PR #11157, opened 2022-01-06
Issue 6320 2

rnveach:
Draft Issue #6320

Start of PR to looking into cases that regression couldn't find. CI is expected to fail under all mutations are handled.

Coding report: http://rveach.no-ip.org/checkstyle/pitest/40/pitest-coding/com.puppycrawl.tools.checkstyle.checks.coding/index.html

Issue #6320, opened 2018-12-27
Pitest: add new mutator REMOVE_CONDITIONALS


rnveach:
This is not an unstable mutator, just disabled by default.

http://pitest.org/quickstart/mutators/#REMOVE_CONDITIONALS

> The remove conditionals mutator will remove all conditionals statements such that the guarded statements always execute
Although not currently enabled by default it is highly recommended that you enable it if you wish to ensure your test suite has full coverage of conditional statements.

**Please note:**
> The reason these are not enabled by default is that there is a large degree of overlap in the tests required to kill these mutations and those required to kill mutations from other default operators such as the conditional boundaries mutator.

Report with only new mutator added, and others disabled: http://rveach.no-ip.org/checkstyle/regression/pitest-reports/30/

This mutator increased the number of timeouts by 6x. And increased the number of memory errors by 93.

Every profile requires 10-15 percent of mutations to be killed, so adding this mutator will take some work. It might be recommended to work by profile instead of doing the entire mutator at once.

-------

<a href="https://www.bountysource.com/issues/68022894-pitest-add-new-mutator-remove_conditionals">
<img src="https://api.bountysource.com/badge/issue?issue_id=68022894">
</a>

rnveach:
@timurt Can you help with killing mutations in xpath?
Specifically:
http://rveach.no-ip.org/checkstyle/regression/pitest-reports/30/pitest-xpath/com.puppycrawl.tools.checkstyle.xpath/ElementNode.java.html
http://rveach.no-ip.org/checkstyle/regression/pitest-reports/30/pitest-xpath/com.puppycrawl.tools.checkstyle.xpath/XpathQueryGenerator.java.html#org.pitest.mutationtest.report.html.SourceFile@2125086c_229

timurt:
@rnveach 

> @timurt Can you help with killing mutations in xpath?

Do you still need help?

romani:
ignore lists were updated to ignore all existing survived lines, only min few cases mutator stay commented out as it is finds a lot uncovered items.
Further process will require covering a line and removing item from list.

rnveach:
New report with only the 1 mutator active: http://rveach.no-ip.org/checkstyle/regression/pitest-reports/39/

Coding package only dropped by 15.

romani:
I see REMOVE_CONDITIONALS activated in all our pitest config in pom.xml, why after such massive fixes, there were no changes in ignore list ?

rnveach:
[...]

PR #11160, opened 2022-01-07
Issue #6320: added test case for EqualsAvoidNullCheck

rnveach:
Issue #6320
Identified at #11157 .

Field needs to come at end of record definition instead of before.

Latest run at https://github.com/checkstyle/checkstyle/runs/4733793364 shows mutations gone with this.


rnveach:
Minor corrections to get the file to compile. Doesn't impact mutations.

romani (approved):

PR #11162, opened 2022-01-07
Issue #6320: resolution for MagicNumberCheck

rnveach:
Issue #6320
Identified at #11157 .

http://rveach.no-ip.org/checkstyle/pitest/40/pitest-coding/com.puppycrawl.tools.checkstyle.checks.coding/MagicNumberCheck.java.html#org.pitest.mutationtest.report.html.SourceFile@6bbc669_489
http://rveach.no-ip.org/checkstyle/pitest/40/pitest-coding/com.puppycrawl.tools.checkstyle.checks.coding/MagicNumberCheck.java.html#org.pitest.mutationtest.report.html.SourceFile@6bbc669_570

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MagicNumberCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MagicNumberCheck.java
@@ -483,28 +483,25 @@ public class MagicNumberCheck extends AbstractCheck {
      * @return {@code true} if {@code ast} is in the scope of a valid hash code method.
      */
     private static boolean isInHashCodeMethod(DetailAST ast) {
+        // find the method definition AST
+        DetailAST methodDefAST = ast.getParent();
+        while (methodDefAST != null
+                && methodDefAST.getType() != TokenTypes.METHOD_DEF) {
+            methodDefAST = methodDefAST.getParent();
+        }
+
         boolean inHashCodeMethod = false;
 
-        // if not in a code block, can't be in hashCode()
-        if (ScopeUtil.isInCodeBlock(ast)) {
-            // find the method definition AST
-            DetailAST methodDefAST = ast.getParent();
-            while (methodDefAST != null
-                    && methodDefAST.getType() != TokenTypes.METHOD_DEF) {
-                methodDefAST = methodDefAST.getParent();
-            }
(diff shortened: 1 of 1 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MagicNumberCheck.java
```diff
@@ -567,14 +564,9 @@ public void setConstantWaiverParentToken(String... tokens) {
      * @param list list of numbers to ignore.
      */
     public void setIgnoreNumbers(double... list) {
-        if (list.length == 0) {
```
rnveach, line 570: This is pretty self-explanatory, both create an empty list and are not different at all.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MagicNumberCheck.java
```diff
@@ -485,26 +485,23 @@ else if (parent.getType() == TokenTypes.UNARY_PLUS) {
     private static boolean isInHashCodeMethod(DetailAST ast) {
         boolean inHashCodeMethod = false;
 
-        // if not in a code block, can't be in hashCode()
-        if (ScopeUtil.isInCodeBlock(ast)) {
```
rnveach, line 489: One of things `isInCodeBlock` does is check if the AST is in a method definition. It does not check that any other thing takes priority, like AST could be a field and it will not stop on a class or record. So the next part of the code is always looking for a method definition, which is also what `isInCodeBlock` looks for. Doesn't care if we pass another class definition (anonymous class) to get to said method definition. As long as it finds a method definition, we can check if its a hash code method or not. The code is `null` safe already that if it scans too far it won't NPE.

So there isn't a way to kill this mutation as it is basically duplicating existing code and is null safe.

I think some this might point to an oversight in the check like are we really in a hash code method if its part of a field in the anonymous class inside the hash code method.

rnveach:
Regression will be at http://rveach.no-ip.org/checkstyle/regression/304/ when it is done.

rnveach:
Minor change for TeamCity:
> 486: isInHashCodeMethod() Scope of variable inHashCodeMethod is too broad

romani (approved):

strkkk (approved):

PR #11165, opened 2022-01-08
Issue #6320: resolution for IllegalTypeCheck

rnveach:
Issue #6320
Identified at #11157 .

http://rveach.no-ip.org/checkstyle/pitest/40/pitest-coding/com.puppycrawl.tools.checkstyle.checks.coding/IllegalTypeCheck.java.html#org.pitest.mutationtest.report.html.SourceFile@56791346_450
http://rveach.no-ip.org/checkstyle/pitest/40/pitest-coding/com.puppycrawl.tools.checkstyle.checks.coding/IllegalTypeCheck.java.html#org.pitest.mutationtest.report.html.SourceFile@56791346_521
http://rveach.no-ip.org/checkstyle/pitest/40/pitest-coding/com.puppycrawl.tools.checkstyle.checks.coding/IllegalTypeCheck.java.html#org.pitest.mutationtest.report.html.SourceFile@56791346_670
http://rveach.no-ip.org/checkstyle/pitest/40/pitest-coding/com.puppycrawl.tools.checkstyle.checks.coding/IllegalTypeCheck.java.html#org.pitest.mutationtest.report.html.SourceFile@56791346_801

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
@@ -445,12 +445,7 @@ public final class IllegalTypeCheck extends AbstractCheck {
     @Override
     public void beginTree(DetailAST rootAST) {
         illegalShortClassNames.clear();
-
-        for (String s : illegalClassNames) {
-            if (s.indexOf('.') == -1) {
-                illegalShortClassNames.add(s);
-            }
-        }
+        illegalShortClassNames.addAll(illegalClassNames);
     }
 
     @Override
@@ -518,13 +513,11 @@ public final class IllegalTypeCheck extends AbstractCheck {
      */
     private boolean isContainVerifiableType(DetailAST modifiers) {
         boolean result = false;
-        if (modifiers.getFirstChild() != null) {
-            for (DetailAST modifier = modifiers.getFirstChild(); modifier != null;
-                     modifier = modifier.getNextSibling()) {
-                if (memberModifiers.contains(modifier.getType())) {
-                    result = true;
-                    break;
-                }
+        for (DetailAST modifier = modifiers.getFirstChild(); modifier != null;
+                 modifier = modifier.getNextSibling()) {
+            if (memberModifiers.contains(modifier.getType())) {
+                result = true;
+                break;
             }
         }
         return result;
@@ -667,7 +660,7 @@ public final class IllegalTypeCheck extends AbstractCheck {
             if (child.getType() == TokenTypes.IDENT) {
                 checkIdent(child);
             }
-            else if (child.getType() == TokenTypes.TYPE_ARGUMENTS) {
+            else {
                 TokenUtil.forEachChild(child, TokenTypes.TYPE_ARGUMENT, this::checkType);
             }
             child = child.getNextSibling();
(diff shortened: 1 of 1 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
```diff
@@ -798,12 +794,10 @@ private static String getImportedTypeCanonicalName(DetailAST importAst) {
         DetailAST toVisitAst = currentNode.getFirstChild();
         while (toVisitAst == null) {
             toVisitAst = currentNode.getNextSibling();
-            if (toVisitAst == null) {
```
rnveach, line 801: The if statement is pointless and so it is safe for it to be removed. Nothing inside the if manipulated `toVisit` which is also the result of this method and nothing inside the method impacts the loop or the result enough that they matter if they are executed or not, even if `toVisit` is not `null` which will end the loop anyways.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
```diff
@@ -667,7 +663,7 @@ private void checkBaseTypes(DetailAST clause) {
             if (child.getType() == TokenTypes.IDENT) {
                 checkIdent(child);
             }
-            else if (child.getType() == TokenTypes.TYPE_ARGUMENTS) {
+            else {
```
rnveach, line 666: The code inside the if statement is checking the children of the child token. When this token is not type arguments, it is specifically comma tokens which have no children. The code in the if is examining the children of the child token, which makes it not matter what is given to it since they have no children. If something ever does have children, we are looking for a very specific child which again contributes to that it will never been found.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
```diff
@@ -518,13 +516,11 @@ private boolean isVerifiable(DetailAST methodOrVariableDef) {
      */
     private boolean isContainVerifiableType(DetailAST modifiers) {
         boolean result = false;
-        if (modifiers.getFirstChild() != null) {
```
rnveach, line 521: The for loop inside this if is already coded that it ends the loop when this same variable returned is `null`, so it is very safe to remove this.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
```diff
@@ -447,9 +447,7 @@ public void beginTree(DetailAST rootAST) {
         illegalShortClassNames.clear();
 
         for (String s : illegalClassNames) {
-            if (s.indexOf('.') == -1) {
```
rnveach, line 450: This is one of the more controversial changes IMO, but I agree it will not have an impact and I am seeing no way to kill the mutation. This if statement prevents fully qualified classes being put into `illegalShortClassNames` but we strictly only match short class names to this map, so there is no way to kill this mutation so we can't check against what won't happen.

rnveach:
Report will be at http://rveach.no-ip.org/checkstyle/regression/306/ when it is done.

romani (approved):
Ok, lets do it

pbludov (approved):

Issue #11140, opened 2022-01-01
Solve fb-contrib errors


pbludov:
[ ] ACEM_ABSTRACT_CLASS_EMPTY_METHODS
[ ] AI_ANNOTATION_ISSUES_NEEDS_NULLABLE
[ ] CE_CLASS_ENVY
[ ] CFS_CONFUSING_FUNCTION_SEMANTICS
[ ] CLI_CONSTANT_LIST_INDEX
[ ] CNC_COLLECTION_NAMING_CONFUSION
[ ] COM_COPIED_OVERRIDDEN_METHOD
[ ] CVAA_CONTRAVARIANT_ELEMENT_ASSIGNMENT
[ ] DLC_DUBIOUS_LIST_COLLECTION
[ ] EXS_EXCEPTION_SOFTENING_NO_CHECKED
[ ] EXS_EXCEPTION_SOFTENING_NO_CONSTRAINTS
[ ] FCCD_FIND_CLASS_CIRCULAR_DEPENDENCY
[ ] IMC_IMMATURE_CLASS_BAD_SERIALVERSIONUID
[ ] IMC_IMMATURE_CLASS_NO_EQUALS
[ ] IMC_IMMATURE_CLASS_NO_TOSTRING
[ ] IMC_IMMATURE_CLASS_PRINTSTACKTRACE
[ ] ITC_INHERITANCE_TYPE_CHECKING
[ ] LII_LIST_INDEXED_ITERATING
[ ] MOM_MISLEADING_OVERLOAD_MODEL
[ ] MUI_CONTAINSKEY_BEFORE_GET
[ ] NAB_NEEDLESS_BOOLEAN_CONSTANT_CONVERSION
[ ] NFF_NON_FUNCTIONAL_FIELD
[ ] NMCS_NEEDLESS_MEMBER_COLLECTION_SYNCHRONIZATION
[ ] OCP_OVERLY_CONCRETE_PARAMETER
[ ] OI_OPTIONAL_ISSUES_USES_IMMEDIATE_EXECUTION
[ ] OPM_OVERLY_PERMISSIVE_METHOD
[ ] PMB_INSTANCE_BASED_THREAD_LOCAL
[ ] PMB_POSSIBLE_MEMORY_BLOAT
[ ] PRMC_POSSIBLY_REDUNDANT_METHOD_CALLS
[ ] PSC_PRESIZE_COLLECTIONS
[...]

PR #11167, opened 2022-01-09
Issue #11140: resolve NAB_NEEDLESS_BOOLEAN_CONSTANT_CONVERSION

rnveach:
Issue #11140

NAB_NEEDLESS_BOOLEAN_CONSTANT_CONVERSION

>NAB_NEEDLESS_BOOLEAN_CONSTANT_CONVERSION
This method assigns a Boxed boolean constant to a primitive boolean variable, or assigns a primitive boolean constant to a Boxed boolean variable. Use the correct constant for the variable desired. Use boolean b = true; boolean b = false; 
or Boolean b = Boolean.TRUE; Boolean b = Boolean.FALSE; Be aware that this boxing happens automatically when you might not expect it. For example, Map statusMap = ... public Boolean someMethod() { statusMap.put("foo", true); //the "true" here is boxed return false; //the "false" here is boxed } 
has two cases of this needless autoboxing. This can be made more efficient by simply substituting in the constant values: Map statusMap = ... public Boolean someMethod() { statusMap.put("foo", Boolean.TRUE); return Boolean.FALSE; }


rnveach:
Cirrus failure doesn't look related.

romani:
@rnveach , methods return the snot Boolean. please explain why autoboxing happening ? And why violation is valid?

rnveach:
> please explain why autoboxing happening ? And why violation is valid?

Autoboxing happens when you are passing a value somewhere and the receiver is expecting an Object and it is being given a primitive type. What Java does behind the scenes, known as autoboxing, is it converts that primitive to an Object to satisfy the declaration of the receiver. Auto-unboxing is the reverse, where the receiver accepts a primitive and it is being sent an Object. I do not know directly how autoboxing happens underneath in the actual implementation, but it happens and most likely needs some extra cycles to do the conversion, especially with an dynamic variable.

In these examples, the methods accept Objects and we are sending them primitives.

Violations are valid in that the changes necessary for the violations to go away are simple and do make the code more easier to understand that it is accepting an Object and not a primitive. All these cases are literals (true/false), so there is no necessary and ugly force boxing needed like `new Boolean(value)` or `Boolean.valueOf(value)`. I would not support ugly boxing like those mentioned in the previous statement. We also gain back those few cycles from the implementation doing the autoboxing.

romani:
Takes object https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#orElse-T-

PR #11177, opened 2022-01-11
Issue #11163: Enforced file size on inputs

### LINKED ISSUE #11163: Enforce file size on Java inputs
From https://github.com/checkstyle/checkstyle/pull/11156#pullrequestreview-845293639

We need to get statistics on how long our Input files and get median value and enforce file size to not above this value.

As we usually do:
Enforce rules with long list of suppression
All suppressions can be resolved in separate PRs

Rationale: big input files a hard to review, hard to debug.

Vyom-Yadav:
Resolves #11163: enforce file size on inputs


src/test/java/com/puppycrawl/tools/checkstyle/AbstractPathTestSupport.java
```diff
@@ -24,11 +24,834 @@
...
 public abstract class AbstractPathTestSupport {
 
+    /**
+     * Input files which have more than 120 lines.
+     */
+    public static final Set<String> FILE_SIZE_SUPPRESSIONS = new HashSet<>(Arrays.asList(
+            "src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc"
+                    + "/nonemptyatclausedescription/InputNonEmptyAtclauseDescription.java",
+            "src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocstyle"
+                    + "/InputJavadocStyle2.java",
```
Vyom-Yadav, line 46: The complete path was used as we have files with the same name.
Example - 
https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/grammar/InputRegressionJavaClass1.java
https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/grammar/antlr4/InputRegressionJavaClass1.java

src/test/java/com/puppycrawl/tools/checkstyle/AbstractPathTestSupport.java
```diff
@@ -110,4 +355,58 @@ protected static String toLfLineEnding(String text) {
...
     }
 
+    /**
+     * Validate that all input files do not exceed specified max length.
+     *
+     * @param theFiles list of files
+     * @throws CheckstyleException if file length is more than max length
+     * @throws IOException if I/O exception occurs while reading
+     */
+    protected static void validateFileLength(
```
rnveach, line 365: Why not just use https://checkstyle.org/config_sizes.html#FileLength ?

Vyom-Yadav: I didn't notice that check earlier, updated it to use it now, also please restart appveyor

rnveach: Restarted but other jobs are failing.

Vyom-Yadav: Yup, just a moment, I filtered for `.java` files, adding other extensions too.

config/checkstyle_resources_suppressions.xml
```diff
@@ -167,4 +167,906 @@
   <suppress checks="PackageDeclarationCheck"
             files="[\\/]InputAstTreeStringPrinterTextBlocksEscapesAreOneChar\.java"/>
 
+  <!-- Until https://github.com/checkstyle/checkstyle/issues/11163 -->
+  <suppress checks="FileLength"
+    files="[\\/]InputNonEmptyAtclauseDescription\.java"/>
```
romani, line 172: Please keep in path name of Check, it will help a lot when we start to resolve such suppression, as we will do them check by check

romani (changes requested):
Item

config/checkstyle_resources_suppressions.xml
```diff
@@ -167,4 +167,906 @@
...
+  <suppress checks="FileLength"
+    files="[\\/]expectedNonTightRbHtmlTagAst\.txt"/>
+  <suppress checks="FileLength"
+    files="[\\/]expectedNonTightRtcHtmlTagAst\.txt"/>
+  <suppress checks="FileLength"
+    files="[\\/]expectedAllStandardJavadocTagsAst\.txt"/>
+  <suppress checks="FileLength"
+    files="[\\/]expectedUppercaseInPackageNameAst\.txt"/>
+  <suppress checks="FileLength"
+    files="[\\/]expectedSpaceBeforeDescriptionInBlockJavadocTagsAst\.txt"/>
```
rnveach, line 1070: I would think we would not make violations on expected files. It is direct output of the input after going through the Checkstyle process. If the input file is flagged to be trimmed, then it will already affect the expected file. But if we are flagging only the expected file, then it may need extra scrutiny if we can even fix the issue.

@romani ping

Vyom-Yadav: > If the input file is flagged to be trimmed, then it will already affect the expected file.

I think that we would have to suppress expected files irrespective of whether the input file for the Checkstyle process is trimmed or not. The input file can be within the limit but the printed AST can exceed the length of 120.
Example - 
https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/asttreestringprinter/ExpectedAstTreeStringPrinterFullOfSinglelineComments.txt
The [input](https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/asttreestringprinter/InputAstTreeStringPrinterFullOfSinglelineComments.java) file is under the limit but the output file is not.

rnveach: It depends on the type of output, but it may be a possibility we can trim the input file more so the output is within the size too. This is why I pinged @romani to confirm.

romani: Yes, we need to exclude txt files, we need to validate only java files.

rnveach: For extra clarification after talking to @romani , we will only do Java files as part of this issue since this is the main chunk of our inputs. We will handle non-Java input files in a new issue. We will exclude "Expected" or "Output" files for now.

nmancus1: We can just use `fileExtensions` property of FileLengthCheck until #11178

romani: https://checkstyle.org/config_sizes.html#FileLength_Properties yes, we can.

config/checkstyle_resources_suppressions.xml
Vyom-Yadav: Done

config/checkstyle_resources_suppressions.xml
Vyom-Yadav: Done

romani (approved):
Ok to merge

rnveach (approved):

rnveach:
> Resolves #11163

This does not resolve the issue since all suppressions are tied to this issue.

PR #11181, opened 2022-01-12
Issue #11140: resolve CNC_COLLECTION_NAMING_CONFUSION

rnveach:
Issue #11140

>CNC_COLLECTION_NAMING_CONFUSION This class defines a field or local collection variable with a name that contains a different type of collection in its name. An example would be a Set called userList. This is confusing to the reader, and likely caused by a previous refactor of type, without changing the name. This detector is obviously only checking for English names.


rnveach:
Azure seems stuck but Failing CI does not look related to the changes made here

romani (approved):

Issue #10848, opened 2021-10-06
Support patterns in switch (preview-feature in Java 17) as described by JEP 406


sviperll:
I have downloaded the latest cli from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

**How it works Now:**

```bash
$ java -version
openjdk version "17" 2021-09-14
OpenJDK Runtime Environment 21.9 (build 17+35)
OpenJDK 64-Bit Server VM 21.9 (build 17+35, mixed mode, sharing)

$ javac --version
javac 17

$ javac --release 17 --enable-preview Main.java
Note: Main.java uses preview features of Java SE 17.
Note: Recompile with -Xlint:preview for details.

$ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
  <property name="severity" value="error"/>
  <property name="fileExtensions" value="java, properties, xml"/>
  <module name="TreeWalker">
    <module name="ConstantName"/>
  </module>
</module>

$ cat Main.java
package com.github.sviperll.checkstylejep406;

public class Main {
    public static void main(String[] args) {
        Object value = Integer.valueOf(42);
        String result = switch (value) {
            case null -> "It's null...";
            case String s -> "It's a string: " + s;
            case Integer i -> "It's an integer: " + i.toString();
            case Object v -> "It's something else: " + v.toString();
        };
        System.out.println(result);
    }
}

$ RUN_LOCALE="-Duser.language=en -Duser.country=US"
$ java $RUN_LOCALE -jar checkstyle-9.0.1-all.jar -c config.xml Main.java
Starting audit...
com.puppycrawl.tools.checkstyle.api.CheckstyleException: Exception was thrown while processing Main.java
	at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:302)
	at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:221)
	at com.puppycrawl.tools.checkstyle.Main.runCheckstyle(Main.java:409)
	at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:332)
	at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:191)
[...]

PR #11100, opened 2021-12-23
Issue #10848: Support patterns in switch

nmancus1:
Closes #10848.

Relevant grammar from https://docs.oracle.com/javase/specs/jls/se17/preview/specs/patterns-switch-jls.html:

```

SwitchBlock:
    { SwitchRule {SwitchRule} }
    { {SwitchBlockStatementGroup} {SwitchLabel :} }

SwitchRule:
    SwitchLabel -> Expression ;
    SwitchLabel -> Block
    SwitchLabel -> ThrowStatement

SwitchBlockStatementGroup:
    SwitchLabel : BlockStatements

SwitchLabel:
    CaseOrDefaultLabel {: CaseOrDefaultLabel }

CaseOrDefaultLabel:
    case CaseLabelElement {, CaseLabelElement }
    default

CaseLabelElement:
    CaseConstant
    Pattern
    null
    default

```

---------------------------------------
[Relevant grammar for patterns](https://openjdk.java.net/jeps/406):

```
Pattern:
  PrimaryPattern
  GuardedPattern

GuardedPattern:
  PrimaryPattern && ConditionalAndExpression

PrimaryPattern:
  TypePattern
  ( Pattern )

```

------------------------------------------
Also, minor update to `instanceof`:

From https://openjdk.java.net/jeps/406:
```
We also change the grammar for instanceof expressions to:

InstanceofExpression:
  RelationalExpression instanceof ReferenceType
  RelationalExpression instanceof PrimaryPattern

```

------------------------------------

Just a test here, I noticed that PMD release in projects files is not up to date.

Diff Regression projects: https://gist.githubusercontent.com/nmancus1/ea9cd53c4dc4be8be3e776d059542f00/raw/6eb067547d67808e3bbab428a4e2dad7a031676d/pmd.properties

Diff Regression config: https://gist.githubusercontent.com/nmancus1/3db3e23a5063857160d7d92cbdfa1bfe/raw/a200e0fb1c4700ecd084f9e0b03bf7dbb9694832/config.xml

---------------------------

## ANTLR regression report

https://nmancus1.github.io/issue-10848/issue-10848-antlr-diff-pmd-checkstyle/index.html

All diffs are exceptions vs files we can parse now.
______________________________________________

## OpenJDK16 Check Regression Reports

https://nmancus1.github.io/issue-10848/issue-10848_check_diff_reports_2021_12_30_openjdk16/diff_checks-only-javadoc-error/index.html

[...]

Issue #5124, opened 2017-09-18
Remove usage of DetailAST.branchContains


rnveach:
Identified at https://github.com/checkstyle/checkstyle/issues/4727#issuecomment-315624202 ,

> branchContains is a dangerous method to use all the time since there is no restriction on how deep it can search.

The referenced issue fixed a bug where we used `branchContains` and it was scanning deeper than the check originally intended. It was supposed to be looking for a variable with a `final` modifier, but because of `branchContains` it scanned into an anonymous class in the assignment and picked up the `final` there.

There may be other cases in our code where it might be searching too much.
Example:
https://github.com/checkstyle/checkstyle/blob/55aa8cedc60d4faa776d7a36f8ae2d0bed0ada45/src/main/java/com/puppycrawl/tools/checkstyle/checks/FinalParametersCheck.java#L144
Because we are doing a `branchContains` on the `METHOD_DEF` it is also searching inside the `SLIST` of the method and may pick-up anonymous classes while it is looking for `PARAMETER_DEF`.

I feel `branchContains` should be completely removed as it's search area can't be controlled. If we don't want to remove it, it's use should be heavily controlled.

We should swap it's current uses with either `findFirstToken` or a new method in each check that specifically controls where and what we are looking for.

romani:
this method just do search in tree, nothing related to any logic of any Check.

user can launch this method of root and receive even more interesting results, that probably not up to its expectations.

This looks like just simple mis-usage of some method by engineer.
We can probably move it Utils , but it can be misused from there too.

Please share more points to support removal of it.
Do you think that in all use cases it is error-prone ?

rnveach:
@romani I will check again, but when I looked the majority of cases where calling the method on modifiers looking for static/public/etc... Unless we can define those modifiers inside or on an annotation, I don't think there are other cases to support.

romani:
ones we come to immutable AST model, such method will go utils for sure.

rnveach:
@romani It can't really go outside the AST class because it has to build up all the children trees recursively and caches them through `getBranchTokenTypes`.
If we were to remove the cache and place it inside a `util` out of the AST, then this is even more reason to just remove it. Without the cache, this method will be even slower if used in areas with deep trees.

romani:
ok, let start this process

rnveach:
Only 3 checks remain to be fixed.
AnnotationUseStyle, EqualsAvoidNull, DesignForExtension.

[...]

PR #11188, opened 2022-01-12
Issue # 11187: OperatorWrapCheck throws NPE on guarded patterns

### LINKED ISSUE #11187: `OperatorWrapCheck` throws NPE on guarded patterns
Link to check documentation: https://checkstyle.sourceforge.io/config_whitespace.html#OperatorWrap

Noticed in https://github.com/checkstyle/checkstyle/pull/11186  at https://github.com/checkstyle/checkstyle/runs/4785736479?check_suite_focus=true :

```

➜  javac --enable-preview --release 17 Test.java             
Note: Test.java uses preview features of Java SE 17.

➜  cat Test.java                                             
public class Test {
    String typeGuardAfterParenthesizedTrueSwitchStatement(Object o) {
        switch (o) {
            case (Integer i) && i == 0: o = String.valueOf(i); return "true";
            case ((Integer i) && i == 2): o = String.valueOf(i); return "second";
            case Object x: return "any";
        }
    }

    String typeGuardAfterParenthesizedTrueSwitchExpression(Object o) {
        return switch (o) {
            case (Integer i) && i == 0: o = String.valueOf(i); yield "true";
            case ((Integer i) && i == 2): o = String.valueOf(i); yield "second";
            case Object x: yield "any";
        };
    }

    String typeGuardAfterParenthesizedTrueIfStatement(Object o) {
        if (o != null && o instanceof ((Integer i) && i == 0)) {
            return "true";
        } else if (o != null && o instanceof (((Integer i) && i == 2)) && (o = i) != null) {
            return "second";
        } else {
            return "any";
        }
    }
}

➜  cat config.xml                                            
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="OperatorWrap">
        </module>
    </module>
</module>

➜  java -jar checkstyle-9.3-SNAPSHOT-all.jar -c config.xml Test.java
Starting audit...
com.puppycrawl.tools.checkstyle.api.CheckstyleException: Exception was thrown while processing Test.java
        at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:305)
        at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:224)
        at com.puppycrawl.tools.checkstyle.Main.runCheckstyle(Main.java:407)
        at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:330)
        at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:189)
        at com.puppycrawl.tools.checkstyle.Main.main(Main.java:126)
Caused by: java.lang.NullPointerException: Cannot invoke "com.puppycrawl.tools.checkstyle.api.DetailAST.getType()" because "result" is null
        at com.puppycrawl.tools.checkstyle.checks.whitespace.OperatorWrapCheck.adjustParens(OperatorWrapCheck.java:452)
        at com.puppycrawl.tools.checkstyle.checks.whitespace.OperatorWrapCheck.getLeftNode(OperatorWrapCheck.java:394)
        at com.puppycrawl.tools.checkstyle.checks.whitespace.OperatorWrapCheck.isNewLineModeViolation(OperatorWrapCheck.java:345)
        at com.puppycrawl.tools.checkstyle.checks.whitespace.OperatorWrapCheck.visitToken(OperatorWrapCheck.java:308)
        at com.puppycrawl.tools.checkstyle.api.AbstractFileSetCheck.process(AbstractFileSetCheck.java:87)
        at com.puppycrawl.tools.checkstyle.Checker.processFile(Checker.java:331)
        at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:292)
        ... 5 more


```

I would expect `OperatorWrapCheck` to process `Test.java` without throwing a null pointer exception.

We can remove file filter for this file from `openjdk17-excluded.files` in PR for this issue, and exclusion from any project files in contribution.

nmancus1:
Closes #11187

Diff Regression config: https://gist.githubusercontent.com/nmancus1/284f66dbf5a076b43708b7685ee9bd53/raw/c595fc26ae622509dadcd25ca6fe02b3db5ffaac/config.xml

Diff Regression projects: https://raw.githubusercontent.com/checkstyle/contribution/e34312d61da0b073f31339cfbf385f289b721fed/checkstyle-tester/projects-to-test-on-for-github-action.properties

Final report: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/586d278_2022024156/reports/diff/index.html

No diffs now that #10848 is closed.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheck.java
@@ -390,6 +390,9 @@ public class OperatorWrapCheck
         if (node.getFirstChild() == null || isAssignToVariable(node)) {
             result = node.getPreviousSibling();
         }
+        else if (isInPatternDefinition(node)) {
+            result = node.getFirstChild();
+        }
         else {
             result = adjustParens(node.getFirstChild(), DetailAST::getNextSibling);
         }
@@ -397,6 +400,30 @@ public class OperatorWrapCheck
             result = result.getLastChild();
         }
         return result;
+    }
+
+    /**
+     * Ascends AST to determine if given node is part of a pattern
+     * definition.
+     *
+     * @param node the node to check
+     * @return true if node is in pattern definition
+     */
+    private static boolean isInPatternDefinition(DetailAST node) {
+        DetailAST parent = node;
+        final int[] tokensToStopOn = {
+            // token we are looking for
+            TokenTypes.PATTERN_DEF,
+            // tokens that mean we can stop looking
+            TokenTypes.EXPR,
+            TokenTypes.RESOURCE,
+            TokenTypes.COMPILATION_UNIT,
+        };
+
+        do {
+            parent = parent.getParent();
+        } while (!TokenUtil.isOfType(parent, tokensToStopOn));
+        return parent.getType() == TokenTypes.PATTERN_DEF;
     }
 
     /**
--- a/.ci/openjdk17-excluded.files
+++ b/.ci/openjdk17-excluded.files
@@ -29,11 +29,6 @@
   <property name="fileNamePattern" value="[\\/]test[\\/]langtools[\\/]tools[\\/]javac[\\/]unicode[\\/]FirstChar2.java$"/>
 </module>
 
-<!-- until https://github.com/checkstyle/checkstyle/issues/11187-->
-<module name="BeforeExecutionExclusionFileFilter">
-  <property name="fileNamePattern" value="[\\/]test[\\/]langtools[\\/]tools[\\/]javac[\\/]patterns[\\/]Guards.java$"/>
-</module>
-
 <!-- non-compilable -->
 <module name="BeforeExecutionExclusionFileFilter">
   <property name="fileNamePattern" value="[\\/]test[\\/]langtools[\\/]tools[\\/]javac[\\/]unicode[\\/]SupplementaryJavaID1.java$"/>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheckTest.java
@@ -139,6 +139,31 @@ public class OperatorWrapCheckTest
     }
 
     @Test
+    public void testGuardedPattern() throws Exception {
+        final String[] expected = {
+            "52:30: " + getCheckMessage(MSG_LINE_NEW, "&&"),
+            "54:31: " + getCheckMessage(MSG_LINE_NEW, "&&"),
+            "62:30: " + getCheckMessage(MSG_LINE_NEW, "&&"),
+            "64:31: " + getCheckMessage(MSG_LINE_NEW, "&&"),
+            "71:23: " + getCheckMessage(MSG_LINE_NEW, "&&"),
+            "72:43: " + getCheckMessage(MSG_LINE_NEW, "&&"),
+            "75:30: " + getCheckMessage(MSG_LINE_NEW, "&&"),
+            "76:44: " + getCheckMessage(MSG_LINE_NEW, "&&"),
+            "77:34: " + getCheckMessage(MSG_LINE_NEW, "&&"),
+        };
+
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputOperatorWrapGuardedPatterns.java"), expected);
+    }
+
+    @Test
+    public void testTryWithResources() throws Exception {
+        final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputOperatorWrapTryWithResources.java"), expected);
+    }
+
+    @Test
     public void testInvalidOption() throws Exception {
 
         try {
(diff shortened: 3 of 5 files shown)


nmancus1:
I am leaving this is draft until https://github.com/checkstyle/checkstyle/pull/11100 is merged

nmancus1:
Github, generate report

nmancus1:
CI failure is not related: `Connect to www.w3.org:80 [www.w3.org/128.30.52.100] failed: Connection timed out`

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheck.java
```diff
@@ -399,6 +402,21 @@ private static DetailAST getLeftNode(DetailAST node) {
...
+     * Ascends AST to determine if given node is part of a pattern
+     * definition.
+     *
+     * @param node the node to check
+     * @return true if node is in pattern definition
+     */
+    private static boolean isInPatternDefinition(DetailAST node) {
+        DetailAST parent = node.getParent();
+        while (parent != null && parent.getType() != TokenTypes.PATTERN_DEF) {
+            parent = parent.getParent();
```
rnveach, line 415: why not?
````
DetailAST parent = node;
do {
    parent = parent.getParent();
} while (parent != null && parent.getType() != TokenTypes.PATTERN_DEF);
````

Also we are passing through parents uncontrolled. It doesn't matter if we pass any other kind of def before hitting the pattern def?

nmancus1: > why not...

Please explain the advantage of this; to me the name of variable declared (`parent`) and initialization (`node.getParent();`) makes sense. Since we are checking operands, there is absolutely no chance that `node` is null (it will always have an operator as a parent).

>  It doesn't matter if we pass any other kind of def before hitting the pattern def?

To the check, no. And a local variable(pattern definition) will always have an enclosing block of some sort, so we certainly could just check for a method defintion, and stop there. But, I generally avoid this sort of "optimization" since pitest normally fails on this sort of thing.

romani: I do not know why, but my brain not always work quick on do-while structure. I prefer while loop in this case.

romani (approved):
last commit is ok to merge

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheck.java
rnveach: Not a big deal in this case, but I see the duplication of `getParent` and possibly an extra unnecessary condition check at the start of the loop and I think how to optimize it more. Hence the do-while version.

rnveach: > On second thought, I might be able to create a case to make this fail.

I am assuming this item is still TODO until told otherwise.

nmancus1: >On second thought, I might be able to create a case to make this fail.

Disregard this.

Replaced while with do-while.

nmancus1:
Github, generate report

nmancus1:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheck.java
rnveach: So I saw the update, but I am still not really seeing a conclusion to this item:
> It doesn't matter if we pass any other kind of def before hitting the pattern def?

For example, it doesn't matter if we pass through an anonymous class to get to the pattern def or patterns are such that we are either in it or not for what we stop on.

nmancus1: > anonymous class to get to the pattern def 

It is not possible to declare an anonymous class within a pattern definition.

>  patterns are such that we are either in it or not for what we stop on.

Please restate this, I am not sure that I understand what you are asking.

romani: Reminder. This is fix for exception, it should merged sooner. If we think there is some false positive, it is better to address in separate issue.

rnveach: I will approve the PR, but we have seen cases before where we either go too deep in the tree or go past too many key nodes when looking for "something" and we are accidentally pulling in nodes wrongly or passing things that should have stopped us. `branchContains` is one good example. https://github.com/checkstyle/checkstyle/issues/5124#issue-258583861 details the specifics of how it was going to far and wasn't be used carefully enough.

Whenever I see a loop where we are scanning up or down uncontrolled, I get concerned. I don't know the new Java syntax well, so I can't explain a possible situation to prompt a more meaningful conversation. I am just making sure this won't be an issue later.

nmancus1:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheck.java
nmancus1: Condition is updated to exit on pattern def or expression node.

nmancus1:
Github, rebase

romani:
Regression , NPE is detected by CI.

nmancus1:
Github, generate report

nmancus1:
Github, generate report

rnveach (approved):

rnveach:
Fix was merged

Issue #11178, opened 2022-01-11
Enforce file size on Non-Java inputs 


rnveach:
Split from https://github.com/checkstyle/checkstyle/issues/11163 ,

This issue should specifically look at Non-Java input files. An input file is anything named "Input*.*".

As we usually do:
Enforce rules with long list of suppression
All suppressions can be resolved in separate PRs

Rationale: big input files a hard to review, hard to debug.

romani:
If we limit java input file, it will limit ast tree file.
Limit on ast tree file will not be very reliable or useful.
Do you want extra limit on grammar regression files ?

rnveach:
> https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/grammar/InputJava9TryWithResources.txt

https://github.com/checkstyle/checkstyle/blob/77852dd5ccda0d69e3146b701549b8c49d5a02c2/src/test/java/com/puppycrawl/tools/checkstyle/grammar/AstRegressionTest.java#L68

We may have to relook at file names. Based on test above, this looks more like an output file which is outside the scope of this issue.

I know we have javadoc input files for Javadoc Parser.
I know we have XML input files for testing configuration files.

Those type of files that are actual input to our application is what I am targeting. It makes it clearer, we can drop TXT files for now.

romani:
Ok to rename ast files to Outputxxxxxx.ast and they should not be validated.
*.javadoc should be limited to 20 lines.
and xml is limited to 20.

rnveach:
Sounds reasonable, but I cant see if we don't have any stats like the other issue. I want to look at some things and will try later today.

rnveach:
We have the following input files under `test\resources` that are not Java.

````
dtd - 1
header - 14
html - 2
js - 1
properties - 41
sarif - 7
sql - 3
translation - 5
translations - 5
txt - 172 (81 are expected files)
xml - 104 (12 are expected files)
````

Some small statistics:

Total Number of Files: 355
Grand Total Lines: 55,003
Average Lines Per File: 155

rnveach:
> rename ast files to Outputxxxxxx.ast and they should not be validated

The convention I am seeing in the files is "Expected". None use "Output".

romani:
Let's use "Expected" prefix

pree-T:
Is this issue still open?

romani:
[...]

PR #11191, opened 2022-01-12
minor: corrected expected file names in grammar package

rnveach:
Identified at https://github.com/checkstyle/checkstyle/issues/11178#issuecomment-1010923795 , renamed expected files to have expected in the name so it is clear they are not inputs.


config/ant-phase-verify.xml
```diff
@@ -118,6 +118,7 @@
 
           <!-- Exclude because printed AST is verified by matching checkstyle output -->
           <exclude name="**/grammar/**/Input*.txt"/>
+          <exclude name="**/grammar/**/Expected*.txt"/>
```
rnveach, line 121: It took me a while to figure this is why violations were showing up but I wasn't finding a matching suppression.

rnveach: All original inputs should be renamed now so this item is all expected.

romani (approved):

PR #11223, opened 2022-01-20
Issue #6320: Kill surviving mutations for REMOVE_CONDITIONALS in UnusedLocalVariableCheck

Vyom-Yadav:
Part of #6320: Kill surviving mutations for REMOVE_CONDITIONALS in UnusedLocalVariableCheck

Diff Regression projects: https://gist.githubusercontent.com/Vyom-Yadav/91807e6244cff60698d9e33e32ba2d51/raw/432e52446da09d37b6c0def38f08c6a5ca6446a5/projects-to-test-on.properties
Diff Regression config: https://gist.githubusercontent.com/Vyom-Yadav/98dceb63a79f4833e85fff9b2e1464a6/raw/ef336038bce0e1a7af42b9032aa41cd1266a2d89/my_checks.xml

---
pit-report (see active mutators at the end): https://vyom-yadav.github.io/DiffReport/pit-reports/202201221220/com.puppycrawl.tools.checkstyle.checks.coding/UnusedLocalVariableCheck.java.html
Diff report: https://vyom-yadav.github.io/DiffReport/diffUnusedKillMutations/index.html

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
@@ -23,9 +23,12 @@ import java.util.ArrayDeque;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Queue;
+import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
@@ -218,10 +221,10 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
     private final Map<DetailAST, TypeDeclDesc> anonInnerAstToTypeDeclDesc;
 
     /**
-     * Queue of tokens of type {@link UnusedLocalVariableCheck#CONTAINERS_FOR_ANON_INNERS}
+     * Set of tokens of type {@link UnusedLocalVariableCheck#CONTAINERS_FOR_ANON_INNERS}
      * and {@link TokenTypes#LAMBDA} in some cases.
      */
-    private final Queue<DetailAST> anonInnerClassHolders;
+    private final Set<DetailAST> anonInnerClassHolders;
 
     /**
      * Name of the package.
@@ -234,23 +237,15 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
     private int depth;
 
     /**
-     * Block containing local anon inner class, keeps updating throughout the check.
-     * Acts as a tracker and checks whether the current node being visited is present
-     * inside the block containing a local anonymous inner class.
-     */
-    private DetailAST blockContainingLocalAnonInnerClass;
-
-    /**
      * Creates a new {@code UnusedLocalVariableCheck} instance.
      */
     public UnusedLocalVariableCheck() {
         variables = new ArrayDeque<>();
         typeDeclarations = new ArrayDeque<>();
-        typeDeclAstToTypeDeclDesc = new HashMap<>();
+        typeDeclAstToTypeDeclDesc = new LinkedHashMap<>();
         anonInnerAstToTypeDeclDesc = new HashMap<>();
-        anonInnerClassHolders = new ArrayDeque<>();
-        blockContainingLocalAnonInnerClass = null;
-        packageName = "";
+        anonInnerClassHolders = new HashSet<>();
+        packageName = null;
         depth = 0;
     }
 
@@ -300,7 +295,7 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
         typeDeclAstToTypeDeclDesc.clear();
         anonInnerAstToTypeDeclDesc.clear();
         anonInnerClassHolders.clear();
-        packageName = "";
+        packageName = null;
         depth = 0;
     }
 
@@ -316,15 +311,15 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
         else if (type == TokenTypes.IDENT) {
             visitIdentToken(ast, variables);
         }
+        else if (type == TokenTypes.LITERAL_NEW
+                && isInsideLocalAnonInnerClass(ast)) {
+            visitLocalAnonInnerClass(ast);
+        }
         else if (TokenUtil.isTypeDeclaration(type)) {
             visitTypeDeclarationToken(ast);
         }
         else if (type == TokenTypes.PACKAGE_DEF) {
             packageName = extractQualifiedName(ast.getFirstChild().getNextSibling());
-        }
-        else if (type == TokenTypes.LITERAL_NEW
-                && isInsideLocalAnonInnerClass(ast)) {
-            visitLocalAnonInnerClass(ast);
         }
     }
 
@@ -335,9 +330,6 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
         }
         else if (ast.getType() == TokenTypes.COMPILATION_UNIT) {
             leaveCompilationUnit();
-        }
-        else if (ast.equals(blockContainingLocalAnonInnerClass)) {
-            blockContainingLocalAnonInnerClass = null;
         }
         else if (isNonLocalTypeDeclaration(ast)) {
             depth--;
@@ -351,9 +343,8 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
      * @param dotAst dotAst
      * @param variablesStack stack of all the relevant variables in the scope
      */
-    private void visitDotToken(DetailAST dotAst, Deque<VariableDesc> variablesStack) {
-        if (blockContainingLocalAnonInnerClass == null
-                && dotAst.getParent().getType() != TokenTypes.LITERAL_NEW
+    private static void visitDotToken(DetailAST dotAst, Deque<VariableDesc> variablesStack) {
+        if (dotAst.getParent().getType() != TokenTypes.LITERAL_NEW
                 && shouldCheckIdentTokenNestedUnderDot(dotAst)) {
             checkIdentifierAst(dotAst.findFirstToken(TokenTypes.IDENT), variablesStack);
         }
@@ -365,10 +356,8 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
      * @param varDefAst varDefAst
      */
     private void visitVariableDefToken(DetailAST varDefAst) {
-        if (blockContainingLocalAnonInnerClass == null) {
-            addLocalVariables(varDefAst, variables);
-            addInstanceOrClassVar(varDefAst);
-        }
+        addLocalVariables(varDefAst, variables);
+        addInstanceOrClassVar(varDefAst);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


Vyom-Yadav:
rnveach mentioned these mutations-
```
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>                        &#38;&#38; TokenUtil.isTypeDeclaration(parentAst.getParent().getType())) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>                        &#38;&#38; isInsideLocalAnonInnerClass(grandParent);</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>                        || parentAst.getType() == TokenTypes.OBJBLOCK</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>                    .getParent().getParent().getType() == TokenTypes.OBJBLOCK) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>                if (!variableDesc.isInstVarOrClassVar()) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>                if (parentAst.getType() == TokenTypes.ENUM_CONSTANT_DEF</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>            if (packageName.isEmpty()) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>            if (scopeAst == null) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        else if (type == TokenTypes.LITERAL_NEW</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        else if (type == TokenTypes.LITERAL_NEW</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        for (int i = 0; i &#60; length &#38;&#38; patternClass.charAt(i) == classToBeMatched.charAt(i); ++i) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        if (TokenUtil.isOfType(ast, SCOPES)) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        if (blockContainingLocalAnonInnerClass == null) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        if (blockContainingLocalAnonInnerClass == null) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        if (blockContainingLocalAnonInnerClass == null) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        if (blockContainingLocalAnonInnerClass == null</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        if (isNonLocalTypeDeclaration(typeDeclAst)) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        if (outerClassQualifiedName == null) {</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        if (parentAst.getType() == TokenTypes.OBJBLOCK</span></pre></td></tr>
+UnusedLocalVariableCheck.java.html:<td class='covered'><pre><span  class='survived'>        return typeDeclAst.getParent().getType() != TokenTypes.SLIST</span></pre></td></tr>
```
All these mutations have been killed, many of these were the result of micro-optimizations I did and pit cannot judge time optimizations unless you write a specific test for it, usually, `n` will be in 10-20's range so removing these won't make a difference in running time, they will make the code cleaner.

Vyom-Yadav:
> Diff report: https://vyom-yadav.github.io/DiffReport/diffUnusedKillMutations/index.html

Files throwing exceptions are ignored, actual removed exceptions were false positives.

nmancus1:
Please add a test case like https://vyom-yadav.github.io/DiffReport/diffUnusedKillMutations/openjdk16/xref/home/vyom/IdeaProjects/contribution/checkstyle-tester/repositories/openjdk16/test/langtools/tools/javac/lambda/LambdaCapture04.java.html#L58 to avoid future false positives

Vyom-Yadav:
> Please add a test case like https://vyom-yadav.github.io/DiffReport/diffUnusedKillMutations/openjdk16/xref/home/vyom/IdeaProjects/contribution/checkstyle-tester/repositories/openjdk16/test/langtools/tools/javac/lambda/LambdaCapture04.java.html#L58 to avoid future false positives

Added

rnveach:
@Vyom-Yadav Are you able to provide the pit report so we can see its 100% now?


> actual removed ... were false positives

https://vyom-yadav.github.io/DiffReport/diffUnusedKillMutations/openjdk16/index.html#A4
This check was just merged and has not been released so I am fine with keeping the PR as 1 big change if we can merge it sooner, but do note that since this is all going under #6320 that means it will not be reported as a bug on the release notes. This issue is marked as "misc" and no one really reads the "misc" notes since we are basically saying they aren't important.

> pit cannot judge time optimizations

This is basically correct. Pit cannot prove what is needed if it cannot be asserted through what is expected. (IE: How can you prove the sky needs to be blue if you have no requirements document detailing what the color of the sky should be.) Time optimizations basically require a time check in this instance, but it is impossible to describe how "fast" something should be through time, especially for computers. Different machines require different lengths of time to execute things and we can't guarantee the same time each run because of Java or machine threading. 1 second on 1 machine, may be 1.1 seconds on another or 2 seconds on the same machine if it is doing other things.

Usually time optimizations amount to less method calls, so mocks can be done to count number of calls and such as an alternative way. I am not sure I fully like this idea since it is usually hard to see how many times a call is appropriate from the human eye and we just have to rely on the value the machine gives.

My advice, is if something looks like it may be more on the critical side to keep, we make it known and agree to it with a comment to the pitest suppression saying its needed and why.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
```diff
@@ -723,21 +731,22 @@ private String getQualifiedTypeDeclarationName(DetailAST typeDeclAst) {
      * @param className className
      * @return the qualified name of type declaration by combining {@code packageName},
      *         {@code outerClassQualifiedName} and {@code className}
+     * @noinspection CallToStringConcatCanBeReplacedByOperator
```
rnveach, line 734: Why is this change ok? It seems odd.

Vyom-Yadav:
```
if (packageName == null) {
}
```
pit mutated it to false, so 
```
qualifiedClassName = packageName + PACKAGE_SEPARATOR + className;
```
resulted in `null.SomeClass`, so to get an exception I replaced it with `concat` to kill the mutation.
Actually, this method was picked from `FinalClassCheck`, it works for this check also but to kill some mutations some changes need to be made, there is difference between the implementation of both these check hence this was required.

Vyom-Yadav:
> My advice, is if something looks like it may be more on the critical side to keep, we make it known and agree to it with a comment to the pitest suppression saying its needed and why.

Those were no big optimizations, just minor ones, won't affect the performance in a real-world scenario.

> but do note that since this is all going under #6320 that means it will not be reported as a bug on the release notes. This issue is marked as "misc" and no one really reads the "misc" notes since we are basically saying they aren't important.

It's fine, there are a lot of pending issues out there that can help me to get into the bugs section or the new section.

rnveach:
>  if you'd like, I can help in killing some of those mutations.

It is very much appreciated. Once this PR is merged we are planning to enable the pitest and move remaining items to suppression list to be fixed by contributors.

> that can help me to get into 

I was bringing it more up about reporting things to users. Bugs should be reported more clearly so users know there was an issue and can take appropriate action to see if they are affected and helps urge them more to upgrade.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
rnveach: Why can't we create a case to kill it without resulting to this which CI is set to avoid doing? We can't create a special package or such possibly matching `null.SomeClass`?

Vyom-Yadav:
> I was bringing it more up about reporting things to users. Bugs should be reported more clearly so users know there was an issue and can take appropriate action to see if they are affected and helps urge them more to upgrade. 

Yes we can increase the depth of release notes by using labels like `false-positive`, `false-negative`, `removed-exception` , `increased-coverage`, `notes` and many more, many organization already use these type of labels https://github.com/pmd/pmd/labels

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
Vyom-Yadav: There were quite a few mutations killed by using `null`, I had seen cases which required using null instead of `""`, but I will look into it again and try to do it with `""`, will post the results here.

rnveach: I am not against removing `null` so if that makes other mutations easier then ok. Why can't we use `null` someway to kill it for `packageName == null`. 
The result of the if/else is a big difference between `SomeClass` and `null.SomeClass`.

rnveach:
> Yes we can increase the depth of release notes 

Again, more/different labels wouldn't help in this case since again we are assigning what is almost a bug fix to a miscellaneous issue. I am talking more about splitting PR/commits then keeping everything in 1 commit.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
Vyom-Yadav: Done

Vyom-Yadav:
> I am talking more about splitting PR/commits then keeping everything in 1 commit.

Many of the changes fixed mutations and fixed bugs, both the things were done simultaneously so we can call this PR a bug fix if we want.

rnveach:
PRs cannot be given a special label when it is tied to an issue. It is the issue that takes precedence and everything tied to the issue is tied to the same label. It doesn't make sense that a PR would have different label than the issue since the issue is entailing the entire fix. We are not going to remove the connection to the issue in this PR.

rnveach (approved):

romani:
@Vyom-Yadav, please generate diff report by GitHub action

Vyom-Yadav:
Github generate report

Vyom-Yadav:
Github, generate report

romani (approved):
ok to merge.

romani:
problems is reported to https://github.com/jqno/equalsverifier/commit/8d9703e732cf278c0b7a5067dbe9938228191793#r65114830

PR #11234, opened 2022-01-25
doc: mention that excludes can be class names

tniessen:
I found it confusing that the `excludes` property only mentioned packages, yet when I specified a package, starred static member imports within that package were not excluded. This change clarifies that the `excludes` property uses packages for starred class imports and class names for starred static member imports.

https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html is a class
https://checkstyle.org/config_imports.html#AvoidStarImport we already have example where `exclude` uses not only package (2nd example).

---

Edit: the requested CLI demonstration of the existing behavior that is merely being documented in this PR:

```
tniessen@ubuntuserver:~/dev/checkstyle$ javac Test.java 
tniessen@ubuntuserver:~/dev/checkstyle$ cat Test.java 
import static java.lang.Math.*; // okay, class exclude applies
import static java.lang.System.*; // violation, package excludes do not apply
import java.io.*; // okay, package exclude applies
tniessen@ubuntuserver:~/dev/checkstyle$ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <module name="TreeWalker">
    <module name="AvoidStarImport">
      <property name="excludes" value="java.io,java.lang.Math"/>
    </module>
  </module>
</module>
tniessen@ubuntuserver:~/dev/checkstyle$ RUN_LOCALE="-Duser.language=en -Duser.country=US"
tniessen@ubuntuserver:~/dev/checkstyle$ java $RUN_LOCALE -jar checkstyle-9.2.1-all.jar -c config.xml Test.java 
Starting audit...
[ERROR] /home/tniessen/dev/checkstyle/Test.java:2:31: Using the '.*' form of import should be avoided - java.lang.System.*. [AvoidStarImport]
Audit done.
```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/AvoidStarImportCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/AvoidStarImportCheck.java
@@ -44,7 +44,8 @@ import com.puppycrawl.tools.checkstyle.api.TokenTypes;
  * </p>
  * <ul>
  * <li>
- * Property {@code excludes} - Specify packages where star imports are allowed.
+ * Property {@code excludes} - Specify packages where starred class imports are
+ * allowed and classes where starred static member imports are allowed.
  * Type is {@code java.lang.String[]}.
  * Default value is {@code ""}.
  * </li>
@@ -152,7 +153,10 @@ public class AvoidStarImportCheck
     /** Suffix for the star import. */
     private static final String STAR_IMPORT_SUFFIX = ".*";
 
-    /** Specify packages where star imports are allowed. */
+    /**
+     * Specify packages where starred class imports are
+     * allowed and classes where starred static member imports are allowed.
+     */
     private final List<String> excludes = new ArrayList<>();
 
     /**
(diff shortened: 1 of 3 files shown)


romani:
Please use our CLI to prove that it works like you describe

tniessen:
Hi @romani, could you please clarify that? It's my first time contributing to checkstyle and I'm not quite sure what I'm supposed to do.

There is an example in the existing documentation:

https://github.com/checkstyle/checkstyle/blob/5a33f82ce7f0f1e605631f1640a76f09d6c0cf2a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/AvoidStarImportCheck.java#L79-L86

rnveach:
> use our CLI to prove that it works like you describe
> could you please clarify that

See https://checkstyle.org/report_issue.html#How_to_report_a_bug.3F especially the section that says "Example of report that we expect". Update your first post with the report and just let us know when it is done.

tniessen:
Thanks for the explanation :) Done @rnveach @romani.

romani:
Github, generate web site

tniessen:
> before we merge, do you think it is reasonable in your example where "allowStaticMemberImports" is default so "false". we do not put violation.

@romani, sorry about the late reply! I believe the current implementation makes sense. `import static java.lang.Math.*` does not produce a violation because of `<property name="excludes" value="java.io,java.lang.Math"/>`.

romani:
Yes, you are right, I am not sure why I was confused.
I created https://github.com/checkstyle/checkstyle/issues/11352 to avoid confusion in future

romani (approved):
Ok to merge

rnveach (approved):

tniessen:
Thank you for reviewing and merging!

Issue #10924, opened 2021-11-04
All checks that use `AbstractCheck#getLine()` should check code points instead of characters for spacing


nmancus1:
Affected checks:
https://checkstyle.sourceforge.io/config_misc.html#CommentsIndentation
https://checkstyle.sourceforge.io/config_blocks.html#RightCurly
https://checkstyle.sourceforge.io/config_blocks.html#LeftCurly
https://checkstyle.sourceforge.io/config_whitespace.html#GenericWhitespace
https://checkstyle.sourceforge.io/config_whitespace.html#NoWhitespaceBefore
https://checkstyle.sourceforge.io/config_whitespace.html#SingleSpaceSeparator
https://checkstyle.sourceforge.io/config_whitespace.html#NoWhitespaceAfter

```bash

➜  checks git:(issue-10920) grep -R 'getLine(' | grep -v "indentCheck\|\.getLine(\|int"
indentation/CommentsIndentationCheck.java:        final String targetSourceLine = getLine(singleLineComment.getLineNo() - 1);
indentation/CommentsIndentationCheck.java:        final String commentLine = getLine(blockComment.getLineNo() - 1);
blocks/RightCurlyCheck.java:        else if (shouldBeAloneOnLine(option, details, getLine(details.rcurly.getLineNo() - 1))) {
blocks/LeftCurlyCheck.java:        final String braceLine = getLine(brace.getLineNo() - 1);
whitespace/WhitespaceAfterCheck.java:            final String line = getLine(targetAST.getLineNo() - 1);
whitespace/WhitespaceAfterCheck.java:            final String line = getLine(ast.getLineNo() - 1);
whitespace/GenericWhitespaceCheck.java:        final String line = getLine(ast.getLineNo() - 1);
whitespace/GenericWhitespaceCheck.java:        final String line = getLine(ast.getLineNo() - 1);
whitespace/NoWhitespaceBeforeCheck.java:        final String line = getLine(ast.getLineNo() - 1);
whitespace/WhitespaceAroundCheck.java:            final String line = getLine(ast.getLineNo() - 1);
whitespace/SingleSpaceSeparatorCheck.java:                            getLine(currentNode.getLineNo() - 1),
whitespace/NoWhitespaceAfterCheck.java:        final String line = getLine(astLineNo - 1);

```

[...]

PR #11240, opened 2022-01-26
Issue #10924: Update NoWhitespaceBeforeCheck 

MUzairS15:
Issue #10924: Updated NoWhitespaceBeforeCheck.

This check already uses code points, refactored to use methods in `CodePointUtil`.
Diff Regression config: https://gist.githubusercontent.com/MUzairS15/12920ff345227511c6d2b7321cb6e344/raw/8a547b379e405964c2fb4a1ec8e719c7a20d0366/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
@@ -23,6 +23,7 @@ import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.CodePointUtil;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 
 /**
@@ -209,19 +210,15 @@ public class NoWhitespaceBeforeCheck
 
     @Override
     public void visitToken(DetailAST ast) {
-        final String line = getLine(ast.getLineNo() - 1);
+        final int[] line = getLineCodePoints(ast.getLineNo() - 1);
         final int before = ast.getColumnNo() - 1;
-        final int[] codePoints = line.codePoints().toArray();
 
-        if ((before == -1 || CommonUtil.isCodePointWhitespace(codePoints, before))
+        if ((before == -1 || CommonUtil.isCodePointWhitespace(line, before))
                 && !isInEmptyForInitializerOrCondition(ast)) {
             boolean flag = !allowLineBreaks;
             // verify all characters before '.' are whitespace
-            for (int i = 0; i <= before - 1; i++) {
-                if (!CommonUtil.isCodePointWhitespace(codePoints, i)) {
-                    flag = true;
-                    break;
-                }
+            if (before > -1 && !CodePointUtil.hasWhitespaceBefore(before, line)) {
+                flag = true;
             }
             if (flag) {
                 log(ast, MSG_KEY, ast.getText());
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
@@ -211,16 +211,16 @@ public class NoWhitespaceBeforeCheck
     @Override
     public void visitToken(DetailAST ast) {
         final int[] line = getLineCodePoints(ast.getLineNo() - 1);
-        final int before = ast.getColumnNo() - 1;
+        final int columnNoBeforeToken = ast.getColumnNo() - 1;
+        final boolean isFirstToken = columnNoBeforeToken == -1;
 
-        if ((before == -1 || CommonUtil.isCodePointWhitespace(line, before))
+        if ((isFirstToken || CommonUtil.isCodePointWhitespace(line, columnNoBeforeToken))
                 && !isInEmptyForInitializerOrCondition(ast)) {
-            boolean flag = !allowLineBreaks;
-            // verify all characters before '.' are whitespace
-            if (before > -1 && !CodePointUtil.hasWhitespaceBefore(before, line)) {
-                flag = true;
-            }
-            if (flag) {
+            final boolean isViolation = !allowLineBreaks
+                    || !isFirstToken
+                    && !CodePointUtil.hasWhitespaceBefore(columnNoBeforeToken, line);
+
+            if (isViolation) {
                 log(ast, MSG_KEY, ast.getText());
             }
         }


MUzairS15:
Github, generate report

MUzairS15:
Pls review

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
```diff
@@ -209,19 +210,15 @@
...
+        if ((before == -1 || CommonUtil.isCodePointWhitespace(line, before))
                 && !isInEmptyForInitializerOrCondition(ast)) {
             boolean flag = !allowLineBreaks;
             // verify all characters before '.' are whitespace
-            for (int i = 0; i <= before - 1; i++) {
-                if (!CommonUtil.isCodePointWhitespace(codePoints, i)) {
-                    flag = true;
-                    break;
-                }
+            if (before > -1 && !CodePointUtil.hasWhitespaceBefore(before, line)) {
```
nick-mancuso, line 220: Please separate update to this check into two commits; for the first one, do only update to use codepoint utils and `getLineCodePoints` method. For the second one, please refactor logic in this block to just have one descriptively named boolean variable comprised of all conditions that we place on `flag` variable now (`!allowLineBreaks && before > -1 && !CodePointUtil.hasWhitespaceBefore(before, line)`). Do you think something like `hasNoWhitespaceBefore` is a good name?

nick-mancuso (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
nick-mancuso: After this update, please regenerate report.

MUzairS15:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
```diff
@@ -215,12 +215,11 @@ public void visitToken(DetailAST ast) {
 
         if ((before == -1 || CommonUtil.isCodePointWhitespace(line, before))
                 && !isInEmptyForInitializerOrCondition(ast)) {
-            boolean flag = !allowLineBreaks;
+            final boolean hasNoWhitespaceBefore = !allowLineBreaks
+                || before > -1
+                && !CodePointUtil.hasWhitespaceBefore(before, line);
             // verify all characters before '.' are whitespace
```
nick-mancuso, line 221: I think we can remove comment now, our intention is very clear.

nick-mancuso (changes requested):
Minor:

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
nick-mancuso: Resolved.

nick-mancuso:
`pitest-coding` failure is unrelated:
`+RequireThisCheck.java.html:<td class='covered'><pre><span  class='survived'>            if (returnedVariable) {</span></pre></td></tr>`

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
MUzairS15: done

nick-mancuso: Resolved.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
nick-mancuso: I see now that we made a mistake in naming logic, this is not what this condition tells us. Let's rename to `isNotFirstTokenOnLine`

MUzairS15: done

MUzairS15: done

MUzairS15: Screenshot 2022-03-15 at 9 48 40 PM

nick-mancuso: I suspected this may be a problem, let's just add this condition into `if` statement guarding log method then.

MUzairS15: need to combine all `if` condition

MUzairS15: done

MUzairS15: @nick-mancuso ping

nick-mancuso: > need to combine all if condition as PMD was failing

I think resulting condition is hard to understand now, it looks like we should go back to variable approach, if nothing else we can name it `isViolation`.

nick-mancuso: > need to combine all if condition as PMD was failing

I think resulting condition is hard to understand now, it looks like we should go back to variable approach, if nothing else we can name it `isViolation`.

MUzairS15: done

nick-mancuso (approved):

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
nick-mancuso: Resolved.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
```diff
@@ -209,21 +210,16 @@
...
-            // verify all characters before '.' are whitespace
-            for (int i = 0; i <= before - 1; i++) {
-                if (!CommonUtil.isCodePointWhitespace(codePoints, i)) {
-                    flag = true;
-                    break;
-                }
-            }
-            if (flag) {
+            final boolean isViolation = !allowLineBreaks
+                    || before > -1
```
strkkk, line 219: ` before > -1` 3 lines above it is `before == -1`
I do not see assignments to this variable, so it can be boolean variable. As a side note - may be refactor code to check it only once

Also, please rename `before` to something more meaningful.

strkkk (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCheck.java
MUzairS15: Refactoring code will involve more conditions to be checked in if block, resulting in hard to understand condition.

MUzairS15: renamed `before` to `columnNoBeforeToken`

pbludov (approved):

romani (approved):

PR #11242, opened 2022-01-27
Issue #10924: Update WhitespaceAroundCheck

MUzairS15:
Issue #10924: `WhitespaceAroundCheck` already uses `codepoint`. Updated this check to use `CodePointUtil` methods.

Diff Regression config: https://gist.githubusercontent.com/MUzairS15/d4bef73243734249b18a2f5fa01ef190/raw/ecbfc3ab7e81187983b6f8ae78547254421f9e1d/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundCheck.java
@@ -687,21 +687,17 @@ public class WhitespaceAroundCheck extends AbstractCheck {
     public void visitToken(DetailAST ast) {
         final int currentType = ast.getType();
         if (!isNotRelevantSituation(ast, currentType)) {
-            final String line = getLine(ast.getLineNo() - 1);
+            final int[] line = getLineCodePoints(ast.getLineNo() - 1);
             final int before = ast.getColumnNo() - 1;
             final int after = ast.getColumnNo() + ast.getText().length();
-            final int[] codePoints = line.codePoints().toArray();
 
-            if (before >= 0) {
-                final char prevChar = Character.toChars(codePoints[before])[0];
-                if (shouldCheckSeparationFromPreviousToken(ast)
-                        && !Character.isWhitespace(prevChar)) {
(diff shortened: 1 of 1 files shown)


MUzairS15:
Check already uses code point refactored to use Utility methods

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundCheck.java
```diff
@@ -687,23 +687,19 @@ public void setAllowEmptyCatches(boolean allow) {
...
+                log(ast, MSG_WS_NOT_PRECEDED, ast.getText());
             }
 
-            if (after < codePoints.length) {
-                final char nextChar = Character.toChars(codePoints[after])[0];
+            if (after < line.length) {
+                final char nextChar = Character.toChars(line[after])[0];
                 if (shouldCheckSeparationFromNextToken(ast, nextChar)
-                        && !Character.isWhitespace(nextChar)) {
+                        && !CommonUtil.isCodePointWhitespace(line, after)) {
```
nick-mancuso, line 702: Why change this method call? We have `nextChar` already, why not use it here, too?

nick-mancuso (changes requested):
Please generate regression report, we must do this if we make changes to check.

MUzairS15:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundCheck.java
MUzairS15: missed

MUzairS15: done

nick-mancuso: Resolved.

nick-mancuso (approved):

nick-mancuso:
Bit of refactoring here made sense, report is clear, I do not see a need to separate commits in this case.

romani (approved):

PR #11257, opened 2022-01-30
infra: jdk17 for equalsverifier

pbludov:
Attempt to solve https://github.com/jqno/equalsverifier/issues/586


romani:
Variable name should be fixed in that repo

PR #11262, opened 2022-01-31
Issue #10924: Update FallThroughCheck to use code points

MUzairS15:
Issue #10924: Updated FallThroughCheck to use code points

Diff Regression config: https://gist.githubusercontent.com/MUzairS15/cb3ffe158a0eb42df8ab37b07e8f6401/raw/1418b03a040cb0727788758b26f4fbcfbb67d862/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
@@ -26,7 +26,7 @@ import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
-import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
+import com.puppycrawl.tools.checkstyle.utils.CodePointUtil;
 
 /**
  * <p>
@@ -462,12 +462,6 @@ public class FallThroughCheck extends AbstractCheck {
     private boolean hasFallThroughComment(DetailAST currentCase, DetailAST nextCase) {
         boolean allThroughComment = false;
         final int endLineNo = nextCase.getLineNo();
-        final int endColNo = nextCase.getColumnNo();
-
-        // Remember: The lines number returned from the AST is 1-based, but
-        // the lines number in this array are 0-based. So you will often
-        // see a "lineNo-1" etc.
-        final String[] lines = getLines();
 
         // Handle:
         //    case 1:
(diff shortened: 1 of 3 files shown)


MUzairS15:
```
src % cat config.xml Test.java
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name = "Checker">

    <module name="TreeWalker">
        <module name="FallThrough"/>
    </module>
</module>

public class Test {
  void fooFallThru() {
    int i = 0;
    switch ("") {

       case "failing": String s = "s🥳d🥳s";
       /*🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳*/ /* fall through*/   case "test":
       case "passing": i = 2;
       /* adsdasdsdadsdasdasdadsadasdasdad */ /* fall through */ default:
    }
  }
}
src % java -jar checkstyle-10.0-all.jar -c config.xml Test.java
Starting audit...
[WARN] /Users/shabana/Documents/OpenSource/CheckStyleTest/src/src/Test.java:7:58: Fall through from previous branch of the switch statement. [FallThrough]
Audit done.

**Comment by MUzairS15** [comment]
Github, generate report

**Comment by nick-mancuso** [comment]
@MUzairS15 please remove usage of `getLines`; we can restrict usage of actual `String` outside of `matchesComment` method.

**Comment by MUzairS15** [comment]
Example updated pls check

> please remove usage of getLines;

done pls check

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java [hunk 1]
```diff
@@ -510,14 +505,18 @@ private boolean hasFallThroughComment(DetailAST currentCase, DetailAST nextCase)
...
      */
     // suppress deprecation until https://github.com/checkstyle/checkstyle/issues/11166
     @SuppressWarnings("deprecation")
-    private boolean matchesComment(Pattern pattern, String line, int lineNo) {
-        final Matcher matcher = pattern.matcher(line);
+    private boolean matchesComment(Pattern pattern, int lineNo) {
+        // Remember: The lines number returned from the AST is 1-based, but
+        // the lines number in this array are 0-based. So you will often
+        // see a "lineNo-1" etc.
+        final String[] lines = getLines();
```
nick-mancuso, line 517: As silly as this sounds, it might be best to pass code points to this method, and convert to a string to avoid using `getLines`.

If nothing else, we could use `getLine`, no need to get all lines in file.

@strkkk what do you think?

nick-mancuso (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
strkkk: Yes, that sounds reasonable. It is better to avoid unnecessary conversions

MUzairS15: done

nick-mancuso: Resolved.

nick-mancuso (approved):

romani (approved):

Issue #9146, opened 2021-01-02
upgrade minimal required jdk version to 11


romani:
World is moving to jdk11 as minimal requirement.

Eclipse requires jdk11 - https://wiki.eclipse.org/Eclipse/Installation

Sonar requires jdk11 - https://docs.sonarqube.org/latest/requirements/requirements/ (they warning us that our sevntu project is still running under jdk8 and will stop to work very soon)

IDEA requires jdk11 - https://www.jetbrains.com/help/idea/installation-guide.html
> You do not need to install Java to run IntelliJ IDEA, because JetBrains Runtime is bundled with the IDE (based on JRE 11). However, to develop Java applications, a standalone JDK is required.  

we discussed it at https://github.com/checkstyle/checkstyle/issues/7281 and waited for Eclipse to move to jdk11

romani:
execution on old jdk requires us to use more old OS instances in CI that cause different problems.
we need to add bunch of CI jobs to recheck build process on other JDKs.

I think it is good time for us to migrate to jdk11 as minimal required.
ALL maintainers should agree on this before approving this issue:
@rnveach , @pbludov , @strkkk , @nmancus1 , @esilkensen , @timurt , please vote

romani:
email from Sonar, subject `[ACTION REQUIRED] Deprecation of Java <11, upgrade before February 1st!`:
```
Hello,

The version of Java installed in your scanner environment must be upgraded to at least Java 11! 
We are ending support for older Java versions soon. You will still be able to use SonarCloud
 for your Java 8 projects, this change only affects the execution of our scanner, 
not the code your project is written in.

Here is how the deprecation will happen:

    Between the 11th and the 15th of January, you will experience a brownout where only the first
 scan of every project with an older version of Java will fail.
    After the 1st of February 2021, all analysis with Java <11 in your scanner environment 
will fail unless you upgrade to a supported version.

Please upgrade these projects now:

    sevntu-checks (checkstyle)

```

nmancus1:
I agree.  I can foresee issues with libraries we use dropping support for Java 8, too. A good article with lots of linked articles (at the bottom of the page): https://nipafx.dev/java-11-migration-guide/

Just a few thoughts:
 - We should create an issue move non-compilable test inputs to compilable
 - Would using the `module` system offer us any advantages?
 - Would we create a separate development branch?
 - Does this bump Checkstyle version number?
 
[...]

PR #11273, opened 2022-02-02
minor: fix mistype in ImportControlCheckTest

pbludov:
Discovered while working in #9146 

`message.startsWith(message)` is always true. It's strange that no tool reports this.


rnveach (approved):
Original code was like this too before the truth transformation. It is strange nothing found we weren't testing something like pitest.

PR #11280, opened 2022-02-04
minor: Set default locale for unit test execution to "en"

hurt:
There are a few unit tests that rely on the default locale language being set to english because they make assertions about localized messages. On systems where the default locale language is not "en", these tests fail.

This change sets the default language for test execution to "en" as surefire argument.

Other than that, I couldn't help but fixing some assertions in `DefaultLoggerTest` to improve the failure messages.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -197,7 +197,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <surefire.options />
+    <surefire.options>-Duser.language=en</surefire.options>
     <projectVersion>${project.version}</projectVersion>
     <antlr4.version>4.9.3</antlr4.version>
     <maven.site.plugin.version>3.11.0</maven.site.plugin.version>


romani:
Please share more details on on your local environment.
We have multiple complaints on this problem but nobody can clearly invest problem.

We have bunch of CI tests that runs in specific locale, see azure execution like test-fr . they are passing.

Tests should run in any locale , enforcement of en locale is not desirable.

romani:
All changes that are not in pom.xml can be sent in separate PR to merged quicker. Thanks a lot to spot them.

rnveach:
@hurt Are you able to provide the requested information on why this PR is needed?

strkkk:
> why this PR is needed?

@rnveach @romani I had issues as well. My locale is russian, and some tests are failing (1 or 2) because of it.

romani:
Please share more details on failure, can we reproduce it CI ?

strkkk:
@romani UT fails because default locale is not supported. It can be reproduced in ci by setting user locale to some other value, e.g. ru_RU

On my laptop it fails with
```
Unsupported language: ru_RU
expected not to be: Empty statement.

	at com.puppycrawl.tools.checkstyle.api.ViolationTest.testLocaleIsSupported
```
and it is annoying.

rnveach:
> setting user locale to some other value,

Is this not what we are doing already?
https://github.com/checkstyle/checkstyle/blob/master/azure-pipelines.yml#L41
https://github.com/checkstyle/checkstyle/blob/master/.ci/validation.sh#L130-L133

rnveach:
Did a rebase to see what CI says and what is new in this PR.

romani:
CI update for RU - https://github.com/checkstyle/checkstyle/pull/11529

rnveach:
> Unsupported language: ru_RU
expected not to be: Empty statement.

I changed the code to the below and the tests passed for me in Eclipse in this class.

````
diff --git a/src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java b/src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java
index efccf13..4a5b233 100644
--- a/src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java
@@ -50,7 +50,7 @@
  */
 public class ViolationTest {
 
-    private static final Locale DEFAULT_LOCALE = Locale.getDefault();
+    private static final Locale DEFAULT_LOCALE = Locale.forLanguageTag("ru_RU");
 
     /**
      * Verifies that the language specified with the system property {@code user.language} exists.
````

**Comment by rnveach** [comment]
@romani @strkkk Could this be because `ViolationTest` uses `Locale.getDefault()` but `Violation` uses `Locale.ROOT`? What is the difference between the 2?

In my JVM, this is the code for `Locale.ROOT`:
> `public static final Locale ROOT = createConstant("", "");`

**Comment by pbludov** [comment]
The origin of this issue is #7129

This test was added to ensure that there is not mistypes like `jp_JP`.
But this test fails if any language not supported by Checkstyle is set as the system language.

The assertion
```java
        final Violation violation = createSampleViolation();
        assertWithMessage("Unsupported language: %s", DEFAULT_LOCALE)
                .that(violation.getViolation())
                .isNotEqualTo("Empty statement.");
```
fails because the default language (en) was used instead. 

To fix this issue, we need to be able to distinguish between the case when the language is specified via the system properties and the case when the system language is used.

This test should not fail if no language was explicitly specified on the command line.

rnveach:
@pbludov Does anything you written conflict with https://github.com/checkstyle/checkstyle/pull/11529#discussion_r846662407 . It seems locale is not being passed to test to work properly.

pbludov:
> @pbludov Does anything you written conflict with [#11529 (comment)](https://github.com/checkstyle/checkstyle/pull/11529#discussion_r846662407) . It seems locale is not being passed to test to work properly.

It is a separate issue. The code
```java
        final String language = DEFAULT_LOCALE.getLanguage();
        assumeFalse(language.isEmpty(), "Locale not set");
```
is wrong, since `language.isEmpty()` is always `false`. Default locale can't be empty. If no `user.language` was specified by the user, the language of the OS will be used.

The issue of not working `argLine` should be fixed independently.

romani:
closed in favor of https://github.com/checkstyle/checkstyle/pull/12053

Issue #11289, opened 2022-02-06
Mark deprecated methods for removal


pbludov:
We are going to make a major version increase, so it is time to drop some deprecated methods.

- [x] AbstractCheck#getFileContents #11166
- [x] Configuration#getAttribute #10164
- [x] Configuration#getAttributeNames #10164
- [x] DefaultConfiguration#addAttribute #10164
- [x] DetailAST#branchContains #5234
- [x] DetailAST#getNumberOfChildren #7512

https://sonarcloud.io/project/issues?id=org.checkstyle%3Acheckstyle&resolved=false&severities=MAJOR 
https://www.baeldung.com/java-deprecated#optional-attributes

/DefaultConfiguration.java
https://github.com/checkstyle/checkstyle/blame/master/src/main/java/com/puppycrawl/tools/checkstyle/DefaultConfiguration.java#L143
since 8.45, forRemoval=true

https://github.com/checkstyle/checkstyle/blame/master/src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractCheck.java#L195
since 9.3

https://github.com/checkstyle/checkstyle/blame/master/src/main/java/com/puppycrawl/tools/checkstyle/api/Configuration.java#L39
since 8.45, forRemoval=true

https://github.com/checkstyle/checkstyle/blame/master/src/main/java/com/puppycrawl/tools/checkstyle/api/DetailAST.java#L101
since 8.43 forRemoval=true

https://github.com/checkstyle/checkstyle/blame/master/src/main/java/com/puppycrawl/tools/checkstyle/api/DetailAST.java#L141
since 8.30, forRemoval=true

it might be good time to remove all that deprecated methods (except for getFileContent as we still use it #10976 
so fix could be simply removal of such methods :) But better to make special issue for this to list it in release notes.
update annotation better to be done sooner .... as it is quick fix and merge, removal of methods might be not simple and affect other projects and execution on them.

_Originally posted by @romani in https://github.com/checkstyle/checkstyle/issues/9146#issuecomment-1030709274_

pbludov:
I'm not sure be can drop these methods now (except for method `NumberOfChildren` which wasn't really used before deprecation).
I changed the description to "Mark for removal"

PR #11291, opened 2022-02-06
Issue #11289: add since and forRemoval properties to Deprecated annotation

pbludov:
Issue #11289

Java 9 introduces two attributes for `@Deprecated` annotation: `since` and `forRemoval`.
The first one should be mandatory. New XPath check added to validate this.


romani (approved):

romani:
TC is not happy https://teamcity.jetbrains.com/viewLog.html?buildId=3782629&tab=Inspection&buildTypeId=Checkstyle_IdeaInspectionsPullRequest

Issue #11292, opened 2022-02-06
Add icon to the GUI window


pbludov:
Currently, the default Java application icon is used in the GUI window. This makes it difficult to use.
We have an icon image for the site `favicon`. It is possible to reuse this icon for GUI.

romani:
share image how it is without icon

pbludov:
The Java mascot:

![image](https://user-images.githubusercontent.com/900805/152688124-892c38cc-12e0-4ee2-9528-18f6f4aec854.png)

pbludov:
Same for Alt+Tab:
![image](https://user-images.githubusercontent.com/900805/152688292-137f41fc-cc0a-4568-a224-449e54796957.png)

pbludov:
Now:
![image](https://user-images.githubusercontent.com/900805/152689102-a01f5085-5040-44af-80e8-3eb77b8c890f.png)

rnveach:
@romani @pbludov Is this issue fixed or is there still more to do?

PR #11293, opened 2022-02-06
Issue #11292: icon for GUI

pbludov:
Issue #11292 

The icon image is same as `checkstyle_logo_small_64.png` from the site source.


src/main/java/com/puppycrawl/tools/checkstyle/gui/Main.java
```diff
@@ -47,6 +49,8 @@ public static void main(final String... args) {
                 final File sourceFile = new File(args[0]);
                 mainFrame.openFile(sourceFile);
             }
+            final URL iconUrl = Main.class.getResource("icon.png");
```
romani, line 52: all images are at https://github.com/checkstyle/resources/tree/master/img/cs-logos-twitter-gplus-backgrounds and in other folders of this repo.
https://github.com/checkstyle/resources/blob/master/img/cs-logos-twitter-gplus-backgrounds/cs-120x120.png is better as CS and pancil are in same size.


and we might figureout from where you got this page and fix it also

romani (changes requested):
item:

src/main/java/com/puppycrawl/tools/checkstyle/gui/Main.java
pbludov: done

romani (approved):

pbludov:
CI is failing. This is a false positive. Spell checker should not process PNG files.

PR #11299, opened 2022-02-07
Issue #11298: Added switch statements and lambda support to WhitespaceAfter

### LINKED ISSUE #11298: WhitespaceAfter does not work on 'switch' and lambda statements
I want to be able to configure Checkstyle to report a violation on missing space after 'switch', which is currently not possible.
WhitespaceAfter detects a missing space after 'for', 'if' etc. but _not_ after 'switch'.
It would be nice if there would be a LITERAL_SWITCH similar to LITERAL_FOR etc.

https://checkstyle.org/config_whitespace.html#WhitespaceAfter
I have downloaded the latest cli from: https://checkstyle.org/cmdline.html#Download_and_Run

**How it works Now:**

```bash
peter@pangolin:~/Downloads/checkstyle$ javac Test.java 

peter@pangolin:~/Downloads/checkstyle$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
	"http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
<module name="Checker">
	<module name="TreeWalker">
		<module name="WhitespaceAfter"/>
	</module>
</module>

peter@pangolin:~/Downloads/checkstyle$ cat Test.java 
class Test {

	void test() {
		for(int i = 0; i < 10; i++) { // Checkstyle catches this, configurable with LITERAL_FOR.
		}
		switch(1) { // Checkstyle does NOT catch this, there is no LITERAL_SWITCH to configure.
			default:
				break;
		}
	}
}

peter@pangolin:~/Downloads/checkstyle$ RUN_LOCALE="-Duser.language=en -Duser.country=US"
peter@pangolin:~/Downloads/checkstyle$ java $RUN_LOCALE -jar checkstyle-9.3-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /home/peter/Downloads/checkstyle/Test.java:4:17: 'for' is not followed by whitespace. [WhitespaceAfter]
Audit done.
```

The problem is that WhitespaceAfter catches missing spaces after 'if', 'for' etc., but _not_ after 'switch'.
Similar to 'for', I suggest a configurable token LITERAL_SWITCH with the "token" property which is configured by default, and a similar output:

```bash
[ERROR] home/peter/Downloads/checkstyle/Test.java:6:21: 'switch' is not followed by whitespace. [WhitespaceAfter]
```

**Additional context**

The code style that I want to impose:

```java
for (int i = 0; i < 10; i++) {
}
switch (1) {
	default:
		break;
}
```

Vyom-Yadav:
Resolves #11298 
Diff Regression config: https://gist.githubusercontent.com/Vyom-Yadav/98dceb63a79f4833e85fff9b2e1464a6/raw/97410d450226ecfe8c45df4f2693cd61c25c3765/my_checks.xml

Diff Regression patch config: https://gist.githubusercontent.com/Vyom-Yadav/273327b9c6cd6963b7dade792bc2c9c2/raw/a2715eb5a51a57dec2e34472a9797abf85f267db/patch_config.xml

---
PMD PR - https://github.com/pmd/pmd/pull/3767
Orekit PR - https://github.com/CS-SI/Orekit/pull/29

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAfterCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAfterCheck.java
@@ -57,7 +57,11 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_DO">
  * DO_WHILE</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#ELLIPSIS">
- * ELLIPSIS</a>.
+ * ELLIPSIS</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_SWITCH">
+ * LITERAL_SWITCH</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LAMBDA">
+ * LAMBDA</a>.
  * </li>
  * </ul>
  * <p>
@@ -153,6 +157,8 @@ public class WhitespaceAfterCheck
             TokenTypes.LITERAL_FOR,
             TokenTypes.DO_WHILE,
             TokenTypes.ELLIPSIS,
+            TokenTypes.LITERAL_SWITCH,
+            TokenTypes.LAMBDA,
         };
     }
 
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -608,7 +608,7 @@ no-error-orekit)
   # no CI is enforced in project, so to make our build stable we should
   # checkout to latest release/development (annotated tag or hash) or sha that have fix we need
   # git checkout $(git describe --abbrev=0 --tags)
-  git checkout "76760bf""bf""b847e227490cd5d3662f""ca087f1a324"
+  git checkout "851de782c6""d16d""f725fa""bb4646ff0dd086723415"
   mvn -e --no-transfer-progress compile checkstyle:check \
     -Dorekit.checkstyle.version=${CS_POM_VERSION}
   cd ..
--- a/config/suppressions.xml
+++ b/config/suppressions.xml
@@ -6,7 +6,7 @@
 
 <suppressions>
   <!-- can't split long messages between lines -->
-  <suppress id="lineLengthXml" files="google_checks\.xml" lines="56,126"/>
+  <suppress id="lineLengthXml" files="google_checks\.xml" lines="56,127"/>
   <!-- don't validate generated files -->
   <suppress id="lineLengthXml" files="releasenotes_old_6\-0_7\-8\.xml"/>
   <suppress id="lineLengthXml" files="releasenotes_old_1\-0_5\-9\.xml"/>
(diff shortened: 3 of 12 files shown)


Vyom-Yadav:
I didn't create a new issue for lambda expressions as enhanced switch also uses the `LAMBDA` token.

Vyom-Yadav:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAfterCheck.java
```diff
@@ -153,6 +157,8 @@
             TokenTypes.LITERAL_FOR,
             TokenTypes.DO_WHILE,
             TokenTypes.ELLIPSIS,
+            TokenTypes.LITERAL_SWITCH,
```
strkkk, line 160: I can agree on lambda to be included in list of default tokens, but switch is not that obvious. I think we can exclude it from list of default tokens. Report shows that it is pretty common to do not put whitespace after switch.

@romani @rnveach @nmancus1 @pbludov what do you think?

Vyom-Yadav: WhitespaceAround also contains `LITERAL_SWITCH`, we can put it in acceptable tokens if not in default.
https://github.com/checkstyle/checkstyle/blob/4b568976ada9210363162e64e232a0c67edbdc74/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundCheck.java#L527

pbludov: > Report shows that it is pretty common to do not put whitespace after switch.

I don't think it's intentional. For example, HBase uses `switch (` too: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/916c493_2022100915/reports/diff/Hbase/xref/home/runner/work/checkstyle/checkstyle/contribution/checkstyle-tester/repositories/Hbase/hbase-examples/src/main/java/org/apache/hadoop/hbase/coprocessor/example/generated/BulkDeleteProtos.java.html#L98

The `switch` is the same statement as `for`, `while`, `catch`. I see no reason to make an exception for it.

peterdemaeyer (approved):
Looks good to me.
I like the fact that you also covered the lambda.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAfterCheck.java
romani: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace demands space in similar structures, so I think it is ok to make is as default.

@Vyom-Yadav , please update inputs for Google style to make it clear our config will start to enforce space after such tokens.

romani: @pbludov , @rnveach please share your position on default

Vyom-Yadav: > please update inputs for Google style to make it clear our config will start to enforce space after such tokens.

Done

pbludov: > @pbludov , @rnveach please share your position on default

SWITCH should be included by default.

romani (approved):

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAfterCheck.java
strkkk: ok, thanks

strkkk (approved):

romani:
Wercker detected regression on PMD, we need to send PR to them.
```
[INFO] --- maven-checkstyle-plugin:3.1.2:check (checkstyle-check) @ pmd-core ---
[INFO] There is 1 error reported by Checkstyle 10.0-SNAPSHOT with /net/sourceforge/pmd/pmd-checkstyle-config.xml ruleset.
[ERROR] src/main/java/net/sourceforge/pmd/cpd/GUI.java:[173,21] (whitespace) WhitespaceAfter: 'switch' is not followed by whitespace.
```

pbludov (approved):

Vyom-Yadav:
@romani All related PR's have been merged

PR #11310, opened 2022-02-12
Issue #11309: Replace usages of deprecated method AccessibleObject.is…

### LINKED ISSUE #11309: Replace usages of deprecated method AccessibleObject.isAccessible
Noticed in https://github.com/checkstyle/checkstyle/pull/10904/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L1521

Now that we have moved to Java 11, we should be able to replace usages of `AccessibleObject#isAccessible` and remove supressions at:

https://github.com/checkstyle/checkstyle/blob/d60a1b6638aa59e6cfb247184ecf7191b0743791/pom.xml#L1520-L1523

nmancus1:
Closes #11309.

From https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/reflect/AccessibleObject.html#trySetAccessible():

> Set the accessible flag for this reflected object to true if possible. This method sets the accessible flag, as if by invoking [setAccessible(true)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/reflect/AccessibleObject.html#setAccessible(boolean)), and returns the possibly-updated value for the accessible flag. If access cannot be enabled, i.e. the checks or Java language access control cannot be suppressed, this method returns false (as opposed to setAccessible(true) throwing InaccessibleObjectException when it fails).

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
@@ -552,9 +552,7 @@ public class AllChecksTest extends AbstractModuleTestSupport {
                     .isEqualTo(Modifier.PUBLIC | Modifier.STATIC | Modifier.FINAL);
 
                 // below is required for package/private classes
-                if (!message.isAccessible()) {
-                    message.setAccessible(true);
-                }
+                message.trySetAccessible();
 
                 if (!INTERNAL_MODULES.contains(module.getSimpleName())) {
                     verifyCheckstyleMessage(usedMessages, module, message);
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
@@ -1283,7 +1283,7 @@ public class XdocsPagesTest {
         while (!Object.class.equals(currentClass)) {
             try {
                 result = currentClass.getDeclaredField(propertyName);
-                result.setAccessible(true);
+                result.trySetAccessible();
                 break;
             }
             catch (NoSuchFieldException ignored) {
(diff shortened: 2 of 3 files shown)


romani (approved):

src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
```diff
@@ -552,9 +552,7 @@ public void testAllCheckstyleMessages() throws Exception {
                     .isEqualTo(Modifier.PUBLIC | Modifier.STATIC | Modifier.FINAL);
 
                 // below is required for package/private classes
-                if (!message.isAccessible()) {
-                    message.setAccessible(true);
-                }
+                message.trySetAccessible();
```
rnveach, line 555: Ignoring inputs, does this only impact Fields or do we need to do methods and constructors?
https://github.com/checkstyle/checkstyle/search?q=setAccessible

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1283,7 +1283,7 @@ private static Field getField(Class<?> fieldClass, String propertyName) {
         while (!Object.class.equals(currentClass)) {
             try {
                 result = currentClass.getDeclaredField(propertyName);
-                result.setAccessible(true);
+                result.trySetAccessible();
```
nmancus1, line 1286: Done for consistency, not required.

src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
rnveach: My fault, resolved.

rnveach (approved):

src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
nmancus1: We really only need to replace usages where we checked if something was accessible first to avoid exception.

Issue #11166, opened 2022-01-08
Fix existing cases of `getFileContents()` usage


romani:
after https://github.com/checkstyle/checkstyle/issues/9889

We need to fix all existing cases of `getFileContents()` usage.

cases:
- [ ] MissingOverrideCheck.java
- [ ] PackageAnnotationCheck.java
- [ ] AvoidEscapedUnicodeCharactersCheck.java
- [ ] FallThroughCheck.java
- [ ] PackageDeclarationCheck.java
- [ ] ImportControlCheck.java
- [ ] UnusedImportsCheck.java
- [ ] JavadocMethodCheck.java (1)
- [ ] JavadocStyleCheck.java
- [ ] JavadocTypeCheck.java
- [ ] JavadocVariableCheck.java
- [ ] MissingJavadocMethodCheck.java
- [ ] MissingJavadocPackageCheck.java
- [ ] MissingJavadocTypeCheck.java
- [ ] WriteTagCheck.java
- [ ] OuterTypeFilenameCheck.java
- [ ] RegexpCheck.java
- [ ] RegexpSinglelineJavaCheck.java
- [ ] MethodLengthCheck.java
- [ ] EmptyLineSeparatorCheck.java
- [ ] JavadocMetadataScraper.java

it is better to resolve each case in separate PR. Each PR should remove the suppression to show the issue is fixed. https://github.com/haanhvu/checkstyle/commit/14391708cba2675bb2cd56135533bc82d6c120c2#diff-c1ffc75b7152435a4d5d40d7ee9b32b57402959b1ad471a41e753ea67554378b

remal:
@romani I have some custom checks for JavaDocs and, I'm afraid, it's not quite clear what exactly is supposed to be used instead of `getFileContents()`. Could you (or someone else) explain what is supposed to be used?

My goal is it check that JavaDoc is presented on every element that matches specific condition, has some tag (`@author`) and the tag's value is equal to some constant.

strkkk:
@remal instead of `getContents()` checks should use only AST and do not rely on text in file.
About the case you described - it seems to me that you dont need a custom check for it, MatchXPathCheck can provide same functionality.

romani:
@remal , please look at
https://github.com/checkstyle/checkstyle/blob/0b3a2606e535eb14d2d9dcd6c79d00db17c22f91/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AtclauseOrderCheck.java#L152

[...]

PR #11312, opened 2022-02-13
11166 remove get file contents method length

strkkk:
Issue #11166 
Remove usage of getFileContents for method length check.

Note - count is changed because now check ignores block comments as well.

Report - https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/80b7f57_2022070929/reports/diff/index.html

Diff Regression config: https://gist.githubusercontent.com/strkkk/31749e0743553a606f69fdfc6670d780/raw/4d8840dce7f80febf98199fab58575402d503fec/confg.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
@@ -19,10 +19,17 @@
 
 package com.puppycrawl.tools.checkstyle.checks.sizes;
 
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
-import com.puppycrawl.tools.checkstyle.api.FileContents;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 
@@ -42,8 +49,7 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
  * Default value is {@code 150}.
  * </li>
  * <li>
- * Property {@code countEmpty} - Control whether to count empty lines and single
- * line comments of the form {@code //}.
+ * Property {@code countEmpty} - Control whether to count empty lines and comments.
  * Type is {@code boolean}.
  * Default value is {@code true}.
  * </li>
@@ -77,7 +83,7 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
  * </pre>
  * <p>
  * To configure the check so that it accepts methods with at most 60 lines,
- * not counting empty lines and single line comments:
+ * not counting empty lines and comments:
  * </p>
  * <pre>
  * &lt;module name="MethodLength"&gt;
@@ -112,7 +118,7 @@ public class MethodLengthCheck extends AbstractCheck {
     /** Default maximum number of lines. */
     private static final int DEFAULT_MAX_LINES = 150;
 
-    /** Control whether to count empty lines and single line comments of the form {@code //}. */
+    /** Control whether to count empty lines and comments. */
     private boolean countEmpty = true;
 
     /** Specify the maximum number of lines allowed. */
@@ -141,9 +147,14 @@ public class MethodLengthCheck extends AbstractCheck {
     public void visitToken(DetailAST ast) {
         final DetailAST openingBrace = ast.findFirstToken(TokenTypes.SLIST);
         if (openingBrace != null) {
-            final DetailAST closingBrace =
-                openingBrace.findFirstToken(TokenTypes.RCURLY);
-            final int length = getLengthOfBlock(openingBrace, closingBrace);
+            final int length;
+            if (countEmpty) {
+                final DetailAST closingBrace = openingBrace.findFirstToken(TokenTypes.RCURLY);
+                length = getLengthOfBlock(openingBrace, closingBrace);
+            }
+            else {
+                length = countUsedLines(openingBrace);
+            }
             if (length > max) {
                 final String methodName = ast.findFirstToken(TokenTypes.IDENT).getText();
                 log(ast, MSG_KEY, length, max, methodName);
@@ -152,30 +163,45 @@ public class MethodLengthCheck extends AbstractCheck {
     }
 
     /**
-     * Returns length of code only without comments and blank lines.
+     * Returns length of code.
      *
      * @param openingBrace block opening brace
      * @param closingBrace block closing brace
      * @return number of lines with code for current block
      */
-    // suppress deprecation until https://github.com/checkstyle/checkstyle/issues/11166
-    @SuppressWarnings("deprecation")
-    private int getLengthOfBlock(DetailAST openingBrace, DetailAST closingBrace) {
-        int length = closingBrace.getLineNo() - openingBrace.getLineNo() + 1;
+    private static int getLengthOfBlock(DetailAST openingBrace, DetailAST closingBrace) {
+        final int startLineNo = openingBrace.getLineNo();
+        final int endLineNo = closingBrace.getLineNo();
+        return endLineNo - startLineNo + 1;
+    }
 
-        if (!countEmpty) {
-            final FileContents contents = getFileContents();
-            final int lastLine = closingBrace.getLineNo();
-            // lastLine - 1 is actual last line index. Last line is line with curly brace,
-            // which is always not empty. So, we make it lastLine - 2 to cover last line that
-            // actually may be empty.
-            for (int i = openingBrace.getLineNo() - 1; i <= lastLine - 2; i++) {
-                if (contents.lineIsBlank(i) || contents.lineIsComment(i)) {
-                    length--;
-                }
+    /**
+     * Count number of used code lines without comments.
+     *
+     * @param ast start ast
+     * @return number of used lines of code
+     */
+    private static int countUsedLines(DetailAST ast) {
+        final Deque<DetailAST> nodes = new ArrayDeque<>();
+        nodes.add(ast);
+        final Set<Integer> usedLines = new HashSet<>();
+        while (!nodes.isEmpty()) {
+            final DetailAST node = nodes.removeFirst();
+            final int lineNo = node.getLineNo();
+            // text block requires special treatment,
+            // since it is the only non-comment token that can span more than one line
+            if (node.getType() == TokenTypes.TEXT_BLOCK_LITERAL_BEGIN) {
+                IntStream.rangeClosed(lineNo, node.getLastChild().getLineNo())
+                    .forEach(usedLines::add);
+            }
+            else {
+                usedLines.add(lineNo);
+                Stream.iterate(
+                    node.getLastChild(), Objects::nonNull, DetailAST::getPreviousSibling
+                ).forEach(nodes::addFirst);
             }
         }
-        return length;
+        return usedLines.size();
     }
 
     /**
@@ -188,11 +214,9 @@ public class MethodLengthCheck extends AbstractCheck {
     }
 
     /**
-     * Setter to control whether to count empty lines and single line comments
-     * of the form {@code //}.
+     * Setter to control whether to count empty lines and comments.
      *
-     * @param countEmpty whether to count empty and single line comments
-     *     of the form //.
+     * @param countEmpty whether to count empty and comments.
      */
     public void setCountEmpty(boolean countEmpty) {
         this.countEmpty = countEmpty;
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheckTest.java
@@ -86,10 +86,40 @@ public class MethodLengthCheckTest extends AbstractModuleTestSupport {
     }
 
     @Test
+    public void testCountEmpty() throws Exception {
+        final int max = 2;
+        final String[] expected = {
+            "24:5: " + getCheckMessage(MSG_KEY, 3, max, "AA"),
+            "41:5: " + getCheckMessage(MSG_KEY, 3, max, "threeLines"),
+            "45:5: " + getCheckMessage(MSG_KEY, 3, max, "threeLinesAndComments"),
+            "53:5: " + getCheckMessage(MSG_KEY, 3, max, "threeLinesWrap"),
+            "63:5: " + getCheckMessage(MSG_KEY, 10, max, "m2"),
+        };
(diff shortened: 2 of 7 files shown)


strkkk:
Github, generate report

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -222,6 +224,21 @@ public void testIsCommentType() {
...
 
+    @ParameterizedTest
+    @ValueSource(ints = {
+        TokenTypes.SINGLE_LINE_COMMENT,
+        TokenTypes.BLOCK_COMMENT_BEGIN,
+        TokenTypes.BLOCK_COMMENT_END,
+        TokenTypes.COMMENT_CONTENT,
+    })
+    public void testIsCommentTypeAst(int type) {
+        final DetailAstImpl ast = new DetailAstImpl();
```
romani, line 235: Can we avoid pure Unit test ? Even it is until, it is better to cover it by full functional test

romani (changes requested):
Items:

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
strkkk: I think it is impossible for such util methods

strkkk:
Github, generate report

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
strkkk: nevermind, I excluded changes in TokenUtil from this PR, it is not needed

strkkk:
Github, generate report

strkkk:
Github, generate report

strkkk:
Github, generate report

.ci/pitest.sh
```diff
@@ -67,6 +67,14 @@ pitest-header)
   checkPitestReport "${ignoredItems[@]}"
   ;;
 
+pitest-sizes)
+  mvn --no-transfer-progress -e -P$1 clean test-compile org.pitest:pitest-maven:mutationCoverage;
+  declare -a ignoredItems=(
+  "MethodLengthCheck.java.html:<td class='covered'><pre><span  class='survived'>                &#38;&#38; isExceedingLimit(length)) {</span></pre></td></tr>"
```
romani, line 73: We do a lot to reduce list of such ignore items.
Please share challenges that force to do this.

We would rather has pure Junit to kill mutation to avoid ignored from pitest

romani (changes requested):
Item:

.ci/pitest.sh
strkkk: I added this check to make sure that we really need to iterate over whole AST to count exact number of lines.
For example, you have config like
```
MethodLengthCheck
countEmpty = false
max = 1000
```
and you have class with 100 methods, each of them is less than 1000 lines and have 200 elements in ast. Apparently, check will produce 0 violations for such class.
And now we can see difference that this "redundant" check gives us:
1. Without it, you will iterate over every node of ast of each method, plus create 100 new boolean arrays and 100 array queues. Basically, for such cases in will result in `O(n + m)` memory complexity and `O(n)` time complexity, where `n` is number of ast nodes in all methods and `m` is number of lines. Plus, all those arrays has to be collected by gc, which adds some penalty.

2. You just make a shortcut check for raw length, thus skip iteration. That will result in O(1) in both space and time complexity.

I have no idea how to explain all this to pitest. Of course, parsing takes a lot more time than inspection itself and show it by single input is impossible. For such cases, difference can be shown on big scale.

strkkk: ok, I removed it. I hope someday we will have tests that can prove not only logic, but also space/time advantages.

strkkk:
Github, generate report

romani:
This method had 180 lines, why we do not report on it any more https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/80b7f57_2022070929/reports/diff/apache-ant/index.html#A1

strkkk:
@romani because now it calculates line length correctly, skipping all comments (single and block)
Previous implementation counted block comments and ignored only single line.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
```diff
@@ -158,24 +162,55 @@ public void visitToken(DetailAST ast) {
...
-                if (contents.lineIsBlank(i) || contents.lineIsComment(i)) {
-                    length--;
-                }
+            rawLength = countUsedLines(openingBrace.getFirstChild(), startLineNo, rawLength);
+        }
+        return rawLength;
+    }
+
+    /**
+     * Collects numbers of lines where code present into given set.
```
romani, line 177: > into given set.

?? just remove this.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
```diff
@@ -158,24 +162,55 @@ public void visitToken(DetailAST ast) {
...
-            final FileContents contents = getFileContents();
-            final int lastLine = closingBrace.getLineNo();
-            // lastLine - 1 is actual last line index. Last line is line with curly brace,
-            // which is always not empty. So, we make it lastLine - 2 to cover last line that
-            // actually may be empty.
-            for (int i = openingBrace.getLineNo() - 1; i <= lastLine - 2; i++) {
-                if (contents.lineIsBlank(i) || contents.lineIsComment(i)) {
-                    length--;
-                }
+            rawLength = countUsedLines(openingBrace.getFirstChild(), startLineNo, rawLength);
```
romani, line 171: it is very weird set of parameters.

lets make code like:
```
if (countEmpty) {
  length = countUsedLines(openingBrace, closingBrace);
} esle {
   length = getLengthOfBlock(openingBrace, closingBrace);
}
```

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
strkkk: my bad, leftover from one of previous versions. Fixed.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
strkkk: This way wont work because of pitest.
To calculate length for boolean array, we need to calculate `endLine - startLine + 1`, but pitest will mutate this into `endLine + startLine + 1`, `endLine + startLine - 1` and both will survive because it allocates more space that we need. 
Therefore rawLength is passed as parameter, mutations here will be killed.

To be honest, in this place pitest is only making things uglier.

romani: try in this direction:
```
./.ci/pitest.sh pitest-sizes
...
>> Line Coverage: 280/280 (100%)
>> Generated 244 mutations Killed 244 (100%)
>> Mutations with no coverage 0. Test strength 100%
>> Ran 411 tests (1.68 tests per mutation)
Enhanced functionality available at https://www.arcmutate.com/
[INFO] BUILD SUCCESS
[INFO] Total time:  50.196 s
[INFO] Finished at: 2022-02-20T19:39:06-08:00
✔ ~/java/github/romani/checkstyle [strkkk/11166_remove_getFileContents_method_length L|✚ 1] 
19:39 $ git diff
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java b/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
index 32ccc87..ad9928e 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
@@ -20,8 +20,12 @@
 package com.puppycrawl.tools.checkstyle.checks.sizes;
 
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Deque;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 import java.util.stream.IntStream;
 
@@ -147,7 +151,14 @@ public class MethodLengthCheck extends AbstractCheck {
         if (openingBrace != null) {
             final DetailAST closingBrace =
                 openingBrace.findFirstToken(TokenTypes.RCURLY);
-            final int length = getLengthOfBlock(openingBrace, closingBrace);
+            final int length;
+            if (countEmpty) {
+                length = getLengthOfBlock(openingBrace, closingBrace);
+            } else {
+                final int startLineNo = openingBrace.getLineNo();
+                final int endLineNo = closingBrace.getLineNo();
+                length = countUsedLines(openingBrace.getFirstChild(), startLineNo);
+            }
             if (length > max) {
                 final String methodName = ast.findFirstToken(TokenTypes.IDENT).getText();
                 log(ast, MSG_KEY, length, max, methodName);
@@ -165,12 +176,7 @@ public class MethodLengthCheck extends AbstractCheck {
     private int getLengthOfBlock(DetailAST openingBrace, DetailAST closingBrace) {
         final int startLineNo = openingBrace.getLineNo();
         final int endLineNo = closingBrace.getLineNo();
-        int rawLength = endLineNo - startLineNo + 1;
-
-        if (!countEmpty) {
-            rawLength = countUsedLines(openingBrace.getFirstChild(), startLineNo, rawLength);
-        }
-        return rawLength;
+        return endLineNo - startLineNo + 1;
     }
 
     /**
@@ -181,9 +187,9 @@ public class MethodLengthCheck extends AbstractCheck {
      * @param length raw length of code block
      * @return number of used lines of code
      */
-    private static int countUsedLines(DetailAST ast, int startLine, int length) {
-        final boolean[] usedLines = new boolean[length];
-        usedLines[0] = true;
+    private static int countUsedLines(DetailAST ast, int startLine) {
+        final Set<Integer> usedLines = new HashSet<Integer>();
+        usedLines.add(0);
         final Deque<DetailAST> nodes = new ArrayDeque<>();
         Stream.iterate(ast, Objects::nonNull, DetailAST::getNextSibling).forEach(nodes::addLast);
         while (!nodes.isEmpty()) {
@@ -196,21 +202,15 @@ public class MethodLengthCheck extends AbstractCheck {
                 // count only lines inside, exclude first and last lines
                 IntStream.range(1, contentLinesCount)
                     .map(idx -> offset + idx)
-                    .forEach(idx -> usedLines[idx] = true);
+                    .forEach(usedLines::add);
             }
             else {
-                usedLines[offset] = true;
+                usedLines.add(offset);
             }
             Stream.iterate(node.getLastChild(), Objects::nonNull, DetailAST::getPreviousSibling)
                 .forEach(nodes::addFirst);
         }
-        int result = 0;
-        for (boolean usedLine : usedLines) {
-            if (usedLine) {
-                result++;
-            }
-        }
-        return result;
+        return usedLines.size();
     }
 
     /**
```

romani: > `if (countEmpty)`

should be visitToken method

src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheckTest.java
```diff
@@ -106,7 +136,24 @@ public void testRecordsAndCompactCtors() throws Exception {
         };
 
         verifyWithInlineConfigParser(
-                getNonCompilablePath("InputMethodLengthRecordsAndCompactCtors.java"),
+            getNonCompilablePath("InputMethodLengthRecordsAndCompactCtors.java"),
+            expected);
```
romani, line 140: please move this formatting out or revert it.

romani (changes requested):
items:

romani:
@strkkk , ping

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
strkkk: I know that it will work with set, because I used it on some of previous versions.

It is not best approach because we know size of this set in advance, so we can use array of flags and avoid useless overhead - boxing, set extension, creation of useless objects. 

In general, pitest makes implementation very inefficient.

src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheckTest.java
strkkk: done

strkkk:
Github, generate report

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/sizes/methodlength/InputMethodLengthCompactCtorsCountEmpty.java
```diff
@@ -0,0 +1,71 @@
...
+            // some comment
+            /*
+            block comment
+             */
+
+        }
+    }
+
+    class LocalRecordHelper {
+        Class<?> m(int x) { // violation
```
romani, line 55: Please use messages for all violations.

romani (changes requested):
Items

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/sizes/methodlength/InputMethodLengthCompactCtorsCountEmpty.java
strkkk: done

romani (approved):
Ok to merge

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
romani: Not always, but I agree.
It is also true to jacoco and all other static code analysis

romani:
GitHub, generate report

romani:
```
Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: unable to parse configuration stream - The markup in the document following the root element must be well-formed.:41:3
    at com.puppycrawl.tools.checkstyle.ConfigurationLoader.loadConfiguration (ConfigurationLoader.java:322)
    at com.puppycrawl.tools.checkstyle.ConfigurationLoader.loadConfiguration (ConfigurationLoader.java:267)
    at
```

pbludov (approved):

strkkk:
Github, generate report

nick-mancuso:
Github, generate site

nick-mancuso:
Link to check documentation: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/f873a9d_2022144552/config_sizes.html#MethodLength

It would be good to have examples with both kinds of comments to show users that  both single line and block comments do not affect check when `countEmpty` is false.

Issue created at https://github.com/checkstyle/checkstyle/issues/11474, we can update examples there.

nick-mancuso (approved):

PR #11325, opened 2022-02-19
 Issue #6320: add remove conditionals mutator for coding package 

rnveach:
Issue https://github.com/checkstyle/checkstyle/issues/6320

Added new mutator and starting list.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2233,8 +2233,7 @@
                 <mutator>INVERT_NEGS</mutator>
                 <mutator>MATH</mutator>
                 <mutator>NEGATE_CONDITIONALS</mutator>
-                <!-- result in 115 extra survived items, too much to keep in ignore list -->
-                <!-- <mutator>REMOVE_CONDITIONALS</mutator> -->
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -2249,7 +2248,7 @@
                 <param>*.Input*</param>
               </excludedTestClasses>
               <coverageThreshold>100</coverageThreshold>
-              <mutationThreshold>100</mutationThreshold>
+              <mutationThreshold>97</mutationThreshold>
```
romani, line 2251: If we reduce it, it might cause leak in other mutators

romani (changes requested):
Item:

pom.xml
rnveach: It is being reduced to 99%. List is being created of outstanding items.

romani (approved):

pom.xml
romani: while we have validation in shell script we are safe.

pbludov (approved):

PR #11133, opened 2021-12-30
Issue #11091: Fail test execution when input file hasn't specified violation messages

Vyom-Yadav:
Resolves #11091

PR #11333, opened 2022-02-22
Pull #11133: refactor check for override annotation - add util method

strkkk:
Same check is in multiple places, so extract common method

Diff Regression config: https://gist.githubusercontent.com/strkkk/26466254f89c32d8f1a6bd53b6f5b251/raw/54d6f5aa9f33b6f47a305772c469d872239049eb/config_single.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
@@ -809,7 +809,7 @@ public final class IllegalTypeCheck extends AbstractCheck {
         final String methodName =
             ast.findFirstToken(TokenTypes.IDENT).getText();
         return isVerifiable(ast) && !ignoredMethodNames.contains(methodName)
-                && !AnnotationUtil.containsAnnotation(ast, "Override");
+                && !AnnotationUtil.hasOverrideAnnotation(ast);
     }
 
     /**
--- a/pom.xml
+++ b/pom.xml
@@ -3292,6 +3292,9 @@
                   com.puppycrawl.tools.checkstyle.checks.coding.MultipleStringLiteralsCheckTest
                 </param>
                 <param>
+                  com.puppycrawl.tools.checkstyle.checks.coding.IllegalThrowsCheckTest
+                </param>
+                <param>
                   com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocMethodCheckTest
                 </param>
                 <param>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/MissingOverrideCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/MissingOverrideCheck.java
@@ -184,12 +184,6 @@ public final class MissingOverrideCheck extends AbstractCheck {
     public static final String MSG_KEY_ANNOTATION_MISSING_OVERRIDE =
         "annotation.missing.override";
 
-    /** {@link Override Override} annotation name. */
-    private static final String OVERRIDE = "Override";
-
-    /** Fully-qualified {@link Override Override} annotation name. */
-    private static final String FQ_OVERRIDE = "java.lang." + OVERRIDE;
-
     /** Compiled regexp to match Javadoc tags with no argument and {}. */
     private static final Pattern MATCH_INHERIT_DOC =
             CommonUtil.createPattern("\\{\\s*@(inheritDoc)\\s*\\}");
@@ -251,21 +245,10 @@ public final class MissingOverrideCheck extends AbstractCheck {
 
             if (check
                 && containsTag
-                && !hasOverrideAnnotation(ast)) {
+                && !AnnotationUtil.hasOverrideAnnotation(ast)) {
                 log(ast, MSG_KEY_ANNOTATION_MISSING_OVERRIDE);
             }
         }
-    }
-
-    /**
-     * Checks to see if the ast has Override annotation.
-     *
-     * @param ast AST node
-     * @return true if has Override annotation
-     */
-    private static boolean hasOverrideAnnotation(DetailAST ast) {
-        return AnnotationUtil.containsAnnotation(ast, OVERRIDE)
-                || AnnotationUtil.containsAnnotation(ast, FQ_OVERRIDE);
     }
 
     /**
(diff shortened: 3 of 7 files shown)


romani:
spelling is faling

strkkk:
> spelling is faling

fixed

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
```diff
@@ -809,7 +809,7 @@ private boolean isCheckedMethod(DetailAST ast) {
         final String methodName =
             ast.findFirstToken(TokenTypes.IDENT).getText();
         return isVerifiable(ast) && !ignoredMethodNames.contains(methodName)
-                && !AnnotationUtil.containsAnnotation(ast, "Override");
+                && !AnnotationUtil.hasOverrideAnnotation(ast);
```
pbludov, line 812: This may be a breaking change, as this check has not previously supported FQDN annotations.

pbludov (approved):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
rnveach: @strkkk Please provide regression.

strkkk:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
strkkk: Report is clean - https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/e081ae1_2022180404/reports/diff/index.html

rnveach: You changed many checks but only ran regression on one. You need to run regression on all the impacted checks.

romani: I can not believe that somebody user FQDN for good reason. It might, but in this case better not classify this update as minor.
@strkkk , please use "Pull #XXXX: " prefix

src/test/java/com/puppycrawl/tools/checkstyle/utils/AnnotationUtilTest.java
```diff
@@ -213,6 +215,75 @@ public void testContainsAnnotationListWithEmptyList() {
...
+            .that(result)
+            .isFalse();
+    }
+
+    @Test
+    public void testHasOverrideOtherAnnotation() {
+        final DetailAstImpl ast = new DetailAstImpl();
+        final DetailAstImpl modifiersAst = create(
+            TokenTypes.MODIFIERS,
+            create(
```
romani, line 243: such test is most toxic test we can have.
They increase illusion that we do testing/coverage and in same time engineers never respect such tests and update them up to new expectation in new code. So their value in long run is very low.
I presume we do this for pitest, do we have ability to extend tests to be Input based ?

romani (changes requested):
items:

strkkk:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
strkkk: I ran it on one check because it is the only one that has some change, for others there is no changes, only method extraction to util.

I added Pull prefix

src/test/java/com/puppycrawl/tools/checkstyle/utils/AnnotationUtilTest.java
strkkk: Is there some examples of tests, that use inputs for util method?

Such tests are added only for pitest, because they actual functionality is tested by checks UTs, but pitest ignores check for checks.

Utils accepts ASTs, alternative is to parse some input, navigate to some node and pass it to util method, but as for me it looks same as creating AST ourselves

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTypeCheck.java
strkkk: @rnveach see https://github.com/checkstyle/checkstyle/pull/11333#issuecomment-1049575520

src/test/java/com/puppycrawl/tools/checkstyle/utils/AnnotationUtilTest.java
romani: if problem come from pitest , we should find a way to resolve it by pitest config.
take a look at: 
https://github.com/checkstyle/checkstyle/blob/aec751dfba46e7ce6eae351cd93fde43f2cf3105/pom.xml#L3272-L3275
we just need to add some tests in pitest-util group for execution.

Historical note: ideally it would be better to run single pitest, so such problem will not exists at all, but unfortunately in this case execution of pitest takes more than 12h, that is why we did a split.

strkkk: done

src/test/java/com/puppycrawl/tools/checkstyle/utils/AnnotationUtilTest.java
```diff
@@ -61,6 +61,13 @@ public void testContainsAnnotationNull() {
         }
     }
 
+    @Test
+    public void testHasOverrideAnnotationNull() {
+        assertWithMessage("Should be false for null ast")
```
romani, line 66: Can we make this test to be Input based ?

romani (changes requested):
Item

src/test/java/com/puppycrawl/tools/checkstyle/utils/AnnotationUtilTest.java
romani: As we do simple refactoring, we should have already all Inputs with all edgy cases

romani: Pitest codding is failing

strkkk: removed. Pitest failure was random, there was no changes in check that failed. Now it is ok

romani (approved):
Ok to merge

rnveach (approved):

PR #11349, opened 2022-02-28
minor: change "nmancus1" to "Nick Mancuso" in release notes

nick-mancuso:
I noticed that I forgot to set my username to my real name in my local git setup.


romani (approved):

rnveach:
@nick-mancuso Did you correct your local so it won't be an issue for future notes?

nick-mancuso:
> @nick-mancuso Did you correct your local so it won't be an issue for future notes?

Yes

Issue #5770, opened 2018-05-03
AbstractJavadocCheck: switch logging to DetailNode


rnveach:
Identified at https://github.com/checkstyle/checkstyle/issues/5758#issuecomment-386167806 ,

We should change `AbstractJavadocCheck` logging from `log(int line...` to `log(DetailNode node...` as we wish to remove access to logging by line only for all checks.

When this issue will be solved, the test `XpathRegressionTest` and xdoc `config_filters` (section `SuppressionXpathFilter`) should be updated.
For now, this test ignores all `AbstractJavadocCheck` tests.

rnveach:
There is an issue doing this.

[To duplicate the AbstractCheck's log(ast)](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractCheck.java#L248) we need access to the [`messages`](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractCheck.java#L310) and [`tabWidth`](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractCheck.java#L52) fields, both of which are `private` and are not accessible in `AbstractJavadocCheck`.

romani:
We need to add new method first and start migration of Checks. Some Checks might be reasonable to stay on line:column. We also should give some period for all possible Checks outside our repo to migrate.

rnveach:
@romani When you say a new method, do you wish a similar one to the AST? If so, the one problem is the required methods mentioned at https://github.com/checkstyle/checkstyle/issues/5770#issuecomment-386693230 are private and would need to be made protected for us to access them. This would mean any check that extends them would have access too as `AbstractCheck` is not an intermediator like `TreeWalker` or `Checker`.

Another option is to just have a new method be a wrapper for the `log(line, column)` until we actually have xpath support for the DetailNode.

romani:
>just have a new method be a wrapper for the log(line, column) until we actually have xpath support for the DetailNode.

Yes

romani:
still needs to be updated:
```
com.puppycrawl.tools.checkstyle.api
AbstractCheckTest.ViolationCheck
AbstractFileSetCheckTest.ViolationFileSetCheck

com.puppycrawl.tools.checkstyle.checks.javadoc
AbstractJavadocCheck
[...]

PR #11363, opened 2022-03-05
Issue #10933: Checkstyle should force logging of AST in checks that require it

### LINKED ISSUE #10933: Checkstyle should force logging of AST in checks that require it
> @pbludov should we somehow limit the usage of "old" `log` method to help prevent this is in the future?

This is a good idea. Developers should be aware of this. We have a [check](https://github.com/checkstyle/checkstyle/blob/38df63e992c1270eac87d5e98d5d8cb281c4a20b/config/checkstyle_sevntu_checks.xml#L215) to forbid some method by its name and number of parameters.

 Unfortunately, the "old" `log` method is still used by Javadoc checks and filters. See #4830 for the details.
This means we should add an entry for `ForbidCertainMethodCheck` and suppress this violation for `*Flter` and `Javadoc*Check`.

_Originally posted by @pbludov in https://github.com/checkstyle/checkstyle/issues/10930#issuecomment-962656895_

-----------------

Confusion between `AbstractCheck#log` methods, as noticed in https://github.com/checkstyle/checkstyle/pull/10930#issuecomment-962656895, can be difficult for contributors to troubleshoot when `XPathRegression...` tests are failing due to suppressions not functioning as expected. It might be a good idea to add some sort of warning about this to the documentation for `AbstractViolationReporter#log` as well. Perhaps we could extend the current summary (`Log a message that has no column information.`) to something like `Log a message that has no column information. Usage of this log method should be restricted to FileSet checks, Javadoc checks, and filters.`

nick-mancuso:
Closes #10933 .

Example of error message:

```
[ERROR] Forbidden method invocation: com.puppycrawl.tools.checkstyle.api.AbstractCheck#log(int,java.lang.String,java.lang.Object[]) [Use of this log method is forbidden, please use AbstractCheck#log(DetailAST ast, String key, Object... args)]


```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1498,6 +1498,28 @@
                 <exclude>**/JavadocPropertiesGenerator.class</exclude>
                 <!-- generated classes, unfortunately use problematic api -->
                 <exclude>**/JavadocParser.class</exclude>
+                <!-- reason at https://github.com/checkstyle/checkstyle/issues/7759 -->
+                <exclude>**/RegexpCheck.class</exclude>
+                <!-- no code in file, no AST to log -->
+                <exclude>**/NoCodeInFileCheck.class</exclude>
+                <!-- until https://github.com/checkstyle/checkstyle/issues/5770 -->
+                <exclude>**/AbstractJavadocCheck.class</exclude>
+                <exclude>**/AtclauseOrderCheck.class</exclude>
+                <exclude>**/JavadocBlockTagLocationCheck.class</exclude>
+                <exclude>**/JavadocMethodCheck.class</exclude>
+                <exclude>**/JavadocMissingLeadingAsteriskCheck.class</exclude>
+                <exclude>**/JavadocMissingWhitespaceAfterAsteriskCheck.class</exclude>
+                <exclude>**/JavadocParagraphCheck.class</exclude>
+                <exclude>**/JavadocStyleCheck.class</exclude>
+                <exclude>**/JavadocTagContinuationIndentationCheck.class</exclude>
+                <exclude>**/JavadocTypeCheck.class</exclude>
+                <exclude>**/MissingDeprecatedCheck.class</exclude>
+                <exclude>**/NonEmptyAtclauseDescriptionCheck.class</exclude>
+                <exclude>**/RequireEmptyLineBeforeBlockTagGroupCheck.class</exclude>
+                <exclude>**/SingleLineJavadocCheck.class</exclude>
+                <exclude>**/SummaryJavadocCheck.class</exclude>
+                <exclude>**/WriteTagCheck.class</exclude>
+                <exclude>**/JavadocMetadataScraper.class</exclude>
               </excludes>
             </configuration>
           </execution>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1498,6 +1498,28 @@
                 <exclude>**/JavadocPropertiesGenerator.class</exclude>
                 <!-- generated classes, unfortunately use problematic api -->
                 <exclude>**/JavadocParser.class</exclude>
+                <!-- reason at https://github.com/checkstyle/checkstyle/issues/7759 -->
+                <exclude>**/RegexpCheck.class</exclude>
+                <!-- no code in file, no AST to log -->
+                <exclude>**/NoCodeInFileCheck.class</exclude>
+                <!-- until https://github.com/checkstyle/checkstyle/issues/5770 -->
```
nick-mancuso, line 1505: I will add comment in #5770 regarding suppressions introduced here.

pom.xml
```diff
@@ -1498,6 +1498,28 @@
...
+                <exclude>**/JavadocMethodCheck.class</exclude>
+                <exclude>**/JavadocMissingLeadingAsteriskCheck.class</exclude>
+                <exclude>**/JavadocMissingWhitespaceAfterAsteriskCheck.class</exclude>
+                <exclude>**/JavadocParagraphCheck.class</exclude>
+                <exclude>**/JavadocStyleCheck.class</exclude>
+                <exclude>**/JavadocTagContinuationIndentationCheck.class</exclude>
+                <exclude>**/JavadocTypeCheck.class</exclude>
+                <exclude>**/MissingDeprecatedCheck.class</exclude>
+                <exclude>**/NonEmptyAtclauseDescriptionCheck.class</exclude>
+                <exclude>**/RequireEmptyLineBeforeBlockTagGroupCheck.class</exclude>
```
romani, line 1518: It might be the only Check need to fix, all others are javadoc and we do no support suppression by ast for them.
So reporting by ast might be too early.
If you agree, please group all javadoc together and keep only java Check until this issue.

nick-mancuso: This is also a Javadoc check, all checks in this group are. I just copy pasted list from https://checkstyle.sourceforge.io/config_filters.html#SuppressionXpathFilter :)

rnveach: I believe the check being discussed here is `JavadocTypeCheck` as it is the only one of the 4 that extends `AbstractCheck`, but even this seems like an outdated javadoc regexp check.

strkkk (approved):

pom.xml
romani: Yes, it is javadoc check
https://github.com/checkstyle/checkstyle/blob/6f000bfbd10948a729befe4ec5cace79f30f2f5b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/RequireEmptyLineBeforeBlockTagGroupCheck.java#L103

Ok, all of them are same group

romani (approved):
Ok to merge

pbludov (approved):

rnveach (approved):

Issue #11365, opened 2022-03-06
FinalClassCheck: False positive with anonymous classes


Vyom-Yadav:
I have read check documentation: https://checkstyle.sourceforge.io/config_design.html#FinalClass
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Test.java

/var/tmp $ cat config.xml
```
```xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="TreeWalker">
        <module name="FinalClass"/>
    </module>
</module>
```
```
/var/tmp $ cat Test.java
```
```java
public class Test {

    class a { // expected no violation
        private a() {
        }
    }

    a obj = new a() {
    };
}
```
```bash
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:3:5: Class a should be declared as final. [FinalClass]
Audit done.
Checkstyle ends with 1 errors.
```
---

**Describe what you expect in detail.**
```bash
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.0-all.jar -c config.xml Test.java
Starting audit...
Audit done.
```
This was supposed to be fixed in #9357. Making the class final results in a compile-time error.

Vyom-Yadav:
I am on it.

romani:
fix is merged

PR #11369, opened 2022-03-06
Issue #11201: Moved methods common in UnusedLocalVariableCheck and FinalClassCheck to util

### LINKED ISSUE #11201: Move methods common in UnusedLocalVariableCheck and FinalClassCheck to util
From https://github.com/checkstyle/checkstyle/pull/10930#discussion_r785397301
To be addressed after merging #10930 
UnusedLocalVariableCheck (#10930) and FinalClassCheck (Check +  #10968) has some methods in common-

```
- extractQualifiedName(..)
- classNameMatchingCount(..)
- getQualifiedClassOrInterfaceName(DetailAST classAst) (known as `getQualifiedClassName` in final class)
- getQualifiedClassOrInterfaceName(String packageName, String outerClassQualifiedName, String className) (known as `getQualifiedClassName` in final class)
```
Also FinalClassCheck and UnusedLocalVariableCheck have a static nested class in common - `ClassDesc`. We may merge both of these classes and extract it somewhere, there are some overlapping properties.

When #10968 will be merged, extract these methods to util and try to remove the suppression for `ExcessiveClassLength` in UnusedLocalVariableCheck.

Vyom-Yadav:
Resolves #11201

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
@@ -34,7 +34,6 @@ import java.util.stream.Collectors;
 import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
-import com.puppycrawl.tools.checkstyle.api.FullIdent;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
 import com.puppycrawl.tools.checkstyle.checks.naming.AccessModifierOption;
 import com.puppycrawl.tools.checkstyle.utils.CheckUtil;
@@ -316,7 +315,7 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
             visitTypeDeclarationToken(ast);
         }
         else if (type == TokenTypes.PACKAGE_DEF) {
-            packageName = extractQualifiedName(ast.getFirstChild().getNextSibling());
+            packageName = CheckUtil.extractQualifiedName(ast.getFirstChild().getNextSibling());
         }
     }
 
@@ -394,18 +393,6 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
     private void visitLocalAnonInnerClass(DetailAST literalNewAst) {
         anonInnerAstToTypeDeclDesc.put(literalNewAst, typeDeclarations.peek());
         anonInnerClassHolders.add(getBlockContainingLocalAnonInnerClass(literalNewAst));
-    }
-
-    /**
-     * Get name of package and super class of anon inner class by concatenating
-     * the identifier values under {@link TokenTypes#DOT}.
-     * Duplicated, until <a>https://github.com/checkstyle/checkstyle/issues/11201</a>
-     *
-     * @param ast ast to extract superclass or package name from
-     * @return qualified name
-     */
-    private static String extractQualifiedName(DetailAST ast) {
-        return FullIdent.createFullIdent(ast).getText();
     }
 
     /**
@@ -554,7 +541,7 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
      */
     private TypeDeclDesc getSuperClassOfAnonInnerClass(DetailAST literalNewAst) {
         TypeDeclDesc obtainedClass = null;
-        final String shortNameOfClass = getShortNameOfAnonInnerClass(literalNewAst);
+        final String shortNameOfClass = CheckUtil.getShortNameOfAnonInnerClass(literalNewAst);
         if (packageName != null && shortNameOfClass.startsWith(packageName)) {
             final Optional<TypeDeclDesc> classWithCompletePackageName =
                     typeDeclAstToTypeDeclDesc.values()
@@ -576,26 +563,6 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
             }
         }
         return obtainedClass;
-    }
-
-    /**
-     * Get the short name of super class of anonymous inner class.
-     * Example-
-     * <pre>
-     * TestClass.NestedClass obj = new Test().new NestedClass() {};
-     * // Short name will be Test.NestedClass
-     * </pre>
-     *
-     * @param literalNewAst ast node of type {@link TokenTypes#LITERAL_NEW}
-     * @return short name of base class of anonymous inner class
-     */
-    public static String getShortNameOfAnonInnerClass(DetailAST literalNewAst) {
-        DetailAST parentAst = literalNewAst.getParent();
-        while (TokenUtil.isOfType(parentAst, TokenTypes.LITERAL_NEW, TokenTypes.DOT)) {
-            parentAst = parentAst.getParent();
-        }
-        final DetailAST firstChild = parentAst.getFirstChild();
-        return extractQualifiedName(firstChild);
     }
 
     /**
@@ -664,42 +631,31 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
     private static TypeDeclDesc getTheNearestClass(String outerTypeDeclName,
             List<TypeDeclDesc> typeDeclWithSameName) {
         return Collections.min(typeDeclWithSameName, (first, second) -> {
-            int diff = Integer.compare(
-                    typeDeclNameMatchingCount(outerTypeDeclName, second.getQualifiedName()),
-                    typeDeclNameMatchingCount(outerTypeDeclName, first.getQualifiedName()));
-            if (diff == 0) {
-                diff = Integer.compare(first.getDepth(), second.getDepth());
-            }
-            return diff;
+            return getTypeDeclarationNameMatchingCountDiff(outerTypeDeclName, first, second);
         });
     }
 
     /**
-     * Calculates and returns the type declaration name matching count.
+     * Get the difference between type declaration name matching count. If the
+     * difference between them is zero, then their depth is compared to obtain the result.
      *
-     * <p>
-     * Suppose our pattern class is {@code foo.a.b} and class to be matched is
-     * {@code foo.a.ball} then type declaration name matching count would be calculated by
-     * comparing every character, and updating main counter when we hit "." to prevent matching
-     * "a.b" with "a.ball". In this case type declaration name matching count
-     * would be equal to 6 and not 7 (b of ball is not counted).
-     * </p>
-     * Duplicated, until <a>https://github.com/checkstyle/checkstyle/issues/11201</a>
-     *
-     * @param patternClass class against which the given class has to be matched
-     * @param classToBeMatched class to be matched
-     * @return class name matching count
+     * @param outerTypeDeclName outer type declaration of anonymous inner class
+     * @param firstTypeDecl first input type declaration
+     * @param secondTypeDecl second input type declaration
+     * @return difference between type declaration name matching count
      */
-    private static int typeDeclNameMatchingCount(String patternClass, String classToBeMatched) {
-        final char packageSeparator = PACKAGE_SEPARATOR.charAt(0);
-        final int length = Math.min(classToBeMatched.length(), patternClass.length());
-        int result = 0;
-        for (int i = 0; i < length && patternClass.charAt(i) == classToBeMatched.charAt(i); ++i) {
-            if (patternClass.charAt(i) == packageSeparator) {
-                result = i;
-            }
+    private static int getTypeDeclarationNameMatchingCountDiff(String outerTypeDeclName,
+                                                               TypeDeclDesc firstTypeDecl,
+                                                               TypeDeclDesc secondTypeDecl) {
+        int diff = Integer.compare(
+            CheckUtil.typeDeclarationNameMatchingCount(
+                outerTypeDeclName, secondTypeDecl.getQualifiedName()),
+            CheckUtil.typeDeclarationNameMatchingCount(
+                outerTypeDeclName, firstTypeDecl.getQualifiedName()));
+        if (diff == 0) {
+            diff = Integer.compare(firstTypeDecl.getDepth(), secondTypeDecl.getDepth());
         }
-        return result;
+        return diff;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/utils/VariableDesc.java
```diff
@@ -0,0 +1,133 @@
...
+
+package com.puppycrawl.tools.checkstyle.utils;
+
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+/**
+ * Maintains information about the variable.
+ */
+public class VariableDesc {
```
Vyom-Yadav, line 28: Though this is just used by UnusedLocaVariable, it has the potential of being used by more, the main reason this was also extracted was due to the fact this is used by TypeDescDecl and I didn't want to give this public access in UnusedLocalVariable. (Then the class would have been accessed like `UnusedLocalVariableCheck.VariableDesc`)

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -319,4 +320,35 @@ public static boolean isBooleanLiteralType(final int tokenType) {
...
+     * Example-
+     * <pre>
+     * TestClass.NestedClass obj = new Test().new NestedClass() {};
+     * // Short name will be Test.NestedClass
+     * </pre>
+     *
+     * @param literalNewAst ast node of type {@link TokenTypes#LITERAL_NEW}
+     * @return short name of base class of anonymous inner class
+     */
+    public static String getShortNameOfAnonInnerClass(DetailAST literalNewAst) {
```
Vyom-Yadav, line 345: This is currently being used only by UnusedLocalVar but will be used by FinalClass for fixing #11365

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -319,4 +320,35 @@ public static boolean isBooleanLiteralType(final int tokenType) {
         return isTrue || isFalse;
     }
 
+    /**
```
pbludov, line 323: Why are these methods in this class? The Javadoc says:
```java
/**
 * Contains utility methods for tokens.
 *
 */
public final class TokenUtil {
```

src/main/java/com/puppycrawl/tools/checkstyle/utils/TypeDeclDesc.java
```diff
@@ -0,0 +1,234 @@
...
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+/**
+ * Maintains information about the type declaration.
+ * Any ast node of type {@link TokenTypes#CLASS_DEF} or {@link TokenTypes#INTERFACE_DEF}
+ * or {@link TokenTypes#ENUM_DEF} or {@link TokenTypes#ANNOTATION_DEF}
+ * or {@link TokenTypes#RECORD_DEF} is considered as a type declaration.
+ */
+public class TypeDeclDesc {
+
```
pbludov, line 35: This class does not look like a utility class. I see two ways:
Move this class to the package `com.puppycrawl.tools.checkstyle`,
or, create a base class for all checks that should use this class and put this class inside of it.

pbludov (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/utils/TypeDeclDesc.java
Vyom-Yadav: > create a base class for all checks that should use this class and put this class inside of it.

This class is used by FinalClass and UnusedLocalVariable, creating a base class for them isn't the best idea as one governs the design and the other governs the coding aspect, though I am ok with the former solution.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
Vyom-Yadav: Shouldn't be there, moved it.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TypeDeclDesc.java
nick-mancuso: I agree with @pbludov 's idea to move this class out of utils, but I wonder if we should make it some abstract class, with each check having it's own subclass with specific implementation details. I suspect this will also help with mutation coverage, and leave details specific to each check out of main implementation.

@Vyom-Yadav right now, are both checks using all methods and fields in this class?

Vyom-Yadav: > it some abstract class, with each check having it's own subclass with specific implementation details.

There is only one implementation of every method, though some classes do not use some of the methods.

> I suspect this will also help with mutation coverage 

Mutation coverage shouldn't be a problem with the current mutators
(+ REMOVE_CONDITIONALS), I tested it with REMOVE_CONDITIONALS as pitest-utils has that mutator commented out and every new line was covered.

> leave details specific to each check out of the main implementation.

This is a good idea, jumping from class to class to check out implementation can be tedious, but most of the fields are just boolean flags. Only `getUpdatedCopyOfVarStack(..)` has an implementation, `declaredAsFinal` and `declaredAsAbstract`are instantiated in the constructor.

> are both checks using all methods and fields in this class?

No only three fields are shared.

1. `qualifiedName`
2. `depth`
3. `typeDeclAst`

rest all the fields are particular to a check. (Some of them are not even applicable to the other check)

nick-mancuso: @Vyom-Yadav if you think abstract class is a good idea, can you make a rough sketch of what this class would look like, and what subclasses might look like, as proposal? Nothing as fancy as lucidcharts is needed, just something in markdown that takes a few minutes.

nick-mancuso: > Mutation coverage shouldn't be a problem

I should clarify; I meant that ideally, we would only have to add one class to pitest for abstract class, since one class would cover all abstract class code. Concrete class in check implementation would cover the rest in it's respective pitest.

Vyom-Yadav: > This is a good idea, jumping from class to class to check out implementation can be tedious, **BUT** most of the fields are just boolean flags.
>> if you think abstract class is a good idea, can you make a rough sketch of what this class would look like

I meant to say usually it is a good idea but in this case, I feel having an abstract class for this might be overkill (We add the complexity of class hierarchy) but if I had to make an implementation it would be something like-
```java
public abstract class AbstractTypeDeclDesc {
    private final String qualifiedName;
    private final int depth;
    private final DetailAST typeDeclAst;

    public AbstractTypeDeclDesc(String qualifiedName, int depth,
                                DetailAST typeDeclAst) {
        this.qualifiedName = qualifiedName;
        this.depth = depth;
        this.typeDeclAst = typeDeclAst;
    }
}
```
Then `FinalClass` will have a subclass called `ClassDesc` which will extend this and the fields, getter/setters it requires. Similarly, `UnusedLocalVar` will also have a subclass extending it and adding its custom fields/methods.

Vyom-Yadav: @nick-mancuso @pbludov Let's agree on a solution and move forward from there.

nick-mancuso: If this is all they have in common, then I do not see the point in combining the two. What benefits would this provide, except to avoid duplicating ~12 lines of code? I am not sure this is a good enough reason to make two checks dependent on some other class.

pbludov: @Vyom-Yadav, this issue is more complicated than I thought. Perhaps it's better to leave these classes as they are for now.

Maybe we need to add some entities to the `api` package to represent types (as we have `FullIdent` for ids).
Something that makes it easier to work with nested classes.
We need to discuss this more carefully.

Vyom-Yadav: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -284,54 +274,27 @@ private void registerNestedSubclassToOuterSuperClasses(String qualifiedClassName
      * Checks if there is a class with same name.
      *
      * @param className name of the class
-     * @param superClassName name of the super class
+     * @param superClsName name of the super class
```
nick-mancuso, line 277: Why change this? Previous naming convention was good.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -284,54 +274,27 @@ private void registerNestedSubclassToOuterSuperClasses(String qualifiedClassName
...
      * @noinspection CallToStringConcatCanBeReplacedByOperator
      */
-    private ClassDesc getNearestClassWithSameName(String className, String superClassName) {
+    private ClassDesc getNearestClassWithSameName(String className, String superClsName) {
         final String dotAndClassName = PACKAGE_SEPARATOR.concat(className);
         return innerClasses.entrySet().stream()
                 .filter(entry -> entry.getKey().endsWith(dotAndClassName))
                 .map(Map.Entry::getValue)
-                .min((first, second) -> {
+                .min((fir, sec) -> {
```
nick-mancuso, line 286: Same as above.

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
```diff
@@ -589,4 +593,90 @@ private static int lastIndexOfNonWhitespace(String line) {
...
+     * comparing every character, and updating main counter when we hit "." to prevent matching
+     * "a.b" with "a.ball". In this case type declaration name matching count
+     * would be equal to 6 and not 7 (b of ball is not counted).
+     * </p>
+     *
+     * @param patternClass class against which the given class has to be matched
+     * @param classToBeMatched class to be matched
+     * @return class name matching count
+     */
+    public static int typeDeclNameMatchingCount(String patternClass, String classToBeMatched) {
```
nick-mancuso, line 612: Why change previous name?

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
```diff
@@ -589,4 +593,90 @@ private static int lastIndexOfNonWhitespace(String line) {
...
+     * @param ast ast to extract superclass or package name from
+     * @return qualified name
+     */
+    public static String extractQualifiedName(DetailAST ast) {
+        return FullIdent.createFullIdent(ast).getText();
+    }
+
+    /**
+     * Get the short name of super class of anonymous inner class.
+     * Example-
```
nick-mancuso, line 665: `Example:`

nick-mancuso (changes requested):
Few from me:

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
Vyom-Yadav: The name is changed wrt to FinalClass but not wrt to UnusedLocalVar
https://github.com/checkstyle/checkstyle/blob/2b7942834e9ecba5b21095133c2c4e6295442535/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L693-L703

`identifierMatchingCount` is way too broad, this method only works with type declarations name, identifier can be of a normal variable too, the current name looks fine to me, but am open to better names.

nick-mancuso: I would prefer the full name then, since we are moving it to util class for general use if we can't think of something better, i.e. `typeDeclarationNameMatchingCount`, this is still shorter than some other method names in this class. As we move it out of private method, we should consider the name more carefully.

nick-mancuso: `lengthOfMatchInTypeDeclarationName`?

Vyom-Yadav: The full name will be better, (`typeDeclarationNameMatchingCount`) there are method chains  like `matcher.results().count();`, so `MatchingCount` is good to have in the end. What do you think?

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
Vyom-Yadav: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: Changed, more explanation comment below.

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
Vyom-Yadav: Changed to `typeDeclarationNameMatchingCount`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
```diff
@@ -664,42 +631,31 @@ private boolean hasSameNameAsSuperClass(String superClassName, TypeDeclDesc type
...
-        final char packageSeparator = PACKAGE_SEPARATOR.charAt(0);
-        final int length = Math.min(classToBeMatched.length(), patternClass.length());
-        int result = 0;
-        for (int i = 0; i < length && patternClass.charAt(i) == classToBeMatched.charAt(i); ++i) {
-            if (patternClass.charAt(i) == packageSeparator) {
-                result = i;
-            }
+    private static int getTypeDeclarationNameMatchingCountDiff(String outerTypeDeclName,
+                                                               TypeDeclDesc firstTypeDecl,
+                                                               TypeDeclDesc secondTypeDecl) {
```
Vyom-Yadav, line 649: This method was required as 
`CheckUtil.typeDeclarationNameMatchingCount(outerTypeDeclName, second.getQualifiedName())`
was exceeding line length and wrapping it on the next line resulted in exceeding max lambda length (10).

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
nick-mancuso: > matcher.results().count();

That's another idea; just have method return char array of longest match and call `Stream.of(results).count()` now that we have Java 11 :)

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
nick-mancuso: Resolved.

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
Vyom-Yadav: I feel it is fine right now, we would have to increase the number of parameters in the function plus the overhead of streams (though it won't bring any drastic performance difference) when we can simply count using the already present `for` loop.

nick-mancuso: Resolved.

pbludov (approved):

romani (approved):

romani:
Cirrus was relaunched

nick-mancuso (approved):

Issue #11370, opened 2022-03-06
Large overhead from using streams in TokenUtil.isOfType


SanderGielisse:
By doing some time analysis using VisualVM, I found that the `isOfType` method of the `TokenUtil` class taking quite a lot of time, as shown below.
![issue0](https://user-images.githubusercontent.com/4587552/156930498-c85d7ad3-dbdb-4a47-a862-52abfb3114fc.png)

The current implementation is;
```
    public static boolean isOfType(int type, int... types) {
        return Arrays.stream(types).anyMatch(tokenType -> tokenType == type);
    }
```

While the slight overhead of using streams (which was probably done to improve readability / maintainability) does not seem like a big issue at first sight, I found that, when executed on some larger projects, this `isOfType` method is called somewhere in the order of several **billion** times... Using the current implementation, I've found cases where around 30% of the time is spent inside of the stream logic (see screenshot). I have tested that a simple naive implementation reduces this to almost 0%. 

Note; this is not a full-benchmark from start-to-end as I stopped it after +/- 120 seconds. So the time spent inside of `isOfType` took 30% of the first 120 seconds of running Checkstyle (quite a lot...), which using the naive implementation dropped to near 0. If it is desired, I could look into doing further benchmarks to give more precise and definitive answers regarding the exact effect, but I hope this issue at least shows a naive implementation is more advantageous in general.

Opened a PR at https://github.com/checkstyle/checkstyle/pull/11371

nick-mancuso:
> By doing some time analysis using VisualVM

@SanderGielisse I am curious, what prompted your investigation?

rnveach:
From my local and running checkstyle on itself,
https://github.com/rnveach/checkstyle/commits/more_audits
````
    public static boolean isOfType(int type, int... types) {
        boolean matching = false;
        synchronized (instance) {
        Checker.GLOBAL_DISPATCHER.fireCustomStarted("isOfType");
        matching = Arrays.stream(types).anyMatch(tokenType -> tokenType == type);
        Checker.GLOBAL_DISPATCHER.fireCustomFinished("isOfType");
        }
        return matching;
    }
````

Output:
````
Run Time: 41.293

Format:
Name	Uses	Time	Min	Max	Average


Parses: (2)
Java	967	5.69	0.0	0.355	0.005
JavaDoc	6022	20.341	0.0	0.6	0.003


Customs: (1)
isOfType	2712864	0.546	0.0	0.002	0.0
````

[...]

PR #11371, opened 2022-03-06
Issue #11370: use naive implementation for isOfType

SanderGielisse:
This PR proposed to use a naive implementation for `TokenUtil.isOfType`, as discussed in #11370. This yields a significant speedup of Checkstyle run-time.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
@@ -281,7 +281,14 @@ public final class TokenUtil {
      * @return true if type matches one of the given types.
      */
     public static boolean isOfType(int type, int... types) {
-        return Arrays.stream(types).anyMatch(tokenType -> tokenType == type);
+        boolean matching = false;
+        for (int tokenType : types) {
+            if (tokenType == type) {
+                matching = true;
+                break;
+            }
+        }
+        return matching;
     }
 
     /**


rnveach:
I am not against this change. There was a topic recently that streams are slower. I do question the metrics mentioned in the issue as they don't line up with mine through an alternative method (time spent 30% versus 1.3%). Other factors could be in play, but I would think the numbers would be closer together.

romani:
>This yields a significant speedup of Checkstyle run-time.

Please show evidence.
We can same level of micro optimization in whole code base, there are numerous methods where old/native be beneficial in performance.
But we balancing performance and maintainability and usually sacrifice performance to keep code easy to read.

SanderGielisse:
> > This yields a significant speedup of Checkstyle run-time.
> 
> Please show evidence. We can same level of micro optimization in whole code base, there are numerous methods where old/native be beneficial in performance. But we balancing performance and maintainability and usually sacrifice performance to keep code easy to read.
> 
> Checkstyle is not executed in production runtime, so performance is not our top priority.

I understand. In the issue #11370 both me and @rnveach did some analysis. We both concluded that in the general case (running Checkstyle on Checkstyle using Checkstyle config) the speed-up is only around 1% on the total runtime. However, when running some specific case (running Checkstyle on Checkstyle using Google config), this yields in the order of several billion `isOfType` calls. The calls come from `OperatorWrapCheck.isNewLineModeViolation`, but my understanding of all the logic behind what calls what is insufficient at this point to provide any concrete answers as to why this is happening.

With regards to the speed up, I think the results below illustrate this well.

> `-c ${workspace_loc}\checkstyle\src\main\resources\google_checks.xml ${workspace_loc}\checkstyle --executeIgnoredModules`

With regards to your timings, I ran the following two commands and obtain the following VisualVM overviews;
`java -jar j_unfixed.jar -c ./google-checks.xml ./checkstyle --executeIgnoredModules`
issue_unfixed

`java -jar j_fixed.jar -c ./google-checks.xml ./checkstyle --executeIgnoredModules`
fixed

The total run-time went from 125 seconds to 85 seconds, a speedup of around 30%. In the naive version, the sampler does not even log `isOfType`, meaning that the time it spends in there is almost negligible. In the stream version, the time spend in there is around 34%, where 27% of which was on `IntPipeline.anyMatch()` (of the total runtime!). In the issue we discussed a possible reason to be the casting to non-primitive type Integer rather than int, which possibly internally traverses the entire array.

I understand that runtime is not a priority of Checkstyle, but I think that even if there exist only a couple specific cases (I do not know the exact domain of cases that trigger this) where the speedup is as significant as 30% of the total runtime, it is well-worth considering the naive approach over the stream based approach. With regards to readability, I'd even argue that the naive approach is as easily interpretable as the stream based one, if not easier. I will definitely not go as far as claiming that naive approaches are better in general, yet I feel like this is different.

rnveach (approved):

strkkk:
@romani I ran some jmh benchmarks with stream and loop. 
Numbers can be incorrect, but pattern is clear - loop works better. My bet is that boxing/autoboxing makes difference.
```
200 elements array, elem present
Benchmark               Mode  Cnt    Score    Error  Units 
BenchmarkLoop.loopFind  avgt   20   53,533 ?  1,851  ns/op
BenchmarkLoop.stream    avgt   20  509,482 ? 33,592  ns/op

20 elements array, elem present
Benchmark               Mode  Cnt    Score    Error  Units
BenchmarkLoop.loopFind  avgt   20   27,146 ?  3,290  ns/op
BenchmarkLoop.stream    avgt   20  102,416 ? 13,638  ns/op

20 elements array, elem may miss
Benchmark               Mode  Cnt   Score   Error  Units
BenchmarkLoop.loopFind  avgt   20  23,249 ? 1,036  ns/op
BenchmarkLoop.stream    avgt   20  95,916 ? 1,293  ns/op

200 elements array, elem may miss
Benchmark               Mode  Cnt    Score    Error  Units
BenchmarkLoop.loopFind  avgt   20   61,033 ?  7,252  ns/op
BenchmarkLoop.stream    avgt   20  647,019 ? 71,335  ns/op
```

strkkk (approved):

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -281,7 +281,14 @@ public static boolean isTypeDeclaration(int type) {
      * @return true if type matches one of the given types.
      */
     public static boolean isOfType(int type, int... types) {
-        return Arrays.stream(types).anyMatch(tokenType -> tokenType == type);
+        boolean matching = false;
```
pbludov, line 284: We can go further and add overloads with one or two int parameters. Like this
```java
public static boolean isOfType(int type, int first, int second) {
    return type == first || type == second;
}
```
The JIT compiler will definitely inline such methods .

SanderGielisse: > We can go further and add overloads with one or two int parameters. Like this
> 
> ```java
> public static boolean isOfType(int type, int first, int second) {
>     return type == first || type == second;
> }
> ```
> 
> The JIT compiler will definitely inline such methods .

Interesting idea, but are there no usages where `types` is constructed dynamically? Then I think this would not work.

With regards to performance, I would be curious to see how much the difference is, but my expectation is it to not be a major difference from the native loop. With regards to maintainability I do feel like this takes a major hit, as you'd need one such method for each length of `types`, but then I am unsure where exactly the performance / maintainability trade-off of Checkstyle lies.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -281,7 +281,14 @@ public static boolean isTypeDeclaration(int type) {
...
+        for (int t : types) {
```
romani, line 285: No single char names, name it "type"

romani (changes requested):
Why we do not violate single char name ?

Items:

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
nick-mancuso: https://github.com/checkstyle/checkstyle/blob/c1c41d49ca6829df3c0dcf3e8ca980838c8ef0bf/config/checkstyle_checks.xml#L758-L761

nick-mancuso: https://github.com/checkstyle/checkstyle/blob/c1c41d49ca6829df3c0dcf3e8ca980838c8ef0bf/config/checkstyle_checks.xml#L758-L761

romani: It probably makes sense to make it false. We like exact names.

nick-mancuso: According to the check documentation, this would also apply to traditional for-loop. I do not think that we should enforce this unless we make property false and suppress warnings for traditional for-loops.

romani: https://github.com/checkstyle/checkstyle/pull/11521

SanderGielisse: Changed to `foundType`, as `type` was already taken by the `type` to find a match for (in the method arguments).

romani: Please use tokenType as it was before

strkkk:
@SanderGielisse please squash commits into one, this is needed to merge changes

pbludov (approved):

romani:
update like:
```
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java b/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
index 9dd5c13..a5ff490 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
@@ -291,7 +291,23 @@ public final class TokenUtil {
         return matching;
     }
 
-    /**
+    public static boolean isOfType(int type, int first) {
+        return type == first;
+    }
+
+    public static boolean isOfType(int type, int first, int second) {
+        return type == first || type == second;
+    }
+
+    public static boolean isOfType(int type, int first, int second, int third) {
+        return type == first || type == second || type == third;
+    }
+
+    public static boolean isOfType(int type, int first, int second, int third, int forth) {
+        return type == first || type == second || type == third || type == forth;
+    }
+
+   /**
      * Determines if the AST belongs to the given types.
      *
      * @param ast the AST node to check
@@ -303,6 +319,22 @@ public final class TokenUtil {
         return ast != null && isOfType(ast.getType(), types);
     }
 
+    public static boolean isOfType(DetailAST ast, int first) {
+        return ast != null && isOfType(ast.getType(), first);
+    }
+
+    public static boolean isOfType(DetailAST ast, int first, int second) {
+        return ast != null && isOfType(ast.getType(), first, second);
+    }
+
+    public static boolean isOfType(DetailAST ast, int first, int second, int third) {
+        return ast != null && isOfType(ast.getType(), first,second, third);
+    }
+
+    public static boolean isOfType(DetailAST ast, int first, int second, int third, int forth) {
+        return ast != null && isOfType(ast.getType(), first,second, third, forth);
+    }
+
     /**
      * Determines if given AST is a root node, i.e. if the type
      * of the token we are checking is {@code COMPILATION_UNIT}.
```

would be good , but it force us to have more test coverage.
image

romani:
@pbludov , please approve if you agree - https://github.com/checkstyle/checkstyle/issues/11527

romani:
rabased and squashed to merge quicker ....

PR #11383, opened 2022-03-09
Issue #11213: Fixed false positive in SummaryJavadocCheck

### LINKED ISSUE #11213: False positive: SummaryJavadocCheck
From [comment1](https://github.com/checkstyle/checkstyle/pull/11051/files#r775167621) and [comment2](https://github.com/checkstyle/checkstyle/pull/11051/files#r775200466)
I have read check documentation: https://checkstyle.sourceforge.io/config_javadoc.html#SummaryJavadoc

```bash
/var/tmp $ javac Test2.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="SummaryJavadocCheck"/>
    </module>
</module>

/var/tmp $ cat Test2.java

/**
 * This is the actual summary.
 * {@summary This is wrong usage}
 */
public class Test2 {
}

/** // expected violation
 * This is summary
 */
class foo {
}

/var/tmp $ java $RUN_LOCALE -jar checkstyle-9.2.1-all.jar -c config.xml Test2.java
Starting audit...
[ERROR] /home/vyom/IdeaProjects/ConfigCheckstyle/src/TestP/Test2.java:1: Summary of Javadoc is missing an ending period. [SummaryJavadoc]
[ERROR] /home/vyom/IdeaProjects/ConfigCheckstyle/src/TestP/Test2.java:8: First sentence of Javadoc is missing an ending period. [SummaryJavadoc]
Audit done.
```
---

```
Starting audit...
[ERROR] /home/vyom/IdeaProjects/ConfigCheckstyle/src/TestP/Test2.java:8: First sentence of Javadoc is missing an ending period. [SummaryJavadoc]
Audit done.
```

If the first line is present then `@summary` is ignored, though a warning is given by the Javadoc tool.
```
Test.java:3: warning: invalid use of @summary
     * {@summary This is wrong usage}
```

Vyom-Yadav:
Resolves #11213 
Diff Regression config: https://gist.githubusercontent.com/Vyom-Yadav/98dceb63a79f4833e85fff9b2e1464a6/raw/9dd91cdba86d998f8fee9d271c0368258aa3deb6/my_checks.xml
Diff Regression projects: https://gist.githubusercontent.com/Vyom-Yadav/91807e6244cff60698d9e33e32ba2d51/raw/0f5845d0050e1b515298271e63c59a829d1d5aae/projects-to-test-on.properties

Final Report: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a906e6a_2022072602/reports/diff/index.html

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
@@ -251,8 +251,8 @@ public class SummaryJavadocCheck extends AbstractJavadocCheck {
     private static final Pattern HTML_ELEMENTS =
             Pattern.compile("<[^>]*>");
 
-    /** Period literal. */
-    private static final String PERIOD = ".";
+    /** Default period literal. */
+    private static final String DEFAULT_PERIOD = ".";
 
     /** Summary tag text. */
     private static final String SUMMARY_TEXT = "@summary";
@@ -274,7 +274,7 @@ public class SummaryJavadocCheck extends AbstractJavadocCheck {
     /**
      * Specify the period symbol at the end of first javadoc sentence.
      */
-    private String period = PERIOD;
+    private String period = DEFAULT_PERIOD;
 
     /**
      * Setter to specify the regexp for forbidden summary fragments.
@@ -658,7 +658,7 @@ public class SummaryJavadocCheck extends AbstractJavadocCheck {
      */
     private static String getFirstSentence(DetailNode ast) {
         final StringBuilder result = new StringBuilder(256);
-        final String periodSuffix = PERIOD + ' ';
+        final String periodSuffix = DEFAULT_PERIOD + ' ';
         for (DetailNode child : ast.getChildren()) {
             final String text;
             if (child.getChildren().length == 0) {
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
@@ -309,11 +309,14 @@ public class SummaryJavadocCheck extends AbstractJavadocCheck {
     @Override
     public void visitJavadocToken(DetailNode ast) {
         final Optional<DetailNode> inlineTag = getInlineTagNode(ast);
-        if (inlineTag.isPresent() && isSummaryTag(inlineTag.get())) {
-            validateSummaryTag(inlineTag.get());
+        final DetailNode inlineTagNode = inlineTag.orElse(null);
+        if (inlineTag.isPresent()
+            && isSummaryTag(inlineTagNode)
+            && isDefinedFirst(inlineTagNode)) {
+            validateSummaryTag(inlineTagNode);
         }
-        else if (inlineTag.isPresent() && isInlineReturnTag(inlineTag.get())) {
-            validateInlineReturnTag(inlineTag.get());
+        else if (inlineTag.isPresent() && isInlineReturnTag(inlineTagNode)) {
+            validateInlineReturnTag(inlineTagNode);
         }
         else if (!startsWithInheritDoc(ast)) {
             validateUntaggedSummary(ast);
@@ -354,6 +357,61 @@ public class SummaryJavadocCheck extends AbstractJavadocCheck {
             .filter(SummaryJavadocCheck::isInlineTagPresent)
             .findFirst()
             .map(SummaryJavadocCheck::getInlineTagNodeWithinHtmlElement);
+    }
+
+    /**
+     * Whether the {@code {@summary}} tag is defined first in the javadoc.
+     *
+     * @param inlineSummaryTag node of type {@link JavadocTokenTypes#JAVADOC_INLINE_TAG}
+     * @return {@code true} if the {@code {@summary}} tag is defined first in the javadoc
+     */
+    private static boolean isDefinedFirst(DetailNode inlineSummaryTag) {
+        boolean isDefinedFirst = true;
+        DetailNode previousSibling = JavadocUtil.getPreviousSibling(inlineSummaryTag);
+        while (previousSibling != null && isDefinedFirst) {
+            switch (previousSibling.getType()) {
+                case JavadocTokenTypes.TEXT:
+                    isDefinedFirst = previousSibling.getText().isBlank();
+                    break;
+                case JavadocTokenTypes.HTML_ELEMENT:
+                    isDefinedFirst = !isTextPresentInsideHtmlTag(previousSibling);
+                    break;
+                default:
+                    break;
+            }
+            previousSibling = JavadocUtil.getPreviousSibling(previousSibling);
+        }
+        return isDefinedFirst;
+    }
+
+    /**
+     * Whether some text is present inside the HTML element or tag.
+     *
+     * @param node DetailNode of type {@link JavadocTokenTypes#HTML_TAG}
+     *             or {@link JavadocTokenTypes#HTML_ELEMENT}
+     * @return {@code true} if some text is present inside the HTML element or tag
+     */
+    public static boolean isTextPresentInsideHtmlTag(DetailNode node) {
+        DetailNode nestedChild = JavadocUtil.getFirstChild(node);
+        if (node.getType() == JavadocTokenTypes.HTML_ELEMENT) {
+            nestedChild = JavadocUtil.getFirstChild(nestedChild);
+        }
+        boolean isTextPresentInsideHtmlTag = false;
+        while (nestedChild != null && !isTextPresentInsideHtmlTag) {
+            switch (nestedChild.getType()) {
+                case JavadocTokenTypes.TEXT:
+                    isTextPresentInsideHtmlTag = !nestedChild.getText().isBlank();
+                    break;
+                case JavadocTokenTypes.HTML_TAG:
+                case JavadocTokenTypes.HTML_ELEMENT:
+                    isTextPresentInsideHtmlTag = isTextPresentInsideHtmlTag(nestedChild);
+                    break;
+                default:
+                    break;
+            }
+            nestedChild = JavadocUtil.getNextSibling(nestedChild);
+        }
+        return isTextPresentInsideHtmlTag;
     }
 
     /**
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/summaryjavadoc/InputSummaryJavadocIncorrect3.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/summaryjavadoc/InputSummaryJavadocIncorrect3.java
@@ -1,0 +1,90 @@
+/*
+SummaryJavadoc
+violateExecutionOnNonTightHtml = (default)false
+forbiddenSummaryFragments = (default)^$
+period = (default).
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.javadoc.summaryjavadoc;
+
+/**
+ * This is the real summary.
+ * {@summary This is not the summary}
+ */
+public class InputSummaryJavadocIncorrect3 {
+
+    /**
+     * {@summary This is the real summary.}
+     * {@summary This is not the summary}
+     */
+    public void testAnotherSummaryTag() {
+    }
+
+    /**
+     * <p>This is summary.</p>
+     * {@summary This is not the summary}
+     */
+    public void testHtmlTags() {
+    }
+
+    /**
+     * <p><i></i></p>
+     * {@summary This is summary}
+     */
+    // violation 2 lines above 'Summary of Javadoc is missing an ending period'
+    public void testHtmlTags2() {
+    }
+
+    // violation below 'Summary javadoc is missing.'
+    /**
+     * @param a parameter
+     * {@summary Wrong usage}
+     */
+    public void method(int a) {
+    }
+
+    // violation below 'First sentence of Javadoc is missing an ending period'
+    /**
+     * <p> This is the summary {@summary This is not summary.}
+     */
+    public void testParagraphTag() {
+    }
+
+    /**
+     * <pre></pre>
+     * {@summary This is summary}
+     */
+    // violation 2 lines above 'Summary of Javadoc is missing an ending period'
+    public void testPreTag() {
+    }
+
+    // violation below 'First sentence of Javadoc is missing an ending period'
+    /**
+     * <i>This will be in italics and is the summary</i>
+     * {@summary This is not summary}
+     */
+    public void testOtherTags() {
+    }
+
+    /**
+     * <br>
+     * <br>
+     * {@summary Line breaks will be there but this will be the summary}
+     */
+    // violation 2 lines above 'Summary of Javadoc is missing an ending period.'
+    public void testBreakTag() {
+    }
+
+    // Until https://github.com/checkstyle/checkstyle/issues/11425
+    // violation below 'Summary javadoc is missing'
+    /**
+     * <ul>
+     *   <li><i>This is the summary</i></li>
+     * </ul>
+     * {@summary This is not the summary.}
+     */
+    public void testMultipleElements() {
+    }
+}
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/summaryjavadoc/InputSummaryJavadocInlineForbidden.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/summaryjavadoc/InputSummaryJavadocInlineForbidden.java
@@ -110,12 +110,11 @@ public class InputSummaryJavadocInlineForbidden {
     void foo12() {
     }
 
+    // violation below 'First sentence of Javadoc is missing an ending period'
     /**
      * Sentence starts as a plain text sentence
      * {@summary ... but ends in the summary tag}
      */
-    // violation 2 lines above 'Summary .* missing an ending period.'
-    // until https://github.com/checkstyle/checkstyle/issues/11213
     public class TestClass {}
 
     /**
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheckTest.java
@@ -57,7 +57,9 @@ public class SummaryJavadocCheckTest extends AbstractModuleTestSupport {
 
     @Test
     public void testInlineCorrect() throws Exception {
-        final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
+        final String[] expected = {
+            "112: " + getCheckMessage(MSG_SUMMARY_FIRST_SENTENCE),
+        };
 
         verifyWithInlineConfigParser(
                 getPath("InputSummaryJavadocInlineCorrect.java"), expected);
@@ -81,7 +83,6 @@ public class SummaryJavadocCheckTest extends AbstractModuleTestSupport {
             "142: " + getCheckMessage(MSG_SUMMARY_FIRST_SENTENCE),
             "147: " + getCheckMessage(MSG_SUMMARY_JAVADOC_MISSING),
             "150: " + getCheckMessage(MSG_SUMMARY_JAVADOC_MISSING),
-            "155: " + getCheckMessage(MSG_SUMMARY_MISSING_PERIOD),
         };
         verifyWithInlineConfigParser(
                 getPath("InputSummaryJavadocIncorrect.java"), expected);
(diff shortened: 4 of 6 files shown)


Vyom-Yadav:
Github, generate report.

Vyom-Yadav:
Github, generate report

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/summaryjavadoc/InputSummaryJavadocInlineForbidden.java
```diff
@@ -114,8 +114,7 @@ void foo12() {
      * Sentence starts as a plain text sentence
      * {@summary ... but ends in the summary tag}
      */
-    // violation 2 lines above 'Summary .* missing an ending period.'
-    // until https://github.com/checkstyle/checkstyle/issues/11213
+    // violation 2 lines above 'Summary of Javadoc is missing an ending period.'
```
Vyom-Yadav, line 117: When the first sentence doesn't end with a period,`@summary` is also included.
eg-
```
/**
 * Sentence starts as a plain text sentence
 *
 *
 * 
 * {@summary ... but ends in the summary tag.}
 */
public class InputSummaryJavadocIncorrect3 {
}
``` 
Screenshot from 2022-03-10 13-51-02

Vyom-Yadav:
Github, generate report

Vyom-Yadav:
The report hasn't got much, if there is something, in particular, you would like me to test, or some project which uses Javadoc extensively, feel free to mention it.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -304,7 +304,7 @@ public void setPeriod(String period) {
     @Override
     public void visitJavadocToken(DetailNode ast) {
         final Optional<DetailNode> inlineSummaryTag = getInlineSummaryTag(ast);
-        if (inlineSummaryTag.isPresent()) {
+        if (inlineSummaryTag.isPresent() && isDefinedFirst(inlineSummaryTag.get())) {
```
romani, line 307: Is it better to update getInlineSummaryTag(ast) to return nothing if summary tag is not first ?

romani (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
Vyom-Yadav: I feel it would be better to keep these methods separate, the first one just gets the inline summary tag, and the second one checks whether it is a suitable candidate for getting into the summary. 
This makes a clear distinction between just having a summary tag and it being applicable for getting into the actual summary part.

Vyom-Yadav:
@romani 
Diff report for world-edit (clean)- https://vyom-yadav.github.io/DiffReport/diffSummaryJavadoc/index.html

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
romani: If we do validation of incorrectly placed tag , I would agree with you, but we just need to search for tag in first place only, if not found, it is not what we need ( no matter where it is located).

@strkkk , can we ask for second opinion from you?

Vyom-Yadav: > If we do validation of incorrectly placed tag , I would agree with you, but we just need to search for tag in first place only, if not found, it is not what we need ( no matter where it is located).

Hmm, now I think about it, I feel that would be a better design (More abstract and easier to debug and manage), I changed to it.

Vyom-Yadav:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -609,7 +652,7 @@ private static String getStringInsideTag(String result, DetailNode detailNode) {
      */
     private static String getFirstSentence(DetailNode ast) {
         final StringBuilder result = new StringBuilder(256);
-        final String periodSuffix = PERIOD + ' ';
+        final String periodSuffix = DEFAULT_PERIOD + ' ';
```
romani, line 655: It might be unrelated bug but this should be `period`
As in Japan language period is different symbol.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -339,12 +339,55 @@ else if (!period.isEmpty()) {
...
+     * @return {@code true} if the {@code {@summary}} tag is defined first in the javadoc
+     */
+    private static boolean isDefinedFirst(DetailNode inlineSummaryTag) {
+        boolean isDefinedFirst = true;
+        DetailNode previousSibling = JavadocUtil.getPreviousSibling(inlineSummaryTag);
+        while (previousSibling != null && isDefinedFirst) {
+            final int siblingType = previousSibling.getType();
+            if (siblingType == JavadocTokenTypes.TEXT) {
+                final String previousSiblingText = previousSibling.getText();
+                if (previousSiblingText.contains(DEFAULT_PERIOD)) {
```
romani, line 361: `period` field should be used.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -339,12 +339,55 @@ else if (!period.isEmpty()) {
...
+                    isDefinedFirst = false;
+                }
+            }
+            else if (siblingType == JavadocTokenTypes.HTML_ELEMENT) {
+                final DetailNode firstChild = JavadocUtil.getFirstChild(previousSibling);
+                final int firstChildType = firstChild.getType();
+                if (firstChildType == JavadocTokenTypes.PARAGRAPH
+                    || firstChildType == JavadocTokenTypes.HTML_TAG
+                    && isPreTag(firstChild)) {
```
romani, line 370: Why pre tag is so special? Any tag can be empty.

romani (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
romani: Please create Input to catch this problem and prove that it will not happen.

Vyom-Yadav: Yes, this is a bug, but we have a problem here.
If the locale is set to English then Javadoc will only consider `. ` as a period, eg-
```
/**                                               
 * This is 1.0 version. There is no other version.
 */                                               
```
The summary will be `This is 1.0 version.` and not `This is 1.`, this logic is not transferable to the Japanese language (according to google translate). There the period symbol is `。` (yes it has that extra space already associated with it), so the `period + ' '` logic won't work for it.
eg-
```
/**                            
 * これは1.0バージョンです。他のバージョンはありません。
 */                            
```
There is no extra space after the `period` symbol, the period symbol itself consists of that extra space. The only solution I see is to make a switch for `period` values and adjust logic based on that. We might like to research it a bit as [java supports many locales](https://www.oracle.com/java/technologies/javase/jdk8-jre8-suported-locales.html).


`. ` = `.` +  ` `
`。` = `。`

Vyom-Yadav: I added logic for the Japanese language.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
Vyom-Yadav: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
Vyom-Yadav:
```
/**
 * <pre>This is inside pre without period</pre>
 * {@summary This wont be included in summary}
 */
public class Test {
}
```
Screenshot from 2022-03-13 10-09-11

```
/**
 * <i>This is inside i tag with period.</i>
 * {@summary This would be inside the summary}
 */
public class Test {
}
```
Screenshot from 2022-03-13 10-11-30

There might be more tags like `pre`, I identified only `<p>` and `<pre>`.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
Vyom-Yadav: The pitest will fail until #11407, I feel it would be better to fix this in #11407 and not include this in the current PR.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
strkkk: I see now there are 2 methods, looks fine for me. 

I suggest to change unnecessary if to one more step in optional transformation. I mean, instead of 
```
        if (node.isPresent() && (!isSummaryTag(node.get()) || !isDefinedFirst(node.get()))) {
            node = Optional.empty();
        }
```
you can just do 
```
return Arrays.stream(javadoc.getChildren())
                .filter(SummaryJavadocCheck::isInlineTagPresent)
                .findFirst()
                .map(SummaryJavadocCheck::getInlineTagNodeWithinHtmlElement);
.filter(node -> !isSummaryTag(node) || !isDefinedFirst(node));
```
this will be easier

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/summaryjavadoc/InputSummaryJavadocInlineForbidden.java
romani: I think it is big in javadoc tool. It is ok to consider summary to be text before inlined summary tag

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
romani: >If the locale is set to English then Javadoc will only consider . as a period

We do not look at locale, period symbol is property of Check. Japanese people set it explicitly.

romani: >yes it has that extra space already associated with it)

We can do space check only if period character is not unicode.

Vyom-Yadav: Yes, let's keep this separate, will add this issue to #11407

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/summaryjavadoc/InputSummaryJavadocInlineForbidden.java
Vyom-Yadav: I feel it is better to align ourselves with the current behavior of the Javadoc tool, if it is truly a bug then we should either report it or search for the bug if it is already present and add an until comment.

romani: @Vyom-Yadav , can you report defect on jdk (javadoc tool) ?
It is really weird that location of `.` make inline tag work differently.

meanwhile I would vote to violate this example and demand dot before inline tag.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
romani: it is same as https://github.com/checkstyle/checkstyle/pull/11383#discussion_r826927226 , lets do violation on `<pre>This is inside pre without period</pre>` code.
if any text started javadoc before inlinetag, we consider it as summary. Any inlined tag after should not be considered.  

no violation on : `<i>This is inside i tag with period.</i>`

As you notice, I am ignoring inlined tag at all for such cases.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/summaryjavadoc/InputSummaryJavadocInlineForbidden.java
Vyom-Yadav: I have filed a report in Oracle JDK, I wasn't able to file a bug report in OpenJDK as I don't have sufficient permissions to do so, oracle doesn't make the issue publically visible, they have given me an internal reference ID and will contact me on mail for further steps or inquiry.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
Vyom-Yadav: I have filed a report, see comment above, what should be done with this PR till then, should go with our presumed bug (change implementation) or the current working?

romani: If at least few character(not html tags) placed before summary inlined tag We consider this character as summary and ignore inlined tag.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/summaryjavadoc/InputSummaryJavadocIncorrect3.java
```diff
@@ -0,0 +1,78 @@
...
+    // violation below 'Summary javadoc is missing.'
+    /**
+     * @param a parameter
+     * {@summary Summary Will be empty}
+     */
+    public void method(int a) {
+    }
+
+    /**
+     * <p> This is {@summary the summary.}
```
romani, line 49: Violation should be there.

romani (changes requested):
Item

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
Vyom-Yadav: Done

Vyom-Yadav:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
Vyom-Yadav: > this will be easier

The condition won't be quite the same but good observation. Changed it.

Vyom-Yadav:
Github, generate report

romani:
@Vyom-Yadav , please rebase and resolve conflict

romani (approved):
Ok to merge

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -334,15 +334,64 @@ else if (!period.isEmpty()) {
...
+        boolean isTextPresentInsideHtmlTag = false;
+        while (nestedChild != null) {
+            final int nestedChildType = nestedChild.getType();
+            if (nestedChildType == JavadocTokenTypes.TEXT
+                && !nestedChild.getText().isBlank()
+                || (nestedChildType == JavadocTokenTypes.HTML_TAG
+                    || nestedChildType == JavadocTokenTypes.HTML_ELEMENT)
+                && isTextPresentInsideHtmlTag(nestedChild)) {
+                isTextPresentInsideHtmlTag = true;
+                break;
```
strkkk, line 390: to make it consistent with previous method's style (no break), it can be
```
boolean isTextPresent = false;
while (nestedChild != null && !isTextPresent) {
   final int nestedChildType = nestedChild.getType();
   isTextPresent = nestedChildType == JavadocTokenTypes.TEXT
                && !nestedChild.getText().isBlank()
                || (nestedChildType == JavadocTokenTypes.HTML_TAG
                    || nestedChildType == JavadocTokenTypes.HTML_ELEMENT
}
```

what do you think?

Vyom-Yadav: Good suggestion, done.

Vyom-Yadav:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -356,6 +358,56 @@ else if (!period.isEmpty()) {
...
+        if (node.getType() == JavadocTokenTypes.HTML_ELEMENT) {
+            nestedChild = JavadocUtil.getFirstChild(nestedChild);
+        }
+        boolean isTextPresentInsideHtmlTag = false;
+        while (nestedChild != null && !isTextPresentInsideHtmlTag) {
+            final int nestedChildType = nestedChild.getType();
+            isTextPresentInsideHtmlTag = nestedChildType == JavadocTokenTypes.TEXT
+                && !nestedChild.getText().isBlank()
+                || (nestedChildType == JavadocTokenTypes.HTML_TAG
+                || nestedChildType == JavadocTokenTypes.HTML_ELEMENT)
```
strkkk, line 404: please add indentation. Without it this line looks like it belongs to main condition chain.
As a side note - may be switch will look better than 5 conditions? e.g.
```
switch(nestedChild.getType()) {
   case JavadocTokenTypes.TEXT:
       isTextPresentInsideHtmlTag = !nestedChild.getText().isBlank();
       break;
   case JavadocTokenTypes.HTML_TAG:
   case JavadocTokenTypes.HTML_ELEMENT:
        isTextPresentInsideHtmlTag = isTextPresentInsideHtmlTag(nestedChild);
        break;
   default:
      isTextPresentInsideHtmlTag =false;
      break;
}
```
it may be a bit more verbose but definitely more readable, as for me

strkkk: same applies for method above, can be like
```
swithc(previousSibling.getType()) {
            case JavadocTokenTypes.TEXT: {
                isDefinedFirst previousSibling.getText().isBlank(); break;
                }
            }
            case JavadocTokenTypes.HTML_ELEMENT:
                isDefinedFirst = !isTextPresentInsideHtmlTag(previousSibling); break;
            }
            default:  isDefinedFirst = true;
```

what you think?

Vyom-Yadav:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
Vyom-Yadav: Good suggestion, implemented.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -309,7 +309,9 @@ public void setPeriod(String period) {
     @Override
     public void visitJavadocToken(DetailNode ast) {
         final Optional<DetailNode> inlineTag = getInlineTagNode(ast);
-        if (inlineTag.isPresent() && isSummaryTag(inlineTag.get())) {
+        if (inlineTag.isPresent()
+            && isSummaryTag(inlineTag.get())
+            && isDefinedFirst(inlineTag.get())) {
```
strkkk, line 314: I suggest to refactor this a bit. Instead of checking for node presense and invoking `get()` a million times we can use java11 `Optional#ifPresentOrElse`
in our case it will look like
```
inlineTag.ifPresentOrElse(
    node -> hereIsExtractedMethodForNonNullableNode(node),
    () -> validateUntaggedSummary(ast)
)
```

?

Vyom-Yadav: Yes, `get()` was called too many times, I extracted it to a variable.
The reason I didn't use `ifPresentOrElse` earlier was due to the logic, the current logic allows us to go to the last else if branch even if the inline tag is present, which `ifPresentOrElse` doesn't allow.

strkkk (approved):

strkkk:
> Final Report(clean): https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/e4ff34b_2022055349/reports/diff/index.html

@Vyom-Yadav usually report is "clean" when there is no diff to save some time for reviewer. Your report has diff, therefore it is not clean.

pbludov (approved):

PR #11390, opened 2022-03-12
minor: Updated Inputs for LeftCurlyCheck

MUzairS15:
Added Inputs with emoji

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/LeftCurlyCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/LeftCurlyCheckTest.java
@@ -455,6 +455,44 @@ public class LeftCurlyCheckTest extends AbstractModuleTestSupport {
     }
 
     @Test
+    public void testLeftCurlyWithEmoji() throws Exception {
+        final String[] expected = {
+            "17:32: " + getCheckMessage(MSG_KEY_LINE_BREAK_AFTER, "{", 32),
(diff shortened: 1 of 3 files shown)


nick-mancuso:
Check requires update, we do not need to close/ open new PRs for each check, you can just edit the PR title and description accordingly.

```
➜  java javac Test.java
➜  java cat Test.java
public class Test {
    void method1()
    /* 😊*/ { }
    void method2()
    { }
}
➜  java cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
    <module name="TreeWalker">
        <module name="LeftCurly">
        </module>
    </module>
</module>
➜  java java -jar checkstyle-10.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /home/nick/IdeaProjects/Tester/src/main/java/Test.java:5:5: '{' at column 5 should be on the previous line. [LeftCurly]
Audit done.


```

MUzairS15:
Thanks for reporting @nick-mancuso, still not getting why it didn't show me🤔

nick-mancuso:
> Thanks for reporting @nick-mancuso, still not getting why it didn't show me🤔

I should’ve made comments above more clear, my edit was to show that I made a mistake, and that it is existing behavior of check is the same with and without emoji in the comment. In my first example, I failed to place a comment in front of the curly braces. The comment in front of the curly braces is what is causing the check to miss the violation, not the emoji in it.


So, we have existing bug in check, and we can just update test inputs and make an issue if there isn’t one already.

romani (approved):

romani:
Please create issue

Issue #7546, opened 2020-02-12
Extend IT tests to able to load few modules from configs


romani:
discussed in https://github.com/checkstyle/checkstyle/pull/7542#discussion_r378190853

we need to extend AbstractGoogleModuleTestSupport to allow load few modules from config file.

In referenced PR we need to load together for each test:
```
<module name="RightCurly">
    <property name="id" value="RightCurlyAlone"/>
    <property name="option" value="alone"/>
    <property name="tokens"
     value="CLASS_DEF, METHOD_DEF, CTOR_DEF, LITERAL_FOR, LITERAL_WHILE, STATIC_INIT,
            INSTANCE_INIT, ANNOTATION_DEF, ENUM_DEF"/>
</module>
<module name="SuppressionXpathSingleFilter">
  <!-- suppresion is required till https://github.com/checkstyle/checkstyle/issues/7541 -->
  <property name="id" value="RightCurlyAlone"/>
  <property name="query" value="//RCURLY[parent::SLIST[count(./*)=1]
                                         or parent::OBJBLOCK[count(./*)=2]]"/>
</module>
```

**Attention**: `id` property in Filter is occupied for business logic meaning, so load of specific Filter might be problematic. It might require update in Filters to make `id` be same as in Check - just identifier.

rnveach:
> we need to extend AbstractGoogleModuleTestSupport to allow load few modules from config file.

Why only a few modules and not the entire config?

romani:
load of whole config might result in big list of `expected` violations, I think it will be hard to maintain such tests. I just keep original ideas of `it` to be module based.

Many years ago I thought that I will easily find open source project(s) that completely follow google style, so it will be our "no violations" test base .... it was naive.

We might create new set of ITs to keep input files and keep violations of whole config, I think it should be separate set ITs.

rnveach:
> load of whole config might result in big list of expected violations

I don't know how hard it will be to make inputs that strict, but it may be good to show that the entire input is compliant except the spots we tell them it will produce violations.

> `<module name="RightCurly">`
`<property name="id" value="RightCurlyAlone"/>`

> `<module name="SuppressionXpathSingleFilter">`
`<property name="id" value="RightCurlyAlone"/>`

We should be able to find config by its module name (like we do now), and if it contains an `id`, find other modules that have the same `id`. The above example shows they both use the same `id` in a property.

romani:
I always ok to have more tests. Please create new issue on this, if you want it.

[...]

PR #11400, opened 2022-03-12
Issue #7546: Extend IT tests to able to load few modules from configs

Binita-tech:
#7546

### DIFF
--- a/src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
+++ b/src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
@@ -130,6 +130,47 @@ public abstract class AbstractGoogleModuleTestSupport extends AbstractItModuleTe
     }
 
     /**
+     * Returns list of {@link Configuration} instance for the list of module name.
+     * This implementation uses {@link #getModuleById(String)} method inside.
+     *
+     * @param moduleIds list of module id.
+     * @return list of {@link Configuration} instances present inside TreeWalker.
+     * @throws CheckstyleException if there is a problem retrieving the module by id.
+     */
+    protected static List<Configuration>
+        getModuleByIds(List<String> moduleIds) throws CheckstyleException {
+        final List<Configuration> result = new ArrayList<>();
+        for (String moduleId: moduleIds) {
+            result.addAll(getModuleById(moduleId));
+        }
+        return result;
+    }
+
+    /**
+     * Returns list of {@link Configuration} instances for the the modules
+     * matching id inside TreeWalker.
+     *
+     * @param moduleId list of module id.
+     * @return list of {@link Configuration} instances inside TreeWalker for the given module ids.
+     * @throws CheckstyleException if there is a problem retrieving the module or config.
+     */
+    protected static List<Configuration>
+        getModuleById(String moduleId) throws CheckstyleException {
+        final List<Configuration> result = new ArrayList<>();
+        for (Configuration currentConfig : CONFIGURATION.getChildren()) {
+            if ("TreeWalker".equals(currentConfig.getName())) {
+                for (Configuration moduleConfig : currentConfig.getChildren()) {
+                    if (moduleConfig.getAttributeNames("id")
+                            && moduleConfig.getAttribute("id").equals(moduleId)) {
+                        result.add(moduleConfig);
+                    }
+                }
+            }
+        }
+        return result;
+    }
+
+    /**
      * Returns a list of all {@link Configuration} instances for the given module name.
      *
      * @param moduleName module name.
--- a/src/it/java/com/google/checkstyle/test/chapter4formatting/rule413emptyblocks/EmptyBlockTest.java
+++ b/src/it/java/com/google/checkstyle/test/chapter4formatting/rule413emptyblocks/EmptyBlockTest.java
@@ -24,6 +24,7 @@ import org.junit.jupiter.api.Test;
 import com.google.checkstyle.test.base.AbstractGoogleModuleTestSupport;
 import com.puppycrawl.tools.checkstyle.api.Configuration;
 import com.puppycrawl.tools.checkstyle.checks.blocks.EmptyBlockCheck;
+import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 
 public class EmptyBlockTest extends AbstractGoogleModuleTestSupport {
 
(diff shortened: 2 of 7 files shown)


src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
```diff
@@ -129,6 +129,52 @@ else if (moduleId == null) {
...
+    /**
+     * Returns list of {@link Configuration} instance for the list of module name.
+     * This implementation uses {@link #getModuleById(String)} method inside.
+     *
+     * @param moduleIds list of module id.
+     * @return list of {@link Configuration} instances present inside TreeWalker.
+     * @throws CheckstyleException if there is a problem retrieving the module by id.
+     */
+    protected static List<Configuration>
+        getModuleByIds(List<String> moduleIds) throws CheckstyleException {
```
rnveach, line 141: Where is this method used?

I am not seeing it, so I can't say why we need `List` and not `String...`.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
```diff
@@ -129,6 +129,52 @@ else if (moduleId == null) {
...
+    /**
+     * Returns list of {@link Configuration} instances for the the modules
+     * matching id inside TreeWalker.
+     *
+     * @param moduleId list of module id.
+     * @return list of {@link Configuration} instances inside TreeWalker for the given module ids.
+     * @throws CheckstyleException if there is a problem retrieving the module or config.
+     */
+    protected static List<Configuration>
+        getModuleById(String moduleId) throws CheckstyleException {
```
rnveach, line 158: Same, I am not seeing where this is used.

src/it/java/com/google/checkstyle/test/chapter4formatting/rule413emptyblocks/EmptyBlockTest.java
```diff
@@ -92,4 +93,14 @@ public void testEmptyBlockCatch() throws Exception {
         verify(checkConfig, filePath, expected, warnList);
     }
 
+    @Test
+    public void testEmptyBlockRightCurlyAlone() throws Exception {
+        final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
+        final Configuration checkConfig = getModuleConfig("EmptyBlock");
```
rnveach, line 99: Why this check for this issue? It doesn't have multiple modules from the config.
https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml#L68
Issue talks about RightCurly, so I would expect to see a test for that unless you can explain what i am missing.
https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml#L85-L104

rnveach (changes requested):

src/it/java/com/google/checkstyle/test/chapter4formatting/rule413emptyblocks/EmptyBlockTest.java
Binita-tech: @rnveach, I had taken reference from the previous PR, Do I have to write test cases for RightCurly as it has multiple modules?
Please guide me, this is the first time I am working on these files.

rnveach: You need to either write tests or write usages showing that this change is being used somewhere and it is working appropriately.

From what I have seen, nothing is using `getModuleByIds`, so we can't see that your changes are working and behaving appropriately.

Binita-tech:
```
    @Test
    public void testEmptyBlockRightCurlyAlone() throws Exception {
        final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
        final Configuration checkConfig = (Configuration) getModuleByIds(" ");
        final String filePath = getPath("InputRightCurlyAloneOrEmptyNoViolations.java");

        final Integer[] warnList = getLinesWithWarn(filePath);
        verify(checkConfig, filePath, expected, warnList);
    }
```
Is writing tests like this in file EmptyBlockTest by using getModuleByIds would be fine?

rnveach: > getModuleByIds(" ");

Not sure how an empty string tests this. But you should be using a module that has multiple ids for, and use an input file that will have violations from both ids. That is how you will show its working.

Binita-tech: As RightCurly have multiple ids, So I should use that code in the input file which has violation from both(shown below), am I saying right?
```
<module name="RightCurly">
    <property name="id" value="RightCurlyAlone"/>
    <property name="option" value="alone"/>
    <property name="tokens"
     value="CLASS_DEF, METHOD_DEF, CTOR_DEF, LITERAL_FOR, LITERAL_WHILE, STATIC_INIT,
            INSTANCE_INIT, ANNOTATION_DEF, ENUM_DEF"/>
</module>
<module name="SuppressionXpathSingleFilter">
  <!-- suppresion is required till https://github.com/checkstyle/checkstyle/issues/7541 -->
  <property name="id" value="RightCurlyAlone"/>
  <property name="query" value="//RCURLY[parent::SLIST[count(./*)=1]
                                         or parent::OBJBLOCK[count(./*)=2]]"/>
</module>
```
And please clear whether changes in file AbstractGoogleModuleTestSupport.java file is done properly? After writing the input file then I have to write tests or write usages to show changes are working appropriately, right?

rnveach: > So I should use that code in the input file which has violation from both(shown below), am I saying right?

RightCurly is different as the second ID is a suppression, which does not print violations it hides them from the other. If we were talking 2 versions of the same check, then what you said would be right.

Also look at main issue's comment:
> we need to load together for each test

So all current RightCurlyCheck's must load both IDs for it's tests. There should be proof that both are loaded, as the suppression will start to hide existing violations being printed.

> please clear whether changes in file AbstractGoogleModuleTestSupport.java file is done properly?

I cannot say because you have no usages of your changes in the test support. Having a usage will tell you if things are working right or not.

Since I mentioned existing tests, if we are not sure which one will have violations suppressed, then create your own input that will prove you are loading both ids and a violation and a suppression are done. If the suppression didn't work, you would see an extra violation. If the violation didn't work, you would see it missing.

I would start with creating your own input. Show me you created an example input that will satisfy the results with 2 items in the config mentioned above. Use https://checkstyle.org/report_issue.html#How_to_report_a_bug.3F as a template and post it to the first post in the PR.

The code does look to be in the right direction outside of my other comments, but without a usage we can't say for sure if they are right or not.

> After writing the input file then I have to write tests or write usages to show changes are working appropriately, right?

Yes.

Binita-tech:
@rnveach , My college exams are going on. I will continue working on this PR after 2 days.

rnveach:
@Binita-tech ping

Binita-tech:
@rnveach I will update changes in this PR  soon. I was having college exam and was working on projects of GSoC.

rnveach:
@Binita-tech do we have some type of timeframe?

Binita-tech:
Closing this due to some time constraint.

Issue #4845, opened 2017-07-28
Test inputs should be completely standalone.


rnveach:
Identified at https://github.com/checkstyle/checkstyle/pull/4844#issuecomment-318740057 ,

Inputs should be completely standalone and not rely on Checkstyle production code or test code.
It creates an usual coupling and requires changes in completely unrelated tests.

Most inputs that have references to checkstyle code: https://github.com/search?utf8=%E2%9C%93&q=path%3Asrc%2Ftest%2Fresources+filename%3AInput*.java+repo%3Acheckstyle%2Fcheckstyle+%22import+com.puppycrawl.tools.checkstyle%22&type=
There could be more as that was just a general search.

Imports have to be examined why they are used, and how best to remove them and still keep the tests original purpose intact.

List of impacted files was added at https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle-input-suppressions.xml#L35 .
Suppressions must be removed as files are worked on to show they are resolved for this issue.

Issue can only be marked completed when all suppressions are identified and handled appropriately.

romani:
I think we can do this by IllegalImport Check(or importorder). We could have special phase and config for this during build.

rnveach:
We should also extend this that we shouldn't import `Inputs` from other inputs unless really necessary.
Re-organizing inputs in #3891 hit an issue with `Input15Extensions`.
Example: https://github.com/search?utf8=%E2%9C%93&q=path%3Asrc%2Ftest%2Fresources+filename%3AInput*.java+repo%3Acheckstyle%2Fcheckstyle+%22import+com.puppycrawl.tools.checkstyle.checks.imports.unusedimports.Input15Extensions%22&type=

pbludov:
This is another blocker for #5102. The  sun packets are widely used in the import order checks.

pbludov:
For the ImportOrderCheck it is possible to add a number of specially arranged classes to test all the order cases. Something like that:
```
src
  test
     resources-importorder
         com
             pkg1
                Class1.java
                Class2.java
         org
             pkg2
                Class3.java
                Class4.java
         net
             pkg3
                Class5.java
                Class6.java
```

romani:
please do not be over crazy with this issue, it could result in huge update.
it is better to do what is required for java9 and that is all, it will be enough for next several years.

rnveach:
This issue isn't even related to java 9, it is about checkstyle's own code being used in inputs causing a needless link between the 2.

[...]

PR #10827, opened 2021-09-26
JavadocMetadataScrapperTest and InlineConfigParser fixes

rnveach:
### reworks JavadocMetadataScrapperTest to make more sense

Moved all tests related to `JavadocMetadataScrapper` into `JavadocMetadataScrapperTest` and use BDD and actually verify outputs from the check. Hopefully, at some point we can use pitest and such on this class to verify it's code.

### changes InlineConfigParser to respect fully qualified name 

This is a simple change to so that when it gets the fully qualified name, it takes as the fully qualified name and doesn't make any other changes. This is all because `JavadocMetadataScrapper` doesn't end with `Check` so BDD can't recognize it. No idea why the class was named this, but BDD should still be changed as I expect the FQN to be the FQN.

Issue #8771, opened 2020-08-28
Improve code coverage for metadata generator classes


gaurabdg:
Classes present in https://github.com/checkstyle/checkstyle/tree/master/src/main/java/com/puppycrawl/tools/checkstyle/meta

PR #11402, opened 2022-03-12
Issue #4845: removed usage of TokenTypes from inputs

rnveach:
Issue #4845

Targetted TokenTypes and switch them to Calendar which also has public static integer fields.

### DIFF
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/meta/javadocmetadatascraper/InputJavadocMetadataScraperAbstractSuperCheck.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/meta/javadocmetadatascraper/InputJavadocMetadataScraperAbstractSuperCheck.java
@@ -12,9 +12,9 @@ import java.util.LinkedList;
 
 import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
-import com.puppycrawl.tools.checkstyle.api.DetailAST;
-import com.puppycrawl.tools.checkstyle.api.TokenTypes;
-import com.puppycrawl.tools.checkstyle.utils.ScopeUtil;
+
+
+
 
 /** //violation
  * <p>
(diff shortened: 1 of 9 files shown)


src/test/resources/com/puppycrawl/tools/checkstyle/meta/javadocmetadatascraper/InputJavadocMetadataScraperAbstractSuperCheck.java
```diff
@@ -12,9 +12,9 @@
 
 import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
-import com.puppycrawl.tools.checkstyle.api.DetailAST;
-import com.puppycrawl.tools.checkstyle.api.TokenTypes;
```
rnveach, line 16: Though this is the meta scapper, I wasn't seeing any uses of these imports so I removed them because of their connection to this issue.
Let me know if there is a reason for these.

romani (approved):
Ok to merge if CI pass

src/test/resources/com/puppycrawl/tools/checkstyle/meta/javadocmetadatascraper/InputJavadocMetadataScraperAbstractSuperCheck.java
romani: Please recheck git history.

rnveach: @romani When I rechecked one file `InputJavadocMetadataScraperNoCodeInFileCheck`, the file had been like that since the beginning.
> Pull #10827: reworks JavadocMetadataScrapperTest to make more sense

This is more me moving files around.

> Issue https://github.com/checkstyle/checkstyle/issues/8771: Improve code coverage for metadata generator classes

This is where the file really started and it is the same. I didn't see anything in PRs, and this issue was for code coverage. I can only assume we wanted more of a representation of the Check without copying the check's full source.

romani: Ok, let's merge this PR

PR #11404, opened 2022-03-12
Issue #4845: removed usage of Checker and similar from inputs

rnveach:
Issue https://github.com/checkstyle/checkstyle/issues/4845

Targeted Checker and ones next to them. The import inputs seem like copy/pastes.


romani (approved):
Ok to merge If CI is green

PR #11411, opened 2022-03-13
minor: compare tokens by type in ScopeUtil

pbludov:
For some reason, the modifiers in `ScopeUtil` are compared by text.
Maybe it comes from very ancient code.

Diff Regression config: https://gist.githubusercontent.com/pbludov/3a8be5da26a64e2be80b26217d052eb6/raw/343dd6f3ef6e8c8d3ae5c06821cc4a4f59fb4101/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
@@ -44,16 +44,20 @@ public final class ScopeUtil {
      */
     public static Scope getDeclaredScopeFromMods(DetailAST aMods) {
         Scope result = null;
-        for (DetailAST token = aMods.getFirstChild(); token != null;
+        for (DetailAST token = aMods.getFirstChild(); token != null && result == null;
                 token = token.getNextSibling()) {
-            if ("public".equals(token.getText())) {
-                result = Scope.PUBLIC;
-            }
-            else if ("protected".equals(token.getText())) {
-                result = Scope.PROTECTED;
-            }
-            else if ("private".equals(token.getText())) {
-                result = Scope.PRIVATE;
+            switch (token.getType()) {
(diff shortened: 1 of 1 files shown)


pbludov:
Github, generate report

romani (approved):

rnveach:
I think regression is missing some impacted checks. The util `getDeclaredScopeFromMods` has impacted checks, but it also impacts other utils and that expands the search.

`getDeclaredScopeFromMods` also impacts `getScope`. That impacts `getSurroundingScope` which includes the missing checks `RequireThis` and `DesignForExtension`. That also impacts `isInScope` which includes the missing check `HiddenField`.

Those 3 checks are not in the regression.

src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
```diff
@@ -44,16 +44,20 @@ private ScopeUtil() {
      */
     public static Scope getDeclaredScopeFromMods(DetailAST aMods) {
         Scope result = null;
-        for (DetailAST token = aMods.getFirstChild(); token != null;
+        for (DetailAST token = aMods.getFirstChild(); token != null && result == null;
```
rnveach, line 47: I am more just curious abut this change. Is it just to exit the loop faster? I am surprised pitest doesn't flag this in some way.

rnveach: Utils doesn't have `REMOVE_CONDITIONALS` active. When it does, this will most likely come up again.

pbludov: Yes this is for fast exit from the loop in cases like `private static final`. No reason to go through the other modifiers if we already met `public` or `private`.

pbludov:
Github, generate report

rnveach (approved):

Issue #11214, opened 2022-01-18
Specify violation messages in input files


Vyom-Yadav:
Single Check might have multiple violations messages, we need to see exact violation message if Check has multiple messages, to be sure that violation of correct type is placed at right line.

Violation messages have not been specified in certain input files, a list of checks in which violation message is not specified in the input file and have more than one violation message key has been specified here-
https://github.com/checkstyle/checkstyle/blob/72f182241a3b8a6d905328dfcfc9739dfffac9bd/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L99

To Do-

- Pick a check (single Check Inputs files update in single Pull Request) To allocate Check update for you please make comment in issue "I am on XxxxxCheck"
- Find all input files that use that check
- Add violation message at the end of the `// violation` comment
- Remove the check from the suppression list

Same has been demonstrated at 461cd048bb059ea6948945104682fb99d4e155dc
If the violation message contains special regex characters, escape that character.
Ignore input files using more than a single check.

Ignore lines with multiple violations, eg-
```
// 2 violations
// 2 violations above
// 4 violations below
```
and any other violation message of this nature.

List of Checks: 

- [x]   RegexpCheck
- [x]   EmptyForInitializerPadCheck
- [ ]   JavadocStyleCheck
- [x]   SummaryJavadocCheck

Old tracker (some might be outdated):

- [x] RightCurlyCheck
- [x] EqualsAvoidNullCheck
- [x] MethodParamPadCheck
- [x] DescendantTokenCheck
- [x] RequireThisCheck
- [x] EqualsHashCodeCheck
- [x] JavaNCSSCheck
- [x] ReturnCountCheck
- [x] AnnotationUseStyleCheck
- [x] EmptyLineSeparatorCheck
- [x] OrderedPropertiesCheck
- [x] AbstractJavadocCheck
- [x] DeclarationOrderCheck
- [x] MissingOverrideCheck
- [x] AbstractClassNameCheck
- [x] MethodCountCheck
- [x] PackageDeclarationCheck
- [x] JavadocTypeCheck
- [x] LeftCurlyCheck
- [ ] RegexpCheck
- [x] WhitespaceAfterCheck
- [x] MultipleVariableDeclarationsCheck
- [x] EmptyForInitializerPadCheck
- [x] UnnecessaryParenthesesCheck
- [x] JavadocMethodCheck
- [x] ModifierOrderCheck
[...]

PR #11414, opened 2022-03-14
Issue #11214: Remove MissingOverrideCheck from InlineConfigParser

AmrDeveloper:
Issue Number: #11214

In MissingOverrideCheck all violation messages are specified correctly and testing passed correctly so we should be remove from the InlineConfigParser list

https://github.com/checkstyle/checkstyle/tree/master/src/test/resources/com/puppycrawl/tools/checkstyle/checks/annotation/missingoverride


Vyom-Yadav:
The commit message although logically correct (removing from a particular file) is a bit misleading, can you change it to something like - 
Remove XXXCheck from unspecified violation message suppression or anything which is a bit more descriptive.

AmrDeveloper:
> The commit message although logically correct (removing from a particular file) is a bit misleading, can you change it to something like - Remove XXXCheck from unspecified violation message suppression or anything which is a bit more descriptive.

You are right, I have changed it now, Thanks

romani (approved):

AmrDeveloper:
@romani Please remove MissingOverrideCheck from the list of cheks on the issue

PR #11428, opened 2022-03-18
infra: clear cache

nick-mancuso:
Just trying to clear cache, builds are failing with the same error again: https://checkstyle.semaphoreci.com/jobs/dde21618-4a6b-4ee3-9b37-993ba00d14ca
`A required class was missing while executing org.antlr:antlr4-maven-plugin:4.9.3:antlr4: org/stringtemplate/v4/STGroup`

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -13,7 +13,7 @@ blocks:
       prologue:
         commands:
           - checkout
-          - cache restore m2
+          - cache delete m2
           - eval "export M2_CACHE_SIZE=$(du -s $HOME/.m2 | cut -f1)"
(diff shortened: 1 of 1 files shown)


strkkk:
why it is a cache issue?

romani:
Try to cat that file, zipinfo on it or ... `file (path)`

nick-mancuso:
semaphore seems to be ok after clearing cache, I am closing PR.

> why it is a cache issue?

Just a guess, but same weird failure across most PRs  in the same CI/job seems like a bad download that was cached to me; if it was an actual problem with dependency/ download, it would be failing in all CI.

PR #11231, opened 2022-01-24
Issue #11087: New check ChainedMethodCallWrap

Vyom-Yadav:
Resolves #11087: New check chained method call wrap

---

Diff Regression config: https://gist.githubusercontent.com/Vyom-Yadav/98dceb63a79f4833e85fff9b2e1464a6/raw/e5d475dd0a62d241e4aa20e58f607672133859fb/my_checks.xml
Diff Regression patch config: https://gist.githubusercontent.com/Vyom-Yadav/273327b9c6cd6963b7dade792bc2c9c2/raw/722734822d2766fded891ddf25819194ba6cdedf/patch_config.xml
Report label: optionalMethodCallsAllProjects

---

Default config of check:
```xml
<module name="ChainedMethodCallWrap">
 <property name="identifierPattern" value="^$"/>
 <property name="maxCallsInSingleLine" value="1" />
 <property name="maxCallsPerLine" value="1" />
</module>
```
---

Config used by checkstyle internally:
```xml
<module name="ChainedMethodCallWrap">
  <property name="identifierPattern" value="^assert.*"/>
  <property name="maxCallsInSingleLine" value="2" />
</module>
```
---

Report for the following config (Match all method calls, report is huge, was used to check for exceptions):

- [x] Updated

```xml
<module name="ChainedMethodCallWrap">
  <property name="identifierPattern" value="^.*$" />
</module>
```
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/3d90a1c_2023113432/reports/diff/index.html

---

Report for the config used by checkstyle:

- [x] Updated

```xml
<module name="ChainedMethodCallWrap">
  <property name="identifierPattern" value="^assert.*"/>
  <property name="maxCallsInSingleLine" value="2" />
</module>
```
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/793d000_2023080701/reports/diff/index.html

---
Report for FullIdent:

- [x] Updated

1. Checkstyle Checks:

    - (**VariableDeclarationUsageDistance shows diff**:
     https://github.com/checkstyle/checkstyle/issues/11335) 
     https://vyom-yadav.github.io/DiffReport/diff-full-ident-main/index.html
2. Sevntu Checks:  https://vyom-yadav.github.io/DiffReport/diff-full-ident-sevntu/index.html

---

- [x] Updated

[...]

PR #1422, opened 2015-07-20
CustomImportOrderCheck. Fix #1263

ivanov-alex:
Reports are pending...

Difference between current solution and solution after the fix is a starting point for reviewing the list of imports is a starting point.
Both solutions show violations only when actual sorting problem exists. So no "false positive" happens in both cases. However, both solutions will return different amount of violations due to different starting point.

**Current solution** takes the very first import as a starting point. It defines "current" group based on this import and every time it see import from the group that should go before "current" group, violation is submitted. Most negative scenario: first import doesn't match any group, it is assigned to default "no group" section. As a result, only imports that cannot be assigned to defined groups on config will be treated as correctly placed and all the imports that match groups defined will receive violations.
For example
Config:
```xml
        <module name="CustomImportOrder">
            <property name="customImportOrderRules" value="SPECIAL_IMPORTS###THIRD_PARTY_PACKAGE###STANDARD_JAVA_PACKAGE"/>
            <property name="specialImportsRegExp" value="^com.google"/>
            <property name="thirdPartyPackageRegExp" value="^org"/>
            <property name="sortImportsInGroupAlphabetically" value="true"/>
            <property name="separateLineBetweenGroups" value="true"/>
        </module>
```
Source:
```java
1 import com.puppycrawl.tools.checkstyle.CustomImportOrderCheck; //doesn't match any group on config
2 
3 import com.google.common.annotations.Beta;
4 import com.google.common.annotations.GwtCompatible;
5
6 import org.androidannotations.helper.ModelConstants.generationSuffix;
7 
8 import java.io.Serializable;
9 import java.util.Iterator;
10 import java.util.Set;
11
12 import javax.annotation.Nullable;
13
14 import com.puppycrawl.tools.checkstyle.ImportOrderCheck; //doesn't match any group on config
15 import net.sourceforge.pmd.lang.ast.Node; //doesn't match any group on config
```
Result (all imports except of 1,14,15 have violations)
```
Starting audit...
test.java:3: warning: Import statement is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group.
test.java:4: warning: Import statement is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group.
test.java:6: warning: Import statement is in the wrong order. Should be in the 'THIRD_PARTY_PACKAGE' group.
test.java:8: warning: Import statement is in the wrong order. Should be in the 'STANDARD_JAVA_PACKAGE' group.
[...]

PR #11431, opened 2022-03-18
minor: Wrapped chained method calls

Vyom-Yadav:
Long method chains are hard to read, wrapped them onto a new line. Prerequisite of https://github.com/checkstyle/checkstyle/pull/11231
Suggested in - https://github.com/checkstyle/checkstyle/pull/11231#discussion_r829070035

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
@@ -727,7 +727,8 @@ public class XdocsPagesTest {
 
         assertWithMessage(
                 fileName + " section '" + sectionName + "' should show properties: " + properties)
-                        .that(properties).isEmpty();
+            .that(properties)
+            .isEmpty();
     }
 
     private static void fixCapturedProperties(String sectionName, Object instance, Class<?> clss,
@@ -1418,7 +1419,8 @@ public class XdocsPagesTest {
 
         assertWithMessage(fileName + " section '" + sectionName
                 + "' has unknown text in 'Example of Usage': " + text)
-            .that(text).isEmpty();
+            .that(text)
+            .isEmpty();
 
         boolean hasCheckstyle = false;
         boolean hasGoogle = false;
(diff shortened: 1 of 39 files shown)


nick-mancuso (approved):

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -727,7 +727,8 @@ private static void validatePropertySection(String fileName, String sectionName,
 
         assertWithMessage(
                 fileName + " section '" + sectionName + "' should show properties: " + properties)
-                        .that(properties).isEmpty();
+                        .that(properties)
```
pbludov, line 730: I think that the wrapping on the line #1422 is more readable.
Such indent looks like it is bound to the string message.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1595,7 +1597,8 @@ public void testAllStyleRules() throws Exception {
 
             assertWithMessage(
                     fileName + " requires the following check(s) to appear: " + styleChecks)
-                            .that(styleChecks).isEmpty();
+                            .that(styleChecks)
+                            .isEmpty();
```
pbludov, line 1601: same here

pbludov (approved):

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Vyom-Yadav: Done.

romani (approved):
This idea of placing each call on separate line might be ok in general for test code. But in main code there will problems, as it extends method size significantly.
Forcing people to refactor method split is not good.

Vyom-Yadav:
> This idea of placing each call on separate line might be ok in general for test code. But in main code there will problems, as it extends method size significantly.
Forcing people to refactor method split is not good.

It can be configured to suit anyone's needs.

PR #11441, opened 2022-03-19
config: remove files from exclude of forbiddenapi; put comment on exc…

romani:
Related to

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1546,8 +1546,7 @@
 
                 <exclude>**/RegexpCheckTest.class</exclude>
 
-                <exclude>**/SuppressionSingleFilterTest.class</exclude>
-
+                <!-- test need to be updated -->
                 <exclude>**/AllSinglelineCommentsTest.class</exclude>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1543,67 +1543,80 @@
...
                 <exclude>**/AllSinglelineCommentsTest.class</exclude>
 
+                <!-- usage of custom internal module -->
                 <exclude>**/FileSetCheckTest.class</exclude>
-
                 <exclude>**/AbstractFileSetCheckTest.class</exclude>
 
-                <exclude>**/OrderedPropertiesCheckTest.class</exclude>
-
+                <!-- test needs to be updated -->
```
nick-mancuso, line 1559: Can we do standard `until https://github.com/checkstyle/checkstyle/issues/11446` now that we have open issue?

nick-mancuso (changes requested):

pom.xml
nick-mancuso: Resolved.

PR #11445, opened 2022-03-20
minor: fix mnemonic for open file button

pbludov:
It is strange that the hotkey for Open file dialog is `Alt+S`. It looks like a mistype.
Should be `Alt+O`.

Maybe it was "Select file" originally, but later the label was changed.


rnveach (approved):
We should follow other standard keys that other applications also follow. Not sure if there is a list somewhere.

romani (approved):

Issue #11446, opened 2022-03-20
Update Tests to use new verify method that use inlined config in Input files


romani:
Test need to be updated:
AllSinglelineCommentsTest
SuppressionCommentFilterTest
EmptyLineSeparatorCheckTest
SuppressWarningsHolderTest
ImportControlCheckTest
ImportOrderCheckTest
SuppressWithNearbyCommentFilterTest
CommentsIndentationCheckTest
SuppressionFilterTest
WriteTagCheckTest
AbstractJavadocCheckTest

nick-mancuso:
I am on it, updating for AllSinglelineCommentsTest

ghulamrabbani883:
Can I work on this issue

romani:
>EmptyLineSeparatorCheckTest

Will be covered at https://github.com/checkstyle/checkstyle/pull/11408
@Kevin222004, fyi.

Kevin222004:
@Rahulkhinchi03  Please give a look to this 

> > EmptyLineSeparatorCheckTest
> 
> Will be covered at #11408 @Kevin222004, fyi.

It will be covering at #11408 it has been already mentioned by @romani , in conversation

Rahulkhinchi03:
@Kevin222004  It was my mistake to open a PR on that. I am closing the PR. Thank you!

Kevin222004:
@romani  ping,

  please let me know if I am wrong

romani:
Yes it should be updated to use execute method

Vyom-Yadav:
@romani Why does `execute` use the traditional way of building configurations and not the parser? If the reason behind it is that it traverses all the lines in the file to `setViolations` then a execute method could be made that just set the configuration using the input file and parser and does not set violations.

nick-mancuso:
> Why does execute use the traditional way of building configurations and not the …

The “check” that I created it for was not a real check, we just needed to execute a fake check with no properties on code to count lines. We should not use this method for real checks, IMO.

https://github.com/checkstyle/checkstyle/pull/11447

romani:
loading of config from Input file very helpful when we do violations in such Input file, so we review config and violations in same place.
In case of test that expect exceptions in initialization or some test that not about functional behavior of test but about following of 100% coverage of tests - we do not need new approach.

Vyom-Yadav:
[...]

PR #11447, opened 2022-03-21
Issue #11446: Update AllSinglelineCommentsTest to use verifyWithInlineConfigParser

nick-mancuso:
Resolves update for AllSinglelineCommentsTest in #11446.

This PR will serve as example for associated issue

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/grammar/comments/AllSinglelineCommentsTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/grammar/comments/AllSinglelineCommentsTest.java
@@ -33,7 +33,6 @@ import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
 import com.puppycrawl.tools.checkstyle.internal.utils.CheckUtil;
-import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 
 public class AllSinglelineCommentsTest extends AbstractModuleTestSupport {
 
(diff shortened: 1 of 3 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/grammar/comments/AllSinglelineCommentsTest.java
```diff
@@ -83,8 +81,8 @@ public boolean isCommentNodesRequired() {
 
         @Override
         public void init() {
-            final int lines = 63;
-            for (int i = 0; i < lines; i++) {
+            final int lines = 70;
+            for (int i = 7; i < lines; i++) {
```
nick-mancuso, line 85: Line number adjustment to allow for config comment

src/test/resources/com/puppycrawl/tools/checkstyle/grammar/comments/InputFullOfSinglelineComments.java
```diff
@@ -1,64 +1,70 @@
...
-{//43
-String//44
-line//45
-=//46
-i//47
-+//48
-""//49
-;//50
+/*
+com.puppycrawl.tools.checkstyle.grammar.comments.AllSinglelineCommentsTest$SinglelineCommentListenerCheck
```
nick-mancuso, line 2: Note the `$`, this is how we can reference a static inner class for usage in `Class#forName` in `InlineConfigParser`.

src/test/java/com/puppycrawl/tools/checkstyle/grammar/comments/AllSinglelineCommentsTest.java
```diff
@@ -48,12 +47,10 @@ protected String getPackageLocation() {
 
     @Test
     public void testAllSinglelineComments() throws Exception {
-        final DefaultConfiguration checkConfig =
-            createModuleConfig(SinglelineCommentListenerCheck.class);
         final String path = getPath("InputFullOfSinglelineComments.java");
         lineSeparator = CheckUtil.getLineSeparatorForFile(path, StandardCharsets.UTF_8);
         final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
-        verify(checkConfig, path, expected);
+        verifyWithInlineConfigParser(path, expected);
```
romani, line 53: This becomes weird. We try to verify.... But in reality it is not verication, it  just execution of module. I think we need new method `execute` that just run module and do not validation.
Update in input file become not required.

Same concept we will use in case we put exceptional values in config of module.
So test expects exception.

romani (changes requested):
Item:

PR #11459, opened 2022-03-24
minor: change java version for modernizer plugin

strkkk:
Update hardcoded version to property


romani:
Better commit prefix - config

Issue #11374, opened 2022-03-07
False Positive: UnusedLocalVariableCheck


Vyom-Yadav:
I have read check documentation: https://checkstyle.sourceforge.io/config_coding.html#UnusedLocalVariable
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Test.java

/var/tmp $ cat config.xml
```
```xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="TreeWalker">
        <module name="UnusedLocalVariable"/>
    </module>
</module>
```
```
/var/tmp $ cat Test.java
```
```java
public class Test {

    static class m {
        static class h {
            public int a = 12;
        }
    }

    class j {
        void method() {
            int a = 22;
            m.h obj = new m.h() {
                int j = a + 12;
            };
            obj.getClass();
        }

        static class m {
            static class h {
            }
        }
    }
}
```
```
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:11:13: Unused local variable 'a'. [UnusedLocalVariable]
Audit done.
Checkstyle ends with 1 errors.

```
---

**Describe what you expect in detail.**
```
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.0-all.jar -c config.xml Test.java
Starting audit...
Audit done.
```

---

Discovered while working on #11365, the source has been identified, the fix has been done locally, will be pushed soon.

Vyom-Yadav:
Approved? This may be a non-common case but the check has support for these types of cases.

nick-mancuso:
I agree that this is a false positive, please add proof of unused variable to issue description (i.e. print value of `j`) to make it obvious to maintainers.

Vyom-Yadav:
Done.

Vishakhakumawat:
I am on it.

[...]

PR #11468, opened 2022-03-26
Fixed false positives related to anonymous inner classes in UnusedLocalVariable and FinalClass

Vyom-Yadav:
Resolves https://github.com/checkstyle/checkstyle/issues/11365 Resolves https://github.com/checkstyle/checkstyle/issues/11374

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
@@ -110,6 +110,12 @@ public class FinalClassCheck
     /** Keeps ClassDesc objects for all inner classes. */
     private Map<String, ClassDesc> innerClasses;
 
+    /**
+     * Maps anonymous inner class's {@link TokenTypes#LITERAL_NEW} node to
+     * the outer class's {@link ClassDesc} object.
+     */
+    private Map<DetailAST, ClassDesc> anonInnerClassToOuterClass;
+
     /** Keeps ClassDesc objects for stack of declared classes. */
     private Deque<ClassDesc> classes;
 
@@ -144,6 +150,7 @@ public class FinalClassCheck
     public void beginTree(DetailAST rootAST) {
         classes = new ArrayDeque<>();
         innerClasses = new HashMap<>();
+        anonInnerClassToOuterClass = new HashMap<>();
         packageName = "";
     }
 
@@ -170,12 +177,8 @@ public class FinalClassCheck
 
             case TokenTypes.LITERAL_NEW:
                 if (ast.getFirstChild() != null
-                        && ast.getLastChild().getType() == TokenTypes.OBJBLOCK) {
-                    for (ClassDesc classDesc : classes) {
-                        if (doesNameOfClassMatchAnonymousInnerClassName(ast, classDesc)) {
-                            classDesc.registerAnonymousInnerClass();
-                        }
-                    }
+                    && ast.getLastChild().getType() == TokenTypes.OBJBLOCK) {
+                    anonInnerClassToOuterClass.put(ast, classes.peek());
                 }
                 break;
 
@@ -220,6 +223,7 @@ public class FinalClassCheck
             classes.pop();
         }
         if (TokenUtil.isRootNode(ast.getParent())) {
+            anonInnerClassToOuterClass.forEach(this::registerAnonymousInnerClassToSuperClass);
             // First pass: mark all classes that have derived inner classes
             innerClasses.forEach(this::registerNestedSubclassToOuterSuperClasses);
             // Second pass: report violation for all classes that should be declared as final
@@ -241,7 +245,7 @@ public class FinalClassCheck
     private static boolean shouldBeDeclaredAsFinal(ClassDesc desc) {
         return desc.isWithPrivateCtor()
                 && !(desc.isDeclaredAsAbstract()
-                    || desc.isWithAnonymousInnerClass())
+                    || desc.isSuperClassOfAnonymousInnerClass())
                 && !desc.isDeclaredAsFinal()
                 && !desc.isWithNonPrivateCtor()
                 && !desc.isWithNestedSubclass();
@@ -259,7 +263,7 @@ public class FinalClassCheck
         final String superClassName = getSuperClassName(currentClass.getClassAst());
         if (superClassName != null) {
             final ClassDesc nearest =
-                    getNearestClassWithSameName(superClassName, qualifiedClassName);
+                    getNearestClassWithSameName(superClassName, qualifiedClassName, false);
             if (nearest == null) {
                 Optional.ofNullable(innerClasses.get(superClassName))
                         .ifPresent(ClassDesc::registerNestedSubclass);
@@ -271,20 +275,74 @@ public class FinalClassCheck
     }
 
     /**
+     * Register to the super class of anonymous inner class that the given class is instantiated
+     * by an anonymous inner class.
+     *
+     * @param literalNewAst ast node of {@link TokenTypes#LITERAL_NEW} representing anonymous inner
+     *                      class
+     * @param outerClass the class under which the anonymous inner class is nested.
+     */
+    private void registerAnonymousInnerClassToSuperClass(DetailAST literalNewAst,
+                                                         ClassDesc outerClass) {
+        final String superClassName = CheckUtil.getShortNameOfAnonInnerClass(literalNewAst);
+        final ClassDesc superClassDesc = getNearestClassWithSameName(
+            superClassName, outerClass.getQualifiedName(), true);
+        if (superClassDesc == null) {
+            Optional.ofNullable(innerClasses.get(superClassName))
+                .ifPresent(ClassDesc::registerSuperClassOfAnonymousInnerClass);
+        }
+        else {
+            superClassDesc.registerSuperClassOfAnonymousInnerClass();
+        }
+    }
+
+    /**
      * Checks if there is a class with same name.
+     * <p>
+     * The parameter {@code isSearchedClsSuperClsOfAnonInnerClass} exists because if the class being
+     * searched for is the super class of anonymous inner class, the rules of evaluation are a bit
+     * different, consider the following example-
+     * <pre>
+     * {@code
+     * public class Main {
+     *     static class a {
+     *         static class b {
+     *         }
+     *     }
+     *
+     *     class j {
+     *         a.b object = new a.b() { // Object of Main.j.a.b and not of Main.a.b
+     *         };
+     *
+     *         static class a {
+     *             static class b {
+     *             }
+     *         }
+     *     }
+     * }
+     * }
+     * </pre>
+     * If the parameter wasn't set to {@code true} then the logic would have falsely evaluated
+     * {@code Main.a.b} to be the super class of the anonymous inner class. Set this to
+     * {@code false} only if the class being searched for is not the super class of anonymous
+     * inner class.
      *
      * @param className name of the class
-     * @param superClassName name of the super class
+     * @param superOrOuterClassName name of the super class or the outer class
+     * @param isSearchedClsSuperClsOfAnonInnerClass Is the class being searched super class of an
+     *                                              anonymous inner class
      * @return true if there is another class with same name.
      * @noinspection CallToStringConcatCanBeReplacedByOperator
      */
-    private ClassDesc getNearestClassWithSameName(String className, String superClassName) {
+    private ClassDesc getNearestClassWithSameName(String className, String superOrOuterClassName,
+                                                  boolean isSearchedClsSuperClsOfAnonInnerClass) {
         final String dotAndClassName = PACKAGE_SEPARATOR.concat(className);
         return innerClasses.entrySet().stream()
                 .filter(entry -> entry.getKey().endsWith(dotAndClassName))
                 .map(Map.Entry::getValue)
                 .min((first, second) -> {
-                    return getClassDeclarationNameMatchingCountDiff(superClassName, first, second);
+                    return getClassDeclarationNameMatchingCountDiff(superOrOuterClassName, first,
+                         second, isSearchedClsSuperClsOfAnonInnerClass);
                 })
                 .orElse(null);
     }
@@ -293,37 +351,27 @@ public class FinalClassCheck
      * Get the difference between class declaration name matching count. If the difference between
      * them is zero, then their depth is compared to obtain the result.
      *
-     * @param superClassName name of the super class
+     * @param superOrOuterClassName name of the super class or the outer class
      * @param firstClass first input class
      * @param secondClass second input class
+     * @param isSearchedClsSuperClsOfAnonInnerClass Is the class being searched super class of an
+     *                                              anonymous inner class
      * @return difference between class declaration name matching count
      */
-    private static int getClassDeclarationNameMatchingCountDiff(String superClassName,
-                                                                ClassDesc firstClass,
-                                                                ClassDesc secondClass) {
+    private static int getClassDeclarationNameMatchingCountDiff(String superOrOuterClassName,
+                                                 ClassDesc firstClass, ClassDesc secondClass,
+                                                 boolean isSearchedClsSuperClsOfAnonInnerClass) {
         int diff = Integer.compare(
-            CheckUtil
-                .typeDeclarationNameMatchingCount(superClassName, secondClass.getQualifiedName()),
-            CheckUtil
-                .typeDeclarationNameMatchingCount(superClassName, firstClass.getQualifiedName()));
+            CheckUtil.typeDeclarationNameMatchingCount(superOrOuterClassName,
+                                                       secondClass.getQualifiedName(),
+                                                       isSearchedClsSuperClsOfAnonInnerClass),
+            CheckUtil.typeDeclarationNameMatchingCount(superOrOuterClassName,
+                                                       firstClass.getQualifiedName(),
+                                                       isSearchedClsSuperClsOfAnonInnerClass));
         if (diff == 0) {
             diff = Integer.compare(firstClass.getDepth(), secondClass.getDepth());
         }
         return diff;
-    }
-
-    /**
-     * Check if class name matches with anonymous inner class name.
-     *
-     * @param ast current ast.
-     * @param classDesc class to match.
-     * @return true if current class name matches anonymous inner
-     *         class name.
-     */
-    private static boolean doesNameOfClassMatchAnonymousInnerClassName(DetailAST ast,
-                                                               ClassDesc classDesc) {
-        final String[] className = classDesc.getQualifiedName().split("\\.");
-        return ast.getFirstChild().getText().equals(className[className.length - 1]);
     }
 
     /**
@@ -394,8 +442,8 @@ public class FinalClassCheck
         /** Does class have nested subclass. */
         private boolean withNestedSubclass;
 
-        /** Does class have anonymous inner class. */
-        private boolean withAnonymousInnerClass;
+        /** Is the class super class of an anonymous inner class. */
+        private boolean superClassOfAnonymousInnerClass;
 
         /**
          *  Create a new ClassDesc instance.
@@ -447,8 +495,8 @@ public class FinalClassCheck
         }
 
         /** Adds anonymous inner class. */
-        private void registerAnonymousInnerClass() {
-            withAnonymousInnerClass = true;
+        private void registerSuperClassOfAnonymousInnerClass() {
+            superClassOfAnonymousInnerClass = true;
         }
 
         /**
@@ -506,12 +554,12 @@ public class FinalClassCheck
         }
 
         /**
-         * Does class have an anonymous inner class.
+         * Is the class super class of an anonymous inner class.
          *
-         * @return true if class has anonymous inner class
+         * @return {@code true} if the class is the super class of an anonymous inner class.
          */
-        private boolean isWithAnonymousInnerClass() {
-            return withAnonymousInnerClass;
+        private boolean isSuperClassOfAnonymousInnerClass() {
+            return superClassOfAnonymousInnerClass;
         }
 
     }
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
@@ -605,16 +605,35 @@ public final class CheckUtil {
      * would be equal to 6 and not 7 (b of ball is not counted).
      * </p>
      *
+     * <p>
+     * Only set {@code isSearchedClassSuperClassOfAnonymousInnerClass} to {@code true} if the class
+     * being searched for is the super class of an anonymous inner class. Suppose our pattern class
+     * is {@code Main.j} and class to be matched is {@code Main.j.a.b}, the matching count is this
+     * case would be 6 instead of being 5. Even if we didn't hit package separator which was our
+     * main criteria for updating count in previous case, the count is updated as if searched class
+     * is super class of anonymous inner class then nested class can also be the target class which
+     * isn't true in case of extending classes as you can't extend a nested class.
+     * </p>
+     *
      * @param patternClass class against which the given class has to be matched
      * @param classToBeMatched class to be matched
+     * @param isSearchedClassSuperClassOfAnonymousInnerClass Is the class being searched super
+     *                                                       class of an anonymous inner class and
+     *                                                       in that case, the pattern class is the
+     *                                                       outer class of the anonymous
+     *                                                       inner class
      * @return class name matching count
      */
     public static int typeDeclarationNameMatchingCount(String patternClass,
-                                                       String classToBeMatched) {
+            String classToBeMatched, boolean isSearchedClassSuperClassOfAnonymousInnerClass) {
         final int length = Math.min(classToBeMatched.length(), patternClass.length());
         int result = 0;
         for (int i = 0; i < length && patternClass.charAt(i) == classToBeMatched.charAt(i); ++i) {
-            if (patternClass.charAt(i) == PACKAGE_SEPARATOR) {
+            if (patternClass.charAt(i) == PACKAGE_SEPARATOR
+                || isSearchedClassSuperClassOfAnonymousInnerClass
+                && i == length - 1
+                && classToBeMatched.length() > i + 1
+                && classToBeMatched.charAt(i + 1) == PACKAGE_SEPARATOR) {
                 result = i;
             }
         }
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
@@ -647,9 +647,9 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
                                                                TypeDeclDesc secondTypeDecl) {
         int diff = Integer.compare(
             CheckUtil.typeDeclarationNameMatchingCount(
-                outerTypeDeclName, secondTypeDecl.getQualifiedName()),
+                outerTypeDeclName, secondTypeDecl.getQualifiedName(), true),
             CheckUtil.typeDeclarationNameMatchingCount(
-                outerTypeDeclName, firstTypeDecl.getQualifiedName()));
+                outerTypeDeclName, firstTypeDecl.getQualifiedName(), true));
         if (diff == 0) {
             diff = Integer.compare(firstTypeDecl.getDepth(), secondTypeDecl.getDepth());
         }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheckTest.java
@@ -167,6 +167,16 @@ public class UnusedLocalVariableCheckTest extends AbstractModuleTestSupport {
     }
 
     @Test
+    public void testUnusedLocalVarInAnonInnerClasses2() throws Exception {
+        final String[] expected = {
+            "20:13: " + getCheckMessage(MSG_UNUSED_LOCAL_VARIABLE, "s"),
+        };
+        verifyWithInlineConfigParser(
+            getPath("InputUnusedLocalVariableAnonInnerClasses2.java"),
+            expected);
+    }
+
+    @Test
     public void testUnusedLocalVarGenericAnonInnerClasses() throws Exception {
         final String[] expected = {
             "13:9: " + getCheckMessage(MSG_UNUSED_LOCAL_VARIABLE, "l"),
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheckTest.java
@@ -188,4 +188,15 @@ public class FinalClassCheckTest
         verifyWithInlineConfigParser(getPath("InputFinalClassInterface.java"), expected);
     }
 
+    @Test
+    public void testFinalClassAnonymousInnerClass() throws Exception {
+        final String[] expected = {
+            "11:9: " + getCheckMessage(MSG_KEY, "b"),
+            "27:9: " + getCheckMessage(MSG_KEY, "m"),
+            "40:9: " + getCheckMessage(MSG_KEY, "q"),
+            "52:13: " + getCheckMessage(MSG_KEY, "b"),
+        };
+        verifyWithInlineConfigParser(getPath("InputFinalClassAnonymousInnerClass.java"), expected);
+    }
+
 }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unusedlocalvariable/InputUnusedLocalVariableAnonInnerClasses2.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unusedlocalvariable/InputUnusedLocalVariableAnonInnerClasses2.java
@@ -1,0 +1,56 @@
+/*
+UnusedLocalVariable
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.unusedlocalvariable;
+
+public class InputUnusedLocalVariableAnonInnerClasses2 {
+
+    static class m {
+        static class h {
+            public int a = 12;
+        }
+    }
+
+    static class j {
+        static void method() {
+            int a = 1000; // ok
+            int s = 13; // violation
+            int q = 14; // ok
+            m.h obj = new m.h() {
+                @Override
+                void method() {
+                    Integer.valueOf(a + s);
+                }
+
+                m.h obj = new m.h() { // ok, inst var
+                    @Override
+                    void method() {
+                        Integer.valueOf(q);
+                    }
+                };
+
+            };
+            obj.method();
+        }
+
+        static class m {
+            static class h {
+                int s = 12;
+                void method() {
+                }
+            }
+        }
+    }
+
+    static class jasper {
+        static class m {
+            static class h {
+                int q = 12;
+            }
+        }
+    }
+
+}
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/design/finalclass/InputFinalClassAnonymousInnerClass.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/design/finalclass/InputFinalClassAnonymousInnerClass.java
@@ -1,0 +1,67 @@
+/*
+FinalClass
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.design.finalclass;
+
+public class InputFinalClassAnonymousInnerClass {
+    static class a {
+        static class b { // violation
+            private b() {
+            }
+        }
+    }
+
+    static class j {
+        a.b object = new a().new b() {
+        };
+
+        class a {
+            class b { // ok
+                private b() {
+                }
+            }
+        }
+        class m { // violation
+            private m() {
+            }
+            class q { // ok
+                private q() {
+                }
+            }
+        }
+    }
+
+    class m { // ok
+        private m() {
+        }
+        class q { // violation
+            private q() {
+            }
+        }
+        j.m.q obj = new j().new m().new q() {
+        };
+    }
+
+    class jasper { // ok
+        private jasper() {
+        }
+        class a {
+            class b { // violation
+                private b() {
+                }
+            }
+            m obj = new m() {
+            };
+        }
+    }
+
+    jasper obj = new com.puppycrawl.tools.checkstyle.checks.design.finalclass
+        .InputFinalClassAnonymousInnerClass().new jasper() {
+    };
+
+    test3 obj2 = new test3() { // test3 is in another file
+    };
+}


Vyom-Yadav:
@nick-mancuso @pbludov @romani @rnveach @strkkk 
From teamcity:
> Parameter name isSearchedClassSuperClassOfAnonInnerClass is too long (41 > 32) 

I don't think the name can be shorter, it won't be that descriptive if we shorten it. 

> 'public' method typeDeclarationNameMatchingCount() with 'boolean' parameter 

I am also not quite happy with it but if we remove the boolean flag, we would have a lot of duplicated code, can I get some feedback on how we should proceed with the design, I don't feel too good with the current design.

strkkk:
> can I get some feedback on how we should proceed with the design

typical approach for such case - make 2 different public methods and make private method with a flag in parameters. In this way all this flag stuff to avoid duplication wont leak outside of class.

> isSearchedClassSuperClassOfAnonInnerClass

I dont like it either. Such long name is even more confusing as for me. May be something like `shouldSearchAnonSuperClass`?

Vyom-Yadav:
Github, generate report

Vyom-Yadav:
Github, generate report

nick-mancuso:
@Vyom-Yadav “and” in issues and PRs is always a red flag to me. Why can’t we separate this into two or more PRs?

Vyom-Yadav:
> @Vyom-Yadav “and” in issues and PR titles is always a red flag to me. Why can’t we separate this into two or more PRs?

@nick-mancuso I am dropping changes done in UnusedLocalVariable to fix false negative, I will make a separate PR for it, the rest of the changes would be difficult to segregate as they are tightly bound and keeping them together would be best IMO.

Vyom-Yadav:
Github, generate report

nick-mancuso:
@Vyom-Yadav NPE in openjdk 17 https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/514d736_2022085653/reports/diff/openjdk17/index.html

Vyom-Yadav:
> @Vyom-Yadav NPE in openjdk 17 https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/514d736_2022085653/reports/diff/openjdk17/index.html

Thats a non compilable file, will send fix when I am near my laptop.

Vyom-Yadav:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/utils/TypeDeclarationDescription.java
```diff
@@ -0,0 +1,90 @@
...
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+/**
+ * Maintains information about the type of declaration.
+ * Any ast node of type {@link TokenTypes#CLASS_DEF} or {@link TokenTypes#INTERFACE_DEF}
+ * or {@link TokenTypes#ENUM_DEF} or {@link TokenTypes#ANNOTATION_DEF}
+ * or {@link TokenTypes#RECORD_DEF} is considered as a type declaration.
+ */
+public class TypeDeclarationDescription {
```
strkkk, line 31: I see no need for separate class in utils package. This is used only in one check, so it is better to move it there and make it private static inner class.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -270,21 +280,75 @@ private void registerNestedSubclassToOuterSuperClasses(String qualifiedClassName
...
         final String dotAndClassName = PACKAGE_SEPARATOR.concat(className);
         return innerClasses.entrySet().stream()
                 .filter(entry -> entry.getKey().endsWith(dotAndClassName))
                 .map(Map.Entry::getValue)
                 .min((first, second) -> {
-                    return getClassDeclarationNameMatchingCountDiff(superClassName, first, second);
+                    return getClassDeclarationNameMatchingCountDiff(superOrOuterClassName, first,
+                         second, shouldSearchAnonSuperClass);
                 })
                 .orElse(null);
```
strkkk, line 353: why you return null here if in registerAnonymousInnerClassToSuperClass method you check for null anyway?

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -293,53 +357,52 @@ private ClassDesc getNearestClassWithSameName(String className, String superClas
...
-    private static int getClassDeclarationNameMatchingCountDiff(String superClassName,
-                                                                ClassDesc firstClass,
-                                                                ClassDesc secondClass) {
-        int diff = Integer.compare(
-            CheckUtil
-                .typeDeclarationNameMatchingCount(superClassName, secondClass.getQualifiedName()),
-            CheckUtil
-                .typeDeclarationNameMatchingCount(superClassName, firstClass.getQualifiedName()));
+    private static int getClassDeclarationNameMatchingCountDiff(String superOrOuterClassName,
+        ClassDesc firstClass, ClassDesc secondClass, boolean shouldSearchAnonSuperClass) {
```
strkkk, line 367: to avoid this stuff with flag and ugly (as for me) code, I can suggest following
```
private static int getClassDeclarationNameMatchingCountDiff(
        ClassDesc firstClass, ClassDesc secondClass, Function<String, Integer> countProvider){
diff = Integer.compare(
countProvider.apply(secondClass.getQualifiedName()),
countProvider.apply(firstClass.getQualifiedName()),
)
// rest is same
```

So, when you invoke this code, you will have something like
```
getClassDeclarationNameMatchingCountDiff(firstClass, secondClass,
 name -> CheckUtil.typeDeclarationMatchingCountAnonSuperClass(superOrOuterClassName, name)
```

??

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -170,12 +181,9 @@ public void visitToken(DetailAST ast) {
 
             case TokenTypes.LITERAL_NEW:
                 if (ast.getFirstChild() != null
-                        && ast.getLastChild().getType() == TokenTypes.OBJBLOCK) {
-                    for (ClassDesc classDesc : classes) {
-                        if (doesNameOfClassMatchAnonymousInnerClassName(ast, classDesc)) {
-                            classDesc.registerAnonymousInnerClass();
-                        }
-                    }
+                    && ast.getLastChild().getType() == TokenTypes.OBJBLOCK) {
```
strkkk, line 184: indentation

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
```diff
@@ -621,6 +621,42 @@ public static int typeDeclarationNameMatchingCount(String patternClass,
...
+     * inner class object of a nested class which isn't true in case of extending classes as you
+     * can't extend nested classes.
+     * </p>
+     *
+     * @param patternTypeDeclaration type declaration against which the given type declaration has
+     *                               to be matched
+     * @param typeDeclarationToBeMatched type declaration to be matched
+     * @return type declaration matching count
+     */
+    public static int typeDeclarationMatchingCountAnonSuperClass(String patternTypeDeclaration,
```
strkkk, line 643: again, is this used somewhere else outside single check? If not, there is no need to put it to utils

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
```diff
@@ -621,6 +621,42 @@ public static int typeDeclarationNameMatchingCount(String patternClass,
...
+        String typeDeclarationToBeMatched) {
+        final int minLength = Math
+            .min(typeDeclarationToBeMatched.length(), patternTypeDeclaration.length());
+        int result = 0;
+        for (int i = 0; i < minLength
+            && patternTypeDeclaration.charAt(i) == typeDeclarationToBeMatched.charAt(i); ++i) {
+            if (patternTypeDeclaration.charAt(i) == PACKAGE_SEPARATOR
+                || i == minLength - 1
+                && typeDeclarationToBeMatched.length() > minLength
+                && typeDeclarationToBeMatched.charAt(minLength) == PACKAGE_SEPARATOR) {
```
strkkk, line 653:
```
i == minLength - 1
                && typeDeclarationToBeMatched.length() > minLength
                && typeDeclarationToBeMatched.charAt(minLength) == PACKAGE_SEPARATOR
```
it looks like there is no need to put this inside cycle. You can just check this before and set result to `minLength - 1` in this case. And if this is false, go for cycle

strkkk (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/utils/TypeDeclarationDescription.java
Vyom-Yadav: For now, this is used by only this check, I plan to use it in UnusedLocalVariable too, had this committed earlier, I will send a separate fix for it.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: Done, used Optional

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -293,53 +357,52 @@ private ClassDesc getNearestClassWithSameName(String className, String superClas
...
-            outerClassQualifiedName = classes.peek().getQualifiedName();
+    private String getQualifiedTypeDeclarationName(DetailAST typeDeclarationAst) {
+        final String className = typeDeclarationAst.findFirstToken(TokenTypes.IDENT).getText();
+        String outerTypeDeclarationQualifiedName = null;
+        if (!typeDeclarations.isEmpty()) {
+            outerTypeDeclarationQualifiedName = typeDeclarations.peek().getQualifiedName();
         }
-        return CheckUtil
-            .getQualifiedTypeDeclarationName(packageName, outerClassQualifiedName, className);
+        return CheckUtil.getQualifiedTypeDeclarationName(packageName,
```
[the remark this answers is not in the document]
Vyom-Yadav: It is intentional, the complete name is built in 2 steps, one method is stateful whereas the other one is stateless.

>Looks confusing

We could refactor it to eliminate the stateful method and do it in the place where the method is called, but I feel this type of method is pretty common, a method calling another method with the same name with different arguments.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -293,53 +357,52 @@ private ClassDesc getNearestClassWithSameName(String className, String superClas
...
+    private static int getClassDeclarationNameMatchingCountDiff(String superOrOuterClassName,
+        ClassDesc firstClass, ClassDesc secondClass, boolean shouldSearchAnonSuperClass) {
+        int diff;
+        if (shouldSearchAnonSuperClass) {
+            diff = Integer.compare(
+                CheckUtil
+                    .typeDeclarationMatchingCountAnonSuperClass(superOrOuterClassName,
+                                                                secondClass.getQualifiedName()),
+                CheckUtil
+                    .typeDeclarationMatchingCountAnonSuperClass(superOrOuterClassName,
```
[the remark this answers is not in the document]
Vyom-Yadav: Sounds good to me too, changed it to `getAnonSuperTypeDeclarationMatchingCount`, could be an interface too, this method will be used to fix #11374

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: Noice.
It is a lot better than using flags, this can be used to replace flags in a lot of cases, noted down.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: Done

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
Vyom-Yadav: Will be used for fixing #11374

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
Vyom-Yadav: Yes, this flag can be extracted, done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -256,90 +267,120 @@ private static boolean shouldBeDeclaredAsFinal(ClassDesc desc) {
...
-        return ast.getFirstChild().getText().equals(className[className.length - 1]);
+                    int diff = Integer.compare(
+                        countProvider.apply(second.getQualifiedName()),
+                        countProvider.apply(first.getQualifiedName())
+                    );
+                    if (diff == 0) {
+                        diff = Integer.compare(first.getDepth(), second.getDepth());
+                    }
+                    return diff;
+                });
```
Vyom-Yadav, line 366: @strkkk Should I refactor this to-
```
.max(Comparator.comparingInt(
    (ClassDesc classDesc) -> countProvider.apply(classDesc.getQualifiedName()))
    .thenComparingInt(classDesc -> -classDesc.getDepth()));
```
The only drawback that I feel there is that we have to use `-` (minus) sign in the latter comparison. What do you think?

Vyom-Yadav:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
strkkk: I thought about this and I find this way too noisy. As for me, this way would be better if there was possibility to use method refs instead of lambdas in both comparisons.

By the way, this comparison method can be more simplified with
```
final int diff = ...
return diff != 0 ? diff : second.getDepth() - first.getDepth();
```

src/main/java/com/puppycrawl/tools/checkstyle/utils/TypeDeclarationDescription.java
strkkk: I get that, but my point is to do not include this-will-be-needed-somewhere-else code until there is an actual need for it.

So I suggest to make extraction to common place with separate commit in next PR for issue you mentioned.

Please keep in mind that actual fix can be changed, and such things "for future" can be useless. Therefore it is better to keep code clean.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
strkkk: original indentation was like this, and there is no reason to change it.
Plus, usually second condition is indented, even if it is on same level as first. For your example I would say that indentation should look like
```
a 
/*1 indent */&& b
/*1 indent */&& (c 
/*2 indent */    || d)
/*1 indent */ && e
```

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
```diff
@@ -621,6 +621,45 @@ public static int typeDeclarationNameMatchingCount(String patternClass,
...
+        int result = 0;
+        final boolean shouldCountBeUpdatedAtLastCharacter =
+            typeDeclarationToBeMatched.length() > minLength
+                && typeDeclarationToBeMatched.charAt(minLength) == PACKAGE_SEPARATOR;
+
+        for (int i = 0; i < minLength
+            && patternTypeDeclaration.charAt(i) == typeDeclarationToBeMatched.charAt(i); ++i) {
+            if (i == minLength - 1
+                && shouldCountBeUpdatedAtLastCharacter
+                || patternTypeDeclaration.charAt(i) == PACKAGE_SEPARATOR) {
```
strkkk, line 656: it is better now, but there is a still duplication that can be avoided
what about
```
        final boolean shouldCountBeUpdatedAtLastCharacter =
            typeDeclarationToBeMatched.length() > minLength;
// in for loop
if (patternTypeDeclaration.charAt(i) == PACKAGE_SEPARATOR && (i != minLength - 1
                || shouldCountBeUpdatedAtLastCharacter) {
```
in this case there is no duplication.

Also I do not like that we compare `typeDeclarationToBeMatched` length with min length, seems a bit odd.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -506,12 +513,12 @@ private boolean isDeclaredAsAbstract() {
         }
 
         /**
-         * Does class have an anonymous inner class.
+         * Is the class super class of an anonymous inner class.
```
strkkk, line 516: Returns whether class is ...

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -256,90 +267,120 @@ private static boolean shouldBeDeclaredAsFinal(ClassDesc desc) {
...
+                                                .getAnonSuperTypeDeclarationMatchingCount(
+                                                    outerTypeDeclName, className);
+                                        });
+        superClassDesc.ifPresentOrElse(ClassDesc::registerSuperClassOfAnonymousInnerClass,
+                                       () -> {
+                                           Optional.ofNullable(innerClasses.get(superClassName))
+                                               .ifPresent(
+                                                   ClassDesc
+                                                       ::registerSuperClassOfAnonymousInnerClass);
+                                       });
```
strkkk, line 312: I do not like this stairs-style, it distracts from following main flow of the logic. Please make it more flat (same for other method with stairs)

I can suggest following:
1. no need for intermediate optional variable. can be one flow
2. registerSuperClassOfAnonymousInnerClass and registerNestedSubclass can accept nulls, therefore eliminating need to make additional wrap into optional here
3. Try to make CheckUtil usage singleline. I guess NeedBraces rule forced you to add return and braces. So, name of the method can be shortened and lambda extracted as local variable

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -256,90 +267,120 @@ private static boolean shouldBeDeclaredAsFinal(ClassDesc desc) {
...
+     *         static class b {
+     *         }
+     *     }
+     *
+     *     class j {
+     *         a.b object = new a.b() { // Object of Main.j.a.b and not of Main.a.b
+     *         };
+     *
+     *         static class a {
+     *             static class b {
```
strkkk, line 334: although it is just an example, please use conventional naming for classes - camelCase

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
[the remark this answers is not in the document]
strkkk: > method calling another method with the same name with different arguments.

this is ok in case of overloading. But it seems confusing to me and can be error-prone (e.g. if you just search for method by text search). I suggest to rename method in check, can be `extractQualifiedTypeName`

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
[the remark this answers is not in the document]
strkkk: My main point is to make name shorter. I think you can safely remove `Declaration` from method names, it wont hurt.

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
strkkk: https://github.com/checkstyle/checkstyle/pull/11468#discussion_r846579696

strkkk (changes requested):
see above

pbludov (approved):
Please add a test case for **local** classes:
```java
class TestClass {
    static class a {
        static class b { // violation
            private b() {}
        }
    }       

    void method() {
        interface a {
            static class b { // ok
                private b() {}
            }
            static final class c {
                private c() {}
            }
        }       

        a.b object = new a.b() {
        };
     
    }
}
```

Vyom-Yadav:
> Please add a test case for **local** classes:

@pbludov FinalClass doesn't support local classes (For both normal and anon inner classes).
Would you like the fix with this PR or in a separate issue?

pbludov:
>  Would you like the fix with this PR or in a separate issue?
It is better to create a separate issue. This case is very unusual.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TypeDeclarationDescription.java
Vyom-Yadav: Got it, done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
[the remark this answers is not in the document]
Vyom-Yadav: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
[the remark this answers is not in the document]
Vyom-Yadav: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: > As for me, this way would be better if there was possibility to use method refs instead of lambdas in both comparisons.

I changed the implementation to use method ref in both comparisions.

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
Vyom-Yadav: > it is better now, but there is a still duplication that can be avoided

Where is it, I am unable to see it (the condition is the same, it is showing outdated as I changed `i` -> `index`).

> Also I do not like that we compare typeDeclarationToBeMatched length with min length, seems a bit odd.

There are cases when its length is less than min length, the condition is necessary for avoiding StringIndexOutOfBoundsException

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -367,24 +394,128 @@ private static String getClassNameFromQualifiedName(String qualifiedName) {
...
+     * or {@link TokenTypes#ENUM_DEF} or {@link TokenTypes#ANNOTATION_DEF}
+     * or {@link TokenTypes#RECORD_DEF} is considered as a type declaration.
+     */
+    private static class TypeDeclarationDescription {
 
-        /** Qualified class name(with package). */
+        /**
+         * Complete type declaration name with package name and outer type declaration name.
+         */
         private final String qualifiedName;
```
strkkk, line 449: just in case - for private static classes there is no need for getters, you can access its fields directly

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
strkkk: To make it clearer, I resolve this and respond in separate comment

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -256,90 +269,104 @@ private static boolean shouldBeDeclaredAsFinal(ClassDesc desc) {
...
+                                                         String outerTypeDeclName) {
+        final String superClassName = CheckUtil.getShortNameOfAnonInnerClass(literalNewAst);
+
+        getNearestClassWithSameName(superClassName, classDesc -> {
+            return getAnonSuperTypeMatchingCount(outerTypeDeclName, classDesc.getQualifiedName());
+        }).ifPresentOrElse(ClassDesc::registerSuperClassOfAnonymousInnerClass, () -> {
+            Optional
+                .ofNullable(innerClasses.get(superClassName))
+                .ifPresent(ClassDesc::registerSuperClassOfAnonymousInnerClass);
+        });
```
strkkk, line 304: it is better now, but can be improved, like
```
Function getAnonCount = classDesc -> getAnonSuperTypeMatchingCount(outerTypeDeclName, classDesc.getQualifiedName());
getNearestClassWithSameName(superClassName,getAnonCount)
.or(Optional.ofNullable(innerClasses.get(superClassName))
.ifPresent(...)
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -256,90 +269,104 @@ private static boolean shouldBeDeclaredAsFinal(ClassDesc desc) {
...
-                nearest.registerNestedSubclass();
-            }
+            getNearestClassWithSameName(superClassName, classDesc -> {
+                return CheckUtil.typeDeclarationNameMatchingCount(qualifiedClassName,
+                                                                  classDesc.getQualifiedName());
+            }).ifPresentOrElse(ClassDesc::registerNestedSubclass, () -> {
+                Optional
+                    .ofNullable(innerClasses.get(superClassName))
+                    .ifPresent(ClassDesc::registerNestedSubclass);
+            });
```
strkkk, line 281: same here

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
strkkk: I will respond a bit later, it requires some long description)

strkkk (changes requested):
see above

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: Yes, we can, I created getters for using method reference here-
```
.max(Comparator
    .comparingInt(countProvider::apply)
    .thenComparingInt(ClassDesc::getDepth)) // <--- Over here
```
Now I created a getter for one field and it would look weird if I didn't create it for the others.

access modifier of getters can be default but XdocsJavaDocsTest doesn't allow it (can be a bug or a feature idk)

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: Done

Vyom-Yadav:
Github, generate report

Vyom-Yadav:
> FinalClassCheck: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/871b2df_2022092810/reports/diff/index.html

created https://github.com/checkstyle/contribution/pull/603

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
Vyom-Yadav: Yes, we can't have two consecutive package separators but the condition is not redundant, this is the only case where we need not evaluate 
`patternTypeDeclaration.charAt(index) == packageSeparator` as the condition before `||` will be true, but in all other cases, this condition is important, I initially had some pitest failures but after looking closely, it was due to missing examples and not redundant conditions.

strkkk:
@Vyom-Yadav is this failure relevant? https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/871b2df_2022092810/reports/diff/openjdk17/index.html#A3
I see you have added this file to exclusions, but why there was no failure before?
Can we add one more test input with simple package-info file to make sure there is no failures?

Vyom-Yadav:
> @Vyom-Yadav is this failure relevant? https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/871b2df_2022092810/reports/diff/openjdk17/index.html#A3

No, the file is non-compilable.

> I see you have added this file to exclusions, but why there was no failure before? Can we add one more test input with simple package-info file to make sure there is no failures?

There was no failure because earlier I was using custom `projects-to-test-on.properties` (see issue description edit history), waiting for https://github.com/checkstyle/contribution/pull/602 to merge, after merging I thought we are good to go, but I missed adding an exclusion to one more file, which was addressed in https://github.com/checkstyle/contribution/pull/603

strkkk (approved):

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/finalclass/InputFinalClassNestedInRecord.java
```diff
@@ -0,0 +1,49 @@
+/*
+FinalClass
+
+
+*/
+
+//non-compiled with javac: Compilable with Java16
```
nick-mancuso, line 7: We generally use the lowest language level that we can here, is there some syntax that prevents us from using 14?

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -367,24 +392,128 @@ private static String getClassNameFromQualifiedName(String qualifiedName) {
...
+        /**
+         * Complete type declaration name with package name and outer type declaration name.
+         */
         private final String qualifiedName;
 
+        /**
+         * Depth of nesting of type declaration. The depth of top-level class is 0,
+         * a class nested under it has a depth of -1 and so on.
+         */
+        private final int depth;
```
nick-mancuso, line 453: Why is depth negative?

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -367,24 +392,128 @@ private static String getClassNameFromQualifiedName(String qualifiedName) {
...
+         * @return ast node of the type declaration
+         */
+        protected DetailAST getTypeDeclarationAst() {
+            return typeDeclarationAst;
+        }
+    }
+
+    /**
+     * Maintains information about class's ctors. {@link TypeDeclarationDescription} does not
+     * maintain information about classes, a subclass called {@link ClassDesc}
```
nick-mancuso, line 506: It seems odd to have this class to have a link to itself and refer to itself in the "third person".

nick-mancuso (changes requested):
Items:

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/finalclass/InputFinalClassNestedInRecord.java
Vyom-Yadav: Yes, it is compilable with java 14, I usually used static declarations in inner classes in these types of examples which are only compilable with 16+, didn't use it in this one, changed it to 14.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: I moved that point to `TypeDeclarationDescription`, done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: Depth is always compared in an inverted manner, this change was done to use method ref-
```
.max(Comparator
    .comparingInt(countProvider::apply)
    .thenComparingInt(ClassDesc::getDepth)) // <--- Over here
```
It looks cleaner compared to using lambda and explicitly using a minus sign.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/finalclass/InputFinalClassNestedInRecord.java
nick-mancuso: Resolved.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
nick-mancuso: Resolved.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
nick-mancuso: > looks cleaner

I am not sure this is a good reason to do something, if it is not logical :) Wouldn't it make sense it have depth be positive, and use `.min` since we are finding the nearest class?

nick-mancuso: @Vyom-Yadav ping

Vyom-Yadav: > I am not sure this is a good reason to do something, if it is not logical :) Wouldn't it make sense it have depth be positive, and use .min since we are finding the nearest class?

Using `.max` shouldn't be a problem as we are finding the class with the most matching count, having negative depth shouldn't be a problem, the deeper you go the more negative it becomes.

nick-mancuso: > the deeper you go the more negative it becomes.

This is exactly my point; this is not intuitive, and you could just use `min` and reverse the sorting order, this still allows fancy syntax, method references, etc. One does not say "we are negative five feet deep in the water" similar to how it would not make sense to say that "we are negative three nested classes deep" .

Think about next maintainer that comes along; when we use depth in the real world, if it is in negative terms, this would be "elevation". Do other checks that use `depth` consider depth to be negative? It does not appear that depth was negative in this check previously either.

Vyom-Yadav: It won't be possible to use `Comparator.comparingInt(..)` with method references because either depth or matching count needs to be with the opposite sign. Inverting the sign of matching count isn't intuitive as well, so I reverted back to our original approach.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -256,90 +267,108 @@ private static boolean shouldBeDeclaredAsFinal(ClassDesc desc) {
...
      *
      * @param className name of the class
-     * @param superClassName name of the super class
-     * @return true if there is another class with same name.
+     * @param countProvider the function to apply to calculate the name matching count
+     * @return {@link Optional} of {@link ClassDesc} object of the nearest class with the same name.
      * @noinspection CallToStringConcatCanBeReplacedByOperator
      */
-    private ClassDesc getNearestClassWithSameName(String className, String superClassName) {
+    private Optional<ClassDesc> getNearestClassWithSameName(String className,
```
nick-mancuso, line 340: @Vyom-Yadav I think we can still use fancy syntax as you proposed before, with no negatives in intuitive manner:

```suggestion
    private Optional<ClassDesc> getNearestClassWithSameName(String className,
                                                            Function<ClassDesc, Integer> countProvider) {
        final String dotAndClassName = PACKAGE_SEPARATOR.concat(className);
        final Comparator<ClassDesc> longestMatch = Comparator.comparingInt(countProvider::apply);

        return innerClasses.entrySet().stream()
                           .filter(entry -> entry.getKey().endsWith(dotAndClassName))
                           .map(Map.Entry::getValue)
                           .min(longestMatch.reversed().thenComparingInt(ClassDesc::getDepth));
    }
```

Please try this.

Vyom-Yadav: Thanks for the suggestion, implemented. This Github feature is also nice.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
nick-mancuso: Resolved at https://github.com/checkstyle/checkstyle/pull/11468#discussion_r873170995

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
nick-mancuso: Resolved.

nick-mancuso (changes requested):
Few minor items, also please generate report after next push, since we changed logic in `getNearest...`

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -367,24 +390,128 @@ private static String getClassNameFromQualifiedName(String qualifiedName) {
...
+     * @param typeDeclarationToBeMatched type declaration to be matched
+     * @return type declaration matching count
+     */
+    private static int getAnonSuperTypeMatchingCount(String patternTypeDeclaration,
+                                                    String typeDeclarationToBeMatched) {
+        final int typeDeclarationToBeMatchedLength = typeDeclarationToBeMatched.length();
+        final int minLength = Math
+            .min(typeDeclarationToBeMatchedLength, patternTypeDeclaration.length());
+        final char packageSeparator = PACKAGE_SEPARATOR.charAt(0);
+        int result = 0;
```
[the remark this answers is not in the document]
Vyom-Yadav: Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -367,24 +390,128 @@ private static String getClassNameFromQualifiedName(String qualifiedName) {
...
+        final boolean shouldCountBeUpdatedAtLastCharacter =
+            typeDeclarationToBeMatchedLength > minLength
+                && typeDeclarationToBeMatched.charAt(minLength) == packageSeparator;
+
+        for (int index = 0; index < minLength
+            && patternTypeDeclaration.charAt(index) == typeDeclarationToBeMatched.charAt(index);
+             ++index) {
+            if (index == minLength - 1
+                && shouldCountBeUpdatedAtLastCharacter
```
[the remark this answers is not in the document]
Vyom-Yadav: 'this' is ambiguous over here, but I assumed you were talking about `++index`, I modified it a bit, please see if it looks fine.

Vyom-Yadav:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -367,24 +390,127 @@ private static String getClassNameFromQualifiedName(String qualifiedName) {
...
+
+        int result = 0;
+        for (int index = 0; index < minLength && patternTypeDeclaration.charAt(
+            index) == typeDeclarationToBeMatched.charAt(index); ++index) {
+            if (index == minLength - 1
+                && shouldCountBeUpdatedAtLastCharacter
+                || patternTypeDeclaration.charAt(index) == packageSeparator) {
+                result = index;
+            }
+        }
```
nick-mancuso, line 430: Please do:

```suggestion
        for (int index = 0; 
                index < minLength 
                    && patternTypeDeclaration.charAt(index) == typeDeclarationToBeMatched.charAt(index); 
                index++) {
                
            if (index == minLength - 1 && shouldCountBeUpdatedAtLastCharacter
                    || patternTypeDeclaration.charAt(index) == packageSeparator) {
                result = index;
            }
        }
```

To make this section easier to read by humans.

nick-mancuso (changes requested):
Clarification:

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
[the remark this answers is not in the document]
nick-mancuso: Resolved

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -367,24 +390,128 @@ private static String getClassNameFromQualifiedName(String qualifiedName) {
...
+            .min(typeDeclarationToBeMatchedLength, patternTypeDeclaration.length());
+        final char packageSeparator = PACKAGE_SEPARATOR.charAt(0);
+        int result = 0;
+        final boolean shouldCountBeUpdatedAtLastCharacter =
+            typeDeclarationToBeMatchedLength > minLength
+                && typeDeclarationToBeMatched.charAt(minLength) == packageSeparator;
+
+        for (int index = 0; index < minLength
+            && patternTypeDeclaration.charAt(index) == typeDeclarationToBeMatched.charAt(index);
+             ++index) {
```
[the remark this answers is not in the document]
Vyom-Yadav: Yes, it does not make a difference in java, [this](https://stackoverflow.com/questions/4831748/is-i-really-faster-than-i-in-for-loops-in-java?noredirect=1&lq=1), clarified it, it was just some medium article that was incorrect which led to this.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
Vyom-Yadav: Done, had to change indent to +2 to adhere to line length.

nick-mancuso (approved):

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
nick-mancuso: Resolved.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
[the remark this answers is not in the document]
nick-mancuso: Resolved.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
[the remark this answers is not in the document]
nick-mancuso: Resolved.

Vyom-Yadav:
@romani Ping

Vyom-Yadav:
Github, generate report

PR #11469, opened 2022-03-26
Issue #11140: Resolve PMB_POSSIBLE_MEMORY_BLOAT issue 1

Rahulkhinchi03:
Related to https://github.com/checkstyle/checkstyle/issues/11140

```
PMB_POSSIBLE_MEMORY_BLOAT
This class defines static fields that are Collections, StringBuffers, or StringBuilders that do not appear to have any way to clear or reduce their size. That is, a collection is defined and has method calls like
{add(), append(), offer(), put(), ...}
with no method calls to removal methods like
{clear(), delete(), pop(), remove(), ...}
This means that the collection in question can only ever increase in size, which is a potential cause of memory bloat.

If this collection is a list, set or otherwise of static things (e.g. a List>String> for month names), consider adding all of the elements in a static initializer, which can only be called once:
 private static List<String> monthNames = new ArrayList<String>(); static { monthNames.add("January"); monthNames.add("February"); monthNames.add("March"); ... } 

```

### DIFF
--- a/config/spotbugs-exclude.xml
+++ b/config/spotbugs-exclude.xml
@@ -77,7 +77,6 @@
     <!-- These fields are restricted caches which must remain populated until the process ends. -->
     <Bug pattern="PMB_POSSIBLE_MEMORY_BLOAT"/>
     <Or>
-      <Class name="com.puppycrawl.tools.checkstyle.PackageObjectFactory"/>
       <Class name="com.puppycrawl.tools.checkstyle.checks.SuppressWarningsHolder"/>
     </Or>
   </Match>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
@@ -20,6 +20,7 @@
 package com.puppycrawl.tools.checkstyle;
 
 import java.io.IOException;
+import java.util.AbstractMap;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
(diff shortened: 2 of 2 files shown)


Rahulkhinchi03:
GitHub, rebase

Rahulkhinchi03:
GitHub, rebase

nick-mancuso:
We only use this map in one method:

```
➜  checkstyle git:(master) grep -R "NAME_TO_FULL_MODULE_NAME\.get(" -B1
src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java-    private Object createFromStandardCheckSet(String name) throws CheckstyleException {
src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java:        final String fullModuleName = NAME_TO_FULL_MODULE_NAME.get(name);
--
...
```

Why can't we just use lazy initialization? It looks like the changes in this PR are just obfuscating the static initialization of ` NAME_TO_FULL_MODULE_NAME`  for spotbugs check. Please explain/ prove how your changes improve performance.

romani:
I would rather keep suppression, than do such update that is hard to read

nick-mancuso:
From documentation:
`If this collection is a list, set or otherwise of static things (e.g. a List>String> for month names), consider adding all of the elements in a static initializer, which can only be called once`

So, the issue is that we never remove or reduce the number of elements in this map within this class, but it is not clear to spotbugs that we do use a static initialization since we do it from another method.

I agree with @romani that we should keep suppression, but let's add a comment to the supression explaining why we keep it (readability) and link to discussion here.

romani:
We should not update code and suppression should stay forever.

romani:
I marked PMB_POSSIBLE_MEMORY_BLOAT as done in issue

PR #11472, opened 2022-03-26
Issue #11446: Update EmptyLineSeparatorCheckTest to use verifyWithInlineConfigParser

Rahulkhinchi03:
Resolves update for EmptyLineSeparatorCheckTest in https://github.com/checkstyle/checkstyle/issues/11446.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheckTest.java
@@ -183,12 +183,9 @@ public class EmptyLineSeparatorCheckTest
     @Test
     public void testJavadocCommentAfterPackageWithImports() throws Exception {
         final DefaultConfiguration checkConfig = createModuleConfig(EmptyLineSeparatorCheck.class);
-        final String[] expected = {
-            "2:1: " + getCheckMessage(MSG_SHOULD_BE_SEPARATED, "/*"),
-        };
-        verify(checkConfig,
-                getPath("InputEmptyLineSeparatorJavadocCommentAfterPackage.java"),
-                expected);
+
+        execute(checkConfig,
+                getPath("InputEmptyLineSeparatorJavadocCommentAfterPackage.java"));
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheckTest.java
```diff
@@ -186,9 +186,8 @@ public void testJavadocCommentAfterPackageWithImports() throws Exception {
         final String[] expected = {
             "2:1: " + getCheckMessage(MSG_SHOULD_BE_SEPARATED, "/*"),
         };
-        verify(checkConfig,
-                getPath("InputEmptyLineSeparatorJavadocCommentAfterPackage.java"),
-                expected);
+        execute(checkConfig,
+                getPath("InputEmptyLineSeparatorJavadocCommentAfterPackage.java"));
```
romani, line 190: Remove "expected" variable

romani (changes requested):
Items:

src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheckTest.java
Rahulkhinchi03: Changes done

romani: You need to convert to new method only places where we expect exception.

All other should be converted to method that use inlined config in Input files.

Rahulkhinchi03: I believe that we should not use `execute `here as I am getting PMD error and If I am using `veriftWithInlineConfigParser` and changing the `config `in input files then all tests are passing.

But after changing the input files, I am getting code coverage issue.

romani: update test to not use "verify"  and use verifyWithInlineConfigParser

Issue #11475, opened 2022-03-27
Unit tests for Checkstyle GUI components


pbludov:
## Issue
Checkstyle has minimalist graphical interface, used by developers for troubleshooting and research purpose.
This feature is not important, and has no much support. There is no tests for GUI in Checkstyle (except for the model).
And there is a reason for this. GUI tests require a physical display and exclusive access to input interfaces.
As a result, it is possible to break the GUI after some refactoring. See #9004 for example.

## Idea
However, this can be done with [caciocavallo](https://github.com/CaciocavalloSilano/caciocavallo), a pure virtual AWT driver.
With Cacio, all GUI windows are invisible and untouchable. This means that we can use it in CI and on the developer's PC without distracting the user with unexpected pop-ups.

## Restrictions
- The behavior of AWT components depends on the toolkit. On Mac, Linux and Windows, anything may be different. And, of course, it will be something different for the virtual toolkit. So it is **impossible** to gain 100% coverage for GUI: we need, for example, some lines of Windows-specific code and such code can be tested only on Windows.
- Another restriction is *modal dialogs*: all of them should be *mocked* to prevent hanging of tests.

## Result
The goal of this issue is to set up initial configuration for GUI tests and write tests for some components to validate this approach.
The improvement of GUI coverage should be done in other issues.

romani:
@pbludov , I presume we need to keep this issue to remove all excludes on gui classes.

Better to create new issue with more details on how exactly do updates what is left to fo

pbludov:
Nick is working on #11482 right now. We should merge it. Then it will be possible to calculate the coverage %.

pbludov:
Continuation of this issue will be done in #11514

PR #11477, opened 2022-03-27
Issue #11475: unit tests for gui

pbludov:
Issue #11475 

Tests for `Main`, `MainFrame` and `TreeTable`.

### DIFF
--- a/config/version-number-rules.xml
+++ b/config/version-number-rules.xml
@@ -11,6 +11,12 @@
     </ignoreVersion>
   </ignoreVersions>
   <rules>
+    <rule groupId="com.github.caciocavallosilano" artifactId="cacio-tta">
+      <ignoreVersions>
+        <!-- The version 1.11 we are on is the last version with Java 11 support. -->
+        <ignoreVersion type="regex">.*</ignoreVersion>
+      </ignoreVersions>
+    </rule>
     <rule groupId="com.google.guava" artifactId="guava">
       <ignoreVersions>
         <ignoreVersion type="regex">.*-android</ignoreVersion>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/gui/MainFrameModel.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/gui/MainFrameModel.java
@@ -204,16 +204,15 @@ public class MainFrameModel {
                 reloadActionEnabled = true;
                 final DetailAST parseTree;
 
-                switch (parseMode) {
-                    case PLAIN_JAVA:
-                        parseTree = JavaParser.parseFile(file, JavaParser.Options.WITHOUT_COMMENTS);
-                        break;
-                    case JAVA_WITH_COMMENTS:
-                    case JAVA_WITH_JAVADOC_AND_COMMENTS:
-                        parseTree = JavaParser.parseFile(file, JavaParser.Options.WITH_COMMENTS);
-                        break;
-                    default:
-                        throw new IllegalArgumentException("Unknown mode: " + parseMode);
+                if (parseMode == ParseMode.PLAIN_JAVA) {
+                    parseTree = JavaParser.parseFile(file, JavaParser.Options.WITHOUT_COMMENTS);
+                }
+                else if (parseMode == ParseMode.JAVA_WITH_COMMENTS
+                        || parseMode == ParseMode.JAVA_WITH_JAVADOC_AND_COMMENTS) {
+                    parseTree = JavaParser.parseFile(file, JavaParser.Options.WITH_COMMENTS);
+                }
+                else {
+                    throw new IllegalArgumentException("Unknown mode: " + parseMode);
                 }
 
                 parseTreeTableModel.setParseTree(parseTree);
--- a/src/test/java/com/puppycrawl/tools/checkstyle/AbstractGuiTestSupport.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/AbstractGuiTestSupport.java
@@ -1,0 +1,69 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2022 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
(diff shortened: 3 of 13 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/gui/MainFrameModel.java
```diff
@@ -204,16 +204,15 @@ public void openFile(File file) throws CheckstyleException {
...
-                    case PLAIN_JAVA:
-                        parseTree = JavaParser.parseFile(file, JavaParser.Options.WITHOUT_COMMENTS);
-                        break;
-                    case JAVA_WITH_COMMENTS:
-                    case JAVA_WITH_JAVADOC_AND_COMMENTS:
-                        parseTree = JavaParser.parseFile(file, JavaParser.Options.WITH_COMMENTS);
-                        break;
-                    default:
-                        throw new IllegalArgumentException("Unknown mode: " + parseMode);
+                if (parseMode == ParseMode.PLAIN_JAVA) {
```
pbludov, line 207: `switch` replaced with `if-else if-else` to gain 100% coverage with java 17.

pbludov:
Master branch dependencies:
```
[INFO] +- com.google.truth:truth:jar:1.1.3:test
[INFO] |  +- junit:junit:jar:4.13.2:test
[INFO] |  |  \- org.hamcrest:hamcrest-core:jar:1.3:test
[INFO] |  +- com.google.auto.value:auto-value-annotations:jar:1.8.1:test
[INFO] |  \- org.ow2.asm:asm:jar:9.1:test
```

Patch dependencies:
```
[INFO] +- com.github.caciocavallosilano:cacio-tta:jar:1.11:test
[INFO] |  +- junit:junit:jar:4.10:test
[INFO] |  |  \- org.hamcrest:hamcrest-core:jar:1.1:test
[INFO] |  +- com.github.caciocavallosilano:cacio-shared:jar:1.11:test
[INFO] |  +- org.assertj:assertj-swing-junit:jar:3.9.2:test
[INFO] |  |  +- org.assertj:assertj-swing:jar:3.9.2:test
[INFO] |  |  |  +- org.assertj:assertj-core:jar:3.11.1:test
[INFO] |  |  |  \- org.easytesting:fest-util:jar:1.2.5:test
[INFO] |  |  \- org.easytesting:fest-reflect:jar:1.4.1:test
[INFO] |  \- com.jidesoft:jide-oss:jar:3.5.11:test
[INFO] +- com.google.truth:truth:jar:1.1.3:test
[INFO] |  +- com.google.auto.value:auto-value-annotations:jar:1.8.1:test
[INFO] |  \- org.ow2.asm:asm:jar:9.1:test
```

Both `com.google.truth` and `com.github.caciocavallosilano:cacio-tta` has dependency on junit4.
This should be fixed in a separate commit.

romani (approved):
Ok, let move to next level of coverage.

nick-mancuso (approved):

config/version-number-rules.xml
```diff
@@ -11,6 +11,12 @@
     </ignoreVersion>
   </ignoreVersions>
   <rules>
+    <rule groupId="com.github.caciocavallosilano" artifactId="cacio-tta">
+      <ignoreVersions>
+        <!-- Version 1.17 is built with Java17 bytecode, we need Java11  -->
+        <ignoreVersion type="exact">1.17</ignoreVersion>
```
rnveach, line 17: Shouldn't this be version 1.17 or later? I doubt they will downgrade to Java 11 afterwards.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractGuiTestSupport.java
```diff
@@ -0,0 +1,69 @@
...
+import java.awt.Component;
+import java.awt.Container;
+import java.awt.GraphicsEnvironment;
+
+import org.junit.jupiter.api.BeforeAll;
+
+/**
+ * Abstract base class for testing GUI components.
+ */
+public abstract class AbstractGuiTestSupport extends AbstractModuleTestSupport {
```
rnveach, line 33: I am not seeing why this needs to be ModuleTest support. I am not seeing any `verify` method calls or anything related with running a module.

config/version-number-rules.xml
pbludov: I tried using version 1.17. It works on my PC but fails in CI because version 1.17 is not compatible with Java11.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractGuiTestSupport.java
pbludov: Done. Changed to `AbstractPathTestSupport`.

config/version-number-rules.xml
rnveach: @pbludov My comment is on the following section:
> type="exact"

Isn't this specifying the version of `caciocavallosilano`? It is saying we will only ignore that version of `caciocavallosilano`, right?

pbludov: Yes. I'd like to see what the next version of `caciocavallosilano` will be.
After that, it will be clear how to compose the pattern.

rnveach: Based on their history, won't it most likely be 18+ if not 2? https://mvnrepository.com/artifact/com.github.caciocavallosilano/cacio-tta
Why can't we just assume that now?

pbludov: The next release should be 1.21. But it could be 2.0. 
I don't know how to write a readable template that will pass 1.11.1, but reject 1.21.
Please suggest.

rnveach: Unless they plan to downgrade Java version, we should just drop the version completely and ignore all versions. If the current one we are on is the last Java 11 supports, then we will have to stick with it always. So it should be similar to what we have for `maven-release-plugin` in this same file.

pbludov: done.

romani:
Appveyor issue is not related to update, we need to bump version of maven, I run into it at contribution repo, fix https://github.com/checkstyle/contribution/commit/07b50f3b82ab745ed3bec6102196f9c92f4dba96#diff-92ab9a36df5d8e9f7076f2fdec59492d1ac2d9cf27ea046767a7fc4d542ef3dcR14

rnveach (approved):

rnveach:
@romani if we are good to merge, then please merge.

PR #11480, opened 2022-03-28
Issue #11163: Enforce file size on InputJavadocStyle2

Rahulkhinchi03:
Resolves update for InputJavadocStyle2 in https://github.com/checkstyle/checkstyle/issues/11163


src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheckTest.java
```diff
@@ -108,25 +108,51 @@ public void testDefaultSettings()
     }
 
     @Test
-    public void testFirstSentence() throws Exception {
```
nick-mancuso, line 111: Why not make the naming convention the same for test method and file/class name?

nick-mancuso (changes requested):
Item:

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheckTest.java
Rahulkhinchi03: You mean `testFirstSentance2a()` ?

nick-mancuso: That seems to make more sense to me, or it would be even better to name the test by some defining characteristic of check config, i.e. `testFirstSentenceCheckHtmlFalseOne` or something.

Rahulkhinchi03: The `InputJavadocStyle` has total 10 input files which must be created into 40 different files. 

I am thinking to name inputs like:
2a, 2b, 2c, 2d
3a, 3b, 3c, 3d
4a, 4b, 4c, 4d..

And for test cases, just like you shared.

nick-mancuso: To me, if I was a new contributor, I would wonder why we have test method like `testFirstSentenceOne` with test input `InputJavadocStyle2a.java`. These two names do not seem to be related. What I am getting at, is why not name the test method and input file similarly?

Example 1:

`testFirstSentenceOne` with input `InputJavadocStyleFirstSentenceOne.java`

nick-mancuso: Or, even better, `testJavadocStyleFirstSentenceOne` and `InputJavadocStyleFirstSentenceOne.java`

romani: Yes, let's match test and input name

Rahulkhinchi03:
@nick-mancuso Why does pitest-coding fails most of the time? Like all the tests pass and just that fails. I tried to found the error and it says actual and ignore txt file differs.

romani (approved):

PR #11499, opened 2022-04-02
Issue #11166: remove usages of getFileContents

strkkk:
Part of #11166
Removes usages from
- EmptyLineSeparatorCheck

### DIFF
--- a/config/suppressions.xml
+++ b/config/suppressions.xml
@@ -114,7 +114,6 @@
   <suppress id="noUsageOfGetFileContentsMethod" files="RegexpCheck.java"/>
   <suppress id="noUsageOfGetFileContentsMethod" files="RegexpSinglelineJavaCheck.java"/>
   <suppress id="noUsageOfGetFileContentsMethod" files="MethodLengthCheck.java"/>
-  <suppress id="noUsageOfGetFileContentsMethod" files="EmptyLineSeparatorCheck.java"/>
   <suppress id="noUsageOfGetFileContentsMethod" files="AvoidEscapedUnicodeCharactersCheck.java"/>
 
   <!-- until https://github.com/checkstyle/checkstyle/issues/5234 -->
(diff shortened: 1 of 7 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractCheck.java
```diff
@@ -298,6 +298,15 @@ public final String getLine(int index) {
         return context.get().fileContents.getLine(index);
     }
 
+    /**
+     * Checks if the current file is a package-info.java file.
+     *
+     * @return true if the package file.
+     */
+    public final boolean inPackageInfo() {
```
romani, line 306: As to me too much details in abstract class, if we can not avoid checking file name, abstract class should have method to return name for Check to do anything required

romani (changes requested):
Can we split PR in two 
With inPackageInfo
And used lines

?

strkkk:
yes, this is better, since now some pitest issues with appear.

Created https://github.com/checkstyle/checkstyle/pull/11502

src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractCheck.java
strkkk: changed, see other pr https://github.com/checkstyle/checkstyle/pull/11502

romani:
please update this PR to not have updates for package-info.

strkkk:
> please update this PR to not have updates for package-info.

I will do it right after other PR is merged

romani:
What PR is blocker ?

pbludov (approved):

rnveach:
@strkkk ping

strkkk:
There are weird pitest failures after my changes, I do not have time to finish it.

PR #11503, opened 2022-04-03
Issue #11446: Update SuppressionCommentFilterTest to use verifyWithIn…

Rahulkhinchi03:
Resolves update for SuppressionCommentFilterTest in https://github.com/checkstyle/checkstyle/issues/11446.


src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionCommentFilterTest.java
```diff
@@ -401,13 +372,18 @@ public void testInvalidCheckFormat() throws Exception {
 
     @Test
     public void testInvalidMessageFormat() throws Exception {
+        final DefaultConfiguration treeWalkerConfig =
+            createModuleConfig(TreeWalker.class);
         final DefaultConfiguration filterConfig =
             createModuleConfig(SuppressionCommentFilter.class);
         filterConfig.addProperty("messageFormat", "e[l");
+        final DefaultConfiguration checkConfig =
+            createModuleConfig(ConstantNameCheck.class);
```
romani, line 381: Check isnot required, as we expect exception on initialization.

src/test/resources/com/puppycrawl/tools/checkstyle/filters/suppressioncommentfilter/InputSuppressionCommentFilter10.java
```diff
@@ -27,11 +18,6 @@
 applyToPrivate = (default)true
 
 
-com.puppycrawl.tools.checkstyle.checks.coding.IllegalCatch
-illegalClassNames = (default)Error, Exception, RuntimeException, Throwable, java.lang.Error, \
-                    java.lang.Exception, java.lang.RuntimeException, java.lang.Throwable
-
-
```
romani, line 34: whole content of this file is not relevant, please remove header comment, and whole body of class.
there should be only empty class.
As we do not use violation of this class.

Apply the same to another Input file.

romani (changes requested):
items for both tests:

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionCommentFilterTest.java
Rahulkhinchi03: Without a check, My exception isn't getting caught.

romani: ok, lets keep it.

src/test/resources/com/puppycrawl/tools/checkstyle/filters/suppressioncommentfilter/InputSuppressionCommentFilter10.java
Rahulkhinchi03: I removed all the files except `Filter config` as, without it, the exception is expected and the exception wasn't getting caught.

romani: please investigate a problem.
you use `execute(treeWalkerConfig` so it means that you get config created in test method (not a Input file).
Execute method does not parse Input file.

Rahulkhinchi03: I tried to remove all the comments from the filter config and found that:
```
/*
SuppressionCommentFilter
offCommentFormat = (default)CHECKSTYLE:OFF
onCommentFormat = (default)CHECKSTYLE:ON


*/
```

I removed other inputs and tests were passing but after removing  `offCommentFormat = (default)CHECKSTYLE:OFF` and  `onCommentFormat = (default)CHECKSTYLE:ON` the test gets fail and ask for exception is expected.

romani: I updated code to make it work, you need debug code.

romani (approved):

src/test/resources/com/puppycrawl/tools/checkstyle/filters/suppressioncommentfilter/InputSuppressionCommentFilter10.java
Rahulkhinchi03: Changes Done. I got it.

PR #11519, opened 2022-04-06
minor: changed the input file path in test file

Kevin222004:
Found at  #11446

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheckTest.java
@@ -517,7 +517,7 @@ public class ImportOrderCheckTest extends AbstractModuleTestSupport {
         checkConfig.addProperty("groups", "/^javax");
 
         try {
-            execute(checkConfig, getPath("InputImportOrder5.java"));
+            execute(checkConfig, getNonCompilablePath("InputImportOrder5.java"));
             assertWithMessage("exception expected").fail();
         }
         catch (CheckstyleException ex) {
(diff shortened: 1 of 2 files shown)


Kevin222004:
@Vyom-Yadav sorry for the delay

romani:
I do not understand, how it use to work, such mistake usually results in test failure

rnveach:
@romani Looking at exception message expected, it is processing a failure on the configuration given before even the file is processed. See `Cannot set property 'groups' to '/^javax'`. This test does not use BDD, so its not getting it's configuration from the input file, but from the java code.

If anything, this shows the input file is not even needed and could be swapped with just random text like "notaRealFile" to make it clear.

romani:
@Kevin222004, if this file is used only by one test method, please keep file but make it's class to be empty.

Kevin222004:
> @Kevin222004, if this file is used only by one test method, please keep file but make it's class to be empty.

The class is already an empty class 

> 
> https://github.com/checkstyle/checkstyle/blob/master/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/imports/importorder/InputImportOrder5.java
> 
> Should be no header comment, no imports.

if I remove imports and header comment it shows the error 
[ERROR] [checkstyle] [ERROR] /home/kevin/Desktop/C_Style/checkstyle/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/imports/importorder/InputImportOrder5.java:0: All check input files in the 'src/(it|test)/resources*/' folder must                    have trailing comments "// ok" and/or "// violation". [StyleValidationCommentInInputs]

romani:
Put on line with class "// ok" to bypass this violation

romani (approved):

Issue #11514, opened 2022-04-05
Improve code coverage for GUI classes


pbludov:
Current coverage:

| Class | Coverage |
| :------- | :-----------: |
| BaseCellEditor | 8% |
| TreeTable.TreeTableCellEditor | 8% |
| TreeTableCellRenderer | 35% |
| TreeTable | 89% | 
| ListToTreeSelectionModelWrapper | 63% |
| ParseTreeTableModel | 80% |
| TreeTableModelAdapter | 56% |
| CodeSelector | 80% | 
| TreeTableModelAdapter.UpdatingTreeModelListener | 60% |

The goal is 100% code coverage.

## Notes

It is possible to run GUI from IDE and collect the coverage report:
![image](https://user-images.githubusercontent.com/900805/161822092-6d68c940-c7b1-4dfe-9f9a-03e43c529b3a.png)

Some code may be unreachable from any user input (mouse, keyboard, changing system properties and so on).
Such code should be considered as "dead" code and should be deleted.

Before deleting any code, make sure that it is not executed in Windows, Mac & Linux.

Please send one PR per class. Use existing GUI tests from the package `com.puppycrawl.tools.checkstyle.gui` for reference.

romani:
@pbludov, we need to enforce current coverage number, to make sure CI will not allow going down.
Human can not track this.

romani:
One more improvement is merged

kpps17:
Hi [pbludov](https://github.com/pbludov), can you assign this to issue to me?
Thanks

romani:
We do not do assignments, just make a comment "I am on it" and start a work. We appreciate your help.

PR #11520, opened 2022-04-06
supplemental: enforce coverage ratio for gui tests to the current values #11514

pbludov:
supplemental PR for #11514 

Coverage percentage should be explicitly specified in the `pom.xml` file to prevent coverage decrease.


romani (approved):

pom.xml
```diff
@@ -875,15 +875,21 @@
                       com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocMethodCheck.RegularClass
                     </exclude>
                     <!-- Swing related classes -->
-                    <exclude>com.puppycrawl.tools.checkstyle.gui.BaseCellEditor*</exclude>
+                    <exclude>com.puppycrawl.tools.checkstyle.gui.BaseCellEditor</exclude>
```
nick-mancuso, line 878: Can we remove all classes in `gui`  from excludes in this PR?

nick-mancuso (changes requested):

pom.xml
rnveach: I am confused why these excludes are changing to only remove the `*` on the end? I believe these stars to are include any subclasses, so we are changing it to account for the subclasses but ignore the primary classes? Shouldn't these excludes be fully removed and we will keep the individual coverage on the bottom?

pbludov: That is exactly what I did. `TreeTable` and `TreeTable.TreeTableCellEditor` have their own coverage config now.
Same for `TreeTableModelAdapter` and `TreeTableModelAdapter.UpdatingTreeModelListener`.

All other classes do not have inner classes. For them, `*` has no effect.

nick-mancuso: @pbludov my mistake, I thought that once we defined coverage rules elsewhere, that we did not need thees excludes anymore. resolved.

nick-mancuso (approved):

rnveach (approved):
I don't see why we are removing the stars. Even if there are no subclasses, for them, they can remain until the entire exclude is removed.

romani:
If subclasses appears, jacoco will fail as far as I remember

Issue #11522, opened 2022-04-08
`checkstyle_checks.xml` should not allow single letter variable names in for loops


nrmancuso:
Noticed at https://github.com/checkstyle/checkstyle/pull/11371#pullrequestreview-932821582:

Right now, there is 85 errors in `master` branch that are suppressed, We need help to remove all suppresions and update names of variables. Examples of updates:
https://github.com/checkstyle/checkstyle/commit/2bf4bf71e2244e0c26d5f969cc2f384c170e6f41

Already DONE:
https://github.com/checkstyle/checkstyle/blob/c1c41d49ca6829df3c0dcf3e8ca980838c8ef0bf/config/checkstyle_checks.xml#L758-L761

Right now, we have `allowOneCharVarInForLoop` property set to `true`, which allows for single letter variable names. For readability, we should enforce the same naming conventions on for loop variables that we do on other local variables. Instead of `i`, let's use `index`.

romani:
Example of violation see in CI failures https://github.com/checkstyle/checkstyle/pull/11521#pullrequestreview-935864566

Rahulkhinchi03:
@nick-mancuso @romani 
We juyst need to set `allowOneCharVarInForLoop` to `false`  and change `i`  to `index` right?

rnveach:
> set allowOneCharVarInForLoop to false 

There is already a PR to do this.
https://github.com/checkstyle/checkstyle/pull/11524/commits

> change i to index 

Violations need to be fixed. If name is `i` and is used as an index, then that change would be valid. This fix depends on the context of the code.
Suppressions in the config attached to this issue need to be removed to show the issue is fixed.

Rahulkhinchi03:
Okay, thanks for the information!

strkkk:
Fix is merged

romani:
https://github.com/checkstyle/checkstyle/blob/ef4b93593a565db16254160b74ad1d5ad52058e0/config/checkstyle-checks.xml#L924-L926

done long time ago

romani:
We need to remove reference to this issue at config/suppressions-xpath.xml

arturict:
I am on it.

romani:
Fix is merged

PR #11524, opened 2022-04-08
Issue #11522: Fix LocalVariableName violation in AbstractItModuleTestSupport.java

nick-mancuso:
First commit has all suppressions and modifies check module.
Second commit is example of fix for one check.

Once PR approved/ merged, we can add a link to second commit as an example in associated issue.


romani (approved):
Ok to merge.
Generation of xpath might be more complicated than actual fix ;)

nick-mancuso:
> Ok to merge.
> Generation of xpath might be more complicated than actual fix ;)

I just wrote a checkstyle script with `-g` option, and regexp to simplify paths, it went much faster than fixing all these :)

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
```diff
@@ -247,10 +247,10 @@ protected final void verify(Checker checker,
             LineNumberReader lnr = new LineNumberReader(
                 new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
             int previousLineNumber = 0;
-            for (int i = 0; i < expected.length; i++) {
-                final String expectedResult = messageFileName + ":" + expected[i];
+            for (int index = 0; index < expected.length; index++) {
+                final String expectedResult = messageFileName + ":" + expected[index];
                 final String actual = lnr.readLine();
-                assertWithMessage("error message %s", i)
+                assertWithMessage("error message %s", index)
```
strkkk, line 253: that seems weird. `index` is index, so error message will be `error message 0` ?

I guess there is no need to pass index here and format string.

By the way, loop can be refactored to for-each loop

strkkk (changes requested):

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
nick-mancuso: > that seems weird. index is index, so error message will be error message 0 ?

I agree, many error messages in this class are strange/ hard to understand. Since this is an "example" can we keep this commit simple, and create an issue to improve error messages in test support?

romani: make sense to keep as is just for students

romani: lets agree to fix it in separate PR,  we have single case like this
```
14:36 $ ag -Q "assertWithMessage(\"error message %s\", i)"
src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java

```

nick-mancuso: I did separate commit, but I think that it actually makes sense to keep regular for loop and similar (but improved) message. Example of new message:

```
error message 57 does not match expected message
expected: …galTokenText.java:161:29: Consider using …
but was : …galTokenText.java:160:29: Consider using …

```

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
```diff
@@ -250,7 +250,7 @@ protected final void verify(Checker checker,
             for (int index = 0; index < expected.length; index++) {
                 final String expectedResult = messageFileName + ":" + expected[index];
                 final String actual = lnr.readLine();
-                assertWithMessage("error message %s", index)
+                assertWithMessage("error message %s does not match expected message", index)
```
romani, line 253: Error Message at position %s of expected, does .....

nick-mancuso: done.

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
```diff
@@ -247,10 +247,11 @@ protected final void verify(Checker checker,
...
                 new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
             int previousLineNumber = 0;
-            for (int i = 0; i < expected.length; i++) {
-                final String expectedResult = messageFileName + ":" + expected[i];
+            for (int index = 0; index < expected.length; index++) {
+                final String expectedResult = messageFileName + ":" + expected[index];
                 final String actual = lnr.readLine();
-                assertWithMessage("error message %s", i)
+                assertWithMessage("Error message at position %s of 'expected' does "
+                        + "not match expected message", index)
```
romani, line 254: not match actual message

:)

nick-mancuso: Long week :), done.

romani (approved):

romani:
it is good point to understand .... as it was "i" all of us was blind, and all our brains refused to read this message, as it become word, brain started to read and nonsense become detected.

strkkk (approved):

PR #11526, opened 2022-04-09
minor: replace Collections.unmodifiable* with static factory methods

strkkk:
1. Replace some usages, where static factory method can be used
2. Simplify some places, where new java11 Collectors methods can be used

Note: some static fields sets where mutable, but now they are immutable.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java
@@ -20,15 +20,11 @@
 package com.puppycrawl.tools.checkstyle.checks.javadoc;
 
 import java.util.ArrayDeque;
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.Deque;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
-import java.util.TreeSet;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
 import com.puppycrawl.tools.checkstyle.JavadocDetailNodeParser;
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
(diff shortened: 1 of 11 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java
```diff
@@ -327,24 +327,24 @@
     public static final String MSG_EXTRA_HTML = "javadoc.extraHtml";
 
     /** HTML tags that do not require a close tag. */
-    private static final Set<String> SINGLE_TAGS = Collections.unmodifiableSortedSet(
-        Arrays.stream(new String[] {"br", "li", "dt", "dd", "hr", "img", "p", "td", "tr", "th", })
-            .collect(Collectors.toCollection(TreeSet::new)));
+    private static final Set<String> SINGLE_TAGS = new TreeSet<>(Set.of(
```
pbludov, line 330: This will create an ordinal (mutable) `TreeSet` and simply call `addAll` method with the inner set items.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java
```diff
@@ -327,24 +327,24 @@
...
+    );
 
     /**
      * HTML tags that are allowed in java docs.
      * From https://www.w3schools.com/tags/default.asp
      * The forms and structure tags are not allowed
      */
     private static final Set<String> ALLOWED_TAGS = Collections.unmodifiableSortedSet(
-        Arrays.stream(new String[] {
+        Stream.of(
```
pbludov, line 340: This can be simple `Set.of`

pbludov (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java
strkkk: Thanks for pointing this.
Actually, the only usage of this set is to invoke `contains` method, so there is no need for TreeSet. Changed to usual `Set.of`

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java
strkkk: yeap, and there is no need for sorted set anyway. Same as above

romani (approved):

pbludov (approved):

PR #11530, opened 2022-04-09
minor: move xml-maven-plugin to config directory outside of ci

rnveach:
Identified at backport when removing the `.ci` folder since it doesn't make use of CI. It was the only thing that couldn't be removed, and it doesn't appear to have a connection to the CI. This file is similar `catalog.xml` which also lives in the config folder to allow offline usage.
https://github.com/rnveach/checkstyle-backport-jre8/tree/master/.ci


rnveach:
pitest failure can be ignored, it is unrelated.

romani (approved):
Please make prefix of commit to reference PR, reason in description is very valuable.

rnveach:
Done.

romani (approved):

Issue #5337, opened 2017-12-09
Deprecate com.puppycrawl.tools.checkstyle.api.AutomaticBean


rnveach:
Discussed at https://github.com/checkstyle/checkstyle/pull/5322#discussion_r155799488 ,

`AutomaticBean` should be renamed to `AbstractAutomaicBean` since it is now abstract.

Additionally we should discuss if we should keep `AutomaticBean` in name.

SinmoWay:
I propose to stop on the name `AbstractAutomatinBean` and do not change anything else.
However, there is an option for the separation of the implementation `AbstractAutomationBean` and `AbstractAutomatic` but I have not yet thought it over well enough.

rnveach:
This change will break maven checkstyle: https://github.com/apache/maven-checkstyle-plugin/blob/ab6a941b61e49fe114c00c3ce412d6b985c95df7/src/main/java/org/apache/maven/plugins/checkstyle/exec/CheckstyleCheckerListener.java#L42
eclipse-cs seems fine.

We will have to create new class with old name and rename old class to new name.

Rahulkhinchi03:
@rnveach is this issue still valid?

rnveach:
Nothing has been said that would say otherwise.

Rahulkhinchi03:
I would like to try this issue. Can you please help in understanding how to mark this API as deprecated? Just a @deprecated works?

romani:
take a look at jacoco config in pom.xml, we have ability to exclude class from code coverage, we do not need to test something that use to work well and now become deprecated.

rnveach:
@Rahulkhinchi03 This issue should be very similar to https://github.com/checkstyle/checkstyle/issues/2551 . Both renamed classes to others and marked them as deprecated and were eventually removed.

stoyanK7:
On it.

romani:
@rnveach , @nrmancuso , I think we need to change label of this issue to bug or misc.

nrmancuso:
Since we only deprecate this class in scope of this issue, and do full removal in https://github.com/checkstyle/checkstyle/issues/12873, this makes sense.

romani:
bug or misc ?

PR #11535, opened 2022-04-10
Issue #5337: Rename AutomaticBean

Rahulkhinchi03:
Resolves Issue #5337

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
@@ -40,7 +40,7 @@ import org.apache.commons.logging.LogFactory;
 
 import com.puppycrawl.tools.checkstyle.api.AuditEvent;
 import com.puppycrawl.tools.checkstyle.api.AuditListener;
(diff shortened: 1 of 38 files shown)


Rahulkhinchi03:
@romani @rnveach @nick-mancuso @strkkk
I added AutomaticBeanTest in excludedTestClasses so that It doesn't affect the code coverage but it still does.

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -465,8 +465,8 @@ protected void setupChild(Configuration childConf)
         try {
             child = moduleFactory.createModule(name);
 
-            if (child instanceof AutomaticBean) {
-                final AutomaticBean bean = (AutomaticBean) child;
+            if (child instanceof AbstractAutomaticBean) {
+                final AbstractAutomaticBean bean = (AbstractAutomaticBean) child;
```
romani, line 469: We can not change all existing numerous third-party Checks.
It will take time before they migrate, we need to support both classes during load of Checks.

romani (changes requested):
Items:

Issue #11427, opened 2022-03-18
Tracker for pitest-coding failures


nick-mancuso:
Since we have been seeing random (but regular) failures in `pitest-coding`, I thought it would be a good idea to track exactly which lines are causing pitest to fail. Maintainers, feel free to just add any failures that you notice below (in issue description).

- [x] `EqualsAvoidNullCheck.java.html:<td class='covered'><pre><span  class='survived'>            if (field.findFirstToken(TokenTypes.IDENT) != null) {</span></pre></td></tr>`

  https://nick-mancuso.github.io/pit-reports-equal/202203230810/com.puppycrawl.tools.checkstyle.checks.coding/EqualsAvoidNullCheck.java.html

  Negated conditional

- [x] `RequireThisCheck.java.html:<td class='covered'><pre><span  class='survived'>            if (returnedVariable) {</span></pre></td></tr>`

  Report at https://nick-mancuso.github.io/pit-reports-requirethis/202203211238/com.puppycrawl.tools.checkstyle.checks.coding/RequireThisCheck.java.html

  Removed  conditional

- [x] `CheckstyleAntTask.java.html:<td class='covered'><pre><span  class='survived'>            if (toFile == null || !useFile) {</span></pre></td></tr>`

  https://pbludov.github.io/pitest-ant-coverage-report/pit-reports/202203270532/com.puppycrawl.tools.checkstyle.ant/CheckstyleAntTask.java.html#org.pitest.mutationtest.report.html.SourceFile@352b92be_720

  removed conditional - replaced equality check with true → SURVIVED

Hopefully we can diagnose issue and make `pitest-coding` more stable.

rnveach:
It may be helpful to know which mutator is the issue as we have added multiple and one line can have multiple mutators involved. In the current example we could be mutating `findFirstToken` or we could be mutating the entire condition to true/false. We may have to look into how we can add this to the report.

nick-mancuso:
In the case of `RequireThisCheck`:

Passing `pitest-coding`:
```
org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_ELSE
>> Generated 904 Killed 894 (99%)
> KILLED 881 SURVIVED 10 TIMED_OUT 13 NON_VIABLE 0 
> MEMORY_ERROR 0 NOT_STARTED 0 STARTED 0 RUN_ERROR 0 
> NO_COVERAGE 0 

```

Failing `pitest-coding`:
```
[...]

PR #11543, opened 2022-04-12
Issue #11527: Extend TokenUtils.isOfType with more overloaded method inprove performance

### LINKED ISSUE #11527: Extend TokenUtils.isOfType with more overloaded method inprove performance
detected at https://github.com/checkstyle/checkstyle/pull/11371#discussion_r830507927

apply patch:
```
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java b/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
index 9dd5c13..a5ff490 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
@@ -291,7 +291,23 @@ public final class TokenUtil {
         return matching;
     }
 
-    /**
+    public static boolean isOfType(int type, int first) {
+        return type == first;
+    }
+
+    public static boolean isOfType(int type, int first, int second) {
+        return type == first || type == second;
+    }
+
+    public static boolean isOfType(int type, int first, int second, int third) {
+        return type == first || type == second || type == third;
+    }
+
+    public static boolean isOfType(int type, int first, int second, int third, int forth) {
+        return type == first || type == second || type == third || type == forth;
+    }
+
+   /**
      * Determines if the AST belongs to the given types.
      *
      * @param ast the AST node to check
@@ -303,6 +319,22 @@ public final class TokenUtil {
         return ast != null && isOfType(ast.getType(), types);
     }
 
+    public static boolean isOfType(DetailAST ast, int first) {
+        return ast != null && isOfType(ast.getType(), first);
+    }
+
+    public static boolean isOfType(DetailAST ast, int first, int second) {
+        return ast != null && isOfType(ast.getType(), first, second);
+    }
+
+    public static boolean isOfType(DetailAST ast, int first, int second, int third) {
+        return ast != null && isOfType(ast.getType(), first,second, third);
+    }
+
+    public static boolean isOfType(DetailAST ast, int first, int second, int third, int forth) {
+        return ast != null && isOfType(ast.getType(), first,second, third, forth);
+    }
+
     /**
      * Determines if given AST is a root node, i.e. if the type
      * of the token we are checking is {@code COMPILATION_UNIT}.
```

fix coverage:
image

Rahulkhinchi03:
Resolves #11527


Rahulkhinchi03:
dd

While creating UTs, I found that this was already `red`.

@romani @nick-mancuso @strkkk @rnveach I am looking for your guidance to make this green.

baratali:
Please add more unit tests to increase code coverage.
Also, please fix spelling mistake in `matc`.
CI errors:
```
[WARNING] Rule violated for class com.puppycrawl.tools.checkstyle.utils.TokenUtil: branches covered ratio is 0.98, but expected minimum is 1.00
```
```
New misspellings found, please review:
matc
```

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -303,6 +303,57 @@ public static boolean isOfType(DetailAST ast, int... types) {
...
 
+    /**
+     * Determines if the AST belongs to the given first parameter.
+     *
+     * @param ast the AST node to check
+     * @param first the acceptable token
+     *
+     * @return true if type matches one of the given parameter.
+     */
+    public static boolean isOfType(DetailAST ast, int first) {
```
baratali, line 314: you should add following methods as well:
```
+    public static boolean isOfType(int type, int first) {
+        return type == first;
+    }
+
+    public static boolean isOfType(int type, int first, int second) {
+        return type == first || type == second;
+    }
+
+    public static boolean isOfType(int type, int first, int second, int third) {
+        return type == first || type == second || type == third;
+    }
+
+    public static boolean isOfType(int type, int first, int second, int third, int forth) {
+        return type == first || type == second || type == third || type == forth;
+    }
```

baratali (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
Rahulkhinchi03: Changes Done. Fix Misspelling `"matc"`.

Rahulkhinchi03:
@baratali I need your guidance to fix the code coverage issue. In my Local, All methods are green.

baratali:
Hey. If you use Intellij IDEA, then it doesn't show branch coverage by default. You need to configure the run and add "Use traces":
2022-04-13 21_49_26-checkstyle – TokenUtil java
Only one condition is missing to cover in `isOfType(DetailAST ast, int... types)` :)
(this exact screenshot is from jacoco.exec report imported to IDEA, but the results are the same if you run it directly in IDEA)
2022-04-13 21_11_33-checkstyle – TokenUtil java

nick-mancuso:
> @baratali @romani @strkkk @rnveach @pbludov @nick-mancuso I need your guidance to fix the code coverage issue. In my Local, All methods are green. 

Please generate report as described in the wiki https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations

Rahulkhinchi03:
@baratali I have tried true and false together but didn't cover the branch. Can you please help me in finding the condition?

baratali:
@Rahulkhinchi03 , please be aware that the full branch coverage means all combinations of that condition:
```
true && true
true && false
false && any
```
IDEA and jacoco doesn't show me what branch extacly is missing. Please try to build coverage report using https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations#how-to-generate-ut-coverage-report. I will try it later too, I used different command yesterday.

Rahulkhinchi03:
@nick-mancuso  @baratali @romani 
I have seen this many times in my PRs that `pitest-coding` gets failed and the error says that `target.txt` differs. Why does this happen? What is the reason behind this?

nick-mancuso:
@Rahulkhinchi03 https://github.com/checkstyle/checkstyle/issues/11427

It is not related to your changes, and we have still not determined what is causing these intermittent failures.

Rahulkhinchi03:
@baratali @romani Please review the changes and let me know if any changes are required.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -291,6 +291,60 @@ public static boolean isOfType(int type, int... types) {
...
+    }
+
+    /**
+     * Determines if the token type belongs to the given first or second parameter.
+     *
+     * @param type the Token Type to check
+     * @param first the acceptable token
+     * @param second the acceptable token
+     *
+     * @return type if it is equals to first or second
```
baratali, line 313: `type if it is equals to first or second`
--->
`true if type equals to first or second`
Please fix `@return` description in other methods too.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -291,6 +291,60 @@ public static boolean isOfType(int type, int... types) {
...
+     *
+     * @param type the Token Type to check
+     * @param first the acceptable token
+     * @param second the acceptable token
+     * @param third the acceptable token
+     * @param forth the acceptable token
+     *
+     * @return type if it is equals to first or second or third or forth parameter.
+     */
+    public static boolean isOfType(int type, int first, int second, int third, int forth) {
```
baratali, line 344: `forth` --> `fourth`

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -303,6 +357,57 @@ public static boolean isOfType(DetailAST ast, int... types) {
...
+        return ast != null && isOfType(ast.getType(), first);
+    }
+
+    /**
+     * Determines if the AST belongs to the given first and second parameters.
+     *
+     * @param ast the AST node to check
+     * @param first the acceptable token
+     * @param second the acceptable token
+     * @return true if type matches one of the given parameter.
```
baratali, line 378: `true if type matches one of the given parameter.`
-->
`true if type matches one of the given types.`

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -303,6 +357,57 @@ public static boolean isOfType(DetailAST ast, int... types) {
...
+     * Determines if the AST belongs to the given first, second, third and forth parameters.
+     *
+     * @param ast the AST node to check
+     * @param first the acceptable token
+     * @param second the acceptable token
+     * @param third the acceptable token
+     * @param forth the acceptable token
+     * @return true if type matches one of the given parameter.
+     */
+    public static boolean isOfType(DetailAST ast, int first, int second, int third, int forth) {
```
baratali, line 407: `forth` --> `fourth`

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -395,4 +395,212 @@ public void testIsBooleanLiteralType() {
...
     }
 
+    @Test
+    public void testAreOnSameLine() {
+        final int type = TokenTypes.COMPILATION_UNIT;
+        final DetailAstImpl astForTest = new DetailAstImpl();
+        final DetailAstImpl astForTest2 = new DetailAstImpl();
+        astForTest.setType(type);
+        astForTest2.setType(type);
+        final boolean result = TokenUtil.areOnSameLine(astForTest, astForTest2);
```
baratali, line 405: we should set a line number here instead of token type, because `TokenUtil.areOnSameLine()` works with ast line number.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -395,4 +395,212 @@ public void testIsBooleanLiteralType() {
...
+        assertWithMessage("Token type did not match")
+            .that(result1)
+            .isTrue();
+        assertWithMessage("Token type did not match")
+            .that(result2)
+            .isTrue();
+    }
+
+    @Test
+    public void testIsOfTypeThreeFalse() {
```
baratali, line 569: incorrect name, you're testing `isOfType` with four type arguments here

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -395,4 +395,212 @@ public void testIsBooleanLiteralType() {
...
+        assertWithMessage("Token type should not match")
+            .that(result2)
+            .isFalse();
+        assertWithMessage("Token type should not match")
+            .that(result3)
+            .isFalse();
+    }
+
+    @Test
+    public void testIsOfTypeThreeTrue() {
```
baratali, line 549: incorrect name, you're testing `isOfType` with four type arguments here

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -395,4 +395,212 @@ public void testIsBooleanLiteralType() {
...
+            .that(result1)
+            .isFalse();
+        assertWithMessage("Token type should not match")
+            .that(result2)
+            .isFalse();
+        assertWithMessage("Token type should not match")
+            .that(result3)
+            .isFalse();
+    }
+
```
baratali, line 547: it seems the existing tests `testIsOfTypeTrue()` and `testIsOfTypeFalse()` test `isOfType()` with three type arguments now. Can you please rename them to `testIsOfTypeThreeTrue()` and `testIsOfTypeThreeFalse()` to match the new naming system and move it below `testIsOfTypeSecondFalse()` for better readability?

baratali (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -395,4 +395,212 @@ public void testIsBooleanLiteralType() {
...
+        assertWithMessage("Token type did not match")
+            .that(result)
+            .isTrue();
+    }
+
+    @Test
+    public void testIsOfTypeTypesTrue() {
+        final int type = TokenTypes.LITERAL_CATCH;
+        final DetailAstImpl astForTest = new DetailAstImpl();
```
nick-mancuso, line 415: @Rahulkhinchi03 we try and avoid pure unit tests like this at all costs. Right now, these are just obfuscating dead code, and adding to technical debt. Before we would resort to pure UT, we need an investigation and explanation of why we can't hit a given condition with actual test cases in check tests.

nick-mancuso (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
romani: All tests should be by inputs, pure Junit for method is absolute last resort with details in comment why it is not possible.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
Rahulkhinchi03: Changes Done

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
Rahulkhinchi03: Changes Done

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
Rahulkhinchi03: Changes Done

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
Rahulkhinchi03: Changes Done

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
Rahulkhinchi03: Changes Done

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
Rahulkhinchi03: Changes Done

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
Rahulkhinchi03: Changes Done

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
Rahulkhinchi03: Changes Done

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
Rahulkhinchi03: @nick-mancuso @romani
I believe that we do not need to create an input file for this. Instead, we can just replace Check Method with the `TokenUtil.java` method.

baratali: `isOfType()` is overloaded, so EmptyLineSeperatorCheck will use the new method you created for 4 type arguments. No need to rename them.
New inputs are needed just to cover the new methods which are not currently covered. So you should create a test with an input for a check (for example, EmptyLineSeperatorCheck) in order for it to call `isOfType()` method instead of creating tests directly in TokenUtilTest.

romani (changes requested):
All unit tests that use creation of DetailAstImp must be removed, test coverage should be done by Input files.

Rahulkhinchi03:
@romani I tried to cover code coverage with the input file. All inputs are covered except for 3 methods:

1. (DetailAst, first, second, third)
2. (DetailAst, first, second)
3. (DetailAst, first, second, third, fourth)

All checks use (ast != null) which makes it not possible to hit the second condition of these methods: 
return `ast != null` && isOfType(ast.getType(), first, second);


I believe that we should use DetailASTImp and create UTs in `TokenUtilTest.java`

Rahulkhinchi03:
@romani We cannot hit `(ast != null) == false` in (DetailAst, first, second, third, fourth) so DetailAst can never be null and we cant hit a check. So we need to create the test in TokenUtilTests.java

Please let me know if we have any other options. 

I am figuring out different methods for now like:

- (DetailAst, first, second, third)
- (DetailAst, first, second)
- (DetailAst, ...types)

romani:
We do not need check for null in this case, remove it.
You can run  regression tests to see in what cases it might be null.

Ideally util methods should be null tolerant. I might agree on pure Unit  test if we can prove that we can not hit it by inputs.

Rahulkhinchi03:
public static boolean isOfType(DetailAST ast, int first, int second,
                                         int third, int fourth) {
        return `ast != null`  && isOfType(ast.getType(), first, second, third, fourth);
    }

The highlighted condition must be false to achieve 100% code coverage. Only 1 out of 4 branches is missing and it is just because of `ast != null`. We cant use this in check as this method is using a check which doesn't allow `(ast!=null) == false`


@romani  I believe that I should create a single test in `TokenUtilTest.java`  with a JavaDoc explaining why it is being used here and not in check?

baratali:
> We do not need check for null in this case, remove it.

Some Checks pass `null` to `isOfType(ast, first)`, so I would keep "not-null check" in all `isOfType()` methods for consistensy.
Here is code coverage when I remove all tests from TokenUtilTest:


----

> We cant use this in check to cover the IsOftype() method it is using a check which doesn't allow (ast!=null) == false

Confirming this too. `isOfType(DetailAST ast, int first, int second, int third, int forth)` is used in two places and both of them can't pass `null` as the first argument `ast`:

- `ScopeUtil` - it checks ast != null before passing it to isOfType()
- `EmptyLineSeparatorCheck` - ast is argument from visitToken(ast) which can't be null.

I think we should create a pure unit test in TokenUtilTest for `isOfType(DetailAST ast, int first, int second, int third, int forth)` to cover it.

----

> I am figuring out different methods for now like:
> (DetailAst, first, second, third)
> (DetailAst, first, second)
> (DetailAst, ...types)

@Rahulkhinchi03 , if you're struggling to find a proper Check and input code to cover these methods, you can try to remove `ast != null` from `isOfType()`, run a regression test and wait for NullPointerException (it should print what java file caused it). Please read about regression testing in https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#report-generation

romani:
>no-exception testing / no-exception-openjdk17

Look inside it might answer some of your questions

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -300,7 +300,114 @@ public static boolean isOfType(int type, int... types) {
...
+     * Determines if the AST belongs to the given first parameter.
+     *
+     * @param ast the AST node to check
+     * @param first the acceptable token
+     *
+     * @return true if type matches one of the given types.
+     */
+    public static boolean isOfType(DetailAST ast, int first) {
+        return isOfType(ast.getType(), first);
+    }
```
baratali, line 328: Please don't remove null check in this method. This condition is already covered and used, see https://github.com/checkstyle/checkstyle/pull/11543#issuecomment-1101406256. So UTs and the report generation will fail on this line.

baratali (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
Rahulkhinchi03: I mistakenly did that. Doing the changes.

Rahulkhinchi03:
GitHub, generate report

Rahulkhinchi03:
GitHub, generate report

baratali:
Based on the exception from github action logs (https://github.com/checkstyle/checkstyle/runs/6078583297?check_suite_focus=true) there is a problem in ImportControlCheck. If we look at that line in java code, we can see it tries to access "file" parameter:
```java
@Override
public Set<String> getExternalResourceLocations() {
  return Collections.singleton(file.toString());
}
```
I can see you deleted "file" parameter from ImportControlCheck in your config file: https://gist.githubusercontent.com/Rahulkhinchi03/84d8508f2db07e445e60a4d9592541eb/raw/4abd5e6ffc3073c8c398c81924e60ec12ebd587a/config.xml.That's why it fails. You should delete `<ImportControl>` module entirely from the config file, it doesn't use isOfType() method. You can remove any module/check from the config if it doesn't use isOfType() method.

Rahulkhinchi03:
GitHub, generate report

Rahulkhinchi03:
GitHub, generate report

Rahulkhinchi03:
GitHub, generate report

Rahulkhinchi03:
GitHub, generate report

Rahulkhinchi03:
GitHub, generate report

nick-mancuso:
@Rahulkhinchi03 please just use https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/checks-nonjavadoc-error.xml , there is no need for patch config

Rahulkhinchi03:
@nick-mancuso I have to use it as
Report Label:  https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/checks-nonjavadoc-error.xml?

nick-mancuso:
@Rahulkhinchi03 carefully read through https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#report-generation to learn about report generation.

Rahulkhinchi03:
GitHub, generate report

Rahulkhinchi03:
GitHub, generate report

nick-mancuso:
@Rahulkhinchi03 make sure that you add modules 
https://github.com/checkstyle/contribution/blob/7fbc89083b260c8b8f0cd1f550a6d958fe5ac00b/checkstyle-tester/my_check.xml#L10 lines 10 - 18

Rahulkhinchi03:
@nick-mancuso What should be the value of severity? It's **`ignore`** in the checkstyle-nonjavadoc file and it's **`warning`** in mycheck.xml

nick-mancuso:
```
➜  checkstyle git:(master) grep -lR "TokenUtil.isOfType(" | grep "Check" | grep -v Util
src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/AbstractClassCouplingCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMissingLeadingAsteriskCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceBeforeCaseDefaultColonCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ExecutableStatementCountCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/NoEnumTrailingCommaCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ParameterAssignmentCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```

Above is a list of all checks that use this method. Add default modules for them to mycheck.xml, which you can copy/paste from https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/checks-nonjavadoc-error.xml.  Note that `AbstractClassCouplingCheck.java` is an abstract class, so include all checks that extend this class as well.

Rahulkhinchi03:
I added all the checks which use the IsOfType() method except the first and (first, second, third, fourth) parameters as these are already covered.

Rahulkhinchi03:
GitHub, generate report

Rahulkhinchi03:
@romani @baratali @nick-mancuso 
Please check out my Report: https://rahulkhinchi03.github.io/

The report is successfully created and everything was processed with no exceptions. 

DetailsAst in `isOfType()` methods is always `non-null`  as there is no such java code that can cause `NullPointerException` which helps us to come to a conclusion that we should continue with pure unit tests and create UTs for isotype() method in `TokenUtilTest.java`

Rahulkhinchi03:
GitHub, rebase

Rahulkhinchi03:
Please check out the report now: https://rahulkhinchi03.github.io/
Please view my comment: https://github.com/checkstyle/checkstyle/pull/11543#issuecomment-1105316549

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -300,7 +300,114 @@ public static boolean isOfType(int type, int... types) {
...
+    /**
+     * Determines if the token type belongs to the given first parameter.
+     *
+     * @param type the Token Type to check
+     * @param first the acceptable token
+     *
+     * @return true if type equals to first
+     */
+    public static boolean isOfType(int type, int first) {
+        return type == first;
```
romani, line 315: Why we need this method?
Remove it

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -300,7 +300,114 @@ public static boolean isOfType(int type, int... types) {
...
+    }
+
+    /**
+     * Determines if the AST belongs to the given first parameter.
+     *
+     * @param ast the AST node to check
+     * @param first the acceptable token
```
pbludov, line 322: `first` -> `type`

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
Rahulkhinchi03: I forgot to remove it during generating the report. Changes Done

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
Rahulkhinchi03: Changes Done

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -341,58 +341,140 @@ public void testIsTypeDeclaration() {
...
+
+    /**
+     * IsOfType{} method with 2 parameter is a general test and not used
+     * in Check UTs as we can't hit all conditions to achieve 100%
+     * Code Coverage.
+     */
+    @Test
+    public void testIsOfTypeSecond() {
+        final int type = TokenTypes.LITERAL_CATCH;
+        final DetailAstImpl astForTest = new DetailAstImpl();
```
romani, line 390: please remove this method and all other methods that do "new DetailAstImpl" that you introducing.
All coverage should be done by Input files and whole Checstyle execution and it is done for Checks. 

You need to extend test of Checks that are using such methods.

romani (changes requested):
items

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
baratali: @romani, why should we remove this method? If we remove it, Checks will keep using 'unoptimized' `isOfType(int type, int... types)` instead, which we try to avoid in this fix.
Currently this new method `isOfType(int type, int first)` is used in UnnecessaryParenthesesCheck.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -341,58 +341,140 @@ public void testIsTypeDeclaration() {
...
+        final DetailAstImpl astForTest2 = new DetailAstImpl();
+        astForTest.setLineNo(type);
+        astForTest2.setLineNo(type);
+        final boolean result = TokenUtil.areOnSameLine(astForTest, astForTest2);
+
+        assertWithMessage("Token type did not match")
+            .that(result)
+            .isTrue();
+    }
```
baratali, line 368: please remove unit tests which are not related to isOfType(). `TokenUtil.areOnSameLine` is already covered by other Check UTs.

baratali (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -341,58 +341,140 @@ public void testIsTypeDeclaration() {
...
+    public void testIsRootNode() {
+        final int type = TokenTypes.COMPILATION_UNIT;
+        final DetailAstImpl astForTest = new DetailAstImpl();
+        astForTest.setType(type);
+        final boolean result = TokenUtil.isRootNode(astForTest);
+
+        assertWithMessage("Token type did not match")
+            .that(result)
+            .isTrue();
+    }
```
baratali, line 380: same as https://github.com/checkstyle/checkstyle/pull/11543#discussion_r857111946

baratali (changes requested):

Rahulkhinchi03:
> Please check out the report now: https://rahulkhinchi03.github.io/
> Please view my comment: https://github.com/checkstyle/checkstyle/pull/11543#issuecomment-1105316549

@romani please check my generated report and comment to see my proposal for this issue.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
Rahulkhinchi03: @romani please check my generated report and comment to see my proposal for this issue.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
Rahulkhinchi03: Changes Done

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
Rahulkhinchi03: Changes Done

baratali:
@Rahulkhinchi03 , you deleted some `ast != null` conditions and added pure unit tests for these methods. This doesn't make sense because the only reason we need these pure unit tests is to cover `ast != null` conditions because it's not possible to cover them via Check tests.
@romani , we need your help to decide whether we should keep `ast != null` in `isOfType()` methods.
I think we should keep all `ast != null` conditions, please see my comment https://github.com/checkstyle/checkstyle/pull/11543#issuecomment-1101406256.

baratali:
@Rahulkhinchi03 , you deleted some `ast != null` conditions and added pure unit tests for these methods. This doesn't make sense because the only reason we need these pure unit tests is to cover `ast != null` conditions because it's not possible to cover them via Check tests.
@romani , we need your help to decide whether we should keep `ast != null` in `isOfType()` methods.
I think we should keep all `ast != null` conditions, please see my comment https://github.com/checkstyle/checkstyle/pull/11543#issuecomment-1101406256.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
rnveach: @romani We should think of adding https://checkstyle.org/config_coding.html#IllegalInstantiation for this .

romani:
>public static boolean isOfType(DetailAST ast, int first) {

We can remove this method, and inline  comparison of type in Check.
If we want to keep this method, we can put check for null in Checks to let them handle nulls. As we already have coverage for this, it shot not be a problem.

Rahulkhinchi03:
@romani  please check my generated report and comment to see my proposal for this issue.
Here: https://github.com/checkstyle/checkstyle/pull/11543#issuecomment-1105316549

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
romani: Ok, let's keep it, but no coverage by pure Unit tests for method.

Rahulkhinchi03: @romani except this, We have to do coverage with pure tests only.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
Rahulkhinchi03: @romani 
Here: https://github.com/checkstyle/checkstyle/pull/11543#issuecomment-1105316549

baratali:
> public static boolean isOfType(DetailAST ast, int first) {
> We can remove this method, and inline comparison of type in Check.
> If we want to keep this method, we can put check for null in Checks to let them handle nulls. As we already have coverage for this, it shot not be a problem.

If we remove `isOfType(DetailAST ast, int first)` method, then I think it will be confusing for Check developers because it would still be possible to execute `isOfType(ast, int)` but it would lead to its unoptimized version `isOfType(ast, int...)`. Therefore, the second option sounds better - I would keep `isOfType(DetailAST ast, int first)` method and move the null check from this method to Check's code (where it's needed).

@Rahulkhinchi03 , could you please prepare such fix?
 - delete null check from all `isOfType()` methods.
 - find code which passes null to `isOfType(DetailAST ast, int first)` and add null check there. Currently I can see 6 places which use this method, but probably not all of them pass null. So this step will probably require some testing and generating diff reports.
 - delete all pure unit tests for `isOfType()` from `TokenUtilTest`, even the existing ones.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -302,7 +302,114 @@ public static boolean isOfType(int type, int... types) {
...
+     * Determines if the token type belongs to one of the given types.
+     *
+     * @param ast the AST node to check
+     * @param first the acceptable token
+     * @param second the acceptable token
+     * @param third the acceptable token
+     * @param fourth the acceptable token
+     * @return true if type matches one of the given types.
+     */
+    public static boolean isOfType(DetailAST ast, int first, int second,
```
romani, line 394: please group all `isOfType(DetailAST ast, .....` 
and then all `boolean isOfType(int type,....`

romani (changes requested):
item:

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -339,60 +339,4 @@ public void testIsTypeDeclaration() {
...
-        assertWithMessage("Token type should not match")
-                .that(result2)
-                .isFalse();
-        assertWithMessage("Token type should not match")
-                .that(result3)
-                .isFalse();
-    }
-
-    @Test
-    public void testIsBooleanLiteralType() {
```
baratali, line 386: this test is not related to isOfType(), you should not remove it

baratali (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
Rahulkhinchi03: Changes Done

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
Rahulkhinchi03: Changes Done

Rahulkhinchi03: I am currently investigating what @baratali said here: https://github.com/checkstyle/checkstyle/pull/11543#issuecomment-1120026036 
Please give me some time to work on this issue.

Rahulkhinchi03: Report with latest commit: https://rahulkhinchi03.github.io/
(for isOfType() with first parameter)

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
romani: Not done, I still see removal.
It might be reason of CI failures.

It is ok to remove only if does not affect CI

romani:
Please review CI failures and fix them, you added more code so it most likely requires test coverage for it.

baratali:
> Report with latest commit: https://rahulkhinchi03.github.io/
> (for isOfType() with first parameter)

Please include `IndentationCheck` to the config too. It uses `isOfType(ast, first)` via `NewHandler` class.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
Rahulkhinchi03: I have already added the code again. Please checkout.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
Rahulkhinchi03: @romani  @baratali Updated Report with latest commit: https://rahulkhinchi03.github.io/
(for isOfType() with first parameter)

romani:
Please fix survival
`+UnusedImportsCheck.java.html:<td class='covered'><pre><span  class='survived'>                &#38;&#38; TokenUtil.isOfType(ast.getPreviousSibling(), TokenTypes.DOT))</span></pre></td></tr>`
https://github.com/checkstyle/checkstyle/runs/6340354326?check_suite_focus=true

You can run on your local `./.ci/pitest.sh pitest-imports` to reproduce problem

nick-mancuso:
@Rahulkhinchi03 ping

Rahulkhinchi03:
@nick-mancuso @baratali I am currently focusing on CI labeled issues. I'll get back to this issue when I'll have some more free time.

Issue #11460, opened 2022-03-24
SuppressionXpathSingleFilter suppresses all violations when used with the default configuration


Vyom-Yadav:
I have read check documentation: https://checkstyle.sourceforge.io/config_filters.html#SuppressionXpathSingleFilter
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Test.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
  <module name="TreeWalker">
    <module name="SuppressionXpathSingleFilter"/>
    <module name="MissingJavadocTypeCheck">
    </module>
  </module>
</module>

/var/tmp $ cat Test.java
public class Test { // expected violation as all properties of filter are null.
}

/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.0-all.jar -c config.xml Test.java
Starting audit...
Audit done.
```
---

**Describe what you expect in detail.**

The default configuration has all the properties set to null but still the violation is suppressed, if I remove or comment the filter then a violation is thrown-
```xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
  <module name="TreeWalker">
<!--    <module name="SuppressionXpathSingleFilter"/>-->
    <module name="MissingJavadocTypeCheck">
    </module>
  </module>
</module>
```
```
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:1:1: Missing a Javadoc comment. [MissingJavadocType]
Audit done.
Checkstyle ends with 1 errors.
```

pbludov:
This case is strange. `SuppressionXpathSingleFilter` with all properties set to `null' is useless.
[...]

PR #11565, opened 2022-04-23
Issue #11460 Fixed the default behavior of SuppressoinXpathSingleFilter

as23415672:
Issue #11460 SuppressionXpathSingleFilter suppresses all violations when used with the default configuration

This problem is caused by XpathFilterElement.accept(TreeWalkerAuditEvent event) returning false when all the properties set to null. This is fixed by adding a function judging if all the properties are set to null.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
@@ -21,6 +21,7 @@ package com.puppycrawl.tools.checkstyle.filters;
 
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
@@ -70,6 +71,9 @@ public class XpathFilterElement implements TreeWalkerFilter {
     /** Xpath query. */
     private final String xpathQuery;
 
+    /** Indicates if all properties are set to null. */
+    private final boolean isEmptyConfig;
+
     /**
      * Creates a {@code XpathElement} instance.
      *
@@ -82,42 +86,11 @@ public class XpathFilterElement implements TreeWalkerFilter {
      */
     public XpathFilterElement(String files, String checks,
                        String message, String moduleId, String query) {
-        filePattern = files;
-        if (files == null) {
-            fileRegexp = null;
-        }
-        else {
-            fileRegexp = Pattern.compile(files);
-        }
-        checkPattern = checks;
-        if (checks == null) {
-            checkRegexp = null;
-        }
-        else {
-            checkRegexp = CommonUtil.createPattern(checks);
-        }
-        messagePattern = message;
-        if (message == null) {
-            messageRegexp = null;
-        }
-        else {
-            messageRegexp = Pattern.compile(message);
-        }
-        this.moduleId = moduleId;
-        xpathQuery = query;
-        if (xpathQuery == null) {
-            xpathExpression = null;
-        }
-        else {
-            final XPathEvaluator xpathEvaluator = new XPathEvaluator(
-                    Configuration.newConfiguration());
-            try {
-                xpathExpression = xpathEvaluator.createExpression(xpathQuery);
-            }
-            catch (XPathException ex) {
-                throw new IllegalArgumentException("Unexpected xpath query: " + xpathQuery, ex);
-            }
-        }
+        this(Optional.ofNullable(files).map(Pattern::compile).orElse(null),
+             Optional.ofNullable(checks).map(CommonUtil::createPattern).orElse(null),
+             Optional.ofNullable(message).map(Pattern::compile).orElse(null),
+             moduleId,
+             query);
     }
 
     /**
@@ -171,11 +144,17 @@ public class XpathFilterElement implements TreeWalkerFilter {
                 throw new IllegalArgumentException("Incorrect xpath query: " + xpathQuery, ex);
             }
         }
+        isEmptyConfig = fileRegexp == null
+                             && checkRegexp == null
+                             && messageRegexp == null
+                             && moduleId == null
+                             && xpathExpression == null;
     }
 
     @Override
     public boolean accept(TreeWalkerAuditEvent event) {
-        return !isFileNameAndModuleAndModuleNameMatching(event)
+        return isEmptyConfig
+                || !isFileNameAndModuleAndModuleNameMatching(event)
                 || !isMessageNameMatching(event)
                 || !isXpathQueryMatching(event);
     }
@@ -261,7 +240,8 @@ public class XpathFilterElement implements TreeWalkerFilter {
 
     @Override
     public int hashCode() {
-        return Objects.hash(filePattern, checkPattern, messagePattern, moduleId, xpathQuery);
+        return Objects.hash(filePattern, checkPattern, messagePattern,
+            moduleId, xpathQuery);
     }
 
     @Override
--- a/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
@@ -327,6 +327,79 @@ public class SuppressionXpathSingleFilterTest
         }
     }
 
+    @Test
+    public void testAllNullConfiguration() throws Exception {
+        final String[] expected = {
+            "18:1: " + getCheckMessage(MissingJavadocTypeCheck.class, MSG_JAVADOC_MISSING),
+        };
+
+        final String[] suppressed = CommonUtil.EMPTY_STRING_ARRAY;
+
+        verifyFilterWithInlineConfigParser(
+                getPath("InputSuppressionXpathSingleFilterAllNullConfiguration.java"),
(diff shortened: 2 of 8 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
```diff
@@ -177,7 +177,22 @@ public XpathFilterElement(Pattern files, Pattern checks, Pattern message,
...
+                || isAllNull();
+    }
+
+    /**
+     * Check if all properties are set to null.
+     *
+     * @return true if all properties are set to null
+     */
+
+    private boolean isAllNull() {
```
romani, line 190: Name it `areAllPropertiesNull`

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
as23415672: Done

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
```diff
@@ -327,6 +327,49 @@ public void testThrowException() throws Exception {
...
+
+        final String[] suppressed = CommonUtil.EMPTY_STRING_ARRAY;
+
+        verifyFilterWithInlineConfigParser(
+                getPath("InputSuppressionXpathSingleFilterAllNullConfiguration.java"),
+                expected, removeSuppressed(expected, suppressed));
+    }
+
+    @Test
+    public void testNullFirstThree() throws Exception {
```
romani, line 344: `testDefaultAllPropertiesExceptForIdAndExpression` put this update in Input name also

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
```diff
@@ -327,6 +327,49 @@ public void testThrowException() throws Exception {
...
+            "20:1: " + getCheckMessage(MissingJavadocTypeCheck.class, MSG_JAVADOC_MISSING),
+        };
+
+        verifyFilterWithInlineConfigParser(
+                getPath("InputSuppressionXpathSingleFilterNullFirstThree.java"),
+                expected, removeSuppressed(expected, suppressed));
+    }
+
+    @Test
+    public void testNullFile() throws Exception {
```
romani, line 359: `testDefaultFileProperty`

romani (changes requested):
TEamcity failure can be ignored, we are fixing it in other PR.

good, last improvements:

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
as23415672: done

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
as23415672: done

romani (approved):
Ok to merge

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
as23415672: ``testDefaultAllPropertiesExceptForIdAndExpression`` is too long, and I renamed it ``testDecideByIdAndExpression``.

romani:
Pitest failure:
>+XpathFilterElement.java.html:<td class='covered'><pre><span  class='survived'>        return fileRegexp == null</span></pre></td></tr>

We need extra test.

pbludov (approved):

pbludov (approved):

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
```diff
@@ -177,7 +177,22 @@ public XpathFilterElement(Pattern files, Pattern checks, Pattern message,
...
+     *
+     * @return true if all properties are set to null
+     */
+
+    private boolean areAllPropertiesNull() {
+        return fileRegexp == null
+                && checkRegexp == null
+                && messageRegexp == null
+                && moduleId == null
+                && xpathExpression == null;
```
strkkk, line 195: all these properties are initialized in constructor. So there is no need to check it for every event. You can just setup field in ctor and check this field first in `accept` method.

strkkk (changes requested):
item

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
```diff
@@ -70,6 +71,9 @@ public class XpathFilterElement implements TreeWalkerFilter {
     /** Xpath query. */
     private final String xpathQuery;
 
+    /** Indicates if all properties are set to Null. */
+    private final boolean allPropertiesAreNull;
```
strkkk, line 75: please rename to something like `isEmptyConfig`

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
```diff
@@ -70,6 +71,9 @@ public class XpathFilterElement implements TreeWalkerFilter {
     /** Xpath query. */
     private final String xpathQuery;
 
+    /** Indicates if all properties are set to Null. */
```
strkkk, line 74: Null -> null

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
```diff
@@ -171,13 +144,29 @@ public XpathFilterElement(Pattern files, Pattern checks, Pattern message,
...
+                || areAllPropertiesNull();
+    }
+
+    /**
+     * Check if all properties are set to null.
+     *
+     * @return true if all properties are set to null
+     */
+
+    private boolean areAllPropertiesNull() {
```
strkkk, line 168: I see no reason for private getter. You can access field directly, without getter

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
```diff
@@ -277,7 +267,8 @@ public boolean equals(Object other) {
                 && Objects.equals(checkPattern, xpathFilter.checkPattern)
                 && Objects.equals(messagePattern, xpathFilter.messagePattern)
                 && Objects.equals(moduleId, xpathFilter.moduleId)
-                && Objects.equals(xpathQuery, xpathFilter.xpathQuery);
+                && Objects.equals(xpathQuery, xpathFilter.xpathQuery)
+                && Objects.equals(allPropertiesAreNull, xpathFilter.allPropertiesAreNull);
```
strkkk, line 271: There is no reason to put this field in hash and equals method. If checks above for other fields were true, this one wont be different. 
I mean that there is no case when this comparison will be not true when we reached it.

strkkk (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
```diff
@@ -171,13 +144,29 @@ public XpathFilterElement(Pattern files, Pattern checks, Pattern message,
...
+                             && xpathExpression == null;
     }
 
     @Override
     public boolean accept(TreeWalkerAuditEvent event) {
         return !isFileNameAndModuleAndModuleNameMatching(event)
                 || !isMessageNameMatching(event)
-                || !isXpathQueryMatching(event);
+                || !isXpathQueryMatching(event)
+                || areAllPropertiesNull();
```
strkkk, line 159: this check can be first in checks chain

as23415672: Done

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
as23415672: Done

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
as23415672: Done

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
as23415672: Done

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
as23415672: Done

pbludov (approved):

Issue #11566, opened 2022-04-23
Resolve IDEA violations from Teamcity 2022.04 (Bundled IntelliJ IDEA 2021.2.3)


pbludov:
Link to the report: https://teamcity.jetbrains.com/viewLog.html?buildId=3844938&tab=Inspection&buildTypeId=Checkstyle_IdeaInspectionsPullRequest#

Mostly **Grammar** and **http links**, but other violations should be fixed.

## General
* [ ] Redundant suppression (Errors) (12)
* [ ] Syntax error (Errors) (2)
## Java/Abstraction issues
* [ ] Private method only used from inner class (Errors) (1)
## Java/Code maturity
* [ ] Commented out code (2)
* [ ] Deprecated member is still used (2)
## Java/Code style issues
* [ ] 'if' statement can be replaced with conditional or boolean expression (Errors) (4)
## Java/Data flow
* [ ] Scope of variable is too broad (Errors) (2)
* [ ] Java/Java language level migration aids/Java 9
* [ ] Null check can be replaced with method call (4)
## Java/JUnit
* [ ] JUnit test method without any assertions (Errors) (8)
## Java/Memory
* [ ] Static collection (73)
## Java/Performance
* [ ] Call to 'set.removeAll(list)' may work slowly (2)
## Java/Probable bugs
* [ ] Result of method call ignored (1)
## Java/Threading issues
* [ ] 'synchronized' method (1)
## Java/Verbose or redundant code constructs
* [ ] Redundant String operation (2)
* [ ] Redundant usage of unmodifiable collection wrappers (2)
## JavaScript and TypeScript/Potentially undesirable code constructs
* [ ] Anonymous function (Errors) (7)
## Proofreading
* [ ] Grammar (379)
## RegExp
* [ ] Duplicate character in character class (2)
## Security
* [ ] Link with unencrypted protocol (64)
## YAML
* [ ] Validation by JSON Schema (10)

romani:
To speed up green state, it might be good to disable some inspections until fix is provided.

pbludov:
I'll try to fix all errors asap, warnings may be postponed.

romani:
I asked TC support to install previous version , they installed 2020.1.1

I activated this version in master build only, build is started
https://teamcity.jetbrains.com/buildConfiguration/Checkstyle_IdeaInspectionsMaster/3847241?hideProblemsFromDependencies=false&hideTestsFromDependencies=false

pbludov:
[...]

PR #11567, opened 2022-04-23
Resolve IDEA violations from Teamcity 2022.04

pbludov:
Issue #11566:
- [x] commented out code
- [x] redundant suppression
- [x] scope of variable is too broad
- [x] JUnit test method without any assertions
- [x] anonymous function (js)
- [x] private method only used from inner class

### DIFF
--- a/config/intellij-idea-inspections.xml
+++ b/config/intellij-idea-inspections.xml
@@ -649,7 +649,8 @@
                    enabled_by_default="false"/>
   <inspection_tool class="AnonymousClassVariableHidesContainingMethodVariable" enabled="true"
                    level="WARNING" enabled_by_default="true"/>
-  <inspection_tool class="AnonymousFunctionJS" enabled="true" level="ERROR"
+  <!-- Anonymous Javascript functions are good for self executing and callbacks. -->
+  <inspection_tool class="AnonymousFunctionJS" enabled="true" level="WARNING"
                    enabled_by_default="true"/>
   <inspection_tool class="AnonymousHasLambdaAlternative" enabled="true" level="WARNING"
                    enabled_by_default="true"/>
--- a/config/intellij-idea-inspections.xml
+++ b/config/intellij-idea-inspections.xml
@@ -4592,24 +4592,19 @@
                    enabled_by_default="true"/>
   <inspection_tool class="TestMethodWithoutAssertion" enabled="true" level="ERROR"
                    enabled_by_default="true">
-    <!-- after https://github.com/powermock/powermock/issues/830
-             all junit4 methods should be removed -->
     <option name="assertionMethods"
-            value="org.junit.Assert,assert.*|fail.*, ,
-    ,org.junit.jupiter.api.Assertions,assert.*|fail.*, ,
-    ,junit.framework.Assert,assert.*|fail.*, ,
+            value="org.junit.jupiter.api.Assertions,assert.*|fail.*, ,
     ,org.mockito.Mockito,verify.*, ,
     ,org.mockito.InOrder,verify, ,
-    ,org.junit.rules.ExpectedException,expect.*, ,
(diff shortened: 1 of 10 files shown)


romani (approved):

romani:
Please do not put all in one PR, it will be hard to review and merge.

pbludov:
> Please do not put all in one PR, it will be hard to review and merge.

OK, no more commits for this PR.

config/intellij-idea-inspections.xml
```diff
@@ -3085,7 +3086,8 @@
                    enabled_by_default="true"/>
   <inspection_tool class="MethodOnlyUsedFromInnerClass" enabled="true" level="ERROR"
                    enabled_by_default="true">
-    <option name="ignoreMethodsAccessedFromAnonymousClass" value="false"/>
+    <!-- Better to keep anonymous classes as small as possible. -->
+    <option name="ignoreMethodsAccessedFromAnonymousClass" value="true"/>
```
romani, line 3090: > TreeTable.java (1) expandSelectedNode() Method expandSelectedNode()

our code:
https://github.com/checkstyle/checkstyle/blob/591b41028cecc786ea94556791a8e53a99defb4c/src/main/java/com/puppycrawl/tools/checkstyle/gui/TreeTable.java#L115-L133

I am ok with this change.

src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java
```diff
@@ -95,18 +95,11 @@ private JavadocUtil() {
...
-
         final List<TagInfo> tags = new ArrayList<>();
-
-        if (getBlockTags) {
+        if (tagType == JavadocTagType.ALL || tagType == JavadocTagType.BLOCK) {
             tags.addAll(BlockTagUtil.extractBlockTags(textBlock.getText()));
         }
-
-        if (getInlineTags) {
+        if (tagType == JavadocTagType.ALL || tagType == JavadocTagType.INLINE) {
```
romani, line 102: > src/main/java/com/puppycrawl/tools/checkstyle/utils
JavadocUtil.java (1) getJavadocTags() Scope of variable getInlineTags is too broad

this fix is controversial.
if demand variable to be moved close to IF , I am agree.

Inlining expressions is not very good as it variable help us to avoid comment over IF. But I agree that name of variable should be changed to `isInlineTags`  same for other variable.

romani (changes requested):
the only item:

src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java
pbludov: Done.

romani (approved):
Ok to merge

src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java
```diff
@@ -95,18 +95,15 @@ private JavadocUtil() {
...
-                                         || tagType == JavadocTagType.BLOCK;
-        final boolean getInlineTags = tagType == JavadocTagType.ALL
-                                          || tagType == JavadocTagType.INLINE;
-
         final List<TagInfo> tags = new ArrayList<>();
-
-        if (getBlockTags) {
+        final boolean isBlockTags = tagType == JavadocTagType.ALL
+                                        || tagType == JavadocTagType.BLOCK;
+        if (isBlockTags) {
```
rnveach, line 101: Not really sure I prefer this inspection. It moves it closer to it's usage, but its the still the same execution. It is not like the move inside if check, where it could prevent it being executed in certain cases.

rnveach (approved):

nick-mancuso (approved):

nick-mancuso (approved):
I am good once CI is happy.

Issue #11528, opened 2022-04-09
Split pitest-coding execution into two executions and investigate instability


romani:
pitest-coding is running too long in comparison to other and fails too often

we need stabilize it and make it quicker to execute.

I am open to any solutions.

related to https://github.com/checkstyle/checkstyle/issues/11427

romani:
@Vyom-Yadav , can you help us with this ?

Vyom-Yadav:
@romani Sure, I am on it.

nick-mancuso:
Let’s make multiple PRs to close this issue. One for each instability, and then final PR to split execution.

romani:
EqualsAvoidNullCheck is fixed, Requirethis still need to be fixed, but we moved it to temp suppression to allow CI be green.

nick-mancuso:
@Vyom-Yadav ping, let's finish this up by splitting execution of pitest-coding so that we can achieve faster feedback loop in https://github.com/checkstyle/checkstyle/pull/11672

nick-mancuso:
Closed via https://github.com/checkstyle/checkstyle/pull/11707

PR #11551, opened 2022-04-19
Issue #11528: Kill surviving mutation in EqualsAvoidNullCheck

Vyom-Yadav:
Issue #11528 
The rationale is explained [here](https://github.com/checkstyle/checkstyle/issues/11427#issuecomment-1102539871).
Does not resolve the complete issue.

PR #11571, opened 2022-04-24
Issue #11528: Kill surviving mutation in RequireThisCheck

Vyom-Yadav:
#11528 

https://github.com/checkstyle/checkstyle/blob/78ef41d24560f880052f02a049da80e560580996/.ci/pitest.sh#L212

### Reports (Clean)-

---

```xml
<module name="RequireThis"/>
```
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/cbab1ad_2022060550/reports/diff/index.html

---

```xml
<module name="RequireThis">
  <property name="validateOnlyOverlapping" value="false"/>
</module>
```

WithoutOpenJDK - https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/db6616b_2022050743/reports/diff/index.html
Report for open JDK - https://vyom-yadav.github.io/DiffReport/diffRequireThisOpenJDK/index.html

### DIFF
--- a/.ci/pitest.sh
+++ b/.ci/pitest.sh
@@ -207,11 +207,7 @@ pitest-coding)
   "VariableDeclarationUsageDistanceCheck.java.html:<td class='covered'><pre><span  class='survived'>        if (!isVarInOperatorDeclaration &#38;&#38; operator.getType() == TokenTypes.LITERAL_IF) {</span></pre></td></tr>"
   "VariableDeclarationUsageDistanceCheck.java.html:<td class='covered'><pre><span  class='survived'>        while (result</span></pre></td></tr>"
   );
-  # Until https://github.com/checkstyle/checkstyle/issues/11427
-  declare -a unstableItems=(
-  "RequireThisCheck.java.html:<td class='covered'><pre><span  class='survived'>            if (returnedVariable) {</span></pre></td></tr>"
-  );
-  checkPitestReport ignoredItems unstableItems
+  checkPitestReport ignoredItems
   ;;
 
 # pitesttyle-gui)
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
@@ -817,14 +817,8 @@ public class RequireThisCheck extends AbstractCheck {
         final Set<DetailAST> returnsInsideBlock = getAllTokensOfType(definitionToken,
             TokenTypes.LITERAL_RETURN, blockEndToken.getLineNo());
 
-        boolean returnedVariable = false;
-        for (DetailAST returnToken : returnsInsideBlock) {
-            returnedVariable = isAstInside(returnToken, ident);
-            if (returnedVariable) {
-                break;
-            }
-        }
-        return returnedVariable;
+        return returnsInsideBlock.stream()
+            .anyMatch(returnToken -> isAstInside(returnToken, ident));
     }
 
     /**
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisCatchVariables.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisCatchVariables.java
@@ -46,4 +46,16 @@ public class InputRequireThisCatchVariables extends Thread {
 
     private static void exceptional(Throwable ex) {}
     private static void debug(String message, Throwable err) {}
+
+    int baseline;
+    public int getBaseline() {
+        int baseline = 12;
+        final int placement = 3;
+        switch (placement) {
+            case 3:
+                baseline = 3 + baseline;
+                return baseline;
+        }
+        return -1;
+    }
 }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisExtendedMethod.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisExtendedMethod.java
@@ -16,4 +16,12 @@ public class InputRequireThisExtendedMethod // ok
     public class Check {
         private Logger log1 = Logger.getLogger(getClass().getName());
     }
+
+    String ELIST;
+    String method(Object a) {
+        String ELIST = "abc";
+        ELIST = this.method(new Check() {
+        });
+        return "";
+    }
 }


Vyom-Yadav:
Github, generate report

Vyom-Yadav:
Github, generate report

Vyom-Yadav:
Github, generate report

Vyom-Yadav:
Github, generate report

Vyom-Yadav:
Github, generate report

Vyom-Yadav:
> Report generation job failed on phase "make_report", step "Generate report".Link: https://github.com/checkstyle/checkstyle/actions/runs/2237142722

Getting out of memory error, will generate this report manually.

Vyom-Yadav:
Github, generate report

Vyom-Yadav:
Github, generate report

Vyom-Yadav:
With openJDK: https://vyom-yadav.github.io/DiffReport/diffRequireThisOpenJDK/index.html

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -756,11 +756,11 @@ private static boolean isUserDefinedArrangementOfThis(AbstractFrame currentFrame
...
         final DetailAST definitionToken = blockFrameNameIdent.getParent();
         final DetailAST blockStartToken = definitionToken.findFirstToken(TokenTypes.SLIST);
-        final DetailAST blockEndToken = getBlockEndToken(blockFrameNameIdent, blockStartToken);
+        final DetailAST blockEndToken = getBlockEndToken(blockStartToken);
 
         boolean userDefinedArrangementOfThis = false;
 
-        final Set<DetailAST> variableUsagesInsideBlock =
+        final List<DetailAST> variableUsagesInsideBlock =
             getAllTokensWhichAreEqualToCurrent(definitionToken, ident,
```
Vyom-Yadav, line 764: Random arrangements of elements in the `Set` was causing this. There will be no performance hit as the `Set`(`HashSet`) was being traversed every time, none of the properties of hashing were used, if anything TC will be reduced (No HashTable made)

This applies to all the places where `Set` is replaced by `List`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -778,27 +778,11 @@ private static boolean isUserDefinedArrangementOfThis(AbstractFrame currentFrame
...
-            for (DetailAST currentRcurly : rcurlyTokens) {
-                final DetailAST parent = currentRcurly.getParent();
-                if (TokenUtil.areOnSameLine(blockStartToken, parent)) {
-                    blockEndToken = currentRcurly;
-                }
-            }
-        }
-        return blockEndToken;
+    private static DetailAST getBlockEndToken(DetailAST blockStartToken) {
+        return blockStartToken.getLastChild();
```
Vyom-Yadav, line 785: I went over the cases, this should be fine. The block end token is the last child of the block start token. Though variables in switches can also be globally (by globally I mean throughout the switch statement) scoped, this shouldn't cause a problem as the variable would have to be initialized separately in every switch case. Diff report is also clean.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -775,32 +775,6 @@ private static boolean isUserDefinedArrangementOfThis(AbstractFrame currentFrame
...
-            for (DetailAST currentRcurly : rcurlyTokens) {
-                final DetailAST parent = currentRcurly.getParent();
-                if (TokenUtil.areOnSameLine(blockStartToken, parent)) {
-                    blockEndToken = currentRcurly;
-                }
-            }
-        }
-        return blockEndToken;
-    }
-
```
Vyom-Yadav, line 803: I went over the cases, this should be fine. The block end token is the last child of the block start token. Though variables in switches can also be globally (by globally I mean throughout the switch statement) scoped, this shouldn't cause a problem as the variable would have to be initialized separately in every switch case. Diff report is also clean.

For switches, the last child will be the last statement in that block.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
romani: >Random arrangements of elements in the Set was causing this.

Unfortunately this a reason to disable pitest mutation or keep suppression forever.
We can not damage logic of Check to make some pitest happy.
All tools have mistakes and false positives, pitest is one of them. Until now we always find a way to change code to not really damage logic. But we were on age few times, and some time did updates in favor of pitest. But it should not be our primary goal to make pitest happy. our code should be logical and robust.

We need to rethink this. Or I need more details on survivals.

Vyom-Yadav: >  We can not damage the logic of Check to make some pitest happy.

It is not being damaged, replacing Sets with Lists shouldn't be a problem, we were actually using Sets as lists earlier also, none of the hashing properties were being utilized, it was just elements arranged randomly.  All the usages of sets were inside for loops. If a regression appears in the future, it will be investigated and fixed.

I don't think this is a false positive and I feel pitest did a good job, RequireThis uses frames, and frames are added/removed in the order in which they are present in code, randomly arranging them just for traversing them later is never a good idea.

> We need to rethink this. Or I need more details on survivals.

Survivals after changing sets to lists?

romani: set does not allow us to put same ast twice, this is main criteria why this collection is selected in many places.

rnveach: How does random order in set cause an issue for killing mutations?

Vyom-Yadav: > set does not allow us to put same ast twice, this is main criteria why this collection is selected in many places.

Why would there be duplicates? none of the algorithms that involve updating the set visit the same ast twice. I even checked it explicitly by modifying the code to compare elements in `list` and in `new HashSet<>(list)`, they were always the same, went through the algorithms manually to check if such case is possible, it is not.
Please share a case where you think there will be duplicate entries.

Vyom-Yadav: > How does random order in set cause an issue for killing mutations?

The reason is similar to https://github.com/checkstyle/checkstyle/issues/11427#issuecomment-1102539871.

nick-mancuso: @Vyom-Yadav could we use a `LinkedHashSet` here? Then, we avoid duplicates and also maintain order.

Vyom-Yadav: I had originally planned to use that only, used it in prev PR, https://github.com/checkstyle/checkstyle/pull/11551#pullrequestreview-946568511, but it was suggested to not use it if hash benefits are not availed, I don't see how can the list end up having duplicates but I am ok with using LinkedHashSet to be on the safer side if everyone agrees.

romani: Why we need order ?

Vyom-Yadav: > Why we need order?

The mutation can be killed without order also but shouldn't be IMO. Also, why pay extra for randomness when we can get the same result from using it in an ordered way.
The scenario-
```
final List<DetailAST> returnsInsideBlock = getAllTokensOfType(definitionToken,
    TokenTypes.LITERAL_RETURN, blockEndToken.getLineNo());                    

for (DetailAST returnToken : returnsInsideBlock) {      
    returnedVariable = isAstInside(returnToken, ident); 
    if (returnedVariable) { // Mutation is here, replaced it with false, never breaking from the loop                             
        break;                                          
    }                                                   
}                                                       
```
Now if we never break, there are cases when the latter `returnToken` won't be inside the tree and `returnedVariable` will return false, which will override the previous value of it being `true`, this can be killed if we add a condition to only change `returnedVariable` when it false. 

This way we will traverse the whole loop irrespective of whether we have found our result or not.
This isn't very good in my opinion. Another reason for replacing sets, which I mentioned earlier too is that we don't avail any hash benefits, duplication is not possible IMO, please share if you think duplication is possible.

pbludov: This can be written with Stream API:

```java
        return returnsInsideBlock.stream()
                .anyMatch(returnToken -> isAstInside(returnToken, ident));
```

nick-mancuso: @Vyom-Yadav ping

Vyom-Yadav: > This can be written with Stream API:

Yes, that will work, we can discuss replacing sets with lists in some other issue.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisCatchVariables.java
```diff
@@ -46,4 +46,16 @@ public void run() {
...
+    public int getBaseline() {
+        int baseline = 12;
+        final int placement = 3;
+        switch (placement) {
+            case 3:
+                baseline = 3 + baseline;
+                return baseline;
+        }
+        return -1;
+    }
```
Vyom-Yadav, line 60: There is no harm in keeping these test cases, they were killing some mutations when sets were replaced with lists. Same for the other example.

Vyom-Yadav:
Github, generate report

.ci/pitest.sh
```diff
@@ -10,16 +10,6 @@ function checkPitestReport() {
   grep -irE "$SEARCH_REGEXP" target/pit-reports \
      | sed -E 's/.*\/([A-Za-z]+.java.html)/\1/' | LC_ALL=C sort > target/actual.txt
   printf "%s\n" "${ignored[@]}" | sed '/^$/d' | LC_ALL=C sort > target/ignored.txt
-  if [ -n "$2" ] ; then
-      local -n unstableMutations=$2
```
romani, line 14: Please restore functionality to support unstable.
Just not use unstable.

romani (changes requested):
Items

.ci/pitest.sh
Vyom-Yadav: Done.

romani (approved):
A way better than original fix

nick-mancuso (approved):

rnveach (approved):

PR #11598, opened 2022-05-02
minor: removes checkstyle specific class in ReverseListIterator

rnveach:
`ReverseListIterator` is basically a utility class. It shouldn't have anything specific to checkstyle similar to `ReverseDescendantIterator` and the others...

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/xpath/iterators/ReverseListIterator.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/xpath/iterators/ReverseListIterator.java
@@ -22,7 +22,6 @@ package com.puppycrawl.tools.checkstyle.xpath.iterators;
 import java.util.ArrayList;
 import java.util.List;
 
-import com.puppycrawl.tools.checkstyle.xpath.AbstractNode;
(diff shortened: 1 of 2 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/xpath/iterators/ReverseListIterator.java
```diff
@@ -22,7 +22,6 @@
 import java.util.ArrayList;
 import java.util.List;
 
-import com.puppycrawl.tools.checkstyle.xpath.AbstractNode;
```
romani, line 25: Can we support not usage by import control Check ?

romani (changes requested):
Items

nick-mancuso (approved):

src/main/java/com/puppycrawl/tools/checkstyle/xpath/iterators/ReverseListIterator.java
rnveach: Done.

romani (approved):
Ok to merge

PR #11603, opened 2022-05-03
Some reworking for test structures

rnveach:
#11604

First commit is a simple synchronize between Test and IT where similar code is.

Second commit is the bigger change. I reworked how we build the Checker tree so it is more streamlined and not passing around the enumeration. `createChecker` will now handle the current given module and identify if it needs a TreeWalker or other similar intermediary between it and the module. It also removed random config names and more closely used the ones they represent.

I could not make similar changes to the IT side as the IT side has some weird complexities that need to be dealt with. We pass the name of the check and not the full class. We have to pull things from an existing config, and we only do partial pulls, still rebuild the missing pieces. There are also some tests that pull in more information from the config to not duplicate it making it all very complicated and needing its own rewrite.

I at least removed and centralized the code for `getModuleConfig` and `getModuleConfigs` for IT.

### DIFF
--- a/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
+++ b/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
@@ -45,6 +45,7 @@ import com.puppycrawl.tools.checkstyle.Checker;
 import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
 import com.puppycrawl.tools.checkstyle.TreeWalker;
 import com.puppycrawl.tools.checkstyle.api.AbstractViolationReporter;
+import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
 import com.puppycrawl.tools.checkstyle.api.Configuration;
 import com.puppycrawl.tools.checkstyle.internal.utils.BriefUtLogger;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
@@ -85,20 +86,89 @@ public abstract class AbstractItModuleTestSupport extends AbstractPathTestSuppor
     protected abstract ModuleCreationOption findModuleCreationOption(String moduleName);
 
     /**
-     * Creates {@link DefaultConfiguration} instance for the given module class.
-     *
-     * @param clazz module class.
-     * @return {@link DefaultConfiguration} instance.
-     */
-    protected abstract DefaultConfiguration createModuleConfig(Class<?> clazz);
-
-    /**
      * Returns test logger.
      *
      * @return logger for tests
      */
     protected final BriefUtLogger getBriefUtLogger() {
         return new BriefUtLogger(stream);
+    }
+
+    /**
+     * Creates a default module configuration {@link DefaultConfiguration} for a given object
+     * of type {@link Class}.
+     *
+     * @param clazz a {@link Class} type object.
+     * @return default module configuration for the given {@link Class} instance.
+     */
+    protected static DefaultConfiguration createModuleConfig(Class<?> clazz) {
+        return new DefaultConfiguration(clazz.getName());
+    }
+
+    /**
+     * Returns {@link Configuration} instance for the given module name pulled
+     * from the {@code masterConfig}.
+     *
+     * @param masterConfig The master configuration to examine.
+     * @param moduleName module name.
+     * @param moduleId module id.
+     * @return {@link Configuration} instance for the given module name.
+     * @throws IllegalStateException if there is a problem retrieving the module
+     *         or config.
+     */
+    protected static Configuration getModuleConfig(Configuration masterConfig, String moduleName,
+            String moduleId) {
+        final Configuration result;
+        final List<Configuration> configs = getModuleConfigs(masterConfig, moduleName);
+        if (configs.size() == 1) {
+            result = configs.get(0);
+        }
+        else if (configs.isEmpty()) {
+            throw new IllegalStateException("no instances of the Module was found: " + moduleName);
+        }
+        else if (moduleId == null) {
+            throw new IllegalStateException("multiple instances of the same Module are detected");
+        }
+        else {
+            result = configs.stream().filter(conf -> {
+                try {
+                    return conf.getProperty("id").equals(moduleId);
+                }
+                catch (CheckstyleException ex) {
+                    throw new IllegalStateException("problem to get ID attribute from " + conf, ex);
+                }
+            })
+            .findFirst()
+            .orElseThrow(() -> new IllegalStateException("problem with module config"));
+        }
+
+        return result;
+    }
+
+    /**
+     * Returns a list of all {@link Configuration} instances for the given
+     * module name pulled from the {@code masterConfig}.
+     *
+     * @param masterConfig The master configuration to examine.
+     * @param moduleName module name.
+     * @return {@link Configuration} instance for the given module name.
+     */
+    protected static List<Configuration> getModuleConfigs(Configuration masterConfig,
+            String moduleName) {
+        final List<Configuration> result = new ArrayList<>();
+        for (Configuration currentConfig : masterConfig.getChildren()) {
+            if ("TreeWalker".equals(currentConfig.getName())) {
+                for (Configuration moduleConfig : currentConfig.getChildren()) {
+                    if (moduleName.equals(moduleConfig.getName())) {
+                        result.add(moduleConfig);
+                    }
+                }
+            }
+            else if (moduleName.equals(currentConfig.getName())) {
+                result.add(currentConfig);
+            }
+        }
+        return result;
     }
 
     /**
@@ -158,9 +228,9 @@ public abstract class AbstractItModuleTestSupport extends AbstractPathTestSuppor
      * @return {@link DefaultConfiguration} for the {@link TreeWalker}
      *     based on the given {@link Configuration} instance.
      */
-    protected final DefaultConfiguration createTreeWalkerConfig(Configuration config) {
+    protected static DefaultConfiguration createTreeWalkerConfig(Configuration config) {
         final DefaultConfiguration dc =
-                new DefaultConfiguration("configuration");
+                new DefaultConfiguration(ROOT_MODULE_NAME);
         final DefaultConfiguration twConf = createModuleConfig(TreeWalker.class);
         // make sure that the tests always run with this charset
         dc.addProperty("charset", StandardCharsets.UTF_8.name());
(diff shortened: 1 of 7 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
```diff
@@ -110,55 +92,44 @@ protected static DefaultConfiguration createModuleConfig(Class<?> clazz) {
...
      */
-    public final Checker createChecker(Configuration moduleConfig,
-                                 ModuleCreationOption moduleCreationOption)
-            throws Exception {
-        final Checker checker = new Checker();
-        checker.setModuleClassLoader(Thread.currentThread().getContextClassLoader());
-
-        if (moduleCreationOption == ModuleCreationOption.IN_TREEWALKER) {
+    protected void addConfigToChecker(Checker checker, Class<?> moduleClass,
+            Configuration moduleConfig) throws Exception {
```
rnveach, line 123: This is the method that can be overridden to allow new walkers and why the old enum way had to be removed.

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
```diff
@@ -85,35 +86,89 @@ public enum ModuleCreationOption {
...
+        }
+        else if (configs.isEmpty()) {
+            throw new IllegalStateException("no instances of the Module was found: " + moduleName);
+        }
+        else if (moduleId == null) {
+            throw new IllegalStateException("multiple instances of the same Module are detected");
+        }
+        else {
+            result = configs.stream().filter(conf -> {
+                try {
```
pbludov, line 134: Please extract this to a helper method.
try/catch lambdas looks weird.

rnveach: Done.

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
```diff
@@ -131,54 +196,54 @@ protected final Checker createChecker(Configuration moduleConfig)
...
+        checker.setModuleClassLoader(Thread.currentThread().getContextClassLoader());
+        // make sure the tests always run with English error messages
+        // so the tests don't fail in supported locales like German
+        final Locale locale = Locale.ENGLISH;
+        checker.setLocaleCountry(locale.getCountry());
+        checker.setLocaleLanguage(locale.getLanguage());
 
         if (moduleCreationOption == ModuleCreationOption.IN_TREEWALKER) {
-            dc = createTreeWalkerConfig(moduleConfig);
+            final Configuration dc = createTreeWalkerConfig(moduleConfig);
```
nick-mancuso, line 219: Can we improve naming while we refactor? `dc` is just `defaultConfiguration` right?

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
```diff
@@ -131,54 +196,54 @@ protected final Checker createChecker(Configuration moduleConfig)
...
      */
-    protected final DefaultConfiguration createTreeWalkerConfig(Configuration config) {
+    protected static DefaultConfiguration createTreeWalkerConfig(Configuration config) {
         final DefaultConfiguration dc =
-                new DefaultConfiguration("configuration");
+                new DefaultConfiguration(ROOT_MODULE_NAME);
         final DefaultConfiguration twConf = createModuleConfig(TreeWalker.class);
         // make sure that the tests always run with this charset
-        dc.addProperty("charset", "iso-8859-1");
+        dc.addProperty("charset", StandardCharsets.UTF_8.name());
```
nick-mancuso, line 246: These two charsets are not equivalent, why did we change this? `StandardCharsets` has a constant for this charset.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
```diff
@@ -79,7 +61,7 @@ public enum ModuleCreationOption {
      *
      * @return stream with log
      */
-    public ByteArrayOutputStream getStream() {
+    protected final ByteArrayOutputStream getStream() {
```
nick-mancuso, line 64: Just curious, why not allow extension of this?

nick-mancuso (changes requested):
Couple questions, only one minor item:

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
rnveach: I see this more as probably fixing a mistake. Both test areas have the same comment, similar code but each one has its own "charset". I am syncing the code between the 2 areas and do not see why the charsets would be different since they are both basically doing the same thing.

This has been here since 2015. The commit was 437986510709d7abc64b59c72779343ac5e77599 and I could not find the PR to see what documentable reason for this charset.

https://checkstyle.org/styleguides/google-java-style-20180523/javaguide.html#s2.2-file-encoding
Google's guide specifically says UTF8, so it makes even weirder we are using a different charset for our tests than they recommend.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
rnveach: I would say because this a simple getter. If we allowed this to be overriden, then we would always have to call the method instead of the field for our uses in the abstract. Overriding this amounts to providing their own custom stream implementation. Nothing in our area uses it, and I see no uses for it being overridden so hence the `final`.

Truthfully I would prefer to remove this method since it has no real uses, except for WriteTagCheck which is just a flawed check in some aspects.

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
rnveach: Done.

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
nick-mancuso: Yeah, I can’t imagine that this would cause any issues. It’s weird that we used two different charsets for basically the same thing. 

Resolved.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
nick-mancuso: Makes sense to me, I wonder why this is not more common now that you explain it.

Resolved.

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
nick-mancuso: Resolved.

nick-mancuso (approved):

pbludov (approved):

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
```diff
@@ -137,39 +94,42 @@ protected ModuleCreationOption findModuleCreationOption(String moduleName) {
...
+     * @throws Exception if an exception occurs during configuration.
      */
-    protected final Checker createChecker(Configuration moduleConfig,
-                                 ModuleCreationOption moduleCreationOption)
-            throws Exception {
-        final Checker checker = new Checker();
-        checker.setModuleClassLoader(Thread.currentThread().getContextClassLoader());
-
-        if (moduleCreationOption == ModuleCreationOption.IN_TREEWALKER) {
+    protected void addConfigToChecker(Checker checker, Class<?> moduleClass,
```
romani, line 122: `addConfigToItsParent(Checker, Configuration)`

moduleClass is already in moduleConfig.

romani (changes requested):
item:

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
rnveach: > addConfigToItsParent
> module does not go directly to Checker

The method is named as such because the end result it is trying to add `moduleConfig` to `Checker` but there could be something intermediary between the 2, such as `TreeWalker`. If `TreeWalker` is the parent, then the suggested method name is misleading as we are doing more than adding it to `TreeWalker`. What we really need is to add it to Checker.

So saying "it does not go directly to Checker" is wrong.

> (Checker, Configuration)
> moduleClass is already in moduleConfig.

This was to simplify this method for overloading. The `Class` is needed to pass to `ModuleReflectionUtil` to know it's heritage. To remove it from the method, means the code creating the class has to be moved in this method, and any overrides of this method will need to duplicate it as the `Class` is always needed.

Also `moduleConfig` only has the class in the Test tier. IT tier works with the configuration and thus does not already have the class in `moduleConfig`. This is what is preventing synchronization right now. When IT tier can be synchronized fully to test, it will need a special method to convert config to class, and thus it makes this method more interoperable that it receive the class instead of creating it itself, as IT will have need for a custom conversion.

romani: this method actually do "configure" so it is instantiation of modules , lets call it `configureChecker`.

```
protected final Checker createChecker(Configuration moduleConfig)
            throws Exception {
        final String moduleName = moduleConfig.getName();
        final Checker checker = new Checker();
        checker.setModuleClassLoader(Thread.currentThread().getContextClassLoader());

        if (ROOT_MODULE_NAME.equals(moduleName)) {
            checker.configure(moduleConfig);
        }
        else { 
           configureChecker(checker, moduleConfig);
        }

        checker.addListener(getBriefUtLogger());
        return checker;
    }

    protected void configureChecker(Checker checker, Configuration moduleConfig) throws Exception {
          final Class<?> moduleClass = Class.forName(moduleName);

        if (ModuleReflectionUtil.isCheckstyleTreeWalkerCheck(moduleClass)
                || ModuleReflectionUtil.isTreeWalkerFilterModule(moduleClass)) {
            final Configuration config = createTreeWalkerConfig(moduleConfig);
            checker.configure(config);
        }
        else {
            final Configuration config = createRootConfig(moduleConfig);
            checker.configure(config);
        }
    }
```

rnveach: Done.

romani: Not done.
`configureChecker(checker, moduleClass, moduleConfig);`

Should be
`configureChecker(checker, moduleConfig);`

As I showed in my code above.

rnveach: See comment at https://github.com/checkstyle/checkstyle/pull/11603#discussion_r867400591

rnveach: I pushed the change.

romani (approved):
Ok to merge

Issue #10043, opened 2021-05-26
SuppressWarnings: support not working on local variables


rnveach:
https://checkstyle.org/config_filters.html#SuppressWarningsFilter_Description

````
$ cat TestClass.java
public class TestClass {
    @SuppressWarnings("MagicNumber")
    private int A = 8;

    @SuppressWarnings("MagicNumber")
    void method() {
        int b = 8;
    }

    void method2() {
        @SuppressWarnings("MagicNumber")
        int c = 8;
    }
}

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
<module name="MagicNumber"/>
<module name="SuppressWarningsHolder" />
    </module>
  <module name="SuppressWarningsFilter" />
</module>

$ java -jar checkstyle-8.42-all.jar -c TestConfig.xml TestClass.java
Starting audit...
[ERROR] TestClass.java:12:17: '8' is a magic number. [MagicNumber]
Audit done.
Checkstyle ends with 1 errors.
````

Suppression on line 3 on field works. Ok.
Suppression on line 7 on method works. Ok.
Suppression on line 12 does not work even though the annotation is added. This is the point of this issue.

I don't see why annotations on variables should not be supported as long as the code compiles, which it does.

Since suppression location is valid to me, it seems we should support this and be able to suppress violations in this manner as this is what check is meant to do.

**Comment by Vyom-Yadav** [comment]
I am on it.

**Comment by Vyom-Yadav** [comment]
@rnveach @romani  I worked on it for quite sometime and found that this is something specific to `MagicNumber`. While debugging also I didn't find any problem in `SuppressWarningsHolder`, it supports local variable too.

**Config**
```
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
        <module name="LocalFinalVariableName"/>
        <module name="SuppressWarningsHolder" />
    </module>
    <module name="SuppressWarningsFilter" />
</module>
```

**Code**
[...]

PR #11609, opened 2022-05-05
Issue #10043: Fixed SuppressWarnings not working on local variables

RandyLovevivi:
Issue #10043

The bug in line 12 is caused by the following line [lastColumn = nextAST.getColumnNo() - 1;](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/SuppressWarningsHolder.java#L338)
It will lead to the 8 of the method2() be out of the range of the SuppressWarnings and will not be suppressed. The end position of the SuppressWarnings will be located to the ";" at the end of the code line. If it minus one, it will not cover the "8", which is just before the ";" .


romani (approved):
Ok to merge

pbludov (approved):

strkkk (approved):

PR #11611, opened 2022-05-06
Issue #11140: solve PRMC_POSSIBLY_REDUNDANT_METHOD_CALLS

pbludov:
Part of #11140 

In fact, this is nano optimization with zero impact. But this forces us to add more variables to the code. This makes the code readable and easy to debug.

There is a false-positive for `Checker` and `TreeWalker`. See https://github.com/mebigfatguy/fb-contrib/issues/329 for the details.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/JavaAstVisitor.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/JavaAstVisitor.java
@@ -146,9 +146,10 @@ public final class JavaAstVisitor extends JavaLanguageParserBaseVisitor<DetailAs
         final DetailAstImpl importRoot = create(ctx.start);
 
         // Static import
-        if (ctx.LITERAL_STATIC() != null) {
+        final TerminalNode literalStaticNode = ctx.LITERAL_STATIC();
+        if (literalStaticNode != null) {
             importRoot.setType(TokenTypes.STATIC_IMPORT);
-            importRoot.addChild(create(ctx.LITERAL_STATIC()));
+            importRoot.addChild(create(literalStaticNode));
         }
 
         // Handle star imports
@@ -501,8 +502,9 @@ public final class JavaAstVisitor extends JavaLanguageParserBaseVisitor<DetailAs
         ctx.arrayDeclarator().forEach(child -> type.addChild(visit(child)));
 
         // If this is an assignment statement, ASSIGN becomes the parent of EXPR
-        if (ctx.ASSIGN() != null) {
-            final DetailAstImpl assign = create(ctx.ASSIGN());
+        final TerminalNode assignNode = ctx.ASSIGN();
+        if (assignNode != null) {
+            final DetailAstImpl assign = create(assignNode);
             variableDef.addChild(assign);
             assign.addChild(visit(ctx.variableInitializer()));
         }
@@ -1588,8 +1590,9 @@ public final class JavaAstVisitor extends JavaLanguageParserBaseVisitor<DetailAs
      */
     private DetailAstImpl getInnerBopAst(JavaLanguageParser.BinOpContext descendant) {
         final DetailAstImpl innerBop = create(descendant.bop);
-        if (!(descendant.expr(0) instanceof JavaLanguageParser.BinOpContext)) {
-            innerBop.addChild(visit(descendant.expr(0)));
+        final JavaLanguageParser.ExprContext expr = descendant.expr(0);
+        if (!(expr instanceof JavaLanguageParser.BinOpContext)) {
+            innerBop.addChild(visit(expr));
         }
         innerBop.addChild(visit(descendant.expr(1)));
         return innerBop;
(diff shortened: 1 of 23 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -678,10 +678,13 @@ private static String getSummarySentence(DetailNode ast) {
                     && ALLOWED_TYPES.get(child.getType())) {
                 result.append(child.getText());
             }
-            else if (child.getType() == JavadocTokenTypes.HTML_ELEMENT
-                    && CommonUtil.isBlank(result.toString().trim())) {
```
pbludov, line 682: No need to call `trim` before `isBlank`.

src/main/java/com/puppycrawl/tools/checkstyle/checks/SuppressWarningsHolder.java
```diff
@@ -234,12 +234,12 @@ public static boolean isSuppressed(AuditEvent event) {
         for (Entry entry : entries) {
             final boolean afterStart = isSuppressedAfterEventStart(line, column, entry);
             final boolean beforeEnd = isSuppressedBeforeEventEnd(line, column, entry);
+            final String checkName = entry.getCheckName();
             final boolean nameMatches =
-                ALL_WARNING_MATCHING_ID.equals(entry.getCheckName())
-                    || entry.getCheckName().equalsIgnoreCase(checkAlias);
-            final boolean idMatches = event.getModuleId() != null
```
pbludov, line 240: `event.getModuleId() != null` is unnecessary if we switch the parts.

rnveach (approved):
I think the changes for this check are a little weird though.

src/main/java/com/puppycrawl/tools/checkstyle/JavaAstVisitor.java
```diff
@@ -1769,12 +1772,15 @@ public DetailAstImpl visitArrayCreatorRest(JavaLanguageParser.ArrayCreatorRestCo
             if (ctx.children.get(i) == ctx.RBRACK()) {
                 arrayDeclarator.addChild(create(ctx.RBRACK()));
             }
-            else if (ctx.children.get(i) == ctx.expression()) {
-                // Handle '[8]', etc.
-                arrayDeclarator.addChild(visit(ctx.expression()));
-            }
             else {
-                addLastSibling(arrayDeclarator, visit(ctx.children.get(i)));
+                final JavaLanguageParser.ExpressionContext expression = ctx.expression();
```
nick-mancuso, line 1776: Can we just extract `final ParseTree child = ctx.children.get(i);` before any logic in the for loop, and use `child` in all conditions? I would like to avoid nested if/else where possible.

src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java
```diff
@@ -181,15 +181,18 @@ public static DetailAST appendHiddenCommentNodes(DetailAST root) {
...
-        if (lastNode != null && ((DetailAstImpl) lastNode).getHiddenAfter() != null) {
-            DetailAST currentSibling = lastNode;
-            for (Token token: ((DetailAstImpl) lastNode).getHiddenAfter()) {
-                final DetailAST newCommentNode =
-                        createCommentAstFromToken((CommonToken) token);
+        if (lastNode != null) {
+            final List<Token> hiddenAfter = ((DetailAstImpl) lastNode).getHiddenAfter();
+            if (hiddenAfter != null) {
+                DetailAST currentSibling = lastNode;
+                for (Token token : hiddenAfter) {
```
nick-mancuso, line 188: Not a big fan of this change, but I really can't think of a better way to do this without a bunch of other refactoring.

src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractFileSetCheck.java
```diff
@@ -79,8 +80,9 @@ public void beginProcessing(String charset) {
     @Override
     public final SortedSet<Violation> process(File file, FileText fileText)
             throws CheckstyleException {
-        final SortedSet<Violation> violations = context.get().violations;
-        context.get().fileContents = new FileContents(fileText);
+        final FileContext fileContext = context.get();
+        final Set<Violation> violations = fileContext.violations;
```
nick-mancuso, line 84: Why do we not need `SortedSet` anymore?

nick-mancuso (changes requested):
Why didn't CircleCI run in this PR?

Minor items:

src/main/java/com/puppycrawl/tools/checkstyle/JavaAstVisitor.java
pbludov: done

src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractFileSetCheck.java
pbludov: fixed

src/main/java/com/puppycrawl/tools/checkstyle/JavaAstVisitor.java
nick-mancuso: Resolved.

src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java
nick-mancuso: Resolved

src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractFileSetCheck.java
nick-mancuso: Resolved

nick-mancuso (approved):

romani (approved):

Issue #11604, opened 2022-05-04
Remove some hardcoding of specifics so Checkstyle can be more Extensible


rnveach:
I am tired of Checkstyle limiting itself to mainly Java files. I am looking for checkstyle to support more different type of unique files (XML, Properties, etc..) at https://github.com/rnveach/checkstyle-extras .

In the process of setting this process up and starting it, it started to become apparent how limited Checkstyle was making itself. Configuration isn't really the issue, as it already supports 3rd party classes. But there are specific other newer additions, that are making this process more cumbersome.

The following (so far) is a list I have identified and looking to have loosen up so Checkstyle can support more.
* Tests only know `Checker` or `TreeWalker` and need some minor re-working to be able to override for newer "walkers ( https://github.com/checkstyle/checkstyle/pull/11603 )
* Suppressions is oddly coded for `Checker` or `TreeWalker` and has no idea of flexibility like configurations do.
* Even some xpath utilities are very `TreeWalker` specific ( https://github.com/checkstyle/checkstyle/pull/11598 )
* After all this, common functionality should be `abstract`ed out, to separate core functionality and walker specific functionality. This includes walkers, checks, xpath nodes, events, etc....

I am looking for these areas to change, not functionally, but in a way that makes way for extensibility by either Checkstyle itself or 3rd parties.

romani:
I have worries to become a base library for some other tools/project. They will force us to have too abstract model extend our contract with others. 

I would rather let all that tools be less dependent on us and be able to resolve their problems without forcing us to extend API.

I ok with some changes, but let's not think that we need to be friendly to others. We will break all contracts and move in direction that we need.

Our history shows us that we do not have time to think and work on our API strategy :( .

PR #11617, opened 2022-05-08
Issue #11604: separates detail ast from xpath

rnveach:
Issue #11604

This separates DetailAST from xpath so that more of its base classes can be expanded upon and make use of xpath for other ASTs without re-inventing everything. Not much was changed besides splitting the classes in half.

Import control was added to enforce this in the future.


src/main/java/com/puppycrawl/tools/checkstyle/api/CheckstyleAST.java
```diff
@@ -0,0 +1,25 @@
...
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.puppycrawl.tools.checkstyle.api;
+
+/** Base interface used for all ASTs of Checkstyle. */
+public interface CheckstyleAST {
```
rnveach, line 23: XPath knows about the underlying AST, so to be able to provide this method without just using `Object` this class was created and all ASTs implement it.

I was originally planning for this to be just `AST` but PMD complained it was too short. I am open to other names.

Right now it is completely empty, but it will have a future purpose for other common areas that fully rely on the specific ASTs provided. The next target area will be the `AstTreeStringPrinter` and making the code agnostic that it doesn't care what AST is given to it.

Future planned methods here will be: print, get line, get column, get type. Some of this when it comes will be breaking changes.

rnveach:
CI is expected to fail since it references the issue which has not been approved.

src/main/java/com/puppycrawl/tools/checkstyle/api/CheckstyleAST.java
nick-mancuso: I would prefer to not do `Checkstyle’ , what about  `AbstractSyntaxTree`?

rnveach: I have no complaints.

rnveach:
@nick-mancuso Let me know if there is anything else. I renamed the one class.

src/main/java/com/puppycrawl/tools/checkstyle/xpath/AbstractElementNode.java
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    /**
+     * Returns following siblings of the current node.
+     *
+     * @return siblings
+     */
+    private List<AbstractNode> getFollowingSiblings() {
+        final List<AbstractNode> siblings = parent.getChildren();
+        return siblings.subList(indexAmongSiblings + 1, siblings.size());
```
nick-mancuso, line 313: Can we return an unmodifiable list here, and below?

src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractSyntaxTree.java
```diff
@@ -0,0 +1,29 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+package com.puppycrawl.tools.checkstyle.api;
+
+/**
+ * Base interface used for all ASTs of Checkstyle.
+ *
+ * @noinspection MarkerInterface
+ */
+public interface AbstractSyntaxTree {
```
nick-mancuso, line 27: I feel that we should have a concrete plan in place with ticket(s) created before we do this.

nick-mancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractSyntaxTree.java
nick-mancuso: In a perfect world, we should be able to put all methods for getting type, line and column number, iterating through children, etc., in here. But, this would be a significant breaking change.

rnveach: This PR will really not work without this interface as we need to be able to provide the underlying node from the abstracts to the implementations. The only other way would be for the methods to return `Object` and that would require more re-casting in places.

> getting type, line and column number, iterating through children, etc., in here.

This is the plan right now. I mentioned `AstTreeStringPrinter` before and that would be one of the main target areas for this interface. After that, we could simplify both our Walkers (Java and Javadoc) since then they could iterate through different nodes the same way.

All of this future plans really falls under https://github.com/checkstyle/checkstyle/issues/11604  for this interface and what it will do.

> this would be a significant breaking change.

It depends on how we go about this. The current plan is to avoid breaking changes right now by having different method names and deprecating old ways. Eventually, it will lead to a break, but I plan to avoid that as much as I can.

src/main/java/com/puppycrawl/tools/checkstyle/xpath/AbstractElementNode.java
rnveach: I was for this change, but the more I look at it, the more it isn't needed and this is because of how we use this method.

`getFollowingSiblings` is a `private` method and is only called by `getFollowingSiblingsIterator`. It usage inside `getFollowingSiblingsIterator` is to just convert it to an array. We would gain nothing by going through the extra step of making sure the list cannot be modified.

`getPrecedingSiblings` is similar, but it isn't converted to an array. It is passed to `ReverseListIterator`.

I can change that one as well, but I see the gain as very minimal since this is openly used in places.

nick-mancuso: Sounds reasonable to just change `getPrecedingSiblings`, but it's up to you.

src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractSyntaxTree.java
rnveach: > Let's at least explicitly mention this new interface, plans for it, and related details in https://github.com/checkstyle/checkstyle/issues/11604.

I added some of this. Let me know if I am missing anything.

src/main/java/com/puppycrawl/tools/checkstyle/xpath/AbstractElementNode.java
nick-mancuso: Resolved.

src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractSyntaxTree.java
nick-mancuso: Resolved

nick-mancuso (approved):

rnveach:
@romani You have any thoughts on this PR?

romani (changes requested):
Yes, let me block it to think twice on API changes

strkkk (approved):

pbludov (changes requested):
Please fix Javadoc

src/main/java/com/puppycrawl/tools/checkstyle/xpath/AbstractElementNode.java
```diff
@@ -0,0 +1,349 @@
...
+     *
+     * @return depth
+     */
+    @Override
+    public int getDepth() {
+        return depth;
+    }
+
+    /**
+     * Returns attribute value. Throws {@code UnsupportedOperationException} in case,
```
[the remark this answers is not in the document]
rnveach: Done.

pbludov (approved):

romani:
Github, rebase

rnveach:
@romani We changed checker to not fail, so git diff is not running because it is only set to run on failure.
Differences are still being reported.
https://github.com/checkstyle/checkstyle/actions/runs/3256078941/jobs/5346075596#step:7:15

rnveach:
Old Pitest run:
https://github.com/checkstyle/checkstyle/actions/runs/3256078941/jobs/5346075596
4 unnecessary mutations

Change file and remove suppressions.

Previous Pitest run:
https://github.com/checkstyle/checkstyle/actions/runs/3261726684 (all pass)

Current Pitest run:
https://github.com/checkstyle/checkstyle/actions/runs/3261918387/jobs/5357869802
+4 new surviving mutations

We have an unstable mutation here. I changed no other code then trying to get checker to pass. I will suppress and mark as unstable.

romani:
@rnveach , please rebase

rnveach:
Done.

romani:
@rnveach , please resolve conflict and change commit message to reference PR not a umbrella issue.

src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractSyntaxTree.java
romani: > The only other way would be for the methods to return Object and that would require more re-casting in places.

I was about to suggest this ....
root reason of this update is not very beneficial for Checkstyle itself, BUT we create new contract(interface) with unclear plan. direct users might start to use it and it will complicate their life. BUT this update is not for checkstyle (java) users.

how much extra casts we would need ? 

I would rather "harm" a bit implementation (not a interface), to let some project to reuse us, but not demand any new API for them, as we will not support any not java languages. We can not support even single language reliably, so support of some other is unrealistic. So no extension on api for them, we need our api to work for us first of all. All plans to make api that will work for all AST based validations is not very realistic.

romani: can you make one more commit, to show what it will be cost to have Object.

rnveach: > can you make one more commit to show what it will be cost to have Object ?

Done. c692c6835aced11f99d64fe7a77661048d8f4693

romani:
I don't see any bad casting. If no objections, let's squash commits.

rnveach:
Squashed.

> I don't see any bad casting

Casting is already in place because the new interface didn't have any methods, so everything needed to be casted to the actual class anyways.

rnveach:
Unrelated error to my PR:
````
No apt package "xmlstarlet", but there is a snap with that name.

Try "snap install xmlstarlet"

apt-get.diagnostics

apt-get install failed

The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install groovy xsltproc xmlstarlet" failed and exited with 100 during .
````

File: config/import-control.xml [hunk 5]
```diff
@@ -240,6 +240,13 @@
     <allow pkg="com.puppycrawl.tools.checkstyle.xpath.iterators" local-only="true"/>
     <allow class="com.puppycrawl.tools.checkstyle.TreeWalkerAuditEvent" local-only="true"/>
 
+    <file name="Abstract.*Node" regex="true">
+      <allow class="com.puppycrawl.tools.checkstyle.api.AbstractSyntaxTree"/>
```
romani, line 244: remove this old line.
also please make `<file name="Abstract(Element|Root)Node" regex="true">`

allowance should be strict to catch any new classes.

src/main/java/com/puppycrawl/tools/checkstyle/xpath/RootNode.java
```diff
@@ -232,27 +99,8 @@ public int getTokenType() {
      * @return underlying node
      */
     @Override
-    public DetailAST getUnderlyingNode() {
+    public Object getUnderlyingNode() {
```
romani, line 102: lets keep more exact type
https://stackoverflow.com/a/14694885

```
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/xpath/ElementNode.java b/src/main/java/com/puppycrawl/tools/checkstyle/xpath/ElementNode.java
index 880210d..10415e6 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/xpath/ElementNode.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/xpath/ElementNode.java
@@ -115,7 +115,7 @@ public class ElementNode extends AbstractElementNode {
      * @return underlying node
      */
     @Override
-    public Object getUnderlyingNode() {
+    public DetailAST getUnderlyingNode() {
         return detailAst;
     }
 
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/xpath/RootNode.java b/src/main/java/com/puppycrawl/tools/checkstyle/xpath/RootNode.java
index 09b3bad..7d61109 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/xpath/RootNode.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/xpath/RootNode.java
@@ -99,7 +99,7 @@ public class RootNode extends AbstractRootNode {
      * @return underlying node
      */
     @Override
-    public Object getUnderlyingNode() {
+    public DetailAST getUnderlyingNode() {
         return detailAst;
     }

```

works fine on my local

romani (changes requested):
last:

src/main/java/com/puppycrawl/tools/checkstyle/xpath/RootNode.java
rnveach: We cannot do this. This goes against the whole purpose of this PR which is to separate XPath from a specific AST. If you look at the original code before my PR, it was already using `DetailAST`. If you wanted to expand xpath to Javadoc node, having Java node classes listed will cause issues.

config/import-control.xml
rnveach: Removed old line.

The purpose of `.*` was to catch any future nodes added and not be specific to specific classes. We are already limiting this to the xpath package. I can change `.*` to `\w` but let me know if you still want this fully removed.

config/import-control.xml
```diff
@@ -240,6 +240,12 @@
     <allow pkg="com.puppycrawl.tools.checkstyle.xpath.iterators" local-only="true"/>
     <allow class="com.puppycrawl.tools.checkstyle.TreeWalkerAuditEvent" local-only="true"/>
 
+    <file name="Abstract.*Node" regex="true">
+      <allow pkg="com.puppycrawl.tools.checkstyle.xpath.iterators"/>
+      <!-- These are more saxon utilities, and should not have anything checkstyle specific -->
```
romani, line 245: >These are more saxon utilities

I am not sure what you mean by this. Rephrase or remove

romani (changes requested):
Items

config/import-control.xml
romani: Ok

config/import-control.xml
rnveach: See same comment further down under `<subpackage name="iterators">`.

This was to denote classes in this package are utilities for saxon only and should not have anything Checkstyle specific, like `DetailAST`.

I duplicated the same note because this was for the same purpose.

romani: Now I see root of confusion.
In bellow block
```
<subpackage name="iterators">
      <!-- These are more saxon utilities, and should not have anything checkstyle specific -->
      <disallow pkg=".....
```

Comment is for whole `subpackage`. It makes sense. I don't know what is most good approach to place comments: above tag or inside tag as first line.

In new `file` tag, comment is placed not on top , so intent to describe sub tag only, not a parent tag.
Please move it at top at least.

rnveach: Done.

romani (approved):
ok to merge after squash

PR #11618, opened 2022-05-08
minor: removes complicated uses of test tier that are not needed

rnveach:
Went through the test tier to see why certain methods are still protected and not final. Found some instances were the methods were used in complex ways that weren't needed. Like manually creating the Checker when the verify method will do this for us, etc...


src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/FileTabCharacterCheckTest.java
```diff
@@ -64,21 +61,4 @@ public void testVerbose() throws Exception {
             expected);
     }
 
-    @Test
-    public void testBadFile() throws Exception {
```
rnveach, line 68: BDD was complaining about this. It was removed because it was not checking anything against the check. The error message is coming from Checker and does not even get to the check. The check does rely on external files.

romani (approved):
ok to merge

PR #11623, opened 2022-05-09
Issue #11214: Specify violation messages for ImportControlCheck

Shubhamxshah:
Issue #11214: Specify violation messages for ImportControlCheck


Rahulkhinchi03:
@romani @nick-mancuso 

Why can't we just use import javax.swing.border.*; `// violation 'Disallowed import'`

instead of:
import javax.swing.border.*; // violation `'Disallowed import - javax.swing.border.*'`

It is good for readability and it is understandable that if we are adding a disallowed import violation on some particular import then it will be a violation. 

`Does it really need to have the full name of the import in the comment?`

nick-mancuso:
@Rahulkhinchi03 unfortunately, bugs are not usually well behaved. This is the advantage of BDD - violations are more nuanced than `violation` or `ok`. We want to verify the actual behavior of the check. If we displayed the name of another import on the wrong line, a simple `violation - 'Disallowed import` as you propose above would never catch this.

This would also be a step backwards, since we had previously verified violation messages in old check tests, and this issue restores that functionality.

Shubhamxshah:
does it needs any changes?

Issue #11584, opened 2022-04-30
WriteTag reports violation with confusing message when there is no javadoc


strkkk:
https://checkstyle.org/config_javadoc.html#WriteTag
```
$ cat Test.java
public class Test {}

$ cat config.xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="TreeWalker">
        <module name="WriteTag"/>
    </module>
</module>

$ java -jar checkstyle-10.2-all.jar -c config.xml Test.java
Starting audit...
[ERROR] C:\workdir\Test.java:1: Type Javadoc comment is missing null tag. [WriteTag]
Audit done.
Checkstyle ends with 1 errors.
```

Expected: no violation, since there is no javadoc.
Also note confusing violation message ("null" tag).

For javadoc presence there are other checks. And docs mention that with default config it does nothing, but it reports weird message instead.

P.S. Checkstyle has `WriteTag` with default config - 
https://github.com/checkstyle/checkstyle/blob/4ad253653d59d77002ae2f61349624dcb18dcc38/config/checkstyle_checks.xml#L653

--------

**Breaking compatibility migration notes:**
after fix and release (10.21.0) Check will not report violations on missed Javadocs.
To archive behavior as before, users should start using [InvalidJavadocPosition](https://checkstyle.org/checks/javadoc/invalidjavadocposition.html), [MissingJavadocMethod](https://checkstyle.org/checks/javadoc/missingjavadocmethod.html) , [MissingJavadocPackage](https://checkstyle.org/checks/javadoc/missingjavadocpackage.html) , [MissingJavadocType](https://checkstyle.org/checks/javadoc/missingjavadoctype.html) to let Checkstyle demand javadoc first and when WriteTag will  demand required tag in existing javadoc

theWolf1198:
I want to work on this, however, I have a question:

This seems like an issue with a confusing message that is printed when there is no default tag, do you want me to refactor the output log message or change the check? 

Thanks!

strkkk:
@theWolf1198 I expected check to do not report violation if there is no javadoc

theWolf1198:
@romani @strkkk, I think that the code should allow empty Javadoc for default WriteTag as there are other checks to check if Javadoc is present or not. 
[...]

PR #11542, opened 2022-04-12
Issue #7576: update doc for SuppressWarningsHolder

Kevin222004:
Fixes Issue #7576 
Screenshot from 2022-04-12 14-18-06

PR #11638, opened 2022-05-15
Issue #11584: Adding changes to WriteTag check to solve null javadoc issue

theWolf1198:
If WriteTag is configured with default settings and there is no Javadoc to check, there used to be a violation with an incorrect error message, hence attempting to solve. 

1. Added check to proceed to writeTag verification only when Javadoc comment is not null. 
2. Added two supplementary checks to throw violation if writeTag configuration has a tag or tagFormat setting and Javadoc comment is null. 
3. Added two unit tests to check on the aforementioned cases and complete code coverage to 100%


Kevin222004:
> Solving issue #11542
Change the number of pr in pr description

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/WriteTagCheck.java
```diff
@@ -283,12 +283,16 @@ public void visitToken(DetailAST ast) {
         final int lineNo = ast.getLineNo();
         final TextBlock cmt =
             contents.getJavadocBefore(lineNo);
-        if (cmt == null) {
+        if (cmt != null) {
+            checkTag(lineNo, cmt.getText());
+        }
+        else if (tag != null) {
             log(lineNo, MSG_MISSING_TAG, tag);
```
strkkk, line 290: this does not solve issue. There should be no message in case there is no javadoc, regardless tags and format

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/WriteTagCheckTest.java
```diff
@@ -206,8 +206,26 @@ public void testEnumsAndAnnotations() throws Exception {
...
+    }
+
+    @Test
+    public void testNoJavadocsWithRequiredTag() throws Exception {
+        final DefaultConfiguration checkConfig = createModuleConfig(WriteTagCheck.class);
+        checkConfig.addProperty("tag", "@incomplete");
+        final String[] expected = {
+            "13: " + getCheckMessage(MSG_MISSING_TAG, "@incomplete"),
+        };
+        verify(checkConfig, getPath("InputWriteTagNoJavadoc.java"), expected);
```
strkkk, line 220: please make a separate test inputs for each test

strkkk (changes requested):
items

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/WriteTagCheckTest.java
theWolf1198: Added separate files for the TCs

romani:
Please keep single commit in PR always. Please fix all CI failures.

strkkk (changes requested):
pls fix ci issues, squash commits into one

Issue #11277, opened 2022-02-03
update code base to have javadoc tag to explain noinspection tag content


romani:
detected at https://github.com/checkstyle/checkstyle/pull/11206#discussion_r797996787

code:
https://github.com/checkstyle/checkstyle/blob/d4228cc1c04f6e18392948986b605cf87738c951/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L628-L635

has problems that such technical excuse of to suppress IDEA inspection is rendered in final javadoc and can confuse users.
we try to keep reasons of suppressions but in same time we do not want to expose it to users.

proposed solution is too use custom tag for this:
```
$ git diff
diff --git a/pom.xml b/pom.xml
index 31c2853..65fddda 100644
--- a/pom.xml
+++ b/pom.xml
@@ -433,6 +433,10 @@
                 <name>noinspection</name>
                 <placement>X</placement>
               </tag>
+              <tag>
+                <name>noinspectionreason</name>
+                <placement>X</placement>
+              </tag>
             </tags>
           </configuration>
         </plugin>
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/Main.java b/src/main/java/com/puppycrawl/tools/checkstyle/Main.java
index fec6cbd..f216a44 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Main.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Main.java
@@ -627,8 +627,10 @@ public final class Main {
 
         /**
          * The checker threads number.
-         * Suppression: CanBeFinal - we use picocli and it use  reflection to manage such fields
-         * This option has been skipped for CLI options intentionally.
+         *
+         * @noinspectionreason CanBeFinal - we use picocli and it use
+         *     reflection to manage such fields
+         *     This option has been skipped for CLI options intentionally.
          *
          * @noinspection CanBeFinal
          */
diff --git a/config/intellij-idea-inspections.xml b/config/intellij-idea-inspections.xml
index de97b29..35a1ab0 100644
--- a/config/intellij-idea-inspections.xml
+++ b/config/intellij-idea-inspections.xml
@@ -2797,7 +2797,7 @@
     <option name="IGNORE_JAVADOC_PERIOD" value="true"/>
     <option name="IGNORE_DUPLICATED_THROWS" value="false"/>
     <option name="IGNORE_POINT_TO_ITSELF" value="false"/>
[...]

PR #11639, opened 2022-05-15
Issue #11566: resolve suppression of DeprecatedIsStillUsed

romani:
Issue #11566

based on feedback - https://youtrack.jetbrains.com/issue/IDEA-293684

### DIFF
--- a/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionEmptyBlockTest.java
+++ b/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionEmptyBlockTest.java
@@ -42,7 +42,7 @@ public class XpathRegressionEmptyBlockTest extends AbstractXpathTestSupport {
                 new File(getPath("SuppressionXpathRegressionEmptyBlockEmpty.java"));
         final DefaultConfiguration moduleConfig =
                 createModuleConfig(EmptyBlockCheck.class);
-        moduleConfig.addAttribute("option", "TEXT");
+        moduleConfig.addProperty("option", "TEXT");
         final String[] expectedViolation = {
             "5:38: " + getCheckMessage(EmptyBlockCheck.class,
                 EmptyBlockCheck.MSG_KEY_BLOCK_EMPTY, "for"),
--- a/src/test/java/com/puppycrawl/tools/checkstyle/api/FileContentsTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/api/FileContentsTest.java
@@ -255,6 +255,8 @@ public class FileContentsTest {
                 .isEqualTo(new Comment(new String[] {"/** *"}, 2, 1, 6).toString());
     }
 
+    // until https://github.com/checkstyle/checkstyle/issues/11723
+    @Deprecated(since = "10.2")
     @Test
     public void testInPackageInfo() {
         final FileContents fileContents = new FileContents(new FileText(
@@ -266,6 +268,8 @@ public class FileContentsTest {
                 .isTrue();
     }
 
+    // until https://github.com/checkstyle/checkstyle/issues/11723
+    @Deprecated(since = "10.2")
     @Test
     public void testNotInPackageInfo() {
         final FileContents fileContents = new FileContents(new FileText(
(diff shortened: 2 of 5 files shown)


strkkk:
@romani as for me, deprecation of the test methods looks weird and confusing. Current way with annotation suppression seems better.

romani:
Javadoc suppression lacking reason of suppression. So suppression is weird.
By placing annotation to test we making our main code better (less noise) to test code. Chance to read this test is significantly less than reading main code. So lets polite test code instead of main code.

strkkk:
this annotation means that annotated method should not be used. Test methods are never used outside of test framework, so this apporach seems weird.
> Javadoc suppression lacking reason of suppression

why not just add explanation?
like `@noinspection DeprecatedIsStillUsed - usages are in unit tests?`
deprecation of unit test method is on same level of weirdness (and there is also no explanation), and main code contains much more other suppressions that are more cryptic than this one

romani:
>why not just add explanation?

It causes another violation as this tag is not block tag, ...

strkkk:
> It causes another violation as this tag is not block tag

I think this is the issue here.

@rnveach @pbludov @nick-mancuso please share your opinion - is ot ok to use Deprecated annotation on test methods or better to add some comment?

pbludov:
I like the idea of new tag to add some explanation right in the Javadoc
Details are here: #11277

rnveach:
Deprecation tag is required for tests because a deprecated method is used in them.

I can understand marking tests seems weird. It looks like we don't have this issue in main code for things like `AbstractCheck#getFileContents` as we use `@SuppressWarnings("deprecation")`. It makes sense there since we don't want to cascade and mark everything deprecated since we can't get away from using our deprecated method.

I am fine with either marking tests as deprecated or using some suppression like `SuppressWarnings` or `noinspectionreason`.

romani:
>think noinspectionreason may be redundant as what exactly will we put in a reason for it?

Something like"Until https://youtrack.jetbrains.com/issue/IDEA-293684" :)

romani:
@strkkk , I created issue to remove deprecated methods and placed them in code and reason for deprecation.

strkkk (approved):

romani:
GitHub, rebase

nick-mancuso (approved):
Better to have annotation on test code than in main code.

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionEmptyBlockTest.java
```diff
@@ -36,6 +36,8 @@ protected String getCheckName() {
         return checkName;
     }
 
+    // until https://github.com/checkstyle/checkstyle/issues/11722
+    @Deprecated(since = "10.2")
```
rnveach, line 40: As I am relooking at this, what is actually deprecated?
I am seeing `addAttribute` as the only thing deprecated and it is our code and our test. This is even xpath tests which aren't really use for code coverage or pitest. Why don't we swap this out for `addProperty`?

This might be a good example to describe what is deprecated with these annotations.

src/test/java/com/puppycrawl/tools/checkstyle/api/FileContentsTest.java
```diff
@@ -266,6 +268,8 @@ public void testInPackageInfo() {
                 .isTrue();
     }
 
+    // until https://github.com/checkstyle/checkstyle/issues/11723
+    @Deprecated(since = "10.2")
     @Test
     public void testNotInPackageInfo() {
```
rnveach, line 274: Should be ok as well.

rnveach (changes requested):

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionEmptyBlockTest.java
romani: awesome catch, fixed.

romani:
@rnveach , do you think link to issue is not good enough ? Should I extend issue description ?

rnveach:
Link to issue is good as long as we are sure to require it, similar to noinspectionreason.

PR #11644, opened 2022-05-15
Issue #11604: allow 3rd party Check providers to group modules under custom parent module

rnveach:
Subset of Issue #11604

Our current module identification process is very strict. This PR is to allow that to be loosened so 3rd parties can create their own sub-modules that our current one doesn't support.

Suppose I had created a new walker `XmlWalker` which extends `AbstractFileSetCheck` and I wanted it to allow children modules like `AbstractXmlCheck`. Checkstyle will recognize `XmlWalker` because of `AbstractFileSetCheck` but it will not identify anything extending `AbstractXmlCheck` even if follows the absolute same makeup as `AbstractCheck`. This is because we only recognize our abstracts (`AbstractFileSetCheck`, `AbstractCheck`, etc...) and nothing else that could be a child module. There is no common class for module children. Our packages.xml file will not help here, because again, we still only pull in things that follow our defined class hierarchy.

With this change we will now be able to pick up these new children, as long as they follow basic rules defined by our `isCheckstyleModule`, and treat them like they are any other module we have for Checkstyle, as long as the parent module will accept them. `isCheckstyleModule` defines things as: must extend `AutomaticBean`, can't be abstract, must have a default constructor, and must not be an internal module to us.

Example of project that has custom children modules: https://github.com/rnveach/checkstyle-extras


rnveach:
@romani You have any thoughts on this PR?

romani (changes requested):
Let me find time to look at it closely

rnveach:
@romani ping

rnveach:
CI is passing here.
https://github.com/checkstyle/checkstyle/actions/runs/3262167551#artifacts
No reported checker differences.

romani:
@rnveach, 
I do not like our override of static code.
It is static only because now it can be, but it doesn't mean that we can't convert to instantiated code.

Can we provide identification function to ctor of https://github.com/checkstyle/checkstyle/blob/a8f065738de6c4ff917a2ee57521f6cf8718f2ba/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java#L122

Or some other place.

rnveach:
The best way to make this non-static is to remove the logic completely of what classes we pull in and move this to the respective parents to handle.
If I call createModule with Joe, then it will return a class it finds for Joe even if it doesn't fit what our modules are normally. It is up to Checker/TreeWalker/etc to determine if that is an acceptable child or not
The only thing we will still force for a class createModule returns is ModuleReflectionUtil#isValidCheckstyleClass. That is: extends AutomaticBean, not abstract, default constructor, and not our XpathFileGenerator
So, for example, if createModule returns an AbstractCheck and Checker is the caller, it would up to Checker to determine if it will accept the class type is returns

src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
```diff
@@ -303,7 +303,7 @@ Definitions.CHECKSTYLE_BUNDLE, getClass(),
     private Map<String, Set<String>> generateThirdPartyNameToFullModuleName(ClassLoader loader) {
         Map<String, Set<String>> returnValue;
         try {
-            returnValue = ModuleReflectionUtil.getCheckstyleModules(packages, loader).stream()
+            returnValue = ModuleReflectionUtil.getValidCheckstyleClasses(packages, loader).stream()
```
rnveach, line 306: This is needed so we can keep the cache of `thirdPartyNameToFullModuleNames`. We need someone to pull in classes to put into the cache, and removing it completely will just load all classes in the classpath. `getCheckstyleModules` was always the method that was making this restrictive.

Passing a Predicate like originally suggested will not assist, one because of the cache. The cache would need the predicate to know what should be populated. However, each walker needs to pass its own predicate which would complicate caching.
Another reason a Predicate will make things more complex is we are currently throwing exceptions if we find multiple classes with the same name. So we would have to add more logic and loops to determine if a Predicate would help any, while breaking some compatibility to allow the Predicate to be given to this class.
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java#L280

Checker and TreeWalker is already ensuring it is only dealing with the children it wants, so this is why more code isn't needed.
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L483
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java#L131

rnveach:
Side note, this new change is what I would consider very simple but archunit changed pretty drastically.
https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=11672&view=logs&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111

src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
romani: Solution is approved, please finalize code changes.
Makes sense to create issue with description of this comment.

Do we need isCheckstyleModule method?

rnveach: `isCheckstyleModule` is still needed for other utils and satisfies the check preventing methods from being too big.

rnveach:
https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=11789&view=logs&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111

I think something is preventing this from passing. I keep copying but it keeps failing. I'll try once more.

src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
romani: I see only single usage (not in tests):
image

please double check your side.
I think we should remove this method.

src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java
```diff
@@ -46,21 +46,21 @@ private ModuleReflectionUtil() {
...
      *
      * @param packages the collection of package names to use
      * @param loader the class loader used to load Checkstyle package names
-     * @return the set of checkstyle's module classes
+     * @return the set of checkstyle's classes
      * @throws IOException if the attempt to read class path resources failed
      * @see #isCheckstyleModule(Class)
      */
-    public static Set<Class<?>> getCheckstyleModules(
+    public static Set<Class<?>> getValidCheckstyleClasses(
```
romani, line 57: I would keep a name getCheckstyleModules and isCheckstyleModule.

getValidCheckstyleClasses and isValidCheckstyleClass are a bit weird names.

romani (changes requested):
item:

src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java
rnveach: What do you want `isValidCheckstyleClass` named? I still think we should keep `isCheckstyleModule`.

src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
romani: I pushed commit with removal of this method.
I still do not understand why we need two different methods to load modules.

So now, can we rename getValidCheckstyleClasses to getCheckstyleModules ?

src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java
romani: https://github.com/checkstyle/checkstyle/pull/11644#discussion_r1008018002

src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
rnveach: My main reason against removal is because `CheckUtil` is our go-to place to get all actual checkstyle modules in the repo to use for documentation and other "what are our checks area".

We are changing this method from `getCheckstyleModules` to `getValidCheckstyleClasses`. So for a trimmed example, instead of getting all filesets we are now changing it to get everything that looks like a module, but we have not confirmed ourselves if it is. So we are possibly getting more classes that are not our modules. We would have to change this method some to ensure we are only reporting our modules in 

That said, these test method is trying to become more open to others repos like sevntu to be used there and maybe it could be used by checkstyle-extras for its own modules. I have an open issue on this, but sevntu is stuck because of eclipse-cs. I am surprised that the test change did not cause any failures.

I want to look at more in my local, but i think I will change my mind and make the change.

romani: >but we have not confirmed ourselves if it is

If we are not confident, we can not merge this PR.

rnveach: I confirmed locally, it seems the method returns the same list for our repo. So I pushed the update. Please look over and I will ensure the CI passes.

src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java
rnveach: Done.

rnveach:
Not understanding why `codecov` is failing as UI is not reporting any decreases I found.

rnveach:
@romani 
> please update PR description to not mention old ideas 

Done.

> Please use PR number in commit

Done.

> Should this be "new feature" or "bug fix" 

I am thinking new feature, as we never really considered allowing these type of sub-modules. We may have had discussions and code changes implying this, but it was never out right said or probably documented.

romani (approved):
We need to improve title of PR to be more easy to understand in release notes.

Maybe later on update PR description to put link to xml linter project as example.
Maybe good to update some doc page on how to do custom parent module.

nrmancuso (commented):
> Maybe good to update some doc page on how to do custom parent module.

Agreed, if we do all this work at allow Checkstyle to be more extensible, but do not have some documentation/ recommendation/ examples about how to do it, the value of resolving #11604 to users is probably low.

@rnveach can we add a healthy documentation update to tasks in #11604 ?

romani:
I recommend to merge, let @rnveach use it in separate project, and then write doc.

nrmancuso:
> I can add documentation, but what are our thoughts regarding it?

I do not know yet, hopefully experience from https://github.com/rnveach/checkstyle-extras will tell :)

> I recommend to merge, let @rnveach use it in separate project, and then write doc.

@romani see https://github.com/checkstyle/checkstyle/pull/11644#pullrequestreview-1170419122, I would just like to have documentation update task added to list at https://github.com/checkstyle/checkstyle/issues/11604. I agree that doc update should wait until checkstyle-extras is closer to completion.

romani:
@rnveach , please and sub task to big issue and let's move further

nrmancuso (approved):
Nice side effect of "killing" seven mutations too :)

PR #11654, opened 2022-05-18
minor: add a test case in RightCurlyCheckTest

yuchesu:
Adding a new test case to RightCurlyCheckTest to test the SAME policy in right curly option. Thanks for your review and consideration.


rnveach:
@yuchesu Please keep the number of commits to 1 in this PR and look over your changes and the CI failure. Your changes should just be adding the 1 test case as you said, and should be void of any of the indentation and superfluous changes.

Also, where is adding this test case coming from? Was it discussed in another issue/PR?

rnveach:
@yuchesu ping

romani:
Why you think we need extra test, please explain how you found a gap.

nick-mancuso:
We have lost contact with author of PR, and we have no ticket on this.

PR #11657, opened 2022-05-19
A new test case created for BlockTagUtil. Thanks for your review and consideration

13542:
Hi, there.
I wrote a small update for adding a new test case for  ExtractBlockTags in BlockTagUtil. Followed the contribution document, and changes are passed by command mvn clean verify. Thank you for your review and consideration.


rnveach:
@13542 Please look over the CI failure. You can run `mvn verify test` locally to ensure it passes before pushing.

Where is adding this test case coming from? Was it discussed in another issue/PR?

romani (changes requested):
Sorry we do allow any more pure Unit tests. All test should be based on Input files and real Check usage.
All utilities should be tested through real Checks

rnveach:
@13542 ping

nick-mancuso:
We have lost contact with PR author, and we have no ticket related to this change.

Issue #11651, opened 2022-05-18
Rework IT's Test Structure


rnveach:
Continuing on https://github.com/checkstyle/checkstyle/pull/11603 and https://github.com/checkstyle/checkstyle/pull/11585 ,

Google and Sun IT's tests follow a different pattern than XPath and Test. Because of this, both test areas should change their test make up based on their respective testing pattern.

Google and Sun does regression on an existing configuration. They do not create modules from memory or scan existing modules. Their methods should revolve around this.
https://github.com/checkstyle/checkstyle/blob/master/src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java#L41
We should remove reliance on ModuleCreationOption and rework the code to know the hierarchy similar to test as we follow the configuration file.
https://github.com/checkstyle/checkstyle/blob/master/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java#L58
In addition, their tests should be standalone in that they are not pulling information from the configuration files but doing a standalone testing. Right now we pull property values and messages from the configuration, making it dependent on the configuration's internal structure if the configuration returned back changes.
https://github.com/checkstyle/checkstyle/blob/master/src/it/java/com/google/checkstyle/test/chapter4formatting/rule462horizontalwhitespace/GenericWhitespaceTest.java#L42
https://github.com/checkstyle/checkstyle/blob/master/src/it/java/com/google/checkstyle/test/chapter5naming/rule528typevariablenames/RecordTypeParameterNameTest.java#L41

XPath's regression very closely follow's test, so it should extend test's structure.

Similar changes to IT and XPath should be copied over to test like usual.

This will be a multi-work PR.

romani:
I am missing a point to depend less on config, whole idea of such tests is to test config as whole. Users do not care about how much modules we have in config.

rnveach:
@romani See PR https://github.com/checkstyle/checkstyle/pull/11652 to understand this point. Let me know if you could point to where the confusion may be.

Tests should not pull information in from the config to build it's results. They should be more independent to actually test the config and the reason is in a future PR I will change the result of the current `getModuleConfig` and thus a simple `checkConfig.getMessages()` would have to be changed to something like `getChildren()[0].getMessages()`.

> Users do not care about how much modules we have in config.

This whole issue is connected to reworking our tests. No code in `main` is planned to be changed. So our users who use our release won't see any difference.

nick-mancuso:
[...]

PR #11661, opened 2022-05-21
Issue #11651: splits xpath it to be based on test tier

rnveach:
Issue #11651

This starts the split, that IT xpath and checkstyle are different form of tests that will be for google/sun's tests as the latter relies on configurations and these won't. This is needed for a future change which will split google/sun's tests more apart from these.

Since IT xpath and checkstyle have different abstract classes, I am keeping them separate. Test's paths are fully focused on `test` folder path, so IT's version need their own paths.

### DIFF
--- a/src/it/java/org/checkstyle/suppressionxpathfilter/AbstractXpathTestSupport.java
+++ b/src/it/java/org/checkstyle/suppressionxpathfilter/AbstractXpathTestSupport.java
@@ -22,6 +22,7 @@ package org.checkstyle.suppressionxpathfilter;
 import static com.google.common.truth.Truth.assertWithMessage;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -31,9 +32,9 @@ import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import org.checkstyle.base.AbstractCheckstyleModuleTestSupport;
 import org.junit.jupiter.api.io.TempDir;
 
+import com.puppycrawl.tools.checkstyle.AbstractModuleTestSupport;
 import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
 import com.puppycrawl.tools.checkstyle.JavaParser;
 import com.puppycrawl.tools.checkstyle.TreeWalker;
@@ -43,7 +44,7 @@ import com.puppycrawl.tools.checkstyle.filters.SuppressionXpathFilter;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 import com.puppycrawl.tools.checkstyle.xpath.XpathQueryGenerator;
 
-public abstract class AbstractXpathTestSupport extends AbstractCheckstyleModuleTestSupport {
+public abstract class AbstractXpathTestSupport extends AbstractModuleTestSupport {
 
     private static final int DEFAULT_TAB_WIDTH = 4;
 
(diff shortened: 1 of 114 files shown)


src/it/java/org/checkstyle/suppressionxpathfilter/AbstractXpathTestSupport.java
```diff
@@ -210,8 +237,7 @@ protected void runVerifications(DefaultConfiguration moduleConfig,
         treeWalkerConfigWithXpath.addChild(createSuppressionXpathFilter(moduleConfig.getName(),
                 generatedXpathQueries));
 
-        final Integer[] warnList = getLinesWithWarn(fileToProcess.getPath());
-        verify(moduleConfig, fileToProcess.getPath(), expectedViolations, warnList);
+        verify(moduleConfig, fileToProcess.getPath(), expectedViolations);
```
romani, line 240: It use to fail if warn trailing comment is not placed in input file.
Is it going to continue to work ?

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionAbbreviationAsWordInNameTest.java
```diff
@@ -39,7 +39,7 @@ protected String getCheckName() {
 
     @Test
     public void testAnnotation() throws Exception {
-        final File fileToProcess = new File(getPath(
+        final File fileToProcess = new File(getXpathPath(
```
romani, line 42: This a bit weird, test class takes files from two locations: compilable and no compilable.
So no matter what type of test it is (xpath or it or regular) we just need getPath method.
Common base class can override it one for all.

romani (changes requested):
Items

src/it/java/org/checkstyle/suppressionxpathfilter/AbstractXpathTestSupport.java
rnveach: No, this is because normal test doesn't have this feature except in BDD, and xpath can't currently be swapped for bdd.

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionAbbreviationAsWordInNameTest.java
rnveach: We would have change the current methods to be non-final, meaning any class can override them. If thats fine, then I will make the change.

src/it/java/org/checkstyle/suppressionxpathfilter/AbstractXpathTestSupport.java
rnveach: @romani Is this an issue?

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionAbbreviationAsWordInNameTest.java
nick-mancuso: > We would have change the current methods to be non-final...

I would prefer to have another level of abstraction here where we set the path for each type of test(regular, xpath, ...) and then allow the respective tests to extend these new classes only. @rnveach what do you think about this?

src/it/java/org/checkstyle/suppressionxpathfilter/AbstractXpathTestSupport.java
rnveach: @romani ping

romani: I am confused that we still put such "warn" in Input files
https://github.com/checkstyle/checkstyle/blob/b6301c189ef5d2d731a755fa9a2e84aa0677494a/src/it/resources/org/checkstyle/suppressionxpathfilter/annotationusestyle/SuppressionXpathRegressionAnnotationUseStyleEight.java#L3

are they redundant after this commit ?
them use to be a marker to show line for reviewer to understand what is target of xpath. Reading of xpath is not pleasant, so extra hint is good help.

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionAbbreviationAsWordInNameTest.java
romani: I lost a reason of my question. I am recalling it.

@nrmancuso, please recheck you point also. let @rnveach know if this is still issue for you.

romani:
I reassigned this PR back to @rnveach for rebase and answering question, it is old PR so it might take some time.

nrmancuso:
Looks like we lost interest in pursuit of this issue, I am going to close it. We can reopen this PR one day in the future or allow some contributor to reuse the changes here.

PR #11676, opened 2022-05-28
Issue #11569: Fixed the false negative in the rule ClassMemberImplied…

### LINKED ISSUE #11569: A false negative about the rule ClassMemberImpliedModifier
I have read check documentation: https://checkstyle.sourceforge.io/config_modifier.html#ClassMemberImpliedModifier

```bash
/var/tmp $ javac Test.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <module name="TreeWalker">
    <module name="ClassMemberImpliedModifier" />
  </module>
</module>

/var/tmp $ cat Test.java
public class Test {
  Object obj = new Object() {
    public record BadRecord() {}  // should report a warning in this line
  };
}

/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.1-all.jar -c config.xml Test.java
Starting audit...
Audit done.
```
---
---
Hi, I found a false negative about the rule ClassMemberImpliedModifier. CheckStyle should have reported a warning in line 3. But it does not. Thanks for your consideration.
Static modifier is expected to be present.

as23415672:
Fixed #11569 
The problem was due to the original implementation didn't consider the case that the declaration of record could be nested in an anonymous object, and ``ScopeUtil.isInBlockOf`` will break the loop once it encounters a ``new`` keyword. I fixed this issue by adding a method to dertermine whether the node is in an anonymous object.
Diff Regression config: https://gist.githubusercontent.com/as23415672/b6efc853253765e2c3ff2ae12f77b6a5/raw/2610c9ad3fe1b3c1213896f23816169f81d11755/my_check_ClassMemberImpliedModifier.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/ClassMemberImpliedModifierCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/ClassMemberImpliedModifierCheck.java
@@ -243,9 +243,33 @@ public class ClassMemberImpliedModifierCheck
      * @return true if ast is in a class, enum, or record
      */
     private static boolean isInTypeBlock(DetailAST ast) {
-        return ScopeUtil.isInClassBlock(ast)
+        return isInObjBlock(ast)
+                || ScopeUtil.isInClassBlock(ast)
                 || ScopeUtil.isInEnumBlock(ast)
                 || ScopeUtil.isInRecordBlock(ast);
+    }
+
+    /**
+     * Returns whether a node is directly contained within an anonymous object block.
+     *
+     * @param node the node to check if directly contained within an anonymous object block.
+     * @return a {@code boolean} value
+     */
+    public static boolean isInObjBlock(DetailAST node) {
+        boolean returnValue = false;
+
+        // Loop up looking for a containing interface block
+        for (DetailAST token = node.getParent();
+             token != null;
+             token = token.getParent()) {
+            if (token.getType() == TokenTypes.OBJBLOCK
+                && token.getParent().getType() == TokenTypes.LITERAL_NEW) {
+                returnValue = true;
+                break;
+            }
+        }
+
+        return returnValue;
     }
 
 }
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/classmemberimpliedmodifier/InputClassMemberImpliedModifierObjNestedRecord.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/classmemberimpliedmodifier/InputClassMemberImpliedModifierObjNestedRecord.java
@@ -1,0 +1,17 @@
+/*
+ClassMemberImpliedModifier
+violateImpliedStaticOnNestedEnum = (default)true
+violateImpliedStaticOnNestedInterface = (default)true
+violateImpliedStaticOnNestedRecord = (default)true
+
+
+*/
+
+//non-compiled with javac: Compilable with Java14
+package com.puppycrawl.tools.checkstyle.checks.modifier.classmemberimpliedmodifier;
+
+public class InputClassMemberImpliedModifierObjNestedRecord {
+    Object obj = new Object() {
+        public record BadRecord() {} // violation
+    };
+}
--- a/pom.xml
+++ b/pom.xml
@@ -3772,6 +3772,9 @@
                 <param>
                   com.puppycrawl.tools.checkstyle.checks.design.FinalClassCheckTest
                 </param>
+                <param>
+                  com.puppycrawl.tools.checkstyle.checks.modifier.ClassMemberImpliedModifierCheckTest
+                </param>
               </targetTests>
               <excludedTestClasses>
                 <param>*.Input*</param>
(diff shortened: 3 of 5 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/ClassMemberImpliedModifierCheck.java
```diff
@@ -243,7 +243,8 @@ public void visitToken(DetailAST ast) {
      * @return true if ast is in a class, enum, or record
      */
     private static boolean isInTypeBlock(DetailAST ast) {
-        return ScopeUtil.isInClassBlock(ast)
+        return ScopeUtil.isInObjBlock(ast)
+                || ScopeUtil.isInClassBlock(ast)
                 || ScopeUtil.isInEnumBlock(ast)
                 || ScopeUtil.isInRecordBlock(ast);
```
nick-mancuso, line 249: All of these structures have an `OBJ_BLOCK` token; can we just check if `ast` is in an `OBJECT_BLOCK` only?

src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
```diff
@@ -181,6 +181,29 @@ public static boolean isInInterfaceBlock(DetailAST node) {
         return isInBlockOf(node, TokenTypes.INTERFACE_DEF);
     }
 
+    /**
+     * Returns whether a node is directly contained within an anonymous object block.
+     *
+     * @param node the node to check if directly contained within an anonymous object block.
+     * @return a {@code boolean} value
+     */
+    public static boolean isInObjBlock(DetailAST node) {
```
nick-mancuso, line 190: Let's keep this as a private method in `ClassMemberImpliedModifier` until we can use it elsewhere.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/classmemberimpliedmodifier/InputClassMemberImpliedModifierObjNestedRecord.java
```diff
@@ -0,0 +1,17 @@
+/*
+ClassMemberImpliedModifier
+violateImpliedStaticOnNestedEnum = (default)true
+violateImpliedStaticOnNestedInterface = (default)true
+violateImpliedStaticOnNestedRecord = (default)true
+
+
+*/
+
+//non-compiled with javac: Compilable with Java16
```
nick-mancuso, line 10: Please use lowest version of Java that allows this syntax.

nick-mancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/ClassMemberImpliedModifierCheck.java
as23415672: No, we need to make sure it is not directly included in a Interface block

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/classmemberimpliedmodifier/InputClassMemberImpliedModifierObjNestedRecord.java
as23415672: Done

src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
as23415672: Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/ClassMemberImpliedModifierCheck.java
nick-mancuso: Ah, yes. This is taken care of by https://checkstyle.sourceforge.io/config_modifier.html#InterfaceMemberImpliedModifier, thanks for the reminder.

nick-mancuso (changes requested):
Please rebase on latest master to fix CI, and generate check regression report via github action as described at https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#report-generation

nick-mancuso:
@as23415672 it appears that report generation is failing because you do not have a fork of checkstyle: https://github.com/checkstyle/checkstyle/actions/runs/2511432701

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/ClassMemberImpliedModifierCheck.java
```diff
@@ -243,9 +243,33 @@ public void visitToken(DetailAST ast) {
...
+    /**
+     * Returns whether a node is directly contained within an anonymous object block.
+     *
+     * @param node the node to check if directly contained within an anonymous object block.
+     * @return a {@code boolean} value
+     */
+    public static boolean isInObjBlock(DetailAST node) {
+        boolean returnValue = false;
+
+        // Loop up looking for a containing interface block
```
strkkk, line 261: `interface block`? looks like a copy-paste

strkkk:
github action for report generation expects PR from user fork, not from some 3rd party repo.

nick-mancuso:
@as23415672 do you plan to continue work on this PR?

nick-mancuso:
We have lost contact with author, anyone is welcome to reuse code here to solve issue.

PR #11684, opened 2022-05-30
Issue #11140: solve UAC_UNNECESSARY_API_CONVERSION_FILE_TO_PATH

pbludov:
Part of #11140


src/main/java/com/puppycrawl/tools/checkstyle/PropertyCacheFile.java
```diff
@@ -120,9 +119,9 @@ public void load() throws IOException {
         // get the current config so if the file isn't found
         // the first time the hash will be added to output file
         configHash = getHashCodeBasedOnObjectContent(config);
-        final File file = new File(fileName);
```
pbludov, line 123: I don't see much difference between old and new code. In any case, it is better to change than to suppress this violation.

nick-mancuso (approved):

strkkk (approved):

PR #11672, opened 2022-05-26
 Issue #13155: Fix false negative in VariableDeclarationDistanceUsage 

Vyom-Yadav:
Resolves #13155

Diff Regression config: https://gist.githubusercontent.com/Vyom-Yadav/98dceb63a79f4833e85fff9b2e1464a6/raw/a875094801a518055bde1c71fa360f703d1ea7f6/my_checks.xml
Report label: AllConfig

---

Final Report: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a99d8b4_2023120550/reports/diff/index.html

---

Coverage for FullIdent, will be incomplete for now, complete coverage is in #11231

PR #11694, opened 2022-06-03
minor: Refactor VariableDeclarationDistanceUsage to reduce complexity

Vyom-Yadav:
Refactoring the check before working on #11672 

To Do-

- [ ] Generate regression report.
- [ ] Fix pitest failures, if any.


Vyom-Yadav:
Does some logical changes as well, better to be in #11672 only.

PR #6781, opened 2019-05-24
Issue #6311: OrderedPropertiesCheck check that properties are ordered

thomassenger:
Issue #6311: OrderedPropertiesCheck 
new pull request on branch issue-6311 , because on master is the diff report

PR #6343, opened 2019-01-03
Issue: #6311 new check: OrderedProperties

thomassenger:
#6311 

The OrderedProperties Check is almost ready. 
There are the following open issues:
- builds only with: mvn package -Dcheckstyle.ant.skip=true -Dcheckstyle.skip=true -DskipTests
-- I need help with failure: XdocsPagesTest.testAllCheckSections:356->validateCheckSection:451->validateParentSection:1300 config_misc.xml section 'OrderedProperties' should have matching parent expected:<[TreeWal]ker> but was:<[Chec]ker>

PR #11702, opened 2022-06-06
minor: use iterator instead of enumeration

nick-mancuso:
Noticed during work on https://github.com/checkstyle/checkstyle/pull/11691, it is better to fix a simple violation like this, than to put a comment on it and let it live in the code base.

Diff Regression config: https://gist.githubusercontent.com/nick-mancuso/e46a3d11fe985fac1c4468c6b0f7fa4e/raw/98085c1a02a72fdf3419e35c5cbfe23cd9aa111c/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/OrderedPropertiesCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/OrderedPropertiesCheck.java
@@ -26,6 +26,7 @@ import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Properties;
 import java.util.regex.Matcher;
@@ -123,7 +124,6 @@ public class OrderedPropertiesCheck extends AbstractFileSetCheck {
      *
      * @param file the file to be processed
      * @param fileText the contents of the file.
-     * @noinspection EnumerationCanBeIteration
      */
     @Override
     protected void processFiltered(File file, FileText fileText) {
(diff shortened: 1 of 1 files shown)


romani:
We need regression report on Check.
Please recheck history of this code, it might be recent and there might be details from contributor why suppression was required.

nick-mancuso:
@romani I cannot find explanation for using enumeration, this area of code has not changed since original implementation at https://github.com/checkstyle/checkstyle/pull/6781/files

nick-mancuso:
Github, generate report

romani:
I couldn't believe that we allowed suppression without discussing it 
Here it is https://github.com/checkstyle/checkstyle/pull/6343#discussion_r258530208

nick-mancuso:
> I couldn't believe that we allowed suppression without discussing it Here it is [#6343 (comment)](https://github.com/checkstyle/checkstyle/pull/6343#discussion_r258530208)

Using iterator as I have here should not be a problem, I am not using keyset.

@romani 
I do not think that we can run diff report on properties files.

romani:
I doubt we have projects for properties, but we run on all, it will be enough.

nick-mancuso:
@rnveach can you use your magic scripts to generate diff report for this PR please?

rnveach:
@nick-mancuso There is no magic. I have dropped maven use as https://github.com/checkstyle/contribution/issues/273#issue-284400200 states, maven is forcing us to ignore non-Java files. My version of the scripts are at https://github.com/checkstyle/contribution/tree/master/checkstyle-tester too.

Config:
````
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">

    <module name="OrderedProperties"/>
</module>
````

Report:
http://rveach.no-ip.org/checkstyle/regression/312/

No differences. If there could be a random hash issue, it would be hard for regression to find it.

romani (approved):
ok to merge

pbludov (approved):

rnveach (approved):
I don't really like using iterators.

PR #11711, opened 2022-06-08
Issue #11710: Disable IDEA Inspection for AnnotationClass

### LINKED ISSUE #11710: Disable IDEA Inspection for  `AnnotationClass`
The `AnnotationClass` inspection is a "Java Language Level" issue, which pertains to Java versions less than 1.5 (Java 5). Since we only support jdk/jre >= 11, this inspection is not relevant to the project. Sadly, the IDEA inspections website has no information about this inspection, but from the "Inspections" view in Intellij: 

`Reports annotation interfaces. Such interfaces are not supported in Java 1.4 and earlier JVM. `

~~The `'assert' statement` inspection pertains to Java versions less than 1.3 (Java 3). Since we only support jdk/jre >= 11, this inspection is not relevant to the project. Sadly, the IDEA inspections website also has no information about this inspection, but from the "Inspections" view in Intellij:~~

~~`Reports assert statements. By default, assert statements are disabled during execution in the production environment. Consider using logger or exceptions instead.
The assert statements are not supported in Java 1.3 and earlier JVM. `~~

We can safely remove ~~these two~~ this inspection since we do not support language levels below 11.

nick-mancuso:
Closes #11710

### DIFF
--- a/config/intellij-idea-inspections.xml
+++ b/config/intellij-idea-inspections.xml
@@ -629,8 +629,9 @@
                    enabled_by_default="false"/>
   <!-- this rule is inspection is only for old java, we are ok to use annotations -->
   <inspection_tool class="Annotation" enabled="false" level="ERROR" enabled_by_default="false"/>
-  <inspection_tool class="AnnotationClass" enabled="true" level="ERROR"
-                   enabled_by_default="true"/>
+  <!-- this rule is inspection is only for old java, we are ok to define annotations -->
+  <inspection_tool class="AnnotationClass" enabled="false" level="ERROR"
+                   enabled_by_default="false"/>
   <inspection_tool class="AnnotationNamingConvention" enabled="true" level="WARNING"
                    enabled_by_default="true">
     <option name="m_regex" value="[A-Z][A-Za-z\d]*"/>
@@ -4456,8 +4457,6 @@
         <option value="ThisEscapedInObjectConstruction"/>
         <!-- it will make code too complicated in some cases -->
         <option value="MultipleReturnPointsPerMethod"/>
-        <!-- MT check markers are annotations -->
-        <option value="AnnotationClass"/>
         <!-- Suppression is only for TokenTypes class -->
         <option value="ClassWithTooManyDependents"/>
         <!-- XDocs UT is very long and complex -->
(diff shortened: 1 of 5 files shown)


config/intellij-idea-inspections.xml
```diff
@@ -629,8 +629,8 @@
                    enabled_by_default="false"/>
   <!-- this rule is inspection is only for old java, we are ok to use annotations -->
   <inspection_tool class="Annotation" enabled="false" level="ERROR" enabled_by_default="false"/>
-  <inspection_tool class="AnnotationClass" enabled="true" level="ERROR"
-                   enabled_by_default="true"/>
+  <inspection_tool class="AnnotationClass" enabled="false" level="ERROR"
+                   enabled_by_default="false"/>
```
romani, line 633: Please put comment on reason of disablement

romani (changes requested):
Items

config/intellij-idea-inspections.xml
nick-mancuso: Done

romani (approved):
Ok to merge

src/main/java/com/puppycrawl/tools/checkstyle/XdocsPropertyType.java
```diff
@@ -29,7 +29,6 @@
  * for clarification of the user-friendly type.
  * For example, some string array property may represent a set of tokens.
  *
```
rnveach, line 31: Please remove empty ending line for javadoc.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/XdocsPropertyType.java
nick-mancuso: Done

rnveach (approved):

config/intellij-idea-inspections.xml
pbludov: "this rule is inspection is only for old java" -> "this inspection is only for old java"?

romani:
rebased.

PR #11721, opened 2022-06-11
infra: test disabling all CI

nick-mancuso:
Test, do not merge


nick-mancuso:
Github, generate report

nick-mancuso:
Only problem left is how to disable idea inspections, removal of config is not enough

romani:
Migration of idea to text config would be very helpful.

@pbludov , can you help to do export or TC config to file? Looks like it will be very good to expedite migration.

Vyom-Yadav:
> Migration of idea to text config would be very helpful.
> 
> @pbludov , can you help to do export or TC config to file? Looks like it will be very good to expedite migration.

We can export project settings to store them in version control. https://www.jetbrains.com/help/teamcity/storing-project-settings-in-version-control.html

nick-mancuso:
We could also add a branch filter in web config, name all PR branches to avoid team city execution, and run team city manually before merge. @romani what do you think about this? If we make a mistake, we still catch inspection issues in master.

romani:
We already agreed to keep filters for PR that are not for merge ( most CIs can be disabled). As such PR is closed and we got agreement in it and it has links to all reports that prove something - we create regular PR that has all enabled and it should pass regular execution of all.

nick-mancuso:
Closing this since we have figured out how to disable all CI that we needed to.

Issue #11698, opened 2022-06-05
Missed limitation: Checkstyle can't parse UTF8 Java File with UTF8 BOM


rnveach:
Identified at https://github.com/rnveach/checkstyle-extras when doing regression on XML on sevntu,

During review of XML regression and a parse error, it was identified sevntu had an XML file with an UTF8 BOM.
https://raw.githubusercontent.com/sevntu-checkstyle/sevntu.checkstyle/master/sevntu-checks/sevntu-checks.xml
https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8

If it wasn't for the BOM, the file would parse correctly. In thinking over the solution for XML, it was identified that this issue also applies to Checkstyle and it's Java parsing.

Code to Construct and Show the case (UTF8 BOM is non-printable):
````
package com.puppycrawl.tools.checkstyle;

import java.io.BufferedWriter;
import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.Paths;

public class CreateJavaBom {

    public static void main(String... args) throws IOException {
        try (BufferedWriter bw = Files.newBufferedWriter(Paths.get("M:\\JavaBOM.java"))) {
            bw.write("\ufeff"); // UTF8 BOM
            bw.write("public class TestClass {");
            bw.newLine();
            bw.write("    void method() {");
            bw.newLine();
            bw.write("    }");
            bw.newLine();
            bw.write("}");
            bw.newLine();
        }

        Main.main("-T", "M:\\JavaBOM.java");
    }

}
````

Output:
````
com.puppycrawl.tools.checkstyle.api.CheckstyleException: IllegalStateException occurred while parsing file M:\JavaBOM.java.
	at com.puppycrawl.tools.checkstyle.JavaParser.parse(JavaParser.java:105)
	at com.puppycrawl.tools.checkstyle.JavaParser.parseFileText(JavaParser.java:122)
	at com.puppycrawl.tools.checkstyle.JavaParser.parseFile(JavaParser.java:142)
	at com.puppycrawl.tools.checkstyle.AstTreeStringPrinter.printFileAst(AstTreeStringPrinter.java:65)
	at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:293)
	at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:189)
	at com.puppycrawl.tools.checkstyle.Main.main(Main.java:126)
[...]

PR #11738, opened 2022-06-13
Issue #11698: reorganizes checkstyle limitations and add bom

rnveach:
Issue #11698

### DIFF
--- a/src/xdocs/writingchecks.xml
+++ b/src/xdocs/writingchecks.xml
@@ -660,10 +660,19 @@ public class MethodLimitCheck extends AbstractCheck
 
     <section name="Limitations">
       <p>
-        There are only a few, but important, limitations for Checkstyle:
+        There are only a few, but important, limitations for Checkstyle. These pertain to Java
+        specifically, as well as regular text files.
+      </p>
+      <p>
+        For Java,
       </p>
       <ul>
-        <li>Java
+        <li>
+          Code has to be compilable by <code>javac</code> to get valid violations.
+          If it is not, you can get hard to understand parse errors.
+        </li>
+        <li>
+          Java
           <a href="https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.5">
             tokens (identifiers, keywords)</a>
           should be written with
@@ -679,14 +688,25 @@ public class MethodLimitCheck extends AbstractCheck
           <a href="https://github.com/checkstyle/checkstyle/issues/10629">
           Issue #10629.</a>
         </li>
-        <li>To get valid violations, code has to be compilable.
-          If it is not, you can get hard to understand parse errors.</li>
         <li>You cannot determine the type of an expression.
         Example: "getValue() + getValue2()"</li>
         <li>You cannot determine the full inheritance hierarchy of type.</li>
-        <li>You cannot see the contents of other files. You have access to the
+      </ul>
+      <p>
+        In addition to Java files, there are similar limitations that apply to all
+        type of text files:
+      </p>
+      <ul>
+        <li>
+          Checkstyle cannot properly handle a file's binary
+          <a href="https://en.wikipedia.org/wiki/Byte_order_mark">
+          Byte Order Mark (BOM)</a>, like UTF8's Unicode \uFEFF.
+        </li>
+        <li>
+          You cannot see the contents of other files. You have access to the
           content of one file only during all Checks execution. All files
-          are processed one by one.</li>
+          are processed one by one.
+        </li>
       </ul>
       <p>
         This means that you cannot implement some code inspection


romani:
GitHub, generate website

nick-mancuso:
Github, generate site

src/xdocs/writingchecks.xml
```diff
@@ -679,14 +688,25 @@ public class MethodLimitCheck extends AbstractCheck
...
         <li>You cannot determine the full inheritance hierarchy of type.</li>
-        <li>You cannot see the contents of other files. You have access to the
+      </ul>
+      <p>
+        In addition to Java files, there are similar limitations that apply to all
+        type of text files:
+      </p>
+      <ul>
+        <li>
+          Checkstyle cannot properly handle a file's binary
```
nick-mancuso, line 701: I am not sure why, but I do not like `properly` here. Maybe `Checkstyle will not be able to parse...`?

src/xdocs/writingchecks.xml
```diff
@@ -679,14 +688,25 @@ public class MethodLimitCheck extends AbstractCheck
...
-        <li>To get valid violations, code has to be compilable.
-          If it is not, you can get hard to understand parse errors.</li>
         <li>You cannot determine the type of an expression.
         Example: "getValue() + getValue2()"</li>
         <li>You cannot determine the full inheritance hierarchy of type.</li>
-        <li>You cannot see the contents of other files. You have access to the
+      </ul>
+      <p>
+        In addition to Java files, there are similar limitations that apply to all
+        type of text files:
```
nick-mancuso, line 697: So, we can't check XML or properties file with a BOM (linelength, etc)?

nick-mancuso (changes requested):
A few items:

src/xdocs/writingchecks.xml
rnveach: `not be able to parse` seems a bit excessive here. In terms of Java not being able to parse, it basically means the file will not be usable in checkstyle at all, like we got an exception on trying to work with it. BOM is not to that extreme. At the most, we will get extra garbage in the stream that should be ignored as it is just the header of the file. This was why I went with "properly handle". It can still be handled, but extra precautions will be needed.

src/xdocs/writingchecks.xml
rnveach: See the end of https://github.com/checkstyle/checkstyle/pull/11738#discussion_r897464531 and my example.

We can, but it requires manually weeding it out. The main point is, Checkstyle will not do this for you. If a 3rd party library like checkstyle extras doesn't take the extra effort to do this personally, then their grammars won't be able to parse those files either.

rnveach: Example on Windows:
````
package com.puppycrawl.tools.checkstyle;

import java.io.BufferedWriter;
import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.Paths;

public class CreateJavaBom {

    public static void main(String... args) throws IOException {
        try (BufferedWriter bw = Files.newBufferedWriter(Paths.get("M:\\JavaBOM.java"))) {
            bw.write("\ufeff"); // UTF8 BOM
            bw.write("public class TestClass {");
            bw.newLine();
            bw.write("    void method() {");
            bw.newLine();
            bw.write("    }");
            bw.newLine();
            bw.write("}");
            bw.newLine();
        }

        try (BufferedWriter bw = Files.newBufferedWriter(Paths.get("M:\\config.xml"))) {
            bw.write("<?xml version=\"1.0\"?>\r\n" //
                    + "<!DOCTYPE module PUBLIC\r\n" //
                    + "          \"-//Puppy Crawl//DTD Check Configuration 1.3//EN\"\r\n" //
                    + "          \"https://checkstyle.org/dtds/configuration_1_3.dtd\">\r\n" //
                    + "\r\n" //
                    + "<module name=\"Checker\">\r\n" //
                    + "    <property name=\"charset\" value=\"UTF-8\"/>\r\n" //
                    + "\r\n" //
                    + "    <module name=\"LineLength\">\r\n" //
                    + "    <property name=\"max\" value=\"0\"/>\r\n" //
                    + "    </module>\r\n" //
                    + "</module>");
        }

        Main.main("-c", "M:\\config.xml", "M:\\JavaBOM.java");
    }

}
````

Untouched Violations:
````
Starting audit...
[ERROR] M:\JavaBOM.java:1: Line is longer than 0 characters (found 25). [LineLength]
[ERROR] M:\JavaBOM.java:2: Line is longer than 0 characters (found 19). [LineLength]
[ERROR] M:\JavaBOM.java:3: Line is longer than 0 characters (found 5). [LineLength]
[ERROR] M:\JavaBOM.java:4: Line is longer than 0 characters (found 1). [LineLength]
Audit done.
Checkstyle ends with 4 errors.
````

Violations with BOM removed:
````
Starting audit...
[ERROR] M:\JavaBOM.java:1: Line is longer than 0 characters (found 24). [LineLength]
[ERROR] M:\JavaBOM.java:2: Line is longer than 0 characters (found 19). [LineLength]
[ERROR] M:\JavaBOM.java:3: Line is longer than 0 characters (found 5). [LineLength]
[ERROR] M:\JavaBOM.java:4: Line is longer than 0 characters (found 1). [LineLength]
Audit done.
Checkstyle ends with 4 errors.
````

First line's count increase by 1 character, though the physical text of the file is unchanged.

src/xdocs/writingchecks.xml
rnveach: > Would you expect Checkstyle to consume/ manipulate/ ignore file contents? What if we wanted to have a check that creates a violation for using BOMs?

Basically Yes.

Checkstyle goes against text files, as our license header says, and even the name of the file we get the contents from is called `FileText`. The BOM is not part of the file's text. It is purely a header to describe the format of the rest of the file. If we wanted binary data, they should really not be using `FileText`. If we were going to support raw binary data and examining it, I would see us needing some intermediary before `FileText` like `RawFileData` or such.

nick-mancuso: > t is purely a header to describe the format of the rest of the file. If we wanted binary data, they should really not be using FileText

This is also a good point.

src/xdocs/writingchecks.xml
nick-mancuso: Thanks for example. This situation could definitely be confusing for someone who is not familiar with BOM.

nick-mancuso (approved):

romani (approved):
ok to merge

PR #11771, opened 2022-06-26
minor: Hardcoding mutation in RequireThis 

Vyom-Yadav:
Hardcoding mutation in `RequireThis`, for testing purposes, DO NOT MERGE

Diff Regression config: https://gist.githubusercontent.com/Vyom-Yadav/98dceb63a79f4833e85fff9b2e1464a6/raw/b424881849408a0cc74460d762ba178bc3957f67/my_checks.xml
Report label: DefaultConfig


Vyom-Yadav:
Github, generate report

Vyom-Yadav:
This condition is logically redundant, `if (true)`, it will go into the branch in all cases, even when `toVisit != null`, in that case, we change the value of `curNode`, but it doesn't matter as when `toVisit != null`, we break from the loop and `curNode` is set back to
 `curNode = toVisit;`

Vyom-Yadav:
Main PR #11775 has been merged.

Issue #11720, opened 2022-06-11
Pitest: Kill all surviving mutations


nick-mancuso:
This will be main tracker issue for killing surviving mutations before we begin work in https://github.com/checkstyle/checkstyle/issues/11719

Now that we have diff report generation as github action, let's modify the procedure from  https://github.com/checkstyle/checkstyle/issues/7797#issue-577522554 a bit to speed this up:

1. Disable all CI first as separate commit(See example at https://github.com/checkstyle/checkstyle/pull/11721). This will help to not waste credits.
2. Pick one single surviving mutation from pitest.sh.
3. Send PR with a single hardcoded mutation only.
4. Generate diff report with hardcoded mutation.
5. Share link to initial commit with hardcoded mutation and initial diff report in PR description. This is in lieu of separate branch/ etc. in previous issue at #7797. 

HARDCODED MUTATION WILL NOT BE IN FINAL PR.

We will have one of three outcomes for each mutation (in order of preference):
1. Find a case that kills the mutation via diff report, add new test in existing PR
2. Safely (no regressions) modify code to be covered, in existing PR
3. Safely (no regressions) remove code to be covered, in existing PR

Case (2) and (3) required a detailed analysis and discussion in issue before making any changes.

-----------------------------------

**We will not be pursuing surviving mutations in IndentationCheck or gui package .**

nick-mancuso:
Once we approve this issue, we should close https://github.com/checkstyle/checkstyle/issues/7797 and any child/ associated issues.

romani:
approved.

we need to have commit that disables all CIs. to make sure we will not eat all credits for CI provided, and we are not banned from CI usage due to misbehavior (overusage). 
so PRs for search of killer will have two commits: 
1) disable all CIs jobs 
2) hardcoded survival mutation

PR should have no CI jobs started or all of them do nothing (skip all jobs).

the only activity happens by comment to trigger Github Action.

Vyom-Yadav:
@rnveach @nick-mancuso 
The new pitest method only examines `mutations.xml` which isn't appropriate to determine lines that are uncovered and are not mutated (due to the absence of some mutator). Such lines can only be taken from the HTML file.

An example of such lines is present over [here](https://vyom-yadav.github.io/DiffReport/pit-blocks-mod-1/index.html) (modified a bit for demonstrating).

[...]

PR #11792, opened 2022-06-29
Issue #11720: Kill surviving mutation in RequireThisCheck related to static blocks

Vyom-Yadav:
#11720
Hardcoded mutation tested at https://github.com/checkstyle/checkstyle/pull/11776

### Reports with hardcoded mutation (clean):

- DefaultConfig: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/dc3d699_2022053018/reports/diff/index.html
- validateOnlyOverlappingFalseWithoutOpenJDK: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/dc3d699_2022072839/reports/diff/index.html

`validateOnlyOverlapping = false` report was not generated with OpenJDK (Out of memory in GitHub actions), though it is very unlikely that it will show any difference, still, if it is required, ping me, and I will generate it locally.

### This mutation falls in the category:

- Safely (no regressions) remove code to be covered

### Rationale:

The condition `definitionToken.getType() == TokenTypes.STATIC_INIT` is same as the condition 
`codeBlockDefinition.getType() == TokenTypes.STATIC_INIT`. A variable declared in static block scope, will always have an enclosing `CLASS_FRAME`.

In the current logic, we skip all frames of type `BLOCK_FRAME`, and `FOR_FRAME` to reach the `CLASS_FRAME`, if we can't reach `CLASS_FRAME` (maybe `METHOD_FRAME` is in between), then we can be assured that there is no static block present.

So if there is a static block, we will surely go into the branch
`if (variableDeclarationFrame.getType() == FrameType.CLASS_FRAME)` in which we check for the condition-
```java
staticContext = codeBlockDefinition.getType() == TokenTypes.STATIC_INIT
    || modifiers.findFirstToken(TokenTypes.LITERAL_STATIC) != null;
```
Hence the line can be safely removed, in theory, we would take a hit in TC, as we would make another traversal in `getCodeBlockDefinitionToken(..)` but in reality, there will be no difference, in real case use scenario, 2-3 more iterations.

---
### Generating reports again:
Diff Regression config: https://gist.githubusercontent.com/Vyom-Yadav/98dceb63a79f4833e85fff9b2e1464a6/raw/5dc88b466a50335d5759f60c1d041a239f11a5a1/my_checks.xml
Diff Regression projects: https://gist.githubusercontent.com/Vyom-Yadav/91807e6244cff60698d9e33e32ba2d51/raw/ccf3b6b9e053f27b371eb6d1e848d7199dd1f567/projects-to-test-on.properties
Report label: validateOnlyOverlappingFalseWithoutOpenJDK

### DIFF
--- a/.ci/pitest-suppressions/pitest-coding-require-this-check-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-coding-require-this-check-suppressions.xml
@@ -4,15 +4,6 @@
     <sourceFile>RequireThisCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.RequireThisCheck</mutatedClass>
     <mutatedMethod>canBeReferencedFromStaticContext</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_ELSE</mutator>
-    <description>removed conditional - replaced equality check with false</description>
-    <lineContent>if (definitionToken.getType() == TokenTypes.STATIC_INIT) {</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>RequireThisCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.RequireThisCheck</mutatedClass>
-    <mutatedMethod>canBeReferencedFromStaticContext</mutatedMethod>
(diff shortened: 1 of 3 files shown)


pbludov (approved):

nick-mancuso (changes requested):
Please generate check regression reports for this PR.

Vyom-Yadav:
Github, generate report

Vyom-Yadav:
Github, generate report

nick-mancuso (approved):

PR #11793, opened 2022-06-29
Issue #11446: Update EmptyLineSeparatorCheckTest to use verifyWithInlineConfigParser

stoyanK7:
Addresses #11446
Updates 1 test in `EmptyLineSeparatorCheckTest` to use `verifyWithInlineConfigParser()`.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheckTest.java
@@ -488,27 +488,25 @@ public class EmptyLineSeparatorCheckTest
 
     @Test
     public void testEmptyLineSeparatorRecordsAndCompactCtors() throws Exception {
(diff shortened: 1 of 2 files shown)


stoyanK7:
@Vyom-Yadav I think you could help me with the failing `pitest-whitespace` check. It has something to do with the 2 new tests I added. I see you wrote the `pitest-survival-check-html.sh` file. Could you explain what it does?

stoyanK7:
Surviving mutations:

romani:
I am ok to restore some test cases to use old approach of testing. Life is not ideal, we already have few Check that can not use such model of tests.

romani:
Try to minimize restoration.

If we have exception of rule, we should have it as minimal as possible.

romani (approved):

romani:
@stoyanK7 ,
please create issue to explain why some test still on old verify method.

and change comment in pom.xml to be `<!-- until ....link to ticket ...-->` , we might come up with ideas later on on how to fix it or minimize it.

Issue #11719, opened 2022-06-11
Pitest: Activate "ALL" mutator group


nick-mancuso:
See https://pitest.org/quickstart/mutators/

**As an initial step, we will generate an initial report with all mutators for each profile and share via github.io in a comment below. This will help us to understand scope of work before we begin.**

In all profiles, we will activate the "ALL" group of mutators, using the following procedure:

1. Pick one profile from checklist below
2. Amend mutators to include the "ALL" group
3. Create suppressions necessary to achieve green state in CI
4. Add suppressions to a main tracker issue (TODO) 

---------------------------

Some notes:

- Mutator [NON_VOID_METHOD_CALLS](http://pitest.org/quickstart/mutators/#NON_VOID_METHOD_CALLS) is considered unstable for certain types like Objects where NullPointerExceptions are likely
- We are excluding `InlineConstantMutator` for all profiles with link to comment from https://github.com/checkstyle/checkstyle/pull/11873#discussion_r920652226

-------------------------

- [x] pitest-misc
- [x] pitest-annotation
- [x] pitest-blocks
- [x] pitest-coding-1
- [x] pitest-coding-2
- [x] pitest-coding-require-this-check
- [x] pitest-design
- [x] pitest-header
- [x] pitest-imports
- [x] pitest-indentation
- [x] pitest-javadoc
- [x] pitest-metrics
- [x] pitest-modifier
- [x] pitest-naming
- [x] pitest-regexp
- [x] pitest-sizes
- [x] pitest-whitespace
- [x] pitest-ant
- [x] pitest-packagenamesloader
- [x] pitest-common
- [x] pitest-common-2
- [x] pitest-main
- [x] pitest-tree-walker
- [x] pitest-api
- [x] pitest-filters
- [x] pitest-utils
- [x] pitest-gui
- [x] pitest-xpath
- [x] pitest-java-ast-visitor

------------------------------
**We will not be pursuing surviving mutations in IndentationCheck or gui package .**

nick-mancuso:
Once we approve this issue, we can close https://github.com/checkstyle/checkstyle/issues/7132 and https://github.com/checkstyle/checkstyle/issues/6320.

Vyom-Yadav:
@nick-mancuso @romani Here is the report with `ALL` mutator group-
https://vyom-yadav.github.io/pit-reports/

rnveach:
[...]

Issue #11871, opened 2022-07-09
Calculate and compare the pitest execution time after including CDG accelerator plugin.


Vyom-Yadav:
From https://github.com/checkstyle/checkstyle/pull/11869#issuecomment-1179587293

The execution time of pitest has reduced after including the CDG accelerator plugin see detailed report at #11869.

 The task of this PR is to calculate the execution time of pitest alone (without setting up the environment in Github actions) and give that report to the maintainers of the plugin.

To calculate the time, create a `pitest-all` profile in local (with all classes present in all the different profiles). Run mutation coverage without the CDG accelerator plugin, and save the log (basically, the time elapsed).

Now run the same profile with the plugin, compare the diff, and report. If the mutation coverage does not take a lot of time (last time we tried the `pitest-all` profile, it took 12 hrs) make a PR, with that profile in a Github action.

romani:
@Vyom-Yadav , can you take most long executed jobs (1-3) and compare results. I hope it would be great feeedback to owners of this library.

romani:
I did testing at https://github.com/checkstyle/checkstyle/pull/12336

romani:
Our tests are complicated and heavy by design, on purpose. Our heavy tests produce more quality value. In such situations we value quality more than speed of tests.

PR #11873, opened 2022-07-09
Issue #11719: Activate ALL group for pitest-javadoc

Vyom-Yadav:
#11719 
Activating `ALL` group for `pitest-javadoc`

Every mutator has been activated except `INLINE_CONSTS`

- Report with `ALL` group: https://vyom-yadav.github.io/pit-reports-all/pitest-javadoc/202207100201/index.html

- Report with `ALL` group except `INLINE_CONSTS`: https://vyom-yadav.github.io/pit-reports-all/pitest-javadoc-without-inline-consts/202207142332/index.html


.ci/pitest-suppressions/pitest-javadoc-suppressions.xml
```diff
@@ -0,0 +1,1973 @@
...
+    <description>removed conditional - replaced equality check with true</description>
+    <lineContent>|| ast.getType() == TokenTypes.COMPACT_CTOR_DEF)</lineContent>
+  </mutation>
+
+  <mutation unstable="false">
+    <sourceFile>WriteTagCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.WriteTagCheck</mutatedClass>
+    <mutatedMethod>logTag</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.InlineConstantMutator</mutator>
+    <description>Substituted 2 with 3</description>
```
nick-mancuso, line 1889: `InlineConstantMutator` could be good for mutating token values in checks, to make us rethink all tokens and related logic in certain checks, methods, etc. But - this mutator is causing a large number of violations that may require a lot of work to create all test cases to cover.

@rnveach @romani @pbludov @baratali @Vyom-Yadav please share your thoughts on the value of this mutator.

Link to docs: https://pitest.org/quickstart/mutators/#INLINE_CONSTS

Link to source: https://github.com/hcoles/pitest/blob/master/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/InlineConstantMutator.java

.ci/pitest-suppressions/pitest-javadoc-suppressions.xml
```diff
@@ -0,0 +1,1973 @@
...
+    <lineContent>currentClassName = &quot;&quot;;</lineContent>
+  </mutation>
+
+  <mutation unstable="false">
+    <sourceFile>AbstractJavadocCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheck</mutatedClass>
+    <mutatedMethod>validateDefaultJavadocTokens</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.InlineConstantMutator</mutator>
+    <description>Substituted 2 with 3</description>
+    <lineContent>final String message = String.format(Locale.ROOT,</lineContent>
```
romani, line 18: I do not understand what is 2 and what is 3 in such code.

```
    private void validateDefaultJavadocTokens() {
        if (getRequiredJavadocTokens().length != 0) {
            final int[] defaultJavadocTokens = getDefaultJavadocTokens();
            Arrays.sort(defaultJavadocTokens);
            for (final int javadocToken : getRequiredJavadocTokens()) {
                if (Arrays.binarySearch(defaultJavadocTokens, javadocToken) < 0) {
                    final String message = String.format(Locale.ROOT,
                            "Javadoc Token \"%s\" from required javadoc "
                                + "tokens was not found in default "
                                + "javadoc tokens list in check %s",
                            javadocToken, getClass().getName());
                    throw new IllegalStateException(message);
                }
            }
        }
    }
```

romani: https://docs.oracle.com/javase/7/docs/api/java/util/Locale.html#ROOT

romani (changes requested):
items:

.ci/pitest-suppressions/pitest-javadoc-suppressions.xml
Vyom-Yadav: I do not either, but taking a conservative guess on what this mutation probably means, here is the decompiled .class file (FernFlower decompiler) related part-
```java
private void validateDefaultJavadocTokens() {                                                                                                                                                                              
    if (this.getRequiredJavadocTokens().length != 0) {                                                                                                                                                                     
        int[] defaultJavadocTokens = this.getDefaultJavadocTokens();                                                                                                                                                       
        Arrays.sort(defaultJavadocTokens);                                                                                                                                                                                 
        int[] var2 = this.getRequiredJavadocTokens();                                                                                                                                                                      
        int var3 = var2.length;                                                                                                                                                                                            
                                                                                                                                                                                                                           
        for(int var4 = 0; var4 < var3; ++var4) {                                                                                                                                                                           
            int javadocToken = var2[var4];                                                                                                                                                                                 
            if (Arrays.binarySearch(defaultJavadocTokens, javadocToken) < 0) {                                                                                                                                             
                String message = String.format(Locale.ROOT, "Javadoc Token \"%s\" from required javadoc tokens was not found in default javadoc tokens list in check %s", javadocToken, this.getClass().getName());        
                throw new IllegalStateException(message);                                                                                                                                                                  
            }                                                                                                                                                                                                              
        }                                                                                                                                                                                                                  
    }                                                                                                                                                                                                                      
                                                                                                                                                                                                                           
}                                                                                                                                                                                                                          
```
Since we used enhanced for loop, I think pit couldn't report this correctly. It inlined some loop variable from `2` to `3` and that test didn't fail.

nick-mancuso: @Vyom-Yadav it is the same idea, it is mutating the value of `javadocToken`

Vyom-Yadav: > @Vyom-Yadav it is the same idea, it is mutating the value of `javadocToken`

Shouldn't it be reported on [this](https://vyom-yadav.github.io/pit-reports-all/pitest-javadoc/202207100201/com.puppycrawl.tools.checkstyle.checks.javadoc/AbstractJavadocCheck.java.html#org.pitest.mutationtest.report.html.SourceFile@10982139_219) line then?

romani: @Vyom-Yadav , please find Junit method that validate message of exception and try to validate whole message.

romani: I do not understand your answer, please share here links to Junit methods that validate such exception

Vyom-Yadav: @romani Yes, it will fail if we remove the first word from the message, but how is it implying the mutation we are talking about right now?
`Substituted 2 with 3`

romani: Please contact pitest maintainer and ask them a question what it could mean.

rnveach: Please remember, pitest is a bytecode manipulator. It doesn't read or understand the source code. If you really want to understand what pitest is doing, you need to look at the bytecode.

> I do not understand what is 2 and what is 3 in such code.

The source,
````
                    final String message = String.format(Locale.ROOT,
                            "Javadoc Token \"%s\" from required javadoc "
                                + "tokens was not found in default "
                                + "javadoc tokens list in check %s",
                            javadocToken, getClass().getName());
````

converts to:

> javap -c -l '.\AbstractJavadocCheck.class'

````
      49: getstatic     #32                 // Field java/util/Locale.ROOT:Ljava/util/Locale;
      52: ldc           #42                 // String Javadoc Token \"%s\" from required javadoc tokens was not found in default javadoc tokens list in check %s
      54: iconst_2
      55: anewarray     #34                 // class java/lang/Object
      58: dup
      59: iconst_0
      60: iload         5
      62: invokestatic  #14                 // Method java/lang/Integer.valueOf:(I)Ljava/lang/Integer;
      65: aastore
      66: dup
      67: iconst_1
      68: aload_0
      69: invokevirtual #36                 // Method java/lang/Object.getClass:()Ljava/lang/Class;
      72: invokevirtual #37                 // Method java/lang/Class.getName:()Ljava/lang/String;
      75: aastore
      76: invokestatic  #38                 // Method java/lang/String.format:(Ljava/util/Locale;Ljava/lang/String;[Ljava/lang/Object;)Ljava/lang/String;
      79: astore        6
````

I'll avoid going too far, but basically pitest is manipulating the iconst_2 ( https://github.com/hcoles/pitest/blob/master/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/InlineConstantMutator.java#L204 ) connected to the following lines:
````
      54: iconst_2
      55: anewarray     #34                 // class java/lang/Object
````

So what is this code doing from and how did it come from the source?

Basically, `String.format` parameters are `format(Locale l, String format, Object... args)` and this all comes down to how `Object...` is handled in the bytecode when calling this method.

The source passes each part of the `args` as separate strings, but the method doesn't accept separate strings. It accepts only an array. What Java does for the developers to make coding easier is it sees and makes this simple conversion for you.

To make this conversion it actually instantiates an Object array for you and puts the arguments you are passing into the array and passes that to the method. In this case we are passing **2** strings to the `args`. So basically your simple `   javadocToken, getClass().getName()` in the source actually becomes `new Object[] {javadocToken, getClass().getName()}` or basically:
````
final Object[] temp = new Object[2];
temp[0] = javadocToken;
temp[1] = getClass().getName();
````

You all can start to see where the 2 comes into play here. It is the number of arguments being passed specifically to the `Object...`.

If you actually embed this change in the source and place it on an isolated line and run pitest, you will see the surviving mutant is only on this line and the rest of the code passes.

Comment (romani) [reply .ci/pitest-suppressions/pitest-javadoc-suppressions.xml:18]: We can not use this mutation, it this specific case can not be excluded by config

File: .ci/pitest-suppressions/pitest-javadoc-suppressions.xml (hunk 1 above)
Comment (Vyom-Yadav) [reply .ci/pitest-suppressions/pitest-javadoc-suppressions.xml]: Given the number of violations and the degree of tediousness v/s effectiveness, I would like to vote in favor of not using this mutator.

File: .ci/pitest-suppressions/pitest-javadoc-suppressions.xml (hunk 2 above)
Comment (romani) [reply .ci/pitest-suppressions/pitest-javadoc-suppressions.xml]: Please create issue and move comment of rnveach to it , keep suppression and put to new issue obove it to explain reason.

Comment (Vyom-Yadav) [reply .ci/pitest-suppressions/pitest-javadoc-suppressions.xml]: Done, created at #11895

File: pom.xml [hunk 3]
```diff
@@ -2920,18 +2920,40 @@
...
+                <mutator>EXPERIMENTAL_SWITCH</mutator>
+                <mutator>RETURNS</mutator>
+                <mutator>EXPERIMENTAL_MEMBER_VARIABLE</mutator>
+                <mutator>NULL_RETURNS</mutator>
+                <mutator>EXPERIMENTAL_BIG_DECIMAL</mutator>
                 <mutator>MATH</mutator>
+                <mutator>EXPERIMENTAL_BIG_INTEGER</mutator>
+                <mutator>INCREMENTS</mutator>
+                <mutator>EXPERIMENTAL_ARGUMENT_PROPAGATION</mutator>
+                <mutator>EXPERIMENTAL_NAKED_RECEIVER</mutator>
```
nick-mancuso, line 2948: I am good with current list of mutators for this profile, including experimental mutators. In fact, I feel that these two experimental mutators in particular are very valuable. Here is an example of questionable code that `EXPERIMENTAL_ARGUMENT_PROPAGATION` caught:

```
    private static DetailNode getNearestNode(DetailNode node) {
        DetailNode tag = JavadocUtil.getNextSibling(node);
        // is replaced with 'DetailNode tag = node;'
        while (tag.getType() == JavadocTokenTypes.LEADING_ASTERISK
                || tag.getType() == JavadocTokenTypes.NEWLINE) {
            tag = JavadocUtil.getNextSibling(tag);
        }
        return tag;
    }

```

nick-mancuso (approved):
I am good with all mutators for this profile.

rnveach (approved):

pom.xml
```diff
@@ -2920,18 +2920,40 @@
...
+                <mutator>CONSTRUCTOR_CALLS</mutator>
+                <mutator>REMOVE_SWITCH</mutator>
+                <!-- Results in hard to debug mutations, see https://github.com/checkstyle/checkstyle/issues/11895 -->
```
romani, line 2951: `<!-- reason of disablement read at https://github.com/checkstyle/checkstyle/issues/11895 -->`

romani (changes requested):
items:

pom.xml
Vyom-Yadav: Done.

romani (approved):

romani:
pitest-javadoc use to be 5min([after acceleration](https://github.com/checkstyle/checkstyle/issues/11871)) to execute, now become 20min.

Vyom-Yadav:
> pitest-javadoc use to be 5min([after acceleration](https://github.com/checkstyle/checkstyle/issues/11871)) to execute, now become 20min.

We increased the mutators so time increased, I believe it will increase for other profiles (which earlier took 10 - 20 min) by a lot.

romani:
I understand, it is just another point of consideration on how/when it is reasonable to activate such mutators and compare with actual benefit they bring.

PR #11884, opened 2022-07-10
Issue #7064: Add influenceFormat property to SuppressWithPlainTextCommentFilter

### LINKED ISSUE #7064: SuppressWithPlainTextCommentFilter should has influenceFormat property
https://checkstyle.org/config_filters.html#SuppressWithPlainTextCommentFilter

We spoke about it some time ago and become an user raised issue at https://github.com/jqno/equalsverifier/pull/245

SuppressWithPlainTextCommentFilter should support influenceFormat property same as at https://checkstyle.org/config_filters.html#SuppressWithNearbyCommentFilter_Properties , as proof of implementation we need to restore code/comments of suppression at equalsverifier.

This should allow user to use suppression on LineLength Check that is now under Checker (#2116).

stoyanK7:
Resolves #7064
## Changes
* Adds `influenceFormat` property along with `commentFormat`
* Adds 3 tests to cover new property
* Updates all other tests with a lot more examples covering the new change
* Updates all other tests' inline configurations to include the new property
* Updates documentation about the new property(I will provide screenshots of the website once the problem below is discussed)
    * Examples of how this prop is used in javadoc and site are not added. I didn't add them yet because I think adding a reference to `SuppressWithNearbyCommentFilter`'s examples might be better. In the end it's the same prop doing the exact same thing. No need to repeat it. Let me know what you think.

--- 
Diff Regression config: https://gist.githubusercontent.com/stoyanK7/0d95fd3ec8caa5cae9eb597ac8c78b8d/raw/c60fcc04f28f853f530efeceb598e6b5173d4855/config.xml


stoyanK7:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java
```diff
@@ -542,50 +603,97 @@ private static final class Suppression {
...
             this.suppressionType = suppressionType;
 
-            final Pattern commentFormat;
-            if (this.suppressionType == SuppressionType.ON) {
-                commentFormat = filter.onCommentFormat;
-            }
-            else {
-                commentFormat = filter.offCommentFormat;
-            }
+            final Pattern filterCommentFormat = getFilterCommentFormat(filter);
```
stoyanK7, line 606: This method extraction is needed due to NCSS line count violation.

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java
```diff
@@ -498,6 +550,8 @@ private enum SuppressionType {
         ON,
         /** Off suppression type. */
         OFF,
+        /** Default suppression type. */
+        DEFAULT,
```
stoyanK7, line 554: I decided to go with `DEFAULT`. Not sure if this is the best choice, let me know what you think.

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java
```diff
@@ -77,6 +77,11 @@
  * </p>
  * <ul>
  * <li>
+ * Property {@code commentFormat} - Specify comment pattern to trigger filter to begin suppression.
```
stoyanK7, line 80: The description of this property is the same as the next one. I think we can maybe mention it is tied to `influenceFormat` or something like that?

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java
```diff
@@ -327,6 +342,12 @@ public class SuppressWithPlainTextCommentFilter extends AutomaticBean implements
     /** Default check format to suppress. By default, the filter suppress all checks. */
     private static final String DEFAULT_CHECK_FORMAT = ".*";
 
+    /** Default regex for lines that should be suppressed. */
+    private static final String DEFAULT_INFLUENCE_FORMAT = "0";
+
+    /** Specify comment pattern to trigger filter to begin suppression. */
+    private Pattern commentFormat = Pattern.compile(DEFAULT_COMMENT_FORMAT);
```
stoyanK7, line 349: I was thinking naming this `defaultCommentFormat` since in the enum down below I chose `DEFAULT`?

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java
stoyanK7: `NEARBY` sounds good.

stoyanK7:
@romani ping

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java
```diff
@@ -542,50 +602,97 @@ private static final class Suppression {
...
 
+        /**
+         * Gets filter comment format based on suppression type.
+         *
+         * @param filter Filter containing the comment formats.
+         * @return comment format.
+         */
+        private Pattern getFilterCommentFormat(SuppressWithPlainTextCommentFilter filter) {
+            final Pattern filterCommentFormat;
+            if (suppressionType == SuppressionType.ON) {
```
strkkk, line 658: please use switch

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java
```diff
@@ -542,50 +602,97 @@ private static final class Suppression {
...
+            }
+            return filterCommentFormat;
+        }
+
+        /**
+         * Gets influence from suppress filter influence format param.
+         *
+         * @param format          influence format to parse
+         * @param influenceFormat raw influence format
+         * @param text            text of the suppression
```
strkkk, line 675: please remove indentation for description here and above

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java
```diff
@@ -631,7 +740,12 @@ && isIdMatch(event)
          * @return true if {@link AuditEvent} is in the scope of the suppression.
          */
         private boolean isInScopeOfSuppression(AuditEvent event) {
-            return lineNo <= event.getLine();
+            final int eventLine = event.getLine();
+            boolean isInScope = lineNo <= eventLine;
+            if (suppressionType == SuppressionType.DEFAULT) {
+                isInScope = eventLine >= firstLine && eventLine <= lastLine;
+            }
+            return isInScope;
```
strkkk, line 748: may be just use ternary operator and make it
```
return suppressionType == SuppressionType.DEFAULT
  ? eventLine >= firstLine && eventLine <= lastLine
  : lineNo <= eventLine;
```

strkkk (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java
```diff
@@ -468,6 +515,10 @@ private Optional<Suppression> getSuppression(FileText fileText, int lineNo) {
             suppression = new Suppression(offCommentMatcher.group(0),
                 lineNo + 1, offCommentMatcher.start(), SuppressionType.OFF, this);
         }
+        if (commentMatcher.find()) {
```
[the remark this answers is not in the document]
stoyanK7: No, no test case for that yet. I think I should tweak it so that there is no overwrite. Do you think this should be the expected behavior?

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java
stoyanK7: Checkstyle does not allow inline conditionals.

stoyanK7:
Github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java
```diff
@@ -103,6 +108,12 @@
  * Type is {@code java.util.regex.Pattern}.
  * Default value is {@code null}.
  * </li>
+ * <li>
+ * Property {@code influenceFormat} - Specify negative/zero/positive value that
+ * defines the number of lines preceding/at/following the suppression comment.
+ * Type is {@code java.lang.String}.
+ * Default value is {@code "0"}.
```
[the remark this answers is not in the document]
stoyanK7: Done! All the points that have thumbs up are done and committed. Only on 2 I have left comment on that are not done yet.

romani:
>as proof of implementation we need to restore code/comments of suppression at equalsverifier.

From issue. Please create PR to that project. You can use Snapshot version in their pom.

romani:
The more I look at this the more I become confident that we need for this functional new filter, as current filter is about suppression from-to comment.  Existing fields first fine for existing design, but not really for new fields.

But we are talking about "around" comment.

Do you have the same filling ?

stoyanK7:
> The more I look at this the more I become confident that we need for this functional new filter,

Do you mean a completely new filter or just change [SuppressWithNearbyCommentCheck](https://checkstyle.org/config_filters.html#SuppressWithNearbyCommentFilter)'s parent to `Checker`?

---
> Existing fields first fine for existing design, but not really for new fields.
> Do you have the same filling ?

Kind of, it didn't require much tweaking to make it work in this case. Mostly copying behavior from `SuppressWithNearbyCommentFilter`. 
* `SuppressWithNearbyCommentFilter` uses internal class `Tag`
* `SuppressWithPlainTextCommentFilter` uses internal class `Suppression`

but both have similar purpose. I get what you mean though, same thought crossed my mind - it feels like this check is doing too much now.

---
>From issue. Please create PR to that project. You can use Snapshot version in their pom.

Should I still do it or shall we first discuss what to do with your proposal?

---
> Can we we start to use this feature for google-checks.xml ? And may be other files

I think yes, absolutely. Suppressions can be done inline.

romani:
completely new filter.

>Should I still do it or shall we first discuss what to do with your proposal?

Let's finish discussion on design.

>Suppressions can be done inline.

Please do now, to prove that it is good.

>Kind of, it didn't require much tweaking 

Let's think of user experience, not our efforts.

----

@strkkk , what is your thoughts on this?

romani:
@stoyanK7 , if you and me think that new Filter will be more logical and has better design, lets create new issue to describe new Filter and convert this PR to use new filter, I pretty sure we will reuse bunch of code(in some form) from other filters , so it will not be very time consuming.

stoyanK7:
On it. Raising the issue now.

strkkk:
> @strkkk , what is your thoughts on this?

new filter seems fine.

@stoyanK7 I think this PR along with the issue can be closed as work will continue in new issues/PR

PR #11888, opened 2022-07-13
Issue #11720: Kill surviving mutation in CheckstyleAntTask related to logging

Vyom-Yadav:
#11720 
Checkstyle Ant Task documentation: https://checkstyle.sourceforge.io/anttask.html

### Diff Reports:
 N/A
### Rationale:

`antTask.execute()` uses `System.currentTimeMillis()` to calculate the time elapsed, if we note the time before and after the execution of the method, that will be more than or equal to the time calculated inside the method.

Now the order of execution of instructions can change in a multithreaded program as the scheduler controls the threads, but all the related methods run on a single thread only.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
@@ -32,6 +32,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.ResourceBundle;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.tools.ant.BuildException;
 import org.apache.tools.ant.Project;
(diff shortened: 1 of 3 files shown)


Vyom-Yadav:
@nick-mancuso @baratali @rnveach @romani Please review, if this method of killing this mutant is good, then I will replicate this across other similar mutations.

src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
```diff
@@ -66,6 +69,9 @@ public class CheckstyleAntTaskTest extends AbstractPathTestSupport {
     private static final String NOT_EXISTING_FILE = "target/not_existing.xml";
     private static final String FAILURE_PROPERTY_VALUE = "myValue";
 
+    @TempDir
+    public File temporaryFolder;
```
nick-mancuso, line 73: What do we need this for?

nick-mancuso (changes requested):
Item:

src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
Vyom-Yadav: My bad, we don't require it, earlier instead of using `CheckstyleAntTaskLogStub` I was logging to a temporary file, removed it now, done.

nick-mancuso (approved):
Good idea!

romani (approved):
Ok to merge.

This is completely valid implementation of test. We can not make Input like test on this, but took most top level classes to operate in test method.

PR #11898, opened 2022-07-16
Issue #11720: Kill surviving mutation in PackageObjectFactory

Vyom-Yadav:
#11720

### Diff Reports: 
N/A

Missing Test Case Was Added

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
@@ -23,12 +23,14 @@ import static com.google.common.truth.Truth.assertWithMessage;
 import static com.puppycrawl.tools.checkstyle.PackageObjectFactory.AMBIGUOUS_MODULE_NAME_EXCEPTION_MESSAGE;
 import static com.puppycrawl.tools.checkstyle.PackageObjectFactory.BASE_PACKAGE;
 import static com.puppycrawl.tools.checkstyle.PackageObjectFactory.CHECK_SUFFIX;
+import static com.puppycrawl.tools.checkstyle.PackageObjectFactory.ModuleLoadOption.SEARCH_REGISTERED_PACKAGES;
 import static com.puppycrawl.tools.checkstyle.PackageObjectFactory.ModuleLoadOption.TRY_IN_ALL_REGISTERED_PACKAGES;
 import static com.puppycrawl.tools.checkstyle.PackageObjectFactory.NULL_LOADER_MESSAGE;
 import static com.puppycrawl.tools.checkstyle.PackageObjectFactory.NULL_PACKAGE_MESSAGE;
 import static com.puppycrawl.tools.checkstyle.PackageObjectFactory.PACKAGE_SEPARATOR;
 import static com.puppycrawl.tools.checkstyle.PackageObjectFactory.STRING_SEPARATOR;
 import static com.puppycrawl.tools.checkstyle.PackageObjectFactory.UNABLE_TO_INSTANTIATE_EXCEPTION_MESSAGE;
+import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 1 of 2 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
```diff
@@ -491,4 +528,8 @@ protected void processFiltered(File file, FileText fileText) {
 
     }
 
+    private static class MockClass {
+        MockClass() {
```
nick-mancuso, line 532: Do we need explicit default constructor?

nick-mancuso (changes requested):
Item:

src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
Vyom-Yadav: If the class has an access modifier `private` then the default constructor by default also has a `private` access modifier, reflection can find that constructor (we won't get `NoSuchMethodException`) but we will get `IllegalAccessException`. So if we use private then the default constructor either needs to be `public` or with the default access modifier.

src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
```diff
@@ -478,6 +479,42 @@ public void testGenerateThirdPartyNameToFullModuleNameWithException() throws Exc
...
+    }
+
+    @Test
+    public void testCreateModuleWithTryInAllRegisteredPackages() {
+        final ClassLoader classLoader = ClassLoader.getSystemClassLoader();
+        final Set<String> packages = Collections.singleton(BASE_PACKAGE);
+        final PackageObjectFactory objectFactory =
+            new PackageObjectFactory(packages, classLoader, SEARCH_REGISTERED_PACKAGES);
+        try {
+            objectFactory.createModule("PackageObjectFactoryTest$MockClass");
```
nick-mancuso, line 502: Please use assertThrows.

nick-mancuso (changes requested):
Item:

src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
Vyom-Yadav: @nick-mancuso Semaphore is complaining, `setAccessable(true)` won't work as I would have to change the code in `PackageObjectFactory` to do so.

rnveach: https://github.com/checkstyle/checkstyle/blob/399f1b48b79731d65e824eb927971ef6174c3463/src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java#L96

nick-mancuso: Let’s skip this, if constructor is not redundant, I am good.

nick-mancuso (approved):

nick-mancuso (changes requested):
@Vyom-Yadav original implementation of was OK, let’s make CI happy so we can merge this.

Vyom-Yadav:
@nick-mancuso Done, semaphore was giving `The constructor PackageObjectFactoryTest.MockClass() is never used locally` so I changed the access modifier of the class so we don't have to deal with constructor.

nick-mancuso (approved):

nick-mancuso:
Build failure in Travis and no-error-xwiki are not related to changes in this PR.

PR #11904, opened 2022-07-17
Pull #11904: Remove pitest mutation checking HTML model

Vyom-Yadav:
Remove `pitest-survival-check-html.sh` as new XML model covers are useful cases.

### One case not covered by the XML model (as it is based on `mutations.xml`):

The new pitest method only examines `mutations.xml` which isn't appropriate to determine lines that are uncovered and are not mutated (due to the absence of some mutator). Such lines can only be taken from the HTML file.

An example of such lines is present over [here](https://vyom-yadav.github.io/DiffReport/pit-blocks-mod-1/index.html) (modified a bit for demonstrating).

Though uncovered lines will result in low line coverage, it is rounded up in certain cases ([example](https://vyom-yadav.github.io/DiffReport/pit-blocks-mod-single-line/index.html)) and the line coverage becomes 100% :( https://github.com/hcoles/pitest/issues/375 

Mutations on lines that have no coverage are of type `NO_COVERAGE`, such mutations are **supported** by the XML report checker.

But it shouldn't be a problem as we have jacoco for line coverage and if there are multiple uncovered lines, then pitest line coverage will also drop and it will fail.

---

The issue for XML report not containing information about uncovered lines is created at https://github.com/hcoles/pitest/issues/1059

---

In some cases when the number of mutators is limited, some uncovered lines are also mutated and mutants are present in the XML report.
In the mock example below, line 277 is uncovered but has a mutation with status `NO_COVERAGE`
Screenshot from 2022-07-20 10-26-36
Screenshot from 2022-07-20 10-28-53

Lines 274, 275, and 276 are uncovered and nothing about them is present in the XML report.
See the report [here](https://vyom-yadav.github.io/DiffReport/pit-blocks-mod-1/com.puppycrawl.tools.checkstyle.checks.blocks/EmptyBlockCheck.java.html) (Mock Example)


romani:
@Vyom-Yadav, please resolve conflict

Please reference PR number in git commit, to preserve reasons why we removed it.

romani:
Next time please do it as prefix "Pull #xxxxx:"

Vyom-Yadav:
The issue for XML report not containing information about uncovered lines is created at https://github.com/hcoles/pitest/issues/1059

romani:
Discussion can be continued, we can reconsider this decision at any point of time.

PR #12001, opened 2022-07-28
Issue #11821: RequireThisCheck now handles try-with resources

### LINKED ISSUE #11821: RequireThisCheck not able to handle try-with-resources
I have read check documentation: https://checkstyle.sourceforge.io/config_coding.html#RequireThis

```bash
/var/tmp $ javac Test.java
```
```bash
/var/tmp $ cat config.xml
```
```XML
<!DOCTYPE module PUBLIC

<module name="Checker">
  <module name="TreeWalker">
    <module name="RequireThisCheck">
      <property name="validateOnlyOverlapping" value="false" />
    </module>
  </module>
</module>
```
```bash
/var/tmp $ cat Test.java
```
```java
import java.io.BufferedReader;
import java.io.IOException;
import java.io.InputStreamReader;

public class Test {

    public BufferedReader instFld2;

     void method() {
        try (BufferedReader instFld2 = new BufferedReader(new InputStreamReader(null, "utf-8"))) {
        }
        catch (IOException e) {
        }
    }
}
```
```bash
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.3.1-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:10:29: Reference to instance variable 'instFld2' needs "this.". [RequireThis]
Audit done.
```

The violation is given for resource definition, which is not expected.

stoyanK7:
Resolves #11821

---
Diff Regression config: https://gist.githubusercontent.com/stoyanK7/915bf3abd0e42ae89e6e6dc1051fbe02/raw/de67d93f46a866635e725941b296a799ac3c1fc3/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
@@ -587,15 +587,7 @@ public class RequireThisCheck extends AbstractCheck {
                 frameStack.addFirst(new BlockFrame(frame, ast));
                 break;
             case TokenTypes.METHOD_DEF:
-                final DetailAST methodFrameNameIdent = ast.findFirstToken(TokenTypes.IDENT);
-                final DetailAST mods = ast.findFirstToken(TokenTypes.MODIFIERS);
-                if (mods.findFirstToken(TokenTypes.LITERAL_STATIC) == null) {
-                    ((ClassFrame) frame).addInstanceMethod(methodFrameNameIdent);
-                }
-                else {
-                    ((ClassFrame) frame).addStaticMethod(methodFrameNameIdent);
-                }
-                frameStack.addFirst(new MethodFrame(frame, methodFrameNameIdent));
+                collectMethodDeclarations(frameStack, ast, frame);
                 break;
             case TokenTypes.CTOR_DEF:
             case TokenTypes.COMPACT_CTOR_DEF:
@@ -652,6 +644,26 @@ public class RequireThisCheck extends AbstractCheck {
         else {
             frame.addIdent(ident);
         }
+    }
+
+    /**
+     * Collects {@code METHOD_DEF} declarations.
+     *
+     * @param frameStack stack containing the FrameTree being built.
+     * @param ast AST to parse.
+     * @param frame current frame.
+     */
+    private static void collectMethodDeclarations(Deque<AbstractFrame> frameStack,
+                                                  DetailAST ast, AbstractFrame frame) {
+        final DetailAST methodFrameNameIdent = ast.findFirstToken(TokenTypes.IDENT);
+        final DetailAST mods = ast.findFirstToken(TokenTypes.MODIFIERS);
+        if (mods.findFirstToken(TokenTypes.LITERAL_STATIC) == null) {
+            ((ClassFrame) frame).addInstanceMethod(methodFrameNameIdent);
+        }
+        else {
+            ((ClassFrame) frame).addStaticMethod(methodFrameNameIdent);
+        }
+        frameStack.addFirst(new MethodFrame(frame, methodFrameNameIdent));
     }
 
     /**
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
@@ -275,7 +275,8 @@ public class RequireThisCheck extends AbstractCheck {
         TokenTypes.PARAMETER_DEF,
         TokenTypes.TYPE_ARGUMENT,
         TokenTypes.RECORD_DEF,
-        TokenTypes.RECORD_COMPONENT_DEF
+        TokenTypes.RECORD_COMPONENT_DEF,
+        TokenTypes.RESOURCE
     );
     /** Set of all assign tokens. */
     private static final BitSet ASSIGN_TOKENS = TokenUtil.asBitSet(
@@ -362,6 +363,8 @@ public class RequireThisCheck extends AbstractCheck {
             TokenTypes.IDENT,
             TokenTypes.RECORD_DEF,
             TokenTypes.COMPACT_CTOR_DEF,
+            TokenTypes.LITERAL_TRY,
+            TokenTypes.RESOURCE,
         };
     }
 
@@ -406,8 +409,13 @@ public class RequireThisCheck extends AbstractCheck {
             case TokenTypes.RECORD_DEF:
                 current.push(frames.get(ast));
                 break;
+            case TokenTypes.LITERAL_TRY:
+                if (ast.getFirstChild().getType() == TokenTypes.RESOURCE_SPECIFICATION) {
+                    current.push(frames.get(ast));
+                }
+                break;
             default:
-                // do nothing
+                break;
         }
     }
 
@@ -425,8 +433,13 @@ public class RequireThisCheck extends AbstractCheck {
             case TokenTypes.RECORD_DEF:
                 current.pop();
                 break;
+            case TokenTypes.LITERAL_TRY:
+                if (current.peek().getType() == FrameType.TRY_WITH_RESOURCES_FRAME) {
+                    current.pop();
+                }
+                break;
             default:
-                // do nothing
+                break;
         }
     }
 
@@ -556,6 +569,12 @@ public class RequireThisCheck extends AbstractCheck {
                     frame.addIdent(parameterIdent);
                 }
                 break;
+            case TokenTypes.RESOURCE:
+                final DetailAST resourceIdent = ast.findFirstToken(TokenTypes.IDENT);
+                if (resourceIdent != null) {
+                    frame.addIdent(resourceIdent);
+                }
+                break;
             case TokenTypes.CLASS_DEF:
             case TokenTypes.INTERFACE_DEF:
             case TokenTypes.ENUM_DEF:
@@ -599,6 +618,11 @@ public class RequireThisCheck extends AbstractCheck {
                 if (isAnonymousClassDef(ast)) {
                     frameStack.addFirst(new AnonymousClassFrame(frame,
                             ast.getFirstChild().toString()));
+                }
+                break;
+            case TokenTypes.LITERAL_TRY:
+                if (ast.getFirstChild().getType() == TokenTypes.RESOURCE_SPECIFICATION) {
+                    frameStack.addFirst(new TryWithResourcesFrame(frame, ast));
                 }
                 break;
             default:
@@ -653,6 +677,11 @@ public class RequireThisCheck extends AbstractCheck {
                 break;
             case TokenTypes.LITERAL_NEW:
                 if (isAnonymousClassDef(ast)) {
+                    frames.put(ast, frameStack.poll());
+                }
+                break;
+            case TokenTypes.LITERAL_TRY:
+                if (ast.getFirstChild().getType() == TokenTypes.RESOURCE_SPECIFICATION) {
                     frames.put(ast, frameStack.poll());
                 }
                 break;
@@ -1228,6 +1257,8 @@ public class RequireThisCheck extends AbstractCheck {
         CATCH_FRAME,
         /** For frame type. */
         FOR_FRAME,
+        /** Try with resources frame type. */
+        TRY_WITH_RESOURCES_FRAME
 
     }
 
@@ -1676,6 +1707,24 @@ public class RequireThisCheck extends AbstractCheck {
             return FrameType.CATCH_FRAME;
         }
 
+        @Override
+        protected AbstractFrame getIfContains(DetailAST identToFind, boolean lookForMethod) {
+            final AbstractFrame frame;
+
+            if (!lookForMethod
+                    && containsFieldOrVariable(identToFind)) {
+                frame = this;
+            }
+            else if (getParent().getType() == FrameType.TRY_WITH_RESOURCES_FRAME) {
+                // Skip try-with-resources frame because resources cannot be accessed from catch
+                frame = getParent().getParent().getIfContains(identToFind, lookForMethod);
+            }
+            else {
+                frame = getParent().getIfContains(identToFind, lookForMethod);
+            }
+            return frame;
+        }
+
     }
 
     /**
@@ -1696,6 +1745,29 @@ public class RequireThisCheck extends AbstractCheck {
         @Override
         public FrameType getType() {
             return FrameType.FOR_FRAME;
+        }
+
+    }
+
+    /**
+     * A frame initiated on entering a try-with-resources construct;
+     * holds local resources for the try block.
+     */
+    private static class TryWithResourcesFrame extends AbstractFrame {
+
+        /**
+         * Creates try-with-resources frame.
+         *
+         * @param parent parent frame.
+         * @param ident ident frame name ident.
+         */
+        protected TryWithResourcesFrame(AbstractFrame parent, DetailAST ident) {
+            super(parent, ident);
+        }
+
+        @Override
+        public FrameType getType() {
+            return FrameType.TRY_WITH_RESOURCES_FRAME;
         }
 
     }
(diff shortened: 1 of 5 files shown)


stoyanK7:
Github, generate report

romani:
@stoyanK7 , each time to get unexpected result in regression testing, please copy input code to UTs to not rely on regression much.

nick-mancuso:
@stoyanK7 please create entirely new test input instead of extending existing tests. Also, please update documentation with new tokens and example of new behavior.

stoyanK7:
@nick-mancuso Will do! I am struggling with those surviving mutations a bit. Any tips on how to deal with them? I'm pretty sure I have to come up with more test cases to cover them buy I can't think of anything.

nick-mancuso:
@stoyanK7 address other CI failures first (NPEs in both openjdk and checkstyle), these might help you to refine code a bit.

romani:
Still pitest red CI

romani:
Github, generate report

stoyanK7:
I know, just fixed the NPEs and now running the pitest locally trying to fix it.

romani:
Github, generate report

romani:
I didn't spot any false positives in report. Looks good.

stoyanK7:
Documentation on [Experimental Switch Mutator](https://pitest.org/quickstart/mutators/#EXPERIMENTAL_SWITCH) is really scarce and I still cannot wrap my head around what it exactly does so I can reproduce behavior. Any ideas?

I noticed other experimental switch mutators are suppressed.

romani:
@nick-mancuso , @Vyom-Yadav , @rnveach ,  @pbludov , please suggest.

stoyanK7:
Alright, turns out documentation does not include this mutator and I was looking at the wrong one. This is what `RemoveSwtich` mutator does:
```java
/**
 * Remove switch statements. We get an array of labels to jump to, plus a
 * default label. We change each label to the default label, thus removing it
 */
public class RemoveSwitchMutator implements MethodMutatorFactory {
```

Taken from [RemoveSwitchMutator.java](https://github.com/hcoles/pitest/blob/master/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/RemoveSwitchMutator.java). I will look at the tests and try to understand it better.

stoyanK7:
I will push my latest work. Pitest would still not pass. Is that okay?

Vyom-Yadav:
@stoyanK7 Good investigation, but keep in mind that pitest operates upon bytecode and in this case, the bytecode (decompiled) will be-
```java
public void leaveToken(DetailAST ast) {                                                                                  
    switch (ast.getType()) {                                                                                             
        case 7:                                                                                                          
        case 8:                                                                                                          
        case 9:                                                                                                          
        case 14:                                                                                                         
        case 15:                                                                                                         
        case 91:                                                                                                         
        case 154:                                                                                                        
        case 157:                                                                                                        
        case 199:                                                                                                        
            this.current.pop();                                                                                          
            break;                                                                                                       
        case 95:                                                                                                         
            if (((AbstractFrame)this.current.peek()).getType() == RequireThisCheck.FrameType.TRY_WITH_RESOURCES_FRAME) { 
                this.current.pop();                                                                                      
            }                                                                                                            
    }                                                                                                                    
                                                                                                                         
}                                                                                                                        
```
The surviving mutation:
```
Source File: "RequireThisCheck.java"
Class: "com.puppycrawl.tools.checkstyle.checks.coding.RequireThisCheck"
Method: "leaveToken"
Line Contents: "switch (ast.getType()) {"
Mutator: "org.pitest.mutationtest.engine.gregor.mutators.experimental.RemoveSwitchMutator_9"
Description: "RemoveSwitch 9 mutation"
Line Number: 424
```
is for `RemoveSwitch 9`, in the decompiled bytecode, the 9th case is `case 199:`, the mutated bytecode(decompiled) will be-
```java
public void leaveToken(DetailAST ast) {                                                                                 
    switch (ast.getType()) {                                                                                            
        case 7:                                                                                                         
        case 8:                                                                                                         
        case 9:                                                                                                         
        case 14:                                                                                                        
        case 15:                                                                                                        
        case 91:                                                                                                        
        case 154:                                                                                                       
        case 157:                                                                                                       
            this.current.pop();                                                                                         
            break;                                                                                                      
        case 95:                                                                                                        
            if (((AbstractFrame)this.current.peek()).getType() == RequireThisCheck.FrameType.TRY_WITH_RESOURCES_FRAME) {
                this.current.pop();                                                                                     
            }                                                                                                           
        case 199:                                                                                                       
    }                                                                                                                   
                                                                                                                        
}                                                                                                                       
```
`199` stands for `TokenTypes.RECORD_DEF` so source code will look like:
```java
public void leaveToken(DetailAST ast) {                                          
    switch (ast.getType()) {                                                     
        case TokenTypes.CLASS_DEF:                                               
        case TokenTypes.INTERFACE_DEF:                                           
        case TokenTypes.ENUM_DEF:                                                
        case TokenTypes.ANNOTATION_DEF:                                          
        case TokenTypes.SLIST:                                                   
        case TokenTypes.METHOD_DEF:                                              
        case TokenTypes.CTOR_DEF:                                                
        case TokenTypes.LITERAL_FOR:                                             
            current.pop();                                                       
            break;                                                               
        case TokenTypes.LITERAL_TRY:                                             
            if (current.peek().getType() == FrameType.TRY_WITH_RESOURCES_FRAME) {
                current.pop();                                                   
            }                                                                    
            break;                                                               
        case TokenTypes.RECORD_DEF:                                              
        default:                                                                 
            // do nothing                                                        
    }                                                                            
}                                                                                
```
And with these changes, I ran the test suite of this PR (After `gh pr checkout 12001`) and none of the tests failed, so this is the mutation you were looking for :)

stoyanK7:
Thank you for the detailed explanation @Vyom-Yadav! There is no way I could have figured that out this quick. I'm pretty sure this mutation kill should be a seperate commit/PR now, my PR has nothing to do with records(I can still try to deal with it if you insist though). Moreover, `RemoveSwitch 6` no longer needs to be suppressed which makes me think that the suppressions shifted with me adding `case TokenTypes.LITERAL_TRY` in the switch statement and this is the reason pitest is failing. Am I right?

Vyom-Yadav:
@stoyanK7 Could be, but you would have to check that, please check that the new mutation is logically the same as the previous mutations, you can use the method I have used above to verify, after that just share the results so we have it documented that we do not have a new surviving mutation.

stoyanK7:
Did a quick test and here are the results:
| TokenType      | Tests fail if removed from switch statement - Master branch | Tests fail if removed from switch statement - Pull Request branch
| ----------- | ----------- | ----------- |
| CLASS_DEF      | 🔴 No       | 🔴 No |
| INTERFACE_DEF      | 🔴 No       | 🔴 No |
| ENUM_DEF      | 🔴 No       | 🔴 No |
| ANNOTATION_DEF      | 🔴 No       | 🔴 No |
| SLIST      | 🟢 Yes       | 🟢 Yes |
| METHOD_DEF      | 🔴 No       | 🔴 No  |
| CTOR_DEF      | 🔴 No       | 🔴 No  |
| LITERAL_FOR      | 🟢 Yes       | 🟢 Yes |
| RECORD_DEF      | 🔴 No       | 🔴 No |
| LITERAL_TRY | Does not exist | 🟢 Yes

romani:
Thanks a lot for green CI 
Can you point to test that killed pitest survival? We are learning advanced pitest now

stoyanK7:
The test that killed the `case TokenTypes.LITERAL_TRY` mutation was `RequireThisCheck.testTryWithResourcesOnlyOverlappingFalse()`.

Removing the case
Screenshot from 2022-08-04 09-54-46
fails the test
Screenshot from 2022-08-04 09-56-50
Screenshot from 2022-08-04 09-59-24

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -1676,6 +1708,24 @@ public FrameType getType() {
             return FrameType.CATCH_FRAME;
         }
 
+        @Override
+        protected AbstractFrame getIfContains(DetailAST nameToFind, boolean lookForMethod) {
```
romani, line 1712: >DetailAST nameToFind

"Name" vs ast confuse my brain. Can we have different name?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisTryWithResourcesOnlyOverlappingFalse.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    void threeResourcesReferencingEachOther() {
+        try (InputStreamReader isr = new InputStreamReader(null, "utf-8"); // ok
+             BufferedReader br = new BufferedReader(isr); // ok
+             Scanner sc = new Scanner(isr.toString() + br.toString())) { } // ok
+        catch (IOException e) { }
+    }
+
+    private String charset = "utf-8";
```
romani, line 39: It took me a while to find this field.
Is it required to have it here? Outside of other Fields

romani (changes requested):
Items

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisTryWithResourcesOnlyOverlappingFalse.java
romani: Can we name all fields with prefix "field" to be more explicit

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
stoyanK7: `identToFind`? In the end, `IDENT`s are being compared.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisTryWithResourcesOnlyOverlappingFalse.java
stoyanK7: > with prefix "field"

No reason apart from usage really, will move it back with the other fields on the top.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -538,6 +551,7 @@ private static boolean isInCompactConstructor(DetailAST ast) {
      * @param frameStack stack containing the FrameTree being built.
      * @param ast AST to parse.
      */
+    // -@cs[ExecutableStatementCount] This method is a big switch and is too hard to remove.
```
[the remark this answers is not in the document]
stoyanK7: Alright, will do. I noticed that there are numerous other files that suppress both `ExecutableStatementCount` and `JavaNCSS` at the same time so I thought why not.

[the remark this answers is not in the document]
stoyanK7: >  Please move case bodies to methods. It might make sense to move out methoddef

Shall I extract all cases or only `METHOD_DEF`?

.ci/jsoref-spellchecker/whitelist.words
```diff
@@ -629,6 +629,7 @@ ioffset
 IPonr
 IPv
 isindex
+isr
```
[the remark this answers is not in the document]
stoyanK7: Done! All other points as well.

Vyom-Yadav:
@stoyanK7 If you need help regarding anything, ping me, I go by pings.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisTryWithResourcesOnlyOverlappingFalse.java
stoyanK7: Oops. Forgot this one in one of the files. Done.

stoyanK7: CI failure looks unrelated, yes?

stoyanK7:
@Vyom-Yadav  I will keep this in mind.

I noticed that you are currently killing mutations left and right. You probably know this already but here is an idea of how to kill `RemoveSwitch` mutations  in `RequireThisCheck.leaveToken()`:

---
`RemoveSwitch` for both `case TokenTypes.METHOD_DEF` and `case TokenTypes.CLASS_DEF` .
Here you need `validateOnlyOverlapping` set to `false`. The following test case fails when either switch-case is removed.
```java
class MethodDefMutation {
    private int x;
    private int y;
    private int z;
    void method(int x) {
        class Local {
            void method2(int y) {
                class Local2 {
                    void method3(int z) {
                        int c = x + y + z;
                    }
                }
                int b = x + y + z; // violation
            }
        }
        int a = x + y + z; // 2 violations
    }
}


```
---

`RemoveSwitch` for both `case TokenTypes.INTERFACE_DEF`. Here you also need `validateOnlyOverlapping` set to `false`.
```java
interface Interface1 {
    int x();
    int y();
    int z();
    default void method() {
        interface Interface2 {
            static int x();
            default void method2() {
                interface Interface3 {
                    static int y();
                    default void method3() {
                        int c = x() + y() + z(); // violation
                    }
                }
                int b = x() + y() + z(); // 2 violations
            }
        }
        int a = x() + y() + z(); // 3 violations
    }
}


```

---
The idea is that you need to nest tokens. Here is a commit with the test cases - https://github.com/stoyanK7/checkstyle/commit/1d2f0cd90ca83acdf8ac9394c4aad0b16895f967 . Hope this helps.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -1303,29 +1346,29 @@ public DetailAST getFrameNameIdent() {
...
         /**
          * Check whether the frame contains a given name.
          *
-         * @param nameToFind IDENT ast of the name we're looking for.
+         * @param identToFind IDENT ast of the name we're looking for.
          * @param lookForMethod whether we are looking for a method name.
          * @return whether it was found.
          */
-        protected AbstractFrame getIfContains(DetailAST nameToFind, boolean lookForMethod) {
+        protected AbstractFrame getIfContains(DetailAST identToFind, boolean lookForMethod) {
```
romani, line 1363: Please move all such renaming in separate commit of this PR, message "minor: renaming to make it clear"
We try our best to not mix functional changes and refactoring.

Please move change for method creation to avoid statement count violation, in separate commit also - "minor: move code to method to avoid violation from zxxx in next commit"

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisTryWithResources.java
```diff
@@ -0,0 +1,114 @@
...
+
+package com.puppycrawl.tools.checkstyle.checks.coding.requirethis;
+
+import java.io.*;
+import java.util.Scanner;
+
+public class InputRequireThisTryWithResources implements AutoCloseable{
+    private BufferedReader bufferedReader;
+    private InputStreamReader streamReader;
+    private Scanner scanner;
```
romani, line 18: Please do me big favor and name Fields by some notation "fieldXxxxx" or "fldXxxx"or "mXxxx". To easily catch reason of violation.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
stoyanK7: Done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisTryWithResources.java
stoyanK7: Done

romani (approved):
Ok to merge

pbludov (approved):

Issue #11865, opened 2022-07-08
Resolve IDEA violations from Teamcity 2022.04 (Bundled IntelliJ IDEA 2022.1.3)


romani:
https://teamcity.jetbrains.com/buildConfiguration/Checkstyle_IdeaInspectionsMaster/3892633?buildTab=Inspection

Java/Code maturity
[Deprecated member is still used (1)](https://teamcity.jetbrains.com/viewLog.html?tab=Inspection&buildId=3892633&embedded=true#)
Java/Javadoc
[Link specified as plain text (20)](https://teamcity.jetbrains.com/viewLog.html?tab=Inspection&buildId=3892633&embedded=true#)
Java/Verbose or redundant code constructs
[Multiple occurrences of the same expression (2)](https://teamcity.jetbrains.com/viewLog.html?tab=Inspection&buildId=3892633&embedded=true#)
Markdown
[Incorrect table formatting (1)](https://teamcity.jetbrains.com/viewLog.html?tab=Inspection&buildId=3892633&embedded=true#)
RegExp
[Regular expression can be simplified (36)](https://teamcity.jetbrains.com/viewLog.html?tab=Inspection&buildId=3892633&embedded=true#)

pbludov:
The last one is dubious:
![изображение](https://user-images.githubusercontent.com/900805/178112176-aa0f1a28-3aac-4c89-901d-b9d194ab23cf.png)

Maybe `'\n'` will be suitable here.

romani:
> Dmitry Treskunov (TeamCity)                                                                    Jul 8, 2022, 19:31 GMT+3
Hi, I've just installed 2021.2.3, could you please try again?
Dmitry Treskunov, TeamCity Team Lead

romani:
This ticket will be good task for migration of TC from UI to config.
to show how we can smoootly fix all violations and update to new Inspection in same PR.

romani:
Sonar plugin execution is hanging https://teamcity-support.jetbrains.com/hc/en-us/requests/4153345

romani:
Student can download from Jetbrains exactly this version of IDEA, and run all on local.

Vyom-Yadav:
Getting a 404 on all links

nick-mancuso:
Yes, whatever build they were obtained from has probably been removed. We can just run it locally to get the same result.

Vyom-Yadav:
[...]

PR #12014, opened 2022-08-01
Issue #11865: solve regexp can be simplified

pbludov:
Part of #11865

All `[A-Za-z0-9_]` replaced with `\w`, `[0-9]` replaced with `\d`.

Also, `[A-Za-z0-9]` replaced with `[A-Za-z\d]`. It looks a bit strange for me, but the inspection works in this way.


pbludov:
The PR is ready for review. I'm not going to assign anyone till GSOC ends.

romani:
@pbludov , please resolve conflict

romani (changes requested):
Ok

romani (approved):
Okok

nick-mancuso (commented):
I am OK to update java code for this inspection, since all our java code is well tested and regexp in Java is system agnostic. Also, we have the intellij inspection to make `\w` and `\d` consistent in our code base (at least in some of our codebase, this is another discussion). But, I am not a big fan of migrating all scripts to this, as I have doubts about it's portability (both for OS and grep/perl/awk, etc.), and we have no way to enforce new format going forward. Also, if we match some unicode digit by accident in a script (`\d` is **not** equivalent to [0-9]), this will be very difficult to debug, especially in cases where we are using the result as an integer. Additionally, by using `\d` in cases like pmd inspections (such as `classPattern`), we are  extending allowed naming conventions in our project. Example: `MyClass८৮੮૮୮௮౮೮` is now a valid identifier. @romani @pbludov are we sure about this update?

pbludov:
I don't like these changes either. Let's drop this PR.

romani:
@pbludov , can we revert \d ? And keep rule suppresed or disabled

rnveach:
@romani @pbludov Is there any piece of this PR we are going to salvage?

romani:
I think we can close PR as even author do not like it, and too much "false positives" or "wont fix".
in case somebody want to continue, we can reopen this PR or reuse branch.

Issue #12017, opened 2022-08-02
Kill surviving mutations in pitest-api


Vyom-Yadav:
Kill all new surviving mutations in `pitest-api`.
See https://github.com/checkstyle/checkstyle/blob/master/.ci/pitest-suppressions/pitest-api-suppressions.xml

romani:
closing in favor of https://github.com/checkstyle/checkstyle/issues/13672

PR #12021, opened 2022-08-02
Issue #12017: Kill surviving mutations in AbstractCheck for tabWidth

Vyom-Yadav:
#12017 

### Rationale:
No need to specifically initialize `tabWidth` to a default value, the default value is inherited from the parent module. Only the root modules need to explicitly instantiate `tabWidth` with a default value (Only `Checker` needs to instantiate the field)

https://github.com/checkstyle/checkstyle/blob/a54849afccd9cdc371c6040398d27225947debe0/src/main/java/com/puppycrawl/tools/checkstyle/api/AutomaticBean.java#L256-L265

### DIFF
--- a/.ci/pitest-suppressions/pitest-api-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-api-suppressions.xml
@@ -3,15 +3,6 @@
   <mutation unstable="false">
     <sourceFile>AbstractCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.api.AbstractCheck</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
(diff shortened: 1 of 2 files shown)


nick-mancuso (approved):

rnveach:
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L454
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L475

> default value is inherited from the parent module

Correct, it is the responsibility of the parent to pass on default properties to the children. You can see the same code in `TreeWalker` as well.

> Only the root modules need to explicitly instantiate tabWidth with a default value (Only Checker needs to instantiate the field)

Unless the children want to override it and do something different for some reason.

However, it is up to each child to pass it along to their own children. Hence me pointing out you can see similar code in `TreeWalker`. If `TreeWalker` did nothing, the children wouldn't get the new value.

rnveach (approved):

nick-mancuso:
Drone is pending for days, let’s merge this.

PR #12022, opened 2022-08-02
Issue #11758: Adding CheckForVariable function to util.sh

### LINKED ISSUE #11758: Adding CheckForVariable function to util.sh
Whenever we add a token or key, it is important to check if it exists before we run the shell script.

Currently, I am using this function in the release project but as we use a lot of tokens in different shell scripts, we need to make a general function in util.sh so that we can call this function and remove repetition of creating this function.

Function:

```
checkForVariable() {
    if [[ "$(declare -p "$1")" =~ ^declare\ -x ]]; then
       echo "Error: Define $1 environment variable"
       exit 1
    fi
}
```

Usage:
`checkForVariable "TOKEN_NAME"`

Rahulkhinchi03:
Fixes #11758

### DIFF
--- a/.ci/prepare-settings.sh
+++ b/.ci/prepare-settings.sh
@@ -1,16 +1,10 @@
 #!/bin/bash
 set -e
 
+source ./.ci/util.sh
+
 mkdir -p .ci-temp
 cp .ci/release-settings.xml .ci-temp/
-
-checkForVariable() {
-  VAR_NAME=$1
-  if [ ! -v "$VAR_NAME" ]; then
-    echo "Error: Define $1 environment variable"
-    exit 1
-  fi
-}
 
 checkForVariable "SONATYPE_USER"
 checkForVariable "SONATYPE_PWD"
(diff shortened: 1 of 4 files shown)


.ci/prepare-settings.sh
```diff
@@ -4,14 +4,6 @@ set -e
 mkdir -p .ci-temp
 cp .ci/release-settings.xml .ci-temp/
 
-checkForVariable() {
```
romani, line 7: Add `source ./.ci/util.sh`

romani (changes requested):
Items

.ci/prepare-settings.sh
Rahulkhinchi03: Done changes.

romani: Not done, recheck push and share testing of this script

Rahulkhinchi03: Done

Rahulkhinchi03: Done

Rahulkhinchi03: sdfjnsd

Testing: Getting the ERROR of checkForVariable. Working.

romani: Now I see `source`

romani (approved):
Ok to merge

PR #12026, opened 2022-08-03
Issue #12017: Kill surviving mutations in AbstractFileSetCheck related to initializing members

Vyom-Yadav:
#12017 

Logic same as #12021


nick-mancuso (approved):

rnveach:
GitHub, generate site

baratali (approved):
It doesn't seem that `fileExtension` value is inherited by child modules because it's not defined in the context:
https://github.com/checkstyle/checkstyle/blob/0666d2e379f3b813085a9c18e1e3548204e1d6c7/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L449-L455
However, it still works because `CommonUtil.matchesFileExtension(file, fileExtensions)` returns true if `fileExtensions == null`, so if Checker passes a file then its child will process this file even if `fileExtensions == null`.

PR #11529, opened 2022-04-09
infra: run checkstyle in ru locale

romani:
related to https://github.com/checkstyle/checkstyle/pull/11280

PR #12031, opened 2022-08-04
Issue #12017: Kill surviving mutations in AbstractFileSetCheck related to custom message keys

Vyom-Yadav:
#12017

### This mutation falls under the category:

- Logic was analyzed to develop the test case.

In general, we make a new test case, but I didn't create one as the check logic coverage won't be affected by modifying the test configuration.

### DIFF
--- a/.ci/pitest-suppressions/pitest-api-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-api-suppressions.xml
@@ -37,15 +37,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>AbstractFileSetCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.api.AbstractFileSetCheck</mutatedClass>
-    <mutatedMethod>log</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to java/util/Map::get</description>
-    <lineContent>getCustomMessages().get(key)));</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>AbstractViolationReporter.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.api.AbstractViolationReporter</mutatedClass>
     <mutatedMethod>setSeverity</mutatedMethod>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/NewlineAtEndOfFileCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/NewlineAtEndOfFileCheckTest.java
@@ -110,8 +110,9 @@ public class NewlineAtEndOfFileCheckTest
 
     @Test
     public void testNoNewlineAtEndOfFile() throws Exception {
+        final String msgKeyNoNewlineEof = "File does not end with a newline :)";
         final String[] expected = {
-            "1: " + getCheckMessage(MSG_KEY_NO_NEWLINE_EOF),
+            "1: " + msgKeyNoNewlineEof,
         };
         verifyWithInlineConfigParser(
                 getPath("InputNewlineAtEndOfFileNoNewline2.java"),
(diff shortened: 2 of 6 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/checks/NewlineAtEndOfFileCheckTest.java
```diff
@@ -110,8 +110,9 @@ public void testNoNewlineLfAtEndOfFile() throws Exception {
 
     @Test
     public void testNoNewlineAtEndOfFile() throws Exception {
+        final String msgKeyNoNewlineEof = "File does not end with a newline :)";
```
rnveach, line 113: This should cause issues with localization because this now means this test can only pass on an english machine if we do not enforce a locale during the run.

Vyom-Yadav: I can annotate it with `@DefaultLocale` to enforce the locale as after the annotated element has been executed the initial default value is restored.

There are similar pieces of code already present: https://github.com/checkstyle/checkstyle/blob/74d8802acfb8833cd34538447d0852baa58e6c5b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/MatchXpathCheckTest.java#L112-L120

Can't a machine in a different locale understand strings in different languages? Like I can write a string in Chinese on my locale. I am not sure how failure will occur. 

I executed the test with `@DefaultLocale("zh-Hant-TW")` and `@DefaultLocale(language = "fr")`, both the times the test passed.

rnveach: Locale testing may not fully work until https://github.com/checkstyle/checkstyle/pull/11529 .

> Can't a machine in a different locale understand strings in different languages? 

The issue is your assertion is only looking for english, while the check execution may use the current locale and we pull in messages from different message files based on the locale requested. If we pull in french messages and compare then to your english message, the assertion will fail even though there is no "issue".

> There are similar pieces of code already present

`InputMatchXpathDoubleBrace.java` uses a custom message that is hardcoded in english in the configuration. This does not count.

rnveach: I guess I missed this, but I see your using a hardcoded messsage in the configuration for your tests.

Why do we need a hardcoded message here?

nick-mancuso:
@Vyom-Yadav please make CI happy

.ci/pitest-suppressions/pitest-api-suppressions.xml
```diff
@@ -18,15 +18,6 @@
     <lineContent>return Arrays.copyOf(fileExtensions, fileExtensions.length);</lineContent>
   </mutation>
 
-  <mutation unstable="false">
```
nick-mancuso, line 21: Why did we need two tests in two different checks to kill one mutation?

nick-mancuso (changes requested):
Question:

.ci/pitest-suppressions/pitest-api-suppressions.xml
Vyom-Yadav: > kill one mutation?

There are **two** mutations, one for:
https://github.com/checkstyle/checkstyle/blob/696cd880e83b6c0af3171bf003ebb142eb4c0d35/src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractFileSetCheck.java#L213-L224

and the other for:
https://github.com/checkstyle/checkstyle/blob/696cd880e83b6c0af3171bf003ebb142eb4c0d35/src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractFileSetCheck.java#L226-L242

But its the limitation of the survival checking method that it cannot differentiate between the two.

nick-mancuso: Please add this explanation to PR description.

Vyom-Yadav: Done

nick-mancuso (approved):

nick-mancuso:
@Vyom-Yadav please resolve conflicts

Vyom-Yadav:
> @Vyom-Yadav please resolve conflicts

Done.

nick-mancuso (approved):

PR #12045, opened 2022-08-09
minor: Fix sorting issue in pitest survival checking script

Vyom-Yadav:
The problem was with the `setDifference(..)` method, both `setOne` and `setTwo` it received were sorted, but as result was a `HashSet`, and the sorting had no effect.

The correct order of elements is displayed now, moreover instead of sorting elements earlier, we sort them while calculating `setDifference(..)`, `setOne` and `setTwo` are of type `HashSet` instead of `TreeSet`, this also executes `setTwo.contains(mutation)` in `O(1)` (on an avg.)


rnveach (approved):
Default sorting is: source file, mutated class, mutated method, line content, mutator, description.

nick-mancuso (approved):

PR #12046, opened 2022-08-10
Issue #11865: Resolve link specefied as plain text idea violations 

Vyom-Yadav:
#11865 

`no_old_refs.sh` was also modified to support anchor tags.

### DIFF
--- a/.ci/no_old_refs.sh
+++ b/.ci/no_old_refs.sh
@@ -20,7 +20,7 @@ if [ -n "$PR_HEAD_REPO_NAME" ]; then
 fi
 
 # collect issues where full link is used
-grep -IPonr "(after|[Tt]il[l]?) $GITHUB_HOST/[\w.-]+/[\w.-]+/issues/\d{1,5}" . \
+grep -IPonr "(after|[Tt]il[l]?) (<a href=\")?$GITHUB_HOST/[\w.-]+/[\w.-]+/issues/\d{1,5}" . \
   | sed -e 's/:[a-zA-Z].*github.com\//:/' >> $MENTIONED_ISSUES_GREP_OUTPUT
 
 # collect checkstyle issues where only hash sign is used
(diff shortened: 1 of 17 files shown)


nick-mancuso:
Github, generate site

.ci/no_old_refs.sh
```diff
@@ -20,7 +20,7 @@ if [ -n "$PR_HEAD_REPO_NAME" ]; then
 fi
 
 # collect issues where full link is used
-grep -IPonr "(after|[Tt]il[l]?) $GITHUB_HOST/[\w.-]+/[\w.-]+/issues/\d{1,5}" . \
+grep -IPonr "(after|[Tt]il[l]?) (<a href=\")?$GITHUB_HOST/[\w.-]+/[\w.-]+/issues/\d{1,5}" . \
```
romani, line 23: if  it optional , why we need this in regexp ?
lets keep regexp simple, please restore to original version.

romani (changes requested):
items:

.ci/no_old_refs.sh
Vyom-Yadav: No violation:
`// Until https://github.com/checkstyle/checkstyle/pull/12048`

IDEA inspections violation (for javadoc):
```
/**                                                         
 * Until https://github.com/checkstyle/checkstyle/pull/12048
 */                  
```
Need to change it to:
```
/**                                                                              
 * Until <a href="https://github.com/checkstyle/checkstyle/pull/12048">#12048</a>
 */                                                                              
```

Vyom-Yadav: It is used in some cases, in some, it isn't, so we need to keep it.

romani: .ci/no_old_refs.sh does not care about IDEA violations, it just search all urls.

if you remove change in this file, what bad will happen ?

Vyom-Yadav: It cannot find `Until <a href="https://github.com/checkstyle/checkstyle/pull/12048">#12048</a>` without the modification.

> what bad will happen ?

If we have an old ref, it won't be able to determine it.

romani: I see

romani (approved):
ok to merge

nick-mancuso (approved):

PR #12051, opened 2022-08-11
minor: Exclude .teamcity dir from ant-phase-verify.xml and spellchecker

Vyom-Yadav:
Excluded as TeamCity config is generated and contains embedded CDATA


nick-mancuso (approved):
As TC config is generated and contains embedded CDATA, I am OK to keep permanent suppression on these files.

romani (approved):

PR #12052, opened 2022-08-11
Issue #11973: Fix NPE on empty switch statements in VariableDeclarati…

### LINKED ISSUE #11973: VariableDeclarationUsageDistance throws NPE on empty switch statements
I have read check documentation: https://checkstyle.org/config_coding.html#VariableDeclarationUsageDistance

```bash
/var/tmp $ javac Test.java
```
```
/var/tmp $ cat config.xml
```
```XML
<!DOCTYPE module PUBLIC

<module name="Checker">
  <module name="TreeWalker">
    </module>
  </module>
</module>
```
```
/var/tmp $ cat Test.java
```
```java
public class Test {
    void method() {
        int i = -1;
        switch (i) {
        }
    }
}
```
```
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.3.1-all.jar -c config.xml Test.java
Starting audit...
com.puppycrawl.tools.checkstyle.api.CheckstyleException: Exception was thrown while processing Test.java
        at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:307)
        at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:224)
        at com.puppycrawl.tools.checkstyle.Main.runCheckstyle(Main.java:407)
        at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:330)
        at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:189)
        at com.puppycrawl.tools.checkstyle.Main.main(Main.java:126)
Caused by: java.lang.NullPointerException
        at com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck.getFirstNodeInsideSwitchBlock(VariableDeclarationUsageDistanceCheck.java:836)
        at com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck.calculateDistanceBetweenScopes(VariableDeclarationUsageDistanceCheck.java:651)
        at com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck.visitToken(VariableDeclarationUsageDistanceCheck.java:410)
        at com.puppycrawl.tools.checkstyle.api.AbstractFileSetCheck.process(AbstractFileSetCheck.java:92)
        at com.puppycrawl.tools.checkstyle.Checker.processFile(Checker.java:333)
        at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:294)
        ... 5 more
```

stoyanK7:
Resolves #11973

---
Diff Regression config: https://gist.githubusercontent.com/stoyanK7/c67715129a58ae4891e4311252de8ea2/raw/01e5f6b1844ceef6fab5c7f14030ecfa12281f48/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
@@ -809,18 +809,8 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
      */
     private static DetailAST getFirstNodeInsideSwitchBlock(
             DetailAST block, DetailAST variable) {
-        final DetailAST currentNode = getFirstCaseGroupOrSwitchRule(block);
         final List<DetailAST> variableUsageExpressions =
-                new ArrayList<>();
-
-        // Checking variable usage inside all CASE_GROUP and SWITCH_RULE ast's.
-        TokenUtil.forEachChild(block, currentNode.getType(), node -> {
-            final DetailAST lastNodeInCaseGroup =
-                node.getLastChild();
-            if (isChild(lastNodeInCaseGroup, variable)) {
-                variableUsageExpressions.add(lastNodeInCaseGroup);
-            }
-        });
+                getVariableUsageExpressionsInsideSwitchBlock(block, variable);
 
         // If variable usage exists in several related blocks, then
         // firstNodeInsideBlock = null, otherwise if variable usage exists
@@ -835,15 +825,32 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
     }
 
     /**
-     * Helper method for getFirstNodeInsideSwitchBlock to return the first CASE_GROUP or
-     * SWITCH_RULE ast.
+     * Helper method for getFirstNodeInsideSwitchBlock to return all variable
+     * usage expressions inside a given switch block.
      *
      * @param block the switch block to check.
-     * @return DetailAST of the first CASE_GROUP or SWITCH_RULE.
+     * @param variable variable which is checked for in switch block.
+     * @return List of usages or empty list if none are found.
      */
-    private static DetailAST getFirstCaseGroupOrSwitchRule(DetailAST block) {
-        return Optional.ofNullable(block.findFirstToken(TokenTypes.CASE_GROUP))
-            .orElseGet(() -> block.findFirstToken(TokenTypes.SWITCH_RULE));
+    private static List<DetailAST> getVariableUsageExpressionsInsideSwitchBlock(DetailAST block,
+                                                                            DetailAST variable) {
+        final Optional<DetailAST> firstToken = TokenUtil.findFirstTokenByPredicate(block, child -> {
+            return child.getType() == TokenTypes.SWITCH_RULE
+                    || child.getType() == TokenTypes.CASE_GROUP;
+        });
+
+        final List<DetailAST> variableUsageExpressions = new ArrayList<>();
+
+        firstToken.ifPresent(token -> {
+            TokenUtil.forEachChild(block, token.getType(), child -> {
+                final DetailAST lastNodeInCaseGroup = child.getLastChild();
+                if (isChild(lastNodeInCaseGroup, variable)) {
+                    variableUsageExpressions.add(lastNodeInCaseGroup);
+                }
+            });
+        });
+
+        return variableUsageExpressions;
     }
 
     /**
--- a/.ci/checker-framework-suppressions/checker-framework-nullness-optional-interning-suppressions.xml
+++ b/.ci/checker-framework-suppressions/checker-framework-nullness-optional-interning-suppressions.xml
@@ -2439,13 +2439,6 @@
 
   <checkerFrameworkError unstable="false">
     <fileName>checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java</fileName>
-    <specifier>introduce.eliminate</specifier>
-    <message>It is bad style to create an Optional just to chain methods to get a value.</message>
-    <lineContent>.orElseGet(() -&gt; block.findFirstToken(TokenTypes.SWITCH_RULE));</lineContent>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
-    <fileName>checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java</fileName>
     <specifier>not.interned</specifier>
     <message>attempting to use a non-@Interned comparison operand</message>
     <lineContent>if (curNode == parent) {</lineContent>
--- a/.ci/pitest-suppressions/pitest-coding-1-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-coding-1-suppressions.xml
@@ -179,13 +179,4 @@
     <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getNextSibling with receiver</description>
     <lineContent>DetailAST exprBetweenBrackets = openingBracket.getNextSibling();</lineContent>
   </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>VariableDeclarationUsageDistanceCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck</mutatedClass>
-    <mutatedMethod>lambda$getFirstCaseGroupOrSwitchRule$2</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::findFirstToken with receiver</description>
-    <lineContent>.orElseGet(() -&gt; block.findFirstToken(TokenTypes.SWITCH_RULE));</lineContent>
-  </mutation>
 </suppressedMutations>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
@@ -309,4 +309,15 @@ public class VariableDeclarationUsageDistanceCheckTest extends
         verifyWithInlineConfigParser(getNonCompilablePath(filename), expected);
     }
 
+    @Test
+    public void testVariableDeclarationUsageDistanceSwitchExpressions2() throws Exception {
+        final int maxDistance = 1;
+        final String[] expected = {
+            "16:9: " + getCheckMessage(MSG_KEY, "i", 2, maxDistance),
+        };
+
+        final String filename = "InputVariableDeclarationUsageDistanceCheckSwitchExpressions2.java";
+        verifyWithInlineConfigParser(getNonCompilablePath(filename), expected);
+    }
+
 }
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceCheckSwitchExpressions2.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceCheckSwitchExpressions2.java
@@ -1,0 +1,92 @@
+/*
+VariableDeclarationUsageDistance
+allowedDistance = 1
+ignoreVariablePattern = (default)
+validateBetweenScopes = true
+ignoreFinal = false
+
+
+*/
+
+//non-compiled with javac: Compilable with Java14
+package com.puppycrawl.tools.checkstyle.checks.coding.variabledeclarationusagedistance;
+
+public class InputVariableDeclarationUsageDistanceCheckSwitchExpressions2 {
+    void issue11973() {
+        int i = -1; // violation 'Distance between.*'i'.*and.*first usage is 2, but allowed 1.'
+        int x = -1; // ok
+        switch (i) {
+            case 1 -> {
+                x++;
+                i++;
+            }
+        };
+
+        int a = -1; // ok
+        int a123 = switch (a) {
+            case 1 -> {
+                yield 2;
+            }
+            default -> {
+                yield a;
+            }
+        };
+
+        int b = -1; // ok
+        int b123 = switch (b) {
+            case 1 -> {
+                yield b;
+            }
+            default -> {
+                yield 1;
+            }
+        };
+
+        int day = 1; // ok
+        int c = -1;
+        int c123 = switch (c) {
+            case 1:
+            case 2:
+            default:
+                c++;
+                c--;
+                yield day++;
+        };
+
+        int day1 = 1; // ok
+        int cc = -1;
+        int d123 = switch (cc) {
+            case 1 -> {
+                cc++;
+                cc--;
+                yield day1;
+            }
+            default -> {
+                yield 0;
+            }
+        };
+
+        int day2 = 1; // ok
+        int ccc = -1;
+        int e123 = switch (ccc) {
+            case 1 -> {
+                ccc++;
+                ccc--;
+                yield ccc;
+            }
+            case 2 -> {
+                yield 2;
+            }
+            default -> {
+                ccc++;
+                ccc--;
+                ccc++;
+                day2++;
+                yield 3;
+            }
+        };
+
+        int u = 0;
+        switch (u) { };
+    }
+}


src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceDefault.java
```diff
@@ -1071,4 +1071,10 @@ private void n() {
         New6.field = (int)a;
     }
 
+    class SwitchStatement {
```
nick-mancuso, line 1074: Please do not extend these input files, make a new one with exact config from issue description. We are trying to reduce the size of our input files, let's not add to our tech debt with this update.

nick-mancuso (changes requested):
Please see failures in pitest and generate report.

Item:

stoyanK7:
@nick-mancuso 

While attempting to kill the mutation I discovered that the following code does not raise any violations. Shall I make an issue for it? I believe the newfound bug first needs to be fixed before I can proceed with the current issue.

```java
boolean issue11973(int a) {
    int sw = 5; // violation 'Distance between variable ''sw'' declaration and its first usage is 2, but allowed 1.'
    return switch (a) {
        case 0 -> {
            a++;
            yield sw + a;
        }
    };
}
```

nick-mancuso:
> @nick-mancuso
> 
> While attempting to kill the mutation I discovered that the following code does not raise any violations. Shall I make an issue for it? I believe the newfound bug first needs to be fixed before I can proceed with the current issue.

Please make an issue for this, make sure to explain why we need to fix new issue before we complete this PR.

romani:
@stoyanK7 , NPE  is blocker to use Check at all, can we continue this PR to at least fix NPE and even fine with false positived.

Exception - force users to disable Check at all in config.
false positive - force users to suppress certain file of user code or specific violation.
false negative - does not bother users at all.

So lets fix exception as primary goal.

stoyanK7:
@romani Alright, will focus my attention to fix this NPE :)

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -810,11 +810,15 @@ private static DetailAST getFirstNodeInsideIfBlock(
     private static DetailAST getFirstNodeInsideSwitchBlock(
             DetailAST block, DetailAST variable) {
         final DetailAST currentNode = getFirstCaseGroupOrSwitchRule(block);
+        int currentNodeType = -1;
```
romani, line 813: Pretty weird code.
Do you try to skip `TokenUtil.forEachChild(block, currentNodeType` ?

If yes, please use IF.

stoyanK7: I think I figured it out, working on the tests right now. Just trying different stuff.

stoyanK7: I will tag you when everything is done.

stoyanK7 (commented):
So `pitest-common-1` is going to fail with the following mutation here:
```
Unnecessary suppressed mutation(s) found and should be removed:

Source File: "VariableDeclarationUsageDistanceCheck.java"
Class: "com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck"
Method: "lambda$getFirstCaseGroupOrSwitchRule$2"
Line Contents: ".orElseGet(() -&gt; block.findFirstToken(TokenTypes.SWITCH_RULE));"
Mutator: "org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator"
Description: "replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::findFirstToken with receiver"
```

Can anybody help me what it does since documentation is scarce(practically nonexistent).
This is what I could find
* [NakedReceiverMutator.java](https://github.com/hcoles/pitest/blob/master/pitest/src/main/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/NakedReceiverMutator.java)
* [NakedReceiverMutatorTest.java](https://github.com/hcoles/pitest/blob/master/pitest/src/test/java/org/pitest/mutationtest/engine/gregor/mutators/experimental/NakedReceiverMutatorTest.java)
* [Official docs on Experimental Naked Receiver (EXPERIMENTAL_NAKED_RECEIVER)](https://pitest.org/quickstart/mutators/#EXPERIMENTAL_NAKED_RECEIVER)

rnveach:
> "replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::findFirstToken with receiver"

I believe the mutation is replacing `block.findFirstToken(TokenTypes.SWITCH_RULE)` with `TokenTypes.SWITCH_RULE`.

stoyanK7:
@rnveach Already tried that and `testVariableDeclarationUsageDistanceSwitchExpressions2()` fails when `block.findFirstToken(TokenTypes.SWITCH_RULE)` is changed to just `block`:
Screenshot from 2022-09-11 15-07-04

rnveach:
@stoyanK7 I ran this on my local and the mutation is being killed.
`Killed by com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheckTest.testVariableDeclarationUsageDistanceSwitchExpressions2`.

@Vyom-Yadav ping

rnveach:
@stoyanK7 Please read the top of the pitest report from CI.

It means you are now killing this mutation and can remove it from the suppressions.

stoyanK7:
@rnveach Oh my god, you are right, thanks! I should have actually read whats up. Removing it right now.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceDefault.java
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
stoyanK7: @romani Should be okay now.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceDefault.java
nrmancuso: >partially https://github.com/checkstyle/checkstyle/issues/11163

No. This PR should consist of only:
1. code change in check
2. new unit test
3. new input with config and example from issue report (please expand a bit to include switch expression, etc.)
4. removal of pitest suppression

stoyanK7: Okay, didn't know you are not okay with cleaning up technical debt along with solving issues around. Will keep this in mind for the future. Restored the commit and added more test cases. Done.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral2.java
```diff
@@ -63,4 +63,10 @@ void method() throws Exception {
             a.equals("");
         }
     }
+
+    void issue11973() {
+        int i = -1;
+        switch (i) {
+        }
+    }
```
nrmancuso, line 71: @stoyanK7 please explain why we cannot add this example to new input file.

nrmancuso (changes requested):

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral2.java
stoyanK7: To **a** new input file or to **the** new input file?

nrmancuso: https://github.com/checkstyle/checkstyle/pull/12052/files#diff-70dea49acafeb398ee0ccd4d208360b12c5c169feafa2f3465f77d15874533c8

stoyanK7: Because we already have it here? I mean if you want to repeat it in each test, sure. Done.

stoyanK7: Do you want it moved to the new file or just added as well?

stoyanK7: No particular reason. It doesn't matter at the end of the day, does it? Moved it over to the new input file. Done

nrmancuso: > It doesn't matter at the end of the day, does it?

Please re-read https://github.com/checkstyle/checkstyle/pull/12052#discussion_r967865041 and try to understand viewpoint of reviewer; new test inputs only means that we know that we did not introduce any change in coverage of some condition that we do not immediately recognize. Keep in mind that we run our regression tests over our own code base, so even though your check input is written for one check, we run all checks over all inputs. Yes, your change was minimal, and probably affected nothing, but then we have some grey area. How do we define a minimal change in test input?

It is easier to just follow 1 issue == 1 PR == one (or more) new tests == corresponding number of test inputs. This allows us to focus on the most critical changes (the change in the check itself).

nrmancuso (changes requested):
Please update documentation for `getFirstCaseGroupOrSwitchRule` to mention that we will return null if switch block is empty.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -839,7 +841,8 @@ private static DetailAST getFirstNodeInsideSwitchBlock(
      * SWITCH_RULE ast.
      *
      * @param block the switch block to check.
-     * @return DetailAST of the first CASE_GROUP or SWITCH_RULE.
+     * @return DetailAST of the first CASE_GROUP or SWITCH_RULE,
+     *         otherwise {@code null} if switch block is empty.
```
romani, line 845: remove this line, it from some old your implementation

romani (changes requested):
please review drone CI falure:
>src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceCheckSwitchExpressions2.java:28: error: yield outside of switch expression
                yield 2;
                ^ 


item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -814,13 +814,15 @@ private static DetailAST getFirstNodeInsideSwitchBlock(
...
 
         // Checking variable usage inside all CASE_GROUP and SWITCH_RULE ast's.
-        TokenUtil.forEachChild(block, currentNode.getType(), node -> {
-            final DetailAST lastNodeInCaseGroup =
-                node.getLastChild();
-            if (isChild(lastNodeInCaseGroup, variable)) {
-                variableUsageExpressions.add(lastNodeInCaseGroup);
-            }
-        });
+        if (currentNode != null) {
```
romani, line 817: lets rename currentNode to blockStatement, wedo not iterate over nodes, there is not "current".
https://docs.oracle.com/javase/specs/jls/se7/html/jls-14.html#jls-14.11

romani: To avoid such nulls, it might be reasonable to make method 
```
getVariableExpressions(block, variable){
Optional.ofNullable(block.findFirstToken(TokenTypes.CASE_GROUP))
            .orElse(() -> block.findFirstToken(TokenTypes.SWITCH_RULE))
. map(
TokenUtil.forEachChild(block, currentNode.getType(), node -> {
                final DetailAST lastNodeInCaseGroup =
                    node.getLastChild();
                if (isChild(lastNodeInCaseGroup, variable)) {
                    variableUsageExpressions.add(lastNodeInCaseGroup);
                }
            })
).
orElseGet( List.empty);
}
```
and combine bodies.


of make it
```
TokenUtil.forEachChild(block, List.of(TokenTypes.CASE_GROUP, TokenTypes.SWITCH_RULE), .......
```

romani (changes requested):
items

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
stoyanK7: Done, created method `getVariableUsageExpressionsInsideSwitchBlock()` so we get rid of null check,

romani: Why not TokenUtil.forEachChild(block, List.of(TokenTypes.CASE_GROUP, TokenTypes.SWITCH_RULE), ....... ?

It more clear

romani: And it is what we actually need to

stoyanK7: TokenUtil.forEachChild takes in an int as a second parameter, not a list.

stoyanK7: Of course, there is the workaround to make two calls to `TokenUtil.forEachChild()` if that is what you wanna do. One for each token `CASE_GROUP` and `SWITCH_RULE`.

romani: All this code want to do is to iterate over each nested block of switch, what is know to be "case" or switchRule. And it probably missed `default` that is not good, but might different issue.
So we just need to iterate over all nodes that are case or switchRule.
So `TokenUtil.forEachChild(block, List.of(TokenTypes.CASE_GROUP, TokenTypes.SWITCH_RULE)` will do exactly what we need.

The only obstacles I can imagine is pitest, which might complain that there is no test for switch with case and switchRule in same time. If this is a case, ok to stay on `.map().or()`

stoyanK7: Done. Method now looking for both `CASE_GROUP` and `SWITCH_RULE` in the same switch block. Even though this is not possible in the Java language to have `CASE_GROUP` and `SWITCH_RULE` in the same switch expression, I think you just want to be on the safe side, am I correct?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -835,15 +825,27 @@ private static DetailAST getFirstNodeInsideSwitchBlock(
...
      */
-    private static DetailAST getFirstCaseGroupOrSwitchRule(DetailAST block) {
-        return Optional.ofNullable(block.findFirstToken(TokenTypes.CASE_GROUP))
-            .orElseGet(() -> block.findFirstToken(TokenTypes.SWITCH_RULE));
+    private static List<DetailAST> getVariableUsageExpressionsInsideSwitchBlock(DetailAST block,
+                                                                            DetailAST variable) {
+        final List<DetailAST> variableUsageExpressions = new ArrayList<>();
+
+        for (int type : List.of(TokenTypes.CASE_GROUP, TokenTypes.SWITCH_RULE)) {
+            TokenUtil.forEachChild(block, type, child -> {
```
romani, line 840: Please share a problem to write util method that takes list of tokens

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
romani: TokenUtil.forEachChild(block, List.of(TokenTypes.CASE_GROUP, TokenTypes.SWITCH_RULE))

stoyanK7: Not a problem at all. I guess I did not pick up your thoughts/what you meant. On it.

stoyanK7: Done. Added method and a test for it.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -321,6 +321,37 @@ public void testForEachChild() {
             .isEqualTo(secondSibling);
     }
 
+    @Test
+    public void testForEachChildMultipleTypes() {
```
romani, line 325: We do not allow pure Unit tests. They fake quality and prime to produce artificial and not real main code 

Can we cover this by Input files.

If not possible I would rather use previous map().or()

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
nrmancuso: @stoyanK7 make sure that this check's test is in the target tests for pitest-utils profile in pom.xml.

stoyanK7: @nrmancuso I think it is there already. All `com.puppycrawl.tools.checkstyle.utils.*` classes are targeted and I see no exclusions for `TokenUtil`.

nrmancuso: Please re-read https://github.com/checkstyle/checkstyle/pull/12052#discussion_r971944458 carefully. Please confirm if this check's test (VariableDecl...CheckTest) is in the `targetTests` for the pitest-utils profile.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -321,6 +330,31 @@ public void testForEachChild() {
...
     }
 
+    @Test
+    public void testForEachChildMultipleTypes() throws Exception {
+        final DetailAST astForTest = getNodeFromFile(TokenTypes.OBJBLOCK,
+                "InputTokenUtilTestForEachChildMultipleTypes.java");
+        final List<String> children = new ArrayList<>(3);
+        final List<Integer> tokensToMatch = List.of(TokenTypes.CLASS_DEF, TokenTypes.VARIABLE_DEF);
+
+        TokenUtil.forEachChild(astForTest, tokensToMatch, child -> {
```
romani, line 340: I might was no clear. I know it is very unusual and sattle.
We allow tests only by Inputs with embedded config.
All other forms of testing are well proven "way to hell" in our repo .

Please no updates in TokenUtilTest.

Let's come back to "map.or" solution , sorry that we confused you and lost valuable time.

Let me know if you interested in more details why we follow unusual philosophy

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
stoyanK7: I think I get what you mean now. We explicitly say which tests pitest should use to kill the mutation. In this case VariableDeclarationDistance uses it, so it should also be mentioned in pom.xml. Is that correct?

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
stoyanK7: > Inputs with embedded config.

This is not possible here since it is not a check/filter itself.

---
> Please no updates in TokenUtilTest
> Let's come back to "map.or" solution 

Reverted back. Done.

---
It's alright. I would not call it wasted time. Dug around the codebase and learned a few things about the rules of this repository and how pull requests should be.


> All other forms of testing are well proven "way to hell" in our repo .
>  Let me know if you interested in more details why we follow unusual philosophy

Yes, I'd like to know why. In my eyes, such a test does not seem harder to maintain than any other check's tests and input files.

romani (approved):
Ok to merge

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
romani: To cover until method we use actual tests of Check.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
romani: We try our best to use tests only like `verifyWithInlineConfigParser(getNonCompilablePath(filename)`.

It makes us confident that we always use compilable input files. And all code in Check can handle any possible-and-compilable combinations of java code and all code in Check are there for a reason (pitest) no extra code and no dead code. Removal/mutation of any part of it immediately detected by Junit tests.

As you might notice in above statement we do not mention about util methods at all , as potentially until method can be copied to class and be part of it. It is our details of implementation that we use util methods, we do not want to let anyone know it, and it subject to change at any time.

If we allow testing of utility methods separately, all logic of all Checks can be generalized and moved to utilities so whole test pressure will also move to utilities. And we will struggle with auto measure of quality of Check. 
In result maintainers will win in short term, user will loose in long term, usage will drop,  maintainers will not found way to invite new contributions, project is dead eventually.

Right now as you might noticed, new contributor does not need to know all ast and all utilities and all rules. All you need is to put a java file to test and place debug in Check visitToken.

romani: We did not born smart, it took us 20 years to do all types of testing to end up where bee are. So are of code is still old style

stoyanK7:
Github, generate report

romani:
OMG, we had exception on project that we in regression list. We might need to have CI script that runs updated in commit check against all such projects to expect no exceptions.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -835,15 +826,29 @@ private static DetailAST getFirstNodeInsideSwitchBlock(
...
+     * @return List of usages or empty list if none are found.
      */
-    private static DetailAST getFirstCaseGroupOrSwitchRule(DetailAST block) {
-        return Optional.ofNullable(block.findFirstToken(TokenTypes.CASE_GROUP))
-            .orElseGet(() -> block.findFirstToken(TokenTypes.SWITCH_RULE));
+    private static List<DetailAST> getVariableUsageExpressionsInsideSwitchBlock(DetailAST block,
+                                                                            DetailAST variable) {
+        return Optional
+                .ofNullable(block.findFirstToken(TokenTypes.CASE_GROUP))
+                .or(() -> Optional.ofNullable(block.findFirstToken(TokenTypes.SWITCH_RULE)))
```
nrmancuso, line 840: Nested optional... this is not good. This return statement is complex and hard to read/ understand, and will be difficult to debug in the future.

Consider that we have three cases of interest to this check:
1. switch block has a CASE_GROUP child
2. switch block has a SWITCH_RULE child
3. switch block has no children that we care about

Then please do something like:

```java
        final Optional<DetailAST> b = TokenUtil.findFirstTokenByPredicate(b,
                c -> c.getType() == TokenTypes.SWITCH_RULE || c.getType() == TokenTypes.CASE_GROUP
        );
```

Now, we do not deal with null, and we have an optional that clearly represents scenario above.

Then:
1. proceed with list creation, we just return an empty list if we have no usages to register
2. if optional block child is present, register variable usages

suggestion:
```
        final List<DetailAST> variableUsageExpressions = new ArrayList<>();

        b.ifPresent(blockStatement -> {
            TokenUtil.forEachChild(block, blockStatement.getType(), child -> {
                final DetailAST lastNodeInCaseGroup = child.getLastChild();
                if (isChild(lastNodeInCaseGroup, variable)) {
                    variableUsageExpressions.add(lastNodeInCaseGroup);
                }
            });
        });

        return variableUsageExpressions;

```
@romani I am open to suggestions or discussion on this, but I do not think current implementation is easy to read/ reason about/ maintain

@stoyanK7 please read more about Optional type at https://stuartmarks.wordpress.com/2016/09/27/vjug24-session-on-optional/ if you are interested, I found this article/presentation last week and found it valuable.

nrmancuso (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
romani: I am ok with proposal.
I am not sure if it is much better, but I. Not firm on this point.

rnveach:
> we had exception on project that we in regression list. 

Regression config is missing turning on `validateBetweenScopes` as seen in regression report
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/4dab5af_2022060049/reports/diff/openjdk17/index.html
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/checks-nonjavadoc-error.xml#L349

Also want to remind we are missing https://github.com/checkstyle/contribution/pull/548 from regression config as well

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
romani: @stoyanK7, please do update.

stoyanK7: Done.

Took the time to watch the 50 minute presentation. I have mixed feelings about this now, at the same time they say single `Optional` is okay but then not every null should use `Optional`/do not use it for simple null check. It is really dependent on situation.

As for me, both are okay here. No complaints.

romani:
@Vyom-Yadav , please glance at failure
Is it `capture#xxx` problem ?
https://github.com/checkstyle/checkstyle/issues/12210

Vyom-Yadav:
@romani For nullness checker it is `capture#`, for `methods-resource-fenum` it is
`Message: "@MustCall method close may not have been invoked on temp-var-20910 or any of its aliases."` 
`temp-var-20910`, this will also vary with runs.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
stoyanK7: I get it now - all in the name of maintainability and keeping the project going and easy to contribute to. Which I have noticed btw. 

What does that mean for the util tests though? Do you want them to move over to such testing `verifyWithInlineConfigParser`? Is it even possible? Or do you expect tests from checks that use util methods to cover them?

romani:
@nrmancuso , lets merge this without Checker green state, in worse case we will add one suppression to our files.
While we are learning this Checker tool, we ok to grow tech debt

nrmancuso (approved):
Checker is already unstable, I am fine to merge on red.

PR #12039, opened 2022-08-06
Added Russian messages.

ZuevKirill95:
doc: Added Russian messages

Issue #7436, opened 2020-01-02
Avoid using Locale.setDefault in tests


pbludov:
Part of migration to Junit5 process https://github.com/checkstyle/checkstyle/issues/6916
Blocker for parallel execution of all unit tests.

By enabling parallel execution of unit tests, they start to fail in random places with messages like this:
```
[ERROR]   AnnotationLocationCheckTest.testAnnotation:204->AbstractModuleTestSupport.verify:220->AbstractModuleTestSupport.verify:256->AbstractModuleTestSupport.verify:292 error message 0 expected:<...Annotation.java:17: [Annotation 'AnnotationAnnotation' have incorrect indentation level 2, expected level should be 0].> but was:<...Annotation.java:17: [Die Annotation 'AnnotationAnnotation' hat eine unerwartete Einrückungstiefe von 2 (erwartet: 0)].>
```
This happens when a test does  `Locale.setDefault(...)`, which affects all other tests running at the same time. To avoid such failure, all unit tests shouldn't alter the global system state.
Instead, the Junit 5 extension [DefaultLocale](https://junit-pioneer.org/docs/default-locale-timezone/) should be used to trick `LocalizedMessage` to load locale-specific message.

rnveach:
We should configure a new `ForbidCertainMethodCheck` from sevntu to forbid the method.

romani:
@pbludov , do you think we will succeed to make MT execution on tests if main code is not ready for MT ?

pbludov:
> What about when we do modifications to other classes?
> do you think we will succeed to make MT execution on tests if main code is not ready for MT ?

I'm actually trying to figure out how ready it is. Definitely, all tests that change static fields or global state are incompatible with parallel execution. For example, `LocalizedMessage` has static field [sLocale](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/LocalizedMessage.java#L64). And modification of this field in a test affects all check tests running concurrently the same way as `Locale.setDefault(...)` does.
But we only need to set this field when initializing. It is necessary that the initialization of this class does not overlap with its use.

The Junit5 documentation recommends using `ResourceLock` annotation to solve this problem.
```java
@ResourceLock(value = "global state", mode = ResourceAccessMode.READ_WRITE)
public class LocalizedMessageTest {
  // ... It is safe to modify sLocale here
}
```
For all check tests:
```java
[...]

Issue #4818, opened 2017-07-25
Look into possibly adding random junit ordering


rnveach:
http://randomjunit.sourceforge.net - Last Published: 2011-07-23 | Version: 1.0.3

This application helps to execute jUnit tests in a random order.
Stack overflow: https://stackoverflow.com/questions/1444314/how-can-i-make-my-junit-tests-run-in-random-order

junit on execution order: https://github.com/junit-team/junit4/wiki/test-execution-order
> JUnit does not specify the execution order of test method invocations ... methods were simply invoked in the order returned by the reflection API
> a predictable failure is better than a random failure

PR #12053, opened 2022-08-11
infra: set English locale for tests

ZuevKirill95:
related to https://github.com/checkstyle/checkstyle/pull/11529

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/AbstractPathTestSupport.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/AbstractPathTestSupport.java
@@ -23,8 +23,11 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.util.Locale;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+
+import org.junit.jupiter.api.BeforeEach;
 
 public abstract class AbstractPathTestSupport {
 
@@ -41,6 +44,15 @@ public abstract class AbstractPathTestSupport {
      * @return path for the package name for the file.
      */
     protected abstract String getPackageLocation();
+
+    /**
+     * Sets the English locale for all tests.
+     * Otherwise, some tests failed in other locales.
+     */
+    @BeforeEach
+    public void setEnglishLocale() {
+        Locale.setDefault(Locale.ENGLISH);
+    }
 
     /**
      * Retrieves the name of the folder location for resources.
--- a/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
@@ -1822,7 +1822,7 @@ public class CheckerTest extends AbstractModuleTestSupport {
             }
             final int childCount = ast.getChildCount();
             if (childCount != METHOD_DEF_CHILD_COUNT) {
-                final String msg = String.format(Locale.getDefault(),
+                final String msg = String.format(Locale.ENGLISH,
                     "AST node in no comment tree has wrong number of children. "
                             + "Expected is %d but was %d",
                     METHOD_DEF_CHILD_COUNT, childCount);
@@ -1836,7 +1836,7 @@ public class CheckerTest extends AbstractModuleTestSupport {
             }
             final int cacheChildCount = ast.getFirstChild().getChildCount();
             if (cacheChildCount != actualChildCount) {
-                final String msg = String.format(Locale.getDefault(),
+                final String msg = String.format(Locale.ENGLISH,
                         "AST node with no comment has wrong number of children. "
                                 + "Expected is %d but was %d",
                         cacheChildCount, actualChildCount);
@@ -1871,6 +1871,7 @@ public class CheckerTest extends AbstractModuleTestSupport {
             return new int[] {TokenTypes.METHOD_DEF};
         }
 
+        // Locale.ENGLISH until #12104
         @Override
         public void visitToken(DetailAST ast) {
             if (ast.findFirstToken(TokenTypes.MODIFIERS).findFirstToken(
@@ -1879,7 +1880,7 @@ public class CheckerTest extends AbstractModuleTestSupport {
             }
             final int childCount = ast.getChildCount();
             if (childCount != METHOD_DEF_CHILD_COUNT) {
-                final String msg = String.format(Locale.getDefault(),
+                final String msg = String.format(Locale.ENGLISH,
                     "AST node in comment tree has wrong number of children. "
                             + "Expected is %d but was %d",
                     METHOD_DEF_CHILD_COUNT, childCount);
@@ -1893,7 +1894,7 @@ public class CheckerTest extends AbstractModuleTestSupport {
             }
             final int cacheChildCount = ast.getFirstChild().getChildCount();
             if (cacheChildCount != actualChildCount) {
-                final String msg = String.format(Locale.getDefault(),
+                final String msg = String.format(Locale.ENGLISH,
                         "AST node with comment has wrong number of children. "
                                 + "Expected is %d but was %d",
                         cacheChildCount, actualChildCount);
--- a/src/test/java/com/puppycrawl/tools/checkstyle/DefaultLoggerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/DefaultLoggerTest.java
@@ -44,7 +44,7 @@ import com.puppycrawl.tools.checkstyle.internal.utils.TestUtil;
 
 public class DefaultLoggerTest {
 
-    private static final Locale DEFAULT_LOCALE = Locale.getDefault();
+    private static final Locale DEFAULT_LOCALE = Locale.ENGLISH;
 
     @AfterEach
     public void tearDown() {
@@ -280,11 +280,12 @@ public class DefaultLoggerTest {
     @Test
     public void testNewCtor() throws Exception {
         final ResourceBundle bundle = ResourceBundle.getBundle(
-                Definitions.CHECKSTYLE_BUNDLE, Locale.ROOT);
+                Definitions.CHECKSTYLE_BUNDLE, Locale.ENGLISH);
         final String auditStartedMessage = bundle.getString(DefaultLogger.AUDIT_STARTED_MESSAGE);
         final String auditFinishedMessage = bundle.getString(DefaultLogger.AUDIT_FINISHED_MESSAGE);
         final String addExceptionMessage = new MessageFormat(bundle.getString(
-                DefaultLogger.ADD_EXCEPTION_MESSAGE), Locale.ROOT).format(new String[] {"myfile"});
+                DefaultLogger.ADD_EXCEPTION_MESSAGE), Locale.ENGLISH).format(new String[] {"myfile"}
+        );
         final String infoOutput;
         final String errorOutput;
         try (MockByteArrayOutputStream infoStream = new MockByteArrayOutputStream()) {
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -135,47 +135,47 @@ test)
 
 test-de)
   mvn -e --no-transfer-progress clean integration-test failsafe:verify \
-    -DargLine='-Duser.language=de -Duser.country=DE -Xms1024m -Xmx2048m'
+    -Dsurefire.options='-Duser.language=de -Duser.country=DE -Xms1024m -Xmx2048m'
   ;;
 
 test-es)
   mvn -e --no-transfer-progress clean integration-test failsafe:verify \
-    -DargLine='-Duser.language=es -Duser.country=ES -Xms1024m -Xmx2048m'
+    -Dsurefire.options='-Duser.language=es -Duser.country=ES -Xms1024m -Xmx2048m'
   ;;
 
 test-fi)
   mvn -e --no-transfer-progress clean integration-test failsafe:verify \
-    -DargLine='-Duser.language=fi -Duser.country=FI -Xms1024m -Xmx2048m'
+    -Dsurefire.options='-Duser.language=fi -Duser.country=FI -Xms1024m -Xmx2048m'
   ;;
 
 test-fr)
   mvn -e --no-transfer-progress clean integration-test failsafe:verify \
-    -DargLine='-Duser.language=fr -Duser.country=FR -Xms1024m -Xmx2048m'
+    -Dsurefire.options='-Duser.language=fr -Duser.country=FR -Xms1024m -Xmx2048m'
   ;;
 
 test-zh)
   mvn -e --no-transfer-progress clean integration-test failsafe:verify \
-    -DargLine='-Duser.language=zh -Duser.country=CN -Xms1024m -Xmx2048m'
+    -Dsurefire.options='-Duser.language=zh -Duser.country=CN -Xms1024m -Xmx2048m'
   ;;
 
 test-ja)
   mvn -e --no-transfer-progress clean integration-test failsafe:verify \
-    -DargLine='-Duser.language=ja -Duser.country=JP -Xms1024m -Xmx2048m'
+    -Dsurefire.options='-Duser.language=ja -Duser.country=JP -Xms1024m -Xmx2048m'
   ;;
 
 test-pt)
   mvn -e --no-transfer-progress clean integration-test failsafe:verify \
-    -DargLine='-Duser.language=pt -Duser.country=PT -Xms1024m -Xmx2048m'
+    -Dsurefire.options='-Duser.language=pt -Duser.country=PT -Xms1024m -Xmx2048m'
   ;;
 
 test-tr)
   mvn -e --no-transfer-progress clean integration-test failsafe:verify \
-    -DargLine='-Duser.language=tr -Duser.country=TR -Xms1024m -Xmx2048m'
+    -Dsurefire.options='-Duser.language=tr -Duser.country=TR -Xms1024m -Xmx2048m'
   ;;
 
 test-ru)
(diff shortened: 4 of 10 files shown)


ZuevKirill95:
Error:
https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=10743&view=logs&j=69ff11bc-5a77-53a4-545f-8dfc2c866a0e&t=fe4ed25a-cc1d-5fd5-85b9-28f4310f8666&l=20

```
HEAD is now at 12c208331 infra: run checkstyle in ru locale
The PR is based on a master that is 25 commit(s) old.
The max allowed is 10.
This PR is too old and should be rebased on origin/master.
```

I suppose it because commit infra: run checkstyle in ru locale too old
Maybe remove this commit? Same changes in https://github.com/checkstyle/checkstyle/pull/12039

rnveach:
@ZuevKirill95 You have to rebase on the latest master and remove all merge commits.

> Maybe remove this commit? Same changes in https://github.com/checkstyle/checkstyle/pull/12039

I am not understanding why we have the same commit in 2 PRs. If one is just for testing then please point it out, but before removing any commits this PR needs confirmation the changes will work before merging.

ZuevKirill95:
> I am not understanding why we have the same commit in 2 PRs.

I'm not sure why
@romani asked me for this in https://github.com/checkstyle/checkstyle/pull/11529#issuecomment-1211483451
Maybe I misunderstood

rnveach:
I thought my understanding about 11529 was that it was to be closed in favor of this PR as we were using that as a test area to identify the issue.

He event said in the comment:
> As you open your PR, I will close my PR.

So yes, 11529 is going away.

romani:
We need to preserve changes at https://github.com/checkstyle/checkstyle/pull/11529/commits they show correct way to proporate locale, please squash them in one commit 

Second commit should be from Kirill

ZuevKirill95:
Help with this error, please
image

ZuevKirill95:
I added "Dsurefire" to witelist

ZuevKirill95:
It seems this error is not related to my changes

image

romani:
@Vyom-Yadav , @nick-mancuso  is this known instability, can we register it in suppression xml  ? to not bother us unexpectedly

romani:
I placed one more commit with Albanian language - https://www.oracle.com/java/technologies/javase/jdk8-jre8-suported-locales.html
we unlikely to get translation to this language :).

romani (approved):
ok to merge

nick-mancuso:
> @Vyom-Yadav , @nick-mancuso is this known instability, can we register it in suppression xml ? to not bother us unexpectedly

I have not seen this particular failure previously, I have noticed a couple of pitest-ant failures and one pitest-javadoc failure lately: https://github.com/checkstyle/checkstyle/actions/workflows/pitest.yml?query=branch%3Amaster+is%3Afailure

Failures have been random enough that I do not think we need to consider "unstable" option yet, I do not want to overuse/ abuse this attribute.

romani:
Does it make sense to collect all instabilities? to notice that they likely to happen, and may be not that often.
Do you have already some issue for this ?

nick-mancuso:
> Does it make sense to collect all instabilities? to notice that they likely to happen, and may be not that often. Do you have already some issue for this ?

We have one for pitest-coding only, but we are preparing to close this. I will create new general tracker.

romani:
```
Pitest-api


Source File: "Violation.java"
Class: "com.puppycrawl.tools.checkstyle.api.Violation$Utf8Control"
Method: "newBundle"
Line Contents: "final String resourceName = toResourceName(bundleName, &quot;properties&quot;);"
Mutator: "org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator"
Description: "replaced call to com/puppycrawl/tools/checkstyle/api/Violation$Utf8Control::toResourceName with argument"
Line Number: 542
```

Restarted

nick-mancuso:
https://github.com/checkstyle/checkstyle/issues/12072

romani:
Strange , stable failure.
```
Source File: "Violation.java"
Class: "com.puppycrawl.tools.checkstyle.api.Violation$Utf8Control"
Method: "newBundle"
Line Contents: "final String resourceName = toResourceName(bundleName, &quot;properties&quot;);"
Mutator: "org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator"
Description: "replaced call to com/puppycrawl/tools/checkstyle/api/Violation$Utf8Control::toResourceName with argument"
Line Number: 542
```

This looks like related to code changes of this PR.
@ZuevKirill95 , can you look what we can do do kill survival?
Might be extra test is required with localized execution.

ZuevKirill95:
I'm new to pit tests

I don't understand how to reproduce this on the local computer
I tried to change this code 

```java
 final String resourceName = toResourceName(bundleName, "properties");
```

to

```java
 final String resourceName = bundleName;
```
and to
```java
 final String resourceName = "properties";
```

but the tests failed. Does this mean that the mutation should not have survived?

Vyom-Yadav:
@ZuevKirill95 Strange, you have got the mutation correct, I cross-checked it, and the actual mutation is `final String resourceName = "properties";` but the test suite is failing with that, however, pitest shows it as survival.

romani:
@Vyom-Yadav , Should we add this survival to suppression file ?

Vyom-Yadav:
@ZuevKirill95 Strangely using `@BeforeEach` instead of `@BeforeAll` kills the mutation, the issue for the same has been opened at https://github.com/hcoles/pitest/issues/1081.

For now, you can use `@BeforeEach` to kill the mutation.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractPathTestSupport.java
```diff
@@ -42,6 +45,15 @@ public abstract class AbstractPathTestSupport {
      */
     protected abstract String getPackageLocation();
 
+    /**
+     * Sets the English locale for all tests.
+     * Otherwise, some tests failed in other locales.
+     */
+    @BeforeAll
+    public static void setEnglishLocale() {
+        Locale.setDefault(Locale.ENGLISH);
```
rnveach, line 54: This is turning into an ugly issue. `ViolationTest` has a static final Locale which gets its value before this method is run. So now we can have a test global variable with its own Locale and then this Before method comes in and set it to its own specific value and possibly conflict with each other. If they both have the same value, there is no issue, but if they are different then this could become messy.

@romani @nick-mancuso In my own project, I had to create a rule that "static final fields" for tests must be a single state throughout the entire life of the tests and can never change. I really found this to be helpful when I introduced random test ordering (https://github.com/checkstyle/checkstyle/issues/4818) as I found bad tests that were modifying values that should have been constant throughout the entire test run and was causing unpredictable issues.  I also forbid mutable fields for tests so 1 test can be completely isolated, but I am willing for this to be ignored for now.

I think we need to incorporate this rule and reconsider these fields, probably starting with this PR.

rnveach (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/AbstractPathTestSupport.java
romani: https://github.com/checkstyle/checkstyle/blob/fe3969a2f16854a93d66e80e0ac2df3aa5a3102c/src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java#L55

Is used only 3 times , we can move definition to each test , to let test control locale .

Goal of this PR to let people with non EN locale to run tests/build without problems. Benefits of this is singnificatly bigger than potential problem that we can resolve later on by overrides and any other way.

romani: All tests which want to play with locale can have separate base class , and do what makes sense to them.

rnveach: > we can move definition to each test 

It will most likely be more involved than this. We are setting the default Locale probably after all our production classes have loaded. Some like Violation, use the preset locale which is loaded before the before test is run. This is starting to look like all tests need to extend a base and we need a basic init routine.

ZuevKirill95: My suggetion add `@DefaultLocale("en")` in each test

This is convenient for those who have a different locale, if you need to locally check that the message is correct in your language, just change locale in test

rnveach: This won't properly handle our production code which is what I was referring to in my previous post.
Example: https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/Violation.java#L68

Just for noting, `@DefaultLocale` saves the current default locale before the test is run, switches the locale to the specified in the annotation, and the restores the previous default locale after the test is run.

rnveach: IMO, I don't think we should do this BeforeEach and setting the locale in it. We will have to setup and maintain more than just the locale for our entire suite to be in that language.

The whole reason for setting the locale to english is because some tests require english language to validate properly.

My suggestion is we modify tests to skip validation when English is required and language in use is not English. The environment is already in place that the language coming in is used everywhere. We fixed the issue with our CI to run tests in any locale. Lets just skip tests that test English specific results.

rnveach: Another reason to not use `Locale.setDefault`. https://github.com/checkstyle/checkstyle/issues/7436

rnveach:
I am marking this as blocked since I don't think its safe to merge in `Locale.setDefault(Locale.ENGLISH);`. We should be thinking about something like #12104 to resolve the underlying issue.

If the rest of the changes can be separated away and won't cause harm, then I am good to merge them in their own separate PR.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractPathTestSupport.java
romani: > Lets just skip tests that test English specific results.

as we use english messages in Input files, we are forced to run in EN locale.

rnveach:
@romani I see it as pointless to continue this PR. My prior denial and block message has not changed. Even the fixes I see you doing now don't fix the immutable issue that I mentioned in my post and that is why it continues to fail.

romani:
locale based test are [failing](https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=17975&view=logs&j=f07bbe47-a53d-57aa-396b-b864e923835d&t=067af9d2-0932-5b80-cadd-10bf65a9f15a&l=1321):
```
[ERROR] Failures: 
[ERROR]   LocalizedMessageTest.testNullArgs:60 Violation should contain exception info
value of           : getMessage()
expected to contain: Error auditing myfile
but was            : Fehler beim Prüfen von myfile
```

romani:
@rnveach, I got a stable form of test executions.

this is good step forward before #12104 because:
- resolves problems for non english contributors (my laptop), it might not be very stable but a way better then stable failure of `mvn test`
- fix our CI test config to finally execute test in locale
- all hacks are located in one commit so it in #12104 we simply revert it as first commit and do what is required to do

and I am switching to brain storm to #12104 that might take a bit to agree by all as it is api changes.

rnveach:
Code has not changed, so my review hasn't changed and this should still be blocked. Even admitting we will revert most of this PR shows this isn't a good step forward. I am against adding more hacks and promising fixes later when the later never comes.

romani:
Right now , and it always was like this, we can pass `mvn test` only on English locale.
We do not change anything. We simply making it work. Nothing new, all playing with locale is not working fake, and now we finally make it clear.

Extra commit, makes life on bigger part of contributors easy and it doesn't not change anything for you.

romani:
@nrmancuso , we are in disagreement, please share second opinion on this PR.
My position is to ease problem for all contributors while we dealing with big gap in our design for locales that existed forever.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractPathTestSupport.java
nrmancuso: @rnveach 

>Before method comes in and set it to its own specific value and possibly conflict with each other. If they both have the same value, there is no issue, but if they are different then this could become messy.

Are we able to make assertions about this case (different locale)?

>My suggestion is we modify tests to skip validation when English is required and language in use is not English.

can you expand on this?

nrmancuso:
Honestly, I am pretty confused about what's going on here. We are referencing closed/blocked/discussion ongoing issue. Why do we need the change in command line args? Can this be merged separately?

src/test/java/com/puppycrawl/tools/checkstyle/AbstractPathTestSupport.java
rnveach: > Are we able to make assertions about this case (different locale)?

If you are talking about code assertions, then yea they can always be added, but it doesn't prevent the issue. And then another issue is what if new code comes in. We won't remember to add more of these checks and the issue could become worse, that is why this isn't a good change.

> My suggestion is we modify tests to skip validation when English is required and language in use is not English.
> can you expand on this?

From what I remember of the issue, basically when a non-English user runs our tests there are failures related to language. Some test(s) is expecting English, and the result is actually the user's non-English locale.

I believe my suggestion was to drop all the code in this PR, and skip over the current test failures ONLY if the user's locale isn't English and validate as many other things as we can. This wouldn't skip anything in CI, as CI doesn't fail on this issue currently.
Basically:
````
if (language == English) {
  validate(....);
}
````

My other suggestion is to actually fix the issue, which is what my other PRs (https://github.com/checkstyle/checkstyle/pull/12511 and https://github.com/checkstyle/checkstyle/pull/12560) were doing. In those PRs, since Checker controls the locale currently (https://checkstyle.org/config.html#Checker_Properties), it would pass it to all the other modules for when a locale is needed and we would drop the global static which is the center of the issue. Basically switching a global static, to a stack based.

**Review by nrmancuso — CHANGES_REQUESTED** [review]
I am ok with these hacks to stabilize test execution for non-english users. This issue doesn't personally affect me, so it is hard for me to weigh risks vs benefits. Ease of contribution workflow provides value to users, contributors, and maintainers alike.

File: src/test/java/com/puppycrawl/tools/checkstyle/AbstractPathTestSupport.java [hunk 2]
```diff
@@ -42,6 +45,15 @@ public abstract class AbstractPathTestSupport {
      */
     protected abstract String getPackageLocation();
 
+    /**
+     * Sets the English locale for all tests.
+     * Otherwise, some tests failed in other locales.
+     */
+    @BeforeEach
+    public void setEnglishLocale() {
+        Locale.setDefault(Locale.ENGLISH);
```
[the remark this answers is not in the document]
romani: image

it is execution before all test methods of all tests.
We can probably can do some other elegant way, but as we consider this as temporal update, let keep it.
I do not know how to make it better.

src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
```diff
@@ -1836,7 +1836,7 @@ public void visitToken(DetailAST ast) {
             }
             final int cacheChildCount = ast.getFirstChild().getChildCount();
             if (cacheChildCount != actualChildCount) {
-                final String msg = String.format(Locale.getDefault(),
+                final String msg = String.format(Locale.ENGLISH,
```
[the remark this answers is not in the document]
romani: done, but the more I think on this, the more I become confident that all test should be english, only rare test methods that does locale specific testing can be different.

src/test/java/com/puppycrawl/tools/checkstyle/DefaultLoggerTest.java
```diff
@@ -277,6 +278,7 @@ public void testCountryIsValid() {
                 .contains(country);
     }
 
+    @DefaultLocale("en")
```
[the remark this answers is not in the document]
romani: done

romani:
> Also, we need agreement to also take an iterative approach to this issue in general, and move forward with https://github.com/checkstyle/checkstyle/pull/12511 even if it is not perfect.

I already posted comment to it, waiting for reply.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractPathTestSupport.java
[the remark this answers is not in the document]
nrmancuso: Ok, yes, I see now that `AbstractModuleTestSupport` extends this class, I am good.

src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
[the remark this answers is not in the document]
nrmancuso: >done, but the more I think on this, the more I become confident that all test should be english, only rare test methods that does locale specific testing can be different.

Agreed

nrmancuso (approved):
I am good, update is not perfect but at least it is a step forward. @romani please make CI happy

romani:
@rnveach , please consider PR one more time.

rnveach:
My point of view that this PR isn't a good change has not changed.

romani:
This very rare case, but I need to use super power to override disagreement, to move energy to fix in better way at next PR, https://github.com/checkstyle/checkstyle/pull/12511 has to be changed a bit to avoid API changes.

@nrmancuso , please comment on that PR too with your thoughts.

PR #12086, opened 2022-08-22
Issue #3770: Extend OverloadMethodsDeclarationOrderCheck to Support Multiple Groupings

### LINKED ISSUE #3770: OverloadMethodsDeclarationOrder: new property to allow few groups of overloaded methods
This is a summary of #3698 .
Successor for: #2652, #3235 .

https://checkstyle.org/config_coding.html#OverloadMethodsDeclarationOrder

property "ModifiersThatCreateGroup" = list of RegExp for modifiers.
It is ok to provide better name for property. 

Examples : 
`ModifiersThatCreateGroup` = "static" - all static methods should be separate (should cover request of this issue), 
"ModifiersThatCreateGroup" = "public, private, protected, package" - 4 groups, each visibility modifier in its own. There will be 'package' artificial visibility modifier.
"ModifiersThatCreateGroup" = "static final , public native, private static, protected abstract" - 4 groups it will keep method with "static final" , "public native", "private static", "protected abstract" is separate groups, modifiers are completely matching , order is matter

Examples:
```
// ModifiersThatCreateGroup = "static"
public class Test {
    void foo() {}
    void foo(String baz) {}
    void bar() {} // OK, bar group
    static void foo() {} // OK, as "static foo" is separate group
}

// ModifiersThatCreateGroup = "static"
public class Test {
    static private void foo() {  }
    void foo() {  } // OK
    void foo(String baz) {  } //OK
    void bar() {  }
    static void foo() {  } // WARN - "static foo" methods are not groupped
}


// ModifiersThatCreateGroup = "static"
public class Test {
    public void foo() {  } //OK
    public void bar(String baz) {  } //OK
    private void foo(String baz) {  } //VIOLATION private foo(String) should be next to foo()
    void bar() {  } // VIOLATION bar should be next to bar(String)
    static public void foo() {  } // OK, "static foo" is separate group.
}

// ModifiersThatCreateGroup = "static"
public class Test {
    public void foo() {  } //OK
    private void bar(String baz) {  } //OK
    private void foo(String baz) {  } //VIOLATION private foo(String) should be next to foo()
    void bar() {  } // VIOLATION bar should be next to bar(String)
    static protected final void foo() {  } // OK, static foo is it's own group. 
}

================

//ModifiersThatCreateGroup = "protected, private, package"
public class Test {
    private void foo() {  } //OK, 2nd group
    void bar() {  }
    public void foo(String baz) {  } //OK, "public foo" does not match special groups so it is default group
    public void foo(String baz) {  } //OK
    private static void foo(int i) {  } // VIOLATION, matches 2nd group, should be close to first method
    private final void foo(int i) {  } // VIOLATION, matches 2nd group, should be close to first method
}

//ModifiersThatCreateGroup = "protected, private, package"
public class Test {
    protected void foo() {  } //OK, matches 1st group
    void bar() {  }
    protected abstract void foo() {  } //VIOlATION, matches 1st group
    void foo(int i) {  } //OK, matches 3rd group
    public final void foo(int i) {  } // OK, "public final " does not match special groups so it is default group
}


================

//ModifiersThatCreateGroup = "public, public final, static"
public class Test {
    private void foo() {  } // default group
    void bar() {  } // defaut group
    static foo() {} // OK, 3rd group
    static final foo() {} // OK, matches to 3rd group
    final void foo(String baz) {  } //VIOLATION, matching to default group
    final public void foo(String baz) {  } //OK, 1st group, Attention: does not match second group, as strict order of modifiers is in regexp
    private void foo(int i) {  } // VIOLATION, default group, should be close to first method
    public final void foo(int i) {  } // OK, 2nd group
    abstract foo() {} //VIOLATION, matches to default, should be close to first method
}


//ModifiersThatCreateGroup = RegExp[] = "^public .*, (protected|package), private, ^static$"
public class Test {
    private void foo() {  } // 3rd group
    void bar() {  } // second group
    static foo() {} // OK, 4th group
    static final foo() {} // OK, matches to default group
    final void foo(String baz) {  } //OK, matching to default group
    public void foo(String baz) {  } //OK, first group
    private void foo(int i) {  } // VIOLATION, 3rd group, should be close to first method
    public final void foo(int i) {  } // VIOLATION, matches to first group
    abstract foo() {} //VIOLATION, matches to default
}
```

austinarbor:
Fixes #3770

This PR completes the work initially started by @ilanKeshet to add the ability to split the overloading check into multiple different groups based on user-supplied regex patterns.

Diff Regression config: https://gist.githubusercontent.com/austinarbor/12a3207dfc1f347fd0da0083384399f0/raw/ba43cf7116d4c7c93109668869e081d1690d8316/config.xml
Diff Regression patch config: https://gist.githubusercontent.com/austinarbor/12a3207dfc1f347fd0da0083384399f0/raw/ba43cf7116d4c7c93109668869e081d1690d8316/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
@@ -20,12 +20,20 @@
 package com.puppycrawl.tools.checkstyle.checks.coding;
 
 import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Locale;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.Scope;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
+import com.puppycrawl.tools.checkstyle.utils.ScopeUtil;
 import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
 
 /**
@@ -34,31 +42,85 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * name but different signatures where the signature can differ by the number of
  * input parameters or type of input parameters or both.
  * </p>
+ * <ul>
+ * <li>
+ * Property {@code modifierGroups} - control how different types of overloaded methods
+ * can be grouped by their modifiers.
+ * Type is {@code java.util.regex.Pattern[]}.
+ * Default value is {@code ""}.
+ * </li>
+ * </ul>
  * <p>
  * To configure the check:
  * </p>
  * <pre>
- * &lt;module name=&quot;OverloadMethodsDeclarationOrder&quot;/&gt;
- * </pre>
- * <p>
- * Example of correct grouping of overloaded methods:
- * </p>
- * <pre>
- * public void foo(int i) {}
- * public void foo(String s) {}
- * public void foo(String s, int i) {}
- * public void foo(int i, String s) {}
- * public void notFoo() {}
+ * &lt;module name="OverloadMethodsDeclarationOrder"/&gt;
  * </pre>
  * <p>
  * Example of incorrect grouping of overloaded methods:
  * </p>
  * <pre>
+ * public void foo() {} // OK
+ * public final void foo(int i) {} // OK
+ * void notFoo() {} // OK
+ * public static void foo(long l) {} // Violation, should be after 'foo(int i)'
+ * protected void foo(String s) {} // OK
+ * void foo(byte b) {} // OK
+ * private static void foo(String s, int i) { } // OK
+ * public void foo(String a, String b) {} // OK
+ * private void notFoo(int i) {} // Violation, should be after 'notFoo()'
+ * public final void foo(long x, long y) {} // Violation, should be after 'foo(String a, String b)'
+ * </pre>
+ * <p>
+ * Example of correct grouping of overloaded methods:
+ * </p>
+ * <pre>
  * public void foo(int i) {} // OK
- * public void foo(String s) {} // OK
- * public void notFoo() {} // violation. Have to be after foo(String s, int i)
- * public void foo(int i, String s) {}
- * public void foo(String s, int i) {}
+ * void foo(String s) {} // OK
+ * private void foo(String s, int i) {} // OK
+ * public void foo(int i, String s) {} // OK
+ * public void notFoo() {} // OK
+ * private void notFoo(int x) {} // OK
+ * </pre>
+ * <p>
+ * To configure the check to group static and private methods separately within their own
+ * individual groups
+ * </p>
+ * <pre>
+ * &lt;module name="OverloadMethodsDeclarationOrder"&gt;
+ *   &lt;property name="modifierGroups" value=".*static.*, ^public .*, (protected|package),
+ *   private"&gt;&lt;/property&gt;
+ * &lt;/module&gt;
+ * </pre>
+ * <p>
+ * Example of code with violation
+ * </p>
+ * <pre>
+ * public void foo() {} // OK
+ * public final void foo(int i) {} // OK
+ * void notFoo() {} // OK
+ * public static void foo(long l) {} // OK
+ * protected void foo(String s) {} // OK
+ * void foo(byte b) {} // OK
+ * private static void foo(String s, int i) { } // Violation, should be after 'foo(long l)'
+ * public void foo(String a, String b) {} // Violation, should be after 'foo()'
+ * private void notFoo(int i) {} // OK
+ * public final void foo(long x, long y) {} // Violation, should be after 'foo(int i)'
+ * </pre>
+ * <p>
+ * Example of compliant code
+ * </p>
+ * <pre>
+ * public void foo() {} // OK
+ * public void foo(String a, String b) {} // OK
+ * public final void foo(int i) {} // OK
+ * public final void foo(long x, long y) {} // OK
+ * void notFoo() {} // OK
+ * public static void foo(long l) {} // OK
+ * private static void foo(String s, int i) { } // OK
+ * protected void foo(String s) {} // OK
+ * void foo(byte b) {} // OK
+ * private void notFoo(int i) {} // OK
  * </pre>
  * <p>
  * Parent is {@code com.puppycrawl.tools.checkstyle.TreeWalker}
@@ -82,6 +144,22 @@ public class OverloadMethodsDeclarationOrderCheck extends AbstractCheck {
      * file.
      */
     public static final String MSG_KEY = "overload.methods.declaration";
+
+    /**
+     * String used as separator between the regex pattern that matches a method's modifiers
+     * and the name of the method.
+     */
+    private static final String SEPARATOR = ":";
+
+    /**
+     * The abstract modifier.
+     */
+    private static final String ABSTRACT_MODIFIER = "abstract";
+
+    /**
+     * Control how different types of overloaded methods can be grouped by their modifiers.
+     */
+    private Pattern[] modifierGroups = new Pattern[0];
 
     @Override
     public int[] getDefaultTokens() {
@@ -118,6 +196,19 @@ public class OverloadMethodsDeclarationOrderCheck extends AbstractCheck {
     }
 
     /**
+     * Setter to control how different types of overloaded methods can be grouped by their
+     * modifiers.
+     *
+     * @param modifierGroupRegex the array of regex patterns
+     */
+    public void setModifierGroups(String... modifierGroupRegex) {
+        modifierGroups = new Pattern[modifierGroupRegex.length];
+        for (int index = 0; index < modifierGroupRegex.length; index++) {
+            modifierGroups[index] = CommonUtil.createPattern(modifierGroupRegex[index]);
+        }
+    }
+
+    /**
      * Checks that if overload methods are grouped together they should not be
      * separated from each other.
      *
@@ -126,27 +217,140 @@ public class OverloadMethodsDeclarationOrderCheck extends AbstractCheck {
      */
     private void checkOverloadMethodsGrouping(DetailAST objectBlock) {
         final int allowedDistance = 1;
+        final Map<String, IndexAndLine> knownMethods = new HashMap<>();
         DetailAST currentToken = objectBlock.getFirstChild();
-        final Map<String, Integer> methodIndexMap = new HashMap<>();
-        final Map<String, Integer> methodLineNumberMap = new HashMap<>();
+
         int currentIndex = 0;
         while (currentToken != null) {
             if (currentToken.getType() == TokenTypes.METHOD_DEF) {
                 currentIndex++;
-                final String methodName =
-                        currentToken.findFirstToken(TokenTypes.IDENT).getText();
-                final Integer previousIndex = methodIndexMap.get(methodName);
-                if (previousIndex != null && currentIndex - previousIndex > allowedDistance) {
-                    final int previousLineWithOverloadMethod =
-                            methodLineNumberMap.get(methodName);
-                    log(currentToken, MSG_KEY,
-                            previousLineWithOverloadMethod);
+                final int modifierGroupIndex = findModifierGroupPatternIndex(currentToken);
+                final String methodKey =
+                    String.format(Locale.ROOT, "%d%s%s", modifierGroupIndex, SEPARATOR,
+                                  currentToken.findFirstToken(TokenTypes.IDENT).getText());
+                final IndexAndLine indexAndLine =
+                    new IndexAndLine(currentIndex, currentToken.getLineNo());
+                final IndexAndLine previousMeta = knownMethods.put(methodKey, indexAndLine);
+                if (previousMeta != null
+                    && currentIndex - previousMeta.getIndex() > allowedDistance) {
+                    final int previousLineWithOverloadMethod = previousMeta.getLineNo();
+                    log(currentToken, MSG_KEY, previousLineWithOverloadMethod);
                 }
-                methodIndexMap.put(methodName, currentIndex);
-                methodLineNumberMap.put(methodName, currentToken.getLineNo());
             }
             currentToken = currentToken.getNextSibling();
         }
     }
 
+    /**
+     * Get the index of the first matching {@code modifierGroup} for the modifiers on the method.
+     *
+     * @param currentToken assumed to be a method
+     * @return index of the first modifier group that matches the modifiers on the method, or -1
+     *     if only the default group matches.
+     */
+    private int findModifierGroupPatternIndex(DetailAST currentToken) {
+        final String methodModifiers = getAllModifiersFrom(currentToken);
+        int modifierIndex = -1;
+        for (int index = 0; index < modifierGroups.length; index++) {
+            final Pattern pattern = modifierGroups[index];
+            final Matcher matcher = pattern.matcher(methodModifiers);
+            if (matcher.find()) {
+                modifierIndex = index;
+                break;
+            }
+        }
+        return modifierIndex;
+    }
+
+    /**
+     * Extract the modifier(s) of a method definition into a string.
+     *
+     * @param token expected to be TokenType.METHOD_DEF .
+     * @return modifier string for {@code token} with 'package' scope added if needed
+     */
+    private static String getAllModifiersFrom(final DetailAST token) {
+        final List<String> collectedModifiers = getExplicitModifiersFrom(token);
+        final String scope = ScopeUtil.getScope(token).getName();
+        if (!collectedModifiers.contains(scope)) {
+            collectedModifiers.add(0, scope);
+        }
+
+        // methods in interfaces can only have public, abstract, default, private,
+        // and static modifiers. a method can be abstract if and only if the only
+        // other modifier present is the public scope. If that is the case, add the implicit
+        // abstract modifier to the list
+        if (ScopeUtil.isInInterfaceBlock(token)
+            && List.of(Scope.PUBLIC.getName()).equals(collectedModifiers)) {
+            collectedModifiers.add(ABSTRACT_MODIFIER);
+        }
+
+        return String.join(" ", collectedModifiers);
+    }
+
+    /**
+     * Get the modifiers associated with a method.
+     *
+     * @param token the method token
+     * @return a list of modifiers found on the method, never empty
+     */
+    private static List<String> getExplicitModifiersFrom(final DetailAST token) {
+        final List<String> collectedModifiers = new LinkedList<>();
+        final DetailAST modifiersAst = token.findFirstToken(TokenTypes.MODIFIERS);
+        DetailAST modifier = modifiersAst.getFirstChild();
+        while (modifier != null) {
+            // skipping annotations
+            if (modifier.getType() != TokenTypes.ANNOTATION) {
+                collectedModifiers.add(modifier.getText());
+            }
+            // advance element in loop
+            modifier = modifier.getNextSibling();
+        }
+        return collectedModifiers;
+    }
+
+    /**
+     * Stores the {@code index} of a method in a class and the {@code lineNo}
+     * on which the method appears.
+     */
+    private static final class IndexAndLine {
+        /**
+         * The index of method within the containing class.
+         */
+        private final int index;
+
+        /**
+         * The line number on which the method appears in the class.
+         */
+        private final int lineNo;
+
+        /**
+         * Create a new composite tuple containing the {@code index} and {@code lineNo}
+         * of a method within a class.
+         *
+         * @param index the index of the method within the class
+         * @param lineNo the line number of the method in the class
+         */
+        private IndexAndLine(int index, int lineNo) {
+            this.index = index;
+            this.lineNo = lineNo;
+        }
+
+        /**
+         * Getter for the index.
+         *
+         * @return the method's index
+         */
+        public int getIndex() {
+            return index;
+        }
+
+        /**
+         * Getter for the line number.
+         *
+         * @return the method's line number
+         */
+        public int getLineNo() {
+            return lineNo;
+        }
+    }
 }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheckTest.java
@@ -61,6 +61,128 @@ public class OverloadMethodsDeclarationOrderCheckTest
     }
 
     @Test
+    public void testOverloadMethodsDeclarationOrderRecords2() throws Exception {
+
+        final String[] expected = {
+            "21:9: " + getCheckMessage(MSG_KEY, 16),
+            "30:9: " + getCheckMessage(MSG_KEY, 25),
+            "33:9: " + getCheckMessage(MSG_KEY, 27),
+        };
+        verifyWithInlineConfigParser(
+            getNonCompilablePath("InputOverloadMethodsDeclarationOrderRecords2.java"),
+            expected);
+    }
+
+    @Test
+    public void testBackwardsCompatibilityCatchAllGroup() throws Exception {
+        final String[] expected = {
+            "20:5: " + getCheckMessage(MSG_KEY, 15),
+            "35:9: " + getCheckMessage(MSG_KEY, 30),
+            "48:5: " + getCheckMessage(MSG_KEY, 45),
+            "80:5: " + getCheckMessage(MSG_KEY, 75),
+        };
+
+        verifyWithInlineConfigParser(
+            getPath("InputOverloadMethodsDeclarationOrder2.java"),
+            expected
+        );
+    }
+
+    @Test
+    public void testBackwardsCompatibilityEmptyModifiers() throws Exception {
+        final String[] expected = {
+            "20:5: " + getCheckMessage(MSG_KEY, 15),
+            "35:9: " + getCheckMessage(MSG_KEY, 30),
+            "48:5: " + getCheckMessage(MSG_KEY, 45),
+            "80:5: " + getCheckMessage(MSG_KEY, 75),
+        };
+
+        verifyWithInlineConfigParser(
+            getPath("InputOverloadMethodsDeclarationOrder2.java"),
+            expected
+        );
+    }
+
+    @Test
+    public void testOverloadMethodsDeclarationOrderStaticGrouped() throws Exception {
+        final String[] expected = {
+            "29:5: " + getCheckMessage(MSG_KEY, 24),
+            "37:5: " + getCheckMessage(MSG_KEY, 34),
+            "39:5: " + getCheckMessage(MSG_KEY, 35),
+            "47:5: " + getCheckMessage(MSG_KEY, 44),
+            "49:5: " + getCheckMessage(MSG_KEY, 45),
+            "57:5: " + getCheckMessage(MSG_KEY, 54),
+            "61:5: " + getCheckMessage(MSG_KEY, 58),
+            "68:5: " + getCheckMessage(MSG_KEY, 65),
+            "79:5: " + getCheckMessage(MSG_KEY, 74),
+            "92:5: " + getCheckMessage(MSG_KEY, 87),
+            "105:5: " + getCheckMessage(MSG_KEY, 102),
+        };
+        verifyWithInlineConfigParser(
+            getPath("InputOverloadMethodsDeclarationOrder3.java"), expected);
+    }
+
+    @Test
+    public void testOverloadMethodsDeclarationOrderAbstractAndPublicOrProtectedAndStatic()
+            throws Exception {
+        final String[] expected = {
+            "28:5: " + getCheckMessage(MSG_KEY, 25),
+            "30:5: " + getCheckMessage(MSG_KEY, 23),
+            "34:5: " + getCheckMessage(MSG_KEY, 26),
+            "44:5: " + getCheckMessage(MSG_KEY, 40),
+        };
+        verifyWithInlineConfigParser(
+            getPath("InputOverloadMethodsDeclarationOrder4.java"), expected);
+    }
+
+    @Test
+    public void testOverloadMethodsDeclarationOrderOrderMatters() throws Exception {
+        final String[] expected = {
+            "16:5: " + getCheckMessage(MSG_KEY, 11),
+            "19:5: " + getCheckMessage(MSG_KEY, 16),
+            "22:5: " + getCheckMessage(MSG_KEY, 19),
+            "30:5: " + getCheckMessage(MSG_KEY, 27),
+            "34:5: " + getCheckMessage(MSG_KEY, 30),
+            "36:5: " + getCheckMessage(MSG_KEY, 32),
+        };
+        verifyWithInlineConfigParser(
+            getPath("InputOverloadMethodsDeclarationOrder5.java"), expected);
+    }
+
+    @Test
+    public void testOverloadMethodsDeclarationOrderComplexRegex() throws Exception {
+        final String[] expected = {
+            "29:5: " + getCheckMessage(MSG_KEY, 26),
+            "39:5: " + getCheckMessage(MSG_KEY, 33),
+            "47:5: " + getCheckMessage(MSG_KEY, 44),
+            "56:5: " + getCheckMessage(MSG_KEY, 51),
+            "64:5: " + getCheckMessage(MSG_KEY, 61),
+            "77:5: " + getCheckMessage(MSG_KEY, 70),
+            "80:5: " + getCheckMessage(MSG_KEY, 74),
+            "92:5: " + getCheckMessage(MSG_KEY, 84),
+            "95:5: " + getCheckMessage(MSG_KEY, 88),
+            "103:5: " + getCheckMessage(MSG_KEY, 100),
+            "106:5: " + getCheckMessage(MSG_KEY, 103),
+        };
+        verifyWithInlineConfigParser(
+            getPath("InputOverloadMethodsDeclarationOrder6.java"), expected);
+    }
+
+    @Test
+    public void testOverloadMethodsDeclarationOrderDefaultModifiers() throws Exception {
+        final String[] expected = {
+            "16:5: " + getCheckMessage(MSG_KEY, 11),
+            "19:5: " + getCheckMessage(MSG_KEY, 13),
+            "35:5: " + getCheckMessage(MSG_KEY, 26),
+            "40:5: " + getCheckMessage(MSG_KEY, 28),
+            "43:5: " + getCheckMessage(MSG_KEY, 37),
+            "48:5: " + getCheckMessage(MSG_KEY, 40),
+        };
+        verifyWithInlineConfigParser(
+            getPath("InputOverloadMethodsDeclarationOrder8.java"), expected);
+    }
+
+    @Test
     public void testTokensNotNull() {
         final OverloadMethodsDeclarationOrderCheck check =
             new OverloadMethodsDeclarationOrderCheck();
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder2.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder2.java
@@ -1,0 +1,100 @@
+/*
+OverloadMethodsDeclarationOrder
+modifierGroups = .*
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.overloadmethodsdeclarationorder;
+
+class InputOverloadMethodsDeclarationOrder2
+{
+    public void overloadMethod(int i) { } // ok
+
+    public void overloadMethod(String s) { } // ok
+
+    public void overloadMethod(boolean b) { } // ok
+
+    public void fooMethod() { } // ok
+
+    // violation below 'All overloaded methods.*Previous overloaded method located at.*'
+    public void overloadMethod(String s, Boolean b, int i) {
+
+    }
+
+    InputOverloadMethodsDeclarationOrder anonymous = new InputOverloadMethodsDeclarationOrder()
+    {
+        public void overloadMethod(int i) { } // ok
+
+        public void overloadMethod(String s) { } // ok
+
+        public void overloadMethod(boolean b) { } // ok
+
+        public void fooMethod() { } // ok
+
+        // violation below 'All overloaded methods.*Previous overloaded method located at.*'
+        public void overloadMethod(String s, Boolean b, int i)
+        {
+            //some foo code
+        }
+    };
+}
+
+interface FooableBackwardsCompatible
+{
+    public abstract void foo(int i); // ok
+    public abstract void foo(String s); // ok
+    public abstract void noFoo(); // ok
+    // violation below 'All overloaded methods.*Previous overloaded method located at.*'
+    public abstract void foo(String s, Boolean b, int i);
+    private void foo() { } // ok
+}
+
+enum FooTypeBackwardsCompatible {
+    Strategy(""),
+    Shooter(""),
+    RPG("");
+
+    private String description;
+
+    private FooTypeBackwardsCompatible(String description) {
+        this.description = description;
+    }
+
+    public String getDescription() { // ok
+        return description;
+    }
+
+    public void setDescription(String description) { // ok
+        this.description = description;
+    }
+
+    public void overloadMethod(int i) { } // ok
+
+    public void overloadMethod(String s) { } // ok
+
+    public void overloadMethod(boolean b) { } // ok
+
+    public void fooMethod() { } // ok
+
+    // violation below 'All overloaded methods.*Previous overloaded method located at.*'
+    public void overloadMethod(String s, Boolean b, int i)
+    {
+        //some foo code
+    }
+}
+
+enum Foo2BackwardsCompatible {
+    VALUE {
+        public void value() {
+            value("");
+        } // ok
+
+        public void middle() { } // ok
+
+        public void value(String s) { } // ok
+    };
+}
+
+@interface ClassPreambleBackwardsCompatible {
+    String author(); // ok
+}
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder5.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder5.java
@@ -1,0 +1,37 @@
+/*
+OverloadMethodsDeclarationOrder
+modifierGroups = public$, public final, static
+
+
+
+*/
+package com.puppycrawl.tools.checkstyle.checks.coding.overloadmethodsdeclarationorder;
+
+class InputOverloadMethodsDeclarationOrder5 {
+    private void foo() { } // ok
+    void bar() { } // ok
+    static void foo(int i) { } // ok
+    static final void foo(String baz) { } // ok
+    // violation below 'All overloaded methods.*Previous overloaded method located at.*'
+    final void foo(String baz, int i) { }
+    final public void foo(String baz, long l) { } // ok
+    // violation below 'All overloaded methods.*Previous overloaded method located at.*'
+    private void foo(int i , String baz) { }
+    public final void foo(int i, long l) { } // ok
+    // violation below 'All overloaded methods.*Previous overloaded method located at.*'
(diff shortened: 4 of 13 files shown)


ilanKeshet:
Hi @austinarbor Thank you for taking on this effort. you have my blessing 🙏

config/suppressions.xml
```diff
@@ -160,6 +160,8 @@
             files="[\\/]src[\\/]test[\\/]java[\\/]com[\\/]puppycrawl[\\/]tools[\\/]checkstyle[\\/]PackageObjectFactoryTest.java"/>
   <suppress id="ImportControlTest"
             files="[\\/]src[\\/]test[\\/]java[\\/]com[\\/]puppycrawl[\\/]tools[\\/]checkstyle[\\/]utils[\\/]TokenUtilTest.java"/>
+  <suppress id="ImportControlTest"
```
austinarbor, line 163: I added this so that I could use reflection in order to achieve 100% code coverage on the `equals` method in the `OverloadDescriptor` inner class of `OverloadMethodsDeclarationOrderCheck`. I had a lot of difficulty getting all of the checks to pass for this class due to:

1. IntelliJ inspections do not allow an inner package-level class declaration which would have allowed me to write non-reflection based unit tests for all of the branches
2. Jacoco requires 100% branch coverage on all PRs

So I think the only options were to allow usage of reflection to reach the 100% coverage, or add an exception in the `pom.xml` to allow a lesser coverage amount for this class. If there are any solutions I didn't think of, let me know.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrderRecords.java
```diff
@@ -21,7 +21,7 @@ public void notFoo() {
         public void foo(int i, String s) { // violation
         }
 
-        public void foo(String s, int i) {
+        public void foo(String s, int i) { // violation
```
austinarbor, line 24: I'm wondering why this wasn't marked as a violation before (along with lines 63, 66)? By rule this should be next to `foo` on line 15. If user only fixes violation for line 21, the next check they will get a violation for this method. Without this PR, it's probably obvious to user what methods need to be grouped together. However, if / when this PR is merged, complex regex patterns can make it not as obvious where a method should be put after. It might become burdensome for users to have to repeatedly run the checks as they fix violations one-by-one instead of seeing them all at once.  I can easily revert to old behavior, but IMO I think this is correct

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheckTest.java
```diff
@@ -75,4 +183,38 @@ public void testTokensNotNull() {
             .isNotNull();
     }
 
+    @Test
+    public void testOverloadDescriptorEquals()
```
nick-mancuso, line 187: Pure unit test is not much better than suppression, we avoid this technical debt at all costs. Please remove this test so that we can see/ work through all pitest failures. If code is not being hit, remove it and run check regression report to help find cases to cover new code.

nick-mancuso (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheckTest.java
nick-mancuso, line 187: Pure unit test is not much better than suppression, we avoid extension of our technical debt at all costs. Please remove this test to let us see all and work through pitest failures. If code is problematic, remove it and run check regression report in order to find test cases to cover new code.

nick-mancuso (changes requested):

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrderRecords.java
nick-mancuso: > I can easily revert to old behavior, but IMO I think this is correct

If this is the case, can we do this in the scope of a separate issue?

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheckTest.java
austinarbor: @nick-mancuso I was only able to get 100% coverage by making the equals method
```java
    @Override
     public boolean equals(final Object o) {
         final OverloadDescriptor other = (OverloadDescriptor) o;
         return Objects.equals(pattern.pattern(), other.pattern.pattern())
                && Objects.equals(methodName, other.methodName);
      }
```

However, when I do this I get two PMD errors:

```
[ERROR] Medium: Equals method for com.puppycrawl.tools.checkstyle.checks.coding.OverloadMethodsDeclarationOrderCheck$OverloadDescriptor assumes the argument is of type OverloadMethodsDeclarationOrderCheck$OverloadDescriptor [com.puppycrawl.tools.checkstyle.checks.coding.OverloadMethodsDeclarationOrderCheck$OverloadDescriptor] At OverloadMethodsDeclarationOrderCheck.java:[line 377] BC_EQUALS_METHOD_SHOULD_WORK_FOR_ALL_OBJECTS
[ERROR] High: com.puppycrawl.tools.checkstyle.checks.coding.OverloadMethodsDeclarationOrderCheck$OverloadDescriptor.equals(Object) does not check for null argument [com.puppycrawl.tools.checkstyle.checks.coding.OverloadMethodsDeclarationOrderCheck$OverloadDescriptor] At OverloadMethodsDeclarationOrderCheck.java:[lines 377-378] NP_EQUALS_SHOULD_HANDLE_NULL_ARGUMENT
```

Given the encapsulation of the class, both of the scenarios (comparing for equality vs a null, and comparing for equality against object with a different class) are impossible to occur. Likewise, I also don't have any way of testing those branches without reflection to meet jacoco's requirements of 100% branch coverage. How should I resolve this?

austinarbor:
@nick-mancuso Other than PMD/branch coverage issue mentioned [here](https://github.com/checkstyle/checkstyle/pull/12086#discussion_r953206083) I have everything else fixed and below validations are passing locally, so I think we are pretty close. Will wait to push up latest changes until I have some direction on the last violation decision.

```shell
mvn clean verify
./.ci/validation.sh checkstyle-and-sevntu # pass
./.ci/validation.sh spotbugs-and-pmd # pass
mvn -e --no-transfer-progress -P"pitest-coding-2" clean test-compile org.pitest:pitest-maven:mutationCoverage # pass
```

nick-mancuso:
Github, generate report

austinarbor:
I added some lesser jacoco requirements for the inner class. I tried to create a jacoco rule to _just_ target `com.puppycrawl.tools.checkstyle.checks.coding.OverloadMethodsDeclarationOrderCheck.OverloadDescriptor#equals(Object)` but I couldn't get it to work correctly

austinarbor:
Github, generate report

austinarbor:
Github, generate report

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheckTest.java
nick-mancuso: We will not extend our technical debt, if we have this much code that we cannot hit, it is time to rethink implementation.

austinarbor: This is the only code that doesn't get hit, it is impossible for this scenario to occur. PMD requires this check to be there.
```java
if (o == null || getClass() != o.getClass()) {
                return false;
}
```

I will move the inner class to https://github.com/checkstyle/checkstyle/tree/master/src/main/java/com/puppycrawl/tools/checkstyle/api and make it public so I can add tests for this branch.

austinarbor: Actually..I think I can use two maps and get rid of the need for the inner class. Will push up later today

austinarbor:
Github, generate report

austinarbor:
Ok! I think I was able to get all of the design and coverage concerns addressed, and latest report and pitests are passing. The 3 cancelled checks are due to brownout by the provider and should be unrelated to my changes. Let me know if anything else is needed from my side! The only other thing I was unsure of is what version to use for the `since` on the new argument  [here](https://github.com/checkstyle/checkstyle/pull/12086/files#diff-cd8c9a8c8d730d1c9a128300a5b0b8302beac605ae7e9d8486e18aae24c44270R5843) so I just used the next minor version. Let me know if this should be changed to something else.

image

nrmancuso:
@austinarbor we have not forgot about your PR, we are in final push of GSOC :) We will come back to PR soon.

austinarbor:
Let me know when you guys are ready and I will rebase off of latest master

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -117,36 +185,167 @@ public void visitToken(DetailAST ast) {
...
      * Checks that if overload methods are grouped together they should not be
      * separated from each other.
      *
-     * @param objectBlock
-     *        is a class, interface or enum object block.
+     * @param objectBlock is a class, interface or enum object block.
      */
     private void checkOverloadMethodsGrouping(DetailAST objectBlock) {
         final int allowedDistance = 1;
+        final Map<String, IndexAndLine> knownMethods = new HashMap<>();
```
nrmancuso, line 212: Can we use a list of `MethodDetails`as proposed below?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -51,14 +72,40 @@
...
  * </p>
  * <pre>
  * public void foo(int i) {} // OK
  * public void foo(String s) {} // OK
- * public void notFoo() {} // violation. Have to be after foo(String s, int i)
- * public void foo(int i, String s) {}
- * public void foo(String s, int i) {}
+ * public void notFoo() {} // OK
+ * private void foo(boolean b) {} // OK
+ * protected void foo(String s, int i) {} // Violation. Has to be after foo(String s)
```
nrmancuso, line 91: please make all `// violation` comments consistent in capitalization and spacing. Also, please do `// violation, should be after 'foo(String s)'` to make examples as easy to read and understand as possible, for all cases.

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
austinarbor: Before I make changes, just to confirm, you are looking for something like:
```java
List<MethodDetails> knownMethods = new ArrayList<>();
for(MethodDetails md: knownMethods) {
    if(md.getMethodName().equals(thisMethodName) && md.getModifiers().equals(thisMethodsModifiers) {
          // confirm method is immediately after previous match
    } else {
         // haven't seen this collection of modifiers yet, add to our list
    }
}
```

Do you think going from O(1) lookup time to O(n) will be a concern?

nrmancuso: @austinarbor this is a good point; I like the map implementation, but I am not a big fan of the "description string". This string seems difficult to reason about/ debug in future maintenance. What do you think about `Map<MethodDetails, Integer> knownMethods`, where integer is the index and we store the line number in `MethodDetails`?

austinarbor: @nrmancuso I had something similar to `Map<MethodDetails, Integer>` when I first opened the PR, but that creates other issues with code coverage as mentioned [here](https://github.com/checkstyle/checkstyle/pull/12086#discussion_r953206083). If we create a `private static` inner class and use it as a key in the map, we need to override `equals` and `hashcode`. 
1. PMD requires `null` check in the equals method of the class
2. It's impossible in the code for the value to ever be null, so we can't test this scenario via normal processing
3. Because the inner class is private we can't write unit test for this scenario without using reflection

Does that make sense?

We can get code coverage to pass if we make the `MethodDetails` a public class outside of the check, (because then we can write unit tests on the public method), but I didn't know where there would be a good place to put it. I also tried to make the inner class `protected` / `package protected` so I can call `equals` from the test class, but that caused other checkstyle checks to fail as well becauase that visibility was disallowed.

nrmancuso: Yes, it has been awhile since we looked at first iteration, thanks for the reminder. One last thought; what do you think of using a simple data class that contains the same information? We could do `Map<MethodDetails, IndexAndLine>` where `MethodDetails` is comprised of the pattern and method name?

austinarbor: Hmm..wouldn't that have the same issue since `MethodDetails` is being used as a key?

nrmancuso: It shouldn’t, since we do not need to override equals/hashcode in this case, right?

austinarbor: Why wouldn't we? We would need to override for any class that's used a key in the map otherwise it will use a referential equality check. In java 17 then we could just use a record, but there isn't a way around overriding equals and hashcode that I'm aware of for java 11....unless there's a trick I don't know about?

nrmancuso: >Why wouldn't we?

I was thinking of some hack with custom map, but this is not good either. Let's proceed with custom description string then, we can get all test cases and behavior of update correct first and give us some time to think on this.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -117,36 +185,167 @@ public void visitToken(DetailAST ast) {
...
+
+            // advance element in loop
+            modifier = modifier.getNextSibling();
+        }
+
+        if (!foundScopeModifier) {
+            // prepend string 'package' to mark default scope.
+            // the package scope modifier is placed at the beginning to conform
+            // with modifiers order as suggested in sections 8.1.1, 8.3.1 and 8.4.3 of the JLS
+            collectedModifiers.add(0, "package");
```
nrmancuso, line 288: What if we are in an interface?

nrmancuso (changes requested):

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder2.java
```diff
@@ -0,0 +1,99 @@
...
+    public void overloadMethod(int i) { }
+
+    public void overloadMethod(String s) { }
+
+    public void overloadMethod(boolean b) { }
+
+    public void fooMethod() { }
+
+    //violation: because overloads never split
+    public void overloadMethod(String s, Boolean b, int i) // violation
```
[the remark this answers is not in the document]
austinarbor: Is the full violation message required in the test file as well? I can't find any examples of the violation message being multiple lines, and I can't get it to work no matter what I try. It seems like the check is only validating a subset of the message. For example, the below passes
```java
    //violation below 'All overloaded methods should be placed next to each other.'
    public void overloadMethod(String s, Boolean b, int i) {

    }
```

The full violation message is quite unwieldy
```
All overloaded methods should be placed next to each other. Placing non-overloaded methods in between overloaded methods with the same type is a violation. Previous overloaded method located at line '15'.
```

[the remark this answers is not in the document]
nrmancuso: You should be able to do something like:
```
    //violation below 'All overloaded methods should be placed .* Previous overloaded method located at line '15'.'


```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
austinarbor: I can't combine the `should be after..` with https://github.com/checkstyle/checkstyle/pull/12086#discussion_r974851851 can I? Seems like I need to do one or the other

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder5.java
```diff
@@ -0,0 +1,23 @@
+/*
+OverloadMethodsDeclarationOrder
+modifierGroups = public final
```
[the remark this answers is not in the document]
austinarbor: Based on description of the issue, I don't think the 7th example is correct. `public` will match `public final` due to the regex. In my test I made the first pattern `public$` to make the test case work, but let me know if you had something else in mind

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder2.java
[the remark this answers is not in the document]
austinarbor: @nrmancuso I don't think this syntax supports multiple lines, see https://github.com/checkstyle/checkstyle/blob/f7ce9ad82249706813e2ec3a454205257566d2a2/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L88 

It is running the regex check one line at a time, so the comment can spill over onto multiple lines. what do you suggest?

austinarbor:
@nrmancuso sorry for the delay, I was traveling quite a bit the past few months and didn't have much time to get back to this until now. I've added some more comments

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
austinarbor: good catch, I think this should be handled now

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
austinarbor: disregard above comment

nrmancuso:
> @nrmancuso sorry for the delay, I was traveling quite a bit the past few months and didn't have much time to get back to this until now. I've added some more comments

No worries, thanks for coming back to your PR :)

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder2.java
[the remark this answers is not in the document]
austinarbor: I just wanted to make sure cutting out a chunk of the message with `.*` to make it fit on a single line would be ok, thanks for clarifying. Does the line number at the end of the error need to be included in the BDD comment? IMO we don't need it because it's being included in the test code, but if you would prefer to have it in the comment as well i'll make sure to include it in the BDD comment also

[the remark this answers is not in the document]
nrmancuso: > IMO we don't need it because it's being included in the test code

Agreed.

austinarbor:
@nrmancuso I think all the requested changes should be good now...I'm not sure what happened with the Travis CI build, do you know what's wrong there?

rnveach:
Travis is known to have issues. Ping us if it looks like a fluke (connection reset, apt-get failures, etc...) and we will restart it. Just beware and verify that it isn't a real issue.

austinarbor:
@rnveach it looks like an `apt-get` issue according to the logs, I didn't change anything between these commits that would break travis. no rush on restarting it

rnveach:
Yes, which is a fluke. Travis is already restarted and passed.

nrmancuso (changes requested):
@austinarbor looks like we are missing some modifiers in test input, please see https://docs.oracle.com/javase/specs/jls/se19/html/jls-8.html#jls-8.4.3. We must cover all modifiers from most recent java grammar. Let's get all test input in place before we proceed further on implementation.

I think tracking/determining implied modifiers will be one of the trickiest parts of this new property, let's make sure it is well tested.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -117,36 +221,200 @@ public void visitToken(DetailAST ast) {
...
+            String defaultModifier = PACKAGE_MODIFIER;
+            if (isInsideInterface) {
+                defaultModifier = PUBLIC_MODIFIER;
+            }
+            collectedModifiers.add(0, defaultModifier);
+        }
+
+        // methods in interfaces can only have public, abstract, default, or static
+        // modifiers. abstract is only valid when used with just the public
+        // modifier, so simply add it to the end of the list if it is missing
```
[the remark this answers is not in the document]
nrmancuso: Let's make sure we have some private methods in interfaces, I do not see this in input.

austinarbor:
@nrmancuso any thoughts on https://github.com/checkstyle/checkstyle/pull/12086#issuecomment-1340475114 ?

nrmancuso:
> @nrmancuso any thoughts on [#12086 (comment)](https://github.com/checkstyle/checkstyle/pull/12086#issuecomment-1340475114) ?

@romani checker requires usage of their utility, RegexUtil, to add a condition like:
```
	if (!RegexUtil.isRegex(regex, 1)) {
	    System.out.println("Input is not a regular expression \"" + regex
		    + "\": " + RegexUtil.regexException(regex).getMessage());
	    System.exit(1);
	}

```
In order to print a user friendly message with failure.


Are we good to use this utility in all `Pattern` setters?

romani:
https://checkerframework.org/releases/1.9.0/tutorial/webpages/user-input-cmd.html
having compile dependency to checker might be too much for now.
after reading documentation with example attentively, I do not think we should use this Check - `Validating User Input`.
We are library code, we do not know where we are configured. We throw exception to let upper level decide what to do with this exception. I think our pitest will  do all required work to demand full coverage for code and validate all indexes of groups.
Again If we fail with exception - it is ok for us.

I am voting to deactivate `Validating User Input`.

nrmancuso:
@austinarbor please just suppress checker failure, we will deal with it in scope of https://github.com/checkstyle/checkstyle/issues/12567

austinarbor:
@nrmancuso I looked at all the failures from the most recent build and none look related to my changes

romani:
Please apply this patch to your branch https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=12634&view=logs&j=6a833393-c551-5000-0916-5000fd3b86e7&t=bdbbd129-77f0-5d73-ce7d-db58d34c8358&l=39

austinarbor:
all set

austinarbor:
@nrmancuso sorry to bother but just wanted to see if you were waiting for anything from me on this

austinarbor:
Any updates on this?

rnveach:
@austinarbor rebase on latest master and look into pitest failures if they still remain.
https://github.com/checkstyle/checkstyle/actions/runs/3869525725/jobs/6625770158#step:6:452
CI should be green, but don't let pitest fool you into thinking you need to remove functionality. We are looking for tests to show your code is needed or prove why it is redundant or not a concern if it is removed.

austinarbor:
@rnveach pitest passes locally, let me know if you want me to run anything else before kicking off other workflows

```shell
 ❯ mvn -e --no-transfer-progress -P"pitest-coding-2" clean test-compile org.pitest:pitest-maven:mutationCoverage
--------------------------------------------------------------------------------
================================================================================
- Timings
================================================================================
> pre-scan for mutations : < 1 second
> scan classpath : < 1 second
> coverage and dependency analysis : 2 seconds
> build mutation tests : < 1 second
> run mutation analysis : 8 minutes and 49 seconds
--------------------------------------------------------------------------------
> Total  : 8 minutes and 52 seconds
--------------------------------------------------------------------------------
================================================================================
- Statistics
================================================================================
>> Line Coverage: 1797/1797 (100%)
>> Generated 3648 mutations Killed 3597 (99%)
>> Mutations with no coverage 0. Test strength 99%
>> Ran 9612 tests (2.63 tests per mutation)
Enhanced functionality available at https://www.arcmutate.com/
[INFO] BUILD SUCCESS
[INFO] Total time:  09:08 min
[INFO] Finished at: 2023-03-05T19:03:18-05:00
```

rnveach:
https://github.com/checkstyle/checkstyle/actions/runs/4338664722/jobs/7575943113#step:6:461
CI says something different. It looks like you are not running the full/right command.

austinarbor:
@rnveach sorry about that, I forgot to run the groovy validation script. should be all set now. As far as I can tell, https://github.com/checkstyle/checkstyle/actions/runs/4338664701/jobs/7575942795 is not related to my changes
```shell
--------------------------------------------------------------------------------
> Total  : 8 minutes and 53 seconds
--------------------------------------------------------------------------------
================================================================================
- Statistics
================================================================================
>> Line Coverage: 1796/1796 (100%)
>> Generated 3635 mutations Killed 3590 (99%)
>> Mutations with no coverage 0. Test strength 99%
>> Ran 9307 tests (2.56 tests per mutation)
Enhanced functionality available at https://www.arcmutate.com/
[INFO] BUILD SUCCESS
[INFO] Total time:  09:06 min
[INFO] Finished at: 2023-03-06T01:07:24-05:00
~/git/my-checkstyle gh-3770-combined *4 +2 !3 ❯ groovy ./.ci/pitest-survival-check-xml.groovy pitest-coding-2                                                                     9m 7s
No new surviving mutation(s) found.
``

**Comment by austinarbor** [comment]
Am I able to add the below  ignore these checker errors? I am not sure what needs/can be done to resolve them other than not using a set at all
```xml
    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java</fileName>
    <specifier>methodref.param</specifier>
    <message>Incompatible parameter type for arg0</message>
    <lineContent>if (collectedModifiers.stream().noneMatch(SCOPE_MODIFIERS::contains)) {</lineContent>
    <details>
      found   : @GuardSatisfied Object
      required: @GuardedBy String
      Consequence: method in @GuardedBy Set&lt;@GuardedBy String&gt;
      @GuardedBy boolean contains(@GuardSatisfied Set&lt;@GuardedBy String&gt; this, @GuardSatisfied Object p0)
      is not a valid method reference for method in @GuardedBy Predicate&lt;@GuardedBy String&gt;
      @GuardedBy boolean test(@GuardedBy Predicate&lt;@GuardedBy String&gt; this, @GuardedBy String p0)
    </details>
  </checkerFrameworkError>

    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java</fileName>
    <specifier>methodref.receiver.bound</specifier>
    <message>Incompatible receiver type</message>
    <lineContent>if (collectedModifiers.stream().noneMatch(SCOPE_MODIFIERS::contains)) {</lineContent>
    <details>
      found   : @GuardedBy Set&lt;@GuardedBy String&gt;
      required: @GuardSatisfied Set&lt;@GuardedBy String&gt;
      Consequence: method
      @GuardedBy Set&lt;@GuardedBy String&gt;
      is not a valid method reference for method in @GuardedBy Set&lt;@GuardedBy String&gt;
      @GuardedBy boolean contains(@GuardSatisfied Set&lt;@GuardedBy String&gt; this, @GuardSatisfied Object p0)
    </details>
  </checkerFrameworkError>
```

austinarbor:
I switched to `Collections.disjoint` instead of a stream and the suppressions are no longer required now

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -34,31 +40,74 @@
...
+ * &lt;module name="OverloadMethodsDeclarationOrder"/&gt;
+ * </pre>
+ * <p>
+ * Example of incorrect grouping of overloaded methods:
+ * </p>
+ * <pre>
+ * public void foo(int i) {} // OK
+ * public void foo(String s) {} // OK
+ * public void notFoo() {} // Violation, should be after 'foo(int i, String s)'
+ * public void foo(String s, int i) {} // OK
```
rnveach, line 64: Example is not correct from my testing. Violation goes on the next line not on `notFoo`. Examples should mimic behavior of code otherwise people will think behavior is not right as example says something else.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -117,36 +218,184 @@ public void visitToken(DetailAST ast) {
...
+            if (isInsideInterface) {
+                defaultModifier = PUBLIC_MODIFIER;
+            }
+            collectedModifiers.add(0, defaultModifier);
+        }
+
+        // methods in interfaces can only have public, abstract, default, private,
+        // static modifiers. abstract is only valid when used with just the public
+        // modifier, so simply add it to the end of the list if it is missing
+        if (isInsideInterface && missingAbstractModifier(collectedModifiers)) {
```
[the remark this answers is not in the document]
austinarbor: We check for `public` inside `missingAbstractModifier` https://github.com/checkstyle/checkstyle/pull/12086/files#diff-7fb7836b6edbcd2dfadb3ab0afa15939587026819fb95b06770c48f7dbb7e04eR353 . According to [JLS](https://docs.oracle.com/javase/specs/jls/se9/html/jls-9.html#jls-9.4), An interface method cannot be abstract if it is static, private, or has default. 

> An interface method lacking a private, default, or static modifier is implicitly abstract. Its body is represented by a semicolon, not a block. It is permitted, but discouraged as a matter of style, to redundantly specify the abstract modifier for such a method declaration

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -117,36 +218,184 @@ public void visitToken(DetailAST ast) {
...
+     * has only one element, and that element is the literal string {@code public}. Based on the
+     * <a href="https://docs.oracle.com/javase/specs/jls/se9/html/jls-9.html#jls-9.4">JLS</a>
+     * this is the only valid combination use of the {@code abstract} modifier when inside of
+     * an interface.
+     *
+     * @param modifiers the list of modifiers for the method
+     * @return {@code true} if {@code abstract} should be added to the modifier list
+     */
+    private static boolean missingAbstractModifier(final List<String> modifiers) {
+        return modifiers.size() == 1 && PUBLIC_MODIFIER.equals(modifiers.get(0));
```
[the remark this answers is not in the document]
austinarbor: I did it this way because it's faster to check the size and explicit existence of `public` as the first item than to check all items to see if they equal to `static`, `default`, or `private`.  Since this only applies to the modifiers on methods, If a method inside of an interface is declared `public static` then it can never be abstract

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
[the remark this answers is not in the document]
austinarbor: we could simplify this to 
```java
if (ScopeUtil.isInInterfaceBlock(token) && List.of(Scope.PUBLIC.getName()).equals(collectedModifiers)) {
  collectedModifiers.add(ABSTRACT_MODIFIER);
}
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -117,36 +218,184 @@ public void visitToken(DetailAST ast) {
...
     }
 
+    /**
+     * Compare the modifiers on a given method with the {@code modifierGroups} and find the
+     * first match to create a unique identifier for the group and method name combination.
+     *
+     * @param currentToken assumed to be a method
+     * @return string that uniquely represents the {@code modifierGroup} and method name
+     */
+    private String extractMethodDescription(DetailAST currentToken) {
```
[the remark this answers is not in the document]
austinarbor: How's this sound?
```java
 /**
     * Get the first matching {@code modifierGroup} for the modifiers on the method
     * and combine it with the method's name (separated by {@code SEPARATOR}) to create
     * a unique identifier for the pair.
     *
     * @param currentToken assumed to be a method
     * @return string that uniquely represents the {@code modifierGroup} and method name
     */
    private String getMethodDescriptor(DetailAST currentToken) {
```

austinarbor:
I have all changes done locally, will wait to push up latest changes until conclusion is reached on [1](https://github.com/checkstyle/checkstyle/pull/12086#discussion_r1128839747), [2](https://github.com/checkstyle/checkstyle/pull/12086#discussion_r1128563596), and [3](https://github.com/checkstyle/checkstyle/pull/12086#discussion_r1128562510)

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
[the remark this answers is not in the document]
rnveach: > We check for public inside missingAbstractModifier 

I see now, so the real issue is the bad method name. `missingAbstractModifier` would have been better to be named `hasOnlyPublicModifier`. You named it in the context of its only usage instead of what it does.

It doesn't seem to matter if this method is going away.

> we could simplify this to the below

With the usage of `ScopeUtil.getDefaultScope`, I would hope most of the custom code would go away.

Must `collectedModifiers` contain all modifiers, including non-explicit modifiers? How do we determine what order non-explicit modifiers are in? Are they always at the end, beginning, etc...? This has to be explicitly defined in the documentation with examples to show the user.
@romani ping

rnveach:
3's code seems to be going away.
1 and 2 is a still bigger discussion.
I recommend to push the code you think is best and I will resolve items in the next pass as I continue looking.

> https://github.com/checkstyle/checkstyle/pull/12086#issuecomment-1226614467

Also your regression report seems to be a no harm regression. We will need a completely separate report to show a usage to help ensure nothing bad shows up.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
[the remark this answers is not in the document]
austinarbor: I don't remember where I saw it, but I thought I saw somewhere that implicit modifiers should be added in the JLS recommended order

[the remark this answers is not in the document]
rnveach: Package makes sense, but the other implicit modifiers took me off guard. Your basically using similar detection defined in Scope, ClassMemberImpliedModifier, InterfaceMemberImpliedModifier, RedundantModifier. The implied check has a whole nest of things like static, abstract, etc.. which can all be implied. https://checkstyle.org/config_modifier.html#InterfaceMemberImpliedModifier_Properties

I am just confirming this is intended before we go further.

rnveach:
@austinarbor Do not resolve comments. Reviewers will do that as we double check.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -34,31 +41,74 @@
...
+ * &lt;module name="OverloadMethodsDeclarationOrder"/&gt;
+ * </pre>
+ * <p>
+ * Example of incorrect grouping of overloaded methods:
+ * </p>
+ * <pre>
+ * public void foo(int i) {} // OK
+ * public void foo(String s) {} // OK
+ * public void notFoo() {} // OK
+ * public void foo(String s, int i) {} // Violation, should be after 'foo(int i, String s)'
```
rnveach, line 65: There is no `foo(int i, String s)` listed. Please review your examples and ensure they are valid and make sense.

rnveach, line 65: Can we use 1 set of code between all "bad" examples. It will show default behavior, what its like with modifiers not set, and then show the difference when modifiers is set. It will show evolution and configuration that can be done with the different options. Each example will obviously show different order of "good" examples based on configuration.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -34,31 +41,74 @@
...
+ * public void notFoo() {} // OK
  * </pre>
  * <p>
- * Example of incorrect grouping of overloaded methods:
+ * To configure the check to group static and private methods separately within their own
+ * individual groups
+ * </p>
+ * <pre>
+ * &lt;module name="OverloadMethodsDeclarationOrder"&gt;
+ *   &lt;property name="modifierGroups" value="static, private"&gt;&lt;/property&gt;
```
rnveach, line 83: Look at examples from main issue. One has a complex regexp.
> "^public .*, (protected|package), private, ^static$"

Add something like this to show it can handle more than just single, simple modifiers.

rnveach, line 83: In this example public isn't a modifier set in the options.

Does this mean for `public` methods they can be in any order compared to private/static? What about if a public is between 2 private/static, can it split the 2 even though its not defined? We need some verbage and example showing what it means when a modifier does not show here.

If it makes sense, it could be a notes.
https://checkstyle.org/config_blocks.html#EmptyCatchBlock_Notes

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -126,27 +210,139 @@ public void visitToken(DetailAST ast) {
...
 
+    /**
+     * Get the first matching {@code modifierGroup} for the modifiers on the method
+     * and combine it with the method's name (separated by {@code SEPARATOR}) to create
+     * a unique identifier for the pair.
+     *
+     * @param currentToken assumed to be a method
+     * @return string that uniquely represents the {@code modifierGroup} and method name
+     */
+    private String getMethodDescriptor(DetailAST currentToken) {
```
rnveach, line 242: Split this into 2 methods:
`findModifierGroupPattern` - it just returns the matched pattern or the default.
I am still looking over the main method for naming and coding.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -126,27 +210,139 @@ public void visitToken(DetailAST ast) {
...
+        final String methodName = currentToken.findFirstToken(TokenTypes.IDENT).getText();
+        String descriptor = DEFAULT_MODIFIER_GROUP_PATTERN + SEPARATOR + methodName;
```
rnveach, line 244: Is there anyway we can remove the complexity of this by combining 2 things? I am just on the fence about this and the usage of a map, but we can't use a custom object or we will have to implement the hashcode method.

1 thought is instead of using the full pattern text, which is just used to show uniqueness, we could just use the pattern index and use -1 or length for the default. No one will see this text regardless, so it can be internal and shorter.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
austinarbor: Yes, the modifier groups only specify what methods must be grouped together, where they appear in relation to each other does not matter as long as they are grouped correctly. In this example, public must appear next to any other methods that are neither static nor private.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
austinarbor: I like the index idea. I can return index from the method and create the key string in the calling method

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
austinarbor: I think if we only return index splitting into two methods is no longer required? If you still want two methods, what would the second method do?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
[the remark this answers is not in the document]
romani: >  How do we determine what order non-explicit modifiers are in?

after reread of this issue, I started to think that original ideas to make order of modifiers matter might be is not good. I might originally proposed it to keep it simple. But it looks like it does not simplify logic. So I am ok to switch from regexp to simply collection/set of modifiers. It will make this module more independent from ModifierOrder module, that is good.

rnveach:
@romani I am assigning you to clear up questions at https://github.com/checkstyle/checkstyle/pull/12086#discussion_r1135948950 on the requirements

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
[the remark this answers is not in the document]
romani: @austinarbor , yes.
hint: better to not use string values, and probably use some enum for this, but your will see that make more sense during implementation.

romani:
@rnveach , PR is back to you, I answered in review thread.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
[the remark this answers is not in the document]
romani: My thoughts:
we need to focus on explicit modifiers, as users not always know implicit rules.
Method without modifiers will be separate group. Methods are usually in class or interface, so same rules of implicit will be and all of them will be in separate group.l, as even without modifiers Check still demands grouping.
Order of groups is not defined, this is good for us now.
User will need to unify modifiers to keep code clean.
If live show us good examples of why implicit matter, we can extend later on.

-----

Let's do only explicit modifiers consideration.

romani:
I replied, let me know what else is required to continue this awesome engagement to this Check

romani:
@austinarbor, ping

austinarbor:
@romani apologies, the new requirements require a lot of changes to current code/tests and I haven't had much free time to work on this. will try and get back to it soon

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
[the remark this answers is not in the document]
austinarbor: With this approach how do you group together methods without any modifiers (i.e. package)?

romani:
We lost connection to author of this PR.
Everyone is welcome to open new PR to reuse this code and finish update.

PR #12088, opened 2022-08-22
minor: add helper method hasModifiers and hasNotModifiers

ZuevKirill95:
added utility methods - hasModifiers and hasNotModifiers

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
@@ -353,4 +353,33 @@ public final class TokenUtil {
                 .collect(BitSet::new, BitSet::set, BitSet::or);
     }
 
+    /**
+     * Checks whether the specified type has all the specified modifiers.
+     *
+     * @param ast the AST node to check
+     * @param modifiersTypes int array of all modifiers types to check.
+     * @return true if AST node has all specified modifiers.
+     */
+    public static boolean hasModifiers(DetailAST ast, int... modifiersTypes) {
+        final DetailAST modifiers = ast.findFirstToken(TokenTypes.MODIFIERS);
+
+        return modifiers != null
+                && Arrays.stream(modifiersTypes)
+                .allMatch(modifier -> modifiers.findFirstToken(modifier) != null);
+    }
+
+    /**
+     * Checks whether the specified type has not all the specified modifiers.
+     *
+     * @param ast the AST node to check
+     * @param modifiersTypes int array of all modifiers types to check.
+     * @return true if AST node has not all specified modifiers.
+     */
+    public static boolean hasNotModifiers(DetailAST ast, int... modifiersTypes) {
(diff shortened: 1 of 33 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -353,4 +353,33 @@ public static BitSet asBitSet(String... tokens) {
...
     }
 
+    /**
+     * Checks whether the specified type has all the specified modifiers.
+     *
+     * @param ast the AST node to check
+     * @param modifiersTypes int array of all modifiers types to check.
+     * @return true if AST node has all specified modifiers.
+     */
+    public static boolean hasModifiers(DetailAST ast, int... modifiersTypes) {
```
rnveach, line 363: Pointing to https://github.com/checkstyle/checkstyle/issues/11527 over past concerns with performance with these type of methods.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -353,4 +353,33 @@ public static BitSet asBitSet(String... tokens) {
...
+    }
+
+    /**
+     * Checks whether the specified type has not all the specified modifiers.
+     *
+     * @param ast the AST node to check
+     * @param modifiersTypes int array of all modifiers types to check.
+     * @return true if AST node has not all specified modifiers.
+     */
+    public static boolean hasNotModifiers(DetailAST ast, int... modifiersTypes) {
```
rnveach, line 378: Same.

rnveach (commented):
Where is this change coming from? What prompted this to be started?

ZuevKirill95:
> Where is this change coming from? What prompted this to be started?

In one of the tasks I needed such a method, but I didn't find it in the util classes

And I decided to suggest such a method in the main code. If it is approved, I am ready to continue working on this PR.

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -395,4 +395,79 @@ public void testIsBooleanLiteralType() {
                 .isFalse();
     }
 
+    @Test
+    public void testAstHasModifiers() {
+        final DetailAstImpl astFinal = new DetailAstImpl();
```
nick-mancuso, line 400: We cover new code by real test input only, we are not extending our tech debt by allowing coverage by synthetic tests.

nick-mancuso (changes requested):
@ZuevKirill95 this change affects 30 checks, this is not a `minor` commit. To move forward with this change, you will need to create an issue for discussion. If issue is approved, we would need to run check regression reports for all affected checks, and ensure that there is no regressions/ changes in check behavior. 

Please create an issue, it would probably be better to close this PR and create a new PR if issue gets approved.

ZuevKirill95:
> Please create an issue, it would probably be better to close this PR and create a new PR if issue gets approved.

Okay, I'll do it. Thanks

PR #12094, opened 2022-08-23
Issue #12080: Github action Update Github milestone and upload '-all'  jar

### LINKED ISSUE #12080: Github action "Update Github"
https://github.com/checkstyle/contribution/blob/master/releasenotes-builder/diagrams-manual-trigger.md

we need to make action to do `run Github action "Update Github '10.3.1'"` (action takes human provided version as input, just to make sure it can not be executed unintentionally)
if user provider input is not matching pom - fail build

so we simply make script that update github milestones and upoad "-all.jar"
see all code in release.sh

Rahulkhinchi03:
Fixes #12080

### DIFF
--- a/.ci/update-github.sh
+++ b/.ci/update-github.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+set -e
+
+source ./.ci/util.sh
+
+checkForVariable "GITHUB_TOKEN"
+
+if [[ -z $1 ]]; then
+  echo "version is not set"
+  echo "Usage: update-github.sh <version>"
+  exit 1
+fi
+
+TARGET_VERSION=$1
+echo TARGET_VERSION="$TARGET_VERSION"
+
+CURRENT_VERSION=$(xmlstarlet sel -N pom=http://maven.apache.org/POM/4.0.0 \
+                             -t -m pom:project -v pom:version pom.xml | sed "s/-SNAPSHOT//")
+echo CURRENT_VERSION="$CURRENT_VERSION"
+
+if [ "$TARGET_VERSION" != "$CURRENT_VERSION" ]; then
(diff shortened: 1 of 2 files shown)


Rahulkhinchi03:
@romani How can I test this? As the script needs an input of the latest version because it will check latest and target. So if I do 10.3.1 then it gives error and if I use 10.3.3 then it doesn't run github tag page script.

.ci/update-github.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -e
+
+checkForVariable "GITHUB_TOKEN"
```
romani, line 5: Did you run this ?
You forgot source of util.sh

.github/workflows/update_github.yml
```diff
@@ -0,0 +1,19 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Target Version without (-SNAPSHOT)'
+        required: true
+jobs:
+  github:
+    name: Update Github ${{ github.event.inputs.version }}
+    runs-on: ubuntu-latest
```
romani, line 15: >CURRENT_VERSION=$(xmlstarlet

Please install xmlstartlet

romani (changes requested):
Items

.ci/update-github.sh
Rahulkhinchi03: Did, Maybe I unintentionally removed it while pushing.  Done Changes

.github/workflows/update_github.yml
Rahulkhinchi03: Done

romani (approved):
ok to merge

romani (approved):
ok to merge

PR #12098, opened 2022-08-24
Issue #12100: Review usages of BeforeAll in tests 

### LINKED ISSUE #12100: Review usages of `BeforeAll` in tests
We have discovered that `BeforeAll` used in conjunction with production static field modifications has allowed some surviving mutations that are false positives. From https://github.com/hcoles/pitest/issues/1081#issuecomment-1225447257:

> From a quick scan through the code I think this is "working as expected", as the test suite has hit a wrinkle with how static state works with coverage based mutation systems.

> Because this code is only run if there is no entry in the static BUNDLE_CACHE, the result of mutation testing will be highly dependent on the order in which tests are run. Adding or removing a test may completely change the order.

> If 20 tests could execute the line of code, only the first one to run (and the first ones to run after each time the cache is cleared) will actually execute the code. Only these tests will be linked by coverage to the mutation, so pitest will only select from this reduced set when the mutant is active.

In #12053 , we worked around this issue by using `BeforeEach`:

>@ZuevKirill95 Strangely using `@BeforeEach` instead of `@BeforeAll` kills the mutation, the issue for the same has been opened at https://github.com/hcoles/pitest/issues/1081.

>For now, you can use `@BeforeEach` to kill the mutation.

We should review all usages of `BeforeAll` to see if we have surviving mutations from this issue, and see if we can use `BeforeEach` instead. We should also consider forbidding usage of `BeforeAll` based on our findings, and probably at least restrict future usage of it.

```bash
➜  checkstyle git:(master) ag -Q "@BeforeAll"      
src/test/java/com/puppycrawl/tools/checkstyle/internal/CommitValidationTest.java
97:    @BeforeAll

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
148:    @BeforeAll

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
35:    @BeforeAll

src/test/java/com/puppycrawl/tools/checkstyle/meta/XmlMetaReaderTest.java
37:    @BeforeAll

src/test/java/com/puppycrawl/tools/checkstyle/AbstractGuiTestSupport.java
38:    @BeforeAll

src/test/java/com/puppycrawl/tools/checkstyle/XpathFileGeneratorAuditListenerTest.java
67:    @BeforeAll

src/it/java/com/google/checkstyle/test/chapter5naming/rule528typevariablenames/MethodTypeParameterNameTest.java
41:    @BeforeAll


```

Vyom-Yadav:
Resolves #12100

### DIFF
--- a/src/it/java/com/google/checkstyle/test/chapter5naming/rule528typevariablenames/MethodTypeParameterNameTest.java
+++ b/src/it/java/com/google/checkstyle/test/chapter5naming/rule528typevariablenames/MethodTypeParameterNameTest.java
@@ -21,31 +21,24 @@ package com.google.checkstyle.test.chapter5naming.rule528typevariablenames;
 
 import java.util.Map;
 
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
 import com.google.checkstyle.test.base.AbstractGoogleModuleTestSupport;
-import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
 import com.puppycrawl.tools.checkstyle.api.Configuration;
 
 public class MethodTypeParameterNameTest extends AbstractGoogleModuleTestSupport {
 
     private static final String MSG_KEY = "name.invalidPattern";
-    private static String format;
 
     @Override
     protected String getPackageLocation() {
         return "com/google/checkstyle/test/chapter5naming/rule528typevariablenames";
     }
 
-    @BeforeAll
-    public static void setConfigurationBuilder() throws CheckstyleException {
-        format = getModuleConfig("ClassTypeParameterName").getProperty("format");
-    }
-
     @Test
     public void testMethodDefault() throws Exception {
         final Configuration checkConfig = getModuleConfig("MethodTypeParameterName");
+        final String format = checkConfig.getProperty("format");
         final Map<String, String> messages = checkConfig.getMessages();
 
         final String[] expected = {
--- a/src/test/java/com/puppycrawl/tools/checkstyle/XpathFileGeneratorAuditListenerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/XpathFileGeneratorAuditListenerTest.java
@@ -26,7 +26,6 @@ import java.io.File;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
 import com.puppycrawl.tools.checkstyle.api.AuditEvent;
@@ -64,8 +63,16 @@ public class XpathFileGeneratorAuditListenerTest {
     private final CloseAndFlushTestByteArrayOutputStream outStream =
             new CloseAndFlushTestByteArrayOutputStream();
 
-    @BeforeAll
-    public static void constructEvents() throws Exception {
+    static {
+        try {
+            constructEvents();
+        }
+        catch (Exception ex) {
+            throw new ExceptionInInitializerError(ex);
+        }
+    }
+
+    private static void constructEvents() throws Exception {
         final TreeWalkerAuditEvent event1 = createTreeWalkerAuditEvent(
                 "InputXpathFileGeneratorAuditListener.java", FIRST_MESSAGE);
 
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/CommitValidationTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/CommitValidationTest.java
@@ -41,7 +41,6 @@ import org.eclipse.jgit.lib.Repository;
 import org.eclipse.jgit.revwalk.RevCommit;
 import org.eclipse.jgit.revwalk.RevWalk;
 import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
 /**
@@ -92,15 +91,10 @@ public class CommitValidationTest {
     private static final CommitsResolutionMode COMMITS_RESOLUTION_MODE =
             CommitsResolutionMode.BY_LAST_COMMIT_AUTHOR;
 
-    private static List<RevCommit> lastCommits;
-
-    @BeforeAll
-    public static void setUp() throws Exception {
-        lastCommits = getCommitsToCheck();
-    }
-
     @Test
-    public void testHasCommits() {
+    public void testHasCommits() throws Exception {
+        final List<RevCommit> lastCommits = getCommitsToCheck();
+
         assertWithMessage("must have at least one commit to validate")
                 .that(lastCommits != null && !lastCommits.isEmpty())
                 .isTrue();
@@ -176,7 +170,8 @@ public class CommitValidationTest {
     }
 
     @Test
-    public void testCommitMessageHasProperStructure() {
+    public void testCommitMessageHasProperStructure() throws Exception {
+        final List<RevCommit> lastCommits = getCommitsToCheck();
         for (RevCommit commit : filterValidCommits(lastCommits)) {
             final String commitMessage = commit.getFullMessage();
             final int error = validateCommitMessage(commitMessage);
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
@@ -30,11 +30,11 @@ import java.util.HashSet;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
@@ -138,22 +138,32 @@ public class XpathRegressionTest extends AbstractModuleTestSupport {
             "SuppressWarningsHolder"
     );
 
-    private static Set<String> simpleCheckNames;
-    private static Map<String, String> allowedDirectoryAndChecks;
-    private static Set<String> internalModules;
+    private static final Set<String> SIMPLE_CHECK_NAMES = getSimpleCheckNames();
+    private static final Map<String, String> ALLOWED_DIRECTORY_AND_CHECKS =
+        getAllowedDirectoryAndChecks();
+
+    private static final Set<String> INTERNAL_MODULES = getInternalModules();
 
     private Path javaDir;
     private Path inputDir;
 
-    @BeforeAll
-    public static void setUpBeforeClass() throws IOException {
-        simpleCheckNames = CheckUtil.getSimpleNames(CheckUtil.getCheckstyleChecks());
+    private static Set<String> getSimpleCheckNames() {
+        try {
+            return CheckUtil.getSimpleNames(CheckUtil.getCheckstyleChecks());
+        }
+        catch (IOException ex) {
+            throw new ExceptionInInitializerError(ex);
+        }
+    }
 
-        allowedDirectoryAndChecks = simpleCheckNames
-                .stream()
-                .collect(Collectors.toMap(id -> id.toLowerCase(Locale.ENGLISH), id -> id));
+    private static Map<String, String> getAllowedDirectoryAndChecks() {
+        return SIMPLE_CHECK_NAMES
+            .stream()
+            .collect(Collectors.toMap(id -> id.toLowerCase(Locale.ENGLISH), Function.identity()));
+    }
 
-        internalModules = Definitions.INTERNAL_MODULES.stream()
+    private static Set<String> getInternalModules() {
+        return Definitions.INTERNAL_MODULES.stream()
             .map(moduleName -> {
                 final String[] packageTokens = moduleName.split("\\.");
                 return packageTokens[packageTokens.length - 1];
@@ -188,7 +198,7 @@ public class XpathRegressionTest extends AbstractModuleTestSupport {
         abstractJavadocCheckNames.addAll(REGEXP_JAVADOC_CHECKS);
         final Set<String> abstractJavadocCheckSimpleNames =
                 CheckUtil.getSimpleNames(abstractJavadocCheckNames);
-        abstractJavadocCheckSimpleNames.removeAll(internalModules);
+        abstractJavadocCheckSimpleNames.removeAll(INTERNAL_MODULES);
         assertWithMessage("INCOMPATIBLE_JAVADOC_CHECK_NAMES should contains all descendants "
                     + "of AbstractJavadocCheck")
             .that(abstractJavadocCheckSimpleNames)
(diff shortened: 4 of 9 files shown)


nick-mancuso:
@Vyom-Yadav please see CI failures, and we can now do this PR in scope of #12100

src/it/java/com/google/checkstyle/test/chapter5naming/rule528typevariablenames/MethodTypeParameterNameTest.java
```diff
@@ -38,7 +38,7 @@ protected String getPackageLocation() {
         return "com/google/checkstyle/test/chapter5naming/rule528typevariablenames";
     }
 
-    @BeforeAll
+    @BeforeEach
     public static void setConfigurationBuilder() throws CheckstyleException {
```
rnveach, line 42: BeforeEach cannot be static. I am not sure if pitest is actually running the method because of this but it is causing a CI failure.

src/test/java/com/puppycrawl/tools/checkstyle/internal/CommitValidationTest.java
```diff
@@ -94,7 +94,7 @@ public class CommitValidationTest {
 
     private static List<RevCommit> lastCommits;
 
-    @BeforeAll
+    @BeforeEach
     public static void setUp() throws Exception {
         lastCommits = getCommitsToCheck();
```
rnveach, line 99: This should not be a before each. This is basically a caching method and is not even part of the pitest suite.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
```diff
@@ -145,7 +144,7 @@ public class XpathRegressionTest extends AbstractModuleTestSupport {
     private Path javaDir;
     private Path inputDir;
 
-    @BeforeAll
+    @BeforeEach
     public static void setUpBeforeClass() throws IOException {
         simpleCheckNames = CheckUtil.getSimpleNames(CheckUtil.getCheckstyleChecks());
```
rnveach, line 149: Same.

rnveach (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/XpathFileGeneratorAuditListenerTest.java
```diff
@@ -64,7 +64,7 @@ public class XpathFileGeneratorAuditListenerTest {
     private final CloseAndFlushTestByteArrayOutputStream outStream =
             new CloseAndFlushTestByteArrayOutputStream();
 
-    @BeforeAll
+    @BeforeEach
     public static void constructEvents() throws Exception {
```
rnveach, line 68: This before is very questionable. It is not modifying any fields and is not doing any assertions. However, removing it results in test failures so obviously there is a pass by reference going on here.

`XpathFileGeneratorAstFilter#accept` is basically populated `XpathFileGeneratorAstFilter#MESSAGE_QUERY_MAP` with pre-filled values to be read in the actual tests.

It is possible to have random test order issues if a new test was added that erased this cache by mistake.

src/it/java/com/google/checkstyle/test/chapter5naming/rule528typevariablenames/MethodTypeParameterNameTest.java
```diff
@@ -38,7 +38,7 @@ protected String getPackageLocation() {
...
         format = getModuleConfig("ClassTypeParameterName").getProperty("format");
```
rnveach, line 43: I feel this before can just be completely removed and inlined.

Vyom-Yadav: It throws an exception, better to be instantiated in the method.

src/test/java/com/puppycrawl/tools/checkstyle/XpathFileGeneratorAuditListenerTest.java
Vyom-Yadav: Reverted back to `@BeforeAll`

src/it/java/com/google/checkstyle/test/chapter5naming/rule528typevariablenames/MethodTypeParameterNameTest.java
rnveach: What is the exception?

src/test/java/com/puppycrawl/tools/checkstyle/XpathFileGeneratorAuditListenerTest.java
rnveach: IMO, a different way should be done instead of this. Like I said, this cause result in random test order issues which will affect pitest.

Vyom-Yadav: Initialized them in a different way.

src/it/java/com/google/checkstyle/test/chapter5naming/rule528typevariablenames/MethodTypeParameterNameTest.java
```diff
@@ -31,18 +31,22 @@
 public class MethodTypeParameterNameTest extends AbstractGoogleModuleTestSupport {
 
     private static final String MSG_KEY = "name.invalidPattern";
-    private static String format;
+    private static final String format = getFormat();
```
nrmancuso, line 34: Why does this have to be a field? Can we just do:
```java
        final String format = checkConfig.getProperty("format");
```
in each unit test? I would rather throw an exception as part of the unit test than when we are initializing them.

src/test/java/com/puppycrawl/tools/checkstyle/internal/CommitValidationTest.java
```diff
@@ -92,11 +92,15 @@ public class CommitValidationTest {
     private static final CommitsResolutionMode COMMITS_RESOLUTION_MODE =
             CommitsResolutionMode.BY_LAST_COMMIT_AUTHOR;
 
-    private static List<RevCommit> lastCommits;
+    private static final List<RevCommit> lastCommits = getLastCommits();
 
-    @BeforeAll
-    public static void setUp() throws Exception {
-        lastCommits = getCommitsToCheck();
+    public static List<RevCommit> getLastCommits() {
```
nrmancuso, line 97: Two questions:
1. Why is this method public?
2. Can we just get last commits in unit test methods?

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
```diff
@@ -138,22 +138,32 @@ public class XpathRegressionTest extends AbstractModuleTestSupport {
...
     );
 
-    private static Set<String> simpleCheckNames;
-    private static Map<String, String> allowedDirectoryAndChecks;
-    private static Set<String> internalModules;
+    private static final Set<String> simpleCheckNames = getSimpleCheckNames();
+    private static final Map<String, String> allowedDirectoryAndChecks =
+        getAllowedDirectoryAndChecks();
+
+    private static final Set<String> internalModules = getInternalModules();
```
nrmancuso, line 145: Can we move initialization of these into the `BeforeEach` below (as well as content of methods that initialize them)?

nrmancuso (changes requested):
Items:

src/it/java/com/google/checkstyle/test/chapter5naming/rule528typevariablenames/MethodTypeParameterNameTest.java
rnveach: https://github.com/checkstyle/checkstyle/blob/399f1b48b79731d65e824eb927971ef6174c3463/src/it/java/com/google/checkstyle/test/chapter5naming/rule51identifiernames/CatchParameterNameTest.java#L40

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
Vyom-Yadav: We can, but why? Initializing `internalModules` doesn't throw an exception and initializing `simpleCheckNames` and `allowedDirectoryAndChecks` initially shouldn't cause any problem.

rnveach: If this just initializes the values once, and they never change during test runs, then it doesn't make sense to run them in `BeforeEach` as it will just generate the same lists every time. `BeforeEach` should really be for areas that need to be reset or prepare for each run.

Vyom-Yadav: They don't change during the tests.

nrmancuso: > If this just initializes the values once, and they never change during test runs, then it doesn't make sense to run them in `BeforeEach` as it will just generate the same lists every time. `BeforeEach` should really be for areas that need to be reset or prepare for each run.

Agreed, let’s not do this.

src/it/java/com/google/checkstyle/test/chapter5naming/rule528typevariablenames/MethodTypeParameterNameTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/CommitValidationTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
Vyom-Yadav: `PMD Failure: com.puppycrawl.tools.checkstyle.internal.XpathRegressionTest:154 Rule:AvoidThrowingRawExceptionTypes Priority:1 Avoid throwing raw exception types..`

Should I create a custom exception?

nrmancuso: No, please avoid custom exceptions at all costs.

Vyom-Yadav: I added it to the suppression list.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
```diff
@@ -138,22 +137,32 @@ public class XpathRegressionTest extends AbstractModuleTestSupport {
...
+        }
+    }
 
-        allowedDirectoryAndChecks = simpleCheckNames
-                .stream()
-                .collect(Collectors.toMap(id -> id.toLowerCase(Locale.ENGLISH), id -> id));
+    private static Map<String, String> getAllowedDirectoryAndChecks() {
+        return SIMPLE_CHECK_NAMES
+            .stream()
+            .collect(Collectors.toMap(id -> id.toLowerCase(Locale.ENGLISH), id -> id));
```
nrmancuso, line 161: `id -> id` can we use identity function here?

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
```diff
@@ -138,22 +137,32 @@ public class XpathRegressionTest extends AbstractModuleTestSupport {
...
 
-    @BeforeAll
-    public static void setUpBeforeClass() throws IOException {
-        simpleCheckNames = CheckUtil.getSimpleNames(CheckUtil.getCheckstyleChecks());
+    private static Set<String> getSimpleCheckNames() {
+        try {
+            return CheckUtil.getSimpleNames(CheckUtil.getCheckstyleChecks());
+        }
+        catch (IOException ex) {
+            throw new RuntimeException(ex);
```
nrmancuso, line 154: Please use `ExceptionInInitializerError`

nrmancuso (changes requested):
Items:

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
Vyom-Yadav: Done.

nrmancuso (commented):
I am not sure if we are ready to forbid `BeforeAll`, but it seems like remaining usages are probably unnecessary. If we turn off these loggers for tests, we should probably do this in the pom or some property file (i.e. globally); especially since one of these is for ArchUnit and we shouldn't have to turn off this property in each ArchUnit test (and we haven't been). I am curious what others think. @romani @rnveach?

rnveach:
We are turning off the loggers for this test, never turning them back on, and it is most likely leaking into other tests and not guaranteed that we are turning them off for all tests (depending on junit's random order). I feel this isn't safe and could cause some weird issue like what we have seen for pitest testing, so something should be done.

If we can disable these loggers for all tests like @nrmancuso said, that would be the best.

romani:
> We have discovered that BeforeAll used in conjunction with production static field modifications has allowed some surviving mutations that are false positives

I confused, I see no updates in survivals. What is benefiting from this test update.
I ok with update, but goal is strange.

Vyom-Yadav:
> I confused, I see no updates in survivals. What is benefiting from this test update.
I ok with update, but goal is strange.

Yes, we did not kill any mutations, but it could have been in-general problematic like rnveach has mentioned in the issue comment.

romani (approved):
Ok to merge

romani:
Lets not finish,l annotation, we will try to recheck this during code review.

rnveach (changes requested):
https://github.com/checkstyle/checkstyle/pull/12098#discussion_r955499871
This item is still left from my original review.

https://github.com/checkstyle/checkstyle/pull/12098#pullrequestreview-1103140612
Can we do this?

rnveach (commented):
After talking with @romani , we are going to delay this PR until GSoC clears up since it no longer is looking related to GSoC or pitest and allow student to focus on more GSoC like Checker.

nrmancuso:
@rnveach @Vyom-Yadav can we proceed?

rnveach:
if @Vyom-Yadav 's time is scarce, I would like to finish resolved Checker to keep on 1 target.

romani:
@Vyom-Yadav, please rebase.

rnveach:
I am still waiting for my previous review items to be finished.
https://github.com/checkstyle/checkstyle/pull/12098#pullrequestreview-1103171646

src/test/java/com/puppycrawl/tools/checkstyle/XpathFileGeneratorAuditListenerTest.java
Vyom-Yadav: Given the design of `XpathFileGeneratorAstFilter`, the only option is to populate the map in a static block, changed to that.

rnveach (approved):

nrmancuso (approved):
@Vyom-Yadav thanks a lot for wrapping this up

PR #12102, opened 2022-08-25
Remove custom cache in Violation.java. It was extra cache for resource bundle cache for messages.

rnveach:
Identified at https://github.com/hcoles/pitest/issues/1081#issuecomment-1226105481
while looking at https://github.com/checkstyle/checkstyle/pull/12053

`java.util.ResourceBundle.getBundle` has its own cache. It's cache is in the form `CacheKey(baseName, locale, module, callerModule)`. So extra cache in our code for this looks like not required.
The double cache has been there since 2002. The commit, 7e083cf6cbbe2b1590b92bf3feae547e5f58d6ac, says it is a 8% performance improvement.

While looking into this, bug was identified and posted at https://github.com/checkstyle/checkstyle/issues/12101 .

`Utf8Control` was made public in 7c2f448ad61f66224168d197dfc6195e7b8d9409 for no reason than for tests and not production code.

`DefaultLoggerTest#testCountryIsValid` and `DefaultLoggerTest#testLanguageIsValid` has no connection to `DefaultLogger`.

`Violation.clearCache()` is removed.

PR is currently the first stage of changes and I am going to add more changes:
Stage 2 is at https://github.com/checkstyle/checkstyle/pull/12105 and waiting for this to be merged.
Stage 3 is planned to be https://github.com/checkstyle/checkstyle/issues/12104 .
Stage 4 was probably going to be some other left over things, possibly things 2 did not finish.
All of this is to basically assist with resolving issues found at https://github.com/checkstyle/checkstyle/pull/12053 .

Issue #12101, opened 2022-08-25
Checker.localeLanguage property work strange with system property user.language 


rnveach:
https://checkstyle.org/config.html#Checker_Properties

````
$ cat TestClass.java
public class TestABbreviation {}

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name = "Checker">
  <property name="charset" value="UTF-8"/>
  <property name="localeLanguage" value="fr"/> <!-- ATTENTION HERE -->

  <property name="severity" value="warning"/>

  <property name="fileExtensions" value="java, properties, xml"/>

  <module name="TreeWalker">
    <module name="AbbreviationAsWordInName">
      <property name="ignoreFinal" value="false"/>
      <property name="allowedAbbreviationLength" value="0"/>
      <property name="allowedAbbreviations" value="AB"/>
      <property name="tokens"
               value="CLASS_DEF, INTERFACE_DEF, ENUM_DEF, ANNOTATION_DEF, ANNOTATION_FIELD_DEF,
                    PARAMETER_DEF, VARIABLE_DEF, METHOD_DEF, PATTERN_VARIABLE_DEF, RECORD_DEF,
          RECORD_COMPONENT_DEF"/>
    </module>
  </module>
</module>

$ java -jar checkstyle-10.3.2-all.jar -c TestConfig.xml TestClass.java
Starting audit...
[WARN] M:\\Test.java:1:14: L'abr├⌐viation dans le nom 'TestABbreviation' ne doit pas 
      contenir plus de '1' lettres majuscules. [AbbreviationAsWordInName]
Audit done.

$ java -D"user.language=fr" -jar .\checkstyle-10.3.2-all.jar -c Test.xml Test.java
D├⌐but de la v├⌐rification...
[WARN] M:\\Test.java:1:14: L'abr├⌐viation dans le nom 'TestABbreviation' ne doit pas 
     contenir plus de '1' lettres majuscules. [AbbreviationAsWordInName]
V├⌐rification termin├⌐e.
````
Starting audit doesn't obey config locale while check messages do.
I expect both outputs to be exactly the same since the configuration file specifies the language, but it turns out starting/finishing audits is in the OS language which for me is English.

Issue is DefaultLogger is only using the default locale
[...]

PR #12105, opened 2022-08-26
Issue #12101: move global locale to localized message 

### LINKED ISSUE #12104: Remove Global Static Locale
https://github.com/checkstyle/checkstyle/blob/a368230b7df8ca70839f97d30a8d8adebd8b9ff3/src/main/java/com/puppycrawl/tools/checkstyle/api/Violation.java#L68
https://github.com/checkstyle/checkstyle/blob/a368230b7df8ca70839f97d30a8d8adebd8b9ff3/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L434

Identified in several issues,
https://github.com/checkstyle/checkstyle/pull/12053#discussion_r955646247
https://github.com/checkstyle/checkstyle/issues/12101
https://github.com/checkstyle/checkstyle/issues/7436

Setting any type of locale in a global area, either by `Violation.setLocale` or `Locale.setDefault`, is a very bad idea.

It is very easy to corrupt the locale if developers creating tests aren't careful as changing this value can impact all other tests since it is global. Even if developers are careful, its hard to remember all the little places that can change the locale without in-depth knowledge all the time.

Example:
https://github.com/checkstyle/checkstyle/blob/d287d78e7ee1218092f0c386cddd78ea81448fc3/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java#L478 -- We set the locale to italy and don't do any reversion. I am not sure how this hasn't been an issue already but I think we have been lucky on test run ordering.
https://github.com/checkstyle/checkstyle/blob/0eadd97b8af5b79e7c49f82958fbb3d863bfdf61/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java#L286 -- We set the locale to English but CI and most users run in English.
https://github.com/checkstyle/checkstyle/blob/d287d78e7ee1218092f0c386cddd78ea81448fc3/src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java#L445 -- Example of where we have to be extra careful and reset the global locale back after testing.

If we ever create a config with a locale set and load it in a test, it would also cause a global issue.

We need to remove this method somehow and pass the locale either by stack or have a central, non-static, area be the main source of it like Checker.

We also need to forbid `Locale.setDefault`. (either ForbiddenAPIs or xpath)

rnveach:
Issue #12101

The issue is our exceptions use LocalizedMessage though it does not have access to the global locale. It is expecting to be supplied the locale, but we never provide it the overridden locale from the config.

As the class name implies, LocalizedMessage should **currently** be the central place for the locale storage. (This does not resolve #12104 but does proceed to better management). The locale does not need to be provided to `LocalizedMessage` in the constructor anymore as anyone using should use the global locale, **currently**.

The second commit converts a bunch of `Violation` uses to `LocalizedMessage`. The areas being used are not actual violations but just being used for locale messages. Violation and LocalizedMessage both use the global locale, so this is really a no change. Any fields dropped when passing to the constructor were not being used anyways. So we basically get the same effect, with less unnecessary coding.

This is built on top of https://github.com/checkstyle/checkstyle/pull/12102 and thus this PR is blocked until the other is merged, hence why this is drafted.

### DIFF
--- a/.ci/pitest-suppressions/pitest-misc-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-misc-suppressions.xml
@@ -209,15 +209,6 @@
 
   <mutation unstable="false">
     <sourceFile>TranslationCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.TranslationCheck</mutatedClass>
-    <mutatedMethod>validateUserSpecifiedLanguageCodes</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/checks/TranslationCheck::getId</description>
-    <lineContent>WRONG_LANGUAGE_CODE_KEY, new Object[] {code}, getId(), getClass(), null);</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>TranslationCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.TranslationCheck$ResourceBundle</mutatedClass>
     <mutatedMethod>getFiles</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator</mutator>
--- a/.ci/pitest-suppressions/pitest-tree-walker-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-tree-walker-suppressions.xml
@@ -3,15 +3,6 @@
   <mutation unstable="false">
     <sourceFile>DetailNodeTreeStringPrinter.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.DetailNodeTreeStringPrinter</mutatedClass>
-    <mutatedMethod>getParseErrorMessage</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/JavadocDetailNodeParser$ParseErrorMessage::getLineNumber</description>
-    <lineContent>parseErrorMessage.getLineNumber(),</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>DetailNodeTreeStringPrinter.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.DetailNodeTreeStringPrinter</mutatedClass>
     <mutatedMethod>parseFile</mutatedMethod>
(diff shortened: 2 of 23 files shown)


.ci/pitest-suppressions/pitest-misc-suppressions.xml
```diff
@@ -207,15 +207,6 @@
     <lineContent>getId(),</lineContent>
   </mutation>
 
-  <mutation unstable="false">
-    <sourceFile>TranslationCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.TranslationCheck</mutatedClass>
-    <mutatedMethod>validateUserSpecifiedLanguageCodes</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/checks/TranslationCheck::getId</description>
-    <lineContent>WRONG_LANGUAGE_CODE_KEY, new Object[] {code}, getId(), getClass(), null);</lineContent>
```
rnveach, line 216: > Any fields dropped when passing to the constructor were not being used anyways. So we basically get the same effect, with less unnecessary coding.

This shows what I am talking about.

`getId` is never used as part of the exception message as we just provide the localized text of the message and the id is not included in it. ID is needed for violations when printing audit events only, which this is never used for.

.ci/pitest-suppressions/pitest-tree-walker-suppressions.xml
```diff
@@ -1,14 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <mutation unstable="false">
-    <sourceFile>DetailNodeTreeStringPrinter.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.DetailNodeTreeStringPrinter</mutatedClass>
-    <mutatedMethod>getParseErrorMessage</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/JavadocDetailNodeParser$ParseErrorMessage::getLineNumber</description>
-    <lineContent>parseErrorMessage.getLineNumber(),</lineContent>
```
rnveach, line 9: Same.

rnveach:
Local testing works:
````
PS M:\> java -jar .\checkstyle-10.3.3-SNAPSHOT-all.jar -c Test.xml Test.java
D├⌐but de la v├⌐rification...
[WARN] M:\\Test.java:1:14: L'abr├⌐viation dans le nom 'TestABbreviation' ne doit pas contenir plus de '1' lettres majuscules. [AbbreviationAsWordInName]
V├⌐rification termin├⌐e.
````

romani (approved):
ok to merge.

it is breaking changes for plugins, but I am ok.

romani:
GitHub, rebase

nrmancuso (approved):

PR #12110, opened 2022-08-27
Issue #12082: Github action Publish release notes Twitter 

### LINKED ISSUE #12082: Github action "Publish release notes outside"
https://github.com/checkstyle/contribution/blob/master/releasenotes-builder/diagrams-manual-trigger.md

we need to make action to do `run Github action "Publish release notes outside '10.3.1'"`  (action takes human provided version as input, just to make sure it can not be executed unintentionally)
if user provider input is not matching pom - fail build

so we simply make/reuse  scripts that does twit (need to know howto share secrets)
and upadate release notes to Github (existing script)

Rahulkhinchi03:
Fixes #12082 

To use Twitter variables as secrets: https://docs.github.com/en/actions/security-guides/encrypted-secrets

### DIFF
--- a/.github/workflows/publish_releasenotes_for_twitter.yml
+++ b/.github/workflows/publish_releasenotes_for_twitter.yml
@@ -1,0 +1,21 @@
+#############################################################################
+# GitHub Action to Publish Tweet of release notes.
+#
+#############################################################################
+name: "Publish Tweet for Release notes"
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Target Version without (-SNAPSHOT)'
+        required: true
+jobs:
+  publish:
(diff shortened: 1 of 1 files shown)


.github/workflows/copy_of_publish_releasenotes_outside.yml
```diff
@@ -0,0 +1,23 @@
+#############################################################################
```
romani, line 1: Rename file

romani (changes requested):
Items

.github/workflows/copy_of_publish_releasenotes_outside.yml
Rahulkhinchi03: Done

romani (approved):
Ok to merge

romani (approved):
ok to merge

Issue #12129, opened 2022-08-28
no-error-pmd execution is failing if version in not changed to 10.4


romani:
right after release 10.3.2 I forgot to update back to 10.4-SNAPSHOT

resulted in build failure:
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/14472/workflows/e0cd3819-c488-4b41-af26-d5debdf521a1/jobs/164841

it became green right after bump
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/14474/workflows/873a3d0a-5015-458c-a481-c651efb3cd09/jobs/164876

problems is that pmd takes this SNAPSHOTS from maven central, it is not what we need!!! :
```
~/java/github/romani/checkstyle/.ci-temp/pmd [master|✔] 
$ ./mvnw verify --show-version --errors --batch-mode                 -DskipTests                 -Dmaven.javadoc.skip=true                 -Dmaven.source.skip=true                 -Dpmd.skip=true                 -Dcheckstyle.skip=false                 -Dcheckstyle.version="10.3.4-SNAPSHOT"
Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
Maven home: /home/rivanov/.m2/wrapper/dists/apache-maven-3.8.6-bin/1ks0nkde5v1pk9vtc31i9d0lcd/apache-maven-3.8.6
Java version: 11.0.14.1, vendor: Amazon.com Inc., runtime: /opt/jvm/amazon-corretto-11.0.14.10.1-linux-x64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.4.0-210-generic", arch: "amd64", family: "unix"
[INFO] Error stacktraces are turned on.
[INFO] Scanning for projects...
[INFO] Inspecting build with total of 39 modules...
[INFO] Installing Nexus Staging features:
[INFO]   ... total of 39 executions of maven-deploy-plugin replaced with nexus-staging-maven-plugin
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO] 
[INFO] PMD                                                                [pom]
[INFO] PMD Core                                                           [jar]
[INFO] PMD Test Schema                                                    [jar]
[INFO] PMD Test Framework                                                 [jar]
[INFO] PMD language module testing utilities                              [jar]
[INFO] PMD C++                                                            [jar]
[INFO] PMD C#                                                             [jar]
[INFO] PMD Dart                                                           [jar]
[...]

PR #12130, opened 2022-08-28
Issue #12129: build checkstyle SNAPSHOT before execution other projects validaiton

romani:
Issue #12129

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -198,6 +198,8 @@ markdownlint)
 no-error-pmd)
   CS_POM_VERSION="$(getCheckstylePomVersion)"
   echo "CS_version: ${CS_POM_VERSION}"
+  mvn -e --no-transfer-progress clean install -Pno-validations
+  echo "Checkout target sources ..."
   checkout_from "https://github.com/pmd/build-tools.git"
   cd .ci-temp/build-tools/
   mvn -e --no-transfer-progress install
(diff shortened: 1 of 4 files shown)


pbludov (approved):

rnveach (changes requested):
I thought CIs already did this?

I am seeing semaphore still doing a clean install before running validation commands, so that seems like an issue and double installs.

Wercker can be ignored since it does 1 install for all runs.

`no-exception-workflow.yml` has something similar but it calls `no-exception-test.sh`. Do we want to make similar changes here?

rnveach:
Double install
https://checkstyle.semaphoreci.com/jobs/9c21040f-4402-4a8a-9cb0-4257aaf17458#L556
https://checkstyle.semaphoreci.com/jobs/9c21040f-4402-4a8a-9cb0-4257aaf17458#L751

romani:
@rnveach , I removed mvn install for checkstyle from semaphore

rnveach:
> no-exception-workflow.yml has something similar but it calls no-exception-test.sh. Do we want to make similar changes here?

So is this a no?

romani:
Such shell file operate on diff.groovy 
https://github.com/checkstyle/checkstyle/blob/master/.ci/no-exception-test.sh

Groovy do build by itself.

rnveach:
> Groovy do build by itself.

So why do we do an install ... https://github.com/checkstyle/checkstyle/blob/master/.github/workflows/no-exception-workflow.yml#L24
Before groovy kicks in ... https://github.com/checkstyle/checkstyle/blob/master/.ci/no-exception-test.sh#L62

Isn't this another example of doing a double install?

romani:
good catch.

https://github.com/checkstyle/contribution/blob/1b3e33600898a849791dc88c90df2bb3f80400e7/checkstyle-tester/diff.groovy#L210-L219

https://github.com/checkstyle/contribution/blob/1b3e33600898a849791dc88c90df2bb3f80400e7/checkstyle-tester/diff.groovy#L28-L33

romani:
I removed install from `.github/workflows/no-exception-workflow.yml`

rnveach:
Can merge when CI passes.

romani:
Cancelled jobs are restarted

nrmancuso (approved):

Issue #11912, opened 2022-07-18
Disable test execution during release


romani:
it takes a while to run them, but if we want to release, we should be able to release.
UT failures is very rare case due to fact that we have bunch of CIs. 

Consider to not execute: other tools too, it might result in empty reports for website, but I doubt anybody need them.

We have special profile `-Pno-validations` we should use it, it will make release quicker and more stable.

Public statement of quality is more accessible in badges of GitHub readme file.
https://github.com/checkstyle/checkstyle

Quality is in details, and amount of activated rules, but HTML report does not show it, so better to not waste time on generation of them.

Rahulkhinchi03:
command:
`mvn -e --no-transfer-progress clean install -Pno-validations`. 

Please tell me, how do you do it now?

I believe I should create a shell script for the above command.

romani:
`mvn -e --no-transfer-progress clean install -Pno-validations` is enough.
I do usually
`mvn -e --no-transfer-progress clean site -Pno-validations`

Rahulkhinchi03:
Shell script with only one command?

Rahulkhinchi03:
This issue needs to be close, working on #12115 ( same task )

romani:
we ok to use `mvn -e --no-transfer-progress site -Pno-validations -Dmaven.javadoc.skip=false`

prove that it will work.
```
✔ ~/java/github/romani/checkstyle [checkstyle-10.3.3|✔] 
12:34 $ mvn -e --no-transfer-progress clean site -Pno-validations -Dmaven.javadoc.skip=false
...
 ~/java/github/romani/checkstyle [checkstyle-10.3.3|✔] 
12:22 $ meld target/site/ ~/java/github/checkstyle/checkstyle.github.io/ &
[1] 24594
```
![image](https://user-images.githubusercontent.com/812984/187091483-d007e2ce-a42e-4bcb-95d5-8f62a98f6af9.png)

nick-mancuso:
>Consider to not execute: other tools too, it might result in empty reports for website, but I doubt anybody need them.

Then why not remove this section?

romani:
I do not know how to do this :)
There might be some "leak" of tool report, I don't mind.

I just try to cut massive execution time.

PR #12131, opened 2022-08-28
Issue #11912: remove all validations during web site build during release

romani:
Issue #11912

see prove that it is ok in issue - https://github.com/checkstyle/checkstyle/issues/11912#issuecomment-1229538320

so we will not have reports in website that nobody care to look and action.
But we will significantly reduce execution time of this release phase.
1 min - without validation
8 min - with validatiion

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -45,7 +45,7 @@ echo "Go to folder where site was build and sources are already at required tag"
 cd target/checkout
 
 echo "Generating web site"
-mvn -e --no-transfer-progress site
+mvn -e --no-transfer-progress site -Pno-validations -Dmaven.javadoc.skip=false
 
 echo "Generating uber jar ...(no clean to keep site resources just in case)"
 mvn -e --no-transfer-progress -Passembly package


release.sh
```diff
@@ -45,7 +45,7 @@ echo "Go to folder where site was build and sources are already at required tag"
 cd target/checkout
 
 echo "Generating web site"
-mvn -e --no-transfer-progress site
+mvn -e --no-transfer-progress site -Pno-validations -Dmaven.javadoc.skip=false
```
pbludov, line 48: Javadoc is a part of the site: https://checkstyle.org/apidocs/index.html
The option `maven.javadoc.skip` will disable generation of API reference.

strkkk (approved):

release.sh
romani: It is skip as FALSE, please look at diff of folders, I prove that API is generated.
Content is same, just timestamp are different

pbludov (approved):

nrmancuso (approved):

rnveach (approved):
I think we should still ensure suite passes when doing the release at the start of the process. We don't have some critical need to critically release on time. We should emphasis a product over timeliness.

romani:
It is too late to care about our over care of quality in each symbol of code.
If leak happened it just should be fixed in next commit, actual user of our product are not affected by our slightly damaged code quality.

In absolutely worse case, we will do one more release.

Issue #12099, opened 2022-08-24
Expand ArchUnit Testing


nrmancuso:
While ArchUnit was introduced in https://github.com/checkstyle/checkstyle/issues/7469, we only did very basic initial integration (see https://github.com/checkstyle/checkstyle/pull/8112). We would like to expand our usage of ArchUnit.

It would be a good first step to see if we can complete the work started at https://github.com/checkstyle/checkstyle/pull/8223, without having ArchUnit add generated files to the project.

References:
https://blog.jdriven.com/2018/10/testing-the-architecture-archunit-in-practice/
https://www.mobileit.cz/Blog/Pages/arch-unit-1.aspx
https://www.mobileit.cz/Blog/Pages/arch-unit-2.aspx
https://www.mscharhag.com/software-development/architecture-validation-with-archunit

Good "alternative" uses:
https://nexocode.com/blog/posts/archunit-forget-architecture-it-s-a-flexible-and-intelligent-linter/

Let's use this issue to discuss usage ideas.

nick-mancuso:
Here might be a good candidate for ArchUnit, if we decide to pursue it:
https://github.com/checkstyle/checkstyle/issues/12100#issuecomment-1225810498

Forbidding mutable test fields.

rnveach:
I would consider a field immutable only if it's value never changes throughout the life of the test run, including the place where it got its value from. If I can get 2 different values from the same method call when executing it at the start of the test run and any time between then and the end of the test run, then the field isn't immutable.

Vyom-Yadav:
> Forbidding mutable test fields.

Using ArchUnit, we can only make sure that fields are `final`, i.e. something like-
```java
@ArchTest                                                                
public static final ArchRule NEW_RULE = fields().that()                   
    .areDeclaredInClassesThat()                                           
    .haveSimpleNameEndingWith("Test")                                     
    .should()                                                             
    .beFinal();                                                          
```

Vyom-Yadav:
> It would be a good first step to see if we can complete the work started at https://github.com/checkstyle/checkstyle/pull/8223, without having ArchUnit add generated files to the project.

The added files by ArchUnit are due to https://www.archunit.org/userguide/html/000_Index.html#_freezing_arch_rules, if we are looking to integrate it, this can act as a suppression file as it will contain all the violations, another method can be to create custom slices without some classes, ignoring those classes.

Vyom-Yadav:
[...]

PR #12135, opened 2022-08-29
Issue #12099: Add ArchUnit test to ensure fields in modules and utils are immutable, and modules are correctly annotated

Vyom-Yadav:
#12099 

### A brief about the changes in this PR:

- There are 4 `ArchRules`, ArchUnit is just like English, so the rules are easy to understand.
  - `fieldsInUtilClassesShouldBeImmutable`
  - `fieldsInStatelessChecksShouldBeImmutable`
  - `classesWithImmutableFieldsShouldBeStateless`
  - `classesWithMutableFieldsShouldBeStateful`
-  Suppression list was added and populated.

### DIFF
--- a/config/pmd-test.xml
+++ b/config/pmd-test.xml
@@ -119,8 +119,8 @@
              from the test method.
            In XdocsPagesTest PMD does not find asserts in lambdas.
            All test classes which starts with XpathRegression have asserts inside parent's method.
-           In ArchUnitTest, ArchUnitSuperClassTest assertion calls are not required as they
-           are called by the library -->
+           In ArchUnitTest, ArchUnitSuperClassTest, ImmutabilityTest assertion calls are not
+           required as they are called by the library -->
       <property name="violationSuppressXPath"
                value="//ClassOrInterfaceDeclaration[@SimpleName='AllChecksTest'
                      or @SimpleName='AstRegressionTest'
@@ -132,6 +132,8 @@
                    | //ClassOrInterfaceDeclaration[@SimpleName='XdocsJavaDocsTest']
                        //MethodDeclaration[@Name='testAllCheckSectionJavaDocs']
                    | //ClassOrInterfaceDeclaration[starts-with(@SimpleName,'XpathRegression')]
+                       //MethodDeclaration
+                   | //ClassOrInterfaceDeclaration[@SimpleName='ImmutabilityTest']
                        //MethodDeclaration
                    | //ClassOrInterfaceDeclaration[@SimpleName='ArchUnitTest']
                        //MethodDeclaration
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
@@ -1,0 +1,470 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code and other text files for adherence to a set of rules.
+// Copyright (C) 2001-2022 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package com.puppycrawl.tools.checkstyle.internal;
+
+import static com.tngtech.archunit.base.DescribedPredicate.doNot;
+import static com.tngtech.archunit.base.DescribedPredicate.not;
+import static com.tngtech.archunit.lang.conditions.ArchPredicates.are;
+import static com.tngtech.archunit.lang.conditions.ArchPredicates.have;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.fields;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import org.junit.jupiter.api.Test;
+
+import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
+import com.puppycrawl.tools.checkstyle.GlobalStatefulCheck;
+import com.puppycrawl.tools.checkstyle.StatelessCheck;
+import com.puppycrawl.tools.checkstyle.meta.ModuleDetails;
+import com.puppycrawl.tools.checkstyle.meta.ModulePropertyDetails;
+import com.puppycrawl.tools.checkstyle.meta.XmlMetaReader;
+import com.puppycrawl.tools.checkstyle.utils.ModuleReflectionUtil;
+import com.tngtech.archunit.base.DescribedPredicate;
+import com.tngtech.archunit.core.domain.JavaClass;
+import com.tngtech.archunit.core.domain.JavaClasses;
+import com.tngtech.archunit.core.domain.JavaField;
+import com.tngtech.archunit.core.domain.JavaModifier;
+import com.tngtech.archunit.core.domain.JavaParameterizedType;
+import com.tngtech.archunit.core.domain.JavaType;
+import com.tngtech.archunit.core.domain.properties.HasName;
+import com.tngtech.archunit.core.importer.ClassFileImporter;
+import com.tngtech.archunit.core.importer.ImportOption;
+import com.tngtech.archunit.lang.ArchCondition;
+import com.tngtech.archunit.lang.ArchRule;
+import com.tngtech.archunit.lang.ConditionEvents;
+import com.tngtech.archunit.lang.SimpleConditionEvent;
+
+public class ImmutabilityTest {
+
+    /**
+     * Immutable types canonical names.
+     */
+    private static final Set<String> IMMUTABLE_TYPES = Set.of(
+        "java.lang.String",
+        "java.lang.Integer",
+        "java.lang.Byte",
+        "java.lang.Character",
+        "java.lang.Short",
+        "java.lang.Boolean",
+        "java.lang.Long",
+        "java.lang.Double",
+        "java.lang.Float",
+        "java.lang.StackTraceElement",
+        "java.math.BigInteger",
+        "java.math.BigDecimal",
+        "java.io.File",
+        "java.util.Locale",
+        "java.util.UUID",
+        "java.net.URL",
+        "java.net.URI",
+        "java.net.Inet4Address",
+        "java.net.Inet6Address",
+        "java.net.InetSocketAddress",
+        "java.util.regex.Pattern"
+    );
+
+    /**
+     * Immutable primitive types.
+     */
+    private static final Set<String> PRIMITIVE_TYPES = Set.of(
+        "byte",
+        "short",
+        "int",
+        "long",
+        "float",
+        "double",
+        "char",
+        "boolean"
+    );
+
+    /**
+     * List of fields that are a zero size array. They are immutable by definition.
+     */
+    private static final Set<String> ZERO_SIZE_ARRAY_FIELDS = Set.of(
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_BIT_SET",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_BYTE_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_DOUBLE_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_INTEGER_OBJECT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_INT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_OBJECT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_STRING_ARRAY"
+    );
+
+    /**
+     * List of fields not following {@link #testUtilClassesImmutability()} rule.
+     */
+    private static final Set<String> SUPPRESSED_FIELDS_IN_UTIL_CLASSES = Set.of(
+        "com.puppycrawl.tools.checkstyle.utils.TokenUtil.TOKEN_IDS",
+        "com.puppycrawl.tools.checkstyle.utils.XpathUtil.TOKEN_TYPES_WITH_TEXT_ATTRIBUTE"
+    );
+
+    /**
+     * List of fields not following {@link #testFieldsInStatelessChecksShouldBeImmutable()} rule.
+     */
+    private static final Set<String> SUPPRESSED_FIELDS_IN_MODULES = Set.of(
+        "com.puppycrawl.tools.checkstyle.checks.FinalParametersCheck.primitiveDataTypes",
+        "com.puppycrawl.tools.checkstyle.checks.SuppressWarningsHolder.ENTRIES",
+        "com.puppycrawl.tools.checkstyle.checks.annotation.MissingDeprecatedCheck.TYPES_HASH_SET",
+        "com.puppycrawl.tools.checkstyle.checks.coding.AvoidDoubleBraceInitializationCheck"
+            + ".HAS_MEMBERS",
+        "com.puppycrawl.tools.checkstyle.checks.coding.AvoidDoubleBraceInitializationCheck"
+            + ".IGNORED_TYPES",
+        "com.puppycrawl.tools.checkstyle.checks.coding.InnerAssignmentCheck"
+            + ".ALLOWED_ASSIGNMENT_CONTEXT",
+        "com.puppycrawl.tools.checkstyle.checks.coding.InnerAssignmentCheck"
+            + ".ALLOWED_ASSIGNMENT_IN_COMPARISON_CONTEXT",
+        "com.puppycrawl.tools.checkstyle.checks.coding.InnerAssignmentCheck.COMPARISON_TYPES",
+        "com.puppycrawl.tools.checkstyle.checks.coding.InnerAssignmentCheck.CONTROL_CONTEXT",
+        "com.puppycrawl.tools.checkstyle.checks.coding.InnerAssignmentCheck"
+            + ".LOOP_IDIOM_IGNORED_PARENTS",
+        "com.puppycrawl.tools.checkstyle.checks.coding.MatchXpathCheck.xpathExpression",
+        "com.puppycrawl.tools.checkstyle.checks.javadoc.AtclauseOrderCheck.DEFAULT_ORDER",
+        "com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocBlockTagLocationCheck.DEFAULT_TAGS",
+        "com.puppycrawl.tools.checkstyle.checks.javadoc.SummaryJavadocCheck.ALLOWED_TYPES",
+        "com.puppycrawl.tools.checkstyle.checks.modifier.ModifierOrderCheck.JLS_ORDER",
+        "com.puppycrawl.tools.checkstyle.checks.modifier.RedundantModifierCheck"
+            + ".TOKENS_FOR_INTERFACE_MODIFIERS",
+        "com.puppycrawl.tools.checkstyle.checks.regexp.RegexpMultilineCheck.detector",
+        "com.puppycrawl.tools.checkstyle.checks.regexp.RegexpSinglelineCheck.detector",
+        "com.puppycrawl.tools.checkstyle.checks.coding.IllegalTokenTextCheck.formatString",
+        "com.puppycrawl.tools.checkstyle.checks.javadoc.WriteTagCheck.tagRegExp",
+        "com.puppycrawl.tools.checkstyle.checks.naming.AbstractNameCheck.format",
+        "com.puppycrawl.tools.checkstyle.checks.whitespace.AbstractParenPadCheck.option"
+    );
+
+    /**
+     * List of classes not following
+     * {@link #testClassesWithImmutableFieldsShouldBeStateless()} rule.
+     */
+    private static final Set<String> SUPPRESSED_CLASSES_FOR_STATELESS_CHECK_RULE = Set.of(
+        "com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocPackageCheck",
+        "com.puppycrawl.tools.checkstyle.checks.javadoc.MissingJavadocMethodCheck",
+        "com.puppycrawl.tools.checkstyle.checks.metrics.ClassDataAbstractionCouplingCheck",
+        "com.puppycrawl.tools.checkstyle.checks.metrics.ClassFanOutComplexityCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.CatchParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.ClassTypeParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.ConstantNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.InterfaceTypeParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.LambdaParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.LocalFinalVariableNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.LocalVariableNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.MemberNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.MethodNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.MethodTypeParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.ParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.PatternVariableNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.RecordComponentNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.RecordTypeParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.StaticVariableNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.whitespace.TypecastParenPadCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.TypeNameCheck"
+    );
+
+    /**
+     * List of classes not following {@link #testClassesWithMutableFieldsShouldBeStateful()} rule.
+     */
+    private static final Set<String> SUPPRESSED_CLASSES_FOR_STATEFUL_CHECK_RULE = Set.of(
+        "com.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck"
+    );
+
+    private static final JavaClasses CHECKSTYLE_CHECKS = new ClassFileImporter()
+        .withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_TESTS)
+        .importPackages("com.puppycrawl.tools.checkstyle")
+        .that(new DescribedPredicate<>("are checkstyle modules") {
+            @Override
+            public boolean apply(JavaClass input) {
+                final Class<?> clazz = input.reflect();
+                return ModuleReflectionUtil.isValidCheckstyleClass(clazz)
+                    && (ModuleReflectionUtil.isCheckstyleTreeWalkerCheck(clazz)
+                        || ModuleReflectionUtil.isFileSetModule(clazz));
+            }
+        });
+
+    /**
+     * ArchCondition for immutable fields.
+     */
+    private static final ArchCondition<JavaField> BE_IMMUTABLE = new ImmutableFieldArchCondition();
+
+    /**
+     * DescribedPredicate defining condition for a class to have immutable fields.
+     */
+    private static final DescribedPredicate<JavaClass> IMMUTABLE_FIELDS =
+        new ImmutableFieldsPredicate();
+
+    /**
+     * Map of module full name to module details.
+     */
+    private static final Map<String, ModuleDetails> MODULE_DETAILS_MAP =
+        XmlMetaReader.readAllModulesIncludingThirdPartyIfAny().stream()
+            .collect(Collectors.toMap(ModuleDetails::getFullQualifiedName,
+                                      Function.identity()));
+
+    /**
+     * Test to ensure that fields in util classes are immutable.
+     */
+    @Test
+    public void testUtilClassesImmutability() {
+        final JavaClasses utilClasses = new ClassFileImporter()
+            .withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_TESTS)
+            .importPackages("com.puppycrawl.tools.checkstyle.utils",
+                            "com.puppycrawl.tools.checkstyle.checks.javadoc.utils");
+
+        final ArchCondition<JavaField> beSuppressedField = new SuppressionArchCondition<>(
+            SUPPRESSED_FIELDS_IN_UTIL_CLASSES, "be suppressed");
+
+        final ArchRule fieldsInUtilClassesShouldBeImmutable = fields()
+            .that()
+            .areDeclaredInClassesThat()
+            .haveSimpleNameEndingWith("Util")
+            .should(BE_IMMUTABLE)
+            .andShould()
+            .beFinal()
+            .andShould()
+            .beStatic()
+            .orShould(beSuppressedField);
+
+        fieldsInUtilClassesShouldBeImmutable.check(utilClasses);
+    }
+
+    /**
+     * Test to ensure modules annotated with {@link StatelessCheck} contain immutable fields.
+     */
+    @Test
+    public void testFieldsInStatelessChecksShouldBeImmutable() {
+        final DescribedPredicate<JavaField> moduleProperties = new ModulePropertyPredicate();
+
+        final ArchCondition<JavaField> beSuppressedField = new SuppressionArchCondition<>(
+            SUPPRESSED_FIELDS_IN_MODULES, "be suppressed");
+
+        final ArchRule fieldsInStatelessChecksShouldBeImmutable = fields()
+            .that()
+            .areDeclaredInClassesThat()
+            .areAnnotatedWith(StatelessCheck.class)
+            .and(are(not(moduleProperties)))
+            .should(BE_IMMUTABLE)
+            .andShould()
+            .beFinal()
+            .orShould(beSuppressedField);
+
+        fieldsInStatelessChecksShouldBeImmutable.check(CHECKSTYLE_CHECKS);
+    }
+
+    /**
+     * Test to ensure classes with immutable fields are annotated with {@link StatelessCheck}.
+     */
+    @Test
+    public void testClassesWithImmutableFieldsShouldBeStateless() {
+        final ArchCondition<JavaClass> beSuppressedClass = new SuppressionArchCondition<>(
+            SUPPRESSED_CLASSES_FOR_STATELESS_CHECK_RULE, "be suppressed");
+
+        final ArchRule classesWithImmutableFieldsShouldBeStateless = classes()
+            .that(have(IMMUTABLE_FIELDS))
+            .and()
+            .doNotHaveModifier(JavaModifier.ABSTRACT)
+            .should()
+            .beAnnotatedWith(StatelessCheck.class)
+            .orShould(beSuppressedClass);
+
+        classesWithImmutableFieldsShouldBeStateless.check(CHECKSTYLE_CHECKS);
+    }
+
+    /**
+     * Test to ensure classes with mutable fields are annotated with {@link FileStatefulCheck} or
+     * {@link GlobalStatefulCheck}.
+     */
+    @Test
+    public void testClassesWithMutableFieldsShouldBeStateful() {
+        final ArchCondition<JavaClass> beSuppressedClass = new SuppressionArchCondition<>(
+            SUPPRESSED_CLASSES_FOR_STATEFUL_CHECK_RULE, "be suppressed");
+
+        final ArchRule classesWithMutableFieldsShouldBeStateful = classes()
+            .that(doNot(have(IMMUTABLE_FIELDS)))
+            .and()
+            .doNotHaveModifier(JavaModifier.ABSTRACT)
+            .should()
+            .beAnnotatedWith(FileStatefulCheck.class)
+            .orShould()
+            .beAnnotatedWith(GlobalStatefulCheck.class)
+            .orShould(beSuppressedClass);
+
+        classesWithMutableFieldsShouldBeStateful.check(CHECKSTYLE_CHECKS);
+    }
+
+    /**
+     * ArchCondition checking fields are immutable.
+     */
+    private static final class ImmutableFieldArchCondition extends ArchCondition<JavaField> {
+        private ImmutableFieldArchCondition() {
+            super("be among immutable types");
+        }
+
+        /**
+         * Whether the raw type of the field is immutable.
+         *
+         * @param javaField java field to examine
+         * @return {@code true} if the raw type of field is immutable.
+         */
+        private static boolean isRawTypeImmutable(JavaField javaField) {
+            final JavaClass rawType = javaField.getRawType();
+            final String rawTypeName = rawType.getName();
+            return PRIMITIVE_TYPES.contains(rawTypeName)
+                || IMMUTABLE_TYPES.contains(rawTypeName);
+        }
+
+        /**
+         * Whether the field is an enum constant or an empty array.
+         *
+         * @param javaField java field to examine
+         * @return {@code true} if the field is an enum constant or an empty array
+         */
+        private static boolean isEnumConstantOrEmptyArray(JavaField javaField) {
+            final JavaClass rawType = javaField.getRawType();
+            return rawType.isEnum()
+                || ZERO_SIZE_ARRAY_FIELDS.contains(javaField.getFullName());
+        }
+
+        /**
+         * Whether the parameterized type of a field is immutable if it contains parameterized
+         * type. Parameterized type refers to the generic type of a field.
+         * {@code List<String>}, here the concrete type of parameterized field is
+         * {@code java.lang.String}.
+         *
+         * @param javaField java field to examine
+         * @return {@code true} if the parameterized type of a field is immutable
+         *         if it contains parameterized type
+         */
+        private static boolean isParameterizedTypeImmutable(JavaField javaField) {
+            boolean isParameterizedTypeImmutable = false;
+            final JavaType javaType = javaField.getType();
+
+            if (javaType instanceof JavaParameterizedType) {
+                final JavaParameterizedType parameterizedType = (JavaParameterizedType) javaType;
+                isParameterizedTypeImmutable = parameterizedType.getActualTypeArguments().stream()
+                    .allMatch(actualTypeArgument -> {
+                        return IMMUTABLE_TYPES.contains(actualTypeArgument.toErasure().getName());
+                    });
+            }
+            return isParameterizedTypeImmutable;
+        }
+
+        @Override
+        public void check(JavaField item, ConditionEvents events) {
+            if (!isRawTypeImmutable(item)
+                && !isEnumConstantOrEmptyArray(item)
+                && !isParameterizedTypeImmutable(item)) {
+                final String message = String
+                    .format(Locale.ROOT, "Field <%s> should %s in %s",
+                            item.getFullName(), getDescription(),
+                            item.getSourceCodeLocation().toString());
+                events.add(SimpleConditionEvent.violated(item, message));
+            }
+        }
+    }
+
+    /**
+     * DescribedPredicate defining condition for a field to be a module property.
+     */
+    private static final class ModulePropertyPredicate extends DescribedPredicate<JavaField> {
+
+        private ModulePropertyPredicate() {
+            super("module properties");
+        }
+
+        /**
+         * Whether a field is a module property or not.
+         *
+         * @param javaField field to check
+         * @return {@code true} if field is a module property
+         */
+        private static boolean isModuleProperty(JavaField javaField) {
+            boolean result = false;
+            final JavaClass containingClass = javaField.getOwner();
+            final ModuleDetails moduleDetails = MODULE_DETAILS_MAP.get(
+                containingClass.getFullName());
+            if (moduleDetails != null) {
+                final List<ModulePropertyDetails> properties = moduleDetails.getProperties();
+                result = properties.stream()
+                    .map(ModulePropertyDetails::getName)
+                    .anyMatch(moduleName -> moduleName.equals(javaField.getName()));
+            }
+            return result;
+        }
+
+        @Override
+        public boolean apply(JavaField input) {
+            return isModuleProperty(input);
+        }
+    }
+
+    /**
+     * DescribedPredicate defining condition for a class to have immutable fields.
+     */
+    private static final class ImmutableFieldsPredicate extends DescribedPredicate<JavaClass> {
+        private ImmutableFieldsPredicate() {
+            super("immutable fields");
+        }
+
+        @Override
+        public boolean apply(JavaClass input) {
+            final Set<JavaField> fields = input.getFields();
+            return fields.stream()
+                .filter(javaField -> {
+                    return !ModulePropertyPredicate.isModuleProperty(javaField)
+                        && !SUPPRESSED_FIELDS_IN_MODULES.contains(javaField.getFullName());
+                })
+                .allMatch(javaField -> {
+                    final Set<JavaModifier> javaFieldModifiers = javaField.getModifiers();
+                    return javaFieldModifiers.contains(JavaModifier.FINAL)
+                        && (ImmutableFieldArchCondition.isRawTypeImmutable(javaField)
+                            || ImmutableFieldArchCondition.isEnumConstantOrEmptyArray(javaField)
+                            || ImmutableFieldArchCondition.isParameterizedTypeImmutable(javaField));
+                });
+        }
+    }
+
+    /**
+     * ArchCondition checking if a type or a member is present in the suppression list.
+     */
+    private static final class SuppressionArchCondition<T extends HasName.AndFullName>
+        extends ArchCondition<T> {
+
+        private final Set<String> suppressions;
+
+        private SuppressionArchCondition(Set<String> suppressions, String description) {
+            super(description);
+            this.suppressions = suppressions;
+        }
+
+        @Override
+        public void check(HasName.AndFullName item, ConditionEvents events) {
+            if (!suppressions.contains(item.getFullName())) {
+                final String message = String.format(
+                    Locale.ROOT, "should %s or resolved.", getDescription());
+                events.add(SimpleConditionEvent.violated(item, message));
+            }
+        }
+    }
+}


src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
```diff
@@ -0,0 +1,149 @@
...
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_BYTE_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_DOUBLE_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_INTEGER_OBJECT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_INT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_OBJECT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_STRING_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.JavadocUtil$JavadocTagType.$VALUES",
+        "com.puppycrawl.tools.checkstyle.utils.JavadocUtil$JavadocTagType.ALL",
+        "com.puppycrawl.tools.checkstyle.utils.JavadocUtil$JavadocTagType.BLOCK",
+        "com.puppycrawl.tools.checkstyle.utils.JavadocUtil$JavadocTagType.INLINE",
```
rnveach, line 81: Please explain how an enum is not immutable?

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
```diff
@@ -0,0 +1,149 @@
...
+        "float",
+        "char",
+        "boolean"
+    );
+
+    /**
+     * List of suppressed fields.
+     */
+    private static final Set<String> SUPPRESSED_FIELDS = Set.of(
+        "com.puppycrawl.tools.checkstyle.checks.javadoc.utils.TagInfo.position",
```
rnveach, line 70: I thought utils had to end with `Util`?

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
```diff
@@ -0,0 +1,149 @@
...
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_BIT_SET",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_BYTE_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_DOUBLE_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_INTEGER_OBJECT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_INT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_OBJECT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_STRING_ARRAY",
```
rnveach, line 77: It is interesting that these are not immutable. I understand they aren't but this means someone can hijack things and modify them. @romani Do we want to consider some type of paradigm switch?

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
```diff
@@ -0,0 +1,149 @@
...
+                    final JavaParameterizedType parameterizedType =
+                        (JavaParameterizedType) javaType;
+                    addViolation = !parameterizedType.getActualTypeArguments().stream()
+                        .allMatch(actualTypeArgument -> {
+                            return DEFAULT_IMMUTABLE_TYPES.contains(
+                                actualTypeArgument.toErasure().getName());
+                        });
+                }
+            }
+            else if ("long".equals(rawTypeName) || "double".equals(rawTypeName)) {
```
rnveach, line 120: Why is long/double separated from primitive types? They are immutable if they have `final` otherwise they are not immutable.

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
Vyom-Yadav: It is, these were old suppression (redundant), I updated and verified the list.

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
Vyom-Yadav: Again, old suppression, updated the list.

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
rnveach: @Vyom-Yadav These items are actually immutable. They are a 0 length array, there is nothing to mutate.
These should be commented as a false positive.

`TOKEN_IDS` is not immutable since it doesn't have a 0 length.

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
```diff
@@ -0,0 +1,148 @@
...
+    private static final Set<String> SUPPRESSED_FIELDS = Set.of(
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_BIT_SET",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_BYTE_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_DOUBLE_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_INTEGER_OBJECT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_INT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_OBJECT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_STRING_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.TokenUtil.TOKEN_IDS",
+        "com.puppycrawl.tools.checkstyle.utils.XpathUtil.TOKEN_TYPES_WITH_TEXT_ATTRIBUTE"
```
rnveach, line 79: BitSet is mutable.

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
Vyom-Yadav: Separated zero-size arrays from actual violations.

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
Vyom-Yadav: Let's fix violations in a separate PR, for now this can stay in suppression list.

rnveach: Agreed, leaving for discussion.

config/pmd-test.xml
```diff
@@ -132,6 +132,8 @@
                        //MethodDeclaration[@Name='testAllCheckSectionJavaDocs']
                    | //ClassOrInterfaceDeclaration[starts-with(@SimpleName,'XpathRegression')]
                        //MethodDeclaration
+                   | //ClassOrInterfaceDeclaration[@SimpleName='UtilClassesImmutabilityTest']
```
rnveach, line 135: Reason has to be added above.

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
```diff
@@ -0,0 +1,181 @@
...
+            .beFinal()
+            .andShould()
+            .beStatic();
+
+        testImmutability.check(importedClasses);
+    }
+
+    private static final class JavaFieldArchCondition extends ArchCondition<JavaField> {
+        private JavaFieldArchCondition() {
+            super("be among default immutable types");
```
rnveach, line 148: What does default mean here?

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
```diff
@@ -0,0 +1,181 @@
...
+import com.tngtech.archunit.lang.ArchRule;
+import com.tngtech.archunit.lang.ConditionEvents;
+import com.tngtech.archunit.lang.SimpleConditionEvent;
+
+public class UtilClassesImmutabilityTest {
+
+    /**
+     * Default immutable types canonical names.
+     */
+    private static final Set<String> DEFAULT_IMMUTABLE_TYPES = Set.of(
```
rnveach, line 46: Same? Why not just call this immutable types?

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
```diff
@@ -0,0 +1,181 @@
...
+import com.tngtech.archunit.core.domain.JavaParameterizedType;
+import com.tngtech.archunit.core.domain.JavaType;
+import com.tngtech.archunit.core.importer.ClassFileImporter;
+import com.tngtech.archunit.core.importer.ImportOption;
+import com.tngtech.archunit.lang.ArchCondition;
+import com.tngtech.archunit.lang.ArchRule;
+import com.tngtech.archunit.lang.ConditionEvents;
+import com.tngtech.archunit.lang.SimpleConditionEvent;
+
+public class UtilClassesImmutabilityTest {
```
rnveach, line 41: Last comment, we talked about checks and immutable types. I assume if we try to move forward with it, its going to use similar code to this.

Why not just try to name this immutability test?

rnveach (changes requested):

config/pmd-test.xml
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
Vyom-Yadav: Removed default.

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
Vyom-Yadav: Removed default.

src/test/java/com/puppycrawl/tools/checkstyle/internal/UtilClassesImmutabilityTest.java
Vyom-Yadav: Done, changed name.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,433 @@
...
+        "com.puppycrawl.tools.checkstyle.checks.naming.TypeNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.whitespace.TypecastParenPadCheck",
+        // Failing 'testClassesWithoutImmutableFieldsAreStateful'
+        "com.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck"
+    );
+
+    /**
+     * ArchCondition for immutable fields.
+     */
+    private static final ArchCondition<JavaField> BE_IMMUTABLE = new ImmutableFieldArchCondition();
```
nrmancuso, line 189: Can we add condition for zero size arrays here to avoid ` .that(are(not(zeroSizeArrays)))` below? It seems like in test methods, this should not be a separate condition.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,433 @@
...
+
+        final DescribedPredicate<JavaField> zeroSizeArrays =
+            new DescribedPredicate<>("zero size arrays") {
+                @Override
+                public boolean apply(JavaField input) {
+                    return ZERO_SIZE_ARRAY_FIELDS.contains(input.getFullName());
+                }
+            };
+
+        final ArchCondition<JavaField> beSuppressed = new ArchCondition<>("be suppressed") {
```
nrmancuso, line 221: If all `beSuppressed` conditions are basically the same (other than the collection we are checking), can we just make some static method where we pass in the name of the collection to build these conditions? 

`final  ArchCondition<JavaField> beSuppressed = getSuppressionCondition(SUPPRESSED_FIELDS_IN_UTIL_CLASSES);`

Or, perhaps even more generically, we pass in a Predicate (java, not archunit).

`final  ArchCondition<JavaField> beMessageKey = getSuppressionConditionByPredicate(it -> it.getName().startsWith("MSG"));`

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,433 @@
...
+                        final String message = String
+                            .format(Locale.ROOT, "Field <%s> should %s in %s",
+                                    item.getFullName(), getDescription(),
+                                    item.getSourceCodeLocation().toString());
+                        events.add(SimpleConditionEvent.violated(item, message));
+                    }
+                }
+            };
+
+        final ArchCondition<JavaClass> beAnnotatedOrSuppressed = new ArchCondition<>(
```
nrmancuso, line 282: I do not see logic here for annotation checking, what am I missing?

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,433 @@
...
+        }
+    }
+
+    private static final class ModulePropertyPredicate extends DescribedPredicate<JavaField> {
+
+        private ModulePropertyPredicate() {
+            super("module properties");
+        }
+
+        private static boolean isModuleProperty(JavaField javaFiled) {
```
nrmancuso, line 389: `javaFiled` -> `javaField`

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,433 @@
...
+        classesWithImmutableFieldsShouldBeStateless.check(checkstyleChecks);
+        classesWithMutableFieldsShouldBeStateful.check(checkstyleChecks);
+    }
+
+    private static final class ImmutableFieldArchCondition extends ArchCondition<JavaField> {
+        private ImmutableFieldArchCondition() {
+            super("be among immutable types");
+        }
+
+        private static boolean isMutable(JavaField javaField) {
```
nrmancuso, line 348: I feel like checking for `isMutable` in `ImmutableFieldArchCondition` is strange... It feels like we should be checking for immutability if anything, or tweak class to be `MutableFieldArchcondition` and negate it.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,433 @@
...
+                    events.add(SimpleConditionEvent.violated(item, message));
+                }
+            }
+        };
+
+        final ArchRule fieldsShouldBeImmutable = fields()
+            .that(are(not(zeroSizeArrays)))
+            .and()
+            .areDeclaredInClassesThat()
+            .haveSimpleNameEndingWith("Util")
```
nrmancuso, line 236: Is it possible to get all classes by package, as for api/util package rule?

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,433 @@
...
+            .orShould()
+            .beAnnotatedWith(GlobalStatefulCheck.class)
+            .orShould(beAnnotatedOrSuppressed);
+
+        fieldsInStatelessChecksAreImmutable.check(checkstyleChecks);
+        classesWithImmutableFieldsShouldBeStateless.check(checkstyleChecks);
+        classesWithMutableFieldsShouldBeStateful.check(checkstyleChecks);
+    }
+
+    private static final class ImmutableFieldArchCondition extends ArchCondition<JavaField> {
```
nrmancuso, line 343: These private classes are not trivial, please add a javadoc to explain what's going on here.

nrmancuso (changes requested):
Few items to start, if we can clean up suppression predicate a bit it will allow us to focus on actual conditions.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,433 @@
...
+
+        final ArchRule classesWithImmutableFieldsShouldBeStateless = classes()
+            .that(have(immutableFields))
+            .and()
+            .haveSimpleNameEndingWith("Check")
+            .and()
+            .doNotHaveModifier(JavaModifier.ABSTRACT)
+            .should()
+            .beAnnotatedWith(StatelessCheck.class)
+            .orShould(beAnnotatedOrSuppressed);
```
nrmancuso, line 324: Let's avoid `Or` in our predicates, and rely on build in `.orShould` to keep predicates as simple as possible.

nrmancuso (changes requested):
One more:

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Nothing, the variable name was incorrect, it only checks suppressions.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: We are getting the required classes only, the ones present in `utils` packages, but there are some classes we need to exclude, like `com.puppycrawl.tools.checkstyle.checks.javadoc.utils.TagInfo` that's why this condition is there.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: The variable name here was incorrect which caused the confusion, corrected now.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done, made a separate class, didn't make a generic (`<T>`) one as it wasn't looking neat.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,467 @@
...
+        final ArchRule fieldsInStatelessChecksAreImmutable = fields()
+            .that()
+            .areDeclaredInClassesThat()
+            .areAnnotatedWith(StatelessCheck.class)
+            .and(are(not(moduleProperties)))
+            .should(BE_IMMUTABLE)
+            .andShould()
+            .beFinal()
+            .andShould(bePrivateExceptMessageKeys)
+            .orShould(beSuppressed);
```
nrmancuso, line 281: Let's try to keep conditions as simple as possible:

```suggestion
            .andShould(bePrivate)
            .orShould(beMessageKeys)
            .orShould(beSuppressed);
```

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,467 @@
...
+     * @noinspectionreason JUnitTestMethodWithNoAssertions - asserts in callstack,
+     *     but not in this method
+     */
+    @Test
+    public void testModuleImmutability() {
+        final JavaClasses checkstyleChecks = new ClassFileImporter()
+            .withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_TESTS)
+            .importPackages("com.puppycrawl.tools.checkstyle.checks");
+
+        final ModulePropertyPredicate moduleProperties = new ModulePropertyPredicate();
```
nrmancuso, line 245: `final DescribedPredicate<JavaField> moduleProperties = new ModulePropertyPredicate();`

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,467 @@
...
+                        final String message = String
+                            .format(Locale.ROOT, "Field <%s> should %s in %s",
+                                    item.getFullName(), getDescription(),
+                                    item.getSourceCodeLocation().toString());
+                        events.add(SimpleConditionEvent.violated(item, message));
+                    }
+                }
+            };
+
+        final ArchRule fieldsInStatelessChecksAreImmutable = fields()
```
nrmancuso, line 272: `Are` -> `ShouldBe`

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,467 @@
...
+    public void testModuleImmutability() {
+        final JavaClasses checkstyleChecks = new ClassFileImporter()
+            .withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_TESTS)
+            .importPackages("com.puppycrawl.tools.checkstyle.checks");
+
+        final ModulePropertyPredicate moduleProperties = new ModulePropertyPredicate();
+
+        final DescribedPredicate<JavaClass> immutableFields = new ImmutableFieldsPredicate();
+
+        final ArchCondition<JavaClass> beAddedToSuppressionList =
```
nrmancuso, line 249: `beAddedToSuppressionList` -> `beSuppressedClass`

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,467 @@
...
+            new JavaClassSuppressionArchCondition(
+                SUPPRESSED_CLASSES, "be added to the suppression list");
+
+        final ArchCondition<JavaField> beSuppressed = new JavaFieldSuppressionArchCondition(
```
nrmancuso, line 253: `beSuppressed` -> `beSuppressedField`

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,467 @@
...
+
+        @Override
+        public void check(JavaClass item, ConditionEvents events) {
+            if (!suppressedClasses.contains(item.getFullName())) {
+                final String message = String.format(
+                    Locale.ROOT, "should %s or resolved.", getDescription());
+                events.add(SimpleConditionEvent.violated(item, message));
+            }
+        }
+    }
```
nrmancuso, line 443: This will allow you to have only one condition class:

```suggestion
    private static final class ArchRuleSuppressionCondition<T extends HasName.AndFullName>
            extends ArchCondition<T> {

        private final Set<String> suppressedClasses;

        private ArchRuleSuppressionCondition(Set<String> suppressedClasses,
                                             String description) {
            super(description);
            this.suppressedClasses = suppressedClasses;
        }

        @Override
        public void check(T item, ConditionEvents events) {
            if (!suppressedClasses.contains(item.getFullName())) {
                final String message = String.format(
                    Locale.ROOT, "should %s or resolved.", getDescription());
                events.add(SimpleConditionEvent.violated(item, message));
            }
        }
    }
```

nrmancuso (changes requested):
Few more items:

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
nrmancuso, line 443: Please do this to avoid duplicated classes:
```suggestion

    private static final class ArchRuleSuppressionCondition<T extends HasName.AndFullName>
            extends ArchCondition<T> {

        private final Set<String> suppressions;

        private ArchRuleSuppressionCondition(Set<String> suppressions,
                                             String description) {
            super(description);
            this.suppressions = suppressions;
        }

        @Override
        public void check(T item, ConditionEvents events) {
            if (!suppressions.contains(item.getFullName())) {
                final String message = String.format(
                    Locale.ROOT, "should %s or resolved.", getDescription());
                events.add(SimpleConditionEvent.violated(item, message));
            }
        }
    }

```

nrmancuso (changes requested):
Item:

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: I already tried this, AndFullName is package private.

nrmancuso: > AndFullName is package private

No, this is not correct: https://github.com/TNG/ArchUnit/blob/c081b98f139546962c98c89f135f78edab27de49/archunit/src/main/java/com/tngtech/archunit/core/domain/properties/HasName.java#L33

Please try again.

Vyom-Yadav: Thanks for clarifying :), done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Both conditions are combined as `orShould(..)` will result in `(a && b && c) || d`, whereas we would like `a && b && (c || d)`.
```java
final ArchRule fieldsInStatelessChecksShouldBeImmutable = fields()                 
    .that()                                                                        
    .areDeclaredInClassesThat()                                                    
    .areAnnotatedWith(StatelessCheck.class)                                        
    .and(are(not(moduleProperties)))                                               
    .should(BE_IMMUTABLE)                                                          
    .andShould()                                                                   
    .beFinal()                                                                     
    .andShould()                                                                   
    .bePrivate()                                                                   
    .orShould()                                                                    
    .haveNameStartingWith("MSG")                                                   
    .orShould(beSuppressedField);                                                 
```
will not give a violation for 
```
public static String MSG_KEY = "variable.declaration.usage.distance";
```
whereas the current `ArchRule` will.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,467 @@
...
+    /**
+     * List of suppressed fields in util classes.
+     */
+    private static final Set<String> SUPPRESSED_FIELDS_IN_UTIL_CLASSES = Set.of(
+        "com.puppycrawl.tools.checkstyle.utils.TokenUtil.TOKEN_IDS",
+        "com.puppycrawl.tools.checkstyle.utils.XpathUtil.TOKEN_TYPES_WITH_TEXT_ATTRIBUTE"
+    );
+
+    /**
+     * List of fields not following {@code testFieldsInStatelessCheckAreImmutable} ArchRule.
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,444 @@
...
+     * Test to ensure modules annotated with {@link StatelessCheck} contain immutable fields
+     * and modules with mutable fields are annotated with {@link FileStatefulCheck} or
+     * {@link GlobalStatefulCheck}.
+     *
+     * @noinspection JUnitTestMethodWithNoAssertions
+     * @noinspectionreason JUnitTestMethodWithNoAssertions - asserts in callstack,
+     *     but not in this method
+     */
+    @Test
+    public void testModuleImmutability() {
```
nrmancuso, line 241: @Vyom-Yadav is there some benefit to having three rules in one unit test? I think it is better to keep paradigm of 1 rule == 1 unit test, especially with complicated rules.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,444 @@
...
+
+        final ArchRule classesWithImmutableFieldsShouldBeStateless = classes()
+            .that(have(immutableFields))
+            .and()
+            .haveSimpleNameEndingWith("Check")
+            .and()
+            .doNotHaveModifier(JavaModifier.ABSTRACT)
+            .should()
+            .beAnnotatedWith(StatelessCheck.class)
+            .orShould(beSuppressedClass);
```
nrmancuso, line 291: Let's have two separate suppression lists, one for this rule and one for the below rule. We can use both lists in a given rule; we can filter by the 'other' list first with some ArchCondition like `shouldBeAnnotatedWith*OtherAnnotation*`.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,444 @@
...
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_OBJECT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_STRING_ARRAY"
+    );
+
+    /**
+     * List of classes not following {@code classesWithImmutableFieldsShouldBeStateless} and
+     * {@code classesWithMutableFieldsShouldBeStateful} ArchRule.
+     */
+    private static final Set<String> SUPPRESSED_CLASSES = Set.of(
+        // Failing 'testClassesWithImmutableFieldsAreStateless'
```
nrmancuso, line 161: Separate lists will eliminate the need for these comments, otherwise we would need to update them to new name of rule.

nrmancuso (changes requested):
Few more items, we are getting close now:

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
rnveach: I brought this up that we could have multiple tests that need mutability checks. I am ok with splitting things, but I don't think we should duplicate code.

nrmancuso: There should be very little/ no code duplication; conditions can be extracted to fields in the test class.

Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Created separate lists.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Created separate lists.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,475 @@
...
+        /**
+         * Whether a field is immutable or not.
+         *
+         * @param javaField field to check
+         * @return {@code true} if field is immutable
+         */
+        private static boolean isImmutable(JavaField javaField) {
+            final JavaClass rawType = javaField.getRawType();
+            boolean isImmutable = true;
+            if (!rawType.isEnum() && !ZERO_SIZE_ARRAY_FIELDS.contains(javaField.getFullName())) {
```
nrmancuso, line 358: Please extract this to a boolean, maybe something like:
`final boolean isEnumConstantOrEmptyArray = rawType.isEnum() || ZERO_SIZE_ARRAY_FIELDS.contains(javaField.getFullName()`

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,475 @@
...
+                final String rawTypeName = rawType.getName();
+                if (!PRIMITIVE_TYPES.contains(rawTypeName)
+                    && !IMMUTABLE_TYPES.contains(rawTypeName)) {
```
nrmancuso, line 361: Please extract this condition to a boolean as well, maybe just called `isImmutableType`

nrmancuso (changes requested):
Immutability is still a somewhat nebulous definition to us, as we have no concrete way to say 100% that something is or isn't immutable. Let's make this method easy to understand and maintain as time goes on and we find more conditions to check.

Please also add some whitespace in here.


Items:

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

Vyom-Yadav:
@nrmancuso I enhanced the JavaDoc of the method to clarify further what are we trying to achieve.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,493 @@
...
+    private static final ArchCondition<JavaField> BE_IMMUTABLE = new ImmutableFieldArchCondition();
+
+    /**
+     * DescribedPredicate defining condition for a class to have immutable fields.
+     */
+    private static final DescribedPredicate<JavaClass> IMMUTABLE_FIELDS =
+        new ImmutableFieldsPredicate();
+
+    /**
+     * Map fo module full name to module details.
```
pbludov, line 206: mistype: `Map fo`

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,493 @@
...
+     */
+    private static final Map<String, ModuleDetails> MODULE_DETAILS_MAP =
+        XmlMetaReader.readAllModulesIncludingThirdPartyIfAny().stream()
+            .collect(Collectors.toMap(ModuleDetails::getFullQualifiedName,
+                                      moduleDetails -> moduleDetails));
```
pbludov, line 211: no op lambda can be replaced with `Function.identity()`

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,493 @@
...
+
+        final ArchCondition<JavaField> beSuppressedField = new SuppressionArchCondition<>(
+            SUPPRESSED_FIELDS_IN_MODULES, "be suppressed");
+
+        final ArchCondition<JavaField> bePrivateExceptMessageKeys =
+            new ArchCondition<>("be private") {
+
+                @Override
+                public void check(JavaField item, ConditionEvents events) {
+                    if (!item.getName().startsWith("MSG")
```
pbludov, line 263: Should it be `.startsWith("MSG_")` ?

pbludov (changes requested):
a bit of minor changes please

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,493 @@
...
+                .filter(javaField -> {
+                    return !ModulePropertyPredicate.isModuleProperty(javaField)
+                        && !SUPPRESSED_FIELDS_IN_MODULES.contains(javaField.getFullName());
+                })
+                .allMatch(javaField -> {
+                    final Set<JavaModifier> javaFieldModifiers = javaField.getModifiers();
+                    boolean result = javaFieldModifiers.contains(JavaModifier.FINAL)
+                        && ImmutableFieldArchCondition.isImmutable(javaField);
+
+                    if (!javaField.getName().contains("MSG")) {
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

pbludov (approved):

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+         * @return {@code true} if field is immutable
+         */
+        private static boolean isImmutable(JavaField javaField) {
+            final JavaClass rawType = javaField.getRawType();
+            boolean isImmutable = true;
+
+            final boolean isEnumConstantOrEmptyArray = !rawType.isEnum()
+                && !ZERO_SIZE_ARRAY_FIELDS.contains(javaField.getFullName());
+
+            if (isEnumConstantOrEmptyArray) {
```
nrmancuso, line 374:
```suggestion
            final boolean isEnumConstantOrEmptyArray = rawType.isEnum()
                || ZERO_SIZE_ARRAY_FIELDS.contains(javaField.getFullName());

            if (!isEnumConstantOrEmptyArray) {
```

nrmancuso (changes requested):
@Vyom-Yadav please fix condition name/ logic

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: My bad, fixed.

nrmancuso (approved):

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+
+        final ArchRule fieldsInStatelessChecksShouldBeImmutable = fields()
+            .that()
+            .areDeclaredInClassesThat()
+            .areAnnotatedWith(StatelessCheck.class)
+            .and(are(not(moduleProperties)))
+            .should(BE_IMMUTABLE)
+            .andShould()
+            .beFinal()
+            .andShould(bePrivateExceptMessageKeys)
```
rnveach, line 283: I think I am getting cross-eyed here, but all fields should be immutable period regardless of public/private etc... for Stateless.

There is a discrepancy here. Test is called `testFieldsInStatelessChecksShouldBeImmutable`. It does not mention the type of fields (private), so one would naturally assume there is no type requirement.

It seems to me `andShould(bePrivateExceptMessageKeys)` so should be a separate test but apply to all modules as there should be no direct communication between isolated checks.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+     */
+    @Test
+    public void testClassesWithImmutableFieldsShouldBeStateless() {
+        final ArchCondition<JavaClass> beSuppressedClass = new SuppressionArchCondition<>(
+            SUPPRESSED_CLASSES_FOR_STATELESS_CHECK_RULE, "be suppressed");
+
+        final ArchRule classesWithImmutableFieldsShouldBeStateless = classes()
+            .that(have(IMMUTABLE_FIELDS))
+            .and()
+            .haveSimpleNameEndingWith("Check")
```
rnveach, line 304: Similar to a previous item, we can't use `ModuleReflectionUtil#isCheckstyleModule`?

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+     * @noinspectionreason JUnitTestMethodWithNoAssertions - asserts in callstack,
+     *     but not in this method
+     */
+    @Test
+    public void testClassesWithImmutableFieldsShouldBeStateless() {
+        final ArchCondition<JavaClass> beSuppressedClass = new SuppressionArchCondition<>(
+            SUPPRESSED_CLASSES_FOR_STATELESS_CHECK_RULE, "be suppressed");
+
+        final ArchRule classesWithImmutableFieldsShouldBeStateless = classes()
+            .that(have(IMMUTABLE_FIELDS))
```
rnveach, line 302: Question: Does this pick up classes with no fields?

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+                            });
+                    }
+                }
+            }
+            return isImmutable;
+        }
+
+        @Override
+        public void check(JavaField item, ConditionEvents events) {
+            if (!isImmutable(item)) {
```
rnveach, line 398: Wouldn't this be better and easier to ready to split into like 3 different methods?
`!isRawTypeImmutable || !isEnumConstantOrEmptyArray || !isParameterizedTypeImmutable`

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+        public boolean apply(JavaClass input) {
+            final Set<JavaField> fields = input.getFields();
+            return fields.stream()
+                .filter(javaField -> {
+                    return !ModulePropertyPredicate.isModuleProperty(javaField)
+                        && !SUPPRESSED_FIELDS_IN_MODULES.contains(javaField.getFullName());
+                })
+                .allMatch(javaField -> {
+                    final Set<JavaModifier> javaFieldModifiers = javaField.getModifiers();
+                    boolean result = javaFieldModifiers.contains(JavaModifier.FINAL)
```
rnveach, line 461: `final` does not make a field immutable. An array list can be made final but you can still add new items to it.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+                        && ImmutableFieldArchCondition.isImmutable(javaField);
+                    // 'MSG_'s in checks are public as they are imported in test files.
+                    if (!javaField.getName().contains("MSG_")) {
+                        result = result && javaFieldModifiers.contains(JavaModifier.PRIVATE);
```
rnveach, line 465: Connected to an earlier comment, I don't think private and immutable should be directly connected.

rnveach (changes requested):
Did we discuss going over these suppressions, either in new issue or ... ?

romani:
I recommend to apply all easy to fix items but all other, move to separate issue.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+        "int",
+        "long",
+        "float",
+        "double",
+        "char",
+        "boolean"
+    );
+
+    /**
+     * List of suppressed fields in util classes.
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+     */
+    private static final Set<String> SUPPRESSED_FIELDS_IN_UTIL_CLASSES = Set.of(
+        "com.puppycrawl.tools.checkstyle.utils.TokenUtil.TOKEN_IDS",
+        "com.puppycrawl.tools.checkstyle.utils.XpathUtil.TOKEN_TYPES_WITH_TEXT_ATTRIBUTE"
+    );
+
+    /**
+     * List of fields not following {@code fieldsInStatelessChecksShouldBeImmutable} ArchRule.
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_BYTE_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_DOUBLE_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_INTEGER_OBJECT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_INT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_OBJECT_ARRAY",
+        "com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_STRING_ARRAY"
+    );
+
+    /**
+     * List of classes not following {@code classesWithImmutableFieldsShouldBeStateless} ArchRule.
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+        "com.puppycrawl.tools.checkstyle.checks.coding.IllegalTokenTextCheck.formatString",
+        "com.puppycrawl.tools.checkstyle.checks.javadoc.WriteTagCheck.tagRegExp",
+        "com.puppycrawl.tools.checkstyle.checks.naming.AbstractNameCheck.format",
+        "com.puppycrawl.tools.checkstyle.checks.whitespace.AbstractParenPadCheck.option"
+    );
+
+    /**
+     * List of fields that are a zero size array. They are immutable by definition.
+     */
+    private static final Set<String> ZERO_SIZE_ARRAY_FIELDS = Set.of(
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+        "com.puppycrawl.tools.checkstyle.checks.naming.ParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.PatternVariableNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.RecordComponentNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.RecordTypeParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.StaticVariableNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.whitespace.TypecastParenPadCheck"
+    );
+
+    /**
+     * List of classes not following {@code classesWithMutableFieldsShouldBeStateful} ArchRule.
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,494 @@
...
+     */
+    private static final Set<String> SUPPRESSED_CLASSES_FOR_STATEFUL_CHECK_RULE = Set.of(
+        "com.puppycrawl.tools.checkstyle.checks.naming.TypeNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck"
+    );
+
+    private static final JavaClasses CHECKSTYLE_CHECKS = new ClassFileImporter()
+        .withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_TESTS)
+        .importPackages("com.puppycrawl.tools.checkstyle.checks");
```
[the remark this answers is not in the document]
Vyom-Yadav: Done, used `ModuleReflectionUtil#isCheckstyleModule`, also I have kept package as `com.puppycrawl.tools.checkstyle.checks` as we aren't checking filters and only checkstyle checks, all the new rules work around `@StatelessCheck` and `@FileStatefulCheck` so only `checks` package was imported.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Removed.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Yes, it does:
```java
@Override                                                                               
public boolean apply(JavaClass input) {                                                 
    final Set<JavaField> fields = input.getFields();                                    
    return fields.stream()                                                              
        .filter(javaField -> {                                                          
            return !ModulePropertyPredicate.isModuleProperty(javaField)                 
                && !SUPPRESSED_FIELDS_IN_MODULES.contains(javaField.getFullName());     
        })                                                                              
        .allMatch(javaField -> {                                                        
            final Set<JavaModifier> javaFieldModifiers = javaField.getModifiers();      
            boolean result = javaFieldModifiers.contains(JavaModifier.FINAL)            
                && ImmutableFieldArchCondition.isImmutable(javaField);                  
            // 'MSG_'s in checks are public as they are imported in test files.         
            if (!javaField.getName().contains("MSG_")) {                                
                result = result && javaFieldModifiers.contains(JavaModifier.PRIVATE);   
            }                                                                           
            return result;                                                              
        });                                                                             
}                                                                                       
```
`allMatch(..)` returns true if the stream is empty.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Yes, it doesn't for an array list, but it gives reference immutability for the field (not for data inside the field), so it will help in partially assuring immutability.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Immutability condition is being equally applied to all fields, it is just that `MSG_` are allowed to be private but other conditions before that are still applicable to them, it is just they are allowed to be public, if a `MSG_` isn't final, we will get a test failure.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Fields should be private see https://stackoverflow.com/q/28410386/14425335

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: The Javadoc mentions the conditions `isImmutable(..)` checks, but if you insist, I can modify the method.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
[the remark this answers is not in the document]
rnveach: > `{@code testName} => {@link #testName}`
> {@link ImmutabilityTest#testFieldsInStatelessChecksShouldBeImmutable()}

Did we really need `ImmutabilityTest`?

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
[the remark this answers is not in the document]
rnveach: > all the new rules work around @StatelessCheck and @FileStatefulCheck so only checks package was imported.

This is technically missing `TreeWalker` which is a fileset check.

Also `ModuleReflectionUtil` has `isCheckstyleTreeWalkerCheck` and `isFileSetModule`.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
rnveach: > if a MSG_ isn't final

We only uppercase fields that are static/final. So all MSGs have to be final, otherwise they would be `msgXxx`.

I also don't see how this applies to my comments. Private should be separate archunit as immutability doesn't care about modifier.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
rnveach: Resolving.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
rnveach: I insist. The javadoc comments can be a bit simpler now.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
rnveach: It won't if this is all connected to Stateless. Reference immutability won't help when multiple threads are rummaging through this check. If the field is mutable (public or private), multiple threads can come in and read and write to the field at the same time. A field must be completely immutable (regardless of modifiers) or we should throw an violation on it.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
rnveach: https://stackoverflow.com/q/28410386/14425335
> That is ONE reason why it is a good idea to keep the fields private. (**Obviously, this only applies to field whose types are mutable reference types.**)

Fields must be immutable for Stateless, regardless of modifiers.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
rnveach: Ok, I think I got lost somewhere on immutability and not needing `final`. It does need `final` otherwise the value can be swapped out. Let me know if I led you astray in any other comments about this. I still stand by private/public has no connection to immutability. `MSG_` is still immutable since it is a final String.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: I removed the thing about checking for public, private, should be fine now.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done, removed public, private part.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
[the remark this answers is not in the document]
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Done, removed public, private part.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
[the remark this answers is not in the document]
Vyom-Yadav: No, we don't, removed for all.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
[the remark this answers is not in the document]
rnveach: @Vyom-Yadav I was expecting `.checks` to be removed so we can pick up `TreeWalker`.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
rnveach: @Vyom-Yadav This is done, but are we moving the `private` test to a new PR for ArchUnit?

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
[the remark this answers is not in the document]
Vyom-Yadav: We are picking up `TreeWalker`, see
```java
|| ModuleReflectionUtil.isFileSetModule(input.reflect()));
```

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: We can, feel free to edit https://github.com/checkstyle/checkstyle/issues/12099#issuecomment-1233749876, so that we can have a clearer goal.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -0,0 +1,469 @@
...
+        "com.puppycrawl.tools.checkstyle.checks.metrics.ClassDataAbstractionCouplingCheck",
+        "com.puppycrawl.tools.checkstyle.checks.metrics.ClassFanOutComplexityCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.CatchParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.ClassTypeParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.ConstantNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.InterfaceTypeParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.LambdaParameterNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.LocalFinalVariableNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.LocalVariableNameCheck",
+        "com.puppycrawl.tools.checkstyle.checks.naming.MemberNameCheck",
```
rnveach, line 174: Just a conversation while waiting for the last item to be done ( https://github.com/checkstyle/checkstyle/pull/12135#discussion_r968698797 ),

> MemberNameCheck

I took a look at this, and this class has no annotation on it. It extends `AbstractAccessControlNameCheck` which also has no annotation on it. It also extends `AbstractNameCheck` which has `@StatelessCheck` on it.

Is this violation solely because the annotation is on parent class, which is 2 level deep?

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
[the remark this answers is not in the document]
rnveach: `TreeWalker` located in `com.puppycrawl.tools.checkstyle`.

I am still seeing `.importPackages("com.puppycrawl.tools.checkstyle.checks")`. which excludes `TreeWalker`.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
rnveach: Added what I understand the `private` check to be as number 6 in the comment recommended.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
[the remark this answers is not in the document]
Vyom-Yadav: My bad, done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Violation is because the class itself has immutable fields (no fields in this case, not the inherited fields) and has no annotation.

rnveach: It has an annotation but not the main class itself. I assume that via reflection the annotation will show up, but I was curious if ArchUnit was not seeing the annotation?

rnveach (approved):

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Vyom-Yadav: Yes, ArchUnit is not seeing it, it also has `@Inherited`, I will report this to ArchUnit team.

Issue #2939, opened 2016-02-19
Annotate all Checks fields that are user defined options


romani:
We have nuance in Checkstyle:
`
All Class fields that are not options should be reseted to initial state as the same instance of check is reused for all files.
`

we need to create special annotation in for checkstyle code to mark all Check fields that are user faced options by it. Create a UT in test part of code to validate all our Checks classes to clean heir fields in `beginTree()` method.

romani:
First discussion was done at #2945 . Change is not complete and abandoned by author, so any contributor is welcome to continue.

rnveach:
> following name for Annotaion
> CheckProperty , as we use "Properties" term already in xdoc html

For XDoc validation, we identify properties from the class' beans. For a field to be considered a bean it must have a setter with the same name as the field.
Since checks are completely isolated, and usually don't have excess setters, why don't we continue using this identification instead of placing an annotation on the field?

> We need additional Annotation - CheckPropertyDependant
> > private String classNameFormat = ".*Exception";
> > private Pattern regexp = Pattern.compile(classNameFormat);

If again we wanted to skip adding an annotation, we could name the dependent field as an extension of the bean.
Ex: rename `regexp` to `classNameFormatRegexp`
It may not be a short name, but it would make it easier to identify and identify which property it goes with.

romani:
> why don't we continue using this identification instead of placing an annotation on the field?

I am fine, as soon as it works :). Annotation was just first idea that had that time. The less updates in Checks the better

romani:
There is another problem thatlean me to annotation, it is about metadata files generation for plugins to ease their live to update their configs after checkstyle library add new property to Check. There is a mail thread with author of eclipse-cs about this, but this is separate problem  and could be addresed sepatately.

rnveach:
Exception at https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/589 demonstrates why `beginTree` specifically needs to be the one to clear the fields and not `finishTree`. We should try to also incorporate this test into this issue.

rnveach:
@romani 
> Create a UT in test part of code to validate all our Checks classes to clean heir fields in beginTree() method.

We are starting to remove these as part of pitest.
Do we have to re-add them and force their existence for pitest using reflection?
Some checks clear out their fields in `leaveToken`, so without forcing reflection I think we would have to build a small army of inputs to verify this behavior, and even then we have to make sure fields will be cleared if an exception inbetween happens.

[...]

PR #12137, opened 2022-08-29
Issue #2939: Add ArchUnit test to ensure properties in modules are annotated

Vyom-Yadav:
#12099
#2939

### Contents of this PR:

1. Adds a new annotation `ModuleProperty`.
2. Verifies that properties are correctly annotated.
3. Modified XmlMetaReader to return Map instead of a List.
4. Added annotation to modules (except `Checker`, `AbstractCheck`, and `AbstractFileSetCheck`)

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/meta/XmlMetaReader.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/meta/XmlMetaReader.java
@@ -22,8 +22,10 @@ package com.puppycrawl.tools.checkstyle.meta;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
 
@@ -67,10 +69,11 @@ public final class XmlMetaReader {
      *
      * @param thirdPartyPackages fully qualified third party package names(can be only a
      *                           hint, e.g. for SevNTU it can be com.github.sevntu / com.github)
-     * @return list of module details found in the classpath satisfying the above conditions
+     * @return map of module full qualified name to module details found in the classpath satisfying
+     *         the above conditions
      * @throws IllegalStateException if there was a problem reading the module metadata files
      */
-    public static List<ModuleDetails> readAllModulesIncludingThirdPartyIfAny(
+    public static Map<String, ModuleDetails> readAllModulesIncludingThirdPartyIfAny(
             String... thirdPartyPackages) {
         final Set<String> standardModuleFileNames = new Reflections(
                 "com.puppycrawl.tools.checkstyle.meta", Scanners.Resources)
@@ -83,7 +86,7 @@ public final class XmlMetaReader {
             allMetadataSources.addAll(thirdPartyModuleFileNames);
         }
 
-        final List<ModuleDetails> result = new ArrayList<>(allMetadataSources.size());
+        final Map<String, ModuleDetails> result = new HashMap<>(allMetadataSources.size(), 1.0F);
         for (String fileName : allMetadataSources) {
             final ModuleType moduleType;
             if (fileName.endsWith("FileFilter.xml")) {
@@ -104,7 +107,7 @@ public final class XmlMetaReader {
                 throw new IllegalStateException("Problem to read all modules including third "
                         + "party if any. Problem detected at file: " + fileName, ex);
             }
-            result.add(moduleDetails);
+            result.put(moduleDetails.getFullQualifiedName(), moduleDetails);
         }
 
         return result;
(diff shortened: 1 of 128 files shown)


Vyom-Yadav:
This does not check properties like `tokens`, we have a manually annotated and verify, should be easy, we only have a few like these.

src/main/java/com/puppycrawl/tools/checkstyle/CheckProperty.java
```diff
@@ -0,0 +1,14 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation indicates that the field is a check property.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface CheckProperty {
```
rnveach, line 13: I am curious on why you choose an annotation for each type of module instead of a singular one like `ModuleProperty`? Also what about TreeWalker property? Technically users can add listeners, and I believe they can have properties too, so should they be added? What about `Checker`, it isn't a Check nor FileFilter?

They are all properties, and we would know if they are for a check or filter based on it's class' heirarchy.

src/main/java/com/puppycrawl/tools/checkstyle/meta/XmlMetaReader.java
```diff
@@ -83,7 +86,7 @@ public static List<ModuleDetails> readAllModulesIncludingThirdPartyIfAny(
             allMetadataSources.addAll(thirdPartyModuleFileNames);
         }
 
-        final List<ModuleDetails> result = new ArrayList<>(allMetadataSources.size());
+        final Map<String, ModuleDetails> result = new HashMap<>(allMetadataSources.size());
```
rnveach, line 89: @nick-mancuso What is your opinion on this change? I understand why it was made, but I am not sure it is warranted.

rnveach (commented):
Even though this is ArchUnit, I think the issue should be attached to https://github.com/checkstyle/checkstyle/issues/2939

src/main/java/com/puppycrawl/tools/checkstyle/CheckProperty.java
Vyom-Yadav: I was thinking of manually annotating properties in `Checker`, `AbstractCheck`, and `AbstractFileSetCheck` as they are only a handful.

If we want to create a test around this, we would need some source to verify that fields are indeed properties (like metadata of checks). Though metadata does not exist for the modules mentioned above, we have `config.xml` defining properties for the checker.

For `AbstractCheck` and `AbstractFileSetCheck`, we can get fields defined, and scour the metadata of checks extending them, to find if that field is a check property or not.

This approach is time-consuming as we would have to scan all the subclasses because some fields (which are not properties) are not in any metadata file, so every subclass would have to be traversed.

I don't think the hassle mentioned above is worth it, we rarely modify properties in the modules mentioned above.

As far as listeners are concerned, we don't have any documented properties regarding them. So it can't be judged whether their fields should be annotated or not. 

> I am curious about why you choose an annotation for each type of module instead of a singular one like `ModuleProperty`

That change can be done, I did that just to be explicit but am ready to change it if required.

romani: Module property.

romani:
Can we point to https://github.com/checkstyle/checkstyle/issues/2939 ? in commit message

src/main/java/com/puppycrawl/tools/checkstyle/meta/XmlMetaReader.java
nick-mancuso: >This class is used by plugins that need load of metadata from XML files.

We cannot change API of this utility if plugins depend on it. However, I would consider allowing class to be extended for our purposes. @rnveach thoughts?

src/main/java/com/puppycrawl/tools/checkstyle/CheckProperty.java
Vyom-Yadav: > Module property.

Done.

Vyom-Yadav:
> Can we point to #2939 ? in commit message

Done.

src/main/java/com/puppycrawl/tools/checkstyle/meta/XmlMetaReader.java
rnveach: https://github.com/checkstyle/eclipse-cs/blob/f877da5251493164033ee65cf71a19dd9904d533/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/meta/MetadataFactory.java#L582
Eclipse-cs is the only registered user right now as I know. Sevntu was attempting to integrate this but has been paused. So public method can't be changed.

I believe Extending class is ok, but be this is an open issue to change the signature some, so it could impact this test some. Also just know this method call isn't a read only, it actually writes files. I know this because I created an open issue about it. Also if we are extending, I would want more than just wrapping the existing public. If we are going to wrap the existing public, then really the code could just be in this test.

Vyom-Yadav:
We are not annotating fields. Other way is present in #12135

PR #12138, opened 2022-08-29
Issue #12116: update publish Twit action to use secrets

### LINKED ISSUE #12116: update publish Twit action to use secrets
https://docs.github.com/en/actions/security-guides/encrypted-secrets#example-using-bash
need special referencing of secret 

https://github.com/checkstyle/checkstyle/pull/12110 introduced action but no referencing.
secrets will be named same as bash environment variable. 

Please show testing prove that you have same secrets defined in your repo and it works by action. 

example of failure: https://github.com/checkstyle/checkstyle/runs/8059289465?check_suite_focus=true

```
Run ./.ci/tweet-releasenotes.sh 10.3.3
Error: Define TWITTER_CONSUMER_KEY environment variable
```

Rahulkhinchi03:
Fixes #12116


Rahulkhinchi03:
GitHub Action: https://github.com/Rahulkhinchi03/checkstyle/runs/8074457595?check_suite_focus=true

sdvsdvsdv

Rahulkhinchi03:
@romani Please note that: GITHUB_TOKEN or READ_ONLY_TOKEN will not work for twitter. We need to use PERSONAL_ACCESS_TOKEN only.

Testing with GITHUB_TOKEN or READ_ONLY_TOKEN gives permission error.

Testing with GitHub TOKEN: `FAIL` : [ACTION](https://github.com/Rahulkhinchi03/checkstyle/runs/8073787642?check_suite_focus=true)

Reason: `Resource not accessible by integration`

.github/workflows/publish_releasenotes_for_twitter.yml
```diff
@@ -18,4 +18,10 @@ jobs:
         uses: actions/checkout@v2
       - name: Run Shell Script
         run: |
-            ./.ci/tweet-releasenotes.sh ${{ github.event.inputs.version }}
+          ./.ci/tweet-releasenotes.sh ${{ github.event.inputs.version }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
romani, line 23: What happens if you remove this line ?
We do not need redefine GitHub token.

romani (changes requested):
Items

.github/workflows/publish_releasenotes_for_twitter.yml
Rahulkhinchi03: Without redefining the GitHub Token and removing this line from workflow: It [FAILS](https://github.com/Rahulkhinchi03/checkstyle/runs/8084952250?check_suite_focus=true)

Rahulkhinchi03: Also found that we do not need to create a secret for PERSONAL_ACESS_TOKEN.

Using `${{ secrets.PERSONAL_ACCESS_TOKEN }}` will automatically creates the token and uses it so we do not need to create a secret for it.

Rahulkhinchi03: The code is updated only.

romani: ok, update code with usage of `GITHUB_READONLY_TOKEN` env variable. We will put there personal token for now.

romani: update tweet-releasenotes.sh to use GITHUB_READONLY_TOKEN to make it clear that it is not general GITHUB_TOKEN

Rahulkhinchi03: Changes Done

Rahulkhinchi03:
@romani Using the token here: [TOKEN](https://github.com/checkstyle/checkstyle/blob/master/.ci/tweet-releasenotes.sh#L34) is not necessary and causes problems of string sometimes. 

I have used it many times and without a token, my Request never got canceled or skipped.

romani (approved):
Ok let's try

PR #12149, opened 2022-09-01
Issue #12099: Add ArchUnit test to ensure classes in api package are not dependent on classes present in utils package

Vyom-Yadav:
#12099

### This PR contains:

1. A rule stating that classes in the API package are not dependent on classes in the utils package.
2. Suppression list containing classes that don't follow this rule.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
@@ -19,7 +19,11 @@
 
 package com.puppycrawl.tools.checkstyle.internal;
 
+import static com.google.common.truth.Truth.assertWithMessage;
 import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.methods;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.noClasses;
+
+import java.util.List;
 
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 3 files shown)


nrmancuso:
@Vyom-Yadav please make CI happy

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
```diff
@@ -74,4 +95,35 @@ public void nonProtectedCheckMethodsTest() {
...
+        };
+
+        final ArchRule testClassesInApiDoNotDependOnClassesInUtil = noClasses()
+            .that(are(not(suppressed)))
+            .should()
+            .dependOnClassesThat()
+            .resideInAnyPackage("com.puppycrawl.tools.checkstyle.utils",
+                                "com.puppycrawl.tools.checkstyle.checks.javadoc.utils");
+
+        testClassesInApiDoNotDependOnClassesInUtil.check(importedClasses);
```
nrmancuso, line 127: Can we add ArchRule#check to valid assert methods?

nrmancuso (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
```diff
@@ -74,4 +95,35 @@ public void nonProtectedCheckMethodsTest() {
...
+            .importPackages("com.puppycrawl.tools.checkstyle.api");
+
+        final DescribedPredicate<JavaClass> suppressed = new DescribedPredicate<>("suppressed") {
+            @Override
+            public boolean apply(JavaClass input) {
+                return API_CLASSES_DEPENDENT_ON_UTILS.contains(input.getFullName());
+            }
+        };
+
+        final ArchRule testClassesInApiDoNotDependOnClassesInUtil = noClasses()
```
nrmancuso, line 120: ArchRule name should read like a rule, not a test: `final ArchRule ClassesInApiDoNotDependOnClassesInUtil`

Our rule is that "Classes in Api (package) Do Not Depend On Classes In Util (package)".

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
```diff
@@ -32,6 +39,20 @@
 
 public class ArchUnitTest {
 
+    /**
+     * List of classes that don't follow the rule
+     * {@code testClassesInApiDoNotDependOnClassesInUtil}.
+     */
+    private static final Set<String> API_CLASSES_DEPENDENT_ON_UTILS = Set.of(
```
nrmancuso, line 46: @Vyom-Yadav do you think it would be better to use `Set<Class<?>>`? This might help us to find problems in the future if we remove/ move/ change a class.

nrmancuso (changes requested):
Items:

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
Vyom-Yadav: There were two reasons why I didn't go for that:

1. It will add a lot of imports if the suppression list is long.
2. Not all classes are public, we won't be able to import them.
(ex - `com.puppycrawl.tools.checkstyle.api.AutomaticBean$PatternConverter`)

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
Vyom-Yadav: Good point, changed to `noClassesInApiShouldDependOnClassesInUtil`

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
Vyom-Yadav: Done.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
nrmancuso: Both good points, agreed.

nrmancuso (approved):
It is good to have this more granular suppression model for the future.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
```diff
@@ -74,4 +120,36 @@ public void nonProtectedCheckMethodsTest() {
 
         checkMethodsShouldNotBeProtectedRule.check(importedClasses);
     }
+
+    /**
+     * The goal is to ensure all classes in api package are not dependent on classes in util
+     * packages.
+     */
```
baratali, line 127: Can you please add a reason to the comment on why we need such rule?
It's not very obvious.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
```diff
@@ -74,4 +120,36 @@ public void nonProtectedCheckMethodsTest() {
...
+        final ArchRule classShouldNotDependOnUtilPackages = noClasses()
+            .should()
+            .dependOnClassesThat()
+            .resideInAnyPackage(utilPackages);
+
+        final EvaluationResult result = classShouldNotDependOnUtilPackages.evaluate(apiPackage);
+        final EvaluationResult filtered = result.filterDescriptionsMatching(description -> {
+            return API_PACKAGE_SUPPRESSION_DETAILS.stream()
+                .noneMatch(description::startsWith);
+        });
```
baratali, line 148: also it would be good to print useless suppressions, so it forces us to remove them if any.
but it can be implemented later, no need to do it right now :)

baratali (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
Vyom-Yadav: Let's implement it later, and do it in a separate issue where we would also make better formatting of errors if any.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
Vyom-Yadav: Added.

baratali (approved):

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
romani: @Vyom-Yadav , please create issue on this to not forget.
One day we will find contributor to make it happen.

pbludov (approved):

src/test/java/com/puppycrawl/tools/checkstyle/internal/ArchUnitTest.java
nrmancuso: @Vyom-Yadav did we create issue on this?

Vyom-Yadav: Created at https://github.com/checkstyle/checkstyle/issues/12189

PR #12154, opened 2022-09-02
Pull #12154: Use OS specific command and correct errorExtractingPattern

Vyom-Yadav:
#### 1. Usage of OS specific command: (Exception without this modification in Windows OS) (Bug Fix)
```
Caught: java.io.IOException: Cannot run program "mvn": CreateProcess error=2, The system cannot find the file specified
java.io.IOException: Cannot run program "mvn": CreateProcess error=2, The system cannot find the file specified
        at error-prone-check.getErrorProneErrors(error-prone-check.groovy:87)
        at error-prone-check.checkErrorProneReport(error-prone-check.groovy:63)
        at error-prone-check.parseArgumentAndExecute(error-prone-check.groovy:41)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at error-prone-check.run(error-prone-check.groovy:19)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
Caused by: java.io.IOException: CreateProcess error=2, The system cannot find the file specified
        ... 10 more
```
#### 3. Removed redundant escapes (`\\]`) (Enhancement)

### DIFF
--- a/.ci/error-prone-check.groovy
+++ b/.ci/error-prone-check.groovy
@@ -6,9 +6,9 @@ import groovy.xml.XmlUtil
 import java.util.regex.Matcher
 import java.util.regex.Pattern
 
-@Field static final String SEPARATOR = System.getProperty("file.separator")
 @Field static final Set<String> PROFILES = Set.of("compile", "test-compile")
-@Field static final String USAGE_STRING = "Usage groovy ./.ci/error-prone-check.groovy" +
+@Field static final String USAGE_STRING = "Usage groovy " +
+        ".${File.separator}.ci${File.separator}error-prone-check.groovy" +
         " (compile | test-compile) [-g | --generate-suppression]\n"
 
 int exitCode = 1
@@ -58,8 +58,8 @@ private int parseArgumentAndExecute(String argument, String flag) {
 private static int checkErrorProneReport(String profile, String flag) {
     final XmlParser xmlParser = new XmlParser()
     final String suppressedErrorsFileUri =
-            ".${SEPARATOR}.ci${SEPARATOR}" +
-                    "error-prone-suppressions${SEPARATOR}${profile}-phase-suppressions.xml"
+            ".${File.separator}.ci${File.separator}" +
+                    "error-prone-suppressions${File.separator}${profile}-phase-suppressions.xml"
     final List<String> errorProneErrors = getErrorProneErrors(profile)
     Set<ErrorProneError> errors = Collections.emptySet()
     if (!errorProneErrors.isEmpty()) {
@@ -84,7 +84,7 @@ private static List<String> getErrorProneErrors(String profile) {
     final List<String> errorProneErrors = [] as ArrayList
     final String command = "mvn -e --no-transfer-progress clean" +
             " ${profile} -Perror-prone-${profile}"
-    final Process process = command.execute()
+    final Process process = getOsSpecificCmd(command).execute()
     process.in.eachLine { line ->
         if (line.startsWith("[ERROR]")) {
             errorProneErrors.add(line)
@@ -96,6 +96,23 @@ private static List<String> getErrorProneErrors(String profile) {
 }
 
 /**
+ * Get OS specific command.
+ *
+ * @param cmd input command
+ * @return OS specific command
+ */
+private static String getOsSpecificCmd(String cmd) {
+    final String osSpecificCmd
+    if (System.getProperty("os.name").toLowerCase().contains('windows')) {
+        osSpecificCmd = "cmd /c ${cmd}"
+    }
+    else {
+        osSpecificCmd = cmd
+    }
+    return osSpecificCmd
+}
+
+/**
  * Get a set of {@link ErrorProneError} from text.
  *
  * @param errorsText errors in text format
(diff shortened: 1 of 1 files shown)


Vyom-Yadav:
Works fine, tested locally on windows.

Vyom-Yadav:
`
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
/home/runner/work/checkstyle/checkstyle/.ci/error-prone-check.groovy: 12: Apparent variable 'SEPARATOR' was found in a static scope but doesn't refer to a local variable, static field or class. Possible causes:
`
There were a lot of bugs fixed related to static fields in `2.5` (See [this](http://groovy-lang.org/changelogs/changelog-2.5.0.html#changelog)). This works correctly in `2.5` but we are using `2.4`, I think we should consider upgrading somehow.

.ci/error-prone-check.groovy
```diff
@@ -8,7 +8,8 @@ import java.util.regex.Pattern
 
 @Field static final String SEPARATOR = System.getProperty("file.separator")
```
nrmancuso, line 9: Why aren't we using `File.separator` ?

.ci/error-prone-check.groovy
```diff
@@ -82,7 +83,14 @@ private static int checkErrorProneReport(String profile, String flag) {
  */
 private static List<String> getErrorProneErrors(String profile) {
     final List<String> errorProneErrors = [] as ArrayList
-    final String command = "mvn -e --no-transfer-progress clean" +
+    final String maven
+    if ("\\".equals(SEPARATOR)) {
+        maven = "mvn.cmd"
```
rnveach, line 88: Why?
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/diff.groovy#L219

Vyom-Yadav: That is a user-defined method:
```groovy
def executeCmd(cmd, dir = new File("").absoluteFile) {
    println "Running command: ${cmd}"
    def osSpecificCmd = getOsSpecificCmd(cmd)
    def proc = osSpecificCmd.execute(null, dir)
    proc.consumeProcessOutput(System.out, System.err)
    proc.waitFor()
    if (proc.exitValue() != 0) {
        throw new GroovyRuntimeException("Error: ${proc.err.text}!")
    }
}
```

.ci/error-prone-check.groovy
Vyom-Yadav: Done.

.ci/error-prone-check.groovy
rnveach: Yes, but it doesn't need to swap from running `mvn` to `mvn.cmd`. Also, I executed this on Windows and it ran fine without the `.cmd`.

Vyom-Yadav: I got `java.io.IOException: Cannot run program "mvn": CreateProcess error=2, The system cannot find the file specified`, are the current changes fine?

romani:
please change commit to `Pull #12154: .....` and explain kin PR description reason on why this update is required.
I am curious why were  ok without in past and who will benefit from this update, and what problem had such people wihtout this update.

nrmancuso (approved):
I am good as long as CI is passing and we have confirmation from @rnveach that this script works. I do not have access to a windows machine, and I do not think it is worth it to add to Windows CI just to prove it.

rnveach:
````
Summary: TotalFiles=3 FilesWithViolations=1 P1=0 P2=0 P3=1

error-prone-check.groovy
    Violation: Rule=UnnecessarySemicolon P=3 Line=124 Msg=[Semi-colons as line endings can be removed safely] Src=[            .compile(".*[\\\\/](.*\\.java):\\[(\\d+).*\\[(\\w+)](.*)");]

[CodeNarc (http://www.codenarc.org/) v1.4]
CodeNarc completed: (p1=0; p2=0; p3=1) 5768ms
````

.ci/error-prone-check.groovy
rnveach: I am good with the new changes and am resolving this.

rnveach (changes requested):
CI failure.

Vyom-Yadav:
> CI failure.

Fixed

romani:
>Usage of OS specific command:

Please share a command and what OS you use.
I still do not understand what problem we have.we never had such problem. In what condition it happens?

Vyom-Yadav:
> > Usage of OS specific command:
> 
> Please share a command and what OS you use. I still do not understand what problem we have.we never had such problem. In what condition it happens?

Only this script uses maven. We never had this problem because the contribution repo is handling it properly. See https://github.com/checkstyle/contribution/blob/1b3e33600898a849791dc88c90df2bb3f80400e7/checkstyle-tester/diff.groovy#L669-L689

You can normally execute `mvn clean verify` in windows, but when you use it like `command.execute()` then it throws and exception.

Vyom-Yadav:
See https://stackoverflow.com/q/9674220/14425335

romani:
Ok it is win os, for new script that we created recently.

rnveach (approved):

PR #12163, opened 2022-09-04
minor: Enable checker framework to see results

Vyom-Yadav:
Enabling [Initialized Fields Checker](https://checkerframework.org/manual/#initialized-fields-checker)

### DIFF
--- a/.github/workflows/checker-framework.yml
+++ b/.github/workflows/checker-framework.yml
@@ -1,0 +1,20 @@
+name: Checker-Framework
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  checker-framework:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Execute checker framework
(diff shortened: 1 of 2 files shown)


Vyom-Yadav:
> The Initialized Fields Checker warns if a constructor does not initialize a field.

Don't think this is really useful for checks as fields are initialized in `beginTree(..)`

nrmancuso:
@Vyom-Yadav what are our options for individual violation suppression of this check? Also, is it possible to restrict the execution of one specific check to certain packages/file paths?

Vyom-Yadav:
@nrmancuso I don't think integrating the initialization checker would be an easy task as we instantiate checks fields by either setter methods (properties) or by using `beginTree(..)`. Also, the check objects are created using reflection by `PackageObjectFactory`. I am going through the documentation to see if there is some way.

nrmancuso:
>I don't think integrating the initialization checker would be an easy task as we instantiate checks fields

Agreed, this is why I asked about restricting this check to certain packages/ file path only. If we can restrict it to api and some other packages only, I think it is ok (although we still use some reflection there, too). If we have to suppress hundreds of errors for checks that we will never fix, I am not sure about the value of this check.

romani:
> api

https://github.com/checkstyle/checkstyle/issues/12164#issuecomment-1236393577

Vyom-Yadav:
Yes, we can.

Issue #12164, opened 2022-09-04
Enable Checker Framework


nrmancuso:
In our relentless pursuit to use all static analysis and code quality tools, we would like to add the [Checker Framework](https://checkerframework.org/) to our project. This framework uses annotations and utilities to enhance type checking and reduce potential defects in code.

Here is the built-in Checkers:

 - [ ] [Nullness Checker](https://checkerframework.org/manual/#nullness-checker) for null pointer errors 
- [ ] [Initialization Checker](https://checkerframework.org/manual/#initialization-checker) to ensure all @NonNull fields are set in the constructor 
- [ ] [Map Key Checker](https://checkerframework.org/manual/#map-key-checker) to track which values are keys in a map 
- [ ] [Optional Checker](https://checkerframework.org/manual/#optional-checker) for errors in using the [Optional](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html) type 
- [ ] [Interning Checker](https://checkerframework.org/manual/#interning-checker) for errors in equality testing and interning 
- [ ] [Called Methods Checker](https://checkerframework.org/manual/#called-methods-checker) for the builder pattern 
- [ ] [Resource Leak Checker](https://checkerframework.org/manual/#resource-leak-checker) for ensuring that resources are disposed of properly 
- [ ] [Fake Enum Checker](https://checkerframework.org/manual/#fenum-checker) to allow type-safe fake enum patterns and type aliases or typedefs 
- [ ] [Tainting Checker](https://checkerframework.org/manual/#tainting-checker) for trust and security errors 
- [ ] [Lock Checker](https://checkerframework.org/manual/#lock-checker) for concurrency and lock errors 
- [ ] [Index Checker](https://checkerframework.org/manual/#index-checker) for array accesses 
- [ ] [Regex Checker](https://checkerframework.org/manual/#regex-checker) to prevent use of syntactically invalid regular expressions 
- [ ] [Format String Checker](https://checkerframework.org/manual/#formatter-checker) to ensure that format strings have the right number and type of % directives 
- [ ] [Internationalization Format String Checker](https://checkerframework.org/manual/#i18n-formatter-checker) to ensure that i18n format strings have the right number and type of {} directives 
- [ ] [Property File Checker](https://checkerframework.org/manual/#propkey-checker) to ensure that valid keys are used for property files and resource bundles 
[...]

PR #12184, opened 2022-09-08
Issue #12164: Enable Optional Checker for detecting errors in using the Optional type

Vyom-Yadav:
#12164 

Enabled the [Optional Checker](https://checkerframework.org/manual/#optional-checker) for detecting errors in using the [Optional](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html) type.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -383,6 +383,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.checkerframework</groupId>
+      <artifactId>checker-qual</artifactId>
+      <version>3.12.0</version>
+    </dependency>
   </dependencies>
 
   <!-- that repositories are required for testing plugin's snapshot version -->
@@ -2380,6 +2385,46 @@
           </plugin>
         </plugins>
       </build>
+    </profile>
+
+    <profile>
+      <id>checker-framework</id>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <version>3.10.1</version>
+            <configuration>
+              <fork>true</fork> <!-- Must fork or else JVM arguments are ignored. -->
+              <compilerArguments>
+                <Xmaxerrs>10000</Xmaxerrs>
+                <Xmaxwarns>10000</Xmaxwarns>
+              </compilerArguments>
+              <annotationProcessorPaths>
+                <path>
+                  <groupId>org.checkerframework</groupId>
+                  <artifactId>checker</artifactId>
+                  <version>3.12.0</version>
+                </path>
+              </annotationProcessorPaths>
+              <annotationProcessors>
+                <!-- Add all the checkers you want to enable here -->
+                <annotationProcessor>org.checkerframework.checker.optional.OptionalChecker</annotationProcessor>
+              </annotationProcessors>
+              <compilerArgs>
+                <!-- <arg>-Awarns</arg> --> <!-- -Awarns turns type-checking errors into warnings. -->
+              </compilerArgs>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+      <dependencies>
+        <dependency>
+          <groupId>org.checkerframework</groupId>
+          <artifactId>checker</artifactId>
+          <version>3.12.0</version>
+        </dependency>
+      </dependencies>
     </profile>
 
     <profile>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java
@@ -34,6 +34,8 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import org.checkerframework.checker.optional.qual.Present;
+
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.transform.TransformerException;
 
@@ -300,7 +302,10 @@ public class JavadocMetadataScraper extends AbstractJavadocCheck {
             final DetailNode propertyNameTag = propertyNameNode.get();
             final String propertyName = getTextFromTag(propertyNameTag);
 
-            final DetailNode propertyType = getFirstChildOfMatchingText(nodeLi, TYPE_TAG)
+            @Present final Optional<DetailNode> firstChildOfMatchingText =
+                getFirstChildOfMatchingText(nodeLi, TYPE_TAG);
+
+            final DetailNode propertyType = firstChildOfMatchingText
                 .orElseThrow(() -> {
                     return new MetadataGenerationException(String.format(
                         Locale.ROOT, PROP_TYPE_MISSING, propertyName)
@@ -323,8 +328,11 @@ public class JavadocMetadataScraper extends AbstractJavadocCheck {
                     validationTypeNode));
             }
 
-            final String defaultValue = getFirstChildOfMatchingText(nodeLi, DEFAULT_VALUE_TAG)
-                .map(defaultValueNode -> getPropertyDefaultText(nodeLi, defaultValueNode))
+            @Present final Optional<String> firstChildOfMatchingTextDefaultText =
+                getFirstChildOfMatchingText(nodeLi, DEFAULT_VALUE_TAG)
+                .map(defaultValueNode -> getPropertyDefaultText(nodeLi, defaultValueNode));
+
+            final String defaultValue = firstChildOfMatchingTextDefaultText
                 .orElseThrow(() -> {
                     return new MetadataGenerationException(String.format(
                         Locale.ROOT, PROP_DEFAULT_VALUE_MISSING, propertyName)
@@ -512,11 +520,13 @@ public class JavadocMetadataScraper extends AbstractJavadocCheck {
      * @param pattern pattern to match against
      * @return the first child node matching the condition
      */
-    private static Optional<DetailNode> getFirstChildOfMatchingText(DetailNode node,
-                                                                    Pattern pattern) {
-        return Arrays.stream(node.getChildren())
+    private static @Present Optional<DetailNode> getFirstChildOfMatchingText(DetailNode node,
+                                                                             Pattern pattern) {
+        @Present final Optional<DetailNode> firstChildOfMatchingText =
+            Arrays.stream(node.getChildren())
                 .filter(child -> pattern.matcher(child.getText()).matches())
                 .findFirst();
+        return firstChildOfMatchingText;
     }
 
     /**
--- a/.github/workflows/checker-framework.yml
+++ b/.github/workflows/checker-framework.yml
@@ -1,0 +1,20 @@
+name: Checker-Framework
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  checker-framework:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Execute checker framework
+        run: mvn -e --no-transfer-progress clean compile -Pchecker-framework


Vyom-Yadav:
```
Error:  /home/runner/work/checkstyle/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java:[304,28] error: [method.invocation.invalid] call to <X>orElseThrow(java.util.function.Supplier<? extends X>) not allowed on the given receiver.
Error:    found   : @MaybePresent Optional</*INFERENCE FAILED for:*/ ? extends Object>
Error:    required: @Present Optional</*INFERENCE FAILED for:*/ ? extends Object>
Error:  /home/runner/work/checkstyle/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java:[328,28] error: [method.invocation.invalid] call to <X>orElseThrow(java.util.function.Supplier<? extends X>) not allowed on the given receiver.
Error:    found   : @MaybePresent Optional</*INFERENCE FAILED for:*/ ? extends Object>
Error:    required: @Present Optional</*INFERENCE FAILED for:*/ ? extends Object>
```

Basically the optional can contain no value and the checker framework doesn't allow calling `orElseThrow(..)` on `@MaybePresent Optional`.
From Docs:
> the Optional Checker will issue a warning if you call [get](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#get()) or [orElseThrow](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#orElseThrow(java.util.function.Supplier)) on a @MaybePresent Optional receiver, because each of these methods throws an exception if the receiver is an absent Optional.

@nrmancuso @romani I don't think this is a bad practice (using `orElseThrow(..)`), I think this is a good practice, better than doing:
```java
if (status.isPresent()) {
    return status.get();
} else {
    throw new Exception(); 
}
```
Thoughts?

nrmancuso:
> I think this is a good practice, better than doing…

Please explain why you think this is a good practice.

Vyom-Yadav:
> > I think this is a good practice, better than doing…
> 
> Please explain why you think this is a good practice.

I meant to say it is better than:
```java
if (status.isPresent()) {
    return status.get();
} else {
    throw new Exception(); 
}
```
throwing exceptions in general is not the best idea, they should be rarely used as part of a program's normal flow.

romani:
I am not sure that extra IF will make code better.

>or [orElseThrow](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#orElseThrow(java.util.function.Supplier)) on a @MaybePresent Optional receiver, because each of these methods throws an exception if the receiver is an absent Optional.

whole idea of this method to throw exception if something is missed.
Can we contact checker maintainers or stackoverflow for details on why it is like this ?  or may be this is defect in Checker ?

romani:
here is details - https://github.com/typetools/checker-framework/issues/2537

src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java
```diff
@@ -300,7 +302,10 @@ private static ModulePropertyDetails createProperties(DetailNode nodeLi) {
             final DetailNode propertyNameTag = propertyNameNode.get();
             final String propertyName = getTextFromTag(propertyNameTag);
 
-            final DetailNode propertyType = getFirstChildOfMatchingText(nodeLi, TYPE_TAG)
+            @Present final Optional<DetailNode> firstChildOfMatchingText =
+                getFirstChildOfMatchingText(nodeLi, TYPE_TAG);
+
+            final DetailNode propertyType = firstChildOfMatchingText
                 .orElseThrow(() -> {
```
nrmancuso, line 309: I am beginning to understand this Checker a bit more... our code does not make sense. If we are guaranteeing that `firstChildOfMatchingText` is present, then we should never need to call `orElseThrow`. Also, when we `map` to some other Optional, the return type of `map` is not `@Present`.

nrmancuso: I agree with this Checker, but we have an issue; compilation is failing, so code must be fixed to activate Optional Checker.

nrmancuso: After applying this patch:
```
➜  checkstyle git:(pull-12184) ✗ git diff | cat
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java b/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java
index fb54dba9c..cd8d99c40 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java
@@ -37,6 +37,8 @@ import java.util.stream.Collectors;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.transform.TransformerException;
 
+import org.checkerframework.checker.optional.qual.Present;
+
 import com.puppycrawl.tools.checkstyle.api.DetailNode;
@@ -300,12 +302,12 @@ public class JavadocMetadataScraper extends AbstractJavadocCheck {
             final DetailNode propertyNameTag = propertyNameNode.get();
             final String propertyName = getTextFromTag(propertyNameTag);
 
-            final DetailNode propertyType = getFirstChildOfMatchingText(nodeLi, TYPE_TAG)
-                .orElseThrow(() -> {
-                    return new MetadataGenerationException(String.format(
-                        Locale.ROOT, PROP_TYPE_MISSING, propertyName)
-                    );
-                });
+            final DetailNode propertyType = getFirstChildOfMatchingText(nodeLi, TYPE_TAG);
+            if (propertyType == null) {
+               throw new MetadataGenerationException(String.format(
+                                        Locale.ROOT, PROP_TYPE_MISSING, propertyName));
+            }
+
             final String propertyDesc = DESC_CLEAN.matcher(
                     constructSubTreeText(nodeLi, propertyNameTag.getIndex() + 1,
                             propertyType.getIndex() - 1))
@@ -315,21 +317,23 @@ public class JavadocMetadataScraper extends AbstractJavadocCheck {
             modulePropertyDetails.setName(propertyName);
             modulePropertyDetails.setType(getTagTextFromProperty(nodeLi, propertyType));
 
-            final Optional<DetailNode> validationTypeNodeOpt = getFirstChildOfMatchingText(nodeLi,
-                VALIDATION_TYPE_TAG);
-            if (validationTypeNodeOpt.isPresent()) {
-                final DetailNode validationTypeNode = validationTypeNodeOpt.get();
+            final DetailNode validationTypeNode = getFirstChildOfMatchingText(nodeLi, VALIDATION_TYPE_TAG);
+            if (validationTypeNode != null) {
                 modulePropertyDetails.setValidationType(getTagTextFromProperty(nodeLi,
                     validationTypeNode));
             }
 
-            final String defaultValue = getFirstChildOfMatchingText(nodeLi, DEFAULT_VALUE_TAG)
-                .map(defaultValueNode -> getPropertyDefaultText(nodeLi, defaultValueNode))
-                .orElseThrow(() -> {
-                    return new MetadataGenerationException(String.format(
-                        Locale.ROOT, PROP_DEFAULT_VALUE_MISSING, propertyName)
-                    );
-                });
+            final DetailNode nodeChild = getFirstChildOfMatchingText(nodeLi, DEFAULT_VALUE_TAG);
+            final String defaultValue;
+            if (nodeChild != null) {
+                defaultValue = getPropertyDefaultText(nodeLi, nodeChild);
+            }
+            else {
+                throw new MetadataGenerationException(
+                        String.format(Locale.ROOT, PROP_DEFAULT_VALUE_MISSING, propertyName)
+                );
+            }
+
             if (!PROPERTIES_TO_NOT_WRITE.contains(defaultValue)) {
                 modulePropertyDetails.setDefaultValue(defaultValue);
             }
@@ -512,11 +516,12 @@ public class JavadocMetadataScraper extends AbstractJavadocCheck {
      * @param pattern pattern to match against
      * @return the first child node matching the condition
      */
-    private static Optional<DetailNode> getFirstChildOfMatchingText(DetailNode node,
+    private static DetailNode getFirstChildOfMatchingText(DetailNode node,
                                                                     Pattern pattern) {
         return Arrays.stream(node.getChildren())
                 .filter(child -> pattern.matcher(child.getText()).matches())
-                .findFirst();
+                .findFirst()
+                .orElse(null);
     }
 
     /**

```

This makes Checker happy other than warnings.

romani: this conflicting with ideas to not use `null` (in some languages it is not a keyword at all, to not let it be used at alll) at all and deal will Optional.
Such refactoring will lead us to remove Optional. Something is wrong here....

Vyom-Yadav: If the goal is to activate the checker anyhow, then `@SuppressWarnings("optional:method.invocation.invalid")` will suppress the violations and build will pass.

nrmancuso: >Such refactoring will lead us to remove Optional. Something is wrong here....

The issue is with the method, since we treat the return value as `@Present` in two cases, and `@MaybePresent` in one case.

nrmancuso: We could also just use optional inline as opposed to the method, and treat each optional separately, with the two that should throw exception as `@Present`

nrmancuso: TLDR;
- if we annotate with `@Present`, absence of value is exceptional
- if we annotate with `@MaybePresent` (or no annotation, as this is default), absence is not exceptional

rnveach: > absence of value is exceptional

Is the "value" in this sentence the possible value inside `Optional` or are we talking about `Optional` itself?

Vyom-Yadav: I don't think annotating with `@Present` and `@MaybePresent` will work here. `java.util.Optional#map(..)` returns an optional of type `@MaybePresent` and `findFirst()` also returns the same. So until we annotate those methods to return of type `@Present` we would get the error.

Vyom-Yadav: > > absence of value is exceptional
> 
> Is the "value" in this sentence the possible value inside `Optional` or are we talking about `Optional` itself?

I think "value" inside the optional.

nrmancuso: > I don't think annotating with `@Present` and `@MaybePresent` will work here. `java.util.Optional#map(..)` returns an optional of type `@MaybePresent` and `findFirst()` also returns the same. So until we annotate those methods to return of type `@Present` we would get the error.

The issue is that we use the return value if the method as two different “types”. We can remove method, inline code, and use the appropriate annotation.

Vyom-Yadav: I am unable to get what you are trying to say, checker will flag even if we use the wrong annotation. eg-
```java
@Present final Optional<String> optional =
        anyThingThatWillReturnOptional(..)
        .map(..);
```
`map(..)` will return `@MaybePresent` and our variable is annotated with `@Present`, the checker will flag this.

romani: Can we suppress without annotation?

romani:
If this check is controversial, can try less problematic rule ? ( Separate PR)

src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java
Vyom-Yadav: We can skip this file by using `-AskipDefs`

nrmancuso: Ok, let’s not block this PR. Let’s:
1. Open an issue for further discussion
2. Add suppression by file mentioned at https://github.com/checkstyle/checkstyle/pull/12184#discussion_r967801525
3. Add comment with “until…” above suppression

romani: Sounds good

pom.xml
```diff
@@ -2382,6 +2387,46 @@
...
+              <annotationProcessorPaths>
+                <path>
+                  <groupId>org.checkerframework</groupId>
+                  <artifactId>checker</artifactId>
+                  <version>3.12.0</version>
+                </path>
+              </annotationProcessorPaths>
+              <annotationProcessors>
+                <!-- Add all the checkers you want to enable here -->
+                <annotationProcessor>org.checkerframework.checker.optional.OptionalChecker</annotationProcessor>
```
romani, line 2412: Please add here all Checkers in comment to let us see all, disabled until issue, so we on activation, move Checker out of comment block.

romani (changes requested):
Items

pom.xml
```diff
@@ -383,6 +383,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.checkerframework</groupId>
+      <artifactId>checker-qual</artifactId>
+      <version>3.12.0</version>
```
nrmancuso, line 389: @Vyom-Yadav why aren't we using newest version?

nrmancuso (changes requested):
Item:

pom.xml
romani: https://checkerframework.org/manual/#installation

3.25

romani: Hmmm, why we are not running our shell script for version check on each PR to prevent such problem?
Is there any objections?

romani: new CI idea is moved to https://github.com/checkstyle/checkstyle/issues/12193

Vyom-Yadav: We aren't using the latest version because of 
```cmnd
[WARNING] 
Dependency convergence error for org.checkerframework:checker-qual:jar:3.12.0:compile paths to dependency are:
+-org.hibernate.search:hibernate-search-build-config:jar:6.2.0-SNAPSHOT
  +-com.puppycrawl.tools:checkstyle:jar:10.3.4-SNAPSHOT:compile
    +-com.google.guava:guava:jar:31.1-jre:compile
      +-org.checkerframework:checker-qual:jar:3.12.0:compile
and
+-org.hibernate.search:hibernate-search-build-config:jar:6.2.0-SNAPSHOT
  +-com.puppycrawl.tools:checkstyle:jar:10.3.4-SNAPSHOT:compile
    +-org.checkerframework:checker-qual:jar:3.25.0:compile

[ERROR] Rule 0: org.apache.maven.plugins.enforcer.DependencyConvergence failed with message:
Failed while enforcing releasability. See above detailed error message.
```

romani: such problems should not block us to use latest version.
We can make version of checker-qual as variable and set it to 3.12.0 before execution of hibernate.
Or we can ask hibernate project to make variable on their side to let us not conflict (we will do this after we become confident that we are going to use checker for sure).

Vyom-Yadav: Problem will occur in our codebase too if we use 3.25.0 without exclusions, I am closing this in favor of #12194, we can continue the discussion over there.

Vyom-Yadav:
Closed in favour of #12194

Issue #12210, opened 2022-09-17
Add method to ignore unstable checker framework violations.


Vyom-Yadav:
From https://github.com/checkstyle/checkstyle/actions/runs/3074040335/jobs/4966560151

link to Suppression :
https://github.com/checkstyle/checkstyle/blob/6335dd84f14a40150e7eb7a037f9a20e5f189236/.ci/checker-framework-suppressions/checker-framework-nullness-optional-interning-suppressions.xml#L2963-L2973

```
New surviving error(s) found:

File Name: "checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/HandlerFactory.java"
Specifier: "argument"
Message: "incompatible argument for parameter constructor of invokeConstructor."
Line Contents: "handlerCtor, indentCheck, ast, parent);"
Line Number: 164
Details: found   : @Initialized @Nullable Constructor&lt;capture#394 extends @Initialized @Nullable Object&gt;
         required: @Initialized @NonNull Constructor&lt;capture#394 extends @Initialized @Nullable Object&gt;
```

https://github.com/checkstyle/checkstyle/blob/6335dd84f14a40150e7eb7a037f9a20e5f189236/.ci/checker-framework-suppressions/checker-framework-nullness-optional-interning-suppressions.xml#L3404-L3414
```
File Name: "checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/HandlerFactory.java"
Specifier: "type.argument"
Message: "incompatible type argument for type parameter T extends Object of invokeConstructor."
Line Contents: "resultHandler = (AbstractExpressionHandler) CommonUtil.invokeConstructor("
Line Number: 163
Details: found   : capture#394[ extends @UnknownKeyFor Object super @KeyForBottom Void]
         required: [extends @UnknownKeyFor Object super @UnknownKeyFor NullType]
```

These errors are unstable because `capture#` keeps changing with every run. Although these errors are present in the current suppression model as `unstable = "true"` but it does not work because unlike pitest unstable suppression model, here the content of the error changes.

Another one is:
https://github.com/checkstyle/checkstyle/blob/6335dd84f14a40150e7eb7a037f9a20e5f189236/.ci/checker-framework-suppressions/checker-framework-methods-resource-fenum-suppressions.xml#L80-L89
```
New surviving error(s) found:

[...]

PR #12211, opened 2022-09-18
minor: do not check details if checker suppression is unstable

nrmancuso:
This is a quick hack until we come up with better idea for #12210

### DIFF
--- a/.ci/checker-framework.groovy
+++ b/.ci/checker-framework.groovy
@@ -1,4 +1,3 @@
-import groovy.transform.EqualsAndHashCode
 import groovy.transform.Field
 import groovy.transform.Immutable
 import groovy.util.slurpersupport.GPathResult
@@ -382,7 +381,6 @@ private static Set<CheckerFrameworkError> setDifference(final Set<CheckerFramewo
 /**
  * A class to represent the XML {@code checkerFrameworkError} node.
  */
-@EqualsAndHashCode(excludes = ["lineNumber", "unstable"])
 @Immutable
 class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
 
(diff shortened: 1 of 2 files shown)


Vyom-Yadav:
Unfortunately, this will not work, we got a new changing variable:

`Message: "@MustCall method close may not have been invoked on temp-var-20910 or any of its aliases."`

`temp-var-\d+` and this one is present in `Message` so ignoring `Details` for unstable errors won't work.

.ci/checker-framework.groovy
```diff
@@ -473,4 +471,112 @@ class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
...
+                details == that.details
+    }
+
+    @Override
+    int hashCode() {
+        int result
+        result = (fileName != null ? fileName.hashCode() : 0)
+        result = 31 * result + (specifier != null ? specifier.hashCode() : 0)
+        result = 31 * result + (message != null ? message.hashCode() : 0)
+        result = 31 * result + (details != null ? details.hashCode() : 0)
```
rnveach, line 508: This won't work if `hashCode` is actually required and used.

> If two objects are equal according to equals() method, then their hash code must be same.

nrmancuso: Good point

nrmancuso: Removed.

rnveach: As long as you override `hashCode` and don't implement similar logic between it and equals, it will be an issue. Removing details and message from just this method won't "Fix" the underlying issue.

Is this hashCode actually used/needed?

nrmancuso: No, I just figured codenarc or something else would have a warning about overriding equals but not hashcode.

nrmancuso: I've removed it, let's see if there are any warnings.

romani:
>temp-var-\d+ and this one is present in Message so ignoring Details for unstable errors won't work.

we can not ignore message, better to apply regexp to it to remove variable part.

.ci/checker-framework.groovy
```diff
@@ -473,4 +471,100 @@ class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
...
+        CheckerFrameworkError that = (CheckerFrameworkError) other
+
+        boolean isEqualExceptDetails = fileName == that.fileName &&
+                lineContent == that.lineContent &&
+                specifier == that.specifier
+
+        if (that.isUnstable()) {
+            // do not compare details for unstable suppression
+            return isEqualExceptDetails &&
+                    isFuzzyMatch(message, that.message)
```
romani, line 493: why we need this ?

can you share expamples of mismatch you try to resolve ?

romani (changes requested):
items:

.ci/checker-framework.groovy
rnveach: https://github.com/checkstyle/checkstyle/blob/master/.ci/checker-framework.groovy#L163
>     final Set<CheckerFrameworkError> errors = new HashSet<>()

It is a hash object. You will need a valid `hashCode` and `equals`.

nrmancuso: > Another question, do we actually have a reason for it to be `equals` or can it be it's own thing like `fuzzyEquals`?

Set.contains relies on equals.

.ci/checker-framework.groovy
romani: https://github.com/checkstyle/checkstyle/issues/12210

nrmancuso:
Closing in favor of https://github.com/checkstyle/checkstyle/pull/12215

Issue #12207, opened 2022-09-16
Debug documentation is not clear


NitishKumar06:
The following link (https://checkstyle.org/idea.html#Debug)  shows how to debug the application. Although, when you look at it carefully, you'll find that the documentation needs to be improved.

The line _Open the Check's source file by double-click on it in a source tree as is shown:_  creates confusion. The image shows that the file that has been made run is present in the `src/main/java/checks`. Althoug when you try to look for the file, you'll find that there's no run option visible on the directory.

The correct directory should be `src/test/java/checks` where you can debug and run the test cases.

Solution: Modify the Debug documentation

NitishKumar06:
I'm working on it. I'd like to know how do we see the changes on our website? In short, how do we run the website locally so that we can see the changes that we implemented?

nrmancuso:
Please see https://github.com/checkstyle/checkstyle/issues/12203#issuecomment-1248034815, you did not need to create this issue. See https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations for how to generate our website.

NitishKumar06:
The rest of the points mentioned are needed to be made fixed upon which I'm working!

romani:
@NitishKumar06 , please send update 
source of page is here - https://github.com/checkstyle/checkstyle/blob/master/src/xdocs/idea.xml

NitishKumar06:
@romani Yes I'm working on it. I'll soon raise a PR as I'm going through other portion of documents as well to keep in mind.

romani:
Fix is merged

PR #12212, opened 2022-09-18
Documentation fixed for the Debug section

NitishKumar06:
Fix: #12207


romani:
Please read CIs error messages and fix problems

NitishKumar06:
@romani Should I go for raising a new PR?

nrmancuso:
> @romani Should I go for raising a new PR?

Please do not close/open PRs, work through problems on your local, make sure that `mvn clean verify` is passing before pushing changes.

romani:
just change commit message and push with force

NitishKumar06:
@romani DONE!

romani:
Please look at CI, you need to rebase on latest code

romani (approved):
Ok to merge

src/xdocs/idea.xml
```diff
@@ -43,7 +43,9 @@
 
     <section name="Debug">
       <p>
-        Open the Check's source file by double-click on it in a source tree as is shown:<br/><br/>
+        Open the Check's source file by double-click on it in a source tree as shown
+        in the figure below. Please note that the path to the file shown in the
+        figure is <code>src/test/java/checks</code><br/><br/>
```
rnveach, line 48: I don't feel this alleviates confusion.
> note that the path to the file shown in the figure 

Image right below still shows main folder and not test. If you read full documentation, this section is saying to open main file to set a breakpoint. It is later down that it says open test and right click that to actually run and debug the test.

I think we need a new first paragraph. Explain this entire section, in short, you need to open main file to set breakpoint you wish and then open test file to run test. In this new paragraph, I would then explain tests are in test folder and main are in main folder. I would then  restore this original paragraph to it's original.

rnveach (changes requested):

src/xdocs/idea.xml
rnveach: I think what your saying is missing is we need another in the list with a picture that basically says "Open the Test's source file by double-click on it in a source tree as is shown:".

I don't think the first image should change as we are first explaining that you need to set a breakpoint in the required place. If you debug without setting a breakpoint, it will just run all the way through and not stop.

If you agree, then feel free to make the change.

NitishKumar06: @rnveach Let me try to present my point in more systematic way.

Documentation is really good, I agree. I went into some confusion and I believe many newbies who will contribute to checkstyle can go into this confusion as well. Here's how:


After the line `Then right-click the corresponding Unit-test file or class definition > Debug "testName"`, there is an image below which shows right clicking on Test file. The confusion that arises here is that: I thought the file is DescendantTokenCheck.java on which debug option is available/breakpoint since at the top corner (see in image) you can see that DescendantTokenCheck.java is opened and not DescendantToeknCheckTest.java

NitishKumar06: checkstyleImage

NitishKumar06: @romani WDYT?

romani: I am fine with change of image

rnveach: @NitishKumar06 I am following the issue.

Your confusion started at `Then right-click the corresponding Unit-test file`. So it doesn't make sense to change things before this line as you said you weren't confused with them and they are the needed pre-cursor to start debugging. There is no point to run in debug mode without a breakpoint.

I still think we need another step in the list with a picture that basically says "Open the Test's source file by double-click on it in a source tree as is shown:". There can be an accompanying picture with it.

NitishKumar06: Yes @rnveach ! This sounds good.

rnveach: @NitishKumar06 I am unsure why you requested a review when my requested changes are not done.

NitishKumar06: Could you please mention which is the right folder to store the pictures as I'm looking forward to modify the picture

rnveach: https://github.com/checkstyle/checkstyle/blob/master/src/xdocs/idea.xml#L48
https://github.com/checkstyle/checkstyle/blob/master/src/site/resources/images/gui_screenshot_select_check_idea.png

NitishKumar06: @rnveach I have done almost the changes you've mentioned. After generating the site locally, I found that the image is too big which won't give a user a good experience. How do I modify this image to have its size the same as the above images?

errorImage

rnveach: Most likely you will have to shrink it in an image editing software.

NitishKumar06: Got it! I'd be applying the changes then. Any more changes apart from that required from your end?

NitishKumar06:
@rnveach I'm trying to squash my commits. When I do `git log` this shows all the commits made by various folks due to which my old commits are present at bottom. Hence I'm not able to squash my commits. Is there any wrong practice that I'm following?

NitishKumar06:
@rnveach @romani There have been conflicts in my branch. I'm closing this PR and here's the new PR raised which will solve this issue: https://github.com/checkstyle/checkstyle/pull/12363

PR #12215, opened 2022-09-18
Issue #12210: Add method to ignore unstable checker framework violations

### LINKED ISSUE #12210: Add method to ignore unstable checker framework violations.

Vyom-Yadav:
Resolves #12210

### DIFF
--- a/.ci/checker-framework.groovy
+++ b/.ci/checker-framework.groovy
@@ -1,4 +1,3 @@
-import groovy.transform.EqualsAndHashCode
 import groovy.transform.Field
 import groovy.transform.Immutable
 import groovy.util.slurpersupport.GPathResult
@@ -382,7 +381,6 @@ private static Set<CheckerFrameworkError> setDifference(final Set<CheckerFramewo
 /**
  * A class to represent the XML {@code checkerFrameworkError} node.
  */
-@EqualsAndHashCode(excludes = ["lineNumber", "unstable"])
 @Immutable
 class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
 
@@ -398,6 +396,21 @@ class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
     List<String> details
     String lineContent
     int lineNumber
+
+    /**
+     * Whether the error is unstable. Unstable errors in suppression list are not flagged as
+     * unnecessary suppressions. An error is considered to be unstable when:
+     * <ul>
+     *     <li>
+     *         Error message changes with each run. Some error messages contains strings like
+     *         {@code temp-var-1234}, the numerical part changes with each run so the error is
+     *         considered unstable. In such cases numerical values in {@code details} and
+     *         {@code message} are replaced with empty string while comparing and hashing errors.
+     *     </li>
+     *     <li>
+     *         The error is in general unstable, it sometimes appears and sometimes it does not.
+     *     </li>
+     */
     boolean unstable
 
     @Override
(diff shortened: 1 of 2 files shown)


romani:
Codenarc failing

Vyom-Yadav:
> Codenarc failing

Done.

.ci/checker-framework.groovy
```diff
@@ -398,6 +396,21 @@ class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
...
+     * unnecessary suppressions. An error is considered to be unstable when:
+     * <ul>
+     *     <li>
+     *         Error message changes with each run. Some error messages contains strings like
+     *         {@code temp-var-1234}, the numerical part changes with each run so the error is
+     *         considered unstable. In such cases numerical values in {@code details} and
+     *         {@code message} are replaced with empty string while comparing and hashing errors.
+     *     </li>
+     *     <li>
+     *         The error is in general unstable, it sometimes appears and sometimes it does not.
```
nrmancuso, line 411: Is this an actual case that we deal with? I thought unstable for checker means that some variable name changes only.

.ci/checker-framework.groovy
```diff
@@ -435,17 +448,96 @@ class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
...
+            i = getDetails().join('') <=> other.getDetails().join('')
+            if (i != 0) {
+                return i
+            }
+        }
+
+        return getLineContent() <=> other.getLineContent()
+    }
+
+    boolean equals(o) {
```
nrmancuso, line 483: Please do o -> object

.ci/checker-framework.groovy
```diff
@@ -398,6 +396,21 @@ class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
     List<String> details
     String lineContent
     int lineNumber
+
+    /**
+     * Whether the error is unstable. Unstable errors in suppression list are not flagged as
```
nrmancuso, line 401: I do not think we need a list here, we can probably just eliminate for sentence and have next sentence lead into explanation in first list item.

nrmancuso (changes requested):
Items:

.ci/checker-framework.groovy
nrmancuso: Also, `@Override`

.ci/checker-framework.groovy
Vyom-Yadav: No, we do not have a case like this, removed it, otherwise would have unnecessarily confused the developer.

.ci/checker-framework.groovy
Vyom-Yadav: Done.

.ci/checker-framework.groovy
Vyom-Yadav: Removed the list.

.ci/checker-framework.groovy
```diff
@@ -435,17 +444,98 @@ class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
...
+        if (this.is(object)) {
+            return true
         }
+        if (!(object instanceof CheckerFrameworkError)) {
+            return false
+        }
+
+        CheckerFrameworkError that = (CheckerFrameworkError) object
+
+        if (this.unstable || that.unstable) {
```
nrmancuso, line 490: Can we just replace everything from here down with `this.compareTo(that) == 0`?

nrmancuso (changes requested):
One more:

nrmancuso (approved):

romani:
Codenarc again

Vyom-Yadav:
> Codenarc again 

Done.

romani (approved):
Ok to merge

.ci/checker-framework.groovy
```diff
@@ -435,17 +444,68 @@ class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
             return i
         }
 
-        i = getMessage() <=> other.getMessage()
-        if (i != 0) {
-            return i
+        if (this.isUnstable() || other.isUnstable()) {
+            final String messageWithoutLineNumber = getMessage().replaceAll('\\d+', '')
+            final String thatMessageWithoutLineNumber = other.getMessage().replaceAll('\\d+', '')
```
rnveach, line 449: Why didn't we do this in the constructor? This is suppose to be an immutable class.

Would also save on execution time, as this gets run probably every time a new entry gets added to the set for nearly everything in the set multiple times.

We obviously don't care about the numbers anyways.

rnveach, line 449: Will numbers that were an issue always be 3 digits? Maybe consider changing the regexp to just match 3 digit numbers.
Also why not match the `#` too?

Vyom-Yadav: > This is suppose to be an immutable class

It still is, the original field is not modified.

> Would also save on execution time, as this gets run probably every time a new entry gets added to the set for nearly everything in the set multiple times.

It will only run for unstable mutations, only a few of them are there.

> We obviously don't care about the numbers anyways.

I feel it would be better if we don't modify the original error itself.

Vyom-Yadav: > Will numbers that were an issue always be 3 digits?

No, they are not, see the updated issue description, we have entries like `temp-var-20910`

>  Also why not match the # too?

Not possible as we have entries like `temp-var-20910`, also if we could do it, there is no point in doing it, would increase the complexity of regex unnecessarily.

Vyom-Yadav:
I think we breach the equals and hashcode contract here, I am investigating it.

rnveach:
If you think this is broken, I recommend adding a print out of full actual so it can be viewed in the logs and debugged easier.

PR #12223, opened 2022-09-22
Pull #12223: Correct the hashCode method in checker-framework.groovy

Vyom-Yadav:
Contents of this PR:

- The `hashCode()` method was not implemented correctly, for two objects `equals()` could return true but the objects had different hashes.
- Now for every error, the hash code is generated from `messageWithoutNumber` and `messageWithoutNumber`.

### DIFF
--- a/.ci/checker-framework.groovy
+++ b/.ci/checker-framework.groovy
@@ -1,5 +1,4 @@
 import groovy.transform.Field
-import groovy.transform.Immutable
 import groovy.util.slurpersupport.GPathResult
 import groovy.util.slurpersupport.NodeChildren
 import groovy.xml.XmlUtil
@@ -381,7 +380,6 @@ private static Set<CheckerFrameworkError> setDifference(final Set<CheckerFramewo
 /**
  * A class to represent the XML {@code checkerFrameworkError} node.
  */
-@Immutable
 class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
 
     /**
@@ -390,12 +388,12 @@ class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
      */
     private static final int LINE_NUMBER_NOT_PRESENT_VALUE = -1
 
-    String fileName
-    String specifier
-    String message
-    List<String> details
-    String lineContent
-    int lineNumber
+    final String fileName
+    final String specifier
+    final String message
+    final List<String> details
+    final String lineContent
+    final int lineNumber
 
     /**
      * Whether the error is unstable. Unstable errors in suppression list are not flagged as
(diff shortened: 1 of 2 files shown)


.ci/checker-framework.groovy
```diff
@@ -390,12 +388,12 @@ class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
...
-    String fileName
-    String specifier
-    String message
-    List<String> details
-    String lineContent
-    int lineNumber
+    final String fileName
+    final String specifier
+    final String message
+    final List<String> details
```
nrmancuso, line 394: Why did we do list instead of multiline string?

nrmancuso (changes requested):
Item:

romani:
I recommend to not mess with hashcode/equals and make special method `compareTo` or somthing like this and use it instead.
We do not need high speed in this script, we need ease of maintenance/change.

nrmancuso:
>I recommend to not mess with hashcode/equals (human is always error prone on this contract) and make special method compareTo or somthing like this and use it instead.

As we use Sets/ TreeSets/ Lists, we must make sure that equals and hashcode are overridden to account for changing messages.

romani:
>As we use Sets/ TreeSets/ Lists, we must make sure that equals and hashcode are overridden

Nobody has to override such methods if objects are immutable or you do not mutate such objects while there in collection.

We just need single place in our code that instead of relying on collection to do search by collection API, do it by our search function that use very specific way of comparison.

rnveach:
> As we use Sets/ TreeSets/ Lists, we must make sure that equals and hashcode are overridden
Nobody has to override such methods if objects are immutable or you do not mutate such objects while there in collection.

Immutability has no play here. Immutable just means the object will not change after creation. I can pass the same values to a constructor twice and have 2 different instances of the "same" inner values. The issue here is determining if 2 different instances are the same. We have to compare the before and after lists, and each one is given their own unique instances. Our class' are not singleton and that is the only way we could get around not overriding equals, is if they amounted to the same instance.

Pitest suppression does the same thing discussed here. The only difference is it doesn't override the hashCode method.
https://github.com/checkstyle/checkstyle/blob/master/.ci/pitest-survival-check-xml.groovy#L316-L317
https://github.com/checkstyle/checkstyle/blob/master/.ci/pitest-survival-check-xml.groovy#L360


> We just need single place in our code that instead of relying on collection to do search by collection API, do it by our search function that use very specific way of comparison.

So you are saying we need to create our version of `contains`/`compareTo`/`equals` which in the end amounts to the same type of code as a  `contains`/`compareTo`/`equals` ? This is a POH tato PAH tato thing, doing our own search is the same as implementing our own `compareTo`/`equals` and using Java's already built utilities. Regardless we have to define when 2 instances are equal to know if they are the same or not.

romani:
Instances of two violations are not same, but we can consider them as same if that is unstable and we want it. But they are not same and equals should give false.

We can merge this PR and in next PR, make it conseptually more robust.

nrmancuso:
I think we should rethink on idea proposed in discord: treat all messages and details the same, i.e. replacement of problematic text by some obvious generic name. Example: `temp-var-123` becomes `'temp-var'`, `capture#123` becomes `'capture'`, etc. This way we only have custom string manipulation method, and once we create suppression object, they are all treated the same.

romani:
I am ok with this

Vyom-Yadav:
> I think we should rethink on idea proposed in discord: treat all messages and details the same, i.e. replacement of problematic text by some obvious generic name. Example: `temp-var-123` becomes `'temp-var'`, `capture#123` becomes `'capture'`, etc. This way we only have custom string manipulation method, and once we create suppression object, they are all treated the same.

That would mean replacing all numbers in details and messages with an empty string. Also, it would be `capture#` and `temp-var-` as creating a regex to remove `#` and `-` is not very reliable as we can have more cases.

nrmancuso:
>we can have more cases.

I think it is better to allow failure for new cases to make us aware of all possible variations and enumerate them in groovy file.

romani:
>That would mean replacing all numbers in details and messages with an empty string.

No just numbers are that has very distinct prefix. Both such numbers are not from our code, and some internal representation of Checker tool.
We can ignore them for now. 
In future as we come to state that we find way to resolve most violations and will plan to resolve all other, we can reconsider our decision update of messages.
We need some stable time to consume this tool and experience it.

rnveach:
@romani @Vyom-Yadav @nrmancuso My recommendation is to drop this PR and proceed with recommendation at https://github.com/checkstyle/checkstyle/issues/12243 .

We should also need to worry less about hash code, and just implement a compare to that will control ordering of violations.

Vyom-Yadav:
> > That would mean replacing all numbers in details and messages with an empty string.
> 
> No just numbers are that has very distinct prefix. Both such numbers are not from our code, and some internal representation of Checker tool. We can ignore them for now. In future as we come to state that we find way to resolve most violations and will plan to resolve all other, we can reconsider our decision update of messages. We need some stable time to consume this tool and experience it.

Agreed, closed in favour of #12244

Issue #12243, opened 2022-10-02
Pitest/Checker: Drop New/Unnecessary Displays And Switch to Git Diff


rnveach:
From trying to work out making CI pass at https://github.com/checkstyle/checkstyle/pull/11617#event-7452762112 I have reached a new level of dislike towards Checker. Since I can't run this on my local, I have to rely on the faulty new/unnecessary display and figure out how to modify the file myself with not good results.

My suggestion is to drop the New/Unnecessary and rely on our git difference to report back any differences, if any, and use this as a way to allow the user to apply the difference patch to their local to pass CI in one go.

This works way more easier for things like arch unit's cycles, which also does not work the same as CI on my local, but I am able to resolve it in one go with the git difference reported by the CI.

I dunno how, but I assume we will still have to continue supporting unstables and recreating the fill on the git difference means keeping that value if it is set. It makes sense for pitest, but I don't remember us having any unstables for checker. This issue will probably have to be taken slowly to ensure it meets our needs.

Vyom-Yadav:
I think the primary problem is the instability with the checker framework, pitest model works smoothly, not sure if git diff is better. 

We can change to `-g` flag in all scripts to get the suppressions in XML, the only drawback is that it won't show line number then because we do not add line number in suppressions, you  would have to search by text in the file.

rnveach:
I am open to anything to make this a better experience.

nrmancuso:
@rnveach to be clear, the goal of this issue is to have each groovy script generate/remove suppressions of violations in their respective files, and rely on git diff to fail build in CI?

rnveach:
Yes. This also provides the benefit that user can copy git diff and apply it to their local (mainly for checker) if all suppressions will stay.

I am open to other options like removing pitest from this, but I have really had enough trying to get checker to pass.

romani:
Git diff can be done in addition to embedded ability to provide "patch".

romani:
Issue is not finished, we have one more PR for it. Based on agreement in private chat we need to keep unstable feature just in case.
And improve new items that we found

nrmancuso:
> Issue is not finished, we have one more PR for it. Based on agreement in private chat we need to keep unstable feature just in case. And improve new items that we found [#12244 (comment)](https://github.com/checkstyle/checkstyle/pull/12244#issuecomment-1272559147)

I am on it

nrmancuso:
[...]

Issue #12230, opened 2022-09-26
checker-framework.groovy (generate): false differences when run locally


rnveach:
When I run `groovy .ci/checker-framework.groovy checker-framework-nullness-optional-interning -g` locally on a clean master, I get the following differences (trimmed due to size) when I don't expect any:
````
diff --git a/.ci/checker-framework-suppressions/checker-framework-nullness-optional-interning-suppressions.xml b/.ci/checker-framework-suppressions/checker-framework-nullness-optional-interning-suppressions.xml
index 775628e..292dbf7 100644
--- a/.ci/checker-framework-suppressions/checker-framework-nullness-optional-interning-suppressions.xml
+++ b/.ci/checker-framework-suppressions/checker-framework-nullness-optional-interning-suppressions.xml
@@ -1,6 +1,2499 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedErrors>
   <checkerFrameworkError unstable="false">
+    <fileName>checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\Checker.java</fileName>
+    <specifier>argument</specifier>
+    <message>incompatible argument for parameter moduleId of Violation.</message>
+    <lineContent>new String[] {ioe.getMessage()}, null, getClass(), null));</lineContent>
+    <details>
+      found   : null (NullType)
+    </details>
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
+    <fileName>checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\Checker.java</fileName>
+    <specifier>method.invocation</specifier>
+    <message>call to addListener(com.puppycrawl.tools.checkstyle.api.AuditListener) not allowed on the given receiver.</message>
+    <lineContent>addListener(counter);</lineContent>
+    <details>
+      found   : @UnderInitialization(com.puppycrawl.tools.checkstyle.api.AutomaticBean.class) @NonNull Checker
+    </details>
+  </checkerFrameworkError>
+
....
 
@@ -5660,10 +8129,10 @@
     <fileName>checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java</fileName>
     <specifier>return</specifier>
     <message>incompatible types in return.</message>
-    <lineContent>return uri;</lineContent>
+    <lineContent>return CommonUtil.class.getResource(name);</lineContent>
     <details>
-      type of expression: @Initialized @Nullable URI
[...]

PR #12244, opened 2022-10-02
Pull #12244: Replace numbers in details and message in checker-framework.groovy

Vyom-Yadav:
This PR contains:

- Replaces `temp-var-\\d+` with `temp-var` in `message` and `capture#\\d+` with `capture` in `details`

### DIFF
--- a/.ci/checker-framework-suppressions/checker-framework-formatter-suppressions.xml
+++ b/.ci/checker-framework-suppressions/checker-framework-formatter-suppressions.xml
@@ -1,27 +1,27 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedErrors>
-  <checkerFrameworkError unstable="false">
+  <checkerFrameworkError>
     <fileName>checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/LocalizedMessage.java</fileName>
     <specifier>i18nformat.key.not.found</specifier>
     <message>a key doesn&apos;t exist in the provided translation file</message>
     <lineContent>final String pattern = resourceBundle.getString(key);</lineContent>
   </checkerFrameworkError>
 
-  <checkerFrameworkError unstable="false">
+  <checkerFrameworkError>
     <fileName>checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java</fileName>
     <specifier>format.string</specifier>
     <message>invalid format string (is a @Format annotation missing?)</message>
     <lineContent>Locale.ROOT, PROP_DEFAULT_VALUE_MISSING, propertyName)</lineContent>
   </checkerFrameworkError>
 
-  <checkerFrameworkError unstable="false">
+  <checkerFrameworkError>
     <fileName>checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java</fileName>
     <specifier>format.string</specifier>
     <message>invalid format string (is a @Format annotation missing?)</message>
     <lineContent>Locale.ROOT, PROP_TYPE_MISSING, propertyName)</lineContent>
   </checkerFrameworkError>
 
-  <checkerFrameworkError unstable="false">
+  <checkerFrameworkError>
     <fileName>checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java</fileName>
     <specifier>i18nformat.key.not.found</specifier>
     <message>a key doesn&apos;t exist in the provided translation file</message>
--- a/.ci/checker-framework-suppressions/checker-framework-index-suppressions.xml
+++ b/.ci/checker-framework-suppressions/checker-framework-index-suppressions.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedErrors>
-  <checkerFrameworkError unstable="false">
+  <checkerFrameworkError>
     <fileName>checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/AstTreeStringPrinter.java</fileName>
     <specifier>argument</specifier>
     <message>incompatible argument for parameter endIndex of substring.</message>
@@ -11,7 +11,7 @@
     </details>
   </checkerFrameworkError>
 
-  <checkerFrameworkError unstable="false">
+  <checkerFrameworkError>
     <fileName>checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/AuditEventDefaultFormatter.java</fileName>
     <specifier>argument</specifier>
     <message>incompatible argument for parameter capacity of StringBuilder.</message>
@@ -22,7 +22,7 @@
     </details>
   </checkerFrameworkError>
 
-  <checkerFrameworkError unstable="false">
+  <checkerFrameworkError>
     <fileName>checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java</fileName>
     <specifier>argument</specifier>
     <message>incompatible argument for parameter bitIndex of get.</message>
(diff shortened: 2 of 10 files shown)


romani:
@rnveach , please confirm execution on your local

rnveach:
@Vyom-Yadav Please specify in the first post what issues this could be resolving (either fully or partially) so atleast we can start keeping track. If it resolved something partially, then please specify which part.

rnveach:
@romani How do you wish me to confirm this?

PR https://github.com/checkstyle/checkstyle/pull/12223 was dropped which is causing weird issues at any time and in all of my PRs that I am trying to work on. There are still false differences at https://github.com/checkstyle/checkstyle/issues/12230 which this PR still gives (114 new survivings, 450 unnecessary). I was hoping we could drop the comparing issue and just go with https://github.com/checkstyle/checkstyle/issues/12243 .

Vyom-Yadav:
@rnveach @romani @nrmancuso The problem with windows is the sequence of output, from logs in local:
```
[WARNING] D:\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\imports\ImportControlLoader.java:[185,78] [argument] incompatible argument for parameter parent of FileImportControl.
  found   : @Initialized @Nullable PkgImportControl
[WARNING]
  required: @Initialized @NonNull PkgImportControl
```

Whereas it should be:
```
[WARNING] D:\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\imports\ImportControlLoader.java:[185,78] [argument] incompatible argument for parameter parent of FileImportControl.
  found   : @Initialized @Nullable PkgImportControl
  required: @Initialized @NonNull PkgImportControl
[WARNING]................
```

As the script is based on the output, we get a lot of errors, tried a bunch of different solutions, but none worked so far.
Problematic code:
https://github.com/checkstyle/checkstyle/blob/fa6856985e3eb32f11ae63e56c19fc2181a50d06/.ci/checker-framework.groovy#L109-L114

Few resources:

- https://stackoverflow.com/questions/4959548/why-does-output-appear-in-wrong-order
- https://stackoverflow.com/questions/4216626/why-is-the-output-from-subprocess-in-wrong-order

Tried:
- https://wiki.sei.cmu.edu/confluence/display/java/FIO07-J.+Do+not+let+external+processes+block+on+IO+buffers
- https://leo3418.github.io/2021/06/20/java-processbuilder-stdout.html

rnveach:
Please atleast do the other things I mentioned.

> As the script is based on the output, we get a lot of errors, tried a bunch of different solutions, but none worked so far.
Problematic code:

I think one issue is the way you are going against the output.

First, you don't really specify a start/end for capturing output. You are just grabbing everything. I wouldn't start capturing until you see `[INFO] --- maven-compiler-plugin`. Maybe hardcode some removals of `[INFO] Changes detected` and `[INFO] Compiling `.

Second, any `[WARNING]` should be removed from the start of the line regardless. I think you may be doing this, but this shouldn't be dependent on anything and just be blind.

When you do the above, you should be left with things like:
````
D:\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\imports\ImportControlLoader.java:[185,78] [argument] incompatible argument for parameter parent of FileImportControl.
  found   : @Initialized @Nullable PkgImportControl
  required: @Initialized @NonNull PkgImportControl
````

Then just identify the "first" line as anything with a file path, and anything else is a consecutive line. You may have to remove extra spaces as I think I was seeing some differences in that regards.

**Comment by Vyom-Yadav** [comment]
> Please atleast do the other things I mentioned.

That is an enhancement and will not solve the current issue you are facing. #12243 just asks for  script to automatically add/remove suppressions and rely on git diff to pass/fail the build.

The real issue, which is making the checker unstable for windows is what is mentioned above.

**Comment by rnveach** [comment]
I was referring to my comment at https://github.com/checkstyle/checkstyle/pull/12244#issuecomment-1272412947to specify what is issues are being partially/fully covered here.

**Comment by Vyom-Yadav** [comment]
> I was referring to my comment at https://github.com/checkstyle/checkstyle/pull/12244#issuecomment-1272412947to specify what is issues are being partially/fully covered here.

It only covers two things, converting `\` to `/` internally, and removing numbers from `temp-var` and `capture`. The former is windows specific, and the latter one applies to any OS.

**Comment by nrmancuso** [comment]
@Vyom-Yadav why are we forced to couple windows support update with other change? “And” in PR description is a red flag to me. Discussion around implementation/ concept around suppression model here is getting messy, if we can separate out windows support changes, let’s do this.

**Comment by romani** [comment]
@Vyom-Yadav , there is conflict, please rebase.

**Comment by romani** [comment]
@Vyom-Yadav , please proceed with update, a bunch of stuff is blocked by this PR

**Comment by Vyom-Yadav** [comment]
@nrmancuso @rnveach @romani Doing more investigation, I don't think this is a script-related issue, it's more likely to be an issue with windows or maven.

Running:
```
mvn -e --no-transfer-progress clean compile -Pchecker-framework-nullness-optional-interning
```
in the cmd prompt, I still get the output:
```
[WARNING] D:\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\imports\ImportControlLoader.java:[185,78] [argument] incompatible argument for parameter parent of FileImportControl.
  found   : @Initialized @Nullable PkgImportControl
[WARNING]
  required: @Initialized @NonNull PkgImportControl
```

instead of:
```
[WARNING] D:\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\imports\ImportControlLoader.java:[185,78] [argument] incompatible argument for parameter parent of FileImportControl.
  found   : @Initialized @Nullable PkgImportControl
  required: @Initialized @NonNull PkgImportControl
[WARNING]................
```

Vyom-Yadav:
Complete log at https://gist.githubusercontent.com/Vyom-Yadav/f2d88a6e765e3ad8547012d2c82b6837/raw/1476e8baa11461808f5e9011e61ae4a0a075e1d2/log.txt

rnveach:
This PR needs to be split up more into pieces we need, and then point those pieces to their respective issues and say how complete the fix is. Windows support will require https://github.com/checkstyle/checkstyle/pull/12244#issuecomment-1272559147 to fix line handling.

For the first PR we need the `Replaces temp-var-\\d+ with temp-var in message and capture#\\d+ with capture in details` in its own PR to merge quickly. We need these values to be replaced before we put them into Details class so they are written to the XML file.

rnveach:
@Vyom-Yadav Can we just trim this PR to just only the replaces? I noticed there is still other things like changing the hash, and removal of unstables.

Vyom-Yadav:
> @Vyom-Yadav Can we just trim this PR to just only the replaces? I noticed there is still other things like changing the hash, and removal of unstables.

Unstable and custom hashing methods were added for handling `capture#` and `temp-var-` only, they can be safely removed.

Vyom-Yadav:
> @nrmancuso @rnveach @romani Doing more investigation, I don't think this is a script-related issue, it's more likely to be an issue with windows or maven.

https://issues.apache.org/jira/browse/MNG-7567

romani:
it is known problem of logging in all places, it might be unsync.
It is much better to not use such builds outputs and use reports that are synced.

romani:
if behavior of logging is not deterministic .... and only on windows
we can have support only for Linux for now, people from windown will use our patch generation in CI.
Please take this from https://github.com/checkstyle/checkstyle/pull/12294

Looks like we need to sort all violations before saving them to file to have predictable behavior. CI will be our source of truth for what suppression file should looks like.

romani (approved):
let merge this  first, it is step forward.

romani:
ok, after discussion in chat with rnveach.

@Vyom-Yadav , please split this PR in few steps and please clearly define what you are fixing.
this Checker adventure becomes messy .... we need to record all details.

romani:
@Vyom-Yadav , if you are around please rebase and resolve conflicts. Please try to keep unstable feature in script, it might be unused now, but it might b used in future.

Vyom-Yadav:
The patch for removing numbers from message and detail has been applied at https://github.com/checkstyle/checkstyle/pull/12294

romani:
@Vyom-Yadav, please recreate your PR with removal of hashcode override and window s support. We still need such updates.

Issue #5168, opened 2017-10-02
Upgrade XML logger to XML 1.1


rnveach:
Identified at https://github.com/checkstyle/checkstyle/pull/5163#issuecomment-333160165 ,

We need to upgrade our XML output to 1.1 to support escaping non-printable characters like hex 0x01 and 0x02 found in openjdk projects.

Before we can upgrade though, we must have maven-checkstyle-plugin ugpgrade their support of this as they won't accept the newer version of XML.
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (default-cli) on project postgresql: Unable to read Checkstyle results xml: /pipeline/source/pgjdbc/pgjdbc/target/checkstyle-result.xml: only 1.0 is supported as <?xml version not '1.1' (position: START_DOCUMENT seen <?xml version="1.1"... @1:19) -> [Help 1]

rnveach:
I have created the issue with maven-checkstyle-plugin at https://issues.apache.org/jira/browse/MCHECKSTYLE-342 .

romani:
@rnveach , I think it is better to close this PR and wait for plugin to resolve issue.

Brijeshthummar02:
@romani should we wait, or update it?

romani:
We can recreate PR to see if issue is resolved

rnveach:
This is exactly what this issue is for, like others we have, to wait for something external to change or for when we break from maven ourselves.

PR #12247, opened 2022-10-02
Issue #5168: upgrades xml output to version 1.1

rnveach:
Issue #5168


rnveach:
Maven plugin has still not resolved the issue

Issue #12251, opened 2022-10-03
Resolve Pitest suppression for AbstractParenPadCheck


Kevin222004:
Related to https://github.com/checkstyle/checkstyle/issues/7797 

 kill mutation 
https://github.com/checkstyle/checkstyle/blob/de37e02b799305fa9309f162f8961c4297fce865/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml#L17
for AbstractParenPadCheck

romani:
Do you understand what mutation is actually done to code?

Kevin222004:
> Do you understand what mutation is actually done to code?
> 
> https://pitest.org/quickstart/mutators/#EXPERIMENTAL_NAKED_RECEIVER

yes I have described full detail in thread

nrmancuso:
Closed via #12258

PR #12258, opened 2022-10-06
Issue #12251: Resolve Pitest suppression for AbstractParenPadCheck

Kevin222004:
Issue #12251: Resolve Pitest suppression for AbstractParenPadCheck

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheckTest.java
@@ -147,6 +147,14 @@ public class ParenPadCheckTest
     }
 
     @Test
+    public void testTrimOptionProperty() throws Exception {
+        final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
+        verifyWithInlineConfigParser(
+                getPath("InputParenPadToCheckTrimFunctionInOptionProperty.java"),
+               expected);
+    }
+
+    @Test
     public void testNospaceWithComplexInput() throws Exception {
         final String[] expected = {
             "55:26: " + getCheckMessage(MSG_WS_FOLLOWED, "("),
--- a/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml
@@ -10,15 +10,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>AbstractParenPadCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.whitespace.AbstractParenPadCheck</mutatedClass>
-    <mutatedMethod>setOption</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to java/lang/String::trim with receiver</description>
-    <lineContent>option = PadOption.valueOf(optionStr.trim().toUpperCase(Locale.ENGLISH));</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>EmptyForInitializerPadCheck.java</sourceFile>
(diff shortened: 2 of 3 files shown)


Kevin222004:
1. pitest report 
https://kevin222004.github.io/BeforeChanges/com.puppycrawl.tools.checkstyle.checks.whitespace/AbstractParenPadCheck.java.html
Surviving Mutation: ```java/lang/String::trim with receiver → SURVIVED``` at line 80

2. Mutated code 
our Surviving mutation is ```java/lang/String::trim with receiver → SURVIVED```
this trim function will remove the extra space added before or after

Now we can't remove or make any changes to this Method

3. Analysis

  a) the trim function will work here as like
```
<property name="option" value=" space "/>
```
as the output of this code we will get only space which is good

b) if we remove the trim function the and set the property
```
<property name="option" value=" space "/>
```
```
 - cannot initialize module ParenPad - Cannot set property 'option' to ' space '
error continue...
```

3. How to kill this

that all have set the property directly ```SPACE``` so the usage of trim function 
nothing in all the condition thats why it is Mutated

```\tSPACE``` so it will appear to code as ```  SPACE```
as know the usage of trim come so the muattion will be killed

**Comment by rnveach** [comment]
> 2. Mutated code

You do not show what the mutated code looks like. I wish to confirm you understand the mutation by seeing the actual code representation, so please provide this.

For example, assume the original code is `a = b + c` and the mutation says "replaced b with zero". I would expect to see you tell me the mutated code looks like `a = 0 + c` (or some form of `a = c`).

I am not saying what you provided for this PR is not correct, I just want to verify you understand what pitest is doing as there will be more like this.

> 3. Analysis
> 3. How to kill this

Just nitpicking, but you say there is two 3s and neither one connect with the number 3 in the thought process issue but you provided your own way that it makes sense to me what your doing.

3 in the thought process is about running actual regression on external projects ( https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#checkstyle-pitest-regression ) . 4 is about what to do for what regression reported back.

Since you provided a PR, and the PR looks like it is passing, I am fine skipping the regression.

File: src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheckTest.java [hunk 1]
```diff
@@ -146,6 +146,14 @@ public void test1322879() throws Exception {
                expected);
     }
 
+    @Test
+    public void testTrimFunc() throws Exception {
```
rnveach, line 150: Please change `testTrimFunc` to `testTrimOptionProperty` and do something similar to the file name.

I wouldn't worry about shortening things like "Func" instead of "Function". It is better to express what is being done as long as it is within a reasonable amount of text, then for it to be vague and lose meaning over time and personnel.

rnveach (changes requested):
CI will show if mutation is resolved, so it is passing proves that.

1 minor item asked for above, and just 1 minor change requested.

src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheckTest.java
Kevin222004: done

Kevin222004:
> You do not show what the mutated code looks like. I wish to confirm you understand the mutation by seeing the actual code representation, so please provide this.
> 
> For example, assume the original code is a = b + c and the mutation says "replaced b with zero". I would expect to see you tell me the mutated code looks like a = 0 + c (or some form of a = c).
> 
> I am not saying what you provided for this PR is not correct, I just want to verify you understand what pitest is doing as there will be more like this.


Ok I have understood what you are exactly asking for 
The function trim is coming from the ```java.lang.String``` package and as of the current testing code the function is returning nothing or it is empty according to pitest documentation  we can Replace return values with an ‘empty’ value for that type as follows

romani (approved):
Awesome hack in config

rnveach (approved):

nrmancuso (approved):
@Kevin222004 it might be a good idea to search pitest suppressions for `trim` and see if we can do similar hack in other check properties.

Issue #12260, opened 2022-10-06
Resolve  Pitest suppression for AbstractParenPadCheck - 2


Kevin222004:
Related to https://github.com/checkstyle/checkstyle/issues/7797

kill mutation

https://github.com/checkstyle/checkstyle/blob/70bbcd835e7d51062abd2801d3329b791854891c/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml#L8

for AbstractParenPadCheck

rnveach:
Fix was merged

PR #12262, opened 2022-10-07
Issue #12260: Resolve Pitest suppression for AbstractParenPadCheck - 2

Kevin222004:
Issue #12260: Resolve Pitest suppression for AbstractParenPadCheck - 2

killing mutation 
https://github.com/checkstyle/checkstyle/blob/70bbcd835e7d51062abd2801d3329b791854891c/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml#L8


Kevin222004:
1. pitest report 
https://kevin222004.github.io/AfterChanges/com.puppycrawl.tools.checkstyle.checks.whitespace/AbstractParenPadCheck.java.html
Surviving Mutation: ```java/lang/String::toUpperCase with receiver → SURVIVED``` at line 80

2. Mutated code 
our Surviving mutation is ```java/lang/String::toUpperCase with receiver → SURVIVED```
this toUpperCase function will change the string as --> SPACE
if we see the without mutation the code looks like.  
```option = PadOption.valueOf(optionStr.trim());``` 

but we can't remove or make any changes to this Method bcs enum has set to be like 
SPACE so all the letter we set to option is must be in Uppercase
 
3. Analysis

  a) the toUpperCase function will work here as like
```
<property name="option" value="space"/>
```
as the output of this code we will get only SPACE which is good

b) if we remove the toUpperCase function and then try to set the property
```
<property name="option" value="space"/>
```
```
 - cannot initialize module ParenPad - Cannot set property 'option' to 'space'
error continue...
```

4. How to kill this

that all have set the property directly ```SPACE``` so the usage of toUpperCase function 
is nothing in all the condition thats why it is Mutated

```space``` so it will appear to code as ```SPACE```
as know the usage of toUpperCase come so the muattion will be killed
https://kevin222004.github.io/Uppercase/com.puppycrawl.tools.checkstyle.checks.whitespace/index.html

File: src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheckTest.java [hunk 1]
```diff
@@ -154,6 +154,14 @@ public void testTrimOptionProperty() throws Exception {
                expected);
     }
 
+    @Test
+    public void testUpperCaseProperty() throws Exception {
+        final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
+        verifyWithInlineConfigParser(
+                getPath("InputParenPadToCheckUpperCaseFunctionInOptionProperty.java"),
+               expected);
```
rnveach, line 162: I noticed this in this PR and previous. `expected` is indented 1 less then where it should.

Please ensure they line up. Please fix previous test in this same PR.

rnveach (changes requested):
1 minor change

src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/parenpad/InputParenPadToCheckUpperCaseFunctionInOptionProperty.java
```diff
@@ -0,0 +1,33 @@
+/*
+ParenPad
+option = space
```
romani, line 3: Instead of creating new input file, can we update one of existing.
We do not expect any changes in validation behavior, so content of file does not matter.
So reusage is good.

romani (changes requested):
Items

src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/parenpad/InputParenPadToCheckUpperCaseFunctionInOptionProperty.java
Kevin222004: done

romani (approved):
Ok to merge

rnveach (approved):

Issue #12261, opened 2022-10-07
Resolve Pitest suppression for EmptyForIteratorPadCheck


Kevin222004:
Related to https://github.com/checkstyle/checkstyle/issues/7797

kill mutation
https://github.com/checkstyle/checkstyle/blob/ba8efc21659a66600a25ed662b62d1b1cb7611ea/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml#L53
and 
https://github.com/checkstyle/checkstyle/blob/ba8efc21659a66600a25ed662b62d1b1cb7611ea/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml#L44

 for EmptyForIteratorPadCheck

romani:
Can we fix all similar issues at one issue and multiple PR or one PR?

nrmancuso:
If we are only adding tests, I think it would be fine to resolve a few of these in the scope of one PR/issue.

rnveach:
My wording to him before this issue was:

> Single mutation issues was just for getting started to ensure we can confirm communication and confirm understanding of pitest and how it functions. Since this check only has 2 mutations, please just create 1 more issue of a single mutation and afterwards I think we can do multiple. However, even with multiple it is best to keep separate PRs between adding tests and changing production code. Adding tests is easier to merge then changing production code which will require a closer review and regression reports.
Beware, if PRs start to get complex we will want them split more, so I would be careful on putting too much into a single PR. It is not that much more work to create separate branches and PRs.

`this check` was `AbstractParenPadCheck` at the time.

Kevin222004:
I have updated the pr description

rnveach:
Fix was merged

PR #12263, opened 2022-10-07
Issue #12261: Resolve Pitest suppression for EmptyForIteratorPadCheck

Kevin222004:
Issue #12261: Resolve Pitest suppression for EmptyForIteratorPadCheck

killed two surviving mutation

https://github.com/checkstyle/checkstyle/blob/ba8efc21659a66600a25ed662b62d1b1cb7611ea/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml#L44
and 
https://github.com/checkstyle/checkstyle/blob/ba8efc21659a66600a25ed662b62d1b1cb7611ea/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml#L53

### DIFF
--- a/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml
@@ -37,24 +37,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>EmptyForIteratorPadCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.whitespace.EmptyForIteratorPadCheck</mutatedClass>
-    <mutatedMethod>setOption</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to java/lang/String::toUpperCase with receiver</description>
-    <lineContent>option = PadOption.valueOf(optionStr.trim().toUpperCase(Locale.ENGLISH));</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>EmptyForIteratorPadCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.whitespace.EmptyForIteratorPadCheck</mutatedClass>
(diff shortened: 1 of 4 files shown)


Kevin222004:
1. pitest report 
https://kevin222004.github.io/AfterChanges/com.puppycrawl.tools.checkstyle.checks.whitespace/EmptyForIteratorPadCheck.java.html
Surviving Mutation: ```java/lang/String::trim with receiver → SURVIVED```
and 
Surviving Mutation: ```java/lang/String::toUpperCase with receiver → SURVIVED``` 

2. Mutated code 
1) trim --> our Surviving mutation is ```java/lang/String::trim with receiver → SURVIVED```
this trim function will remove the extra space added before or after
while setting the property ```SPACE`` to option
if we see the without mutation the code looks like.  
```option = PadOption.valueOf(optionStr.trim());``` 

 2) toUppercase --> our Surviving mutation is ```java/lang/String::toUpperCase with receiver → SURVIVED```
this toUpperCase function will change the string as --> SPACE
if we see the without mutation the code looks like.  
```option = PadOption.valueOf(optionStr.trim());``` 

but we can't remove or make any changes to this Method bcs enum has set to be like 
SPACE so all the letter we set to option is must be in Uppercase
 
3. Analysis

  a) the toUpperCase function and trim function will work here as like
```
<property name="option" value="  space"/>
```
as the output of this code we will get only SPACE which is good

b) if we remove the toUpperCase function and then try to set the property
```
<property name="option" value="space"/>
```
```
 - cannot initialize module ParenPad - Cannot set property 'option' to 'space'
error continue...
```
and similarly
if we remove the trim function the and set the property
```
<property name="option" value=" space "/>
```
```
 - cannot initialize module ParenPad - Cannot set property 'option' to ' space '
error continue...
```



4. How to kill this

that all have set the property directly ```SPACE``` the usage of toUpperCase function 
and trimfunction is nothing in all the condition thats why it is Mutated

```space``` and ```   SPACE``` so it will appear to code as ```SPACE```
as know the usage of toUpperCase and trim  come so the muattion will be killed
https://kevin222004.github.io/EmptyforIteratorPad/com.puppycrawl.tools.checkstyle.checks.whitespace/EmptyForIteratorPadCheck.java.html

rnveach (approved):

Issue #3095, opened 2016-04-13
Add COMPILATION_UNIT token in Ast Tree, remove EOF token


romani:
base on https://groups.google.com/forum/#!topic/checkstyle/4njXEcUsf34

Javadoc claim that root token exists but it does not exists after parsing:
http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#EOF

ToDo:
investigate a reason why root node was removed from Ast Tree.
remove 
https://github.com/checkstyle/checkstyle/blob/5bf4e04862b6aa897dc295284def928123f1bb6b/config/import-control.xml#L166-L169

**Rationale of update and hints on how to update affected Checks**:
https://github.com/checkstyle/checkstyle/pull/10574#issuecomment-898806357

**Migration note, affecting all xpath queries**:
https://github.com/checkstyle/checkstyle/issues/3095#issuecomment-897257902

rnveach:
GUI creates it's own root
https://github.com/checkstyle/checkstyle/blob/a73ff0890c27cbe8affc1a55f2710a231687bd85/src/main/java/com/puppycrawl/tools/checkstyle/gui/ParseTreeTablePresentation.java#L257-L261

nmancus1:
Proposed result:

```bash

➜  src cat Test.java
package com.puppycrawl.tools.checkstyle.grammar.antlr4;

import java.io.BufferedReader;
import java.io.FileNotFoundException;

public class Test {}

@interface NonNull{}

➜  src javac Test.java

➜  src java -jar ~/IdeaProjects/checkstyle/target/checkstyle-8.45-SNAPSHOT-all.jar -t Test.java
PACKAGE_DEF -> package [1:0]
|--ANNOTATIONS -> ANNOTATIONS [1:47]
|--DOT -> . [1:47]
|   |--DOT -> . [1:39]
|   |   |--DOT -> . [1:28]
|   |   |   |--DOT -> . [1:22]
|   |   |   |   |--DOT -> . [1:11]
|   |   |   |   |   |--IDENT -> com [1:8]
|   |   |   |   |   `--IDENT -> puppycrawl [1:12]
|   |   |   |   `--IDENT -> tools [1:23]
|   |   |   `--IDENT -> checkstyle [1:29]
|   |   `--IDENT -> grammar [1:40]
|   `--IDENT -> antlr4 [1:48]
`--SEMI -> ; [1:54]
[...]

PR #12268, opened 2022-10-07
Update Xpath suppression example

findepi:
In local testing it was found that SuppressionXpathSingleFilter for suppressing RedundantModifier in interface definitions did not work. A variant with leading double slash does work.

### DIFF
--- a/src/xdocs/config_filters.xml
+++ b/src/xdocs/config_filters.xml
@@ -1562,7 +1562,7 @@ public class FileOne {}
           <pre>
 &lt;module name="SuppressionXpathSingleFilter"&gt;
   &lt;property name="checks" value="RedundantModifier"/&gt;
-  &lt;property name="query" value="/INTERFACE_DEF//*"/&gt;
+  &lt;property name="query" value="//INTERFACE_DEF//*"/&gt;
 &lt;module/&gt;
           </pre>
         </div>


nrmancuso:
@findepi thanks a lot for bringing this to our attention: we should have taken care of this update in the scope of https://github.com/checkstyle/checkstyle/issues/3095. See details about why we need to update examples at https://github.com/checkstyle/checkstyle/issues/3095#issuecomment-897257902.

Please do the following:
1. Make your commit as `doc: update XPath examples (related to #3095)`
2. Please check examples in [MatchXpath](https://checkstyle.sourceforge.io/config_coding.html#MatchXpath), [SuppressionXpathFilter](https://checkstyle.sourceforge.io/config_filters.html#SuppressionXpathFilter), and 
[SuppressionXpathSingleFilter](https://checkstyle.sourceforge.io/config_filters.html#SuppressionXpathSingleFilter) 
3.Update any top level nodes (`/CLASS_DEF`, `/INTERFACE_DEF`, `/ENUM_DEF`, and `/RECORD_DEF`) by preceding them with `/COMPILATION_UNIT` instead of `//` . Let's not do `//`, as this matches nodes at any depth and is not equivalent to the previous intention of the examples.

src/xdocs/config_filters.xml
```diff
@@ -1562,7 +1562,7 @@ public class FileOne {}
           <pre>
 &lt;module name="SuppressionXpathSingleFilter"&gt;
   &lt;property name="checks" value="RedundantModifier"/&gt;
-  &lt;property name="query" value="/INTERFACE_DEF//*"/&gt;
+  &lt;property name="query" value="//INTERFACE_DEF//*"/&gt;
```
nrmancuso, line 1565:
```suggestion
  &lt;property name="query" value="/COMPILATION_UNIT/INTERFACE_DEF//*"/&gt;
```

romani:
moved to https://github.com/checkstyle/checkstyle/pull/12276

Issue #12275, opened 2022-10-08
Resolve Pitest suppression for EmptyForInitializerPad-2


Kevin222004:
Related to https://github.com/checkstyle/checkstyle/issues/7797

kill mutation

https://github.com/checkstyle/checkstyle/blob/357ba1ed99f82b07d5663d23091ca9c10bb382b8/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml#L4-L9

 for EmptyForInitializerPad

Kevin222004:
As of know it is different mutation, so I have taken just one mutation in this pr

romani:
fix is merged

Issue #7797, opened 2020-03-08
Resolve Pitest Issues


rnveach:
This issue is the main tracker of all pitest work and will be split into multiple individual issues as new work is identified. Nothing can be done in this issue.

Shown in https://github.com/checkstyle/checkstyle/blob/master/.ci/pitest.sh ,

Checkstyle uses pitest (mutation testing) to test if there is enough tests to cover all code in main source. If new code is written, or existing code is re-written, and no tests are added, pitest will fail and require more tests to cover the new changes. It is extremely important to get familar with pitest as it is an integral part of working with Checkstyle.

Checkstyle adopted pitest after we already had a bunch of modules and tests. Sadly, not all tests cover all the code we have and pitest is flagging additional lines that need to be resolved. Following pitest's terminology, lines that still need to be resolved are called Surviving mutations. The overview of all unresolved mutations can be found at https://github.com/checkstyle/checkstyle/blob/master/.ci/pitest.sh .

If at anytime, more help is needed, don't hesitate to ping an admin for guidance.

For each issue the following steps need to be taken:
1) The pitest report needs to be run on a clean checkstyle master branch and uploaded to the issue on exactly the type of mutation(s) that is surviving. It is possible that more than 1 mutation could be surviving on the same line, but it is also possible removing 1 mutation will remove the others too.
2) The reported pitest mutation must be hardcoded into a new checkstyle branch and the branch uploaded to github and reported in the issue to verify the user understands the mutation and how pitest works behind the scenes. See https://pitest.org/quickstart/mutators/
3) Regression must be run with the hardcoded mutation branch and the regression uploaded to the issue to show if regression can find test cases that can be added to our own UTs to kill the mutations. See https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#checkstyle-pitest-regression . We recommend running regression on all projects.
**Note**: Step 3 may not be able to be done if the item to kill is throwing an exception that will terminate the checkstyle run. Regression is more for logic surrounding violations being printed or not.

4a) If regression does find a case, or a case is found by another means, a PR must be submitted adding only the new case to checkstyle. The mutated branch must not be used and can be deleted.
4aa) In the changes for the PR, `pitest.sh` has to be modified to remove the line in question to show the mutation is killed and no longer a concern.
4ab) With the PR a new pitest must be run and uploaded to the PR to show that the new test case does kill the mutation.

[...]

PR #12277, opened 2022-10-08
Issue #12275: Resolve Pitest suppression for EmptyForInitializerPad-2

Kevin222004:
Issue #12275: Resolve Pitest suppression for EmptyForInitializerPad-2

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/a85bd5c1897e158b6495850ba792bf4b/raw/37385b138ced4f80b43ccf865900a812500f4f42/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyForInitializerPadCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyForInitializerPadCheck.java
@@ -148,11 +148,9 @@ public class EmptyForInitializerPadCheck
     @Override
     public void visitToken(DetailAST ast) {
         if (!ast.hasChildren()) {
-            // empty for initializer. test pad before semi.
-            final DetailAST semi = ast.getNextSibling();
-            final int semiLineIdx = semi.getLineNo() - 1;
-            final int[] line = getLineCodePoints(semiLineIdx);
-            final int before = semi.getColumnNo() - 1;
+            final int lineIdx = ast.getLineNo() - 1;
+            final int[] line = getLineCodePoints(lineIdx);
+            final int before = ast.getColumnNo() - 1;
             // don't check if semi at beginning of line
             if (!CodePointUtil.hasWhitespaceBefore(before, line)) {
                 if (option == PadOption.NOSPACE
--- a/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml
@@ -1,15 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
   <mutation unstable="false">
-    <sourceFile>EmptyForInitializerPadCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.whitespace.EmptyForInitializerPadCheck</mutatedClass>
-    <mutatedMethod>visitToken</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getNextSibling with receiver</description>
-    <lineContent>final DetailAST semi = ast.getNextSibling();</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>EmptyLineSeparatorCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.whitespace.EmptyLineSeparatorCheck</mutatedClass>
     <mutatedMethod>checkToken</mutatedMethod>


Kevin222004:
1. pitest report 

    https://kevin222004.github.io/general/com.puppycrawl.tools.checkstyle.checks.whitespace/EmptyForInitializerPadCheck.java.html
    `replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getNextSibling with receiver`


2. Mutated code 

    current code 
    `final DetailAST semi = ast.getNextSibling();`
    is Mutated to 
    `final DetailAST semi = ast;`


3. Analysis

    If we change `final DetailAST semi = ast.getNextSibling();` to `final DetailAST semi = ast;` 
    it want make any difference because in this check we just care about the `parenPad-> (` and `;`
    `.getNextSibling()` will point the condition of for statment statment if this is for loop and in the
    source code it returns as `null`.

    So the usage of `.getNextSibling()` is currently nothing so I think it is not possible to create a Unit
    test file   


4. How to kill this

    removed the `.getNextSibling()`

Report of killed Mutation 
https://kevin222004.github.io/emptpad/com.puppycrawl.tools.checkstyle.checks.whitespace/EmptyForInitializerPadCheck.java.html

romani:
Ok, such code most likely existed for some good reason we need to find such java code.
Please provide regression diff report.

Hint: https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports#how-to-find-java-code-that-can-kill-mutation

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyForInitializerPadCheck.java
```diff
@@ -149,7 +149,7 @@ public int[] getRequiredTokens() {
     public void visitToken(DetailAST ast) {
         if (!ast.hasChildren()) {
             // empty for initializer. test pad before semi.
-            final DetailAST semi = ast.getNextSibling();
+            final DetailAST semi = ast;
```
romani, line 152: Remove this variable, just use `ast`

romani (changes requested):
Items

Kevin222004:
Diff Regression config: https://gist.githubusercontent.com/Kevin222004/a85bd5c1897e158b6495850ba792bf4b/raw/b41a01d49b4aa0474ae31deef8f9da7e1740db8c/check.xml
Report label: validateBetweenScopesFalse

Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

Kevin222004:
[]https://github.com/checkstyle/checkstyle/commit/998c65dc977d457eeb645c91834018655280ef63(url)

Kevin222004:
#https://github.com/checkstyle/checkstyle/commit/998c65dc977d457eeb645c91834018655280ef63

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyForInitializerPadCheck.java
Kevin222004: done

Kevin222004:
https://github.com/checkstyle/checkstyle/commit/998c65dc977d457eeb645c91834018655280ef63

Kevin222004:
GitHub, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyForInitializerPadCheck.java
```diff
@@ -149,10 +149,9 @@ public int[] getRequiredTokens() {
     public void visitToken(DetailAST ast) {
         if (!ast.hasChildren()) {
             // empty for initializer. test pad before semi.
-            final DetailAST semi = ast.getNextSibling();
```
romani, line 152: Remove comments

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyForInitializerPadCheck.java
Kevin222004: done

Kevin222004:
Github, generate report

romani (approved):
ok to merge

rnveach:
> Diff Regression config: https://gist.githubusercontent.com/Kevin222004/a85bd5c1897e158b6495850ba792bf4b/raw/b3d6fd5577d468d81339aa9f80b1581729b22bb5/config.xml

https://checkstyle.org/config_whitespace.html#EmptyForInitializerPad
This config is not enough.
Please read https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#checkstyle-pitest-regression and expand config to all permutations which this check can be.

> so I think it is not possible to create a Unit test file

If nothing changes on new regression, prove to me you stepped through the source and there is no way to create an input file to kill this mutation and changing production code is the only option.
See https://github.com/checkstyle/checkstyle/issues/7797 , 4b .

Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

Kevin222004:
Updated config https://gist.githubusercontent.com/Kevin222004/a85bd5c1897e158b6495850ba792bf4b/raw/37385b138ced4f80b43ccf865900a812500f4f42/config.xml

Kevin222004:
@rnveach please look at the updated config

rnveach:
Updated config is better. It shows there is still no regression.

romani:
AST structure might be changed from initial implementation.
https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#FOR_INIT
SEMI is sibling  of FOR_INIT (the only token this Check run on)

Kevin222004:
> To me it seems odd that we are removing references to a `;` when code examples talk about `;`.

@rnveach @romani if it is not looking good then we can make changes like this https://github.com/checkstyle/checkstyle/pull/12277#pullrequestreview-1135271872

rnveach:
> prove to me you stepped through the source and there is no way to create an input file to kill this mutation and changing production code is the only option

In short, there is no input file we can create here.

In this case AST is `FOR_INIT` and the next sibling is `SEMI`. When looking at the code after the mutation, there is checks done against Line and Column numbers. My first attempt to examine this was to try and get the 2 ASTs on different lines, as this would cause violations to be on different lines and prove we need the violation to be on only one of them.

This failed to be done because of this check and the way our ASTs are formed. `EmptyForInitializer` is only for for loops which have no initialization (IE: `for (;`). Even with the `FOR_INIT` and `SEMI` on different lines, our AST structure always makes them the same line/column.
````
for (;
 ;) {}
````
creates:
````
        |      |--LITERAL_FOR -> for [3:0]
        |      |  |--LPAREN -> ( [3:4]
        |      |  |--FOR_INIT -> FOR_INIT [3:5]
        |      |  |--SEMI -> ; [3:5]
````
We have both tokens appear to be on the same line/column. It doesn't matter if there is comments inside the `FOR_INIT` as it is still the same behavior.

Since we can't get these 2 tokens to have different line/column, there isn't a way to kill the mutation as it is now.

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyForInitializerPadCheck.java [hunk 3]
```diff
@@ -148,11 +148,9 @@ public int[] getRequiredTokens() {
     @Override
     public void visitToken(DetailAST ast) {
         if (!ast.hasChildren()) {
-            // empty for initializer. test pad before semi.
-            final DetailAST semi = ast.getNextSibling();
-            final int semiLineIdx = semi.getLineNo() - 1;
+            final int semiLineIdx = ast.getLineNo() - 1;
```
rnveach, line 151: This is no longer a semi, and needs to be renamed. Please make it `lineIdx`.

rnveach (changes requested):
1 minor change.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyForInitializerPadCheck.java
Kevin222004: done

rnveach: Not done, `lineIdx` should have a lowercase `l` not an uppercase one.

rnveach (approved):

Issue #12280, opened 2022-10-09
Resolve Pitest suppression for SingleSpaceSeparatorCheck


Kevin222004:
Related to https://github.com/checkstyle/checkstyle/issues/7797

kill mutation
https://github.com/checkstyle/checkstyle/blob/fa6856985e3eb32f11ae63e56c19fc2181a50d06/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml#L120-L136

Kevin222004:
I am working on it, send pr when its approved

rnveach:
I am closing this and moving it under #12594 .

PR #12283, opened 2022-10-09
Issue #12280: Resolve Pitest suppression for SingleSpaceSeparatorCheck

Kevin222004:
Issue #12280: Resolve Pitest suppression for SingleSpaceSeparatorCheck

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/55bba7110290f08feb12e3c4f36976e4/raw/a3bd29dd8d9035b48d5bbfba68042e33e4b82526/CONFIG.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
@@ -179,7 +179,6 @@ public class SingleSpaceSeparatorCheck extends AbstractCheck {
      */
     private void visitEachToken(DetailAST node) {
         DetailAST currentNode = node;
-        final DetailAST parent = node.getParent();
 
         do {
             final int columnNo = currentNode.getColumnNo() - 1;
@@ -199,7 +198,7 @@ public class SingleSpaceSeparatorCheck extends AbstractCheck {
                 currentNode = currentNode.getFirstChild();
             }
             else {
-                while (currentNode.getNextSibling() == null && currentNode.getParent() != parent) {
+                while (currentNode.getNextSibling() == null && currentNode.getParent() != null) {
                     currentNode = currentNode.getParent();
                 }
                 currentNode = currentNode.getNextSibling();
--- a/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml
@@ -116,22 +116,4 @@
     <description>replaced call to com/puppycrawl/tools/checkstyle/utils/CodePointUtil::trim with argument</description>
     <lineContent>final int[] substringAfterToken = CodePointUtil.trim(</lineContent>
   </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>SingleSpaceSeparatorCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.whitespace.SingleSpaceSeparatorCheck</mutatedClass>
-    <mutatedMethod>visitEachToken</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/api/DetailAST::getParent</description>
-    <lineContent>final DetailAST parent = node.getParent();</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>SingleSpaceSeparatorCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.whitespace.SingleSpaceSeparatorCheck</mutatedClass>
-    <mutatedMethod>visitEachToken</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getParent with receiver</description>
-    <lineContent>final DetailAST parent = node.getParent();</lineContent>
-  </mutation>
 </suppressedMutations>


Kevin222004:
1. pitest report 

    https://kevin222004.github.io/general/com.puppycrawl.tools.checkstyle.checks.whitespace/SingleSpaceSeparatorCheck.java.html
    `removed call to com/puppycrawl/tools/checkstyle/api/DetailAST::getParent`
    `replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getParent with receiver`


2. Mutated code 

    current code 
    `final DetailAST parent = node.getParent();`
    is Mutated to 
    `final DetailAST parent = node;`


3. Analysis

    The current situation of code `final DetailAST parent = node.getParent();`
    in this if we see closely to the method from which this line of code belong 
    ```
        private void visitEachToken(DetailAST node) {
        DetailAST currentNode = node;
        final DetailAST parent = node.getParent();

    ```
    Now here In ```private void visitEachToken(DetailAST node) {``` node is always be our compilation unit
    or we say the package  ex( the AST I have provided of one test file).

    Now the `final DetailAST parent = node.getParent();` here parent is calling getParent() to get 
    the parent node of node which is compilation unit now obvivously compilation is going to the root of tree
    so it become always `null` 

    now `final DetailAST parent = node.getParent();` the parent is using by method at one place only 
    at   `while (currentNode.getNextSibling() == null && currentNode.getParent() != parent) {`
    now the parent is always being `null` so we change it as null

    AST for test file https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/singlespaceseparator/InputSingleSpaceSeparatorMinColumnNo.java
    ```

    ```


4. How to kill this

    removed the `final DetailAST parent = node.getParent();` and change the parent as null in method

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
```diff
@@ -199,7 +198,7 @@ private void visitEachToken(DetailAST node) {
                 currentNode = currentNode.getFirstChild();
             }
             else {
-                while (currentNode.getNextSibling() == null && currentNode.getParent() != parent) {
+                while (currentNode.getNextSibling() == null && currentNode.getParent() != null) {
```
romani, line 201: this will go up to a ROOT ast, we should not do this, as it will walk through all nodes in tree. It is too much.

romani (changes requested):
item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
Kevin222004: yes, my bad, sorry this is being huge mistake of my. I have not noticed it's a loop

Kevin222004:
as of know the whole approach is wrong

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
Kevin222004: @romani it is possible for any situation where the parent will get null rather then  Root. I am not seen any condition for this as of know

Kevin222004: so I think if we write it as null it will still walk till root ast

romani: I recommed to apply mutation as pitest suggest and run diff report on it (do not forget to disable all CI to speedup execution)

Kevin222004:
I am generating a diff report to just analyize

Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
rnveach: So you know you don't need CI to run regression.
https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#checkstyle-tester
All this can be run on your local. While we do recommend running against all projects, you can trim the list of projects to get a quick idea if something can be found. It can also let you try different code changes when you do find a difference or such.

romani: @Kevin222004 ,
try to remove `currentNode.getParent() != null` completely.

Kevin222004: > @Kevin222004 , try to remove `currentNode.getParent() != null` completely.

As of know it is becoming quite complicated to me if we remove the `currentNode.getParent() != null` it will working fine and it will traverse the tree as expected. 
but it is giving error when the violation is coming in file. I will still trying to find out the exact reason if you gave any hint. it would be helpfull

romani: Just remove it and run diff report generation on your local or by GitHub action.
It is easy.

rnveach:
@Kevin222004 ping

Kevin222004:
> @Kevin222004 ping

sorry for no activity on this pr from, long after https://github.com/checkstyle/checkstyle/pull/12283#discussion_r991607090
I have tried to remove it, the tree is traversing perfectly as expectation but  wherever violation is expected it is giving 
some error so unit test are failing . I am not able to work on it as it take time. because of examination, but i am  trying to figure it out. and complete the pr

rnveach:
Let us know when you think you may return.
I would recommend to provide the difference reports that were requested, and post that URL in https://github.com/checkstyle/checkstyle/pull/12283#discussion_r991607090 with a brief summary (differences or no differences) and we will continue from there.

Kevin222004:
> Let us know when you think you may return.

for this pr and all the issue related to Pitest I will comeback around December starting

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
Kevin222004: Ok so as a minor update if i remove this   `currentNode.getParent() != null` completely according to https://github.com/checkstyle/checkstyle/pull/12283#discussion_r991607090  so unit test cases are failing in this situation so regression would not be generated there.
 please gave a look at this as well  https://github.com/checkstyle/checkstyle/pull/12283#issuecomment-1272589581

Kevin222004: @romani ping

romani: this this while condition as as.
and apply mutation to `final DetailAST parent = node.getParent();` to make it `final DetailAST parent = node;`
and see what is missed.

sounds like we can remove `final DetailAST parent = node.getParent();` completely and make while condition as
`currentNode.getParent() != node`

rnveach:
I am taking over resolving this mutation, so I am closing this PR for my own.

Issue #12087, opened 2022-08-22
Create IDEA Inspections Docker image for CircleCI


nrmancuso:
After https://github.com/checkstyle/checkstyle/issues/11876#issuecomment-1222297812, it has become apparent that TeamCity is too complicated for our needs; we have to maintain at least five different configs for one project, and still have configuration in the web UI to contend with, just to run IDEA inspections via command line in some container. See evidence [here](https://github.com/checkstyle/contribution/pull/638#issuecomment-1221630110).

We do not need TeamCity to run IDEA inspections via command line in a container; we can do this ourselves and avoid complicated TeamCity config altogether. This will also make it easy for maintainers that do not use Intellij to run inspections locally.

My proposal:

1. Create docker image, using https://circleci.com/developer/images/image/cimg/openjdk as a base
2. Add latest IDEA binary to docker image
3. Use this image in CircleCI job to run IDEA inspections

Dockerfile, etc. can live in https://github.com/checkstyle/contribution/tree/master/devops/docker

Implementation should be such that we can have a simple/ fast process in CircleCI:
1. checkout
2. run inspections with `-v 3` verbosity
3. check that report directory is empty, if not, fail build

We should also include a script to run this locally (this should be created as the image is developed to use for testing locally anyway), and a `.bat` file for Windows users.

See https://github.com/checkstyle/checkstyle/blob/a152a3a704d3373f34886df322aeb5fbc518003c/.ci/idea_inspection.bat and https://github.com/checkstyle/checkstyle/blob/a152a3a704d3373f34886df322aeb5fbc518003c/.ci/idea_inspection.sh as reference (these are not for containers)

UPDATE: https://hub.docker.com/r/checkstyle/idea-docker/tags

strkkk:
@nick-mancuso https://github.com/checkstyle/checkstyle/issues/12075 can be a replacement for idea analysis in circleci. wdyt?

nick-mancuso:
@strkkk Unless I am missing something, it does not support inspections config in project. So, until it does, it is not a real replacement, and has nowhere close to the number of inspections available.

romani:
migration is merged

romani:
I detached pullRequest builds 
![image](https://user-images.githubusercontent.com/812984/194786928-2f8b6b30-210d-46a4-a07c-f2445d81e5b7.png)
with hope it will stop execution of TC on each PR

PR #12287, opened 2022-10-10
doc: remove teamcity badge but keep links in case we will restore it

romani:
to test https://github.com/checkstyle/checkstyle/issues/12087#issuecomment-1272669800


romani:
All green, merging ..

PR #12295, opened 2022-10-10
infra: remove outdated jobs from travis

rnveach:
All jobs were commented out and confirmed moved somewhere else.


romani (approved):
ok to merge

PR #12309, opened 2022-10-16
infra: enable printing diff to monitor what is going on in Checker

romani:
extra fix after:
https://github.com/checkstyle/checkstyle/commit/dbd312a6ee808708cf35f1b24ef4ed41987f1ecc

### DIFF
--- a/.github/workflows/checker-framework.yml
+++ b/.github/workflows/checker-framework.yml
@@ -40,11 +40,11 @@ jobs:
         run: groovy ./.ci/checker-framework.groovy ${{ matrix.profile }} || true
 
       - name: Git Diff on fail
-        if: failure()
-        run: git diff > target/${{ matrix.profile }}.patch
+        # if: failure()
+        run: git diff && git diff > target/${{ matrix.profile }}.patch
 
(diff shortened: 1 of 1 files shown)


.github/workflows/checker-framework.yml
```diff
@@ -40,8 +40,9 @@ jobs:
         run: groovy ./.ci/checker-framework.groovy ${{ matrix.profile }} || true
 
       - name: Git Diff on fail
-        if: failure()
-        run: git diff > target/${{ matrix.profile }}.patch
+        # if: failure()
+        # run: git diff > target/${{ matrix.profile }}.patch
+        run: git diff
 
       - name: Share patch as artifact to apply on local
```
rnveach, line 47: Patch needs the same changes as well.

romani: I did this on purpose, but as you mention this also , ok, we print in output + in file and store as artifacts

rnveach: Why `git diff && git diff`?

rnveach (approved):

PR #12313, opened 2022-10-19
Pull #12312: Remove custom equals and hashCode method and correct compareTo method in checker-framework.groovy

Vyom-Yadav:
Contents of this PR:

- Removed custom hashcode and equals method, now both of those methods are generated using `@EqualsAndHashCode`
- Resolved a few IDEA violations generated locally.

### DIFF
--- a/.ci/checker-framework.groovy
+++ b/.ci/checker-framework.groovy
@@ -8,9 +8,9 @@ import groovy.xml.XmlUtil
 import java.util.regex.Matcher
 import java.util.regex.Pattern
 
-@Field static final String USAGE_STRING = "Usage groovy " +
+@Field static final String USAGE_STRING = 'Usage groovy ' +
         ".${File.separator}.ci${File.separator}checker-framework.groovy" +
-        " [profile] [--list]\n"
+        ' [profile] [--list]\n'
 
 int exitCode = 1
 if (args.length == 1) {
@@ -27,14 +27,14 @@ System.exit(exitCode)
  * @param argument command line argument
  * @return {@code 0} if command executes successfully, {@code 1} otherwise
  */
-private int parseArgumentAndExecute(String argument) {
+private int parseArgumentAndExecute(final String argument) {
     final int exitCode
     final Set<String> profiles = getCheckerFrameworkProfiles()
     if (profiles.contains(argument)) {
         exitCode = checkCheckerFrameworkReport(argument)
     }
-    else if (argument == "--list") {
-        println "Supported profiles:"
+    else if (argument == '--list') {
+        println 'Supported profiles:'
         profiles.each { println it }
         exitCode = 0
     }
@@ -54,10 +54,10 @@ private static Set<String> getCheckerFrameworkProfiles() {
     final GPathResult mainNode = new XmlSlurper().parse(".${File.separator}pom.xml")
     final NodeChildren ids = mainNode.profiles.profile.id as NodeChildren
     final Set<String> profiles = new HashSet<>()
-    ids.each { node ->
+    ids.each { final node ->
         final GPathResult id = node as GPathResult
         final String idText = id.text()
-        if (idText.startsWith("checker-framework-")) {
+        if (idText.startsWith('checker-framework-')) {
             profiles.add(idText)
         }
     }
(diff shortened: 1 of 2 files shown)


.ci/checker-framework.groovy
```diff
@@ -71,14 +71,14 @@ private static Set<String> getCheckerFrameworkProfiles() {
...
  * @return {@code 0} if checker framework report is as expected, {@code 1} otherwise
  */
-private static int checkCheckerFrameworkReport(String profile) {
+private static int checkCheckerFrameworkReport(final String profile) {
     final XmlParser xmlParser = new XmlParser()
     final String suppressedErrorsFileUri =
             ".${File.separator}.ci${File.separator}" +
                     "checker-framework-suppressions${File.separator}${profile}-suppressions.xml"
-    final List<String> checkerFrameworkErrors = getCheckerFrameworkErrors(profile)
+    final List<List<String>> checkerFrameworkErrors = getCheckerFrameworkErrors(profile)
```
Vyom-Yadav, line 79: This was incorrect, the wrong datatype was being used.

.ci/checker-framework.groovy
```diff
@@ -109,27 +110,27 @@ private static int checkCheckerFrameworkReport(String profile) {
...
     final String command = "mvn -e --no-transfer-progress clean compile -P${profile}"
     final Process process = getOsSpecificCmd(command).execute()
-    process.in.eachLine { line ->
+    process.in.eachLine { final line ->
         checkerFrameworkLines.add(line)
         println(line)
     }
     process.waitFor()
-    final List<List<String>> checkerFrameworkErrors = [][] as ArrayList
+    final List<List<String>> checkerFrameworkErrors = [] as ArrayList
```
Vyom-Yadav, line 122: [][] as ArrayList is `ArrayList<Object>` which isn't the type we wanted, this works and doesn't give any violation, please verify this change.

romani:
Codenarc and checker CIs are failing

.ci/checker-framework.groovy
```diff
@@ -443,68 +439,18 @@ class CheckerFrameworkError implements Comparable<CheckerFrameworkError> {
             return i
         }
 
-        if (this.isUnstable() || other.isUnstable()) {
```
rnveach, line 446: Why is `compareTo` changing when this says it is just changing equals/hash? The 2 are not connected together so I don't understand why we are including them together.

.ci/checker-framework.groovy
```diff
@@ -27,14 +27,14 @@ System.exit(exitCode)
  * @param argument command line argument
  * @return {@code 0} if command executes successfully, {@code 1} otherwise
  */
-private int parseArgumentAndExecute(String argument) {
+private int parseArgumentAndExecute(final String argument) {
     final int exitCode
-    final Set<String> profiles = getCheckerFrameworkProfiles()
+    final Set<String> profiles = checkerFrameworkProfiles
```
rnveach, line 32: I am personally liking `()` for method calls. Makes it clear to differentiate methods and fields/variables.

.ci/checker-framework.groovy
rnveach: > please verify this change

How should we be verifying this?

.ci/checker-framework.groovy
romani: I always prefer to write in Groovy like in Java, to be ready to convert script to java project easily and do not deal with remembering syntax sugar of Groovy.

.ci/checker-framework.groovy
romani: Can we use `new arrayList<>()` to keep it clear ?

Vyom-Yadav: codenarc gives a violation on using that.

romani: I am ok disable this rule. 
I make code very confusing, I hope they will improve such rule in future

Vyom-Yadav: Disabled and change to `new ArrayList<>()`

.ci/checker-framework.groovy
Vyom-Yadav: `compareTo` is also incorrect, we do not need any special comparison when an error is `unstable`, added it to the commit message and PR title.

.ci/checker-framework.groovy
Vyom-Yadav: Reverted it.

.ci/checker-framework.groovy
rnveach: I am seeing a lot more changing than unstable. Message and details changed. I assume this is to rewriting `temp-var` and `capture`?

Please put all changes to first post as well since you are documenting this as your issue.

rnveach (approved):
Still looking for https://github.com/checkstyle/checkstyle/pull/12313#discussion_r1001342112 to occur.

romani (approved):

.ci/checker-framework.groovy
Vyom-Yadav: Done.

rnveach:
@romani @Vyom-Yadav I just noticed, PR number in commit does not match this PR. It is a number for another PR.

I don't know how this will impact release notes, but we should add validation that `Pull #` can only be used for the current PR.

Issue #12282, opened 2022-10-09
Resolve Pitest suppression for ParenPad


Kevin222004:
Related to https://github.com/checkstyle/checkstyle/issues/7797

kill mutation
https://github.com/checkstyle/checkstyle/blob/fa6856985e3eb32f11ae63e56c19fc2181a50d06/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml#L93-L100

Gairick52:
@Kevin222004 please suggest me some good first issues

romani:
@Gairick52 , please read https://github.com/checkstyle/checkstyle/wiki/To-GSoC-participant, there are first,second,thrird issue to let you gradually be introduced to project.

Gairick52:
@romani thank you very much

Kevin222004:
@rnveach We have currently this line of code which is mutated https://github.com/checkstyle/checkstyle/blob/ad27aa0feb4804cf3a14430b2cb90c3d13dbe532/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheck.java#L257

and the source code for this mutated code is just to remove it

rnveach:
> the source code for this mutated code is just to remove it

The mutation is not to remove it.

Kevin222004:
@rnveach yes we don't have to remove the switch case according to wiki I have also done like that before. and know it is clear to me 

when we are making the changes like that I am getting errors in three input file i am sharing one with you 
https://github.com/checkstyle/checkstyle/blob/ad27aa0feb4804cf3a14430b2cb90c3d13dbe532/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/parenpad/InputParenPadLeftRightAndNoSpace1.java#L122-L126
here the violation is missing in line 125 at 2 places 
according to the current scenario checkstyle don't generate the violation for any case for `QUESTION` I just want to know it is expected a violation in such condition like line no 126 in same file according to its config comment msg

rnveach:
If you make a change to mimic a mutation and our test fails, then either the mutation you applied isn't correct or there is more going on with pitest than the mutation alone (caching, etc...).  Normally tests do not fail with the mutation applied.

Please provide me the step by step we request. Provide me the mutated branch pushed to the server. Let me know if tests fail with the mutation applied.

romani:
Please send PR with mutation to let us see all details.

Please try to remove all other "case" one by one.

[...]

PR #12318, opened 2022-10-21
Issue #12282: Resolve Pitest suppression for ParenPad

Kevin222004:
Issue #12282: Resolve Pitest suppression for ParenPad

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/e206f8b15b0d8c04e1d6b1fbded90dbe/raw/6b6a30c0a864caad500abed316e2e7a51d99c59b/config.xml

### DIFF
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/parenpad/InputParenPadForSynchronized.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/parenpad/InputParenPadForSynchronized.java
@@ -1,0 +1,23 @@
+/*
+ParenPad
+option = (default)nospace
+tokens = LITERAL_SYNCHRONIZED, CTOR_DEF, CTOR_CALL, LITERAL_IF
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.whitespace.parenpad;
+
+class InputParenPadForSynchronized {
+
+    private static InputParenPadForSynchronized instance = null;
+    private InputParenPadForSynchronized() {
+    }
+
+    public synchronized static InputParenPadForSynchronized getInstance() {
+        if(instance == null ) { // violation
+            instance = new InputParenPadForSynchronized();
+        }
+        return instance;
+    }
+}
--- a/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-whitespace-suppressions.xml
@@ -85,15 +85,6 @@
     <sourceFile>ParenPadCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck</mutatedClass>
     <mutatedMethod>visitToken</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.RemoveSwitchMutator_3</mutator>
-    <description>RemoveSwitch 3 mutation</description>
-    <lineContent>switch (ast.getType()) {</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>ParenPadCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck</mutatedClass>
-    <mutatedMethod>visitToken</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.RemoveSwitchMutator_7</mutator>
     <description>RemoveSwitch 7 mutation</description>
     <lineContent>switch (ast.getType()) {</lineContent>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheckTest.java
@@ -513,6 +513,16 @@ public class ParenPadCheckTest
                 getPath("InputParenPadCheckEmoji.java"), expected);
     }
 
+    @Test
+    public void testParenPadForSynchronized() throws Exception {
+
+        final String[] expected = {
+            "18:29: " + getCheckMessage(MSG_WS_PRECEDED, ")"),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputParenPadForSynchronized.java"), expected);
+    }
+
     /**
      * Pitest requires us to specify more concrete lower bound for condition for
      * ParenPadCheck#isAcceptableToken as nodes of first several types like CTOR_DEF,


Kevin222004:
Github, generate report

romani:
Tests are failing, no need for report.
Sorry ch mutation should not be survival.
Did you try to remove others? May be we misunderstood indexation of cases PR may be ... Please do several more removals of others case ( but remove only one at the time)

rnveach:
case 0 fails tests.
case 1 fails tests.
case 2 fails tests.
case 3 fails tests.
case 4 fails tests.
case 5 fails just 1 test.
case 6 (`ENUM_CONSTANT_DEF`) does NOT fail tests.
case 7 fails tests.
case 8 (`LITERAL_SYNCHRONIZED`) does NOT fail tests.
case 9 fails tests.
case 10 fails just 1 test.

rnveach:
@Kevin222004 Leave this branch and issue up, I will continue to investigate. Feel free to move to a different suppression.

rnveach:
@Kevin222004 I believe the issue is ordering. This does actually have a bytecode connection as the internals of the switch have a different order than the source. We will need to update our wiki with this information.

I believe the tokens you want to investigate are `ENUM_CONSTANT_DEF` and `LITERAL_SYNCHRONIZED`.

rnveach:
> Report generation job failed on phase "make_report", step "Generate report".

Error that failed report is: `Caused by: java.lang.OutOfMemoryError: Java heap space`.
> Diff Regression config: https://gist.githubusercontent.com/Kevin222004/e206f8b15b0d8c04e1d6b1fbded90dbe/raw/6c6ca45c3c652fe259dc5844d874b1091554ae0b/config.xml

This may show that CI can't handle big loads and you may have to trim the projects or use your local with more memory.
One thought is for tokens, I don't believe you need to separate out individual tokens. They all could have been set as all the acceptable tokens.

romani:
Am I understand this correctly?

>`<description>RemoveSwitch 3 mutation`

Means 3rd group of cases in switch have survavals. If there are few cases, some of them.

rnveach:
> Means 3rd group of cases in switch have survavals. If there are few cases, some of them.

No. It doesn't count by groups, only by case. However order of cases is based on bytecode and not source.

rnveach:
Imagine you have:
````
case 3: // 1st case, 1st group
  ...; break;
case 1: // 2nd case, 2nd group
case 2: // 3rd case, 2nd group
  ...; break;
default: // default, 3rd group
  ...; break;
````
Pitest actually sees and counts:
````
default: // default
 functGroup3(); break; // group 3 logic
case 1: // 0th case
 functGroup2(); break; // group 2 logic
case 2: // 1st case
 functGroup2(); break; // group 2 logic
case 3: // 2nd case
 functGroup1(); break; // group 1 logic
````

**Comment by Kevin222004** [comment]
@rnveach thanks for the lots of information

>67 and 155 actually represent token IDs. 67 = LITERAL_SYNCHRONIZED (case 3) and 155 = ENUM_CONSTANT_DEF (case 7).

can you please tell how you decided it that the int 67 is for  LITERAL_SYNCHRONIZED and same for 155

---------------------------------
Is it expected that if we make changes and test tare not failing the mutation will kill 
I have done for the `LITERAL_SYNCHRONIZED` I change the label accordingly, tests are not failing but still mutation is not killed 
it is look like here i also cant create any test because all the different cases are available in input file and if it is not for what i can create it

**Comment by rnveach** [comment]
> can you please tell how you decided

I didn't decide, I determined. ANTLR is the one that decides on the numbering.

> that the int 67 is for LITERAL_SYNCHRONIZED and same for 155

Either System.out the value of these tokens or you can look in `JavaLanguageLexer` once the file is generated.

> Is it expected that if we make changes and test tare not failing the mutation will kill

If you apply the mutation, and tests are failing, then either mutation should have been killed, or there is other facts at play (like caching).
If you apply the mutation, and tests are passing, then the mutation is surviving and we need to determine what we can do to kill it.

> it is look like here i also cant create any test because all the different cases are available in input file and if it is not for what i can create it

Please do things in order.
Update the branch with only the mutation so we can make sure it is correct.
Once that is done, provide us regression. If regression finds differences, then confirm and trim them to an input.
If regression is clean, then tell us why in details why you can't create an input.

**Comment by Kevin222004** [comment]
Github, generate report

**Comment by Kevin222004** [comment]
@rnveach all the unit test are passing but  all the it testing are failing

**Comment by rnveach** [comment]
Only `src\test` is used for pitest. Anything else, like Checkstyle or such that is failing can be ignored for the mutation only branch.

**Comment by romani** [comment]
Сcan you copy code from https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/6aa50c2_2022192038/reports/diff/java-design-patterns/index.html#A1 to our Inputs to recheck is pitest is become killed

**Comment by Kevin222004** [comment]
@romani I have created the test file as you asked but it is throwing all the time exception that AST is null. All the things are looking good but, i don't understand that, I am updating the branch

**Comment by nrmancuso** [comment]
> but it is throwing all the time exception that AST is null. 

Every case of NullPointerException is a new test case that kills your mutation :)

**Comment by Kevin222004** [comment]
@nrmancuso Yes, the issue has been resolved i will update the branch

**Comment by Kevin222004** [comment]
https://kevin222004.github.io/solvedParenPad/com.puppycrawl.tools.checkstyle.checks.whitespace/ParenPadCheck.java.html
Line:-250

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheck.java [hunk 1]
```diff
@@ -262,8 +262,8 @@ public void visitToken(DetailAST ast) {
                 break;
             case TokenTypes.ANNOTATION:
             case TokenTypes.ENUM_CONSTANT_DEF:
-            case TokenTypes.LITERAL_NEW:
             case TokenTypes.LITERAL_SYNCHRONIZED:
+            case TokenTypes.LITERAL_NEW:
```
romani, line 266: Please restore order as it was before.

romani (changes requested):
Items

src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/parenpad/InputParenPadForSynchronized.java
```diff
@@ -0,0 +1,34 @@
...
+
+package com.puppycrawl.tools.checkstyle.checks.whitespace.parenpad;
+
+/**
+ *
+ * Thread-safe Singleton class.
+ * The instance is lazily initialized and thus needs synchronization
+ * mechanism.
+ *
+ */
```
romani, line 17: Please remove all unrelated to problem code.
We need to minimize code in input to keep only what matters.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheck.java
Kevin222004: done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/parenpad/InputParenPadForSynchronized.java
Kevin222004: done

romani (approved):
ok to merge

rnveach (approved):

Issue #12229, opened 2022-09-26
Rename all 'Checker-Framework / checker-..' actions 


romani:
On mobile long names are not good
We have "Checker-Framework / checker-..."
is what visible.
We need to rename them to be 'Checker / checker-...'

NitishKumar06:
I'm working on this.

@romani Could you provide a link/reference where could I find "Checker-Framerwork/checker-...", It'd be helpful!

romani:
Just search it in yml files, all is in sources of our repository

rnveach:
@romani This issue is complete, right?

PR #12269, opened 2022-10-07
Issue #12229: Renamed Checker-Frameework to Checker-checker

NitishKumar06:
Fix: #12229

PR #12324, opened 2022-10-22
Issue #12229: make Checker names more compact to show profile

romani:
Issue #12229


rnveach:
Is this taking over https://github.com/checkstyle/checkstyle/pull/12269 ? If so, I would comment or block out the other PR so it is clearer and make sure we don't have unnecessary duplicate efforts?

rnveach (commented):
> Checker / analyse (checker-framework-nullness-optional-interning) (pull_request) 

Is it possible and sense to try to make it `Checker / analyse (nullness-optional-interning)`?

rnveach:
Completely unrelated to this PR and can be done in another, but it looks like we can do the same for pitest.

> Pitest / pitest (pitest-coding-2) (pull_request)

rnveach (approved):
ok, didn't realize what is being displayed is the profile name. I don't know if there is a way to override github workflow display.

romani:
@rnveach , please look now.
Not ideal, but show enough to distinguish

romani:
Screenshot_20221022-072814

rnveach:
another option is to keep profile names as they were, and change checker groovy file to use full profile name.

romani:
lets ask @nrmancuso  for second opinion

.github/workflows/checker-framework.yml
```diff
@@ -7,19 +7,19 @@ on:
...
-          - checker-framework-regex-property-key-compiler-message
-          - checker-framework-purity-value-returns
+          - checker-nullness-optional-interning
+          - checker-methods-resource-fenum
+          - checker-lock-tainting
+          - checker-index
+          - checker-formatter
+          - checker-signature-gui-units-init
+          - checker-regex-property-key-compiler-message
+          - checker-purity-value-returns
```
nrmancuso, line 22: Let's do this below, in job:
```yml

      - name: Execute checker framework
        env:
          MAVEN_OPTS: "-Xmx4g"
        run: groovy ./.ci/checker-framework.groovy checker-${{ matrix.profile }}

```

then we can just do: `- purity-value-returns`

.github/workflows/checker-framework.yml
```diff
@@ -7,19 +7,19 @@ on:
   pull_request:
 
 jobs:
-  checker-framework:
+  analyse:
```
nrmancuso, line 10: We can rename to just be `check` to save two more letters for phone-friendly CI diagnostics :)

.github/workflows/checker-framework.yml
romani: done

.github/workflows/checker-framework.yml
romani: done

romani:
Screenshot_20221022-075515

rnveach (approved):

nrmancuso (approved):
Much easier to read and live with.

nrmancuso:
We missed new prefix in other job steps, PR is updated

nrmancuso (approved):
We needed to rename the files as well, since they contain profile name. It is better to keep all consistent.

Issue #12326, opened 2022-10-22
Resolve Pitest suppression for AccessModifierOption


shubh220922:
Reference from https://github.com/checkstyle/checkstyle/issues/7797

Kill the mutation

https://github.com/checkstyle/checkstyle/blob/4c7c3cc832a2b6ddefa529a816008880a9db9255/.ci/pitest-suppressions/pitest-naming-suppressions.xml#L12-L19

shubh220922:
I am on it.

romani:
Fix is merged

PR #12340, opened 2022-10-23
Issue #12326: Resolve Pitest suppression for AccessModifierOption

shubh220922:
Referencing from #12326 
also related to #12341

image

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/naming/ParameterNameCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/naming/ParameterNameCheckTest.java
@@ -69,6 +69,17 @@ public class ParameterNameCheckTest
     }
 
     @Test
+    public void testWhitespaceInAccessModifierProperty() throws Exception {
+        final String pattern = "^h$";
+        final String[] expected = {
+            "14:69: " + getCheckMessage(MSG_INVALID_PATTERN, "parameter1", pattern),
+            "18:31: " + getCheckMessage(MSG_INVALID_PATTERN, "parameter2", pattern),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputParameterNameWhitespaceInAccessModifierProperty.java"), expected);
+    }
+
+    @Test
     public void testDefault()
             throws Exception {
         final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/naming/parametername/InputParameterNameWhitespaceInAccessModifierProperty.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/naming/parametername/InputParameterNameWhitespaceInAccessModifierProperty.java
@@ -1,0 +1,23 @@
+/*
+ParameterName
+format = ^h$
+ignoreOverridden = (default)false
+accessModifiers = \tpublic
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.naming.parametername;
+
+public class InputParameterNameWhitespaceInAccessModifierProperty {
+
+    public InputParameterNameWhitespaceInAccessModifierProperty(int parameter1) {} // violation
+
+    public void v1(int h) { // ok
+        new Object () {
+            public void i(int parameter2) {} // violation
+        };
+    }
+
+    void method2(int V3) {}
+}
--- a/.ci/pitest-suppressions/pitest-naming-suppressions.xml
+++ b/.ci/pitest-suppressions/pitest-naming-suppressions.xml
@@ -10,15 +10,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>AccessModifierOption.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.naming.AccessModifierOption</mutatedClass>
-    <mutatedMethod>getInstance</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to java/lang/String::trim with receiver</description>
-    <lineContent>return valueOf(AccessModifierOption.class, modifierName.trim().toUpperCase(Locale.ENGLISH));</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>ParameterNameCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.naming.ParameterNameCheck</mutatedClass>
     <mutatedMethod>setAccessModifiers</mutatedMethod>


shubh220922:
Github, generate report

romani:
We do not allow pure Junit test methods anymore.
Please make coverage by Input files and real Checks.

shubh220922:
ok @romani

nrmancuso:
@shubh220922 please hardcode mutation first and try to understand why this mutation survives our tests.

shubh220922:
@nrmancuso okay going through this.

nrmancuso:
@shubh220922 also, please keep hack like https://github.com/checkstyle/checkstyle/commit/4f9c2a47f9e4941f79559626a00b217ccdb35ad4 in mind to kill mutation (using config input to kill mutation)

shubh220922:
@nrmancuso Should I write input files for the tests previously present in the file as well or only for mine.

nrmancuso:
@shubh220922 please do not modify existing test inputs

shubh220922:
I'll work on this pr within the next 3 days. My exams are going on.

shubh220922:
@nrmancuso I've been trying to understand the meaning of ```replaced call to ..... with receiver```. What is this receiver? Also I found this term ```naked receiver```. I surfed all of the internet for this but couldn't understand what it means. Could you please help me figure this out.
image

nrmancuso:
@shubh220922 please read pitest documentation for this mutation operator; share specific points you need clarification on https://pitest.org/quickstart/mutators/#EXPERIMENTAL_NAKED_RECEIVER

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/AccessModifierOption.java
```diff
@@ -63,7 +63,7 @@ private String getName() {
      * @return the AccessModifier associated with given access modifier name.
      */
     public static AccessModifierOption getInstance(String modifierName) {
-        return valueOf(AccessModifierOption.class, modifierName.trim().toUpperCase(Locale.ENGLISH));
+        return valueOf(AccessModifierOption.class, modifierName.toUpperCase(Locale.ENGLISH));
```
romani, line 66: https://github.com/checkstyle/checkstyle/pull/12263/files#diff-81239df2f2f1b6a10b954c1086a273293352aa3dde3554cc04ce0005661f673cR3

romani (changes requested):
item:

shubh220922:
Okay so I have studied PIT as much as I could and the following is what I have understood;

1. PIT mutates each line by changing some bits of code on that line for example it can simply remove the ```trim()``` in the line
```return valueOf(AccessModifierOption.class, modifierName.trim().toUpperCase(Locale.ENGLISH));``` and make it 
```return valueOf(AccessModifierOption.class, modifierName.toUpperCase(Locale.ENGLISH));```

_Well why is it doing that?_ 
Simply to check whether the test suits that we have already written are able to catch that change in code or not. If all our tests pass without a problem even though ```trim()``` was removed then that means that our test suits are not good/enough so we need to write more tests to catch that change in code that PIT made (or kill that mutation).

Now some parts where I have a question:

Question1)

Q.1.1) What is the purpose of input files at all. 

Q.1.2) How running tests on input files through ```verifyWithInlineConfigParser()``` method and showing respective error on respective lines in input files going to kill mutation in a totally different java file where the mutation was reported through PIT.

Q.1.3) **Most importantly why is this way of testing that checkstyle is using better than writing pure JUnit tests?**

Question2) 

Q.2.1) Say if I make the input files then how do I know what code to write in it. 

Q.2.2) How violations being reported on the lines in the input files going to fix the mutation? 

Q.2.3) How do we know what config to mention at the top of the input files as the documentation is not present for ```AccessModifierOption``` as I think it is an enum thats why.

_I do have a lot in my head but I can't visually graph what is happening between the files in the sense which file is communicating with which file for what reason. As I know about conventional pure JUnit test cases, it is difficult to understand this a little bit and I cannot proceed further until some help is provided. I need to connect what is happening._

@nrmancuso @romani

**Comment by nrmancuso** [comment]
@shubh220922 please move comment at https://github.com/checkstyle/checkstyle/pull/12340#issuecomment-1317104456 to a discussion, this is probably a better place for such general questions than a PR

**Comment by shubh220922** [comment]
@nrmancuso I have moved to discussion.

**Comment by shubh220922** [comment]
Now in order to find what violation I will get at respective lines, I will have to check through running the config but the following is being displayed:

```
PS C:\Users\shubh\OneDrive\Desktop\checkstyle project> cat config.xml                                                                      
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN"

<module name="Checker">
    <property name="charset" value="UTF-8" />
    <module name="TreeWalker">
        <module name="AccessModifierOption">
        </module>
    </module>
</module>
PS C:\Users\shubh\OneDrive\Desktop\checkstyle project> cat test.java                                                                       
/*
AccessModifierOption
option = \tSPACE


*/

package com.puppycrawl.tools.checkstyle.checks.naming.accessmodifieroption;

public class InputAccessModifierOption {

    class classOne {

    }

    class classTwo {

    }

    class classThree {

    }

    public void method1() {

    }

    public void method2() {

    }

}
PS C:\Users\shubh\OneDrive\Desktop\checkstyle project> java -jar "C:\Users\shubh\Downloads\checkstyle-10.4-all.jar" -c config.xml test.java
com.puppycrawl.tools.checkstyle.api.CheckstyleException: cannot initialize module TreeWalker - cannot initialize module AccessModifierOption - Unable to instantiate 'AccessModifierOption' class, it is also not possible to instantiate
        ... 5 more
Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Unable to instantiate 'AccessModifierOption' class, it is also not possible to instantiate it as .AccessModifierOption, AccessModifierOptionCheck, .AccessModifierOptionCheck. Please recheck that class name is specified as canonical name or read how to configure short name usage https://checkstyle.org/config.html#Packages. Please also recheck that provided ClassLoader to Checker is configured correctly.
        at com.puppycrawl.tools.checkstyle.PackageObjectFactory.createModule(PackageObjectFactory.java:214)
        at com.puppycrawl.tools.checkstyle.TreeWalker.setupChild(TreeWalker.java:120)
        ... 7 more

```


It is giving me this error.

**Comment by nrmancuso** [comment]
@shubh220922 you found your failing test case ;)

**Comment by nrmancuso** [comment]
@shubh220922 you must use a check that uses AccessModifierOptions

**Comment by shubh220922** [comment]
@nrmancuso i have made some changes.. Thanks to your link given at https://github.com/checkstyle/checkstyle/pull/12340#issuecomment-1318476079.

But I am getting this error when I am running the test that I have made:

image

**Comment by nrmancuso** [comment]
@shubh220922 please only create a new test and input in JavadocMethodCheckTest, with tab character added to access modifier property value. You only need a very simple input, the idea is that we need check to read the property in only.

**Comment by shubh220922** [comment]
Okay fixed some things. Mutation doesn't seem to be killed. Can you check my changes

**Comment by shubh220922** [comment]
Mutation didn't get killed. What am I missing @nrmancuso.

**Comment by Kevin222004** [comment]
@nrmancuso this is really quite suspicious. The current test case must resolve the mutation according to the config but it is not.
I have also make test to resolve it and it won't work if we are creating the test cases for the check which don't belong to package naming in which enum AccessModiferOption is created  https://github.com/checkstyle/checkstyle/tree/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming. 

I have created a test case in one of this check from naming package and mutation were killed.

i am also curious to know why this is happening :), because according to the concept of mutation testing if we don't have strong enough test case for the code which we write so it will be mutated and in this currently updated pr this test is enough to kill trim
func mutatiion

**Comment by nrmancuso** [comment]
@shubh220922 this means that the string is getting trimmed before we reach this method (`AccessModifierOption#getInstance`). Debug the code and follow the call stack to investigate. Provide your full analysis here.

**Comment by shubh220922** [comment]
@nrmancuso the following is what I got from debugging:

I have started debugging from ```valueOf(AccessModifierOption.class, modifierName.trim().toUpperCase(Locale.ENGLISH));``` as in the image below:

image

I specially tagged this ```modifierName="\tpublic"``` to understand whether it is getting trimmed before reaching ```AccessModifierOption.getInstance()``` or not.

The value of ```token``` below is ```\tpublic``` which means that ```AccessModifierOption.getInstance(token)``` is receiving the non-trimmed version of ```\tpublic``` as in the image below:

Image1>
image

But in the Image2 ```result``` seems to contains ```public```.
image

Following the stack of execution the value of ```value``` as in the image stays ```\tpublic``` until it reaches the line ```value=convertForCopy(value, type);``` as in the image below:

image

After crossing this line, the value of ```value``` starts to be shown as```AccessModifierOption[1]@2407```

image

I am thinking where exactly is the problem coming from?!

**Comment by nrmancuso** [comment]
@shubh220922 ok, we have missed a nuance here. `AccessModifierOption` lives in the `naming` package, so due to pitest config in pom.xml we only target tests in the `naming` package. Please do:
1. Remove new test + input
2. Add test + input (with same `\t` trick in config) using https://checkstyle.sourceforge.io/config_naming.html#ParameterName or some other naming check
3. Create issue to move `AccessModifierOption` to `api` package, since we use this enum outside of `naming` and plan to use it more extensively.

**Comment by romani** [comment]
>Create issue to move AccessModifierOption to api package,

We do all possible to avoid API package.
We want to really treat API as contract. So if we can avoid api we should avoid.

**Comment by nrmancuso** [comment]
> > Create issue to move AccessModifierOption to api package,
> 
> We do all possible to avoid API package. We want to really treat API as contract. So if we can avoid api we should avoid.

@romani let's create issue to move AccessModifierOption, we can discuss final location in issue. It doesn't make sense to keep AccessModifierOption in naming if we use it elsewhere.

**Comment by shubh220922** [comment]
@nrmancuso mvn passed successfully. Mutation killed. 
image

**Comment by shubh220922** [comment]
Q.1)
@nrmancuso as you mentioned in https://github.com/checkstyle/checkstyle/pull/12340#issuecomment-1323675196 about the config of pitest in pom.xml preventing us from killing mutation of ```accessModifierOption``` in other packages, does this rule apply to all the packages? We necessarily need to make input and test in the same package as the surviving mutation always? 

Q.2) How did you figure this out in https://github.com/checkstyle/checkstyle/pull/12340#issuecomment-1323675196 as I wasn't able to reach any conclusion through debugging.

**Comment by nrmancuso** [comment]
@shubh220922 test was right, but mutation was not killed. This means that we did not use this test to cover the mutation we are working on. We cannot put all tests on all packages, as pitest would take a very long time to run. I initially missed this since mutation class was suppressed in `naming` package, but used in other packages. This is what prompted me to mention that we should create an issue to move AccessModifiers to a more general location.

**Comment by shubh220922** [comment]
@nrmancuso 

> This means that we did not use this test to cover the mutation we are working on.

Please see if my interpretation of this is correct or not:

### Statement 1

_Due to certain config of pitest present in pom.xml that **we have set**, **it only allows for killing those mutations for which the tests and inputs are written within the same package as the main java file containing mutation**_

_It is thereby not possible to write tests and inputs in a different package than the package that contains the file containing the mutation_ **due to the config of pit in pom.xml**

### Question 1.1
Did _**we**_ set this config of PIT in pom.xml? Or it is given by default by PIT?

### Question 2

> I initially missed this since mutation class was suppressed in `naming` package
Could you please help me understand this. 

### Question 2.1
Where is the suppression exactly? 
### Question 2.2
Which mutation class is being suppressed here? 
### Question 2.3
The suppression of the mentioned **mutation class** is suppressing what function here?

**Comment by shubh220922** [comment]
@nrmancuso should i create issue to move ```AccesssModifierOption``` to ```api``` package?

**Comment by nrmancuso** [comment]
@shubh220922 https://github.com/checkstyle/checkstyle/issues/12445

**Comment by nrmancuso** [comment]
@shubh220922 

>Did we set this config of PIT in pom.xml? Or it is given by default by PIT?

See for yourself in the pom file.

>Where is the suppression exactly?

This is what you have removed in this PR, read the suppression you've removed for all details. To be clear: most pitest mutation suppression files are named by package, like `pitest-<package name>suppressions.xml`, so when I say that it was suppressed in naming package, this is what I was referring to.

File: src/test/java/com/puppycrawl/tools/checkstyle/checks/naming/ParameterNameCheckTest.java [hunk 2]
```diff
@@ -68,6 +68,17 @@ public void testSpecified()
                 getPath("InputParameterNameOne.java"), expected);
     }
 
+    @Test
+    public void testTrimPropertyInConfig() throws Exception {
```
nrmancuso, line 72: `testWhitespaceInAccessModifierProperty`

src/test/resources/com/puppycrawl/tools/checkstyle/checks/naming/parametername/InputParameterNameAccessModifierChecker.java
```diff
@@ -0,0 +1,23 @@
...
+format = ^h$
+ignoreOverridden = (default)false
+accessModifiers = \tpublic
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.naming.parametername;
+
+public class InputParameterNameAccessModifierChecker {
```
nrmancuso, line 12: `InputParameterNameWhitespaceInAccessModifierProperty`, we have another tool called checker, let's avoid this term.

nrmancuso (changes requested):
Items:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/naming/parametername/InputParameterNameAccessModifierChecker.java
shubh220922: Done.

src/test/java/com/puppycrawl/tools/checkstyle/checks/naming/ParameterNameCheckTest.java
shubh220922: Done.

nrmancuso (approved):

shubh220922:
Travis CI failing! Fixing it.

romani (approved):

src/test/resources/com/puppycrawl/tools/checkstyle/checks/naming/parametername/InputParameterNameWhitespaceInAccessModifierProperty.java
```diff
@@ -0,0 +1,23 @@
+/*
+ParameterName
+format = ^h$
+ignoreOverridden = (default)false
+accessModifiers = \tpublic
```
nrmancuso, line 5: @rnveach

rnveach: > config like this: public\t,\t,

nrmancuso: In other PR, we had `public, \t`, case taken from this PR is with tab character "on" the access modifier. The first trim in both methods took care of the single `\t`, the second trim takes care of the `\t` "on" the access modifier.

PR #12342, opened 2022-10-24
Issue #12243: Improve message capture

### LINKED ISSUE #12243: Pitest/Checker: Drop New/Unnecessary Displays And Switch to Git Diff

nrmancuso:
Closes #12243

### DIFF
--- a/.ci/checker-framework-suppressions/checker-formatter-suppressions.xml
+++ b/.ci/checker-framework-suppressions/checker-formatter-suppressions.xml
@@ -1,28 +1,28 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedErrors>
   <checkerFrameworkError unstable="false">
-    <fileName>checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/LocalizedMessage.java</fileName>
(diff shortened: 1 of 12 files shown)


nrmancuso:
closed, in favor of https://github.com/checkstyle/checkstyle/pull/12368

Issue #5385, opened 2017-12-25
Convert sevntu-checkstyle-check to ant run


rnveach:
See https://issues.apache.org/jira/browse/MCHECKSTYLE-346 .
We need to deprecate all usage of `maven-checkstyle-plugin` so we can start implementing backward breaking changes.

We must convert our `sevntu-checkstyle-check` run to an ant run.
Specified section: https://github.com/checkstyle/checkstyle/blob/06ff5677fa5a19afe3f55078fff79ead61068a51/pom.xml#L558-L597

romani:
this was  interesting hack a while ago, as it make self dependency of snapshot checkstyle to previous checkstyle version `${maven.sevntu-checkstyle-check.checkstyle.version}`.

It was not done as ANT, only because I personally do not like ANT at all, so I try to minimize its usage as much as reasonable. May be is it a reason to make it by ANT....

lets come initial reasons ...

@rnveach , what force us to migrate from maven-checkstyle-plugin here ?
braking changes like at #5324 , was not detected by this validation....
so to make it clear "we must" or "we should" migrate.

rnveach:
> braking changes like at #5324 , was not detected by this validation....

This maven plugin run doesn't use snapshot version of checkstyle. If we used snapshot version, problem would be the same. This is why we only saw issues in wercker, as wercker used maven plugin with snapshot version.

If we leave this area the same, it will be broken once we release and try to upgrade our project to new Checkstyle version.

romani:
Ok, make sense.
I am ok with change, but ant is not an option, we can use maven exec plugin to run validation as by cli execution. It might be easier to setup, it will be based on release version of checkstyle again.
If you are ok, please remove ant from title and let's keep it as option of implementation.

nrmancuso:
Closed via #12346

PR #12346, opened 2022-10-26
Issue #5385: convert maven-checkstyle-plugin to ant

rnveach:
Issue #5385

New profile is needed, see note.

### DIFF
--- a/.ci/validation.cmd
+++ b/.ci/validation.cmd
@@ -9,8 +9,13 @@
 SET OPTION=%1
 
 if "%OPTION%" == "sevntu" (
-  call mvn -e --no-transfer-progress verify -DskipTests -DskipITs -Dpmd.skip=true^
-    -Dspotbugs.skip=true -Djacoco.skip=true -Dxml.skip=true^
+  call mvn -e --no-transfer-progress verify -Pno-validations,sevntu^
+    || goto :ERROR
+  goto :END_CASE
+)
+
+if "%OPTION%" == "run_checkstyle" (
+  call mvn -e --no-transfer-progress clean compile antrun:run@ant-phase-verify^
     || goto :ERROR
   goto :END_CASE
 )
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -662,9 +662,9 @@ no-error-checkstyles-sevntu)
   CS_POM_VERSION="$(getCheckstylePomVersion)"
   echo CS_version: "${CS_POM_VERSION}"
   mvn -e --no-transfer-progress clean install -Pno-validations
-  mvn -e --no-transfer-progress compile verify \
+  mvn -e --no-transfer-progress compile verify -Psevntu \
     -Dmaven.sevntu-checkstyle-check.checkstyle.version="${CS_POM_VERSION}" \
-    -Dmaven.test.skip=true -Dcheckstyle.ant.skip=true -Dpmd.skip=true -Dspotbugs.skip=true \
+    -Dmaven.test.skip=true -Dpmd.skip=true -Dspotbugs.skip=true \
     -Djacoco.skip=true -Dforbiddenapis.skip=true -Dxml.skip=true
   ;;
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,14 @@ jobs:
         - DEPLOY="true"
         - USE_MAVEN_REPO="true"
 
+    # sevntu (openjdk11)
+    - jdk: openjdk11
+      env:
+        - DESC="sevntu"
+        - CMD="mvn -e --no-transfer-progress compile antrun:run@ant-phase-verify-sevntu -Psevntu"
+        - DEPLOY="true"
+        - USE_MAVEN_REPO="true"
+
     # codecov
     - jdk: openjdk11
       env:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -39,10 +39,14 @@ environment:
   # We do matrix as AppVeyor could fail to finish simple "mvn -e verify"
   #    if he loose maven cache (happens from time to time)
   matrix:
-    # checkstyle and sevntu.checkstyle (JDK11)
+    # sevntu (JDK11)
     - JAVA_HOME: C:\Program Files\Java\jdk11
-      DESC: "checkstyle and sevntu.checkstyle (JDK11)"
+      DESC: "sevntu (JDK11)"
       CMD: "./.ci/validation.cmd sevntu"
+    # run checkstyle (JDK11)
+    - JAVA_HOME: C:\Program Files\Java\jdk11
+      DESC: "run checkstyle (JDK11)"
+      CMD: "./.ci/validation.cmd run_checkstyle"
     # verify without checkstyle (JDK11)
     - JAVA_HOME: C:\Program Files\Java\jdk11
       DESC: "verify without checkstyle (JDK11)"
--- a/pom.xml
+++ b/pom.xml
@@ -652,55 +652,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-checkstyle-plugin</artifactId>
-        <version>${maven.checkstyle.plugin.version}</version>
-        <dependencies>
-          <dependency>
-            <groupId>com.github.sevntu-checkstyle</groupId>
-            <artifactId>sevntu-checks</artifactId>
-            <version>${maven.sevntu.checkstyle.plugin.version}</version>
-          </dependency>
-          <dependency>
-            <groupId>com.puppycrawl.tools</groupId>
-            <artifactId>checkstyle</artifactId>
-            <version>${maven.sevntu-checkstyle-check.checkstyle.version}</version>
-          </dependency>
-        </dependencies>
-        <!-- Specifying configuration here will take effect on the execution
-             of "mvn checkstyle:checkstyle",
-             but will not take effect on the execution of "mvn site"  -->
-        <executions>
-          <execution>
-            <id>sevntu-checkstyle-check</id>
-            <phase>verify</phase>
-            <configuration>
-              <configLocation>
-                ${project.basedir}/config/checkstyle_sevntu_checks.xml
-              </configLocation>
-              <failOnViolation>true</failOnViolation>
-              <includeResources>false</includeResources>
-              <includeTestResources>false</includeTestResources>
-              <logViolationsToConsole>true</logViolationsToConsole>
-              <maxAllowedViolations>0</maxAllowedViolations>
-              <violationSeverity>error</violationSeverity>
-              <cacheFile>${project.build.directory}/cachefile_sevntu</cacheFile>
-              <sourceDirectories>
-                <sourceDirectory>${project.basedir}/src</sourceDirectory>
-              </sourceDirectories>
-              <excludes>
-                **/it/resources/**/*,**/it/resources-noncompilable/**/*,
-                ,**/test/resources/**/*,**/test/resources-noncompilable/**/*,
-                **/gen/**
-              </excludes>
-            </configuration>
-            <goals>
-              <goal>check</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
 
       <plugin>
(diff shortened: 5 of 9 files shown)


rnveach:
Sevntu will have to be updated at https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/5cc4928729a91f71fc3d674a06ea6599fe8066fa/.ci/validation.sh#L102 when this is merged

.travis.yml
```diff
@@ -36,6 +36,14 @@ jobs:
         - DEPLOY="true"
         - USE_MAVEN_REPO="true"
 
+    # sevntu (openjdk11)
+    - jdk: openjdk11
+      env:
+        - DESC="sevntu"
+        - CMD="mvn -e --no-transfer-progress clean verify -Pno-validations,sevntu"
```
romani, line 43: If we want precise execution
Better to use exact plugin execution, try  https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations#how-to-generate-checkstyle-report

.ci/validation.sh
```diff
@@ -669,9 +669,9 @@ no-error-checkstyles-sevntu)
   CS_POM_VERSION="$(getCheckstylePomVersion)"
   echo CS_version: "${CS_POM_VERSION}"
   mvn -e --no-transfer-progress clean install -Pno-validations
-  mvn -e --no-transfer-progress compile verify \
+  mvn -e --no-transfer-progress compile verify -Psevntu \
```
romani, line 672: Do you think jar usage is better than classpath from target folder?
Potentially we can skip extra build step and run it more easily on local.
Or it is good live example of ant usage of custom jar ?

.ci/validation.cmd
```diff
@@ -9,6 +9,12 @@
 SET OPTION=%1
 
 if "%OPTION%" == "sevntu" (
+  call mvn -e --no-transfer-progress verify -Pno-validations,sevntu^
+    || goto :ERROR
+  goto :END_CASE
+)
+
+if "%OPTION%" == "verify_with_checkstyle" (
   call mvn -e --no-transfer-progress verify -DskipTests -DskipITs -Dpmd.skip=true^
```
romani, line 18: Why we need this job now?

pom.xml
```diff
@@ -2169,14 +2118,64 @@
...
+                    <ant antfile="config/ant-phase-verify-sevntu.xml" />
+                  </target>
+                </configuration>
+              </execution>
+            </executions>
+            <dependencies>
+              <dependency>
+                <groupId>com.puppycrawl.tools</groupId>
+                <artifactId>checkstyle</artifactId>
+                <version>${maven.sevntu-checkstyle-check.checkstyle.version}</version>
```
romani, line 2159: This variable might need rename to skip maven.
As all is done by ant

romani (changes requested):
Items

.ci/validation.cmd
rnveach: This and the changes to appveyor.yml is to keep CI running the same things as before my change.

Before when we ran `verify` phase, it ran our ANT runs and maven-checkstyle-plugin run all in 1 go.

Now because we had to move sevntu run to a profile, it requires 2 separate commands.

I thought this would be better than combining both calls and results into 1 appveyor run. Similar to how we have other CIs split simple things out into different runs.

pom.xml
rnveach: I assumed `maven` was not related to the maven plugin as there are other checkstyle version properties that start with maven also.

maven.sevntu.checkstyle.plugin.version
maven.sevntu-checkstyle-check.checkstyle.version

Should I rename them all and just remove `maven` from the beginning since they are dependency versions and it doesn't really matter how they are run.

.ci/validation.sh
rnveach: > classpath from target folder

I am unsure what this part of the statement means. Sevntu profile run is using officially released checkstyle version. It is not using the master version, which is what is stored in target. Only results and such are put into target folder.

> Potentially we can skip extra build step and run it more easily on local.
> Or it is good live example of ant usage of custom jar ?

If your referring to JAR usage, we technically have to package the JAR for us to use it. Also, if we switch to JAR usage, we have to then go through the trouble of downloading the sevntu dependency to add it and use it in the run. I think it is easier to maintain if we continue to rely on maven.

I also believe we added plenty of examples to https://github.com/sevntu-checkstyle/checkstyle-samples and can point people to that if they need different ways of running checkstyle.

.ci/validation.cmd
romani: Name is very confusing. I am not sure how to name it better

.ci/validation.sh
romani: We build snapshot jar

And reference it at
`Dmaven.sevntu-checkstyle-check.checkstyle.version="${CS_POM_VERSION`
We did this as maven does not have ability to work on folder with classes.
As we use our own ant task, we can do same as main execution.

We want to test that code in PR is not making breaking changes to sevntu.

.ci/validation.cmd
romani: If we need only checkstyle to be executed, can we do 
https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations#how-to-generate-checkstyle-report

.ci/validation.sh
rnveach: @romani What is the change you want me to make? I am not understanding if it was stated.

pom.xml
rnveach: Is this a change to make in this PR?

.ci/validation.sh
romani: I still do not see reason to run sevntu as separate profile.
With new profile, we stik sevntu from "mvn verify"  - bad.

I do not see good reason to keep running on released version of checkstyle. We need to run on checkstyle target folder + sevntu jar.

We use sevntu as maven execution only as convenience, or my lack of knowledge at that time to run ANT smartly.
As we remove maven plugin,  sevntu should join our regular ant phase, it is what we need and it will be more beneficial.

So there will be no special profile, sevntu runs on freshly compiled sources.
It will allow us to remove special job in CI where we try to run sevntu on SNAPSHOT checkstyle jar.

pom.xml
romani: please read https://github.com/checkstyle/checkstyle/pull/12346/files#r1042257603 , it might result in natural steep to "My recommendations"

.travis.yml
romani: please read https://github.com/checkstyle/checkstyle/pull/12346#discussion_r1042257603

.ci/validation.sh
rnveach: If we want to run Sevntu on an unsupported version of Checkstyle, then lets make this another PR with its own separate isolation.

But just to note, it is a bad idea to run Sevntu on unsupported versions of Checkstyle. When sevntu was stuck on 10.0, it didn't make sense to try and run it on 10.4, and if there was a breaking enough change, this means DISABLING the entire sevntu run, even though we could just run it against the supported 10.0 . Checking sevntu works with the latest CS version is a separate no-violation regression that isn't tested here.

romani: from rnveach:

my concern I brought up is we were breaking sevntu

Your saying you want to force sevntu to run on an unconfirmed version of CS by running on snapshot instead of release

we only use release version of sevntu, so I am less worried about sevntu doing a breaking change

The question basically comes down to, if CS starts doing breaking changes. Do you still want to run sevntu against our code base or not? If you do, we can't run sevntu with the snapshot version of CS.

Imagine we had to change the name of something common between the 2. CI will report the issue in CS snapshot. Sevntu which relies on the released version of CS, can't make a change for something that is breaking in snapshot. It will never compile since the change is only in snapshot.

There are things only we can find in CI as it is. Checker and ArchUnit is one. Pitest is a better example.

romani: I am ok to keep us less dependent on satellites as whole idea of migration from maven to ant to let us move forward easier with breaking changes and let satellites to catch up.

.ci/validation.cmd
romani: better to rename as suggested above.
and update command to make only ANT execution of checksyle.

.travis.yml
romani: better to remove "verify" and execute ant maven task only.

pom.xml
romani: > My recommendations:

yes, we have small mess in naming 
https://github.com/checkstyle/checkstyle/blob/bef49aea4a923bc003b0ffd1338b8f1606f2badc/pom.xml#L209-L225

I think it is better to remove "maven" from name, we can do this in separate PR/issue, to not complicate current PR diff.

.travis.yml
rnveach: @romani So you are aware, the profile still exists so it still needs to be part of the command executed.

If that is ok, this is Done in the next push.

.ci/validation.cmd
rnveach: > name it run_checkstyle
> only checkstyle to be executed, can we do

Both are done.

rnveach:
Review items are Done, imo.

.travis.yml
```diff
@@ -34,6 +34,16 @@ jobs:
         - DEPLOY="true"
         - USE_MAVEN_REPO="true"
 
+    # sevntu (openjdk11)
+    - jdk: openjdk11
+      env:
+        - DESC="sevntu"
+        - CMD1="mvn -e --no-transfer-progress clean compile antrun:run@ant-phase-verify-sevntu"
+        - CMD2=" -Psevntu"
+        - CMD="$CMD1$CMD2"
```
rnveach, line 43: Line was too long, so I had to split the 2.

.travis.yml
romani: Yes it is better expected to have profile

pom.xml
romani: Please create issue to not forget to do this

.travis.yml
romani: Can we remove "clean" ?
Host is always clean after clone

pom.xml
rnveach: Done.

.travis.yml
rnveach: Done.

appveyor.yml
```diff
@@ -39,10 +39,14 @@ environment:
...
   matrix:
-    # checkstyle and sevntu.checkstyle (JDK11)
+    # sevntu (JDK11)
     - JAVA_HOME: C:\Program Files\Java\jdk11
-      DESC: "checkstyle and sevntu.checkstyle (JDK11)"
+      DESC: "sevntu (JDK11)"
       CMD: "./.ci/validation.cmd sevntu"
+    # verify with checkstyle (JDK11)
+    - JAVA_HOME: C:\Program Files\Java\jdk11
+      DESC: "verify with checkstyle (JDK11)"
```
romani, line 48: Remove 'verify'

romani (changes requested):
Items

appveyor.yml
rnveach: Done.

romani (approved):
Ok to merge

nrmancuso (changes requested):
From https://github.com/checkstyle/checkstyle/issues/5385#issuecomment-374036395:

>I am ok with change, but ant is not an option, we can use maven exec plugin to run validation as by cli execution. It might be easier to setup, it will be based on release version of checkstyle again.

This is the last comment in the issue, are we disregarding it? If we use the exec plugin, wouldn't we be able to have sevntu checks running in the `verify` phase again? It seems like a waste to spin up CI containers and run maven just for a 15 second task.

romani:
For now we don't know how to put it in `verify`

nrmancuso:
@rnveach did we try https://www.mojohaus.org/exec-maven-plugin/ as specified at https://github.com/checkstyle/checkstyle/issues/5385#issuecomment-374036395? This way we would be able to keep execution of sevntu (on release version of checkstyle) in `verify` phase.

nrmancuso (approved):
Ok, migration away from maven-checkstyle-plugin is a big step forward. I hope we can reunite sevntu checkstyle execution with `verify` phase one day.

PR #12349, opened 2022-10-28
Issue #11747: Migrate codenarc to travis CI

### LINKED ISSUE #11747: Move codenarc to non GitHub CI
From https://github.com/checkstyle/checkstyle/pull/11739#issuecomment-1157245833

I would move this check to Travis or other not high demand CI. It it most likely that such groovy script(s) will never be updated after initial creation, so better to have it run in some less critical CI.

Vyom-Yadav:
Resolves #11747 

### Proof of working:
Failing due to extra import - https://app.travis-ci.com/github/checkstyle/checkstyle/jobs/586922249


.ci/codenarc.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/bash
 # Attention, there is no "-x" to avoid problems on Travis
-set -e
```
Vyom-Yadav, line 3: No need to exit if `groovy ./.ci/codenarc.groovy` fails, we still need to echo the reason for failure, for example, if groovy is not installed or `JAVA_HOME` is not set.

romani: This is required to fail fast and do not continue execution I'd something existing with error code.

Please share more details and log on this update

romani (changes requested):
Items

.ci/codenarc.sh
Vyom-Yadav: Actually `set -e` can stay, error in groovy installation or `JAVA_HOME` would be shown in `before_script`

romani: Yes, placing `-e` in Travis level code is damaging it's execution, but in script it is ok

romani (approved):
Ok to merge

.travis.yml
```diff
@@ -43,13 +44,26 @@ jobs:
...
     - env:
         - DESC="ensure that all modules are used in no exception configs"
         - CMD1="export PULL_REQUEST=$TRAVIS_PULL_REQUEST"
         - CMD2="./.ci/validation.sh verify-no-exception-configs"
         - CMD="$CMD1 && $CMD2"
 
+before_script:
+  - |
+    if [[ $TRAVIS_CPU_ARCH == 'ppc64le' ]]; then
+        export JAVA_HOME=/usr/lib/jvm/adoptopenjdk-11-hotspot-ppc64el
```
rnveach, line 62: @romani Should this actually be in `./.ci/travis.sh install-adoptium-jdk` as it is the JRE and to keep it all together as needed for travis.

rnveach (approved):
Just the one comment.

.travis.yml
nrmancuso: Let's keep evidence of weird architecture in yml file, to remind us of this when adding new jobs here.

nrmancuso (approved):

Issue #12355, opened 2022-11-01
Feature request: add FATAL SeverityLevel


pcbouman-eur:
I have read check documentation: https://checkstyle.org/config_xxxxxx.html#NameOfAffectedCheck
I have downloaded the latest cli from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and show small examples below.

To summarize my feature request, I would like to have checkstyle support the same severity levels as PMD, which has a *fatal* severity level that checkstyle lacks. Technically, this seems to involve adding a single additional entry to the [SeverityLevel.java](../blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/SeverityLevel.java) enum and doing so seems to not break any tests or running `mvn package` on the full source tree of checkstyle.

**How it works Now:**

I have deviated from the template because it is not helpful in communicating my feature request clearly, given that checkstyle currently works as intended, as long as my configuration XML defines rules as follows:

```xml
		<module name="MissingJavadocType">
			<property name="severity" value="error" />
		</module>
```

I am using both checkstyle and PMD in a automated grading platform (Codegrade) to generate automated style feedback for student assignments. This platform supports four severity levels for general linters: "info", "warning", "error" and "fatal", which all exist in PMD, but of which only three exist in checkstyle.

What I would like define a rule as follows (note the severity level `fatal`), so I can use the same levels that are available in PMD. Checkstyle obviously fails reading in the configuration because I use a non-existing SeverityLevel:

```xml
		<module name="MissingJavadocType">
			<property name="severity" value="fatal" />
		</module>
```

Which leads to the CLI output:

```
$ java -jar checkstyle-10.4-all.jar -c checkstyle.xml FruitBasket.java
Exception in thread "main" java.lang.IllegalArgumentException: No enum constant com.puppycrawl.tools.checkstyle.api.SeverityLevel.FATAL
        at java.base/java.lang.Enum.valueOf(Enum.java:273)
        at com.puppycrawl.tools.checkstyle.api.SeverityLevel.getInstance(SeverityLevel.java:67)
        at com.puppycrawl.tools.checkstyle.ConfigurationLoader$InternalLoader.endElement(ConfigurationLoader.java:575)
        .... and the stack trace goes on
```

In practice, if I still want to use the fatal severity level supported by my other tooling, I have two options:

* Work with the `ignored` severity level and use wrapper scripts remap the output of checkstyle bumping up the severity levels by one
[...]

PR #12356, opened 2022-11-01
doc: expanded documentation on severities

rnveach:
Created for https://github.com/checkstyle/checkstyle/issues/12355#issuecomment-1298701324 and not connected to the issue since this is just for documenting what was seen as missing, imo.

### DIFF
--- a/src/xdocs/property_types.xml
+++ b/src/xdocs/property_types.xml
@@ -831,16 +831,88 @@
 
     <section name="SeverityLevel">
       <p>
-        This type represents the severity level of a check violation. The
-        valid options are:
+        This type represents the severity level of a module violation.
+        The following table describes the valid options:
       </p>
 
-      <ul>
-        <li><code>ignore</code></li>
-        <li><code>info</code></li>
-        <li><code>warning</code></li>
-        <li><code>error</code></li>
-      </ul>
+      <div class="wrapper">
+        <table>
+          <tr>
+            <td>Option</td>
+            <td>Definition</td>
+            <td><a href="cmdline.html">CLI</a></td>
+            <td><a href="anttask.html">Ant task</a></td>
+          </tr>
+
+          <tr>
+            <td><code>ignore</code></td>
+            <td>
+              If a violation occurs, modules with this severity are ignored as if they never
+              occurred.
+              This severity can be used to disable a module in the configuration.
+            </td>
+            <td>
+              Such violation not printed in output, however,
+              depending on the setting of
+              <a href="cmdline.html#Command_line_usage"><code>executeIgnoredModules</code></a>
+              when executed, the module may still run and if an exception occurs, the exception
+              will turn into an error and act as one.
+            </td>
+            <td>
+              Such violation not printed in output, however,
+              depending on the setting of
+              <a href="anttask.html#Parameters"><code>executeIgnoredModules</code></a>
+              when executed, the module may still run and if an exception occurs, the exception
+              will turn into an error and act as one.
+            </td>
+          </tr>
+
+          <tr>
+            <td><code>info</code></td>
+            <td>
+              If a violation occurs, modules with this severity are displayed as informational.
+            </td>
+            <td>
+              Displays the violation, but not fail the execution.
+            </td>
+            <td>
+              Displays the violation, but not fail the execution.
+            </td>
+          </tr>
+
+          <tr>
+            <td><code>warning</code></td>
+            <td>
+              This severity behaves exactly the same as <code>info</code> in Checkstyle.
+              It can be used to let user bring more attention to reviewers of such violations.
+            </td>
+            <td>
+              Displays the violation, but not fail the execution.
+            </td>
+            <td>
+              Displays the violation, might fail the execution.<br/>
+              <a href="anttask.html#Parameters"><code>maxWarnings</code></a> can be used to pass
+              the execution until a certain number of warnings are found.
+            </td>
+          </tr>
+
+          <tr>
+            <td><code>error</code></td>
+            <td>
+              If a violation occurs, modules with this severity are displayed as error and should
+              be considered as a failure.
+            </td>
+            <td>
+              Displays the violation, fail the execution.
+            </td>
+            <td>
+              Displays the violation, might fail the execution.<br/>
+              <a href="anttask.html#Parameters"><code>maxErrors</code></a> can be used to pass
+              the execution until a certain number of error are found.
+            </td>
+          </tr>
+        </table>
+      </div>
     </section>
 
     <section name="String">


rnveach:
run inspections is unrelated to this PR.

src/xdocs/property_types.xml
```diff
@@ -831,16 +831,62 @@
...
+            <td>Option</td>
+            <td>Definition</td>
+          </tr>
+
+          <tr>
+            <td><code>ignore</code></td>
+            <td>
+              If a violation occurs, modules with this severity are ignored as if they never
+              occurred.
+              Checkstyle will not display the violation and will not fail the build.
```
romani, line 850: I have some worries that we need to phrase it as "Checksyle's Audit Listeners (reports)" as decision on what to show and how if up to module that does report generation.

src/xdocs/property_types.xml
```diff
@@ -831,16 +831,62 @@
...
+              module may still run and if an exception occurs, the exception will turn into an
+              error and act as one.
+            </td>
+          </tr>
+
+          <tr>
+            <td><code>info</code></td>
+            <td>
+              If a violation occurs, modules with this severity are displayed as informational.
+              Checkstyle will display the violation, but will not fail the build.
```
romani, line 862: > build

should be "execution". We can be executed as CLI , so no "build".

src/xdocs/property_types.xml
```diff
@@ -831,16 +831,62 @@
...
+              This severity can be used for normal logging.
+            </td>
+          </tr>
+
+          <tr>
+            <td><code>warning</code></td>
+            <td>
+              This severity behaves exactly the same as <code>info</code> in Checkstyle.
+              It can be used to stress something that <b>MAY</b> cause future problems and may
```
romani, line 871: `used to let user bring user more attention to such violations ..`

we do not have opinion on what is problem, user define it by sevverity.

src/xdocs/property_types.xml
```diff
@@ -831,16 +831,62 @@
...
+              error and act as one.
+            </td>
+          </tr>
+
+          <tr>
+            <td><code>info</code></td>
+            <td>
+              If a violation occurs, modules with this severity are displayed as informational.
+              Checkstyle will display the violation, but will not fail the build.
+              This severity can be used for normal logging.
```
romani, line 863: > normal logging

hmmm, I do not know how to say it better. May be remove line is better.

src/xdocs/property_types.xml
```diff
@@ -831,16 +831,62 @@
...
+              <a href="anttask.html#Parameters"><code>maxWarnings</code></a> can be used to pass
+              the build until a certain number of warnings are found.
+            </td>
+          </tr>
+
+          <tr>
+            <td><code>error</code></td>
+            <td>
+              If a violation occurs, modules with this severity are displayed as error and should
+              be considered as a failure.
```
romani, line 883: modules are not shown, we show violations. you might mean `violation with`.

>be considered as a failure.

Something tel me that we should avoid phases on "failure", as it is decision of plugin/cli/.... to make decision. Plugin can do failure based on "gates" , for example info > 200.

src/xdocs/property_types.xml
```diff
@@ -831,16 +831,62 @@
...
+              Depending on how Checkstyle is executed, there may be some additional settings which
+              can be set to not fail the build or until a certain number of errors occur.
```
romani, line 885: do we have such settings in ant ?

again, "fail" ... "build".

romani (changes requested):
items:

nrmancuso:
Github, generate site

src/xdocs/property_types.xml
```diff
@@ -831,16 +831,62 @@
...
+          </tr>
+
+          <tr>
+            <td><code>ignore</code></td>
+            <td>
+              If a violation occurs, modules with this severity are ignored as if they never
+              occurred.
+              Checkstyle will not display the violation and will not fail the build.
+              This severity can be used to disable a module in the configuration. However,
+              depending on the setting of <code>executeIgnoredModules</code> when executed, the
```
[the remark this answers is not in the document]
rnveach: Done.

Link is https://checkstyle.org/cmdline.html#Command_line_usage

src/xdocs/property_types.xml
rnveach: Done for all 4.

src/xdocs/property_types.xml
rnveach: Done.

New sentence is `It can be used to let user bring more attention to reviewers of such violations.` for grammar.

src/xdocs/property_types.xml
rnveach: Removed.

src/xdocs/property_types.xml
rnveach: > modules are not shown, we show violations. you might mean violation with.

Modules are the ones where severities are set in the config. I think I was trying to reference that.

How about we rewrite these sentences to:
* If a violation with this severity occurs, it is ...

> we should avoid phases on "failure"

I am avoiding using the same words as the name of the severity. Otherwise we are saying "Error is an error" and it means nothing for describing to the user if they were already confused about "error" in the first place.

I don't have anything else as I do think this is a violation that needs corrective action. Do we want to use that wording instead?

src/xdocs/property_types.xml
rnveach: https://checkstyle.org/anttask.html#Parameters
> maxErrors 	The maximum number of errors that are tolerated before breaking the build or setting the failure property.
maxWarnings 	The maximum number of warnings that are tolerated before breaking the build or setting the failure property.

I changed build to execution.

romani:
Github, generate site

src/xdocs/property_types.xml
```diff
@@ -831,16 +831,62 @@
...
+            <td>Definition</td>
+          </tr>
+
+          <tr>
+            <td><code>ignore</code></td>
+            <td>
+              If a violation occurs, modules with this severity are ignored as if they never
+              occurred.
+              Checkstyle Audit Listeners will not display the violation and will not fail the
+              execution.
```
romani, line 851: I think we can remove this sentence, as we clearly state that we remove module from execution or move below mention of option

src/xdocs/property_types.xml
```diff
@@ -831,16 +831,62 @@
...
+            </td>
+          </tr>
+
+          <tr>
+            <td><code>error</code></td>
+            <td>
+              If a violation occurs, modules with this severity are displayed as error and should
+              be considered as a failure.
+              Depending on how Checkstyle is executed, there may be some additional settings which
+              can be set to not fail the execution or until a certain number of errors occur.
```
romani, line 885: I think we should mention that CLI and Ant will fail. We do not know about any others behavior of plugins

romani (changes requested):
Items

src/xdocs/property_types.xml
romani: >be considered as a failure.

Hmm, I still think it not good wording. I will think on it tomorrow

src/xdocs/property_types.xml
rnveach: I moved it below the option and shortened it.

src/xdocs/property_types.xml
rnveach: > I think we should mention that CLI and Ant will fail. 

Done.

> We do not know about any others behavior of plugins

Agreed, but we should document what we believe the expected behavior is. Plugins can always change it for their needs, but our view of the severity will not change. Ex: We feel errors should fail the execution even if another plugin says at least X violations must occur first.

romani:
Github, generate site

src/xdocs/property_types.xml
romani: audit listeners does not make decision on failure/(error-code) of execution
https://github.com/checkstyle/checkstyle/blob/a8f065738de6c4ff917a2ee57521f6cf8718f2ba/src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java

decusion on when to fail in 
wrapper of Checkstyle: CLI, ANT, any plugin ....

example:
https://github.com/checkstyle/checkstyle/blob/b1c44756f64cb1009e9648cf107e7e467c27197c/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L130 

lets be explicit. to make it clear that each wrapper works differently and better to refer to its documentation to know when execution fail.

romani: may be we should make two more columns: CLI and ANT  and describe what such wrapers are doing with events severity.

CLI does not fail on WARN, and has no arguments to contomize failure behavior.
ANT is more advanced and has limits on error and warn severity.

romani:
@rnveach , I pushed extra change , please review , if you agree squash them.
image

src/xdocs/property_types.xml
rnveach: This is why I didn't want to get too depth into specifics with my original wording. I was giving an overview and then saying actual behavior depends on execution method. We could link to those sections so users can see more details over there. I worry if we put all details here, we will forget to update it later when we add new functionality.

rnveach:
Github, generate site

rnveach:
I am ok.

romani (approved):
Ok to merge

nrmancuso (approved):

PR #12365, opened 2022-11-04
infra: added quarterly cache clean up (travis)

rnveach:
After thinking about random things late at night, and then seeing sevntu has a 2 gig cache file. Checkstyle seems to be around 600 megs.

https://app.travis-ci.com/github/sevntu-checkstyle/sevntu.checkstyle/caches
https://app.travis-ci.com/github/checkstyle/checkstyle/caches

Command from https://www.howtogeek.com/howto/ubuntu/delete-files-older-than-x-days-on-linux .

This won't be perfect as we could removing dependencies that are still used but have not updated in this timeframe.

### DIFF
--- a/.ci/travis.sh
+++ b/.ci/travis.sh
@@ -138,6 +138,10 @@ jacoco)
   fi
   ;;
 
+quarterly-cache-cleanup)
+  find ~/.m2 -maxdepth 4 -type d -mtime +90 -exec rm -rf {} \;
+  ;;
+
 *)
   echo "Unexpected argument: $1"
   sleep 5s
--- a/.ci/jsoref-spellchecker/whitelist.words
+++ b/.ci/jsoref-spellchecker/whitelist.words
@@ -830,6 +830,7 @@ markdownlint
 Massol
 matchxpath
 mavenbadge
+maxdepth
 maxmem
 maxmethods
 mct
@@ -878,6 +879,7 @@ mozilla
 MRELEASE
 mstudman
 MTAGLIST
+mtime
 MUI
 multicharacter
 multiline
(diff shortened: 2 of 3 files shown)


rnveach:
Run reduced the cache to 198 megs (reduced to 30% of the original).

romani:
Interesting idea.
We definitely keep there all our old dependencies and all their depencies.

But let's make it 90 days. During 30 days we update not all depencies, but bunch of them, we can presume that during x3 time period we most likely refresh all dependencies.
Or we can take some stat of dependency commits and make it different time period.

rnveach:
> 's make it 90 days

Done.

romani (approved):
Ok to merge

rnveach:
New push has increased the cache size by 100 megs so far.

We delete old files after the run is complete, so this means we have ~100 megs of 30 day old dependencies still in use.

romani:
We increase size, but think it is not old dependencies, but just depencies that doesn't change much.

.ci/travis.sh
```diff
@@ -138,6 +138,10 @@ jacoco)
   fi
   ;;
 
+quarterly-cache-cleanup)
+  find ~/.m2 -mtime +90 -exec rm {} \;
+  ;;
```
nrmancuso, line 143: It would be awesome if we could base cache cleaning on changed SHA of pom.xml

romani: We changing versions too often in pom.xml

rnveach: Also pom can change for other things besides version upgrades.

I will come back to this PR to unblock it.

romani: I do not recommend to block this PR, we can always improve later on.

nrmancuso (approved):
I would rather destroy cache completely at least once a month, but this is better than allowing accumulation of all old dependencies.

romani (approved):

PR #12367, opened 2022-11-04
infra: reorg circleci to have one parameterized job for validation

nrmancuso:
Reorganizes circleci config, functionality is the same. Also, I added a cache in a second commit to help CI to be more stable and speed up builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,216 +1,24 @@
-version: 2
-jobs:
+version: 2.1
 
-  no-exception-lucene-and-others-javadoc:
-    docker:
-      - image: &default-image checkstyle/jdk-11-groovy-git-mvn:11.0.13__3.0.9__2.25.1__3.6.3
+commands:
+  run-maven-script:
+    parameters:
+      script-name:
+        description: "name of script"
+        type: string
+        default: ""
+      args:
+        description: "arguments to script"
+        type: string
+        default: ""
     steps:
       - checkout
       - run:
-          command: |
-            ./.ci/no-exception-test.sh no-exception-lucene-and-others-javadoc
+          command: << parameters.script-name >> << parameters.args >>
       - run:
           command: ./.ci/validation.sh git-diff
-  no-exception-cassandra-storm-tapestry-javadoc:
-    docker:
-      - image: *default-image
-    steps:
-      - checkout
-      - run:
-          command: |
-            ./.ci/no-exception-test.sh no-exception-cassandra-storm-tapestry-javadoc
-      - run:
-          command: ./.ci/validation.sh git-diff
-  no-exception-hadoop-apache-groovy-scouter-javadoc:
-    docker:
-      - image: *default-image
-    steps:
-      - checkout
-      - run:
-          command: |
-            ./.ci/no-exception-test.sh no-exception-hadoop-apache-groovy-scouter-javadoc
-      - run:
-          command: ./.ci/validation.sh git-diff
-  no-exception-only-javadoc:
-    docker:
-      - image: *default-image
-    steps:
-      - checkout
-      - run:
-          command: |
-            ./.ci/no-exception-test.sh no-exception-only-javadoc
-      - run:
-          command: ./.ci/validation.sh git-diff
-  no-error-xwiki:
-    docker:
-      - image: *default-image
-    steps:
-      - checkout
-      - run:
-          command: |
-            ./.ci/validation.sh no-error-xwiki
-      - run:
-          command: ./.ci/validation.sh git-diff
-  no-error-pmd:
-    docker:
-      - image: *default-image
-    steps:
-      - checkout
-      - run:
-          command: |
-            ./.ci/validation.sh no-error-pmd
-      - run:
-          command: ./.ci/validation.sh git-diff
-  no-error-hibernate-search:
-    docker:
-      - image: cimg/openjdk:18.0.1
-    steps:
-      - checkout
-      - run:
-          command: |
-            ./.ci/validation.sh no-error-hibernate-search
-      - run:
-          command: ./.ci/validation.sh git-diff
-  no-exception-struts:
-    docker:
-      - image: *default-image
-    steps:
-      - checkout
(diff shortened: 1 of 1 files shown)


rnveach:
@nrmancuso Not saying it needs to be done in this PR, but can we add caching to circleci? I have seen a few runs fail because of failure to download dependencies, and it seems like we are downloading them all the time.

nrmancuso:
> @nrmancuso Not saying it needs to be done in this PR, but can we add caching to circleci? I have seen a few runs fail because of failure to download dependencies, and it seems like we are downloading them all the time.

It is already done :)

.circleci/config.yml
```diff
@@ -261,31 +78,74 @@ jobs:
...
+                "no-exception-hibernate-orm",
+                "no-exception-spotbugs",
+                "no-exception-spoon",
+                "no-exception-spring-framework",
+                "no-exception-hbase",
+                "no-exception-Pmd-elasticsearch-lombok-ast",
+                "no-exception-alot-of-projects",
+                "no-warning-imports-guava",
+                "no-warning-imports-java-design-patterns",
+                "sonarqube",
```
nrmancuso, line 147: We should probably consider moving these tasks to more aptly named files. I propose the following:
1. All `no-exception` should live in `no-exception-test.sh`, and only these tests
2. All `no-error` tests should live in a script called `no-error-test.sh`
3. All `no-warning` test should live in a script called `no-warning-test.sh`
4. Then we have `misc.sh for stuff like sonar.

Thoughts?

romani: 1 and 2 ok.

3 and 4 are rare, better to keep in validation.sh

nrmancuso: Then perhaps 3 and 4 can be combined into misc?

romani: I don't like name - misc. What is a problem with 'validation' ?

rnveach: Agree with 1 and 2.

I don't have any issues with 3, but I don't know how many no-warnings we have. Also, how is that different than no-error? Warnings don't fail the build normally.

I agree with @romani on 4, better in validation.sh for now. validation is the misc of files.

.circleci/config.yml
```diff
@@ -261,31 +78,74 @@ jobs:
             export PR_NUMBER=$CIRCLE_PR_NUMBER
             ./.ci/validation.sh git-check-pull-number
 
+  validation:
+    docker:
+      - image: &default-image checkstyle/jdk-11-groovy-git-mvn:11.0.13__3.0.9__2.25.1__3.6.3
```
rnveach, line 83: Just pointing this out, that this says groovy is 3.0.9 (unless I am reading this wrong), but we have specific version needed for other groovy script, 2.4.8 ( see https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#setup ) . Maybe we should consider synching so they are the same especially if we ever want to groovy in these CIs.

.circleci/config.yml
```diff
@@ -1,216 +1,33 @@
...
+          name: Restore Maven repo cache
+          keys:
+            - mvn-cache-{{ checksum "pom.xml" }}
+      - run:
+          command: << parameters.script-name >> << parameters.args >>
+      - run:
+          command: ./.ci/validation.sh git-diff
+      - save_cache:
+          name: Save Maven repo cache
+          key: mvn-cache-{{ checksum "pom.xml" }}
```
rnveach, line 26: Pointing to other conversation at https://github.com/checkstyle/checkstyle/pull/12365#discussion_r1014564736 and that whole PR ,

POM changes too often and we agreed to quarterly cache period. This should be updated to `(month - 1) / 3`.

rnveach (changes requested):

.circleci/config.yml
nrmancuso: Quarterly is much too long, Circleci does not even allow a cache to live that long anyway. Max lifetime of a CircleCI cache is two weeks.

rnveach: Ok, then why don't we just make this a static field? We don't change that much in a 2 weeks period that we would grow our cache out of control.

nrmancuso: This allows us to identify a given cache in case of some issue (bad download, experimental dependency cached, dependabot PR, etc).

romani: Non of cache model of very good, it all depends ..., usage of any is good enough. 
Circle CI recommend this caching for maven, ok.

rnveach:
@nrmancuso I just noticed this is still in draft. Is that intentional?

.circleci/config.yml
nrmancuso: Yes, long term I would like to avoid custom images altogether (other than IDEA) to make it easier to have consistent versions. Java version is not the same, either. But - by the same token, it is also not a bad idea to keep similar jobs running in different versions.

nrmancuso:
> @nrmancuso I just noticed this is still in draft. Is that intentional?

Thanks for the reminder :)

.circleci/config.yml
rnveach: Java makes sense.

Groovy does not as we code for specific version and if you deviate from that version by 1 bit, the code fails to compile. This is what prompted the documentation at contribution for tester. Groovy version should be consistent everywhere if we are ever to use it in more CIs.

rnveach (approved):

.circleci/config.yml
```diff
@@ -1,216 +1,33 @@
...
-      - image: *default-image
-    steps:
-      - checkout
-      - run:
-          command: |
-            ./.ci/validation.sh sonarqube
-      - run:
-          command: ./.ci/validation.sh git-diff
+commands:
+  run-script:
```
romani, line 4: Run_maven_script or something with maven in name 
as it deals with maven cache

romani (changes requested):
Items

.circleci/config.yml
nrmancuso: done

Issue #12221, opened 2022-09-21
checker-framework.groovy (differences): false positives when running checker-framework-nullness-optional-interning


rnveach:
When running `groovy ./.ci/checker-framework.groovy checker-framework-nullness-optional-interning` on a clean master, I get:
````
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  12:34 min
[INFO] Finished at: 2022-09-20T20:47:21-04:00
[INFO] ------------------------------------------------------------------------
New surviving error(s) found:

File Name: "checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\Checker.java"
Specifier: "argument"
Message: "incompatible argument for parameter moduleId of Violation."
Line Contents: "new String[] {ioe.getMessage()}, null, getClass(), null));"
Line Number: 342
Details: found   : null (NullType)
....
````

I get 308 new errors and 643 unnecessary errors.

I a running on Windows.

**Comment by nrmancuso** [comment]
@rnveach please share your groovy version

**Comment by nrmancuso** [comment]
Similar issue here, on linux:
```
➜  checkstyle git:(master) java --version
openjdk 18.0.2 2022-07-19
OpenJDK Runtime Environment (build 18.0.2+0)
OpenJDK 64-Bit Server VM (build 18.0.2+0, mixed mode)
➜  checkstyle git:(master) groovy --version
Groovy Version: 2.5.18 JVM: 18.0.2 Vendor: Oracle Corporation OS: Linux

```

Seeing as how checker relies on javac, it is possible that compiler warnings/errors can be different with each java version/vendor.

Vyom-Yadav:
I am on it and will share the reason by tomorrow.

Vyom-Yadav:
New errors coming up could be due to version/vendor of javac, I sent a fix for proper equals and hashCode contract in #12223, please check after applying that patch.

Vyom-Yadav:
> > I sent a fix for proper equals and hashCode contract in #12223, please check after applying that patch.
> 
> No real difference on my end.

Please try with the version and distribution used in CI, it looks to be version dependent.

nrmancuso:
> > > I sent a fix for proper equals and hashCode contract in #12223, please check after applying that patch.
> > 
> > 
> > No real difference on my end.
> 
> Please try with the version and distribution used in CI, it looks to be version dependent.

[...]

PR #12368, opened 2022-11-05
infra: run one checker to maintain windows compatibility

nrmancuso:
In order to prove resolution of #12221 and https://github.com/checkstyle/checkstyle/issues/12243, we should keep one checker execution in windows CI. I have added `|| VER>NUL` to the end of the checker command to allow checker execution to fail and CI still pass, until we resolve the above issues. `|| VER>NUL` should be removed in the scope of the above issues.

### DIFF
--- a/.ci/checker-framework.groovy
+++ b/.ci/checker-framework.groovy
@@ -112,7 +112,7 @@ private static int checkCheckerFrameworkReport(final String profile) {
  */
 private static List<List<String>> getCheckerFrameworkErrors(final String profile) {
     final List<String> checkerFrameworkLines = new ArrayList<>()
-    final String command = "mvn -e --no-transfer-progress clean compile -P${profile}"
+    final String command = "mvn -X -e --no-transfer-progress clean compile -P${profile}"
     final Process process = getOsSpecificCmd(command).execute()
     process.in.eachLine { final line ->
         checkerFrameworkLines.add(line)
--- a/.ci/validate-checker.cmd
+++ b/.ci/validate-checker.cmd
@@ -1,0 +1,2 @@
+@echo on
+groovy ./.ci/checker-framework.groovy checker-signature-gui-units-init
--- a/.ci/validate-checker.cmd
+++ b/.ci/validate-checker.cmd
@@ -1,2 +1,3 @@
 @echo on
+JAVA_HOME="C:\Program Files\OpenJDK\jdk-17"
 mvn -X -e --no-transfer-progress clean compile -Pchecker-signature-gui-units-init
--- a/.ci/validate-checker.cmd
+++ b/.ci/validate-checker.cmd
@@ -1,2 +1,2 @@
 @echo on
-groovy ./.ci/checker-framework.groovy checker-signature-gui-units-init
+mvn -X -e --no-transfer-progress clean compile -Pchecker-signature-gui-units-init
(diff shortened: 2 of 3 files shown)


romani (approved):
Ok to merge

nrmancuso:
Link to passing checker build in windows: https://cirrus-ci.com/task/4578533786779648

Checker is failing, but execution of checker works

rnveach:
> execution of checker works

@nrmancuso My concern is checker run is not producing anything. There are no new errors, and everything is being removed. It is almost like checker is not running at all and we are saying to remove all violations.

How can we say checker is working in this situation?

rnveach (changes requested):
Just saying, I think my above comment is a blocker until it is decided otherwise.

romani:
I thought Nick is just doing initial CI work to let us fix windows and quickly remove hack in this command to enable it.

We kind of merging infra change that should be easier activated when we ready for win support

rnveach:
Unless @nrmancuso has something else, it will continue to not function even if groovy supports Windows fully. There is no output from Checker at all. The execution is not being skipped, it is just not happening.

nrmancuso:
> Unless @nrmancuso has something else, it will continue to not function even if groovy supports Windows fully. There is no output from Checker at all. The execution is not being skipped, it is just not happening.

From the output, we can see that groovy script runs. This was really my only intention; from the execution time, it looks like checker based compilation is not working, and of course, we still have the maven bug that we are trying to work around for windows. We cannot claim "windows support" until this job runs exactly as it does in Linux CI. I am not sure how `echo` works in windows, but I have turned it on in the script now, hopefully we can get some more detailed output to figure out why checker compilation is not happening.

rnveach:
@nrmancuso If it is unclear what the issue is, lets report this to Checker. I am ok to merge this it if we create a new issue that it is not working in cirrus and to follow up with Checker.

nrmancuso:
> @nrmancuso If it is unclear what the issue is, lets report this to Checker. I am ok to merge this it if we create a new issue that it is not working in cirrus and to follow up with Checker.

I would like to debug a bit more before we do this.

rnveach:
> I am not sure how echo works in windows, but I have turned it on in the script now

Echo just prints the commands being used. It doesn't do anything different in terms of maven execution. I believe it is on by default.
https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/echo

nrmancuso:
Ok, run of checker without script (just maven) and debug on results in:
```
[WARNING] Unable to autodetect 'javac' path, using 'javac' from the environment.
```

I wonder if we use other javac that we might not pick up all annotation processors or something. I also suspect that JAVA_HOME variable is not set by choco install, this could be a problem as well. @rnveach thoughts?

nrmancuso:
Trying to set the `JAVA_HOME` environment variable, I have tried the following and none have worked:
```
JAVA_HOME="C:\Program Files\OpenJDK\jdk-17"
set JAVA_HOME="C:\Program Files\OpenJDK\jdk-17"
set JAVA_HOME "C:\Program Files\OpenJDK\jdk-17"
set JAVA_HOME = "C:\Program Files\OpenJDK/jdk-17"
```
Also in the cirrus environment variables: `JAVA_HOME: "C:\Program Files\OpenJDK\jdk-17"`

nrmancuso:
I am closing this PR. Until our Checker script works reliably in Windows, I do not see value in continuing work here. Anyone is welcome to take over/ reuse code.

PR #12380, opened 2022-11-07
infra: re-enable nondex and upgrade to 2.1

rnveach:
https://github.com/TestingResearchIllinois/NonDex/issues/175

https://github.com/TestingResearchIllinois/NonDex/compare/nondex-1.1.2...master (note, no tag for 2.1)

nondex 2.1 was released and should support Java 11.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -27,22 +27,6 @@ blocks:
               cache store m2 $HOME/.m2
             fi
       jobs:
-        # NonDex supports only Java 8 till https://github.com/TestingResearchIllinois/NonDex/pull/155
-        # - name: NonDex (openjdk11)
-        #   priority:
-        #     - value: 90
-        #       when: true
-        #   matrix:
-        #     - env_var: CMD
-        #       values:
-        #         - .ci/validation.sh nondex
-        #   commands:
-        #     - sem-version java 11
-        #     - echo "eval of CMD is starting";
-        #     - echo "CMD=$CMD";
-        #     - eval $CMD;
-        #     - echo "eval of CMD is completed";
-
         - name: Validation (openjdk11, fast pool)
           matrix:
             - env_var: CMD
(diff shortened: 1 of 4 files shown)


rnveach:
CI results coming....

https://github.com/checkstyle/checkstyle/actions/runs/3413639313/jobs/5680589386
Pitest seems to have an unstable mutation

rnveach:
nondex doesn't appear to work on windows.

rnveach:
@romani Command is failing at https://github.com/checkstyle/checkstyle/blob/7215503167e1254d0b06c8396256959cdef62418/.ci/validation.sh#L91

Running `grep -RlE 'td class=.x' .nondex/ | cat` on my linux machine after nondex runs, produces an empty result.

Should there be some type of HTML file generated? I am unfamilar with this and the commands being run.

rnveach:
It seems the above command is putting `cat: '': No such file or directory` into the output file, when we are expecting it to be empty for no failures.

romani:
New version might use new HTML we just need to have any grep to know if there is some violations

rnveach:
@romani There has been no changes to the report logic itself from the bottom version we are upgrading from, 1.1.2. The last time the HTML report logic was modified was 7 years ago, which was version 1.0.1. It doesn't look like the report itself has changed in this upgrade.

romani:
We need to reinvestigate ... I have no idea what is going on.

rnveach:
https://checkstyle.semaphoreci.com/jobs/252a2b91-026f-4a8c-b8a8-fc974781f352#L4764
Fix seems to be working, but someone with more bash experience will have to verify.

.semaphore/semaphore.yml
```diff
@@ -27,21 +27,20 @@ blocks:
...
-        #     - eval $CMD;
-        #     - echo "eval of CMD is completed";
+        - name: NonDex (openjdk11)
+          priority:
+            - value: 90
+              when: true
+          matrix:
+            - env_var: CMD
+              values:
+                - .ci/validation.sh nondex
```
romani, line 37: Can this execution join `Validation (openjdk11, fast pool)` job matrix ?

romani (changes requested):
Items

.semaphore/semaphore.yml
rnveach: Done.

I don't know but I assume it can.

romani (approved):
Ok to merge

nrmancuso (approved):

nrmancuso:
Drone build pending is expected, and failure at https://app.travis-ci.com/github/checkstyle/checkstyle/jobs/588305771#L1718 is not related (cache clearing)

Issue #12370, opened 2022-11-05
Improper text in NewLineAtEndOfFileCheck documentation


NitishKumar06:
The following link (https://checkstyle.org/config_misc.html#NewlineAtEndOfFile) explains the check for adding a new line at the end of file. There's a line mentioned which says _Example (line 36 should not be in diff):_.

Although when you look at the image below, you find that there is no line number  present / way to find out the line number 36.

The IDE is capable of showing the line 36 but the line numbers are not visible in the website.

Would it be a good idea to modify the documentation little bit in order to add line number/ remove the text?

strkkk:
I think it can be simpler, just 2 lines and refer to them as line 1 and line 2.

romani:
and use output of `diff -u`

romani:
Probably better to send PR and discuss exact wording in PR.

NitishKumar06:
I believe it'd be a good idea if we remove "line 36 should not be in diff" since different files have different line numbers which may leads to confusion.

PR #12384, opened 2022-11-09
Issue #12370: Removed the unwanted text for NewLineCheck

NitishKumar06:
Fix: #12370

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/NewlineAtEndOfFileCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/NewlineAtEndOfFileCheck.java
@@ -38,7 +38,7 @@ import com.puppycrawl.tools.checkstyle.api.FileText;
  * command does not show previous lines as changed.
  * </p>
  * <p>
- * Example (line 36 should not be in diff):
+ * Example (the line with 'No newline at end of file' should not be in the diff):
  * </p>
  * <pre>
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/NewlineAtEndOfFileCheck.java
```diff
@@ -38,7 +38,7 @@
  * command does not show previous lines as changed.
  * </p>
  * <p>
- * Example (line 36 should not be in diff):
+ * Example:
```
rnveach, line 41: To me, I think it would be better to say `Example: (the line with 'No newline at end of file' should not be in the diff)`.

NitishKumar06: Exactly! That's way better approach then

rnveach (approved):

romani (approved):
Thanks a lot

NitishKumar06:
@romani @rnveach I just found out that the content on the website don't show these changes because the HTML file was not modified. I'd be raising a PR to fix that as well.

NitishKumar06:
@rnveach Can you see the change in the content here: https://checkstyle.org/config_misc.html#NewlineAtEndOfFile?
I couldn't :(

rnveach:
Our website isn't updated until after a release. It is not real time.

PR #12388, opened 2022-11-11
Issue #12181: Fail 'Publish GitHub page' when curl fails

### LINKED ISSUE #12181: 'R: Publish GitHub Page' failing at perform curl command but action status is succcess
https://github.com/checkstyle/checkstyle/actions/runs/3007811873

such execution should be failed.

stoyanK7:
Resolves #12181 

---
From curl docs:
> [--fail-with-body](https://curl.se/docs/manpage.html#--fail-with-body)
(HTTP) Return an error on server errors where the HTTP response code is 400 or greater). In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This flag will still allow curl to output and save that content but also to return error 22. 

---
Failed run: https://github.com/stoyanK7/checkstyle/actions/runs/3444671922/jobs/5747521215

```
curl: (22) The requested URL returned error: 404
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/reference/repos#update-a-release"
}
Error: Process completed with exit code 22.
```

### DIFF
--- a/.ci/update-github-page.sh
+++ b/.ci/update-github-page.sh
@@ -89,7 +89,7 @@ EOF
 echo "$JSON"
 
 echo "Updating Github tag page"
-curl \
+curl --fail-with-body \
   -X PATCH https://api.github.com/repos/checkstyle/checkstyle/releases/"$RELEASE_ID" \
   -H "Accept: application/vnd.github+json" \
   -H "Authorization: token $BUILDER_GITHUB_TOKEN" \


.ci/update-github-page.sh
```diff
@@ -93,4 +93,5 @@ curl \
   -X PATCH https://api.github.com/repos/checkstyle/checkstyle/releases/"$RELEASE_ID" \
   -H "Accept: application/vnd.github+json" \
   -H "Authorization: token $BUILDER_GITHUB_TOKEN" \
-  -d "${JSON}"
+  -d "${JSON}" \
+  --fail-with-body
```
nrmancuso, line 97: Please move to line with `curl`, typically data argument is last, preceded by headers, preceded by url and request method.

nrmancuso (changes requested):
Item:

.ci/update-github-page.sh
stoyanK7: Done.

nrmancuso (approved):

.ci/update-github-page.sh
```diff
@@ -89,7 +89,7 @@ EOF
 echo "$JSON"
 
 echo "Updating Github tag page"
-curl \
+curl --fail-with-body \
```
rnveach, line 92: Shouldn't this be a requirement on all `curls` like we do with `e` for mvn? It seems to me we should want to fail on all curl failures, just that this was one example.
Example: https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle_non_main_files_checks.xml#L26-L32

rnveach (changes requested):

.ci/update-github-page.sh
nrmancuso: @rnveach I agree, @stoyanK7 please create an issue for creation of new Regexp Singleline module. There might be some exceptions in case of retries (I think we have some scripts that try repeatedly to download something), we can deal with that in PR for new issue.

romani: Created
https://github.com/checkstyle/checkstyle/issues/12389

Let's unblock this PR

romani (approved):
Ok to merge.

Better to not resolve review item above to let be easy seen, as it referenced in new issue

rnveach (approved):

PR #12386, opened 2022-11-10
Issue #11446: Replace verify method in ImportControlCheckTest

arinmodi:
Subtask of #11446  : Update The ImportControlCheckTest

PR #12395, opened 2022-11-12
Update Test Configuration To Support Adding TreeWalker Property

arinmodi:
In test Input Files, Checks Modules and Properties are passed in comments at the top file and all these modules are added as child of TreeWalker.

But There is no way to configure the property of TreeWalker from Input Files Comments. So I added the function which will set the property of treeWalker.

For ref, #12386

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputConfiguration.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputConfiguration.java
@@ -84,9 +84,8 @@ public final class TestInputConfiguration {
 
     public DefaultConfiguration createConfiguration() {
         final DefaultConfiguration root = new DefaultConfiguration(ROOT_MODULE_NAME);
-        final DefaultConfiguration treeWalker =
-                new DefaultConfiguration(TreeWalker.class.getName());
         root.addProperty("charset", StandardCharsets.UTF_8.name());
+        final DefaultConfiguration treeWalker = createTreeWalker();
         childrenModules
                 .stream()
                 .map(ModuleInputConfiguration::createConfiguration)
@@ -94,7 +93,7 @@ public final class TestInputConfiguration {
                     if (CHECKER_CHILDREN.contains(moduleConfig.getName())) {
                         root.addChild(moduleConfig);
                     }
-                    else {
+                    else if (!treeWalker.getName().equals(moduleConfig.getName())) {
                         treeWalker.addChild(moduleConfig);
                     }
                 });
@@ -104,9 +103,8 @@ public final class TestInputConfiguration {
 
     public DefaultConfiguration createConfigurationWithoutFilters() {
         final DefaultConfiguration root = new DefaultConfiguration(ROOT_MODULE_NAME);
-        final DefaultConfiguration treeWalker =
-                new DefaultConfiguration(TreeWalker.class.getName());
         root.addProperty("charset", StandardCharsets.UTF_8.name());
+        final DefaultConfiguration treeWalker = createTreeWalker();
         childrenModules
                 .stream()
                 .map(ModuleInputConfiguration::createConfiguration)
@@ -115,12 +113,23 @@ public final class TestInputConfiguration {
                     if (CHECKER_CHILDREN.contains(moduleConfig.getName())) {
                         root.addChild(moduleConfig);
                     }
-                    else {
+                    else if (!treeWalker.getName().equals(moduleConfig.getName())) {
                         treeWalker.addChild(moduleConfig);
                     }
                 });
         root.addChild(treeWalker);
         return root;
+    }
+
+    private DefaultConfiguration createTreeWalker() {
+        final DefaultConfiguration treeWalker;
+        if (childrenModules.get(0).getModuleName().equals(TreeWalker.class.getName())) {
+            treeWalker = childrenModules.get(0).createConfiguration();
+        }
+        else {
+            treeWalker = new DefaultConfiguration(TreeWalker.class.getName());
+        }
+        return treeWalker;
     }
 
     public static final class Builder {
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportControlCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportControlCheckTest.java
@@ -423,17 +423,12 @@ public class ImportControlCheckTest extends AbstractModuleTestSupport {
 
     @Test
     public void testFileNameNoExtension() throws Exception {
-        final DefaultConfiguration checkConfig = createModuleConfig(ImportControlCheck.class);
-        checkConfig.addProperty("file",
-                getResourcePath("InputImportControlFileNameNoExtension.xml"));
-        final DefaultConfiguration treewalkerConfig = createModuleConfig(TreeWalker.class);
-        treewalkerConfig.addProperty("fileExtensions", "");
-        treewalkerConfig.addChild(checkConfig);
         final String[] expected = {
-            "11:1: " + getCheckMessage(MSG_DISALLOWED, "java.awt.Image"),
+            "13:1: " + getCheckMessage(MSG_DISALLOWED, "java.awt.Image"),
         };
 
-        verify(treewalkerConfig, getPath("InputImportControlFileNameNoExtension"), expected);
+        verifyWithInlineConfigParser(
+                getPath("InputImportControlFileNameNoExtension"), expected);
     }
 
     /**
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/importcontrol/InputImportControlFileNameNoExtension
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/importcontrol/InputImportControlFileNameNoExtension
@@ -1,14 +1,16 @@
 /*
-ImportControl
-file = InputImportControlFileNameNoExtension.xml
-path = (default).*
+com.puppycrawl.tools.checkstyle.TreeWalker
+fileExtensions =
 
+ImportControl
+file = (file)InputImportControlFileNameNoExtension.xml
+path = (default).*
 
 */
 
 package com.puppycrawl.tools.checkstyle.checks.imports.importcontrol;
 
-import java.awt.Image; // violation
+import java.awt.Image; // violation 'Disallowed import - java.awt.Image'
 
 public @interface InputImportControlFileNameNoExtension {
 }


rnveach:
This is not a config update. Please use `Pull #12395: Update Test Configuration To Support Adding TreeWalker Property` as your commit message so release notes will connect information for this commit to this PR.

Also please update 1 existing test that will make use of this feature to prove it is now working.

nrmancuso:
@arinmodi place all details and context of why were are making this change in PR description so that we do not need to dig into other PR to find information later.

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -184,7 +187,13 @@ private static void setModules(TestInputConfiguration.Builder testInputConfigBui
                     new ModuleInputConfiguration.Builder();
             setModuleName(moduleInputConfigBuilder, inputFilePath, lines.get(lineNo));
             setProperties(moduleInputConfigBuilder, inputFilePath, lines, lineNo + 1);
-            testInputConfigBuilder.addChildModule(moduleInputConfigBuilder.build());
+            if (TREE_WALKER_MODULE_NAME.equals(lines.get(lineNo))) {
+                testInputConfigBuilder.setTreeWalkerModuleInputConfiguration(
```
romani, line 191: Please hide all logic of special handling of Treewalker at testInputConfigBuilder

Treewalker is just another module. As it appears we add it to configbuilder. And config builder internally figure out how to replace all properties new Treewalker to existing Treewalker created by default.

romani (changes requested):
We need removal of suppression
https://github.com/checkstyle/checkstyle/pull/12386/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L1804

Good step forward, items to improve

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
arinmodi: Do you mean I remove this condition of checking is the module TreeWalker or not ?

and add it as child module ? just like another modules

romani: You will need this condition in Builder class. Yes, almost like just another module.

arinmodi: Got it

arinmodi: Is it ok ?

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
```diff
@@ -231,8 +257,10 @@ protected final void verifyWithInlineConfigParser(String filePath, String... exp
             throws Exception {
         final TestInputConfiguration testInputConfiguration =
                 InlineConfigParser.parse(filePath);
+        final DefaultConfiguration treeWalker = processTreeWalkerModuleFromFile(
+                testInputConfiguration);
         final DefaultConfiguration parsedConfig =
-                testInputConfiguration.createConfiguration();
+                testInputConfiguration.createConfiguration(treeWalker);
```
romani, line 263: Can you avoid such change in this test method ?
Just load Treewalker as another module, if it already exists, put properties from  file to existing module

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
romani: You can put Treewalker before check in file to make more natural

arinmodi:
> We need removal of suppression https://github.com/checkstyle/checkstyle/pull/12386/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L1804
> 
> Good step forward, items to improve

If I remove that suppression than I have to replace all verify with verifyInLineConfigParser in that files, which doesn't belong to this PR scope.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
arinmodi: Done

src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputConfiguration.java
```diff
@@ -104,9 +110,13 @@ public DefaultConfiguration createConfiguration() {
...
         final DefaultConfiguration root = new DefaultConfiguration(ROOT_MODULE_NAME);
-        final DefaultConfiguration treeWalker =
-                new DefaultConfiguration(TreeWalker.class.getName());
         root.addProperty("charset", StandardCharsets.UTF_8.name());
+        DefaultConfiguration treeWalkerModule = new DefaultConfiguration(
+                TreeWalker.class.getName());
+        if (treeWalkerModuleInputConfiguration != null) {
+            treeWalkerModule = treeWalkerModuleInputConfiguration.createConfiguration();
+        }
+        final DefaultConfiguration treeWalker = treeWalkerModule;
```
romani, line 119: Too much variables about the same.
Use if-else assign directly to `DefaultConfiguration treeWalker`

Is it possible to avoid treeWalkerModuleInputConfiguration field ?
By treatment of Treewalker as just another child module.
And during construction of config search of it in children by name?

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputConfiguration.java
arinmodi: I think removing treeWalkerModuleInputConfiguration would not be good, cause than we have to traverse in childModules list each time and if found we have to remove that from childModules also. So new Field should be good, there is just a  one simple check.

In short with this field time Complexity will reduce to just O(1) from O(n), n is number of child in list.

romani: In tests we do not care much of performance especially if that makes code easier.

arinmodi: > In tests we do not care much of performance especially if that makes code easier. We value maintainability more than performance in this case.

Oh! That's why test are taking too much time to run

romani: No, because we parse files for each test. Parsing is most time consuming part.

arinmodi: @romani, mvn clean verify taking too much time, is there any other command that will just check the updated files ?

romani: We are not that advanced, sorry.
How much it takes on your local?

You need to run verify only at the end, before push. In all other cases just run single test from your IDE

arinmodi: it takes around 8-10 minutes,

image

The main problem is to wait 8-10 minutes,  just to check errors of checkstyle

arinmodi: BTW this is done

src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputConfiguration.java
```diff
@@ -82,11 +82,22 @@ public List<TestInputViolation> getFilteredViolations() {
         return Collections.unmodifiableList(filteredViolations);
     }
 
+    public DefaultConfiguration getTreeWalker() {
+        DefaultConfiguration treeWalker = new DefaultConfiguration(TreeWalker.class.getName());
+        for (ModuleInputConfiguration moduleInputConfiguration : childrenModules) {
+            if (treeWalker.getName().equals(moduleInputConfiguration.getModuleName())) {
+                treeWalker = moduleInputConfiguration.createConfiguration();
+                childrenModules.remove(moduleInputConfiguration);
```
romani, line 90: Too complicated.

Can you do logic at code around 
```
forEach(moduleConfig -> {
                    if (CHECKER_CHILDREN.contains(moduleConfig.getName())) {
                        root.addChild(
```

romani (changes requested):
Itet

src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputConfiguration.java
arinmodi: No problem in that but than I have to copy same code to two functions, will it be ok ?

romani: You need to skip Treewalker in foreach

arinmodi: What if TreeWalker specified after other check modules than it will be lost if i overwrite.

romani: Last Treewalker wins.
You can put Treewalker above Check in header to initialize in required order

arinmodi: So, is that compulsory to define the treeWalker at the top ?

romani: I am ok with this.
If not defined, we use default.

arinmodi: This is last. I got an idea,

While adding child i will check is that treeWalker or not, if it is I will add that at position 0, in this way it always stays at top in list even if defined last in input file, what do you think ?

romani: You can sort modules before foreach

arinmodi: how about one position variable, which will store the index of treeWalker in list ?

arinmodi: Done,

TreeWalker Must Specified At The Top otherwise it takes default treeWalker.

src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputConfiguration.java
```diff
@@ -104,9 +109,14 @@ public DefaultConfiguration createConfiguration() {
 
     public DefaultConfiguration createConfigurationWithoutFilters() {
         final DefaultConfiguration root = new DefaultConfiguration(ROOT_MODULE_NAME);
-        final DefaultConfiguration treeWalker =
-                new DefaultConfiguration(TreeWalker.class.getName());
         root.addProperty("charset", StandardCharsets.UTF_8.name());
+        final DefaultConfiguration treeWalker;
+        if (childrenModules.get(0).getModuleName().equals(TreeWalker.class.getName())) {
+            treeWalker = childrenModules.get(0).createConfiguration();
```
romani, line 115: Please make a method
initializeTreewaker

romani (changes requested):
Items`

src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputConfiguration.java
romani: It is not just checkstyle errors, there a lot of tool and validations.
CI do for you many hours of extra validations.

All together makes it a bit slow but no regression leaked to user - our paramount goal

src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputConfiguration.java
arinmodi: Done

romani (approved):
Good

src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputConfiguration.java
```diff
@@ -82,19 +82,29 @@ public List<TestInputViolation> getFilteredViolations() {
         return Collections.unmodifiableList(filteredViolations);
     }
 
+    public DefaultConfiguration initializeTreeWalker() {
```
rnveach, line 85: I am not seeing any external uses, so this should be private. Also please move this to be after it's usage.

rnveach, line 85: When I think `initialize`, I think we are populating a field or class somewhere. This is creating or pulling out an existing one and returning. It isn't doing populating anything.

I think it should be renamed to `createTreeWalker`.

rnveach (changes requested):
My only 2 items.

rnveach:
@arinmodi Please also make CI happy. One failure is PR is based on old master.

src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputConfiguration.java
arinmodi: Done

arinmodi: Done

arinmodi:
why git Diff command failing in the Checker / check (lock-tainting) (pull_request), is it from my end or CI ?

nrmancuso:
Checker is restarted, bad download

arinmodi:
Now is it good to merge ?

I want to complete the other pr dependent on this

rnveach (approved):

PR #12397, opened 2022-11-12
minor: prevent find running for quarterly cleanup if it doesn't exist

rnveach:
Prevent build failures on other PRs.

### DIFF
--- a/.ci/travis.sh
+++ b/.ci/travis.sh
@@ -10,6 +10,9 @@ case $1 in
 
 init-m2-repo)
   if [[ $RUN_JOB == 1 ]]; then
+    MVN_REPO=$(mvn -e --no-transfer-progress help:evaluate -Dexpression=settings.localRepository \
+      -q -DforceStdout);
+    echo "Maven Repo Located At: " "$MVN_REPO"
     MVN_SETTINGS=${TRAVIS_HOME}/.m2/settings.xml
     if [[ -f ${MVN_SETTINGS} ]]; then
       if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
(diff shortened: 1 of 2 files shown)


.ci/travis.sh
```diff
@@ -139,7 +142,13 @@ jacoco)
   ;;
 
 quarterly-cache-cleanup)
-  find ~/.m2 -maxdepth 4 -type d -mtime +90 -exec rm -rf {} \;
+  MVN_REPO=$(mvn -e --no-transfer-progress help:evaluate -Dexpression=settings.localRepository \
+    -q -DforceStdout);
+  if [[ -d ${MVN_REPO} ]]; then
+    find "$MVN_REPO" -maxdepth 4 -type d -mtime +90 -exec rm -rf {} \;
```
nrmancuso, line 148: I would prefer a simply appending `|| true`, if cache is where we expect, OK. If not, download dependencies and run maven. Adding another maven command just adds fragility and makes for a new point of failure.

nrmancuso (changes requested):
Item:

.ci/travis.sh
rnveach: if `find` errors out, will it display that with `|| true`? Also do you want me to remove the display from the init too?

nrmancuso: >if find errors out, will it display that with || true?

yes:
```
➜  checkstyle git:(fix-link-check) find notpresent -maxdepth 4 -type d -mtime +90 -exec rm -rf {} \;        

find: ‘notpresent’: No such file or directory
➜  checkstyle git:(fix-link-check) echo $?
1
➜  checkstyle git:(fix-link-check) find notpresent -maxdepth 4 -type d -mtime +90 -exec rm -rf {} \; || true

find: ‘notpresent’: No such file or directory
➜  checkstyle git:(fix-link-check) echo $?                                                                  
0

```

If build is failing, what value will location of maven repo provide?

rnveach: It will tell us if maven repo changes location, which would be one cause for the failure.

nrmancuso: I would prefer to keep scripts as simple as possible, and add debugging commands as needed when a problem arises. But let's get a second opinion. @romani ?

romani: I always prefer simplicity of bash scripts. I agree with Nick.
But extra logging is also ok.

romani (approved):

nrmancuso (approved):
Ok, let's merge this.

nrmancuso:
Failure in https://github.com/checkstyle/checkstyle/actions/runs/3453271573/jobs/5763738631 is not related to changes here.

PR #12401, opened 2022-11-13
infra: migrate releasenotes-gen to codeship

nrmancuso:
As releasenotes-gen only runs on master, we can migrate this task to codeship. Currently, this job is failing in CircleCi due to missing github token: https://app.circleci.com/pipelines/github/checkstyle/checkstyle/16059/workflows/82c3e207-5f67-4958-84a0-c4428ea215c8/jobs/197310

PR #12410, opened 2022-11-14
infra: remove releasenotes-gen from circleci

nrmancuso:
releasenotes-gen job is failing without github token, we have a PR to re-enable this job in another CI at https://github.com/checkstyle/checkstyle/pull/12401, but that PR may take some time. Better to disable this since it will constantly fail anyway in repo branches (master, dependabot, etc., not from forked PRs).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,9 +201,6 @@ workflows:
       - validate-with-script:
           name: "checkchmod"
           command: "./.ci/checkchmod.sh"
-      - validate-with-script:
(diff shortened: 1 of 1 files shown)


rnveach (approved):
I would recommend to just comment it out and say "Until ...".

nrmancuso:
> I would recommend to just comment it out and say "Until ...".

No until needed, this job should never be allowed to run to completion in CircleCI for security reasons.

PR #12411, opened 2022-11-15
infra: disable sonar temporarily

nrmancuso:
Fixes failing CI and removes sonar from Github action


.circleci/config.yml
```diff
@@ -43,8 +43,6 @@ jobs:
       - image: << parameters.image-name >>
     steps:
       - checkout
-      - run:
-          command: apt-get install -y jq
```
nrmancuso, line 47: This was causing other CI to fail, since we use different images that require `sudo` or use `apt-get` instead.

romani (approved):

romani:
merging quicker, to fix CI

PR #12414, opened 2022-11-15
infra: remove --fail-with-body from no_old_refs.sh and add suppression

nrmancuso:
Noted at https://github.com/checkstyle/checkstyle/actions/runs/3474029994/jobs/5806727798, we found another old version of curl :)


rnveach (approved):
I am good, but I think the suppressions should have the CI so we can start saying which CIs are good/bad and maybe just do a quick look instead of waiting for an error.

rnveach:
Merging to unblock PRs

nrmancuso:
>I am good, but I think the suppressions should have the CI so we can start saying which CIs are good/bad and maybe just do a quick look instead of waiting for an error.

@rnveach this issue is related to the image used, not the CI; is some cases like travis, we are tied to only a few images, but in others like CircleCI this could be basically anything from dockerhub

PR #12439, opened 2022-11-22
minor: extract Set declaration

nrmancuso:
See https://sonarcloud.io/organizations/checkstyle/rules?open=java%3AS5778&rule_key=java%3AS5778 for details.

>When verifying that code raises a runtime exception, a good practice is to avoid having multiple method calls inside the tested code, to be explicit about which method call is expected to raise the exception.

image


nrmancuso:
CI failure is from missing package: https://app.travis-ci.com/github/checkstyle/checkstyle/jobs/589129155

rnveach (approved):

PR #12446, opened 2022-11-23
infra: remove --fail-with-body from pr-description.sh

nrmancuso:
Noticed at https://github.com/checkstyle/checkstyle/actions/runs/3531403281/jobs/5924533074, check of PR description was silently failing.

```
0s
Run ./.ci/pr-description.sh
curl: option --fail-with-body: is unknown
curl: try 'curl --help' or 'curl --manual' for more information
Commit messages from github: ...
Issue number: 
PR validation succeeded.
```

### DIFF
--- a/.ci/pr-description.sh
+++ b/.ci/pr-description.sh
@@ -4,7 +4,7 @@ set -e
 
 if [[ ! $PULL_REQUEST =~ ^([0-9]*)$ ]]; then exit 0; fi
 LINK_COMMITS=https://api.github.com/repos/checkstyle/checkstyle/pulls/$PULL_REQUEST/commits
-COMMITS=$(curl --fail-with-body -s -H "Authorization: token $READ_ONLY_TOKEN" "$LINK_COMMITS" \
+COMMITS=$(curl -s -H "Authorization: token $READ_ONLY_TOKEN" "$LINK_COMMITS" \
              | jq '.[0] | .commit.message')
 echo 'Commit messages from github: '"${COMMITS:0:60}"...
 ISSUE_NUMBER=$(echo "$COMMITS" | sed -e 's/^.*Issue //' | sed -e 's/:.*//')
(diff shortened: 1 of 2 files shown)


rnveach:
> was silently failing

What should we be doing to avoid these silent fails?

nrmancuso:
> > was silently failing
> 
> What should we be doing to avoid these silent fails?

Do less piping and always check exit value of curl commands maybe?

rnveach:
> Do less piping and always check exit value of curl commands maybe?

I am not a bash programmer so I can't really provide valuable insight. :)

Is it piping that is hiding the failure or is it the `${}`? Can we atleast confirm this?

romani:
this option is ~1 year old only in curl - https://curl.se/mail/lib-2021-03/0056.html
action is using latest-ubuntu, so sad.

romani (approved):
commit is updated.

ok to merge as CI pass

rnveach:
> we can not use bash feature fail on failure in middle of pipe ... because ... in this script

So it is the pipe that causing the hidden failure?
Can we add it to other scripts so they don't have hidden issues?

> we use grep -E "$REGEXP" | cat in this script, so we expect failed commands

Is the failure because the regexp doesn't match like if they don't have issue number and thats what we are checking?

romani:
Pipe is not causing it.
Without bash flag that I mentioned above, failure in the middle swallowed.

rnveach:
ok, it makes sense now. Maybe we should see if we can silo off commands that can fail and those that shouldn't from the grep.

@romani What about this? It has no grep.
````
COMMITS=$(curl -s -H "Authorization: token $READ_ONLY_TOKEN" "$LINK_COMMITS" \
             | jq '.[0] | .commit.message')
````

rnveach (approved):

romani:
If we refactor code to avoid greps or narrow them to specific segments, I am ok to enable  failures in pipelines.
I already did somewhere enablement of such  mode in code and disablement after.

Issue #12079, opened 2022-08-21
Github action "Release perform"


romani:
https://github.com/checkstyle/contribution/blob/master/releasenotes-builder/diagrams-manual-trigger.md

we need to make action to do `run Github action "Push perform '10.3.1'"` (action takes human provided version as input, just to make sure it can not be executed unintentionally)
if user provider input is not matching pom - fail build

so we simply make script that do checkout to required tag and do `mvn:release perform`.
this phase suppose to not push anything to github,
but it should able to deploy jars to maven central.

For testing purposes in your test repo you can do release to some folder, you need to configure release plugin to do so.

romani:
details on file that is created at release:prepare phase - https://github.com/checkstyle/checkstyle/issues/12078#issuecomment-1221438812

Rahulkhinchi03:
Why do we have to use the gpg key here? Can we not do release perform without the gpg key? also, this command doesn't have
 `-gpg`

romani:
We have to use gpg and we already use it during release

Rahulkhinchi03:
Also:
> Note that the setup-java action configures Maven's settings.xml file for you automatically and imports the GPG key from the given stored secret.

We already have our own configured settings.xml, what if we are left with 2 settings.xml files at the end?

Also what about the values? I believe that we need to try and you can do testing by your gpg key, I faced a lot of issues while generating and using gpg key.

romani:
We will try to skip our specific file, in worse case we will use our settings in addition

romani:
@stoyanK7 , please help to test https://github.com/checkstyle/checkstyle/pull/12467#discussion_r1033071815
this might speed up execution it should help in testing.

stoyanK7:
On it.

romani:
This is commands that I run to make all that this action need to do manually - https://github.com/checkstyle/checkstyle/wiki/Semi-automated-release
"Release:Perform"

stoyanK7:
Back on it.

romani:
testing on local for deploy:
```
From 3a6d340dd01b6a2c98aac8aa29dd688b6c2be708 Mon Sep 17 00:00:00 2001
From: stoyanK7 <stoyank127@gmail.com>
Date: Fri, 3 Feb 2023 10:11:35 +0100
Subject: [PATCH] doc: temp release commit

---
 .ci/maven-release-perform.sh | 13 +++++++------
 pom.xml                      | 20 +++++---------------
[...]

PR #12467, opened 2022-11-28
infra: disable test during release phases to speed up execution 

romani:
(CI should be green to start release)
during release it is too late to run tests or any other validation.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -35,11 +35,12 @@ SKIP_CHECKSTYLE="-Dcheckstyle.ant.skip=true -Dcheckstyle.skip=true"
 SKIP_OTHERS="-Dpmd.skip=true -Dspotbugs.skip=true -Djacoco.skip=true -Dxml.skip=true"
 
 echo "Version bump in pom.xml (release:prepare) ..."
-mvn -e --no-transfer-progress -Pgpg release:prepare -B -Darguments="$SKIP_TEST $SKIP_CHECKSTYLE \
-  $SKIP_OTHERS"
+mvn -e --no-transfer-progress -Pgpg release:prepare -B \
+  -Darguments="$SKIP_TEST $SKIP_CHECKSTYLE $SKIP_OTHERS"
 
 echo "Deployment of jars to maven central (release:perform) ..."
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -35,11 +35,12 @@ SKIP_CHECKSTYLE="-Dcheckstyle.ant.skip=true -Dcheckstyle.skip=true"
 SKIP_OTHERS="-Dpmd.skip=true -Dspotbugs.skip=true -Djacoco.skip=true -Dxml.skip=true"
 
 echo "Version bump in pom.xml (release:prepare) ..."
-mvn -e --no-transfer-progress -Pgpg release:prepare -B -Darguments="$SKIP_TEST $SKIP_CHECKSTYLE \
-  $SKIP_OTHERS"
+mvn -e --no-transfer-progress -Pgpg release:prepare -B \
+  -Darguments="$SKIP_TEST $SKIP_CHECKSTYLE $SKIP_OTHERS"
```
rnveach, line 39: Why not just use the `no-validations` profile? You will basically end up reproducing it if you want to exclude everything.

rnveach (changes requested):

release.sh
romani: there are nuances.
we should not skip javadoc for example. I remember something else was not ok to use it. 

We might need to make special profile for release, but it is not very easy to experiment on release process - recovering from release failure is mess. I think in december I will be have time to focus heavily on release to optimize it more.

rnveach: Even with profile, you can turn things on that profile turns off. It can also work with double profiles.
Example:
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/.ci/validation.sh#L50

rnveach (approved):

release.sh
rnveach: If you need to test a release to maven folder, you can try:
https://github.com/rnveach/checkstyle-backport-jre8/blob/master/pom.xml#L656

nrmancuso (approved):

release.sh
romani: we will try at https://github.com/checkstyle/checkstyle/issues/12079

Issue #11807, opened 2022-07-01
Null pointer exception with records in RequireThisCheck


Vyom-Yadav:
I have read check documentation: https://checkstyle.sourceforge.io/config_coding.html#RequireThis
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Test.java
```
```bash
/var/tmp $ cat config.xml
```
```XML
<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
        "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">

<module name="Checker">
  <module name="TreeWalker">
    <module name="RequireThisCheck"/>
  </module>
</module>
```
```bash
/var/tmp $ cat Test.java
```
```java
public record Test(int a) {

    void method() {
        String a = "BUG";
        a = a.substring(0, 1);
    }
}
```
```bash
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.3.1-all.jar -c config.xml Test.java
Starting audit...
com.puppycrawl.tools.checkstyle.api.CheckstyleException: Exception was thrown while processing Test.java
        at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:305)
        at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:224)
        at com.puppycrawl.tools.checkstyle.Main.runCheckstyle(Main.java:407)
        at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:330)
        at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:189)
        at com.puppycrawl.tools.checkstyle.Main.main(Main.java:126)
Caused by: java.lang.NullPointerException
        at com.puppycrawl.tools.checkstyle.checks.coding.RequireThisCheck$ClassFrame.hasFinalField(RequireThisCheck.java:1557)
        at com.puppycrawl.tools.checkstyle.checks.coding.RequireThisCheck.canAssignValueToClassField(RequireThisCheck.java:932)
        at com.puppycrawl.tools.checkstyle.checks.coding.RequireThisCheck.getClassFrameWhereViolationIsFound(RequireThisCheck.java:731)
[...]

PR #12471, opened 2022-11-29
Issue #11807: fixes exception on records and fields in RequireThis

rnveach:
Issue #11807

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
@@ -1562,10 +1562,18 @@ public class RequireThisCheck extends AbstractCheck {
         public boolean hasFinalField(final DetailAST instanceMember) {
             boolean result = false;
             for (DetailAST member : instanceMembers) {
-                final DetailAST mods = member.getParent().findFirstToken(TokenTypes.MODIFIERS);
-                final boolean finalMod = mods.findFirstToken(TokenTypes.FINAL) != null;
-                if (finalMod && isAstSimilar(member, instanceMember)) {
+                final DetailAST parent = member.getParent();
+                if (parent.getType() == TokenTypes.RECORD_COMPONENT_DEF) {
                     result = true;
+                }
+                else {
+                    final DetailAST mods = parent.findFirstToken(TokenTypes.MODIFIERS);
+                    final boolean finalMod = mods.findFirstToken(TokenTypes.FINAL) != null;
+                    if (finalMod && isAstSimilar(member, instanceMember)) {
+                        result = true;
+                    }
+                }
+                if (result) {
                     break;
                 }
             }
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisRecordsWithCheckFields.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisRecordsWithCheckFields.java
@@ -1,0 +1,48 @@
+/*
+RequireThis
+checkFields = true
+checkMethods = (default)true
+validateOnlyOverlapping = (default)true
+
+
+*/
+
+//non-compiled with javac: Compilable with Java14
+package com.puppycrawl.tools.checkstyle.checks.coding.requirethis;
+
+public record InputRequireThisRecordsWithCheckFields(String a) {
+    void method() {
+        String a = "BUG";
+        a = a.substring(0, 1); // ok, field 'a' is final and cannot be changed
+    }
+
+    void method2() {
+        String x = a; // ok, not overlapping
+        String y = this.a; // ok, uses 'this'
+        String a = this.a; // ok, local variable assigned from field
+        a += a; // ok, local variable
+    }
+}
+class InputRequireThisRecordsWithCheckFieldsClass {
+    public final String a;
+
+    public InputRequireThisRecordsWithCheckFieldsClass(String a) {
+        this.a = a;
+    }
+
+    void method() {
+        String a = "BUG";
+        a = a.substring(0, 1); // ok, field 'a' is final and cannot be changed
+    }
+
+    void method2() {
+        String x = a; // ok, not overlapping
+        String y = this.a; // ok, uses 'this'
+        String a = this.a; // ok, local variable assigned from field
+        a += a; // ok, local variable
+    }
+
+    public String a() {
+        return a; // ok
+    }
+}
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisRecordsWithCheckFieldsOverlap.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisRecordsWithCheckFieldsOverlap.java
@@ -1,0 +1,48 @@
+/*
+RequireThis
+checkFields = true
+checkMethods = (default)true
+validateOnlyOverlapping = false
+
+
+*/
+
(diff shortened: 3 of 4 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -1562,11 +1562,14 @@ public boolean hasStaticMethod(final DetailAST ident) {
         public boolean hasFinalField(final DetailAST instanceMember) {
             boolean result = false;
             for (DetailAST member : instanceMembers) {
-                final DetailAST mods = member.getParent().findFirstToken(TokenTypes.MODIFIERS);
-                final boolean finalMod = mods.findFirstToken(TokenTypes.FINAL) != null;
-                if (finalMod && isAstSimilar(member, instanceMember)) {
-                    result = true;
-                    break;
+                final DetailAST parent = member.getParent();
+                if (parent.getType() != TokenTypes.RECORD_COMPONENT_DEF) {
```
rnveach, line 1566: I wasn't fully sure, but I assume this `RECORD_COMPONENT_DEF` is not a final field by default?

rnveach:
Regression: http://rveach.no-ip.org:81/checkstyle/regression/317/
No differences.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisRecordsWithCheckFields.java
```diff
@@ -0,0 +1,18 @@
...
+
+*/
+
+//non-compiled with javac: Compilable with Java14
+package com.puppycrawl.tools.checkstyle.checks.coding.requirethis;
+
+public record InputRequireThisRecordsWithCheckFields(int a) {
+    void method() {
+        String a = "BUG";
+        a = a.substring(0, 1); // violation
```
romani, line 16: Add message

romani (changes requested):
Items

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisRecordsWithCheckFields.java
rnveach: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
nrmancuso: record components are implicitly  `final`.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisRecordsWithCheckFields.java
```diff
@@ -0,0 +1,18 @@
...
+
+*/
+
+//non-compiled with javac: Compilable with Java14
+package com.puppycrawl.tools.checkstyle.checks.coding.requirethis;
+
+public record InputRequireThisRecordsWithCheckFields(int a) {
+    void method() {
+        String a = "BUG";
+        a = a.substring(0, 1); // violation 'Reference to instance variable 'a' needs "this."'
```
nrmancuso, line 16: This should not be a violation, usage of `this` here would not compile due to different types (`int` vs `String`).

nrmancuso: I suggest to make any test input for records have the equivalent class also in the input, to make it easier to compare behavior between the two. Also, we should have the `// ok` equivalent as well.

romani: +1

nrmancuso:
It would be a good idea to run check regression over the projects file that found the NPE: https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/latest-projects-to-test-on.properties. The older project file has projects that have little or no record usage.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
rnveach: Let me know if this code is not correct then. This is basically saying these things aren't final.

rnveach: The code should be "fixed" now.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisRecordsWithCheckFields.java
```diff
@@ -0,0 +1,34 @@
...
+
+*/
+
+//non-compiled with javac: Compilable with Java14
+package com.puppycrawl.tools.checkstyle.checks.coding.requirethis;
+
+public record InputRequireThisRecordsWithCheckFields(int a) {
+    void method() {
+        String a = "BUG";
+        a = a.substring(0, 1); // ok, field 'a' is final and cannot be changed
```
nrmancuso, line 16: Please make local variable and record component/ field the same type, I think this example is confusing. Also, let's add some violations in here such as:
```
    void m() {
        int x = a; // violation, access of field 'a' needs 'this'
        int y = this.a; // ok, uses 'this'
        int a = this.a; // ok, local variable assigned from field
        a += a; // ok, local variable
    }
```

nrmancuso (changes requested):
Item:

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisRecordsWithCheckFields.java
nrmancuso: Ah, I see that validateOnlyOverlapping is true, so this would not be checked I suppose.

nrmancuso: We should probably make another test with this property set to false

rnveach: Done.

rnveach:
Regression: http://rveach.no-ip.org:81/checkstyle/regression/318/

Only 1 difference, no new violation.

nrmancuso (changes requested):
Compilation is failing on input files.

Item:

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisRecordsWithCheckFieldsOverlap.java
```diff
@@ -0,0 +1,48 @@
...
+package com.puppycrawl.tools.checkstyle.checks.coding.requirethis;
+
+public record InputRequireThisRecordsWithCheckFieldsOverlap(String a) {
+    void method() {
+        String a = "BUG";
+        a = a.substring(0, 1); // ok, field 'a' is final and cannot be changed
+    }
+
+    void method2() {
+        String x = a; // violation
```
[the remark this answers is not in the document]
rnveach: This done and compilation.

nrmancuso (approved):

romani (approved):

PR #12487, opened 2022-12-02
Issue #12486: NoWhitespaceAfter shouldn't check synchronized method

### LINKED ISSUE #12486: NoWhitespaceAfter synchronized method
Hello,

I have the following checkstyle configuration in `pom.xml`:
```xml
<module name="NoWhitespaceAfter">
      <property name="tokens" value="ARRAY_INIT,AT,INC,DEC,UNARY_MINUS,UNARY_PLUS,BNOT,LNOT,DOT,TYPECAST,ARRAY_DECLARATOR,INDEX_OP,LITERAL_SYNCHRONIZED,METHOD_REF" />
</module>
```

With a `mvn verify`, the following piece of code reports an error `'synchronized' is followed by whitespace`:

```java
public class TestClass {
    private synchronized Row waitForNextRow() {
    }
}
```

It seems to be a false positive because the check should only apply on structures like:
```java
synchronized(lock) { ... }
```

Here is the output:
```
$ cat TestClass.java
public class TestClass {
    private synchronized Row waitForNextRow() {
    }
}

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
<module name="NoWhitespaceAfter">
      <property name="tokens" value="ARRAY_INIT,AT,INC,DEC,UNARY_MINUS,UNARY_PLUS,BNOT,LNOT,DOT,TYPECAST,ARRAY_DECLARATOR,INDEX_OP,LITERAL_SYNCHRONIZED,METHOD_REF" />
</module>
    </module>
</module>

$ java -jar checkstyle-10.5.0-all.jar -c TestConfig.xml TestClass.java
Starting audit...
[ERROR] TestClass.java:2:13: 'synchronized' is followed by whitespace. [NoWhitespaceAfter]
Audit done.
```

Thanks & best regards.

nrmancuso:
Closes #12486 

Diff Regression config: https://gist.githubusercontent.com/nrmancuso/8875948c95d7ad62310b0ecd72c527ff/raw/df0c55c7670d01c7db1ba09bc8219c96c67b85c9/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceAfterCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceAfterCheck.java
@@ -283,12 +283,8 @@ public class NoWhitespaceAfterCheck extends AbstractCheck {
      * @return true if whitespace after ast should be checked
      */
     private static boolean shouldCheckWhitespaceAfter(DetailAST ast) {
-        boolean checkWhitespace = true;
         final DetailAST previousSibling = ast.getPreviousSibling();
-        if (previousSibling != null && previousSibling.getType() == TokenTypes.ANNOTATIONS) {
-            checkWhitespace = false;
-        }
-        return checkWhitespace;
+        return previousSibling == null || previousSibling.getType() != TokenTypes.ANNOTATIONS;
     }
 
     /**
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceAfterCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceAfterCheck.java
@@ -55,6 +55,13 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
  * <pre>
  * public void foo(final char @NotNull [] param) {} // No violation
  * </pre>
+ * <p>
+ * Note: This check processes the
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_SYNCHRONIZED">
+ * LITERAL_SYNCHRONIZED</a> token only when it appears as a part of a
+ * <a href="https://docs.oracle.com/javase/specs/jls/se19/html/jls-14.html#jls-14.19">
+ * synchronized statement</a>, i.e. {@code synchronized(this) {}}.
+ * </p>
  * <ul>
  * <li>
  * Property {@code allowLineBreaks} - Control whether whitespace is allowed
@@ -284,7 +291,10 @@ public class NoWhitespaceAfterCheck extends AbstractCheck {
      */
     private static boolean shouldCheckWhitespaceAfter(DetailAST ast) {
         final DetailAST previousSibling = ast.getPreviousSibling();
-        return previousSibling == null || previousSibling.getType() != TokenTypes.ANNOTATIONS;
+        final boolean isSynchronizedMethod = ast.getType() == TokenTypes.LITERAL_SYNCHRONIZED
+                        && ast.getFirstChild() == null;
+        return !isSynchronizedMethod
+                && (previousSibling == null || previousSibling.getType() != TokenTypes.ANNOTATIONS);
     }
 
     /**
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceAfterCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceAfterCheck.java
@@ -50,11 +50,9 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
  * INDEX_OP</a> will be ignored.
  * </p>
  * <p>
- * If the annotation is between the type and the array, the check will skip validation for spaces
+ * If the annotation is between the type and the array, like {@code char @NotNull [] param},
+ * the check will skip validation for spaces.
  * </p>
- * <pre>
- * public void foo(final char @NotNull [] param) {} // No violation
- * </pre>
  * <p>
  * Note: This check processes the
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_SYNCHRONIZED">
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/nowhitespaceafter/InputNoWhitespaceAfterSynchronized.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/nowhitespaceafter/InputNoWhitespaceAfterSynchronized.java
@@ -1,0 +1,59 @@
+/*
+NoWhitespaceAfter
+allowLineBreaks = (default)true
+tokens = LITERAL_SYNCHRONIZED
+
+
+*/
+package com.puppycrawl.tools.checkstyle.checks.whitespace.nowhitespaceafter;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+
+public class InputNoWhitespaceAfterSynchronized<T> {
+
+    private final Object lock = new Object();
+
+    void m1() {
+        synchronized (lock) {} // violation
+        synchronized(lock) {} // ok
+    }
+
+    synchronized int m2() { // ok
+        return 2;
+    }
+
+    synchronized private int[] m3() { // ok
+        return new int[]{2};
+    }
+
+    synchronized int[] m4() { // ok
+        return new int[]{2};
+    }
+
+    private final synchronized int m5() { // ok
+        return 2;
+    }
+
+    synchronized T m6() { // ok
+        return null;
+    }
+
+    @SyncAnno synchronized T m7() { // ok
+        return null;
+    }
+
+    @SyncAnno private synchronized T m8() { // ok
+        return null;
+    }
+
+    // Note that we do not check 'synchronized' as a modifier,
+    // convention is to have type parameters of a generic method
+    // be surrounded by whitespace.
+    synchronized <T2> void m9() {} // ok
+    synchronized<T2> void m10() {} // ok
+}
+
+@Target(ElementType.METHOD)
+@interface SyncAnno {
+}
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/whitespace/NoWhitespaceAfterCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/whitespace/NoWhitespaceAfterCheck.xml
@@ -30,7 +30,14 @@
  &lt;/p&gt;
  &lt;pre&gt;
  public void foo(final char @NotNull [] param) {} // No violation
- &lt;/pre&gt;</description>
+ &lt;/pre&gt;
+ &lt;p&gt;
+ Note: This check processes the
+ &lt;a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_SYNCHRONIZED"&gt;
+ LITERAL_SYNCHRONIZED&lt;/a&gt; token only when it appears as a part of a
+ &lt;a href="https://docs.oracle.com/javase/specs/jls/se19/html/jls-14.html#jls-14.19"&gt;
+ synchronized statement&lt;/a&gt;, i.e. {@code synchronized(this) {}}.
+ &lt;/p&gt;</description>
          <properties>
             <property default-value="true" name="allowLineBreaks" type="boolean">
                <description>Control whether whitespace is allowed
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/whitespace/NoWhitespaceAfterCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/whitespace/NoWhitespaceAfterCheck.xml
@@ -26,11 +26,9 @@
  INDEX_OP&lt;/a&gt; will be ignored.
  &lt;/p&gt;
  &lt;p&gt;
- If the annotation is between the type and the array, the check will skip validation for spaces
+ If the annotation is between the type and the array, like {@code char @NotNull [] param},
+ the check will skip validation for spaces.
  &lt;/p&gt;
- &lt;pre&gt;
- public void foo(final char @NotNull [] param) {} // No violation
- &lt;/pre&gt;
  &lt;p&gt;
  Note: This check processes the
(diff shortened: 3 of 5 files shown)


nrmancuso:
Github, generate report

nrmancuso:
Github, generate report

nrmancuso:
Github, generate report

nrmancuso:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceAfterCheck.java
```diff
@@ -283,12 +283,11 @@ private static DetailAST getWhitespaceFollowedNode(DetailAST ast) {
      * @return true if whitespace after ast should be checked
      */
     private static boolean shouldCheckWhitespaceAfter(DetailAST ast) {
-        boolean checkWhitespace = true;
         final DetailAST previousSibling = ast.getPreviousSibling();
-        if (previousSibling != null && previousSibling.getType() == TokenTypes.ANNOTATIONS) {
-            checkWhitespace = false;
-        }
-        return checkWhitespace;
+        final boolean isSynchronizedMethod = ast.getType() == TokenTypes.LITERAL_SYNCHRONIZED
```
nrmancuso, line 287: If `synchronized`  is used for a method, it is a modifier and therefore doesn't have children. If `synchronized` is used in a synchronized statement, it has RPAREN, etc. as children.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/nowhitespaceafter/InputNoWhitespaceAfterSynchronized.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+
+public class InputNoWhitespaceAfterSynchronized<T> {
+
+    private final Object lock = new Object();
+
+    void m1() {
+        synchronized (lock) {} // violation
```
nrmancuso, line 18: A bunch of variations on this can already be found in https://github.com/checkstyle/checkstyle/blob/edcc686e3099db8ccffb70b03cd5dcc0c862177b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/nowhitespaceafter/InputNoWhitespaceAfterTestSynchronized.java#L11

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceAfterCheck.java
strkkk, line 287: I think it is better to make it like
```
final boolean isSynchronizedBlock = ast.getType() == TokenTypes.LITERAL_SYNCHRONIZED 
 && ast.getFirstChild() != null;
return isSynchronizedBlock || previousSibling == null || previousSibling.getType() != TokenTypes.ANNOTATIONS;
```
looks clearer and parens are not needed

src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/nowhitespaceafter/InputNoWhitespaceAfterSynchronized.java
```diff
@@ -0,0 +1,53 @@
...
+
+    synchronized int[] m4() { // ok
+        return new int[]{2};
+    }
+
+    private final synchronized int m5() { // ok
+        return 2;
+    }
+
+    synchronized T m6() { // ok
```
strkkk, line 38: please also add something like
`synchronized <T> void method() {}`

strkkk (changes requested):

src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/nowhitespaceafter/InputNoWhitespaceAfterSynchronized.java
nrmancuso: Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceAfterCheck.java
nrmancuso: Consider:

Current: !(A ^ B) ^ (C ^ !D)
Proposed: (A ^ !B) v C v !D

These are not equivalent; also short circuit of proposed expression causes some issues with array declarations.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/nowhitespaceafter/InputNoWhitespaceAfterSynchronized.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @SyncAnno synchronized T m7() { // ok
+        return null;
+    }
+
+    @SyncAnno private synchronized T m8() { // ok
+        return null;
+    }
+
+    synchronized <T2> void m9() {}
```
strkkk, line 50: it looks like this case 
`synchronized<T> void method() {}` should report a violation, but it does not (?)

nrmancuso: I disagree. In this context, `synchronized` is a modifier, not part of a statement.

Take a look at all of the tokens available to this check:
```
            TokenTypes.ARRAY_INIT,
            TokenTypes.AT,
            TokenTypes.INC,
            TokenTypes.DEC,
            TokenTypes.UNARY_MINUS,
            TokenTypes.UNARY_PLUS,
            TokenTypes.BNOT,
            TokenTypes.LNOT,
            TokenTypes.DOT,
            TokenTypes.TYPECAST,
            TokenTypes.ARRAY_DECLARATOR,
            TokenTypes.INDEX_OP,
            TokenTypes.LITERAL_SYNCHRONIZED,
            TokenTypes.METHOD_REF,
```

`synchronized` already looks out of place, and we can see that there is no modifiers at all in this collection of tokens. So I would think that this check should not validate "no whitespace after" `synchronized` as a modifier at all.

If we did, then we would need to consider adding support for all modifiers to this check, since according to your example (`synchronized<T> void method() {}`), the following should all be violations, too:
```
public class Test {
    synchronized private static<T> void m1(){}
    synchronized private<T> void m2(){}
    private<T> void m3(){}
    strictfp<T> void m4(){}
}
```
Typically, only generic types have the type parameter immediately follow the type name; generic method type parameters are always surrounded by whitespace.

Take the openjdk19u repo as an example:
```
➜  jdk19u git:(master) find . -name "*.java" | wc -l
47947
➜  jdk19u git:(master) ag " [a-z]+<[A-Z]>" --java | wc -l 
113
➜  jdk19u git:(master) ag " [a-z]+ <[A-Z]>" --java | wc -l
2441


```

Surrounding type parameters on generic methods by whitespace is about 22 times more prevalent than not.

nrmancuso: From https://checkstyle.sourceforge.io/config_whitespace.html#GenericWhitespace:

>Left angle bracket ("<"):

> - should be preceded with whitespace only in generic methods definitions.
> - should not be preceded with whitespace when it is preceded method name or constructor.
> - should not be preceded with whitespace when following type name.
> - should not be followed with whitespace in all cases.

rnveach: I feel this should be moved to a new issue and not part of this as it is a bug fix, especially if the token wasn't reported before.

I truthfully can't see someone wanting this option and it is probably better to wait for a request.

strkkk: ok, I agree

strkkk (approved):

rnveach (approved):

src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/nowhitespaceafter/InputNoWhitespaceAfterSynchronized.java
romani: I found this also very confusing.
Please add above/below new test code
`synchronized<T2> void m9NoSpace() {} // ok, as we ignore tokens that are used in modifiers`

To make it clear that we noticed it and do this on purpose

romani (changes requested):
Item: https://github.com/checkstyle/checkstyle/pull/12487/files#r1041634426

nrmancuso:
Github, generate site

nrmancuso:
Github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/NoWhitespaceAfterCheck.java
```diff
@@ -55,6 +55,24 @@
...
  * </pre>
+ * <p>
+ * This check processes the
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_SYNCHRONIZED">
+ * LITERAL_SYNCHRONIZED</a> token only when it appears as a part of a
+ * <a href="https://docs.oracle.com/javase/specs/jls/se19/html/jls-14.html#jls-14.19">
+ * synchronized statement</a>.
+ * </p>
+ * <p>
+ * Example:
```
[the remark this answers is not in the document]
nrmancuso: Example above this was already in the description.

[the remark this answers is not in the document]
nrmancuso: It looks like the existing example had no issues.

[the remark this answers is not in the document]
romani: It might be good idea to make check on this.
@rnveach , is your xml Checker can handle this ?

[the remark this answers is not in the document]
rnveach: CS Extras has basic checks for XML. Indentation, Naming, and Illegal Token and Illegal Token Text. If those work, you can use them, otherwise you will need to create an issue for what your asking for.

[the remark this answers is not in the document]
romani: unfortunately we escape all and place it in `<description>` tag
```
 &lt;/pre&gt;
 &lt;p&gt;
 This check processes the
 &lt;a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_SYNCHRONIZED"&gt;
 LITERAL_SYNCHRONIZED&lt;/a&gt; token only when it appears as a part of a
 &lt;a href="https://docs.oracle.com/javase/specs/jls/se19/html/jls-14.html#jls-14.19"&gt;
 synchronized statement&lt;/a&gt;.
 &lt;/p&gt;
 &lt;p&gt;
 Example:
 &lt;/p&gt;
 &lt;pre&gt;
 void m1() {
     synchronized(this) {} // ok
     synchronized (this) {} // violation
 }
 synchronized &amp;lt;T&amp;gt; void m2() {} // ok
 synchronized&amp;lt;T&amp;gt; void m3() {} // ok
 &lt;/pre&gt;</description>
```

xml validation will not work, we might need to update [JavadocMetadataScraper ](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java)  to not allow `source` tags (there might be already bunch of bad examples that needs to be ignored and fixed separately).
I remember that we were very strict for on this for properties descriptions.
May be for Check descriptions we need more relaxed rules or same. Please share your ideas.

@nrmancuso , while discussion is ongoing, please remove example code and lets merge this fix.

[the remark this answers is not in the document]
romani: >It looks like the existing example had no issues.

@nrmancuso , we need to check how it looks in Sonar or EclipseCS embedded rendering.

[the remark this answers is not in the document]
romani:
```
~/java/github/romani/checkstyle [master|✚ 1] 
$ git diff
diff --git a/src/xdocs/config_whitespace.xml b/src/xdocs/config_whitespace.xml
index 55e3ac4..053c32f 100644
--- a/src/xdocs/config_whitespace.xml
+++ b/src/xdocs/config_whitespace.xml
@@ -1347,12 +1347,10 @@ import static java.math.BigInteger.ZERO;
           will be ignored.
         </p>
         <p>
-         If the annotation is between the type and the array, the check will skip validation for
+         If the annotation is between the type and the array like
+         <code>char @NotNull [] param</code>, the check will skip validation for
          spaces
         </p>
-        <source>
-public void foo(final char @NotNull [] param) {} // No violation
-        </source>
       </subsection>
 
       <subsection name="Properties" id="NoWhitespaceAfter_Properties">

```
image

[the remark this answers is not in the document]
nrmancuso: Issue is opened for check to place violations on source tags in description subsection: https://github.com/rnveach/checkstyle-extras/issues/7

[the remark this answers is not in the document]
nrmancuso: Issue is opened to forbid `<pre>` tags in descriptions at https://github.com/checkstyle/checkstyle/issues/12514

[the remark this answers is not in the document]
nrmancuso: Done

nrmancuso:
Github, generate site

src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/nowhitespaceafter/InputNoWhitespaceAfterSynchronized.java
nrmancuso: Done

romani (approved):

PR #12300, opened 2022-10-13
Issue #9984: restore verify-no-exception-configs

rnveach:
Issue #9984

PR #12504, opened 2022-12-05
infra: avoid travis package management

nrmancuso:
Similar to https://github.com/checkstyle/contribution/pull/700#event-7956954470, avoidance of travis package management will hopefully help to stabilize travis build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,18 +8,15 @@ cache:
   directories:
     - ~/.m2
 
-addons:
-  apt:
-    packages:
-      - groovy
-      - xsltproc
-      - xmlstarlet
-
 branches:
   only:
     - master
 
+before_install:
+  - sudo apt update
+
 install:
+  - sudo apt install -y groovy xsltproc xmlstarlet
   - ./.ci/travis.sh install-adoptium-jdk
   - ./.ci/travis.sh install-custom-mvn
 


romani (approved):
Ok to merge.

@rnveach is required to review as add-on we're added recently at
https://github.com/checkstyle/checkstyle/pull/12300/files

And my https://github.com/checkstyle/contribution/pull/441/files

rnveach:
Travis failed :)
10 minute timeout after 20 minutes.

.travis.yml
```diff
@@ -8,18 +8,15 @@ cache:
...
-
 branches:
   only:
     - master
 
+before_install:
+  - sudo apt update
+
 install:
+  - sudo apt install -y groovy xsltproc xmlstarlet
```
rnveach, line 19: Is this more a temporary solution and that is why we aren't putting these into install scripts?

nrmancuso: No, we typically install dependencies in CI, other than the complicated ones below.

rnveach (approved):

Bananeweizen:
I'm not a Linux guy at all, but doesn't this kind of setup mean that the version of xmlstarlet etc. might be different on every build, depending on what is available in the package managers database? In my private CI setups (not on Travis, but in AWS) I therefore have fixed versions on each thing installed on a CI system, and rather have a Dependabot-like update for those installations.
Sorry if I'm just adding noise. :)

romani:
@Bananeweizen, please never hesitate to share some thoughts, we open to all people in world for review and ideas, not only maintainers

PR #12510, opened 2022-12-08
Issue #11604: allow 3rd party checks to print their own custom asts

rnveach:
Issue #11604

Currently 3rd parties have no way to provide their own AST to do some of the basics that Checkstyle does like printing the ASTs for users to understand it.

This provides a new option in the CLI so custom printers can be used. Other printers were very slightly reworked to this new method to show its usage.

It is hoped to use this method for other CLI functionality like `printXpathBranch` and `printSuppressions`.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Main.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Main.java
@@ -48,6 +48,7 @@ import com.puppycrawl.tools.checkstyle.api.AutomaticBean;
 import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
 import com.puppycrawl.tools.checkstyle.api.Configuration;
 import com.puppycrawl.tools.checkstyle.api.RootModule;
+import com.puppycrawl.tools.checkstyle.api.TreeStringPrinter;
 import com.puppycrawl.tools.checkstyle.utils.ChainedPropertyUtil;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 import com.puppycrawl.tools.checkstyle.utils.XpathUtil;
@@ -87,6 +88,10 @@ public final class Main {
 
     /** Exit code returned when execution finishes with {@link CheckstyleException}. */
     private static final int EXIT_WITH_CHECKSTYLE_EXCEPTION_CODE = -2;
+
+    /** Error message when there is a failure with the custom tree printer. */
+    private static final String CUSTOM_TREE_PRINTER_ERROR =
+            "Unable to work with custom tree printer";
 
     /**
      * Client code should not create instances of this class, but use
@@ -286,31 +291,29 @@ public final class Main {
 
         // create config helper object
         if (options.printAst) {
-            // print AST
-            final File file = filesToProcess.get(0);
-            final String stringAst = AstTreeStringPrinter.printFileAst(file,
-                    JavaParser.Options.WITHOUT_COMMENTS);
-            System.out.print(stringAst);
+            callTreeStringPrinter(filesToProcess.get(0), AstTreeStringPrinter.class);
+        }
+        else if (options.printAstWithComments) {
+            callTreeStringPrinter(filesToProcess.get(0), AstWithCommentsTreeStringPrinter.class);
+        }
+        else if (options.printJavadocTree) {
+            callTreeStringPrinter(filesToProcess.get(0), DetailNodeTreeStringPrinter.class);
+        }
+        else if (options.printTreeWithJavadoc) {
+            callTreeStringPrinter(filesToProcess.get(0), AstAndDetailNodeTreeStringPrinter.class);
+        }
+        else if (Objects.nonNull(options.customTreePrinterClass)) {
+            try {
+                callTreeStringPrinter(filesToProcess.get(0),
+                        Class.forName(options.customTreePrinterClass));
+            }
+            catch (ClassNotFoundException | IOException ex) {
+                throw new CheckstyleException(CUSTOM_TREE_PRINTER_ERROR, ex);
+            }
         }
         else if (Objects.nonNull(options.xpath)) {
             final String branch = XpathUtil.printXpathBranch(options.xpath, filesToProcess.get(0));
             System.out.print(branch);
-        }
-        else if (options.printAstWithComments) {
-            final File file = filesToProcess.get(0);
-            final String stringAst = AstTreeStringPrinter.printFileAst(file,
-                    JavaParser.Options.WITH_COMMENTS);
-            System.out.print(stringAst);
-        }
-        else if (options.printJavadocTree) {
-            final File file = filesToProcess.get(0);
-            final String stringAst = DetailNodeTreeStringPrinter.printFileAst(file);
-            System.out.print(stringAst);
-        }
-        else if (options.printTreeWithJavadoc) {
-            final File file = filesToProcess.get(0);
-            final String stringAst = AstTreeStringPrinter.printJavaAndJavadocTree(file);
-            System.out.print(stringAst);
         }
         else if (hasSuppressionLineColumnNumber) {
             final File file = filesToProcess.get(0);
@@ -339,6 +342,31 @@ public final class Main {
         }
 
         return result;
+    }
+
+    /**
+     * Calls the {@link TreeStringPrinter} class's method with the provided file.
+     *
+     * @param file The file to parse.
+     * @param clss The custom implementation of {@link TreeStringPrinter} to use for printing.
+     * @throws IOException Failed to open a file
+     * @throws CheckstyleException error while parsing the file
+     * @noinspection UseOfSystemOutOrSystemErr
+     * @noinspectionreason UseOfSystemOutOrSystemErr - driver class for Checkstyle requires
+     *      usage of System.out and System.err
+     */
+    private static void callTreeStringPrinter(File file, Class<?> clss)
+            throws IOException, CheckstyleException {
+        final String output;
+        try {
+            output = ((TreeStringPrinter) clss.getDeclaredConstructor().newInstance())
+                    .printFileAst(file);
+        }
+        catch (ClassCastException | ReflectiveOperationException ex) {
+            throw new CheckstyleException(CUSTOM_TREE_PRINTER_ERROR, ex);
+        }
+
+        System.out.print(output);
     }
 
     /**
@@ -754,6 +782,11 @@ public final class Main {
                         + "be specified.")
         private boolean printTreeWithJavadoc;
 
+        /** Option that controls whether to print the custom AST of the file. */
+        @Option(names = {"-pc", "--printWithCustomTree"},
+            description = "Prints Tree based on the provided custom class.")
+        private String customTreePrinterClass;
+
         /** Option that controls whether to print debug info. */
         @Option(names = {"-d", "--debug"},
                 description = "Prints all debug logging of CheckStyle utility.")
@@ -827,11 +860,11 @@ public final class Main {
             }
             // ensure there is no conflicting options
             else if (printAst || printAstWithComments || printJavadocTree || printTreeWithJavadoc
-                || xpath != null) {
+                || xpath != null || customTreePrinterClass != null) {
                 if (suppressionLineColumnNumber != null || configurationFile != null
                         || propertiesFile != null || outputPath != null
                         || parseResult.hasMatchedOption(OUTPUT_FORMAT_OPTION)) {
-                    result.add("Option '-t' cannot be used with other options.");
+                    result.add("Tree Options cannot be used with other options.");
                 }
                 else if (filesToProcess.size() > 1) {
                     result.add("Printing AST is allowed for only one file.");
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/TreeStringPrinter.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/TreeStringPrinter.java
@@ -1,0 +1,41 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code and other text files for adherence to a set of rules.
+// Copyright (C) 2001-2022 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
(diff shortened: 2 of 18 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/Main.java
```diff
@@ -754,6 +777,11 @@ private static class CliOptions {
                         + "be specified.")
         private boolean printTreeWithJavadoc;
 
+        /** Option that controls whether to print the custom AST of the file. */
+        @Option(names = {"-pc", "--printWithCustomTree"},
```
pbludov, line 781: A multi-character short option may be confusing, see
https://picocli.info/#_short_posix_options 

Although the library supports multi-character short names, it collects them into groups:
```bash
  -o=<outputPath>           Sets the output file. Defaults to stdout.
  -p=<propertiesFile>       Sets the property files to load.
      -pc, --printWithCustomTree=<customTreePrinterClass>
                            Prints Tree based on the provided custom class.
```
It looks like `-pc` is an extension to `-p`. We can change it to `-tc` so the usage will be

```bash
  -t, --tree                Prints Abstract Syntax Tree(AST) of the checked file. The option cannot
                              be used other options and requires exactly one file to run on to be
                              specified.
  -T, --treeWithComments    Prints Abstract Syntax Tree(AST) with comment nodes of the checked
                              file. The option cannot be used with other options and requires
                              exactly one file to run on to be specified.
      -tc, --printWithCustomTree=<customTreePrinterClass>
                            Prints Tree based on the provided custom class.
```

src/main/java/com/puppycrawl/tools/checkstyle/Main.java
```diff
@@ -341,6 +341,29 @@ else if (hasSuppressionLineColumnNumber) {
...
+     * @throws CheckstyleException error while parsing the file
+     */
+    private static void callTreeStringPrinter(File file, Class<?> clss)
+            throws IOException, CheckstyleException {
+        final String output;
+        try {
+            output = ((TreeStringPrinter) clss.getDeclaredConstructor().newInstance())
+                    .printFileAst(file);
+        }
+        catch (NoSuchMethodException | InvocationTargetException | InstantiationException
```
pbludov, line 359: This can be simple
```java
        catch (ReflectiveOperationException ex) {
```

src/main/java/com/puppycrawl/tools/checkstyle/Main.java
```diff
@@ -286,31 +288,29 @@ private static int runCli(CliOptions options, List<File> filesToProcess)
...
-            final String stringAst = AstTreeStringPrinter.printJavaAndJavadocTree(file);
-            System.out.print(stringAst);
+            callTreeStringPrinter(filesToProcess.get(0), AstAndDetailNodeTreeStringPrinter.class);
+        }
+        else if (Objects.nonNull(options.customTreePrinterClass)) {
+            try {
+                callTreeStringPrinter(filesToProcess.get(0),
+                        Class.forName(options.customTreePrinterClass));
+            }
+            catch (ClassNotFoundException | IOException | CheckstyleException ex) {
```
pbludov, line 307: No need to catch `CheckstyleException` here. It just wraps one `CheckstyleException` with another:
```bash
com.puppycrawl.tools.checkstyle.api.CheckstyleException: Unable to work with custom tree printer
	at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:308)
	at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:197)
	at com.puppycrawl.tools.checkstyle.Main.main(Main.java:132)
Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Unable to work with custom tree printer
	at com.puppycrawl.tools.checkstyle.Main.callTreeStringPrinter(Main.java:360)
	at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:304)
	... 2 more
```

pbludov, line 307: No need to catch `CheckstyleException` here. It just wraps one `CheckstyleException` with another:
```bash
java -jar checkstyle.jar -pc com.puppycrawl.tools.checkstyle.utils.CheckUtil

com.puppycrawl.tools.checkstyle.api.CheckstyleException: Unable to work with custom tree printer
	at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:308)
	at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:197)
	at com.puppycrawl.tools.checkstyle.Main.main(Main.java:132)
Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Unable to work with custom tree printer
	at com.puppycrawl.tools.checkstyle.Main.callTreeStringPrinter(Main.java:360)
	at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:304)
	... 2 more
```

src/main/java/com/puppycrawl/tools/checkstyle/Main.java
pbludov, line 359: There should be an entry for `ClassCastException`, otherwise no readable error will be shown to the user:
```bash
java -jar checkstyle.jar -pc com.puppycrawl.tools.checkstyle.Main

Exception in thread "main" java.lang.ClassCastException: class com.puppycrawl.tools.checkstyle.Main cannot be cast to class com.puppycrawl.tools.checkstyle.api.TreeStringPrinter (com.puppycrawl.tools.checkstyle.Main and com.puppycrawl.tools.checkstyle.api.TreeStringPrinter are in unnamed module of loader 'app')
	at com.puppycrawl.tools.checkstyle.Main.callTreeStringPrinter(Main.java:356)
	at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:304)
	at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:197)
	at com.puppycrawl.tools.checkstyle.Main.main(Main.java:132)

Comment (pbludov) [reply src/main/java/com/puppycrawl/tools/checkstyle/Main.java:359]: Suggested change:
```java
- catch (NoSuchMethodException | InvocationTargetException | InstantiationException
-                 | IllegalAccessException ex) {
+ catch (ClassCastException | ReflectiveOperationException ex) {
```

src/main/java/com/puppycrawl/tools/checkstyle/Main.java
pbludov, line 781: A multi-character short name may be confusing, see https://picocli.info/#_short_posix_options

Although the library supports multi-character short names, it collects them into groups:
```bash
  -o=<outputPath>           Sets the output file. Defaults to stdout.
  -p=<propertiesFile>       Sets the property files to load.
      -pc, --printWithCustomTree=<customTreePrinterClass>
                            Prints Tree based on the provided custom class.
```

This can be changed to `-tc`:

```bash
  -t, --tree                Prints Abstract Syntax Tree(AST) of the checked file. The option cannot
                              be used other options and requires exactly one file to run on to be
                              specified.
  -T, --treeWithComments    Prints Abstract Syntax Tree(AST) with comment nodes of the checked
                              file. The option cannot be used with other options and requires
                              exactly one file to run on to be specified.
      -tc, --printWithCustomTree=<customTreePrinterClass>
                            Prints Tree based on the provided custom class.
```

pbludov (changes requested):
Couple of minors

src/main/java/com/puppycrawl/tools/checkstyle/Main.java
rnveach: I am fine with this change, but just asking if there is someway we can change `printWithCustomTree` to line up with `tc`? Maybe `treeWithCustomPrinter`?

On a side note, you don't think it may be confused with `treeWithComments`?

src/main/java/com/puppycrawl/tools/checkstyle/Main.java
rnveach: Done in the next push.

src/main/java/com/puppycrawl/tools/checkstyle/Main.java
rnveach: Done in the next push.

src/main/java/com/puppycrawl/tools/checkstyle/Main.java
```diff
@@ -827,11 +860,11 @@ private List<String> validateCli(ParseResult parseResult, List<File> filesToProc
             }
             // ensure there is no conflicting options
             else if (printAst || printAstWithComments || printJavadocTree || printTreeWithJavadoc
-                || xpath != null) {
+                || xpath != null || customTreePrinterClass != null) {
                 if (suppressionLineColumnNumber != null || configurationFile != null
                         || propertiesFile != null || outputPath != null
                         || parseResult.hasMatchedOption(OUTPUT_FORMAT_OPTION)) {
-                    result.add("Option '-t' cannot be used with other options.");
+                    result.add("Tree Options cannot be used with other options.");
```
rnveach, line 867: This wording was identified as out of date and confusing for options where `-t` is not used.

Examples:
````
PS M:\> java -jar .\checkstyle-10.5.0-all.jar -s out -t .
Option '-t' cannot be used with other options.
PS M:\> java -jar .\checkstyle-10.5.0-all.jar -s out -T .
Option '-t' cannot be used with other options.
PS M:\> java -jar .\checkstyle-10.5.0-all.jar -s out -J .
Option '-t' cannot be used with other options.
````

File: src/main/java/com/puppycrawl/tools/checkstyle/Main.java (hunk 1 above)
Comment (pbludov) [reply src/main/java/com/puppycrawl/tools/checkstyle/Main.java:786]: Yes, it is confusing. Other options may be: `-C` or no short option at all.

**Review by pbludov — APPROVED** [review]

File: src/main/java/com/puppycrawl/tools/checkstyle/Main.java (hunk 1 above)
Comment (rnveach) [reply src/main/java/com/puppycrawl/tools/checkstyle/Main.java:786]: I removed the short. I was going down that path before, but the `CliOptionsXdocsSyncTest` was still picking it up as a short option even though it wasn't short at all. I changed the code there.

**Review by nrmancuso — CHANGES_REQUESTED** [review]
What alternatives have been considered to solve this problem? Extension of CLI/API like this deserves it's own issue/discussion IMO. Won't we need to add "alternative" options for others, such as `-g`, `-s`, and `-b`?

**Comment by rnveach** [comment]
> Won't we need to add "alternative" options for others, such as -g, -s, and -b?

See the end of my first post at https://github.com/checkstyle/checkstyle/pull/12510#issue-1485542892 .
`It is hoped to use this method for other CLI functionality like printXpathBranch and printSuppressions.`

I was planning things in future PRs after this one.

> deserves it's own issue

I can create the issue if that is what is needed.

> What alternatives have been considered to solve this problem?

There is no alternative in Checkstyle as we don't allow any extension of the CLI. If this PR isn't acceptable, then there isn't a lot of good options.

3rd parties could create a hack of a check who's only function is to print the tree of the file it is given, which would require a configuration for nothing more than specifying the tree printer which I am passing in as an argument now.

Other than that, 3rd parties will have to duplicate Checkstyle's CLI as they will need the same options (plus their own) if we were to allow them to use https://github.com/checkstyle/contribution/issues/604 .

**Comment by rnveach** [comment]
> Is allowing extension of CLI another option?

**Comment by rnveach** [comment]
> Is allowing extension of CLI another option? 

I am open to this option, but I am not fully seeing how to be honest.

Any extension is based picocli allowing this as it is the main control of the CLI and probably some overridable methods to allow such custom execution.

https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L629
I am going to assume first this would have to be made a public class so 3rd parties can extend it. I am not sure if picocli will accept that.

https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L113-L114
Somehow this has to be overridable that 3rd parties can provide their new CLI instance with their new options.

https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L188
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L195
Then we would need to override these remaining methods for the main logic and ensure they stay of main repo's way if a 3rd party option is provided.

If this is possible, I am seeing this require main repo make classes public and methods protected to be able to do an `extends` on it.

If we prefer this way, I can look into it.

**Comment by nrmancuso** [comment]
If we want to inject dependencies, we could do something like:

```
public class Main {
    
    CliOptions clioptions; // injected, can read concrete class name from property file, bean, etc.
    CliRunner cliRunner; // injected, replaces `runCli` and other functionality, also can specify class via property file, bean, etc.

    public static void main(String... args){
        Main m = new Main();
        m.run(args);
    }

    private void run(String... args) {
        int e = cliRunner.execute(cliOptions);
        Runtime.getRuntime().exit(e);
    }
...
```

File: src/main/java/com/puppycrawl/tools/checkstyle/api/TreeStringPrinter.java [hunk 5]
```diff
@@ -0,0 +1,41 @@
...
+package com.puppycrawl.tools.checkstyle.api;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Interface for printing ASTs to String.
+ */
+@FunctionalInterface
+public interface TreeStringPrinter {
```
strkkk, line 29: I dont like this design, to be honest.

method name like `printFileAst` assumes that 
1. there is no return, similar to `System.out.println()` and similar methods.
2. Result is the AST, which is weird since return type is String, which means any text.

Plus, name of the interface is confusing, since even existing implementations dont do any printing of the AST, they just convert it into string. Printing is performed outside, in Main.java line 369.

Therefore I suggest to rename it to something like FileConverter (or something like this) with meaningful method name.

strkkk (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/api/TreeStringPrinter.java
rnveach: @nrmancuso @strkkk @pbludov 
Is there any agreement we can make on design before I continue? @nrmancuso Was mentioning extension over something being built into the project. For @strkkk 's comment, everything was originally `printFileAst` so thats why I went with this, but if we head down extension, then naming becomes not an issue as 3rd parties will be the ones to name it themselves.

nrmancuso: Can we make https://github.com/checkstyle/checkstyle/blob/a8f065738de6c4ff917a2ee57521f6cf8718f2ba/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L629 a public class to allow extension of options? There is bound to be more things that some third party may need to do with the CLI, and this way they would not be blocked by limitations of checksytle CLI.

romani:
I do not like this extension of our API, i think it should be done completely in thirdparty projects such custom support.

romani:
no more ideas or points to reconcile disfigurements.
I am closing this PR.
We will focus in our api on what we need, thirparty projects can clone us and hack us as they want to get what they need, we will not make it as our contract to support.
We already have bunch of implementations and other unfortunate decisions in our API we need to focus on our needs only.

romani:
can we use this property and print tree in format thatwe need at https://github.com/checkstyle/checkstyle/issues/14631  ?

PR #12511, opened 2022-12-08
Issue #12104: prepare order of execution for global locale issue

rnveach:
Issue #12104

This is the first PR to fix the overall issue. This is a sort of prep PR and is more minor and smaller. Unfortunately some context here might not make sense without the final product but I will try to explain as much as I can.

The point of this PR is:
* Ensure we are executing tests in the correct lifeycycle order. `configure`/`finishLocalSetup` should run first before `setupChild` which also should run before any custom "adds" to the root module. finishLocalSetup is executing before setupChild https://github.com/checkstyle/checkstyle/blob/b6301c189ef5d2d731a755fa9a2e84aa0677494a/src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java#L191-L207
 * I did this by adding custom code on the locale to print exceptions if things went out of order. It is not code that will remain for the PR.
* Ensure anything we are adding that is breaking the above bullet, is custom, to not interfere with the coming changes.
* Ensure everything we pull from the configuration extends `AutomaticBean` as we have declared all modules should be using it, which gives everything access to `contextualize`.
 * Source: https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java#L75-L76

------

### **Main point**

The changes I am planning for the bigger stage is anything going through `add*` in `Checker` (our root module) needs to call `contextualize` if allowed. I am planning to give all modules the `Locale` through the `contextualize` by adding this on [the end of `finishLocalSetup`](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L432). There are too many different types of areas that need the Locale in Checkstyle from Listeners to Filters, so this is the best way to spread the locale around in the current system.

---------

The lifecycle order is not a big deal as our Main is already following it. I confirmed maven-checkstyle is also following it.

A peek at the future PR can be seen at https://github.com/rnveach/checkstyle/commits/issue_12104 but it will be in some flux.

---------

how to install different locale on ubuntu host:
recheck installed locales:
```
$ locale -a
C
C.UTF-8
en_AG
en_AG.utf8
en_AU.utf8
en_BW.utf8
en_CA.utf8
en_DK.utf8
en_GB.utf8
en_HK.utf8
en_IE.utf8
en_IN
en_IN.utf8
en_NG
en_NG.utf8
en_NZ.utf8
en_PH.utf8
en_SG.utf8
en_US.utf8
en_ZA.utf8
en_ZM
en_ZM.utf8
en_ZW.utf8
POSIX
```

```
sudo apt-get install language-pack-ru
sudo update-locale LANG=ru_RU.utf8
reboot
```

all locales
```
$ locale -a
C
C.UTF-8
en_AG
en_AG.utf8
en_AU.utf8
en_BW.utf8
en_CA.utf8
en_DK.utf8
en_GB.utf8
en_HK.utf8
en_IE.utf8
en_IN
en_IN.utf8
en_NG
en_NG.utf8
en_NZ.utf8
en_PH.utf8
en_SG.utf8
en_US.utf8
en_ZA.utf8
en_ZM
en_ZM.utf8
en_ZW.utf8
POSIX
ru_RU.utf8
ru_UA.utf8
```

current locale
```
$ locale
LANG=ru_RU.utf8
LANGUAGE=ru
LC_CTYPE="ru_RU.utf8"
.....
$ mvn clean test -Dtest=CheckstyleAntTaskTest
.....
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTaskTest
[ERROR] Tests run: 42, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 0.838 s <<< FAILURE! - in com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTaskTest
[ERROR] com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTaskTest.testExecuteIgnoredModules  Time elapsed: 0.094 s  <<< FAILURE!
Content of file with violations differs from expected
expected: Starting audit...
but was : Начинаем проверку...
	at com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTaskTest.testExecuteIgnoredModules(CheckstyleAntTaskTest.java:390)
	at [[Reflective call: 4 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at [[Testing framework: 27 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at [[Testing framework: 9 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at [[Testing framework: 21 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at org.apache.maven.surefire.junitplatform.LazyLauncher.execute(LazyLauncher.java:50)
	at org.apache.maven.surefire.junitplatform.JUnitPlatformProvider.execute(JUnitPlatformProvider.java:184)
	at org.apache.maven.surefire.junitplatform.JUnitPlatformProvider.invokeAllTests(JUnitPlatformProvider.java:148)
	at org.apache.maven.surefire.junitplatform.JUnitPlatformProvider.invoke(JUnitPlatformProvider.java:122)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:385)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:162)
	at org.apache.maven.surefire.booter.ForkedBooter.run(ForkedBooter.java:507)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:495)
```

workaround (use existing/installed on your local "en" locale):
```
$ export LC_ALL=en_US.utf8
$ mvn verify
.... passes ...
```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
@@ -137,7 +137,7 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
      * The instance needs to be contextualized and configured.
      */
     public Checker() {
-        addListener(counter);
+        listeners.add(counter);
         log = LogFactory.getLog(Checker.class);
     }
 
@@ -469,11 +469,9 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
         try {
             child = moduleFactory.createModule(name);
 
-            if (child instanceof AbstractAutomaticBean) {
-                final AbstractAutomaticBean bean = (AbstractAutomaticBean) child;
-                bean.contextualize(childContext);
-                bean.configure(childConf);
-            }
+            final AbstractAutomaticBean bean = (AbstractAutomaticBean) child;
+            bean.contextualize(childContext);
+            bean.configure(childConf);
         }
         catch (final CheckstyleException ex) {
             throw new CheckstyleException("cannot initialize module " + name
@@ -482,19 +480,20 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
         if (child instanceof FileSetCheck) {
             final FileSetCheck fsc = (FileSetCheck) child;
             fsc.init();
-            addFileSetCheck(fsc);
+            fsc.setMessageDispatcher(this);
+            fileSetChecks.add(fsc);
         }
         else if (child instanceof BeforeExecutionFileFilter) {
             final BeforeExecutionFileFilter filter = (BeforeExecutionFileFilter) child;
-            addBeforeExecutionFileFilter(filter);
+            beforeExecutionFileFilters.addBeforeExecutionFileFilter(filter);
         }
         else if (child instanceof Filter) {
             final Filter filter = (Filter) child;
-            addFilter(filter);
+            filters.addFilter(filter);
         }
         else if (child instanceof AuditListener) {
             final AuditListener listener = (AuditListener) child;
-            addListener(listener);
+            listeners.add(listener);
         }
         else {
             throw new CheckstyleException(name
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/testmodules/TestBeforeExecutionFileFilter.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/testmodules/TestBeforeExecutionFileFilter.java
@@ -19,11 +19,18 @@
 
 package com.puppycrawl.tools.checkstyle.internal.testmodules;
 
+import com.puppycrawl.tools.checkstyle.api.AutomaticBean;
 import com.puppycrawl.tools.checkstyle.api.BeforeExecutionFileFilter;
 
-public final class TestBeforeExecutionFileFilter implements BeforeExecutionFileFilter {
+public final class TestBeforeExecutionFileFilter extends AutomaticBean
+        implements BeforeExecutionFileFilter {
 
     private boolean called;
+
+    @Override
+    protected void finishLocalSetup() {
+        // no code
+    }
 
     @Override
     public boolean accept(String uri) {
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -106,17 +106,6 @@
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/Checker.java</fileName>
-    <specifier>method.invocation</specifier>
-    <message>call to addListener(com.puppycrawl.tools.checkstyle.api.AuditListener) not allowed on the given receiver.</message>
-    <lineContent>addListener(counter);</lineContent>
-    <details>
-      found   : @UnderInitialization(com.puppycrawl.tools.checkstyle.AbstractAutomaticBean.class) @NonNull Checker
-      required: @Initialized @NonNull Checker
-    </details>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java</fileName>
     <specifier>argument</specifier>
     <message>incompatible argument for parameter arg0 of add.</message>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
@@ -75,6 +75,7 @@ import com.puppycrawl.tools.checkstyle.checks.NewlineAtEndOfFileCheck;
 import com.puppycrawl.tools.checkstyle.checks.TranslationCheck;
 import com.puppycrawl.tools.checkstyle.checks.coding.HiddenFieldCheck;
 import com.puppycrawl.tools.checkstyle.filters.SuppressionFilter;
+import com.puppycrawl.tools.checkstyle.filters.SuppressionXpathFilter;
 import com.puppycrawl.tools.checkstyle.internal.testmodules.DebugAuditAdapter;
 import com.puppycrawl.tools.checkstyle.internal.testmodules.DebugFilter;
 import com.puppycrawl.tools.checkstyle.internal.testmodules.TestBeforeExecutionFileFilter;
@@ -117,6 +118,8 @@ public class CheckerTest extends AbstractModuleTestSupport {
     @Test
     public void testDestroy() throws Exception {
         final Checker checker = new Checker();
+        checker.setModuleClassLoader(Thread.currentThread().getContextClassLoader());
+        checker.finishLocalSetup();
         final DebugAuditAdapter auditAdapter = new DebugAuditAdapter();
         checker.addListener(auditAdapter);
         final TestFileSetCheck fileSet = new TestFileSetCheck();
@@ -154,6 +157,8 @@ public class CheckerTest extends AbstractModuleTestSupport {
     public void testAddListener() throws Exception {
         final Checker checker = new Checker();
         final DebugAuditAdapter auditAdapter = new DebugAuditAdapter();
+        checker.setModuleClassLoader(Thread.currentThread().getContextClassLoader());
+        checker.finishLocalSetup();
         checker.addListener(auditAdapter);
 
         // Let's try fire some events
@@ -208,6 +213,8 @@ public class CheckerTest extends AbstractModuleTestSupport {
     @Test
     public void testRemoveListener() throws Exception {
         final Checker checker = new Checker();
+        checker.setModuleClassLoader(Thread.currentThread().getContextClassLoader());
+        checker.finishLocalSetup();
         final DebugAuditAdapter auditAdapter = new DebugAuditAdapter();
         final DebugAuditAdapter aa2 = new DebugAuditAdapter();
         checker.addListener(auditAdapter);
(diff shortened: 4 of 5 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -138,7 +138,7 @@ public class Checker extends AutomaticBean implements MessageDispatcher, RootMod
      * The instance needs to be contextualized and configured.
      */
     public Checker() {
-        addListener(counter);
+        listeners.add(counter);
```
rnveach, line 141: We won't have the childContext when this is called since we haven't read in the config yet, so I feel it is better to just add the listener directly like this as it will not need any of that information. This will clearly show it is something different.

rnveach: We should similarly be doing the same with https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L486-L498 as we need to differentiate items being brought in from the config versus those being added from outside forces through our `add*` methods in `Checker`. I am starting to think those add methods should be removed as they seem more for test only.

nrmancuso: I do not understand how this will make any difference in this case. Do we have a test that will fail without directly accessing collections?

nrmancuso:
>A peek at the future PR can be seen at https://github.com/rnveach/checkstyle/commits/issue_12104 but it will be in some flux.

Future PR will be comprised of `start` and following commit?

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
rnveach: https://github.com/rnveach/checkstyle/commit/9a5598a18525509a6467e3138d5b5d6d7df246f9#diff-650e834bca871d6bf6c1b3800ff4116b2d7ab4fe5de6489a042652c2f72e024dR555-R558
https://github.com/rnveach/checkstyle/commit/9a5598a18525509a6467e3138d5b5d6d7df246f9#diff-650e834bca871d6bf6c1b3800ff4116b2d7ab4fe5de6489a042652c2f72e024dL501-R504

The `add` methods will be changed to methods for outside us only. This is to prevent accidentally `contextualize`ing the object twice, once inside the `setupChild` method and again inside `add`. If it is something this class absolutely needs, then it should handle all its needs itself instead of mixing it up with outside forces.

The way I currently coded it, there will be no failures, but improper use will lead to things not working like if they expect the locale, hence why this PR is titled "order of execution".

rnveach:
> Future PR will be comprised of start and following commit?

Correct, anything not this PR and not in master.

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
rnveach: Imagine a bank can have money deposited through a teller or through an armored car which the bank owns. You will want different checks for each. The personnel depositing money through the teller is fraud in some way. On the other end, you will want to make sure the armored car has the right credentials to be let into the vault to directly deposit.

I am adding this separation here to ensure there are no mistakes by confusing layers. Luckily, most of this being done is just very simple methods that have no big impact.

nrmancuso: What will prevent some future bank manager from sending the armored car through the teller? Especially since the interface specifies this?

rnveach: There is nothing. How can i tell a class to not use it's own public method? I am open to suggestions.

nrmancuso:
After thinking on changes here, and further proposed changes at https://github.com/checkstyle/checkstyle/pull/12511#discussion_r1046651114, it concerns me that we are making a bunch of changes in the API package just for tests.

If we can have issues with global locale value in actual usage, we should reconsider Checker design. If issue is with tests only, changes should be confined to test packages, IMO. @romani thoughts?

rnveach:
Really this is just to make overall change easier to review. These changes are for the next PR which is the big one, which really couldn't be split.

If you want, I can split test changes between API changes, which maybe I should have done to begin with. Test changes are just to ensure are tests are following proper order for future PR. API changes ties in to proper order which next PR will require more of, or else things won't work as expected.

rnveach:
@romani ping

rnveach:
@romani ping

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -482,19 +480,20 @@ protected void setupChild(Configuration childConf)
         if (child instanceof FileSetCheck) {
             final FileSetCheck fsc = (FileSetCheck) child;
             fsc.init();
-            addFileSetCheck(fsc);
+            fsc.setMessageDispatcher(this);
+            fileSetChecks.add(fsc);
```
romani, line 484: why we try to not use addXXXX methods ? in this specific case it is not single statement wrapper.

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
romani: > I am starting to think those add methods should be removed as they seem more for test only, but it seems maven-checkstyle uses one of them.

we support Bean based configuration of Checkstyle to let users use us without XML, for example Sonar plugin or any other plugin with different persistence of module configs.

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
romani: this most likely an answer - https://github.com/checkstyle/checkstyle/pull/12511/files#r1046651114

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -469,11 +469,9 @@ protected void setupChild(Configuration childConf)
...
             child = moduleFactory.createModule(name);
 
-            if (child instanceof AbstractAutomaticBean) {
-                final AbstractAutomaticBean bean = (AbstractAutomaticBean) child;
-                bean.contextualize(childContext);
-                bean.configure(childConf);
-            }
+            final AbstractAutomaticBean bean = (AbstractAutomaticBean) child;
+            bean.contextualize(childContext);
+            bean.configure(childConf);
```
romani, line 474: removal of this IF can mean breaking changes, as some thirdparty modules might stop to work.
If module does not need properties or any Checkers/Treewalker properties, it is ok to be not AbstractAutomaticBean, even we do not recommend it and highly expect.

lets keep it as it was before, with IF, to not deal with breaking old and not not well done modules. At least we need to consider to make it as separate and very explicit breaking change.

romani (changes requested):
item:

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
romani: does it make sense do make private method something line `addFileSetCheckInternal` to keep two statement together.

romani: if we do this update only for future update, lets not do this now, it is very small, it will looks more natural in next PR when addListenter start to have new code that we do not want to apply each time.

romani:
if you agree with all above, we can keep changes only for Tests, and make commit as `minor: ......... for #12104`

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
rnveach: I don't remember details now, but I assume this was required by pitest.

romani: lets restore this block and add pitest to suppression for now to clearly see it. As we agree on PR , covering pitest will be final polishing.

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
romani: @rnveach , please reply.

src/test/java/com/puppycrawl/tools/checkstyle/internal/testmodules/TestBeforeExecutionFileFilter.java
```diff
@@ -19,12 +19,19 @@
 
 package com.puppycrawl.tools.checkstyle.internal.testmodules;
 
+import com.puppycrawl.tools.checkstyle.api.AutomaticBean;
 import com.puppycrawl.tools.checkstyle.api.BeforeExecutionFileFilter;
 
-public final class TestBeforeExecutionFileFilter implements BeforeExecutionFileFilter {
+public final class TestBeforeExecutionFileFilter extends AutomaticBean
```
romani, line 25: we can move all updates for ` extends AutomaticBean` to separate minor commit and merge it sooner.
all production classes that are from `implements BeforeExecutionFileFilter` , `implements AuditListener`, `implements Filter ` are beans, so making them more real is good at any time.

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
romani: this method was used only here
image

it is public, so theoretically can be used by plugins, but we are kind deprecating it ? we need to figureout how much it will affect plugins/extensions.

romani (changes requested):
items:

romani:
@rnveach , please share you thoughts on my comments. Do you want continue or we need concurrent PR with different approach?

rnveach:
I don't understand what you consider a "big API change" or how you will remove the Global Static Locale, which is the heart of this issue, while avoiding what I had to accomplish in my PR by passing the locale from Checker where it is decided.

> we need to figureout how much it will affect plugins/extensions.

https://github.com/checkstyle/checkstyle/issues/12104#issuecomment-1414938002

romani:
ok, I am assigning this PR to me.
Looks like I need to provide one more PR with draft implementation of different approach, so we will have more visual definitions.

Issue #7416, opened 2019-12-29
Simplify JavadocStyleCheck


rnveach:
JavadocStyleCheck currently does 4 things, which are duplicates of other checks.
https://github.com/checkstyle/checkstyle/blob/7c693ccc937109e4498c7756af0b5548348f42a7/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java#L326-L349

1) Ensure package-info.java has a Javadoc.
2) Checks the first sentence ends with a period and doesn't contain a specific fragement.
3) Checks that the javadoc isn't empty.
4) Checks that the tags have respective close for an open.

1 is a duplicate of MissingJavadocPackage. 2 is a duplicate of SummaryJavadoc but doesn't have a scope restriction. 3 is a duplicate of MissingJavadocMethod and MissingJavadocType. 4 is the only thing unique to this check.

I say we start new issues to remove the duplicate functionality, and provide the missing functionality to the other checks. This will help identify what is left for this check to be converted to an AbstractJavadocCheck.

romani:
@rnveach ,

> 4. Checks that the tags have respective close for an open.
 4 is the only thing unique to this check.

I think that all antlr based checks do this, as it is requirement of proper parsing.
if this is true , we can simply remove this Check, and make migration note on what Check(configs) to use instead.

rnveach:
@romani I assume you are referring to https://checkstyle.sourceforge.io/writingjavadocchecks.html#Tight-HTML_rules .

This violation is disabled by default. I don't know the validation the check does versus tight html to know if they are the same or not. I can definitely run regression to see if they are as another issue.

Are you ok with me to proceed with everything else and start issues for each one?

romani:
yes , https://checkstyle.org/config_javadoc.html#SummaryJavadoc , property `violateExecutionOnNonTightHtml` is disabled by default. But we can ask users to activate it if they care.

> I don't know the validation JavadocStyleCheck does versus tight html to know if they are the same or not

please briefly take a look. It is kind does not matter .... we will not support custom parsing of HTML ... so if any diff exists it does not matter that much.

> Are you ok with me to proceed with everything else and start issues for each one?

if you prove that usage `violateExecutionOnNonTightHtml=true` is better/same we simply need to remove Check, I am not sure why might need separate issues.

rnveach:
> 2 ... SummaryJavadoc (since 6.0) but doesn't have a scope restriction.

[...]

PR #12512, opened 2022-12-09
Issue #11166: Fix existing case of getFileContents() usage In JavaDocStyleCheck

arinmodi:
Part of #11166 : Fix existing case of getFileContents() usage In JavaDocStyleCheck

Diff Regression config: https://gist.githubusercontent.com/arinmodi/51e37018c484192ab9f6408df457b944/raw/833988a313a7a5add43c65718f5122a2cc9c0509/javadocstyle.xml


rnveach:
@arinmodi Please work on fixing CI.

We will eventually need regression, but please note CI is already reporting regression issues with an NPE. I would look into fixing most of CI before providing us a regression report to not duplicate CI's reporting already.
https://github.com/checkstyle/checkstyle/actions/runs/3660015653/jobs/6186672529#step:5:4540

arinmodi:
@rnveach, I found out what the error is and where it is in my code, but I am not able to locate the test file which generated this error.

Where can i find that ?

rnveach:
The CI specifies the command that is run, you can look there and see where it is being cloned from and pull in test cases that way.
The one I mentioned before is `.ci/no-exception-test.sh openjdk17-with-checks-nonjavadoc-error`.

arinmodi:
how to solve the error related to the coverage, it is saying ``` Failed to execute goal org.jacoco:jacoco-maven-plugin:0.8.8:check (default-check) on project checkstyle: Coverage checks have not been met. See log for details```. 

Do I need to add test ?

rnveach:
Yes, you must add tests as it is saying you have code missing coverage reported by jacoco.
https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=12449&view=logs&j=eb841e9f-6e1e-5c29-3ad5-74ec2c658069&t=a9b761e6-b8d5-50cf-467a-54967189cab2&l=1218
> [WARNING] Rule violated for class com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocStyleCheck: branches covered ratio is 0.94, but expected minimum is 1.00

https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations

arinmodi:
Hey What's the meaning of 1``` of 6 branch missed ```,```1 of 4 branch missed```, In code converge ?

rnveach:
You are going to want to become familar with code coverage and branch coverage. You should use google to help you.
Example: https://linearb.io/blog/what-is-branch-coverage/

arinmodi:
Github, generate report

arinmodi:
Hey @rnveach, is my approach of getting javadoc correct, or am i doing something wrong ?

rnveach:
@arinmodi Honestly, you will know if you have the correct code to get the Javadoc when regression returns no differences with your changes.

I will say your code is already looking complex and I think these changes might become part of a bigger issue.

Javadoc identification is not an easy process. Take a look at `InvalidJavadocPositionCheck` which was built around utilities to know if something is a javadoc, and it is in the correct location. `AbstractJavadocCheck` uses the same code as its process.
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/InvalidJavadocPositionCheck.java#L106-L107
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck.java#L304

Since this check says Javadoc, I am sure this is on the radar for "Why is this an AbstractCheck instead of an AbstractJavadocCheck" since it is not making use of the javadoc parser. It has been said before these things should be swapped but nothing has progressed on that. https://github.com/checkstyle/checkstyle/issues/7416 is the closest I could find for `JavadocStyleCheck`.

romani:
If no regression is archive-able we can postpone migration to javadoc parser to next  issue or PR.

arinmodi:
Github, generate report

arinmodi:
Hey @romani @rnveach,

is storing all the javadoc comments of file in arraylist good idea ?

Than On Visit Token, we can do just linear search to find corresponding java doc or may be we can do binary search also.

Thanks

arinmodi:
is ```grammar``` refers to the AST tree which is build from file ?

rnveach:
Yes.

arinmodi:
> > ...storing all the javadoc comments of file in...
> 
> https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java#L56 https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java#L215 That is virtually what the code is doing now.
> 
> I would question why we are doing something special for just this check alone and not do it for all the others. And even if we did that, I would question how is it any different then making use of the forbidden `FileContents`.
> 
> The point of this issue is to not rely on the file contents, but rely on the grammars we have built as they are the only things that can parse the file correctly.

So, Using this approach is good, https://github.com/arinmodi/checkstyle/commit/244fcfdbd123ce3ccba4d6abf648f5f872b13b7c, right ?

rnveach:
I will point back to my previous comment at https://github.com/checkstyle/checkstyle/pull/12512#issuecomment-1345693313 about the current state of the code

arinmodi:
@rnveach,

I was trying to generate code coverage report and got this error,

image

I think it's related to some suppression, but not able to find where should I add that suppression

arinmodi:
Github, generate report

arinmodi:
Github, generate report

romani:
It is Junit validation, no suppression, just follow failure and add annotation

arinmodi:
@romani, 

image

In my local these changes passes the test but in regression report it shows something else here, 

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/79c5f59_2022210031/reports/diff/checkstyle/xref/home/runner/work/checkstyle/checkstyle/contribution/checkstyle-tester/repositories/checkstyle/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocstyle/pkginfo/invalidformat/package-info.java.html#L19

is it intended or bug ?

arinmodi:
Hey @rnveach @romani, I am seeing weird behavior, during the test it's working good but during ```mvn clean verify``` it is generating violation which shouldn't be there, is there different AST format for test and actual code ?

nrmancuso:
> Hey @rnveach @romani, I am seeing weird behavior, during the test it's working good but during `mvn clean verify` it is generating violation which shouldn't be there, is there different AST format for test and actual code ?

If you are running the test via the “play” button in your IDE, it has probably cached something. You should consider `mvn clean verify` as your source of truth on local.

arinmodi:
So, how do i clear that cache ?

arinmodi:
> > Hey @rnveach @romani, I am seeing weird behavior, during the test it's working good but during `mvn clean verify` it is generating violation which shouldn't be there, is there different AST format for test and actual code ?
> 
> If you are running the test via the “play” button in your IDE, it has probably cached something. You should consider `mvn clean verify` as your source of truth on local.

Hey I don't think that can be the reason for this cause I am printing on console when running the test and it is printing. So it is not related to the cache.

arinmodi:
```mvn clean verify``` generating some violations but if I run ```mvn verify``` without any code change than it is showing ```Build Success```.

Any Specific Reason for this ?

**Comment by arinmodi** [comment]
Github, generate report

**Comment by arinmodi** [comment]
Github, generate report

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java [hunk 1]
```diff
@@ -387,30 +390,270 @@ public int[] getAcceptableTokens() {
...
+     *
+     * @param comment as Javadoc Comment String
+     * @param commentDetails as Line Number and Column Number Details
+     *        commentDetails[0] as Start Line Number
+     *        commentDetails[1] as Start Column Number
+     *        commentDetails[2] as End Line Number
+     *        commentDetails[3] as End Column Number
+     * @return TextBlock of Javadoc comment
+     */
+    public TextBlock generateTextBlock(String comment, int... commentDetails) {
```
romani, line 510: If this is not avoidable, it is better to consider to upgrade Check to be javadoc check
https://checkstyle.org/writingjavadocchecks.html

romani (changes requested):
Question:

nrmancuso:
We lost contact with author, anybody is welcome to reuse code to solve issue

Issue #12520, opened 2022-12-12
Simplify JavadocStyleCheck: remove functionality for missing package-info Javadoc


rnveach:
Identified at https://github.com/checkstyle/checkstyle/issues/7416 item 1,

https://checkstyle.org/config_javadoc.html#JavadocStyle

We are attempting to trim back JavadocStyleCheck to help with a conversion to AbstractJavadocCheck. The first thing is to remove functionality that is identified as already being available in other checks.

> Ensure package-info.java has a Javadoc.
> is a duplicate of MissingJavadocPackage (Since 8.22).

This will remove this functionality and guide users on how to migrate their check.

========

**Migration Notes**

Users will have to add `MissingJavadocPackage` to their configuration, if it is not there already. No other options are needed.

Old Config Example:
````
<module name="Checker">
    <module name="TreeWalker">
<module name="JavadocStyle"/>
    </module>
</module>
````

Migrated Config Example:
````
<module name="Checker">
    <module name="TreeWalker">
<module name="JavadocStyle"/>
<module name="MissingJavadocPackage"/>
    </module>
</module>
````

https://www.baeldung.com/java-package-info
https://stackoverflow.com/questions/22095487/why-is-package-info-java-useful

nrmancuso:
Closed via #12521

PR #12521, opened 2022-12-12
Issue #12520: removes missing package-info Javadoc check in JavadocStyle

rnveach:
Issue #12520

Regression: http://rveach.no-ip.org/checkstyle/regression/319/

Please note, configuration was created to show that with the functionality removed is exactly the same as `MissingJavadocPackage`. The extra properties added were to allow regression to see them as the same to remove duplicates and trim the report.

http://rveach.no-ip.org/checkstyle/regression/319/checkstyle/index.html#A1
Is an example of a false negative as the old check was not identifying javadocs in the correct location.

http://rveach.no-ip.org/checkstyle/regression/319/checkstyle/index.html#A2
Is an example of a false positive as the new check only verifies the package has a styled javadoc. This was only being picked up by JavadocStyle as it saw there was no javadoc and it was in a package-info file, which it probably shouldn't have been doing. Basically the check if looking at all token types in the file even if they weren't a package definition. There are other "missing" checks which should handle this case, so there should be nothing lost in the end.

-----

property still in use by other Checks
```
~/java/github/romani/checkstyle [master|✔] 
$ ag -Q "javadoc.missing\""
src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java
308:    public static final String MSG_JAVADOC_MISSING = "javadoc.missing";

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
184:    public static final String MSG_JAVADOC_MISSING = "javadoc.missing";

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/MissingJavadocTypeCheck.java
191:    public static final String MSG_JAVADOC_MISSING = "javadoc.missing";

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
236:    public static final String MSG_SUMMARY_JAVADOC_MISSING = "summary.javaDoc.missing";

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/MissingJavadocMethodCheck.java
270:    public static final String MSG_JAVADOC_MISSING = "javadoc.missing";

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/MissingJavadocPackageCheck.java
80:    public static final String MSG_PKG_JAVADOC_MISSING = "package.javadoc.missing";

```


src/test/java/com/puppycrawl/tools/checkstyle/utils/CheckUtilTest.java
```diff
@@ -443,6 +443,16 @@ public void testParseClassNames() {
             .isEqualTo(expected);
     }
 
+    @Test
+    public void testIsPackageInfo() {
+        assertWithMessage("Result is expected")
+            .that(CheckUtil.isPackageInfo("Bland.java"))
+            .isEqualTo(false);
```
romani, line 450: can we cover this by Inputs ?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocstyle/bothfiles/package-info.java
```diff
@@ -13,4 +13,16 @@
 
 */
 
-package com.puppycrawl.tools.checkstyle.checks.javadoc.javadocstyle.bothfiles; // violation
+/** Example */ // violation
```
romani, line 16: please add violation message.

we are changing input file by structure (extra comment).

romani (changes requested):
https://checkstyle.org/config_javadoc.html#MissingJavadocPackage

>Checks for missing Javadoc comments in package-info.java files. 

we need to update description to mention that this Check does only package javadoc. If any type are declared in such file in addition, they will not be validated.

But in this case validation of other types in package-info.java should be on  JavadocStyle. can we keep this easily ? or we can create separate issue to cover this very rare usage of code.

items:

src/test/java/com/puppycrawl/tools/checkstyle/utils/CheckUtilTest.java
rnveach: We can, but as the test shows, it doesn't matter. The contents of the file does not matter as this method doesn't read it.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocstyle/bothfiles/package-info.java
rnveach: Done.

rnveach:
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/16604/workflows/083adf23-1c23-4b3c-92af-f2525931c94b/jobs/216874
CircleCI is having issues.
> error checking out branch: object not found

I already repushed and rebased on master.

src/test/java/com/puppycrawl/tools/checkstyle/utils/CheckUtilTest.java
romani: So we can make input file that has only empty class, so it will be clear that it doesn't matter.
But we do all by Inputs as much as possible.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocstyle/bothfiles/package-info.java
```diff
@@ -13,4 +13,16 @@
 
 */
 
-package com.puppycrawl.tools.checkstyle.checks.javadoc.javadocstyle.bothfiles; // violation
+/** Example */ // violation 'First sentence should end with a period.'
+@Example
+package com.puppycrawl.tools.checkstyle.checks.javadoc.javadocstyle.bothfiles;
```
romani, line 18: Why we changing input ?
It was only only one line, no comment, but it becomes a way different

romani (changes requested):
Items

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocstyle/bothfiles/package-info.java
rnveach: Reverted and added new test.

src/test/java/com/puppycrawl/tools/checkstyle/utils/CheckUtilTest.java
rnveach: Done.

rnveach:
@nrmancuso ping
https://github.com/checkstyle/checkstyle/actions/runs/3742296373/jobs/6353051245#step:7:15
> removed call to com/puppycrawl/tools/checkstyle/grammar/java/JavaLanguageLexer::removeErrorListeners

I remember an issue of this.

src/test/java/com/puppycrawl/tools/checkstyle/utils/CheckUtilTest.java
rnveach: Made non UT test.

romani (approved):
ok to  merge

src/test/java/com/puppycrawl/tools/checkstyle/utils/CheckUtilTest.java
romani: thanks a lot.

nrmancuso:
Github, generate site

nrmancuso:
rebased on latest master

nrmancuso:
Github, generate site

nrmancuso (approved):

nrmancuso:
Restarted pitest-common, looks like random failure.

nrmancuso:
Al CI failures are known (Travis and XML issues)

PR #12523, opened 2022-12-13
Issue #12522: Empty ErrorStream in checker-framework.groovy to prevent buffer exhaustion and external processes from blocking

### LINKED ISSUE #12522: Empty ErrorStream in `checker-framework.groovy` to prevent buffer exhaustion and external processes from blocking
Currently, `checker-framework.groovy` only empties the standard output stream and ignores the error stream. 
https://github.com/checkstyle/checkstyle/blob/f0633dcf0c0b4430ab6ded44bdd6aa3327b31039/.ci/checker-framework.groovy#L116-L121

This can cause the error stream buffer to fill completely which can block external processes. 

A better approach would be to redirect the error stream to prevent buffer from filling completely.

Source: https://wiki.sei.cmu.edu/confluence/display/java/FIO07-J.+Do+not+let+external+processes+block+on+IO+buffers

Vyom-Yadav:
Resolves #12522

### DIFF
--- a/.ci/checker-framework.groovy
+++ b/.ci/checker-framework.groovy
@@ -113,12 +113,24 @@ private static int checkCheckerFrameworkReport(final String profile) {
 private static List<List<String>> getCheckerFrameworkErrors(final String profile) {
     final List<String> checkerFrameworkLines = new ArrayList<>()
     final String command = "mvn -e --no-transfer-progress clean compile -P${profile}"
-    final Process process = getOsSpecificCmd(command).execute()
-    process.in.eachLine { final line ->
-        checkerFrameworkLines.add(line)
-        println(line)
+    final ProcessBuilder processBuilder = new ProcessBuilder(getOsSpecificCmd(command).split(' '))
+    processBuilder.redirectErrorStream(true)
+    final Process process = processBuilder.start()
+
+    BufferedReader reader = null
+    try {
+        reader = new BufferedReader(new InputStreamReader(process.inputStream))
+        String lineFromReader = reader.readLine()
+        while (lineFromReader != null) {
+            println(lineFromReader)
+            checkerFrameworkLines.add(lineFromReader)
+            lineFromReader = reader.readLine()
+        }
+        process.waitFor()
+    } finally {
+        reader.close()
     }
-    process.waitFor()
+
     final List<List<String>> checkerFrameworkErrors = new ArrayList<>()
     for (int index = 0; index < checkerFrameworkLines.size(); index++) {
         final String line = checkerFrameworkLines.get(index)


.ci/checker-framework.groovy
```diff
@@ -113,10 +113,16 @@ private static int checkCheckerFrameworkReport(final String profile) {
...
     final String command = "mvn -e --no-transfer-progress clean compile -P${profile}"
-    final Process process = getOsSpecificCmd(command).execute()
-    process.in.eachLine { final line ->
-        checkerFrameworkLines.add(line)
-        println(line)
+    final ProcessBuilder processBuilder = new ProcessBuilder(getOsSpecificCmd(command).split(' '))
+    processBuilder.redirectErrorStream(true)
+    final Process process = processBuilder.start()
+
+    final BufferedReader reader = new BufferedReader(new InputStreamReader(process.inputStream))
```
nrmancuso, line 120: Do we need to close this when we are done?

.ci/checker-framework.groovy
```diff
@@ -113,10 +113,16 @@ private static int checkCheckerFrameworkReport(final String profile) {
 private static List<List<String>> getCheckerFrameworkErrors(final String profile) {
     final List<String> checkerFrameworkLines = new ArrayList<>()
     final String command = "mvn -e --no-transfer-progress clean compile -P${profile}"
-    final Process process = getOsSpecificCmd(command).execute()
-    process.in.eachLine { final line ->
-        checkerFrameworkLines.add(line)
-        println(line)
+    final ProcessBuilder processBuilder = new ProcessBuilder(getOsSpecificCmd(command).split(' '))
+    processBuilder.redirectErrorStream(true)
```
nrmancuso, line 117: Will this obfuscate errors produced by maven during command execution?

nrmancuso (changes requested):
Items:

.ci/checker-framework.groovy
Vyom-Yadav: I don't think currently maven by default writes anything to `System.err`, all the info, error, and warning logs are written to `System.out`. 

See the [default SLF4J configuration for Maven ](https://github.com/apache/maven/blob/master/apache-maven/src/assembly/maven/conf/logging/simplelogger.properties). Output may go to `System.err` if changed to:
```properties
org.slf4j.simpleLogger.logFile=System.err
```

This makes me question whether this PR is required or not, but I will still go with this patch as this is a good practice, we should empty the error stream.

More details about maven logging at: https://maven.apache.org/maven-logging.html

nrmancuso (approved):
Restarted CircleCI

romani (approved):
ok to merge

.ci/checker-framework.groovy
```diff
@@ -113,12 +113,20 @@ private static int checkCheckerFrameworkReport(final String profile) {
...
+    final Process process = processBuilder.start()
+
+    final BufferedReader reader = new BufferedReader(new InputStreamReader(process.inputStream))
```
rnveach, line 120: Shouldn't this be part of a try with resource?

.ci/checker-framework.groovy
```diff
@@ -113,12 +113,20 @@ private static int checkCheckerFrameworkReport(final String profile) {
...
+    String lineFromReader = reader.readLine()
+    while (lineFromReader != null) {
+        println(lineFromReader)
+        checkerFrameworkLines.add(lineFromReader)
```
rnveach, line 124: Maybe not related to this PR, but why do we need to add everything to an array and not process it in real time? It doesn't seem like we re-use old indexes from the array.

rnveach (changes requested):

.ci/checker-framework.groovy
Vyom-Yadav: try with resources is not supported at this language level, I used try-finally, checked exception can be treated like unchecked in groovy, automatically are added to method signature.

.ci/checker-framework.groovy
Vyom-Yadav: Yes, it can be done with an additional boolean control variable in the same loop, let's keep it for a separate PR.

.ci/checker-framework.groovy
```diff
@@ -113,12 +113,24 @@ private static int checkCheckerFrameworkReport(final String profile) {
...
+        reader = new BufferedReader(new InputStreamReader(process.inputStream))
+        String lineFromReader = reader.readLine()
+        while (lineFromReader != null) {
+            println(lineFromReader)
+            checkerFrameworkLines.add(lineFromReader)
+            lineFromReader = reader.readLine()
+        }
+        process.waitFor()
+    } finally {
+        reader.close()
```
rnveach, line 131: For writting code this way, you need to verify the reader isn't null, otherwise you will get a NPE.

rnveach (changes requested):
Last

.ci/checker-framework.groovy
rnveach: Then please work on this. I am marking this resolved for now.

.ci/checker-framework.groovy
Vyom-Yadav: I thought about this but why would it be null? None of the constructors throw an exception.

Vyom-Yadav: @rnveach Ping, let's get this done.

rnveach (approved):

PR #12531, opened 2022-12-14
Issue #12515: Support the latest pattern matching syntax in switch labels

nrmancuso:
#12515

Looks like we actually lucked out with this change, and can keep backwards compatibility.

It will be easier to do this update before update at https://github.com/checkstyle/checkstyle/pull/12516, the inverse will make testing more difficult.

----------------------------------

### Check Regression Reports

https://nrmancuso.github.io/issue-12512/reports_logs/issue-12515_check_diff_reports_2022_12_22/diff_checks-only-javadoc-error/index.html

- Symmetrical parsing errors only (difference in line number)

https://nrmancuso.github.io/issue-12512/reports_logs/issue-12515_check_diff_reports_2022_12_22/diff_part1/index.html

- clean

https://nrmancuso.github.io/issue-12512/reports_logs/issue-12515_check_diff_reports_2022_12_22/diff_part2/index.html

- clean

https://nrmancuso.github.io/issue-12512/reports_logs/issue-12515_check_diff_reports_2022_12_22/diff_part3/index.html

- clean

https://nrmancuso.github.io/issue-12512/reports_logs/issue-12515_check_diff_reports_2022_12_22/diff_part4/index.html

- clean

https://nrmancuso.github.io/issue-12512/reports_logs/issue-12515_check_diff_reports_2022_12_22/diff_part5/index.html

- clean

https://nrmancuso.github.io/issue-12512/reports_logs/issue-12515_check_diff_reports_2022_12_22/diff_sevntu-check-regression_part_1/index.html

- clean 

https://nrmancuso.github.io/issue-12512/reports_logs/issue-12515_check_diff_reports_2022_12_22/diff_sevntu-check-regression_part_2/index.html

 - clean

-------------------
### AST Regression Report

https://nrmancuso.github.io/issue-12512/reports_logs/diff/index.html

- Symmetrical parsing errors only (difference in line number)

PR #12539, opened 2022-12-17
Issue #11163: Allow permanent suppression on AST regression test inputs

nrmancuso:
Part of #11163 , noticed at https://github.com/checkstyle/checkstyle/pull/12531#discussion_r1049074314


romani:
https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/checks/naming/abbreviationaswordinname/InputAbbreviationAsWordInNameType6.java

This is just another big input file, we can split it into two and it will be easier to debug from Junit test method

romani (changes requested):
See above

nrmancuso:
I am closing this after some discussion with @romani , we have agreed that it is good to limit file length of AST input too (not just BDD input).

Issue #12557, opened 2022-12-30
Restore pitest.sh to ease operation on local


romani:
We lost pitest.sh files that was very convenient to use on local to validate pitest.

Right now it is two steps in Github action and users need to know how to craft command line.

we need to restore our easy ability as it was before `./.ci/pitest.sh ant` and it does all required.

romani:
@nrmancuso , 
https://github.com/checkstyle/checkstyle/blob/069905ab1bdd6a866a6b2c1090c192e82e83100c/.github/workflows/pitest.yml#L61-L67

do we have any reasons to keep this steps separately ?

Vyom-Yadav:
Do we want the complete file back? A short simple script to run both commands with user input will do the job.

romani:
No restoration of what was before. We need new wrapper that execute mvn and groovy sequentially.

Vyom-Yadav:
On it.

nrmancuso:
> @nrmancuso ,
> 
> https://github.com/checkstyle/checkstyle/blob/069905ab1bdd6a866a6b2c1090c192e82e83100c/.github/workflows/pitest.yml#L61-L67
> 
> do we have any reasons to keep this steps separately ? keeping them as one step will do same in CI and ease users life on local to run pitest.

No need to separate them, we can just replace with one step that uses the new script.

romani:
PR is sent

nrmancuso:
Completed via #12561

PR #12561, opened 2022-12-31
Issue #12557: Restore pitest.sh to ease operation on local

romani:
Issue #12557

### DIFF
--- a/.ci/pitest.sh
+++ b/.ci/pitest.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Attention, there is no "-x" to avoid problems on CircleCI
+set -e
+
+function list_profiles() {
+  POM_PATH="$(dirname "${0}")/../pom.xml"
+  cat "$POM_PATH" | sed -n -e 's/^.*<id>\(pitest-[^<]\+\)<\/id>.*$/\1/p' | sort
+}
+
+case $1 in
+
+--list)
+  echo "Supported profiles:"
+  list_profiles "${0}"
+  ;;
+
+*)
+  PROFILES=$(list_profiles "${0}");
+
+  if [[ $(echo "$PROFILES" | grep -w -- "${1}" | cat) != "" ]]; then
+    mvn -e --no-transfer-progress -P"$1" clean test-compile org.pitest:pitest-maven:mutationCoverage
+    if [[ "$2" != "" ]]; then
+      groovy ./.ci/pitest-survival-check-xml.groovy "$1" "$2"
+    else
+      groovy ./.ci/pitest-survival-check-xml.groovy "$1"
+    fi
+  else
+    echo "Unexpected argument: $*"
+    echo "Usage $0 <profile> <extra argument for pitest-survival-check-xml.groovy>"
+    echo "To see the full list of supported profiles run '$0 --list'"
+    false
+  fi
+  ;;
+
+esac
--- a/.github/workflows/pitest.yml
+++ b/.github/workflows/pitest.yml
@@ -60,11 +60,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Generate ${{ matrix.profile }} report
         run: |
-          mvn -e --no-transfer-progress -P"pitest-${{ matrix.profile }}" clean test-compile \
-          org.pitest:pitest-maven:mutationCoverage
-      - name: Check ${{ matrix.profile }} report
-        run: |
-          groovy ./.ci/pitest-survival-check-xml.groovy pitest-${{ matrix.profile }}
+          ./.ci/pitest.sh "pitest-${{ matrix.profile }}" --generate-suppression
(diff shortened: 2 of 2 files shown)


.ci/pitest.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+# Attention, there is no "-x" to avoid problems on CircleCI
+set -e
+
+case $1 in
+
+pitest-annotation \
+|pitest-ant \
+|pitest-api \
+|pitest-blocks \
```
Vyom-Yadav, line 10: Can these not be hard coded and derived from `pom.xml`

.ci/pitest.sh
```diff
@@ -0,0 +1,60 @@
...
+#   # post validation is skipped, we do not test gui throughly
+#   ;;
+
+--list)
+  echo "Supported profiles:"
+  pomXmlPath="$(dirname "${0}")/../pom.xml"
+  sed -n -e 's/^.*<id>\(pitest-[^<]\+\)<\/id>.*$/\1/p' < "${pomXmlPath}" | sort
+  ;;
+
+*)
```
Vyom-Yadav, line 51: Why have separate logic for this? The groovy script has `--list` functionality.

Vyom-Yadav (changes requested):
The script has a `-g` flag too, please add that too.

.ci/pitest.sh
romani: it has bad output:
```
$ groovy ./.ci/pitest-survival-check-xml.groovy --list
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.reflection.CachedClass$3$1 (file:/opt/groovy/groovy-2.4.7/lib/groovy-2.4.7.jar) to method java.lang.Object.finalize()
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.reflection.CachedClass$3$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
Supported profiles:
....
```

better to not bother user with all this noise.

.ci/pitest.sh
romani: I made bash more smart.

.ci/pitest.sh
```diff
@@ -0,0 +1,31 @@
...
+
+--list)
+  echo "Supported profiles:"
+  list_profiles "${0}"
+  ;;
+
+*)
+  PROFILES=$(list_profiles "${0}");
+
+  if [[ $(echo $PROFILES | grep -w -- ${1} | cat) != "" ]]; then
```
romani, line 20: read 
 `--` - https://stackoverflow.com/a/2427987
`-w` - https://stackoverflow.com/a/8063284

Vyom-Yadav: Isn't 
```bash
if [[ $(echo "$PROFILES" | grep -w "${1}") != "" ]]; then
```
sufficient?

`--` is not required as `$1` would be `pitest-XYZ` not something like `--pitest-XYZ`, also `cat` can also be omitted, it is already printed to output stream.

Vyom-Yadav:
Please address https://github.com/checkstyle/checkstyle/pull/12561#pullrequestreview-1233328774 too

.ci/pitest.sh
Vyom-Yadav: Resolved

.ci/pitest.sh
Vyom-Yadav: Resolved

.ci/pitest.sh
romani: Without `--` , `--lis` is failing in weird way.

Cat is good to avoid failure is grep found nothing.

Vyom-Yadav: Good points, resolved.

romani:
I hope I fixed last violation in shell script, @Vyom-Yadav , please finish review.

Vyom-Yadav:
Please add the `[-g | --generate-suppression]` flag, from the actual script:
```
Usage groovy ./.ci/pitest-survival-check-xml.groovy [profile] [-g | --generate-suppression]
To see the full list of supported profiles run
groovy ./.ci/ pitest-survival-check-xml.groovy --list
```

romani:
I don't see it in GitHub action. Am I missing something?

romani:
Script will be reused in both usages, did skip g on good reason or just missed?

Vyom-Yadav:
> Script will be reused in both usages, did skip g on good reason or just missed?

Yes, it will be used for both, the `-g` flag is pretty helpful when you just want the suppressions.

romani:
done

nrmancuso (approved):

nrmancuso:
CI failure is in travis, it is well known.

PR #12563, opened 2022-12-31
infra: allow trigger action manually after version bump

romani:
Release Notes action was skipped on commit of version bump 

https://github.com/checkstyle/checkstyle/actions/workflows/releasenotes-gen.yml

same as: https://github.com/checkstyle/contribution/commit/6c1e4973f94f63fbd70c4fd749dcbe19e3b08cd1#diff-3352b8e68cb99ef1d26fdb226a4caf7e9e5b524dee8dda727210453eec3d9b0dR10

### DIFF
--- a/.github/workflows/releasenotes-gen.yml
+++ b/.github/workflows/releasenotes-gen.yml
@@ -1,6 +1,7 @@
 name: "Generate release notes"
 
 on:
+  workflow_dispatch: null
   push:
     branches:
       - master


.github/workflows/releasenotes-gen.yml
```diff
@@ -1,6 +1,7 @@
 name: "Generate release notes"
 
 on:
+  workflow_dispatch: null
```
nrmancuso, line 4: Please specify branch here, like:
```
workflow_dispatch:
  branches: [ master ]

```
If we only run action on push to master, we should only dispatch from master.

nrmancuso (changes requested):
Item:

.github/workflows/releasenotes-gen.yml
romani: done

nrmancuso: Same as https://github.com/checkstyle/checkstyle/pull/12580#discussion_r1061487465

nrmancuso (approved):

PR #12576, opened 2023-01-02
supplemental: contribution#273 converts diff.groovy to use Checkstyle's CLI

rnveach:
Issue https://github.com/checkstyle/contribution/issues/273
https://github.com/checkstyle/contribution/pull/747#issuecomment-1368682775

**Note:** `export MAVEN_OPTS="-Xmx2048m"` is now pointless, seeing if we need to convert this to CLI option.

### DIFF
--- a/.ci/no-exception-test.sh
+++ b/.ci/no-exception-test.sh
@@ -9,7 +9,10 @@ guava-with-google-checks)
   CS_POM_VERSION="$(getCheckstylePomVersion)"
   BRANCH=$(git rev-parse --abbrev-ref HEAD)
   echo CS_version: "$CS_POM_VERSION"
-  checkout_from https://github.com/checkstyle/contribution
+  checkout_from https://github.com/rnveach/contribution
+  cd .ci-temp/contribution
+  git checkout issue_273
+  cd ../..
   cd .ci-temp/contribution/checkstyle-tester
   sed -i.'' 's/^guava/#guava/' projects-to-test-on.properties
   sed -i.'' 's/#guava|/guava|/' projects-to-test-on.properties
@@ -17,10 +20,10 @@ guava-with-google-checks)
   cp src/main/resources/google_checks.xml .ci-temp/google_checks.xml
   sed -i.'' 's/warning/ignore/' .ci-temp/google_checks.xml
   cd .ci-temp/contribution/checkstyle-tester
-  export MAVEN_OPTS="-Xmx2048m"
+  export JDK_JAVA_OPTIONS="-Xmx2048m"
   groovy ./diff.groovy --listOfProjects projects-to-test-on.properties \
       --patchConfig ../../google_checks.xml \
-      --mode single --allowExcludes -xm "-Dcheckstyle.failsOnError=false" \
+      --mode single --allowExcludes --continueOnError \
       -p "$BRANCH" -r ../../..
   cd ../..
   removeFolderWithProtectedFiles contribution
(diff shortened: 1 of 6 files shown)


.ci/no-exception-test.sh
```diff
@@ -20,7 +20,7 @@ guava-with-google-checks)
   export MAVEN_OPTS="-Xmx2048m"
   groovy ./diff.groovy --listOfProjects projects-to-test-on.properties \
       --patchConfig ../../google_checks.xml \
-      --mode single --allowExcludes -xm "-Dcheckstyle.failsOnError=false" \
```
nrmancuso, line 23: Let's use long options, i.e. `--continueOnError`

rnveach:
We need to rework suppressions, which are no longer valid.

Example:
> repositories/hibernate-orm/**/hibernate-orm/documentation/**/*
> Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: IllegalStateException occurred while parsing file /root/project/.ci-temp/contribution/checkstyle-tester/repositories/hibernate-orm/documentation/src/main/docbook/devguide/en-US/chapters/events/extras/JpaCallbacksExample.java.

Now needs to be `repositories/hibernate-orm/documentation/`.

.ci/no-exception-test.sh
rnveach: Done.

rnveach:
no-exception-cassandra-storm-tapestry-javadoc was missed in this repo.

no-exception-checkstyle-sevntu and no-exception-checkstyle-sevntu-javadoc seems either like a CLI parsing issue or we are actually excluding all files from this run making regression pointless.

I am also not seeing why `continueOnError` is not working for it.

rnveach:
CI is passing now.

rnveach:
> export MAVEN_OPTS="-Xmx2048m" 

All of CI passed without this. If wanted, I can make this a non-maven option so local runs know how memory they need.

Let me know.

nrmancuso:
>Note: export MAVEN_OPTS="-Xmx2048m" is now pointless, seeing if we need to convert this to CLI option.

We can still pass these args via environment variable like `JDK_JAVA_OPTIONS="-Xmx2048m"`, I would prefer to avoid another diff.groovy option.

rnveach:
Why wouldn't we use the existing `extraRegressionOptions` ?

nrmancuso:
>to convert this to CLI option.

This sounded like a new CLI option (picocli) to me. Either using `xo` or environment variable is good with me.

rnveach:
`xo` is passed to the CLI after the program name. It will interpret this as an argument to Checkstyle and not to Java itself. Maybe `xo` should be arguments to Java itself?

nrmancuso:
I think it is best to have xo be checkstyle (picocli) options, and let us use environment variables for Java args. This way we have flexibility.

rnveach:
Updated to use JDK_JAVA_OPTIONS.

rnveach:
Rebased.

config/projects-to-test/openjdk-17-projects-to-test-on.config
```diff
@@ -5,4 +5,4 @@
 # This file has been created to alleviate confusion in the contribution
 # repository's projects-to-test-on.properties file; these projects
 # must be used with special configurations.
-openjdk17|git|https://github.com/openjdk/jdk17.git|master||
+openjdk17|git|https://github.com/openjdk/jdk17.git|master|/openjdk17/test/.*/MalformedDataRB_en.properties|
```
romani, line 8: please explain,
It is not clear why this was not required before.
if there is good reason , lets merge such changes sooner.

romani:
conceptually this PR is good. please assign it back to me when it is ready for merge.

config/projects-to-test/openjdk-17-projects-to-test-on.config
rnveach: Reasoning at https://github.com/checkstyle/contribution/pull/747#discussion_r1059674964

rnveach:
@romani It is blocked by https://github.com/checkstyle/contribution/pull/747 and must be almost merged at the same time as this depends on the other and without this merged https://github.com/checkstyle/contribution/pull/747 will break CS.

Issue #12587, opened 2023-01-05
Resolve surviving modifier mutations for pitest


rnveach:
Mutations:
https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-modifier-suppressions.xml

This issue will resolve the remaining suppression and remove the entire file. POM is already at 100% for this profile.

nrmancuso:
Completed via #12588

PR #12588, opened 2023-01-05
Issue #12587: resolves modifier pitest suppression

rnveach:
Issue #12587

https://checkstyle.org/config_modifier.html#RedundantModifier_Properties

Report: http://rveach.no-ip.org/checkstyle/regression/322/
No differences.

The old code is more a minor speed up in processing time and we don't lose much by removing it.

The method is `processMethods` and the AST given to it is always a `METHOD_DEF`. Usually we go 1 up on this before starting the loop, so it never starts with `METHOD_DEF`. However, even if it does start with `METHOD_DEF` all the cases inside the loop are not `METHOD_DEF`, so it goes straight to the last case which is the same the code we are removing, which is go up 1 parent.

Thus, its the same exact processing, except it takes a few more condition checks to realize this. It isn't a big jump in processing time, so I am ok to just remove the impacted code.

If we care about fixing this so processing time isn't increased (we are talking 4 extra condition checks comparing integers), then the only way is to either keep the suppression or I believe we can use reflection to create a test and kill the mutation the only way possible (direct calling the method).


romani (approved):
Ok to merge

nrmancuso (approved):

Issue #12592, opened 2023-01-06
Resolve surviving sizes mutations for pitest


rnveach:
Mutations:
https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-sizes-suppressions.xml

This issue will resolve the remaining suppression and remove the entire file. POM needs to be updated as well.

nrmancuso:
Completed via #12593

PR #12593, opened 2023-01-06
Issue #12592: resolves sizes pitest suppressions

rnveach:
Issue #12592

Fresh regression to come.


src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ExecutableStatementCountCheck.java
```diff
@@ -252,7 +252,6 @@ private static final class Context {
          */
         private Context(DetailAST ast) {
             this.ast = ast;
-            count = 0;
```
rnveach, line 255: The field is `int` and is `0` by default and is not `final`. So this assignment is unneceessary.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ExecutableStatementCountCheck.java
```diff
@@ -212,7 +212,7 @@ private void leaveContainerNode(DetailAST ast) {
      */
     private void visitSlist(DetailAST ast) {
         final DetailAST contextAST = context.getAST();
-        DetailAST parent = ast.getParent();
+        DetailAST parent = ast;
```
rnveach, line 215: Similar reasoning to https://github.com/checkstyle/checkstyle/pull/12588#issue-1520971449 . `isContainerNode` checks for specific types, and `SLIST` isn't one of them, so it just adds a bit more processing time to figure out it has to go up one anyways.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.java
```diff
@@ -272,14 +272,13 @@ private static int countUsedLines(DetailAST ast) {
         final Deque<DetailAST> nodes = new ArrayDeque<>();
         nodes.add(ast);
         final BitSet usedLines = new BitSet();
-        final int startLineNo = ast.getLineNo();
```
rnveach, line 275: This methods counts lines used in method. This logic returns the `cardinality` of the `BitSet` which is just how many lines were used. It doesn't matter what the starting line is as long as count the lines correctly (IE: Lines 17 and 18 is still 2 lines, just like lines 1 and 2).

With this `startLineNo` variable set to `0` for this mutation, all the other changes are basic math on their removal. `X - 0` is the same as just `X`. So the whole variable can just be removed.

rnveach:
@nrmancuso @Vyom-Yadav 
> `<sourceFile>RecordComponentNumberCheck.java</sourceFile>`

It looks like we have an overlap in the profiles. `RecordComponentNumberCheck` is part of sizes but is also appearing in the misc profile.

pom.xml
```diff
@@ -2918,9 +2918,6 @@
                 <param>com.puppycrawl.tools.checkstyle.checks.NoCodeInFileCheck*</param>
                 <param>com.puppycrawl.tools.checkstyle.checks.OrderedPropertiesCheck*</param>
                 <param>com.puppycrawl.tools.checkstyle.checks.OuterTypeFilenameCheck*</param>
-                <param>
-                  com.puppycrawl.tools.checkstyle.checks.sizes.RecordComponentNumberCheck*
```
rnveach, line 2922: https://github.com/checkstyle/checkstyle/actions/runs/3851529734/jobs/6562806022#step:6:337

This is the only one I am seeing not directly in `com.puppycrawl.tools.checkstyle.checks`. I did not recheck the others listed.

src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/RecordComponentNumberCheckTest.java
```diff
@@ -153,4 +155,20 @@ public void testRecordComponentNumberPrivateModifier() throws Exception {
         verifyWithInlineConfigParser(
                 getNonCompilablePath("InputRecordComponentNumberPrivateModifier.java"), expected);
     }
+
+    @Test
+    public void testSetAccessModifiers() throws Exception {
```
romani, line 160: we need details on what this exactly is covering, code is changing it might be unrelated test soon.
We agreed to put javadoc explanation on reasoning of pure unit test.

may be copy to pitest message ?

romani (changes requested):
itms:

src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/RecordComponentNumberCheckTest.java
romani: should we create util method for clone to let us ease skip/suppression of mutation on it:
```
    <description>replaced call to java/util/Arrays::copyOf with argument</description>
    <lineContent>Arrays.copyOf(accessModifiers, accessModifiers.length);</lineContent>
```

we do bunch of copy of content for over-security and fanatism of best practices. Such tests has no quality value, so we can make a code that should be skipped by pitest.

romani: or it will be just another junit method in all Test classes that check that clone is done for not simple type, as we have test methods for tokens.

rnveach: Done.

rnveach: > should we create util method for clone to let us ease skip/suppression of mutation on it:

A util for production code, will end up in the same situation as `Arrays.copyOf` is also a util method.
A util for test code seems fine to me, but it is really not a lot of code involved

romani: Please rename to `testCloneInAccessModifiersProperty()`

rnveach: Done.

romani (approved):
Ok to merge.

I am not resolving review items to make it readable for reviews.
If we have more cases like this, we need to consider to create one util method and let it be skipped by pitest at all.

src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/OuterTypeNumberCheckTest.java
```diff
@@ -104,4 +110,37 @@ public void testWithRecords() throws Exception {
...
 
+    /**
+     * Checks if the private field {@code currentDepth} and {@code outerNum} is
+     * properly cleared during the start of processing the next file in the
+     * check as they are file specific values.
+     *
+     * @throws Exception if there is an error.
+     */
+    @Test
+    public void testClearState() throws Exception {
```
nrmancuso, line 121: I wonder... could https://github.com/checkstyle/checkstyle/blob/33c291e2dc1f846362902ed2906d84acb3777b74/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java#L251 kill these mutations by sequential execution of check on files?

nrmancuso (changes requested):
Item:

src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/OuterTypeNumberCheckTest.java
rnveach: No. It has to be the same check instance. BDD, like the verify methods, recreate the entire config each run.

nrmancuso (approved):

Issue #12594, opened 2023-01-06
Resolve surviving pitest-java-ast-visitor mutations for pitest


rnveach:
Mutations:
https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-java-ast-visitor-suppressions.xml

This issue will resolve the remaining suppression and remove the entire file. POM does not need to be updated.

rnveach:
Issue is fixed

PR #12595, opened 2023-01-06
Issue #12594: resolves java-ast-visitor pitest suppressions

rnveach:
Issue #12594

ANTLR regression to come.


src/main/java/com/puppycrawl/tools/checkstyle/JavaAstVisitor.java
```diff
@@ -558,9 +558,8 @@ public DetailAstImpl visitClassOrInterfaceType(
         for (ParserRuleContext extendedContext : ctx.extended) {
             final DetailAstImpl dot = create(extendedContext.start);
             DetailAstPair.makeAstRoot(currentAST, dot);
-            final List<ParseTree> childList = extendedContext
-                    .children.subList(1, extendedContext.children.size());
```
rnveach, line 562: The mutation is basically saying the `subList` doesn't need to be done. The only thing the `subList` is doing is dropping the first item.

The first item in one example is a `period` and a terminal node (from `InputAntlr4AstRegressionWeirdCtor`, code `final Map.Entry<Class<?>, Integer> entry`). The reason this `subList` isn't needed is because the node is a terminal node. When this terminal node is allowed to pass to `visit`, it goes through the specific and default handling of terminal nodes which is to return `null`. When `null` gets to `addAstChild` it basically just drops it, so this terminal node isn't added to the tree there. A few lines above in the visitor's code, you can see the dot is specifically being handled by the `create` method there. So we aren't dropping a tree node, thus there is no harm in removing this code.

Rewriting of the code is the result of the `subList` call not being needed, thus we don't need to create the `childList` variable.

src/main/java/com/puppycrawl/tools/checkstyle/JavaAstVisitor.java
```diff
@@ -1336,7 +1335,7 @@ public DetailAstImpl visitSuperExp(JavaLanguageParser.SuperExpContext ctx) {
             superSuffixParent = bop;
         }
         else {
-            DetailAstImpl firstChild = superSuffixParent.getFirstChild();
+            DetailAstImpl firstChild = superSuffixParent;
```
rnveach, line 1338: Similar reasoning to https://github.com/checkstyle/checkstyle/pull/12588#issue-1520971449 . The loop keeps going down through the children until it finds a `null` first child. This just adds 1 more loop processing time since it doesn't automatically go into the first child, however, this prevents a possible (but probably unlikely) NPE that `superSuffixParent`'s first child is `null` already.

rnveach:
Regression (when complete) will be located at http://rveach.no-ip.org/checkstyle/regression/325/

I will update this comment but I expect it to be empty or only have an exception stack trace difference.

config/pitest-suppressions/pitest-java-ast-visitor-suppressions.xml
```diff
@@ -1,74 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
nrmancuso, line 1: Why are we forced to kill eight mutations in one PR? It would be better to separate out mutations that are killed by test only to another PR.

For removal of code, we usually demand a single PR per mutation.

rnveach: It was just because they were simple fixes. I had to plan to split out more if I had to do more removals.

nrmancuso: Ok, changes seem simple enough. But, with this PR we have a new concept (I think): changing the visitor without changing grammar. When we make grammar + visitor updates, we run regression testing on all checks with all projects.  What do you think about this?

rnveach: I think it is pointless to run check regression in this case.

The most important thing with this type of change is to first ask did the tree change? My ANTLR regression proves that the tree did not change for atleast the files it scans in our repo. I consider our most complex repo in our projects to test on and when I am regression hunting, majority times now I find cases in our own repo.

If the tree has absolutely no changes, how will the checks be impacted to show a change in regression? I believe this is the first time we have done what I believe to be a no change to regression.

If you are thinking this is not enough, then maybe your real concern is there is not enough ANTLR regression since I ran just the 1 project? I can run it on another of your choosing, but please note it took 77 minutes to run ANTLR regression on this 1. Bigger repos like openjdk will take hours and lock me out until it completes, which I am unsure how long that will be as I avoid bigger repos. I can also run check regression if you want, but I think it won't show anything (besides exceptions).

Also to note, we are now 100% with mutation on AST which should mean there is nothing outside our repo for our ANTLR besides these small little data points of concern.

rnveach: Side note for the future, maybe we should consider an alternative like ANTLR hashcode regression and see if that can run faster than full ANTLR print regression since we are not printing out each node, one by one. This would have to be a special custom Check, which we have not introduced into regression yet.

Another factor in the slowdown is we do javadoc ANTLR as well in the same run, so maybe that should be separated between Java only and Java and Javadoc. We would only be able to do Javadoc only printing regression with another special custom Check.

romani (approved):

config/pitest-suppressions/pitest-java-ast-visitor-suppressions.xml
romani: we can run with and without to be safe. as it is not javadoc it will not be very slow, so activation of all Checks in default (or any config, for example our config) will be ok.

rnveach: I am attempting to run regression.

rnveach: Regression: http://rveach.no-ip.org/checkstyle/regression/332/

No differences.

nrmancuso: Thanks for check regression test.

nrmancuso (approved):

nrmancuso:
@rnveach travis even liked your PR, it is hard to argue with that :)

Issue #12601, opened 2023-01-07
Resolve surviving pitest-whitespace mutations for pitest


rnveach:
Mutations:
https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-whitespace-suppressions.xml

This issue will resolve the remaining suppressions and remove the entire file. POM does not need to be updated.

5-6 PRs are expected. I will close the issue when done.

rnveach:
Fix was merged

PR #12603, opened 2023-01-07
Issue #12601: removes code for mutation in EmptyLineSeparator

rnveach:
Issue #12601


src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
```diff
@@ -621,14 +621,12 @@ private List<Integer> getEmptyLines(DetailAST ast) {
      */
     private static List<Integer> getEmptyLinesToLog(List<Integer> emptyLines) {
         final List<Integer> emptyLinesToLog = new ArrayList<>();
-        if (emptyLines.size() >= 2) {
```
rnveach, line 624: I assume this became a surviving mutation with other kills.

This prevents lines with only 1 empty line to be excluded.

The way the code is written, 1 line will never get added to the result. `previousEmptyLineNo` is set to the value of that 1 line. Then the condition to add to the result is `previousEmptyLineNo + 1 == emptyLineNo` which is based on the same 1 line. It logically impossible to for the equality `X + 1 == X` to be true.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
```diff
@@ -621,14 +621,12 @@ private List<Integer> getEmptyLines(DetailAST ast) {
...
-            int previousEmptyLineNo = emptyLines.get(0);
```
rnveach, line 625: This follows the same logic as the previous one.

The for loop still investigates the 1st item and it is impossible to match it as I explained before to add to the result.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
```diff
@@ -619,16 +619,14 @@ private List<Integer> getEmptyLines(DetailAST ast) {
      * @param emptyLines list of empty lines.
      * @return list of empty lines to log.
      */
-    private static List<Integer> getEmptyLinesToLog(List<Integer> emptyLines) {
+    private static List<Integer> getEmptyLinesToLog(Iterable<Integer> emptyLines) {
```
rnveach, line 622: https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=12734&view=logs&j=eb841e9f-6e1e-5c29-3ad5-74ec2c658069&t=a9b761e6-b8d5-50cf-467a-54967189cab2&l=1262
PMD said this had to change.

rnveach:

rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/328/
No differences.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
nrmancuso: This link isn't working for me, is it related to removing the call to `size()` and being able to use an interface higher in the inheritance chain?

rnveach: > is it related to removing the call to size() and get() and being able to use an interface higher in the inheritance chain?

Yes, it is similar to if I use `ArrayList` as a type, but could downgrade it to `List` or even further as `Collection`.

romani (approved):
ok to merge

rnveach:
@nrmancuso Last one. With approval I will rebase and remove the file completely.

nrmancuso (approved):

PR #12604, opened 2023-01-07
Issue #12601: removes code for SingleSpaceSeparator

rnveach:
Issue #12601

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
@@ -179,7 +179,6 @@ public class SingleSpaceSeparatorCheck extends AbstractCheck {
      */
     private void visitEachToken(DetailAST node) {
         DetailAST currentNode = node;
-        final DetailAST parent = node.getParent();
 
         do {
             final int columnNo = currentNode.getColumnNo() - 1;
@@ -199,7 +198,7 @@ public class SingleSpaceSeparatorCheck extends AbstractCheck {
                 currentNode = currentNode.getFirstChild();
             }
             else {
-                while (currentNode.getNextSibling() == null && currentNode.getParent() != parent) {
+                while (currentNode.getNextSibling() == null && currentNode.getParent() != null) {
                     currentNode = currentNode.getParent();
                 }
                 currentNode = currentNode.getNextSibling();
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
```diff
@@ -179,7 +179,6 @@ public void beginTree(DetailAST rootAST) {
      */
     private void visitEachToken(DetailAST node) {
         DetailAST currentNode = node;
-        final DetailAST parent = node.getParent();
```
rnveach, line 182: `visitEachToken` is only called in `beginTree` on the root node. Well the parent of the root node is always `null` so this code is completely pointless and the variable is replace with just `null`.

rnveach:

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
nrmancuso: This is a good point; what do you think about migrating checks like this (that specify/allow no tokens/traverse the entire tree) to set COMPILATION_UNIT as the only token?

nrmancuso (approved):

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/SingleSpaceSeparatorCheck.java
rnveach: Looking at this again, it seems to me `visitEachToken` is trying to replicate `TreeWalker.processIter`. We can just set the check to stop on every node as an alternative, but it is slightly overkill. It doesn't need to stop on nodes that are not in the source file, but we don't maintain such a list and it probably wouldn't be reasonable to maintain one.

Outside of that, I think it is fine that `beginTree` was used here, even with `COMPILATION_UNIT`. I see Javadoc checks do what you discuss, and I always think it would be better if they could just use a `beginTree`, or like I suggested, find actual nodes to stop on.

rnveach: > We can just set the check to stop on every node as an alternative

I think this may not have been done due to comments. `TokenUtil.getAllTokenIds()` returns all nodes, even comments, however, we have the property `validateComments` to say if comments should be stopped at and the 2 must be in synch otherwise we get the exception:
> Check 'com.puppycrawl.tools.checkstyle.checks.whitespace.SingleSpaceSeparatorCheck' waits for comment type token ('BLOCK_COMMENT_BEGIN') and should override 'isCommentNodesRequired()' method to return 'true'

romani (approved):

PR #12605, opened 2023-01-07
Issue #12601: removes code for EmptyLineSeparatorCheck

rnveach:
Issue #12601


src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
```diff
@@ -677,7 +677,7 @@ else if (!hasEmptyLineAfter(ast)) {
      * @return true, if there is an element.
      */
     private static boolean isLineEmptyAfterPackage(DetailAST ast) {
-        DetailAST nextElement = ast.getNextSibling();
+        DetailAST nextElement = ast;
```
rnveach, line 680: Similar reasoning to https://github.com/checkstyle/checkstyle/pull/12588#issue-1520971449 . The code below will call `getNextSibling` on the AST and I could not kill the mutation with the conditions of the while loop, so it will just cost us a bit more processing time to get the same results.

rnveach:

nrmancuso (approved):

romani (approved):

PR #12607, opened 2023-01-07
Issue #12601: adds test for whitespace mutation 

rnveach:
Issue #12601

Edit by @nrmancuso to help find example in the future: Mutation killed by combination of XPath suppression and check

### DIFF
--- a/config/pitest-suppressions/pitest-whitespace-suppressions.xml
+++ b/config/pitest-suppressions/pitest-whitespace-suppressions.xml
@@ -73,15 +73,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>ParenPadCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck</mutatedClass>
-    <mutatedMethod>visitResourceSpecification</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::findFirstToken with receiver</description>
-    <lineContent>processLeft(ast.findFirstToken(TokenTypes.LPAREN));</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>SingleSpaceSeparatorCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.whitespace.SingleSpaceSeparatorCheck</mutatedClass>
     <mutatedMethod>visitEachToken</mutatedMethod>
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheck.java
```diff
@@ -298,7 +298,7 @@ private void visitTokenWithOptionalParentheses(DetailAST ast) {
      * @param ast the token to check.
      */
     private void visitResourceSpecification(DetailAST ast) {
-        processLeft(ast.findFirstToken(TokenTypes.LPAREN));
+        processLeft(ast);
```
rnveach, line 301: Regression found nothing. I assume this just adds a bit more processing to skip over the `LPAREN` and goes in it.

romani: please share result report of regression.

so Looks like pitest is prone to force us to remove performance optimizations, as it definitely not detected by tests.

It is interesting how far it will lead us.
I am ok to sacrifice performance in favor of quality, as it will give more value, but still fill it uncomfortable.

romani: strange method looks like not even ready to get any other token:
https://github.com/checkstyle/checkstyle/blob/0b82f9c27a9a06320e829ad287fd7e1b8171b53e/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/AbstractParenPadCheck.java#L83-L88

rnveach: > share result report of regression.

Regression is coming.

> I am ok to sacrifice performance in favor of quality

I am ok for only small performance. Bigger ones I am not in favor of.

> strange method looks like not even ready to get any other token

It doesn't do anything specific to paren. I am going to relook at this, I think this may be a case where we need weird/non-normal, multiline Java code.

rnveach: Ok, now it makes more sense. AST is `RESOURCE_SPECIFICATION`. It is a type that is created by us and not seen in the actual source. It's line/column number is based solely on the first child token. The first child token is `LPAREN` which is what the original code was looking for. Based on this, `RESOURCE_SPECIFICATION` and `LPAREN` have the same line/column position, so they are indistinguishable besides text and type.

`processLeft` does not really care about the AST. It doesn't look at text/type/siblings/children/parent. It just looks at the line/column and the line's text.

Now there is a caveat. The AST sent to `processLeft` is sent to the `log` method. Since we are changing the logging AST, this means it also changes our XPath path to the violation since it is dependent on the exact node. It **MAY** be possible to kill the mutation with XPath suppressions as that cares about the type, but I am not fully sure, even for creating one using our BDD.

@romani Do we have example tests where we have a check and xpath filter using our BDD?

romani: I see from code of Check, that it execute on ast that is not left paren, so javadoc might be outdated.

rnveach: Resolved.

rnveach:
No regression needed as a test was made.

nrmancuso (approved):

romani (approved):

Issue #11407, opened 2022-03-13
False Negative: SummaryJavadocCheck


Vyom-Yadav:
have read check documentation: https://checkstyle.sourceforge.io/config_javadoc.html#SummaryJavadoc
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Test.java

/var/tmp $ cat config.xml
```
```xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="TreeWalker">
        <module name="SummaryJavadoc"/>
    </module>
</module>
```
```
/var/tmp $ cat Test.java
```
```java
/**
 * This is 1 version
 */
public class Test {
}

/**
 * This is 1.0 version
 */
class Test2 {
}
```
```
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:1: First sentence of Javadoc is missing an ending period. [SummaryJavadoc]
Audit done.
Checkstyle ends with 1 errors.
```
---

**Describe what you expect in detail.**

```
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:1: First sentence of Javadoc is missing an ending period. [SummaryJavadoc]
[ERROR] /var/tmp/Test.java:7: First sentence of Javadoc is missing an ending period. [SummaryJavadoc]
Audit done.
Checkstyle ends with 2 errors.
```

The `.` in `1.0` is falsely considered as the ending period which causes this bug.

Vyom-Yadav:
On it.

nick-mancuso:
Please print out AST with javadoc for your example, and share it in a comment here.

romani:
Please share javadoc tool output on this, how html look like for this methods?

Vyom-Yadav:
AST:-
```
[...]

PR #12609, opened 2023-01-07
 Issue #11407: False Negative: SummaryJavadocCheck

Kevin222004:
Issue #11407: False Negative: SummaryJavadocCheck


Kevin222004:
@Vyom-Yadav please look at test cases https://github.com/checkstyle/checkstyle/blob/d0bb9bc8ef9fbcab25fbf8d8778e2cdeabd008d0/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/summaryjavadoc/InputSummaryJavadocPeriodAtEndJapanese.java added in Japanese. i guess it was expected

src/main/java/com/puppycrawl/tools/checkstyle/JavadocDetailNodeParser.java
```diff
@@ -473,7 +473,7 @@ private static String getNodeClassNameWithoutContext(ParseTree node) {
      * <tbody>
      * <thead>
      * <tfoot>
-     * }
+     * }.
```
Kevin222004, line 476: I will update this it is not at good position

src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java
```diff
@@ -270,7 +270,7 @@ public static Configuration loadConfiguration(String config,
     /**
      * Returns the module configurations from a specified input source.
      * Note that if the source does wrap an open byte or character
-     * stream, clients are required to close that stream by themselves
+     * stream, clients are required to close that stream by themselves.
```
rnveach, line 273: I disagree with this change if fixing the Check flagged this. I am not saying the change is completely invalid, but it doesn't seem tied to this issue.

https://checkstyle.org/config_javadoc.html#SummaryJavadoc_Description
> Check also violate Javadoc that does not contain first sentence....

The first sentence is "Returns the module configurations from a specified input source." and it ends with a period. The check should not have produced a violation here.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java
Kevin222004: ok got it. I have to just look for first sentence in the scope of this issue. 
current code is looking for https://github.com/checkstyle/checkstyle/blob/887aab9402e4e24b2e9c5d7b005e9383c889929e/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java#L340
period in whole summary so i have also make changes accordingly but i will update this as to see the period only first sentece

Kevin222004: @rnveach I am converting this pr as a draft right know i am not finding any proper fix for this according to current code 

i also dont understand why this method is name as https://github.com/checkstyle/checkstyle/blob/811983e0f89edf2419b9165ca60bb04a7c3c1ea6/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java#L718
`getFirstSentence`  it is not giving the first sentence at all it is just returning the string of whole javadoc

Vyom-Yadav: This method returns first sentence till `periodSuffix`. If you javadoc summary just contains one sentence then it will return the complete javadoc. Try this method with multiple sentences in your javadoc.

Kevin222004:
I am really sorry but i am closing this pr for know i will surely come back with some new fresh idea after working a lot on this their are still many confusion to me i have to study about this more.

rnveach:
@Kevin222004 You are free to close your own PRs if that is what you wish. There is no restriction preventing you from doing this.

Issue #12615, opened 2023-01-10
Resolve surviving pitest-blocks mutations for pitest


rnveach:
Mutations:
https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-blocks-suppressions.xml

This issue will resolve the remaining suppressions and remove the entire file. POM does need to be updated.

~Part of this issue is blocked by https://github.com/hcoles/pitest/issues/1137 (target pitest release 1.10.5) as those suppressions are basically false positives and cannot be resolved.~

nrmancuso:
Completed via #12618

PR #12618, opened 2023-01-10
Issue #12615: removes code for RightCurly

rnveach:
Issue #12615

This is one PR because everything is for the same check and will make things complicated to separate. Analysis is done from a clean branch.


rnveach:
Regression is coming and will be added here.

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -551,17 +551,13 @@ private static Details getDetails(DetailAST ast) {
             switch (ast.getType()) {
                 case TokenTypes.LITERAL_TRY:
                 case TokenTypes.LITERAL_CATCH:
-                case TokenTypes.LITERAL_FINALLY:
```
rnveach, line 554: This can be removed because the code for `getDetailsForTryCatchFinally` gives the same results as `getDetailsForOthers` for the `LITERAL_FINALLY` token. This is because `finally` must always have `{}`s and the constant tree structure.

````
        |      |  `--LITERAL_FINALLY -> finally [5:2]
        |      |      `--SLIST -> { [5:10]
...
        |      |          |--SEMI -> ; [6:9]
        |      |          `--RCURLY -> } [7:0]
````

````
        private static Details getDetailsForTryCatchFinally(DetailAST ast) {
            final DetailAST lcurly;
            DetailAST nextToken;
            final int tokenType = ast.getType();
            if (false) {
...
            }
            else {
                nextToken = ast.getNextSibling(); // == null
                lcurly = ast.getLastChild(); // == `{` (always)

            }

            final boolean shouldCheckLastRcurly;
            if (true) {
                shouldCheckLastRcurly = true;
                nextToken = getNextToken(ast); // is always what comes after `}`
            }
            else {
...
            }

            final DetailAST rcurly = lcurly.getLastChild(); // == `}`
            return new Details(lcurly, rcurly, nextToken, shouldCheckLastRcurly);
        }
````

````
        private static Details getDetailsForOthers(DetailAST ast) {
            DetailAST rcurly = null;
            final DetailAST lcurly;
            final int tokenType = ast.getType();
            if (false) {
...
            }
            else {
                lcurly = ast.findFirstToken(TokenTypes.SLIST); // == `{` (always)
                if (true) {
                    // SLIST could be absent if method is abstract
                    rcurly = lcurly.getLastChild(); // == `}` (always)
                }
            }
            return new Details(lcurly, rcurly, getNextToken(ast), true);
        }
````

In short, 1 is:
````
lcurly = ast.getLastChild();
shouldCheckLastRcurly = true;
nextToken = getNextToken(ast);
rcurly = lcurly.getLastChild();
````

2 is:
````
lcurly = ast.findFirstToken(TokenTypes.SLIST);
rcurly = lcurly.getLastChild();
nextToken = getNextToken(ast)
````

The only difference is lcurly (`ast.getLastChild()` versus `ast.findFirstToken(TokenTypes.SLIST)` which is equivalent by the tree.

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java [hunk 2]
```diff
@@ -551,17 +551,13 @@ private static Details getDetails(DetailAST ast) {
...
-                    details = getDetailsForTryCatchFinally(ast);
+                    details = getDetailsForTryCatch(ast);
                     break;
                 case TokenTypes.LITERAL_IF:
-                case TokenTypes.LITERAL_ELSE:
```
rnveach, line 558:
````
        private static Details getDetailsForIfElse(DetailAST ast) {
            final boolean shouldCheckLastRcurly;
            final DetailAST lcurly;
            DetailAST nextToken = ast.findFirstToken(TokenTypes.LITERAL_ELSE); // always `null`, `else` can't have `else` as child

            if (true) {
                shouldCheckLastRcurly = true;
                nextToken = getNextToken(ast);
                lcurly = ast.getLastChild();
            }
            else {
....
            }

            DetailAST rcurly = null;
            if (lcurly.getType() == TokenTypes.SLIST) {
                rcurly = lcurly.getLastChild();
            }
            return new Details(lcurly, rcurly, nextToken, shouldCheckLastRcurly);
        }
````

````
        private static Details getDetailsForOthers(DetailAST ast) {
            DetailAST rcurly = null;
            final DetailAST lcurly;
            final int tokenType = ast.getType();
            if (false) {
...
            }
            else {
                lcurly = ast.findFirstToken(TokenTypes.SLIST);
                if (lcurly != null) {
                    // SLIST could be absent if method is abstract
                    rcurly = lcurly.getLastChild();
                }
            }
            return new Details(lcurly, rcurly, getNextToken(ast), true);
        }
````

In short, 1 is:
````
shouldCheckLastRcurly = true;
nextToken = getNextToken(ast);
lcurly = ast.getLastChild();
rcurly is `null` or `lcurly.getLastChild()` if lcurly is an slist
````
2 is:
````
lcurly = ast.findFirstToken(TokenTypes.SLIST);
rcurly is `null` or `lcurly.getLastChild()` if lcurly found an slist (same as is an slist)
nextToken = getNextToken(ast);
````

The only difference is lcurly (ast.getLastChild() versus ast.findFirstToken(TokenTypes.SLIST) which is the difference between both being the same if the `else` has a `{` or one will have the next token and the other will have `null`. So it comes down to how we handle `lcurly` in the logic.
Example:
original   - else[75x8] - Details [rcurly=null, lcurly=;[75x12], nextToken=}[76x4], shouldCheckLastRcurly=true]
mutated  - else[75x8] - Details [rcurly=null, lcurly=null, nextToken=}[76x4], shouldCheckLastRcurly=true]

However in this case, it doesn't come down to `lcurly` but `rcurly`.
https://github.com/checkstyle/checkstyle/blob/54616f662130aa0ca5afe45f72d36cdefb78a781/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java#L308
We don't validate if there is no `rcurly`.

`rcurly` is always `null` if `lcurly` doesn't find an `SLIST`. `rculy` always has the `}` if `lcurly` does find the `SLIST`.

So in the end, they are both the same.

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java [hunk 3]
```diff
@@ -551,17 +551,13 @@ private static Details getDetails(DetailAST ast) {
...
-                    details = getDetailsForIfElse(ast);
+                    details = getDetailsForIf(ast);
                     break;
                 case TokenTypes.LITERAL_DO:
-                case TokenTypes.LITERAL_WHILE:
```
rnveach, line 562:
````
        private static Details getDetailsForLoops(DetailAST ast) {
            DetailAST rcurly = null;
            final DetailAST lcurly;
            final DetailAST nextToken;
            final int tokenType = ast.getType();
            final boolean shouldCheckLastRcurly;
            if (false) {
...
            }
            else {
                shouldCheckLastRcurly = true;
                lcurly = ast.findFirstToken(TokenTypes.SLIST);
                if (lcurly != null) {
                    // SLIST could be absent in code like "while(true);"
                    rcurly = lcurly.getLastChild();
                }
                nextToken = getNextToken(ast);
            }
            return new Details(lcurly, rcurly, nextToken, shouldCheckLastRcurly);
        }
````

````
        private static Details getDetailsForOthers(DetailAST ast) {
            DetailAST rcurly = null;
            final DetailAST lcurly;
            final int tokenType = ast.getType();
            if (false) {
...
            }
            else {
                lcurly = ast.findFirstToken(TokenTypes.SLIST);
                if (lcurly != null) {
                    // SLIST could be absent if method is abstract
                    rcurly = lcurly.getLastChild();
                }
            }
            return new Details(lcurly, rcurly, getNextToken(ast), true);
        }
````

If you look closely, the code remaining is the same, so there is no differences between the 2.

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java [hunk 4]
```diff
@@ -551,17 +551,13 @@ private static Details getDetails(DetailAST ast) {
...
-                case TokenTypes.LITERAL_FOR:
```
rnveach, line 563: Exact same as https://github.com/checkstyle/checkstyle/pull/12618#discussion_r1066076220 .

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -673,35 +669,19 @@ private static boolean isTokenWithNoChildSlist(int tokenType) {
...
-        private static Details getDetailsForLoops(DetailAST ast) {
+        private static Details getDetailsForDoLoops(DetailAST ast) {
+            final DetailAST lcurly = ast.findFirstToken(TokenTypes.SLIST);
+            final DetailAST nextToken = ast.findFirstToken(TokenTypes.DO_WHILE);
             DetailAST rcurly = null;
-            final DetailAST lcurly;
-            final DetailAST nextToken;
-            final int tokenType = ast.getType();
-            final boolean shouldCheckLastRcurly;
-            if (tokenType == TokenTypes.LITERAL_DO) {
```
rnveach, line 687: With the cascade of both at https://github.com/checkstyle/checkstyle/pull/12618#discussion_r1066077473 , this always become true and thus all this code can be reduced.

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -648,7 +644,7 @@ private static Details getDetailsForOthers(DetailAST ast) {
             final int tokenType = ast.getType();
             if (isTokenWithNoChildSlist(tokenType)) {
                 final DetailAST child = ast.getLastChild();
-                lcurly = child.getFirstChild();
+                lcurly = child;
```
rnveach, line 647: Tokens that enter here can only be in `TOKENS_WITH_NO_CHILD_SLIST` which is `TokenTypes.CLASS_DEF, TokenTypes.ENUM_DEF, TokenTypes.ANNOTATION_DEF, TokenTypes.INTERFACE_DEF, TokenTypes.RECORD_DEF`.

`child` is always the `OBJBLOCK` of these tokens since they are all definitions. Since these are objects, the first child is always the `{`.

Since `OBJBLOCK` is a token we create and is not defined in the source, it always has the same line/column as the first child. This AST is not sent the `log` and we don't look at the text/type of the AST in this check.

So for all purposes, the 2 are identical and results in no differences.

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -461,12 +461,12 @@ private static boolean skipDoubleBraceInstInit(Details details) {
     private static boolean isBlockAloneOnSingleLine(Details details) {
         DetailAST nextToken = details.nextToken;
 
-        while (nextToken != null && nextToken.getType() == TokenTypes.LITERAL_ELSE) {
-            nextToken = Details.getNextToken(nextToken);
+        if (nextToken != null && nextToken.getType() == TokenTypes.LITERAL_ELSE) {
+            nextToken = null;
```
rnveach, line 465: This is only for `ALONE_OR_SINGLELINE` and `TokenUtil.areOnSameLine(details.lcurly, details.rcurly)` must be `true` for this difference to matter.

If the token doesn't enter the loop, then it is never changed. So this literally only applies when the `nextToken` is an `LITERAL_ELSE`. So this looks like it comes down to the difference between `Details.getNextToken` and `null`.

It is actually all about `Details.getNextToken` and the loop conditions and the tree. `getNextToken` only looks at the next siblings or the next parent. `LITERAL_ELSE` has no siblings, so the parent only goes up the chain and grabs the next token after it, which can be whatever expression is next.

So what matters next is the condition on the return, `nextToken == null || !TokenUtil.areOnSameLine(details.rcurly, nextToken) || isRightcurlyFollowedBySemicolon(details)`. `nextToken == null` doesn't matter since that is what my mutated code is doing.

`!TokenUtil.areOnSameLine(details.rcurly, nextToken) || isRightcurlyFollowedBySemicolon(details)` is where I believe I found a test to kill this mutation.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTestAloneOrSingleline.java
```diff
@@ -244,4 +244,9 @@ public void codeAfterLastRightCurly() {
 
     interface Interface2
     { int i = 1; public void meth1(); }
+
+    private void foo28() {
+        if (true) { return; } else { } if (false) { // 2 violations
```
rnveach, line 249: This mutation kills https://github.com/checkstyle/checkstyle/pull/12618#discussion_r1066151940 .
It needs to be funky like this since it checks line positions.

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
rnveach: Should be killed now. Resolved.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTestAloneOrSingleline.java
```diff
@@ -244,4 +244,14 @@ public void codeAfterLastRightCurly() {
...
+        }
+    }
+
+    private void foo31() {
+        do { var2 ++; }
+        while (var2 < 15); while (var1 < 10) { var1++; }
```
rnveach, line 255: This mutation kills https://github.com/checkstyle/checkstyle/pull/12618#discussion_r1066187654 .
It needs to be funky like this since it checks line positions.

rnveach: https://github.com/checkstyle/checkstyle/actions/runs/3886458243/jobs/6631584122#step:6:416
This isn't working with this PR. I will have to relook at this.

rnveach: I was made a mistake in my testing and this doesn't help pitest.

romani (approved):
ok to merge

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTestAloneOrSingleline.java
nrmancuso: Good find :)

nrmancuso (approved):

Issue #12626, opened 2023-01-12
Resolve surviving pitest-coding-require-this-check mutations for pitest


rnveach:
Mutations:
https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-coding-require-this-check-suppressions.xml

This issue will resolve the remaining suppressions and remove the entire file. POM does need to be updated.

nrmancuso:
@rnveach which PR will have pom update?

rnveach:
@nrmancuso It depends which PR is the last to merge? I can't specify 100% when there are 2 open PRs. While there is a rounding issue with pitest, most likely, neither one will register 100% without the other.

rnveach:
Fixes were merged

PR #12614, opened 2023-01-10
Issue #12613: `pitest-survival-check-xml.groovy` should automatically update list of suppressions

Vyom-Yadav:
Resolves #12613
Resolves #12153

PR #12628, opened 2023-01-12
Issue #12626: removes code for mutations in RequireThis

rnveach:
Issue #12626

Regression and reasons to come.


src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -458,7 +458,6 @@ private void processIdent(DetailAST ast) {
         }
         switch (parentType) {
             case TokenTypes.ANNOTATION_MEMBER_VALUE_PAIR:
-            case TokenTypes.ANNOTATION:
```
rnveach, line 461: This case does nothing. The default does nothing if `checkFields` is `false`. So the only thing that matters here is when `checkFields` is `true`.

For a violation to show up, we need a frame returned from `getFieldWithoutThis` on the AST.

It seems to me we need an annotation that is also named after a field to confuse the logic. (Still investigating)

rnveach: > we need an annotation that is also named after a field to confuse the logic

This seems is what is needed, so I found a test to kill the mutation.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -534,7 +533,7 @@ private AbstractFrame getFieldWithoutThis(DetailAST ast, int parentType) {
      */
     private static boolean isInCompactConstructor(DetailAST ast) {
         boolean isInCompactCtor = false;
-        DetailAST parent = ast.getParent();
```
rnveach, line 537: This is only called through `processIdent`, so the AST is always an `IDENT`.

This is very similar to other things we have seen.

The loop below looks for some criteria and until it finds it it goes up the parentage. Well this mutation is removing going up the first parent. The only way we can say its needed is if the AST is what we are looking for, `COMPACT_CTOR_DEF` and we won't since the AST is always `IDENT`.

The AST is never `null` as seen from the original code, otherwise it would be an NPE on getting the parent.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -609,7 +608,7 @@ private static void collectDeclarations(Deque<AbstractFrame> frameStack, DetailA
             case TokenTypes.LITERAL_NEW:
                 if (isAnonymousClassDef(ast)) {
                     frameStack.addFirst(new AnonymousClassFrame(frame,
-                            ast.getFirstChild().toString()));
+                            ast.toString()));
```
rnveach, line 611: The token is `LITERAL_NEW`, so the mutation difference here is giving a frame the name `new` (from `LITERAL_NEW`) or giving what we wanted which was the name of the class being instantiated.

`AnonymousClassFrame` doesn't track the `AST` so it all depends on how it uses the name.

The only place it seems to be used is `logViolation`. It compares the name to the result of `getNearestClassFrameName` but it looks like it is getting the name from the same frame as defined as the "nearest".

It is comparing apples to apples, so it seems it may not care what the name is as long as it matches itself. Maybe we could confuse it with the same name multiple times, but this may require a deeper analysis.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -920,7 +905,7 @@ private boolean canBeReferencedFromStaticContext(DetailAST ident) {
      *         definition was not found.
      */
     private static DetailAST getCodeBlockDefinitionToken(DetailAST ident) {
-        DetailAST parent = ident.getParent();
```
rnveach, line 923: The AST given is always `IDENT`.

This is the same thing we have seen everywhere. We just do an extra `getParent` to get back to where we would be without this mutation.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -1215,7 +1200,7 @@ private String getNearestClassFrameName() {
      */
     private static boolean isLambdaParameter(DetailAST ast) {
         DetailAST parent;
-        for (parent = ast.getParent(); parent != null; parent = parent.getParent()) {
```
rnveach, line 1218: Similar to https://github.com/checkstyle/checkstyle/pull/12628#discussion_r1068204516 .

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -886,28 +885,14 @@ private static boolean isAstInside(DetailAST tree, DetailAST ast) {
      * @param ident ident token.
      * @return true if field can be referenced from a static context.
      */
-    private boolean canBeReferencedFromStaticContext(DetailAST ident) {
-        AbstractFrame variableDeclarationFrame = findFrame(ident, false);
-        while (variableDeclarationFrame.getType() == FrameType.BLOCK_FRAME
```
rnveach, line 891: This mutation had a cascading effect, and so the end result is the frame of the `ident` does not matter.

According to the callers, the frame can only be `CLASS_FRAME`, `METHOD_FRAME`, `BLOCK_FRAME`. So the question is why does it not matter if act on a `BLOCK_FRAME`.

This requires more in-depth analysis.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -886,28 +885,14 @@ private static boolean isAstInside(DetailAST tree, DetailAST ast) {
...
-            || variableDeclarationFrame.getType() == FrameType.FOR_FRAME) {
```
rnveach, line 892: This mutation had a cascading effect, and so the end result is the frame of the `ident` does not matter.

According to the callers, the frame can only be `CLASS_FRAME`, `METHOD_FRAME`, `BLOCK_FRAME`.

This means the frame can never be `FOR_FRAME`, so that is why it can be removed without issue.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
rnveach: So I am not finding anything more to kill this. It does seem like the name doesn't matter since we are comparing the same frame to each other for anonymous classes. `getNearestClassFrameName` finds the first `CLASS_FRAME`. We can't put some other class into the anonymous class and it doesn't seem like the frame scans upper frames that much.

I don't agree that this change is valid, but I can't find anyway to kill it naturally.

Is it ok to make a reflective test to kill this?

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheckTest.java
```diff
@@ -529,6 +529,21 @@ public void testUnusedMethod() throws Exception {
             .isEqualTo("CATCH_FRAME");
     }
 
+    @Test
+    public void testUnusedMethodFor() throws Exception {
```
rnveach, line 533: New surviving from cascades.
https://github.com/checkstyle/checkstyle/actions/runs/3903501555/jobs/6667951329#step:6:418

I assume it cascade from https://github.com/checkstyle/checkstyle/pull/12628#discussion_r1068222005 .

rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/338/
No differences.

rnveach:
I am still trying to analyze one more mutation, but I am not optimistic on finding a solution. This check is too complex to analyze some of these easily.

rnveach:
Only blocked for merging, I think I have a new way to kill a mutation in a new PR.

rnveach:
I am giving up on these mutations, I can't crack them.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -886,28 +886,14 @@ private static boolean isAstInside(DetailAST tree, DetailAST ast) {
...
-            variableDeclarationFrame = variableDeclarationFrame.getParent();
-        }
-
+    private static boolean canBeReferencedFromStaticContext(DetailAST ident) {
         boolean staticContext = false;
 
-        if (variableDeclarationFrame.getType() == FrameType.CLASS_FRAME) {
```
rnveach, line 898: This mutation is what cascades a bunch of the others.

The issue is the above while and frame loop is what keeps looking the top level class. That is what caused the NPE at https://github.com/checkstyle/checkstyle/issues/12629 when I fooled into thinking it found the top level class when it found the catch frame.

`getCodeBlockDefinitionToken` also does its own scan for the top method/static init.

We need `variableDeclarationFrame` to find a non-class frame that will give us a different static value then the `getCodeBlockDefinitionToken` will. I could not find anything even with weird code.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
rnveach: This is cascaded from https://github.com/checkstyle/checkstyle/pull/12628#discussion_r1068792628 and depends on it's analysis.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
nrmancuso: So, all anonymous class frames are named `new`?

nrmancuso: >We can't put some other class into the anonymous class

Not sure what exactly this means, can you elaborate?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
nrmancuso: Associated suppression:
```
  <mutation unstable="false">
    <sourceFile>RequireThisCheck.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.RequireThisCheck</mutatedClass>
    <mutatedMethod>canBeReferencedFromStaticContext</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
    <description>removed conditional - replaced equality check with true</description>
    <lineContent>if (variableDeclarationFrame.getType() == FrameType.CLASS_FRAME) {</lineContent>
  </mutation>

```

nrmancuso (approved):
After playing with this check and digging through the code, I can't dispute any of the changes here; however, I can say that this check needs to get put on a leash. I am OK with aggressive changes to get it under control.

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheckTest.java
romani: as we agreed in private chat....
lets keep survival not a hacky junit.

sadly , there is no ability to make comment in generated xml files to warn users to not play with this survival, as it too expensive for now.

what we can do .... is keep such test methods with comment "until XXXXX" to remove them and convert to suppression if suppression file is about to be removed.

romani:
the only item https://github.com/checkstyle/checkstyle/pull/12628/files#r1070612087

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheckTest.java
nrmancuso: > sadly , there is no ability to make comment in generated xml files to warn users to not play with this survival, as it too expensive for now.

Why can't we make comment in xml file?

romani: we do `diff` on generated and git file. 
we can potentially remove all comments from files before diff. But it will make extra complication for people on local, `git diff` will not work anymore.

nrmancuso: > we do diff on generated and git file

Where do we do this? One of the reasons for migrating to xml file was to be able to place comments and have permanent suppression list.

nrmancuso: We have comments in files already: https://github.com/checkstyle/checkstyle/blob/72fa0536c643e56934cb7095061150ebf17a84e6/config/pitest-suppressions/pitest-common-2-suppressions.xml#L183

rnveach:
@romani @nrmancuso I am looking for an answer to https://github.com/checkstyle/checkstyle/pull/12628#discussion_r1068301085 .

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheckTest.java
rnveach: @romani I think we are going to have some other issues because of git diff. We don't validate git diff on a success. Someone could put things in the wrong order and it will cause a weird git diff when some failure does occur and the user copies the git diff.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
nrmancuso: Can we just remove name field from this class if it doesn’t matter?

rnveach: We can remove the name field, but we can't remove the override for `getFrameName`. Returning `null` for it gives an NPE. Giving the AST in the constructor to the super of `AnonymousClassFrame` gives an invalid difference in the tests as it then thinks its a class instead of an anonymous at https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java#L493 .

I am still not liking some of this code for this check.

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheckTest.java
romani: We try to validate that after execution of CI , git state is not changed.

If we missed it somewhere we should fill gap.

We can make single issue to re-review pure unit tests after we finish with pitest.

rnveach: Comment on the method? I thought we were saying to remove this method and leave suppression?

romani: I just try to think how to avoid anyone else to try to tackle it.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
romani: lets make reflection hack to kill mutation and in comment explain that we keep this junit until some issue where we will need to rethink such survivals and kill them with better design or ......
I am ok sacrifice our rule on "no pure junit", as we canont keep comments in suppression xml files and we can do this in java.

romani:
@rnveach , replied.

romani (approved):
ok to merge on green CI

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheckTest.java
nrmancuso: We do not run `git diff` on pitest: https://github.com/checkstyle/checkstyle/actions/runs/3924059900/jobs/6708062064

>We try to validate that after execution of CI , git state is not changed.

> If we missed it somewhere we should fill gap.

We do not write to file automatically, git diff validation will have no affect on this. We need to be able to place comments on suppressions, this is one of the advantages of new suppression system.

>We can make single issue to re-review pure unit tests after we finish with pitest.

We agreed that suppression is better than pure UTs for difficult mutations, we have ability to leave comments on suppressions, we should do it.

Please create a permanent suppression list at the bottom of the suppression file and move this one there.

romani: update of files is in progress - https://github.com/checkstyle/checkstyle/pull/12614

rnveach: @romani @nrmancuso Please communicate with each other and let me know which way you both want to handle this as you both gave conflicting changes requested.

nrmancuso: Whatever @romani wants to do is fine, let's get this merged.

romani: @rnveach , please make a comment above to explain reason. And may be create ticket that we will use to as single marker to quickly find such compromise unit tests.

rnveach: Done.

romani:
@rnveach , please fix violation https://app.circleci.com/pipelines/github/checkstyle/checkstyle/17077/workflows/05180a89-33d8-4121-bfcc-e0345ca934e5/jobs/235500/artifacts

romani:
@rnveach , please find time to resolve violation and merge PR.

rnveach:
CI is green.

Issue #12637, opened 2023-01-14
Resolve surviving pitest-metrics-check mutations for pitest


rnveach:
Mutations:
https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-metrics-suppressions.xml

This issue will resolve the remaining suppressions and remove the entire file. POM does need to be updated.

romani:
Fix is merged

PR #12639, opened 2023-01-14
Issue #12637: removes code for the metrics profile

rnveach:
Issue #12637

Regression to come.


src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/AbstractClassCouplingCheck.java
```diff
@@ -160,7 +160,6 @@ public final void setExcludedClasses(String... excludedClasses) {
     public void setExcludeClassesRegexps(String... from) {
         Arrays.stream(from)
                 .map(CommonUtil::createPattern)
-                .distinct()
```
rnveach, line 163: Similar to others, this just saves us processing time. It doesn't matter how many we have of the same string. I would also think this should be somewhat rare unless they are doing something funky in the config.

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/BooleanExpressionComplexityCheck.java
```diff
@@ -353,7 +353,6 @@ private final class Context {
          */
         private Context(boolean checking) {
             this.checking = checking;
-            count = 0;
```
rnveach, line 356: Similar to a previous one, it is assigned 0 even if we don't do anything so this assignment is unnecessary.

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
```diff
@@ -266,7 +266,7 @@ public final class NPathComplexityCheck extends AbstractCheck {
     private final TokenEnd processingTokenEnd = new TokenEnd();
 
     /** NP value for current range. */
-    private BigInteger currentRangeValue = INITIAL_VALUE;
```
rnveach, line 269: This has to be removed so the test works that it is reset in the `beginTree` otherwise the test can't tell the difference.

src/test/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheckTest.java
```diff
@@ -341,6 +346,7 @@ public void testVisitTokenBeforeExpressionRange() {
         astTernary.addChild(astTernaryTrue);
 
         final NPathComplexityCheck npathComplexityCheckObj = new NPathComplexityCheck();
+        npathComplexityCheckObj.beginTree(null);
```
rnveach, line 349: Will get a NPE without this because of the previous change.

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
```diff
@@ -428,7 +428,7 @@ public void leaveToken(DetailAST ast) {
     private void visitConditional(DetailAST ast, int basicBranchingFactor) {
         int expressionValue = basicBranchingFactor;
         DetailAST bracketed;
-        for (bracketed = ast.findFirstToken(TokenTypes.LPAREN).getNextSibling();
+        for (bracketed = ast.findFirstToken(TokenTypes.LPAREN);
```
rnveach, line 431: TBD

rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/340/

No differences.

config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
```diff
@@ -4296,6 +4296,13 @@
     <lineContent>private Deque&lt;Counter&gt; counters;</lineContent>
   </checkerFrameworkError>
 
+  <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java</fileName>
+    <specifier>initialization.field.uninitialized</specifier>
+    <message>the default constructor does not initialize field currentRangeValue</message>
```
rnveach, line 4302: @romani @nrmancuso @Vyom-Yadav This check conflicts with pitest when it is the default value is assigned. I am sure we have a check to also avoid default assignments as they are already defaulted to when we don't assign them.
Pitest reasoning at https://github.com/checkstyle/checkstyle/pull/12639#discussion_r1070235437

romani (approved):
I am fine with placing new violation in Checker suppression.
For now we value pitest more.

nrmancuso (approved):

nrmancuso:
Ok to merge, @rnveach please rebase and resolve conflicts

rnveach:
Rebased and made 100% and removed file.

Issue #12641, opened 2023-01-15
Resolve surviving pitest-design-check mutations for pitest


rnveach:
Mutations:
https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-design-suppressions.xml

This issue will resolve the remaining suppressions and remove the entire file. POM does need to be updated.

bharatsheoran03:
I want to work on this issue, but I don't know how to do as I am beginner and new to open source. Proper guidance can help to resolve this issue.

rnveach:
@bharatsheoran03 This issue is already being worked on.

Please look over https://github.com/checkstyle/checkstyle/wiki/To-GSoC-participant to begin.

rnveach:
Fix was merged

PR #12643, opened 2023-01-15
Issue #12641: removes code for the design profile

rnveach:
Issue #12641

Regression and reasons to come.


src/main/java/com/puppycrawl/tools/checkstyle/checks/design/HideUtilityClassConstructorCheck.java
```diff
@@ -248,10 +248,7 @@ public boolean isHasPublicCtor() {
          */
         public void invoke() {
             final DetailAST objBlock = ast.findFirstToken(TokenTypes.OBJBLOCK);
-            hasNonStaticMethodOrField = false;
-            hasNonPrivateStaticMethodOrField = false;
             hasDefaultCtor = true;
-            hasPublicCtor = false;
```
rnveach, line 254: This method is the first thing called after the class is instantiated, so this might as well be part of the constructor. All these assignments are the already default values assigned in the constructor, so there is no reason to re-assign them.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/InnerTypeLastCheck.java
```diff
@@ -94,7 +94,7 @@ public class InnerTypeLastCheck extends AbstractCheck {
     );
 
     /** Meet a root class. */
-    private boolean rootClass = true;
+    private boolean rootClass;
```
rnveach, line 97: This value is assigned in the `beginTree` so its pointless to assign it here.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/VisibilityModifierCheck.java
```diff
@@ -476,12 +476,10 @@
...
 
     /** Set of ignore annotations short names. */
-    private Set<String> ignoreAnnotationShortNames =
-            getClassShortNames(DEFAULT_IGNORE_ANNOTATIONS);
+    private Set<String> ignoreAnnotationShortNames;
 
     /** Set of immutable classes short names. */
-    private Set<String> immutableClassShortNames =
-        getClassShortNames(DEFAULT_IMMUTABLE_TYPES);
+    private Set<String> immutableClassShortNames;
```
rnveach, line 482: These 2 are similar to the previous. They are assigned default values in the `beginTree`.

`immutableClassShortNames` is given `immutableClassCanonicalNames`, and `immutableClassCanonicalNames`'s default value is `DEFAULT_IMMUTABLE_TYPES`.
`ignoreAnnotationShortNames` is given `ignoreAnnotationCanonicalNames`, and `ignoreAnnotationCanonicalNames`'s default value is `DEFAULT_IGNORE_ANNOTATIONS`.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/InnerTypeLastCheck.java
```diff
@@ -127,7 +127,7 @@ public void visitToken(DetailAST ast) {
             rootClass = false;
         }
         else {
-            DetailAST nextSibling = ast.getNextSibling();
```
rnveach, line 130: This is connected to `CLASS_MEMBER_TOKENS`.

`ScopeUtil.isInCodeBlock` passes but `CLASS_MEMBER_TOKENS` has the specific tokens: VARIABLE_DEF, METHOD_DEF, CTOR_DEF, INSTANCE_INIT, STATIC_INIT, COMPACT_CTOR_DEF. However `ast` is from the visit method, and it only stops on CLASS_DEF, INTERFACE_DEF, RECORD_DEF before going on to the siblings.

Since neither list ever has a matching token, it means this mutation can never be killed.

rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/341/

No differences.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/VisibilityModifierCheck.java
```diff
@@ -963,7 +939,7 @@ private static boolean isPrimitive(DetailAST type) {
      */
     private static String getCanonicalName(DetailAST type) {
         final StringBuilder canonicalNameBuilder = new StringBuilder(256);
-        DetailAST toVisit = type.getFirstChild();
+        DetailAST toVisit = type;
```
rnveach, line 942: This is called from `visitImport` and `getTypeName`.

For `visitImport`, `type` is the the first child of the FQCN being imported, so its always the period. You can't import anything from the blank package, so we are always guaranteed a period first. This only looks at `IDENT` tokens, so regardless, it always skips over anything that is a period, and we can't import from the blank package so there is no way to kill it here. It is just slightly more processing to come to the same result.

For `getTypeName`, the `type` is always the `TYPE` token like in `VARIABLE_DEF`. As we said above, we only care about `IDENT` in this method, so we always skip `TYPE` for the next token.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/VisibilityModifierCheck.java
```diff
@@ -825,7 +822,7 @@ private boolean isImmutableField(DetailAST variableDef) {
         if (isFinalField(variableDef)) {
             final DetailAST type = variableDef.findFirstToken(TokenTypes.TYPE);
             final boolean isCanonicalName = isCanonicalName(type);
-            final String typeName = getTypeName(type, isCanonicalName);
+            final String typeName = getCanonicalName(type);
```
rnveach, line 825: There is a big difference here, so this should be killable.
Example: 
Code: `ImmutableSet<String>`
Results: `ImmutableSet` vs `String`

rnveach: My original analysis was on a clean master. It turns out this is now a cascade from https://github.com/checkstyle/checkstyle/pull/12643#discussion_r1070767748 .

When we removed the `getFirstChild` in `getCanonicalName`, it became the exact same as `type.getFirstChild().getText()`.

The only time it is not exactly the same is on primitive types as they are not `IDENT`s but `LITERAL_...`. 

Unfortunately it doesn't matter with the if/else after the call in `isImmutableField` because the other checks are working. If we say "int" is immutable, it has no type arguments and returns true for the if. Primitives are not canonical and `isPrimitive` will identify the primitive and returns true for the else.

The other usage is in `getTypeArgsClassNames`. The problem is primitives can't be a type argument (IE: Class<T>). So I am not seeing a way for this to be killed.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/VisibilityModifierCheck.java
```diff
@@ -661,8 +659,7 @@ private boolean hasIgnoreAnnotation(DetailAST variableDef) {
      */
     private void visitImport(DetailAST importAst) {
         if (!isStarImport(importAst)) {
-            final DetailAST type = importAst.getFirstChild();
-            final String canonicalName = getCanonicalName(type);
+            final String canonicalName = getCanonicalName(importAst);
```
rnveach, line 662: This looks like another one connected to https://github.com/checkstyle/checkstyle/pull/12643#discussion_r1070767748 . The first child is always a period while the main ast is `IMPORT` token.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -171,7 +171,7 @@ public void visitToken(DetailAST ast) {
             case TokenTypes.INTERFACE_DEF:
             case TokenTypes.RECORD_DEF:
                 final TypeDeclarationDescription description = new TypeDeclarationDescription(
-                    extractQualifiedTypeName(ast), typeDeclarations.size(), ast);
```
rnveach, line 174: Last TBD.

config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
```diff
@@ -2743,6 +2743,20 @@
...
+    <message>the default constructor does not initialize field ignoreAnnotationShortNames</message>
+    <lineContent>private Set&lt;String&gt; ignoreAnnotationShortNames;</lineContent>
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/design/VisibilityModifierCheck.java</fileName>
+    <specifier>initialization.field.uninitialized</specifier>
+    <message>the default constructor does not initialize field immutableClassShortNames</message>
+    <lineContent>private Set&lt;String&gt; immutableClassShortNames;</lineContent>
+  </checkerFrameworkError>
```
romani, line 2758: should we disable this Rule is Checker ?

I see we do this suppression again and again, and there are ~100  of such suppression already.

romani (approved):
ok to merge

config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
rnveach: See my previous comment at https://github.com/checkstyle/checkstyle/pull/12639#discussion_r1070312108

This only conflicts with pitest.

romani: yes, so that is why I am questioning role of rule in our project, looks like very conflicting with java Checks that has initialization methods `beginTree` and `beginProcessing`.

rnveach: I am fine to disable, but this makes me question if there are any rules from this process that we do like.

romani: We will prove by practice that some rules unfortunately does not match us or conflicting with tools that we like more.
For now we suppressing. If we endup in suppressing of all eventually, it would be good to show/share publicly this experience.

Nyamsons:
I want to generate a code

romani:
@rnveach , please rebase and merge this PR.
@nrmancuso  does not have time to review.

config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
nrmancuso: Let’s make an issue for disablement if this rule and discuss there.

nrmancuso (approved):

rnveach:
Rebased and made 100%.

romani (approved):

Issue #12647, opened 2023-01-15
pitest: groovy does not run when profile fails


rnveach:
Identified at https://github.com/checkstyle/checkstyle/actions/runs/3925617482/jobs/6710655116#step:6:436 ,

No differences were reported which is what the groovy script displays. Coverage report is still uploaded so it can be viewed there for now, but if we start adding suppressions, then the report may become less clear.

I just expect groovy to run regardless if it passes/fails to display the removals/additions.

rnveach:
Fix is merged

PR #12648, opened 2023-01-15
Issue #12647: ensure groovy runs when pitests fail 

rnveach:
Issue #12647

### DIFF
--- a/.ci/pitest.sh
+++ b/.ci/pitest.sh
@@ -18,12 +18,15 @@ case $1 in
   PROFILES=$(list_profiles "${0}");
 
   if [[ $(echo "$PROFILES" | grep -w -- "${1}" | cat) != "" ]]; then
-    mvn -e --no-transfer-progress -P"$1" clean test-compile org.pitest:pitest-maven:mutationCoverage
+    RESULT=0
+    mvn -e --no-transfer-progress -P"$1" clean test-compile \
+      org.pitest:pitest-maven:mutationCoverage | RESULT=1
     if [[ "$2" != "" ]]; then
       groovy ./.ci/pitest-survival-check-xml.groovy "$1" "$2"
     else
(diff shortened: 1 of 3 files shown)


rnveach:
Output during a failure: https://github.com/checkstyle/checkstyle/actions/runs/3926061148/jobs/6711436469#step:6:62

.ci/pitest.sh
```diff
@@ -18,12 +18,15 @@ case $1 in
   PROFILES=$(list_profiles "${0}");
 
   if [[ $(echo "$PROFILES" | grep -w -- "${1}" | cat) != "" ]]; then
-    mvn -e --no-transfer-progress -P"$1" clean test-compile org.pitest:pitest-maven:mutationCoverage
+    RESULT=0
+    mvn -e --no-transfer-progress -P"$1" clean test-compile \
+      org.pitest:pitest-maven:mutationCoverage | RESULT=1
```
rnveach, line 23: I don't understand why this drops the maven failure.

https://github.com/checkstyle/checkstyle/actions/runs/3926061139/jobs/6711436264
> SC2030 (info): Modification of RESULT is local (to subshell caused by pipeline).
> SC2031 (info): RESULT was modified in a subshell. That change might be lost.
I also don't know what to do to please shellcheck.

Vyom-Yadav: Why not just `set +e` before and set it back to `set -e` . If the maven command execution fails for XYZ  reason, then the pit report won't be created (most likely) and the groovy command will subsequently fail.

romani: better to play with "set + e", we have examples in our scripts

rnveach: @Vyom-Yadav If you wish to take on this issue, please do. I am not familar enough with bash to resolve this.

Vyom-Yadav: Created #12686, you can close this.

PR #12649, opened 2023-01-16
supplemental: changes to assembly profile for contribution#747

rnveach:
Supplemental for https://github.com/checkstyle/contribution/pull/747#discussion_r1064284794

> let's combine no-validation and assembly.
> 
> But assembly should skip tests jaring and assembly plugin. So all 2 parameters should be part of assembly profile.
test jars assemble required only during release process. -Passembly was purely a profile to make "-all" jar that is not maven concept of jars.


romani (approved):
Ok to merge

romani:
merging without @nrmancuso , as he has lack of time for review.

PR #12652, opened 2023-01-16
Issue #12645: All GitHub CI workflows should have concurrency of 1

### LINKED ISSUE #12645: All Github CI workflows should have concurrency of 1
Currently, all of our Github workflows will continue to run after new pushes to a PR. This is usually pointless, and consumes a lot of extra resources (credits + occupied containers). This causes other workflows to be queued for long periods, and also can slow down the release process.

Solution: we need to set all github workflows to have a concurrency of 1, and automatically cancel in progress workflows.

Example of yaml update:
```
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

We also need to create some check to make sure that all github workflows have this concurrency setting.

nrmancuso:
Closes #12645

Example of failure: https://circleci.com/gh/checkstyle/checkstyle/234145

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -1107,6 +1107,27 @@ ci-temp-check)
     exit $fail
   ;;
 
+  check-github-workflows-concurrency)
+    GITHUB_WORKFLOW_FILES=$(find .github/workflows -maxdepth 1 -not -type d -name "*.y*ml")
+
+    FILES_NO_CONCURRENCY=()
+    for f in $GITHUB_WORKFLOW_FILES; do
+      if ! grep -wq "concurrency:" "$f"; then
+            FILES_NO_CONCURRENCY+=( $f )
+      fi
+    done
+
+    if [[ ${#FILES_NO_CONCURRENCY[@]} -gt 0 ]]; then
+      echo "The following Github workflows are missing a concurrency block:"
+    fi
+
+    for value in "${FILES_NO_CONCURRENCY[@]}"; do
+      echo "$value"
+    done
+
+    exit ${#FILES_NO_CONCURRENCY[@]}
+    ;;
+
 *)
   echo "Unexpected argument: $1"
   sleep 5s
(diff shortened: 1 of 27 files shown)


.github/workflows/update_sources.yml
```diff
@@ -9,6 +9,11 @@ on:
       version:
         description: 'Target Version without (-SNAPSHOT)'
         required: true
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
nrmancuso, line 15: Github CI uses concurrency groups to determine concurrency. So, we need to create a group name that is unique to each PR and each workflow in that PR. However, when it comes to tags and master, `github.event.pull_request.number` will not exist. `||` allow us to specify an alternative, which in this case is the fully formed `ref`.

romani (approved):
ok to merge

rnveach:
@nrmancuso CI is failing

.ci/validation.sh
```diff
@@ -1105,6 +1105,27 @@ ci-temp-check)
     exit $fail
   ;;
 
+  check-github-workflows-concurrency)
+    GITHUB_WORKFLOW_FILES=$(find .github/workflows -maxdepth 1 -not -type d -name "*.y*ml")
```
rnveach, line 1109: > "*.y*ml"

Is this to include `yaml` files too? I don't see any changes in this PR for `yaml`.

nrmancuso:
> @nrmancuso CI is failing

Yes, this is intentional :)

.ci/validation.sh
romani: I do not see any "yaml" at https://github.com/checkstyle/checkstyle/tree/master/.github/workflows

nrmancuso: Yes, just future-proofing in case someone uses yaml extension.

rnveach (approved):

PR #12659, opened 2023-01-17
infra: fix indentation and add echo of output

nrmancuso:
We have a failure to generate diff report when we use more than one parameter. See examples below.

Ok, we need to merge this to see results from `echo`.

My guess is that we are appending to GITHUB_OUTPUT in such a way as to cause issues when we later retrieve this information.

Diff Regression config: https://gist.githubusercontent.com/nrmancuso/51b18335d59bc81bb09370564e258f51/raw/5dbbab498fceb4b15f5395cebcf133f2d7a02ba4/config.xml

### DIFF
--- a/.github/workflows/diff_report.yml
+++ b/.github/workflows/diff_report.yml
@@ -85,6 +85,9 @@ jobs:
           sed 's/Report label: //' temp > report_label
           echo "report_label=$(cat report_label)" >> "$GITHUB_OUTPUT"
 
+          echo "GITHUB_OUTPUT:"
+          echo "$GITHUB_OUTPUT"
+
       - name: Set branch
         id: branch
         run: |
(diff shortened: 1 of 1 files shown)


nrmancuso:
Github, generate report

.github/workflows/diff_report.yml
```diff
@@ -106,19 +106,20 @@ jobs:
...
-         fi
-
-         if [ -n "$DIFF_CONFIG_LINK" ]; then
-           wget -q "$DIFF_CONFIG_LINK" -O diff_config.xml
-         fi
-
-         if [ -n "$PATCH_CONFIG_LINK" ]; then
-           wget -q "$PATCH_CONFIG_LINK" -O patch_config.xml
-         fi
+          LINK="${LINK_FROM_PR:-$DEFAULT_PROJECTS_LINK}"
```
nrmancuso, line 109: Indentation was -1 for this block, not sure if this had something to do with problems or just one of the links below.

nrmancuso:
Github, generate report

nrmancuso:
`Unable to process file command 'output' successfully.` I suspect that new bash syntax in workflow is the issue, looking deeper.

nrmancuso:
Github, generate report

nrmancuso:
Github, generate report

nrmancuso (commented):
Action will not run with changed workflow file, we need to merge this to get `echo`s.

.github/workflows/diff_report.yml
romani: It should not, we have `|` that is multiline marker

.github/workflows/diff_report.yml
```diff
@@ -85,6 +85,8 @@ jobs:
           sed 's/Report label: //' temp > report_label
           echo "report_label=$(cat report_label)" >> "$GITHUB_OUTPUT"
 
+          echo "$GITHUB_OUTPUT"
+
```
romani, line 89: Let put some text upfront to explain output

romani (changes requested):
Items

.github/workflows/diff_report.yml
nrmancuso: Multiline marker does count on increased indentation in most interpreters, but it should have been good since it was indented more than it's key. Better to fix it either way. This file is too complex for most interpreters, I think this is why yamllint failed to find bad indentation.

.github/workflows/diff_report.yml
nrmancuso: Done.

romani: it is kind of weird, that GITHUB_OUTPUT is not self printed.
it is desinged to be self printed.
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

is this PR is experiment ? to "hack" Gitub and print it.
I thought that github redirect all output to his internals, that is why they need this variable.

nrmancuso: I am not sure what "self-printed" means, but `output` in this context is the output of a given step, to be passed to the next one.

nrmancuso: >is this PR is experiment ? to "hack" Gitub and print it.

Sort of. It is a good sanity check to know what we are passing to the next step. You cannot ssh into containers like you can with CircleCI for debugging, so it will also be helpful for that.

nrmancuso:
>may be we reached deprecation time ?

No, I think this warning is from `khan/pull-request-comment-trigger@master`, not us, since I removed all usage of this command in actionlint PR. But, rocket still works, I do not think this is the case.

romani (approved):
Ok to merge but remove this print if it doesn't work

Issue #12553, opened 2022-12-27
Add 'when' to default IllegalIdentifierName pattern


nrmancuso:
I have read check documentation: https://checkstyle.sourceforge.io/config_naming.html#IllegalIdentifierName
I have compiled snapshot from #12531 
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

How it works now:
```bash                                                                                                          
➜  src cat Test.java 
public class Test {
    String m(Integer integer) {
        return switch(integer) {
          case Integer when when when < 0 -> when + " is negative";
          case null, default -> "";
        };
    }
}

➜  src cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <module name="TreeWalker">
    <module name="IllegalIdentifierName"/>
  </module>
</module>

➜  src javac --enable-preview --release 19 Test.java
Note: Test.java uses preview features of Java SE 19.
Note: Recompile with -Xlint:preview for details.

➜  src java -jar checkstyle-10.5.1-SNAPSHOT-all.jar -c config.xml Test.java
Starting audit...
Audit done.

```

Expected:
```
➜  src java -jar checkstyle-10.5.1-SNAPSHOT--all.jar -c config.xml Test.java
Starting audit...
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:4:13: Name 'when' must match pattern '(?i)^(?!(record|yield|var|permits|sealed|when|_)$).+$'. [IllegalIdentifierName]
Audit done.
Checkstyle ends with 1 errors.

```

romani:
better to focus on Check deal with single "when" , multiple only as super stability. I hope it will be blocked by javac in final version.

Kevin222004:
I am on it

Kevin222004:
so i want be able to see the ast till https://github.com/checkstyle/checkstyle/pull/12531 not get merged and i get the snapshot of that right

nrmancuso:
@Kevin222004 you do not need to wait; since this check just looks at identifiers of other structures, it does not depend on #12531. You could make a branch off of the branch at #12531 to work on this if you wanted to, though. It goes something like this:

```
git checkout master
git fetch upstream pull/12531/head:pull-12531
git checkout pull-12531
git checkout -b issue-12553

```

[...]

PR #12665, opened 2023-01-19
Issue #12553: Add 'when' to default IllegalIdentifierName pattern

Kevin222004:
Issue #12553: Add 'when' to default IllegalIdentifierName 

------------------------------------------------------

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/f8cbb850ce72234e08c6963c6f4147b6/raw/25d74a45948b256b83ba02fb72a9882214916120/name.xml

### DIFF
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/naming/illegalidentifiername/InputIllegalIdentifierNameWhen.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/naming/illegalidentifiername/InputIllegalIdentifierNameWhen.java
@@ -1,0 +1,55 @@
+/*
+IllegalIdentifierName
+format = (default)(?i)^(?!(record|yield|var|permits|sealed|when|_)$).+$
+tokens = (default)CLASS_DEF, INTERFACE_DEF, ENUM_DEF, ANNOTATION_DEF, ANNOTATION_FIELD_DEF, \
+         PARAMETER_DEF, VARIABLE_DEF, METHOD_DEF, ENUM_CONSTANT_DEF, PATTERN_VARIABLE_DEF, \
+         RECORD_DEF, RECORD_COMPONENT_DEF, LAMBDA
+
+
+*/
+
+//non-compiled with javac: Compilable with Java19
+package com.puppycrawl.tools.checkstyle.checks.naming.illegalidentifiername;
+
+public class InputIllegalIdentifierNameWhen {
+
+    int When = 10; // violation 'Name 'When' must match pattern .*'
+    int when = 10; // violation 'Name 'when' must match pattern .*'
+
+    int otherwhen = 11; // ok
+}
+
+class InputJava19PatternsTrickyWhenUsage {
+    static String m1(Object when) { // violation 'Name 'when' must match pattern .*'
+        when = new Object();
+        return switch (when) {
+            case Integer i when i >= 0 -> i.toString();
+            default -> "2";
+        };
+    }
+
+    static void m2(){
+        int when = Integer.parseInt("when"); // violation 'Name 'when' must match pattern .*'
+    }
+
+    static String m3(Object x) {
+        return switch (x) {
+            // this is horrible, but it compiles
+            case Integer when when when >= 0 -> when.toString();
+            // violation above 'Name 'when' must match pattern .*'
+            default -> "2";
+        };
+    }
+
+    @when static class A {
+        static class when{} // violation 'Name 'when' must match pattern .*'
+    }
+
+    static class B<when> {
+        when getWhen() {
+            return null;
+        }
+    }
+}
+
+@interface when{} // violation 'Name 'when' must match pattern .*'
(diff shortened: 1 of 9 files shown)


Kevin222004:
I have remove this  test case from previous latest commit to avoid ci failure because this is not compilable with java19 i will add after resolving this
```
    void m1(Object o) {
        switch (o) {
            case Integer when when i1 == 0: // violation
            case Integer noWhen when i1 == 1: // ok
                m2(i1);
                break;
        }
    }
    
    void m2() {}
```

Kevin222004:
GitHub, generate report

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/naming/illegalidentifiername/InputIllegalIdentifierNameWhen.java
```diff
@@ -0,0 +1,54 @@
...
+public class InputIllegalIdentifierNameWhen {
+
+    int When = 10; // violation
+    int when = 10; // violation
+
+    int otherwhen = 11; // ok
+}
+
+class InputJava19PatternsTrickyWhenUsage {
+    static String m1(Object when) { // violation
```
romani, line 23: add message to all "// violation"

Do you any suggestion on how to enforce this rule by Checkstyle and put in suppression all existing files, to let us violate new files.

romani (changes requested):
items:

Kevin222004:
GitHub, generate report

Kevin222004:
GitHub, generate report

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/naming/illegalidentifiername/InputIllegalIdentifierNameWhen.java
Kevin222004: it is becoming difficult to put violation message here bcs msg is too long if we use violation above and below still message become longer than 100 line, 
although if we not consider the 100 line it is failing only with 
```
Actual and expected violations differ.
expected to match: 16:(?:\d+:)?\s.*Name 'When' must match pattern '\(?i\)^\(?!\(record|yield|var|permits|sealed|when|_\)$\).+$'.*
but was          : 16:9: Name 'When' must match pattern '(?i)^(?!(record|yield|var|permits|sealed|when|_)$).+$'.
```  

we can write the message here something like this 
`    int When = 10; // violation 'Name 'When' must match pattern .*' ` this is working fine is it good to show message like this 

> Do you any suggestion on how to enforce this rule by Checkstyle and put in suppression all existing files, to let us violate new files.

i have to think on this right know i have not any suggestion once the issue https://github.com/checkstyle/checkstyle/issues/11214 will resolve and all the suppression from bdd is removed then if we not write the violation message on that file the mvn it would not pass 
right know  i want have any idea

Kevin222004: AS of know message written

romani:
Code is good, but report is full of noise only due to change in message.
Please update config to override message to make it as it was before without 'when'.

Kevin222004:
GitHub, generate report

Kevin222004:
> Code is good, but report is full of noise only due to change in message. Please update config to override message to make it as it was before without 'when'. https://checkstyle.org/config.html#Custom_messages

done

romani (approved):
ok to merge .

But I have bad feeling that a lot of items become to be violation. This might be a sign that it is not good to make it behavior by default.
Lets ask other maintainers for opinion.

nrmancuso:
> But I have bad feeling that a lot of items become to be violation. This might be a sign that it is not good to make it behavior by default.

After looking at the reports and thinking on this, I tend to agree. Maybe we should wait on this update at least until we are sure that `when` will make it into Java officially. I was not aware that `when` was such a common naming convention. Also, considering that `when` is limited to only case labels, it may not be such a bad practice to use it elsewhere.

Kevin222004:
can't we make it as a separate property know

rnveach:
User can always turn off what conflicts for them.

https://checkstyle.org/config_naming.html#IllegalIdentifierName
> Checks identifiers with a pattern for a set of illegal names, such as those that are restricted or contextual keywords. 

I can't believe we never ran into this for `record`.

> wait on this update at least until we are sure that when will make it into Java officially

I atleast agree with this. This is not a change in our grammar, so we don't need to worry about failed parsing. My mindset is unofficial Java things are basically just test grounds and they could come and go. It seems odd to support something like this officially from us when it is not even official for Java.

I am more for not supporting unofficial things, but if we want, we can atleast add documentation for users on how to add it if they wish.

Kevin222004:
What would be the next step, should we wait for the official update of java for when

rnveach:
@nrmancuso @romani ping. We need clarification on issue direction.

romani:
We can not merge it until `when` decome a final and official. For now it is in-preview feature of jdk, so might change.

romani:
So others also does not like this update as it affects existing code bases too much.
So having it by default will be too aggressive.

After thinking on this, we should not do this as default behavior. But if no reason to work by default, no reason in this update.

@Kevin222004 , sorry that we wasted your time

PR #12670, opened 2023-01-21
 Issue #11589: adds descriptions for ResultOfMethodCallIgnored

rnveach:
#11589

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
@@ -331,11 +331,11 @@ public class PropertyCacheFileTest extends AbstractPathTestSupport {
     }
 
     /**
-     * Temp comment.
-     * until #11589
+     * Test functionality when toByteArray throws an exception.
      *
      * @noinspection ResultOfMethodCallIgnored
-     * @noinspectionreason ResultOfMethodCallIgnored - temporary suppression until #11589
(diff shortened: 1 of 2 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
```diff
@@ -335,7 +335,7 @@ public void testChangeInConfig() throws Exception {
      * until #11589
```
romani, line 335: should be removed.

src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
```diff
@@ -335,7 +335,7 @@ public void testChangeInConfig() throws Exception {
...
      *
      * @noinspection ResultOfMethodCallIgnored
-     * @noinspectionreason ResultOfMethodCallIgnored - temporary suppression until #11589
+     * @noinspectionreason ResultOfMethodCallIgnored - Setup for mockito for which method to mock.
```
romani, line 338: can we be specific , something like `toByteArray only mocked to throw exception`

romani (changes requested):
items:

src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
rnveach: Done.

romani (approved):

Issue #6223, opened 2018-11-20
Update shell scripts to split sha to avoid pollution in spelling white list


romani:
Workaround to avoid having parts of sha in white list: https://github.com/checkstyle/checkstyle/commit/4cda37a73a4749750768a624732294e534f58550

This is NOT super ideal but better than pollution of whitelist of our spell checker.

rnveach:
@romani The only example I am seeing is `cbeaff` which is a commented link.
https://github.com/checkstyle/checkstyle/blob/master/.ci/test-spelling-unknown-words.sh#L3

Am I missing others is this issue outdated now?

rnveach:
I am only seeing the following:
https://github.com/checkstyle/checkstyle/blob/d26483f8a010f30bb8787037cc2c2f9164a102f4/.ci/validation.sh#L625
https://github.com/checkstyle/checkstyle/blob/d26483f8a010f30bb8787037cc2c2f9164a102f4/.ci/validation.sh#L640
https://github.com/checkstyle/checkstyle/blob/d26483f8a010f30bb8787037cc2c2f9164a102f4/.ci/validation.sh#L649
https://github.com/checkstyle/checkstyle/blob/d26483f8a010f30bb8787037cc2c2f9164a102f4/.ci/validation.sh#L742

@romani How do we know how to split these? Some of them already seem split, so is this issue outdated?

romani:
All is already split, as such sha are updated frequently.

We need to cleanup whitelist file.

We should remove whole content from such file, run spell check, it will generate patch file, we apply patch, difference detected by git should be sent as PR.

rnveach:
> We need to cleanup whitelist file. As bunch is leaked to this file

Normal clean up will occur in https://github.com/checkstyle/checkstyle/issues/5741 . This is only related to SHAs.

Script does a diff, so words automatically removed appear in the diff and thus if the diff does not fail then no words were added or removed. There is also special text by the script when words are removed by the verbage `There are now fewer misspellings than before`.
https://github.com/checkstyle/checkstyle/blob/858dcaa32a00038a157390ace599d9732ca86d7b/.ci/test-spelling-unknown-words.sh#L70

> We should remove whole content from such file, run spell check, it will generate patch file,

The following shows nothing can be removed with even the removal of the whole file.
````
$ git status

HEAD detached at origin/master
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
[...]

PR #12672, opened 2023-01-21
minor: exclude spellcheck's own whitelist words

rnveach:
Bug identified at https://github.com/checkstyle/checkstyle/issues/6223#issuecomment-1399347234


rnveach:
@romani Can you update this script to print a "git diff" instead of a normal diff? It would make things easier like other areas to just apply the git patch.

romani (approved):

romani:
I don't know all nuances of patch/diff
I still periodically have problems to apply patch.

We tried git diff in other PR for pitest, and there are problems to reuse. But here command always worked for me and others.

rnveach:
> But here command always worked for me and others.

Can you provide command for this?

PR #12673, opened 2023-01-21
minor: moves files generated by spellchecker to temp directory

rnveach:
All temp files should be moved to `.ci-temp` as that is our temporary working directory when doing CI commands. It makes it easier to manage these files if they are separated from long term files managed by the git repository.

### DIFF
--- a/.ci/test-spelling-unknown-words.sh
+++ b/.ci/test-spelling-unknown-words.sh
@@ -9,11 +9,12 @@ spellchecker='config/jsoref-spellchecker'
 temp='.ci-temp'
 whitelist_path="$spellchecker/whitelist.words"
 dict="$temp/english.words"
-word_splitter="$spellchecker/spelling-unknown-word-splitter.pl"
-run_output="$spellchecker/unknown.words"
+word_splitter="$temp/spelling-unknown-word-splitter.pl"
+run_output="$temp/unknown.words"
+
+mkdir -p $temp
 
 if [ ! -e "$dict" ]; then
-  mkdir -p $temp
   echo "Retrieve cached english.words from checkstyle.sourceforge.io"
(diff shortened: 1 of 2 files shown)


.ci/test-spelling-unknown-words.sh
```diff
@@ -9,11 +9,12 @@ spellchecker='config/jsoref-spellchecker'
 temp='.ci-temp'
 whitelist_path="$spellchecker/whitelist.words"
 dict="$temp/english.words"
-word_splitter="$spellchecker/spelling-unknown-word-splitter.pl"
-run_output="$spellchecker/unknown.words"
+word_splitter="$temp/spelling-unknown-word-splitter.pl"
+run_output="$temp/unknown.words"
+
+mkdir -p $temp
```
rnveach, line 15: This was moved here as all files need it. There is no file existence check for `$run_output`, so it would always need to be run regardless.

.ci/test-spelling-unknown-words.sh
```diff
@@ -72,6 +73,8 @@ diff_output=$(diff -U1 "$whitelist_path" "$run_output" |grep -v "$spellchecker"
 if [ -z "$diff_output" ]; then
   echo "No new words and misspellings found."
   rm $dict
+  rm $word_splitter
+  rm $run_output
```
rnveach, line 77: https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=12962&view=logs&j=6a833393-c551-5000-0916-5000fd3b86e7&t=bdbbd129-77f0-5d73-ce7d-db58d34c8358&l=36
Clean up has to be done on an exit 0.
None of the others really need to change because they only occur when there will be a failure anyways.

romani (approved):

Issue #7488, opened 2020-01-17
LocalizedMessage should accept message argument in string only


romani:
Detected by IDEA inspection and Sonar Inspection `Make "args" transient or serializable.`
https://sonarcloud.io/project/issues?id=org.checkstyle%3Acheckstyle&issues=AW9t2w2MYD2QG1pPXIUJ&open=AW9t2w2MYD2QG1pPXIUJ

Code:
https://github.com/checkstyle/checkstyle/blob/2f9f8c620eaa953a0d5822c17f59fd396f09c2ad/src/main/java/com/puppycrawl/tools/checkstyle/api/LocalizedMessage.java#L84-L89

We did not change it as it is API class, so it should be done separately and clear consideration of breaking affects.

Expected: change `Object[]` to `String[]` to simply do string concatenation without any thoughts on special format/serialization of particular type.
If some localization is required, it should be done by module/Check that report such violation.

rnveach:
> If some localization is required, it should be done by module/Check that report such violation.

It seems wrong that a class that deals with message localization will not localize the arguments too.

romani:
I am not sure how it can localize something that he do not understand and can not differentiate.
This class is controlling only template string value, so it make good decision on what to use as base, but all other is just some string values that might be not a human words at all.

rnveach:
This issue has a large impact on everything from LocalizedMessage and Violation all the way to modules. Modules cannot have integer arguments anymore, they need to be converted to Strings.

rnveach:
@romani ping

Please clarify the extent of the issue. Are we converting the field only, or are we converting the constructor arguments as well?

romani:
Both. To make it clear that all consideration of how to serialize of arguments should be done outside of this API class. It will print just strings.

rnveach:
Then this a pretty big break. Method signature is changing, so I am sure all checks will need to be recompiled from source. Sevntu is surely impacted. I wasn't seeing any usages in 3rd parties of `Violation` but now this extends much further.

romani:
Yes, I can imagine, we can start with extension of API with new ctor that takes array of string and store it as array of string. And we mark old ctor and field as deprecated to remove in few years.

And we will see migration nuances of sevntu.

rnveach:
@romani I assume this is all coming solely because this class is serializable and everything associated to it needs to be the same. Can you update issue with why this class needs to be serializable?

[...]

Issue #4739, opened 2017-07-17
Non-serializable field 'args' in a serializable LocalizedMessage


romani:
from Intellij Idea inspection:
src/main/java/com/puppycrawl/tools/checkstyle/api/LocalizedMessage.java
81: args Non-serializable field 'args' in a Serializable class

We need to investigate how to make a fix and minimize affect to users.
Looks like LocalizedMessage is not completely serializable so it NOT serializable, nobody use it  as serializable .....

rnveach:
> nobody use it as serializable

Why do we serialize it then?

rnveach:
Similar to #7488

PR #12675, opened 2023-01-22
Issue #12681: removes serializable from violation and audit event

rnveach:
Issue #12681 

We have comments for this class that even say we do no serialize it.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java
@@ -19,8 +19,6 @@
 
 package com.puppycrawl.tools.checkstyle.api;
 
-import java.util.EventObject;
-
 /**
  * Raw event for audit.
  * <p>
@@ -37,11 +35,10 @@ import java.util.EventObject;
  *
  * @see AuditListener
  */
-public final class AuditEvent
-    extends EventObject {
+public final class AuditEvent {
 
-    /** Record a version. */
-    private static final long serialVersionUID = -3774725606973812736L;
+    /** The object on which the Event initially occurred. */
+    private final Object source;
     /** Filename event associated with. **/
     private final String fileName;
     /** Violation associated with the event. **/
@@ -72,11 +69,25 @@ public final class AuditEvent
      * @param src source of the event
      * @param fileName file associated with the event
      * @param violation the actual violation
+     * @throws IllegalArgumentException if {@code src} is {@code null}.
      */
     public AuditEvent(Object src, String fileName, Violation violation) {
-        super(src);
+        if (src == null) {
+            throw new IllegalArgumentException("null source");
+        }
+
+        source = src;
         this.fileName = fileName;
         this.violation = violation;
+    }
+
+    /**
+     * The object on which the Event initially occurred.
+     *
+     * @return the object on which the Event initially occurred
+     */
+    public Object getSource() {
+        return source;
     }
 
     /**
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/Violation.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/Violation.java
@@ -19,7 +19,6 @@
 
 package com.puppycrawl.tools.checkstyle.api;
 
-import java.io.Serializable;
 import java.text.MessageFormat;
 import java.util.Arrays;
 import java.util.Locale;
@@ -32,16 +31,12 @@ import com.puppycrawl.tools.checkstyle.LocalizedMessage;
  * message.properties files. The underlying implementation uses
  * java.text.MessageFormat.
  *
- * @noinspection SerializableHasSerializationMethods, ClassWithTooManyConstructors
- * @noinspectionreason SerializableHasSerializationMethods - we do not serialize this class
+ * @noinspection ClassWithTooManyConstructors
  * @noinspectionreason ClassWithTooManyConstructors - immutable nature of class requires a
  *      bunch of constructors
  */
 public final class Violation
-    implements Comparable<Violation>, Serializable {
-
-    /** A unique serial version identifier. */
-    private static final long serialVersionUID = 5675176836184862150L;
+    implements Comparable<Violation> {
 
     /** The default severity level if one is not specified. */
     private static final SeverityLevel DEFAULT_SEVERITY = SeverityLevel.ERROR;
@@ -64,15 +59,7 @@ public final class Violation
     /** Key for the violation format. **/
     private final String key;
 
-    /**
-     * Arguments for MessageFormat.
-     *
-     * @noinspection NonSerializableFieldInSerializableClass
-     * @noinspectionreason NonSerializableFieldInSerializableClass - usage of
-     *      'Serializable' for this api class
-     *      is considered as mistake now, but we do not break api without
-     *      good reason
-     */
+    /** Arguments for MessageFormat. */
     private final Object[] args;
 
     /** Name of the resource bundle to get violations from. **/
--- a/src/test/java/com/puppycrawl/tools/checkstyle/api/AuditEventTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/api/AuditEventTest.java
@@ -20,6 +20,7 @@
 package com.puppycrawl.tools.checkstyle.api;
 
 import static com.google.common.truth.Truth.assertWithMessage;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import org.junit.jupiter.api.Test;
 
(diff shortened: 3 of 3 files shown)


rnveach:
Test failed at https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=12967&view=logs&j=eb841e9f-6e1e-5c29-3ad5-74ec2c658069&t=a9b761e6-b8d5-50cf-467a-54967189cab2&l=1264 .

============

Notes from https://github.com/checkstyle/checkstyle/issues/4739

Serialization was added at https://github.com/checkstyle/checkstyle/commit/d2af4accdde3eac0a2103353f490e97bf62f61d2#diff-52efb9bb7c26b577e3885a9539eac7d5R42 titled Apply parts of patch #1952556 fixing some FindBugs bugs. Thanks to Travis Schneeberger.

http://git.net/ml/java.audit.checkstyle.devel/2008-04/msg00067.html

Findbugs does produce this error if I remove the serialization:

    [INFO] Class com.puppycrawl.tools.checkstyle.api.AuditEvent defines non-transient non-serializable instance field localizedMessage [com.puppycrawl.tools.checkstyle.api.AuditEvent] In AuditEvent.java SE_BAD_FIELD

This is because AuditEvent is serialized through inheritance of EventObject.

============

Now this is become why is `AuditEvent` serialized and need `EventObject`.

It was added in 1d21af32b9949396bb3f010cd3f3addec4075b5d , 21 years ago, as the first cut.

rnveach:
Only failure is simple `RedundantThis`.

I like this solution more than swapping to `String[]`. Either way, `MessageFormat` will handle properly formatting the message.

src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java
```diff
@@ -37,11 +35,10 @@
  *
  * @see AuditListener
  */
-public final class AuditEvent
-    extends EventObject {
+public final class AuditEvent {
```
romani, line 38: referenced issue is not very clear on what actually is breaking change in this PR.

conceptually removal of extends in implements is breaking changes.
Doe sit make sense to make specific issue on this and merge in this release ? and final update in LocalizeMessage will be done in next release or when change is ready.

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java
rnveach: Referenced issue was another way to resolve this issue, but I am recommending a different way. https://github.com/checkstyle/checkstyle/issues/7488#issuecomment-1402023707 says referenced issue is not good anyways.

> specific issue on this and merge in this release ? 

Yes, I recommend this.

> final update in LocalizeMessage will be done in next release or when change is ready

What final update are you referring to? Just adding documentation update referenced in https://github.com/checkstyle/checkstyle/issues/7488#issuecomment-1402023707 ?

romani: Yes, there will be no final update as it too big to break.

romani: Better to close #7488
And quickly create issue to explain what we breaking in this PR.

rnveach: > quickly create issue to explain what we breaking in this PR.

Done.

> Better to close https://github.com/checkstyle/checkstyle/issues/7488

Can't, it requires documentation update discussed at https://github.com/checkstyle/checkstyle/issues/7488#issuecomment-1402130818 .

rnveach:
@romani ping

src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java
romani: We should describe reason of removal `extends EventObject` in that issue. It might affect some plugins, developer should get details why we removed such extends/implements.

romani:
I am confused by reference of this PR in second commit. I think we can use single issue to explain all reasons and all updates.

rnveach:
Do you want me to combine both commits? 2nd commit was from CI failure on first commit. It is not a class mentioned in main issue, so I made it a second commit.

src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java
romani: I updated https://github.com/checkstyle/checkstyle/issues/12681 to contain all changes in API at one issue.

@rnveach , please squash both commits in one, as we try our best to be compile-able on each commit.

romani:
@rnveach , how we can be confident that this is not HUGE breaking compatibility for Modules and for plugins in our organization ? it will drive bump of first or second digit in version.

src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java
rnveach: > we try our best to be compile-able on each commit

Commits are compilable separate. It was just another CI failure that a serializable class had non-serializable fields.

> please squash both commits in one

Done.

rnveach:
> how we can be confident that this is not HUGE breaking compatibility for Modules and for plugins in our organization ? 

If you can tell me how this can break 3rd parties, then we should add a CI test to confirm. I can only go by what I think I know, but I know I have been wrong before and accidentally broke 3rd parties when it wasn't expected.

https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/config/import-control.xml
Sevntu, our largest 3rd party module, does not directly use Violation or Audit Event. I specifically made the import controls non-wildcards so we can identify which core classes it was using.

https://github.com/checkstyle/sonar-checkstyle/blob/master/config/import-control.xml
Sonar, has a custom audit listener which captures Audit Events.
https://github.com/checkstyle/sonar-checkstyle/blob/2f9d1a95aa97f30af243f2162198a8f43767a302/src/main/java/org/sonar/plugins/checkstyle/CheckstyleAuditListener.java#L95
We didn't change any of the method signatures, and the class isn't custom extended.

We did change the location of the `getSource` from the super to the main as the super went away. I am not sure the effect on the byte code and JRE connection, but this may break compatibility. This may require unknown 3rd parties who call this method to be source recompiled and there is no way around this. I think they will need to be source recompiled if they extend the changed classes in any way, or if they assign an instance of the changed class to a `Serializable` only type.

However, Sonar does not use this method and so can't be impacted if this did break compatibility. Even if we did break compatibility with Sonar, they always release a new JAR with every CS release and I don't think users are custom embedding new CS jars manually into Sonar or making 3rd parties on our 3rd party Sonar.

https://github.com/sevntu-checkstyle/sevntu.checkstyle/tree/master/sevntu-checkstyle-sonar-plugin
I don't know what this means for Sevntu's sonar as it is a few versions behind, but if main sonar isn't impacted I would think Sevntu is not either.

https://github.com/checkstyle/eclipse-cs/blob/783ad3c3b27c5c3956de788075d2c59366412767/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/Auditor.java#L309
Eclipse-CS is the same as Sonar in all ways mentioned above. They do not have an import control yet, but I wasn't see anything manually.

I did not see anyone accessing the Violation class directly, so it's change seem nominal.

romani (approved):
Thanks a lot for detailed reply.

I am ok to merge. we will update only second digit of our version as change should not affect if usage of our internals is not deep.

src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java
```diff
@@ -74,11 +71,20 @@ public AuditEvent(Object src, String fileName) {
      * @param violation the actual violation
      */
     public AuditEvent(Object src, String fileName, Violation violation) {
-        super(src);
+        source = src;
```
nrmancuso, line 74: We should continue to throw an exception here for a null source, as we did previously  (via `EventObject`).

nrmancuso (changes requested):
I also agree with minor version bump only.

Minor item:

src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java
rnveach: Done.

nrmancuso:
@rnveach failure at https://checkstyle.semaphoreci.com/jobs/24484e3d-790b-4885-a873-161e1fb8d6ce is legit

rnveach:
Fixed

nrmancuso (approved):
Good to merge as CI passes.

Issue #4930, opened 2017-08-10
LocalizedMessage should use ModuleName in compareTo method


romani:
base on discussion at https://github.com/checkstyle/checkstyle/pull/4899/commits/0dc6ed77dbc99c1b5934dd06415f367a89625eab#r131655519

LocalizedMessage.compareTo use ModuleId but does not use ModuleName ..... strange, ModuleName should be used

shashwat70:
I'm on it!

shashwat70:
Do we need to introduce a new property named ``` moduleName ``` ? @romani

rnveach:
@shashwat70 See the method `getSourceName()` in this class. The module name is the name of the class reporting the violation.

shashwat70:
Are moduleName and moduleID both supposed to be used in the compareTo method? Or do I need to replace moduleID with moduleName? @rnveach @romani

romani:
@shashwat70 ,

> Are moduleName and moduleID both supposed to be used in the compareTo method? 

yes. You meant `sourceName` instead `moduleName`.
We need both fields to be used in compare method.

we need to rename it to be `moduleName` to match naming convention with `moduleId` (could be done as separate issue, as it is API, we need to deprecate `getSourceName` method and keep it for some time before deletion).

shashwat70:
https://github.com/checkstyle/checkstyle/compare/master...shashwat70:compareto?expand=1
@romani @rnveach can you please take a look at these changes and let me know if this is what was expected? There are some issues with pitest. I'll send a PR as soon as I resolve them.

romani:
abandoned PR - https://github.com/checkstyle/checkstyle/pull/8192
everyone is welcome to reuse this code and send new PR to fix an issue.

rnveach:
> LocalizedMessage.compareTo use ModuleId but does not use ModuleName ..... strange

Adding `ModuleName` (more technically `sourceClass`) to comparison will mean 2 different modules who print the same violation message will no longer be seen as same and each one will be unique. One good example is `AbstractJavadocCheck#violateExecutionOnNonTightHtml`.

romani:
Fix  the s merged

PR #12676, opened 2023-01-22
Issue #4930: adds sourceClass to Violation compareTo method

rnveach:
Issue #4930

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
@@ -521,6 +521,7 @@ public class TreeWalkerTest extends AbstractModuleTestSupport {
 
         final String[] expected = {
             "6:9: " + getCheckMessage(WhitespaceAfterCheck.class, "ws.notFollowed", "if"),
+            "6:9: " + getCheckMessage(WhitespaceAroundCheck.class, "ws.notFollowed", "if"),
         };
 
         verify(treeWalkerConfig,
--- a/src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java
@@ -25,6 +25,8 @@ import static com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_OBJECT_ARRA
 import org.junit.jupiter.api.Test;
 import org.junitpioneer.jupiter.DefaultLocale;
 
+import com.puppycrawl.tools.checkstyle.checks.annotation.AnnotationLocationCheck;
+import com.puppycrawl.tools.checkstyle.checks.annotation.AnnotationOnSameLineCheck;
 import nl.jqno.equalsverifier.EqualsVerifier;
 import nl.jqno.equalsverifier.EqualsVerifierReport;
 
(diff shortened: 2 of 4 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
```diff
@@ -523,6 +523,7 @@ public void testMultiCheckOrder() throws Exception {
 
         final String[] expected = {
             "6:9: " + getCheckMessage(WhitespaceAfterCheck.class, "ws.notFollowed", "if"),
+            "6:9: " + getCheckMessage(WhitespaceAroundCheck.class, "ws.notFollowed", "if"),
```
rnveach, line 526: Example of https://github.com/checkstyle/checkstyle/issues/4930#issuecomment-1399412618
> Adding ModuleName (more technically sourceClass) to comparison will mean 2 different modules who print the same violation message will no longer be seen as same and each one will be unique.

src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java
```diff
@@ -131,6 +133,23 @@ public void testCompareToWithDifferentModuleId() {
                 .isTrue();
     }
 
+    @Test
+    public void testCompareToWithDifferentClass() {
```
romani, line 137: let keep having comments over pure unit tests to share why it is not possible to cover this by Input files.

romani (changes requested):
items:

src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java
rnveach: Technically it is possible, as shown by the change in test above. The issue is it becomes connected with those specific modules and I still don't feel pure utils should really be connected to modules.

rnveach:
@romani ping

src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java
romani: We try to ban all pure unit tests.
We allow them only as requirements for coverage and only because it is impossible to cover by actual execution.

In your tests, there is null based message - it is not possible, so ok to use pure unit tests with comment above test to explain reason.

But comparison of two messages from checks, should be covered by execution of both Checks or please share why not possible.

src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java
```diff
@@ -131,6 +133,23 @@ public void testCompareToWithDifferentModuleId() {
...
+        final Violation messageNull = createSampleViolationWithClass(null);
+
+        assertWithMessage("Invalid comparing result")
+                .that(message1.compareTo(messageNull) > 0)
+                .isTrue();
+        assertWithMessage("Invalid comparing result")
+                .that(messageNull.compareTo(message1) < 0)
+                .isTrue();
+        assertWithMessage("Invalid comparing result")
+                .that(message1.compareTo(message2) < 0)
```
romani, line 149: This should be possible to cover by two Check execution, right ?

romani (changes requested):
Items:

romani:
compareTo is used at:
```
	at com.puppycrawl.tools.checkstyle.api.Violation.compareTo(Violation.java:401)
	at com.puppycrawl.tools.checkstyle.api.Violation.compareTo(Violation.java:40)
	at java.base/java.util.TreeMap.compare(TreeMap.java:1291)
	at java.base/java.util.TreeMap.put(TreeMap.java:536)
	at java.base/java.util.TreeSet.add(TreeSet.java:255)
	at com.puppycrawl.tools.checkstyle.api.AbstractCheck.log(AbstractCheck.java:277)
	at com.puppycrawl.tools.checkstyle.checks.annotation.AnnotationLocationCheck.checkAnnotations(AnnotationLocationCheck.java:366)
	at com.puppycrawl.tools.checkstyle.checks.annotation.AnnotationLocationCheck.visitToken(AnnotationLocationCheck.java:336)
```

so it is used to keep our violation sorted.

src/test/java/com/puppycrawl/tools/checkstyle/api/ViolationTest.java
romani: ok,I see in code bunch of other tests like this.

romani (approved):
ok to merge on green CI

PR #12677, opened 2023-01-22
minor: replaces unncessary isTrue assertions with better ones

rnveach:
The `isTrue` were starting to annoy me.

The same cannot be done with `isFalse`. For example there is no `doesNotStartWith`.


romani (approved):
ok to merge

Issue #11055, opened 2021-12-19
AbstractExpressionHandler.getFirstAstNode might throw NPE


romani:
https://sonarcloud.io/project/issues?id=org.checkstyle%3Acheckstyle&issues=AXyeERcVIEIrBWM4sA6A&open=AXyeERcVIEIrBWM4sA6A

https://github.com/checkstyle/checkstyle/blob/e0aa61acf1f06e6ff0a1db0b90114e1becfdffbb/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java#L430-L434

disabled in Sonar until this issue, so remove Wontfix in Sonar after fix.

![image](https://user-images.githubusercontent.com/812984/146663340-65e11594-3493-4ef3-9417-60a2f6193b15.png)

theexplorist:
Hi @romani want to work on this but what is the expected fix here?

rnveach:
For these type of issues, we should confirm if an Exception is even possible. Do some regression to search for a case.

For this we need to identify which part of the code the exception is possible from.

If we can confirm that an Exception isn't possible logically and in regression, then we need to look to see how we can rewrite the code to hide the possible issue from the static scanner.

See how I handled a similar case in https://github.com/checkstyle/checkstyle/issues/11054 which is now resolved and closed.

theexplorist:
> For these type of issues, we should confirm if an Exception is even possible. Do some regression to search for a case.
> 
> For this we need to identify which part of the code the exception is possible from.
> 
> If we can confirm that an Exception isn't possible logically and in regression, then we need to look to see how we can rewrite the code to hide the possible issue from the static scanner.
> 
> See how I handled a similar case in #11054 which is now resolved and closed.

@rnveach Can you please shed some more light on regression if possible, I surely want to work on this!

rnveach:
> please shed some more light on regression 

https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#checkstyle-tester
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/README_MANUAL_EXECUTION.md#executing-diffgroovy
https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#checkstyle-pitest-regression

rnveach:
@romani How do you want to resolve this issue?

If `ast` coming into this method is `null`, then the result will be `null`.
[...]

PR #12680, opened 2023-01-24
Issue #11055: rewrites code to avoid possible NPE

rnveach:
Issue #11055

There is no surviving mutations on this, so a lot of our tests says it has to be like this. The code is identical to our TreeWalker code. The code is looking for a node by line/column, and since we don't keep our nodes ordered in source order, doing a random scan is really the only way. There are times we do go to null in this code. There are times were we going through multiple ASTs before we exit with our "answer". I don't see any "find by..." AST here then we already have.

If anything, having this as a do/while shows us we don't need to do the condition check first as long as regression shows differences.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
@@ -431,7 +431,7 @@ public abstract class AbstractExpressionHandler {
 
         DetailAST curNode = ast;
         DetailAST realStart = ast;
-        while (curNode != null) {
+        do {
             if (curNode.getLineNo() < realStart.getLineNo()
                     || curNode.getLineNo() == realStart.getLineNo()
                     && curNode.getColumnNo() < realStart.getColumnNo()) {
@@ -443,7 +443,7 @@ public abstract class AbstractExpressionHandler {
                 curNode = curNode.getParent();
             }
             curNode = toVisit;
-        }
+        } while (curNode != null);
         return realStart;
     }
 


rnveach:
Sonar: https://sonarcloud.io/summary/new_code?id=org.checkstyle%3Acheckstyle&pullRequest=12680

@romani How do we identify if this is better or resolves the possible NPE?

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
```diff
@@ -443,7 +443,7 @@ protected static DetailAST getFirstAstNode(DetailAST ast) {
                 curNode = curNode.getParent();
             }
             curNode = toVisit;
-        }
+        } while (curNode != null);
```
romani, line 446: This is same as before we need to avoid null at all if we can not, we should keep suppression as permanent

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
rnveach: We cannot, so lets just keep it as permanent.

There is no surviving mutations on this, so a lot of our tests says it has to be like this. The code is identical to our TreeWalker code. The code is looking for a node by line/column, and since we don't keep our nodes ordered in source order, doing a random scan is really the only way. I don't see any magical "find by..." ast here then we already have.

romani: Please copy this explanation to issue and let update sonar wantfix to "reason at #12680"

rnveach: > Please copy this explanation to issue 

Done.

> let update sonar wantfix to "reason at https://github.com/checkstyle/checkstyle/pull/12680"

Please do, I do not know how to update sonar.

romani:
ok , lets make this supresison to be permanent

PR #12684, opened 2023-01-28
Issue #12571: update all release action to have 'Setup local maven ca…

### LINKED ISSUE #12571: update all release action to have 'Setup local maven cache' step to optimize download of artifacts
all actions that executes `mvn` takes a lot of time to download jars .... we need to use maven cache as we do in other actions that works on PRs 

example: https://github.com/checkstyle/checkstyle/blob/bef49aea4a923bc003b0ffd1338b8f1606f2badc/.github/workflows/pitest.yml#L54

stoyanK7:
Resolves #12571

### DIFF
--- a/.github/workflows/bump-version-and-update-milestone.yml
+++ b/.github/workflows/bump-version-and-update-milestone.yml
@@ -30,6 +30,11 @@ jobs:
     steps:
       - name: Checkout the latest code
         uses: actions/checkout@v3
+      - name: Setup local maven cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: checkstyle-maven-cache-${{ hashFiles('**/pom.xml') }}
       - name: Modify File
         run: |
           ./.ci/bump-version.sh ${{ github.event.inputs.version }}
--- a/.github/workflows/checker-framework.yml
+++ b/.github/workflows/checker-framework.yml
@@ -40,6 +40,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Setup local maven cache
+        uses: actions/cache@v2
+        with:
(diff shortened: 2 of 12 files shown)


stoyanK7:
@romani In response to [your comment in the issue](https://github.com/checkstyle/checkstyle/issues/12571#issuecomment-1410405862)
> it was and issue at "release: prepare" https://github.com/checkstyle/checkstyle/actions/runs/4054790160/jobs/6977062039

Here are the results for `R: Maven Release Prepare ` in particular:
| Job | Time |
| :---:   | :---: |
| [First job without cache. Uploads cache after task is done](https://github.com/stoyanK7/checkstyle/actions/runs/4057026843/jobs/6982267658) | 3m 13s    |
| [Same job running for a second time with the cache from the first job](https://github.com/stoyanK7/checkstyle/actions/runs/4057076737/jobs/6982378354) | 1m 26s    |

romani:
If we do caching by pom.xml, it will be always different when we do version bump.
So there will be no benefits for release actions.

@nrmancuso , do you have any good ideas from your experience on this?

nrmancuso:
> If we do caching by pom.xml, it will be always different when we do version bump. So there will be no benefits for release actions.
> 
> @nrmancuso , do you have any good ideas from your experience on this?

I would say that it is good to create release with no cache, red CI from some bad package can be fixed whenever, but blocking release because we wanted to save 1-3 minutes is not good. 

But if you really want to use cache for release, just hash some other file.

romani (approved):
Ok to merge.

I have I my mind idea to create new file from pom.xml but removal of first 16 lines
https://github.com/checkstyle/checkstyle/blob/master/pom.xml
And making cache on such new file, for release related actions only. This speed up is good I think.

PR #12700, opened 2023-01-31
Issue #12694: Push 'all' jar action is not working and not failing execution

### LINKED ISSUE #12694: Push 'all' jar action is not working and not failing execution
https://github.com/checkstyle/checkstyle/actions/runs/4054967613/jobs/6977468997#step:5:207

`RELEASE_ID=$(curl -s -X GET -H "Authorization: token $GITHUB_TOKEN" \` somthing like this is required
we need to echo variables and add auth token to curl.

lets recheck all curls that we use in actions to have token

stoyanK7:
Resolves #12694

---
Adds `GITHUB_TOKEN` authorization header where it was missing. I checked all actions jobs and the `.sh` files they are calling.

---
The reason for [the main issue was different](https://github.com/checkstyle/checkstyle/issues/12694#issuecomment-1410750115) but I am adding the headers anyways to keep things consistent.

### DIFF
--- a/.ci/close-create-milestone.sh
+++ b/.ci/close-create-milestone.sh
@@ -8,6 +8,7 @@ checkForVariable "GITHUB_TOKEN"
 
 echo "Close previous milestone at github"
 MILESTONE_NUMBER=$(curl -s \
+                -H "Authorization: token $GITHUB_TOKEN" \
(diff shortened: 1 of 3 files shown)


romani (approved):
Ok to merge

Issue #11109, opened 2021-12-26
False Negative: UnnecessaryParenthesesCheck


Vyom-Yadav:
I have read check documentation: https://checkstyle.sourceforge.io/config_coding.html#UnnecessaryParentheses
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Test.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="TreeWalker">
        <module name="UnnecessaryParentheses"/>
    </module>
</module>

/var/tmp $ cat Test.java
public class Test {
    void method() {
        boolean a = true;
        boolean b = false;
        if(a || (b ^ a) || b) { // expected violation
        }
        if(a || (b | a) || b) { // expected violation
        }
        if(a || (b & a) || b) { // expected violation
        }
        int c = 1;
        int d = 12;
        if(a || (c ^ d) >> 1 == 1 || b) { // ok
        }
    }
}
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-9.2-all.jar -c config.xml Test.java
Starting audit...
Audit done.
```

---

**Describe what you expect in detail.**
```
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-9.2-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /home/vyom/IdeaProjects/ConfigCheckstyle/src/TestP/Test.java:5:17: Unnecessary parentheses around expression. [UnnecessaryParentheses]
[ERROR] /home/vyom/IdeaProjects/ConfigCheckstyle/src/TestP/Test.java:7:17: Unnecessary parentheses around expression. [UnnecessaryParentheses]
[ERROR] /home/vyom/IdeaProjects/ConfigCheckstyle/src/TestP/Test.java:9:17: Unnecessary parentheses around expression. [UnnecessaryParentheses]
Audit done.
Checkstyle ends with 3 errors.

```
[...]

Issue #11197, opened 2022-01-15
A false positive about the document case


Zustin:
I have read check documentation: https://checkstyle.sourceforge.io/config_naming.html#ClassTypeParameterName
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Test.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
    "https://checkstyle.org/dtds/configuration_1_3.dtd">
    
<module name="Checker">
  <property name="severity" value="error"/>
  <property name="fileExtensions" value="java, properties, xml"/>
  <module name="TreeWalker">
    <module name="ClassTypeParameterName">
      <property name="format" value="(^[A-Z][0-9]?)$|([A-Z][a-zA-Z0-9]*[T]$)"/>
    </module>
  </module>
</module>


/var/tmp $ cat Test.java
class MyClass1<T> {}        // violation
class MyClass2<t> {}        // violation
class MyClass3<abc> {}      // violation
class MyClass4<LISTENER> {} // violation
class MyClass5<RequestT> {} // OK

/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-X.XX-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /home/vanguard/projects/SAMutator/tools/Test.java:2:16: Name 't' must match pattern '(^[A-Z][0-9]?)$|([A-Z][a-zA-Z0-9]*[T]$)'. [ClassTypeParameterName]
[ERROR] /home/vanguard/projects/SAMutator/tools/Test.java:3:16: Name 'abc' must match pattern '(^[A-Z][0-9]?)$|([A-Z][a-zA-Z0-9]*[T]$)'. [ClassTypeParameterName]
[ERROR] /home/vanguard/projects/SAMutator/tools/Test.java:4:16: Name 'LISTENER' must match pattern '(^[A-Z][0-9]?)$|([A-Z][a-zA-Z0-9]*[T]$)'. [ClassTypeParameterName]
Audit done.
Checkstyle ends with 3 errors.
```
---
**Describe what you expect in detail.**
---
[...]

PR #12703, opened 2023-02-02
Issue #11109: False Negative: UnnecessaryParenthesesCheck2

Kevin222004:
Issue #11109: False Negative: UnnecessaryParenthesesCheck2

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/d7bafc525a9bbbd310d26fc1450d8c73/raw/f08834a6a50b6d149d81930a3ce5894a729d3c65/upara.xml

### DIFF
--- a/config/checkstyle-checks.xml
+++ b/config/checkstyle-checks.xml
@@ -528,7 +528,15 @@
     <module name="StringLiteralEquality"/>
     <module name="SuperClone"/>
     <module name="SuperFinalize"/>
-    <module name="UnnecessaryParentheses"/>
+    <module name="UnnecessaryParentheses">
+      <property name="tokens"
+                value="EXPR, IDENT, NUM_DOUBLE, NUM_FLOAT, NUM_INT, NUM_LONG, STRING_LITERAL,
+                       LITERAL_NULL, LITERAL_FALSE, LITERAL_TRUE, ASSIGN, BAND_ASSIGN, BOR_ASSIGN,
+                       BSR_ASSIGN, BXOR_ASSIGN, DIV_ASSIGN, MINUS_ASSIGN, MOD_ASSIGN, PLUS_ASSIGN,
+                       SL_ASSIGN, SR_ASSIGN, STAR_ASSIGN, LAMBDA, TEXT_BLOCK_LITERAL_BEGIN, LAND,
+                       LOR, LITERAL_INSTANCEOF, GT, LT, GE, LE, EQUAL, NOT_EQUAL, UNARY_MINUS,
+                       UNARY_PLUS, INC, DEC, LNOT, BNOT, POST_INC, POST_DEC, BOR, BXOR, BAND"/>
+    </module>
     <module name="UnnecessarySemicolonAfterOuterTypeDeclaration"/>
     <module name="UnnecessarySemicolonAfterTypeMemberDeclaration"/>
     <module name="UnnecessarySemicolonInEnumeration"/>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
@@ -222,6 +222,41 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * }
  * </pre>
  * <p>
+ * To configure the check to detect unnecessary parentheses around bitwise inclusive OR
+ * {@code '|'}, bitwise AND {@code '&amp;'}, bitwise exclusive OR {@code '^'}:
+ * </p>
+ * <pre>
+ * &lt;module name=&quot;UnnecessaryParentheses&quot;&gt;
+ *   &lt;property name=&quot;tokens&quot; value=&quot;BOR, BAND, BXOR&quot; /&gt;
+ * &lt;/module&gt;
+ * </pre>
+ * <pre>
+ * class Test {
+ *
+ *     void method() {
+ *         int x = 9, y = 8;
+ *         if(x&gt;= 0 ^ (x&lt;=8 &amp; y&lt;=11) // violation, unnecessary parenthesis
+ *             ^ y&gt;=8) {
+ *             return;
+ *         }
+ *         if(x&gt;= 0 ^ x&lt;=8 &amp; y&lt;=11 ^ y&gt;=8) { // ok
+ *             return;
+ *        }
+ *        if(x&gt;= 0 || (x&lt;=8 &amp; y&lt;=11) // violation, unnecessary parenthesis
+ *            &amp;&amp; y&gt;=8) {
+ *            return;
+ *        }
+ *        if(x&gt;= 0 || x&lt;=8 &amp; y&lt;=11 &amp;&amp; y&gt;=8) { // ok
+ *            return;
+ *        }
+ *        if(x&gt;= 0 &amp; (x&lt;=8 ^ y&lt;=11) &amp; y&gt;=8) { // ok
+ *            return;
+ *        }
+ *     }
+ *
+ * }
+ * </pre>
+ * <p>
  * Parent is {@code com.puppycrawl.tools.checkstyle.TreeWalker}
  * </p>
  * <p>
@@ -340,10 +375,14 @@ public class UnnecessaryParenthesesCheck extends AbstractCheck {
         TokenTypes.STAR_ASSIGN,
     };
 
-    /** Token types for conditional and relational operators. */
-    private static final int[] CONDITIONALS_AND_RELATIONAL = {
+    /** Token types for conditional operators. */
+    private static final int[] CONDITIONAL_OPERATOR = {
         TokenTypes.LOR,
         TokenTypes.LAND,
+    };
+
+    /** Token types for relation operator. */
+    private static final int[] RELATIONAL_OPERATOR = {
         TokenTypes.LITERAL_INSTANCEOF,
         TokenTypes.GT,
         TokenTypes.LT,
@@ -363,6 +402,13 @@ public class UnnecessaryParenthesesCheck extends AbstractCheck {
         TokenTypes.BNOT,
         TokenTypes.POST_INC,
         TokenTypes.POST_DEC,
+    };
+
+    /** Token types for bitwise binary operator. */
+    private static final int[] BITWISE_BINARY_OPERATORS = {
+        TokenTypes.BXOR,
+        TokenTypes.BOR,
+        TokenTypes.BAND,
     };
 
     /**
@@ -464,6 +510,9 @@ public class UnnecessaryParenthesesCheck extends AbstractCheck {
             TokenTypes.BNOT,
             TokenTypes.POST_INC,
             TokenTypes.POST_DEC,
+            TokenTypes.BXOR,
+            TokenTypes.BOR,
+            TokenTypes.BAND,
         };
     }
 
@@ -598,7 +647,7 @@ public class UnnecessaryParenthesesCheck extends AbstractCheck {
     }
 
     /**
-     * Checks if conditional, relational, unary and postfix operators
+     * Checks if conditional, relational, bitwise binary operator, unary and postfix operators
      * in expressions are surrounded by unnecessary parentheses.
      *
      * @param ast the {@code DetailAST} to check if it is surrounded by
@@ -608,20 +657,84 @@ public class UnnecessaryParenthesesCheck extends AbstractCheck {
      */
     private static boolean unnecessaryParenAroundOperators(DetailAST ast) {
         final int type = ast.getType();
-        final int parentType = ast.getParent().getType();
-        final boolean isConditional = TokenUtil.isOfType(type, CONDITIONALS_AND_RELATIONAL);
-        boolean result = TokenUtil.isOfType(parentType, CONDITIONALS_AND_RELATIONAL);
-        if (isConditional) {
-            if (type == TokenTypes.LOR) {
-                result = result && !TokenUtil.isOfType(parentType, TokenTypes.LAND);
-            }
-            result = result && !TokenUtil.isOfType(parentType, TokenTypes.EQUAL,
-                TokenTypes.NOT_EQUAL);
+        final boolean isConditionalOrRelational = TokenUtil.isOfType(type, CONDITIONAL_OPERATOR)
+                        || TokenUtil.isOfType(type, RELATIONAL_OPERATOR);
+        final boolean isBitwise = TokenUtil.isOfType(type, BITWISE_BINARY_OPERATORS);
+        final boolean hasUnnecessaryParentheses;
+        if (isConditionalOrRelational) {
+            hasUnnecessaryParentheses = checkConditionalOrRelationalOperator(ast);
+        }
+        else if (isBitwise) {
+            hasUnnecessaryParentheses = checkBitwiseBinaryOperator(ast);
         }
         else {
-            result = result && TokenUtil.isOfType(type, UNARY_AND_POSTFIX);
+            hasUnnecessaryParentheses = TokenUtil.isOfType(type, UNARY_AND_POSTFIX)
+                    && isBitWiseBinaryOrConditionalOrRelationalOperator(ast.getParent().getType());
         }
-        return result;
+        return hasUnnecessaryParentheses;
+    }
+
+    /**
+     * Check if conditional or relational operator has unnecessary parentheses.
+     *
+     * @param ast to check if surrounded by unnecessary parentheses
+     * @return true if unnecessary parenthesis
+     */
+    private static boolean checkConditionalOrRelationalOperator(DetailAST ast) {
+        final int type = ast.getType();
+        final int parentType = ast.getParent().getType();
+        final boolean isParentEqualityOperator =
+                TokenUtil.isOfType(parentType, TokenTypes.EQUAL, TokenTypes.NOT_EQUAL);
+        final boolean result;
+        if (type == TokenTypes.LOR) {
+            result = !TokenUtil.isOfType(parentType, TokenTypes.LAND)
+                    && !TokenUtil.isOfType(parentType, BITWISE_BINARY_OPERATORS);
+        }
+        else if (type == TokenTypes.LAND) {
+            result = !TokenUtil.isOfType(parentType, BITWISE_BINARY_OPERATORS);
+        }
+        else {
+            result = true;
+        }
+        return result && !isParentEqualityOperator
+                && isBitWiseBinaryOrConditionalOrRelationalOperator(parentType);
+    }
+
+    /**
+     * Check if bitwise binary operator has unnecessary parentheses.
+     *
+     * @param ast to check if surrounded by unnecessary parentheses
+     * @return true if unnecessary parenthesis
+     */
+    private static boolean checkBitwiseBinaryOperator(DetailAST ast) {
+        final int type = ast.getType();
+        final int parentType = ast.getParent().getType();
+        final boolean result;
+        if (type == TokenTypes.BOR) {
+            result = !TokenUtil.isOfType(parentType, TokenTypes.BAND, TokenTypes.BXOR)
+                    && !TokenUtil.isOfType(parentType, RELATIONAL_OPERATOR);
+        }
+        else if (type == TokenTypes.BXOR) {
+            result = !TokenUtil.isOfType(parentType, TokenTypes.BAND)
+                    && !TokenUtil.isOfType(parentType, RELATIONAL_OPERATOR);
+        }
+        // we deal with bitwise AND here.
+        else {
+            result = !TokenUtil.isOfType(parentType, RELATIONAL_OPERATOR);
+        }
+        return result && isBitWiseBinaryOrConditionalOrRelationalOperator(parentType);
+    }
+
+    /**
+     * Check if token type is bitwise binary or conditional or relational operator.
+     *
+     * @param type Token type to check
+     * @return true if it is bitwise binary or conditional operator
+     */
+    private static boolean isBitWiseBinaryOrConditionalOrRelationalOperator(int type) {
+        return TokenUtil.isOfType(type, CONDITIONAL_OPERATOR)
+                || TokenUtil.isOfType(type, RELATIONAL_OPERATOR)
+                || TokenUtil.isOfType(type, BITWISE_BINARY_OPERATORS);
     }
 
     /**
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesOperator.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesOperator.java
@@ -1,0 +1,111 @@
+/*
+UnnecessaryParentheses
+tokens = EXPR, IDENT, NUM_DOUBLE, NUM_FLOAT, NUM_INT, NUM_LONG, \
+         STRING_LITERAL, LITERAL_NULL, LITERAL_FALSE, LITERAL_TRUE, ASSIGN, \
+         BAND_ASSIGN, BOR_ASSIGN, BSR_ASSIGN, BXOR_ASSIGN, DIV_ASSIGN, \
+         MINUS_ASSIGN, MOD_ASSIGN, PLUS_ASSIGN, SL_ASSIGN, SR_ASSIGN, STAR_ASSIGN, \
+         LAMBDA, TEXT_BLOCK_LITERAL_BEGIN, LAND, LITERAL_INSTANCEOF, GT, LT, GE, \
+         LE, EQUAL, NOT_EQUAL, UNARY_MINUS, UNARY_PLUS, INC, DEC, LNOT, BNOT, \
+         POST_INC, POST_DEC, BOR, BAND, BXOR
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.unnecessaryparentheses;
+
+public class InputUnnecessaryParenthesesOperator {
+        boolean a = true;
+        boolean b = false;
+    void method() {
+        if(a || (b ^ a) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a || (b | a) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a || (b & a) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && (b ^ a) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && (b | a) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && (b & a) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && (b ^ a) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && (b | a) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && (b & a) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a || (b ^ a) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a || (b | a) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a || (b & a) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+    }
+
+    void method2() {
+        if(a || (((b ^ a))) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a || ((b | a)) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a || ((b & a)) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && (b ^ a) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && ((b | a)) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && (((b & a))) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && ((b ^ a)) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && (((b | a))) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a && ((b & a)) || b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a || ((b ^ a)) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a || (((b | a))) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+        if(a || ((b & a)) && b) { // violation 'Unnecessary parentheses around expression'
+        }
+        int x = 10, y = 38;
+        if(x>= 0 != (x<=8 ^ y<=11) && y>=8) {
+            return; // ok
+        }
+        if(x>= 0 != (x<=8 || y<=11) && y>=8) {
+            return; // ok
+        }
+        if(x== 0 != (x<=8 ^ y<=11) && y==8) {
+            return; // ok
+        }
+        boolean v = true, w = false;
+        if ((v & w) != // ok
+            (w | v)) {
+        }
+        if(x>= 0 | (x<=8 | y<=11) | y>=8) {
+            return; // violation above 'Unnecessary parentheses around expression'
+        }
+    }
+
+    void method3() {
+        int code1 = 9;
+        if ((code1 & 1) == 8) {}// ok
+        if ((code1 | 1) == 8) {} // ok
+        if ((code1 ^ 1) == 8) {} // ok
+
+        if ((code1 & 1) != 8) {} // ok
+        if ((code1 | 1) != 8) {} // ok
+        if ((code1 ^ 1) != 8) {} // ok
+    }
+    void method4() {
+        int a = 2;
+         if ((~a) < -27 // violation 'Unnecessary parentheses around expression'
+             && a-- < 30) { // ok
+            return;
+        }
+        if ((~a) <= -27 // violation 'Unnecessary parentheses around expression'
+             && a-- < 30) { // ok
+            return;
+        }
+    }
+}
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -647,7 +647,7 @@ no-error-orekit)
   checkout_from https://github.com/Hipparchus-Math/hipparchus.git
   cd .ci-temp/hipparchus
   # checkout to version that Orekit expects
-  SHA_HIPPARCHUS="fc""c4cd17097a92952a92ec00f755eb""fb99b02001"
+  SHA_HIPPARCHUS="815ad2bf9ce764e4498911d2145c49165f5f3333"
   git checkout $SHA_HIPPARCHUS
   mvn -e --no-transfer-progress install -DskipTests
   cd -
@@ -656,7 +656,7 @@ no-error-orekit)
   # no CI is enforced in project, so to make our build stable we should
   # checkout to latest release/development (annotated tag or hash) or sha that have fix we need
   # git checkout $(git describe --abbrev=0 --tags)
-  git checkout "dc""db3086ea33f7cb1582ed3f3618f0ba5e9a628f"
+  git checkout "a32b4629b2890fc198b19a95a714d67b87d7943d"
   mvn -e --no-transfer-progress compile checkstyle:check \
     -Dorekit.checkstyle.version="${CS_POM_VERSION}"
   cd ..
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheckTest.java
@@ -277,4 +277,65 @@ public class UnnecessaryParenthesesCheckTest extends AbstractModuleTestSupport {
         verifyWithInlineConfigParser(
                 getPath("InputUnnecessaryParenthesesIdentifier.java"), expected);
     }
+
+    @Test
+    public void testOperator1() throws Exception {
+        final String[] expected = {
+            "20:17: " + getCheckMessage(MSG_EXPR),
+            "22:17: " + getCheckMessage(MSG_EXPR),
+            "24:17: " + getCheckMessage(MSG_EXPR),
+            "26:17: " + getCheckMessage(MSG_EXPR),
+            "28:17: " + getCheckMessage(MSG_EXPR),
+            "30:17: " + getCheckMessage(MSG_EXPR),
+            "32:17: " + getCheckMessage(MSG_EXPR),
+            "34:17: " + getCheckMessage(MSG_EXPR),
+            "36:17: " + getCheckMessage(MSG_EXPR),
+            "38:17: " + getCheckMessage(MSG_EXPR),
+            "40:17: " + getCheckMessage(MSG_EXPR),
+            "42:17: " + getCheckMessage(MSG_EXPR),
+            "47:19: " + getCheckMessage(MSG_EXPR),
+            "49:18: " + getCheckMessage(MSG_EXPR),
+            "51:18: " + getCheckMessage(MSG_EXPR),
+            "53:17: " + getCheckMessage(MSG_EXPR),
+            "55:18: " + getCheckMessage(MSG_EXPR),
+            "57:19: " + getCheckMessage(MSG_EXPR),
+            "59:18: " + getCheckMessage(MSG_EXPR),
+            "61:19: " + getCheckMessage(MSG_EXPR),
+            "63:18: " + getCheckMessage(MSG_EXPR),
+            "65:18: " + getCheckMessage(MSG_EXPR),
+            "67:19: " + getCheckMessage(MSG_EXPR),
+            "69:18: " + getCheckMessage(MSG_EXPR),
+            "85:20: " + getCheckMessage(MSG_EXPR),
+            "102:14: " + getCheckMessage(MSG_EXPR),
+            "106:13: " + getCheckMessage(MSG_EXPR),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputUnnecessaryParenthesesOperator.java"), expected);
+    }
+
+    @Test
+    public void testOperator2() throws Exception {
+        final String[] expected = {
+            "66:18: " + getCheckMessage(MSG_EXPR),
+            "67:17: " + getCheckMessage(MSG_EXPR),
+            "68:25: " + getCheckMessage(MSG_EXPR),
+            "82:14: " + getCheckMessage(MSG_EXPR),
+            "83:19: " + getCheckMessage(MSG_EXPR),
+            "92:21: " + getCheckMessage(MSG_EXPR),
+            "95:19: " + getCheckMessage(MSG_EXPR),
+            "98:20: " + getCheckMessage(MSG_EXPR),
+            "101:21: " + getCheckMessage(MSG_EXPR),
+            "107:20: " + getCheckMessage(MSG_EXPR),
+            "110:21: " + getCheckMessage(MSG_EXPR),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputUnnecessaryParenthesesOperator2.java"), expected);
+    }
+
+    @Test
+    public void testOperator3() throws Exception {
+        final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
+        verifyWithInlineConfigParser(
+                getPath("InputUnnecessaryParenthesesOperator3.java"), expected);
+    }
 }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesOperator2.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesOperator2.java
@@ -1,0 +1,119 @@
+/*
+UnnecessaryParentheses
+tokens = EXPR, IDENT, NUM_DOUBLE, NUM_FLOAT, NUM_INT, NUM_LONG, \
+         STRING_LITERAL, LITERAL_NULL, LITERAL_FALSE, LITERAL_TRUE, ASSIGN, \
+         BAND_ASSIGN, BOR_ASSIGN, BSR_ASSIGN, BXOR_ASSIGN, DIV_ASSIGN, \
+         MINUS_ASSIGN, MOD_ASSIGN, PLUS_ASSIGN, SL_ASSIGN, SR_ASSIGN, STAR_ASSIGN, \
+         LAMBDA, TEXT_BLOCK_LITERAL_BEGIN, LAND, LITERAL_INSTANCEOF, GT, LT, GE, \
+         LE, EQUAL, NOT_EQUAL, UNARY_MINUS, UNARY_PLUS, INC, DEC, LNOT, BNOT, \
+         POST_INC, POST_DEC, BOR, BAND, BXOR
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.unnecessaryparentheses;
+
+public class InputUnnecessaryParenthesesOperator2 {
+
+    void method() {
+        boolean a = true;
+        boolean b = false;
+
+        if ((b || a) & // ok
+                (b || a)) {
+        }
+        if ((b && a) & // ok
+                (b && a)) {
+        }
+        if ((b || a) | // ok
+                (b && a)) {
+        }
+
+        int c = 1;
+        int d = 12;
+        if(a || (c ^ d) >> 1 == 1 || b) { // ok
+        }
+
+        if ((b || a) & // ok
+                (b || a)) {
+        }
+        if ((b || a) | // ok
+                (b || a)) {
+        }
+        if ((b || a) ^ // ok
+                (b || a)) {
+        }
+        if ((b && a) & // ok
+                (b && a)) {
+        }
+        if ((b && a) | // ok
+                (b && a)) {
+        }
+        if ((b && a) ^ // ok
+                (b && a)) {
+        }
+        if ((b || a) & // ok
+                (b && a)) {
+        }
+    }
+
+    public void test2() {
+        int a = 9;
+        int b = 10;
+        if (a > b) {
+            System.out.println("a is greater then b");
+        }
+        else if ((b < a) & ( // violation 'Unnecessary parentheses around expression'
+                (8>9) | // violation 'Unnecessary parentheses around expression'
+                        (!m3()))) { // violation 'Unnecessary parentheses around expression'
+             System.out.println();
+        }
+
+    }
+
+    boolean m3() {
+        return true;
+    }
+
+    void method3() {
+        int x = 10, y = 5, z = 5;
+
+        if (x == ((y<z) ? y : z) &
+            ((x>y & y>z) // violation 'Unnecessary parentheses around expression'
+                | (!(x<z) & y>z))) { // violation 'Unnecessary parentheses around expression'
+                return;
+        }
+        if(x>= 0 & (x<=8 | y<=11) & y>=8) { // ok
+            return;
+        }
+        if(x>= 0 ^ (x<=8 | y<=11) ^ y>=8) { // ok
+            return;
+        }
+        if(x>= 0 || (x<=8 & y<=11) && y>=8) {
+            return; // violation above 'Unnecessary parentheses around expression'
+        }
+        if(x> 0 & (x<=8 & y<=11) & y>=8) { // violation 'Unnecessary parentheses around expression'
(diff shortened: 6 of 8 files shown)


Kevin222004:
GitHub, generate report

Kevin222004:
Please start reivew

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -612,6 +627,13 @@ private static boolean unnecessaryParenAroundOperators(DetailAST ast) {
         final boolean isConditional = TokenUtil.isOfType(type, CONDITIONALS_AND_RELATIONAL);
         boolean result = TokenUtil.isOfType(parentType, CONDITIONALS_AND_RELATIONAL);
         if (isConditional) {
+            if (parentType == TokenTypes.BAND || parentType == TokenTypes.BXOR
+                    || parentType == TokenTypes.BOR) {
```
nrmancuso, line 631: Please create an array as we have done for `CONDITIONALS_AND_RELATIONAL`, let's call it `BITWISE_BINARY_OPERATORS`.  Then we can use `TokenUtil.isOfType` here.

nrmancuso: Code here gets weird as we consider operator precedence; please add a comment here about why parenthesis are unnecessary in this case.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesOperator.java
```diff
@@ -0,0 +1,75 @@
...
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.unnecessaryparentheses;
+
+public class InputUnnecessaryParenthesesOperator {
+        boolean a = true;
+        boolean b = false;
+    void method() {
```
nrmancuso, line 19: Please add cases with more than one set of parenthesis, like `(a || ((b ^ a)) || b)`

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesOperator.java
```diff
@@ -0,0 +1,75 @@
...
+         LAMBDA, TEXT_BLOCK_LITERAL_BEGIN, LAND, LITERAL_INSTANCEOF, GT, LT, GE, \
+         LE, EQUAL, NOT_EQUAL, UNARY_MINUS, UNARY_PLUS, INC, DEC, LNOT, BNOT, \
+         POST_INC, POST_DEC
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.unnecessaryparentheses;
+
+public class InputUnnecessaryParenthesesOperator {
```
nrmancuso, line 16: Please add test cases with relational operators, shifts, and arithmetic operators (both ok and violation). I suspect that we will have problems with these using the existing logic.

nrmancuso (changes requested):
Please review orekit failures and provide a summary of what we found there.

Items:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesOperator.java
Kevin222004: yes you are right the current logic is not good lots of the changes in the report and in orkiet are unexpected. i have started to studying the check once again

nrmancuso:
@Kevin222004 ping, let's continue here before opening new PRs (re: https://github.com/checkstyle/checkstyle/issues/11197#issuecomment-1483860269)

Kevin222004:
sure i just forget this

Kevin222004:
@nrmancuso Ci will fail because of to cyclomatic complexity but please start review for logic i am working on it

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesOperator.java
Kevin222004: Please look at new implementation.

I have to make if else(cyclomatic complexity 16) to switch(cyclomatic complexity 13) to reduce cyclomatic complexity .
I am trying to make it 10

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesOperator.java
Kevin222004: done

Kevin222004:
https://kevin222004.github.io/Regression/

romani:
@nrmancuso , please continue review

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -609,11 +628,26 @@ else if (assignDepth >= 1) {
...
         final int parentType = ast.getParent().getType();
-        final boolean isConditional = TokenUtil.isOfType(type, CONDITIONALS_AND_RELATIONAL);
-        boolean result = TokenUtil.isOfType(parentType, CONDITIONALS_AND_RELATIONAL);
-        if (isConditional) {
-            if (type == TokenTypes.LOR) {
-                result = result && !TokenUtil.isOfType(parentType, TokenTypes.LAND);
+        final boolean isConditionalOrBitwiseBinaryOperator =
+                isConditionalOrBitwiseBinary(type);
+        boolean result = isConditionalOrBitwiseBinary(parentType);
+        if (isConditionalOrBitwiseBinaryOperator) {
```
nrmancuso, line 634: Are we forced to couple checking of conditional and binary operators? Can we do something like:
```
if (isConditionalOperator) {
...
}
else if (isBitWiseBinaryOperator) {
...
}

```

nrmancuso (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: we  can make implemenation like 
```
    private static boolean unnecessaryParenAroundOperators(DetailAST ast) {
        final int type = ast.getType();
        final int parentType = ast.getParent().getType();
        final boolean isConditional = TokenUtil.isOfType(type, CONDITIONALS_AND_RELATIONAL);
        final boolean isBitwise = TokenUtil.isOfType(type, BITWISE_BINARY_OPERATORS);
        boolean result = isConditionalOrBitwiseBinary(parentType);
        if (isConditional) {
            if (type == TokenTypes.LOR) {
                result = result(parentType, TokenTypes.LAND)
                        && result(parentType, BITWISE_BINARY_OPERATORS);
            }
            if (type == TokenTypes.LAND) {
                result = result(parentType, BITWISE_BINARY_OPERATORS);
            }
            result = result && !TokenUtil.isOfType(parentType, TokenTypes.EQUAL,
                TokenTypes.NOT_EQUAL);
        }
        else if (isBitwise) {
             if (type == TokenTypes.BOR) {
                result = result(parentType, TokenTypes.BAND)
                        && result(parentType, TokenTypes.BXOR);
            }
            if (type == TokenTypes.BXOR) {
                result = result(parentType, TokenTypes.BAND);
            }
            result = result && !TokenUtil.isOfType(parentType, TokenTypes.EQUAL,
                TokenTypes.NOT_EQUAL);
        }
        else {
            result = result && TokenUtil.isOfType(type, UNARY_AND_POSTFIX);
        }
        return result;
    }
```

but  cyclomatic complexity is 12  here but allowed is 10. I have tried a lot but  
The way i implement(switch case) is the only way which i found

nrmancuso: Please do:
```
if (isConditionalOperator) {
    result = checkConditionalOperator(...);
}
else if (isBitWiseBinaryOperator) {
    result = checkBitWiseBinaryOperator(...);
}
else {
...
}

```

And remove other helper methods (like `result()` so that we can clearly see logic involved in each check.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -610,20 +629,67 @@ private static boolean unnecessaryParenAroundOperators(DetailAST ast) {
         final int type = ast.getType();
         final int parentType = ast.getParent().getType();
         final boolean isConditional = TokenUtil.isOfType(type, CONDITIONALS_AND_RELATIONAL);
-        boolean result = TokenUtil.isOfType(parentType, CONDITIONALS_AND_RELATIONAL);
+        final boolean isBitwise = TokenUtil.isOfType(type, BITWISE_BINARY_OPERATORS);
+        boolean result = TokenUtil.isOfType(parentType, CONDITIONALS_AND_RELATIONAL)
```
nrmancuso, line 633: `final boolean result;`, handle all logic in if/else if/ else

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -610,20 +629,67 @@ private static boolean unnecessaryParenAroundOperators(DetailAST ast) {
...
+    /**
+     * Check if Bitwise binary operator has unnecessary parentheses.
+     *
+     * @param ast to check if surrounded by unnecessary parentheses
+     * @return true if Unnecessary parenthesis
+     */
+    private static boolean checkBitwiseBinaryOperator(DetailAST ast) {
+        final int type = ast.getType();
+        final int parentType = ast.getParent().getType();
+        boolean result = TokenUtil.isOfType(parentType, CONDITIONALS_AND_RELATIONAL)
```
nrmancuso, line 679: Same as above, push all logic into if/else if/else

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -610,20 +629,67 @@ private static boolean unnecessaryParenAroundOperators(DetailAST ast) {
...
+    /**
+     * Check if conditional or logical operator has unnecessary parentheses.
+     *
+     * @param ast to check if surrounded by unnecessary parentheses
+     * @return true if Unnecessary parenthesis
+     */
+    private static boolean checkConditionalOperator(DetailAST ast) {
+        final int type = ast.getType();
+        final int parentType = ast.getParent().getType();
+        boolean result = TokenUtil.isOfType(parentType, CONDITIONALS_AND_RELATIONAL)
```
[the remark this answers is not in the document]
Kevin222004: DONE

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
nrmancuso: Why are we required to consider parent at all in this method for conditional/bitwise operators? We should handle all logic for these in `check...` method.

Kevin222004: I am sorry :) but i think i am missing something from https://github.com/checkstyle/checkstyle/pull/12703#discussion_r1148748480 to understand what changes you suggesting.  

i think you want me to remove `TokenUtil.isOfType(parentType, CONDITIONALS_AND_RELATIONAL || TokenUtil.isOfType(parentType, BITWISE_BINARY_OPERATORS`  am i right?

I have tried to not try to make any new logic for the token i have only enhance the current code to make not more confusion. the logic is same

nrmancuso: Current code (in master branch) in this method is not easy to understand, this is why we exceeded cyclomatic complexity with your changes. We need to push logic to check operators out to helper methods such that method reads like:
```
final boolean result;
if (isConditionalOperator) {
    result = checkConditionalOperator(...);
}
else if (isBitWiseBinaryOperator) {
    result = checkBitWiseBinaryOperator(...);
}
else {
// handle unary and postfix here
...
}
```

Instead of doing what I suggested, your changes to the method have some repetitive logic (checking of parents for bitwise and conditional operators), and couple checking of bitwise operators and conditional operators.

I am open to other ways to handle update to this method, but it must be easy to understand and reason about, which neither the existing code, nor the most recent changes in this PR are.

Kevin222004: please check again i have update the branch

Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -608,22 +627,67 @@ else if (assignDepth >= 1) {
...
+        if (type == TokenTypes.LOR) {
+            result = !TokenUtil.isOfType(parentType, TokenTypes.LAND)
+                    && !TokenUtil.isOfType(parentType, BITWISE_BINARY_OPERATORS);
+        }
+        else if (type == TokenTypes.LAND) {
+            result = !TokenUtil.isOfType(parentType, BITWISE_BINARY_OPERATORS);
+        }
+        return result && !TokenUtil.isOfType(parentType, TokenTypes.EQUAL, TokenTypes.NOT_EQUAL)
+                && (TokenUtil.isOfType(parentType, CONDITIONALS_AND_RELATIONAL)
+                || TokenUtil.isOfType(parentType, BITWISE_BINARY_OPERATORS));
```
nrmancuso, line 666: Please extract this to helper method, let's call it 'isBitWiseBinaryOrConditionalOperator`, we can use it in three places.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -608,22 +627,67 @@ else if (assignDepth >= 1) {
...
+     * @param ast to check if surrounded by unnecessary parentheses
+     * @return true if Unnecessary parenthesis
+     */
+    private static boolean checkBitwiseBinaryOperator(DetailAST ast) {
+        final int type = ast.getType();
+        final int parentType = ast.getParent().getType();
+        boolean result = true;
+        if (type == TokenTypes.BOR) {
+            result = !TokenUtil.isOfType(parentType, TokenTypes.BAND)
+                && !TokenUtil.isOfType(parentType, TokenTypes.BXOR);
```
nrmancuso, line 681: Please use varargs method, we can combine these method calls

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -608,22 +627,67 @@ else if (assignDepth >= 1) {
...
+        final int parentType = ast.getParent().getType();
+        boolean result = true;
+        if (type == TokenTypes.LOR) {
+            result = !TokenUtil.isOfType(parentType, TokenTypes.LAND)
+                    && !TokenUtil.isOfType(parentType, BITWISE_BINARY_OPERATORS);
+        }
+        else if (type == TokenTypes.LAND) {
+            result = !TokenUtil.isOfType(parentType, BITWISE_BINARY_OPERATORS);
+        }
+        return result && !TokenUtil.isOfType(parentType, TokenTypes.EQUAL, TokenTypes.NOT_EQUAL)
```
nrmancuso, line 664: `TokenUtil.isOfType(parentType, TokenTypes.EQUAL, TokenTypes.NOT_EQUAL)` -> extract to boolean value with name like `isParentEqualityOperator`

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -608,22 +627,78 @@ else if (assignDepth >= 1) {
      */
     private static boolean unnecessaryParenAroundOperators(DetailAST ast) {
         final int type = ast.getType();
-        final int parentType = ast.getParent().getType();
         final boolean isConditional = TokenUtil.isOfType(type, CONDITIONALS_AND_RELATIONAL);
-        boolean result = TokenUtil.isOfType(parentType, CONDITIONALS_AND_RELATIONAL);
+        final boolean isBitwise = TokenUtil.isOfType(type, BITWISE_BINARY_OPERATORS);
+        final boolean result;
```
nrmancuso, line 632: `result` -> `hasUnecessaryParenthesis`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -608,22 +627,78 @@ else if (assignDepth >= 1) {
...
+     *
+     * @param ast to check if surrounded by unnecessary parentheses
+     * @return true if Unnecessary parenthesis
+     */
+    private static boolean checkConditionalOperator(DetailAST ast) {
+        final int type = ast.getType();
+        final int parentType = ast.getParent().getType();
+        final boolean isParentEqualityOperator =
+                TokenUtil.isOfType(parentType, TokenTypes.EQUAL, TokenTypes.NOT_EQUAL);
+        boolean result = true;
```
nrmancuso, line 657: Can we do similar to above, i.e.:
```
final boolean result;
if (type == LOR) {
...
}
else if (type == LAND) {
...
}
else {
...
}
return result;
```
?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -608,22 +627,78 @@ else if (assignDepth >= 1) {
...
+     *
+     * @param ast to check if surrounded by unnecessary parentheses
+     * @return true if Unnecessary parenthesis
+     */
+    private static boolean checkBitwiseBinaryOperator(DetailAST ast) {
+        final int type = ast.getType();
+        final int parentType = ast.getParent().getType();
+        final boolean isParentEqualityOperator =
+                TokenUtil.isOfType(parentType, TokenTypes.EQUAL, TokenTypes.NOT_EQUAL);
+        boolean result = true;
```
nrmancuso, line 680: Same as above, can we do exhaustive if/else if/else?

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -608,22 +627,78 @@ else if (assignDepth >= 1) {
      */
     private static boolean unnecessaryParenAroundOperators(DetailAST ast) {
         final int type = ast.getType();
-        final int parentType = ast.getParent().getType();
         final boolean isConditional = TokenUtil.isOfType(type, CONDITIONALS_AND_RELATIONAL);
-        boolean result = TokenUtil.isOfType(parentType, CONDITIONALS_AND_RELATIONAL);
+        final boolean isBitwise = TokenUtil.isOfType(type, BITWISE_BINARY_OPERATORS);
+        final boolean result;
         if (isConditional) {
```
nrmancuso, line 633: `isConditional` -> `isConditionalOrRelational`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -608,22 +627,78 @@ else if (assignDepth >= 1) {
...
         return result;
     }
 
+    /**
+     * Check if conditional or logical operator has unnecessary parentheses.
+     *
+     * @param ast to check if surrounded by unnecessary parentheses
+     * @return true if Unnecessary parenthesis
+     */
+    private static boolean checkConditionalOperator(DetailAST ast) {
```
nrmancuso, line 652: `checkConditionalOrRelationalOperator`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
nrmancuso: Please leave comment in else branch that this is where we deal with bitwise AND

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: I think to change `boolean result = true` as `final boolean result;` is not possible 

and i think you suggested to change the name result to hasUnecessaryParenthesis

nrmancuso: >I think to change boolean result = true as final boolean result; is not possible

Please explain why

>and i think you suggested to change the name result to hasUnecessaryParenthesis

Only do this in the 'unnecessaryParenthesis` method, elsewhere we still do not know if parenthesis are unnecessary yet :)

Kevin222004: Ahh get it https://github.com/checkstyle/checkstyle/pull/12703#discussion_r1153999664 my mistake

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

romani:
GitHub, generate report

Kevin222004:
@romani the i don't see any issue in report overall

romani:
@nrmancuso , please finish review.

nrmancuso:
@Kevin222004 please review report again, we still have false positives like https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/04c2f97_2023174930/reports/diff/lombok-ast/xref/home/runner/work/checkstyle/checkstyle/contribution/checkstyle-tester/repositories/lombok-ast/src/printer/lombok/ast/printer/TextFormatter.java.html#L131

We need to be very careful, since not only is this a false positive, but removal of these parenthesis would cause code to not compile. Let's create a new test input and begin adding cases like this to it and work through issues. Ping me when you are ready for review again, since code is likely to change after this update.

Kevin222004:
@nrmancuso you'r right i need to be check once again(will take care from next time to not make this mistake).

Kevin222004:
I am seeing issue only with relational operator

Kevin222004:
GitHub, generate report

Kevin222004:
@nrmancuso can you please help here https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/925e6cf_2023064434/reports/diff/apache-struts/xref/home/runner/work/checkstyle/checkstyle/contribution/checkstyle-tester/repositories/apache-struts/core/src/main/java/org/apache/struts2/url/StrutsUrlDecoder.java.html#L86:~:text=86%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20b%20%3D%20(byte)%20((((convertHexDigit(bytes%5Bix%2B%2B%5D)%20%3C%3C%204)%20%26%200xff)%20%2B%20convertHexDigit(bytes%5Bix%2B%2B%5D)%20%26%200xff)%20%26%200xff)%3B
```
public class Test {
     public void method() {
        b = (byte) ((((convertHexDigit(bytes[ix++]) << 4) & 0xff) + convertHexDigit(bytes[ix++]) & 0xff) & 0xff);

    }
} 

<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
    
<module name="Checker">
    <module name="TreeWalker">
<module name="UnnecessaryParentheses"/>
  

  </module>
</module>

kevin@inspiron-15-5510:~/Desktop/check_style$ java -jar /home/kevin/checkstyle-10.9.4-SNAPSHOT-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /home/kevin/Desktop/check_style/Test.java:3:21: Unnecessary parentheses around expression. [UnnecessaryParentheses]
Audit done.
```
To me it is looking fine but i doubted this case that output might effect
is it fine?

Kevin222004:
@nrmancuso ready for review

nrmancuso (changes requested):
Report is looking good, but code is becoming more complicated than it might have to be. As we just check type of token, and type of parent, and check precedence, I do not think some of these details like which operator group a token belongs to is helping us to simplify comparisons. @Kevin222004 can we do something like:
```
    getOperatorPrecedence(int t) {
        final int precedence;
        switch (t) {
            case TokenTypes.GT:
            case TokenTypes.LT:
            case TokenTypes.GE:
            case TokenTypes.LE:
            case TokenTypes.LITERAL_INSTANCEOF:
                precedence = 9;
                break;
            case TokenTypes.EQUAL:
            case TokenTypes.NOT_EQUAL:
                precedence = 8;
                break;
...
            default: throw new IllegalArgument...
    }
    return precendence;
}

```
Then in unnecessaryParenAroundOperators, we could probably just do something like:
```
return getOperatorPrecedence(childType) > getOperatorPrecedence(parentType);

```

Does this make sense?

Kevin222004:
GitHub, generate report

Kevin222004:
This might be a little [error](https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=13970&view=logs&jobId=e663805c-ee13-5a24-f979-d725e0346c8e&j=d72e04f6-d294-5bfe-b3cd-ffcb523f40f9&t=a021e897-39dd-596d-375e-f941387b47d5#:~:text=%5BERROR%5D%20%5Bcheckstyle%5D%20%5BERROR%5D%20/Users/runner/work,7%27%20is%20a%20magic%20number.%20%5BMagicNumber%5D) but i dont know hoe to fix this :)

Kevin222004:
This might be a little [error](https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=13970&view=logs&jobId=e663805c-ee13-5a24-f979-d725e0346c8e&j=d72e04f6-d294-5bfe-b3cd-ffcb523f40f9&t=a021e897-39dd-596d-375e-f941387b47d5#:~:text=%5BERROR%5D%20%5Bcheckstyle%5D%20%5BERROR%5D%20/Users/runner/work,7%27%20is%20a%20magic%20number.%20%5BMagicNumber%5D) but i dont know hoe to fix this :)

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesIfStatement2.java
```diff
@@ -59,29 +59,29 @@ public void checkBooleanStatements() {
                         && (a // violation 'Unnecessary parentheses around expression'
                         && (b==c))) // violation 'Unnecessary parentheses around expression'
                 || (a // violation 'Unnecessary parentheses around expression'
-                || a!=(b<=c)) // ok
-                || (a==(b!=d==(c==b) && a!=(b<=c)))) { // violation 'parentheses around expression'
+                || a!=b<=c) // ok
+                || (a==(b!=d==(c==b) && a!=b<=c))) { // violation 'parentheses around expression'
                                                        // after '||'
```
Kevin222004, line 64: @nrmancuso All the implemenattion looks correct to me new report has no major difference then previous one.

All the new changes in new report then previous one are something similar to the code for i am leaving this comment in and they all are correct as per rule of operator precedence. we can remove this braces.

In the previous implementation, it is been not considered.  But know as per the new implementation we directly check the precedence of operators so i guess this may be bug which resolved by new implemnattion

nrmancuso: @Kevin222004 thanks a lot for finding this. You are correct, this should be a violation. However, I think this many changes in behavior in one PR/issue might be too crazy. I propose the following:
1. Let's go back to https://github.com/checkstyle/checkstyle/commit/925e6cf8f2acbbee30959031578ca10ef9c7f0f7, since we add only new functionality here
2. Let's open an issue for what you found above, and we can address this later.

From what I understand, this check has been controversial in the past for being too strict, so we should be careful in changing behavior too much. As you return to old commit, please update this documentation with a couple examples of new tokens: https://checkstyle.org/config_coding.html#UnnecessaryParentheses:~:text=The%20partial%20support%20for%20operator%20precedence%20includes%20cases%20of%20the%20following%20type%3A

Kevin222004: @nrmancuso i update the code as previous implementation but

 if i open the new issue regarding the new code changes then, then lets say if we expand the code as in https://github.com/checkstyle/checkstyle/commit/925e6cf8f2acbbee30959031578ca10ef9c7f0f7 then it might be very complex to understand.
 
I dont think their would be any other better way to implement this as it is right know 

>From what I understand, this check has been controversial in the past for being too strict, so we should be careful in changing behavior too much

I think we can make issue some how for selected operator like Bitwise for that we can remove parentheseis if more then 1 is their. because normally the presence of  operator like bitwise  is difficult to remember and also to having parantehsis is a good way rather than bad to understand code simple [please look] (https://github.com/CS-SI/Orekit/pull/32#:~:text=I%20understand%20it%20follows%20Java%20well%2Ddefined%20operator%20priority%20semantics%2C%20but%20these%20semantics%20are%20sometimes%20hard%20to%20remember%20for%20mere%20humans.%0AI%20usually%20use%20parentheses%20as%20a%20kind%20of%20in%2Dcode%20help%20to%20ease%20maintenance%2C%20so%20for%20me%2C%20these%20extra%20parentheses%20are%20rather%20good%20style%20than%20bad%20style.)

Kevin222004: https://github.com/checkstyle/checkstyle/pull/12703#discussion_r1158481390 done

Kevin222004: @nrmancuso ping

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesIfStatement2.java
```diff
@@ -49,8 +49,8 @@ public void checkBooleanStatements() {
         if ((b == c) == a
                 && (
                         ( // violation 'Unnecessary parentheses around expression'
-                            (b==c)==(d>=b)==a!=(c==d))
-                    || (b<=c)!=a==(c>=d))) {
+                            (b==c)==d>=b==a!=(c==d))
+                    || b<=c!=a==c>=d)) {
```
nrmancuso, line 53: @Kevin222004 can't we restore these now?

Kevin222004: yup will update

nrmancuso:
Github, generate site

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesIfStatement2.java
Kevin222004: @nrmancuso please let me know if doc update is good will make push for this change

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/UnnecessaryParenthesesCheck.xml
```diff
@@ -63,9 +63,13 @@
  }
  if ((++f) &amp;gt; g &amp;amp;&amp;amp; a) { // violation, unnecessary paren
  }
+ if ((a &amp;amp; d) | b) {} // unnecessary paren
```
nrmancuso, line 66: Please make these consistent  with other examples, like `// violation, unnecessary paren`

nrmancuso (changes requested):
Last item:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesIfStatement2.java
nrmancuso: Left last item from me below

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/UnnecessaryParenthesesCheck.xml
Kevin222004: done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesIfStatement2.java
Kevin222004: done

Kevin222004:
Please re-run semaphore

nrmancuso (approved):
Code is a bit verbose, but I think it is good to be clear and deliberate about our logic around operators and semi-support of operator precedence. Maybe the next reviewer can help to refine code a bit. This one was a lot of work, thanks a lot @Kevin222004 !

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -417,6 +438,9 @@ public int[] getDefaultTokens() {
             TokenTypes.BNOT,
             TokenTypes.POST_INC,
             TokenTypes.POST_DEC,
+            TokenTypes.BXOR,
+            TokenTypes.BOR,
+            TokenTypes.BAND,
```
romani, line 443: I am not sure that this is good default behavior.
Technically it is extra parentheses BUT compiler does not case and will optimize it as required, for human extra symbols do a lot of good job to show priority of execution. Nobody remember all this priorities, no need for human to even think of this, extra parentheses make it a way easier to read by all engineers.

extra parentheses are cheaper than mistakes caused by their absence.
At least we do not make behavior over strict by default.

@rnveach , @nrmancuso  , please vote.

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: @romani I also think default behavior is not good

nrmancuso: I would prefer this check to report all unnecessary parentheses and have behavior similar to the same IDEA inspection, and allow user to ignore/suppress/remove tokens as they see fit, but I can understand not making these default.

romani: Ok, I we think it's not good default behavior, let's remove token from default collection. User can extend if he wants to be strict.
We can make example how to make it, to let it quickly be found.

Kevin222004:
Github, generate site

Kevin222004:
Github, generate site

romani (approved):
ok to merge.

Thanks a lot for big update.

config/checkstyle-checks.xml
```diff
@@ -523,7 +523,15 @@
     <module name="StringLiteralEquality"/>
     <module name="SuperClone"/>
     <module name="SuperFinalize"/>
-    <module name="UnnecessaryParentheses"/>
+    <module name="UnnecessaryParentheses">
+      <property name="tokens"
+                value="EXPR, IDENT, NUM_DOUBLE, NUM_FLOAT, NUM_INT, NUM_LONG, STRING_LITERAL,
```
rnveach, line 528: I am not seeing where it was said the new tokens should not be added by default and why.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -88,6 +88,10 @@
  * }
  * if ((++f) &gt; g &amp;&amp; a) { // violation, unnecessary paren
  * }
+ * if ((a &amp; d) | b) {} // violation, unnecessary paren
```
rnveach, line 91: More curious, we are saying we have only partial support. Is there something missing for support for the tokens we are adding?

@romani Do we have something somewhere that says what full support is?

romani: Better to extend doc to define what we can not do

Kevin222004: can i change line no 70 as 
`The full support for operator precedence includes cases of the following type:`
and will remove the new line added and create a new section with text 
`The partial support for operator precedence includes cases of the following type:`
and add the example of new token

Kevin222004: look like
image

rnveach: I think it should say what the partial is not supporting. How am I to know as a user if my case is something not supported or is supported and just broken? One case I shouldn't report an issue since it isn't supported and the other I should report it since it is broken.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -222,6 +226,41 @@
  * }
  * </pre>
  * <p>
+ * To configure the check to detect unnecessary Parentheses around bitwise inclusive OR
```
rnveach, line 229: `Parentheses` => `parentheses`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -222,6 +226,41 @@
...
+ * {@code '|'}, bitwise AND {@code '&amp;'}, Bitwise exclusive OR {@code '^'}:
```
rnveach, line 230: `Bitwise` => `bitwise`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -607,21 +660,85 @@ else if (assignDepth >= 1) {
...
+        }
+        else {
+            hasUnnecessaryParentheses = TokenUtil.isOfType(type, UNARY_AND_POSTFIX)
+                    && isBitWiseBinaryOrConditionalOrRelationalOperator(ast.getParent().getType());
+        }
+        return hasUnnecessaryParentheses;
+    }
+
+    /**
+     * Check if conditional or logical operator has unnecessary parentheses.
```
rnveach, line 682: Shouldn't logical be relational? The method is named `OrRelational` but your javadoc says logical.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -607,21 +660,85 @@ else if (assignDepth >= 1) {
...
+     *
+     * @param ast to check if surrounded by unnecessary parentheses
+     * @return true if Unnecessary parenthesis
```
rnveach, line 685: `Unnecessary` => `unnecessary`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -607,21 +660,85 @@ else if (assignDepth >= 1) {
...
+        return result && isBitWiseBinaryOrConditionalOrRelationalOperator(parentType);
+    }
+
+    /**
+     * Check if token type is BitWise Binary Or Conditional or Relational Operator.
+     *
+     * @param type Token type to check
+     * @return true if it is BitWise Binary Or Conditional Operator
+     */
+    private static boolean isBitWiseBinaryOrConditionalOrRelationalOperator(int type) {
```
rnveach, line 738: `BitWise` => `Bitwise`

You did this casing in other areas.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -607,21 +660,85 @@ else if (assignDepth >= 1) {
...
+        }
+        // we deal with bitwise AND here.
+        else {
+            result = !TokenUtil.isOfType(parentType, RELATIONAL_OPERATOR);
+        }
+        return result && isBitWiseBinaryOrConditionalOrRelationalOperator(parentType);
+    }
+
+    /**
+     * Check if token type is BitWise Binary Or Conditional or Relational Operator.
```
rnveach, line 733: Same with casing here.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -607,21 +660,85 @@ else if (assignDepth >= 1) {
...
+     *
+     * @param type Token type to check
+     * @return true if it is BitWise Binary Or Conditional Operator
```
rnveach, line 736: Same.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: please give look

Kevin222004:
Github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
rnveach: I am still confused by the new text. It is not clear to me what you are saying is not supported under `Partial not supported for operator`. Are you showing false positives and false negatives? What is exactly not supported by this PR with the new tokens?
If there are things supported by this PR, an example or 2 should be added to `The partial support for operator precedence includes `.

I would expect things like: `violation not reported because we are not type aware` and `violation reported but this could fail if X was Y`.

config/checkstyle-checks.xml
rnveach: @Kevin222004 ping

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
Kevin222004: @rnveach  This pr should support for three new token but they are not going to be supported by default user have to add this token by themself.   

> I would expect things like: violation not reported because we are not type aware

it is already mentioned https://checkstyle.org/config_coding.html#UnnecessaryParentheses:~:text=The%20check%20is%20not%20%22type%20aware%22%2C%20that%20is%20to%20say%2C%20it%20can%27t%20tell%20if%20parentheses%20are%20unnecessary%20based%20on%20the%20types%20in%20an%20expression.%20The%20check%20is%20partially%20aware%20about%20operator%20precedence%20but%20unaware%20about%20operator%20associativity.%20It%20won%27t%20catch%20cases%20such%20as%3A 

violation reported but this could fail if X was Y.

I have added separate section for it and put the example to show what is violated 

I will change the new text some thing like  `The  partial support for operator precedence includes bitwise operator but they are not supported by default`
is it fine

rnveach: I am looking for a response at https://github.com/checkstyle/checkstyle/pull/12703#discussion_r1166947780 then

config/checkstyle-checks.xml
Kevin222004: as per https://github.com/checkstyle/checkstyle/pull/12703#discussion_r1164105414 it is decided to not support this three token by default so i have removed the token from getDefaultToken() and i have update this because it is showing me error if i dont add new token with all the existing one

Kevin222004: @rnveach please take look

Kevin222004: @rnveach ping

Kevin222004:
@rnveach please review

Kevin222004:
@rnveach ping

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -89,6 +89,15 @@
  * if ((++f) &gt; g &amp;&amp; a) { // violation, unnecessary paren
  * }
  * </pre>
+ * <p>
+ * Partial not supported for operator precedence includes cases of the following type:
```
rnveach, line 93: Probably a continuation of https://github.com/checkstyle/checkstyle/pull/12703#discussion_r1166953481

I am still not following what is "partially supported" and what is missing.

> I will change the new text some thing like The  partial support for operator precedence includes bitwise operator but they are not supported by default
is it fine

This is not what partial support means. As an analogy, you don't really say a car does not support driving until a key is turned in the ignition.

If you support everything your tokens does when enabled, then this paragraph needs to be removed. Otherwise, it needs to be updated specifically with what is and what is not supported.

https://github.com/checkstyle/checkstyle/pull/12703#discussion_r1167617023
> I would expect things like: `violation not reported because we are not type aware` and `violation reported but this could fail if X was Y`.

Kevin222004: >I am still not following what is "partially supported" and what is missing

I am sorry for this confusion.

> If you support everything your tokens does when enabled 

yes, in these three new tokens, support everything similar as https://docs.oracle.com/javase/tutorial/java/nutsandbolts/operators.html#:~:text=logical%20AND,%7C%7C
 in these three tokens, there is no operator associativity. that's why it is similar to logical or, logical and
 
 apart of this they wont support cases like https://checkstyle.org/config_coding.html#UnnecessaryParentheses:~:text=The%20check%20is%20not,if%20(p%20%3D%3D%20(q%20%3C%3D%20r))%20%7B%7D. 
 
 so i am removing the para

Kevin222004: @rnveach ping

rnveach (approved):

Issue #12705, opened 2023-02-02
execution of codenarc is failing on grapes very often


romani:
https://checkstyle.semaphoreci.com/jobs/a4f70d7c-9969-4e1e-917b-f13e3b735e25

execution of codenarc is failing on grapes very often
```
./.ci/codenarc.sh Failed in 00:06
WARNING: An illegal reflective access operation has occurred 00:01
WARNING: Illegal reflective access by org.codehaus.groovy.reflection.CachedClass (file:/usr/share/groovy/lib/groovy-2.4.17.jar) to method java.lang.Object.finalize() 00:01
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.reflection.CachedClass 00:01
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations 00:01
WARNING: All illegal access operations will be denied in a future release 00:01
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed: 00:06
General error during conversion: Error grabbing Grapes -- [download failed: org.slf4j#slf4j-api;1.7.25!slf4j-api.jar] 00:06
java.lang.RuntimeException: Error grabbing Grapes -- [download failed: org.slf4j#slf4j-api;1.7.25!slf4j-api.jar] 00:06
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) 00:06
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) 00:06
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) 00:06
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490) 00:06
	at org.codehaus.groovy.reflection.CachedConstructor.invoke(CachedConstructor.java:83) 00:06
	at org.codehaus.groovy.reflection.CachedConstructor.doConstructorInvoke(CachedConstructor.java:77) 00:06
	at org.codehaus.groovy.runtime.callsite.ConstructorSite$ConstructorSiteNoUnwrap.callConstructor(ConstructorSite.java:84) 00:06
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCallConstructor(CallSiteArray.java:59) 00:06
[...]

PR #12706, opened 2023-02-02
Issue 12705: remove .m2 for codenarc execution to make it more stable on download

romani:
Issue #12705

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -33,6 +33,7 @@ blocks:
 
         - name: codenarc analysis
           commands:
+            - rm -rf $HOME/.m2
             - ./.ci/codenarc.sh
 
         - name: ensure that all modules are used in no exception configs


.semaphore/semaphore.yml
```diff
@@ -33,6 +33,7 @@ blocks:
 
         - name: codenarc analysis
           commands:
+            - rm -rf $HOME/.m2
```
nrmancuso, line 36: I am not sure of cache implementation in semaphore, but this will probably affect other jobs. Plus, we are still wasting time in container to restore cache, just to delete it. It is better to create new "no cache" build like so and add codenarc job to it:

```yml
  - name: "Linux openjdk build (no cache)"
    task:
      prologue:
        commands:
          - checkout
          - sudo apt-get update
          - sudo apt-get install -y ant groovy xsltproc xmlstarlet
          - export GRADLE_OPTS=-Dorg.gradle.console=plain

```

nrmancuso (changes requested):
Item:

.semaphore/semaphore.yml
romani: It is few seconds, now we wasting more on mentors side to recheck CI status.

Semaphore has parallelization of 5 task, they must run them in separate containers.

We can keep issue open to do bigger update as you suggested.
I just tried to render some fix in 10 minutes that I got. And ease pain to recheck CI status.

nrmancuso: Ok, I added comment to issue, first priority is stable CI.

nrmancuso (approved):
Let's merge this

Issue #11637, opened 2022-05-15
Use Shellcheck to resolve shell script violations


Rahulkhinchi03:
ShellCheck is an open-source static analysis tool that automatically finds bugs in your shell scripts. (https://github.com/koalaman/shellcheck)

`./.ci/*sh` has all the shell scripts files and we have disabled all the violations at the `.shellcheckrc` file.
We are currently invoking spellcheck through Terminal (You can download it through apt or similar package managers)

TO-DO:
Pick a suppression (single suppression  fix in single Pull Request) To allocate a suppression update for you please make comment in the issue "I am on XxxxxSuppression"
Remove Your Selected Suppression from the `.shellcheckrc` file.
Run shellcheck through Terminal and See Errors
Fix the Errors and make a PR.

LIST OF SUPPRESSIONS:

- [ ] SC2086 # (info): Double quote to prevent globbing and word splitting.
- [ ] SC2006 # (style): Use $(...) notation instead of legacy backticks ....
- [ ] SC2046 # (warning): Quote this to prevent word splitting.
- [ ] SC2207 # (warning): Prefer mapfile or read -a to split command output.
- [ ] SC2002 # (style): Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
- [ ] SC2061 # (warning): Quote the parameter to -name so the shell won't interpret it.
- [ ] SC2035 # (info): Use ./glob or -- glob so names with dashes won't become options.
- [ ] SC2236 # (style): Use -n instead of ! -z.
- [ ] SC2016 # (info): Expressions don't expand in single quotes, use double quotes for that.
- [ ] SC2185 # (info): Some finds don't have a default path. Specify '.' explicitly.
- [ ] SC2155 # (warning): Declare and assign separately to avoid masking return values.
- [ ] SC2071 # (error): > is for string comparisons. Use -gt instead.
- [ ] SC2034 # (warning): RUN_JOB appears unused. Verify use (or export if used externally).
- [ ] SC2216 # (warning): Piping to 'true', a command that doesn't read stdin.
- [ ] SC2013 # (info): To read lines rather than words, pipe/redirect to a 'while read' loop.
- [ ] SC2206 # (warning): Quote to prevent word splitting/globbing.
- [ ] SC2143 # (style): Use grep -q instead of comparing output with [ -n .. ].
- [ ] SC2004 # (style): $/${} is unnecessary on arithmetic variables.
[...]

PR #12715, opened 2023-02-03
Issue #11637: Resolve shellcheck SC1004

Kevin222004:
part of :- #11637


romani (approved):
Spell check is in GitHub action, green

Issue #12591, opened 2023-01-05
Resolve surviving annotation mutations for pitest


rnveach:
Mutations:
https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-annotation-suppressions.xml

This issue will resolve the remaining suppression and remove the entire file. POM is not at 100% for this profile, so the profile needs to be updated.

This is blocked by https://github.com/hcoles/pitest/issues/1137 (target pitest release 1.10.5) as all the suppressions are basically false positives and cannot be resolved.

rnveach:
Issue is now unblocked with pitest bump, I will be resolving the issue.

nrmancuso:
Completed via #12719

PR #12719, opened 2023-02-04
Issue #12591: resolves annotation pitest suppressions

rnveach:
Issue #12591


src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/SuppressWarningsCheck.java
```diff
@@ -242,18 +242,15 @@ public void visitToken(final DetailAST ast) {
...
                                 break;
-                            // param in constant case
-                            // ex: public static final String UNCHECKED = "unchecked";
-                            // @SuppressWarnings(UNCHECKED)
-                            // or
-                            // @SuppressWarnings(SomeClass.UNCHECKED)
-                            case TokenTypes.IDENT:
-                            case TokenTypes.DOT:
-                                break;
                             default:
```
rnveach, line 245: Pitest will not handle cases where execution of cases match `default` case. It is basically required to merge cases with default. Since we use default, I see no reason to attach cases with default as it defeats the purpose of the default, which is all remaining items (unlisted).

I moved comments to default so information is not lost.

romani (approved):
Ok to merge

nrmancuso (approved):

PR #12725, opened 2023-02-05
Issue #12379: Checker false violation removals on other failures

### LINKED ISSUE #12379: Checker: false violation removals on other failures
Identified at https://github.com/checkstyle/checkstyle/actions/runs/3405402498/jobs/5663345165 ,

`xml-maven-plugin` failed and checker reported back all suppressions should be removed even though Checker never actually run.

We should skip anything else besides Checker to ensure groovy script behaves correctly.

romani:
fixes #12379

### DIFF
--- a/.ci/checker-framework.groovy
+++ b/.ci/checker-framework.groovy
@@ -112,7 +112,8 @@ private static int checkCheckerFrameworkReport(final String profile) {
  */
 private static List<List<String>> getCheckerFrameworkErrors(final String profile) {
     final List<String> checkerFrameworkLines = new ArrayList<>()
-    final String command = "mvn -e --no-transfer-progress clean compile -P${profile}"
+    final String command = "mvn -e --no-transfer-progress clean compile" +
+        " -P${profile},no-validations"
     final ProcessBuilder processBuilder = new ProcessBuilder(getOsSpecificCmd(command).split(' '))
     processBuilder.redirectErrorStream(true)
     final Process process = processBuilder.start()


.ci/checker-framework.groovy
```diff
@@ -112,7 +112,8 @@ private static int checkCheckerFrameworkReport(final String profile) {
  */
 private static List<List<String>> getCheckerFrameworkErrors(final String profile) {
     final List<String> checkerFrameworkLines = new ArrayList<>()
-    final String command = "mvn -e --no-transfer-progress clean compile -P${profile}"
+    final String command = "mvn -e --no-transfer-progress clean compile"
+        + " -P${profile},no-validations"
```
rnveach, line 116: CI is failing, `+` has to be on previous line.
https://github.com/checkstyle/checkstyle/actions/runs/4097261782/jobs/7065502375#step:6:15

rnveach (changes requested):

.ci/checker-framework.groovy
romani: Yes, we should disable this rule one day in future.

rnveach: My understanding was that this is something in groovy, not a check we turned on.
https://stackoverflow.com/a/5079914/1016482

rnveach: I rebased and fixed the code.

rnveach:
CI is all green.

https://github.com/checkstyle/checkstyle/actions/runs/4116348792/jobs/7106284737#step:6:64
Checker still ran and looks like it is outputting the expected information.

rnveach (approved):

.ci/checker-framework.groovy
romani: thanks a lot, I just pushed a fix and got surprised that I have no PR

Issue #12720, opened 2023-02-04
Make action to trigger all release steps


romani:
so we have all steps/phases of release automated but still required to be manually executed sequentially.
https://github.com/checkstyle/contribution/blob/master/releasenotes-builder/diagrams-manual-trigger.md

Ideally all of them should be auto triggered by themself.
One of the approach is: each action create git tag, other action triggered by tag creation.
https://github.com/checkstyle/contribution/blob/master/releasenotes-builder/diagrams.md

May be there is already some other way to execution actions one after another.

romani:
@stoyanK7 , this is very last issue to close a project. this is kind of binding of all actions to make release to be done by one manual action from maintainer.

stoyanK7:
@romani I already did some "research" on it yesterday. I had the idea to create one master workflow file to take care of all jobs.

GitHub allows reusing workflows: https://docs.github.com/en/actions/using-workflows/reusing-workflows#example-reusable-workflow

One limitation is that
> You can call a maximum of 20 reusable workflows from a single workflow file.

but from what I see, we fit in that number.

What do you think?

rnveach:
Just my thought if we have scripts fully automated (either from one or chain), we create an issue to do release and script executes from issue and posts statuses in issue until release is done, and it can report failures and such. Eventually we could have triggers do resumes from after failure point.

stoyanK7:
Just pasting what is possible: https://docs.github.com/en/actions/using-workflows/about-workflows#triggering-a-workflow

> Workflow triggers are events that cause a workflow to run. These events can be:
> - Events that occur in your workflow's repository
> - Events that occur outside of GitHub and trigger a repository_dispatch event on GitHub
> - Scheduled times
> - Manual

> For example, you can configure your workflow to run when a push is made to the default branch of your repository, when a release is created, or when an issue is opened.

romani:
I am in favor of simple and reliable actions, we do not need fancy reporting now. We just need it to work.

stoyanK7:
Already on it. Goes a bit slow since I need to do a lot of tweaks to point to my repo. I will open draft PR in the following days.

stoyanK7:
Since the release will be performed by one single workflow file triggered by a maintainer, can we rename the issue? Something in the lines of `Make action to trigger all release steps`?

romani:
renamed

stoyanK7:
[...]

PR #12741, opened 2023-02-14
Issue #12720: Make action to trigger all release steps

stoyanK7:
Resolves #12720

Continues from https://github.com/checkstyle/checkstyle/pull/12733. I can not reopen PR since it was from my `master` branch and I force-pushed. Probably there is a way to fix that but it is faster to open this one.

The first two commits are just for my convenience. We will remove them before merging. Look at commit `Issue #12720: Make action to trigger all release steps` to see the relevant changes to the issue.

Success run: https://github.com/stoyanK7/checkstyle/actions/runs/4177134639
Tags at my repo: https://github.com/stoyanK7/checkstyle/tags
Milestones: https://github.com/stoyanK7/checkstyle/milestones?state=closed
Commits: https://github.com/stoyanK7/checkstyle/commits/master
Sourceforge: https://stoyank7-checkstyle.sourceforge.io/

PR #12733, opened 2023-02-10
Issue #12720: Make action to be triggered by tag with name 'prepare/10.3.1'

stoyanK7:
Resolves #12720

---
The main file to look at is `release.yml`. I have deleted some files and changed others temporarily to be able to execute action in my repo.

---
Latest action: https://github.com/stoyanK7/checkstyle/actions/runs/4143720138

PR #12538, opened 2022-12-16
Issue #12079: Github action "Release perform"

stoyanK7:
Resolves #12079 

---
I cannot wrap my head around how any of this realease/deploy works and I am stuck. This is my progress until now. I feel like I'm a just doing trial and error hoping something works at this point.

Running
```
 ./.ci/maven-release-perform.sh 10.5.1
```
yields
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:39 min
[INFO] Finished at: 2022-12-16T19:13:34+01:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.sonatype.plugins:nexus-staging-maven-plugin:1.6.13:deploy
 (injected-nexus-deploy) on project checkstyle: Failed to deploy artifacts: Could not transfer artifact
 com.puppycrawl.tools:checkstyle:jar:test-sources:10.5.1-20221216.181333-36 
from/to sonatype-nexus-snapshots (https://oss.sonatype.org/content/repositories/snapshots/): 
Transfer failed for https://oss.sonatype.org/content/repositories/snapshots/com/puppycrawl
/tools/checkstyle/10.5.1-SNAPSHOT/checkstyle-10.5.1-20221216.181333-36-test-sources.jar 
401 Unauthorized -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal 
org.sonatype.plugins:nexus-staging-maven-plugin:1.6.13:deploy (injected-nexus-deploy) 
on project checkstyle: Failed to deploy artifacts: Could not transfer artifact 
com.puppycrawl.tools:checkstyle:jar:test-sources:10.5.1-20221216.181333-36 
from/to sonatype-nexus-snapshots (https://oss.sonatype.org/content/repositories/snapshots/):
 Transfer failed for https://oss.sonatype.org/content/repositories/snapshots/com/puppycrawl
/tools/checkstyle/10.5.1-SNAPSHOT/checkstyle-10.5.1-20221216.181333-36-test-sources.jar 401 Unauthorized

```

PR #12745, opened 2023-02-15
Issue #12720: Fix issues with maven-release-perform workflow

stoyanK7:
Part of #12720 

---
The maven-release-perform workflow was not tested as an action. Now that I understand better what is happening, a few things needed to be changed to make it work.

Proof that it works: https://github.com/stoyanK7/checkstyle/actions/runs/4183171423/jobs/7247190315

### DIFF
--- a/.ci/maven-release-perform.sh
+++ b/.ci/maven-release-perform.sh
@@ -17,17 +17,18 @@ CURRENT_VERSION=$(getCheckstylePomVersionWithoutSnapshot)
 echo CURRENT_VERSION="$CURRENT_VERSION"
 
 if [ "$TARGET_VERSION" != "$CURRENT_VERSION" ]; then
-  echo "Target Version and current Version do not match"
+  echo "[ERROR] Target Version and current Version do not match."
   exit 1;
 fi
 
 SKIP_TEST="-DskipTests -DskipITs"
 SKIP_CHECKSTYLE="-Dcheckstyle.ant.skip=true -Dcheckstyle.skip=true"
-SKIP_OTHERS="-Dpmd.skip=true -Dspotbugs.skip=true -Djacoco.skip=true -Dxml.skip=true"
+SKIP_OTHERS="-Dpmd.skip=true -Dspotbugs.skip=true -Djacoco.skip=true -Dxml.skip=true \
+  --no-transfer-progress"
 
 git checkout "checkstyle-$TARGET_VERSION"
 echo "Deploying jars to maven central (release:perform) ..."
 mvn -e --no-transfer-progress -Pgpg release:perform \
-  -DconnectionUrl=scm:git:git@github.com:checkstyle/checkstyle.git \
+  -DconnectionUrl=scm:git:https://github.com/checkstyle/checkstyle.git \
   -Dtag=checkstyle-"$TARGET_VERSION" \
   -Darguments="$SKIP_TEST $SKIP_CHECKSTYLE $SKIP_OTHERS"
--- a/.ci/prepare-settings.sh
+++ b/.ci/prepare-settings.sh
@@ -20,7 +20,7 @@ replace SONATYPE_PWD "$SONATYPE_PWD"
 replace GPG_PASSPHRASE "$GPG_PASSPHRASE"
 replace GPG_KEY "$GPG_KEY"
 
-
+mkdir -p ~/.m2
 TEMP_SETTING=./.ci-temp/release-settings.xml
 SETTING=~/.m2/settings.xml
 
@@ -34,4 +34,5 @@ if [ -f "$SETTING" ]; then
     echo "no backup is required as content is same"
   fi
 fi
+
 cp "$TEMP_SETTING" "$SETTING"
--- a/.github/workflows/maven-release-perform.yml
+++ b/.github/workflows/maven-release-perform.yml
@@ -16,22 +16,28 @@ concurrency:
 
 jobs:
   perform:
-    name: Release perform {{ inputs.version }}
+    name: Release perform ${{ inputs.version }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout the latest code
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          ref: master
       - name: Setup local maven cache
         uses: actions/cache@v2
         with:
           path: ~/.m2/repository
           key: checkstyle-maven-cache-${{ hashFiles('**/pom.xml') }}
-      - name: Set up JDK 11
-        uses: actions/setup-java@v3
-        with:
-          java-version: 11
+      - name: Configure GPG
+        env:
+          GPG_SECRET_KEY: ${{ secrets.GPG_SECRET_KEY }}
+        run: |
+          mkdir -p ~/.gnupg/
+          echo "$GPG_SECRET_KEY" > ~/.gnupg/private.key
+          gpg --batch --import ~/.gnupg/private.key
+          chmod 600 ~/.gnupg/private.key
+          chmod 700 ~/.gnupg
       - name: Prepare release settings
         env:
           SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
--- a/pom.xml
+++ b/pom.xml
@@ -759,6 +759,17 @@
         <version>3.1.0</version>
       </plugin>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-gpg-plugin</artifactId>
+        <version>3.0.1</version>
+        <configuration>
+          <gpgArguments>
+            <arg>--pinentry-mode</arg>
+            <arg>loopback</arg>
+          </gpgArguments>
+        </configuration>
+      </plugin>
+      <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>versions-maven-plugin</artifactId>
         <version>${maven.versions.plugin.version}</version>
--- a/config/jsoref-spellchecker/whitelist.words
+++ b/config/jsoref-spellchecker/whitelist.words
@@ -492,6 +492,7 @@ githubusercontent
 Gjs
 globbing
 GMetrics
+gnupg
 google
 googleapis
 googleblog
@@ -1031,6 +1032,7 @@ pguyot
 php
 picocli
 pid
+pinentry
 PIPESTATUS
 pitest
 pkga


.ci/maven-release-perform.sh
```diff
@@ -15,8 +15,8 @@ CURRENT_VERSION=$(xmlstarlet sel -N pom=http://maven.apache.org/POM/4.0.0 \
                              -t -m pom:project -v pom:version pom.xml | sed "s/-SNAPSHOT//")
 echo CURRENT_VERSION="$CURRENT_VERSION"
 
-if [ "$TARGET_VERSION" != "$CURRENT_VERSION" ]; then
-  echo "Target Version and current Version do not match"
+if [ "$TARGET_VERSION" == "$CURRENT_VERSION" ]; then
```
stoyanK7, line 18: We need to ensure that the versions are not the same, rather than the opposite since `release-prepare` bumps the version in the pom before that. Say we are releasing `10.7.1`, `release-prepare` would have bumped version in pom to `10.7.2` before that.

.ci/maven-release-perform.sh
```diff
@@ -26,7 +26,7 @@ SKIP_OTHERS="-Dpmd.skip=true -Dspotbugs.skip=true -Djacoco.skip=true -Dxml.skip=
 
 git checkout "checkstyle-$TARGET_VERSION"
 echo "Deploying jars to maven central (release:perform) ..."
-mvn -e --no-transfer-progress -Pgpg release:perform \
+mvn -e --no-transfer-progress -B release:perform \
```
stoyanK7, line 29: Batch mode `-B` needs to be used since download progress was present and was clogging the logs.

.github/workflows/maven-release-perform.yml
```diff
@@ -31,15 +35,24 @@ jobs:
...
+          server-password: SONATYPE_PWD
+          gpg-private-key: ${{ secrets.GPG_KEY }}
+          gpg-passphrase: GPG_PASSPHRASE
+      - name: Download xmlstarlet
+        run: |
+          sudo apt-get install xmlstarlet
+      - name: Setup SSH
+        uses: webfactory/ssh-agent@v0.7.0
+        with:
+          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
```
stoyanK7, line 51: We need to set up an SSH key for the `release:perform` because it uses `-DconnectionUrl=scm:git:git@github.com:checkstyle/checkstyle.git` to pull the repo.

.github/workflows/maven-release-perform.yml
```diff
@@ -31,15 +35,24 @@ jobs:
...
+      - name: Run Shell Script
         env:
           SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
           SONATYPE_PWD: ${{ secrets.SONATYPE_PWD }}
           GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
-          GPG_KEY: ${{ secrets.GPG_KEY }}
-        run: |
-          ./.ci/prepare-settings.sh
```
stoyanK7, line 42: No need to use that because `setup-java` takes care of preparing the GPG key and passphrase.

.ci/maven-release-perform.sh
stoyanK7: Also, `-Pgpg` mode should be excluded because `setup-java` takes care of that.

.github/workflows/maven-release-perform.yml
stoyanK7: We have a manual way of doing this in `copy-site-to-sourceforge.yml`:
https://github.com/checkstyle/checkstyle/blob/ab04b31535028b6afbe45e2c50fe458e5cb43328/.github/workflows/copy-site-to-sourceforge.yml#L25-L35

I don't know if you prefer using `webfactory/ssh-agent@v0.7.0` or doing it manually. Let me know so we can keep it consistent.

romani: Usage of this third-party action does not make code even smaller

.ci/maven-release-perform.sh
romani: Version much match, as we executing maven on source checkouted by tag, under tag there is strict version without snapshot.
If we want to improve, we can remove sed of snapshot removal

romani (changes requested):
See above

.ci/maven-release-perform.sh
stoyanK7: Oh my god, completely messed this up in my mind. I thought `release:perform` bumps to next version for some reason. Just double checked.

My bad, forget what I said. Fixed back. Done.

.github/workflows/maven-release-perform.yml
stoyanK7: Just a suggestion, why even use SSH? We can skip the SSH setup and just do
```
  -DconnectionUrl=scm:git:https://github.com/stoyanK7/checkstyle.git \
```
instead of what we do right now:
https://github.com/checkstyle/checkstyle/blob/e487327aef4fbb8f36a60a29d487375947d7a488/.ci/maven-release-perform.sh#L31

.ci/maven-release-perform.sh
stoyanK7: https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#publishing-using-apache-maven

The generated `$HOME/.m2/settings.xml` by the action would be:
```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
  <servers>
    <server>
      <id>maven</id>
      <username>${env.MAVEN_USERNAME}</username>
      <password>${env.MAVEN_CENTRAL_TOKEN}</password>
    </server>
    <server>
      <id>gpg.passphrase</id>
      <passphrase>${env.MAVEN_GPG_PASSPHRASE}</passphrase>
    </server>
  </servers>
</settings>
```

Extra setup that I did as well: https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#extra-setup-for-pomxml

The action takes care of importing the gpg key into the keychain.

The `GPG_KEY` secret should be exported as `gpg --armor --export-secret-keys YOUR_ID`: https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#gpg

Proof is the action itself `maven-gpg-plugin`: https://github.com/stoyanK7/checkstyle/actions/runs/4183171423/jobs/7247190315#step:7:3731

.github/workflows/maven-release-perform.yml
stoyanK7: Or do we want to stay with the ssh?

romani: we do copy to sourceforge, as it is or original , backup website and most used/indexed web side.
ssh is requied only because it is the only way I know how to upload zip to sourceforge.
it is not an operation of maven deploy.

.github/workflows/maven-release-perform.yml
romani: this script is good even for local laptop to setup environment.
We like CI to do same on our behalf. All CIs have problems and one day this will be required to be done manually on local.

stoyanK7: So you do not want to rely on setup-java at all, but rather scripts, got it.

romani: I would rather preserve knowledge on how to setup and CI scripts and config usually very good textual presentation of this that is never outdated in comparison to wiki pages.

stoyanK7: Done.

.github/workflows/maven-release-perform.yml
stoyanK7: Done.

.ci/prepare-settings.sh
```diff
@@ -20,7 +20,7 @@ replace SONATYPE_PWD "$SONATYPE_PWD"
 replace GPG_PASSPHRASE "$GPG_PASSPHRASE"
 replace GPG_KEY "$GPG_KEY"
 
-
+mkdir -p ~/.m2
```
stoyanK7, line 23: We assume that `~/.m2` directory exists but that is not the case on runners:

https://github.com/stoyanK7/checkstyle/actions/runs/4195984802/jobs/7276471446#step:5:10
```
cp: cannot create regular file '/home/runner/.m2/settings.xml': No such file or directory
```

.github/workflows/maven-release-perform.yml
```diff
@@ -16,22 +16,28 @@ concurrency:
...
         with:
           path: ~/.m2/repository
           key: checkstyle-maven-cache-${{ hashFiles('**/pom.xml') }}
-      - name: Set up JDK 11
-        uses: actions/setup-java@v3
-        with:
-          java-version: 11
+      - name: Configure GPG
+        env:
+          GPG_SECRET_KEY: ${{ secrets.GPG_SECRET_KEY }}
```
stoyanK7, line 34: You will have to export your secret key like this:
```sh
gpg --list-secret-keys
gpg --output private.pgp --armor --export-secret-key <email-here>
cat private.pgp
```
Then copy whole file contents(including `-----BEGIN PGP PRIVATE KEY BLOCK-----` and `-----END PGP PRIVATE KEY BLOCK-----`) and create new actions secret `GPG_SECRET_KEY`.

From: https://unix.stackexchange.com/a/482559

.github/workflows/maven-release-perform.yml
```diff
@@ -16,22 +16,28 @@ concurrency:
...
+        run: |
+          mkdir -p ~/.gnupg/
+          echo "$GPG_SECRET_KEY" > ~/.gnupg/private.key
+          gpg --batch --import ~/.gnupg/private.key
+          chmod 600 ~/.gnupg/private.key
+          chmod 700 ~/.gnupg
```
stoyanK7, line 40: These two are required because of ` gpg: WARNING: unsafe permissions on homedir`
 https://github.com/stoyanK7/checkstyle/actions/runs/4204292665/jobs/7294759223#step:6:1802

Fix from:  https://gist.github.com/oseme-techguy/bae2e309c084d93b75a9b25f49718f85

pom.xml
```diff
@@ -758,6 +758,17 @@
         <artifactId>maven-deploy-plugin</artifactId>
         <version>3.1.0</version>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-gpg-plugin</artifactId>
+        <version>3.0.1</version>
+        <configuration>
+          <gpgArguments>
+            <arg>--pinentry-mode</arg>
```
stoyanK7, line 767: Needed due to ` gpg: signing failed: Inappropriate ioctl for device`: https://github.com/stoyanK7/checkstyle/actions/runs/4204200206/jobs/7294547925#step:6:3125

Answer from: https://stackoverflow.com/a/53992951/9553927

stoyanK7:
Successful run: https://github.com/stoyanK7/checkstyle/actions/runs/4204312844/jobs/7294807962

stoyanK7:
@romani Successful run: https://github.com/stoyanK7/checkstyle/actions/runs/4204312844/jobs/7294807962

.ci/maven-release-perform.sh
stoyanK7: Returned back to `-Dgpg`. Done.

stoyanK7:
@romani ping

romani:
I am blocking this PR for now until we finish fully migration to automated actions.
I see this PR as just internal enhancements that I not agree with, I need to look at them after all other is working.

stoyanK7:
@romani I do not agree that these are internal enhancements. Please do not forget that when we created the action and merged it two weeks ago at https://github.com/checkstyle/checkstyle/pull/12538, it was not tested in GitHub actions, just locally.

Now that I am running the action in combination with the others in https://github.com/checkstyle/checkstyle/pull/12741, I noticed it has problems.

The important changes in this PR are:
- Moving the `-DconnectionUrl` parameter from SSH to HTTPS so we do not have to bother setting it(SSH) up.
  - When running this locally, SSH is already set up so we do not run into this issue, but GitHub actions does
- Configuring the private GPG key for signing. You need to add it to the GPG keychain before being able to sign.
  - Same as the above. GPG private key is on local machine, but it needs importing in Github actions.
- Configuring the `maven-gpg-plugin` to write the passphrase for us. If it is not configured, it fails: https://github.com/checkstyle/checkstyle/pull/12745#discussion_r1109840426

---
**The setup of the action is failing with what we have right now**: https://github.com/stoyanK7/checkstyle/actions/runs/4225401227/jobs/7337635648

.ci/maven-release-perform.sh
romani: @stoyanK7 , why we need `-B` ? if we have problems with too much logs , ` --no-transfer-progress` should care about this.

stoyanK7: https://stackoverflow.com/a/48929096/9553927
> Batch mode causes Maven to not display "Progress: 125/150kB" style lines when running. If you are running Maven on some server and then checking the logs afterwards, these progress lines take up 90% of the log and make it basically impossible to find the stuff that matters. Setting batch mode prevents this. Apart from that, I don't know any other use for batch mode. As others have said, I have never seen Maven prompt for anything during a build, regardless of whether interactive or batch mode is set.

---
https://github.com/checkstyle/checkstyle/pull/12733#issuecomment-1428406622

> Operation was cancelled so there is no reasoning. There is an excessive amount of downloads in the logs. Please ensure maven commands are turning off downloads, otherwise logs become hard to read. Also make sure you run in Batch mode to ensure it is not trying to get input from a "user".

---

> > 
> 
>     excessive amount of downloads in the logs
>     Batch mode
> 
> Looks like update needs to be made to 

https://github.com/checkstyle/checkstyle/pull/12733#issuecomment-1428425571

.github/workflows/maven-release-perform.yml
romani: usage of git by ssh , allows to avoid usage of passwords/tokens during push.
I use it on local a lot, I thought everyone is using it.
we do push to sources only in release:prepare not a release:perform so usage of http should be ok.

.ci/maven-release-perform.sh
stoyanK7: For some reason `--no-transfer-progress` isn't getting picked up. We also use batch mode in mvn:preprare.

romani:
@stoyanK7 , please reply to https://github.com/checkstyle/checkstyle/pull/12745#discussion_r1112184997 this looks like last items in this PR to resolve, all others are ok to merge.

.ci/maven-release-perform.sh
stoyanK7: Here are logs without `-B`: https://github.com/stoyanK7/checkstyle/actions/runs/4225686733/jobs/7338275777#step:6:16641

stoyanK7: @romani 
Done. Removed.

Uh, just found out that `--no-transfer-progess` needs to be added as an argument after `release:perform` to work. I was wondering why it was not working and hence added `-B`

Here is the latest run with the latest changes:
https://github.com/stoyanK7/checkstyle/actions/runs/4225875334/jobs/7338675470

romani: Strange , such argument is for mvn , not any other plugins, but it is ok to keep it as it works

romani (approved):
Ok to merge 
I am not resolving discussions as there are a lot of  valuable discussions

PR #12737, opened 2023-02-11
Issue #11338: FinalClassCheck should report private classes without c…

Kevin222004:
Issue #11338: FinalClassCheck should report private classes without constructor

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/2ce4726e677546482c436a000f4b01d9/raw/3faec59c1d2757c5398a5b325d32e43f12c9ca73/finalclass.xml

PR #12750, opened 2023-02-17
supplemental: make private class final with default constructor

Kevin222004:
supplemental: make private class final with default constructor

part of :- https://github.com/checkstyle/checkstyle/pull/12737

### DIFF
--- a/config/checker-framework-suppressions/checker-signature-gui-units-init-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-signature-gui-units-init-suppressions.xml
@@ -387,28 +387,28 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/gui/MainFrame.java</fileName>
     <specifier>call.ui</specifier>
     <message>Calling a method with UIEffect effect from a context limited to SafeEffect effects.</message>
-    <lineContent>private class ExpandCollapseAction extends AbstractAction {</lineContent>
+    <lineContent>private final class ExpandCollapseAction extends AbstractAction {</lineContent>
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/gui/MainFrame.java</fileName>
     <specifier>call.ui</specifier>
     <message>Calling a method with UIEffect effect from a context limited to SafeEffect effects.</message>
-    <lineContent>private class FileSelectionAction extends AbstractAction {</lineContent>
+    <lineContent>private final class FileSelectionAction extends AbstractAction {</lineContent>
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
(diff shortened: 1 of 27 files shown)


Kevin222004:
@romani please merge soon if possible

romani:
Please reference issue in tail of commit message.

@rnveach , we are changing API classes, should we do something to notify users?

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/finalclass/InputFinalClassClassWithPrivateCtorWithNestedExtendingClass.java
```diff
@@ -10,17 +10,18 @@
 public class InputFinalClassClassWithPrivateCtorWithNestedExtendingClass {
     class A {
         private A() {}
-        private class ExtendA extends A {}
+        private final class ExtendA extends A {}
```
rnveach, line 13: Inputs should not be modified.

rnveach (changes requested):

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/finalclass/InputFinalClassClassWithPrivateCtorWithNestedExtendingClass.java
Kevin222004: @rnveach that no issue i will remove it. after our new changes that is expected

rnveach:
Only way to notify users is by issue label. If we consider this a breaking change, it will need its own issue.

However, all of the classes are `private static`. They are technically internal classes, and if they were "extensible", we would be getting compile errors. Users outside of Checkstyle can't access a `private static` class without reflection, and I don't think you can extend with reflection.

> in this case why we need final on them

Because we are following the check which is active and will violate this going forward.

If you disagree with changes to check, then please discuss in the issue.
If you wish to not continue using this check, then please create a new issue.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/finalclass/InputFinalClassClassWithPrivateCtorWithNestedExtendingClass.java
Kevin222004: done

Kevin222004:
travis is not because of changes

Kevin222004:
@romani @rnveach please merge this so this https://github.com/checkstyle/checkstyle/pull/12737 you find easy to review

rnveach (approved):
I am fine to merge this without any special issue since all classes are private.

romani (approved):

PR #12763, opened 2023-02-22
Issue #12762: Assert variable is not empty in checkForVariable

stoyanK7:
From https://github.com/checkstyle/checkstyle/issues/12762#issuecomment-1439990111
> new problem is [here](https://github.com/checkstyle/checkstyle/actions/runs/4242897227/jobs/7374970451#step:6:1806):

Turns out that GitHub secrets are not set and environment variables are created with empty string as value but `checkForVariable` does not catch that:
https://github.com/checkstyle/checkstyle/actions/runs/4242897227/jobs/7374970451#step:5:4
### Before(`master`)
```
 stoyan @  zenbook:  ~/open-source/checkstyle                                                                                                                                       master
$ ./.ci/creation-of-issue-in-other-repos.sh                                                                                                                                         3:52:46 pm
Error: Define GITHUB_TOKEN environment variable
______________________________________________________________________________________________________________________________________________________________________________________________
 stoyan @  zenbook:  ~/open-source/checkstyle                                                                                                                                       master
$ export GITHUB_TOKEN=                                                                                                                                                              3:52:47 pm
______________________________________________________________________________________________________________________________________________________________________________________________
 stoyan @  zenbook:  ~/open-source/checkstyle                                                                                                                                       master
$ ./.ci/creation-of-issue-in-other-repos.sh                                                                                                                                         3:52:52 pm
version is not set
Usage: ./.ci/creation-of-issue-in-other-repos.sh <version>

```
Test action with nonexistent secret `GITHUB_TOKEN: ${{ secrets.NON_EXISTENT_SECRET_123 }}`
https://github.com/stoyanK7/checkstyle/actions/runs/4243962020/jobs/7377408744#step:5:8

---
### After(`this PR`)
```
 stoyan @  zenbook:  ~/open-source/checkstyle                                                                                                                  issue/12762-checkforVariale
$ ./.ci/creation-of-issue-in-other-repos.sh                                                                                                                                         3:51:22 pm
Error: Define GITHUB_TOKEN environment variable
______________________________________________________________________________________________________________________________________________________________________________________________
 stoyan @  zenbook:  ~/open-source/checkstyle                                                                                                                  issue/12762-checkforVariale
$ export GITHUB_TOKEN=                                                                                                                                                              3:51:26 pm
______________________________________________________________________________________________________________________________________________________________________________________________
 stoyan @  zenbook:  ~/open-source/checkstyle                                                                                                                  issue/12762-checkforVariale
$ ./.ci/creation-of-issue-in-other-repos.sh                                                                                                                                         3:51:36 pm
Error: Set value to GITHUB_TOKEN environment variable
______________________________________________________________________________________________________________________________________________________________________________________________
 stoyan @  zenbook:  ~/open-source/checkstyle                                                                                                                  issue/12762-checkforVariale
$ export GITHUB_TOKEN=a                                                                                                                                                             3:51:37 pm
______________________________________________________________________________________________________________________________________________________________________________________________
 stoyan @  zenbook:  ~/open-source/checkstyle                                                                                                                  issue/12762-checkforVariale
$ ./.ci/creation-of-issue-in-other-repos.sh                                                                                                                                         3:51:43 pm
version is not set
Usage: ./.ci/creation-of-issue-in-other-repos.sh <version>

```
Test action with nonexistent secret `GITHUB_TOKEN: ${{ secrets.NON_EXISTENT_SECRET_123 }}`
https://github.com/stoyanK7/checkstyle/actions/runs/4243984054/jobs/7377463718#step:5:7

### Documentation
https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html
> -v varname
> True if the shell variable varname is set (has been assigned a value).

> -z string
> True if the length of string is zero.

### DIFF
--- a/.ci/util.sh
+++ b/.ci/util.sh
@@ -12,6 +12,12 @@ function checkForVariable() {
     echo "Error: Define $1 environment variable"
     exit 1
   fi
+
+  VAR_VALUE="${!VAR_NAME}"
+  if [ -z "$VAR_VALUE" ]; then
+    echo "Error: Set not empty value to $1 environment variable"
+    exit 1
+  fi
 }
 
 function getMavenProperty {


.ci/util.sh
```diff
@@ -12,6 +12,12 @@ function checkForVariable() {
     echo "Error: Define $1 environment variable"
     exit 1
   fi
+
+  VAR_VALUE="${!VAR_NAME}"
+  if [ -z "$VAR_VALUE" ]; then
```
romani, line 17: Hm `if [ ! -v "$VAR_NAME" ];` , can you receched docs for it, is way we tried to do this, it use to work. Is it a nuances of GitHub initialization of variable ?

romani (changes requested):
Items

.ci/util.sh
stoyanK7: `-v` only checks if a variable is initialized, it does not care what value is given to the variable. I put documentation reference in PR description: https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html

Github actions uses an empty string if a secret is not set:
https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow
> If a secret has not been set, the return value of an expression referencing the secret (such as ${{ secrets.SuperSecret }} in the example) will be an empty string.

So the 
```yml
env:
   GPG_KEY: ${{ secrets.GPG_KEY }}
```
would be equal to
```sh
export GPG_KEY=""
```

The `if [ ! -v "$VAR_NAME" ];` statement would see that the variable is initialized but does not see that it is an empty string. Hence, we need an extra check.

romani: Ok, please make message `Set not empty value to ...`

nrmancuso: I propose we keep old `-v`check and add new `-z` check, with an error message that makes it very clear why check is failing. Github actions can be tricky to debug, any extra details will prove very helpful.

stoyanK7: Done.

romani: Yes, we keep both

romani (approved):
Ok to merge

PR #12777, opened 2023-02-27
Pull #12777: emphasize mutability of collectors

rickie:
Consider the Javadoc of `Collectors#toSet`:

> Returns a Collector that accumulates the input elements into a new Set. There are no guarantees on the type, mutability, serializability, or thread-safety of the Set returned; if more control over the returned Set is required, use `toCollection(Supplier)`. 

The occurrences that do need a guarantee on mutability are updated to use the `toCollection(Supplier)` collector. There is also a similar `toList` occurrence that is fixed.

Our custom Error Prone check, [CollectorMutability](https://error-prone.picnic.tech/bugpatterns/CollectorMutability/), that we have in [Error Prone Support](https://github.com/picnicSupermarket/error-prone-support/) identified these occurrences.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalCatchCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalCatchCheck.java
@@ -20,6 +20,7 @@
 package com.puppycrawl.tools.checkstyle.checks.coding;
 
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -154,7 +155,8 @@ public final class IllegalCatchCheck extends AbstractCheck {
     /** Specify exception class names to reject. */
     private final Set<String> illegalClassNames = Arrays.stream(new String[] {"Exception", "Error",
         "RuntimeException", "Throwable", "java.lang.Error", "java.lang.Exception",
-        "java.lang.RuntimeException", "java.lang.Throwable", }).collect(Collectors.toSet());
+        "java.lang.RuntimeException", "java.lang.Throwable", })
+            .collect(Collectors.toCollection(HashSet::new));
 
     /**
      * Setter to specify exception class names to reject.
--- a/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
@@ -581,7 +581,7 @@ public abstract class AbstractModuleTestSupport extends AbstractPathTestSupport 
     protected static String[] removeSuppressed(String[] actualViolations,
                                                String... suppressedViolations) {
         final List<String> actualViolationsList =
-            Arrays.stream(actualViolations).collect(Collectors.toList());
+            Arrays.stream(actualViolations).collect(Collectors.toCollection(ArrayList::new));
         actualViolationsList.removeAll(Arrays.asList(suppressedViolations));
         return actualViolationsList.toArray(CommonUtil.EMPTY_STRING_ARRAY);
     }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/JavaAstVisitorTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/JavaAstVisitorTest.java
@@ -19,6 +19,7 @@
 
 package com.puppycrawl.tools.checkstyle;
 
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.truth.Truth.assertWithMessage;
 
 import java.io.File;
@@ -36,6 +37,7 @@ import org.antlr.v4.runtime.CharStreams;
 import org.antlr.v4.runtime.CommonTokenStream;
 import org.junit.jupiter.api.Test;
 
+import com.google.common.collect.ImmutableSet;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.FileContents;
 import com.puppycrawl.tools.checkstyle.api.FileText;
@@ -107,14 +109,14 @@ public class JavaAstVisitorTest extends AbstractModuleTestSupport {
                 .map(Method::getName)
                 .collect(Collectors.toSet());
 
-        final Set<String> filteredVisitMethodNames = Arrays.stream(visitMethods)
+        final ImmutableSet<String> filteredVisitMethodNames = Arrays.stream(visitMethods)
                 .filter(method -> method.getName().contains("visit"))
+                // remove overridden 'visit' method from ParseTreeVisitor interface in
+                // JavaAstVisitor
+                .filter(method -> !"visit".equals(method.getName()))
                 .filter(method -> method.getModifiers() == Modifier.PUBLIC)
                 .map(Method::getName)
-                .collect(Collectors.toSet());
-
-        // remove overridden 'visit' method from ParseTreeVisitor interface in JavaAstVisitor
-        filteredVisitMethodNames.remove("visit");
+                .collect(toImmutableSet());
 
         final String message = "Visit methods in 'JavaLanguageParserBaseVisitor' generated from "
                 + "production rules and labeled alternatives in 'JavaLanguageParser.g4' should "
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
@@ -193,7 +193,7 @@ public class XpathRegressionTest extends AbstractModuleTestSupport {
         final Set<Class<?>> abstractJavadocCheckNames = CheckUtil.getCheckstyleChecks()
                 .stream()
                 .filter(AbstractJavadocCheck.class::isAssignableFrom)
-                .collect(Collectors.toSet());
+                .collect(Collectors.toCollection(HashSet::new));
         // add the extra checks
         abstractJavadocCheckNames.addAll(REGEXP_JAVADOC_CHECKS);
         final Set<String> abstractJavadocCheckSimpleNames =
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/CheckUtil.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/CheckUtil.java
@@ -87,7 +87,7 @@ public final class CheckUtil {
             }
 
             return name;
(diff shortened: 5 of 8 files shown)


romani:
Thanks a lot for fix 
Can we start using this custom rule in our build , we have it here https://github.com/Vyom-Yadav/checkstyle/blob/master/.github/workflows/error-prone.yml

Please change commit message to be `Pull #12777: ....` to preserve details on what and how was activated.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalCatchCheck.java
```diff
@@ -154,7 +155,8 @@ public final class IllegalCatchCheck extends AbstractCheck {
     /** Specify exception class names to reject. */
     private final Set<String> illegalClassNames = Arrays.stream(new String[] {"Exception", "Error",
         "RuntimeException", "Throwable", "java.lang.Error", "java.lang.Exception",
-        "java.lang.RuntimeException", "java.lang.Throwable", }).collect(Collectors.toSet());
+        "java.lang.RuntimeException", "java.lang.Throwable", })
+            .collect(Collectors.toCollection(HashSet::new));
```
rnveach, line 159: I don't use these type of methods much, so I am not really following. Can you explain what the difference or harm is from the old code?

> occurrences that do need a guarantee on mutability are updated 

Don't we want these to be immutable?

rickie:
Yes, enabling the custom rule is definitely possible. Though, it is important to note and explain that this rule can be considered as opinionated because its focus on immutability. In this PR I only updated the occurrences that are *mutable*, which is the second suggested fix of the custom rule. Enabling this rule would also result in rewriting all other usages to an  *immutable* variant, which is the first (and therefore default) suggested fix of the custom rule. This basically means that all other occurrences will be rewritten to `com.google.common.collect.ImmutableSet#toImmutableSet` (likewise for `List` and `Map`). This could result in quite some changes.  

We would be interested in your opinion on the rule and are definitely willing to enable it this repository.  

So to be complete, let me ask a few questions:  
- Are you also okay with the changes that rewrite collectors to the *immutable* variant? 
- Can we do both the applying, introducing Error Prone Support (with enabling the check) all in the same PR? I'm also fine with splitting those if necessary.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalCatchCheck.java
rickie: > I don't use these type of methods much, so I am not really following. Can you explain what the difference or harm is from the old code?  

In the cases that are updated in this PR we are making some assumptions about the mutability of the sets (and lists). To be precise: we are using a collector to create a set and later modify the set, so the set is mutated. The Javadoc of the previously used method clearly states that: "There are no guarantees on the type, mutability, serializability, or thread-safety of the Set returned; if more control over the returned Set is required, use `toCollection(Supplier)`."   
    
Since we want to have more control over the returned set (because we need to have the guarantee that the set is mutable), we need to rewrite this code to the `toCollection(Supplier)`.  

There are some [extra examples on the documentation](https://error-prone.picnic.tech/bugpatterns/CollectorMutability/) of this check for some examples and its suggested fixes.

rnveach: Thanks for the explanation.

> There are no guarantees on ... mutability, 

Now I am just curious. What causes it to change mutability for the old method call? We have not had any issues with this code and I am sure it has been around for months. Is this just JRE implementation specific?

rickie: Hmm, I'm not sure I can answer that 🤔. I'd suggest looking at the actual [implementation](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/stream/Collectors.java#L289) of the method, maybe that explains it?

rnveach: Alright. ;)
Thanks anyways.

romani: Even in most cases such updates looks like more verbose but more exact.
the  most obvious reason to be verbose is to at least demand engineer to decide on mutable vs immutable. That part is good.

In our project we do not benefit much with different types of Sets, Lists, so in most cases it will be default HashSet, ArrayList, HashMap. But think it is still good to think on type of collection.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
```diff
@@ -581,7 +581,7 @@ private static String getMessageBundle(String className) {
     protected static String[] removeSuppressed(String[] actualViolations,
                                                String... suppressedViolations) {
         final List<String> actualViolationsList =
-            Arrays.stream(actualViolations).collect(Collectors.toList());
+            Arrays.stream(actualViolations).collect(Collectors.toCollection(ArrayList::new));
         actualViolationsList.removeAll(Arrays.asList(suppressedViolations));
```
romani, line 585: unrelated but actualViolationsList should  be Set.

src/test/java/com/puppycrawl/tools/checkstyle/JavaAstVisitorTest.java
```diff
@@ -111,7 +112,7 @@ public void testAllVisitMethodsAreOverridden() {
                 .filter(method -> method.getName().contains("visit"))
                 .filter(method -> method.getModifiers() == Modifier.PUBLIC)
                 .map(Method::getName)
-                .collect(Collectors.toSet());
+                .collect(Collectors.toCollection(HashSet::new));
 
         // remove overridden 'visit' method from ParseTreeVisitor interface in JavaAstVisitor
         filteredVisitMethodNames.remove("visit");
```
romani, line 118: we can filter it out in chain calls above and make it immutable set.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
```diff
@@ -193,7 +193,7 @@ public void validateIncompatibleJavadocCheckNames() throws IOException {
         final Set<Class<?>> abstractJavadocCheckNames = CheckUtil.getCheckstyleChecks()
                 .stream()
                 .filter(AbstractJavadocCheck.class::isAssignableFrom)
-                .collect(Collectors.toSet());
+                .collect(Collectors.toCollection(HashSet::new));
         // add the extra checks
         abstractJavadocCheckNames.addAll(REGEXP_JAVADOC_CHECKS);
```
romani, line 198: we can make abstractJavadocCheckNames immutable and add it to mutable collection that is initialized by REGEXP_JAVADOC_CHECKS

src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/CheckUtil.java
```diff
@@ -87,7 +87,7 @@ public static Set<String> getSimpleNames(Set<Class<?>> checks) {
             }
 
             return name;
-        }).collect(Collectors.toSet());
+        }).collect(Collectors.toCollection(HashSet::new));
```
romani, line 90: we can try immutbale here

romani (changes requested):
let use this rule.

>Are you also okay with the changes that rewrite collectors to the immutable variant?

yes, most valuable consideretion during code writing.

>all in the same PR? I'm also fine with splitting those if necessary.

activation in separate PR is ok, we just want to be sure we will not degrade in future changes. 

items:

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
rnveach: I am not sure it can be since it is a String. I think order matters, as it is the order of violations. Other methods are list too. I think this should be a separate PR to confirm.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
rickie: The implementation of `REGEXP_JAVADOC_CHECKS` uses `Set#of`. This is the Javadoc of `Set#of`:

>  Returns an unmodifiable set containing four elements. See Unmodifiable Sets for details.
....
Throws:
IllegalArgumentException – if there are any duplicate elements
NullPointerException – if an element is null
Since:
9 static <E> Set<E> of(E e1, E e2, E e3, E e4) {
        return new ImmutableCollections.SetN<>(e1, e2, e3, e4);
    }

So we cannot add the `abstractJavadocCheckNames` to that collection. Should we update the implementation of `REGEXP_JAVADOC_CHECKS` or should we combine the two sets in a way?

src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/CheckUtil.java
rickie: This wouldn't work because `AllChecksTest#testAllModulesAreReferencedInConfigFile` will mutate the collection right?

rickie:
> let use this rule.

Nice! Let's do it.

> activation in separate PR is ok, we just want to be sure we will not degrade in future changes.

Okay, will create a PR in the coming days! 

---

Added a commit and responded to the comments @romani 😄.

rickie:
Rebased.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
romani: ok, one unmodifiable is enough, so ok to keep `toCollection(HashSet::new)`

src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/CheckUtil.java
romani: ok, I see that tests use a lot returned collection to add/remove elements from it.
let keep it HashSet

romani:
@rickie , code is good, lets finalize polishing of code to pass CI.

rickie:
Thanks for the suggestions. Squashed the commits and fixed the violation! Should be good now @romani 😄. 

I will look into enabling the rule we discussed ~soon.

romani:
Please make commit message like `Pull #12777: ...` to preserve link to this conversation as it is not very obvious why we are doing this.

romani (approved):
Ok to merge

rnveach (approved):

Issue #7587, opened 2020-02-23
Update doc for DescendantToken


timurt:
https://checkstyle.org/config_misc.html#DescendantToken

Add example for each existing config to show what code is ok and what code cause violation.
Good example is https://checkstyle.org/config_naming.html#AbstractClassName

please play with module/Check by means of https://checkstyle.org/cmdline.html to understand how it works.

DXTkastb:
I'm on it.

timurt:
issue can be taken by anyone

harshit82:
@romani @timurt
I am a newbie to open source. I am interested to work on this issue but need some guidance.
Can I work on this?

nmancus1:
@harshit82 you are welcome to send a PR, please read through all details at https://checkstyle.org/beginning_development.html to set up the project, and take time to read through https://github.com/checkstyle/checkstyle/issues/7587#issue-569522224.

harshit82:
@romani @nmancus1 can you explain what exactly do I need to do. Which files do I need to update and what changes do I need to make? Is it DescendantTokenCheck.xml or DescendantTokenCheck.java?
I have gone through all the links provided.

romani:
>Is it DescendantTokenCheck.xml or DescendantTokenCheck.java?

Yes.

ShubhraSuman:
Can I work on this one?

romani:
You can work on all issues that does not have comment "I am in it" from some other people.

Please do not comment on a few issues. Choose one and focus on it , do not block others to pickup issue.

ConnectBhawna:
I would like to work on this issue.
Can you please assign it to me :)

romani:
No assignments, just make comment "I am in it"

ConnectBhawna:
Okk sure thanks @romani . If i need any help i will contact you :)

romani:
no PR in 3 weeks, this issue is open for all to be fixed.

athuma:
Hi @romani i am new to open source and would like to contribute i am familiar with java but i don't know where to start can you please help me how to start?

romani:
Please read https://github.com/checkstyle/checkstyle/wiki/Checkstyle-GSoC-2023-Project-Ideas

athuma:
@romani i have gone through it and i think i can do spellcheck of Identifiers by english dictionary project will be submitting my application through GSOC thanks for your help

athuma:
i am working on this issue

[...]

PR #12791, opened 2023-03-03
Example added for Switch with no default case in Descendant Token

athuma:
Issue #7587 : add code examples


romani (commented):
This Check is unusual, I doubt we can use same java code to cover all properties, as some of them are very specific to tokens.

We just need to add java examples under each config to show code structure with and without violation.
Some example configs property values need to be updated to have counts to be 1-3 to not make huge java code snippet.

athuma:
So should i add example for each config and then raise a PR? @romani

romani:
Update all in this PR, single commit

PR #12793, opened 2023-03-03
Pull #12793: Remove execution of upload -all jar from new-milestone-and-issue-in-other-repo

stoyanK7:
Removes execution of `upload-all-jar.sh` from `new-milestone-and-issue-in-other-repo.yml`. The workflow was doing too much and the upload of `-jar` is an important part of the release process. We already have a separate workflow for uploading all jar:
https://github.com/checkstyle/checkstyle/blob/master/.github/workflows/upload-all-jar.yml

After merging PR, https://github.com/checkstyle/checkstyle/wiki/Semi-automated-release has to be updated.

### DIFF
--- a/.github/workflows/new-milestone-and-issue-in-other-repo.yml
+++ b/.github/workflows/new-milestone-and-issue-in-other-repo.yml
@@ -2,8 +2,8 @@
 # GitHub Action to Create New Milestone and issue in Other Repositories.
 #
 #############################################################################
-name: "R: Close/Create Milestone, Upload '-all' jar, Create issues"
-run-name: "R: Close/Create Milestone, Upload '-all' jar, Create issues ${{ inputs.version }}"
+name: "R: Close/Create Milestone, Create issues"
+run-name: "R: Close/Create Milestone, Create issues ${{ inputs.version }}"
 on:
   workflow_dispatch:
     inputs:
@@ -50,12 +50,6 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
         run: |
           ./.ci/close-create-milestone.sh
-
-      - name: Upload '-all' jar
-        env:
-          GITHUB_TOKEN: ${{ github.token }}
-        run: |
-          ./.ci/upload-all-jar.sh ${{ inputs.version }}
 
       - name: Creation of issue in other Repositories
         env:


romani (approved):
Ok to merge

Better to not mix beautification of issue tracking and actual deploy of Uber jar (more critical action)

rnveach:
@stoyanK7 
> After merging PR, https://github.com/checkstyle/checkstyle/wiki/Semi-automated-release has to be updated.

Are you providing this update?

.github/workflows/new-milestone-and-issue-in-other-repo.yml
```diff
@@ -2,8 +2,8 @@
 # GitHub Action to Create New Milestone and issue in Other Repositories.
 #
 #############################################################################
-name: "R: Close/Create Milestone, Upload '-all' jar, Create issues"
-run-name: "R: Close/Create Milestone, Upload '-all' jar, Create issues ${{ inputs.version }}"
+name: "R: Close/Create Milestone, Create issues"
+run-name: "R: Close/Create Milestone, Create issues ${{ inputs.version }}"
```
rnveach, line 6: Is name and run-name basically the same, but one says what name it goes by under a run? If not, what is the difference and why do we have these as 2 different names?

romani:
>Are you providing this update?

No, manually should be done by maintainers

.github/workflows/new-milestone-and-issue-in-other-repo.yml
romani: Name for action title.
Run-name is for execution instance

romani:
wiki is updated - https://github.com/checkstyle/checkstyle/wiki/Semi-automated-release

`R: upload uber('-all') jar to Github`

.github/workflows/new-milestone-and-issue-in-other-repo.yml
rnveach: > why do we have these as 2 different names?

romani: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#name

romani: Action is workflow. We have name for workflow, it is like a name of class. And we have name for execution (run-name), we make it same as name of workflow plus run parameters. Execution is like instance of workflow.

rnveach: But that doesn't say why they have to be different and can't be the same. For example, why aren't they both `"R: Close/Create Milestone, Create issues ${{ inputs.version }}"`. We didn't use to do this name and run-name in other workflows.

stoyanK7: `${{ inputs.xx }}` context is not supported in `name:`, only in `run-name`.

rnveach: @stoyanK7 Thanks. I did not know this.

Just to confirm, you are talking only about the top level name? I see sub-names have `${{}}`.

rnveach (approved):

.github/workflows/new-milestone-and-issue-in-other-repo.yml
stoyanK7: Yes, the sub-level names are names for `job`s, not the workflow itself.

rnveach: Thanks.

PR #12795, opened 2023-03-04
Issue #11374: Fix false positive related to anon inner classes in UnusedLocalVariableCheck

### LINKED ISSUE #11374: False Positive: UnusedLocalVariableCheck

Vyom-Yadav:
Resolves #11374

Diff Regression config: https://gist.githubusercontent.com/romani/e93272bbd35d686c134d5e6197322096/raw/4d9182b034271dcab66f9d497c5d04737715119d/config-pull-12795.xml

single check config: https://gist.githubusercontent.com/Vyom-Yadav/98dceb63a79f4833e85fff9b2e1464a6/raw/113875d5a22338b01bac435d7dae45446e6e1d2b/my_checks.xml

Final Report (clean): https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/3e532cd_2023140921/reports/diff/index.html

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
@@ -486,7 +486,7 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
         else {
             final List<TypeDeclDesc> typeDeclWithSameName = typeDeclWithSameName(shortNameOfClass);
             if (!typeDeclWithSameName.isEmpty()) {
-                obtainedClass = getTheNearestClass(
+                obtainedClass = getTheNearestClassOfAnonInnerClass(
                         anonInnerAstToTypeDeclDesc.get(literalNewAst).getQualifiedName(),
                         typeDeclWithSameName);
             }
@@ -555,10 +555,11 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
      * @param typeDeclWithSameName typeDeclarations which have the same name as the super class
      * @return the nearest class
      */
-    private static TypeDeclDesc getTheNearestClass(String outerTypeDeclName,
-            List<TypeDeclDesc> typeDeclWithSameName) {
+    private static TypeDeclDesc getTheNearestClassOfAnonInnerClass(
+        String outerTypeDeclName, List<TypeDeclDesc> typeDeclWithSameName) {
+
         return Collections.min(typeDeclWithSameName, (first, second) -> {
-            return getTypeDeclarationNameMatchingCountDiff(outerTypeDeclName, first, second);
+            return getAnonTypeDeclarationNameMatchingCountDiff(outerTypeDeclName, first, second);
         });
     }
 
@@ -571,13 +572,13 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
      * @param secondTypeDecl second input type declaration
      * @return difference between type declaration name matching count
      */
-    private static int getTypeDeclarationNameMatchingCountDiff(String outerTypeDeclName,
-                                                               TypeDeclDesc firstTypeDecl,
-                                                               TypeDeclDesc secondTypeDecl) {
+    private static int getAnonTypeDeclarationNameMatchingCountDiff(String outerTypeDeclName,
+                                                                   TypeDeclDesc firstTypeDecl,
+                                                                   TypeDeclDesc secondTypeDecl) {
         int diff = Integer.compare(
-            CheckUtil.typeDeclarationNameMatchingCount(
+            CheckUtil.getAnonSuperTypeMatchingCount(
                 outerTypeDeclName, secondTypeDecl.getQualifiedName()),
-            CheckUtil.typeDeclarationNameMatchingCount(
+            CheckUtil.getAnonSuperTypeMatchingCount(
                 outerTypeDeclName, firstTypeDecl.getQualifiedName()));
         if (diff == 0) {
             diff = Integer.compare(firstTypeDecl.getDepth(), secondTypeDecl.getDepth());
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
@@ -26,7 +26,6 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Optional;
-import java.util.function.Function;
 import java.util.function.ToIntFunction;
 
 import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
@@ -260,13 +259,42 @@ public class FinalClassCheck
         final String superClassName = getSuperClassName(currentClass.getTypeDeclarationAst());
         if (superClassName != null) {
             final ToIntFunction<ClassDesc> nestedClassCountProvider = classDesc -> {
-                return CheckUtil.typeDeclarationNameMatchingCount(qualifiedClassName,
+                return typeDeclarationNameMatchingCount(qualifiedClassName,
                                                                   classDesc.getQualifiedName());
             };
             getNearestClassWithSameName(superClassName, nestedClassCountProvider)
                 .or(() -> Optional.ofNullable(innerClasses.get(superClassName)))
                 .ifPresent(ClassDesc::registerNestedSubclass);
         }
+    }
+
+    /**
+     * Calculates and returns the type declaration name matching count.
+     *
+     * <p>
+     * Suppose our pattern class is {@code foo.a.b} and class to be matched is
+     * {@code foo.a.ball} then type declaration name matching count would be calculated by
+     * comparing every character, and updating main counter when we hit "." to prevent matching
+     * "a.b" with "a.ball". In this case type declaration name matching count
+     * would be equal to 6 and not 7 (b of ball is not counted).
+     * </p>
+     *
+     * @param patternClass class against which the given class has to be matched
+     * @param classToBeMatched class to be matched
+     * @return class name matching count
+     */
+    private static int typeDeclarationNameMatchingCount(String patternClass,
+                                                       String classToBeMatched) {
+        final int length = Math.min(classToBeMatched.length(), patternClass.length());
+        int result = 0;
+        for (int idx = 0;
+             idx < length && patternClass.charAt(idx) == classToBeMatched.charAt(idx);
+             ++idx) {
+            if (patternClass.charAt(idx) == PACKAGE_SEPARATOR.charAt(0)) {
+                result = idx;
+            }
+        }
+        return result;
     }
 
     /**
@@ -283,7 +311,8 @@ public class FinalClassCheck
         final String superClassName = CheckUtil.getShortNameOfAnonInnerClass(literalNewAst);
 
         final ToIntFunction<ClassDesc> anonClassCountProvider = classDesc -> {
-            return getAnonSuperTypeMatchingCount(outerTypeDeclName, classDesc.getQualifiedName());
+            return CheckUtil.getAnonSuperTypeMatchingCount(outerTypeDeclName,
+                                                           classDesc.getQualifiedName());
         };
         getNearestClassWithSameName(superClassName, anonClassCountProvider)
             .or(() -> Optional.ofNullable(innerClasses.get(superClassName)))
@@ -317,8 +346,8 @@ public class FinalClassCheck
      * }
      * }
      * </pre>
-     * If the {@link Function} {@code countProvider} hadn't used
-     * {@link FinalClassCheck#getAnonSuperTypeMatchingCount} to
+     * If the {@link ToIntFunction} {@code countProvider} hadn't used
+     * {@link CheckUtil#getAnonSuperTypeMatchingCount} to
      * calculate the matching count then the logic would have falsely evaluated
      * {@code Main.One.Two} to be the super class of the anonymous inner class.
      *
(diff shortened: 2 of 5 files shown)


Vyom-Yadav:
Github, generate report

romani:
Please recheck CI failures

Vyom-Yadav:
@nrmancuso @romani @rnveach Let's get this in the review stage. Only cosmetic changes have been made to `FinalClassCheck`.

romani:
@Vyom-Yadav , your report is only on UnusedLocalVariable but you change two classes.
please put in regression config both classes.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
```diff
@@ -655,9 +655,9 @@ private static int getTypeDeclarationNameMatchingCountDiff(String outerTypeDeclN
                                                                TypeDeclDesc firstTypeDecl,
                                                                TypeDeclDesc secondTypeDecl) {
         int diff = Integer.compare(
-            CheckUtil.typeDeclarationNameMatchingCount(
+            CheckUtil.getAnonSuperTypeMatchingCount(
```
romani, line 658: this change is weird, naming is confusing. 
host method is ` getTypeDeclarationNameMatchingCountDiff` and whole logic is in `getAnonSuperTypeMatchingCount`

why "AnnonSuper"  is so critical ?

it make sense for registerAnonymousInnerClassToSuperClass, but not in this context.
looks like getAnonSuperTypeMatchingCount name is root of confusion.

romani (changes requested):
items

Vyom-Yadav:
> your report is only on UnusedLocalVariable but you change two classes.
please put in regression config both classes.

I have made changes to both but changes to FinalClassCheck are non functional changes, there was a util method used by both `FinalClass` and `UnusedLocalVar` but now it is only used by `FinalClass` so it was moved out of util. Similarily there was a method in `FinalClass` that is now used by both `FinalClass` and `UnusedLocalVar` so it was moved to util.

No functional changes have been made to `FinalClassCheck`.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
Vyom-Yadav: The matching count way is different for anon inner classes and regular classes, that is why this is specifically mentioned. 
For regular classes we follow:
```java
/**                                                                                    
 * Calculates and returns the type declaration name matching count.                    
 *                                                                                     
 * <p>                                                                                 
 * Suppose our pattern class is {@code foo.a.b} and class to be matched is             
 * {@code foo.a.ball} then type declaration name matching count would be calculated by 
 * comparing every character, and updating main counter when we hit "." to prevent matc
 * "a.b" with "a.ball". In this case type declaration name matching count              
 * would be equal to 6 and not 7 (b of ball is not counted).                           
 * </p>                                                                                
 *                                                                                     
 * @param patternClass class against which the given class has to be matched           
 * @param classToBeMatched class to be matched                                         
 * @return class name matching count                                                   
 */                                                                                    
private static int typeDeclarationNameMatchingCount(String patternClass,               
                                                   String classToBeMatched) {          
    final int length = Math.min(classToBeMatched.length(), patternClass.length());     
    int result = 0;                                                                    
    for (int idx = 0;                                                                  
         idx < length && patternClass.charAt(idx) == classToBeMatched.charAt(idx);     
         ++idx) {                                                                      
        if (patternClass.charAt(idx) == PACKAGE_SEPARATOR.charAt(0)) {                 
            result = idx;                                                              
        }                                                                              
    }                                                                                  
    return result;                                                                     
}                                                                                      
```

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
```diff
@@ -591,27 +591,42 @@ private static int lastIndexOfNonWhitespace(String line) {
...
-     * @return class name matching count
+     * @param patternTypeDeclaration type declaration against which the given type declaration has
+     *                               to be matched
+     * @param typeDeclarationToBeMatched type declaration to be matched
+     * @return type declaration matching count
      */
-    public static int typeDeclarationNameMatchingCount(String patternClass,
-                                                       String classToBeMatched) {
-        final int length = Math.min(classToBeMatched.length(), patternClass.length());
+    public static int getAnonSuperTypeMatchingCount(String patternTypeDeclaration,
```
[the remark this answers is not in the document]
Vyom-Yadav: Yes, there is no boolean flag or something, instead, we use it particularly in an `ToIntFunction`, and the logic corresponds to anon inner classes.
```java
final ToIntFunction<ClassDesc> anonClassCountProvider = classDesc -> {            
    return CheckUtil.getAnonSuperTypeMatchingCount(outerTypeDeclName,             
                                                   classDesc.getQualifiedName()); 
};                                                                                
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
romani: it is ok to have different methods for each type of strucutre, but why 
`getTypeDeclarationNameMatchingCountDiff` is not
`getAnnonTypeDeclarationNameMatchingCountDiff` ?

as we use `CheckUtil.getAnonSuperTypeMatchingCount` in it.

Vyom-Yadav: Changed to `getAnonTypeDeclarationNameMatchingCountDiff`, done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
```diff
@@ -638,7 +638,7 @@ private boolean hasSameNameAsSuperClass(String superClassName, TypeDeclDesc type
     private static TypeDeclDesc getTheNearestClass(String outerTypeDeclName,
             List<TypeDeclDesc> typeDeclWithSameName) {
         return Collections.min(typeDeclWithSameName, (first, second) -> {
-            return getTypeDeclarationNameMatchingCountDiff(outerTypeDeclName, first, second);
+            return getAnonTypeDeclarationNameMatchingCountDiff(outerTypeDeclName, first, second);
```
romani, line 641:
```
private static TypeDeclDesc getTheNearestClass(....

   return getAnonTypeDeclaration
```

Why Nearest is always Anon class ? 
looks like we need to rename getTheNearestClass to getTheNearestAnonClass

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
Vyom-Yadav: The nearest class is not anon class, this method returns the superclass of anon-inner class, getNearestClass returns the nearest super class that matches the anon inner class name the most.

romani: Usually super class mean class that we extended. Do you mean nesting class? Class that nest another class

Vyom-Yadav: Super class here is:
```java
static class j {                                          
    static void method() {                                                              
        m.h obj = new m.h() { // super class of this anon inner class is j.m.h                                                                                     
        };                                                
    }                                                     
                                                          
    static class m {                                      
        static class h {                                                                    
        }                                                 
    }                                                     
}                                                         
```

romani: ok , we search class name for anonymous class.
so `getTheNearestClass ..... by means of ... getTypeDeclarationNameMatchingCountDiff` is logical
you propose
   `getTheNearestClass by means of getAnonTypeDeclarationNameMatchingCountDiff` that sounds strange and nonsense.

problems is name of `public static int getAnonSuperTypeMatchingCount(String patternTypeDeclaration`
you renamed to be `AnonSuper` but name should stay and handing of annon properly in addition is details of implmentation

romani:
@Vyom-Yadav, please finish or abandon PR.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
Vyom-Yadav: @romani Consider this example:
```java
static class j {                                          
    static void method() {
        m.h obj = new m.h() { // super class of this anon inner class is j.m.h
        };
    }                                                
                                                          
    static class m {                                      
        static class h {  // m.h                                         
        }                                                 
    }    

    static class k {                                      
        static class m {
          static class h { // k.m.h
          }                                         
        }                                                 
    }                                                     
}   
```
Here:
`superClassName = m.h`
`outerTypeDeclName = j`
`typeDeclWithSameName = [ j.m.h , j.k.m.h ]` // because of `m.h`

`getAnonSuperTypeMatchingCount(outerTypeDeclName, typeDeclWithSameName[i])` Now `typeDeclWithSameName[i]` is a **potential anon super type**, hence the name `getAnonSuperTypeMatchingCount` because we get the matching count for the super class of anon inner class.

romani: please do me a favor and rename 
`getTheNearestClass`
to 
`getTheNearestClassOfAnonInnerClass`
 to make it clear it is not general method but very specific for anonymous and it it is not only do min of collection , it actually does `getAnon.......` so it is fully Anon related method.

just name of method should be changed.

image

romani: I renaming it myself to stop this conversation today.

romani:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.java
```diff
@@ -269,6 +268,35 @@ private void registerExtendedClass(String qualifiedClassName,
...
+     * comparing every character, and updating main counter when we hit "." to prevent matching
+     * "a.b" with "a.ball". In this case type declaration name matching count
+     * would be equal to 6 and not 7 (b of ball is not counted).
+     * </p>
+     *
+     * @param patternClass class against which the given class has to be matched
+     * @param classToBeMatched class to be matched
+     * @return class name matching count
+     */
+    private static int typeDeclarationNameMatchingCount(String patternClass,
```
nrmancuso, line 286: Ok, both this method and `getAnonSuperTypeMatchingCount` are hard to understand, and bring up a lot of questions (why prefix ++ sometimes, etc.). Additionally, when we find the next bug, it is tempting to extend logic within the loops, so my concern is also for maintainability.

Is there a reason that we cannot use a more general approach to both of these methods? Something like:
1. Turn both strings into lists of type Character
2. Take intersection of lists
3. Remove everything after after final `.`
4. Apply any special logic
5. Return count of remaining characters

1-3 could certainly be extracted to some util method like 'String getIntersection(String x, String y)`. Then, each one of these "specialty" methods can apply custom logic and return a count.

nrmancuso (changes requested):
Item:

romani:
@Vyom-Yadav, do you have time to finish this PR?

Vyom-Yadav:
I don't think I'll have time to wrap this up soon, but I'll try. (You may close it to ease project management, I can pick this later then)

romani:
@kkoutsilis, can you help us to finish this defect?
Bug fix is stuck in almost done state. But this Check is one of most complicated.

kkoutsilis:
> @kkoutsilis, can you help us to finish this defect? Bug fix is stuck in almost done state. But this Check is one of most complicated.

Will have a look, thanks @romani

kkoutsilis:
Hey, @romani what needs to be done for this to be considered fixed? Is it the refactoring mentioned [here](https://github.com/checkstyle/checkstyle/pull/12795/files/4183859b0e430873d6345ae898cab91f588ebafa#r1421488948)? Apart from this comment I believe this bug fix works.

romani:
Yes, just extra refactoring to keep code on leash of easy to understand.

romani:
If refactoring is not possible or not make it better, please do comments and thoughts and we will move forward.

romani:
@kkoutsilis , is this PR non-recoverable ?
we can close it and start from scratch.

kkoutsilis:
> @kkoutsilis , is this PR non-recoverable ?
> 
> we can close it and start from scratch.

Hey @romani, sorry I should have mentioned that before but I'm been away for the past week and haven't had access to my PC. I couldn't get something fully working with what was proposed in that comment. I can have another look this week when I'm back or feel free to close this one and start from scratch. Sorry for the inconvenience.

romani:
no problem, just want to be sure that you did not burn out on this PR.

kkoutsilis:
> no problem, just want to be sure that you did not burn out on this PR.

Hey @romani, I can't seem to crack it, there are always a couple of test cases that fail if I take the approach mentioned [here](https://github.com/checkstyle/checkstyle/pull/12795#discussion_r1421488948). 

It's been some time since I worked on it but now that I am back and look at it again I believe the issue with this approach is that if for example, we have 
```
a = "a.b.c.f"
b = "a.b.c.foo"
c = "a.b.c.f.bar"
```
Intersection of a with b = "a.b.c.f" 
Removing everything after the last "."  we get "a.b.c.", all good this is fine

Intersection of a with c = "a.b.c.f" 
Removing everything after the last "."  we get"a.b.c.",  in this case, this is wrong, we should not remove anything as the intersection is correct.

I am not yet sure how to differentiate between those two cases, did not get anything fully working. I might be missing something obvious but hope this makes sense. My hunch is that we need the loop for this to work, but not 100% sure yet.

I can either work on it for a couple more days when I get some more free time and see if I get something fully working or leave it for someone more experienced. What's your thoughts on this?

romani:
Hard to understand method is better than defect in Check.

Please do your judgement on better implementation and let's see your code  in your PR. 

Strive for ideal code will not block merge to fix defect.

romani:
Activity is moved to https://github.com/checkstyle/checkstyle/pull/15969

Issue #12497, opened 2022-12-03
Create GitHub action to put current milestone on issue of closed PR


romani:
We put milestone manually after we merge PR.

This should be automatic.
https://www.jessesquires.com/blog/2022/08/04/automatically-assign-milestones-with-github-actions/
https://github.com/orgs/community/discussions/26724#discussioncomment-3253107

Or something special for us. As sometimes pull request needs to marked with milestone as it is referred in commit, but it is rare case we can skip it if this is complicated.

rnveach:
It can't be automatic if the issue already has a milestone set. We sometimes have issues that span multiple releases.

romani:
I think it is fine to have latest milestone. Long lasting usually misc issues, user unlikely interested in watching them.

ayush27prasad:
@romani I am working on this issue, could you please assign it to me?

romani:
@Kevin222004, can you help with this issue?

Kevin222004:
I am on it

stoyanK7:
@Kevin222004 Are you still working on it? I am interested to take it.

Kevin222004:
@stoyanK7 please take it I am busy with exam till 2march 
And then still I have to learn more about this. 

Please work on it

stoyanK7:
Thanks a lot.

On it then.

rnveach:
Not working at https://github.com/checkstyle/checkstyle/pull/12797#issuecomment-1454895114

romani:
@stoyanK7, please take a look

stoyanK7:
@rnveach 

> We should fail on any error.

From what I remember, `--fail-with-body` was not available in all images so I didn't even bother adding it: https://github.com/checkstyle/checkstyle/issues/12451
I will try adding it to see if it is supported now.

---
> https://github.com/checkstyle/checkstyle/actions/runs/4332960192/jobs/7565786160
> {
  "message": "Resource not accessible by integration",
  "documentation_url": "https://docs.github.com/rest/reference/issues#update-an-issue"
}

It seems that the `GITHUB_TOKEN` does not have sufficient permissions to update the issue milestone which is strange. 
We should have the right permissions for this:
 https://github.com/checkstyle/checkstyle/blob/bbd4975a606d9c7670f695a6160f3e6c8575a463/.github/workflows/set-milestone-on-referenced-issue.yml#L16-L18

However, at the `Set up job` step of the failing job, this is shown:
[...]

PR #12799, opened 2023-03-05
Pull #12799: Switch to PAT secret in set-milestone workflow

stoyanK7:
Switches to `PAT` secret in `.github/workflows/set-milestone-on-referenced-issue.yml`.

From https://github.com/checkstyle/checkstyle/issues/12497#issuecomment-1455018698:
`GITHUB_TOKEN` does not receive write permissions for some reason.

Switching to `PAT` as suggested in https://github.com/checkstyle/checkstyle/issues/12497#issuecomment-1455130026
> as we have special logic in action `${{ github.event.pull_request.merged == true }}` that limits execution a lot, and content is trusty in this case. We should use PAT.

### DIFF
--- a/.github/workflows/set-milestone-on-referenced-issue.yml
+++ b/.github/workflows/set-milestone-on-referenced-issue.yml
@@ -2,7 +2,7 @@
 # GitHub action to set latest milestone on issue of merged PR.
 #
 #############################################################################
-name: 'Milestone for PR'
+name: 'Milestone issue closed by PR'
 
 on:
   pull_request:
--- a/.github/workflows/set-milestone-on-referenced-issue.yml
+++ b/.github/workflows/set-milestone-on-referenced-issue.yml
@@ -13,10 +13,6 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
-permissions:
-  issues: write
-  pull-requests: write
-
 jobs:
   set-milestone:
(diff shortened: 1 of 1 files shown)


rnveach (approved):
Is there anyway to show this will work?

stoyanK7:
I can do it on my fork, probably will work, but can't guarantee that it will work on `checkstyle` repo :/

.github/workflows/set-milestone-on-referenced-issue.yml
```diff
@@ -13,10 +13,6 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
romani, line 13: please change name from `name: 'Milestone for PR'`
to name: 'Milestone issue closed by PR'

romani (changes requested):
minor

romani (approved):
ok to merge

rnveach:
@romani When this is merged, please confirm the first run works without issue.

stoyanK7:
@rnveach 
> Also, is there some way we can identify what requires us to use a PAT to try to avoid this in the future.

We use PAT only for workflows that require stuff to be done in other repos since `GITHUB_TOKEN` is only limited to current repo where workflow is executed from. 

I'm curious if PAT works out here. I will try to find out if it works.

rnveach:
Its odd then a PAT is required since this isn't in other repo, unless, it thinks it is because we are going through the github API and it doesn't know where it came from.

PR #12800, opened 2023-03-06
Issue #11214: Add violation messages in JavadocTypeCheck

naotokuwa:
Issue: #11214
I finished adding violation messages except for 2 parts.
The first part I have a trouble is this file.
`src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadoctype/InputJavadocTypeJavadoc_2.java
`
In this file, I need to put `\$Revision.*\$` as a version format and I could not figure out how to do the escaping for special regex characters. I tried below patterns, but it did not work. Could you guide me how to fix it?

Things I tried
```
\\$Revision.*\\$
\\$Revision\.*\\$
\\$Revision.\*\\$
\\$Revision\.\*\\$
```

The second part I have a trouble is **25th** line of this file. 
`src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadoctype/InputJavadocTypeTypeParamsTags_1.java`
`public class InputJavadocTypeTypeParamsTags_1<A,B1,C456 extends Comparable> // violation 'missing'`
Since the class header is already long, I cannot put enough words to explain the violation message. Can I wrap it to 2nd line? If so, could you tell me how to wrap it? When I tested, the comment is not caught by tester and it failed the test.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
@@ -103,7 +103,6 @@ public final class InlineConfigParser {
             "com.puppycrawl.tools.checkstyle.checks.whitespace.EmptyLineSeparatorCheck",
             "com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheck",
             "com.puppycrawl.tools.checkstyle.checks.coding.PackageDeclarationCheck",
-            "com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocTypeCheck",
             "com.puppycrawl.tools.checkstyle.checks.regexp.RegexpCheck",
             "com.puppycrawl.tools.checkstyle.checks.whitespace.EmptyForInitializerPadCheck",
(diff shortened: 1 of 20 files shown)


romani:
>  Can I wrap it to 2nd line?

you can use "violation below", "violation above", "violation X lines below", Example: 
https://github.com/checkstyle/checkstyle/blob/0f3620329667200f70bf7d38de2523802eada195/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/summaryjavadoc/InputSummaryJavadocIncorrect.java#L23

naotokuwa:
@romani 
Thank you for your help.
In terms of first problem, currently, I only put `tag @version must match pattern` which is a valid expression. However, when I put the format above (ex,"\\$Revision.*\\$") after "pattern", it will give me a test failure.
In terms of second problem, I got it. Thank you.

naotokuwa:
I updated below file to fit the long comment. Since one line is added for a violation comment, I incremented the line number of each expected result in the test file as well.

romani:
It is ok to skip this details, reader just need to see type of violation, as sometimes check has multiple violations messages so better to show what type of violation it does.

naotokuwa:
Thank you

PR #12801, opened 2023-03-06
Pull #12801: Fix 'set-milestone-on-referenced-issue'

stoyanK7:
Moves from `pull-request:close` event trigger to a `push` trigger.

The issue with milestone: https://github.com/stoyanK7/checkstyle/issues/36
Job: https://github.com/stoyanK7/checkstyle/actions/runs/4346403881/jobs/7592443158

I believe this is the reason for our failure with making this job work yesterday: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories
> With the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository. The `GITHUB_TOKEN` has read-only permissions in pull requests from forked repositories.

### DIFF
--- a/.ci/set-milestone-on-referenced-issue.sh
+++ b/.ci/set-milestone-on-referenced-issue.sh
@@ -5,28 +5,24 @@ set -e
 source ./.ci/util.sh
 
 if [[ -z $1 ]]; then
-  echo "[ERROR] Pull request number is not set."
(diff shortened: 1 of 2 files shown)


romani (approved):
Let's try

romani:
It works!!!

Issue #5722, opened 2018-04-12
update DTD public names


romani:
public DTD name format
`prefix//owner_of_the_DTD//description_of_the_DTD//ISO 639_language_identifier`
according to http://xmlwriter.net/xml_guide/doctype_declaration.shtml#PublicDTD

attention to 
> The "DTD_location" is used to find the public DTD if it cannot be located by the "DTD_name".

DTD name is first step to load dtd, second step is remote url. ConfigurationLoader do use  "DTD_name"
to map to internally stored DTD files. So we have to preserve mapping of DTD_name at ConfigurationLoader.
But names still have to changes in all our repo to have good name and all next configurations will follow new format.

TODO:
- `Puppy Crawl` ==> `Checkstyle`
- `DTD Check Configuration` ==>  `DTD Checkstyle Configuration`
- `DTD Import Control` ==>  `DTD ImportControl Configuration`
- `DTD Package Names` ==>  `DTD Package Names Configuration`
- `DTD Suppressions` ==>  `DTD SuppressionFilter Configuration`
- `DTD Suppressions Xpath Experimental` ==>  `DTD SuppressionXpathFilter Experimental Configuration`
...

It should not affect anybody, as is has only description role.

diff should be like:
```
--- a/config/checkstyle_checks.xml
+++ b/config/checkstyle_checks.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <!DOCTYPE module PUBLIC
-          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
           "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
```

All files to update:
`grep -Liran "//Puppy Crawl" --include="*.xml" --exclude-dir="target" *`

romani:
@rnveach , please let me know your opinion on TODO part. I have concern about "...Configuration", It might be not required as DTD(Document Type Definition) kind of mean configuration and xml is mostly for configuration but in the same time we have xml report, that is not configuration and one day we will get DTD/XSD for it too.

please share your opinion.

rnveach:
Struts 2 seems to name its DTD after the file name.
https://github.com/apache/struts/blob/6e96f11debc4fa52c65a12b28fea82b514b96abd/core/src/main/resources/struts-2.5.dtd#L29
[...]

PR #12820, opened 2023-03-11
Issue #11230: fix FallThroughCheck for labeled statements

### LINKED ISSUE #11230: FallThrough check false positive with labelled break
I have read check documentation: https://checkstyle.sourceforge.io/config_coding.html#FallThrough

```cmd
> javac Test.java
(success)

> type config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <module name="TreeWalker">
    <module name="FallThrough">
    </module>
  </module>
</module>

> type Test.java
public class Test {
    public int test(int a, int b) {
        top: switch (a) {
            case 1:
                switch (b) {
                    case 10:
                        break top;
                    default:
                        return 11;
                }
            case 2: // line 11
                return 2;
        }
        return -1;
    }
}

> java -jar checkstyle-0.2.1-all.jar -c config.xml Test.java
Starting audit...
[ERROR] Test.java:11:13: Fall through from previous branch of the switch statement. [FallThrough]
Audit done.
```

---

Expected result is no errors.  This code never falls through from case 1 to case 2.  The problem seems to only happen with a labelled break.  If I use `return 11` instead of `break top` then checkstyle does not find a problem.

strkkk:
Closes #11230

Diff Regression config: https://gist.githubusercontent.com/strkkk/26466254f89c32d8f1a6bd53b6f5b251/raw/4ddccb1907ad938ea84048819f86498321931590/config_single.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
@@ -19,6 +19,8 @@
 
 package com.puppycrawl.tools.checkstyle.checks.coding;
 
+import java.util.HashSet;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -266,7 +268,7 @@ public class FallThroughCheck extends AbstractCheck {
         if (!isLastGroup || checkLastCaseGroup) {
             final DetailAST slist = ast.findFirstToken(TokenTypes.SLIST);
 
-            if (slist != null && !isTerminated(slist, true, true)
+            if (slist != null && !isTerminated(slist, true, true, new HashSet<>())
                 && !hasFallThroughComment(ast, nextGroup)) {
                 if (isLastGroup) {
                     log(ast, MSG_FALL_THROUGH_LAST);
@@ -285,10 +287,11 @@ public class FallThroughCheck extends AbstractCheck {
      * @param ast root of given subtree
      * @param useBreak should we consider break as terminator
      * @param useContinue should we consider continue as terminator
+     * @param labels label names
      * @return true if the subtree is terminated.
      */
     private boolean isTerminated(final DetailAST ast, boolean useBreak,
-                                 boolean useContinue) {
+                                 boolean useContinue, Set<String> labels) {
         final boolean terminated;
 
         switch (ast.getType()) {
@@ -298,35 +301,50 @@ public class FallThroughCheck extends AbstractCheck {
                 terminated = true;
                 break;
             case TokenTypes.LITERAL_BREAK:
-                terminated = useBreak;
+                terminated = useBreak || hasOuterLabel(ast, labels);
                 break;
             case TokenTypes.LITERAL_CONTINUE:
-                terminated = useContinue;
+                terminated = useContinue || hasOuterLabel(ast, labels);
                 break;
             case TokenTypes.SLIST:
-                terminated = checkSlist(ast, useBreak, useContinue);
+                terminated = checkSlist(ast, useBreak, useContinue, labels);
                 break;
             case TokenTypes.LITERAL_IF:
-                terminated = checkIf(ast, useBreak, useContinue);
+                terminated = checkIf(ast, useBreak, useContinue, labels);
                 break;
             case TokenTypes.LITERAL_FOR:
             case TokenTypes.LITERAL_WHILE:
             case TokenTypes.LITERAL_DO:
-                terminated = checkLoop(ast);
+                terminated = checkLoop(ast, labels);
                 break;
             case TokenTypes.LITERAL_TRY:
-                terminated = checkTry(ast, useBreak, useContinue);
+                terminated = checkTry(ast, useBreak, useContinue, labels);
                 break;
             case TokenTypes.LITERAL_SWITCH:
-                terminated = checkSwitch(ast, useContinue);
+                terminated = checkSwitch(ast, useContinue, labels);
                 break;
             case TokenTypes.LITERAL_SYNCHRONIZED:
-                terminated = checkSynchronized(ast, useBreak, useContinue);
+                terminated = checkSynchronized(ast, useBreak, useContinue, labels);
+                break;
+            case TokenTypes.LABELED_STAT:
+                labels.add(ast.getFirstChild().getText());
+                terminated = isTerminated(ast.getLastChild(), useBreak, useContinue, labels);
                 break;
             default:
                 terminated = false;
         }
         return terminated;
+    }
+
+    /**
+     * Checks if given break or continue ast has outer label.
+     *
+     * @param ast break or continue node
+     * @param labels label names in scope of case group
+     * @return true if local label used
+     */
+    private static boolean hasOuterLabel(DetailAST ast, Set<String> labels) {
+        return ast.getChildCount() == 2 && !labels.contains(ast.getFirstChild().getText());
     }
 
     /**
@@ -336,10 +354,11 @@ public class FallThroughCheck extends AbstractCheck {
      * @param slistAst SLIST to check
      * @param useBreak should we consider break as terminator
      * @param useContinue should we consider continue as terminator
+     * @param labels label names
      * @return true if SLIST is terminated.
      */
     private boolean checkSlist(final DetailAST slistAst, boolean useBreak,
-                               boolean useContinue) {
+                               boolean useContinue, Set<String> labels) {
         DetailAST lastStmt = slistAst.getLastChild();
 
         if (lastStmt.getType() == TokenTypes.RCURLY) {
@@ -347,7 +366,7 @@ public class FallThroughCheck extends AbstractCheck {
         }
 
         return lastStmt != null
-            && isTerminated(lastStmt, useBreak, useContinue);
+            && isTerminated(lastStmt, useBreak, useContinue, labels);
     }
 
     /**
@@ -357,17 +376,18 @@ public class FallThroughCheck extends AbstractCheck {
      * @param ast IF to check
      * @param useBreak should we consider break as terminator
      * @param useContinue should we consider continue as terminator
+     * @param labels label names
      * @return true if IF is terminated.
      */
     private boolean checkIf(final DetailAST ast, boolean useBreak,
-                            boolean useContinue) {
+                            boolean useContinue, Set<String> labels) {
         final DetailAST thenStmt = ast.findFirstToken(TokenTypes.RPAREN)
                 .getNextSibling();
         final DetailAST elseStmt = thenStmt.getNextSibling();
 
         return elseStmt != null
-                && isTerminated(thenStmt, useBreak, useContinue)
-                && isTerminated(elseStmt.getFirstChild(), useBreak, useContinue);
+                && isTerminated(thenStmt, useBreak, useContinue, labels)
+                && isTerminated(elseStmt.getFirstChild(), useBreak, useContinue, labels);
     }
 
     /**
@@ -375,9 +395,10 @@ public class FallThroughCheck extends AbstractCheck {
      * if allowed break, continue.
      *
      * @param ast loop to check
+     * @param labels label names
      * @return true if loop is terminated.
      */
-    private boolean checkLoop(final DetailAST ast) {
+    private boolean checkLoop(final DetailAST ast, Set<String> labels) {
         final DetailAST loopBody;
         if (ast.getType() == TokenTypes.LITERAL_DO) {
             final DetailAST lparen = ast.findFirstToken(TokenTypes.DO_WHILE);
@@ -387,7 +408,7 @@ public class FallThroughCheck extends AbstractCheck {
             final DetailAST rparen = ast.findFirstToken(TokenTypes.RPAREN);
             loopBody = rparen.getNextSibling();
         }
-        return isTerminated(loopBody, false, false);
+        return isTerminated(loopBody, false, false, labels);
     }
 
     /**
@@ -397,13 +418,15 @@ public class FallThroughCheck extends AbstractCheck {
      * @param ast loop to check
      * @param useBreak should we consider break as terminator
      * @param useContinue should we consider continue as terminator
+     * @param labels label names
      * @return true if try/catch/finally block is terminated
      */
     private boolean checkTry(final DetailAST ast, boolean useBreak,
-                             boolean useContinue) {
+                             boolean useContinue, Set<String> labels) {
         final DetailAST finalStmt = ast.getLastChild();
         boolean isTerminated = finalStmt.getType() == TokenTypes.LITERAL_FINALLY
-                && isTerminated(finalStmt.findFirstToken(TokenTypes.SLIST), useBreak, useContinue);
+                && isTerminated(finalStmt.findFirstToken(TokenTypes.SLIST),
+                                useBreak, useContinue, labels);
 
         if (!isTerminated) {
             DetailAST firstChild = ast.getFirstChild();
@@ -413,7 +436,7 @@ public class FallThroughCheck extends AbstractCheck {
             }
 
             isTerminated = isTerminated(firstChild,
-                    useBreak, useContinue);
+                    useBreak, useContinue, labels);
 
             DetailAST catchStmt = ast.findFirstToken(TokenTypes.LITERAL_CATCH);
             while (catchStmt != null
@@ -421,7 +444,7 @@ public class FallThroughCheck extends AbstractCheck {
                     && catchStmt.getType() == TokenTypes.LITERAL_CATCH) {
                 final DetailAST catchBody =
                         catchStmt.findFirstToken(TokenTypes.SLIST);
-                isTerminated = isTerminated(catchBody, useBreak, useContinue);
+                isTerminated = isTerminated(catchBody, useBreak, useContinue, labels);
                 catchStmt = catchStmt.getNextSibling();
             }
         }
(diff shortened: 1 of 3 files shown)


strkkk:
github, generate report

romani:
We have suspicion that it failing to fetch files without token

nrmancuso:
> We have suspicion that it failing to fetch files without token . We migrated all actions for release to have token, at least GitHub auto generated token.

Failure to generate report is from old DTD in config.

strkkk:
Report is clean

romani:
>Failure to generate report is from old DTD in config.

Please share details. Others might run into this. I am interested in: what is old? How it becomes old ? How few people got this old ?

nrmancuso:
@romani 

>what is old?

Change in DTD is described here: https://github.com/checkstyle/checkstyle/issues/5722

>How it becomes old ? How few people got this old ?

I would assume anyone that is still using the old DTD has just been a long time contributor, and reused some config gist they had laying around instead of using the latest at https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/my_check.xml

I made some comment in a PR at some point mentioning all maintainers as FYI about this, I can't find it now. The exception message from SAX makes it pretty obvious what failure is.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java
```diff
@@ -0,0 +1,60 @@
...
+FallThrough
+checkLastCaseGroup = (default)false
+reliefPattern = (default)falls?[ -]?thr(u|ough)
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.fallthrough;
+
+public class InputFallThroughLabeledBreak {
```
nrmancuso, line 11: Please add a couple more test cases:
1. break to outer labelled block, like:
```
outer: { switch(a) {case 1: break outer; ... }}
```
2. Break to label with comment, like `break /* comment */ top`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -298,30 +301,35 @@ private boolean isTerminated(final DetailAST ast, boolean useBreak,
                 terminated = true;
                 break;
             case TokenTypes.LITERAL_BREAK:
-                terminated = useBreak;
+                terminated = useBreak
+                    || ast.getChildCount() == 2 && !labels.contains(ast.getFirstChild().getText());
```
nrmancuso, line 305: Why isn't it enough to just check if a given `break` is breaking to a label? Why do we need label names?

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
strkkk: it is needed to handle weirdest cases like
```
        case 1:
            inner: inner2: switch (b) {
            case 10:
                break inner;
            default:
                return 11;
            }
        case 2: // violation 'Fall through from previous branch of the switch statement.'
```

In such case we need to collect label names and check if break is used with label that is outside of case block

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java
strkkk: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
nrmancuso: Leaving this unresolved so that others can see it.

nrmancuso (approved):

romani:
@rnveach , after your review, please assign to me, I will use this bug fix as reason for release test.

strkkk:
@romani some checks are skipped, I believe reason is the same as in https://github.com/checkstyle/checkstyle/pull/12831#issuecomment-1472400097

I submitted the appeal for on circleci site, but I have no idea how long it will take to process it

rnveach:
I am using blocked so we don't merge this until all checks are run to not damage master. In the meantime, I will review.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -266,7 +268,7 @@ public void visitToken(DetailAST ast) {
         if (!isLastGroup || checkLastCaseGroup) {
             final DetailAST slist = ast.findFirstToken(TokenTypes.SLIST);
 
-            if (slist != null && !isTerminated(slist, true, true)
+            if (slist != null && !isTerminated(slist, true, true, new HashSet<>())
```
rnveach, line 271: I don't use labels in Java code, but I noticed this is the only time we are "erasing" this labels list. We never seem to be removing anything from it, so things just keep growing.

Can we have 2 very distinct sets of code in the same case group use the same label names? For example, this compiles for me:
`````
        int a = 0;
        switch (a) {
        case 0:
            mylabel: while (a == 0) {
                break mylabel;
            }

            int b = 0;
            b++;

            mylabel: while (b == 0) {
                break mylabel;
            }
            break;
        default:
        }
`````
I also don't get any compile errors if a label is defined and never used.

Comment (strkkk) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java:271]: I dont get the point :)

This set is created every time we visit TokenTypes.CASE_GROUP, it is collected by GC after method invocation is completed. It doesn't really matter if labels are duplicated or not for this check. Same labels can be on same level, but not in hierarchy, e.g. this wont compile:
```
label_a: switch (1) {
        case 1: {
            label_a: break;
 }
```
we collect all labels inside case group and one set is enough for it

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java [hunk 4]
```diff
@@ -298,30 +301,35 @@ private boolean isTerminated(final DetailAST ast, boolean useBreak,
                 terminated = true;
                 break;
             case TokenTypes.LITERAL_BREAK:
-                terminated = useBreak;
+                terminated = useBreak
+                    || ast.getChildCount() == 2 && !labels.contains(ast.getFirstChild().getText());
                 break;
             case TokenTypes.LITERAL_CONTINUE:
                 terminated = useContinue;
                 break;
```
rnveach, line 309: What about a continue label? We don't need similar label tracking?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
rnveach: Like I said, I don't really use labels so I am not sure if I can go further.

So it doesn't matter where the label is located in terms of the check? It just seems odd we are not taking these contexts into account and only capturing a partial look of the code.

romani: @strkkk , I still think we should not care  about labels at all.
read https://github.com/checkstyle/checkstyle/issues/11230#issuecomment-1055297662

I am complaetely ok to to make note in description that we do not care where label of break is pointing, if it is present we consider it as good code for FallThrough.

all madness with detection of infinite loop is out of this module consideration.

romani (changes requested):
from issue:
> The problem seems to only happen with a labelled break.

we should treat `break;` and `break somelabel;` as same code.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
romani: infinite loop is outside of this Check targets.

rnveach:
> we should treat break; and break somelabel; as same code.

We do already in release artifact according to the code changed in this PR.
See https://github.com/checkstyle/checkstyle/pull/12820#discussion_r1135514169

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java
```diff
@@ -0,0 +1,77 @@
...
+        outer: outer2: switch (a) {
+            case 1:
+                inner1: inner2: switch (b) {
+                case 10:
+                    break outer; // ok
+                default:
+                    return 11;
+                }
+            case 2:
+                inner: for (;;) {
```
rnveach, line 35: Make a similar case like this but have the for loop have a condition.

Also similar with a while loop like:
````
		case 1:
            while(cond) {
                break top;
            }
		case 2: // line 11
````
or point me to where I can find them.

**Review by rnveach — CHANGES_REQUESTED** [review]
Just some more tests as I try to understand.

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java (hunk 3 above)
Comment (strkkk) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java:271]: I do not really understand what the problem is, to be honest.

> So it doesn't matter where the label is located in terms of the check? It just seems odd we are not taking these contexts into account and only capturing a partial look of the code. We have had bugs where we were not emptying caches correctly, so I am just checking if there is possible issues with your code.

This HashSet of labels is created for every CASE_GROUP. The idea is very simple - we only need to check label names inside case group because any other labelled break will mean that label is outside and code will exit somewhere outside CASE_GROUP. 

So, if we see something like `break my_cool_label` we only need to check if this `my_cool_label` in CASE_GROUP. If it is, there is no termination. If it is not - it is terminated.

@romani 
infinite loop is here just to make example shorter, since the only thing that matters is break.

Example code is weird, but it is compilable and is supported in this PR. I dont see much sense in reverting changes and correcting documentation

File: src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java (hunk 5 above)
Comment (strkkk) [reply src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java]: I added examples for while loop.

Regarding for cycle - I dont see what difference condition makes, I made infinite only to make it shorter. I brought condition back

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java (hunk 4 above)
Comment (strkkk) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java:308]: yes, it needs. Fixed

File: src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java (hunk 5 above)
Comment (rnveach) [reply src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java]: @strkkk I don't have an issue with this PR, or the issue, or the shorthand loops which look like infinity.

>  I brought condition back

Please have both if you removed one. I just wanted to see how fall through behaves in the different situations with your changes.

The while loop I asked for is just me seeing something that @romani said in the main issue I wanted confirm.

====

My thought process:

https://github.com/checkstyle/checkstyle/issues/11230#issuecomment-1474870614
> nested switch shoudl not be something special

https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java#L390
Sees this code and wonders why it isn't `true, true` as it seems that would also fix the issue. What fails if I turned this into `true, true`.

https://github.com/checkstyle/checkstyle/blob/ea919d5434653046e133aa7ea2f5b410d5a2c683/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThrough3.java#L74-L77
This violation on 78 goes away which is what I asked to be created.

I just wanted to see how your code behaves when it is a `break <label>` instead of just a break.

Playing in eclipse made me believe the condition changes making it drop through or not. Thats why I asked for a similar case and then one adding a condition.

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java (hunk 4 above)
Comment (rnveach) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java:308]: Unless I am missing it, I am not seeing new test inputs for `continue <label>`. Please add some and also please regenerate regression. Will be interesting if anything new for `continue` comes up.

File: src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java [hunk 6]
```diff
@@ -0,0 +1,101 @@
...
+                }
+            case 10:  // violation 'Fall through from previous branch of the switch statement.'
+                inner1:
+                {
+                    switch (a) {
+                    case 1:
+                        break outer; // ok
+                    }
+                }
+            case 11:
```
rnveach, line 95: Isn't this technically a fall-through, and a false negative, because the switch statement doesn't have a default case?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
rnveach: I resolving this as I believe my concerns are not validated, but I still find it just odd that we don't care where the go to actually goes to and we are never doing any clean up.

romani:
>we should support compilable code, even if it weird

Yes, but I miss a point why we need to track labels? We should be ok with any break

nrmancuso:
@romani maybe this will make it clear to you:

```
➜  src javac Test.java

➜  src java Test      
Calling m1!
falling through to default
in m1 default
Calling m2!
Done!

➜  src cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <module name="TreeWalker">
    <module name="FallThrough">
    </module>
  </module>
</module>

➜  src java -jar checkstyle-10.7.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:10:9: Fall through from previous branch of the switch statement. [FallThrough]
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:23:9: Fall through from previous branch of the switch statement. [FallThrough]
Audit done.

```

Violation at 23:9 is false positive that this PR fixes.

romani:
I still not sure why we need to trace labels, if we detected label usage I am ok to skip whole branch of cas and all inner switch.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
strkkk: I dont see much difference between `break` and `comtinue`, so imo same file is ok. I will rename it.
I will create new report as well, but I really doubt there will be something - continue with break is even less frequent than labelled break.

strkkk:
github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
strkkk: done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java
strkkk: yes, it looks like a false negative, but I am not sure we can handle this case properly without dataflow analysis.

I will create a new issue

strkkk: Update - it looks like this will become an issue after merge of this PR, since there is no handling for such case in master (labeled break)

So it should be created after merge or just ignored since this is not a frequent case

rnveach:
@strkkk is this PR still blocked?

romani:
I most likely blocked by me, I need to digest Nick explanation, I have some worries on our direction.

romani:
I am ok with a change, point that it didn't get is that we do not track labels outside of switch, so any jump to outside of switch considered as break.
It is ok.

@strkkk , please rebase PR to prepare for final reviews.

romani:
@relentless-pursuit , please help us to finish this PR, looks like a little is left to be done, but still needs to be done.

relentless-pursuit:
> @relentless-pursuit , please help us to finish this PR, looks like a little is left to be done, but still needs to be done.

Is it

> @relentless-pursuit , please help us to finish this PR, looks like a little is left to be done, but still needs to be done.

Is someone working on it? Or, open to be taken?

romani:
Nobody is working on the this PR, it is confirmed as abandoned.

relentless-pursuit:
> Nobody is working on the this PR, it is confirmed as abandoned. I will close this PR as soon as you open new PR.
> 
> We appreciate you help.

Sure, will do that. Thanks.

nrmancuso:
I am closing this PR in favor of https://github.com/checkstyle/checkstyle/pull/14199

PR #12821, opened 2023-03-11
Pull #12821: Make grep always return exit code 0 in set-milestone

stoyanK7:
Resolves https://github.com/checkstyle/checkstyle/issues/12497#issuecomment-1462092794

> This still seems to be broken.
https://github.com/checkstyle/checkstyle/actions/runs/4369202639/jobs/7642723623#step:3:46
> Please review and confirm.

Success job run: https://github.com/stoyanK7/checkstyle/actions/runs/4393492340/jobs/7693921915

### DIFF
--- a/.ci/set-milestone-on-referenced-issue.sh
+++ b/.ci/set-milestone-on-referenced-issue.sh
@@ -17,9 +17,14 @@ echo "COMMITS_ARRAY=$COMMITS_ARRAY"
 
 echo "Extracting commit messages from commits array that match '^(Pull|Issue) #[0-9]+'."
 COMMIT_MESSAGES=$(echo "$COMMITS_ARRAY" | jq -r .[].message \
-  | grep -E "^(Pull|Issue) #[0-9]+")
+  | grep -E "^(Pull|Issue) #[0-9]+" || true)
 echo "COMMIT_MESSAGES=
 $COMMIT_MESSAGES"
+
+if [ -z "$COMMIT_MESSAGES" ]; then
+  echo "[WARN] No issue numbers found in commit messages."
+  exit 0
+fi
 
 echo "Extracting issue numbers from commit messages."
 ISSUE_NUMBERS=$(echo "$COMMIT_MESSAGES" | grep -oE "#[0-9]+" | cut -c2-)
@@ -27,8 +32,8 @@ echo "ISSUE_NUMBERS=
 $ISSUE_NUMBERS"
 
 if [ -z "$ISSUE_NUMBERS" ]; then
-  echo "[ERROR] No issue numbers found in commit messages."
-  exit 0
+  echo "[ERROR] No issue numbers found in commit messages but was expecting some."
+  exit 1
 fi
 
 echo "Fetching latest milestone."


.ci/set-milestone-on-referenced-issue.sh
```diff
@@ -17,7 +17,7 @@ echo "COMMITS_ARRAY=$COMMITS_ARRAY"
 
 echo "Extracting commit messages from commits array that match '^(Pull|Issue) #[0-9]+'."
 COMMIT_MESSAGES=$(echo "$COMMITS_ARRAY" | jq -r .[].message \
-  | grep -E "^(Pull|Issue) #[0-9]+")
+  | grep -E "^(Pull|Issue) #[0-9]+" || true)
```
stoyanK7, line 20: The solution was taken from here:
https://stackoverflow.com/a/61321426/9553927

grep returns an exit code 1 if it doesn't find lines that match the regex. In this case, we force it to return 0.

rnveach:
https://github.com/stoyanK7/checkstyle/actions/runs/4393492340/jobs/7693921915#step:3:52
> Error:  No issue numbers found in commit messages.

I find it odd this ends with an error, but "passes". I think this is just an issue in the wording and this should really be a warning, but....

https://github.com/checkstyle/checkstyle/blob/4361a49d385235b9cf44aa6e8f22cb2c2c23c0c9/.ci/set-milestone-on-referenced-issue.sh#L23

Shouldn't we have warning here there is no issues/pulls if `COMMIT_MESSAGES` is completely empty and then,

https://github.com/checkstyle/checkstyle/blob/4361a49d385235b9cf44aa6e8f22cb2c2c23c0c9/.ci/set-milestone-on-referenced-issue.sh#L29

Make this a real error with exit 1 saying we found issues/pulls but we found no numbers. Something must be up...

stoyanK7:
@romani proposed
> Yes, if not issue or pull, we should skip such commit

However, I could not come up with a solution that skips the workflow if a commit does not start with `Issue #` or `Pull #`. The best we can do is just exit with `0` like this PR does. With what GitHub supports, it is not possible to skip the workflow in our scenario. Here is where I got to and then gave up:

To skip the worflow, we would need an `if` check like this
```
jobs:
  set-milestone:
    runs-on: ubuntu-latest
    if: <any commit in push starts with 'Issue #' or 'Pull #'>
    ...
```

In a `push` triggered workflow, commits can be accessed with `github.event.commits` and the commits array looks like this:
```
[
  {
    "author": {
      "email": "stoyank127@gmail.com",
      "name": "stoyanK7",
      "username": "stoyanK7"
    },
    "committer": {
      "email": "stoyank127@gmail.com",
      "name": "stoyanK7",
      "username": "stoyanK7"
    },
    "distinct": true,
    "id": "e724125b7aa6727d55c5fe4448f34c71ab96f146",
    "message": "Pull #12821: Run set-milestone only on pull and issue commits",
    "timestamp": "2023-03-11T18:52:35+01:00",
    "tree_id": "d90485292819d32649aaf6d1b5c6514bb47dfee3",
    "url": "https://github.com/stoyanK7/checkstyle/commit/e724125b7aa6727d55c5fe4448f34c71ab96f146"
  },
  {
    "author": {
      "email": "stoyank127@gmail.com",
      "name": "stoyanK7",
      "username": "stoyanK7"
    },
    "committer": {
      "email": "stoyank127@gmail.com",
      "name": "stoyanK7",
      "username": "stoyanK7"
    },
    "distinct": true,
    "id": "fa043c5536ea7b83c9243a7ece39ca3747e46f4c",
    "message": "infra: disable all CI services",
    "timestamp": "2023-03-11T18:53:10+01:00",
    "tree_id": "575347cbe6dc2dd79bb057a73023c88ff48598a5",
    "url": "https://github.com/stoyanK7/checkstyle/commit/fa043c5536ea7b83c9243a7ece39ca3747e46f4c"
  },
  {
    "author": {
      "email": "49699333+dependabot[bot]@users.noreply.github.com",
      "name": "dependabot[bot]",
      "username": "dependabot[bot]"
    },
    "committer": {
      "email": "stoyank127@gmail.com",
      "name": "stoyanK7",
      "username": "stoyanK7"
    },
    "distinct": true,
    "id": "fde28c8805b1cb58dbd30da6bf68fdaf66a320ad",
    "message": "dependency: bump org.eclipse.jgit\n\nBumps org.eclipse.jgit from 6.4.0.202211300538-r to 6.5.0.202303070854-r.\n\n---\nupdated-dependencies:\n- dependency-name: org.eclipse.jgit:org.eclipse.jgit\n  dependency-type: direct:development\n  update-type: version-update:semver-minor\n...\n\nSigned-off-by: dependabot[bot] <support@github.com>",
    "timestamp": "2023-03-11T18:54:53+01:00",
    "tree_id": "575347cbe6dc2dd79bb057a73023c88ff48598a5",
    "url": "https://github.com/stoyanK7/checkstyle/commit/fde28c8805b1cb58dbd30da6bf68fdaf66a320ad"
  }
]
```
With GitHub's [Object filters](https://docs.github.com/en/actions/learn-github-actions/expressions#object-filters) we could write `github.event.commits.*.message` and would extract only the commit messages in a JSON array:
```
[
  "Pull #12821: Run set-milestone only on pull and issue commits",
  "infra: disable all CI services",
  "dependency: bump or..."
]
```

GitHub offers built-in functions that can be used in expressions but none help accomplish the task:
 - [contains( search, item )](https://docs.github.com/en/actions/learn-github-actions/expressions#contains) can make **exact**(not partial) string comparison between the `item` string and all strings in the `search` array.
 - [startsWith( searchString, searchValue )](https://docs.github.com/en/actions/learn-github-actions/expressions#startswith) only accepts a single string but we have an array of commits to go over.

Moreover, [regular expressions are not supported by GitHub actions](https://github.com/orgs/community/discussions/26186#discussioncomment-3250722) so nothing can be done there too.

rnveach:
> To skip the worflow, we would need an if check like 

I don't understand why we need to modify the YML file. You are already doing a skip with that `exit 0`. I am just asking to re-arrange it in the script file.

stoyanK7:
@rnveach I just put out there the initial path I took. It seemed most straightforward to skip the workflow completely, but it's not possible to do so. Just my reasons for not doing so.

rnveach:
I am not against your change here, but I don't understand why you are saying what I am asking is not possible.  This is patch I am asking. Maybe you can explain why this patch isn't possible or won't work.

````
diff --git a/.ci/set-milestone-on-referenced-issue.sh b/.ci/set-milestone-on-referenced-issue.sh
old mode 100755
new mode 100644
index cc1bf81..9b8cce7
--- a/.ci/set-milestone-on-referenced-issue.sh
+++ b/.ci/set-milestone-on-referenced-issue.sh
@@ -21,13 +21,18 @@
 echo "COMMIT_MESSAGES=
 $COMMIT_MESSAGES"
 
+if [ -z "$COMMIT_MESSAGES" ]; then
+  echo "[WARN] No issue numbers found in commit messages."
+  exit 0
+fi
+
 echo "Extracting issue numbers from commit messages."
 ISSUE_NUMBERS=$(echo "$COMMIT_MESSAGES" | grep -oE "#[0-9]+" | cut -c2-)
 echo "ISSUE_NUMBERS=
 $ISSUE_NUMBERS"
 
 if [ -z "$ISSUE_NUMBERS" ]; then
-  echo "[ERROR] No issue numbers found in commit messages."
+  echo "[ERROR] No issue numbers found in commit messages but was expecting some."
   exit 0
 fi
 ````

.ci/set-milestone-on-referenced-issue.sh
romani: I more like ` | cat` to avoid grep failures.

romani:
Extra validation and exit it good.

.ci/set-milestone-on-referenced-issue.sh
nrmancuso: @romani  ` | cat`  is not intuitive, we try to be Java friendly above all, `|| true` is more clear. Also - shellcheck does not like useless `cat`s.

stoyanK7:
Done.

Actions: https://github.com/stoyanK7/checkstyle/actions/runs/4397126469/jobs/7700047258#step:3:49

stoyanK7:
> Just my reasons for not doing so.

I was talking about skipping the workflow, not your proposal. I agree with your proposal. This is a misunderstanding :D

.ci/set-milestone-on-referenced-issue.sh
romani: I thought we disabled such rule

nrmancuso: @romani That was for `wc -l`

romani: I am not an expert here ...... but ` | cat` works well in sequential greps.
may be 
`grep x || true | grep y || true | grep z || true`
vs
`grep x | cat | grep y | cat | grep z | cat`

At least in second you do not need to remember some nuances, cat just re-prints, so if we deal with string chain-pipe updates, it might be better.

lets keep "|| true".

romani (approved):
ok to merge

rnveach (approved):

PR #12543, opened 2022-12-21
Issue #11968: false Postive In VariableDeclarationUsageDistance

arinmodi:
Closes : #11968 : VariableDeclarationUsageDistance not able to handle try with resources properly

Diff Regression config: https://gist.githubusercontent.com/arinmodi/911bd345758c522c77efe6b82afc8c4e/raw/13d8c333a277c0cdb45ce865d43ea3dfde9f27f6/VariableDeclarationUsageDistance.xml

Issue #11968, opened 2022-07-23
VariableDeclarationUsageDistance not able to handle try with resources properly


Vyom-Yadav:
I have read check documentation: https://checkstyle.org/config_coding.html#VariableDeclarationUsageDistance
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Test.java
```
```
/var/tmp $ cat config.xml
```
```XML
<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
        "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">

<module name="Checker">
  <module name="TreeWalker">
    <module name="VariableDeclarationUsageDistance">
      <property name="validateBetweenScopes" value="true"/>
      <property name="allowedDistance" value="1"/>
    </module>
  </module>
</module>
```
```
/var/tmp $ cat Test.java
```
```java
public class Test {
    void method() {
        String a = ""; // no violation expected
        try (AutoCloseable i = new java.io.StringReader(a)) {
        }
        catch (Exception e) {
            throw new RuntimeException(e);
        }
    }
}
```
```
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.3.1-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:3:9: Distance between variable 'a' declaration and its first usage is 2, but allowed 1.  Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value). [VariableDeclarationUsageDistance]
Audit done.
Checkstyle ends with 1 errors.
```
No error was expected, the variable is used just after it was declared, plus in the source code, there is no mention of `TokenTypes.RESOURCES` so this seems like a false positive.

jxr98:
I am on it.

nrmancuso:
@jxr98 this is a great improvement on this check, we found a few more false positives than we initially thought :)

PR #12822, opened 2023-03-11
Issue #11968: corrected false positive in VariableDeclarationUsageDistanceCheck

jxr98:
Issue: #11968 

- I used some code from Pull Request: #12543.
- I removed unnecessary suppressed mutation 
```
<mutation unstable="false">
    <sourceFile>VariableDeclarationUsageDistanceCheck.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck</mutatedClass>
    <mutatedMethod>getFirstNodeInsideTryCatchFinallyBlocks</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
    <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getLastChild with receiver</description>
    <lineContent>final DetailAST finalBlock = currentNode.getLastChild();</lineContent>
  </mutation>
```
- This PR solved Issue: #11968 as follows:
```
less config.xml
```
```
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name = "Checker">

    <!-- do not change severity to 'error', as that will hide errors caused by exceptions -->

    <!-- haltOnException is required for exception fixes and reporting of all exceptions -->

    <!-- BeforeExecutionFileFilters is required for sources of java9 -->
        <property name="fileNamePattern" value="module\-info\.java$" />
    </module>

    <module name="TreeWalker">
         <!-- Example of checkstyle Check usage -->
         <!-- PLEASE CHANGE IT TO CHECK YOU ARE TESTING !!!! -->
          <property name="id" value="Default"/>
        </module>
              
          <property name="id" value="Custom"/>
        </module> 
    </module>
    <!-- Example of filter -->
    <!--
    <module name="SeverityMatchFilter">
        <property name="acceptOnMatch" value="false"/>
    </module>
    -->

    <!-- as we run on regression even on non-compiled files we need to skip exceptions on them -->
    <module name="SuppressionSingleFilter">
      <property name="message" value="Exception occurred while parsing"/>
      <property name="checks" value="Checker"/>
    </module>
</module>
```
```
less test.java
```
```
public class Test {
    void method() {
        String a = ""; // no violation expected
        try (AutoCloseable i = new java.io.StringReader(a)) {
        }
        catch (Exception e) {
            throw new RuntimeException(e);
        }
    }
}
```
```
$ java -jar checkstyle-10.9.3-SNAPSHOT-all.jar -c config.xml Test.java
Starting audit...
Audit done.
```

### DIFF
--- a/config/pitest-suppressions/pitest-coding-1-suppressions.xml
+++ b/config/pitest-suppressions/pitest-coding-1-suppressions.xml
@@ -156,15 +156,6 @@
   <mutation unstable="false">
     <sourceFile>VariableDeclarationUsageDistanceCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck</mutatedClass>
-    <mutatedMethod>getFirstNodeInsideTryCatchFinallyBlocks</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getLastChild with receiver</description>
-    <lineContent>final DetailAST finalBlock = currentNode.getLastChild();</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>VariableDeclarationUsageDistanceCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck</mutatedClass>
     <mutatedMethod>isInitializationSequence</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
     <description>removed call to com/puppycrawl/tools/checkstyle/api/DetailAST::getText</description>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
@@ -692,10 +692,9 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
                 if (isChild(currentStatementAst, variableAst)) {
                     variableUsageExpressions.add(currentStatementAst);
                 }
-                // If expression doesn't contain variable and this variable
-                // hasn't been met yet, then distance + 1.
+                // If expression hasn't been met yet, then distance + 1.
                 else if (variableUsageExpressions.isEmpty()
-                        && currentStatementAst.getType() != TokenTypes.VARIABLE_DEF) {
+                        && !isZeroDistanceToken(currentStatementAst.getType())) {
                     distance++;
                 }
             }
@@ -1009,6 +1008,40 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
     private boolean isVariableMatchesIgnorePattern(String variable) {
         final Matcher matcher = ignoreVariablePattern.matcher(variable);
         return matcher.matches();
+    }
+
+    /**
+     * Check if the token should be ignored for distance counting.
+     * For example,
+     * <pre>
+     *     try (final AutoCloseable t = new java.io.StringReader(a);) {
+     *     }
+     * </pre>
+     * final is a zero-distance token and should be ignored for distance counting.
+     * <pre>
+     *     class Table implements Comparator&lt;Integer&gt;{
+     *     }
+     * </pre>
+     * An inner class may be defined. Both tokens implements and extends
+     * are zero-distance tokens.
+     * <pre>
+     *     public int method(Object b){
+     *     }
+     * </pre>
+     * public is a modifier and zero-distance token. int is a type and
+     * zero-distance token.
+     *
+     * @param type
+     *        Token type of the ast node.
+     * @return true if it should be ignored for distance counting, otherwise false.
+     */
+    private static boolean isZeroDistanceToken(int type) {
+        return type == TokenTypes.VARIABLE_DEF
+                || type == TokenTypes.TYPE
+                || type == TokenTypes.MODIFIERS
+                || type == TokenTypes.RESOURCE
+                || type == TokenTypes.EXTENDS_CLAUSE
+                || type == TokenTypes.IMPLEMENTS_CLAUSE;
     }
 
 }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
@@ -85,7 +85,6 @@ public class VariableDeclarationUsageDistanceCheckTest extends
         final String[] expected = {
             "17:9: " + getCheckMessage(MSG_KEY, "first", 5, 1),
             "29:9: " + getCheckMessage(MSG_KEY, "allInvariants", 2, 1),
-            "59:9: " + getCheckMessage(MSG_KEY, "a", 2, 1),
         };
         verifyWithInlineConfigParser(
                 getPath("InputVariableDeclarationUsageDistanceGeneral2.java"), expected);
@@ -321,17 +320,34 @@ public class VariableDeclarationUsageDistanceCheckTest extends
     }
 
     @Test
-    public void testGeneral3Class() throws Exception {
+    public void testGeneralClass3() throws Exception {
         final String[] expected = {
-            "28:9: " + getCheckMessage(MSG_KEY, "a", 4, 1),
             "46:9: " + getCheckMessage(MSG_KEY, "a", 2, 1),
-            "65:9: " + getCheckMessage(MSG_KEY, "a", 4, 1),
-            "106:9: " + getCheckMessage(MSG_KEY, "m", 4, 1),
         };
 
         verifyWithInlineConfigParser(
-                getPath("InputVariableDeclarationUsageDistanceGeneral3.java"),
-                expected);
+                getPath("InputVariableDeclarationUsageDistanceGeneral3.java"), expected);
+    }
+
+    @Test
+    public void testGeneralClass4() throws Exception {
+        final String[] expected = {
+            "26:9: " + getCheckMessage(MSG_KEY, "z", 3, 1),
+        };
+
+        verifyWithInlineConfigParser(
+                getPath("InputVariableDeclarationUsageDistanceGeneral4.java"), expected);
+    }
+
+    @Test
+    public void testVariableDeclarationUsageDistanceTryResources() throws Exception {
+        final String[] expected = {
+            "19:9: " + getCheckMessage(MSG_KEY, "a", 2, 1),
+            "20:9: " + getCheckMessage(MSG_KEY, "b", 2, 1),
+        };
+
+        verifyWithInlineConfigParser(
+                getPath("InputVariableDeclarationUsageDistanceTryResources.java"), expected);
     }
 
 }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral2.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral2.java
@@ -55,8 +55,7 @@ public class InputVariableDeclarationUsageDistanceGeneral2 {
     }
 
     void method() throws Exception {
-        // Until https://github.com/checkstyle/checkstyle/issues/11968
-        String a = ""; // violation 'Distance between .* declaration and its first usage is 2.'
+        String a = "";
         try (AutoCloseable i = new java.io.StringReader(a)) {
         }
         finally {
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral3.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral3.java
@@ -21,11 +21,11 @@ public class InputVariableDeclarationUsageDistanceGeneral3 {
     class Parent {
         void mm() {
         }
+        <T> void xx(List<T> m){}
     }
 
     public void method2() {
-        // until https://github.com/checkstyle/checkstyle/issues/13011
-        Integer a = 5;  // violation 'Distance .* is 4.'
+        Integer a = 5;
         class AClass extends Parent {
             int i;
 
@@ -61,8 +61,7 @@ public class InputVariableDeclarationUsageDistanceGeneral3 {
     }
 
     public void method5() {
-        // until https://github.com/checkstyle/checkstyle/issues/13011
-        Object a = new Object(); // violation 'Distance .* is 4.'
+        Object a = new Object();
         class AClass implements Ani {
             @Override
             public void method4(Object o) {
@@ -102,8 +101,7 @@ public class InputVariableDeclarationUsageDistanceGeneral3 {
     }
 
     public void method8() {
-        // until https://github.com/checkstyle/checkstyle/issues/13011
-        Integer m = 10; // violation 'Distance .* is 4.'
+        Integer m = 10;
         class BClass extends Parent{
             public BClass getNewInstance(){
                 return new BClass(){
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral4.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral4.java
@@ -1,0 +1,46 @@
+/*
+VariableDeclarationUsageDistance
+allowedDistance = 1
+ignoreVariablePattern = (default)
+validateBetweenScopes = true
+ignoreFinal = false
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.variabledeclarationusagedistance;
+
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+
+public class InputVariableDeclarationUsageDistanceGeneral4 {
+
+    class Parent {
+        void mm() {
+        }
+        <T> void xx(List<T> m){}
+    }
+    public void method9() {
+        // until https://github.com/checkstyle/checkstyle/issues/13011
+        Integer z = 5; // violation 'Distance .* is 3.'
+        Iterator<Integer> mn = new HashSet<Integer>().iterator();
+        class BClass extends Parent {
+            Boolean h = false;
+            @Override
+            public <t> void xx(List<t> a) {
+                if (a.get(0).toString().compareTo("1") > 0) {
+                    System.out.println("temp");
+                }
+            }
+            @Override
+            public void mm() {
+            }
+            void nn(Boolean m) {
+                if(z<0){
+                    System.out.println();
+                }
+            }
+        }
+    }
+}
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceTryResources.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceTryResources.java
@@ -1,0 +1,118 @@
+/*
+VariableDeclarationUsageDistance
+allowedDistance = 1
+ignoreVariablePattern = (default)
+validateBetweenScopes = true
+ignoreFinal = false
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.variabledeclarationusagedistance;
+
+import java.io.*;
+import java.nio.file.Files;
+import java.util.*;
+
+public class InputVariableDeclarationUsageDistanceTryResources {
+    public int methodTry() {
+        String a = ""; // violation 'Distance .* is 2.'
+        String b = "abc"; // violation 'Distance .* is 2.'
+        System.out.println();
+        try (AutoCloseable j = new java.io.StringReader(b);
+             final AutoCloseable i = new java.io.StringReader(a);
+             final AutoCloseable k = new java.io.StringReader(b);) {
+            b.replace(a.charAt(0), 'b');
+            String c = b.toString();
+        } catch (Exception e) {
+            System.out.println(a);
+            throw new RuntimeException(e);
+        }
+        return 0;
+    }
+    String method2() throws Exception {
+        String b = "";
+        FileReader fr = new FileReader(b);
+        BufferedReader br = new BufferedReader(fr);
+        try {
+            return br.readLine();
+        } finally {
+            br.close();
+            fr.close();
+        }
+    }
+    void method() {
+        byte[] buf = new byte[10];
+        File b = new File("");
+        try (AutoCloseable j = new java.io.StringReader(b.toString());
+             InputStream finstr = Files.newInputStream(b.toPath());
+             BufferedInputStream in = new BufferedInputStream(finstr, buf.length)) {
+            int length;
+            while ((length = in.read(buf)) != -1) {
+                System.out.println(buf.toString());
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+    String method3() throws Exception {
+        String b = "";
+        FileReader fr = new FileReader("");
+        BufferedReader br = new BufferedReader(fr);
+        try {
+            return br.readLine();
+        } catch (Exception e) {
+            FileReader t = new FileReader(b);
+        } finally {
+            br.close();
+            fr.close();
+            return b;
+        }
+    }
+    class myClass {
+        String m, n;
+        myClass(String a, String b) {
+            this.m = a;
+            this.n = b;
+        }
+        AutoCloseable getAutoCloseable(int m) {
+            return new AutoCloseable() {
+                @Override
+                public void close() throws Exception {
+
+                }
+            };
+        }
+    }
+    void method4() throws Exception {
+        String b = "abc";
+        myClass a = new myClass("", b);
+        Scanner scanner = new Scanner(System.in);
+        scanner.nextBigInteger(100);
+        try (final AutoCloseable t = a.getAutoCloseable(3);) {
+            Integer number = 10;
+            for (int i = 0; i < 10; i++) {
+                System.out.println(i);
+            }
+            b.substring(0, a.toString().length());
+        }
+    }
+    private void method5(OutputStream out, HashMap<String, myClass> fieldStatistics)
+            throws IOException {
+        out.write(fieldStatistics.size());
+        final boolean[] states = {fieldStatistics.isEmpty(), fieldStatistics.isEmpty()};
+        Object[] keys = fieldStatistics.keySet().toArray(new String[0]);
+        Object[] values = fieldStatistics.values().toArray(new myClass[0]);
+        for (int i = 0; i < states.length; i++) {
+            if (states[i]) {
+                out.write(((String) keys[i]).getBytes());
+                myClass statistics = (myClass) values[i];
+                assert statistics.hashCode() >= 0;
+                out.write(statistics.hashCode());
+                out.write(statistics.m.getBytes());
+            }
+        }
+    }
+}


jxr98:
Github, generate report

jxr98:
I have encountered an impediment. As in the pull request: #12543, @romani mentioned that 'our test input code base is missing for this Check. We should not rely much on regressions, all/most possible usages of structures/.... should be in our tests'. I don't know what it means. Also, the changed logic in https://github.com/jxr98/checkstyle/blob/c7ed4aecff9947da4bb8a4ea620f756fe689b61c/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L698-L699 solves the issue: #11968. But I am not sure if the logic is correct. How can I verify that?

nrmancuso:
@jxr98 if one issue indirectly solves another, it is always good to do:
1. Send PR for one issue
2. Get PR merged
3. Leave comment in other issue, showing proof by CLI with snapshot/ released version of Checkstyle with your changes that other issue is fixed

As for
>our test input code base is missing for this Check. We should not rely much on regressions, all/most possible usages of structures/.... should be in our tests

This just means that we should not always count on our regression testing to catch problems, it is better to have lots of different input in our "unit" tests and  catch problems there. Regression testing can take hours, where unit tests take ~5 minutes :) If we find something interesting in regression testing, it is always a good idea to add it to unit tests.

jxr98:
GitHub, generate report

jxr98:
@nrmancuso Thanks. This PR is to solve the Issue: #11968. I am waiting for the regression reports to figure out what to do next.

romani:
Ci report is broken, please use old way manual on your local https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#checkstyle-tester

jxr98:
> Ci report is broken, please use old way manual on your local https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#checkstyle-tester

Hi @romani. Thanks for reminding. But a regression report was generated. I am trying to understand the logic in the function `calculateDistanceInSingleScope()` and DetailAST.
https://github.com/checkstyle/checkstyle/blob/d4f9fa33c16cd5543fa587afc905965cc87867dc/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L611-L675
I have read the [documentation](https://checkstyle.org/config_coding.html#VariableDeclarationUsageDistance). But still cannot understand how to calculate distance between usage and declaration in differenct scopes. Is there any other documentation and tutorial that I can refer to? 

The regression test is not as quick as the unit test. The `VariableDeclarationUsageDistanceCheck` is one of the suppressed checks. We need the unit test to verify the correctness of the logic in `VariableDeclarationUsageDistanceCheck`. So first I need to specify violation messages in that check files to 'unsupress' the check first as in issue: #11214

romani:
ok, it means we need a bit more knowledge on this module.
Please help to fix violation trailaing messages for this Check at https://github.com/checkstyle/checkstyle/issues/11214 in separate PR, to see this Check in action.

jxr98:
> where is suppression ? for #11214 ?
Here the last one
https://github.com/checkstyle/checkstyle/blob/fe36c48a8b75bb97f5b4c617e0a79dfb724ead0a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L99-L114

romani:
@jxr98 , ok so reason is
https://github.com/checkstyle/checkstyle/blob/fe36c48a8b75bb97f5b4c617e0a79dfb724ead0a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L95-L99

so, please create PR to fix this Check, such issue is "second" issue.

jxr98:
The errors from the failed test.
image
I think the first one is okay because `TokenTypes.LITERAL_CATCH` is of type int.
I am not sure about the second error.

romani:
We do not allow new mutation survival. Please investigate how to cover them by test. You can apply suggested mutation to code and run regression report on your local on all projects that we have in config. It will show what java code input can cover this.

jxr98:
Hi @romani. Thanks for explanation. About the nakedReceiverMutator, I didn't find any userful information about it. Does this mutator mean to replace `variableUsageNode = variableUsageExpressions.get(0).getFirstChild();` with `variableUsageExpressions.get(0).getFirstChild();`

nrmancuso:
@jxr98 we have a wiki on pitest, see here: https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports#experimental-naked-receiver

jxr98:
> @jxr98 we have a wiki on pitest, see here: https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports#experimental-naked-receiver

Thanks you!

romani:
@jxr98 , if you are stuck with such update, it is not easy update, please send us fix for some "third" issue in parallel.

jxr98:
Okay, will solve the two mutations first.

jxr98:
Solved the second mutation, but haven't figured out how to solve the first. I have run regression tests on most of the projects of the first mutation, but there was no difference. Will run regression tests on all projects tmr.

romani:
@Vyom-Yadav , please assist here.

jxr98:
I have removed the surviving mutations and an unnecessary supressed mutation. Ready for review.

romani:
@jxr98 , while we are working on review, please start working on other issue.

Vyom-Yadav:
@jxr98 Please generate diff report

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
```diff
@@ -75,6 +75,7 @@ public void testGeneralLogic() throws Exception {
             "1001:9: " + getCheckMessage(MSG_KEY, "a", 3, 1),
             "1036:9: " + getCheckMessage(MSG_KEY, "c", 3, 1),
             "1066:9: " + getCheckMessage(MSG_KEY, "a", 4, 1),
+            "1075:9: " + getCheckMessage(MSG_KEY, "a", 2, 1),
```
Vyom-Yadav, line 78: Please do not add new test cases in existing input files. We have a line limit of 120 but this file is already very big, so it is exempted. 

You can create a new input file for every case you are adding.

Vyom-Yadav:
@jxr98 I don't think it is logically correct yet.
In https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c7ed4ae_2023173053/reports/diff/elasticsearch/xref/home/runner/work/checkstyle/checkstyle/contribution/checkstyle-tester/repositories/elasticsearch/src/test/java/org/elasticsearch/index/engine/InternalEngineTests.java.html#L1656.

It changed the violation from:
`Distance between variable 'indexSettingsService' declaration and its first usage is 5, but allowed 1. Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value).`

to:
`Distance between variable 'indexSettingsService' declaration and its first usage is 4, but allowed 1. Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value).`

Actually, there should be no violation at all, example is similar to:
`config.xml`
```xml
<!DOCTYPE module PUBLIC

<module name="Checker">
  <module name="TreeWalker">
    </module>
  </module>
</module>
```
`Test.java`
```java
public class Test {
    void method() {
        IndexSettingsService indexSettingsService = new IndexSettingsService(shardId.index(), indexSettings); // no violation
        Store store = createStore();
        Translog translog = createTranslog();
        final Engine engine = new InternalEngine(config(indexSettingsService, store, translog, createMergeScheduler()));
    }
}
```
```bash
$ java -jar /home/vyom/Downloads/checkstyle-10.9.2-all.jar -c config.xml Test.java
Starting audit...
Audit done.
```

Variable initialization statements are not considered as distance, the same should be the behavior with `try-with-resources`.

In the link I posted above, the user can't adjust the distance anyhow to solve the problem.

Current logic does not relate with `TokenTypes.RESOURCES` or `TokenTypes.LITERAL_TRY`, which should be incorporated to solve the problem.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral2.java
```diff
@@ -54,13 +58,38 @@ private void p() {
...
+            while((length=in.read(buf))!=-1){
+                System.out.println(buf.toString());
+            }
+        }
+    }
+
+    void methodTry() {
+        File a = new File(""); // violation 'Distance .* is 2.'
+        p();
+        p();
```
Vyom-Yadav, line 84: For all examples, don't have same-name method calls in between. If you use the same name method calls, they are considered initialization. Considering it as initialization is broken and it is a known bug and it will lead to wrong test results. PR for this is already open https://github.com/checkstyle/checkstyle/pull/11672

jxr98:
Thanks for explanation. It still take some time to understand the logic. But does this check have other bugs? This is the [regresion test report](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c7ed4ae_2023173053/reports/diff/elasticsearch/index.html#A1). Why the distance here is 2? I think it is only 1.

jxr98:
GitHub, generate report

jxr98:
Hi everyone. I have made two changes in `searchVariableUsageExpressions()`. 
1. Skipped the `Resource` nodes which don't contain the variable. 
2. Check the `TYPE` or `MODIFIERS` nodes which don't contain the variable. The `TYPE` or `MODIFIERS` nodes shouldn't cause `distance++`.
I have run some regression tests locally. It seemed like it was right. I haven't completed the tests cases. Just want to check if my modification is right. Thanks for reviewing.

romani:
@jxr98, pitest still a problem, please fix

jxr98:
working on pitest

nrmancuso:
@Vyom-Yadav please continue review

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral2.java
```diff
@@ -53,14 +56,4 @@ private void p() {
         } else {
         }
     }
-
-    void method() throws Exception {
-        // Until https://github.com/checkstyle/checkstyle/issues/11968
-        String a = ""; // violation 'Distance between .* declaration and its first usage is 2.'
```
romani, line 59: please restore.

We do not remove Input data, we need all types of weird code to be in out test suite.
All we do is extension of inputs or changing trailing comments.

config/pitest-suppressions/pitest-coding-1-suppressions.xml
```diff
@@ -156,10 +156,10 @@
   <mutation unstable="false">
     <sourceFile>VariableDeclarationUsageDistanceCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck</mutatedClass>
-    <mutatedMethod>getFirstNodeInsideTryCatchFinallyBlocks</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getLastChild with receiver</description>
-    <lineContent>final DetailAST finalBlock = currentNode.getLastChild();</lineContent>
+    <mutatedMethod>calculateDistanceBetweenScopes</mutatedMethod>
```
romani, line 159: we need to kill this mutation.
we do not allow new technical dept.

romani (changes requested):
after  quick glance:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral2.java
jxr98: I have the same type of tests in the new input file. But I will restore it.

config/pitest-suppressions/pitest-coding-1-suppressions.xml
jxr98: Done.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral2.java
romani: Please restore.

jxr98: issue11968 is the issue I am addressing, so I deleted this comment

Vyom-Yadav:
GitHub, generate report

romani:
@jxr98 , please put all cases that are not pointed by mentors to Inputs. If you see some other interesting code sample, please add it also.

We should not rely much on regression report, ideally all possible code should be in our Inputs.

jxr98:
The regression test code:
```
public class Test {
        Integer a = 1;
        class Inner {
            int i;

            Inner(int i) {
                this.i = i;
            }

            public int result() {
                return a * 1000 + i;
            }
        }
        SAM s = v -> new Inner(v) {
        }.result();
        if (s.m(2) != 1002) {
            throw new AssertionError("Unexpected value!");
        }
    }

    interface SAM {
        int m(int v);
    }
}
```
[1](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/252e206_2023054402/reports/diff/openjdk17/xref/home/runner/work/checkstyle/checkstyle/contribution/checkstyle-tester/repositories/openjdk17/test/langtools/tools/javac/lambda/LambdaConv30.java.html#L34) This changs from 4 to 2 because I added a rule to prevent Modifier and Type to cause distance to increase one. The decreasing 2 are from `MODIFIERS[11x12]` and `TYPE[11x19]`. The AST is as following. 
image. Another 1 distance is caused by `CTOR_DEF[7x12]` which I haven't added a rule for it.

jxr98:
The regression test code:
```
public class Test2 {
    public static void main() {

        ImageComparator comparator = new StrictImageComparator();
        String timestampName = timeStamp(imageID + "-" + lafShortName(), "png");
        class ImageWaitable implements Waitable {
            BufferedImage image;

            @Override
            public Object actionProduced(Object obj) {
                image = supplier.get();
                doSave(image, timestampName);
                return comparator.compare(reference, image) ? null : image;
            }

            @Override
            public String getDescription() {
                return "Waiting screen image to change";
            }
        }
        ImageWaitable waitable = new ImageWaitable();
        Waiter waiter = new Waiter(waitable);
        waiter.setTimeouts(timeouts);
        waiter.waitAction(null);
        return waitable.image;
    }
}
```
[2](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/252e206_2023054402/reports/diff/openjdk17/xref/home/runner/work/checkstyle/checkstyle/contribution/checkstyle-tester/repositories/openjdk17/test/jdk/sanity/client/lib/Extensions/src/org/jemmy2ext/JemmyExt.java.html#L254) This changes from 7 to 5. The decreasd 2 are caused by `MODIFIERS[9x12]` and `TYPE[10x19]`. The AST is as following.
image. The other 4 are caused by `implements[6x28]`, `PARAMETERS[10x41]`, `EXPR[11x22]` and `EXPR[12x22]`.

jxr98:
> 

Yes, agreed. Some test cases are not related with try resources. Will add them in separate issues.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -692,10 +692,17 @@ else if (variableUsageExpressions.isEmpty()) {
...
+                    currentStatementAst = currentStatementAst.getNextSibling();
+                    continue;
+                }
                 // If expression doesn't contain variable and this variable
                 // hasn't been met yet, then distance + 1.
                 else if (variableUsageExpressions.isEmpty()
-                        && currentStatementAst.getType() != TokenTypes.VARIABLE_DEF) {
+                        && currentStatementAst.getType() != TokenTypes.VARIABLE_DEF
+                        && currentStatementAst.getType() != TokenTypes.TYPE
+                        && currentStatementAst.getType() != TokenTypes.MODIFIERS) {
```
Vyom-Yadav, line 705: The regression report looks correct but I am skeptical about the logic present here. Why is it not modified at lines 605-654? Have you tried that approach?

jxr98: I combined several if conditions into the `isStatementVariableDefinition()` function.

jxr98:
GitHub, generate report

jxr98:
Hi @Vyom-Yadav, could you plz continue reviewing? Thanks.

romani:
Why distance becomes 2 ?
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/585d640_2023042012/reports/diff/spring-integration/index.html#A1

It should stay 4, as before, as first usage is in class, so class line definition is start of usage.

jxr98:
```
public class Test {
    public void method() {
        AtomicInteger instances = new AtomicInteger();

        class InterceptorFactory extends HelloWorldInterceptorFactory {

            @Override
            public TcpConnectionInterceptorSupport getInterceptor() {
                return new TcpConnectionInterceptorSupport() {

                    private final int instance = instances.incrementAndGet();

                    @Override
                    public void addNewConnection(TcpConnection connection) {
                        addOrder.add(this.instance);
                        interceptorAddCalled.countDown();
                        super.addNewConnection(connection);
                    }

                    @Override
                    public synchronized void removeDeadConnection(TcpConnection connection) {
                        super.removeDeadConnection(connection);
                        // can be called multiple times on different threads.
                        if (!remOrder.contains(this.instance)) {
                            remOrder.add(this.instance);
                            interceptorRemCalled.countDown();
                        }
                    }

                };
            }

        }
    }
}
```
The previous four was cause by `extends`(8:19), `modifier Override`(7:12), `type TcpConnectionInterceptorSupport`(8:19) and the expression where variable `instances` is used. The distance decreased two because modifiers and type are also treated as variable definition in the [code snippet](https://github.com/jxr98/checkstyle/blob/585d6401feff77f3c1178143b1b363220ef0fe31/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L1014-L1028). I have read the documentation for variable distance check. From my opinion, modifiers and types are indicators of definitions. But I am not sure if it is right in all real-world cases.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -1011,4 +1011,18 @@ private boolean isVariableMatchesIgnorePattern(String variable) {
...
+     *        Type of the statement.
+     * @return true if the statement is, otherwise false.
+     */
+    private static boolean isStatementVariableDefinition(int type) {
+        return type == TokenTypes.VARIABLE_DEF
+                || type == TokenTypes.TYPE
+                || type == TokenTypes.MODIFIERS
+                || type == TokenTypes.RESOURCE;
+    }
+
```
Vyom-Yadav, line 1027: This method is confusing, `isStatementVariableDefinition`, I think the correct way of checking whether any AST token is part of the variable definition is by checking if one of its ancestor ast is `VARIABLE_DEF`.

If this method does not aim to do that, the name should be changed or split into different methods if a single method can not have a proper name.

Vyom-Yadav (changes requested):
Please rebase onto latest master

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
jxr98: > This method is confusing, `isStatementVariableDefinition`, I think the correct way of checking whether any AST token is part of the variable definition is by checking if one of its ancestor ast is `VARIABLE_DEF`.
> 
> If this method does not aim to do that, the name should be changed or split into different methods if a single method can not have a proper name.

Okay. Done. I have changed the name of the function.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
romani: Combining is ok. 
But method name should describe what it does.

You do `!myfinctio(ast)) {distance++; }`

So your name should about ignoring of distance extension.
Name it as `isZeroDistanceToken`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
jxr98: Because I think the code will be easier to understand with conditions combined into a function.

Vyom-Yadav: But the function can't be named appropriately, combining conditions that do a lot of things without appropriate naming is not worth it.

jxr98: Romani suggested a name as isZeroDistanceToken. It makes more sense than isDefinition.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
jxr98: Thanks. The name makes more sense.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -1011,4 +1011,19 @@ private boolean isVariableMatchesIgnorePattern(String variable) {
         return matcher.matches();
     }
 
+    /**
+     * Check if the statement is zero-distance token.
```
romani, line 1015: Check if the toke should ignored for distance counting .

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -1011,4 +1011,19 @@ private boolean isVariableMatchesIgnorePattern(String variable) {
...
+     * Modifiers, types and resources are indicators of function or class or variable definitions.
```
romani, line 1016: This is not required, as it is clearly visible from code.
Remove it.
We need in comment something that code might not be conveying to reader.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
jxr98: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
jxr98: Done.

jxr98:
image
Could we set the time limit larger? The tests were easy to fail due to internet connection.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -1011,4 +1011,18 @@ private boolean isVariableMatchesIgnorePattern(String variable) {
         return matcher.matches();
     }
 
+    /**
+     * Check if the token should be ignored for distance counting.
+     *
+     * @param type
+     *        Type of the statement.
+     * @return true if the statement is, otherwise false.
```
romani, line 1019: Remove all `statement` words from this method

romani (changes requested):
Items

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral3.java
```diff
@@ -25,7 +25,7 @@ void mm() {
 
     public void method2() {
         // until https://github.com/checkstyle/checkstyle/issues/13011
-        Integer a = 5;  // violation 'Distance .* is 4.'
+        Integer a = 5;  // violation 'Distance .* is 2.'
         class AClass extends Parent {
```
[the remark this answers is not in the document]
jxr98: Yes. But I haven't looked into that issue yet.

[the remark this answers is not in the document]
jxr98: Has solved `extends` and `implements`. The issue of `@Override` still exists. It may be solved by adding a case for function definition in the `calculateDistanceBetweenScopes()` function.

jxr98:
One CI test failed: 
image

jxr98:
GitHub, generate report

jxr98:
GitHub, generate report

jxr98:
I will run the whole regression test locally, and upload the report.

romani:
Please upload report to GitHub, see instructions at https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/README_MANUAL_EXECUTION.md#executing-diffgroovy

jxr98:
Regression report: https://jxr98.github.io/xinrui_report/

romani:
Spring, sputbugs, strut are good.

jxr98:
https://jxr98.github.io/xinrui_report/spoon/index.html#A1
This is due to the method_definition. I will add the code for the method_definition in a separate PR because it is kind of hard to maintain all changes in one pr.

jxr98:
GitHub, generate report

romani:
We should split this extra problems on distance counting to separate issues.
If we reported violation before your changes and continue to report violation and only distance value is different and use to be bad and continue to be bad - this is acceptable regression. It should be separate issue to resolve it. Just create new issue as we see a problem.

romani:
GitHub, generate report

romani:
While we having issue with GitHub action, please share report from local execution.

jxr98:
> While we having issue with GitHub action, please share report from local execution.
> 
> Code looks ok, if regression report is good, we can finalize review.

I haven't changed anything later than https://jxr98.github.io/xinrui_report/. I didn't run openjdk17 project bc there was an error about the config file. I checked all differences. 

https://jxr98.github.io/xinrui_report/Orekit/index.html#A1
I counted this manually. If the calls to same function count as 1 distance, then it is right.

The only suspicious one is https://jxr98.github.io/xinrui_report/jOOL/index.html#A2. I haven't figured out the pattern. Will add to this issue#13011 later.

romani:
GitHub, generate report

romani:
GitHub, generate report

romani:
GitHub, generate report

romani:
GitHub, generate report

romani:
GitHub, generate report

romani:
GitHub, generate report

romani:
GitHub, generate report

romani:
GitHub, generate report

romani:
GitHub, generate report

romani:
GitHub, generate report

romani:
GitHub, generate report

romani:
@jxr98 , please register separate issue on https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c05c8be_2023022727/reports/diff/jOOL/index.html#A2 distance should 1, so no violation.

romani (approved):
As separate issue is created , I am ok to merge

romani:
@jxr98 , please do final action.

@Vyom-Yadav , please finalize review.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -1011,4 +1011,20 @@ private boolean isVariableMatchesIgnorePattern(String variable) {
         return matcher.matches();
     }
 
+    /**
+     * Check if the token should be ignored for distance counting.
+     *
+     * @param type
+     *        Type.
+     * @return true if it is, otherwise false.
+     */
```
Vyom-Yadav, line 1020:
```suggestion
/**                                                                                                         
 * Check if the token should be ignored for distance counting.                                              
 *                                                                                                          
 * @param type the token type of the ast node                                                               
 * @return {@code true} if it is a zero distance token and the token should be ignored for distance counting
 */                                                                                                         
```

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
```diff
@@ -323,15 +322,32 @@ public void testVariableDeclarationUsageDistanceSwitchExpressions2() throws Exce
...
         };
 
         verifyWithInlineConfigParser(
-                getPath("InputVariableDeclarationUsageDistanceGeneral3.java"),
-                expected);
+                getPath("InputVariableDeclarationUsageDistanceGeneral3.java"), expected);
+    }
+
+    @Test
+    public void testGeneral4Class() throws Exception {
```
Vyom-Yadav, line 333: `testGeneralClass4`, please change it at other places in the test file :)

Vyom-Yadav (changes requested):
Last Items:

Vyom-Yadav:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c05c8be_2023022727/reports/diff/jOOL/index.html#A2

Rightly as @romani said, this should be a separate issue. Variables declared in methods but used in local classes should be either completely ignored or some intelligent (looks challenging) logic should be put in place.

romani:
@jxr98 , please reply "done" to each review item. It really speedups review. Please share link to issue.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
jxr98: Done.

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
jxr98: Done. Thanks for pointing this out.

jxr98:
Thanks for reminding. I just finished it. New issue #13038. Thanks

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
Vyom-Yadav: Resolved

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
Vyom-Yadav: Resolved.

Vyom-Yadav (approved):
Great Work @jxr98 !

jxr98:
> Great Work @jxr98 !

Thanks!

romani:
@nrmancuso , please finalize. 
This PR was open for a while, we need fresh eyes to look at it

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -692,10 +692,10 @@ else if (variableUsageExpressions.isEmpty()) {
                 if (isChild(currentStatementAst, variableAst)) {
                     variableUsageExpressions.add(currentStatementAst);
                 }
-                // If expression doesn't contain variable and this variable
+                // If expression is not zero-distance token and this variable
                 // hasn't been met yet, then distance + 1.
```
nrmancuso, line 696: We can remove this comment, new method name and clear logic make it obvious what is going on here.

nrmancuso (changes requested):
Items:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral3.java
```diff
@@ -10,8 +10,7 @@
 
 package com.puppycrawl.tools.checkstyle.checks.coding.variabledeclarationusagedistance;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
```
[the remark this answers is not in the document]
romani: I thought we have a check to catch star imports.
Let's look why it is not caught by CI.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -1011,4 +1011,20 @@ private boolean isVariableMatchesIgnorePattern(String variable) {
...
+     *        Token type of the ast node.
+     * @return true if it should be ignored for distance counting, otherwise false.
+     */
+    private static boolean isZeroDistanceToken(int type) {
+        return type == TokenTypes.VARIABLE_DEF
+                || type == TokenTypes.TYPE
+                || type == TokenTypes.MODIFIERS
+                || type == TokenTypes.RESOURCE
+                || type == TokenTypes.EXTENDS_CLAUSE
+                || type == TokenTypes.IMPLEMENTS_CLAUSE;
```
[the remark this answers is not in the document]
romani: Please put explanation to javadoc.
Let's code remember what is not obvious.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
jxr98: Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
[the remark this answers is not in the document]
jxr98: Done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral3.java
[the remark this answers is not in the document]
nrmancuso: @romani i do not think we such validation on test inputs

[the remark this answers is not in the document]
romani: Ok, inputs can use any style of imports. The wilder the better :).

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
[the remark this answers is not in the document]
nrmancuso: I can't help but feel that we are missing tokens from this list. One comes immediately to mind: `LITERAL_PERMITS`. Please create an test using this token to see if we need it, input file can be very simple.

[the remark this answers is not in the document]
jxr98: I searched online, and it seemed that a local class can't be a sealed class. So the keyword permit won't be a problem.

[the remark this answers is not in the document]
nrmancuso: >I searched online, and it seemed that a local class couldn't be a sealed class. 

javac should be the source of truth.


```
➜  src cat Test.java
public class Test {
    void m(){
        sealed interface I permits C { }
        final class C implements I { }
    }
}%                                                                                                         ➜  src cat Test.java
public class Test {
    void m(){
        sealed interface I permits C { }
        final class C implements I { }
    }
}
➜  src javac Test.java 
Test.java:3: error: sealed or non-sealed local classes are not allowed
        sealed interface I permits C { }
        ^
Test.java:3: error: invalid permits clause
        sealed interface I permits C { }
                           ^
  (interface must be sealed)
2 errors
➜  src javac --version
javac 19.0.1

```

nrmancuso (approved):

Issue #11641, opened 2022-05-15
SuppressWarningsHolder should allow usage of simple name of Checks in aliasList


romani:
https://checkstyle.org/config_annotation.html#SuppressWarningsHolder

```
$ cat Test.java
class Test {
  @SuppressWarnings("paramnum")
  public void needsLotsOfParameters( int a, 
  int b, int c, int d, int e, int f, int g, int h) {
    // ...
  }
}

$ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
       "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
	<property name="charset" value="UTF-8" />
	
   
<module name="TreeWalker">
   <module name="ParameterNumber">
   </module>
   
   <module name="SuppressWarningsHolder">
        <property name="aliasList" value="ParameterNumber=paramnum" />
  </module>
</module>

   <module name="SuppressWarningsFilter" />
</module>

$ java -jar checkstyle-10.1-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /home/kevin/Desktop/C_Style/Test.java:4:15: More than 7 parameters (found 8). [ParameterNumber]
Audit done.
```

Expected: no violation.

workaround is to use fully qualified name of Check.
see details - https://github.com/checkstyle/checkstyle/pull/11542#issuecomment-1126821468

Kevin222004:
we have to create a method in https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/SuppressWarningsHolder.java
am i right

romani:
Create unit test method to reproduce problem by Input file first. 
Second , update holder to have in map of names simple names also .

rnveach:
https://checkstyle.org/config_annotation.html#SuppressWarningsHolder_Properties
>  The attribute is the fully qualified name of the Check and value is its alias. 

Documentation states that it must be the fully qualified name.

What documentation may not state, is that it looks like it can also be the Check's ID instead of the name.
https://github.com/checkstyle/checkstyle/blob/399f1b48b79731d65e824eb927971ef6174c3463/src/main/java/com/puppycrawl/tools/checkstyle/checks/SuppressWarningsHolder.java#L242

[...]

PR #12833, opened 2023-03-13
supplemental: test cases added

Kevin222004:
supplemental: test cases added

While working on https://github.com/checkstyle/checkstyle/issues/11641 I noticed that I don't found test cases which use alias list property

### DIFF
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/suppresswarningsholder/InputSuppressWarningsHolderAlias.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/suppresswarningsholder/InputSuppressWarningsHolderAlias.java
@@ -1,0 +1,56 @@
+/*
+com.puppycrawl.tools.checkstyle.checks.SuppressWarningsHolder
+aliasList = com.puppycrawl.tools.checkstyle.checks.sizes.ParameterNumberCheck=paramnum
+
+com.puppycrawl.tools.checkstyle.filters.SuppressWarningsFilter
+
+com.puppycrawl.tools.checkstyle.checks.sizes.ParameterNumberCheck
+max = (default)7
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.suppresswarningsholder;
+
+public class InputSuppressWarningsHolderAlias {
+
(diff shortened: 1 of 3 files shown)


src/test/resources/com/puppycrawl/tools/checkstyle/checks/suppresswarningsholder/InputSuppressWarningsHolderAlias.java
```diff
@@ -0,0 +1,56 @@
+/*
+com.puppycrawl.tools.checkstyle.checks.SuppressWarningsHolder
+aliasList = com.puppycrawl.tools.checkstyle.checks.sizes.ParameterNumberCheck=paramnum
```
romani, line 3: It is not same as `<property name="aliasList" value="ParameterNumber=paramnum" />`
As you reference in issue.
You should have `aliasList=ParameterNumber=paramnum`
Or second = to be escaped.

romani (changes requested):
Items

src/test/resources/com/puppycrawl/tools/checkstyle/checks/suppresswarningsholder/InputSuppressWarningsHolderAlias.java
romani: Or we need both formats in tests

romani: Or we need both formats in tests

Kevin222004: @romani the issue https://github.com/checkstyle/checkstyle/issues/11641 is all about supporting the simple name `aliasList=ParameterNumber=paramnum`.

 Right now it only supports the fully qualified name https://github.com/checkstyle/checkstyle/pull/11542#issuecomment-1126821468  

I will most probably send pr for fixing this by today https://github.com/checkstyle/checkstyle/issues/11641.

romani: Ok.

PR #12850, opened 2023-03-16
minor: move run key at last spot

stoyanK7:
We keep `run` key always at last spot. For example,
https://github.com/checkstyle/checkstyle/blob/665f58f178e282e5565b5782ab1b33237d32f547/.github/workflows/release-maven-perform.yml#L38-L54
https://github.com/checkstyle/checkstyle/blob/665f58f178e282e5565b5782ab1b33237d32f547/.github/workflows/release-update-github-page.yml#L41-L45


romani:
pitest failed on download

Issue #12898, opened 2023-03-23
Typo in Documentation of VariableDeclarationUsageDistance


jxr98:
In [documentation](https://checkstyle.org/config_coding.html#VariableDeclarationUsageDistance), 
```
final int PI;
PI = 3.14;
```
There's a typo. PI should be of type double or float.

jxr98:
If it is approved, I think I can work on it •ᴗ•.

romani:
@jxr98, please send pr

romani:
Fix merged

PR #12899, opened 2023-03-23
Issue #12898: changed type of PI from int to double

jxr98:
Issue #12898
I changed the type of PI from int to double in the [documentation](https://checkstyle.org/config_coding.html#VariableDeclarationUsageDistance).

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/TokenTypes.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/TokenTypes.java
@@ -312,7 +312,7 @@ public final class TokenTypes {
      *
      * <p>For example:</p>
      * <pre>
-     * final int PI = 3.14;
+     * final double PI = 3.14;
      * </pre>
      * <p>parses as:</p>
      * <pre>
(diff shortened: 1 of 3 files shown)


nrmancuso:
@jxr98 please grep by `int PI` to see if we copy-pasted this elsewhere in documentation

jxr98:
> @jxr98 please grep by `int PI` to see if we copy-pasted this elsewhere in documentation

Done.
image

src/main/java/com/puppycrawl/tools/checkstyle/api/TokenTypes.java
```diff
@@ -312,7 +312,7 @@ public final class TokenTypes {
      *
      * <p>For example:</p>
      * <pre>
-     * final int PI = 3.14;
+     * final double PI = 3.14;
```
romani, line 315: `--LITERAL_INT`

This line below should be changed .

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/api/TokenTypes.java
jxr98: Oh right. Thanks.

romani (approved):
Ok to merge

PR #12904, opened 2023-03-24
Issue #12890: Kill surviving mutation in PackageDeclarationCheck.java

### LINKED ISSUE #12890: Kill surviving mutation in PackageDeclarationCheck.java
Kill mutation :

https://github.com/checkstyle/checkstyle/blob/54df97c85d5fae09e38cda3c83781ac83b857930/config/pitest-suppressions/pitest-coding-1-suppressions.xml#L75-L82

shubh220922:
Resolves #12890 .

Mutation killed.

### DIFF
--- a/config/pitest-suppressions/pitest-coding-1-suppressions.xml
+++ b/config/pitest-suppressions/pitest-coding-1-suppressions.xml
@@ -64,15 +64,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>PackageDeclarationCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.PackageDeclarationCheck</mutatedClass>
-    <mutatedMethod>beginTree</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable defined</description>
-    <lineContent>defined = false;</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>ReturnCountCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.ReturnCountCheck</mutatedClass>
     <mutatedMethod>leaveToken</mutatedMethod>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/PackageDeclarationCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/PackageDeclarationCheckTest.java
@@ -38,6 +38,17 @@ public class PackageDeclarationCheckTest extends AbstractModuleTestSupport {
     }
 
     @Test
+    public void testInMultiThreadedEnvironment() throws Exception {
+
+        final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
+
+        verifyWithInlineConfigParser(
+                getPath("InputPackageDeclarationPlain.java"),
+                getPath("InputPackageDeclarationWithCommentOnly.java"),
+                expected);
+    }
+
+    @Test
     public void testDefaultNoPackage() throws Exception {
 
         final String[] expected = {


shubh220922:
@romani @nrmancuso 

Code was redundant. Removed it to kill mutation. No test cases required.

nrmancuso:
“code was redundant” - please provide an explanation with all details

shubh220922:
@nrmancuso @romani 

### Complete analysis report

Now we know that in java if we declare a boolean value, it is declared ``false`` by default. This is basically the heart of why the mutation is surviving at the line ``defined = false;`` in the ``beginTree()`` method.  We are making a boolean value equal to false when it is already equal to false and hence, when PIT comes to that line and removes assignment to false that is, it changes ``defined = false`` to ``defined = defined`` even then none of the tests fail and the code doesn't break as writing ``defined = defined`` doens't change anything. To better understand what I am saying, refer the below image. 
image

The value of ``defined`` is always false. 

### How can we kill this mutation

#### Method 1: Remove the piece of code (works! (this needs review of maintainer though))

The problem with this method is the very reason the ``beginTree()`` method ever existed. It was to initialize ``defined`` as ``false`` and to not bother about what its default value is as it is a good coding practice to initialize the variables irrespective of what their default values are.

Removing the code would mean we are relying on the default value of ``defined`` which is not good "coding practice". 

Work-around)
We can mention a small comment in the source code addressing this and remove ``beginTree()`` method 
anyways.

#### Method 2: Write Input files to kill mutation 
To the best of what I know, this method will not work for the following reason.

When we study the call stack of how ``beginTree()`` is being called, we notice that it is being called from ``TreeWalker.java``(image below) (notice ``check.beingTree()``, that's where it's being called from): 
image

Now the thing is no matter whether the ``package`` in the input file is defined or not, ``check.beginTree()`` would be invoked always. This means we will again execute the ``beginTree()`` method and again execute the line ``defined = false;``. 

Note that the purpose of ``defined`` is to be true when there is a package in input file and false when the ``package`` is not mentioned. In our case, ``defined`` is always initialized to ``false`` because ``beginTree()`` is being called always and that's a problem as no matter which input file we write, none of our test cases will ever be able to differentiate if a value (``defined`` in this case) which is ``false`` by default is being initialized to ``false`` again. 

(I'm not sure if a work-around for this is possible. Maintainer help required.)

### Method 3: Write pure JUnit truth assert statements 

We can write pure JUnit assert statements to exclusively call ``beginTree()`` method. But, then again the issue is what am I going to write in the test that helps me catch the fact that assignment to ``defined`` has been removed. This is not possible as disscussed above as even when the assignment is removed, value of ``defined`` is false. I cannot write any JUnit tests to catch the difference as there is not difference. 

@nrmancuso  what do you think. Did I miss anything?

shubh220922:
@nrmancuso 
CI failure not related to my PR?
https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=13758&view=logs&j=e663805c-ee13-5a24-f979-d725e0346c8e&t=40ff6e43-c5d7-5753-f947-23012624b0c6&l=724

shubh220922:
@romani @nrmancuso 
This is from the screenshot of checkstyle:
image
If ``beginTree()`` is only used for initialization or carrying out some starting work then i don't think there is anything wrong with removing it from this file as it is upto us to do that "starting work".

shubh220922:
@nrmancuso @romani 
Please review and let me know if I need to change anything.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/PackageDeclarationCheck.java
```diff
@@ -145,11 +145,6 @@ public int[] getAcceptableTokens() {
         return getRequiredTokens();
     }
 
-    @Override
-    public void beginTree(DetailAST ast) {
-        defined = false;
```
romani, line 150: https://github.com/checkstyle/checkstyle/blob/655f743503209544b1b87dc016da4ebfd10cee12/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/PackageDeclarationCheck.java#L103

It means that Check instance will stay same between two input java classes. It has to be reset his state on each new file.

To reproduce problem, handle two files, one with violation and without. There should be violation only on first.

Recheck git history on changes for this field to get more details why it is required.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/PackageDeclarationCheck.java
nrmancuso: @shubh220922 it is likely that we would need two concurrent executions of check in different threads to kill this mutation, if I understand documentation correctly.

shubh220922: @nrmancuso ok I'm looking into it.

shubh220922: @nrmancuso 
I studied multi threading as much as possible. I would request you to send link of the documentation you are talking about in https://github.com/checkstyle/checkstyle/pull/12904#discussion_r1163522418 and also point me to any pre existing multi threaded test in checkstyle's codebase as I have not written multi threaded tests before.

shubh220922: @nrmancuso @romani please let me know what I can do here.

nrmancuso: Looks like we have different instances of the check between runs, @shubh220922 you should be able to determine this by checking the hascode of the check object.

romani: in real execution Check instance is always the same for all files so this is why cleanup is required.
In test we need to simulate same,  it might require verify method that takes two Input classes for execution, I think we already has such method.

shubh220922: @romani could you be more precise. We do have ``verify()`` method that takes 2 input files at once but, it is only for those files that have same violations. Please refer the method below:
```java
/**
     * Performs verification of two files with their given file paths using specified
     * configuration of one file only. Also performs verification of the config specified
     * in the input file. This method needs to be implemented when two given files need to be
     * checked through a single check only.
     *
     * @param filePath1 file path of first file to verify
     * @param filePath2 file path of second file to verify
     * @param expected an array of expected messages
     * @throws Exception if exception occurs during verification process
     */
    protected final void verifyWithInlineConfigParser(String filePath1,
                                                      String filePath2,
                                                      String... expected)
            throws Exception {
                InlineConfigParser.parse(filePath1);
                testInputConfiguration.createConfiguration();
        verifyViolations(parsedConfig, filePath1, testInputConfiguration.getViolations());
        verifyViolations(parsedConfig, filePath2, testInputConfiguration.getViolations());
        verify(createChecker(parsedConfig),
                new File[] {new File(filePath1), new File(filePath2)},
                filePath1,
                expected);
    }
```
The method: 
```java
verify(createChecker(parsedConfig),
                new File[] {new File(filePath1), new File(filePath2)},
                filePath1,
                expected);
```
takes only one violation.

romani: >We do have verify() method that takes 2 input files at once but, it is only for those files that have same violations. 

Nothing in doc of it and implementation mention it.
It takes config from first input file, and validate both files as one run. 
Make two inputs with same inlined config. And use this method

romani: Second file should make `ast` as null
https://github.com/checkstyle/checkstyle/blob/70946a8f01ae95c5debf3f7ea10557daadea1cd6/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/PackageDeclarationCheck.java#L155

Please investigate if this possible.

Issue #7608, opened 2020-02-24
Update doc for DesignForExtension


pbludov:
https://checkstyle.org/config_design.html#DesignForExtension

Add example for each existing config to show what code is ok and what code cause violation.
Good example is https://checkstyle.org/config_naming.html#AbstractClassName
Look at referenced commits of #5832, to see what and how should be changed in code.

Please play with module/check by means of https://checkstyle.org/cmdline.html to understand how it works.

------------

PLEASE READ https://github.com/checkstyle/checkstyle/wiki/Good-practice-on-how-to-update-examples

sulkykookie:
I'm on it.

ImmortalRabbit:
I am on it.

aryaniiit002:
i want to work on this, can i work on it?

mananmehta3:
If the issue is not assigned I would like to work on it.

nick-mancuso:
@mananmehta3 if “I am on it” comments are more than a month old with no PR, feel free to add “I am on it” comment and proceed.

mananmehta3:
The I am on it message on this issue is around a year old and there is no PR.
So I guess that is okay.

Chandan-DK:
Hi. I just checked https://checkstyle.org/config_design.html#DesignForExtension.
It has examples for all the configs. So, why isn't this issue closed?
Thank you

romani:
https://checkstyle.org/config_design.html#DesignForExtension_Examples

We need example of code for default config.
We need to make all java codes to be same code to show user how only changes in config makes behavior (location of violations) different.

Chandan-DK:
I will work on this issue. 
I'll ask for any clarifications required when I start working on it

romani:
I highly recommend to review other PRs for doc update, example https://github.com/checkstyle/checkstyle/pull/12187 , to see where new contributors have most of misunderstanding and how to make update less time consuming for you.

Chandan-DK:
> We need example of code for default config. We need to make all java codes to be same code to show user how only changes in config makes behavior (location of violations) different.

@romani 
1. You mean that the examples for default config and all these properties (https://checkstyle.org/config_design.html#DesignForExtension_Properties)
should be similar to each other? 
If yes, shall I make changes to the existing examples provided for the properties? Because they are slightly different from each other 

[...]

PR #12908, opened 2023-03-25
Issue #7608: Update doc for DesignForExtension #7608 

HarshSharma0801:
Issue: #7608 
Added example for each existing config to show what code is ok and what code cause violation.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/design/DesignForExtensionCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/design/DesignForExtensionCheck.java
@@ -180,18 +180,28 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * to be designed for extension.
  * </p>
  * <pre>
- * &lt;module name=&quot;DesignForExtension&quot;&gt;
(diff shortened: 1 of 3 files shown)


romani:
please make single commit in PR and make sure CI is green.

src/main/java/com/puppycrawl/tools/checkstyle/checks/design/DesignForExtensionCheck.java
```diff
@@ -180,18 +180,28 @@
...
- *   public int foo2() {return 8;} // violation
+ *   public int foo2() {return 8;}  // violation, required annotation is not present .
+ *
+ *
+ *   &#64;Override
+ *   public int foo3() {
+ *       return 9 ;         // OK
+ *    }
+ *
+ *    public int foo() {return 7;}  // violation, required annotation is not present .
```
romani, line 202: please make changes to examples to make them same for java code in all examples.
the only diffrence should be is trailing comments that changing due to different config.

example: https://checkstyle.org/config_annotation.html#AnnotationLocation_Examples

romani (changes requested):
items:

PR #12911, opened 2023-03-25
Issue #12840: Add repository owner environment variable in release process

### LINKED ISSUE #12840: Add repository owner environment variable in release process
Whenever I am testing changes in the release process, I often have to make such changes to point to my local fork.

https://github.com/stoyanK7/checkstyle/commit/2600b41317e42d2b6a283ac53b3c35009fcf5401
Screenshot from 2023-03-14 18-49-36

---
We can use the [`github.repository_owner` variable](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) and set it as an env variable for example. This will ease troubleshooting/testing those workflows in the future if they break.
```yml
      - name: Run Shell Script
        env:
          REPOSITORY_OWNER: ${{ github.repository_owner }}
        run: |
          ./.ci/some-release-script.sh
```

```sh
checkForVariable "REPOSITORY_OWNER"
...
-X GET https://api.github.com/repos/$REPOSITORY_OWNER/checkstyle/milestones?state=open \
```

I should have done this proposal long ago to be honest..

stoyanK7:
Resolves #12840 

---
Workflow run: https://github.com/stoyanK7/checkstyle/actions/runs/4520382682
The run fails, but that is because I do not have Twitter credentials. I had to manually point to my Sourceforge project, but that is not tied with GitHub so nothing can be done there.

What I noticed is that we can maybe make `release-maven-perform` have a different profile (https://github.com/stoyanK7/checkstyle/commit/08f79b22ba7c8f381be22b968ca01b3409ba272b) to deploy to local if `$REPOSITORY_OWNER` is not `checkstyle`. That way anyone can run/debug the whole release process on their fork. This is a separate issue from this one though.

### DIFF
--- a/.ci/release-close-create-milestone.sh
+++ b/.ci/release-close-create-milestone.sh
@@ -5,15 +5,16 @@ set -e
 source ./.ci/util.sh
 
 checkForVariable "GITHUB_TOKEN"
+checkForVariable "REPOSITORY_OWNER"
 
 echo "Close previous milestone at github"
 MILESTONE_NUMBER=$(curl --fail-with-body -s \
                 -H "Authorization: token $GITHUB_TOKEN" \
-                -X GET https://api.github.com/repos/checkstyle/checkstyle/milestones?state=open \
+                -X GET https://api.github.com/repos/"$REPOSITORY_OWNER"/checkstyle/milestones?state=open \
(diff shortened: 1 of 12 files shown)


romani (approved):
Ok to merge

PR #12941, opened 2023-03-31
doc: remove extra space

7h3-3mp7y-m4n:


romani (changes requested):
Do same at https://github.com/checkstyle/checkstyle/blob/bf9b7b2cd8599df78d183244b437875c59c4e21c/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/MethodTypeParameterNameCheck.java#L46

Content should be same

7h3-3mp7y-m4n:
ok

PR #12946, opened 2023-04-01
Issue #12794: Diff-Report action does not post link to report in GitHub comment

### LINKED ISSUE #12794: Diff-Report action does not post link to report in GitHub comment
There is no link posted by the GitHub action when using `Report label:` property.

I suspect that because we append to the `message` it goes on to the next line and only first line is picked up by upcoming steps or it has something to do with `set-output`.

Noticed at: https://github.com/checkstyle/checkstyle/pull/11231#issuecomment-1454660189, https://github.com/checkstyle/checkstyle/pull/11231#issuecomment-1408476323

stoyanK7:
Resolves #12794

Multiline strings should be handled differently. Taken from GitHub actions documentation: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-a-multiline-string. We already applied the same solution in `site` workflow.

Before:
https://github.com/stoyanK7/checkstyle/pull/62#issuecomment-1493068289
Screenshot from 2023-04-01 20-56-58

Now:
https://github.com/stoyanK7/checkstyle/pull/62#issuecomment-1493073677
Screenshot from 2023-04-01 20-57-29

### DIFF
--- a/.github/workflows/diff-report.yml
+++ b/.github/workflows/diff-report.yml
@@ -226,7 +226,13 @@ jobs:
 
       - name: Set output
         id: out
-        run: echo "::set-output name=message::$(cat message)"
+        run: |
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          {
+            echo "message<<$EOF"
+            cat message
+            echo "$EOF"
+          } >> "$GITHUB_OUTPUT"
 
   # should be always last step
   send_message:
@@ -256,7 +262,13 @@ jobs:
 
       - name: Set message
         id: out
-        run: echo "::set-output name=message::$(cat message)"
+        run: |
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          {
+            echo "message<<$EOF"
+            cat message
+            echo "$EOF"
+          } >> "$GITHUB_OUTPUT"
 
       - name: Comment PR
         uses: checkstyle/contribution/comment-action@master


romani (approved):
Ok to merge

.github/workflows/diff-report.yml
```diff
@@ -256,7 +262,13 @@ jobs:
...
       - name: Set message
         id: out
-        run: echo "::set-output name=message::$(cat message)"
+        run: |
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          {
+            echo "message<<$EOF"
+            cat message
+            echo "$EOF"
+          } >> "$GITHUB_OUTPUT"
```
nrmancuso, line 271: I think we need to make it obvious to others that may not be experienced with bash what is going on here. As GitHub had their special `set-output` command that we replaced, we should create a script to use for appending multiple lines to workflow output. Please create an issue on this, I propose we do something like `./append-to-github-output.sh message`.

This will also help in future maintainence/debugging efforts, since we can check for existence of `message`, etc.

nrmancuso (changes requested):
Item:

.github/workflows/diff-report.yml
stoyanK7: Agreed. Opened at https://github.com/checkstyle/checkstyle/issues/12954.

nrmancuso (approved):

PR #12959, opened 2023-04-05
Issue #12954: Create script for appending text to $GITHUB_OUTPUT

### LINKED ISSUE #12954: Create script for appending text to $GITHUB_OUTPUT
From https://github.com/checkstyle/checkstyle/pull/12946#discussion_r1155903342

> I think we need to make it obvious to others that may not be experienced with bash what is going on here. As GitHub had their special `set-output` command that we replaced, we should create a script to use for appending multiple lines to workflow output. Please create an issue on this, I propose we do something like ` ./append-to-github-output.sh` message.
> 
> This will also help in future maintainence/debugging efforts, since we can check for existence of `message`, etc.

stoyanK7:
Resolves #12954
Test on local fork: https://github.com/stoyanK7/checkstyle/pull/62#issuecomment-1496137559
Screenshot from 2023-04-05 12-40-07

### DIFF
--- a/.ci/append-to-github-output.sh
+++ b/.ci/append-to-github-output.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+NAME=$1
+VALUE=$2
+
+if [[ -z "$NAME" || -z "$VALUE" ]]; then
+  echo "not all parameters are set"
+  echo "Usage: $BASH_SCRIPT <variable name> <variable value>"
+  exit 1
+fi
+
+# Select random value for EOF as a delimiter.
+EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+{
+  echo "$NAME<<$EOF"
+  echo "$VALUE"
+  echo "$EOF"
+} >> "$GITHUB_OUTPUT"
--- a/.github/workflows/diff-report.yml
+++ b/.github/workflows/diff-report.yml
@@ -227,12 +227,8 @@ jobs:
       - name: Set output
         id: out
         run: |
-          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
-          {
-            echo "message<<$EOF"
-            cat message
-            echo "$EOF"
-          } >> "$GITHUB_OUTPUT"
+          cd checkstyle
+          ./.ci/append-to-github-output.sh "message" "$(cat ../message)"
 
   # should be always last step
   send_message:
@@ -240,6 +236,8 @@ jobs:
     needs: [ parse_body, make_report ]
     if: failure() || success()
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
 
       - name: Get message
         env:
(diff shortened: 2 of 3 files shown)


.github/workflows/diff-report.yml
```diff
@@ -227,19 +227,17 @@ jobs:
...
+          ./.ci/append-to-github-output.sh "message" "$(cat ../message)"
 
   # should be always last step
   send_message:
     runs-on: ubuntu-latest
     needs: [ parse_body, make_report ]
     if: failure() || success()
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
romani, line 240: Does it make sense to use `fetch-depth: 0`

And align affected scripts to use v3 of action.

.ci/append-to-github-output.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+NAME=$1
+VALUE=$2
+
```
nrmancuso, line 7: Please add `cd $(dirname $0)` so that we can remove path from file name that we `cat`

.github/workflows/diff-report.yml
```diff
@@ -227,19 +227,17 @@ jobs:
...
         id: out
         run: |
-          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
-          {
-            echo "message<<$EOF"
-            cat message
-            echo "$EOF"
-          } >> "$GITHUB_OUTPUT"
+          cd checkstyle
+          ./.ci/append-to-github-output.sh "message" "$(cat ../message)"
```
nrmancuso, line 231: Let's handle `cat` in script

nrmancuso (changes requested):

.github/workflows/diff-report.yml
stoyanK7: Opened PR at https://github.com/checkstyle/checkstyle/pull/12960.

.ci/append-to-github-output.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+NAME=$1
+VALUE=$2
```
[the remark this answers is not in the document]
stoyanK7: We can. I was trying to be flexible with the script.

.github/workflows/diff-report.yml
romani: I am ok to keep script to deal with multiple line variables

.ci/append-to-github-output.sh
romani: Please confirm that this good idea.
I don't think so

nrmancuso: This helps us to avoid confusing paths in yaml. This script is for a very specific use case, better to make yaml as simple as possible.

nrmancuso: > 

This allows us to run this script from the pwd so that we do not need to worry about paths to message file

.github/workflows/diff-report.yml
nrmancuso: Idea of the script was to make yaml as simple as possible and keep all details inside script with comments and descriptive variable names.

.ci/append-to-github-output.sh
romani: One day such code from yml will go to shell script. So nested script, that we introduce now, should be simple to let it used by any means from any place.

.github/workflows/diff-report.yml
romani: If this was  original intend, we need to move all code out to shell script , not a part.

stoyanK7: I'd say it belongs in a separate file because it is the exact same code in both `site` and `diff-report`.

nrmancuso (approved):
ok

PR #12970, opened 2023-04-10
Issue #12696: Use Environment Files instead of set-output in GitHub Actions

### LINKED ISSUE #12696: Use Environment Files instead of `set-output` in GitHub Actions.
From https://github.com/checkstyle/checkstyle/actions/runs/4055751603/jobs/6979309020

`
Warning: The set-output command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
`

See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ for the new way to `set-output`.

stoyanK7:
Resolves #12696

### DIFF
--- a/.ci/append-to-github-output.sh
+++ b/.ci/append-to-github-output.sh
@@ -5,7 +5,7 @@ set -e
 NAME=$1
 VALUE=$2
 
-if [[ -z "$NAME" || -z "$VALUE" ]]; then
+if [[ "$#" != "2" ]]; then
   echo "not all parameters are set"
   echo "Usage: $BASH_SCRIPT <variable name> <variable value>"
   exit 1
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -23,6 +23,3 @@ jobs:
 
       - name: Check workflow files
         uses: docker://rhysd/actionlint:latest
-        with:
-          # Until https://github.com/checkstyle/checkstyle/issues/12696
-          args: -ignore ".* was deprecated.*"
--- a/.github/workflows/diff-report.yml
+++ b/.github/workflows/diff-report.yml
@@ -58,6 +58,9 @@ jobs:
       - run: 'echo We print it here for this action to work'
         if: 'true'
 
+      - name: Checkout repository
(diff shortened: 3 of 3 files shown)


.ci/append-to-github-output.sh
```diff
@@ -5,7 +5,7 @@ set -e
 NAME=$1
 VALUE=$2
 
-if [[ -z "$NAME" || -z "$VALUE" ]]; then
```
stoyanK7, line 8: Instead of asserting that 2 variables with non-zero length are passed, we can assert that just 2 variables are passed.
https://github.com/stoyanK7/checkstyle/actions/runs/4658168765/jobs/8243534520#step:6:29
```
Run grep "^Diff Regression projects:" text | cat > temp
not all parameters are set
Usage:  <variable name> <variable value>
```

There are checks below to ensure that those strings are not-empty.
https://github.com/checkstyle/checkstyle/blob/1346ff94acef40b7772b5170beeb5fbc5d5b7d5a/.github/workflows/diff-report.yml#L118-L135

romani (approved):
Ok to merge

.ci/append-to-github-output.sh
```diff
@@ -5,7 +5,7 @@ set -e
...
+if [[ $# != 2 ]]; then
```
nrmancuso, line 8: Why are we using string comparison operator on a number?

nrmancuso (changes requested):
Item:

.ci/append-to-github-output.sh
romani: https://stackoverflow.com/a/18568726/1015848

stoyanK7: I went with `!=` for readability. But you are right, I should put them in quotes.

Done.

nrmancuso: I meant to use `-ne` for integer comparison ;) but if we make everything string by `"`, that's fine too.

nrmancuso (approved):

PR #12993, opened 2023-04-17
infra: fix file download location and script pwd

nrmancuso:
Report is still failing, we are not running script from checkstyle repo directory, and we generate report from general workspace directory, so we need to download files there.

### DIFF
--- a/.ci/diff-report.sh
+++ b/.ci/diff-report.sh
@@ -5,7 +5,7 @@ source ./.ci/util.sh
 
 case $1 in
 
-# Downloads all files necessary to generate regression report.
+# Downloads all files necessary to generate regression report to the parent directory.
 download-files)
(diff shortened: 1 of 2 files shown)


.ci/diff-report.sh
```diff
@@ -17,27 +17,27 @@ download-files)
   curl --fail-with-body -X GET "${LINK}" \
     -H "Accept: application/vnd.github+json" \
     -H "Authorization: token $GITHUB_TOKEN" \
-    -o project.properties
+    -o ../project.properties
```
romani, line 20: `mkdir -p .ci-temp`

And keep all files in it. We should not go out of our repo folder.

romani (changes requested):
Items

Issue #12671, opened 2023-01-21
Resolve pitest for FinalLocalVariableCheck


Kevin222004:
Part of https://github.com/checkstyle/checkstyle/issues/12341

This issue contains all the mutations for `FinalLocalVariableCheck` from

https://github.com/checkstyle/checkstyle/blob/6647ab368fb58e4ca8c71c1ffc029d0122bce227/config/pitest-suppressions/pitest-coding-2-suppressions.xml#L39-L91

 profile pitest-coding-2-suppressions.xml

Kevin222004:
@romani please approve this so I start working on this

romani:
I highly recommend to not send PR as soon as at least one survival is killed.
It is completely ok to send few PRs for same issue. The smaller PR the easier review and quicker merge.

stoyanK7:
@Kevin222004 Out of curiosity, have you managed to kill any of those mutations?

Kevin222004:
@stoyanK7 no I have not done anything to kill this mutation

romani:
I removing self assignment

coderdeadpool:
I am on Mutation: 
https://github.com/checkstyle/checkstyle/blob/5768a5e30a00ad8c2ae80da5fb67961219acf7b6/.ci/pitest-suppressions/pitest-coding-2-suppressions.xml#L57-L64

coderdeadpool:
I am on Mutation:
https://github.com/checkstyle/checkstyle/blob/b9cd0cb3553a21795d27a0238625c96ed8f8a33d/config/pitest-suppressions/pitest-coding-2-suppressions.xml#L12-L19

coderdeadpool:
I am on Mutation:
https://github.com/checkstyle/checkstyle/blob/e1e578b472d2730f82b0963173ab462be5e54901/config/pitest-suppressions/pitest-coding-2-suppressions.xml#L12-L19

romani:
issue is resolved by inscope of project "Pitest Resolution" by some referenced above commits.

Issue #12341, opened 2022-10-23
Resolve Pitest Suppressions


nrmancuso:
This issue will be the main tracker issue for resolving Pitest suppressions.

## Summary
Surviving pit mutations are a form of [technical debt](https://en.wikipedia.org/wiki/Technical_debt), and we would like to kill all surviving mutations. **It is important to first understand what pitest is and how it works**. Please visit https://pitest.org/ to learn about this tool. Simply put, pitest changes our code using mutation operators (this changed code is now called a **mutation**), then runs our tests against this changed code. If some test fails due to the changed code, then the mutation is considered "killed". If not, then the mutation has survived, and we need to (ideally) find a test input to kill the mutation.

**The point of this issue is to find a test case that kills a surviving mutation.**

### How does Pitest work?
Let's take a look at a suppressed surviving mutation:
https://github.com/checkstyle/checkstyle/blob/5768a5e30a00ad8c2ae80da5fb67961219acf7b6/.ci/pitest-suppressions/pitest-blocks-suppressions.xml#L156-L163

The above suppression is for a mutation operator that took this original line of code:
https://github.com/checkstyle/checkstyle/blob/5768a5e30a00ad8c2ae80da5fb67961219acf7b6/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java#L651

And mutated it into this code (now called a mutant or mutation):
```java
lcurly = lcurly;
```

The mutation is considered to have "survived", which means that no test failed after the code was changed. We want to find a test case that will fail if this mutation is in the code.

### How do I select a mutation to work on?

Visit https://github.com/checkstyle/checkstyle/tree/master/.ci/pitest-suppressions, take a look through any of the suppression files there, and pick one mutation to work on. 

### How do I resolve a pitest suppression?
1. Pick a single mutation suppression from a file in https://github.com/checkstyle/checkstyle/tree/master/.ci/pitest-suppressions
2. Create a new issue, with a permalink to the suppression.
3. Send PR with single hardcoded mutation only (ex: change `lcurly = child.getFirstChild();` to `lcurly =lcurly;`) 
4. If any tests or continuous integration testing (ex: `no-exception testing`) are failing, you might have found a test case to kill the mutation. 
5. If a test case is not found in step (4), then generate a check regression report. See https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#report-generation for details. Any differences or exceptions in the report are probably a new test case.
[...]

PR #13015, opened 2023-04-20
Issue #12671: Resolve pitest for FinalLocalVariableCheck

ShivamPandey00:
Related to #12671 #12341 

Mutation killed:
https://github.com/checkstyle/checkstyle/blob/e1e578b472d2730f82b0963173ab462be5e54901/config/pitest-suppressions/pitest-coding-2-suppressions.xml#L12-L19

Diff Regression config: https://gist.githubusercontent.com/coderdeadpool/96fa62c511add1eb980efe8de851256d/raw/57420e4b601c7ed073340dd039e5c4af7a0d1ccd/gistfile1.txt

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
@@ -346,7 +346,6 @@ public class FinalLocalVariableCheck extends AbstractCheck {
                     || findLastCaseGroupWhichContainsSlist(parentAst.getParent()) == parentAst) {
                     containsBreak = scopeStack.peek().containsBreak;
                     scope = scopeStack.pop().scope;
-                    prevScopeUninitializedVariables.pop();
                 }
                 if (containsBreak || shouldUpdateUninitializedVariables(parentAst)) {
                     updateAllUninitializedVariables(prevScopeUninitializedVariableData);
--- a/config/pitest-suppressions/pitest-coding-2-suppressions.xml
+++ b/config/pitest-suppressions/pitest-coding-2-suppressions.xml
@@ -12,15 +12,6 @@
   <mutation unstable="false">
     <sourceFile>FinalLocalVariableCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.FinalLocalVariableCheck</mutatedClass>
-    <mutatedMethod>leaveToken</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to java/util/Deque::pop</description>
-    <lineContent>prevScopeUninitializedVariables.pop();</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>FinalLocalVariableCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.FinalLocalVariableCheck</mutatedClass>
     <mutatedMethod>updateUninitializedVariables</mutatedMethod>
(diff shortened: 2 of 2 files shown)


coderdeadpool:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
```diff
@@ -346,7 +346,6 @@ public void leaveToken(DetailAST ast) {
                     || findLastCaseGroupWhichContainsSlist(parentAst.getParent()) == parentAst) {
                     containsBreak = scopeStack.peek().containsBreak;
                     scope = scopeStack.pop().scope;
-                    prevScopeUninitializedVariables.pop();
```
nrmancuso, line 349: Please provide analysis of why we can’t hit/do not need this code, and why we cannot kill it with a test.

nrmancuso (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
coderdeadpool: @nrmancuso The conditions required to reach the `prevScopeUninitializedVariables.pop();` where the mutation occurs might be difficult to achieve in a test case. Also, The mutation might be indirectly covered by other parts of the code, Like https://github.com/checkstyle/checkstyle/blob/1db88b8d0600f10426e76d9024fb0d9ae3c5c2b9/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java#L461-L470 making it difficult to isolate and kill with a single test case.

romani: @coderdeadpool , can you prove ? Without maybe and difficult

coderdeadpool: I have tried several test cases but none of them are able to kill this mutation without changing the main code.
Instead i have created a test case which can kill the mutation that is created by removing `prevScopeUninitializedVariables.pop();`. So by doing this we can kill this mutation without creating any more mutations.
Also, that test case is killing most of the mutations of pitest-coding-2 profile

romani: we can always explain removal of code by explanation like this.
Challenge here is to analytically prove that it is not required, based on algorithm.

Example: "we do not need pop, because wedo not need scopes top"

coderdeadpool: @romani I am still in the process of understanding the codebase and trying to understand everything. Doing my best to grasp the concepts and structure.

romani: Good, keep digging.

coderdeadpool: @romani Sorry for the delay.
I have a doubt. Since the previous scope's uninitialized variables have already been updated with the call to `updateAllUninitializedVariables(prevScopeUninitializedVariableData)` , it may not be necessary to remove the top of the `prevScopeUninitializedVariables`  stack. Please correct me if am wrong.

coderdeadpool: Okay, so we do need to pop.

romani: we need to find code where previous scope of variable become matter.

Kevin222004:
@romani can we close this please

Issue #13026, opened 2023-04-23
Modify the instructions for checks related to 'record' in WriteTagCheck


Sloent:
I have read check documentation: https://checkstyle.org/config_javadoc.html#WriteTag
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run

\S  - A non-whitespace character: [^\s]

Here are the results of the CI report:

`type config.xml`
```
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
    "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
    "https://checkstyle.org/dtds/configuration_1_2.dtd">

<module name="Checker">
   <module name="TreeWalker">
   <module name="WriteTag">
      <property name="tag" value="@incomplete" />
      <property name="tagFormat" value="\S." />
      <property name="tokens" value="RECORD_DEF" />
      </module>
   </module>
</module>
```
`javac record_test.java`
```
public class record_test {
    /**
     * @incomplete This class needs more code...
     */
    record Person(String name, int age) {
        public Person {
            if (name == null) {
                throw new IllegalArgumentException("Name cannot be null");
            }
            if (age < 0 || age > 150) {
                throw new IllegalArgumentException("Invalid age: " + age);
            }
        }
    }
}
```
`set RUN_LOCALE="-Duser.language=en -Duser.country=US"`
`java %RUN_LOCALE% -jar checkstyle-10.9.3-all.jar -c config.xml record_test.java`
```
Starting audit...
[INFO] ....record_test.java:5: Javadoc tag @incomplete=This class needs more code... [WriteTag]
Audit done.
```

Expected is no violation.

romani:
@Sloent , please report issue as requested in issue template by CLI

Sloent:
I have updated the content of the question

romani:
@Sloent , if you see where is problem - please send PR with fix

nrmancuso:
I am failing to understand the issue here.

I would expect the following two violations, since the `@incomplete` tag is present, and we use `info` severity, but the regexp doesn't match:

```
➜  src cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
    "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <property name="severity" value="error"/>
[...]

PR #13034, opened 2023-04-27
Issue #13026: Modify the instructions for checks related to record in WriteTagCheck

Sloent:
Issue: #13026 
Modify violation messages associated with 'record'.
Update doc:
image


src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/WriteTagCheck.java
```diff
@@ -163,6 +163,10 @@
  * }
  * </pre>
  * <p>
+ * For all the above examples, the handling of 'record' introduced in Java 14 and 'class'
+ * is the same.
```
romani, line 167: this sounds strange ... as it is also applicable to interfaces and all other target tokens.

romani (changes requested):
Root problem is not clear goal of this ancient Check.
and not clear documentation.

Confusion is that this Check demands:
- if no target javadoc tag give violation
- if tag is present, put violation on tag that match regexp.

I think we need to update description of this Check to describe this.


from doc:
>Requires user defined Javadoc tag to be present in Javadoc comment with defined format.

this is bad wording, as it is not exactly what happening.
proposal: `Requires user defined Javadoc tag to be present in Javadoc comment, tag that match format will be marked as violation.`

>  Property tagSeverity is used for severity of events when the tag exists. 

` Property tagSeverity is used for severity of violations when the tag exists. `


-----

And we need extend example to show it.

Sloent:
> - if no target javadoc tag give violation
> - if tag is present, put violation on tag that match regexp.

So I need to update the documentation for this check and add examples to describe the two requirements above?  Then do I need to keep the other modifications I have made so far?

Sloent:
I got it

romani:
Do not hesitate to close this PR and open new with changes I pointed at https://github.com/checkstyle/checkstyle/pull/13034#pullrequestreview-1404044934

Or simply change things PR.

Sloent:
Maybe I stayed up too hard this month and had a fever of 38.5 degrees last night.  In addition, I have to hand in some course work and a paper for the math school competition when I return from the holiday.  Maybe I can't deal with this issue these days.  After I solve other problems, I will open a new PR according to your requirement.

rnveach:
CI needs to be green. Needs to be rebased.

romani:
I lost reason of this update.
if we do extension of doc we kind should do same for all others modules, but take it for granted that user knows that  interface/class/enum/record are kind the same , they are types that can have javadoc.

PR #13073, opened 2023-05-21
Issue #11514: Added Test cases to TreeTableTest.java for TreeTebleMod…

MrPatchz:
Partial of #11514 for improving code coverage for GUI.

Added test cases for TreeTableModelAdapter class in TestTableTest.java:
- Build successful with mvn clean verify
- Added test case for getValueAt method
- Added test case for getColumnClass method
- Added test case for isCellEditable method

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1004,7 +1004,7 @@
                     <limit>
                       <counter>LINE</counter>
                       <value>COVEREDRATIO</value>
-                      <minimum>0.64</minimum>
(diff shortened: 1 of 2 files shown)


romani (changes requested):
update coverage at 
https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L1000-1012

MrPatchz:
@romani Updated pom.xml with the new coverage. Thanks!

Issue #13108, opened 2023-05-30
Resolves Pitest for AbstractSuperCheck


Kevin222004:
Part of https://github.com/checkstyle/checkstyle/issues/13089

This issue will cover all the suppressed mutation for AbstractSuperCheck.java

https://github.com/checkstyle/checkstyle/blob/d60bae5293a4221f4e5b28da1300441a90be5d44/config/pitest-suppressions/pitest-coding-1-suppressions.xml#L3-L10

romani:
@Kevin222004 , lets not create too much tickets.
please copy https://github.com/checkstyle/checkstyle/issues/13089 to new ticket created by you and simply reference only it in all PRs

you can always update ticket to put link to PR for specific mutation, to see that is in flight , what is done and what is still not done.

nrmancuso:
Closed via #13111

PR #13111, opened 2023-05-30
Issue #13108: Resolves Pitest for AbstractSuperCheck

Kevin222004:
Issue #13108: Resolves Pitest for AbstractSuperCheck

covering 
https://github.com/checkstyle/checkstyle/blob/f633d9b61116664b6c00154ad96f6633e0addea2/config/pitest-suppressions/pitest-coding-1-suppressions.xml#L3-L11

### DIFF
--- a/config/pitest-suppressions/pitest-coding-1-suppressions.xml
+++ b/config/pitest-suppressions/pitest-coding-1-suppressions.xml
@@ -1,15 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
   <mutation unstable="false">
-    <sourceFile>AbstractSuperCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.AbstractSuperCheck$MethodNode</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable callingSuper</description>
-    <lineContent>callingSuper = false;</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>DeclarationOrderCheck.java</sourceFile>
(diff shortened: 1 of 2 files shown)


Kevin222004:
Test cases not found.
https://github.com/checkstyle/checkstyle/blob/f633d9b61116664b6c00154ad96f6633e0addea2/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/AbstractSuperCheck.java#L215-L217
In the constructor only the parameter ast is taken for `private final DetailAST method;` every time when the object of `MethodNode` class is created then `private boolean callingSuper;` is been set false by the constructor intially. 

By default in java the value of none initalize boolean variable is false;
https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html#:~:text=boolean-,false,-Local%20variables%20are 
so if it is been removed it not make any changes.

Kevin222004:
Regression :- 
https://kevin222004.github.io/RegressionReport/

https://kevin222004.github.io/Regression-r1/

rdiachenko:
> This might not be a good practice to leave a variable on default initialization and not initialize it by ourselves. There are lots of mutations similar to this.
> 
> @romani @Vyom-Yadav @rdiachenko do you have any suggestions for this type of case?

If it's a primitive type like `boolean` I don't see why we shouldn't rely on default initialization. If you check other checks we do rely on default `false` for boolean properties.

rdiachenko (approved):
looks good to me

nrmancuso (changes requested):
@Kevin222004 different SHA of patch branch in reports and PR is questionable; if we are removing code, regression reports should be generated on final SHA of the commit in your PR.

Kevin222004:
@nrmancuso I am also not sure how this happens, I will provide the new one.

Vyom-Yadav (approved):
LGTM!

Kevin222004:
@nrmancuso 
new report :- https://kevin222004.github.io/reports/ASC1/2023-05-31-T-13-58-32/test-report/index.html

romani (approved):
Ok to merge

PR #13103, opened 2023-05-29
Issue #13097: Split annotation xdocs into separate pages

stoyanK7:
Part of #13097

PR #13117, opened 2023-05-31
minor: disable section validation

stoyanK7:
From https://github.com/checkstyle/checkstyle/pull/13103#issuecomment-1570440101

>  I think test is just strict enforcement of doc to be present. Back in several years, web site was very very very inconsistent. So such hack in Junit helped put situation under control.
> 
> Now layout are changing, we do not expect new Checks in near future, so we disable this check or some part of it and create issue to come back to it at the end , when we will have already final layout.
> 
> It was part of project to setup best practices ( wiki page) and enforce some of them. Please create create two issues to deal with it later at summer.


romani:
CI is not happy

nrmancuso:
@stoyanK7 please ping me once CI is happy

stoyanK7:
@nrmancuso Ping.

nrmancuso:
@stoyanK7 are we able to just use `@Disabled` annotation with comment like `until #13118`?

nrmancuso (approved):
Ok, we have git history of test to base new one on, I am good.

stoyanK7:
> @stoyanK7 are we able to just use `@Disabled` annotation with comment like `until #13118`?

You're right. I went a bit too harsh on those tests by deleting them 😂. Though they will have to be rewritten.

Issue #13095, opened 2023-05-29
Resolves Pitest for VariableDeclarationUsageDistanceCheck


Kevin222004:
Part of https://github.com/checkstyle/checkstyle/issues/13089

This issue will cover all the suppressed mutation for VariableDeclarationUsageDistanceCheck.

https://github.com/checkstyle/checkstyle/blob/6f451fbb49afc21b0e04b46c0576c7a984419a88/config/pitest-suppressions/pitest-coding-1-suppressions.xml#L111-L190

Kevin222004:
I am on it.

romani:
Resolved at scope of https://github.com/checkstyle/checkstyle/issues/13109

Point `6)`

PR #13120, opened 2023-06-01
Issue #13095: kill mutation for VariableDeclarationUsageDistanceCheck 4

Kevin222004:
Issue #13095: kill mutation for VariableDeclarationUsageDistanceCheck 4

# Check
https://checkstyle.org/config_coding.html#VariableDeclarationUsageDistance

# Mutation 
https://github.com/checkstyle/checkstyle/blob/81448ea434cda883b748d3d2c77b2f58a8b22bbe/config/pitest-suppressions/pitest-coding-1-suppressions.xml#L102-L109

4) Explanation:-
My thinking is that instead of removing the condition, we can continue if the token is SEMI and set the result to true. Now this won't have any problem as other branches can alter the result in the subsequent iterations and result is by default true.

A single dangling ; would be EMPTY_STAT and not SEMI. Please run regression on this method.

5)  Regression 

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/2e4b5ad_2023084207/reports/diff/index.html

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/2e4b5ad_2023213552/reports/diff/index.html

----------------------

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/12b3bfed9dcea32830f532159c5023c6/raw/1d2cf31a526f452572fc41bf922762ecc5c6b7d6/variable.xml
Report label: New-Regression-1

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
@@ -462,52 +462,34 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
         DetailAST currentSiblingAst = variableUsageAst;
         String initInstanceName = "";
 
-        while (result
-                && !isUsedVariableDeclarationFound
-                && currentSiblingAst != null) {
-            switch (currentSiblingAst.getType()) {
-                case TokenTypes.EXPR:
-                    final DetailAST methodCallAst = currentSiblingAst.getFirstChild();
-
-                    if (methodCallAst.getType() == TokenTypes.METHOD_CALL) {
-                        final String instanceName =
-                            getInstanceName(methodCallAst);
-                        // method is called without instance
-                        if (instanceName.isEmpty()) {
-                            result = false;
-                        }
-                        // differs from previous instance
-                        else if (!instanceName.equals(initInstanceName)) {
-                            if (initInstanceName.isEmpty()) {
-                                initInstanceName = instanceName;
-                            }
-                            else {
-                                result = false;
-                            }
-                        }
+        while (result && !isUsedVariableDeclarationFound && currentSiblingAst != null) {
+            if (currentSiblingAst.getType() == TokenTypes.EXPR
+                    && currentSiblingAst.getFirstChild().getType() == TokenTypes.METHOD_CALL) {
+                final DetailAST methodCallAst = currentSiblingAst.getFirstChild();
+                final String instanceName = getInstanceName(methodCallAst);
+                if (instanceName.isEmpty()) {
+                    result = false;
+                }
+                else if (!instanceName.equals(initInstanceName)) {
+                    if (initInstanceName.isEmpty()) {
+                        initInstanceName = instanceName;
                     }
                     else {
-                        // is not method call
                         result = false;
                     }
-                    break;
+                }
 
-                case TokenTypes.VARIABLE_DEF:
-                    final String currentVariableName = currentSiblingAst
-                        .findFirstToken(TokenTypes.IDENT).getText();
-                    isUsedVariableDeclarationFound = variableName.equals(currentVariableName);
-                    break;
-
-                case TokenTypes.SEMI:
-                    break;
-
-                default:
-                    result = false;
             }
-
+            else if (currentSiblingAst.getType() == TokenTypes.VARIABLE_DEF) {
+                final String currentVariableName =
+                        currentSiblingAst.findFirstToken(TokenTypes.IDENT).getText();
+                isUsedVariableDeclarationFound = variableName.equals(currentVariableName);
+            }
+            else {
+                result = currentSiblingAst.getType() == TokenTypes.SEMI;
+            }
             currentSiblingAst = currentSiblingAst.getPreviousSibling();
         }
-
         return result;
     }
 
--- a/config/pitest-suppressions/pitest-coding-1-suppressions.xml
+++ b/config/pitest-suppressions/pitest-coding-1-suppressions.xml
@@ -18,13 +18,4 @@
     <lineContent>return Collections.unmodifiableSet(methodCalls);</lineContent>
   </mutation>
 
-  <mutation unstable="false">
-    <sourceFile>VariableDeclarationUsageDistanceCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck</mutatedClass>
-    <mutatedMethod>isInitializationSequence</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
-    <description>removed conditional - replaced equality check with true</description>
-    <lineContent>while (result</lineContent>
-  </mutation>
-
 </suppressedMutations>


Kevin222004:
Not Ready to merge Regression started at https://github.com/Kevin222004/Kevin222004.github.io/actions/runs/5140708944/jobs/9252421769?pr=14

Kevin222004:
Regression :- https://kevin222004.github.io/reports/M5/2023-06-01-T-08-28-54/test-report/index.html

romani:
@Kevin222004 , please remove mutation survival from config 
https://github.com/checkstyle/checkstyle/actions/runs/5140688341/jobs/9252379713?pr=13120#step:7:24

romani:
@Kevin222004 , please remove mutation survival from config 
https://github.com/checkstyle/checkstyle/actions/runs/5140688341/jobs/9252379713?pr=13120#step:7:24

Kevin222004:
@romani ready for review

Kevin222004:
@rdiachenko please review this as well

romani:
> but yes the whole loop can run for more time then it was running with result

this is stable side-effect of pitest demands. Previosly, all times we followed pitest. I am not sure when to stop this fanaticism, but I ok to keep do code change in favor of pitest.

romani (approved):
this code looks so familiar ..... 
@Vyom-Yadav , did we merge your update for this method ?

rdiachenko:
> > but yes the whole loop can run for more time then it was running with result
> 
> this is stable side-effect of pitest demands. Previosly, all times we followed pitest. I am not sure when to stop this fanaticism, but I ok to keep do code change in favor of pitest.

@romani even though, it is a stable side-effect, we are lucky there are no heavy operations in the loop. I'm more for not following pitest in such cases and put this case to a list of petentially bad side effects.

@Kevin222004 could you resolve a conflict and regenerate report please?

Kevin222004:
> @Kevin222004 could you resolve a conflict and regenerate report please?

started

Kevin222004:
Regression 1 :- https://kevin222004.github.io/reports/M5/2023-06-03-T-07-27-19/test-report/index.html

romani:
@Kevin222004 , please create issue on side effects, to let you be able to modify it as we go.

@rdiachenko , for now lets keep moving with following pitest, in this particular case it is nano optimization, so better to have NOT pitest survival, it will help later on in maintenance and defending no survival from other contributors.
Unfortunately presence of any let contributors possibility to appeal to do same, and hard time for maintainers to agree or push back.

Kevin222004:
Regression :- 2 https://kevin222004.github.io/reports/M5/2023-06-03-T-13-23-06/test-report/index.html

rdiachenko (approved):
lgtm

Vyom-Yadav:
>  @Vyom-Yadav , did we merge your update for this method ?

@romani No, would you like to merge it first?

romani:
@Vyom-Yadav , your update/PR still not ready, there items to resolve.

Vyom-Yadav:
@Kevin222004 Instead of removing the condition try:
```java
while (result                                                                         
        && !isUsedVariableDeclarationFound                                            
        && currentSiblingAst != null) {                                               
    switch (currentSiblingAst.getType()) {                                            
        case TokenTypes.EXPR:                                                         
            final DetailAST methodCallAst = currentSiblingAst.getFirstChild();        
                                                                                      
            if (methodCallAst.getType() == TokenTypes.METHOD_CALL) {                  
                final String instanceName =                                           
                    getInstanceName(methodCallAst);                                   
                // method is called without instance                                  
                if (instanceName.isEmpty()) {                                         
                    result = false;                                                   
                }                                                                     
                // differs from previous instance                                     
                else if (!instanceName.equals(initInstanceName)) {                    
                    if (initInstanceName.isEmpty()) {                                 
                        initInstanceName = instanceName;                              
                    }                                                                 
                    else {                                                            
                        result = false;                                               
                    }                                                                 
                }                                                                     
            }                                                                         
            else {                                                                    
                // is not method call                                                 
                result = false;                                                       
            }                                                                         
            break;                                                                    
                                                                                      
        case TokenTypes.VARIABLE_DEF:                                                 
            final String currentVariableName = currentSiblingAst                      
                .findFirstToken(TokenTypes.IDENT).getText();                          
            isUsedVariableDeclarationFound = variableName.equals(currentVariableName);
            break;                                                                    
                                                                                      
        default:                                                                      
            result = currentSiblingAst.getType() == TokenTypes.SEMI;                  
    }                                                                                 
                                                                                      
    currentSiblingAst = currentSiblingAst.getPreviousSibling();                       
}                                                                                     
```

My thinking is that instead of removing the condition, we can continue if the token is `SEMI` and set the result to `true`. Now this won't have any problem as other branches can alter the `result` in the subsequent iterations and `result` is by default `true`.

A single dangling `;` would be `EMPTY_STAT` and not `SEMI`. Please run regression on this method.

Vyom-Yadav:
We would be able to kill the mutation this way and have the condition to save time too.

Kevin222004:
Github, generate report

Vyom-Yadav (changes requested):
https://github.com/checkstyle/checkstyle/pull/13120#issuecomment-1579916446

Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

Kevin222004:
@Vyom-Yadav Reports are clean

Vyom-Yadav:
@Kevin222004 Please put reports in the PR description. Finding final reports from comments can be tricky. PR description should be considered the source of truth for all changes. Please meticulously do it in all the PRs you create.

Vyom-Yadav (approved):
LGTM!

Vyom-Yadav:
@rdiachenko @romani Please review again, the complete logic was changed after your review.

rdiachenko:
@Kevin222004 please resolve conflicts

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -498,11 +498,8 @@ else if (!instanceName.equals(initInstanceName)) {
                     isUsedVariableDeclarationFound = variableName.equals(currentVariableName);
                     break;
 
-                case TokenTypes.SEMI:
-                    break;
-
                 default:
-                    result = false;
+                    result = currentSiblingAst.getType() == TokenTypes.SEMI;
```
rdiachenko, line 502: This refactoring makes code less readable, creates questions when you read it and error-prone to extend the logic in the main loop if there is such a need. I'd vote for not following pitest here.

@romani thoughts?

Vyom-Yadav: > error-prone to extend the logic in the main loop if there is such a need.

Extending any logic in this check is quite difficult. This loop is just for detecting if the statements result in an initialization sequence or not. Have you seen https://github.com/checkstyle/checkstyle/pull/13120#issuecomment-1579916446?

We can either not follow pitest here or go with the current solution. In any case, I don't think we should get rid of `result` as a loop-controlling variable.

romani:
@Kevin222004 , please resolve conflict meanwhile

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
rdiachenko: > Have you seen https://github.com/checkstyle/checkstyle/pull/13120#issuecomment-1579916446?

Yes, would you remember this when you open this code in few months?

Among 3 options we have - not follow pitest, remove `result`, move case logic to `default` - I wouldn't follow pitest here, you sacrifice readability and simplicity over making a tool happy.

Vyom-Yadav: I am ok with not following pitest here:
1. It decreases performance by removing `result` (only on paper).
2. It reduces readability.

Kevin222004: should we go with removing the `result`?

romani: Let me review this.
I need to meditate a bit on this.

Kevin222004: @romani ping

romani: after meditation on this.
General code of this method is not good, tools by all means point us to this code to rewrite it.
We do not have time to rewrite code to better (it will be even outside of this method).
If rewriting of this code is inevitable, lets not have pitest suppressions as area for next engineer to extend it more.
So in next iteration over such code we will be constrained by tools from the beginning and it will force us to write better logic.

romani (approved):
ok to merge.

Reason: https://github.com/checkstyle/checkstyle/pull/13120/files#r1234129636

romani:
@Kevin222004 , please resolve IDEA inspection.

Kevin222004:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
rdiachenko: So, we make bad code to be worse to enforce us to make it better when we rewrite it :) The problem with this approach is that most of the code will never be rewritten and making it worse by following tools will eventually lead to codebase which is hard to read, understand and maintain.

I'm ok to merge this specific case, but there should be some boundaries.

Kevin222004:
Github, generate report

Kevin222004:
@romani Please check failure are not related to change

Kevin222004:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
rdiachenko: @Kevin222004 why did you rewrite the whole logic to if conditions? We agreed to go with `result = currentSiblingAst.getType() == TokenTypes.SEMI;` in the default statement.

Kevin222004: @rdiachenko due to ci failure.
I am sorry, I am not remembering an exact message.

rdiachenko:
@Kevin222004 please update repo and make sure all checks are green.

romani:
Rebased in GitHub

rdiachenko (approved):
lgtm

PR #13121, opened 2023-06-01
Issue #13095: kill mutation for VariableDeclarationUsageDistanceCheck 5

Kevin222004:
1 Issue https://github.com/checkstyle/checkstyle/issues/13095: kill mutation for VariableDeclarationUsageDistanceCheck 5

------------

# Check:-
https://checkstyle.org/config_coding.html#VariableDeclarationUsageDistance

-----------

# mutation covered
https://github.com/checkstyle/checkstyle/blob/6f451fbb49afc21b0e04b46c0576c7a984419a88/config/pitest-suppressions/pitest-coding-1-suppressions.xml#L111-L118

---------

# Explaination
ok so the whole loop will start if 
https://github.com/checkstyle/checkstyle/blob/c5566a6ea8c336ddedde446648b03ef56a73b7ba/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L613
currentScopeAst != null;

then 
https://github.com/checkstyle/checkstyle/blob/c5566a6ea8c336ddedde446648b03ef56a73b7ba/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L614-L619

here `searchResult` will find the first usage of variable
know lets take a example 
```
    public void testMethod9() {
        boolean result = false;
        boolean b1 = true;
        if (b1) {
            result = true;
        }
    }
```

ast for if block in method 
```
    |       |--LITERAL_IF -> if [19:8]
    |       |   |--LPAREN -> ( [19:11]
    |       |   |--EXPR -> EXPR [19:12]
    |       |   |   `--IDENT -> b1 [19:12]
    |       |   |--RPAREN -> ) [19:14]
    |       |   `--SLIST -> { [19:16]
    |       |       |--EXPR -> EXPR [20:19]
    |       |       |   `--ASSIGN -> = [20:19]
    |       |       |       |--IDENT -> result [20:12]
    |       |       |       `--LITERAL_TRUE -> true [20:21]
    |       |       |--SEMI -> ; [20:25]
    |       |       `--RCURLY -> } [21:8]
    |       `--RCURLY -> } [22:4]

```

lets suppose we have variable `result` know what does `searchResult` give the result as `LITERAL_IF -> if [19:8]` where it is used.

and variable usage expression will also get same values.

in this example size of list of variableUsageExpressions would be one. so if statment will execute. and over their `blockWithVariableUsage = variableUsageExpressions.get(o)` means in our example again it would be assigned as `LITERAL_IF -> if [19:8]`

in switch case 
https://github.com/checkstyle/checkstyle/blob/c5566a6ea8c336ddedde446648b03ef56a73b7ba/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L639-L641

`getFirstNodeInsideIfBlock` will execute this method basically provide 
https://github.com/checkstyle/checkstyle/blob/c5566a6ea8c336ddedde446648b03ef56a73b7ba/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L743-L745

`SLIST -> { [19:16]` in all the case if the variable is used in initalization then it will return `LITERAL_IF -> if [19:8]` this is same for all other case except default one.

and then here lets suppose if variable is used in 
initalization then 
https://github.com/checkstyle/checkstyle/blob/c5566a6ea8c336ddedde446648b03ef56a73b7ba/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L655-L657

currentScopeAst will become null and variable usage ast become blockWithVariableUsage hence in this case their is no issue.

but lets say if it is not null in the above example. then currentScope would become exprWithVariableUsage
and variableUsageAst become exprWithVariableUsage.

and the loop start again with the value of exprWithVariableUsage

and again `searchResult` will find the usage but know the scope will start from exprWithVariableUsage and again `blockWithVariableUsage` will get it value and here `variableUsageAst` is assgined as `blockWithVariableUsage`. 
know loop will run till currentScopeASt would not become null and if it is null then value of variableUsageAst will become `blockWithVariableUsage`
 so at the end the value will become only `blockWithVariableUsage` so istead of assigning it to
 `currentScopeAst` it will not make any issue if we assign it as blockWithVariableUsage 

-------------------

# Regression :- 
https://kevin222004.github.io/reports/M6/2023-06-01-T-14-00-21/test-report/index.html

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
@@ -23,7 +23,6 @@ import java.util.AbstractMap.SimpleEntry;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map.Entry;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -653,8 +652,7 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
                         exprWithVariableUsage = blockWithVariableUsage.getFirstChild();
                 }
                 currentScopeAst = exprWithVariableUsage;
-                variableUsageAst =
-                        Objects.requireNonNullElse(exprWithVariableUsage, blockWithVariableUsage);
+                variableUsageAst = blockWithVariableUsage;
             }
 
             // If there's no any variable usage, then distance = 0.
--- a/config/pitest-suppressions/pitest-coding-1-suppressions.xml
+++ b/config/pitest-suppressions/pitest-coding-1-suppressions.xml
@@ -93,15 +93,6 @@
   <mutation unstable="false">
     <sourceFile>VariableDeclarationUsageDistanceCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck</mutatedClass>
-    <mutatedMethod>calculateDistanceBetweenScopes</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator</mutator>
-    <description>replaced call to java/util/Objects::requireNonNullElse with argument</description>
-    <lineContent>Objects.requireNonNullElse(exprWithVariableUsage, blockWithVariableUsage);</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>VariableDeclarationUsageDistanceCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck</mutatedClass>
     <mutatedMethod>isInitializationSequence</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
     <description>removed conditional - replaced equality check with true</description>


Kevin222004:
GitHub, generate report

Kevin222004:
@nrmancuso @romani @Vyom-Yadav @rdiachenko ready for review
generating new report started

Kevin222004:
Regression after code changes :- https://kevin222004.github.io/reports/M6/2023-06-01-T-14-00-21/test-report/index.html

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -652,9 +652,11 @@ private static Entry<DetailAST, Integer> calculateDistanceBetweenScopes(
                     default:
                         exprWithVariableUsage = blockWithVariableUsage.getFirstChild();
                 }
-                currentScopeAst = exprWithVariableUsage;
                 variableUsageAst =
                         Objects.requireNonNullElse(exprWithVariableUsage, blockWithVariableUsage);
+                if (exprWithVariableUsage != null) {
+                    currentScopeAst = variableUsageAst;
```
rdiachenko, line 658: The logic is hard to read / understand. Reading this line you need to keep in mind both the `if` condition and the condition of `Objects.requireNonNullElse`.

It is also strange to me that the mutated code hasn't been touched at all. You just changed unrelated code. Have you tried to hardcode mutation? What was the result?

Can you try to replace a call to `Objects.requireNonNullElse` and your new condition with the following?
```
                if (exprWithVariableUsage == null) {
                    variableUsageAst = blockWithVariableUsage;
                } else {
                    variableUsageAst = exprWithVariableUsage;
                    currentScopeAst = variableUsageAst;
                }
```

Kevin222004: > tried to hardcode mutation? What was the result?
Yes the report is clean https://github.com/checkstyle/checkstyle/pull/13121#issuecomment-1571651427

> Can you try to replace a call to Objects.requireNonNullElse and your new condition with the following?
@rdiachenko In the first try I have done the same as you but i have not include `currentScopeAst = variableUsageAst;`
in else thats why mutation has not killed at that time and I have make this type of changes.

I have updated the pr with the changes you suggest it will kill the mutation

Kevin222004: done

rdiachenko (approved):
lgtm

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -652,9 +651,13 @@ private static Entry<DetailAST, Integer> calculateDistanceBetweenScopes(
...
-                currentScopeAst = exprWithVariableUsage;
-                variableUsageAst =
-                        Objects.requireNonNullElse(exprWithVariableUsage, blockWithVariableUsage);
+                if (exprWithVariableUsage == null) {
+                    variableUsageAst = blockWithVariableUsage;
+                }
+                else {
+                    variableUsageAst = exprWithVariableUsage;
+                    currentScopeAst = variableUsageAst;
+                }
```
Vyom-Yadav, line 660: We are not actually killing the mutation. We are just playing smart with pitest so it cannot create a mutation that passes all test cases.

The real problem is assigning `variableUsageAst` to `blockWithVariableUsage` in every circumstance i.e. `variableUsageAst = blockWithVariableUsage;`.

---

The condition:
```java
 variableUsageAst = Objects.requireNonNullElse(exprWithVariableUsage, blockWithVariableUsage);
```
translates to:
```java
if (exprWithVariableUsage == null) {          
    variableUsageAst = blockWithVariableUsage;
} else {                                      
    variableUsageAst = exprWithVariableUsage; 
}                                             
```
Now why this won't pass pitest is due to the mutation `exprWithVariableUsage == null` -> `true` which again results in the initial mutation condition i.e.  `variableUsageAst = blockWithVariableUsage;`

---

Now we take advantage of another fact that `currentScopeAst == null` in every iteration and it is assigned value only at `currentScopeAst = exprWithVariableUsage;` Now there is no need to assign a value to `currentScopeAst` when `exprWithVariableUsage` is `null` because we already set `currentScopeAst == null`. 

So using this information to our advantage we put `currentScopeAst = exprWithVariableUsage;` in the second branch of if-else so that the mutation `exprWithVariableUsage == null` -> `true` fails a test case as in that condition `currentScopeAst` would not be assigned to `exprWithVariableUsage`.

It is basically:
```java
currentScopeAst = exprWithVariableUsage;        
if (exprWithVariableUsage == null) {            
    variableUsageAst = blockWithVariableUsage;  
} else {                                        
    variableUsageAst = exprWithVariableUsage;   
}                                               
```

versus 

```java
if (exprWithVariableUsage == null) {          
    variableUsageAst = blockWithVariableUsage;
} else {                                      
    // try replacing exprWithVariableUsage with blockWithVariableUsage,
    // still all test cases would pass 
    variableUsageAst = exprWithVariableUsage;
    currentScopeAst = exprWithVariableUsage;  
}                                             
```

BUT this doesn't change the fact that `variableUsageAst = blockWithVariableUsage;`  can be done in **any** condition. Hence, the mutation is not actually killed but we just used certain facts to our advantage to kill the mutation.

This would reduce the count but we won't get any actual benefit in code quality by using this method.

@nrmancuso @romani @rdiachenko Thoughts on this way? I would say we should tackle the original mutation `variableUsageAst = blockWithVariableUsage;` instead of using certain facts to our advantage.

romani: we do not try to kill survival. Our goal to avoid survival. Highly preferable by reasonable code.
It happens all the time with all quality tools.

@Kevin222004 , why we can not simply apply pitest suggestion ?

Kevin222004: @romani we can apply Pitest's suggestion I have debugged the whole code with that as well at the end output of the method will not affect it.

I don't want to make any new changes here and this was the possible way to avoid the mutation. I have really not think as https://github.com/checkstyle/checkstyle/pull/13121#discussion_r1213914694

romani: Instead of your changes, let's do what pitest suggests.

Kevin222004: done

romani (approved):
ok to merge

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
Vyom-Yadav: > we can apply Pitest's suggestion I have debugged the whole code with that as well at the end output of the method will not affect it.

@Kevin222004 Another important aspect of this project is explaining why you think it will work. Sometimes there might be gaps in your calculation or sometimes we are thinking of something else and your calculation can correct us.

So please share why you think this won't be a problem in the PR description. This way even if it comes up as a regression later, we can find faults in our thinking if we have something (like a slight explanation as to why it is correct) present.

Vyom-Yadav (changes requested):
Explanation is pending:
https://github.com/checkstyle/checkstyle/pull/13121#discussion_r1216124199

Kevin222004:
> Explanation is pending: [#13121 (comment)](https://github.com/checkstyle/checkstyle/pull/13121#discussion_r121612419

sorry for delay will do it by today

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
Kevin222004: @Vyom-Yadav I have tried to explain such whole thing if you have any question in that let me know

Vyom-Yadav (approved):
LGTM!

Issue #13109, opened 2023-05-30
Resolve pitest survivals in modules


Kevin222004:
Part of https://github.com/checkstyle/checkstyle/issues/13089

-----------

**1) FullIdent**
- [x] NonVoidMethodCallMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13178  
- [x] NakedReceiverMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13178

------------------

**2) FileContents**
- [x] ArgumentPropagationMutator (2)
        https://github.com/checkstyle/checkstyle/pull/13156 (Covered both)
        
-----------------

**3) FileText**
- [x] MemberVariableMutator (2)   
        https://github.com/checkstyle/checkstyle/pull/13157
        https://github.com/checkstyle/checkstyle/pull/13158
          
-----------------

**4) AbstractSuperCheck$MethodNode**
- [x] MemberVariableMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13111

------------------

**5) DeclarationOrderCheck**
- [x] MemberVariableMutator (2)
        https://github.com/checkstyle/checkstyle/pull/13104
        https://github.com/checkstyle/checkstyle/pull/13123

------------------

**6) VariableDeclarationUsageDistanceCheck**
- [x] RemoveSwitchMutator_1 (1)
        https://github.com/checkstyle/checkstyle/pull/13099
- [x] ArgumentPropagationMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13121
- [x] NonVoidMethodCallMutator (2)
        https://github.com/checkstyle/checkstyle/pull/13098
        https://github.com/checkstyle/checkstyle/pull/13098 
- [x] RemoveSwitchMutator_4 (1)
        https://github.com/checkstyle/checkstyle/pull/13096
- [x] NakedReceiverMutator (2)
        https://github.com/checkstyle/checkstyle/pull/13098
        https://github.com/checkstyle/checkstyle/pull/13122
- [x] RemoveSwitchMutator_6 (1)
        https://github.com/checkstyle/checkstyle/pull/13096

---------------------------

**7) EqualsAvoidNullCheck$FieldFrame**
- [x] ArgumentPropagationMutator (2)
       https://github.com/checkstyle/checkstyle/pull/13124
       https://github.com/checkstyle/checkstyle/pull/13125
       
------------------------

**8) PackageDeclarationCheck**
- [x] MemberVariableMutator (1)
[...]

PR #13133, opened 2023-06-01
Issue #13109: Kill mutation for PackageDeclarationCheck

Kevin222004:
Issue #13109: Kill mutation for PackageDeclarationCheck

-------------

# Mutation
https://github.com/checkstyle/checkstyle/blob/81448ea434cda883b748d3d2c77b2f58a8b22bbe/config/pitest-suppressions/pitest-coding-1-suppressions.xml#L48-L55

### DIFF
--- a/config/pitest-suppressions/pitest-coding-1-suppressions.xml
+++ b/config/pitest-suppressions/pitest-coding-1-suppressions.xml
@@ -19,15 +19,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>PackageDeclarationCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.PackageDeclarationCheck</mutatedClass>
-    <mutatedMethod>beginTree</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable defined</description>
-    <lineContent>defined = false;</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>VariableDeclarationUsageDistanceCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck</mutatedClass>
     <mutatedMethod>isInitializationSequence</mutatedMethod>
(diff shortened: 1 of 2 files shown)


romani:
why draft ? no regression report is executed.

we already discussed somewhere that all reset in beginTree are for case when same instance of Check is used on two files, Check is state-ful, and we should not affect new files by values in fields from previous java file.

Kevin222004:
@romani 
Regression https://kevin222004.github.io/reports/p1/2023-06-02-T-08-07-37/test-report/index.html

romani:
How long it takes for your local to generate such report ?
Do you able to do something else meanwhile ?

Kevin222004:
@romani i am not generating report on local 
I generate using https://github.com/Kevin222004/Kevin222004.github.io

Kevin222004:
@romani https://kevin222004.github.io/Regression/

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/PackageDeclarationCheck.java
```diff
@@ -145,11 +145,6 @@ public int[] getAcceptableTokens() {
         return getRequiredTokens();
     }
 
-    @Override
-    public void beginTree(DetailAST ast) {
-        defined = false;
```
romani, line 150: reminder, all initialization in this method, is reset of Check state before starting new file, to reproduce problem we need to run same Check instance on 2 special input files to let "defined = true" affect validation in second file even it has no java code that make "defined = true".

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/PackageDeclarationCheck.java
romani: Do not deal with pure Junit method , we should cover it by Inputs.
New or existing verify method that takes creates check ones and execute two files sequentially.

Kevin222004: @romani, I have tried to run 2 files sequentially but the only thing changed here is if I make https://github.com/checkstyle/checkstyle/blob/24a4b06182f9b9066a684715f324d2bbf7099d10/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/PackageDeclarationCheck.java#L148-L151 
defined = true then my test case will fail but it is still not resolving our actual mutation which is to remove that line 
`defined = false`

Kevin222004: Done

romani (approved):
Ok to merge.
Just Input based test extension, best scenario.

I move it further for review by others to learn on this.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/packagedeclaration/InputPackageDeclarationCheck.java
```diff
@@ -0,0 +1,7 @@
+package com.puppycrawl.tools.checkstyle.checks.coding.packagedeclaration;
```
rdiachenko, line 1: So, why do we need this new input file if it is not used in the new test?

rdiachenko (approved):
lgtm

Vyom-Yadav (approved):
LGTM!

PR #13151, opened 2023-06-03
Issue #13328: Kill mutation for UnusedLocalVariableCheck 8

Kevin222004:
Issue #13328: Kill mutation for UnusedLocalVariableCheck 8

------

# Check :- 
https://checkstyle.org/checks/coding/unusedlocalvariable.html#UnusedLocalVariable

--------

# Mutation
https://github.com/checkstyle/checkstyle/blob/554a87ea9b5ae27f9519e8dbca15a116bcc108a4/config/pitest-suppressions/pitest-coding-2-suppressions.xml#L75-L82

-----------

# Explaination

the updated scope is changes as only literalNewAst Actually it is not affecting any thing if we make it as literalNewast or objBlock as its last child

----------

# Regression :- 

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/0dbdfb487f8e9a050ed0e6356f1a35b4/raw/31f7927f400e11e4285e86fd086b373095227848/unusedLocal.xml
Report label: Final-Regression-1


Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
```diff
@@ -1016,7 +1016,7 @@ public DetailAST getTypeDeclAst() {
          * @return copy of variables in instanceAndClassVar stack with updated scope.
          */
         public Deque<VariableDesc> getUpdatedCopyOfVarStack(DetailAST literalNewAst) {
-            final DetailAST updatedScope = literalNewAst.getLastChild();
+            final DetailAST updatedScope = literalNewAst;
```
romani, line 1019: Remove `updatedScope` at all.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
Kevin222004: Done

romani (approved):
Ok to merge if reports are clean

Kevin222004:
Please ignore survivng mutation it will be covering at https://github.com/checkstyle/checkstyle/pull/13331

rdiachenko (approved):
lgtm

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
```diff
@@ -1016,11 +1016,10 @@ public DetailAST getTypeDeclAst() {
          * @return copy of variables in instanceAndClassVar stack with updated scope.
          */
         public Deque<VariableDesc> getUpdatedCopyOfVarStack(DetailAST literalNewAst) {
-            final DetailAST updatedScope = literalNewAst.getLastChild();
             final Deque<VariableDesc> instAndClassVarDeque = new ArrayDeque<>();
             instanceAndClassVarStack.forEach(instVar -> {
                 final VariableDesc variableDesc = new VariableDesc(instVar.getName(),
-                        instVar.getTypeAst(), updatedScope);
+                        instVar.getTypeAst(), literalNewAst);
```
Vyom-Yadav, line 1022: I am not ok with pitest complaining here. The logic is correct but it breaks the integrity of the check. We use last-child because we have a set definition of scope.

https://github.com/checkstyle/checkstyle/blob/3142055d05641087c2b4b1775a39548ef02f2f41/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L152-L159

The check by design uses `{` or `for` for determining scope. Extending this to `literalNew` is error-prone as it serves no value. The check is designed to have only the above-mentioned token as scope and this change can make extending this check difficult because one more case has to be handled. 

I vote for suppressing this pitest mutation. 

@romani @rdiachenko Please share your thoughts.

Vyom-Yadav (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
romani: If this value does not matter in current context, let's put null there or use some ctor without this value.

Vyom-Yadav: The value of `scope` matters, it is used to pop out the variables at the right time (from the proper scope). I am saying this should stay: 
```java
final DetailAST updatedScope = literalNewAst.getLastChild();
```

And pitest should be suppressed in this case. This is an intentional check design otherwise every other parent-level token can be considered as `scope` but we have limited it due to intentional design and ease of maintenance.

romani:
@Kevin222004 , please rebase to resolve conflict.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
romani: but you might mean that we do validation based on scope token type
https://github.com/checkstyle/checkstyle/blob/d4b1aec93f4c9e3d6b8eeed8556187e342c1bb33/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L310-L312
so any not SCOPES will be skiped here 
https://github.com/checkstyle/checkstyle/blob/d4b1aec93f4c9e3d6b8eeed8556187e342c1bb33/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L423-L424

Vyom-Yadav: Pitest is logically correct, but it does not know the convention of this check. We only allow following scopes: 

https://github.com/checkstyle/checkstyle/blob/3142055d05641087c2b4b1775a39548ef02f2f41/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L152-L159

> but this method also use parent, not a last child:

Of course, for every type of ast, the relation would be different. Some would have

https://github.com/checkstyle/checkstyle/blob/3142055d05641087c2b4b1775a39548ef02f2f41/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L152-L159

as parent and other would have them as child.

Here `SLIST` is parent of variable def: (`SLIST`)
```
--SLIST -> { [26:47]
   |--VARIABLE_DEF -> VARIABLE_DEF [27:8]
   |   |--MODIFIERS -> MODIFIERS [27:8]
   |   |--TYPE -> TYPE [27:8]
   |   |   `--LITERAL_INT -> int [27:8]
   |   |--IDENT -> sameName [27:12]
   |   `--ASSIGN -> = [27:21]
   |       `--EXPR -> EXPR [27:23]
   |           `--NUM_INT -> 10 [27:23]
```

Here it is the child of `LITERAL_NEW`: (`OBJBLOCK`)
```
--LITERAL_NEW -> new [17:19]
   |--IDENT -> Test [17:23]
   |--LPAREN -> ( [17:27]
   |--ELIST -> ELIST [17:28]
   |--RPAREN -> ) [17:28]
   `--OBJBLOCK -> OBJBLOCK [17:30]
```

romani: main point here. Check already in state without convention, as one method use lastChild and second method getParent.
So now we align them to be all getParent.

So we need to find a case where OBJBLOCK become matter for LITERAL_NEW, so it will kill mutation by extra test case.

Strange that regression did not find it, but it might be ok, unused variables are not very common in code.
We have two choices: 
- follow mutation and experience possible defect later on
- go deeper in recreation of example that can trigger this.

@Kevin222004 , if we do not care on scope here, can we put null in it ?

romani: according to https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_NEW
we just need a bit more tests for anonymous classes.
Usually anonymous classes are not very big, so causing unused variable is very unlikely to find by regression.

@Kevin222004, please reverse engineer java code where becomes to matter.

Vyom-Yadav: @romani Both codes are logically equivalent, and no test case would be there, IMO. 

> So now we align them to be all getParent.

That does not make any sense. The relationship does not matter, the token matters. We restrict it to `SLIST`, `LITERAL_FOR`, `OBJBLOCK`. I see no point adding `LITERAL_NEW` to it when **it can be** covered by `OBJBLOCK`.

```
--LITERAL_NEW -> new [17:19]
   |--IDENT -> Test [17:23]
   |--LPAREN -> ( [17:27]
   |--ELIST -> ELIST [17:28]
   |--RPAREN -> ) [17:28]
   `--OBJBLOCK -> OBJBLOCK [17:30]
```

We use `scope` over here (while leaving a token):
https://github.com/checkstyle/checkstyle/blob/86b95ff9f34805da669c0d99325abef06d04ac51/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L423-L432

So it does not matter if we do checking at `OBJBLOCK` or `LITERAL_NEW`, the result would be the same.

Why complicate the check more by adding a new scope when the current definition of scope makes everything work.

Kevin222004: @romani 

> @Kevin222004 , if we do not care on scope here, can we put null in it ?

No, we can't remove the scope, it matters here

https://github.com/checkstyle/checkstyle/pull/13151#discussion_r1252103046
No test was found. I tried some various way but it not matter wether it is `OBJBLOCK` or `LITERAL_NEW`

Kevin222004: Once again after evaluating and following pitest of pr https://github.com/checkstyle/checkstyle/pull/13152,(following pitest) it looks to me that the design needs to be changed or the logic looks very clean to me so let it be suppressed

romani: I lost track of this PR, I will do deep dive one more time to make decision.

rdiachenko: @Vyom-Yadav 
> Why complicate the check more by adding a new scope when the current definition of scope makes everything work.

We are not extending `SCOPES` with a new scope.

> So it does not matter if we do checking at OBJBLOCK or LITERAL_NEW, the result would be the same.

I didn't get it, why the result would be the same? If it's the same and we are not extending `SCOPES` and there is no input test, why shouldn't we accept Kevin's change?

Vyom-Yadav: > We are not extending SCOPES with a new scope.

We are not adding anything to `SCOPES` static field but now a variable has `LITERAL_NEW` as scope which is not what the check promises. Sort of violates the contract.

> I didn't get it, why the result would be the same? If it's the same and we are not extending SCOPES and there is no input test, why shouldn't we accept Kevin's change?

The result would be the same because no instead of both `OBJBLOCK` and `LITERAL_NEW` don't contribute anything to the violations (as they are non-local). On top of that, it won't affect the stack order because they are popped off at the right time. (With this change, at `LITERAL_NEW` instead of `OBJBLOCK`)

rdiachenko: I tried to play with code preserving `OBJBLOCK` and resolving the mutation - https://github.com/checkstyle/checkstyle/pull/13546/files. 

@Vyom-Yadav what do you think about such a change, assuming it passes all needed ci checks?

romani:
@Kevin222004 , conflict again

Kevin222004:
https://github.com/checkstyle/checkstyle/pull/13151#issuecomment-1685585916 Resolved

rdiachenko (approved):
lgtm

I tried a different approach here https://github.com/checkstyle/checkstyle/pull/13546#issuecomment-1781890727 but it is less safe and may break existing logic. I'd rather go with the current approach with is simpler and safer.

rdiachenko:
@Kevin222004 please resolve conflicts

romani (approved):
I am ok to merge.

Absence of any mutation will not next contributor to extend list of situations.

Vyom-Yadav (approved):
Though this is not what I wanted, but I can see why this is the better solution.

While designing this check, I didn't keep local classes in mind ('cause colliding variable names in local classes is really rare, and then their nesting, it can get reallyyy complex).

Pitest is indirectly indicating that only, this breaks the contract of the check but the logic is working fine. I will come back to this check sometime with hopefully a better design.

---

LGTM!

romani:
By merging this we close a door to all other leaks for survival, one day we will see how to make design better, we already spent too much time on this, we need a break.

PR #13156, opened 2023-06-04
Issue #13109: Kill mutation for FileContents

Kevin222004:
Issue #13109: Kill mutation for FileContents

---------------------

# Mutation covered 

https://github.com/checkstyle/checkstyle/blob/82b6ac46193fafdfbcfc5dacbb848004c36a82e4/config/pitest-suppressions/pitest-api-suppressions.xml#L3-L19

--------------------

# Explaination 
Test cases expanded.
If we see `return Collections.unmodifiableMap(cppComments);` is mutated to `return cppComments;`. basically  Collections. unmodifiable will not allow us to make any updates to the map. if we try to update it it will throw `UnsupportedOperationException`. so in this test case if we try to remove data from the comment so it will throw the exception

### DIFF
--- a/config/pitest-suppressions/pitest-api-suppressions.xml
+++ b/config/pitest-suppressions/pitest-api-suppressions.xml
@@ -1,24 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
   <mutation unstable="false">
-    <sourceFile>FileContents.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.api.FileContents</mutatedClass>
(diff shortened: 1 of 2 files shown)


Kevin222004:
Ready for review

romani (approved):
again `replaced call to java/util/Collections::unmodifiableMap with argument`

but
interesting case.
it is API, so we are ok to merge by pure unit test it as it is api, and we do not know how it will be used so we need to test api method.

rdiachenko (approved):
lgtm

Vyom-Yadav (approved):
LGTM!

PR #13158, opened 2023-06-04
Issue #13109: Kill mutation for FileText 2

Kevin222004:
Issue #13109: Kill mutation for FileText 2

--------------

# Mutation covered 
https://github.com/checkstyle/checkstyle/blob/82b6ac46193fafdfbcfc5dacbb848004c36a82e4/config/pitest-suppressions/pitest-api-suppressions.xml#L75-L82

--------------

# Explaination
Our code is
```
    public FileText(FileText fileText) {
        file = fileText.file;
        charset = fileText.charset;
        fullText = fileText.fullText;
        lines = fileText.lines.clone();
        if (fileText.lineBreaks == null) {
            lineBreaks = null;
        }
        else {
            lineBreaks = fileText.lineBreaks.clone();
        }
    }

```

Here the part of code  
```
        if (fileText.lineBreaks == null) {
            lineBreaks = null;
        }
```

is mutated as 
```
        if (fileText.lineBreaks == null) {
        }
```

Here the condition in if statment is 
something like 
```
fileText.lineBreaks == null
```

know if we see logically their is 2 possiblity 
either `fileText.lineBreaks` is `null` or it is
`not null`  if it is null then the if block will
execute and in the case of `not null` else will execute.

know instead of doing this `fileText.lineBreaks == null` in if block and set the lineBreaks to null 
 we can change the if condition to 
```
        if (fileText.lineBreaks != null) {
            lineBreaks = fileText.lineBreaks.clone();
        }
```
so the new code look like 
```
    public FileText(FileText fileText) {
        file = fileText.file;
        charset = fileText.charset;
        fullText = fileText.fullText;
        lines = fileText.lines.clone();
        if (fileText.lineBreaks != null) {
            lineBreaks = fileText.lineBreaks.clone();
        }
        else {
            lineBreaks = null;
        }
    }
```

Here `lineBreaks` set to null which is its default value and this is constructor. so if we remove the default value it won't effect anything.
in this case when `fileText.lineBreaks == null` by default it will asign  `lineBreaks = null;`
it will not make any effect to code

so finall code we write as 
```
    public FileText(FileText fileText) {
        file = fileText.file;
        charset = fileText.charset;
        fullText = fileText.fullText;
        lines = fileText.lines.clone();
        if (fileText.lineBreaks != null) {
            lineBreaks = fileText.lineBreaks.clone();
        }
    }
```

### DIFF
--- a/config/archunit-store/slices-should-be-free-of-cycles-suppressions
+++ b/config/archunit-store/slices-should-be-free-of-cycles-suppressions
@@ -3,9 +3,9 @@ Cycle detected: Slice api -> \
                 Slice api\
   1. Dependencies of Slice api\
     - Method <com.puppycrawl.tools.checkstyle.api.AbstractFileSetCheck.process(java.io.File, com.puppycrawl.tools.checkstyle.api.FileText)> calls method <com.puppycrawl.tools.checkstyle.utils.CommonUtil.matchesFileExtension(java.io.File, [Ljava.lang.String;)> in (AbstractFileSetCheck.java:97)\
-    - Constructor <com.puppycrawl.tools.checkstyle.api.FileText.<init>(java.io.File, java.util.List)> gets field <com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_STRING_ARRAY> in (FileText.java:136)\
+    - Constructor <com.puppycrawl.tools.checkstyle.api.FileText.<init>(java.io.File, java.util.List)> gets field <com.puppycrawl.tools.checkstyle.utils.CommonUtil.EMPTY_STRING_ARRAY> in (FileText.java:133)\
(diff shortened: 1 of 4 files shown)


Kevin222004:
How to do regression testing here?

Kevin222004:
ready for review

romani:
resolve Checker violation.

This class is used by all Checks and Filters, I think this is a case where we do not need diff regression. Green CI is enough.

Kevin222004:
Done
I have suppressed the new surviving one.
Reason
`<message>the constructor does not initialize fields: lineBreaks</message>`
if I will initialize the lineBreak then it is still the same pitest issue their

romani (approved):
Ok to merge

Vyom-Yadav (approved):
LGTM!

PR #13178, opened 2023-06-07
Issue #13109: Kill mutation for FullIdent.java

Kevin222004:
Issue #13109: Kill mutation for FullIdent.java

--------------

# Mutation Covered 
https://github.com/checkstyle/checkstyle/blob/b8753a7a025f706d41e6e7b1de302d94da3b8c0f/config/pitest-suppressions/pitest-api-suppressions.xml#L57-L73

---------------

# Explaination
So we have code :- 
https://github.com/checkstyle/checkstyle/blob/b8753a7a025f706d41e6e7b1de302d94da3b8c0f/src/main/java/com/puppycrawl/tools/checkstyle/api/FullIdent.java#L119-L128
in which 
https://github.com/checkstyle/checkstyle/blob/b8753a7a025f706d41e6e7b1de302d94da3b8c0f/src/main/java/com/puppycrawl/tools/checkstyle/api/FullIdent.java#L120
is mutated as 
`DetailAST expression = arrayDeclarator`

Here expression is assigned as `arrayDeclarator.getFirstChild` 
when the loop is going to start it will check whether the expression is null or not know if we remove `.getFirstChild` then `expression=arrayDeclarator` know arrayDeclarator is never going to be null so the loop will always execute, in the loop we have if a condition like `expression.getType() == TokenTypes.EXPR` which always becomes false because ast is an arrayDeclarator token.
and then at last `expression = expression.getFirstChild()` hence here is the case why the removal will not affect.
the expression here is again assigned as `expression.getFirstChild`. so it is not possible a test case in which we can create a test where `.getFirstChild()` will fail if we assigned it or not it will always execute.

by removal of this effect would the loop run one time more than it was running before.

------------

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/1f89420b96e95d45a945d0f8846d0d0f/raw/ed5ab114bdbdd1359c85db6edf11c4126a0e5965/fullident.xml
Report label: Regression-2

### DIFF
--- a/config/pitest-suppressions/pitest-api-suppressions.xml
+++ b/config/pitest-suppressions/pitest-api-suppressions.xml
@@ -55,24 +55,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>FullIdent.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.api.FullIdent</mutatedClass>
-    <mutatedMethod>isArrayTypeDeclaration</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getFirstChild with receiver</description>
-    <lineContent>DetailAST expression = arrayDeclarator.getFirstChild();</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>FullIdent.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.api.FullIdent</mutatedClass>
-    <mutatedMethod>isArrayTypeDeclaration</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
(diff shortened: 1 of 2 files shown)


rdiachenko:
@Kevin222004 please generate a report

Kevin222004:
@rdiachenko How can i generate report for this change

rdiachenko:
> @rdiachenko How can i generate report for this change

Take few checks which use that class and generate report for them.

romani:
Not a class, but affected.

This class is unfortunate API, so by API rules we do not know it can be used, it just should work in all cases.

romani (approved):
Ok to merge as some extra testing is provided

Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

rdiachenko (approved):
lgtm

Vyom-Yadav (approved):
https://github.com/checkstyle/checkstyle/blob/3f77dd11d7e7ac8e23ecc817a1caeeea7d1a489d/src/main/java/com/puppycrawl/tools/checkstyle/api/FullIdent.java#L82-L86

These lines cover the prior conditions of the ast passed to the method to be non null and not of type `EXPR` (as if ast is of type `EXPR` a case exists where if `EXPR` is passed, it will also break from the loop).

---

LGTM!

PR #13192, opened 2023-06-10
Issue #13109: Kill mutation for FileText

Kevin222004:
Issue #13109: Kill mutation for FileText

---------

# mutation covered 
https://github.com/checkstyle/checkstyle/blob/e2d41a956becef80a6ba59ed1cb11a76906ffe59/config/pitest-suppressions/pitest-api-suppressions.xml#L2-L10

-----

# Explaination
One more bad side effect of pitest.

https://github.com/checkstyle/checkstyle/blob/63cd22df005ed5113247de84d3920d0893533dc1/src/main/java/com/puppycrawl/tools/checkstyle/api/FileText.java#L131
this line is mutated and it is declared in the constructor. and  https://github.com/checkstyle/checkstyle/blob/63cd22df005ed5113247de84d3920d0893533dc1/src/main/java/com/puppycrawl/tools/checkstyle/api/FileText.java#L78 it is a final we cannot remove it from the constructor. 

How to resolve this mutation?

### DIFF
--- a/config/pitest-suppressions/pitest-api-suppressions.xml
+++ b/config/pitest-suppressions/pitest-api-suppressions.xml
@@ -4,15 +4,6 @@
     <sourceFile>FileText.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.api.FileText</mutatedClass>
     <mutatedMethod>&lt;init&gt;</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable charset</description>
-    <lineContent>charset = null;</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>FileText.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.api.FileText</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
     <description>removed call to java/nio/charset/CharsetDecoder::onMalformedInput</description>
     <lineContent>decoder.onMalformedInput(CodingErrorAction.REPLACE);</lineContent>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/FileText.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/FileText.java
@@ -75,7 +75,7 @@ public final class FileText {
      * The charset used to read the file.
      * {@code null} if the file was reconstructed from a list of lines.
      */
-    private final Charset charset;
+    private Charset charset;
 
     /**
      * The lines of the file, without terminators.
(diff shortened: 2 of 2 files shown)


Kevin222004:
@rdiachenko can we discuss this mutation

romani:
```
/**
     * The charset used to read the file.
     * {@code null} if the file was reconstructed from a list of lines.
     */
```

Please recheck how charset is used outside of this class. We cover this mutation, we need test that execute some module that care about charset.

config/pitest-suppressions/pitest-api-suppressions.xml
```diff
@@ -1,14 +1,5 @@
...
 <suppressedMutations>
-  <mutation unstable="false">
-    <sourceFile>FileText.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.api.FileText</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable charset</description>
-    <lineContent>charset = null;</lineContent>
-  </mutation>
-
```
Vyom-Yadav, line 11: This is an equivalent mutation according to pitest docs. See https://pitest.org/quickstart/mutators/#EXPERIMENTAL_MEMBER_VARIABLE.

This mutator is replacing:
```java
charset = null;
```
with 
```java
charset = null;
```
Yes, it is not doing anything, the code isn't actually mutated. Pitest blindly replaces member variable values with default values. So no test would be able to kill this mutation. 

I opened https://github.com/hcoles/pitest/issues/1230 for this.

@romani @rdiachenko Please share your thoughts on this one.

romani: We can estimate how much value this mutator  introduced. If we fixed a lot from it already and this is just rare case of unfortunate behavior, we can keep special issue in our project (linked to pitest) to wait pitest fix and keep them in survivals.

If almost all are same problem, we can deactivate whole mutator in pom.xml .

@Vyom-Yadav , what is your opinion on this?

Vyom-Yadav: @romani There are 114 mutations from this mutator right now:
```bash
$ grep -ir "MemberVariableMutator" config/pitest-suppressions | sed '/^\s*$/d' | wc -l
114
```

I manually went through a few mutations to see the value this introduces. From what I saw, there is a mix of actual and false-positive mutations. I would vote for not-deactivating this mutator as it has significant cases that carry value. Let's trust pitest maintainers to fix this issue like they previously did at https://github.com/hcoles/pitest/issues/1066

romani: I have no worries that they will fix it, I just try to think on how to postpone some resolution to the end in hope that fix can be done in next 1-3 months.

romani: @Kevin222004 , as you go through survivals please mark them in issue description by ` delayed until hcoles/pitest#1066` to make it clear that you looked at it and think that better to not do it.

Kevin222004: ok, will update issue description as i went through this type of mutation

romani:
I think we should close this PR.
@Kevin222004, please do if you agree and marked it in tracking issue by comment.

Kevin222004:
@romani can we close this

romani:
No need to ask, it is your PR, just close it

PR #13212, opened 2023-06-13
Issue #13109: Kill mutation for CustomImportOrderCheck 4

Kevin222004:
Issue #13109: Kill mutation for CustomImportOrderCheck 4

----

# check 
https://checkstyle.org/config_imports.html#CustomImportOrder

------

# Mutation 
https://github.com/checkstyle/checkstyle/blob/79d3c5536760d2645b0676c5a7f31da3bb0eed72/config/pitest-suppressions/pitest-imports-suppressions.xml#L30-L37

----------

# Explaination

This is setter for  https://github.com/checkstyle/checkstyle/blob/79d3c5536760d2645b0676c5a7f31da3bb0eed72/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java#L726
 which is https://github.com/checkstyle/checkstyle/blob/79d3c5536760d2645b0676c5a7f31da3bb0eed72/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java#L809-L817 
we are not directly using this property hence all the rule will be first stored in https://github.com/checkstyle/checkstyle/blob/79d3c5536760d2645b0676c5a7f31da3bb0eed72/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java#L720

hence in all the code wherever we define the property all the logic has been written using  
https://github.com/checkstyle/checkstyle/blob/79d3c5536760d2645b0676c5a7f31da3bb0eed72/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java#L720
so no test case is possible

--------

# Regression

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/2463f07cae06e0317712d3aae38b3d0b/raw/42a3cd2e7a12eb3b02ccac0906aace21ddaa02ce/custom.xml
Report label: Report 1

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java
@@ -716,14 +716,14 @@ public class CustomImportOrderCheck extends AbstractCheck {
     /** Pattern used to separate groups of imports. */
     private static final Pattern GROUP_SEPARATOR_PATTERN = Pattern.compile("\\s*###\\s*");
 
+    /** Specify format of order declaration customizing by user. */
+    private static final String DEFAULT_CUSTOM_IMPORT_ORDER_RULES = "";
+
     /** Processed list of import order rules. */
     private final List<String> customOrderRules = new ArrayList<>();
 
     /** Contains objects with import attributes. */
     private final List<ImportDetails> importToGroupList = new ArrayList<>();
-
-    /** Specify format of order declaration customizing by user. */
-    private String customImportOrderRules = "";
 
     /** Specify RegExp for SAME_PACKAGE group imports. */
     private String samePackageDomainsRegExp = "";
@@ -807,13 +807,12 @@ public class CustomImportOrderCheck extends AbstractCheck {
      *        user value.
      */
     public final void setCustomImportOrderRules(final String inputCustomImportOrder) {
-        if (!customImportOrderRules.equals(inputCustomImportOrder)) {
+        if (!DEFAULT_CUSTOM_IMPORT_ORDER_RULES.equals(inputCustomImportOrder)) {
             for (String currentState : GROUP_SEPARATOR_PATTERN.split(inputCustomImportOrder)) {
                 addRulesToList(currentState);
             }
             customOrderRules.add(NON_GROUP_RULE_GROUP);
         }
-        customImportOrderRules = inputCustomImportOrder;
     }
 
     @Override
--- a/config/pitest-suppressions/pitest-imports-suppressions.xml
+++ b/config/pitest-suppressions/pitest-imports-suppressions.xml
@@ -28,15 +28,6 @@
   </mutation>
 
   <mutation unstable="false">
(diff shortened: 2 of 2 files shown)


Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java
```diff
@@ -807,13 +807,12 @@ public final void setSortImportsInGroupAlphabetically(boolean value) {
      *        user value.
      */
     public final void setCustomImportOrderRules(final String inputCustomImportOrder) {
-        if (!customImportOrderRules.equals(inputCustomImportOrder)) {
+        if (!CUSTOM_IMPORT_ORDER_RULES.equals(inputCustomImportOrder)) {
```
rdiachenko, line 810: I don't get it :) Why do we need `CUSTOM_IMPORT_ORDER_RULES` when we just compare the input with empty string in this place only?

romani: We have fields as constants for default values of all users accessible properties.
 But I agree it looked weird even for me.

If expression as inputCustomImportOrder.isEmpty will looks better. 

Naming constant as DEFAULT_xxxxxx would make it more clear, but all other default value constants are not named like this. Should we rename all of them to be more explicit?

But end up in gray area of ... What is default value of property for this module, as for user it is String.
https://checkstyle.org/config_imports.html#CustomImportOrder_Properties

rdiachenko: I see, so it's a Checkstyle's requirement that the user's property field must be present even though it is not actually used. This allows setter to be called and use incoming value in the way we want.

I don't see we use DEFAULT_xxx that much in other checks, but I agree that `DEFAULT_CUSTOM_IMPORT_ORDER_RULES.equals(inputCustomImportOrder)` is more readable than `CUSTOM_IMPORT_ORDER_RULES.equals(inputCustomImportOrder)`

Kevin222004: Ok then I have changed name as `DEFAULT_CUSTOM_IMPORT_ORDER_RULES`

romani: It is not a requirement, but end up our pattern for most of Checks.

rdiachenko (approved):
lgtm

Vyom-Yadav (approved):
LGTM!

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java
Kevin222004: @rdiachenko @romani ping

Issue #13100, opened 2023-05-29
Create tests from xdoc examples


stoyanK7:
For example,
![Screenshot from 2023-05-21 16-09-17](https://github.com/checkstyle/checkstyle/assets/23459549/0fd4c763-ad6b-4544-b9d5-787dedb5dc91)
should turn into
```java
/*xml
<module name="Checker">
  <module name="TreeWalker">
    <module name="AbstractClassName"/>
  </module>
</module>


*/

package com.puppycrawl.tools.checkstyle.checks.naming.abstractclassname;

// xdoc section -- start
class Test1{
  abstract class AbstractFirstClass {} // OK
  abstract class SecondClass {} // violation
  class AbstractThirdClass {} // violation
  class FourthClass {} // OK
}
// xdoc section -- end
```

---
Edit: As time progressed since this issue was created, a few things changed. Example configurations should be in XML (as shown above) and the code should have delimiters surrounding the class definition. Moreover, examples will be excluded from xdoc-javadoc validation. When this issue is closed, there will be a list of all modules below in `XdocsJavadocsTest`. This list can be removed along with the `if`-statements using it.

For convenience, this issue takes care of multiple things - moving examples to test files, placing macros in place of the examples, adding id to each example, and removing examples from class javadoc.

- [x] AbbreviationAsWordInName
- [x] AbstractClassName (needs anchors)
- [x] AnnotationLocation
- [x] AnnotationOnSameLine
- [x] AnnotationUseStyle
- [x] AnonInnerLength
- [x] ArrayTrailingComma
- [x] ArrayTypeStyle
- [x] AtclauseOrder
- [x] AvoidDoubleBraceInitialization
- [x] AvoidEscapedUnicodeCharacters
- [x] AvoidInlineConditionals
- [x] AvoidNestedBlocks
- [x] AvoidNoArgumentSuperConstructorCall
- [x] AvoidStarImport
- [x] AvoidStaticImport
- [x] BooleanExpressionComplexity
- [x] CatchParameterName
- [x] ClassDataAbstractionCoupling
- [x] ClassFanOutComplexity
- [x] ClassMemberImpliedModifier
- [x] ClassTypeParameterName
- [x] CommentsIndentation
- [x] ConstantName
- [x] CovariantEquals
- [x] CustomImportOrder
- [x] CyclomaticComplexity
- [x] DeclarationOrder
- [x] DefaultComesLast
- [x] DescendantToken
- [x] DesignForExtension
[...]

Issue #12155, opened 2022-09-02
BDD configuration comments should be in XML.


nrmancuso:
Consider:

```bash
~/I/checkstyle/s/t/resources/c/p/t/c/checks/im/importorder master *1 !2 > cat InputImportOrderStaticGroupOrderBottom_Negative2.java
/*
ImportOrder
option = top
groups = org, java
ordered = (default)true
separated = true
separatedStaticGroups = (default)false
caseSensitive = (default)true
staticGroups = (default)
sortStaticImportsAlphabetically = true
useContainerOrderingForStatic = (default)false
tokens = (default)STATIC_IMPORT


*/

package com.puppycrawl.tools.checkstyle.checks.imports.importorder;

import org.antlr.v4.runtime.*; // ok

import static java.lang.Math.PI; // violation
import static org.antlr.v4.runtime.Recognizer.EOF; // ok

import java.util.Set; // ok

public class InputImportOrderStaticGroupOrderBottom_Negative2
{

}

```

If I want to test or use this config locally or set up an example for an issue report, I need to manually translate this into a configuration module:

```
~/IdeaProjects/checkstyle_tester/src > cat import_config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
        <module name="ImportOrder">
          <property name="option" value="top"/>
          <property name="groups" value="org, java"/>
          <property name="ordered" value="true"/>
          <property name="separated" value="true"/>
          <property name="separatedStaticGroups" value="false"/>
          <property name="caseSensitive" value="true"/>
          <property name="staticGroups" value=""/>
          <property name="sortStaticImportsAlphabetically" value="true"/>
          <property name="useContainerOrderingForStatic" value="false"/>
          <property name="tokens" value="STATIC_IMPORT"/>
        </module>
    </module>
</module>


```

I am proposing that we migrate from our custom config/ config parser, to simply extract xml from block comment in header to create check module configs.

Example:
```
~/I/checkstyle/s/t/resources/c/p/t/c/checks/im/importorder master *1 !2 > cat InputImportOrderStaticGroupOrderBottom_Negative2.java
/*
<module name="Checker">
    <property name="charset" value="UTF-8"/>

[...]

PR #13218, opened 2023-06-13
Issue #13100: Create tests from EmptyLineSeparator xdoc examples

stoyanK7:
Part of #13100 

https://checkstyle.sourceforge.io/config_whitespace.html#EmptyLineSeparator

### DIFF
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/Example1.java
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/Example1.java
@@ -1,0 +1,31 @@
+/*
+EmptyLineSeparator
+
+
+*/
+
+///////////////////////////////////////////////////
+//HEADER
+///////////////////////////////////////////////////
+package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator;
+import java.io.Serializable;
+// violation 2 lines above ''package' should be separated from previous line'
+// violation 2 lines above ''import' should be separated from previous line'
+
+class Example1 {
+
+  int var1 = 1;
+  int var2 = 2; // violation ''VARIABLE_DEF' should be separated from previous line'
+
+
+  int var3 = 3;
+
+
+  void method1() {}
+  void method2() { // violation ''METHOD_DEF' should be separated from previous line'
+      int var4 = 4;
+
+
+      int var5 = 5;
+  }
+}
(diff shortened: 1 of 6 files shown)


src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/Example1.java
```diff
@@ -0,0 +1,37 @@
+/*
+EmptyLineSeparator
+allowNoEmptyLineBetweenFields = (default)false
+allowMultipleEmptyLines = (default)true
```
nrmancuso, line 4: Do we have validation on defaults? Example: if we change the default value of this property to false, will we fail somewhere? Otherwise, I do not see the value in having a bunch of default property values in these configs. This seems analogous to the extraneous `// ok` comments; it is just noise.

cc @romani @rnveach

nrmancuso (changes requested):
Item:

nrmancuso:
@stoyanK7 did we decide not to pursue xml configs in comments?

stoyanK7:
@nrmancuso 
> did we decide not to pursue xml configs in comments?

Are you talking about https://github.com/checkstyle/checkstyle/issues/12155? Everybody is fine with the decision.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/Example1.java
nrmancuso: > Example: if we change the default value of this property to false, will we fail somewhere?

To be clear, default value of the property refers to the actual default value defined in the check.

nrmancuso:
> @nrmancuso
> 
> > did we decide not to pursue xml configs in comments?
> 
> Are you talking about #12155? Everybody is fine with the decision.

Yes . If we are doing this, aren’t we making more work by not doing this in these example files now? If the parser doesn’t complain now (I.e. ignores the xml), just define the check module as we do now, then the xml. We can remove all old module definitions later by `sed`.

stoyanK7:
This doesn't work
> com.puppycrawl.tools.checkstyle.api.CheckstyleException: cannot initialize module com.puppycrawl.tools.checkstyle.TreeWalker - cannot initialize module com.puppycrawl.tools.checkstyle.checks.whitespace.EmptyLineSeparatorCheck - Property '<module' does not exist, please check the documentation

stoyanK7:
We can put it in a comment below
```
/*
EmptyLineSeparator

*/
/*
<module name="EmptyLineSeparator"/>
*/
```

stoyanK7:
Update: The decision from Discord is to not do the XML comments but rather convert them later. If we do XML now, later when we remove it, expected violation line numbers will have to be touched as well.

stoyanK7:
Removed `(default)` properties.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/Example1.java
stoyanK7: @nrmancuso Done. Removed `(default)` properties.

romani (approved):
Ok to merge

PR #13219, opened 2023-06-13
Issue #13100: Create tests from FileTabCharacter xdoc examples 

stoyanK7:
Part of #13100

### DIFF
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/filetabcharacter/Example4.xml
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/filetabcharacter/Example4.xml
@@ -1,0 +1,13 @@
+/*
+FileTabCharacter
+fileExtensions = java, xml
+
+
+*/
+// xdoc section -- start
+<?xml version="1.0" encoding="UTF-8" ?>
+<UserAccount>
+	<FirstName>John</FirstName> // violation 'File contains tab characters'
+	<LastName>Doe</LastName>    <!-- OK, only first occurrence in file reported -->
+</UserAccount>
+// xdoc section -- end
--- a/pom.xml
+++ b/pom.xml
@@ -1593,6 +1593,7 @@
                 <exclude>.ci-temp/**</exclude>
                 <!-- Specially crafted test files -->
                 <exclude>src/test/resources/**</exclude>
+                <exclude>src/xdocs-examples/resources/**</exclude>
                 <!-- Generated files -->
                 <exclude>src/xdocs/releasenotes.xml</exclude>
                 <exclude>src/xdocs/releasenotes_old*.xml</exclude>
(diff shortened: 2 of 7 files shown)


src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/filetabcharacter/Example4.xml
```diff
@@ -0,0 +1,13 @@
+/*
+FileTabCharacter
+eachLine = (default)false
+fileExtensions = java, xml
+
+
+*/
```
nrmancuso, line 7: While the html config "comment" is debatable (I am not a big fan), this is not valid xml. As we demand compilation for java files, we should have valid xml. Are we able to simply put this comment inside of a valid multiline xml comment? If fix is that easy, we should also update html file.

nrmancuso (changes requested):
Item:

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/filetabcharacter/Example4.xml
stoyanK7: Me too, but this is the only way for now. I tried putting the comment as such
```diff
+<!--
/*
FileTabCharacter
eachLine = (default)false
fileExtensions = java, xml


*/
+-->
```

but the `InlineConfigParser` is strict about it
> Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Config not specified on top.Please see other inputs for examples of what is required.


https://github.com/checkstyle/checkstyle/blob/79d3c5536760d2645b0676c5a7f31da3bb0eed72/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L164-L167

Maybe we can create an issue to extend and allow others?

nrmancuso: Please do, and take a few minutes to scope work required to allow us to parse config comment successfully with inner Java comment and without (only module config in xml comment).

stoyanK7: Opened at https://github.com/checkstyle/checkstyle/issues/13222

stoyanK7: I will remove the commit when #13223 gets merged. I continue with adding other tests.

stoyanK7: @nrmancuso I see that the PR/Issue might drag for some time. Let's leave it as is and we can come back to this issue at some point.

nrmancuso: Agreed.

nrmancuso (approved):

romani (approved):
Ok to merge

PR #13223, opened 2023-06-14
Issue #13222: Accept more types of comments in InlineConfigParser

### LINKED ISSUE #13222: Accept more types of comments in `InlineConfigParser`
From https://github.com/checkstyle/checkstyle/pull/13219#discussion_r1228944093

> While the html config "comment" is debatable (I am not a big fan), this is not valid xml. As we demand compilation for java files, we should have valid xml. Are we able to simply put this comment inside of a valid multiline xml comment? If fix is that easy, we should also update html file.

---
All inline test configurations have to start with `/*` which is an issue with files whose comments are different. For example `xml`,`html` with `<!--` and `-->`. The purpose of this issue is to allow more types of comments be accepted in test configurations.

stoyanK7:
Resolves #13222

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
@@ -19,6 +19,8 @@
 
 package com.puppycrawl.tools.checkstyle.bdd;
 
+import static com.puppycrawl.tools.checkstyle.utils.CommonUtil.getFileExtension;
+
 import java.io.File;
 import java.io.IOException;
 import java.io.StringReader;
@@ -116,6 +118,12 @@ public final class InlineConfigParser {
             "com.puppycrawl.tools.checkstyle.checks.TranslationCheck"
     );
 
+    /** File extensions and the allowed comment delimiters.  */
+    private static final Map<String, String[]> FILE_EXTENSIONS_COMMENT_DELIMITERS =
+            Map.of("java", new String[] {"/*", "*/"},
+                   "html", new String[] {"<!--", "-->"},
+                   "xml", new String[] {"<!--", "-->"});
+
     /** Stop instances being created. **/
     private InlineConfigParser() {
     }
@@ -161,11 +169,17 @@ public final class InlineConfigParser {
     private static void setModules(TestInputConfiguration.Builder testInputConfigBuilder,
                                    String inputFilePath, List<String> lines)
             throws Exception {
-        if (!lines.get(0).startsWith("/*")) {
+        final String fileExtension = getFileExtension(inputFilePath);
+        final String[] commentDelimiters = FILE_EXTENSIONS_COMMENT_DELIMITERS
+                .getOrDefault(fileExtension, FILE_EXTENSIONS_COMMENT_DELIMITERS.get("java"));
+        final String openingComment = commentDelimiters[0];
+
+        if (!lines.get(0).startsWith(openingComment)) {
             throw new CheckstyleException("Config not specified on top."
                 + "Please see other inputs for examples of what is required.");
         }
         int lineNo = 1;
+        final String closingComment = commentDelimiters[1];
         do {
             final ModuleInputConfiguration.Builder moduleInputConfigBuilder =
                     new ModuleInputConfiguration.Builder();
@@ -175,7 +189,7 @@ public final class InlineConfigParser {
             do {
                 lineNo++;
             } while (lines.get(lineNo).isEmpty() || !lines.get(lineNo - 1).isEmpty());
-        } while (!lines.get(lineNo).startsWith("*/"));
+        } while (!lines.get(lineNo).startsWith(closingComment));
     }
 
     private static String getFullyQualifiedClassName(String filePath, String moduleName)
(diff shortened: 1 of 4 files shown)


nrmancuso (changes requested):
So, with current implementation, it looks like this is ok:
```
/*

MyCheck

—>
```
Seems like we should have some simple validation on this. @romani ?

If we have file extension easily available, it might be better to have a map or something of file extension -> valid comment delimiters.

stoyanK7:
You're right. Very short-sighted of me. Let me see.

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -175,7 +175,15 @@ private static void setModules(TestInputConfiguration.Builder testInputConfigBui
             do {
                 lineNo++;
             } while (lines.get(lineNo).isEmpty() || !lines.get(lineNo - 1).isEmpty());
-        } while (!lines.get(lineNo).startsWith("*/"));
+        } while (!startsWithClosingComment(lines.get(lineNo)));
+    }
+
+    private static boolean startsWithOpeningComment(String line) {
+        return line.startsWith("/*") || line.startsWith("<!--");
```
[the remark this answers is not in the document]
stoyanK7: Please check https://github.com/checkstyle/checkstyle/pull/13219#discussion_r1228944093.
> While the html config "comment" is debatable (I am not a big fan), this is not valid xml. As we demand compilation for java files, we should have valid xml. Are we able to simply put this comment inside of a valid multiline xml comment? If fix is that easy, we should also update html file.

We want to have it for inline configs in `xml`, `html`. Migration to `/*xml` is a separate issue.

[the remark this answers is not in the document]
romani: Ok, it might be better to limit usage of not java files as much as possible, and keep them as exception for now and deal with them at the end of project. 
I am ok to have such Examples to out of our project at all. 
Let's focus on java problem first.

stoyanK7:
@nrmancuso Done. I took your suggestion into consideration.

nrmancuso (approved):
Awesome, I am glad this worked out to be a simple update.

src/test/resources/com/puppycrawl/tools/checkstyle/filters/suppresswithnearbytextfilter/InputSuppressWithNearbyTextFilterMessagePattern.xml
```diff
@@ -20,7 +20,7 @@ maximum = (default)0
 ignoreComments = (default)false
 
 
-*/
+-->
 // violation 9 lines above 'Line matches the illegal pattern .*'
 
 <?xml version="1.0"?>
```
romani, line 26: This is not valid xml .
`<?xml` must be first line.
`// violation` not allowed content.

I loosing reason of update. What is a benefit ?

romani (changes requested):
Items

src/test/resources/com/puppycrawl/tools/checkstyle/filters/suppresswithnearbytextfilter/InputSuppressWithNearbyTextFilterMessagePattern.xml
stoyanK7: > This is not valid xml .
> ?xml must be first line.

Yeah, but the rule of having the config on top will be broken if we go this path. Depends on which you believe is more-important - having config always on top or 100% validity.

> I loosing reason of update. What is a benefit ?

To make test file valid with native file extension comments. Moreover, helps when editing in an IDE.

> `// violation` not allowed content

Agreed. It is a separate issue at https://github.com/checkstyle/checkstyle/issues/13233. We can deal with that later.

romani: We just treat all not java files as text file.
Extension doesn't matter at all.

We can rename all such SQL, bash, xml file to be `some-name.xml.txt` and keep it clear it is txt with xml flavor/content, but it is txt.
It is txt, as we do not parse content as we do in java files. All of them are plain text.

And in plain text files we do have some lines that similar to java comments. But many other languages use such comments too. So `/*` is our start marker for special content.

One more time: I probably asked a long time ago to use xml and SQL and bash files just to show user that we can do something in text files. All of them are plain text files.

nrmancuso: > This is not valid xml . `<?xml` must be first line.

I do not think this is accurate for xml v1. 

As for comments, we could add one more forward slash to make a valid comment. However, I do not want to waste a bunch of time debating this or tweaking stuff. If we cannot reach agreement quickly and implementation easily, we should move ahead with imaginary input file extensions.

romani: > with imaginary input file extensions.

not imaginary, they are `.txt`. and we can name it  as we want:
`InputSuppressWithNearbyTextFilterMessagePattern-xml.txt`
`InputSuppressWithNearbyTextFilterMessagePatternXml.txt`
`InputSuppressWithNearbyTextFilterMessagePattern.xml.txt`

romani: Better to do in separate issue https://github.com/checkstyle/checkstyle/issues/13267

PR #13224, opened 2023-06-14
Issue #13100: Create tests from LambdaBodyLength xdoc examples

stoyanK7:
Part of #13100

https://checkstyle.sourceforge.io/config_sizes.html#LambdaBodyLength_Examples

### DIFF
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/lambdabodylength/Example1.java
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/lambdabodylength/Example1.java
@@ -1,0 +1,43 @@
+/*
+LambdaBodyLength
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.sizes.lambdabodylength;
+
+class Example1 {
+  Runnable r = () -> { // OK, length is 10
+    System.out.println(2); // line 2 of lambda
+    System.out.println(3);
+    System.out.println(4);
+    System.out.println(5);
+    System.out.println(6);
+    System.out.println(7);
+    System.out.println(8);
+    System.out.println(9);
+  }; // line 10
+  Runnable r2 = () -> { // violation 'length is 11'
+    System.out.println(2); // line 2 of lambda
+    System.out.println(3);
+    System.out.println(4);
+    System.out.println(5);
+    System.out.println(6);
+    System.out.println(7);
+    System.out.println(8);
+    System.out.println(9);
+    System.out.println(10);
+  }; // line 11
+  Runnable r3 = () -> // violation 'length is 11'
+    "someString".concat("1") // line 1 of lambda
+                .concat("2")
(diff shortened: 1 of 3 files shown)


src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/lambdabodylength/Example1.java
```diff
@@ -0,0 +1,45 @@
+/*
+LambdaBodyLength
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.sizes.lambdabodylength;
+
+class Example1 {
+  Runnable r = () -> { // violation 'length is 11 .* (max allowed is 10).'
```
nrmancuso, line 10: Seems reasonable to move "good" lambda to be `r`, and place `ok` comment with explanation there.

nrmancuso (changes requested):
Item:

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/lambdabodylength/Example1.java
romani: We can skip `.* (max allowed is 10).` as example CI fig is above all is clear and short.

Better to not use regexp in Examples.

romani: My request above sounds like a regexp check idea for Examples. Is is doable in simple way ?

stoyanK7: > Seems reasonable to move "good" lambda to be r, and place ok comment with explanation there.

> We can skip .* (max allowed is 10). as example CI fig is above all is clear and short.

Done.

> My request above sounds like a regexp check idea for Examples. Is is doable in simple way?

A regexp to check for regex expressions in the `// violation` comments? We have to be very specific - for example escapes `\.`, `\+`, `\(`, `\)`, `\[`, `\]` or common expressions `.*`

romani: >common expressions .*

Let's target this most common mistake. We will see if we can add other also later on.

nrmancuso (approved):

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/lambdabodylength/Example1.java
stoyanK7: > Let's target this most common mistake. We will see if we can add other also later on.

https://github.com/checkstyle/checkstyle/issues/13244 is where discussion continues on this.

romani (approved):

PR #13237, opened 2023-06-16
Issue #13195: Mobile page rendering requires extra font difference for properties

### LINKED ISSUE #13195: Mobile page rendering requires extra font difference for properties
Detected at https://github.com/checkstyle/checkstyle/pull/13103#issuecomment-1585814118

There only pop point to improve is to make name of property to be italic to look different than description.

Screenshot_20230610-135311

stoyanK7:
Resolves #13195

### DIFF
--- a/src/site/resources/css/site.css
+++ b/src/site/resources/css/site.css
@@ -163,4 +163,8 @@ a[title="toTop"] {
     display: block;
     text-align: left;
   }
+
+  section[id$="Properties"] .wrapper .bodyTable tr td:first-child {
+    font-style: italic;
+  }
 }


stoyanK7:
Github, generate site

nrmancuso (approved):
I do not have experience with CSS/ web dev in general, but change looks good to me.

nrmancuso:
@rnveach please take a look, I know you are more experienced with web dev.

romani:
Github, generate site

romani:
Previous website is not opening for me

stoyanK7:
@romani Weird. Works for me. Here is a direct link to some properties: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/0b86150_2023191601/checks/annotation/annotationlocation.html#AnnotationLocation_Properties

romani:
It works now

src/site/resources/css/site.css
```diff
@@ -163,4 +163,8 @@ a[title="toTop"] {
     display: block;
     text-align: left;
   }
+
+  section[id$="_Properties"] .wrapper .bodyTable tr td:first-child {
```
rnveach, line 167: Looks like there is some issue in new pages if the ID starts with `_`. Originally IDs were like `id="AnnotationLocation_Properties"` before we split. The `_` was added to provide a split between check name and section name. The `_` was not removed when it should have been.

rnveach (changes requested):

src/site/resources/css/site.css
romani: Yes, better to be without leading `_`

stoyanK7: Done.

stoyanK7:
Github, generate site

src/site/resources/css/site.css
rnveach: @stoyanK7 I don't see any other code changes but removing the `_`. Is this a `find` and not a complete `match`?

Looking at https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/b00c60f_2023120428/checks/annotation/annotationlocation.html#AnnotationLocation_Description
It seems we still have the same pattern `check name _ section name`. I don't see the point in duplicating the check name with the same page name.
These `href`s should be changed for the new page style.

rnveach: This should be a complete match when the `href`s are fixed.

stoyanK7: Mentioned in issue.

romani:
@rnveach , it looks like ready for final review.

rnveach (approved):

PR #13261, opened 2023-06-19
Issue #13109: Kill mutation for RegexpHeaderCheck

Kevin222004:
Issue #13109: Kill mutation for RegexpHeaderCheck

--- 

# Check  :-  
https://checkstyle.org/config_header.html#RegexpHeader

------

# Mutation
https://github.com/checkstyle/checkstyle/blob/7d35a52067c994eb12312b8fff0a072a198d2c00/config/pitest-suppressions/pitest-header-suppressions.xml#L48-L64

-------

# Explaination
Test added


src/test/java/com/puppycrawl/tools/checkstyle/checks/header/RegexpHeaderCheckTest.java
```diff
@@ -380,4 +380,20 @@ public void testNoHeaderMissingErrorInCaseHeaderSizeEqualToFileSize() throws Exc
         verify(checkConfig, getPath("InputRegexpHeaderMulti52.java"), expected);
     }
 
+    @Test
+    public void testHeaderWithInvalidRegexp2() throws Exception {
```
rdiachenko, line 384: It is almost the same as `testHeaderWithInvalidRegexp()` with the difference in the input file. If you keep existing test and just update its input and expected exception message, would that be enough, so we don't need `testHeaderWithInvalidRegexp2()`?

romani: Yes, let's not be blocked by semaphore.

romani (approved):
Ok to merge

Vyom-Yadav (approved):
LGTM!

rdiachenko (approved):
lgtm

PR #13268, opened 2023-06-20
minor: Kill Mutation for AbstractHeaderCheck

Kevin222004:
minor: Kill Mutation for  AbstractHeaderCheck

for  https://github.com/checkstyle/checkstyle/pull/13269

PR #13269, opened 2023-06-20
Issue #13501: Kill mutation for AbstarctHeaderCheck

Kevin222004:
Issue #13501: Kill mutation for AbstarctHeaderCheck

------

# Check
https://checkstyle.org/config_header.html

-------

# Mutation 
https://github.com/checkstyle/checkstyle/blob/e0d3066777834ba62c489d86bd3edc9727aa79e3/config/pitest-suppressions/pitest-header-suppressions.xml#L3-L46

-------

# Explaination
at https://github.com/checkstyle/checkstyle/pull/13268, specific message https://github.com/checkstyle/checkstyle/pull/13268#issuecomment-1657179042


src/main/java/com/puppycrawl/tools/checkstyle/checks/header/AbstractHeaderCheck.java
```diff
@@ -60,8 +60,7 @@ public abstract class AbstractHeaderCheck extends AbstractFileSetCheck
 
     /** Specify the character encoding to use when reading the headerFile. */
     @XdocsPropertyType(PropertyType.STRING)
-    private Charset charset = createCharset(System.getProperty("file.encoding",
-        StandardCharsets.UTF_8.name()));
+    private final Charset charset = createCharset(StandardCharsets.UTF_8.name());
```
romani, line 63: this system property might be tricky
https://stackoverflow.com/a/362006

romani: what it mean is that we become to ignore this system property, and just use UTF-8 by default.
I am ok, as we have setCharset property to be explicit from our config.

But I think we need to create special issue on this to let users see it release notes as explicit deprecation, not a some pitest refactoring.

romani: discussion happens at https://github.com/checkstyle/checkstyle/pull/13268#discussion_r1249956750 for this problem

romani:
@Kevin222004 , is this duplicate PR of https://github.com/checkstyle/checkstyle/pull/13268

Can we close it ?

Kevin222004:
@romani Yes it is created for discussion purpose 
https://github.com/checkstyle/checkstyle/pull/13269#issuecomment-1657689688

after https://github.com/checkstyle/checkstyle/pull/13268#issuecomment-1657179042 I am closing that 
and update this in few hours and mark for review

romani:
Please add suppression for Checker with same comment above https://github.com/checkstyle/checkstyle/actions/runs/5714370827/job/15481588407?pr=13269#step:6:1658
Please also mention that it is conflict with pitest but value pitest above checker.

Kevin222004:
@romani https://github.com/checkstyle/checkstyle/pull/13269#issuecomment-1658425962 done

config/checkstyle-checks.xml
```diff
@@ -46,6 +46,13 @@
     <property name="files"
               value=".*[\\/]src[\\/](test|it|xdocs-examples)[\\/].*(?&lt;!Support)\.java"/>
   </module>
+  <!-- require for charset testing -->
+  <module name="SuppressionSingleFilter">
+    <property name="checks" value="RegexpSinglelineJava"/>
+    <property name="files"
+              value=".*[\\/]src[\\/]test[\\/]java[\\/].*[\\/]RegexpHeaderCheckTest.java"/>
+    <property name="lines" value="398"/>
```
romani, line 54: Suppression by line is very fragile.
Please share link to check that forcing you to do this.

Is it https://github.com/checkstyle/checkstyle/blob/42f9ceb39a5e4b79e2733c9cedc1766fcafd5cbb/config/checkstyle-checks.xml#L827 ?

If so , use id of Check, not name.
Please use xpath suppression to suppress violation on special this method

romani (changes requested):
Items:

config/checkstyle-checks.xml
Kevin222004: @romani does xpath query generator works only when a violation message contains column no 
https://github.com/checkstyle/checkstyle/blob/34e7d5c714b2df62781790537c6127b5712ec097/src/main/java/com/puppycrawl/tools/checkstyle/xpath/XpathQueryGenerator.java#L111-L112
https://github.com/checkstyle/checkstyle/blob/34e7d5c714b2df62781790537c6127b5712ec097/src/main/java/com/puppycrawl/tools/checkstyle/xpath/XpathQueryGenerator.java#L126-L127

To generate a query

Kevin222004: I have removed line no. I think Xpath Query may not be possible because this error i on line not on any ast token

romani: Please use suppression like https://github.com/checkstyle/checkstyle/blob/19b2594c7766801859e73f2194b55fb00d26f36c/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L819

Kevin222004: done

src/test/java/com/puppycrawl/tools/checkstyle/checks/header/RegexpHeaderCheckTest.java
```diff
@@ -379,4 +379,38 @@ public void testNoHeaderMissingErrorInCaseHeaderSizeEqualToFileSize() throws Exc
...
+        verify(checkConfig, path, expected);
+    }
+
+    @Test
+    public void testCharsetProperty2() throws Exception {
+        final DefaultConfiguration checkConfig = createModuleConfig(RegexpHeaderCheck.class);
+        checkConfig.addProperty("charset", "US-ASCII");
+        checkConfig.addProperty("headerFile", getPath("InputRegexpHeader7.header"));
+        final String[] expected = {
+            // -@cs[RegexpSinglelineJava] need for testing
```
romani, line 398: Try to use `id`.
Please put comment on whole method.

romani (changes requested):
Last:

src/test/java/com/puppycrawl/tools/checkstyle/checks/header/RegexpHeaderCheckTest.java
Kevin222004: It will not work if I put comment id 
image

It will also not work if i put this line at top of method 
image
and
image

romani (approved):
Ok to merge

Vyom-Yadav (approved):
LGTM! @Kevin222004 @romani I think we discussed creating an issue about removing `file.encoding`

romani:
After deeper analysis, I recalled it, it is not required, this property is not effective at all. It remnants of something old.

rdiachenko (approved):
lgtm

rdiachenko:
rebased

PR #13258, opened 2023-06-19
Issue #13097: Split misc xdocs into separate pages

stoyanK7:
Part of #13097

PR #13285, opened 2023-06-21
Pull #13285: Ensure category index page tables are in sync with checks.xml page table

stoyanK7:
Identified at https://github.com/checkstyle/checkstyle/pull/13258#discussion_r1236976318

The tables in `index.xml` in each category(`blocks`, `naming`,etc..) are an exact copy of `checks.xml`. It turned out that the script doesn't extract them correctly, hence we need this check.

Test run found 3 issues merged. Here is one
```
The summary for check WhitespaceAfter in the file "src/xdocs/checks/whitespace/index.xml" should match the summary for this check in the file "src/xdocs/checks.xml"
expected: Checks that a token is followed by whitespace, with the exception that it does not check for whitespace after the semicolon of an empty for iterator.
but was : Checks that there is at least one whitespace after the leading asterisk.
```

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
@@ -285,6 +285,36 @@ public class XdocsPagesTest {
         }
     }
 
+    @Test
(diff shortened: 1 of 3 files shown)


nrmancuso (approved):
Thanks for this.

romani (approved):

PR #13287, opened 2023-06-21
Pull #13287: Automatically scroll to first h2 in mobile views

stoyanK7:
Resolves problem where anchors cannot be added to menu items in `site.xml`
https://github.com/checkstyle/checkstyle/blob/fbbf38a0d9e3517cbea6e41ab95293d6ae20bbb6/src/site/site.xml#L83
doing `checks/annotation/annotationlocation.html#AnnotationLocation` doesn't work since when one clicks on the link, the menu collapses. Look at https://stoyank7.github.io/checks/annotation/index.html. Try to click any annotation check in menu and see how the menu collapses once the page is opened. 

PR changes work only in mobile view (screens below `823px`) and applies to all pages.

### DIFF
--- a/src/site/resources/js/scroll-to-h2.js
+++ b/src/site/resources/js/scroll-to-h2.js
@@ -1,0 +1,16 @@
+window.addEventListener("load", function () {
+    if (window.innerWidth > 823) {
+        return;
+    }
+
+    const anchorIsAlreadySelected = document.URL.includes("#");
+    if (anchorIsAlreadySelected) {
+        return;
+    }
+
+    const elements = document.querySelectorAll("h2 > a[name]");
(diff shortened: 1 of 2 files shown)


stoyanK7:
Github, generate site

stoyanK7:
Maybe having it for all pages is too broad? We can limit it to checks and filters only.

romani:
if anchor can not be added in left panel, it ok, lets see what we can do at the end of summer.
Lets not spend time on this now.

stoyanK7:
Opened https://github.com/checkstyle/checkstyle/issues/13292. I still believe javascript is the only way here but again, we can deal with this later and open this MR at some point in time.

PR #13297, opened 2023-06-23
Issue #13283: Move Filters and File Filters to be a sibling of Checks in site menu

### LINKED ISSUE #13283: Move Filters and File Filters to be a sibling of Checks in site menu
Right now `Filters` and `FileFilters` are under `Configuration`
Screenshot from 2023-06-21 13-56-56

We can move them to be a sibling of `Checks` like this. They are already a big package of modules.
Screenshot from 2023-06-21 13-56-18

stoyanK7:
Resolves #13283

### DIFF
--- a/src/site/site.xml
+++ b/src/site/site.xml
@@ -68,26 +68,6 @@
     <menu name="Documentation">
       <item name="Configuration" href="config.html">
         <item name="Property Types" href="property_types.html"/>
-        <item name="Filters" href="filters/index.html" collapse="true">
-          <item name="SeverityMatchFilter" href="filters/severitymatchfilter.html"/>
-          <item name="SuppressionCommentFilter" href="filters/suppressioncommentfilter.html"/>
-          <item name="SuppressionFilter" href="filters/suppressionfilter.html"/>
-          <item name="SuppressionSingleFilter" href="filters/suppressionsinglefilter.html"/>
-          <item name="SuppressionXpathFilter" href="filters/suppressionxpathfilter.html"/>
-          <item name="SuppressionXpathSingleFilter"
-                href="filters/suppressionxpathsinglefilter.html"/>
-          <item name="SuppressWarningsFilter" href="filters/suppresswarningsfilter.html"/>
-          <item name="SuppressWithNearbyCommentFilter"
-                href="filters/suppresswithnearbycommentfilter.html"/>
-          <item name="SuppressWithNearbyTextFilter"
-                href="filters/suppresswithnearbytextfilter.html"/>
-          <item name="SuppressWithPlainTextCommentFilter"
-                href="filters/suppresswithplaintextcommentfilter.html"/>
-        </item>
-        <item name="File Filters" href="filefilters/index.html" collapse="true">
-          <item name="BeforeExecutionExclusionFileFilter"
-                href="filefilters/beforeexecutionexclusionfilefilter.html"/>
-        </item>
         <item name="System Properties" href="config_system_properties.html"/>
       </item>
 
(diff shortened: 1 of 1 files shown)


stoyanK7:
Github, generate site

romani (approved):
Ok to merge

src/site/site.xml
```diff
@@ -338,6 +318,27 @@
         </item>
       </item>
 
+      <item name="Filters" href="filters/index.html" collapse="true">
```
nrmancuso, line 321: Why are these collapsed by default when other single level items are expanded?

nrmancuso (changes requested):
Item:

rnveach:
Names are too large to fix in navigation.
image

image

src/site/site.xml
stoyanK7: All categories are collapsed - `Annotations`, `Blocks`, etc...

The idea was to keep the original length of the menu. I am not sure if making it longer would be good. Especially on mobile.

nrmancuso: We moved these out of checks since they are not the same, so this is not a good comparison :) If you look at other siblings of these two that are only single depth, they are already expanded.

stoyanK7:
Github, generate site

stoyanK7:
@rnveach Good catch. I have added an extra commit fix. Please check it out. I tested manually on desktop and mobile and it looks good to me now.

src/site/site.xml
romani: I am fine with model: expanding of one collapse others.

nrmancuso (approved):

src/site/resources/css/site.css
```diff
@@ -98,10 +98,10 @@ p, td, table.bodyTable td, th, table.bodyTable th, li, pre, #breadcrumbs span, #
 }
 
 #leftColumn {
-  width: 270px;
+  min-width: 270px;
+  width: fit-content;
```
rnveach, line 102: Min width on my laptop is 250. If we are going to do this, we may as well decrease this.

src/site/resources/css/site.css
```diff
@@ -134,6 +134,10 @@ a[title="toTop"] {
   z-index: 10;
 }
 
+.poweredBy {
+  display: block;
```
rnveach, line 138: But first, why this?

src/site/resources/js/checkstyle.js
```diff
@@ -25,3 +25,21 @@ window.addEventListener("scroll", function () {
...
+        return;
+    }
+
+    // Else calculate margin based on left column width
+    const leftColumnWidth = leftColumn.offsetWidth;
+    bodyColumn.style.marginLeft = `${leftColumnWidth + 27}px`;
+}
+
+window.addEventListener('load', setBodyColumnMargin);
+window.addEventListener('resize', setBodyColumnMargin);
```
rnveach, line 45: And why this? What was wrong with the old CSS handling?

rnveach (commented):
@stoyanK7 Since @romani wants to do a release, it is probably better this to be a single PR and move the 2nd commit to another PR.

stoyanK7:
@rnveach I am removing the second commit from here. I will open another PR for it and answer your reviews there.

romani:
we will deal with better rendering at https://github.com/checkstyle/checkstyle/issues/13304

romani:
linkcheck failure will be fixed at - https://github.com/checkstyle/checkstyle/pull/13305

src/site/site.xml
romani: we will vote at https://github.com/checkstyle/checkstyle/issues/13306

PR #13299, opened 2023-06-23
Issue #13321: Kill mutation for ConfigurationLoader

Kevin222004:
Issue #13321: Kill mutation for ConfigurationLoader

# Mutation 
https://github.com/checkstyle/checkstyle/blob/03aeb21b90499fd51d397b4de224a05a2b0fa05e/config/pitest-suppressions/pitest-common-suppressions.xml#L138-L290


src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java
```diff
@@ -492,7 +471,15 @@ private final class InternalLoader
          */
         private InternalLoader()
                 throws SAXException, ParserConfigurationException {
-            super(createIdToResourceNameMap());
+            super(Map.of(
```
nrmancuso, line 474: Please add comment like `// map from local resources and dtd ids` above superclass constructor call. We should have probably done this in other PR too

src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java
```diff
@@ -492,7 +471,15 @@ private final class InternalLoader
...
+                DTD_PUBLIC_ID_1_0, DTD_CONFIGURATION_NAME_1_0,
+                DTD_PUBLIC_ID_1_1, DTD_CONFIGURATION_NAME_1_1,
+                DTD_PUBLIC_ID_1_2, DTD_CONFIGURATION_NAME_1_2,
+                DTD_PUBLIC_ID_1_3, DTD_CONFIGURATION_NAME_1_3,
+                DTD_PUBLIC_CS_ID_1_0, DTD_CONFIGURATION_NAME_1_0,
+                DTD_PUBLIC_CS_ID_1_1, DTD_CONFIGURATION_NAME_1_1,
+                DTD_PUBLIC_CS_ID_1_2, DTD_CONFIGURATION_NAME_1_2,
+                DTD_PUBLIC_CS_ID_1_3, DTD_CONFIGURATION_NAME_1_3));
```
nrmancuso, line 482: Please line wrap `));`, always think about how to reduce diff on extension of collections.

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java
Kevin222004: > Please add comment like // map from local resources to dtd ids above superclass constructor call.

Done

> We should have probably done this in other PR too.

I update that too

src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java
Kevin222004: done

nrmancuso (approved):

rdiachenko (approved):
lgtm

src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java
```diff
@@ -492,7 +471,17 @@ private final class InternalLoader
          */
         private InternalLoader()
                 throws SAXException, ParserConfigurationException {
-            super(createIdToResourceNameMap());
+            // map from local resources to dtd ids
+            super(Map.of(
```
romani, line 475: this is hacking to only cheat Pitest.

but conceptually this collection is required to load old config files that have reference to old schema name.

can we create config xml file with `-//Puppy Crawl//DTD Check Configuration 1.0//EN` instead of `-//Checkstyle//DTD Checkstyle Configuration 1.0//EN` and make sure config is loadable.

This test is ok to be not Input based.

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java
nrmancuso: > this is hacking to only cheat Pitest.

I disagree. It looks like the method that built this map was written in 2015, before we migrated to JDK11. `Map.of` was released in JDK9. The method we removed did nothing but create a map anyway; we are doing better now that we are returning an Unmodifiable map. We can write any tests we want, but the fact is that passing an Unmodifiable map to the super constructor is better that what we were doing previously.


Since we can have no code before the super constructor call, pre-JDK9 code required some other method to exist. We do not need it any longer.

romani: I have no problem to make `createIdToResourceNameMap()` to return immutable map.

Pitest clearly point us that we have no test for old configs.

romani: @Kevin222004 , please finish this PR.

Kevin222004: Done Test added

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
```diff
@@ -695,4 +696,27 @@ public void testConfigWithIgnoreExceptionalAttributes() {
...
+        };
+
+        for (String configFile : configFiles) {
+            final DefaultConfiguration config =
+                    (DefaultConfiguration) ConfigurationLoader.loadConfiguration(
+                            new InputSource(Files.newInputStream(Paths.get(
+                                    getPath(configFile)))),
+                            new PropertiesExpander(new Properties()),
+                            IgnoredModulesOptions.OMIT);
+            assertThat(config.getPropertyNames()).asList().isEmpty();
```
romani, line 719: I do not understand this assert.
please use getChildren() , config  should have not 0 amount of modules that are loaded.

romani (changes requested):
really good.

last item:

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
```diff
@@ -695,4 +695,29 @@ public void testConfigWithIgnoreExceptionalAttributes() {
...
+        for (String configFile : configFiles) {
+            final DefaultConfiguration config =
+                    (DefaultConfiguration) ConfigurationLoader.loadConfiguration(
+                            new InputSource(Files.newInputStream(Paths.get(
+                                    getPath(configFile)))),
+                            new PropertiesExpander(new Properties()),
+                            IgnoredModulesOptions.OMIT);
+            assertWithMessage("should not have children")
+                    .that(config.getChildren())
+                    .isEmpty();
```
romani, line 720: Should not be empty.....

Strange that it is empty , please look for a reason.
Confiy should be loadable and produce violation if required.

Can we regular verify test method execution?

romani (changes requested):
Items:

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
Kevin222004: @romani If i am not wrong all the content of all the xml contains checker only 

```
<module name="Checker">
</module>
```
if the file is like 
```
<module name="Checker">
    <module name="TreeWalker">
        
    </module>
</module>

```
then the child is not empty, because it is treewalker

> Can we regular verify test method execution? 

for this I have to extend AbstractModuleTestSupport and this test class already extends AbstractPathTestSupport

romani: we can skip regular verify method execution here, but we need ^^^^

Kevin222004: Done please check

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
```diff
@@ -695,4 +695,44 @@ public void testConfigWithIgnoreExceptionalAttributes() {
...
+                    .that(config.getAttribute("severity"))
+                    .isEqualTo("error");
+
+            assertWithMessage("")
+                    .that(config.getAttribute("fileExtensions"))
+                    .isEqualTo("java, properties, xml");
+
+            assertWithMessage("")
+                    .that(config.getChildren().length)
+                    .isEqualTo(1);
```
romani, line 735: Please do assert on Treewalker properties and that Treewalker has a child module.
We have them in config, we should recheck that all parts of config are loaded

romani (changes requested):
Last item

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
Kevin222004: Done

romani (approved):
Ok to merge

PR #13308, opened 2023-06-26
 Issue #13304: Rendering issues with website 

### LINKED ISSUE #13304: Rendering issues with website
from https://github.com/checkstyle/checkstyle/pull/13297#pullrequestreview-1497065219

Investigate rendering issue detected in this PR - https://github.com/checkstyle/checkstyle/pull/13297#issuecomment-1605250984

stoyanK7:
Resolves #13304 

This PR makes the menu stretch based on content width.

Some names are too large to fit in the navigation.
image

### DIFF
--- a/src/site/resources/css/site.css
+++ b/src/site/resources/css/site.css
@@ -98,9 +98,9 @@ p, td, table.bodyTable td, th, table.bodyTable th, li, pre, #breadcrumbs span, #
 }
 
 #leftColumn {
-  width: 270px;
+  min-width: 250px;
+  width: fit-content;
 }
-
 
 .releaseDate {
   position: absolute;
@@ -132,6 +132,10 @@ a[title="toTop"] {
   text-decoration: none;
   font-weight: 900;
   z-index: 10;
+}
+
+.poweredBy {
+  display: block;
 }
 
 @media screen and (max-width: 823px) {
--- a/src/site/resources/js/checkstyle.js
+++ b/src/site/resources/js/checkstyle.js
@@ -25,3 +25,21 @@ window.addEventListener("scroll", function () {
         scrollButton.style.display = "none";
     }
 });
+
+function setBodyColumnMargin() {
+    const leftColumn = document.querySelector('#leftColumn');
+    const bodyColumn = document.querySelector('#bodyColumn');
+
+    // If in mobile view use margin as defined in site.css
+    if (window.innerWidth < 823) {
+        bodyColumn.style.marginLeft = '1.5em';
+        return;
+    }
+
+    // Else calculate margin based on left column width
+    const leftColumnWidth = leftColumn.offsetWidth;
+    bodyColumn.style.marginLeft = `${leftColumnWidth + 27}px`;
+}
+
+window.addEventListener('load', setBodyColumnMargin);
+window.addEventListener('resize', setBodyColumnMargin);


stoyanK7:
Github, generate site

src/site/resources/css/site.css
```diff
@@ -134,6 +134,10 @@ a[title="toTop"] {
   z-index: 10;
 }
 
+.poweredBy {
+  display: block;
```
stoyanK7, line 138: > But first, why this?

Without `display: block;`, the `poweredBy` images stack next to each other and the menu becomes too wide. The idea of this style is to make them stack below each other. Ideally, they would be stacked next to and above each other like before. But to make that work, they need to be placed inside some container(`div`) and I couldn't find a way to place them inside a `div`. `site.xml` is pretty strict about that.
```html
...
    </ul>
      <a href="https://github.com/checkstyle/checkstyle" title="GitHub" class="poweredBy">
        <img class="poweredBy"  alt="GitHub" src="images/github_logo_social_coding_outlined.png"     />
      </a>
      <a href="https://twitter.com/checkstyle_java/" title="Twitter" class="poweredBy">
        <img class="poweredBy"  alt="Twitter" src="images/twitter_button.png"     />
      </a>
      <a href="https://stackoverflow.com/questions/tagged/checkstyle" title="Stackoverflow" class="poweredBy">
        <img class="poweredBy"  alt="Stackoverflow" src="images/stackoverflow.jpeg"     />
      </a>
      <a href="https://groups.google.com/forum/#!forum/checkstyle" title="GoogleGroups" class="poweredBy">
        <img class="poweredBy"  alt="GoogleGroups" src="images/groups.png"     />
      </a>
      <a href="https://www.ej-technologies.com/products/jprofiler/overview.html" title="JProfiler" class="poweredBy">
        <img class="poweredBy"  alt="JProfiler" src="https://www.ej-technologies.com/images/product_banners/jprofiler_medium.png"     />
      </a>
      </div>
...
```

Screenshot from 2023-06-26 12-06-23

src/site/resources/js/checkstyle.js
```diff
@@ -25,3 +25,21 @@ window.addEventListener("scroll", function () {
...
+        return;
+    }
+
+    // Else calculate margin based on left column width
+    const leftColumnWidth = leftColumn.offsetWidth;
+    bodyColumn.style.marginLeft = `${leftColumnWidth + 27}px`;
+}
+
+window.addEventListener('load', setBodyColumnMargin);
+window.addEventListener('resize', setBodyColumnMargin);
```
stoyanK7, line 45: > And why this? What was wrong with the old CSS handling?

Old CSS handling relied on a fixed `margin-left` to move the main content from the menu. Now that the width of the menu can change, the left margin needs to be calculated based on that. CSS cannot deal with such dynamic calculations to my knowledge.
https://github.com/checkstyle/checkstyle/blob/3d8ed42ab60c0399550b3e7d54ca9dd767583101/src/site/resources/css/site.css#L96-L102

src/site/resources/css/site.css
```diff
@@ -98,10 +98,10 @@ p, td, table.bodyTable td, th, table.bodyTable th, li, pre, #breadcrumbs span, #
 }
 
 #leftColumn {
-  width: 270px;
+  min-width: 270px;
```
stoyanK7, line 101: > Min width on my laptop is 250. If we are going to do this, we may as well decrease this.

I am fine with that.

stoyanK7:
@rnveach I answered your reviews from https://github.com/checkstyle/checkstyle/pull/13297#pullrequestreview-1497065219

romani:
From phone looks good.
But even in desktop mode all content is vertical, there is no left panel (but this is same as live version of website).
We need to play with PR version on desktop.

stoyanK7:
> From phone looks good.

Phone should be unaffected. We are targeting desktop here.

stoyanK7:
> But even in desktop mode all content is vertical, there is no left panel (but this is same as live version of website).
We need to play with PR version on desktop.

What do you mean?

romani:
I just stating that on phone still ok, we need desktop testing.

romani:
desktop rendering still a problem:
image

nrmancuso:
@romani what browser are you using? It looks fine for me in Google Chrome and Firefox on Linux.

romani:
I update my post, I might be on old browser.
If this is issue of old browser, I am ok to ignore old browser issue.

nrmancuso (approved):
Looks good for me, tested on:
1. Mobile (chrome for iOS)
2. Mobile (safari)
3. Desktop (Chrome on Fedora Linux)
4. Desktop (Firefox on Fedora Linux)
5. Desktop (Chrome on OSX)

src/site/resources/css/site.css
rnveach: Ok.

src/site/resources/js/checkstyle.js
rnveach: @romani @nrmancuso This will break the page when Javascript is disabled.
image

rnveach: @romani @nrmancuso This will break the page when Javascript is disabled.
image

romani: 20 years ago disabled JavaScript was normal, but today browser without JavaScript is not working application :).

If fix is easy, we can consider to support non JavaScript browser.

rnveach: Javascript disable is still normal today as a security function. https://noscript.net/ is still being actively developed and used, especially in the Tor browser.

romani: If we end up in problem to make it for secure mode of browser, I recommend to merge as is, it will work for 90% user, and create issue to experiment of only css solution.

I am not expert here, I can not lead on how to do this.

stoyanK7: > @stoyanK7 Something like https://stackoverflow.com/a/38150661/1016482 won't work?

I tried but couldn't make it work since we are trying to set a `margin`, not a `width`. I found https://stackoverflow.com/questions/49286748/css-set-the-margin-of-a-div-equal-to-another-divs-width but it uses predefined `width` which is not applicable if we want to stretch the menu based on the content inside.

The only viable non-javascript solution I can think of is using grid. Though we are dropping some browser compatibility for that - browser versions before 2015-2012: https://caniuse.com/css-grid

src/site/resources/css/site.css
stoyanK7: Done.

src/site/resources/js/checkstyle.js
romani: @rnveach , I think we need merge as is, and create new issue to improve, but we will need your experience in html to fix it.

PR #13317, opened 2023-06-26
Issue #13321: Kill mutation for SummaryJavaDocCheck-7

Kevin222004:
Issue #13321: Kill mutation for SummaryJavaDocCheck-7

-----

# Check :- 
https://checkstyle.org/checks/javadoc/summaryjavadoc.html#SummaryJavadoc

-----

# Mutation 
https://github.com/checkstyle/checkstyle/blob/08da65a9f63912571441fde44965b827c4977614/config/pitest-suppressions/pitest-javadoc-suppressions.xml#L925-L932

-----

# Explaination
this method is used by two method 

https://github.com/checkstyle/checkstyle/blob/08da65a9f63912571441fde44965b827c4977614/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java#L464-L475

Please guide if this is wrong. as per https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.html#CUSTOM_NAME whatever the tag will come it will always be true.

another and main reason the method is returning https://github.com/checkstyle/checkstyle/blob/08da65a9f63912571441fde44965b827c4977614/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java#L484 
the tag with given name then that task is already done by only https://github.com/checkstyle/checkstyle/blob/08da65a9f63912571441fde44965b827c4977614/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java#L493 if they don't match it is false.

-----

# Regression :- 

-----

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/61eaa69fe98585b6ee1882ca082362bc/raw/316ca02ce9fd290119d9bfc560f2b11e84a1b685/Summary.xml
Report label: Regression-2


Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

romani:
Pitest are failing

romani:
Not all tags are Custom named.
Please review https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.html#JAVADOC_INLINE_TAG

There are bunch of standard inline tags please make test with recognized inlined tag or show existing line in Input and explain why it doesn't help to kill survival

Kevin222004:
https://github.com/checkstyle/checkstyle/pull/13317#issuecomment-1615317660 Thanks for clarifying @romani
but the main reason to remove is 
https://github.com/checkstyle/checkstyle/blob/3142055d05641087c2b4b1775a39548ef02f2f41/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java#L486-L494

this method check 2 condition that the name of inline tag should be match to `child1.getText()` which is text of JavadocInline Tag so if they mismatch then it will be false.

Know this method is only called by 2 method which are 
https://github.com/checkstyle/checkstyle/blob/3142055d05641087c2b4b1775a39548ef02f2f41/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java#L464-L466
and 
https://github.com/checkstyle/checkstyle/blob/3142055d05641087c2b4b1775a39548ef02f2f41/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java#L474-L476
 
in which as a name parameter already we are passing the name as `@summary` and `@return` as form of Summary_text and Return_text 

so if this name will not match with the name of javadoc tag then method return false and in return statment both the condition are in && so if the removed statement is true still it will fail

romani:
@Kevin222004 , Please resolve conflict meanwhile

romani:
Yes, we check only predefined inline tags
Like https://github.com/checkstyle/checkstyle/blob/3142055d05641087c2b4b1775a39548ef02f2f41/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java#L262

, and all of them are Custom for us.
So we do not need this extra condition.

romani (approved):
Ok to merge

Vyom-Yadav (approved):
LGTM!

rdiachenko:
@Kevin222004 please resolve a conflict

rdiachenko (approved):
lgtm

PR #13325, opened 2023-06-27
minor: fix broken links

stoyanK7:
- Fixes broken links which were not found by linkcheck. I used `git grep` to find those
 - Removes 3 config files that I forgot to remove (I think those were removed but I rebased wrongly when there were lots of merge conflicts and they remained)


src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/RegexpSinglelineJavaCheck.java
```diff
@@ -32,7 +32,8 @@
  * </p>
  * <p>
  * This class is variation on
- * <a href="https://checkstyle.org/config_regexp.html#RegexpSingleline">RegexpSingleline</a>
+ * <a href="https://checkstyle.org/checks/regexp/regexpsingleline.html#RegexpSingleline">
```
romani, line 35: Please remove `#RegexpSingleline`

Same for others.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/RegexpSinglelineJavaCheck.java
stoyanK7: Done.

Issue #13328, opened 2023-06-29
Targeted Pitest Mutation in July 2023


Kevin222004:
extended from https://github.com/checkstyle/checkstyle/issues/13109 

------

# Pitest-Import Profile

**5) UnusedImportsCheck**
- [x] MemberVariableMutator (1)
       https://github.com/checkstyle/checkstyle/pull/13332

-----------------

# Pitest-api Profile 

**7) FullIdent**
- [x] NonVoidMethodCallMutator (1)
      https://github.com/checkstyle/checkstyle/pull/13179 

--------------------

# Pitest-PackageNamesLoader Profile

- [x] ArgumentPropagationMutator (3)
- [x] NonVoidMethodCallMutator (2)
- [x] EmptyObjectReturnValsMutator (1)
    all the mutations covered by 
    https://github.com/checkstyle/checkstyle/pull/13286

---------

Pitest-coding-1 

**14) EqualsAvoidNullCheck**
- [x] ArgumentPropagationMutator (2)
       https://github.com/checkstyle/checkstyle/pull/13127

**16) VariableDeclarationUsageDistanceCheck**
- [x] RemoveConditionalMutator_EQUAL_IF (1)
       https://github.com/checkstyle/checkstyle/pull/13120

------------

# Pitest-coding-2

**29) UnusedLocalVariableCheck$TypeDeclDesc**
- [x] NonVoidMethodCallMutator (1)
       https://github.com/checkstyle/checkstyle/pull/13331
       
**22) StringLiteralEqualityCheck**
- [x] NakedReceiverMutator (1)
       https://github.com/checkstyle/checkstyle/pull/13277

**26) OneStatementPerLineCheck**
- [x] MemberVariableMutator (2)
      https://github.com/checkstyle/checkstyle/pull/13281
      https://github.com/checkstyle/checkstyle/pull/13163 

**34) IllegalInstantiationCheck**
- [x] MemberVariableMutator (1)
      https://github.com/checkstyle/checkstyle/pull/13284

**27) HiddenFieldCheck$FieldFrame**
- [x] RemoveConditionalMutator_EQUAL_IF (1)

Vyom-Yadav:
@Kevin222004 Are you updating the issue body?

romani:
Issue is closed. There is separate issue for all items that were left to be done.

PR #13333, opened 2023-07-02
Issue #13328: Kill mutation ImportOrder

Kevin222004:
Issue #13328: Kill mutation ImportOrder

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/61eaa69fe98585b6ee1882ca082362bc/raw/316ca02ce9fd290119d9bfc560f2b11e84a1b685/Summary.xml
Diff Regression projects: https://gist.githubusercontent.com/Kevin222004/21e3934e85f802e2fbd48af06d122364/raw/604256badd733d8568064f371d55657c04b00dfd/test-projects-2.properties
Report label: new-Regression-2

PR #13334, opened 2023-07-02
Issue #13328: Kill mutation ImportOrderCheck  

Kevin222004:
Issue #13328: Kill mutation ImportOrderCheck  

--------------

# Check 
https://checkstyle.org/checks/imports/importorder.html#ImportOrder

------------

# Mutation 
https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/config/pitest-suppressions/pitest-imports-suppressions.xml#L57-L64

--------------

# Explaination 
begin tree always reset data. for `lastImportStatic = false` so we need to create two input somehow lastImportStatic become true while the sequentially second file will be come in process and that is easy we only have to set last import of first file as true.

lets start from visitToken
In visit token the parameter `isStatic` will check wether the current import is static or not.
https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L800-L810

To move further 
all the condition 
https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L814-L825

call `doVisitToken` know our target is to check that output get affected if remove `lastimportStatic` from beginTree 
https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L845

know in `doVisitToken` only 2 parameter is iportant for us `isStatic` and `previous` 
https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L815-L816
https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L818-L820

because they and the reason is they are only one right know which related with `lastimportstatic` and violation throwing 
and in the whole method previous is used only at 
https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L856-L858

know this condition is not big deal we can execute that and if we look closure on https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L923 this will execute when https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L925 which is by default true so that is not issue in that if never execute because in from visittoken condition.

know in else either https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L936 
needs to be true or in that either https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L940-L942 
this is need to be true

**case 1**
lets see the previous first we know that we need previos true and 
https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L814-L817

lets suppose we set option as above so previous import needs to be nonStatic and current ast need to be static so from here it is for sure that
we want previous as true `previous = true` know suppose option is set to be true so previous value is comming from visitToken so we need top set `isStatic = false` and `lastimportStatic = true` so in previous file lastimportstatic needs to be non static it means false then if it already same then no need to be reset so begin tree is redundant.

know lets suppose option is set to Bottom or under
https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L818-L821

in this case to set previous true we need to set `lastimportstatic = true` mean previous file import needs to be static and in current file ast needs to be non static it means `isStatic=fasle` and `isLastImport=true` in this case https://github.com/checkstyle/checkstyle/blob/b8915dab95432f57df7218f4f16ef7306e049e54/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L856-L858 this never will be equal so conditon will never execute

this is alsosimilar if both are static 

so no test cases looks possible to clear the state using input 

--------------------------------------

# Regression 

Report-1 :- https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/920122e_2023074505/reports/diff/index.html

Report-2 :- https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/920122e_2023090228/reports/diff/index.html

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/e3a00111b4bb7b8c054483466b88606d/raw/4895c88b1daf686479058c972bad0c27cd0818ef/IOC.xml
Report label: new-Regression-1

### DIFF
--- a/config/pitest-suppressions/pitest-imports-suppressions.xml
+++ b/config/pitest-suppressions/pitest-imports-suppressions.xml
@@ -54,14 +54,14 @@
 
 
 
-  <mutation unstable="false">
-    <sourceFile>ImportOrderCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck</mutatedClass>
-    <mutatedMethod>beginTree</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable lastImportStatic</description>
-    <lineContent>lastImportStatic = false;</lineContent>
-  </mutation>
+
+
+
+
+
+
+
+
 
   <mutation unstable="false">
     <sourceFile>ImportOrderCheck.java</sourceFile>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheckTest.java
@@ -24,12 +24,16 @@ import static com.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck.MS
 import static com.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck.MSG_SEPARATED_IN_GROUP;
 import static com.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck.MSG_SEPARATION;
 
+import java.io.File;
+import java.util.Optional;
+
 import org.antlr.v4.runtime.CommonToken;
 import org.junit.jupiter.api.Test;
 
 import com.puppycrawl.tools.checkstyle.AbstractModuleTestSupport;
 import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
 import com.puppycrawl.tools.checkstyle.DetailAstImpl;
+import com.puppycrawl.tools.checkstyle.JavaParser;
 import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
@@ -781,5 +785,32 @@ public class ImportOrderCheckTest extends AbstractModuleTestSupport {
         verifyWithInlineConfigParser(
                 getPath("InputImportOrderTestTrimInOption.java"),
                 expected);
+    }
+
+    /**
+     * Finding the appropriate input for testing the "lastImportStatic"
+     * field is challenging. Removing it from the reset process might
+     * create an opportunity for the module to enter an incorrect state,
+     * leading to hard-to-detect and unstable violations that will affect our users.
+     *
+     * @throws Exception when code tested throws exception
+     */
+    @Test
+    public void testClearState() throws Exception {
+        final ImportOrderCheck check = new ImportOrderCheck();
+        final DetailAST root = JavaParser.parseFile(
+                new File(getPath("InputImportOrderBeginTree.java")),
+                JavaParser.Options.WITHOUT_COMMENTS);
+        final Optional<DetailAST> staticImport = TestUtil.findTokenInAstByPredicate(root,
+            ast -> ast.getType() == TokenTypes.STATIC_IMPORT);
+
+        assertWithMessage("Ast should contain STATIC_IMPORT")
+                .that(staticImport.isPresent())
+                .isTrue();
+        assertWithMessage("State is not cleared on beginTree")
+                .that(TestUtil.isStatefulFieldClearedDuringBeginTree(check, staticImport.get(),
+                        "lastImportStatic", lastImportStatic -> !((boolean) lastImportStatic)))
+                .isTrue();
+
     }
 }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/importorder/InputImportOrderBeginTree.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/importorder/InputImportOrderBeginTree.java
@@ -1,0 +1,6 @@
+package com.puppycrawl.tools.checkstyle.checks.imports.importorder;
+
+import static java.io.IOException.*; // ok
+
+public class InputImportOrderBeginTree {
+}


Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

Vyom-Yadav (approved):
LGTM!

Vyom-Yadav:
@Kevin222004 Logic looks correct to me, but to be on the safer side, please generate a report with both changes in this PR and in https://github.com/checkstyle/checkstyle/pull/13333. You can do it in a separate PR and link it back here.

Kevin222004:
@Vyom-Yadav new report https://github.com/checkstyle/checkstyle/pull/13378

Report-1 :-  https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/43193d6_2023095732/reports/diff/index.html
Report-2  :- https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/43193d6_2023115512/reports/diff/index.html

rdiachenko (approved):
lgtm

played with this trying multiple inputs and combinations, didn't manage to find a failure case

romani:
This is great example of scary update :).
Right now, yes, maybe we can remove this cleanup.
But it gives opportunity for module to come to wrong state and eventually produce hard to catch unstable violations that will drive crazy our users.

What if we stay on safe side with clean up of all fields in class to default state ? I know we fighting a lot on such pitest in beginTree in other Checks, and before this I was ok, and we able to find coverage.

Proposal: make method `resetToDefaultIgnoredByPitest`  and put there only fields resets we afraid to remove. And we exclude this method from pitest to consider. We do javadoc over it to explain that we damage design a bit in favor of pitest. It doesn't not mean we can easily apply same pattern to other Modules. Allowance to be in such method will be only by very special allowance.

Let's vote, If all agree, we can do this.

rnveach (changes requested):
As I mentioned in discord, we have a special set of utils to kill mutations like this.

`isStatefulFieldClearedDuringBeginTree`
https://github.com/checkstyle/checkstyle/blob/b2df719ed76cba27e35ad402b204b7a4f90526b6/src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/TestUtil.java#L134
It looks like `visitToken` always set the field at https://github.com/checkstyle/checkstyle/blob/e1fc7ce702124f686d0660d8a991a6a03c3fdad1/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L830 .
All that is needed is to have the field change its value using the `isStatefulFieldClearedDuringBeginTree`.

rdiachenko:
I'd go with what @rnveach suggests instead of having pitest specific methods in the check's class. It looks clenar and avoids questions like why we have such tool related methods in the mian code. If we do introduce `resetToDefaultIgnoredByPitest` sort of stuff, it opens opportunities to add similar things for violations from other tools if we can't cover them, which will make main code even more obscure.

Kevin222004:
Ok I have updated the pr with the test case suggested at https://github.com/checkstyle/checkstyle/pull/13334#pullrequestreview-1540385100

src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheckTest.java
```diff
@@ -782,4 +786,23 @@ public void testTrimOption() throws Exception {
                 getPath("InputImportOrderTestTrimInOption.java"),
                 expected);
     }
+
+    @Test
+    public void testClearState() throws Exception {
```
romani, line 791: Please do summary of my comment from PR to explain reason why we allow pure unit test.
We have a rule that all pure unit test must have explanation comment.

romani (changes requested):
Item:

src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheckTest.java
romani: Please add this case to our issue that collect bad sude effects of pitest - extra code that does help to make code more robust to future changes that might miss extended testing.

Kevin222004: Done

romani (approved):
ok to merge

config/pitest-suppressions/pitest-imports-suppressions.xml
```diff
@@ -54,14 +54,14 @@
...
-    <lineContent>lastImportStatic = false;</lineContent>
-  </mutation>
+
+
+
+
+
+
+
+
```
rnveach, line 64: Why are we adding empty lines to this file?

Kevin222004: To avoid Conflicts

rnveach: Is the plan to remove them at some point if we can't remove the entire file?

Kevin222004: Yes, If the entire file is not removed then I will update this and remove it. so at the, it looks clean

But right know in this file 4 mutations are still left which is possible to cover

rnveach (approved):

Issue #13321, opened 2023-06-27
Resolve pitest survivals in modules - 2


Kevin222004:
Split from https://github.com/checkstyle/checkstyle/issues/13109
Part of https://github.com/checkstyle/checkstyle/issues/13089

------------------

**17) SummaryJavadocCheck**
- [x] ArgumentPropagationMutator (2)
           https://github.com/checkstyle/checkstyle/pull/13315
           https://github.com/checkstyle/checkstyle/pull/13330
- [x] NonVoidMethodCallMutator (4)
          https://github.com/checkstyle/checkstyle/pull/13313
          https://github.com/checkstyle/checkstyle/pull/13312
          https://github.com/checkstyle/checkstyle/pull/13323
          https://github.com/checkstyle/checkstyle/pull/13319
- [x]  RemoveConditionalMutator_EQUAL_ELSE (2)
          https://github.com/checkstyle/checkstyle/pull/13312
          https://github.com/checkstyle/checkstyle/pull/13313
- [x] RemoveConditionalMutator_EQUAL_IF (4)
          https://github.com/checkstyle/checkstyle/pull/13323
          https://github.com/checkstyle/checkstyle/pull/13319
          https://github.com/checkstyle/checkstyle/pull/13316
          https://github.com/checkstyle/checkstyle/pull/13317
- [x] NakedReceiverMutator (3)
         https://github.com/checkstyle/checkstyle/pull/13311
         https://github.com/checkstyle/checkstyle/pull/13323
         https://github.com/checkstyle/checkstyle/pull/13318
         
----------------------------------

**25) JavadocUtil**
- [x] RemoveConditionalMutator_EQUAL_ELSE (1)
         https://github.com/checkstyle/checkstyle/pull/13379

--------------------

**26) XpathUtil**
- [x] NonVoidMethodCallMutator (2)
        https://github.com/checkstyle/checkstyle/pull/13379 (Cover 2 mutation)

----------------------

**28) AnnotationUtil**
- [x] NonVoidMethodCallMutator (2)
       https://github.com/checkstyle/checkstyle/pull/13379 (cover 2 Mutation) 
- [x] RemoveConditionalMutator_EQUAL_IF (2)
       https://github.com/checkstyle/checkstyle/pull/13379 (cover 2 Mutation)

----------------------

**29) ScopeUtil**
- [x] NonVoidMethodCallMutator (2)
[...]

Issue #13038, opened 2023-04-30
VariableDeclarationUsageDistanceCheck doesn't handle method definition properly


jxr98:
I have read check documentation: https://checkstyle.org/config_coding.html#VariableDeclarationUsageDistance
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below.

```bash
/var/tmp $ javac Test1.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name = "Checker">
    <property name="charset" value="UTF-8"/>

    <!-- do not change severity to 'error', as that will hide errors caused by exceptions -->
    <property name="severity" value="warning"/>

    <!-- haltOnException is required for exception fixes and reporting of all exceptions -->
    <property name="haltOnException" value="false"/>

    <!-- BeforeExecutionFileFilters is required for sources of java9 -->
    <module name="BeforeExecutionExclusionFileFilter">
        <property name="fileNamePattern" value="module\-info\.java$" />
    </module>

    <module name="TreeWalker">
         <!-- Example of checkstyle Check usage -->
         <!-- PLEASE CHANGE IT TO CHECK YOU ARE TESTING !!!! -->
              
        <module name="VariableDeclarationUsageDistance">
          <property name="id" value="Custom"/>
          <property name="validateBetweenScopes" value="true"/>
          <property name="allowedDistance" value="1"/>
        </module> 
    </module>
    <!-- Example of filter -->
    <!--
    <module name="SeverityMatchFilter">
        <property name="severity" value="warning"/>
        <property name="acceptOnMatch" value="false"/>
    </module>
    -->

    <!-- as we run on regression even on non-compiled files we need to skip exceptions on them -->
    <module name="SuppressionSingleFilter">
      <property name="message" value="Exception occurred while parsing"/>
      <property name="checks" value="Checker"/>
    </module>

/var/tmp $ cat Test1.java

import java.util.Iterator;

public class Test1 {
    public void method(){
        int a=1;
        class GeneralLogic {
            public <T> Iterator<T> method(T[] b) throws Exception{
                if(a>0){
                    System.out.println(b.length);
                }
                return null;
            }
        }
    }
}


[...]

PR #13366, opened 2023-07-06
Issue #13321: Kill mutation for TranslationCheck-1

Kevin222004:
Issue #13321: Kill mutation for TranslationCheck-1

---------

# Check
https://checkstyle.org/checks/misc/translation.html#Translation

---------

# Mutation 
https://github.com/checkstyle/checkstyle/blob/2be5780673b9ba5397526aadab981be68f874172/config/pitest-suppressions/pitest-misc-suppressions.xml#L174-L190

---------

# Explaination 
To Kill the mutation unmodifiableCollectionUtil I have used avoid call to for out class created to wrap that method in util by default avoidcall to is applied to https://pitest.org/quickstart/maven/#:~:text=If%20a%20list%20is%20not%20explicitly%20supplied%20then%20PIT%20will%20default%20to%20a%20list%20of%20common%20logging%20packages%20as%20follows 
so add log method to avoid the error at https://github.com/checkstyle/checkstyle/blob/2be5780673b9ba5397526aadab981be68f874172/src/main/java/com/puppycrawl/tools/checkstyle/checks/TranslationCheck.java#L672

https://github.com/checkstyle/checkstyle/blob/0831953d82df93e5ebe7f178c2f2b904af8b8f3b/src/main/java/com/puppycrawl/tools/checkstyle/api/Violation.java#L327

-------

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/TranslationCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/TranslationCheck.java
@@ -24,7 +24,6 @@ import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.Locale;
 import java.util.Map;
@@ -51,6 +50,7 @@ import com.puppycrawl.tools.checkstyle.api.FileText;
 import com.puppycrawl.tools.checkstyle.api.MessageDispatcher;
 import com.puppycrawl.tools.checkstyle.api.Violation;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
+import com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil;
 
 /**
  * <p>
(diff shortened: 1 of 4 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/TranslationCheck.java
```diff
@@ -664,7 +664,7 @@ private void logException(Exception exception, File file) {
                 Definitions.CHECKSTYLE_BUNDLE,
                 key,
                 args,
-                getId(),
+                null,
```
romani, line 667: This will mean that:
If used define id for module - https://checkstyle.org/config.html#Id
It will not be printed to user in report instead of module name, so just module name will be used. If user has two instances of same module type he will not know what module instance caused it.

Example of `id` printing instead  of module name in CLI output - https://github.com/checkstyle/checkstyle/issues/13038 

We can not cover such mutation by our Inputs with embedded config. But we should create test method that validate violation `id` with config defined in java code and assert on violation object field or checking content of some test audit listener https://checkstyle.org/config.html#Audit_Listeners we should already have some tests for this.

romani (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/TranslationCheck.java
Kevin222004: ok i got it Thanks.

I found this type of testing at https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/AuditEventDefaultFormatterTest.java

i have update an exisiting test is it fine

romani (approved):
Ok to merge

romani:
We just extended test, I am merging

PR #13381, opened 2023-07-08
Issue #13321: Kill mutation for RegexpCheck-2

Kevin222004:
Issue #13321: Kill mutation for RegexpCheck-2

----------

# check
https://checkstyle.org/checks/regexp/regexpsinglelinejava.html#RegexpSinglelineJava
https://checkstyle.org/checks/regexp/regexpsingleline.html#RegexpSingleline

---------

# Mutation 
https://github.com/checkstyle/checkstyle/blob/4b7a2dfd9d98e8fdd59a220e45e365aa6a6507b3/config/pitest-suppressions/pitest-regexp-suppressions.xml#L66-L73
https://github.com/checkstyle/checkstyle/blob/4b7a2dfd9d98e8fdd59a220e45e365aa6a6507b3/config/pitest-suppressions/pitest-regexp-suppressions.xml#L84-L91

### DIFF
--- a/config/pitest-suppressions/pitest-regexp-suppressions.xml
+++ b/config/pitest-suppressions/pitest-regexp-suppressions.xml
@@ -53,22 +53,4 @@
     <description>Removed assignment to member variable duplicateLimit</description>
     <lineContent>this.duplicateLimit = duplicateLimit;</lineContent>
   </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>RegexpSinglelineCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.regexp.RegexpSinglelineCheck</mutatedClass>
-    <mutatedMethod>beginProcessing</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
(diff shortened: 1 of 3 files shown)


Kevin222004:
@romani can you give some guidance here 
I tried to find it but i don't understand what does this do exactly

romani:
https://github.com/checkstyle/checkstyle/blob/4969acc6b75906221c0f5ea5b30d5e1b451ed13b/src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/DetectorOptions.java#L264

https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#compile-java.lang.String-int-

Zero is default, https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/share/classes/java/util/regex/Pattern.java#L1068 we do not need to  explicitly set it.

This is bad side of pitest fanatizm, code is more explicit and doesn't require user to know or remember default implementations.

I am still ok to merge it, as it is in Builder pattern.

romani (approved):
Ok to merge

rdiachenko (approved):
lgtm

Vyom-Yadav (approved):
LGTM!

PR #13397, opened 2023-07-14
Pull #13397: update link check to only check internal links in PRs

nrmancuso:
We should only check for internal links in PRs.

Linkcheck run should pass in this PR even though we have a bad external link (stickler-ci) in master: https://circleci.com/gh/checkstyle/checkstyle/335960

Link to failure in master: https://github.com/checkstyle/checkstyle/actions/runs/5546064017/jobs/10125837894

----------------
Proof that `--check-external` works: https://app.circleci.com/pipelines/github/checkstyle/checkstyle/19614/workflows/2475ba7b-ee8e-4d65-96b3-fbc06dd143ab/jobs/336405

Proof that default works (internal only): https://circleci.com/gh/checkstyle/checkstyle/336829

### DIFF
--- a/.ci/run-link-check-plugin.sh
+++ b/.ci/run-link-check-plugin.sh
@@ -11,9 +11,19 @@ mvn -e --no-transfer-progress clean site -Dcheckstyle.ant.skip=true -DskipTests 
    -Dpmd.skip=true -Dspotbugs.skip=true -Djacoco.skip=true -Dcheckstyle.skip=true
 mkdir -p .ci-temp
 
+OPTION=$1
+
 echo "------------ grep of linkcheck.html--BEGIN"
-grep -E "doesn't exist|externalLink" target/site/linkcheck.html | grep -v 'Read timed out' \
-  | sed 's/<\/table><\/td><\/tr>//g' | sort > .ci-temp/linkcheck-errors-sorted.txt
+LINKCHECK_ERRORS=$(grep -E "doesn't exist|externalLink" target/site/linkcheck.html \
+  | grep -v 'Read timed out' | sed 's/<\/table><\/td><\/tr>//g' || true)
+
+if [[ $OPTION == "--skip-external" ]]; then
+  echo "Checking internal (checkstyle website) links only."
+  echo "$LINKCHECK_ERRORS" | grep -v 'externalLink' | sort > .ci-temp/linkcheck-errors-sorted.txt
+else
+  echo "Checking internal (checkstyle website) and external links."
+  echo "$LINKCHECK_ERRORS" | sort > .ci-temp/linkcheck-errors-sorted.txt
+fi
 
 sort config/linkcheck-suppressions.txt | sed 's/<\/table><\/td><\/tr>//g' \
   > .ci-temp/linkcheck-suppressions-sorted.txt
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -221,7 +221,7 @@ workflows:
       - validate-with-maven-script:
           name: "linkcheck-plugin"
           image-name: "cimg/openjdk:11.0.19"
-          command: "./.ci/run-link-check-plugin.sh"
+          command: "./.ci/run-link-check-plugin.sh --skip-external"
 
   idea:
     jobs:


nrmancuso:
Need to test bad internal link

romani:
I think it will be better to make argument for execution to validate external or not.
And make it activated in non pull requests.

It will allow us to do PR from branch of main repo to test it. We almost never to PRs from branches of main repo.

.circleci/config.yml
```diff
@@ -222,7 +222,10 @@ workflows:
           name: "linkcheck-plugin"
           image-name: "cimg/openjdk:11.0.19"
           command: "./.ci/run-link-check-plugin.sh"
-
+      - validate-with-maven-script:
+          name: "linkcheck-plugin-EXTERNAL"
+          image-name: "cimg/openjdk:11.0.19"
+          command: "./.ci/run-link-check-plugin.sh --check-external"
```
romani, line 228: We should not do this in Circle CI.

GitHub Action is master push based, it run full validation of links, good. But circleci ok to skip checking external. It will be stable if we skip doing it.

Chance that we leak in PR external link that doesn't work is low.

Ok to keep only to test that it works as separate commit for removal before merge

romani (changes requested):
Items

.circleci/config.yml
romani: Ok this is what is done already.

nrmancuso:
Link to test for internal links: https://circleci.com/gh/checkstyle/checkstyle/336829

nrmancuso:
>If you change code, please use reference to pull number

Done

.circleci/config.yml
romani: I keep it unresolved to show how it was tested

romani (approved):
Ok to merge

.github/workflows/run-link-check.yml
```diff
@@ -30,4 +30,4 @@ jobs:
           key: checkstyle-maven-cache-${{ hashFiles('**/pom.xml') }}
 
       - name: Check links
-        run: ./.ci/run-link-check-plugin.sh
+        run: "./.ci/run-link-check-plugin.sh --check-external"
```
rnveach, line 33: > We should only check for internal links in PRs.

Isn't this checking both internal and external?

nrmancuso: Yes, that is correct. We have some redundancy here between the github action and circleci.

romani: There might be some confusion on name of argument, is it what in addition or do-only.

If there better name please share how to name it.

rnveach: I am confused then, but not on argument name.

https://github.com/search?q=repo%3Acheckstyle%2Fcheckstyle%20%22run-link-check-plugin%22&type=code
Why does GH get lucky and tests both but circleci doesn't?

nrmancuso: GitHub action runs in master on both internal and external, for PRs we are only checking internal, for stability reasons.

rnveach: Should we document that, especially on circle?

romani: Code should be self explanatory.
And now I see a way to improve.
Action should do: `./.ci/run-link-check-plugin.sh`
Circleci: `./.ci/run-link-check-plugin.sh --skip-external`

nrmancuso: Done

.ci/run-link-check-plugin.sh
```diff
@@ -11,9 +11,19 @@ mvn -e --no-transfer-progress clean site -Dcheckstyle.ant.skip=true -DskipTests
...
 
+SKIP_EXTERNAL_LINKS=$1
+
 echo "------------ grep of linkcheck.html--BEGIN"
-grep -E "doesn't exist|externalLink" target/site/linkcheck.html | grep -v 'Read timed out' \
-  | sed 's/<\/table><\/td><\/tr>//g' | sort > .ci-temp/linkcheck-errors-sorted.txt
+LINKCHECK_ERRORS=$(grep -E "doesn't exist|externalLink" target/site/linkcheck.html \
+  | grep -v 'Read timed out' | sed 's/<\/table><\/td><\/tr>//g' || true)
+
+if [[ $SKIP_EXTERNAL_LINKS == "--skip-external" ]]; then
```
romani, line 20: This reads like nonsense.
Please rename SKIP_EXTERNAL_LINKS to PARAMETER

romani (changes requested):
Last:

.ci/run-link-check-plugin.sh
nrmancuso: God point, renamed to OPTION, as parameter would imply some possibly changing argument.

romani (approved):
Ok to merge

rnveach (approved):

PR #13407, opened 2023-07-18
Issue #13399: Implement module properties macro

### LINKED ISSUE #13399: Implement module properties macro
After https://github.com/checkstyle/checkstyle/pull/13357 is merged, we will have xdoc templates that use macros for examples. We can create a macro for the properties table that can look like:
```xml
    <macro name="properties">
      <param name="modulePath"
             value="src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/AbstractClassNameCheck.java"/>
    </macro>
```

When the generation of xdocs is run, the macro should produce the following result by crawling the java module
https://github.com/checkstyle/checkstyle/blob/41572dac51680bac020cac3dfc3d866320b38e73/src/xdocs/checks/naming/abstractclassname.xml#L24-L63

stoyanK7:
Resolves #13399 

Creates `PropertiesMacro` which is responsible for creating the table with 5 properties - name, description, type, default value, since. Lots of code has been taken from `XdocsPagesTest` as it already does what we want to.

The way the macro works is:
1. Call `ClassAndPropertiesSettersJavadocScraper` to scrape all the setter javadocs of parent classes - for example `AbstractNameCheck` which holds properties such as `applyToPublic`, `applyToProtected`. Properties that are not specified in checks classes but are inherited. This step is done only once when the macro is executed for the first time.
2. Process the module given to the macro - gather the class and setters javadocs from the file
3. Write the header row - name, description, type, default value, since
4. Get a set of the properties via the class property descriptors
5. Filter this set of properties - remove ones that are not supposed to be documented and add ones that are supposed to be - `tokens`, `fileExtensions`, etc..
6. Iterate trough this set of properties and write the 5 columns
    - Property name is coming from the set
    - Description is scraped off the setter javadoc. A description is any text from the beginning of the Javadoc until we meet a single asterisk on a line. For example:
    ```java
    /**
     * This over here is a setter description. The alone asterisk below is what says the description is over
     *   <---- end of description
     * @param ..l.
     */
    ```

    - Type comes from the methods taken from `XdocsPagesTest`
    - Same for default value - from `XdocsPagesTest`
    - Since version comes from one of 3 places(in order)
       1. Exception list - This is a list where properties are inherited from a parent but the property was introduced in a version after the creation of the check
       2. If property is inerited from parent and not part of module file - use `@since xxx` from class javadoc
       3. Else use the `@since xxx` version from property setter Javadoc


src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMetadataScraper.java
```diff
@@ -0,0 +1,96 @@
...
+
+import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.DetailNode;
+import com.puppycrawl.tools.checkstyle.api.JavadocTokenTypes;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheck;
+import com.puppycrawl.tools.checkstyle.meta.ModulePropertyDetails;
+import com.puppycrawl.tools.checkstyle.utils.BlockCommentPosition;
+import org.apache.tools.ant.taskdefs.Java;
```
stoyanK7, line 30: .

nrmancuso:
@stoyanK7 let's get to reviewable state

romani:
@stoyanK7 , please keep fixing CI failures - https://ci.appveyor.com/project/Checkstyle/checkstyle/builds/47874621/job/t52fqlfkkwhbsbsm#L310 to be ready for merge at any time.

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndSettersJavadocScraper.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    /**
+     * Set the module javadoc.
+     *
+     * @param moduleJavadoc the module javadoc.
+     */
+    public static void setModuleJavadoc(DetailNode moduleJavadoc) {
+        ClassAndSettersJavadocScraper.moduleJavadoc = moduleJavadoc;
+    }
```
nrmancuso, line 67: How is this used?

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndSettersJavadocScraper.java
```diff
@@ -0,0 +1,155 @@
...
+    private static final Map<String, DetailNode> MODULE_PROPERTY_SETTER_JAVADOCS = new TreeMap<>();
+
+    /**
+     * Set the module name to be scraped.
+     *
+     * @param moduleName the module name.
+     */
+    public static void setModuleName(String moduleName) {
+        ClassAndSettersJavadocScraper.moduleName = moduleName;
+    }
```
nrmancuso, line 58: Can we just call this `init/initialize` and handle clearing static context and all initialization here to not put this on the caller?

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndSettersJavadocScraper.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Checks if the given method is a method of the module being scraped. Traverses
+     * parent nodes until it finds the class definition and checks if the class name
+     * is the same as the module name.
+     *
+     * @param methodDef the method definition.
+     * @return true if the method is a method of the given module, false otherwise.
+     */
+    private static boolean isMethodOfScrapedModule(DetailAST methodDef) {
```
nrmancuso, line 126: Help me to understand what the conditions are where we can have a method that is NOT a method of the module being scraped.

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,861 @@
...
+
+            value = newArray;
+        }
+
+        if (value != null && Array.getLength(value) > 0) {
+            final String[] newArray = new String[Array.getLength(value)];
+
+            for (int index = 0; index < newArray.length; index++) {
+                newArray[index] = ((Pattern) Array.get(value, index)).pattern();
+            }
```
romani, line 618: transformation/map of one array to another array ....
can we use some fancy streams/map ?

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,861 @@
...
+            else {
+                final Object[] array = (Object[]) value;
+                valuesStream = Arrays.stream(array);
+            }
+            result = valuesStream
+                .map(String.class::cast)
+                .sorted()
+                .collect(Collectors.joining(COMMA_SPACE));
+            valuesStream.close();
```
romani, line 670: use try-with-resourcse structure

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,861 @@
...
+     * Gets the class of the given field.
+     *
+     * @param field the field to get the class of.
+     * @return the class of the field.
+     * @throws MacroExecutionException if an error occurs during getting the class.
+     */
+    private static Class<?> getFieldClass(Field field) throws MacroExecutionException {
+        Class<?> result = null;
+
+        if (field != null) {
```
romani, line 724: hmm ..... in what cases field is null ?
soundlike this condition shuld be out of this methods, as if upper level do bad field calculation it should handle it itself, `getFiledClass(null)` sounds weird.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,861 @@
...
+    /**
+     * Fixes the captured properties. This method removes properties that are not documented and
+     * adds properties that have to be documented but are not captured.
+     *
+     * @param instance the instance of the module.
+     * @param clss the class of the module.
+     * @param properties the properties of the module.
+     */
+    private static void fixCapturedProperties(
+            Object instance, Class<?> clss, Set<String> properties) {
```
romani, line 806: can we split this method to
`removeNotProperties`
`addNonExplicitProperties` ?

it will be easier to understand, as for now it is "fix" ==== "do something unclear"

romani (changes requested):
items:

src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
```diff
@@ -516,6 +516,10 @@ public void testAllCheckstyleModulesInCheckstyleConfig() throws Exception {
     public void testAllCheckstyleChecksHaveMessage() throws Exception {
         for (Class<?> module : CheckUtil.getCheckstyleChecks()) {
             final String name = module.getSimpleName();
+            // Module is not a check
+            if ("ClassAndSettersJavadocScraper".equals(name)) {
+                continue;
+            }
```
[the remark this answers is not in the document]
rnveach: IMO, we really shouldn't have to keep overriding our internal modules to be skipped each time. At the least, we should have some uniformity in this area.

Examples:
https://github.com/checkstyle/checkstyle/blob/0dba32b6c8c29977d74af675078501fa8052c10f/src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java#L186
https://github.com/checkstyle/checkstyle/blob/0dba32b6c8c29977d74af675078501fa8052c10f/src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/CheckUtil.java#L195

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,1057 @@
...
+        }
+        else if (fieldClass == URI.class || fieldClass == String.class) {
+            if (value != null) {
+                result = '"' + value.toString() + '"';
+            }
+        }
+        else if (fieldClass == Pattern.class) {
+            if (value != null) {
+                result = '"' + value.toString().replace("\n", "\\n").replace("\t", "\\t")
+                        .replace("\r", "\\r").replace("\f", "\\f") + '"';
```
nrmancuso, line 758: Please share an example of a check that requires this replacement. Also, is it safe to just always assume that we need to do `\` -> `\\`?

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,1057 @@
...
+
+    /**
+     * Fixes the captured properties. This method removes properties that are not documented and
+     * adds properties that have to be documented but are not captured.
+     *
+     * @param instance the instance of the module.
+     * @param clss the class of the module.
+     * @param properties the properties of the module.
+     */
+    private static void fixCapturedProperties(Object instance,
```
nrmancuso, line 1001: Please return new collection, we should always avoid modifying collections and elements of collections.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,1057 @@
...
+     * Extract the description of the setter from the Javadoc.
+     *
+     * @param javadoc the Javadoc to extract the description from.
+     * @param moduleName the name of the module.
+     * @return the description of the setter.
+     * @throws MacroExecutionException if an error occurs during extraction.
+     */
+    private static String getDescription(DetailNode javadoc, String moduleName)
+            throws MacroExecutionException {
+        final Stack<DetailNode> stack = new Stack<>();
```
nrmancuso, line 566: Stack -> Deque

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,1057 @@
...
+    private static void writePropertyDescriptionCell(Sink sink, String propertyName,
+                                                     String moduleName, DetailNode javadoc)
+            throws MacroExecutionException {
+        sink.rawText(INDENT_LEVEL_14);
+        sink.tableCell();
+        if ("tokens".equals(propertyName)) {
+            sink.text("tokens to check");
+        }
+        else {
+            System.out.println(propertyName);
```
nrmancuso, line 354: Leftover from debugging?

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,1057 @@
...
+     * @param moduleName the name of the module.
+     * @throws MacroExecutionException if an error occurs during writing.
+     */
+    private static void writeTablePropertiesRows(Sink sink, String moduleName)
+            throws MacroExecutionException {
+        final Map<String, DetailNode> unmodifiableJavadocs =
+                ClassAndPropertiesSettersJavadocScraper.getJavadocs();
+        final Map<String, DetailNode> javadocs = new LinkedHashMap<>(unmodifiableJavadocs);
+        final Object instance = SiteUtil.getModuleInstance(moduleName);
+        final Class<?> clss = instance.getClass();
```
nrmancuso, line 282: Can we do `final AbstractCheck checkInstance = (AbstractCheck) SiteUtil.getModuleInstance(moduleName);` and just pass this to other methods to reduce the number of parameters and casts?

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,1057 @@
...
+            if (isInHrefAttribute && node.getType() == JavadocTokenTypes.ATTR_VALUE) {
+                final String href = node.getText();
+                // Do not touch external links
+                if (!href.contains("https://checkstyle.org/")) {
+                    continue;
+                }
+
+                final Path templatePath = SiteUtil.getTemplatePath(moduleName.replace("Check", ""));
+
+                String finalHref = href.replace("https://checkstyle.org/", "");
```
nrmancuso, line 597: Should we do a StringBuilder here?

nrmancuso (changes requested):
Few more:

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndSettersJavadocScraper.java
stoyanK7: Inner class.

I updated Javadoc to make it more obvious what we are avoiding here.

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndSettersJavadocScraper.java
stoyanK7: Removed.

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndSettersJavadocScraper.java
```diff
@@ -0,0 +1,155 @@
...
+ * given checkstyle module.
+ */
+@FileStatefulCheck
+public class ClassAndSettersJavadocScraper extends AbstractJavadocCheck {
+
+    /** Name of the module being scraped. */
+    private static String moduleName;
+
+    /** The module javadoc. */
+    private static DetailNode moduleJavadoc;
```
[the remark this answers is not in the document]
stoyanK7: Removed.

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndSettersJavadocScraper.java
```diff
@@ -0,0 +1,155 @@
...
+    public static void clearModulePropertySetterJavadocs() {
+        MODULE_PROPERTY_SETTER_JAVADOCS.clear();
+    }
+
+    /**
+     * Get the module property setter javadocs.
+     *
+     * @return the module property setter javadocs map.
+     */
+    public static Map<String, DetailNode> getModulePropertySetterJavadocs() {
```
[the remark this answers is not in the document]
stoyanK7: Renamed to just `JAVADOCS` and `getJavadocs`

config/checkstyle-non-main-files-checks.xml
```diff
@@ -57,7 +57,7 @@
     <property name="id" value="lineLength"/>
     <!-- catch lines above 100 symbols -->
     <property name="format"
-             value="^(?!(.*value=.*Example\d+|.*href=|.*http(s)?:|import |(.* )?package |.* files=|.*\.dtd| \* \{@code| \* com\.)).{101,}$"/>
+             value="^(?!(.*value=.*|.*href=|.*http(s)?:|import |(.* )?package |.* files=|.*\.dtd| \* \{@code| \* com\.)).{101,}$"/>
```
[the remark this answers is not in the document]
stoyanK7: Because the macro value is too long
```xml
          <macro name="properties">
            <param name="modulePath"
                   value="src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/AnnotationLocationCheck.java"/>
          </macro>
```

Initially, the exemption was added for examples but now we need it here too.

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -145,12 +145,12 @@ else if (module == RegexpSinglelineCheck.class
      * @return the value of the field.
      * @throws MacroExecutionException if the value could not be retrieved.
      */
-    private static String getFieldValue(Field field, Object instance)
+    public static Object getFieldValue(Field field, Object instance)
```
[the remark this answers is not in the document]
stoyanK7: This is functionality taken from `XdocsPagesTest`. I initially took it for violation messages and narrowed it down to return a `String` but now that I need it for properties, I made it public and return an `Object`. Better than duplicating code.

src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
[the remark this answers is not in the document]
stoyanK7: Added to `isFromAllowedPackages`

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: Yes. Removed.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: This method and others I have taken as-is from `XdocsPagesTest` - `getProperties`, `fixCapturedProperties`, `getType`, `getDefaultValue`, `getPatternArrayPropertyValue`, `getStringArrayPropertyValue`, `getIntArrayPropertyValue`, `getFieldClass`, `subtractTokens`.

To answer your question - running the code from `XdocsPagesTest`, `JavadocStyle.endOfSentenceFormat` pops up
Screenshot from 2023-08-29 14-15-51


> Also, is it safe to just always assume that we need to do \ -> \\

I think so. Makes `value` readable:

Value before
```
([.?!][ 	

<])|([.?!]$)
```

Value after
```
([.?!][ \t\n\r\f<])|([.?!]$)
```

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,861 @@
...
+            stream = Arrays.stream((int[]) value);
+        }
+        String result = stream
+                .mapToObj(TokenUtil::getTokenName)
+                .sorted()
+                .collect(Collectors.joining(COMMA_SPACE));
+        if (result.isEmpty()) {
+            result = CURLY_BRACKETS;
+        }
+        stream.close();
```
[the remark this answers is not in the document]
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,861 @@
...
+
+    /**
+     * Gets the name of the bean property's default value for the Pattern array class.
+     *
+     * @param fieldValue The bean property's value
+     * @return String form of property's default value
+     */
+    private static String getPatternArrayPropertyValue(Object fieldValue) {
+        Object value = fieldValue;
+        String result;
```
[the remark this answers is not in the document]
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,861 @@
...
+        if (value instanceof Collection) {
+            final Collection<?> collection = (Collection<?>) value;
+            final Pattern[] newArray = new Pattern[collection.size()];
+            final Iterator<?> iterator = collection.iterator();
+            int index = 0;
+
+            while (iterator.hasNext()) {
+                final Object next = iterator.next();
+                newArray[index] = (Pattern) next;
+                index++;
```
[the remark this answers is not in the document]
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: Done. Simplified.

stoyanK7:
Resolving comments right now, will move to ready for review when done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: `HeaderCheck.header` for example.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: It's not certain that it's always an `AbstractCheck`. For example filters.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
romani: https://checkstyle.org/checks/header/header.html#Properties

Field can not be null.
Value of field can be null. Do we have issue with method name?

stoyanK7:
Moving back to ready for review even though CI will fail. Moved quite a bit of stuff around.

src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/AnnotationLocationCheck.java
```diff
@@ -65,16 +65,16 @@
...
- * Default value is {@code true}.
+ * Default value is {@code false}.
  * </li>
  * <li>
- * Property {@code allowSamelineParameterizedAnnotation} - Allow one and only parameterized
+ * Property {@code allowSamelineSingleParameterlessAnnotation} - Allow single parameterless
  * annotation to be located on the same line as target element.
  * Type is {@code boolean}.
- * Default value is {@code false}.
+ * Default value is {@code true}.
```
stoyanK7, line 77: The reason why the metadata file and the javadoc are changed is because of the order of the properties. Properties didn't have any order until now(or do they?) so it's impossible to match their order. The order now follows the order of the setters in the module class.

rnveach: XDocs Order: https://github.com/checkstyle/checkstyle/blob/54649c930bfd0fc09f59fb5afdd8c6df82c680a5/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1551

XdocsJavaDocsTest took XDoc as it is  (order and all) and made sure it was in the correct format as the Javadoc.

JavadocMetadataScraper just took what was in Javadoc and placed it into meta file.

It seems to me, javadoc should be ordered just like xdoc, which is alphabetical order by TreeSet.

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/AbstractAccessControlNameCheck.java
```diff
@@ -113,7 +113,7 @@ private static boolean isPublic(DetailAST modifiers) {
     }
 
     /**
-     * Sets whether we should apply the check to public members.
+     * Setter to control whether we should apply the check to public members.
```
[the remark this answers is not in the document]
rnveach: This was a limitation since Checkstyle does not support multi-file.

src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/AnnotationLocationCheck.java
rnveach: Looks like your issue is at:
https://github.com/checkstyle/checkstyle/pull/13407/files#diff-582038065de9c242c07931401cfa262fcfd745638bf7c42736dba5793a3294cdR401

You switch over from the TreeSet to a LinkedHashSet. You shouldn't use hash sorting as different VMs can possibly create different hashes. Even if we were guaranteed hashes, we want alpha ordering here.

nrmancuso: > I recommend creating a new issue for this and this being done first. It won't be easy to review if there is a bunch of property changes like this.

I second this.

romani: Can we implement alphabet ordering now, but do not migrate numerous checks for now. So we can do:
 1)migration to macroses in smaller batches 

2)make massive updates for ordering first and then separate PRs wave to use macroses.

We will see later on what is less on effort.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: Simplified.

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,493 @@
...
+    private static void processModule(String moduleName, File moduleFile)
+            throws MacroExecutionException {
+        if (!moduleFile.isFile()) {
+            final String message = String.format(Locale.ROOT,
+                    "File %s is not a file. Please check the 'modulePath' property.", moduleFile);
+            throw new MacroExecutionException(message);
+        }
+        ClassAndPropertiesSettersJavadocScraper.clear();
+        ClassAndPropertiesSettersJavadocScraper.setModuleName(moduleName);
+        final Checker checker = new Checker();
```
romani, line 153: please recheck it that is easy to make ClassAndPropertiesSettersJavadocScraper be created for each processModule as new class, it will be a way easier to manage it in future refactoring.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,493 @@
...
+        final Map<String, DetailNode> javadocs = new LinkedHashMap<>(unmodifiableJavadocs);
+        final Object instance = SiteUtil.getModuleInstance(currentModuleName);
+        final Class<?> clss = instance.getClass();
+
+        final Set<String> properties = SiteUtil.getProperties(clss);
+        final Set<String> allProperties = SiteUtil
+                .fixCapturedProperties(instance, clss, properties);
+
+        for (String property : allProperties) {
+            javadocs.putIfAbsent(property, PARENT_MODULES_PROPERTIES_JAVADOCS.get(property));
```
romani, line 247: please move this putIfAbsent to block that calculates allProperties.

we need keep code in phases: data preparation, execution of writes.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,493 @@
...
+            throws MacroExecutionException {
+        final Map<String, DetailNode> unmodifiableJavadocs =
+                ClassAndPropertiesSettersJavadocScraper.getJavadocs();
+        final Map<String, DetailNode> javadocs = new LinkedHashMap<>(unmodifiableJavadocs);
+        final Object instance = SiteUtil.getModuleInstance(currentModuleName);
+        final Class<?> clss = instance.getClass();
+
+        final Set<String> properties = SiteUtil.getProperties(clss);
+        final Set<String> allProperties = SiteUtil
+                .fixCapturedProperties(instance, clss, properties);
```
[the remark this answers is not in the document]
romani: can we combine getProperties and fixCapturedProperties in other method ? to get properties ones (by one method) in a way we need it here.

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,493 @@
...
+    private static final String INDENT_LEVEL_18 = SiteUtil.getNewlineAndIndentSpaces(18);
+    /** A newline with 20 spaces of indentation. */
+    private static final String INDENT_LEVEL_20 = SiteUtil.getNewlineAndIndentSpaces(20);
+
+    /** Path to main source code folder. */
+    private static final String MAIN_FOLDER_PATH = Paths.get(
+            "src", "main", "java", "com", "puppycrawl", "tools", "checkstyle").toString();
+
+    /** List of files who are parent modules and contain certain properties that checks inherit. */
+    private static final List<File> PARENT_MODULES_WITH_PROPERTIES = List.of(
```
nrmancuso, line 80: `PARENT_MODULES_WITH_PROPERTIES` -> `MODULE_SUPER_CLASS_FILES`

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,493 @@
...
+        processModule(currentModuleName, moduleFile);
+        writePropertiesTable((XdocSink) sink);
+    }
+
+    /**
+     * Collect the properties setters javadocs of the parent modules.
+     *
+     * @throws MacroExecutionException if an error occurs during processing.
+     */
+    private static void processParentModules() throws MacroExecutionException {
```
nrmancuso, line 125: We need to revisit usage of `parent` and `parentModule` throughout this entire PR, this is very confusing. Take `RecordCompoenentNumberCheck` . It has a parent (`TreeWalker`) and a super class: `AbstractCheck`.

nrmancuso (changes requested):
Few more:

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -269,4 +367,763 @@ public static String getParentModule(Class<?> moduleClass)
...
+
+    /**
+     * Get a set of properties for the given class.
+     *
+     * @param clss the class to get the properties for.
+     * @return a set of properties for the given class.
+     */
+    public static Set<String> getProperties(Class<?> clss) {
+        final Set<String> result = new TreeSet<>();
+        final PropertyDescriptor[] map = PropertyUtils.getPropertyDescriptors(clss);
```
nrmancuso, line 379: `map` -> propertyDescriptors`

nrmancuso (changes requested):
Few more:

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraper.java
```diff
@@ -0,0 +1,161 @@
...
+    /** Name of the module being scraped. */
+    private static String moduleName = "";
+
+    /**
+     * Set the name of the module to be scraped.
+     *
+     * @param moduleName the module name.
+     */
+    public static void setModuleName(String moduleName) {
+        ClassAndPropertiesSettersJavadocScraper.moduleName = moduleName;
```
romani, line 65: we can get name of class from file name, do we need this setter ?

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: I was following the workings of the metadata scraper.

https://github.com/checkstyle/checkstyle/blob/c789b7dedc8823a64481da8bf48c297dbb7a2261/src/main/java/com/puppycrawl/tools/checkstyle/meta/MetadataGeneratorUtil.java#L57-L79

I don't mind changing to an instance-based scraping as long as you can point me to how this whole configuration would look like because I couldn't find what to do.

rnveach: I don't think we do it anywhere, even in main. https://github.com/checkstyle/checkstyle/blob/c789b7dedc8823a64481da8bf48c297dbb7a2261/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L393

@romani There is no way to provide TreeWalker an instance of a check. It is completely controlled by a configuration, which is a string class name.
https://github.com/checkstyle/checkstyle/blob/c789b7dedc8823a64481da8bf48c297dbb7a2261/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java#L133

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -269,4 +367,763 @@ public static String getParentModule(Class<?> moduleClass)
...
+     */
+    public static Set<String> fixCapturedProperties(
+            Object instance, Class<?> clss, Set<String> properties) {
+        final Set<String> result = new TreeSet<>(properties);
+        // remove global properties that don't need documentation
+        removeGlobalProperties(clss, result);
+
+        removeUndocumentedProperties(clss, result);
+
+        addNonExplicitProperties(instance, clss, result);
```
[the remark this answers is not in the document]
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,493 @@
...
+        new File(Paths.get(MAIN_FOLDER_PATH,
+                CHECKS, "javadoc", "AbstractJavadocCheck.java").toString()),
+        new File(Paths.get(MAIN_FOLDER_PATH,
+                "api", "AbstractFileSetCheck.java").toString()),
+        new File(Paths.get(MAIN_FOLDER_PATH,
+                CHECKS, "header", "AbstractHeaderCheck.java").toString())
+    );
+
+    /** Mao of all parent modules properties and their javadocs. */
+    private static final Map<String, DetailNode> PARENT_MODULES_PROPERTIES_JAVADOCS =
```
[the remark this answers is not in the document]
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,493 @@
...
+                CHECKS, NAMING, "AbstractNameCheck.java").toString()),
+        new File(Paths.get(MAIN_FOLDER_PATH,
+                CHECKS, "javadoc", "AbstractJavadocCheck.java").toString()),
+        new File(Paths.get(MAIN_FOLDER_PATH,
+                "api", "AbstractFileSetCheck.java").toString()),
+        new File(Paths.get(MAIN_FOLDER_PATH,
+                CHECKS, "header", "AbstractHeaderCheck.java").toString())
+    );
+
+    /** Mao of all parent modules properties and their javadocs. */
```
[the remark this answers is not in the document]
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraper.java
```diff
@@ -0,0 +1,161 @@
...
+public class ClassAndPropertiesSettersJavadocScraper extends AbstractJavadocCheck {
+
+    /**
+     * Map of scraped javadocs - name of property, javadoc detail node.
+     * The class javadoc is stored too, with the key being the module name.
+     */
+    private static final Map<String, DetailNode> JAVADOCS = new LinkedHashMap<>();
+
+    /** List of setter methods that the scraper ignores but shouldn't. */
+    private static final Set<String> SETTER_EXCEPTIONS = Set.of(
```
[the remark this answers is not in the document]
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraper.java
```diff
@@ -0,0 +1,161 @@
...
+ * given checkstyle module.
+ */
+@FileStatefulCheck
+public class ClassAndPropertiesSettersJavadocScraper extends AbstractJavadocCheck {
+
+    /**
+     * Map of scraped javadocs - name of property, javadoc detail node.
+     * The class javadoc is stored too, with the key being the module name.
+     */
+    private static final Map<String, DetailNode> JAVADOCS = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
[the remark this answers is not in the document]
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,493 @@
...
+     * @throws MacroExecutionException if link to the document file cannot be constructed.
+     */
+    private static void writeLinkToToken(Sink sink, String document, String tokenName)
+            throws MacroExecutionException {
+        sink.link(
+                SiteUtil.getLinkToDocument(currentModuleName,
+                        document)
+                        + "#"
+                        + tokenName
+        );
```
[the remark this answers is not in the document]
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -269,4 +401,843 @@ public static String getParentModule(Class<?> moduleClass)
...
+
+    /** Utility class for extracting description from a method's Javadoc. */
+    private static final class DescriptionExtractor {
+
+        /** Variable that tells if we are in a CODE_LITERAL token. */
+        private static boolean isInCodeLiteral;
+        /** Variable that tells if we are in an HTML_ELEMENT token. */
+        private static boolean isInHtmlElement;
+        /** Variable that tells if we are in an HREF attribute. */
+        private static boolean isInHrefAttribute;
```
stoyanK7, line 1091: IDEA inspections:
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/20544/workflows/29b03a15-8b8d-485a-bd8a-2bd63d1b2aa4/jobs/373704/artifacts
```
Static field <code>isInHrefAttribute</code> may not be initialized during class initialization #loc
...
```
Initialize it with `false` and checkstyle complains it is initialized to default value. Place it back in method and method becomes too complex. I am stuck with this one.

stoyanK7: https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&sinceLeakPeriod=true&pullRequest=13407&id=org.checkstyle%3Acheckstyle
Regarding the Sonarqube code smells - the first 2 ones with the brackets are not applicable because line length would become more than 100

Regarding the 2 methods that are too complex `getDescriptionFromJavadoc` and `getDefaultValue`, I don't think there's anything to gain from splitting them.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,403 @@
...
+        }
+        else {
+            final String type = SiteUtil.getType(field, propertyName, currentModuleName, instance);
+            final String relativePathToPropertyTypes = SiteUtil
+                    .getLinkToDocument(currentModuleName, "property_types.xml");
+            final String url = String
+                    .format(Locale.ROOT, "%s#%s", relativePathToPropertyTypes,
+                            type
+                                    .replace("[", ".5B")
+                                    .replace("]", ".5D"));
```
romani, line 267: a way better to read:
```
            final String relativePathToPropertyTypes = 
                    SiteUtil.getLinkToDocument(currentModuleName, "property_types.xml");
            final String ecapedType = type.replace("[", ".5B").replace("]", ".5D")        
            final String url = 
                    String.format(Locale.ROOT, "%s#%s", relativePathToPropertyTypes, ecapedType);
```

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,403 @@
...
+     * @param document the document to link to.
+     * @param tokenName the name of the token.
+     * @throws MacroExecutionException if link to the document file cannot be constructed.
+     */
+    private static void writeLinkToToken(Sink sink, String document, String tokenName)
+            throws MacroExecutionException {
+        final String link = SiteUtil.getLinkToDocument(currentModuleName,
+                        document)
+                        + "#"
+                        + tokenName;
```
romani, line 338:
```
        final String link = SiteUtil.getLinkToDocument(currentModuleName, document)
                        + "#" + tokenName;
```

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -269,4 +401,843 @@ public static String getParentModule(Class<?> moduleClass)
...
+     * certain conditions are met.
+     *
+     * @param instance the instance of the module.
+     * @param clss the class of the module.
+     * @return the non explicit properties.
+     */
+    private static Set<String> getNonExplicitProperties(
+            Object instance, Class<?> clss) {
+        final Set<String> result = new TreeSet<>();
+        if (AbstractCheck.class.isAssignableFrom(clss)) {
```
nrmancuso, line 566: I do not understand why we must pass around `clss` and use `Class#isAssignableFrom` everywhere. This method is slower and has a different implication than the `instanceOf` keyword. I am good with this code for now, but we should improve in the future at some point.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,403 @@
...
+        sink.rawText(INDENT_LEVEL_14);
+        sink.tableCell();
+
+        if (SiteUtil.TOKENS.equals(propertyName)) {
+            final AbstractCheck check = (AbstractCheck) instance;
+            final List<Integer> tokens = SiteUtil.subtractTokens(
+                    check.getDefaultTokens(), check.getRequiredTokens());
+            writeListOfTokens(sink, tokens);
+        }
+        else if ("javadocTokens".equals(propertyName)) {
```
romani, line 366: SiteUtils.JAVADOC_TOKENS

romani (changes requested):
item

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -269,4 +401,843 @@ public static String getParentModule(Class<?> moduleClass)
...
+         * @noinspectionreason TooBroadScope - complex nature of method requires large scope
+         */
+        // -@cs[NPathComplexity] Splitting would not make the code more readable
+        // -@cs[CyclomaticComplexity] Splitting would not make the code more readable.
+        private static String getDescriptionFromJavadoc(DetailNode javadoc, String moduleName)
+                throws MacroExecutionException {
+            resetValues();
+            final StringBuilder description = new StringBuilder(128);
+            final Deque<DetailNode> queue = new ArrayDeque<>();
+            pushOntoQueue(queue, getDescriptionNodes(javadoc));
```
nrmancuso, line 1111: No need to reinvent the wheel: ` getDescriptionNodes(j).forEach(queue::push);`

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -269,4 +401,843 @@ public static String getParentModule(Class<?> moduleClass)
...
+
+            // Perform DFS traversal on description nodes
+            while (!queue.isEmpty()) {
+                final DetailNode node = queue.pop();
+                pushOntoQueue(queue, Arrays.asList(node.getChildren()));
```
nrmancuso, line 1116: `Lists.reverse(Arrays.asList(n.children())).forEach(queue::push);`

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,403 @@
...
+     * @param sink sink to write to.
+     * @param propertyName the name of the property.
+     * @param propertyJavadoc the Javadoc of the property.
+     * @param instance the instance of the module.
+     * @param moduleJavadoc the Javadoc of the module.
+     * @throws MacroExecutionException if an error occurs during writing.
+     */
+    private static void writePropertyRow(Sink sink, String propertyName,
+                                         DetailNode propertyJavadoc, Object instance,
+                                            DetailNode moduleJavadoc)
```
romani, line 179: too late to do no w ..... just a note for future improvement in design:
there should be no dependency on DetailNode, all this data should be taken before and stored in some POJO objects, and serializer/formatter to xdoc sink just operates on simple objects to put data in final stream.
The same as we have in release notes. One module create structure/object by some logic, Another module do not care how value a calculated and simply place them in result file and care on formatting only.

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraper.java
```diff
@@ -0,0 +1,168 @@
...
+        clear();
+        moduleName = newModuleName;
+    }
+
+    /** Clear all scraped information. */
+    private static void clear() {
+        JAVADOC_FOR_MODULE_OR_PROPERTY.clear();
+        moduleName = null;
+    }
+
```
nrmancuso, line 75: This method seems pointless now, we should probably just clear the map in `initialize`, but it is not a big deal.

nrmancuso (changes requested):
Last from me:

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
nrmancuso: Let's create issue for general improvements, and discuss there. 

> POJO

I was thinking the same, to keep all extraction in one method and then deal with only what we need to avoid `isAssignable` checks everywhere, etc.

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraper.java
```diff
@@ -0,0 +1,168 @@
...
+    @Override
+    public void visitJavadocToken(DetailNode ast) {
+        final DetailAST blockCommentAst = getBlockCommentAst();
+        if (BlockCommentPosition.isOnMethod(blockCommentAst)) {
+            final DetailAST methodDef = getParentAst(blockCommentAst, TokenTypes.METHOD_DEF);
+            if (methodDef != null) {
+                final String methodName = methodDef.findFirstToken(TokenTypes.IDENT).getText();
+                if (CheckUtil.isSetterMethod(methodDef) && isMethodOfScrapedModule(methodDef)
```
nrmancuso, line 99: @stoyanK7 @romani we have a conflict with https://github.com/checkstyle/checkstyle/pull/13607

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -269,4 +401,843 @@ public static String getParentModule(Class<?> moduleClass)
...
+            }
+        }
+
+        return result;
+    }
+
+    /**
+     * Checks if the property is a global property. Global properties that are not
+     * documented.
+     *
```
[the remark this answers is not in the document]
stoyanK7: Added in Javadoc - `example id, severity, tabWidth`

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraper.java
stoyanK7: Not sure what to do here. Can `isSetterMethod ` be moved to `TokenUtil`? It has a few methods such as  `isBooleanLiteralType`, `isTypeDeclaration`, `isCommentType`..

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
nrmancuso: need a reverse in here, too, missed this in example

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraper.java
romani: Please copy isSetterMethod to scrapper.
Let's not deal with dependencies for now.
Setter is very specific term that point of different interpretations.

nrmancuso: I would be fine to leave it in checkutil, let's see what @romani thinks

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
stoyanK7: Not sure as well, but if it works, let's leave it. I added a draft task with a link to your comment on the project board.

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -269,4 +401,843 @@ public static String getParentModule(Class<?> moduleClass)
...
+     * Scrape the Javadocs of the class and its properties setters with
+     * ClassAndPropertiesSettersJavadocScraper.
+     *
+     * @param moduleName the name of the module.
+     * @param moduleFile the module file.
+     * @throws MacroExecutionException if an error occurs during processing.
+     */
+    private static void processModule(String moduleName, File moduleFile)
+            throws MacroExecutionException {
+        if (!moduleFile.isFile()) {
```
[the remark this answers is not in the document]
stoyanK7: Method is also used by `processSuperclasses()`

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
rnveach: It could just be hindsight. The original tests were built out slowly, and they eventually got pretty massive. It was rare we optimized it. When I first started creating this, I think I prefer to keep the class because I actually wanted to avoid instancing it. The code is about the class structure and not its instance. However, to get the default value of the fields, we were forced to use the instance since fields could be initialized directly, through a method, or in the constructor.

I was going to say that we were using the class and you can't use instanceof with it, but it does look like we pass around the actual instance in the same corners, so we can switch.

> Maybe this is to "cheat" Intellij?

No, but we did allow more suppressions than normal in this code since it was test only and so massive.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -0,0 +1,403 @@
...
+    private static void writePropertyTypeCell(Sink sink, String propertyName,
+                                              Field field, Object instance)
+            throws MacroExecutionException {
+        sink.rawText(INDENT_LEVEL_14);
+        sink.tableCell();
+
+        if (SiteUtil.TOKENS.equals(propertyName)) {
+            final AbstractCheck check = (AbstractCheck) instance;
+            final List<Integer> tokens = SiteUtil
+                    .subtractTokens(check.getAcceptableTokens(), check.getRequiredTokens());
```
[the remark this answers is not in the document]
rnveach: The list is displaying tokens that users can customize/personalize for their own needs. We subtract "required" tokens because they can't be customized and are mandatory to always be included whether the user lists them or not.

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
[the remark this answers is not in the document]
rnveach: Global = properties that come from the base classes and are common to all modules.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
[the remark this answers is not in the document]
nrmancuso: Ah, I see. Let's rename this to `configurabletokens` to make this obvious.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
stoyanK7: Done. No problem, I caught it.

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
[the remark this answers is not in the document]
stoyanK7: Done. Added explanation to javadoc

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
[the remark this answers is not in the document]
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
romani: https://github.com/checkstyle/checkstyle/issues/13673

romani (approved):
Ok to merge.

Thanks a lot for your hard work on this

nrmancuso (approved):
This is good for now, let's merge this and improve later.

PR #13412, opened 2023-07-20
Issue #13321: Kill mutation for DetectorOption

Kevin222004:
Issue #13321: Kill mutation for DetectorOption

-----

# Mutation 
https://github.com/checkstyle/checkstyle/blob/2335abaea101774a23d0bdaf55f19e4f0cad0d37/config/pitest-suppressions/pitest-regexp-suppressions.xml#L3-L10

-----

# Explaination

It is some inner class, and all fields already not initialized, so with such update we do implementation consistent.

All the time whenever we use this message field in check we use the build method and the build method is 
https://github.com/checkstyle/checkstyle/blob/2335abaea101774a23d0bdaf55f19e4f0cad0d37/src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/DetectorOptions.java#L246-L250

message is set like `message = Optional.ofNullable(message).orElse("");` so if the message is null then it will be `""` empty string so removal will not make issue

-----

# Regression :- 

-----

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/af6965cd82853221b89f6ba3b81b996a/raw/89f2f0efe08703af1c3c9b9f3b5d2330a064dea3/RegexpSingleline.xml
Report label: Regression-2

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/DetectorOptions.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/DetectorOptions.java
@@ -41,7 +41,7 @@ public final class DetectorOptions {
      */
     private String format;
     /** The message to report on detection. If blank, then use the format. */
-    private String message = "";
+    private String message;
     /** Minimum number of times regular expression should occur in a file. */
     private int minimum;
     /** Maximum number of times regular expression should occur in a file. */
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpsingleline/InputRegexpSinglelineSemantic10.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpsingleline/InputRegexpSinglelineSemantic10.java
@@ -1,0 +1,19 @@
+/*
+RegexpSingleline
+format = SYSTEM\\.(OUT)|(ERR)\\.PRINT(LN)?\\(
+message =
+ignoreCase = true
+minimum = (default)0
+maximum = (default)0
+fileExtensions = (default)all files
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.regexp.regexpsingleline;
+
+public class InputRegexpSinglelineSemantic10 {
+    public static void main(String[] args) {
+        System.out.println("str"); // violation ''
+    }
+}
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -4392,7 +4392,7 @@
   <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/DetectorOptions.java</fileName>
     <specifier>initialization.fields.uninitialized</specifier>
-    <message>the constructor does not initialize fields: reporter, format, suppressor, pattern</message>
+    <message>the constructor does not initialize fields: reporter, format, message, suppressor, pattern</message>
     <lineContent>private DetectorOptions() {</lineContent>
   </checkerFrameworkError>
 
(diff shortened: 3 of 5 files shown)


Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

rdiachenko (approved):
lgtm

src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpsingleline/InputRegexpSinglelineSemantic10.java
```diff
@@ -0,0 +1,19 @@
...
+fileExtensions = (default)all files
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.regexp.regexpsingleline;
+
+public class InputRegexpSinglelineSemantic10 {
+    public static void main(String[] args) {
+        System.out.println("str"); // violation
```
romani, line 17: Can we use `// violation ''`
?

To make it clear that message via empty.

romani (changes requested):
Items

src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpsingleline/InputRegexpSinglelineSemantic10.java
Kevin222004: > Can we use // violation ''
 ?

Yes, I updated the code

src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/DetectorOptions.java
```diff
@@ -41,7 +41,7 @@ public final class DetectorOptions {
      */
     private String format;
     /** The message to report on detection. If blank, then use the format. */
-    private String message = "";
+    private String message;
```
romani, line 44: This is property of check.
We make it nullable, but user can not set null by config, so user can not bring Check back to default state.
Not good.

Even Javadoc tells us that empty is special value. We need to change javadoc.

But really we need to change code. Can remove `Optional.ofNullable(message).orElse("")` and treat message as not nullable?

romani (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/DetectorOptions.java
Kevin222004: I got your point. but the check which use this builder Pattern such like https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/RegexpSinglelineCheck.java 

In this check itself docs says that by default value is null so even if user can't set null but still by default value is null and everytime https://github.com/checkstyle/checkstyle/blob/58fdccf566a46b2ad4eea3987be3157d323bb3a8/src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/RegexpSinglelineCheck.java#L218-L227 
we call the build method to get the value 
https://github.com/checkstyle/checkstyle/blob/58fdccf566a46b2ad4eea3987be3157d323bb3a8/src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/RegexpSinglelineCheck.java#L206
so when user doesn't set any value then message become null so we can't remove ` Optional.ofNullable(message).orElse("")`. 

and I also https://checkstyle.org/property_types.html don't see this in docs

romani: This is my bad.
It is not a Module! 
It is some inner class, and all fields already not initialized, so with such update we do implementation consistent.

romani (approved):
Ok to merge

Vyom-Yadav (approved):
LGTM!

PR #13423, opened 2023-07-24
Issue #13422: Implement module violation messages macro 

### LINKED ISSUE #13422: Implement module violation messages macro
The initial idea of having an annotation for message keys has been scraped (https://github.com/checkstyle/checkstyle/issues/13114). Since we now use macros for examples, it makes sense to make one for violation messages.

The macro should be called like
```xml
        <macro name="violation-messages">
            <param name="path" value="AnnotationLocation"/>
        </macro>
```

and generate the following content
```xml
        <ul>
          <li>
            <a href="https://github.com/search?q=path%3Asrc%2Fmain%2Fresources%2Fcom%2Fpuppycrawl%2Ftools%2Fcheckstyle%2Fchecks%2Fannotation%20path%3A**%2Fmessages*.properties+repo%3Acheckstyle%2Fcheckstyle+%22annotation.location%22">
            annotation.location</a>
          </li>
          <li>
            <a href="https://github.com/search?q=path%3Asrc%2Fmain%2Fresources%2Fcom%2Fpuppycrawl%2Ftools%2Fcheckstyle%2Fchecks%2Fannotation%20path%3A**%2Fmessages*.properties+repo%3Acheckstyle%2Fcheckstyle+%22annotation.location.alone%22">
            annotation.location.alone</a>
          </li>
        </ul>
```

stoyanK7:
Resolves #13422


src/main/java/com/puppycrawl/tools/checkstyle/site/XdocsTemplateSink.java
```diff
@@ -62,4 +67,11 @@ public void body_() {
         flush();
         init();
     }
+
+    @Override
+    public void link(String name) {
```
stoyanK7, line 72: I need to mention in Javadoc that I added this method because the default `XdocSink` adds a `class="externalLink"` that I don't want.

This is a trimmed-down version of the default `Xhtml5BaseSink`
https://github.com/apache/maven-doxia/blob/24ab8ba193ec44007333775576277f9aa9ea165f/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java#L1572-L1598

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,274 @@
...
+            throws MacroExecutionException {
+        try {
+            // required for package/private classes
+            field.trySetAccessible();
+            return field.get(instance).toString();
+        }
+        catch (IllegalAccessException ex) {
+            throw new MacroExecutionException("Couldn't get field value", ex);
+        }
+    }
```
stoyanK7, line 173: All this is a literal copy of https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/CheckUtil.java.

But I have no access to it.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,274 @@
...
+     * Returns the message keys for the given fields.
+     *
+     * @param messageKeyFields the fields.
+     * @param instance the instance of the module.
+     * @return the message keys.
+     * @throws MacroExecutionException if the message keys could not be retrieved.
+     */
+    private static Set<String> getMessageKeys(Set<Field> messageKeyFields, Object instance)
+            throws MacroExecutionException {
+        final Set<String> messageKeys = new TreeSet<>();
```
nrmancuso, line 148: Why TreeSet?

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,274 @@
...
+     * @param clss the class of the field.
+     * @param messageKey the message key.
+     */
+    private static void createListItem(Sink sink, Class<?> clss, String messageKey) {
+        final String messageKeyUrl = constructMessageKeyUrl(clss, messageKey);
+
+        // Place the <li>.
+        sink.rawText(getNewlineAndIndentSpaces(10));
+        // This is a hack to prevent a newline from being inserted by the default sink.
+        // Once we get rid of the custom parser, we can remove this.
```
nrmancuso, line 227: `until ...`

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,274 @@
...
+        for (Field field : fields) {
+            if (field.getName().startsWith("MSG_")) {
+                checkstyleMessages.add(field);
+            }
+        }
+
+        // deep scan class through hierarchy
+        final Class<?> superModule = clss.getSuperclass();
+
+        if (superModule != null && shouldScanDeepClassForMessages(superModule)) {
```
nrmancuso, line 102: Why not add null check to `shouldScanDeepClassForMessages`?

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,274 @@
...
+            throws MacroExecutionException {
+        final Set<Field> checkstyleMessages = new HashSet<>();
+
+        final Field[] fields = clss.getDeclaredFields();
+
+        for (Field field : fields) {
+            if (field.getName().startsWith("MSG_")) {
+                checkstyleMessages.add(field);
+            }
+        }
```
nrmancuso, line 97:
```
Set s = Arrays.stream(fields)
  .filter(f -> ...)
  .Collect(Collections.toSet);
```

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,274 @@
...
+     * @return the message keys.
+     * @throws MacroExecutionException if the message keys could not be retrieved.
+     */
+    private static Set<String> getMessageKeys(Set<Field> messageKeyFields, Object instance)
+            throws MacroExecutionException {
+        final Set<String> messageKeys = new TreeSet<>();
+        for (Field field : messageKeyFields) {
+            messageKeys.add(getFieldValue(field, instance));
+        }
+        return messageKeys;
```
nrmancuso, line 152:
```
return messageKeyFields.stream()
  .map(f -> getfieldvalue(f, instance))
  .collect(...);
```

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
nrmancuso: Hm, this brings up a good point. Should this be considered `internal`? @romani @rnveach

nrmancuso (changes requested):
Items contingent on response at :

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
stoyanK7: We want to preserve the order of the keys. Using HashSet messed up the order and failed XdocsJavadocsTest. That's what's being done in `XdocsPagesTest`

https://github.com/checkstyle/checkstyle/blob/0c2ab5338db569b4699d5d20121d1bd508394eb9/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1381-L1389

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * Constructs a URL to GitHub that searches for the message key.
+     *
+     * @param clss the class of the module.
+     * @param messageKey the message key.
+     * @return the URL to GitHub.
+     */
+    private static String constructMessageKeyUrl(Class<?> clss, String messageKey) {
+        return "https://github.com/search?q="
```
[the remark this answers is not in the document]
stoyanK7: Url was taken from https://github.com/checkstyle/checkstyle/blob/0c2ab5338db569b4699d5d20121d1bd508394eb9/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1427-L1433

Right now we don't want to escape this text because we convert from template to regular xdoc. Regular xdoc expects it to be unescaped and default xdoc parser and sink will take care of escaping when generating the website(`mvn site`):
https://github.com/apache/maven-doxia/blob/master/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java#L1595
This was part of the reason why I overwrote `link(x)` in `XdocsTemplateSink`

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
stoyanK7: Screenshot from 2023-07-27 13-39-40

IntelliJ even says it's unnecessary

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,274 @@
...
+                + "path%3Asrc%2Fmain%2Fresources%2F"
+                + clss.getPackage().getName().replace(".", "%2F")
+                + "%20path%3A**%2Fmessages*.properties+repo%3Acheckstyle%2F"
+                + "checkstyle+%22" + messageKey + "%22";
```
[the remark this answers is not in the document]
stoyanK7: Yes, I need to add this to the `link(x)` I overwrote in `XdocsTemplateSink`
https://github.com/apache/maven-doxia/blob/24ab8ba193ec44007333775576277f9aa9ea165f/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java#L1561

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
nrmancuso: Then why do we have it here?

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
romani: it is Util in test area of code, I am ok make all methods that we need public

stoyanK7: They are public, the thing is, I can not import code from `test` in `main` :/

romani: macros code will go to separate project, as we should not export by jar such doxia site classes :).
please keep them as copy, but you can make SiteUtil.java to make copy-paste more explicit and declarative.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,280 @@
...
+
+        if (superModule != null && shouldScanDeepClassForMessages(superModule)) {
+            checkstyleMessages.addAll(getCheckMessageKeyFields(superModule));
+        }
+
+        // special cases that require additional classes
+        try {
+            if (clss == RegexpMultilineCheck.class) {
+                checkstyleMessages.addAll(getCheckMessageKeyFields(Class
+                    .forName("com.puppycrawl.tools.checkstyle.checks.regexp.MultilineDetector")));
```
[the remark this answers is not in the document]
stoyanK7: > 'com.puppycrawl.tools.checkstyle.checks.regexp.MultilineDetector' is not public in 'com.puppycrawl.tools.checkstyle.checks.regexp'. Cannot be accessed from outside package

Same below.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
nrmancuso: Hm, why are we forced to use `Set` at all?

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
stoyanK7: I changed this code quite a bit. We don't need the `shouldScanDeepClassForMessages` because we want to scan the parent classes of every check for violation messages. We need the `superModule != null` because there is recursion.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
stoyanK7: Method is removed.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
stoyanK7: Done.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
stoyanK7: I changed the method(`getCheckMessageKeys`) to get the `String` of the message keys. Doing this lambda would require me to do 
```java
    final List<String> checkstyleMessageKeys = Arrays.stream(fields)
            .filter(field -> field.getName().startsWith("MSG_"))
            .map(field -> {
                try {
                    return getFieldValue(field, checkInstance);
                }
                catch (IllegalAccessException exception) {
                    throw new RuntimeException(exception);
                }
            })
```
which doesn't look good in my opinion.

nrmancuso: Agreed.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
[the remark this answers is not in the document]
nrmancuso: Ok, we got it from elsewhere, let's stick with current way of doing this.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
stoyanK7: I got rid of the `Set` and just used a `List`.

@rnveach could you please tell if using a `Set` here is a must? Could there be some duplicates or?

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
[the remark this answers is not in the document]
stoyanK7: I did one final try with `UrlEncoder.encode` and it works. Almost. 
```java
URLEncoder.encode("....", StandardCharsets.UTF_8);
```
https://www.baeldung.com/java-url-encoding-decoding

A space can either be encoded as a `%20` or `+`. This function encodes it as a `+`. The problem is that the following test  expects both `%20` and `+`.

https://github.com/checkstyle/checkstyle/blob/0c2ab5338db569b4699d5d20121d1bd508394eb9/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1427-L1433

Shall we move this to an issue and deal with it after all violation messages are generated from the macro? Then I can do the encoding and change the test.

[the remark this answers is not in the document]
nrmancuso: Yes, let's make another issue.

nrmancuso (approved):

nrmancuso:
I would like to make sure we have response from @rnveach at https://github.com/checkstyle/checkstyle/pull/13423#discussion_r1279229444 before we merge.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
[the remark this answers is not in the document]
stoyanK7: Opened at https://github.com/checkstyle/checkstyle/issues/13490

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
romani: No duplicates are allowed in messages

nrmancuso: > No duplicates are allowed in messages

Please elaborate on this; do we have some way to enforce this or do we make sure by Set API?

config/suppressions.xml
```diff
@@ -179,4 +179,9 @@
   <!-- DetailAST API requires large number of methods -->
   <suppress checks="MethodCount"
             files="[\\/]src[\\/]main[\\/]java[\\/]com[\\/]puppycrawl[\\/]tools[\\/]checkstyle[\\/]DetailAstImpl.java"/>
+
+  <!-- The numbers are explained by the method they are calling.
+       Until https://github.com/checkstyle/checkstyle/issues/13426 -->
+  <suppress checks="MagicNumber"
```
romani, line 185: Sorry , not good reason.
With such reasoning we can explain all.

You might need `final int baseIndent`, 
`indentLevel1` ...

config/spotbugs-exclude.xml
```diff
@@ -280,4 +280,10 @@
     name="com.puppycrawl.tools.checkstyle.site.XdocsTemplateParser"/>
     <Bug pattern="STT_TOSTRING_STORED_IN_FIELD"/>
   </Match>
+  <Match>
+    <!-- We cast to XdocSink to avoid unwanted newlines.
+         Until https://github.com/checkstyle/checkstyle/issues/13426 -->
+    <Class name="com.puppycrawl.tools.checkstyle.site.ViolationMessagesMacro"/>
+    <Bug pattern="PDP_POORLY_DEFINED_PARAMETER"/>
```
romani, line 287: Please share error message. To let me see on what code id complaining on.
Can we use more fine grained suppression?

config/intellij-idea-inspections.xml
```diff
@@ -3025,6 +3026,8 @@
                    enabled_by_default="false"/>
   <inspection_tool class="MagicNumber" enabled="true" level="ERROR" enabled_by_default="true">
     <scope name="Tests" level="ERROR" enabled="false"/>
+    <!-- Temporary method until https://github.com/checkstyle/checkstyle/issues/13426 -->
+    <scope class="ViolationMessagesMacro" level="ERROR" enabled="false"/>
```
romani, line 3030: Please share what force you to do this

pom.xml
```diff
@@ -863,6 +863,8 @@
                     <!-- until https://github.com/checkstyle/checkstyle/issues/13393 -->
                     <exclude>com.puppycrawl.tools.checkstyle.site.XdocsTemplateParser</exclude>
                     <exclude>com.puppycrawl.tools.checkstyle.site.ExampleMacro</exclude>
+                    <exclude>com.puppycrawl.tools.checkstyle.site.ViolationMessagesMacro</exclude>
+                    <exclude>com.puppycrawl.tools.checkstyle.site.SiteUtil</exclude>
                     <!-- Parser related classes -->
```
romani, line 868: Can we exclude by `*`?

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.maven.doxia.macro.AbstractMacro;
+import org.apache.maven.doxia.macro.Macro;
+import org.apache.maven.doxia.macro.MacroExecutionException;
+import org.apache.maven.doxia.macro.MacroRequest;
+import org.apache.maven.doxia.module.xdoc.XdocSink;
+import org.apache.maven.doxia.sink.Sink;
+import org.codehaus.plexus.component.annotations.Component;
+
+/**
+ * A macro that inserts an unordered list of the violation messages.
```
romani, line 34: Why unordered?
Ordering is good.
Even we are ordering in xdoc, it will be easy diff for us.

src/xdocs/checks/annotation/annotationlocation.xml
```diff
@@ -233,8 +233,8 @@ private boolean field2;
 @Override //ok
 public boolean equals(Object obj) { return true; }
 @Mock DataLoader loader; //ok
-@SuppressWarnings("deprecation") DataLoader loader; //ok
-@SuppressWarnings("deprecation") public int foo() { return 1; } //violation
+@SuppressWarnings(&quot;deprecation&quot;) DataLoader loader; //ok
+@SuppressWarnings(&quot;deprecation&quot;) public int foo() { return 1; } //violation
```
romani, line 237: Not sure why messages related PR has this change.

Can we simply move templatization to separate PR and merge it quicker. To let this PR be only messages related.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
stoyanK7: Unordered as in `<ul>`, bullet points - https://www.w3schools.com/html/html_lists_unordered.asp

config/spotbugs-exclude.xml
stoyanK7: > [ERROR] Medium: Method com.puppycrawl.tools.checkstyle.site.ViolationMessagesMacro.createListItem(Sink, Class, String) defines parameters more abstractly than needed to function properly [com.puppycrawl.tools.checkstyle.site.ViolationMessagesMacro] At ViolationMessagesMacro.java:[line 83] PDP_POORLY_DEFINED_PARAMETER

```java
        final XdocSink xdocSink = (XdocSink) sink;
```

src/xdocs/checks/annotation/annotationlocation.xml
stoyanK7: This is the underlying parser escaping all text inside `<source>` elements. I am having a hard time overriding it.

config/suppressions.xml
stoyanK7: Done. Removed suppression

config/intellij-idea-inspections.xml
stoyanK7: Removed suppression. Same as https://github.com/checkstyle/checkstyle/pull/13423#discussion_r1280641475

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
stoyanK7: Done. Removed `unordered`

config/spotbugs-exclude.xml
romani: > PDP_POORLY_DEFINED_PARAMETER This method defines parameters at a more abstract level than is actually needed to function correctly, as the code casts these parameters to more concrete types. Since this method is not derivable, you should just define the parameters with the type that is needed.

https://fb-contrib.sourceforge.net/bugdescriptions.html

romani: `createListItem((XdocSink)sink,....)`

src/xdocs/checks/annotation/annotationlocation.xml
romani: I meant , why we create new template file in this PR. We just need replacement of content to macros in existing templates

romani: It is ok to not migrate all pages now , but at least some, but still in separate easy to review/merge PRs.

config/intellij-idea-inspections.xml
```diff
@@ -2594,6 +2594,7 @@
     <scope name="Tests" level="ERROR" enabled="false">
       <option name="ignoreInstanceofOnLibraryClasses" value="false"/>
     </scope>
+    <scope class="SiteUtil" level="ERROR" enabled="false"/>
```
romani, line 2597: Explain this please

romani (changes requested):
Items

src/xdocs/checks/annotation/annotationlocation.xml
romani: Please choose 10 modules and migrate them to templates ASAP, it will be your base set of modules to test your macros.
We still be able to make implementation that satisfies some of such modules and work on extension in separate PRs.

romani: If that is kind of automate-able, please migrate all, I will merge quickly.
It will be easily review as they should be no diff (except escaping and indentation maybe)

stoyanK7: Opened at https://github.com/checkstyle/checkstyle/issues/13498.

config/intellij-idea-inspections.xml
stoyanK7: Violation at

https://app.circleci.com/pipelines/github/checkstyle/checkstyle/19794/workflows/5625e8ee-9bc6-4358-8ecb-295367dc2619/jobs/343601/artifacts

```xml
<problem>
<file>
file://$PROJECT_DIR$/src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
</file>
<line>108</line>
<module>project</module>
<package>com.puppycrawl.tools.checkstyle.site</package>
<entry_point TYPE="method" FQNAME="com.puppycrawl.tools.checkstyle.site.ViolationMessagesMacro java.util.Set<java.lang.reflect.Field> getCheckMessageKeyFields(java.lang.Class<?> clss)"/>
<problem_class id="InstanceofChain" severity="ERROR" attribute_key="ERRORS_ATTRIBUTES">Chain of 'instanceof' checks</problem_class>
<description>
Chain of class equality checks indicates abstraction failure #loc
</description>
<highlighted_element>if</highlighted_element>
<language>JAVA</language>
<offset>12</offset>
<length>2</length>
</problem>
</problems>
```

It's talking about 
```java
            if (module == RegexpMultilineCheck.class) {
                checkstyleMessageKeys.addAll(getCheckMessageKeys(
                        Class.forName(
                                "com.puppycrawl.tools.checkstyle.checks.regexp.MultilineDetector"
                        ), checkInstance)
                );
            }
            else if (module == RegexpSinglelineCheck.class
                    || module == RegexpSinglelineJavaCheck.class) {
                checkstyleMessageKeys.addAll(getCheckMessageKeys(
                        Class.forName(
                                "com.puppycrawl.tools.checkstyle.checks.regexp.SinglelineDetector"
                        ), checkInstance)
                );
            }
```

These are edge cases in my opinion. Method copied from https://github.com/checkstyle/checkstyle/blob/34e7d5c714b2df62781790537c6127b5712ec097/src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/CheckUtil.java#L211

src/xdocs/checks/annotation/annotationlocation.xml
stoyanK7: Done. Changed an existing template instead.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
romani: Set, will hide duplicates to make illusion in html.
Better to use List to simply print as is.

The only extra benefit from fancier collection I see, is ordering.
It will be good to order alphabetically in html.
But we can do this later on, after transition to templates is done.

stoyanK7: > It will be good to order alphabetically in html.

It is already done in `ViolationMessagesMacro::createListOfMessages`
```java
        Collections.sort(messageKeys);
```

config/intellij-idea-inspections.xml
romani: We suppress inspections by javadoc tags, please suppress by it

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -0,0 +1,171 @@
...
+                        Class.forName(
+                                "com.puppycrawl.tools.checkstyle.checks.regexp.MultilineDetector"
+                        ), checkInstance)
+                );
+            }
+            else if (module == RegexpSinglelineCheck.class
+                    || module == RegexpSinglelineJavaCheck.class) {
+                checkstyleMessageKeys.addAll(getCheckMessageKeys(
+                        Class.forName(
+                                "com.puppycrawl.tools.checkstyle.checks.regexp.SinglelineDetector"
```
romani, line 90: https://www.tutorialspoint.com/get-the-fully-qualified-name-of-a-class-in-java

Use java method to get fully qualified name.

romani (changes requested):
Items

config/intellij-idea-inspections.xml
romani: And this violation is right, it is hack.
We should have separate ticket to improve it by refactoring of modules to have  all MSG fields in module as field, all child, nested, composited classes should take by constructor content of message.
We already discussed it multiple times, and now we see real need for this

romani: Just to be clear: we suppress in javadoc `noinspection` tag by link to issue where we will do it in better way.

We will continue extend this if-else while converting more and more to use of this macros, and in separate PR will do optimization to avoid such hack.

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
stoyanK7: Same as https://github.com/checkstyle/checkstyle/pull/13423#discussion_r1272913610. Class is not public and I have no access to it
> 'com.puppycrawl.tools.checkstyle.checks.regexp.MultilineDetector' is not public in 'com.puppycrawl.tools.checkstyle.checks.regexp'. Cannot be accessed from outside package

config/intellij-idea-inspections.xml
stoyanK7: Opened at https://github.com/checkstyle/checkstyle/issues/13500

stoyanK7: Done. Suppressed.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
rnveach: You can't use an object that is based on hashcode. It is JVM specific and different runs can generate different orders. We can't use insertion order here because reflection is not guaranteed an order and we want to ensure our documentation is consistent. This is why we originally used `TreeSet`.

I recommend keeping this a copy/paste and just create new issues to fix anything.

nrmancuso: @rnveach awesome explanation, thanks a lot. @stoyanK7 please restore copy pasta and add a comment about why we use treeset

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,126 @@
...
+    public void execute(Sink sink, MacroRequest request) throws MacroExecutionException {
+        // until https://github.com/checkstyle/checkstyle/issues/13426
+        if (!(sink instanceof XdocSink)) {
+            throw new MacroExecutionException("Expected Sink to be an XdocSink.");
+        }
+        final String checkName = (String) request.getParameter("checkName");
+        final Object instance = SiteUtil.getModuleInstance(checkName);
+        final Class<?> clss = instance.getClass();
+        final List<String> messageKeys = SiteUtil.getCheckMessageKeys(clss, instance);
+        createListOfMessages((XdocSink) sink, clss, messageKeys);
```
rnveach, line 48: So what is the plan in terms of disabling javadoc. Are we not going to have these type of generation in meta and have our macros get the data from meta? To me this makes the most sense, instead of meta having another custom implementation.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.maven.doxia.macro.MacroRequest;
+import org.apache.maven.doxia.module.xdoc.XdocSink;
+import org.apache.maven.doxia.sink.Sink;
+import org.codehaus.plexus.component.annotations.Component;
+
+/**
+ * A macro that inserts a list of the violation messages.
+ */
+@Component(role = Macro.class, hint = "violation-messages")
+public class ViolationMessagesMacro extends AbstractMacro {
```
rnveach, line 37: So we are not going to disable the violation portion of the xdocspages test for classes we have converted to use macros to show it isn't needed anymore?

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -0,0 +1,175 @@
...
+import com.puppycrawl.tools.checkstyle.PackageObjectFactory;
+import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
+import com.puppycrawl.tools.checkstyle.checks.regexp.RegexpMultilineCheck;
+import com.puppycrawl.tools.checkstyle.checks.regexp.RegexpSinglelineCheck;
+import com.puppycrawl.tools.checkstyle.checks.regexp.RegexpSinglelineJavaCheck;
+
+/**
+ * Utility class for site generation.
+ */
+public final class SiteUtil {
```
rnveach, line 41: If we are leaving this in production, then why don't we remove the test code and have it use this instead?

romani (approved):
Ok to merge

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
romani: This implementation will substitute  getting data from metadata.
For now validation javadoc vs xdoc help to test this implementation until it gets its maturity.
Such "another implementation" will me main implementation and source of truth soon.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
romani: Not for now, but as we convert all modules to templates, and all continue  to match, we can disable it. May be we will make some hack to allow migration suppression for such validation.

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
romani: This code is still planned to be moved to separate project. I am ok with such transitional copy paste of utils.

src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java
stoyanK7: @nrmancuso Done at https://github.com/checkstyle/checkstyle/pull/13505.

PR #13425, opened 2023-07-24
Issue #13321: Kill mutation for CheckUtil

Kevin222004:
Issue #13321: Kill mutation for CheckUtil

-----

# Mutation 
https://github.com/checkstyle/checkstyle/blob/f4a33f90bf8eda8ef2cdf7f66fd60878d2aa1736/config/pitest-suppressions/pitest-utils-suppressions.xml#L75-L82

-----

# Explaination

Usage of method in check :- https://checkstyle.org/checks/coding/nestedifdepth.html#NestedIfDepth

https://github.com/checkstyle/checkstyle/blob/f4a33f90bf8eda8ef2cdf7f66fd60878d2aa1736/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java#L145-L149
This method is used by https://github.com/checkstyle/checkstyle/blob/f4a33f90bf8eda8ef2cdf7f66fd60878d2aa1736/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java#L121-L126
this method is used by NestedIfDepthCheck and the parameter ast is always LiteralIf and in the mutated method the parameter passed is always parent of literalIF 
https://github.com/checkstyle/checkstyle/blob/f4a33f90bf8eda8ef2cdf7f66fd60878d2aa1736/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java#L146
so here the ast is always the parent of LITERAL_IF token abd it will always be `TokenTypes.SLIST` whether it is used in method, constructor or in Switch cases

-----

# Regression :- 

-----

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/b79c1d6f36c0d6db4516ff0aa57b8aee/raw/6e708ea645d0a75bfde0bd2baaff2adbf01730e0/NestedIf.xml
Report label: Regression-1

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
@@ -143,8 +143,7 @@ public final class CheckUtil {
      * @return whether the toke does represent an SLIST as part of an ELSE
      */
     private static boolean isElseWithCurlyBraces(DetailAST ast) {
-        return ast.getType() == TokenTypes.SLIST
-            && ast.getChildCount() == 2
+        return ast.getChildCount() == 2
             && isElse(ast.getParent());
     }
 
--- a/config/pitest-suppressions/pitest-utils-suppressions.xml
+++ b/config/pitest-suppressions/pitest-utils-suppressions.xml
@@ -72,14 +72,14 @@
     <lineContent>for (DetailAST token = node.getParent();</lineContent>
   </mutation>
 
-  <mutation unstable="false">
-    <sourceFile>CheckUtil.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.utils.CheckUtil</mutatedClass>
-    <mutatedMethod>isElseWithCurlyBraces</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
-    <description>removed conditional - replaced equality check with true</description>
-    <lineContent>return ast.getType() == TokenTypes.SLIST</lineContent>
-  </mutation>
+
+
+
+
+
+
+
+
 
   <mutation unstable="false">
     <sourceFile>CheckUtil.java</sourceFile>


Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
```diff
@@ -143,8 +143,7 @@ private static boolean isElse(DetailAST ast) {
      * @return whether the toke does represent an SLIST as part of an ELSE
      */
     private static boolean isElseWithCurlyBraces(DetailAST ast) {
```
rdiachenko, line 145: This method is used in `NestedIfDepthCheck` only which makes sense to move it there as for me.

Regarding `NestedIfDepthCheck`, it is super weird check:
```
// nesting == 2
if (true) {
} else {
    if (true) {
    } else {
        if (false) { // I would expect violation here, but I get nothing
        } else {
        }
    }
}

// nesting == 2
if (true) {
    if (true) {
        if (false) { // violation here occurred as expected
        } else {
        }
    } else {
    }
} else {
}
```
@romani do you know why this check handles only nested `if` and ignores nesting in `else` blocks?

romani: I do not see any good reason https://checkstyle.org/checks/coding/nestedifdepth.html#NestedIfDepth , it is  defect.
@rdiachenko , please create issue.

------

if util method is used only in one Check, it should be moved to Check and covered.

Kevin222004: Done The method is moved to check itself

romani: @Kevin222004 , please create issue to speed up a process and share link here

romani (approved):
Ok to merge.
Great update, less confusing util methods that are used ones and less pure unit testing.

But don't forget to open issue on Check.

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
rdiachenko: Issue created: https://github.com/checkstyle/checkstyle/issues/13493

rdiachenko (approved):
lgtm

romani:
@Kevin222004 , please resolve conflict

Issue #13421, opened 2023-07-23
Add `@since` in javadoc of each property setter


stoyanK7:
A requirement for implementing module properties macro
https://github.com/checkstyle/checkstyle/issues/13399

One of the 5 attributes of each property is the `since` attribute. In order for us to implement the properties macro, this information has to be stored somewhere in the class java file. Since we decided to use the setters of a module for generating the properties, the `since` has to be added in the property javadoc.

![Screenshot from 2023-07-23 19-23-42](https://github.com/checkstyle/checkstyle/assets/23459549/7ae92234-ab22-4ab2-bafa-0faf63a75c69)

For example
```diff
    /**
     * Setter to allow single parameterless annotation to be located on the same line as
     * target element.
     *
     * @param allow User's value of allowSamelineSingleParameterlessAnnotation.
+     * @since 6.1
     */
    public final void setAllowSamelineSingleParameterlessAnnotation(boolean allow) {
        allowSamelineSingleParameterlessAnnotation = allow;
    }
```

relentless-pursuit:
I am on it.

nrmancuso:
We also need to add a check on this.

romani:
@relentless-pursuit, please avoid all issues that are assigned to project.

PR #13438, opened 2023-07-26
Issue #13421: Add since in javadoc of each property setter for block checks

stoyanK7:
Part of #13421 

---
Adds `@since` to each property setter.

Taken from https://checkstyle.sourceforge.io/checks/blocks/index.html


romani:
@stoyanK7 , why draft?

stoyanK7:
@romani We need to add a test first.

romani (approved):
ok to merge.

Looks like since for "tokens" is skipped as it is more complicated, we can do it later on

PR #13479, opened 2023-07-29
Issue #13321: Kill mutation for ScopeUtil

Kevin222004:
Issue #13321: Kill mutation for ScopeUtil

------

# Mutation 
https://github.com/checkstyle/checkstyle/blob/bb64c20cbcf6e543b297b6e809f379bd8746b5e7/config/pitest-suppressions/pitest-utils-suppressions.xml#L345-L352

------

# Explaination
I have change the design. And this will not be cheat to pitest because. we have three if statment so each time all the condition will be checked. Instead of doing that I convert it into if-else if-else would be better.
I have changed the position of 
```
        else if (node.getType() == TokenTypes.RESOURCE) {
            localVariableDef = node.getChildCount() > 1;
        }

        // catch parameter?
        else if (node.getType() == TokenTypes.PARAMETER_DEF) {
            final DetailAST parent = node.getParent();
            localVariableDef = parent.getType() == TokenTypes.LITERAL_CATCH;
        }
```

the reason is 
```
        else if (node.getType() == TokenTypes.RESOURCE) {
            localVariableDef = node.getChildCount() > 1;
        }
```
condition inside it which is assigned to `localVariableDef`.

if we follow mutation then also it might not possible to find test using regression as well.

-------

# Regression 
Diff Regression config: https://gist.githubusercontent.com/Kevin222004/811066dafa00871bdf97dc6ec3d58962/raw/863854d2d20d1a8a493a34c237e0452375d6f34a/su1.xml
Report label: Regression-2


Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

nrmancuso:
>if we follow mutation then also it might not possible to find test using regression as well.

Why "might not be possible"? What is preventing us from hardcoding this mutation and generating tests?

romani:
@Kevin222004 , please reply question above. If you already tried, share a link to PR where you tried.

romani (approved):
After reading code closely, it is clear that old code just did 3 matching of node.getType() to specific token type. So it is clear switch or sequential if-else.

I am ok to merge

Kevin222004:
https://github.com/checkstyle/checkstyle/pull/13479#issuecomment-1656812049 
@nrmancuso I have started regression after hardcoding mutation at https://github.com/checkstyle/checkstyle/pull/13485.

The thing I was trying to mention here is that this mutation is happening because of its design and the way the how conditions are there.

rdiachenko:
@Kevin222004 please resolve a conflict

Kevin222004:
Conflict resolved

rdiachenko (approved):
lgtm

PR #13398, opened 2023-07-15
minor: update pitest-xpath profile

Kevin222004:
minor: update pitest-xpath profile

PR #13492, opened 2023-07-31
Pull #13492: Fix Pitest script to recognize unstable mutations

Vyom-Yadav:
From https://github.com/checkstyle/checkstyle/pull/13398#discussion_r1265538825

### DIFF
--- a/.ci/pitest-survival-check-xml.groovy
+++ b/.ci/pitest-survival-check-xml.groovy
@@ -267,7 +267,7 @@ private static int printComparisonToConsole(Set<Mutation> survivingMutations,
         println 'No new surviving mutation(s) found.'
     }
     else if (survivingUnsuppressedMutations.isEmpty()
-            && !hasOnlyStableMutations(extraSuppressions)) {
+            && hasOnlyUnstableMutations(extraSuppressions)) {
         exitCode = 0
         println 'No new surviving mutation(s) found.'
     }
@@ -293,13 +293,13 @@ private static int printComparisonToConsole(Set<Mutation> survivingMutations,
 }
 
 /**
- * Whether a set has only stable mutations.
+ * Whether a set has only unstable mutations.
  *
  * @param mutations A set of mutations
- * @return {@code true} if a set has only stable mutations
+ * @return {@code true} if a set has only unstable mutations
  */
-private static boolean hasOnlyStableMutations(Set<Mutation> mutations) {
-    return mutations.every { !it.isUnstable() }
+private static boolean hasOnlyUnstableMutations(Set<Mutation> mutations) {
+    return mutations.every { it.isUnstable() }
 }
 
 /**
--- a/config/pitest-suppressions/pitest-xpath-suppressions.xml
+++ b/config/pitest-suppressions/pitest-xpath-suppressions.xml
@@ -9,15 +9,6 @@
     <lineContent>return Collections.unmodifiableList(siblings.subList(0, indexAmongSiblings));</lineContent>
   </mutation>
 
-  <mutation unstable="false">
-    <sourceFile>AbstractElementNode.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.xpath.AbstractElementNode</mutatedClass>
-    <mutatedMethod>iterateAxis</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.RemoveSwitchMutator_8</mutator>
-    <description>RemoveSwitch 8 (case value 8)</description>
-    <lineContent>switch (axisNumber) {</lineContent>
-  </mutation>
-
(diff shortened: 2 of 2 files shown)


rnveach:
https://github.com/checkstyle/checkstyle/actions/runs/5718351146/job/15493978082?pr=13492
Shows this will not work with comments.

https://github.com/checkstyle/checkstyle/actions/runs/5718351146/job/15493981725?pr=13492#step:7:51
Shows this will not work with unstable mutations.

rnveach:
This change should be in the groovy script instead and it looks like it is.
https://github.com/checkstyle/checkstyle/blob/master/.ci/pitest-survival-check-xml.groovy#L283

It needs to be looked into why this piece of the script didn't catch the issue.

.ci/pitest-survival-check-xml.groovy
```diff
@@ -293,13 +293,13 @@ private static int printComparisonToConsole(Set<Mutation> survivingMutations,
...
+ * Whether a set has only unstable mutations.
  *
  * @param mutations A set of mutations
- * @return {@code true} if a set has only stable mutations
+ * @return {@code true} if a set has only unstable mutations
  */
-private static boolean hasOnlyStableMutations(Set<Mutation> mutations) {
-    return mutations.every { !it.isUnstable() }
+private static boolean hasOnlyUnstableMutations(Set<Mutation> mutations) {
+    return mutations.every { it.isUnstable() }
```
Vyom-Yadav, line 302: The problem was that `!hasOnlyStableMutations != hasOnlyUnstableMutations`, `!hasOnlyStableMutations` could mean that it has both stable and unstable and having unstable made `hasOnlyStableMutations` return false, which made `!hasOnlyStableMutations` true. BUT this does not mean that it only has unstable, it could be a mix of both.

Vyom-Yadav:
> https://github.com/checkstyle/checkstyle/actions/runs/5718351146/job/15493978082?pr=13492 Shows this will not work with comments.
> 
> https://github.com/checkstyle/checkstyle/actions/runs/5718351146/job/15493981725?pr=13492#step:7:51 Shows this will not work with unstable mutations.
> 
> You can't compare a custom made file with a generated file, as the custom changes will also be pinged as differences.

@rnveach Yes, this is a problem. The current automatic fixing system has these limitations. Its algorithm is simple; it replaces the file contents with the surviving mutations. 

Instead, we should impose a fine-grained automatic survival addition/removal system that only removes extra mutations with the comment directly above it or add new mutations. It should not interfere with unstable mutations and comments not related to that particular mutation.

nrmancuso:
My vision with the groovy based system was to serialize/deserialize the data from the file and and data from the report, and compare only the data so that we can link issues, have comments, etc. File generation and git diff will not work well with this method, nor should they have to.

Vyom-Yadav:
Agreed, we are at a stage where the mutations are stable (not a lot of addition or removal), so having some complicated logic for writing-to/removing-from suppression file isn't worth it.

> My vision with the groovy based system was to serialize/deserialize the data from the file and data from the report, and compare only the data so that we can link issues, have comments, etc.

Yes, this is currently what we have (+ blind write to file), and I think this is sufficient. 
We can vote for removing writing to suppression file + git diff (automatic patch).

romani (approved):
I agree with this update. We are not redesign anything and just fix defect.

Ok to merge.

nrmancuso (approved):

rnveach (approved):

rdiachenko (approved):
lgtm

rdiachenko:
rebased

romani:
Pitest failure:
```
diff --git a/config/pitest-suppressions/pitest-common-suppressions.xml b/config/pitest-suppressions/pitest-common-suppressions.xml
index 1ed5890..3b8d487 100644
--- a/config/pitest-suppressions/pitest-common-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-suppressions.xml
@@ -45,6 +45,[15](https://github.com/checkstyle/checkstyle/actions/runs/6176878912/job/16766769295?pr=13492#step:7:16) @@
     <lineContent>final String stripped = CommonUtil.relativizeAndNormalizePath(basedir, fileName);</lineContent>
   </mutation>
 
+    <mutatedMethod>finishLocalSetup</mutatedMethod>
+    <description>removed call to com/puppycrawl/tools/checkstyle/LocalizedMessage::setLocale</description>
+    <lineContent>LocalizedMessage.setLocale(locale);</lineContent>
+  </mutation>
+
   <mutation unstable="false">
     <sourceFile>Checker.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.Checker</mutatedClass>
```

Vyom-Yadav:
@romani It's green

romani:
Restarted https://github.com/checkstyle/checkstyle/actions/runs/6176878912/job/16779176282

PR #13495, opened 2023-08-01
infra: remove version check for groovy .travis.yml

romani:


.travis.yml
```diff
@@ -41,7 +41,7 @@ before_script:
         export JAVA_HOME=/usr/lib/jvm/adoptopenjdk-11-hotspot-ppc64el
         export PATH=$JAVA_HOME/bin:$PATH
     fi
-  - groovy --version
+  # - groovy --version
```
Vyom-Yadav, line 44: The problem is with `JAVA_HOME` not being correctly set up.

romani:
Yes, I just tested an easy way to disable it or make it always green.

Vyom-Yadav:
> IMO we need to stop using Travis completely, it is not worth the effort to maintain. We have plenty of reliable CI.

Do we want to migrate to existing ones like GA, CircleCI, etc? If we are looking for new ones, then GitLab CI can be a viable option. They got plans for open source https://about.gitlab.com/solutions/open-source/join/

romani:
I doubt they can allow use them just as CI, they need hosting of sources.

Vyom-Yadav:
> I doubt they can allow use them just as CI, they need hosting of sources.

There is a dedicated option to use GitLab CI for externally hosted projects which automatically creates a mirror from GitHub.

rnveach:
> for externally hosted projects which automatically creates a mirror from GitHub

But this will not work for PRs, right?

Vyom-Yadav:
I will create a local copy of Checkstyle to demonstrate and check it.

PR #13505, opened 2023-08-03
Pull #13505: Restore usage of TreeSet

stoyanK7:
As suggested in https://github.com/checkstyle/checkstyle/pull/13423#discussion_r1281289602

>  please restore copy pasta and add a comment about why we use treeset

In https://github.com/checkstyle/checkstyle/pull/13423#discussion_r1275385868 there is a thread about a chunk of code that is a direct copy of those two:

https://github.com/checkstyle/checkstyle/blob/ec5cd7e2ffd7c5c3a1b2cd4e28e4d3c37bb723fc/src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/CheckUtil.java#L211-L213
https://github.com/checkstyle/checkstyle/blob/0c2ab5338db569b4699d5d20121d1bd508394eb9/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1381-L1389

There was some uncertainty as to why `TreeSet` is used, but @rnveach explained that in https://github.com/checkstyle/checkstyle/pull/13423#discussion_r1281288043

> You can't use a collection that is based on hashcode. It is JVM specific and different runs can generate different hashes and orders. We can't use insertion order here because reflection is not guaranteed an order and we want to ensure our documentation is consistent in order. It will be bad if one run says A,B,C and then another run says C,A,B. This is why we originally used TreeSet, to say order is alphabetic.
I recommend keeping this a copy/paste and just create new issues to fix anything.

This PR restores the check message keys extraction process to use TreeSet.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
@@ -20,10 +20,10 @@
 package com.puppycrawl.tools.checkstyle.site;
 
 import java.lang.reflect.Field;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.HashSet;
 import java.util.Locale;
 import java.util.Set;
+import java.util.TreeSet;
 
 import org.apache.maven.doxia.macro.MacroExecutionException;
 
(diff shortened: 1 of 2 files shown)


stoyanK7:
I will open new issue as suggested in https://github.com/checkstyle/checkstyle/pull/13423#discussion_r1281288043 after this is merged.

> I recommend keeping this a copy/paste and just create new issues to fix anything.

But isn't the current way of having a list and sorting it good enough?

romani (approved):
Ok to merge

PR #13513, opened 2023-08-04
Issue #13328: remove OBJBLOCK scope from UnusedLocalVariableCheck

rdiachenko:
Temp PR to test whether we need  OBJBLOCK scope

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
@@ -155,7 +155,6 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
     private static final int[] SCOPES = {
         TokenTypes.SLIST,
         TokenTypes.LITERAL_FOR,
-        TokenTypes.OBJBLOCK,
     };
 
     /**


nrmancuso:
New test input found: https://github.com/openjdk/jdk17/blob/bb3d226a4eedb3d3ef73ef91683f45f1db43a74e/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java#L93

romani:
@Kevin222004 , please help to move this input file to our source base, we should not rely on regression report too much.

Kevin222004:
@rdiachenko @romani I will add this test case.

But this will not change make any effect on mutation killing at https://github.com/checkstyle/checkstyle/pull/13151

romani:
Yes, not related to pitest killing. Just a fact to rely less on regression report.
And maybe in future, pitest will be able to catch such survivors.

romani:
Resolve.java is insanely big file, 5K lines.


exception:
```
Caused by: java.lang.NullPointerException
    at com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck.checkIdentifierAst (UnusedLocalVariableCheck.java:749)
    at com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck.visitDotToken (UnusedLocalVariableCheck.java:331)
    at com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck.visitToken (UnusedLocalVariableCheck.java:288)
    at com.puppycrawl.tools.checkstyle.TreeWalker.notifyVisit (TreeWalker.java:335)
    at com.puppycrawl.tools.checkstyle.TreeWalker.processIter (TreeWalker.java:406)
    at com.puppycrawl.tools.checkstyle.TreeWalker.walk (TreeWalker.java:273)
    at com.puppycrawl.tools.checkstyle.TreeWalker.processFiltered (TreeWalker.java:154)
   
```

code that trigger:
```
    private static void checkIdentifierAst(DetailAST identAst, Deque<VariableDesc> variablesStack) {
        for (VariableDesc variableDesc : variablesStack) {
            if (identAst.getText().equals(variableDesc.getName()) <!----- NPE here
```

Vyom-Yadav:
From discord chat: (https://discord.com/channels/845645228467159061/1137114234031968487/1138557270842675321)

None of the tests failed because `SCOPES` is only used at:
```
if (TokenUtil.isOfType(ast, SCOPES)) {
    logViolations(ast, variables);    
}                                     
```

and since variables under OBJBLOCK are not local, there is no chance of them adding/removing something to the violation list. I also checked whether their prescence in stack would effect other variables and the answer in 'no'.

If a OBJBLOCK scoped var is present in `variables` stack, then it must be an instance var in an anonymous inner class. Now we re-process the entire block (See `CONTAINERS_FOR_ANON_INNERS`) of anon inner class again. So even if it blocked some token by being on the top of stack in the first pass, in the second pass that won't be true (because `SCOPES` is only used on `leaveToken` for main pass) and everything would be processed correctly. 

Does it mean, it is not required? No, I would still advocate that we should keep it. Reasons:

1. Acts as a declaration of what scopes we allow.
2. Removes unnecessary entires from `variables` stack in the first pass.

With pitest we want to optimize the code, but not by removing every condition so it is only computationally correct but when a human tries to understand it, it blows their mind.

PR #13543, opened 2023-08-09
Issue #13167: Fix NPE in UnusedLocalVariable check

### LINKED ISSUE #13167: Null pointer Exception in UnusedLocalVariableCheck
I have read check documentation: https://checkstyle.org/config_coding.html#UnusedLocalVariable

```
kevin@inspiron-15-5510:~/Desktop/check_style$ cat Test.java
import javax.swing.*;
import java.awt.*;
import java.util.function.Supplier;

public class Test {
   private final NotNullLazyValue<JComponent> myComponent = NotNullLazyValue.lazy(() -> {
      JPanel results = new JPanel(new GridBagLayout()) {};
       return results;
   });
}
class NotNullLazyValue<T> {
    public NotNullLazyValue() {
    }
    public static <T> NotNullLazyValue<T> lazy(Supplier<T> supplier) {
        return new NotNullLazyValue<>();
    }
}

kevin@inspiron-15-5510:~/Desktop/check_style$ javac Test.java

kevin@inspiron-15-5510:~/Desktop/check_style$ cat config.xml 
<?xml version="1.0"?>

<module name="Checker">
    <property name="charset" value="UTF-8" />
    <property name="severity" value="warning" />
    <property name="haltOnException" value="false" />

    <module name="TreeWalker">

        <module name="UnusedLocalVariable">
        </module>

    </module>
</module>

kevin@inspiron-15-5510:~/Desktop/check_style$ java -jar /home/kevin/Downloads/checkstyle-10.12.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /home/kevin/Desktop/check_style/Test.java:1: Got an exception - java.lang.NullPointerException: Cannot invoke "com.puppycrawl.tools.checkstyle.api.DetailAST.getType()" because "parentAst" is null
        at com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck.getBlockContainingLocalAnonInnerClass(UnusedLocalVariableCheck.java:480)
        at com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck.visitLocalAnonInnerClass(UnusedLocalVariableCheck.java:404)
        at com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck.visitToken(UnusedLocalVariableCheck.java:312)
        at com.puppycrawl.tools.checkstyle.api.AbstractFileSetCheck.process(AbstractFileSetCheck.java:98)
        at com.puppycrawl.tools.checkstyle.Checker.processFile(Checker.java:334)
        at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:293)
        at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:223)
        at com.puppycrawl.tools.checkstyle.Main.runCheckstyle(Main.java:415)
        at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:338)
        at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:195)
        at com.puppycrawl.tools.checkstyle.Main.main(Main.java:130)
 [Checker]
Audit done.
kevin@inspiron-15-5510:~/Desktop/check_style$ 
```
---

# Describe what you expect in detail.
Should Not throw Null pointer Exception

---

Vyom-Yadav:
Resolves #13167

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
@@ -413,15 +413,17 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
     private static DetailAST getBlockContainingLocalAnonInnerClass(DetailAST literalNewAst) {
         DetailAST currentAst = literalNewAst;
         DetailAST result = null;
-        while (!TokenUtil.isOfType(currentAst, CONTAINERS_FOR_ANON_INNERS)) {
-            if (currentAst.getType() == TokenTypes.LAMBDA
-                    && currentAst.getParent()
-                    .getParent().getParent().getType() == TokenTypes.OBJBLOCK) {
-                result = currentAst;
-                break;
+        DetailAST topMostLambdaAst = null;
+        while (currentAst != null && !TokenUtil.isOfType(currentAst, CONTAINERS_FOR_ANON_INNERS)) {
+            if (currentAst.getType() == TokenTypes.LAMBDA) {
+                topMostLambdaAst = currentAst;
             }
             currentAst = currentAst.getParent();
             result = currentAst;
+        }
+
+        if (currentAst == null) {
+            result = topMostLambdaAst;
         }
         return result;
     }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unusedlocalvariable/InputUnusedLocalVariableLambdaExpression.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unusedlocalvariable/InputUnusedLocalVariableLambdaExpression.java
@@ -1,0 +1,92 @@
+/*
+UnusedLocalVariable
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.unusedlocalvariable;
+
+import java.util.function.Supplier;
+
+public class InputUnusedLocalVariableLambdaExpression {
+    private final LambdaTest<String> myComponent = LambdaTest.lazy(() -> {
+        String foo = "";
+        String hoo = ""; // violation
+        new Runnable() {
+            String hoo = "";
+
+            @Override
+            public void run() {
+                String j = hoo; // violation
+                String ja = foo;
+                ja += "asd";
+            }
+        };
+        return "";
+    });
+
+    final LambdaTest<String> nestedLambdas = LambdaTest.lazy(() -> {
+        String foo = "";
+        String hoo = "";
+        String hoo2 = ""; // violation
+        String hoo3 = ""; // violation
+        final LambdaTest<String> myComponent = LambdaTest.lazy(() -> { // violation
+            final LambdaTest<String> myComponent3 = LambdaTest.lazy(() -> { // violation
+                new Runnable() {
+                    String hoo2 = "";
+
+                    @Override
+                    public void run() {
+                        String j = hoo; // violation
+                        String ja = foo;
+                        ja += hoo2;
+                    }
+                };
+                return "";
+            });
+            new Runnable() {
+                String hoo3 = "";
+
+                @Override
+                public void run() {
+                    String j = hoo3; // violation
+                    String ja = foo;
+                    ja += "asd";
+                }
+            };
+            return "";
+        });
+        new Runnable() {
+            String hoo2 = "";
+
+            @Override
+            public void run() {
+                String j = hoo2;
+                String ja = foo; // violation
+                j += hoo2;
+            }
+        };
+        return "";
+    });
+
+    final LambdaTest<String> nestedLambdas2 = LambdaTest.lazy(() -> {
+        String k = ""; // violation
+        final LambdaTest<String> nestedLambdas = LambdaTest.lazy(() -> {
+            new LambdaTest<String>() {
+                void foo() {
+                    String j = k;
+                    j += "asd";
+                }
+            };
+            return "";
+        });
+        return nestedLambdas.toString();
+    });
+}
+
+class LambdaTest<T> {
+    String k = "";
+    public static <T> LambdaTest<T> lazy(Supplier<T> supplier) {
+        return new LambdaTest<>();
+    }
+}
--- a/config/pitest-suppressions/pitest-coding-2-suppressions.xml
+++ b/config/pitest-suppressions/pitest-coding-2-suppressions.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressedMutations>
+  <!-- until https://github.com/checkstyle/checkstyle/issues/14894 -->
+  <mutation unstable="false">
+    <sourceFile>UnusedLocalVariableCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck</mutatedClass>
+    <mutatedMethod>getBlockContainingLocalAnonInnerClass</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NegateConditionalsMutator</mutator>
+    <description>negated conditional</description>
+    <lineContent>if (currentAst.getType() == TokenTypes.LAMBDA) {</lineContent>
+  </mutation>
+
+  <mutation unstable="false">
+    <sourceFile>UnusedLocalVariableCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck</mutatedClass>
+    <mutatedMethod>getBlockContainingLocalAnonInnerClass</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
+    <description>removed conditional - replaced equality check with true</description>
+    <lineContent>if (currentAst.getType() == TokenTypes.LAMBDA) {</lineContent>
+  </mutation>
+
+  <mutation unstable="false">
+    <sourceFile>UnusedLocalVariableCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck</mutatedClass>
+    <mutatedMethod>isInsideLocalAnonInnerClass</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NegateConditionalsMutator</mutator>
+    <description>negated conditional</description>
+    <lineContent>if (currentAst.getType() == TokenTypes.SLIST) {</lineContent>
+  </mutation>
+
+  <mutation unstable="false">
+    <sourceFile>UnusedLocalVariableCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck</mutatedClass>
+    <mutatedMethod>isInsideLocalAnonInnerClass</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
+    <description>removed conditional - replaced equality check with true</description>
+    <lineContent>if (currentAst.getType() == TokenTypes.SLIST) {</lineContent>
+  </mutation>
+</suppressedMutations>
--- a/config/spotbugs-exclude.xml
+++ b/config/spotbugs-exclude.xml
@@ -315,4 +315,11 @@
     </Or>
     <Bug pattern="CT_CONSTRUCTOR_THROW"/>
   </Match>
+  <Match>
+    <!-- variable is just being saved for further analysis and isn't the main decider
+         of the loop if it is done or not -->
+    <Class name="com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck"/>
+    <Method name="getBlockContainingLocalAnonInnerClass"/>
+    <Bug pattern="SLS_SUSPICIOUS_LOOP_SEARCH"/>
+  </Match>
 </FindBugsFilter>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheckTest.java
@@ -328,6 +328,36 @@ public class UnusedLocalVariableCheckTest extends AbstractModuleTestSupport {
     }
 
     @Test
+    public void testUnusedLocalVarLambdas() throws Exception {
+        final String[] expected = {
+            "14:9: " + getCheckMessage(MSG_UNUSED_LOCAL_VARIABLE, "hoo"),
+            "20:17: " + getCheckMessage(MSG_UNUSED_LOCAL_VARIABLE, "j"),
+            "31:9: " + getCheckMessage(MSG_UNUSED_LOCAL_VARIABLE, "hoo2"),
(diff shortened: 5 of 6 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
```diff
@@ -462,15 +462,34 @@ private static boolean isNonLocalTypeDeclaration(DetailAST typeDeclAst) {
...
+     * @param ast ast node
+     * @return {@code true} if ast is nested under tokens of type
+     *         {@link UnusedLocalVariableCheck#CONTAINERS_FOR_ANON_INNERS}
+     *         or {@link TokenTypes#LAMBDA}
+     */
+    public static boolean isTokenNestedUnderContainerForAnonInners(DetailAST ast) {
+        boolean result = false;
+        DetailAST parentAst = ast.getParent();
+        while (parentAst != null) {
+            if (TokenUtil.isOfType(parentAst, CONTAINERS_FOR_ANON_INNERS)
```
nrmancuso, line 487: Please rename this collection to something like `CLASS_BODY_DECLARATION_TOKENS` in a separate commit first, `container` is nonsense in this context. Then update the method naming, etc. appropriately.

nrmancuso (changes requested):
Item:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unusedlocalvariable/InputUnusedLocalVariableLambdaExpression.java
```diff
@@ -0,0 +1,92 @@
...
+public class InputUnusedLocalVariableLambdaExpression {
+    private final LambdaTest<String> myComponent = LambdaTest.lazy(() -> {
+        String foo = ""; // ok
+        String hoo = ""; // violation
+        new Runnable() {
+            String hoo = "";
+
+            @Override
+            public void run() {
+                String j = hoo; // violation
```
romani, line 20: Please define violation messages

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
romani: @rdiachenko , please step in to help review.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unusedlocalvariable/InputUnusedLocalVariableLambdaExpression.java
Vyom-Yadav: We do not define violation messages when the check has only one kind of violation message.

romani: Hmm, I thought We started in all places to make it clear that violation message meaningful on a line, and it doesn't not create extra effort on new message addition.

@rnveach , do you remember what is our rule for this now?

nrmancuso: We should have violation message for all violations, regardless of number of messages in check. This is a good sanity check, and will also alleviate extra work in the future if we add a new message.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
```diff
@@ -462,15 +462,34 @@ private static boolean isNonLocalTypeDeclaration(DetailAST typeDeclAst) {
...
             }
             currentAst = currentAst.getParent();
-            result = currentAst;
+        }
+        return result;
+    }
+
+    /**
+     * Whether the ast is nested under tokens of type
+     * {@link UnusedLocalVariableCheck#CONTAINERS_FOR_ANON_INNERS} or {@link TokenTypes#LAMBDA}.
```
nrmancuso, line 476: Please remove mentions of specific tokens, such a list is bound to get outdated/ignored, and is an implementation detail that the caller should not have to worry about.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
```diff
@@ -462,15 +462,34 @@ private static boolean isNonLocalTypeDeclaration(DetailAST typeDeclAst) {
...
+     *
+     * @param ast ast node
+     * @return {@code true} if ast is nested under tokens of type
+     *         {@link UnusedLocalVariableCheck#CONTAINERS_FOR_ANON_INNERS}
+     *         or {@link TokenTypes#LAMBDA}
```
nrmancuso, line 481: Same as above.

nrmancuso (changes requested):
Another item as you push changes for collection:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unusedlocalvariable/InputUnusedLocalVariableLambdaExpression.java
rnveach: @romani @nrmancuso If it was written down somewhere (either in code or the top of issues) or controlled by CI, then it doesn't exist. If this should be a rule, then a test should enforce it.

https://github.com/checkstyle/checkstyle/issues/11214
> Single Check might have multiple violations messages, we need to see exact violation message if Check has multiple messages, to be sure that violation of correct type is placed at right line.

I consider this the main issue for this and it is documented only for checks with multiple messages. I scanned some of the checks listed and did not see any single message checks.

https://github.com/checkstyle/checkstyle/blob/2335abaea101774a23d0bdaf55f19e4f0cad0d37/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L118-L119
It is even documented in code that it is only for checks with multiple messages.

====================

I was not involved in the discussions or the decision to do this, so I don't know anything more than, as I said, what is written.

romani: Thanks a lot.
If code checks for not single message, let's keep it and Finish phase of migration to update multiple messages Checks first.
And maybe be later we will do same update for single-message, we will vote on this later on.

I recall this review item for now.

romani: Thanks a lot.
If code checks for not single message, let's keep it and Finish phase of migration to update multiple messages Checks first.
And maybe be later we will do same update for single-message, we will vote on this later on.

I recall this review item for now.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
Vyom-Yadav: Created https://github.com/checkstyle/checkstyle/pull/14090 to change the variable name.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
```diff
@@ -397,16 +397,18 @@ private static boolean isNonLocalTypeDeclaration(DetailAST typeDeclAst) {
...
-        while (!TokenUtil.isOfType(currentAst, CONTAINERS_FOR_ANON_INNERS)) {
-            if (currentAst.getType() == TokenTypes.LAMBDA
-                    && currentAst.getParent()
-                    .getParent().getParent().getType() == TokenTypes.OBJBLOCK) {
-                result = currentAst;
-                break;
+        DetailAST topMostLambdaAst = null;
+        while (currentAst != null && !TokenUtil.isOfType(currentAst, CONTAINERS_FOR_ANON_INNERS)) {
+            if (currentAst.getType() == TokenTypes.LAMBDA) {
+                topMostLambdaAst = currentAst;
```
Vyom-Yadav, line 403: This is a time-complexity mutation. The reason I have added this to the suppressions list is because removing this condition will make the entire `COMPILATION_UNIT` to be processed again. And having many nested lambdas can make the entire `COMPILATION_UNIT` be processed multiple times.

If the penalty was small, I would have inclined towards pitest, but processing the entire `COMPILATION_UNIT` multiple times isn't a wise option IMO.

rnveach:
Regression for this PR: https://rveach.no-ip.org/checkstyle/regression/345/

1 NPE is removed, no other changes.

rnveach:
As discussed, analysis is right at https://github.com/checkstyle/checkstyle/pull/13543#discussion_r1412796525 . This would be a very complex check, which I am unfamilar with, to fix pitest.

If we are good, I can write a reasoning for the suppression and create a new issue to resolve pitest suppressions. Maybe it can be done as part of the GSoC project.

rnveach:
It seems to me this check has a check inside itself. We have [the first set of visits which is collecting information](https://github.com/checkstyle/checkstyle/blob/55f5c966bfa61b387110db6c4780f259a21676c8/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L221-L255), then [we have a custom tree walker with its own visits using the information collected from the first](https://github.com/checkstyle/checkstyle/blob/55f5c966bfa61b387110db6c4780f259a21676c8/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L614-L627). It is not re-walking the entire tree in both times, but this is what makes this check complex by itself.

Vyom-Yadav:
Yeah, I remember some of these mutations to be like that. One mutation was particularly like processing the whole `COMPILATION_UNIT` again, so the statement that pitest suggests is actually to prevent that from happening. It's a shame I never got the time to complete this 😞.

 
> It seems to me this check has a check inside itself. We have [the first set of visits which is collecting information](https://github.com/checkstyle/checkstyle/blob/55f5c966bfa61b387110db6c4780f259a21676c8/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L221-L255), then [we have a custom tree walker with its own visits using the information collected from the first](https://github.com/checkstyle/checkstyle/blob/55f5c966bfa61b387110db6c4780f259a21676c8/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L614-L627). It is not re-walking the entire tree in both times, but this is what makes this check complex by itself.

Yeah, it does that. The reason for two traversals is that a class may be defined further down the AST but may be used before, so scoping, etc., becomes difficult in that scenario, particularly with anon inner classes.

rdiachenko:
@Vyom-Yadav 

> And having many nested lambdas can make the entire `COMPILATION_UNIT` be processed multiple times.

Do you have any real examples with many nested lambdas? I'm interested in whether this is a genuine concern or just something we think is problematic when, in practice, it isn't.

> If the penalty was small, I would have inclined towards pitest, but processing the entire `COMPILATION_UNIT` multiple times isn't a wise option IMO.

What's the penalty? Do you have some numbers at hand?

Vyom-Yadav:
> Do you have any real examples with many nested lambdas? I'm interested in whether this is a genuine concern or just something we think is problematic when, in practice, it isn't.

I don't know if there is some pattern using nested lambdas. It is not a rare case, that I'm sure. 

> What's the penalty? Do you have some numbers at hand?

Penalty is directly proportional to the number of nodes in the ast. It will traverse the entire ast again (maybe multiple times too).

rdiachenko:
I didn't find a quick and clear way to test Checkstyle performance for the fix with/without optimisation, so I ended up running both versions against the following list of projects via regression diff report:

```
checkstyle|git|https://github.com/checkstyle/checkstyle.git|master|...|
sevntu-checkstyle|git|https://github.com/sevntu-checkstyle/sevntu.checkstyle|master||
checkstyle-sonar|git|https://github.com/checkstyle/sonar-checkstyle|master||
guava|git|https://github.com/google/guava|v28.2||

spotbugs|git|https://github.com/spotbugs/spotbugs|3.1.2||
pmd|git|https://github.com/pmd/pmd|pmd_releases/6.21.0|**/pmd/pmd-java/src/test/**/*,**/pmd/cpd/files/*
spoon|git|https://github.com/INRIA/spoon.git|spoon-core-10.1.0|**/src/test/resources/**/*
lombok-ast|git|https://github.com/rzwitserloot/lombok.ast|v0.2|**/lombok-ast/test/**/*

spring-framework|git|https://github.com/spring-projects/spring-framework|v4.1.6.RELEASE||
hibernate-orm|git|https://github.com/hibernate/hibernate-orm|4.2.19.Final|**/hibernate-orm/documentation/**/*
elasticsearch|git|https://github.com/elastic/elasticsearch|v1.5.2||
java-design-patterns|git|https://github.com/iluwatar/java-design-patterns|dd855a376bc025aa61f6816584f79eb9854fe5d7||
MaterialDesignLibrary|git|https://github.com/navasmdc/MaterialDesignLibrary|1.3||
Hbase|git|https://github.com/apache/hbase|1.1.0.1||
Orekit|git|https://github.com/CS-SI/Orekit|8.0.1||

# Projects which contain a lot of labmda expressions
infinispan|git|https://github.com/infinispan/infinispan|7.2.5.Final||
protonpack|git|https://github.com/poetix/protonpack|protonpack-1.7||
jOOL|git|https://github.com/jOOQ/jOOL|version-0.9.7||
RxJava|git|https://github.com/ReactiveX/RxJava|v1.0.9||
Vavr|git|https://github.com/vavr-io/vavr|v0.9.0||
```

For this check's config:

```
...
    <module name="TreeWalker">
         <module name="UnusedLocalVariable">
         </module>
    </module>
...
```

I thought if we had a performance problem, the unoptimised fix would give x-times higher processing time. Here are the actual numbers (I ran the regression diff report 3 times for each build and choose the median):

```
# optimised fix
time groovy diff.groovy -r /var/tmp/checkstyle \
-b master -p fixNullPointerExInUnusedLocalVar -c my_check.xml -l projects-to-test-on.properties

1434.43s user 71.09s system 225% cpu 11:08.89 total

# not optimised fix
time groovy diff.groovy -r /var/tmp/checkstyle \
-b master -p fixNullPointerExInUnusedLocalVar-wo-optim -c my_check.xml -l projects-to-test-on.properties

1417.58s user 70.16s system 224% cpu 11:03.38 total
```

There is no difference. I suggest we remove optimisation and use a [simplified fix](https://github.com/rdiachenko/checkstyle/commit/becac625e959642a658fa0b49ce943ea6fd76365) that avoids two `if (currentAst.getType() == TokenTypes.LAMBDA)` mutations.

romani:
I would love to have no pitest suppression,  rather than nano optimization. In long run no pitest will be more beneficial.

romani:
@nrmancuso , @rnveach , please share your opinion.

rnveach:
I am ok for this PR as it is now or closing it.

I don't agree with applying the pitest mutations to the code since that would degrade this check more than it is. We should already look into the issue that it is reporting the same token 2x already.

nrmancuso:
> @nrmancuso , @rnveach , please share your opinion.

My opinion is that we have already spent too much effort on this pitest suppression, let’s either merge or close this. I think it is personally ridiculous to traverse the entire AST up to the root node when we know that we can (and should) stop when we reach the enclosing anonymous class.

We are humans that can think and reason, not robots that only deal in pitest logic.

romani:
All suppression in this PR smells badly.
https://github.com/checkstyle/checkstyle/issues/14894
we will unblock users from NPE, and will deal with better code later on.

romani (approved):
Ok to merge.
We will improve in separate PR

Issue #13501, opened 2023-08-02
Resolve pitest survivals in modules - 3


Kevin222004:
Split from 13321
Part of https://github.com/checkstyle/checkstyle/issues/13089

--------------

**1) JavaParser**

- [x]  NullReturnValsMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13655
- [x]  NakedReceiverMutator (1)
          https://github.com/checkstyle/checkstyle/pull/13658
------

**2) DetailAstImpl**

- [x]  NonVoidMethodCallMutator (2)
        https://github.com/checkstyle/checkstyle/pull/13652 (Cover 2 Mutations)
- [x]  MemberVariableMutator (2)
       https://github.com/checkstyle/checkstyle/pull/13654   
https://github.com/checkstyle/checkstyle/commit/fe0df9c8b2472b720342e2255851fb37ceb11de2

-----

**7) JavadocTypeCheck**

- [x] NonVoidMethodCallMutator (2)
        https://github.com/checkstyle/checkstyle/pull/13562 
        https://github.com/checkstyle/checkstyle/pull/13563
- [x]  MemberVariableMutator (1)
         https://github.com/checkstyle/checkstyle/pull/13571
- [x]  RemoveConditionalMutator_EQUAL_IF (4)
        https://github.com/checkstyle/checkstyle/pull/13562 (Cover 3 Mutation)
         https://github.com/checkstyle/checkstyle/pull/13563
- [x] NakedReceiverMutator (1)
          https://github.com/checkstyle/checkstyle/pull/13564

------

**9) AtclauseOrderCheck**

- [x]  NakedReceiverMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13515
        

------

**11) HtmlTag**

- [x]  ArgumentPropagationMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13516
        
-------

**12) JavadocContentLocationCheck**

- [x]  MemberVariableMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13517
        
------

**13) JavadocParagraphCheck**

- [x]  RemoveConditionalMutator_EQUAL_ELSE (3)
        https://github.com/checkstyle/checkstyle/pull/13590
- [x]  RemoveConditionalMutator_EQUAL_IF (1)
        https://github.com/checkstyle/checkstyle/pull/13590

**16) JavadocTagContinuationIndentationCheck**

- [x]  ArgumentPropagationMutator (1)
[...]

PR #13549, opened 2023-08-14
Issue #13501: Kill mutation for MultilineDetector

Kevin222004:
Issue #13501: Kill mutation for MultilineDetector

-----

# Mutation
https://github.com/checkstyle/checkstyle/blob/8134680099d82c13c8231ca1b96048a01773a356/config/pitest-suppressions/pitest-regexp-suppressions.xml#L4-L11


src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/MultilineDetector.java
```diff
@@ -116,7 +116,7 @@ private void findMatch() {
             // OK http://blog.igorminar.com/2008/05/catching-stackoverflowerror-and-bug-in.html
             // http://programmers.stackexchange.com/questions/
             //        209099/is-it-ever-okay-to-catch-stackoverflowerror-in-java
-            options.getReporter().log(1, MSG_STACKOVERFLOW, matcher.pattern().toString());
+            options.getReporter().log(1, MSG_STACKOVERFLOW);
         }
```
Kevin222004, line 120: These changes are actually wrong https://github.com/checkstyle/checkstyle/blob/8134680099d82c13c8231ca1b96048a01773a356/src/main/resources/com/puppycrawl/tools/checkstyle/checks/regexp/messages.properties#L8
because we need to mention the pattern here is required

But I think we can cover this only while we have some real input. I have tried some way on my local but it doesn't work

Kevin222004: I am not sure whether this type of case is possible in real code

romani: It was added at this commit https://github.com/checkstyle/checkstyle/commit/6f089534ea0b9680f0c76bc1fb3c04d11954a278

Please recheck PR and issue

romani: Do we still have test that was created originally? Recheck git history if it was removed or updated.

Can we extend it to validate full message? It is ok cover this case by pure Junit test. It has no much value for users, it is extreme edge case.

Kevin222004: https://github.com/checkstyle/checkstyle/blob/8134680099d82c13c8231ca1b96048a01773a356/src/test/java/com/puppycrawl/tools/checkstyle/checks/regexp/RegexpMultilineCheckTest.java#L182-L196
Yes the test still exist.

romani: Can we extend message verification?

Kevin222004: image
Here, their is something fault in code itself I am trying to find the cause. why this is happening may be becuase of exception

romani: If you fail to extend, push in GitHub what you tried, to let mentors to help easier.

Kevin222004: I have not make any changes just change a minor change in exisiting test

I have update the branch

romani: Please fix https://github.com/checkstyle/checkstyle/blob/d832cae442e754f6d76729ff4794cb6c6302a971/src/main/resources/com/puppycrawl/tools/checkstyle/checks/regexp/messages.properties#L8
To use index of parameter 0 instead of 1.

```
[ERROR]   RegexpMultilineCheckTest.testNoStackOverflowError:196->AbstractModuleTestSupport.verify:312->AbstractModuleTestSupport.verify:350->AbstractModuleTestSupport.verify:372->AbstractModuleTestSupport.verify:402->AbstractModuleTestSupport.lambda$verify$1:405 Violations for /tmp/junit13558589523051257823/junit13264957617699487808.tmp differ.
value of: iterable.onlyElement()
expected: 1: java.util.regex.Matcher caused a java.lang.StackOverflowError for pattern (x|y)* . (you may be scanning a binary file instead of text?)
but was : 1: java.util.regex.Matcher caused a java.lang.StackOverflowError for pattern '{1}' (you may be scanning a binary file instead of text?).
```

romani:
@Kevin222004 , please fix index in message.

src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/MultilineDetector.java
Kevin222004: done

romani (approved):
Good catch of defect.

Ok to merge.

Vyom-Yadav (approved):
LGTM!

PR #13564, opened 2023-08-17
Issue #13501: Kill mutation for JavadocTypeCheck3

Kevin222004:
Issue #13501: Kill mutation for JavadocTypeCheck3

----

# Check
https://checkstyle.org/checks/javadoc/javadoctype.html#JavadocType

-------

# Mutation 
https://github.com/checkstyle/checkstyle/blob/d2c985ecdb5deb34ce5e535034060826d2f9bff4/config/pitest-suppressions/pitest-javadoc-suppressions.xml#L579-L586

We filter only params at https://github.com/checkstyle/checkstyle/blob/0fa8d8b3e951ddee26785a2af50c7ff7c159f7c3/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java#L619 so it is double check, not required.
-----

# Regression :- 

Report-1 :- https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/6d4c16b_2023215123/reports/diff/index.html

Report-2 :- https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/6d4c16b_2023234000/reports/diff/index.html

-----

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/e591a1cf82070108242c5569a7236a01/raw/9b17bf12fa20506af727bb7cae5c18b5b6dfe4ec/JavadocTypeCheck.xml
Report label: Regression-2

### DIFF
--- a/config/pitest-suppressions/pitest-javadoc-suppressions.xml
+++ b/config/pitest-suppressions/pitest-javadoc-suppressions.xml
@@ -576,14 +576,14 @@
     <lineContent>return astType == TokenTypes.VARIABLE_DEF</lineContent>
   </mutation>
 
-  <mutation unstable="false">
-    <sourceFile>JavadocTypeCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocTypeCheck</mutatedClass>
-    <mutatedMethod>checkTypeParamTag</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to java/util/stream/Stream::filter with receiver</description>
-    <lineContent>.filter(JavadocTag::isParamTag)</lineContent>
-  </mutation>
+
+
+
+
+
+
+
+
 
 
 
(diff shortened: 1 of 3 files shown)


Kevin222004:
Github, generate report

Kevin222004:
Github, generate report

Kevin222004:
Non ast based check

romani (approved):
Ok to merge

rdiachenko (approved):
lgtm

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
```diff
@@ -739,7 +739,6 @@ private void checkTypeParamTag(DetailAST ast,
 
         final boolean found = tags
             .stream()
-            .filter(JavadocTag::isParamTag)
```
Vyom-Yadav, line 742: > We filter only params at .... so it is double check, not required.

Nope, it is not. At ` CheckUtil.getTypeParameterNames(ast); ` we only get the parameter names from the Method or class, etc. We don't check anything regarding Javadoc.

This mutation can be killed by:
```java
/**                                                                   
 *                                                                    
 * @link <T>                                                          
 */                                                                   
protected class InnerPublic2<T> // violation 'missing @param <T> tag.'
{                                                                     
}                                                                     
```
If we remove `.filter(JavadocTag::isParamTag)` then it won't give a violation which is wrong. We just compare the first argument of the tag, it may or may not be the same as the value that `@param` is supposed to have.

Vyom-Yadav (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
Kevin222004: Thanks a lot @Vyom-Yadav for test case

romani (approved):
Ok to merge

Issue #13610, opened 2023-08-25
Use parent module macro in all templates


stoyanK7:
In https://github.com/checkstyle/checkstyle/pull/13435 we implemented the parent module macro and applied it only to a single file `src/xdocs/checks/whitespace/whitespaceafter.xml.template`.

All module xdocs now have templates and it is time to use this macro everywhere.

I will do this PR in chunks (per category) - annotation, blocks, naming, etc..

- [x] [Annotations](https://checkstyle.org/checks/annotation/index.html)
- [x]     [Block Checks](https://checkstyle.org/checks/blocks/index.html)
 - [x]    [Class Design](https://checkstyle.org/checks/design/index.html)
- [x]     [Coding](https://checkstyle.org/checks/coding/index.html)
- [x]     [Headers](https://checkstyle.org/checks/header/index.html)
- [x]     [Imports](https://checkstyle.org/checks/imports/index.html)
 - [x]    [Javadoc Comments](https://checkstyle.org/checks/javadoc/index.html)
- [x]     [Metrics](https://checkstyle.org/checks/metrics/index.html)
- [x]     [Miscellaneous](https://checkstyle.org/checks/misc/index.html)
- [x]     [Modifiers](https://checkstyle.org/checks/modifier/index.html)
- [x]     [Naming Conventions](https://checkstyle.org/checks/naming/index.html)
- [x]     [Regexp](https://checkstyle.org/checks/regexp/index.html)
- [x]     [Size Violations](https://checkstyle.org/checks/sizes/index.html)
- [x]     [Whitespace](https://checkstyle.org/checks/whitespace/index.html)
- [x] [Filters](https://checkstyle.org/filters/index.html)
- [x] [File Filters](https://checkstyle.org/filefilters/index.html)

stoyanK7:
Macro was applied to all templates

PR #13620, opened 2023-08-25
Issue #13610: Use parent module macro in modifiers templates

stoyanK7:
Part of #13610

We see no diff in generated files because formatting matches exactly


stoyanK7:
Github, generate site

romani (approved):

PR #13654, opened 2023-08-27
Issue #13501: Kill mutation for DetailAstImpl-7

Kevin222004:
Issue #13501: Kill mutation for DetailAstImpl-7

https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java

-------

# Mutations 
https://github.com/checkstyle/checkstyle/blob/88486f7ca9795f613d28fa6c2a53beccef521dfc/config/pitest-suppressions/pitest-common-2-suppressions.xml#L66-L73

------------

# Explaination

Test added

### DIFF
--- a/config/pitest-suppressions/pitest-common-2-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-2-suppressions.xml
@@ -63,14 +63,14 @@
 
 
 
-  <mutation unstable="false">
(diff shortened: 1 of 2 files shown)


romani (approved):
Ok to merge.

No input based test as we need to simulate tree creation with removal, but real logic do not remove children

PR #13658, opened 2023-08-28
Issue #13501: Kill mutation for JavaParser2

Kevin222004:
Issue #13501: Kill mutation for JavaParser2

https://github.com/Kevin222004/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java

-------

# Mutations 
https://github.com/checkstyle/checkstyle/blob/7007bef1f43a621757ae0b885818cf8baa0e6ef1/config/pitest-suppressions/pitest-common-2-suppressions.xml#L102-L109

------------

# Explaination

https://github.com/Kevin222004/checkstyle/blob/7007bef1f43a621757ae0b885818cf8baa0e6ef1/src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java#L140

The code is mutated as to remove `file.getAbsoluteFile()` --> `file` getAbsoluteFile this method will give the absolute path of file.

As per my research, the file from the parameter is already an absolute file path. I also don't think that test case is even possible because if we provide only filename then it will become exception. 

--------------

# Regression

### DIFF
--- a/config/pitest-suppressions/pitest-common-2-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-2-suppressions.xml
@@ -99,14 +99,14 @@
 
 
 
-  <mutation unstable="false">
-    <sourceFile>JavaParser.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.JavaParser</mutatedClass>
-    <mutatedMethod>parseFile</mutatedMethod>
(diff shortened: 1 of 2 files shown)


romani (approved):
Creation by File is more exact, that by absolute pass of it.

rdiachenko:
rebased

Issue #13693, opened 2023-09-04
Use module properties macro in all templates


romani:
macros is implemented at https://github.com/checkstyle/checkstyle/issues/13422
we need to use it in all templates

this is example of update we need to make for template files - https://github.com/checkstyle/checkstyle/pull/13681/files#diff-01387dd92ef7af5a9a1f274744b44e12fbeda8302eb79072785f7d6c9d559727L40

`mvn clean test -Dtest=XdocsPagesTest,XdocsJavaDocsTest`

romani:
We can not migrate modules that has tokens as property as xdoc validation demands them to be the last in table. 

But for now macros use simple alphabetical ordering

rnveach:
https://github.com/checkstyle/checkstyle/issues/13666 should be done first for each page before adding the macro to the page. The re-ordering should be a manual process of moving the property to a new position so we are sure the new macro isn't causing issues, as we found one.

nrmancuso:
Current state of migration:
```bash
➜  checkstyle git:(master)  for f in $(find . -name "*.xml.template"); do grep -l "<table>" $f | xargs basename 2>&1 | grep xml.template; done
customimportorder.xml.template
nonemptyatclausedescription.xml.template
writetag.xml.template
missingjavadocmethod.xml.template
summaryjavadoc.xml.template
javadocparagraph.xml.template
javadocstyle.xml.template
javadoctagcontinuationindentation.xml.template
javadoctype.xml.template
javadocvariable.xml.template
missingjavadoctype.xml.template
requireemptylinebeforeblocktaggroup.xml.template
singlelinejavadoc.xml.template
npathcomplexity.xml.template
javancss.xml.template
commentsindentation.xml.template
finalparameters.xml.template
orderedproperties.xml.template
todocomment.xml.template
trailingcomment.xml.template
uncommentedmain.xml.template
uniqueproperties.xml.template
arraytypestyle.xml.template
newlineatendoffile.xml.template
translation.xml.template
avoidescapedunicodecharacters.xml.template
indentation.xml.template
classmemberimpliedmodifier.xml.template
redundantmodifier.xml.template
interfacememberimpliedmodifier.xml.template
catchparametername.xml.template
illegalidentifiername.xml.template
packagename.xml.template
patternvariablename.xml.template
recordcomponentname.xml.template
recordtypeparametername.xml.template
abbreviationaswordinname.xml.template
constantname.xml.template
membername.xml.template
methodname.xml.template
[...]

PR #13694, opened 2023-09-04
Issue #13693: use property macros in few coding modules

romani:
Issue #13693

codign package files from top util defaultcomeslast

### DIFF
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/DeclarationOrderCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/DeclarationOrderCheck.xml
@@ -45,10 +45,10 @@
  &lt;/pre&gt;</description>
          <properties>
             <property default-value="false" name="ignoreConstructors" type="boolean">
-               <description>control whether to ignore constructors.</description>
+               <description>Control whether to ignore constructors.</description>
             </property>
             <property default-value="false" name="ignoreModifiers" type="boolean">
-               <description>control whether to ignore modifiers (fields, ...).</description>
+               <description>Control whether to ignore modifiers (fields, ...).</description>
             </property>
          </properties>
          <message-keys>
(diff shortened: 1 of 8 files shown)


src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/DeclarationOrderCheck.xml
```diff
@@ -45,10 +45,10 @@
  &lt;/pre&gt;</description>
          <properties>
             <property default-value="false" name="ignoreConstructors" type="boolean">
-               <description>control whether to ignore constructors.</description>
+               <description>Control whether to ignore constructors.</description>
             </property>
             <property default-value="false" name="ignoreModifiers" type="boolean">
-               <description>control whether to ignore modifiers (fields, ...).</description>
+               <description>Control whether to ignore modifiers (fields, ...).</description>
```
nrmancuso, line 51: If we are making improvements,  we need to reword this description, it is nonsense. `Configures check to ignore modifiers on target tokens when considering declaration order.` or something is much better.

nrmancuso (changes requested):
Item:

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/DeclarationOrderCheck.xml
romani: Technically we are not improving. 
We just migrate to macros as it, and this is just minor side effect.

I can do one more commit to change wording. As you suggested.

nrmancuso: Ok, I did not realize change in capitalization was from generation, we don’t need to do this.

romani: please proceed with review.

nrmancuso (approved):

Issue #13672, opened 2023-08-31
Resolve pitest survivals in modules - 4


Kevin222004:
Split from 13321
Part of https://github.com/checkstyle/checkstyle/issues/13089

--------------

**2) DetailAstImpl**

- [x]  NonVoidMethodCallMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13649
- [x]  MemberVariableMutator (3)
        https://github.com/checkstyle/checkstyle/pull/13391 (Cover 2 Mutation)
        https://github.com/checkstyle/checkstyle/pull/13649
        
------

**4) XmlLoader**

- [x] ArgumentPropagationMutator (1)
      https://github.com/checkstyle/checkstyle/pull/13800

------

**5) JavadocNodeImpl**

- [x] ArgumentPropagationMutator (2)
       https://github.com/checkstyle/checkstyle/pull/13340 (Cover 2)
- [x]  NakedReceiverMutator (1)
       https://github.com/checkstyle/checkstyle/pull/13340   
-----

**15) JavadocMissingLeadingAsteriskCheck**

- [x]  RemoveConditionalMutator_EQUAL_IF (1)
        https://github.com/checkstyle/checkstyle/pull/13663

----

**17) SummaryJavadocCheck**

- [x]  RemoveConditionalMutator_EQUAL_IF (1)
         https://github.com/checkstyle/checkstyle/pull/13329

----------

52) SuppressionXpathFilter
 
- [x]  NonVoidMethodCallMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13788

--------
 
53) SuppressionFilter
 
- [x]  MemberVariableMutator (2)
         https://github.com/checkstyle/checkstyle/pull/13788
- [x]  EmptyObjectReturnValsMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13788
        
-------
 
**56) TreeWalker**
 
- [x]  ArgumentPropagationMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13684
 --------

**63) LocalizedMessage**

- [x]  ArgumentPropagationMutator (1)
       https://github.com/checkstyle/checkstyle/pull/13800

------

**67) ConfigurationLoader**

- [x] NonVoidMethodCallMutator (9)
        https://github.com/checkstyle/checkstyle/pull/13299 (covered 8)
        https://github.com/checkstyle/checkstyle/pull/13801

--------

**75) AbstractRootNode**

[...]

PR #13181, opened 2023-06-08
Issue #13109: Kill mutation for FileText3

Kevin222004:
Issue #13109: Kill mutation for FileText3

---------

# Mutation Covered
https://github.com/checkstyle/checkstyle/blob/958051594a64ea5a228359919ed7426e69a242d7/config/pitest-suppressions/pitest-api-suppressions.xml#L12-L28

PR #13728, opened 2023-09-11
Issue #13672: Kill mutation related to Charset in FileText

Vyom-Yadav:
Issue #13672: Kill mutation related to Charset in FileText

originally discussed at https://github.com/checkstyle/checkstyle/pull/13181#discussion_r1308725119

---

# Mutation Covered
https://github.com/checkstyle/checkstyle/blob/958051594a64ea5a228359919ed7426e69a242d7/config/pitest-suppressions/pitest-api-suppressions.xml#L12-L28

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/NonEmptyAtclauseDescriptionCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/NonEmptyAtclauseDescriptionCheckTest.java
@@ -22,10 +22,16 @@ package com.puppycrawl.tools.checkstyle.checks.javadoc;
 import static com.google.common.truth.Truth.assertWithMessage;
 import static com.puppycrawl.tools.checkstyle.checks.javadoc.NonEmptyAtclauseDescriptionCheck.MSG_KEY;
 
+import java.nio.charset.CodingErrorAction;
+
 import org.junit.jupiter.api.Test;
 
 import com.puppycrawl.tools.checkstyle.AbstractModuleTestSupport;
+import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
+import com.puppycrawl.tools.checkstyle.TreeWalker;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
+import de.thetaphi.forbiddenapis.SuppressForbidden;
 
 public class NonEmptyAtclauseDescriptionCheckTest
         extends AbstractModuleTestSupport {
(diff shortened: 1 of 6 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/LineLengthCheckTest.java
```diff
@@ -118,4 +124,28 @@ public void testLineLengthIgnoringImportStatements() throws Exception {
...
+        checkConfig.addProperty("max", "1");
+
+        final DefaultConfiguration checkerConfig = createRootConfig(checkConfig);
+        checkerConfig.addProperty("charset", "IBM1098");
+
+        final Path tempFile = Files.createTempFile("InputLineLengthCharset.java", null);
+
+        try (OutputStream fileOutputStream = Files.newOutputStream(tempFile)) {
+            fileOutputStream.write(0x80);
+            fileOutputStream.write(0x80);
```
romani, line 144: we need a comment over test method to explain it.
`This test exists to make to show how we drop symbols that can not be mapped during loading by some encoding, example of unmapped symbols can be found at https://www.fileformat.info/info/charset/x-IBM1098/decode.htm 0x80, 0x81, 0xCF`


or better:
https://www.fileformat.info/info/charset/ISO-8859-6/decode.htm `a1 | (unmappable)` but  mappable at https://www.fileformat.info/info/charset/windows-1251/list.htm `Ў 	CYRILLIC CAPITAL LETTER SHORT U (U+040E) 	a1`
so we can create file that has visual character and it works with one charset and does not work in another charset. We can have single test method that just try run same file two time by different charsets.

romani (changes requested):
items:

src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/LineLengthCheckTest.java
Vyom-Yadav: Done, I added Javadoc comments and converted it to a non-temporary input file-based test.

romani (approved):
This is very good.

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/NonEmptyAtclauseDescriptionCheckTest.java
```diff
@@ -95,4 +101,30 @@ public void testCheckTwo() throws Exception {
...
+
+    /**
+     * This tests that the check does not fail when the input file contains unmappable characters.
+     * The test file contains the character ü which is not mappable to US-ASCII. It makes sure that
+     * unmappable characters are replaced with the default replacement character using
+     * {@link CodingErrorAction#REPLACE}.
+     *
+     * @throws Exception exception
+     */
+    @SuppressForbidden
```
rnveach, line 113: What is suppress forbidden for?

romani: Verify method, as we supposed to use verify that use input based config. But we need to set charset, and this type set we do not support yet.
So for now old style testing on Input file, without embedded config.

Issue #11902, opened 2022-07-16
Surviving pitest mutations for conditions reducing time complexity at PackageObjectFactory


Vyom-Yadav:
Two mutation cases to discuss, these conditions are present to reduce the time complexity.
https://github.com/checkstyle/checkstyle/blob/a583508cfe3478137d58f3776f860dd4fa548a03/config/pitest-suppressions/pitest-common-suppressions.xml#L210-L217

1.
```
<mutation unstable="false">                                                                          
  <sourceFile>PackageObjectFactory.java</sourceFile>                                                 
  <mutatedClass>com.puppycrawl.tools.checkstyle.PackageObjectFactory</mutatedClass>                  
  <mutatedMethod>createModule</mutatedMethod>                                                        
  <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
  <description>removed conditional - replaced equality check with true</description>                 
  <lineContent>if (thirdPartyNameToFullModuleNames == null) {</lineContent>                          
</mutation>                                                                                          
```
Complete relevant code https://github.com/checkstyle/checkstyle/blob/678c547af6de156c604595fe42ffdd0cb05c0023/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java#L189-L193

This null check is present to reduce the time complexity, the method `generateThirdPartyNameToFullModuleName(..)` uses `moduleClassLoader` (final instance field) and `packages` (another final instance field) so even if  `thirdPartyNameToFullModuleNames` is recalculated again and again the result will be the same. The only way it can change is if you access the internal fields, for eg - `packages` using reflections and add an element to the set. But in the current code those fields are not modified. 

2.
```
<mutation unstable="false">                                                                          
  <sourceFile>PackageObjectFactory.java</sourceFile>                                                 
  <mutatedClass>com.puppycrawl.tools.checkstyle.PackageObjectFactory</mutatedClass>                  
  <mutatedMethod>createModule</mutatedMethod>                                                        
  <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
  <description>removed conditional - replaced equality check with true</description>                 
  <lineContent>if (!name.contains(PACKAGE_SEPARATOR)) {</lineContent>                                
</mutation>                                                                                          
```

[...]

PR #13733, opened 2023-09-12

rdiachenko:
Issue #11902

Attempt to kill mutation preserving one time field initialization via Optional.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
@@ -27,6 +27,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -187,10 +188,11 @@ public class PackageObjectFactory implements ModuleFactory {
             instance = createFromStandardCheckSet(name);
             // find the name in third party map
             if (instance == null) {
-                if (thirdPartyNameToFullModuleNames == null) {
-                    thirdPartyNameToFullModuleNames =
-                            generateThirdPartyNameToFullModuleName(moduleClassLoader);
-                }
+                final Optional<Map<String, Set<String>>> cache =
+                        Optional.ofNullable(thirdPartyNameToFullModuleNames);
+                thirdPartyNameToFullModuleNames = cache.orElseGet(() -> {
+                    return generateThirdPartyNameToFullModuleName(moduleClassLoader);
+                });
                 instance = createObjectFromMap(name, thirdPartyNameToFullModuleNames);
             }
         }
--- a/config/pitest-suppressions/pitest-common-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-suppressions.xml
@@ -210,15 +210,6 @@
   <mutation unstable="false">
     <sourceFile>PackageObjectFactory.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.PackageObjectFactory</mutatedClass>
-    <mutatedMethod>createModule</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
-    <description>removed conditional - replaced equality check with true</description>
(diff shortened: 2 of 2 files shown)


romani:
I think this is better than crazy wrappers method and extra config in pom.xml

src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
```diff
@@ -187,10 +188,11 @@ public Object createModule(String name) throws CheckstyleException {
             instance = createFromStandardCheckSet(name);
             // find the name in third party map
             if (instance == null) {
-                if (thirdPartyNameToFullModuleNames == null) {
-                    thirdPartyNameToFullModuleNames =
-                            generateThirdPartyNameToFullModuleName(moduleClassLoader);
-                }
+                final Optional<Map<String, Set<String>>> cache =
+                        Optional.ofNullable(thirdPartyNameToFullModuleNames);
```
nrmancuso, line 192: I am surprised that none of our static analysis caught this. IMO, this is abuse of `Optional`.  Collections are better to be empty than optional. Typically `Optional` is used for a single object. Can we do:
1. Make this field final 9`thirdPartyNameToFullModuleNames`)
2. Initialize to empty Collection in constructor
3. Populate the Collection in this method
?

nrmancuso (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
romani: Let's do this in separate issue, we need to clean up pitest, project is close to completion.

nrmancuso: I would much rather we allow a suppression for pitest than allow Optional collections into the codebase. We have a mutable collection as a class field, this should be addressed anyway.

rdiachenko: > Can we do: ...

The mutation survives in this case. We either use Optional or use a wrapper method with exclusion in the pom.xml unless you have other ideas how to kill it.

romani: Optional is more elegant than another options

nrmancuso: IMO: best option is to leave this suppression with an explanation, this is why we have a suppression system.

Optional collection is a hack just like some wrapper method. We should not allow pitest to mangle production code like this; even a suppression in the pom is better than this. I don't see any reason to change code in cases where we have lazy loading, etc. for performance reasons. We are becoming too fanatic with our desire for empty suppression files.

The existing pitest suppression is the most targeted way to suppress this code; we should use it.

rnveach: The other option seems also just as easy, write a test to cover this situation.

The whole point of writing tests, besides coverage, is to also show expected behavior and functionality. It is expected we have lazy loading here to ensure we don't waste CPU cycles reloading the same data.

We have written tests using Mockito before and it provides the ability count how many times a method is called. We also use it in the releasenotes-builder tests.
https://stackoverflow.com/a/39452689/1016482

We should not be afraid to write these type of tests as this is the only way to prove this behavior is expected.

10 years down the road, no one will understand why we wrote such bad looking code if we merge this PR as it is. I am sorry, but I see this production code rewrite as just a method we have found to "outsmart" pitest, and have not actually solved the problem of showcasing lazy loading.

romani: Some ideas came to my mind.
Make method 
```
private initializeThirdPartyModuleMapIfRequired() {
if (thirdPartyNameToFullModuleNames == null) {
                    thirdPartyNameToFullModuleNames =
                            generateThirdPartyNameToFullModuleName(moduleClassLoader);
 }
}
```

And add it to pitest exclude in pom.xml

nrmancuso: If lazy initialization is a concern elsewhere, we could just make a generic util method that takes a supplier and use that.

romani: rethingking code in different way - https://github.com/checkstyle/checkstyle/pull/13744

PR #13744, opened 2023-09-16

romani:
#11902

continuation of https://github.com/checkstyle/checkstyle/pull/13733

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
@@ -187,11 +187,7 @@ public class PackageObjectFactory implements ModuleFactory {
             instance = createFromStandardCheckSet(name);
             // find the name in third party map
             if (instance == null) {
-                if (thirdPartyNameToFullModuleNames == null) {
-                    thirdPartyNameToFullModuleNames =
-                            generateThirdPartyNameToFullModuleName(moduleClassLoader);
-                }
-                instance = createObjectFromMap(name, thirdPartyNameToFullModuleNames);
+                instance = createObjectFromClassPath(name);
             }
         }
         if (instance == null) {
@@ -218,6 +214,21 @@ public class PackageObjectFactory implements ModuleFactory {
     }
 
     /**
+     * Calls method that does classpath scanning if required.
+     * This operation is very slow so it should be done only once.
+     *
+     * @return the map of third party Checkstyle module names to the set of their fully qualified
+     *      names
+     */
+    private Map<String, Set<String>> getThirdPartyModulesWithInitializationIfRequired() {
+        if (thirdPartyNameToFullModuleNames == null) {
+            thirdPartyNameToFullModuleNames =
+                    generateThirdPartyNameToFullModuleName(moduleClassLoader);
+        }
+        return thirdPartyNameToFullModuleNames;
+    }
+
+    /**
      * Create object from one of Checkstyle module names.
      *
      * @param name name of module.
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
```diff
@@ -217,6 +213,21 @@ Definitions.CHECKSTYLE_BUNDLE, getClass(),
         return instance;
     }
 
+    /**
+     * Calls method that does classpath scanning if required.
+     * This operation is very slow so it should be done only ones.
```
rnveach, line 218: `ones` =>` once`

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
romani: fixed

rnveach:
https://github.com/checkstyle/checkstyle/pull/13733#discussion_r1325246539
> The other option seems also just as easy, write a test to cover this situation.
> We have written tests using Mockito before and it provides the ability count how many times a method is called.

romani:
I don't believe in mocking much, it is problematic in long run, too much hacks to make it running, pitest does not like mocking.
In scope of this project we do functional coverage and if some coverage can not be archived functionality, we sacrifice it or avoid it.

In this case we avoid dealing with pitest.

romani:
@rnveach , do you agree on usage of this utility method?

The only problem I see in it, as I do not know any other places where it might be used, and we already moving util methods that  are used by single module back to modules. But I am ok for now to make it until method.

nrmancuso:
> @rnveach , do you agree on usage of this utility method?
> 
> The only problem I see in it, as I do not know any other places where it might be used, and we already moving util methods that are used by single module back to modules. But I am ok for now to make it until method.

Yeah, I am just wondering about alternatives, I thought I remembered similar lazy loading causing surviving mutations elsewhere. Extension of util methods is also bad practice, I agree (but this seems like a reasonable case where we do not care about any check logic at all).

Also - if we stick with changes in this PR, we can use TestUtil to write a test instead of having a suppression in the pom.

romani:
>we can use TestUtil to write a test instead of having a suppression in the pom.

Our religion is no tests for utils. All utilities for modules logic.

romani:
I created separate PR to let all compare https://github.com/checkstyle/checkstyle/pull/13757
I don't think such implementation becomes better. It is more lambda fancy but also has weirdness with readability.

src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java
```diff
@@ -217,6 +213,21 @@ Definitions.CHECKSTYLE_BUNDLE, getClass(),
...
+     * @return the map of third party Checkstyle module names to the set of their fully qualified
+     *      names
+     */
+    private Map<String, Set<String>> getThirdPartyModulesWithInitializationIfRequired() {
+        if (thirdPartyNameToFullModuleNames == null) {
+            thirdPartyNameToFullModuleNames =
+                    generateThirdPartyNameToFullModuleName(moduleClassLoader);
+        }
+        return thirdPartyNameToFullModuleNames;
+    }
```
nrmancuso, line 229: This implementation is better if we have no need for lazy loading elsewhere.

Vyom-Yadav: I am ok with this. This is pretty easy to read. Lambdas aren't helping in any way here. Simple, straightforward code, no need of lambdas.

romani: @rnveach , please review this, if you approve it, we will merge this PR.

rnveach: I already merged the other PR which was opened. It was not clear I was to pick between these.

romani: @rnveach , just share your opinion on this update, if you like it more I can rebase it  to keep this code

rnveach: I prefer us to write a test to cover lazy loading than either method.

romani: It is next step with fancy mocking.
If we start doing it, this code will be easier to mock i think.

We can even test it by Junit with call of private method.

rnveach: Since the other PR is merged, I don't see a point in doing any steps except reverting the code and switching to an actual test to cover lazy loading, or an explanation why we can't create a test and keep the code as it is now.

romani:
I am closing this, looks like team does not like it.

Issue #13746, opened 2023-09-16
PropertyCacheFileTest.testPathToCacheContainsOnlyFileName is failing


romani:
during execution of `mvn test` with  code of https://github.com/checkstyle/checkstyle/pull/13744

unexpected failure in junit on my local:
```
[INFO] Running com.puppycrawl.tools.checkstyle.PropertyCacheFileTest
[ERROR] Tests run: 15, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.380 s <<< FAILURE! -- in com.puppycrawl.tools.checkstyle.PropertyCacheFileTest
[ERROR] com.puppycrawl.tools.checkstyle.PropertyCacheFileTest.testPathToCacheContainsOnlyFileName 
-- Time elapsed: 0.004 s <<< ERROR!
java.nio.file.FileSystemException: temp.cache: this is a folder
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:100)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:219)
	at java.base/java.nio.file.spi.FileSystemProvider.newOutputStream(FileSystemProvider.java:478)
	at java.base/java.nio.file.Files.newOutputStream(Files.java:220)
	at com.puppycrawl.tools.checkstyle.PropertyCacheFile.persist(PropertyCacheFile.java:150)
	at com.puppycrawl.tools.checkstyle.PropertyCacheFileTest.testPathToCacheContainsOnlyFileName(PropertyCacheFileTest.java:276)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:727)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
[...]

PR #13747, opened 2023-09-16
Issue #13746: fix junit method to use temp file

romani:
#13746

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
@@ -33,7 +33,6 @@ import java.io.ObjectOutputStream;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
```diff
@@ -268,16 +267,14 @@ public void testCacheDirectoryDoesNotExistAndShouldBeCreated() throws IOExceptio
     @Test
     public void testPathToCacheContainsOnlyFileName() throws IOException {
         final Configuration config = new DefaultConfiguration("myName");
-        final String fileName = "temp.cache";
-        final Path filePath = Paths.get(fileName);
-        final PropertyCacheFile cache = new PropertyCacheFile(config, fileName);
+        final String filePath = File.createTempFile("junit", null, temporaryFolder).getPath();
```
rnveach, line 270: `getPath` => `getName`

rnveach (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
romani: Why?
We use to have path.

rnveach: `testPathToCacheContainsOnlyFileName `

romani: hmm, in this case it creates junit15239200722429684874.tmp

and after digging deeper, this test failed becuse for some reason my local I got this file as folder, but it suppose to be a file,
weird, .... I am closing PR as no change should be done, and i do not know how to reproduce problem with folder creation instead of file

PR #13752, opened 2023-09-18
Issue #11902: resolve pitest

romani:
part of Issue #11902


romani:
this update is bumping `mvn verify` to 36 minutes, not acceptable performance

Issue #13666, opened 2023-08-30
Assert that properties are ordered alphabetically in xdoc


stoyanK7:
As suggested in https://github.com/checkstyle/checkstyle/pull/13407#discussion_r1309128616

We need to rearrange properties in alphabetical order(xdoc and javadoc since they are connected) and add a test for that. This will ease reviews when applying the properties macro to checks.

https://checkstyle.org/checks/annotation/annotationusestyle.html for example, has properties that do not follow alphabetical order:

- elementStyle
- closingParens
- trailingArrayComma

Edit: This issue is only for adding the test with suppressions. I will deal with ordering the properties at https://github.com/checkstyle/checkstyle/issues/13668

Update: almost all are ordered with reference to this issue. #13668 is closed.

stoyanK7:
Do we still want to keep `tokens`/`javadocTokens` at the end or do we want true alphabetical order?

romani:
Let's do alphabetical for now, after migration to macroses it will be easy to change to any alphabetical-special.

stoyanK7:
Already went with alphabetical and `javadocTokens`, `tokens` at the end. Please check #13677

romani:
>This will ease reviews when applying the properties macro to checks.

I found it as not easy, a way easier is to migrate to macros, so ordering is enforcing by macros.

PR #13754, opened 2023-09-19
Issue #13666: fix ordering of properties in few modules

romani:
Issue #13666

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagContinuationIndentationCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagContinuationIndentationCheck.java
@@ -38,16 +38,16 @@ import com.puppycrawl.tools.checkstyle.utils.JavadocUtil;
  * </p>
  * <ul>
  * <li>
+ * Property {@code offset} - Specify how many spaces to use for new indentation level.
+ * Type is {@code int}.
+ * Default value is {@code 4}.
+ * </li>
+ * <li>
  * Property {@code violateExecutionOnNonTightHtml} - Control when to print violations
  * if the Javadoc being examined by this check violates the tight html rules defined at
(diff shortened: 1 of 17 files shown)


nrmancuso (approved):

rnveach:
I think our documentation at https://checkstyle.org/property_types.html#Pattern.5B.5D is either wrong or misleading.
https://checkstyle.org/checks/metrics/classfanoutcomplexity.html#Properties

> The string representation is parsed as a set of forward slash ('/') separated patterns.

When glancing at this, i thought it meant each pattern is separated by a `/`.

But looking at the code at https://github.com/checkstyle/checkstyle/blob/5a0f923a6d48a9152b963468bfece7b6e9f28295/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/AbstractClassCouplingCheck.java#L160 , it takes input as a string array, which is separated by the comma and not the `/`.

Even `createPattern` method that it calls does nothing with `/`, so I don't understand where the `/` is coming from.

https://github.com/search?q=repo%3Acheckstyle%2Fcheckstyle+%22Pattern.5B.5D%22+path%3A%2F%5Esrc%5C%2Fxdocs%5C%2Fchecks%5C%2F%2F&type=code
We have 6 checks with this type.

It looks like documentation is coming from https://github.com/checkstyle/checkstyle/blob/5a0f923a6d48a9152b963468bfece7b6e9f28295/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L759 , which is a serious problem now as we have a documented type being parsed differently in different checks, meaning we can't even unify documentation without first changing production code.

@romani If you agree, create an issue on this.

rnveach:
Truthfully we shouldn't allow global types on https://checkstyle.org/property_types.html
if it is not listed in https://github.com/checkstyle/checkstyle/blob/5a0f923a6d48a9152b963468bfece7b6e9f28295/src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java#L110-L111 .

We only have a global pattern of `Pattern` and not `Pattern[]`.

romani:
@rnveach , please continue to review this PR (ordering)

PR #13757, opened 2023-09-20

romani:
Issue #11902


nrmancuso (approved):
I am conceptually OK with this, but I don't see why we would need such a generic implementation for a private method in a single check. My proposal was to add this to CheckUtil and use it in other places where had issues with pitest failing on lazy loading, I seemed to remember some others.

Is the plan to move this to CheckUtil later?

https://github.com/checkstyle/checkstyle/pull/13744/files#r1332246323

romani:
I try to keep our rule to have something in util only if non single usage is present.

rnveach (approved):

PR #3438, opened 2016-09-11
Issue #2999: allow regex in importcontrol and subpackage element

vboerchers:
This pull request fixes Issue #2999. Notes:
* I have added the possibility to define packages not only to the `subpackage` element but also to the root element `importcontrol`.
* The travis will continue to fail until the new DTD [import_control_1_2.dtd](https://github.com/vboerchers/checkstyle/blob/master/src/main/resources/com/puppycrawl/tools/checkstyle/checks/imports/import_control_1_2.dtd) is available as http://www.puppycrawl.com/dtds/import_control_1_2.dtd

Issue #2999, opened 2016-03-03
ImportControl: allow regex in subpackage elements.


effjay:
Hi all,

at the moment it is not possible to use the same import control configuration for multiple subpackages. 

What I need to do is this:

```
<subpackage pkg="de.main">
   <subpackage pkg="*"> 
      <subpackage pkg="service">
         <allow...> 
      </subpackage>
   </subpackage>
</subpackage>
```

This way, the package control configuration for the service subpackage would match all modules at once, like de.main.mod1.service as well as de.main.mod2.service and so on. If someone creates a mod3 module, then the configuration would still work.

I couldn't get it to work, as wildcards and regex are only valid on allow or disallow elements right now.

For this, a new attribute might need to be created for the "subpackage" element: "regex"="true/false". That is required to be consistent with the design of  the "allow" element.

This will mean that http://www.puppycrawl.com/dtds/import_control_1_1.dtd must be changed to version 1.2 as well.

Thanks for any help on this, best regards
Daniel

romani:
Issues was discussed in mailist https://groups.google.com/forum/m/#!topic/checkstyle/Jp4jktHA4VI

DyosTheNerd:
As I also would like to see this feature for my next project, but never really used github, how can I start contributing?

rnveach:
@DyosTheNerd 
Check out the following links:
http://checkstyle.sourceforge.net/contributing.html
http://checkstyle.sourceforge.net/beginning_development.html (has more specific details of the 2)

Besides knowing Java, you will have to be able to work with Maven and Git.
Questions may be asked on Gitter or Google Groups.

romani:
@DyosTheNerd , please review https://github.com/sevntu-checkstyle/sevntu.checkstyle/wiki it was written for beginners.

vboerchers:
@DyosTheNerd

What is your status on this issue? I have a need for this feature too and I am considering working on it.

romani:
@vboerchers , nothing was done for 2 month it usually mean that he has no time to do this.
please be welcome to provide a patch.

vboerchers:
I have a question regarding XML design: I have started by
`<subpackage pkg=".*" regex="true">` since this is the way it's implemented for `<allow>` and `<disallow>` but I'm wondering if

    <subpackage regex=".*">
 
wouldn't be nicer. What do you think?

vboerchers:
[...]

PR #13760, opened 2023-09-21
Issue #13672: Kill mutation for FileImportControl

Kevin222004:
Issue #13672: Kill mutation for FileImportControl

# Mutation 
https://github.com/checkstyle/checkstyle/blob/ccc12eb6d7f6c8e0876077e7883cf8f59a541a37/config/pitest-suppressions/pitest-imports-suppressions.xml#L12-L19

# Explaination

This constructor
https://github.com/checkstyle/checkstyle/blob/ccc12eb6d7f6c8e0876077e7883cf8f59a541a37/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java#L44-L49
 is used called only at line 185
https://github.com/checkstyle/checkstyle/blob/ccc12eb6d7f6c8e0876077e7883cf8f59a541a37/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportControlLoader.java#L181-L186

so this elseif statment will execute in case 
```
 <file name="InputImportControl.*" regex="true">
    <disallow class="java.awt.Image"/>
  </file>
```

where we have file know encloseInGroup is used to non-capture the group I have done some research at
https://stackoverflow.com/questions/3512471/what-is-a-non-capturing-group-in-regular-expressions

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java
@@ -43,26 +43,9 @@ class FileImportControl extends AbstractImportControl {
      */
     /* package */ FileImportControl(PkgImportControl parent, String name, boolean regex) {
         super(parent, MismatchStrategy.DELEGATE_TO_PARENT);
-
         this.regex = regex;
-        if (regex) {
-            this.name = encloseInGroup(name);
-            patternForExactMatch = createPatternForExactMatch(this.name);
-        }
-        else {
-            this.name = name;
-            patternForExactMatch = null;
-        }
-    }
-
-    /**
-     * Enclose {@code expression} in a (non-capturing) group.
-     *
-     * @param expression the input regular expression
-     * @return a grouped pattern.
-     */
-    private static String encloseInGroup(String expression) {
-        return "(?:" + expression + ")";
+        this.name = name;
+        patternForExactMatch = createPatternForExactMatch(name);
     }
 
     /**
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -2845,17 +2845,6 @@
 
   <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java</fileName>
-    <specifier>assignment</specifier>
-    <message>incompatible types in assignment.</message>
-    <lineContent>patternForExactMatch = null;</lineContent>
-    <details>
-      found   : null (NullType)
-      required: @Initialized @NonNull Pattern
-    </details>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java</fileName>
     <specifier>return</specifier>
     <message>incompatible types in return.</message>
     <lineContent>return finestMatch;</lineContent>
--- a/config/pitest-suppressions/pitest-imports-suppressions.xml
+++ b/config/pitest-suppressions/pitest-imports-suppressions.xml
@@ -1,24 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
   <mutation unstable="false">
-    <sourceFile>FileImportControl.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.imports.FileImportControl</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable patternForExactMatch</description>
-    <lineContent>patternForExactMatch = null;</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>FileImportControl.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.imports.FileImportControl</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl::encloseInGroup with argument</description>
-    <lineContent>this.name = encloseInGroup(name);</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>PkgImportControl.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.imports.PkgImportControl</mutatedClass>
     <mutatedMethod>&lt;init&gt;</mutatedMethod>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportControlCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportControlCheckTest.java
@@ -452,6 +452,14 @@ public class ImportControlCheckTest extends AbstractModuleTestSupport {
     }
 
     @Test
+    public void testImportControlFileName2() throws Exception {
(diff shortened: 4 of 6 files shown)


Kevin222004:
In explanation, I have only mentioned where it is used. And tried to understand the significance of it but i don't understand why the reason to do this. Please let me know what value this will create

romani:
hmm, I do not see any example of "file" tag https://checkstyle.org/checks/imports/importcontrol.html#ImportControl

romani:
@rnveach , do you remember why you used non captured group as wrapper?

romani:
pitest blocks failed due to crash of pitest execution.

imports pitest has:
```
 <suppressedMutations>
+    <sourceFile>FileImportControl.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.imports.FileImportControl</mutatedClass>
+    <lineContent>if (regex) {</lineContent>
+  </mutation>
```

rnveach:
> @rnveach , do you remember why you used non captured group as wrapper?

What code are you talking about? I am not following.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java
```diff
@@ -55,16 +55,6 @@ class FileImportControl extends AbstractImportControl {
...
     }
 
-    /**
-     * Enclose {@code expression} in a (non-capturing) group.
-     *
-     * @param expression the input regular expression
-     * @return a grouped pattern.
-     */
-    private static String encloseInGroup(String expression) {
-        return "(?:" + expression + ")";
```
romani, line 65: @rnveach , This code.


If we don't remember why we did this, we will remove it

rnveach: The comment above its call says `ensure that fullPackage is a self-contained regular expression`.

You had to search a bit more in the commit you found.
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/PkgImportControl.java#L155-L156 was renamed from `ImportControl`, and then from `PkgControl` before that, and that is where the code originally came from. I did not add it.

It appears the code really came from https://github.com/checkstyle/checkstyle/commit/63e8c1dd4e1da443c8533d40640fe1a3b371afe8#diff-f68bc4c425df6a25439881dbcd97576b489f7d71fc7706a450e11a101f0d3e77R148-R149 .
https://github.com/checkstyle/checkstyle/issues/2999

Some things of interest:
https://github.com/checkstyle/checkstyle/pull/3438#issuecomment-252503843
https://github.com/checkstyle/checkstyle/pull/3438#discussion_r82990960
https://github.com/checkstyle/checkstyle/pull/3438#discussion_r83542445

romani: Ok, thanks a lot, after reading all of this, there is no special reason of such code.

Last reason could be performance, as regexp does not need do creation of group. But as we wrap whole regexp it is does not guarantee that all nested groups will be downgraded to non-captured groups.

I am ok to remove this method.

romani (approved):
Ok to merge

romani:
@Kevin222004 , please resolve pitest failure:
```

 <suppressedMutations>
+    <sourceFile>FileImportControl.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.imports.FileImportControl</mutatedClass>
+    <lineContent>if (regex) {</lineContent>
+  </mutation>
+
```

Please prepare PR for merge.

romani (changes requested):
See reason above

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java
rnveach: @romani 

> (?:.*(Panel|View|Dialog))

What happens when the regexp changes to `.*(Panel|View|Dialog)`? Since this now a capturing group, will it cause any issues?

romani: If no code deals with group, no changes.
We do not use group, we just match, true or false.

rnveach: @romani To me the following lines explain why we need `encloseInGroup`.
https://github.com/checkstyle/checkstyle/blob/4d113f93dd9a4a731515f2e5da76112e37afd49b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/PkgImportControl.java#L128-L130

We can't rely on the users adding `()` themselves and we enclose it in a `()` in case they don't.

rnveach (approved):
It seems to me we could remove `name` completely, and transfer everything to `patternForExactMatch`. We would just have to sanitize name that we put into `patternForExactMatch` if it is not `regexp`.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java
rnveach: > If no code deals with group, no changes.
We do not use group

Or if we do not combine 2 regexps.

rnveach: @romani https://github.com/checkstyle/checkstyle/pull/13760#discussion_r1337408244 . I also approved this PR as I think the code removal is fine.

romani:
@Kevin222004 , please resolve a conflict.

Let us know if you donot know how to resolve this survival, so mentors will look closer.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java
```diff
@@ -43,26 +42,8 @@ class FileImportControl extends AbstractImportControl {
      */
     /* package */ FileImportControl(PkgImportControl parent, String name, boolean regex) {
         super(parent, MismatchStrategy.DELEGATE_TO_PARENT);
```
Kevin222004, line 44: This is the constructor. so after loading all the configuration and setup when `visitToken` in ImportControl
https://github.com/checkstyle/checkstyle/blob/83788d6ff5fb659f8590906557c46b79a018f21e/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportControlCheck.java#L237
after executing this statement 
https://github.com/checkstyle/checkstyle/blob/83788d6ff5fb659f8590906557c46b79a018f21e/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java#L80
after this method 
https://github.com/checkstyle/checkstyle/blob/83788d6ff5fb659f8590906557c46b79a018f21e/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java#L90-L102

where we already check that if regex is true or not so that's why replacing this https://github.com/checkstyle/checkstyle/blob/83788d6ff5fb659f8590906557c46b79a018f21e/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java#L48-L55
if as true will not make any affect. https://github.com/checkstyle/checkstyle/blob/83788d6ff5fb659f8590906557c46b79a018f21e/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java#L33
The only usage of this field is at https://github.com/checkstyle/checkstyle/blob/83788d6ff5fb659f8590906557c46b79a018f21e/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java#L95-L97
so we can remove this field  and directly initialize that in method it will not make any issue

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java
```diff
@@ -93,6 +74,7 @@ protected boolean matchesExactly(String pkg, String fileName) {
             result = false;
         }
         else if (regex) {
+            final Pattern patternForExactMatch = createPatternForExactMatch(name);
```
romani, line 77: This is performance issue.
Compilation of pattern should be done ones, not on each try of match

romani (changes requested):
Items

romani (approved):
Ok to merge

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/FileImportControl.java
romani: Done

PR #13768, opened 2023-09-23
Issue #13672: Kill mutation for Xpath

Kevin222004:
Issue #13672: Kill mutation for Xpath

------------

# Mutation 

https://github.com/checkstyle/checkstyle/blob/c525c6732e2c1f15bd3f85f943770c649f1ab775/config/pitest-suppressions/pitest-xpath-suppressions.xml#L30-L46
and
https://github.com/checkstyle/checkstyle/blob/c525c6732e2c1f15bd3f85f943770c649f1ab775/config/pitest-suppressions/pitest-xpath-suppressions.xml#L3-L10

----------

# Explaination

Test added

### DIFF
--- a/config/pitest-suppressions/pitest-xpath-suppressions.xml
+++ b/config/pitest-suppressions/pitest-xpath-suppressions.xml
@@ -1,14 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <mutation unstable="false">
-    <sourceFile>AbstractElementNode.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.xpath.AbstractElementNode</mutatedClass>
-    <mutatedMethod>getPrecedingSiblings</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator</mutator>
-    <description>replaced call to java/util/Collections::unmodifiableList with argument</description>
-    <lineContent>return Collections.unmodifiableList(siblings.subList(0, indexAmongSiblings));</lineContent>
-  </mutation>
-
   <mutation unstable="true">
     <sourceFile>AbstractRootNode.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.xpath.AbstractRootNode</mutatedClass>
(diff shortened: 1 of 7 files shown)


config/pitest-suppressions/pitest-xpath-suppressions.xml
```diff
@@ -27,24 +18,6 @@
...
-  </mutation>
-
-  <mutation unstable="true">
-    <sourceFile>AbstractRootNode.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.xpath.AbstractRootNode</mutatedClass>
-    <mutatedMethod>iterateAxis</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.RemoveSwitchMutator_8</mutator>
-    <description>RemoveSwitch 8 (case value 8)</description>
-    <lineContent>switch (axisNumber) {</lineContent>
-  </mutation>
```
Kevin222004, line 46: Their is no case in which the case value 8 is used

config/checkstyle-checks.xml
```diff
@@ -613,6 +613,11 @@
       <property name="files" value=".*[\\/]src[\\/](test|it|xdocs-examples)[\\/]"/>
       <property name="checks" value="MethodCount"/>
     </module>
+    <module name="SuppressionXpathSingleFilter">
+      <property name="files"
+                value=".*[\\/]xpath[\\/]AbstractElementNode.java"/>
+      <property name="checks" value="ImportControl"/>
```
romani, line 619: I do not see any criminal to allow new import in this class.
Please remove suppression and add allowance for required import.

romani (changes requested):
item:

config/checkstyle-checks.xml
Kevin222004: done

src/test/resources/com/puppycrawl/tools/checkstyle/xpath/abstractrootnode/InputAbstractRootNode.java
```diff
@@ -0,0 +1,18 @@
...
+com.puppycrawl.tools.checkstyle.checks.coding.IllegalTokenTextCheck
+format = [^a-zA-z0-9]*
+ignoreCase = (default)false
+message = (default)
+tokens = CHAR_LITERAL
+
+*/
+package com.puppycrawl.tools.checkstyle.xpath.abstractrootnode;
+
+public class InputAbstractRootNode {
```
romani, line 15: SuppressionXpathFilter vs InputAbstractRootNode is red flag.

please move this code to SuppressionXpathFilterTest.
fact that we have some abstract class does not mean that we need to test is as abstract. We test all from user perspective, user knows only SuppressionXpathFilter and amount of abstract classes or classes used by composite that are under it does not matter.

if you keep all tests at level of SuppressionXpathFilter we can refactor all easily and we will never need to change a test.

config/suppressions-xpath.xml
```diff
@@ -515,4 +515,8 @@
           checks="LocalVariableNameCheck"
           query="//METHOD_DEF[./IDENT[@text='getLineNumbersFromExpected']]
                 //VARIABLE_DEF/IDENT[@text='i']"/>
+  <suppress-xpath
+          files="AbstractRootNodeTest.java"
+          checks="AbstractClassName"
```
romani, line 520: not good, see a reason in another comment.

romani (changes requested):
last:

src/test/resources/com/puppycrawl/tools/checkstyle/xpath/abstractrootnode/InputAbstractRootNode.java
Kevin222004: Done

config/suppressions-xpath.xml
Kevin222004: Done

romani (approved):
Awesome

PR #13771, opened 2023-09-23
Issue #13672: Kill mutation for Xpath-2

Kevin222004:
Issue #13672: Kill mutation for Xpath-2

-----------

# Mutation
Test

### DIFF
--- a/config/pitest-suppressions/pitest-xpath-suppressions.xml
+++ b/config/pitest-suppressions/pitest-xpath-suppressions.xml
@@ -4,15 +4,6 @@
     <sourceFile>AbstractRootNode.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.xpath.AbstractRootNode</mutatedClass>
     <mutatedMethod>iterateAxis</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.RemoveSwitchMutator_1</mutator>
-    <description>RemoveSwitch 1 (case value 1)</description>
-    <lineContent>switch (axisNumber) {</lineContent>
-  </mutation>
-
-  <mutation unstable="true">
-    <sourceFile>AbstractRootNode.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.xpath.AbstractRootNode</mutatedClass>
-    <mutatedMethod>iterateAxis</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.RemoveSwitchMutator_2</mutator>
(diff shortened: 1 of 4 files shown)


romani:
> know ATTRIBUTE and ANCESTOR_OR_SELF will never be at root

it might be nonsense by user can write it (as mistake) but we would rather return no result than exception.
same with Regexp, there are uncountable amount of ways to make nonsense, and we do not block users to make it.

can you share CLI output that we do not throw excpeption on cases when such axis are used at root ?

Kevin222004:
OK, I was wrong about `ANCESTOR_OR_SELF` After re-thinking I have found this test will convert this in InputBased and kill Mutation 

I am also re-thinking for `ATTRIBUTE` axis

```
kevin@kevin-Inspiron-15-5510:~/Desktop/check_style$ cat Test.java 
package com.puppycrawl.tools.checkstyle.filters.suppressionxpathfilter;

public class Test {

    char a = '&'; // filtered violation
}
 kevin@kevin-Inspiron-15-5510:~/Desktop/check_style$ cat config.xml 
<?xml version="1.0"?>

<module name="Checker">
    <module name="TreeWalker">
        <module name="IllegalTokenText">
            <property name="format" value="[^a-zA-z0-9]*"/>
            <property name="tokens" value="CHAR_LITERAL"/>
        </module>
        <module name="SuppressionXpathFilter">
            <property name="file" value="/home/kevin/Desktop/check_style/Suppression-Xpath.xml" />
        </module>
    </module>
</module>
kevin@kevin-Inspiron-15-5510:~/Desktop/check_style$ cat Suppression-Xpath.xml 
<!DOCTYPE suppressions PUBLIC
        "-//Checkstyle//DTD SuppressionXpathFilter Experimental Configuration 1.2//EN"
        "https://checkstyle.org/dtds/suppressions_1_2_xpath_experimental.dtd">
<suppressions>
    <suppress-xpath
            files="Test.java"
            checks="IllegalTokenTextCheck"
            query="//ancestor-or-self::*"/>
</suppressions>
kevin@kevin-Inspiron-15-5510:~/Desktop/check_style$ java -jar /home/kevin/Downloads/checkstyle-10.12.3-all.jar -c config.xml Test.java
Starting audit...
Audit done.
kevin@kevin-Inspiron-15-5510:~/Desktop/check_style$ 
```

romani:
>I am also re-thinking for ATTRIBUTE axis

I don't see test for this, what are your thoughts on this?

Kevin222004:
After re-thinking on `ATTRIBUTE` I don't find any test case. The attribute is used only as a ElementNode

romani (approved):
Ok to merge if ci pass

PR #13778, opened 2023-09-25
Issue #13666: fix property order for few modules

romani:
Issue #13666

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java
@@ -68,6 +68,12 @@ import com.puppycrawl.tools.checkstyle.api.DetailAST;
  * </pre>
  * <ul>
  * <li>
+ * Property {@code arrayInitIndent} - Specify how far an array initialisation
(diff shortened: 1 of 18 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java
```diff
@@ -68,6 +68,12 @@
  * </pre>
  * <ul>
  * <li>
+ * Property {@code arrayInitIndent} - Specify how far an array initialisation
```
rnveach, line 71: Not related but this should be `initialization`.

rnveach (approved):

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java
romani: Will do in separate PR

romani: @rnveach , please consider https://github.com/checkstyle/checkstyle/issues/13783

PR #13794, opened 2023-09-28
Issue #13672: Kill mutation for SuppressionXpathSingleFilter

Kevin222004:
Issue #13672: Kill mutation for SuppressionXpathSingleFilter

### DIFF
--- a/config/pitest-suppressions/pitest-filters-suppressions.xml
+++ b/config/pitest-suppressions/pitest-filters-suppressions.xml
@@ -174,15 +174,6 @@
   <mutation unstable="false">
     <sourceFile>SuppressionXpathSingleFilter.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.filters.SuppressionXpathSingleFilter</mutatedClass>
-    <mutatedMethod>setFiles</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable files</description>
-    <lineContent>this.files = null;</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>SuppressionXpathSingleFilter.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.filters.SuppressionXpathSingleFilter</mutatedClass>
     <mutatedMethod>setMessage</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
     <description>Removed assignment to member variable message</description>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
@@ -26,11 +26,13 @@ import static com.puppycrawl.tools.checkstyle.checks.naming.AbstractNameCheck.MS
 
 import java.io.File;
 import java.nio.charset.StandardCharsets;
+import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
 import org.junit.jupiter.api.Test;
 
 import com.puppycrawl.tools.checkstyle.AbstractModuleTestSupport;
+import com.puppycrawl.tools.checkstyle.JavaParser;
 import com.puppycrawl.tools.checkstyle.TreeWalkerAuditEvent;
 import com.puppycrawl.tools.checkstyle.api.FileContents;
 import com.puppycrawl.tools.checkstyle.api.FileText;
@@ -398,6 +400,55 @@ public class SuppressionXpathSingleFilterTest
         verifyFilterWithInlineConfigParser(
                 getPath("InputSuppressionXpathSingleFilterDecideById.java"),
                 expected, removeSuppressed(expected, suppressed));
+    }
+
+    /**
+     * This test is required to cover pitest mutation
+     * for reset of 'file' field in SuppressionXpathSingleFilter.
+     * This not possible to reproduce by natural execution of Checkstyle
+     * as config is never changed in runtime.
+     * Projects that use us by api can reproduce this case.
+     * We need to allow users to reset module property to default state.
+     *
+     * @throws Exception when there is problem to load Input file
+     */
+    @Test
+    public void testUpdateFilterSettingsInRunTime() throws Exception {
+        final File file = new File(getPath("InputSuppressionXpathSingleFilterComplexQuery.java"));
+
+        final SuppressionXpathSingleFilter filter = new SuppressionXpathSingleFilter();
+        filter.setChecks("MagicNumber");
+        filter.finishLocalSetup();
+
+        final Violation violation = new Violation(27, 21, TokenTypes.NUM_DOUBLE, "",
+                        "", null, null, null,
+                        MagicNumberCheck.class, null);
+
+        final FileContents fileContents =
+                new FileContents(new FileText(file, StandardCharsets.UTF_8.name()));
+
+        final TreeWalkerAuditEvent ev = new TreeWalkerAuditEvent(fileContents, file.getName(),
+                violation, JavaParser.parseFile(file, JavaParser.Options.WITHOUT_COMMENTS));
+
+        assertWithMessage("match ia expected as 'files' is defaulted")
+                .that(filter.accept(ev))
+                .isFalse();
+
+        // set non-default value for files
+        filter.setFiles(Pattern.quote(file.getPath() + ".never.match"));
+        filter.finishLocalSetup();
+
+        assertWithMessage("no match is expected due to weird value in 'files'")
+                .that(filter.accept(ev))
+                .isTrue();
+
+        // reset files to default value of filter
+        filter.setFiles(null);
+        filter.finishLocalSetup();
+
+        assertWithMessage("match ia expected as 'files' is defaulted")
+                .that(filter.accept(ev))
+                .isFalse();
     }
 
     private static SuppressionXpathSingleFilter createSuppressionXpathSingleFilter(


config/pitest-suppressions/pitest-filters-suppressions.xml
```diff
@@ -351,15 +351,6 @@
...
   </mutation>
 
-  <mutation unstable="false">
-    <sourceFile>SuppressionXpathSingleFilter.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.filters.SuppressionXpathSingleFilter</mutatedClass>
-    <mutatedMethod>setFiles</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable files</description>
-    <lineContent>this.files = null;</lineContent>
-  </mutation>
```
Kevin222004, line 361: @romani this mutation is not possible to cover with unit test as well becuase the mutation is about to remove the assignment so by default it become null. if we remove the assignment then the code becomes non-compilable 

https://kevin222004.github.io/PitestReport/com.puppycrawl.tools.checkstyle.filters/SuppressionXpathSingleFilter.java.html

romani: @Kevin222004 , you junit method should be improved.

it shuold set `filter.setFiles("some file");` make sure that xpath is matching.
in same junit and on the same instance of `filter` do
`filter.setFiles(null);` and repeat `accept` and assert on no match of xpath.

with mitation it will match, so junit will fail.

and we need comment over pure junit method to explain why we do non-Input based testing.

Kevin222004: Ok I have update the test case can you please gave a look to improve it more looks like wrong query is also getting accepted

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
```diff
@@ -400,6 +401,49 @@ public void testDecideById() throws Exception {
...
+
+        final TreeWalkerAuditEvent ev = new TreeWalkerAuditEvent(fileContents, file.getName(),
+                violation, JavaParser.parseFile(file, JavaParser.Options.WITHOUT_COMMENTS));
+
+        assertWithMessage("")
+                .that(filter.accept(ev))
+                .isTrue();
+
+        try {
+            filter.setFiles(null);
```
romani, line 434: Null should not cause any exception, as we simulate not setting it at all

romani:
@Kevin222004 , I think you can make this PR as ready for review as ask mentors to help.

romani (approved):
ok to merge if CI pass

romani:
Windows failure:
```
[ERROR]   SuppressionXpathSingleFilterTest.testUpdateFilterSettingsInRunTime:437 » PatternSyntax Unknown character property name {In/Isr} near index 4
C:\projects\checkstyle\src\test\resources\com\puppycrawl\tools\checkstyle\filters\suppressionxpathsinglefilter\InputSuppressionXpathSingleFilterComplexQuery.java.never.match
    ^
[INFO] 

```

rnveach:
You didn't sanitize the path before passing it on. Files is a regexp field.

romani:
fixed.

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
```diff
@@ -400,6 +401,59 @@ public void testDecideById() throws Exception {
...
+        filter.setFiles(null);
+        filter.finishLocalSetup();
+
+        assertWithMessage("match ia expected as 'files' is defaulted")
+                .that(filter.accept(ev))
+                .isFalse();
+    }
+
+    private static String sanitizeWindowsPathForRegexp(String value) {
+        return value.replaceAll("\\\\", "\\\\\\\\\\\\\\\\");
```
rnveach, line 454: https://github.com/checkstyle/checkstyle/blob/a89ddac62254f3b94a2d97231bf6a878563d74af/config/suppressions.xml#L12

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
```diff
@@ -400,6 +401,59 @@ public void testDecideById() throws Exception {
...
+        // reset files to default value of filter
+        filter.setFiles(null);
+        filter.finishLocalSetup();
+
+        assertWithMessage("match ia expected as 'files' is defaulted")
+                .that(filter.accept(ev))
+                .isFalse();
+    }
+
+    private static String sanitizeWindowsPathForRegexp(String value) {
```
rnveach, line 453: This has nothing to do with Windows. It is just linux was not smart enough to notice the issue.

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
```diff
@@ -400,6 +401,59 @@ public void testDecideById() throws Exception {
...
+        filter.setFiles(sanitizeWindowsPathForRegexp(file.getPath() + ".never.match"));
+        filter.finishLocalSetup();
+
+        assertWithMessage("no match is expected due to weird value in 'files'")
+                .that(filter.accept(ev))
+                .isTrue();
+
+        // reset files to default value of filter
+        filter.setFiles(null);
+        filter.finishLocalSetup();
```
rnveach, line 446: I am pretty sure this goes against the normal lifecycle. We finish the setup 3 separate times. What is the mutation we are trying to kill?

rnveach (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
romani: Mutation is in diff as removal, in suppression files.

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
romani: What name you suggest to use?
In Linux there is no `\` in path, so I tried by name of method explaining what am I targeting to address.

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
romani: It was good form for humans to deal with, do you really think we need this fancy escaping for execution in windows in Junit?

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
rnveach: > In Linux there is no \ in path

This is why linux did not find the issue. It is a sanitation issue and the only way to identify it was to give it something that wasn't regexp safe. The issue would have presented itself if the linux path had some things that weren't regexp safe.

> What name you suggest to use?

`sanitizePathForRegexp`

In addition, I recommend to just use `Pattern.quote`.
https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#quote-java.lang.String-

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
rnveach: I assume this is it.
https://github.com/checkstyle/checkstyle/blob/a89ddac62254f3b94a2d97231bf6a878563d74af/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilter.java#L125

https://pitest.org/quickstart/mutators/#EXPERIMENTAL_MEMBER_VARIABLE
> Please Note: This mutator is likely to create equivalent mutations if a member variable is explicitly initialized with the Java default value for the specific type of the member variable as in

This seems like another example we should report to pitest to change.
````
variable = assignment;
````

````
into
````
if (assignment == null) new RuntimeException();
variable = assignment;
````

Similar to what it does for other mutators like RETURN_VALS.
https://pitest.org/quickstart/mutators/#return-values-mutator-return_vals

Comment (romani) [reply src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java:446]: I interpreted pitest as:
```
public void setFiles(String files) {
        if (files == null) {
            
        }
        else {
            this.files = Pattern.compile(files);
        }
    }
```
and covered ^^^ mutation.

File: src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java (hunk 4 above)
Comment (romani) [reply src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java]: Pattern.quote is used.

File: src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java (hunk 3 above)
Comment (romani) [reply src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java]: method is removed.

File: src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java (hunk 5 above)
Comment (nrmancuso) [reply src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java:447]: Why can't we change this code to:
```
 if (files != null) {
     this.files = Pattern.compile(files);
 }

```
?

romani: We try the best to allow user to be able to put mutate move modules to default state of properties.
In other words, user created module, changed values of properties, changed his mind and want to change property value back to default.

it is not possible in XML in most cases where null is default value. But xml config is not whole world for us, I hope one day there will be more formats we support. If our java beans modules continue to allow setting default value, we will always be good.

So this is extreme edge case for us, I doubt much users care about this, but  as a library we should not have opinion on our modules usage, and just allow to set default value back by setter.

nrmancuso: Do users really consider check module properties mutable? I understand that due to the way we instantiate/configure checks that it is not reasonable to assign all properties via constructor as final fields, but usage of check modules as mutable objects seems like a bad idea to me.

romani: Modules were mutable for 20 years.
We do not know how our users use us.
But ideally module should be initialized by explicit default values and work same as without being explicit.

nrmancuso (approved):
IMO, the pitest suppression is better to keep than this crazy test. But if we want to translate tech debt from suppression files to tests, let's do it.

romani:
Keep him mind that survival is here because we use null for default value of property. All modules which doing this are prone to it. By default value is null and there is special behavior for such null. But user can not set null by xml config, if user want something to be unset, he should not have property tag in config.

But this creates inconsistency. All other fields can be defined in xml with explicit value that is equal to default and it works as default. But nullable properties are different.

That is why I demanding in new design not null property, and be like empty string or anything that can be set in xml as value.

Ideally we should refactor module and do not allow java nulls , but this is breaking compatibility and a way more complicated, out of scope of this project. 

If we have time, we need to update all modules to have ability to set default value explicitly iby supported configs.

rnveach:
> Ideally we should refactor module and do not allow java nulls 
> but this is breaking compatibility
> a way more complicated

I agree on refactoring to remove nulls. I disagree that this is complicated. Why can't we just remove the `null` and be that? What more needs to be involved that makes this complicated.

Instead of doing all at once, just do one by one when we run into it with pitest.

romani:
I have no time for this update now, I still have macros project as not finished. 

All is doable just takes a bit of planning, investigation, scope definition, ..... . 

We can have this battle later on, we need to finish projects and clean up our PRs list,  and start new project without backlog, too much is assigned to be to unblock.

romani:
Complications of to think to make breaking changes or not. 
https://checkstyle.org/filters/suppressionxpathsinglefilter.html#Properties 
All nulls.
We usually break something when we have a reason, now reason is weak.
This fix just bubble up problem that run into before and decided to postpone and even now I do not have energy to think on this and define what it should be.

Values in xdoc are values that user can set to java beans, so our API is java beans. Xml config is just convenience format, the only firmat, and unfortunately it does not have null. We already have somewhere idea to support json like config, and it will have null definition :) so we will be able to cover such mutation by json config implementation.

I am in position do not break, do not be involved in new project, and simple put coverage for pitest. Imagine we need this test for jacoco, it is same.

I can put extra comment on this test that it is required until we support config with null as first class element (json).

Each module should get explicitly provided default values, and it should work same as no values provided.

>What plugins do this hack with modules and nulls?

I do not have exact links, Imagine you render some form for user, there are some fields, you already put default values in them to show user what is default, it is convenient for module creation you just take data entry fields values and set in bean.
If you need to deal with nulls , it is unfortunate.

Issue #13626, opened 2023-08-25
Rewrite metadata module


stoyanK7:
One of the goals of this Xdoc automation project is to gut the Javadocs. The problem is that they can not be deleted immediately. The metadata scraper relies on these Javadocs to generate the XML metadata files:
https://github.com/checkstyle/checkstyle/blob/98e74d072a8c20ab4cf9f0fdcb35c3cbdb4e7aaf/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java#L48-L52

After the properties macro (https://github.com/checkstyle/checkstyle/pull/13407) is merged, we will have all the functionality to make this rewrite possible - properties macro,  parent module macro, violation messages macro. Additionally, a module will have to be written to get the description from the Xdoc.

All this functionality can be moved to a shared module where the macros and the metadata generator can be used.

After completion of this issue, the Javadocs of all modules can be removed.

romani:
I recommend to tackle this issue in few steps, to swap functionality step by step.

romani:
@SteLeo1602 , please confirm that you fixed this issue.

SteLeo1602:
> [@SteLeo1602](https://github.com/SteLeo1602) , please confirm that you fixed this issue.

I do confirm that I have gutted javadocs for all of the modules, but there is still an [issue ](https://github.com/checkstyle/checkstyle/issues/17746) where the code optimization has to occur.

romani:
Thanks a lot.
Separate issue is still open.
I am closing this epic issue.

Issue #13722, opened 2023-09-09
Decide on how to process validation-type


stoyanK7:
In https://github.com/checkstyle/checkstyle/pull/13670 we reuse functionality from site generation in the metadata module in an effort to start removing Javadocs.

One property that is mentioned in Javadocs but we do not have a way to extract is `validation-type`
https://github.com/checkstyle/checkstyle/blob/c2e65aa2e9f6854f3902c8c018301ba97e8a1706/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/DescendantTokenCheck.xml#L19-L22

https://github.com/checkstyle/checkstyle/blob/c2e65aa2e9f6854f3902c8c018301ba97e8a1706/src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java#L46-L49

How should we proceed with the extraction of this property?

---
One idea is to mention it as a Javadoc param above the setter and get it from there:
```java
    /**
     * Setter to specify set of tokens with limited occurrences as descendants.
     *
     * @param limitedTokensParam tokens to ignore.
     * @validationType tokenTypesSet
     * @since 3.2
     */
    public void setLimitedTokens(String... limitedTokensParam) {
        ....
    }
```

same problems is blocking macros usage at MultipleStringLiterals https://github.com/checkstyle/checkstyle/pull/13721

romani:
It was in javadoc of class as tag, let's keep it as javadoc tag, but in property javadoc.

rnveach:
We shouldn't need to do anything more special than what we are doing now.

The field declares it is an array of tokens.
https://github.com/checkstyle/checkstyle/blob/f22c784ff4d2ae8a5dc8d4ffbed2b3cc6b641c8c/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MultipleStringLiteralsCheck.java#L110-L111
Meta is suppose to be using XDOC's logic in the "final" form.

https://github.com/checkstyle/checkstyle/blob/f22c784ff4d2ae8a5dc8d4ffbed2b3cc6b641c8c/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1473-L1474
`BitSet` is always treated as int array.
[...]

PR #13721, opened 2023-09-08
usageof propery macros for MultipleStringLiterals generate bad link to property type

romani:
Issue #13693

problem detected at https://github.com/checkstyle/checkstyle/pull/13717#issuecomment-1710874863

we need to fix link.
This PR is just to show a problem and to share a code to reuse for functional fix.

PR #13812, opened 2023-10-02
Issue #13693: use marco property in module templates

romani:
Issue #13693

no updates in javadoc until #13626

this is RP to agree on, if all good, other PR, will be batched with 5-7 modules.

### DIFF
--- a/src/xdocs/checks/annotation/annotationusestyle.xml
+++ b/src/xdocs/checks/annotation/annotationusestyle.xml
@@ -87,43 +87,23 @@
             </tr>
             <tr>
               <td>closingParens</td>
-              <td>
-                Define the policy for ending parenthesis.
-              </td>
-              <td>
-                <a href="../../property_types.html#ClosingParensOption">ClosingParensOption</a>
-              </td>
-              <td>
-                <code>never</code>
-              </td>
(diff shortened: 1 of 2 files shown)


rnveach:
If we are going to work on expanding the properties macro, why don't we fix the issue identified with it creating the URLs?

romani:
Please share link to such issue

rnveach:
https://github.com/checkstyle/checkstyle/pull/13721
https://github.com/checkstyle/checkstyle/issues/13722

romani:
Thanks a lot.
I will skip updates in modules that has affected types of property, where link is changing.
We are waiting for such issue to be resolved.

But let's keep merging updates where link is not changing, and just some html indentation affected.

romani:
@rnveach , I can reduce batch size to 5  to make it easier to review.
I such PRs we just need to be sure that links are same, I will try to be attentive and put on a side issues problematic cases.

rnveach:
There wouldn't be a concern if we just fixed the problem sooner than later.

rnveach (approved):

romani:
@rnveach , I try to help in this project only for no brain activity, and keep my brain time only for pitest project.

To let both project to be finished.

nrmancuso (approved):

Issue #13809, opened 2023-10-01
Resolve pitest survivals in modules - 5


Kevin222004:
Split from 13321
Part of https://github.com/checkstyle/checkstyle/issues/13089

# 4) [Pitest-Filters](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-filters-suppressions.xml)

**1) SuppressionCommentFilter**

- [x] NonVoidMethodCallMutator (1) 
        https://github.com/checkstyle/checkstyle/pull/13833

------------

 **3) SuppressFilterElement**
 
- [x] NonVoidMethodCallMutator (3)
        https://github.com/checkstyle/checkstyle/pull/13841 (Covered 3 mutation)
- [x] MemberVariableMutator (6)   
       https://github.com/checkstyle/checkstyle/pull/13841 (Covered 6 Mutation)

-----------
 
**5) SuppressWithPlainTextCommentFilter**
 
- [x] NonVoidMethodCallMutator (2)
       https://github.com/checkstyle/checkstyle/pull/13888
       
-------
 
**7) XpathFilterElement**

- [x] NonVoidMethodCallMutator (3)
      https://github.com/checkstyle/checkstyle/pull/13841 (Covered 3 Mutation)
- [x] MemberVariableMutator (3) 
       https://github.com/checkstyle/checkstyle/pull/13841 (Covered 3 Mutation)
- [x] MemberVariableMutator (7)
https://github.com/checkstyle/checkstyle/pull/13841 (Covered 7 mutation) 
 
----------
 
**8) SuppressWithPlainTextCommentFilter$Suppression**
 
- [x] MemberVariableMutator (2)
       https://github.com/checkstyle/checkstyle/pull/13888
       
-------------

# 6) [Pitest-Common](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-common-suppressions.xml)

**1) PackageObjectFactory**

some related issue https://github.com/checkstyle/checkstyle/issues/11902

- [x] NonVoidMethodCallMutator (1)
       https://github.com/checkstyle/checkstyle/pull/13924
- [x] RemoveConditionalMutator_EQUAL_IF (2)
       https://github.com/checkstyle/checkstyle/pull/13924 
- [x] NakedReceiverMutator (1)
        https://github.com/checkstyle/checkstyle/pull/13927 

-----------

**3) ConfigurationLoader$InternalLoader**

[...]

PR #13821, opened 2023-10-04
Issue #13809: kill mutation for JavadocPropertiesGenerator

Kevin222004:
Issue #13809: kill mutation for JavadocPropertiesGenerator

# Mutation

https://github.com/checkstyle/checkstyle/blob/a7fa83a3de92486dd7033b45bc5364ca3eae9725/config/pitest-suppressions/pitest-common-2-suppressions.xml#L111-L118

# Explaintaion
Test

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/JavadocPropertiesGeneratorTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/JavadocPropertiesGeneratorTest.java
@@ -355,4 +355,23 @@ public class JavadocPropertiesGeneratorTest extends AbstractPathTestSupport {
         }
     }
 
+    @Test
+    public void testGetFirstJavadocSentence(@SysErr Capturable systemErr,
+                                            @SysOut Capturable systemOut) throws Exception {
+        final String expectedContent = "EOF1=First Javadoc Sentence.";
+
+        JavadocPropertiesGenerator.main(
+                getPath("InputJavadocPropertiesGeneratorGetFirstJavadocSentence.java"),
+            "--destfile", DESTFILE_ABSOLUTE_PATH);
+        assertWithMessage("Unexpected error log")
+            .that(systemErr.getCapturedData())
+            .isEqualTo("");
+        assertWithMessage("Unexpected output log")
+            .that(systemOut.getCapturedData())
+            .isEqualTo("");
+        final String fileContent = FileUtils.readFileToString(DESTFILE, StandardCharsets.UTF_8);
+        assertWithMessage("File content is not expected")
+            .that(fileContent.trim())
+            .isEqualTo(expectedContent.trim());
+    }
 }
--- a/config/pitest-suppressions/pitest-common-2-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-2-suppressions.xml
@@ -13,15 +13,6 @@
   <mutation unstable="false">
     <sourceFile>JavadocPropertiesGenerator.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.JavadocPropertiesGenerator</mutatedClass>
-    <mutatedMethod>getFirstJavadocSentence</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/api/DetailAST::getNextSibling</description>
-    <lineContent>child = child.getNextSibling()) {</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>JavadocPropertiesGenerator.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.JavadocPropertiesGenerator</mutatedClass>
     <mutatedMethod>main</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
     <description>replaced call to picocli/CommandLine::setUsageHelpWidth with receiver</description>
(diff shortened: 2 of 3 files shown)


romani:
This is internal class to generate one txt file.
If content of file is same, just apply mutation to make pitest happy.

We do not need to test this class much, it doesn't execute on unpredictable user code.

Kevin222004:
As per the exisiting test if we follow the mutation then it must be timeout but the mutation is surviving.

romani:
Tests are failing https://app.travis-ci.com/github/checkstyle/checkstyle/builds/266475895#L1555

romani:
this class is executed at
https://github.com/checkstyle/checkstyle/blob/d74f431438e0568ca1ba06e9ec789580eaf7416d/pom.xml#L781-L797

result file is not in git, so please show CLI prove that content of it is not changed.

src/test/resources/com/puppycrawl/tools/checkstyle/javadocpropertiesgenerator/InputJavadocPropertiesGeneratorCorrect2.java
```diff
@@ -0,0 +1,16 @@
+package com.puppycrawl.tools.checkstyle.javadocpropertiesgenerator;
+
+public class InputJavadocPropertiesGeneratorCorrect2 {
+
+    // note: &#064; is the html escape for '@',
+    // used here to avoid confusing the javadoc tool
+    /**
+     * An annotation of a package, type, field, parameter or variable.
+     * An annotation may occur anywhere modifiers occur (it is a
+     * type of modifier) and may also occur prior to a package definition.
```
romani, line 10: Can we remove text that does not matter.
If we care about first line, let's keep only it.

src/test/resources/com/puppycrawl/tools/checkstyle/javadocpropertiesgenerator/InputJavadocPropertiesGeneratorCorrect2.java
```diff
@@ -0,0 +1,16 @@
...
+     * The notable children are: The annotation name and either a single
+     * default annotation value or a sequence of name value pairs.
+     * Annotation values may also be annotations themselves.
+     */
+    public static final @Deprecated int EOF1 = 1, EOF2 = 2;
```
romani, line 15: We never Going to do this if our token Java class.
Having same javadoc for different token is nonsense.
Technically possible, but nonsense.
Why we need test for nonsense ?
This generator class is internal tool, it should work only on out TokenTypes.java.

 Is there a way to refactor java code to avoid mutation survival and not supporting of this type of javadoc on multiple?
I would rather have let java code to throw exception in such code to clearly highlight it not good way to write a code in token class.

romani (changes requested):
Items:

src/test/resources/com/puppycrawl/tools/checkstyle/javadocpropertiesgenerator/InputJavadocPropertiesGeneratorCorrect2.java
Kevin222004: I have make it simple.
the difference is becuase of the single line comment above the javadoc.

src/test/resources/com/puppycrawl/tools/checkstyle/javadocpropertiesgenerator/InputJavadocPropertiesGeneratorCorrect2.java
Kevin222004: done

src/test/java/com/puppycrawl/tools/checkstyle/JavadocPropertiesGeneratorTest.java
```diff
@@ -355,4 +355,23 @@ public void testParseError() throws Exception {
         }
     }
 
+    @Test
+    public void testCorrect2(@SysErr Capturable systemErr, @SysOut Capturable systemOut)
```
romani, line 359: Make method name to describe a reason.
Same for Input file name.
And if more wording is required you can put it in comment of Input.

src/test/resources/com/puppycrawl/tools/checkstyle/javadocpropertiesgenerator/InputJavadocPropertiesGeneratorCorrect2.java
```diff
@@ -0,0 +1,10 @@
+package com.puppycrawl.tools.checkstyle.javadocpropertiesgenerator;
+
+public class InputJavadocPropertiesGeneratorCorrect2 {
+
+    // used here to avoid confusing the javadoc tool
+    /**
+     * An annotation of a package, type, field, parameter or variable.
```
romani, line 7: Put here some words to explain reason of this test

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/JavadocPropertiesGeneratorTest.java
Kevin222004: done

src/test/resources/com/puppycrawl/tools/checkstyle/javadocpropertiesgenerator/InputJavadocPropertiesGeneratorCorrect2.java
Kevin222004: done

romani (approved):
Ok to merge

romani (changes requested):
CI must be green

romani (approved):
Ok to merge

PR #13826, opened 2023-10-04
Issue #13809: kill mutation for AuditEventDefaultFormatter

Kevin222004:
Issue #13809: kill mutation for AuditEventDefaultFormatter

# Explaintation

I didn't find any test case for this and the reason may be in StringBuilder even after we assign the capacity if the character or string goes beyond the limit of assigned capacity. it will automatically expands the necessary memory to accommodate character to string builder.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4372,6 +4372,8 @@
               <excludedMethods>
                 <!-- caching of search results, big performance optimization -->
                 <param>lazyLoad</param>
+                <!-- performance optimization -->
+                <param>initStringBuilderWithOptimalBuffer</param>
               </excludedMethods>
               <coverageThreshold>100</coverageThreshold>
               <mutationThreshold>96</mutationThreshold>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/AuditEventDefaultFormatter.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/AuditEventDefaultFormatter.java
@@ -55,9 +55,7 @@ public class AuditEventDefaultFormatter implements AuditEventFormatter {
             severityLevelName = severityLevel.getName().toUpperCase(Locale.US);
         }
 
-        // Avoid StringBuffer.expandCapacity
-        final int bufLen = calculateBufferLength(event, severityLevelName.length());
-        final StringBuilder sb = new StringBuilder(bufLen);
+        final StringBuilder sb = initStringBuilderWithOptimalBuffer(event, severityLevelName);
 
         sb.append('[').append(severityLevelName).append("] ")
             .append(fileName).append(':').append(event.getLine());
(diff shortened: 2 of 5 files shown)


romani:
inspection is not happy - https://app.circleci.com/pipelines/github/checkstyle/checkstyle/21119/workflows/2dfef869-c3b0-470b-b761-a7c7e3fefc70/jobs/398013/artifacts

src/main/java/com/puppycrawl/tools/checkstyle/AuditEventDefaultFormatter.java
```diff
@@ -55,9 +52,7 @@ public String format(AuditEvent event) {
             severityLevelName = severityLevel.getName().toUpperCase(Locale.US);
         }
 
-        // Avoid StringBuffer.expandCapacity
-        final int bufLen = calculateBufferLength(event, severityLevelName.length());
-        final StringBuilder sb = new StringBuilder(bufLen);
+        final StringBuilder sb = new StringBuilder();
```
romani, line 55: this is perofrmance optimization.
Lets add this method to pom.xml for pitest to avoid.
Method of this method is pretty uniq, it is good.

romani (changes requested):
items:

pom.xml
```diff
@@ -4369,6 +4369,7 @@
                 <avoidCallsTo>com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil</avoidCallsTo>
               </avoidCallsTo>
               <excludedMethods>
+                <param>format</param>
```
romani, line 4372: I meant calculateBufferLength

romani (changes requested):
Items

pom.xml
Kevin222004: sorry 

Done

Kevin222004: Not done. it is not mutated because of method calculateBufferLength. 
If I put the method calculateBufferLength then only this method will be ignored. format method is implemented only in 2 classes 
1) in this AuditEventDefaultFormatter 
2) https://github.com/checkstyle/checkstyle/blob/9e931263c2eb441fa7f893001e18005b1a2097a5/src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/AuditEventUtFormatter.java#L35 --> for which we are not doing any mutation testing.

I have also tried to use `<avoidCallsTo>java.lang.StringBuilder</avoidCallsTo>`  This is also not working.

romani:
@Kevin222004 , please move forward to other mutation s, I will handle this update to figure out how to avoid mutation.

romani (approved):
ok to merge.

I did this change :).

Kevin222004:
@romani thanks, I have Fixed the ci

rdiachenko (approved):
lgtm

PR #13841, opened 2023-10-07
Issue #13809: Kill mutation in Filters

Kevin222004:
Issue #13809: Kill mutation in Filters

this is typical extra data storage that pitest catch as not critical for execution.
I think it is fine to remove it here as cost of taking String value of regexp from Pattern object is cheap, as it is stored as is (String) in Patter class (it is single getter to get String on Pattern class).

https://github.com/openjdk/jdk/blob/ecd25e7d6f9d69f9dbdbff0a4a9b9d6b19288593/src/java.base/share/classes/java/util/regex/Pattern.java#L1138-L1144

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
@@ -43,20 +43,11 @@ public class SuppressFilterElement
     /** The regexp to match file names against. */
     private final Pattern fileRegexp;
 
-    /** The pattern for file names. */
-    private final String filePattern;
-
     /** The regexp to match check names against. */
     private final Pattern checkRegexp;
 
-    /** The pattern for check class names. */
-    private final String checkPattern;
-
     /** The regexp to match message names against. */
     private final Pattern messageRegexp;
-
-    /** The pattern for message names. */
-    private final String messagePattern;
 
     /** Module id filter. */
     private final String moduleId;
@@ -86,21 +77,18 @@ public class SuppressFilterElement
      */
     public SuppressFilterElement(String files, String checks,
                            String message, String modId, String lines, String columns) {
-        filePattern = files;
         if (files == null) {
             fileRegexp = null;
         }
         else {
             fileRegexp = Pattern.compile(files);
         }
-        checkPattern = checks;
         if (checks == null) {
             checkRegexp = null;
         }
         else {
             checkRegexp = Pattern.compile(checks);
         }
-        messagePattern = message;
         if (message == null) {
             messageRegexp = null;
         }
@@ -136,30 +124,9 @@ public class SuppressFilterElement
      */
     public SuppressFilterElement(Pattern files, Pattern checks, Pattern message, String moduleId,
             String lines, String columns) {
-        if (files == null) {
-            filePattern = null;
-            fileRegexp = null;
-        }
-        else {
-            filePattern = files.pattern();
-            fileRegexp = files;
-        }
-        if (checks == null) {
-            checkPattern = null;
-            checkRegexp = null;
-        }
-        else {
-            checkPattern = checks.pattern();
-            checkRegexp = checks;
-        }
-        if (message == null) {
-            messagePattern = null;
-            messageRegexp = null;
-        }
-        else {
-            messagePattern = message.pattern();
-            messageRegexp = message;
-        }
+        fileRegexp = files;
+        checkRegexp = checks;
+        messageRegexp = message;
         this.moduleId = moduleId;
         if (lines == null) {
             linesCsv = null;
@@ -224,8 +191,8 @@ public class SuppressFilterElement
 
     @Override
     public int hashCode() {
-        return Objects.hash(filePattern, checkPattern, messagePattern, moduleId, linesCsv,
-                columnsCsv);
+        return Objects.hash(getPatternSafely(fileRegexp), getPatternSafely(checkRegexp),
+                getPatternSafely(messageRegexp), moduleId, linesCsv, columnsCsv);
     }
 
     @Override
@@ -237,12 +204,29 @@ public class SuppressFilterElement
             return false;
         }
         final SuppressFilterElement suppressElement = (SuppressFilterElement) other;
-        return Objects.equals(filePattern, suppressElement.filePattern)
-                && Objects.equals(checkPattern, suppressElement.checkPattern)
-                && Objects.equals(messagePattern, suppressElement.messagePattern)
+        return Objects.equals(getPatternSafely(fileRegexp),
+                    getPatternSafely(suppressElement.fileRegexp))
+                && Objects.equals(getPatternSafely(checkRegexp),
+                    getPatternSafely(suppressElement.checkRegexp))
+                && Objects.equals(getPatternSafely(messageRegexp),
+                    getPatternSafely(suppressElement.messageRegexp))
                 && Objects.equals(moduleId, suppressElement.moduleId)
                 && Objects.equals(linesCsv, suppressElement.linesCsv)
                 && Objects.equals(columnsCsv, suppressElement.columnsCsv);
     }
 
+    /**
+     * Util method to get pattern String value from Pattern object safely, return null if
+     * pattern object is null.
+     *
+     * @param pattern pattern object
+     * @return value of pattern or null
+     */
+    private static String getPatternSafely(Pattern pattern) {
+        String result = null;
+        if (pattern != null) {
+            result = pattern.pattern();
+        }
+        return result;
+    }
 }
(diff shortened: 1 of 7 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
```diff
@@ -105,7 +105,7 @@ public SuppressFilterElement(String files, String checks,
             messageRegexp = null;
         }
         else {
-            messageRegexp = Pattern.compile(message);
+            messageRegexp = Pattern.compile(messagePattern);
```
Kevin222004, line 108: This will not make any cause in code as messagePattern is already equal to message

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
```diff
@@ -111,23 +111,23 @@ public XpathFilterElement(Pattern files, Pattern checks, Pattern message,
         }
         else {
             filePattern = files.pattern();
-            fileRegexp = files;
+            fileRegexp = Pattern.compile(filePattern);
```
rnveach, line 114: This change makes no sense to de-pattern a pattern then to re-pattern it right back again.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
romani: @Kevin222004 , try to remove fields of type String that are string representation of compiled pattern.
all their usage in hashCode/equals change to `xxxx.pattern()`.

romani: @Kevin222004 , can you split this PR in two ? to let us merge what is easy and discuss/review deeply only narrow set of cases

romani: for null cases make method `getPatternSafely(pattern) { if (pattern == null) { return null} else {return patter.patern()}}
﻿` 
to make `xxxx.pattern()` without exceptions.

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
```diff
@@ -237,27 +206,4 @@ private List<Item> getItems(TreeWalkerAuditEvent event) {
         }
         return items;
     }
-
-    @Override
-    public int hashCode() {
-        return Objects.hash(filePattern, checkPattern, messagePattern,
```
romani, line 243: please restore such methods and do `filePattern.patern(),.....` to simulate how it was before.

romani (changes requested):
items:

romani:
@Kevin222004 , please consider https://github.com/checkstyle/checkstyle/pull/13903 , please reuse this code to make CI green

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
```diff
@@ -85,22 +76,19 @@ public class SuppressFilterElement
      * @param columns columns CSV values and ranges for column number filtering.
      */
     public SuppressFilterElement(String files, String checks,
-                           String message, String modId, String lines, String columns) {
-        filePattern = files;
+                                 String message, String modId, String lines, String columns) {
```
romani, line 79: Restore indentation to minimize diff

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
```diff
@@ -38,26 +38,17 @@
  *
  */
 public class SuppressFilterElement
-    implements Filter {
+        implements Filter {
```
romani, line 41: Restore indentation to minimize diff

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
romani: Done

romani:
@Kevin222004 , please update Checker

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
Kevin222004: Done

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
Kevin222004: done

romani:
@Kevin222004 , please remove second commit, make PR ready for review and merge.
Regression testing for this update doesn't not required and doesn't make sense, as this purely config nuances.

Kevin222004:
OK I just need diff of checker.

romani:
For now, let's just suppress Checker.

romani:
pitest passed on my local, so we are moving forward with review.

romani (approved):
ok to merge.

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
```diff
@@ -136,46 +124,25 @@ public SuppressFilterElement(String files, String checks,
...
+        messageRegexp = message;
         this.moduleId = moduleId;
         if (lines == null) {
             linesCsv = null;
             lineFilter = null;
         }
         else {
             linesCsv = lines;
-            lineFilter = new CsvFilterElement(lines);
+            lineFilter = new CsvFilterElement(linesCsv);
```
rnveach, line 137: Why can't we leave this as it was before? It is logically equivalent.
Why don't we remove `linesCsv` if it is always the assignment of `lines`?

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
```diff
@@ -136,46 +124,25 @@ public SuppressFilterElement(String files, String checks,
...
         }
         if (columns == null) {
             columnsCsv = null;
             columnFilter = null;
         }
         else {
             columnsCsv = columns;
-            columnFilter = new CsvFilterElement(columns);
+            columnFilter = new CsvFilterElement(columnsCsv);
```
rnveach, line 145: Same.

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
```diff
@@ -253,11 +223,28 @@ public boolean equals(Object other) {
...
         final XpathFilterElement xpathFilter = (XpathFilterElement) other;
-        return Objects.equals(filePattern, xpathFilter.filePattern)
-                && Objects.equals(checkPattern, xpathFilter.checkPattern)
-                && Objects.equals(messagePattern, xpathFilter.messagePattern)
+        return Objects.equals(getPatternSafely(fileRegexp),
+                getPatternSafely(xpathFilter.fileRegexp))
+                && Objects.equals(getPatternSafely(checkRegexp),
+                getPatternSafely(xpathFilter.checkRegexp))
+                && Objects.equals(getPatternSafely(messageRegexp),
+                getPatternSafely(xpathFilter.messageRegexp))
```
rnveach, line 231: lines starting with `getPatternSafely` should indented +4 over the other lines.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
romani: restored.

romani: we use it to store a value, may be we can take it from CsvFilterElement but it is abit more refactoring with no much value.

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
romani: restored

src/main/java/com/puppycrawl/tools/checkstyle/filters/XpathFilterElement.java
romani: done

rdiachenko (approved):
lgtm

romani:
@Kevin222004 , please fix Checker https://github.com/checkstyle/checkstyle/actions/runs/6616385677/job/17970549707?pr=13841#step:6:640

romani:
I can not reproduce it on local, restarted.

romani:
@rnveach , CI is green, please finish review

rnveach (approved):

Vyom-Yadav (approved):
LGTM!

nrmancuso (approved):

PR #13843, opened 2023-10-07
Issue #13693: migrate filter xdoc termplates to property macros

romani:
Issue #13693

### DIFF
--- a/src/xdocs/filters/suppressioncommentfilter.xml
+++ b/src/xdocs/filters/suppressioncommentfilter.xml
@@ -57,7 +57,7 @@
             </tr>
             <tr>
               <td>checkC</td>
-              <td>Control whether to check C style comments (<code>/* ... */</code>).</td>
+              <td>Control whether to check C style comments (<code>&#47;* ... *&#47;</code>).</td>
               <td><a href="../property_types.html#boolean">boolean</a></td>
               <td><code>true</code></td>
               <td>3.5</td>
--- a/config/checkstyle-non-main-files-suppressions.xml
+++ b/config/checkstyle-non-main-files-suppressions.xml
@@ -566,27 +566,8 @@
          files="src[\\/]xdocs[\\/]checks[\\/]whitespace[\\/]whitespaceafter.xml.template"/>
   <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]checks[\\/]whitespace[\\/]whitespacearound.xml.template"/>
-  <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]filefilters[\\/]beforeexecutionexclusionfilefilter.xml.template"/>
-  <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]filters[\\/]severitymatchfilter.xml.template"/>
-  <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]filters[\\/]suppressioncommentfilter.xml.template"/>
-  <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]filters[\\/]suppressionfilter.xml.template"/>
-  <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]filters[\\/]suppressionsinglefilter.xml.template"/>
-  <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]filters[\\/]suppressionxpathfilter.xml.template"/>
-  <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]filters[\\/]suppressionxpathsinglefilter.xml.template"/>
+  <!-- no properties by design -->
   <suppress id="propertiesMacroMustExist"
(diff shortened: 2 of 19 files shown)


src/xdocs/filters/suppressioncommentfilter.xml
```diff
@@ -57,7 +57,7 @@
             </tr>
             <tr>
               <td>checkC</td>
-              <td>Control whether to check C style comments (<code>/* ... */</code>).</td>
+              <td>Control whether to check C style comments (<code>&#47;* ... *&#47;</code>).</td>
```
rnveach, line 60: Why?

romani: Looks like simple escaping logic,  `/` is part of special symbols in xml.
Xdoc becomes more and more generated , do you think we need to care about this?

rnveach: No, but escaping is usually reserved for if something could break without escaping, like `<` in html as its used as the start of a tag and a parser could mistake it as one.

I don't see why `/` would hurt something and need to be escaped.

romani: Xdoc files receiving long lines now as wrapping is not easy.

rnveach: @romani Let me know if this isn't something you can look into. As I said, I don't think this is an issue, but I am wondering what prompted doxia to escape this. I assume it is no code of ours that is causing this to be escaped.

romani: @nrmancuso , please share your thoughts on this side effect.

romani: @rnveach , this is not an escaping, I did not find escaping logic in macros or any util.
it is content like it is in javadoc comment:
https://github.com/checkstyle/checkstyle/blob/0685947a4b9618a5b1f0cb518b5ddd2fc00d6fed/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressionCommentFilter.java#L75

nrmancuso: This should be directly translated from https://github.com/checkstyle/checkstyle/blob/0685947a4b9618a5b1f0cb518b5ddd2fc00d6fed/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressionCommentFilter.java#L149, escaping is required for comment end delimiter to be in Javadoc comment.

nrmancuso:
GitHub, generate site

src/xdocs/filters/suppressioncommentfilter.xml
rnveach: > escaping is required for comment end delimiter to be in Javadoc comment.

Ahhh, this explains it. Our macro pulls in the text to display directly from the source files. So whatever we have escaped in our Javadoc will be escaped the same way in the Doxia, unless we undo the manual escaping somewhere.

Unless we do something special, I assume Javadoc macros ( Example `{@` ) are not converted to doxia friendly tags?

As @nrmancuso started doing at https://github.com/checkstyle/checkstyle/pull/13843#issuecomment-1753030514 , we need to see if this is displayed properly in the site or if it will be a problem.

romani: @rnveach , please finish review.

romani:
GitHub, generate website

nrmancuso (approved):
Generated website looks good.

src/xdocs/filters/suppressioncommentfilter.xml
rnveach: We do some javadoc conversions at https://github.com/checkstyle/checkstyle/blob/ede308a9724a3a9fe1e0eab89da86f1d3ae110f5/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java#L1129 like `{@code`.

Example: https://github.com/checkstyle/checkstyle/blob/ede308a9724a3a9fe1e0eab89da86f1d3ae110f5/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java#L1175-L1177

rnveach: The following are the tokens this method apparently doesn't recognize and does nothing for them:
````
ATTRIBUTE
HTML_ELEMENT_END
HTML_ELEMENT_START
HTML_TAG
JAVADOC_INLINE_TAG
JAVADOC_INLINE_TAG_START
LEADING_ASTERISK
NEWLINE
WS
````

rnveach (approved):

src/xdocs/filters/suppressioncommentfilter.xml
romani: @rnveach ,  we can not afford now support of any kind of javadoc. We need to make it work for cases we have now first of all. 
Migration process with bubble up them, I already found few more edge cases.

rnveach: Those token supports are not needed, but the issue is this utility will not report errors on tokens we skip over and do nothing with. It could be possible to misconvert the javadoc because of this. Unless we are going to review every text in detail that there are no discrepancies, I think the utility needs a bit of an update.

PR #13861, opened 2023-10-09
Issue #13693: migrate part 1 of coding package to property macros

romani:
Issue #13693

PR #13863, opened 2023-10-10
infra: do not allow in-progress site action to be cancelled by skipped run

nrmancuso:
Noticed at https://github.com/checkstyle/checkstyle/pull/13861#issuecomment-1755265092, skipped runs count towards concurrency. Updating the `cancel-in-progress` configuration to be `false` will protect in-progress actions from being cancelled.

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -27,7 +27,7 @@ permissions:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.issue.number || github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 jobs:
   parse_pr_info:


romani:
I do not understand this, please explain a bit more.

rnveach:
I am assuming if the job is already running, and someone comes in and requests their own PR site to be run, that it cancels the first job thinking it is a duplicate.

If so, my only question is can we make it so, it has to be the same PR to be considered a duplicate?

nrmancuso:
The issue is that GitHub "runs" actions triggered by comments for every comment in a PR, even if the comment doesn't contain the trigger words. So if I intentionally trigger a report to be generated, then make some comment (with any content) after that, GitHub will cancel the report that is is progress. I do not remember this behavior previously, it seems like something changed on GitHub's side.

> it has to be the same PR to be considered a duplicate?

Yes, this is what the current settings do, except they cancel whatever is in progress. When I originally set this up, the idea was that the site action runs quickly and this wouldn't matter, while also preventing duplicate executions.

The changes in this PR will just cue the workflows, so that comments following the site generation comment will be "skipped" after the existing site workflow.

romani (approved):
Please put pull number in commit to quickly lookup this decision. I think one day someone will try to change it back.

rnveach:
@nrmancuso Thanks for the explanation. So basically the trigger word is checked last, while the group and cancel in progress are checked first, and this is done on every post in a PR.

I am good from my side to approved this. See https://github.com/checkstyle/checkstyle/pull/13863#pullrequestreview-1678115935 .

Is there any way now to either prevent duplicate runs in the PR , or cancel the previous run if someone accidentally triggers the keyword with this change?

nrmancuso:
>I assume every post includes updated posts, or is it just brand new posts?

Actions triggered by comment can be trigged by edit, yes.

>Is there any way now to either prevent duplicate runs in the PR 

Define a duplicate run

> cancel the previous run if someone accidentally triggers the keyword twice with this change?

You can do this through the GitHub actions UI.

rnveach:
> You can do this through the GitHub actions UI.

I was referring to automatically so we don't have to be watching 24/7.

rnveach:
Yes, which is why I was asking if it can still be restored after this PR is merged, or are we just stuck?

rnveach (approved):

PR #13870, opened 2023-10-11
Issue #13809: kill mutation for JavadocPropertiesGenerator-2

Kevin222004:
Issue #13809: kill mutation for JavadocPropertiesGenerator-2

# Mutation

https://github.com/checkstyle/checkstyle/blob/649b669336ebdb914936b65e582eb00a5a9ed17f/config/pitest-suppressions/pitest-common-2-suppressions.xml#L13-L20

# Explaintaion

we don't need help when we run cmd from the config


romani (approved):
Yes, we do not need help rendering beautification for internal execution as execution not done by humans.

It was done during migration as sweeping update for all CLI like execution.

PR #13900, opened 2023-10-18
Issue #13809: Kill mutation ConfigurationLoader2

Kevin222004:
Issue #13809: Kill mutation ConfigurationLoader2

Test

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
@@ -769,4 +769,24 @@ public class ConfigurationLoaderTest extends AbstractPathTestSupport {
                     .contains("query");
         }
     }
+
+    @Test
+    public void testDefaultValuesForNonDefinedProperties() throws Exception {
+        final Properties props = new Properties();
+        props.setProperty("checkstyle.charset.base", "UTF");
+
+        final File file = new File(
+                getPath("InputConfigurationLoaderDefaultProperty.xml"));
+        final DefaultConfiguration config =
+            (DefaultConfiguration) ConfigurationLoader.loadConfiguration(
+                    file.toURI().toString(), new PropertiesExpander(props));
+
+        final Properties expectedPropertyValues = new Properties();
+        expectedPropertyValues.setProperty("tabWidth", "2");
+        expectedPropertyValues.setProperty("basedir", ".");
+        // charset property uses 2 variables, one is not defined, so default becomes a result value
+        expectedPropertyValues.setProperty("charset", "ASCII");
+        verifyConfigNode(config, "Checker", 0, expectedPropertyValues);
+    }
+
 }
--- a/src/xdocs/config.xml
+++ b/src/xdocs/config.xml
@@ -245,6 +245,11 @@
       </source>
 
       <p>
+        ATTENTION: default value is applied to module property value if at least one expansion
+        property is not defined.
+      </p>
+
+      <p>
         This feature is a great help when setting up a centralized
         configuration file (e.g. one file for the whole company) to
         lower configuration maintenance costs.  Projects that are happy
--- a/config/pitest-suppressions/pitest-common-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-suppressions.xml
@@ -28,15 +28,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>ConfigurationLoader.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.ConfigurationLoader</mutatedClass>
-    <mutatedMethod>replaceProperties</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to java/lang/StringBuilder::length</description>
-    <lineContent>sb.replace(0, sb.length(), defaultValue);</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>LocalizedMessage.java</sourceFile>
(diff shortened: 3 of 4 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
```diff
@@ -769,4 +769,20 @@ public void testLoadConfiguration3() throws Exception {
                     .contains("query");
         }
     }
+
+    @Test
+    public void testLengthInReplaceProperties() throws Exception {
+        final String[][] testValues = {
+                {"${a}"},
+                {"x${a}"},
```
romani, line 777: Why multi dimensional array ?

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
```diff
@@ -769,4 +769,20 @@ public void testLoadConfiguration3() throws Exception {
...
+        };
+        final Properties props = new Properties();
+        for (String[] testValue : testValues) {
+            final String value = (String) getReplacePropertiesMethod().invoke(
+                    null, testValue[0], new PropertiesExpander(props), "Default");
```
romani, line 782: Can we load whole xml file and check values of Checker property?

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
```diff
@@ -769,4 +769,20 @@ public void testLoadConfiguration3() throws Exception {
...
+            assertWithMessage("\"" + testValue[0] + "\"")
+                    .that(value)
+                    .isEqualTo("Default");
```
romani, line 785: Why `{"x${a}"}` becomes Default? 
Should  be `Default`.

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
Kevin222004: converted into single dimension

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
Kevin222004: The default value is provided as "Default" thats why

romani: Should  be `xDefault`.
As x is before variable.

romani:
Kevin requested help on my comments , we need to help him.

romani:
your code is correct by it is really hard to understand why this works like this.

lets make test full functional.
```

<!DOCTYPE module PUBLIC

<module name="Checker">
    <module name="TreeWalker">
        <module name="MemberName">
            <property name="severity" value="${severity}"  default="ignore"/>
        </module>
    </module>
</module>
 
```

load this config as it is done at https://github.com/checkstyle/checkstyle/blob/bd3a9bb9ec9cf804f628171c2e7557b0afbcfd93/src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java#L81
and confirm that property is set to default value.

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
romani: I recalling this as default is for whole property value

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
romani: need to be converted to more full functional test based on xml files

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
romani: same as above, should be like https://github.com/checkstyle/checkstyle/blob/bd3a9bb9ec9cf804f628171c2e7557b0afbcfd93/src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java#L81

romani (approved):
I updated code to required state

romani:
Github, generate website

src/xdocs/config.xml
```diff
@@ -244,6 +244,11 @@
 &lt;/module&gt;
       </source>
 
+      <p>
+        ATTENTION: default value is applied to module property value if at least one expansion
+        property is not defined.
```
rnveach, line 249: I assume this is referring to property values with multiple expansions in a single property?

This needs to be expanded. In addition why don't we add an example?

romani: Not only multiple expansion, any .

src/test/java/com/puppycrawl/tools/checkstyle/ConfigurationLoaderTest.java
```diff
@@ -769,4 +769,24 @@ public void testLoadConfiguration3() throws Exception {
                     .contains("query");
         }
     }
+
+    @Test
+    public void testDefaultValuesForNonDefinedProperties() throws Exception {
```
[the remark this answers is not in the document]
romani: Whole test class is not using our general format. I still do not feel good to migrate this also to our general for modules format.

Tests are very similar to each other.

src/xdocs/config.xml
romani: I thought about example, but it is not easy to explain and might be too technical, and looks like nobody dealing with this, so we found this nuance only during pitest.

I think we don't need to make our doc to be too much extended and hard to read.

PR #13905, opened 2023-10-18
Issue #13809: Kill mutation in Checker2

Kevin222004:
Issue #13809: Kill mutation in Checker2

https://github.com/checkstyle/checkstyle/blob/223ed3981ebf088c005dc7c8bf2d075194f77ae1/config/pitest-suppressions/pitest-common-suppressions.xml#L39-L46

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
@@ -1706,6 +1706,50 @@ public class CheckerTest extends AbstractModuleTestSupport {
         verify(checkerConfig, filePath, expected);
     }
 
+    /**
+     * This test is created to cover pitest mutation over Error message.
+     * Usually Exceptions are thrown , not Error, so we need to mock creation of Error.
+     *
+     * @noinspection SerializableInnerClassWithNonSerializableOuterClass
+     * @noinspectionreason SerializableInnerClassWithNonSerializableOuterClass - mocked file
+     *      for test does not require serialization
+     */
+    @Test
+    public void testCatchErrorWithFileName() throws Exception {
+        final String errorMessage = "Error was thrown while processing ";
+        final Error expectedError = new IOError(new InternalError(errorMessage));
+
+        final File file = new File(getPath("InputChecker.java")) {
+            private static final long serialVersionUID = 1L;
+
+            /**
+             * Test is checking catch clause when exception is thrown.
+             *
+             * @noinspection ProhibitedExceptionThrown
+             * @noinspectionreason ProhibitedExceptionThrown - we require mocked file to
+             *      throw exception as part of test
+             */
+            @Override
+            public long lastModified() {
+                throw expectedError;
+            }
+        };
+
+        final Checker checker = new Checker();
+        final List<File> filesToProcess = new ArrayList<>();
+        filesToProcess.add(file);
+        try {
+            checker.process(filesToProcess);
+            assertWithMessage("IOError is expected!").fail();
+        }
+        // -@cs[IllegalCatchExtended] Testing for catch Error is part of 100% coverage.
+        catch (Error error) {
+            assertWithMessage("Error message is not expected")
+                    .that(error.getMessage())
+                    .isEqualTo(errorMessage + file.getPath());
+        }
+    }
+
     public static class DefaultLoggerWithCounter extends DefaultLogger {
 
         private int fileStartedCount;
--- a/config/pitest-suppressions/pitest-common-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-suppressions.xml
@@ -12,15 +12,6 @@
   <mutation unstable="false">
     <sourceFile>Checker.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.Checker</mutatedClass>
-    <mutatedMethod>processFiles</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to java/io/File::getPath</description>
-    <lineContent>throw new Error(&quot;Error was thrown while processing &quot; + file.getPath(), error);</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>Checker.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.Checker</mutatedClass>
     <mutatedMethod>setSeverity</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
     <description>removed call to com/puppycrawl/tools/checkstyle/api/SeverityLevel::getInstance</description>
--- a/config/suppressions.xml
+++ b/config/suppressions.xml
@@ -12,7 +12,7 @@
   <suppress id="lineLengthXml" files="[\\/]meta[\\/]"/>
 
   <suppress checks="FileLength"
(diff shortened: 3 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -314,7 +314,8 @@ private void processFiles(List<File> files) throws CheckstyleException {
                 }
 
                 // We need to catch all errors to put a reason failure (file name) in error
-                throw new Error("Error was thrown while processing " + file.getPath(), error);
+                throw new Error("Error was thrown while processing "
+                        + file.getAbsolutePath(), error);
```
romani, line 318: Can we keep getPath ?
Looks like just extra assert is required in tests

romani (changes requested):
CI is red 

Items

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
Kevin222004: I think yes can you please guide in which sceanrio we are throwing this error

romani: I think test is failing at https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=17141&view=logs&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111&l=1322

We have full coverage so you can easily find it by message or change type of exception and some test will fail

romani:
@Kevin222004 , please resolve conflict

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
Kevin222004: restored changes

src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
```diff
@@ -1706,6 +1706,49 @@ public void testViolationMessageOnIoException() throws Exception {
         verify(checkerConfig, filePath, expected);
     }
 
+    /**
+     * Test doesn't need to be serialized.
```
romani, line 1710: `This test is created to cover pitest mutation over Error message. Usually Exceptions are thrown , not Error, so we need to mock creation of Error.`

romani (changes requested):
Good.
Small item to improve before merge:

romani:
@Kevin222004 , please update this PR, looks like easy change is required.

src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
Kevin222004: done

romani (approved):
Ok to merge

nrmancuso (changes requested):
Please remove file length suppression and synthetic test, we should be able to kill this mutation with code improvement in Checker. We can extract duplicate calls to `File#getPath` in `Checker#process` to a variable before `try` block.

romani:
```✘-1 ~/java/github/romani/checkstyle [Kevin222004/Checker2 L|✚ 2⚑ 1] 
08:15 $ git diff
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
index 83d2096..083adff 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
@@ -283,40 +283,39 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
     // -@cs[CyclomaticComplexity] no easy way to split this logic of processing the file
     private void processFiles(List<File> files) throws CheckstyleException {
         for (final File file : files) {
-            String fileName = null;
+            String absolutePath = file.getAbsolutePath();
             try {
-                fileName = file.getAbsolutePath();
                 final long timestamp = file.lastModified();
-                if (cacheFile != null && cacheFile.isInCache(fileName, timestamp)
-                        || !acceptFileStarted(fileName)) {
+                if (cacheFile != null && cacheFile.isInCache(absolutePath, timestamp)
+                        || !acceptFileStarted(absolutePath)) {
                     continue;
                 }
                 if (cacheFile != null) {
-                    cacheFile.put(fileName, timestamp);
+                    cacheFile.put(absolutePath, timestamp);
                 }
-                fireFileStarted(fileName);
+                fireFileStarted(absolutePath);
                 final SortedSet<Violation> fileMessages = processFile(file);
-                fireErrors(fileName, fileMessages);
-                fireFileFinished(fileName);
+                fireErrors(absolutePath, fileMessages);
+                fireFileFinished(absolutePath);
             }
             // -@cs[IllegalCatch] There is no other way to deliver filename that was under
             // processing. See https://github.com/checkstyle/checkstyle/issues/2285
             catch (Exception ex) {
-                if (fileName != null && cacheFile != null) {
-                    cacheFile.remove(fileName);
+                if (absolutePath != null && cacheFile != null) {
+                    cacheFile.remove(absolutePath);
                 }
 
                 // We need to catch all exceptions to put a reason failure (file name) in exception
                 throw new CheckstyleException("Exception was thrown while processing "
-                        + file.getPath(), ex);
+                        + absolutePath, ex);
             }
             catch (Error error) {
-                if (fileName != null && cacheFile != null) {
-                    cacheFile.remove(fileName);
+                if (absolutePath != null && cacheFile != null) {
+                    cacheFile.remove(absolutePath);
                 }
 
                 // We need to catch all errors to put a reason failure (file name) in error
-                throw new Error("Error was thrown while processing " + file.getPath(), error);
+                throw new Error("Error was thrown while processing " + absolutePath, error);
             }
         }
     }
diff --git a/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java b/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
index 3598117..71f2da4 100644
--- a/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
@@ -53,6 +53,7 @@ import java.util.SortedSet;
 import java.util.TreeSet;
 
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
@@ -1715,6 +1716,7 @@ public class CheckerTest extends AbstractModuleTestSupport {
      *      for test does not require serialization
      */
     @Test
+    @Disabled
     public void testCatchErrorWithFileName() throws Exception {
         final String errorMessage = "Error was thrown while processing ";
         final Error expectedError = new IOError(new InternalError(errorMessage));
```

inthis case testExceptionWithCacheAndNoFileName failing 
```
Error cause differs from IOError
value of            : throwable.getCause()
expected instance of: java.io.IOError
but was instance of : java.lang.InternalError
with value          : java.lang.InternalError: Java Virtual Machine is broken or has run out of resources necessary for it to continue operating.

	at com.puppycrawl.tools.checkstyle.CheckerTest.testCatchErrorWithCacheWithNoFileName(CheckerTest.java:1256)

```

nrmancuso:
Changes at https://github.com/checkstyle/checkstyle/pull/13905#issuecomment-1784141080 are not what I am requesting.

romani:
But it was strange that we operated by absolute path, but reported some path. We might try to remove test of security exception. It may might be any operation on path are you exception prone, so that is why it was in try

nrmancuso:
See https://github.com/checkstyle/checkstyle/pull/13981

romani:
merged in form of https://github.com/checkstyle/checkstyle/pull/13981

PR #13923, opened 2023-10-21
Issue #13693: migrate whitespace modules to property macro

romani:
Issue #13693

several updated in javadoc that to match what actually should be.
property description is taken from setter, it is good.


rnveach:
CI is failing.

romani:
`Error: /home/runner/work/checkstyle/checkstyle/./src/xdocs/checks/whitespace/emptyforiteratorpad.xml:25,52: Insert 1 spaces. Expected 6 found 5 spaces before start element <subsection>`

romani:
fixed

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/FileTabCharacterCheck.java
```diff
@@ -51,7 +51,7 @@
  * Default value is {@code false}.
  * </li>
  * <li>
- * Property {@code fileExtensions} - Specify file type extension of files to process.
+ * Property {@code fileExtensions} - Specify the file type extension of files to process.
```
rnveach, line 54: Is this something added in this PR?

romani: It is value from setter, so it has to be the same in javadoc, test failure forcing us to do this

romani: I hope we will remove properties from javadoc one day completely.

rnveach: > value from setter, so it has to be the same in javadoc

Which we had the `XdocsJavadoc` doing before this migration. So why is it only being caught now?

rnveach: I see now. This is coming from the abstract class. `XdocsJavadoc` only looks at one file since it follows our limitations.

The `the` is coming from the field and the setter in the abstract class.

rnveach (approved):

PR #13924, opened 2023-10-22
Issue #13809: Kill mutation in PackageObjectFactory

Kevin222004:
Issue #13809: Kill mutation in PackageObjectFactory

# Explaination

if the string contains packageSeparator then the instance will be null. Probably no test case is possible.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
@@ -493,6 +493,35 @@ public class PackageObjectFactoryTest {
             .isInstanceOf(MockClass.class);
     }
 
+    /**
+    * This test case is designed to verify the behavior of the PackageObjectFactory's
+    * createModule method when it is provided with a fully qualified class name
+    * (containing a package separator).
+    * It ensures that ModuleReflectionUtil.getCheckstyleModules is not executed in this case.
+    */
+    @Test
+    public void testCreateObjectWithNameContainingPackageSeparatorWithoutSearch() throws Exception {
+        final ClassLoader classLoader = ClassLoader.getSystemClassLoader();
+        final Set<String> packages = Collections.singleton(BASE_PACKAGE);
+        final PackageObjectFactory objectFactory =
+            new PackageObjectFactory(packages, classLoader, TRY_IN_ALL_REGISTERED_PACKAGES);
+
+        try (MockedStatic<ModuleReflectionUtil> utilities =
+                     mockStatic(ModuleReflectionUtil.class)) {
+            utilities.when(() -> ModuleReflectionUtil.getCheckstyleModules(packages, classLoader))
+                    .thenThrow(new IllegalStateException("creation of objects by fully qualified"
+                            + " class names should not result in search of modules in classpath"));
+
+            final String fullyQualifiedName = MockClass.class.getName();
+            assertWithMessage("class name is not in expected format")
+                    .that(fullyQualifiedName).contains(".");
+            final Object object = objectFactory.createModule(fullyQualifiedName);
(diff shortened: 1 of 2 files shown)


romani:
CI is timeouting during test execution, that is weird, please investigate.

romani:
Rebased by GitHub

rnveach:
I thought we mentioned this area before. This is another time saving area.

If name provided has a period in it, then it is always a full class path. They will never be in `createFromStandardCheckSet` and `createObjectFromClassPath` and they will always return `null`.

I don't agree with the way the code was changed.

rdiachenko:
@Kevin222004 
Instead of changing the main logic, could you try to update this test `testCreateObjectWithNameContainingPackageSeparator` as follows:
```
    @Test
    public void testCreateObjectWithNameContainingPackageSeparator() throws Exception {
        final ClassLoader classLoader = ClassLoader.getSystemClassLoader();
        final Set<String> packages = Collections.singleton(BASE_PACKAGE);
        final PackageObjectFactory objectFactory =
            new PackageObjectFactory(packages, classLoader, TRY_IN_ALL_REGISTERED_PACKAGES);

        try (MockedStatic<ModuleReflectionUtil> utilities =
                     mockStatic(ModuleReflectionUtil.class)) {
            utilities.when(() -> ModuleReflectionUtil.getCheckstyleModules(packages, classLoader))
                    .thenThrow(new IllegalStateException("mock exception"));

            final Object object = objectFactory.createModule(MockClass.class.getName());
            assertWithMessage("Object should be an instance of MockClass")
                    .that(object)
                    .isInstanceOf(MockClass.class);
        }
    }
```

When this test is executed, `!name.contains(PACKAGE_SEPARATOR)` condition is skipped, thus, the code which throws mocked exception is also skipped. On the other hand, when pitest replaces this condition with `true`, the logic inside is executed and mocked exception is thrown which kills mutation.

romani:
>thenThrow(new IllegalStateException("mock exception"))

thenThrow(new IllegalStateException("creation of objects by fully qualified names Shou not call for search of modules in classpath"))

And we need to add comment above test method to explain why we need this test.

src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
```diff
@@ -493,6 +493,32 @@ public void testCreateObjectWithNameContainingPackageSeparator() throws Exceptio
...
+    public void testCreateObjectWithNameContainingPackageSeparator2() throws Exception {
+        final ClassLoader classLoader = ClassLoader.getSystemClassLoader();
+        final Set<String> packages = Collections.singleton(BASE_PACKAGE);
+        final PackageObjectFactory objectFactory =
+            new PackageObjectFactory(packages, classLoader, TRY_IN_ALL_REGISTERED_PACKAGES);
+
+        try (MockedStatic<ModuleReflectionUtil> utilities =
+                     mockStatic(ModuleReflectionUtil.class)) {
+            utilities.when(() -> ModuleReflectionUtil.getCheckstyleModules(packages, classLoader))
+                    .thenThrow(new IllegalStateException("mock exception"));
```
romani, line 513: thenThrow(new IllegalStateException("creation of objects by fully qualified names Shou not call for search of modules in classpath"))

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
Kevin222004: done

romani (approved):
I updated wording a bit.

ok to merge if CI is green.

rdiachenko (approved):
lgtm

PR #13176, opened 2023-06-07
Pull #13176: Fail linkcheck on internal broken site link

stoyanK7:
Extends script to fail on broken links between xdocs.

---
## EDIT: I had excluded any checks in and to `apidocs/` when I ran those scripts. Therefore some 1300 errors that were present before do not show.

I broke two links before running the experiments below.
```diff
modified: src/xdocs/config_filters.xml
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ src/xdocs/config_filters.xml:122 @
          (<code>&lt;module name=&quot;TreeWalker&quot;/&gt;</code>) and only
          applies to checks which are also defined within this module.
          To filter non-TreeWalker checks like <code>RegexpSingleline</code>,
-          a <a href="config_filters.html#SuppressWithPlainTextCommentFilter">
+          a <a href="asdasdfilters.html#SuppressWithPlainTextCommentFilter">
          SuppressWithPlainTextCommentFilter</a> or similar filter must be used.
        </p>
      </subsection>
      <subsection name="Notes" id="SuppressionCommentFilter_Notes">
        <p>
          <code>offCommentFormat</code> and <code>onCommentFormat</code> must have equal
-          <a href="https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Matcher.html#groupCount()">
+          <a href="https://docs.e.com/en/java/javase/11/docs/api/java.base/java/util/regex/Matcher.html#groupCount()">
            paren counts</a>.
        </p>
        <p>

```

### Before changes
The script greps `externalLink`, finds broken external link to oracle but not the internal one.
```sh
$ ./.ci/run-link-check-plugin.sh
...
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:52 min
[INFO] Finished at: 2023-06-07T10:37:42+02:00
[INFO] ------------------------------------------------------------------------
------------ grep of linkcheck.html--BEGIN
<td><i><a class="externalLink" href="https://docs.e.com/en/java/javase/11/docs/api/java.base/java/util/regex/Matcher.html#groupCount()">https://docs.e.com/en/java/javase/11/docs/api/java.base/java/util/regex/Matcher.html#groupCount()</a>: java.net.UnknownHostException : docs.e.com</i></td></tr></table></td></tr>
------------ grep of linkcheck.html--END
Exit code:1
```

In the report the 2 errors are shown.
image

### After changes

[...]

PR #13925, opened 2023-10-22
Issue #13885: Migrate DescendantToken to properties macros

### LINKED ISSUE #13885: Migrate DescendantToken to properties macros
there is problem to migrate DescendantToken to property macro.

```
diff --git a/config/checkstyle-non-main-files-suppressions.xml b/config/checkstyle-non-main-files-suppressions.xml
index 9c93ac93b..e0d773e8d 100644
--- a/config/checkstyle-non-main-files-suppressions.xml
+++ b/config/checkstyle-non-main-files-suppressions.xml
@@ -398,8 +398,6 @@
          files="src[\\/]xdocs[\\/]checks[\\/]misc[\\/]avoidescapedunicodecharacters.xml.template"/>
          files="src[\\/]xdocs[\\/]checks[\\/]misc[\\/]commentsindentation.xml.template"/>
-         files="src[\\/]xdocs[\\/]checks[\\/]misc[\\/]descendanttoken.xml.template"/>
          files="src[\\/]xdocs[\\/]checks[\\/]misc[\\/]finalparameters.xml.template"/>
```

```
diff --git a/src/xdocs/checks/misc/descendanttoken.xml.template b/src/xdocs/checks/misc/descendanttoken.xml.template
index bfcadc6b9..fd17ab36e 100644
--- a/src/xdocs/checks/misc/descendanttoken.xml.template
+++ b/src/xdocs/checks/misc/descendanttoken.xml.template
@@ -25,87 +25,10 @@
 
       <subsection name="Properties" id="Properties">
         <div class="wrapper">
-          <table>
-            <tr>
-              <th>name</th>
-              <th>description</th>
-              <th>type</th>
-              <th>default value</th>
-              <th>since</th>
-            </tr>
-            <tr>
-              <td>limitedTokens</td>
-              <td>Specify set of tokens with limited occurrences as descendants.</td>
-              <td>
-                subset of tokens
-                <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">TokenTypes</a>
-              </td>
-              <td><code>{}</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>maximumDepth</td>
-              <td>Specify the maximum depth for descendant counts.</td>
-              <td><a href="../../property_types.html#int">int</a></td>
-              <td><code>2147483647</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>maximumMessage</td>
-              <td>Define the violation message when the maximum count is exceeded.</td>
-              <td><a href="../../property_types.html#String">String</a></td>
-              <td><code>null</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>maximumNumber</td>
-              <td>Specify a maximum count for descendants.</td>
-              <td><a href="../../property_types.html#int">int</a></td>
-              <td><code>2147483647</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>minimumDepth</td>
-              <td>Specify the minimum depth for descendant counts.</td>
-              <td><a href="../../property_types.html#int">int</a></td>
-              <td><code>0</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>minimumMessage</td>
-              <td>Define the violation message when the minimum count is not reached.</td>
-              <td><a href="../../property_types.html#String">String</a></td>
-              <td><code>null</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>minimumNumber</td>
-              <td>Specify a minimum count for descendants.</td>
-              <td><a href="../../property_types.html#int">int</a></td>
-              <td><code>0</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>sumTokenCounts</td>
-              <td>
-                Control whether the number of tokens found should be calculated
-                from the sum of the individual token counts.
-              </td>
-              <td><a href="../../property_types.html#boolean">boolean</a></td>
-              <td><code>false</code></td>
-              <td>5.0</td>
-            </tr>
-            <tr>
-              <td>tokens</td>
-              <td>tokens to check</td>
-              <td>set of any supported
-                <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">
-                    tokens</a>
-              </td>
-              <td><code>empty</code></td>
-              <td>3.2</td>
-            </tr>
-          </table>
+          <macro name="properties">
+            <param name="modulePath"
+                   value="src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java"/>
+         </macro>
         </div>
       </subsection>
```
Failure:
```
descendanttoken.xml section 'DescendantToken' should have all the acceptable tokens
expected: set of any supported tokens
but was : TokenTypes
Expected :set of any supported tokens
Actual   :TokenTypes
<Click to see difference>


	at com.puppycrawl.tools.checkstyle.internal.XdocsPagesTest.validatePropertySectionPropertyTokens(XdocsPagesTest.java:1069)
	at com.puppycrawl.tools.checkstyle.internal.XdocsPagesTest.validatePropertySectionProperties(XdocsPagesTest.java:964)
	at com.puppycrawl.tools.checkstyle.internal.XdocsPagesTest.validatePropertySection(XdocsPagesTest.java:805)
	at com.puppycrawl.tools.checkstyle.internal.XdocsPagesTest.validateCheckSection(XdocsPagesTest.java:662)
	at com.puppycrawl.tools.checkstyle.internal.XdocsPagesTest.testAllCheckSections(XdocsPagesTest.java:579)
	at [[Reflective call: 4 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at [[Testing framework: 27 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at [[Testing framework: 9 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at [[Testing framework: 22 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at com.intellij.junit5.JUnit5IdeaTestRunner.startRunnerWithArgs(JUnit5IdeaTestRunner.java:57)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater$1.execute(IdeaTestRunner.java:38)
	at com.intellij.rt.execution.junit.TestsRepeater.repeat(TestsRepeater.java:11)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:35)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:232)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:55)
```

relentless-pursuit:
Resolves #13885

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
@@ -37,6 +37,7 @@ import org.apache.maven.doxia.module.xdoc.XdocSink;
 import org.apache.maven.doxia.sink.Sink;
 import org.codehaus.plexus.component.annotations.Component;
 
+import com.puppycrawl.tools.checkstyle.PropertyType;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailNode;
 import com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheck;
@@ -49,6 +50,12 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  */
 @Component(role = Macro.class, hint = "properties")
 public class PropertiesMacro extends AbstractMacro {
+
+    /** Represents the relative path to the property types XML. */
+    private static final String PROPERTY_TYPES_XML = "property_types.xml";
+
+    /** Represents the format string for constructing URLs with two placeholders. */
+    private static final String URL_F = "%s#%s";
 
     /** Reflects start of a code segment. */
     private static final String CODE_START = "<code>";
@@ -311,18 +318,42 @@ public class PropertiesMacro extends AbstractMacro {
         }
         else {
             final String type = SiteUtil.getType(field, propertyName, currentModuleName, instance);
-            final String relativePathToPropertyTypes =
-                    SiteUtil.getLinkToDocument(currentModuleName, "property_types.xml");
-            final String escapedType = type
-                    .replace("[", ".5B")
-                    .replace("]", ".5D");
-            final String url =
-                    String.format(Locale.ROOT, "%s#%s", relativePathToPropertyTypes, escapedType);
-            sink.link(url);
-            sink.text(type);
-            sink.link_();
+            if (PropertyType.TOKEN_ARRAY.getDescription().equals(type)) {
+                processLinkForTokenTypes(sink);
+            }
+            else {
+                final String relativePathToPropertyTypes =
+                        SiteUtil.getLinkToDocument(currentModuleName, PROPERTY_TYPES_XML);
+                final String escapedType = type
+                        .replace("[", ".5B")
+                        .replace("]", ".5D");
+
+                final String url =
+                        String.format(Locale.ROOT, URL_F, relativePathToPropertyTypes, escapedType);
+
+                sink.link(url);
+                sink.text(type);
+                sink.link_();
+            }
         }
         sink.tableCell_();
+    }
+
+    /**
+     * Writes a formatted link for "TokenTypes" to the given sink.
+     *
+     * @param sink The output target where the link is written.
+     * @throws MacroExecutionException If an error occurs during the link processing.
+     */
+    private static void processLinkForTokenTypes(Sink sink)
+            throws MacroExecutionException {
+        final String link =
+                SiteUtil.getLinkToDocument(currentModuleName, SiteUtil.PATH_TO_TOKEN_TYPES);
+
+        sink.text("subset of tokens ");
+        sink.link(link);
+        sink.text("TokenTypes");
+        sink.link_();
     }
 
     /**
--- a/src/xdocs/checks/misc/descendanttoken.xml
+++ b/src/xdocs/checks/misc/descendanttoken.xml
@@ -36,10 +36,7 @@
             <tr>
               <td>limitedTokens</td>
               <td>Specify set of tokens with limited occurrences as descendants.</td>
-              <td>
-                subset of tokens
-                <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">TokenTypes</a>
-              </td>
+              <td>subset of tokens <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">TokenTypes</a></td>
               <td><code>{}</code></td>
               <td>3.2</td>
             </tr>
@@ -87,10 +84,7 @@
             </tr>
             <tr>
               <td>sumTokenCounts</td>
-              <td>
-                Control whether the number of tokens found should be calculated
-                from the sum of the individual token counts.
-              </td>
+              <td>Control whether the number of tokens found should be calculated from the sum of the individual token counts.</td>
               <td><a href="../../property_types.html#boolean">boolean</a></td>
               <td><code>false</code></td>
               <td>5.0</td>
--- a/config/checkstyle-non-main-files-suppressions.xml
+++ b/config/checkstyle-non-main-files-suppressions.xml
@@ -348,8 +348,6 @@
   <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]checks[\\/]misc[\\/]commentsindentation.xml.template"/>
   <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]checks[\\/]misc[\\/]descendanttoken.xml.template"/>
-  <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]checks[\\/]misc[\\/]finalparameters.xml.template"/>
   <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]checks[\\/]misc[\\/]indentation.xml.template"/>
--- a/config/import-control.xml
+++ b/config/import-control.xml
@@ -97,6 +97,7 @@
     <allow class="com.puppycrawl.tools.checkstyle.utils.CheckUtil" local-only="true"/>
     <allow class="com.puppycrawl.tools.checkstyle.utils.JavadocUtil" local-only="true"/>
     <allow class="com.puppycrawl.tools.checkstyle.utils.TokenUtil" local-only="true"/>
+    <allow class="com.puppycrawl.tools.checkstyle.PropertyType"/>
   </subpackage>
 
   <subpackage name="utils">
--- a/src/xdocs/checks/misc/descendanttoken.xml.template
+++ b/src/xdocs/checks/misc/descendanttoken.xml.template
@@ -25,87 +25,10 @@
 
       <subsection name="Properties" id="Properties">
         <div class="wrapper">
-          <table>
-            <tr>
-              <th>name</th>
-              <th>description</th>
-              <th>type</th>
-              <th>default value</th>
-              <th>since</th>
-            </tr>
-            <tr>
-              <td>limitedTokens</td>
-              <td>Specify set of tokens with limited occurrences as descendants.</td>
-              <td>
-                subset of tokens
-                <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">TokenTypes</a>
-              </td>
-              <td><code>{}</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>maximumDepth</td>
-              <td>Specify the maximum depth for descendant counts.</td>
-              <td><a href="../../property_types.html#int">int</a></td>
-              <td><code>2147483647</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>maximumMessage</td>
-              <td>Define the violation message when the maximum count is exceeded.</td>
-              <td><a href="../../property_types.html#String">String</a></td>
-              <td><code>null</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>maximumNumber</td>
-              <td>Specify a maximum count for descendants.</td>
-              <td><a href="../../property_types.html#int">int</a></td>
-              <td><code>2147483647</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>minimumDepth</td>
-              <td>Specify the minimum depth for descendant counts.</td>
-              <td><a href="../../property_types.html#int">int</a></td>
-              <td><code>0</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>minimumMessage</td>
-              <td>Define the violation message when the minimum count is not reached.</td>
-              <td><a href="../../property_types.html#String">String</a></td>
-              <td><code>null</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>minimumNumber</td>
-              <td>Specify a minimum count for descendants.</td>
-              <td><a href="../../property_types.html#int">int</a></td>
-              <td><code>0</code></td>
-              <td>3.2</td>
-            </tr>
-            <tr>
-              <td>sumTokenCounts</td>
-              <td>
-                Control whether the number of tokens found should be calculated
-                from the sum of the individual token counts.
-              </td>
-              <td><a href="../../property_types.html#boolean">boolean</a></td>
-              <td><code>false</code></td>
-              <td>5.0</td>
-            </tr>
-            <tr>
-              <td>tokens</td>
-              <td>tokens to check</td>
-              <td>set of any supported
-                <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">
-                    tokens</a>
-              </td>
-              <td><code>empty</code></td>
-              <td>3.2</td>
-            </tr>
-          </table>
+          <macro name="properties">
+            <param name="modulePath"
+                   value="src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java"/>
+         </macro>
         </div>
       </subsection>
 


src/xdocs/checks/misc/descendanttoken.xml
```diff
@@ -36,10 +36,7 @@
             <tr>
               <td>limitedTokens</td>
               <td>Specify set of tokens with limited occurrences as descendants.</td>
-              <td>
-                subset of tokens
-                <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">TokenTypes</a>
-              </td>
+              <td><a href="../../property_types.html#subset of tokens TokenTypes">subset of tokens TokenTypes</a></td>
```
relentless-pursuit, line 39: trying to see if this new changes after macros migration be restored to original

relentless-pursuit:
Github, generate website

romani:
@relentless-pursuit , in case you just experimenting, you can do this on your local - https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations#how-to-generate-website-only

relentless-pursuit:
> @relentless-pursuit , in case you just experimenting, you can do this on your local - https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations#how-to-generate-website-only

thank you for the reference

relentless-pursuit:
@Romani, @nrmancuso in my opinion, we need to correct the xdocs here. On referring to the `propertyName` i.e. `limitedTokens`, it is of `String[]`, but in the xdocs we are treating it differently. I also see in XdocsJavaDocsTest.java, 
  ```
  private static final Set<String> NON_BASE_TOKEN_PROPERTIES = Collections.unmodifiableSet(
        Arrays.stream(new String[] {
            "AtclauseOrder - target",
            "DescendantToken - limitedTokens",
            "IllegalType - memberModifiers",
            "MagicNumber - constantWaiverParentToken",
            "MultipleStringLiterals - ignoreOccurrenceContext",
        }).collect(Collectors.toSet()));
```
I am not sure of the above relevance here. 

Shouldn't the xml be like this?
```
          <tr>
              <td>limitedTokens</td>
              <td>Specify set of tokens with limited occurrences as descendants.</td>
              <td><a href="../../property_types.html#String.5B.5D">String[]</a></td>
              <td><code>{}</code></td>
              <td>3.2</td>
            </tr>
```

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java [hunk 2]
```diff
@@ -156,7 +156,7 @@ public class DescendantTokenCheck extends AbstractCheck {
      */
     private boolean sumTokenCounts;
     /** Specify set of tokens with limited occurrences as descendants. */
-    @XdocsPropertyType(PropertyType.TOKEN_ARRAY)
+    @XdocsPropertyType(PropertyType.STRING_ARRAY)
     private int[] limitedTokens = CommonUtil.EMPTY_INT_ARRAY;
```
romani, line 160: Tokens are integer values internally.
Externally they look like string values, but they are actually name of int constant from TokenTypes.java

In xml all looks like a string, but modules has strickt type so xml config loader does conversion.
But conceptually modules can be configured in runtime by strick types.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java
romani: Please share problems that you have to let see how vwe can handle this problem

relentless-pursuit: While figuring out the `DescendantTokenCheck.java`, I saw the return type of property name limitedToken to be String array. However, I guess you have clarified that. 

```
 /**
     * Setter to specify set of tokens with limited occurrences as descendants.
     *
     * @param limitedTokensParam tokens to ignore.
     * @since 3.2
     */
    public void setLimitedTokens(String... limitedTokensParam) {
    ....
    }
```

relentless-pursuit: The question remains, if i go without the recent changes, the xml representation looks like below:
`<td><a href="../../property_types.html#subset of tokens TokenTypes">subset of tokens TokenTypes</a></td`

and, it is extracting `TOKEN_ARRAY` from `PropertyType.java`. 

The issue remains that for it to match with before macros migration i.e. 
```
            <tr>
              <td>limitedTokens</td>
              <td>Specify set of tokens with limited occurrences as descendants.</td>
              <td>
                subset of tokens
                <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">TokenTypes</a>
              </td>
              <td><code>{}</code></td>
              <td>3.2</td>
            </tr>
```
But, to match it to this, the code doesn't go inside the below section of `writePropertyTypeCell` method. 
```
            else {
                final List<String> configurableTokens = SiteUtil
                        .getDifference(check.getAcceptableTokens(),
                                check.getRequiredTokens())
                        .stream()
                        .map(TokenUtil::getTokenName)
                sink.text("subset of tokens");

                writeTokensList(sink, configurableTokens, SiteUtil.PATH_TO_TOKEN_TYPES);
            }
```
Because it is failing this condition: `if (SiteUtil.TOKENS.equals(propertyName))`

One approach would be to have a specific use-case of limitedTokens, and create a condition to match it with the original xdocs. I hope I didn't confuse you. Please let me know if anything wasn't clear from my side.

relentless-pursuit: What is the relevance of the below code? Any idea?
```
private static final Set<String> NON_BASE_TOKEN_PROPERTIES = Collections.unmodifiableSet(
      Arrays.stream(new String[] {
          "AtclauseOrder - target",
          "DescendantToken - limitedTokens",
          "IllegalType - memberModifiers",
          "MagicNumber - constantWaiverParentToken",
          "MultipleStringLiterals - ignoreOccurrenceContext",
      }).collect(Collectors.toSet()));
```

relentless-pursuit: > Hmm, this actually means it should be String[].
> 
> https://github.com/checkstyle/checkstyle/blob/0f9d09e1d27c01d2fbdefd6c035dc4f4a9a97cb7/src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java#L304
> 
> All others have it as in[] . So converter is required. But here look like we missed it, and it should be String[]. As type of properties is API, we can not break it.

About the converter, can you guide me further? And, what is the way forward from here?

relentless-pursuit: My learnings so far: 
```
    @XdocsPropertyType(PropertyType.STRING_ARRAY)
    private int[] limitedTokens = CommonUtil.EMPTY_INT_ARRAY;
```
it retuns int[], which possibly is the **inconsistency** highlighted. And, changing it to `String[]` might lead to **issues in backward compatibility** by breaking things. 

About, the **conversion**, i assume it is already being done in the setter method of `setLimitedTokens`. 
```
    public void setLimitedTokens(String[] limitedTokensParam) {
        limitedTokens = new int[limitedTokensParam.length];

        int maxToken = 0;
        for (int i = 0; i < limitedTokensParam.length; i++) {
            limitedTokens[i] = TokenUtil.getTokenId(limitedTokensParam[i]);
            if (limitedTokens[i] >= maxToken + 1) {
                maxToken = limitedTokens[i];
            }
        }
        counts = new int[maxToken];
    }
```
When i inspect the custom annotation of `XdocsPropertyType`, the javadoc comments highlight something
```
/**
 * The type of property used in documentation and configuration files
 * for clarification of the user-friendly type.
 * For example, some string array property may represent a set of tokens.
 */

```
So, i assume the statment `some string array property may represent a set of tokens` hinted towards the original code i.e. `@XdocsPropertyType(PropertyType.TOKENS_ARRAY)`. 

Based on maintainers further guidance, i will proceed with what the ideal xml be like. And, please correct me if i was wrong somewhere in the above learnings.

relentless-pursuit: @Romani, do you want me to replace the var-args for this?
Original:
`public void setLimitedTokens(String... limitedTokensParam) `
Possible requested change:
`public void setLimitedTokens(String[] limitedTokensParam) `

romani: Metadata 
https://github.com/checkstyle/checkstyle/blob/b31faf461e8fe543db5791462888ed26de017c93/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/DescendantTokenCheck.xml#L19-23

Already stated it is array of string type, so our direction to `String[]` is good.

Change of `String...` to `String[]` is not going to do difference. It is same.
I think nothing is required to be changed.

src/xdocs/checks/misc/descendanttoken.xml
```diff
@@ -36,10 +36,7 @@
             <tr>
               <td>limitedTokens</td>
               <td>Specify set of tokens with limited occurrences as descendants.</td>
-              <td>
-                subset of tokens
-                <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">TokenTypes</a>
```
romani, line 41: Let's put this valuable link to actual type of string values to property/setter description `<td>Specify set of tokens .....` in this case users will know that it is same.

romani (changes requested):
Lest item:

relentless-pursuit:
Github, generate website

src/xdocs/checks/misc/descendanttoken.xml
relentless-pursuit: @romani, i have made changes. please let me know if understood the requested changes whenever convenient for you.

romani:
GitHub, generate website

src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java
```diff
@@ -297,6 +303,8 @@ private void countTokens(DetailAST ast, int depth) {
 
     /**
      * Setter to specify set of tokens with limited occurrences as descendants.
+     * Refer to the link:
+     * <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">here</a>.
```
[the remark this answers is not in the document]
romani: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/5cec21e_2023125750/checks/misc/descendanttoken.html#Properties

Look at `tokens` for how to make link

src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java
```diff
@@ -43,7 +43,8 @@
  * </p>
  * <ul>
  * <li>
- * Property {@code limitedTokens} - Specify set of tokens with limited occurrences as descendants.
+ * Property {@code limitedTokens} - Specify set of
+ * <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">tokens</a>.
```
romani, line 47: Restore `with limited occurrences as descendants`

romani (changes requested):
Items

romani (approved):
Ok to merge

romani:
GitHub, generate website

romani:
There is CI failure 
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/21545/workflows/690654bd-f24d-4a13-a647-aa9b9551a8c0/jobs/416476

Please address, there might be suppression file should be extended 
```
----------- grep of linkcheck.html--BEGIN
Checking internal (checkstyle website) links only.
35a36,38
> <td><i><a href="apidocs/com/puppycrawl/tools/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html</a>: doesn't exist.</i></td></tr>
> <td><i><a href="apidocs/com/puppycrawl/tools/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">../../../../../com/puppycrawl/tools/checkstyle/checks/../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html</a>: doesn't exist.</i></td></tr>
> <td><i><a href="apidocs/com/puppycrawl/tools/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">./com/puppycrawl/tools/checkstyle/checks/../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html</a>: doesn't exist.</i></td></tr>
```

relentless-pursuit:
> There is CI failure https://app.circleci.com/pipelines/github/checkstyle/checkstyle/21545/workflows/690654bd-f24d-4a13-a647-aa9b9551a8c0/jobs/416476
> 
> Please address, there might be suppression file should be extended
> 
> ```
> ----------- grep of linkcheck.html--BEGIN
> Checking internal (checkstyle website) links only.
> 35a36,38
> > <td><i><a href="apidocs/com/puppycrawl/tools/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html</a>: doesn't exist.</i></td></tr>
> > <td><i><a href="apidocs/com/puppycrawl/tools/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">../../../../../com/puppycrawl/tools/checkstyle/checks/../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html</a>: doesn't exist.</i></td></tr>
> > <td><i><a href="apidocs/com/puppycrawl/tools/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">./com/puppycrawl/tools/checkstyle/checks/../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html</a>: doesn't exist.</i></td></tr>
> ```

done

romani:
@rnveach ,PR is fully ready

relentless-pursuit:
> @rnveach ,PR is fully ready

@rnveach, is this good to be merged?

config/linkcheck-suppressions.txt
```diff
@@ -1339,3 +1339,6 @@
 <td><i><a href="com/puppycrawl/tools/checkstyle/xpath/iterators/ReverseListIterator.html#%3Cinit%3E(java.util.Collection)">com/puppycrawl/tools/checkstyle/xpath/iterators/ReverseListIterator.html#%3Cinit%3E(java.util.Collection)</a>: doesn't exist.</i></td></tr></table></td></tr>
 <td><i><a href="com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.html#copyOfArray(T%5B%5D,int)">com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.html#copyOfArray(T%5B%5D,int)</a>: doesn't exist.</i></td></tr>
 <td><i><a href="com/puppycrawl/tools/checkstyle/site/XdocsTemplateSink.CustomPrintWriter.html#%3Cinit%3E(java.io.Writer)">com/puppycrawl/tools/checkstyle/site/XdocsTemplateSink.CustomPrintWriter.html#%3Cinit%3E(java.io.Writer)</a>: doesn't exist.</i></td></tr>
+<td><i><a href="apidocs/com/puppycrawl/tools/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html</a>: doesn't exist.</i></td></tr>
```
rnveach, line 1342: Why is this added?
Prove to me that link check is wrong and these links do exist.

src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java
```diff
@@ -155,8 +157,12 @@ public class DescendantTokenCheck extends AbstractCheck {
...
      */
     private boolean sumTokenCounts;
-    /** Specify set of tokens with limited occurrences as descendants. */
-    @XdocsPropertyType(PropertyType.TOKEN_ARRAY)
+    /**
+     * Specify set of
+     * <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">tokens</a>
+     * with limited occurrences as descendants.
+     */
+    @XdocsPropertyType(PropertyType.STRING_ARRAY)
```
rnveach, line 165: I am still reading prior discussion https://github.com/checkstyle/checkstyle/pull/13925#discussion_r1368922550 , but this change doesn't seem valid.
Property is an array of tokens.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java
rnveach: > Because it is failing this condition: if (SiteUtil.TOKENS.equals(propertyName))
> One approach would be to have a specific use-case of limitedTokens, and create a condition to match it with the original xdocs. 

We should not have to, and really avoid, creating a custom override list of types for specific checks and properties. Everything is documented in class that this is a token list, so we should be using that.

It looks to me that code needs to be re-written to take actual type from `XdocsPropertyType`, which I am sure I mentioned in some other issue and then just have somewhat special handling of token lists as there is no pre-defined acceptable/required list like with natural `tokens` property.

config/linkcheck-suppressions.txt
relentless-pursuit: > Why is this added? Prove to me that link check is wrong and these links do exist.

to overcome the below CI failure
```
----------- grep of linkcheck.html--BEGIN
Checking internal (checkstyle website) links only.
35a36,38
> <td><i><a href="apidocs/com/puppycrawl/tools/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html</a>: doesn't exist.</i></td></tr>
> <td><i><a href="apidocs/com/puppycrawl/tools/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">../../../../../com/puppycrawl/tools/checkstyle/checks/../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html</a>: doesn't exist.</i></td></tr>
> <td><i><a href="apidocs/com/puppycrawl/tools/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">./com/puppycrawl/tools/checkstyle/checks/../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html</a>: doesn't exist.</i></td></tr>
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java
relentless-pursuit: That can be may be substantiated by @Romani. But, i will just add some more context to it. If we refer `DescendantTokenCheck.xml`, it says limitedTokens is of `String[]`.
```
            <property default-value=""
                      name="limitedTokens"
                      type="java.lang.String[]"
                      validation-type="tokenTypesSet">
               <description>Specify set of tokens with limited occurrences as descendants.</description>
            </property>
```

config/linkcheck-suppressions.txt
rnveach: So then what is your reasoning that you think CI was NOT telling you something is wrong with your changes, and instead is a bug with link check?

src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java
rnveach: Then it sounds like a bug in the metadata as the property is a representation of tokens, and no 3rd party utility will know this since we say its just a simple string array.

rnveach: Even if we are suppose to take the type from `validation-type`, which is still slightly confusing, especially since we have no documentation of the metadata and what any of it really means. It doesn't change that this is a token array, and how will metadata know to use the `validation-type` anyways.

This doesn't seem to be making much sense to me.

config/linkcheck-suppressions.txt
relentless-pursuit: I guess the CI wasn't failing before my last push. It started failing on making these changes: `<td>Specify set of <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">tokens</a> with limited occurrences as descendants.</td>`. But, fair point. When we click on the tokens, in the generated website, it redirects to the desired section. But, i do see the message in CI clearly. <a href="apidocs/com/puppycrawl/tools/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html">../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html</a>. Possibly it is doing some combination and giving links not valid.

src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java
rnveach: https://github.com/checkstyle/checkstyle/blob/8896be8cd4e90c59fdba8e048000b0f3472245c9/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsJavaDocsTest.java#L374

https://github.com/checkstyle/checkstyle/blob/8896be8cd4e90c59fdba8e048000b0f3472245c9/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsJavaDocsTest.java#L107-L113

It looks like there is another area that is just a custom hack list which should be using `XdocsPropertyType`.

All these partial GSoC fixes are just causing a mess with the code.

relentless-pursuit: Sure, i am leaning with the go. Let me know how to proceed from here.

config/linkcheck-suppressions.txt
rnveach: I recommend the first thing you do is figure out which of your changes are causing the failures. This will require you to try reverting stuff and running link check on your local. When you figure out which lines are the issue, tell us those lines, then generate that into whatever website is needed and test out the link and check it actually works.

If you need help let us know, but I can tell you, I can already tell you from what I saw, CI was not wrong about the failure.

src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java
rnveach: How to proceed is at https://github.com/checkstyle/checkstyle/pull/13925#discussion_r1375306101 .

After you do that, I recommend to create an issue for https://github.com/checkstyle/checkstyle/pull/13925#discussion_r1375308387 and work on it also since it will be similar areas to the 2.

romani: This is very confusing update , I agree, and kept it unresolved to share through process.

Confusion is that we use this annotation on fields but it actually should represent type of setters. We naively thought that all our properties has fully aligned setter and field. But here are exceptions. 
To avoid doing breaking change on setter type we should put annotation with different type on class field.

I am ok to create breaking  compatibility issue  and resolve it separately. But for now it is like what we have in this PR.
Do you want us to create this issue ?

config/linkcheck-suppressions.txt
romani: Hm,  generated website and all links are working. That is why I thought it is some limitation of our plugin execution.

src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java
rnveach: > Confusion is that we use this annotation on fields but it actually should represent type of setters
> We naively thought that all our properties has fully aligned setter and field

There is too much baggage here, but basically this has been known since the beginning. `XdocsPagesTest` handles identifying types as we have created since the beginning. I may be biased since I created it and worked it out, but it has stood the test of time for this long.

You can't go by setter alone, and you can't go by field alone. There won't always be a field, and due to limitations placed by ourselves on allowing extension of the bean utility, setters are purposely restricted to simple types. We have shown both of those things in this PR. `aliasList` has no field, and this class' setter is a simple string array for a token array type.

> But here are exceptions

To me there are very few exceptions. I pointed out now how a custom override list can be maintained through our annotations.

If you want to see how types are identified, look at https://github.com/checkstyle/checkstyle/blob/bd3a9bb9ec9cf804f628171c2e7557b0afbcfd93/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1369 and the surrounding code.

The field is the primary authority. If it doesn't exist, we go against the setter. See above on why we can't solely rely on one or the other.

> To avoid doing breaking change on setter type we should put annotation with different type on class field.

I am not sure I understand, but I am not sure it really matters. Annotation aren't breaking compatibility. Changing setter type won't really break compatibility as long as values in configs can remain inputted the same.

> Do you want us to create this issue ?

You can go about this as you want, assuming we still need a consensus, but we need to start planning better and providing actual fixes and less overrides.

If you want my opinion, we should start fixing issues where it has been pointed out things are wrong as it will just come up and block future migration work.

I don't see the change in this PR being the way.

config/linkcheck-suppressions.txt
rnveach: I believe I do, but I don't know for sure as I am not looking into it. If you want to see how I would tell if I am right or wrong, my instructions are at https://github.com/checkstyle/checkstyle/pull/13925#discussion_r1375309638 .

We shouldn't just assume CI is wrong. It seems like that is what is being done and it doesn't seem like anyone has looked into it to confirm that X link in Y file is being reported as wrong and it is shown to be wrong. If this is a bug with the utility, we should report it.

src/main/java/com/puppycrawl/tools/checkstyle/checks/DescendantTokenCheck.java
relentless-pursuit: @rnveach, i have reverted back to keep changes for token_array. though, my CI is failing, pointing 
`<td><i><a href="property_types.html#subset_of_tokens_TokenTypes">../../property_types.html#subset_of_tokens_TokenTypes</a>: doesn't exist.</i></td></tr>`. 

Even if i didn't add any logic in this recent push, simply relying on some other fix that was merged which also had a failed macros. I am currently looking into ti. But, your review would be insightful.

romani:
@relentless-pursuit , 

>`<td><a href="../../property_types.html#subset of tokens TokenTypes">subset of tokens TokenTypes</a></td>`

I am not sure why reverted back to this, this is also a form that @rnveach , does not like.

rnveach:
> "../../property_types.html#subset of tokens TokenTypes"

This fails linkcheck.

relentless-pursuit:
> @relentless-pursuit ,
> 
> > `<td><a href="../../property_types.html#subset of tokens TokenTypes">subset of tokens TokenTypes</a></td>`
> 
> I am not sure why reverted back to this, this is also a form that @rnveach , does not like.
> 
> I am reading attentively his comments, to understand how to explain what we did.

I haven't really added any logic. I just tried with macros migration. And, it resulted in the following xml. If i compare with original, it considered the entire `subset of token TokenTypes` as a link. Now, it just takes TokenTypes as a link. Should the effort be to be in sync with original?

relentless-pursuit:
> @relentless-pursuit ,
> 
> > `<td><a href="../../property_types.html#subset of tokens TokenTypes">subset of tokens TokenTypes</a></td>`
> 
> I am not sure why reverted back to this, this is also a form that @rnveach , does not like.
> 
> I am reading attentively his comments, to understand how to explain what we did.

Also, a bit of clarity on how the ideal xml should be like be helpful in closing this PR>

romani:
Ideal is xml and link should be same as before by content, spacing/formatting differences are ok.

src/xdocs/checks/misc/descendanttoken.xml
romani: I am ok, with such update.
I can move conversation with @rnveach  to separate issue to discuss it separately and not blocking this migration.

romani:
@rnveach , please review.

rnveach:
CI is failing.

romani:
@rnveach , do you in generally agree with this approach?

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -311,20 +317,45 @@ else if (SiteUtil.JAVADOC_TOKENS.equals(propertyName)) {
...
-                    SiteUtil.getLinkToDocument(currentModuleName, "property_types.xml");
-            final String escapedType = type
-                    .replace("[", ".5B")
-                    .replace("]", ".5D");
-            final String url =
-                    String.format(Locale.ROOT, "%s#%s", relativePathToPropertyTypes, escapedType);
-            sink.link(url);
-            sink.text(type);
-            sink.link_();
+            if ("subset of tokens TokenTypes".equals(type)) {
```
rnveach, line 320: Truthfully, I question this kind of matching.

We are losing type safety by not matching on a type like `XdocsPagesTest` did, by requiring a match on a custom string. This may be a limitation of how we are coding this macro, but custom strings are usually not a good idea. It is easy to misspell or typo something and go unnoticed.

I also question where these strings are coming from, and why don't we match on defined public string from there, similar to what we did with `PROPERTY_TYPES_XML`. Something like `if (type.equals(TOK_STR_1 + TOK_STR_2))`.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -311,20 +317,45 @@ else if (SiteUtil.JAVADOC_TOKENS.equals(propertyName)) {
...
+                processLinkForTokenTypes(sink);
+            }
+            else {
+                final String relativePathToPropertyTypes =
+                        SiteUtil.getLinkToDocument(currentModuleName, PROPERTY_TYPES_XML);
+                final String escapedType = type
+                        .replace("[", ".5B")
+                        .replace("]", ".5D");
```
rnveach, line 328: I remember there was an issue where we should be using some URL building util instead of manually doing this.

config/linkcheck-suppressions.txt
rnveach: @romani @relentless-pursuit Since this code is now being removed from the PR, the invalid links were coming from the Javadoc. Not the XDocs.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
relentless-pursuit: > Truthfully, I question this kind of matching.
> 
> We are losing type safety by not matching on a type like `XdocsPagesTest` did, by requiring a match on a custom string. This may be a limitation of how we are coding this macro, but custom strings are usually not a good idea. It is easy to misspell or typo something and go unnoticed.
> 
> I also question where these strings are coming from, and why don't we match on defined public string from there, similar to what we did with `PROPERTY_TYPES_XML`. Something like `if (type.equals(TOK_STR_1 + TOK_STR_2))`.

I agree with type-safety issue. I personally didn't like the string literals but i assume since it was more of a hack, and a sustainable solution is waiting, i didn't create constants. However, in my coming push, i have refactored the code as such:
 `if (type.equals(SUBSET_TOKENS + TOKENS_PROPERTY)) {`
 
 I created a constant for the same and used an existing one. Please let me know your feedback.

relentless-pursuit: > PROPERTY_TYPES_XM

Or, do you want me to extract 'subset of tokens' from somewhere inside the code?

relentless-pursuit: from `TOKEN_ARRAY("subset of tokens TokenTypes");`

being used here:
```
  @XdocsPropertyType(PropertyType.TOKEN_ARRAY)
    private int[] limitedTokens = CommonUtil.EMPTY_INT_ARRAY;
```

We see the same in xdocs. 

And, all the below are using `TOKEN_ARRAY`
```
private static final Set<String> NON_BASE_TOKEN_PROPERTIES = Collections.unmodifiableSet(
      Arrays.stream(new String[] {
          "AtclauseOrder - target",
          "DescendantToken - limitedTokens",
          "IllegalType - memberModifiers",
          "MagicNumber - constantWaiverParentToken",
          "MultipleStringLiterals - ignoreOccurrenceContext",
      }).collect(Collectors.toSet()));
```

relentless-pursuit: We see the same issue in migration of all below files. Maybe, this hack for this PR will help migration of the above checks mentioend.

relentless-pursuit: And, in my code flow, it directly enters else block with value as `"subset of tokens TokenTypes"`
  ```
private static void writePropertyTypeCell(Sink sink, String propertyName,
                                              Field field, Object instance)
            throws MacroExecutionException {
        sink.rawText(INDENT_LEVEL_14);
        sink.tableCell();

        if (SiteUtil.TOKENS.equals(propertyName)) {
            final AbstractCheck check = (AbstractCheck) instance;
            if (check.getRequiredTokens().length == 0
                    && Arrays.equals(check.getAcceptableTokens(), TokenUtil.getAllTokenIds())) {
                sink.text("set of any supported");
                writeLink(sink);
            }
            else {
                final List<String> configurableTokens = SiteUtil
                        .getDifference(check.getAcceptableTokens(),
                                check.getRequiredTokens())
                        .stream()
                        .map(TokenUtil::getTokenName)
                sink.text(SUBSET_TOKENS);

                writeTokensList(sink, configurableTokens, SiteUtil.PATH_TO_TOKEN_TYPES);
            }
        }
        else if (SiteUtil.JAVADOC_TOKENS.equals(propertyName)) {
            final AbstractJavadocCheck check = (AbstractJavadocCheck) instance;
            final List<String> configurableTokens = SiteUtil
                    .getDifference(check.getAcceptableJavadocTokens(),
                            check.getRequiredJavadocTokens())
                    .stream()
                    .map(JavadocUtil::getTokenName)
            sink.text("subset of javadoc tokens");
            writeTokensList(sink, configurableTokens, SiteUtil.PATH_TO_JAVADOC_TOKEN_TYPES);
        }
        else {
            final String type = SiteUtil.getType(field, propertyName, currentModuleName, instance);
            if ((SUBSET_TOKENS + " " + TOKEN_TYPES_PROPERTY).equals(type)) {
                processLinkForTokenTypes(sink);
            }
            else {
                final String relativePathToPropertyTypes =
                        SiteUtil.getLinkToDocument(currentModuleName, PROPERTY_TYPES_XML);
                final String escapedType = type
                        .replace("[", ".5B")
                        .replace("]", ".5D");
                final String url =
                        String.format(Locale.ROOT, URL_F, relativePathToPropertyTypes, escapedType);
                sink.link(url);
                sink.text(type);
                sink.link_();
            }
        }
        sink.tableCell_();
    }
```
    
   And, if we ignore the  `if ((SUBSET_TOKENS + " " + TOKEN_TYPES_PROPERTY).equals(type)) {` added by me, it goes on to write the xdocs with link followed by type which is nothing but value held by `TOKENS_ARRAY`. So, if i don't add any hack and just proceed with normal migration, you see something like this:
`<td><a href="../../property_types.html#subset of tokens TokenTypes">subset of tokens TokenTypes</a></td>`

Comment (relentless-pursuit) [reply src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java]: And, that leads to linkcheck error in CI. So, either way, it is broken.

Comment (relentless-pursuit) [reply src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java]: And, if migration is a priority, we should proceed with this PR. Otherwise, in whatever little time, i have understood the code, it needs to be re-written to accomodate all edge cases we have been encountering with properties migration. So, if @rnveach, if you agree, we can proceed with the PR and i can start working with your guiding light on how to get rid of hacks so far.

Comment (rnveach) [reply src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java]:
```suggestion
            if (PropertyType.TOKEN_ARRAY.getDescription().equals(type)) {
```

This resolves the type safety and usage of raw text concern.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
rnveach: If we don't do it in this PR, it needs to be added to the other PR.

config/linkcheck-suppressions.txt
romani: I remember that there was some problem in apidoc 
https://github.com/checkstyle/checkstyle/pull/13176

Let's keep growing such suppression for now.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -311,20 +320,47 @@ else if (SiteUtil.JAVADOC_TOKENS.equals(propertyName)) {
...
+            }
+            else {
+                final String relativePathToPropertyTypes =
+                        SiteUtil.getLinkToDocument(currentModuleName, PROPERTY_TYPES_XML);
+
+                final String encodedType = URLEncoder.encode(type, StandardCharsets.UTF_8);
+
+                final String encTypeF = encodedType
+                        .replace("%5B", ".5B")
+                        .replace("%5D", ".5D");
```
rnveach, line 334: I was under the impression these were not needed when we use the URL encoder. Is something wrong?

relentless-pursuit: If i just use the encoder, it alters the url for the following files in this manner
```
src/xdocs/checks/coding/illegalcatch.xml
src/xdocs/checks/coding/illegalinstantiation.xml 
src/xdocs/checks/coding/illegalthrows.xml 
src/xdocs/checks/whitespace/filetabcharacter.xml
```
```
-              <td><a href="../../property_types.html#String.5B.5D">String[]</a></td>
+              <td><a href="../../property_types.html#String%5B%5D">String[]</a></td>
```

rnveach: ok, that is interesting. Doxia is basically creating an invalid link drop as it is using `.` instead of `%`. This is why the URL encoder uses `%`.

@romani I question, should we still be using URL encoder?

rnveach (approved):

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
romani: @relentless-pursuit , can we avoid `encodedType = URLEncoder.encode`?

relentless-pursuit: Yes, i will restore it back to original.

relentless-pursuit: > @relentless-pursuit , can we avoid `encodedType = URLEncoder.encode`?

done

romani (approved):
ok to merge if CI pass

PR #13913, opened 2023-10-19
Issue #13809: Kill mutation in Checker-3

Kevin222004:
Issue #13809: Kill mutation in Checker-3

--------

Mutation 

https://github.com/checkstyle/checkstyle/blob/8c8bc0c6102823206ca3a562b01b26882fc242fa/config/pitest-suppressions/pitest-common-suppressions.xml#L21-L28

PR #13928, opened 2023-10-22
Issue #13809: Kill mutation in Tree-walker

Kevin222004:
Issue #13809: Kill mutation in Tree-walker

### DIFF
--- a/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
+++ b/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
@@ -93,15 +93,6 @@
   <mutation unstable="false">
     <sourceFile>TreeWalker.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.TreeWalker</mutatedClass>
-    <mutatedMethod>getExternalResourceLocations</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to java/util/stream/Stream::map with receiver</description>
-    <lineContent>.map(ExternalResourceHolder.class::cast)</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>TreeWalker.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.TreeWalker</mutatedClass>
     <mutatedMethod>lambda$createNewCheckSortedSet$3</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.EmptyObjectReturnValsMutator</mutator>
     <description>replaced return value with &quot;&quot; for com/puppycrawl/tools/checkstyle/TreeWalker::lambda$createNewCheckSortedSet$3</description>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
@@ -388,8 +388,10 @@ public final class TreeWalker extends AbstractFileSetCheck implements ExternalRe
         return Stream.concat(filters.stream(),
                 Stream.concat(ordinaryChecks.stream(), commentChecks.stream()))
             .filter(ExternalResourceHolder.class::isInstance)
-            .map(ExternalResourceHolder.class::cast)
-            .flatMap(resource -> resource.getExternalResourceLocations().stream())
+            .flatMap(resource -> {
+                return ((ExternalResourceHolder) resource)
+                        .getExternalResourceLocations().stream();
+            })
             .collect(Collectors.toSet());
     }
 


src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
```diff
@@ -388,8 +388,10 @@ public Set<String> getExternalResourceLocations() {
         return Stream.concat(filters.stream(),
                 Stream.concat(ordinaryChecks.stream(), commentChecks.stream()))
             .filter(ExternalResourceHolder.class::isInstance)
-            .map(ExternalResourceHolder.class::cast)
```
nrmancuso, line 391: @Kevin222004 I am curious; does this suffer the same surviving mutation?

```
            .map(r -> (ExternalResourceHolder) r)
```

config/pitest-suppressions/pitest-tree-walker-suppressions.xml
```diff
@@ -90,15 +90,6 @@
     <lineContent>Comparator.nullsLast(Comparator.naturalOrder()))</lineContent>
   </mutation>
 
-  <mutation unstable="false">
-    <sourceFile>TreeWalker.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.TreeWalker</mutatedClass>
-    <mutatedMethod>getExternalResourceLocations</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to java/util/stream/Stream::map with receiver</description>
```
nrmancuso, line 98: @Kevin222004 doesn't this effectively mean that we completely remove the call to `map` from this chain of method calls?

romani: there is compilation error:
```
[INFO] BUILD FAILURE
[INFO] Total time:  8.178 s
[INFO] Finished at: 2023-10-22T06:44:59-07:00
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project checkstyle: Compilation failure: Compilation failure: 
[ERROR] /home/rivanov/java/github/romani/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java:[391,42] cannot find symbol
[ERROR]   symbol:   method getExternalResourceLocations()
[ERROR]   location: variable resource of type java.lang.Object
[ERROR] /home/rivanov/java/github/romani/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java:[392,21] incompatible types: inference variable T has incompatible bounds
[ERROR]     equality constraints: java.lang.String
[ERROR]     lower bounds: java.lang.Object
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException


$ git diff
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java b/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
index ae9a0e6..9c3d5c8 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
@@ -388,7 +388,6 @@ public final class TreeWalker extends AbstractFileSetCheck implements ExternalRe
         return Stream.concat(filters.stream(),
                 Stream.concat(ordinaryChecks.stream(), commentChecks.stream()))
             .filter(ExternalResourceHolder.class::isInstance)
-            .map(ExternalResourceHolder.class::cast)
             .flatMap(resource -> resource.getExternalResourceLocations().stream())
             .collect(Collectors.toSet());
     }

```

I think it is reasonable to contact pitest.

@Vyom-Yadav , did you experience the same ?

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
Kevin222004: @nrmancuso Yes, it will survive the same.

https://kevin222004.github.io/PitestReport/com.puppycrawl.tools.checkstyle/TreeWalker.java.html

config/pitest-suppressions/pitest-tree-walker-suppressions.xml
romani: I think mutations are by design create code that is not compile able.
Compiler smart and can block nonsense code writing, but jvm can execute  any bitecode.
So we can not create issue on pitest, it is by design.

@nrmancuso .

romani (approved):
Ok to merge

config/pitest-suppressions/pitest-tree-walker-suppressions.xml
rnveach: Pitest is maniupulation of the byte code. Compilation is turning source code to byte code. There is no "compilable" code as byte code pitest modifies is already compiled from source.

In this instance, it probably means you did not successfully recreate what pitest is doing by modifying the source instead. It also could be that pitest is not implementing some safeties in its mutations that the compiler checks for and adds. You will need to look at the original bytecode and the modified bytecode to see what exactly pitest is doing to even know if it is a bug in pitest or a misunderstanding by the user. We had to do this quite a bit for this project, which hopefully we aren't lowering our standards.

rnveach: You may want to submit this issue to pitest as it seems it will always happily ignore type casts while still being able to call methods that require the cast in the first place. This seems like pitest is bypassing what the compiler checks and the JVM assumes the compiler already did all the proper checks in the first place to not care at run time.

romani: As few people are unsure, let's create issue in pitest to get official position on what is correct.

But in same time let's proceed with PR acceptance.

rdiachenko (approved):
lgtm

config/pitest-suppressions/pitest-tree-walker-suppressions.xml
nrmancuso: Fact: we still don’t know what the actual mutation is. That is enough to block this PR.

rnveach:
https://github.com/checkstyle/checkstyle/pull/13928#discussion_r1369993041

config/pitest-suppressions/pitest-tree-walker-suppressions.xml
romani: It is ok that pitest mutate code to non compilable code, mutation already imply that code becomes weird and error prone. Java compiler catch most basic and obvious nonsense code.

Vyom-Yadav:
I don't think it is straight away removing call to `map`, I pointed out exactly the same type of mutation in https://github.com/checkstyle/checkstyle/pull/13913#pullrequestreview-1691211278.

That comment contains the bytecode generated by PIT after inserting the mutation. Decompiling does not guarantee compilable code, so I wasn't sure whether it was PIT or decompilers limitation.

rnveach:
https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports#decompiler-usage-to-see-bytecode

Its misleading that it is a decompiler. You need to look at the mutated bytecode, and then create an understanding from that what it is doing. I am sure it is possible to have bytecode that will not translate into any standard Java, that doesn't mean the bytecode isn't valid and executable.

Vyom-Yadav:
Here is the mutated bytecode for this particular mutation: 
[mutant1.class.gz](https://github.com/checkstyle/checkstyle/files/13169613/mutant1.class.gz)
 
Continuing my comment at https://github.com/checkstyle/checkstyle/pull/13928#issuecomment-1779865155, even the **non-mutated** code, decompiled by fernflower isn't compilable. 

Here is what fernflower gave me:
```java
public Set<String> getExternalResourceLocations() {                                                                                              
    Stream var10000 = Stream.concat(this.filters.stream(), Stream.concat(this.ordinaryChecks.stream(), this.commentChecks.stream()));            
    Objects.requireNonNull(ExternalResourceHolder.class);                                                                                        
    var10000 = var10000.filter(ExternalResourceHolder.class::isInstance);                                                                        
    Objects.requireNonNull(ExternalResourceHolder.class);                                                                                        
    return (Set)var10000.map(ExternalResourceHolder.class::cast).flatMap((resource) -> {                                                         
        return resource.getExternalResourceLocations().stream();                                                                                 
    }).collect(Collectors.toSet());                                                                                                              
}                                                                                                                                                
```
```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/vyom/IdeaProjects/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java:[394,28] cannot find symbol
  symbol:   method getExternalResourceLocations()
  location: variable resource of type java.lang.Object
[INFO] 1 error
[INFO] -------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] Total time:  01:27 min
[INFO] Finished at: 2023-10-26T00:34:48+05:30
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project checkstyle: Compilation failure
[ERROR] /home/vyom/IdeaProjects/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java:[394,28] cannot find symbol
[ERROR]   symbol:   method getExternalResourceLocations()
[ERROR]   location: variable resource of type java.lang.Object
[ERROR] 
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException

vyom at fedora in ~/IdeaProjects/checkstyle (fixLinkks●) 
$ git diff          
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java b/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
index ae9a0e646..31175d9c3 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
@@ -27,6 +27,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
@@ -385,12 +386,13 @@ public final class TreeWalker extends AbstractFileSetCheck implements ExternalRe
 
     @Override
     public Set<String> getExternalResourceLocations() {
-        return Stream.concat(filters.stream(),
-                Stream.concat(ordinaryChecks.stream(), commentChecks.stream()))
-            .filter(ExternalResourceHolder.class::isInstance)
-            .map(ExternalResourceHolder.class::cast)
-            .flatMap(resource -> resource.getExternalResourceLocations().stream())
-            .collect(Collectors.toSet());
+        Stream var10000 = Stream.concat(this.filters.stream(), Stream.concat(this.ordinaryChecks.stream(), this.commentChecks.stream()));
+        Objects.requireNonNull(ExternalResourceHolder.class);
+        var10000 = var10000.filter(ExternalResourceHolder.class::isInstance);
+        Objects.requireNonNull(ExternalResourceHolder.class);
+        return (Set)var10000.map(ExternalResourceHolder.class::cast).flatMap((resource) -> {
+            return resource.getExternalResourceLocations().stream();
+        }).collect(Collectors.toSet());
     }
 
     /**
```

So I think we should at least bring this to pitest maintainers attention.

rnveach:
Provide us the difference between the mutated and non-mutated code from JavaP.

Vyom-Yadav:
> Provide us the difference from JavaP between the mutated and non-mutated byte code.

https://www.diffchecker.com/LhJhXQVV/

rnveach:
Thank you.

This makes sense. This is what [Naked Receiver](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports#experimental-naked-receiver) does.

The only difference is pitest can modify the bytecode and can bypass the type check. This is what I surmised at https://github.com/checkstyle/checkstyle/pull/13928#discussion_r1369478579 .

I still think we should report this case to pitest because according to the bytecode it creates an equivalent mutation when it doesn't when compared to the source compiler. We should avoid these cases and this PR in hopes that pitest acknowledges and provides a fix.

romani:
Type checking is compiler job, jvm bite code executes just commands, and if some casting problems happens ( if compiler missed it) it will be run time exception.

rnveach:
An exception would kill the mutation. Since the mutation survives, I assume there is no runtime exception happening.

romani:
Yes, because code generate proper types of collection elements, and only single mutation happening at the time.

Vyom-Yadav:
We don't hit a runtime exception as we are already filtering based on whether it is an instance or not. The compiler complains for this but jvm can execute those instructions.

A very particular case, pitest may not be able to handle such types of cases.

romani:
@Vyom-Yadav, how you get mutated bytecode ?

rnveach:
@romani https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports#decompiler-usage-to-see-bytecode

Vyom-Yadav:
I don't think pitest can solve this. If anything, if a method was invoked at A then that method invokes B and so on...

Somewhere down in the tree there is a cast. Pitest would possibly have to ignore all the methods in the tree to get rid of this mutation. IT MAY NOT BE a direct child parent relationship with these type of method calls.

Our best option rn is to work around like we are doing rn. This mutation only survives because this is only required by compiler. Maybe a pitest limitation. 

But let's create an issue to discuss this with the maintainers.

romani:
@Kevin222004 , please create issue on pitest.

Vyom-Yadav:
> @Vyom-Yadav, how you get mutated bytecode ?

I have made some modifications to the test suite of pitest, so when a class is mutated, I save the mutated bytecode. See https://github.com/hcoles/pitest/compare/master...Vyom-Yadav:pitest:minorMod

rnveach:
There is also https://github.com/hcoles/pitest/issues/391#issuecomment-326663476
https://github.com/pitest/export-plugin

Vyom-Yadav:
Nice find, https://github.com/hcoles/pitest/issues/432, it was moved to the main repo.

Vyom-Yadav (approved):
LGTM! Mutant has been discussed. This is the best way to proceed IMO.

romani:
I stored both link to https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports

and clear answer from author that pitest does not try code to be mathching to compiler expectations.

nrmancuso:
I am prepared to approve this PR once we have issue opened in pitest repo.

romani:
issue is opened https://github.com/hcoles/pitest/issues/1270

nrmancuso (approved):

PR #13878, opened 2023-10-13
Issue #13853: migrate FinalLocalVariable xdoc to use property macro

AayushSaini101:
Related: https://github.com/checkstyle/checkstyle/issues/13853

Issue #13853, opened 2023-10-09
migrate FinalLocalVariable xdoc to use property macro


romani:
following diff should be result in chnages of ordering in corresponding xdoc file:
```
diff --git a/src/xdocs/checks/coding/finallocalvariable.xml.template b/src/xdocs/checks/coding/finallocalvariable.xml.template
index 4fc6691..880d27f 100644
--- a/src/xdocs/checks/coding/finallocalvariable.xml.template
+++ b/src/xdocs/checks/coding/finallocalvariable.xml.template
@@ -25,48 +25,10 @@
 
       <subsection name="Properties" id="Properties">
         <div class="wrapper">
-          <table>
-            <tr class="header">
-              <th>name</th>
-              <th>description</th>
-              <th>type</th>
-              <th>default value</th>
-              <th>since</th>
-            </tr>
-            <tr>
-              <td>validateEnhancedForLoopVariable</td>
-              <td>
-                Control whether to check
-                <a href="https://docs.oracle.com/javase/specs/jls/se11/html/jls-14.html#jls-14.14.2">
-                enhanced for-loop</a> variable.
-              </td>
-              <td>
-                <a href="../../property_types.html#boolean">boolean</a>
-              </td>
-              <td>
-                <code>false</code>
-              </td>
-              <td>6.5</td>
-            </tr>
-            <tr>
-              <td>tokens</td>
-              <td>tokens to check</td>
-              <td>
-                subset of tokens
-                <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#VARIABLE_DEF">
-                VARIABLE_DEF</a>
-                , <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#PARAMETER_DEF">
-                PARAMETER_DEF</a>
-                .
-              </td>
-              <td>
-                <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#VARIABLE_DEF">
-                  VARIABLE_DEF</a>
-                  .
-              </td>
-              <td>3.2</td>
-            </tr>
-          </table>
+          <macro name="properties">
+            <param name="modulePath"
+                   value="src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java"/>
+          </macro>
         </div>
       </subsection>
 

```

AayushSaini101:
[...]

PR #13931, opened 2023-10-22
Issue #13853: migrate FinalLocalVariable xdoc to use property macro

AayushSaini101:
Related:  https://github.com/checkstyle/checkstyle/issues/13853

Old PR: https://github.com/checkstyle/checkstyle/pull/13878

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
@@ -22,6 +22,7 @@ package com.puppycrawl.tools.checkstyle.site;
 import java.io.File;
 import java.lang.reflect.Field;
 import java.util.Arrays;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -67,6 +68,12 @@ public class PropertiesMacro extends AbstractMacro {
     private static final String INDENT_LEVEL_18 = SiteUtil.getNewlineAndIndentSpaces(18);
     /** A newline with 20 spaces of indentation. */
     private static final String INDENT_LEVEL_20 = SiteUtil.getNewlineAndIndentSpaces(20);
+
+    /**
+     * This property is used to change the existing properties for javadoc.
+     * Tokens always present at the end of all properties.
+    */
+    private static final String TOKENS_PROPERTY = SiteUtil.TOKENS;
 
     /** The name of the current module being processed. */
     private static String currentModuleName = "";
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -68,7 +69,10 @@ public class PropertiesMacro extends AbstractMacro {
     /** A newline with 20 spaces of indentation. */
     private static final String INDENT_LEVEL_20 = SiteUtil.getNewlineAndIndentSpaces(20);
 
-    /** The name of the current module being processed. */
+    /** The property being process with all tokens. */
+    private static final String TOKENS_PROPERTY = SiteUtil.TOKENS;
+
+    /** The file of the current module being processed. */
```
nrmancuso, line 75: Why is this changed?

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -165,13 +169,34 @@ private static void writeTablePropertiesRows(Sink sink)
         final Map<String, DetailNode> propertiesJavadocs = SiteUtil
                 .getPropertiesJavadocs(properties, currentModuleName, currentModuleFile);
 
-        for (String property : properties) {
+        final List<String> orderProperties = orderProperties(properties);
```
nrmancuso, line 172:
```suggestion
        final List<String> orderedProperties = orderProperties(properties);
```

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -165,13 +169,34 @@ private static void writeTablePropertiesRows(Sink sink)
...
+    private static List<String> orderProperties(Set<String> properties) {
+
+        final List<String> validProperties = new LinkedList<>(properties);
+
+        if (validProperties.remove(TOKENS_PROPERTY)) {
+            validProperties.add(TOKENS_PROPERTY);
+        }
+
+        return validProperties;
+
```
nrmancuso, line 197: Please return immutable list here, as suggested in other PR:
```suggestion
      if (properties.remove(TOKENS_PROPERTY)) {
            properties.add(TOKENS_PROPERTY);
        }
        return List.copyOf(properties);
```

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
AayushSaini101: @nrmancuso This property is used to correct the ordered properties.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
AayushSaini101: Done @nrmancuso

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
AayushSaini101: Rewrite the conditions, direclty removing from the set and adding the set causing an error, hence creating list first and then after performing the operation returning immutable list

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
romani: It might looks now different than it was before.
But now I see no difference, and GitHub shows diff with extra lines change but content is not changed

AayushSaini101: Added @romani

nrmancuso: Why did we change `name` to `file`? I would expect something that is called a `file` to be of type `File`. Please revert this change.

romani: @AayushSaini101, please restore this line as it was before.

AayushSaini101: @nrmancuso @romani Done

romani (approved):
Ok to merge

nrmancuso (approved):

PR #13943, opened 2023-10-24
Issue #13809: Kill mutation in TokenUtil

Kevin222004:
Issue #13809: Kill mutation in TokenUtil

# Explaintaion 

https://github.com/checkstyle/checkstyle/blob/0322ba1be8489d1df373013efc9078de4d209eb4/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java#L80 
we are passing object as a param so I think it will not make a difference either we pass only name or whole field

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
@@ -97,7 +97,7 @@ public final class TokenUtil {
         return Arrays.stream(cls.getDeclaredFields())
             .filter(fld -> Modifier.isPublic(fld.getModifiers()) && fld.getType() == Integer.TYPE)
             .collect(Collectors.toUnmodifiableMap(
-                Field::getName, fld -> getIntFromField(fld, fld.getName()))
+                Field::getName, fld -> getIntFromField(fld, null))
             );
     }
 
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -6407,6 +6407,17 @@
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java</fileName>
+    <specifier>argument</specifier>
+    <message>incompatible argument for parameter object of getIntFromField.</message>
+    <lineContent>Field::getName, fld -&gt; getIntFromField(fld, null))</lineContent>
+    <details>
+      found   : null (NullType)
+      required: @Initialized @NonNull Object
+    </details>
+  </checkerFrameworkError>
+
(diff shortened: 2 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -97,7 +97,7 @@ public static Map<String, Integer> nameToValueMapFromPublicIntFields(Class<?> cl
         return Arrays.stream(cls.getDeclaredFields())
             .filter(fld -> Modifier.isPublic(fld.getModifiers()) && fld.getType() == Integer.TYPE)
             .collect(Collectors.toUnmodifiableMap(
-                Field::getName, fld -> getIntFromField(fld, fld.getName()))
+                Field::getName, fld -> getIntFromField(fld, fld))
```
nrmancuso, line 100: `getIntFromField` is basically a wrapper for `Field#getInt`. From reading the docs of [Field](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/reflect/Field.html#getInt(java.lang.Object)), it looks like we should be passing the class itself to this method. I am not sure how this ever worked in the first place. Please do:

```suggestion
                Field::getName, fld -> getIntFromField(fld, cls))
```

nrmancuso (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
rnveach: > I am not sure how this ever worked in the first place

This is because we are going against a static field. It has no instance to speak of, so most likely reflection is ignoring the 2nd parameter we pass. We also have similar funky uses in `TokenUtilTest` as `getIntFromField` is `public` when it should not be.

Truthfully this 2nd parameter should be `null`, as there is no instance.

```suggestion
                Field::getName, fld -> getIntFromField(fld, null))
```

romani (approved):
Ok to merge.

@Kevin222004 , please update Checker config

nrmancuso (approved):
I really don't see any other method in java.lang.reflect.Field that we can use to avoid second parameter. I am good with `null` to make it clear that the second parameter doesn't matter in the case of static fields.

nrmancuso:
Assigned to @rnveach to complete review, I am not sure if we want to open a ticket for refactor of `getIntFromField` or not, I don't think it is a big deal in this case.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
nrmancuso: @rnveach leaving this for you to resolve.

rnveach (approved):

rnveach:
@Kevin222004 If you want an easy assignment, `getIntFromField` can be refactored as mentioned at https://github.com/checkstyle/checkstyle/pull/13943#discussion_r1372055700 .

PR #13944, opened 2023-10-24
Issue #13809: Kill mutation in XpathUtil

Kevin222004:
Issue #13809: Kill mutation in XpathUtil

https://github.com/checkstyle/checkstyle/blob/0322ba1be8489d1df373013efc9078de4d209eb4/config/pitest-suppressions/pitest-utils-suppressions.xml#L354-L361

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.java
@@ -20,10 +20,12 @@
 package com.puppycrawl.tools.checkstyle.utils;
 
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * <p>Note: it simply wraps the existing JDK methods to provide a workaround
@@ -61,6 +63,21 @@ public final class UnmodifiableCollectionUtil {
      */
     public static <T> List<T> unmodifiableList(List<T> collection) {
         return Collections.unmodifiableList(collection);
+    }
+
+    /**
+     * Returns an unmodifiable view of a List containing elements of a specific type.
+     *
+     * @param items The List of items to make unmodifiable.
+     * @param elementType The Class object representing the type of elements in the list.
+     * @param <S> The generic type of elements in the input Collection.
+     * @param <T> The type of elements in the resulting unmodifiable List.
+     * @return An unmodifiable List containing elements of the specified type.
+     */
+    public static <S, T> List<T> unmodifiableList(Collection<S> items, Class<T> elementType) {
+        return items.stream()
+                .map(elementType::cast)
+                .collect(Collectors.toUnmodifiableList());
     }
 
     /**
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/XpathUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/XpathUtil.java
@@ -214,8 +214,6 @@ public final class XpathUtil {
         final XPathDynamicContext xpathDynamicContext = xpathExpression
                 .createDynamicContext(rootNode);
         final List<Item> items = xpathExpression.evaluate(xpathDynamicContext);
-        return items.stream()
-                .map(NodeInfo.class::cast)
-                .collect(Collectors.toUnmodifiableList());
+        return UnmodifiableCollectionUtil.unmodifiableList(items, NodeInfo.class);
     }
 }
--- a/config/checker-framework-suppressions/checker-methods-resource-fenum-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-methods-resource-fenum-suppressions.xml
@@ -244,6 +244,21 @@
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.java</fileName>
+    <specifier>methodref.param</specifier>
+    <message>Incompatible parameter type for obj</message>
+    <lineContent>.map(elementType::cast)</lineContent>
+    <details>
+      found   : @FenumUnqualified Object
+      required: S extends @FenumTop Object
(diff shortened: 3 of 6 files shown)


nrmancuso (changes requested):
We need to understand how pitest is actually mutating this code before we make changes like this. The code proposed by this PR is objectively less clear and concise than what we have now. We should not allow pitest to decrease our code quality, especially if we don’t even understand what it is doing under the hood.

src/main/java/com/puppycrawl/tools/checkstyle/utils/XpathUtil.java
```diff
@@ -214,8 +214,11 @@ public static List<NodeInfo> getXpathItems(String xpath, AbstractNode rootNode)
         final XPathDynamicContext xpathDynamicContext = xpathExpression
                 .createDynamicContext(rootNode);
         final List<Item> items = xpathExpression.evaluate(xpathDynamicContext);
-        return items.stream()
-                .map(NodeInfo.class::cast)
-                .collect(Collectors.toUnmodifiableList());
+        final List<NodeInfo> nodeInfoList = new ArrayList<>();
```
romani, line 217: please make method `UnmodifiableCollectionUtil.unmodifiableList(items, NodeInfo.class)`
to keep what pitest naturaly see as not required code is special place/class.

romani (changes requested):
items:

romani (approved):
Ok to merge.

@Vyom-Yadav , please share your opinion on cast problem, looks like this becomes hot topic.

romani:
@Kevin222004 , please rebase to resolve conflict

src/main/java/com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.java
```diff
@@ -63,6 +67,20 @@ public static <T> List<T> unmodifiableList(List<T> collection) {
...
 
+    /**
+     * Returns an unmodifiable view of a List containing elements of a specific type.
+     *
+     * @param items The List of items to make unmodifiable.
+     * @param elementType The Class object representing the type of elements in the list.
+     * @param <T> The type of elements in the resulting unmodifiable List.
+     * @return An unmodifiable List containing elements of the specified type.
+     */
+    public static <T> List<T> unmodifiableList(Collection<Item> items, Class<T> elementType) {
```
rdiachenko, line 78: Why do you use exact type here (which is `Item`). This method is supposed to be generic. Can we update it as follows?
```
public static <S, T> List<T> unmodifiableList(Collection<S> items, Class<T> elementType) { ... }
```
here, `S` means source type and `T` means target type.

romani: @Kevin222004 , please fix as suggested.

Kevin222004: done

rdiachenko: Do we need this block after the above change? Could you remove it if not needed?
```
    <file name="UnmodifiableCollectionUtil">
       <allow pkg="net.sf.saxon"/>
     </file>
```

Kevin222004: @rdiachenko removed.

romani:
@Kevin222004 , please address Checker

rdiachenko (approved):
lgtm

@Kevin222004 please resolve checker errors

Vyom-Yadav (approved):
LGTM! Please fix CI errors.

src/main/java/com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.java
```diff
@@ -63,6 +65,21 @@ public static <T> List<T> unmodifiableList(List<T> collection) {
...
+     * @param elementType The Class object representing the type of elements in the list.
+     * @param <S> The generic type of elements in the input Collection.
+     * @param <T> The type of elements in the resulting unmodifiable List.
+     * @return An unmodifiable List containing elements of the specified type.
+     */
+    public static <S, T> List<T> unmodifiableList(Collection<S> items, Class<T> elementType) {
+        return items.stream()
+                .map(elementType::cast)
+                .collect(Collectors.toUnmodifiableList());
+    }
```
nrmancuso, line 81: Why did we make this about collections, when the issue is with the calls to `Class#cast`?  If we want to make some hack to bypass pitest, why not just have a util method to cast the stream as an intermediate operation?

Pseudocode:
```
<R, T> Stream<T> castedStreamOF(Stream<R> s, Class<T> c) {
  return stream.map(c::cast)
}

```

Then we can collect or perform subsequent stream operations elsewhere more readily.

nrmancuso (dismissed):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.java
romani: It has unmodified collection, that pitest will complain about, not sure why it is not doing it now.

And current implementation is easier, and we don't want to make it highly reusable, at least for now.

nrmancuso: Why is it a public method in a utility class if we don't want it to be reusable?

romani: Pitest limitation to allow easy exclude from mutation by class name

nrmancuso (commented):
The changes here don't add up for me. In order to remove one pitest suppression (that should just be permanent IMO), we have added:
 - 2 checker suppressions (tech debt)
 - avoidCallsTo suppression in pom.xml (tech debt)
 - yet another specialized utility method that won't get used anywhere else (tech debt)
 - Production code (the reason why we are really here) that is much less clear than before

At this point, we are shuffling tech debt around. I won't block this, but I cannot approve either.

nrmancuso:
@romani let’s just merge this and move on. We have already spent too much time and resources on this PR.

PR #13948, opened 2023-10-24
Issue #13809: Kill mutation in CommonUtil

Kevin222004:
Issue #13809: Kill mutation in CommonUtil


src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java
```diff
@@ -123,20 +123,8 @@ public static boolean matchesFileExtension(File file, String... fileExtensions)
...
-                final String extension = fileExtensions[i];
-                if (extension.startsWith(EXTENSION_SEPARATOR)) {
-                    withDotExtensions[i] = extension;
-                }
-                else {
-                    withDotExtensions[i] = EXTENSION_SEPARATOR + extension;
-                }
-            }
-
             final String fileName = file.getName();
```
Kevin222004, line 126: @romani do we really required to do this at the end `fileName.endsWith(fileExtension)` we are checking only wether file names end with fileExtension.

romani: I am surprised on this code 
The only place where we do mentioning `.` in extension name is at https://checkstyle.org/config.html#TreeWalker_Properties

All other places are without dot.

romani: Extension is not a standard , it is just agreement that file name ends with some characters.

We should not parse java file that is named as `Test,java` so looks like we missed test on names without dots as separator of name and extension.

Kevin222004: Test added

romani (approved):
Ok to merge

PR #13284, opened 2023-06-21
Issue #13328: Kill mutation for IllegalInstantiationCheck

Kevin222004:
Issue #13328: Kill mutation for IllegalInstantiationCheck

-----

Check :
https://checkstyle.org/config_coding.html#IllegalInstantiation

-----

# Mutation 
https://github.com/checkstyle/checkstyle/blob/6324a9d693324eb755c2cdfdbd90c2300d3c7f02/config/pitest-suppressions/pitest-coding-2-suppressions.xml#L30-L37

-----

# Explaintaion 
Tests Added

-------

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/3bd236a24ed8eacd760e6b12b47f7455/raw/97d7a64aa97fc35488a6db5d7345ee23ae9b9507/olc.xml
Diff Regression projects: https://gist.githubusercontent.com/Kevin222004/9600f179b602d4c971bdb0a050099005/raw/360a95ed7bb60d7a0956e531199d484c4d6f6617/test-projects.properties
Report label: Regression-1

PR #13953, opened 2023-10-26
Issue #13809: Kill mutation in CommonUtil

Kevin222004:
Issue #13809: Kill mutation in CommonUtil

regression is taken from https://github.com/checkstyle/checkstyle/pull/13284
gist: https://gist.github.com/romani/f15f3afabac0f9ad1e63f429b62d445f

Diff Regression config: https://gist.githubusercontent.com/romani/f15f3afabac0f9ad1e63f429b62d445f/raw/40dfb3e50dc3e5606844f00d074f572be4682079/Config.xml
Report label: Regression-with-path-for-basedir

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java
@@ -246,7 +246,7 @@ public final class CommonUtil {
     }
 
     /**
-     * Constructs a normalized relative path between base directory and a given path.
+     * Constructs a relative path between base directory and a given path.
      *
      * @param baseDirectory
      *            the base path to which given path is relativized
@@ -255,14 +255,14 @@ public final class CommonUtil {
      * @return the relative normalized path between base directory and
      *     path or path if base directory is null.
      */
-    public static String relativizeAndNormalizePath(final String baseDirectory, final String path) {
+    public static String relativizePath(final String baseDirectory, final String path) {
         final String resultPath;
         if (baseDirectory == null) {
             resultPath = path;
         }
         else {
-            final Path pathAbsolute = Paths.get(path).normalize();
-            final Path pathBase = Paths.get(baseDirectory).normalize();
+            final Path pathAbsolute = Paths.get(path);
+            final Path pathBase = Paths.get(baseDirectory);
             resultPath = pathBase.relativize(pathAbsolute).toString();
         }
         return resultPath;
--- a/config/pitest-suppressions/pitest-common-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-suppressions.xml
@@ -5,8 +5,8 @@
     <mutatedClass>com.puppycrawl.tools.checkstyle.Checker</mutatedClass>
     <mutatedMethod>acceptFileStarted</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/utils/CommonUtil::relativizeAndNormalizePath with argument</description>
-    <lineContent>final String stripped = CommonUtil.relativizeAndNormalizePath(basedir, fileName);</lineContent>
+    <description>replaced call to com/puppycrawl/tools/checkstyle/utils/CommonUtil::relativizePath with argument</description>
+    <lineContent>final String stripped = CommonUtil.relativizePath(basedir, fileName);</lineContent>
   </mutation>
 
   <mutation unstable="false">
--- a/config/pitest-suppressions/pitest-utils-suppressions.xml
+++ b/config/pitest-suppressions/pitest-utils-suppressions.xml
@@ -10,24 +10,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>CommonUtil.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.utils.CommonUtil</mutatedClass>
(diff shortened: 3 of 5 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java
```diff
@@ -261,8 +261,8 @@ public static String relativizeAndNormalizePath(final String baseDirectory, fina
             resultPath = path;
         }
         else {
-            final Path pathAbsolute = Paths.get(path).normalize();
-            final Path pathBase = Paths.get(baseDirectory).normalize();
+            final Path pathAbsolute = Paths.get(path);
```
romani, line 264: Looks like we need to test message of violation and validate how path is starting https://www.google.com/amp/s/www.geeksforgeeks.org/path-normalize-method-in-java-with-examples/amp/

To be OS agnostic we usually avoid checking full path in violation, we might do only file name

romani (changes requested):
Items :

src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java
romani: What we can test is ... We set path with `.` and/or `..` but have in violation path without such symbols.

Kevin222004: I have tried multiple ways but at the i didnt find a way in which relativize path is getting affected.

romani: yes, such normalization is done by jdk
https://github.com/openjdk/jdk/blob/db3402577a2c14a41045753a1ffe2829a6bdda91/src/java.base/unix/classes/sun/nio/fs/UnixPath.java#L478-L495

we already have test on this https://github.com/checkstyle/checkstyle/blob/bd3a9bb9ec9cf804f628171c2e7557b0afbcfd93/src/test/java/com/puppycrawl/tools/checkstyle/utils/CommonUtilTest.java#L221-L224

romani: image

romani (approved):
ok to merge

rnveach:
1) Where is regression showing this won't be an issue? This is used by Checker.
2) This method is not really called "relativizeAnd**Normalize**Path" if there is no more normalize.

romani:
> Where is regression showing this won't be an issue? This is used by Checker.

regression is only covering behavior on config, to make this regression something weird should be done with paths to validate. non of our tools can do this.

> This method is not really called "relativizeAndNormalizePath" if there is no more normalize.

relativize does normalize. Do you propose to change name of util method ?

rnveach:
Regression should also cover file paths which is done by Checker, as 2 paths need to line up for a match for differencing. If it truthfully has no impact, regression will show no differences either way and will show we won't have anything to worry about after merge.

I am leaning to changing the util name and javadoc since I assume it was named normalize for the normalize that happened in it.

romani:
method is renamed.

romani:
Github, generate report

romani:
@rnveach , please review.

rnveach:
@romani Please make sure you are using a config to hit the code. `baseDir` has to be set to trigger as shown in the method in this pr.

src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java
rnveach: > We do not need to care about path in zips 

You may want to recheck this. Some users put configuration files into external JARs controlled by Maven. Those are basically ZIP files.

romani: We will happily fix issue if they will show how to reproduce issue.

romani:
Github, generate repor

romani:
Github, generate report

rdiachenko (approved):
lgtm

normalization is done as part of `relativize` for UnixPath

romani:
regression report needs to pass, I am not sure why it is failing with exception.

Kevin222004:
Github, generate report

Vyom-Yadav:
@romani @Kevin222004 I think there is a problem with relativizing path.
```
Caused by: java.lang.IllegalArgumentException: 'other' is different type of Path
    at sun.nio.fs.UnixPath.relativize (UnixPath.java:429)
    at sun.nio.fs.UnixPath.relativize (UnixPath.java:43)
    at com.puppycrawl.tools.checkstyle.utils.CommonUtil.relativizeAndNormalizePath (CommonUtil.java:266)
```
You may want to take a look at https://stackoverflow.com/q/16299604/15412365

romani:
problem is at `relativizeAndNormalizePath` , so it is execution without update of this PR.
so exception is not related to PR update, and this is just bad config.

romani:
Github, generate report

romani:
Github, generate report

romani:
@rnveach , report is ready.

romani (approved):
Ok to merge

rnveach:
@romani So we are not going to do anything on https://github.com/checkstyle/checkstyle/pull/13953#discussion_r1375561273 ?

romani:
@rnveach , yes, let's have regression if it's actually matters.
I don't know how to deal with such path in zip for checkstyle.

rnveach (approved):

nrmancuso (approved):

rnveach:
Apparently this change failed in backport (JDK 8). Nothing changed in the javadoc for `Path.relativize`.

Issue #13772, opened 2023-09-23
IllegalImport with class pattern fails on static import


Bananeweizen:
`IllegalImport` with property `illegalClasses` configured doesn't detect illegal static imports, because it just does a text comparison of the import statement (which contains the method names for static imports): https://github.com/checkstyle/checkstyle/blob/c525c6732e2c1f15bd3f85f943770c649f1ab775/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/IllegalImportCheck.java#L245

```bash
/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <property name="severity" value="error"/>
  <property name="fileExtensions" value="java, properties, xml"/>
  <module name="TreeWalker">
    <module name="IllegalImport">
      <property name="illegalClasses" value="org.junit.jupiter.api.Assertions"/>
    </module>
  </module>
</module>

/var/tmp $ cat TestStatic.java
import static org.junit.jupiter.api.Assertions.fail;

public class TestStatic {

  void useTheImport() {
    fail();
  }
}

/var/tmp $ cat Test.java
import org.junit.jupiter.api.Assertions;

public class Test {

  void useTheImport() {
    Assertions.fail();
  }
}

❯ java %RUN_LOCALE% -jar checkstyle-10.12.3-all.jar -c config.xml Test.java
Starting audit...
[ERROR] C:\dev\eclipse-cs\ws\Reproducer\Test.java:1:1: Illegal import - org.junit.jupiter.api.Assertions. [IllegalImport]
Audit done.
Checkstyle ends with 1 errors.

❯ java %RUN_LOCALE% -jar checkstyle-10.12.3-all.jar -c config.xml TestStatic.java
Starting audit...
Audit done.
```
Both files should report the same issue.

Causes https://github.com/checkstyle/eclipse-cs/issues/523

nrmancuso:
Link to check documentation: https://checkstyle.sourceforge.io/checks/imports/illegalimport.html#IllegalImport

AayushSaini101:
I am on it

nrmancuso:
I am removing 'approved' label, since this seems like the expected check behavior.

From https://checkstyle.sourceforge.io/checks/imports/illegalimport.html#IllegalImport:

[...]

PR #13960, opened 2023-10-28
Issue #13772: IllegalImport with class pattern fails on static import

AayushSaini101:
Related: https://github.com/checkstyle/checkstyle/issues/13772

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/IllegalImportCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/IllegalImportCheck.java
@@ -139,7 +139,7 @@ public class IllegalImportCheck
         illegalPkgs = from.clone();
         illegalPkgsRegexps.clear();
         for (String illegalPkg : illegalPkgs) {
-            illegalPkgsRegexps.add(CommonUtil.createPattern("^" + illegalPkg + "\\..*"));
+            illegalPkgsRegexps.add(CommonUtil.createPattern(illegalPkg + "\\..*"));
         }
     }
 
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/IllegalImportCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/IllegalImportCheckTest.java
@@ -157,4 +157,25 @@ public class IllegalImportCheckTest extends AbstractModuleTestSupport {
                 getPath("InputIllegalImportDefault7.java"), expected);
     }
 
+    @Test
+    public void testIllegalPackagesAndClassesRegularExpression2()
+            throws Exception {
+
+        final String[] expected = {
+            "15:1: " + getCheckMessage(MSG_KEY, "java.util.List"),
+            "16:1: " + getCheckMessage(MSG_KEY, "java.util.List"),
+            "19:1: " + getCheckMessage(MSG_KEY, "java.util.Enumeration"),
+            "20:1: " + getCheckMessage(MSG_KEY, "java.util.Arrays"),
+            "33:1: " + getCheckMessage(MSG_KEY, "java.awt.Component"),
+            "34:1: " + getCheckMessage(MSG_KEY, "java.awt.Graphics2D"),
+            "35:1: " + getCheckMessage(MSG_KEY, "java.awt.HeadlessException"),
+            "36:1: " + getCheckMessage(MSG_KEY, "java.awt.Label"),
+            "37:1: " + getCheckMessage(MSG_KEY, "java.util.Date"),
+            "38:1: " + getCheckMessage(MSG_KEY, "java.util.Calendar"),
+            "39:1: " + getCheckMessage(MSG_KEY, "java.util.BitSet"),
+        };
+        verifyWithInlineConfigParser(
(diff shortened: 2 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/IllegalImportCheck.java
```diff
@@ -139,7 +139,7 @@ public final void setIllegalPkgs(String... from) {
         illegalPkgs = from.clone();
         illegalPkgsRegexps.clear();
         for (String illegalPkg : illegalPkgs) {
-            illegalPkgsRegexps.add(CommonUtil.createPattern("^" + illegalPkg + "\\..*"));
+            illegalPkgsRegexps.add(CommonUtil.createPattern(illegalPkg + "\\..*"));
```
rnveach, line 142: Won't this possibly create false matches like if my illegal package is `blah` and my input file has a package `com.other.blah.other`?

AayushSaini101: @rnveach This will cause an error when the blah is found in the package, and this is an illegal package

rnveach: I am not sure I follow what you mean by an error. I assume violation.

But my question was would this be a false match. If I say my illegal package is `blah`, I expect to only match `blah` or `blah.other` depending on the rest of the criteria for this module. I wouldn't expect a match on `com.other.blah.other`.

AayushSaini101: > I am not sure I follow what you mean by an error. I assume violation.
> 
> But my question was would this be a false match. If I say my illegal package is `blah`, I expect to only match `blah` or `blah.other` depending on the rest of the criteria for this module. I wouldn't expect a match on `com.other.blah.other`.

Thanks @rnveach In this case it will give false violation i removed the ^ symbol, can we add some conditions like only static or some more keywords allow only

rnveach: You would need to examine the module closer to see if there is some conditional statements that would have to be added. Like other issues, I have not looked into the full details of the issue and code.

However, to me it does seem odd that this issue can be resolved with what amounts to a configuration change from looking at this PR. If your changing the code in the property setter, your basically saying either how we read the value from the configuration, or the contents of the configuration, are wrong.

romani:
While you try to do code analysis, you can create diff report by GitHub action and it might catch regression  for you, so you will have ready test input file where regression is visible.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/IllegalImportCheck.java
AayushSaini101: @romani @rnveach There are the proper usage how to use import in different, For this issue, We need a condition only for handling the static instead of removing the "^" from the import check.

- Single Type Import:

This is the most common usage of the import statement and is used to import a single class or interface.

`import java.util.ArrayList;`
Import All Classes from a Package:
You can import all classes and interfaces from a specific package by using the * wildcard.


`import java.util.*;`
This will import all classes and interfaces within the java.util package.

- Static Import:

The import statement can also be used to import static members (fields and methods) of a class. This is often used for constants or utility methods.


`import static java.lang.Math.*;`
This allows you to use static members of the Math class directly without prefixing with Math..

- On-Demand Type Import:

This is similar to importing all classes from a package but only imports the specific class you use in your code.


```
import java.util.ArrayList;
import java.util.HashMap;
```
In this example, only ArrayList and HashMap are imported from the java.util package, even though other classes and interfaces are available in that package.

- Import for Nested Classes:

If you have a nested class, you can import it in a similar way, using the nested class name.


`import mypackage.OuterClass.InnerClass;`
This imports the InnerClass which is a nested class within OuterClass.

rnveach: > java.util.HashMap
> mypackage.OuterClass.InnerClass

Just pointing out that you can't identify an inner class from a regular class without multi-file and classpath support.

AayushSaini101: @rnveach @romani can we add separate property for this for handling static import?

romani: It is completely impossible for checkstyle to cover it, so we simply need to skip such cases from validation.

aayushRedHat:
@romani @rnveach  What update i need to do this in issue ?

rnveach:
@aayushRedHat See https://github.com/checkstyle/checkstyle/pull/13960#discussion_r1377565209

romani:
@aayushRedHat, please create new PR with account aayushRedHat, to have in CI all 113 executions

aayushRedHat:
> @aayushRedHat, please create new PR with account aayushRedHat, to have in CI all 113 executions

Sure @romani Closing This PR

PR #13975, opened 2023-10-31
Issue #13809: Kill mutation in Tree-Walker

Kevin222004:
Issue #13809: Kill mutation in Tree-Walker

Test added

PR #13976, opened 2023-10-31
Issue #13809: Kill mutation in Tree-Walker-2

Kevin222004:
Issue #13809: Kill mutation in Tree-Walker-2

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
@@ -33,6 +33,7 @@ import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -679,6 +680,38 @@ public class TreeWalkerTest extends AbstractModuleTestSupport {
         verify(checkerConfig, filePath, expected);
     }
 
+    /**
+     * This test is checking that Checks execution ordered by name.
+     *
+     * @throws Exception if file is not found
+     */
+    @Test
+    public void testOrderOfCheckExecution() throws Exception {
+
+        final DefaultConfiguration configuration1 = createModuleConfig(AaCheck.class);
+        configuration1.addProperty("id", "2");
+        final DefaultConfiguration configuration2 = createModuleConfig(BbCheck.class);
+        configuration2.addProperty("id", "1");
+
+        final DefaultConfiguration treeWalkerConfig = createModuleConfig(TreeWalker.class);
+        treeWalkerConfig.addChild(configuration2);
+        treeWalkerConfig.addChild(configuration1);
+
+        final List<File> files =
+                Collections.singletonList(new File(getPath("InputTreeWalker2.java")));
+        final Checker checker = createChecker(treeWalkerConfig);
+
+        try {
+            checker.process(files);
+            assertWithMessage("exception is expected").fail();
+        }
+        catch (CheckstyleException exception) {
+            assertWithMessage("wrong order of Check executions")
+                    .that(exception.getCause().getMessage())
+                    .isEqualTo(AaCheck.class.toString());
+        }
+    }
+
     public static class BadJavaDocCheck extends AbstractCheck {
 
         @Override
@@ -769,6 +802,54 @@ public class TreeWalkerTest extends AbstractModuleTestSupport {
         @Override
         public boolean isCommentNodesRequired() {
             return true;
+        }
+
+    }
+
+    public static class AaCheck extends AbstractCheck {
+
+        @Override
+        public int[] getDefaultTokens() {
+            return new int[0];
+        }
+
+        @Override
+        public int[] getAcceptableTokens() {
+            return new int[0];
+        }
+
+        @Override
+        public int[] getRequiredTokens() {
+            return new int[0];
+        }
+
+        @Override
+        public void beginTree(DetailAST rootAST) {
+            throw new IllegalStateException(AaCheck.class.toString());
+        }
+
+    }
+
+    public static class BbCheck extends AbstractCheck {
+
+        @Override
+        public int[] getDefaultTokens() {
+            return new int[0];
+        }
+
+        @Override
+        public int[] getAcceptableTokens() {
+            return new int[0];
+        }
+
+        @Override
+        public int[] getRequiredTokens() {
+            return new int[0];
+        }
+
+        @Override
+        public void beginTree(DetailAST rootAST) {
+            throw new IllegalStateException(BbCheck.class.toString());
         }
 
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
```diff
@@ -424,10 +424,7 @@ private void processIter(DetailAST root, AstState astState) {
      */
     private static SortedSet<AbstractCheck> createNewCheckSortedSet() {
         return new TreeSet<>(
-                Comparator.<AbstractCheck, String>comparing(check -> check.getClass().getName())
-                        .thenComparing(AbstractCheck::getId,
-                                Comparator.nullsLast(Comparator.naturalOrder()))
-                        .thenComparing(AbstractCheck::hashCode));
+                Comparator.comparing(AbstractCheck::hashCode));
```
Kevin222004, line 427: I understand this code but I am not able to create any test cases for this.

romani: Not clear what mutation you tried to kill 
As you already have some success at https://github.com/checkstyle/checkstyle/pull/13975

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
```diff
@@ -424,8 +424,7 @@ private void processIter(DetailAST root, AstState astState) {
      */
     private static SortedSet<AbstractCheck> createNewCheckSortedSet() {
         return new TreeSet<>(
-                Comparator.<AbstractCheck, String>comparing(check -> check.getClass().getName())
-                        .thenComparing(AbstractCheck::getId,
+                Comparator.comparing(AbstractCheck::getId,
                                 Comparator.nullsLast(Comparator.naturalOrder()))
```
romani, line 428: https://pitest.org/quickstart/mutators/#EMPTY_RETURNS

We need to have different Checks in config. To let violations be sorted by check name.
But it might require non regular verify method, as we need strictly check order of violations, existing verify method might do comparison in non strict ways or by its own sorting.

Please look inside verify method on how it compares violations.

romani: One more mutation is required to be removed https://github.com/checkstyle/checkstyle/actions/runs/6723258174/job/18272949364?pr=13976#step:7:24

romani (approved):
ok to merge

strange, on pushing from local a bit different code but Github is not showing it.

rnveach:
https://github.com/checkstyle/checkstyle/pull/14015
https://github.com/checkstyle/checkstyle/pull/13976

One of these need to be closed or explained why they look so identical.

romani:
There was some problem with GitHub, it didn't show in PR my latest push, so I created new PR, looks like now all is good.

rnveach:
@nrmancuso @romani It looks like URL got "Check no closed issue references" is being generated wrong.
https://github.com/checkstyle/checkstyle/actions/runs/6831526581/job/18593284692?pr=13976
> https://github.com/romani/checkstyle/blob/Kevin222004/treeWalker-2/config/checkstyle-non-main-files-suppressions.xml#L256

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
```diff
@@ -679,6 +680,41 @@ public void testExternalResourceFiltersWithNoExternalResource() throws Exception
         verify(checkerConfig, filePath, expected);
     }
 
+    /**
+     * This test is required for pitest coverage.
+     * Order of Check execution is not visible from outside
+     * as we order violations by line and column.
+     * This test is checking that Checks execution ordered by name and then by id.
```
rnveach, line 687: > Order of Check execution is not visible from outside

This is somewhat incorrect. This code came from random ordering of Javadoc checks. Multiple checks printed the same violation, but different executions picked which check was being printed as the source of the violation. This would cause differences in difference reports as we display who is the source of the violation.

The specific code from pitesting shows, we order **Checks** by class name, id, and then by hashcode. Those are the ordering of the execution of checks. Order of violations is defined elsewhere.

> ordered by name and then by id.

Show how this test ensures checks are ordered by id too? I only see somewhat reliance on check name ordering.

rnveach (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
rnveach: The whole javadoc should just be:

`This test is checking that Checks execution ordered by name.`

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
```diff
@@ -679,6 +680,41 @@ public void testExternalResourceFiltersWithNoExternalResource() throws Exception
...
+        final DefaultConfiguration treeWalkerConfig = createModuleConfig(TreeWalker.class);
+        treeWalkerConfig.addChild(configuration2);
+        treeWalkerConfig.addChild(configuration1);
+
+        final List<File> files =
+                Collections.singletonList(new File(getPath("InputTreeWalker2.java")));
+        final Checker checker = createChecker(treeWalkerConfig);
+
+        try {
+            checker.process(files);
```
rnveach, line 708: Why can't this be a BDD test? I am not seeing anything really preventing it.

rnveach (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
romani: All tries to make it in bad style did not kill mutation. Because mutation on order of checks, order of checks is not publicly visible.
I tried bunch and gave up.

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
romani: applied.

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
nrmancuso: Have we tried something like the following?
- config with two checks
- both checks have bad config (maybe property name)
- we assert that it is the first check in the config that reported in the exception
?

romani: I am not , but exception will happen before Check is added to sorted-set of Checks of Treewker.
Order of check to load from config is defined by xml, predictable order, but mutation is for order of execution in runtime (traverse of tree).

I tried to keep test clear that first execution on beginTree will throw exception, so it very clear definition that Checks are ordered.
After bunch of time with this mutation, I am very ok to keep it as non bdd, because final variant of test is still easy to understand, I hope.

nrmancuso: Yes, test is easy to understand.

nrmancuso (approved):
A hack for sure, but let's not spend too much time on this.

PR #13981, opened 2023-11-02
Issue #13809: Kill mutation in Checker

nrmancuso:
Part of #13809 

https://sonarcloud.io/project/issues?resolved=false&severities=MAJOR&sinceLeakPeriod=true&pullRequest=13981&id=org.checkstyle%3Acheckstyle&open=AYuOAJXquUWup32NLidh

Change from method call -> variable must be enough to fool sonar into thinking this is new code; sonar violation should be suppressed after merge to master.

### DIFF
--- a/config/pitest-suppressions/pitest-common-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-suppressions.xml
@@ -12,15 +12,6 @@
   <mutation unstable="false">
     <sourceFile>Checker.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.Checker</mutatedClass>
-    <mutatedMethod>processFiles</mutatedMethod>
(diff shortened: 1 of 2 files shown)


romani:
We can use probably https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/lang/AssertionError.html

It is pretty generic.

All other subclasses of Error is very specific.

@rnveach , are you ok to throw AssertionError ?

rnveach:
@romani AssertionError seems unrelated to this issue of killing the mutation with file path.

> Thrown to indicate that an assertion has failed. 

As for the question, I am not sure I see this area as asserting anything, so I don't see using AssertionError making more sense than just a general error.

romani (approved):
ok to merge, 
we will suppress in sonar as wontfix.

Issue #13544, opened 2023-08-09
Re-enable ArchUnit Testing


Kevin222004:
Temporarily we are skipping the ArchUnit Testing to avoid conflicts in Pitest-Resolution in summer 2023 at https://github.com/checkstyle/checkstyle/pull/13542. 

The issue is opened to re-enable them all at the end of summer.

romani:
Not approved until end of pitest project

romani:
Moved to https://github.com/checkstyle/checkstyle/issues/14053

PR #14009, opened 2023-11-08
Issue #13544: Re-enable ArchUnit Testing

Kevin222004:
Issue #13544: Re-enable ArchUnit Testing


romani:
Let's last pack of PRs for pitest to pass before we activate this.

romani:
Bunch of changes https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=17687&view=logs&jobId=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111


@nrmancuso , @rnveach , please share your opinion on usage of this cycle validation, I have fillings that it is too error prone to be used.

rnveach:
I haven't seen how we are suppose to resolve the cycle issues. I can't say if its good or bad until then.

nrmancuso:
> I haven't seen how we are suppose to resolve the cycle issues. I can't say if its good or bad until then.

Yeah, I agree here. If this is going to be like checker and just noise in every PR, we should leave it deactivated until we have a clear plan on how to fix these issues.

romani:
Problem is that we do not make issues, fact that we have line numbers inside this file causing it to be updated. 
May be trim of line numbers will help, I am not sure.

@Vyom-Yadav , what do you think?

nrmancuso:
> Problem is that we do not make issues

This is not a problem at all, and more reason not to keep this test. We aren’t adding to this tech debt, and it just creates noise in PRs and confuses contributors.

romani:
I meant issues of cycles.
Ideally we should be just notified on new cycle creation by failure of this test, but unfortunately if failing all the time for any refactoring, and even we introduce new cycle, we will not look at and simply update this file.

Vyom-Yadav:
> Problem is that we do not make issues, fact that we have line numbers inside this file causing it to be updated.
May be trim of line numbers will help, I am not sure.

@romani Aren't line numbers excluded by default. According to the [docs](https://www.archunit.org/userguide/html/000_Index.html#_usage), it should be:

> On the first run all violations of that rule will be stored as the current state. On consecutive runs only new violations will be reported. By default FreezingArchRule will ignore line numbers, i.e. **if a violation is just shifted to a different line, it will still count as previously recorded and will not be reported.**

Vyom-Yadav:
I agree with @romani, getting rid of cycles would only involve package refactoring and not logical changes (most proly). If we create a good 3rd/4th issue for this, like we did for specifying violation messages, maybe over time contributors can pick this up and help us.

I think we should include this rule if line-number issue isn't there.

romani:
FreezingArchRule ignore them, but git is not :),
Should we try to update our file to remove line numbers?

nrmancuso:
>  If we create a good 3rd/4th issue for this, like we did for specifying violation messages

Ok, please do this and break it down into logical sections that contributors can handle before we merge this PR to help us understand the scope of this work.

romani:
We can not fix this issue by outside contributors, as even I don't know how to resolve it, it will be significant design change, it will evolve bunch of activities to propose and get agreement from all maintainers.

rnveach:
So where is follow up issues?

romani:
@rnveach , what do you mean ?
We just discussing a plan, if you agree to not activate we will try to remove line numbers or completely remove implementation (in separate followup issues).

rnveach:
I can't agree or not to agree without knowing how we resolve this. If we don't do this, then close the pr.

romani:
Ok now we know that nobody what to keep it, so we need find easy remedy to avoid line numbers in report or we deleting  this validation until special project to fix it completely.

romani:
New issue is created to investigate more on options before final deprecation.
https://github.com/checkstyle/checkstyle/issues/14053

PR #14032, opened 2023-11-16
Issue #11608: Detect return after if in SimplifyBooleanReturnCheck

### LINKED ISSUE #11608: SimplifyBooleanReturn false negatives
I have read check documentation: https://checkstyle.org/config_coding.html#SimplifyBooleanReturn

```bash
/var/tmp $ javac Test.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <module name="TreeWalker">
    <module name="SimplifyBooleanReturn"/>
  </module>
</module>

/var/tmp $ cat Test.java
// Taken from tests: https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/simplifybooleanreturn/InputSimplifyBooleanReturn.java
public class Test
{
    public static boolean isOddMillis()
    {
        boolean even = System.currentTimeMillis() % 2 == 0;

        // can be simplified to "if (even)"
        if (even == true) { // violation
            return false;
        }
        else {
            return true;
        }
        // return can be simplified to "return !even"
    }

    public static boolean isOddMillis2()
    {
        boolean even = System.currentTimeMillis() % 2 == 0;
        // can be simplified to "return !even"
        if (!even) // violation
            return true;
        else
            return false;
    }

    public static boolean giveMeTrue()
    {
        boolean tt = isOddMillis() || true;
        boolean ff = isOddMillis() && false;
        return !false || (true != false);
    }

    public boolean ifNoElse()
    {
        if (isOddMillis()) {
            return true;
        }
        return false;
    }

    boolean a() {
        boolean asd = false;
        boolean dasa = true;

        if (asd) {
            return true;
        } else {
            return dasa;
        }
    }

    boolean aa() {
        boolean asd = false;
        boolean dasa = true;

        if (asd) {
            return dasa;
        } else {
            return true;
        }
    }

    boolean b() {
        boolean asd = false;

        if(asd);
        else;

        return true;
    }
}


/var/tmp $ java -jar $RUN_LOCALE -jar checkstyle-10.2-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/checkstyle/Test.java:9:9: Conditional logic can be removed. [SimplifyBooleanReturn]
[ERROR] /var/tmp/checkstyle/Test.java:22:9: Conditional logic can be removed. [SimplifyBooleanReturn]
Audit done.
```

---

I expect this check to also handle other cases from https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/simplifybooleanreturn/InputSimplifyBooleanReturn.java, though it seems like only the first two cases are currently being handled.

jungm:
Resolves #11608

Diff Regression config: https://gist.githubusercontent.com/jungm/21e7f2e5c9372238726f891840e31106/raw/f57c6c1239a5b766c9e20f0a1d32eff00a4948c1/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/SimplifyBooleanReturnCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/SimplifyBooleanReturnCheck.java
@@ -94,19 +94,23 @@ public class SimplifyBooleanReturnCheck
         // thenStatement
         // [ LITERAL_ELSE (with the elseStatement as a child) ]
 
-        // don't bother if this is not if then else
-        final DetailAST elseLiteral =
-            ast.findFirstToken(TokenTypes.LITERAL_ELSE);
-        if (elseLiteral != null) {
-            final DetailAST elseStatement = elseLiteral.getFirstChild();
+        // skip '(' and ')'
+        final DetailAST condition = ast.getFirstChild().getNextSibling();
+        final DetailAST thenStatement = condition.getNextSibling().getNextSibling();
 
-            // skip '(' and ')'
-            final DetailAST condition = ast.getFirstChild().getNextSibling();
-            final DetailAST thenStatement = condition.getNextSibling().getNextSibling();
-
-            if (canReturnOnlyBooleanLiteral(thenStatement)
-                && canReturnOnlyBooleanLiteral(elseStatement)) {
-                log(ast, MSG_KEY);
+        if (canReturnOnlyBooleanLiteral(thenStatement)) {
+            final DetailAST elseLiteral = ast.findFirstToken(TokenTypes.LITERAL_ELSE);
+            if (elseLiteral != null) {
+                final DetailAST elseStatement = elseLiteral.getFirstChild();
+                if (canReturnOnlyBooleanLiteral(elseStatement)) {
+                    log(ast, MSG_KEY);
+                }
+            }
+            else {
+                final DetailAST statementAfterLiteralIf = ast.getNextSibling();
+                if (isBooleanLiteralReturnStatement(statementAfterLiteralIf)) {
+                    log(ast, MSG_KEY);
+                }
             }
         }
     }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/SimplifyBooleanReturnCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/SimplifyBooleanReturnCheckTest.java
@@ -39,6 +39,7 @@ public class SimplifyBooleanReturnCheckTest
         final String[] expected = {
             "22:9: " + getCheckMessage(MSG_KEY),
             "35:9: " + getCheckMessage(MSG_KEY),
+            "74:9: " + getCheckMessage(MSG_KEY),
         };
         verifyWithInlineConfigParser(
                 getPath("InputSimplifyBooleanReturn.java"), expected);
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/simplifybooleanreturn/InputSimplifyBooleanReturn.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/simplifybooleanreturn/InputSimplifyBooleanReturn.java
@@ -60,11 +60,30 @@ public class InputSimplifyBooleanReturn
         }
     }
 
+    public boolean tryToProvokeNPE2() {
+        if (false) {
+        }
+        else if (true) {
+            return true;
+        }
+        return false;
+    }
+
     public boolean ifNoElse()
     {
+        if (isOddMillis()) { // violation
+            return true;
+        }
+        return false;
+    }
+
+    public boolean ifNoElse2() {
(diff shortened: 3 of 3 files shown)


jungm:
XWiki CI fail is an actual violation: [ExternalExtensionCheck.java:103:9](https://github.com/xwiki/xwiki-commons/blob/2dc71724c7def7470bb2c602f881aba8c5b34cd4/xwiki-commons-tools/xwiki-commons-tool-enforcers/xwiki-commons-tool-enforcer-dependencies/src/main/java/org/xwiki/tool/enforcer/ExternalExtensionCheck.java#L103)

jungm:
GitHub, generate report

nrmancuso:
@jungm Please do not extend existing test inputs, let’s create new inputs with cases from the issue and new violations found in the regression report.

jungm:
Looking through the regression report I'm quite often seeing this pattern:
```java
  public boolean someFunction() {
    if (condition1) return false;
    if (condition2) return false;
    if (condition3) return false;

    return true;
  }
```

Is this something that should be a violation? Obviously gets detected now but was fine before

nrmancuso:
@jungm Let's consider the check documentation:

> Checks for over-complicated boolean return statements.

Your example can be simplified to:
```java

  public boolean someFunction() {
    return !condition1 && !condition2 && !condition3;
  }
```

I think this aligns with the check's goals of simplification. We would have three violations on your example, correct?

jungm:
@nrmancuso nope, only the last `if` would be a violation as it's directly followed by a `return` statement with a literal boolean, all ifs before that should be passing

Theoretically something like this is also passing right now because the final `return` statement is not a literal boolean, this might be something I can improve on
```java
  public boolean someFunction() {
    if (condition1) return false;
    if (condition2) return false;
    return !condition3;
  }
```

Specifically what I'm worried about is cases [like this](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/5d1dffa_2023083659/reports/diff/elasticsearch/xref/home/runner/work/checkstyle/checkstyle/.ci-temp/contribution/checkstyle-tester/repositories/elasticsearch/src/main/java/org/apache/lucene/queryparser/classic/QueryParserSettings.java.html#L381) that would result in a huge boolean expression

nrmancuso:
>Specifically what I'm worried about is cases [like this](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/5d1dffa_2023083659/reports/diff/elasticsearch/xref/home/runner/work/checkstyle/checkstyle/.ci-temp/contribution/checkstyle-tester/repositories/elasticsearch/src/main/java/org/apache/lucene/queryparser/classic/QueryParserSettings.java.html#L381) that would result in a huge boolean expression


Yes, we cannot evaluate expressions, and the number of conditions in any given expression can be infinite :) I would suggest to create some input like this as a new test case, along with the equivalent code with an `else` branch to see how the check behaves.

Now, we do have a new nuance; would your changes set off a chain of violations that keep surfacing as the violation on the previous `if` is resolved?

romani:
>Looking through the regression report I'm quite often seeing this pattern:

Please update your comment to share several links. Report has too much differences. Better to be explicit. 

People might like aligned `return false` as it makes code easy to read. 
Refactoring like https://github.com/checkstyle/checkstyle/pull/14032#issuecomment-1814402983 will not make code quicker but will make it more hard to read.

But whole idea of check to improve easiness to read. Nobody cares on compact code that is harder to read ( at least this is target for another Check).

jungm:
I've got to admit, I'm a bit confused on where to continue on this right now. What steps should I take? Improve tests? PR in the no-error projects to gather feedback from real users before doing anything else?

@romani, I've added some examples to my comment


Regarding this part:
>People might like aligned return false as it makes code easy to read.
Refactoring like https://github.com/checkstyle/checkstyle/pull/14032#issuecomment-1814402983 will not make code quicker but will make it more hard to read.

This is exactly the concern I have in mind; the pattern I've mentioned earlier is mainly present in (probably autogenerated) `equals` methods where, at least in my opinion, optimizing it into one condition (or trying to fool the check by not returning a literal boolean) does not benefit the code quality.
I can imagine that some property like `ignoreChainedIfs` could be introduced to circumvent this. But I'm really not sure on that one

romani:
Thanks for sharing this comment. Without it we presume that you just need find time to comeback to PR.

@nrmancuso , can you help here ?

nrmancuso:
I would suggest a good next step would be to compare the new behavior of this check to the PMD rule that it was modeled after, as mentioned in https://github.com/checkstyle/checkstyle/pull/14032#issuecomment-1814416539.

romani:
@jungm , please send PRs to xwiki and Equalsverifier that are not inside of "equals" or other methods that in pattern of:
```
  public boolean someFunction() {
    if (condition1) return false;
    if (condition2) return false;
    return !condition3;
  }
```

lets get feedback at least on simple cases.

nrmancuso:
We have lost contact with author, anyone is welcome to reuse changes here to close the issue.

PR #14041, opened 2023-11-18
Issue #13772: IllegalImport with class pattern fails on static import

aayushRedHat:
Related: https://github.com/checkstyle/checkstyle/issues/13772

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/IllegalImportCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/IllegalImportCheck.java
@@ -139,7 +139,7 @@ public class IllegalImportCheck
         illegalPkgs = from.clone();
         illegalPkgsRegexps.clear();
         for (String illegalPkg : illegalPkgs) {
-            illegalPkgsRegexps.add(CommonUtil.createPattern("^" + illegalPkg + "\\..*"));
+            illegalPkgsRegexps.add(CommonUtil.createPattern("^(static|" + illegalPkg + ")\\..*"));
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/IllegalImportCheck.java
```diff
@@ -139,7 +139,7 @@ public final void setIllegalPkgs(String... from) {
         illegalPkgs = from.clone();
         illegalPkgsRegexps.clear();
         for (String illegalPkg : illegalPkgs) {
-            illegalPkgsRegexps.add(CommonUtil.createPattern("^" + illegalPkg + "\\..*"));
+            illegalPkgsRegexps.add(CommonUtil.createPattern("^(static|" + illegalPkg + ")\\..*"));
```
rnveach, line 142: This is not much different from https://github.com/checkstyle/checkstyle/pull/13960#discussion_r1377565209 . This now looks like it can ignore the illegal package definition completely for the word `static`. Even if it can't be valid java to show that issue, this still doesn't seem right.

rnveach (changes requested):

nrmancuso (changes requested):
Few things here:
1. The changes in this PR will have no effect on the behavior of this check unless we literally have a package name that contains `static`
2. Issue was originally for `illegalClasses`, but for some reason the regexp for `illegalPkgs` is changed
3. I have removed approved label, since I do not see any issue with the current behavior of the check when compared to it's documentation.

If we were to do any work related to #13772, it would be to expand documentation to include some examples similar to what is in the issue description to make static import/method checking obvious.

I am blocking this PR until we have a resolution in #13772.

romani:
@nrmancuso, please consider https://github.com/checkstyle/checkstyle/issues/13772#issuecomment-1830673466

nrmancuso:
I am removing my assignment here, we need to resolve discussion in issue and decide on what changes (if any) we need to make.

PR #13964, opened 2023-10-28
Issue #13839: migrate SuppressWarningsHolder to use property macro

relentless-pursuit:
Resolves: #13839

PR #13670, opened 2023-08-30
Issue #13626: Reuse in metadata module scrapper all methods to get data from javadoc and reflection only

stoyanK7:
Resolves #13626

PR #14048, opened 2023-11-19
Issue #13693: migrate header modules to use property macro

romani:
Issue #13693

all extra refactoring was enforced by our verify validation.

setCharset
https://github.com/checkstyle/checkstyle/commit/326e3ae67b8cc2a1dcbcf2356b30fd9a4afa964a#diff-fa56dd39f138d19b44a0aa110150c621c7ba641aead7a198d35db46fdbc767a2R71

setHeaderFile
https://github.com/checkstyle/checkstyle/commit/6f1f7bd3fdcca84404e97964b3e48de6db5024ec#diff-fa56dd39f138d19b44a0aa110150c621c7ba641aead7a198d35db46fdbc767a2R69

setHeader
https://github.com/checkstyle/checkstyle/commit/6f1f7bd3fdcca84404e97964b3e48de6db5024ec#diff-fa56dd39f138d19b44a0aa110150c621c7ba641aead7a198d35db46fdbc767a2R81

### DIFF
--- a/config/checkstyle-non-main-files-suppressions.xml
+++ b/config/checkstyle-non-main-files-suppressions.xml
@@ -252,10 +252,6 @@
          files="src[\\/]xdocs[\\/]checks[\\/]coding[\\/]unusedlocalvariable.xml.template"/>
   <!-- Properties macro suppressions until https://github.com/checkstyle/checkstyle/issues/13693 -->
   <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]checks[\\/]header[\\/]header.xml.template"/>
-  <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]checks[\\/]header[\\/]regexpheader.xml.template"/>
-  <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]checks[\\/]imports[\\/]customimportorder.xml.template"/>
   <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]checks[\\/]javadoc[\\/]atclauseorder.xml.template"/>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/header/AbstractHeaderCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/header/AbstractHeaderCheck.java
@@ -77,7 +77,7 @@ public abstract class AbstractHeaderCheck extends AbstractFileSetCheck
     }
 
     /**
-     * Setter to specify the charset to use when reading the headerFile.
+     * Setter to specify the character encoding to use when reading the headerFile.
      *
      * @param charset the charset name to use for loading the header from a file
      */
@@ -86,7 +86,7 @@ public abstract class AbstractHeaderCheck extends AbstractFileSetCheck
     }
 
     /**
-     * Setter to specify the name of the file containing the required header..
+     * Setter to specify the name of the file containing the required header.
      *
      * @param uri the uri of the header to load.
      * @throws CheckstyleException if fileName is empty.
@@ -147,8 +147,9 @@ public abstract class AbstractHeaderCheck extends AbstractFileSetCheck
     }
 
     /**
-     * Set the header to check against. Individual lines in the header
-     * must be separated by '\n' characters.
+     * Specify the required header specified inline.
+     * Individual header lines must be separated by the string
+     * {@code "\n"}(even on platforms with a different line separator).
      *
      * @param header header content to check against.
      * @throws IllegalArgumentException if the header cannot be interpreted
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/header/HeaderCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/header/HeaderCheck.java
@@ -68,7 +68,7 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * <li>
  * Property {@code header} - Specify the required header specified inline.
  * Individual header lines must be separated by the string {@code "\n"}
- * (even on platforms with a different line separator), see examples below.
+ * (even on platforms with a different line separator).
  * Type is {@code java.lang.String}.
  * Default value is {@code null}.
  * </li>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
@@ -140,6 +140,16 @@ public final class SiteUtil {
     private static final Set<String> FILESET_PROPERTIES =
             getProperties(AbstractFileSetCheck.class);
 
+    /**
+     * Check and property name.
+     */
+    private static final String HEADER_CHECK_HEADER = "HeaderCheck.header";
+
+    /**
+     * Check and property name.
+     */
+    private static final String REGEXP_HEADER_CHECK_HEADER = "RegexpHeaderCheck.header";
+
     /** Set of properties that are undocumented. Those are internal properties. */
     private static final Set<String> UNDOCUMENTED_PROPERTIES = Set.of(
         "SuppressWithNearbyCommentFilter.fileContents",
@@ -151,11 +161,26 @@ public final class SiteUtil {
         // static field (all upper case)
         "SuppressWarningsHolder.aliasList",
         // loads string into memory similar to file
-        "HeaderCheck.header",
-        "RegexpHeaderCheck.header",
+        HEADER_CHECK_HEADER,
+        REGEXP_HEADER_CHECK_HEADER,
         // until https://github.com/checkstyle/checkstyle/issues/13376
         "CustomImportOrderCheck.customImportOrderRules"
     );
+
+    /**
+     * Frequent version.
+     */
+    private static final String VERSION_6_9 = "6.9";
+
+    /**
+     * Frequent version.
+     */
+    private static final String VERSION_5_0 = "5.0";
+
+    /**
+     * Frequent version.
+     */
+    private static final String VERSION_3_2 = "3.2";
 
     /**
      * Frequent version.
@@ -169,11 +194,19 @@ public final class SiteUtil {
     private static final Map<String, String> SINCE_VERSION_FOR_INHERITED_PROPERTY = Map.ofEntries(
         Map.entry("MissingDeprecatedCheck.violateExecutionOnNonTightHtml", V824),
         Map.entry("NonEmptyAtclauseDescriptionCheck.violateExecutionOnNonTightHtml", "8.3"),
+        Map.entry("HeaderCheck.charset", VERSION_5_0),
+        Map.entry("HeaderCheck.fileExtensions", VERSION_6_9),
+        Map.entry("HeaderCheck.headerFile", VERSION_3_2),
+        Map.entry(HEADER_CHECK_HEADER, VERSION_5_0),
+        Map.entry("RegexpHeaderCheck.charset", VERSION_5_0),
+        Map.entry("RegexpHeaderCheck.fileExtensions", VERSION_6_9),
+        Map.entry("RegexpHeaderCheck.headerFile", VERSION_3_2),
+        Map.entry(REGEXP_HEADER_CHECK_HEADER, VERSION_5_0),
         Map.entry("NonEmptyAtclauseDescriptionCheck.javadocTokens", "7.3"),
-        Map.entry("FileTabCharacterCheck.fileExtensions", "5.0"),
+        Map.entry("FileTabCharacterCheck.fileExtensions", VERSION_5_0),
         Map.entry("LineLengthCheck.fileExtensions", V824),
         Map.entry("ParenPadCheck.option", "3.0"),
-        Map.entry("TypecastParenPadCheck.option", "3.2")
+        Map.entry("TypecastParenPadCheck.option", VERSION_3_2)
     );
 
     /** Map of all superclasses properties and their javadocs. */
@@ -803,7 +836,8 @@ public final class SiteUtil {
         final Class<?> fieldClass = getFieldClass(field, propertyName, moduleName, classInstance);
         String result = null;
         if (CHARSET.equals(propertyName)) {
-            result = "the charset property of the parent Checker module";
+            result = "the charset property of the parent"
+                    + " <a href=\"https://checkstyle.org/config.html#Checker\">Checker</a> module";
         }
         else if (classInstance instanceof PropertyCacheFile) {
             result = "null (no cache file)";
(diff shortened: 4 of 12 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/SuppressWarningsHolder.java
```diff
@@ -47,9 +47,10 @@
  * <ul>
  * <li>
  * Property {@code aliasList} - Specify aliases for check names that can be used in code
- * within {@code SuppressWarnings}.
+ * within {@code SuppressWarnings} in a format of comma separated attribute=value entries.
+ * The attribute is the fully qualified name of the Check and value is its alias.
  * Type is {@code java.lang.String[]}.
- * Default value is {@code null}.
+ * Default value is {@code ""}.
```
romani, line 53: Not expected , but I will improve after merge of blocker PR.

rnveach: How is it not expected and how will it be improved?

romani: By doing https://github.com/checkstyle/checkstyle/pull/13964#discussion_r1399822130

rnveach: How will going against the setter improve the default value? There is no default value in the setter.

romani: I have idea to:
If field is not found, default is null for sure. We have scalar types and array. 
For scaral type null is null, for array it will be `{}`

romani: fixed.

romani:
GitHub, generate website

romani:
GitHub, generate website

romani:
Travis is crashing in JVM, restarting usually helps, I sent mail to support to look if it is related to some infra update on their side.

src/xdocs/checks/header/header.xml
```diff
@@ -61,13 +61,8 @@ line 5: ////////////////////////////////////////////////////////////////////
...
               <td><a href="../../property_types.html#String">String</a></td>
-              <td>
-                <code>
-                  the charset property of the parent<a href="../../config.html#Checker">
-                  Checker</a> module
-                </code>
-              </td>
-              <td>5.0</td>
+              <td><code>the charset property of the parent &lt;a href=&quot;https://checkstyle.org/config.html#Checker&quot;&gt;Checker&lt;/a&gt; module</code></td>
+              <td>4.4</td>
```
rnveach, line 65: Version number changes are not expected. Please show proof 5.0 is wrong.

src/xdocs/checks/header/header.xml
```diff
@@ -78,21 +73,17 @@ line 5: ////////////////////////////////////////////////////////////////////
...
-              <td>
-                Specify the required header specified inline. Individual header lines
-                must be separated by the string <code>&quot;\n&quot;</code> (even on platforms with
-                a different line separator), see examples below.
-              </td>
+              <td>Specify the required header specified inline. Individual header lines must be separated by the string <code>"\n"</code>(even on platforms with a different line separator).</td>
               <td><a href="../../property_types.html#String">String</a></td>
               <td><code>null</code></td>
-              <td>5.0</td>
+              <td>4.4</td>
```
rnveach, line 79: Same.

src/xdocs/checks/header/header.xml
```diff
@@ -78,21 +73,17 @@ line 5: ////////////////////////////////////////////////////////////////////
...
             </tr>
             <tr>
               <td>headerFile</td>
               <td>Specify the name of the file containing the required header.</td>
               <td><a href="../../property_types.html#URI">URI</a></td>
               <td><code>null</code></td>
-              <td>3.2</td>
+              <td>4.4</td>
```
rnveach, line 86: Same.

src/xdocs/checks/header/regexpheader.xml
```diff
@@ -100,13 +100,8 @@ line 6: ^\W*$
...
               <td><a href="../../property_types.html#String">String</a></td>
-              <td>
-                <code>
-                  the charset property of the parent<a href="../../config.html#Checker">
-                  Checker</a> module
-                </code>
-              </td>
-              <td>5.0</td>
+              <td><code>the charset property of the parent &lt;a href=&quot;https://checkstyle.org/config.html#Checker&quot;&gt;Checker&lt;/a&gt; module</code></td>
+              <td>4.4</td>
```
rnveach, line 104: Same.

src/xdocs/checks/header/regexpheader.xml
```diff
@@ -133,7 +122,7 @@ line 6: ^\W*$
               <td>Specify the name of the file containing the required header.</td>
               <td><a href="../../property_types.html#URI">URI</a></td>
               <td><code>null</code></td>
-              <td>3.2</td>
+              <td>4.4</td>
```
rnveach, line 125: Same.

src/main/java/com/puppycrawl/tools/checkstyle/checks/header/AbstractHeaderCheck.java
```diff
@@ -77,19 +77,21 @@ protected List<String> getHeaderLines() {
     }
 
     /**
-     * Setter to specify the charset to use when reading the headerFile.
+     * Setter to specify the character encoding to use when reading the headerFile.
      *
      * @param charset the charset name to use for loading the header from a file
+     * @since 4.4
```
rnveach, line 83: Also need commits showing this version number is correct.

src/main/java/com/puppycrawl/tools/checkstyle/checks/header/AbstractHeaderCheck.java
```diff
@@ -77,19 +77,21 @@ protected List<String> getHeaderLines() {
...
         this.charset = createCharset(charset);
     }
 
     /**
-     * Setter to specify the name of the file containing the required header..
+     * Setter to specify the name of the file containing the required header.
      *
      * @param uri the uri of the header to load.
      * @throws CheckstyleException if fileName is empty.
+     * @since 4.4
```
rnveach, line 94: Same.

src/main/java/com/puppycrawl/tools/checkstyle/checks/header/AbstractHeaderCheck.java
```diff
@@ -147,11 +149,13 @@ private static Charset createCharset(String name) {
...
     /**
-     * Set the header to check against. Individual lines in the header
-     * must be separated by '\n' characters.
+     * Specify the required header specified inline.
+     * Individual header lines must be separated by the string
+     * {@code "\n"}(even on platforms with a different line separator).
      *
      * @param header header content to check against.
      * @throws IllegalArgumentException if the header cannot be interpreted
+     * @since 4.4
```
rnveach, line 158: Same.

src/main/java/com/puppycrawl/tools/checkstyle/checks/header/HeaderCheck.java
```diff
@@ -68,7 +68,7 @@
  * <li>
  * Property {@code header} - Specify the required header specified inline.
  * Individual header lines must be separated by the string {@code "\n"}
- * (even on platforms with a different line separator), see examples below.
+ * (even on platforms with a different line separator).
```
rnveach, line 71: This is not expected. Why is this changed?

If it is because Javadoc has no examples, this is still copied to xdoc which does.

rnveach (changes requested):

src/xdocs/checks/header/header.xml
romani: Already in PR description

rnveach: Yes, PR description shows original version is correct. Please ensure you are showing before and after on release commits.

> https://github.com/checkstyle/checkstyle/commit/326e3ae67b8cc2a1dcbcf2356b30fd9a4afa964a#diff-fa56dd39f138d19b44a0aa110150c621c7ba641aead7a198d35db46fdbc767a2R71

https://github.com/checkstyle/checkstyle/blob/326e3ae67b8cc2a1dcbcf2356b30fd9a4afa964a/pom.xml#L5
Version 5.0 .

config/checkstyle-non-main-files-suppressions.xml
```diff
@@ -251,10 +251,6 @@
   <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]checks[\\/]coding[\\/]unusedlocalvariable.xml.template"/>
   <!-- Properties macro suppressions until https://github.com/checkstyle/checkstyle/issues/13693 -->
-  <suppress id="propertiesMacroMustExist"
```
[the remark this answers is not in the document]
romani: fixed

src/xdocs/checks/header/header.xml
romani: all since are reverted. map is updated with values from xdoc

src/xdocs/checks/header/header.xml
romani: all since are reverted. map is updated with values from xdoc

src/xdocs/checks/header/header.xml
romani: all since are reverted. map is updated with values from xdoc

src/xdocs/checks/header/regexpheader.xml
romani: all since are reverted. map is updated with values from xdoc

src/xdocs/checks/header/regexpheader.xml
romani: all since are reverted. map is updated with values from xdoc

src/main/java/com/puppycrawl/tools/checkstyle/checks/header/AbstractHeaderCheck.java
romani: all since are reverted. map is updated with values from xdoc

src/main/java/com/puppycrawl/tools/checkstyle/checks/header/AbstractHeaderCheck.java
romani: all since are reverted. map is updated with values from xdoc

src/main/java/com/puppycrawl/tools/checkstyle/checks/header/AbstractHeaderCheck.java
romani: all since are reverted. map is updated with values from xdoc

romani:
> The point of macros is to document everything in our Java files to be pulled over for XDoc documentation.

Yes, but ....
During implementation phase we too much narrowed our minds to xdoc. And we blindly thought that `@since` is also going to serve for xdoc. But I started such PRs you become confused, as you brain is clear xdoc project, and you brain properly interpret `@since` as tag that defines when method is appeared in particular class - https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#@since .
And here there are tiny nuances:
0) Xdoc needs see only properties in result object (that is combination of multiple layers(abstract/parent classes)) 
1) We did not accounted that CS modules has a lot of abstract classes, definitely our big miss.
2) We started to use javadoc general `@since` and it worked well until we came to our old Checks and all this activity is started, and confusion is growing.

Example why general javadoc `@since` is not going to work for us:
Imagine we have check:
```
/** @since 13.0.0 */
class SomeRuleCheck extends AbstractCheck {
...
/** @since 13.0.0 */
setProp1(){}
/** @since 13.0.0 */
setProp2(){}
...
}  
```
we decided to make refactoring in classes to create some abstract class (user and xdoc should not notice anything):
```
/** @since 13.1.1 */
class SomeAbstractCheck extends AbstractCheck {
...
/** @since 13.1.1 */
setProp1(){}
...
} 

/** @since 13.0.0 */
class SomeRuleCheck extends SomeAbstractCheck {
...
/** @since 13.0.0 */
setProp2(){}
...
}  
```

as you can see setProp1 changed since from 13.0.0 to 13.1.1. from Java (javadoc of class) perspective and engineers - correct, from user perspective  someRuleCheckInstance.setProp1 works on cs version 13.0.0.
WE SHOULD NOT reuse general `@since` for xdoc, it will not works clearly as it hard to serve two purposes in same time.
We need `@xdoc-since` tag that will have a value that will serve single purpose - version when such property is accessible by user of object/instance (NOT a some class).

but here is another problem, it is not easy, I failed multiple times to find proper value even for `@xdoc-since` (`@object-since`), at least do not agree with me on what I propose, so I try to avoid this very complicated process and move forward with workaround we already have, and deal with elimination of such hardcoded map in separate PR when we will have nuances gathered and we can make better decision on what to do.

> I have laid out a few times how the logic should work. If there is still confusion on it, I can give it again.

Example that I showed above clearly shows that it will be error-prone for xdoc.

> One of my goals is to minimize this list based on how since values should be documented.

yes, but lets sacrifice (gambit) this for a bit, finish migration of properties, resolve all such problems with unsynced descriptions and typos and ...... and remove copy-paste in templates to unblock https://github.com/checkstyle/checkstyle/pull/13670 and deal with imperfection of hardcoded list for modules with abstract classes. That we are not going to refactor and it is our legacy. 
Something tells me that after we finish we might encounter more problems, that might change our thinking again.
Main point is to deal with version and only versions in separate PRs, so easy to review and focus on nuances and debate what version is actually correct 4.4 or 5.0.

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -151,12 +161,27 @@ public final class SiteUtil {
...
+    /**
+     * Frequent version.
+     */
+    private static final String V50 = "5.0";
+
+    /**
+     * Frequent version.
+     */
+    private static final String V32 = "3.2";
+
```
nrmancuso, line 184: These are hard to read, please just suppress multiple string literal check for this file or add some underscores like `VERSION_3_2` and `VERSION_8_24`

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -169,11 +194,19 @@ public final class SiteUtil {
     private static final Map<String, String> SINCE_VERSION_FOR_INHERITED_PROPERTY = Map.ofEntries(
         Map.entry("MissingDeprecatedCheck.violateExecutionOnNonTightHtml", V824),
         Map.entry("NonEmptyAtclauseDescriptionCheck.violateExecutionOnNonTightHtml", "8.3"),
+        Map.entry("HeaderCheck.charset", V50),
+        Map.entry("HeaderCheck.fileExtensions", V69),
+        Map.entry("HeaderCheck.headerFile", V32),
+        Map.entry(HEADERCHECK_HEADER, V50),
```
nrmancuso, line 200: Translation of `HeaderCheck` to upper snake case is `HEADER_CHECK`, please do for all.

nrmancuso (changes requested):
We need to move forward with migration to templates, I am OK with some hacks that are contained in SiteUtil/etc, with the understanding that we will address these in the future.

Items:

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
romani: done

src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
romani: done

nrmancuso (approved):
I am good, let's move forward with this migration.

romani:
@rnveach , please approve or recall review to let me proceed with migration. We will come back to version discussion a bit later.

romani:
I don't see new messages, so I consider reassignment back to us as sign to be skipped from review.

We are moving forward, we will be back to versions after migration

PR #14051, opened 2023-11-20
Issue #4845: Removed Checkstyle Dependency For CustomOrderImport

piyush-1234:
Issue: #4845


src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/imports/customimportorder/InputCustomImportOrderThirdPartyAndSpecial2.java
```diff
@@ -39,11 +39,6 @@
 
 import org.apache.commons.collections.*; // violation 'Extra separation in import group before .*'
 
-import com.puppycrawl.tools.checkstyle.checks.imports.AbstractImportControl; // SAME_PACKAGE(6)
-
-
-import com.puppycrawl.tools.checkstyle.checks.imports.AbstractImportRule; // violation 'Extra separation in import group before .*'
```
romani, line 45: we can not simply remove lines, we need to replace them with some another "com" that is available in our classpath or what is better in jdk.

in this particulr case we have to keep `import com.puppycrawl.tools.checkstyle.checks.imports.AbstractImportControl; // SAME_PACKAGE(6)` as it is same import as package of class. it is test case for a reason. Please read Check 
https://checkstyle.org/checks/imports/customimportorder.html#CustomImportOrder to understand how it work and what is matter for it.

romani (changes requested):
item:

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/imports/customimportorder/InputCustomImportOrderThirdPartyAndSpecial2.java
piyush-1234: thanx for providing reference doc

piyush-1234: hi @romani , one question, in above class from line no 42 to 45, thats a same package group import, so in this case both same package import should be as it is, right? as it belong to the same package (6) group, and even if we go to remove this dependency we have replace it with the same kind of dependency of SAME PACKAGE (6) which should be of classspath, and we dont change it with other type of dependency like thirdparty and other, as it is grped as same package (6) import.

romani: > both same package import should be as it is, right? 

yes, no changes should be as it by design, this Check is very special if it use SAME PACKAGE.
All imports lines related to SAME PACKAGE should stay as is.
All others should be updated.

piyush-1234: right. thanx.

romani: All permanent suppression should be moved to https://github.com/checkstyle/checkstyle/blob/11e2b4baee11523db5c15602380775e95fb437ba/config/checkstyle-input-suppressions.xml#L8

romani:
please do not close PRs you can reuse it.
please read https://checkstyle.org/beginning_development.html#Starting_Development on how to manage git branches when changes are required.

Issue #14053, opened 2023-11-21
ArchUnit FreezingArchRule ability to skip line numbers and any other extra changes 


romani:
Based on discussion at https://github.com/checkstyle/checkstyle/pull/14009
It becomes clear that validation of cycles by ArchUnit FreezingArchRule has problems due to fact that base file that we store in our git repo contains line numbers.
Example: https://github.com/checkstyle/checkstyle/blob/76a65e13e6e75b36106f49b7d27b2727d17dc153/config/archunit-store/slices-should-be-free-of-cycles-suppressions#L10

and almost each change in files that referenced in this suppression, results in no problems in ArchUnit validation as it  skipped by design.
According to the docs, it should be:

>On the first run all violations of that rule will be stored as the current state. On consecutive runs only new violations will be reported. By default FreezingArchRule will ignore line numbers, i.e. if a violation is just shifted to a different line, it will still count as previously recorded and will not be reported.

BUT, as we store this file in git, this file is changing, and our CI detects not committed changes and failing.

-------

We need to update execution of test method that triggers validation and update of this file to have post validation update on file to remove or substitute to dummy `0`  value for line number.

If approach above is not going to help, we need to disable this validation as we do not know how to freeze old problems and only update this file if we agree to extend our cycles. We should not update this file at any other case.

Referenced PR is example how to activate back validation and play with extra updates to such file.

romani:
https://github.com/checkstyle/checkstyle/pull/14009/files#diff-a976ae03b947162410f0c24b8a0892a6b4410b110c297dfee26878446137c367R6

This diff is good example that file is just snapshot of our current dependencies of packages. So not only changes in line numbers, in signatures of methods. BUT also new line might appear.

So I propose to:
In test method, before arch execution, copy suppression file to `.ci-temp` folder
Let ArchUnit use file from temp folder.
It will update file in that folder, it test is not failing, nobody will care that such file is changed.
If test is failing, and we agree to extend our suppression with extra cycle, we will instruct contributor to copy file to `config` folder and make change to be part of commit.

nrmancuso:
@romani doesn't only one test in https://github.com/checkstyle/checkstyle/pull/14009/files#diff-a976ae03b947162410f0c24b8a0892a6b4410b110c297dfee26878446137c367R6 depend on a file being checked in? Can we enable the other tests?

>So I propose to...

[...]

PR #14060, opened 2023-11-22
Issue #14053: remove cycles

romani:
Issue #14053

I do not know how to make it work
I just added few empty lines to
https://github.com/checkstyle/checkstyle/blob/e4942b9158f910f1089ada94f78e4a94e65d3f23/src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractFileSetCheck.java#L94
and it test failed.

we have `freeze.refreeze=true` and it regenerate file each time, but it should be `freeze.refreeze=false` , as "true"  should be used in case to allow new vioaltions.

but even I did this, freeze file generation is not stable, I increased number of nodes and file become different, ok, I added few empty lines and some cycles are removed from file, even I did nothing.

I think something is not working for this Rule and freeze file, Freeze might work for other rules, but cycle+freeze is weird.

-------

We might need to create issue to resolve cycles and at the end enable this Rule.
One more idea is to write our own Store Persisting - https://www.archunit.org/userguide/html/000_Index.html#_violation_store
to keep only fact of cycle:
```
✔ ~/java/github/romani/checkstyle [master|⚑ 2] 
$ ag --nofilename -A 5 "Cycle detected:" | grep -vE "Dependencies|Method|Const"
Cycle detected: Slice api -> \
                Slice utils -> \
                Slice api\

Cycle detected: Slice api -> \
                Slice utils -> \
                Slice checks.javadoc -> \
                Slice api\

Cycle detected: Slice api -> \
                Slice utils -> \
                Slice checks.javadoc -> \
                Slice checks.naming -> \
                Slice api\

Cycle detected: Slice api -> \
                Slice utils -> \
                Slice checks.javadoc.utils -> \
                Slice api\

Cycle detected: Slice api -> \
                Slice utils -> \
                Slice checks.naming -> \
                Slice api\

Cycle detected: Slice api -> \
                Slice utils -> \
                Slice xpath -> \
                Slice api\

Cycle detected: Slice checks.javadoc -> \
                Slice checks.naming -> \
                Slice utils -> \
                Slice checks.javadoc\

Cycle detected: Slice checks.javadoc -> \
                Slice utils -> \
                Slice checks.javadoc\

Cycle detected: Slice checks.naming -> \
                Slice utils -> \
                Slice checks.naming\

Cycle detected: Slice utils -> \
                Slice xpath -> \
                Slice utils\
```


romani:
while we have implementations in our api folders, we can not fix:
```
Cycle detected: Slice api -> \
                Slice utils -> \
                Slice ....... -> \
                Slice api\
```

src/test/resources/archunit.properties
nrmancuso, line 1: Is it possible that we are hitting some of the max numbers in this file, and that is why test isn't functioning as expected?

nrmancuso (changes requested):
Item:

src/test/resources/archunit.properties
romani: Maybe, but made multiplied numbers by million, still a problem.

romani:
The biggest problem is facts at https://github.com/checkstyle/checkstyle/pull/14060#issuecomment-1826795578
We can not  resolve cycles until we have actual implementation in our API package. Moving all to interfaces is not easy approach, as it will trigger heated debates.
Issue to arch project will look like "how to activate this in our big project", it would still be good to add it from scratch,  to not involve them in confusing history.

Referenced PR should be we add test class and store in first commit, second commit we add new empty lines and we have some issue. We add new cycle, violation is weird.

But still better to remove and start from scratch. Test is minimalistic.

nrmancuso (approved):
Ok, I really don't agree with the need to do all this extra work before we report some issue to ArchUnit, but let's not waste any more time on this.

Issue #14019, opened 2023-11-11
Resolve pitest survivals in modules


Kevin222004:
# 1) [Pitest-Filters](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-filters-suppressions.xml)

-------

**1) SuppressWithNearbyCommentFilter**

- [ ] NonVoidMethodCallMutator (1)

-------
 
**2) SuppressWithNearbyTextFilter**

some details at https://github.com/checkstyle/checkstyle/issues/12952
 
- [ ] NonVoidMethodCallMutator (1)
- [ ] RemoveConditionalMutator_EQUAL_IF (1)
 
--------

**3) XpathFilterElement**

- [ ] ArgumentPropagationMutator (1)
- [ ] NonVoidMethodCallMutator (1)
- [ ] NakedReceiverMutator (1)

-----------

**4) SuppressionXpathSingleFilter** 
 
- [ ] MemberVariableMutator (1) https://github.com/checkstyle/checkstyle/pull/13794
- [ ] MemberVariableMutator (1) should be fixed similar as above
- [ ] MemberVariableMutator (1) should be fixed similar as above

-------------

--------------------------------------

# MemberVariableMutator all are delayed until hcoles/pitest#1230

# 4) [Pitest-Filters](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-filters-suppressions.xml)

-------

**1) SuppressFilterElement**

- [ ] MemberVariableMutator (10)     `10 mutations delayed until hcoles/pitest#1230`

----------

**2 SuppressWithNearbyCommentFilter$Tag**

- [ ] MemberVariableMutator (2)  `2 mutations delayed until hcoles/pitest#1230`
 
--------

**3) SuppressWithPlainTextCommentFilter$Suppression**
 
- [ ] MemberVariableMutator (2) `2 mutations delayed until hcoles/pitest#1230`

--------

suniti0804:
I worked on ScopeUtil#getScopeFromMods

piyush-1234:
i am on ModuleReflectionUtil#getCheckstyleModules

Zopsss:
I'm working on  `NonVoidMethodCallMutator on this.severity = SeverityLevel.getInstance(severity);` &  `MemberVariableMutator on this.severity = SeverityLevel.getInstance(severity);` of `Checker`

mahfouz72:
[...]

Issue #13142, opened 2023-06-03
Bad Side effects of Pitest


Kevin222004:
This issue is been created to make a list of some bad side effects of Pitest on the codebase.

# 0 All caching of operations

Surviving pitest mutations for conditions reducing time complexity at PackageObjectFactory - https://github.com/checkstyle/checkstyle/issues/11902

# 1 creation of mutable collection consider as can be avoided 
 https://github.com/checkstyle/checkstyle/pull/13127 
and https://github.com/checkstyle/checkstyle/pull/13126  
```
        public Set<DetailAST> getMethodCalls() {
            return Collections.unmodifiableSet(methodCalls);
        }
```
this code has been mutated as 
```
        public Set<DetailAST> getMethodCalls() {
            return methodCalls;
        }
```
Modifying the code with Pitest from `return Collections.unmodifiableSet(methodCalls)` to `return new HashSet<>(methodCalls)` eliminates the guarantee of returning an unmodifiable set, potentially leading to unintended modifications and inconsistent behavior.

but in same time if method is API, it is good to add pure unit test to kill mutation - [Example](https://github.com/checkstyle/checkstyle/pull/13156#pullrequestreview-1461216335).
-------------------------------
 
# 2 does not allow performance optimization in logic 

https://github.com/checkstyle/checkstyle/pull/13557 RequireEmptyLineBeforeBlockTagGroup
https://github.com/checkstyle/checkstyle/pull/13120 VariableDeclarationUsageDistanceCheck
https://github.com/checkstyle/checkstyle/pull/13149  UnusedLocalVariableCheck
https://github.com/checkstyle/checkstyle/pull/13196   MagicNumberCheck
https://github.com/checkstyle/checkstyle/pull/13416 BlockCommentPosition
https://github.com/checkstyle/checkstyle/pull/13430  CheckUtil
```
        while (result
                && !isUsedVariableDeclarationFound
                && currentSiblingAst != null) {
```
mutated as 
```
        while (!isUsedVariableDeclarationFound
                && currentSiblingAst != null) {
```
by the removal of the result, the output will not affect.
but the whole loop can run for more time than it was running with the result. This type of change is not good as it might lead to a problem.

------

It forcing to write less explicit code even it is redundant. So user readability is tiny decreasing sometimes.
Example: https://github.com/checkstyle/checkstyle/pull/13381#issuecomment-1627715592

-----------

# 3. Extra code that does help to make code more robust to future changes that might miss extended testing.

[...]

PR #14061, opened 2023-11-23
Issue #14019: Resolve pitest utils suppressions for ScopeUtil#getScopeFromMods

suniti0804:
Issue #14019 

### Explanation

Mutation - 

https://github.com/checkstyle/checkstyle/blob/11e2b4baee11523db5c15602380775e95fb437ba/config/pitest-suppressions/pitest-utils-suppressions.xml#L39-L46

### Dependency Tree

```
Method
    getScopeFromMods(DetailAST)
Usages or usages of base method in Project Files
  Unclassified
    checkstyle
      com.puppycrawl.tools.checkstyle.checks.coding
        DeclarationOrderCheck
          processModifiersSubState(DetailAST, ScopeState, boolean)
          final Scope access = ScopeUtil.getScopeFromMods(modifiersAst);
```

### Modules 

1. `DeclarationOrderCheck`

Diff Regression config: https://gist.githubusercontent.com/suniti0804/35fc6bb1108b1e7ddb7c5f13471510b8/raw/fffe9dbf99d906a8e1855e38dcbf64c57fca430d/pull-14061-regerssion-config

Diff Regression projects: https://gist.githubusercontent.com/suniti0804/1d7bd7fbc50bf6c93896fbe2bbbf2c4a/raw/f71611ef97d0737b6d8bb1c1253cab18b9342429/projects-to-test-on-for-github-action.properties

Report label: Issue#14061-Report

### Analysis for ScopeUtil#getScopeFromMods Method Modification

The original code was: `.orElseGet(() -> getDefaultScope(aMods.getParent()));`
The mutated code is: `.orElseGet(() -> getDefaultScope(aMods));`

In the original code, `getDefaultScope()` was being called with the parent of `aMods` as the argument. In the mutated code, `getDefaultScope()` is being called with `aMods` itself.

**Justification for the Change:**

1. All unit tests are passing after the change, which means that the modification has not broken any existing functionality.

2. The `getDefaultScope()` method determines the default scope based on the type of the AST node. It does not require a parent node specifically, and there are no operations in the method that would fail if a non-parent node were passed. Therefore, it is safe to pass `aMods` instead of `aMods.getParent()`.

3. The `getDefaultScope()` method checks if the `DetailAST` object is in an enum block, an interface or annotation block, or none of these. It does not rely on the hierarchical relationship of the nodes. Therefore, replacing `aMods.getParent()` with `aMods` does not alter the logic or the context of the application.

4. There doesn't seem to be any side-effects of this change as the `getDefaultScope()` method only checks for the type of the node and doesn't depend on the node's parent-child relationship.

The mutation doesn't seem to break the functionality because `getDefaultScope()` doesn't require a parent node as an argument. It only needs a node of an appropriate type, which `aMods` is.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
@@ -85,7 +85,7 @@ public final class ScopeUtil {
      */
     public static Scope getScopeFromMods(DetailAST aMods) {
         return Optional.ofNullable(getDeclaredScopeFromMods(aMods))
-                .orElseGet(() -> getDefaultScope(aMods.getParent()));
+                .orElseGet(() -> getDefaultScope(aMods));
     }
 
     /**
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -6333,7 +6333,7 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java</fileName>
     <specifier>introduce.eliminate</specifier>
     <message>It is bad style to create an Optional just to chain methods to get a value.</message>
-    <lineContent>.orElseGet(() -&gt; getDefaultScope(aMods.getParent()));</lineContent>
+    <lineContent>.orElseGet(() -&gt; getDefaultScope(aMods));</lineContent>
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
--- a/config/pitest-suppressions/pitest-utils-suppressions.xml
+++ b/config/pitest-suppressions/pitest-utils-suppressions.xml
@@ -26,13 +26,4 @@
     <description>removed conditional - replaced equality check with true</description>
     <lineContent>token != null &amp;&amp; !returnValue;</lineContent>
   </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>ScopeUtil.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.utils.ScopeUtil</mutatedClass>
-    <mutatedMethod>lambda$getScopeFromMods$1</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getParent with receiver</description>
-    <lineContent>.orElseGet(() -&gt; getDefaultScope(aMods.getParent()));</lineContent>
-  </mutation>
 </suppressedMutations>


suniti0804:
Github, generate report

romani:
>In conclusion, it seems safe to remove this condition based on the successful passing of all tests and the lack of any adverse effects on the rest of the code.

unfortunately it is not enough as we do not have in our Inputs all possible forms of java code. Regression report is required.
Please do also analytical analysis on why this is extra operation.
review https://github.com/checkstyle/checkstyle/issues/13142 and all referenced PR to see how contributors shared analysis of code to share reason why it is actually extra operation.

romani:
Github, generate report

romani (changes requested):
please do analysis https://github.com/checkstyle/checkstyle/pull/14061#issuecomment-1824541504

romani:
report is good.
I am ok to merge.

@suniti0804 , please share analytical based on code (not a report or test execution results) prove that removal is ok, to make sure you understand what is going on.

romani (approved):

src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
```diff
@@ -85,7 +85,7 @@ public static Scope getScope(DetailAST ast) {
      */
     public static Scope getScopeFromMods(DetailAST aMods) {
         return Optional.ofNullable(getDeclaredScopeFromMods(aMods))
-                .orElseGet(() -> getDefaultScope(aMods.getParent()));
```
nrmancuso, line 88: We need an actual analysis of why this change is OK.

nrmancuso (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
suniti0804: done, please check the PR description

nrmancuso:
@suniti0804 your analysis needs some further consideration:

> In the original code, getDefaultScope() was being called with the parent of aMods as the argument. In the mutated code, getDefaultScope() is being called with aMods itself.

Yes, we can see this from the diff in this PR.

>Justification for the Change:

>All unit tests are passing after the change, which means that the modification has not broken any existing functionality.

What does this prove? How is this different than a surviving mutation?

>The getDefaultScope() method determines the default scope based on the type of the AST node. It does not require a parent node specifically, and there are no operations in the method that would fail if a non-parent node were passed. Therefore, it is safe to pass aMods instead of aMods.getParent().

How do you know this? What is a parent node? What is a non-parent node?

>The getDefaultScope() method checks if the DetailAST object is in an enum block, an interface or annotation block, or none of these. It does not rely on the hierarchical relationship of the nodes. Therefore, replacing aMods.getParent() with aMods does not alter the logic or the context of the application.

Why doesn't this method rely on the relationship of the two different nodes?

>There doesn't seem to be any side-effects of this change as the getDefaultScope() method only checks for the type of the node and doesn't depend on the node's parent-child relationship.

Again, why? "There doesn't seem to be any side-effects of this change ..." is not an analysis.

>The mutation doesn't seem to break the functionality because getDefaultScope() doesn't require a parent node as an argument. It only needs a node of an appropriate type, which aMods is.

What type is "appropriate"?

nrmancuso:
@suniti0804 ping, let's get this PR across the finish line before we open more.

suniti0804:
sure @nrmancuso , let me try - 

1. **What does this prove? How is this different than a surviving mutation?**

Passing unit tests ensure that the existing functionality remains intact after the modification. A surviving mutation is a mutation that doesn't cause the tests to fail, which could indicate an inadequacy in the tests rather than the code's correctness. In this case, the mutation was intentional, and the passing tests indicate that the mutation didn't introduce a regression.

2. **How do you know this? What is a parent node? What is a non-parent node?**

In an Abstract Syntax Tree (AST), a parent node is one that has child nodes under it. A non-parent node does not have any child nodes. I've studied the getDefaultScope method's implementation, and it doesn't contain any operations that specifically require a parent node. It operates based on the type of the node, not its position in the tree, hence it is safe to pass aMods instead of aMods.getParent().

3. **Why doesn't this method rely on the relationship of the two different nodes?**

The getDefaultScope() method operates by checking the type of the DetailAST object itself, not its relationship to other nodes. The method checks whether the object is in an enum block, an interface or annotation block, or none of these. It doesn't check or rely on the node's position within the tree.

4. **Again, why? "There doesn't seem to be any side-effects of this change ..." is not an analysis.**

I've carefully analyzed the getDefaultScope method, and it doesn't have any operations that depend on a node's relationship to other nodes. Therefore, changing the node that's passed to it shouldn't have side effects as long as the node is of an appropriate type.

5. **What type is "appropriate"?**

An 'appropriate' type is one that corresponds to one of the types that getDefaultScope checks for: enum block, interface or annotation block, or none of these. Since aMods is a DetailAST object, it's an appropriate type to pass to the getDefaultScope method.

nrmancuso:
We are still struggling to understand why this change is ok.

>The getDefaultScope() method operates by checking the type of the DetailAST object itself, not its relationship to other nodes.

>I've carefully analyzed the getDefaultScope method, and it doesn't have any operations that depend on a node's relationship to other nodes.

We are under the  false impression that node's relationships do not matter when it seems like this is the only thing that matters. We also missed analysis of the actual AST. This is my take on what is happening here:

See https://github.com/checkstyle/checkstyle/blob/5eb384ef18e5b3cf7c4f46e733437cf7e4df46d3/src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java#L104

In plain words, this is what ScopeUtil#getDefaultScope is doing:
1. Recurse up through all parents of the given AST to see if one is an enum definitition
    a. if so, check the type of this AST and assign a scope
 2. Recurse up through all parents of the given AST to see if we are in an interface definition
    a. if so, assign public scope
3. Otherwise, assign package scope

So, we have three cases: we are in an enum, we are in an interface, and everything else.

Take this AST:

```
➜  src cat Test.java
public class Test {
    interface I {
        String S = "s";
    }
}

➜  src java -jar checkstyle-10.12.1-all.jar -t Test.java
COMPILATION_UNIT -> COMPILATION_UNIT [1:0]
`--CLASS_DEF -> CLASS_DEF [1:0]
    |--MODIFIERS -> MODIFIERS [1:0]
    |   `--LITERAL_PUBLIC -> public [1:0]
    |--LITERAL_CLASS -> class [1:7]
    |--IDENT -> Test [1:13]
    `--OBJBLOCK -> OBJBLOCK [1:18]
        |--LCURLY -> { [1:18]
        |--INTERFACE_DEF -> INTERFACE_DEF [2:4]
        |   |--MODIFIERS -> MODIFIERS [2:4]
        |   |--LITERAL_INTERFACE -> interface [2:4]
        |   |--IDENT -> I [2:14]
        |   `--OBJBLOCK -> OBJBLOCK [2:16]
        |       |--LCURLY -> { [2:16]
        |       |--VARIABLE_DEF -> VARIABLE_DEF [3:8] <-- currently we pass in this node
        |       |   |--MODIFIERS -> MODIFIERS [3:8]  <-- changes in this PR pass in this node
        |       |   |--TYPE -> TYPE [3:8]
        |       |   |   `--IDENT -> String [3:8]
        |       |   |--IDENT -> S [3:15]
        |       |   |--ASSIGN -> = [3:17]
        |       |   |   `--EXPR -> EXPR [3:19]
        |       |   |       `--STRING_LITERAL -> "s" [3:19]
        |       |   `--SEMI -> ; [3:22]
        |       `--RCURLY -> } [4:4]
        `--RCURLY -> } [5:0]
```

If we are recursing up through all parents, does it matter if pass in `MODIFIERS` instead of `VARIABLE_DEF`? It shouldn't. So this mutation seems like another case of pitest finding our optimizations.

nrmancuso (approved):
Ok to merge when CI passes, I am going to rebase this.

PR #14069, opened 2023-11-27
infra: remove concurrency configuration for diff report action

nrmancuso:
Tired of messing with concurrency configuration, seems like Github changed some behavior on their side again. We cannot block diff report generation.


rnveach:
````
The following Github workflows are missing a concurrency block:
.github/workflows/diff-report.yml

Exited with code exit status 1
````

rnveach:
Just pointing out CI is still failing as mentioned above.

Issue #14084, opened 2023-12-01
Resolve CheckerFramework violations


romani:
CheckerFramework was applied  to project in scope of https://github.com/checkstyle/checkstyle/issues/12164

in scope of this issue we need to fix violations that are reasonable. 
Here is suppression list https://github.com/checkstyle/checkstyle/tree/master/config/checker-framework-suppressions

We need each remove suppression xml block, run corresponding checker scripts and resolve violation kin main code.
It is highly recommended to use one PR for one violation as fix might be controvercial.

vivek-0509:
Hi, @romani 

While working on this  issue ,i found that adding the annotation `@IntRange(from=2, to=36)` to the radix parameter in CheckUtil.java can fix 8 radix validation violations. However, the utils package's import-control disallows this import.

Could we add a  exception for `@IntRange` in the utils subpackage? For example:

```<subpackage name="utils"> <allow class="org.checkerframework.checker.index.qual.IntRange"/> </subpackage> ```

Alternatively, if the utils package is intentionally kept free of external framework dependencies (to keep it a pure utility layer with minimal coupling), I understand keeping these as suppressions.

Please let me know which approach is preferred.

Thanks!

romani:
we did not find a good value from `IntRange` , I think we need to disable rules that produces this completely.
May be in future we will change out mind, but for now we are not mature enough to handle them, and simply annotating whole code wit such annotations is also not good approach.

Please find way to disable such rules completely.

vivek-0509:
> we did not find a good value from `IntRange` , I think we need to disable rules that produces this completely. May be in future we will change out mind, but for now we are not mature enough to handle them, and simply annotating whole code wit such annotations is also not good approach.
> 
> Please find way to disable such rules completely.

Understood. Before proceeding, I wanted to clarify one point. Since the Checker Framework does not allow disabling @IntRange or any other validation independently, all of these checks are bundled together under IndexChecker. Currently, the project does not use any of the IndexChecker annotations, and all reported violations are already suppressed in suppression file, which means IndexChecker runs without providing any real benefit while still adding maintenance overhead and a small build time cost. Because of this, a clean removal of IndexChecker would simplify the setup without requiring any code changes. Should I go ahead and remove IndexChecker?

romani:
> Should I go ahead and remove IndexChecker?

yes, we need to reduce volume of violations/suppressions and we can always come back to this Checker after we cleanup code for less complicated violations.

[...]

PR #14085, opened 2023-12-01
Issue #14084: fix Checker violation for nullable return

romani:
Issue #14084

https://stackoverflow.com/a/35875417

### DIFF
--- a/config/import-control.xml
+++ b/config/import-control.xml
@@ -19,6 +19,7 @@
   <allow pkg="java.io"/>
   <allow pkg="java.net"/>
   <allow pkg="java.util"/>
+  <allow class="javax.annotation.Nullable"/>
   <allow pkg="javax.xml.parsers"/>
   <allow pkg="org.apache.commons.beanutils"/>
   <allow pkg="org.apache.commons.logging"/>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/LeftCurlyCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/LeftCurlyCheck.java
@@ -21,6 +21,8 @@ package com.puppycrawl.tools.checkstyle.checks.blocks;
 
 import java.util.Locale;
 
+import javax.annotation.Nullable;
+
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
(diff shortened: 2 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/LeftCurlyCheck.java
```diff
@@ -307,7 +309,7 @@ private static DetailAST getBraceFromSwitchMember(DetailAST ast) {
      * @return {@code DetailAST} if the first child is {@code TokenTypes.SLIST},
      *     {@code null} otherwise.
      */
-    private static DetailAST getBraceAsFirstChild(DetailAST ast) {
+    private static @Nullable DetailAST getBraceAsFirstChild(DetailAST ast) {
```
nrmancuso, line 312: Please make annotation above method definition. We should also make a check module for this. As we grow in annotation usage, we will not want to inline all of them.

nrmancuso (changes requested):
Item:

nrmancuso (approved):
Awesome!

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/LeftCurlyCheck.java
nrmancuso: https://github.com/checkstyle/checkstyle/issues/14101#issue-2026222144

nrmancuso: @romani assigned you for approval ^

romani: approved, I surprised that it was not done from the beginning.

config/import-control.xml
```diff
@@ -20,6 +20,7 @@
   <allow pkg="java.net"/>
   <allow pkg="java.util"/>
   <allow pkg="javax.xml.parsers"/>
+  <allow class="javax.annotation.Nullable"/>
```
rnveach, line 23: Move up one line so it is alphabetical.

rnveach (approved):

config/import-control.xml
romani: done

rnveach (approved):

Issue #14129, opened 2023-12-11
File name too long error


rickie:
I use Linux Ubuntu 20.04.1 on my machine and my file system is encrypted with eCryptfs. As a result, there is a maximum allowed filename size for folders and files. For context, see [this StackOverflow question](https://unix.stackexchange.com/questions/32795/what-is-the-maximum-allowed-filename-and-folder-size-with-ecryptfs).

If I do a `mvn clean install` on Checkstyle `master`, I end up with the following error:

> [ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/rick/repos/oss/checkstyle/src/test/resources/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierNestedClassInPublicInterfaceRedundantModifiers.java:[14,13] error while writing com.puppycrawl.tools.checkstyle.checks.modifier.redundantmodifier.InputRedundantModifierNestedClassInPublicInterfaceRedundantModifiers.PublicInnerInterface.PublicInnerInnerInterface.PublicInnerClassInNestedPublicInterfaces: /home/rick/repos/oss/checkstyle/target/test-classes/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierNestedClassInPublicInterfaceRedundantModifiers$PublicInnerInterface$PublicInnerInnerInterface$PublicInnerClassInNestedPublicInterfaces.class: File name too long

The problem is that the names of all the inner classes and interfaces are chained together with a `$` in between for the file names. This results in fairly long file names as can be seen in the example.

I realize this is quite an uncommon case and asking the following feels like a long shot, but here is my question. 
Would you be open to fixing this problem? If yes, I'm open to creating a PR for it based on your decision on how to best fix this. The simple way would of course be renaming some of the classes and interfaces, to prevent this problem from happening. However, I can see how this is not an ideal way of fixing this.

<details>
<summary>To give you a potential solution, see this diff.</summary>

 ```diff
diff --git a/src/test/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheckTest.java b/src/test/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheckTest.java
index af4601e7f..c7a01f7c6 100644
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheckTest.java
[...]

PR #12063, opened 2022-08-15
Issue #2160: Activate error-prone

Vyom-Yadav:
Resolves #2160

Enabling error-prone for the project, all `.*[\\/]resources[\\/].*` are excluded from error-prone evaluation.
https://errorprone.info/docs/flags

---

**Proof of Working:**

- `New surviving error(s) found:` https://github.com/checkstyle/checkstyle/runs/7965373088?check_suite_focus=true
- `Unnecessary suppressed error(s) found and should be removed:` https://github.com/checkstyle/checkstyle/runs/7965413923?check_suite_focus=true
- Both `New surviving error(s) found:` and `Unnecessary suppressed error(s) found and should be removed:` https://github.com/checkstyle/checkstyle/runs/7965481307?check_suite_focus=true

---

### Reason for adding `jvm.config` file:
`--add-exports` and `--add-opens` is required on JDK 16 and newer due to [JEP 396: Strongly Encapsulate JDK Internals by Default](https://openjdk.java.net/jeps/396)

PR #14136, opened 2023-12-12
Issue #14137: Introduce Error Prone Support

rickie:
This PR is based on a suggestion by @romani in: https://github.com/checkstyle/checkstyle/issues/14129. The idea is to add Error Prone Support to Checkstyle in this PR and after that discuss which checks we want to enable.

I added Error Prone Support as part of the already existing Maven profile that runs Error Prone. The reason for this is that Error Prone Support is an extension and "just adds some checks". Configuration wise, this is the simplest. The only difference the user could notice is that the documentation URL from the Error Prone Support checks are slightly different. See an example below.

Error from an Error Prone check from Google:
> Error:  /home/runner/work/checkstyle/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocNodeImpl.java:[183,51] [ArrayHashCode] hashcode method on array does not hash array contents
    (see https://errorprone.info/bugpattern/ArrayHashCode)

Error from an Error Prone check from Error Prone Support:
> Error:  /home/runner/work/checkstyle/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/checks/TranslationCheck.java:[257,88] [CollectorMutability] Avoid `Collectors.to{List,Map,Set}` in favor of collectors that emphasize (im)mutability
    (see https://error-prone.picnic.tech/bugpatterns/CollectorMutability)

Checkstyle is configured to only consider checks that have `SeverityLevel.ERROR`, see [here](https://github.com/checkstyle/checkstyle/blob/master/.ci/error-prone-check.groovy#L89). In Error Prone Support there are 3 checks with error severity. So adding the tool itself requires changes only in the `pom.xml`.
Now we can simply increase the severity of Error Prone Support checks as follows:
```
-Xplugin:ErrorProne -Xep:CollectorMutability:ERROR
```

In the follow up issue (https://github.com/checkstyle/checkstyle/issues/14137) we can discuss which checks we want to enable. I'm definitely open to creating PRs for that!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -233,8 +233,9 @@
     <junit.version>5.10.1</junit.version>
     <forbiddenapis.version>3.6</forbiddenapis.version>
     <json-schema-validator.version>1.2.0</json-schema-validator.version>
-    <error-prone.version>2.23.0</error-prone.version>
     <checkerframework.version>3.42.0</checkerframework.version>
+    <error-prone.version>2.23.0</error-prone.version>
+    <error-prone-support.version>0.14.0</error-prone-support.version>
     <doxia.version>1.12.0</doxia.version>
   </properties>
 
@@ -2385,6 +2386,11 @@
                       <artifactId>error_prone_core</artifactId>
                       <version>${error-prone.version}</version>
                     </path>
+                    <path>
+                      <groupId>tech.picnic.error-prone-support</groupId>
+                      <artifactId>error-prone-contrib</artifactId>
+                      <version>${error-prone-support.version}</version>
+                    </path>
                   </annotationProcessorPaths>
                 </configuration>
               </execution>
@@ -2438,6 +2444,11 @@
                       <artifactId>error_prone_core</artifactId>
                       <version>${error-prone.version}</version>
                     </path>
+                    <path>
+                      <groupId>tech.picnic.error-prone-support</groupId>
+                      <artifactId>error-prone-contrib</artifactId>
+                      <version>${error-prone-support.version}</version>
+                    </path>b
                   </annotationProcessorPaths>
                 </configuration>
               </execution>


romani:
Example of error level rule https://error-prone.picnic.tech/bugpatterns/FluxFlatMapUsage/


Please share where `-Xplugin:ErrorProne -Xep:CollectorMutability:ERROR` will be placed to change default value of severity for this rule/pattern.

Please use in commit message prefix `Issue #14137: .......` we keep wide explanations linked to commit to easily restore later on.

As already noticed we use hacky groovy script to put on leash reporting of errorprone. As you very knowledgeable in this tool, do you know better way to get report from tool? We just need print to contributor not all messages but messages that requires to fix.

Is there ability to exclude pattern/rule from execution at all? Or change severity to IGNORE/OFF.

romani (approved):
Ok to merge.
Next PRs will be more interesting.

romani:
@Vyom-Yadav , please read this as you were involved in error prone activation in our project

rnveach:
https://github.com/checkstyle/checkstyle/actions/runs/7187166866/job/19574155042?pr=14136
Log file jumped from 100 kb to 2.5 megs in this PR.

rickie:
Thanks for all the questions, let me know if something is unclear or I'm missing something. 

> Please share where `-Xplugin:ErrorProne -Xep:CollectorMutability:ERROR` will be placed to change default value of severity for this rule/pattern.

After the `-Xplugin:ErrorProne` we have to add exta configuration ([here](https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L2379)) if we want to. Now that I'm writing down, I  we need to have a place in the `pom.xml` where we define this property as we will use it in two places, namely in the [error-prone-compile](https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L2346) and [error-prone-test-compile](https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L2398) profiles. Do we define a property and use it in both profiles?

W.r.t. how we can best configure the checks, I'd like to propose to use a similar approach to what we have [here](https://github.com/PicnicSupermarket/error-prone-support/blob/master/pom.xml#L1727-L1761), which allows us to specifically disable and increase the severity of checks.

> As already noticed we use hacky groovy script to put on leash reporting of errorprone. As you very knowledgeable in this tool, do you know better way to get report from tool? We just need print to contributor not all messages but messages that requires to fix.

Currently there is nothing yet that gives us the reports easily. There is a GitHub issue related to this: https://github.com/google/error-prone/issues/3766, that's all...

> Is there ability to exclude pattern/rule from execution at all? Or change severity to IGNORE/OFF.

We can use the command-line flags to enable and disable checks, see the docs [here](https://errorprone.info/docs/flags).
So we simply disable rules by adding `-Xep:NameOfTheCheck:OFF`, or use one of their convenience flags like `-XepDisableAllWarnings`. 

> As far as I remember we focused on fixing ERROR only as we did not able to find way to on/off rules or certain violation. To enable all and iteratively and slowly clean up backlog.

I tried searching fo a discussion related to this in the old [Error Prone activation PR](https://github.com/checkstyle/checkstyle/pull/12063#discussion_r950340236) but I couldn't find any. We can iteratively enable all of the rules, I don't see blockers (yet?). If you know what was blocking you from doing this, let me know, I'm curious!

> Our suppression model: [master/config/error-prone-suppressions/compile-phase-suppressions.xml](https://github.com/checkstyle/checkstyle/blob/master/config/error-prone-suppressions/compile-phase-suppressions.xml?rgh-link-date=2023-12-13T00%3A46%3A09Z)

Thanks for sharing. Is there a specific reason for not using `@SuppressWarnings("ArrayHashCode" /* Explanation why we suppress this check. */)` for the suppression? 

Oh: I think I found the reason in the old Error Prone PR: https://github.com/checkstyle/checkstyle/pull/12063#discussion_r950340236, am I correct?

> [checkstyle/checkstyle/actions/runs/7187166866/job/19574155042?pr=14136](https://github.com/checkstyle/checkstyle/actions/runs/7187166866/job/19574155042?pr=14136)
Log file jumped from 100 kb to 2.5 megs in this PR.

Good point, I noticed a few of our checks are [creating quite some output](https://github.com/checkstyle/checkstyle/actions/runs/7187166866/job/19574155042?pr=14136). What we can do is already disable some of the rules like `StaticImport`, `LexicographicalAnnotationListing`, and `JUnitMethodDeclaration`. I suggest we disable them in follow up PRs.

rickie:
Rebased.

FYI: The build is failing because the linked issue is not approved.

romani:
>Do we define a property and use it in both profiles?

Yes. We can define properties in pom xml, we already have bunch.

>propose to use a similar approach to what we have [here](https://github.com/PicnicSupermarket/error-prone-support/blob/master/pom.xml#L1727-L1761),

Works for us too.

>So we simply disable rules by adding -Xep:NameOfTheCheck:OFF

Yes, this is good for us, as we have rule that if we enforce some rule we do it as build failure, no leaks are allowed.

>I don't see blockers (yet?). If you know what was blocking you from doing this, let me know, I'm curious!

No blockers, but we would need your help to make it, we are sinking in PRs, maintainers are overloaded.

>Is there a specific reason for not using @SuppressWarnings("ArrayHashCode"

We keep culture of "no pollution in main code for tools". So we keep main code clean and all hacks and unfortunate cases outside, let let readers do not be bothered by extra details. If we put all to code it will have a lot of annotations :). But sometimes we need to suppression on specific line, not whole method.

>I suggest we disable them in follow up PRs.

Yes, please. I updated issue with opinion on what should apply and what we should disable.

romani:
@Vyom-Yadav , fyi

rickie:
> Yes. We can define properties in pom xml, we already have bunch.

Nice sounds good, will do that as part of the follow up PRs :).

> we would need your help to make it

Sounds good, I'll try to keep it doable and not overload with PRs. After this one is merged I'll open the first one(s).

> We keep culture of "no pollution in main code for tools". 

Interesting, cool approach! Thanks for explaining.

> Yes, please. I updated issue with opinion on what should apply and what we should disable.

Yes perfect. Will start creating them after we merge this PR. Oh and if it goes too fast or too slow after we merge this one, just let me know and we will work it out :).

rickie:
Rebased and resolved conflict.

nrmancuso (approved):
@rickie thanks a lot for taking this up!

Vyom-Yadav:
Sorry, Late to the party, but this is really cool, thanks a lot, @rickie for taking charge here.

> Checkstyle is configured to only consider checks that have SeverityLevel.ERROR

Just curious, do you, in your org, consider the warnings as well? Are there any good checks that give out warnings but are good to have as errors? That way we can further improve the quality.

romani:
We usually apply/enforce all rules, if we do not see some rule doing any good, of it is against our style or .., we just disable it at all.
There should be no violations of any severity.

rickie:
> Sorry, Late to the party, but this is really cool, thanks a lot, @rickie for taking charge here.

No problem, I really like doing this @Vyom-Yadav.

> Just curious, do you, in your org, consider the warnings as well? Are there any good checks that give out warnings but are good to have as errors? That way we can further improve the quality.

Yes we do! I think we have quite the same approach & goals as Checkstyle. Namely, try to enable as much static analysis as possible. For our Error Prone usage, it means that we use the following [command-line flags](https://errorprone.info/docs/flags): `-XepAllDisabledChecksAsWarnings` and `-XepAllSuggestionsAsWarnings` (fun fact: the last flag we contributed ourselves to Error Prone). These flags combined enable all checks that Error Prone provides, see the full list of checks here: https://errorprone.info/bugpatterns. After enabling all of them with these flags, we selectively disable a few of the checks.

If you want to see an example, here is our configuration from Error Prone Support: https://github.com/PicnicSupermarket/error-prone-support/blob/master/pom.xml#L1701-L1757.

This is roughly similar to what we have in our internal setup at Picnic. 

I'd say after we went through the checks of Error Prone Support as defined in #14137, we can take a look at the checks from Error Prone that are listed under "WARNING" and "INFO" https://errorprone.info/bugpatterns. WDYT?

romani:
Yes, please, let's try to activate all. It helps a lot to maintenance of project by small team.

Vyom-Yadav:
Good suggestion. As we have default `ERROR` in our [script](https://github.com/checkstyle/checkstyle/blob/master/.ci/error-prone-check.groovy#L89), we can enable all, look at the logs to find some value checks from the warning which we can later change to `ERROR` severity level.

Issue #14120, opened 2023-12-08
Make Checkstyle an (automatic) module


future2r:
Checkstyle 10.x requires JDK 11 to run. Wouldn't it be the right time to finally make it a Java module (introduced with JDK 9)?

I have implemented some Checks for Checkstyle by my own. My project complains about an '_unstable module name derived from the file name_'. Even Maven warns about not uploading my project to a public repository, since the build is far from being stable.

Since the filename is something like `checkstyle-10.12.6.jar`, Java ends up with the module name `checkstyle`. Usually, a good naming would be the name of the main package, in this case `com.puppycrawl.tools.checkstyle`. This is more expressive and more unique.

You may achieve this just by adding a `module-info.java` into your source root. For the beginning you may export all your packages and open those, that require reflection.

If this is not an option, please add `AutomaticModuleName` to the `MANIFEST.MF`. In this case, at least the module name is no longer constructed from the file name.

https://dzone.com/articles/automatic-module-name-calling-all-java-library-maintainers

romani:
>Even Maven warns about not uploading my project to a public repository, since the build is far from being stable.

Please share details. We use maven for long time, never had a problem.

>Since the filename is something like checkstyle-10.12.6.jar

I thought all maven artifacts are named like this, org name is in path/folders of maven repository. Please share details/examples that some other old projects are migrating.

>Usually, a good naming would be the name of the main package, in this case com.puppycrawl.tools.checkstyle

Did you miss version and extension?

future2r:
It is not about Maven, it is about the Java 9 module system.

This is the `module-info.java` of the project with my implemented checks:

```
module com.qualitype.modaplex.checkstyle {

    requires java.xml;
    requires transitive checkstyle;

    exports com.qualitype.modaplex.checkstyle;
}
```
**Java Compiler (since Java 9):**
"Name of automatic module 'checkstyle' is unstable, it is derived from the module's file name."

**Maven:**
"[WARNING] Required filename-based automodules detected: [checkstyle-10.12.6.jar]. Please don't publish this project to a public artifact repository!"

Please check this for more background: https://dzone.com/articles/automatic-module-name-calling-all-java-library-maintainers

romani:
I am not sure I can find time to investigate all of this.
Can you contribute module definition to us ?
 Maybe by example and initial implementation we will be onboarder quicker. I will invite plugin maintainers to double check that we will not affect them.

future2r:
Sure, I can try.

[...]

PR #14138, opened 2023-12-13
Issue: #14120 Automatic Module

future2r:
Part of #14120 
Added `Automatic-Module-Name` to `MANIFEST.MF`.

When using Checkstyle as a dependency (e.g. when implementing own checks), this will fix the warnings issued by the Java compiler and by Maven about an unstable automatic module name that is derived from the file name.

With this change, Checkstyle is still treated as an automatic module (as before) but with a stable name.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1557,6 +1557,9 @@
               <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
               <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
             </manifest>
+            <manifestEntries>
+              <Automatic-Module-Name>${project.groupId}.${project.artifactId}</Automatic-Module-Name>
+            </manifestEntries>
           </archive>
           <excludes>
             <exclude>**/Input*.*</exclude>


nrmancuso (approved):
I am good, @future2r please make CI happy.

romani (approved):
I am ok to merge but this PR needs review of sonar and eclipscs plugin maintainers.
Just to be sure there is no side effects.

pom.xml
```diff
@@ -1552,6 +1552,9 @@
               <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
               <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
             </manifest>
+            <manifestEntries>
+              <Automatic-Module-Name>com.puppycrawl.tools.checkstyle</Automatic-Module-Name>
```
rnveach, line 1556: Can we change it to something like:
````suggestion
              <Automatic-Module-Name>${project.groupId}.${project.artifactId}</Automatic-Module-Name>
````

romani:
@muhlba91, @Calixte, @Bananeweizen, please test that jar is loading fine with this update.

pom.xml
future2r: Sounds like a good idea. Please keep in mind, that the Java module name may be the same as the Maven identifier but does not need to be. Often Maven identifiers have a dash. This is not allowed in a module name. But in your case, it will work of course.

rnveach (changes requested):
From comment at 
https://github.com/checkstyle/checkstyle/pull/14139#issuecomment-1855262066 ,
I am confused, is this PR unnecessary if we merge the module-info PR?

future2r:
> From comment at [#14139 (comment)](https://github.com/checkstyle/checkstyle/pull/14139#issuecomment-1855262066) , I am confused, is this PR unnecessary if we merge the module-info PR?

Yes it is. But adding a `module-info.java` has a much bigger impact and needs to be tested very thoroughly and of course fine-tuned (see my notes there). Adding just the `Automatic-Module-Name` is an interim solution to get rid of some problems very quickly.

Once more I suggest this document, it explains everything in detail, much better than I can do here.
https://dzone.com/articles/automatic-module-name-calling-all-java-library-maintainers

future2r:
> What is the harm if we keep both and do not remove this? Will module-info always take precedence, or will it cause issues if the 2 don't match up?

As far as I know, there is no harm. But it is useless. If there is a `module-info.java`, the `Automatic-Module-Name` is ignored. I don't know, maybe there will be a warning during packaging.
I suggest removing it. Otherwise, this question will be raised again and again until someone finally removes it.

muhlba91:
i agree. i‘d also suggest go just remove it; also because that way you know exactly what module name/info is taken and if a new module were to be introduced without a module info i‘d expect the build to throw errors as well.
some automatic setting could lead more likely to some inconsistencies or work/more impactful changes later when removed…

Bananeweizen:
> @muhlba91, @Calixte, @Bananeweizen, please test that jar is loading fine with this update.

EclipseCS should not be affected by this change or by the other module creation PR. It doesn't load the original jar at all. At _build_ time, the JAR is unbundled and the sources are repackaged into the EclipseCS sources. So as long as the classes are still organized as before, everything should be fine in our world. (and yes, I want to get rid of the unbundling, but there is so much else to fix with classpaths etc. before that can happen).

romani:
Test failing: `Proper commit message should adhere to the following rules:`

romani:
@rnveach, PR is rebased, commits are fixed.

romani:
@future2r

>When using Checkstyle as a dependency (e.g. when implementing own checks), this will fix the warnings issued by the Java compiler and by Maven about an unstable automatic module name that is derived from the file name.

please share link to your project and exact warning that you have to let us know what we are doing.

rnveach (approved):

romani:
@future2r , if your project is open source please share link to it to show how you experience this warming problem, it might help us understand how to make module-info.java

Issue #12586, opened 2023-01-05
Optimize testing methods like verifyWithInlineConfigParser to run Check one time over file


romani:
identified at https://github.com/checkstyle/checkstyle/pull/12322#discussion_r1061072126

code:
```
protected final void verifyWithInlineConfigParser(String filePath, String... expected)
            throws Exception {
        final TestInputConfiguration testInputConfiguration =
                InlineConfigParser.parse(filePath);
        final DefaultConfiguration parsedConfig =
                testInputConfiguration.createConfiguration();
        verifyViolations(parsedConfig, filePath, testInputConfiguration.getViolations());
        verify(parsedConfig, filePath, expected);
    }
```

does parsing and execution of Check twice. It was done on purpose at the begining of project on migration to new testing model.
as now we are in final stages of migration completion, we see weirdness of design more acute.
We need to finish migration to new model.
After resolving https://github.com/checkstyle/checkstyle/issues/11214
we need to refactor such verify methods to minimize execution of Check.

Vyom-Yadav:
I am on it.

romani:
@Vyom-Yadav , please fail at least referenced PR being merged.

Vyom-Yadav:
> @Vyom-Yadav , please fail at least referenced PR being merged.

I didn't understand, fail?

shubh220922:
@Vyom-Yadav
> I didn't understand, fail? 

He means 'wait' maybe as the referenced PR has not been merged yet.

shubh220922:
@Vyom-Yadav related PR has been merged.

suniti0804:
I am on it

PR #14145, opened 2023-12-14
Issue #12586: Refactor test method to execute Check only once for AbsractJavadocCheckTest

suniti0804:
Issue #12586 

This commit resolves an issue where the Check was being executed twice in the 2 test methods of class `AbstractJavadocCheckTest`.  This was due to the `verifyWithInlineConfigParser` method running the Check twice. 

This change ensures that the test methods now expects the correct number of Javadocs, fixing the error in the test.

_Alternatively we can - 
Overload the 'verifyWithInlineConfigParser' method and add only one argument - (String filePath). Please let me know which way will suit us more._

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
@@ -299,6 +299,24 @@ public abstract class AbstractModuleTestSupport extends AbstractPathTestSupport 
     }
 
     /**
+     * Performs verification of the file with the given file path using specified configuration
+     * and the array of expected messages. Unlike verifyWithInlineConfigParser, this method
+     * does not perform an additional verification step, thereby executing the Check only once.
+     *
+     * @param filePath file path to verify.
+     * @param expected an array of expected messages.
+     * @throws Exception if exception occurs during verification process.
+     */
+    protected final void verifyOnceWithInlineConfigParser(String filePath, String... expected)
+            throws Exception {
+        final TestInputConfiguration testInputConfiguration =
+                InlineConfigParser.parse(filePath);
+        final DefaultConfiguration parsedConfig =
+                testInputConfiguration.createConfiguration();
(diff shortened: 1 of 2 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
```diff
@@ -298,6 +298,24 @@ protected final void verifyWithInlineConfigParser(String filePath1,
...
+    /**
+     * Performs verification of the file with the given file path using specified configuration
+     * and the array of expected messages. Unlike verifyWithInlineConfigParser, this method
+     * does not perform an additional verification step, thereby executing the Check only once.
+     *
+     * @param filePath file path to verify.
+     * @param expected an array of expected messages.
+     * @throws Exception if exception occurs during verification process.
+     */
+    protected final void verifyOnceWithInlineConfigParser(String filePath, String... expected)
```
romani, line 310: As it is performance issue.
We don't need new method, we just need to fix existing method and fix it for all executions, all tests.
I hope in final result you archive significant performance improvement of ` mvn test` so reduce significantly whole CI execution, and execution on local

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
suniti0804: Do we need to add an overloaded method or do we need to modify the content of existing method only. Also, modifying it for one case might not suit the other case imo. We can try if that's the requirement.

romani: Let's try migrate all at one PR, they should work the same way, you can reuse existing method and simply optimize it, no overloaded methods are required.

nrmancuso:
@suniti0804 do you plan to continue on here?

nrmancuso:
We have lost contact with this contributor, anyone is welcome to reuse the work here.

PR #14156, opened 2023-12-19
Issue #13307: support for 'X violations above:'

### LINKED ISSUE #13307: Add support to our Inputs for lines with multiple violations
This was a problem for long time but become very acute at https://github.com/checkstyle/checkstyle/pull/13276#discussion_r1237941452

We need to update our Abstract Test classes to allow lines like: 
```
    Runnable noop = () ->{}; // 4 violations, see messages below 
                          // ''->' is not followed by whitespace'
                          // ''{' is not followed by whitespace'
                          // ''{' is not preceded with whitespace'
                          // ''}' is not preceded with whitespace'
```

so all messages are just on new lines below ` 4 violations, see messages below `
So Reviewer will clearly see what is a violation.

example: 
https://github.com/checkstyle/checkstyle/blob/3d8ed42ab60c0399550b3e7d54ca9dd767583101/src/test/resources/com/puppycrawl/tools/checkstyle/checks/annotation/annotationlocation/InputAnnotationLocationIncorrect.java#L38
but there are more in code.

pattern that used at [PR](https://github.com/checkstyle/checkstyle/pull/13951), may be extra `:` is required to signal that second line is continuation:
```
    /**
     *  <p> <li> <TR> <Td> <tH> <.....
     *      <tFoot> // ok
     * @param field5 </p> value to which {@link #field5} is to be set to
     */
    // 4 violations 4 lines above:
    //                            'Unclosed HTML tag found: p'
    //                            'tag P_TAG_START'
    //                            'tag LI_TAG_START'
    //                            'tag BODY_TAG_START'
    private void setField5(int field5) {this.field5 = field5;}
```

As we process lines one by one, of we detect multiple-line pattern we look ahead for defined set of lines. Processing of such lines with violation only will not match any pattern so they will be skipped by general iteration over lines. Location of single quoted message can be at any indentation, to let user user any formatting to show message as he wants.

romani:
Fixes #13307

VIOLATIONS_ABOVE_PATTERN_WITH_MESSAGES was MULTIPLE_VIOLATIONS_ABOVE_PATTERN_WITH_MESSAGES but PMD complained on length, it kind of right, so I removed MULTIPLE_.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
@@ -53,7 +53,7 @@ public final class InlineConfigParser {
 
     /**
      * Pattern for lines under
-     * {@link InlineConfigParser#MULTIPLE_VIOLATIONS_SOME_LINES_ABOVE_PATTERN}.
+     * {@link InlineConfigParser#VIOLATIONS_SOME_LINES_ABOVE_PATTERN}.
      */
     private static final Pattern VIOLATION_MESSAGE_PATTERN = Pattern
             .compile(".*//\\s*(?:['\"](.*)['\"])?$");
@@ -159,7 +159,7 @@ public final class InlineConfigParser {
      * Messages are matched by {@link InlineConfigParser#VIOLATION_MESSAGE_PATTERN}
      * </p>
      */
-    private static final Pattern MULTIPLE_VIOLATIONS_SOME_LINES_ABOVE_PATTERN = Pattern
+    private static final Pattern VIOLATIONS_SOME_LINES_ABOVE_PATTERN = Pattern
             .compile(".*//\\s*(\\d+) violations (\\d+) lines above:$");
 
     /** The String "(null)". */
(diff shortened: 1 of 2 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -132,6 +132,21 @@ public final class InlineConfigParser {
...
+     *   // X violations above:
+     *   //                    'violation message1'
+     *   //                    'violation messageX'
+     * }
+     *
+     * Messages are matched by {@link InlineConfigParser#VIOLATION_MESSAGE_PATTERN}
+     * </p>
+     */
+    private static final Pattern VIOLATIONS_ABOVE_PATTERN_WITH_MESSAGES = Pattern
+            .compile(".*//\\s*(\\d+) violations above:\\s*(?:['\"](.*)['\"])?$");
```
rnveach, line 148: What is the purpose of `(?:['\"](.*)['\"])?` ? I don't see it being used anywhere and it seems to be all over the place in this class.

romani: Good point, we do not use it, we need to remove it.

romani: In all other places, it is to get a message from line and match it to what Check is generating

romani: fixed.

rnveach: Not really. Now that you removed the rest, it now raises why do we need `\\s*` on the end? We don't allow trailing spaces, and `*` says it is completely optional. So it should be removed as well.

romani: fixed.

rnveach (approved):

nrmancuso (approved):

Issue #14178, opened 2023-12-23
Javadoc on Module property should be able to define special default value


romani:
detected at: 

CustomImportOrderCheck.customImportOrderRules has very special class field that defined it default value.
https://github.com/checkstyle/checkstyle/blob/3ba30a040638c3ff14c44676a2f019930bfaf827/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java#L349-L358

https://github.com/checkstyle/checkstyle/blob/3ba30a040638c3ff14c44676a2f019930bfaf827/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java#L260-L261

usage of mactros results in:
`    -Property {@code customImportOrderRules} - Specify format of order declaration customizing by user. Type is {@code java.lang.String}. Default value is {@code null}.`
`    +Property {@code customImportOrderRules} - Specify format of order declaration customizing by user. Type is {@code java.lang.String}. Default value is {@code ""}.`

we need to remove temporal workaround with hardcoded default value and find way to make it declarative in javadoc of setter.

romani:
my bad ..... `{@code ""}` is what macros demands and it is right , it is a value of DEFAULT_CUSTOM_IMPORT_ORDER_RULES

romani:
One of the approach to address this is to make new javadoc tag:
`@defaultvalue DEFAULT_CUSTOM_IMPORT_ORDER_RULES`

relentless-pursuit:
I am on it.

romani:
@relentless-pursuit, please beware that issue is not approved, so please send draft PR to let other maintenance agree on design and you will not waste time on redoing

relentless-pursuit:
Sure

relentless-pursuit:
@Romani, any reference PR for creating custom Javadocs tags?

romani:
Look at history  commit of this line https://github.com/checkstyle/checkstyle/blob/d7f97f34021bab6e5027a947eabbcd40300d4567/pom.xml#L531

romani:
@relenless-pursuit, are you doing something on this issue ?

@rnveach, @nrmancuso, lets agree on name of javadoc tag, see https://github.com/checkstyle/checkstyle/issues/14178#issuecomment-1868553955 . It might be useful more widely as we start removal of javadoc for all propertties from content of javadoc over class.

rnveach:
I don't agree with this issue.
[...]

Issue #13376, opened 2023-07-07
Place back removed property `customImportOrderRules` from `CustomImportOrderCheck`


stoyanK7:
In #13362 we restored validation that was removed in https://github.com/checkstyle/checkstyle/pull/13117/files#diff-350da613cecad468adf5dcb54826f8aa2cca63e699aee79a7d8bad4f8257f48d. In this time window where the validation was absent, a problem slipped - property `customImportOrderRules` was removed from `CustomImportOrderCheck` in https://github.com/checkstyle/checkstyle/commit/7d35a52067c994eb12312b8fff0a072a198d2c00. This in turn broke an assertion in `XdocsPagesTest` that we brought back. The `getField()` method now returns null 
https://github.com/checkstyle/checkstyle/blob/1ef4cd7e10a71dcbc584ad306049ec355d3888de/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L944-L946

Comments from the discussion at https://github.com/checkstyle/checkstyle/pull/13362#discussion_r1253244719

> We should not be allowing pitest to dictate class design/common patterns in Checkstyle. My expectation has always been that a user facing property has an associated field in check class (or superclass).

> Let's do hack and skip this validation for this module and this property. Create separate issue to comeback to it and fix it. Let's merge this PR with hack.

romani:
Property will not be restored.

We need to take content for javadoc from setter.

PR #14180, opened 2023-12-23
Issue 13693 - CustomImportOrderCheck

### LINKED ISSUE #13693: Use module properties macro in all templates

romani:
closes #13693

https://checkstyle.org/checks/imports/customimportorder.html#Properties

nuance with not easy property default value is postponed until https://github.com/checkstyle/checkstyle/issues/14178 all explanation why we need override is here.

### DIFF
--- a/config/checkstyle-non-main-files-suppressions.xml
+++ b/config/checkstyle-non-main-files-suppressions.xml
@@ -250,11 +250,6 @@
          files="src[\\/]xdocs[\\/]checks[\\/]coding[\\/]unnecessarysemicoloninenumeration.xml.template"/>
   <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]checks[\\/]coding[\\/]unusedlocalvariable.xml.template"/>
-  <!-- Properties macro suppressions until https://github.com/checkstyle/checkstyle/issues/13693 -->
-  <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]checks[\\/]imports[\\/]customimportorder.xml.template"/>
-
-  <!-- no properties by design -->
   <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]checks[\\/]design[\\/]finalclass.xml.template"/>
   <suppress id="propertiesMacroMustExist"
--- a/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
@@ -187,7 +187,7 @@ public final class SiteUtil {
     /**
      * Frequent version.
      */
-    private static final String V824 = "8.24";
+    private static final String VERSION_8_24 = "8.24";
 
     /**
      * Frequent version.
@@ -228,7 +228,7 @@ public final class SiteUtil {
      * @noinspectionreason JavacQuirks until #14052
      */
     private static final Map<String, String> SINCE_VERSION_FOR_INHERITED_PROPERTY = Map.ofEntries(
-        Map.entry("MissingDeprecatedCheck.violateExecutionOnNonTightHtml", V824),
+        Map.entry("MissingDeprecatedCheck.violateExecutionOnNonTightHtml", VERSION_8_24),
         Map.entry("NonEmptyAtclauseDescriptionCheck.violateExecutionOnNonTightHtml", "8.3"),
         Map.entry("HeaderCheck.charset", VERSION_5_0),
         Map.entry("HeaderCheck.fileExtensions", VERSION_6_9),
@@ -253,9 +253,9 @@ public final class SiteUtil {
         Map.entry("OrderedPropertiesCheck.fileExtensions", "8.22"),
         Map.entry("UniquePropertiesCheck.fileExtensions", VERSION_5_7),
         Map.entry("TranslationCheck.fileExtensions", VERSION_3_0),
-        Map.entry("LineLengthCheck.fileExtensions", V824),
+        Map.entry("LineLengthCheck.fileExtensions", VERSION_8_24),
         // until https://github.com/checkstyle/checkstyle/issues/14052
-        Map.entry("JavadocBlockTagLocationCheck.violateExecutionOnNonTightHtml", V824),
+        Map.entry("JavadocBlockTagLocationCheck.violateExecutionOnNonTightHtml", VERSION_8_24),
         Map.entry("JavadocMissingLeadingAsteriskCheck.violateExecutionOnNonTightHtml", "8.38"),
         Map.entry(
             "RequireEmptyLineBeforeBlockTagGroupCheck.violateExecutionOnNonTightHtml",
--- a/src/xdocs/checks/imports/customimportorder.xml
+++ b/src/xdocs/checks/imports/customimportorder.xml
@@ -129,7 +129,7 @@ import com.puppycrawl.tools.checkstyle.checks.imports.AvoidStarImportCheck;
             <tr>
               <td>customImportOrderRules</td>
               <td>Specify ordered list of import groups.</td>
-              <td><a href="../../property_types.html#String[]">String[]</a></td>
+              <td><a href="../../property_types.html#String.5B.5D">String[]</a></td>
               <td><code>{}</code></td>
               <td>5.8</td>
             </tr>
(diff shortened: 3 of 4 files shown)


romani:
I will put annotation on method:)

rnveach (changes requested):
CI is failing.

Nothing has been said why this has to be a custom override and we don't accept the default of the system we have set up, especially for an custom list of 1. I am also against adding something like this to a javadoc, as there is nothing requiring we update it if we change the default value.

https://github.com/checkstyle/checkstyle/issues/13376 doesn't say what is "broken" about our validation.

What is the current default displayed for this type, and why do we have that if we don't want it showing up?

romani:
>Nothing has been said why this has to be a custom override

I thought https://github.com/checkstyle/checkstyle/issues/14178 explains all.

>and we don't accept the default of the system we have set up,

We need override to avoid default value charges, from "" to null. As it is pointed in issue, this property is not simple as it doesn't have class field that represents it, so we do not have easy way to get default value, as it is in very special constant. I do not know what is best way to describe default values for such cases, so to avoid long discussions I am postponing it for late to get all cases like this, as we do not know them now.

>What is the current default displayed for this type,

Empty string.

>why do we have that if we don't want it showing up ?

We do what to keep empty string. Our parents implementation is limited to get default from class field but in this case we do not have class field, so all end up to null, we do not want null.

rnveach:
> I thought https://github.com/checkstyle/checkstyle/issues/14178 explains all.

It explains more after your update.

> all end up to null, we do not want null

This goes back to my question `why do we have that if we don't want it showing up?`. Why does all our other work arounds for our set up not fix the issue and we need to spend time on what is right now a 1 time work around?

romani:
>why do we have that if we don't want it showing up?

Our logic is correct: get default value field. But we have nuance, there is no field here, default value is different field and applies is setter by special logic. It is what it is.  Each time I touch some check for migration, so crazy nuance pop up.  So now we know that not only for since we need so declaration in class, but also for default. Making generalization on early stages is not good, it is error prone. A way better to do automation when we see all edge cases together. @relentless-pursuit
Already try to help in it but issue is not approved, I hope draft PR will make it a bit more clear.

>Why does all our other work arounds for our set up not fix the issue

We never run into this before 

>we need to spend time on what is right now

Issue is not approved, nobody has time to investigate. I think we will catch more cases like this. I will be more clear what to do later, in more right way. I still believe we do not know enough to make automation in right way.

>which we don't know if it will ever be used again?

In this case we loose nothing. If we lucky that much, it will be simple. But I don't think so, there should be more cases like this, it sounds like pitest demands to remove absolete code

rnveach:
I am still not following. You say our code/logic is correct, but then go on to say it isn't correct for this check. If we need an override, then either something is wrong in the code or this check doesn't follow 99% of our other rules we have set up and working.

Why can't we correct code when field can not be found to display blanks instead of `null`?
Why can't we rename field to match setter so default can be identified?
Why is `null` not valid and empty string makes a difference? And then this goes back to why aren't we correcting all `null`s like the first question?

romani:
> or this check doesn't follow 99% of our other rules

Check is not following our naive/simple rules. It is first time we run into such implementation. 

>Why can't we correct code when field can not be found to display blanks instead of null?

it will affect bunch of other Checks that already expect null as fallback.

>Why can't we rename field to match setter so default can be identified?

it is constant, it is not a field that can set new value.

>Why is null not valid and empty string makes a difference?

because it will change default value in xdoc/html and we are not ready to change it, "null" visible for users is bad design of Checks, user can not set `null`, so we should not show what user can not set in XML config.

romani:
I am going to finish migration this week, and start fixing issues that we postponed until end of macros migration to eliminate unfortunate cases in SiteUtils.java

romani:
@rnveach, please consider this PR to be merged before better fix is provided.
Better to have 100% migrated modules to easily see all changes in result xml files on each update of meta model calculation.
So there will be no conversation on what is correct default value or any other values, in separate PR we will just affect versions, nothing else.

rnveach:
> Why can't we correct code when field can not be found to display blanks instead of null?
> it will affect bunch of other Checks that already expect null as fallback.

This is just documentation. It doesn't impact check logic, so others that expect null as fallback will still have it.

> Why can't we rename field to match setter so default can be identified?
> it is constant

Constants are not properties. Properties are mutable.
https://github.com/checkstyle/checkstyle/blob/8531fec143058ddd92fab14839f2e9389ac6cf5c/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java#L361

> "null" visible for users is bad design of Checks, user can not set null, so we should not show what user can not set in XML config.

So if `null` is never a good choice, then this is why we must swap the default to blanks for this kind of situation in the macro code.

=====

I am not seeing anything to approve here when we are fixing a problem that doesn't yet exist. Everything is leading back to changing displaying `null` to just blanks in this scenario. We don't need 100% conversion to fix things that are wrong. We have done it in other PRs.

romani:
we do not store user provided value as-is,
https://github.com/checkstyle/checkstyle/blob/8531fec143058ddd92fab14839f2e9389ac6cf5c/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java#L264C19-L264C48
final type is List, so it is conceptually null-able.

User can not set `null` even to java bean, it will be NPE. So having `null` is completely wrong and confusing value.

> So if null is never a good choice, then this is why we must swap the default to blanks for this kind of situation in the macro code.

We do have nulls already in xdoc visible for user, to avoid "breaking changes" like updates and discussions, it was easier to just keep it as before and postpone it until a time it is actually required.

> Everything is leading back to changing displaying null to just blanks in this scenario. 

some modules are good with `null` for java beans. I do not like fact that we show users `null` but rethinking on this should be separate project, it is not easy in many cases.

> I am not seeing anything to approve here when we are fixing a problem that doesn't yet exist.

we are making our meta model be aware that some Modules has design defect and it is what it is for now.


>We don't need 100% conversion to fix things that are wrong. We have done it in other PRs.

nothing is wrong with default value, it is `""`. The only problem that it is not easy to get this value, as it is part of code in setter, not a declaration.

romani:
@rnveach, @nrmacuso, please consider this update.
it is not conflicting with update in Check that we will do, but it will make special list for "null"s for properties, I will update issue to reword it to be for fix Header Checks instead. In this case we are confident and secure that no more null will sneak to design easily.

rnveach:
To me, everything I said above is the same but now pointed at `PROPERTY_SPECIAL_DEFAULTS`. If `null` isn't valid, then we shouldn't display it for any of the others.

romani:
We can merge this and deal with null in separate PRs. If we can not find easy way to change Checks, hack will stay for longer, but main point that nobody can leak null anymore, as system will push to make it empty by default.

rnveach:
The difference in a suppression is we don't have to change any code to suppress. Here we have to create a workaround, and I don't agree with it as we have come up with atleast 2-3 other solutions where we follow our current rules, make sense, and don't need an override we don't need anywhere else in 90% of our checks.

romani:
https://github.com/checkstyle/checkstyle/issues/14216 is approved and will be fixed separately.
This update is just to finish macros udpates  and start dealing with https://github.com/checkstyle/checkstyle/issues/14234 .
We can do it righaway, if you approve issue.

romani:
@nrmancuso , I am assigning this PR to you to confirm that it is not blocking with your PR of type change.

romani:
rebased to resolve conflict.

nrmancuso (approved):

romani:
@rnveach , what comment define you request for change? I see label but not sure what you do not like.
I hope current update is small step forward with gating nulls until we make breaking changes.

rnveach:
https://github.com/checkstyle/checkstyle/pull/14217#issuecomment-1879912059

romani:
Yes, Main point is to migrate to macros.
And deal with Checks unfortunate design later on .

rnveach:
So then you understand why this PR is still awaiting changes?

romani:
@rnveach , only macros usage in this PR now.

config/checkstyle-non-main-files-suppressions.xml
```diff
@@ -250,11 +250,6 @@
          files="src[\\/]xdocs[\\/]checks[\\/]coding[\\/]unnecessarysemicoloninenumeration.xml.template"/>
   <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]checks[\\/]coding[\\/]unusedlocalvariable.xml.template"/>
-  <!-- Properties macro suppressions until https://github.com/checkstyle/checkstyle/issues/13693 -->
-  <suppress id="propertiesMacroMustExist"
-         files="src[\\/]xdocs[\\/]checks[\\/]imports[\\/]customimportorder.xml.template"/>
-
-  <!-- no properties by design -->
```
rnveach, line 257: These 2 lines can't be removed unless you are saying these checks have properties now.

rnveach (changes requested):

config/checkstyle-non-main-files-suppressions.xml
romani: done

romani: reverted, it is required to combine two parts of `no properties by design`, block above is same reason `no properties by design`

Issue #14137, opened 2023-12-12
Enable checks from Error Prone Support


rickie:
In PR https://github.com/checkstyle/checkstyle/pull/14136 I added Error Prone Support to Checkstyle. As @romani suggested in https://github.com/checkstyle/checkstyle/issues/14129, I'd like to create this issue to discuss which checks from [Error Prone Support](https://github.com/picnicSupermarket/error-prone-support/) are useful for the Checkstyle project to enable. 

The full list of checks can be found here: https://error-prone.picnic.tech/.

1. [AmbiguousJsonCreator](https://error-prone.picnic.tech/bugpatterns/AmbiguousJsonCreator/)
1. [AssertJIsNull](https://error-prone.picnic.tech/bugpatterns/AssertJIsNull/)
1. [AutowiredConstructor](https://error-prone.picnic.tech/bugpatterns/AutowiredConstructor/)
1. [CanonicalAnnotationSyntax](https://error-prone.picnic.tech/bugpatterns/CanonicalAnnotationSyntax/)
1. [CollectorMutability](https://error-prone.picnic.tech/bugpatterns/CollectorMutability/)
1. [DirectReturn](https://error-prone.picnic.tech/bugpatterns/DirectReturn/)
1. [EmptyMethod](https://error-prone.picnic.tech/bugpatterns/EmptyMethod/)
1. [ExplicitEnumOrdering](https://error-prone.picnic.tech/bugpatterns/ExplicitEnumOrdering/)
1. [FormatStringConcatenation](https://error-prone.picnic.tech/bugpatterns/FormatStringConcatenation/)
1. [IdentityConversion](https://error-prone.picnic.tech/bugpatterns/IdentityConversion/)
1. [ImmutablesSortedSetComparator](https://error-prone.picnic.tech/bugpatterns/ImmutablesSortedSetComparator/)
1. [IsInstanceLambdaUsage](https://error-prone.picnic.tech/bugpatterns/IsInstanceLambdaUsage/)
1. [JUnitClassModifiers](https://error-prone.picnic.tech/bugpatterns/JUnitClassModifiers/)
1. [JUnitMethodDeclaration](https://error-prone.picnic.tech/bugpatterns/JUnitMethodDeclaration/)
1. [JUnitValueSource](https://error-prone.picnic.tech/bugpatterns/JUnitValueSource/)
[...]

Issue #8252, opened 2020-05-24
Sonar violations: JUnit5 test classes and methods should have default package visibility


romani:
https://sonarcloud.io/project/issues?id=org.checkstyle%3Acheckstyle&open=AXItYgyWEeM3Dea7mBcs&resolved=false&severities=INFO

https://rules.sonarsource.com/java/RSPEC-5786

![image](https://user-images.githubusercontent.com/812984/82742993-96533480-9d19-11ea-9dc5-4fb470ea305e.png)

romani:
Should we follow this rule ?

rnveach:
I am fine with test only code removing the modifiers.

pbludov:
> Should we follow this rule ?
> 
> We already had discussion on this somewhere, but I do not remember where. What is left in my memory is to keep visibility as public, as it doesn't help at all to be less be public

The transition to JUnit5 was automated by IntellyJ IDEA. The access modifiers were not changed during that process, we agreed that this can be done in a separate issue.

At first glance, it looks like an easy sed task:
`sed -i "s/public void test/void test/g"`
since we already has a rule to enforce the `test` prefix in all test methods.

pbludov:
I think that all test classes should be non-public and even final, to make it impossible to use one test from another and create unnecessary dependencies.

romani:
please read - https://stackoverflow.com/a/55230350/1015848 nothing about encapsulation and less visibility (that I agreee), just a matter of style.

> and even final

it is separate issue ... we can do it even for public classes.

> create unnecessary dependencies

still be possible in the same package. Inheritance from tests classes is very weird, I have never see this in my life.

romani:
I marked as won't fix all such violations in sonar https://sonarcloud.io/summary/new_code?id=org.checkstyle%3Acheckstyle

There will be more to appear as we can not disable rule, but we have bulk update now in UI for we will mark as won't fix periodically.

PR #14192, opened 2023-12-27
Issue #14137: Disable Error Prone Support checks

rickie:
Related issue https://github.com/checkstyle/checkstyle/issues/14137.

This PR disables a few checks from Error Prone Support. Additionally, the goal of this PR is to agree on how we are going to configure the rest of the checks in the `pom.xml`. If we, at some point, decide to either disable or enable more checks from Error Prone itself that have severity `WARN` or `INFO` we can do that in the same configuration section (see [docs](https://errorprone.info/bugpatterns)).

At Picnic we have the convention to add a comment for every rule explaining why we disable it. If we would like the same here, I need your help to write the correct reasoning :). We can also decide to leave it like this. For now, I disabled two extra rules that are that will cause more noise than improvements.

[In the previous PR](https://github.com/checkstyle/checkstyle/pull/14136#issuecomment-1853104700) @rnveach mentioned:

> Log file jumped from 100 kb to 2.5 megs in this PR.  

With disabling these checks we are down to normal numbers again.

More PRs to follow :rocket:!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -237,6 +237,20 @@
     <error-prone.version>2.24.0</error-prone.version>
     <error-prone-support.version>0.14.0</error-prone-support.version>
     <doxia.version>1.12.0</doxia.version>
+    <error-prone.configuration-args>
+      <!-- Reason at https://github.com/checkstyle/checkstyle/issues/8252. -->
+      -Xep:JUnitClassModifiers:OFF
+      <!-- Reason at https://github.com/checkstyle/checkstyle/issues/8252. -->
+      -Xep:JUnitMethodDeclaration:OFF
+      <!-- We prefer different style of writing tests. -->
+      -Xep:JUnitValueSource:OFF
+      <!-- Until https://github.com/checkstyle/checkstyle/issues/14194. -->
+      -Xep:LexicographicalAnnotationListing:OFF
+      <!-- This check is not finalized. -->
+      -Xep:MethodReferenceUsage:OFF
+      <!-- We prefer to qualify all static method calls with class name. -->
+      -Xep:StaticImport:OFF
+    </error-prone.configuration-args>
   </properties>
 
   <dependencies>
@@ -2380,7 +2394,7 @@
                     <arg>-Xpkginfo:always</arg>
                     <arg>-XDcompilePolicy=simple</arg>
                     <arg>
-                      -Xplugin:ErrorProne
+                      -Xplugin:ErrorProne ${error-prone.configuration-args}
                     </arg>
                   </compilerArgs>
                   <annotationProcessorPaths>
@@ -2438,7 +2452,8 @@
                     <arg>-XDcompilePolicy=simple</arg>
                     <arg>
                       -Xplugin:ErrorProne \
-                      -XepExcludedPaths:.*[\\/]resources[\\/].*
+                      -XepExcludedPaths:.*[\\/]resources[\\/].* \
+                      ${error-prone.configuration-args}
                     </arg>
                   </compilerArgs>
                   <annotationProcessorPaths>
@@ -2451,7 +2466,7 @@
                       <groupId>tech.picnic.error-prone-support</groupId>
                       <artifactId>error-prone-contrib</artifactId>
                       <version>${error-prone-support.version}</version>
-                    </path>b
+                    </path>
                   </annotationProcessorPaths>
                 </configuration>
               </execution>


pom.xml
```diff
@@ -237,6 +237,19 @@
...
+      Error Prone Support configuration:
+      -->
+      -Xep:JUnitClassModifiers:OFF
+      -Xep:JUnitMethodDeclaration:OFF
+      <!-- Check is not configurable yet. -->
+      -Xep:LexicographicalAnnotationListing:OFF
+      <!-- Too many false positives. -->
+      -Xep:MethodReferenceUsage:OFF
+      -Xep:JUnitValueSource:OFF
+      -Xep:StaticImport:OFF
```
romani, line 251: > Too many false positives.

better to say `Disabled as we do not see value in such rules`.
Rule might work well, but our team does think it matters or do code less explicit.

pom.xml
```diff
@@ -237,6 +237,19 @@
...
     <error-prone-support.version>0.14.0</error-prone-support.version>
     <doxia.version>1.12.0</doxia.version>
+    <error-prone.configuration-args>
+      <!--
+      Error Prone Support configuration:
+      -->
+      -Xep:JUnitClassModifiers:OFF
+      -Xep:JUnitMethodDeclaration:OFF
+      <!-- Check is not configurable yet. -->
+      -Xep:LexicographicalAnnotationListing:OFF
```
romani, line 247: in issue I see 
>please activate:
[LexicographicalAnnotationAttributeListing](https://error-prone.picnic.tech/bugpatterns/LexicographicalAnnotationAttributeListing/)

please share a reason.

romani (changes requested):
items:

pom.xml
rickie: Correct, note the slight difference, we want to enable LexicographicalAnnotation**Attribute**Listing, but here I disabled the _LexicographicalAnnotationListing_. That simply sorts annotations lexicographically, which is not that desirable in most cases.

pom.xml
rickie: I pushed a proposal, let me know if you agree to group the checks like this.

pom.xml
romani: @rnveach , @nrmancuso , do we want  to try sorting of annotations? 
I do think we have much of them, 1-2 is max on method I think

pom.xml
```diff
@@ -2451,7 +2461,7 @@
                       <groupId>tech.picnic.error-prone-support</groupId>
                       <artifactId>error-prone-contrib</artifactId>
                       <version>${error-prone-support.version}</version>
-                    </path>b
+                    </path>
```
romani, line 2464: Please remove this extra char

romani (changes requested):
Please keep single commit always to avoid CI failures

pom.xml
romani: We can disable it for now, and enable it in separate PR.
This is not a blocker to merge this PR

romani: When we introduce new tools, we usually suppress all one by one , explicitly and in next PRs, enable one by one.

pom.xml
rickie: I noticed this got introduced in the previous PR, so I dropped it in this commit.

pom.xml
rickie: > @rnveach , @nrmancuso , do we want to try sorting of annotations?

Sorry for not checking this one with you guys, I didn't think too much about it as we don't use it internally. If there aren't that many annotations, it might be a nice addition.

> When we introduce new tools, we usually suppress all one by one , explicitly and in next PRs, enable one by one.

Do you mean that we should split the _current_ PR into multiple PRs? It's fine by me if you prefer this.

In any case, for next PRs I will keep it strictly to enabling **one** check per PR.

rickie:
> Please keep single commit always to avoid CI failures

Noted, rebased and squashed the commits.

pom.xml
romani: No split for now, let's merge as is.

In next PRs we can use this approach, if convenient.

romani (approved):
Ok to merge

pom.xml
romani: In next PRs we can reuse https://github.com/checkstyle/checkstyle/tree/master/config/error-prone-suppressions to activate rule and simply put all/controversial violations to suppression and deal with them separately. At least nobody will merge a code to introduce more violations, there are a lot concurrent PRs now.

nrmancuso: As for annotation sorting, I prefer to have simple `@Blah` above all annotations like `@Blah(value={x, y, x})`, I do not care so much about lexicographical order.

romani: Ordering usually becomes matter, if there are numerous annotations.

@nrmancuso , do you mean you like piramid style of annotations? The most parametrized the bottom

nrmancuso: > "Pyramid"

Yes! This is a good way to describe it. I feel like the basic `@Blah` annotations sort of get lost, especially in Spring Boot applications.

romani: Ok, good to know, but this discussion should not block this PR, please review PR to let us unleash disablement of some rules. We can reconsider disablement in separate PR.

pom.xml
```diff
@@ -237,6 +237,15 @@
     <error-prone.version>2.24.0</error-prone.version>
     <error-prone-support.version>0.14.0</error-prone-support.version>
     <doxia.version>1.12.0</doxia.version>
+    <error-prone.configuration-args>
+      <!-- Disabled rules as we do not see value in them. -->
+      -Xep:JUnitClassModifiers:OFF
```
nrmancuso, line 242:
```suggestion
      <!-- until https://github.com/checkstyle/checkstyle/issues/8252 -->
      -Xep:JUnitClassModifiers:OFF
```

pom.xml
```diff
@@ -237,6 +237,15 @@
...
+      -Xep:JUnitMethodDeclaration:OFF
```
nrmancuso, line 243:
```suggestion
      <!-- until https://github.com/checkstyle/checkstyle/issues/8252 -->
      -Xep:JUnitMethodDeclaration:OFF
```

pom.xml
```diff
@@ -237,6 +237,15 @@
...
+      -Xep:JUnitValueSource:OFF
```
nrmancuso, line 244: Why not activate this pattern?

pom.xml
```diff
@@ -237,6 +237,15 @@
...
+      -Xep:LexicographicalAnnotationListing:OFF
+      -Xep:MethodReferenceUsage:OFF
```
nrmancuso, line 246: We shouldn't have many (any?)  violations on this one, since we have an IDEA inspection for this. 

For reference: https://error-prone.picnic.tech/bugpatterns/MethodReferenceUsage/#methodreferenceusage

pom.xml
```diff
@@ -237,6 +237,15 @@
...
+      -Xep:StaticImport:OFF
```
nrmancuso, line 247:
```suggestion
      <!-- we prefer to qualify all static method calls with class name -->
      -Xep:StaticImport:OFF
```

nrmancuso (changes requested):
Items:

pom.xml
```diff
@@ -237,6 +237,15 @@
     <error-prone.version>2.24.0</error-prone.version>
     <error-prone-support.version>0.14.0</error-prone-support.version>
     <doxia.version>1.12.0</doxia.version>
+    <error-prone.configuration-args>
+      <!-- Disabled rules as we do not see value in them. -->
```
nrmancuso, line 241: We can remove this since we will comment on each pattern individually.

nrmancuso, line 241: We can remove this since we will comment on each pattern individually.

pom.xml
```diff
@@ -237,6 +237,15 @@
...
+      -Xep:JUnitClassModifiers:OFF
+      -Xep:JUnitMethodDeclaration:OFF
+      -Xep:JUnitValueSource:OFF
+      -Xep:LexicographicalAnnotationListing:OFF
```
nrmancuso, line 245:
```suggestion
       <!-- until https://github.com/checkstyle/checkstyle/issues/14194 -->
      -Xep:LexicographicalAnnotationListing:OFF
```

nrmancuso (changes requested):

pom.xml
romani: We need to close this issue, we are not going to follow this style for now, may be in future we will, but I don't see much reason.

romani: Issue is closed.
Please make comment `<!-- reason at https://github.com/checkstyle/checkstyle/issues/8252 -->`

pom.xml
romani: https://error-prone.picnic.tech/bugpatterns/JUnitValueSource/

We do not use this pattern of test design, we eliminating all tests that are unit tests, we do all tests by full checkstyle execution.

pom.xml
romani: If no violations present now, let's activate.
If there are some let's add it to https://github.com/checkstyle/checkstyle/issues/14137 and we can use comment `until https://github.com/checkstyle/checkstyle/issues/14137`

I highly recommend to merge basic suppression sooner and deal with all activations in separate PRs.

pom.xml
romani: We do most of code review in phone browsers, no ide, so code must be explicit.

pom.xml
romani: https://github.com/checkstyle/checkstyle/wiki/Checkstyle-GSoC-2021-Project-Ideas#project-name-adaptation-of-behavior-driven-development-ideas-for-testing-of-source-code-validation-algorithms

pom.xml
nrmancuso: >We need to close this issue, we are not going to follow this style for now, may be in future we will, but I don't see much reason.

Reasoning is sound: we do not want to allow/encourage sharing of test methods across test classes.

romani: Changing visibility does not protect again reuse, we are open for any reuse of public method, usually it is nonsense and not possible to have any reason.

pom.xml
rickie: The only problem is that this is the only rule in Error Prone Support that is not ready for use yet. For historical reasons this check is already "available" like this, but it's better for now to disable it. We also disabled it internally.

rickie: For now I added this: `<!-- This check is not finalized. -->`. Is this okay?

pom.xml
rickie: For now I added: `<!-- We prefer different style of writing tests. -->`. 

Let me know if you prefer something else.

A quick search showed me that in the project there are only 2 parameterized tests, both use `@ValueSource`.

rickie:
> We usually practice the same :) I can assist with this task.

Nice, let's do it! 

I added the comments.

nrmancuso (approved):
Let’s do it

nrmancuso:
@rickie thank you very much!

PR #14199, opened 2023-12-29
Issue #11230: fix FallThroughCheck for labeled statements

relentless-pursuit:
Resolves: #11230

This PR aims to fix FallThroughCheck for labeled statements. 

Diff regression config: https://gist.githubusercontent.com/relentless-pursuit/5f484c60cf4933dcc3cadc564e7e574f/raw/70d35777b174a9d980d09644bb45d1c4fc483365/config_fallThroughCheck.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
@@ -19,7 +19,9 @@
 
 package com.puppycrawl.tools.checkstyle.checks.coding;
 
+import java.util.HashSet;
 import java.util.Optional;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
@@ -150,8 +152,8 @@ public class FallThroughCheck extends AbstractCheck {
         if (!isLastGroup || checkLastCaseGroup) {
             final DetailAST slist = ast.findFirstToken(TokenTypes.SLIST);
 
-            if (slist != null && !isTerminated(slist, true, true)
-                && !hasFallThroughComment(ast)) {
+            if (slist != null && !isTerminated(slist, true, true, new HashSet<>())
+                    && !hasFallThroughComment(ast)) {
                 if (isLastGroup) {
                     log(ast, MSG_FALL_THROUGH_LAST);
                 }
@@ -165,14 +167,17 @@ public class FallThroughCheck extends AbstractCheck {
     /**
      * Checks if a given subtree terminated by return, throw or,
      * if allowed break, continue.
+     * When analyzing fall-through cases in switch statements, a Set of String labels
+     * is used to keep track of the labels encountered in the enclosing switch statements.
      *
      * @param ast root of given subtree
      * @param useBreak should we consider break as terminator
      * @param useContinue should we consider continue as terminator
+     * @param labelsForCurrentSwitchScope the Set labels for the current scope of the switch
      * @return true if the subtree is terminated.
      */
     private boolean isTerminated(final DetailAST ast, boolean useBreak,
-                                 boolean useContinue) {
+                                 boolean useContinue, Set<String> labelsForCurrentSwitchScope) {
         final boolean terminated;
 
         switch (ast.getType()) {
@@ -182,35 +187,64 @@ public class FallThroughCheck extends AbstractCheck {
                 terminated = true;
                 break;
             case TokenTypes.LITERAL_BREAK:
-                terminated = useBreak;
+                terminated =
+                        useBreak || hasLabel(ast, labelsForCurrentSwitchScope);
                 break;
             case TokenTypes.LITERAL_CONTINUE:
-                terminated = useContinue;
+                terminated =
+                        useContinue || hasLabel(ast, labelsForCurrentSwitchScope);
                 break;
             case TokenTypes.SLIST:
-                terminated = checkSlist(ast, useBreak, useContinue);
+                terminated =
+                        checkSlist(ast, useBreak, useContinue, labelsForCurrentSwitchScope);
                 break;
             case TokenTypes.LITERAL_IF:
-                terminated = checkIf(ast, useBreak, useContinue);
+                terminated =
+                        checkIf(ast, useBreak, useContinue, labelsForCurrentSwitchScope);
                 break;
             case TokenTypes.LITERAL_FOR:
             case TokenTypes.LITERAL_WHILE:
             case TokenTypes.LITERAL_DO:
-                terminated = checkLoop(ast);
+                terminated = checkLoop(ast, labelsForCurrentSwitchScope);
                 break;
             case TokenTypes.LITERAL_TRY:
-                terminated = checkTry(ast, useBreak, useContinue);
+                terminated =
+                        checkTry(ast, useBreak, useContinue, labelsForCurrentSwitchScope);
                 break;
             case TokenTypes.LITERAL_SWITCH:
-                terminated = checkSwitch(ast, useContinue);
+                terminated =
+                        checkSwitch(ast, useContinue, labelsForCurrentSwitchScope);
                 break;
             case TokenTypes.LITERAL_SYNCHRONIZED:
-                terminated = checkSynchronized(ast, useBreak, useContinue);
+                terminated =
+                        checkSynchronized(ast, useBreak, useContinue, labelsForCurrentSwitchScope);
+                break;
+            case TokenTypes.LABELED_STAT:
+                labelsForCurrentSwitchScope.add(ast.getFirstChild().getText());
+                terminated =
+                        isTerminated(ast.getLastChild(), useBreak, useContinue,
+                                labelsForCurrentSwitchScope);
                 break;
             default:
                 terminated = false;
         }
         return terminated;
+    }
+
+    /**
+     * Checks if given break or continue ast has outer label.
+     *
+     * @param statement break or continue node
+     * @param labelsForCurrentSwitchScope the Set labels for the current scope of the switch
+     * @return true if local label used
+     */
+    private static boolean hasLabel(DetailAST statement, Set<String> labelsForCurrentSwitchScope) {
+        return Optional.ofNullable(statement)
+                .map(DetailAST::getFirstChild)
+                .filter(child -> child.getType() == TokenTypes.IDENT)
+                .map(DetailAST::getText)
+                .filter(label -> !labelsForCurrentSwitchScope.contains(label))
+                .isPresent();
     }
 
     /**
@@ -220,10 +254,11 @@ public class FallThroughCheck extends AbstractCheck {
      * @param slistAst SLIST to check
      * @param useBreak should we consider break as terminator
      * @param useContinue should we consider continue as terminator
+     * @param labels label names
      * @return true if SLIST is terminated.
      */
     private boolean checkSlist(final DetailAST slistAst, boolean useBreak,
-                               boolean useContinue) {
+                               boolean useContinue, Set<String> labels) {
         DetailAST lastStmt = slistAst.getLastChild();
 
         if (lastStmt.getType() == TokenTypes.RCURLY) {
@@ -236,7 +271,7 @@ public class FallThroughCheck extends AbstractCheck {
         }
 
         return lastStmt != null
-            && isTerminated(lastStmt, useBreak, useContinue);
+            && isTerminated(lastStmt, useBreak, useContinue, labels);
     }
 
     /**
@@ -246,17 +281,18 @@ public class FallThroughCheck extends AbstractCheck {
      * @param ast IF to check
      * @param useBreak should we consider break as terminator
      * @param useContinue should we consider continue as terminator
+     * @param labels label names
      * @return true if IF is terminated.
      */
     private boolean checkIf(final DetailAST ast, boolean useBreak,
-                            boolean useContinue) {
+                            boolean useContinue, Set<String> labels) {
         final DetailAST thenStmt = getNextNonCommentAst(ast.findFirstToken(TokenTypes.RPAREN));
 
         final DetailAST elseStmt = getNextNonCommentAst(thenStmt);
 
         return elseStmt != null
-                && isTerminated(thenStmt, useBreak, useContinue)
-                && isTerminated(elseStmt.getLastChild(), useBreak, useContinue);
+                && isTerminated(thenStmt, useBreak, useContinue, labels)
+                && isTerminated(elseStmt.getLastChild(), useBreak, useContinue, labels);
     }
 
     /**
@@ -279,9 +315,10 @@ public class FallThroughCheck extends AbstractCheck {
      * if allowed break, continue.
      *
      * @param ast loop to check
+     * @param labels label names
      * @return true if loop is terminated.
      */
-    private boolean checkLoop(final DetailAST ast) {
+    private boolean checkLoop(final DetailAST ast, Set<String> labels) {
         final DetailAST loopBody;
         if (ast.getType() == TokenTypes.LITERAL_DO) {
             final DetailAST lparen = ast.findFirstToken(TokenTypes.DO_WHILE);
@@ -291,7 +328,7 @@ public class FallThroughCheck extends AbstractCheck {
             final DetailAST rparen = ast.findFirstToken(TokenTypes.RPAREN);
             loopBody = rparen.getNextSibling();
         }
-        return isTerminated(loopBody, false, false);
+        return isTerminated(loopBody, false, false, labels);
     }
 
     /**
@@ -301,13 +338,15 @@ public class FallThroughCheck extends AbstractCheck {
      * @param ast loop to check
      * @param useBreak should we consider break as terminator
      * @param useContinue should we consider continue as terminator
+     * @param labels label names
      * @return true if try/catch/finally block is terminated
      */
     private boolean checkTry(final DetailAST ast, boolean useBreak,
-                             boolean useContinue) {
+                             boolean useContinue, Set<String> labels) {
         final DetailAST finalStmt = ast.getLastChild();
         boolean isTerminated = finalStmt.getType() == TokenTypes.LITERAL_FINALLY
-                && isTerminated(finalStmt.findFirstToken(TokenTypes.SLIST), useBreak, useContinue);
+                && isTerminated(finalStmt.findFirstToken(TokenTypes.SLIST),
+                useBreak, useContinue, labels);
 
         if (!isTerminated) {
             DetailAST firstChild = ast.getFirstChild();
@@ -317,7 +356,7 @@ public class FallThroughCheck extends AbstractCheck {
             }
 
             isTerminated = isTerminated(firstChild,
-                    useBreak, useContinue);
+                    useBreak, useContinue, labels);
 
             DetailAST catchStmt = ast.findFirstToken(TokenTypes.LITERAL_CATCH);
             while (catchStmt != null
@@ -325,7 +364,7 @@ public class FallThroughCheck extends AbstractCheck {
                     && catchStmt.getType() == TokenTypes.LITERAL_CATCH) {
                 final DetailAST catchBody =
                         catchStmt.findFirstToken(TokenTypes.SLIST);
-                isTerminated = isTerminated(catchBody, useBreak, useContinue);
+                isTerminated = isTerminated(catchBody, useBreak, useContinue, labels);
                 catchStmt = catchStmt.getNextSibling();
             }
         }
@@ -338,15 +377,18 @@ public class FallThroughCheck extends AbstractCheck {
      *
      * @param literalSwitchAst loop to check
      * @param useContinue should we consider continue as terminator
+     * @param labels label names
      * @return true if switch is terminated
      */
-    private boolean checkSwitch(final DetailAST literalSwitchAst, boolean useContinue) {
+    private boolean checkSwitch(DetailAST literalSwitchAst,
+                                boolean useContinue, Set<String> labels) {
         DetailAST caseGroup = literalSwitchAst.findFirstToken(TokenTypes.CASE_GROUP);
         boolean isTerminated = caseGroup != null;
         while (isTerminated && caseGroup.getType() != TokenTypes.RCURLY) {
             final DetailAST caseBody =
                 caseGroup.findFirstToken(TokenTypes.SLIST);
-            isTerminated = caseBody != null && isTerminated(caseBody, false, useContinue);
+            isTerminated = caseBody != null
+                    && isTerminated(caseBody, false, useContinue, labels);
             caseGroup = caseGroup.getNextSibling();
         }
         return isTerminated;
@@ -359,12 +401,13 @@ public class FallThroughCheck extends AbstractCheck {
      * @param synchronizedAst synchronized block to check.
      * @param useBreak should we consider break as terminator
      * @param useContinue should we consider continue as terminator
+     * @param labels label names
      * @return true if synchronized block is terminated
      */
     private boolean checkSynchronized(final DetailAST synchronizedAst, boolean useBreak,
-                                      boolean useContinue) {
+                                      boolean useContinue, Set<String> labels) {
         return isTerminated(
-            synchronizedAst.findFirstToken(TokenTypes.SLIST), useBreak, useContinue);
+            synchronizedAst.findFirstToken(TokenTypes.SLIST), useBreak, useContinue, labels);
     }
 
     /**
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheckTest.java
@@ -365,4 +365,30 @@ public class FallThroughCheckTest extends AbstractModuleTestSupport {
                 getPath("InputFallThrough8.java"),
                 expected);
     }
+
+    @Test
+    public void testLabeledBreak() throws Exception {
+        final String[] expected = {
+            "51:17: " + getCheckMessage(MSG_FALL_THROUGH),
+            "55:17: " + getCheckMessage(MSG_FALL_THROUGH),
+            "69:17: " + getCheckMessage(MSG_FALL_THROUGH),
+            "78:17: " + getCheckMessage(MSG_FALL_THROUGH),
+            "88:17: " + getCheckMessage(MSG_FALL_THROUGH),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputFallThroughLabeledBreak.java"),
+                expected);
+
+    }
+
+    @Test
+    public void testSwitchLabeledRules() throws Exception {
+        final String[] expected = {
+
+        };
+
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputFallThroughSwitchRules.java"),
+                expected);
+    }
 }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java
@@ -1,0 +1,104 @@
+/*
+FallThrough
+checkLastCaseGroup = (default)false
+reliefPattern = (default)falls?[ -]?thr(u|ough)
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.fallthrough;
+
+public class InputFallThroughLabeledBreak {
+
+    int method(int a, int b) {
+        top: switch (a) {
+            case 1:
+                switch (b) {
+                    case 10:
+                        break top; // ok as terminates outer 'top' switch
+                    default:
+                        return 11;
+                }
+            case 2:
+                return 2;
+        }
+        outer: outer2: switch (a) {
+            case 1:
+                inner1: inner2: switch (b) {
+                    case 10:
+                        break outer; // ok as terminates outer 'outer' switch
+                    default:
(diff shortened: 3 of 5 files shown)


nrmancuso:
@relentless-pursuit ping, do you plan to continue this PR?

relentless-pursuit:
> @relentless-pursuit ping, do you plan to continue this PR?

Yes, @nrmancuso. Will resume from 1st of Feb. Was caught up with something.

nrmancuso:
> @nrmancuso, in general i struggle with jacoco coverage. And, especially with this, i am struggling to create a jacoco coverage for the while loop i.e. `while (parent !=null)`. Can you provide me some guidance on this?

If we can’t hit that code, just remove it, let CI run, then generate check regression reports.

romani:
@relentless-pursuit , please don't open new PRs, let's finish hanging PRs like this

relentless-pursuit:
> @relentless-pursuit , please don't open new PRs, let's finish hanging PRs like this

Sure.

relentless-pursuit:
@romani, @nrmancuso can we proceed with the review?

relentless-pursuit:
I have used the example mentioned in the test case. Also, tried to replicate the issue with jar with code changes. I didn't see any error. Please let me know if something else needs to be done.

nrmancuso (changes requested):
@relentless-pursuit please do:
1. Generate regression reports
2. Extend test inputs to match https://github.com/checkstyle/checkstyle/pull/12820/files
3. Explain why the implementation has deviated significantly from https://github.com/checkstyle/checkstyle/pull/12820/files#r1141171046

relentless-pursuit:
> 3\. Explain why the implementation has deviated significantly from https://github.com/checkstyle/checkstyle/pull/12820/files#r1141171046

I was focussed on solving the issue based on the input test file. But, after i carefully read and re-read the PR conversation, i realised it is a bit complicated. I was just testing my code for test input mentioned in the issue. Now, I have been able to more or less understand what is aimed here. But, i have a concern. The test input as a part of this PR: https://github.com/checkstyle/checkstyle/pull/12820/files doesn't any longer fit to the recent code base of checksyle as we have changed the method `fallThrough` and done away with `matchesComment`. Hence, when i replicated the PR, and tested with recent Checkstyle master, it showed 2 Fall though violations. 

Functions like `matchesComment` to identify comments based on regular expressions and specific line numbers within the source code, which no longer exist. To my understanding the old PR was almost complete but maybe faced issue on these lines. It couldn't keep up with the changes made so far.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java
```diff
@@ -0,0 +1,101 @@
...
+                    }
+                case 10:  // violation 'Fall through from previous branch of the switch statement.'
+                    inner1:
+                    {
+                        switch (a) {
+                            case 1:
+                                break outer; // ok
+                        }
+                    }
+                case 11:
```
relentless-pursuit, line 95: If we observe, in the ```
Violation lines for /Users/piyushkumarsadangi/Development/checkstyle/checkstyle/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java differ.
unexpected (2): 37, 95
---
expected      : [50, 54, 68, 77, 87]
but was       : [37, 50, 54, 68, 77, 87, 95]

```

Now, ideally we shouldn't be getting any violation at this line: `case 11:`. As we exit the `outer:switch (a)`. However, we continue to get a violation now, which must not have been the case when the old PR was being authored, because the code base has undergone some changes. Now, the question remains do we revert back to old changes, or have a fusion of both worlds, line-number based approach and AST. I am not sure on this. Please let me know if i have understood right. I will continue to understand the existing code, and maybe a change here and there can still get us no violation at 37 and 95. I hope I didn't complicate with my observations.

**Comment by nrmancuso** [comment]
@relentless-pursuit thanks a lot for the explanation; I had completely forgotten about the overhaul to this check. I hope that it makes your life easier in this PR :) Let's generate a regression report while I begin review.

File: src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java (hunk 1 above)
Comment (nrmancuso) [reply src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java:95]: Yes, 37 and 95 should not be violations.

**Comment by relentless-pursuit** [comment]
@nrmancuso, I have been debugging for quite some time to see why 37 and 95 are false-positive. And, what i figured out is that they don't go inside `case TokenTypes.SLIST:` inside the method `isTerminated`. Hence, we don't get to execute the `hasOuterLabels` method leading to false positive. 

Now, my understanding of why it goes inside the `case TokenTypes.SLIST:` and `case TokenTypes.LITERAL_BREAK:` is 
when visitToken encounters a switch statement, it examines each case group. The SLIST under a case group includes all the statements related to that case. SLIST acts as the parent node for a sequence of statements. When a `break` statement (`LITERAL_BREAK`) appears directly within the `SLIST` of a case, it is considered a direct child of that `SLIST`. In such scenarios, the `break` statement successfully prevents fall-through. 

Example:
```
case 1:
    switch (b) {
        case 10:
            break top; // ok
        default:
            return 11;
    }


But, if a `break` statment is inside another strcuture like `for` loop in the example, the `for` loop itself becomes the direct child of the `SLIST` associated with the case. The `break` statement is a child of the for loop structure, not the `SLIST` directly. Hence, not directly processed under `LITERAL_BREAK` by the switch-case logic because it's considered part of the for loop's structure, which is inside the `SLIST` like in below example. 

```
        outer: outer2: switch (a) {
            case 1:
                inner1: inner2: switch (b) {
                    case 10:
                        break outer; // ok
                    default:
                        return 11;
                }
            case 2:
                inner: for (int i = 0;i < 3;i++) {
                    break outer2; // ok
                }
            case 3:
                return 2; **// We experience here false-positive**
        }
```

To summarise, `break` statement if directly within the case's statement list (not nested in other structures) it's processed as a `LITERAL_BREAK` in the AST.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -182,61 +186,83 @@ private boolean isTerminated(final DetailAST ast, boolean useBreak,
...
+
+    /**
+     * Checks if given break or continue ast has outer label.
+     *
+     * @param ast break or continue node
+     * @param labels label names in scope of case group
+     * @return true if local label used
+     */
+    private static boolean hasOuterLabel(DetailAST ast, Set<String> labels) {
+        return ast.getChildCount() == 2 && !labels.contains(ast.getFirstChild().getText());
```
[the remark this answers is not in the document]
relentless-pursuit: > Please extract `ast.getChildCount() == 2` to some meaningfully named variable so that the next person to code in here knows why this matters.

Specifically, on observing for the below piece fo code which leads to Fall-through false positive, the child count is 3. 
```
 case 2:
                inner: for (int i = 0;i < 3;i++) {
                    break outer2; // ok
                }
```

Some logs from the console:
```
Child 0: Type = 86, Text = break
Child 1: Type = 144, Text = //
Child 2: Type = 73, Text = }
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -246,17 +272,18 @@ private boolean checkSlist(final DetailAST slistAst, boolean useBreak,
...
+                            boolean useContinue, Set<String> labels) {
         final DetailAST thenStmt = getNextNonCommentAst(ast.findFirstToken(TokenTypes.RPAREN));
 
         final DetailAST elseStmt = getNextNonCommentAst(thenStmt);
 
         return elseStmt != null
-                && isTerminated(thenStmt, useBreak, useContinue)
-                && isTerminated(elseStmt.getLastChild(), useBreak, useContinue);
+                && isTerminated(thenStmt, useBreak, useContinue, labels)
+                && isTerminated(elseStmt.getFirstChild(), useBreak, useContinue, labels);
```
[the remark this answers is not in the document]
relentless-pursuit: The the author of the old PR and some review comments that followed in the old PR agreed on tracing labels for 
`continue` statement as well. Hence, the scope of the PR is not just limited to break but `continue` aswell. 

Reference: https://github.com/checkstyle/checkstyle/pull/12820#discussion_r1141009389

If you find it not required for this PR, i will remove it.

relentless-pursuit:
> @relentless-pursuit thanks a lot for the explanation; I had completely forgotten about the overhaul to this check. I hope that it makes your life easier in this PR :) Let's generate a regression report while I begin review.

I feel the existing code is not sufficient. We still have some work to be done. And, that might break something from past. But, I will proceed with regression report generation.

nrmancuso:
@relentless-pursuit how's it going here? Can we help with something?

relentless-pursuit:
> @relentless-pursuit how's it going here? Can we help with something?

@nrmancuso Sorry for the delay. I have generated a regression report. I did it after long time. let me know if i have done it right, or if the report looks fine. Meanwhile, I had posted some concerns here: https://github.com/checkstyle/checkstyle/pull/14199#issuecomment-1971802298

Please let me know your feedback on the same.

nrmancuso:
> @nrmancuso sorry for the ping. Any feedback on the development so far on this PR?

No need to be sorry :) Feedback from me is that first we need to achieve expected check behavior in new test cases, with no failures in CI. Then we need to generate a regression report. If this is clean, we proceed with review and code refinement. If not, we found new test cases to add and work on. Then rinse, repeat :)

relentless-pursuit:
> > @nrmancuso sorry for the ping. Any feedback on the development so far on this PR?
> 
> No need to be sorry :) Feedback from me is that first we need to achieve expected check behavior in new test cases, with no failures in CI. Then we need to generate a regression report. If this looks good, we proceed with review and code refinement. If not, we found new test cases to add and work on. Then rinse, repeat :)

@nr

> > @nrmancuso sorry for the ping. Any feedback on the development so far on this PR?
> 
> No need to be sorry :) Feedback from me is that first we need to achieve expected check behavior in new test cases, with no failures in CI. Then we need to generate a regression report. If this looks good, we proceed with review and code refinement. If not, we found new test cases to add and work on. Then rinse, repeat :

> @nrmancuso, I have been debugging for quite some time to see why 37 and 95 are false-positive. And, what i figured out is that they don't go inside `case TokenTypes.SLIST:` inside the method `isTerminated`. Hence, we don't get to execute the `hasOuterLabels` method leading to false positive.
> 
> Now, my understanding of why it goes inside the `case TokenTypes.SLIST:` and **NOT** `case TokenTypes.LITERAL_BREAK:` is when visitToken encounters a switch statement, it examines each case group. The SLIST under a case group includes all the statements related to that case. `SLIST` acts as the parent node for a sequence of statements.
> 
> When a `break` statement (`LITERAL_BREAK`) appears directly within the `SLIST` of a case, it is considered a direct child of that `SLIST`. In such scenarios, the `break` statement successfully prevents fall-through.
> 
> ```
> case 1:
>     switch (b) {
>         case 10:
>             break top; // ok
>         default:
>             return 11;
>     }
> ```
> 
> But, if a `break` statment is inside another strcuture like `for` loop in the example, the `for` loop itself becomes the direct child of the `SLIST` associated with the case. The `break` statement is a child of the `for` loop structure, not the `SLIST` directly. Hence, not directly processed under `LITERAL_BREAK` by the switch-case logic because it's considered part of the for loop's structure, which is inside the `SLIST` like in below example.
> 
> ```
>         outer: outer2: switch (a) {
>             case 1:
>                 inner1: inner2: switch (b) {
>                     case 10:
>                         break outer; // ok
>                     default:
>                         return 11;
>                 }
>             case 2:
>                 inner: for (int i = 0;i < 3;i++) {
>                     break outer2; // ok
>                 }
>             case 3: // We experience here false-positive by encountering violation of Fall through
>                 return 2; 
>         }
> ```
> 
> To summarise, `break` statement if directly within the case's statement list (not nested in other structures) it's processed as a `LITERAL_BREAK` in the Checkstyle's existing code.

Kindly ignore. After careful debugging, I realised that `37` and `95` were given false-positive because of not being able to realise break statement because of the // ok comment. Hence, altered the input test file by adding the `// ok` comment to as `// ok below as ...`

relentless-pursuit:
> > @nrmancuso sorry for the ping. Any feedback on the development so far on this PR?
> 
> No need to be sorry :) Feedback from me is that first we need to achieve expected check behavior in new test cases, with no failures in CI. Then we need to generate a regression report. If this looks good, we proceed with review and code refinement. If not, we found new test cases to add and work on. Then rinse, repeat :)

@nrmancuso, updated the regression report. I think the CI failure is unrelated. Please let me know if it is good enough to begin review.

relentless-pursuit:
@nrmancuso ping

nrmancuso:
> @nrmancuso ping

Need a bit of time to get to this one, thanks for the reminder

relentless-pursuit:
> > @nrmancuso ping
> 
> Need a bit of time to get to this one, thanks for the reminder

Sure, will wait.

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheckTest.java
```diff
@@ -365,4 +365,20 @@ public void testReliefCommentBetweenMultipleComment() throws Exception {
                 getPath("InputFallThrough8.java"),
                 expected);
     }
+
+    @Test
```
nrmancuso, line 369: Please add another test case with similar inputs, except with switch labeled rules. See https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-SwitchRule

nrmancuso (changes requested):
Looking good, please resolve/mark outdated all items (in conversation) from you that have been addressed

Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -182,48 +185,65 @@ private boolean isTerminated(final DetailAST ast, boolean useBreak,
...
 
+    /**
+     * Checks if given break or continue ast has outer label.
+     *
+     * @param ast break or continue node
+     * @param labels label names in scope of case group
+     * @return true if local label used
+     */
+    private static boolean hasOuterLabel(DetailAST ast, Set<String> labels) {
+        final boolean childCount = ast.getChildCount() == 2;
```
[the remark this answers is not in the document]
relentless-pursuit: > Why do we have this condition?

The name could be deceptive. But, this method helps us to avoid fall-through violation. 

```
        top: switch (a) {
            case 1:
                switch (b) {
                    case 10:
                        break top; // ok as terminates outer 'top' switch
                    default:
                        return 11;
                }
            case 2:
                return 2;
        }
```

In the above scenario,` hasOuterLabel` helps us to come out of the entire scope of `top`. And, this prevents fall-through as we don't fall onto case 2 after `break top; `

```
        top: switch (a) {
            case 1:
                inner: switch (b) {
                    case 10:
                        break inner;                     
                        return 11;
                }
            case 2: // Fall through
                return 2;
        }
```

Now, here, we have again a labeled break. We break out of scope of inner switch block and control comes to case 2 causing Fall throgh violation. In nutshell, this method only helps in detecting a labeled break, be it inner or outer.

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheckTest.java
relentless-pursuit: > Please add another test case with similar inputs, except with switch labeled rules. See https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-SwitchRule

I am a bit confused. I am reading the documentation. And, it mentions of SwitchRule and SwitchLabel. 

```
int method(int a, int b) {
    top:
    switch (a) {
        case 1 -> {
            switch (b) {
                case 10 -> { break top; } // ok as terminates outer 'top' switch
                default -> { return 11; }
            }
        }
        case 2 -> { return 2; }
    }
}

```

Something like this? Like replacing with `->` ?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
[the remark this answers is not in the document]
nrmancuso: fwiw, I doubt that your second example would compile; we will never reach the `return 11;` statement.


> The name could be deceptive.

Yes, it is. This is not good. 

>childCount

This sounds like it should be a numerical value.


Please do something like this to make it very clear what is happening here:
```suggestion

    private static boolean hasOuterLabel(DetailAST statement, Set<String> labels) {
        final boolean isLabeledStatement = ast.getChildCount() == 2;
        final String identifier = ast.getFirstChild().getText
        return childCount && !labels.contains(identifier);
    }
````

Or to be even more clear and avoid possibility of  `;` being called an identifier or a possible NPE, we could do something like this:
```suggestion

    private static boolean hasOuterLabel(DetailAST statement, Set<String> labels) {       
        return Optional.of(statement)
            .map(DetailAST::getFirstChild)  // get first child
            .filter(child -> child.getType() == TokenTypes.IDENT) // see if it is an identifier
            .map(DetailAST::getText) // if so, get the name
            .filter(label -> !labels.contains(label)) // see if this label is in labels
            .isPresent();
    }

```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -169,10 +171,11 @@ public void visitToken(DetailAST ast) {
      * @param ast root of given subtree
      * @param useBreak should we consider break as terminator
      * @param useContinue should we consider continue as terminator
+     * @param labels label names
      * @return true if the subtree is terminated.
      */
     private boolean isTerminated(final DetailAST ast, boolean useBreak,
-                                 boolean useContinue) {
+                                 boolean useContinue, Set<String> labels) {
```
nrmancuso, line 178: Each Set of labels is only for a given "level" of switch, right?

Example of my understanding:
```
        top: switch (a) { // 'top' is in one Set of labels
            case 1:
                inner: switch (b) { // 'inner' is in a different Set of labels
                    case 10:
                        break inner;                     
                }
            case 2: // Fall through
                return 2;
        }
```

If my understanding is correct, can we somehow make this obvious? Perhaps improve the Javadoc, and rename this Set to something that makes this obvious?

nrmancuso (changes requested):
Item:

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheckTest.java
nrmancuso: > Like replacing with -> ?

Basically, yes. There are some different AST elements around this, so it will be good to make sure we are covered there.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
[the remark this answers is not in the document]
relentless-pursuit: > fwiw,

Yes, it overlooked it. but, the idea was to explain what haOuterLabel does. it is not a part of the PR.

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheckTest.java
nrmancuso: Just wanted to be clear here, there is no possibility of fallthrough with this syntax, I just want to make sure there is no possibility of NPE.

relentless-pursuit: > Basically, yes. There are some different AST elements around this, so it will be good to make sure we are covered there.

Have added a test-case and corresponding input file using switch rules. Let me know your thoughts on the same.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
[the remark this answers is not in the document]
relentless-pursuit: > fwiw, I doubt that your second example would compile; we will never reach the `return 11;` statement.
> 
> > The name could be deceptive.
> 
> Yes, it is. This is not good.
> 
> > childCount
> 
> This sounds like it should be a numerical value.
> 
> Please do something like this to make it very clear what is happening here:
> 
> Or to be even more clear and avoid possibility of `;` being called an identifier or a possible NPE, we could do something like this:

Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
relentless-pursuit: > Each Set of labels is only for a given "level" of switch, right?
> 
> Example of my understanding:
> 
> ```
>         top: switch (a) { // 'top' is in one Set of labels
>             case 1:
>                 inner: switch (b) { // 'inner' is in a different Set of labels
>                     case 10:
>                         break inner;                     
>                 }
>             case 2: // Fall through
>                 return 2;
>         }
> ```
> 
> If my understanding is correct, can we somehow make this obvious? Perhaps improve the Javadoc, and rename this Set to something that makes this obvious?

Yes, you are right. I have made changes to the the name.

nrmancuso (changes requested):
Please make CI happy, we are almost there:

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheckTest.java
nrmancuso: Looks good. Thanks!

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -182,48 +187,78 @@ private boolean isTerminated(final DetailAST ast, boolean useBreak,
...
     }
 
+    /**
+     * Checks if given break or continue ast has outer label.
+     *
+     * @param ast break or continue node
+     * @param labelsForCurrentSwitchScope the Set labels for the current scope of the switch
+     * @return true if local label used
+     */
+    private static boolean hasLabels(DetailAST ast, Set<String> labelsForCurrentSwitchScope) {
```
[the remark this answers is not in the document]
relentless-pursuit: Yes, thanks for pointing it out. Corrected.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -182,48 +187,78 @@ private boolean isTerminated(final DetailAST ast, boolean useBreak,
...
+        return Optional.ofNullable(ast)
+                .filter(node -> node.getChildCount() == 2)
+                .map(node -> node.getFirstChild())
```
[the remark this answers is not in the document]
relentless-pursuit: Done. Waiting for CI's status.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -182,48 +187,78 @@ private boolean isTerminated(final DetailAST ast, boolean useBreak,
...
+     * @param statement break or continue node
+     * @param labelsForCurrentSwitchScope the Set labels for the current scope of the switch
+     * @return true if local label used
+     */
+    private static boolean hasLabel(DetailAST statement, Set<String> labelsForCurrentSwitchScope) {
+        return Optional.ofNullable(statement)
+                .filter(node -> node.getChildCount() == 2)
+                .map(DetailAST::getFirstChild)
+                .map(DetailAST::getText)
+                .map(text -> !labelsForCurrentSwitchScope.contains(text))
```
nrmancuso, line 246:
```suggestion
                .map(label -> !labelsForCurrentSwitchScope.contains(label))
```

nrmancuso (changes requested):
One more minor one:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
nrmancuso: @relentless-pursuit was there an issue with filtering by TokenTypes.IDENT as suggested in https://github.com/checkstyle/checkstyle/pull/14199#discussion_r1525031818?

relentless-pursuit: @nrmancuso I missed it somehow. However, i see failed cases, and more violations expected than it is now. 

```
    private static boolean hasLabel(DetailAST statement, Set<String> labelsForCurrentSwitchScope) {
        return Optional.ofNullable(statement)
                .filter(child -> child.getType() == TokenTypes.IDENT)
                .map(DetailAST::getText)
                .map(text -> !labelsForCurrentSwitchScope.contains(text))
                .orElse(Boolean.FALSE);
    }
```

```
Violation lines for /Users/piyushkumarsadangi/Development/checkstyle/checkstyle/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughLabeledBreak.java differ.
unexpected (7): 22, 33, 38, 64, 74, 83, 98
---
expected      : [51, 55, 69, 78, 88]
but was       : [22, 33, 38, 51, 55, 64, 69, 74, 78, 83, 88, 98]
```

```
    int method(int a, int b) {
        top: switch (a) {
            case 1:
                switch (b) {
                    case 10:
                        break top; // ok as terminates outer 'top' switch
                    default:
                        return 11;
                }
            case 2: // Expecting a Fall-through violation here, which shouldn't be the case
                return 2;
        }
        
        ...
```

relentless-pursuit: @nrmancuso , please ignore my above comment. I hurriedly tested without carefully observing your comment. 
Yes, it seems to be working. 

```
    private static boolean hasLabels(DetailAST statement, Set<String> labels) {
        return Optional.of(statement)
                .map(DetailAST::getFirstChild)  // get first child
                .filter(child -> child.getType() == TokenTypes.IDENT) // see if it is an identifier
                .map(DetailAST::getText) // if so, get the name
                .filter(label -> !labels.contains(label)) // see if this label is in labels
                .isPresent();
    }
```

I will push my code changes again with your suggestions. Thanks for your patience.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -182,48 +187,78 @@ private boolean isTerminated(final DetailAST ast, boolean useBreak,
...
+     * @param labelsForCurrentSwitchScope the Set labels for the current scope of the switch
+     * @return true if local label used
+     */
+    private static boolean hasLabel(DetailAST statement, Set<String> labelsForCurrentSwitchScope) {
+        return Optional.ofNullable(statement)
+                .map(DetailAST::getFirstChild)
+                .filter(child -> child.getType() == TokenTypes.IDENT)
+                .map(DetailAST::getText)
+                .map(label -> !labelsForCurrentSwitchScope.contains(label))
+                .orElse(Boolean.FALSE);
```
nrmancuso, line 247:
```suggestion
                .ifPresent();
```

is cleaner and should make checker happy

relentless-pursuit: I actually had isPresent() but the I guess one of the CI job failed, stating some issue. I can't find it. And, had to look for something else. Will try to revert it.

nrmancuso: I am good to suppress checker violation for this

relentless-pursuit:
> @relentless-pursuit please make CI happy, then I will hand off to the next reviewer. Thanks for this!

@nrmancuso, It tried looking into the logs, but couldn't find why it is failing. Though, i can see 2 warnings for `FallThrougCheck.java` which is related to `Optional` use, one of them being used before this PR. And, I did add the suppression too. Can you guide me where in the logs, it is leading to failure? Is it failing for some threshold level of warnings allowed, and we have a new warning added due to `hasLabel` `Optional` approach.

nrmancuso:
Let’s try to apply patch one more time, it looks like ordering changed

relentless-pursuit:
> Let’s try to apply patch one more time, it looks like ordering changed

I tried that by placing suppressions of `FallThroughCheck.java` together. But, the failure persists.

nrmancuso:
> > Let’s try to apply patch one more time, it looks like ordering changed
> 
> I tried that by placing suppressions of `FallThroughCheck.java` together. But, the failure persists.

you should only `git apply …` the patch that is provided in the failing CI job

relentless-pursuit:
> > > Let’s try to apply patch one more time, it looks like ordering changed
> > 
> > 
> > I tried that by placing suppressions of `FallThroughCheck.java` together. But, the failure persists.
> 
> you should only `git apply …` the patch that is provided in the failing CI job

The failing job is suceeding now. I assumed patch meant something else, i.e. hiding the suppresion. However, I wasn't aware of git apply. I learnt something new. Thank you.

relentless-pursuit:
@nrmancuso ping

nrmancuso (approved):
Good stuff!

relentless-pursuit:
@romani ping

romani (approved):
ok to merge

PR #14205, opened 2023-12-31
infra: enable spotbugs

romani:
by request from https://github.com/spotbugs/spotbugs/issues/2759?notification_referrer_id=NT_kwDOAAxnuLE4NzA2MDcxNTc2OjgxMjk4NA#issuecomment-1856012271

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,8 @@ jobs:
           command: |
             ./.ci/validation.sh git-diff
             ./.ci/validation.sh ci-temp-check
+      - store_artifacts:
+          path: target
       - save_cache:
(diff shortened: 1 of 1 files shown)


rnveach:
https://github.com/marketplace/actions/tar-action
https://github.com/marketplace/actions/action-build-zip

romani:
From that issue

>So until the issue is fixed in BCEL (and maybe SpotBugs) I think that you should try a find a way to run SpotBugs before Jacoco

We can probably skip jacoco when we run spotbug in such CI job.
https://www.eclemma.org/jacoco/trunk/doc/instrument-mojo.html#skip

romani:
spotbugs team found a problem

PR #14209, opened 2023-12-31
Issue #13693: migrate misc modules to property macros

romani:
Issue #13693

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/TranslationCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/TranslationCheck.java
@@ -93,6 +93,11 @@ import com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil;
  * of default translation files in project.
  * </p>
  * <p>
+ * Note: If your project uses preprocessed translation files and the original files do not have the
+ * {@code properties} extension, you can specify additional file extensions
+ * via the {@code fileExtensions} property.
+ * </p>
+ * <p>
  * Attention: the check will perform the validation of ISO codes if the option
  * is used. So, if you specify, for example, "mm" for language code,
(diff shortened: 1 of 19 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/TranslationCheck.java
```diff
@@ -113,10 +113,7 @@
  * Default value is {@code "^messages.*$"}.
  * </li>
  * <li>
- * Property {@code fileExtensions} - Specify file type extension to identify
- * translation files. Setting this property is typically only required if your
- * translation files are preprocessed and the original files do not have
- * the extension {@code .properties}
+ * Property {@code fileExtensions} - Specify the file extensions of the files to process.
```
romani, line 116: Even we loose a bit of details, I still think it is ok. I do not think this is critical details.

nrmancuso: I do not agree here. Why do we need to change this?

romani: Because content comes from abstract class method, it is same for many other Checks.

nrmancuso:
```java
 * <p>
 * Note: If your project uses preprocessed translation files and the original files do not have the
 * {@code properties} extension, you can specify additional file extensions via the {@code fileExtensions} property.
 * </p>


```

nrmancuso: Please add this to check description then ^

romani: done

nrmancuso (approved):

PR #14229, opened 2024-01-05
Issue #14137: Enable `CollectorMutability` check

rickie:
Issue https://github.com/checkstyle/checkstyle/issues/14137.

This PR enables the CollectorMutablility check, see [documentation](https://error-prone.picnic.tech/bugpatterns/CollectorMutability/).

Initially this was brought up in https://github.com/checkstyle/checkstyle/pull/12777#pullrequestreview-1325151741.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -238,6 +238,8 @@
     <error-prone-support.version>0.14.0</error-prone-support.version>
     <doxia.version>1.12.0</doxia.version>
     <error-prone.configuration-args>
+      <!-- Prefer usage of immutable collectors. -->
+      -Xep:CollectorMutability:ERROR
       <!-- Reason at https://github.com/checkstyle/checkstyle/issues/8252. -->
       -Xep:JUnitClassModifiers:OFF
       <!-- Reason at https://github.com/checkstyle/checkstyle/issues/8252. -->
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
@@ -32,12 +32,13 @@ import java.util.Locale;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.puppycrawl.tools.checkstyle.api.AuditEvent;
 import com.puppycrawl.tools.checkstyle.api.AuditListener;
 import com.puppycrawl.tools.checkstyle.api.BeforeExecutionFileFilter;
@@ -222,7 +223,7 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
 
         final List<File> targetFiles = files.stream()
                 .filter(file -> CommonUtil.matchesFileExtension(file, fileExtensions))
-                .collect(Collectors.toList());
+                .collect(ImmutableList.toImmutableList());
         processFiles(targetFiles);
 
         // Finish up
@@ -251,7 +252,7 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
                 return ((ExternalResourceHolder) resource)
                         .getExternalResourceLocations().stream();
             })
-            .collect(Collectors.toSet());
+            .collect(ImmutableSet.toImmutableSet());
     }
 
     /** Notify all listeners about the audit start. */
--- a/config/checker-framework-suppressions/checker-lock-tainting-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-lock-tainting-suppressions.xml
@@ -100,7 +100,7 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/PropertiesExpander.java</fileName>
     <specifier>methodref.receiver.bound</specifier>
     <message>Incompatible receiver type</message>
-    <lineContent>.collect(Collectors.toMap(Function.identity(), properties::getProperty));</lineContent>
+    <lineContent>.collect(ImmutableMap.toImmutableMap(Function.identity(), properties::getProperty));</lineContent>
     <details>
       found   : @GuardedBy Properties
       required: @GuardSatisfied Properties
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -775,7 +775,7 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/PropertiesExpander.java</fileName>
     <specifier>methodref.return</specifier>
     <message>Incompatible return type</message>
-    <lineContent>.collect(Collectors.toMap(Function.identity(), properties::getProperty));</lineContent>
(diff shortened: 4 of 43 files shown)


pom.xml
```diff
@@ -238,6 +238,8 @@
     <error-prone-support.version>0.14.0</error-prone-support.version>
     <doxia.version>1.12.0</doxia.version>
     <error-prone.configuration-args>
+      <!-- Prefer usage of immutable collectors. -->
```
rickie, line 241: Open question; do we have a section with disabled and enabled checks? Or do we sort them lexicographically (which is what I did right now).

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -222,7 +223,7 @@ public int process(List<File> files) throws CheckstyleException {
 
         final List<File> targetFiles = files.stream()
                 .filter(file -> CommonUtil.matchesFileExtension(file, fileExtensions))
-                .collect(Collectors.toList());
+                .collect(ImmutableList.toImmutableList());
```
rickie, line 226: Opened a PR to get some feedback as the `mvn clean install` failed locally. The `Immutable*` import was not allowed. Isn't it and should I change this? Or do we need to add this somewhere such that it is allowed?

rickie (commented):
Some context and questions.

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
rickie: Example of what I mean: https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=18567&view=logs&j=e663805c-ee13-5a24-f979-d725e0346c8e&t=40ff6e43-c5d7-5753-f947-23012624b0c6&l=1975.

pom.xml
romani: Sorting is appreciated.

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
romani: Long time ago we decided to reduce dependency to guava, as there was was problems in some plugins. I even don't remember root reason.

Can we archive the same by jdk methods ?

If not possible, we can reconsider old decision on guava, but it will take a bit longer to accept, as I need to recheck all old problems to be sure other will not be affected.

rickie: Hmm, interesting. Internally we indeed like Guava and use these methods (although if there is a good JDK alternative we prefer that). In the `CollectorMutability` check we have the [following code](https://github.com/PicnicSupermarket/error-prone-support/blob/a2f44f82f2c815450c4fc3ee8b35452c03d053f1/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CollectorMutability.java#L61):
```
if (!ThirdPartyLibrary.GUAVA.isIntroductionAllowed(state)
```
This piece of code basically makes the check do nothing as it doesn't want to introduce new Guava dependencies. Maybe we can introduce a different collector that still clearly communicates the immutability. It would not be something we prefer internally but would make sense if the introduction of new Guava dependencies is not desired.

I was playing around and found the `Collectors.toUnmodifiableList()` alternative to `ImmutableList.toImmutableList()`. I pushed a commit with a change where I applied this. Curious to hear what you think.

If we like the suggestion we can consider updating Error Prone Support to have that as alternative. We are always open to discussing improvements to Error Prone Support such that our checks are more useful to others.

Tagging @Stephan202 to get his opinion on this change proposal.

Stephan202: Yeah, I think that it makes sense to suggest `Collectors.toUnmodifiable{List,Map,Set}` when `!ThirdPartyLibrary.GUAVA.isIntroductionAllowed(state)`, instead of backing out early like we do now. (We should then add a guard to prevent suggesting those variants when the build targets JDK < 10.)

romani:
Just jdk methods is much better than guava.


This PR is going complicated, we got conflict between tools https://app.circleci.com/pipelines/github/checkstyle/checkstyle/22723/workflows/15aec14e-3fa8-4c87-ba69-5f2f37c0e3e5/jobs/467800/artifacts

Pitest and Checkers suppression just need to be updated for new line content.

rickie:
> This PR is going complicated, we got conflict between tools

I think the tools correctly flagged something. We used the `toUnmodifiableSet` twice so I updated the code. It should be fine like this. I'll take a look at the errors and see if we can get a green build.

@Stephan202 [agrees](https://github.com/checkstyle/checkstyle/pull/14229#discussion_r1443505966) with my proposal, so I'll create a issue or PR in Error Prone Support and link it here.

Till that time, do we want to disable the check and link the issue I'll create just like the other checks we disabled? Then we can at least merge these changes and add the check in a separate PR if we released the fix. Let me know what works for you guys.

rickie:
API rate limit exceeded 😅.

romani:
idea still not happy: https://app.circleci.com/pipelines/github/checkstyle/checkstyle/22740/workflows/8162acc2-64a6-4dc1-adcc-e6db56da86c3/jobs/468544/artifacts
```
<problem>
<file>
file://$PROJECT_DIR$/src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
</file>
<line>61</line>
<module>project</module>
<package>com.puppycrawl.tools.checkstyle.site</package>
<entry_point TYPE="field" FQNAME="com.puppycrawl.tools.checkstyle.site.PropertiesMacro NON_BASE_TOKEN_PROPERTIES"/>
<problem_class id="PublicStaticCollectionField" severity="ERROR" attribute_key="ERRORS_ATTRIBUTES">'public static' collection field</problem_class>
<description>
'public static' collection field <code>NON_BASE_TOKEN_PROPERTIES</code>, compromising security #loc
</description>
<highlighted_element>NON_BASE_TOKEN_PROPERTIES</highlighted_element>
<language>JAVA</language>
<offset>36</offset>
<length>25</length>
</problem>
```

romani:
>Till that time, do we want to disable the check and link the issue I'll create just like the other checks we disabled?

can we reuse our fancy dynamic suppressions ?
 https://github.com/checkstyle/checkstyle/blob/720ad558e82d9a6c357ec11325601da5331fe736/config/error-prone-suppressions/compile-phase-suppressions.xml#L3-L8

here is how we run it https://github.com/checkstyle/checkstyle/blob/720ad558e82d9a6c357ec11325601da5331fe736/.github/workflows/error-prone.yml#L38 , so it will upadate file and you just add change it commit.

romani (approved):
Ok to merge.

Please share feedback on why you think it is better to keep whole rule disabled only because of one edge case and not to deal with our suppression file?

rickie:
Ah sorry, I misread that, I'm also fine with enabling the check and suppressing that one occurrence. Will update the PR.

romani:
https://github.com/checkstyle/checkstyle/actions/runs/7450003885/job/20267925289?pr=14229#step:6:334

```
Description: "Avoid `Collectors.to{List,Map,Set}` in favor of collectors that emphasize (im)mutability"
```
Vs
```
Description: "Avoid Collectors.to{List,Map,Set} in favor of collectors that emphasize (im)mutability"
```

There is quotes over collection name, it is a diff

rickie:
Done, sorry I didn't follow-up directly, was busy at work :).

romani (approved):
Ok to merge

rnveach (approved):

rnveach:
@rickie There is a conflict.

PR #14230, opened 2024-01-05
Issue #14137: Enable `AmbiguousJsonCreator` check

rickie:
Issue #14137.

Based on the answers in #14229, I will update this PR accordingly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -238,6 +238,7 @@
     <error-prone-support.version>0.14.0</error-prone-support.version>
     <doxia.version>1.12.0</doxia.version>
     <error-prone.configuration-args>
+      -Xep:AmbiguousJsonCreator:ERROR
       <!-- Reason at https://github.com/checkstyle/checkstyle/issues/8252. -->
       -Xep:JUnitClassModifiers:OFF
       <!-- Reason at https://github.com/checkstyle/checkstyle/issues/8252. -->


romani (approved):
Ok to merge if ci pass

pom.xml
```diff
@@ -238,6 +238,8 @@
     <error-prone-support.version>0.14.0</error-prone-support.version>
     <doxia.version>1.12.0</doxia.version>
     <error-prone.configuration-args>
+      <!-- `JsonCreator.Mode` should be set for single-argument creators. -->
+      -Xep:AmbiguousJsonCreator:ERROR
```
rnveach, line 242: Do we really need to write a reason for each type of check we are turning on? From the main issue this will be a long list, and we don't even document our own checks on why we are turning them on. We only do this if we are turning them off.

rickie: This is completely up to your preference 😄. I'm fine with both options, we can simply omit it.

rnveach: @romani

romani: I vote to skip comment.
We do comment only when we disable rule , OFF .

rickie: Deal, makes sense. Will update!

rickie:
@romani @rnveach Can you help me understand what the problem is? I'm not sure how I can get to the actual error?

rnveach:
@romani no exception CI failed with no reason or error.
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/22732/workflows/6464b014-40c3-46e4-ab5f-d8cc32953a5c/jobs/468173

rnveach (approved):

Issue #6207, opened 2018-11-10
Expand XPath IT Regression Testing


rnveach:
Identified at https://github.com/checkstyle/checkstyle/pull/6198#issuecomment-437386863 ,

We need to expand XPath IT Regression to all checks that support it.

All checks listed under [Currently, filter supports the following checks](http://checkstyle.sourceforge.net/config_filters.html#SuppressionXpathFilter_Description) must appear with regression tests in https://github.com/checkstyle/checkstyle/tree/master/src/it/java/org/checkstyle/suppressionxpathfilter to showcase suppression by XPath works for the check.
Every log statement must be tested from the Check, so if 1 check has 2+ log statements, the XPath test must have 2+ tests for each message.

Tests can be copy of examples from our normal test tier at https://github.com/checkstyle/checkstyle/tree/master/src/test/java/com/puppycrawl/tools/checkstyle/checks just smaller.

romani:
@varunu28 , can you help us with this issue ?
It is simple but it will good start to understand how actual development should be done -through test. Test driven development.

varunu28:
@romani @rnveach Should I be working on this by making 1 PR for every test?

rnveach:
@varunu28 That is fine with us for the first few. Every PR should be isolated, so you can open multiple PRs after the first few to make sure we get the format right.

romani:
yes, simple PR are easy to craft, easy to validate and merge quickly, If any nuance happens , discussion is started, it does not block all other changes/PRs.

rnveach:
@varunu28 has stated he has no more time for this issue and to assign it to someone else.

mincong-h:
Hi, is this issue still valid? I can contribute some of these missing tests.

rnveach:
It is still valid as long as we are missing these tests.
You are free to contribute.

mincong-h:
Before moving further on this ticket, I just want to make an exhaustive list of checks to be tested. So that we are all aware of the test effort. They are extracted from https://checkstyle.sourceforge.io/config_filters.html#SuppressionXpathFilter_Description

- [ ]     AbstractClassName
- [ ]     AnonInnerLength
- [ ]     ArrayTypeStyle
- [ ]     AvoidInlineConditionals
- [ ]     AvoidNestedBlocks
- [ ]     BooleanExpressionComplexity
- [ ]     CatchParameterName
- [ ]     ClassDataAbstractionCoupling
- [ ]     ClassFanOutComplexity
[...]

PR #14257, opened 2024-01-13
Issue #6207: Add Xpath Regression Test for PackageName

TanayMorakhia:
part of #6207

### DIFF
--- a/src/it/resources/org/checkstyle/suppressionxpathfilter/packagename/SuppressionXpathRegressionPackageNameTwo.java
+++ b/src/it/resources/org/checkstyle/suppressionxpathfilter/packagename/SuppressionXpathRegressionPackageNameTwo.java
@@ -1,0 +1,7 @@
+/* test comment */ package
+        org.checkstyle. // warn
+        suppressionxpathfilter.
+        packagename;
+
+public interface SuppressionXpathRegressionPackageNameTwo {
+}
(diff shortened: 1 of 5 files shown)


TanayMorakhia:
I think the test fail is not because of this PR.

src/it/resources/org/checkstyle/suppressionxpathfilter/packagename/SuppressionXpathRegressionPackageNameTwo.java
```diff
@@ -0,0 +1,4 @@
+package org.checkstyle.suppressionxpathfilter.packagename; // warn
```
romani, line 1: Same as in Two.
Please make package name wrapped in few lines, to make it a bit different.
And please add comment before `package`

romani (approved):
Items

romani (changes requested):
See above

src/it/resources/org/checkstyle/suppressionxpathfilter/packagename/SuppressionXpathRegressionPackageNameTwo.java
TanayMorakhia: done

romani (approved):
Ok to merge

PR #14263, opened 2024-01-15
Issue #14137: Enable `EmptyMethod` check

rickie:
Issue #14137.

This PR enables the https://error-prone.picnic.tech/bugpatterns/EmptyMethod/ check.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -244,6 +244,7 @@
       -Xep:CanonicalAnnotationSyntax:ERROR
       -Xep:DirectReturn:ERROR
       -Xep:CollectorMutability:ERROR
+      -Xep:EmptyMethod:ERROR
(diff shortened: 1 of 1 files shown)


rickie:
Unrelated error it seems?

> The version of Java (11.0.20.1) used to run this analysis is deprecated, and SonarCloud no longer supports it. Please upgrade to Java 17 or later.
As a temporary measure, you can set the property 'sonar.scanner.force-deprecated-java-version' to 'true' to continue using Java 11.0.20.1
This workaround will only be effective until January 28, 2024. After this date, all scans using the deprecated Java 11 will fail.

nrmancuso (approved):
Awesome!

nrmancuso:
We are aware of sonar failure, that can be ignored until we resolve/disable it

PR #14269, opened 2024-01-16
Issue #11163: Enforced File size AbstractJavadocPosition

MANISH-K-07:
Part of #11163 
Enforced file size for InputAbstractJavadocPosition
The decided file size limit was 120 lines and I have implemented the same.

Part of #13213
While modifying InputAbstractJavadocPosition.java ,  unnecessary '//ok' comment was found which has been removed as part of the same commit.


romani (approved):
Ok to merge

PR #14271, opened 2024-01-16
Issue #11163: Enforced file size on NonTightHtmlTags

MANISH-K-07:
Part of #11163

Enforced file size for InputAbstractJavadocHtmlTags.java
The decided file size limit was 120 lines and I have implemented the same.


romani (approved):
Ok to merge

PR #14286, opened 2024-01-17
Issue #13213: Removed //ok atclauseorder

MANISH-K-07:
Part of #13213 
Removed //ok comments from atclauseorder module.


romani (approved):
ok to merge if CI pass

PR #14297, opened 2024-01-18
Issue #13213: Removed //ok from illegaltokentext

mahfouz72:
remove //ok form illegal token text


romani:
Do not close PRs, just do fix and  push to GitHub with force to the same branch

PR #14316, opened 2024-01-20
Issue #14137: Enable `MockitoMockClassReference` check

rickie:
Issue #14137.

This PR enables the https://error-prone.picnic.tech/bugpatterns/MockitoMockClassReference/ check.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -247,6 +247,8 @@
       -Xep:EmptyMethod:ERROR
       -Xep:ExplicitEnumOrdering:ERROR
       -Xep:IsInstanceLambdaUsage:ERROR
+      -Xep:MockitoMockClassReference:ERROR
+      -Xep:NestedOptionals:ERROR
       <!-- Reason at https://github.com/checkstyle/checkstyle/issues/8252. -->
       -Xep:JUnitClassModifiers:OFF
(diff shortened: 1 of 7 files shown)


romani (approved):
Ok to merge

Needs more reviews as there is code changes

pom.xml
```diff
@@ -254,6 +254,7 @@
       -Xep:JUnitValueSource:OFF
       <!-- Until https://github.com/checkstyle/checkstyle/issues/14194. -->
       -Xep:LexicographicalAnnotationListing:OFF
+      -Xep:MockitoMockClassReference:ERROR
```
rnveach, line 257: I think there should be a clear separation between OFFs and ERRORs. You added this one right after an OFF with a reason for being OFF. At just a glance, it looks like this new addition is following the reason of the other, but it is not.

I recommend putting all ERRORs on top with no comments, and OFFs one the bottom with comments.

rnveach (changes requested):

pom.xml
rickie: Added a commit for these changes!

rickie:
Rebased and resolved conflict!

rnveach (approved):

PR #14308, opened 2024-01-19
Issue checkstyle #11163: Enforced file size on inputs on InputComment…

mahfouz72:
issue #11163:
force file size on inputs on InputCommentsIndentationCommentIsAtTheEndOfBlock

PR #14335, opened 2024-01-23
Issue #13213: Removed //ok leftout checks

MANISH-K-07:
Part of #13213 
Removed '//ok' comments from leftout files of /noncompilable.

Also, noticed that there were few unnecessary suppress tags in checkstyle-input-suppressions.xml
The reason was due to the files changed in the recent PR [#14308](https://github.com/checkstyle/checkstyle/pull/14308).
I noticed that the comments were removed as part of enforcing file size but the respective suppress tags have not been removed.
The PR however was merged and thus the unnecessary suppress tags, which I have removed now.


MANISH-K-07:
The only two failing checks are 
Cirrus - JDK17
check (index) because of the Out of Memory exception..

Hope this is fine.

romani (approved):
ok to merge

Issue #13345, opened 2023-07-03
Enable examples tests


stoyanK7:
This Issue deals with enabling Example Checks.
If you plan to help us to fix issue, thanks a lot first of all, and secondly please make a comment "I am on xxxCheck" to let others know that you are working on it to avoid conflicts.

Some examples from the website are uncompilable and untested ..... conversion of `.txt` to `.java` and removal of `@Disabled `annotatioon in Test class will make them active.

Examples of ideal updates : 
- https://github.com/checkstyle/checkstyle/commit/7123cc2b7ca95f5f349e54776a02724f2e7b0d69
- https://github.com/checkstyle/checkstyle/commit/ddbcf447682a58a9de237391dad99e96903cb969
- https://github.com/checkstyle/checkstyle/commit/2cec4d51ffb0478689039251c66a4e0253d6ec9f

These commits implement your idea of `example same by java code and different by violation comments` .

Please create 1 PR for each Check , please look at referenced PR commits as example of update.

**_RULES :_**

- The java code/structure is same for all examples ( _meaning they should follow one template and look the same_ )
- Only comments are changing in content from Example1 to Example2 and so on....
- If code line is present in one Example on Line X, it should be located on same line in all other Examples. 
- Lines raising violations should be commented using " `// violation` "
- If line contains only violation comment but violation is not raised, it should be empty line or " `// ok, because ....` "

Visual sample of how examples should be same by java structure/statements but different by comments 
([taken from PR](https://github.com/checkstyle/checkstyle/pull/14166/files)):
![image](https://github.com/checkstyle/checkstyle/assets/812984/db385c6a-2957-460a-b83a-bd3f7fd90f8c)
 one more example([html](https://checkstyle.org/checks/naming/staticvariablename.html#Example1-code), [pull request](https://github.com/checkstyle/checkstyle/pull/14345/files)):
![image](https://github.com/checkstyle/checkstyle/assets/812984/e293b8d5-bcf0-419e-85b1-0aca357d35f7)
one more example (java code reuse same line numbers and only comments are changing by content):
![image](https://github.com/checkstyle/checkstyle/assets/812984/8c85ede3-fcc6-4b12-b692-2bbb1ee837dc)
  

List of remaining checks' tests:

 - [x] SuppressWithNearbyTextFilterExamplesTest.java
 - [x] SuppressWarningsFilterExamplesTest.java
 - [x] SuppressionSingleFilterExamplesTest.java
[...]

PR #14341, opened 2024-01-23
Issue #13345: Enable TestCases for GenericWhitespaceCheckExamplesTest

chrccl:
Issue #13345: Enable TestCases for GenericWhitespaceCheckExamplesTest


chrccl:
Hi @romani, can you help me with the failure checks? When i run mvn verify locally I got this error:
GenericWhitespaceCheckError
I don't understand why miss a violation when the test on the example was:
GenericWhitespaceCode

Zopsss:
Hey @chrccl the error is trying to say that there are only 5 violations but you have provided 6 violations. Remove the following line from the `testExample2` method and try again:

```
  "21:26: " + getCheckMessage(MSG_WS_FOLLOWED, ">"),
```

chrccl:
Hi @Zopsss  thank you for helping! Incredibly I've got the same error:
UpdateGenericCheckstyleWhitespaceError
I don't understand why I have to delete this line:
 ` "21:26: " + getCheckMessage(MSG_WS_FOLLOWED, ">"),`
In Example2.java in that line (row = 21, column = 26) there is a violation! Why don't I have to indicate it in the testExample2 method?

MANISH-K-07:
> Hey @chrccl the error is trying to say that there are only 5 violations but you have provided 6 violations. Remove the following line from the `testExample2` method and try again:
> 
> ```
>   "21:26: " + getCheckMessage(MSG_WS_FOLLOWED, ">"),
> ```

@chrccl , if you have noticed, the whitespace after '>'  in line:17 is not a violation according to the Google Java Style Guide. 
I think the whitespace after the '>' in line:21 is also not being considered a violation. Maybe, try making the comment as '//ok, because.... ' and removing the violation as expected from 'testExample2' like Zopsss already suggested.
Or if the whole point of the check is to expect a violation for the whitespace, I think adding another whitespace ought to throw the violation.

Do correct me if I am wrong!!

chrccl:
> > Hey @chrccl the error is trying to say that there are only 5 violations but you have provided 6 violations. Remove the following line from the `testExample2` method and try again:
> > ```
> >   "21:26: " + getCheckMessage(MSG_WS_FOLLOWED, ">"),
> > ```
> 
> @chrccl , if you have noticed, the whitespace after '>' in line:17 is not a violation according to the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace). I think the whitespace after the '>' in line:21 is also not being considered a violation. Maybe, try making the comment as '//ok, because.... ' and removing the violation as expected from 'testExample2' like Zopsss already suggested. Also modify the genericwhitespace.xml file accordingly. Or if the whole point of the check is to expect a violation for the whitespace, I think maybe adding another whitespace ought to throw the violation.
> 
> Do correct me if I am wrong!!

Hi @MANISH-K-07 and thank you too for helping me! The whitespace after '>' in line:17 is not a violation because after there isn't a method name or a constructor like in line: 21 (https://checkstyle.org/checks/whitespace/genericwhitespace.html#GenericWhitespace)
However, I tried both the idea of adding another whitespace and indicating that line with the comment 'ok, because...'; thus, modifying the genericwhitespace.xml file as well. Regarding the "double whitespace," "mvn verify" still gives me the same error. As for the second idea, the error now is as follows:
UpdateGenericCheckstyleWhitespaceError
Does anyone have any ideas?

MANISH-K-07:
I can see in your snapshot that the issue with line:21 has been resolved?

Coming to the error at present, that is in line:22.
The problem is with the error message which you have provided in the GenericWhitespaceCheckExamplesTest.java ?

chrccl:
> I can see in your snapshot that the issue with line:21 has been resolved?
> 
> Coming to the error at present, that is in line:22. The problem is with the error message which you have provided in the GenericWhitespaceCheckExamplesTest.java ? How about trying to change that message accordingly....?

The problem is that I provide exactly what it wants... if you see, on the left of the image there is GenericWhitespaceCheckExamplesTest.java and for line 22 I pass "MSG_WS_ILLEGAL_FOLLOW" that is the message for that violation.

MANISH-K-07:
> > I can see in your snapshot that the issue with line:21 has been resolved?
> > Coming to the error at present, that is in line:22. The problem is with the error message which you have provided in the GenericWhitespaceCheckExamplesTest.java ? How about trying to change that message accordingly....?
> 
> The problem is that I provide exactly what it wants... if you see, on the left of the image there is GenericWhitespaceCheckExamplesTest.java and for line 22 I pass "MSG_WS_ILLEGAL_FOLLOW" that is the message for that violation. For line 21 I don't have error anymore

Is it not showing expected to match as "is not followed by whitespace.." ?
Whereas you have passed the 'MSG_WS_ILLEGAL_FOLLOW' which is showing "is followed by an illegal character." msg

MANISH-K-07:
One step at a time.. great that line:21 isn't an issue now!!

Is it not showing expected to match as "is not followed by whitespace.." ?
Whereas you have passed the 'MSG_WS_ILLEGAL_FOLLOW' which is showing "is followed by an illegal character." msg
Maybe change the error msg in the violation comment accordingly to the second one....?

chrccl:
> One step at a time.. great that line:21 isn't an issue now!!
> 
> Is it not showing expected to match as "is not followed by whitespace.." ? Whereas you have passed the 'MSG_WS_ILLEGAL_FOLLOW' which is showing "is followed by an illegal character." msg Maybe change the error msg in the violation comment accordingly to the second one....?

Changing the violation comment works!!!
UpdateGenericCheckstyleWhitespaceError
Thank you very much for the help!!!

MANISH-K-07:
Thought So!!
See.. One step at a time always works....

Happy to Help!!

sktpy:
@chrccl, you are indeed correct.
As per the documentation `List a = new ArrayList<> ();` should throw a violation as the right right angle bracket (">") precedes a constructor in this case, and therefore should not be followed with whitespace, as evident in the example shown in the [documentation](https://checkstyle.org/checks/whitespace/genericwhitespace.html#Example2-code).

I have created an issue for the same, so that it doesn't goes unnoticed in this closed PR.

Also, for future reference, there is no need to close an existing PR and reopen another for doing some changes.
You can add additional changes by pushing new commits to your branch, just make sure to squash any additional commits. ([Reference](https://checkstyle.org/beginning_development.html#Starting_Development))

chrccl:
Perfect, can you link me the issue that you have created, please? Thank you
for the tip for changing the PR instead of closing it!
Do you think that I have to create the issue for this thing linking my new
PR?
Thank you again for your help!

Il giorno mer 24 gen 2024 alle ore 19:38 Saket ***@***.***>
ha scritto:

sktpy:
@chrccl https://github.com/checkstyle/checkstyle/issues/14344

PR #14355, opened 2024-01-26
Issue #13213: Removed //ok from declarationorder

Tahanima:
Part of #13213
Removed `//ok` comments from **InputDeclarationOrder2.java**


romani (approved):
Thanks!

PR #14362, opened 2024-01-27
minor: Simplify `PropertyCacheFileTest#testNonExistentResource`

Stephan202:
As of 0bb9038582406af853dd064d9c699ddc93c0994d the code under test no longer uses `ByteStreams.toByteArray`.

### DIFF
--- a/config/error-prone-suppressions/test-compile-phase-suppressions.xml
+++ b/config/error-prone-suppressions/test-compile-phase-suppressions.xml
@@ -6,11 +6,4 @@
     <description>Return value of &apos;findFirst&apos; must be used</description>
     <lineContent>.findFirst();</lineContent>
   </error>
-
-  <error>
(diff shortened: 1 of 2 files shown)


Stephan202:
(In case one wonders how I noticed this: while updating [our integration test](https://github.com/PicnicSupermarket/error-prone-support/tree/4849bb5ae242eccf51169aaa14045fa9c3dc993d/integration-tests), a [new Refaster rule](https://github.com/PicnicSupermarket/error-prone-support/blob/4849bb5ae242eccf51169aaa14045fa9c3dc993d/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/InputStreamRules.java#L30-L40) modified the test code in a way that is not compatible with Mockito, causing a test failure. Closer inspection made me realize that the test can be simplified.)

nrmancuso (approved):
Thanks for noticing this!

rnveach (approved):

PR #14377, opened 2024-01-29
Issue #14346: Bump maximum heap size for checker framework execution

### LINKED ISSUE #14346: Expanding maxmem to resolve OOM Exception (Checker checks)
Identified in various random PRs
for example [Issue #11163: Enforced file size commentsindentation #14338
](https://github.com/checkstyle/checkstyle/pull/14338)
image

As per my observations, the build is a clean success and the problem with Checker checks does not show on my local system.
I'm not sure if this is relevant but my system has JDK x64 bit and I have preset the heap size of JVM to 8192m.
I agree that the failing CI on the server is a huge pain and needs to be resolved asap.

image

Ever since the OOM exception started to pop up, I have tried to find a workaround by searching for all possible alternatives to resolve this issue and in the process read the [Oracle Java Documentation](https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/memleaks002.html). After a bit of understanding, the only thing that came to my mind was to expand the memory size of the JVM's heap. The [OOM exception/issue](https://github.com/checkstyle/checkstyle/actions/runs/7630359181/job/20785856626?pr=14338#step:6:101) at hand is the result of the Gc using up >98% of the mem.
As per my understanding of the [pom.xml](https://github.com/checkstyle/checkstyle/blob/master/pom.xml) , the preset initialmem and maxmem for the Checker checks are 1024m & **_4048m_** respectively, which obviously is the root cause of the problem at present. You can check the below snap, which is one of the examples..

image

**_Proposed fix :_**
I think expanding the maxmem to higher size (like 8192m as in my local system or whatever limit that you recommend) is most likely to resolve this issue. This can be implemented only for the Checker check(index), as it is the constantly failing check or since like in the current PR where few other Checker checks are also throwing the same said exception unexpectedly, the mem can be expanded for the framework completely (all 8 checks).
I can try to implement the said fix by changing the preset maxmem for Checker Framework so that this is of no longer concern to the project.

nrmancuso:
Closes #14346

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2525,7 +2525,7 @@
               <fork>true</fork>
               <failOnError>false</failOnError>
               <meminitial>1024m</meminitial>
-              <maxmem>4048m</maxmem>
+              <maxmem>8192m</maxmem>
               <compilerArguments>
                 <Xmaxerrs>10000</Xmaxerrs>
                 <Xmaxwarns>10000</Xmaxwarns>
@@ -2584,7 +2584,7 @@
               <fork>true</fork>
               <failOnError>false</failOnError>
               <meminitial>1024m</meminitial>
-              <maxmem>4048m</maxmem>
+              <maxmem>8192m</maxmem>
               <compilerArguments>
                 <Xmaxerrs>10000</Xmaxerrs>
                 <Xmaxwarns>10000</Xmaxwarns>
@@ -2643,7 +2643,7 @@
               <fork>true</fork>
               <failOnError>false</failOnError>
               <meminitial>1024m</meminitial>
-              <maxmem>4048m</maxmem>
+              <maxmem>8192m</maxmem>
               <compilerArguments>
                 <Xmaxerrs>10000</Xmaxerrs>
                 <Xmaxwarns>10000</Xmaxwarns>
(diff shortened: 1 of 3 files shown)


nrmancuso:
We need to get CI back to green state, cc @MANISH-K-07

rnveach:
> We need to get CI back to green state

So who is going to fix Cirrus? It is still failing even in this PR. :)

>  Failed to start an instance: FAILED_PRECONDITION: Monthly compute limit exceeded!

pom.xml
```diff
@@ -2525,7 +2525,7 @@
               <fork>true</fork>
               <failOnError>false</failOnError>
               <meminitial>1024m</meminitial>
-              <maxmem>4048m</maxmem>
+              <maxmem>8192</maxmem>
```
rnveach, line 2528: I assume the `m` is understood? It sort of makes more sense to keep it so we are explicit. Other items like `meminitial` still use the `m`.

rnveach (approved):

pom.xml
nrmancuso: Yes, `m` is not required, but I added it back to keep them consistent.

romani:
>So who is going to fix Cirrus?

Not fix, we need to slow down contributions (less PRs executions) or migrate off this CI.

MANISH-K-07:
> We need to get CI back to green state, cc @MANISH-K-07

Apologies for not being able to fix it on time, I was having trouble applying the patch files for some reason!!

romani:
@MANISH-K-07, if you see a way to improve our groovy script and fail when Checker is failed, please send separate PR, we would rather catch this problem sooner and more explicitly next time.

MANISH-K-07:
> @MANISH-K-07, if you see a way to improve our groovy script and fail when Checker is failed, please send separate PR, we would rather catch this problem sooner and more explicitly next time.

@romani ,
Consider it done !!

Issue #13999, opened 2023-11-06
Resolve Pitest Suppression in Pitest-Javadoc Profile


Kevin222004:
As a part of this issue, we have to kill the mutation listed below and remove the suppression from [pitest-javadoc-suppression.xml](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-javadoc-suppressions.xml)

# How to Kill the Mutation

## 1) Take 1 mutation and apply to the code

Pick up the one mutation from the [pitest-javadoc-suppression.xml file](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-javadoc-suppressions.xml) or for better understanding you can generate the pitest report as well by Following [Steps](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports)
 and finally hardcode the mutation.  

**How to hardcode the mutation**
For better understanding of mutatnt, That how it works. Please checkout [Pitest Doc](https://pitest.org/) In this you will find the detail explaination of each mutant with example.

## 2) Send the pr and generate the report

After hardcoding the code please send a pull request and try to generate the regression Report. If Regression report found the diff then you got your test cases. use that part of code as your input and create a test.

**How to generate the Regression Report**
To learn how to generate the regression report Please go through the Following link.
https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#report-generation 

After reading this you found the 2 method two genrate the report please try to generate the report with Github Action.
For example https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#generation-examples 
Basically you have to generate the 2 report in with project file 
1) [Project-list-1](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/github-action-projects1.properties)
2) [Project-list-2](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/github-action-projects2.properties)
3) [Project-list-3](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/projects-to-test-on-for-github-action.properties)

## 3) Removal of code
If no diff found from all the report just follow the mutation apply the changes and update the pr.

Apart of this for more detailed explaination please go through [Wiki-Page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports) and [Pit Documentation](https://pitest.org/)

[...]

PR #14393, opened 2024-01-31
Issue #13999: Killed mutation for ScopeUtil

Rohanraj123:
Related #13999 

Mutation : 

https://github.com/checkstyle/checkstyle/blob/60267bb9848194182de1f6a0980999a5f5bcbe28/config/pitest-suppressions/pitest-utils-suppressions.xml#L12-L19

Diff Regression Config : https://raw.githubusercontent.com/Rohanraj123/Rohanraj123.github.io/main/my_check.html

Diff Regression Projects : https://raw.githubusercontent.com/Rohanraj123/Rohanraj123.github.io/main/ProjectsConfig

### DIFF
--- a/config/pitest-suppressions/pitest-utils-suppressions.xml
+++ b/config/pitest-suppressions/pitest-utils-suppressions.xml
@@ -12,15 +12,6 @@
   <mutation unstable="false">
     <sourceFile>ScopeUtil.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.utils.ScopeUtil</mutatedClass>
-    <mutatedMethod>getSurroundingScope</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getParent with receiver</description>
-    <lineContent>for (DetailAST token = node.getParent();</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>ScopeUtil.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.utils.ScopeUtil</mutatedClass>
     <mutatedMethod>isInBlockOf</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
     <description>removed conditional - replaced equality check with true</description>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtilTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtilTest.java
@@ -24,12 +24,19 @@ import static com.puppycrawl.tools.checkstyle.internal.utils.TestUtil.isUtilsCla
 
 import org.junit.jupiter.api.Test;
 
+import com.puppycrawl.tools.checkstyle.AbstractModuleTestSupport;
 import com.puppycrawl.tools.checkstyle.DetailAstImpl;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.Scope;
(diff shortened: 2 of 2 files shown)


Rohanraj123:
Github, generate report

nrmancuso:
@Rohanraj123 few things:

- How did you build this dependency tree? What does it mean?
- How did you determine that we needed to generate a check regression report for NestedSwitchCheck?
- Why didn't we simply hardcode the surviving mutation that pitest found?

Also, we will need to run check regression testing on ALL checks that use this method (with a few different configurations):
```
➜  checkstyle ag -lQ "ScopeUtil.getSurroundingScope(" --ignore "*Test.java"
src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/design/DesignForExtensionCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/MissingJavadocMethodCheck.java
src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/MissingJavadocTypeCheck.java

```

I suggest that you put this PR into draft mode while you work on these tasks.

Rohanraj123:
@nrmancuso I am working on asked questions but you can please first confirm me that changes made was okay or not.

nrmancuso:
> can you please first confirm me that changes made was okay or not

No one can answer this question until you have answered my questions above, and generated a report :) We have no way to know how your changes have affected all of the checks that use this method. I suggest you strictly follow steps at https://github.com/checkstyle/checkstyle/issues/12341.

Rohanraj123:
> * How did you build this dependency tree? What does it mean?

I have updated dependency tree. Letme first explain what my understanding of dependency tree is : Which are other checks methods which depends on the specific method. If yes then I think now dependency tree must be understandable. And also Updated the checks needed to generate regression report over. 

> * How did you determine that we needed to generate a check regression report for NestedSwitchCheck?

Yeah I was wrong here I have updated checks

> * Why didn't we simply hardcode the surviving mutation that pitest found?

If you mean by adding suppressions then this is something which is not allowed.

nrmancuso:
> I have updated dependency tree. Letme first explain what my understanding of dependency tree is …

Let me try to be more clear: why do we need this “dependency tree”? Where in https://github.com/checkstyle/checkstyle/issues/12341 is it advised to do this? 

> If you mean by adding suppressions then this is something which is not allowed.

Please carefully read though https://github.com/checkstyle/checkstyle/issues/12341. You need to first change the code to exactly what pitest is mutating to.

Rohanraj123:
> Please carefully read though #12341. You need to first change the code to exactly what pitest is mutating to.

I have read #12341 . I have removed the unwanted material.

src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
```diff
@@ -131,23 +131,21 @@ else if (isInInterfaceOrAnnotationBlock(ast)) {
      */
     public static Scope getSurroundingScope(DetailAST node) {
         Scope returnValue = null;
-        for (DetailAST token = node.getParent();
```
nrmancuso, line 134: The only change in this PR right now should be here, we should hard code exactly what pitest is mutating this line to: `DetailAST token = node;`

nrmancuso:
@Rohanraj123 IMO, we don’t need a custom projects file right now.

src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
Rohanraj123: done !

Rohanraj123:
> @Rohanraj123 IMO, we don’t need a custom projects file right now.

No worries I hardcoded the mutation and revert my changes that I made earlier.

PR #14401, opened 2024-02-01
Issue #14084: Remove checker violation for nullable return in abstrac…

Rohanraj123:
Related #14084

PR #14403, opened 2024-02-01
Issue #14084: fix Checker violation for nullable return RightCurlyCheck

MANISH-K-07:
Part of #14084 

Updated RightCurlyCheck.java as per requirements of this issue.

### DIFF
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -11,13 +11,6 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java</fileName>
     <specifier>toarray.nullable.elements.not.newarray</specifier>
     <message>call of toArray on collection of non-null elements yields an array of possibly-null elements; omit the argument to toArray or make it an explicit array constructor</message>
-    <lineContent>return result.toArray(CommonUtil.EMPTY_STRING_ARRAY);</lineContent>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java</fileName>
-    <specifier>toarray.nullable.elements.not.newarray</specifier>
-    <message>call of toArray on collection of non-null elements yields an array of possibly-null elements; omit the argument to toArray or make it an explicit array constructor</message>
     <lineContent>return result.toArray(EMPTY_MODIFIER_ARRAY);</lineContent>
   </checkerFrameworkError>
 
--- a/src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java
@@ -368,6 +368,7 @@ public abstract class AbstractAutomaticBean
 
         @SuppressWarnings("unchecked")
         @Override
+        @Nullable
         public Object convert(Class type, Object value) {
             final StringTokenizer tokenizer = new StringTokenizer(
                 value.toString().trim(), COMMA_SEPARATOR);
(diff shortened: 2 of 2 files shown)


MANISH-K-07:
Response to comment at https://github.com/checkstyle/checkstyle/pull/14401#pullrequestreview-1859119526

@nrmancuso ,
Without casting return type, we find new surviving mutations like the one below for example: 

```
New surviving error(s) found:

    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java</fileName>
    <specifier>argument</specifier>
    <message>incompatible argument for parameter ast2 of TokenUtil.areOnSameLine.</message>
    <lineContent>&amp;&amp; !TokenUtil.areOnSameLine(rcurly, Details.getNextToken(tokenAfterNextToken));</lineContent>
    <details>
    </details>
  </checkerFrameworkError>
```

romani:
If you need our assistance , please move PR in ready for review

MANISH-K-07:
> If you need our assistance , please move PR in ready for review

@romani 
That would be very much helpful indeed. I have been trying but new surviving mutations are getting induced each time....

rdiachenko:
@MANISH-K-07 

> Without casting return type, we find new surviving mutations like the one below for example

You don't need casting here. What you see is that after making `getNextToken` nullable, the checker error propagated to the caller `!TokenUtil.areOnSameLine(rcurly, Details.getNextToken(tokenAfterNextToken))`. The new error message says the second argument for `TokenUtil.areOnSameLine` is expected to be not null but nulluble value is provided. So, as a fix you may try to mark that second argument with nullable annotation.

Rohanraj123:
@MANISH-K-07 I think to fix this issue . You need to update all the other methods who call this `getNextToken `method . 
List of methods : 
```
1. getDetailsForOthers
2. getDetailsForIf
3. getDetailsForTryCatch
4. getDetailsForSwitch
5. isBlockAloneOnSingleLine
6. skipDoubleBraceInstInit
```

Add `@Nullable` annotation over these methods can resovle the issue i guess .

Rohanraj123:
@MANISH-K-07 I think updating those methods which are calling `getNextToken `method with @NonNull annotation can resolve the issue.
List of methods : 
```
1. skipDoubleBraceInstInit do `Details.getNextToken(tokenAfterNextToken` -> `Objects.requireNonNull(Details.getNextToken(tokenAfterNextToken))`
2. add @NonNull over getDetailsForTryCatch
3. add @NonNull over getDetailsForIf
4. add @NonNull over getDetailsForOthers
```

MANISH-K-07:
@Rohanraj123 ,

I blank pushed a commit to check what the surviving mutations were....
Will work on your suggestions now

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -22,6 +22,9 @@
 import java.util.Arrays;
 import java.util.Locale;
 
+import javax.annotation.Nullable;
+import javax.annotation.Nonnull;
+
```
Vyom-Yadav, line 27: Checker Framework does not use these annotations. It allows them (https://checkerframework.org/manual/#nullness-related-work), but I'd prefer using checker framework's Nonnull and Nullable annotations. Please see https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/NonNull.html.

Also, please attentively read how the nullness checker works: https://checkerframework.org/manual/#nullness-checker

Vyom-Yadav (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
romani: >It treats annotations from other tools as if you had written the corresponding annotation from the Nullness Checker,

it does

> but I'd prefer using checker framework's Nonnull and Nullable annotations.

it will force us to add this tool to our compile dependecy, that will grow our jar, lets avoid this for now until we find this tool critical for us.

PR #14410, opened 2024-02-02
Issue #13999: Resolve Pitest suppression for `ancestor.getType() == TokenTypes.LITERAL_TRY` of JavadocMethod

sktpy:
Part of #13999 

([Check Documentation](https://checkstyle.org/checks/javadoc/javadocmethod.html))

**Mutation:**
https://github.com/checkstyle/checkstyle/blob/943576fc0da0f0e00b15b7d77f15b16dc8823416/config/pitest-suppressions/pitest-javadoc-suppressions.xml#L111-L118
<hr>

### Definitions:
1. **and-expression**:-
```java
ancestor.getType() == TokenTypes.LITERAL_TRY
&& ancestor.findFirstToken(TokenTypes.LITERAL_CATCH) != null	
```
2. **first-operand**: `ancestor.getType() == TokenTypes.LITERAL_TRY`
3. **second-operand**: `ancestor.findFirstToken(TokenTypes.LITERAL_CATCH) != null`
4. **mutant**: replacing **first-operand** with `true`.
<hr>

### Why mutant survives:
- **mutant** does not alters the result of **and-expression** because `LITERAL_CATCH` cannot exist without being the child of `LITERAL_TRY`. 
- This makes evaluating **first-operand** not a necessity because **second-operand** being evaluated as `true` implicitly means that the type of `ancestor` is `LITERAL_TRY`.
<hr>

### Why mutant cannot be killed:
- For this, there needs to be an input where `LITERAL_CATCH` exist without a preceding `LITERAL_TRY` block
- This results in compilation error and hence is not possible.
<hr>

### Why not remove second-operand:
- The **and-expression** is meant to evaluate if the `throw` is inside a `try` block which has a corresponding `catch` block.
- Unlike `catch` block, which cannot exist without a `try` block, `try` block _can_ exist without a `catch` block.
<hr>

### ⚠️ Performance Impact: 
- Benefit due to short circuiting in **and-expression** should be weighed in before merging this PR.
- If `ancestor` is not `LITERAL_TRY`, `findFirstToken()` is not only useless for the reason that it won't find a `LITERAL_CATCH` ;
- But along with that, it may impact performance if `ancestor` has a significant number of children.
- This is avoided by short circuiting in **and-expression** but results in the surviving **mutant**.
<hr>

### Change: 
Updated the code as Pitest mutated, except it is now the last operand of the or-statement to help mitigate the performance impact by a bit.
(Although this shifting of operands may produce negligible performance gains, still better than none)
<hr>

### Usage:
Definition and the only usage for the method with **mutant** is in `JavadocMethodCheck`. (1 usage)
`JavadocMethodCheck` does not affect any other class.
**Call Stack:**
```
isInIgnoreBlock:691, JavadocMethodCheck (com.puppycrawl.tools.checkstyle.checks.javadoc)
  getThrowed:640, JavadocMethodCheck (com.puppycrawl.tools.checkstyle.checks.javadoc)
    checkComment:442, JavadocMethodCheck (com.puppycrawl.tools.checkstyle.checks.javadoc)
      processAST:392, JavadocMethodCheck (com.puppycrawl.tools.checkstyle.checks.javadoc)
        visitToken:374, JavadocMethodCheck (com.puppycrawl.tools.checkstyle.checks.javadoc)
          notifyVisit:335, TreeWalker (com.puppycrawl.tools.checkstyle)
            processIter:408, TreeWalker (com.puppycrawl.tools.checkstyle)
              ...
```
<hr>

### Regression:
**Result:** No differences found. ([Report](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/2c6df13_2024140229/reports/diff/index.html))

Diff Regression config: https://gist.githubusercontent.com/sktpy/44a31ea4e946a0755fa3e331fd65de98/raw/233b4d68f7cd46f957ee37e94813922d5fd1733c/check.xml

Diff Regression projects: https://gist.githubusercontent.com/sktpy/49239940660ceaef88d224d6abc8878e/raw/058cbb1b662c9276e934fdffbb99cec760332398/projects-to-test-on.properties

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
@@ -688,12 +688,11 @@ public class JavadocMethodCheck extends AbstractCheck {
     private static boolean isInIgnoreBlock(DetailAST methodBodyAst, DetailAST throwAst) {
         DetailAST ancestor = throwAst;
         while (ancestor != methodBodyAst) {
-            if (ancestor.getType() == TokenTypes.LITERAL_TRY
-                    && ancestor.findFirstToken(TokenTypes.LITERAL_CATCH) != null
-                    || ancestor.getType() == TokenTypes.LAMBDA
-                    || ancestor.getType() == TokenTypes.OBJBLOCK) {
-                // throw is inside a try block, and there is a catch block,
-                // or throw is inside a lambda expression/anonymous class/local class
+            if (ancestor.getType() == TokenTypes.LAMBDA
+                    || ancestor.getType() == TokenTypes.OBJBLOCK
+                    || ancestor.findFirstToken(TokenTypes.LITERAL_CATCH) != null) {
+                // throw is inside a lambda expression/anonymous class/local class,
+                // or throw is inside a try block, and there is a catch block
                 break;
             }
             if (ancestor.getType() == TokenTypes.LITERAL_CATCH
--- a/config/pitest-suppressions/pitest-javadoc-suppressions.xml
+++ b/config/pitest-suppressions/pitest-javadoc-suppressions.xml
@@ -111,15 +111,6 @@
   <mutation unstable="false">
     <sourceFile>JavadocMethodCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocMethodCheck</mutatedClass>
-    <mutatedMethod>isInIgnoreBlock</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
-    <description>removed conditional - replaced equality check with true</description>
-    <lineContent>if (ancestor.getType() == TokenTypes.LITERAL_TRY</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>JavadocMethodCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocMethodCheck</mutatedClass>
     <mutatedMethod>processThrows</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
     <description>removed call to com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck$Token::getText</description>


sktpy:
GitHub, generate report

rnveach:
@sktpy How did you get that exclamation icon in your post?

sktpy:
@rnveach I have Characters app ([link](https://apps.gnome.org/en-GB/Characters/)), it can fetch results for emojis directly in system search or you can select one in-app.
You can also use unicodes for the corresponding emojis.
There are probably other better ways.

sktpy:
`ci/circleci: run-inspections` is reporting urelated errors regarding `YAMLSchemaValidation` for [site.yml](https://github.com/checkstyle/checkstyle/blob/master/.github/workflows/site.yml). ([Artifacts](https://app.circleci.com/pipelines/github/checkstyle/checkstyle/1/workflows/7fa59043-69d8-4618-bd5f-7e522e0050cb/jobs/498800/artifacts))

sktpy:
Clicked on update branch here which created two commits, standby while I fix it

sktpy:
GitHub, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
```diff
@@ -688,12 +688,11 @@ private static DetailAST getFirstClassNameNode(DetailAST ast) {
     private static boolean isInIgnoreBlock(DetailAST methodBodyAst, DetailAST throwAst) {
         DetailAST ancestor = throwAst;
         while (ancestor != methodBodyAst) {
-            if (ancestor.getType() == TokenTypes.LITERAL_TRY
-                    && ancestor.findFirstToken(TokenTypes.LITERAL_CATCH) != null
```
rnveach, line 692: Why was the catch moved to the bottom of the if?

> except it is now the last operand of the or-statement to help mitigate the performance impact by a bit.

What performance impact are you referring to?

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
```diff
@@ -688,12 +688,11 @@ private static DetailAST getFirstClassNameNode(DetailAST ast) {
...
-                    || ancestor.getType() == TokenTypes.LAMBDA
-                    || ancestor.getType() == TokenTypes.OBJBLOCK) {
-                // throw is inside a try block, and there is a catch block,
-                // or throw is inside a lambda expression/anonymous class/local class
+            if (ancestor.getType() == TokenTypes.LAMBDA
+                    || ancestor.getType() == TokenTypes.OBJBLOCK
+                    || ancestor.findFirstToken(TokenTypes.LITERAL_CATCH) != null) {
+                // throw is inside a lambda expression/anonymous class/local class,
+                // or throw is inside a try block, and there is a catch block
```
rnveach, line 695: `inside a try block, and there is a catch block` => `inside a catch block of a try`

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
sktpy: > What performance impact are you referring to?

Referring to the performance impact that is mentioned in the PR description.
<br>

> Why was the catch moved to the bottom of the if?

Let's say if the type of `ancestor` is either `LAMBDA` or `OBJBLOCK`. 
If the operands were in original ordering, `findFirstToken()` would be the first to run in every iteration regardless the type, checking all the children of the ancestor.
This will all be in vain because it wouldn't find `LITERAL_CATCH` because it can only exist as the child for `ancestor` if the type is `LITERAL_TRY`.
Shifting that to the last, helps us avoid this because it will exit immediately if the type matches either `LAMBDA` or `OBJBLOCK`.

This wouldn't be helpful for cases where the type doesn't matches either `LAMBDA` or `OBJBLOCK` or if the `ancestor` is in some intermediate part of such type. Still some optimization is better than none.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
sktpy: My understanding is that this checks if `throw` is inside the `try` block which has a succeeding `catch` block.
Wouldn't "`inside a catch block of a try`" mean that the `throw` _is not_ inside the `try` block but in the `catch` block of that `try` block?

sktpy: > **Unrelated:** I clicked on `re-request review` to signify that the review items were answered and were ready for review.
Please do let me know(for future cases) if I am not supposed to do that and reviewer(s) will check the PR again once they get time to do so.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
rnveach: I believe you are wrongly assuming one token is more likely to appear than another. If "ancestor is either LAMBDA or OBJBLOCK", but what if ancestor is CATCH then it negates your argument as putting it first would trigger it the fastest. OBJBLOCK is more likely to be seen than any other because more structures use it, but the rarity of LAMBDA versus CATCH is highly dependent on each Java source. One could have a ton of Lambdas or a ton of catches for whichever reason. For example, my projects don't use lambdas but does have some catches.

We can leave it like this for now.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
rnveach: The point of the rewording was that the original seemed to equally mention try and catch when both tokens were written in the code. Now that one is removed, it should be more written about what is remaining which is in the catch block.

> mean that the throw is not inside the try block but in the catch block of that try block

I am not sure what this code is exactly referencing. So your saying the code is basically `throw is inside a block with a catch block`?

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
sktpy: True, occurrence of such tokens is code dependent, but my main reason to change the ordering was that the `getType()` is comparatively much faster than `findFirstToken()`.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
sktpy: While the try code is removed, the meaning of `ancestor.findFirstToken(TokenTypes.LITERAL_CATCH) != null` remains the same, as this being evaluated as `true` implicitly means that the type of ancestor is `LITERAL_TRY` (the reason why mutation was able to survive).

Now the literal meaning of this is that the `throw is inside an ancestor which has a child with the type "LITERAL_CATCH"`. --- (1)
But as we know that such ancestor can only be the type `LITERAL_CATCH`.

Do you think that the comment can be reworded as **(1)** or perhaps even better:
`throw is inside a try block, and that try block has a corresponding catch block`

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
rnveach: I glossed over `findFirstToken`. I am resolving this now.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
rnveach: I am just going to resolve this to move on.

rnveach (approved):

romani (approved):
thanks a lot for wide explanation.
I placed your PR as example to issue description to let others do similar.

sktpy:
The pleasure was all mine, mutant hunting is a lot of fun  👾 👾 👾

nrmancuso (approved):

PR #14461, opened 2024-02-11
Issue #11446: Update TestTabViolation Method Of Checkertest

piyush-1234:
Issue: #11446

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
@@ -1408,13 +1408,11 @@ public class CheckerTest extends AbstractModuleTestSupport {
 
     @Test
     public void testTabViolationDefault() throws Exception {
-        final DefaultConfiguration checkConfig =
-            createModuleConfig(VerifyPositionAfterTabFileSet.class);
         final String[] expected = {
-            "2:9: violation",
-            "3:17: violation",
+            "10:17: violation",
+            "13:33: violation",
         };
-        verify(checkConfig, getPath("InputCheckerTabCharacter.txt"),
+        verifyWithInlineConfigParser(getPath("InputCheckerTabCharacter.txt"),
             expected);
     }
 
@@ -1425,8 +1423,8 @@ public class CheckerTest extends AbstractModuleTestSupport {
         final DefaultConfiguration checkerConfig = createRootConfig(checkConfig);
         checkerConfig.addProperty("tabWidth", "4");
         final String[] expected = {
-            "2:5: violation",
-            "3:9: violation",
+            "10:13: violation",
+            "13:33: violation",
         };
         verify(checkerConfig, getPath("InputCheckerTabCharacter.txt"),
             expected);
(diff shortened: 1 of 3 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
```diff
@@ -1424,12 +1422,12 @@ public void testTabViolation() throws Exception {
...
-            "2:5: violation",
-            "3:9: violation",
-        };
-        verify(checkerConfig, getPath("InputCheckerTabCharacter.txt"),
-            expected);
+
+        execute(checkerConfig, getPath("InputCheckerTabCharacter.txt"));
+
+        assertWithMessage("Tab width should be set to 4")
+            .that(checkerConfig.getProperty("tabWidth"))
```
romani, line 1429: Please explain why not a verify

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
piyush-1234: execute method covered this below line of setTabWidth method

`this.tabWidth = tabWidth;`

```
/**
     * Set the tab width to report audit events with.
     *
     * @param tabWidth an {@code int} value
     */
    public final void setTabWidth(int tabWidth) {
        this.tabWidth = tabWidth;
    }
```

here is the surviving mutation which i want to cover through this test so that pitest common get pass

```
<mutation unstable="false">
+    <mutatedMethod>setTabWidth</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
+    <description>Removed assignment to member variable tabWidth</description>
+    <lineContent>this.tabWidth = tabWidth;</lineContent>
+  </mutation>
```


while using verifyWithInlineConfigParser is not covering that same line (this.tabWidth = tabWidth;)
this was the first reason

second reason i found that is, a repeatation of same test method is i use verifyWithInlineConfigParser in **testTabViolation** method. see below

```
@Test
    public void testTabViolationDefault() throws Exception {
        final String[] expected = {
            "10:17: violation",
            "13:33: violation",
        };
        verifyWithInlineConfigParser(getPath("InputCheckerTabCharacter.txt"),
            expected);
    }
    
    @Test
    public void testTabViolation() throws Exception {
        final String[] expected = {
            "10:17: violation",
            "13:33: violation",
        };
        verifyWithInlineConfigParser(getPath("InputCheckerTabCharacter.txt"),
            expected);
    }
```

may be i am wrong, not very much sure about analysis.

but pitest is still failing, 
i need to figure out that.

romani: Not a good reason.
Please do update like at https://github.com/checkstyle/checkstyle/pull/14462 to cover coverage of Treewalker

romani (approved):
Good step forward.
Checker is not migrated completely, but good step forward

PR #14464, opened 2024-02-11
Issue #6207: Add XPath IT Regression test for IllegalInstantiationCheck

mahfouz72:
part of #6207 [doc](https://checkstyle.org/checks/coding/illegalinstantiation.html#IllegalInstantiation)
add xPath regression test for   IllegalInstantiationCheck
1 test for normal case, 1 for anonymous class, 1 for interface
the input files aren't compiled by IntelliJ because there is a deprecation warning for the Boolean(boolean) and Integer constructors


romani (approved):
Ok to merge

Issue #14122, opened 2023-12-08
Recfactor classes to avoid throwing exceptions from constructors


romani:
detected at https://github.com/checkstyle/checkstyle/pull/14117?notification_referrer_id=NT_kwDOAAxnuLE4Njc0Mzg0MTcxOjgxMjk4NA#issuecomment-1844198684

exceptions in ctors are added in scope of https://github.com/checkstyle/checkstyle/pull/5753
it is update resulted by pitest https://github.com/checkstyle/checkstyle/pull/5753/files#r189127556 

we need to remove exceptions and find differen way to fix pitest survivals

abhinavs1920:
Can you please assign this issue to me?

romani:
@abhinavs1920, if you are new to our code base, I highly recommend https://github.com/checkstyle/checkstyle/wiki/To-New-Contributor

This refactoring might be tricky. 
We do not assignments, you just need to make a comment "I am on it" and start working on it.

relentless-pursuit:
@Romani, is the refactoring and pitest removal part of the same isssue? I was going through the code of `DefaultLogger.java` and I see a lot of overloaded constructors. One standard approach would be to use a builder pattern. And, i am just mentioning this from a surface-level investigation. But, how would that affect the client side of creating an instance of `DefaultLogger`? So, is the idea of refactoring to ensure we continue to create objects the way we do, but the exceptions shall not be thrown while instantiating?

romani:
It is refactoring. Pitest should not be increased.

>One standard approach would be to use a builder pattern

This is API class , we can not fix it, it will break users code.

>but the exceptions shall not be thrown while instantiating?

Yes, exception in ctor is bad design.

relentless-pursuit:
> It is refactoring. Pitest should not be increased.
> 
> > One standard approach would be to use a builder pattern
> 
> This is API class , we can not fix it, it will break users code.
> 
> > but the exceptions shall not be thrown while instantiating?
> 
> Yes, exception in ctor is bad design.

Understood. Will spend some time on it and see if i can reach a solution. On it.

romani:
Try update single class and send PR for review

relentless-pursuit:
> Try update single class and send PR for review

Sure

Youssefalaa7:
I am on CsvFilterElement

romani:
final fix for what should permanently should be not fixed is done.

PR #14475, opened 2024-02-13
Issue #14122: Recfactor classes to avoid throwing exceptions from constructor

relentless-pursuit:
Partial of #14122

This PR aims to refactor the class to avoid exception from being thrown from the constructor.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java
@@ -58,17 +58,29 @@ public class DefaultLogger extends AbstractAutomaticBean implements AuditListene
     public static final String AUDIT_FINISHED_MESSAGE = "DefaultLogger.auditFinished";
 
     /** Where to write info messages. **/
-    private final PrintWriter infoWriter;
+    private PrintWriter infoWriter;
     /** Close info stream after use. */
     private final boolean closeInfo;
 
     /** Where to write error messages. **/
-    private final PrintWriter errorWriter;
+    private PrintWriter errorWriter;
     /** Close error stream after use. */
     private final boolean closeError;
 
     /** Formatter for the log message. */
     private final AuditEventFormatter formatter;
+
+    /** Options for the info stream. */
+    private OutputStreamOptions infoStreamOptions;
+
+    /** Options for the error stream. */
+    private OutputStreamOptions errorStreamOptions;
+
+    /** The OutputStream for info messages. */
+    private OutputStream infoStream;
+
+    /** The OutputStream for error messages. */
+    private OutputStream errorStream;
 
     /**
      * Creates a new {@code DefaultLogger} instance.
@@ -128,31 +140,27 @@ public class DefaultLogger extends AbstractAutomaticBean implements AuditListene
                          OutputStream errorStream,
                          OutputStreamOptions errorStreamOptions,
                          AuditEventFormatter messageFormatter) {
-        if (infoStreamOptions == null) {
-            throw new IllegalArgumentException("Parameter infoStreamOptions can not be null");
-        }
-        closeInfo = infoStreamOptions == OutputStreamOptions.CLOSE;
-        if (errorStreamOptions == null) {
-            throw new IllegalArgumentException("Parameter errorStreamOptions can not be null");
-        }
-        closeError = errorStreamOptions == OutputStreamOptions.CLOSE;
-        final Writer infoStreamWriter = new OutputStreamWriter(infoStream, StandardCharsets.UTF_8);
-        infoWriter = new PrintWriter(infoStreamWriter);
 
-        if (infoStream == errorStream) {
-            errorWriter = infoWriter;
-        }
-        else {
-            final Writer errorStreamWriter = new OutputStreamWriter(errorStream,
-                    StandardCharsets.UTF_8);
-            errorWriter = new PrintWriter(errorStreamWriter);
-        }
-        formatter = messageFormatter;
+        this.infoStream = infoStream;
+        this.infoStreamOptions = infoStreamOptions;
+        this.errorStream = errorStream;
+        this.errorStreamOptions = errorStreamOptions;
+        this.formatter = messageFormatter;
+
+        closeInfo = infoStreamOptions == OutputStreamOptions.CLOSE;
+        closeError = errorStreamOptions == OutputStreamOptions.CLOSE;
+
+        finishLocalSetup();
     }
 
     @Override
     protected void finishLocalSetup() {
-        // No code by default
+        if (infoStream == null || errorStream == null) {
+            throw new IllegalArgumentException("Error and info streams cannot be null.");
+        }
+        infoWriter = new PrintWriter(new OutputStreamWriter(infoStream, StandardCharsets.UTF_8));
+        errorWriter = (infoStream == errorStream) ? infoWriter :
+                new PrintWriter(new OutputStreamWriter(errorStream, StandardCharsets.UTF_8));
     }
 
     /**
--- a/src/test/java/com/puppycrawl/tools/checkstyle/DefaultLoggerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/DefaultLoggerTest.java
@@ -140,34 +140,35 @@ public class DefaultLoggerTest {
 
     @Test
     public void testNullInfoStreamOptions() {
-        final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-        final IllegalArgumentException ex = assertThrows(IllegalArgumentException.class,
-                () -> new DefaultLogger(outputStream, (OutputStreamOptions) null),
-                "IllegalArgumentException expected");
-        assertWithMessage("Invalid error message")
-                .that(ex)
-                .hasMessageThat()
-                        .isEqualTo("Parameter infoStreamOptions can not be null");
+        final OutputStream outputStream = new ByteArrayOutputStream();
+        final DefaultLogger dl =
+                new DefaultLogger(outputStream, (OutputStreamOptions) null);
+
+        dl.addException(new AuditEvent(5000, "myfile"), new IllegalArgumentException("upsss"));
+        dl.auditFinished(new AuditEvent(6000, "myfile"));
+        assertWithMessage("Message should contain exception info")
+                .that(outputStream.toString())
+                .contains("java.lang.IllegalArgumentException: upsss");
     }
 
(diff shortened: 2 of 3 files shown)


romani:
please investigate `finishLocalSetup`
we can move validation of ctor parameters to this method, we just need some initialiation method that is called before all.

relentless-pursuit:
> please investigate `finishLocalSetup` we can move validation of ctor parameters to this method, we just need some initialiation method that is called before all.

@romani, i have tried to use `finishLocalSetup` in the refactored code. please let me know your feedback on the same.

src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java
```diff
@@ -128,13 +134,9 @@ public DefaultLogger(OutputStream infoStream,
                          OutputStream errorStream,
                          OutputStreamOptions errorStreamOptions,
                          AuditEventFormatter messageFormatter) {
-        if (infoStreamOptions == null) {
-            throw new IllegalArgumentException("Parameter infoStreamOptions can not be null");
-        }
+        this.infoStreamOptions = infoStreamOptions;
+        this.errorStreamOptions = errorStreamOptions;
```
romani, line 138: we can safely remove this validation, as logic is null safe.
We will keep junit tests that pass nulls, to control code that it is still null safe.

romani (changes requested):
item:

src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java
```diff
@@ -152,7 +154,12 @@ public DefaultLogger(OutputStream infoStream,
 
     @Override
     protected void finishLocalSetup() {
-        // No code by default
+        if (infoStreamOptions == null) {
```
relentless-pursuit, line 157: @Romani, as you mentioned above https://github.com/checkstyle/checkstyle/pull/14475/files#r1546355855, logic is null-safe. Hence, we can remove the validation. I assume you were talking about null-check. I moved them to finishLocalSetup. Or, null-check should be scrapped entirely?

romani: If null is not possible, we do not need null check.

relentless-pursuit: > If null is not possible, we do not need null check.

I am bit confused. I am initialising them but, i think they can be null. If anyone passes null to the constructor, i guess the values would be still null. Shouldn't we require an exception for that?

romani: If npe is possible, it is better to throw exact exception

romani: please remove such validation.
Logic is null safe.

relentless-pursuit: How do i design the test case which takes null value?

src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java
```diff
@@ -70,6 +70,12 @@ public class DefaultLogger extends AbstractAutomaticBean implements AuditListene
     /** Formatter for the log message. */
     private final AuditEventFormatter formatter;
 
+    /** Options for the info stream. */
+    private OutputStreamOptions infoStreamOptions;
+
+    /** Options for the error stream. */
+    private OutputStreamOptions errorStreamOptions;
+
```
romani, line 78: looks like this not required also, if we removing validatin

romani (changes requested):
item:

src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java
romani: you have junit test (not Input based) for null already, keep it, it will work as test for NONE.

src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java
relentless-pursuit: dome

src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java
relentless-pursuit: i have tried to modify the test case. please let me know your feedback.

src/test/java/com/puppycrawl/tools/checkstyle/DefaultLoggerTest.java
```diff
@@ -139,35 +138,26 @@ public void testNewCtorWithTwoParameters() {
...
+    public void testNullInfoStreamOptions() throws Exception {
         final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-        final IllegalArgumentException ex = assertThrows(IllegalArgumentException.class,
-                () -> new DefaultLogger(outputStream, (OutputStreamOptions) null),
-                "IllegalArgumentException expected");
-        assertWithMessage("Invalid error message")
-                .that(ex)
-                .hasMessageThat()
-                        .isEqualTo("Parameter infoStreamOptions can not be null");
+        final DefaultLogger logger = new DefaultLogger(outputStream, OutputStreamOptions.NONE);
```
romani, line 143: Please keep nulls for ctors, and expect null to be working as none

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/DefaultLoggerTest.java
relentless-pursuit: Done

src/test/java/com/puppycrawl/tools/checkstyle/DefaultLoggerTest.java
```diff
@@ -141,33 +140,25 @@ public void testNewCtorWithTwoParameters() {
...
-                () -> new DefaultLogger(outputStream, (OutputStreamOptions) null),
-                "IllegalArgumentException expected");
-        assertWithMessage("Invalid error message")
-                .that(ex)
-                .hasMessageThat()
-                        .isEqualTo("Parameter infoStreamOptions can not be null");
+        final DefaultLogger logger =
+                new DefaultLogger(outputStream, (OutputStreamOptions) null);
+
+        final boolean closeError = TestUtil.getInternalState(logger, "closeError");
```
romani, line 146: Test should be same as testNewCtorWithTwoParameters() the only difference is null in ctor.

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/DefaultLoggerTest.java
relentless-pursuit: I have done that except i made a slight modification by replacing IllegalStateException with IllegalArgumentException to satisfu the original condition of the unit test.

romani (approved):
Ok to merge

rnveach (changes requested):
CI is failing.

It looks like the first post in this PR is incorrect. This does not resolve the issue, and is only the first. It needs to be fixed.

I also find it hard to believe we barely have any tests to show impact to this change.

@romani @nrmancuso If `OutputStreamWriter` not receives a `null` stream, it will throw an NPE. This is why I can't believe no tests are showing this. Also this technically doesn't resolve the issue since we are still throwing an exception from the constructor.

Are we now saying this is acceptable behavior to throw this exception instead of an `IllegalArgumentException`?

romani:
>OutputStreamWriter not receives a null stream, it will throw an NPE.

I thought we did all to prove that npe is not possible, we have Junit pure unit tests. @relentless-pursuit , do you see a way to reproduce npe ? I do not.

rnveach:
````
new OutputStreamWriter(null, StandardCharsets.UTF_8);

Exception in thread "main" java.lang.NullPointerException
	at java.base/java.io.Writer.<init>(Writer.java:174)
	at java.base/java.io.OutputStreamWriter.<init>(OutputStreamWriter.java:126)
```

`infoStream != errorStream` for this to be hit. The IAE was the only thing blocking it.

romani:
@relentless-pursuit , please add Junit for
`final DefaultLogger dl = new DefaultLogger(null, null, null, null);`

It is a a case that was problem even before, but @rnveach raising it as we try to make ctor that does not throw exception, and this old problem still present.


I think we should just store all args to fields, and do all initialization or validation in finishXxxxxxxx method, and be absolutely safe.

relentless-pursuit:
> @relentless-pursuit , please add Junit for `final DefaultLogger dl = new DefaultLogger(null, null, null, null);`
> 
> It is a a case that was problem even before, but @rnveach raising it as we try to make ctor that does not throw exception, and this old problem still present.
> 
> I think we should just store all args to fields, and do all initialization or validation in finishXxxxxxxx method, and be absolutely safe.

@ormani, @rnveach i have tried to understand the review comments and go through the codebase again. I have tried to incroporate changes. But, I am getting failure wrt coverage issue. However, is the existing code atleast in sync with the review comments?

src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java
```diff
@@ -58,18 +58,30 @@ public class DefaultLogger extends AbstractAutomaticBean implements AuditListene
     public static final String AUDIT_FINISHED_MESSAGE = "DefaultLogger.auditFinished";
 
     /** Where to write info messages. **/
-    private final PrintWriter infoWriter;
+    private PrintWriter infoWriter;
```
romani, line 61: loosing of this `final` is concerning.

romani (changes requested):
items to get clarity:

src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java
romani: I am blocking this PR until https://github.com/spotbugs/spotbugs/issues/2984

romani (approved):
ok to merge.

relentless-pursuit:
Github, generate report for JavadocType/Example2

romani:
@relentless-pursuit , this PR needs to be rebased, treewalker got new property that test-configs already using.

relentless-pursuit:
Github, generate report for AbstractClassName/Example1

rnveach:

PR #14480, opened 2024-02-14
Issue #14084: Remove checker violation for checker

Rohanraj123:
Related #14084

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
@@ -344,7 +344,7 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
             log.debug("IOException occurred.", ioe);
             fileMessages.add(new Violation(1,
                     Definitions.CHECKSTYLE_BUNDLE, EXCEPTION_MSG,
-                    new String[] {ioe.getMessage()}, null, getClass(), null));
+                    new Object[] {ioe.getMessage()}, null, getClass(), null));
         }
         // -@cs[IllegalCatch] There is no other way to obey haltOnException field
         catch (Exception ex) {
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -35,17 +35,6 @@
   <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/Checker.java</fileName>
     <specifier>argument</specifier>
-    <message>incompatible argument for parameter args of Violation constructor.</message>
-    <lineContent>new String[] {ioe.getMessage()}, null, getClass(), null));</lineContent>
-    <details>
-      found   : @Initialized @Nullable String @Initialized @NonNull []
-      required: @Initialized @NonNull Object @Initialized @NonNull []
-    </details>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/Checker.java</fileName>
-    <specifier>argument</specifier>
     <message>incompatible argument for parameter customMessage of Violation constructor.</message>
     <lineContent>new String[] {ioe.getMessage()}, null, getClass(), null));</lineContent>
     <details>


src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -35,6 +35,8 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import javax.annotation.Nonnull;
+
```
Vyom-Yadav, line 39: This is not the annotation checker framework works with. I'd suggest reading the checker documentation. It is a complex topic, but achievable by breaking the problem into bits. https://checkerframework.org/manual/#nullness-checker

Rohanraj123: > This is not the annotation checker framework works with

@NonNull is a  default or expected value. So I think using @NonNull annotation makes no sense. Letme study more on this topic to understand it better.

Vyom-Yadav: Two annotations with the same name don't mean they are the same. Checker framework expects this annotation: https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/NonNull.html. You are importing the wrong one.

Vyom-Yadav: This annotation will work with Checker, no need to use the Checker one. In short, this is fine

romani: https://checkerframework.org/manual/#nullness-related-work

>If your code is already annotated with a different nullness annotation, the Checker Framework can type-check your code. It treats annotations from other tools as if you had written the corresponding annotation from the Nullness Checker

> javax.annotation.Nonnull ‍ ⇒  ‍org.checkerframework.checker.nullness.qual.NonNull ‍

why to use checker annotation ?

romani: https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/NonNull.html
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/annotation/RetentionPolicy.html#RUNTIME

this means we have to keep Checkeras our compile dependency that grow size of `-all` jar, as users do not have benefits.

Vyom-Yadav: Yeah, that's why I added:

> This annotation will work with Checker, no need to use the Checker one. In short, this is fine

Rohanraj123: > https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/NonNull.html https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/annotation/RetentionPolicy.html#RUNTIME
> 
> this means we have to keep Checkeras our compile dependency that grow size of `-all` jar, as users do not have benefits.

Cool means using javax.annotation.Nonnull will be okay.

Vyom-Yadav: Yes

Rohanraj123:
@Vyom-Yadav Let me correct if I am wrong. As per the violation 
Found : @Initialized @Nullable String @Initialized @NonNull []
Req : @Initialized @NonNull Object @Initialized @NonNull []

https://github.com/checkstyle/checkstyle/blob/7724223713fc66a244011e76b91dcc31cf458493/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L344-L349

Changing String to Object and adding @Nonnull annotation before Object initlisation should resolve this issue .  Like : 

Object @Nonnull [] args = {ioe.getMessage()}; because it is getting the exception message so always a non null quantity

nrmancuso:
@Rohanraj123 comment at https://github.com/checkstyle/checkstyle/pull/14480#issuecomment-1951335993 is very difficult to understand, please always use code blocks for code/CLI output. It is not clear to me what the question here is. Please read through the checker docs and test things out locally to unblock yourself.

Rohanraj123:
> comment at [#14480 (comment)](https://github.com/checkstyle/checkstyle/pull/14480#issuecomment-1951335993) is very difficult to understand, please always use code blocks for code/CLI output. It is not clear to me what the question here is.

@nrmancuso I have updated the comment.

Rohanraj123:
> Please read through the checker docs and test things out locally to unblock yourself.

Reading checker docs sir. Also can you please help with something to test things locally.  I dont know why this command : `groovy ./.ci/checker-framework.groovy checker-nullness-optional-interning` isnt working in my terminal. Always shows lot of violations even if there isnt .

nrmancuso:
>isnt working in my terminal

Please google the output, also share here and maybe we can recognize something.

Rohanraj123:
> Please google the output, also share here and maybe we can recognize something.

This is something that i found on google. 
https://github.com/checkstyle/checkstyle/issues/12230

I think Groovy Script execution is still not properly set. This is the same issue I am facing as mentioned in this issue.

romani:
CI is source of truth, if it fails, we cannot merge.

@Rohanraj123 , if this update is complicated, please abandon it and move to another issue.

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -344,7 +344,7 @@ private SortedSet<Violation> processFile(File file) throws CheckstyleException {
             log.debug("IOException occurred.", ioe);
             fileMessages.add(new Violation(1,
                     Definitions.CHECKSTYLE_BUNDLE, EXCEPTION_MSG,
-                    new String[] {ioe.getMessage()}, null, getClass(), null));
+                    new Object[] {ioe.getMessage()}, null, getClass(), null));
```
romani, line 347: it complain about nulls for `new Violation` not a `String[]`

PR #14483, opened 2024-02-14
Issue #14019: Resolve pitest supressions for ScopeUtil#getSurrounding…

mahfouz72:
issue #14019:
## Mutation
https://github.com/checkstyle/checkstyle/blob/7b8ceb178cfe52019b7214e23a89e7c3b48c24d9/config/pitest-suppressions/pitest-utils-suppressions.xml#L12-L19

## Modules
```
RequireThisCheck          
DesignForExtensionCheck   
JavadocStyleCheck         
javadocTypeCheck          
javadocVariableCheck      
MissingJavadocMethodCheck
MissingJavadocTypeCheck  
``` 
Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/7bdb426c3d091c2936e3e9726cd8cd47350e1d3b/check.xml

Diff Regression projects:  https://gist.githubusercontent.com/mahfouz72/649bd1dced8e2cde1d5b5aa148ab93a1/raw/db4ff628cf5ecdf578729cd4811f880de1fca868/projects-to-test-on.properties

### DIFF
--- a/config/pitest-suppressions/pitest-utils-suppressions.xml
+++ b/config/pitest-suppressions/pitest-utils-suppressions.xml
@@ -12,15 +12,6 @@
   <mutation unstable="false">
     <sourceFile>ScopeUtil.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.utils.ScopeUtil</mutatedClass>
-    <mutatedMethod>getSurroundingScope</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/api/DetailAST::getParent with receiver</description>
-    <lineContent>for (DetailAST token = node.getParent();</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>ScopeUtil.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.utils.ScopeUtil</mutatedClass>
     <mutatedMethod>isInBlockOf</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
     <description>removed conditional - replaced equality check with true</description>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheck.java
@@ -300,11 +300,9 @@ public class JavadocStyleCheck
             final Scope surroundingScope = ScopeUtil.getSurroundingScope(ast);
 
             check = customScope.isIn(scope)
-                    && (surroundingScope == null || surroundingScope.isIn(scope))
-                    && (excludeScope == null
-                        || !customScope.isIn(excludeScope)
-                        || surroundingScope != null
-                            && !surroundingScope.isIn(excludeScope));
+                    && surroundingScope.isIn(scope)
+                    && (excludeScope == null || !customScope.isIn(excludeScope)
+                            || !surroundingScope.isIn(excludeScope));
         }
         return check;
     }
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
@@ -357,16 +357,11 @@ public class JavadocTypeCheck
      * @return whether we should check a given node.
      */
     private boolean shouldCheck(DetailAST ast) {
-        final Scope customScope = ScopeUtil.getScope(ast);
         final Scope surroundingScope = ScopeUtil.getSurroundingScope(ast);
 
-        return customScope.isIn(scope)
-            && (surroundingScope == null || surroundingScope.isIn(scope))
-            && (excludeScope == null
-                || !customScope.isIn(excludeScope)
-                || surroundingScope != null
-                && !surroundingScope.isIn(excludeScope))
-            && !AnnotationUtil.containsAnnotation(ast, allowedAnnotations);
+        return surroundingScope.isIn(scope)
+                && (excludeScope == null || !surroundingScope.isIn(excludeScope))
+                && !AnnotationUtil.containsAnnotation(ast, allowedAnnotations);
     }
 
     /**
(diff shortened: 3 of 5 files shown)


mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

nrmancuso:
@mahfouz72 what you’ve hardcoded looks correct to me; sometimes, we get cascading survivals like this. I can help take a look in the next few days, but I encourage you to begin trying to analyze the code to understand why this is happening.

romani:
Sounds like just another case of point 2 from https://github.com/checkstyle/checkstyle/issues/13142

nrmancuso:
> Sounds like just another case of point 2 from #13142 

Yea, this is a performance optimization that pitest is catching here, but doesn’t explain the new surviving mutations.

nrmancuso (changes requested):
I am fine to remove this surviving mutation and extend list to include the newly discovered mutations. The new surviving mutations seem to be more interesting and likely to clean up our codebase (or find better test cases in regression reports). @mahfouz72 please update the suppressed surviving mutation list to make CI happy.

Edit: looks like code coverage dropped, we need to investigate this.

romani:
jacoco coverage is reduced https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=19714&view=logs&j=e663805c-ee13-5a24-f979-d725e0346c8e&t=40ff6e43-c5d7-5753-f947-23012624b0c6&l=1788

somthing is weird.... not good.

nrmancuso:
> jacoco coverage is reduced https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=19714&view=logs&j=e663805c-ee13-5a24-f979-d725e0346c8e&t=40ff6e43-c5d7-5753-f947-23012624b0c6&l=1788
> 
> somthing is weird.... not good.

Yeah, just noticed this. I wonder if this is due to utility usage being spread across several pitest profiles.

nrmancuso:
Ok, @mahfouz72 it looks like we are now handling an extra condition in this utility method that we were previously forcing checks to take care of. Now we have a check on `Scope#isIn` on the original node:
```
                if (returnValue == null || returnValue.isIn(tokenScope)) {
                    returnValue = tokenScope;
                }
```

Please investigate each new surviving mutation and confirm this.

nrmancuso:
@mahfouz72 this is probably from what I mentioned in https://github.com/checkstyle/checkstyle/pull/14483#issuecomment-1946162898, we might not have all tests in the profile that is failing

mahfouz72:
- I tried to investigate the code to hardcode the mutation but got many cascading survivals and the coverage drop idk why is that. is this normal or it is just a tricky one?
- I think killing the mutation with tests would be a better solution
- mutation is killed and no new survivals

### Explanation
consider this AST
 assume we want to pass the `CLASS_DEF` node to the `getSurroundingScope()`
```
`--CLASS_DEF -> CLASS_DEF [1:0]     <--- for the original code we start from the parent of this node which is null so loop breaks and null is returned
    |--MODIFIERS -> MODIFIERS [1:0]                                                                                     
    |--LITERAL_CLASS -> class [1:0]
    |--IDENT -> Test [1:6]
    `--OBJBLOCK -> OBJBLOCK [1:11]
        |--LCURLY -> { [1:11]
        `--RCURLY -> } [2:0]
```
```
`--CLASS_DEF -> CLASS_DEF [1:0]  <-- the mutated code starts from the node itself and as it is a `TypeDeclaration`  we executed getScope method and searched for modifiers the returned value will be package
    |--MODIFIERS -> MODIFIERS [1:0]
    |--LITERAL_CLASS -> class [1:0]
    |--IDENT -> Test [1:6]
    `--OBJBLOCK -> OBJBLOCK [1:11]
        |--LCURLY -> { [1:11]
        `--RCURLY -> } [2:0]
```
So I called  `getSurroundingScope()` with `CLASS_DEF` node and assert that the scope has to be null

@nrmancuso @romani  please review

mahfouz72:
Github, generate report

nrmancuso:
It looks like we have a chance to clean up a lot of code in a few checks here, we typically use synthetic tests only as a last resort. I would prefer to at least investigate the following, it would be a good exercise if nothing else:
1. Hardcode mutation
2. Remove code where we lost coverage
3. Add failing tests to corresponding pitest profile


And see where we land. @rnveach what do you think?

mahfouz72:
@nrmancuso I am ok with that but is it always valid to hard code mutation because there is a case that the output differs before and after hardcoding mutation (When we pass a `CLASS_DEF` and it is the first node in the tree) See https://github.com/checkstyle/checkstyle/pull/14483#issuecomment-1947124314

nrmancuso:
> there is a case that the output differs before and after hardcoding mutation

If you can find a test case that uses real code and a check config, this is fine. Otherwise, we are just keeping around a bunch of code that checks duplicate conditions for no reason.

mahfouz72:
@nrmancuso please review this all is good except these two suppressed mutations pitest reported no coverage for them and when I remove them the normal JUnit test for `JavadocStyleCheck` fails

mahfouz72:
Github, generate report

mahfouz72:
@nrmancuso @romani please review all is good except the two suppressed mutations pitest reported them as surviving mutation  and when I remove them the normal JUnit test for `JavadocStyleCheck` fails (the coverage didn't drop)
```
Line Coverage (for mutated classes only): 1757/1757 (100%)
>> Generated 3413 mutations Killed 3381 (99%)
>> Mutations with no coverage 0. Test strength 99%
>> Ran 10548 tests (3.09 tests per mutation)
```
when I run `mvn verify` I got this error 

 `Rule violated for class com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocStyleCheck: branches covered ratio is 0.97, but expected minimum is 1.00
`

romani:
>when I run mvn verify I got this error

Unfortunately this is hard blocker to merge.

nrmancuso:
@mahfouz72 generate code coverage report and figure out where coverage dropped, see our wiki for how to do this

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

nrmancuso (commented):
I am good with these changes, @mahfouz72 thanks a lot for this update! Killing this mutation ended up helping us to clean up code in a few checks. I am good to approve this once we have a clean check regression report.

nrmancuso (approved):
Awesome.

romani (approved):
good job!!!
it was not easy but you did it.

PR #14484, opened 2024-02-14
Issue #14482: Remove UnmodifiableCollectionUtil and use pitest funmod…

### LINKED ISSUE #14482: Remove UnmodifiableCollectionUtil and use pitest funmodifiablecollection feature
UnmodifiableCollectionUtil as created at https://github.com/checkstyle/checkstyle/pull/13127/files
to bypass pitest issue mutation of unmodified collections.

Pitest released a fix at https://github.com/hcoles/pitest/issues/1300 and we upgraded to the version of pitest at https://github.com/checkstyle/checkstyle/commit/68f7af17f6a0020a720745bdce2d19b7d8dab5c5

We need to activate pitest feature and revert UnmodifiableCollectionUtil.

Kevin222004:
part of #14482

### DIFF
--- a/config/jsoref-spellchecker/whitelist.words
+++ b/config/jsoref-spellchecker/whitelist.words
@@ -476,6 +476,7 @@ ftpclient
 fullident
 fullname
 func
+funmodifiablecollection
 Fwhitespace
 Fx
 gav
--- a/pom.xml
+++ b/pom.xml
@@ -3076,6 +3076,9 @@
                 <avoidCallsTo>com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil</avoidCallsTo>
                 <avoidCallsTo>org.apache.commons.logging</avoidCallsTo>
               </avoidCallsTo>
+              <features>
+                <feature>+funmodifiablecollection</feature>
+              </features>
               <excludedMethods>
                 <!-- destroy method was added in case module had to free up resources
                 before ending, but currently it does ThreadLocal cleanup,
@@ -3313,6 +3316,9 @@
               <avoidCallsTo>
                 <avoidCallsTo>com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil</avoidCallsTo>
               </avoidCallsTo>
+              <features>
+                <feature>+funmodifiablecollection</feature>
+              </features>
               <excludedTestClasses>
                 <param>*.Input*</param>
               </excludedTestClasses>
@@ -4408,6 +4414,9 @@
                 <!-- performance optimization -->
                 <param>initStringBuilderWithOptimalBuffer</param>
               </excludedMethods>
+              <features>
+                <feature>+funmodifiablecollection</feature>
+              </features>
               <coverageThreshold>100</coverageThreshold>
               <mutationThreshold>96</mutationThreshold>
               <timeoutFactor>${pitest.plugin.timeout.factor}</timeoutFactor>
@@ -4481,6 +4490,9 @@
                 <!-- cause of https://github.com/checkstyle/checkstyle/issues/3605 -->
                 <param>setFeaturesBySystemProperty</param>
               </excludedMethods>
+              <features>
+                <feature>+funmodifiablecollection</feature>
+              </features>
               <avoidCallsTo>
                 <!-- cause of https://github.com/checkstyle/checkstyle/issues/3605 -->
                 <avoidCallsTo>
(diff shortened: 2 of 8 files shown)


romani:
https://github.com/checkstyle/checkstyle/actions/runs/7906562343/job/21581681565?pr=14484#step:6:456

Sounds like it is not working.

https://pitest.org/quickstart/maven/#features

>Available options are shown in the console output when verbose logging is enabled.

We need to check if such feature is printed.

From https://github.com/checkstyle/checkstyle/actions/runs/7906562343/job/21581681565?pr=14484#step:6:224

>[INFO] Found plugin : Filter mutations to defensive return wrappers such as unmodifiableCollection

romani:
Why only 4 survived? We have a bit more usages.

src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java
```diff
@@ -499,7 +499,7 @@ public void removeChildren() {
     public List<Token> getHiddenBefore() {
         List<Token> returnList = null;
         if (hiddenBefore != null) {
-            returnList = UnmodifiableCollectionUtil.unmodifiableList(hiddenBefore);
+            returnList = Collections.unmodifiableList(hiddenBefore);
```
romani, line 502: Pitest has defect to detect usage of this method without return

romani (changes requested):
Items to wait for fix in pitest:

src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java
```diff
@@ -511,11 +510,9 @@ public List<Token> getHiddenBefore() {
      * @return list of comment tokens
      */
     public List<Token> getHiddenAfter() {
```
nrmancuso, line 512: What is the impact of returning an empty list instead of `null` here?

romani: I do not know, we can think of this later on in separate PR.

romani: there is discussion at https://github.com/hcoles/pitest/issues/1300

romani:
@Kevin222004 , let's revert DetailAstImpl cases that pitest still complain on.
At least we Can merge all others.

src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java
rnveach: > What is the impact of returning an empty list instead of null here?

It looks like the only impact here is changing `JavaParser` which is the only caller of this code and has a `null` check. This is probably left over code from the mentality if it isn't "populated" set it to `null`.

https://github.com/search?q=repo%3Acheckstyle%2Fcheckstyle%20getHiddenAfter&type=code

This wouldn't really impact downstream applications if someone was using this public method, since I would assume most programmers would treat `null` and an empty list the same.

romani (approved):
ok to merge.

I hope there will be more fixes in pitest to resolve this completetly, but we can merge this improvement now.

src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java
romani: do you agree to update this in separate issue ?

nrmancuso: > do you agree to update this in separate issue ?

The title of the issue for this PR is `Remove UnmodifiableCollectionUtil...`, and NOT returning null here would allow us to readily do a simple return. So - why should this be done in a separate issue? In 2024, returning a `null` collection is frowned upon anyway.

nrmancuso: >since I would assume most programmers would treat null and an empty list the same.

Not sure I agree here, but at least going from `null` to an empty list should be a backwards compatible change (not breaking)

romani: Ok, looks like others wants to kill this null now.
@Kevin222004 , please do it.

nrmancuso: @Kevin222004 please ping me once we are good here

Kevin222004: @nrmancuso sure.

I have not started working. I will finish this as soon as possible.

romani:
@nrmancuso , I am ok to merge this PR. We don't know when other fixes in pitest library will be done

nrmancuso (approved):
Ok to merge after CI passes, I have left a comment in the issue to make sure that we don't close it.

nrmancuso:
CI failure is not related to changes here.

PR #14499, opened 2024-02-18
Issue #14019: Kill mutation in ModuleReflectionUtil

mahfouz72:
Issue #14019: 
## Mutation
https://github.com/checkstyle/checkstyle/blob/a4411aa08c3a9744139067dc38ced282adbf9fcc/config/pitest-suppressions/pitest-utils-suppressions.xml#L3-L10

## Regression

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/471fd9cd53eb2c56da8a5748f832a3987c9235ff/check.xml

Diff Regression projects: https://gist.githubusercontent.com/mahfouz72/649bd1dced8e2cde1d5b5aa148ab93a1/raw/db4ff628cf5ecdf578729cd4811f880de1fca868/projects-to-test-on.properties

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtilTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtilTest.java
@@ -20,10 +20,14 @@
 package com.puppycrawl.tools.checkstyle.utils;
 
 import static com.google.common.truth.Truth.assertWithMessage;
+import static com.puppycrawl.tools.checkstyle.PackageObjectFactory.BASE_PACKAGE;
 import static com.puppycrawl.tools.checkstyle.internal.utils.TestUtil.isUtilsClassHasPrivateConstructor;
 
 import java.io.File;
+import java.io.IOException;
+import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 
 import org.junit.jupiter.api.Test;
 
@@ -73,6 +77,21 @@ public class ModuleReflectionUtilTest {
         assertWithMessage("Should return true when checkstyle module is passed")
                 .that(ModuleReflectionUtil.isCheckstyleModule(RootModuleClass.class))
                 .isTrue();
+    }
+
+    /**
+    * This test case is designed to verify the behavior of getCheckstyleModules method.
+    * It is provided with a package name that does not contain any checkstyle modules.
+    * It ensures that ModuleReflectionUtil.getCheckstyleModules is returning an empty set.
+    */
+    @Test
+    public void testGetCheckStyleModules() throws IOException {
+        final ClassLoader classLoader = ClassLoader.getSystemClassLoader();
+        final Set<String> packages = Collections.singleton(BASE_PACKAGE + ".checks.javadoc.utils");
+
+        assertWithMessage("specified package has no checkstyle modules")
+                .that(ModuleReflectionUtil.getCheckstyleModules(packages, classLoader))
+                .isEmpty();
     }
 
     @Test
--- a/config/pitest-suppressions/pitest-utils-suppressions.xml
+++ b/config/pitest-suppressions/pitest-utils-suppressions.xml
@@ -1,13 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <mutation unstable="false">
-    <sourceFile>ModuleReflectionUtil.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.utils.ModuleReflectionUtil</mutatedClass>
-    <mutatedMethod>getCheckstyleModules</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
-    <description>replaced call to java/util/stream/Stream::filter with receiver</description>
-    <lineContent>.filter(ModuleReflectionUtil::isCheckstyleModule)</lineContent>
-  </mutation>
 
   <mutation unstable="false">
     <sourceFile>ScopeUtil.java</sourceFile>


src/test/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtilTest.java
```diff
@@ -75,6 +79,17 @@ public void testIsCheckstyleModule() {
...
 
+    @Test
+    public void testGetCheckstyleModules() throws IOException {
+        final ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+        final Set<String> packages =
+                Collections.singleton(BASE_PACKAGE + ".checks" + ".javadoc" + ".utils");
+
+        assertWithMessage("specified package has no checkstyle modules")
+                .that(ModuleReflectionUtil.getCheckstyleModules(packages, classLoader))
+                .isEmpty();
```
romani, line 90: Sorry we do not accept pure Junit test.
We need to cover by verifyxxxxx somehow.
If not possible, we need to write long explanation and put it over test method.

romani (changes requested):
Items

nrmancuso:
@mahfouz72 we still need to follow https://github.com/checkstyle/checkstyle/issues/12341

src/test/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtilTest.java
mahfouz72: @romani how am I supposed to cover this with verify? we aren't testing a check or a check property here how verify with input file can help?

mahfouz72:
@nrmancuso Sure. I thought regression aren't necessary here because we aren't testing any checks. but if it is required what checks to put in the config.xml file?

nrmancuso:
> @nrmancuso Sure. I thought regression aren't necessary here because we aren't testing any checks. but if it is required what checks to put in the config.xml file?

Please point to where we did the first two steps in the flow chart.

mahfouz72:
> Please point to where we did the first two steps in the flow chart.

okay let me explain my thoughts more clearly. check this
```
public static Set<Class<?>> getCheckstyleModules(
            Collection<String> packages, ClassLoader loader) throws IOException {
        final ClassPath classPath = ClassPath.from(loader);
        return packages.stream()
                .flatMap(pkg -> classPath.getTopLevelClasses(pkg).stream())
                .map(ClassPath.ClassInfo::load)
                .filter(ModuleReflectionUtil::isCheckstyleModule)     <-- mutated line
                .collect(Collectors.toUnmodifiableSet());
    }
```
I know that we have to hardcode mutation and check test and report as  first steps but I decided to skip this  because:
- removing this line that pitest mutated is not a good change as we only need checkstyle modules to be picked up. We don't want  random files and that is the purpose of `getCheckstyleModule()` filtering files in package and return the set of checkstyle's module classes if we remove `.filter` it will return all the classes in the package
- I found a test case that can kill this mutation by testing the function on one of our packages that has no checkstyle modules
@nrmancuso please correct me if anything wrong. and if it is always required to follow steps at #12341 in all cases please let me know

nrmancuso:
>if it is always required to follow steps at https://github.com/checkstyle/checkstyle/issues/12341 in all cases please let me know

The first two steps require the least amount of thinking, time, and work :) if `mvn clean verify` fails when this line is removed, you have found a way to kill the mutation with no actual analysis of the code.

If we were to remove this call to `Stream#filter`,  what would break (or could) break? Or is this just a performance optimization?

mahfouz72:
Github, generate report

src/test/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtilTest.java
romani: you skipping this conditions:
https://github.com/checkstyle/checkstyle/blob/a2d834fb1517cf9056dc70f4c8f1c4999d676f52/src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java#L74-L78

can we make internal test module that do some violation or make it abstract class to let initialization to fail with exception.
Even easier: use XpathFileGenerator as module in Input file, and assert that it did not run.

mahfouz72: @romani sorry but please can you explain more what should I do. I think i am not understanding well

src/test/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtilTest.java
```diff
@@ -179,6 +185,14 @@ public void testKeepEclipseHappy() {
             .isEqualTo(1);
     }
 
+    @Test
+    public void testGetCheckstyleModules() throws Exception {
+        final String filePath = getPath("InputUtil.java");
```
nrmancuso, line 190:
```suggestion
        final String filePath = getPath("InputModuleReflectionUtilLoadInternalFilter.java");
```

Should make CI happy.

src/test/resources/com/puppycrawl/tools/checkstyle/utils/modulereflectionutil/InputUtil.java
```diff
@@ -0,0 +1,9 @@
+/*
+com.puppycrawl.tools.checkstyle.XpathFileGeneratorAstFilter
+
+*/
+
+package com.puppycrawl.tools.checkstyle.utils.modulereflectionutil;
+
+public class InputUtil {
```
nrmancuso, line 8:
```suggestion
public class InputModuleReflectionUtilLoadInternalFilter {
```

mahfouz72:
@nrmancuso @romani I tried to cover by verify but it didn't work. I debugged and we didn't even reach to test the mutated method maybe I didn't get the flow well. how to test this mutated method using verify exactly? and can we use pure junit as I did before .. like this
 ```
assertWithMessage("specified package has no checkstyle modules")
                .that(ModuleReflectionUtil.getCheckstyleModules(packages, classLoader))
                .isEmpty();
```

**Comment by nrmancuso** [comment]
@mahfouz72 what were the results of removing the code and running `mvn clean verify`? Are you able to answer my questions at https://github.com/checkstyle/checkstyle/pull/14499#issuecomment-1951336284? Until we understand exactly why the code is here, and what would fail if it is not present, we really can’t proceed. Even if we were to accept a pure unit test, it would require a detailed explanation containing at least what I have asked above, in addition to why we can’t hit the code.

**Comment by nrmancuso** [comment]
> This is as far as I understand. please correct me if I am missing anything

You know more about this area of the code now than I do :) Thanks fo the detailed explanation.

@romani I am OK with a pure unit test for this one, as long as we have a good explanation of why (a summation of @mahfouz72 's comments from above) in a comment on the test method. I can't think of any reasonable way to test this otherwise, and the benefit of filtering modules here is clear.

**Comment by mahfouz72** [comment]
killed the mutation using pure unit test waiting for @romani approval and checking if test comment is good

File: src/test/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtilTest.java [hunk 4]
```diff
@@ -75,6 +79,21 @@ public void testIsCheckstyleModule() {
                 .isTrue();
     }
 
+    /**
+    * This test case is designed to verify the behavior of getCheckstyleModules method.
+    * It is provided with a package name that does not contain any checkstyle modules.
+    * It ensures that ModuleReflectionUtil.getCheckstyleModules is returning an empty set.
+    */
```
nrmancuso, line 86: @mahfouz72 it might be good to make sure that @romani is good with this spending time updating this test. If this is the route we take, please mention your research here:
 - exactly which part of the code we are hitting with this test (the filter operation)
 - why we couldn't make any normal test fail
 - why we didn't want to remove the code (we don't want to load everything in the class path)

mahfouz72: okay I am waiting for @romani approval before updating the comment

mahfouz72: @romani ping

romani: please consider https://github.com/checkstyle/checkstyle/pull/14499/files#r1499361210

src/test/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtilTest.java
mahfouz72: @romani please check this https://github.com/checkstyle/checkstyle/pull/14499#issuecomment-1952291697, we are not even hitting our mutated function when using verify and input config. I have tried many other tests that use verify and we aren't hitting it.

src/test/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtilTest.java
romani: I am confirming that making Input based test is possible but will be big hack.
patch to get to call of this method:
```
 $ git diff
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java b/src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java
index 1124219..9c6815b 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java
@@ -60,7 +60,7 @@ public final class ModuleReflectionUtil {
         return packages.stream()
                 .flatMap(pkg -> classPath.getTopLevelClasses(pkg).stream())
                 .map(ClassPath.ClassInfo::load)
-                .filter(ModuleReflectionUtil::isCheckstyleModule)
+                //.filter(ModuleReflectionUtil::isCheckstyleModule)
                 .collect(Collectors.toUnmodifiableSet());
     }
 
diff --git a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
index 0b9de5c..f45ed14 100644
--- a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
@@ -179,7 +179,8 @@ public final class InlineConfigParser {
             // Inlined config is not supported for non java files.
             "com.puppycrawl.tools.checkstyle.checks.OrderedPropertiesCheck",
             "com.puppycrawl.tools.checkstyle.checks.UniquePropertiesCheck",
-            "com.puppycrawl.tools.checkstyle.checks.TranslationCheck"
+            "com.puppycrawl.tools.checkstyle.checks.TranslationCheck",
+        "LineSeparatorOption"
     );
 
     /** Stop instances being created. **/
@@ -318,6 +319,8 @@ public final class InlineConfigParser {
                 "com.puppycrawl.tools.checkstyle.checks.whitespace.NoWhitespaceAfterCheck");
         moduleMappings.put("SummaryJavadoc",
                 "com.puppycrawl.tools.checkstyle.checks.javadoc.SummaryJavadocCheck");
+        moduleMappings.put("LineSeparatorOption",
+                "LineSeparatorOption");
         if (moduleMappings.containsKey(moduleName)) {
             fullyQualifiedClassName = moduleMappings.get(moduleName);
         }
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/finalparameters/InputFinalParameters.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/finalparameters/InputFinalParameters.java
index a69b300..636a960 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/finalparameters/InputFinalParameters.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/finalparameters/InputFinalParameters.java
@@ -1,5 +1,5 @@
 /*
-FinalParameters
+LineSeparatorOption
 ignorePrimitiveTypes = (default)false
 tokens = (default)METHOD_DEF, CTOR_DEF
```

romani (approved):
ok to merge, thanks a lot

PR #14503, opened 2024-02-18
improve text block grammar

nrmancuso:
Minor update to ease review at https://github.com/checkstyle/checkstyle/pull/14390

Our grammar was previously able to parse this illegal escape sequence in a text block:
```
➜  src cat Test.java 

public class Test {
    String s = """
            \{}
            """;
}
➜  src javac --enable-preview --release 21 Test.java
Test.java:3: error: processor missing from string template expression
    String s = """
               ^
Note: Test.java uses preview features of Java SE 21.
Note: Recompile with -Xlint:preview for details.
1 error
➜  src java -jar checkstyle-10.13.0-all.jar -t Test.java
COMPILATION_UNIT -> COMPILATION_UNIT [2:0]
`--CLASS_DEF -> CLASS_DEF [2:0]
    |--MODIFIERS -> MODIFIERS [2:0]
    |   `--LITERAL_PUBLIC -> public [2:0] blocks
    |--LITERAL_CLASS -> class [2:7]
    |--IDENT -> Test [2:13]
    `--OBJBLOCK -> OBJBLOCK [2:18]
        |--LCURLY -> { [2:18]
        |--VARIABLE_DEF -> VARIABLE_DEF [3:4]
        |   |--MODIFIERS -> MODIFIERS [3:4]
        |   |--TYPE -> TYPE [3:4]
        |   |   `--IDENT -> String [3:4]
        |   |--IDENT -> s [3:11]
        |   |--ASSIGN -> = [3:13]
        |   |   `--EXPR -> EXPR [3:15]
        |   |       `--TEXT_BLOCK_LITERAL_BEGIN -> """ [3:15]
        |   |           |--TEXT_BLOCK_CONTENT -> \n            \{}\n             [3:18]
        |   |           `--TEXT_BLOCK_LITERAL_END -> """ [5:12]
        |   `--SEMI -> ; [5:15]
        `--RCURLY -> } [6:0]
```

Successfully parsing such escapes made it difficult to add support for text
block templates in https://github.com/checkstyle/checkstyle/pull/14390.

------------------------------------------------------------
Reports:

Link to report generation: https://github.com/nrmancuso/checkstyle-diff-report-generator/actions/runs/7961175874

Projects file: projects-to-test-on.properties
Patch branch: improve-text-block-grammar

https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/improve-text-block-grammar/2024-02-19-T-17-44-24/sevntu-check-regression_part_1/index.html

[...]

PR #14525, opened 2024-02-23
Issue #3502: update VariableDeclarationUsageDistance to do violation to move variables closer to block

### LINKED ISSUE #3502: VariableDeclarationUsageDistance: no violation to move variables closer to block
```
$ cat TestClass.java
public class TestClass {
    void method() {
        int a = 3;
        int b = 2;

        run();
        run();
        run();
        run();
        run();
        run();
        run();
        run();
        run();
        run();
        run();
        run();

        for (int i = 0; i < 10; i++) {
            a = a + b;
        }
    }

    public void run() {};
}

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">

    <module name="TreeWalker">
<property name="validateBetweenScopes" value="false" />
</module>
    </module>
</module>

$ java -jar checkstyle-7.1.2-all.jar -c TestConfig.xml TestClass.java
Starting audit...
Audit done.
```

I use `VariableDeclarationUsageDistance` to make sure my variables are close to where they are being used. It makes reading the code easier most times.
In the example, there are no violations that variables a and b should be closer to the block that uses them. They are not very close to the block at all, and are not declared `final`. I am expecting them to put out violations since they are so far away, just like if the for loop didn't exist.
On the [XDoc](http://checkstyle.sourceforge.net/config_coding.html#VariableDeclarationUsageDistance) in the `Notes` section, I understand from case 1 that the check will not require me to put variables inside the block. I am just asking that the variables be placed close enough to the block to satisfy the distance, assuming the block in question is considered 1 unit as a whole.

From the example above, I am expecting the following violations:
```
[ERROR] TestClass.java:3: Distance between variable 'a' declaration and its first usage is 13, but allowed 1.  Consider to make that variable as final if you still need to store its value in advance (before method calls that might do side effect on original value). [VariableDeclarationUsageDistance]
[ERROR] TestClass.java:4: Distance between variable 'b' declaration and its first usage is 13, but allowed 1.  Consider to make that variable as final if you still need to store its value in advance (before method calls that might do side effect on original value). [VariableDeclarationUsageDistance]
```

I am only expecting no violations with the config, if the code is defined as the following:

```
    void method() {
        run();
        run();
        run();
        run();
        run();
        run();
        run();
        run();
        run();
        run();
        run();
        run();

        int a = 3;
        int b = 2;

        for (int i = 0; i < 10; i++) {
            a = a + b;
        }
    }
```

I cannot turn on validateBetweenScopes because it acts as if the braces do not exist and with a distance of 1 it will ask for variables to be moved inside the braces which interfers with loops.

mahfouz72:
Resolves #3502:
- updated VariableDeclarationUsageDistance to do violation to move variables closer to the block
- updated corresponding tests to match the new functionality of the check

## Regression
I will generate 2 reports difference is expected for report-2

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/ce878f7db10dd474a7f6aa53ada50c7c/raw/68d7ec7bcef08b49d6deaa712d2e18cf8c328dcd/check1.xml
Report label: report-2

I generated two report
- [report-1](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/17c1f8f_2024173315/reports/diff/index.html) (no diff) with this [config](https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/e82a7576e6372a18bb4794217b6713b475f0e4f9/check.xml) to check that the check behavior didn't change when `validateBetweenScopes` is on 

- [report-2](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/17c1f8f_2024192425/reports/diff/index.html) (expected diff) with this [config](https://gist.githubusercontent.com/mahfouz72/ce878f7db10dd474a7f6aa53ada50c7c/raw/68d7ec7bcef08b49d6deaa712d2e18cf8c328dcd/check1.xml) to check that we raise a violation to move variable near to the block when `validateBetweenScopes` is off as stated in the issue

### DIFF
--- a/src/it/java/com/google/checkstyle/test/chapter4formatting/rule4822variabledistance/VariableDeclarationUsageDistanceTest.java
+++ b/src/it/java/com/google/checkstyle/test/chapter4formatting/rule4822variabledistance/VariableDeclarationUsageDistanceTest.java
@@ -43,6 +43,7 @@ public class VariableDeclarationUsageDistanceTest extends AbstractGoogleModuleTe
             "219:9: " + getCheckMessage(clazz, msgExt, "t", 5, 3),
             "483:9: " + getCheckMessage(clazz, msgExt, "myOption", 7, 3),
             "495:9: " + getCheckMessage(clazz, msgExt, "myOption", 6, 3),
+            "508:9: " + getCheckMessage(clazz, msgExt, "count", 4, 3),
         };
 
         final Configuration checkConfig =
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
@@ -306,7 +306,7 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
         while (!firstUsageFound && currentAst != null) {
             if (currentAst.getFirstChild() != null) {
                 if (isChild(currentAst, variableIdentAst)) {
-                    dist = getDistToVariableUsageInChildNode(currentAst, variableIdentAst, dist);
+                    dist = getDistToVariableUsageInChildNode(currentAst, dist);
                     variableUsageAst = currentAst;
                     firstUsageFound = true;
                 }
@@ -324,11 +324,10 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
      * Returns the distance to variable usage for in the child node.
      *
      * @param childNode child node.
-     * @param varIdent variable variable identifier.
      * @param currentDistToVarUsage current distance to the variable usage.
      * @return the distance to variable usage for in the child node.
      */
-    private static int getDistToVariableUsageInChildNode(DetailAST childNode, DetailAST varIdent,
+    private static int getDistToVariableUsageInChildNode(DetailAST childNode,
                                                          int currentDistToVarUsage) {
         DetailAST examineNode = childNode;
         if (examineNode.getType() == TokenTypes.LABELED_STAT) {
@@ -336,6 +335,7 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
         }
 
         int resultDist = currentDistToVarUsage;
+
         switch (examineNode.getType()) {
             case TokenTypes.SLIST:
                 resultDist = 0;
@@ -345,17 +345,12 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
             case TokenTypes.LITERAL_DO:
             case TokenTypes.LITERAL_IF:
             case TokenTypes.LITERAL_SWITCH:
-                if (isVariableInOperatorExpr(examineNode, varIdent)) {
-                    resultDist++;
-                }
-                else {
-                    // variable usage is in inner scope
-                    // reset counters, because we can't determine distance
-                    resultDist = 0;
-                }
+                // variable usage is in inner scope, treated as 1 block
+                // or in operator expression, then distance + 1
+                resultDist++;
                 break;
             default:
-                if (examineNode.findFirstToken(TokenTypes.SLIST) == null) {
+                if (childNode.findFirstToken(TokenTypes.SLIST) == null) {
                     resultDist++;
                 }
                 else {
@@ -706,22 +701,6 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
                 break;
             }
             ast = ast.getNextSibling();
-        }
-
-        // Variable may be met in ELSE declaration
-        // So, check variable usage in these declarations.
-        if (!isVarInOperatorDeclaration) {
-            final DetailAST elseBlock = operator.getLastChild();
-
-            if (elseBlock.getType() == TokenTypes.LITERAL_ELSE) {
-                // Get IF followed by ELSE
-                final DetailAST firstNodeInsideElseBlock = elseBlock.getFirstChild();
-
-                if (firstNodeInsideElseBlock.getType() == TokenTypes.LITERAL_IF) {
-                    isVarInOperatorDeclaration =
-                        isVariableInOperatorExpr(firstNodeInsideElseBlock, variable);
-                }
-            }
         }
 
         return isVarInOperatorDeclaration;
--- a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
@@ -301,7 +301,6 @@ public final class InlineConfigParser {
 
     private static String getFullyQualifiedClassName(String filePath, String moduleName)
             throws CheckstyleException {
-        String fullyQualifiedClassName;
         // This is a hack until https://github.com/checkstyle/checkstyle/issues/13845
         final Map<String, String> moduleMappings = new HashMap<>();
         moduleMappings.put("ParameterNumber",
@@ -320,6 +319,7 @@ public final class InlineConfigParser {
                 "com.puppycrawl.tools.checkstyle.checks.javadoc.SummaryJavadocCheck");
         moduleMappings.put("LineLength",
                 "com.puppycrawl.tools.checkstyle.checks.sizes.LineLengthCheck");
+        String fullyQualifiedClassName;
         if (moduleMappings.containsKey(moduleName)) {
             fullyQualifiedClassName = moduleMappings.get(moduleName);
         }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceCloseToBlock.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceCloseToBlock.java
@@ -1,0 +1,59 @@
+/*
+VariableDeclarationUsageDistance
+allowedDistance = 1
+ignoreVariablePattern = (default)
+validateBetweenScopes = false
+ignoreFinal = false
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.variabledeclarationusagedistance;
+
+public class InputVariableDeclarationUsageDistanceCloseToBlock {
+    void method() {
+        int a = 3; // violation 'Distance between .* declaration and its first usage is 13.'
+        int b = 2; // violation 'Distance between .* declaration and its first usage is 13.'
+
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+
+        for (int i = 0; i < 10; i++) {
+            a = a + b;
+        }
+    }
+
+     void method2() {
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+        run();
+
+        int a = 3;
+        int b = 2;
+
+        for (int i = 0; i < 10; i++) {
+            a = a + b;  // distance = 1
+        }
+    }
+
+    public void run() {};
+}
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceScopes.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceScopes.java
@@ -269,8 +269,8 @@ public class InputVariableDeclarationUsageDistanceScopes {
     }
 
     public int[] getSelectedIndices() {
-        int[] sel = new int[5];
-        String model = "";
+        int[] sel = new int[5]; // violation 'Distance .* is 2.'
+        String model = ""; // violation 'Distance between .* declaration and its first usage is 2.'
         int a = 0;
         a++;
         for (int index = 0; index < 5; ++index) {
@@ -513,7 +513,7 @@ public class InputVariableDeclarationUsageDistanceScopes {
     public int testIssue32_11(String toDir)
             throws Exception
     {
-        int count = 0;
+        int count = 0; // violation 'Distance .* is 4.'
         String[] files = {}; // violation 'Distance .* is 2.'
 
         System.identityHashCode("Data archival started");
@@ -852,7 +852,7 @@ public class InputVariableDeclarationUsageDistanceScopes {
 
 class New4 {
     void a() {
-        int a = 1;
+        int a = 1; // violation 'Distance .* is 5.'
         System.lineSeparator();
         System.lineSeparator();
         System.lineSeparator();
@@ -864,7 +864,7 @@ class New4 {
     }
 
     void b() {
-        int a = 1;
+        int a = 1; // violation 'Distance .* is 5.'
         System.lineSeparator();
         System.lineSeparator();
         System.lineSeparator();
@@ -876,7 +876,7 @@ class New4 {
     }
 
     void c() {
-        int a = 1;
+        int a = 1; // violation 'Distance .* is 5.'
         System.lineSeparator();
         System.lineSeparator();
         System.lineSeparator();
@@ -888,7 +888,7 @@ class New4 {
     }
 
     void d() {
-        int a = 1;
+        int a = 1; // violation 'Distance .* is 5.'
         System.lineSeparator();
         System.lineSeparator();
         System.lineSeparator();
@@ -900,7 +900,7 @@ class New4 {
     }
 
     void f() {
-        int a = 1;
+        int a = 1; // violation 'Distance .* is 5.'
         System.lineSeparator();
         System.lineSeparator();
         System.lineSeparator();
@@ -910,7 +910,7 @@ class New4 {
     }
 
     void h() {
-        int a = 1;
+        int a = 1; // violation 'Distance .* is 5.'
         System.lineSeparator();
         System.lineSeparator();
         System.lineSeparator();
@@ -921,7 +921,7 @@ class New4 {
     }
 
     void i() {
-        int a = 1;
+        int a = 1; // violation 'Distance .* is 2.'
         switch (Math.max(1, 2)) {
         case 1:
             System.lineSeparator();
@@ -942,7 +942,7 @@ class New4 {
     }
 
     void k() {
-        int a = 1;
+        int a = 1; // violation 'Distance .* is 5.'
         System.lineSeparator();
         System.lineSeparator();
         System.lineSeparator();
@@ -998,7 +998,7 @@ class New4 {
     }
 
     void test() {
-        int a = 0;
+        int a = 0; // violation 'Distance between .* declaration and its first usage is 4.'
 
         System.lineSeparator();
         System.lineSeparator();
--- a/config/pitest-suppressions/pitest-coding-1-suppressions.xml
+++ b/config/pitest-suppressions/pitest-coding-1-suppressions.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressedMutations>
+  <mutation unstable="false">
+    <sourceFile>VariableDeclarationUsageDistanceCheck.java</sourceFile>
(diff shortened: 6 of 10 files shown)


mahfouz72:
Github, generate report

mahfouz72:
this change changes the behavior of the check should I add violation for all places that was good before for example this function in it chapter4 was good but now there is a violation reported 
```
public int testIssue32_11(String toDir)
            throws Exception
    {
        int count = 0;     <--- this line was good before now it has a violation (max distance  is 3 and it is 4 now it eas 0 before)
        String[] files = {};

        System.identityHashCode("Data archival started");
        files.notify();
        System.identityHashCode("sss");

        if (files == null || files.length == 0) {
            System.identityHashCode("No files on a remote site");
        }
        else {
            System.identityHashCode("Files on remote site: " + files.length);

            for (String ftpFile : files) {
                if (files.length == 0) {
                    "".concat("");
                    ftpFile.concat(files[2]);
                    count++;
                }
            }
        }

        System.lineSeparator();

        return count;
    }
```

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

mahfouz72:
I generated two report
- report-1 (no diff) with this [config](https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/e82a7576e6372a18bb4794217b6713b475f0e4f9/check.xml) to check that the check behavior didn't change when `validateBetweenScopes` is on 
- report-2 (expected diff) with this [config](https://gist.githubusercontent.com/mahfouz72/ce878f7db10dd474a7f6aa53ada50c7c/raw/68d7ec7bcef08b49d6deaa712d2e18cf8c328dcd/check1.xml) to check that we raise a violation to move variable near to the block when `validateBetweenScopes` is off as stated in the issue

nrmancuso:
@mahfouz72 when we move onto bug fixes, new features, etc. like this PR (awesome!), please always be the first reviewer. Imagine you have never seen this code before, and call out anything that is not 100% obvious to the reviewer. This helps review to go much faster, and this practice will help you to find issues/improvements in your changes before anyone else. Remember: there is a good chance that some reviewers may have never even seen this part of the code before, or completely forgotten if even if they have :)

You can see an example of this at https://github.com/checkstyle/checkstyle/pull/14503.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -324,38 +324,28 @@ else if (currentAst.getType() != TokenTypes.VARIABLE_DEF) {
...
-                    resultDist++;
-                }
-                else {
-                    // variable usage is in inner scope
-                    // reset counters, because we can't determine distance
-                    resultDist = 0;
-                }
+                // variable usage is in inner scope, treated as 1 block
+                // or in operator expression, then distance + 1
+                resultDist++;
```
mahfouz72, line 345: here we were checking if the variable isn't in operator expression then it is in inner scope so we set the distance to 0 because `validateBetweenScopes` property is false.

my changes were not to set dist to zero but to add one. As we are not interested in the actual distance we just treat the whole inner scope as one block.So, even if the property is off we will get a violation to move the variable near the block. now we don't need this if condition` if(isVariableInOperatorExpr(examineNode, varIdent)) `as in both cases (inner scope and in operator expression) we add 1 to the distance.

the other removals was reported by Pitest and they are just redundant conditions and it is safe to remove like the switch case `LITERAL_SWITCH `is now being handled in the default case same for this condition

```
if (examineNode.getType() == TokenTypes.LABELED_STAT) {
            examineNode = examineNode.getFirstChild().getNextSibling();}
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -708,22 +698,6 @@ private static boolean isVariableInOperatorExpr(
...
-            final DetailAST elseBlock = operator.getLastChild();
-
-            if (elseBlock.getType() == TokenTypes.LITERAL_ELSE) {
-                // Get IF followed by ELSE
-                final DetailAST firstNodeInsideElseBlock = elseBlock.getFirstChild();
-
-                if (firstNodeInsideElseBlock.getType() == TokenTypes.LITERAL_IF) {
-                    isVarInOperatorDeclaration =
-                        isVariableInOperatorExpr(firstNodeInsideElseBlock, variable);
-                }
```
mahfouz72, line 723: these were surviving mutation after my changes as this part of the function was useful when it was used by `getDistToVariableUsageInChildNode`  but after debugging it was redundant when used by `calculateDistanceBetweenScopes` as the work done here to recurse down to check every node (else statments) in If-else block. can be covered  by the method `calculateDistanceBetweenScopes`. as there is a loop in it that checks  If variable usage exists in a single scope, then look into this scope and update the current scope in each iteration so we checking all the else statements in it

 this mutation didn't survive before becuase this recursion with neccessary for `getDistToVariableUsageInChildNode`  because we we don't have such a loop in this method that checks every node (else statements) we were just passing the first If node and let the recursion do the work for us

src/it/java/com/google/checkstyle/test/chapter4formatting/rule4822variabledistance/VariableDeclarationUsageDistanceTest.java
```diff
@@ -43,6 +43,7 @@ public void testArrayTypeStyle() throws Exception {
             "219:9: " + getCheckMessage(clazz, msgExt, "t", 5, 3),
             "483:9: " + getCheckMessage(clazz, msgExt, "myOption", 7, 3),
             "495:9: " + getCheckMessage(clazz, msgExt, "myOption", 6, 3),
+            "508:9: " + getCheckMessage(clazz, msgExt, "count", 4, 3),
         };
```
mahfouz72, line 47: it may be obvious. but this added violation because we are now making the check to do a violation to move the variable near the block our check didn't make this violation before the same was applied for all other input files

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceScopes.java
```diff
@@ -269,8 +269,9 @@ public Object readObject(String in) throws Exception {
     }
 
     public int[] getSelectedIndices() {
+        // violation below, 'Distance between .* declaration and its first usage is 2.'
         int[] sel = new int[5];
-        String model = "";
```
rnveach, line 273: Remove new line and use ` // violation 'Distance .* is 2.'` like you did below.

It best to avoid adding new lines in existing inputs, unless it is specifically part of the issue.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceCloseToBlock.java
```diff
@@ -0,0 +1,37 @@
...
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.variabledeclarationusagedistance;
+
+public class InputVariableDeclarationUsageDistanceCloseToBlock {
+    void method() {
+        int a = 3; // violation 'Distance between .* declaration and its first usage is 13.'
+        int b = 2; // violation 'Distance between .* declaration and its first usage is 13.'
```
rnveach, line 16: Add example in this same test showing what the code looks like with no violations.

rnveach (changes requested):

rnveach:
> I generated two report

If you are going to give this summary for reviewers, I highly recommend you copy the report links too. This helps that we don't have to hunt them down especially when the PR posts get excessively long and GH starts to trim them.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -324,38 +324,28 @@ else if (currentAst.getType() != TokenTypes.VARIABLE_DEF) {
...
-     * @param varIdent variable variable identifier.
      * @param currentDistToVarUsage current distance to the variable usage.
      * @return the distance to variable usage for in the child node.
      */
-    private static int getDistToVariableUsageInChildNode(DetailAST childNode, DetailAST varIdent,
+    private static int getDistToVariableUsageInChildNode(DetailAST childNode,
                                                          int currentDistToVarUsage) {
-        DetailAST examineNode = childNode;
-        if (examineNode.getType() == TokenTypes.LABELED_STAT) {
-            examineNode = examineNode.getFirstChild().getNextSibling();
```
rnveach, line 335: Can you explain why this issue impacts labels specifically? Nothing in the issue mentioned this type.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceCloseToBlock.java
mahfouz72: @rnveach done

rnveach: Not done. The code before and after looks nothing a like. I am expecting a fix of the original code in the issue and what it will look like in the future with no violations as if the developer fixed it himself.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceScopes.java
mahfouz72: @rnveach done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
mahfouz72: @rnveach Yes issue has nothing to do with labels but this condition was a surviving mutation reported by Pitest after my changes. it is just a redundant condition now and is being handled by the default case in the switch. we don't need to get the sibling of its child (loop statement) because the distance will be added by 1 but if we remove this condition completely the label will hit default case in the switch and the distance will be added by 1 also 
So it is safe to remove

mahfouz72:
> If you are going to give this summary for reviewers, I highly recommend you copy the report links too. This helps that we don't have to hunt them down especially when the PR posts get excessively long and GH starts to trim them.

noted. please check the PR description

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceCloseToBlock.java
mahfouz72: @rnveach please check the updated example

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
rnveach: Can you generate a custom diff report? Make `allowedDistance` = 1, before image (base) being the code in this PR as it is now, the after image (patch) being the code in this PR with this label code restored. If no differences are found, then I am good.

rnveach: > it is just a redundant condition now and is being handled by the default case in the switch

The code in the default switch has nothing in common with the label code. One gets the first child and next sibling, the other calculates distance value.

> was a surviving mutation reported by Pitest after my changes

Pitest just reports that there are no tests to prove the reported code is necessary or not, not the worth of keeping the production code. It is up to us to determine by examination if pitest is reporting an actual issue with the production code, or if simply more tests are needed.

mahfouz72: > The code in the default switch has nothing in common

I didn't mean common in code. I mean that after we finish the switch block, we will achieve the same result whether we remove the condition or not

- if we leave the condition we will [get the first child and next sibling](https://github.com/checkstyle/checkstyle/blob/ccde012394294f31d1cc28c51e03c6a622424663/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L335) which is `LITERAL_WHILE` or `LATERAL_FOR` then we will hit [this line](https://github.com/checkstyle/checkstyle/blob/9ab3556eec80d5b2c3a7430d518a6ac6a93dcf72/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L345) where we increment the distance by 1
- if we remove the condition we will hit [this line in the default case](https://github.com/checkstyle/checkstyle/blob/ccde012394294f31d1cc28c51e03c6a622424663/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L359) because `LABELED_STAT.findFirstToken(TokenType.SLIST) == null` so we increment the distance by 1 also 

this was not the case before my changes. the condition was necessary because when we  [get the first child and next sibling](https://github.com/checkstyle/checkstyle/blob/ccde012394294f31d1cc28c51e03c6a622424663/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L335) we enter the switch case and set distance to zero this can't be done in the default 
but now we want to add 1 not set to zero therefore this condition is redundant now

mahfouz72: @rnveach please review if this explanation is good if not I can generate report as you requested above

mahfouz72: @rnveach please review if this explanation is enough if not I can generate report as you requested above

rnveach: Please try generating the report.

mahfouz72: @rnveach I tried so many times and got these errors while generating the report can you please tell me how to solve it? I used to use GitHub actions I haven't used manual before
this is the command that I ran ` D:\contribution\checkstyle-tester> groovy diff.groovy --localGitRepo  D:\checkstyle --baseBranch issue-3502 --patchBranch issue-3502-revertlabel --config my_check.xml --listOfProjects projects-to-test-on.properties`

```
[INFO] BUILD SUCCESS
[INFO] Total time:  01:28 min
[INFO] Finished at: 2024-02-28T02:39:20+02:00
Testing Checkstyle started
   [delete] Deleting directory D:\contribution\checkstyle-tester\src\main\java
Cloning git repository 'checkstyle' to repositories\checkstyle folder ...
Cloning into 'repositories\checkstyle'...
error: RPC failed; curl 92 HTTP/2 stream 0 was not closed cleanly: CANCEL (err 8)
error: 3941 bytes of body are still expected
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
error: RPC failed; curl 92 HTTP/2 stream 0 was not closed cleanly: CANCEL (err 8)
error: 4652 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
Cloning into 'repositories\checkstyle'...
error: 5107 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
Cloning into 'repositories\checkstyle'...
error: 5561 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
Cloning into 'repositories\checkstyle'...
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at diff.generateCheckstyleReport(diff.groovy:271)
        at diff$generateCheckstyleReport$11.callCurrent(Unknown Source)
        at diff.launchCheckstyleReport(diff.groovy:232)
        at diff$launchCheckstyleReport$7.callCurrent(Unknown Source)
        at diff.run(diff.groovy:31)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
```

rnveach: @mahfouz72 Restore the code and add a suppression to the pitest file in this PR. I talked to another admin and we add a suppression for now.

Afterwards, can you report your issue to https://github.com/checkstyle/contribution/issues/ .

nrmancuso: I am good to extend suppression list in a case like this, we can do a follow up to kill this mutation to make life easier for both @mahfouz72 and reviewers, and get the fix to users.

mahfouz72: @rnveach @nrmancuso I extended suppression and reported[ issue](https://github.com/checkstyle/contribution/issues/839) for this. is it an individual issue? also, I am wondering if I can use GitHub actions in case that the base isn't  master branch like in this case

rnveach: I don't use GH actions, so I can't say for sure, but I am pretty sure GH actions only works with master and the code in the PR. My ask was you use a separate branch of code, so to use that with GH actions you would need another PR or to crash this PR's state for the run.

rnveach (approved):

mahfouz72:
@nrmancuso are we ok on this?

mahfouz72:
@nrmancuso @rnveach 
finally [report](https://mahfouz72.github.io/) with this PR as base and patch is the same but with hardcoding the mutations no diff found on 26 projects 
so now we will merge this extended suppresion list or just proceed and remove the mutated code

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -320,6 +319,7 @@ private static String getFullyQualifiedClassName(String filePath, String moduleN
                 "com.puppycrawl.tools.checkstyle.checks.javadoc.SummaryJavadocCheck");
         moduleMappings.put("LineLength",
                 "com.puppycrawl.tools.checkstyle.checks.sizes.LineLengthCheck");
+        String fullyQualifiedClassName;
```
nrmancuso, line 322: Love it when we find new violations within the project :)

nrmancuso (approved):
Thanks a lot for this! Please send follow up PRs to address the new suppressed mutations.

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
mahfouz72: same here 😄

Issue #14436, opened 2024-02-06
Rename Java Input files and test methods to specify Test functionality


MANISH-K-07:
Response to discussion at  https://github.com/checkstyle/checkstyle/pull/14369#discussion_r1469873309

Findings from PR #14369 : 

[Input file](https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableOnInnerClassFields.java) -
```
public class InputJavadocVariableNoJavadocOne          //naming like '1' or 'One' need to be omitted
public class InputJavadocVariableOnInnerClassFields    //preferred based on test functionality
```
[Test Method](https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheckTest.java) - 

```
@Test
public void testScopesOne() {                              //naming like '1' or 'One' need to be omitted
public void testJavadocVariableOnInnerClassFields() {      //preferred based on test functionality
```

Nonsense test method/input file naming conventions like this need to be avoided. We must come up with some descriptive name to differentiate these tests, considering the content and their functionality. Naming conventions like "`........One`" in both test method names and test inputs are being followed at present which tell us nothing about the purpose of a given test. Imagine that it is required to come back to this test class in near future. Why does this test exist? What did the author want to validate? 

**_A precise naming convention needs to be enforced to give a better idea of what we are testing, without even needing to open up the input file and inspect the configuration and code. Precise naming as mentioned above is an outcome of clear understanding of the check and what it tests, it's functionality or purpose._** 

Fixing all the Java Input files in checkstyle is possible only through active contributions. So, first of all, a huge Thanks to every new contributor. If you are looking forward to working on this issue, **_be sure to analyze the test and look at the example PR :_** 

`Examples of ideal updates` : 

- https://github.com/checkstyle/checkstyle/commit/6d4fe4f19264b5e4f47fe3cbdc3015f61c7a8c58
- https://github.com/checkstyle/checkstyle/commit/b08ea45624d10cd6a3eac7da8c902de3dbb98f8c
- https://github.com/checkstyle/checkstyle/commit/8673001d9b0edacfa0d54bec85b26db46d1cce4d

`Each PR must contain one complete module with all existing filenames and test method names analysed and renamed.`
[...]

PR #14549, opened 2024-02-25
Issue #13213: Remove '//ok' comments from InputUnnecessaryParenthesesIfStatement2

CubeTures:
Issue #13213 
I will get the suppressions too.

PR #14550, opened 2024-02-26
Issue #14436: Enforce Naming Convention for MissingCtor

CubeTures:
Issue #14436 
InputMissingCtor2.java -> InputMissingCtorNestedClasses.java
Still need to change MissingCtorCheckTest.java.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingCtorCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingCtorCheckTest.java
@@ -69,7 +69,7 @@ public class MissingCtorCheckTest extends AbstractModuleTestSupport {
         };
 
         verifyWithInlineConfigParser(
-                getPath("InputMissingCtor2.java"),
+                getPath("InputMissingCtorNestedClasses.java"),
                 expected);
     }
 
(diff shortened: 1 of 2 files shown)


CubeTures:
Will squash and fix commit titles as per rules.

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingCtorCheckTest.java
```diff
@@ -69,7 +69,7 @@ public void testMissingCtorClassOnOneLine() throws Exception {
         };
 
         verifyWithInlineConfigParser(
-                getPath("InputMissingCtor2.java"),
+                getPath("InputMissingCtorNestedClasses.java"),
```
MANISH-K-07, line 72: @CubeTures ,
The issue requires renaming of test method names along with the java input files.
Please do change test name above : 

testMissingCtorClassOnOneLine() --> `testMissingCtorNestedClasses`

MANISH-K-07 (commented):
Please do reuse the existing PR.
Look at my comment at https://github.com/checkstyle/checkstyle/pull/14549#pullrequestreview-1899861230

It is recommended to update a whole module as part of the issue instead of single file to avoid ambiguity.

Make changes to input file [InputMissingCtor.java](https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/missingctor/InputMissingCtor.java) and the respective test method also.
I would recommend something like : 
InputMissingCtor --> `InputMissingCtorCheckPublicClasses`

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingCtorCheckTest.java
CubeTures: Thanks. I'm new to using git so I had actually made two commits instead of just the one. In the second commit I had renamed the test method name as well. Thanks for your help and I'll be sure to find a new name for the original InputMissingCtor.java as well!

PR #14553, opened 2024-02-26
Issue #11340: Add ignoreAnnotatedBy property to ParameterNumberCheck

### LINKED ISSUE #11340: Add ignoreMethodAnnotatedBy property to ParameterNumberCheck
Check name: [ParameterNumberCheck](https://checkstyle.org/config_sizes.html#ParameterNumber)

https://checkstyle.org/checks/sizes/parameternumber.html#ParameterNumber

## Problem

https://www.baeldung.com/jackson-annotations#1-jsoncreator
https://fasterxml.github.io/jackson-annotations/javadoc/2.7/com/fasterxml/jackson/annotation/JsonCreator.html
If I implement external API, I have no power over the number of parameters, e.g. when creating a DTO class for REST API and deserializing JSON to an object by constructor annotated by @JsonCreator

```
class MyDto {
  private final String a1;
  private final String a2;
  private final String a3;
  private final String a4;
  private final String a5;
  private final String a6;
  private final String a7;
  private final String a8;
  private final String a9;

  @JsonCreator
  MyDto(String a1, String a2, String a3, String a4, String a5, String a6, String a7, String a8, String a9) {
    this.a1 = a1;
    this.a2 = a2;
    this.a3 = a3;
    this.a4 = a4;
    this.a5 = a5;
    this.a6 = a6;
    this.a7 = a7;
    this.a8 = a8;
    this.a9 = a9;
  }
}
```

## Solution

I'd like to configure check to ignore this constructor. In general, ignore all methods annotated by `@JsonCreator`, e.g.

```
 <module name="ParameterNumber">
   <property name="max" value="5"/>
   <property name="ignoreMethodAnnotatedBy" value="JsonCreator"/>
 </module>
```

The PR with the proposal is here https://github.com/checkstyle/checkstyle/pull/11339

Check already has property `ignoreOverriddenMethods`  so new property will make it more general.

sktpy:
Resolves #11340 

Old Doc([site](https://checkstyle.org/checks/sizes/parameternumber.html)), New Doc([site](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/4cc9297_2024153913/checks/sizes/parameternumber.html#Properties))
<hr>

### Regression: ([Report](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a62d9ca_2024194155/reports/diff/index.html))

**Differences Found** (expected):
`Deprecated` is used as the value for the `ignoreAnnotatedBy`(selected for the purpose that it would produce some differences being a commonly used annotation)

**Removed violations** (**22**): Example ([link](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a62d9ca_2024194155/reports/diff/guava/xref/home/runner/work/checkstyle/checkstyle/.ci-temp/contribution/checkstyle-tester/repositories/guava/android/guava/src/com/google/common/collect/ImmutableSortedMapFauxverideShim.java.html#L107)), rest of the occurrences are similar.
```java
@Deprecated
public static <K, V> ImmutableSortedMap<K, V> of(K k1, V v1, K k2, V v2, K k3, V v3, K k4, V v4) {
```
  -   Method above contains 8 params, while maximum(by default) allowed are 7, 
  -   This is no longer a violation because it is annotated by `Deprecated` which is specified in `ignoreAnnotatedBy` property in the patch-config. 
  -   This is the intended behaviour.
<hr>

Diff Regression config: https://gist.githubusercontent.com/sktpy/0c6a38735b3206b86b593a490206e76d/raw/4c033289ef4802e25c473cec5ef58ff49ea8d788/base-config.xml
Diff Regression patch config: https://gist.githubusercontent.com/sktpy/43c4a398aadd414fbe69d75953704597/raw/7f468b2a7e04762d763d7148d81a640dad733091/patch-config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
@@ -19,6 +19,9 @@
 
 package com.puppycrawl.tools.checkstyle.checks.sizes;
 
+import java.util.Collections;
+import java.util.Set;
+
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
@@ -31,6 +34,12 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
  * Checks the number of parameters of a method or constructor.
  * </p>
  * <ul>
+ * <li>
+ * Property {@code ignoreAnnotatedBy} - Ignore methods and constructors
+ * annotated with the specified annotation(s).
+ * Type is {@code java.lang.String[]}.
+ * Default value is {@code ""}.
+ * </li>
  * <li>
  * Property {@code ignoreOverriddenMethods} - Ignore number of parameters for
  * methods with {@code @Override} annotation.
@@ -87,6 +96,11 @@ public class ParameterNumberCheck
     private boolean ignoreOverriddenMethods;
 
     /**
+     * Ignore methods and constructors annotated with the specified annotation(s).
+     */
+    private Set<String> ignoreAnnotatedBy = Collections.emptySet();
+
+    /**
      * Setter to specify the maximum number of parameters allowed.
      *
      * @param max the max allowed parameters
@@ -104,6 +118,16 @@ public class ParameterNumberCheck
      */
     public void setIgnoreOverriddenMethods(boolean ignoreOverriddenMethods) {
         this.ignoreOverriddenMethods = ignoreOverriddenMethods;
+    }
+
+    /**
+     * Setter to ignore methods and constructors annotated with the specified annotation(s).
+     *
+     * @param annotationNames specified annotation(s)
+     * @since 10.15.0
+     */
+    public void setIgnoreAnnotatedBy(String... annotationNames) {
+        ignoreAnnotatedBy = Set.of(annotationNames);
     }
 
     @Override
@@ -132,16 +156,33 @@ public class ParameterNumberCheck
     }
 
     /**
-     * Determine whether to ignore number of parameters for the method.
+     * Determine whether to ignore number of parameters.
      *
      * @param ast the token to process
-     * @return true if this is overridden method and number of parameters should be ignored
-     *         false otherwise
+     * @return true if number of parameters should be ignored.
      */
     private boolean shouldIgnoreNumberOfParameters(DetailAST ast) {
-        // if you override a method, you have no power over the number of parameters
-        return ignoreOverriddenMethods
-                && AnnotationUtil.hasOverrideAnnotation(ast);
+        return isIgnoredOverriddenMethod(ast) || isAnnotatedByIgnoredAnnotations(ast);
+    }
+
+    /**
+     * Checks if method is overridden and should be ignored.
+     *
+     * @param ast method definition to check
+     * @return true if method is overridden and should be ignored.
+     */
+    private boolean isIgnoredOverriddenMethod(DetailAST ast) {
+        return ignoreOverriddenMethods && AnnotationUtil.hasOverrideAnnotation(ast);
+    }
+
+    /**
+     * Checks if method or constructor is annotated by ignored annotation(s).
+     *
+     * @param ast method or constructor definition to check
+     * @return true if annotated by ignored annotation(s).
+     */
+    private boolean isAnnotatedByIgnoredAnnotations(DetailAST ast) {
+        return AnnotationUtil.containsAnnotation(ast, ignoreAnnotatedBy);
     }
 
 }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheckTest.java
@@ -123,4 +123,35 @@ public class ParameterNumberCheckTest
                 getPath("InputParameterNumber2.java"), expected);
     }
 
+    @Test
+    public void testIgnoreAnnotatedBy() throws Exception {
+        final String[] expected = {
+            "23:10: " + getCheckMessage(MSG_KEY, 2, 3),
+            "30:10: " + getCheckMessage(MSG_KEY, 2, 4),
+            "35:14: " + getCheckMessage(MSG_KEY, 2, 3),
+            "43:9: " + getCheckMessage(MSG_KEY, 2, 4),
+            "58:30: " + getCheckMessage(MSG_KEY, 2, 3),
+            "62:29: " + getCheckMessage(MSG_KEY, 2, 3),
+            "77:34: " + getCheckMessage(MSG_KEY, 2, 4),
+            "97:10: " + getCheckMessage(MSG_KEY, 2, 3),
+            "106:14: " + getCheckMessage(MSG_KEY, 2, 3),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputParameterNumberIgnoreAnnotatedBy.java"), expected);
+    }
+
+    @Test
+    public void testIgnoreAnnotatedByFullyQualifiedClassName() throws Exception {
+        final String[] expected = {
+            "15:10: " + getCheckMessage(MSG_KEY, 2, 3),
+            "17:10: " + getCheckMessage(MSG_KEY, 2, 3),
+            "23:10: " + getCheckMessage(MSG_KEY, 2, 3),
+            "27:10: " + getCheckMessage(MSG_KEY, 2, 3),
+            "41:14: " + getCheckMessage(MSG_KEY, 2, 3),
+            "45:14: " + getCheckMessage(MSG_KEY, 2, 3),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.java"),
+                expected);
+    }
 }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.java
@@ -1,0 +1,62 @@
+/*
+ParameterNumber
+max = 2
+ignoreAnnotatedBy = java.lang.Deprecated, InnerClass.InnerAnno, Session
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.sizes.parameternumber;
+
+public class InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName {
+    @java.lang.Deprecated
+    void method3(int a, int b, int c) {}
+
+    @Deprecated
+    void method4(int a, int b, int c) {} // violation, 'More than 2 parameters (found 3).'
+
+    void method5(int a, int b, int c) {} // violation, 'More than 2 parameters (found 3).'
+
+    @Session
+    void method6(int a, int b, int c) {}
+
+    @InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.Session
+    void method7(int a, int b, int c) {} // violation, 'More than 2 parameters (found 3).'
+
+    @com.puppycrawl.tools.checkstyle.checks.sizes.parameternumber
+            .InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.Session
+    void method8(int a, int b, int c) {} // violation, 'More than 2 parameters (found 3).'
+
+    @com.puppycrawl.tools.checkstyle.checks.sizes.parameternumber
+            .InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.Session
+    void method8(int a, int b) {}
+
+    private static class InnerClass {
+        private @interface InnerAnno {}
+
+        @InnerClass.InnerAnno
+        void method1(int a, int b, int c) {
+        }
+
+        @InnerAnno
+        void method2(int a, int b, int c) { // violation, 'More than 2 parameters (found 3).'
+        }
+
+        @Bit
+        void method3(int a, int b, int c) { // violation, 'More than 2 parameters (found 3).'
+        }
+
+        @InnerAnno
+        void method2(int a, int b) {
+        }
+
+        @Bit
+        void method3(int a, int b) {
+        }
+
+        void method4(int a, int b) {
+        }
+    }
+
+    @interface Session {}
+    @interface Bit{}
+}
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/Example4.txt
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/Example4.txt
@@ -1,0 +1,31 @@
+/*xml
+<module name="Checker">
+  <module name="TreeWalker">
+    <module name="ParameterNumber">
+      <property name="ignoreAnnotatedBy" value="JsonCreator"/>
+    </module>
+  </module>
+</module>
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.sizes.parameternumber;
+
+// xdoc section -- start
+class Example {
+    @JsonCreator
+    Example(int a, int b, int c, int d,
+            int e, int f, int g, int h) {}
+
+    // violation below, 'More than 7 parameters (found 8)'
+    Example(String a, String b, String c, String d,
+            String e, String f, String g, String h) {}
+
+    // annotation name must be an exact match
+    // violation below, 'More than 7 parameters (found 8)'
+    @com.puppycrawl.tools.checkstyle.checks.sizes.parameternumber.Example.JsonCreator
+    Example(float a, float b, float c, float d,
+            float e, float f, float g, float h) {}
+
+    @interface JsonCreator {}
+}
+// xdoc section -- end
--- a/src/xdocs/checks/sizes/parameternumber.xml
+++ b/src/xdocs/checks/sizes/parameternumber.xml
@@ -25,6 +25,13 @@
               <th>since</th>
             </tr>
             <tr>
+              <td>ignoreAnnotatedBy</td>
+              <td>Ignore methods and constructors annotated with the specified annotation(s).</td>
+              <td><a href="../../property_types.html#String.5B.5D">String[]</a></td>
+              <td><code>{}</code></td>
+              <td>10.15.0</td>
+            </tr>
+            <tr>
               <td>ignoreOverriddenMethods</td>
               <td>Ignore number of parameters for methods with <code>@Override</code> annotation.</td>
               <td><a href="../../property_types.html#boolean">boolean</a></td>
@@ -114,6 +121,43 @@ public void needsLotsOfParameters(int a,
     ...
 }
         </source>
+        <p id="Example4-config">
+          To configure the check to ignore methods and constructors annotated with
+          <code>JsonCreator</code> annotation:
+        </p>
+        <source>
+&lt;module name=&quot;Checker&quot;&gt;
+  &lt;module name=&quot;TreeWalker&quot;&gt;
+    &lt;module name=&quot;ParameterNumber&quot;&gt;
+      &lt;property name=&quot;ignoreAnnotatedBy&quot; value=&quot;JsonCreator&quot;/&gt;
+    &lt;/module&gt;
+  &lt;/module&gt;
+&lt;/module&gt;
+        </source>
+        <p id="Example4-code">Example:</p>
+        <source>
+class Example {
+    @JsonCreator
+    Example(int a, int b, int c, int d,
+            int e, int f, int g, int h) {}
+
+    // violation below, 'More than 7 parameters (found 8)'
+    Example(String a, String b, String c, String d,
+            String e, String f, String g, String h) {}
+
+    // annotation name must be an exact match
+    // violation below, 'More than 7 parameters (found 8)'
+    @com.puppycrawl.tools.checkstyle.checks.sizes.parameternumber.Example.JsonCreator
+    Example(float a, float b, float c, float d,
+            float e, float f, float g, float h) {}
+
+    @interface JsonCreator {}
+}
+        </source>
+        <p>
+          Note: Annotation names specified in the <code>ignoreAnnotatedBy</code> property
+          must be an exact match for the annotation name on the method or constructor.
+        </p>
       </subsection>
 
       <subsection name="Example of Usage" id="Example_of_Usage">
--- a/src/xdocs/checks/sizes/parameternumber.xml.template
+++ b/src/xdocs/checks/sizes/parameternumber.xml.template
@@ -63,6 +63,25 @@
                  value="resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/Example3.txt"/>
           <param name="type" value="code"/>
         </macro>
+        <p id="Example4-config">
+          To configure the check to ignore methods and constructors annotated with
+          <code>JsonCreator</code> annotation:
+        </p>
+        <macro name="example">
+          <param name="path"
+                 value="resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/Example4.txt"/>
+          <param name="type" value="config"/>
+        </macro>
+        <p id="Example4-code">Example:</p>
+        <macro name="example">
+          <param name="path"
+                 value="resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/Example4.txt"/>
+          <param name="type" value="code"/>
+        </macro>
+        <p>
+          Note: Annotation names specified in the <code>ignoreAnnotatedBy</code> property
+          must be an exact match for the annotation name on the method or constructor.
+        </p>
       </subsection>
 
       <subsection name="Example of Usage" id="Example_of_Usage">
--- a/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionParameterNumberTest.java
+++ b/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionParameterNumberTest.java
@@ -103,4 +103,30 @@ public class XpathRegressionParameterNumberTest extends AbstractXpathTestSupport
         runVerifications(moduleConfig, fileToProcess, expectedViolations, expectedXpathQueries);
     }
 
+    @Test
+    public void testIgnoreAnnotatedBy() throws Exception {
+        final String filePath =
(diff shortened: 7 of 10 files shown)


sktpy:
GitHub, generate report

sktpy:
GitHub, generate website

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/Example4.txt
```diff
@@ -0,0 +1,23 @@
...
+    <module name="ParameterNumber">
+      <property name="ignoreAnnotatedBy" value="Session"/>
+    </module>
+  </module>
+</module>
+*/
+
+// xdoc section -- start
+@Session
+public void ignoredAnnotatedMethod(int a, int b, int c, int d,
```
sktpy, line 13: Have kept the xdoc examples for ignoreAnnotatedBy in the non-enabled way to not grow this PR more than required, as rest of examples are currently non-enabled and it wouldn't make much sense to have just one example enabled while the rest remain the old way.
All the examples for this check can be enabled at once in a separate PR as part of the ongoing issue to Enable the Examples.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
```diff
@@ -132,16 +157,33 @@ public void visitToken(DetailAST ast) {
...
+     * Determine whether to ignore number of parameters.
      *
-     * @param ast the token to process
-     * @return true if this is overridden method and number of parameters should be ignored
-     *         false otherwise
+     * @param ast ast
+     * @return true if number of parameters should be ignored.
      */
     private boolean shouldIgnoreNumberOfParameters(DetailAST ast) {
-        // if you override a method, you have no power over the number of parameters
```
sktpy, line 142: Have removed this comment, because it is more from a philosophical point of view to why the functionality to ignore overridden method was added but does not imparts any helpful technical info.

sktpy:
Failing checks in CI([cirrus](https://github.com/checkstyle/checkstyle/pull/14553/checks?check_run_id=21998086645), [xwiki](https://app.circleci.com/pipelines/github/checkstyle/checkstyle/24150/workflows/20ec76ce-7477-45f8-a4d6-2cdef9a76e9c/jobs/529429/parallel-runs/0/steps/0-103?invite=true#step-103-835196_151)) are unrelated.

sktpy:
GitHub, generate website

nrmancuso (changes requested):
Very good, few items:

src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheckTest.java
```diff
@@ -123,4 +123,20 @@ public void testIgnoreOverriddenMethodsFalse()
                 getPath("InputParameterNumber2.java"), expected);
     }
 
+    @Test
+    public void testIgnoreAnnotatedBy() throws Exception {
```
nrmancuso, line 127: Please add another test, showing that we can support fully qualified names on annotations, like `com.puppycrawl.tools.checkstyle.checks.sizes.parameternumber.MyAnno`.

nrmancuso (changes requested):
One more:

src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheckTest.java
sktpy: Should this be the case when the fully qualified name is specified in the property, or it should be able to derive it from just `MyAnno` (if this is the case, it isn't supported yet)

nrmancuso: >Should this be the case when the fully qualified name is specified in the property

Yes, thanks for asking.

>or it should be able to derive it from just MyAnno

No, we should try to keep this check as simple as possible, this would probably involve looking at the package, imports, etc.

nrmancuso: Also, it would be ideal to add an example with a fully qualified name to the docs to make it obvious that we need to qualify the name in the property to have a match.

sktpy:
GitHub, generate website

sktpy:
GitHub, generate website

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
```diff
@@ -32,6 +34,12 @@
  * </p>
  * <ul>
  * <li>
+ * Property {@code ignoreAnnotatedBy} - Ignore number of parameters for
+ * methods and constructors annotated with the specified annotation(s).
```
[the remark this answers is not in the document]
sktpy: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
```diff
@@ -106,6 +120,17 @@ public void setIgnoreOverriddenMethods(boolean ignoreOverriddenMethods) {
...
     }
 
+    /**
+     * Setter to ignore number of parameters for
+     * methods and constructors annotated with the specified annotation(s).
+     *
+     * @param userAnnotations user's value.
+     * @since 10.14.0
+     */
+    public void setIgnoreAnnotatedBy(String... userAnnotations) {
```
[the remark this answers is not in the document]
sktpy: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
```diff
@@ -132,16 +157,33 @@ public void visitToken(DetailAST ast) {
...
+     * @return true if ignoreOverriddenMethods property is true and the method is overridden.
+     */
+    private boolean isIgnoredOverriddenMethod(DetailAST ast) {
+        return ignoreOverriddenMethods && AnnotationUtil.hasOverrideAnnotation(ast);
+    }
+
+    /**
+     * Checks if the constructor or method is annotated by ignored annotation(s).
+     *
+     * @param ast ast
```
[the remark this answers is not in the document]
sktpy: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
```diff
@@ -132,16 +157,33 @@ public void visitToken(DetailAST ast) {
     }
 
     /**
-     * Determine whether to ignore number of parameters for the method.
+     * Determine whether to ignore number of parameters.
      *
-     * @param ast the token to process
-     * @return true if this is overridden method and number of parameters should be ignored
-     *         false otherwise
+     * @param ast ast
```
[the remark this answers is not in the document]
sktpy: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
```diff
@@ -132,16 +157,33 @@ public void visitToken(DetailAST ast) {
...
      */
     private boolean shouldIgnoreNumberOfParameters(DetailAST ast) {
-        // if you override a method, you have no power over the number of parameters
-        return ignoreOverriddenMethods
-                && AnnotationUtil.hasOverrideAnnotation(ast);
+        return isIgnoredOverriddenMethod(ast) || isAnnotatedByIgnoredAnnotations(ast);
+    }
+
+    /**
+     * Checks if ignoreOverriddenMethods property is true and the method is overridden.
```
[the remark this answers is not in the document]
sktpy: Done.
Was thinking the same thing while writing that, but for some reason carried on, might as well have written the exact code instead :joy:

src/test/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheckTest.java
sktpy: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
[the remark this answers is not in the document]
sktpy: Had written in that way to be consistent with the description of the `ignoreOverriddenMethods` property of the check, now that we have changed it, should I do the same for `ignoreOverriddenMethods` perhaps in a minor:... pr after this one gets merged?

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/Example4.txt
```diff
@@ -0,0 +1,19 @@
...
+  </module>
+</module>
+*/
+
+// xdoc section -- start
+class Example {
+    @JsonCreator
+    Example(int a, int b, int c, int d,
+            int e, int f, int g, int h) {}
+
```
nrmancuso, line 16:
```suggestion
    @JsonCreator
    Example(int a, int b, int c, int d,
            int e, int f, int g, int h) {}
     
     // violation below, 'More than 7 parameters (found 8)'
     Example(String a, String b, String c, String d,
            String e, String f, String g, String h) {}
            
     // annotation name must be an exact match
     // violation below, 'More than 7 parameters (found 8)'
     @com.puppycrawl.tools.checkstyle.checks.sizes.ParameterNumber.JsonCreator
     Example(float a, float b, float c, float d,
            float e, float f, float g, float h) {}
            

```

src/xdocs/checks/sizes/parameternumber.xml
```diff
@@ -112,6 +119,60 @@
...
+
+    @interface JsonCreator {}
+}
+        </source>
+        <p>
+          Fully qualified class name(s) can also specified in ignoreAnnotatedBy property.
+          Note that when specifying annotation(s), only methods and constructors annotated with
+          the exact specified annotation(s) will be ignored, any variation, even if it refers to
+          the same annotation, will result in it being checked, and produce violation if any.
+        </p>
```
nrmancuso, line 152: Let's be as concise as possible:
```suggestion
        <p>
          Note: annotation names specified in the <code>ignoreAnnotatedBy</code> property must be an exact match for the annotation name on the method or constructor.
        </p>
```

This statement + the extension to the example I've provided above should sufficiently convey this concept.

src/xdocs/checks/sizes/parameternumber.xml
```diff
@@ -112,6 +119,60 @@
...
+        <p id="Example5-config">
```
nrmancuso, line 153: Please completely remove this example, for two reasons:
1. With my proposed example above, we accurately demonstrate the check's behavior in this respect
2. I can't think of any practical reason that anyone would want to ignore the number of parameters in a method definition based on the method is `Deprecated` or not. It is OK (even desireable) to use weird code in tests, but in user facing documentation, all examples should have reasonable usage that makes sense and is familiar to developers.

@sktpy what are your thoughts on this?

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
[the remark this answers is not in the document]
nrmancuso: > Had written it that way to be consistent with the description of `ignoreOverriddenMethods` property of the check, now that we have changed it, should I do the same for `ignoreOverriddenMethods` perhaps in a minor:... pr after this one gets merged?

@sktpy please restate this, I do not understand this question.

src/xdocs/checks/sizes/parameternumber.xml
nrmancuso: Also, apologies if we had a misunderstanding at https://github.com/checkstyle/checkstyle/pull/14553#discussion_r1505368278, my intention was to show the check's behavior on some code like this, not to make an entirely new example.

sktpy: I agree, only reason I went for `Deprecated` was because its FQCN was concise and I thought being a widely used annotation, it would have delivered what we were trying to convey easily.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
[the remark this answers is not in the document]
sktpy: The description of `ignoreOverriddenMethods` is as following:
https://github.com/checkstyle/checkstyle/blob/7e6b50ff6983d37e2b7235fc1be3adb7a525dd1e/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java#L86

This was the reason I had kept the description of `ignoreAnnotatedBy` in the similiar way, but since now it is changed as suggested to not have `...number of parameters for...`, My question was that should the description of `ignoreOverriddenMethods` be also changed in a similar way for the purpose of consistency? (after this is merged, in a minor pr)

sktpy:
GitHub, generate website

src/xdocs/checks/sizes/parameternumber.xml
sktpy: Done.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/Example4.txt
sktpy: Done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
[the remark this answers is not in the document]
nrmancuso: Ah! I understand now. Thanks for the explanation. Yes, we could send another PR with `doc:` prefix for this, no need to open an issue for a simple documentation update.

sktpy:
Failing CI is unrelated, build is failing on master for line length violation in the release note.

nrmancuso:
I am in agreement with sentiments at https://github.com/checkstyle/checkstyle/pull/14553#discussion_r1503208990, namely we can focus on making examples all the "same" when we migrate to actual example test execution. We don't need to make this PR any larger than it is right now.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
```diff
@@ -106,6 +119,16 @@ public void setIgnoreOverriddenMethods(boolean ignoreOverriddenMethods) {
         this.ignoreOverriddenMethods = ignoreOverriddenMethods;
     }
 
+    /**
+     * Setter to ignore methods and constructors annotated with the specified annotation(s).
+     *
+     * @param annotationNames specified annotation(s)
+     * @since 10.14.0
```
romani, line 126: 10.15.0

src/test/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.java
```diff
@@ -0,0 +1,62 @@
...
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.sizes.parameternumber;
+
+public class InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName {
+    @java.lang.Deprecated
+    void method3(int a, int b, int c) {}
+
+    @Deprecated
+    void method4(int a, int b, int c) {} // violation, 'More than 2 parameters (found 3).'
```
romani, line 15: This needs to be discussed.
When use define `java.lang.Deprecated` he usually does not mean to use  match fully and exactly.

It can be instruction to  figure out exactly a type not a string values. We already have Checks that does this. By import we can be 100% sure on type.

romani (changes requested):
Items

src/test/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.java
romani: @nrmancuso , @rnveach , what are your thoughts on this?

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
sktpy: Done.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.java
sktpy: This might be helpful here: https://github.com/checkstyle/checkstyle/pull/14553#discussion_r1505333655

nrmancuso: I think we should go with most simple implementation first (what's in this PR), we can iterate on this as needed. Worst case is that some user has to put both the fully qualified name and the simple annotation name in the config.

sktpy: Do we have an update here?

nrmancuso: I am good to proceed as-is, @rnveach please share your thoughts

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
```diff
@@ -86,6 +94,11 @@ public class ParameterNumberCheck
     /** Ignore number of parameters for methods with {@code @Override} annotation. */
     private boolean ignoreOverriddenMethods;
 
+    /**
+     * Ignore methods and constructors annotated with the specified annotation(s).
+     */
+    private Set<String> ignoreAnnotatedBy = Set.of();
```
rnveach, line 100: @nrmancuso @romani We have `ignoreOverriddenMethods` which also ignores methods based on annotation of `Override`. Is there some reason we aren't combining these 2 to make the configuration for this check more reasonable?

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
```diff
@@ -32,6 +34,12 @@
  * </p>
  * <ul>
  * <li>
+ * Property {@code ignoreAnnotatedBy} - Ignore methods and constructors
+ * annotated with the specified annotation(s).
```
rnveach, line 38: > Ignore ... constructors annotated with the specified annotation(s).

Where is the test that shows this?

src/xdocs/checks/sizes/parameternumber.xml.template
```diff
@@ -63,6 +63,25 @@
...
+        <p id="Example4-config">
+          To configure the check to ignore methods and constructors annotated with
+          <code>JsonCreator</code> annotation:
+        </p>
+        <macro name="example">
+          <param name="path"
+                 value="resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/Example4.txt"/>
+          <param name="type" value="config"/>
+        </macro>
+        <p id="Example4-code"> Example:</p>
```
rnveach, line 75: Remove space before `Example:`

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
sktpy: I believe you have missed:
https://github.com/checkstyle/checkstyle/pull/14553/files#diff-22ceb90d0660b452426c9a948115933dd3e1afd09a4473174d6179ced132df60
and
https://github.com/checkstyle/checkstyle/pull/14553/files#diff-297d5bdb06dede8c554b1542df9781b8833099a36eb71ae102368195f8b6dcd8

Are they not enough/missing something?

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
sktpy: I was thinking the same, but then I read:
https://github.com/checkstyle/checkstyle/issues/11340#issuecomment-1052311653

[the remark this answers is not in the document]
sktpy: Done.
I was thinking the same before, but then saw that we would be needing additional import statement for Collections,
but `emptySet()` is indeed a bit more readable, thanks for pointing it out,
will keep this in mind for future to not be afraid of additional import statements.

src/xdocs/checks/sizes/parameternumber.xml.template
sktpy: Done.
Hawk eye :eagle:

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
nrmancuso: We could deprecate this property, and remove it it the next major version bump.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
rnveach: I see them now. I recommend when you submit links, you also include the line numbers so I know exactly where you are saying the cases are. Pointing to the whole file doesn't help, when I can already see the file but not the specific cases.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.java
rnveach: I think we should be consistent with how we treat annotation properties like this with other checks. I don't find it friendly to have to specify both class name and FQCN.
https://checkstyle.org/checks/coding/illegaltype.html#IllegalType

rnveach: We need to make our behavior consistent with other checks and friendly to the user.

We have some check somewhere that requires FQCN and will be smart and reduce it to simple based on the import in the class.
https://checkstyle.org/checks/coding/illegaltype.html
https://checkstyle.org/checks/coding/illegalinstantiation.html

I also feel this shouldn't be delayed if it is going to create almost a breaking change in the fix.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
sktpy: I didn't included the line numbers because everything in those two files are the cases concerning these,
if I pointed to the specific line numbers they wouldn't be too helpful without going through the config at the top,
at that point I would have created a whole file in the comment itself.

sktpy: But if you still believe that including the line number in this case can somehow be helpful,
please do tell me so that I can do it for future cases.

rnveach: I wouldn't have mentioned it if I didn't think it would be helpful. This item is resolved.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
sktpy: Do we have a resolution here?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.java
sktpy: Do we have a resolution here?

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.java
nrmancuso: We can leave both for now, let’s create an issue to discuss deprecation of ignoreOverriddenMethods

sktpy: https://github.com/checkstyle/checkstyle/issues/14663

nrmancuso: @rnveach please continue review

src/test/resources/com/puppycrawl/tools/checkstyle/checks/sizes/parameternumber/InputParameterNumberIgnoreAnnotatedByFullyQualifiedClassName.java
romani: I agree we can extend to FQCN later on. Lets have basic implementation for now.
With current traffic of PRs I do not have time to mediate on consistency.

rnveach (approved):

rnveach:
https://github.com/sktpy/checkstyle/tree/issue-11340-ignore-annotated-by
This branch is [1 commit ahead of](https://github.com/sktpy/checkstyle/compare/checkstyle%3Acheckstyle%3Amaster...issue-11340-ignore-annotated-by), [52 commits behind](https://github.com/sktpy/checkstyle/compare/issue-11340-ignore-annotated-by...checkstyle%3Acheckstyle%3Amaster)

sktpy:
Rebased over latest master.

PR #14576, opened 2024-03-01
Issue #6207: Add XPath IT Regression test for SuperFinalize

CubeTures:
Issue #6207 

Not complete; I need some help with completing this push.
I'm getting the following errors when running the test. How can I fix this? It seems like I'm putting in the wrong required parameters for the expected value, but I think I'm doing the correct process. Any guidance would be very much appreciated. 
image

### DIFF
--- a/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionSuperFinalizeTest.java
+++ b/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionSuperFinalizeTest.java
@@ -1,0 +1,129 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code and other text files for adherence to a set of rules.
+// Copyright (C) 2001-2024 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+///////////////////////////////////////////////////////////////////////////////////////////////
+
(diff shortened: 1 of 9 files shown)


CubeTures:
Ok I think it has to do with the SuperFinalizeCheck.java file. It's basically completely empty at the moment, so that's likely what's causing the issue. I'll keep messing with it and see where I get.

romani:
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/SuperFinalizeCheck.java

https://github.com/checkstyle/checkstyle/blob/73571a13c77c116253afac18b1cf9f59eafcac87/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/AbstractSuperCheck.java#L38-L45

try to put `public static final String MSG_KEY = "missing.super.call";` in SuperFinalizeCheck.java

CubeTures:
It seems that still failed. I also tried with the messages "finalize" and "super.finalize" and neither of those worked.

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionSuperFinalizeTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
+import com.puppycrawl.tools.checkstyle.checks.coding.SuperFinalizeCheck;
+import static com.puppycrawl.tools.checkstyle.checks.coding.SuperFinalizeCheck.MSG_KEY;
```
nrmancuso, line 30:
```suggestion
import static com.puppycrawl.tools.checkstyle.checks.coding.AbstractSuperCheck.MSG_KEY;
```

nrmancuso:
@CubeTures please squash all commits into one and make sure that `mvn clean verify` is passing locally before pushing up your changes.

CubeTures:
> @CubeTures please squash all commits into one and make sure that `mvn clean verify` is passing locally before pushing up your changes.

Ok, I will do that. What should I do in the future if I need help with a code then? Should I create the PR as a draft?

Additionally, the suggested changes still didn't change the error messages received... I'm unsure what the problem is at this point. Since I've tried all sorts of different MSG_KEYs.

CubeTures:
Actually, I think I found it! The original test for SuperFinalize had arguments with the verifications. I added them and stopped getting the error.

romani:
Please fix CI failures

CubeTures:
I am working on it. Unfortunately I ran into this latest error. This feels like its an error with the expected line itself. @romani have you ever seen something like this?
image

PR #14588, opened 2024-03-02
Issue #12586: Optimize testing methods to run check one time over a file

mahfouz72:
First step for #12586:   

this is not the final fix I just want to make sure that I am on the right direction also I faced some errors want to discuss them. IMO once this optimization is completed and reaches its final form, it will provide significant benefits. specifically aims to improve the process of ` mvn test` and `mvn clean verify.`

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
@@ -254,8 +254,11 @@ public abstract class AbstractModuleTestSupport extends AbstractPathTestSupport 
                 InlineConfigParser.parse(filePath);
         final DefaultConfiguration parsedConfig =
                 testInputConfiguration.createConfiguration();
-        verifyViolations(parsedConfig, filePath, testInputConfiguration.getViolations());
-        verify(parsedConfig, filePath, expected);
+        final List<String> actualViolations = getActualViolationsForFile(parsedConfig, filePath);
+        verifyViolations(filePath, testInputConfiguration.getViolations(), actualViolations);
+        assertWithMessage("Violations for %s differ.", filePath)
+            .that(actualViolations)
+            .containsExactlyElementsIn(expected);
     }
 
     /**
@@ -315,6 +318,25 @@ public abstract class AbstractModuleTestSupport extends AbstractPathTestSupport 
         verify(createChecker(parsedConfig), inputs, ImmutableMap.of(
             filePath1, expectedFromFile1,
             filePath2, expectedFromFile2));
+    }
+
+    /**
+     * Performs verification of the file with the given file path using specified configuration
+     * and the array expected messages. Also performs verification of the config specified in
+     * input file
+     *
+     * @param filePath file path to verify.
+     * @param expected an array of expected messages.
+     * @throws Exception if exception occurs during verification process.
+     */
+    protected void verifyWithInlineConfigParserTwice(String filePath, String... expected)
+            throws Exception {
+        final TestInputConfiguration testInputConfiguration =
+                InlineConfigParser.parse(filePath);
+        final DefaultConfiguration parsedConfig =
+                testInputConfiguration.createConfiguration();
+        verifyViolations(parsedConfig, filePath, testInputConfiguration.getViolations());
+        verify(parsedConfig, filePath, expected);
     }
 
     /**
@@ -492,6 +514,33 @@ public abstract class AbstractModuleTestSupport extends AbstractPathTestSupport 
                                   List<TestInputViolation> testInputViolations)
             throws Exception {
         final List<String> actualViolations = getActualViolationsForFile(config, file);
+        final List<Integer> actualViolationLines = actualViolations.stream()
+                .map(violation -> violation.substring(0, violation.indexOf(':')))
+                .map(Integer::valueOf)
+                .collect(Collectors.toUnmodifiableList());
+        final List<Integer> expectedViolationLines = testInputViolations.stream()
+                .map(TestInputViolation::getLineNo)
+                .collect(Collectors.toUnmodifiableList());
+        assertWithMessage("Violation lines for %s differ.", file)
+                .that(actualViolationLines)
+                .isEqualTo(expectedViolationLines);
+        for (int index = 0; index < actualViolations.size(); index++) {
+            assertWithMessage("Actual and expected violations differ.")
+                    .that(actualViolations.get(index))
+                    .matches(testInputViolations.get(index).toRegex());
+        }
+    }
+
+    /**
+     * Performs verification of violation lines.
+     *
+     * @param file file path.
+     * @param testInputViolations List of TestInputViolation objects.
+     * @param actualViolations for a file
+     */
+    private static void verifyViolations(String file,
+                                  List<TestInputViolation> testInputViolations,
+                                  List<String> actualViolations) {
         final List<Integer> actualViolationLines = actualViolations.stream()
                 .map(violation -> violation.substring(0, violation.indexOf(':')))
                 .map(Integer::valueOf)
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/WriteTagCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/WriteTagCheckTest.java
@@ -60,7 +60,7 @@ public class WriteTagCheckTest extends AbstractModuleTestSupport {
         final String[] expected = {
             "15: " + getCheckMessage(MSG_WRITE_TAG, "@author", "Daniel Grenner"),
         };
-        verifyWithInlineConfigParser(getPath("InputWriteTag.java"), expected);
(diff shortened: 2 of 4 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
```diff
@@ -235,8 +235,12 @@ protected final void verifyWithInlineConfigParser(String filePath, String... exp
...
         final DefaultConfiguration parsedConfig =
                 testInputConfiguration.createConfiguration();
-        verifyViolations(parsedConfig, filePath, testInputConfiguration.getViolations());
-        verify(parsedConfig, filePath, expected);
+        final List<String> actualViolations = getActualViolationsForFile(parsedConfig, filePath);
+        verifyViolations(filePath, testInputConfiguration.getViolations(), actualViolations);
+        
+        assertWithMessage("Violations for %s differ.", filePath)
+            .that(actualViolations)
+            .containsExactlyElementsIn(expected);
```
mahfouz72, line 243: Instead of running a check twice. once to verify the violation in `verifyViolations()` against testInputConfigurationViolation that got parsed from the input file. and another one with `verify()` to verify violations against the expected array.

I extracted running the check out  to `getActualViolationsForFile()` once and then used these actualViolations to verify them against the  testInputConfigurationViolation with `verifyViolations()` and then verify them with the expected array with  assertion

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
```diff
@@ -489,6 +493,32 @@ private void verifyViolations(Configuration config,
...
+                .collect(Collectors.toUnmodifiableList());
+        assertWithMessage("Violation lines for %s differ.", file)
+                .that(actualViolationLines)
+                .isEqualTo(expectedViolationLines);
+        for (int index = 0; index < actualViolations.size(); index++) {
+            assertWithMessage("Actual and expected violations differ.")
+                    .that(actualViolations.get(index))
+                    .matches(testInputViolations.get(index).toRegex());
+        }
+    }
```
mahfouz72, line 521: this is not supposed to be here at the final fix . I just overloaded the normal verifyViolation method to test my new changes. but the other remaining test methods still use the old verifyViolation So I just did that to make the old tests method work but at the end of the fix it should be only one verifyViolation method

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/WriteTagCheckTest.java
```diff
@@ -188,6 +188,7 @@ public void testWriteTagRecordsAndCompactCtors() throws Exception {
             getNonCompilablePath("InputWriteTagRecordsAndCompactCtors.java"), expected);
     }
 
+
     @Override
     protected void verify(Checker checker,
                           File[] processedFiles,
```
mahfouz72, line 194: this test is giving an error after my change. it uses a strange verify method overridden from the abstractTestSupport that doesn't verify the actual violation with the expected array and just reads the line (violation comment ) and asserts that it is equal to the expected array but the check doesn't getTheActualViolation by running check on the file like we do normally. may be I am missing something so please correct me because this tests are falling after my optimization

mahfouz72, line 194: some of these tests here are not passing because it is using an overridden verify method not the normal one in the abstractTestSupport. any guidance on how to solve this will be helpful

nrmancuso: I am not an expert on this part of the code, but you may find it useful to look back through git blame to see why these tests are different than others, and investigate the possibility of migrating the tests that deviate to the normal verify methods. In fact, this may even be some prework that you could do before solving this issue.

mahfouz72: @nrmancuso I looked in git blame didn't help much. but I think that the reason this test is different is that the default severity for this check is info so the `getViolationForFile()` is returning an empty array as info sevirty ( I am not sure please confirm that) . so this verify uses the output stream to test with it not the actual violation returned from the file

mahfouz72: is migrating back to use `verify` instead of `verifyWithInlineConfigParser` will be a valid solution

mahfouz72:
tried to override verifyWithInlineConfigParser() to use the old way of testing but got error as usage of verify is forbbeden 
```
forbidden method invocation: com.puppycrawl.tools.checkstyle.AbstractModuleTestSupport#verify(com.puppycrawl.tools.checkstyle.api.Configuration, java.lang.String, java.lang.String[]) [Use inline config parser instead.]
[ERROR]   in com.puppycrawl.tools.checkstyle.checks.javadoc.WriteTagCheckTest
```

romani:
We just need to reuse existing method in final fix. During transition to final state we can suppress such violation

nrmancuso:
@mahfouz72 where are we at here? Can we assist?

mahfouz72:
@nrmancuso I just need time to figure out why there are surviving mutations and how to kill them. it would be helpful If you could review my changes and tell me if there is something that is not correct and is the reason for all those survivals

mahfouz72:
@nrmancuso I just need time to get back to this PR figure out why there are surviving mutations and how to kill them. it would be helpful If you could review my changes and tell me if there is something that is not correct and is the reason for all those survivals

romani:
This means that we actually used double execution to kill mutation :).
We can execute some tests twice to kill them again :).

mahfouz72:
@romani I optimized one test only and still got survivals I don't why we still have tests that execute check twice

mahfouz72:
@romani do you have idea why this happened? .I debugged  and we didn't change the logic of testing at all the logic still the same we now just run the checker one time why we got this survivals?

romani:
Let's be not fanatic about to convert all at ones, keep treewalker on double run method for now. In next step/PR we will reduce double run to few tests methods, and may be we will keep such few test methods on double run.

Create verify method that kept old behavior of double run, and use it to satisfy pitest.

mahfouz72:
@romani Please review I kept tree walker tests on double run to satisfy pitest
I also kept writeTagCheckTest on double run due to https://github.com/checkstyle/checkstyle/pull/14588#discussion_r1510001407
In the next PR, we can fix other verify methods

romani:
Please fix https://app.circleci.com/pipelines/github/checkstyle/checkstyle/25485/workflows/7c31ed43-d542-45a6-983b-307fb0c38674/jobs/588676/artifacts

romani:
Please fix https://app.circleci.com/pipelines/github/checkstyle/checkstyle/25485/workflows/7c31ed43-d542-45a6-983b-307fb0c38674/jobs/588676/artifacts

mahfouz72:
@romani done can you please explain to me why we should make the method static and no warnings on other methods?
https://github.com/checkstyle/checkstyle/commit/6034763dc743d6739e60b763438eba5e12dad642

romani:
static modifier is kind of synonym to "state less" method, so does not depends on class fields, so can be used anywhere and does no side effects and simply return you a value based on parameters.

romani (approved):
ok to merge.

in next PRs we will try to minimize usage of double execution in other test methods.

nrmancuso (changes requested):
Items:

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
```diff
@@ -509,6 +531,33 @@ private void verifyViolations(Configuration config,
...
+                .map(TestInputViolation::getLineNo)
+                .collect(Collectors.toUnmodifiableList());
+        assertWithMessage("Violation lines for %s differ.", file)
+                .that(actualViolationLines)
+                .isEqualTo(expectedViolationLines);
+        for (int index = 0; index < actualViolations.size(); index++) {
+            assertWithMessage("Actual and expected violations differ.")
+                    .that(actualViolations.get(index))
+                    .matches(testInputViolations.get(index).toRegex());
+        }
```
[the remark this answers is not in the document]
mahfouz72: I tried a few but this one makes the violation diff more readable.

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
```diff
@@ -317,6 +320,25 @@ protected final void verifyWithInlineConfigParser(String filePath1,
...
+    /**
+     * Performs verification of the file with the given file path using specified configuration
+     * and the array expected messages. Also performs verification of the config specified in
+     * input file
+     *
+     * @param filePath file path to verify.
+     * @param expected an array of expected messages.
+     * @throws Exception if exception occurs during verification process.
+     */
+    protected void verifyWithInlineConfigParserDoubleRun(String filePath, String... expected)
```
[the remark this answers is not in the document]
mahfouz72: done

nrmancuso (approved):

PR #14594, opened 2024-03-03
Issue #12020: False positive in RightCurly with 'alone_or_singleline'

### LINKED ISSUE #12020: False positive in RightCurly with 'alone_or_singleline' RightCurlyOption in try/catch
I have read check documentation: https://checkstyle.org/checks/blocks/rightcurly.html#RightCurly

Related documentation: https://checkstyle.org/property_types.html#RightCurlyOption

From `alone_or_singleline`:
> The brace must be alone on the line, yet single-line format of block is allowed. 

```bash
/var/tmp $ javac RCurlyTest.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">

    <module name="TreeWalker">
        <module name="RightCurly">
            <property name="option" value="alone_or_singleline"/>
        </module>
    </module>
</module>

/var/tmp $ cat RCurlyTest.java
public class RCurlyTest {
	public static void main(String args[]) {
		try { System.out.println(args[0]); } catch (Exception e) {}
	}
}

/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.3.1-all.jar -c config.xml RCurlyTest.java
Starting audit...
[ERROR] [...]/RCurlyTest.java:3:44: '}' at column 38 should be alone on a line. [RightCurly]
Audit done.
```
---

Since the whole try..catch construct is in a single line, I expect the RightCurly test to pass (i.e. no errors raised).

As a side-note. Column number in the error message seems to be wrong when tabs are used in the source file. In the above example you can see that:

 * RCurlyTest.java:3:44 is correct and honours the top level tabWidth property in the config.xml file
 * However the message incorrectly refers to "column 38"

mahfouz72:
Resolves #12020:  
tryBlock at single line was raising a violation in `ALONE_OR_SINGLELINE`  option

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/b3282ceb40c5ca4a56ede9f6c5194a3f/raw/bb787f9f9d73023e38b43faae582a6e4f9d6d20d/rightcurly.xml

[report](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/9179d1b_2024114344/reports/diff/index.html) expected difference. violation has been removed for all single line try-catch blocks when `ALONE_OR_SINGLELINE`  option

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
@@ -328,9 +328,16 @@ public class RightCurlyCheck extends AbstractCheck {
             nextToken = Details.getNextToken(nextToken);
         }
 
-        if (nextToken != null && nextToken.getType() == TokenTypes.DO_WHILE) {
-            final DetailAST doWhileSemi = nextToken.getParent();
-            nextToken = Details.getNextToken(doWhileSemi);
+        // sibling tokens should be allowed on a single line
+        final int[] tokensWithBlockSibling = {
+            TokenTypes.DO_WHILE,
+            TokenTypes.LITERAL_FINALLY,
+            TokenTypes.LITERAL_CATCH,
+        };
+
+        if (TokenUtil.isOfType(nextToken, tokensWithBlockSibling)) {
+            final DetailAST parent = nextToken.getParent();
+            nextToken = Details.getNextToken(parent);
         }
 
         return TokenUtil.areOnSameLine(details.lcurly, details.rcurly)
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTryWithResourceAloneSingle.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTryWithResourceAloneSingle.java
@@ -32,7 +32,6 @@ class InputRightCurlyTryWithResourceAloneSingle {
                 BufferedReader br2 = new BufferedReader(br1)) { ; }
         catch (IOException e) { ; }
         try (BufferedReader br1 = new BufferedReader(null);
-                // violation below ''}' at column 64 should be alone on a line'
                 BufferedReader br2 = new BufferedReader(br1)) {} catch (IOException e) { ; }
         try (BufferedReader br1 = new BufferedReader(null);
                 BufferedReader br2 = new BufferedReader(br1)) {
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheckTest.java
@@ -307,6 +307,13 @@ public class RightCurlyCheckTest extends AbstractModuleTestSupport {
     }
 
     @Test
+    public void testAloneOrSingleLineTryCatchBlock() throws Exception {
+        final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
+        verifyWithInlineConfigParser(
+                getPath("InputRightCurlyTestAloneOrSinglelineTryCatchBlock.java"), expected);
+    }
+
+    @Test
     public void testCatchWithoutFinally() throws Exception {
         final String[] expected = {
             "19:9: " + getCheckMessage(MSG_KEY_LINE_SAME, "}", 9),
@@ -387,9 +394,8 @@ public class RightCurlyCheckTest extends AbstractModuleTestSupport {
     public void testTryWithResourceAloneSingle() throws Exception {
         final String[] expected = {
             "27:9: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 9),
-            "36:64: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 64),
-            "44:15: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 15),
-            "46:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "43:15: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 15),
+            "45:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
         };
         verifyWithInlineConfigParser(
                 getPath("InputRightCurlyTryWithResourceAloneSingle.java"), expected);
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTestAloneOrSinglelineTryCatchBlock.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTestAloneOrSinglelineTryCatchBlock.java
@@ -1,0 +1,30 @@
+/*
+RightCurly
+option = ALONE_OR_SINGLELINE
+tokens = LITERAL_TRY, LITERAL_CATCH, LITERAL_FINALLY
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.blocks.rightcurly;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+
+public class InputRightCurlyTestAloneOrSinglelineTryCatchBlock {
+
+     private void foo() {
+
+            try {
+                int i = 5; int b = 10;
+            }
+            catch (Exception e) { }
+     }
+     private void testSingleLineTryBlock() {
+         try { } catch (Exception e) { }
+         try {int x = 5;} catch (RuntimeException e) { } catch (Exception e) { }
+         try { } catch (RuntimeException e) { } catch (Exception e) { } finally { foo();}
+         try (BufferedReader br1 = new BufferedReader(null)) {} catch (IOException e) { ; }
+    }
+
+}


src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTryWithResourceAloneSingle.java
```diff
@@ -31,9 +31,6 @@ void test() throws IOException {
         try (BufferedReader br1 = new BufferedReader(null);
                 BufferedReader br2 = new BufferedReader(br1)) { ; }
         catch (IOException e) { ; }
-        try (BufferedReader br1 = new BufferedReader(null);
-                // violation below ''}' at column 64 should be alone on a line'
-                BufferedReader br2 = new BufferedReader(br1)) {} catch (IOException e) { ; }
```
mahfouz72, line 36: Is this supposed to be a violation and not treated as one block in a single line? because after my changes the check detects this as a single block on a line so, no violation. all the similar cases for other tokens are not detected as violation also examples:

```
PS D:\test> cat config.xml
<?xml version="1.0"?>

 <module name="TreeWalker">
        <module name="RightCurly">
            <property name="option" value="alone_or_singleline"/>
        </module>
    </module>
</module>
PS D:\test> cat src/Test.java
public class Test {

        if (true
                && false ) { } else { }

        do {int a = 5;
        int x = 7;
        int m = 3;} while (true);
    }
    public int add (int n
            , int m
            , int x) {  return n + m + x;}
}

PS D:\test> java  -jar checkstyle-10.13.0-all.jar -c config.xml src/Test.java
Starting audit...
Audit done.
PS D:\test> 

```
so if these cases are supposed to be a violation. I can fix it but in this PR? or open new issue

mahfouz72:
Github, generate report

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTestAloneOrSingleline.java
```diff
@@ -10,6 +10,9 @@
 
 package com.puppycrawl.tools.checkstyle.checks.blocks.rightcurly;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+
 public class InputRightCurlyTestAloneOrSingleline {
```
rnveach, line 16: Move new test to a whole new file. We don't like modifying existing tests and causing line differences.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTryWithResourceAloneSingle.java
rnveach, line 36: Test input code should not be removed. It should stay as is, and either be a violation or not.

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -328,9 +328,12 @@ private static boolean isBlockAloneOnSingleLine(Details details) {
             nextToken = Details.getNextToken(nextToken);
         }
 
-        if (nextToken != null && nextToken.getType() == TokenTypes.DO_WHILE) {
-            final DetailAST doWhileSemi = nextToken.getParent();
-            nextToken = Details.getNextToken(doWhileSemi);
+        if (nextToken != null && (nextToken.getType() == TokenTypes.DO_WHILE
+                || nextToken.getType() == TokenTypes.LITERAL_FINALLY
+                || nextToken.getType() == TokenTypes.LITERAL_CATCH)) {
+
```
rnveach, line 334: Remove empty line.

rnveach (changes requested):

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTryWithResourceAloneSingle.java
rnveach: This is alone or singleline. The try isn't a single line obviously, but is it alone? If it is alone, then no violation is expected.

> alone 	The brace must be alone on the line.

This does not look to be alone, so a violation is expected IMO.

rnveach:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/9179d1b_2024114344/reports/diff/apache-ant/index.html#A1
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/9179d1b_2024114344/reports/diff/apache-ant/xref/home/runner/work/checkstyle/checkstyle/.ci-temp/contribution/checkstyle-tester/repositories/apache-ant/src/tests/junit/org/apache/tools/ant/taskdefs/optional/EchoPropertiesTest.java.html#L236

The option is alone or single line. I am expecting no violations if the code is EITHER on a single line or if its right curly is alone. Everything on this one line looks to be on a single line.

This looks like what is reported in issue unless I am missing something.

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
mahfouz72: @rnveach done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTryWithResourceAloneSingle.java
mahfouz72: @rnveach done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTestAloneOrSingleline.java
mahfouz72: done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTryWithResourceAloneSingle.java
mahfouz72:
```
private static boolean isBlockAloneOnSingleLine(Details details) {
       .......
        return TokenUtil.areOnSameLine(details.lcurly, details.rcurly)
            && (nextToken == null || !TokenUtil.areOnSameLine(details.rcurly, nextToken)
                || isRightcurlyFollowedBySemicolon(details));
    }
```
this is the code to detect a single block in a line it is checking if the rcurly and lcurly in single line and the rcurly and the next token is not in the same line

so in this case 
```
try( x = 5;
                 y = 6) {} catch {}
                           l r
```
the rcurly and lcurly on same line so single block in a line

mahfouz72: the reason for this was detected as a violation before because the check wasn't even try to  detect try-catch block as block in a single line but this is not the case now as after this PR the check detect it as a single line so no violation

mahfouz72: I will open a new issue for this as this problem is not a part of the issue

rnveach: I don't know what I am being pinged in this section for.

rnveach (approved):

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyTryWithResourceAloneSingle.java
mahfouz72: to review the new issue if it should be approved

rnveach: I don't review issues for approval. That is @romani

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -328,9 +328,11 @@ private static boolean isBlockAloneOnSingleLine(Details details) {
             nextToken = Details.getNextToken(nextToken);
         }
 
-        if (nextToken != null && nextToken.getType() == TokenTypes.DO_WHILE) {
-            final DetailAST doWhileSemi = nextToken.getParent();
-            nextToken = Details.getNextToken(doWhileSemi);
+        if (nextToken != null && (nextToken.getType() == TokenTypes.DO_WHILE
+                || nextToken.getType() == TokenTypes.LITERAL_FINALLY
+                || nextToken.getType() == TokenTypes.LITERAL_CATCH)) {
```
nrmancuso, line 333: Let's do a favor for the next person to come across this code: please extract these tokens to some descriptively named collection, and do something like this:
```suggestion
        final int[] myAwesomeCollection = {TokenTypes.DO_WHILE, ...};
        if (TokenUtil.isOfType(nextToken, myAwesomeCollection) {
```

I am wondering why there isn't more tokens in this condition (LITERAL_FOR, LITERAL_IF, etc.), so it would be good to point it out by the name of this collection.

nrmancuso (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
mahfouz72: > please extract these tokens to some descriptively named collection,

@nrmancuso done I like this update but I am not sure what descriptive name could be nothing on my mind. we just checking this tokens as we have  to get parent first before getting the next token so ...? 🤔 




> I am wondering why there isn't more tokens in this condition

yea I was wondering the same but I analyzed it and it turned out that if the next token is IF or for then it is ok we don't need to update it as it will never be a second of block of related group of blocks ( no relation between If and for and its previous tokens) 
but in case that let's say the next token is a catch (the current token is try) then we have update the next token to be the one after the catch as try-catch block is treated as one block in alone or single line option consider this
```
try{ }catch{ }
```
when we want to validate that try is a block on single line we have to check that is not in the same line with the next token to its next token (it may be method braces or something else)
but consider` for( ... ) if( )` no relation between this two and they can't both be considered as block alone on single line so that's why if the next token is If or for we leave it as it is and not get its next token

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -328,9 +328,10 @@ private static boolean isBlockAloneOnSingleLine(Details details) {
             nextToken = Details.getNextToken(nextToken);
         }
 
-        if (nextToken != null && (nextToken.getType() == TokenTypes.DO_WHILE
-                || nextToken.getType() == TokenTypes.LITERAL_FINALLY
-                || nextToken.getType() == TokenTypes.LITERAL_CATCH)) {
+        final int[] tokensToCheck = {
+            TokenTypes.DO_WHILE, TokenTypes.LITERAL_FINALLY, TokenTypes.LITERAL_CATCH,
+        };
+        if (TokenUtil.isOfType(nextToken, tokensToCheck)) {
```
nrmancuso, line 334:
```suggestion
        // sibling tokens should be allowed on a single line
        final int[] tokensWithBlockSibling = {
            TokenTypes.DO_WHILE,
            TokenTypes.LITERAL_FINALLY,
            TokenTypes.LITERAL_CATCH,
        };
        
        if (TokenUtil.isOfType(nextToken, tokensWithBlockSibling)) {
```

nrmancuso (changes requested):
Last item:

nrmancuso:
Github, generate report

nrmancuso (approved):
Thank you very much @mahfouz72 !!

Issue #13309, opened 2023-06-26
False positive error from JavadocType with record


turbanoff:
I have read check documentation: https://checkstyle.org/checks/javadoc/javadoctype.html
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac YOUR_FILE.java

/var/tmp $ cat config.xml
<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="TreeWalker">
        <module name="JavadocType">
            <property name="allowMissingParamTags" value="true"/>
            <property name="allowUnknownTags" value="true"/>
        </module>
    </module>
</module>

/var/tmp $ cat MyTransaction.java
package bkk;

/**
 * One Transaction.
 *
 * @param transactionId  unique ID of the transaction in format: {@code <first Code>:<second Code>}
 */
public record MyTransaction(String transactionId) {
}

/var/tmp $ java -Duser.language=en -Duser.country=US -jar checkstyle-10.12.1-all.jar -c config.xml MyTransaction.java
Starting audit...
[ERROR] D:\Projects\app4Test\checkstyle-test\src\main\java\bkk\MyTransaction.java:6:4: Unused @param tag for 'transactionId'. [JavadocType]
Audit done.
Checkstyle ends with 1 errors.
```
---

I expect no warning or errors here. Javadoc is properly written in the proper place.

---

turbanoff:
Here
![image](https://github.com/checkstyle/checkstyle/assets/741251/b01d6e21-0e74-47a3-a295-498d0e2bca13)
[MyTransaction.zip](https://github.com/checkstyle/checkstyle/files/11870164/MyTransaction.zip)

romani:
I am on phone, please show html on constructor detailed section.

turbanoff:
Constructor + getter.
![изображение](https://github.com/checkstyle/checkstyle/assets/741251/9544eaf5-41ea-4077-9c9b-8fb3b1d48123)

sktpy:
**Preliminary Investigation:**
[...]

PR #14603, opened 2024-03-04
Issue #13309: False positive error from JavadocType with record

sktpy:
Resolves https://github.com/checkstyle/checkstyle/issues/13309

<hr>

**Regression:** ([Report](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/0fbf9da_2024125210/reports/diff/index.html)) No Differences Found
Expected to have some differences(specifically removed violations) as we have fixed a false positive,
but we do not see any due the very specific scope of this false positive. ([Reference](https://github.com/checkstyle/checkstyle/issues/13309#issuecomment-1971651949))

The fix is working as intended and can be verified using the added test inputs, which previously would have produced a false violation.

Diff Regression config: https://gist.githubusercontent.com/sktpy/44a31ea4e946a0755fa3e331fd65de98/raw/df79e294a1b3f1fc37a6fdc971273c2dd746cea2/check.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
@@ -190,7 +190,7 @@ public class JavadocTypeCheck
 
     /** Pattern to match type name within angle brackets in javadoc param tag. */
     private static final Pattern TYPE_NAME_IN_JAVADOC_TAG =
-            Pattern.compile("\\s*<([^>]+)>.*");
+            Pattern.compile("^<([^>]+)");
 
     /** Pattern to split type name field in javadoc param tag. */
     private static final Pattern TYPE_NAME_IN_JAVADOC_TAG_SPLITTER =
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheckTest.java
@@ -358,6 +358,35 @@ public class JavadocTypeCheckTest extends AbstractModuleTestSupport {
     }
 
     @Test
+    public void testJavadocTypeParamDescriptionWithAngularTags() throws Exception {
+        final String[] expected = {
+            "44:4: " + getCheckMessage(MSG_UNUSED_TAG, "@param", "<P>"),
+            "46:1: " + getCheckMessage(MSG_MISSING_TAG, "@param <U>"),
+            "50:4: " + getCheckMessage(MSG_UNUSED_TAG, "@param", "region"),
+        };
+
+        verifyWithInlineConfigParser(
+                getPath("InputJavadocTypeParamDescriptionWithAngularTags.java"), expected);
+    }
+
+    @Test
+    public void testJavadocTypeRecordParamDescriptionWithAngularTags() throws Exception {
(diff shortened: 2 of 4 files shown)


sktpy:
GitHub, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
```diff
@@ -190,7 +190,7 @@ public class JavadocTypeCheck
 
     /** Pattern to match type name within angle brackets in javadoc param tag. */
     private static final Pattern TYPE_NAME_IN_JAVADOC_TAG =
-            Pattern.compile("\\s*<([^>]+)>.*");
```
sktpy, line 193: The following removals weren't the cause of the issue, but are removed because they are not required.

### Removal of `\\s`:
We are guaranteed to have no space before type name, because the input is already trimmed in the following line (performed before this method is called):
https://github.com/checkstyle/checkstyle/blob/fa8b95c964251f2156db988c8d05215ea0a80368/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/utils/BlockTagUtil.java#L81

Call stack:
```java
extractBlockTags:57, BlockTagUtil (com.puppycrawl.tools.checkstyle.checks.javadoc.utils)
getJavadocTags:102,  JavadocUtil (com.puppycrawl.tools.checkstyle.utils)
getJavadocTags:374,  JavadocTypeCheck (com.puppycrawl.tools.checkstyle.checks.javadoc)
visitToken:323,      JavadocTypeCheck (com.puppycrawl.tools.checkstyle.checks.javadoc)
```
<hr>

### Removal of `>`:
If the type name starts with `<` it must be ended by `>` for it to be valid input, in any other case javadoc will throw an error and such tag wouldn't be rendered.
<hr>

### Removal of `>.*`:
What is beyond the type name does not concerns us in this particular method, hence no need to check anything.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
```diff
@@ -190,7 +190,7 @@ public class JavadocTypeCheck
...
+            Pattern.compile("^<([^>]+)");
```
sktpy, line 193: Addition of `^` here is what fixes this issue.

**Finding nemo:**
Let's consider the input submitted by author of the issue:
```java
/**
 * One Transaction.
 *
 * @param transactionId  unique ID of the transaction in format: {@code <first Code>:<second Code>}
 */
 ```
 
The pattern in question is used by the following method to match for generic tag type name:
https://github.com/checkstyle/checkstyle/blob/fa8b95c964251f2156db988c8d05215ea0a80368/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java#L496

The pattern is used on the tag's `firstArg`, which in this case would be:
`transactionId  unique ID of the transaction in format: {@code <first Code>:<second Code>}`

The old pattern matched angular tag type content anywhere in the line, matching `<first Code>` here, wrongly identifying it it as generic type parameter (which is not) and returning `first Code` as the param name, but the param here is `transactionId`.

The param name is always at the beggining of the tag's `firstArg`, therefore matching the pattern at the begging of line solved the problem.

rnveach (approved):

nrmancuso (approved):
Thanks a lot for the easy to review PR!

Issue #3110, opened 2016-04-16
some messages are still hardcoded in english


rnveach:
CS still has some output events that are hardcoded in the code and are only in English. Some already have translations in message files, but we don't use them.
We translate check messages to their specific language so non-english users can understand the issue, we should do the same with every output of CS (minus XML attributes and such) so there is no confusion.

Examples:
https://github.com/checkstyle/checkstyle/blob/e18195de775c9ee707f31760820bc45cd1c9aa91/src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java#L130-L143
https://github.com/checkstyle/checkstyle/blob/e18195de775c9ee707f31760820bc45cd1c9aa91/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L192

MaggieMa21:
I am on this.

subkrish:
@romani @rnveach Could you provide any other examples for translated messages?

romani:
Why this example is not enough ? Please share your problem

subkrish:
@romani I am having a trouble in trying to understand what class to import to use localized messages. I was trying to follow how localized messages have been used elsewhere. But due to different implementations, I wasn't sure how to implement it.

romani:
it is better to try another issue for you.

subkrish:
@romani Okay but I give trying to solve the issue my best shot and if I fall short I'll work on a different issue.

romani:
@subkrish , please proceed with hardcoded lines at:
- Main
- Checker
- please recheck that non of Exception throw message is left with hardcoded content.

subkrish:
@romani @rnveach Looks like the message for this https://github.com/checkstyle/checkstyle/blob/e18195de775c9ee707f31760820bc45cd1c9aa91/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L192 is not present in the 'messages.properties' file and the subsequent other languages files. Do I add them for the languages present and then remove the hardcoded lines? I might have to use a translator for the other languages messages.

rnveach:
> Do I add them for the languages present and then remove the hardcoded lines? 

Yes.

>  I might have to use a translator for the other languages messages.

That is what everyone does. No one in Checkstyle of us know all these languages.

subkrish:
for the Main file changes should also be made to command line option that only print in english right?
For example:
```
// show version and exit if it is requested
            if (commandLine.hasOption(OPTION_V_NAME)) {
[...]

PR #14614, opened 2024-03-05
Issue #3110: BugFix in Checker.java removed hardcoded exception messages of finishLocalSetup method

MANISH-K-07:
Part of #3110 

All hardcoded messages in finishLocalSetup  method of Checker.java have been removed and replaced with keys pointing to respective messages.properties files to support non-locale translations.

I have used Google translate to extract translations of the messages.
PS : Simplified Chinese has been used for messages_zh.properties

### DIFF
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
@@ -1,12 +1,10 @@
+Checker.finishLocalSetup=if no custom moduleFactory is set, moduleClassLoader must be specified
 DefaultLogger.addException=Error auditing {0}
 DefaultLogger.auditFinished=Audit done.
 DefaultLogger.auditStarted=Starting audit...
 general.exception=Got an exception - {0}
 general.fileNotFound=File not found!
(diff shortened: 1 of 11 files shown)


MANISH-K-07:
If these changes are ideal, I will proceed with working on one complete file per each PR.

romani (approved):
Ok to merge

src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
```diff
@@ -1,12 +1,16 @@
+#
+# for Checker.java
+#
```
nrmancuso, line 3: This comment leads me to believe that this entire file is for Checker messages, or at least all messages up to the next comment like this; the other comments are equally unhelpful and misleading IMO. We don't need to do this, please remove these. The key should tell us everything we need to know (it does).

nrmancuso (changes requested):
Item:

src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
MANISH-K-07: Done..
Rebased on master

nrmancuso (approved):
Good to merge as CI passes

PR #12640, opened 2023-01-14
Issue #9994: LITERAL_SWITCH token support in RightCurlyCheck

Kevin222004:
Issue #9994: LITERAL_SWITCH token support in RightCurlyCheck

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/75ac442a047d7fe25bedb544f9acd754/raw/52982ec6530ca6a4bfb5d24054bf816649cbe4fd/config.xml 

Diff Regression patch config: https://gist.githubusercontent.com/Kevin222004/7049aace77660fb0f0ac515ee7996c83/raw/1ef1a12723d35019f756bb0b11ed855e1d86ca92/patch_config.xml

PR #14623, opened 2024-03-07
Issue #14620:  `LITERAL_CASE` token support in RightCurlyCheck

### LINKED ISSUE #14620: LITERAL_CASE token support in RightCurlyCheck
Currently [RightCurlyCheck](https://checkstyle.org/checks/blocks/rightcurly.html#RightCurly) accepts a less tokens than [LeftCurlyCheck](https://checkstyle.org/checks/blocks/leftcurly.html#LeftCurly) (19 tokens)
This issue aims to add support for `LITERAL_CASE` token in RightCurlyCheck 
this issue is a child issue for https://github.com/checkstyle/checkstyle/issues/3547

For LeftCurlyCheck:

```
PS D:\test> cat config.xml                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">

 <module name="TreeWalker">

      <module name="LeftCurly">
            <property name="tokens" value="LITERAL_CASE "/>
          <property name="option" value="eol"/>
        </module>
    </module>
</module>
PS D:\test> cat src/Test.java                                                
class Test{

    void test(){
        int x = 1;
        switch(x){
            case 1:
            {        // violation
                System.out.println("x is 1");
                break;
            }
            case 2:
            {     // violation
                System.out.println("x is 2");
                break;
            }
            default:
                System.out.println("x is neither 1 nor 2");
        }
    }
}
PS D:\test> java  -jar checkstyle-10.13.0-all.jar -c config.xml src/Test.java
Starting audit...
[ERROR] D:\test\src\Test.java:7:13: '{' at column 13 should be on the previous line. [LeftCurly]
[ERROR] D:\test\src\Test.java:12:13: '{' at column 13 should be on the previous line. [LeftCurly]
Audit done.
```
for RightCurlyCheck:

```
PS D:\test> cat config.xml                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">

 <module name="TreeWalker">

      <module name="RightCurly">
            <property name="tokens" value="LITERAL_CASE"/>
          <property name="option" value="alone"/>
        </module>
    </module>
</module>
PS D:\test> cat src/Test.java                                                
class Test{

    void test(){
        int x = 1;
        switch(x){
            case 1:
            {
                System.out.println("x is 1");
                break;
            } case 2:                       // expected violation '}' should be alone in the line
            {
                System.out.println("x is 2");
                break;
            } default:                        // expected violation '}' should be alone in the line
                System.out.println("x is neither 1 nor 2");
        }
    }
}
```
Expected:- Violation for the right curly brace for  `LITERAL_CASE `

mahfouz72:
Resolves #14620 :

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/c82dbc1b4a05567e245729eccfc6cf62/raw/c4958b5ee00ded9c8f89d3af0a50fa871b352205/rcurlybase.xml

Diff Regression patch config: https://gist.githubusercontent.com/mahfouz72/fb26eda6b5a3264c0c42351e612cc4f6/raw/daf318509889f654cd61d58c9e039c9ba1505716/rcurlypatch.xml

### Reports

- alone option ([link](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/ddf54a1_2024014514/reports/diff/index.html))

- alone_or_singleline option ([link](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/ddf54a1_2024024511/reports/diff/index.html  ))

- same option ([link](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/ddf54a1_2024001633/reports/diff/index.html))

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
@@ -21,6 +21,7 @@ package com.puppycrawl.tools.checkstyle.checks.blocks;
 
 import java.util.Arrays;
 import java.util.Locale;
+import java.util.Optional;
 
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
@@ -32,8 +33,8 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
 /**
  * <p>
  * Checks the placement of right curly braces (<code>'}'</code>) for code blocks. This check
- * supports if-else, try-catch-finally blocks, switch statements, while-loops, for-loops,
- * method definitions, class definitions, constructor definitions,
+ * supports if-else, try-catch-finally blocks, switch statements, switch cases, while-loops,
+ *  for-loops, method definitions, class definitions, constructor definitions,
  * instance, static initialization blocks, annotation definitions and enum definitions.
  * For right curly brace of expression blocks of arrays, lambdas and class instances
  * please follow issue
@@ -155,6 +156,7 @@ public class RightCurlyCheck extends AbstractCheck {
             TokenTypes.RECORD_DEF,
             TokenTypes.COMPACT_CTOR_DEF,
             TokenTypes.LITERAL_SWITCH,
+            TokenTypes.LITERAL_CASE,
         };
     }
 
@@ -432,6 +434,9 @@ public class RightCurlyCheck extends AbstractCheck {
                 case TokenTypes.LITERAL_SWITCH:
                     details = getDetailsForSwitch(ast);
                     break;
+                case TokenTypes.LITERAL_CASE:
+                    details = getDetailsForCase(ast);
+                    break;
                 default:
                     details = getDetailsForOthers(ast);
                     break;
@@ -458,6 +463,38 @@ public class RightCurlyCheck extends AbstractCheck {
                 nextToken = getNextToken(switchNode);
             }
             return new Details(lcurly, rcurly, nextToken, true);
+        }
+
+        /**
+         * Collects details about case statements.
+         *
+         * @param caseNode case statement to gather details about
+         * @return new Details about given case statement
+         */
+        private static Details getDetailsForCase(DetailAST caseNode) {
+            final DetailAST caseParent = caseNode.getParent();
+            final int parentType = caseParent.getType();
+            final Optional<DetailAST> lcurly;
+            final DetailAST statementList;
+
+            if (parentType == TokenTypes.SWITCH_RULE) {
+                statementList = caseParent.findFirstToken(TokenTypes.SLIST);
+                lcurly = Optional.ofNullable(statementList);
+            }
+            else {
+                statementList = caseNode.getNextSibling();
+                lcurly = Optional.ofNullable(statementList)
+                         .map(DetailAST::getFirstChild)
+                         .filter(node -> node.getType() == TokenTypes.SLIST);
+            }
+            final DetailAST rcurly = lcurly.map(DetailAST::getLastChild)
+                    .filter(child -> !isSwitchExpression(caseParent))
+                    .orElse(null);
+            final Optional<DetailAST> nextToken =
+                    Optional.ofNullable(lcurly.map(DetailAST::getNextSibling)
+                    .orElseGet(() -> getNextToken(caseParent)));
+
+            return new Details(lcurly.orElse(null), rcurly, nextToken.orElse(null), true);
         }
 
         /**
--- a/src/main/resources/google_checks.xml
+++ b/src/main/resources/google_checks.xml
@@ -96,7 +96,7 @@
       <property name="tokens"
                value="CLASS_DEF, METHOD_DEF, CTOR_DEF, LITERAL_FOR, LITERAL_WHILE, STATIC_INIT,
                     INSTANCE_INIT, ANNOTATION_DEF, ENUM_DEF, INTERFACE_DEF, RECORD_DEF,
-                    COMPACT_CTOR_DEF, LITERAL_SWITCH"/>
+                    COMPACT_CTOR_DEF, LITERAL_SWITCH, LITERAL_CASE"/>
     </module>
     <module name="SuppressionXpathSingleFilter">
       <!-- suppresion is required till https://github.com/checkstyle/checkstyle/issues/7541 -->
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheckTest.java
@@ -827,4 +827,158 @@ public class RightCurlyCheckTest extends AbstractModuleTestSupport {
         verifyWithInlineConfigParser(
                 getNonCompilablePath("InputRightCurlyTestSwitchExpression7.java"), expected);
     }
+
+    @Test
+    public void testCaseBlocksInSwitchStatementAlone() throws Exception {
+        final String[] expected = {
+            "33:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "44:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "73:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "78:44: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 44),
+            "90:39: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 39),
+            "97:42: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 42),
+            "107:35: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 35),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputRightCurlyCaseBlocksInSwitchStatementAlone.java"), expected);
+    }
+
+    @Test
+    public void testCaseBlocksInSwitchStatementAlone2() throws Exception {
+        final String[] expected = {
+            "17:15: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 15),
+            "26:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "38:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "41:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "49:24: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 24),
+            "49:45: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 45),
+            "57:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "68:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "68:26: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 26),
+            "75:34: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 34),
+            "75:47: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 47),
+            "82:23: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 23),
+            "102:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputRightCurlyCaseBlocksInSwitchStatementAlone2.java"), expected);
+    }
+
+    @Test
+    public void testCaseBlocksInSwitchStatementAloneOrSingleLine() throws Exception {
+        final String[] expected = {
+            "33:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "44:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "73:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "78:51: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 51),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputRightCurlyCaseBlocksInSwitchStatementAloneOrSingleline.java"),
+                expected);
+    }
+
+    @Test
+    public void testCaseBlocksInSwitchStatementAloneOrSingle2() throws Exception {
+        final String[] expected = {
+            "18:15: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 15),
+            "27:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "39:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "42:15: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 15),
+            "50:23: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 23),
+            "71:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "80:23: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 23),
+            "100:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputRightCurlyCaseBlocksInSwitchStatementAloneOrSingleline2.java"),
+                expected);
+    }
+
+    @Test
+    public void testCaseBlocksInSwitchStatementSame() throws Exception {
+        final String[] expected = {
+            "33:16: " + getCheckMessage(MSG_KEY_LINE_BREAK_BEFORE, "}", 16),
+            "44:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "73:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "78:51: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 51),
+            "106:23: " + getCheckMessage(MSG_KEY_LINE_BREAK_BEFORE, "}", 23),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputRightCurlyCaseBlocksInSwitchStatementSame.java"), expected);
+    }
+
+    @Test
+    public void testCaseBlocksInSwitchStatementSame2() throws Exception {
+        final String[] expected = {
+            "18:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "27:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "39:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "42:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "50:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "70:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "77:34: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 34),
+            "77:47: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 47),
+            "85:23: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 23),
+            "103:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputRightCurlyCaseBlocksInSwitchStatementSame2.java"), expected);
+    }
+
+    @Test
+    public void testCaseBlocksWithSwitchRuleAlone() throws Exception {
+        final String[] expected = {
+            "32:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "44:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "45:27: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 27),
+            "56:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "69:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "74:46: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 46),
+            "83:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "83:36: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 36),
+            "93:31: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 31),
+            "94:31: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 31),
+        };
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputRightCurlyCaseBlocksWithSwitchRuleAlone.java"),
+                expected);
+    }
+
+    @Test
+    public void testCaseBlocksWithSwitchRuleAloneOrSingleLine() throws Exception {
+
+        final String[] expected = {
+            "27:19: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 19),
+            "41:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "53:23: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 23),
+            "54:27: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 27),
+            "64:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "77:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "82:46: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 46),
+            "91:13: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 13),
+            "102:31: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 31),
+        };
+        final String fileName = "InputRightCurlyCaseBlocksWithSwitchRuleAloneOrSingleline.java";
+        verifyWithInlineConfigParser(getNonCompilablePath(fileName), expected);
+    }
+
+    @Test
+    public void testCaseBlocksWithSwitchExpressionAlone() throws Exception {
+        final String[] expected = {
+            "63:31: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 31),
+            "86:42: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 42),
+        };
+        final String fileName = "InputRightCurlyCaseBlocksWithSwitchExpressionAlone.java";
+        verifyWithInlineConfigParser(getNonCompilablePath(fileName), expected);
+    }
+
+    @Test
+    public void testCaseBlocksWithSwitchExpressionAloneOrSingleLine() throws Exception {
+        final String[] expected = {
+            "63:31: " + getCheckMessage(MSG_KEY_LINE_ALONE, "}", 31),
+        };
+        final String fileName =
+                "InputRightCurlyCaseBlocksWithSwitchExpressionAloneOrSingleline.java";
+        verifyWithInlineConfigParser(getNonCompilablePath(fileName), expected);
+    }
 }
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksWithSwitchExpressionAloneOrSingleline.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksWithSwitchExpressionAloneOrSingleline.java
@@ -1,0 +1,93 @@
+/*
+RightCurly
+option = ALONE_OR_SINGLELINE
+tokens = LITERAL_CASE, LITERAL_IF
+
+*/
+
+//non-compiled with javac: Compilable with Java14
+package com.puppycrawl.tools.checkstyle.checks.blocks.rightcurly;
+
+public class InputRightCurlyCaseBlocksWithSwitchExpressionAloneOrSingleline {
+
+    int foo(int yield) {
+        return switch (yield) {
+            case 1 -> 2;
+            case 2 -> 3;
+            case 3 -> 4;
+            default -> 5;
+        };
+    }
+
+    void test() {
+        int x = 1;
+        String s = switch (x) {
+            case 1 -> {
+
+                yield "and that's it";
+            }  case 2 -> "x is 2";
+            default -> "x is neither 1 nor 2";
+        };
+    }
+
+    void test2() {
+        int x = 1;
+        String s = switch (x) {
+            case 1 -> {
+
+                yield "and that's it";
+            }  case 2 -> { yield "x is 2";}
+            default -> "x is neither 1 nor 2";
+        };
+    }
+
+    void test3() {
+        int x = 1;
+         String s = switch (x) {
+            case 1 -> {
+
+                String s1 = "and that's it";
+                yield s1;
+            } case 2 -> { yield "x is 2";}
+            default -> "x is neither 1 nor 2";
+        };
+    }
+
+    void test4() throws Exception {
+           int a = 0;
+           int b = switch (a) {
+                case 0 -> {
+                    int x = 5;
+                    int y = 6;
+                    if (a == 2) {
+                        y = 7;}   // violation '}' at column 31 should be alone on a line'
+                    throw new Exception();
+                }  default -> 2;
+            };
+    }
+
+    void test5() throws Exception {
+        int a = 0;
+        int b = switch (a) {
+            case 0 -> {throw new Exception();}
+            default -> 2;
+        };
+    }
+
+    void test6() {
+         int y = 0;
+         int x = 0;
+         final boolean a = switch (y) {
+             case 1 -> {x = 1; yield true;
+             }case 2 -> {
+                 x = 1;
+                 yield true;} case 3 -> {
+                 x = 1;
+                 if (x == 1) {yield true;} // ok single line
+                 yield false;
+             }
+             default -> throw new RuntimeException();
+         };
+    }
+
+}
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksWithSwitchRuleAloneOrSingleline.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksWithSwitchRuleAloneOrSingleline.java
@@ -1,0 +1,104 @@
+/*
+RightCurly
+option = ALONE_OR_SINGLELINE
+tokens = LITERAL_CASE
+
+*/
+
+//non-compiled with javac: Compilable with Java14
+package com.puppycrawl.tools.checkstyle.checks.blocks.rightcurly;
+
+public class InputRightCurlyCaseBlocksWithSwitchRuleAloneOrSingleline {
+
+    public void testt0() {
+        int mode = 0;
+        switch (mode) {
+            case 1 -> {int x = 1;}   // ok, single line
+            case 2 -> {int x = 0;}
+            case 3 -> System.out.println("x is 3");
+        }
+    }
+
+    public static void test0() {
+        int mode = 0;
+        switch (mode) {
+            case 1 -> {
+                int
+            x = 1;} // violation '}' at column 19 should be alone on a line'
+            case 2 -> {
+                int x = 0;
+            }
+            case 3 -> System.out.println("x is 3");
+        }
+    }
+
+    public static void test() {
+        int mode = 0;
+        switch (mode) {
+            case 1 -> {
+                int x = 1;
+
+            } default -> {          // violation '}' at column 13 should be alone on a line'
+                int x = 0;
+            }
+        }
+    }
+
+    public static void test1() {
+        int k = 0;
+        switch (k) {
+            case 1 -> {
+
+
+           int x = 1; } case 2 -> {       // violation '}' at column 23 should be alone on a line'
+                int x = 2;}     // violation '}' at column 27 should be alone on a line'
+
+        }
+    }
+
+    public static void test2() {
+         int k = 0;
+         switch (k) {
+            case 1 -> {int x = 1;
+
+            }case 2 -> throw new RuntimeException();
+            // violation above '}' at column 13 should be alone on a line'
+         }
+    }
+
+    public static void test3() {
+         int k = 0;
+         switch (k) {
+            case 1 -> {
+                 int x = 1;
+            }
+            case 2 -> {
+                    int x = 2;
+            } }      // violation '}' at column 13 should be alone on a line'
+    }
+
+    public static void test4() {
+        int mode = 0;
+        switch (mode) { case 0 -> {int x = 1;} default -> {int x = 0;} }
+        // violation above '}' at column 46 should be alone on a line'
+    }
+
+    public static void test5() {
+        int mode = 0;
+        switch (mode) {
+            case 0 -> {
+                int x = 1;
+            } case 1 -> {int x = 2;}  // violation '}' at column 13 should be alone on a line'
+        }
+    }
+
+    public static void test8() {
+        int mode = 0;
+        int x = 0;
+        switch (mode) {
+            case 0 -> throw new RuntimeException();
+            case 1 -> x = 1;
+            case 2 -> {x = 1; }      // ok, single line
+            case 3 -> {x = 5; } }   // violation '}' at column 31 should be alone on a line'
+    }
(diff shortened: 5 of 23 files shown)


mahfouz72:
@nrmancuso @rnveach can you guide me on how to solve this CI failure related to checker I am not familiar with it


also the purpose of this function is to  skip switch expressions when we `getDetailsForSwitch()` token why is that ? and should I do the same here in this PR https://github.com/checkstyle/checkstyle/blob/7e9960b8bfcf60f82a9236af846efe40883fa528/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java#L462-L472

and please review changes if all is good

nrmancuso:
>@nrmancuso @rnveach can you guide me on how to solve this CI failure related to checker I am not familiar with it

Let's wait until we are getting close to merge to deal with this, code might change by then :) But what Checker is telling us is that basically it expects the DetailAST to be non-null, and you are passing in a potentially null object instead.

>also the purpose of this function is to skip switch expressions when we getDetailsForSwitch() token why is that ? and should I do the same here in this PR

Ah, I had forgotten about this. I do not totally understand this "limitation", but since we have an open issue on it, let's also avoid switch expressions here. You can use `git blame` on this method to find the associated PR and read more if you are interested.

So, the answer to 

>should I do the same here in this PR

is yes. Make sure we have some in our inputs to verify this behavior.

Looking good so far!

nrmancuso:
@mahfouz72 I am going to place this PR into draft mode while you keep working :)

mahfouz72:
Github, generate report

mahfouz72:
Github, generate site

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

nrmancuso (commented):
@mahfouz72 let's try to make Checker happy and place `@Nullable` annotation on `Details` constructor arguments `lcurly` and `rcurly`.

mahfouz72:
Github, generate report

mahfouz72:
> If this doesn't help, we can just update suppression file.

this works but produced new survivals

```
    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java</fileName>
    <specifier>assignment</specifier>
    <message>incompatible types in assignment.</message>
    <lineContent>this.lcurly = lcurly;</lineContent>
    <details>
    </details>
  </checkerFrameworkError>

    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java</fileName>
    <specifier>assignment</specifier>
    <message>incompatible types in assignment.</message>
    <lineContent>this.rcurly = rcurly;</lineContent>
    <details>
    </details>
  </checkerFrameworkError>
```

mahfouz72:
@nrmancuso checker is falling even after adding suppression. Check PR description for report and site

mahfouz72:
@nrmancuso checker is falling even after adding suppression. Please Check PR description for report and site

nrmancuso:
> @nrmancuso checker is falling even after adding suppression build successbut process exit with code 1 idk why. also Please Check PR description for report and site

Did you download the checker patch and use `git apply`?  Please try to make checker happy while I find time to review, I need to dedicate some time on looking through all test cases. Thank you very much for making this easy to review by generating reports and site ahead of time and placing them in the PR description!

mahfouz72:
> Did you download the checker patch

@nrmancuso no can you please guide me on how to do this or send a doc for this

nrmancuso:
> > Did you download the checker patch
> 
> @nrmancuso no can you please guide me on how to do this or send a doc for this

Download artifact from link in checker execution (https://github.com/checkstyle/checkstyle/actions/runs/8220570380/artifacts/1312622476) unzip, and do `git apply …` where … the the name of the unzipped file

mahfouz72:
@nrmancuso thank you checker is happy now. Waiting for your review. TYT

src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheckTest.java
```diff
@@ -827,4 +827,150 @@ public void testSwitchExpression7() throws Exception {
         verifyWithInlineConfigParser(
                 getNonCompilablePath("InputRightCurlyTestSwitchExpression7.java"), expected);
     }
+
+    @Test
+    public void testCaseBlocksInSwitchStatement1() throws Exception {
```
nrmancuso, line 832: Please name test inputs/methods in some way to make it obvious what our motivation is for each test. Also, `Test` in the file name is redundant and makes these file names longer than they need to be. The next developer to come to this file should be able to easily look at the test method name and know what it's intention is. We can draw inspiration from the code under test, the configuration, some bug, etc. For instance, this method could be called `testCaseBlocksInSwitchStatementAlone` to give us an obvious hint as to what the configuration is.

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -460,6 +464,36 @@ private static Details getDetailsForSwitch(DetailAST switchNode) {
...
+                    lcurly = ast.getFirstChild();
+                }
+                else {
+                    lcurly = null;
+                }
+            }
+
+            if (isSwitchExpression(caseParent)) {
+                lcurly = null;
+            }
```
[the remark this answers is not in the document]
nrmancuso: After you do this, please generate another check regression test report

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -155,6 +155,7 @@ public int[] getAcceptableTokens() {
             TokenTypes.RECORD_DEF,
             TokenTypes.COMPACT_CTOR_DEF,
             TokenTypes.LITERAL_SWITCH,
+            TokenTypes.LITERAL_CASE,
```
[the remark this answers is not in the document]
mahfouz72: @nrmancuso idk but I remember that I read a comment from a maintainer in (issue/PR) that the defaultTokens should be limited. also there are a lot of default tokens in lcurly and not default here

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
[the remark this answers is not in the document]
mahfouz72: @nrmancuso done

src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheckTest.java
mahfouz72: @nrmancuso done

mahfouz72:
Github, generate report

mahfouz72:
Github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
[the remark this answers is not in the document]
mahfouz72: report: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/0a64bbd_2024044920/reports/diff/index.html 
I expected violation in more projects only two projects make this kind of violation we and jdk17 😄

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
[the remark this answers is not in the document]
mahfouz72: report: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/0a64bbd_2024044920/reports/diff/index.html 
I expected violation in more projects only two projects make this kind of violation we and jdk17 😄

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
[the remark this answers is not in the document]
nrmancuso: > @nrmancuso idk but I remember that I read a comment from a maintainer in (issue/PR) that the defaultTokens should be limited. also there are a lot of default tokens in lcurly and not default here

Please always share links to important points like this, I think I remember things all the time only to discover I am delusional 😂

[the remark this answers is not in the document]
mahfouz72: okay let me think where it was 😄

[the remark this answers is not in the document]
mahfouz72: @nrmancuso https://github.com/checkstyle/checkstyle/pull/12640#issuecomment-1434108894

mahfouz72:
@nrmancuso Spotbugs reported this
`[ERROR] Medium: Method com.puppycrawl.tools.checkstyle.checks.blocks.RightCurlyCheck$Details.getDetailsForCase(DetailAST) uses immediate execution of a block of code that is often not used [com.puppycrawl.tools.checkstyle.checks.blocks.RightCurlyCheck$Details] At RightCurlyCheck.java:[line 487] OI_OPTIONAL_ISSUES_USES_IMMEDIATE_EXECUTION`

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -460,6 +465,34 @@ private static Details getDetailsForSwitch(DetailAST switchNode) {
...
+            final Optional<DetailAST> switchRuleLcurly = Optional.ofNullable(ast)
+                    .filter(node -> parentType == TokenTypes.SWITCH_RULE);
+
+            final DetailAST lcurly = caseGroupLcurly.orElse(switchRuleLcurly.orElse(null));
+
+            DetailAST rcurly = null;
+            if (lcurly != null && !isSwitchExpression(caseParent)) {
+                rcurly = lcurly.getLastChild();
+            }
+            return new Details(lcurly, rcurly, getNextToken(caseParent), true);
```
nrmancuso, line 493: There is a few ways we can do this, but we can clean this up a bit more and do the same for rcurly:
```suggestion
          final DetailAST statementList = caseParent.findFirstToken(TokenTypes.SLIST);

            final Optional<DetailAST> lcurly;
            if (parentType == TokenTypes.SWITCH_RULE) {
                lcurly = Optional.ofNullable(statementList);
            }
            else {
                lcurly = Optional.ofNullable(statementList)
                        .map(DetailAST::getFirstChild)
                        .filter(node -> node.getType() == TokenTypes.SLIST);
                        
            }

            final DetailAST rcurly = lcurly.map(DetailAST::getLastChild)
                    .filter(child -> !isSwitchExpression(caseParent))
                    .orElse(null);
            
            return new Details(lcurly.orElse(null), rcurly, getNextToken(caseParent), true);
```

mahfouz72:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
mahfouz72: @nrmancuso very good done

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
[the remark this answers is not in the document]
nrmancuso: Ok, thanks for that! I've added that link to the original issue report. I thought I remembered the same, but can't always trust myself :)

nrmancuso:
> @nrmancuso Spotbugs reported this I am not sure why this is reported and how to solve it `[ERROR] Medium: Method com.puppycrawl.tools.checkstyle.checks.blocks.RightCurlyCheck$Details.getDetailsForCase(DetailAST) uses immediate execution of a block of code that is often not used [com.puppycrawl.tools.checkstyle.checks.blocks.RightCurlyCheck$Details] At RightCurlyCheck.java:[line 487] OI_OPTIONAL_ISSUES_USES_IMMEDIATE_EXECUTION`

@mahfouz72 For future reference, `Optional#orElse(...)` ALWAYS executes the `...` code. `Optional#orElseGet(...)` only executes the `...` code when the optional is not present (that's why it takes a `Supplier`). For most of our uses, it's not a big deal; however, in real life, you could imagine that our code might be like:

```java

MyResource r = Optional.ofNullable(getDatabaseObject())
                                   .orElseGet(() -> httpClient.fetchMyResource()));

```

obviously, we don't want unnecessary I/O (http calls) to fetch resources.

nrmancuso (approved):
Good stuff!

romani (approved):
ok to merge

mahfouz72:
@rnveach ping

src/main/resources/google_checks.xml
```diff
@@ -96,7 +96,7 @@
       <property name="tokens"
                value="CLASS_DEF, METHOD_DEF, CTOR_DEF, LITERAL_FOR, LITERAL_WHILE, STATIC_INIT,
                     INSTANCE_INIT, ANNOTATION_DEF, ENUM_DEF, INTERFACE_DEF, RECORD_DEF,
-                    COMPACT_CTOR_DEF, LITERAL_SWITCH"/>
+                    COMPACT_CTOR_DEF, LITERAL_SWITCH, LITERAL_CASE"/>
```
rnveach, line 99: Show proof from google's style guide that this is the correct setting for cases.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/Example5.java
```diff
@@ -41,5 +40,6 @@ public void method() {
         x = 1; }
     // violation above, 'should be alone on a line.'
   }
+
```
rnveach, line 43: Please remove unnecessary change.

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
```diff
@@ -460,6 +465,34 @@ private static Details getDetailsForSwitch(DetailAST switchNode) {
...
+            final Optional<DetailAST> lcurly;
+
+            if (parentType == TokenTypes.SWITCH_RULE) {
+                lcurly = Optional.ofNullable(statementList);
+            }
+            else {
+                lcurly = Optional.ofNullable(statementList)
+                        .map(DetailAST::getFirstChild)
+                        .filter(node -> node.getType() == TokenTypes.SLIST);
+
```
rnveach, line 487: Remove unnecessary empty line.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementALoneOrSingleline.java
```diff
@@ -0,0 +1,111 @@
...
+            case 2:{
+                    int x = 2;
+                    break;
+            } }      // violation '}' at column 13 should be alone on a line'
+    }
+
+    public static void test4() {
+        int mode = 0;
+        switch (mode) { case 0: {int x = 1;} break; default : int x = 5; }
+        // violation above '}' at column 44 should be alone on a line'
```
rnveach, line 79: Why is this not ok? It is a single line.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementALoneOrSingleline.java
```diff
@@ -0,0 +1,111 @@
...
+        switch (mode) { case 0: {int x = 1;}     // ok, single line
+            break; default : break; }
+    }
+
+    public static void test8() {
+        int mode = 0;
+        switch (mode) {
+            case 0: int x = 1;
+            case 1: x = 1; break;
+            case 2: {x = 1; break;} // ok, single line
```
rnveach, line 106: I am not seeing this so please add or point to where it is. This should be done for all the options, both regular switches and expression switches.

Cases where there are statements before, after, or both the `{}`s.

Example:
{x = 1;} break;
x = 1; { break; }
Etc...

Some weird cases where `{}` is on lines with expression but not on the same line as the other `{}`.

Example:
{
 x = 1; }
{ x = 1;
}
{ x =
1; }

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/Example3.java
```diff
@@ -16,10 +16,9 @@ class Example3 {
 
   public void method0() {
     int mode = 0;
+    int x;
     switch (mode) {
-      case 1:
-        int x = 1;
-        break;
+      case 1: {x = 1;}   // violation '}' at column 22 should be alone on a line'
```
rnveach, line 21: Keep original case. Make this one `int y = 1;`

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/Example5.java
```diff
@@ -16,10 +16,9 @@ class Example5 {
 
   public void method0() {
     int mode = 0;
+    int x;
     switch (mode) {
-      case 1:
-        int x = 1;
-        break;
+      case 1: {x = 1;}   // ok, RightCurly is in single line
```
rnveach, line 21: Same.

rnveach (changes requested):

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementALoneOrSingleline.java
rnveach: Also casing in the file name for `ALone` is not correct.

src/main/resources/google_checks.xml
mahfouz72: I added it in favor of  [rule 4.1.2](https://google.github.io/styleguide/javaguide.html#s4.1.2-blocks-k-r-style) and I found nothing says that this rule should not be applied for cases

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/Example5.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/RightCurlyCheck.java
mahfouz72: done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementALoneOrSingleline.java
mahfouz72: done thank you for pointing this

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementALoneOrSingleline.java
mahfouz72: > Cases where there are statements before, after, or both the `{}`s.

done add some case like this but in switch statement only  it won't compile in switch rule all statements after `->` has to be enclosed in one block



> Some weird cases where `{}` is on lines with expression but not on the same line as the other `{}`.

add cases like this also

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/Example3.java
mahfouz72: done

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/Example5.java
mahfouz72: done

src/main/resources/google_checks.xml
nrmancuso: If Google style guide is slowing down acceptance of this PR, I am good to make an issue to able it separately @rnveach

mahfouz72:
Github, generate report

mahfouz72:
Github, generate site

mahfouz72:
@rnveach ping

src/main/resources/google_checks.xml
rnveach: I am good here. I question if us forcing `alone` violates https://google.github.io/styleguide/javaguide.html#s4.1.3-braces-empty-blocks if it is empty, but this would be more than just this 1 token impacted.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksWithSwitchExpressionAloneOrSingleline.java
```diff
@@ -0,0 +1,93 @@
...
+    void test4() throws Exception {
+           int a = 0;
+           int b = switch (a) {
+                case 0 -> {
+                    int x = 5;
+                    int y = 6;
+                    if (a == 2) {
+                        y = 7;}   // violation '}' at column 31 should be alone on a line'
+                    throw new Exception();
+                }  default -> 2;
```
rnveach, line 65: Please confirm if this token validates this type of line. I noticed there is a violation on another file in a similar manner, but there isn't one here as it is not alone or single line.

https://github.com/checkstyle/checkstyle/pull/14623/files#diff-01f1a1b1234b46ef06ab0849a66bdb1efd5d8cdb080408432bd308213423a598R32

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementAlone2.java
```diff
@@ -0,0 +1,119 @@
...
+
+    public static void test15() {
+        int mode = 0;
+        switch (mode) {case 0: { } case 1: {  } } // 2 violations
+    }
+
+    public static void test16() {
+        int mode = 0;
+        switch (mode) {
+            case 0: int x = 1; { } break;  // violation '}' at column 34 should be alone on a line'
```
rnveach, line 81: So are we violating anywhere in a case that has a `{}`? It doesn't have to start the case? Middle and End is violated too?

@romani @nrmancuso Please confirm if this is acceptable as the issue only shows `{}` that start the cases.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksWithSwitchExpressionAloneOrSingleline.java
mahfouz72: > but there isn't one here as it is not alone or single line.

there is no violation here because this is switch expression and we are supposed to ignore it . the other example is normal switch rule

rnveach: So `case` includes switch statements (using `:` and `->`) but not switch expressions where the switch is used in said expression?

mahfouz72: as per my understanding In a switch expression, rather than just executing code blocks, the switch itself produces a value, which can then be assigned to a variable or directly used in an expression.
ex:
```
sout(switch(...) )
int a = switch () {...}
return switch (){....}
```
if switch doesn't produce a value to be used and it just executes code blocks then it is switch statement or rule
Check [JEP](https://openjdk.org/jeps/361) for more info

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementAlone2.java
mahfouz72: @rnveach Yes now it is violating the rcurly anywhere should it be at the start only? this was the case before I didn't even think of this syntax I was working on as we either put one block with braces `case x: {.....} `
or don't put braces at all `case x:.. `
but after your comment above I reconsidered that we can have the braces in the middle of the block
Please tell me what is the appropriate action here should I just leave this example but no violation?

rnveach: I leave this to the issue approvers I pinged. This seems like a weird case where we pick up all braces inside a case. For a similar train of thought, does this mean if we configure this check for "METHOD_DEF", then all solo braces inside a method itself should also be handled by this token. It is not an accurate analogy since they would be braces inside braces, but also technically the braces in the case tokens are not even connected to the case as shown in the AST. They are just a pure `SLIST`.

@mahfouz72 Can you add other test cases to show how your changes will fair against braces inside braces. Example: `case X: { { int x = 0; } }`. Are the inners validated as well?

rnveach: When I see examples using braces on `->` for cases, I agree that we should only be doing it when the braces surround the entire case (including any break). I am assuming it is invalid syntax to have anything after the braces in this case like `case X -> { int x = 0; } break;`. I don't use this new syntax, so I am less familiar with it.

I still await confirmation from others on non `->` but I am leaning towards having it be handled by `->` even if `->` can't be written like that.

rnveach: @mahfouz72 LeftCurly ( https://checkstyle.org/checks/blocks/leftcurly.html#LeftCurly ) supports `case`s. Please check if there is any conflicts on how you handle `}` versus how it handles `{`. If both line up, I am more inclined to approve the implementation.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementSame2.java
```diff
@@ -0,0 +1,119 @@
...
+
+            } case 1: {  }         // violation '}' at column 13 should be alone on a line'
+            default : {break;}
+        }
+    }
+
+    public static void test15() {
+        int mode = 0;
+        switch (mode) {case 0: { } case 1: {  } }
+        // violation above '}' at column 47 should be alone on a line'
```
rnveach, line 78: https://checkstyle.org/property_types.html#RightCurlyOption
> same
> Works like alone_or_singleline but .... If no next part of a multi-block statement present, brace must be alone on line. It also allows single-line format of multi-block statements. 

Why isn't this acceptable since it is all on a single line? I see other single lines being allowed, so I am confused.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementAlone2.java
nrmancuso: These blocks (`{}`) should not be checked. You can pretty much put blocks anywhere in Java.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementSame2.java
nrmancuso: >If no next part of a multi-block statement present, brace must be alone on line

I would not consider this a "multi block statement".

nrmancuso: I would consider it analogous to this case:
```
➜  src cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <module name="TreeWalker">
    <module name="RightCurly">
      <property name="option" value="same"/>
    </module>
  </module>
</module>                                                                                                

➜  src cat Test.java 
public class Test {

        if (args.length > 0) {}  if (args.length > 0) {}
    }

}
                                                                                                              
➜  src javac Test.java 

➜  src java -jar checkstyle-10.13.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:4:31: '}' at column 31 should be alone on a line. [RightCurly]
Audit done.
```

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementAlone2.java
mahfouz72: > Please check if there is any conflicts on how you handle `}` versus how it handles `{`

https://github.com/checkstyle/checkstyle/blob/f4b71587d76f15661e97c571a6183cb28d73b9eb/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/LeftCurlyCheck.java#L301-L314

the left curly is checking SLIST if it is the first child only



> `case X: { { int x = 0; } }`. Are the inners validated as well

no,  the inners are not validated. the `SLIST` that is a direct child of the main `SLIST` of the case only is validated but this is a nested SLIST (grandchild) and not checked



> but I am leaning towards having it be handled like `->`

I agree. to be in sync with leftcurly and with `->` I see it is better to check the braces only when it is the first child of the case any other inner braces should be ignored as it is not directly connected to the case

rnveach: > the left curly is checking SLIST if it is the first child only

Then I think this is how this check's implementation should be.

This issue started because Left check supports more tokens than Right check. If they are going to support the same tokens, they should validate the same cases. It doesn't make sense for one to support more or less than the other. It will have to be a new issue if they support case that we feel should not be supported, and vice versa.

mahfouz72: > Then I think this is how this check's implementation should be

yes I pushed update to validate onlynthe rculy that is a first child in case

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementSame2.java
rnveach: So I assume we are saying this works exactly as `alone_or_singleline` and `}` must always end the line? If so, documentation follows that.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksWithSwitchRuleAloneOrSingleline.java
```diff
@@ -0,0 +1,104 @@
...
+                 int x = 1;
+            }
+            case 2 -> {
+                    int x = 2;
+            } }      // violation '}' at column 13 should be alone on a line'
+    }
+
+    public static void test4() {
+        int mode = 0;
+        switch (mode) { case 0 -> {int x = 1;} default -> {int x = 0;} }
```
rnveach, line 82: Continuing from https://github.com/checkstyle/checkstyle/pull/14623#discussion_r1555180807, shouldn't this be a violation since it doesn't end the line?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementAlone2.java
rnveach: > the left curly is checking SLIST if it is the first child only

Just confirming, so as long as SLIST is the first child we will validate it even if there is more which comes after the `}`? It could be more code, break, or just the next case/default, etc...?

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/Example3.java
```diff
@@ -16,10 +16,12 @@ class Example3 {
 
   public void method0() {
     int mode = 0;
+    int x;
     switch (mode) {
       case 1:
-        int x = 1;
+        int y = 1;
         break;
+      case 2: {x = 1;}   // violation '}' at column 22 should be alone on a line'
```
rnveach, line 24: Add another example with code before the `{}` to show no violation.

rnveach (changes requested):

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/Example3.java
rnveach: I am leaning to having some verbiage explain how {}s in case works.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksWithSwitchRuleAloneOrSingleline.java
mahfouz72: the thing is both `LITERAL_CASE` and `LITERAL_DEFAULT` are both under `SWITCH_RULE` token
```
            |   |--SWITCH_RULE -> SWITCH_RULE [7:10]     <----
            |   |   |--LITERAL_CASE -> case [7:10]
            |   |   |   `--EXPR -> EXPR [7:15]
            |   |   |       `--NUM_INT -> 0 [7:15]
            |   |   |--LAMBDA -> -> [7:17]
            |   |   `--SLIST -> { [7:20]
            |   |       `--RCURLY -> } [7:21]
            |   |--SWITCH_RULE -> SWITCH_RULE [7:23]      <-----
            |   |   |--LITERAL_DEFAULT -> default [7:23]
            |   |   |--LAMBDA -> -> [7:31]
            |   |   `--SLIST -> { [7:34]
            |   |       `--RCURLY -> } [7:35]

```
So, this example is the same as  ` switch (mode) { case 0 -> {int x = 1;} case 1 -> {int x = 0;} }`  and this is ok because it is a single line format of a block in this example `}` shouldn't be at the end of the line.

This should be violation if the nextToken is not `SWITCH_RULE` example  `switch (mode) { case 0 -> {int x = 1;} int y;`
here will be a violation and `}` should be at the end of the line

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementAlone2.java
mahfouz72: @rnveach Yes exactly

src/test/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksInSwitchStatementSame2.java
mahfouz72: > So I assume we are saying this works exactly as `alone_or_singleline` and `}` must always end the line?

Yes but why always? if we have multiple single line blocks format then it is ok to not be at the end of the line 
example from if statements:
```
PS D:\test> cat src/Test3.java
public class Test3 {
  public static void test() {
      int mode = 0;
      if(mode == 0) {} else if(mode == 1) { } else {}
  }
}
PS D:\test> java  -jar checkstyle-10.14.2-all.jar -c config.xml  src/Test3.java
Starting audit...
Audit done.
```

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/Example3.java
mahfouz72: @rnveach done and added  explanation in a comment beside the example

mahfouz72:
Github, generate site

mahfouz72:
Github, generate report

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/InputRightCurlyCaseBlocksWithSwitchRuleAloneOrSingleline.java
mahfouz72: Yes I was missing that we should not consider this as a multiblock statement So it should be violation
Done

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/blocks/rightcurly/Example3.java
mahfouz72: @rnveach ping

mahfouz72:
@rnveach ping

rnveach:
I see production code changes after https://github.com/checkstyle/checkstyle/pull/14623#issuecomment-2042857667 , so we need a new report.

mahfouz72:
@rnveach I will generate new one but It was just a removal of a word typed by mistake in the test method so this is the final report https://github.com/checkstyle/checkstyle/pull/14623#issuecomment-2042857667 you can check this

mahfouz72:
Github, generate report

rnveach:
I think I am good with the PR. I want to review the latest report first.

mahfouz72:
okay and to ease review for you. I reviewed it first and all lgtm expect [this](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a77fb5d_2024140817/reports/diff/checkstyle/index.html#A9)  I am not sure about it. the thing is that according to AST both `LITERAL_CASE` and `LITERAL_DEFAULT` are under the same parent (`CASE_GROUP`) and the `{ }` are for both of them 

```
        |       |   |--CASE_GROUP -> CASE_GROUP [6:14]
        |       |   |   |--LITERAL_CASE -> case [6:14]
        |       |   |   |   |--EXPR -> EXPR [6:19]
        |       |   |   |   |   `--NUM_INT -> 1 [6:19]
        |       |   |   |   `--COLON -> : [6:21]
        |       |   |   |--LITERAL_DEFAULT -> default [7:14]
        |       |   |   |   `--COLON -> : [7:21]
        |       |   |   `--SLIST -> SLIST [7:23]
        |       |   |       `--SLIST -> { [7:23]
        |       |   |           `--RCURLY -> } [8:14]

```

rnveach:
> expect [this](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/da58ea6_2024154338/reports/diff/checkstyle/index.html#A13) I am not sure about it

How does LeftCurly handle this case if only `LITERAL_CASE` token is added? We are basing all the logic here off of duplicating Left's logic just for Rights.

mahfouz72:
Github, generate report

rnveach:
Please let me know what to expect from https://github.com/checkstyle/checkstyle/pull/14623#issuecomment-2057418330 .

mahfouz72:
leftCurlyCheck handles it differently. the braces in our example are not considered with case they are considered with the default only and this is correct IMO, So I made a small change to handle it as leftCurlyCheck does

for leftCurlyCheck when `LITERAL_CASE` only is added
```
case 0:
default: 
{               // no violation 
} ; 
```
so we should expect now after the last change. that [this](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/da58ea6_2024154338/reports/diff/checkstyle/index.html#A13) should have no violation

mahfouz72:
@rnveach report is good. The violation was removed from https://github.com/checkstyle/checkstyle/pull/14623#issuecomment-2053676049 and there are no differences between the last report and the previous one rather than this case. I think we are good to go now

rnveach:
@mahfouz72 Did you regenerate all 3 reports in https://github.com/checkstyle/checkstyle/pull/14623#issue-2174208475 ? I would want to see all 3 since this is a new token and it has many different options.

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

mahfouz72:
@rnveach please check https://github.com/checkstyle/checkstyle/pull/14623#issue-2174208475  reports for all 3 options has been regenerated and all lgtm
I updated PR description it has the final reports now

rnveach (approved):

PR #14635, opened 2024-03-10
Issue #13038: handled inner class method definition in VariableDeclarationUsageDistanceCheck

jxr98:
Issue: #13038 

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
@@ -415,6 +415,9 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
                             getFirstNodeInsideTryCatchFinallyBlocks(blockWithVariableUsage,
                                 variable);
                         break;
+                    case TokenTypes.METHOD_DEF:
+                        exprWithVariableUsage = blockWithVariableUsage.getNextSibling();
+                        break;
                     default:
                         exprWithVariableUsage = blockWithVariableUsage.getFirstChild();
                 }
@@ -784,7 +787,8 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
                 || type == TokenTypes.MODIFIERS
                 || type == TokenTypes.RESOURCE
                 || type == TokenTypes.EXTENDS_CLAUSE
-                || type == TokenTypes.IMPLEMENTS_CLAUSE;
+                || type == TokenTypes.IMPLEMENTS_CLAUSE
+                || type == TokenTypes.CLASS_DEF;
     }
 
 }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral4.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral4.java
@@ -22,8 +22,7 @@ public class InputVariableDeclarationUsageDistanceGeneral4 {
         <T> void xx(List<T> m){}
     }
     public void method9() {
-        // until https://github.com/checkstyle/checkstyle/issues/13011
-        Integer z = 5; // violation 'Distance .* is 3.'
+        Integer z = 5;
         Iterator<Integer> mn = new HashSet<Integer>().iterator();
         class BClass extends Parent {
             Boolean h = false;
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
@@ -352,9 +352,7 @@ public class VariableDeclarationUsageDistanceCheckTest extends
 
     @Test
     public void testGeneralClass4() throws Exception {
-        final String[] expected = {
-            "26:9: " + getCheckMessage(MSG_KEY, "z", 3, 1),
-        };
+        final String[] expected = {};
 
         verifyWithInlineConfigParser(
                 getPath("InputVariableDeclarationUsageDistanceGeneral4.java"), expected);
@@ -407,5 +405,14 @@ public class VariableDeclarationUsageDistanceCheckTest extends
         };
         verifyWithInlineConfigParser(
                 getPath("InputVariableDeclarationUsageDistanceCloseToBlock.java"), expected);
+    }
+
+    @Test
+    public void testVariableDeclarationUsageDistanceMethodDefinition() throws Exception {
+        final String[] expected = {
+            "18:9: " + getCheckMessage(MSG_KEY, "c", 2, 1),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputVariableDeclarationUsageDistanceMethodDefinition.java"), expected);
     }
 }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceMethodDefinition.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceMethodDefinition.java
@@ -1,0 +1,30 @@
+/*
+VariableDeclarationUsageDistance
+allowedDistance = 1
+ignoreVariablePattern = (default)
+validateBetweenScopes = true
+ignoreFinal = false
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.variabledeclarationusagedistance;
+
+import java.util.Iterator;
+public class InputVariableDeclarationUsageDistanceMethodDefinition {
+    public void method(){
+        int a=1;
+        int b=2;
+        int c=3;// violation 'Distance .* is 2.'
+        class GeneralLogic {
+            public <T> Iterator<T> method(T[] b) throws Exception{
+                if(a>0){
+                    System.out.println(b.length);
+                }
+                return null;
+            }
+        };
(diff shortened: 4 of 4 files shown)


jxr98:
GitHub, generate report

jxr98:
I've read the pitest wiki and the pitest mutation overview. I understand `removed conditional - replaced equality check with true` mutation, but don't see how it applies to the [code](https://github.com/jxr98/checkstyle/blob/49133d8cc3e38f3d843cf0c3dd918bd9a2f5314e/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L784-L794). I have tried to replace the first `==` with equal and remove it. It doesn't seem right. Can anyone help me? Thanks.

nrmancuso:
> I've read the pitest wiki and the pitest mutation overview. I understand `removed conditional - replaced equality check with true` mutation, but don't see how it applies to the [code](https://github.com/jxr98/checkstyle/blob/49133d8cc3e38f3d843cf0c3dd918bd9a2f5314e/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L784-L794). I have tried to replace the first `==` with equal and remove it. It doesn't seem right. Can anyone help me? Thanks.
> 
> image

I am fairly certain this means that now the ONLY tokens that are passed to this method are in these conditions, so we might as well return true from the isZeroDistanceToken method, which means that where ever this method is used could probably be mutated to just `true` as well. Try to remove this method completely and see what happens.

jxr98:
> > I've read the pitest wiki and the pitest mutation overview. I understand `removed conditional - replaced equality check with true` mutation, but don't see how it applies to the [code](https://github.com/jxr98/checkstyle/blob/49133d8cc3e38f3d843cf0c3dd918bd9a2f5314e/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L784-L794). I have tried to replace the first `==` with equal and remove it. It doesn't seem right. Can anyone help me? Thanks.
> > image
> 
> I am fairly certain this means that now the ONLY tokens that are passed to this method are in these conditions, so we might as well return true from the isZeroDistanceToken method, which means that where ever this method is used could probably be mutated to just `true` as well. Try to remove this method completely and see what happens.

I've tried to return true or false in the isZeroDistanceToken method and remove it completely, but some of the tests we have failed after that.

romani:
as a hack, please make method as public and write junit test for this method to clearly show that survival can gone if we do have explicit test.

jxr98:
> as a hack, please make method as public and write junit test for this method to clearly show that survival can gone if we do have explicit test.

Thanks. The method is very straightforward containing several or operator as follows 
image, and the code is [here](https://github.com/jxr98/checkstyle/blob/49133d8cc3e38f3d843cf0c3dd918bd9a2f5314e/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java#L784-L794). From debugging and experience, the survivals are `Expr`, `if`, `for`, `return`, `{`, etc. The conditions in the method can't be complete I guess. 🤔

nrmancuso:
> but some of the tests we have failed after that. Still can't figure out why

If this is true, then we are missing tests from the pitest profile in pom.xml for this check. Please push up your changes and show this failure so that we can help debug.

jxr98:
> > but some of the tests we have failed after that. Still can't figure out why
> 
> If this is true, then we are missing tests from the pitest profile in pom.xml for this check. Please push up your changes and show this failure so that we can help debug.

I've pushed the changes I made. If the method just return true directly, the failed tests are shown as in the file below. 
[com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheckTest.txt](https://github.com/checkstyle/checkstyle/files/14843742/com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheckTest.txt)

It looks like the pom file includes the target class and target test of this check correctly. I'd really appreciate it if you could help me debug. I guess I might make some mistakes where I'm not aware of right now.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral4.java
```diff
@@ -22,8 +22,7 @@ void mm() {
         <T> void xx(List<T> m){}
     }
     public void method9() {
-        // until https://github.com/checkstyle/checkstyle/issues/13011
```
rnveach, line 25: First I would confirm that these 2 issues are exactly the same and explain it so we understand. I find it odd that a method issue would fix a class issue, but maybe no one noticed the cases were extremely similar.

Also, apparently you reported both issues. :)

nrmancuso:
> It looks like the pom file includes the target class and target test of this check correctly. I'd really appreciate it if you could help me debug. I guess I might make some mistakes where I'm not aware of right now.

There is definitely something weird going on here. I've checked this code out locally, and tried to hardcode the mutation in a number of ways, but the tests are always failing. The only thing I can think of is that the bytecode inlines this method and the actual mutation isn't what is shown in the suppression. 

I suggest that you check the bytecode for this file, you can see how to do this at https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports#decompiler-usage-to-see-bytecode.

@Vyom-Yadav have you seen anything like this before?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -784,7 +787,8 @@ private static boolean isZeroDistanceToken(int type) {
                 || type == TokenTypes.MODIFIERS
                 || type == TokenTypes.RESOURCE
                 || type == TokenTypes.EXTENDS_CLAUSE
-                || type == TokenTypes.IMPLEMENTS_CLAUSE;
+                || type == TokenTypes.IMPLEMENTS_CLAUSE
+                || type == TokenTypes.CLASS_DEF;
```
rnveach, line 791: > removed conditional - replaced equality check with true

This is my local when I run the profile.
Image

I removed one of these lines completely, or something similar to changing `|| type == TokenTypes.EXTENDS_CLAUSE` to `|| false`, and all the tests passed. This shows the mutation. The mutation in this case is reversed from the text that says true.

So now we must identify how to kill these mutations, or understand why they can't be killed. I would look over https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports#how-to-find-java-code-that-can-kill-mutation and https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#checkstyle-pitest-regression for the next steps. You can also look over other pitest issues that describe the more complex process. We do not normally accept new code with new surviving mutations.

nrmancuso: >At a glance this sort of makes sense. extends and implements both have class as a parent.

Yes, this sounds like what is happening here. We should probably add some test cases for local enums, records, and interfaces as well.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral4.java
jxr98: Oo right they r similar.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
jxr98: Thank u so much!!! I see. Trying to solve it.

romani:
@jxr98, please find time to finish PR

jxr98:
Okay sry a bit busy recently. Will finish soon.

romani:
@MohanadKh03 , please help us here .
We appreciate your help.

MohanadKh03:
@romani Sorry for the late reply , Sure 
It seems the issue itself is fixed and the only thing remaining is resolving the mutation right ?

romani:
We don't know :)

romani:
@MohanadKh03, I hope you are doing well. 
Please let me know I can answer any your questions.
We appreciate your help on this PR.

MohanadKh03:
@romani Hey thanks for the ping I am doing fine just got a bit busy
ill be able to go through this PR next week or sth but Ill see if there is another issue/PR hanging that wouldnt need much pitest resolution such as this one and go through it

romani:
Thanks a lot for response, if restore PR, I can help with pitest coverage ideas.
Together we will be able to finish it.

MohanadKh03:
So after rebasing locally and doing some investigations it seems that the issue is fixed but the mutations are caused due to these conditions 
```java
|| type == TokenTypes.IMPLEMENTS
|| type == TokenTypes.EXTENDS
|| type == TokenTypes.CLASS_DEF
```
so as far as I can see what is causing the problem is that if the token type is a class then it 100% is a class therefore there will always be a mutation in these conditions
as @nrmancuso mentioned in an earlier comment we probably need to add tests for some inner classes , records and hopefully the mutations will be killed but I do wonder do we need to add another condition for records ? maybe idk

Ill try all this out in a new PR as an extension of this one alright ? @romani Also if there is something I am missing about my conclusion please let me know if there is sth else that is probably causing this mutation

nrmancuso:
@MohanadKh03 thanks for your analysis :) I find that when working on Checkstyle, test driven development is a great way to approach most issues. Take the time to craft some interesting input files and tests, and let them guide your changes.

romani:
Please create PR, we will discuss all in it.
We might be even luck to find case by regression.

romani:
>wonder do we need to add another condition for records ? maybe idk

Just drop it (record ) to Input file and lets see how it works.

There more tests the better for project.

romani:
Energy is moved to https://github.com/checkstyle/checkstyle/pull/16026

Issue #14631, opened 2024-03-09
update JavadocTokenTypes.java to new format of AST print


romani:
Same as https://github.com/checkstyle/checkstyle/issues/14516 but for javadoc 

https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.java

https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.html

https://checkstyle.org/cmdline.html , see -j or -J .

```
SteLeo@DESKTOP-11C7U5R MINGW64 ~/exception_literal test
$ cat Test.java
* @exception SQLException if query is not correct
SteLeo@DESKTOP-11C7U5R MINGW64 ~/exception_literal test
$ java -jar checkstyle-10.16.0-all.jar -j Test.java | sed "s/\[[0-9]\+:[0-9]\+\]//g"
JAVADOC -> JAVADOC
|--LEADING_ASTERISK -> *
|--WS ->
|--JAVADOC_TAG -> JAVADOC_TAG
|   |--EXCEPTION_LITERAL -> @exception
|   |--WS ->
|   |--CLASS_NAME -> SQLException
|   |--WS ->
|   `--DESCRIPTION -> DESCRIPTION
|       `--TEXT -> if query is not correct
`--EOF -> <EOF>
```
Good example on how update should looks like is at https://github.com/checkstyle/checkstyle/pull/14862

To generate AST, use cmd :  `java -jar checkstyle-10.13.0-all.jar -J Test.java | sed "s/\[[0-9]\+:[0-9]\+\]//g"`
The sed command removes the line:column numbers for our documentation.

Each token update should be in its own Pull Request.
Share whole  cli output for javadoc chunk in PR description.

~ATTENTION: do not forget to escape `>` to `&gt`. And also escape `<` if appeared.~

**Watch videos on how to do new PR and do exactly this update , link to youtube video is at this page** https://checkstyle.org/beginning_development.html#Starting_Development 

Good example of update: https://github.com/checkstyle/checkstyle/pull/14632 please provide all details as provided in description of this PR
One more good example of code changes: https://github.com/checkstyle/checkstyle/pull/16086/files

Documentation should be improved for:
- [x] AREA_HTML_TAG_NAME  
- [x] AREA_TAG  
- [x] ARGUMENT  
- [ ] ATTR_VALUE  
- [x] ATTRIBUTE 
[...]

PR #14636, opened 2024-03-10
Issue #14631: Updated RETURN_LITERAL of JavadocTokenTypes.java to new format of AST print

SteLeo1602:
Issue: #14631 

**Updated:**
image

**Old:**
return_literal

**Test.java code used**
```
public class Test {
    /**
     * Some java doc.
     *
     * @return true if file exists
     * @param fileExists file exists
     */
    public boolean fileVerify(boolean fileExists) {
        return fileExists;
    }
}
```
**Run command**
`java -jar checkstyle-10.14.0-all.jar -J Test.java | sed 's/->/-\&gt;/g; s/\[[0-9]\+:[0-9]\+\]//g'`

**Output**
```
COMPILATION_UNIT -&gt; COMPILATION_UNIT
`--CLASS_DEF -&gt; CLASS_DEF
    |--MODIFIERS -&gt; MODIFIERS
    |   `--LITERAL_PUBLIC -&gt; public
    |--LITERAL_CLASS -&gt; class
    |--IDENT -&gt; Test
    `--OBJBLOCK -&gt; OBJBLOCK
        |--LCURLY -&gt; {
        |--METHOD_DEF -&gt; METHOD_DEF
        |   |--MODIFIERS -&gt; MODIFIERS
        |   |   |--BLOCK_COMMENT_BEGIN -&gt; /*
        |   |   |   |--COMMENT_CONTENT -&gt; *\r\n     * Some java doc.\r\n     *\r\n     * @return true if file exists\r\n     * @param fileExists file exists\r\n
        |   |   |   |   `--JAVADOC -&gt; JAVADOC
        |   |   |   |       |--NEWLINE -&gt; \r\n
        |   |   |   |       |--LEADING_ASTERISK -&gt;      *
        |   |   |   |       |--TEXT -&gt;  Some java doc.
        |   |   |   |       |--NEWLINE -&gt; \r\n
        |   |   |   |       |--LEADING_ASTERISK -&gt;      *
        |   |   |   |       |--NEWLINE -&gt; \r\n
        |   |   |   |       |--LEADING_ASTERISK -&gt;      *
        |   |   |   |       |--WS -&gt;
        |   |   |   |       |--JAVADOC_TAG -&gt; JAVADOC_TAG
        |   |   |   |       |   |--RETURN_LITERAL -&gt; @return
        |   |   |   |       |   |--WS -&gt;
        |   |   |   |       |   `--DESCRIPTION -&gt; DESCRIPTION
        |   |   |   |       |       |--TEXT -&gt; true if file exists
        |   |   |   |       |       `--NEWLINE -&gt; \r\n
        |   |   |   |       |--LEADING_ASTERISK -&gt;      *
        |   |   |   |       |--WS -&gt;
        |   |   |   |       |--JAVADOC_TAG -&gt; JAVADOC_TAG
        |   |   |   |       |   |--PARAM_LITERAL -&gt; @param
        |   |   |   |       |   |--WS -&gt;
        |   |   |   |       |   |--PARAMETER_NAME -&gt; fileExists
        |   |   |   |       |   |--WS -&gt;
        |   |   |   |       |   `--DESCRIPTION -&gt; DESCRIPTION
        |   |   |   |       |       |--TEXT -&gt; file exists
        |   |   |   |       |       |--NEWLINE -&gt; \r\n
        |   |   |   |       |       `--TEXT -&gt;
        |   |   |   |       `--EOF -&gt; <EOF>
        |   |   |   `--BLOCK_COMMENT_END -&gt; */
        |   |   `--LITERAL_PUBLIC -&gt; public
        |   |--TYPE -&gt; TYPE
        |   |   `--LITERAL_BOOLEAN -&gt; boolean
        |   |--IDENT -&gt; fileVerify
        |   |--LPAREN -&gt; (
        |   |--PARAMETERS -&gt; PARAMETERS
        |   |   `--PARAMETER_DEF -&gt; PARAMETER_DEF
        |   |       |--MODIFIERS -&gt; MODIFIERS
        |   |       |--TYPE -&gt; TYPE
        |   |       |   `--LITERAL_BOOLEAN -&gt; boolean
        |   |       `--IDENT -&gt; fileExists
        |   |--RPAREN -&gt; )
        |   `--SLIST -&gt; {
        |       |--LITERAL_RETURN -&gt; return
        |       |   |--EXPR -&gt; EXPR
        |       |   |   `--IDENT -&gt; fileExists
        |       |   `--SEMI -&gt; ;
        |       `--RCURLY -&gt; }
        `--RCURLY -&gt; }
```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.java
@@ -43,11 +43,11 @@ public final class JavadocTokenTypes {
      * <pre>{@code @return true if file exists}</pre>
      * <b>Tree:</b>
      * <pre>{@code
-     *   |--JAVADOC_TAG[4x3] : [@return true if file exists]
-     *       |--RETURN_LITERAL[4x3] : [@return]
-     *       |--WS[4x10] : [ ]
-     *       |--DESCRIPTION[4x11] : [true if file exists]
-     *           |--TEXT[4x11] : [true if file exists]
+     * JAVADOC_TAG -> JAVADOC_TAG
+     *  |--RETURN_LITERAL -> @return
+     *  |--WS ->
+     *  `--DESCRIPTION -> DESCRIPTION
+     *      |--TEXT -> true if file exists
      * }</pre>
      *
      * @see


src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.java
```diff
@@ -43,11 +43,13 @@ public final class JavadocTokenTypes {
      * <pre>{@code @return true if file exists}</pre>
      * <b>Tree:</b>
      * <pre>{@code
-     *   |--JAVADOC_TAG[4x3] : [@return true if file exists]
-     *       |--RETURN_LITERAL[4x3] : [@return]
-     *       |--WS[4x10] : [ ]
-     *       |--DESCRIPTION[4x11] : [true if file exists]
-     *           |--TEXT[4x11] : [true if file exists]
+     * JAVADOC_TAG -&gt; JAVADOC_TAG
```
romani, line 46: `-&gt;` in html is weird, it should be `->`.
Please investigate why we needed html escaping in Tokentypes.java but don't need here.

romani (changes requested):
Your image shows a problem, thanks for sharing:

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.java
```diff
@@ -43,11 +43,13 @@ public final class JavadocTokenTypes {
...
+     *    |--RETURN_LITERAL -&gt; @return
+     *    |--WS -&gt;
+     *    `--DESCRIPTION -&gt; DESCRIPTION
+     *        |--TEXT -&gt; true if file exists
+     *        |--NEWLINE -&gt; \r\n
+     *        `--TEXT -&gt;
      * }</pre>
```
romani, line 53: This is weird AST node. It might appear there only due to fact that it is last javadoc tag.
Please try to add one more tag after `return`, for example `see`. And share ast.
If no such node, please use AST without such node.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.java
MANISH-K-07: Yes indeed, this is because the target javadoc tag is last of the block.
However, we can simply discard these lines and copy until line 48. I think adding another tag will not make much difference as we are trimming the AST for only the part we require ( return_literal )

The only part we need would be : 
```
JAVADOC_TAG -&gt; JAVADOC_TAG
  |--RETURN_LITERAL -&gt; @return
  |--WS -&gt;
```

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.java
MANISH-K-07: @romani ,
I believe this is because we use a paragraph code block inside javadoc in JavadocTokenTypes.java ( `@code {` )
Whereas, in TokenTypes.java , the paragraph tag is used plainly and any '>' or '<' or even '@' are treated as HTML tags

From TokenTypes.java : 
```
     * <p>For example:</p>
     * <pre>
     * public class MyClass {
     *     private int foo;
     *     {foo = 10;}
     * }
     * </pre>
```

From JavadocTokenTypes.java : 
```
     * <p><b>Example:</b></p>
     * <pre>{@code @return true if file exists}</pre>
     * <b>Tree:</b>
     * <pre>{@code
     *   |--JAVADOC_TAG[4x3] : [@return true if file exists]
     *       |--RETURN_LITERAL[4x3] : [@return]
     *       |--WS[4x10] : [ ]
     *       |--DESCRIPTION[4x11] : [true if file exists]
     *           |--TEXT[4x11] : [true if file exists]
     * }</pre>
```

So yepp, the `@code` is what makes the difference here. It basically tells the paragraph of javadoc that the code inside this block is source code and should not be considered as Html tags

In case my explanation isn't good, maybe related  :  https://stackoverflow.com/questions/19302183/what-does-the-code-java-annotation-do#:~:text=%7B%40code%20...%7D%20is,differently%20than%20the%20other%20text.

SteLeo1602: Thanks

SteLeo1602: Done

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.java
SteLeo1602: Done, swapped them so return is not the last tag.

MANISH-K-07: @SteLeo1602 , do you see any diff from the previous change, except like I said, one last line of AST trimmed ?

SteLeo1602: I see, ``--NEWLINE ->; \r\n` seems to unrelate

MANISH-K-07: Yes, by convention, all javadoc tags begin at a new line. So, I don't see why we would require adding that to each and every example AST. Due to this, I believe dropping these would increase readability and decrease confusion among users.
Let's see if @romani agrees with me or not ;)

MANISH-K-07:
Response to https://github.com/checkstyle/checkstyle/pull/14636#discussion_r1518726253 : 

> `-&gt;` in html is weird, it should be `->`.
> Please investigate why we needed html escaping in Tokentypes.java but don't need here.

@romani , I have posted the possible explanation at https://github.com/checkstyle/checkstyle/pull/14636#discussion_r1518730602

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.java
```diff
@@ -43,11 +43,11 @@ public final class JavadocTokenTypes {
      * <pre>{@code @return true if file exists}</pre>
      * <b>Tree:</b>
      * <pre>{@code
-     *   |--JAVADOC_TAG[4x3] : [@return true if file exists]
-     *       |--RETURN_LITERAL[4x3] : [@return]
-     *       |--WS[4x10] : [ ]
-     *       |--DESCRIPTION[4x11] : [true if file exists]
-     *           |--TEXT[4x11] : [true if file exists]
+     * JAVADOC_TAG ->; JAVADOC_TAG
```
romani, line 46: Remove `;`

It was part of html escaping symbol.

romani (changes requested):
Items

romani (approved):
Ok to merge

PR #14639, opened 2024-03-10
Issue #6207: Add Xpath regression test for LocalFinalVariableName

Lmh-java:
Part of issue #6207

### DIFF
--- a/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionLocalFinalVariableNameTest.java
+++ b/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionLocalFinalVariableNameTest.java
@@ -1,0 +1,115 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code and other text files for adherence to a set of rules.
+// Copyright (C) 2001-2024 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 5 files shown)


MANISH-K-07 (commented):
@Lmh-java ,
I think it would be better if we cover one input for Ctors and class fields

Lmh-java:
Sure, I can do Ctors, but can you elaborate more on class fields? I don't think we need class fields here, since this is the check for final local variables inside a method (correct me if I am wrong)

MANISH-K-07:
Thanks for the reply.. otherwise I wouldn't have gone through the check now ;)
Please hold on until we get a bit of clarity....

@romani ,
Could you please tell me why LocalFinalVariableNameCheck is only testing for three tokens (basically just parameters and try-with-resources block).
I believe this should apply to final variables in Ctors as well? Infact, it should apply to any block in general ??
Please correct me in case my understanding of this check is wrong 😅

Lmh-java:
I took a close look at the checker. Here is a screenshot of the class-level comment.
image
So it is only checking `VARIABLE_DEF`, `PARAMETER_DEF`, and 'RESOURCE'.
The checker also re-writes `mustCheckName(DetailAST)` to only focus on local variables. The method detects variables declared in "a code block, a for initializer, or a catch parameter".

Therefore, I think I might need to include extra tests for:
- [ ] final variables declaration inside a `for` initializer.
- [ ] parameters of a method that are marked as final.
And these tests would have different Xpath structures.

Hi @MANISH-K-07 , what's your opinion about this?

MANISH-K-07:
> Hi @MANISH-K-07 , what's your opinion about this?

Your post is basically a duplicate of my question at https://github.com/checkstyle/checkstyle/pull/14639#issuecomment-1987340360.  😂
You are partially right though, That is what the description says but I'm not sure what the Check is actually testing. I will run a few test cases as soon as I get on my laptop, and get back with answers !!

> @romani ,
> Could you please tell me why LocalFinalVariableNameCheck is only testing for three tokens (basically just parameters and try-with-resources block).
> I believe this should apply to final variables in Ctors as well? Infact, it should apply to any block in general ??
> Please correct me in case my understanding of this check is wrong 😅

Meanwhile, Let's wait for @romani to have a look at my query at https://github.com/checkstyle/checkstyle/pull/14639#issuecomment-1987340360
;)

romani (approved):
Ok to merge.

This Check only for local variables , not parameters.
We should have separate Check to cover parameters

PR #14397, opened 2024-01-31
Issue #14376: Removed Parent POM & Updated maven-release-plugin Config to fix Broken mvn-Enforcer Config

MANISH-K-07:
Closes  #14376 

@romani  , @nrmancuso 

In order to remove the parent pom usage, we needed to apply the parent pom's plugin config. See https://github.com/sonatype/oss-parents/blob/master/codehaus-parent/pom.xml for this configuration.

Without updating this configuration, we get the following error:
 
```
[ERROR] Project does not define required minimum version of Maven.
[ERROR] Update the pom.xml to contain maven-enforcer-plugin to
[ERROR] force the Maven version which is needed to build this project.
```

We also had failures in the [release-dry-run workflow](https://app.circleci.com/pipelines/github/checkstyle/checkstyle/23415/workflows/1abd9b3b-bf74-424b-afd6-cf50631e7dd2/jobs/497715?invite=true#step-103-22861_57):

```
launch (Launcher.java:226)
    at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode (Launcher.java:407)
    at org.codehaus.plexus.classworlds.launcher.Launcher.main (Launcher.java:348)
Caused by: org.apache.maven.plugin.PluginContainerException: An API incompatibility was encountered while executing org.apache.maven.plugins:maven-release-plugin:2.2.1:prepare: java.lang.NoSuchMethodError: 'org.apache.maven.settings.RuntimeInfo org.apache.maven.settings.Settings.getRuntimeInfo()'
```

org.sonatype.oss oss-parent 9 : 
```
  <!-- This parent is requirements for deploy of artifacts to maven central -->
  <parent>
    <groupId>org.sonatype.oss</groupId>
    <artifactId>oss-parent</artifactId>
    <version>9</version>
  </parent>
```

PR #14642, opened 2024-03-11
Issue #14376: Added JAR generation config in maven-source to replace OSS parent pom

### LINKED ISSUE #14376: broken maven-enforcer configuration
The POM has 2 different maven-enforcer configurations, because the parent (sonatype-oss) and the checkstyle pom.xml use different execution ids.

This is from Sonatype parent pom.xml:
```
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-enforcer-plugin</artifactId>
                <version>1.2</version>
                <executions>
                    <execution>
                        <id>enforce-maven</id>
                        <goals>
                            <goal>enforce</goal>
                        </goals>
                        <configuration>
                            <rules>
                                <requireMavenVersion>
                                    <version>(,2.1.0),(2.1.0,2.2.0),(2.2.0,)</version>
                                    <message>Maven 2.1.0 and 2.2.0 produce incorrect GPG signatures and checksums respectively.</message>
                                </requireMavenVersion>
                            </rules>
                        </configuration>
                    </execution>
                </executions>
```

From checkstyle pom.xml: https://github.com/checkstyle/checkstyle/blob/50683f10e9d33649dd8d1c42fe0549c003cedccd/pom.xml#L1352-L1356

Note the **2 different ids**. This is problematic for instance when executing the versions plugin:
```
mvn versions:display-plugin-updates
```
which *wrongly* complains about no maven version being defined:
```
[ERROR] Project does not define required minimum version of Maven.
[ERROR] Update the pom.xml to contain maven-enforcer-plugin to
[ERROR] force the Maven version which is needed to build this project.
```
That's because it uses the first value it finds in the effective pom, which is the one from the Sonatype parent.

To fix this, I would recommend binding the execution of the Sonatype defined id `enforce-maven` against phase `none` (to effectively disable it). Alternatively the checkstyle pom execution id could be changed to `enforce-maven` to overwrite the parent configuration.

UPDATE: If I get it right, that parent POM is deprecated, so just removing it would of course also fix this issue: https://github.com/sonatype/oss-parents

MANISH-K-07:
Aims to close #14376 

@romani , @nrmancuso , This PR is a Follow-up of #14397 in response to https://github.com/checkstyle/checkstyle/issues/14376#issuecomment-1969038627

In order to remove the parent pom usage, we needed to apply the parent pom's plugin config. See https://github.com/sonatype/oss-parents/blob/master/codehaus-parent/pom.xml for this configuration.

We also had problems with release :
https://github.com/checkstyle/checkstyle/actions/runs/8066068513/job/22033791002#step:6:399
```
Error: ROR]   Rule "javadoc-staging" failures
Error: ROR]     * Missing: no javadoc jar found in folder '/com/puppycrawl/tools/checkstyle/10.14.0'
```
After digging in, I found that we had a dummy maven-source-plugin for build helpin our pom which was fine until we used oss parent. I believe this is the root cause for previous failures.
```
      <!-- Generate test sources as well as main sources. -->
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-source-plugin</artifactId>
      </plugin>
```
For JAR generation, I have introduced related configs to the plugins to execute.

https://maven.apache.org/plugins/maven-source-plugin/
Specified goals : `jar` and `test-jar` , `jar-no-fork`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,13 +3,6 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
-  <!-- This parent is requirements for deploy of artifacts to maven central -->
-  <parent>
-    <groupId>org.sonatype.oss</groupId>
-    <artifactId>oss-parent</artifactId>
-    <version>9</version>
-  </parent>
-
   <groupId>com.puppycrawl.tools</groupId>
   <artifactId>checkstyle</artifactId>
   <version>10.14.1-SNAPSHOT</version>
@@ -487,9 +480,11 @@
         </plugin>
         <plugin>
           <artifactId>maven-release-plugin</artifactId>
-          <!-- version is same as in supper-pom as it is better to use same version
-               as in sonatype-nexus-staging -->
           <version>2.1</version>
+          <configuration>
+            <mavenExecutorId>forked-path</mavenExecutorId>
+            <useReleaseProfile>false</useReleaseProfile>
+          </configuration>
         </plugin>
         <!-- from super-pom END -->
         <plugin>
@@ -550,6 +545,14 @@
               </tag>
             </tags>
           </configuration>
+          <executions>
+            <execution>
+              <id>attach-javadocs</id>
+              <goals>
+                <goal>jar</goal>
+              </goals>
+            </execution>
+          </executions>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
@@ -697,8 +700,7 @@
             <execution>
               <id>attach-sources</id>
               <goals>
-                <goal>test-jar</goal>
-                <goal>jar</goal>
+                <goal>jar-no-fork</goal>
               </goals>
             </execution>
           </executions>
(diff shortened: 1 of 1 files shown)


romani:
Please recheck why dry run of release phase did not catch fact of no javadoc.

We have it in CI somewhere.
If you run dry run, do you see javadoc zip, can you share list of jar/zip we have.

Here is list of files we need https://repo1.maven.org/maven2/com/puppycrawl/tools/checkstyle/10.14.1/

MANISH-K-07:
> Please recheck why dry run of release phase did not catch fact of no javadoc.

@romani , I think that might have something to do with parameter `failOnWarnings` unspecified for `maven-javadoc-plugin`

> If you run dry run, do you see javadoc zip, can you share list of jar/zip we have.

Unfortunately, no. The plugin config seems to be running but is not bundling jar.
I have added all related configs from parent but still am missing something here....

romani:
>This issue seems to be killing a lot of my productive hours 😞

abandon this PR for now, we can come back to it at any point of time in future, in future you will be more experienced and you will handle it.

MANISH-K-07:
> > This issue seems to be killing a lot of my productive hours 😞
> 
> abandon this PR for now, we can come back to it at any point of time in future, in future you will be more experienced and you will handle it.

Sure, I will temporarily put this on hold until Gsoc or until some new thought comes to my mind :)
Please don't close this PR, I don't want this issue to be gone unnoticed or lost in time. An open (draft) PR will keep reminding me that I've left a pending issue.

nrmancuso:
If we are abandoning this, please close it, Github keeps a reference to this code forever.

romani:
@MANISH-K-07 , please find time to finish this PR

romani:
Everyone is welcome to reuse this code and make new PR with good fix

PR #14652, opened 2024-03-12
Issue #3110: Bugfix in Checker.java removed English hardcoded msgs & Added new method to extract non-locale translations

MANISH-K-07:
Part of #3110 

The following methods required message keys : 

- processFiles (Exception msg)
- setupChild
- setCharset

Checker.java file now has only two more hardcoded strings left to completely be compatible with non-locale translations : 

1. destroy
2. processFiles (Error msg)

For maximum code optimization, I've designed a new method  `getLocalizedMessage` that takes in a String arg `messageKey` and returns another String `localizedMessage` that is extracted from _.properties_ files instead of returning _LocalizedMessage_ object, to make things even more simpler.

Thanks a lot @rdiachenko, for pointing this out coz, reused and optimized code is obviously the best !!

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
@@ -88,11 +88,6 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
     /** The audit event filters. */
     private final FilterSet filters = new FilterSet();
 
-    /** Message used by finishLocalSetup method. */
-    private final LocalizedMessage finishLocalSetupMsg = new LocalizedMessage(
-            Definitions.CHECKSTYLE_BUNDLE, getClass(),
-                    "Checker.finishLocalSetup");
-
     /** The basedir to strip off in file names. */
     private String basedir;
 
@@ -313,8 +308,8 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
                 }
 
                 // We need to catch all exceptions to put a reason failure (file name) in exception
-                throw new CheckstyleException("Exception was thrown while processing "
-                        + filePath, ex);
+                throw new CheckstyleException(
+                        getLocalizedMessage("Checker.processFilesException", filePath), ex);
             }
             catch (Error error) {
                 if (fileName != null && cacheFile != null) {
@@ -445,8 +440,7 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
 
         if (moduleFactory == null) {
             if (moduleClassLoader == null) {
-                final String finishLocalSetupMessage = finishLocalSetupMsg.getMessage();
-                throw new CheckstyleException(finishLocalSetupMessage);
+                throw new CheckstyleException(getLocalizedMessage("Checker.finishLocalSetup"));
             }
 
             final Set<String> packageNames = PackageNamesLoader
@@ -486,8 +480,8 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
             }
         }
         catch (final CheckstyleException ex) {
-            throw new CheckstyleException("cannot initialize module " + name
-                    + " - " + ex.getMessage(), ex);
+            throw new CheckstyleException(
+                    getLocalizedMessage("Checker.setupChildModule", name, ex.getMessage()), ex);
         }
         if (child instanceof FileSetCheck) {
             final FileSetCheck fsc = (FileSetCheck) child;
@@ -507,8 +501,8 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
             addListener(listener);
         }
         else {
-            throw new CheckstyleException(name
-                    + " is not allowed as a child in Checker");
+            throw new CheckstyleException(
+                    getLocalizedMessage("Checker.setupChildNotAllowed", name));
         }
     }
 
@@ -620,8 +614,8 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
     public void setCharset(String charset)
             throws UnsupportedEncodingException {
         if (!Charset.isSupported(charset)) {
-            final String message = "unsupported charset: '" + charset + "'";
-            throw new UnsupportedEncodingException(message);
+            throw new UnsupportedEncodingException(
+                    getLocalizedMessage("Checker.setCharset", charset));
         }
         this.charset = charset;
     }
@@ -651,6 +645,21 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
         if (cacheFile != null) {
             cacheFile.reset();
         }
+    }
+
+    /**
+     * Extracts localized messages from properties files.
+     *
+     * @param messageKey the key pointing to localized message in respective properties file.
+     * @param args the arguments of message in respective properties file.
+     * @return a string containing extracted localized message
+     */
+    private String getLocalizedMessage(String messageKey, Object... args) {
+        final LocalizedMessage localizedMessage = new LocalizedMessage(
+            Definitions.CHECKSTYLE_BUNDLE, getClass(),
+                    messageKey, args);
+
+        return localizedMessage.getMessage();
     }
 
 }
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
@@ -1,4 +1,8 @@
 Checker.finishLocalSetup=if no custom moduleFactory is set, moduleClassLoader must be specified
+Checker.processFilesException=Exception was thrown while processing {0}
+Checker.setCharset=unsupported charset: ''{0}''
+Checker.setupChildModule=cannot initialize module {0} - {1}
+Checker.setupChildNotAllowed={0} is not allowed as a child in Checker
 DefaultLogger.addException=Error auditing {0}
 DefaultLogger.auditFinished=Audit done.
 DefaultLogger.auditStarted=Starting audit...
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/messages_ru.properties
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/messages_ru.properties
@@ -1,5 +1,9 @@
 Checker.finishLocalSetup=если нет обычая moduleFactory установлен, \
   moduleClassLoader должно быть указано
+Checker.processFilesException=Исключение возникло при обработке {0}
+Checker.setCharset=неподдерживаемая кодировка: ''{0}''
+Checker.setupChildModule=не могу инициализировать модуль {0} - {1}
+Checker.setupChildNotAllowed={0} не разрешен в качестве вложенного модуля в Checker
 DefaultLogger.addException=Ошибка проверки {0}
 DefaultLogger.auditFinished=Проверка завершена.
 DefaultLogger.auditStarted=Начинаем проверку...
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/messages_zh.properties
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/messages_zh.properties
@@ -1,4 +1,8 @@
 Checker.finishLocalSetup=如果没有定制 moduleFactory 已设置, moduleClassLoader 必须指定
+Checker.processFilesException=处理时抛出异常 {0}
+Checker.setCharset=不支持的字符集: ''{0}''
+Checker.setupChildModule=无法初始化模块 {0} - {1}
+Checker.setupChildNotAllowed={0} 不允许作为Checker的子模块
 DefaultLogger.addException=检查错误： {0}
 DefaultLogger.auditFinished=检查完成。
 DefaultLogger.auditStarted=开始检查……
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -24,6 +24,17 @@
   <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/Checker.java</fileName>
     <specifier>argument</specifier>
+    <message>incompatible argument for parameter args of Checker.getLocalizedMessage.</message>
+    <lineContent>getLocalizedMessage(&quot;Checker.setupChildModule&quot;, name, ex.getMessage()), ex);</lineContent>
+    <details>
+      found   : @Initialized @Nullable String
+      required: @Initialized @NonNull Object
+    </details>
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/Checker.java</fileName>
+    <specifier>argument</specifier>
     <message>incompatible argument for parameter args of Violation constructor.</message>
     <lineContent>new String[] {ioe.getMessage()}, null, getClass(), null));</lineContent>
     <details>
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/messages_de.properties
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/messages_de.properties
@@ -1,5 +1,9 @@
 Checker.finishLocalSetup=wenn kein Brauch moduleFactory eingestellt ist, moduleClassLoader \
   muss angegeben werden
+Checker.processFilesException=Während der Verarbeitung wurde eine Ausnahme ausgelöst {0}
+Checker.setCharset=Nicht unterstützter Zeichensatz: ''{0}''
+Checker.setupChildModule=Modul kann nicht initialisiert werden {0} - {1}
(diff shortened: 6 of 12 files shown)


MANISH-K-07:
The only english hardcoded strings I have left in Checker.java are : 

https://github.com/checkstyle/checkstyle/blob/646f49995cad29f62d489e14ba910788fc953aa6/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L349-L360

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -65,6 +65,9 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
     /** Message to use when an exception occurs and should be printed as a violation. */
     public static final String EXCEPTION_MSG = "general.exception";
 
+    /** One single White space. */
+    public static final String SINGLE_WHITESPACE = " ";
```
rdiachenko, line 69: Why do we need it as a separate field? I'd suggest to use it in place.

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -88,11 +91,41 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
...
+
     /** Message used by finishLocalSetup method. */
     private final LocalizedMessage finishLocalSetupMsg = new LocalizedMessage(
             Definitions.CHECKSTYLE_BUNDLE, getClass(),
                     "Checker.finishLocalSetup");
 
+    /** Message used by processFiles method to show Error. */
+    private final LocalizedMessage processFilesErrorMsg = new LocalizedMessage(
+            Definitions.CHECKSTYLE_BUNDLE, getClass(),
+                    "Checker.processFilesError");
```
rdiachenko, line 107: That's a lot of repetitive code.

1. Could you try to single out common part into a method to end up with something like this: `private final LocalizedMessage processFilesErrorMsg = localizedMessage("Checker.processFilesError");`.
2. Do we actually need those fields? Why can't we use inplace values? Considering, they are used once in rare cases when exception is thrown, I think we a good to have them closer to the usage.

src/main/resources/com/puppycrawl/tools/checkstyle/messages_ru.properties
```diff
@@ -1,5 +1,11 @@
+Checker.destroy=Невозможно сохранить файл кэша.
 Checker.finishLocalSetup=если нет обычая moduleFactory установлен, \
   moduleClassLoader должно быть указано
+Checker.processFilesError=Во время обработки возникла ошибка
+Checker.processFilesException=Исключение возникло при обработке
+Checker.setCharset=неподдерживаемая кодировка
+Checker.setupChildModule=не могу инициализировать модуль
+Checker.setupChildNotAllowed=в детстве не допускается Checker
```
rdiachenko, line 8: Please update to: `не разрешен в качестве потомка в Checker`

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
MANISH-K-07: The string literal "` `" has been used at a total of 4 places in the file. Apparently, we have a check called _MultipleStringLiterals_ that allows no repetitions (by default max allowed = 1).

src/main/resources/com/puppycrawl/tools/checkstyle/messages_ru.properties
MANISH-K-07: Update done.
I have been using Google Translate to obtain the non-locale translations

MANISH-K-07:
@rdiachenko ,
I am not that familiar with pitests side of our project. ( More into development and unit testing than mutation hunting 😅 )
There are few surviving mutations in pitest-commons 
https://github.com/checkstyle/checkstyle/actions/runs/8274073597/job/22638881050?pr=14652#step:6:476
Could you please guide me on how to deal with these ?

Lmh-java:
@MANISH-K-07 I guess it is basically because you've introduced a new method (`getLocalizedMessage()`), but you don't have any tests for it. Therefore, some mutations for it in Pitest is not killed.

src/main/resources/com/puppycrawl/tools/checkstyle/messages_zh.properties
```diff
@@ -1,4 +1,10 @@
+Checker.destroy=无法保留缓存文件.
 Checker.finishLocalSetup=如果没有定制 moduleFactory 已设置, moduleClassLoader 必须指定
+Checker.processFilesError=处理时抛出错误
+Checker.processFilesException=处理时抛出异常
+Checker.setCharset=不支持的字符集
+Checker.setupChildModule=无法初始化模块
+Checker.setupChildNotAllowed=不允许作为儿童 Checker
```
Lmh-java, line 7: Hi, could you please update Checker.setupChildNotAllowed to `不允许作为Checker的子模块`. Since this would make more sense in Mandarin 😊

MANISH-K-07:
> @MANISH-K-07 I guess it is basically because you've introduced a new method (`getLocalizedMessage()`), but you don't have any tests for it. Therefore, some mutations in Pitest are not killed.

I actually had the same surviving mutations during first push itself i.e. before introducing the newly designed method. So, I don't think the Pitest failure is coz of the `getLocalizedMessage` method. I suspect the fault here being with call to getMessage().
Thanks for your opinion anyways :)

Also, regarding the Chinese msg change, I'll get it done as soon as I get on my laptop

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
rdiachenko: 1. Could you make it `private` since it is used within this class only?
2. Please rename it as `WHITESPACE_SEPARATOR`
3. Update comment as follows: `Separator to use in string concatination.`

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -190,7 +188,8 @@ public void destroy() {
                 cacheFile.persist();
             }
             catch (IOException ex) {
-                throw new IllegalStateException("Unable to persist cache file.", ex);
+                final String destroyMessage = getLocalizedMessage("Checker.destroy");
```
rdiachenko, line 191: Why do we need it as a separate variable? Can we use it inplace `throw new IllegalStateException(getLocalizedMessage("Checker.destroy"), ex);`?

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -653,4 +660,18 @@ public void clearCache() {
         }
     }
 
+    /**
+     * Extracts Localized messages from properties files.
```
rdiachenko, line 664: Please make it lowercase `Localized`

rdiachenko:
> @rdiachenko , I am not that familiar with pitests side of our project. ( More into development and unit testing than mutation hunting 😅 ) There are few surviving mutations in pitest-commons https://github.com/checkstyle/checkstyle/actions/runs/8274073597/job/22638881050?pr=14652#step:6:476 Could you please guide me on how to deal with these ?

Some (not all) of your changes led to the following surviving mutations:
```
+    <mutatedMethod>destroy</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
+    <description>removed call to com/puppycrawl/tools/checkstyle/Checker::getLocalizedMessage</description>
+    <lineContent>final String destroyMessage = getLocalizedMessage(&quot;Checker.destroy&quot;);</lineContent>
+  </mutation>
+
+    <mutatedMethod>destroy</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator</mutator>
+    <description>replaced call to com/puppycrawl/tools/checkstyle/Checker::getLocalizedMessage with argument</description>
+    <lineContent>final String destroyMessage = getLocalizedMessage(&quot;Checker.destroy&quot;);</lineContent>
+  </mutation>
+
+    <mutatedMethod>processFiles</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
+    <description>removed call to com/puppycrawl/tools/checkstyle/Checker::getLocalizedMessage</description>
+    <lineContent>final String processFilesErrorMessage = getLocalizedMessage(</lineContent>
+  </mutation>
+
+    <mutatedMethod>processFiles</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator</mutator>
+    <description>replaced call to com/puppycrawl/tools/checkstyle/Checker::getLocalizedMessage with argument</description>
+    <lineContent>final String processFilesErrorMessage = getLocalizedMessage(</lineContent>
+  </mutation>
```
It means, that there are no real tests for code you introduced. To kill survivors you need to add such tests to cover new logic with getting localized message. To check whether pitest passes on localhost against specific pitest profile you may run this: `.ci/pitest.sh pitest-common`.

If you are not familiar with pitest at all, I'd suggest to start with this resource https://pitest.org

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
MANISH-K-07: Done..
Actually, I left the field `public static` so that it can be called later when required somewhere else while I work on rest of the files of this issue. I though this way, we could avoid multiple private field declarations for every class :)

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
MANISH-K-07: Done..
Even more optimization, that's fantastic

MANISH-K-07:
@rdiachenko ,
I have browsed through https://pitest.org/ for the past few hours and now have a basic understanding of PIT.
Looks like we will need to cover tests that fail when call to getLocalizedMessage() is removed or replaced with arg for two places.

I will work on covering tests and send a follow-up PR as soon as this is merged. As this PR introduces a new method, I didn't wanna complicate things here :)

src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
```diff
@@ -1,4 +1,10 @@
+#Checker.destroy=Unable to persist cache file.
```
rdiachenko, line 1: Please remove all commented lines and add them as part of a new PR when you work on it.

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
rdiachenko: > later when required somewhere else

This "later" rarely comes in practice. Better and easier to apply restrictions now and loosen them later if needed than the other way around.

src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
MANISH-K-07: Done..
Archived on local and removed for remote branch.

rdiachenko (approved):
lgtm

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -68,6 +68,9 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
     /** The extension separator. */
     private static final String EXTENSION_SEPARATOR = ".";
 
+    /** Separator to use in string concatenation. */
+    private static final String WHITESPACE_SEPARATOR = " ";
```
nrmancuso, line 72:
```suggestion
    private static final String SINGLE_SPACE = " ";
```
We have no need to be ambiguous here. "Whitespace" could be multiple spaces, tabs, newlines, carriage returns, form feeds, vertical tabs, etc (or any combination thereof).

nrmancuso (changes requested):
Minor item:

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
MANISH-K-07: I have originally used var `SINGLE_WHITESPACE` but later changed it after review https://github.com/checkstyle/checkstyle/pull/14652#discussion_r1524076122.

Done now.... re-changed it after re-review 😄

MANISH-K-07: I have originally used var `SINGLE_WHITESPACE` but later changed it after review https://github.com/checkstyle/checkstyle/pull/14652#discussion_r1524076122.

Done now.... re-changed it after re-review 😄

src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
```diff
@@ -1,4 +1,8 @@
 Checker.finishLocalSetup=if no custom moduleFactory is set, moduleClassLoader must be specified
+Checker.processFilesException=Exception was thrown while processing
+Checker.setCharset=unsupported charset
+Checker.setupChildModule=cannot initialize module
+Checker.setupChildNotAllowed=is not allowed as a child in Checker
```
romani, line 5: All of them should have `{0}` to clearly define that some value will be there

romani (changes requested):
Major item:

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
romani: If we do https://github.com/checkstyle/checkstyle/pull/14652#discussion_r1526999404 this constant is not required at all

nrmancuso: Yep, please do this @MANISH-K-07

MANISH-K-07: Sure.. already on it :)

src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
MANISH-K-07: Done and rebased....

Hacked the new method `getLocalizedMessage` to take another argument _**Varargs**_. Also, modified the driver codes to take params. There weren't much major changes required here bcoz `getMessage` already calls format on args if varargs is not null, just like our getCheckMessage :)

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -486,8 +480,9 @@ protected void setupChild(Configuration childConf)
             }
         }
         catch (final CheckstyleException ex) {
-            throw new CheckstyleException("cannot initialize module " + name
-                    + " - " + ex.getMessage(), ex);
+            throw new CheckstyleException(
+                    getLocalizedMessage("Checker.setupChildModule", name)
+                            + " - " + ex.getMessage(), ex);
```
romani, line 485: All should be in message, nobody limit you to have `{1}`.
All should be in parameters of message. No any concatenation of content in code.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
MANISH-K-07: >  I was trying a rather different approach with one of the exceptions but there were conflicts

@romani , 
Actually, the change you suggested now is what I tried previously. I removed call to getMessage from hardcoded string and pointed it to .properties file as {1}
However, there were conflicts among two ci.
The `getLocalizedMessage` takes varargs as arguments, so we need to pass param Obj but the _ex.getMessage()_ returns a string (not a string object). So, without casting, we get a survivng error in `Checker / check (nullness-optional-interning) (pull_request)`
When we cast to type object (or even use string constructor invocation), we have violations with idea inspections saying the cast is redundant. I did not see a simpler way here to tackle both these ci other than pulling out ex.getMessage()
If there is any better way then I am all ears :)

romani: If static analysis tools blocks to make better code , we do suppression of them.

MANISH-K-07: I kept seeing in many other cases that the maintainers generally recommend not to extend surviving errors. So, I thought we won't be allowing it here too.

> If static analysis tools blocks to make better code , we do suppression of them.

One suppression coming right up....

romani: Checker is not fully enforced tool, I would rather sacrifice it, in favor all other, we will deal with it later on.

Main point that whole message must be in property file. All other are internal details that we can change at any time. Message is property file is kind of a contract to all users, it should not change without notifying them, user should be able to translate and change it if required.

MANISH-K-07: Please ignore this push. It is to obtain patch

romani (approved):
ok to merge

romani:
@rdiachenko , @nrmancuso , @rnveach , FYI, better to have Checker suppression extension than not good code.

Issue #6294, opened 2018-12-17
pitest: increase mutation coverage for indentation profile to 100%


rnveach:
We should increase coverage for indentation profile up to 100%.
This issue is a subtask of #3708

SGanguly1999:
I am on it.

rohit9625:
Is somebody working on this issue?

nrmancuso:
@rohit9625 no one is working on this issue, please be aware that this is most difficult pitest coverage issue we have. If you want to begin working on it, please kill one mutation per PR only.

sktpy:
Is there any update that I might need to be aware of(asking since it has been a while since the issue was created), else I can start working on this issue.

nrmancuso:
> Is there any thing that I might need to be aware of before I take this issue?(asking because it has been a while since this was created), else I can start working on it. 🚀

@sktpy 

1. ROI is low
2. Difficulty will be likely be high
3. Surviving mutations can be found at https://github.com/checkstyle/checkstyle/blob/72fa0536c643e56934cb7095061150ebf17a84e6/config/pitest-suppressions/pitest-indentation-suppressions.xml#L2

sktpy:
> ROI is low

If that is the case, will focus on the issues that will be more impactful for our end users, but will visit this one periodically so that I can meet and greet boss level mutants. :space_invader:

vivek-0509:
Hey @romani ,
While investigating PIT mutations in the indentation module, I encountered a suppressed mutation related to the call to primordialHandler.checkIndentation() in IndentationCheck.beginTree(). 
https://github.com/checkstyle/checkstyle/blob/aac12dfeb30a44cbc478826ecf1e400cb25216d1/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java#L357-L359
On closer inspection, this call serves no functional purpose because PrimordialHandler.checkIndentation() has an empty implementation (// nothing to check). As a result, the mutation is equivalent: removing the call produces no observable behavior change, which explains why it cannot be killed by tests.
https://github.com/checkstyle/checkstyle/blob/aac12dfeb30a44cbc478826ecf1e400cb25216d1/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/PrimordialHandler.java#L36-L40

[...]

PR #14664, opened 2024-03-15
Issue #6294: kill mutation in pitest-indentation profile

mahfouz72:
Issue #6294 

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/56ef607d74d50a56a43a242f8217dc30/raw/af6f018051816d9ca6f5edfd9d670bac3dd3f75f/indentation.xml

### DIFF
--- a/config/pitest-suppressions/pitest-indentation-suppressions.xml
+++ b/config/pitest-suppressions/pitest-indentation-suppressions.xml
@@ -1,13 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <mutation unstable="false">
-    <sourceFile>AbstractExpressionHandler.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.indentation.AbstractExpressionHandler</mutatedClass>
-    <mutatedMethod>findSubtreeAst</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.ConditionalsBoundaryMutator</mutator>
-    <description>changed conditional boundary</description>
-    <lineContent>if (colNum == null || thisLineColumn &lt; colNum) {</lineContent>
-  </mutation>
 
   <mutation unstable="false">
     <sourceFile>AbstractExpressionHandler.java</sourceFile>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
@@ -477,7 +477,7 @@ public abstract class AbstractExpressionHandler {
             final Integer colNum = astSet.getStartColumn(lineNum);
 
             final int thisLineColumn = expandedTabsColumnNo(tree);
-            if (colNum == null || thisLineColumn < colNum) {
+            if (colNum == null || thisLineColumn <= colNum) {
                 astSet.addAst(tree);
             }
 


mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

mahfouz72:
report generation failed with out of memory error how to solve this?

rnveach:
With indentation you may have to run on your local. I dunno if splitting up the config more will help. I have a VM for regression and I had to give it a lot of memory for indentation runs, like 4-8 gigs. https://cwiki.apache.org/confluence/display/MAVEN/OutOfMemoryError

mahfouz72:
Github, generate report

mahfouz72:
> With indentation you may have to run on your local.

got out of memory error even on local. I am trying @romani suggestion

mahfouz72:
Github, generate report

mahfouz72:
@rnveach @romani  there is a diff in the report and I used it as testcase but still this case didn't kill the mutation when I tried to run it on my local please educate me how is this possible

rnveach:
First, did you try to copy the file in its entirety and use that to see if you see the difference when running the test before trying to kill the mutation. Sometimes trimming the file can cause what is unique about the file to be removed. Mutation and test cases can get tricky in these kind of situations.

If you can't reconfirm the regression's findings with the file as it is, it could mean there is like a memory leak crossing over into the new file's run. We have had a few issues like that before.

mahfouz72:
> id you try to copy the file in its entirety (no trimming)

@rnveach I copied it without trimming, but when running the test after hardcoding the mutation refactor, the difference in the report doesn't appear.

rnveach:
Then either there is a leak between files (like an instance field not cleared on a new file), or there is an order issue (like a hash set returned a different order of a list because the ordinal hashes were different between runs), or something of the likes that would cause this type of issue. Either option is not very easy to track down.

mahfouz72:
unfortunately, I have no idea where to start I am not familiar with this process. how to debug regression? it would be helpful if you could provide some resources or references to old issues/PR that dealt with similar cases

romani:
It means mutation that you did for regression report is not mutation that pitest is doing to this code.

mahfouz72:
it was [conditionals_boundary](https://pitest.org/quickstart/mutators/#CONDITIONALS_BOUNDARY) mutator  `<` to `<=` and that is what I did

rnveach:
Regression on my local did not find the case.
https://rveach.no-ip.org/checkstyle/regression/343/
https://github.com/checkstyle/checkstyle/compare/master...rnveach:checkstyle:help_Issue_6294

mahfouz72:
yes, I also generated a report on my local and found no diff. So now how am I supposed to proceed on this what is the next step?

> I would show us pitest report with your mutated code

after hardcode the mutation pitest reported new survival which is the code before mutation
`<` mutated to `<=` and it survived now after hardcoding `<=` mutated to  `<` and obviously it survived too. So hardcoding mutation isn't a solution and we didn't find a case in regression that will kill this so how to deal with this case?

rnveach:
> So now how am I supposed to proceed on this what is the next step?

See https://github.com/checkstyle/checkstyle/issues/14019#issue-1988998627
> To understanding how to kill the mutation
Visit [Wiki page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports), [Pitest-docs](https://pitest.org/) and expected actions and workflow at https://github.com/checkstyle/checkstyle/issues/12341.

mahfouz72:
yes. we should now accept what pitest says apply the mutation in code and analyze why is not possible to kill the mutation by test. but what I say is that hardcoding mutation is producing new survivals so we don't solve anything here
`<=` mutated to `<` and `<` mutated to `<=`  it is something like circular survivals 😄 
Please correct me if I am missing something because I haven't faced this before

rnveach:
https://github.com/checkstyle/checkstyle/issues/12341
You are only looking at step 3 - 5. Step 6 has to be done. If analysis finds nothing then some PR will have to be provided with the mutation removed.

I would be hesitant to modify indentation without a concrete reason.

mahfouz72:
now I know why we haven't killed mutation in pitest profile before it seems tricky. I think that If I leave the indentation profile for now and start with another simpler profile to get enough experience and get back to indentation would be better.
please share your thoughts about this do you recommend that?

rnveach:
The process for pitest will not really differ between profiles. Even simplier classes can have complex situations. What would make this easy is understanding the full process, the class you are focusing on, and some luck finding regression cases or a logical mind to sift through the code to create a case or reason why it can't be killed.

PR #14673, opened 2024-03-17
Issue #14084: fix Checker violation for nonnegative integer argument

Lmh-java:
Part of Issue #14084 
Solve suppression for `DetailAstImpl` by introducing `NonNegative` annotation.

### DIFF
--- a/config/checker-framework-suppressions/checker-index-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-index-suppressions.xml
@@ -14,17 +14,6 @@
   <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java</fileName>
     <specifier>argument</specifier>
-    <message>incompatible argument for parameter bitIndex of BitSet.get.</message>
-    <lineContent>return getBranchTokenTypes().get(tokenType);</lineContent>
-    <details>
-      found   : int
-      required: @NonNegative int
-    </details>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java</fileName>
-    <specifier>argument</specifier>
     <message>incompatible argument for parameter bitIndex of BitSet.set.</message>
     <lineContent>branchTokenTypes.set(type);</lineContent>
     <details>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java
@@ -384,7 +384,7 @@ public final class DetailAstImpl implements DetailAST {
 
     @Override
     public boolean branchContains(int tokenType) {
-        return getBranchTokenTypes().get(tokenType);
+        return tokenType >= 1 && getBranchTokenTypes().get(tokenType);
     }
 
     @Override
--- a/src/test/java/com/puppycrawl/tools/checkstyle/DetailAstImplTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/DetailAstImplTest.java
@@ -355,8 +355,9 @@ public class DetailAstImplTest extends AbstractModuleTestSupport {
 
     @Test
     public void testBranchContains() {
-        final DetailAstImpl root = createToken(null, TokenTypes.CLASS_DEF);
-        final DetailAstImpl modifiers = createToken(root, TokenTypes.MODIFIERS);
+        final DetailAstImpl root = createToken(null, TokenTypes.COMPILATION_UNIT);
+        final DetailAstImpl classDef = createToken(root, TokenTypes.CLASS_DEF);
+        final DetailAstImpl modifiers = createToken(classDef, TokenTypes.MODIFIERS);
         createToken(modifiers, TokenTypes.LITERAL_PUBLIC);
 
         assertWithMessage("invalid result")
@@ -364,6 +365,12 @@ public class DetailAstImplTest extends AbstractModuleTestSupport {
                 .isTrue();
         assertWithMessage("invalid result")
                 .that(root.branchContains(TokenTypes.OBJBLOCK))
+                .isFalse();
+        assertWithMessage("invalid result")
+                .that(root.branchContains(TokenTypes.COMPILATION_UNIT))
+                .isTrue();
+        assertWithMessage("negative tokenType")
+                .that(root.branchContains(-1))
                 .isFalse();
     }
 


src/main/java/com/puppycrawl/tools/checkstyle/api/DetailAST.java
```diff
@@ -19,6 +19,8 @@
 
 package com.puppycrawl.tools.checkstyle.api;
 
+import org.checkerframework.checker.index.qual.NonNegative;
```
romani, line 22: this is apart I am not sure about Checker: we looks like have to change our api and force all our user to depend on Checker, that is not good.
We recently resolved big pain of dependency to guava, that is not api based dependency.
Do we have any other way resolve this ?
if not I would rather disable validation of Checker for cases that require `NonNegative`.
Can we remove it from `api` class only ?

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/api/DetailAST.java
MANISH-K-07: @romani ,
Sorry to involve.. But why are we not using  **jsr-305** annotations here ?
I think we have a Nonnegative annotation in javax. 
https://www.javadoc.io/doc/com.google.code.findbugs/jsr305/latest/javax/annotation/Nonnegative.html

@Lmh-java , could you please explain why you have introduced checker's annotations here? 
I understand why there are no ci troubles here. Checker framework treats annotations from other tools as if they were extracted from corresponding annotation in Checker itself. 
But by doing this, `we will only be increasing our JAR.`

Lmh-java: checkerframework does not support `Nonnegative` from javax. I was trying to use it first until I figured out it does not recognize that annotation. Therefore, I turned to use `NonNegative` from the checker framework.

MANISH-K-07: Yes, looks that way , I was just browsing through checker release docs : https://checkerframework.org/releases/2.3.1/api/index.html?javax/annotation/package-summary.html

> Do we have any other way resolve this ?

@romani , I'm afraid we don't have any alternative to migrate **jsr 305** other than checker framework.
Most of the projects are removing their javax annotations rather than using checker imports.
Jenkins for example : https://groups.google.com/g/jenkinsci-dev/c/uE1wwtVi1W0

Lmh-java: Yea, same here. I've just looked at Jenkin's discussion posts, but I will try to solve this problem by throwing an `UncheckException`. Not very sure if that works, but I will test it out.

romani: >It seems checkerframework does not support Nonnegative from javax.

Can we check if there is issue registed for it on checker ? If not, let's create.
If there is defect in checker, I would rather postpone dealing with such violations. There are tons of others, no need to be stuck with problematic.

MANISH-K-07: The Nonnegative annotation is just one of many..
These are all the annotations in package javax.annotations (jsr 305) : 
https://www.javadoc.io/doc/com.google.code.findbugs/jsr305/latest/javax/annotation/package-summary.html

And checker is only supporting three of them : 
https://checkerframework.org/releases/2.3.1/api/index.html?javax/annotation/package-summary.html

Lmh-java: > And checker is only supporting three of them   

Yea, I have seen that as well. 

> Can we check if there is issue registed for it on checker ? If not, let's create.

I have opened an issue to their repo already and I am waiting for their response. Meanwhile, I will be dealing with other violations. We can postpone this violation for now.

Lmh-java: I can solve this problem by introducing a throw of `IllegalArgumentException` as an unchecked exception. But I still feel it's not elegant and not necessary,

romani (changes requested):
Items

config/checker-framework-suppressions/checker-index-suppressions.xml
```diff
@@ -11,17 +11,6 @@
     </details>
   </checkerFrameworkError>
 
-  <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java</fileName>
-    <specifier>argument</specifier>
-    <message>incompatible argument for parameter bitIndex of BitSet.get.</message>
-    <lineContent>return getBranchTokenTypes().get(tokenType);</lineContent>
```
romani, line 18: This is details of implementation that we use bitset.  But we are changing API method. It is not good.
API is generic, and if somebody wants to put -1 in our method, they have full right in this, we should not throw exception, and just return false.

romani (changes requested):
Items

config/checker-framework-suppressions/checker-index-suppressions.xml
Lmh-java: Done

Lmh-java:
@romani, the checker framework decides to add the `@Nonnegative` from `javax` as an alias to their annotation in the future update. I will keep track of that so we can bump the version of the checker framework in the future to better solve other problems.
The issue I opened: https://github.com/typetools/checker-framework/issues/6507

MANISH-K-07:
> @romani, the checker framework may add the `@Nonnegative` from `javax` as an alias to their annotation in the future update. I will keep track of that so we can bump the version of the checker framework in the future to better solve other suppressed violations. The issue I opened: [typetools/checker-framework#6507](https://github.com/typetools/checker-framework/issues/6507)

They had majorly introduced the nullness annotations in release [2.1.14](https://github.com/typetools/checker-framework/releases/tag/checker-framework-2.1.14). Unfortunately, this is outdated by 8 solid years :)
It would be great if they add the update coz we don't have any other option to jsr 305, but I'm not sure if this will be happening anytime soon, given their update history 😅

src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java
```diff
@@ -384,7 +384,7 @@ private BitSet getBranchTokenTypes() {
 
     @Override
     public boolean branchContains(int tokenType) {
-        return getBranchTokenTypes().get(tokenType);
+        return tokenType >= 1 && getBranchTokenTypes().get(tokenType);
```
romani, line 387: this is good.
```
✔ ~/java/github/checkstyle/checkstyle
$ head -n 25 ./target/generated-sources/antlr/com/puppycrawl/tools/checkstyle/grammar/java/JavaLanguageLexer.java
// Generated from com/puppycrawl/tools/checkstyle/grammar/java/JavaLanguageLexer.g4 by ANTLR 4.13.1
package com.puppycrawl.tools.checkstyle.grammar.java;

import com.puppycrawl.tools.checkstyle.grammar.CommentListener;
import com.puppycrawl.tools.checkstyle.grammar.CompositeLexerContextCache;
import com.puppycrawl.tools.checkstyle.grammar.CrAwareLexerSimulator;

import org.antlr.v4.runtime.Lexer;
import org.antlr.v4.runtime.CharStream;
import org.antlr.v4.runtime.Token;
import org.antlr.v4.runtime.TokenStream;
import org.antlr.v4.runtime.*;
import org.antlr.v4.runtime.atn.*;
import org.antlr.v4.runtime.dfa.DFA;
import org.antlr.v4.runtime.misc.*;

@SuppressWarnings({"all", "warnings", "unchecked", "unused", "cast", "CheckReturnValue", "this-escape"})
public class JavaLanguageLexer extends Lexer {
	static { RuntimeMetaData.checkVersion("4.13.1", RuntimeMetaData.VERSION); }

	protected static final DFA[] _decisionToDFA;
	protected static final PredictionContextCache _sharedContextCache =
		new PredictionContextCache();
	public static final int
		COMPILATION_UNIT=1, PLACEHOLDER1=2, NULL_TREE_LOOKAHEAD=3, BLOCK=4, MODIFIERS=5, 
```

romani (approved):
ok to merge

config/checker-framework-suppressions/checker-index-suppressions.xml
nrmancuso: >API is generic, and if somebody wants to put -1 in our method, they have full right in this, we should not throw exception, and just return false.

> But we are changing API method. It is not good.

We are still changing the behavior of our API for this update (from `IndexOutOfBoundsException` -> silent fail). Can we just annotate the method's parameter to specify that it should be nonnegative? Dancing around the dependency on Checker at compile time is getting old; if we are going to use this tool, let's use it correctly.

romani: I am afraid to use checker as compile time dependency for our API

nrmancuso: > I am afraid to use checker as compile time dependency for our API .

Then I don’t see any point in using it at all. Changing a part of our API to silently fail to be able to remove some suppression from a file is not an improvement to the project.

nrmancuso: I cannot approve this PR unless we are throwing an IllegalArgumentException.

nrmancuso: We should absolutely throw an exception here. If we cannot come to an agreement on this particular checker (index) then we should avoid sending PRs until we agree on how we are handling this. Changing production code in our core API to silently fail is not a positive change.

nrmancuso:
> @romani, the checker framework may add the `@Nonnegative` from `javax` as an alias to their annotation in the future update. I will keep track of that so we can bump the version of the checker framework in the future to better solve other suppressed violations. The issue I opened: [typetools/checker-framework#6507](https://github.com/typetools/checker-framework/issues/6507)

I propose that we gather all annotations that we will require to complete index checker integration and share this with checker maintainers. Once they support this, we can resume integration with index checker.

romani:
Sure, one day we will do summary. But if we pause, we will never find next problems for us. Our problem is that part of our code is API. But as slowly start thinking of Checker, on single and simple case we will eventually find more items that we can not do, and we will have better summary.

Problem here is that we have API `branchContains` . API method should not throw exceptions if that is possible. Bitset is details of implementation, we should not leak fact that we are limited to Bitset. 
Non of search like API throws exceptions if parameters are not good, they just return nothing.

romani:
@rnveach , please share your opinion.

nrmancuso:
> Bitset is details of implementation, we should not leak fact that we are limited to Bitset.

I agree with this statement, we should create an issue to discuss not using bitset.

rnveach:
> please share your opinion.

Which am I sharing on?

romani:
>create an issue to discuss not using bitset.

It is ok to use it, such a performance optimization is far future, and I am not sure about it. For now all we say is, if use do `contains`, we do not throw exception  on not expect content, we just return `false`.

config/checker-framework-suppressions/checker-index-suppressions.xml
romani: @rnveach , please share your opinion here on exception or just return of false

romani:
@rnveach , here is a discussion https://github.com/checkstyle/checkstyle/pull/14673#discussion_r1539414373

config/checker-framework-suppressions/checker-index-suppressions.xml
rnveach: I question the whole point of the new code check. We are reliant on ANTLR generating these numbers. I am not sure we have any contract with them which they should not be negative. Even if they did specify it somewhere in some document, I would think this is something we shouldn't enforce since its such low level and behinds the scene. As long as the method doesn't have a problem with positive/zero/negative, then I don't think we should check this.

romani: It would not be a concern if we have tokens as enum , so it will be type safe. But we have `int` so user can put anything to such methods that is comply by type, for example index of some loop in custom Check.

rnveach:
Here is my answer - https://github.com/checkstyle/checkstyle/pull/14673#discussion_r1544998932

Let me know when we have decided on a group decision.

romani:
@rnveach , I replied you on May12 https://github.com/checkstyle/checkstyle/pull/14673#discussion_r1597848629 , please consider this PR one more time.

rnveach:

Issue #3544, opened 2016-11-11
clean API methods and other methods from word "Error"


romani:
Error is a severity of violation. We report violations to user, severity of it is up to user.

There are number of methods, even non API, that mislead by name - should be fixed. Proposed term - violation .

old API methods could be marked as deprecated for few releases to let users be ready.

rnveach:
@romani confirm issue and names are still valid.

> LocalizedMessage ==>LocalizedViolation

This was renamed to Violation.

> AuditEvent ===> Violation
Error ==> Violation

I think this was a mistake naming 2 different classes with the same name.

romani:
Fixed.
`Error ==> Violation` was removed

Brijeshthummar02:
@romani is issue still open?

romani:
Issue is not approved, better to avoid it, not clear how to fix it.

PR #14687, opened 2024-03-19
Issue #3544: cleaned API methods related to MessageDispatcher & Renamed to ViolationDispatcher

MANISH-K-07:
Part of #3544 

Renamed MessageDispatcher --> `ViolationDispatcher`
Cleaned up all related methods.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
@@ -50,17 +50,17 @@ import com.puppycrawl.tools.checkstyle.api.FileSetCheck;
 import com.puppycrawl.tools.checkstyle.api.FileText;
 import com.puppycrawl.tools.checkstyle.api.Filter;
 import com.puppycrawl.tools.checkstyle.api.FilterSet;
-import com.puppycrawl.tools.checkstyle.api.MessageDispatcher;
 import com.puppycrawl.tools.checkstyle.api.RootModule;
 import com.puppycrawl.tools.checkstyle.api.SeverityLevel;
 import com.puppycrawl.tools.checkstyle.api.SeverityLevelCounter;
 import com.puppycrawl.tools.checkstyle.api.Violation;
+import com.puppycrawl.tools.checkstyle.api.ViolationDispatcher;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 
 /**
  * This class provides the functionality to check a set of files.
  */
-public class Checker extends AbstractAutomaticBean implements MessageDispatcher, RootModule {
+public class Checker extends AbstractAutomaticBean implements ViolationDispatcher, RootModule {
 
     /** Message to use when an exception occurs and should be printed as a violation. */
     public static final String EXCEPTION_MSG = "general.exception";
@@ -513,7 +513,7 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
      * @param fileSetCheck the additional FileSetCheck
      */
     public void addFileSetCheck(FileSetCheck fileSetCheck) {
-        fileSetCheck.setMessageDispatcher(this);
+        fileSetCheck.setViolationDispatcher(this);
         fileSetChecks.add(fileSetCheck);
     }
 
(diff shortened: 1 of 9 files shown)


MANISH-K-07:
I have no idea why we are missing our checkstyle workflow in ci ? 👀

MANISH-K-07:
@romani , ping pong please as this might've been missed :)

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -50,17 +50,17 @@
 import com.puppycrawl.tools.checkstyle.api.FileText;
 import com.puppycrawl.tools.checkstyle.api.Filter;
 import com.puppycrawl.tools.checkstyle.api.FilterSet;
-import com.puppycrawl.tools.checkstyle.api.MessageDispatcher;
 import com.puppycrawl.tools.checkstyle.api.RootModule;
 import com.puppycrawl.tools.checkstyle.api.SeverityLevel;
 import com.puppycrawl.tools.checkstyle.api.SeverityLevelCounter;
 import com.puppycrawl.tools.checkstyle.api.Violation;
+import com.puppycrawl.tools.checkstyle.api.ViolationDispatcher;
```
rdiachenko, line 57: Considering this interface seats in the api package, should we go via deprication process first, so we introduce a new interface but still keep the old one marked as deprecated? Otherwise, this PR may contain breaking changes. @romani what are your thoughts on this?

romani: Yes, we need to think on deprecation first. To let all plugins to migrate first, and if no known tools left, do deletion.
Is it possible?

src/main/java/com/puppycrawl/tools/checkstyle/api/ViolationDispatcher.java
```diff
@@ -24,7 +24,7 @@
 /**
  * Used by FileSetChecks to distribute AuditEvents to AuditListeners.
  */
-public interface MessageDispatcher {
+public interface ViolationDispatcher {
```
romani, line 27: This by fact, new class.
This good opportunity to make it with better design.

Major concerns on `void fireErrors(String fileName, SortedSet<Violation> errors);`
Name of method is not good. Should be `fireViolations` or similar.

@rdiachenko , what would you recommend?
We can change all.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/api/ViolationDispatcher.java
romani:
```
public interface ViolationDispatcher{
  void fireFileStarted(String fileName);
  void fireFileFinished(String fileName);
  void fireErrors(String fileName, SortedSet<Violation> errors);
}
```

Name of class and method becomes weird. Before update it was also not good, but it was too abstract it was less a concern, `MessageDispatcher`.

Problem is that we mixing `fireFileXxxx` with `fireViolation` and call it ViolationDispatrer, that is good name if only `fireViolation` be in interface.

May be we should just rename method in this interface to `fireViolation` and name of class `EventDispatcher`. Do we have any EventXxxx in our code base?

romani: https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java

MANISH-K-07: > We can not use Event word.

@romani , this PR is already outdated by 2 weeks. So, how do you suggest we finish this?

romani: @rdiachenko , is there any suggestion on api ?
@nrmancuso , @rnveach , please share your opinion.

rnveach: I think we first need to confirmation we are breaking compatibility with this issue and PR. We are renaming an interface that sits in the API class. I am also not seeing any deprecation in this PR as issue states.

If we are breaking compatibility, then issues needs label.

romani: we put labels after a fix, if we agree on fix, label is just reflection.
I am ok to break compatibility if we got better api that is easier to understand.

romani:
no energy to continue, PR is closed.
Issue is updated to reference this experiment.

Issue #14715, opened 2024-03-24
Enforce new naming convention "InputXpath{Checkname}Xxxx.java" in IT area


MANISH-K-07:
Follow-up of #6981 

Until now, XpathRegressionTest validated the pattern `SuppressionXpathRegressionXxxx.java` for input filenames
All our existing files are named that way, so we bypass them until this issue is fixed.

The pattern has been updated to `InputXpath{Checkname}Xxxx.java` as part of PR #14713 
All existing modules have been suppressed from validation. 

As part of this issue, one module needs to be picked and all input files need to be renamed. Note that the xpath queries need to be verified in respective `XpathRegression{Checkname}Test` after filename change. Also, the set `RENAME_INPUT_XPATH` needs to be updated by removing the updated module to re-enable pattern validation.

The list of suppressions can be found at :
https://github.com/checkstyle/checkstyle/blob/cd1afd907ff8f80b2953617e736040642dc47a1d/src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java#L121-L261

Example of update : https://github.com/checkstyle/checkstyle/commit/e70b8b5814b67cfc2c3ac47cde84822895c399d6

Kexin2000:
Because #14719  fixes the failing build in master, list in the first post does not have a MethodTypeParameterName. I am on that :)

MANISH-K-07:
I am on VisibilityModifier

romani:
@MANISH-K-07, I highly recommend to do fixes by packages or by packs of 20 files. to fnish this update quicker and not making a an epic from it.

romani:
@MANISH-K-07 , please update issue description to reference list of suppression and do udpates based on it.
checkboxes will be a problem to maintain, as fix is easy

Lmh-java:
I am on RecordTypeParameterName

Lmh-java:
I am on AbbreviationAsWordInName

romani:
@Lmh-java, we appreciate your help with easy issues, you are always welcome, but I recommend you next month to focus on more complicated issues to increase your chances to be selected to gsoc.

MANISH-K-07:
@romani , like you said, let's not make an epic of this issue. I'll close this with multiple PRs at a time in 2 batches like @rnveach suggested at https://github.com/checkstyle/checkstyle/pull/14725#issuecomment-2024283600

romani:
ok, lets let only @MANISH-K-07  to work on this issue.

PR #14727, opened 2024-03-28
Issue #14715: Enforce new naming convention on RecordTypeParameterName in IT area

Lmh-java:
Part of #14715

### DIFF
--- a/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionRecordTypeParameterNameTest.java
+++ b/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionRecordTypeParameterNameTest.java
@@ -39,9 +39,9 @@ public class XpathRegressionRecordTypeParameterNameTest extends AbstractXpathTes
     }
 
     @Test
-    public void testOne() throws Exception {
+    public void testTypeDeclared() throws Exception {
         final File fileToProcess = new File(getNonCompilablePath(
-                "SuppressionXpathRegressionRecordTypeParameterName1.java"));
+                "InputXpathRecordTypeParameterNameTypeDeclared.java"));
 
         final DefaultConfiguration moduleConfig =
                 createModuleConfig(RecordTypeParameterNameCheck.class);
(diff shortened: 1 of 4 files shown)


src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionRecordTypeParameterNameTest.java
```diff
@@ -41,7 +41,7 @@ protected String getCheckName() {
     @Test
     public void testOne() throws Exception {
         final File fileToProcess = new File(getNonCompilablePath(
-                "SuppressionXpathRegressionRecordTypeParameterName1.java"));
+                "InputXpathRecordTypeParameterNameExtend.java"));
```
MANISH-K-07, line 44:
```suggestion
                "InputXpathRecordTypeParameterNameTypeDeclared.java"));
```

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionRecordTypeParameterNameTest.java
```diff
@@ -67,7 +67,7 @@ public void testOne() throws Exception {
     @Test
     public void testTwo() throws Exception {
         final File fileToProcess = new File(getNonCompilablePath(
-                "SuppressionXpathRegressionRecordTypeParameterName2.java"));
+                "InputXpathRecordTypeParameterNameDefault.java"));
```
MANISH-K-07, line 70:
```suggestion
                "InputXpathRecordTypeParameterNameTypeDefault.java"));
```

MANISH-K-07 (commented):
IMO, it would be better if the record names were same as filenames just like we do with classes by convention.

Items :

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionRecordTypeParameterNameTest.java
Lmh-java: I don't see the point why you would name it like that. Could you explain it a bit? Basically, why would you add "TypeXXXX" to the file name?

Lmh-java: The test is testing the generic var with `extends` and `&` synatx.

MANISH-K-07: Please read about TypeVariables in java.
https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/lang/reflect/TypeVariable.html

Make changes if you feel my suggestions make more sense than suffixes "Extend" and "Default"

Lmh-java: Sorry for the late reply, done

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionRecordTypeParameterNameTest.java
Lmh-java: done

rnveach:
@MANISH-K-07 I assume you are done?

MANISH-K-07 (commented):
Yes @rnveach , 
I think everything looks good here but I'm not in the power to approve so please approve if you agree with the suggestions..

rnveach (approved):

PR #14016, opened 2023-11-11
Issue #13553: False positive in FallThroughCheck on last case

aayushRedHat:
Related: https://github.com/checkstyle/checkstyle/issues/13553

Diff Regression config: https://gist.githubusercontent.com/aayushRedHat/849313f94f6df07629e364070d5ed19a/raw/3fc325f6c384a7ca918dc4a5875fa26cc1444679/gistfile1.txt

Diff Regression projects: https://gist.githubusercontent.com/aayushRedHat/085b2f4c08f7e2859c92e0d2cb2144c9/raw/358a901648397fd9f66978d4ce51636694546177/gistfile1.txt

Issue #13553, opened 2023-08-15
False positive in FallThroughCheck on last case


Kevin222004:
I have read check documentation: https://checkstyle.org/checks/coding/fallthrough.html#FallThrough
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

-------

# Config
```
kevin@inspiron-15-5510:~/Desktop/check_style$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN" 
      "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
    <module name="TreeWalker">
        <module name="FallThrough">
            <property name="checkLastCaseGroup" value="true"/>
        </module>
    </module>
</module>
```

------

# code 
```
kevin@inspiron-15-5510:~/Desktop/check_style$ cat Test.java
class Test {
   void method5(int i, int j, boolean cond) {
      while (true) {
          switch (i){
          case 5:
               i++;
               /* block */ /* fallthru */ // comment
           }
       }
    }
}
```

----

# output 
```
$ java -jar checkstyle-10.12.4-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:5:11: Fall through from the last branch of the switch statement. [FallThrough]
Audit done.
Checkstyle ends with 1 errors.
```

------

# Expected
No violation

Update in doc to remove:
>The comment containing these words must be all on one line

AayushSaini101:
I am on it

AayushSaini101:
@romani @Kevin222004 So we have to resolve that issue: when only one case in switch and checklist property is set true and check style must not return any error? Is it

romani:
Cases in switch can be few.
But last case should not be highlighted with violation.

AayushSaini101:
> Cases in switch can be few. But last case should not be highlighted with violation.
@romani can you please highlight this better ? The checklastgroup property is used to check the last group but according to you we need to remove the violation for the last group that is check then what is need to of checklastgroup property?

Kevin222004:
https://checkstyle.sourceforge.io/checks/coding/fallthrough.html#FallThrough:~:text=The%20check%20honors,ugly%2C%20but%20possible). 
[...]

PR #12966, opened 2023-04-09
Issue #11166: remove getFileContent() usage from FallThroughCheck

Kevin222004:
Resolves https://github.com/checkstyle/checkstyle/issues/12948

partial for Issue #11166: remove getFileContent() usage from FallThroughCheck

--------------------

# Regression

final report :- https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/b242e7a_2023192425/reports/diff/index.html

final report 2 :- https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/b242e7a_2023213303/reports/diff/index.html

-----------------------------------------------------------------------------

Diff Regression config: https://gist.githubusercontent.com/Kevin222004/ebc8a446892fe0a9aaf746b1b1b43c05/raw/4e77df9181e2cbd8dd9e227169310190e2eb8f5b/fall.xml
Diff Regression projects: https://gist.githubusercontent.com/Kevin222004/9600f179b602d4c971bdb0a050099005/raw/360a95ed7bb60d7a0956e531199d484c4d6f6617/test-projects.properties
Report label: final-report-2.1

PR #14734, opened 2024-03-29
Issue #13553: False positive in FallThroughCheck on last case

Lmh-java:
Fix for Issue #13553, Continue of #14016

## Config
```xml
<module name="FallThrough">
    <property name="checkLastCaseGroup" value="true"/>
</module>
```

## Regression Report
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8a0a354_2024060435/reports/diff/checkstyle/index.html
5 violations are removed in Checkstyle code base. Differences in this report are intended. 

For example:
https://github.com/checkstyle/checkstyle/blob/4fe811819594747bcce6f62c216ea133c3cd2954/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThrough4.java#L109

```java
void method5(int i, int j, boolean cond) {
      while (true) {
          switch (i){
          case 5: // violation 'Fall\ through from the last branch of the switch statement'
              i++;
              /* block */ /* fallthru */ // comment
          }
      }
   }
```
This is no longer a violation in this case. Same for the other 4 cases.

Diff Regression config: https://gist.githubusercontent.com/Lmh-java/55bdd9e289522aff23374ad7b5571c2c/raw/4e52c69d3888eff5f976f0ae87c058e98c0dd72d/input-fall-thru-config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
@@ -20,9 +20,11 @@
 package com.puppycrawl.tools.checkstyle.checks.coding;
 
 import java.util.HashSet;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
@@ -42,9 +44,9 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * "fallthru", "fall thru", "fall-thru",
  * "fallthrough", "fall through", "fall-through"
  * "fallsthrough", "falls through", "falls-through" (case-sensitive).
- * The comment containing these words must be all on one line,
- * and must be on the last non-empty line before the {@code case} triggering
- * the warning or on the same line before the {@code case}(ugly, but possible).
+ * The comment containing these words must be on the last non-empty line
+ * before the {@code case} triggering the warning or on the same line before
+ * the {@code case}(ugly, but possible).
  * </p>
  * <p>
  * Note: The check assumes that there is no unreachable code in the {@code case}.
@@ -454,11 +456,19 @@ public class FallThroughCheck extends AbstractCheck {
      * @return true if relief comment found
      */
     private boolean hasReliefComment(DetailAST ast) {
-        return Optional.ofNullable(getNextNonCommentAst(ast))
-                .map(DetailAST::getPreviousSibling)
-                .map(previous -> previous.getFirstChild().getText())
-                .map(text -> reliefPattern.matcher(text).find())
-                .orElse(Boolean.FALSE);
+        final DetailAST nonCommentAst = getNextNonCommentAst(ast);
+        boolean result = false;
+        if (nonCommentAst != null) {
+            final int prevLineNumber = nonCommentAst.getPreviousSibling().getLineNo();
+            result = Stream.iterate(nonCommentAst.getPreviousSibling(),
+                            Objects::nonNull,
+                            DetailAST::getPreviousSibling)
+                    .takeWhile(sibling -> sibling.getLineNo() == prevLineNumber)
+                    .map(DetailAST::getFirstChild)
+                    .filter(Objects::nonNull)
+                    .anyMatch(firstChild -> reliefPattern.matcher(firstChild.getText()).find());
+        }
+        return result;
     }
 
 }
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/FallThroughCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/FallThroughCheck.xml
@@ -15,9 +15,9 @@
  "fallthru", "fall thru", "fall-thru",
  "fallthrough", "fall through", "fall-through"
  "fallsthrough", "falls through", "falls-through" (case-sensitive).
- The comment containing these words must be all on one line,
- and must be on the last non-empty line before the {@code case} triggering
- the warning or on the same line before the {@code case}(ugly, but possible).
+ The comment containing these words must be on the last non-empty line
+ before the {@code case} triggering the warning or on the same line before
+ the {@code case}(ugly, but possible).
  &lt;/p&gt;
  &lt;p&gt;
  Note: The check assumes that there is no unreachable code in the {@code case}.
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheckTest.java
@@ -252,7 +252,6 @@ public class FallThroughCheckTest extends AbstractModuleTestSupport {
         final String[] expected = {
             "48:11: " + getCheckMessage(MSG_FALL_THROUGH_LAST),
             "83:11: " + getCheckMessage(MSG_FALL_THROUGH_LAST),
-            "112:11: " + getCheckMessage(MSG_FALL_THROUGH_LAST),
         };
         verifyWithInlineConfigParser(
                 getPath("InputFallThrough4.java"),
@@ -331,11 +330,7 @@ public class FallThroughCheckTest extends AbstractModuleTestSupport {
     public void testLastLine() throws Exception {
         final String[] expected = {
             "21:13: " + getCheckMessage(MSG_FALL_THROUGH),
-            // until https://github.com/checkstyle/checkstyle/issues/13553
-            "33:13: " + getCheckMessage(MSG_FALL_THROUGH),
             "99:39: " + getCheckMessage(MSG_FALL_THROUGH_LAST),
-            // until https://github.com/checkstyle/checkstyle/issues/13553
-            "107:11: " + getCheckMessage(MSG_FALL_THROUGH_LAST),
         };
         verifyWithInlineConfigParser(
                 getPath("InputFallThroughLastLineCommentCheck.java"),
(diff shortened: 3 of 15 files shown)


Lmh-java:
GitHub, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -454,11 +454,24 @@ private boolean hasFallThroughComment(DetailAST currentCase) {
...
+                final DetailAST firstChild = previousSibling.getFirstChild();
+                if (firstChild != null && reliefPattern.matcher(firstChild.getText()).find()) {
+                    hasReliefComment = true;
+                    break;
+                }
+                previousSibling = previousSibling.getPreviousSibling();
+            }
+        }
+
+        return hasReliefComment;
```
Lmh-java, line 474: Refactor `hasReliefComment`

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/FallThroughCheck.xml
```diff
@@ -15,9 +15,9 @@
  "fallthru", "fall thru", "fall-thru",
  "fallthrough", "fall through", "fall-through"
  "fallsthrough", "falls through", "falls-through" (case-sensitive).
- The comment containing these words must be all on one line,
- and must be on the last non-empty line before the {@code case} triggering
- the warning or on the same line before the {@code case}(ugly, but possible).
+ The comment containing these words must be on the last non-empty line
+ before the {@code case} triggering the warning or on the same line before
+ the {@code case}(ugly, but possible).
```
Lmh-java, line 20: Remove doc as required

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheckTest.java
```diff
@@ -383,12 +373,45 @@ public void testLabeledBreak() throws Exception {
...
+        final String[] expected = {
+            "21:9: " + getCheckMessage(MSG_FALL_THROUGH),
+            "45:9: " + getCheckMessage(MSG_FALL_THROUGH),
+            "54:9: " + getCheckMessage(MSG_FALL_THROUGH),
+            "60:9: " + getCheckMessage(MSG_FALL_THROUGH),
+            "77:9: " + getCheckMessage(MSG_FALL_THROUGH),
+            "94:9: " + getCheckMessage(MSG_FALL_THROUGH),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputFallThroughWithoutReliefPattern.java"),
```
Lmh-java, line 414: Add sufficient test cases to enhance code cov and pitest.

Lmh-java:
@nrmancuso @romani please take a look.

nrmancuso (approved):

nrmancuso:
Rebasing on latest master

romani (approved):
Ok to merge

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/FallThroughCheck.xml
rnveach: > > Update in doc to remove:

> The comment containing these words must be all on one line

I don't remember this conversation regarding this but does this mean if I have a suppression word as "suppress me", it will match things like "suppress\nme"? Do you have a test example showing this?

rnveach: Your updated documentation says:
>  ...on the same line before the ...

How is that different then `...must be all on one line...`?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
```diff
@@ -454,11 +454,24 @@ private boolean hasFallThroughComment(DetailAST currentCase) {
      * @return true if relief comment found
      */
     private boolean hasReliefComment(DetailAST ast) {
-        return Optional.ofNullable(getNextNonCommentAst(ast))
-                .map(DetailAST::getPreviousSibling)
-                .map(previous -> previous.getFirstChild().getText())
-                .map(text -> reliefPattern.matcher(text).find())
-                .orElse(Boolean.FALSE);
```
rnveach, line 461: More just curious why we can't keep this original structure of the code and have to revert to basically a pre-Java 8 version of the code?

rnveach:
> https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8a0a354_2024060435/reports/diff/index.html
> https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8a0a354_2024060435/reports/diff/diff_config.html

Config is when `checkLastCaseGroup` is true, but it looks like code is also executed when it is false too.
We need another regression with other value of `checkLastCaseGroup`.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
nrmancuso: We could certainly create a list of siblings here and filter them to make this method more concise and avoid mutable data.

Lmh-java: Sure. I am not very familiar with this new style as well. I will try to change this to the new style.

Lmh-java: Done, please have a look.

```java
result = Stream.iterate(nonCommentAst.getPreviousSibling(),
                            Objects::nonNull, DetailAST::getPreviousSibling)
                    .takeWhile(sibling -> sibling.getLineNo() == prevLineNumber)
                    .filter(Objects::nonNull)
                    .anyMatch(firstChild -> reliefPattern.matcher(firstChild.getText()).find());
```

I didn't really create a list of siblings, but I used `iterate` to iterate over all thes sblings. I think this way might be more efficient.

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/FallThroughCheck.xml
Lmh-java: > > > Update in doc to remove:
> 
> > The comment containing these words must be all on one line
> 
> I don't remember this conversation regarding this but does this mean if I have a suppression word as "suppress me", it will match things like "suppress \n me"? Do you have a test example showing this?

No, I don't think "suppress \n me" will match.
This is matching by the pattern
```java
```

Lmh-java: > Your updated documentation says:
> 
> > ...on the same line before the ...
> 
> How is that different then `...must be all on one line...`?

I didn't see the conversion about this part either, but I think "must be all on one line" is redundant and confusing. I don't really understand what it means. The subject of this sentence is "the comment containing these words". However, we only need one "comment" containing these words to suppress this check for that branch.

If we remove this "must be all on one line.", and updated it to
>The comment containing these words must be on the last non-empty line
 before the {@code case} triggering the warning or on the same line before
 the {@code case}(ugly, but possible).

I think this might be able to explain everything?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
rnveach: @nrmancuso Please assist.

@Lmh-java If you think this a 1 to 1, then regression should return the same results.

Lmh-java:
GitHub, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FallThroughCheck.java
nrmancuso: @Lmh-java very good, takeWhile is a great stream operation to use here. 🚀

Lmh-java: >very good, takeWhile is a great stream operation to use here. 🚀

Thank you! 

>If you think this a 1 to 1, then regression should return the same results.

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/7e9dba3_2024170824/reports/diff/index.html
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8a0a354_2024060435/reports/diff/index.html

@rnveach These two regression report has the same result. Please take a look.

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/FallThroughCheck.xml
rnveach: It looks like comment in issue came from https://github.com/checkstyle/checkstyle/pull/12966#issuecomment-1675649800 . This was about "the last comment on that line". This is why the issue has `/* block */ /* fallthru */ // comment`.

If you agree, please restore original wording and add some wording that it can be any comment on that line.

Can you point to keywords where the suppression text is not the last comment on the line. If we don't have one, then create it and point to it.

Example:
/* trigger */ /* comment */ // last
/* comment */ /* fallthru */ // last

rnveach:
Latest review discussion is at https://github.com/checkstyle/checkstyle/pull/14734#discussion_r1597719839

romani:
Moved to https://github.com/checkstyle/checkstyle/pull/15405

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/FallThroughCheck.xml
romani: Let's continue here https://github.com/checkstyle/checkstyle/pull/15405/files?diff=unified&w=0#r1697058346

Issue #14716, opened 2024-03-24
ClassDefHandler should exclude indentation check of modifiers like ANNOTATIONS, STATIC & STRICTFP which are preceeded by 


kalpadiptyaroy:
I have read check documentation: https://checkstyle.org/checks/xxxxxx/nameofaffectedcheck.html
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac YOUR_FILE.java
#[[MAKE SURE THERE IS SUCCESSFUL COMPILATION]]

/var/tmp $ cat config.xml
#[[PLACE YOUR OUTPUT HERE]]

/var/tmp $ cat YOUR_FILE.java
#[[PLACE YOU OUTPUT HERE]]

/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-X.XX-all.jar -c config.xml YOUR_FILE.java
#[[PLACE YOUR OUTPUT HERE]]
```

For Windows users, please use `type` instead of `cat` and run

```cmd
set RUN_LOCALE="-Duser.language=en -Duser.country=US"
java %RUN_LOCALE% -jar checkstyle-X.XX-all.jar -c config.xml YOUR_FILE.java
```

in place of the last 2 commands above.

---

**Describe what you expect in detail.**

---

Still not clear ???
see example - https://checkstyle.org/report_issue.html#How_to_report_a_bug.3F

ATTENTION: FAILURE TO FOLLOW THE ABOVE TEMPLATE WILL RESULT IN THE ISSUE BEING CLOSED.

rnveach:
@kalpadiptyaroy There is literally no information about what this issue is about. Issue title is not enough.
https://checkstyle.org/report_issue.html#How_to_report_a_bug.3F

rnveach:
Expected output says to remove line 12, which is
````
  public
      static class C4{}  // this is a violation too, can't make check happy, expects 2
````

I disagree with this issue, specifically the words `should exclude`. Every line that is code should be violated for indentation. It doesn't make sense to exclude lines completely. We should fix any issues so violations can be resolved.

kalpadiptyaroy:
@rnveach Updated the title. If there is still a disagreement.

Let's consult with @nrmancuso, and know his opinion as well.

[...]

PR #14743, opened 2024-03-30
Issue #14716: Fix the violation for modifiers wrapped in class def

kalpadiptyaroy:
Issue #14716: Fix the violation for modifiers wrapped in class definition
Diff Regression config: https://gist.githubusercontent.com/kalpadiptyaroy/187788693b6445aed065ce39a171c502/raw/cdb3feb926b07fd8c07df72997df9de4d38957ee/checks.xml
Diff Regression projects: https://gist.githubusercontent.com/kalpadiptyaroy/80471b85dff631584c7668ba773186c0/raw/69bc6e343a083698e61a0afd039a69cb0c1b28a2/my-projects-to-test-on.properties

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
@@ -24,6 +24,7 @@ import java.util.Arrays;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
+import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
 
 /**
  * Abstract base class for all handlers.
@@ -500,6 +501,10 @@ public abstract class AbstractExpressionHandler {
              modifier != null;
              modifier = modifier.getNextSibling()) {
             if (isOnStartOfLine(modifier)
+                 && (!TokenUtil.isOfType(modifier,
+                        TokenTypes.ANNOTATION, TokenTypes.STRICTFP, TokenTypes.LITERAL_STATIC)
+                    || modifier.getPreviousSibling() == null
+                    || !isPrecededByAnyModifier(modifier))
                 && !getIndent().isAcceptable(expandedTabsColumnNo(modifier))) {
                 logError(modifier, "modifier",
                     expandedTabsColumnNo(modifier));
@@ -588,6 +593,22 @@ public abstract class AbstractExpressionHandler {
                 && isOnStartOfLine(lparen)) {
             logError(lparen, "lparen", expandedTabsColumnNo(lparen));
         }
+    }
+
+    /**
+     * Check whether given node a preceding access modifier or not.
+     *
+     * @param node the node whose predecessors are to be checked.
+     * @return true if node is preceded by an access modifier else false.
+     */
+    private boolean isPrecededByAnyModifier(DetailAST node) {
+        DetailAST tempNode = node;
+        while (tempNode.getPreviousSibling() != null
+            && !TokenUtil.isOfType(tempNode.getPreviousSibling(),
+            TokenTypes.LITERAL_PUBLIC, TokenTypes.FINAL, TokenTypes.ABSTRACT)) {
+            tempNode = tempNode.getPreviousSibling();
+        }
+        return tempNode.getPreviousSibling() != null;
     }
 
 }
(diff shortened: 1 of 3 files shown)


kalpadiptyaroy:
We can test any potential scenario by adding that case to this file. 
`InputIndentationClassDeclarationWrapped`. 

Welcome to every suggestion.

kalpadiptyaroy:
GitHub, generate report

kalpadiptyaroy:
GitHub, generate report

kalpadiptyaroy:
Above report is without openjdk17 (excluded since it is raising **OutOfMemoryException**)

kalpadiptyaroy:
GitHub, generate report

kalpadiptyaroy:
Generating a new report - including new popular open-source **java** repos.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -353,4 +353,20 @@ public static BitSet asBitSet(String... tokens) {
                 .collect(BitSet::new, BitSet::set, BitSet::or);
     }
 
+    /**
+     * Check whether given node a preceding access modifier or not.
+     *
+     * @param node the node whose predecessors are to be checked.
+     * @return true if node is preceded by an access modifier else false.
+     */
+    public static boolean isPrecededByAnyModifier(DetailAST node) {
```
romani, line 362: please move it to src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
and test it by `InputIndentationxxxxxxx.java` not a pure junit.

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
```diff
@@ -500,6 +501,10 @@ protected void checkModifiers() {
              modifier != null;
              modifier = modifier.getNextSibling()) {
             if (isOnStartOfLine(modifier)
+                 && (!TokenUtil.isOfType(modifier,
+                        TokenTypes.ANNOTATION, TokenTypes.STRICTFP, TokenTypes.LITERAL_STATIC)
+                    || modifier.getPreviousSibling() == null
+                    || !isPrecededByAnyModifier(modifier))
```
rnveach, line 507: @romani @nrmancuso 

I am continuing on the discussion at https://github.com/checkstyle/checkstyle/issues/14716#issuecomment-2028461416 , I don't see how skipping specific tokens will serve a benefit. I don't think we should be reviewing this PR until we have some agreement here.

Including @nrmancuso 's original comments, the input file in this PR needs to be vastly expanded.

We need a duplicate input file with all indentation at 0 showing the violations produced by bad code. Regression was done on good formatted code, so it will not provide less value for this check in those terms. We should also do force strict as it will tell us exactly what indentation should be and when it is absolutely correct and not just meeting the min requirement.

We also need inputs where these tokens being skipped are completely alone on a line and are the first AST in the class def. All the current examples start with public, abstract, final which are not any of the tokens listed here.

rnveach (changes requested):

kalpadiptyaroy:
@rnveach - I agree with you. We must rigorously see the regression of this patch on dirty codes elaborating as many scenario's as possible and then make a move.

I will try and keep preparing samples of dirty codes with wrappings in class defs in as many possible ways as I think of and compare the regression reports.
I will be discussing that in issue thread and we only come back to PR when everyone is in agreement and confident of this move.

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
nrmancuso: >There doesn't seem to be any clear reason why we are targeting these tokens

Yep, my sentiments exactly.

romani:
@kalpadiptyaroy, please find time to finish this PR

kalpadiptyaroy:
@romani @rnveach - Its quite some time. I have not been able to focus on this PR. But I would like to put an effort once again to arrive at a common conclusion.

After a quick re-cap, I agree with your views. But since I am a new contributor to this project, so it is difficult for me to visualize how to treat all identifiers equally.

May I request you guys to provide a good and details explanation of how you people are visualizing the identifiers to treated equally? Examples are really appreciated.

Thanks :)

rnveach:
Indentation is not a check that a new contributor should take on. Even maintainers don't want to touch this check as it is too complex, and has too many issues. Proper fixes for this check will require changing how we handle our AST, and that will be a such a huge break that no one will want to move forward with that.

You either have to dive deep into this check and understand the issue better why some modifiers are being treated differently, or close this PR and find another issue which would be recommended for your current experience.

romani:
Look like PR is too complicated.
I am closing it to let someone more experienced to continue work.

@kalpadiptyaroy , please be welcome with contribution to other issues. Sorry that this update was too complicated for both sides.

Issue #14726, opened 2024-03-27
new Check: ConstructorsDeclarationGrouping to check the grouping of overloaded constructors


Zopsss:
initially discuses at https://github.com/checkstyle/checkstyle/issues/14415 and https://github.com/checkstyle/checkstyle/pull/14398
Reason of new check creation is desire to cover Google style request for this - https://github.com/checkstyle/checkstyle/issues/14273

I have downloaded the latest cli from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

**How it works Now:**
Checkstyle have a module called  [OverloadMethodsDeclarationOrder](https://checkstyle.org/checks/coding/overloadmethodsdeclarationorder.html) which checks that if the overloaded methods are grouped together or not, if there is another method between overloaded methods then it'll give you an error. Currently, in checkstyle there is no module to check the grouping of overloaded constructors in a class, enum, abstract class, records. This new module will check that the overloaded constructors are grouped together or not. If there is another method or anything else between the constructors ( expect comments ), it will give an error.

**How I want check to work**
```xml
$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
    <module name="TreeWalker">
        <module name="ConstructorsDeclarationGrouping"/>
    </module>
</module>
```

**Test.java**
```java
$ cat Test.java 
public class Test {
    public Test(){}

    public Test(int a){}

    public void testing() {}

    public Test(int a, int b) {}

}
```

**Expected output**
The module should check the grouping of overloaded constructors and give an error if they don't follow the rule. It should give an error something like this:

```
$ java -jar checkstyle-10.13.0-all.jar -c config.xml Test.java 
Starting audit...
[ERROR] C:\checkstyle testing\Test.java:8:3: All overloaded constructors should be grouped together. Previous overloaded constructor located at line ''4''.[ConstructorsDeclarationGrouping]
Audit done.
Checkstyle ends with 1 errors.
```

---
**Some other examples showing how the check should work**
Example 1: This is fine
```java
public class Example1 {
    Example1 () { }

    Example1 (int a) { }

    Example1 (int a, int b) { }

    void testing () { }
}
```

[...]

PR #14749, opened 2024-04-01
Issue #14726: new check: ConstructorsDeclarationGroupingCheck

Zopsss:
Issue: #14726 

Implementing a new module called ConstructorsDeclarationGroupingCheck

---

New module config: https://gist.githubusercontent.com/Zopsss/b3dabe65814aae54ff98c62f9fe54d3e/raw/6372644a5aa81acba09f1ddc607a79cd0b56de58/ConstructorsDeclarationGrouping.xml

Diff Regression projects: https://gist.githubusercontent.com/Zopsss/2aaf1a18ce741b3d498b7e99c743845a/raw/3b3d859ba7d9955b8f8668aa9509d5eb8bdcd312/project-list-2.properties

Contrib repo PR: https://github.com/checkstyle/contribution/pull/851


Zopsss:
hey @nrmancuso @rnveach @romani I've implemented this check locally and added test cases. Everything was working fine then I ran the mvn clean verify, it was giving me these erorrs:

```
[ERROR] [checkstyle] [ERROR] C:\Open Source\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\filters\SuppressWithPlainTextCommentFilter.java:328:9: All overloaded constructors should be grouped together. Previous constructor was at line number '322'. [ConstructorsDeclarationGrouping]
[ERROR] [checkstyle] [ERROR] C:\Open Source\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\XMLLoggerTest.java:431:9: All overloaded constructors should be grouped together. Previous constructor was at line number '300'. [ConstructorsDeclarationGrouping]]
```

these errors are the part of this new check. It is giving error for [SuppressWithPlaintextCommentFilter.java](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java) & [XMLLoggerTest.java](https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java)

 It is showing that  in [SuppressWithPlainTextCommentFilter.java](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java) line number [328](https://github.com/checkstyle/checkstyle/blob/f7e952b401ef0056cc2eac78a894bbc2e8ce8481/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java#L328),it does not have constructors grouped together and previous constructor was at line number [322](https://github.com/checkstyle/checkstyle/blob/f7e952b401ef0056cc2eac78a894bbc2e8ce8481/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java#L322). But the line number 322 is a comment not a constructor. The visitToken() method gets only called for CTOR_DEF & COMPACT_CTOR_DEF tokens, so I'm confused how it is giving error for a comment.

And in XMLLoggerTest, it giving me an error at line [431](https://github.com/checkstyle/checkstyle/blob/f7e952b401ef0056cc2eac78a894bbc2e8ce8481/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java#L431), and the previous constructor it is showing is at line [300](https://github.com/checkstyle/checkstyle/blob/f7e952b401ef0056cc2eac78a894bbc2e8ce8481/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java#L300), which is not a constructor. 

I also tried to replicate these files into my input files for test cases, but they were working fine there. All of my test cases were passing even after replicating these files. I've also added test cases to check for constructors grouping in nested enum, class, records, etc. And they are all also working fine.

Here is the visitToken() method which contains all the logic of the check.

https://github.com/checkstyle/checkstyle/blob/feba1977f846e7cb4fbed8ee76239ca9ee7a5140/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java#L93

I tried reproducing these errors in the input test files, but they were working fine and not giving any errors. can pls you help me with this one? I'm confused how it is giving errors for something which is not a constructor when the visitToken() gets called for only constructors and stores only constructors information.

Zopsss:
hey @romani @nrmancuso @rnveach the new module implementation is almost done, all the CIs are passing expect semaphore. It was complaining about creating a new PR at contribution repo for adding the new module and something about pitest.

I have made the PR about adding the new module in the contribution repo and added its link in the PR description but I'm not familiar with the Pitest error it is giving. Can you help me with this?

Here is the error link: https://checkstyle.semaphoreci.com/jobs/c43bc698-c72f-472b-82e6-ceadb034f92e

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_de.properties
```diff
@@ -3,6 +3,7 @@ assignment.inner.avoid=Innere Zuweisungen sollten vermieden werden.
 avoid.clone.method=Die Methode clone() sollte vermieden werden.
 avoid.double.brace.init=Vermeiden Sie doppelte geschweifte Klammern.
 avoid.finalizer.method=Die Methode finalize() sollte vermieden werden.
+constructors.declaration.grouping=All overloaded constructors should be grouped together. Previous constructor was at line number ''{0}''.
```
Zopsss, line 6: I have not translated the error message into different languages. Because the translated messages are in some kind of encoded format and I'm not so sure how that works.

romani: >All overloaded constructors should be grouped together. Previous constructor was at line number ''{0}''.

make message to be `All overloaded constructors should be grouped together. Previous constructor was at line number '5'.`and use Google translator. and use all wording from Google in other languages properties. Community will fix wording eventually.

romani:
> [INFO] --- spotbugs-maven-plugin:4.8.3.1:check (default) @ checkstyle --- com.puppycrawl.tools.checkstyle.checks.coding.ConstructorsDeclarationGroupingCheck.visitToken(DetailAST) checks a map with containsKey(), before using get() [com.puppycrawl.tools.checkstyle.checks.coding.ConstructorsDeclarationGroupingCheck] At ConstructorsDeclarationGroupingCheck.java:[line 94] MUI_CONTAINSKEY_BEFORE_GET

please fix it, and rebase on most recent master.

Zopsss:
> please fix it, and rebase on most recent master.

to fix that error, I did the following changes:

Before:
https://github.com/checkstyle/checkstyle/blob/7a3c06b28506755428eaf309a29966f4235e85e1/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java#L85-L87

After:
https://github.com/checkstyle/checkstyle/blob/c2eae4fb38ad6f3f8882b7d834645e8738fde3ef/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java#L85-L88

but now `Pitest / test (coding - 1)` CI is failing. Before changing the code it was passing but after the change it started giving error. Here is the error link: https://github.com/checkstyle/checkstyle/actions/runs/8604545703/job/23578894912?pr=14749#step:6:499

there are also many new mutation survivals, but this was not happening before the code change. Can you please explain me the reason behind failing Pitest CI when the code is not changed that much and its logic is also the same as before. @romani

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_de.properties
Zopsss: Isn't the new message you suggested is same as the old message? Expect the `'5'`, which I believe that would be a hardcode value instead of a placeholder? 

Btw I've translated the messages into different languages as you asked. @romani

src/xdocs/checks/coding/index.xml
```diff
@@ -395,6 +395,16 @@
...
             </td>
           </tr>
+          <tr>
+            <td>
+              <a href="constructorsdeclarationgrouping.html#ConstructorsDeclarationGrouping">
+                ConstructorsDeclarationGrouping
+              </a>
+            </td>
+            <td>
+              Checks that overloaded constructors are grouped together.
```
romani, line 405: Remove `overliaded` it is not applicable for constructors.
Remove it completely from all your PR

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
```diff
@@ -0,0 +1,104 @@
...
+ * <li>
+ * {@code constructors.declaration.grouping}
+ * </li>
+ * </ul>
+ *
+ * @since 10.16.0
+ */
+
+@FileStatefulCheck
+public class ConstructorsDeclarationGroupingCheck extends AbstractCheck {
```
romani, line 52: Can you make it stateless?
It will work well when we make checkstyle multithreading.

romani (changes requested):
Items

src/xdocs/checks/coding/index.xml
Zopsss: Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
Zopsss: Firstly I tried making it stateless using other ways but the `mvn clean verify` was giving errors so I have added the suppressions to make it work.

https://github.com/checkstyle/checkstyle/blob/fde8f4aba57f969543a0bf3b9a67896ead38763a/src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java#L156-L157

https://github.com/checkstyle/checkstyle/blob/fde8f4aba57f969543a0bf3b9a67896ead38763a/src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java#L193

After adding the suppressions, `mvn clean verify` was running perfectly

mahfouz72: > mutating them would make test cases fail.

@Zopsss how is that? if the tests fail when hardcoding the mutation then how they survive in the first place maybe you hardcode the mutation in the wrong way

Zopsss: > how is that?

NVM it was showing that many surviving mutations because I didn't added the Test class in the TargetTests of pitest-coding-1. Thanks for helping me out. Now it is showing only 3 surviving mutations :)


@romani Looking into the surviving mutations now 🫡

Zopsss:
GitHub, generate report

Zopsss:
GitHub, generate report

Zopsss:
GitHub, generate report

Zopsss:
GitHub, generate report

romani:
OOM - https://github.com/checkstyle/checkstyle/actions/runs/8642972336/job/23695060359#step:9:2443
please use smaller projects for validation.

Zopsss:
GitHub, generate report

Zopsss:
GitHub, generate report

Zopsss:
GitHub, generate report

Zopsss:
hey @romani the reports looks fine to me. The check is working perfectly. Is there anything else left to do?

Zopsss:
GitHub, generate site

Zopsss:
@romani ping

config/checkstyle-non-main-files-suppressions.xml
```diff
@@ -278,5 +278,7 @@
          files="src[\\/]xdocs[\\/]checks[\\/]whitespace[\\/]nowhitespacebeforecasedefaultcolon.xml.template"/>
   <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]filters[\\/]suppresswarningsfilter.xml.template"/>
+  <suppress id="propertiesMacroMustExist"
+            files="src[\\/]xdocs[\\/]checks[\\/]coding[\\/]constructorsdeclarationgrouping.xml.template"/>
```
romani, line 282: Please put it to "coding" group of suppression

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionConstructorsDeclarationGroupingTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    protected String getCheckName() {
+        return clazz.getSimpleName();
+    }
+
+    @Test
+    public void testOne() throws Exception {
+        final File fileToProcess = new File(
+                getPath("SuppressionXpathRegressionConstructorsDeclarationGrouping1.java"));
```
romani, line 44: In issue https://github.com/checkstyle/checkstyle/issues/14715
We try better names of tests and inputs.
Please use good meaningful names.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * A key is pointing to the warning message text in "messages.properties"
+     * file.
+     */
+    public static final String MSG_KEY = "constructors.declaration.grouping";
+
+    /**
+     * Specifies different Object Blocks scope.
+     */
+    private final Map<DetailAST, Integer> allObjBlocks = new HashMap<>();
```
romani, line 63: This is a state.
If we can not avoid state ( fields of class) , we should clean/reset fields at the begging of each file validation. Same instance of Check is reused over all files to validate.

 If you see no way to make it stateless, it is ok. You can just explain a reason on why stateless is not possible, and stay with stateful.

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
```diff
@@ -152,7 +152,9 @@ public class ImmutabilityTest {
         "com.puppycrawl.tools.checkstyle.checks.coding.IllegalTokenTextCheck.formatString",
         "com.puppycrawl.tools.checkstyle.checks.javadoc.WriteTagCheck.tagRegExp",
         "com.puppycrawl.tools.checkstyle.checks.naming.AbstractNameCheck.format",
-        "com.puppycrawl.tools.checkstyle.checks.whitespace.AbstractParenPadCheck.option"
+        "com.puppycrawl.tools.checkstyle.checks.whitespace.AbstractParenPadCheck.option",
+        "com.puppycrawl.tools.checkstyle.checks.coding.ConstructorsDeclarationGroupingCheck"
+            + ".allObjBlocks"
```
romani, line 157: Not ok, this points to fact that you marked Checks steless, but it is stateful by facts.

romani (changes requested):
Items

config/checkstyle-non-main-files-suppressions.xml
Zopsss: done.

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionConstructorsDeclarationGroupingTest.java
Zopsss: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
Zopsss: > This is a state. If we can not avoid state ( fields of class) , we should clean/reset fields at the begging of each file validation. Same instance of Check is reused over all files to validate.

Done. Used the `beginTree()` method to achieve this:

https://github.com/checkstyle/checkstyle/blob/cadb0dc3a4635dd53eaa24ad6d41fd6e2cef7fd2/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java#L83-L86


---

> If you see no way to make it stateless, it is ok. You can just explain a reason on why stateless is not possible, and stay with stateful.

I guess we can't make this check stateless because the check requires tracking the scope of the overloaded constructors, so we need a global variable to keep track of this. A local variable will not work in this situation. Here is the example of what I'm trying to say:

https://github.com/checkstyle/checkstyle/blob/cadb0dc3a4635dd53eaa24ad6d41fd6e2cef7fd2/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/constructorsdeclarationgrouping/InputConstructorsDeclarationGrouping.java#L29-L53

@romani

src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java
Zopsss: Done. Removed this suppression, explained the reason why this check cannot be stateless here: https://github.com/checkstyle/checkstyle/pull/14749#discussion_r1564656284

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionConstructorsDeclarationGroupingTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+        );
+
+        runVerifications(moduleConfig, fileToProcess, expectedViolation, expectedXpathQueries);
+    }
+
+    @Test
+    public void testTwo() throws Exception {
+        final File fileToProcess = new File(
+                getPath("InputXpathConstructorsDeclarationGroupingTwo.java"));
```
romani, line 76: InputXpathConstructorsDeclarationGroupingEnum.java

Please rename same approach other input.

romani (changes requested):
Items

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionConstructorsDeclarationGroupingTest.java
Zopsss: done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
```diff
@@ -0,0 +1,109 @@
...
+
+import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+/**
+ * <p>
+ * Checks that all constructors are grouped together.
+ * If there is anything between overloaded constructors ( expect comments )
```
romani, line 33: Please forget word `overloaded` and never use it in code and comments. :)

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public void beginTree(DetailAST rootAst) {
+        allObjBlocks.clear();
+    }
+
+    @Override
+    public void visitToken(DetailAST ast) {
+        final DetailAST currObjBlock = ast.getParent();
+        final Integer previousCtorLineNo = allObjBlocks.get(currObjBlock);
```
romani, line 91: Why we can not simply check previous sibling on each ctor? 
We might need some collection to keep track if ctor is first or not. This collection will just track that ctor is detected, so all other ctors should be in siblings.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
Zopsss: so sorry, I forgot to remove it from here :(

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
Zopsss: This will work in most of the situations. We can keep track if the current CTOR is first or not and check for its previous sibling, but if there is another block between the constructors then this approach will not work. For example:

https://github.com/checkstyle/checkstyle/blob/5c4be0c514d0a63aa89d6e6ea711c4d0bf294632/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/constructorsdeclarationgrouping/InputConstructorsDeclarationGrouping.java#L29-L53

Here, the `Testing(double d) {}` occurs after an inner class and based on our check's implementation we will loose track of the previous CTOR and its position if we use a normal variable like a boolean or something to keep track of the previous constructor as it will get reset. That's why I used HashMap to keep track of all blocks and see if there is a constructor occurred in current block or not. We're getting the previous CTOR's line number of the current block and checking if it is null or not, null means there is no previous CTOR in the current block. If there is no CTOR occurred, then put the current block in the HashMap and store the constructor's line number. But if there is a constructor already occurred in the current block, then we check for the previous sibling, if the previous sibling is not a constructor then we will log the violation and if the previous sibling is a constructor then we will update the previous constructor's line number to current constructor's line number.

And we have provided the last occurred constructor's line number in the violation so user can know where the previous constructor is located. @romani

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
Zopsss: I have updated the docs, comments & error messages and make sure `overloaded` word is not used anywhere.

Zopsss:
The `mvn clean verify` was not working properly locally, in this branch as well as master branch, Idk why that was happening. So I have pushed the changes without running it, I added some new examples in the template file but didn't generate those examples in the xml file, so I guess that will give some errors. I will fix those soon once `mvn clean verify` issue is resolved.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
Zopsss: I have updated all the docs, comments, etc to exclude the word `overloaded`. I will look into the failing CIs soon as mentioned here https://github.com/checkstyle/checkstyle/pull/14749#issuecomment-2054180664

romani:
Did you share with us how it is failing on your local?

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionConstructorsDeclarationGroupingTest.java
```diff
@@ -0,0 +1,105 @@
...
+                        + "/OBJBLOCK/CTOR_DEF/IDENT"
+                        + "[@text='InputXpathConstructorsDeclarationGroupingClass']"
+
+        );
+
+        runVerifications(moduleConfig, fileToProcess, expectedViolation, expectedXpathQueries);
+    }
+
+    @Test
+    public void testTwo() throws Exception {
```
romani, line 74: We usually demand 3 tests for xpath, please add one more for inner class in class or better on https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#COMPACT_CTOR_DEF 

And change name of method avoid abstract Two, and use Enum
. Same for others test methods.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
```diff
@@ -0,0 +1,109 @@
...
+
+import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+/**
+ * <p>
+ * Checks that all constructors are grouped together.
+ * If there is any code between the constructors ( expect comments )
```
romani, line 33: >expect

Comment are ignored.

romani (changes requested):
Items

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_de.properties
```diff
@@ -3,6 +3,7 @@ assignment.inner.avoid=Innere Zuweisungen sollten vermieden werden.
 avoid.clone.method=Die Methode clone() sollte vermieden werden.
 avoid.double.brace.init=Vermeiden Sie doppelte geschweifte Klammern.
 avoid.finalizer.method=Die Methode finalize() sollte vermieden werden.
+constructors.declaration.grouping=Alle Konstruktoren sollten zusammengefasst werden. Der vorherige Konstruktor befand sich in der Zeilennummer „{0}“.
```
MANISH-K-07, line 6: @Zopsss , looks like you have been copy-pasting entire line into the translator?
The quotes are also been styled i18n for few files.. Please look into such

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionConstructorsDeclarationGroupingTest.java
Zopsss: Done. Added new test for records. Here it is https://github.com/checkstyle/checkstyle/pull/14749/files#diff-7ed07ebc16b7c739ede9580361207c06573eef922d5c593acac87b48c315ac58

And also renamed the test methods to testClass, testEnum, testRecords.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
Zopsss: done. updated the comment as below:

https://github.com/checkstyle/checkstyle/blob/34095f7f434f46efe609d76b692db8175140abca/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java#L32-L35

Zopsss:
GitHub, generate site

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_de.properties
Zopsss: Fixed. thanks for pointing this out.

Zopsss:
> Did you share with us how it is failing on your local?

I guess there was some issue with Maven. I restarted the laptop and then it started working fine.

config/checkstyle-non-main-files-suppressions.xml
```diff
@@ -234,6 +234,8 @@
          files="src[\\/]xdocs[\\/]checks[\\/]coding[\\/]nofinalizer.xml.template"/>
   <suppress id="propertiesMacroMustExist"
          files="src[\\/]xdocs[\\/]checks[\\/]coding[\\/]overloadmethodsdeclarationorder.xml.template"/>
+  <suppress id="propertiesMacroMustExist"
+            files="src[\\/]xdocs[\\/]checks[\\/]coding[\\/]constructorsdeclarationgrouping.xml.template"/>
```
romani, line 238: Align indentation of attribute with other lines like this

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
romani: Ok

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
romani: Ok

romani:
GitHub, generate report

config/checkstyle-non-main-files-suppressions.xml
Zopsss: done

Zopsss:
> GitHub, generate report

I've generated the regression reports above, here are those:

https://github.com/checkstyle/checkstyle/pull/14749#issuecomment-2050078570
https://github.com/checkstyle/checkstyle/pull/14749#issuecomment-2050161180

there were no false positives/negatives found. The report was fine.

Zopsss:
> GitHub, generate report

I've generated the regression reports above, here are those:

https://github.com/checkstyle/checkstyle/pull/14749#issuecomment-2050078570
https://github.com/checkstyle/checkstyle/pull/14749#issuecomment-2050161180

there were no false positives/negatives found. The report was fine. @romani

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_de.properties
nrmancuso: Constructors are "overloaded" by definition, we don't need to include this term in the message at all.

Zopsss: Yes, I've removed the term "overloaded" from everywhere including the error message. Here's the new error message:

https://github.com/checkstyle/checkstyle/blob/2c8f77f847bc4b8b71910ff62b61ebab5a8a2fb9/src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties#L6

I've also updated error message in other languages too.

Zopsss:
@romani

Zopsss:
@romani can you please review the PR?

Zopsss:
@nrmancuso If you have time can also you please review this PR? it would be really helpful if you or @romani  reviews this PR

Zopsss:
Hey @nrmancuso thanks for self assigning the PR... Here are the regression reports and documentation reports:

https://github.com/checkstyle/checkstyle/pull/14749#issuecomment-2060505881

I needed to divide the project list into 2 parts because it was giving `Out Of Memory` error for single project list.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
nrmancuso: I am "unresolving" this item to continue discussion. We should be able to make this check stateless by using all class definitions as required tokens (CLASS_DEF_, ENUM_DEF, etc...), then checking class members. This should allow us to make this check stateless, perhaps at the cost of a slightly larger memory footprint. @Zopsss @romani thoughts?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
nrmancuso: Looks like [this](https://github.com/checkstyle/checkstyle/pull/14749#discussion_r1583033202) approach might also mitigate the issue here.

src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/constructorsdeclarationgrouping/InputXpathConstructorsDeclarationGroupingRecords.java
```diff
@@ -0,0 +1,20 @@
...
+      this(a,a);
+    }
+
+    void foo() {}
+
+    public MyRecord {} // warn
+
+    public MyRecord(int x, int y, int z) {
+      this(x+y, z);
+    }
```
nrmancuso, line 18: @Zopsss @romani it looks like we are emulating the behavior of OverloadMethodsDeclarationOrder where a user may need to run the check multiple times in order to make checkstyle happy. Is there some technical limitation that prevents us from using the following logic:

 - First constructor is where constructors should live
 - All other constructors are flagged

?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
Zopsss: > We should be able to make this check stateless by using all class definitions as required tokens (CLASS_DEF_, ENUM_DEF, etc...), then checking class members.

do you mean we should use an approach like this?

https://github.com/checkstyle/checkstyle/blob/55f5c966bfa61b387110db6c4780f259a21676c8/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java#L78-L92

The [OverloadMethodsDeclarationOrderCheck](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java) uses this approach and it is stateless however the `visitToken()` gets called every time for class, enum, interface, etc... Then we're travelling the whole block to check that all the methods are grouped together. By this approach we can make the check stateless but then we will need to travel the whole class, enum, records, etc, we will be checking each and every tokens of the block, this is a time consuming method as we're travelling through the whole block and checking every tokens. 

We're currently calling the Check for only `CTOR` and `COMPACT_CTOR` tokens only, by using this implementation we're saving a lot of time instead of travelling the whole block. This is an optimized method to check that all constructors are grouped together or not.

We're also clearing the HashMap every time before calling the `visitToken()`, so the HashMap does not contain any unnecessary data:

https://github.com/checkstyle/checkstyle/blob/cadb0dc3a4635dd53eaa24ad6d41fd6e2cef7fd2/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java#L83-L86 

If we make the current implementation stateless then we cannot use this optimized approach. The check will take more time to execute. So I guess if we're able to make the Check optimized then we should be using the current approach. The Check implementation will also become complex if we use the OverloadMethodsDeclarationOrder's approach. Correct me if I'm wrong.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
Zopsss: Yeah by using that approach we will be operating the check statelessly but that approach will be more time consuming, am I right? I've given my opinion [here](https://github.com/checkstyle/checkstyle/pull/14749#discussion_r1583106507)

src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/constructorsdeclarationgrouping/InputXpathConstructorsDeclarationGroupingRecords.java
Zopsss: yes we can do this, I agree with this. Waiting for @romani's opinion. If he agrees then I'll start working on the new implementation

rnveach: @nrmancuso @romani If this is the way to want to move, we should probably create an issue for the other check.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
rnveach: Sorry, I didn't read all this but just a few comments. We aren't multithreaded any time soon, so its not a big deal if we can't change the type of check there. Since we don't have an implementation of multithreading there is no telling how big a performance there may be. I don't see stateless being a big performance improvement. It just means we don't have to instantiate the check for each new file it processes. While new instances are cheap, it really depends on how we will do that instantiation and populate any property values.

src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/constructorsdeclarationgrouping/InputXpathConstructorsDeclarationGroupingRecords.java
Zopsss: @nrmancuso I've changed the implementation according to your new idea. Should I push the new changes or wait for other maintainers approval?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java
Zopsss: After reading @rnveach's comment, do we still need to transition to the stateless check or is it okay to keep this check stateful? @nrmancuso @romani please tell me what do you think about this?

src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/constructorsdeclarationgrouping/InputXpathConstructorsDeclarationGroupingRecords.java
nrmancuso: > @nrmancuso I've changed the implementation according to your new idea. Should I push the new changes or wait for other maintainers approval?

If you’ve already done the work, you can checkout a new branch from what you’ve done, and open a draft PR. Then you can reset this branch from the origin. This way you have both versions preserved and we can see what the other looks like.

Zopsss: Done. Here is the draft PR: https://github.com/checkstyle/checkstyle/pull/14844

nrmancuso: >Done. Here is the draft PR: https://github.com/checkstyle/checkstyle/pull/14844

This new PR makes it even more obvious that we should do [this](https://github.com/checkstyle/checkstyle/pull/14749#discussion_r1583033202), IMO.

This is my vision:
1. Create a list of all children of the class/enum/whatever def
2. Create sublist from first constructor , taking while the AST is a constructor
3. Take the difference of the original list and the sublist
4. Filter out non-constructors
5. log all remaining items

Zopsss: made the implementation stateless on that PR, can you please check that out? and sorry for the late reply I wasn't active for few days

Zopsss:
The work on this module will continue here: https://github.com/checkstyle/checkstyle/pull/14844

The module implementation has been changed based on this [comment](https://github.com/checkstyle/checkstyle/pull/14749#discussion_r1586244932).

Issue #14599, opened 2024-03-03
Parsing performance regression testing


nrmancuso:
As discovered at https://github.com/checkstyle/checkstyle/issues/14566 (and elsewhere previously), parsing performance is a critical, user-facing metric that we should be consistently measuring and verifying. We have discussed this topic previously, but no action has been taken and we released a "bad" version to users.

A quick hack can be to time the execution of OpenJDK no-error CI tasks; however, this can end up bring pretty inconsistent due to the fact that we download the repo to be parsed during the report generation. Additionally, variables like the amount of resources allocated to the JVM, etc. play a role.

A slightly better approach would be to improve our AST regression testing to show checkstyle execution time as a difference between master and the feature branch (as a percentage), since this will be executed on the same machine, sequentially, and have identical resource allocation.

Probably the best approach would be to create a new script to time only checkstyle execution, and capture other metrics like memory usage, etc, without all the I/O involved with report generation.

For example, our generated report could tell us that parsing was 33% faster on the feature branch, used 10% less memory, etc.

Lmh-java:
@nrmancuso I would like to give it a shot. But, I want to discuss a few things about this to get more insights before everything starts.

I want to follow the **best approach**. Therefore, I will have to introduce a new test task and create a new script that benchmarks multiple metrics for checkstyle execution. I think it would be better if we also integrate this into the CI (just like how we use the "GitHub, generate report" command).

To benchmark, ideally, it would be beneficial for us to run Checkstyle on some large projects that take relatively longer execution time. Do you have some suggestions on which project should we use as the benchmark sample?

PS: if any of my understanding above is inaccurate, please correct me :)

nrmancuso:
>To benchmark, ideally, it would be beneficial for us to run Checkstyle on some large projects that take relatively longer execution time. Do you have some suggestions on which project should we use as the benchmark sample? Or do we want to do it over many different projects? (In this case, this will be a part of the original regression test)

The latest openjdk has proven to be a good codebase to test on for both performance and memory consumption considerations.

Lmh-java:
I will start right now to build a new github action task for measuring both performance and memory consumption on openJDK.

nrmancuso:
> I will start right now to build a new github action task for measuring both performance and memory consumption on openJDK.

I would suggest to map out your high level plan here to make sure we are all on the same page.

Lmh-java:
# Draft Plan

## Goal
[...]

PR #14754, opened 2024-04-02
Issue #14599: Create performance regression test CI

Lmh-java:
Draft for Issue #14599

### DIFF
--- a/.ci/check-performance-regression.sh
+++ b/.ci/check-performance-regression.sh
@@ -1,0 +1,103 @@
+#!/bin/bash
+
+set -e
+
+# max difference tolerance in %
+THRESHOLD_PERCENTAGE=10
+# baseline of execution time in seconds
+BASELINE_SECONDS=415.72
+
+# JDK version
+JDK_VERSION=17
+# sample project path
+SAMPLE_PROJECT="./.ci-temp/jdk$JDK_VERSION"
+# suppression file
+SUPPRESSION_FILE="./config/projects-to-test/openjdk$JDK_VERSION-excluded.files"
+# config file
+CONFIG_FILE="./config/benchmark-config.xml"
+
+# execute a command and time it
+# $TEST_COMMAND: command being timed
+time_command() {
+  # execute the command and time it
+  /usr/bin/time -o time.temp -q -f "%e" "$@" &>result.tmp
+
+  cat time.temp
+}
+
+# execute the benchmark a few times to calculate the average metrics
+# $JAR_PATH: path of the jar file being benchmarked
+execute_benchmark() {
+  local JAR_PATH=$1
+  if [ -z "$JAR_PATH" ]; then
+      echo "Missing JAR_PATH as an argument."
+      exit 1
+  fi
+
+  local TOTAL_SECONDS=0
+  local NUM_EXECUTIONS=3
+
+  [ ! -d "$SAMPLE_PROJECT" ] &&
+    echo "Directory $SAMPLE_PROJECT DOES NOT exist." | exit 1
+
+  # add suppressions to config file
+  sed -i "/  <!-- Filters -->/r $SUPPRESSION_FILE" \
+        $CONFIG_FILE
+
+  for ((i = 0; i < NUM_EXECUTIONS; i++)); do
+    local CMD=(java -jar "$JAR_PATH" -c "$CONFIG_FILE" \
+      -x .git -x module-info.java "$SAMPLE_PROJECT")
+    local BENCHMARK=($(time_command "${CMD[@]}"))
+    TOTAL_SECONDS=$(echo "$TOTAL_SECONDS + $BENCHMARK" | bc)
+  done
+
+  # average execution time in patch
+  local AVERAGE_IN_SECONDS=$(echo "scale=2; $TOTAL_SECONDS / $NUM_EXECUTIONS" | bc)
+  echo "$AVERAGE_IN_SECONDS"
+}
+
+# compare baseline and patch benchmarks
+# $EXECUTION_TIME_SECONDS execution time of the patch
+compare_results() {
+  local EXECUTION_TIME_SECONDS=$1
+  if [ -z "$EXECUTION_TIME_SECONDS" ]; then
+        echo "Missing EXECUTION_TIME_SECONDS as an argument."
+        exit 1
+    fi
+  # Calculate percentage difference for execution time
+  local DEVIATION_IN_SECONDS=$(echo "scale=4; \
+    ((${EXECUTION_TIME_SECONDS} - ${BASELINE_SECONDS}) / ${BASELINE_SECONDS}) * 100" | bc)
+  echo "Execution Time Difference: $DEVIATION_IN_SECONDS%"
+
+  # Check if differences exceed the maximum allowed difference
+  if (( $(echo "$DEVIATION_IN_SECONDS > $THRESHOLD_PERCENTAGE" | bc -l) )); then
+    echo "Difference exceeds the maximum allowed difference (${DEVIATION_IN_SECONDS}% \
+     > ${THRESHOLD_PERCENTAGE}%)!"
+    exit 1
+  else
+    echo "Difference is within the maximum allowed difference (${DEVIATION_IN_SECONDS}% \
+     <= ${THRESHOLD_PERCENTAGE}%)."
+    exit 0
+  fi
+}
+
+# package patch
+mvn -e --no-transfer-progress -Passembly,no-validations package
+
+# start benchmark
+echo "Benchmark launching..."
+AVERAGE_IN_SECONDS="$(execute_benchmark "$(find "./target/" -type f -name "checkstyle-*-all.jar")")"
+
+# print the command execution result
+echo "================ MOST RECENT COMMAND RESULT ================="
+cat result.tmp
+
+echo "===================== BENCHMARK SUMMARY ===================="
+echo "Execution Time Baseline: ${BASELINE_SECONDS} s"
+echo "Average Execution Time: ${AVERAGE_IN_SECONDS} s"
+echo "============================================================"
+
+# compare result with baseline
+compare_results "$AVERAGE_IN_SECONDS"
+
+exit $?
--- a/.github/workflows/check-performance-regression.yml
+++ b/.github/workflows/check-performance-regression.yml
@@ -1,0 +1,54 @@
+#####################################################################################
+# GitHub Action to test performance regression.
+#
+# Workflow starts when:
+# 1) push to master
+# 2) PR created or pushed
+#
+#####################################################################################
+name: Check-Performance-Regression
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches: '*'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    if: github.repository == 'checkstyle/checkstyle'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up JDK 11
+        uses: actions/setup-java@v4
+        with:
+          java-version: 11
+          distribution: 'temurin'
+
+      - name: Checkout patch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          path: ./checkstyle
+
+      - name: Clone JDK 17 Repo
+        uses: actions/checkout@v4
+        with:
+          repository: openjdk/jdk17
+          path: ./checkstyle/.ci-temp/jdk17
+
+      - name: Setup local maven cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.m2/repository
+          key: checkstyle-maven-cache-${{ hashFiles('**/pom.xml') }}
+
+      - name: Run performance test
+        run: |
+          cd checkstyle
+          bash ./.ci/check-performance-regression.sh
--- a/config/benchmark-config.xml
+++ b/config/benchmark-config.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+  "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
+<module name="Checker">
+  <!-- Filters -->
+
+  <property name="severity" value="ignore"/>
+  <module name="TreeWalker">
+    <module name="NoCodeInFileCheck">
+    </module>
+  </module>
+</module>


.ci/baseline.benchmark
```diff
@@ -0,0 +1,15 @@
...
+Running benchmark 2/3...
+================== BENCHMARK RESULT #2 ==================
+Execution Time: 8.37 s
+============================================================
+Running benchmark 3/3...
+================== BENCHMARK RESULT #3 ==================
+Execution Time: 8.50 s
+============================================================
+===================== BENCHMARK SUMMARY ====================
+Average Execution Time: 8.56 s
```
Lmh-java, line 14: This baseline file is generated by the shell script and is parsable by the same shell script. So whenever we want to change the baseline, we can use the same tool to generate this.

config/checkstyle-non-main-files-checks.xml
```diff
@@ -19,7 +19,7 @@
   <module name="RegexpOnFilename">
     <property name="id" value="executablesLocation"/>
     <property name="folderPattern" value="[\\/].ci([\\/]|$)"/>
-    <property name="fileNamePattern" value="\.(bat|cmd|groovy|sh|md)$"/>
+    <property name="fileNamePattern" value="\.(bat|cmd|groovy|sh|md|benchmark)$"/>
```
Lmh-java, line 22: Exclude `benchmark` files since they must are located in `.ci` folder but are not executable.

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,48 @@
...
+    echo "Differences exceed the maximum allowed difference (${MAX_DIFFERENCE}%)!"
+    exit 1
+  else
+    echo "Differences are within the maximum allowed difference."
+    exit 0
+  fi
+}
+
+# parse baseline benchmark
+BASELINE=($(parse_benchmark_result "./.ci/baseline.benchmark"))
```
Lmh-java, line 35: We may want to read the baseline from the `master` branch later after this purposal is accepted.

.ci/run-benchmark.sh
```diff
@@ -0,0 +1,69 @@
...
+SAMPLE_PROJECT="./.ci/jdk22"
+
+# generate benchmark config file
+generate_config() {
+  local CONFIG="<?xml version=\"1.0\"?>
+<!DOCTYPE module PUBLIC \"-//Puppy Crawl//DTD Check
+Configuration 1.3//EN\"
+\"http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd\">
+<module name=\"Checker\">
+  <module name=\"TreeWalker\"/>
```
romani, line 22: https://github.com/checkstyle/checkstyle/blob/b0c91b99d18f64eec419247010a7187120eedaca/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java#L153

We need at least one module.

Please make this config as static file.

.ci/baseline.benchmark
```diff
@@ -0,0 +1,15 @@
+Running benchmark 1/3...
```
romani, line 1: This is overengineering.
This file should have only one digit `8.56`
And probably no files , and just variable in shell file that defined our baseline value.

.ci/run-benchmark.sh
```diff
@@ -0,0 +1,69 @@
...
+    echo "Directory $SAMPLE_PROJECT DOES NOT exists." | exit 1
+
+  generate_config
+  for ((i = 1; i <= NUM_RUNS; i++)); do
+    echo "Running benchmark ${i}/${NUM_RUNS}..."
+    local CMD="java -jar $JAR_PATH -c ./.ci/benchmark-config.xml -x .git $SAMPLE_PROJECT"
+    local BENCHMARK=($(time_command "$CMD"))
+    TOTAL_TIME=$(echo "$TOTAL_TIME + ${BENCHMARK[0]}" | bc)
+    echo "================== BENCHMARK RESULT #${i} =================="
+    echo "Execution Time: ${BENCHMARK[0]} s"
```
romani, line 55: Remove all this output, just focus on accumulation time values to summ them.

romani (changes requested):
Items

.ci/baseline.benchmark
Lmh-java: Sure, I will make it a variable inside the script to avoid adding additional file type check suppression.

.ci/run-benchmark.sh
rnveach: Shouldn't we choose a low cost module to be included? "Low cost" meaning a module which does extremely little work. Maybe just a quick visit and no begin/end.

Any module we add will be included in the performance test and will skew the results of calculating the grammar itself.

If we do this, we should add a warning comment too.

Lmh-java: Yes, if we only want to benchmark the grammar parsing, using a low cost module would definitely be better. I assume [`ConstantName`](https://checkstyle.sourceforge.io/checks/naming/constantname.html) is a good one, since it only runs regex on texts. Do you have recommandations on better modules in our case? I can even create an empty check just for testing purpose later if you think it would be more accurate and more helpful.

rnveach: I do not have a recommendation currently off the top of my head. I wouldn't consider regexp low cost, especially if it executes the regexp match, but if there is nothing lower than that is fine.

> create an empty check just for testing purpose 

I am fine with this, but I would look for approval from @romani .

Lmh-java: Thanks for you advice, I will try `NodeCodeInFileCheck` for now. 😄

romani: Yes, https://checkstyle.org/checks/misc/nocodeinfile.html#NoCodeInFile is awesome candidate.

We also use https://github.com/checkstyle/contribution/blob/cdeb66c6bc719eb0cf27a3cc0aadafbe2c016a47/checkstyle-tester/my_check.xml#L23

.ci/run-benchmark.sh
Lmh-java: done

.ci/baseline.benchmark
Lmh-java: done

.ci/run-benchmark.sh
Lmh-java: done, I have also added all necessary filters to prevent checkstyle stops with errors

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,103 @@
...
+  # shellcheck disable=SC2086
+  local TIME_OUTPUT=$(command time -p $TEST_COMMAND 2>&1)
+  # Extract execution time
+  local EXECUTION_TIME=$(echo "$TIME_OUTPUT" | awk '/real/ {print $2}')
+
+  echo "${EXECUTION_TIME}"
+}
+
+# insert all the filters into the config file
+generate_config() {
```
romani, line 32: No generation is required, just make config as you need and store it in git

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,103 @@
...
+export MAVEN_OPTS='-Xmx2000m'
+mvn -e --no-transfer-progress -Passembly,no-validations package
+
+generate_config
+
+# run benchmark
+echo "Benchmark launching..."
+AVG_TIME="$(run_benchmark "$(find "./target/" -type f -name "checkstyle-*-all.jar")")"
+echo "===================== BENCHMARK SUMMARY ===================="
+echo "Average Execution Time: ${AVG_TIME} s"
```
romani, line 98: Please print here EXECUTION_TIME_BASELINE.

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,103 @@
...
+# $TEST_COMMAND: command being timed
+time_command() {
+  # Run the command with time
+  local TEST_COMMAND=$1
+  if [ -z "$TEST_COMMAND" ]; then
+        echo "Missing TEST_COMMAND as an argument."
+        exit 1
+    fi
+  # shellcheck disable=SC2086
+  local TIME_OUTPUT=$(command time -p $TEST_COMMAND 2>&1)
```
romani, line 24: Why disabled?

romani (changes requested):
Items

.ci/check-performance-regression.sh
Lmh-java: done by using an array according to https://www.shellcheck.net/wiki/SC2086

.ci/check-performance-regression.sh
Lmh-java: done

.ci/check-performance-regression.sh
Lmh-java: done

.github/workflows/check-performance-regression.yml
```diff
@@ -0,0 +1,53 @@
...
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          path: ./checkstyle
+
+      - name: Clone JDK 17 Repo
+        uses: actions/checkout@v4
+        with:
+          repository: openjdk/jdk17
+          path: ./checkstyle/.ci/jdk17
```
romani, line 42: Please make a folder `.ci-temp` on that CI host.

And do  `./checkstyle/.ci-temp/jdk17`

We have a rule to not change repository by CI tasks. This folder already knows by gitignore file

romani (changes requested):
Good  job. Last item:

.github/workflows/check-performance-regression.yml
Lmh-java: done

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,91 @@
...
+# compare baseline and patch benchmarks
+# $EXECUTION_TIME execution time of the patch
+compare_results() {
+  local EXECUTION_TIME=$1
+  if [ -z "$EXECUTION_TIME" ]; then
+        echo "Missing EXECUTION_TIME as an argument."
+        exit 1
+    fi
+  # Calculate percentage difference for execution time
+  local EXECUTION_TIME_DIFFERENCE=$(echo "scale=4; \
```
nrmancuso, line 61:
```suggestion
  local DEVIATION_IN_SECONDS=$(echo "scale=4; \
```

config/benchmark-config.xml
nrmancuso, line 1: Why can't we reuse the existing file filter file?

config/jsoref-spellchecker/whitelist.words
nrmancuso, line 1: If we reuse the existing file filter file, we can avoid extension of this list.

Lmh-java:
Thanks for your response. It might take longer to apply all these changes than I usually do. I am having some final exams this week. I am still working on this. Sorry for the delay.

nrmancuso:
> Thanks for your response. It might take a little bit longer to apply all these changes than usual. I am having some final exams this week. I am still working on this. Sorry for the delay.

No worries :) We are all busy people trying our best. Thanks for your contributions!

config/jsoref-spellchecker/whitelist.words
Lmh-java: done

Lmh-java: done

config/benchmark-config.xml
Lmh-java: Removed the duplicated content and use `sed` to insert the file filter list. After the new property is introduced to treewalker (#14779), we can use that to reduce additional time cost.

.ci/check-performance-regression.sh
Lmh-java: done

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,91 @@
...
+      exit 1
+  fi
+
+  local TOTAL_TIME=0
+  local NUM_RUNS=3
+
+  [ ! -d "$SAMPLE_PROJECT" ] &&
+    echo "Directory $SAMPLE_PROJECT DOES NOT exists." | exit 1
+
+  for ((i = 1; i <= NUM_RUNS; i++)); do
```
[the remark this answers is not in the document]
Lmh-java: done

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,91 @@
...
+  if [ -z "$JAR_PATH" ]; then
+      echo "Missing JAR_PATH as an argument."
+      exit 1
+  fi
+
+  local TOTAL_TIME=0
+  local NUM_RUNS=3
+
+  [ ! -d "$SAMPLE_PROJECT" ] &&
+    echo "Directory $SAMPLE_PROJECT DOES NOT exists." | exit 1
```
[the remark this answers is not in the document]
Lmh-java: done

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,91 @@
...
+
+# sample project path
+SAMPLE_PROJECT="./.ci-temp/jdk17"
+CONFIG_FILE="./config/benchmark-config.xml"
+
+# run a command and time it
+# $TEST_COMMAND: command being timed
+time_command() {
+  # Run the command with time
+  local TIME_OUTPUT=$(command time -p "$@" 2>&1)
```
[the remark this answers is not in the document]
Lmh-java: done

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,91 @@
...
+
+  for ((i = 1; i <= NUM_RUNS; i++)); do
+    local CMD=(java -jar "$JAR_PATH" -c "$CONFIG_FILE" \
+      -x .git -x module-info.java "$SAMPLE_PROJECT")
+    local BENCHMARK=($(time_command "${CMD[@]}"))
+    TOTAL_TIME=$(echo "$TOTAL_TIME + ${BENCHMARK}" | bc)
+  done
+
+  # average execution time in patch
+  local AVG_TIME=$(echo "scale=2; $TOTAL_TIME /   $NUM_RUNS" | bc)
```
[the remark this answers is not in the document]
Lmh-java: done

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+set -e
+
+# max difference tolerance in %
+MAX_DIFFERENCE=10
+# baseline of execution time in second
+EXECUTION_TIME_BASELINE=415.72
```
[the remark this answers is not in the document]
Lmh-java: done

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+set -e
+
+# max difference tolerance in %
+MAX_DIFFERENCE=10
```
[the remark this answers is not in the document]
Lmh-java: done

Lmh-java:
> Show evidence of testing (prove that this will actually fail if we have an execution time that is an outlier). In order to do this, you can make a temporary commit that rebases this PR on the commit before we merged https://github.com/checkstyle/checkstyle/pull/14568

I am working on this right now.

Lmh-java:
@nrmancuso The CI fails in https://github.com/Lmh-java/checkstyle/pull/6.

I checkout 38708754639278abe108c37bef025e76c38e7882 (The commit before https://github.com/checkstyle/checkstyle/pull/14568), and cherry-picked https://github.com/checkstyle/checkstyle/pull/14754 for the test.

Lmh-java:
@nrmancuso, by the way, do you think it would be better if we add a line of
```yml
    if: github.repository == 'checkstyle/checkstyle'
```
to this workflow? so that we can make sure it only runs on our end, instead of on the developer ends.

nrmancuso:
> @nrmancuso, by the way, do you think it would be better if we add a line of
> 
> ```yaml
>     if: github.repository == 'checkstyle/checkstyle'
> ```
> 
> to this workflow? so that we can make sure it only runs on our end, instead of on the developer ends.

Yeah, we have done this in most others, I agree.

Lmh-java:
> > @nrmancuso, by the way, do you think it would be better if we add a line of
> > ```yaml
> >     if: github.repository == 'checkstyle/checkstyle'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to this workflow? so that we can make sure it only runs on our end, instead of on the developer ends.
> 
> Yeah, we have done this in most others, I agree.

Done

.ci/check-performance-regression.sh
```diff
@@ -0,0 +1,98 @@
...
+  fi
+
+  local TOTAL_SECONDS=0
+  local NUM_EXECUTIONS=3
+
+  [ ! -d "$SAMPLE_PROJECT" ] &&
+    echo "Directory $SAMPLE_PROJECT DOES NOT exist." | exit 1
+
+  # add suppressions to config file
+  sed -i '/  <!-- Filters -->/r ./config/projects-to-test/openjdk17-excluded.files' \
```
nrmancuso, line 39: Please extract this filename to a variable declared above with SAMPLE_PROJECT, etc., and do anything else to make it easy/obvious to update this script to run on any java version.

@Lmh-java how did we settle on 17, instead of 21?

nrmancuso (changes requested):
Last from me:

.ci/check-performance-regression.sh
nrmancuso: I am good with 17 for now, just curious about why we did this.

Lmh-java: I am still working on #14779 right now. If that's merged, we can directly use that new property to skip all exceptions. That will be faster than a long list of suppressions :).

Lmh-java: Done

nrmancuso (approved):
Good stuff! Thanks a lot @Lmh-java this will be a big help during syntax updates.

Issue #13758, opened 2023-09-21
ImportOrder: replace Pattern[] with String[] type for properties staticGroups and groups


romani:
from https://github.com/checkstyle/checkstyle/pull/13754#issuecomment-1726796580

We think our documentation at https://checkstyle.org/property_types.html#Pattern.5B.5D is either wrong or misleading.

Found looking at https://checkstyle.org/checks/metrics/classfanoutcomplexity.html#Properties

> The string representation is parsed as a set of forward slash ('/') separated patterns.

When glancing at this, i thought it meant each pattern is separated by a `/`.

But looking at the code at https://github.com/checkstyle/checkstyle/blob/5a0f923a6d48a9152b963468bfece7b6e9f28295/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/AbstractClassCouplingCheck.java#L160 , it takes input as a string array, which is separated by the comma and not the `/`.

Even `createPattern` method that it calls does nothing with `/`, so I don't understand where the `/` is coming from.

https://github.com/search?q=repo%3Acheckstyle%2Fcheckstyle+%22Pattern.5B.5D%22+path%3A%2F%5Esrc%5C%2Fxdocs%5C%2Fchecks%5C%2F%2F&type=code
We have 6 checks with this type.

It looks like documentation is coming from https://github.com/checkstyle/checkstyle/blob/5a0f923a6d48a9152b963468bfece7b6e9f28295/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L759 , which is a serious problem now as we have a documented type being parsed differently in different checks, meaning we can't even unify documentation without first changing production code.

------

We need review all Patern[] usage and probably move  ImportOrderCheck to have special type.
how we can support list of Pattern, when "," can be used in regexp.

-----

agreement on how to fix we archived at https://github.com/checkstyle/checkstyle/pull/14764#issuecomment-2041485231

---------

**Migration Notes**:

`ImportOrder` is update to replace `Pattern[]` with `String[]` type for properties `staticGroups` and `groups`.
user who use xml config are not affected, but plugins who might have type validation during Check configuration (before Check execution) might be affected. Validation type in metadata is different from 10.17.0 for this Check.

Lmh-java:
Hi, I would give a try on this issue.

Lmh-java:
[...]

PR #14740, opened 2024-03-30
Issue #13758: Add pattern array converter for auto bean

Lmh-java:
For Issue #13758.

1. Create auto configuration for `Pattern[]` with a fixed separator (in this PR)
2. Update `AbstractClassCouplingCheck` to use this new property type (in this PR)
3. Update `ImportOrderCheck` to replace use of `Pattern[]` with `String[]` (In this PR)
------

Agreement on update is archived at https://github.com/checkstyle/checkstyle/pull/14764#issuecomment-2041485231

PR #14764, opened 2024-04-06
Issue #13758: Fix documentation of pattern array

Lmh-java:
Issue #13758: Fix documentation of pattern array
Reword documentation for `Pattern[]` as discussed in https://github.com/checkstyle/checkstyle/issues/13758#issuecomment-2040166442

### DIFF
--- a/src/xdocs/property_types.xml
+++ b/src/xdocs/property_types.xml
@@ -673,7 +673,7 @@
     <section name="Pattern[]">
       <p>
         This type represents a set of patterns. The string representation
-        is parsed as a set of forward slash ('/') separated patterns.
+        is parsed as a set of comma (',') separated patterns.
       </p>
     </section>
 


romani (approved):
Ok to merge

rnveach (changes requested):
I am not sure why this is split from https://github.com/checkstyle/checkstyle/pull/14740 . It seems they are inter-connected based on what is decided on the other PR. This change also being in that PR and discussed in this PR's first post seems to align with that.

If minds are changed in the other PR, it will impact this one.

rnveach:
Documentation is still misleading as we don't support `Pattern[]` in our setter(s). `AbstractAutomaticBean` only has `Pattern` and not `Pattern[]`.

rnveach:
Also if this updated documentation is correct, please show that all current fake implementations follow it. This area is global documentation, so everything has to follow it otherwise the documentation is wrong.

https://checkstyle.org/checks/imports/importorder.html#Properties
> Specify list of type import groups. Every group identified either by a common prefix string, or by a regular expression enclosed in forward slashes (e.g. /regexp/). 

Wouldn't this mean this documentation is wrong since every group is a comma? Not really sure what a common prefix string is.

rnveach:
> leave the update for Pattern[] delimiter for the future, since it will break the backward compatibility

Yes suggestion in issue is breaking compatibility. There is no break in https://github.com/checkstyle/checkstyle/pull/14740 which I see.

Lmh-java:
>If this updated documentation is correct, please show that all current fake implementations follow it

Reply to https://github.com/checkstyle/checkstyle/pull/14764#issuecomment-2041146853:

By searching through GitHub [link](https://github.com/search?q=repo%3Acheckstyle%2Fcheckstyle+%22Pattern.5B.5D%22+path%3A%2F%5Esrc%5C%2Fxdocs%5C%2Fchecks%5C%2F%2F&type=code), I only see 3 checks using `Pattern[]` right now. Here is a list:
1. ImportOrder
`groups` and `staticGroups` property.

These two properties are special, since they have a mix of plain string and regex expression. Because of that, these two properties require users to enclose regex expression with `/`. Plain string represents the so-called "common prefix string", which is the common package name. For example, `java.lang.util` and `java.lang.annotation` share the "common prefix string" of "java.lang". Eventually, these plain strings are converted to regex expression as the code shown below.
https://github.com/checkstyle/checkstyle/blob/e30260d88bfc5ae5ff110560e28371a8fd8074ba/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L747-L777

IMO, this design is somehow over-engineering, since we can directly tell users to add a `^` and a `.` to encolse the "common prefix string" to turn the plain string into a regex (also to avoid confusing), instead of mixing regex and plain string together. However, such update will break the compatibility.

3. ClassFanOutComplexity
`excludeClassesRegexps` property (inherited from ClassDataAbstractionCouplingCheck).
This is separated by comma.
4. ClassDataAbstractionCoupling
`excludeClassesRegexps ` property (inherited from ClassDataAbstractionCouplingCheck).
This is separated by comma.

romani:
ImportOrder is using comma separated list, and each element of it additionally special in interpretation, that is described in property description, so it is ok, not ideal design, but ok.

rnveach:
I think it is better to go with https://github.com/checkstyle/checkstyle/pull/14740 then confuse users between this being a string or a pattern.

romani:
>then confuse users between this being a string or a pattern.

if property is simple patterns, so Pattern[] is not confusing.
but this Check is not simple Pattern[], it is very special interpretation of Strings, that mostly Pattern for each element.n 
If property description explains all detail of type/data interpretation it is all that wee need.
Such types are useful only for UI plugins which can do extra validation of content of property to give validation before config is created+executed.

rnveach:
https://github.com/checkstyle/checkstyle/pull/14740/files is a simple pattern. If we need to change only `ImportOrderCheck` to be `String[]` type because it is so special, then I am fine with that.

> PR with make Pattern[], as first class type - https://github.com/checkstyle/checkstyle/pull/14740 , such type will have design problem when , is used in regexp intentionally.

This is already a design problem before that PR, so that PR has no negatives and still the positive that we support what we say we do, `Pattern[]`.

romani:
ok as we agree on update in package, not individually

@Lmh-java , please make such updates in one PR to make it sure that all updates will be done in same time and all see all changes together, they are not big changes, so single PR is fine.

romani:
lets put all changes to https://github.com/checkstyle/checkstyle/pull/14740

Issue #14765, opened 2024-04-07
OverloadMethodDeclarationOrder check ignores anything besides methods


Zopsss:
I have read check documentation: https://checkstyle.org/checks/coding/overloadmethodsdeclarationorder.html
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

The [OverloadMethodDeclarationOrderCheck](https://checkstyle.org/checks/coding/overloadmethodsdeclarationorder.html) gives us error if there is another method between the overloaded methods. But it does not gives error when there is anything else besides method between overloaded methods like a variable, array, inner interface, inner class, etc.

This is how the check works currently:

#### Test.java
```java
public class Test {
    void test() {}

    void test(int x) {}

    private interface Testing {}

    String str;

    void test(String str) {}
}
```

#### config.xml
```xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
    <module name="TreeWalker">
        <module name="OverloadMethodsDeclarationOrder"/>
    </module>
</module>
```

#### Output
```
➜ java -jar ../checkstyle-10.15.0-all.jar -c config.xml Test.java
Starting audit...
Audit done.
```

The check should be giving error for the last method
```java
void test(String str) {}
```

---

**Describe what you expect in detail.**

The check should be give an error something like this for the above example:

```
➜ java -jar ../checkstyle-10.15.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] C:\checkstyle testing\OverloadMethods\Test.java:10:5: 
All overloaded methods should be placed next to each other. Placing non-overloaded methods
 in between overloaded methods with the same type is a violation. 
Previous overloaded method located at line '4'. [OverloadMethodsDeclarationOrder]
Audit done.
Checkstyle ends with 1 errors.
```

---

Zopsss:
I'm working on it!

nrmancuso:
Closed via #14766

PR #14766, opened 2024-04-08
Issue #14765: fix OverloadMethodsDeclarationOrderCheck

Zopsss:
Issue #14765

- [x] Fixes OverloadMethodsDelcarationCheck. Now the module gives error when there is variable, interface, class, etc between overloaded methods.
- [x] Added new test cases
- [x] Updated Documentation 

---

Diff Regression config: https://gist.githubusercontent.com/Zopsss/80b98438b46ad1111cf47c520e8ea669/raw/d98d1c801e2c5abd3e6e8834b749dbc95769db53/OverloadMethodsDeclarationOrderCheck.xml

Diff Regression projects: https://gist.githubusercontent.com/Zopsss/22adadb570e4deb95296917244c580b3/raw/29ea756eada8915637b76678db4f46048c198808/projects-to-test-on-for-github-action.properties


Zopsss:
GitHub, generate site

Zopsss:
GitHub, generate report

Zopsss:
GitHub, generate report

Zopsss:
GitHub, generate report

Zopsss:
The website & report looks good to me. 

@romani can you please review the PR?

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/Example2.txt
```diff
@@ -12,4 +12,6 @@ public void foo(String s) {} // OK
 public void notFoo() {} // violation. Have to be after foo(String s, int i)
 public void foo(int i, String s) {}
 public void foo(String s, int i) {}
+private interface Testing() {}
+public void foo(double d) {} // violation. Have to be after foo(String s, int i)
```
romani, line 16: All examples should be same by structure , please add such lines to Example1 in correct order/grouping.

romani (changes requested):
item:

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/Example2.txt
Zopsss: > this looks like false positive

I analyzed the report, there is a semicolon between the method it is giving error for and its above method.

```java
    public void visit(Node.IncludeAction n) throws JasperException {
        JspUtil.checkAttributes("Include action", n, includeActionAttrs,
                err);
        n.setPage(getJspAttribute(null, "page", null, null, n
                        .getAttributeValue("page"), java.lang.String.class, n,
                false));
        visitBody(n);
    };

    public void visit(Node.ForwardAction n) throws JasperException { // violation
        JspUtil.checkAttributes("Forward", n, forwardActionAttrs, err);
        n.setPage(getJspAttribute(null, "page", null, null, n
                        .getAttributeValue("page"), java.lang.String.class, n,
                false));
        visitBody(n);
    }
```

Here you can see the semicolon at the end of the first method which is meaningless and causing the error. So it is not a false positive and our check is working fine. @romani

Zopsss: > All examples should be same by structure , please add such lines to Example1 in correct order/grouping.

done.

Zopsss: > Should we change message ?

I guess we should also change the module name as it contains "order" but we're checking the grouping of the methods not the ordering. The new name can be something like [ConstructorsDeclarationGrouping](https://github.com/checkstyle/checkstyle/issues/14726). What do you think about this?

Zopsss: > Here you can see the semicolon at the end of the first method which is meaningless and causing the error. So it is not a false positive and our check is working fine

@romani ping

romani: >Here you can see the semicolon at the end of the first method which is meaningless and causing the error.

uhh, it is hard to catch but, Check is right. `;` is empty statement
Please add this case to your Inputs, and make a comment over it to explain.

romani:
GitHub, generate website

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/Example2.txt
Zopsss: @romani ping

romani (approved):
Ok to merge.

romani:
GitHub, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -110,7 +110,11 @@ private void checkOverloadMethodsGrouping(DetailAST objectBlock) {
                 final String methodName =
                         currentToken.findFirstToken(TokenTypes.IDENT).getText();
                 final Integer previousIndex = methodIndexMap.get(methodName);
-                if (previousIndex != null && currentIndex - previousIndex > allowedDistance) {
+                final DetailAST previousSibling = currentToken.getPreviousSibling();
+                final boolean isMethod = previousSibling.getType() == TokenTypes.METHOD_DEF;
+
+                if (previousIndex != null
+                        && (currentIndex - previousIndex > allowedDistance || !isMethod)) {
```
rnveach, line 117: It seems to me it would be faster to check the boolean first, then to do the computation, so I would swap them.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder.java
```diff
@@ -60,6 +60,13 @@ public void fooMethod()
         public void overloadMethod(String s, Boolean b, int i) // violation
         {
             //some foo code
+        };
+
+        public void overloadMethod(double d) // violation
+                                            // because there is an unnecessary semicolon at the
+                                            // end of the method
```
rnveach, line 67: where is this semicolon at the end of **this** method? Also, this issue isn't about a semicolon so I am not sure why this specific method is being added.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/Example2.txt
rnveach, line 16: This new stuff to the example isn't completely making sense.

I understand why you added `interface` to both, but there is no example of `foo(double d)` in Example 1. So you lose the reason why you added interface to both if your not keeping everything else in synch too.

This needs to be fixed up a bit more.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
```diff
@@ -110,7 +110,11 @@ private void checkOverloadMethodsGrouping(DetailAST objectBlock) {
                 final String methodName =
                         currentToken.findFirstToken(TokenTypes.IDENT).getText();
                 final Integer previousIndex = methodIndexMap.get(methodName);
-                if (previousIndex != null && currentIndex - previousIndex > allowedDistance) {
+                final DetailAST previousSibling = currentToken.getPreviousSibling();
+                final boolean isMethod = previousSibling.getType() == TokenTypes.METHOD_DEF;
```
rnveach, line 114: What if previous is a comment? Will this cause a conflict?

It doesn't look like it because this check isn't comment aware. However, add some examples of this for proof (IE: overloaded methods right next to each other but a comment between them).

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
romani: This checks is not comment aware, so comments are ignored completely.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
Zopsss: done. here is the new changes:

https://github.com/checkstyle/checkstyle/blob/24ab0e93ad43a656fca70f6a7fd16b9d559bafdf/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java#L116-L117

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder.java
Zopsss: sorry the explanation comment was not written properly. I have explained it properly now:

https://github.com/checkstyle/checkstyle/blob/24ab0e93ad43a656fca70f6a7fd16b9d559bafdf/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder.java#L65-L68

There is a semicolon at the end of the above method which is separating the overloaded methods.  That's why it is causing the violation. We found this test case here: https://github.com/checkstyle/checkstyle/pull/14766#pullrequestreview-1986501881

Mr.Roman suggested to add this to the input test files: https://github.com/checkstyle/checkstyle/pull/14766#discussion_r1561053086

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/Example2.txt
Zopsss: Done. Here is the new changes: https://github.com/checkstyle/checkstyle/blob/24ab0e93ad43a656fca70f6a7fd16b9d559bafdf/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/Example2.txt#L10-L15

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheck.java
Zopsss: > What if previous is a comment? Will this cause a conflict?

No. The check does not care about comments. I have updated the input files and added this test case:

https://github.com/checkstyle/checkstyle/blob/24ab0e93ad43a656fca70f6a7fd16b9d559bafdf/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder.java#L107-L116

Also mentioned this in the documentation examples:

https://github.com/checkstyle/checkstyle/blob/24ab0e93ad43a656fca70f6a7fd16b9d559bafdf/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/Example1.txt#L10-L16

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder.java
rnveach: This comment spanning so many lines isn't much better.

Can we not use the comment like "Violation above" or "Violation below" here?

`at the end of the above method` => `at the end of the previous method`

rnveach:
@romani @nrmancuso @Zopsss 
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/1f7b5dc_2024204006/reports/diff/apache-struts/index.html#A1
> Placing non-overloaded methods in between overloaded methods with the same type is a violation. 

Shouldn't this verbiage be updated to something like `Placing anything in between overloaded methods`?

Zopsss:
> Placing anything in between overloaded methods

I guess this will be a better option as we ignore the comments in this check:

`Placing anything in between overloaded methods, expect comments`

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder.java
Zopsss: done. Here's the new changes:

https://github.com/checkstyle/checkstyle/blob/922cba38624e05652b2d9cb37f13e342d0d4c03b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrder.java#L65-L74

@rnveach

Zopsss:
Why linkcheck-plugin is failing? Even though I changed only the input file? 

https://app.circleci.com/pipelines/github/checkstyle/checkstyle/25181/workflows/01a7d70f-631b-4419-9430-902834a1d7e5/jobs/574997/steps

Zopsss:
Why linkcheck-plugin is failing? Even though I changed only the input file?

https://app.circleci.com/pipelines/github/checkstyle/checkstyle/25181/workflows/01a7d70f-631b-4419-9430-902834a1d7e5/jobs/574997/steps

@rnveach

romani:
Failed CI is restarted

Zopsss:
> Failed CI is restarted 

Thanks now all CIs are passing.

I've made the changes according to your suggestions, is there anything else left to do? @rnveach

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrderRecords.java
```diff
@@ -60,7 +60,7 @@ public void foo() { // violation
         public MyClass() {
         }
 
-        public void foo(int i, String s) {
+        public void foo(int i, String s) { // violation
         }
 
         public void foo(String s, int i) {
```
nrmancuso, line 66: Why isn't this a violation?

nrmancuso (commented):
Question:

nrmancuso:
> @romani @nrmancuso @Zopsss https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/1f7b5dc_2024204006/reports/diff/apache-struts/index.html#A1
> 
> > Placing non-overloaded methods in between overloaded methods with the same type is a violation.
> 
> Shouldn't this verbiage be updated to something like `Placing anything in between overloaded methods`?

If `type` in the message refers to the return type of the method, this should definitely be removed since this has nothing to do with whether a method can be considered overloaded or not. 

https://docs.oracle.com/javase/tutorial/java/javaOO/methods.html#:~:text=Definition%3A%20Two%20of%20the%20components,%2C%20int%2C%20double%2C%20double)

> Definition: Two of the components of a method declaration comprise the method signature—the method's name and the parameter types.

https://docs.oracle.com/javase/specs/jls/se21/html/jls-8.html#jls-8.4.9

>If two methods of a class (whether both declared in the same class, or both inherited by a class, or one declared and one inherited) have the same name but signatures that are not override-equivalent, then the method name is said to be overloaded.

>This fact causes no difficulty and never of itself results in a compile-time error. There is no required relationship between the return types or between the throws clauses of two methods with the same name, unless their signatures are override-equivalent.

Does this check have logic based on return type?

Perhaps a better message would be `Overloaded method declarations should be grouped together`.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrderRecords.java
Zopsss: it is a violation. I've added the `// violation` comment to it.

Zopsss:
> Does this check have logic based on return type?

No the check does not care about the return type, it only cares about the method name.

> Perhaps a better message would be `Overloaded method declarations should be grouped together.`

Can it be something like this?

`Overloaded method declarations should be grouped together. Placing anything in between them is a violation, expect comment. Previous overloaded method is located at line '0'.`

This message is shorter than the current error message and also simple & easy to understand.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrderRecords.java
nrmancuso: > it is a violation. I've added the `// violation` comment to it to indicate that's a violation

CI is green and no changes have been pushed up, so I am not sure what this comment means.

Zopsss: > I am not sure what this comment means.

This was the new violation introduced due to the changes in the check implementation. I've added the violation comment and handled it here:

https://github.com/checkstyle/checkstyle/blob/922cba38624e05652b2d9cb37f13e342d0d4c03b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/OverloadMethodsDeclarationOrderCheckTest.java#L62

So there's no CI failing and everything is working fine as expected.

nrmancuso: This comment is for line 66, which is why I have placed it there.

Zopsss: Sorry, but I'm not able to understand what do you mean by line 66. 

https://github.com/checkstyle/checkstyle/blob/922cba38624e05652b2d9cb37f13e342d0d4c03b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/InputOverloadMethodsDeclarationOrderRecords.java#L57-L67

As you can see here, there is a class between the `foo()` & `foo(int i, String s)`. That's why there is a violation at `foo(int i, String s)` indicating that it should be placed next to `foo()` i.e the class should not be between the overloaded methods. There is no violation for line 66 because there is the overloaded `foo(int i, String s)` above it.

rnveach: So to sum it up, we just check if one overloaded method is sitting against another. Not that all overloaded methods need to be moved to one spot.

In the long run, both mean the same thing, but in a single slice of time, it is a difference. 66 would be a violation if only line 63 was moved. However, technically line 57 could be moved to 63 and everything would be compliant in terms of this check.

nrmancuso: Ah, so a user would have to run multiple checkstyle executions to make this check happy. I see.

rnveach:
@Zopsss We have come to a decision on new message.

Original message is `All overloaded methods should be placed next to each other. Placing non-overloaded methods in between overloaded methods with the same type is a violation. Previous overloaded method located at line X`.

New message is we trim message to just `All overloaded methods should be placed next to each other. Previous overloaded method located at line X.` So we basically remove the existing 2nd sentence.

Zopsss:
done. Updated the error message to the new one:

https://github.com/checkstyle/checkstyle/blob/8c91a6bfda8827449ee4a8ff544ab25bb018df53/src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties#L47

also updated it in other languages too

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/Example1.txt
```diff
@@ -10,6 +10,8 @@
 public void foo(int i) {}
 public void foo(String s) {}
 public void foo(String s, int i) {}
+// comments between overloaded methods are allowed.
```
rnveach, line 13: Add this comment to Example 2 to keep everything in sync. It is like you are saying 1 is what is wrong, and 2 is what everything looks like fixed.

rnveach (changes requested):
Last change.

rnveach:
I am good with code and regression.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/Example1.txt
Zopsss: Done. Just changed the comment location so the Example 2 does not look confusing and also changed the interface name used in the Example 2 from `Testing` to `ExampleInterface`

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/overloadmethodsdeclarationorder/Example1.txt
```diff
@@ -8,8 +8,10 @@
 
 // xdoc section -- start
 public void foo(int i) {}
+// comments between overloaded methods are allowed.
 public void foo(String s) {}
 public void foo(String s, int i) {}
 public void foo(int i, String s) {}
 public void notFoo() {}
+private interface Testing() {}
```
rnveach, line 16: This needs to be ExampleInterface then...

rnveach (changes requested):

Zopsss:
My IntelliJ automatically added unicodes to the translated messages.. fixing this asap. Idk why this happened. So sorry for this

Zopsss:
Updated the xdoc examples and resolved the failing CIs. 

PR is now ready to be merged. Sorry it took a little longer, my IntelliJ was not working properly idk why that happened.

rnveach (approved):

MANISH-K-07:
> PR is now ready to be merged. Sorry it took a little longer, my IntelliJ was not working properly idk why that happened.

@Zopsss , not just IntelliJ, IDEs in general are not automatically configured to detect these.
It is better to use primitive editors for properties files to avoid such unwanted changes..

Zopsss:
> @Zopsss , not just IntelliJ, IDEs in general are not automatically configured to detect these. It is better to use primitive editors for properties files to avoid such unwanted changes..

Okkay got it! Thanks for the help.

Zopsss:
@nrmancuso can you merge the PR please? Other maintainers have approved the changes.

nrmancuso (approved):

nrmancuso:
Good stuff @Zopsss !

Issue #12542, opened 2022-12-18
new Treewalker property to skip all nested Modules if there is parsing exception


romani:
It is our biggest limitation to not work on non compilable files
https://checkstyle.org/writingchecks.html#Limitations
> Code has to be compilable by javac to get valid violations. If it is not, you can get hard to understand parse errors. 

All of this is blocker for users:
https://github.com/checkstyle/checkstyle/issues?q=is%3Aopen+is%3Aissue+label%3Aantlr
but in most cases they mostlikely to ignore all Checkstyle activity on such files and let them use fancy new jdk syntax for business reasons.

We rely on plugin to cover this problem from us
But we still have problems with new syntax of Java

We need new property like `skipFileOnJavaParseException` or some thing similar.
If user wants new features of Java and ok to skip checkstyle in such files, it should be easy to do
We should not limit our users and let them decide : choose one checkstyle of new jdk features
By default it is disabled so exception crash all as before
But we can hint users after they create issue on use, about it, and they will be unblocked

We will introduce a new property called `skipFileOnJavaParseException`:
```
<module name="TreeWalker">
  <property name="skipFileOnJavaParseException" value="false"/>

```

romani:
@rnveach , @nrmancuso , @pbludov , @strkkk , please share your thoughts

rnveach:
Why not switch it to Javadoc's method of handling parsing failures and instead print a violation?

romani:
It still very complicated concept to suppress. Most users do not know that we have filters and how to manage them.
As they switched to new jdk, and what to proceed with jdk, better to not force them to update suppression each time someone what to use new syntax.

nrmancuso:
From the description, it seems to me that this exact behavior can be achieved by the `haltOnException` property and a `SuppressionSingleFilter` modlue:

```
➜  src cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <module name="TreeWalker">
    <module name="AbbreviationAsWordInName"/>
  </module>
</module>

➜  src cat Test.java 
public clazz Test {}

➜  src java -jar checkstyle-10.4-all.jar -c config.xml Test.java
Starting audit...
com.puppycrawl.tools.checkstyle.api.CheckstyleException: Exception was thrown while processing Test.java
[...]

Issue #12803, opened 2023-03-07
All interface methods should allow throwing CheckstyleException


rnveach:
Identified at https://github.com/checkstyle/checkstyle/pull/12151#discussion_r1125699128 ,

We saw that filters do not allow users to throw a `CheckstyleException` in the accept method. The only way they can throw exceptions is using unchecked exceptions like `IllegalStateException`.

We should look over all our interfaces that are part of our API, and allow users to throw `CheckstyleException` where appropriate and they are not allowed to throw anything else.

https://github.com/checkstyle/checkstyle/blob/5506d9587faf953e5e01825f4b1b50a81b83fdfe/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyTextFilter.java#L228-L238

https://github.com/checkstyle/checkstyle/blob/5506d9587faf953e5e01825f4b1b50a81b83fdfe/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyTextFilter.java#L194-L201

but Checks does not  this also:
https://github.com/checkstyle/checkstyle/blob/5506d9587faf953e5e01825f4b1b50a81b83fdfe/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/AbstractSuperCheck.java#L75-L81

sawaisinghh:
we can update the API interface to include a "throws CheckstyleException" declaration in the method signature. This will inform users of the API that they are allowed to throw this exception in the accept method.

Boukhalfa11:
Can I work in this issue please ? @romani @rnveach

Boukhalfa11:
okey, I am on it

romani:
Code that is in  referenced PR above, can be reused.

mohitsatr:
on it.

romani:
We do same pattern of catch-trow in other places
https://github.com/checkstyle/checkstyle/blob/5506d9587faf953e5e01825f4b1b50a81b83fdfe/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L178-L188

and we convert all exceptions to Checkstyle 
in processFiles 
https://github.com/checkstyle/checkstyle/blob/5506d9587faf953e5e01825f4b1b50a81b83fdfe/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L305-L311

most public method that runs all is 
[...]

PR #14779, opened 2024-04-10
Issue #12542: new TreeWalker property to skip exceptions

Lmh-java:
Issue #12542

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
@@ -79,11 +79,15 @@ public final class TreeWalker extends AbstractFileSetCheck implements ExternalRe
     /** A factory for creating submodules (i.e. the Checks) */
     private ModuleFactory moduleFactory;
 
+    /** Control whether to skip files with Java parsing errors. */
+    private boolean skipFileOnJavaParseException;
+
     /**
      * Creates a new {@code TreeWalker} instance.
      */
     public TreeWalker() {
         setFileExtensions("java");
+        setSkipFileOnJavaParseException(false);
     }
 
     /**
@@ -93,6 +97,16 @@ public final class TreeWalker extends AbstractFileSetCheck implements ExternalRe
      */
     public void setModuleFactory(ModuleFactory moduleFactory) {
         this.moduleFactory = moduleFactory;
+    }
+
+    /**
+     * Setter to control whether to skip files with Java parsing errors.
+     *
+     *  @param skipFileOnJavaParseException whether to always check for a trailing comma.
+     *  @since 10.15.1
+     */
+    public void setSkipFileOnJavaParseException(boolean skipFileOnJavaParseException) {
+        this.skipFileOnJavaParseException = skipFileOnJavaParseException;
     }
 
     @Override
@@ -149,23 +163,30 @@ public final class TreeWalker extends AbstractFileSetCheck implements ExternalRe
         // check if already checked and passed the file
         if (!ordinaryChecks.isEmpty() || !commentChecks.isEmpty()) {
             final FileContents contents = getFileContents();
-            final DetailAST rootAST = JavaParser.parse(contents);
-            if (!ordinaryChecks.isEmpty()) {
-                walk(rootAST, contents, AstState.ORDINARY);
+            try {
+                final DetailAST rootAST = JavaParser.parse(contents);
+                if (!ordinaryChecks.isEmpty()) {
+                    walk(rootAST, contents, AstState.ORDINARY);
+                }
+                if (!commentChecks.isEmpty()) {
+                    final DetailAST astWithComments = JavaParser.appendHiddenCommentNodes(rootAST);
+                    walk(astWithComments, contents, AstState.WITH_COMMENTS);
+                }
+                if (filters.isEmpty()) {
+                    addViolations(violations);
+                }
+                else {
+                    final SortedSet<Violation> filteredViolations =
+                            getFilteredViolations(file.getAbsolutePath(), contents, rootAST);
+                    addViolations(filteredViolations);
+                }
+                violations.clear();
             }
-            if (!commentChecks.isEmpty()) {
-                final DetailAST astWithComments = JavaParser.appendHiddenCommentNodes(rootAST);
-                walk(astWithComments, contents, AstState.WITH_COMMENTS);
+            catch (CheckstyleException ex) {
+                if (!skipFileOnJavaParseException) {
+                    throw ex;
+                }
             }
-            if (filters.isEmpty()) {
-                addViolations(violations);
-            }
-            else {
-                final SortedSet<Violation> filteredViolations =
-                    getFilteredViolations(file.getAbsolutePath(), contents, rootAST);
-                addViolations(filteredViolations);
-            }
-            violations.clear();
         }
     }
 
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
@@ -1,5 +1,6 @@
 Checker.finishLocalSetup=if no custom moduleFactory is set, moduleClassLoader must be specified
-Checker.processFilesException=Exception was thrown while processing {0}
+Checker.processFilesException=Exception was thrown while processing {0}. Set ''skipFileOnJavaParseException''\
+    property of ''TreeWalker'' module to ''true'' to skip all these exceptions.
 Checker.setCharset=unsupported charset: ''{0}''
 Checker.setupChildModule=cannot initialize module {0} - {1}
 Checker.setupChildNotAllowed={0} is not allowed as a child in Checker
--- a/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
@@ -1068,7 +1068,7 @@ public class CheckerTest extends AbstractModuleTestSupport {
         catch (CheckstyleException ex) {
             assertWithMessage("Error message is not expected")
                 .that(ex.getMessage())
-                .isEqualTo("Exception was thrown while processing " + filePath);
+                .contains("Exception was thrown while processing " + filePath);
         }
     }
 
@@ -1098,7 +1098,7 @@ public class CheckerTest extends AbstractModuleTestSupport {
         catch (CheckstyleException ex) {
             assertWithMessage("Error message is not expected")
                 .that(ex.getMessage())
-                .isEqualTo("Exception was thrown while processing " + filePath);
+                .contains("Exception was thrown while processing " + filePath);
 
             // destroy is called by Main
             checker.destroy();
--- a/src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
@@ -673,6 +673,34 @@ public class TreeWalkerTest extends AbstractModuleTestSupport {
         }
     }
 
+    @Test
+    public void testSkipFileOnJavaParseException() throws Exception {
+        final DefaultConfiguration config = createModuleConfig(TreeWalker.class);
+        config.addProperty("skipFileOnJavaParseException", "true");
+        config.addChild(createModuleConfig(ConstantNameCheck.class));
+
+        final List<File> files =
+                Arrays.asList(
+                        new File(getNonCompilablePath("InputTreeWalkerSkipParsingException.java")),
+                        new File(getPath("InputTreeWalkerProperFileExtension.java")),
+                        new File(getNonCompilablePath("InputTreeWalkerSkipParsingException.java")));
+
+        final Checker checker = createChecker(config);
+        try {
+            checker.process(files);
+        }
+        catch (CheckstyleException exception) {
+            assertWithMessage("exception is NOT expected").fail();
+        }
+    }
+
+    @Test
+    public void testSkipFileOnJavaParseExceptionConfig() throws Exception {
+        final String path = getNonCompilablePath("InputTreeWalkerSkipParsingExceptionConfig.java");
+        final String[] expected = {};
+        verifyWithInlineXmlConfig(path, expected);
+    }
+
     public static class BadJavaDocCheck extends AbstractCheck {
 
         @Override
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSkipParsingExceptionConfig.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSkipParsingExceptionConfig.java
@@ -1,0 +1,11 @@
+/*xml
+<module name="Checker">
+  <module name="TreeWalker">
+    <property name="skipFileOnJavaParseException" value="true"/>
+    <module name="ConstantName"/>
+  </module>
+</module>
+*/
+
+//non-compiled syntax: bad file for testing
+public clazz InputTreeWalkerSkipParsingExceptionConfig {}
--- a/src/xdocs/config.xml
+++ b/src/xdocs/config.xml
@@ -539,6 +539,13 @@
               <td><code>.java</code></td>
               <td>3.0</td>
             </tr>
+            <tr>
+              <td>skipFileOnJavaParseException</td>
+              <td>Control whether to skip files with Java parsing errors.</td>
+              <td><a href="property_types.html#Boolean">boolean</a></td>
+              <td><code>false</code></td>
+              <td>10.15.1</td>
+            </tr>
           </table>
         </div>
       </subsection>
--- a/config/ant-phase-verify.xml
+++ b/config/ant-phase-verify.xml
@@ -166,7 +166,8 @@
           <exclude name="**/InputMainFrameModelIncorrectClass.java"/>
           <exclude name="**/InputBeforeExecutionExclusionFileFilterIncorrectClass.java"/>
           <exclude name="**/InputJavaParser.java"/>
-
+          <exclude name="**/InputTreeWalkerSkipParsingException.java"/>
+          <exclude name="**/InputTreeWalkerSkipParsingExceptionConfig.java"/>
         </fileset>
       </path>
       <formatter type="plain"/>
@@ -203,6 +204,8 @@
           <exclude name="**/InputMainFrameModelIncorrectClass.java"/>
           <exclude name="**/InputBeforeExecutionExclusionFileFilterIncorrectClass.java"/>
           <exclude name="**/InputJavaParser.java"/>
+          <exclude name="**/InputTreeWalkerSkipParsingException.java"/>
+          <exclude name="**/InputTreeWalkerSkipParsingExceptionConfig.java"/>
           <!-- This check does not apply to grammar input files -->
           <exclude name="**/grammar/**/*"/>
           <!-- No need to check xpathmapper files -->
--- a/src/test/java/com/puppycrawl/tools/checkstyle/MainTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/MainTest.java
@@ -918,7 +918,7 @@ public class MainTest {
                 + new File(getNonCompilablePath("InputMainIncorrectClass.java")).getPath());
         assertWithMessage("Unexpected system error log")
                 .that(systemErr.getCapturedData())
-                .contains(exceptionMessage);
+                .contains(exceptionMessage.trim());
     }
 
     @Test
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSkipParsingException.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSkipParsingException.java
@@ -1,0 +1,3 @@
+//non-compiled syntax: bad file for testing
+
+public clazz InputTreeWalkerSkipParsingException {}


src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
```diff
@@ -1,5 +1,6 @@
 Checker.finishLocalSetup=if no custom moduleFactory is set, moduleClassLoader must be specified
-Checker.processFilesException=Exception was thrown while processing {0}
+Checker.processFilesException=Exception was thrown while processing {0}. Set ''skipFileOnJavaParseException''\
+    property of ''TreeWalker'' module to ''true'' to skip all these exceptions.
```
Lmh-java, line 3: If this message works, I will translate this to other languages as well.

Lmh-java:
I think the 2 failing CIs are related to 2 new suppressions of non-compilable files. These new suppressions do not exist on the master branch, so these two non-compilable files will still fail the CI.

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
```diff
@@ -95,6 +98,16 @@ public void setModuleFactory(ModuleFactory moduleFactory) {
         this.moduleFactory = moduleFactory;
     }
 
+    /**
+     * Setter to control whether to skip files with Java parsing errors.
+     *
+     *  @param skipFileOnJavaParseException whether to always check for a trailing comma.
+     *  @since 10.15.1
```
romani, line 105: 10.16.0 , it will be new feature, that people awaits alot.

romani (changes requested):
item:

src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
romani: We should not change this message, we should not promote this property to be used.
User need to understand context of problem to make decision himself, as this skip can result in more uncertainty and confusion.

romani:
recheck what is missing at https://github.com/checkstyle/contribution/blob/a2f8b1575c42b9156fdcedabbb528296eb32e09a/checkstyle-tester/projects-to-test-on.properties#L7
to skip such files

please send PR to that repo.

Lmh-java:
@romani PR is created here https://github.com/checkstyle/contribution/pull/852.

romani:
Merged, CIs are restarted

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
Lmh-java: done

src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
Lmh-java: removed

src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
```diff
@@ -1068,7 +1068,7 @@ public void testHaltOnException() throws Exception {
         catch (CheckstyleException ex) {
             assertWithMessage("Error message is not expected")
                 .that(ex.getMessage())
-                .isEqualTo("Exception was thrown while processing " + filePath);
+                .contains("Exception was thrown while processing " + filePath);
```
romani, line 1071: Why this matter?

romani (changes requested):
Last:

src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
Lmh-java: done

src/xdocs/config.xml
```diff
@@ -539,6 +539,13 @@
               <td><code>.java</code></td>
               <td>3.0</td>
             </tr>
+            <tr>
+              <td>skipFileOnJavaParseException</td>
+              <td>Control whether to skip files with Java parsing errors.</td>
+              <td><a href="property_types.html#boolean">boolean</a></td>
+              <td><code>false</code></td>
+              <td>10.15.1</td>
```
romani, line 547: 10.16.0

romani (changes requested):
last:

romani (approved):
Ok to merge.
Thanks a lot!

Lmh-java:
Sorry for didn't catch that

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
```diff
@@ -149,23 +162,30 @@ protected void processFiltered(File file, FileText fileText) throws CheckstyleEx
...
+                    addViolations(filteredViolations);
+                }
+                violations.clear();
             }
-            else {
-                final SortedSet<Violation> filteredViolations =
-                    getFilteredViolations(file.getAbsolutePath(), contents, rootAST);
-                addViolations(filteredViolations);
+            catch (CheckstyleException ex) {
+                if (!skipFileOnJavaParseException) {
```
rnveach, line 185: This catch spans almost the entire method and has the possibility to catch more than just parsing exceptions. There is no check that we are catching parse exceptions. We could be catching much more.

This should be trimmed back to ensure it only catches parse exceptions.

There is issue https://github.com/checkstyle/checkstyle/issues/12803 to allow us to have more methods that throw checkstyle exceptions.

rnveach (commented):
I don't see any regression provided to ensure this is working.

We need 2 reports. One with default treewalker (expecting no changes), another with new property turned on (show exceptions removed).

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
```diff
@@ -149,23 +162,30 @@ protected void processFiltered(File file, FileText fileText) throws CheckstyleEx
...
+                    throw ex;
+                }
```
rnveach, line 187: @romani We are not going to display anything when exception occurs? We are going to **completely** hide that this happened? Even with `haltOnException`, we display a message to the user to notify this happened.
https://github.com/checkstyle/checkstyle/blob/86195e06404df68905c90232f71a1c88491f9ad8/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L349C61-L365

This seems very risk to completely ignore it.

rnveach (changes requested):

rnveach:
@romani In addition, is this something to enable for google/sun styles?

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
Lmh-java: @rnveach I agree, this is definitely not the best way to suppress expections. However, I was trying to only enclose `final DetailAST rootAST = JavaParser.parse(contents);` with try-catch. But, in that case, I need to use a `return` to block the following procedure, which will cause our CI to complain, since it is not allowed to have a return in a void method. Do you agree to suppress it?

Lmh-java:
> @romani In addition, is this something to enable for google/sun styles?
> 
> What about our default regression configurations?

@rnveach I was also thinking about this the other day. I think we can copy the original 2 check config files an make another 2 with this property on. I.e., we have 4 built-in config: 1. Sun 2. Google 3. Sun No Exception 4. Google No Exception. 

My reason is: there are users who want to use the original check with exceptions, and for other group of user, in their cases, they might want to ignore the exceptions.

About our default regressions: 
I would like to use it in the new performance CI (#14599), since this property costs less than a long list of suppressions and will make the benchmark more accurate. I think that's probably same in other regression tests.

rnveach:
> I think we can copy the original 2 check config files 

I am against duplicating the configs. There is more work that goes into ensuring how they are built and we don't need the headache of maintaining 2 copies.

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
rnveach: I do not agree to suppress yet. There should ways around this that follow our style rules.

Lmh-java: I can also add a local variable flag. When an exception occurs set the flag to true, and we guard the following procedure with this flag. If this flag is true, then we do not proceed the following procedure. What about this?

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
romani: >We are not going to display anything when exception occurs? We are going to completely hide that this happened?

If user explicitly configure to skip, we do skip swallow and no noise warning or any other messages. User should use it with caution.

romani:
>is this something to enable for google/sun styles?

No, we are not going to promote this property, user should use it only if they really want it, and they understand results.

Lmh-java:
GitHub, generate report

Lmh-java:
GitHub, generate report

Lmh-java:
GitHub, generate report

Lmh-java:
Base with filter suppression, patch without filter suppression and skip property is on:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c6a72a2_2024061201/reports/diff/index.html

Base with filter suppression, patch with filter suppression and skip property is off:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c6a72a2_2024070234/reports/diff/index.html

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
romani: Good catch, we should catch only parse exceptions, not CheckstyleException.

romani: We need to target with try catch only https://github.com/checkstyle/checkstyle/blob/86195e06404df68905c90232f71a1c88491f9ad8/src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java#L103

Lmh-java: done, work around by using a `skip` flag.

Lmh-java: @romani @rnveach This workaround using a `skip` flag seems bring us more suppressions with nullness. 
https://github.com/checkstyle/checkstyle/blob/e84dc9c47b2d593f52691dc19039b8c46fe2b261/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java#L169-L198
```
New surviving error(s) found:

    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java</fileName>
    <specifier>argument</specifier>
    <message>incompatible argument for parameter ast of TreeWalker.walk.</message>
    <lineContent>walk(rootAST, contents, AstState.ORDINARY);</lineContent>
    <details>
    </details>
  </checkerFrameworkError>

    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java</fileName>
    <specifier>argument</specifier>
    <message>incompatible argument for parameter root of JavaParser.appendHiddenCommentNodes.</message>
    <lineContent>final DetailAST astWithComments = JavaParser.appendHiddenCommentNodes(rootAST);</lineContent>
    <details>
    </details>
  </checkerFrameworkError>

    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java</fileName>
    <specifier>argument</specifier>
    <message>incompatible argument for parameter rootAST of TreeWalker.getFilteredViolations.</message>
    <lineContent>getFilteredViolations(file.getAbsolutePath(), contents, rootAST);</lineContent>
    <details>
    </details>
  </checkerFrameworkError>
```

I felt like this structure of code is too complicated. This is the core code of the project, it would be hard for later maintainers if we do more workarounds In this case, do you prefer to still work on this workaround (add a dummy initial value for rooAST instead of null), or prefer to suppress a single return statement, or there are any better solutions. Please share your thoughts. I am out of ideas.

romani (approved):

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
```diff
@@ -149,23 +162,40 @@ protected void processFiltered(File file, FileText fileText) throws CheckstyleEx
...
-            }
-            if (!commentChecks.isEmpty()) {
-                final DetailAST astWithComments = JavaParser.appendHiddenCommentNodes(rootAST);
-                walk(astWithComments, contents, AstState.WITH_COMMENTS);
+            DetailAST rootAST = null;
+            // whether skip the procedure after parsing Java files.
+            boolean skip = false;
+
+            try {
+                rootAST = JavaParser.parse(contents);
```
nrmancuso, line 170: I am open to creating a Try monad for this purpose, so that we can return a value or exception and deal with the result accordingly.

Lmh-java: Sure, thanks for your advice. I will try to create a try monad :)

nrmancuso: Other maintainers might not like this, it might be good to get some feedback first

romani: Please better to postpone monad improvement to next PR. Let's merge functional fix first.

Lmh-java: @romani @nrmancuso @rnveach Could you please take a look at https://github.com/checkstyle/checkstyle/pull/14779#discussion_r1564914632. The new workaround is kind of wired to me.

nrmancuso: > @romani @nrmancuso @rnveach Could you please take a look at [#14779 (comment)](https://github.com/checkstyle/checkstyle/pull/14779#discussion_r1564914632). The new workaround is kind of wired to me.

Yes, I do not like this update, which is why I recommended a Try monad for to avoid mutable flags and other hacks.

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
nrmancuso: > I am for displaying a violation which can be suppressed, similar to halt on exception.

Yep, same here. Otherwise, how can we easily determine what is being skipped?

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
romani: >it would be hard for later maintainers

If we want to have better code, please do this actively, right now this issue stuck in not done state, and users still having unfortunate decisions to disable checkstyle because it doesn't support modern language features.
You will not avoid catch of exception, so Checker will be not happy, and you will do just playing with creation of method that hide such exceptions handling, but conceptually all will stay the same.

If we start new method creation, you will stick in review more. My suggestion is pass review quicker, get feature out to users, and work in separate PR in beautiful code as much as we need, no rush. You can create new  PR, right now and collaborate on ideal form of code. 
There is no problem with mutation of variable, we do this in a lot of places, one day we will improve, but existence of this doesn't block users to receive bug fixes and new features.

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
romani: If user use this option, he is completely ok to skip checkstyle on some classes.  Some Users value new syntax features more than checkstyle validation, it is normal, we should allow this.

It is same as all plugins have `skip` options, to allow keep plugin plugin still in build but let users decide when to activate it to fail a build or ignore it when they want this.

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
Lmh-java: >My suggestion is pass review quicker, get feature out to users, and work in separate PR in beautiful code as much as we need, no rush.

Thanks for your advice. I will aim to finish this PR right now by suppressing the mutation test shown above. :)

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
nrmancuso: > If user use this option, he is completely ok to skip checkstyle on some classes.

I’m not debating this; it is the essence of this feature. My question is how can a user *know* what is being skipped?

romani: No way, we do not provide any other communication message to users other than violations. Do we want to place violation with info level on whole file. But if we do this, we need to give user ability to adjust severity of violation as they want, even they might want to put "ignore". But it is extension of public config/design, I don't have good ideas for now, I would wait for feedback from users on how they use it in real life.

Lmh-java:
GitHub, generate report

Lmh-java:
GitHub, generate report

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
nrmancuso: We can just follow the existing model of haltOnException, we have no need to reinvent the wheel.

Lmh-java:
Base with filter suppression, patch without filter suppression and skip property is on:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/79ca895_2024210432/reports/diff/index.html

Base with filter suppression, patch with filter suppression and skip property is off:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/79ca895_2024204023/reports/diff/index.html

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
romani: https://github.com/checkstyle/checkstyle/blob/174bccc47ad62dae2b28c3fd10baf3a20809d425/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java#L351-L369

We do conversion of exception to violation by haltOnException.
Default severity is error 
https://github.com/checkstyle/checkstyle/blob/174bccc47ad62dae2b28c3fd10baf3a20809d425/src/main/java/com/puppycrawl/tools/checkstyle/api/Violation.java#L43

We need some property that do parse exception handling only and second property to define severity. @nrmancuso , can you suggest good names ?

nrmancuso: >second property to define severity

This is a really good idea. How about:
```xml
<module name="Checker">
  <property name="skipFileOnJavaParseException" value="true"/>
  <property name="parseExceptionSeverity" value="warning"/>
  <module name="TreeWalker">
    <module name="OverloadMethodsDeclarationOrder"/>
  </module>
</module>
```

Lmh-java: Sure, I have already come up with a functional workaround for the try-catch.

I would like to conclude everything here. I will print a long when there is a Java parsing error. AIso, a new property named `parseExceptionSeverity` will be added to control the log serverity.

For the new severity, do we prefer to open a new PR or just keep everything in here?

Lmh-java: @nrmancuso @romani @rnveach
According to https://github.com/checkstyle/checkstyle/pull/14779#discussion_r1573101500, do we move these two new properties to `Checker` instead of `TreeWalker`?

If we log the message about Java parsing exception in TreeWalker, there is still another layer of try-catch in `Checker`, which will print the exception information again. Therefore, just want to confirm that, we are discarding all the changes in `TreeWalker` and move what is here now into `Checker`?

rnveach: > we log the message 

It sounds like there may be some confusion, especially on log.

If `skipFileOnJavaParseException` is turned on, we are talking about writing the exception as a violation and never throwing the exception. The severity of the violation will be determined by `parseExceptionSeverity`.
If `skipFileOnJavaParseException` is turned off, we follow default behavior by throwing the exception. We let what other mechanisms we have take over as that is the default behavior.

Unless I am missing something, `skipFileOnJavaParseException` is really only about stopping the exception from skipping other `TreeWalker` processing.

nrmancuso: @Lmh-java we want to emulate the `haltOnException` property. It is not clear to me why checker should care. Parsing exception happens, we catch it and log the message. Is this behavior not possible?

nrmancuso: >According to https://github.com/checkstyle/checkstyle/pull/14779#discussion_r1573101500, do we move these two new properties to Checker instead of TreeWalker?

This is my mistake, these should live in TreeWalker for sure, thanks for calling this out.

Lmh-java: > Is this behavior not possible?

I think this is possible. let me try. :)

nrmancuso:
> https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/79ca895_2024210432/reports/diff/index.html

I would expect no TreeWalker checks to run on files that we are skipping. I see that we are now passing a null AST to checks; I am not sure about what side effects this could possibly have. If we want to explore a bit, we could use the same config that we use in our no-exception openjdk executions (without the file filters) to see what other checks are effected. We will need to add a test reproducing the issue seen in the regression report.


@Lmh-java please help us to understand what we need to do to avoid passing a null AST to TreeWalker checks.

Lmh-java:
GitHub, generate report

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
Lmh-java: That's a great idea. Thanks for your advice. I didn't think about `Optional` before. I will use that instead.

Lmh-java:
> > https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/79ca895_2024210432/reports/diff/index.html
> 
> I would expect no TreeWalker checks to run on files that we are skipping. I see that we are now passing a null AST to checks; I am not sure about what side effects this could possibly have. If we want to explore a bit, we could use the same config that we use in our no-exception openjdk executions (without the file filters) to see what other checks are effected. We will need to add a test reproducing the issue seen in the regression report.
> 
> @Lmh-java please help us to understand what we need to do to avoid passing a null AST to TreeWalker checks.

This is a mistake. I introduced `skip` to skip the checks when exception happened. However, somehow it did not skip checks. I should not pass `null` to checks. Otherwise there will be lots of changes with some specific checks. They are hard to detect. Let me fix this.

Passing `null` AST to major checks will not cause any problem. However, for some special checks (such as `NoCodeInFile` shown in the regression report), there will be a different. In this case of `NoCodeInFile`, there is a guard for `null`.
```java
@Override
public void finishTree(DetailAST ast) {
    if (ast == null) {
        log(DEFAULT_LINE_NUMBER, MSG_KEY_NO_CODE);
    }
}
```
so it will be logged.

src/xdocs/config.xml
```diff
@@ -539,6 +539,20 @@
               <td><code>.java</code></td>
               <td>3.0</td>
             </tr>
+            <tr>
+              <td>javaParseExceptionSeverity</td>
+              <td>Severity level to log Java parsing exceptions when they are skipped.</td>
+              <td><a href="property_types.html#SeverityLevel">SeverityLevel</a></td>
+              <td><code>error</code></td>
+              <td>10.16.1</td>
```
romani, line 547: 10.17.0

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSkipParsingExceptionConfig.java
```diff
@@ -0,0 +1,12 @@
+/*xml
+<module name="Checker">
+  <module name="TreeWalker">
+    <property name="skipFileOnJavaParseException" value="true"/>
+    <property name="javaParseExceptionSeverity" value="ignore"/>
+    <module name="ConstantName"/>
```
romani, line 6: Add here check that participate in diff report.

romani (changes requested):
Items

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSkipParsingExceptionConfig.java
romani: NoCodeInFile

Lmh-java: I have finished these updates. Waiting for #14852 to be solved, so we can test that with this PR.

romani: Performance test  PR is not related.
Ok Let keep, `name="ConstantName"/>`  , but please create new test for NoCodeInFile, to show how it works.

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
```diff
@@ -673,6 +679,129 @@ public void testOrderOfCheckExecution() throws Exception {
...
+
+        try {
+            execute(config, files);
+            assertWithMessage("Exception is expected").fail();
+        }
+        catch (CheckstyleException exception) {
+            assertWithMessage("Error message is unexpected")
+                    .that(exception.getMessage())
+                    .contains("Exception was thrown while processing ");
+        }
```
nrmancuso, line 724: Please use https://github.com/checkstyle/checkstyle/blob/bfb3e5f5e416211bc95799426edba4581b442d0d/src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/TestUtil.java#L393 in all cases like this.

nrmancuso (changes requested):
Item:

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSkipParsingExceptionConfig.java
Lmh-java: There are already some tests for NoCodeInFile.
For example, in the following test we do not expect any violations from NoCodeInFile if there is a java parsing exception.
```java
public void testSkipFileOnJavaParseExceptionSkipChecks() throws Exception {
        final DefaultConfiguration config = createModuleConfig(TreeWalker.class);
        config.addProperty("skipFileOnJavaParseException", "true");
        config.addProperty("javaParseExceptionSeverity", "ignore");
        config.addChild(createModuleConfig(NoCodeInFileCheck.class));

        final Checker checker = createChecker(config);

        final File[] files = {
            new File(getNonCompilablePath("InputTreeWalkerSkipParsingException.java")),
            new File(getPath("InputTreeWalkerProperFileExtension.java")),
        };
        final Map<String, List<String>> expectedViolation = new HashMap<>();
        expectedViolation.put(getPath("InputTreeWalkerProperFileExtension.java"),
                new ArrayList<>());

        verify(checker, files, expectedViolation);
    }
```

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
Lmh-java: Done

romani (approved):
ok to merge

romani:
GitHub, generate report

romani:
@Lmh-java , please update config to make severity as IGNORE and generate one more report

romani:
@Lmh-java , please extend test to kill mutation https://github.com/checkstyle/checkstyle/actions/runs/9040829453/job/24845471896?pr=14779#step:7:24

You can reproduce it on your local by `./.ci/pitest.sh "pitest-tree-walker"` command in terminal.

Lmh-java:
GitHub, generate report

Lmh-java:
> @Lmh-java , please update config to make severity as IGNORE and generate one more report

Done, waiting for the report

Lmh-java:
> @Lmh-java , please extend tests to kill mutation https://github.com/checkstyle/checkstyle/actions/runs/9040829453/job/24845471896?pr=14779#step:7:24
> 
> You can reproduce it on your local by `./.ci/pitest.sh "pitest-tree-walker"` command in terminal.

Done

romani:
Last update, please add this Checker stuff to suppression https://github.com/checkstyle/checkstyle/actions/runs/9052825433/job/24870931959?pr=14779#step:6:1707

nrmancuso (approved):
We could stand to improve this code in the future, but let's get this feature into user's hands first.

src/xdocs/config.xml
```diff
@@ -539,6 +539,20 @@
               <td><code>.java</code></td>
               <td>3.0</td>
             </tr>
+            <tr>
+              <td>javaParseExceptionSeverity</td>
+              <td>Severity level to log Java parsing exceptions when they are skipped.</td>
```
rnveach, line 544: => `Specify the severity level...`

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
```diff
@@ -95,6 +105,27 @@ public void setModuleFactory(ModuleFactory moduleFactory) {
...
+     *
+     *  @param skipFileOnJavaParseException whether to always check for a trailing comma.
+     *  @since 10.17.0
+     */
+    public void setSkipFileOnJavaParseException(boolean skipFileOnJavaParseException) {
+        this.skipFileOnJavaParseException = skipFileOnJavaParseException;
+    }
+
+    /**
+     * Sets severity level to log Java parsing exceptions when they are skipped.
```
rnveach, line 119: => `Setter to specify the severity level...`

rnveach (changes requested):
Still reviewing the code.

Add an example to https://checkstyle.org/config.html#TreeWalker_Examples and explain when they would use such a feature and why. Be sure to explain we will only skip checks under TreeWalker and not under Checker.

rnveach:
@romani 
> What about our default regression configurations?
> I do not think we need this flag in our configs. We should clearly see exception in report.

Our default regression configuration ignores these exceptions. So it doesn't make sense we wouldn't add this.
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/my_check.xml#L45

rnveach:
@Lmh-java Can you organize your regression reports? They seem a bit scattered. Also, the ones I am seeing only run against 1 project. We normally request for them to be run on all projects. This can impact more than just JDK 17 projects. If older projects show no differences, then it shows there is no harm.

rnveach (changes requested):

romani:
@Lmh-java, please do us a favor 

Please send PR to remove this block
https://github.com/checkstyle/contribution/blob/fb4a65b361fd6c6887444eb0e89a9191ec89d789/checkstyle-tester/my_check.xml#L43-47
And use your new properties.

src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java
```diff
@@ -149,21 +180,41 @@ protected void processFiltered(File file, FileText fileText) throws CheckstyleEx
...
-                walk(astWithComments, contents, AstState.WITH_COMMENTS);
+            DetailAST rootAST = null;
+            // whether skip the procedure after parsing Java files.
+            boolean skip = false;
+            try {
+                rootAST = JavaParser.parse(contents);
             }
-            if (filters.isEmpty()) {
-                addViolations(violations);
+            catch (CheckstyleException ex) {
```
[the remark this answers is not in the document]
romani: @Lmh-java , please add this file to our Inputs https://github.com/openjdk/jdk/blob/master/test/langtools/tools/javac/diags/examples/VarargsAndReceiver.java without header comment.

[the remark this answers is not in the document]
rnveach: From discussions in discord, it is better to catch `Exception` instead of `Throwable`.

Issue #13845, opened 2023-10-08
SuppressWarningsHolder naming inconsistency giving java.lang.ClassNotFoundException


relentless-pursuit:
Upon trying to work on enabling example tests for `SuppressWarningsHolderExamplesTests`, I constantly encountered with `java.lang.ClassNotFoundException: com.puppycrawl.tools.checkstyle.checks.annotation.SuppressWarningsHolderCheck`. 

Upon debugging the code, I realised that the code appends 'Check' to the fully qualified name of the class. While, we don't have `SuppressWarningsHolderCheck.java` rather `SuppressWarningsHolder.java`. In the method getFullyQualifiedClassName, a specific block of code highlighted below is doing the append operation. 
```
if (!fullyQualifiedClassName.endsWith("Filter")) {
                fullyQualifiedClassName += "Check";
            }
```

Infact, SuppressWarningsHolder is part of check module and needs to be refactored as `SuppressWarningsHolderCheck.java`.

nrmancuso:
@relentless-pursuit please do:
1. Add a link to the check documentation
2. Add permalink to the referenced code
3. Investigate historical reason for this naming (git blame/ GitHub history can be useful for this)

romani:
> needs to be refactored as SuppressWarningsHolderCheck.java


this is big breaking compatibility for our users to change name of module.
and this module is not a Check, it is bad hack, and one day we will need remove it.

relentless-pursuit:
> > needs to be refactored as SuppressWarningsHolderCheck.java
> 
> this is big breaking compatibility for our users to change name of module. and this module is not a Check, it is bad hack, and one day we will need remove it.

But, the code seems to consider it as a check. If you see my recent PR,  https://github.com/checkstyle/checkstyle/pull/13848, I had explicitly add the fully qualified name in the configuration of xml file. If i just use <module name="SuppressWarningsHolder"/>, it complains of `com.puppycrawl.tools.checkstyle.checks.annotation.SuppressWarningsHolderCheck`. As you mentioned refactoring it to a check class won't be a good idea. But, then the other way around would be to ensure the the the fully qualified name is returned correctly i.e. `com.puppycrawl.tools.checkstyle.checks.annotation.SuppressWarningsHolder`. Meanwhile, I will follow @nrmancuso instructions to add more substatance to the issue.

romani:
We need to extend our test code to handle this unfortunate module in special way, config in example should be same as users have in config.

[...]

PR #14784, opened 2024-04-11
Issue #13345: enable example test for SeverityMatchFilterExamplesTest

Kexin2000:
part of issue: #13345
this is the first example test for filter, I've encountered some issues and need to discuss potential solutions. @romani @nrmancuso 
1. My current approach to writing filter examples is inspired by [this file](https://github.com/checkstyle/checkstyle/blob/571e547a9f7b0a05503b8d76f34181059a8c3bbf/src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/filters/suppresswithplaintextcommentfilter/Example9.java#L4): demonstrating the function of the filter through other checks. I used ParameterNameCheck and MethodNameCheck, setting the severity of ParameterNameCheck to info, while the MethodNameCheck defaults to error. However, when I run SeverityMatchFilterExamplesTest, the code enters line 382 of the following code block and throws an exception. The reason is that the path where SeverityMatchFilterExamplesTest is located does not contain module name parameters for 'parametername' and 'methodname'. So, I added 'parametername' and 'methodname' to the hashmap 'moduleMappings'. I think this approach should be acceptable, as I used 'git blame' to investigate the historical reason for 'moduleMappings'. The existence of this variable is to address some special cases in #13345, such as SuppressWarningsHolder #13845.

https://github.com/checkstyle/checkstyle/blob/571e547a9f7b0a05503b8d76f34181059a8c3bbf/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L372-L390

2. While running SeverityMatchFilterExamplesTest, there's a similar issue with a path mismatch. The code will throw an exception at line 602 of the following code block because there is no message in the filters package. So, I added a messages.properties file under src/main/resources/com/puppycrawl/tools/checkstyle/filters, and for subsequent filter example tests, more violation messages can be added to this file to resolve this issue. Is my solution feasible? (Messages.properties files for other languages also need to be added.)

https://github.com/checkstyle/checkstyle/blob/571e547a9f7b0a05503b8d76f34181059a8c3bbf/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java#L600-L610


Kexin2000:
The work on this PR is not yet complete. To avoid unnecessary work, I want to first confirm a suitable solution.

src/xdocs/filters/severitymatchfilter.xml
```diff
@@ -48,13 +48,19 @@
...
       <subsection name="Examples" id="Examples">
-        <p id="Example1-config">
+        <p id="Example-config">
           For example, the following configuration fragment directs the
           Checker to not report audit events with severity
           level <code>info</code>:
         </p>
         <source>
 &lt;module name=&quot;Checker&quot;&gt;
+  &lt;module name=&quot;TreeWalker&quot;&gt;
```
romani, line 58: please do us a favor and move this example to "Example of Usage"

src/main/resources/com/puppycrawl/tools/checkstyle/filters/messages.properties
```diff
@@ -0,0 +1 @@
+name.invalidPattern=Name ''{0}'' must match pattern ''{1}''.
```
romani, line 1: what forcing you to add this file and such property?

romani (changes requested):
items:

romani:
>  So, I added 'parametername' and 'methodname' to the hashmap 'moduleMappings'. I think this approach should be acceptable,

yes, it is defect of our tes suite as it expects modules to be in same package and target module - Filter in this case.

>  So, I added a messages.properties file under src/main/resources/com/puppycrawl/tools/checkstyle/filters, and for subsequent filter example tests, more violation messages can be added to this file to resolve this issue. Is my solution feasible?

not good, can we update internal suite code to take messages from file we will provide as hardcoded values, same as we did for fully qualified names ( item above)

src/main/resources/com/puppycrawl/tools/checkstyle/filters/messages.properties
romani: see guidance at https://github.com/checkstyle/checkstyle/pull/14784#issuecomment-2049962607

Kexin2000: done

src/xdocs/filters/severitymatchfilter.xml
romani: I was confused, I recall this

src/xdocs/filters/severitymatchfilter.xml
```diff
@@ -55,12 +55,29 @@
...
         </source>
+        <p id="Example1-code">
+          Code Example:
+        </p>
+        <source>
+public class Example1 {
+
+  public void method1(int V1){} // ok, ParameterNameCheck's severity is info
+
+  public void Method2(){} // violation
```
romani, line 78: `// violation, as default severity is error`

src/xdocs/filters/severitymatchfilter.xml
```diff
@@ -55,12 +55,29 @@
...
   &lt;/module&gt;
 &lt;/module&gt;
         </source>
+        <p id="Example1-code">
+          Code Example:
+        </p>
+        <source>
+public class Example1 {
+
+  public void method1(int V1){} // ok, ParameterNameCheck's severity is info
```
romani, line 76: `// ok, ParameterName's severity is info`

src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
```diff
@@ -629,12 +629,21 @@ private static String getMessageBundle(String className) {
...
+                "com.puppycrawl.tools.checkstyle.checks.naming.messages");
+
         if (endIndex < 0) {
             messageBundle = messages;
         }
         else {
             final String packageName = className.substring(0, endIndex);
-            messageBundle = packageName + "." + messages;
+            if ("com.puppycrawl.tools.checkstyle.filters".equals(packageName)) {
+                messageBundle = messageBundleMappings.get(className.substring(endIndex + 1));
```
romani, line 642: Please change logic to use value from manual mapping if mapping is present.
If present in special map then use it , else use packageName

romani (changes requested):
Items

romani:
@Kexin2000 ,please find time to finish PR

romani:
@Kexin2000 ,please find time to finish PR.
Let us know if somebody else can finish this PR

romani:
will be done at https://github.com/checkstyle/checkstyle/pull/15021

Issue #14747, opened 2024-04-01
ParenPad incorretly flags unsupported token


mahfouz72:
I have read check documentation: https://checkstyle.org/checks/whitespace/parenpad.html#ParenPad
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```
PS D:\test> cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

 <module name="TreeWalker">
     <module name="ParenPad"/>
    </module>
</module>

PS D:\test> cat src/Main.java
import static java.lang.StringTemplate.STR;
record ColoredPoint(int x, int y, String c) {
}
record Rectangle(ColoredPoint upperLeft, ColoredPoint lowerRight) {
}

public class Main {
    public static void main(String[] args) {
        ColoredPoint cp = new ColoredPoint(1, 2, "red");
        Rectangle r = new Rectangle(cp, new ColoredPoint(3, 4, "blue"));
        printUpperLeftColoredPoint(r);

    }
    static void printUpperLeftColoredPoint(Object obj) {
            if (obj instanceof Rectangle(  ColoredPoint(_, int y, _), _)) {  // unsupported token
                System.out.println(y);
            }
        }
}
PS D:\test> java  -jar checkstyle-10.14.2-all.jar -c config.xml  src/Main.java
Starting audit...
[ERROR] D:\test\src\Main.java:19:41: '(' is followed by whitespace. [ParenPad]
Audit done.
Checkstyle ends with 1 errors.
PS D:\test> 

```
**Describe what you expect in detail.**
no violation as `RECORD_PATTERN` token is not supported in the check

romani:
```
$ java -jar checkstyle-10.15.0-all.jar -t Test.java
....
|--LITERAL_IF -> if [15:12]
|   |--LPAREN -> ( [15:15]
|   |--EXPR -> EXPR [15:20]
|   |   `--LITERAL_INSTANCEOF -> instanceof [15:20]
|   |       |--IDENT -> obj [15:16]
|   |       `--RECORD_PATTERN_DEF -> RECORD_PATTERN_DEF [15:31]
[...]

Issue #4175, opened 2017-04-09
ParenPad: deep scan picks up excess tokens not set in config


rnveach:
Identified in https://github.com/checkstyle/checkstyle/pull/4109#issuecomment-291590567 ,

ParenPad is doing a deep scan for tokens to check at https://github.com/checkstyle/checkstyle/blob/e70836bb3dec30734f5fd51b63d4dc6cbe24679d/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheck.java#L179-L198 .
This caused tokens that were specifically not set in the configuration to be picked up for violations as seen in Issue #3048.
It is weird to validate tokens outside of check's configuration and users won't be expecting this. It even caused no differences to appear in regression reports for default module even when new token was added to the check. We only saw differences when running on very limited/specific tokens.

We need to remove this deep scan from the check.
We then need to identify if there are any differences in the regression report. If there are any differences, those areas have to be identified and added as new tokens to the check.
When all this is done, we expect no differences for default module in regression on final implementation. Differences on specific tokens is ok.

rnveach:
modifying `AbstractParenPadCheck`'s `processLeft` and `processRight` to notify of violations for possible unacceptable token parents comes up with the following list after running `ParenPadCheckTest`:
````
\checkstyle\src\test\resources\com\puppycrawl\tools\checkstyle\checks\whitespace\InputParenPad.java
([26x47] - parent = TYPECAST
([26x52] - parent = STAR
)[26x58] - parent = STAR
)[26x59] - parent = TYPECAST
([60x49] - parent = TYPECAST
([60x55] - parent = STAR
)[60x62] - parent = STAR
)[60x63] - parent = TYPECAST
([94x48] - parent = TYPECAST
([94x53] - parent = STAR
)[94x60] - parent = STAR
)[94x62] - parent = TYPECAST
([113x21] - parent = TYPECAST
([123x28] - parent = LAND
)[123x44] - parent = LAND
([131x22] - parent = TYPECAST
([131x27] - parent = STAR
)[131x33] - parent = STAR
)[132x16] - parent = TYPECAST
([160x27] - parent = LNOT
)[160x48] - parent = LNOT
([167x40] - parent = PLUS
([192x47] - parent = ASSIGN
\checkstyle\src\test\resources\com\puppycrawl\tools\checkstyle\checks\whitespace\InputWhitespace.java
([29x18] - parent = PLUS
[...]

Issue #5945, opened 2018-06-18
New Check: RightCurlyOfExpression


romani:
based on discussion at #4058 (PR #4173)

right curly of expression block is different from other blocks so rules or RightCurly Check does not work well as for example for `alone` results in weird code as `}` in expression is always followed by smth, `same` does not make sense .

[Array Creation Expressions](https://docs.oracle.com/javase/specs/jls/se9/html/jls-15.html#jls-15.10.1)
[Lambda expression](https://docs.oracle.com/javase/specs/jls/se9/html/jls-15.html#jls-15.27 )
[Class Instance Creation Expressions](https://docs.oracle.com/javase/specs/jls/se9/html/jls-15.html#jls-15.10.1)

on_same_line
```
ignoredAnnotations = Arrays.stream(new String[] {"Test", "Before", "After",
        "BeforeClass", "AfterClass", }).collect(Collectors.toSet());
final Class<?>[] param = new Class<?>[] {String.class, String.class,
       String.class, Attributes.class,};
```
on_new_line
```
ignoredAnnotations = Arrays.stream(new String[] {"Test", "Before", "After",
        "BeforeClass", "AfterClass",
}).collect(Collectors.toSet());
final Class<?>[] param = new Class<?>[] {String.class, String.class,
       String.class, Attributes.class,
};
```
on_new_line_or_singleline
```
ignoredAnnotations = Arrays.stream(new String[] {"Test", "Before",}).collect(Collectors.toSet());
ignoredAnnotations = Arrays.stream(new String[] {"Test", "Before", "After",
        "BeforeClass", "AfterClass",
}).collect(Collectors.toSet());
final Class<?>[] param = new Class<?>[] {String.class, String.class,
       String.class, Attributes.class,
};
final Class<?>[] param = new Class<?>[] {String.class, Attributes.class, };
```

For multi-dimensional arrays we validate only outer curly brace
```
int array[][] = { {1, 2}, {3,4} };
``` 
we will validate only "};".

rnveach:
xdoc in https://github.com/checkstyle/checkstyle/pull/5947 has to be updated to point to new check

SGanguly1999:
I am on it.

Issue #5234, opened 2017-10-25
Remove DetailAST.branchContains


rnveach:
Split from https://github.com/checkstyle/checkstyle/issues/5124 ,

We should remove `branchContains` completely from `DetailAST`.

This issue is on hold until the other issue is finished and when we decide when we wish to break compatibility.

romani:
We need to deprecate it, completely forbid to use it in our code by some Check and .... - all in scope of #5124

remove in several versions completely.

rnveach:
@romani What do you think of also removing the Javadoc equivalent which also isn't used in checkstyle anywhere.
https://github.com/checkstyle/checkstyle/blob/6bab0788365ba9afb713e14d9f6263f1f60b7aa2/src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java#L230

rnveach:
This is blocked by https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/669 .

romani:
From https://github.com/checkstyle/checkstyle/pull/7424#issuecomment-569995176

Can we have some/any utility method that can be used instead ? to mention it in migration notes and let users use it to keep functionality the same, even it is not efficient approach.

Update of legacy Checks might be very painful process, users will likely to agree on anything to keep it as is.

We will test such migration on sevntu Checks. If we succeed sevntu migration, we can sure that user will have barrable size of problems during upgrade.

rnveach:
Yes we can always create a utility that behaves the same but then we will have to keep it in main repo and require it to meet all our requirements, 100% coverage and pitest passage. All this will do is kick the can down the road as users will just switch to the new utility and not rework this code to use this type of method. My aim was to completely remove this method.

If we are to build a duplicate of this then why not just leave the method as it is.

romani:
>require it to meet all our requirements, 100% coverage and pitest passage.

Deprecation side effect stuff can have pure UTs, so coverage will be technical.

>If we are to build a duplicate of this then why not just leave the method as it is.

We remove it from API, deprecation in Util will be less accurate. We can remove it later on to force users just copy method code to their codebase if they want no heavy refactoring.

nmancus1:
Can we just create a separate issue to deprecate now, then take care of removal after https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/669 ?

rnveach:
[...]

PR #14792, opened 2024-04-13
Issue #14747: fix ParenPad to not flag unsupported Tokens

mahfouz72:
Resolves: #14747 
Resolves: #4175 

- removed deep scan from `processExpression` and just scanned the first level of a given node (the direct children only not the whole subtree)
- adding some tokens that got missed from having no deep scan

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/2b480a919e3a98f5609aeab17fb79b1b/raw/23cfa105f73bc7149494f1e0712be2376ad13fbd/parenpadbase.xml

Diff Regression patch config: https://gist.githubusercontent.com/mahfouz72/79f7d3b270b52b91da18557854ce6e60/raw/9d6c487e7b1df6d511ec2a808566db21759b91e1/parenpadpatch.xml


src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheck.java
```diff
@@ -253,31 +267,10 @@ private void processExpression(DetailAST ast) {
...
-                currentNode = currentNode.getFirstChild();
-                continue;
-            }
-
-            // Go up after processing the last child
-            while (currentNode.getNextSibling() == null && currentNode.getParent() != ast) {
-                currentNode = currentNode.getParent();
-            }
             currentNode = currentNode.getNextSibling();
         }
```
mahfouz72, line 271: removed the deep scan and just check the first child and all of its siblings

```
PS D:\test> cat src/Test3.java

public class Test3 {
    int i =  (( (5*4) + ( 4 + 2)));
}
PS D:\test> java  -jar checkstyle-10.14.2-all.jar -t  src/Test3.java
COMPILATION_UNIT -> COMPILATION_UNIT [2:0]
`--CLASS_DEF -> CLASS_DEF [2:0]
    |--MODIFIERS -> MODIFIERS [2:0]
    |   `--LITERAL_PUBLIC -> public [2:0]
    |--LITERAL_CLASS -> class [2:7]
    |--IDENT -> Test3 [2:13]
    `--OBJBLOCK -> OBJBLOCK [2:19]
        |--LCURLY -> { [2:19]
        |--VARIABLE_DEF -> VARIABLE_DEF [3:4]
        |   |--MODIFIERS -> MODIFIERS [3:4]
        |   |--TYPE -> TYPE [3:4]
        |   |   `--LITERAL_INT -> int [3:4]
        |   |--IDENT -> i [3:8]
        |   |--ASSIGN -> = [3:10]
        |   |   `--EXPR -> EXPR [3:13]
        |   |       |--LPAREN -> ( [3:13]         <-- this and all its siblings only checked 
        |   |       |--LPAREN -> ( [3:14]
        |   |       |--PLUS -> + [3:22]
        |   |       |   |--LPAREN -> ( [3:16]          <-- no deep scan so this will not be picked
        |   |       |   |--STAR -> * [3:18]
        |   |       |   |   |--NUM_INT -> 5 [3:17]
        |   |       |   |   `--NUM_INT -> 4 [3:19]
        |   |       |   |--RPAREN -> ) [3:20]
        |   |       |   |--LPAREN -> ( [3:24]
        |   |       |   |--PLUS -> + [3:28]
        |   |       |   |   |--NUM_INT -> 4 [3:26]
        |   |       |   |   `--NUM_INT -> 2 [3:30]
        |   |       |   `--RPAREN -> ) [3:31]
        |   |       |--RPAREN -> ) [3:32]
        |   |       `--RPAREN -> ) [3:33]
        |   `--SEMI -> ; [3:34]
        `--RCURLY -> } [4:0]

```

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheck.java
```diff
@@ -306,6 +299,15 @@ private static int[] makeAcceptableTokens() {
...
             TokenTypes.RECORD_DEF,
+            TokenTypes.TYPECAST,
+            TokenTypes.STAR,
+            TokenTypes.PLUS,
+            TokenTypes.MINUS,
+            TokenTypes.DIV,
+            TokenTypes.MOD,
+            TokenTypes.LAND,
+            TokenTypes.LOR,
+            TokenTypes.LNOT,
```
mahfouz72, line 310: add some tokens that got missed due to having no deep scan (we may need to add some more tokens but those were the obvious one from the test and input files)

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/ParenPadCheck.java
```diff
@@ -253,31 +267,10 @@ private void processExpression(DetailAST ast) {
...
-    /**
-     * Checks whether AcceptableTokens contains the given ast.
-     *
-     * @param ast the token to check.
-     * @return true if the ast is in AcceptableTokens.
-     */
-    private boolean isAcceptableToken(DetailAST ast) {
-        return acceptableTokens.get(ast.getType());
-    }
-
```
mahfouz72, line 280: we don't need this anymore we needed it to avoid checking the same node twice. while doing a deep scan when we found a subtree that is acceptableToken we don't visit it because it will be picked up with `visitToken()` later

now there is no deep scan for tokens so we don't need this check. I removed this method, the related unnecessary class fields and the test corresponding to it

mahfouz72:
Github, generate report

mahfouz72:
Github, generate site

mahfouz72:
there are huge differences.
I want to know if I'm on the right track before proceeding to add more tokens.

romani:
Please extend link check suppression files with lines as CI suggesting.

mahfouz72:
@romani done. should I add all tokens that got missed and identified in the regression report in this PR?

romani:
I do not recommend to rush to add tokens, some of them not added for good reason. Let's add one by one, with full attention to regression diff report.

mahfouz72:
one by one in separate PRs or in this PR?
and also what about the tokens added till now? all of them were added to cover the missing parnes that are not checked in input files after having no deepscan

nrmancuso:
>I do not recommend to rush to add tokens, some of them not added for good reason. Let's add one by one, with full attention to regression diff report.

@romani I don't think we can do this without a bunch of hacks; as soon as we stop "deep scanning", we need to extend the tokens for this check to keep the behavior consistent.

@mahfouz72 can you help us to understand what other tokens we may need to add here, and how you are discovering them?

mahfouz72:
@nrmancuso we need to add any possible token that could be used under `EXPR` token. for example, all mathematical, logical, and bitwise operators should be added. for now, I added some of them I discovered those from the failing unit tests after having no deep scan. but there are more tokens that we didn't use in the input file

Examples:
- bitwise:  & ,  | ,   ~ ,  ^  ,  << ,  >>
- comparison : <   <=   > >=  ==  !=
- assignments :   +=   -=    =     /=  *=  etc..

I am discovering them from the regression report and in general as I stated above I need to think of any token that can be used under expression and can be surrounded by paren

what do you think should I start working in this way? and pay full attention to the report to have consistent behaviour

rnveach:
@nrmancuso @romani Could this be a sign we need a ParenPadExpression check? Will someone ever want different spacing for different expression tokens?

mahfouz72:
> Will someone ever want different spacing for different expression tokens?

IMO, No, no one will need to violate this  `x = (3+5)`  but not this `x += (3+5)`. That way I see this step https://github.com/checkstyle/checkstyle/pull/14792#issuecomment-2088501154 making the check weird and unnecessary a mess of tokens (we are talking about 15 ~ 20 new tokens)

but at the same time, how to pick those cases that got missed after having no deep scan without adding all those tokens....

I don't know if this a good design but can we leave the deep scan and while scanning we skip tokens that are not any of the tokens mentioned here https://github.com/checkstyle/checkstyle/pull/14792#issuecomment-2088501154 so we will have 
`isMathematicalOperator()` , `isBitwiseOperator()`,  etc... and skip the validation while deep scanning based on the type of token
So basiaclly this solution is almost the same as  https://github.com/checkstyle/checkstyle/pull/14792#issuecomment-2088501154 but we won't explicity  add them as new token of the check we will be checking them internally while deep scanning

rnveach:
> but can we leave the deep scan and while scanning we skip tokens that are not any of the tokens mentioned here 

What is the difference between leaving the deep scanning (dropping the issue) or leaving the deep scanning while checking this internal only list ? My understanding was you built the list from the deep scanning (and/or regression).

I was thinking along the lines we break this check apart. We drop expression support here and create an expression only check (we want to do this for another check, https://github.com/checkstyle/checkstyle/issues/5945 ), specify all the tokens you suggested, but make them non-configurable (must stop on them) and no deep scanning.

The problem with deep scanning is its easy to lose control and its harder to understand what it is exactly looking at. We had issues before where a check went beyond its boundaries in scanning. We are sort of having a discussion like this for UnusedLocalVariableCheck regarding pitest. Another example is https://github.com/checkstyle/checkstyle/issues/5234 and https://github.com/checkstyle/checkstyle/issues/5124 which I specifically mentioned `branchContains is a dangerous method to use all the time since there is no restriction on how deep it can search`.

mahfouz72:
> What is the difference between leaving the deep scanning (dropping the issue) or leaving the deep scanning while checking this internal only list ?

if we leave deep scan and check  this internal list only. we will avoid validating tokens thay definitely should not be validated example: `RECORD_PATREN_DEF` that I have in the issue related to this PR

one of the main problems of the deepscan that we was checking token that not in configuration if we enforce checking this internal list only we will avoid this problem



> We drop expression support here and create an expression only check

this is a good solution I am leaning to it

mahfouz72:
we can simplify this check by making the required Tokens `LPAREN`, `RPAREN`. So we always stop on both tokens only and check the Pad for each and every `LPAREN` or `RPAREN` as simple as that. In this case no need for a special check for expressions we will stop on all.

but with this design, the check will not be configurable users can't customize it for specific tokens so I dunno if this is good

nrmancuso:
>but with this design, the check will not be configurable users can't customize it for specific tokens so I dunno if this is good

I propose that we do the following:
1. Open an issue for making a simple check that is only configurable like EsLint's [check](https://eslint.style/rules/js/space-in-parens#options).
2. Close this
3. Allow us to discuss this further in the issue

Generally, I like to start with some simple concept like this and get it out to users and get real feedback, instead of trying to dream up everything a user could ask for initially (this is always doomed to fail, since lots of folks write code differently).

Also, what users may have liked 10 years ago is not necessarily what they like now. 

I think it is time to start over with this check.

mahfouz72:
https://github.com/checkstyle/checkstyle/issues/14896#issue-2317262264

PR #14795, opened 2024-04-15
Issue #14715: Enforced new naming convention in IT area 16

MANISH-K-07:
Part of #14715


src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionIllegalTypeTest.java
```diff
@@ -63,7 +63,7 @@ public void testOne() throws Exception {
     @Test
     public void testTwo() throws Exception {
```
romani, line 64: Do you plan to fix such names to match input suffix?

romani: Or better to do this in separate PR ? I think so

romani (approved):

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionIllegalTypeTest.java
MANISH-K-07: @romani , in fact I've been changing respective test method names too in these PRs itself..

The problem with a few input files is that there is no proper design on how they have been introduced to our code by the authors. I'm unable to come up with meaningful suffixes for such files, and am leaving those abstract (One, Two, ....)

Rest assured, the input file suffixes and test names, I always make sure they match.
Even here, `testTwo` because it takes file `InputXpathIllegalTypeTwo`

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionInnerAssignmentTest.java
```diff
@@ -38,9 +38,9 @@ protected String getCheckName() {
     }
 
     @Test
-    public void testFile1() throws Exception {
+    public void testInnerAssignment() throws Exception {
```
MANISH-K-07, line 41: like here, name also has been changed :)

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionInterfaceIsTypeTest.java
```diff
@@ -65,9 +65,9 @@ public void testOne() throws Exception {
     }
 
     @Test
-    public void testTwo() throws Exception {
+    public void testAllowMarkerFalse() throws Exception {
```
MANISH-K-07, line 68: like here, test names also have been changed :)

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionIllegalTypeTest.java
romani: Ok, thanks a lot for details

PR #14804, opened 2024-04-16
Issue #13086: fix InnerAssignmentCheck In SwitchRuleAndExpression

### LINKED ISSUE #13086: InnerAssignmentCheck failed for one line code in Java 14 switch expression
I have read check documentation: https://checkstyle.org/config_coding.html#InnerAssignment

```bash
/var/tmp $ javac TestInnerAssignmentCheck.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <module name="TreeWalker">
    <module name="InnerAssignment"/>
  </module>
</module>

/var/tmp $ cat TestInnerAssignmentCheck.java
public class TestInnerAssignmentCheck {
    private boolean flag;

    void getOperation(final String operation) {
        switch (operation) {
            case "Y" -> flag = true;  // violation 
            case "N" -> {
                flag = false;
            }
            default -> throw new UnsupportedOperationException();
        }
    }
}

/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.11.0-all.jar -c config.xml TestInnerAssignmentCheck.java
Starting audit...
[ERROR] /Users/hael/Code/Learn/Tools/checkstyle/TestInnerAssignmentCheck.java:6:30:
     Inner assignments should be avoided. [InnerAssignment]
Audit done.
```

---

Checkstyle behaves differently between one line code and code block for Java 14's Switch Expression. In the given code above, checkstyle will complain about `case "Y" -> flag = true;` but not `case "N" -> {flag = false;}`. I'm afraid that it's because `flag = true;` is equivalent to `return flag = true;`? But the return value is not used here in the switch expression, we just want to do the value assignment. 

Expected: no violations on such code.

mahfouz72:
Resolves #13086 :

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/70e0fb4b76668949fd9e0fd3744deddd/raw/0b035184445fee1009e49543af7b1daa6b52f79f/innerass.xml

Diff Regression projects: https://gist.githubusercontent.com/mahfouz72/297d7353d2d0bcec178e1b098109cd9e/raw/546dcdbdff363110fa265ea80d70e36f079c6ee8/projects-to-test-on.properties

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheck.java
@@ -101,6 +101,7 @@ public class InnerAssignmentCheck
             TokenTypes.RESOURCE_SPECIFICATION,
         },
         {TokenTypes.EXPR, TokenTypes.LAMBDA},
+        {TokenTypes.EXPR, TokenTypes.SWITCH_RULE, TokenTypes.LITERAL_SWITCH, TokenTypes.SLIST},
     };
 
     /**
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheckTest.java
@@ -109,4 +109,24 @@ public class InnerAssignmentCheckTest
             .isNotNull();
     }
 
+    @Test
+    public void testInnerAssignmentSwitchAndSwitchExpression() throws Exception {
+        final String[] expected = {
+            "28:23: " + getCheckMessage(MSG_KEY),
+            "38:25: " + getCheckMessage(MSG_KEY),
+            "40:25: " + getCheckMessage(MSG_KEY),
+            "41:26: " + getCheckMessage(MSG_KEY),
+            "49:25: " + getCheckMessage(MSG_KEY),
+            "51:31: " + getCheckMessage(MSG_KEY),
+            "52:26: " + getCheckMessage(MSG_KEY),
+            "59:42: " + getCheckMessage(MSG_KEY),
+            "61:34: " + getCheckMessage(MSG_KEY),
+            "94:25: " + getCheckMessage(MSG_KEY),
+            "96:26: " + getCheckMessage(MSG_KEY),
+            "98:27: " + getCheckMessage(MSG_KEY),
+        };
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputInnerAssignmentSwitchAndSwitchExpression.java"),
+                expected);
+    }
 }
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/InputInnerAssignmentSwitchAndSwitchExpression.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/InputInnerAssignmentSwitchAndSwitchExpression.java
@@ -1,0 +1,101 @@
+/*
+InnerAssignment
+
+*/
+
+//non-compiled with javac: Compilable with Java14
+package com.puppycrawl.tools.checkstyle.checks.coding.innerassignment;
+
+public class InputInnerAssignmentSwitchAndSwitchExpression {
+
+    public void test1(int mode) {
+        int x = 0;
+        switch (mode) {
+            case 2 -> {
+                x = 2;
+            }
+            case 1 -> x = 1;
+        }
+    }
+
+    public void test2(int mode) {
+        int x = 0, y = 0;
+        switch (mode) {
+            case 2, 4, 6 -> {
+                x = 2;
+            }
+            case 1, 3, 5 -> {
+                x = y = 1;  // violation
+            }
+            case 0, 7, 8 -> x = 1;
+        }
+    }
+
+    public void test3(int mode) {
+        int x = 0;
+        x = switch (mode) {
+            case 2 -> {
+                yield x = 2; // violation
+            }
+            case 1 -> x = 1;  // violation
+            default -> x = 0; // violation
+        };
+    }
+
+    public void test4(int mode) {
+        int x = 0;
+        x = switch (mode) {
+            case 2, 4, 6 -> {
+                yield x = 2; // violation
+            }
+            case 1, 3, 5 -> x = 1;  // violation
+            default -> x = 0; // violation
+        };
+    }
+
+    public void test5(String operation) {
+        boolean innerFlag, flag;
+        switch (operation) {
+            case "Y" -> flag = innerFlag = true; // violation
+            case "N" -> {
+                flag = innerFlag = false; // violation
+            }
+        }
+    }
+
+    public void test6(int mode) {
+        int x = 0;
+        switch (mode) {
+            case 2: {
+                x = 2;
+            }
+            case 1:
+                x = 1;
+        }
+    }
+
+    public void test7(int mode) {
+        int x = 0;
+        switch (mode) {
+            case 0:
+            case 1:
+            case 2: {
+                x = 2;
+            }
+            case 4:
+            case 5:
+                x = 1;
+        }
+    }
+
+    public void test8(int mode) {
+        int x = 4;
+        System.out.println(switch (x) {
+            case 1 -> x = 1;   // violation
+            case 2 -> {
+                 yield x = 2;      // violation
+            }
+            default ->  x = 3;     // violation
+        });
+    }
+}
--- a/src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/Example2.java
+++ b/src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/Example2.java
@@ -1,0 +1,47 @@
+/*xml
+<module name="Checker">
+  <module name="TreeWalker">
+    <module name="InnerAssignment"/>
+  </module>
+</module>
+*/
+
+//non-compiled with javac: Compilable with Java14
+package com.puppycrawl.tools.checkstyle.checks.coding.innerassignment;
+
+
+// xdoc section -- start
+public class Example2 {
+  public void test1(int mode) {
+    int x = 0;
+    x = switch (mode) {
+      case 1 -> x = 1;  // violation
+      case 2 -> {
+        yield x = 2; // violation
+      }
+      default -> x = 0; // violation
+    };
+  }
+  public void test2(int mode) {
+    int x = 0;
+    switch(mode) {
+      case 2 -> {
+        x = 2;
+      }
+      case 1 -> x = 1;
+    }
+  }
+  public void test3(int mode) {
+    int x = 0, y = 0;
+    switch(mode) {
+      case 1:
+      case 2: {
+        x = y = 2; // violation
+      }
+      case 4:
+      case 5:
+        x = 1;
+    }
+  }
+}
+// xdoc section -- end
--- a/src/xdocs-examples/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheckExamplesTest.java
+++ b/src/xdocs-examples/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheckExamplesTest.java
@@ -45,4 +45,17 @@ public class InnerAssignmentCheckExamplesTest extends AbstractExamplesModuleTest
 
         verifyWithInlineConfigParser(getPath("Example1.java"), expected);
     }
+
+    @Test
+    public void testExample2() throws Exception {
+        final String[] expected = {
+            "18:19: " + getCheckMessage(MSG_KEY),
+            "20:17: " + getCheckMessage(MSG_KEY),
+            "22:20: " + getCheckMessage(MSG_KEY),
+            "39:15: " + getCheckMessage(MSG_KEY),
+        };
+
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("Example2.java"), expected);
+    }
 }
--- a/src/xdocs/checks/coding/innerassignment.xml
+++ b/src/xdocs/checks/coding/innerassignment.xml
@@ -57,7 +57,7 @@ while ((line = bufferedReader.readLine()) != null); // OK
   &lt;/module&gt;
 &lt;/module&gt;
         </source>
-        <p id="Example1-code">Example:</p>
+        <p id="Example1-code">Example 1:</p>
         <source>
 public class Example1 {
   void foo() throws IOException {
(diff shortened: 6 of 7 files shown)


mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

mahfouz72:
report is good

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheck.java
```diff
@@ -101,6 +101,7 @@ public class InnerAssignmentCheck
             TokenTypes.RESOURCE_SPECIFICATION,
         },
         {TokenTypes.EXPR, TokenTypes.LAMBDA},
+        {TokenTypes.EXPR, TokenTypes.SWITCH_RULE, TokenTypes.LITERAL_SWITCH, TokenTypes.SLIST},
```
romani, line 104: Why we need TokenTypes.SLIST ? And TokenTypes.LITERAL_SWITCH

romani (changes requested):
Question

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheck.java
mahfouz72: To make sure that we are in normal Switch context not Switch Expression. if it is  `{TokenTypes.EXPR, TokenTypes.SWITCH_RULE}` only then we will have no violation if the assignment is in switch expression
ex 
```java
    public void test2(int mode) {
        int x = 0;
        x = switch (mode) {
            case 2 -> {
                yield x = 2; 
            }
            case 1 -> x = 1;  // this will be ok
            default -> throw new UnsupportedOperationException();
        };
    }
``` 

Switch Expression AST:
```
        |--EXPR -> EXPR [8:10]
        |       |   `--ASSIGN -> = [8:10]   <<---
        |       |       |--IDENT -> x [8:8]
        |       |       `--LITERAL_SWITCH -> switch [8:12]   <<---
        |       |           |--LPAREN -> ( [8:19]
        |       |           |--EXPR -> EXPR [8:20]
        |       |           |   `--IDENT -> mode [8:20]
        |       |           |--RPAREN -> ) [8:24]
        |       |           |--LCURLY -> { [8:26]
        |       |           |--SWITCH_RULE -> SWITCH_RULE [9:12]   <<---
        |       |           |   |--LITERAL_CASE -> case [9:12]
        |       |           |   |   `--EXPR -> EXPR [9:17]
        |       |           |   |       `--NUM_INT -> 1 [9:17]
        |       |           |   |--LAMBDA -> -> [9:19]
        |       |           |   |--EXPR -> EXPR [9:24]         <<---
        |       |           |   |   `--ASSIGN -> = [9:24]
        |       |           |   |       |--IDENT -> x [9:22]
        |       |           |   |       `--NUM_INT -> 1 [9:26]
        |       |           |   `--SEMI -> ; [9:27]
        |       |           `--RCURLY -> } [10:8]
```
switch rule AST

```
        `--SLIST -> { [6:33]          <<---
        |       |--LITERAL_SWITCH -> switch [7:8]   <<---
        |       |   |--LPAREN -> ( [7:15]
        |       |   |--EXPR -> EXPR [7:16]
        |       |   |   `--IDENT -> mode [7:16]
        |       |   |--RPAREN -> ) [7:20]
        |       |   |--LCURLY -> { [7:22]
        |       |   |--SWITCH_RULE -> SWITCH_RULE [8:12]          <<---
        |       |   |   |--LITERAL_CASE -> case [8:12]
        |       |   |   |   `--EXPR -> EXPR [8:17]
        |       |   |   |       `--NUM_INT -> 1 [8:17]
        |       |   |   |--LAMBDA -> -> [8:19]
        |       |   |   |--EXPR -> EXPR [8:24]       <<---
        |       |   |   |   `--ASSIGN -> = [8:24]
        |       |   |   |       |--IDENT -> x [8:22]
        |       |   |   |       `--NUM_INT -> 1 [8:26]
        |       |   |   `--SEMI -> ; [8:27]
        |       |   `--RCURLY -> } [9:8]

```

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/InputInnerAssignmentSwitchAndSwitchExpression.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public void test2(int mode) {
+        int x = 0;
+        x = switch (mode) {
+            case 2 -> {
+                yield x = 2; // violation
+            }
+            case 1 -> x = 1;  // violation
+            default -> throw new UnsupportedOperationException();
```
nrmancuso, line 28: Please add inner assignment to default branch, and create another method with the same pattern, but with switch labeled statement groups. Also, add an example with multiple case constants in one or more labels.

nrmancuso (changes requested):
Please review documentation to see if we need to update there as well.

Item:

mahfouz72:
Github, generate site

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/InputInnerAssignmentSwitchAndSwitchExpression.java
mahfouz72: done and updated documentation to add examples with switch

nrmancuso:
@mahfouz72 please generate a regression report against openjdk21

nrmancuso (approved):

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheck.java
mahfouz72: @romani ping

romani (approved):
Ok to merge

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheckTest.java
```diff
@@ -99,6 +99,7 @@ public void testInnerAssignmentNotInLoopContext() throws Exception {
     public void testTokensNotNull() {
         final InnerAssignmentCheck check = new InnerAssignmentCheck();
         assertWithMessage("Acceptable tokens should not be null")
+
```
rnveach, line 102: remove unnecessary change.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/InputInnerAssignmentSwitchAndSwitchExpression.java
```diff
@@ -0,0 +1,85 @@
...
+            }
+            case 1, 3, 5 -> x = 1;
+        }
+    }
+
+    public void test3(int mode) {
+        int x = 0;
+        x = switch (mode) {
+            case 2 -> {
+                yield x = 2; // violation
```
rnveach, line 35: So why does this create violations but something like test 2 doesn't? Is this because it is basically nested inside an assignment already?

Is there somewhere (official?) that shows what inner assignments are in this case?

What if we had a case like (also add it):
````
System.out.println(switch (mode) {
            case 2 -> {
                yield x = 2;
````

This wouldn't be an exception?

File: src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/InputInnerAssignmentSwitchAndSwitchExpression.java [hunk 5]
```diff
@@ -0,0 +1,85 @@
...
+            }
+            case 1 -> x = 1;
+        }
+    }
+
+    public void test2(int mode) {
+        int x = 0;
+        switch(mode) {
+            case 2, 4, 6 -> {
+                x = 2;
```
rnveach, line 25: Create another test method like:
````
   public void test2(int mode) {
        int x = 0;
        switch(mode) {
            case 2, 4, 6 -> {
                x = y = 2;
````

I would expect a violation here.

File: src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/Example2.java [hunk 6]
```diff
@@ -0,0 +1,47 @@
...
+<module name="Checker">
+  <module name="TreeWalker">
+    <module name="InnerAssignment"/>
+  </module>
+</module>
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.innerassignment;
+
+//non-compiled with javac: Compilable with Java14
```
rnveach, line 11: @romani @nrmancuso Do we check that these files are compilable with different java versions? I don't remember seeing this before with examples.
Shouldn't this also be before the package?

rnveach (changes requested):
There is not really any descriptions update in this PR.
https://checkstyle.org/checks/coding/innerassignment.html#Description
> Except for the loop idioms, all assignments should occur in their own top-level statement to increase readability. 

Is this wording technically not correct anymore? I wouldn't really consider cases to be top level statements.

rnveach:
> Diff Regression projects: https://gist.githubusercontent.com/mahfouz72/297d7353d2d0bcec178e1b098109cd9e/raw/546dcdbdff363110fa265ea80d70e36f079c6ee8/projects-to-test-on.properties

We normally provide regression on all projects. Even if no others have this form of code, it will show this doesn't impact older code.

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheckTest.java
mahfouz72: done. it was by mistake

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/InputInnerAssignmentSwitchAndSwitchExpression.java
mahfouz72: > So why does this create violations but something like test 2 doesn't

because this is an assignment inside expressions ( switch expressions ), while test2 is an assignment inside normal switch statements

> Is there somewhere (official?)

I didn't find any but what I know from our docs is that inner assignments are the assignments inside expressions. So basically the assignments inside switch expressions `x = switch(){....} `  should be avoided

> What if we had a case like (also add it):

Yes, violation and added

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/InputInnerAssignmentSwitchAndSwitchExpression.java
mahfouz72: added in the same method

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/Example2.java
mahfouz72: I put it before the package

mahfouz72:
> We normally provide regression on all projects. Even if no others have this form of code, it will show this doesn't impact older code. Please update the file.

Yes this is only to generate another report I generated two

all projects:  https://github.com/checkstyle/checkstyle/pull/14804#issuecomment-2060099135
java 21: https://github.com/checkstyle/checkstyle/pull/14804#issuecomment-2066948221

mahfouz72:
> Is this wording technically not correct anymore?

I don't see anything wrong. Can you please point to an example that you think violates this wording

`x = switch(..){ case 1 -> x = 3} `  (assignment is not in its own top level (nested inside another assignment)  so violation
`switch(..){ case 1 -> x = 3} `  ( I guess this is considered  as a top-level so no violation )

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/Example2.java
nrmancuso: @rnveach We only compile with the specified version.

nrmancuso:
> There is not really any descriptions update in this PR. https://checkstyle.org/checks/coding/innerassignment.html#Description
> 
> > Except for the loop idioms, all assignments should occur in their own top-level statement to increase readability.
> 
> Is this wording technically not correct anymore? I wouldn't really consider cases to be top level statements.

I think this still holds for switch expressions; the only assignment happening should be to a variable declared outside(top level) of the switch expression, from a value yielded by the switch expression.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/InputInnerAssignmentSwitchAndSwitchExpression.java
nrmancuso: >Is there somewhere (official?) that shows what inner assignments are in this case?

Doubt it. A better way to describe what this check does is "place violations (mostly) anywhere that assigned values are used directly". Ideally, we should only do one thing at a time; assign a value, or use it, not both.

rnveach: > yield x = 2;

Just to confirm, since I am not use to this syntax, is this the same as `x = 2; yield x;`?
If we had `y = 2` before this line would that be a violation too?

@nrmancuso @romani Are we ok that test8 is a violation for this check?

rnveach:
I think I need some clarification then on what is a "top level statement" is.

I tried google. The first thing said `Top-level statements allows you to write executable code directly at the root of a file, eliminating the need for wrapping your code in a class or method. `. However that seems to be a C# link. Then I got a reddit for Java help where someone replied to the poster to clarify this exact statement and it was never answered.

I always considered this check to mean something like "the statement must only contain the assignment itself", but then the whole switch is considered a single statement (unless I have the Java grammar wrong).

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/innerassignment/InputInnerAssignmentSwitchAndSwitchExpression.java
mahfouz72: > Just to confirm, since I am not use to this syntax, is this the same as `x = 2; yield x;`?

Yes I think So

> If we had `y = 2` before this line would that be a violation too?

this will have no violation due to https://github.com/checkstyle/checkstyle/blob/405db25d46c4cb9ac561f91e04c59aebac50baf6/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheck.java#L94

nrmancuso: Yes, assigned values are used directly in all three cases.

romani:
>but then the whole switch is considered a single statement (unless I have the Java grammar wrong).

It is probably very confusing to consider switch expressions as statement. It better to think of this like inlined function in such cases.

Detect that we have is that we report violation on switch that statement like.

rnveach:
I am not calling into question what we print out as violations. My whole comments was the documentation and is it clear enough on what we say this check does.

romani:
https://checkstyle.org/checks/coding/innerassignment.html#Description

I don't know how to improve our doc, we have have not very strict target of vali

PR #14819, opened 2024-04-21
Issue #11166: remove getFileContents from EmptyLineSeparator

mahfouz72:
Issue #11166 : 
refactor  EmptyLineSeparatorCheck to use AST-based logic not `getFileContents()`

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/9384f843297bb2e10c074bcb33ddc46902d6bd52/check.xml

**WIP**. just wants to get an initial review from CIs and maintainers

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
@@ -622,19 +622,16 @@ public class EmptyLineSeparatorCheck extends AbstractCheck {
      * @return true if token have empty line after.
      */
     private boolean hasEmptyLineAfter(DetailAST token) {
-        DetailAST lastToken = token.getLastChild().getLastChild();
-        if (lastToken == null) {
-            lastToken = token.getLastChild();
+        DetailAST currentEnd = token.getLastChild().getLastChild();
+        if (currentEnd == null) {
+            currentEnd = token.getLastChild();
         }
-        DetailAST nextToken = token.getNextSibling();
-        if (TokenUtil.isCommentType(nextToken.getType())) {
-            nextToken = nextToken.getNextSibling();
+        DetailAST nextBegin = token.getNextSibling();
+        if (TokenUtil.isCommentType(nextBegin.getType())) {
+            nextBegin = nextBegin.getNextSibling();
         }
-        // Start of the next token
-        final int nextBegin = nextToken.getLineNo();
-        // End of current token.
-        final int currentEnd = lastToken.getLineNo();
-        return hasEmptyLine(currentEnd + 1, nextBegin - 1);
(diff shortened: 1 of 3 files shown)


mahfouz72:
dealing with comment tokens in the AST was weird to me (the ordering and the location)

@romani @nrmancuso @rnveach 
do we have any check that is aware of comments using AST-based logic not file content to take it as a reference?

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
```diff
@@ -652,28 +649,62 @@ private static Optional<DetailAST> findCommentUnder(DetailAST packageDef) {
...
+        final DetailAST modifersAst = ast.findFirstToken(TokenTypes.MODIFIERS);
+        final DetailAST commentAst;
+
+        if (modifersAst != null) {
+            commentAst = modifersAst.getFirstChild();
+        }
+        else {
+            commentAst = ast.getFirstChild();
+        }
+        return commentAst;
```
mahfouz72, line 707: I will work more on this method. I believe that this will cause NPE. more cases need to be handled to get the first comment token

mahfouz72, line 707: I will work more on this method. I believe that there are more cases that need to be handled to get the first comment token

mahfouz72:
Github, generate report

mahfouz72:
> dealing with comment tokens in the AST was weird to me. why it is under modifiers of the next subtree. is there a reason to build AST like this?

@nrmancuso ping

nrmancuso:
> dealing with comment tokens in the AST was weird to me. why it is under modifiers of the next subtree. is there a reason to build AST like this?

Yeah, comment placement in the AST is super weird, and has been for a long time. We have at least one issue on this. I can’t really imagine this check working well without getting file contents honestly. 

IMO, this check should continue to use getFileContents

mahfouz72:
> Yeah, comment placement in the AST is super weird, and has been for a long time. We have at least one issue on this. I can’t really imagine this check working well without getting file contents honestly.

Yes, I tried a few times to continue this PR but things always get very messy and error-prone. I couldn't reach a good design. we should consider refactoring comments AST first before migrating this check to AST-based logic

> IMO, this check should continue to use getFileContents

same and just wanted to know your opinion

nrmancuso:
Let’s make sure other maintainers are on board. @rnveach @romani ?

rnveach:
I am.

nrmancuso:
We will not be able to complete this issue before GSOC. @mahfouz72 let's close this for now, and come back after summer. All your work will be saved via the link in the issue, and you can keep this branch in your fork to pick up later.

PR #14831, opened 2024-04-25
Issue #13213: Remove '//ok' comments from InputJavadocMethodExtraThrows

prathm3:
Issue: #13213 

Removed //ok from InputJavadocMethodExtraThrows also removed suppression for that from xml file


romani (approved):
Ok to merge. Thanks a lot

PR #14839, opened 2024-04-28
minor: removes extra method call from UnusedLocalVariableCheck

rnveach:
This removes an if condition, and moves the if condition inside the method call to the outside. Nothing else was changed besides indentation.

The conditions we removed/moved before this change basically amounts to:
> TokenUtil.isTypeDeclaration && isNonLocalTypeDeclaration

[`isNonLocalTypeDeclaration`](https://github.com/checkstyle/checkstyle/blob/55f5c966bfa61b387110db6c4780f259a21676c8/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L389) is basically:
> TokenUtil.isTypeDeclaration && !slist

So basically we are calling `TokenUtil.isTypeDeclaration` twice, and my changes remove the extra call.

Also, this code is in the `visitToken` and this change [aligns us with the `leaveToken` which has the same if condition check](https://github.com/checkstyle/checkstyle/blob/55f5c966bfa61b387110db6c4780f259a21676c8/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L251).

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java
@@ -232,8 +232,8 @@ public class UnusedLocalVariableCheck extends AbstractCheck {
         else if (isInsideLocalAnonInnerClass(ast)) {
             visitLocalAnonInnerClass(ast);
         }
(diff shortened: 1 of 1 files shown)


romani (approved):
Confirmed

PR #14858, opened 2024-05-05
Issue #14825: fix false positive for WhitespaceAroundCheck

### LINKED ISSUE #14825: WhitespaceAround reports a violation when switch expression is passed as a method argument
I have read check documentation: https://checkstyle.sourceforge.io/checks/whitespace/whitespacearound.html

```cmd
C:\DEV\checkstyle>javac --version
javac 21.0.1

C:\DEV\checkstyle>javac Example.java

C:\DEV\checkstyle>type config.xml
<module name="Checker">
    <module name="SuppressWarningsFilter"/>

    <module name="TreeWalker">
        <module name="WhitespaceAround"/>
    </module>
</module>

C:\DEV\checkstyle>type Example.java
package com.company.example;

import java.util.Optional;

public class Example {
        Optional<String> someMethod(int arg) {
                return Optional.ofNullable(switch (arg) {
                        case 123 -> "yes";
                        default -> null;
                });
        }
}

C:\DEV\checkstyle>set RUN_LOCALE="-Duser.language=en -Duser.country=US"

C:\DEV\checkstyle>java %RUN_LOCALE% -jar checkstyle-10.15.0-all.jar -c config.xml Example.java
Starting audit...
[WARN] C:\DEV\checkstyle\Example.java:7:44: 'switch' is not preceded with whitespace. [WhitespaceAround]
Audit done.
```

---

After the introduction of switch expressions (e.g., `return switch (...)`) the `switch` keyword can now be used anywhere where an expression could be used previously.

As whitespace is not required before any other expression, it should also not be required before `switch` (when it is used as part of a larger expression). Therefore, I'd expect Checkstyle to allow this:
```
return Optional.ofNullable(switch (arg) {
        case 123 -> "yes";
        default -> null;
});
```

Now it only allows this:
```
return Optional.ofNullable(
	switch (arg) {                     // <-- with an extra line break
		case 123 -> "yes";
		default -> null;
	}
);
```
or this
```
return Optional.ofNullable( switch (arg) {   // <-- notice the space before `switch`
	case 123 -> "yes";
	default -> null;
});
```

I think this may have been a case missed in: https://github.com/checkstyle/checkstyle/issues/8687

strkkk:
Closes #14825

Diff Regression config: https://gist.githubusercontent.com/strkkk/26466254f89c32d8f1a6bd53b6f5b251/raw/3dfd508a98b976b267532ea4d4c0986d7b78c11c/config_single.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundCheck.java
@@ -71,6 +71,10 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
  * of the catch block (left curly bracket) is not separated from the end
  * of the catch block (right curly bracket).
  * </p>
+ * <p>
+ * Note: <a href="https://openjdk.org/jeps/361">
+ * Switch expressions</a> are ignored by this check.
+ * </p>
  * <ul>
  * <li>
  * Property {@code allowEmptyCatches} - Allow empty catch bodies.
@@ -497,22 +501,31 @@ public class WhitespaceAroundCheck extends AbstractCheck {
      */
     private boolean isNotRelevantSituation(DetailAST ast, int currentType) {
         final int parentType = ast.getParent().getType();
-        final boolean starImport = currentType == TokenTypes.STAR
-                && parentType == TokenTypes.DOT;
-        final boolean insideCaseGroup = parentType == TokenTypes.CASE_GROUP;
-
-        final boolean starImportOrSlistInsideCaseGroup = starImport || insideCaseGroup;
-        final boolean colonOfCaseOrDefaultOrForEach =
-                isColonOfCaseOrDefault(parentType)
-                        || isColonOfForEach(parentType);
-        final boolean emptyBlockOrType =
-                isEmptyBlock(ast, parentType)
+        final boolean result;
+        switch (parentType) {
+            case TokenTypes.DOT:
+                result = currentType == TokenTypes.STAR;
+                break;
+            case TokenTypes.LITERAL_DEFAULT:
+            case TokenTypes.LITERAL_CASE:
+            case TokenTypes.CASE_GROUP:
+                result = true;
+                break;
+            case TokenTypes.FOR_EACH_CLAUSE:
+                result = ignoreEnhancedForColon;
+                break;
+            case TokenTypes.EXPR:
+                result = currentType == TokenTypes.LITERAL_SWITCH;
+                break;
+            case TokenTypes.ARRAY_INIT:
+            case TokenTypes.ANNOTATION_ARRAY_INIT:
+                result = currentType == TokenTypes.RCURLY;
+                break;
+            default:
+                result = isEmptyBlock(ast, parentType)
                     || allowEmptyTypes && isEmptyType(ast);
-
-        return starImportOrSlistInsideCaseGroup
-                || colonOfCaseOrDefaultOrForEach
-                || emptyBlockOrType
-                || isArrayInitialization(currentType, parentType);
+        }
+        return result;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundCheck.java
```diff
@@ -497,22 +497,37 @@ public void visitToken(DetailAST ast) {
...
+                result = isEmptyBlock(ast, parentType)
                     || allowEmptyTypes && isEmptyType(ast);
-
-        return starImportOrSlistInsideCaseGroup
-                || colonOfCaseOrDefaultOrForEach
-                || emptyBlockOrType
-                || isArrayInitialization(currentType, parentType);
+            }
+        }
+        return result;
```
strkkk, line 530: I had to refactor this method since complexity was >10 and caused violation

nrmancuso:
Hi @strkkk :) please generate a regression report when you get a chance, you can assign me to this PR when you are ready.

strkkk:
Github, generate report

romani:
No violations on this by this fix: `switch' is not followed by whitespace.
`var otherSuffix = switch(locale.getLanguage()) {`

`return switch(exp) {`

Requested in issue:
`return Optional.ofNullable(switch (arg) {`

Something is weird, I think we mixing switch as statement and like expression. In statement, spaces around might make sense, in expression form we should skip it completely for preceded and following spaces. So there should be two violations disappear https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/e6cdb69_2024152818/reports/diff/openjdk17/index.html#A4 .

If we agree on this we should just mention in doc that we don't handle switch expressions at all.

Thought?

strkkk:
@romani it makes sense imo. There is WhitespaceAfter check if user wants to force whitespace after switch.

nrmancuso:
>If we agree on this we should just mention in doc that we don't handle switch expressions at all.

Yes, I am in agreement, @strkkk let's update the docs to reflect this.

strkkk:
ok, done. I added a note

nrmancuso (approved):

romani:
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/25506/workflows/53e74f0a-dbb1-4cfc-9309-183ef8929151/jobs/589623/artifacts is red

strkkk:
fixed

romani:
3 more CI failures

strkkk:
@romani build is green

romani (changes requested):
We need a bit more test inputs to get all kind of `ok` code that previously was violated.

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/e6cdb69_2024152818/reports/diff/openjdk17/index.html#A1

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/whitespace/whitespacearound/InputWhitespaceAroundSwitchExpressions.java
```diff
@@ -0,0 +1,38 @@
...
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.whitespace.whitespacearound;
+
+import java.util.Optional;
+
+public class InputWhitespaceAroundSwitchExpressions {
+
+    Optional<String> someMethod(int arg) {
+        return Optional.ofNullable(switch (arg) { // ok - switch expression
```
romani, line 28: Please add one more code like this but without space after switch

romani (changes requested):
Items

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/whitespace/whitespacearound/InputWhitespaceAroundSwitchExpressions.java
romani: `{ // ok - switch expression is skipped `

strkkk: done

romani (approved):
Ok to merge

rnveach (approved):

romani:
@strkkk , if will be awesome if you fix your account in CircleCI, this CI is not running for your account. This created a leak.

PR #14863, opened 2024-05-06
infra: change reason of spotbugs disablement in circleci

romani:
enablement based on request at https://github.com/spotbugs/spotbugs/issues/2819#issuecomment-2095294403

UPDATE: disablement is staying, reason is changing.


romani:
It not clear why we have such violations, over CIs are not reporting them. If them are false positives, we need to report them  to spot bugs team

romani:
I have the same 23 problems on my local.

romani:
we will continue to keep disablement until new issue is resolved.
most(maybe all) violations are false positives.

nrmancuso (approved):

Issue #14724, opened 2024-03-27
[JavadocType] add additional allowedTags feature


xenoterracide:
Since, java 8 there are new javadoc tags, but you can also add arbitrary ones

https://nipafx.dev/javadoc-tags-apiNote-implSpec-implNote/

**How it works Now:**

```
/**
 * Interface for objects that have a name.
 *
 * @apiNote you must implement one of {@link #getName()} or {@link #name()}.
 */
interface Nameable {}
```

```
[ant:checkstyle] [ERROR] /home/xeno/IdeaProjects/spring-app-commons/module/model/src/main/java/com/xenoterracide/model/Nameable.java:11:4: Unknown tag 'apiNote'. [JavadocType]
```

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

Although the tags mentioned above are used by java itself, they aren't enabled by default, and you have to pass options to javadoc to use them. note: the weird quoting mentioned in the doc appears to be a shell escape as it's completely irrelevant in gradle

```kts
tasks.withType<Javadoc>().configureEach {
  (options as StandardJavadocDocletOptions).apply {
    addMultilineStringsOption("tag").value = listOf(
      "apiSpec:a:API Spec:",
      "apiNote:a:API Note:",
      "implSpec:a:Implementation Spec:",
      "implNote:a:Implementation Note:",
    )
  }
}
```

I think the best option is to introduce a property of "additional tags"

```xml
      <property name="additionalTags" value="apiNote,apiSpec,implNote,implSpec,arbitraryInput" />
```

**Additional context**

it might be reasonable to consider adding `apiNote,apiSpec,implNote,implSpec` as either defaults for this new property, or just have them be recognized without that. I'm indifferent. I just don't like the idea that by allowing unknown tags, it means anything. Meaning typos won't be caught, or other junk.

sktpy:
In the meantime you can suppress violation via [SuppressionSingleFilter](https://checkstyle.org/filters/suppressionsinglefilter.html#SuppressionSingleFilter),
Refer https://github.com/checkstyle/checkstyle/issues/14618#issuecomment-1987370899 for more info.

It is highly likely that property would not be added until the check is redesigned to be AST based.
<br>

> I just don't like the idea that by allowing unknown tags, it means anything.

I agree with this. https://github.com/checkstyle/checkstyle/issues/14618#issuecomment-1986874359

romani:
[...]

PR #14882, opened 2024-05-17
Issue #14573: fix javadoc type check for empty tags

### LINKED ISSUE #14573: JavadocType: False negative for unknown tag with no description
([Documentation](https://checkstyle.org/checks/javadoc/javadoctype.html))

**config.xml**
```xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <module name="TreeWalker">
    <module name="JavadocType"/>
  </module>
</module>
```
<br>

**Test.java**
```java
/**
 * @val
 */
class Test {}

```
<br>

**Current CLI:**
```
java -jar checkstyle-10.14.0-all.jar -c config.xml Test.java
Starting audit...
Audit done.
```
<br>

**Expected:**
```
[ERROR] Test.java:2:4: Unknown tag 'val'. [JavadocType]
```
<hr>

**Explanation:**
- `val` is an unknown tag (allowUnknownTags by default is false), and hence should produce a violation.
- Note that this false negative **_only occurs when_** the tag is not accompanied by any description,  violation is produced as expected when the unknown tag is associated with description.

**Attention note from maintainers:**
https://github.com/checkstyle/checkstyle/blob/72cf172972605f79fa8e7ddeea4623529d30f730/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java#L149-L150
this Check is not AST based, so fix might be not trivial. We can accept fix only if it is simple and does not cause more regressions or false positives. Ideally we should  reimplement this Check to be ast based, example: https://github.com/checkstyle/checkstyle/blob/72cf172972605f79fa8e7ddeea4623529d30f730/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AtclauseOrderCheck.java#L110

strkkk:
Closes #14573

Diff Regression config: https://gist.githubusercontent.com/strkkk/26466254f89c32d8f1a6bd53b6f5b251/raw/cf451b3aa6366addc027a08423786d455bc79aa7/config_single.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
@@ -379,10 +379,9 @@ public class JavadocTypeCheck
         final JavadocTags tags = JavadocUtil.getJavadocTags(textBlock,
             JavadocUtil.JavadocTagType.BLOCK);
         if (!allowUnknownTags) {
-            for (final InvalidJavadocTag tag : tags.getInvalidTags()) {
-                log(tag.getLine(), tag.getCol(), MSG_UNKNOWN_TAG,
-                    tag.getName());
-            }
+            tags.getInvalidTags().forEach(tag -> {
+                log(tag.getLine(), tag.getCol(), MSG_UNKNOWN_TAG, tag.getName());
+            });
         }
         return tags.getValidTags();
     }
@@ -481,11 +480,16 @@ public class JavadocTypeCheck
                         || recordComponentNames.contains(paramName);
 
                 if (!found) {
-                    final String actualParamName =
-                        TYPE_NAME_IN_JAVADOC_TAG_SPLITTER.split(tag.getFirstArg())[0];
-                    log(tag.getLineNo(), tag.getColumnNo(),
-                        MSG_UNUSED_TAG,
-                        JavadocTagInfo.PARAM.getText(), actualParamName);
+                    if (paramName.isEmpty()) {
+                        log(tag.getLineNo(), tag.getColumnNo(), MSG_UNUSED_TAG_GENERAL);
+                    }
+                    else {
+                        final String actualParamName =
+                            TYPE_NAME_IN_JAVADOC_TAG_SPLITTER.split(tag.getFirstArg())[0];
+                        log(tag.getLineNo(), tag.getColumnNo(),
+                            MSG_UNUSED_TAG,
+                            JavadocTagInfo.PARAM.getText(), actualParamName);
+                    }
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


strkkk:
Github, generate report

strkkk:
There are cases like this https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/e5f329e_2024235419/reports/diff/spotbugs/index.html#A1, when annotation is considered as tag. In theory, we can analyse context and ignore text inside `<code>` or `<pre>` tags

strkkk:
Problem is in BlockTagUtil#extractBlockTags, it does not take context into consideration. There are 2 possible solutions:
1. Consider tags <pre> and <code> and ignore its content
2. Adjust regex to make it parse tag only if it starts with lowercase letter. I couldn't find some article / spec to back it up, but it seems reasonable

romani:
This Check is not AST based, so amount of problems in this Check is uncountable and we will never fix them. We need to migrate to AST.

Without migration, we just need to make sure that diff report is decently ok or better then before in behavior

strkkk:
> Without migration, we just need to make sure that diff report is decently ok or better then before in behavior

it seems ok to me if we ignore cases with code snippets with annotations in javadocs

romani:
> [openjdk17](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/e5f329e_2024235419/reports/diff/openjdk17/index.html) ( ±1954, +1954 )

this is too much, we can NOT implSpec, apiNote to known tags  in this Check in config.
Please add filter to you config_single.xml to skip such violations, and get cleaner report.
**We will need to post this filter in migration notes**, as such tags become more and more popular, we already have issue somewhere to add them to support.

> Consider tags and  and ignore its content

How complicated to add this ? User do not heavily like false positives, and "ok" with false-negatives, especially in javadoc.

romani:
This is where extra property is requested https://github.com/checkstyle/checkstyle/issues/14724#issuecomment-2021940770 it will ease a problem, but it is separate issue.

romani:
@strkkk , do you have time to finish this PR ?

romani:
Github, generate report for JavadocType/Example1

romani:
failed due to `Error:  Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.3:site (default-site) on project sample: Error during page generation: Error rendering Maven report: Failed during checkstyle configuration: cannot initialize module TreeWalker - Property 'skipFileOnJavaParseException' does not exist, please check the documentation -> [Help 1]`

rebase is required, PR is too old.

romani:
Github, generate report for JavadocType/Example1

romani:
Github, generate report for JavadocType/Example1

romani:

romani:

romani:

romani:
@MohanadKh03,

Please help us to finish this PR.
This comment is where we lost connection with author https://github.com/checkstyle/checkstyle/pull/14882#issuecomment-2143650370

In addition to Check changes there will be dive to our automated diff report generation that helps a lot to reduce leaks of bad behavior.

MohanadKh03:
Sure ill need a few days before starting this though
Is the check going to be rewritten to be AST-based or ?

romani:
@MohanadKh03 , thanks a lot !! sure any time
It is pleasure to work with you.

better to avoid conversion to AST, as it will be huge effort.
If we can fix this Check to work better and without false positives, false negatives are fine. It will be huge win. 

according to diff report that I see above just fix is archive-able.

romani:
Github, generate report for JavadocType/Example7

romani:
Github, generate report for JavadocType/Example7

romani (approved):
report above is good.

confusion was on what config to test.

ok to merge

PR #14884, opened 2024-05-19
spelling: added detailed message to spelling test script

SteLeo1602:
Added additional details to the message to clarify for the beginners what they need to do exactly and ensure they won't confuse the command with the rest of the report.

### DIFF
--- a/.ci/test-spelling-unknown-words.sh
+++ b/.ci/test-spelling-unknown-words.sh
@@ -55,7 +55,8 @@ echo "Run w"
 
 printDetails() {
   echo ''
-  echo 'If you are ok with the output of this run, you will need to'
+  echo 'If you are ok with the output of this run,'
+  echo 'you will need to run the following entire multiline command:'
 }
 
 echo "Review results"


romani:
>[ERROR] [checkstyle] [ERROR] C:\projects\checkstyle\.ci\test-spelling-unknown-words.sh:58: Line should not be longer than 100 symbols [lineLength]

Please fix CI failure

.ci/test-spelling-unknown-words.sh
```diff
@@ -55,7 +55,8 @@ echo "Run w"
 
 printDetails() {
   echo ''
-  echo 'If you are ok with the output of this run, you will need to'
+  echo 'If you are ok with the output of this run, you will need \
+to run the following entire multiline command:'
```
romani, line 59: I would do this in two separate echo commands, no reason on line wrap here.

romani: `echo 'If you are ok with the output of this run,'`
Other text by separate echo command.

SteLeo1602: Done

Issue #14910, opened 2024-05-30
Add new section 4.8.5.2 Class Annotations to google style guide implementation


Zopsss:
I have read check documentation: https://checkstyle.org/checks/annotation/annotationlocation.html
I have downloaded the latest cli from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

**How it works Now:**
![Screenshot 2024-05-30 181511](https://github.com/checkstyle/checkstyle/assets/86093112/6a4ab013-8bf8-42b8-9201-fdfab33ec453)
( Diff Link: https://github.com/checkstyle/checkstyle/pull/14901/files/4bcfed8ef4a4446fbda32e24fd067abd933030ce#diff-3f79fa1ed75e6db042f0bb87505b89c47280f86dc510d90ed8d039c77869ce33R764-R775 )

There was a new section added in the latest version of google style guide implementation, section `4.8.5.2`. This section was separated from the original section `4.8.5`, dividing it into sub-sections.

This rule has already being followed in the current implementation, for example:

#### ClassAnnotations.java
```java
import java.beans.JavaBean;

/**
 * testing.
 */
@Deprecated
@JavaBean
public class ClassAnnotations {
  // testing class.


  /**
   * Testing.
   */
  @Deprecated @JavaBean // warning, ensuring there's only one annotation per-line.
  class Test {

  }

  @Deprecated
  @JavaBean
  /** // warning, ensuring the annotations comes immediately after the documentation.
   * Testing.
   */
  class Test2 {

  }
}
```

#### Output:
```
java -jar .\checkstyle-10.17.0-all.jar -c .\google_checks.xml .\ClassAnnotations.java
Starting audit...
[WARN] C:\checkstyle testing\TestingGoogleStyle\.\ClassAnnotations.java:15:15: Annotation 'JavaBean' should be alone on line. [AnnotationLocationMostCases]
[WARN] C:\checkstyle testing\TestingGoogleStyle\.\ClassAnnotations.java:22:3: Javadoc comment is placed in the wrong location. [InvalidJavadocPosition]
Audit done.
```

These 2 warnings ensures that there is only one annotation per-line and annotation comes immediately after the documentation, not before the documentation.

**Describe the solution you'd like**
[...]

Issue #265, opened 2014-09-02
New Check: Utf8EncodingCheck


rdiachenko:
Source files have to be UTF-8 encoded: http://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s2.2-file-encoding

PR #14925, opened 2024-06-01
Issue #14910: Add new section 4.8.5.2 in google style guide implementation

Zopsss:
Issue #14910 

~~Draft PR~~ because I had some doubts regarding failing xdoc tests.

I was adding a new section `4.8.5.2` to the google style guide implementation, to add support for this section, I required to combine 2 different modules ( AnnotationLocation & InvalidJavadocPosition ) in single test case. 

The test case is working fine but the problem is with Xdoc tests. I added 2 new configs and 1 test case in the coverage page, something like this:

Screenshot 2024-06-02 012656

Here, 2 `config` are for AnnotationLocation module and InvalidJavadocPosition module. The single `test` is the test file which uses these 2 configs internally to ensure we're following the rule.

But the `XdocPagesTest` is failing, it requires all the `config` to be paired with their individual `test` which is not the case here. In this case, it requires to have one `test` for AnnotationLocation and other `test` for InvalidJavadocPosition. 

And also it requires to have the test file linked in the `test` to be same name as its module name. In this case, for AnnotationLocation's `test`, it requires to be linked to a `AnnotationLocationTest` file and for InvalidJavadocPosition's `test`, it requires to be linked to a `InvalidJavadocPositionTest` file. In short, we cannot name the test file differently from its module name.

#### What we need for the new Xdoc model:

- Remove the pairing of `config` and `test` and allow us to have single `test` for multiple `config`. 
- The name of the test file used in the `test` does not have to be same as its module name.

### DIFF
--- a/config/sevntu-suppressions.xml
+++ b/config/sevntu-suppressions.xml
@@ -39,6 +39,7 @@
 
   <!-- Suppressed alternative method of pulling partial configuration from file -->
   <suppress checks="CheckstyleTestMakeup" files="[\\/]RightCurlyTest\.java"/>
+  <suppress checks="CheckstyleTestMakeup" files="[\\/]ClassAnnotationsTest\.java"/>
 
   <!-- No need for constructor with cause -->
   <suppress checks="CauseParameterInException" files=".*MetadataGenerationException\.java"/>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
@@ -226,6 +226,62 @@ public class XdocsPagesTest {
     private static final Set<String> GOOGLE_MODULES = Collections.unmodifiableSet(
         CheckUtil.getConfigGoogleStyleModules());
 
+    // Contains all the sections which are not migrated to chapter wise testing
+    // in google style documentation.
+    // This list will be removed once all the sections are migrated.
+    // until https://github.com/checkstyle/checkstyle/issues/14937
+    private static final Set<String> PER_MODULE_TESTS_RULES_LIST = Set.of(
+            "2.1 File name",
+            "2.3.1 Whitespace characters",
+            "2.3.2 Special escape sequences",
+            "2.3.3 Non-ASCII characters",
+            "3 Source file structure",
+            "3.2 Package statement",
+            "3.3.1 No wildcard imports",
+            "3.3.2 No line-wrapping",
+            "3.3.3 Ordering and spacing",
+            "3.4.1 Exactly one top-level class declaration",
+            "3.4.2.1 Overloads: never split",
+            "4.1.1 Use of optional braces",
+            "4.1.2 Nonempty blocks: K & R style",
+            "4.1.3 Empty blocks: may be concise",
+            "4.2 Block indentation: +2 spaces",
+            "4.3 One statement per line",
+            "4.4 Column limit: 100",
+            "4.5.1 Where to break",
+            "4.5.2 Indent continuation lines at least +4 spaces",
+            "4.6.1 Vertical Whitespace",
+            "4.6.2 Horizontal whitespace",
+            "4.8.2.1 One variable per declaration",
+            "4.8.2.2 Declared when needed",
+            "4.8.3.2 No C-style array declarations",
+            "4.8.4.1 Indentation",
+            "4.8.4.2 Fall-through: commented",
+            "4.8.4.3 Presence of the default label",
+            "4.8.5 Annotations",
+            "4.8.6.1 Block comment style",
+            "4.8.7 Modifiers",
+            "4.8.8 Numeric Literals",
+            "5.2.1 Package names",
+            "5.2.2 Class names",
+            "5.2.3 Method names",
+            "5.2.5 Non-constant field names",
+            "5.2.6 Parameter names",
+            "5.2.7 Local variable names",
+            "5.2.8 Type variable names",
+            "5.3 Camel case: defined",
+            "6.2 Caught exceptions: not ignored",
+            "6.4 Finalizers: not used",
+            "7.1.1 General form",
+            "7.1.2 Paragraphs",
+            "7.1.3 Block tags",
+            "7.2 The summary fragment",
+            "7.3 Where Javadoc is used",
+            "7.3.1 Exception: self-explanatory methods",
+            "7.3.2 Exception: overrides",
+            "7.3.4 Non-required Javadoc"
+    );
+
     /**
      * Generate xdoc content from templates before validation.
      * This method will be removed once
@@ -1773,16 +1829,30 @@ public class XdocsPagesTest {
             Set<String> styleChecks, String styleName, String ruleName) {
         final Iterator<Node> itrChecks = checks.iterator();
         final Iterator<Node> itrConfigs = configs.iterator();
+        final boolean isRegularDocumentation = "sun".equals(styleName)
+               || "google".equals(styleName) && PER_MODULE_TESTS_RULES_LIST.contains(ruleName);
 
+        if (isRegularDocumentation) {
+            validateModuleWiseTesting(itrChecks, itrConfigs, styleChecks, styleName, ruleName);
+        }
+        else {
+            validateChapterWiseTesting(itrChecks, itrConfigs, styleChecks, styleName, ruleName);
+        }
+
+        assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' has too many configs")
+                .that(itrConfigs.hasNext())
+                .isFalse();
+    }
+
+    private static void validateModuleWiseTesting(Iterator<Node> itrChecks,
+          Iterator<Node> itrConfigs, Set<String> styleChecks, String styleName, String ruleName) {
         while (itrChecks.hasNext()) {
             final Node module = itrChecks.next();
             final String moduleName = module.getTextContent().trim();
             final String href = module.getAttributes().getNamedItem("href").getTextContent();
-            // until https://github.com/checkstyle/checkstyle/issues/13132
-            final boolean moduleIsConfig = href.startsWith("config_");
             final boolean moduleIsCheck = href.startsWith("checks/");
 
-            if (!moduleIsConfig && !moduleIsCheck) {
+            if (!moduleIsCheck) {
                 continue;
             }
 
@@ -1842,10 +1912,91 @@ public class XdocsPagesTest {
                 }
             }
         }
+    }
 
-        assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' has too many configs")
-                .that(itrConfigs.hasNext())
+    private static void validateChapterWiseTesting(Iterator<Node> itrChecks,
+          Iterator<Node> itrConfigs, Set<String> styleChecks, String styleName, String ruleName) {
+        boolean hasChecks = false;
+
+        while (itrChecks.hasNext()) {
+            final Node module = itrChecks.next();
+            final String moduleName = module.getTextContent().trim();
+            final String href = module.getAttributes().getNamedItem("href").getTextContent();
+            final boolean moduleIsCheck = href.startsWith("checks/");
+
+            if (!moduleIsCheck) {
+                continue;
+            }
+
+            hasChecks = true;
+
+            assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' module '"
+                    + moduleName + "' shouldn't end with 'Check'")
+                    .that(moduleName.endsWith("Check"))
+                    .isFalse();
+
+            styleChecks.remove(moduleName);
+
+            Node config = null;
+
+            try {
+                config = itrConfigs.next();
+            }
+            catch (NoSuchElementException ignore) {
+                assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' module '"
+                        + moduleName + "' is missing the config link: config").fail();
+            }
+
+            final String configUrl = config.getAttributes().getNamedItem("href")
+                    .getTextContent();
+
+            final String expectedUrl = "https://github.com/search?q="
+                    + "path%3Asrc%2Fmain%2Fresources%20path%3A**%2F" + styleName
+                    + "_checks.xml+repo%3Acheckstyle%2Fcheckstyle+" + moduleName;
+
+            assertWithMessage("google_style.xml rule '" + ruleName + "' module '"
+                    + moduleName + "' should have matching config url")
+                    .that(configUrl)
+                    .isEqualTo(expectedUrl);
+
+        }
+
+        if (itrConfigs.hasNext()) {
+            assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' should have checks"
+                    + " if it has config/test links")
+                    .that(hasChecks)
+                    .isTrue();
+
+            final Node config = itrConfigs.next();
+            final String configUrl = config.getAttributes().getNamedItem("href")
+                    .getTextContent();
+            final String[] parts = ruleName.split(" ", 2);
+            final String extractedRuleName = parts[1].trim().replaceAll(" ", "");
+
+            assertWithMessage("google_style.xml rule '" + ruleName + "' rule '"
+                    + "' should have matching test url")
+                    .that(configUrl)
+                    .startsWith("https://github.com/checkstyle/checkstyle/"
+                            + "blob/master/src/it/java/com/google"
+                            + "/checkstyle/test/");
+
+            assertWithMessage("google_style.xml rule '" + ruleName
+                    + "' should have matching test url")
+                    .that(configUrl)
+                    .endsWith("/" + extractedRuleName + "Test.java");
+
+            assertWithMessage("google_style.xml rule '" + ruleName
+                    + "' should have a test that exists")
+                    .that(new File(configUrl.substring(53).replace('/',
+                            File.separatorChar)).exists())
+                    .isTrue();
+        }
+        else {
+            assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' should have no"
+                 + " checks if it has no config/test links")
+                .that(hasChecks)
                 .isFalse();
+        }
     }
 
     @Test
--- a/config/jsoref-spellchecker/whitelist.words
+++ b/config/jsoref-spellchecker/whitelist.words
@@ -197,6 +197,7 @@ cimg
 Cinit
 circleci
 cirrusci
+classannotations
 classdataabstractioncoupling
 classfanoutcomplexity
 classloader
--- a/src/it/java/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/ClassAnnotationsTest.java
+++ b/src/it/java/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/ClassAnnotationsTest.java
@@ -1,0 +1,68 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code and other text files for adherence to a set of rules.
+// Copyright (C) 2001-2024 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package com.google.checkstyle.test.chapter4formatting.rule4852classannotations;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+import com.google.checkstyle.test.base.AbstractGoogleModuleTestSupport;
+import com.puppycrawl.tools.checkstyle.api.Configuration;
+import com.puppycrawl.tools.checkstyle.checks.annotation.AnnotationLocationCheck;
+import com.puppycrawl.tools.checkstyle.checks.javadoc.InvalidJavadocPositionCheck;
+
+public class ClassAnnotationsTest extends AbstractGoogleModuleTestSupport {
+
+    @Override
+    protected String getPackageLocation() {
+        return "com/google/checkstyle/test/chapter4formatting/rule4852classannotations";
+    }
+
+    @Test
+    public void testAnnotation() throws Exception {
+        final Class<AnnotationLocationCheck> clazz1 = AnnotationLocationCheck.class;
+        final Class<InvalidJavadocPositionCheck> clazz2 = InvalidJavadocPositionCheck.class;
+
+        final Configuration annotationConfig = getModuleConfig("AnnotationLocation",
+                "AnnotationLocationMostCases");
+        final Configuration invalidJavadocPositionConfig =
+                getModuleConfig("InvalidJavadocPosition");
+        final List<Configuration> childrenConfigs = new ArrayList<>();
+        childrenConfigs.add(annotationConfig);
+        childrenConfigs.add(invalidJavadocPositionConfig);
+        final Configuration treeWalkerConfig = createTreeWalkerConfig(childrenConfigs);
+
+        final String annotationAloneMsg = "annotation.location.alone";
+        final String invalidJavadocPositionMsg = "invalid.position";
+
+        final String[] expected = {
+            "17:5: " + getCheckMessage(clazz2, invalidJavadocPositionMsg),
+            "25:22: " + getCheckMessage(clazz1, annotationAloneMsg, "SomeAnnotation2"),
+            "28:22: " + getCheckMessage(clazz1, annotationAloneMsg, "SomeAnnotation2"),
+            "29:5: " + getCheckMessage(clazz2, invalidJavadocPositionMsg),
+        };
+
+        final String filePath = getPath("InputClassAnnotations.java");
+
+        final Integer[] warnList = getLinesWithWarn(filePath);
+        verify(treeWalkerConfig, filePath, expected, warnList);
+    }
+}
--- a/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/InputClassAnnotations.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/InputClassAnnotations.java
@@ -1,0 +1,33 @@
+package com.google.checkstyle.test.chapter4formatting.rule4852classannotations;
+
+public class InputClassAnnotations {
+    public @interface SomeAnnotation1 { }
+
+    public @interface SomeAnnotation2 { }
+
+    /**
+     * Inner class 2.
+     */
+    @SomeAnnotation1
+    @SomeAnnotation2
+    class Inner1 {}
+
+    @SomeAnnotation1
+    @SomeAnnotation2
+    /** // warn
+     * Inner class 2.
+     */
+    class Inner2 {}
+
+    /**
+     * Inner class 3.
+     */
+    @SomeAnnotation1 @SomeAnnotation2 // warn
+    class Inner3 {}
+
+    @SomeAnnotation1 @SomeAnnotation2 // warn
+    /** // warn
+     * Inner class 4.
+     */
+    class Inner4 {}
+}
--- a/src/xdocs/google_style.xml
+++ b/src/xdocs/google_style.xml
@@ -1365,8 +1365,23 @@
                   </a>
                 </td>
                 <td>
-                  <a href="styleguides/google-java-style-20180523/javaguide.html#s4.8.5-annotations">
+                  <a href="styleguides/google-java-style-20220203/javaguide.html#s4.8.5-annotations">
                     4.8.5 Annotations</a>
+                </td>
+                <td>↓</td>
+                <td/>
+              </tr>
+              <tr>
+                <td><a name="4.8.5.2"/>
+                  <a href="#4.8.5.2">
+                    <span class="wrapper inline">
+                      <img src="images/anchor.png" alt=""/>
+                    </span>
+                  </a>
+                </td>
+                <td>
+                  <a href="styleguides/google-java-style-20220203/javaguide.html#s4.8.5.2-class-annotation-style">
+                    4.8.5.2 Class Annotations</a>
                 </td>
                 <td>
                   <span class="wrapper inline">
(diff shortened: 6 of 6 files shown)


romani:
```
[ERROR] Failures: 
[ERROR]   XdocsPagesTest.testAllStyleRules:1651->validateStyleModules:1810 
google_style.xml rule '4.8.5.2 Class Annotations' module 'AnnotationLocation' 
has mismatched config/test links
expected: test
but was : config

```

romani:
yes, you just need to update xdoc validation XdocsPagesTest.testAllStyleRules . 
section `4.8.5.2 Class Annotations` should be `ClassAnnotationsTest` or as before each Check has special test (to keep compatibillity with old style).
So is test is notmatching a Check, we check that xxxxTest  is matching title of section, without digits and special symbols (`-`, `:`) and it should be camel case.
please update this test in this PR, do not hesitate to change code of this test.

Zopsss:
How can I solve this error:

https://ci.appveyor.com/project/Checkstyle/checkstyle/builds/49945078/job/rruisaqwb5xrrwyq#L311

It is for following line:

https://github.com/checkstyle/checkstyle/blob/33f16c552746ef1a6af9cbcc6c94826df776acdc/src/it/java/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/ClassAnnotationsTest.java#L51

I tried using `DefaultConfiguration` and `Configuration` both as the variable type but it is still giving the same error

rdiachenko:
> How can I solve this error:
> 
> https://ci.appveyor.com/project/Checkstyle/checkstyle/builds/49945078/job/rruisaqwb5xrrwyq#L311
> 
> It is for following line:
> 
> https://github.com/checkstyle/checkstyle/blob/33f16c552746ef1a6af9cbcc6c94826df776acdc/src/it/java/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/ClassAnnotationsTest.java#L51
> 
> I tried using `DefaultConfiguration` and `Configuration` both as the variable type but it is still giving the same error

It is related to a method used to create config (not a type): `createTreeWalkerConfig(...)`. More details in [this javadoc](https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/CheckstyleTestMakeupCheck.java#L52).

I see the same creation method used in `RightCurlyTest`, and I also see it is suppressed in `sevntu-suppressions.xml` config:
```
  <!-- Suppressed alternative method of pulling partial configuration from file -->
  <suppress checks="CheckstyleTestMakeup" files="[\\/]RightCurlyTest\.java"/>
```

You should probably extend that suppression for your test.

Zopsss:
> You should probably extend that suppression for your test.

thanks, now it is working fine. 

CircleCI CIs are failing:

https://app.circleci.com/pipelines/github/checkstyle/checkstyle/25711/workflows/82fbba72-a93e-4413-96fd-257f97beca74/jobs/598142

https://app.circleci.com/pipelines/github/checkstyle/checkstyle/25711/workflows/0b2da461-e9a2-4ba0-9cf5-0b7d146db772/jobs/598145

But I guess we do not need to worry about it. Maybe these are the reasons:

> CircleCI Security Alert:  We have become aware of a website impersonating CircleCI, presumably for malicious purposes. Stay alert and stay safe. [Read More](https://discuss.circleci.com/t/circleci-security-alert-warning-fraudulent-website-impersonating-circleci/50899)

> CircleCI ran into an issue setting up the infrastructure for this job. Try re-running the build and if this issue persists, open a [Support ticket](https://support.circleci.com/hc/en-us).

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -226,6 +226,61 @@ public class XdocsPagesTest {
     private static final Set<String> GOOGLE_MODULES = Collections.unmodifiableSet(
         CheckUtil.getConfigGoogleStyleModules());
 
+    // Contains all the sections which are not migrated to chapter wise testing
+    // in google style documentation.
+    // This list will be removed once all the sections are migrated.
```
romani, line 231: please put here `until #xxxxxx`
to make sure in scope of what ticket we will do final cleanup.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1773,6 +1828,93 @@ private static void validateStyleModules(Set<Node> checks, Set<Node> configs,
             Set<String> styleChecks, String styleName, String ruleName) {
         final Iterator<Node> itrChecks = checks.iterator();
         final Iterator<Node> itrConfigs = configs.iterator();
+        final boolean isRegularDocumentation = "google".equals(styleName)
+                && PER_MODULE_TESTS_RULES_LIST.contains(ruleName) || "sun".equals(styleName);
```
romani, line 1832:
```
final boolean isPerModuleTesting = "sun".equals(styleName) 
   ||  "google".equals(styleName) && PER_MODULE_TESTS_RULES_LIST.contains(ruleName) 
```

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1773,6 +1828,93 @@ private static void validateStyleModules(Set<Node> checks, Set<Node> configs,
...
+
+        if (isRegularDocumentation) {
+            while (itrChecks.hasNext()) {
```
romani, line 1835: please move to new method - validateModuleWiseTesting

romani (changes requested):
items:

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: Currently, we does not have any issue which talks about migrating to chapter wise testing, so which issue should I link here?

romani: lets create such issue and reference it here :)

Zopsss: I have opened the issue: https://github.com/checkstyle/checkstyle/issues/14937. I'll link this as you requested.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: Done.

https://github.com/checkstyle/checkstyle/blob/7e868f5cd0424fa926ed50a7426158666c3941a1/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1835-L1837

https://github.com/checkstyle/checkstyle/blob/7e868f5cd0424fa926ed50a7426158666c3941a1/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1847-L1848

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: done

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: Linked the issue: https://github.com/checkstyle/checkstyle/blob/7e868f5cd0424fa926ed50a7426158666c3941a1/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L229-L233

Zopsss:
@romani  I've made changes according to your suggestions. The XdocPagesTest now has support to check Chapter Wise Testing approach. I've tested it locally with different inputs and it was working fine.

 I guess the PR is now ready to be merged.

config/sevntu-suppressions.xml
```diff
@@ -39,6 +39,7 @@
 
   <!-- Suppressed alternative method of pulling partial configuration from file -->
   <suppress checks="CheckstyleTestMakeup" files="[\\/]RightCurlyTest\.java"/>
+  <suppress checks="CheckstyleTestMakeup" files="[\\/]ClassAnnotationsTest\.java"/>
```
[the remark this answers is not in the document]
romani: It was checkstyle violation.
`ERROR] [checkstyle] [ERROR] C:\projects\checkstyle\src\it\java\com\google\checkstyle\test\chapter4formatting\rule4852classannotations\ClassAnnotationsTest.java:51:46: DefaultConfiguration was not assigned by an allowed create configuration method. [CheckstyleTestMakeup]`

When I removed such suppression, `mvn verify` passed, so it means this suppression is not required.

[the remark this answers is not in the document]
romani: Let's try to remove it to let me see this anomaly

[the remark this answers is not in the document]
Zopsss: done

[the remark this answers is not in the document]
Zopsss: Failing CI: https://ci.appveyor.com/project/Checkstyle/checkstyle/builds/49952243/job/96b020g8euk1q1ld#L311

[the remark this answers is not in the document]
romani: It is weird why only windows is affected, Linux execution is green.
We will deal with this in separate issue.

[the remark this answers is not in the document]
romani: Please restore suppression.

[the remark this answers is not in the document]
Zopsss: done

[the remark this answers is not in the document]
rnveach: Violation looks correct. `createTreeWalkerConfig` is not an allowed method. You can see it is extremely rare for us to use it in our code base.

I mentioned before this method of picking and pulling configs just doesn't seem like the best approach. How will we validate the checks we manually pick are the same ones we have identified in the sections and that we didn't miss one or add one? The whole reason we had a 1 to 1 check/test is this was a way to validate we were testing against what we are saying.

Is there a reason we just don't apply the whole google config against the input file?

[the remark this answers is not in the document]
romani: https://github.com/checkstyle/checkstyle/issues/14939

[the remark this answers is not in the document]
rnveach: My comment is not on the inconsistent CI, but on the code itself.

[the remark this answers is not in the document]
romani: >Is there a reason we just don't apply the whole google config against the input file?

It might be too much violations, or too hard to make Input.
I remember we tried at initial year to make it and there was some mess. 
I am not sure on this.
It is easy to make 0 violations Inputs, and with violations might have bunch unrelated violations.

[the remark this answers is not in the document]
Zopsss: > How will we validate the checks we manually pick are the same ones we have identified in the sections and that we didn't miss one or add one? The whole reason we had a 1 to 1 check/test is this was a way to validate we were testing against what we are saying in the documentation.


You have a valid point. We cannot test if all the mentioned modules in a section are used in the test or if any module is missing in the test. This approach focuses on combining different modules into one test so we can test the whole section in one go. For this, I guess we need to manually ensure that all the modules are valid and no modules is left behind.

Zopsss:
GitHub, generate site

config/sevntu-suppressions.xml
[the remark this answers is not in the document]
rnveach: Manual is ok for a time, but I see this as being a concern. All reviewers must ensure every tests matches google doc each time an update is done.

I don't see how it is bad we have unrelated violations if we apply the whole google config against input files. The whole point of this is we are showing we follow google's guide, so any and all violations which are displayed should be part of what we are showcasing. If a violation is "wrong", then this would be a bug in our config and would need to be fixed.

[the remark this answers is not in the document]
Zopsss: >  My motto is if you don't enforce something by CI, someone will find a way to do something the wrong way. 

Why CI? We can write test in XdocPagesTest so why we should enforce something by CI?

> I don't see how it is bad we have unrelated violations if we apply the whole google config against input files. The whole point of this is we are showing we follow google's guide, so any and all violations which are displayed should be part of what we are showcasing. If a violation is "wrong", then this would be a bug in our config and would need to be fixed.

You're correct but checking the input against whole config will require us to take small things into account like adding javadocs, maintaining indentation ( google uses 2 spaces for indentation ), etc. This can become annoying to do as we need to ensure all these small details in every tests we write. Romani also said in the discord that, earlier we tried this approach but it was a mess or something like that. IMO we should properly discuss this before moving forward. I'm fine with abandoning the idea of validating the modules being used in the test. We can implement it after we come to a solid decision.

romani:
@Zopsss , please create issue to discuss and do xdoc pages validation for test presense.
in this PR do hardcodes list of sections where we simply skip validation of tests in table.
We will discuss all details on how to do this in separate Issue of this summer project.

config/sevntu-suppressions.xml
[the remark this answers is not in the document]
rnveach: > We can write test in XdocPagesTest 

Tests are run by CI, so if it is in a test, it is enforced. I wasn't implying writing in the CI's YML directly, sorry for that confusion.

> checking the input against whole config will require us to take small things into account like adding javadocs, maintaining indentation ( google uses 2 spaces for indentation ), etc. This can become annoying to do as we need to ensure all these small details in every tests we write. 

Yes, this is why I was suggesting this not be part of this GSoC. It does also seem odd to show these files off as following GSoC, when as you bring up, they do not follow them. :)

> I'm fine with abandoning the idea of validating the modules being used in the test. We can implement it after we come to a solid decision.

I agree.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1818,34 +1891,111 @@ private static void validateStyleModules(Set<Node> checks, Set<Node> configs,
...
+        while (itrChecks.hasNext()) {
+            final Node module = itrChecks.next();
+            final String moduleName = module.getTextContent().trim();
+            final String href = module.getAttributes().getNamedItem("href").getTextContent();
+            // until https://github.com/checkstyle/checkstyle/issues/13132
+            final boolean moduleIsConfig = href.startsWith("config_");
+            final boolean moduleIsCheck = href.startsWith("checks/");
+
+            if (!moduleIsConfig && !moduleIsCheck) {
+                skip = true;
```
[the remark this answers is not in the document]
rnveach: In master, this type of condition was hit on a link to an issue ( `https://github.com/checkstyle/checkstyle/issues/265` ).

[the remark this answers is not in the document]
rnveach: I am seeing separation a bit better now, skip will skip all test config links if it is set. This could lead to false results where we completely ignore column despite what is in check column.

If column should be empty, it should be validated as empty. If it shouldn't be empty, it should be validated as not empty.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1818,34 +1891,111 @@ private static void validateStyleModules(Set<Node> checks, Set<Node> configs,
...
+
+            assertWithMessage("google_style.xml rule '" + ruleName + "' module '"
+                    + moduleName + "' should have matching config url")
+                    .that(configUrl)
+                    .isEqualTo(expectedUrl);
+
+        }
+
+        if (!skip) {
+            if (itrConfigs.hasNext()) {
```
rnveach, line 1969: Explain chapter testing. My understanding was, 1 test for many checks. We would still have many configs for many checks.

I am just seeing checks for test links here, and no more config links?

rnveach:
GitHub, generate website

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
rnveach: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a4.8.5.2

rnveach: I am going to resolve this for now.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1818,34 +1891,111 @@ private static void validateStyleModules(Set<Node> checks, Set<Node> configs,
...
             }
         }
+    }
 
-        assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' has too many configs")
-                .that(itrConfigs.hasNext())
-                .isFalse();
+    private static void validateChapterWiseTesting(Iterator<Node> itrChecks,
+          Iterator<Node> itrConfigs, Set<String> styleChecks, String styleName, String ruleName) {
+        boolean skip = !itrConfigs.hasNext();
```
rnveach, line 1922: Rename skip to `hasChecks`. Set it to false for now.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
rnveach, line 1933: Remove this skip assignment here.

If we have checks, we must have atleast 1 config link and 1 test link. If we have no checks, we must have no configs or links.

After this if statement and before first assert set `hasChecks` to true. We are validating a check at this point, we so know we have a check. If we have no links, it will be caught during `itrConfigs.next`. If we have links but it is a test link, it will be caught during expected config link check.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1818,34 +1891,111 @@ private static void validateStyleModules(Set<Node> checks, Set<Node> configs,
...
+                    + "_checks.xml+repo%3Acheckstyle%2Fcheckstyle+" + moduleName;
+
+            assertWithMessage("google_style.xml rule '" + ruleName + "' module '"
+                    + moduleName + "' should have matching config url")
+                    .that(configUrl)
+                    .isEqualTo(expectedUrl);
+
+        }
+
+        if (!skip) {
```
rnveach, line 1968: Remove this line `if (!skip) {`.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1818,34 +1891,111 @@ private static void validateStyleModules(Set<Node> checks, Set<Node> configs,
...
+            if (itrConfigs.hasNext()) {
+                final Node config = itrConfigs.next();
```
rnveach, line 1970: Assert before accessing `itrConfigs` that `hasChecks` is `true`.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1818,34 +1891,111 @@ private static void validateStyleModules(Set<Node> checks, Set<Node> configs,
...
+
+                assertWithMessage("google_style.xml rule '" + ruleName
+                        + "' should have a test that exists")
+                        .that(new File(configUrl.substring(53).replace('/',
+                                File.separatorChar)).exists())
+                        .isTrue();
+            }
+            else {
+                assertWithMessage("google_style.xml rule '" + ruleName + "' should have a test url")
+                        .fail();
```
rnveach, line 1996: Change this to assert `hasChecks` is `false`.

rnveach (changes requested):
Some improvements. My main complaint is `if (!skip) {` didn't seem proper. It was possibly to basically not validate any test links if it was ever set improperly.

Zopsss:
> My main complaint is `if (!skip) {` didn't seem proper.

I guess you're misunderstanding it's purpose. Let me explain the whole flow of the method, that way it will be clear to you:

There is a list of sections of google style: https://github.com/checkstyle/checkstyle/blob/8db57f20d1729eeb519e3f52ef4a6701f66ab87f/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L233

this list contains all the google style sections which follows per-module approach. Though this list does not contains sections like: [2.2](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a2.2), [3.3.4](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a3.3.4), [5.2.4](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a5.2.4), [6.1](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a6.1), [6.3](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a6.3).

We will be removing sections one-by-one from this list as we migrate them to chapter wise testing approach, at the end this list will be empty and removed. I did not included the above mentioned sections as they does not comes under any of the approach ( nor in chapter-wise neither in per-module ) because at the end we want to remove this list, if we keep those sections under in this list then this list cannot be empty and removed. 

In the [validateStyleModules()](https://github.com/checkstyle/checkstyle/blob/8db57f20d1729eeb519e3f52ef4a6701f66ab87f/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1828) method we're checking if the current documentation is sun style? if yes then simply follow per-module testing approach, if the current documentation is google and the current section is not in the list I mentioned above then it means we need to validate chapter-wise approach. So all the sections I mentioned above ( 2.2, 3.3.4, etc ) which are not included in that list will also be counted in the chapter-wise approach. Here is how we're checking this condition:

https://github.com/checkstyle/checkstyle/blob/8db57f20d1729eeb519e3f52ef4a6701f66ab87f/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1832-L1840

Now let's come to the `skip` part:

https://github.com/checkstyle/checkstyle/blob/8db57f20d1729eeb519e3f52ef4a6701f66ab87f/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1922

this is checking if the current section does not have any check link ( like [3.3.4](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a3.3.4), [5.2.4](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a5.2.4), [6.1](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a6.1), [6.3](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a6.3) ) then we will be skipping validation for that section.

It will set `skip` to true if the current section does not have any check link. otherwise it'll set false.

---

https://github.com/checkstyle/checkstyle/blob/8db57f20d1729eeb519e3f52ef4a6701f66ab87f/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1932-L1934

this is checking if the current section has link but it is not valid ( like [2.2](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a2.2) ) then we will be skipping config link validation.

The above condition will be hit only for section 2.2.

---

https://github.com/checkstyle/checkstyle/blob/8db57f20d1729eeb519e3f52ef4a6701f66ab87f/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1968

this is used to check if we need to skip the validation of test links for current section or not. If `skip` is set to true then we need to skip the validation.

If the current section does not have any check link ( like [3.3.4](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a3.3.4), [5.2.4](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a5.2.4), [6.1](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a6.1), [6.3](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a6.3) ) or if it has but the link is not valid ( like [2.2](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a2.2) ) then it helps us to skip the test link validation for them.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: We cannot change this and other suggestions you gave. Pls refer to this comment detailed explanation: https://github.com/checkstyle/checkstyle/pull/14925

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: > https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a4.8.5.2

the `skip` isn't used for this, please refer to this comment https://github.com/checkstyle/checkstyle/pull/14925#issuecomment-2157832607

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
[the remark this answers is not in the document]
Zopsss: > In master, this type of condition was hit on a link to an issue (
> 
> https://github.com/checkstyle/checkstyle/blob/2c232e33c8edccfd7f7658917f7f38208f5a3f2e/src/xdocs/google_style.xml#L199
> 
> ).

yes it is used for this line. Pls refer https://github.com/checkstyle/checkstyle/pull/14925#issuecomment-2157832607

rnveach:
> The above condition will be hit only for section 2.2.

For your current code, yes, however, 2 things.

1) I am looking to the future and what looks appropriate in the validation. We should expect random links to appear in the column and to not impact the correct validation. This is an HTML website, so we can literally put things in new places we haven't before.
2) [When you expand this to the (old) section 7.1.1 and the likes](https://checkstyle.org/google_style.html#a7.1.1), this is where I believe the validation will start to be funky. Our link finder doesn't look for specific links in the column. [It finds all links including issues, etc.](https://github.com/checkstyle/checkstyle/blob/a9086cbaaa2311fb30fc40494a5fc132a947219b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L1772)

> this is checking if the current section has link but it is not valid ( like [2.2](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8db57f2_2024193251/google_style.html#a2.2) )

"Not valid" is too strong a word. Link is just not a check link. It can be an issue link, link to another part of our website, to a point in the google style, or a JLS link, both of which are technically valid and feasible to have in this document where appropriate. They do appear in this document in random places.

This will (and eventually) hit in other areas like https://checkstyle.org/google_style.html#a4.2 , https://checkstyle.org/google_style.html#a4.4 , and https://checkstyle.org/google_style.html#a4.6.2 .

This is part of where I think your logic falls apart when it is connected with, ...

> if (!moduleIsConfig && !moduleIsCheck) { 
     skip = true; 
> then we will be skipping config link validation.

This is where I have issues with the code as it is.

Just because a link isn't a check doesn't mean we need to skip all further config link validation.

Imagine we have this in a section ([similar to the (old) section 7.1.1](https://checkstyle.org/google_style.html#a7.1.1)):
`````
1st Check Link
Issue Link
2nd Check Link
`````

You will see and process the 1st check link and it's config. However, once you see the issue link, you will basically signal a "stop" and do no validation on test links. You will still process the 2nd check link and it's config (which is why I say skip is slightly confusing), but your ending logic for the test link will never process, though we should, since we saw 2 check links.

This is what led to my statement `It doesn't make sense to skip (test link check) on unrelated content. If we have checks, we must have atleast 1 config link and 1 test link. If we have no checks, we must have no configs or links.`

> I have added the following code which checks if the section does not have any checks then it should also not contain any config/test url links

Which is all surrounded by the `skip` check.

File: src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java [hunk 11]
```diff
@@ -1786,8 +1858,8 @@ private static void validateStyleModules(Set<Node> checks, Set<Node> configs,
                 continue;
             }
 
-            assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' module '" + moduleName
-                        + "' shouldn't end with 'Check'")
+            assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' module '"
+                    + moduleName + "' shouldn't end with 'Check'")
```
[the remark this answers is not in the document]
Zopsss: done

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1801,13 +1873,14 @@ private static void validateStyleModules(Set<Node> checks, Set<Node> configs,
                 }
                 catch (NoSuchElementException ignore) {
                     assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' module '"
-                            + moduleName + "' is missing the config link: " + configName).fail();
+                            + moduleName + "' is missing the config link: " + configName)
+                            .fail();
```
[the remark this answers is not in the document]
Zopsss: done

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1801,13 +1873,14 @@ private static void validateStyleModules(Set<Node> checks, Set<Node> configs,
...
                 }
 
                 assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' module '"
-                                + moduleName + "' has mismatched config/test links")
-                    .that(config.getTextContent().trim())
-                    .isEqualTo(configName);
+                        + moduleName + "' has mismatched config/test links")
+                        .that(config.getTextContent().trim())
+                        .isEqualTo(configName);
```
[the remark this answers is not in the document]
Zopsss: done

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: ignore my above comment, pushed new changes according to your suggestions

Zopsss:
@rnveach pushed new changes according to your suggestions. Thanks for all the help!

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: done

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: done

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: done

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: done

rnveach (approved):

romani (approved):
ok to merge

rdiachenko (approved):
lgtm

rdiachenko:
rebased

PR #14926, opened 2024-06-02
minor: move wiki content of useful commands to code

romani:
content from https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations
is copied to code to let all use it as if we have `just` tool.

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -1130,7 +1130,7 @@ jacoco)
     echo "Reporting to codecov"
     bash <(curl --fail-with-body -s https://codecov.io/bash)
   else
-    echo "No reporting to codecov outside CI"
+    echo "Result is at target/site/jacoco/index.html"
   fi
   ;;
 
@@ -1184,6 +1184,49 @@ check-wildcards-on-pitest-target-classes)
     echo "$CLASSES_NO_WILDCARD"
   fi
   exit "$CLASSES_NO_WILDCARD_COUNT"
+  ;;
+
+verify)
+  mvn -e --no-transfer-progress clean verify
+  ;;
+
+package-all-jar)
+  mvn -e --no-transfer-progress clean package -Passembly
+  ;;
+
+website-only)
+  mvn -e --no-transfer-progress clean site -Pno-validations
+  ;;
+
+pmd)
+  mvn -e --no-transfer-progress clean test-compile pmd:check
+  ;;
+
+spotbugs)
+  mvn -e --no-transfer-progress clean test-compile spotbugs:check
+  ;;
+
+checkstyle)
+  mvn -e --no-transfer-progress clean compile antrun:run@ant-phase-verify
+  ;;
+
+forbiddenapis)
+  mvn -e --no-transfer-progress \
+    clean compile test-compile forbiddenapis:testCheck@forbiddenapis-test
+  ;;
+
+run-test)
+  if [[ -z "$2" ]] ; then
+    echo "Error: test class is not defined."
+    echo "Example: mvn -e --no-transfer-progress clean test -Dtest=XdocsPagesTest,XdocsJavaDocsTest"
+    echo "Example: mvn -e --no-transfer-progress clean test -Dtest=CheckerTest#testDestroy"
+    exit 1
+  fi
+  mvn -e --no-transfer-progress clean test -Dtest="$2"
+  ;;
+
+sevntu)
+  mvn -e --no-transfer-progress clean compile checkstyle:check@sevntu-checkstyle-check
   ;;
 
 *)


.ci/validation.sh
```diff
@@ -1186,6 +1186,63 @@ check-wildcards-on-pitest-target-classes)
...
+test-class)
+  if [[ -z "$2" ]] ; then
+    echo "Error: test class is not defined."
+    echo "Example: mvn -e --no-transfer-progress clean test -Dtest=XdocsPagesTest,XdocsJavaDocsTest"
+    exit 1
+  fi
+  mvn -e --no-transfer-progress clean test -Dtest="$2"
+  ;;
+
+test-method)
```
nrmancuso, line 1227: This can also be used for test classes, right?

.ci/validation.sh
```diff
@@ -1186,6 +1186,63 @@ check-wildcards-on-pitest-target-classes)
...
+  if [[ -n "$2" ]] ; then
+    echo "Error: test method is not defined."
+    echo "Example: mvn -e --no-transfer-progress clean test -Dtest=CheckerTest#testDestroy"
+    exit 1
+  fi
+  mvn -e --no-transfer-progress clean test -Dtest="$2"
+  ;;
+
+jacoco-local)
```
nrmancuso, line 1236: Why is this "local"?

nrmancuso (commented):
Questions:

.ci/validation.sh
romani: Yes, but from users perspective it is different

.ci/validation.sh
romani: We already have `jacoco` in this file.

rnveach: Non-local:
https://github.com/checkstyle/checkstyle/blob/dd4ac5ca90d9f14255c2c7043d423502ca880219/.ci/validation.sh#L1122-L1135

Local:
https://github.com/checkstyle/checkstyle/blob/dd4ac5ca90d9f14255c2c7043d423502ca880219/.ci/validation.sh#L1236-L1240

rnveach: What does `jacoco:check@default-check ` do that it isn't needed in local? Why don't we just have 1 call the other?

romani: That is why I told that wiki commands  and CI command  very different and on local you don't need much of such details.

I just do it in continuation of our "just" discussion. I was told we can put all from wiki :).
I need to keep commands simple to let it be shared with not experienced contributors.

romani: No merged, as is , with minimal turning, or better to stay at wiki and not even try to mix stuff that should not be mixed.

romani: Goal is to let Nick reference command in shell by exact line , not a wiki page.

romani: I can rename old jacoco to jacoco-ci

nrmancuso: > Goal is to let Nick reference command in shell by exact line , not a wiki page.

I am not going to use this, if you are just doing this for me, we can close this PR. The whole point of just is that I can simply type 'just' in the project root, and it tells me each command and exactly what it does. This script will not do that.

romani: I use to have a rule: if command in such validation.sh is not in CI , it should be removed, as it might not be actual and might not work.

But I am ok to keep in such files some commands, and yes it might be sometime more convenient to use code and not a wiki.
So I am ok to keep such commands in this file.
Just share with me good names and I will put them.

romani: I want this PR to be merged :)

nrmancuso: I am good with whatever naming that @rnveach suggests

nrmancuso: > I want this PR to be merged :)

It is still not clear why or who is going to use this,

romani: I will use it, now I will know that .ci/validation.sh prints all commands and I can grep

romani: `jacoco-local)` is removed.

.ci/validation.sh
romani: please suggest better name, if no ideas for now, lets merge and after usage for some time, ti will come.
different names will help easily to grep.

nrmancuso: `run-test`?

romani: What will be name for second command?
They should be similar in naming.
run-test
run-test-method
?

nrmancuso: The parameter is used the same way, why do we need two commands?

romani: combined to "run-test"

nrmancuso (approved):

nrmancuso:
Ok to merge

romani:
@rnveach mentioned that he is not using this at all.

If any improvement noticed, I will improve in separate PR

PR #14973, opened 2024-06-11
Issue #14891: MissingSwitchDefault should not violate switch that con…

### LINKED ISSUE #14891: Add Support for Java 21 Record Pattern : MissingSwitchDefault
child of #14890 

I have read check documentation: https://checkstyle.org/checks/coding/missingswitchdefault.html#MissingSwitchDefault

---

```java

PS D:\CS\test> javac src/RecordPatterns.java --enable-preview --source 22
PS D:\CS\test> cat src/RecordPatterns.java
import java.util.List;

public class RecordPatterns {
    sealed interface I permits C, D { }
    final class C implements I { }
    final class D implements I { }
    record Pair<T>(T x, T y) { }

    void test() {
        C c = new C();
        D d = new D();

        Pair<I> p2 = new Pair<>(c, d);
        // exhaustiveness check by the compiler
        switch (p2) {   // violation ??
            case Pair<I>(I i, C j) -> System.out.println("Pair (C|D, C)");
            case Pair<I>(I i, D j) -> System.out.println("Pair (C|D, D)");
        }
    }

        new RecordPatterns().test();
    }
}
PS D:\CS\test> cat config.xml

<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
 <module name="TreeWalker">
    <module name="MissingSwitchDefault"/>
 </module>
</module>
PS D:\CS\test> java  -jar checkstyle-10.14.2-all.jar -c config.xml  src/RecordPatterns.java
Starting audit...
[ERROR] D:\CS\test\src\RecordPatterns.java:15:9: switch without "default" clause. [MissingSwitchDefault]
Audit done.

``` 

I expect no violation here since there is an exhaustiveness Check by the compiler. This means that all possible inputs must be covered. So there is no need for default

---

mahfouz72:
closes #14891 :

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/fd6bb80bfe66feb1d7bf946caf87c0a7be981f6c/check.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
@@ -50,9 +50,13 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * inputs must be covered.
  * </p>
  * <p>
- * See the <a href="https://docs.oracle.com/javase/specs/jls/se17/html/jls-15.html#jls-15.28">
+ * See the <a href="https://docs.oracle.com/javase/specs/jls/se22/html/jls-15.html#jls-15.28">
  *     Java Language Specification</a> for more information about switch statements
  *     and expressions.
+ * </p>
+ * <p>
+ * See the <a href="https://docs.oracle.com/javase/specs/jls/se22/html/jls-14.html#jls-14.30">
+ *     Java Language Specification</a> for more information about patterns.
  * </p>
  * <p>
  * Parent is {@code com.puppycrawl.tools.checkstyle.TreeWalker}
@@ -116,7 +120,8 @@ public class MissingSwitchDefaultCheck extends AbstractCheck {
     }
 
     /**
-     * Checks if a switch block contains a case label with a pattern variable definition.
+     * Checks if a switch block contains a case label with a pattern variable definition
+     * or record pattern definition.
      * In this situation, the compiler enforces the given switch block to cover
      * all possible inputs, and we do not need a default label.
      *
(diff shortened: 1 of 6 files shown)


mahfouz72:
Github, generate report

romani:
@nrmancuso , @rnveach , try think of we should collect such test bundles (config , list of projects) in our test-configs repository.
This specific is simple default config, but what if any non default come up let's seriously consider to preserve it, or signal to me and Piyush to review.

@relentless-pursuit , as you finish generation of basic set of configs, you will be immediately huge help for others. People will not use gist anymore for simple cases.

nrmancuso (approved):
Let’s go!

rnveach (commented):
Add one more input example where the switch does have a default.

https://checkstyle.org/checks/coding/missingswitchdefault.html#Description
> This check does not validate switch statements that use pattern or null labels. Rationale: Switch statements that use pattern or null labels are checked by the compiler for exhaustiveness.

You made a point to update the javadoc. Shouldn't this be updated as well?

> See the Java Language Specification for more information about switch statements and expressions

Shouldn't we update this to the new JLS also show the new patterns?

mahfouz72:
> > This check does not validate switch statements that use pattern or null labels. Rationale: Switch statements that use pattern or null labels are checked by the compiler for exhaustiveness.
> 
> You made a point to update the javadoc. Shouldn't this be updated as well?

I assumed that the word "pattern" in "that use pattern or null labels." includes both normal type patterns and record pattern

mahfouz72:
Github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
```diff
@@ -55,6 +55,10 @@
  *     and expressions.
  * </p>
  * <p>
+ * See the <a href="https://docs.oracle.com/javase/specs/jls/se17/html/jls-14.html#jls-14.30">
```
rnveach, line 58: > new record pattern syntax introduced in Java 21

I would expect use to use the new JLS in which it was introduced. Update the one before it as well.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
mahfouz72: done

mahfouz72:
Github, generate site

rnveach (approved):

PR #14974, opened 2024-06-11
minor: Add isUnamed method to TokenUtil

mahfouz72:
since unnamed variables are just identifiers with `_` as their text. and we will need to check whether the given local variables, lambda parameters, etc... are unnamed in several checks. I thought it would be good if we separate this method outside the logic of the checks and to be in `TokenUtil`

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
@@ -395,4 +395,34 @@ public class TokenUtilTest {
                 .isFalse();
     }
 
+    @Test
+    public void testIdentIsUnnamedTrue() {
+        final DetailAstImpl identifierAst = new DetailAstImpl();
+        identifierAst.setType(TokenTypes.IDENT);
+        identifierAst.setText("_");
+        final boolean result = TokenUtil.isUnnamed(identifierAst);
+
+        assertWithMessage("Result is not expected")
+                .that(result)
+                .isTrue();
+    }
+
+    @Test
+    public void testIdentIsUnnamedFalse() {
+        final DetailAstImpl identifierAst = new DetailAstImpl();
+        final DetailAstImpl stringLiteralAst = new DetailAstImpl();
+        stringLiteralAst.setType(TokenTypes.STRING_LITERAL);
+        identifierAst.setType(TokenTypes.IDENT);
+        identifierAst.setText("x");
+        stringLiteralAst.setText("_");
+        final boolean result = TokenUtil.isUnnamed(identifierAst);
+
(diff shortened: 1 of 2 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
```diff
@@ -395,4 +395,34 @@ public void testIsBooleanLiteralType() {
...
     }
 
+    @Test
+    public void testIdentIsUnnamedTrue() {
+        final DetailAstImpl identifierAst = new DetailAstImpl();
+        identifierAst.setType(TokenTypes.IDENT);
+        identifierAst.setText("_");
+        final boolean result = TokenUtil.isUnnamed(identifierAst);
+
+        assertWithMessage("Result is not expected")
```
romani, line 405: All tests should be through real check execution.
In other cases we can generate fancy utils with fancy code and and fancy pure junit test method and will still not work in real code and real Check.

romani (changes requested):
Items:

src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java
romani: We already had bunch of such examples in past.
Pure unit testing is forbidden, and allowed only with huge comment above to explain all nuances of why it is not possible to do it through Check. 
Unfortunately there are bunch bad examples in our code base, we were not smart from the beginning.

mahfouz72: Yes, this method will be tested using real check execution once we support unnamed variables in the checks. I think this is the same for all utils here. all TokenUtil Tests are pure JUnit tests but at the same time they are being tested using real check execution in the checks test

mahfouz72: So, I think pureJunit here in TokenUtil is ok as long as we use these utils in some checks then we are sure that they are tested in real check execution in the checks test

mahfouz72: I have a claim it may be wrong but we can even remove almost all tests from `TokenUtilTest` and code coverage and pitest will not complain if we use the utils method appropriately in the checks

romani: I would like to remove all and make sure all is covered by regular Check tests, one day we will do this.

nrmancuso (changes requested):
This should be done in the scope of a bug fix/new check to show necessity. Let’s close this and keep it in mind for later.

PR #14978, opened 2024-06-11
Issue #14688 : LocalVaiableNameCheck should allow unnamed variables b…

### LINKED ISSUE #14688: LocalVariableNameCheck incorrectly flags variables named `_` (see JEP456)
https://openjdk.org/jeps/456 allows `_` to be used as a variable name to signify that it is "unused".

checkstyle issues the following warning on `var _`:

```
Name '_' must match pattern '^[a-z][a-zA-Z0-9]*$'. [LocalVariableName]Lint(Java)
```

Seems like the [regex in `LocalVariableNameCheck`](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/LocalVariableNameCheck.java#L74) needs to be updated to support this new language feature.

-----------------------------------------
I have read check documentation: https://checkstyle.org/checks/naming/localvariablename.html#LocalVariableName

```
import java.util.Queue;

class Test {
    void test() {
        q.add(1);
        while (!q.isEmpty()) {
            var _ = q.poll();   // violation
        }
    }
}
PS D:\CS\test> cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
 <module name="TreeWalker">
    <module name="LocalVariableName">
    </module>
 </module>
</module>
Starting audit...
[ERROR] D:\CS\test\src\Test.java:9:17: Name '_' must match pattern '^[a-z][a-zA-Z0-9]*$'. [LocalVariableName]
Audit done.
```

---

I want this check to allow unnamed variables by default.

---

mahfouz72:
closes #14688 :

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/ec67844f3c91caaff9c853fa121a42c69ec2e0df/check.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/LocalVariableNameCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/LocalVariableNameCheck.java
@@ -41,7 +41,7 @@ import com.puppycrawl.tools.checkstyle.utils.ScopeUtil;
  * <li>
  * Property {@code format} - Sets the pattern to match valid identifiers.
  * Type is {@code java.util.regex.Pattern}.
- * Default value is {@code "^[a-z][a-zA-Z0-9]*$"}.
+ * Default value is {@code "^([a-z][a-zA-Z0-9]*|_)$"}.
  * </li>
  * </ul>
  * <p>
@@ -71,7 +71,7 @@ public class LocalVariableNameCheck
 
     /** Creates a new {@code LocalVariableNameCheck} instance. */
     public LocalVariableNameCheck() {
-        super("^[a-z][a-zA-Z0-9]*$");
+        super("^([a-z][a-zA-Z0-9]*|_)$");
     }
 
     /**
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/naming/LocalVariableNameCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/naming/LocalVariableNameCheck.xml
@@ -18,7 +18,7 @@
  initialization expressions&lt;/a&gt;
  in FOR loop if one char variable name is prohibited by {@code format} regexp.</description>
             </property>
-            <property default-value="^[a-z][a-zA-Z0-9]*$"
+            <property default-value="^([a-z][a-zA-Z0-9]*|_)$"
                       name="format"
                       type="java.util.regex.Pattern">
                <description>Sets the pattern to match valid identifiers.</description>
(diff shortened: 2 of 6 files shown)


mahfouz72:
Github, generate report

mahfouz72:
this check will break on legacy code that can use `_` as a valid identifier (java 9 and below) no violation on underscores.
if we don't want to do this we should introduce a new property

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/LocalVariableNameCheck.java
```diff
@@ -116,7 +117,7 @@ protected final boolean mustCheckName(DetailAST ast) {
             final boolean isFinal = modifiersAST.findFirstToken(TokenTypes.FINAL) != null;
             result = !isFinal && ScopeUtil.isLocalVariableDef(ast);
         }
-        return result;
+        return result && !TokenUtil.isUnnamed(ast.findFirstToken(TokenTypes.IDENT));
```
rnveach, line 120: There are 2 ways to handle this issue. Either we update the existing property regexp to allow underscores, or we update the code to fully ignore single underscore variable names.

The regexp allows anyone to change it and what it accepts. Changing it would allow the users to backward compatible this change.

Change the code hard codes that this is to be a full on ignore which can't be undone. This is similar to ignore `final` variables, as another check handles this. If someone wanted to configure CS to violate underscore variables again, they would have to use another check, `IllegalTokenText` or possibly `IllegalIdentifierName`.

Are we sure we want to basically break compatibility and hard code the ignore? I was originally thinking we would just update the property. I am leaning more to the property just to remain compatible with those not on 21 yet. I remember somewhere we said we wanted to keep backwards compatibility with another check.

@nrmancuso Feel free to review.

nrmancuso: > I was originally thinking we would just update the property.

Yes, please do this @mahfouz72

mahfouz72:
Github, generate site

mahfouz72:
Github, generate site

mahfouz72:
Github, generate report

nrmancuso:
Report is a bit hard to consume, but looks good when you click into projects with an asymmetrical diff and find the places where unnamed variables are present, you can see that there violations have been removed.

Example: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/d3e56f0_2024095635/reports/diff/checkstyle/xref/home/runner/work/checkstyle/checkstyle/.ci-temp/contribution/checkstyle-tester/repositories/checkstyle/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/java21/InputUnnamedVariableBasic.java.html#L6

rnveach:
Since we will probably have more regexp property updates, change the message of the check to a static message so we can see new/removed violations.
Ex: `<message key="name.invalidPattern" value="Hello World" />`

mahfouz72:
Github, generate report

rnveach (approved):

PR #14997, opened 2024-06-15
Issue #14971: Add input file in indentation check for pattern matchin…

### LINKED ISSUE #14971: Add Check Support for Java 21 Pattern Matching for Switch Syntax: Indentation
child of #14961 
I have read check documentation: https://checkstyle.org/checks/misc/indentation.html#Indentation

```
record Point(int x, int y) { }
record square(Point upperLeft, Point lowerRight) { }

public class Test {
    public void test() {
        Point p = new Point(1, 2);
        switch (p) {
            case Point(int x, int y)
            when x > 0 && y > 0 -> System.out.println(x + y); // violation
            case Point(int x, int y) -> System.out.println("error");
        }
    }
}
PS D:\CS\test> cat config.xml                                                    
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
 <module name="TreeWalker">
    <module name="Indentation">
    </module>
 </module>
</module>
Starting audit...
[ERROR] D:\CS\test\src\Test.java:9:13: 'case' child has incorrect indentation level 12, expected level should be 16. [Indentation]
Audit done.
```
---

I consider this a false positive. I want no violation on `when`

---

mahfouz72:
closes #14971

### DIFF
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationPatternMatchingForSwitch.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationPatternMatchingForSwitch.java
@@ -1,0 +1,95 @@
+//non-compiled with javac: Compilable with Java21        //indent:0 exp:0
+package com.puppycrawl.tools.checkstyle.checks.indentation.indentation;//indent:0 exp:0
+//indent:0 exp:0
+//indent:0 exp:0
+/* Config:                                                                  //indent:0 exp:0
+ * basicOffset = 4                                                          //indent:1 exp:1
+ * braceAdjustment = 0                                                      //indent:1 exp:1
+ * caseIndent = 4                                                           //indent:1 exp:1
+ * tabWidth = 4                                                             //indent:1 exp:1
+ * throwsIndent = 8                                                         //indent:1 exp:1
+ * forceStrictCondition = true                                             //indent:1 exp:1
+ */                                                                         //indent:1 exp:1
+public class InputIndentationPatternMatchingForSwitch {                     //indent:0 exp:0
+    record Point(int x, int y) { }                                           //indent:4 exp:4
+    record square(Point upperLeft, Point lowerRight) { }                    //indent:4 exp:4
+
+    public void test1() {   //indent:4 exp:4
+        Point p = new Point(1, 2); //indent:8 exp:8
+        switch (p) { //indent:8 exp:8
+            case Point(int x, int y) //indent:12 exp:12
+            when x > 0 && y > 0 -> System.out.println(x + y); //indent:12 exp:16 warn
+            case Point(int x, int y) -> System.out.println("error"); //indent:12 exp:12
+        } //indent:8 exp:8
+        switch (p) { //indent:8 exp:8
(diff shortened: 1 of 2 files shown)


nrmancuso (approved):
Looks good.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationPatternMatchingForSwitch.java
```diff
@@ -0,0 +1,86 @@
...
+                    Point(int x, int y) //indent:20 exp:20
+                            when x > 0 && y > 0 -> System.out.println(x + y); //indent:28 exp:28
+            case Point(int x, int y) -> System.out.println("error"); //indent:12 exp:12
+        } //indent:8 exp:8
+        switch (p) { //indent:8 exp:8
+            case    //indent:12 exp:12
+                    Point(int x, int y) //indent:20 exp:20
+            when x > 0 && y > 0 -> System.out.println(x + y); //indent:12 exp:16 warn
+            case Point(int x, int y) -> System.out.println("error"); //indent:12 exp:12
+        } //indent:8 exp:8
```
rnveach, line 56: Add an input like: (indentation is intentional)
````
        switch (p) { //indent:8 exp:8
            case    //indent:12 exp:12
Point(int x, int y)
when
x > 0 && y > 0
->
System.out.println(x + y);
            case Point(int x, int y) -> System.out.println("error"); //indent:12 exp:12
        } //indent:8 exp:8
````

rnveach (commented):
Just to be clear, we are only validating `LITERAL_WHEN` and things around it, correct?

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationPatternMatchingForSwitch.java
mahfouz72: > Just to be clear, we are only validating `LITERAL_WHEN` and things around it, correct?

Yes



> Add an input like: (indentation is intentional)

done.

> System.out.println(x + y);

this is weird indent:0 expected:0

rnveach: > this is weird indent:0 expected:0

Exactly. :)

This is a bug. Most likely this line is not being validated.

You will only see issues with this check if you do bad indentation. You other examples were "too good".

After talking with @nrmancuso , move this to a new issue. We need to continue opening other issues and will probably return to indentation later.

rnveach: @mahfouz72 Is this done?

mahfouz72: https://github.com/checkstyle/checkstyle/issues/15098

rnveach (approved):

Issue #15026, opened 2024-06-18
Add support to check input file against whole google configuration for Google IT tests


Zopsss:
Continuation of https://github.com/checkstyle/checkstyle/issues/14929

In the previous issue, we added support to check the input file against the provided list of modules. The verifyWithGooglexxx method for this issue takes file path of the input file and list of modules to load from the google config and check against it the input file. There is no need to provide `expected` violation array. No need to use `// warn` message, instead we require to use `// violation 'violation message'`, following the traditional `verifyWithInlineConfigParser()` way.

In the scope of this issue, we require to create a new `verifyWithWholeGoogleConfig()` ( name can be improved ) method which will only take file path of the input file as input. It will load the whole google config and check it against the input file. We don't need to provide any kind of module list, just the file path of input file.

This way we will need to compulsory follow all the rules of google style guide and show violations only for the required modules.

rnveach:
I am approving this since we decided to go with this approach. We will be using the first few PRs to see how this goes since there was concerns in prior releases on doing this approach.

Issue #14937, opened 2024-06-04
Migrate google style guide tests to follow chapter wise testing


Zopsss:
Currently the google style guide tests are implemented as `Per Module Testing` approach. We need to migrate these tests to use `Chapter Wise Testing` approach.

#### What is Per Module Testing?
The testing approach in which every modules used in a rule will have its special test file. All of these different test files will test different areas of the rule to ensure that we follow the rule correctly. 

#### What is Per Module Testing?
The testing approach in which there will be only one test file which will combine all the different modules used for the corresponding rule and ensure we follow the rule correctly. We combine different modules into single test and test the rule as whole.

The support to test `Chapter wise testing`'s documentation was added in this PR: https://github.com/checkstyle/checkstyle/pull/14925

#### Final Goal:
Migrate all google tests to chapter wise testing by combining different modules into single test file and update the coverage page accordingly. 

Following sections needs to be migrated:

- [x] 2.1 File name
- [x] 2.3.1 Whitespace characters
- [x] 2.3.2 Special escape sequences
- [x] 2.3.3 Non-ASCII characters
- [x] 3 Source file structure
- [x] 3.2 Package statement
- [x] 3.3.1 No wildcard imports
- [x] 3.3.2 No line-wrapping
- [x] 3.3.3 Ordering and spacing
- [x] 3.4.1 Exactly one top-level class declaration
- [x] 3.4.2.1 Overloads: never split
- [x] 4.1.1 Use of optional braces
- [x] 4.1.2 Nonempty blocks: K & R style
- [x] 4.1.3 Empty blocks: may be concise
- [x] 4.2 Block indentation: +2 spaces
- [x] 4.3 One statement per line
- [x] 4.4 Column limit: 100
- [x] 4.5.1 Where to break
- [x] 4.5.2 Indent continuation lines at least +4 spaces
- [x] 4.6.1 Vertical Whitespace
- [x] 4.6.2 Horizontal whitespace
- [x] 4.8.2.1 One variable per declaration
- [x] 4.8.2.2 Declared when needed
- [x] 4.8.3.2 No C-style array declarations
- [x] 4.8.4.1 Indentation
- [x] 4.8.4.2 Fall-through: commented
- [x] 4.8.4.3 Presence of the default label
- [x] 4.8.6.1 Block comment style
- [x] 4.8.7 Modifiers
- [x] 4.8.8 Numeric Literals
- [x] 5.2.1 Package names
- [x] 5.2.2 Class names
[...]

PR #15028, opened 2024-06-18
Issue #15026: Added support to check input file against whole google configuration for google IT tests

Zopsss:
issue #15026 

created a new method: `verifyWithWholeGoogleConfig()` which helps us to check the input file against whole google configuration. We just require to pass the file path to the method and everything else will work like a magic 🔮🪄

### DIFF
--- a/src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
+++ b/src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
@@ -20,13 +20,19 @@
 package com.google.checkstyle.test.base;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Properties;
 import java.util.Set;
 
 import org.checkstyle.base.AbstractItModuleTestSupport;
 
 import com.puppycrawl.tools.checkstyle.ConfigurationLoader;
+import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
 import com.puppycrawl.tools.checkstyle.PropertiesExpander;
+import com.puppycrawl.tools.checkstyle.TreeWalker;
 import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
 import com.puppycrawl.tools.checkstyle.api.Configuration;
 import com.puppycrawl.tools.checkstyle.internal.utils.CheckUtil;
@@ -40,10 +46,33 @@ public abstract class AbstractGoogleModuleTestSupport extends AbstractItModuleTe
 
     private static final Set<Class<?>> CHECKSTYLE_MODULES;
 
+    private static final Set<String> CHECKER_CHILDREN = new HashSet<>(Arrays.asList(
+            "BeforeExecutionExclusionFileFilter",
+            "SeverityMatchFilter",
+            "SuppressionFilter",
+            "SuppressionSingleFilter",
+            "SuppressWarningsFilter",
+            "SuppressWithNearbyTextFilter",
+            "SuppressWithPlainTextCommentFilter",
+            "JavadocPackage",
+            "NewlineAtEndOfFile",
+            "UniqueProperties",
+            "OrderedProperties",
+            "RegexpMultiline",
+            "RegexpSingleline",
+            "RegexpOnFilename",
+            "FileLength",
+            "LineLength",
+            "FileTabCharacter"
+    ));
+
     static {
         try {
+            final Properties properties = new Properties();
+            properties.put("org.checkstyle.google.severity", "error");
+            final PropertiesExpander expander = new PropertiesExpander(properties);
             CONFIGURATION = ConfigurationLoader.loadConfiguration(XML_NAME,
-                    new PropertiesExpander(System.getProperties()));
+                    expander);
         }
         catch (CheckstyleException ex) {
             throw new IllegalStateException(ex);
@@ -110,4 +139,71 @@ public abstract class AbstractGoogleModuleTestSupport extends AbstractItModuleTe
         return getModuleConfigsByIds(CONFIGURATION, moduleIds);
     }
 
+    /**
+     * Performs verification of the file with the given file path against the whole config.
+     *
+     * @param filePath file path to verify.
+     * @throws Exception if exception occurs during verification process.
+     */
+    protected void verifyWithWholeConfig(String filePath) throws Exception {
+        verifyWithItConfig(CONFIGURATION, filePath);
+    }
+
+    // until https://github.com/checkstyle/checkstyle/issues/14937
+    /**
+     * Performs verification of the file with the given file path against config.
+     * It uses the specified list of modules to load them from config for validation.
+     *
+     * @param listOfModules list of modules to load from config.
+     * @param filePath file path to verify.
+     * @throws Exception if exception occurs during verification process.
+     */
+    protected final void verifyWithConfigParser(String[] listOfModules,
+           String filePath) throws Exception {
+        final DefaultConfiguration rootConfig = new DefaultConfiguration(ROOT_MODULE_NAME);
+        final DefaultConfiguration treeWalkerConfig =
+                new DefaultConfiguration(TreeWalker.class.getSimpleName());
+
+        for (String module : listOfModules) {
+            final List<Configuration> children =
+                    getConfigChildren(module);
+            if (CHECKER_CHILDREN.contains(module)) {
+                for (Configuration child : children) {
+                    rootConfig.addChild(child);
+                }
+            }
+            else {
+                for (Configuration child : children) {
+                    treeWalkerConfig.addChild(child);
+                }
+            }
+        }
+
+        if (treeWalkerConfig.getChildren().length > 0) {
+            rootConfig.addChild(treeWalkerConfig);
+        }
+
+        verifyWithItConfig(rootConfig, filePath);
+    }
+
+    /**
+     * Gets the specified children module(s) from the config.
+     *
+     * @param module the children module to get. It can be either a module name or module id.
+     * @return list of children modules.
+     * @throws CheckstyleException if there is a problem for getting the module.
+     */
+    private static List<Configuration> getConfigChildren(String module)
+            throws CheckstyleException {
+        final List<Configuration> children = new ArrayList<>();
+
+        try {
+            children.add(getModuleConfig(module));
+        }
+        catch (IllegalStateException ex) {
+            children.addAll(getModuleConfigsByIds(module));
+        }
+
+        return children;
+    }
 }
--- a/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
+++ b/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
@@ -33,23 +33,18 @@ import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.text.MessageFormat;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
-import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import com.puppycrawl.tools.checkstyle.AbstractPathTestSupport;
 import com.puppycrawl.tools.checkstyle.Checker;
-import com.puppycrawl.tools.checkstyle.ConfigurationLoader;
 import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
-import com.puppycrawl.tools.checkstyle.PropertiesExpander;
 import com.puppycrawl.tools.checkstyle.TreeWalker;
 import com.puppycrawl.tools.checkstyle.api.AbstractViolationReporter;
 import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
@@ -83,26 +78,6 @@ public abstract class AbstractItModuleTestSupport extends AbstractPathTestSuppor
 
     private static final Pattern WARN_PATTERN = CommonUtil
             .createPattern(".* *// *warn *|/[*]\\*?\\s?warn\\s?[*]/");
-
-    private static final Set<String> CHECKER_CHILDREN = new HashSet<>(Arrays.asList(
-            "BeforeExecutionExclusionFileFilter",
-            "SeverityMatchFilter",
-            "SuppressionFilter",
-            "SuppressionSingleFilter",
-            "SuppressWarningsFilter",
-            "SuppressWithNearbyTextFilter",
-            "SuppressWithPlainTextCommentFilter",
-            "JavadocPackage",
-            "NewlineAtEndOfFile",
-            "UniqueProperties",
-            "OrderedProperties",
-            "RegexpMultiline",
-            "RegexpSingleline",
-            "RegexpOnFilename",
-            "FileLength",
-            "LineLength",
-            "FileTabCharacter"
-    ));
 
     private final ByteArrayOutputStream stream = new ByteArrayOutputStream();
 
@@ -318,7 +293,8 @@ public abstract class AbstractItModuleTestSupport extends AbstractPathTestSuppor
             final Configuration config = createTreeWalkerConfig(moduleConfig);
             checker.configure(config);
         }
-        else if (ROOT_MODULE_NAME.equals(moduleConfig.getName())) {
+        else if (ROOT_MODULE_NAME.equals(moduleConfig.getName())
+                || "Checker".equals(moduleConfig.getName())) {
             checker.configure(moduleConfig);
         }
         else {
@@ -481,21 +457,16 @@ public abstract class AbstractItModuleTestSupport extends AbstractPathTestSuppor
     }
 
     /**
-     * Performs verification of the file with the given file path against google config.
-     * It uses the specified list of modules to load them from google config for validation.
-     * And also uses the array of expected messages for verification.
+     * Performs the verification of the file with the given file path and config.
      *
-     * @param listOfModules list of modules to load from google config.
-     * @param filePath file path to verify.
+     * @param config config to check against.
+     * @param filePath input file path.
      * @throws Exception if exception occurs during verification process.
      */
-    public final void verifyWithGoogleConfigParser(String[] listOfModules,
-           String filePath) throws Exception {
-        final DefaultConfiguration googleConfig =
-                createConfigForModulesOfGoogleConfig(listOfModules);
+    protected void verifyWithItConfig(Configuration config, String filePath) throws Exception {
         final List<TestInputViolation> violations =
-                InlineConfigParser.getViolationsFromInputFile(filePath);
-        final List<String> actualViolations = getActualViolationsForFile(googleConfig, filePath);
+            InlineConfigParser.getViolationsFromInputFile(filePath);
+        final List<String> actualViolations = getActualViolationsForFile(config, filePath);
 
         verifyViolations(filePath, violations, actualViolations);
     }
@@ -508,7 +479,7 @@ public abstract class AbstractItModuleTestSupport extends AbstractPathTestSuppor
      * @return list of actual violations.
      * @throws Exception if exception occurs during verification process.
      */
-    private List<String> getActualViolationsForFile(DefaultConfiguration config,
+    private List<String> getActualViolationsForFile(Configuration config,
           String file) throws Exception {
         stream.flush();
         stream.reset();
@@ -551,81 +522,6 @@ public abstract class AbstractItModuleTestSupport extends AbstractPathTestSuppor
 
             return actualViolations;
         }
-    }
-
-    /**
-     * Creates a google configuration based on the list of modules.
-     *
-     * @param listOfModules list of modules to load from google config.
-     * @return google configuration based on the list of modules.
-     * @throws CheckstyleException if there is a problem loading the configuration.
-     */
-    private static DefaultConfiguration createConfigForModulesOfGoogleConfig(
-            String... listOfModules) throws CheckstyleException {
-        final Configuration masterGoogleConfig = getMasterGoogleConfig();
-        final DefaultConfiguration root = new DefaultConfiguration(ROOT_MODULE_NAME);
-        final DefaultConfiguration treeWalkerConfig =
-                new DefaultConfiguration(TreeWalker.class.getSimpleName());
-
-        for (String module : listOfModules) {
-            final List<Configuration> children =
-                    getGoogleConfigChildren(masterGoogleConfig, module);
-            if (CHECKER_CHILDREN.contains(module)) {
-                for (Configuration child : children) {
-                    root.addChild(child);
-                }
-            }
-            else {
-                for (Configuration child : children) {
-                    treeWalkerConfig.addChild(child);
-                }
-            }
-        }
-
-        if (treeWalkerConfig.getChildren().length > 0) {
-            root.addChild(treeWalkerConfig);
-        }
-        return root;
-    }
-
-    /**
-     * Returns the whole google configuration.
-     *
-     * @return the google config present in the google_checks.xml
-     * @throws IllegalStateException if there is a problem loading the configuration.
-     */
-    private static Configuration getMasterGoogleConfig() {
-        final Configuration masterGoogleConfig;
-        try {
-            masterGoogleConfig =
-                    ConfigurationLoader.loadConfiguration("/google_checks.xml",
-                            new PropertiesExpander(System.getProperties()));
-        }
-        catch (CheckstyleException ex) {
-            throw new IllegalStateException(ex);
-        }
-        return masterGoogleConfig;
-    }
-
-    /**
-     * Gets the specified children module(s) from google config.
-     *
-     * @param masterGoogleConfig the master google config.
-     * @param module the children module to get. It can be either a module name or module id.
-     * @return list of children modules.
-     * @throws CheckstyleException if there is a problem for getting the module.
-     */
-    private static List<Configuration> getGoogleConfigChildren(Configuration masterGoogleConfig,
-           String module) throws CheckstyleException {
-        final List<Configuration> children = new ArrayList<>();
-        try {
-            children.add(getModuleConfig(masterGoogleConfig, module, null));
-        }
-        catch (IllegalStateException ex) {
-            children.addAll(getModuleConfigsByIds(masterGoogleConfig, module));
-        }
-
-        return children;
     }
 
     /**
--- a/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/InputClassAnnotations.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/InputClassAnnotations.java
@@ -1,35 +1,44 @@
 package com.google.checkstyle.test.chapter4formatting.rule4852classannotations;
 
+/**
+ * Test class for checking the rule 4852.
+ */
 public class InputClassAnnotations {
-    public @interface SomeAnnotation1 { }
+  /**
+   * Custom annotation 1.
+   */
+  public @interface SomeAnnotation1 { }
 
-    public @interface SomeAnnotation2 { }
+  /**
+   * Custom annotation 2.
+   */
+  public @interface SomeAnnotation2 { }
 
-    /**
-     * Inner class 2.
-     */
-    @SomeAnnotation1
-    @SomeAnnotation2
-    class Inner1 {}
+  /**
+   * Inner class 2.
+   */
+  @SomeAnnotation1
+  @SomeAnnotation2
+  class Inner1 {}
 
-    @SomeAnnotation1
-    @SomeAnnotation2
-    /** // violation 'Javadoc comment is placed in the wrong location.'
-     * Inner class 2.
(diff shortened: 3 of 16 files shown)


src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
```diff
@@ -500,6 +500,28 @@ public final void verifyWithGoogleConfigParser(String[] listOfModules,
...
     }
 
+    /**
+     * Performs verification of the file with the given file path against the whole google config.
+     *
+     * @param filePath file path to verify.
+     * @throws Exception if exception occurs during verification process.
+     */
+    public void verifyWithWholeGoogleConfig(String filePath) throws Exception {
+        final Configuration googleConfig = getMasterGoogleConfig();
```
rnveach, line 510: This type of method should not be in `AbstractIt`. This class is used for multiple different areas of IT, including Sun and CS's regression area. It doesn't make sense for sun to be able to see google's config. This is why the abtsracts are separated out like this. This type of mixing is almost like you gave production code access to test code. They should be kept separate.

Why can't we have:
* `verifyWithItConfig` method in this abstract class, which will take in the entire config.
* `verifyWithWholeConfig` in `AbstractGoogleModuleTestSupport` which will call `verifyWithItConfig` and pass it Google's config.

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
```diff
@@ -491,7 +491,7 @@ protected final void verify(Checker checker,
      */
     public final void verifyWithGoogleConfigParser(String[] listOfModules,
```
rnveach, line 492: Same here.

rnveach (changes requested):

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
rnveach: Same with `getMasterGoogleConfig`, `getGoogleConfigChildren`, and anything else that says "google" in this abstract IT class.

rnveach: It looks like some methods could just be renamed and not split apart.

`getGoogleConfigChildren` really has no code specific to google, except in method name and one parameter name. So it can be renamed.

`getMasterGoogleConfig` looks like it could be made abstract and renamed to `getItConfig`, and the method in `AbstractGoogleModuleTestSupport` could just return `CONFIGURATION` which it already loads and caches for you.

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
romani: Agree, nothing special for Google here. It just execution based on some config. Each exact test will define what config to use.

Zopsss: so we will need to make changes for `verifyWithGoogleConfigParser()` too to meet these criteria.

rnveach: >Each exact test will define what config to use.

`AbstractGoogle` should define the config to use. Test should define the verify method to use.

romani: Yes, we can have some abstract class for all Google ITs to define path it.
Parent class for `abstract class for all Google ITs` should just receive any path to config and just load it.

Zopsss: I have moved all the google related methods to the `AbstractGoogleModuleTestSupport` and kept only required methods in `AbstractItModuleTestSupport`

created a new [verifyWithItConfig()](https://github.com/checkstyle/checkstyle/blob/c3f84d08cce5ad488d54a0ca210bfcb0b475bb69/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java#L465) as @rnveach suggested.

This method takes a configuration and file path and performs the verification. We're not loading configuration in this method, we're receiving it from parameters, here is why I think it is better:

In future if we decide to migrate sun style to chapter wise testing or add new style which will follow chapter wise testing, it can work in the same way as google. We will load it's configuration from it's abstract class and pass it to this method as we're doing currently ( providing config from `AbstractGoogleModuleTestSupport` ). This will remove the overhead of loading all types of configuration from one place. We can provide the configuration of different styles from their corresponding abstract class. This is more flexible way.

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
Zopsss: > getGoogleConfigChildren really has no code specific to google, except in method name and one parameter name. So it can be renamed.

This method is only used inside [verifyWithGoogleConfigParser()](https://github.com/checkstyle/checkstyle/blob/c3f84d08cce5ad488d54a0ca210bfcb0b475bb69/src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java#L163C23-L163C51) for creating the google config. So instead of renaming it, I have moved it to `AbstractGoogleModuleTestSupport` with other methods.

> getMasterGoogleConfig looks like it could be made abstract and renamed to getItConfig, and the method in AbstractGoogleModuleTestSupport could just return CONFIGURATION which it already loads and caches for you.

In the new changes, I created the `verifyWithItConfig()` method which will directly receive the configuration from `AbstractGoogleModuleTestSupport` as its parameter. So we will not need `getItConfig` to get the configuration.

Zopsss: Here's the new code for you to understand it better:

https://github.com/checkstyle/checkstyle/blob/c3f84d08cce5ad488d54a0ca210bfcb0b475bb69/src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java#L138-L210

https://github.com/checkstyle/checkstyle/blob/c3f84d08cce5ad488d54a0ca210bfcb0b475bb69/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java#L458-L471

also take a look at this comment https://github.com/checkstyle/checkstyle/pull/15028#discussion_r1646095246, here I have explained the reason behind why I kept working of `verifyWithItConfig` like this.

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
Zopsss: Now the `AbstractItModuleTestSupport` has only `verifyWithItConfig()` and some other helper methods to perform the verification. All the google related methods has been moved to `AbstractGoogleModuleTestSupport`

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
```diff
@@ -110,4 +135,77 @@ protected static List<Configuration> getModuleConfigsByIds(String... moduleIds)
...
+    }
+
+    /**
+     * Gets the specified children module(s) from google config.
+     *
+     * @param module the children module to get. It can be either a module name or module id.
+     * @return list of children modules.
+     * @throws CheckstyleException if there is a problem for getting the module.
+     */
+    private static List<Configuration> getGoogleConfigChildren(String module)
```
romani, line 198: `getConfigChildren`

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
```diff
@@ -110,4 +135,77 @@ protected static List<Configuration> getModuleConfigsByIds(String... moduleIds)
...
+
+        if (treeWalkerConfig.getChildren().length > 0) {
+            googleConfig.addChild(treeWalkerConfig);
+        }
+
+        verifyWithItConfig(googleConfig, filePath);
+    }
+
+    /**
+     * Gets the specified children module(s) from google config.
```
romani, line 192: remove "google"

src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/InputClassAnnotations.java
```diff
@@ -1,35 +1,44 @@
 package com.google.checkstyle.test.chapter4formatting.rule4852classannotations;
 
+/**
+ * Test class for checking the rule 4852.
+ */
```
romani, line 5: can we avoid such javadocs ?
this is already in name of fodler and package

src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/InputClassAnnotations.java
```diff
@@ -1,35 +1,44 @@
...
 public class InputClassAnnotations {
-    public @interface SomeAnnotation1 { }
+  /**
+   * Custom annotation 1.
+   */
```
romani, line 9: lets skip such comments.
Let minimize content that is not related to testing of chapter.

romani (changes requested):
last minors:

src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/InputClassAnnotations.java
Zopsss: We require javadocs for interfaces, classes, etc otherwise google config will give warnings

src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/InputClassAnnotations.java
Zopsss: We're checking against the whole google config so as of it's rules, we need javadoc for classes, interfaces, etc

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
Zopsss: done

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
Zopsss: done

romani (approved):
ok to merge.

@rnveach , please share your concerns and think if we can improve in separate PR on them.
This PR is "blocking" migration to whole configs, that will be significant effort, but it will NOT touch Abstract classe, so we can do fixes in parallel.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
```diff
@@ -110,4 +135,77 @@ protected static List<Configuration> getModuleConfigsByIds(String... moduleIds)
         return getModuleConfigsByIds(CONFIGURATION, moduleIds);
     }
 
+    /**
+     * Performs verification of the file with the given file path against the whole google config.
+     *
+     * @param filePath file path to verify.
+     * @throws Exception if exception occurs during verification process.
+     */
+    public void verifyWithWholeGoogleConfig(String filePath) throws Exception {
```
rnveach, line 144: No public methods except for tests. If something can be made `private`, then it should be that. Otherwise it should be `protected`.

I wouldn't name methods `Google` while in a `Google` abstract class. It sounds like code which calls `toString` on a `String`. Make sure to do this for others and update the javadocs.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
```diff
@@ -110,4 +135,77 @@ protected static List<Configuration> getModuleConfigsByIds(String... moduleIds)
...
+    }
+
+    /**
+     * Gets the specified children module(s) from the config.
+     *
+     * @param module the children module to get. It can be either a module name or module id.
+     * @return list of children modules.
+     * @throws CheckstyleException if there is a problem for getting the module.
+     */
+    private static List<Configuration> getConfigChildren(String module)
```
rnveach, line 198: This looks like it should be in `AbstractItModuleTestSupport`. I am not seeing anything that says "Google".

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
```diff
@@ -110,4 +135,77 @@ protected static List<Configuration> getModuleConfigsByIds(String... moduleIds)
...
+     */
+    public void verifyWithWholeGoogleConfig(String filePath) throws Exception {
+        final Configuration[] children = CONFIGURATION.getChildren();
+        final DefaultConfiguration root = new DefaultConfiguration(ROOT_MODULE_NAME);
+
+        for (Configuration child : children) {
+            root.addChild(child);
+        }
+
+        verifyWithItConfig(root, filePath);
```
rnveach, line 152: > verifyWithItConfig method in this abstract class, which will take in the entire config.

Why can't we replace `root` parameter with `CONFIGURATION`? The code above looks like it is not "Google" specific.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
```diff
@@ -110,4 +135,77 @@ protected static List<Configuration> getModuleConfigsByIds(String... moduleIds)
...
+                    treeWalkerConfig.addChild(child);
+                }
+            }
+        }
+
+        if (treeWalkerConfig.getChildren().length > 0) {
+            googleConfig.addChild(treeWalkerConfig);
+        }
+
+        verifyWithItConfig(googleConfig, filePath);
```
rnveach, line 188: Same as other `verifyWithItConfig` comment.

rnveach (changes requested):

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
romani: as we extend abstract classes, we inherit from most basic methods without "Google" in name, so in final Test class it might be confusing what method to use as it has all methods available for use. Imaginary: verifyWithWholeGoogleConfig , verifyWithItConfig, verifyWithConfig, verifyWithModules

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
Zopsss: It is used inside verifyWithGoogleConfigParser() :

https://github.com/checkstyle/checkstyle/blob/aff00e1ab05cc59fc35158b99cf380ae1cabe3ca/src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java#L169-L172

It can't be moved to IT abstract class

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
Zopsss: Following up Roman's reply... Currently we're able to access the `verifyWithItConfig` in test files. I couldn't find a way to make it not accessible there. It would be helpful if you can suggest something about this. Thanks for pointing this out

rnveach: > Currently we're able to access the verifyWithItConfig in test files. I couldn't find a way to make it not accessible there. It would be helpful if you can suggest something about this. 

There is no way to prohibit this in Java. We would have to make the method private but then `AbstractGoogle` wouldn't be able to use it. We are stuck for now.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
rnveach: Can't `verifyWithGoogleConfigParser` still access it when we switch the method to `protected`? I am still not seeing an issue.

Zopsss: Yes it will be accessible. I kept it public because all other `verifyWithxxxxx` methods in `AbstractModuleTestSupport` are also public. 

I'll change this.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
Zopsss: we cannot directly pass the `CONFIGURATION` to the `verifyWithItConfig()` because `verifyWithItConfig()` uses [getActualViolationsForFile()](https://github.com/checkstyle/checkstyle/blob/aff00e1ab05cc59fc35158b99cf380ae1cabe3ca/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java#L481) under the hood and this method uses `createChecker()` inside it:

https://github.com/checkstyle/checkstyle/blob/aff00e1ab05cc59fc35158b99cf380ae1cabe3ca/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java#L484-L487

The `createChecker()` method requires the top level module name to be `root` otherwise it gives error. 

That's why we're getting the whole configuration, changing the top level module name and then passing it to the `verifyWithItConfig()`:

https://github.com/checkstyle/checkstyle/blob/aff00e1ab05cc59fc35158b99cf380ae1cabe3ca/src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java#L144-L153


This is also the same reason behind when using the `verifyWithGoogleConfigParser()` method, we create the config with the `root` as top level module:

https://github.com/checkstyle/checkstyle/blob/aff00e1ab05cc59fc35158b99cf380ae1cabe3ca/src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java#L163-L165

rnveach: > The createChecker() method requires the top level module name to be root otherwise it gives error.

This was resolved with the git diff I provided in discord.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
rnveach: I am going to drop this move for now.

I still think `AbstractItConfig` is the way to go, but it is more work than we need right now. It will have to be revisted when Sun is updated.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
rnveach: > I wouldn't name methods Google while in a Google abstract class. 

I still think a lot of "Google" words in this abstract should be removed. Method names and local variables.

Truthfully the only "Google" things in this class appear to be `CONFIGURATION` and `XML_NAME` and neither is named google either. It is because it is Google's configuration and Google's XML. The name is trimmed because the abstract class is Google's.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
rnveach: This is also dropped for the same reason, but also this whole method should be removed eventually when we switch everything to the whole config.

rnveach: @Zopsss Where is issue to eventually go back to convert other tests? This should be added to the end there and a comment `until http://....` should be added to this method.

Zopsss: https://github.com/checkstyle/checkstyle/issues/14937

Zopsss: added `until` message

https://github.com/checkstyle/checkstyle/blob/1b71a042246823ec7fdc3a752575c900c2b2aa14/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java#L459-L467

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
Zopsss: done added your changes

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
Zopsss: done. removed `google` from every method and javadoc. Changed verify methods names to 

`verifyWithWholeConfig()`
`verifyWithConfigParser()` as discussed in discord

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
Zopsss: > This is also the same reason behind when using the verifyWithGoogleConfigParser() method, we create the config with the root as top level module:

I have still not changed the code for `verifyWithGoogleConfigParser`, I could not find a way to create config which will have same Checker properties as our google config and I guess it is okay to keep this code as we're not checking against the whole config file, we're checking against only some specific modules

src/main/resources/google_checks.xml
```diff
@@ -21,7 +21,7 @@
 
   <property name="charset" value="UTF-8"/>
 
-  <property name="severity" value="warning"/>
+  <property name="severity" value="${org.checkstyle.google.severity}" default="warning"/>
```
romani, line 24: this is not misc update anymore,
I created https://github.com/checkstyle/checkstyle/issues/15053 please move it out to separate PR and we will merge quicker.

romani (changes requested):
items:

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
rnveach: @Zopsss `until` was placed in wrong location. `until` is basically reminding us the mentioned method is to be removed. `verifyWithItConfig` is the way we will be proceeding now (unless something comes up), so this method should stay.

The note should be on `verifyWithConfigParser` in `AbstractGoogle`. We are only keeping it until all tests move over to full config testing. Once we do, it can be removed completely.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
rnveach: As I mentioned in other comment, `verifyWithGoogleConfigParser` should be completely removed eventually, so it is ok that it doesn't do everything since I see it as the old/bad form of testing.

>  I could not find a way to create config which will have same Checker properties as our google config

Just to provide you clarity, what you are missing in the code is calling `CONFIGURATION.getProperties()`. You call get children which only returns the children modules. This is why the original code for `verifyWithWholeConfig` did not pull in the `Checker` properties either.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
rnveach: @Zopsss This doesn't look completely done.

> final DefaultConfiguration googleConfig = new DefaultConfiguration(ROOT_MODULE_NAME);

rnveach: This one is resolved now.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
rnveach: @Zopsss This doesn't look completely done. I see one left.

> final DefaultConfiguration googleConfig = new DefaultConfiguration(ROOT_MODULE_NAME);

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
rnveach: I am resolving this. If there is any further questions or clarifications needed, please reach out in discord.

Zopsss: > As I mentioned in other comment, verifyWithGoogleConfigParser should be completely removed eventually, so it is ok that it doesn't do everything since I see it as the old/bad form of testing.

I think we should keep this method for very large input files. For now lets keep this `until` message, we can discuss this later in discord

> Just to provide you clarity, what you are missing in the code is calling CONFIGURATION.getProperties(). You call get children which only returns the children modules. This is why the original code for verifyWithWholeConfig did not pull in the Checker properties either.

yep I found the way after I commented

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
Zopsss: done

src/main/resources/google_checks.xml
rnveach: Documentation also needs to be added for this new property. Similar to https://checkstyle.org/google_style.html#Google_Suppressions .

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
rnveach: @Zopsss 
> No public methods except for tests. If something can be made private, then it should be that. Otherwise it should be protected.

This is not done either. Let me know if there is some conflict.

Zopsss: > @Zopsss This doesn't look completely done. I see one left.
> 
> > final DefaultConfiguration googleConfig = new DefaultConfiguration(ROOT_MODULE_NAME);

In this method, we are only loading some specific modules so we should create a config which contains only these modules and their properties. If we create the config whose root will be google_checks.xml's Checker module then we will be copying its properties to the new generated config. So should we really change this and use original Checker module of google_checks.xml as the top level module?

rnveach: I was only commenting to remove `google` from the variable name `googleConfig`. I wasn't trying to change any logic.

src/main/resources/google_checks.xml
Zopsss: made PR on the linked issue, pls check it out

rnveach:
Blocked from merging by https://github.com/checkstyle/checkstyle/pull/15072 , but code updates and reviews can still be done.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
Zopsss: oh, sorry I misunderstood. Changed the variable name now to `rootConfig`

Zopsss: > @Zopsss
> 
> > No public methods except for tests. If something can be made private, then it should be that. Otherwise it should be protected.
> 
> This is not done either. Let me know if there is some conflict.
> 
> `verifyWithConfigParser` and `verifyWithWholeConfig` are both `public`.

done. made them protected

romani:
@rnveach , please review this PR as if it does not have update in google_config.xml (moved to separate PR).
if no other comments, please approve PR and I will make sure that https://github.com/checkstyle/checkstyle/pull/15072 is merged first and this PR is rebased.

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
```diff
@@ -110,4 +139,71 @@ protected static List<Configuration> getModuleConfigsByIds(String... moduleIds)
...
+     * Performs verification of the file with the given file path against the whole config.
+     *
+     * @param filePath file path to verify.
+     * @throws Exception if exception occurs during verification process.
+     */
+    protected void verifyWithWholeConfig(String filePath) throws Exception {
+        verifyWithItConfig(CONFIGURATION, filePath);
+    }
+
+    // until https://github.com/checkstyle/checkstyle/issues/14937
```
rdiachenko, line 152: Assuming this method will be removed upon migration completion, I'm fine to keep its logic as it is now.

rdiachenko (approved):
lgtm

src/it/java/com/google/checkstyle/test/base/AbstractGoogleModuleTestSupport.java
Zopsss: This method might stay, we will use it for complicated input files for which following all googles rules would be difficult.

rnveach: This would still go back to why don't we go against the entire config which started us on this path. I understand not everything can be done in the short time of GSoC, but even complicated input files should be converted eventually.

Zopsss: Yess, In GSoC time period I'm not sure if I can convert all tests to use whole config as I have other issues to work on too. But after GSoC, I can complete it.

rnveach:
@Zopsss Please rebase.

romani (approved):
ok to merge

rnveach (approved):

PR #15031, opened 2024-06-19
Issue #14937: Migrated section 3.4.2.1 Overloads: Never Split to follow Chapter Wise Testing approach

Zopsss:
issue #14937 

migrated the section and added some new test cases in `records` input file


src/xdocs/google_style.xml
```diff
@@ -562,7 +562,7 @@
                 </td>
                 <td>
                   <a href="styleguides/google-java-style-20220203/javaguide.html#s3.4.2.1-overloads-never-split">
-                    3.4.2.1 Overloads: never split</a>
+                    3.4.2.1 Overloads Never Split</a>
                 </td>
```
Zopsss, line 566: had to modify the section name to make XdocsPagesTest happy

config/checkstyle-resources-suppressions.xml
```diff
@@ -804,6 +804,8 @@
     files="[\\/]it[\\/]resources[\\/]com[\\/]google[\\/]checkstyle[\\/]test[\\/]chapter2filebasic[\\/]rule231filetab[\\/]InputFileTabCharacter\.java"/>
   <suppress checks="FileLength"
     files="[\\/]it[\\/]resources[\\/]com[\\/]google[\\/]checkstyle[\\/]test[\\/]chapter2filebasic[\\/]rule232specialescape[\\/]InputIllegalTokenText\.java"/>
+  <suppress checks="FileLength"
+    files="[\\/]it[\\/]resources[\\/]com[\\/]google[\\/]checkstyle[\\/]test[\\/]chapter3filestructure[\\/]rule3421overloadsplit[\\/]InputOverloadsNeverSplit\.java"/>
```
romani, line 808: this will be a pain.

https://github.com/checkstyle/checkstyle/blob/104cf7723d2e7b38230e4792417462c52444043f/config/checkstyle-resources-checks.xml#L41-L44

we did this limit 120 for Inputs that are in test, that contributors are using to debug, so small files means easier to debug.
In ITs we do not need debug almost ever. So we do not need this limit.
let put single suppression for all ITs for FileLength.

romani (changes requested):
Is it possible to have at least one Input that have no violations ? like: InputCorrectOverloadsNeverSplit

all other Inputs can have violations and any name we like
we will use such Inputs to run google formatter to make sure we are not conflicting with it, files will be selected by  pattern `**/it/**/InputCorrect*` 

item:

Zopsss:
CI is failing but it does not affect the PR

https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=21783&view=logs&j=d72e04f6-d294-5bfe-b3cd-ffcb523f40f9

Zopsss:
> Is it possible to have at least one Input that have no violations ? like: InputCorrectOverloadsNeverSplit
> 
> all other Inputs can have violations and any name we like we will use such Inputs to run google formatter to make sure we are not conflicting with it, files will be selected by pattern `**/it/**/InputCorrect*`


I think we should do this when we're using `verifyWithWholeGoogleConfig()` because then we will be following all the rules of google style. 

With the current method we're only following some specific rules. If we run google formatter against these inputs then it will format all the bad code and it will be hard to find if we're following the style guide properly or not.

romani:
Sounds good. We will do this as separate phase of updates.

config/checkstyle-resources-suppressions.xml
Zopsss: done 

https://github.com/checkstyle/checkstyle/blob/1922ffb7a499c0bb0993022ba97fe7728f66aa67/config/checkstyle-resources-suppressions.xml#L199-L202

romani (approved):
Ok to merge

romani:
@rdiachenko , canceled jobs can be ignored for this PR.

src/xdocs/google_style.xml
rnveach: @romani Are these section names suppose to be a duplicate of the style guide's own names?

If so, I think this means we should change xdoc's logic. Possibly remove non-alphabetic characters, remove spaces, and equals ignoring case.

Zopsss: I have fixed this in my other PR: https://github.com/checkstyle/checkstyle/pull/15030#discussion_r1647950400

When it will get merged, I will fix all these edited names. Till then we have to keep it to make tests happy

src/xdocs/google_style.xml
```diff
@@ -335,7 +335,7 @@
...
                 <td>
                   <a href="styleguides/google-java-style-20220203/javaguide.html#s3.2-package-statement">
-                    3.2 Package Statement</a>
+                    3.2 Package statement</a>
```
Zopsss, line 338: Reverted all section names I modified in previous PRs

src/xdocs/google_style.xml
Zopsss: Reverted the section name to the original

Zopsss:
@rdiachenko @romani conflicts resolved

rdiachenko (approved):
lgtm

Issue #8905, opened 2020-10-20
Add SARIF as a supported output format


adverma98:
**Is your feature request related to a problem? Please describe.**
No, it's not related to any existing problems.

**Describe the solution you'd like**

Github's new [code scanning](https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/about-code-scanning) product allows for teams to integrate third party static analysis and linting tools.  They use the [SARIF format](https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/uploading-a-sarif-file-to-github) for uploading and reporting the output of these analyses/linters.

It would be very useful if checkstyle was able to output SARIF results directly along with the already existing formats.

**Additional context**
Some resources that might be useful:
- [SARIF description](https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/sarif-support-for-code-scanning) 
- [SARIF website](https://sarifweb.azurewebsites.net/)
- [SARIF tutorials](https://github.com/microsoft/sarif-tutorials)
- [Specification](https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html)

-----

To avoid problems in schema validation for urls to files, user should use system property https://checkstyle.org/config_system_properties.html#Base_directory_support
`java -jar target/checkstyle-8.42-SNAPSHOT-all.jar -f sarif -c src/main/resources/sun_checks.xml src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractCheck.java` and setting `basedir` using `<property name="basedir" value="${basedir}"/>` in config.

chao2zhang:
I could pick this up if no one else claimed it.

michaelcfanning:
It would be great to see this work get done. We're glad to help if we can (we're MS engineer w/a lot of experience in SARIF). Don't forget to run the [SARIF Validator](https://sarifweb.azurewebsites.net/Validation) and to enable the `GitHub ingestion rules` in particular (which cover some corner cases in the format that provide for best integration with GitHub Advanced Security). I also wouldn't be surprised if those validations are a bit out of sync with current GHAS (which has evolved in recent months). If our team can fix up any gaps/inconsistencies you observe, glad to do that.

@eddynaka @yongyan-gh

romani:
Issue is approved, just send a PR.

strkkk:
[...]

PR #15097, opened 2024-06-22
Issue #14137: Enable `IdentityConversion` check

rickie:
Issue #14137.

This enables the `IdentityConversion` check, see docs here: https://error-prone.picnic.tech/bugpatterns/IdentityConversion/.

This is the last PR of #14137! 

https://github.com/checkstyle/checkstyle/blob/a2ba2b27bc7d59f9619d96663881f000260a913f/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java#L140-L148

### DIFF
--- a/config/error-prone-suppressions/compile-phase-suppressions.xml
+++ b/config/error-prone-suppressions/compile-phase-suppressions.xml
@@ -13,4 +13,11 @@
     <description>Avoid `Collectors.to{List,Map,Set}` in favor of collectors that emphasize (im)mutability</description>
     <lineContent>}).collect(Collectors.toSet()));</lineContent>
   </error>
+  <!-- This identity conversion prevents potential `NullPointerException`s from being thrown. -->
+  <error>
+    <sourceFile>SarifLogger.java</sourceFile>
+    <bugPattern>IdentityConversion</bugPattern>
+    <description>This method invocation appears redundant; remove it or suppress this warning and add a comment explaining its purpose</description>
+    <lineContent>.replace(VERSION_PLACEHOLDER, String.valueOf(version))</lineContent>
+  </error>
 </suppressedErrors>
--- a/pom.xml
+++ b/pom.xml
@@ -248,6 +248,7 @@
       -Xep:EmptyMethod:ERROR
       -Xep:ExplicitEnumOrdering:ERROR
       -Xep:FormatStringConcatenation:ERROR
+      -Xep:IdentityConversion:ERROR
       -Xep:ImmutablesSortedSetComparator:ERROR
       -Xep:IsInstanceLambdaUsage:ERROR
       -Xep:MockitoMockClassReference:ERROR


config/error-prone-suppressions/compile-phase-suppressions.xml
```diff
@@ -13,4 +13,10 @@
     <description>Avoid `Collectors.to{List,Map,Set}` in favor of collectors that emphasize (im)mutability</description>
     <lineContent>}).collect(Collectors.toSet()));</lineContent>
   </error>
+  <error>
+    <sourceFile>SarifLogger.java</sourceFile>
+    <bugPattern>IdentityConversion</bugPattern>
+    <description>Avoid or clarify identity conversions</description>
+    <lineContent>.replace(VERSION_PLACEHOLDER, String.valueOf(version))</lineContent>
```
romani, line 20: can you share reason of suppression in comment above?
We always do this to let next reader of this code understand our position or reasoning of ignoring best practices.

romani (changes requested):
CI is red.

item:

config/error-prone-suppressions/compile-phase-suppressions.xml
rickie: It's not an ideal situation but the `version` here can be `null` while the replace will throw an NPE if it receives `null. Let me know if I should update the comment.

rickie (commented):
Updated the commit :).

config/error-prone-suppressions/compile-phase-suppressions.xml
romani: in change of code:
```
✔ ~/java/github/romani/checkstyle [master|✚ 1⚑ 2] 
06:38 $ git diff
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java b/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
index 82e848b..c8cdecb 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
:...skipping...
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java b/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
index 82e848b..c8cdecb 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
@@ -31,6 +31,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 
+import org.apache.commons.lang3.StringUtils;
+
 import com.puppycrawl.tools.checkstyle.api.AuditEvent;
 import com.puppycrawl.tools.checkstyle.api.AuditListener;
 import com.puppycrawl.tools.checkstyle.api.SeverityLevel;
@@ -140,7 +142,7 @@ public class SarifLogger extends AbstractAutomaticBean implements AuditListener
     public void auditFinished(AuditEvent event) {
         final String version = SarifLogger.class.getPackage().getImplementationVersion();
         final String rendered = report
-            .replace(VERSION_PLACEHOLDER, String.valueOf(version))
+            .replace(VERSION_PLACEHOLDER, StringUtils.defaultString(version))
             .replace(RESULTS_PLACEHOLDER, String.join(",\n", results));
         writer.print(rendered);
         if (closeStream) {

```
test failures:
```
diff (-expected +actual):
    @@ -13,8 +13,8 @@
               "organization": "Checkstyle",
               "rules": [
               ],
    -          "semanticVersion": "null",
    -          "version": "null"
    +          "semanticVersion": "",
    +          "version": ""
```

romani:
@rickie , please resolve conflict.
I am ok with update.

config/error-prone-suppressions/compile-phase-suppressions.xml
romani: I do not know if such fields ok to me empty if not properly defined.
https://github.com/checkstyle/checkstyle/blob/a2ba2b27bc7d59f9619d96663881f000260a913f/src/test/resources/com/puppycrawl/tools/checkstyle/sariflogger/ExpectedSarifLoggerLineOnly.sarif#L1-L20

I am ok to keep suppression.

romani (approved):
Ok to merge

nrmancuso (approved):

rickie:
I think it's ready to merge, right 😄?

romani:
@rnveach , please finalize review

rnveach (approved):
I am ok to continue how we are doing, but...

> I do not know if such fields ok to me empty if not properly defined.

I believe we should decide how we are going to control our own output. I don't know SARIF all that well, but I would think this is our output we can control what goes out into it just like XML or such.

romani:
I asked a question from author of this format type https://github.com/checkstyle/checkstyle/issues/8905#issuecomment-2191672813

Issue #20, opened 2013-10-16
FinalLocalVariable does not check for loop variable 


romani:
assigned to Bhavik Patel  by mail-list at 18 Jan 2015.

Created: 2013-02-12
Creator: Nathan 
SF: #sf695

Checkstyle version: 5.6
Check: http://checkstyle.sourceforge.net/config_coding.html#FinalLocalVariable

With FinalLocalVariable enabled, checkstyle does not complain about following code:
```
for (String s : new String[] {"A", "B", "C"}) {
System.out.println(s);
}
```
Should require:
```
for (final String s : new String[] {"A", "B", "C"}) {
System.out.println(s);
}
```

Bhavik3:
check-style has a bid code. it will take time to go through all of them. so can you give me portion of it related to this issue? @romani

romani:
target of fix - https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java

how to : http://checkstyle.sourceforge.net/writingchecks.html
some first steps guidance - https://github.com/sevntu-checkstyle/sevntu.checkstyle/wiki

Bhavik3:
Hi,

in visitToken Method, for the case of token being VARIABLE_DEF, we are ignoring tokens having FOR_EACH_CLAUSE parent token>>

case TokenTypes.VARIABLE_DEF:
            if ((aAST.getParent().getType() != TokenTypes.OBJBLOCK)
                && (aAST.getParent().getType() != TokenTypes.FOR_EACH_CLAUSE))
            {
                insertVariable(aAST);
            }
            break;

I can change it to>>

case TokenTypes.VARIABLE_DEF:
            if (aAST.getParent().getType() != TokenTypes.OBJBLOCK)
            {
                insertVariable(aAST);
            }
            break;

i have not study whole code yet. so it may happen that i am not considering some point. so I think it solves the issue. @romani

Bhavik3:
class Blah
{
    static
    {
        for(int a : some.getInts())
        {

        }
    }
}

This is the portion of testcase given for FinalLocalVariable. but i don't understand what is some.getInts() actually. if "some.getInts()" method is returning array of ints, then there should not be problem.
@romani

romani:
[...]

PR #15113, opened 2024-06-24
Issue #14979: New Check MissingNullCaseInSwitch

### LINKED ISSUE #14979: Add Check Support for Java 21 Pattern Matching for Switch Syntax: New Check MissingNullCaseInSwitch
child of #14961 :

Switch statements and expressions in Java traditionally throw a `NullPointerException` if the selector expression evaluates to null. This necessitates an external null check before entering the switch. However, when Java introduces the new pattern matching for switch syntax, it becomes possible to integrate a `null` check within the switch statement by allowing a new null case label. This enhancement simplifies the code and eliminates the risk of `NullPointerException` by ensuring the null case is explicitly handled within the switch statement itself.

We should implement a new check named `MissingNullCaseInSwitch`. This check will enforce that any switch statement with a **reference type selector** must include a case null label.

We have a similar check ([EqualsAvoidNull](https://checkstyle.org/checks/coding/equalsavoidnull.html#EqualsAvoidNull)) to avoid a potential NullPointerException when using a string variable in LHS with `equals()`. We didn't make anything to allow suppression here even though it may be guaranteed that the LHS is not null.

---
**Config**
```XML
$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
    <module name="TreeWalker">
        <module name="MissingNullCaseInSwitch"/>
    </module>
</module>
```
---
**Examples**

```Java
public class Test {
    public enum C {A}

    public void testPatternMatching(Point p) {
        switch (p) {  // violation
            case Point(int x, int y) -> System.out.println("error");
        }
        // after fix
        switch (p) {
            case Point(int x, int y) -> System.out.println("error");
            case null -> System.out.println("default");
        }
    }
    public void testStrings(String s) {
        switch (s) {  // violation
            case "a" -> System.out.println("a");
        }
    }
    
    public void testPrimitives(int i, char ch) {
        switch (i) {  // ok, NPE can't happen
            case 0 -> System.out.println("0");
        }
        switch (ch) {  // ok, NPE can't happen
            case 'a' -> System.out.println("a");
        }
    }
}

```
---

**Expected Output**
```powershell
Starting audit...
[ERROR] D:\Test.java:5:9: Switch using reference types should have a null case. [MissingNullCaseInSwitch]
[ERROR] D:\Test.java:16:9: Switch using reference types should have a null case. [MissingNullCaseInSwitch]
[ERROR] D:\Test.java:22:9: Switch using reference types should have a null case. [MissingNullCaseInSwitch]
Audit done.
```

---
**Limitations**

switch can handle both primitive types and their corresponding boxed types. However, there is a  limitation when dealing with boxed types in switch. because we need to be aware of how to distinguishes between primitives and their boxed types.

**Example**  

```Java
	public void testBoxed(Integer i, char ch) {
		switch (i) {    
		        case 0 -> System.out.println("0");
                       case null -> System.out.println("1");
		}
	}
```
we can add case `null` in the above example because the switch selector is of a reference type (Integer). However, the check can't violate this example becuase we have no way to distinguitsh 0 (int) and their boxed type 0 (Integer)

---
 This check does not support enums; syntactically, these are no different from other constants.
**Example**  

```Java
     // Checkstyle is not type-aware. The
   // value of 'A' could be an enum constant, a primitive, or a reference type.
  void testEnums(E obj) {
        switch (obj) {  // Enum is a reference type it can be null
            case A -> {}
            case B -> {} 
            default -> {}
        }
    }
        switch (ast.getType()) { 
            case TokenTypes.METHOD_CALL:{} // this is just an integer can't be null
        }
    }
}
enum E {A, B, C}
class TokenTypes {
    public static final int METHOD_CALL = 2;
}
class DetailAST {
    public int getType() { return 0; }
}
```

mahfouz72:
closes #14979 : 

New module config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/d1ea609b49955fc8c28ba1236625f2032edc7621/check.xml

---
**Site** 
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c30e2a6_2024151038/index.html

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c30e2a6_2024151038/checks.html#

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c30e2a6_2024151038/checks/coding/index.html#

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c30e2a6_2024151038/checks/coding/missingnullcaseinswitch.html#

---

**Report:** https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/b2743e3_2024192510/reports/diff/index.html

---

Contribution repo: https://github.com/checkstyle/contribution/pull/875


mahfouz72:
Github, generate site

mahfouz72:
Github, generate site

mahfouz72:
Github, generate site

mahfouz72:
@romani Why security test fails?

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

nrmancuso:
Please create a PR in contribution: https://checkstyle.semaphoreci.com/jobs/60aa228f-6ae8-4ca0-b195-31bdce26d07d

```
Please create PR to repository https://github.com/checkstyle/contribution00:01
and add your new Check 00:01
   to file checkstyle-tester/checks-nonjavadoc-error.xml00:01
or to file checkstyle-tester/checks-only-javadoc-error.xml00:01
Place the contribution repository PR link in the description of this PR.00:01
PR for contribution repository will be merged right after this PR.
```

mahfouz72:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,195 @@
...
+     */
+    private static boolean hasNullCaseLabel(DetailAST detailAST) {
+        return TokenUtil.findFirstTokenByPredicate(detailAST, ast -> {
+            final DetailAST caseLabel = ast.findFirstToken(TokenTypes.LITERAL_CASE);
+            DetailAST expr = null;
+            if (caseLabel != null) {
+                expr = caseLabel.findFirstToken(TokenTypes.EXPR);
+            }
+            return expr != null && expr.findFirstToken(TokenTypes.LITERAL_NULL) != null;
+        }).isPresent();
```
nrmancuso: It is not clear to me why we need to lean so heavily on findFirstTokenByPredicate in this check. Can we just make some collection of  LITERAL_CASE DetailASTs, then simplify all of these predicates to only deal with one LITERAL_CASE at a time? Then in visitToken, we just do something like this:
```
        final boolean hasNullCase =  cases.stream()
          .filter(MissingNullCaseInSwitchCheck::hasLiteralNull)
          .isPresent();
```

Then, this method looks like:
```
return Optional.ofNullable(caseLabel.findFirsttoken(TokenTypes.EXPR)
    .map(expr -> expr.findFirstToken(TokenTypes.LITERAL_NULL)
    .isPresent();


```

Think about how we will extend this check in the future without refactoring. If we start with a collection of cases, and build check logic around this, I think it will be more maintainable going forward.

nrmancuso: This all all pseudocode obvs.

src/xdocs/checks/coding/missingnullcaseinswitch.xml
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+  void testPrimitives(int obj) {
+    switch (obj) {   // ok, this is a primitive type can't be null
+      case 1 : {}
+    }
+  }
+
+  void testBoxedType(Integer obj) {
+    // ok, we can't tell if '1' is of a reference or primitive type
```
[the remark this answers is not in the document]
mahfouz72: from another class outside this file or do you mean so inner class?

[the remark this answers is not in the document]
nrmancuso: Some class that exists in core Java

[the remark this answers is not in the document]
mahfouz72: what do yo think about this?
```Java
public class Example1 {

  void testString(String obj) {
    // violation below, 'switch using reference types should have a null case.'
    switch (obj) {
      case "something" : {}
    }
    switch (obj) {
      case null : {}
      case "something" : {}
      default: {}
    }
  }

  void testPatterns(Object obj) {
    // violation below, 'switch using reference types should have a null case.'
    switch (obj) {
      case Integer i : {} break;
      default : {}
    }
  }

  void testPrimitives(int obj) {
    switch (obj) {   // ok, this is a primitive type can't be null
      case 1 : {}
    }
  }

  void testBoxedType() {
    switch (getInteger()) { // ok, we can't tell if '1' is an Integer type
      case 1 : {}
    }
  }

  void testIdent() {
    // ok, we can't tell if Ident 'ONE' is of a reference type,
    // i.e. Enum or primitive type i.e. local f
    switch (getEnum()) {
      case E.ONE: {}
    }
  }

  enum E { ONE }
  public static final int ONE = 5;
  public static Integer getInteger() { return 1; }
  public static E getEnum() { return E.ONE; }
}
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: > It is not clear to me why we need to lean so heavily on findFirstTokenByPredicate in this check.

no specific reason I just liked it and found it handy.

> Can we just make some collection of LITERAL_CASE DetailASTs, then simplify all of these predicates to only deal with one LITERAL_CASE at a time?

Good, but I think this will make the check Filestateful do we encourage this? I think somewhere else we stated that the check should always be stateless if possible

mahfouz72: >  It is not clear to me why we need to lean so heavily on findFirstTokenByPredicate in this check.

 No specific reason I just liked it and found it handy and also inspired by `MissingDefaultCheck`

> Can we just make some collection of LITERAL_CASE DetailASTs, then simplify all of these predicates to only deal with one LITERAL_CASE at a time?

Good, but I think this will make the check Filestateful do we encourage this? I think somewhere else we stated that the check should always be stateless if possible

nrmancuso: No, we don’t need to make this check stateful. Right now, we repeatedly iterate over all children of a given switch via findFirstTokenByPredicate. I am proposing we just make a collection instead.

mahfouz72: this `LITERAL_CASE` collection will be a class field, right? Isn't this a state

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,195 @@
...
+import com.puppycrawl.tools.checkstyle.StatelessCheck;
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
+
+/**
+ * <p>
+ * Checks that switch statement or expression that uses reference type in case labels has
+ * a {@code null} case label.
```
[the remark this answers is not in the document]
mahfouz72: done just corrected it's selector to its selector

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,195 @@
...
+ * </p>
+ * <p>
+ * Rationale: switch statements and expressions in Java throw a
+ * {@code NullPointerException} if the selector expression evaluates to {@code null}.
+ * With the introduction of patterns in case label,
+ * It is now possible to integrate a null check within the switch,
```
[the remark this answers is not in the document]
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,195 @@
...
+ * <li>
+ * This check does not differentiate between primitive types and
+ * their boxed counterparts in case labels.
+ * Therefore, any case label with an integer will not violate this check,
+ * even if it is a reference type {@code Integer}.
+ * </li>
+ * <li>
+ * This check will not violate case labels with idents even if
+ * this is an identifier of Enums. We can't tell if the identifier
+ * is an Enum or just a normal final variable.
```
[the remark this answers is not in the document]
mahfouz72: done different than is autocorrect by IntelliJ to different from

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,195 @@
...
+ * </p>
+ * <p>
+ * Limitations due to not being type-aware:
+ * </p>
+ * <ul>
+ * <li>
+ * This check does not differentiate between primitive types and
+ * their boxed counterparts in case labels.
+ * Therefore, any case label with an integer will not violate this check,
+ * even if it is a reference type {@code Integer}.
```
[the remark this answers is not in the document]
mahfouz72: done

src/xdocs/checks/coding/missingnullcaseinswitch.xml
```diff
@@ -0,0 +1,148 @@
...
+    // violation below, 'switch using reference types should have a null case.'
+    switch (obj) {
+      case &quot;something&quot; : {}
+    }
+    switch (obj) {
+      case null : {}
+      case &quot;something&quot; : {}
+      default : {}
+    }
+  }
```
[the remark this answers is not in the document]
mahfouz72: done

src/xdocs/checks/coding/missingnullcaseinswitch.xml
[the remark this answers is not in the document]
mahfouz72: please review the example again if it is good or still needs improvements I added the methods in the same class

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
nrmancuso: >class field, right

Nope. Just make a List/Set/whatever makes sense in visitToken.

nrmancuso: See https://github.com/nrmancuso/checkstyle/blob/b8615479028fb40523376d6093db01df9db6876a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ConstructorsDeclarationGroupingCheck.java#L92 as an example.

mahfouz72:
Github, generate report

mahfouz72:
Github, generate site

src/xdocs/checks/coding/missingnullcaseinswitch.xml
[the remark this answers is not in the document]
nrmancuso: We should not be using the term "Ident" anywhere, unless we are referring to the actual TokenType, in which case we should make that clear. However, we should not be referencing TokenTypes at all in this example, since this is an implementation detail.

In both of your examples above, we _could_ tell the type of the given variable if we wanted to, since everything is in the example. We need to use methods/constants from core java to drive the point home that Checkstyle has no idea about the type of the value resulting from these expressions:

```
    // This example is ok, because Checkstyle is not type-aware. The
    // value of 't' could be an enum constant, a primitive, or a reference
    // type.
    void m1() {
        var t = TimeUnit.DAYS;
        switch (t) {
            ...
        }
    }
    
    // This is example is also ok, because we do not know the return type
    // of a given method.
    void m2() {
        var t = List.of().size();
        switch (t) {
            ...
        }
    }
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,210 @@
...
+     */
+    private static List<DetailAST> getAllCaseLabels(DetailAST detailAST) {
+        final List<DetailAST> caseLabels = new ArrayList<>();
+        DetailAST ast = detailAST.getFirstChild();
+        while (ast != null) {
+            // case group token may have several LITERAL_CASE tokens
+            TokenUtil.forEachChild(ast, TokenTypes.LITERAL_CASE, caseLabels::add);
+            ast = ast.getNextSibling();
+        }
+        return caseLabels;
```
nrmancuso, line 132: Please return an immutable list.

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
```diff
@@ -35,6 +35,7 @@ missing.ctor=Class should define a constructor.
 missing.package.declaration=Missing package declaration.
 missing.super.call=Method ''{0}'' should call ''super.{0}''.
 missing.switch.default=switch without \"default\" clause.
+missing.switch.nullcase=switch using reference types should have a null case.
```
nrmancuso, line 38:
```suggestion
missing.switch.nullcase=Switch using reference types should have a null case.
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,210 @@
...
+    /**
+     * Checks if the case label has a string literal.
+     *
+     * @param caseLabel the AST node representing {@code LITERAL_CASE}
+     * @return true if the case label has a string literal
+     */
+    private static boolean hasStringLiteral(DetailAST caseLabel) {
+        return Optional.ofNullable(caseLabel.findFirstToken(TokenTypes.EXPR))
+                .map(exp -> exp.findFirstToken(TokenTypes.STRING_LITERAL))
+                .isPresent();
```
nrmancuso, line 183: We can suppress these checker violations. I can see a bit of value in them sometimes, but in cases like this, we would double the lines of code needed to do the same thing and decrease readability.

nrmancuso (changes requested):
Last from me, also see my comment above about the documentation example:

src/xdocs/checks/coding/missingnullcaseinswitch.xml
```diff
@@ -0,0 +1,149 @@
...
+  <head>
+    <title>MissingNullCaseInSwitch</title>
+  </head>
+  <body>
+    <section name="MissingNullCaseInSwitch">
+      <p>Since Checkstyle 10.18.0</p>
+      <subsection name="Description" id="Description">
+        <p>
+          Checks that a given switch statement or expression that use a reference type
+          in its selector expression has a {@code null} case label.
```
nrmancuso, line 14: We need to use xml tags here, this does not render correctly.

src/xdocs/checks/coding/missingnullcaseinswitch.xml
```diff
@@ -0,0 +1,149 @@
...
+        <p>
+         Rationale: switch statements and expressions in Java throw a
+          <code>NullPointerException</code> if the selector expression evaluates
+          to <code>null</code>. As of Java 21,
+          it is now possible to integrate a null check within the switch,
+          eliminating the risk of <code>NullPointerException</code> and simplifies the code
+          as there is no need for an external null check before entering the switch.
+        </p>
+        <p>
+          Limitations due to not being type-aware:
```
nrmancuso, line 25:
```suggestion
          Limitations due to Checkstyle not being type-aware:
```

nrmancuso (changes requested):
Few more after rereading the docs:

src/xdocs/checks/coding/missingnullcaseinswitch.xml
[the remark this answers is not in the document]
mahfouz72: > In both of your examples above, we _could_ tell the type of the given variable if we wanted to, since everything is in the example. We need to use methods/constants from core java to drive the point home that Checkstyle has no idea about the type of the value resulting from these expressions:

Yes, understood thank you

src/xdocs/checks/coding/missingnullcaseinswitch.xml
```diff
@@ -0,0 +1,149 @@
...
+
+  void testPatterns(Object obj) {
+    // violation below, 'switch using reference types should have a null case.'
+    switch (obj) {
+      case Integer i : {} break;
+      default : {}
+    }
+  }
+
+  void testPrimitives(int obj) {
```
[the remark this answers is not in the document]
mahfouz72: done

src/xdocs/checks/coding/missingnullcaseinswitch.xml
```diff
@@ -0,0 +1,149 @@
...
+        </p>
+        <p>
+          Limitations due to not being type-aware:
+        </p>
+        <ul>
+          <li>
+             This check does not differentiate between primitive types and
+             their boxed counterparts in case labels.
+             Therefore, any case label with an integer will not violate this check,
+             even if it is a reference type, i.e. <code>Integer</code>.
```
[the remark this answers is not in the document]
mahfouz72: done

src/xdocs/checks/coding/missingnullcaseinswitch.xml
mahfouz72: done

src/xdocs/checks/coding/missingnullcaseinswitch.xml
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: done

mahfouz72:
Github, generate site

mahfouz72:
Github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72:
```
+    <sourceFile>MissingNullCaseInSwitchCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.MissingNullCaseInSwitchCheck</mutatedClass>
+    <mutatedMethod>getAllCaseLabels</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator</mutator>
+    <description>replaced call to java/util/Collections::unmodifiableList with argument</description>
+    <lineContent>return Collections.unmodifiableList(caseLabels);</lineContent>
+  </mutation>
```

I thought this was an issue with Pitest and solved by https://github.com/checkstyle/checkstyle/pull/14484 ?

mahfouz72: seems like this is not activated for the pitest-coding-2 profile yet. we can suppress for now ?

nrmancuso: I think the suppression would be more LOC than just adding the feature to this profile, right?

mahfouz72: right, done

nrmancuso (approved):
Very good. This is a great foundation to build on, and a very practical check that will help people like me that forget that switches can throw a NPE when the selector expression is null!

mahfouz72:
@rnveach @romani Please finalize the review

config/checkstyle-checks.xml
```diff
@@ -523,6 +523,14 @@
       -->
       <property name="severity" value="ignore"/>
     </module>
+    <module name="MissingNullCaseInSwitch">
+      <!--
+        We will not use this check because case null
+        is only supported by newer versions of Java (21).
+        Using a 'case null' will not compile in older Java code.
```
rnveach, line 530: Examples in issue show violations only show up when using `->` in the case. Aren't we good to enable this since we only use `:` cases currently? What earliest Java version is `->` cases supported?

src/xdocs/checks.xml
```diff
@@ -791,6 +791,17 @@
               when the <code>@inheritDoc</code> javadoc tag is present.
             </td>
           </tr>
+          <tr>
+            <td>
+              <a href="missingnullcaseinswitch.html#MissingNullCaseInSwitch">
```
rnveach, line 796: This URL is wrong. This is why it appears in linkcheck. The suppression added is also wrong.

Make sure you verify the site and links you are adding.

src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/missingnullcaseinswitch/InputXpathMissingNullCaseInSwitchNested.java
```diff
@@ -0,0 +1,22 @@
...
+    void test(Object obj) {
+       int x =  switch (obj) {
+            case Rectangle(ColoredPoint _, ColoredPoint _) -> 1;
+            case ColoredPoint(int _, int _, String _) -> 2;
+            case null -> 3;
+            default -> {
+                int y = switch (obj) { // warn
+                    case Rectangle(ColoredPoint _, ColoredPoint _) -> 1;
+                    case ColoredPoint(int _, int _, String _) -> 2;
+                    default -> 4;
```
rnveach, line 14: Doesn't the default case cover the null case too? This code has no issues of a NPE because of the default right?

src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/missingnullcaseinswitch/InputXpathMissingNullCaseInSwitchSimple.java
```diff
@@ -0,0 +1,16 @@
+
```
rnveach, line 1: Remove empty line.

src/xdocs/checks/coding/missingnullcaseinswitch.xml
```diff
@@ -0,0 +1,142 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+  <head>
+    <title>MissingNullCaseInSwitch</title>
+  </head>
+  <body>
+    <section name="MissingNullCaseInSwitch">
+      <p>Since Checkstyle 10.18.0</p>
```
rnveach, line 10: Release is tomorrow. You might as well bump this just to be on the safe side.

src/xdocs/checks/coding/missingnullcaseinswitch.xml
```diff
@@ -0,0 +1,142 @@
...
+      <subsection name="Description" id="Description">
+        <p>
+          Checks that a given switch statement or expression that use a reference type
```
rnveach, line 13: I am slightly confused as the check is named `...InSwitch` not `...InSwitchOrExpression...`.

You are referring to switch expressions? Or is how you wrote it the way it is worded in the JLS? (Link?)

src/xdocs/checks/coding/missingnullcaseinswitch.xml
```diff
@@ -0,0 +1,142 @@
...
+          eliminating the risk of <code>NullPointerException</code> and simplifies the code
+          as there is no need for an external null check before entering the switch.
+        </p>
+        <p>
+          Limitations due to Checkstyle not being type-aware:
+        </p>
+        <ul>
+          <li>
+            This check does not support enums; syntactically,
+            these are no different from other constants.
```
rnveach, line 30: Missing
https://github.com/checkstyle/checkstyle/issues/14979
> switch can handle both primitive types and their corresponding boxed types. However, there is a limitation when dealing ....

Or if something has changed, let me know.

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/Example1.java
```diff
@@ -0,0 +1,60 @@
...
+    switch (i) {   // ok, this is a primitive type can't be null
+      case 1 : {}
+    }
+  }
+
+  // This example is ok, because Checkstyle is not type-aware. The
+  // value of 't' could be an enum constant, a primitive, or a reference type.
+  void testEnum() {
+    var t = TimeUnit.DAYS;
+    switch (t) {
```
rnveach, line 46: Issue needs to be updated to reflect state of this check.

https://github.com/checkstyle/checkstyle/issues/14979
> public void testEnums(C c) {
switch (c) {  // violation

It seems we aren't validating these anymore.

rnveach (changes requested):

config/checkstyle-checks.xml
rnveach: I see regression and tests have `:`, so it was slightly misleading to me from the issue. I will resolve this.

src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/missingnullcaseinswitch/InputXpathMissingNullCaseInSwitchNested.java
rnveach: I tested this on my local, and default does not cover the null case. I find this a bit odd TBH.

Please add this to the documentation.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,205 @@
...
+     */
+    private static List<DetailAST> getAllCaseLabels(DetailAST detailAST) {
+        final List<DetailAST> caseLabels = new ArrayList<>();
+        DetailAST ast = detailAST.getFirstChild();
+        while (ast != null) {
+            // case group token may have several LITERAL_CASE tokens
+            TokenUtil.forEachChild(ast, TokenTypes.LITERAL_CASE, caseLabels::add);
+            ast = ast.getNextSibling();
+        }
+        return Collections.unmodifiableList(caseLabels);
```
rnveach, line 127: Did something require this to unmodifable? It seems odd to do this for a list that will never leave this class.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,205 @@
...
+     */
+    private static boolean hasLiteralNull(DetailAST caseLabel) {
+        return Optional.ofNullable(caseLabel.findFirstToken(TokenTypes.EXPR))
+                .map(exp -> exp.findFirstToken(TokenTypes.LITERAL_NULL))
+                .isPresent();
+    }
+
+    /**
+     * Checks if the given case AST node has a pattern variable declaration label.
+     * or record pattern definition label.
```
rnveach, line 144: You have 2 period for the same sentence. Either remove one or convert them to their own, separate sentences.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,205 @@
...
+     * In this case, if the switch selector evaluates to null,
+     * a {@code NullPointerException} will be thrown.
```
rnveach, line 146: This sentence doesn't seem to have anything to do with the method. This seems to be more about the check overall. The method javadoc should be specific to what the implementation of the method is doing to convey this to the user. We are just checking if this has a pattern case. The reason for **why** we are checking it is more for the caller and it's logic.

I would remove it or move it somewhere more pertinent.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,205 @@
...
+        return Collections.unmodifiableList(caseLabels);
+    }
+
+    /**
+     * Checks if the given case AST node has a null label.
+     *
+     * @param caseLabel the AST node representing {@code LITERAL_CASE}
+     * @return true if the case has {@code null} label, false otherwise
+     */
+    private static boolean hasLiteralNull(DetailAST caseLabel) {
```
rnveach, line 136: `caseLabel` -> `caseAST`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,205 @@
...
+    /**
+     * Checks if the given case AST node has a pattern variable declaration label.
+     * or record pattern definition label.
+     * In this case, if the switch selector evaluates to null,
+     * a {@code NullPointerException} will be thrown.
+     *
+     * @param caseLabel the AST node representing {@code LITERAL_CASE}
+     * @return true if case has a pattern in its label
+     */
+    private static boolean hasPatternCaseLabel(DetailAST caseLabel) {
```
rnveach, line 151: `caseLabel` -> `caseAST`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    /**
+     * Gets all case labels in the given switch AST node.
+     *
+     * @param detailAST the AST node representing {@code LITERAL_SWITCH}
+     * @return a list of all case labels in the switch
+     */
+    private static List<DetailAST> getAllCaseLabels(DetailAST detailAST) {
```
rnveach, line 119: `detailAST` -> `switchAST`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * Checks if the given case contains a string in its label.
+     * In this case, if the switch selector evaluates to null,
+     * a {@code NullPointerException} will be thrown.
+     *
+     * @param caseLabel the AST node representing {@code LITERAL_CASE}
+     * @return true if switch block contains a string case label
+     */
+    private static boolean hasStringCaseLabel(DetailAST caseLabel) {
```
rnveach, line 165: Same as others and for ones further below too.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,205 @@
...
+    private static boolean hasPatternCaseLabel(DetailAST caseLabel) {
+        return caseLabel.findFirstToken(TokenTypes.RECORD_PATTERN_DEF) != null
+               || caseLabel.findFirstToken(TokenTypes.PATTERN_VARIABLE_DEF) != null
+               || caseLabel.findFirstToken(TokenTypes.PATTERN_DEF) != null;
+    }
+
+    /**
+     * Checks if the given case contains a string in its label.
+     * In this case, if the switch selector evaluates to null,
+     * a {@code NullPointerException} will be thrown.
```
rnveach, line 160: Same as previous javadoc.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,205 @@
...
+        DetailAST ast = caseLabel.getFirstChild().findFirstToken(TokenTypes.PLUS);
+        boolean hasStringInConcatenatedExpression = false;
+        while (ast != null) {
+            if (ast.findFirstToken(TokenTypes.STRING_LITERAL) != null) {
+                hasStringInConcatenatedExpression = true;
+                break;
+            }
+            if (ast.getFirstChild() != null
+                    && ast.getFirstChild().getType() == TokenTypes.PLUS) {
+                ast = ast.getFirstChild();
```
rnveach, line 197: Doesn't this seem like a concern that we are going deeper into the tree, without ever returning?

For example: a + b + "c"
````
        |      |  |  |--LITERAL_CASE -> case [4:2]
        |      |  |  |  |--EXPR -> EXPR [4:10]
        |      |  |  |  |  `--PLUS -> + [4:10]
        |      |  |  |  |      |--PLUS -> + [4:8]
        |      |  |  |  |      |  |--IDENT -> a [4:7]
        |      |  |  |  |      |  `--IDENT -> b [4:9]
        |      |  |  |  |      `--STRING_LITERAL -> "c" [4:11]
````

You will miss out on the string literal, because you kept going deep into the plus which just has 2 idents.

File: src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheckTest.java [hunk 32]
```diff
@@ -0,0 +1,116 @@
...
+        final MissingNullCaseInSwitchCheck check = new MissingNullCaseInSwitchCheck();
+        assertWithMessage("Acceptable tokens should not be null")
+            .that(check.getAcceptableTokens())
+            .isNotNull();
+        assertWithMessage("Default tokens should not be null")
+            .that(check.getDefaultTokens())
+            .isNotNull();
+        assertWithMessage("Required tokens should not be null")
+            .that(check.getRequiredTokens())
+            .isNotNull();
```
rnveach, line 49: Why aren't we checking exact matches?

rnveach (changes requested):

src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/missingnullcaseinswitch/InputXpathMissingNullCaseInSwitchNested.java
rnveach: It seems if you don't add a `null` case, Java adds a `Objects.requireNonNull` before the switch. With a `null` case, that code isn't inserted. In my test bed, a `null` case was represented as `-1`.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
rnveach: Cases have to be constants, so a variable shouldn't be possible. However `case 1 + 2 + "Hello" + "lol":` was acceptable for me.

src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/missingnullcaseinswitch/InputXpathMissingNullCaseInSwitchNested.java
mahfouz72: > Please add this to the documentation.

can you clarify what to add exactly? I think It is clear that `case null` is required to eliminate the risk of NPE, not the default.  as in this example  in the doc
```Java
 void testPatterns(Object obj) {
    // violation below, 'Switch using reference types should have a null case.'
    switch (obj) {
      case Integer i : {} break;
      default : {}
    }
  }
```

src/xdocs/checks/coding/missingnullcaseinswitch.xml
mahfouz72: Yes, I was referring to switch expressions. I didn't take this wording from JLS. I took it from existing documentation in our checks.

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c30e2a6_2024151038/checks/coding/missingswitchdefault.html

> See the [Java Language Specification](https://docs.oracle.com/javase/specs/jls/se22/html/jls-15.html#jls-15.28) for more information about switch statements and expressions. 

the word switch in the check name is a general name for any switch. I think we shouldn't name it `InSwitchOrExpression`. we don't have to be very explicit in the name. but in the description I added it to make it clear that we verify both as both are subjected to NPE if selector expression is null.

src/xdocs/checks/coding/missingnullcaseinswitch.xml
mahfouz72: https://github.com/checkstyle/checkstyle/pull/15113#discussion_r1654777167

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/Example1.java
mahfouz72: Issue is updated

src/xdocs/checks/coding/missingnullcaseinswitch.xml
nrmancuso: This is pedantic IMO, but we can do ‘switch statement or switch expression’ here. The existing wording should be understood as ‘switch (statement or expression)’

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
nrmancuso: Data should always be final/unmodifiable unless we need to modify it.

mahfouz72: Collections should be unmodifiable unless we need to modify them on purpose. This is a common best practice we use it a lot in our codebase with collections.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
nrmancuso: Yeah, let’s just remove it

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: > You will miss out on the string literal, because you kept going deep into the plus which just has 2 idents.

before going deeper I check the current plus children if there is a string we will break, if not then go deeper to the next plus. With: 

```Java
 if (ast.findFirstToken(TokenTypes.STRING_LITERAL) != null) {
                hasStringInConcatenatedExpression = true;
                break;
   }
    ...
```

> For example: a + b + "c"


I already have this example and it works correctly

```
        |      |  |  |--LITERAL_CASE -> case [4:2]
        |      |  |  |  |--EXPR -> EXPR [4:10]
        |      |  |  |  |  `--PLUS -> + [4:10]   <---- I will search for string literal before going deep
        |      |  |  |  |      |--PLUS -> + [4:8]
        |      |  |  |  |      |  |--IDENT -> a [4:7]
        |      |  |  |  |      |  `--IDENT -> b [4:9]
        |      |  |  |  |      `--STRING_LITERAL -> "c" [4:11] <--- we found it so we break the loop and return true

```
https://github.com/mahfouz72/checkstyle/blob/76bfa57748922ebb410db76347405a8978856ad5/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithStringLiterals.java#L97-L101

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheckTest.java
mahfouz72: the purpose of this method is to just kill mutation So nullity check was enought. but yes makes sense to be an exact match

mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: done for all similar below

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: done

src/xdocs/checks/coding/missingnullcaseinswitch.xml
mahfouz72: bumped to 10.19.0

src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/missingnullcaseinswitch/InputXpathMissingNullCaseInSwitchSimple.java
mahfouz72: done

src/xdocs/checks.xml
mahfouz72: done

src/xdocs/checks/coding/missingnullcaseinswitch.xml
mahfouz72: I think check-since workflow fails because it requires the version to be 10.18.0 after the release it should be green after restarting

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheckTest.java
rnveach: Not done, messages were not updated.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
rnveach: > before going deeper I check the current plus children if there is a string we will break

So then this will "break" if I hid string literal better?
`case 1 + 2 + (3 + "lol" + 4) + 5:`

mahfouz72: we have something similar.

https://github.com/checkstyle/checkstyle/blob/5d14d6b97361177a5df7763b2e6da80118871816/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithStringLiterals.java#L103-L107

but we might have false negatives when the expression becomes very complex. which contains a lot of nested parentheses with a depth of more than 1 and some integers 
something like `(1 + 2 + (1 + (2 + "a" + 3)) + 5)`  
Since we don't do a full search we might miss some string literals if it is hidden well inside a bunch of parenthes.
To have no false negatives we should do a BFS (deep scan like `branchContains()` ) and visit each token starting from the first `plus`,  like we have in our tree walker, Or is it ok to have some false negatives in this weird and uncommon syntax? 

Note that the string will be caught if it is in an expression with any number of  `+`. The false negative may happen in case of nested parens only.

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheckTest.java
mahfouz72: sorry missed the message. Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
rnveach: > is it ok to have some false negatives in this weird and uncommon syntax

IMO, if we have a false and know it is false, we should fix it. I see this more as an issue with implementation than cases being too complex. If we don't fix this here, this needs to be a new issue. Lines in our inputs would have to be associated to the issue with a "until ...".

> visit each token starting from the first plus, like we have in our tree walker

This is basically what you would need to do if you are not going to explicitly handle the tokens. We have other checks which basically implement their own "treewalker" processing. You can use `TreeWalker`'s code as a basis, but pitest will most likely require it to be trimmed back in certain places.

src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/missingnullcaseinswitch/InputXpathMissingNullCaseInSwitchNested.java
rnveach: Maybe it is just me, but it was not clear that `default` doesn't cover the `case null`. I would assume most people think `default` covers all cases not explicitly defined. Now it is like it "covers all cases not explicitly defined excluding `null` cases".

src/xdocs/checks/coding/missingnullcaseinswitch.xml
rnveach: Ok, I think my mind just went a different direction when I was reading this.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithPrimitives.java
```diff
@@ -0,0 +1,87 @@
...
+public class InputMissingNullCaseInSwitchWithPrimitives {
+    void testIntegers(int obj) {
+        switch (obj) {   // ok, case labels are primitive types
+            default -> {}
+            case 1 + 4 -> {}
+            case 2 -> {}
+        }
+        switch (obj) {
+           // case null -> {} // this will not compile
+            case 1 + 4 -> {}
```
rnveach, line 18: Line up comment with case.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithPrimitives.java
```diff
@@ -0,0 +1,87 @@
...
+
+    void testChar(char c) {
+        switch (c) {   // ok, case labels are primitive types
+            default -> {}
+            case 'a' + 'c' -> {}
+            case 'b' -> {}
+        }
+        switch (c) {
+         // case null -> {} // this will not compile
+            case 'a' -> {}
```
rnveach, line 41: Please at least be consistent with how you format things.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithPrimitives.java
```diff
@@ -0,0 +1,87 @@
...
+        final int x = 1;
+        final int y = 0;
+        switch (obj) {   // ok, case labels are idents we ignore it
+            default -> {}
+            case x -> {}
+            case y -> {}
+        }
+        switch (obj) {
+           // case null -> {}
+            case x -> {}
```
rnveach, line 61: same.

rnveach (changes requested):

src/xdocs/checks/coding/missingnullcaseinswitch.xml
rnveach: I am sorry, but due to release issues it is probably now safer to revert this to 10.18.0 . I wrongly assumed release would go without an issue.

mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: done I used treewalker implementation we now sure that the string will always be detected even if it is hidden in complex expression.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithPrimitives.java
```diff
@@ -0,0 +1,87 @@
...
+
+    void boxedTypes(Integer i) {
+        switch (i) {   // case label are of a refetence type (Integer) but we can't detect
+            default -> {}
+            case 1 -> {}
+            case 2 -> {}
+        }
+
+        switch (i) {
+            case null -> {} // this compiles
```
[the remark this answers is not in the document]
mahfouz72: it works as intended, auto unboxing should occur when we encounter a case label that has an integer and the obj is not null. this may not be very accurate but this is as far as I understand.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,201 @@
...
+
+import com.puppycrawl.tools.checkstyle.StatelessCheck;
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
+
+/**
+ * <p>
+ * Checks that a given switch statement or expression that use a reference type in its selector
```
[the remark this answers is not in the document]
mahfouz72: please see the update I added another section below to explain what we validate specifically

src/xdocs/checks/coding/missingnullcaseinswitch.xml.template
```diff
@@ -0,0 +1,93 @@
...
+    <section name="MissingNullCaseInSwitch">
+      <p>Since Checkstyle 10.19.0</p>
+      <subsection name="Description" id="Description">
+        <p>
+          Checks that a given switch statement or expression that use a reference type
+          in its selector expression has a <code>null</code> case label.
+        </p>
+        <p>
+         Rationale: switch statements and expressions in Java throw a
+          <code>NullPointerException</code> if the selector expression evaluates
```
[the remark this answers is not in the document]
mahfouz72: done

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithPrimitives.java
mahfouz72: done for all below

mahfouz72:
github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,201 @@
...
+ * <p>
+ * Limitations due to Checkstyle not being type-aware:
+ * </p>
+ * <ul>
+ * <li>
+ * This check does not support enums; syntactically, these are no different from other constants.
+ * </li>
+ * </ul>
+ * <p>
+ * See the <a href="https://docs.oracle.com/javase/specs/jls/se22/html/jls-15.html#jls-15.28">
```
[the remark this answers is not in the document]
rnveach: Not technically. I said to move it with rationale, but it was moved before rationale.

Now that there is other updates, I am good with its new basic placement, but I question why there is a gap between it and "Specifically, this check validates ...". To me they seem like they are the same thought.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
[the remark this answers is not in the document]
rnveach: Let's rearrange this a bit more if you agree.

Description
Rationale
Validates
Limitations

IMO, this is a more natural flow. I would also update limitations with a bit more verbiage before "Limitations:". Something like we know there are more cases where there are reference types, but we can't identify them because it requires type knowledge.

> This check does not support enums

Isn't there another option where we can't tell the difference between primitive and their class versions? Are there any other limitations not really being mentioned?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
```diff
@@ -0,0 +1,184 @@
...
+     * It may contain a single string literal or a string literal
+     * in a concatenated expression.
+     *
+     * @param caseAST the AST node representing {@code LITERAL_CASE}
+     * @return true if switch block contains a string case label
+     */
+    private static boolean hasStringCaseLabel(DetailAST caseAST) {
+        DetailAST curNode = caseAST;
+        boolean hasStringCaseLabel = false;
+        while (curNode != null) {
```
rnveach, line 170: Since this is a new validation, I just wanted to verify you tried to ensure by debugging that we won't go beyond the current `caseAST`?

These type of mini-walkers have this issue where they go beyond where they start and look into too much. From glancing at the code, I am not seeing anything that stops this except a `null` AST or a string literal. I would expect it to end when it comes back to `caseAST` again (IE finishes the current case or it's case group, however we work this).

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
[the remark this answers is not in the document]
rnveach: Maybe we don't need a fully documented limitations section. We define what we validate. I am not seeing how that doesn't cover the limitations fully. Why not just trim limitations to a simple sentence, no list. Let me know if you agree.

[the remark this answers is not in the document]
mahfouz72: I tend to agree the list was suitable when we don't specifically mention what we support we just say reference type then we exclude some items in the limitation now we have mentioned what we validate exactly, there is no reason for a separate limitation list I just will leave one sentence (not a list) for enums to be super clear that we don't support it

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
mahfouz72: Thanks for pointing this out. Yes I missed a condition to not go beyond the current case label expression 

```
        |       |               |   |--LITERAL_CASE -> case [99:12] <--- we should just iterate on this subtree
        |       |               |   |   `--EXPR -> EXPR [99:33]   
        |       |               |   |       `--PLUS -> + [99:33]
        |       |               |   |           |--PLUS -> + [99:27]
        |       |               |   |           |   |--PLUS -> + [99:23]
        |       |               |   |           |   |   |--PLUS -> + [99:19]
        |       |               |   |           |   |   |   |--NUM_INT -> 1 [99:17]
        |       |               |   |           |   |   |   `--NUM_INT -> 2 [99:21]
        |       |               |   |           |   |   `--NUM_INT -> 3 [99:25]
        |       |               |   |           |   `--STRING_LITERAL -> "a" [99:29]
        |       |               |   |           `--NUM_INT -> 5 [99:35]
        |       |               |   |--LAMBDA -> -> [99:37]    <--- once we reach  `->` or `:` it means that we finished the deepscan of case expr tree then we can break
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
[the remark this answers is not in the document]
mahfouz72: > Description
> Rationale
> Validates
> Limitations

> Maybe we don't need a fully documented limitations section. We define what we validate. I am not seeing how that doesn't cover the limitations fully. Why not just trim limitations to a simple sentence (or 2), no list. Let me know if you agree.

done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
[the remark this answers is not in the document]
mahfouz72: I moved it right after the Rationale IMO, it is better and more readable to be in a separate section right after please let me know if you have any comment

mahfouz72:
github, generate site

mahfouz72:
github, generate report

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithPrimitives.java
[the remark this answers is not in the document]
rnveach: The bytecode is like: `invokedynamic #20,  0             // InvokeDynamic #0:typeSwitch:(Ljava/lang/Object;I)I`
It is then followed by a `tableSwitch` with a set of integers and addresses associated to those integers.

It takes the `i` as the first parameter, some constant 0 as the second parameter, and it returns an integer (I believe unrelated to what `i` really is).

IMO this is how I believe the bytecode works: the return value is the value to use in the later `tableSwitch` look up. `typeSwitch`'s basic logic is it first checks if `i` is null and if so returns -1. If that doesn't match, it then unboxes, and just returns the unboxed value as the result.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java
rnveach: I assume this is done?

I am not really liking the `TokenUtil.isOfType` to call, but it seems necessary since our end AST is in different positions and some of the other tokens around it could be missing.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithPrimitives.java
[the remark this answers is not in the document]
rnveach: It looks like the bytecode of a pattern case with a when is the same as if the user wrote out an if statement inside the case. It did do an auto-unbox when doing the comparison.

Interestingly it did a loop on the switch and increased the value of the second parameter to 1 to handle when the `when` didn't match.

Java code:
````
		Object obj = 0;

		switch (obj) {
		case Integer i when i == 0: {
			System.out.println("pattern");
		}
			break;
		case null: {
			System.out.println("null");
			break;
		}
		default: {
			System.out.println("default");
		}
		}
````
ChatGPT Java code from the bytecode:
````
        Integer obj = Integer.valueOf(0);
        Integer temp = obj;
        int index = 0;

        while (true) {
            switch (typeSwitch(temp, index)) {
                case -1:
                    System.out.println("null");
                    break;
                case 0:
                    Integer intObj = (Integer) temp;
                    if (intObj.intValue() != 0) {
                        index = 1;
                        continue;
                    } else {
                        System.out.println("pattern");
                    }
                    break;
                default:
                    System.out.println("default");
                    break;
            }
            break;  // Exit the while loop
        }
    }

    public static int typeSwitch(Object obj, int index) {
        if (obj == null) {
            return -1; // Representing a null case
        }

        if (index == 0) {
            return 0; // Non-null, default behavior
        }

        // Other cases can be handled here if necessary
        return 1; // Default case
    }
````

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java (hunk 40 above)
Comment (mahfouz72) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingNullCaseInSwitchCheck.java]: > I assume this is done?

Yes, this is done. we can resolve it

**Comment by mahfouz72** [comment]
@romani ping

**Comment by nrmancuso** [comment]
We can do simultaneous review to get this across the finish line

**Comment by mahfouz72** [comment]
@rnveach @romani ping please

File: src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/missingnullcaseinswitch/InputXpathMissingNullCaseInSwitchNested.java (hunk 17 above)
Comment (romani) [reply src/it/resources-noncompilable/org/checkstyle/suppressionxpathfilter/missingnullcaseinswitch/InputXpathMissingNullCaseInSwitchNested.java:14]: It is different to keep compatibility.
https://stackoverflow.com/a/72475165/1015848

File: src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/Example1.java [hunk 41]
```diff
@@ -0,0 +1,60 @@
...
+      case SECONDS -> {}
+    }
+  }
+
+  // This is example is also ok, because we do not know the return type
+  // of a given method. It could be an 'int' or an 'Integer'.
```
romani, line 52: ` of variable t`

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithStringLiterals.java
```diff
@@ -0,0 +1,119 @@
...
+*/
+
+//non-compiled with javac: Compilable with Java17
+package com.puppycrawl.tools.checkstyle.checks.coding.missingnullcaseinswitch;
+
+public class InputMissingNullCaseInSwitchWithStringLiterals {
+
+    void testSwitchRule(String obj) {
+        switch (obj) {   // violation, 'Switch using reference types should have a null case.'
+            default -> {}
```
romani, line 13: Is null case is a feature of jdk17 or 21 ?
If of 21, we should not do violations on jdk17 code.
If jdk17, we need to update documentation description of Check

romani (changes requested):
Did you thought on how to avoid violation in case variable is marked by NotNull annotation?
Such markings of method arguments become popular.
Can xpath help us or we need property?

Items

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithStringLiterals.java
mahfouz72: 17 as a preview 21 final. That's why I put 21 in the doc because it is the final. Do you think we should put the first preview?

romani: If 21 is official, we should have some way to not violate jdk17 code that doesn't use preview features in compiler.

https://www.google.com/amp/s/www.geeksforgeeks.org/string-in-switch-case-in-java/amp/

We can not do violations on jdk17 code.

@nrmancuso , do we need jdk property to Check ?

Nuance is that in big companies checkstyle configuration is single file and projects are on different jdk due to long migration. Some old projects stay on old jdk for very long.

mahfouz72: I don't think we need property here. This is a single decision either you have a new enough Java version then you can use the check or an old java version so you will not use it.

> Nuance is that in big companies checkstyle configuration is single file and projects are on different jdk  

Then don't use this check at all. This check is applicable only for 17-18-19-20 previews and Java 21+

Am I getting you correctly? can you share a specific example where we need a property in the check

romani: Usually all Checks that we have can run on any jdk. 
At least we need to clearly state this in description that usage on jdk before 21 is prone to false positives and advised to not be used.

So user will do decision himself on how to  activate this Check.

nrmancuso: > At least we need to clearly state this in description that usage on jdk before 21 is prone to false positives and advised to not be used.

Yes, we can provide some details about Java version in the description, this is a good idea. As far as false positives go, we don’t need to mention this. We can just talk about when `when` was released and how this version of Java is the intended target of this check. 

We do not need a property here. You either use the check or not.

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/Example1.java
mahfouz72: done

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingnullcaseinswitch/InputMissingNullCaseInSwitchWithStringLiterals.java
mahfouz72: > As of Java 21, it is now possible to integrate a null check within the switch, eliminating the risk of NullPointerException

Is this enough?

romani:
Github, generate site

romani:
Screenshot_20240712-094558

We need very clear disclaimer that this Check should be activated only for jdk21.

romani:
Github, generate site

romani (approved):
Ok to merge

mahfouz72:
@rnveach please ping

PR #15116, opened 2024-06-24
Issue #14937: Migrated section 4.1.2 Nonempty blocks K and R style to chapter wise testing

Zopsss:
#14937


src/it/java/com/google/checkstyle/test/chapter4formatting/rule412nonemptyblocks/NonemptyBlocksKRStyleTest.java
```diff
@@ -0,0 +1,94 @@
...
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package com.google.checkstyle.test.chapter4formatting.rule412nonemptyblocks;
+
+import org.junit.jupiter.api.Test;
+
+import com.google.checkstyle.test.base.AbstractGoogleModuleTestSupport;
+
+// -@cs[AbbreviationAsWordInName] Test should be named as its chapter name.
+public class NonemptyBlocksKRStyleTest extends AbstractGoogleModuleTestSupport {
```
romani, line 27: `NonemptyBlocksKrStyleTest`
Please respect, Abbreviation rules of GOOGLE style.

romani (changes requested):
Item

src/it/java/com/google/checkstyle/test/chapter4formatting/rule412nonemptyblocks/NonemptyBlocksKRStyleTest.java
Zopsss: The section name is: `Nonempty blocks: K & R Style`

Here `K` and `R` both different words, that's why they're in capital.


Based on the implementation of XdocPagesTest validation, we need to use the `NonemptyBlocksKRStyleTest`, if we change the name implementation it'd affect other names too.

romani: But we violate Google style on Abbreviations.
Let's update XdocPagesTest validation to change remove ` & ` , to let it become KR as abbreviation, and later on end up as 'Kr'

romani: https://checkstyle.sourceforge.io/styleguides/google-java-style-20180523/javaguide.html#s5.3-camel-case

Zopsss: done

https://github.com/checkstyle/checkstyle/blob/9c780b1ce791087f3a9963561f315e3621aa93e8/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L2010-L2016

Zopsss: removed the suppression

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -2008,6 +2007,14 @@ else if (part.matches("\\d+")) {
             }
         }
 
+        // If found Abbreviation, make the next letter lowercase
+        for (int index = 0; index < extractedRuleName.length() - 1; index++) {
+            if (Character.isUpperCase(extractedRuleName.charAt(index))) {
+                extractedRuleName.setCharAt(
+                        index + 1, Character.toLowerCase(extractedRuleName.charAt(index + 1)));
+            }
```
romani, line 2015: I have worries on this.
Lets do as I suggested simple removal of " & " from test to make words around be single word.
"K & R" --> "KR" , you should do this after `ruleNameWithoutPrecedingNumbers.split`

romani (changes requested):
item:

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: > "K & R" --> "KR" , you should do this after ruleNameWithoutPrecedingNumbers.split

We're already following this. After `ruleNameWithoutPrecedingNumbers.split` we get an array of words. Then we make the first letter of the word capital and other letters to lower case.

Here after removing the `&`, we get an array of words where `K` and `R` both are considered as different words. After that we combine `K`, `R` to `KR` and then it gets converted to `Kr`.

I think this logic is fine there should not be any problem with this.

Zopsss: > "K & R" --> "KR" , you should do this after ruleNameWithoutPrecedingNumbers.split

We're already following this. After `ruleNameWithoutPrecedingNumbers.split` we get an array of words. Then we make the first letter of the word capital and other letters to lower case in foreach loop. Here the special symbols gets ignored so the `&` will get ignored.

In the array of words `K` and `R` both are considered as different words. After that we combine `K`, `R` to `KR` in the foreach loop and then it gets converted to `Kr` in the last loop.

I think this logic is fine there should not be any problem with this.

Zopsss: So you're suggesting to combine `K` and `R` into one word `KR` before `for (String part : parts)` this means that we're combing all the words into a single string like `NonemptyblocksKRstyle`. Then we cannot split the string or perform the camel case operation. 

If we write some logic for whenever we encounter a capital letter, we make the next letter lower case then it'll not work for: `Non-ASCII characters`

romani: update is pushed

romani (approved):
ok to merge

src/it/resources/com/google/checkstyle/test/chapter4formatting/rule412nonemptyblocks/InputRightCurly.java
```diff
@@ -1,9 +1,11 @@
 package com.google.checkstyle.test.chapter4formatting.rule412nonemptyblocks;
 
-public class InputRightCurly2 {
+public class InputRightCurly {
     public static void main(String[] args) {
         boolean after = false;
         try {
-        } finally { after = true; }
+        } finally {
```
rdiachenko, line 7: I noticed we changed original formatting in input files. This is just one example, there others. Considering it doesn't affect existing violations under the test, I'm ok with the changes. However, in next PRs I suggest to not mix formatting with migration. cc @romani

rdiachenko (approved):

src/it/resources/com/google/checkstyle/test/chapter4formatting/rule412nonemptyblocks/InputRightCurly.java
romani: @Zopsss , please restore, in this Check, such formatting matters.

romani: @Zopsss , I merged to move forward big changes, please send separate PR to restore location of curly brace

Zopsss: It was a violation that's why I had to change it. Anyways I've restored it here: https://github.com/checkstyle/checkstyle/pull/15148#issue-2377311149

PR #15117, opened 2024-06-25
Issue #15016: Remove underscore from illegal identifers format

### LINKED ISSUE #15016: Add Check Support for Java 21 Unnamed Variables & Patterns Syntax: IllegalIdentifierName
child of https://github.com/checkstyle/checkstyle/issues/14942
I have read check documentation: https://checkstyle.org/checks/naming/illegalidentifiername.html

```
import java.util.Queue;

public class Test {
    void test() {
        q.add(1);
        var _ = q.poll();
    }
}
PS D:\CS\test> cat config.xml                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
         <module name="IllegalIdentifierName"/>
    </module>
</module>
Starting audit...
[ERROR] D:\CS\test\src\Test.java:8:13: Name '_' must match pattern '(?i)^(?!(record|yield|var|permits|sealed|_)$).+$'. [IllegalIdentifierName]
Audit done.
PS D:\CS\test> 
```

Describe what you expect in detail.

 `IllegalIdentifierName` should allow unnamed pattern variables by default

mahfouz72:
closes #15016 :  

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/ecdaf1a2931b1a7d14f2d0c6d68993923bbc0c91/check.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/IllegalIdentifierNameCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/IllegalIdentifierNameCheck.java
@@ -27,9 +27,9 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 /**
  * <p>
  * Checks identifiers with a pattern for a set of illegal names, such as those
- * that are restricted or contextual keywords. Examples include "yield", "record", "_", and
+ * that are restricted or contextual keywords. Examples include "yield", "record", and
  * "var". Please read more at
- * <a href="https://docs.oracle.com/javase/specs/jls/se14/html/jls-3.html#jls-3.9">
+ * <a href="https://docs.oracle.com/javase/specs/jls/se22/html/jls-3.html#jls-3.9">
  * Java Language Specification</a> to get to know more about restricted keywords. Since this
  * check uses a pattern to specify valid identifiers, users can also prohibit the usage
  * of certain symbols, such as "$", or any non-ascii character.
@@ -38,7 +38,7 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
  * <li>
  * Property {@code format} - Sets the pattern to match valid identifiers.
  * Type is {@code java.util.regex.Pattern}.
- * Default value is {@code "(?i)^(?!(record|yield|var|permits|sealed|_)$).+$"}.
+ * Default value is {@code "(?i)^(?!(record|yield|var|permits|sealed)$).+$"}.
  * </li>
  * <li>
  * Property {@code tokens} - tokens to check
@@ -94,7 +94,7 @@ public class IllegalIdentifierNameCheck extends AbstractNameCheck {
      * Creates a new {@code IllegalIdentifierNameCheck} instance.
      */
     public IllegalIdentifierNameCheck() {
-        super("(?i)^(?!(record|yield|var|permits|sealed|_)$).+$");
+        super("(?i)^(?!(record|yield|var|permits|sealed)$).+$");
     }
 
     @Override
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/naming/illegalidentifiername/InputIllegalIdentifierNameUnnamedVariables.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/naming/illegalidentifiername/InputIllegalIdentifierNameUnnamedVariables.java
@@ -1,0 +1,46 @@
+/*
+IllegalIdentifierName
+format = (default)(?i)^(?!(record|yield|var|permits|sealed)$).+$
+tokens = (default)CLASS_DEF, INTERFACE_DEF, ENUM_DEF, ANNOTATION_DEF, ANNOTATION_FIELD_DEF, \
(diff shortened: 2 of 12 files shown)


mahfouz72:
Github, generate site

nrmancuso (approved):

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/IllegalIdentifierNameCheck.java
```diff
@@ -27,9 +27,9 @@
 /**
  * <p>
  * Checks identifiers with a pattern for a set of illegal names, such as those
- * that are restricted or contextual keywords. Examples include "yield", "record", "_", and
+ * that are restricted or contextual keywords. Examples include "yield", "record", and
  * "var". Please read more at
- * <a href="https://docs.oracle.com/javase/specs/jls/se14/html/jls-3.html#jls-3.9">
+ * <a href="https://docs.oracle.com/javase/specs/jls/se22/html/jls-3.html#jls-3.9">
```
rnveach, line 32: @mahfouz72 @nrmancuso What do you guys think of making this an issue? Finding all old JLS and moving them to 22.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/naming/illegalidentifiername/InputIllegalIdentifierNameUnnamedVariables.java
```diff
@@ -0,0 +1,46 @@
...
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.naming.illegalidentifiername;
+
+public class InputIllegalIdentifierNameUnnamedVariables {
+    void m(Object o) {
+        int _ = sideEffect();
+        for (Integer _ : new int[0]) {
```
rnveach, line 16: Please add other examples like `__` (double underscore) and `_BAD` and `BAD_`. I just want to verify that they won't be picked up.

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/IllegalIdentifierNameCheck.java
rnveach, line 32: @mahfouz72 @nrmancuso Now that I check the JLS, I have a question and concern.

https://docs.oracle.com/javase/specs/jls/se22/html/jls-3.html#jls-3.9
It is listed as a reserved keyword. Check description says restricted or **contextual** keywords. I assume this is why this won't be a violation with Java 22+?

However, with Java 16 and below,
https://docs.oracle.com/javase/specs/jls/se16/html/jls-3.html#jls-3.9
The JLS says this is a **keyword**. It doesn't say what type, except it is a keyword. My assumption is we should violate everything listed there for Java 16 and below.

Doesn't this change mean we are breaking with old Java versions? We discussed adding a Java version to another check for a similar situation. Shouldn't we do the same here?

Funny enough, starting with Java 8 and below,
https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.9
The JLS does not even list the underscore at all. I assume this also means we shouldn't have a violation here with Java 8 and below.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/IllegalIdentifierNameCheck.java
rnveach: This can be ignored for now, but me and @nrmancuso are talking this over in private.

mahfouz72: Good idea I am ok with this it can be a separate issue to keep tacker of all checks or I can update it in other PRs (normal updates) whenever I see an old link in the doc 
I think a separate issue would be better

mahfouz72: But does updating links add value? I think most of the links either 22 or 14 will have no difference in the content. there will be a difference ofc in links related to the new syntax (patterns and switch)

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/naming/illegalidentifiername/InputIllegalIdentifierNameUnnamedVariables.java
mahfouz72: done

rnveach (approved):

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/IllegalIdentifierNameCheck.java
nrmancuso: @rnveach you can create an issue for this, but we will not cover it during GSOC.

rnveach: Done.

PR #15127, opened 2024-06-25
Issue #15123: False Positive when case null in switch statements group

### LINKED ISSUE #15123: MissingSwitchDefault : False positive when `case null` in switch labeled statement group
Check documentation : https://checkstyle.org/checks/coding/missingswitchdefault.html#MissingSwitchDefault

---
```
public class Test {
    void m(E i) {
        switch (i) {  // violation
            case FIRST :
            case SECOND :
            case null :
                System.out.println("NULL first and second");
        }
    }
    enum E {
        FIRST, SECOND
    }
}
PS D:\CS\test> cat config.xml                                                    
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
  <module name="TreeWalker">
            <module name="MissingSwitchDefault"/>
        </module>
</module>
PS D:\CS\test> java  -jar checkstyle-10.17.0-all.jar -c config.xml  src/Test.java
Starting audit...
[ERROR] D:\CS\test\src\Test.java:3:9: switch without "default" clause. [MissingSwitchDefault]
Audit done.
PS D:\CS\test> 

```

---
I expect no violation because there is a case null in the switch so exhaustiveness is being checked. the issue is that the check doesn't detect case null if it is in a labeled group.

mahfouz72:
closes #15123 

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/fd6bb80bfe66feb1d7bf946caf87c0a7be981f6c/check.xml

### DIFF
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingswitchdefault/InputMissingSwitchDefaultCheckNullCaseLabel.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingswitchdefault/InputMissingSwitchDefaultCheckNullCaseLabel.java
@@ -15,6 +15,24 @@ public class InputMissingSwitchDefaultCheckNullCaseLabel {
             case SECOND: counter--;
             case null: counter=counter+2;
         }
+        switch (i) {
+            case FIRST:
+            case SECOND:
+            case null:
+                counter=counter+2;
+        }
+        switch (i) {
+            case FIRST:
+            case null:
+            case SECOND:
+                counter=counter+2;
+        }
+        switch (i) {
+            case null:
+            case FIRST:
+            case SECOND:
+                counter=counter+2;
+        }
     }
     enum SwitchInput {
         FIRST,
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
@@ -176,13 +176,13 @@ public class MissingSwitchDefaultCheck extends AbstractCheck {
     /**
      * Checks if the case contains null label.
      *
-     * @param ast the switch statement we are checking
+     * @param detailAST the switch statement we are checking
      * @return returnValue the ast of null label
      */
-    private static boolean hasNullCaseLabel(DetailAST ast) {
-        final DetailAST firstChild = ast.getFirstChild();
-        return firstChild != null
-                && TokenUtil.isOfType(firstChild.getFirstChild(), TokenTypes.LITERAL_NULL);
-
+    private static boolean hasNullCaseLabel(DetailAST detailAST) {
+        return TokenUtil.findFirstTokenByPredicate(detailAST.getParent(), ast -> {
+            final DetailAST expr = ast.findFirstToken(TokenTypes.EXPR);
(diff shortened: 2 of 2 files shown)


mahfouz72:
Github, generate report

mahfouz72:
Sounds like we don't have similar examples in the projects
```
        switch (i) { 
            case FIRST :
            case SECOND :
            case null :
                System.out.println("NULL first and second");
        }
```

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/missingswitchdefault/InputMissingSwitchDefaultCheckNullCaseLabel.java
```diff
@@ -15,6 +15,24 @@ public void switchWithNullLabelTest(SwitchInput i) {
             case SECOND: counter--;
             case null: counter=counter+2;
```
rnveach, line 16: So we had an existing case here.

@mahfouz72 Can you clarify why this is an issue if we had an existing case? I am not seeing any real differences between this and your next new input.

nrmancuso: His comment was in reference to no diffs in the report.

mahfouz72: > I am not seeing any real differences between this and your next new input.

in my new input the case null is grouped with other cases but in old input the case null is an alone case not grouped 


```Java
case SECOND:             // no statement here so both cases grouped under same CASR_GROUP
case null: counter=counter+2;
```

```Java
case SECOND: counter--; // each case is under its CASE_GROUP
case null: counter=counter+2; 
```

rnveach: Thanks, I see now. We were only check first item in group and in your example, we have multiple.

````
        |      |  |--CASE_GROUP -> CASE_GROUP [9:12]
        |      |  |  |--LITERAL_CASE -> case [9:12]
        |      |  |  |  |--EXPR -> EXPR [9:17]
        |      |  |  |  |  `--IDENT -> FIRST [9:17]
        |      |  |  |  `--COLON -> : [9:22]
        |      |  |  |--LITERAL_CASE -> case [10:12]
        |      |  |  |  |--EXPR -> EXPR [10:17]
        |      |  |  |  |  `--IDENT -> SECOND [10:17]
        |      |  |  |  `--COLON -> : [10:23]
        |      |  |  |--LITERAL_CASE -> case [11:12]
        |      |  |  |  |--EXPR -> EXPR [11:17]
        |      |  |  |  |  `--LITERAL_NULL -> null [11:17]
````

rnveach (approved):

rnveach:
@mahfouz72 One last concern. Diff report shows no differences.

Have we looked for new projects to expand regression on?

I noticed we don't have openjdk 22 listed. Not sure what Java version this issue is connected to.

rnveach:
Ok, I am good to merge this but I thought part of this project was also finding new projects for regression as well as best practices examples.

PR #15148, opened 2024-06-27
minor: Restored formatting for InputRightCurly.java

Zopsss:
Detected at: https://github.com/checkstyle/checkstyle/pull/15116#discussion_r1655465265

### DIFF
--- a/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule412nonemptyblocks/InputRightCurly.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule412nonemptyblocks/InputRightCurly.java
@@ -4,8 +4,7 @@ public class InputRightCurly {
     public static void main(String[] args) {
         boolean after = false;
         try {
-        } finally {
-            after = true;
-        }
+        } finally { after = true; }
+        // violation above ''{' at column 19 should have line break after.'
     }
 }


src/it/resources/com/google/checkstyle/test/chapter4formatting/rule412nonemptyblocks/InputRightCurly.java
```diff
@@ -4,7 +4,8 @@ public class InputRightCurly {
     public static void main(String[] args) {
         boolean after = false;
         try {
-        } finally {
+        } // violation ''}' at column 9 should be on the same line as the next part'
+        finally {
```
Zopsss, line 8: Restored the old curly location and added the violation message

romani: @Zopsss , please restore as it was original and just place a comment if there is violation.

rnveach: Original had no violation, so if there is one now that is a difference and the line isn't restored.

Zopsss: corrected the restored code. The violation is occurring because of `{`

https://github.com/checkstyle/checkstyle/blob/148e78d25899444ba5531ff03f1db5078af7a899/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule412nonemptyblocks/InputRightCurly.java#L6-L8

It was not occurring before because we were using the `RightCurly` module only. In the new test, we're using both module: `LeftCurly` & `RightCurly` so it became violation for LeftCurly

romani (approved):
ok to merge.

no problem, we just show how system works now.

src/it/resources/com/google/checkstyle/test/chapter4formatting/rule412nonemptyblocks/InputRightCurly.java
rnveach: > LeftCurly & RightCurly so it became violation for LeftCurly

This makes sense, but this still seems a bit odd. Why is there a violation for left curly but right curly is saying everything is ok?

I think we need to review this and the style guide. This looks like it might be a bug of some kind.

rnveach: I am going to approve this. I still think it is odd Right is fine.

rnveach (approved):

rdiachenko (approved):
lgtm

PR #15154, opened 2024-06-27
Issue #15015: Update CatchParameterName format to allow unnamed catch…

### LINKED ISSUE #15015: Add Check Support for Java 21 Unnamed Variables & Patterns Syntax: CatchParameterName
child of https://github.com/checkstyle/checkstyle/issues/14942
I have read check documentation: https://checkstyle.org/checks/naming/catchparametername.html#CatchParameterName

```
public class Test {
    void test() {
        try {
            int x = 1/0;
        } catch (Exception _) {

        }
    }
}
PS D:\CS\test> cat config.xml                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
         <module name="CatchParameterName"/>
    </module>
</module>
Starting audit...
[ERROR] D:\CS\test\src\Test.java:5:28: Name '_' must match pattern '^(e|t|ex|[a-z][a-z][a-zA-Z]+)$'. [CatchParameterName]
Audit done.
PS D:\CS\test> 
```

Describe what you expect in detail.

`CatchParameterName` should allow unnamed pattern variables by default

mahfouz72:
closes #15015: 

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/696b257d7628c4a99328382b0878d6a07a46ff1a/check.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/CatchParameterNameCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/CatchParameterNameCheck.java
@@ -35,6 +35,7 @@ import com.puppycrawl.tools.checkstyle.api.TokenTypes;
  * <li>allows {@code e} abbreviation (suitable for exceptions end errors)</li>
  * <li>allows {@code ex} abbreviation (suitable for exceptions)</li>
  * <li>allows {@code t} abbreviation (suitable for throwables)</li>
+ * <li>allows {@code _} for unnamed catch parameters</li>
  * <li>prohibits numbered abbreviations like {@code e1} or {@code t2}</li>
  * <li>prohibits one letter prefixes like {@code pException}</li>
  * <li>prohibits two letter abbreviations like {@code ie} or {@code ee}</li>
@@ -44,7 +45,7 @@ import com.puppycrawl.tools.checkstyle.api.TokenTypes;
  * <li>
  * Property {@code format} - Sets the pattern to match valid identifiers.
  * Type is {@code java.util.regex.Pattern}.
- * Default value is {@code "^(e|t|ex|[a-z][a-z][a-zA-Z]+)$"}.
+ * Default value is {@code "^(e|t|ex|[a-z][a-z][a-zA-Z]+|_)$"}.
  * </li>
  * </ul>
  * <p>
@@ -67,7 +68,7 @@ public class CatchParameterNameCheck extends AbstractNameCheck {
      * Creates a new {@code CatchParameterNameCheck} instance.
      */
     public CatchParameterNameCheck() {
-        super("^(e|t|ex|[a-z][a-z][a-zA-Z]+)$");
+        super("^(e|t|ex|[a-z][a-z][a-zA-Z]+|_)$");
     }
 
     @Override
--- a/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionCatchParameterNameTest.java
+++ b/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionCatchParameterNameTest.java
@@ -40,7 +40,7 @@ public class XpathRegressionCatchParameterNameTest extends AbstractXpathTestSupp
 
     @Test
     public void testSimple() throws Exception {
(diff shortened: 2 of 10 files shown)


mahfouz72:
Github, generate site

mahfouz72:
Github, generate report

nrmancuso (changes requested):
Please update documentation (add a bullet point) to mention that we allow underscores for unnamed variables, and update/remove the bullet point that mentions that we do not allow symbols.

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/CatchParameterNameCheck.java
```diff
@@ -35,16 +35,16 @@
  * <li>allows {@code e} abbreviation (suitable for exceptions end errors)</li>
  * <li>allows {@code ex} abbreviation (suitable for exceptions)</li>
  * <li>allows {@code t} abbreviation (suitable for throwables)</li>
+ * <li>allows {@code _} for unnamed catch parameter</li>
```
nrmancuso, line 38:
```suggestion
 * <li>allows {@code _} for unnamed catch parameters</li>
```
Other items are plural, let's be consistent

nrmancuso (changes requested):
minor item, last from me:

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/CatchParameterNameCheck.java
mahfouz72: done

nrmancuso (approved):
Good stuff.

mahfouz72:
Github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/CatchParameterNameCheck.java
```diff
@@ -35,16 +35,16 @@
  * <li>allows {@code e} abbreviation (suitable for exceptions end errors)</li>
  * <li>allows {@code ex} abbreviation (suitable for exceptions)</li>
  * <li>allows {@code t} abbreviation (suitable for throwables)</li>
+ * <li>allows {@code _} for unnamed catch parameters</li>
  * <li>prohibits numbered abbreviations like {@code e1} or {@code t2}</li>
  * <li>prohibits one letter prefixes like {@code pException}</li>
  * <li>prohibits two letter abbreviations like {@code ie} or {@code ee}</li>
- * <li>prohibits any other characters than letters</li>
```
rnveach, line 41: If we are going to remove this then there needs to be something equivalent that mentions the rest of the prohibitions not mentioned above. This statement covered those before.

For example, we don't say `$` is prohibited, nor underscores (excluding `_` alone) nor numbers. We say numbered abbreviations above, but its not just abbreviations, it is all numbers.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/CatchParameterNameCheck.java
mahfouz72: prohibits any other characters than letters and underscore?

rnveach: > prohibits any other characters than letters and underscore?

The concern with just saying "underscore" is you have to make sure they understand underscore with anything else is disallowed. It is just a single `_` which is allowed.

Truthfully, I think original wording is fine. We say first we allow `_`. We can say we don't allow anything other than letters. It should be understood it is top to bottom order. Allows overrides the prohibits.

mahfouz72: > Truthfully, I think original wording is fine. We say first we allow `_`. We can say we don't allow anything other than letters. It should be understood it is top to bottom order. Allows overrides the prohibits.

make sense. Original wording is restored

rnveach (approved):

PR #15166, opened 2024-06-28
Issue #14963: Add new property to skip unnamed parameters in FinalPar…

### LINKED ISSUE #14963: Add Check Support for Java 21 Unnamed Variables & Patterns Syntax: FinalParameter
child of #14942 

I have read check documentation: https://checkstyle.org/checks/misc/finalparameters.html#FinalParameters

```
import java.util.Queue;

class Test {
    void test() {
        q.add(1);
        for (Integer _ : q) {  // violation
            q.poll();
        }
        try {
            int x = 1 / 0;
        } catch (Exception _) {       // violation
            System.out.println("division by zero");
        }
    }
}

PS D:\CS\test> cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
 <module name="TreeWalker">
    <module name="FinalParameters">
        <property name="tokens" value="FOR_EACH_CLAUSE, LITERAL_CATCH"/>
    </module>
 </module>
</module>
Starting audit...
[ERROR] D:\CS\test\src\Test.java:8:14: Parameter _ should be final. [FinalParameters]
[ERROR] D:\CS\test\src\Test.java:13:18: Parameter _ should be final. [FinalParameters]
Audit done.


```
---

check doc : 

> Rationale: Changing the value of parameters during the execution of the method's algorithm can be confusing and should be avoided. A great way to let the Java compiler prevent this coding style is to declare parameters final.

JLS:

> If a declaration does not include an identifier, but instead includes the reserved keyword _ (underscore), then the entity cannot be referred to by name.

So I consider this as a false positive unnamed variables can't be referred to by name and Their value can't change so there is no point to require it to be final

---

mahfouz72:
closes #14963:   

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/25589ae66b95b075875747c055df30118c0861d3/check.xml

Diff Regression patch config: https://gist.githubusercontent.com/mahfouz72/33967bf86e4a29cafa429771b0e95808/raw/72f89107ab1b81c270885ee5042ccaf535221089/check_patch.xml

Report Label: property is false

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/FinalParametersCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/FinalParametersCheck.java
@@ -48,6 +48,13 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * Default value is {@code false}.
  * </li>
  * <li>
+ * Property {@code ignoreUnnamedParameters} -
+ * Ignore <a href="https://docs.oracle.com/en/java/javase/21/docs/specs/unnamed-jls.html">
+ * unnamed parameters</a>.
+ * Type is {@code boolean}.
+ * Default value is {@code true}.
+ * </li>
+ * <li>
  * Property {@code tokens} - tokens to check
  * Type is {@code java.lang.String[]}.
  * Validation type is {@code tokenSet}.
@@ -103,6 +110,12 @@ public class FinalParametersCheck extends AbstractCheck {
     private boolean ignorePrimitiveTypes;
 
     /**
+     * Ignore <a href="https://docs.oracle.com/en/java/javase/21/docs/specs/unnamed-jls.html">
+     * unnamed parameters</a>.
+     */
+    private boolean ignoreUnnamedParameters = true;
+
+    /**
      * Setter to ignore primitive types as parameters.
      *
      * @param ignorePrimitiveTypes true or false.
@@ -110,6 +123,18 @@ public class FinalParametersCheck extends AbstractCheck {
      */
     public void setIgnorePrimitiveTypes(boolean ignorePrimitiveTypes) {
         this.ignorePrimitiveTypes = ignorePrimitiveTypes;
+    }
+
+    /**
+     * Setter to ignore
+     * <a href="https://docs.oracle.com/en/java/javase/21/docs/specs/unnamed-jls.html">
+     * unnamed parameters</a>.
+     *
+     * @param ignoreUnnamedParameters true or false.
+     * @since 10.18.0
+     */
+    public void setIgnoreUnnamedParameters(boolean ignoreUnnamedParameters) {
+        this.ignoreUnnamedParameters = ignoreUnnamedParameters;
     }
 
     @Override
@@ -195,7 +220,8 @@ public class FinalParametersCheck extends AbstractCheck {
      */
     private void checkParam(final DetailAST param) {
         if (param.findFirstToken(TokenTypes.MODIFIERS).findFirstToken(TokenTypes.FINAL) == null
-                && !isIgnoredParam(param)
+                && !isIgnoredPrimitiveParam(param)
+                && !isIgnoredUnnamedParam(param)
                 && !CheckUtil.isReceiverParameter(param)) {
             final DetailAST paramName = param.findFirstToken(TokenTypes.IDENT);
             final DetailAST firstNode = CheckUtil.getFirstNode(param);
@@ -210,7 +236,7 @@ public class FinalParametersCheck extends AbstractCheck {
      * @param paramDef {@link TokenTypes#PARAMETER_DEF PARAMETER_DEF}
      * @return true if param has to be skipped.
      */
-    private boolean isIgnoredParam(DetailAST paramDef) {
+    private boolean isIgnoredPrimitiveParam(DetailAST paramDef) {
         boolean result = false;
         if (ignorePrimitiveTypes) {
             final DetailAST type = paramDef.findFirstToken(TokenTypes.TYPE);
@@ -223,6 +249,17 @@ public class FinalParametersCheck extends AbstractCheck {
             }
         }
         return result;
+    }
+
+    /**
+     *  Checks for skip current param due to <b>ignoreUnnamedParameters</b> option.
+     *
+     * @param paramDef parameter to check
+     * @return true if the parameter should be skipped due to the ignoreUnnamedParameters option.
+     */
+    private boolean isIgnoredUnnamedParam(final DetailAST paramDef) {
+        final DetailAST paramName = paramDef.findFirstToken(TokenTypes.IDENT);
+        return ignoreUnnamedParameters && paramName != null && "_".equals(paramName.getText());
     }
 
 }
--- a/src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/finalparameters/Example4.java
+++ b/src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/finalparameters/Example4.java
@@ -1,0 +1,54 @@
+/*xml
+<module name="Checker">
+  <module name="TreeWalker">
+    <module name="FinalParameters">
+      <property name="tokens" value="FOR_EACH_CLAUSE, LITERAL_CATCH"/>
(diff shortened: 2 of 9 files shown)


mahfouz72:
github, generate report

mahfouz72:
github, generate site

mahfouz72:
github, generate report

mahfouz72:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/e53d9a2_2024010200/reports/diff/checkstyle/index.html#A3 This exception due to record pattern syntax in enhanced for that is no longer compilable syntax. It was in preview and removed in the final release

nrmancuso:
> https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/e53d9a2_2024010200/reports/diff/checkstyle/index.html#A3 This exception due to record pattern syntax in enhanced for that is no longer compilable syntax. It was in preview and removed in the final release

Please always show this sort of thing via javac.

src/main/java/com/puppycrawl/tools/checkstyle/checks/FinalParametersCheck.java
```diff
@@ -48,6 +48,13 @@
  * Default value is {@code false}.
  * </li>
  * <li>
+ * Property {@code ignoreUnnamedParameters} -
```
nrmancuso, line 51: Not really for this line, but please add a new example that shows how this property works, with a brief explanation about why someone may want to use it.

nrmancuso (changes requested):
Code looks good, one item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/FinalParametersCheck.java
mahfouz72: should we always do this?  I think we introduced new properties in other PRs without adding new examples for it

nrmancuso: > should we always do this?

Maybe not always, but probably most of the time.

mahfouz72:
github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/FinalParametersCheck.java
mahfouz72: @nrmancuso

mahfouz72:
github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/FinalParametersCheck.java
mahfouz72: @nrmancuso done

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/finalparameters/Example4.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  void testForEachParameters() {
+    int[] l = {1, 2, 3};
+    int x = 0;
+    for (int number: l) { // violation, 'Parameter number should be final'
+      System.out.println(number);
+    }
+    for (int _: l) {  // ok, unnamed catch parameter is ignored, it can't be used
```
nrmancuso, line 39: Looks like copypasta
```suggestion
    for (int _: l) {  // ok, unnamed enhanced for loop variable, it is implicitly final
```

nrmancuso (changes requested):
Last item:

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/finalparameters/Example4.java
mahfouz72: > Looks like copypasta

Yes :) thank you. Updated the above comment also to be consistent in wording.
`it can't be used` =>` it is implicitly final`

nrmancuso (approved):

src/xdocs/checks/misc/finalparameters.xml
```diff
@@ -144,6 +152,57 @@ public class Example3 {
   public void methodOne(final int x) { }
   public void methodTwo(int x) { }
   public static void main(String[] args) { } // violation, 'args should be final'
+}
+        </source>
+        <p id="Example4-config">
+          To configure the check to enforce final parameters for catch and
+          for-each blocks while ignoring unnamed parameters.
+          Since unnamed parameters cannot be read from or written to,
+          marking them as final is unnecessary:
```
rnveach, line 161: The wording is a bit weird. We usually say "To configure...: config example" (notice no period so it is a complete thought), but instead there is another sentence with an explanation instead of a config, making the first sentence not really a complete sentence since it says "To configure" without showing them how to configure.

I would remove the second sentence, replace the period of the first sentence with `:`, and use the Java code comments as a way to convey the reason you tried to show in the second sentence. If you really needed to explain more, it should be done before "To configure" or after the Java example IMO.

src/xdocs/checks/misc/finalparameters.xml
```diff
@@ -144,6 +152,57 @@ public class Example3 {
...
+    try {
+      int x = 1 / 0;
+    }
+    catch (Exception e) { // violation, 'Parameter e should be final'
+      System.out.println(e);
+    }
+    try {
+      int x = 1 / 0;
+    }
+    catch (Exception _) { // ok, unnamed catch parameter, it is implicitly final
```
rnveach, line 189: Add another example with a `final` unnamed and just say it is ok to have the final declared.

rnveach (changes requested):

rnveach:
@romani @nrmancuso
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/e53d9a2_2024010200/reports/diff/checkstyle/index.html#A3
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/e53d9a2_2024010200/reports/diff/checkstyle/index.html#A8
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/e53d9a2_2024010200/reports/diff/checkstyle/index.html#A10
Regression found an exception.

@mahfouz72 Move exception in master to new issue. Our checks should not throw an exception on compilable code.

rnveach:
Maybe I am not finding all the regressions. I should see regression of:
* Base is default with all tokens, Patch is default with `ignoreUnnamedParameters` set to true with all tokens (expecting unnamed differences only)
* Base is default with all tokens, Patch is default with `ignoreUnnamedParameters` set to false with all tokens (expecting no differences)

mahfouz72:
> If it is no longer compilable, why is it still in our repo and still marked as compilable ("Compilable with Java20")? Is this also something we should discuss removing?

I suggested somewhere that we should open an issue to remove the no longer compilable code in our repo. Such code was a preview in some version we supported it but it was removed. So I am good with this as an issue

src/xdocs/checks/misc/finalparameters.xml
mahfouz72: done completely removed it the code examples and code comments speak well for this point.

src/xdocs/checks/misc/finalparameters.xml
mahfouz72: done

mahfouz72:
github, generate report

mahfouz72:
github, generate report

rnveach:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/22a829d_2024194022/reports/diff/index.html
Change in exceptions is fine. I am asking more questions outside of this.

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/22a829d_2024183400/reports/diff/index.html

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/22a829d_2024183400/reports/diff/openjdk17/index.html#A2
This is an interesting case. Method parameters can't be unnamed, but this file isn't compilable. Pre-java 9 they could use `_` as a named method parameter. Point of ignoring property is for Java 21+ projects to use and like I said they can't have a unnamed method parameter, so this behavior seems ok to me.

rnveach (approved):

PR #15181, opened 2024-07-01
Issue #14937: Migrated section 7.1.2 Paragraphs to chapter wise testing

Zopsss:
#14937


src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputIncorrectJavadocParagraphCheck.java
```diff
@@ -2,23 +2,33 @@
 
 /**
  * Some Javadoc.
- * <p> //warn
- * /^ WARN/   Some Javadoc.<p> //warn
- *
+ * <p>
+ * // 2 violations above:
```
romani, line 6: Please move all such marker comments outside of javadoc comment at all.
```
// Violation 2 lines below....
/**
 * Some Javadoc.
 * <p> some text
 *    Some Javadoc.<p>
 */
```
```

**Review by romani — CHANGES_REQUESTED** [review]
Item same for all others

File: src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputIncorrectJavadocParagraphCheck.java (hunk 1 above)
Comment (romani) [reply src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputIncorrectJavadocParagraphCheck.java:6]: Not done

File: src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java [hunk 2]
```diff
@@ -162,6 +162,21 @@ public final class InlineConfigParser {
...
+     *   //                            'violation message1'
+     *   //                            'violation messageX'
+     * }
+     *
+     * Messages are matched by {@link InlineConfigParser#VIOLATION_MESSAGE_PATTERN}
+     * </p>
+     */
+    private static final Pattern VIOLATIONS_SOME_LINES_BELOW_PATTERN = Pattern
+            .compile(".*//\\s*(\\d+) violations (\\d+) lines below:$");
+
```
Zopsss, line 179: Added support for new kind of violation msg similar to 

https://github.com/checkstyle/checkstyle/blob/8d5e2e2c212f3b988abd2a03fb980141ecae42e3/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L150-L163

so we can display violations in a more readable way

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -716,11 +738,13 @@ private static List<TestInputViolation> getExpectedViolationsForSpecificLineAbov
...
+        final int violationLine =
             Integer.parseInt(matcher.group(2));
-        final int violationLineNum = lineNo - linesAbove + 1;
-        return getExpectedViolationsForSpecificLineAbove(lines,
+        final int violationLineNum =
+                isAbove ? lineNo - violationLine + 1
+                        : lineNo + violationLine + 1;
+        return getExpectedViolationsForSpecificLine(lines,
             lineNo, violationLineNum, matcher);
     }
```
Zopsss, line 749: Had to modify this method to identify which violation message we're checking and also it'd be helpful if you can suggest a better name for: `violationLine`

I could also create a new method similar to the existing method, that way I would not need to modify the existing method but these methods would be very identical to each other, the only difference would be:

For checking "lines above message": `lineNo - linesAbove + 1`
For checking "lines below message": `lineNo - linesBelow + 1`

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -690,7 +712,7 @@ else if (useFilteredViolations) {
         }
     }
 
-    private static List<TestInputViolation> getExpectedViolationsForSpecificLineAbove(
+    private static List<TestInputViolation> getExpectedViolationsForSpecificLine(
```
Zopsss, line 715: Renamed this method as it is not specific to only `Lines Above` messages anymore

Zopsss (commented):
Added support for new kind of Message:

src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputIncorrectJavadocParagraphCheck.java
Zopsss: done

romani:
Condl

Zopsss:
conflicts resolved

src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputIncorrectJavadocParagraphCheck.java
```diff
@@ -1,24 +1,36 @@
 package com.google.checkstyle.test.chapter7javadoc.rule712paragraphs;
 
+// 2 violations 5 lines below:
+//                           '.* tag should be placed immediately before the first word'
```
romani, line 4: Can we use exact value instead of .* ?

Let's treat .* as completely forbidden characters at the beginning of end of message

romani (changes requested):
Item

src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputIncorrectJavadocParagraphCheck.java
Zopsss: I guess we cannot. In this case `<p>`  is the tag we're talking about, but the `<` and `>` causes mismatching in comparing violation messages as we compare messages using regex. 

I have used `.*` in my other PRs too because of the similar reason. Special symbols in violation messages causes mismatching in comparing messages due to regex used in comparing messages.

romani: At PRs we merged it was clear what we violating, now it is not clear, even I don't understand it.
Can we we use `\` to escape them? It is better than not clear what is causing violation.

Zopsss: done, escaped `<p>` using `\` as you suggested:

https://github.com/checkstyle/checkstyle/blob/b0464ec57c620df634808ed56b071faf17564c35/src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputIncorrectJavadocParagraphCheck.java#L3-L5

romani (approved):
Ok to merge

src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputIncorrectJavadocParagraphCheck.java
```diff
@@ -27,51 +39,84 @@ class InputIncorrectJavadocParagraphCheck {
...
          *
          * <p>Some Javadoc.
          *
          * @since 8.0
          */
         public static final byte NUL = 0;
 
+        // 2 violations 5 lines below:
+        //                            '\<p\> tag should be placed immediately before the first word'
+        //                            '\<p\> tag should be preceded with an empty line.'
```
rdiachenko, line 107: It looks like a waste of space, especially for long messages. Can we use 2 spaces as an indent for multiple violations?

example:
```
         // 2 violations 5 lines below:
         //   '\<p\> tag should be placed immediately before the first word'
         //   '\<p\> tag should be preceded with an empty line.'
         // violation 4 lines below '\<p\> tag should be placed immediately before the first word'
```

Same for all other cases.

romani: Ok, let's do it.
When we initially created this approach it was convenient to do indentation to far right to make it more clear that such comments are very very special to not mix them.
But let's try +2 indentation.

Zopsss: done

https://github.com/checkstyle/checkstyle/blob/cf4fb7f7c11e8680e4ae5021ce971583f967b64b/src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputIncorrectJavadocParagraph.java#L18-L20

Zopsss: also renamed Input files and removed trailing `Check` from their name

romani (approved):

rdiachenko (approved):
lgtm

Issue #14805, opened 2024-04-17
Remove Support for String Template Syntax


nrmancuso:
We always take a risk supporting preview features, and try to wait until second/third/final preview to add support for new syntax, to improve the chance that a given feature will become a part of the language permanently. Of course, this is also driven by user demand.

Unfortunately, String Templates have been completely removed from JDK 23 (they are not even a preview feature). Since we have really only begun to implement them (text block templates are not supported, and no checks have really been updated), it is better to remove support completely now, than to end up supporting syntax that is not valid and updating a bunch of checks to only have to rewrite the grammar, build the AST, and update the checks all over again once the syntax is finalized.

**Enabling us to start with a clear slate when the new syntax is revealed will help us to deliver support much faster.**

It is doubtful that whatever the new (post JDK 23) String Template syntax will look the same, so we will have to break compatibility with the JDK 22 String Template syntax regardless.

From  https://mail.openjdk.org/pipermail/amber-spec-experts/2024-April/004106.html:

>So, to be clear: there will be no string template feature, even with --enable-preview, in JDK 23.

nrmancuso:
Closed via #15212

romani:
@nrmancuso , if you put both labels , we have release notes build failure.

PR #15192, opened 2024-07-03
supplemental: Drop STRING_TEMPLATE_CONTENT token from IllegalTokenTex…

mahfouz72:
supplemental before #14805


rnveach (approved):

nrmancuso (approved):
Very good, thanks for breaking this up

PR #15199, opened 2024-07-04
Issue #14985: New Check WhenShouldBeUsed

### LINKED ISSUE #14985: Add Check Support for Java 21 Pattern Matching for Switch Syntax: New Check WhenShouldBeUsed
child of #14961: 

pattern matching for switch, allows more concise and readable code when dealing with various object types. This feature includes the `when` keyword to add conditions to case labels directly, improving readability.

I propose to Introduce a new check to enforce the use of the `when` keyword in switch case patterns to replace single if statements, in the case body, enhancing code readability and consistency.

---
**Config**
```XML
$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
    <module name="TreeWalker">
        <module name="WhenShouldBeUsed"/>
    </module>
</module>
```
---
**Example**

```Java
void test(Object response) {
    switch (response) {
        case String s -> {
            if(s.length() > 80) {     // violation
                System.out.println("This string is too long");
            }
         }
        default -> System.out.println("Unknown type");
    }
     switch (response) {
        case String s when s.length() > 80 -> {  // ok
                System.out.println("This string is too long");
         }
        default -> System.out.println("Unknown type");
    }
}
```
**Expected Output**
```

Starting audit...
[ERROR] D:\Test.java:4:12: when can be used instead of single if statement. [WhenShouldBeUsed]
Audit done.
```

mahfouz72:
closes #14985 

New module config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/2029ee1be198c9c838baf28fcd4f2a2fa5282047/check.xml


src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * @param statementList the AST node representing {@code SLIST} of the current case
+     * @return true if the case block has a single if statement, false otherwise
+     */
+    private static boolean hasSingleIfStatementInCaseBlock(DetailAST statementList) {
+        final int[] acceptableStatementsInCaseBlock = {
+            TokenTypes.LITERAL_BREAK,
+            TokenTypes.EMPTY_STAT,
+            TokenTypes.RCURLY,
+        };
```
rnveach, line 139: Please move this to a static final field.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,159 @@
...
+ * </p>
+ * <p>
+ * Java 21 has introduced enhancements for switch statements and expressions
+ * that allow the use of patterns in case labels. The {@code when} keyword is used to specify
+ * condition for a case label, also called as guarded case labels. This syntax is more readable
+ * and concise than the single {@code if} statement inside the pattern match block.
+ * </p>
+ * <p>
+ * See the <a href="https://docs.oracle.com/javase/specs/jls/se22/html/jls-14.html#jls-Guard">
+ *     Java Language Specification</a> for more information about guarded case labels.
```
rnveach, line 43: Please remove extra padding.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
```diff
@@ -0,0 +1,108 @@
...
+
+public class InputWhenShouldBeUsedSwitchExpression {
+
+    void test(Object j, int x) {
+        int a = switch (j) {
+            case Point(int _, int y): // violation, 'When should be used instead of a single if*.'
+                if (y == 0) {
+                    yield 0;
+                }
+            case Integer _: if (x == 0)  yield 4;
```
rnveach, line 17: Maybe I am a bit confused on this syntax, but if you corrected the violations here, would it be the same code?

`Point` will fall through to the Integer if `y != 0`. There would be no fallthrough on a `when` and wouldn't it not hit `Integer` since we triggered as a point?

It seems if you fixed this code, you would need an extra `Point` without a `when` for the fallthrough?

Also, add or point to examples where there is no fall through and we use `:` on case.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
```diff
@@ -0,0 +1,108 @@
...
+                }
+            }
+            // violation below, 'When should be used instead of a single if*.'
+            case ColoredPoint(Point(_, _), String _) : {
+                if (x == 0) yield 4;
+            }
+            default: { yield 3;}
+        };
+        int bb = switch (j) {
+            case Point(int _, int y) when y == 0:  {
```
rnveach, line 34: Add or point to an example where a `when` has a single if statement inside. Will we get a violation?

src/xdocs/checks/coding/whenshouldbeused.xml
```diff
@@ -0,0 +1,122 @@
...
+  <body>
+    <section name="WhenShouldBeUsed">
+      <p>Since Checkstyle 10.18.0</p>
+      <subsection name="Description" id="Description">
+        <p>
+          Ensures that <code>when</code> is used instead of a single
+          <code>if</code> statement inside a case block.
+        </p>
+        <p>
+          Java 21 has introduced enhancements for switch statements and expressions
```
rnveach, line 17: Isn't this a `Rationale`?

rnveach (changes requested):

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
rnveach, line 17: Can we have an example of a violation where case has a `{}` surrounding it?

rnveach (changes requested):

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
mahfouz72: > There would be no fallthrough on a when and wouldn't it not hit Integer since we triggered as a point? 

That's what I expected. I though that there would be no fall through to a guarded case label. But this is wrong we fall through to the next case without stopping on `when`. sounds weird to me. 

```
PS D:\CS\test\src> cat Test.java
public class Test {
    void test(Object j, int x) {
        int a = switch (j) {
            case Point(int _, int y):
                if (y == 0) {
                    yield 0;
                }
            case Integer _:
                if (x == 0) {
                    yield 4;
                }
            default: { yield 3; }
        };
        System.out.println(a);
    }
    void test2(Object j, int x) {
        int b = switch (j) {
            case Point(int _, int y):
                if (y == 0) {
                    yield 0;
                }
            case Integer _ when x == 0:
                    yield 4;
            default: { yield 3; }
        };
        System.out.println(b);
    }
        Test t = new Test();

        t.test(new Point(1, 1), 0); // 4
        t.test2(new Point(1, 1), 0); // 4

        t.test(new Point(1, 1), 1); // 3
        t.test2(new Point(1, 1), 1); // 4
    }

    record Point(int x, int y) { }
}
PS D:\CS\test\src> javac Test.java
PS D:\CS\test\src> java Test
4
4
3
4
```

--- 
after this what do you think on how to proceed with this check do we need to check for fall through now? 

sonarline don't mention how they handle this in their [check](https://rules.sonarsource.com/java/tag/java21/RSPEC-6916/). The examples are with enhanced switch only where we can't have a fall through

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,159 @@
...
+        final int[] acceptableStatementsInCaseBlock = {
+            TokenTypes.LITERAL_BREAK,
+            TokenTypes.EMPTY_STAT,
+            TokenTypes.RCURLY,
+        };
+        boolean hasSingleIfStatementInCaseBlock = false;
+        boolean hasIfBranch = false;
+        DetailAST ast = statementList.getFirstChild();
+
+        while (ast != null) {
```
nrmancuso, line 144: Can we just make a list of siblings to process here? ;)

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
```diff
@@ -66,3 +66,4 @@ unnecessary.semicolon=Unnecessary semicolon.
 unused.local.var=Unused local variable ''{0}''.
 variable.declaration.usage.distance=Distance between variable ''{0}'' declaration and its first usage is {1}, but allowed {2}.
 variable.declaration.usage.distance.extend=Distance between variable ''{0}'' declaration and its first usage is {1}, but allowed {2}.  Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value).
+when.should.be.used=When should be used instead of a single if statement inside a pattern match body.
```
nrmancuso, line 69:
```suggestion
when.should.be.used='when' expression should be used instead of a single 'if' statement in switch rule or case statement.
```

nrmancuso (changes requested):
Few things I've noticed:

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
nrmancuso: @rnveach thoughts?

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
mahfouz72: @nrmancuso please share your thoughts

romani:
Fix name of Check in commit message in in PR

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
nrmancuso: Sounds like we should handle most simple case with -> syntax only

rnveach: @mahfouz72 I recommend downloading and installing Eclipse. Sonarlint comes with or can be downloaded from the marketplace. I am not near a computer able to test it right now.

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
rnveach: I am ok with the general change. I am not sure if there is commonality with other check messages on if we make them technical or keep them in plain English.

nrmancuso:
WhenShouldbeUsed -> WhenShouldBeUsed

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
mahfouz72: @rnveach sonar lint flags this

src/xdocs/checks/coding/whenshouldbeused.xml
mahfouz72: I don't know I felt like this is a general explanation for the syntax and the `Rationale` is the last scentence only in this paragragh. Do you think we should add a `Rationale:` ?

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
mahfouz72: I have few comments here
> 'when' expression 

is this term is worded like this somewhere ? why we have this extra "expression"

>  instead of a single 'if' statement in switch rule or case statement.

this is not always true because the case label must be a pattern to use `when`. So we may have switch rule and single if statement and we shouldn't suggest using `when`

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
mahfouz72: > Will we get a violation?

Yes because you can combine the if condition with when. I added an example for this

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
mahfouz72: https://github.com/checkstyle/checkstyle/blob/d2fb8bada0ea4998b6643a4660b676f76d1c543e/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchRule.java#L98-L111

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done ;).

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
mahfouz72: done

rnveach: I leave it to you and @nrmancuso if we should restrict check like mentioned at https://github.com/checkstyle/checkstyle/pull/15199#discussion_r1671559164 .

sonarlint only violates the `Integer` and not the `Point`? If you resolve the one violation, does it report anything else?
Also your example in the screenshot is not the same, as you encase the `Integer` case with `{}` but not the example above. Does sonarlint only violate this because of the encapsulation or it does it without it?

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
rnveach: Since we violate this case, to me it seems to me that the message is slightly misleading. We say use when instead of if but they are already using when. I am not sure if we should change it, but I am pointing it out.

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
rnveach: > 'when' expression
is this term is worded like this somewhere ?

https://docs.oracle.com/javase/specs/jls/se22/html/jls-14.html#jls-14.11.1
>  A case label with case patterns may have an optional when expression, known as a guard

I thought the same thing when I saw this. :)

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
mahfouz72: > sonarlint only violates the Integer and not the Point

Yes because the case label `Point` with has `{}`. I don't get the point of why they violate it only when we have `{}`

>  If you resolve the one violation, does it report anything else?

nope once you move `if` to `when` then nothing reported

>  Does sonarlint only violate this because of the encapsulation

Yes, it doesn't violate any case that has no `{}` . IDK why may be a bug ?

> : Even though you used {} in the screenshot, because this is using :, this is still a fall through right?

Yes this is a fall-through whether we have curly brackets or not

mahfouz72: Example when we have `{}` with point. fall through still exists because it is `:` not `->`

nrmancuso: There is no fall through in your example since we are using patterns in a switch expression

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
nrmancuso: >this is not always true because the case label must be a pattern to use when. So we may have switch rule and single if statement and we shouldn't suggest using when

hopefully it is always true for code we are placing violations on :)

nrmancuso:
I am good to suppress all checker violations

rnveach:
@nrmancuso I am assigning because of https://github.com/checkstyle/checkstyle/pull/15199#discussion_r1676593624 and if we should switch to only validate if there is no fall throughs.

src/xdocs/checks/coding/whenshouldbeused.xml
rnveach: I think so, Yes. You are explaining the general syntax but it helps lead into the real rationale if they didn't know what the syntax was.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
rnveach: Can we make a post somewhere for those less familar with the syntax when it is a fallthrough and when it is not?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,176 @@
...
+        final DetailAST statementList = getStatementListOfCase(ast);
+        final boolean hasPatternLabel = hasPatternLabel(ast);
+        if (hasPatternLabel) {
+            final List<DetailAST> blockStatements = getBlockStatements(statementList);
+            final boolean hasAcceptableStatementsOnly = blockStatements.stream()
+                .allMatch(WhenShouldBeUsedCheck::isAcceptableStatement);
+            final boolean hasSingleIfWithNoElse = blockStatements.stream()
+                .filter(statement -> {
+                    return statement.getType() == TokenTypes.LITERAL_IF
+                            && statement.findFirstToken(TokenTypes.LITERAL_ELSE) == null;
```
nrmancuso, line 112: Please make this a private method. I suspect that as we add features to this check we will have more conditions in this area, let's establish a good pattern now.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,176 @@
...
+    public static final String MSG_KEY = "when.should.be.used";
+
+    /**
+     *  An array of acceptable statements inside the case block.
+     */
+    private static final int[] ACCEPTABLE_STATEMENTS_INSIDE_CASE_BLOCK = {
+        TokenTypes.LITERAL_IF,
+        TokenTypes.LITERAL_BREAK,
+        TokenTypes.EMPTY_STAT,
+        TokenTypes.RCURLY,
```
nrmancuso, line 83: How is this a statement?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,176 @@
...
+ * <ul>
+ * <li>
+ * {@code when.should.be.used}
+ * </li>
+ * </ul>
+ *
+ * @since 10.18.0
+ */
+
+@FileStatefulCheck
```
nrmancuso, line 67: I don't understand how this check is stateful.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    /**
+     * Get the statement list token of the case block.
+     *
+     * @param caseAST the AST node representing {@code LITERAL_CASE}
+     * @return the AST node representing {@code SLIST} of the current case
+     */
+    private static DetailAST getStatementListOfCase(DetailAST caseAST) {
```
nrmancuso, line 126:
```suggestion
    private static DetailAST getStatementList(DetailAST caseAST) {
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,176 @@
...
+               .orElse(statementList);
+    }
+
+    /**
+     * Get all statements inside the case block.
+     *
+     * @param statementList the AST node representing {@code SLIST} of the current case
+     * @return statements inside the current case block
+     */
+    private static List<DetailAST> getBlockStatements(DetailAST statementList) {
```
nrmancuso, line 141: Ideally, we should never pass in a null DetailAST here. How can we prevent this?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @Override
+    public int[] getRequiredTokens() {
+        return new int[] {TokenTypes.LITERAL_CASE};
+    }
+
+    @Override
+    public void visitToken(DetailAST ast) {
+        final DetailAST statementList = getStatementListOfCase(ast);
```
nrmancuso, line 103: Better to use Optional here to make it clear that this might not be present and how we deal with this case below, in `if (hasPatternLabel)`.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,176 @@
...
+        final boolean hasPatternLabel = hasPatternLabel(ast);
+        if (hasPatternLabel) {
+            final List<DetailAST> blockStatements = getBlockStatements(statementList);
```
nrmancuso, line 106: pseudo code:
```suggestion
            final List<DetailAST> blockStatements = statementList.map(WhenShouldBeUsedCheck::getBlockStatements)
                     .orElseGet(List::new);
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,176 @@
...
+            final boolean hasAcceptableStatementsOnly = blockStatements.stream()
+                .allMatch(WhenShouldBeUsedCheck::isAcceptableStatement);
+            final boolean hasSingleIfWithNoElse = blockStatements.stream()
```
nrmancuso, line 109: Please let each statement breathe in here (empty line between each), we have no reason to be concerned about file length in this check :)

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,176 @@
...
+     * Get all statements inside the case block.
+     *
+     * @param statementList the AST node representing {@code SLIST} of the current case
+     * @return statements inside the current case block
+     */
+    private static List<DetailAST> getBlockStatements(DetailAST statementList) {
+        final List<DetailAST> blockStatements = new ArrayList<>();
+        DetailAST ast = Optional.ofNullable(statementList)
+                .map(DetailAST::getFirstChild)
+                .orElse(null);
```
nrmancuso, line 145: I think Optional usage here is overkill, especially if we don't pass in a null statementList as I've suggested above.

nrmancuso (changes requested):
Second pass:

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
nrmancuso: Conceptually, we are simply following exist Java paradigms; expressions yield a single value

 I think we can simplify things by first adding support only for arrow syntax, then creating an issue to add support for expressions in general to keep things as simple as possible.

@rnveach @mahfouz72 what do you think about this?

mahfouz72: I am good with this. Let's just support `->` only for now. and add support for `:` in a separate issue later because it seems like there are nuances with it that will make us slow.

mahfouz72: > There is no fall through in your example since we are using patterns in a switch expression

actually, [this comment](https://github.com/checkstyle/checkstyle/pull/15199#discussion_r1671306450) shows that there is a fall-through

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: This is not a statement actually but it is a `SLIST` child according to AST. should we change the name of this field?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: Once I add a class filed (`ACCEPTABLE_STATEMENTS_INSIDE_CASE_BLOCK`) in this class. Some test fails and require the check to be FileStateful, not Stateless. It considered this final static array as a state. I didn't understand also it is final so shouldn't consider a state as per my understanding

nrmancuso: Ok, we need to just move this to the private method then, representing this check as stateless is more important than having this declared as a field IMO

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: I just added another condition `hasPatternLabel && statementList != null` if one of them is false we don't need to continue

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: Is it needs to be Optional can't it be the  `DetailAST` and just check for non-nullity before enter if

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: I don't think we need this for now. we added `statementList != null` in the if so we don't continue with a null slist. 

Do you think we still should follow your sugesstion?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done we skip all methods if statementList is null.

src/xdocs/checks/coding/whenshouldbeused.xml
mahfouz72: done

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
rnveach: > first adding support only for arrow syntax

I am good with this too.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: I annotated it with stateless and suppressed it in ImmutabilityTest

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/InputWhenShouldBeUsedSwitchExpression.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,183 @@
...
+    @Override
+    public int[] getRequiredTokens() {
+        return new int[] {TokenTypes.LITERAL_CASE};
+    }
+
+    @Override
+    public void visitToken(DetailAST ast) {
+        final boolean hasPatternLabel = hasPatternLabel(ast);
+        final DetailAST statementList = getStatementList(ast);
+        final boolean isInSwitchRule = ast.getParent().getType() == TokenTypes.SWITCH_RULE;
```
nrmancuso, line 104: Please add `// until ...` comment here

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
nrmancuso: This is good

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,183 @@
...
+        DetailAST ast = statementList.getFirstChild();
+        while (ast != null) {
+            blockStatements.add(ast);
+            ast = ast.getNextSibling();
+        }
+        return Collections.unmodifiableList(blockStatements);
+    }
+
+    /**
+     * Check if the statement is an acceptable statement inside the case block.
```
nrmancuso, line 150: Please explain the concept of an acceptable statement here.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
```diff
@@ -0,0 +1,183 @@
...
+     *
+     * @param ast the AST node representing the statement
+     * @return true if the statement is an acceptable statement, false otherwise
+     */
+    private static boolean isAcceptableStatement(DetailAST ast) {
+        return TokenUtil.isOfType(ast, ACCEPTABLE_STATEMENTS_INSIDE_CASE_BLOCK);
```
nrmancuso, line 156: Please move the collection/array to this method, I don't want to start this check off with suppressions if we can help it. This collection is only used here, there is no need for it to be a field.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
nrmancuso: Just move it to the private method and rename it :)

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/Example1.java
```diff
@@ -0,0 +1,46 @@
...
+      // violation below, 'When should be used instead of a single if*.'
+      case String s -> {
+        if (s.isEmpty()) {
+          System.out.println("empty string");
+        }
+      }
+      default -> {}
+    }
+    switch (o) {
+      case String s -> { // ok, not a single if statement inside the case block
```
nrmancuso, line 26: Please do not use trailing comments in examples

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
nrmancuso: How about `'when' expression should be used instead of a single 'if' statement in switch block.`?

nrmancuso (changes requested):
Looking very good, few more:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
nrmancuso: https://github.com/checkstyle/checkstyle/pull/15199#discussion_r1677206142

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
mahfouz72: good I think you mean case block 
'when' expression should be used instead of a single 'if' statement in case block.

mahfouz72: > this is not always true because the case label must be a pattern to use when. So we may have switch rule and single if statement and we shouldn't suggest using when

but still, this isn't satisfied. Do you think it is ok to not mention this explicit fact in the message that `when` can be used with patterns only ?

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/Example1.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/WhenShouldBeUsedCheck.java
mahfouz72: done

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
nrmancuso: Yes, let’s not mention this in the message.

mahfouz72: done

nrmancuso:
I am prepared to approve after final documentation review and report analysis

@mahfouz72 we are close, please generate reports

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/Example1.java
```diff
@@ -0,0 +1,47 @@
...
+    switch (o) {
+      // violation below,  ''when' expression should be used*.'
+      case String s -> {
+        if (s.isEmpty()) {
+          System.out.println("empty string");
+        }
+      }
+      default -> {}
+    }
+    switch (o) {
```
nrmancuso, line 25:
```suggestion
  
    switch (o) {
```

nrmancuso (changes requested):
Last from me, pending reports

mahfouz72:
Github, generate report

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/whenshouldbeused/Example1.java
mahfouz72: done

src/xdocs/checks/coding/whenshouldbeused.xml
nrmancuso: @mahfouz72 please always share a link to created issues: https://github.com/checkstyle/checkstyle/issues/15270

nrmancuso (approved):
Already looked at the report, meant to approve this yesterday, my apologies. Good stuff.

PR #15210, opened 2024-07-08
infra: prove of concept for gogole formatter

romani:
prove of concept


romani:
Maintainers disagree on this approach that introduces this test dependency, falling back to https://github.com/checkstyle/checkstyle/issues/15236

PR #15211, opened 2024-07-08
Infra google formatter correct

romani:
prove of concept
to show how to we can convert Input fie with violations to correct file and prove that checktyle is not reporting any violation on file passed by formatter

https://github.com/google/google-java-format/releases

file was formatted by:
`rivanov@p5510:/var/tmp$ java -jar google-java-format-1.22.0-all-deps.jar --replace /home/rivanov/java/github/romani/checkstyle/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule43onestatement/InputOneStatementPerLine.java`

 but we will do new file creation, we need to name it same but with "correct" word somewhere. `InputOneStatementPerLineCorrect.java`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -412,6 +412,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>com.google.googlejavaformat</groupId>
+      <artifactId>google-java-format</artifactId>
+      <version>1.22.0</version>
+      <scope>test</scope>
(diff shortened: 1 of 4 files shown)


romani:
Maintainers disagree on this approach that introduces this test dependency, falling back to https://github.com/checkstyle/checkstyle/issues/15236

PR #15213, opened 2024-07-08
Issue #14937: Removed PER_MODULE_TESTS_RULES_LIST from XdocsPagesTest

### LINKED ISSUE #14937: Migrate google style guide tests to follow chapter wise testing

Zopsss:
Closes #14937 

Removed the `PER_MODULE_TESTS_RULES_LIST` used in migration of sections to chapter wise testing

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
@@ -226,14 +226,6 @@ public class XdocsPagesTest {
     private static final Set<String> GOOGLE_MODULES = Collections.unmodifiableSet(
         CheckUtil.getConfigGoogleStyleModules());
 
-    // Contains all the sections which are not migrated to chapter wise testing
-    // in google style documentation.
-    // This list will be removed once all the sections are migrated.
-    // until https://github.com/checkstyle/checkstyle/issues/14937
-    private static final Set<String> PER_MODULE_TESTS_RULES_LIST = Set.of(
-            "4.8.5 Annotations"
-    );
-
     /**
      * Generate xdoc content from templates before validation.
      * This method will be removed once
@@ -1781,14 +1773,13 @@ public class XdocsPagesTest {
             Set<String> styleChecks, String styleName, String ruleName) {
         final Iterator<Node> itrChecks = checks.iterator();
         final Iterator<Node> itrConfigs = configs.iterator();
-        final boolean isRegularDocumentation = "sun".equals(styleName)
-               || "google".equals(styleName) && PER_MODULE_TESTS_RULES_LIST.contains(ruleName);
+        final boolean isGoogleDocumentation = "google".equals(styleName);
 
-        if (isRegularDocumentation) {
-            validateModuleWiseTesting(itrChecks, itrConfigs, styleChecks, styleName, ruleName);
+        if (isGoogleDocumentation) {
+            validateChapterWiseTesting(itrChecks, itrConfigs, styleChecks, styleName, ruleName);
         }
         else {
-            validateChapterWiseTesting(itrChecks, itrConfigs, styleChecks, styleName, ruleName);
+            validateModuleWiseTesting(itrChecks, itrConfigs, styleChecks, styleName, ruleName);
         }
 
         assertWithMessage(styleName + "_style.xml rule '" + ruleName + "' has too many configs")


src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -226,14 +226,6 @@ public class XdocsPagesTest {
     private static final Set<String> GOOGLE_MODULES = Collections.unmodifiableSet(
         CheckUtil.getConfigGoogleStyleModules());
 
-    // Contains all the sections which are not migrated to chapter wise testing
-    // in google style documentation.
-    // This list will be removed once all the sections are migrated.
-    // until https://github.com/checkstyle/checkstyle/issues/14937
```
rnveach, line 232: Looks like this shouldn't be a minor commit. I assume this closes the issue also, so the first post should say so.

rnveach (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
romani: Prefix should be referencing #14937 .

This PR will close this issue

rdiachenko (approved):
lgtm

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
romani: We have strange 4.8.5.2

https://github.com/checkstyle/checkstyle/blob/b16e96a757ccee5be66fabb8e666823ca3cb558a/src/it/java/com/google/checkstyle/test/chapter4formatting/rule4852classannotations/ClassAnnotationsTest.java#L36
Vs
https://checkstyle.org/google_style.html#a4.8.5

@Zopsss , please clarify what is going on

Zopsss:
GitHub, generate site

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
Zopsss: > We have strange 4.8.5.2

4.8.5.2 was converted to whole config testing, I took this section as an example when I was implementing `verifyWithWholeConfig` to show that how this method works in our tests. 

> Vs
> https://checkstyle.org/google_style.html#a4.8.5

This section was divided into smaller sections like 4.8.5.1, 4.8.5.2, 4.8.5.3 so there is no test for 4.8.5 but there are tests for its subsections, here's the how current website looks like:

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a0c4884_2024024517/google_style.html#a4.8.5

Zopsss: > Prefix should be referencing https://github.com/checkstyle/checkstyle/issues/14937 .

Done, renamed the commit and PR

romani: Ok, please create 4..8.5.1 Type-use annotations in table.

We need all paragraphs.
Please recheck that all paragraphs are present.

rnveach: @romani Depending on what you consider a paragraph to be present, there could be more missing.

For example, we have 2.2 and then 2.3.1. While if you look at other similar numbered sections, we have 3.3 and 3.3.1 . It looks like we are missing 2.3 in this instance.
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a0c4884_2024024517/styleguides/google-java-style-20220203/javaguide.html#s2.3-special-characters

7.3.3 is also missing.

5.1's text mentions 5.2, but it's text is different from the label. `of` versus `by`.

I see these as a separate issue and should not be handled in this PR.

rnveach (approved):

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
romani: I agree, will do separately https://github.com/checkstyle/checkstyle/issues/15217

romani (approved):

Issue #15214, opened 2024-07-08
Migrate google style guide tests to follow whole config testing approach


Zopsss:
In #14937, all the google style guide tests were Migrated to `Chapter Wise Testing` approach. 

The goal of this issue is to migrate all the sections to follow whole config testing approach. In this type of approach, the input file will be tested against the whole google configuration and ensure that we follow all the rules of google style guide in our input files.

Example of problem with not-whole config validation - https://github.com/checkstyle/checkstyle/pull/15248#discussion_r1676836582

Following sections needs to be migrated:
- [x] 2.1 File name
- [x] 2.3.1 Whitespace characters
- [x] 2.3.2 Special escape sequences
- [x] 2.3.3 Non-ASCII characters
- [x] 3 Source file structure
- [x] 3.2 Package statement
- [x] 3.3.1 No wildcard imports
- [x] 3.3.2 No line-wrapping
- [x] 3.3.3 Ordering and spacing
- [x] 3.4.1 Exactly one top-level class declaration
- [x] 3.4.2.1 Overloads: never split
- [x] 4.1.1 Use of optional braces
- [x] 4.1.2 Nonempty blocks: K & R style
- [x] 4.1.3 Empty blocks: may be concise
- [x] 4.2 Block indentation: +2 spaces
- [x] 4.3 One statement per line
- [x] 4.4 Column limit: 100
- [x] 4.5.1 Where to break
- [x] 4.5.2 Indent continuation lines at least +4 spaces
- [x] 4.6.1 Vertical Whitespace
- [x] 4.6.2 Horizontal whitespace
- [x] 4.8.2.1 One variable per declaration
- [x] 4.8.2.2 Declared when needed
- [x] 4.8.3.2 No C-style array declarations
- [x] 4.8.4.1 Indentation
- [x] 4.8.4.2 Fall-through: commented
- [x] 4.8.4.3 Presence of the default label
- [x] 4.8.5.3 Methods And Constructors Annotations
- [x] 4.8.5.4 Field Annotations
- [x] 4.8.6.1 Block comment style
- [x] 4.8.7 Modifiers
- [x] 4.8.8 Numeric Literals
- [x] 5.2.1 Package names
- [x] 5.2.2 Class names
- [x] 5.2.3 Method names
- [x] 5.2.5 Non-constant field names
- [x] 5.2.6 Parameter names
- [x] 5.2.7 Local variable names
- [x] 5.2.8 Type variable names
[...]

PR #15221, opened 2024-07-09
Issue #15214: migrated section 3.3.3 Ordering and spacing to whole config testing

Zopsss:
#15214


romani (approved):
Even I see bunch of changes in Inputs
I looked through all of them, it is ok to be less crazy over import `.*` it is not primarily point of testing.

Issue #15231, opened 2024-07-10
Create new GitHub  workflow similar to diff-report.yml but to use test-configs repository configs 


relentless-pursuit:
Rather than 'Github, generate report' we move to specific check.  And, the testc-config's README.md would be adapted likewise
Example:
`Github, generate report for SummaryJavadoc/Example1`

Existing format used to run regression reports from the comment section:
From https://github.com/checkstyle/test-configs/blob/1e98f67239a38c969893ef095b5026a4ad42a6d2/SummaryJavadoc/Example1/README.md

romani:
merging this to https://github.com/checkstyle/checkstyle/issues/15234

PR #15232, opened 2024-07-10
New workflow like diff-report.yml to use test-configs repo

relentless-pursuit:
Resolves: #15231

### DIFF
--- a/.github/workflows/diff-report.yml
+++ b/.github/workflows/diff-report.yml
@@ -97,6 +97,36 @@ jobs:
           # shellcheck disable=SC2005
           echo "$(cat "$GITHUB_OUTPUT")"
 
+      - name: Extract config name
+        id: extract_config_name
+        run: |
+          CONFIG_NAME=$(grep -oP '(?<=Diff Regression config: ).*?(?=/config.xml)' .ci-temp/text | awk -F'/' '{print $(NF-1)"/"$NF}')
+          echo "config_name=$CONFIG_NAME" >> $GITHUB_OUTPUT
+          echo "Extracted config name: $CONFIG_NAME"
+
+      - name: Extract check and example
+        id: extract_check_example
+        run: |
+          CHECK_EXAMPLE=$(echo "${{ github.event.comment.body }}" | grep -oP '(?<=Github, generate report ).*')
+          echo "check_example=$CHECK_EXAMPLE" >> $GITHUB_OUTPUT
+          echo "Extracted CHECK_EXAMPLE: $CHECK_EXAMPLE"
+          if [ -z "$CHECK_EXAMPLE" ]; then
+            echo "No valid check example found. Exiting."
+            exit 1
+          fi
+
+      - name: Validate check example
+        run: |
+          CONFIG_NAME="${{ steps.extract_config_name.outputs.config_name }}"
+          CHECK_EXAMPLE="${{ steps.extract_check_example.outputs.check_example }}"
+          if [ "$CONFIG_NAME" != "$CHECK_EXAMPLE" ]; then
+            echo "Error: The check example in the comment does not match the configuration in the PR description."
+            echo "PR Config: $CONFIG_NAME"
+            echo "Comment Example: $CHECK_EXAMPLE"
+            exit 1
+          fi
+          echo "Check example matches configuration."
+
   make_report:
     runs-on: ubuntu-latest
     needs: parse_body


romani (changes requested):
New workflow should be created.

Please share links to PRs how it works in your repository.
Workflows can we executed only after merge, so PR doesn't test workflow.

relentless-pursuit:
> New workflow should be created.
> 
> Please share links to PRs how it works in your repository. Workflows can we executed only after merge, so PR doesn't test workflow.



> New workflow should be created.
> 
> Please share links to PRs how it works in your repository. Workflows can we executed only after merge, so PR doesn't test workflow.

We can. Here is the link for the PR used for testing. Pardon me it is quite messy. I had created this org and sent inviation. I am not sure if you have accepted the invitation or not.

https://github.com/checkstyle-gsoc/checkstyle/pull/2

also, adding the link to workflow.
https://github.com/checkstyle-gsoc/checkstyle/blob/master/.github/workflows/experiment-automate-regression.yml

The only change in workflow is the report name, my AWS account and some repo change as i am doing under my org. Hence, had to tweak the workflow to understand my forked branches.

romani:
good, please create new workflow file.

.github/workflows/diff-report.yml
```diff
@@ -97,6 +97,36 @@ jobs:
           # shellcheck disable=SC2005
           echo "$(cat "$GITHUB_OUTPUT")"
 
+      - name: Extract config name
+        id: extract_config_name
+        run: |
+          CONFIG_NAME=$(grep -oP '(?<=Diff Regression config: ).*?(?=/config.xml)' .ci-temp/text | awk -F'/' '{print $(NF-1)"/"$NF}')
+          echo "config_name=$CONFIG_NAME" >> $GITHUB_OUTPUT
+          echo "Extracted config name: $CONFIG_NAME"
```
romani, line 105: what for we have this block?
we already print all variables where links are stored.

.github/workflows/diff-report.yml
```diff
@@ -97,6 +97,36 @@ jobs:
...
+
+      - name: Extract check and example
+        id: extract_check_example
+        run: |
+          CHECK_EXAMPLE=$(echo "${{ github.event.comment.body }}" | grep -oP '(?<=Github, generate report ).*')
```
romani, line 110: we need trim spaces right in comment content.
We need to have trigger phrase to be `Github, generate report for `.
To get Check name , we need to change comment by replace of `Github, generate report for ` with ``.
all other content is path in test-config repo.

.github/workflows/diff-report.yml
```diff
@@ -97,6 +97,36 @@ jobs:
...
+          echo "check_example=$CHECK_EXAMPLE" >> $GITHUB_OUTPUT
```
romani, line 111: when you do echo, do it for humans.
`echo "Check Example=..."`
or exactly varaible name - tracing.
` echo "CHECK_EXAMPLE=$CHECK_EXAMPLE"`

.github/workflows/diff-report.yml
```diff
@@ -97,6 +97,36 @@ jobs:
...
+            echo "No valid check example found. Exiting."
+            exit 1
+          fi
+
+      - name: Validate check example
+        run: |
+          CONFIG_NAME="${{ steps.extract_config_name.outputs.config_name }}"
+          CHECK_EXAMPLE="${{ steps.extract_check_example.outputs.check_example }}"
+          if [ "$CONFIG_NAME" != "$CHECK_EXAMPLE" ]; then
+            echo "Error: The check example in the comment does not match the configuration in the PR description."
```
romani, line 123: this is not required.
As we reusing existing workflow code, we should just keep all existing abilities to parse PR body but if we detect value in CHECK_EXAMPLE we use only it and all other variables are set to ``.

We can do use `Github, generate report for PR description links`. So if CHECK_EXAMPLE is equal `PR description links` we will use old location of links in PR description.

romani (changes requested):
item:

.github/workflows/diff-report.yml
relentless-pursuit: So, whenever a user is trying to put configs in PR description, i am extrating the modulename and comparing with the github comment, if they match, only proceed otherwise the report fails.

.github/workflows/diff-report.yml
relentless-pursuit: if the whole idea is to get rid of this workflow, then why bother about changing anything. Because, won't it create inconvenience?

romani: Existing workflow in in high usage, we can not introduce defects in it.
Making copy, will let us ability to merge, test thoughtly and substitute only when we are confident.

.github/workflows/diff-report.yml
romani: We don't need this at all.
Let's skip handling any content from PR description for now

romani:
Closing in favor of https://github.com/checkstyle/checkstyle/pull/15235

PR #15212, opened 2024-07-08
Issue #14805: Remove support for string templates syntax

mahfouz72:
closes #14805

-----------------------

### Reports
Projects file: projects-to-test-on.properties
Patch branch: pull-15212

https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/pull-15212/2024-07-10-T-00-20-42/part1/index.html

https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/pull-15212/2024-07-10-T-00-20-42/part4/index.html

https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/pull-15212/2024-07-10-T-00-20-42/part5/index.html

https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/pull-15212/2024-07-10-T-00-20-42/part3/index.html

https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/pull-15212/2024-07-10-T-00-20-42/part2/index.html

https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/pull-15212/2024-07-10-T-00-20-42/part6/index.html

https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/pull-15212/2024-07-10-T-00-20-42/sevntu-check-regression_part_1/index.html

https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/pull-15212/2024-07-10-T-00-20-42/sevntu-check-regression_part_2/index.html

All previous reports are good. differences are:
-  removal of violations in files that have string template
-  line numbers in exception messages ([VarargsAndReceiver](https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/pull-15212/2024-07-10-T-00-20-42/sevntu-check-regression_part_2/openjdk21/index.html#A195) file)

https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/pull-15212/2024-07-10-T-00-20-42/antlr-report-checkstyle/index.html

Report is good difference are:
- String templates trees are removed and printing exception message instead
- line numbers in exception messages

I just don't understand [this](https://checkstyle-reports.s3.us-east-2.amazonaws.com/reports/pull-15212/2024-07-10-T-00-20-42/antlr-report-checkstyle/checkstyle/index.html#A2220). Is this expected? it is still stackoverflow but the difference in messages is not just line numbers

PR #15245, opened 2024-07-13
Issue #15241: Reduce baseline execution time value

### LINKED ISSUE #15241: Reduce performance threshold value
Don't forget to downgrade value in performance threshold.

_Originally posted by @romani in https://github.com/checkstyle/checkstyle/issues/15212#issuecomment-2221876446_

mahfouz72:
closes #15241 

new baseline comes from the average execution time after string template removal  https://github.com/checkstyle/checkstyle/pull/15212#issuecomment-2221074819

### DIFF
--- a/.ci/check-performance-regression.sh
+++ b/.ci/check-performance-regression.sh
@@ -5,7 +5,7 @@ set -e
 # max difference tolerance in %
 THRESHOLD_PERCENTAGE=10
 # baseline of execution time in seconds
-BASELINE_SECONDS=415.72
+BASELINE_SECONDS=384.02
 
 # JDK version
 JDK_VERSION=17
@@ -64,9 +64,12 @@ compare_results() {
         echo "Missing EXECUTION_TIME_SECONDS as an argument."
         exit 1
     fi
-  # Calculate percentage difference for execution time
+  # Calculate absolute percentage difference for execution time
   local DEVIATION_IN_SECONDS=$(echo "scale=4; \
     ((${EXECUTION_TIME_SECONDS} - ${BASELINE_SECONDS}) / ${BASELINE_SECONDS}) * 100" | bc)
+  DEVIATION_IN_SECONDS=$(echo "scale=4; \
+    if ($DEVIATION_IN_SECONDS < 0) -($DEVIATION_IN_SECONDS) else $DEVIATION_IN_SECONDS" | bc)
+
   echo "Execution Time Difference: $DEVIATION_IN_SECONDS%"
 
   # Check if differences exceed the maximum allowed difference


rnveach (approved):
If we don't want to forget to update this, we should abs the `DEVIATION_IN_SECONDS` so it can't be negative.

romani (approved):
@rnveach , gas good idea to enhance this script

nrmancuso (changes requested):
Please do https://github.com/checkstyle/checkstyle/pull/15245#pullrequestreview-2176067278

mahfouz72:
but we can  get negative percentage in normal circumstances it is an average time no guarantee that it will be always greater than the Baseline 

example: https://github.com/checkstyle/checkstyle/actions/runs/9910196565/job/27380011798?pr=15199

rnveach:
> we can get negative percentage in normal circumstances it is an average time so no guarantee that it will be always greater than the Baseline

Just like we can high positives in "normal" circumstances. We have no guarantees it will be within the threshold and won't randomly take longer for unexplained reasons. We could hit some PC lag for random reasons.

rnveach:
> example: https://github.com/checkstyle/checkstyle/actions/runs/9910196565/job/27380011798?pr=15199
> Difference is within the maximum allowed difference (-.8100%      <= 10%).

This actually proves we have no problems using abs normally. Even +0.81% (8/10s of a percentage) is still less than 10% and wouldn't falsely trigger.

mahfouz72:
Make sense, I got it. After taking the abs value now it will trigger:
-  if it goes below the threshold on the negative side. An example of this would be the string template removal PR, performance improved so we go negative. (CI will be red so we need to decrease the baseline)
- if it goes above the threshold on the positive side. An example of this would be a syntax update PR that affects performance badly  (CI will be red so we need to redo PR to think about how to optimize parsing)

We will determine if it is case 1 or 2 based on the PR activity either removal or introducing syntax.

@nrmancuso @rnveach what do you feel about decreasing the threshold percentage for example from 10% to 7%

rnveach:
> what do you feel about decreasing the threshold percentage for example from 10% to 7%

I am not sure what analysis went into for creating these numbers. @nrmancuso was mostly incontrol of that.

I would like to know what the values have generally been. Like the average of the last 10-20 runs. If we can gather that from master runs (I assume this runs all the time), then we can use those as our basis. Your example showed <1%, so I wouldn't be too surpirsed if it could be low, however, we need to also take into account if we want to support local runs which could be faster and slower.

mahfouz72:
`Failed to execute goal org.pitest:pitest-maven:1.16.1:mutationCoverage (default-cli) on project checkstyle: Execution default-cli of goal org.pitest:pitest-maven:1.16.1:mutationCoverage failed: Licence expired on 2024-06-13 and grace period ended.`

romani:
Let me contact them to renew

mahfouz72:
Yes only pitest

nrmancuso (approved):

PR #15102, opened 2024-06-22
I15089 windows system properties

romani:
Closes #15089

PR #15255, opened 2024-07-14
doc: remove Localization section from system properties web page

romani:
taken from https://github.com/checkstyle/checkstyle/pull/15102

just noticed, that content is irrelevan for system properties.

https://checkstyle.org/checks/naming/abbreviationaswordinname.html#Violation_Messages

last list very clearly shows where are translations.

### DIFF
--- a/src/xdocs/config_system_properties.xml
+++ b/src/xdocs/config_system_properties.xml
@@ -16,15 +16,6 @@
       </macro>
     </section>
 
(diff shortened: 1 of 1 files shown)


rnveach (approved):
> last list very clearly shows where are translations.

I am not seeing anything mention translations. It mentions message customization which is not the same as translation.

nrmancuso (approved):

PR #15257, opened 2024-07-14
Issue #13999: Remove the suppression for AbstractJavadocCheck

Rohanraj123:
Related #13999

### DIFF
--- a/config/pitest-suppressions/pitest-javadoc-suppressions.xml
+++ b/config/pitest-suppressions/pitest-javadoc-suppressions.xml
@@ -13,15 +13,6 @@
     <sourceFile>AbstractJavadocCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheck</mutatedClass>
     <mutatedMethod>walk</mutatedMethod>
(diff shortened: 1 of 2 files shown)


Rohanraj123:
Github. generate website

romani:
to fix pitest failure , just rebase

Rohanraj123:
@romani I want you to see the workflow explained by @nrmancuso 
[]([Here](https://github.com/checkstyle/checkstyle/issues/12341))

It says if tests fails after hardcoding the mutation then hardcoded mutation might be wrong. But if some tests are failing due to mutation then that means test cases are detecting the changes and that means mutations should be killed. .

Even in my case. [Here](https://github.com/checkstyle/checkstyle/actions/runs/9933675425/job/27436811304?pr=15257)

When i change it from true to false some tests fails. that means mutations should be killed but it still survived .

romani:
If you mutating manually and test is failing, it means that you mutating wrong place and need to attentively read mutation or experiment on different mutation to be confident what mutation survived, sometimes it is not obvious.

romani:
@Rohanraj123, do you have energy to finish this PR ?

romani:
We lost connection to contributor

PR #8697, opened 2020-08-15
Issue #8662: Switch expression syntax check update for EmptyBlockCheck

nmancus1:
Issue #8662: Switch expression syntax check update for EmptyBlockCheck

Diff Regression projects: https://gist.githubusercontent.com/nmancus1/e3988f8092d919b5cbe70f5359c4d9e8/raw/a0f6f5860f8025c19868061dee6e0e40960b992f/projects-to-test-on.properties

Diff Regression config: https://gist.githubusercontent.com/nmancus1/658282e07a843a40b8136e75c3e95ecc/raw/a0fe890093bb0512ec3f833533da90ce6c8212b8/EmptyBlock.xml

Issue #8662, opened 2020-08-11
Java 14 SwitchExpression Check Support validation: EmptyBlock


nmancus1:
Child of #8658 
Check documentation: https://checkstyle.sourceforge.io/config_blocks.html#EmptyBlock

_Checks for empty blocks. This check does not validate sequential blocks._

```
➜  src cat TestClass.java 
public class TestClass {

    int howMany1(int k) {
        switch (k) {
            case 1:

            case 2:

            case 3:

            default:

        }
        return k;
    }

    int howMany2(int k) {
        return switch (k) {
            case 1 -> {

            }
            case 2 -> {

            }
            case 3 -> {

            }
            default -> {
                yield k;
            }
        };
    }

}                                                                                                                             
➜  src /usr/lib/jvm/java-14-openjdk/bin/javac --enable-preview --source 14 TestClass.java                               
TestClass.java:21: error: switch rule completes without providing a value
            }
            ^
  (switch rules in switch expressions must either provide a value or throw)
TestClass.java:24: error: switch rule completes without providing a value
            }
            ^
  (switch rules in switch expressions must either provide a value or throw)
TestClass.java:27: error: switch rule completes without providing a value
            }
            ^
  (switch rules in switch expressions must either provide a value or throw)
3 errors

```
Switch rules must provide a definite value, therefore the block of a switch rule cannot be empty.  We need to check for no false positives:

```
➜  src /usr/lib/jvm/java-14-openjdk/bin/javac --enable-preview --source 14 TestClass.java                               
➜  src cat config.xml                                                                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="TreeWalker">
        <module name="EmptyBlock">
            <property name="tokens" value="LITERAL_DEFAULT,LITERAL_CASE"/>
        </module>
    </module>
</module>
➜  src cat TestClass.java                                                                                               
public class TestClass {

    int howMany1(int k) {
        switch (k) {
            case 1:{} // violation

            case 2:{} // violation

            case 3:{} // violation

        }
        return k;
    }

    int howMany2(int k) {
        return switch (k) {
            case 1 -> {
                yield 2;
            }
            case 2 -> {
                yield 3;
            }
            case 3 -> {
                yield 4;
            }
            default -> {
                yield k;
[...]

PR #3918, opened 2017-03-04
Issue #3839 : EmptyBlock: can't get violation from case token

ps-sp:
Issue #3839 : Added special checks in EmptyBlockCheck.java for LITERAL_CASE to detect related empty blocks.

PR #15273, opened 2024-07-15
Issue #15029: Update EmptyBlockCheck to Violate Empty Case and Default Blocks wit…

### LINKED ISSUE #15029: Update EmptyBlockCheck to Violate Empty Case Blocks with Arrow Syntax

I have read check documentation: https://checkstyle.org/checks/blocks/emptyblock.html#EmptyBlock

```
PS D:\CS\test> javac  src/Test.java       
 PS D:\CS\test> cat  src/Test.java         

public class Test {

    void test(Object obj) {
        int y = 0;
        switch (obj) {
            case ColoredPoint(int x, _, _) when(x >= 2) -> { }  // expected violation
            case ColoredPoint(int x, _, _) -> { }  // expected violation
            default -> { }  // expected violation
        }
        switch (obj) {
            case ColoredPoint(int x, _, _) when(x >= 2) : { } break;  // violation
            case ColoredPoint(int x, _, _) : { } // violation
            default : { }  // violation
        }
    }
}   
PS D:\CS\test> cat config.xml                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="EmptyBlockCheck">
            <property name="tokens" value="LITERAL_CASE, LITERAL_DEFAULT "/>
        </module>
    </module>
</module>
Starting audit...
[ERROR] D:\CS\test\src\Test.java:14:59: Must have at least one statement. [EmptyBlock]
[ERROR] D:\CS\test\src\Test.java:15:46: Must have at least one statement. [EmptyBlock]
[ERROR] D:\CS\test\src\Test.java:16:23: Must have at least one statement. [EmptyBlock]
Audit done.
PS D:\CS\test> 

```

**Describe what you want in detail**

nothing in the doc says that switch rule blocks (arrow syntax) are not validated also It can be empty so I am expecting a violation on empty blocks in the switch rule

mahfouz72:
closes #15029 

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/8ee6d2081667934094dc86562f74ccc96f54bfb3/check.xml


romani:
Please consider Checker violation.

mahfouz72:
We usually suppress such violations. Isn’t this acceptable?

mahfouz72:
Github, generate report

romani:
>We usually suppress such violations. Isn’t this acceptable?

Yes and no :). We have it to not grow list of problems.

https://github.com/checkstyle/checkstyle/actions/runs/9942558796/job/27464302193?pr=15273#step:6:1718

>
```<fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java</fileName>
    <specifier>introduce.eliminate</specifier>
    <message>It is bad style to create an Optional just to chain methods to get a non-optional value.</message>
    <lineContent>.orElse(null);</lineContent>
```

@nrmancuso , will make decision to fix or extend suppression.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/InputEmptyBlockSwitchExpressions.java
```diff
@@ -113,9 +113,9 @@ void howMany7(Nums k) {
     void howMany8(Nums k) {
         switch (k) {
             case ONE -> System.out.println("case two");
-            case TWO, THREE -> {} // ok, empty blocks are allowed for cases
+            case TWO, THREE -> {} // violation, 'Must have at least one statement'
```
rnveach, line 116: Text was added in https://github.com/checkstyle/checkstyle/issues/8662 . `CASE` was a token of the check then, so I am slightly confused why we originally said this was ok to be allowed.

@nrmancuso When you review.

rnveach:
CI is failing.

It is not clear from from inputs which token this flagging under now. Is this from case / default as talked about at https://github.com/checkstyle/checkstyle/issues/15029#issuecomment-2177544141?

Regression was only run on default. It seems to me we need regression on case as well.

Since we ran regression on default, we have a spot at https://checkstyle.org/checks/blocks/emptyblock.html#Example3-config for example we can contribute to for this new update.

===

https://checkstyle.org/checks/blocks/emptyblock.html#Description
> This check does not validate sequential blocks. 

What is a sequential block?

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/InputEmptyBlockSwitchExpressions.java
mahfouz72: > I am slightly confused why we originally said this was ok to be allowed.

me too,  and in the same file ` case : { } ` is a violation

It was added based on this review https://github.com/checkstyle/checkstyle/pull/8697#discussion_r471072392

> This behavior is (poorly) [documented](https://checkstyle.sourceforge.io/config_blocks.html#EmptyBlock), let's leave it as it is.

rnveach: We don't have to have an exhaustive list in these cases, do we? I could see those not needing anything in the blocks but I don't think adding documentation would be a bad idea.

mahfouz72:
github, generate report

mahfouz72:
> It is not clear from from inputs which token this flagging under now. Is this from case / default as talked about at https://github.com/checkstyle/checkstyle/issues/15029#issuecomment-2177544141?
 
 This PR should fix `case` and `default` with arrow syntax any other token shouldn't be affected.
 
 > What is a sequential block?
 
 I have no idea. The doc for this check is not good should we update it? in another PR? 
 
 > Also, no violations for fallthrough

  I don't get this also do you understand what is trying to say?
  
 also, we have some incorrect renderring of tags: `Verification empty block is done for single nearest {@code case} or {@code default}.
 `

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/InputEmptyBlockSwitchExpressions.java
mahfouz72: > We don't have to have an exhaustive list in these cases

No, this is true in switch expressions only.

> I don't think adding documentation would be a bad idea.

Updating example 3 with some cases and default with arrow syntax is good

mahfouz72:
github, generate report

mahfouz72:
github, generate report

mahfouz72:
Reports:

1. `LITERAL_DEFAULT` : https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/9179e85_2024173139/reports/diff/index.html

- violation added for empty default blocks with arrow syntax (expected)

2. `LITERAL_CASE` : https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/9179e85_2024165045/reports/diff/index.html

- violation added for empty case blocks with arrow syntax (expected)

3. All other tokens : https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/9179e85_2024165045/reports/diff/index.html

- no difference (expected)

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/InputEmptyBlockSwitchExpressions.java
mahfouz72: Updated example 3 with more cases and default example with `->` and `:`

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/Example3.java
```diff
@@ -0,0 +1,42 @@
...
+
+    switch (a) {
+      case 1 -> { } // violation, 'Must have at least one statement'
+      default -> { } // violation, 'Must have at least one statement'
+    }
+
+    switch (a) {
+      case 1 : { }  // violation, 'Must have at least one statement'
+      default : { }  // violation, 'Must have at least one statement'
+    }
```
nrmancuso, line 32: Let me play devils advocate here.

Suppose that you are a developer that wrote the second case here:
```java
    switch (a) {
      case 1 : { }  // violation, 'Must have at least one statement'
      default : { }  // violation, 'Must have at least one statement'
    }
```

You can "fix" these violations like this:
```java
    switch (a) {
      case 1 :
      default :
    }
```
Now suppose you wrote this:
```java
    switch (a) {
      case 1 -> { } // violation, 'Must have at least one statement'
      default -> { } // violation, 'Must have at least one statement'
    }
```

How would you fix this code?

mahfouz72: I would completely remove the case label. 
Let me play in reverse and consider this similar case.

```Java
        int x = 0;
        if (x == 1) {   // violation
          
        } else {  // violation

        }
```
How would you fix this code? you can't remove the `{}` but you will remove the if statement.

mahfouz72: @nrmancuso ping

nrmancuso: After playing around with this locally, I don’t think the compiler will let us have a case where we would be in the position to place a violation on code that would fail compilation.

I was concerned that we would place a violation on code, user would remove it, then the code would not compile. I do not think this is possible.

I am good here.

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
```diff
@@ -271,21 +272,25 @@ private boolean checkIsAllLinesAreWhitespace(int lineFrom, int lineTo) {
...
+            leftCurly = parent.findFirstToken(TokenTypes.SLIST);
         }
-        else if (slistAST == null) {
-            leftCurly = ast.findFirstToken(TokenTypes.LCURLY);
+        else if (parentType == TokenTypes.CASE_GROUP) {
+            final DetailAST statementList = ast.getNextSibling();
+            leftCurly = Optional.ofNullable(statementList)
+                         .map(DetailAST::getFirstChild)
+                         .filter(node -> node.getType() == TokenTypes.SLIST)
+                         .orElse(null);
```
nrmancuso, line 287: I think this is not a great usage of Optional (defaulting to a null value); it would make more sense for this method to return an Optional, if anything. How can we improve this?

nrmancuso (changes requested):
Items:

config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
```diff
@@ -1679,6 +1679,31 @@
...
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java</fileName>
+    <specifier>introduce.eliminate</specifier>
+    <message>It is bad style to create an Optional just to chain methods to get a non-optional value.</message>
+    <lineContent>.orElseGet(() -&gt; ast.findFirstToken(TokenTypes.LCURLY));</lineContent>
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
```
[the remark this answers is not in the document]
mahfouz72: done

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/Example3.java
```diff
@@ -0,0 +1,42 @@
...
+</module>
+*/
+
+//non-compiled with javac: Compilable with Java17
+package com.puppycrawl.tools.checkstyle.checks.blocks.emptyblock;
+
+// xdoc section -- start
+public class Example3 {
+
+  private void test(int a) {
```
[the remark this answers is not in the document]
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
```diff
@@ -175,8 +176,9 @@ public int[] getRequiredTokens() {
 
     @Override
     public void visitToken(DetailAST ast) {
-        final DetailAST leftCurly = findLeftCurly(ast);
-        if (leftCurly != null) {
+        final Optional<DetailAST> lcurlyOptional = findLeftCurly(ast);
```
nrmancuso, line 179: Please never name Optional variables as `Optional...`. We force the developer to think about the semantics of this type by contract.

```suggestion
        final Optional<DetailAST> leftCurly = findLeftCurly(ast);
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
```diff
@@ -270,23 +272,28 @@ private boolean checkIsAllLinesAreWhitespace(int lineFrom, int lineTo) {
...
+        if (parentType == TokenTypes.SWITCH_RULE) {
+            lcurlyOptional = Optional.ofNullable(parent.findFirstToken(TokenTypes.SLIST));
         }
-        else if (slistAST == null) {
-            leftCurly = ast.findFirstToken(TokenTypes.LCURLY);
+        else if (parentType == TokenTypes.CASE_GROUP) {
+            final DetailAST statementList = ast.getNextSibling();
+            lcurlyOptional = Optional.ofNullable(statementList)
+                         .map(DetailAST::getFirstChild)
+                         .filter(node -> node.getType() == TokenTypes.SLIST);
```
nrmancuso, line 287: This can be one assignment, I am not sure what we gain by having this extra variable declaration

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
```diff
@@ -270,23 +272,28 @@ private boolean checkIsAllLinesAreWhitespace(int lineFrom, int lineTo) {
...
         }
         else {
-            leftCurly = slistAST;
+            DetailAST leftCurly = ast.findFirstToken(TokenTypes.SLIST);
+            if (leftCurly == null) {
```
nrmancuso, line 291: This can be separated into another `else if`, please do this (via Optional#of since we will now have a not null check); also make a comment in each branch here about what type we expect in each.

nrmancuso:
```
        else if (ast.findFirstToken(TokenTypes.SLIST) != null) {
            lcurlyOptional = Optional.of(ast.findFirstToken(TokenTypes.SLIST));
        }
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
```diff
@@ -270,23 +272,28 @@ private boolean checkIsAllLinesAreWhitespace(int lineFrom, int lineTo) {
...
+                leftCurly = ast.findFirstToken(TokenTypes.LCURLY);
+            }
+            lcurlyOptional = Optional.ofNullable(leftCurly);
```
nrmancuso, line 294: If we do ^, then you end up with a very succinct catch all:
```suggestion
        else {
            lcurlyOptional = Optional.ofNullable(ast.findFirstToken(TokenTypes.LCURLY));
        }
```

This makes a lot of sense; if we match none of the special cases above, we just try to get the first left curly that exists if it is present.

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
```diff
@@ -270,23 +272,28 @@ private boolean checkIsAllLinesAreWhitespace(int lineFrom, int lineTo) {
...
-    private static DetailAST findLeftCurly(DetailAST ast) {
-        final DetailAST leftCurly;
-        final DetailAST slistAST = ast.findFirstToken(TokenTypes.SLIST);
-        if ((ast.getType() == TokenTypes.LITERAL_CASE
-                || ast.getType() == TokenTypes.LITERAL_DEFAULT)
-                && ast.getNextSibling() != null
-                && ast.getNextSibling().getFirstChild() != null
-                && ast.getNextSibling().getFirstChild().getType() == TokenTypes.SLIST) {
-            leftCurly = ast.getNextSibling().getFirstChild();
+    private static Optional<DetailAST> findLeftCurly(DetailAST ast) {
```
nrmancuso, line 275:
```suggestion
    private static Optional<DetailAST> getLeftCurly(DetailAST ast) {
```

nrmancuso: `find` doesn't imply `get` :)

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
mahfouz72: good catch :). done

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
mahfouz72: > This makes a lot of sense; if we match none of the special cases above, we just try to get the first left curly that exists if it is present.

I agree, done.

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
mahfouz72: good, done

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
mahfouz72: yea done didn't add much value

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
mahfouz72: done. renamed optional to leftCurly and the value we get from the Optional to leftCurlyAST is it good ?

nrmancuso (changes requested):
Last from me: please add a test case for `text` block option, where we have some comments explaining why we have no actual code.

After you push your next changes, let's generate the report again

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java
nrmancuso: yep

mahfouz72:
github, generate report

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/InputEmptyBlockSwitchExpressions.java
nrmancuso: > Text was added in #8662 . `CASE` was a token of the check then, so I am slightly confused why we originally said this was ok to be allowed.
> 
> @nrmancuso When you review.

Yeah this was weird, I think this should be a violation

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/InputEmptyBlockCaseAndDefaultWithTextOption.java
```diff
@@ -0,0 +1,94 @@
+/*
+EmptyBlock
+option = text
```
rnveach, line 3: Example using statement.

rnveach:
@nrmancuso 
> What is a sequential block?
I have no idea. The doc for this check is not good should we update it? in another PR?

Do you have any ideas about this? It is specifically written in the documentation for this check.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/InputEmptyBlockCaseAndDefaultWithTextOption.java
```diff
@@ -0,0 +1,94 @@
...
+tokens = LITERAL_DEFAULT, LITERAL_CASE, LITERAL_SWITCH
```
rnveach, line 4: More curious. Why did you add `switch` here, when you don't add it in the updated example.

rnveach (changes requested):

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/InputEmptyBlockCaseAndDefaultWithTextOption.java
mahfouz72: We have this already (The input file that I modifed in this PR)
https://github.com/checkstyle/checkstyle/blob/3369a7ab5dd8e07d5299e6016a88e38e0cba09a9/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/InputEmptyBlockSwitchExpressions.java

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/emptyblock/InputEmptyBlockCaseAndDefaultWithTextOption.java
mahfouz72: Copy paste from another file and forget to remove it. but actually, we don't need it here, removed

nrmancuso:
> @nrmancuso
> 
> > What is a sequential block?
> > I have no idea. The doc for this check is not good should we update it? in another PR?
> 
> Do you have any ideas about this? It is specifically written in the documentation for this check.

I have no idea what this means, I am sorry 😂

rnveach:
@mahfouz72 @nrmancuso 
Text was added at https://github.com/checkstyle/checkstyle/commit/350801597ccce93d019ab03f5f171d5acc27e3d9 . Looks like conversation started at https://github.com/checkstyle/checkstyle/pull/3918#discussion_r107571758 .

It is meant to say we don't support `case 0: {} {}`.

rnveach (approved):

Issue #15218, opened 2024-07-09
Format google IT tests' input files to follow all the rules of google java style guide


Zopsss:
All google tests are being migrated to whole config testing in this issue: https://github.com/checkstyle/checkstyle/issues/15214

We detected that Google java style simply reference to google-java-format as source of truth for specific chapter about line length (not whole) https://github.com/checkstyle/checkstyle/pull/14901/files#r1639771738 , same is location of `{` at https://github.com/checkstyle/checkstyle/pull/14901/files#r1621156853 . So we need to use this tool for Inputs and see cases where we are conflicting to make decision who is right. For other chapters we will still look at style guide first as source of truth.

As part of this issue, all the google tests' input files needs to be formatted to follow [google java style guide](https://google.github.io/styleguide/javaguide.html) rules. The input files should be formatted by [google-java-format](https://github.com/google/google-java-format) tool locally. This tool helps us to format the java file to follow google java style guide rules. Though this tool formats only some specific parts of the input file, it can fix incorrect spacing, indentation, brackets/curly braces placements, etc but it cannot fix incorrect variable names, missing javadoc comments, etc.

The input files will require to follow all the rules of the style guide, expect it will contain violations only corresponding to their section. In some cases, the input file will require some extra violations outside its corresponding section, which will be required to test the section properly, example: https://github.com/checkstyle/checkstyle/pull/15220#discussion_r1671190653

---

Initially, we thought of adding google-java-format tool as a dependency to our project ( [PR](https://github.com/checkstyle/checkstyle/commit/04ccd77368ac3761e1d54d9c89c738450cbb3415#diff-3323f2bf6b63b5111ea2e99bde3afbdbb048701d56afaee6170bbc10a4fe6e46) in which we tried to did this ) but after doing discussion on this topic, maintainers have come upon the idea of not adding this tool as a dependency for now. In future, we might add support for this tool, but for now we're not doing this, this tool should be used locally to format the input file but should not be added to CS as a dependency.

---

Input files of following sections needs to be formatted to follow all rules of google java style guide:

- [x] 2.1 File name
- [x] 2.3.1 Whitespace characters
- [x] 2.3.2 Special escape sequences
- [x] 2.3.3 Non-ASCII characters
- [x] 3 Source file structure
- [x] 3.2 Package statement
- [x] 3.3.2 No line-wrapping
- [x] 3.3.3 Ordering and spacing
- [x] 3.4.1 Exactly one top-level class declaration
[...]

PR #15278, opened 2024-07-15
Issue #15218: formatted input file for section 4.8.7 modifiers to follow google style guide rules

Zopsss:
#15218

### DIFF
--- a/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule487modifiers/InputModifierOrder.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule487modifiers/InputModifierOrder.java
@@ -1,300 +1,289 @@
 package com.google.checkstyle.test.chapter4formatting.rule487modifiers;
 
-strictfp abstract class InputModifierOrder
-// violation above ''abstract' modifier out of order with the JLS suggestions.'
-{
+// violation below ''abstract' modifier out of order with the JLS suggestions.'
+strictfp abstract class InputModifierOrder {
+  transient private String dontSaveMe;
+  // violation above ''private' modifier out of order with the JLS suggestions.'
+
+  volatile public int whatImReading;
+  // violation above ''public' modifier out of order with the JLS suggestions.'
+
+  public volatile boolean sModifierOrderVar = false;
+
+  /**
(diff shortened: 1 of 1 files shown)


src/it/resources/com/google/checkstyle/test/chapter4formatting/rule487modifiers/InputModifierOrder.java
```diff
@@ -1,300 +1,289 @@
...
 
-strictfp abstract class InputModifierOrder
-// violation above ''abstract' modifier out of order with the JLS suggestions.'
-{
-    transient private String dontSaveMe;
-    // violation above ''private' modifier out of order with the JLS suggestions.'
+// violation below ''abstract' modifier out of order with the JLS suggestions.'
+strictfp abstract class InputModifierOrder {
+  transient private String dontSaveMe;
+  // violation above ''private' modifier out of order with the JLS suggestions.'
```
rdiachenko, line 6: Why did you move violation comment? It does not look like a part of formatting.

In general, I think we should better be consistent in cases with a single violation. It is more readable to have `violation below` comment instead of `violation above` for simple cases. Here you mix both, let's be consistent.

romani: It might be due to fact that formatter wants `InputModifierOrder {` but comment is in middle .

Did formatter move a comment ?

Zopsss: > It might be due to fact that formatter wants InputModifierOrder { but comment is in middle .

Yes

> Did formatter move a comment ?

No, I moved the comment and corrected the `{` placement

romani: >It is more readable to have violation below comment instead of violation above for simple cases. Here you mix both, let's be consistent.

We already have mix, if we do  this now, it will be more changes that you/we might not like also.
We still did not make good rules on when use each variant of comment.

rdiachenko (approved):
lgtm

PR #15304, opened 2024-07-17
Issue #15025: Add support for LITERAL_WHEN in WhiteSpaceAfterCheck

mahfouz72:
closes #15025 

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/759257e4c69433ba141d623872b21d1c3dd3a4f3/check.xml
Diff Regression patch config: https://gist.githubusercontent.com/mahfouz72/33967bf86e4a29cafa429771b0e95808/raw/49ec3a90de0ff4ebd87979177c2cda622f751c4e/check_patch.xml

PR #15311, opened 2024-07-18
Issue #15027: Add support for LITERAL_WHEN in whitespaceAroundCheck

### LINKED ISSUE #15027: Add Check Support for Java 21 Pattern Matching for Switch Syntax: WhitespaceAround

I have read check documentation: https://checkstyle.org/checks/whitespace/whitespacearound.html#WhitespaceAround

```bash
➜  src cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="WhitespaceAround"/>
    </module>
</module>
➜  src cat Test.java 

public class Test {

    void test(Object obj) {
        int y = 0;
        if(y >= 2) { }      // violation
        switch (obj) {
            case ColoredPoint(int x, _, _) when(x >= 2) -> { }   // expected violation
            default -> { }
        }
    }

    void test2(Object obj) {
        int y=0; // 2 violations, no space before and after '='
        switch (obj) {
            case ColoredPoint(int x, _, _)when(x >= 2) -> { } // (expected) 2 violations, no space before and after 'when'
            case ColoredPoint(int x, _, _) when(x >= 0)  -> { }  // (expected) 1 violation, no space after 'when'
            default -> { }
        }
    }
}
Starting audit...
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:8:9: 'if' is not followed by whitespace. [WhitespaceAround]
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:16:14: '=' is not followed by whitespace. [WhitespaceAround]
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:16:14: '=' is not preceded with whitespace. [WhitespaceAround]
Audit done.

```

---

violation on `when`.

mahfouz72:
closes #15027 

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/2c9cfc2b9a7daad757c39ec06d41a75da0c95730/check.xml
Diff Regression patch config: https://gist.githubusercontent.com/mahfouz72/33967bf86e4a29cafa429771b0e95808/raw/f607f60b0a9a5ad9ecf1e13fea9e31d4dddc32da/check_patch.xml


mahfouz72:
github, generate report

src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
```diff
@@ -179,7 +179,9 @@ public class AllChecksTest extends AbstractModuleTestSupport {
                 "ARRAY_INIT",
                 "ELLIPSIS",
                 // these are covered by GenericWhitespaceCheck
-                "WILDCARD_TYPE", "GENERIC_END", "GENERIC_START")
+                "WILDCARD_TYPE", "GENERIC_END", "GENERIC_START",
+                // 'when' keyword can be used by newer java versions only
```
nrmancuso, line 183: We don't need this here, let's remove this comment.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/whitespace/whitespacearound/InputWhitespaceAroundLiteralWhen.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class InputWhitespaceAroundLiteralWhen {
+
+    void test(Object o) {
+        switch (o) {
+            case Integer i when(i == 0) -> {} // violation, ''when' is not followed by whitespace'
+            // violation below, ''when' is not followed by whitespace'
+            case String s when(s.equals("a")) -> {}
+            // violation below, ''when' is not followed by whitespace'
+            case Point(int x, int y ) when(x>=0 && y >=0) -> {}
```
nrmancuso, line 18: Same as https://github.com/checkstyle/checkstyle/pull/15304#discussion_r1683362869, let's find other tokens to put here.

nrmancuso (changes requested):
Items:

src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
mahfouz72: done

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/whitespace/whitespacearound/InputWhitespaceAroundLiteralWhen.java
mahfouz72: done

nrmancuso (approved):

nrmancuso:
We are also missing checks in this PR

nrmancuso (approved):

src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
```diff
@@ -179,7 +179,8 @@ public class AllChecksTest extends AbstractModuleTestSupport {
                 "ARRAY_INIT",
                 "ELLIPSIS",
                 // these are covered by GenericWhitespaceCheck
-                "WILDCARD_TYPE", "GENERIC_END", "GENERIC_START")
+                "WILDCARD_TYPE", "GENERIC_END", "GENERIC_START",
+                "LITERAL_WHEN")
```
rnveach, line 183: Show me that when is covered here.

mahfouz72: can you restate. I don't understand. what do you mean by "covered here"

nrmancuso: @mahfouz72 what would happen if we don't exclude this token? It should not appear in our code until we migrate to java 21.

mahfouz72: there is a test that validate that all checks tokens are used in google config https://github.com/checkstyle/checkstyle/blob/42158b1fbad0ee3ef2d60a246635c805910c4057/src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java#L391

mahfouz72: So we should add it to the config or exclude it like I did to make the test passes

mahfouz72: added `LITERAL_WHEN` to the config.

rnveach: @mahfouz72 What is your understanding of what it means to put the token here at this location?

mahfouz72: Ideally, we should have all checks tokens in our checkstyle config and we have a test to enforce that. This list is to exclude some tokens from being added in the config (suppress the test for this specific token). an example for this `GENERIC_START` which should be in the config of the whitespaceAround check but we didn't add it there because we have a separate check for it so we added it in this exclude list

mahfouz72: My understanding at first is why we would need to add LITERAL_WHEN in the config? we don't use this syntax in our code what will we gain if we add it in the config? that's why I excluded it

mahfouz72: Please correct me if anything is wrong and clarify this comment https://github.com/checkstyle/checkstyle/pull/15311#discussion_r1685632454

rnveach: So this area is to document what tokens are excluded from our config as well as the other configs. When we add these tokens here, we document why. The comment that is next to the token you added is `these are covered by GenericWhitespaceCheck`. To me, that comment coveys why we are adding this token here. So my question at the time was, how does `GenericWhitespaceCheck` cover this token for another check to satisfy why we aren't adding it.

mahfouz72: ahh, The comment was misleading I didn't want to say that it applies to `LITERAL_WHEN`. actually there was another comment (https://github.com/checkstyle/checkstyle/pull/15311#discussion_r1683363896)    `// 'when' keyword can be used by newer java versions only` but I removed it based on @nrmancuso review he was probably was trying to say to remove the token at all from the list but i removed the comment only

mahfouz72: Okay now we are good I added the token to the config and removed it from the exclude list. I just was confused why we added to our config a token that we are not using. but as @nrmancuso said [here](https://github.com/checkstyle/checkstyle/pull/15311#discussion_r1685742936) there should be no issues with this. we will use it when we migrate to java 17

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundCheck.java
```diff
@@ -389,6 +389,7 @@ public int[] getAcceptableTokens() {
             TokenTypes.GENERIC_START,
             TokenTypes.GENERIC_END,
             TokenTypes.ELLIPSIS,
+            TokenTypes.LITERAL_WHEN,
```
rnveach, line 392: @nrmancuso Is there a reason we don't make this default? We should our reasoning to the first post in the issue.

It seems like this won't hurt anything unless its used and if it is used, it would only be on a project which supports it. I am not seeing any negatives yet on why this can't be defaulted.

src/main/resources/google_checks.xml
```diff
@@ -126,7 +126,8 @@
                     LITERAL_FOR, LITERAL_IF, LITERAL_RETURN, LITERAL_SWITCH, LITERAL_SYNCHRONIZED,
                     LITERAL_TRY, LITERAL_WHILE, LOR, LT, MINUS, MINUS_ASSIGN, MOD, MOD_ASSIGN,
                     NOT_EQUAL, PLUS, PLUS_ASSIGN, QUESTION, RCURLY, SL, SLIST, SL_ASSIGN, SR,
-                    SR_ASSIGN, STAR, STAR_ASSIGN, LITERAL_ASSERT, TYPE_EXTENSION_AND"/>
+                    SR_ASSIGN, STAR, STAR_ASSIGN, LITERAL_ASSERT,
+                    TYPE_EXTENSION_AND, LITERAL_WHEN"/>
```
rnveach, line 130: @nrmancuso @romani ping.

@mahfouz72 If we are changing google checks, we need documentation that the google style guide supports this change.
Basically what specific rule supports this.
https://checkstyle.org/styleguides/google-java-style-20180523/javaguide.html

For our config, we like to use as much as possible, but only if we really agree with the rule being applied to our repo. If we don't agree with it, we usually still add it but set it's violations to be ignored so we are still atleast no exception testing it.

rnveach (changes requested):

src/main/resources/google_checks.xml
rnveach: @Zopsss ping as well

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/whitespace/whitespacearound/InputWhitespaceAroundLiteralWhen.java
```diff
@@ -0,0 +1,42 @@
...
+            default -> {}
+        }
+
+         switch (o) {
```
rnveach, line 22: Indentation. Switch is wrong location.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/whitespace/whitespacearound/InputWhitespaceAroundLiteralWhen.java
```diff
@@ -0,0 +1,42 @@
+/*
+WhitespaceAround
+tokens = LITERAL_WHEN
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.whitespace.whitespacearound;
+
+public class InputWhitespaceAroundLiteralWhen {
```
rnveach, line 10: Add some examples of a new line before/after `when`.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundCheck.java
nrmancuso: Good point, let’s make this default

rnveach: @mahfouz72 Please make this default also. This should also revert `checkstyle-checks.xml` to what it was before.

src/main/resources/google_checks.xml
Zopsss: [4.6.2 Horizontal whitespace](https://checkstyle.org/styleguides/google-java-style-20180523/javaguide.html#s4.6.2-horizontal-whitespace) section defines rules for Whitespace around code. The documentation does not specifically mentions anything about the `when` keyword but the google-java-format adds whitespace around `when`. Here's the formatted code:

```java
record ColoredPoint(int p, int x, String c) {}

record Rectangle(ColoredPoint upperLeft, ColoredPoint lowerRight) {}

public class Test {

  void test(Object obj) {
    int y = 0;
    if (y >= 2) {} // violation
    switch (obj) {
      case ColoredPoint(int x, _, _) when (x >= 2) -> {} // expected violation
      default -> {}
    }
  }

  void test2(Object obj) {
    int y = 0; // 2 violations, no space before and after '='
    switch (obj) {
      case ColoredPoint(int x, _, _) when (x
          >= 2) -> {} // (expected) 2 violations, no space before and after 'when'
      case ColoredPoint(int x, _, _) when (x
          >= 0) -> {} // (expected) 1 violation, no space after 'when'
      default -> {}
    }
  }
}
```

`when` was introduced in Java 21, the last major update for style guide was before Java 21 released so it does not contains rule for newly introduced keywords. 

But I think we should add support for this keyword, it is ensures basic spacing between `when` and other code and it is overall improving the coding style

Zopsss: If we agree with this keyword then I think it's we should test cases for it in [InputWhitespaceAroundBasic.java](https://github.com/checkstyle/checkstyle/blob/master/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule462horizontalwhitespace/InputWhitespaceAroundBasic.java) file.

rnveach: @Zopsss Even though it is not specifically mentioned, are there any parts of the style which indicate this spacing follows the heart of the guide? Unfortunately we can't go by what "looks good" since that is subjective.

This is the closest I am seeing
> Separating any reserved word

`when` is a keyword like the others
https://docs.oracle.com/javase/specs/jls/se22/html/jls-3.html#jls-3.9
But it isn't require to have a `(` after it, but the google documentation also clearly states the spaces should only appear in very specific places.

@romani @nrmancuso What do you think?

Zopsss: > But it isn't require to have a `(` after it

I see. If that's the case then based on style guide, we don't need any kind rule for `when` as it does not mentions anything explicitly about it. 

Though the formatter does adds spacing around it and I also think we should add support for this token to maintain consistent spacing. I'll be waiting for other maintainers opinion on this.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundCheck.java
mahfouz72: done added to the default tokens and revered checkstyle-checks.xml

src/main/resources/google_checks.xml
mahfouz72: > Separating any reserved word, such as if, for or catch, from an open parenthesis (() that follows it on that line

`when` is a reserved word and it may have an open parenthesis so I think google style should support this token

> But it isn't require to have a ( after it

why does this information affect our decision? Yes it is not required but It still may have an open parenthesis

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/whitespace/whitespacearound/InputWhitespaceAroundLiteralWhen.java
mahfouz72: done

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/whitespace/whitespacearound/InputWhitespaceAroundLiteralWhen.java
mahfouz72: done

src/main/resources/google_checks.xml
rnveach: @mahfouz72 Please read https://checkstyle.org/styleguides/google-java-style-20180523/javaguide.html#s4.6.2-horizontal-whitespace .
Google config follows style guide and style guide seems to be very strict in this case on what is supposed to happen with spacing.

romani: @mahfouz72 , please use formatter https://github.com/google/google-java-format on input to confirm that `when` in spaced on both sides.

mahfouz72: @Zopsss used the formatter and confirmed that it adds spaces around `when` https://github.com/checkstyle/checkstyle/pull/15311#discussion_r1686824725

Zopsss: > @mahfouz72 , please use formatter https://github.com/google/google-java-format on input to confirm that `when` in spaced on both sides.

Formatter adds spaces on both sides of `when` but there's no spacing rule defined for Java 21 keywords in the style guide because style guide is more older than the Java 21 release.

romani: @mahfouz72 , please add example of "when" to new Input at folder
https://github.com/checkstyle/checkstyle/tree/master/src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting
and create test method at https://github.com/checkstyle/checkstyle/blob/master/src/it/java/com/google/checkstyle/test/chapter4formatting/rule462horizontalwhitespace/HorizontalWhitespaceTest.java

mahfouz72: https://github.com/checkstyle/checkstyle/pull/15359

rnveach: Same state as other PR. @nrmancuso @romani

romani: Yes, this PR is above my expectations.
Thanks a lot.

mahfouz72:
@rnveach @nrmancuso we should unblock this aslo

nrmancuso:
> @rnveach @nrmancuso we should unblock this aslo

done, let's party

rnveach (approved):

PR #15313, opened 2024-07-18
Issue #15292: add OS-specific commands to Running > Command Line > Usage by Classpath Update

### LINKED ISSUE #15292: Potential Typo in Running > Command Line > Usage by Classpath Update
Hi folks, new-to-checkstyle user here. Apologies for not using the bug report template, but that seems like overkill for this issue.

I was looking at [Running > Command Line > Usage by Classpath Update](https://checkstyle.sourceforge.io/cmdline.html#Usage_by_Classpath_update), and the last example under "Run checkstyle with [custom module(s) (Root module, Checks, etc)](https://checkstyle.sourceforge.io/extending.html) in configuration file:" has the following command:

```
java -classpath MyCustom.jar;checkstyle-10.17.0-all.jar \
    com.puppycrawl.tools.checkstyle.Main -c config.xml Check.java
```

The semicolon here seems to be a typo (unless I'm missing something); I believe this should be

```
java -classpath MyCustom.jar:checkstyle-10.17.0-all.jar \
    com.puppycrawl.tools.checkstyle.Main -c config.xml Check.java
```

Usually not a big deal, but I noticed that this was also [picked up by a 2018 SO post](https://stackoverflow.com/a/50673260).

Am I missing something? If not, happy to submit a one-line PR to fix this here:

https://github.com/checkstyle/checkstyle/blob/eb6f9e5379c66d268b5cf4e77f758b9ec2fed656/src/xdocs/cmdline.xml.vm#L339

(or, have an existing contributor do it if that'd be faster!)

mattxwang:
Pretty straightforward: adds two different sets of commands for Windows vs Linux/Unix, quotes the entire classpath, and adds a trailing note. Open to any suggestions (copy, placement, etc.).

Closes #15292.

### DIFF
--- a/src/xdocs/cmdline.xml.vm
+++ b/src/xdocs/cmdline.xml.vm
@@ -334,17 +334,34 @@ java -D&lt;property&gt;=&lt;value&gt;  \
           (Root module, Checks, etc)</a> in configuration file:
         </b>
       </p>
+
+      <p>For Linux/Unix OS:</p>
       <div class="wrap-content">
         <source>
-          java -classpath MyCustom.jar;checkstyle-${projectVersion}-all.jar \
+          java -classpath "MyCustom.jar:checkstyle-${projectVersion}-all.jar" \
           &#xa0;&#xa0;&#xa0;&#xa0;com.puppycrawl.tools.checkstyle.Main -c config.xml Check.java
         </source>
       </div>
+
+      <p>For Windows OS:</p>
+      <div class="wrap-content">
+        <source>
+          java -classpath "MyCustom.jar;checkstyle-${projectVersion}-all.jar" ^
+          &#xa0;&#xa0;&#xa0;&#xa0;com.puppycrawl.tools.checkstyle.Main -c config.xml Check.java
+        </source>
+      </div>
+
       <p>
         <b>Note</b>: Custom modules should be specified with the class'
         <a href="writingchecks.html#Integrating_Checks">full classpath</a>
         in the configuration file and the compiled class be located in the custom
         JAR for Checkstyle to find.
+      </p>
+
+      <p>
+        <b>Note</b>: You will need to use the appropriate path
+        separator (typically <code>;</code> on Windows, <code>:</code> on
+        Linux/Unix and macOS).
       </p>
     </section>
   </body>


mattxwang:
Since I'm a first-time contributor, I think a maintainer may need to approve some of the CI workflows!

romani:
GitHub, generate website

romani (approved):
Ok to merge.

Thanks a lot for update

romani:
There is already 

>Note: You will need to use the appropriate path separator (typically ; on Windows, : on Linux/Unix and macOS).

:) . Looks like people are not able to read to the end. We might need to move this note above example.

romani:
If we have energy to change more, I would make new separate section above existing and move this example to it.

Current `Usage By Classpath Update` is ancient approach to run on classes, nobody is doing this nowadays.

But usage with custom jar is still actual and the in demand.

src/xdocs/cmdline.xml.vm
```diff
@@ -334,18 +334,35 @@ java -D&lt;property&gt;=&lt;value&gt;  \
...
+          java -classpath "MyCustom.jar:checkstyle-${projectVersion}-all.jar" \
           &#xa0;&#xa0;&#xa0;&#xa0;com.puppycrawl.tools.checkstyle.Main -c config.xml Check.java
         </source>
       </div>
+
+      <p>For Windows OS:</p>
+      <div class="wrap-content">
+        <source>
+          java -classpath "MyCustom.jar;checkstyle-${projectVersion}-all.jar" ^
+          &#xa0;&#xa0;&#xa0;&#xa0;com.puppycrawl.tools.checkstyle.Main -c config.xml Check.java
```
rnveach, line 350: DOS:
````
M:\>java -classpath "MyCustom.jar;checkstyle-10.17.0-all.jar" ^
More?               com.puppycrawl.tools.checkstyle.Main -c config.xml Check.java
Files to process must be specified, found 0.

````

PowerShell:
````
PS M:\> java -classpath "MyCustom.jar;checkstyle-10.17.0-all.jar" ^
Error: Could not find or load main class ^
Caused by: java.lang.ClassNotFoundException: ^
PS M:\>               com.puppycrawl.tools.checkstyle.Main -c config.xml Check.java
````

I don't feel this current documentation is enough for Windows users. I also find it odd the first thing we would add here would be DOS' instead of PS. On Windows 10, you have to specifically go through the trouble to open DOS. Only option given in Windows Explorer is PS.

rnveach (changes requested):

src/xdocs/cmdline.xml.vm
romani: @mattxwang , please share pictures of how it executed on your side.

romani: @mattxwang , ping.

rnveach:

mattxwang:
Ah so sorry, I had missed the first few pings on this.

> Please create a new PR with other admins as the reviewers.

Will do when I have more time! Sorry for the inconvenience.

PR #15315, opened 2024-07-18
minor: fix XmlMetaReaderTest

mahfouz72:


romani:
Any ideas how we missed this ?

mahfouz72:
Two checks were merged together. The first branch (first check) incremented a number by 1 to make its CI pass. The second branch (second check) also incremented the same number by 1 to make its CI pass. However, after merging the first check, we should have updated the code of the second check to increment the number by 2 instead of 1.

rnveach:
Basically Git saw the update as a no change and removed it from merger of the 2nd PR. We won't be able to catch things like this unless we always rebase on head of master.

PR #15321, opened 2024-07-19
Issue #15034: Add new input file for pattern matching for switch in N…

### LINKED ISSUE #15034: Add Check Support for Java 21 Pattern Matching for Switch Syntax: NeedBraces

I have read check documentation: https://checkstyle.org/checks/blocks/needbraces.html

```
PS D:\CS\test> cat config.xml                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="NeedBraces">
            <property name="tokens" value="LITERAL_CASE, LITERAL_DEFAULT"/>
        </module>
    </module>
</module>

public class Test {

    void test(Object obj) {
        switch (obj) {
            case ColoredPoint(_, _, _):
            case Rectangle(_, _) : // violation
                break;
            default : // expected violation
        }
        switch (obj) {
            case ColoredPoint(int x, _, _) when (x >= 2)   // violation
                    :
                break;
            case ColoredPoint(int x, _, _)    // violation
                    when(x >= 2) :
                break;
            case ColoredPoint(int x, _, _)
                    when(x >= 2) : {}
            break;
            default :   // expected violation
        }
        switch (obj) {
            case ColoredPoint(int x, _, _) when (x >= 2) // violation
                    -> throw new RuntimeException();
            case ColoredPoint(int x, _, _)
                    when(x >= 2) -> {}
            default -> throw new RuntimeException();  // violation
        }
    }
}
Starting audit...
[ERROR] D:\CS\test\src\Test.java:9:13: 'case' construct must use '{}'s. [NeedBraces]
[ERROR] D:\CS\test\src\Test.java:14:13: 'case' construct must use '{}'s. [NeedBraces]
[ERROR] D:\CS\test\src\Test.java:17:13: 'case' construct must use '{}'s. [NeedBraces]
[ERROR] D:\CS\test\src\Test.java:26:13: 'case' construct must use '{}'s. [NeedBraces]
[ERROR] D:\CS\test\src\Test.java:30:13: 'default' construct must use '{}'s. [NeedBraces]
Audit done.
Checkstyle ends with 5 errors.
PS D:\CS\test> 

```

---

we are ok here. This can be closed or add an input file for cases with `when`

mahfouz72:
closes #15034

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheckTest.java
@@ -276,4 +276,48 @@ public class NeedBracesCheckTest extends AbstractModuleTestSupport {
             expected);
     }
 
+    @Test
+    public void testPatternMatchingForSwitch() throws Exception {
+        final String[] expected = {
+            "17:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "21:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "25:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "28:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
+            "53:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "55:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "58:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "61:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
+            "74:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "78:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "80:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "83:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
+            "90:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "91:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+        };
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputNeedBracesPatternMatchingForSwitch.java"),
+            expected);
+    }
+
+    @Test
+    public void testPatternMatchingForSwitchAllowSingleLine() throws Exception {
+        final String[] expected = {
+            "17:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "21:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "25:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "28:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
+            "55:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "58:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "61:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
+            "74:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "80:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "83:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
+            "90:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+        };
+        verifyWithInlineConfigParser(
+                getNonCompilablePath(
+                        "InputNeedBracesPatternMatchingForSwitchAllowSingleLine.java"),
+            expected);
+    }
+
 }
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitch.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitch.java
@@ -1,0 +1,101 @@
+/*
+NeedBraces
+allowSingleLineStatement = false
+allowEmptyLoopBody = (default)false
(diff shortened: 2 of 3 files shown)


src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitch.java
```diff
@@ -0,0 +1,100 @@
...
+
+            default :  // violation
+                yield 0;
+        };
+
+        int b =  switch (o) {
+            case Integer i when (i == 0) -> i;
+            case String s when(s.equals("a")) -> s.length();  // violation
+            case Point(int x, int y ) when(x>=0 && y >=0) ->  // violation
+                x + y;
```
mahfouz72, line 91: 2 false positives, we can't have braces here, otherwise we need to introduce an unnecessary yield.

```Java
            case String s when(s.equals("a")) -> {s.length();} // non-compliant, Switch expression rule should produce result in all execution paths
            case String s when(s.equals("a")) -> {yield s.length();} // compliant but unnecessary yield
```

interesting that line 88 is ok (and this is expected) while line 89 and 90 is violations

mahfouz72: https://github.com/checkstyle/checkstyle/pull/15321#discussion_r1684288238

mahfouz72: added until comment

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitch.java
```diff
@@ -0,0 +1,101 @@
...
+                break;
+            }
+            case String s when(s.equals("a")) : {
+                System.out.println("a");
+                break;
+            }
+            case Point(int x, int y ) when(x>=0 && y >=0) : {
+                System.out.println("point");
+                break;
+                }
```
nrmancuso, line 44: Please fix indentation here and anywhere else.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitch.java
```diff
@@ -0,0 +1,101 @@
...
+         switch (o) {
+            case Integer i when (i == 0) : {
+                System.out.println("zero");
+                break;
+            }
+            case String s when(s.equals("a")) : {
+                System.out.println("a");
+                break;
+            }
+            case Point(int x, int y ) when(x>=0 && y >=0) : {
```
nrmancuso, line 41:
```suggestion
            case Point(int x, int y ) when (x>=0 && y >=0) : {
```
Please fix all of these

src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheckTest.java
```diff
@@ -276,4 +276,47 @@ public void testNeedBracesSwitchExpression() throws Exception {
...
+            "90:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "91:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+        };
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputNeedBracesPatternMatchingForSwitch.java"),
+            expected);
+    }
+
+    @Test
+    public void testPatternMatchingForSwitch2() throws Exception {
```
nrmancuso, line 303: Please never do this; test names should do a good job of describing what we are testing; we can draw inspiration from configurations, specific code under test, etc.

src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheckTest.java
```diff
@@ -276,4 +276,47 @@ public void testNeedBracesSwitchExpression() throws Exception {
...
+            "55:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "58:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "61:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
+            "74:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "80:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "83:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
+            "90:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+        };
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputNeedBracesPatternMatchingForSwitch2.java"),
```
nrmancuso, line 318: Same for file naming, it should be consistent with the test method name.

nrmancuso (changes requested):
Items:

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitch.java
mahfouz72: done

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitch.java
mahfouz72: done

src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheckTest.java
mahfouz72: done

src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheckTest.java
mahfouz72: done

nrmancuso (approved):

Issue #15256, opened 2024-07-14
 InputEqualsHashCodeEqualsParameter.java has wrong location of violation comments


romani:
detected at https://github.com/checkstyle/checkstyle/pull/15021

InputEqualsHashCodeEqualsParameter.java has wrong location of violation comments.

https://github.com/checkstyle/checkstyle/blob/6d8f8f2f9064ab7706fd5c1dd5c715fe0b4ac599/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/equalshashcode/InputEqualsHashCodeEqualsParameter.java#L15-L16

violation is actually only on method, not on class.

Reason is wrong implementation of regexp to match comment wihtout message but with some words as reasons or explanation
example: `// violation, no correct `equals``
vs comment with violation message
`// violation 'without .* of 'equals()'.'`

Sugested regexp: `".*//\\s*violation,\\s+(?:.*)?$"` from https://github.com/checkstyle/checkstyle/pull/15021#discussion_r1676942913

 
Expected: update remove bdd/InlineConfigParser.java to match to `// violations, bla bla bla`. Remove extra comments from this InputEqualsHashCodeEqualsParameter file.

If there is "violation" word and this line does not match any pattern - throw exception.

rnveach:
We need some examination of why we aren't catching this and how we can avoid these type of things in the future. I tried a `// violation` in Google's IT, and it picked it up as a missing violation.

romani:
Reason is at https://github.com/checkstyle/checkstyle/pull/15021#discussion_r1657178202

There are multiple matchers to trailing comments, we should match to VIOLATION_WITH_EXPLANATION_PATTERN
But content of such regexp is wrong, not matching to name of field , just a defect.

As there was no usage of it, it wasn't detected.

After fix in referenced PR, it will be in use, so if any other regression happens, build failure will happen in test.

rnveach:
> Reason is at https://github.com/checkstyle/checkstyle/pull/15021#discussion_r1657178202

I disagree. Link is why current expression doesn't match the violation text, but it is not why we missed the violation comment completely.

We should identify anything written as `// violation` as a violation and then fail with an error if we can't match it with a specific message. We shouldn't have violation comments go completely unnoticed as it did to create this issue.

romani:
Ok, this is slightly different issue with bigger scope .

@SteLeo1602 , please extend this sequence of matchers with one more matcher at the end that match any comment with `// violation .*`

[...]

PR #15021, opened 2024-06-18
Issue #13345: Enabled SeverityMatch example tests

SteLeo1602:
Issue: #13345

Enabled the example tests for SeverityMatch filter.

PR #15327, opened 2024-07-20
Issue #15256: Fixed location of violation comments

SteLeo1602:
Issue: #15256

Updated bdd/InlineConfigParser.java to match violation comments
Removed extra comments from this InputEqualsHashCodeEqualsParameter file.
Added additional matcher that matches the following pattern: `//.*violation.*`

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
@@ -96,7 +96,7 @@ public final class InlineConfigParser {
 
     /** A pattern to find the string: "// violation, explanation". */
     private static final Pattern VIOLATION_WITH_EXPLANATION_PATTERN = Pattern
-            .compile(".*//\\s*violation,\\s.+\\s(?:['\"](.*)['\"])?$");
+            .compile(".*//\\s*violation,\\s+(?:.*)?$");
 
     /** A pattern to find the string: "// X violations". */
     private static final Pattern MULTIPLE_VIOLATIONS_PATTERN = Pattern
@@ -174,6 +174,10 @@ public final class InlineConfigParser {
      */
     private static final Pattern VIOLATIONS_SOME_LINES_BELOW_PATTERN = Pattern
             .compile(".*//\\s*(\\d+) violations (\\d+) lines below:$");
+
+    /** A pattern that matches any comment by default. */
+    private static final Pattern VIOLATION_DEFAULT = Pattern
+            .compile("//.*violation.*");
 
     /** The String "(null)". */
     private static final String NULL_STRING = "(null)";
@@ -764,6 +768,8 @@ public final class InlineConfigParser {
                 VIOLATIONS_SOME_LINES_ABOVE_PATTERN.matcher(lines.get(lineNo));
         final Matcher violationsSomeLinesBelowMatcher =
                 VIOLATIONS_SOME_LINES_BELOW_PATTERN.matcher(lines.get(lineNo));
+        final Matcher violationsDefault =
+                VIOLATION_DEFAULT.matcher(lines.get(lineNo));
         if (violationMatcher.matches()) {
             final String violationMessage = violationMatcher.group(1);
             final int violationLineNum = lineNo + 1;
@@ -796,11 +802,8 @@ public final class InlineConfigParser {
             inputConfigBuilder.addViolation(violationLineNum, violationMessage);
         }
         else if (violationWithExplanationMatcher.matches()) {
-            final String violationMessage = violationWithExplanationMatcher.group(1);
             final int violationLineNum = lineNo + 1;
-            checkWhetherViolationSpecified(specifyViolationMessage, violationMessage,
-                    violationLineNum);
-            inputConfigBuilder.addViolation(violationLineNum, violationMessage);
+            inputConfigBuilder.addViolation(violationLineNum, null);
         }
         else if (violationSomeLinesAboveMatcher.matches()) {
             final String violationMessage = violationSomeLinesAboveMatcher.group(2);
@@ -858,6 +861,10 @@ public final class InlineConfigParser {
         else if (useFilteredViolations) {
             setFilteredViolation(inputConfigBuilder, lineNo + 1,
                     lines.get(lineNo), specifyViolationMessage);
+        }
+        else if (violationsDefault.matches()) {
+            final int violationLineNum = lineNo + 1;
+            inputConfigBuilder.addViolation(violationLineNum, null);
         }
     }
 
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/equalshashcode/InputEqualsHashCodeEqualsParameter.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/equalshashcode/InputEqualsHashCodeEqualsParameter.java
@@ -12,7 +12,7 @@ public class InputEqualsHashCodeEqualsParameter {
             return true;
         }
     }
-    public static class TestClass2 { // violation, no correct `equals`
+    public static class TestClass2 {
         public int hashCode() { // violation 'without .* of 'equals()'.'
             return 1;
         }
@@ -20,7 +20,7 @@ public class InputEqualsHashCodeEqualsParameter {
             return true;
         }
     }
-    public static class TestClass3 { // violation, no `hashCode`
+    public static class TestClass3 {
         public boolean equals(Object o) { // violation 'without .* of 'hashCode()'.'
             return true;
         }
@@ -47,7 +47,7 @@ public class InputEqualsHashCodeEqualsParameter {
             return true;
         }
     }
-    public static class TestClass6 { // violation, no `hashCode` implementation
+    public static class TestClass6 {
         public static int hashCode(int i) {
             return 1;
         }
@@ -55,7 +55,7 @@ public class InputEqualsHashCodeEqualsParameter {
             return true;
         }
     }
-    public static class TestClass7 { // violation, no `equals` implementation
+    public static class TestClass7 {
         public int hashCode() { // violation 'without .* of 'equals()'.'
             return 1;
         }
@@ -67,23 +67,23 @@ public class InputEqualsHashCodeEqualsParameter {
         public native int hashCode();
         public native boolean equals(Object o);
     }
-    public static class TestClass9 { // violation, no `equals` implementation
+    public static class TestClass9 {
         public native int hashCode(); // violation 'without .* of 'equals()'.'
     }
-    public static class TestClass10 { // violation, no `hashCode` implementation
+    public static class TestClass10 {
         public native boolean equals(Object o); // violation 'without .* of 'hashCode()'.'
     }
     public static abstract class TestClass11 {
         public abstract int hashCode();
         public abstract boolean equals(Object o);
     }
-    public static abstract class TestClass12 { // violation, no `equals` implementation
+    public static abstract class TestClass12 {
         public int hashCode() { // violation 'without .* of 'equals()'.'
             return 1;
         }
         public abstract boolean equals(Object o);
     }
-    public static abstract class TestClass13 { // violation, no `hashCode` implementation
+    public static abstract class TestClass13 {
         public abstract int hashCode();
         public boolean equals(java.lang.Object o) { // violation 'without .* of 'hashCode()'.'
             return true;
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTag.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTag.java
@@ -2,14 +2,14 @@
 WriteTag
 tag = @author
 tagFormat = \\S
-tagSeverity = (default)info
+tagSeverity = error
 tokens = (default)INTERFACE_DEF, CLASS_DEF, ENUM_DEF, ANNOTATION_DEF, RECORD_DEF
 
 
 */
 
 package com.puppycrawl.tools.checkstyle.checks.javadoc.writetag;
-// violation 3 lines below , 'param @author must match pattern '\\S'.*'
+// violation 3 lines below 'Javadoc tag @author=Daniel Grenner'
 /**
  * Testing tag writing
  * @author Daniel Grenner
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport.java
@@ -19,7 +19,7 @@ package com.puppycrawl.tools.checkstyle.checks.regexp.regexpmultiline;
  */
 public class InputRegexpMultilineMultilineSupport {
     void method() { // violation below
-// abc - violation
+// abc
 // def
 // abc
     }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport2.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport2.java
@@ -18,10 +18,10 @@ package com.puppycrawl.tools.checkstyle.checks.regexp.regexpmultiline;
  * matchAcrossLines = true
  */
 public class InputRegexpMultilineMultilineSupport2 {
-    void method() { // violation below
-// abc - violation
-// def // violation below
-// abc - violation
+    void method() {
+// abc // violation
+// def
+// abc // violation
     }
 
     void method2() {
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/nocodeinfile/Example1.java
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/nocodeinfile/Example1.java
@@ -9,7 +9,7 @@
 // violation 8 lines above
 
 // xdoc section -- start
-// the violation is on first line of file
+/* the violation is on first line of file */
 // public class Example1 {
 // single-line comment is not code
 // }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheckTest.java
@@ -342,7 +342,7 @@ public class ImportOrderCheckTest extends AbstractModuleTestSupport {
     @Test
     public void testWildcard() throws Exception {
         final String[] expected = {
-            "24:1: " + getCheckMessage(MSG_ORDERING, "javax.crypto.Cipher"),
+            "21:1: " + getCheckMessage(MSG_ORDERING, "javax.crypto.Cipher"),
         };
 
         verifyWithInlineConfigParser(
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/importorder/InputImportOrder_Wildcard.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/importorder/InputImportOrder_Wildcard.java
@@ -15,9 +15,6 @@ tokens = (default)STATIC_IMPORT
 */
 
 package com.puppycrawl.tools.checkstyle.checks.imports.importorder;
-// groups are configured as follows
-// com.puppycrawl,*,java
-// the trailing javax.crypto.Cipher; should be flagged as a violation.
 
 import javax.crypto.BadPaddingException;
 import java.util.List; //comment test
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagDoubleTag.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagDoubleTag.java
@@ -2,15 +2,15 @@
 WriteTag
 tag = @doubletag
 tagFormat = \\S
-tagSeverity = (default)info
+tagSeverity = error
 tokens = (default)INTERFACE_DEF, CLASS_DEF, ENUM_DEF, ANNOTATION_DEF, RECORD_DEF
 
 
 */
 
 package com.puppycrawl.tools.checkstyle.checks.javadoc.writetag;
-// violation 6 lines below , '@doubletag appeared at the same time.*'
-// violation 6 lines below , '@doubletag appeared at the same time.*'
+// violation 6 lines below 'Javadoc tag @doubletag=first text'
+// violation 6 lines below 'Javadoc tag @doubletag=second text'
 /**
  * Testing tag writing
  * @author Daniel Grenner
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagIncomplete.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagIncomplete.java
@@ -2,14 +2,14 @@
 WriteTag
 tag = @incomplete
 tagFormat = \\S
-tagSeverity = (default)info
+tagSeverity = error
 tokens = (default)INTERFACE_DEF, CLASS_DEF, ENUM_DEF, ANNOTATION_DEF, RECORD_DEF
 
 
 */
 
 package com.puppycrawl.tools.checkstyle.checks.javadoc.writetag;
-// violation 4 lines below , 'Must contain non-whitespace characters.*'
+// violation 4 lines below 'Javadoc tag @incomplete=This class needs more code...'
 /**
  * Testing tag writing
  * @author Daniel Grenner
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagMissingFormat.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagMissingFormat.java
@@ -2,14 +2,14 @@
 WriteTag
 tag = @author
 tagFormat = (default)null
-tagSeverity = (default)info
+tagSeverity = error
 tokens = (default)INTERFACE_DEF, CLASS_DEF, ENUM_DEF, ANNOTATION_DEF, RECORD_DEF
 
 
 */
 
 package com.puppycrawl.tools.checkstyle.checks.javadoc.writetag;
-// violation 3 lines below , 'Missing @author tag.*'
+// violation 3 lines below 'Javadoc tag @author=Daniel Grenner'
 /**
  * Testing tag writing
  * @author Daniel Grenner
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagSeverity.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagSeverity.java
@@ -2,7 +2,7 @@
 WriteTag
 tag = @author
 tagFormat = \\S
-severity = ignore
+tagSeverity = error
 tokens = (default)INTERFACE_DEF, CLASS_DEF, ENUM_DEF, ANNOTATION_DEF, RECORD_DEF
 
 
@@ -10,7 +10,7 @@ tokens = (default)INTERFACE_DEF, CLASS_DEF, ENUM_DEF, ANNOTATION_DEF, RECORD_DEF
 
 package com.puppycrawl.tools.checkstyle.checks.javadoc.writetag;
 
-// violation 3 lines below , 'Must contain non-whitespace characters.*'
+// violation 3 lines below 'Javadoc tag @author=Daniel Grenner'
 /**
  * Testing tag writing
  * @author Daniel Grenner
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpSemantic2.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpSemantic2.java
@@ -10,7 +10,6 @@ ignoreComments = (default)false
 
 */
 // violation 11 lines above 'Required pattern .* missing in file.'
-// location of violation is first line of file
 package com.puppycrawl.tools.checkstyle.checks.regexp.regexp;
 
 import java.awt.*;
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpTrailingComment11.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpTrailingComment11.java
@@ -10,7 +10,7 @@ ignoreComments = true
 
 */
 // violation 11 lines above 'Required pattern .* missing'
-// violation due to ignoreComments=true
+// due to ignoreComments=true
 package com.puppycrawl.tools.checkstyle.checks.regexp.regexp;
 
 public class InputRegexpTrailingComment11 {
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator.java
@@ -12,18 +12,18 @@ tokens = (default)PACKAGE_DEF, IMPORT, STATIC_IMPORT, CLASS_DEF, INTERFACE_DEF, 
 
 package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator; //no violation: trailing comment
 import java.io.Serializable; // violation ''import' should be separated from previous line.'
-import java.util.ArrayList; /*no violation: trailing comment*/
+import java.util.ArrayList; /*ok: trailing comment*/
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.Collections;
-/* no violation: block comment after token*/
+/* ok: block comment after token*/
 
 import java.io.PrintWriter;
-//no violation: single-line comment after token
+//ok: single-line comment after token
 
-import javax.swing.AbstractAction; /* no violation: no trailing comment
+import javax.swing.AbstractAction; /* ok: no trailing comment
 */
 
 import org.apache.commons.beanutils.locale.converters.ByteLocaleConverter;
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator2.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator2.java
@@ -10,20 +10,20 @@ tokens = (default)PACKAGE_DEF, IMPORT, STATIC_IMPORT, CLASS_DEF, INTERFACE_DEF, 
 
 */
 
-package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator; //no violation: trailing comment
+package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator; //ok: trailing comment
 import java.io.Serializable; // violation ''import' should be separated from previous line.'
-import java.util.ArrayList; /*no violation: trailing comment*/
+import java.util.ArrayList; /*ok: trailing comment*/
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.Collections;
-/* no violation: block comment after token*/
+/* ok: block comment after token*/
 
 import java.io.PrintWriter;
-//no violation: single-line comment after token
+//ok: single-line comment after token
 
-import javax.swing.AbstractAction; /* no violation: no trailing comment
+import javax.swing.AbstractAction; /* ok: no trailing comment
 */
 
 import org.apache.commons.beanutils.locale.converters.ByteLocaleConverter;
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorNoViolationOnEmptyLineBeforeComments.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorNoViolationOnEmptyLineBeforeComments.java
@@ -16,7 +16,7 @@ import java.lang.Object; // no violation
 
 import java.lang.Class; // no violation
 
-// no violation
+// ok below
 import java.lang.Long;
 
 /*
@@ -31,12 +31,12 @@ import java.lang.Double;
 import java.lang.Object;
 
 // .
-// no violation
+// ok below
 import java.lang.Boolean;
-// no violation
+// ok below
 import java.lang.Byte;
 
-// no violation
+// ok below
 /* no violation */
 import java.lang.Short;
 import java.lang.Number;
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorSingleLineCommentAfterPackage.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorSingleLineCommentAfterPackage.java
@@ -12,7 +12,6 @@ tokens = (default)PACKAGE_DEF, IMPORT, STATIC_IMPORT, CLASS_DEF, INTERFACE_DEF, 
 
 package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator; // trailing comment is allowed.
 // violation ''//' should be separated from previous line.'
-// violation at line above
 
 import java.util.Map;
 
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorWithComments.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorWithComments.java
@@ -19,13 +19,13 @@ package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator; //
 
 import java.lang.Object; // violation ''import' has more than 1 empty lines before.'
 
-import java.lang.Class; // no violation
+import java.lang.Class;
 
 
 // violation ''//' has more than 1 empty lines before.'
 import java.lang.Integer;
 
-// no violation
+// ok, because no more than 1 empty lines before
 import java.lang.Long;
 
 
@@ -35,7 +35,7 @@ import java.lang.Long;
 import java.lang.Float;
 
 /*
- * no violation
+ * something
  */
 import java.lang.Double;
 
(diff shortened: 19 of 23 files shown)


SteLeo1602:
I'm not quite sure what to do with violation expectations lines. Do I work to match all violations location's expectations or do I fix something within the existing code to prevent those errors in CI?

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -615,6 +619,8 @@ private static void setViolations(TestInputConfiguration.Builder inputConfigBuil
                 VIOLATIONS_SOME_LINES_ABOVE_PATTERN.matcher(lines.get(lineNo));
         final Matcher violationsSomeLinesBelowMatcher =
                 VIOLATIONS_SOME_LINES_BELOW_PATTERN.matcher(lines.get(lineNo));
+        final Matcher violationsDefault =
+                VIOLATION_DEFAULT.matcher(lines.get(lineNo));
```
romani, line 623: add usage of it before `else if (useFilteredViolations) {`

something like:
```
else if (violationsDefault.matches()) {
            final int violationLineNum = lineNo + 1;
            inputConfigBuilder.addViolation(violationLineNum, null);
        }
else if (useFilteredViolations) {
```

romani (changes requested):
item

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -98,7 +98,7 @@ public final class InlineConfigParser {
 
     /** A pattern to find the string: "// violation, explanation". */
     private static final Pattern VIOLATION_WITH_EXPLANATION_PATTERN = Pattern
-            .compile(".*//\\s*violation,\\s.+\\s(?:['\"](.*)['\"])?$");
+            .compile("\".*//\\\\s*violation,\\\\s+(?:.*)?$\"");
```
rnveach, line 101: I don't know why, but you encase this string in another set of quotes and looks like you double escaped the escapes.

If you used Eclipse, I would say you pasted the old regexp and string inside the string here in order to bring the old PR forward. Newer versions of Eclipse will escape whatever you paste into a string in Java.

rnveach (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
SteLeo1602: Done

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
SteLeo1602: Fixed

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -706,6 +709,10 @@ else if (multipleViolationsBelowMatcher.matches()) {
                         inputConfigBuilder.addViolation(actualLineNumber, null);
                     });
         }
+        else if (violationsDefault.matches()) {
+                    final int violationLineNum = lineNo + 1;
+                    inputConfigBuilder.addViolation(violationLineNum, null);
```
romani, line 714: Please move code to left, to have no extra spaces on left.

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
romani: https://ci.appveyor.com/project/Checkstyle/checkstyle/builds/50256080/job/drp3uw8ceedrvos0#L111

Please always run `mvn verify` before pushing to GitHub, and resolve all violations from such execution

SteLeo1602: Done. Do I fix all of the parts where CI complains that violations differ?

romani: Yes, we need green CI.
```
Violation lines for C:\projects\checkstyle\src\test\resources\com\puppycrawl\tools\checkstyle\checks\javadoc\writetag\InputWriteTagSeverity.java differ.
missing (1): 13
---
expected   : [13]
but was    : []

```

Looks like there's is more problems, please investigate

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -177,6 +177,10 @@ public final class InlineConfigParser {
     private static final Pattern VIOLATIONS_SOME_LINES_BELOW_PATTERN = Pattern
             .compile(".*//\\s*(\\d+) violations (\\d+) lines below:$");
 
+    /** A pattern that matches any comment by default. */
+    private static final Pattern VIOLATION_DEFAULT = Pattern
+            .compile("//.*violation.*");
```
Zopsss, line 182: suspicious regex. This would match all single line comments which has `violation` word in it. We need to match only comments which starts with `violation`. 

@romani  correct me if I'm wrong

rnveach: > We need to match only comments which starts with violation

This is not what was agreed upon in issue.
See https://github.com/checkstyle/checkstyle/issues/15256#issuecomment-2232264814

romani: @rnveach, 

this become too aggressive and we have no way to use `violation` word at all
Example:
https://github.com/checkstyle/checkstyle/blob/f89958aa517585f932a012e3f876dfb361f3543f/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/nocodeinfile/Example1.java#L9-L12

What is your suggestions to rephrase comment ?

romani:
InputImportOrder_Wildcard

>// groups are configured as follows
// com.puppycrawl,*,java
// the trailing javax.crypto.Cipher; should be flagged as a violation.

such comments should be just removed.

romani:
```
✔ ~/java/github/romani/checkstyle [SteLeo1602/FixViolations L|✚ 1⚑ 2] 
 $ git diff
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagDoubleTag.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagDoubleTag.java
index 14c0d8c..8403c5b 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagDoubleTag.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagDoubleTag.java
@@ -2,15 +2,15 @@
 WriteTag
 tag = @doubletag
 tagFormat = \\S
-tagSeverity = (default)info
+tagSeverity = error
 tokens = (default)INTERFACE_DEF, CLASS_DEF, ENUM_DEF, ANNOTATION_DEF, RECORD_DEF
 
 
 */
 
 package com.puppycrawl.tools.checkstyle.checks.javadoc.writetag;
-// violation 6 lines below , '@doubletag appeared at the same time.*'
-// violation 6 lines below , '@doubletag appeared at the same time.*'
+// violation 6 lines below 'Javadoc tag @doubletag=first text'
+// violation 6 lines below 'Javadoc tag @doubletag=second text'
 /**
  * Testing tag writing
  * @author Daniel Grenner

```

romani:
```
✔ ~/java/github/romani/checkstyle [SteLeo1602/FixViolations L|✚ 4⚑ 2] 
22:32 $ git diff
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpSemantic2.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpSemantic2.java
index dd18428..80afa66 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpSemantic2.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpSemantic2.java
@@ -10,7 +10,6 @@ ignoreComments = (default)false
 
 */
 // violation 11 lines above 'Required pattern .* missing in file.'
-// location of violation is first line of file
 package com.puppycrawl.tools.checkstyle.checks.regexp.regexp;
 
 import java.awt.*;
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpTrailingComment11.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpTrailingComment11.java
index 5bff0fd..771773e 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpTrailingComment11.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexp/InputRegexpTrailingComment11.java
@@ -10,7 +10,7 @@ ignoreComments = true
 
 */
 // violation 11 lines above 'Required pattern .* missing'
-// violation due to ignoreComments=true
+// due to ignoreComments=true
 package com.puppycrawl.tools.checkstyle.checks.regexp.regexp;
 
 public class InputRegexpTrailingComment11 {
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport.java
index b5436bb..c1e0920 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport.java
@@ -19,7 +19,7 @@ package com.puppycrawl.tools.checkstyle.checks.regexp.regexpmultiline;
  */
 public class InputRegexpMultilineMultilineSupport {
     void method() { // violation below
-// abc - violation
+// abc
 // def
 // abc
     }
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport2.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport2.java
index 3ceb5f3..74e3376 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport2.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport2.java
@@ -19,9 +19,10 @@ package com.puppycrawl.tools.checkstyle.checks.regexp.regexpmultiline;
  */
 public class InputRegexpMultilineMultilineSupport2 {
     void method() { // violation below
-// abc - violation
-// def // violation below
-// abc - violation
+// abc
+// def
+// abc
+        // violation above
     }
 
     void method2() {

```

romani:
```
✔ ~/java/github/romani/checkstyle [SteLeo1602/FixViolations L|✚ 6⚑ 2] 
22:45 $ git diff
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator.java
index 8391313..e10ed6d 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator.java
@@ -12,18 +12,18 @@ tokens = (default)PACKAGE_DEF, IMPORT, STATIC_IMPORT, CLASS_DEF, INTERFACE_DEF,
 
 package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator; //no violation: trailing comment
 import java.io.Serializable; // violation ''import' should be separated from previous line.'
-import java.util.ArrayList; /*no violation: trailing comment*/
+import java.util.ArrayList; /*ok: trailing comment*/
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.Collections;
-/* no violation: block comment after token*/
+/* ok: block comment after token*/
 
 import java.io.PrintWriter;
-//no violation: single-line comment after token
+//ok: single-line comment after token
 
-import javax.swing.AbstractAction; /* no violation: no trailing comment
+import javax.swing.AbstractAction; /* ok: no trailing comment
 */
 
 import org.apache.commons.beanutils.locale.converters.ByteLocaleConverter;
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator2.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator2.java
index 1d7f4e2..69823e6 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator2.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparator2.java
@@ -21,7 +21,7 @@ import java.util.Collections;
 /* no violation: block comment after token*/
 
 import java.io.PrintWriter;
-//no violation: single-line comment after token
+//ok: single-line comment after token
 
 import javax.swing.AbstractAction; /* no violation: no trailing comment
 */
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorSingleLineCommentAfterPackage.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorSingleLineCommentAfterPackage.java
index 6ac871a..ac46184 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorSingleLineCommentAfterPackage.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorSingleLineCommentAfterPackage.java
@@ -12,7 +12,7 @@ tokens = (default)PACKAGE_DEF, IMPORT, STATIC_IMPORT, CLASS_DEF, INTERFACE_DEF,
 
 package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator; // trailing comment is allowed.
 // violation ''//' should be separated from previous line.'
-// violation at line above
+
 
 import java.util.Map;
 
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorWithComments.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorWithComments.java
index 29b3738..0069ff2 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorWithComments.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorWithComments.java
@@ -19,7 +19,7 @@ package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator; //
 
 import java.lang.Object; // violation ''import' has more than 1 empty lines before.'
 
-import java.lang.Class; // no violation
+import java.lang.Class; // ok
 
 
 // violation ''//' has more than 1 empty lines before.'
@@ -35,7 +35,7 @@ import java.lang.Long;
 import java.lang.Float;
 
 /*
- * no violation
+ * something
  */
 import java.lang.Double;
 
@@ -46,29 +46,29 @@ import java.lang.Double;
 import java.lang.Character;
 
 /**
- * no violation
+ * something
  */
 import java.lang.String;
 
 /*
- * no violation
+ * something
  */
-/* no violation */
+/* something */
 import java.lang.Object;
 
 
 // violation ''//' has more than 1 empty lines before.'
 // .
 import java.lang.Boolean;
-// no violation
+// ok
 import java.lang.Byte;
 
-// no violation
-/* no violation */
+// ok
+/* something */
 import java.lang.Short;
 
 
-/* violation */
+/* something */
 // violation above ''/\*' has more than 1 empty lines before.'
 import java.lang.Number;
 
@@ -76,7 +76,7 @@ import java.lang.Number;
 import java.lang.Runnable; // violation ''import' has more than 1 empty lines before.'
 import java.lang.Thread;
 
-// no violation
+// ok
 
 
 // violation ''//' has more than 1 empty lines before.'
@@ -93,13 +93,13 @@ public class InputEmptyLineSeparatorWithComments {
 
     public int testViolationWithoutComment = 1; // violation ''VARIABLE_DEF' .*1 empty lines .*'
 
-    public int testNoViolationWithoutComment = 2; // no violation
+    public int testNoViolationWithoutComment = 2; // ok
 
 
     // violation ''//' has more than 1 empty lines before.'
     public int testViolationWithSingleLineComment = 3;
 
-    // Should not have violations
+    // Should be ok
     public int testNoViolationWithSingleLineComment = 4;
 
 
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorWithComments2.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorWithComments2.java
index af681f9..fc0c6c5 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorWithComments2.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorWithComments2.java
@@ -19,56 +19,56 @@ package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator; //
 
 import java.lang.Object; // ok
 
-import java.lang.Class; // no violation
+import java.lang.Class; // ok
 
 
 // ok
 import java.lang.Integer;
 
-// no violation
+// ok
 import java.lang.Long;
 
 
 /*
- * violation
+ * javadoc
  */
 import java.lang.Float;
 
 /*
- * no violation
+ * javadoc
  */
 import java.lang.Double;
 
 
 /**
- * violation
+ * javadoc
  */
 import java.lang.Character;
 
 /**
- * no violation
+ * javadoc
  */
 import java.lang.String;
 
 /*
- * no violation
+ * javadoc
  */
-/* no violation */
+/* javadoc */
 import java.lang.Object;
 
 
 // ok
 // .
 import java.lang.Boolean;
-// no violation
+// ok
 import java.lang.Byte;
 
-// no violation
-/* no violation */
+// ok
+/* javadoc */
 import java.lang.Short;
 
 
-/* violation */
+/* ok javadoc */
 // .
 import java.lang.Number;
 
@@ -76,7 +76,7 @@ import java.lang.Number;
 import java.lang.Runnable;
 import java.lang.Thread;
 
-// no violation
+// ok
 
 
 // ok
@@ -84,7 +84,7 @@ import java.lang.StringBuilder;
 
 
 /*
- * violation
+ * javadoc
  */
 import static java.lang.Math.abs;
 
@@ -105,26 +105,26 @@ public class InputEmptyLineSeparatorWithComments2 {
 
     /*
      * Should have
-     * violation
+     * somthing
      */
     public int testViolationWithMultilineComment = 5;
 
     /*
      * Should not have
-     * violation
+     * somthing
      */
     public int testNoViolationWithMultilineComment = 6;
 
 
     /**
      * Should have
-     * violation
+     * someting
      */
     public int testViolationWithJavadoc = 7;
 
     /**
      * Should not have
-     * violation
+     * somthing
      */
     public int testNoViolationWithJavadoc = 8;
 
@@ -133,29 +133,29 @@ public class InputEmptyLineSeparatorWithComments2 {
     } // ok
 
     public void testNoViolationWithoutComment() {
-    } // no violation
+    } // ok
 
 
-    // Should have violation
+    // Should be ok
     public void testViolationWithSingleLineComment() {
     }
 
     // Should not have
-    // a violation
+    // a somthing
     public void testNoViolationWithSingleLineComment() {
     }
 
 
     /*
      * Should have
-     * violation
+     * somthing
      */
     public void testViolationWithMultilineComment() {
     }
 
     /*
      * Should not have
-     * violation
+     * something
      */
     public void testNoViolationWithMultilineComment() {
     }
@@ -163,14 +163,14 @@ public class InputEmptyLineSeparatorWithComments2 {
 
     /**
      * Should have
-     * violation
+     * something
      */
     public void testViolationWithJavadoc() {
     }
 
     /**
      * Should not have
-     * violation
+     * something
      */
     public void testNoViolationWithJavadoc() {
     }
@@ -180,7 +180,7 @@ public class InputEmptyLineSeparatorWithComments2 {
 
     public static class Class2 { } // ok
 
-    // no violation
+    // ok
     public static class Class3 { }
 
 
@@ -190,46 +190,46 @@ public class InputEmptyLineSeparatorWithComments2 {
 
     // ok
     public
-    // no violation
+    // ok
     static class Class5 { }
 
 
     // ok
     public
-    /* no violation */
+    /* javadoc */
     static class Class6 { }
 
     /*
      * Should not have
-     * violation
+     * something
      */
     public static class Class7 { }
 
 
     /*
      * Should have
-     * violation
+     * something
      */
     public static class Class8 { }
 
     /**
      * Should not have
-     * violation
+     * something
      */
     public static class Class9 { }
 
 
     /**
      * Should have
-     * violation
+     * something
      */
     public static class Class10 {
         {
-            // no violation
+            // ok
         }
     }
 
-    // no violation
+    // ok
     public interface Interface1 { }
 
 
@@ -241,15 +241,15 @@ public class InputEmptyLineSeparatorWithComments2 {
     interface Interface3 { }
 
 
-    /* violation */
+    /* something */
     /* . */
     /* . */
     /* . */
     interface Interface4 { }
 
-    // no violation
+    // ok
 
-    // no violation
+    // ok
     // .
     // .
     // .
@@ -266,14 +266,14 @@ public class InputEmptyLineSeparatorWithComments2 {
 
 
     // ok
-    // no violation
+    // ok
 
-    // no violation
+    // ok
     public enum Enum3 { }
-    // no violation
+    // ok
 
     public enum Enum4 { }
-    // no violation
+    // ok
 
     public enum Enum5 { }
 
@@ -294,7 +294,7 @@ public class InputEmptyLineSeparatorWithComments2 {
         abs(2);
     }
 
-    // no violation
+    // ok
 
 
     // ok
@@ -302,7 +302,7 @@ public class InputEmptyLineSeparatorWithComments2 {
        abs(1);
     }
 
-    /* no violation */
+    /* something */
     { }
 
 
@@ -318,24 +318,24 @@ public class InputEmptyLineSeparatorWithComments2 {
         testNoViolationWithJavadoc = 1;
     }
 
-    // no violation
+    // ok
     /* . */ public InputEmptyLineSeparatorWithComments2(int i) {
         testNoViolationWithJavadoc = 1;
     }
 
-    // no violation
-    // no violation
+    // ok
+    // ok
 
-    // no violation
+    // ok
     public InputEmptyLineSeparatorWithComments2(int i, int j) {
         testNoViolationWithJavadoc = 1;
     }
 
 
     // ok
-    // no violation
+    // ok
 
-    // no violation
+    // ok
 
 
 
diff --git a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/packageinfo/test4/package-info.java b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/packageinfo/test4/package-info.java
index 23fbb2d..23b6a3e 100644
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/packageinfo/test4/package-info.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/packageinfo/test4/package-info.java
@@ -11,7 +11,7 @@ tokens = (default)PACKAGE_DEF, IMPORT, STATIC_IMPORT, CLASS_DEF, INTERFACE_DEF,
 */
 
 //for test there's warning when single-line comment isn't separated from PACKAGE_DEF by line
-//violation is expected after the next line. Starting * on next comment is to kill mutation
+//Starting * on next comment is to kill mutation
 //*because utility doesn't check javadoc but any comment with starting *.
 package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator.packageinfo.test4; // violation ''package' should be separated from previous line.'
 

```

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
romani: workaround is `/* the violation is on first line of file */`

I am ok to use this workaround/hack.

romani:
to fix problems in Filters
```
✔ ~/java/github/romani/checkstyle [SteLeo1602/FixViolations L|✚ 1⚑ 2] 
23:01 $ git diff
diff --git a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
index dd3b1e7..3249bf9 100644
--- a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
@@ -709,14 +709,14 @@ public final class InlineConfigParser {
                         inputConfigBuilder.addViolation(actualLineNumber, null);
                     });
         }
-        else if (violationsDefault.matches()) {
-            final int violationLineNum = lineNo + 1;
-            inputConfigBuilder.addViolation(violationLineNum, null);
-        }
         else if (useFilteredViolations) {
             setFilteredViolation(inputConfigBuilder, lineNo + 1,
                     lines.get(lineNo), specifyViolationMessage);
         }
+        else if (violationsDefault.matches()) {
+            final int violationLineNum = lineNo + 1;
+            inputConfigBuilder.addViolation(violationLineNum, null);
+        }
     }

```

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
Zopsss: we can face this problem in other files too. we will need to use this workaround everywhere.

Zopsss:
> such comments should be just removed.

I think these comments should stay as they're explaining the check's behavior. Can we just convert them to block comments instead, as you suggested here: https://github.com/checkstyle/checkstyle/pull/15327#discussion_r1692541776?

romani:
We can , but such comments are from time when we don't have BDD and had to jump between input file and Java test to see where is violation and what is config, right now all at one place: config slightly above and violation pointer a line below.

It is duplicate information now.

Zopsss:
Ohh, I see. If that's the case then it is okay to remove them 👍🏻

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
rnveach: > this become too aggressive and we have no way to use violation word at all.

This issue will never be truly be "resolved" unless we have a catch all. This issue started because of a violation message in an input file which didn't follow our patterns and as such was missed. What is to say someone won't do something similar in the future. We say violation is a keyword to all our BDDs, but we don't want to make it truly a keyword.

If we want to drop this and not do a catch all, then confirm with @nrmancuso for his approval and update the issue.

> What is your suggestions to rephrase comment ?

My current suggestion would be to update our BDD to use a less common word.

romani: I am good with workaround.
No concerns for now.

nrmancuso: I am good with ignoring block comments, as long as we have no valid violation comments in them anywhere

romani:
@SteLeo1602, please fix CI failures

SteLeo1602:
> @SteLeo1602, please fix CI failures

Done

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/nocodeinfile/Example1.java
```diff
@@ -9,7 +9,6 @@
 // violation 8 lines above
 
 // xdoc section -- start
-// the violation is on first line of file
```
romani, line 12: `/* the violation is on first line of file */`

same for all files in nocodeinfile folder

romani (changes requested):
last:

romani:
Github, generate website

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/nocodeinfile/Example1.java
romani: fyi - https://checkstyle.org/checks/misc/nocodeinfile.html#Example2-code

SteLeo1602: Done

SteLeo1602:
Github, generate website

romani (approved):
Ok to merge

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -649,11 +655,8 @@ else if (violationBelowWithExplanationMatcher.matches()) {
             inputConfigBuilder.addViolation(violationLineNum, violationMessage);
         }
         else if (violationWithExplanationMatcher.matches()) {
-            final String violationMessage = violationWithExplanationMatcher.group(1);
             final int violationLineNum = lineNo + 1;
-            checkWhetherViolationSpecified(specifyViolationMessage, violationMessage,
-                    violationLineNum);
-            inputConfigBuilder.addViolation(violationLineNum, violationMessage);
+            inputConfigBuilder.addViolation(violationLineNum, null);
```
rnveach, line 659: Why was this changed?

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
```diff
@@ -712,6 +715,10 @@ else if (useFilteredViolations) {
             setFilteredViolation(inputConfigBuilder, lineNo + 1,
                     lines.get(lineNo), specifyViolationMessage);
         }
+        else if (violationsDefault.matches()) {
+            final int violationLineNum = lineNo + 1;
+            inputConfigBuilder.addViolation(violationLineNum, null);
```
rnveach, line 720: https://github.com/checkstyle/checkstyle/issues/15256#issuecomment-2233441811
> match all things violation, then figure out which it is, or throw an exception if it doesn’t match

This code makes it look like we capture this as a violation and continue on. I was expecting a blanket throw exception. Am I missing something? We prefer to see the violation message and not just the word violation by itself.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/equalshashcode/InputEqualsHashCodeEqualsParameter.java
```diff
@@ -12,15 +12,15 @@ public boolean equals(String o) {
...
         public int hashCode() { // violation 'without .* of 'equals()'.'
             return 1;
         }
         public boolean equals(String o) {
             return true;
         }
     }
-    public static class TestClass3 { // violation, no `hashCode`
+    public static class TestClass3 {
         public boolean equals(Object o) { // violation 'without .* of 'hashCode()'.'
```
rnveach, line 24: `VIOLATION_WITH_EXPLANATION_PATTERN` has a comma while these don't, so the new regexp doesn't match these.

Why are we updating the regexp if it is not needed for these to match?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTag.java
```diff
@@ -2,14 +2,14 @@
 WriteTag
 tag = @author
 tagFormat = \\S
-tagSeverity = (default)info
+tagSeverity = error
```
rnveach, line 5: Why did this change?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport.java
```diff
@@ -19,7 +19,7 @@
  */
 public class InputRegexpMultilineMultilineSupport {
     void method() { // violation below
-// abc - violation
+// abc
```
rnveach, line 22: I think it would make more sense to remove the `violation below` and keep the single line `violation`.

rnveach (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
romani: We support violation by itself. There are a lot of Inputs with just `// violation `

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
romani: To fix a problem that we started from. Without it migration of example can not happen

rnveach: This doesn't make things clearer. We said we missed violation from bad violation text in input and changed regexp to match. Why does this need to change as well?

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
rnveach: How did we support them when we just added this regexp? Show me we will get an exception with a violation comment that doesn't match our other regexps like issue talked about.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTag.java
romani: Because this input started to work. And this update to make match what is in input

src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport2.java
```diff
@@ -19,9 +19,10 @@
  */
 public class InputRegexpMultilineMultilineSupport2 {
     void method() { // violation below
-// abc - violation
-// def // violation below
-// abc - violation
+// abc
+// def
+// abc
+// violation above
```
Zopsss, line 25: It would be good if we make it single line violation instead of using violation above/below as rnveach suggested here: https://github.com/checkstyle/checkstyle/pull/15327/files#r1699267278

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
romani: comparison of 
https://github.com/checkstyle/checkstyle/blob/e071377c4d6e049491aef610b04bc106e8a8b6a3/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L80-L81
vs
https://github.com/checkstyle/checkstyle/blob/e071377c4d6e049491aef610b04bc106e8a8b6a3/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L100-L101
```
.compile(".*//\\s*violation,?\\s*(?:['\"](.*)['\"])?$")
.compile(".*//\\s*violation,\\s.+\\s(?:['\"](.*)['\"])?$");

with extra spaces (to spot a difference):
.compile(".*//\\s*violation,?\\s*   (?:['\"](.*)['\"])?$")
.compile(".*//\\s*violation,\\s.+\\s(?:['\"](.*)['\"])?$");
```

so VIOLATION_WITH_EXPLANATION_PATTERN kind of copy paste, and it demand message `'some text' or "some text"` to be present, but role of it to allow any text without `"` or `'`

we changing VIOLATION_WITH_EXPLANATION_PATTERN to do no message match, and handling of this pattern should have null in violation message parameter - no matching by message will happen.

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
romani: we always supported it , and still allow if message is single in check - https://github.com/checkstyle/checkstyle/commit/a3c1783c90371f582d853d2e1d490f1ff847e30e#diff-7059762335dfd1e655cf558d06cfbe29db28808a2c154035b57c9d8dd2026740R10

but if people did typo , like extra space or some other, it did not match. What we do now is aggressive match of any "violation" word at any place to expected violation, so in this case we eliminating any possible typos or ...

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/equalshashcode/InputEqualsHashCodeEqualsParameter.java
romani: it match https://github.com/checkstyle/checkstyle/blob/e071377c4d6e049491aef610b04bc106e8a8b6a3/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L80-L81

attentioon to `,?`

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTag.java
romani: this check is most bad from design perspective (design of Check) I always struggle to understand how it should work. One day we will deprecate it, but for now we support it as is, some people use it as it has feature that non of others our Check can do, I do not remember which feature.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport.java
romani: we matching regexp over comments, lets keep it, it is just a test.
`violation below` is used to not overlap comment marker with target code(code comments).

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/equalshashcode/InputEqualsHashCodeEqualsParameter.java
rnveach: Which is a different pattern and now brings up a separate question. Why do we have these similar patterns? We removed quotes from the other one so why are we keeping it in this one?

IMO, very similar patterns with no real reason on which to use will just confuse users. I am fine with this being a separate issue, but it should be an issue.

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
rnveach: 1) This doesn't change from the issue about `throw an exception if it doesn’t match`.
2) It doesn't explain how we handled `// violation` without the regular expression we just added now.
3) Why didn't this catch the issue's violation message.

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
rnveach: 1) If `VIOLATION_WITH_EXPLANATION_PATTERN` no longer has an explanation matching, then we need to remove it from the field name.
2) I thought we wanted violation message matching as part of BDD? I remember seeing an issue to convert other inputs to this format. Others have violation message matching and it seems like this is the only one who won't, so again, why are we doing this?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport.java
rnveach: Violation below text does not match `VIOLATION_BELOW_PATTERN` or `VIOLATION_BELOW_WITH_EXPLANATION_PATTERN` which brings up a point in a similar comment about why we aren't throwing an exception on wrong violation messages.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/equalshashcode/InputEqualsHashCodeEqualsParameter.java
romani: They are separate.
With quotes is more strict as it demands quoted text to match violation message.

Without quotes is free text , whatever author think is reasonable to share that is not a message.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport.java
romani: No exception - Because we now has most aggressive match by word `violation` at any place.
And it demands it to be declared in expected array, and author with not be able to put nonsense there, as expected array is line, column, and message match.

rnveach: That is not what 2 of 3 admins asked for in issue.

https://github.com/checkstyle/checkstyle/issues/15256#issuecomment-2231354613
> We should identify anything written as // violation as a violation and then fail with an error if we can't match it with a specific message like VIOLATION_WITH_EXPLANATION_PATTERN

https://github.com/checkstyle/checkstyle/issues/15256#issuecomment-2233441811
> Ideally we match all things violation, then figure out which it is, or throw an exception if it doesn’t match.

src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
romani: > This doesn't change from the issue about throw an exception if it doesn’t match

done.

romani: explanation is at https://github.com/checkstyle/checkstyle/pull/15021#discussion_r1657176223
I am posting here one more time.
image
it is how regexp it written.

> If we supported bland violation matching, why didn't this catch the issue's violation message?

it was not bland, it was match to simple strict `// violation` or `// violation, 'message'` or  `// violation, "message"`

src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport.java
romani: excetion is placed.

>Google IT no longer has an expected array, and it uses our BDD now (right?).

it uses, and it uses messages in all places as we have numerous Checks and message is always required (not enforced automatically) as it is not clear what violation is there.

romani: > I think it would make more sense to remove the violation below and keep the single line violation.

please give exact code to use, I do not mind to use any there.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/regexp/regexpmultiline/InputRegexpMultilineMultilineSupport2.java
SteLeo1602: Done

romani:
@rnveach , please finish review.

rnveach:
This PR has too much going on, not enough answers, and doesn't cover what was asked for in issue. I would look for a way to trim this into smaller PRs.

We are expecting an exception if we have a line with "violation" and it doesn't match any of our other regular expressions. I still am not seeing this.

I can scan through and see me asking question and not getting answers or getting vague answers which don't satisfy me. Nothing has changed since the last time I posted/reviewed.

All this is blocking this PR.

romani:
this PR is already a split to make another a bit easier to consume. It is kind of blocker for https://github.com/checkstyle/checkstyle/pull/15021 and any similar updates that outside contributors helps us to do. 
I think we goes too deep in ideal design of of this model, this is not a goal of this issue.
We need to create new issue and start conceptual conversation on how it should work.
This is not external functionality we can improve at any time, nobody will be affected.

romani:
discussion on what is better plan for matching is moved to https://github.com/checkstyle/checkstyle/issues/15643

PR #15334, opened 2024-07-21
Issue #15322: NeedBraces check should violate switch expressions with a…

### LINKED ISSUE #15322: NeedBraces Check Should violate Switch Expressions with Arrow Syntax That Directly Yields a Value
detected at https://github.com/checkstyle/checkstyle/pull/15321#discussion_r1684288238
I have read check documentation:https://checkstyle.org/checks/blocks/needbraces.html#NeedBraces

```
D:\CS\test
javac src/Test.java
D:\CS\test
cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
  <module name="TreeWalker">
    <module name="NeedBraces">
      <property name="tokens" value="LAMBDA, LITERAL_CASE"/>
    </module>
  </module>
</module>
D:\CS\test
cat src/Test.java
import java.util.Arrays;
import java.util.List;

public class Test {
    int test(Object o, String z) {
        List<Integer> numbers = Arrays.asList(1, 2, 3);
        numbers = numbers.stream()
                .map(x -> 0) // violation
                .toList();

        numbers = numbers.stream()
                .map(x -> z.length()) // violation
                .toList();

        return switch (o) {
            case Integer _ -> 5; // should be violation same as the line below and lambda examples above
            case String s -> s.length(); // violation
            default -> 0;
        };
    }
}
D:\CS\test
Starting audit...
[ERROR] D:\CS\test\src\Test.java:8:24: '->' construct must use '{}'s. [NeedBraces]
[ERROR] D:\CS\test\src\Test.java:12:24: '->' construct must use '{}'s. [NeedBraces]
[ERROR] D:\CS\test\src\Test.java:17:13: 'case' construct must use '{}'s. [NeedBraces]
Audit done.
z

```

---

I expect a violation on line 16. We can add braces and the `yield` keyword. We should be consistent with Lambdas. We violate them and the fix will be add braces and the `return` keyword. If some users don't want to violate these examples they should enable `allowSingleLineStatement`.

 ---

mahfouz72:
closes #15322 

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/adbff014913347bd87c907aad81c7fec584f71f8/check.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java
@@ -177,8 +177,7 @@ public class NeedBracesCheck extends AbstractCheck {
                 break;
             case TokenTypes.LITERAL_CASE:
             case TokenTypes.LITERAL_DEFAULT:
-                result = hasUnbracedStatements(ast)
-                    && !isSwitchLabeledExpression(ast);
+                result = hasUnbracedStatements(ast);
                 break;
             case TokenTypes.LITERAL_ELSE:
                 result = ast.findFirstToken(TokenTypes.LITERAL_IF) == null;
@@ -416,35 +415,6 @@ public class NeedBracesCheck extends AbstractCheck {
      */
     private static boolean isInSwitchRule(DetailAST ast) {
         return ast.getParent().getType() == TokenTypes.SWITCH_RULE;
-    }
-
-    /**
-     * Checks if current expression is a switch labeled expression. If so,
-     * braces are not allowed e.g.:
-     * <p>
-     * {@code
-     * case 1 -> 4;
-     * }
-     * </p>
-     *
-     * @param ast the ast to check
-     * @return true if current expression is a switch labeled expression.
-     */
-    private static boolean isSwitchLabeledExpression(DetailAST ast) {
-        final DetailAST parent = ast.getParent();
-        return switchRuleHasSingleExpression(parent);
-    }
-
-    /**
-     * Checks if current switch labeled expression contains only a single expression.
-     *
-     * @param switchRule {@link TokenTypes#SWITCH_RULE}.
-     * @return true if current switch rule has a single expression.
-     */
-    private static boolean switchRuleHasSingleExpression(DetailAST switchRule) {
-        final DetailAST possibleExpression = switchRule.findFirstToken(TokenTypes.EXPR);
-        return possibleExpression != null
-                && possibleExpression.getFirstChild().getFirstChild() == null;
     }
 
     /**
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitchAllowSingleLine.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitchAllowSingleLine.java
@@ -84,10 +84,10 @@ public class InputNeedBracesPatternMatchingForSwitchAllowSingleLine {
                 yield 0;
         };
 
-        int b =  switch (o) {
+        int b = switch (o) {
             case Integer i when (i == 0) -> i;
             case String s when (s.equals("a")) -> s.length();
-            case Point(int x, int y ) when (x>=0 && y >=0) ->  // violation
+            case Point(int x, int y ) when (x>=0 && y >=0) -> // violation
                 x + y;
             case Integer i when (i == 9) -> {
                 int n = i;
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheckTest.java
@@ -246,7 +246,10 @@ public class NeedBracesCheckTest extends AbstractModuleTestSupport {
             "59:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
             "73:47: " + getCheckMessage(MSG_KEY_NEED_BRACES, "->"),
             "80:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
-            };
+            "87:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "88:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "89:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+        };
         verifyWithInlineConfigParser(
                 getNonCompilablePath("InputNeedBracesTestSwitchExpressionNoSingleLine.java"),
             expected);
@@ -291,8 +294,10 @@ public class NeedBracesCheckTest extends AbstractModuleTestSupport {
             "78:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
             "80:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
             "83:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
+            "88:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "89:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
             "90:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
-            "91:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "case"),
+            "96:13: " + getCheckMessage(MSG_KEY_NEED_BRACES, "default"),
         };
         verifyWithInlineConfigParser(
                 getNonCompilablePath("InputNeedBracesPatternMatchingForSwitch.java"),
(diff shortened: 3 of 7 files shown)


mahfouz72:
Github, generate report for NeedBraces/all-examples-in-one

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitchAllowSingleLine.java
```diff
@@ -84,10 +84,11 @@ case Point(int x, int y ) when (x>=0 && y >=0) :  // violation
                 yield 0;
         };
 
-        int b =  switch (o) {
+        int b;
+        b =  switch (o) {
```
mahfouz72, line 88: This change is for code coverage line 439 (  `|| switchAST.getParent().getParent().getType() == TokenTypes.EXPR;`)

mahfouz72:
@nrmancuso report is good and the PR ready for review

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitchAllowSingleLine.java
nrmancuso: What if we have doubly nested switch expressions? Please add this to the new test cases.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitchAllowSingleLine.java
```diff
@@ -84,10 +84,11 @@ case Point(int x, int y ) when (x>=0 && y >=0) :  // violation
...
             case Integer i when (i == 0) -> i;
             case String s when (s.equals("a")) -> s.length();
-            case Point(int x, int y ) when (x>=0 && y >=0) ->  // violation
+            case Point(int x, int y ) when (x>=0 && y >=0) ->
```
nrmancuso, line 91: Please add new tests cases, using all brace options, on input like this:
```java
public class Test {
    
    int m1(int x) {
        return switch (x) {
            case 1 -> 1;
            case 2 -> 2;
            default -> throw new IllegalStateException();
        };
    }

    int m2(int x) {
        return switch (x) {
            case 1 -> { yield 1; }
            case 2 -> { yield 2; }
            default -> { throw new IllegalStateException(); }
        };
    }
}

```

So that we can see the expected behavior for both switch labeled expressions AND blocks in their equivalent forms.

nrmancuso (changes requested):
Items:

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitchAllowSingleLine.java
mahfouz72: done added to the new input

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/blocks/needbraces/InputNeedBracesPatternMatchingForSwitchAllowSingleLine.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java
```diff
@@ -419,32 +419,26 @@ private static boolean isInSwitchRule(DetailAST ast) {
...
-     */
-    private static boolean switchRuleHasSingleExpression(DetailAST switchRule) {
-        final DetailAST possibleExpression = switchRule.findFirstToken(TokenTypes.EXPR);
-        return possibleExpression != null
-                && possibleExpression.getFirstChild().getFirstChild() == null;
+        final DetailAST caseParent = ast.getParent();
+        final DetailAST switchAST = caseParent.getParent();
+        final boolean isSwitchExpression = switchAST.getParent().getType() == TokenTypes.EXPR
+                || switchAST.getParent().getParent().getType() == TokenTypes.EXPR;
+        final boolean caseParentHasExpression = caseParent.findFirstToken(TokenTypes.EXPR) != null;
```
nrmancuso, line 440:
```suggestion
        final boolean isChildOfExpression = caseParent.findFirstToken(TokenTypes.EXPR) != null;
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java
```diff
@@ -419,32 +419,26 @@ private static boolean isInSwitchRule(DetailAST ast) {
...
-     *
-     * @param switchRule {@link TokenTypes#SWITCH_RULE}.
-     * @return true if current switch rule has a single expression.
-     */
-    private static boolean switchRuleHasSingleExpression(DetailAST switchRule) {
-        final DetailAST possibleExpression = switchRule.findFirstToken(TokenTypes.EXPR);
-        return possibleExpression != null
-                && possibleExpression.getFirstChild().getFirstChild() == null;
+        final DetailAST caseParent = ast.getParent();
+        final DetailAST switchAST = caseParent.getParent();
```
nrmancuso, line 437: Hm, I figured this would give us some checkstyle violations, please do `switchAst`

nrmancuso (changes requested):
Last from me:

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java
rnveach: `AST` is approved by our config.
https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle-checks.xml#L819

Just look at our AST class if you need to know why we did that. Hint: `DetailAST`

romani: Better to use `Ast`.
Let's not grow bad wording, even it is allowed. 

We might need to suppress certain violations to let it be forbidden for new code. But this could be code out side of this PR.

nrmancuso (dismissed):

src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java
```diff
@@ -419,32 +419,26 @@ private static boolean isInSwitchRule(DetailAST ast) {
...
-     * Checks if current expression is a switch labeled expression. If so,
-     * braces are not allowed e.g.:
-     * <p>
-     * {@code
-     * case 1 -> 4;
+     * Checks if the current case or default label is a switch labeled expression.
+     * If so, braces are not allowed e.g.:
+     * <pre>
+     * return switch(x) {
+     *    case Integer i -&gt; 3; // no braces allowed
```
rnveach, line 426: Braces are allowed if you add a yield:
````
		return switch (x) {
		case Integer i -> {
			yield 3;
		}
````

Isn't this the same with lambdas were you have to add a return if you specifically add the braces since it was implied before?

Comment (mahfouz72) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java:426]: > Braces are allowed if you add a yield:

Yes, but we should not enforce this code to have braces for a single return value like this. because if we want to fix this violation we will have to add **unnecessary** `yield`.

Comment (mahfouz72) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java:426]: > Isn't this the same with lambdas were you have to add a return if you specifically add the braces since it was implied before?

Yes actually
[

](url)

we can remove braces and return from lambda but we will have a check violation
The check doesn't allow us to use the expression lambda. I don't think it is a good idea to violate this kind of code (code where we need to add unnecessary keywords such as return and yield to satisfy the check)

Comment (rnveach) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java:426]: If our check supports lambdas and does the same enforcement with lambdas and needing to write `return`s, then I question why this issue is valid for switches but not for lambdas.
https://checkstyle.org/checks/blocks/needbraces.html#Example6-config

I didn't provide comments on approval of the issue, so I am now asking to reconsider why this issue is valid.
@nrmancuso

Comment (rnveach) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java:426]: The reason we added lambda support was to make it clear where the lambda started and ended if it spanned more than a single line.

Comment (nrmancuso) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java:426]: Check behavior should be consistent between lambdas/blocks/returns and switch rule/blocks/yields IMO

Comment (mahfouz72) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/NeedBracesCheck.java:426]:
```
D:\CS\test
javac src/Test.java
D:\CS\test
cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
  <module name="TreeWalker">
    <module name="NeedBraces">
      <property name="tokens" value="LAMBDA, LITERAL_CASE"/>
    </module>
  </module>
</module>
D:\CS\test
cat src/Test.java
import java.util.Arrays;
import java.util.List;

public class Test {
    int test(Object o, String z) {
        List<Integer> numbers = Arrays.asList(1, 2, 3);
        numbers = numbers.stream()
                .map(x -> 0) // violation
                .toList();

        numbers = numbers.stream()
                .map(x -> z.length()) // violation
                .toList();

        return switch (o) {
            case Integer _ -> 5; // should be violation same as line below and lambda examples above
            case String s -> s.length(); // violation
            default -> 0;
        };
    }
}
D:\CS\test
Starting audit...
[ERROR] D:\CS\test\src\Test.java:8:24: '->' construct must use '{}'s. [NeedBraces]
[ERROR] D:\CS\test\src\Test.java:12:24: '->' construct must use '{}'s. [NeedBraces]
[ERROR] D:\CS\test\src\Test.java:17:13: 'case' construct must use '{}'s. [NeedBraces]
Audit done.
```

---

So, we are saying that line 16 (` case Integer _ -> 5;` ) is a false negative? note that in all cases we have to introduce unnecessary `return` or yield

rnveach: To me, that is what it sounds like. If someone doesn't want the violation they could turn on `allowSingleLineStatement`, create a suppression, or remove the token.

I think we should redo the issue first and confirm before we redo the PR.

> value="LAMBDA, LITERAL_CASE"

I believe I would expect a violation on the default tokens if that token was added too.

nrmancuso: I have reviewed and reapproved the issue, thanks a lot for catching this @rnveach

nrmancuso:
I am removing my assignment and dismissing my review, @rnveach please reassign me after you are good so that I can review again.

@mahfouz72 I am sorry that we did not catch this in the issue itself

mahfouz72:
@rnveach please review the new changes based on the new issue description

mahfouz72:
Github, generate report for NeedBraces/all-examples-in-one

rnveach:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/894d35c_2024002603/reports/diff/openjdk21/index.html#A1
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/894d35c_2024002603/reports/diff/openjdk21/xref/home/runner/work/checkstyle/checkstyle/.ci-temp/contribution/checkstyle-tester/repositories/openjdk21/src/java.base/share/classes/java/util/Formatter.java.html#L4749
Example 4 has `allowSingleLineStatement` on. Does the "line" start with the `case`?

Also this regression is incomplete. 4 is the only differences shown because it is the only valid check configuration connected to the issue. Case and default tokens are not on by default, and example 4 is the only one that turns it on and it has `allowSingleLineStatement` on. I would like to see a run with `allowSingleLineStatement` off for this too.

mahfouz72:
github, generate report

mahfouz72:
@rnveach done, and yes the line start with case

rnveach (approved):

nrmancuso (approved):

Issue #15324, opened 2024-07-19
Enforce preceding line break for opening braces of a case/default under switch in google_checks.xml


Zopsss:
I have read check documentation: https://checkstyle.sourceforge.io/checks/blocks/leftcurly.html
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

Detected at: https://github.com/checkstyle/checkstyle/pull/14901#discussion_r1621156853

As per the newly added `Exception` case in the section [4.1.2 Nonempty blocks: K & R style](https://google.github.io/styleguide/javaguide.html#s4.1.2-blocks-k-r-style):

> Exception: In places where these rules allow a single statement ending with a semicolon (;), a block of statements can appear, and the opening brace of this block is preceded by a line break. Blocks like these are typically introduced to limit the scope of local variables, for example inside switch statements.

 
If a block of code appears, then the opening brace ( `{` ) of that block should be preceded by a line break, for example:

```java
case 42:
  {
    int x = foo();
    frob(x);
  }
```

Here a block of code appeared and its `{` is preceded by a line break but Checkstyle does not follow this updated rule, as of now we don't enforce a line break before the `{` for such code blocks, for example:

```java
/** Testing. */
public class TestingLeftCurly {
  static void testing() {
    int xyz = 10;

    switch (xyz) {
      case 42:
        { // false positive warning
          xyz++;
          foo(xyz);
        }
        break;

      case 10: { // false negative, should give warning here
        xyz++;
        foo(xyz);
      }
      break;

      default:
        break;
    }
  }

  static void foo(int xyz) {
    // do nothing.
  }
}
```

**Output:**

```
$ java -jar .\checkstyle-10.17.0-all.jar -c .\google_checks.xml .\TestingLeftCurly.java
Starting audit...
[WARN] C:\checkstyle testing\TestingGoogleStyle\.\TestingLeftCurly.java:8:9: '{' at column 9 should be on the previous line. [LeftCurly]
Audit done.
```

As you can see, there's no violation for `case 10: {` while there should be one. And there is a false positive violation for `case 42: \n{` while it is the correctly formatted code.

---

**Describe what you expect in detail.**

[...]

PR #15335, opened 2024-07-21
Issue #15324: Enforce preceding line break for opening braces of Case and Default blocks in Google checks

kvlohith1:
Issue #15324

### DIFF
--- a/src/main/resources/google_checks.xml
+++ b/src/main/resources/google_checks.xml
@@ -78,12 +78,19 @@
                value="LITERAL_DO, LITERAL_ELSE, LITERAL_FOR, LITERAL_IF, LITERAL_WHILE"/>
     </module>
     <module name="LeftCurly">
+      <property name="id" value="LeftCurlyEol"/>
       <property name="tokens"
                value="ANNOTATION_DEF, CLASS_DEF, CTOR_DEF, ENUM_CONSTANT_DEF, ENUM_DEF,
-                    INTERFACE_DEF, LAMBDA, LITERAL_CASE, LITERAL_CATCH, LITERAL_DEFAULT,
-                    LITERAL_DO, LITERAL_ELSE, LITERAL_FINALLY, LITERAL_FOR, LITERAL_IF,
-                    LITERAL_SWITCH, LITERAL_SYNCHRONIZED, LITERAL_TRY, LITERAL_WHILE, METHOD_DEF,
-                    OBJBLOCK, STATIC_INIT, RECORD_DEF, COMPACT_CTOR_DEF"/>
+                    INTERFACE_DEF, LAMBDA, METHOD_DEF, OBJBLOCK,
+                    STATIC_INIT, RECORD_DEF, COMPACT_CTOR_DEF"/>
+    </module>
+    <module name="LeftCurly">
+      <property name="id" value="LeftCurlyNl"/>
+      <property name="option" value="nl"/>
+      <property name="tokens"
+                value="LITERAL_CASE, LITERAL_DEFAULT, LITERAL_IF, LITERAL_ELSE, LITERAL_SWITCH,
+                       LITERAL_SYNCHRONIZED, LITERAL_FOR, LITERAL_WHILE, LITERAL_DO,
+                       LITERAL_CATCH, LITERAL_TRY, LITERAL_FINALLY" />
     </module>
     <module name="RightCurly">
       <property name="id" value="RightCurlySame"/>
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/main/InputMainComplexityOverflow.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/main/InputMainComplexityOverflow.java
@@ -8,14 +8,22 @@ public class InputMainComplexityOverflow {
 
   public void provokeNpathIntegerOverflow()
     {
-      if (true) {
-        if (true) {
-          if (true) {
-            if (true) {
-              if (true) {
-                if (true) {
-                  if (true) {
-                    if (true) {
+      if (true)
+        {
+        if (true)
+          {
+          if (true)
+            {
+            if (true)
+              {
+              if (true)
+                {
+                if (true)
+                  {
+                  if (true)
+                    {
+                    if (true)
+                      {
                     }
                   }
                 }
@@ -24,14 +32,22 @@ public class InputMainComplexityOverflow {
           }
         }
       }
-      if (true) {
-        if (true) {
-          if (true) {
-            if (true) {
-              if (true) {
-                if (true) {
-                  if (true) {
-                    if (true) {
+      if (true)
+        {
+        if (true)
+          {
+          if (true)
+            {
+            if (true)
+              {
+              if (true)
+                {
+                if (true)
+                  {
+                  if (true)
+                    {
+                    if (true)
+                      {
                     }
                   }
                 }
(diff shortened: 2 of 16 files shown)


mahfouz72:
Thanks for your contribution but this issue is assigned to someone else and it is a part of GSoC project. I think you should take issues that has no assignees only

src/main/resources/google_checks.xml
```diff
@@ -78,12 +78,19 @@
...
+                    INTERFACE_DEF, LAMBDA, METHOD_DEF, OBJBLOCK, LITERAL_SWITCH,
+                    STATIC_INIT, RECORD_DEF, COMPACT_CTOR_DEF, LITERAL_CATCH, LITERAL_TRY,
+                    LITERAL_WHILE, LITERAL_DO, LITERAL_ELSE, LITERAL_FINALLY, LITERAL_FOR,
+                    LITERAL_IF, LITERAL_SYNCHRONIZED"/>
+    </module>
+    <module name="LeftCurly">
+      <property name="id" value="LeftCurlyNl"/>
+      <property name="option" value="nl"/>
+      <property name="tokens"
+                value="LITERAL_CASE, LITERAL_DEFAULT" />
```
Zopsss, line 93: Rule says that a block of statements can appear anywhere in the code and opening curly braces of such blocks are preceded by line break. 

> Exception: In places where these rules allow a single statement ending with a semicolon (;), a block of statements can appear, and the opening brace of this block is preceded by a line break. Blocks like these are typically introduced to limit the scope of local variables, for example inside switch statements.

Here by the `block of statements` means a block of code, something like:

```
{
  int x = 0;
  foo(x++);
}
```

These block of code can appear anywhere, inside an if statement, method, switch statement, etc. Opening braces of such blocks needs to be preceded with a line break.

Here, you have added only LITERAL_CASE & LITERAL_DEFAULT in the tokens, by this way the rule will only be true for switch statements, but we need to follow this rule everywhere else too. 

Your approach is good but the tokens you have used are incorrect. Maybe `INSTANCE_INIT` will do the work.

Zopsss (changes requested):
item:

Zopsss:
@kvlohith1 I appreciate your effort for solving the problem but this issue is part of my GSoC project and it is already assigned to me. 

Please look for other issues to work on and also please comment `I'm on it` or `I'm working on this` to let others know that you have started working on the issue. Go through the [contributing](https://checkstyle.sourceforge.io/contributing.html) page to know all the rules for contribution.

src/main/resources/google_checks.xml
Zopsss: I will solve this issue later, you may close this PR.

romani:
Let's do it together, @kvlohith1 , if finish it quicker it would be awesome.

src/main/resources/google_checks.xml
Zopsss: nvm, you can continue your work as maintainer suggested here: https://github.com/checkstyle/checkstyle/pull/15335#issuecomment-2241740935.

romani: @Zopsss , you are first reviewer of this PR, I will look only after you finish.

Zopsss: @kvlohith1 ping

Pls go through my comment: https://github.com/checkstyle/checkstyle/pull/15335#discussion_r1685715555. This is a difficult task, if you're a beginner I would suggest you to work on [good-first issues](https://github.com/checkstyle/checkstyle/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) to get some experience. But if you still want to work on this issue, we're here to help you. Let us know what you decide :)

kvlohith1: Hi @Zopsss, 
I am not a beginner (albeit a beginner to Checkstyle), I would like to still work on it if thats alright.

Zopsss: alrighty then let's solve this issue! You can go through my this comment: https://github.com/checkstyle/checkstyle/pull/15335#discussion_r1685715555. We need to find out a way to detect such blocks. Play with our CLI and generate ASTs to find a way to detect such blocks, it would be good if we're able to achieve this by modifying our google config only. If we cannot achieve this only by modifying our config then we would need to make modifications in check's implementation. If you face some problems or have any doubts, don't hesitate to ask us :)

kvlohith1: Apologies, I have been working on this, feels a little too complicated.

Zopsss: Did you tried generating ASTs and found any patterns? If yes then please share

src/main/resources/google_checks.xml
```diff
@@ -81,16 +81,16 @@
...
     </module>
     <module name="LeftCurly">
       <property name="id" value="LeftCurlyNl"/>
       <property name="option" value="nl"/>
       <property name="tokens"
-                value="LITERAL_CASE, LITERAL_DEFAULT" />
+                value="LITERAL_CASE, LITERAL_DEFAULT, LITERAL_IF, LITERAL_ELSE, LITERAL_SWITCH,
+                       LITERAL_SYNCHRONIZED, LITERAL_FOR, LITERAL_WHILE, LITERAL_DO,
+                       LITERAL_CATCH, LITERAL_TRY, LITERAL_FINALLY" />
     </module>
```
Zopsss, line 94: Umm, this is not we want. I guess you're getting wrong idea about the feature we want. 

Go through this stackoverflow post: https://stackoverflow.com/questions/1563030/anonymous-code-blocks-in-java

We want such anonymous code block's opening curly brace ( `{` ) to be on new line. Example from the post:

```java
    // in
    {
        // out
    }
}
```

Here, we have anonymous inner block inside the main method. We want to ensure that such block's opening brace is preceded with a line break.

```java
    // in
    final String str = "s"; { // warn, incorrect placement of "{"
        // out
    }
    final int xyz = 19;
    { // ok, correct placement. "{" is preceded with a line break
       // code here
    }
}
```

In the above example, for `String str` line, we should give warning like `"{" should be alone on the line`.

Here's a small explanation about the from google side: https://github.com/google/styleguide/commit/49488412b2f59843fce2433bd834a3fd9700604e#r143141977, we have some examples in the issue's description too: https://github.com/checkstyle/checkstyle/issues/15324.

With the current changes you made, we're enforcing the opening curly braces of various blocks like if/else, try/catch/finally to be on new line but the rule isn't about that.

src/main/resources/google_checks.xml
```diff
@@ -81,16 +81,16 @@
       <property name="id" value="LeftCurlyEol"/>
       <property name="tokens"
                value="ANNOTATION_DEF, CLASS_DEF, CTOR_DEF, ENUM_CONSTANT_DEF, ENUM_DEF,
-                    INTERFACE_DEF, LAMBDA, METHOD_DEF, OBJBLOCK, LITERAL_SWITCH,
-                    STATIC_INIT, RECORD_DEF, COMPACT_CTOR_DEF, LITERAL_CATCH, LITERAL_TRY,
-                    LITERAL_WHILE, LITERAL_DO, LITERAL_ELSE, LITERAL_FINALLY, LITERAL_FOR,
-                    LITERAL_IF, LITERAL_SYNCHRONIZED"/>
+                    INTERFACE_DEF, LAMBDA, METHOD_DEF, OBJBLOCK,
+                    STATIC_INIT, RECORD_DEF, COMPACT_CTOR_DEF"/>
```
Zopsss, line 85: pls revert this. This was correct, no need to modify this.

src/test/resources/com/puppycrawl/tools/checkstyle/main/InputMainComplexityOverflow.java
```diff
@@ -8,14 +8,22 @@ public class InputMainComplexityOverflow {
...
+            if (true)
+              {
+              if (true)
+                {
+                if (true)
+                  {
+                  if (true)
+                    {
+                    if (true)
+                      {
```
Zopsss, line 26: This is incorrect placement of "{". I have explained the issue in detail in the above comment

Zopsss (changes requested):
explained the issue in more detail:

Zopsss:
Also squash the commits into one. We should have only one commit per PR

romani:
This looks like abandoned

PR #15337, opened 2024-07-21
Issue #14975: New Check SealedShouldHavePermitsList

### LINKED ISSUE #14975: Add Check Support for Java 17 Sealed Classes: New Check SealedClassesShouldPermitsSubClasses
child of #14969

> the sealed class may omit the permits clause and the Java compiler will infer the permitted subclasses from the declarations in the source file. (The subclasses may be auxiliary or nested classes.) 

A sealed class restricts which other classes can extend it. This is typically achieved using the permits clause.
However, if a sealed class does not specify any permitted subclasses using the permits clause, it implicitly allows any class within the same file to extend it. This can lead to unintended inheritance classes that should not have been allowed to extend the sealed class can do so at any time. 

I propose the implementation of a check that violates any sealed class or interface without a `permits` clause. This check will help ensure that we explicitly defining permitted subclasses, therefore avoiding unintended extensions.

--- 
**Config**

```XML
$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
    <module name="TreeWalker">
        <module name="PatternVariableAssignment"/>
    </module>
</module>
```

---
**Example**

```Java
public class SealedClasses {
     sealed class A {} // violation
     final class B extends A {}
     final class C extends A {}
     final class D { }    // this is not intended to be a subclass of A but it can do so anytime as there is no restrictive permits clause
 }
```
---
**Expected Output**

```
$ java -jar checkstyle-10.19.0-all.jar -c config.xml SealedClasses.java 
Starting audit...
[ERROR] D:\SealedClasses.java:2:5: subclasses are missing from permits clause of a sealed class
Audit done.
```

mahfouz72:
closes #14975:

New module config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/bae6f13dbf50787449429a666ce99615294a0d80/check.xml


mahfouz72:
github, generate report

mahfouz72:
Github, generate site

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/sealedshouldhavepermitslist/InputSealedShouldHavePermitsListInnerClass.java
```diff
@@ -0,0 +1,21 @@
...
+package com.puppycrawl.tools.checkstyle.checks.design.sealedshouldhavepermitslist;
+
+public class InputSealedShouldHavePermitsListInnerClass {
+     sealed class A {} // violation
+     final class B extends A {}
+     final class C extends A {}
+     final class D { } // this can extend A, so as any other class in the compilation unit
+}
+
+class corrected {
```
nrmancuso, line 16: Please always follow common Java paradigms to allow us to focus on what is important here, please do for all:

```suggestion
class InputSealedShouldHavePermitsListInnerClassCorrect {
```

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/sealedshouldhavepermitslist/InputSealedShouldHavePermitsListInnerClass.java
```diff
@@ -0,0 +1,21 @@
...
+*/
+
+//non-compiled with javac: Compilable with Java17
+package com.puppycrawl.tools.checkstyle.checks.design.sealedshouldhavepermitslist;
+
+public class InputSealedShouldHavePermitsListInnerClass {
+     sealed class A {} // violation
+     final class B extends A {}
+     final class C extends A {}
+     final class D { } // this can extend A, so as any other class in the compilation unit
```
nrmancuso, line 13: Let's spice up these inputs a bit with non-sealed subclasses, static classes, etc.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/sealedshouldhavepermitslist/InputSealedShouldHavePermitsListTopLevelSealedInterface.java
```diff
@@ -0,0 +1,19 @@
+/*
+SealedShouldHavePermitsList
+
+*/
+
+//non-compiled with javac: Compilable with Java17
+package com.puppycrawl.tools.checkstyle.checks.design.sealedshouldhavepermitslist;
+
+public sealed interface InputSealedShouldHavePermitsListTopLevelSealedInterface { // violation
```
nrmancuso, line 9: Similar to above, let's spice these up with static classes, records, enums, etc

nrmancuso (changes requested):
Pretty much a slam dunk, few minor items:

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/sealedshouldhavepermitslist/InputSealedShouldHavePermitsListTopLevelSealedInterface.java
mahfouz72: I think it is done. Please review. Sorry for the late reply it was done  yesterday but I forgot to reply :)

nrmancuso (approved):
Good stuff.

romani:
CI restarted

mahfouz72:
Rebased.

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/sealedshouldhavepermitslist/Example1.java
```diff
@@ -0,0 +1,27 @@
...
+// xdoc section -- start
+public class Example1 {
+  sealed class A {}   // violation
+  final class B extends A {}
+  final class C extends A {}
+  final class D { } // this can extend A, so as any other class in the same file
+}
+
+class CorrectedExample1 {
+  sealed class A permits B, C  {} // ok, explicitly declared permitted subclasses
```
rnveach, line 22: Double space after C.

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/sealedshouldhavepermitslist/Example1.java
```diff
@@ -0,0 +1,27 @@
...
+</module>
+*/
+
+
+//non-compiled with javac: Compilable with Java17
+package com.puppycrawl.tools.checkstyle.checks.design.sealedshouldhavepermitslist;
+
+// xdoc section -- start
+public class Example1 {
+  sealed class A {}   // violation
```
rnveach, line 15: Triple space here.

rnveach (changes requested):
Simple changes.

Also, please ensure the branch is rebased on the latest master.

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/sealedshouldhavepermitslist/Example1.java
mahfouz72: done

rnveach (approved):

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/sealedshouldhavepermitslist/Example1.java
```diff
@@ -0,0 +1,27 @@
...
+  final class C extends A {}
+  final class D { } // this can extend A, so as any other class in the same file
+}
+
+class CorrectedExample1 {
+  sealed class A permits B, C {} // ok, explicitly declared permitted subclasses
+  final class B extends A {}
+  final class C extends A {}
+  final class D { } // this can not extend A
+}
```
romani, line 26: please add example:
```
public class Example2 {
  sealed class A {} // violation
  // bunch of code
  final class D { } // this can extend A, so as any other class in the same file
  // bunch of code
}
```
and this becomes very intereting .... we will annoy users to have any sealed class. 
and force  users to have 
```
public class Example1 {
  sealed class A {} permits SealingA
  final class SealingA extends A {}
  final class D { } // this can extend A, so as any other class in the same file
}
```

where users need to make fake SealingA as just double confirmation that nobody else will extend.
this is not ok.

What you tried to do is catch only cases like:
```
public class Example1 {
  sealed class A {}  
  final class LazyExtensionA extends A {} // violation , 'allowance by permit clause at A should be explicit'
}
```
and
```
public class Example2 {
  // no violation, as nobody even try to use hack:
  //        ` it implicitly allows any class within the same compilation unit to extend it.` 
  sealed class A {} 
  // bunch of code
  final class D { } // this can extend A, and until it is not extending, there is no problem
  // bunch of code
}
```

https://docs.oracle.com/javase/specs/jls/se21/html/jls-8.html#jls-8.1.6

>If C is not an enum class, then its permitted direct subclasses are those classes declared in the same compilation unit as C ([§7.3](https://docs.oracle.com/javase/specs/jls/se21/html/jls-7.html#jls-7.3)) which have a canonical name ([§6.7](https://docs.oracle.com/javase/specs/jls/se21/html/jls-6.html#jls-6.7)) and whose direct superclass is C.
That is, the permitted direct subclasses are inferred as the classes in the same compilation unit that specify C as their direct superclass. The requirement for a canonical name means that no local classes or anonymous classes will be considered.


So we are fighting with implicit nuances of jdk, and by means of Check we want to do all explicit.
But Explicit declaration of permits is required when we detecting extending, so code owner just got lazy to explicitly declare permits.

romani (changes requested):
item:

src/xdocs-examples/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/sealedshouldhavepermitslist/Example1.java
mahfouz72: I got confused a little because we have a non-compile able example here

> please add example:
> 
> public class Example2 {
>   sealed class A {} // violation
>   // bunch of code
>   final class D { } // this can extend A, so as any other class in the same file
>   // bunch of code
> }
> 

---

```
D:\CS\test cat src/Test.java
public class Test {
    sealed class A { }
    final class D { }
}
D:\CS\test javac src/Test.java
src\Test.java:2: error: sealed class must have subclasses
    sealed class A { }
           ^
1 error
```
To make the example compilable it should be:

```
D:\CS\test cat src/Test.java
public class Test {
    sealed class A { }   // violation by us, it should be ( sealed class A permits D { } ) 
    final class D extends A { }
}
```
btw, the above example is a violation by IDEA inspections.

Does this satisfy your thoughts? If not then please share a compilable example with what you expect vs what we actually do

nrmancuso: @romani please prove that your concerns are valid by using `javac` on your examples above.

mahfouz72: I think his thought are around idea that we may have a sealed class that have no subclasses in this case to satisfy the check we will make a fake class and add it to the permitted list of the sealed class 
but this will not compile sealed class mush have subclasses
@romani please confirm if I understand you correctly

romani: Yes. Current logic kind forbid usage of single sealed class. It demands subclass.

mahfouz72: Then, please finish the review and merge :)

mahfouz72: @romani This is also non-compilable
```
public class Example2 {
  sealed class A {} // violation
  // bunch of code
  
}
```
---

```
D:\CS\test cat src/Test.java
public class Test {
    sealed class A {}
}
D:\CS\test javac src/Test.java
src\Test.java:2: error: sealed class must have subclasses
    sealed class A {}
           ^
1 error
```

> We need tests on sealed classes at top level of same file (compilation unit).
https://github.com/checkstyle/checkstyle/blob/108ec4cd543c93d44f2edec9fbea13070337441d/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/design/sealedshouldhavepermitslist/InputSealedShouldHavePermitsListTopLevelSealedClass.java

romani: very smart compiler, awesome.

romani:
example is taken from https://openjdk.org/jeps/409
and we violating it, very bold position.
```
$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <module name="TreeWalker">
    <module name="SealedShouldHavePermitsList"/>
  </module>
</module>

$ cat Test.java 
abstract sealed class Test {
    final class A extends Test {  }
    final class B extends Test {  }
    final class C extends Test {  }
}

$ javac Test.java 
$ java -jar /home/rivanov/java/github/romani/checkstyle/target/checkstyle-10.18.0-SNAPSHOT-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:1:1: Sealed classes or interfaces should explicitly declare permitted subclasses. [SealedShouldHavePermitsList]
Audit done.
```

Do you think we need to exclude this pattern from validation ?
Checksyle will demand to put canonical names, it code will be this ugly:


--------

Example from https://docs.oracle.com/en/java/javase/17/language/sealed-classes-and-interfaces.html
```
$ javac Figure.java
$ cat Figure.java 
public sealed class Figure
    // The permits clause has been omitted
    // as its permitted classes have been
    // defined in the same file.
{ }

final class Circle extends Figure {
    float radius;
}
non-sealed class Square extends Figure {
    float side;
}
sealed class Rectangle extends Figure {
    float length, width;
}
final class FilledRectangle extends Rectangle {
    int red, green, blue;
}

$ cat -n Figure.java 
     1	public sealed class Figure
     2	    // The permits clause has been omitted
     3	    // as its permitted classes have been
     4	    // defined in the same file.
     5	{ }
     6	
     7	final class Circle extends Figure {
     8	    float radius;
     9	}
    10	non-sealed class Square extends Figure {
    11	    float side;
    12	}
    13	sealed class Rectangle extends Figure {
    14	    float length, width;
    15	}
    16	final class FilledRectangle extends Rectangle {
    17	    int red, green, blue;
    18	}

$ java -jar /home/rivanov/java/github/romani/checkstyle/target/checkstyle-10.18.0-SNAPSHOT-all.jar -c config.xml Figure.java
Starting audit...
[ERROR] /var/tmp/Figure.java:1:1: Sealed classes or interfaces should explicitly declare permitted subclasses. [SealedShouldHavePermitsList]
[ERROR] /var/tmp/Figure.java:13:1: Sealed classes or interfaces should explicitly declare permitted subclasses. [SealedShouldHavePermitsList]
Audit done.
```

ok, both violations are ok, please add such code to Inputs as it (only name of top level class should be different).

romani:
@nrmancuso ,
what about `Rationale: If a sealed class or interface omits the permits clause, it implicitly allows any class in the same file(compilation unit) to extend it, potentially leading to unwanted inheritance if compilation unit is big (several scrolls of screen). By including the permits clause you explicitly list permitted subclasses(that might not being seen on screen with sealed), enhancing control and reducing the risk of unwanted extensions.` ?

wording can be changed a bit to be better English.

nrmancuso:
>Do you think we need to exclude this pattern from validation ?

Placing violations on this pattern is the point of this check :) This check is for demanding that we declare all permitted subclasses. If a user can dream up some scenario where this doesn't work for them, we can create an issue and either add some property or help them to formulate an XPath suppression.

>ok, both violations are ok, please add such code to Inputs as it (only name of top level class should be different).

I have added this example to the tests.

>@nrmancuso , what about

I think this is wordy and wholly unnecessary. Consider how the majority of users will come to use any check (not just this one): they experience some pain, and come looking for some way to prevent this pain via Checkstyle. It could be that some junior engineer doesn't understand the implications/semantics of sealed classes, or maybe they are just crazy about EVERYTHING being explicit and use checks like https://checkstyle.org/checks/modifier/classmemberimpliedmodifier.html#ClassMemberImpliedModifier, or maybe it is the scenario that you came up with where they are too lazy to scroll down :). 

Good technical writing is concise and direct, which the current description is.

Additionally:

>compilation unit is big (several scrolls of screen)

In production code, classes that are NOT a couple hundred lines of code long are the exception :)

romani:
>they experience some pain, and come looking for some way to prevent this pain via Checkstyle.

This is a case for very advanced and caring engineers. 1 out 100.

I am about people that are 99 out of 100: they experience some pain due to Check violation, and come looking at documentations for a reason why it matters.

When is read this Check I was slightly confused why this matters in certain cases. All examples around are small and compact, it was not obvious "why" it matters.

There are lot of examples in our doc where we collect reasons "why" it matters, as nobody can remember all reasons, and when required to show all to persuade others, something might be forgotten and sound not very persuasive.

I want to preserve this reason. There might be more later on to add, it is ok.

Please help me to correct wording, or if you don't want to participate in this, we will go with my variant and in future somebody will improve it.

romani:
@mahfouz72 , please do me favor to do

>We can add Input that have long/big classes (we can suppress violation that Input cannot be longer than 120 lines), so it will be very clear that it becomes very matter on big classes.

And put a comment over class `all child classes are not visible on screen, placing all of them in permits clause help reader to avoid scrolling and investigating`

romani:
@nrmancuso , one more version:
`Improved text:
When a permits clause is omitted from a sealed class, any class within the same compilation unit can extend it by default. This differs from other sealed classes where permitted subclasses are explicitly declared, making them readily visible to the reader. Without a permits clause, identifying potential subclasses requires searching the entire compilation unit, which can be challenging, especially in large files with complex class hierarchies.`

nrmancuso:
> Rationale: When a permits clause is omitted from a sealed class, any class within the same compilation unit can extend it. This differs from other sealed classes where permitted subclasses are explicitly declared, making them readily visible to the reader. Without a permits clause, identifying potential subclasses requires searching the entire compilation unit, which can be challenging, especially in large files with complex class hierarchies.

I made this concept more clear in the example, no users are going through unit tests looking for check concepts.

>@nrmancuso , one more version:

updated to use this

nrmancuso:
Github, generate website

romani:
>  no users are going through unit tests looking for check concepts.

we can simply make a link from description to github to Input. We should keep this in mind. If something is hard to explain and better to show - we can place a link to Input.

romani (approved):
ok to merge

PR #15341, opened 2024-07-22
Issue #15045: Add LITERAL_WHEN as decision point in CyclomaticComplex…

### LINKED ISSUE #15045: Add Check Support for Java 21 Pattern Matching for Switch Syntax:  CyclomaticComplexity

I have read check documentation:https://checkstyle.org/checks/metrics/cyclomaticcomplexity.html

---

```
➜  src cat config.xml                               
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="CyclomaticComplexity">
            <property name="max" value="3"/>
        </module>
    </module>
</module>
➜  src cat Test.java                                
import java.awt.color.ICC_ColorSpace;

record ColoredPoint(int p, int x, boolean c) {
}
record Rectangle(ColoredPoint upperLeft, ColoredPoint lowerRight) {
}

public class Test {

    // violation,  Cyclomatic Complexity is 4
    void test(Object obj) {  // 1, function declaration
        switch (obj) {
            case ColoredPoint(int a, int b, _)  // 2 , case
                    when (a >= b) -> {        // should we add 1 for when ? it was literally a `&&` before
            }
            case ColoredPoint(int a, int b, _) -> {}  // 3, case
            case Rectangle(_, _) -> {}  // 4, case
        }
    }

    // violation,  Cyclomatic Complexity is 5
    void test2(Object obj) {  // 1, function declaration
        // I interpreted `when` as `&&`
        if (obj instanceof ColoredPoint(int a, int b, _) && a >= b) { // 2, if and 3, '&&'

        } else if (obj instanceof ColoredPoint(int a, int b, _)) { // 4 if

        } else if (obj instanceof Rectangle(_, _)) {   // 5 if

        }
    }
}
Starting audit...
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:11:5: Cyclomatic Complexity is 4 (max allowed is 3). [CyclomaticComplexity]
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:23:5: Cyclomatic Complexity is 5 (max allowed is 3). [CyclomaticComplexity]
Audit done.

```

---

Since we add 1 for `&&`. I think it makes sense that `when` should be considered as +1 here. It is just the same as `&&` in if conditions see the similarities between the two examples

mahfouz72:
closes #15045 

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/bdde1a4a901d1eca9e79c017dc8037fd989a020c/check.xml
Diff Regression patch config: https://gist.githubusercontent.com/mahfouz72/33967bf86e4a29cafa429771b0e95808/raw/6ec96ab6424c29e5c6ad0fb9d403cdda19054609/check_patch.xml

### DIFF
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/cyclomaticcomplexity/InputCyclomaticComplexityWhenSwitchAsSinglePoint.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/cyclomaticcomplexity/InputCyclomaticComplexityWhenSwitchAsSinglePoint.java
@@ -1,0 +1,59 @@
+/*
+CyclomaticComplexity
+max = 0
+switchBlockAsSingleDecisionPoint = true
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.metrics.cyclomaticcomplexity;
+
+public class InputCyclomaticComplexityWhenSwitchAsSinglePoint {
+
+    // violation below, 'Cyclomatic Complexity is 5 (max allowed is 0)'
+    void testIf(Object o) {   // 1, function declaration
+        if (o instanceof Integer i && i == 9) {}  // 2, if - 3, &&
+        else if (o instanceof String s && s.length() == 9) {} // 4, if - 5, &&
+    }
+
+    // violation below, 'Cyclomatic Complexity is 4 (max allowed is 0)'
+    void testSwitch(Object o) {   // 1, function declaration
+        switch (o) {  // 2, switch
+            case Integer i when i == 9 -> {} // 3, when
+            case String s when s.length() == 9 -> {} // 4, when
+            default -> {}
+        }
+    }
+
+    // violation below, 'Cyclomatic Complexity is 5 (max allowed is 0)'
+    void testSwitch2(Object o) {  // 1, function declaration
+        switch (o) {    // 2 switch
+            case Integer i -> {}
+            case Point(int x, int y) when (x == 0 && y == 0) -> {} // 3, when - 4, &&
+            case Point(int x, int y) when x == 0 -> {} // 5, when
+            default -> {}
+        }
+    }
+
+    // violation below, 'Cyclomatic Complexity is 5 (max allowed is 0)'
+    void testSwitch3(Object o) {  // 1, function declaration
+        switch (o) {    // 2 switch
+            case Integer i -> {}
+            case Point(int x, int y) -> {
+                if (x == 0 && y == 0) { // 3, if - 4, &&
+
+                }
+            }
+            case Point2(int x) -> {
+                if (x == 0) { // 5 if
+
+                }
+            }
+            default -> {}
+        }
+    }
+
+    record Point(int x, int y) { }
+
+    record Point2(int x) { }
+}
--- a/src/xdocs/checks/metrics/cyclomaticcomplexity.xml
+++ b/src/xdocs/checks/metrics/cyclomaticcomplexity.xml
@@ -19,13 +19,21 @@
           static initializers and instance initializers</a>.
         </p>
         <p>
-          The complexity is equal to the number of decision points <code> + 1</code>.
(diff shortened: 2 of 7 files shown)


src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/cyclomaticcomplexity/InputCyclomaticComplexityWhenSwitchAsSinglePoint.java
```diff
@@ -0,0 +1,39 @@
...
+
+   // violation below, 'Cyclomatic Complexity is 5 (max allowed is 0)'
+   void testSwitch2(Object o) {  // 1, function declaration
+     switch (o) {    // 2 switch
+         case Integer i -> {}
+         case Point(int x, int y) when (x == 0 && y == 0) -> {} // 3, when - 4, &&
+         case Point(int x, int y) when x == 0 -> {} // 5, when
+         default ->  {}
+      }
+   }
```
mahfouz72, line 36: I am not sure about it, but shouldn't it be more reasonable if we ignore everything inside the switch when `switchBlockAsSingleDecisionPoint` is true? because basically, we should treat switch as a single decision point so considering `when` and `&&` as decision points and adding 1 to them is odd to me when this option is true

mahfouz72:
github, generate site

mahfouz72:
github, generate report

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/cyclomaticcomplexity/InputCyclomaticComplexityWhenSwitchAsSinglePoint.java
nrmancuso: Please add traditional switches in here with nested `if` statements to allow us to compare how this check handles them

mahfouz72: done, it got counted. 

https://github.com/checkstyle/checkstyle/blob/87d431a7a0f438e5a858cf11ff3acf025148f936/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/CyclomaticComplexityCheck.java#L256-L259

looks like we only ignore `case` when this option is true but anything else inside the switch will be counted

nrmancuso (approved):

src/xdocs/checks/metrics/cyclomaticcomplexity.xml
```diff
@@ -22,7 +22,7 @@
           The complexity is equal to the number of decision points <code> + 1</code>.
           Decision points: <code>if</code>, <code>while</code>
           , <code>do</code>, <code>for</code>, <code>?:</code>, <code>catch</code>
-          , <code>switch</code>, <code>case</code>
+          , <code>switch</code>, <code>case</code>, <code>when</code>
           statements and operators <code>&amp;&amp;</code> and <code>||</code>
```
rnveach, line 26: > when statements

https://docs.oracle.com/javase/specs/jls/se22/html/jls-14.html#jls-Guard
> may have an optional when expression, known as a guard

This is not a statement.

rnveach (changes requested):

mahfouz72:
github, generate site

src/xdocs/checks/metrics/cyclomaticcomplexity.xml
mahfouz72: done, how do you feel about this being a list? I think this is more readable

nrmancuso: List would definitely be better here.

mahfouz72: ok done it is list now

rnveach (approved):

PR #15345, opened 2024-07-23
Issue #15038: Add new input file for pattern matching in switch synta…

### LINKED ISSUE #15038: Add Check Support for Java 21 Pattern Matching for Switch Syntax: FallThrough

I have read check documentation:https://checkstyle.org/checks/coding/fallthrough.html#FallThrough

```
PS D:\CS\test> cat config.xml                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="FallThrough">

        </module>
    </module>
</module>
public class Test {

    void test(Object obj) {
        switch (obj) {
            case ColoredPoint(int x,_,_) when (x >= 2):
                System.out.println("coloredPoint");
            case Rectangle(_,_) : // violation
                System.out.println("Rectangle");
                break;
            default:
                throw new IllegalStateException("Unexpected value: " + obj);
        }
         switch (obj) {
            case ColoredPoint(int x,_,_) when (x >= 2):
                System.out.println("coloredPoint");
                break;
            case Rectangle(var y,_) :
                System.out.println("Rectangle");
                break;
            default:
                throw new IllegalStateException("Unexpected value: " + obj);
        }
        switch (obj) {
            case ColoredPoint(int x,_,_) when (x >= 2) ->
                System.out.println("coloredPoint");
            case Rectangle(var y,_) ->
                System.out.println("Rectangle");
            default->
                throw new IllegalStateException("Unexpected value: " + obj);
        }
    }
}
Starting audit...
[ERROR] D:\CS\test\src\Test.java:9:13: Fall through from previous branch of the switch statement. [FallThrough]
Audit done.
```

we are ok here. we can close or approve for input file. 
note that:
- the switch rule (arrow syntax) isn't subject to fall through by design
- fall through to a case label declaring a pattern will not compile 
     Example:
 ```
       switch (obj) {
            case ColoredPoint(int x,_,_) when (x >= 2):
                System.out.println("coloredPoint");
            case Rectangle(var y,_) :       // compile-time error Illegal fall-through to a pattern
                System.out.println("Rectangle");
                break;
            default:
                throw new IllegalStateException("Unexpected value: " + obj);
        }
```

### PR DESCRIPTION
**Pull request by mahfouz72**
closes #15038

### DIFF
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughWithPatternMatchingCheckLastCase.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughWithPatternMatchingCheckLastCase.java
@@ -1,0 +1,99 @@
+/*
+FallThrough
+checkLastCaseGroup = true
+reliefPattern = (default)falls?[ -]?thr(u|ough)
+
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.coding.fallthrough;
+
+public class InputFallThroughWithPatternMatchingCheckLastCase {
+
+    void testSwitchStatements(Object o) {
+        switch (o) {
+            case Integer i when i > 0:
+                System.out.println("Positive integer");
+            // violation below, 'Fall\ through from previous branch of the switch'
+            case String _ when o.toString().isEmpty():
+                System.out.println("empty string");
+            default:
+            // 2 violations above:
+            //                     'Fall\ through from previous branch of the switch'
+            //                     'Fall\ through from the last branch of the switch'
+                System.out.println("default");
+        }
+
+        switch (o) {
+            case Integer i when i > 0: {
+                System.out.println("Positive integer");
+            }
+            // violation below, 'Fall\ through from previous branch of the switch'
+            case String _ when o.toString().isEmpty(): {
+                System.out.println("empty string");
+            }
+            default: {
+            // 2 violations above:
+            //                     'Fall\ through from previous branch of the switch'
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/fallthrough/InputFallThroughWithPatternMatchingCheckLastCase.java [hunk 1]
```diff
@@ -0,0 +1,99 @@
...
+            case Integer i when i > 0:
+                System.out.println("Positive integer");
+                break;
+            case String s when o.toString().isEmpty():
+                System.out.println("empty string");
+                break;
+            default:   // violation 'Fall\ through from the last branch of the switch'
+                System.out.println("default");
+        }
+    }
```
mahfouz72, line 64: This is a bug. Why fall thru ? we have a break in the previous branch

```
D:\CS\test cat src/Test.java
public class Test {
    void test(int x) {
        switch (x) {
            case 1 : System.out.println("1"); break;
            case 2 :  System.out.println("2"); break;
            default : System.out.println("default");  // violation
        }

    }
}
D:\CS\test cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
  <module name="TreeWalker">
    <module name="FallThrough">
      <property name="checkLastCaseGroup" value="true"/>
    </module>
  </module>
</module>
D:\CS\test java  -jar checkstyle-10.17.0-all.jar -c config.xml src/Test.java
Starting audit...
[ERROR] D:\CS\test\src\Test.java:6:13: Fall through from the last branch of the switch statement. [FallThrough]
Audit done.
```

nrmancuso (approved):

Issue #15025, opened 2024-06-18
Add Check Support for Java 21 Pattern Matching for Switch Syntax: WhitespaceAfter


mahfouz72:
child of #14961 

I have read check documentation: https://checkstyle.org/checks/whitespace/whitespaceafter.html#WhitespaceAfter
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```

➜  src cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
<module name="Checker">
    <property name="charset" value="UTF-8"/>
    <module name="TreeWalker">
        <module name="WhitespaceAfter"/>
    </module>
</module>
➜  src cat Test.java 
record ColoredPoint(int p, int x, String c) { }

public class Test {

    void test(Object obj) {
        int y = 0;
        if(y>=2) { }            // violation
        switch (obj) {
            case ColoredPoint(int x, _, _) when(x >= 2) -> {}     // expected violation

            default -> {}
        }
    }
}
➜  src javac --enable-preview --release 21 Test.java                        
Note: Test.java uses preview features of Java SE 21.
Note: Recompile with -Xlint:preview for details.
➜  src java -jar checkstyle-10.18.1-SNAPSHOT-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:7:9: 'if' is not followed by whitespace. [WhitespaceAfter]
Audit done.
Checkstyle ends with 1 errors.


```
---

**Describe what you want in detail**

violation on `when` because there is no white space after when and before the parens

rnveach:
Token to add is `LITERAL_WHEN` and should be able to turn on/off.

I am good with this.

nrmancuso:
I have updated the example now that we have added the `EXPR` token, this is a false negative.

rnveach:
Issue is not done until subsequent PRs are merged for the IT

PR #15359, opened 2024-07-24
supplemental: Extend ITs input section 4.6.2 to have examples with LI…

mahfouz72:
closes: #15027 
closes: #15025

### DIFF
--- a/src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting/rule462horizontalwhitespace/InputWhitespaceAroundWhen.java
+++ b/src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting/rule462horizontalwhitespace/InputWhitespaceAroundWhen.java
@@ -1,0 +1,78 @@
+//non-compiled with javac: Compilable with Java21
+
+package com.google.checkstyle.test.chapter4formatting.rule462horizontalwhitespace;
+
+class InputWhitespaceAroundWhen {
+
+  /** method. */
+  void test(Object o) {
+    switch (o) {
+      case Integer i when(i == 0) -> {
+      }
+      // 2 violations 2 lines above:
+      //              ''when' is not followed by whitespace.'
+      //              ''when' is not followed by whitespace.'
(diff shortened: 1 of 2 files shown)


mahfouz72:
@romani @rdiachenko @Zopsss  Please review, the test will not pass until we have the token in our config right?

src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting/rule462horizontalwhitespace/InputWhitespaceAroundWhen.java
```diff
@@ -0,0 +1,46 @@
...
+            //              ''when' is not preceded with whitespace.'
+            default -> { }
+        }
+
+        switch (o) {
+            case Integer i when (i == 0) -> { }
+            case String s when (s.equals("a")) -> { }
+            case Point(int x, int y) when (x >= 0 && y >= 0) -> { }
+            default -> { }
+        }
```
Zopsss, line 42: can you also add test cases where `when` is not followed by brackets? something like:

```java
String day = "Monday";

switch (day) {
    case String s when s.startsWith("M"):
        System.out.println("It's a Monday!");
        break;
    case String s when s.startsWith("T"):
        System.out.println("It's a Tuesday!");
        break;
    default:
        System.out.println("It's another day.");
}
```

Zopsss (commented):
item:

nrmancuso:
I am putting this in draft until we resolve the linked issues

Zopsss:
>  Please review, the test will not pass until we have the token in google config right?

Yes but before that we need the token support in Check itself. After that we can add it in google config

romani:
Please format this file by CLI of google-java-format and restore only required for this Input missing space to keep violations valid.
So in this case file will look natural for good style except for required lines that are with violation by design.

nrmancuso (approved):

romani (approved):
Look ok

rnveach (approved):

PR #15372, opened 2024-07-26
Issue #15214: migrated section 5.2.2 class names to whole config testing

Zopsss:
#15214


romani (approved):
ok to merge

this input looks like was missed in formatting iteration, so there extra changes, but it is ok.
in this final phase we forced to use all formatting from style guide

Zopsss:
> this input looks like was missed in formatting iteration

naah, I just moved all the extra classes/interfaces inside main class and added indentation to prevent `Top class should be in its own file` violation.

rdiachenko (approved):
lgtm

reviewed with `hide whitespaces` split diff

Issue #15159, opened 2024-06-28
RedundantModifer Should Violate Final Modifer on Unnamed Variables if jdkVersion GTE 22


mahfouz72:
child of #14942 

Check documentation: https://checkstyle.org/checks/modifier/redundantmodifier.html#RedundantModifier

---

```
PS D:\CS\test> cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
  <module name="TreeWalker">
    <module name="RedundantModifier"/>
    <property name="jdkVersion" value="21"/>
  </module>
</module>
PS D:\CS\test> cat src/Test.java
public class Test {
    void m(Object o) throws Exception {
        try (final var a = lock()) {  // violation , resource is effective final
        }

        if (o instanceof R(final int _)) { // expected violation, unnamed pattern variable is effective final

        }

        final int _ = sideEffect(); // expected violation, unnamed variable is effective final
    }
    AutoCloseable lock() { return null; }
    record R(int x) { }
    int sideEffect() { return 0; }
}
PS D:\CS\test> java  -jar checkstyle-10.17.0-all.jar -c config.xml  src/Test.java                                           
Starting audit...
[ERROR] D:\CS\test\src\Test.java:3:14: Redundant 'final' modifier. [RedundantModifier]
Audit done.
Checkstyle ends with 1 errors.
PS D:\CS\test> 
```

---

[JLS](https://docs.oracle.com/javase/specs/jls/se22/html/jls-6.html#jls-6.3) :

> If a declaration does not include an identifier, but instead includes the keyword _ (underscore), then the entity cannot be referred to by name. The following kinds of entity may be declared using an underscore:

Like resources, unnamed variables are effectively final which means that whether the unnamed variable has final modifier or not. the variable can't be reassigned. I expect a violation on the redundant modifer check on unnamed variables.

---
**Describe what you want in detail.**

We should update `RedundantModifer`  to violate the `final` modifier on unnamed variables if `jdkVersion` is >= 22. If a user is on old Java versions (below Java 9). the final modifier will not be redundant because `_` is a normal identifier in this case. 

Example:

```
PS D:\CS\test> cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
[...]

PR #15376, opened 2024-07-26
Issue #10798: New property to violate strictfp as redundant modifier …

### LINKED ISSUE #10798: RedundantModifierCheck: New property 'jdkVersion' to violate strictfp as redundant modifier if version LTE 17
http://checkstyle.sourceforge.net/config_modifier.html#RedundantModifier
Reason: since java 17 there is no need for this modifier, see details in [JEP 306](https://openjdk.java.net/jeps/306)

Suggestion is to add new property `jdkVersion` where user can define jdk version to let us make decision on dedundancy of `strictfp`. Before 17th not redundant, on 17 and afrter - redundant. No jdk specified, no violation on  `strictfp`.

```bash
/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">

    <module name="TreeWalker">
        <module name="RedundantModifier">
         <property name="jdkVersion" value="17">
        </module>
    </module>
</module>

/var/tmp $ cat Test.java
class Test {
  void strictfp method(){} // violation
}
```
Expected - violation reported on line with violation comment

jdkversion values should be integers "1,2,3,4,5,6,7,8,9,10....", as well as support old maven versioning of 1.8, etc...

mahfouz72:
closes #10798

### DIFF
--- a/config/ant-phase-verify.xml
+++ b/config/ant-phase-verify.xml
@@ -53,6 +53,8 @@
                 file="config/suppressions.xml"/>
       <property key="checkstyle.suppressions-xpath.file"
                 file="config/suppressions-xpath.xml"/>
+      <property key="checkstyle.java.version"
+                value="${mvn.java.version}"/>
     </checkstyle>
 
     <tstamp>
--- a/pom.xml
+++ b/pom.xml
@@ -1578,6 +1578,7 @@
               <target>
                 <property name="mvn.project.build.directory" value="${project.build.directory}" />
                 <property name="mvn.runtime_classpath" refid="maven.runtime.classpath" />
+                <property name="mvn.java.version" value="${java.version}"/>
                 <ant antfile="config/ant-phase-verify.xml" />
               </target>
             </configuration>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
@@ -60,6 +60,10 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * {@code record} definitions that are declared as {@code final} and nested
  * {@code record} definitions that are declared as {@code static}.
  * </li>
+ * <li>
+ * {@code strictfp} modifier when using JDK 17 or later. See reason at
+ * <a href="https://openjdk.org/jeps/306">JEP 306</a>
+ * </li>
  * </ol>
  * <p>
  * interfaces by definition are abstract so the {@code abstract} modifier is redundant on them.
@@ -144,6 +148,13 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * </pre>
  * <ul>
  * <li>
+ * Property {@code jdkVersion} - Set the JDK version that you are using.
+ * Old JDK version numbering is supported (e.g. 1.8 for Java 8)
+ * as well as just the major JDK version alone (e.g. 8) is supported.
+ * Type is {@code java.lang.String}.
+ * Default value is {@code "21"}.
+ * </li>
+ * <li>
  * Property {@code tokens} - tokens to check
  * Type is {@code java.lang.String[]}.
  * Validation type is {@code tokenSet}.
@@ -201,6 +212,45 @@ public class RedundantModifierCheck
         TokenTypes.LITERAL_STATIC,
         TokenTypes.ABSTRACT,
     };
+
+    /**
+     *  Constant for jdk 21 version number.
+     */
+    private static final int JDK_21 = 21;
+
+    /**
+     *  Constant for jdk 17 version number.
+     *
+     */
+    private static final int JDK_17 = 17;
+
+    /**
+     * Set the JDK version that you are using.
+     * Old JDK version numbering is supported (e.g. 1.8 for Java 8)
+     * as well as just the major JDK version alone (e.g. 8) is supported.
+     *
+     */
+    private int jdkVersion = JDK_21;
+
+    /**
+     * Setter to set the JDK version that you are using.
+     * Old JDK version numbering is supported (e.g. 1.8 for Java 8)
+     * as well as just the major JDK version alone (e.g. 8) is supported.
+     *
+     * @param jdkVersion the Java version
+     * @since 10.18.0
+     */
+    public void setJdkVersion(String jdkVersion) {
+        final String singleVersionNumber;
+        if (jdkVersion.startsWith("1.")) {
+            singleVersionNumber = jdkVersion.substring(2);
+        }
+        else {
+            singleVersionNumber = jdkVersion;
+        }
+
+        this.jdkVersion = Integer.parseInt(singleVersionNumber);
+    }
 
     @Override
     public int[] getDefaultTokens() {
@@ -260,6 +310,10 @@ public class RedundantModifierCheck
 
         if (isInterfaceOrAnnotationMember(ast)) {
             processInterfaceOrAnnotation(ast);
+        }
+
+        if (jdkVersion >= JDK_17) {
+            checkForRedundantModifier(ast, TokenTypes.STRICTFP);
         }
     }
 
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/modifier/RedundantModifierCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/modifier/RedundantModifierCheck.xml
@@ -33,6 +33,10 @@
  {@code record} definitions that are declared as {@code final} and nested
  {@code record} definitions that are declared as {@code static}.
  &lt;/li&gt;
+ &lt;li&gt;
+ {@code strictfp} modifier when using JDK 17 or later. See reason at
+ &lt;a href="https://openjdk.org/jeps/306"&gt;JEP 306&lt;/a&gt;
+ &lt;/li&gt;
  &lt;/ol&gt;
  &lt;p&gt;
  interfaces by definition are abstract so the {@code abstract} modifier is redundant on them.
@@ -116,6 +120,11 @@
  }
  &lt;/pre&gt;</description>
          <properties>
+            <property default-value="21" name="jdkVersion" type="java.lang.String">
+               <description>Set the JDK version that you are using.
+ Old JDK version numbering is supported (e.g. 1.8 for Java 8)
+ as well as just the major JDK version alone (e.g. 8) is supported.</description>
+            </property>
             <property default-value="METHOD_DEF,VARIABLE_DEF,ANNOTATION_FIELD_DEF,INTERFACE_DEF,CTOR_DEF,CLASS_DEF,ENUM_DEF,RESOURCE,ANNOTATION_DEF,RECORD_DEF"
                       name="tokens"
                       type="java.lang.String[]"
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/ConfigurationUtil.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/ConfigurationUtil.java
@@ -44,6 +44,7 @@ public final class ConfigurationUtil {
         props.setProperty("checkstyle.regexp.header.file", "file");
         props.setProperty("checkstyle.importcontrol.file", "file");
         props.setProperty("checkstyle.importcontroltest.file", "file");
+        props.setProperty("checkstyle.java.version", "value");
 
         return loadConfiguration(path, props);
     }
--- a/src/xdocs/checks/modifier/redundantmodifier.xml
+++ b/src/xdocs/checks/modifier/redundantmodifier.xml
@@ -34,6 +34,10 @@
             <code>record</code> definitions that are declared as <code>final</code> and nested
             <code>record</code> definitions that are declared as <code>static</code>.
           </li>
+          <li>
+            <code>strictfp</code> modifier when using JDK 17 or later. See reason at
+            <a href="https://openjdk.org/jeps/306">JEP 306</a>
+          </li>
         </ol>
 
         <p>
@@ -154,6 +158,13 @@ public class ClassExtending extends ClassExample {
               <th>since</th>
             </tr>
             <tr>
+              <td>jdkVersion</td>
+              <td>Set the JDK version that you are using. Old JDK version numbering is supported (e.g. 1.8 for Java 8) as well as just the major JDK version alone (e.g. 8) is supported.</td>
+              <td><a href="../../property_types.html#String">String</a></td>
+              <td><code>&quot;21&quot;</code></td>
+              <td>10.18.0</td>
+            </tr>
+            <tr>
               <td>tokens</td>
               <td>tokens to check</td>
               <td>subset of tokens
@@ -236,8 +247,8 @@ public class Example1 {
   abstract interface I {
     public abstract void m();
     // 2 violations above:
-    //                   'Redundant 'public' modifier'
-    //                   'Redundant 'abstract' modifier'
+    //    'Redundant 'public' modifier'
+    //    'Redundant 'abstract' modifier'
     public int x = 0; // violation, 'Redundant 'public' modifier'
   }
 
@@ -245,6 +256,7 @@ public class Example1 {
         A, B, C
   }
 
+  // violation below, 'Redundant 'strictfp' modifier'
   public strictfp class Test { }
 
   AutoCloseable lock() {
@@ -283,8 +295,8 @@ public class Example2 {
   abstract interface I {
     public abstract void m();
     // 2 violations above:
-    //                   'Redundant 'public' modifier'
-    //                   'Redundant 'abstract' modifier'
+    //    'Redundant 'public' modifier'
+    //    'Redundant 'abstract' modifier'
     public int x = 0;
   }
 
@@ -297,6 +309,55 @@ public class Example2 {
   AutoCloseable lock() {
     return null;
   }
+}
+        </source>
+        <p id="Example3-config">
+          To configure the check if you are using JDK 11:
+        </p>
+        <source>
+&lt;module name=&quot;Checker&quot;&gt;
+  &lt;module name=&quot;TreeWalker&quot;&gt;
+    &lt;module name=&quot;RedundantModifier&quot;&gt;
+        &lt;property name=&quot;jdkVersion&quot; value=&quot;11&quot;/&gt;
+    &lt;/module&gt;
+  &lt;/module&gt;
+&lt;/module&gt;
+        </source>
+        <p id="Example3-code">
+          Example:
+        </p>
+        <source>
+public class Example3 {
+
+  void test() {
+    // violation below, 'Redundant 'final' modifier'
+    try (final var a = lock()) {
+
+    } catch (Exception e) {
+
+    }
+  }
+
+  // violation below, 'Redundant 'abstract' modifier'
+  abstract interface I {
+    public abstract void m();
+    // 2 violations above:
+    //    'Redundant 'public' modifier'
+    //    'Redundant 'abstract' modifier'
+    public int x = 0; // violation, 'Redundant 'public' modifier'
+  }
+
+  static enum E { // violation, 'Redundant 'static' modifier'
+        A, B, C
+  }
+
+  // ok below, 'strictfp' is not redundant before JDK 17
+  public strictfp class Test { }
+
+  AutoCloseable lock() {
+    return null;
+  }
+
 }
         </source>
       </subsection>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,10 +137,11 @@ workflows:
           name: "no-error-xwiki"
           image-name: "cimg/openjdk:18.0.1"
           command: "./.ci/validation.sh no-error-xwiki"
-      - validate-with-maven-script:
-          name: "no-error-pmd"
-          image-name: *cs_img
-          command: "./.ci/validation.sh no-error-pmd"
+      #      until https://github.com/checkstyle/checkstyle/issues/15475
+      #      - validate-with-maven-script:
+      #          name: "no-error-pmd"
+      #          image-name: *cs_img
+      #          command: "./.ci/validation.sh no-error-pmd"
       - validate-with-maven-script:
           name: "no-exception-struts"
           image-name: *cs_img
--- a/config/checkstyle-checks.xml
+++ b/config/checkstyle-checks.xml
@@ -819,7 +819,9 @@
       <property name="severity" value="ignore"/>
     </module>
     <module name="ModifierOrder"/>
-    <module name="RedundantModifier"/>
+    <module name="RedundantModifier">
+      <property name="jdkVersion" value="${checkstyle.java.version}"/>
+    </module>
 
     <!-- Naming Conventions -->
     <module name="AbbreviationAsWordInName">
--- a/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
@@ -165,6 +165,8 @@ public final class SiteUtil {
         // loads string into memory similar to file
         HEADER_CHECK_HEADER,
         REGEXP_HEADER_CHECK_HEADER,
+        // property is an int, but we cut off excess to accommodate old versions
+        "RedundantModifierCheck.jdkVersion",
         // until https://github.com/checkstyle/checkstyle/issues/13376
         "CustomImportOrderCheck.customImportOrderRules"
     );
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheckTest.java
@@ -354,4 +354,72 @@ public class RedundantModifierCheckTest
         verifyWithInlineConfigParser(
                 getNonCompilablePath("InputRedundantModifierSealedClasses.java"), expected);
     }
+
+    @Test
+    public void testStrictfpWithVersionBeforeJava9() throws Exception {
+        final String[] expected = {
+            "25:5: " + getCheckMessage(MSG_KEY, "abstract"),
+            "27:9: " + getCheckMessage(MSG_KEY, "public"),
+            "27:16: " + getCheckMessage(MSG_KEY, "static"),
+            "34:9: " + getCheckMessage(MSG_KEY, "final"),
+        };
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputRedundantModifierStrictfpWithVersionBeforeJava9.java"),
+                expected);
+    }
+
+    @Test
+    public void testStrictfpWithOldVersion() throws Exception {
+        final String[] expected = {
+            "25:5: " + getCheckMessage(MSG_KEY, "abstract"),
+            "27:9: " + getCheckMessage(MSG_KEY, "public"),
+            "27:16: " + getCheckMessage(MSG_KEY, "static"),
+            "34:9: " + getCheckMessage(MSG_KEY, "final"),
+        };
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputRedundantModifierStrictfpWithOldVersion.java"),
+                expected);
+    }
+
+    @Test
+    public void testStrictfpWithJava17() throws Exception {
+        final String[] expected = {
+            "15:19: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "18:5: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "21:5: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "24:5: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "27:14: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "30:5: " + getCheckMessage(MSG_KEY, "abstract"),
+            "30:14: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "34:9: " + getCheckMessage(MSG_KEY, "public"),
+            "34:16: " + getCheckMessage(MSG_KEY, "static"),
+            "34:23: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "42:9: " + getCheckMessage(MSG_KEY, "final"),
+            "42:15: " + getCheckMessage(MSG_KEY, "strictfp"),
+        };
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputRedundantModifierStrictfpWithJava17.java"),
+                expected);
+    }
+
+    @Test
+    public void testStrictfpWithDefaultVersion() throws Exception {
+        final String[] expected = {
+            "14:19: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "17:5: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "20:5: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "23:5: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "26:14: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "29:5: " + getCheckMessage(MSG_KEY, "abstract"),
+            "29:14: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "33:9: " + getCheckMessage(MSG_KEY, "public"),
+            "33:16: " + getCheckMessage(MSG_KEY, "static"),
+            "33:23: " + getCheckMessage(MSG_KEY, "strictfp"),
+            "41:9: " + getCheckMessage(MSG_KEY, "final"),
+            "41:15: " + getCheckMessage(MSG_KEY, "strictfp"),
+        };
+        verifyWithInlineConfigParser(
+                getNonCompilablePath("InputRedundantModifierStrictfpWithDefaultVersion.java"),
+                expected);
+    }
 }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
@@ -149,6 +149,8 @@ public class XdocsPagesTest {
             // loads string into memory similar to file
             "Header.header",
             "RegexpHeader.header",
+            // property is an int, but we cut off excess to accommodate old versions
+            "RedundantModifier.jdkVersion",
             // until https://github.com/checkstyle/checkstyle/issues/13376
             "CustomImportOrder.customImportOrderRules"
     );
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierStrictfpWithDefaultVersion.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierStrictfpWithDefaultVersion.java
@@ -1,0 +1,46 @@
+/*
+RedundantModifier
+tokens = (default)METHOD_DEF, VARIABLE_DEF, ANNOTATION_FIELD_DEF, INTERFACE_DEF, \
+         CTOR_DEF, CLASS_DEF, ENUM_DEF, RESOURCE
+
+
+*/
+
+//non-compiled with javac: Compilable with Java17
+package com.puppycrawl.tools.checkstyle.checks.modifier.redundantmodifier;
+
+public class InputRedundantModifierStrictfpWithDefaultVersion {
+
+    public static strictfp class Test { }
+    // violation above, 'Redundant 'strictfp' modifier'
+
+    strictfp interface MyInterface { }
+    // violation above, 'Redundant 'strictfp' modifier'
+
+    strictfp enum MyEnum {}
+    // violation above, 'Redundant 'strictfp' modifier'
+
+    strictfp record MyRecord(int x) {}
+    // violation above, 'Redundant 'strictfp' modifier'
+
+    abstract strictfp class MyAbstractClass { }
+    // violation above, 'Redundant 'strictfp' modifier'
+
+    abstract strictfp interface MyStrictFPInterface {
+        // 2 violations above:
+        //                  'Redundant 'abstract' modifier'
+        //                  'Redundant 'strictfp' modifier'
+        public static strictfp enum MyInnerEnum { }
+        // 3 violations above:
+        //                   'Redundant 'public' modifier'
+        //                   'Redundant 'static' modifier'
+        //                   'Redundant 'strictfp' modifier'
+    }
+
+    final class OtherClass {
+        final strictfp void m1() {}
+        // 2 violations above:
+        //                  'Redundant 'final' modifier'
+        //                  'Redundant 'strictfp' modifier'
+    }
+}
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierStrictfpWithJava17.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierStrictfpWithJava17.java
@@ -1,0 +1,47 @@
+/*
+RedundantModifier
+tokens = (default)METHOD_DEF, VARIABLE_DEF, ANNOTATION_FIELD_DEF, INTERFACE_DEF, \
+         CTOR_DEF, CLASS_DEF, ENUM_DEF, RESOURCE
+jdkVersion = 17
+
+
+*/
+
+//non-compiled with javac: Compilable with Java17
+package com.puppycrawl.tools.checkstyle.checks.modifier.redundantmodifier;
+
+public class InputRedundantModifierStrictfpWithJava17 {
+
+    public static strictfp class Test { }
+    // violation above, 'Redundant 'strictfp' modifier'
+
+    strictfp interface MyInterface { }
+    // violation above, 'Redundant 'strictfp' modifier'
+
+    strictfp enum MyEnum {}
+    // violation above, 'Redundant 'strictfp' modifier'
+
+    strictfp record MyRecord(int x) {}
+    // violation above, 'Redundant 'strictfp' modifier'
+
+    abstract strictfp class MyAbstractClass { }
+    // violation above, 'Redundant 'strictfp' modifier'
+
+    abstract strictfp interface MyStrictFPInterface {
+        // 2 violations above:
+        //                  'Redundant 'abstract' modifier'
+        //                  'Redundant 'strictfp' modifier'
+        public static strictfp enum MyInnerEnum { }
+        // 3 violations above:
+        //                   'Redundant 'public' modifier'
+        //                   'Redundant 'static' modifier'
+        //                   'Redundant 'strictfp' modifier'
+    }
+
+    final class OtherClass {
+        final strictfp void m1() {}
+        // 2 violations above:
+        //                  'Redundant 'final' modifier'
+        //                  'Redundant 'strictfp' modifier'
+    }
+}
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierStrictfpWithOldVersion.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierStrictfpWithOldVersion.java
@@ -1,0 +1,37 @@
+/*
+RedundantModifier
+tokens = (default)METHOD_DEF, VARIABLE_DEF, ANNOTATION_FIELD_DEF, INTERFACE_DEF, \
+         CTOR_DEF, CLASS_DEF, ENUM_DEF, RESOURCE
+jdkVersion = 11
+
+
+*/
+
+//non-compiled with javac: Compilable with Java17
+package com.puppycrawl.tools.checkstyle.checks.modifier.redundantmodifier;
+
+public class InputRedundantModifierStrictfpWithOldVersion {
+
+    public static strictfp class Test { }
+
+    strictfp interface MyInterface { }
+
+    strictfp enum MyEnum {}
+
+    strictfp record MyRecord(int x) {}
+
+    abstract strictfp class MyAbstractClass { }
+
+    abstract strictfp interface MyStrictFPInterface {
+        // violation above, 'Redundant 'abstract' modifier'
+        public static strictfp enum MyInnerEnum { }
+        // 2 violations above:
+        //                   'Redundant 'public' modifier'
+        //                   'Redundant 'static' modifier'
+    }
+
+    final class OtherClass {
+        final strictfp void m1() {}
+        // violation above, 'Redundant 'final' modifier'
+    }
+}
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierStrictfpWithVersionBeforeJava9.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierStrictfpWithVersionBeforeJava9.java
@@ -1,0 +1,37 @@
+/*
+RedundantModifier
+tokens = (default)METHOD_DEF, VARIABLE_DEF, ANNOTATION_FIELD_DEF, INTERFACE_DEF, \
+         CTOR_DEF, CLASS_DEF, ENUM_DEF, RESOURCE
+jdkVersion = 1.8
+
+
+*/
+
+//non-compiled with javac: Compilable with Java17
+package com.puppycrawl.tools.checkstyle.checks.modifier.redundantmodifier;
+
+public class InputRedundantModifierStrictfpWithVersionBeforeJava9 {
+
+    public static strictfp class Test { }
+
+    strictfp interface MyInterface { }
+
+    strictfp enum MyEnum {}
+
+    strictfp record MyRecord(int x) {}
+
+    abstract strictfp class MyAbstractClass { }
+
+    abstract strictfp interface MyStrictFPInterface {
+        // violation above, 'Redundant 'abstract' modifier'
+        public static strictfp enum MyInnerEnum { }
+        // 2 violations above:
+        //                   'Redundant 'public' modifier'
+        //                   'Redundant 'static' modifier'
+    }
+
+    final class OtherClass {
+        final strictfp void m1() {}
+        // violation above, 'Redundant 'final' modifier'
+    }
+}
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierIt.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierIt.java
@@ -2,7 +2,7 @@
 RedundantModifier
 tokens = (default)METHOD_DEF, VARIABLE_DEF, ANNOTATION_FIELD_DEF, INTERFACE_DEF, \
          CTOR_DEF, CLASS_DEF, ENUM_DEF, RESOURCE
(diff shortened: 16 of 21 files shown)


mahfouz72:
we will use this property in #15159

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -169,6 +172,11 @@
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#RECORD_DEF">
  * RECORD_DEF</a>.
  * </li>
+ * <li>
+ * Property {@code jdkVersion} - Allow the check to be aware of the Java version.
```
romani, line 176: java versions are very complicated and uncertain term. We need to mention that we use major part of it, and 

https://www.java.com/releases/
we will use only first number 
```
2024-07-16 	CPU 	22.0.2 	
		"    	21.0.4 	
		"    	17.0.12 	
		"    	11.0.24 	
		"    	8u421
```

so 8, 11, ... 
we might just give link to https://en.wikipedia.org/wiki/Java_version_history , attention to `J2SE 1.4` , yes, version was `1.4`.

romani (changes requested):
item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: https://openjdk.org/projects/jdk/

OpenJDK lists the versions with the major part only

mahfouz72:
```
src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierStrictfpWithJava17.java:24: error: cannot find symbol
    strictfp record MyRecord(int x) {}
             ^
  symbol:   class record
  location: class InputRedundantModifierStrictfpWithJava17
```

the file is in the noncompilable folder and has `//non-compilable with javac: Compilable with Java17`. why javac 11 job process it?

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
romani: Ok, put this link to doc. We just need to explain what format we expect.

Should we make property as integer? To double hit on version format.

mahfouz72: Yep, I made it `int` and mentioned that we use major numbers only do we still need to add a link?

mahfouz72:
github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
romani: I tried to search in Google what is "major version of java" and nothing found. So it is not a term, some people  can be confused.

@nrmancuso , @rnveach , what we can do to make it clear? 
May be int type of property is enough to hint what we mean

mahfouz72: > May be int type of property is enough to hint what we mean

I agree with this. It is pretty clear from the property type that it should be a single number only. Also, I think it is common for all Java users to mention the major part only when they asked for their Java version number.

mahfouz72:
github, generate site

mahfouz72:
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/26925/workflows/1b392721-c26a-4bd4-bbb9-4e6b5bfc15a9/jobs/649099

```
[INFO] There are 8 errors reported by Checkstyle 10.18.0-SNAPSHOT with /net/sourceforge/pmd/pmd-checkstyle-config.xml ruleset.
[ERROR] src/main/java/net/sourceforge/pmd/lang/java/ast/ConstantFolder.java:[24,7] (modifier) RedundantModifier: Redundant 'strictfp' modifier.
[ERROR] src/main/java/net/sourceforge/pmd/lang/java/ast/ConstantFolder.java:[141,12] (modifier) RedundantModifier: Redundant 'strictfp' modifier.
[ERROR] src/test/java/net/sourceforge/pmd/lang/java/ast/ConstantExpressionsTests.java:[108,5] (modifier) RedundantModifier: Redundant 'strictfp' modifier.
```
I am not sure what we do to fix this. They should override this property in their config to what suits them

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -261,6 +292,10 @@ public void visitToken(DetailAST ast) {
         if (isInterfaceOrAnnotationMember(ast)) {
             processInterfaceOrAnnotation(ast);
         }
+
+        if (jdkVersion >= JAVA_17) {
+            checkForRedundantModifier(ast, TokenTypes.STRICTFP);
```
romani, line 297: This is breaking behavior compatibility.
I am ok to make default to be most recent version because this modifier is super rear in use.

Please put in issue example on violation message that users might get after update upgrade to latest checkstyle and what change to config they should do. Let's presume they should put `11` in this property to bring behavior back to previous.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -60,6 +60,10 @@
  * {@code record} definitions that are declared as {@code final} and nested
  * {@code record} definitions that are declared as {@code static}.
  * </li>
+ * <li>
+ * {@code strictfp} modifier on java 17 or later. See reason at
```
nrmancuso, line 64: What constructs can this modifier be placed on? records, classes, interfaces, methods, abstract methods, ...?

nrmancuso (changes requested):
Please add an example with this property.

Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: Any , https://github.com/checkstyle/checkstyle/blob/37dd5e815caf7b324cb15fd2da386ab299bc1ea4/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierStrictfpWithJava17.java

That's why I kept it general on the doc and didn't mention specific constructs

mahfouz72: > Please add an example with this property.

I think your request will be blocked until https://github.com/checkstyle/checkstyle/pull/15376

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
nrmancuso: The term is "major version", however this is fairly convoluted by things like the JVM version, classfile versions, etc. We should use "major version"; this, coupled with the fact that this property is an int, is a clear signal that we are looking for 1.5 == 5, 22 == 22, etc.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
nrmancuso: >I am ok to make default to be most recent version because this modifier is super rear in use.

How can we keep this updated? Ideally we would have some source of truth in this repo that declares what java language version we support up to, which is then referenced by some test for this, our docs, etc.

romani: I don't know.

rnveach: We keep this updated by passing [`java.version`](https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L224) to the Checkstyle run and pass this to the property of the check.

I think this would also solidify the other discussion that we must follow the pattern of this field. [In old versions, it was `1.8` for Java 8.](https://github.com/rnveach/checkstyle-backport-jre8/blob/master/pom.xml#L79) I don't know if it supported `8` instead.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
rnveach: My thoughts at https://github.com/checkstyle/checkstyle/pull/15376#discussion_r1693589245 . We could support `1.5` as well as `5`.

Either way, I think this will require a clear description in the documentation.

romani: We should consider to review how maven ,Gradle , have this version. It should be variable of build system to let it be convenient.

If all build system use their own way, we can have our own versioning 5, 6, 7, 8, 9 , 10, 11, .... 22 . We just need to explain this in description

nrmancuso: >Should be variable of build system to let it be convenient

This makes sense, I am good with this proposal.

Maven was `1.x` until Java 9 (for the compiler plugin), gradle is the same (sourceCompatibility/targetCompatibility). We should just make this field a string, and accept 1.x for Java * and below, and X for Java 9 and above, as these build systems have. This way, if someone complains or actual usage shows us otherwise, we can support versions like 8 as well without breaking anything.

nrmancuso: @mahfouz72 let's do this ^, and we can create a ticket to add support for versions like `8` and discuss mapping 1.8 -> 8 there.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
nrmancuso: `java.version` won't be the latest version of java that we support, though. Either way, we can discuss this elsewhere, let's create a ticket if we want to pursue this.

nrmancuso: We can do


>make default to be most recent version because this modifier is super rear in use.

So - Java 21 will be our default since this is the highest version that we support by language level

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
nrmancuso, line 64:
```suggestion
 * {@code strictfp} modifier when building with java 17 or later. See reason at
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -184,6 +193,7 @@
  *
  * @since 3.0
  */
+
```
nrmancuso, line 196: Please revert this

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -202,6 +212,27 @@ public class RedundantModifierCheck
...
+    /**
+     * Constant for Java 17 version number.
+     */
+    private static final int JAVA_17 = 17;
+
+    /**
+     * Allow the check to be aware of the Java version.
+     *
+     */
+    private int jdkVersion = JAVA_17;
```
nrmancuso, line 224:
```suggestion
    private int jdkVersion = JAVA_21;
```

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/modifier/RedundantModifierCheck.xml
```diff
@@ -116,6 +120,9 @@
  }
  &lt;/pre&gt;</description>
          <properties>
+            <property default-value="17" name="jdkVersion" type="int">
```
nrmancuso, line 123:
```suggestion
            <property default-value="21" name="jdkVersion" type="int">
```

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/modifier/RedundantModifierCheck.xml
```diff
@@ -116,6 +120,9 @@
...
+               <description>Allow the check to be aware of the Java version.</description>
```
nrmancuso, line 124:
```suggestion
               <description>Set the Java version that you are using.</description>
```

nrmancuso: We aren't allowing anything, this property is always set now

nrmancuso (changes requested):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: > How can we keep this updated?

how about documenting that
in the check update doc? mention that some checks have `be considered` property these should be consider to be updated

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: done the property field  is string

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: done

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/modifier/RedundantModifierCheck.xml
mahfouz72: done

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/modifier/RedundantModifierCheck.xml
mahfouz72: done

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/Example1.java
```diff
@@ -24,15 +24,16 @@ void test() {
   abstract interface I {
     public abstract void m();
     // 2 violations above:
-    //                   'Redundant 'public' modifier'
-    //                   'Redundant 'abstract' modifier'
+    //    'Redundant 'public' modifier'
+    //    'Redundant 'abstract' modifier'
```
mahfouz72, line 28: because of https://github.com/checkstyle/checkstyle/pull/15379#discussion_r1693661024

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
romani: We still need documentation on what versions  this field can take. Kind of an enumeration.
Versions like 1.8 does not make sense to support as before "17" all was compatible.
We can just explain in property description that we support versions 11,12,13,14,15,16,17,... and up.

Do we have example that users can get value `11` from build system?

Please update checkstyle config to use this property, it will be example for ANT :).

We have to eat food we produce.

rnveach: > java.version won't be the latest version of java that we support, though

I thought we were talking about connecting it to our (CS) version, not for users.

mahfouz72: > We keep this updated by passing [java.version](https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L224) to the Checkstyle run and pass this to the property of the check.


done

pom.xml
```diff
@@ -1578,6 +1578,7 @@
               <target>
                 <property name="mvn.project.build.directory" value="${project.build.directory}" />
                 <property name="mvn.runtime_classpath" refid="maven.runtime.classpath" />
+                <property name="java.version" value="${java.version}"/>
```
romani, line 1581: Can it work without assigning to itself?

config/ant-phase-verify.xml
```diff
@@ -53,6 +53,8 @@
                 file="config/suppressions.xml"/>
       <property key="checkstyle.suppressions-xpath.file"
                 file="config/suppressions-xpath.xml"/>
+      <property key="checkstyle.java.version"
```
romani, line 56: checkstyle.source.java.version

Source are on syntax of jdk11, but actual jdk that compiled can be jdk21.

romani (commented):
Quick glance and pointing:

pom.xml
romani: I got it, it is declaration for ant

config/ant-phase-verify.xml
romani: Ok, here we define target jdk https://github.com/checkstyle/checkstyle/blob/6a3ab019020cf67973e4f3fdc8ace6d197eadddc/pom.xml#L224

romani: Actually, let's keep it as is

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -261,6 +296,10 @@ public void visitToken(DetailAST ast) {
         if (isInterfaceOrAnnotationMember(ast)) {
             processInterfaceOrAnnotation(ast);
         }
+
+        if (Integer.parseInt(jdkVersion) >= JDK_17) {
```
romani, line 300: https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html

How we are going to handle 1.x version ?

What if we split by `.` and take as int first digit.
In case 1.x it will be 1, but we don't care until it becomes 17 :).
We are fully backwards compatible until 17, so anything less than 17 is whatever.

mahfouz72: Yes, we should split 1.x on `.` and take the x. "x" in this case will be any number between 0 and  8. Starting from java 9 it will be 9, 10, 11, .. , 22 only. 

Right now we only support the single-number versions. So people on Java 1.6 should use 6. But we also have the property as a string to keep us flexible if we want to support mapping from 1.x to x. @nrmancuso suggested adding support for this mapping in another issue. Do we need to support it in this PR?

romani: My point it that until 17 we don't need versions at all , so any integer part of not number that is less than 17 is the same version, same behavior.
We don't need to have 21 as value in code as it also meaning

nrmancuso: Yeah, right now we only care if this number is less than 17.

It might be a good idea to abstract the internal representation of this property from the API of the check. We could create a LanguageLevel enum and use logic like this:
 - anything that contains a `.` or up to 16 is LEVEL_LT_17
 - 17 and above can be LEVEL_17

If we separate the language level of the check from the user-facing jdkversion property, we can make "buckets" like this for this check.

Once we add some new logic here, we can just add a new language level enum constant. Each enum can have a `intValue` field.
 -

mahfouz72: > My point it that until 17 we don't need versions at all , so any integer part of not number that is less than 17 is the same version, same behavior.
 
 Yes exactly whatever version number less than 17 will not affect the check behavior. It will be affected only by 17 and 21 (21 is another issue- unnamed variables)
 
 > We don't need to have 21 as value in code as it also meaningless for us, it is just above 17 - all that we need to care.
 
Yes, for now it is meaningless, but we will need 21 in https://github.com/checkstyle/checkstyle/issues/15159 `if (jdkVersion > 21)` violate final on unnamed

nrmancuso: Something like this:
```
    enum JavaLanguageLevel {

        LEVEL_LT_17(16),
        LEVEL_17(17),
        ;

        private final int value;

        JavaLanguageLevel(int value) {
            this.value = value;
        }

        public int getIntValue() {
            return value;
        }

        public JavaLanguageLevel getLevel(String jdkVersion) {
            final JavaLanguageLevel level;
            if (jdkVersion.startsWith("1.")
                    || Integer.parseInt(jdkVersion) < 17) {
                level = LEVEL_LT_17;
            } else {
                level = LEVEL_17;
            }
            return level;
        }
    }
```

romani: ok looks like we in some agreement, let put this in code to discuss details of implementation.

mahfouz72: > It might be a good idea to abstract the internal representation of this property from the API of the check. We could create a LanguageLevel enum and use logic like this:

anything that contains a . or up to 16 is LEVEL_LT_17
17 and above can be LEVEL_17


Very good, done

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/JavaLanguageLevel.java
```diff
@@ -0,0 +1,80 @@
...
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package com.puppycrawl.tools.checkstyle.checks.modifier;
+
+/**
+ * Represents the Java language level.
+ *
+ * @see RedundantModifierCheck
+ */
+public enum JavaLanguageLevel {
```
nrmancuso, line 27: We should make this a inner enum of RedundantModifierCheck, it is unlikely that any other check will need this exact logic. If we start to use this property elsewhere, we can extract and extend this enum.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -261,6 +297,10 @@ public void visitToken(DetailAST ast) {
         if (isInterfaceOrAnnotationMember(ast)) {
             processInterfaceOrAnnotation(ast);
         }
+
+        if (javaLanguageLevel.getIntValue() >= JavaLanguageLevel.LEVEL_17.getIntValue()) {
```
nrmancuso, line 301: We can make a method within the enum to make this read nicely and abstract away the fact that these are backed by integers (or maybe even better: combine these two conditions into a GTE sort of thing)?
```suggestion
        if (javaLanguageLevel == JavaLanguageLevel.LEVEL_17 || javaLanguageLevel.isGreaterThan(JavaLanguageLevel.LEVEL_17)) {
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -202,6 +211,33 @@ public class RedundantModifierCheck
...
+    /**
+     * Constant for jdk 21 version number.
+     */
+    private static final int JDK_21 = 21;
+
+    /**
+     * Set the JDK version that you are using.
+     *
+     */
+    private String jdkVersion = String.valueOf(JDK_21);
```
nrmancuso, line 223:
```suggestion
    /**
     * Set the JDK version that you are using.
     *
     */
    private String jdkVersion = String.valueOf(LEVEL_21.intValue());
```

nrmancuso: This will probably make things a bit easier to reason about and help us to remain "stateless"

nrmancuso (changes requested):
Looking good :)

Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: Yes I think it will be better if we have a GTE method this will always be our check. 
is level >= 17 ? is level >= 21 ?  is level >= 30 ?

something like: 
  ```
      private boolean isGreaterThanOrEqual(JavaLanguageLevel level) {
            return this.getIntValue() >= level.getIntValue();
        }

```

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java (hunk 14 above)
Comment (mahfouz72) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java:223]: Ok but still tests require this check to be filesateful. It is due to the new javaLanguageLevel field

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java (hunk 13 above)
Comment (nrmancuso) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java:301]: yep, sounds good

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java (hunk 14 above)
Comment (nrmancuso) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java:223]: ah shoot, right

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java (hunk 13 above)
Comment (nrmancuso) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java:301]: fwiw, I tend to do `private boolean isGreaterThanOrEqual(JavaLanguageLevel that) ` on instance methods to make things super obvious (another good name is `other`).

**Comment by romani** [comment]
Please remove/postpone any variables/enums for 21 to next  issue.
This issue is only for jdk 17 handling.

**Comment by rnveach** [comment]
@romani Can you point to where 21 issue is. If there is no issue, then @mahfouz72 will need to create one.

**Comment by romani** [comment]
It was referenced at https://github.com/checkstyle/checkstyle/pull/15376#discussion_r1702750768

>Yes, for now it is meaningless, but we will need 21 in https://github.com/checkstyle/checkstyle/issues/15159 if (jdkVersion > 21) violate final on unnamed

**Comment by nrmancuso** [comment]
> Please remove/postpone any variables/enums for 21 to next issue. This issue is only for jdk 17 handling.

We are setting Java 21 as the default. I explicitly requested an enum constant to be created for this.

**Comment by romani** [comment]
After thinking...
If we sent 17 as default, we will update default to 21 in next PR. Changing of default is breaking compatibility, so better to make 21 as default from the beginning.

Do we plan to merge two issues in one PR? Or do them by two separate PRs?

**Comment by mahfouz72** [comment]
@romani @rnveach I kept anything on 17 and removed everything related to 21 even the default value. adding `LEVEL_21` to enum will result in surviving mutations that can't be killed because rn 17 and 21 behave similarly. In another PR we will introduce 21 with its new behavior ( violate final on underscore if java >= 21 )

**Comment by mahfouz72** [comment]
@romani 

> Changing of default is breaking compatibility, so better to make 21 as default from the beginning.

Even if we do not release the old default? I will change it right after we merge this PR

> Changing of default is breaking compatibility, so better to make 21 as default from the beginning.

It should be 2 PRs

**Comment by romani** [comment]
Github, generate website

**Comment by mahfouz72** [comment]
https://checkstyle.semaphoreci.com/jobs/71ce4a21-6f9e-460f-830c-ad260546a6c9

> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.1.0:run (ant-phase-verify) on project sevntu-checks: An Ant BuildException has occured: The following error occurred while executing this line:01:23
[ERROR] /home/semaphore/checkstyle/.ci-temp/sevntu.checkstyle/sevntu-checks/config/ant-phase-verify.xml:25: Unable to create Root Module: config {../../../config/checkstyle-checks.xml}.01:23
[ERROR] around Ant part ...<ant antfile="config/ant-phase-verify.xml" />... @ 10:50 in /home/semaphore/checkstyle/.ci-temp/sevntu.checkstyle/sevntu-checks/target/antrun/build-main.xml: unable to parse configuration stream: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Property ${checkstyle.java.version} has not been set


Do we need to update anything in sevntu repo or something? mvn verify is successful in local. So i don't understand how to fix this

**Comment by rnveach** [comment]
> Do we need to update anything in sevntu repo 

Yes.

https://github.com/checkstyle/checkstyle/blob/d325ed0fb7ede71fe1327bc2bfef362d51bdaf99/.ci/validation.sh#L713
We pass CS config to sevntu to run on sevntu.

https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/pom.xml#L22
Sevntu is set up to use CS's release config to run on its own repo. This CI tests any updates to the config while we are working with the snapshot.

I would hold off on any merges to sevntu. This update won't have any impact on sevntu anyways.

File: pom.xml (hunk 9 above)
Comment (rnveach) [inline pom.xml:1581]: The others are `mvn...`, so I would expect we follow the same pattern here.

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java [hunk 15]
```diff
@@ -202,6 +211,33 @@ public class RedundantModifierCheck
...
+    private JavaLanguageLevel javaLanguageLevel = JavaLanguageLevel.getLevel(jdkVersion);
+
+    /**
+     * Setter to set the JDK version that you are using.
+     *
+     * @param jdkVersion the Java version
+     * @since 10.18.0
+     */
+    public void setJdkVersion(String jdkVersion) {
+        this.jdkVersion = jdkVersion;
```
rnveach, line 237: Why do we need to save the string? It looks like we fully rely on the enumeration.

src/xdocs/checks/modifier/redundantmodifier.xml
```diff
@@ -153,6 +157,13 @@ public class ClassExtending extends ClassExample {
               <th>default value</th>
               <th>since</th>
             </tr>
+            <tr>
+              <td>jdkVersion</td>
+              <td>Set the JDK version that you are using.</td>
```
rnveach, line 162: Documentation does not mention how we support <= 8.

rnveach (changes requested):
quick first review

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -524,4 +564,64 @@ private static List<DetailAST> getMethodAnnotationsList(DetailAST methodDef) {
...
+            final JavaLanguageLevel level;
+            final int jdk17 = 17;
+            if (jdkVersion.startsWith("1.")
+                    || Integer.parseInt(jdkVersion) < jdk17) {
+                level = LEVEL_LT_17;
+            }
+            else {
+                level = LEVEL_17;
+            }
+            return level;
```
rnveach, line 612: As I was talking with @nrmancuso , this enumeration just seems weird to me.

`LEVEL_LT_17(16)` is odd, since now we say its 16 for anything less than 17.
Similarly, we say anything above 17 is just 17.

This `getLevel` method also never fails. I could pass in a property value of `abc` and it would pass and assume I am on 17 (or above). The user would never get any feedback on an issue with what they set it as, and in our case, we don't set it ourselves.

I understand we don't need all possible exact values, but it also seems just weird we would do this for what mostly seems like a simple integer. Even if we add validation to this method to ensure the property value is valid, it seems we could just as easily use the code to manage an `int` over an enumeration.

I don't see the code getting more complex with a simple `int`. `isGreaterThanOrEqual` would just turn into `>=`. We would also not need to worry about breaking compatibility here with future updates.

pom.xml

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: no reason just dummy saving. shouldn't we save it? and why?

src/xdocs/checks/modifier/redundantmodifier.xml
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: done

mahfouz72:
github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
rnveach: If we are switching to the int, then I feel we don't need this. To me this looks like it exists only to make default value on documentation easier to print.

Setter should be able to be a String while a single field is an int. They are not required to be the same type.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -202,6 +214,52 @@ public class RedundantModifierCheck
...
+     * use the single version number (e.g., 9, 10, 11, etc.).
+     *
+     */
+    private String jdkVersion = String.valueOf(JDK_21);
+
+    /**
+     * Integer representation of the JDK version.
+     *
+     */
+    private int jdkVersionValue = Integer.parseInt(jdkVersion);
```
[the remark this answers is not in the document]
rnveach: Similar to another comment, we should only need one. It can be a different type than the setter.

config/pmd-main.xml
```diff
@@ -66,4 +66,16 @@
...
+  <rule ref="category/java/design.xml/SingularField">
+    <properties>
+      <!-- we are intentionally adding 'jdkVersion'
+           as a user-facing property. However, we do not read this field directly;
+           instead, we read from another int field that abstracts the internal implementation
+           of this property.
+      -->
+      <property name="violationSuppressXPath"
+         value="//ClassDeclaration[@SimpleName='RedundantModifierCheck']
+                       //FieldDeclaration//VariableDeclarator[@Name='jdkVersion']"/>
```
rnveach, line 78: Is this similar to my other comment on a single field?

mahfouz72: yes, If we can make the property is int then we don't need this suppression

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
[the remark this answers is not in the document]
mahfouz72: I think the type in the doc is generated from the type of the field. So, we can't say `int x;` and expect it to be a String in the doc. but let me try again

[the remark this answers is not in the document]
mahfouz72: Yup, the doc and class fields have to be the same type. :(

when I changed the type to int. and leave only one field
```
     <li>
    -Property {@code jdkVersion} - Set the JDK version that you are using. We support JDK versions in both notations for versions up to and including 8 (e.g., 5 and 1.5, 8 and 1.8). For versions 9 and later, use the single version number (e.g., 9, 10, 11, etc.). Type is {@code int}. Default value is {@code 21}.
    +Property {@code jdkVersion} - Set the JDK version that you are using. We support JDK versions in both notations for versions up to and including 8 (e.g., 5 and 1.5, 8 and 1.8). For versions 9 and later, use the single version number (e.g., 9, 10, 11, etc.). Type is {@code java.lang.String}. Default value is {@code "21"}.
     </li>
```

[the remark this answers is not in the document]
mahfouz72: So thats why I have 2 differents fields one for a user-facing property (String) and the equivalent integer for us to work with.
This is a violation by PMD and IDEA inspections they said `jdkVersion` can be a local varaiable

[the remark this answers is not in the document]
mahfouz72: This also results in a pitest issue because we aren't using this variable (so removing its assignment(default value) will survive) I tried a workaround to satisfy pitest (used the `jdkVersion` field in the intialization of `jdkVersionValue` -` private int jdkVersionValue = Integer.pareInt(jdkVersion);` ), spotbugs is complaining (The object under construction remains partially initialized and may be vulnerable to Finalizer attacks.). I can look more into this to find a solution that satisfy all tools but I think supporting the single version number even when jdk <= 8 will be better at this point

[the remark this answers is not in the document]
mahfouz72: do we agree on any suppressions for any tool?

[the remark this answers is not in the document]
rnveach: https://github.com/checkstyle/checkstyle/pull/15376#discussion_r1706279547
> -.....Type is {@code int}.
> +.....Type is {@code java.lang.String}.

This is what we want. No suppression needed. Our type is string since we are allowing "1.X", which is not an integer.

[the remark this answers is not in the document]
rnveach: We can override property type display in `XDocsPagesTest`. Add property to `PROPERTIES_ALLOWED_GET_TYPES_FROM_METHOD`. Similar list is in `SiteUtil` for xdoc macro side.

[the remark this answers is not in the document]
mahfouz72: I added `RedundantModifier.jdkVersion` to both lists and still the property type is regenerated from the field (as int)
commit : https://github.com/checkstyle/checkstyle/commit/3b67a960fcd3599cbac80ad4604b61f9307a363b
CI failures: https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=23078&view=logs&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111&l=1828


why not to avoid all of these and have just a single string field and parse it to integer when needed in methods. I don't see the code getting complex with a simple `Integer.parseInt()` when needed.

romani (approved):
Ok to merge

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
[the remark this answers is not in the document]
mahfouz72: > why not to avoid all of these and have just a single string field and parse it to integer when needed in methods. I don't see the code getting complex with a simple Integer.parseInt() when needed.

I made this

mahfouz72:
github, generate site

mahfouz72:
@romani please consider https://github.com/checkstyle/checkstyle/pull/15376#discussion_r1706917123

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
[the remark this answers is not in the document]
rnveach: > why not to avoid all of these and have just a single string field and parse it to integer when needed in methods

I think you should try it out.

Run that version of the code on a big project. Turn halt on exception off, and set the property to "abc".

Examine what happens and also when it happens.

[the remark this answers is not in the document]
romani: We need to keep String as interface for user.
I am ok with any small performance issues, we can resolve them separately.

We should be able to introduce new class field that will keep int version, so justification of this extra field will be: metadata model defect/feature that is using class field type and not a setter type.

If extra class field is caught be some validation, we can suppress them. We just need to minimize suppressions.

May be one day we will improve metadata and this weird code will be optimized.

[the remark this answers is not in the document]
mahfouz72: Please see current code and let me now if you have any comments for now. 

I used only one field with String type.

> Run that version of the code on a big project. Turn halt on exception off, and set the property to "abc".
>
>Examine what happens and also when it happens.
>
>Then run a similar experiment with another check like JavaNCSS and set classMaximum to "abc". Both fields of these checks are internally int.

In both checks, a `NumberFormatException` is thrown

I am good with the current code we can improve it in the future if needed. adding an "extra" field that keeps the int version will cost us a lot of violations by pitest, PMD, spotbugs and IDEA inspections :).

Also making a single int field with a string setter is not working properly, metadata generation is weird and it always regenerates the type from the class field.

IMO, we can keep the current code as it is and parse the string field to int when needed. also consider that parsing to int will be in very minimal cases till now we will need it for strictfp and unnamed variables only so I think performance issues is not a big concern here.

rnveach:
> I would hold off on any merges to sevntu. This update won't have any impact on sevntu anyways.

Since I am last and I am good with the property names, you can start sending PRs to sevntu and contribution to fix the CI here.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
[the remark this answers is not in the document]
rnveach: > In both checks, a NumberFormatException is thrown

If that is all you found, then I don't feel you did one or more of the following:
> Run that version ... on a big project
> Turn halt on exception off
> Examine if they are different and how.

=====

It has been an eternity since I have written this code. `PROPERTIES_ALLOWED_GET_TYPES_FROM_METHOD` allows us to use the setter's type ONLY if the field cannot be found. If we make the field invisible to use that override, we lose the default value of the field.

I have submitted a PR to update PROPERTIES_ALLOWED_GET_TYPES_FROM_METHOD logic so it uses the setter even if the field exists. This override is only used by 4 other checks before this PR, so it should be minimal impact.

The PR is at https://github.com/checkstyle/checkstyle/pull/15466 .

Here is your new patch. It will require the other PR to function properly and CI pass. Feel free to add better verbage.
````
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java b/src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
index 804761e..5538766 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
@@ -232,7 +232,7 @@
      * use the single version number (e.g., 9, 10, 11, etc.).
      *
      */
-    private String jdkVersion = String.valueOf(JDK_21);
+    private int jdkVersion = JDK_21;
 
     /**
      * Setter to set the JDK version that you are using.
@@ -245,10 +245,10 @@
      */
     public void setJdkVersion(String jdkVersion) {
         if (jdkVersion.startsWith("1.")) {
-            this.jdkVersion = jdkVersion.substring(2);
+            this.jdkVersion = Integer.parseInt(jdkVersion.substring(2));
         }
         else {
-            this.jdkVersion = jdkVersion;
+            this.jdkVersion = Integer.parseInt(jdkVersion);
         }
     }
 
@@ -312,7 +312,7 @@
             processInterfaceOrAnnotation(ast);
         }
 
-        if (Integer.parseInt(jdkVersion) >= JDK_17) {
+        if (jdkVersion >= JDK_17) {
             checkForRedundantModifier(ast, TokenTypes.STRICTFP);
         }
     }
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java b/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
index bbdd341..82add5d 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
@@ -165,6 +165,8 @@
         // loads string into memory similar to file
         HEADER_CHECK_HEADER,
         REGEXP_HEADER_CHECK_HEADER,
+        // property is an int, but we cut off excess to accommodate old versions
+        "RedundantModifierCheck.jdkVersion",
         // until https://github.com/checkstyle/checkstyle/issues/13376
         "CustomImportOrderCheck.customImportOrderRules"
     );
diff --git a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
index 44f9ab2..813e589 100644
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
@@ -149,6 +149,8 @@
             // loads string into memory similar to file
             "Header.header",
             "RegexpHeader.header",
+            // property is an int, but we cut off excess to accommodate old versions
+            "RedundantModifier.jdkVersion",
             // until https://github.com/checkstyle/checkstyle/issues/13376
             "CustomImportOrder.customImportOrderRules"
     );
````

**Comment by mahfouz72** [comment]
Test commit: https://github.com/checkstyle/checkstyle/pull/15376/commits/4baa2606b7e7d7a47dc253d9105b1c056ddeadc9
build URL: https://checkstyle.semaphoreci.com/workflows/4d7cd252-50aa-4653-b70d-3df56357de34?pipeline_id=f93bb76c-1cab-4317-9456-04accba9d232

sempahoreci build is successful

sventu PR: https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/1036
contrbuiton repo PR: https://github.com/checkstyle/contribution/pull/886

**Comment by mahfouz72** [comment]
github, generate site

File: src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/ConfigurationUtil.java [hunk 20]
```diff
@@ -44,6 +44,7 @@ public static Configuration loadConfiguration(String path) throws CheckstyleExce
         props.setProperty("checkstyle.regexp.header.file", "file");
         props.setProperty("checkstyle.importcontrol.file", "file");
         props.setProperty("checkstyle.importcontroltest.file", "file");
+        props.setProperty("checkstyle.java.version", "value");
```
rnveach, line 47: I feel you made this invalid by not using a number. If this doesn't fail when it loads the CS config in the test, then I would want to know why as that just questions why we then need to populate it.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/naming/illegalidentifiername/InputIllegalIdentifierNamePatternMatchingForSwitch.java
```diff
@@ -0,0 +1,35 @@
+/*
+IllegalIdentifierName
+format = (default)(?i)^(?!(record|yield|var|permits|sealed|when)$).+$
+tokens = (default)CLASS_DEF, INTERFACE_DEF, ENUM_DEF, ANNOTATION_DEF, ANNOTATION_FIELD_DEF, \
```
rnveach, line 4: Why is this check, `IllegalIdentifierName` being modified in this PR for `RedundantModifier`?

src/xdocs/checks/modifier/redundantmodifier.xml
```diff
@@ -153,6 +157,13 @@ public class ClassExtending extends ClassExample {
               <th>default value</th>
               <th>since</th>
             </tr>
+            <tr>
+              <td>jdkVersion</td>
+              <td>Set the JDK version that you are using. We support JDK versions in both notations for versions up to and including 8 (e.g., 5 and 1.5, 8 and 1.8). For versions 9 and later, use the single version number (e.g., 9, 10, 11, etc.).</td>
```
rnveach, line 162: I find it odd you say "both" versions before you identify what the different versions are. I would avoid pronouns as it is the documentation conveying to the user what is going on, not us directly.

`We support JDK versions in both notations for versions up to and including 8 (e.g., 5 and 1.5, 8 and 1.8). For versions 9 and later, use the single version number (e.g., 9, 10, 11, etc.).`
==>
`Old JDK version numbering is supported (e.g. 1.8 for Java 8) as well as just the major JDK version alone (e.g. 8) is supported.`

rnveach (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/ConfigurationUtil.java
mahfouz72: I am not sure but maybe when the parser encounters `${checkstyle.java.version}`, it looks for a corresponding entry in the properties list. If it doesn't find one, it fails with the error (`unable to parse configuration stream
 Property ${checkstyle.java.version} has not been set`) to enforce that all placeholders must be populated so that the configuration can be successfully loaded. if a placeholder like` ${checkstyle.java.version} `is found in the configuration file but isn't populated in the properties passed to` loadConfiguration()`, the parser can't resolve the placeholder, leading to the error.
 
However, even though the property setting is required to prevent the error, setting it to an invalid value like "value" won't cause an error because we are not actually loading the modules (checks) here. we are just parsing the config stream to test what tokens are missing from `checkstyle-checks.xml`

.ci/validation.sh
```diff
@@ -211,28 +211,29 @@ markdownlint)
...
-                -Dmaven.source.skip=true \
-                -Dpmd.skip=true \
-                -Dcheckstyle.skip=false \
-                -Dcheckstyle.version="${CS_POM_VERSION}"
-  cd ..
-  removeFolderWithProtectedFiles build-tools
-  removeFolderWithProtectedFiles pmd
-  ;;
+# until https://github.com/checkstyle/checkstyle/issues/15475
+# no-error-pmd)
```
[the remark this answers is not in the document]
mahfouz72: done

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/naming/illegalidentifiername/InputIllegalIdentifierNamePatternMatchingForSwitch.java
mahfouz72: Removed.
probably a leakage between branches (check out to this branch without committing this file in its branch)

src/xdocs/checks/modifier/redundantmodifier.xml
mahfouz72: done

src/test/java/com/puppycrawl/tools/checkstyle/internal/utils/ConfigurationUtil.java
rnveach: Ok thanks, that explained it. It isn't until `rootModule.configure` for when we start creating checks, and assigning properties. That is where it should fail on an invalid property value.

rnveach (approved):

PR #15377, opened 2024-07-26
Issue #15036: Add new input for nestedIfDepth

### LINKED ISSUE #15036: Add Check Support for Java 21 Pattern Matching for Switch Syntax: NestedIfDepth

I have read check documentation:https://checkstyle.org/checks/coding/nestedifdepth.html

```

public class Test {

    void test(Object obj) {
        switch (obj) {
            case ColoredPoint(int x, int y, String z) when (x >= 2)
                    -> {
                if (x >= 2) {
                    if(y >= 2) {
                        if (z.equals("red")) {

                        }
                    }
                }
            }
            default -> {}
        }
    }
}
PS D:\CS\test> cat config.xml                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="NestedIfDepth">

        </module>
    </module>
</module>
Starting audit...
[ERROR] D:\CS\test\src\Test.java:12:25: Nested if-else depth is 2 (max allowed is 1). [NestedIfDepth]
Audit done.

```

---

`when` can't be nested and nested `ifs` in the switch blocks work correctly. We are ok here. This can be closed.

mahfouz72:
closes #15036

IMO, very unneeded input

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/NestedIfDepthCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/NestedIfDepthCheckTest.java
@@ -59,6 +59,17 @@ public class NestedIfDepthCheckTest extends AbstractModuleTestSupport {
     }
 
     @Test
+    public void testInsideCaseBody() throws Exception {
+
+        final String[] expected = {
+            "17:25: " + getCheckMessage(MSG_KEY, 2, 1),
+        };
+
+        verifyWithInlineConfigParser(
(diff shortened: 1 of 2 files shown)


romani:
I like weird code in inputs, it will prevent regression for sure one time in future.

If someone will try to change grammar of java , all of this becomes very useful.
Diff report generation use this input for all types of Checks .

It is not wasted time to merge do this PR

nrmancuso (approved):

PR #15398, opened 2024-07-28
Issue #15214: migrated section 4.8.6.1 block comment style to whole config testing

Zopsss:
#15214

### DIFF
--- a/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4861blockcommentstyle/InputCommentsIndentationInSwitchBlock.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4861blockcommentstyle/InputCommentsIndentationInSwitchBlock.java
@@ -1,5 +1,6 @@
 package com.google.checkstyle.test.chapter4formatting.rule4861blockcommentstyle;
 
+/** some javadoc. */
 public class InputCommentsIndentationInSwitchBlock {
 
   private static void fooSwitch() {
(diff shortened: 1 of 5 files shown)


romani (approved):
ok to merge.

diff is not easy, mostly change in line numbers.
there are  bunch of other fixes for unrelated Checks - looks good.

src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4861blockcommentstyle/InputCommentsIndentationInSwitchBlock.java
```diff
@@ -18,75 +19,73 @@ private static void fooSwitch() {
...
       case "5":
-        // violation 2 lines below '.* indentation should be the same level as line 27, 29.'
+        // violation 2 lines below '.* indentation should be the same level as line 30.'
         foo1();
            // odd indentation comment
+        break;
       case "6":
         int k = 7;
-        // fall through
+        break;
```
rdiachenko, line 33: Why did you replace `// fall through` with `break` here, but didn't do it few lines below (line 43)?

Zopsss: sorry my bad. In some places it was giving `fall through` violation so I had to add break statement everywhere, I might have accidentally removed the comment here and added the break statement here. I have revert it, the comment is back now.

It is strange that if we have `// fall through` comment instead of `break;` then we don't get the fall through violation.

In some places this comment was inside the block so I have moved them outside to prevent violation. Example: https://github.com/checkstyle/checkstyle/pull/15398/files?diff=unified&w=0#diff-a2fac517f687a7b67e6183b98412714725d08a574219ae15e00d974d9ad23532R44-R47

rdiachenko (approved):
lgtm

Issue #15400, opened 2024-07-28
Remove condition for triggering new workflow on active PR


relentless-pursuit:
Remove below as it doesn't seem to work

```
on:
  pull_request:
    types:
      - opened
```

romani:
It works in different ways: it triggers on operating of PR.

https://github.com/checkstyle/test-configs/issues/131#issuecomment-2254302694

relentless-pursuit:
> It works in different ways: it triggers on operating of PR.
> 
> [checkstyle/test-configs#131 (comment)](https://github.com/checkstyle/test-configs/issues/131#issuecomment-2254302694)

Okay, i checked with closed PRs. I didn't see that for merged PRs. For closed PRs, it doesn't work though. For now, i have removed it and can come back to it later.

PR #15401, opened 2024-07-28
Remove condition for triggering new workflow on active PR

relentless-pursuit:
Resolves: #15400

### DIFF
--- a/.github/workflows/regression-report.yml
+++ b/.github/workflows/regression-report.yml
@@ -20,9 +20,6 @@ env:
 on:
   issue_comment:
     types: [ created ]
-  pull_request:
-    types:
-      - opened
 
 permissions:
   contents: read


.github/workflows/regression-report.yml
```diff
@@ -19,10 +19,7 @@ env:
 
 on:
   issue_comment:
-    types: [ created ]
-  pull_request:
-    types:
-      - opened
+    types: [ created, edited ]
```
romani, line 23: We don't need "edited"

romani (changes requested):
Items

.github/workflows/regression-report.yml
relentless-pursuit: done

romani (approved):
Ok to merge

PR #15411, opened 2024-07-31
supplemental: added test cases for JUnit test method names in google style guide section 5.2.3 method names

Zopsss:
Detected at: https://github.com/checkstyle/checkstyle/pull/14901/files/4bcfed8ef4a4446fbda32e24fd067abd933030ce#r1679695992

Added test case for JUnit test methods. The test works same even if I remove the `@Test` annotation. I'm not sure if we should be concerned about this or not.

### DIFF
--- a/src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
@@ -1,5 +1,7 @@
 package com.google.checkstyle.test.chapter5naming.rule523methodnames;
 
+import org.junit.jupiter.api.Test;
+
 /**
  * Test input for MethodNameCheck specifically whether the method name equals the class name.
  *
@@ -22,6 +24,27 @@ public class InputMethodName {
 
   void fO() {} // violation 'Method name 'fO' must match pattern'
 
+  @Test
+  void testing_foo() {}
+
+  @Test
+  void testing_Foo() {}
+
+  @Test
+  void testing_fOo() {}
+
+  @Test
+  void testingFoo() {}
+
+  @Test
+  void testingFoo_foo() {}
+
+  @Test
+  void testing_0123() {}
+
+  @Test
+  void TestingFooBad() {} // violation 'Method name 'TestingFooBad' must match pattern'
+
   class InnerFoo {
     void foo() {}
 
@@ -38,6 +61,20 @@ public class InputMethodName {
     void f() {} // violation 'Method name 'f' must match pattern'
 
     void fO() {} // violation 'Method name 'fO' must match pattern'
+
+    void testing_foo() {}
+
+    void testing_Foo() {}
+
+    void testing_fOo() {}
+
+    void testingFoo() {}
+
+    void testingFoo_foo() {}
+
+    void testing_0123() {}
+
+    void TestingFooBad() {} // violation 'Method name 'TestingFooBad' must match pattern'
   }
 
   InnerFoo anon =


src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
```diff
@@ -22,6 +24,24 @@ void f() {} // violation 'Method name 'f' must match pattern'
 
   void fO() {} // violation 'Method name 'fO' must match pattern'
 
+  @Test
+  void testing_foo() {}
+
+  @Test
+  void testing_Foo() {}
```
Zopsss, line 31: I think there should be violation for this method, after underscore the first character should be lowercase letter. @rdiachenko what do you think about this?

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
```diff
@@ -22,6 +24,27 @@ void f() {} // violation 'Method name 'f' must match pattern'
...
+  void testing_fOo() {}
+
+  @Test
+  void testingFoo() {}
+
+  @Test
+  void testingFoo_foo() {}
+
+  @Test
+  void testing_0123() {}
```
Zopsss, line 43: same as above, after underscore the first letter should not be a number? But I'm not so sure about this as documentation does not mentions anything about using digits.

[5.2.3 Method names](https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names)

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
Zopsss: but documentation does not explicitly mentions anything about this so I'm not sure if we should give violation for this or not

[5.2.3 Method names](https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names)

rdiachenko: It should be a violation here. The doc mentions this in the following sentence:

> with each component written in `lowerCamelCase`

where `lowerCamelCase` is example of correct component naming. If the first character after underscore were allowed, they would use this wording instead:

> with each component written in `UpperCamelCase`

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
```diff
@@ -22,6 +24,27 @@ void f() {} // violation 'Method name 'f' must match pattern'
 
   void fO() {} // violation 'Method name 'fO' must match pattern'
 
+  @Test
+  void testing_foo() {}
```
rdiachenko, line 28: Please add similar method but without `@Test` annotation, I think we should expect violation for non-test methods according to the doc:

> Underscores may appear in JUnit **test** method names

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
rdiachenko: They say nothing about numbers. Could you check if there are rules in other sections regarding the usage of numbers in names?

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
Zopsss: I see. So there's a defect from our side

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
Zopsss: I found this from [5.1 Rules common to all identifiers](https://google.github.io/styleguide/javaguide.html#s5.1-identifier-names):

> Identifiers use only ASCII letters and digits, and, in a small number of cases noted below, underscores

though this section is not about method naming but it does mentions that numbers are allowed.

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
Zopsss: done, pushed the changes:

https://github.com/checkstyle/checkstyle/blob/29d91cab6fffc7375b24dd7656e0f08ca1b117fb/src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java#L65-L77

though we're checking the JUnit methods and normal methods same way, there's no difference in violations. We are not checking if the method is a test method or not

Zopsss: MethodName module from google_checks.xml:

https://github.com/checkstyle/checkstyle/blob/e071377c4d6e049491aef610b04bc106e8a8b6a3/src/main/resources/google_checks.xml#L353-L357

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
romani: from coverage table:[5.2.3 Method names](https://checkstyle.org/styleguides/google-java-style-20180523/javaguide.html#s5.2.3-method-names) 	[MethodName](https://checkstyle.org/checks/naming/methodname.html#MethodName)
No ability to distinguish verb at the beginning of name, no way distinguish Test method from others so "_" is allowed in names

I do not understand why this is not a violation:
```
@Test
  void testing_Foo() {}
```

to archive allowance of `_` in test methods we should make Xpath suppression that check that method has `Test` annotation (just text comparison, as there bunch of test libraries in this world, they did not mean only junit).
In addition to this suppression instance we need to add checking for message that contains `_` in message.

in this case  we can remove this `no way distinguish Test method from others so "_" is allowed in names` from comments.

------------

>testing_Foo

from abbreviation perspective this is correct,  camel case is checked by  AbbreviationAsWordInName.
`_` is separator of words. In style guide it is covered by `Divide this result into words, splitting on spaces and any remaining punctuation (typically hyphens). `

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
romani: number that starts a word is ok, they have example: `"Turn on 2SV" --->	turnOn2sv`

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
romani: @Zopsss , please create separate issue on this to clarify it separately.
we can merge this PR and issue will point to merged code and question correctness of violations or missed violations.

Zopsss: > @Zopsss, please explain why there is no violation.

defect in our regex:

https://github.com/checkstyle/checkstyle/blob/9da5170cdd0d9e82803c83eb2cc23a0c08d370ee/src/main/resources/google_checks.xml#L353-L357

We only check for starting 2 characters, after that `\w` comes which allows for underscore, digits and alphabets. Then we don't validate anything. That's why there's no violation

Zopsss: > @Zopsss , please create separate issue on this to clarify it separately.

done https://github.com/checkstyle/checkstyle/issues/15425

Zopsss: > @Zopsss, please explain why there is no violation.

as you explained below in your comment, it is a defect our Check. Once it is fixed, we will have violation for this method

romani: >to archive allowance of _ in test methods we should make Xpath suppression that check that method has Test annotation (just text comparison, as there bunch of test libraries in this world, they did not mean only junit).

Let's create issue on this. It is still a gap.

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
Zopsss: but numbers directly after `_` are ok? Nobody write name likes this, so I'm bit confuse

romani: Turn on 2SV --> Turn_on_2SV

> Nobody write name likes this

https://en.wikipedia.org/wiki/Snake_case , it is one of styles.
Java trending more to CamleCase style.

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
romani: https://github.com/checkstyle/checkstyle/commit/0eadd97b8af5b79e7c49f82958fbb3d863bfdf61 is where `_` is explicitly disappear and  becomes covered by `\w`.

romani: postponed until https://github.com/checkstyle/checkstyle/issues/15429

romani (approved):
this input show false negatives
that will be addressed in scope of https://github.com/checkstyle/checkstyle/issues/15429

rdiachenko (approved):
lgtm

PR #15430, opened 2024-08-03
Issue #15047: Add support for pattern matching in switch in NPathComp…

### LINKED ISSUE #15047: Add Check Support for Java 21 Pattern Matching for Switch Syntax: NPathComplexity

I have read check documentation:https://checkstyle.org/checks/metrics/npathcomplexity.html

---

```
c cat config.xml                               
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="NPathComplexity">
            <property name="max" value="1"/>
        </module>
    </module>
</module>
➜  src cat Test.java                                
record ColoredPoint(int p, int x, boolean c) { }

public class Test {
    void test(Object obj) {
        switch (obj) {
            case ColoredPoint(int a, int b, _) when (a >= b) : {} break;
            case ColoredPoint(int a, int b, _) when (b > 100) : {} break;
            case ColoredPoint(int a, int b, _) when (b != 1000) : {} break;
            case ColoredPoint(int a, int b, _) when (b != 100000) : {} break;
            case Rectangle(_,_) : {}
            default : System.out.println("none");
        };
    }

    void test2(Object obj) {
        switch (obj) {
            case ColoredPoint(int a, int b, _) when (a >= b) -> {}
            case ColoredPoint(int a, int b, _) when (b > 100) -> {}
            case ColoredPoint(int a, int b, _) when (b != 1000) -> {}
            case ColoredPoint(int a, int b, _) when (b != 100000) -> {}
            case Rectangle(_,_) -> {}
        };
    }
}
Starting audit...
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:5:5: NPath Complexity is 6 (max allowed is 1). [NPathComplexity]
[ERROR] /home/nick/IdeaProjects/tester/src/Test.java:16:5: NPath Complexity is 2 (max allowed is 1). [NPathComplexity]
Audit done.


```

---

I expect both methods to have the same NPathComplexity. There is a problem in counting when case labels are in the switch rule `(->)`

mahfouz72:
closes #15047

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
@@ -139,6 +139,14 @@ public final class NPathComplexityCheck extends AbstractCheck {
      */
     public static final String MSG_KEY = "npathComplexity";
 
+    /** Tokens that are considered as case labels. */
+    private static final int[] CASE_LABEL_TOKENS = {
+        TokenTypes.EXPR,
+        TokenTypes.PATTERN_DEF,
+        TokenTypes.PATTERN_VARIABLE_DEF,
+        TokenTypes.RECORD_PATTERN_DEF,
+    };
+
     /** Default allowed complexity. */
     private static final int DEFAULT_MAX = 200;
 
@@ -508,17 +516,21 @@ public final class NPathComplexityCheck extends AbstractCheck {
     }
 
     /**
-     * Counts number of case constants (EXPR) tokens in a switch labeled rule.
+     * Counts number of case constants tokens in a switch labeled rule.
      *
      * @param ast switch rule token.
-     * @return number of case constant (EXPR) tokens.
+     * @return number of case constant tokens.
      */
     private static int countCaseConstants(DetailAST ast) {
         final AtomicInteger counter = new AtomicInteger();
         final DetailAST literalCase = ast.getFirstChild();
 
-        TokenUtil.forEachChild(literalCase,
-            TokenTypes.EXPR, node -> counter.getAndIncrement());
+        for (DetailAST node = literalCase.getFirstChild(); node != null;
+                    node = node.getNextSibling()) {
+            if (TokenUtil.isOfType(node, CASE_LABEL_TOKENS)) {
+                counter.getAndIncrement();
+            }
+        }
 
         return counter.get();
     }
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/npathcomplexity/InputNPathComplexityPatternMatchingForSwitch.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/npathcomplexity/InputNPathComplexityPatternMatchingForSwitch.java
@@ -1,0 +1,105 @@
+/*
+NPathComplexity
+max = 1
+
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.metrics.npathcomplexity;
+
+public class InputNPathComplexityPatternMatchingForSwitch {
+
+    // violation below, 'NPath Complexity is 3 (max allowed is 1)'
+    void testPatternWithRule(Object o) {
+        switch (o) {
+            case Integer _ -> {}
+            case String _ -> {}
+            default -> {}
+        }
+    }
+
+    // violation below, 'NPath Complexity is 3 (max allowed is 1)'
+    void testPatternWithStatement(Object o) {
+        switch (o) {
+            case Integer _: {}
+            case String _: {}
+            default : {}
+        }
+    }
+
+    // violation below, 'NPath Complexity is 3 (max allowed is 1)'
+    void testRecordPatternWithRule(Object o) {
+        switch (o) {
+            case A(int x) -> {}
+            case B(String s) -> {}
+            default -> {}
+        }
+    }
+
+    // violation below, 'NPath Complexity is 3 (max allowed is 1)'
+    void testRecordPatternWithStatement(Object o) {
+        switch (o) {
+            case A(int x) : {} break;
+            case B(String s) : {}
+            default : {}
+        }
+    }
+
+    // violation below, 'NPath Complexity is 3 (max allowed is 1)'
+    void testGuardsInRule(Object o) {
+        switch (o) {
+            case Integer i when i > 0 -> {}
+            case String s when s.length() > 0 -> {}
+            default -> {}
+        }
+    }
+
+    // violation below, 'NPath Complexity is 3 (max allowed is 1)'
+    void testGuardsInStatement(Object o) {
+        switch (o) {
+            case Integer i when i > 0 : {} break;
+            case String s when s.length() > 0 : {}
+            default : {}
+        }
+    }
+
+    // violation below, 'NPath Complexity is 4 (max allowed is 1)'
+    void testMultiCaseLabelWithRule(Object o) {
+        switch (o) {
+            case Integer _, String _ , Double _ -> {}
+            default -> {}
+        }
+    }
+
+    // violation below, 'NPath Complexity is 4 (max allowed is 1)'
+    void testMultiCaseLabelWithStatement(Object o) {
+        switch (o) {
+            case Integer _ :
+            case String _  :
+            case Double _  : {}
+            default : {}
+        }
+    }
+
+    // violation below, 'NPath Complexity is 3 (max allowed is 1)'
+    void testExprInRule(int x) {
+        switch (x) {
+            case 1 -> {}
+            case 2 -> {}
+            default -> {}
+        }
+    }
+
+    // violation below, 'NPath Complexity is 3 (max allowed is 1)'
+    void testExprInStatement(int x) {
+        switch (x) {
+            case 1 : {}
+            case 2 : {}
+            default : {}
+        }
+    }
+
+    record A(int x) {}
+    record B(String s) {}
+}
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheckTest.java
@@ -250,6 +250,28 @@ public class NPathComplexityCheckTest extends AbstractModuleTestSupport {
     }
 
     @Test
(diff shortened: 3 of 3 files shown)


mahfouz72:
Github, generate report for NPathComplexity/Example2

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
```diff
@@ -508,17 +507,24 @@ private static int countCaseTokens(DetailAST ast) {
...
         final DetailAST literalCase = ast.getFirstChild();
 
-        TokenUtil.forEachChild(literalCase,
-            TokenTypes.EXPR, node -> counter.getAndIncrement());
+        for (DetailAST node = literalCase.getFirstChild(); node != null;
+                    node = node.getNextSibling()) {
+            if (node.getType() == TokenTypes.EXPR
+                    || node.getType() == TokenTypes.PATTERN_VARIABLE_DEF
+                    || node.getType() == TokenTypes.RECORD_PATTERN_DEF
+                    || node.getType() == TokenTypes.PATTERN_DEF) {
```
nrmancuso, line 524: Please extract this to a Set and do `caseConstantTokens.contains(node.getType)`

nrmancuso (changes requested):
Report looks good, one minor item:

nrmancuso (approved):

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
```diff
@@ -508,17 +508,28 @@ private static int countCaseTokens(DetailAST ast) {
...
-     * @return number of case constant (EXPR) tokens.
+     * @return number of case constant tokens.
      */
     private static int countCaseConstants(DetailAST ast) {
         final AtomicInteger counter = new AtomicInteger();
         final DetailAST literalCase = ast.getFirstChild();
 
-        TokenUtil.forEachChild(literalCase,
-            TokenTypes.EXPR, node -> counter.getAndIncrement());
+        final int[] caseLabelTokens = {
```
rnveach, line 520: move to a constant field.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/npathcomplexity/InputNPathComplexityPatternMatchingForSwitch.java
```diff
@@ -0,0 +1,105 @@
...
+    void testGuardsInRule(Object o) {
+        switch (o) {
+            case Integer i when i > 0 -> {}
+            case String s when s.length() > 0 -> {}
+            default -> {}
+        }
+    }
+
+    // violation below, 'NPath Complexity is 3 (max allowed is 1)'
+    void testGuardsInStatement(Object o) {
```
rnveach, line 59: I think there may be more to this. I rewrote the code as it I thought it would appear without `when` and got a different number.

````
$ cat TestClass.java
public class TestClass {
    void testGuardsInStatement(Object o) {
        switch (o) {
            case Integer i when i > 0 : {} break;
            case String s when s.length() > 0 : {}
            default : {}
        }
    }

    void testGuardsInStatement(Object o) {
        switch (o) {
            case Integer i : { if (i > 0) { } } break;
            case String s : { if (s.length() > 0) { } }
            default : {}
        }
    }
}

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">

    <module name="TreeWalker">
    <module name="NPathComplexity">
    </module>
    </module>
</module>

$ java -jar checkstyle-10.17.0-all.jar -c TestConfig.xml TestClass.java
Starting audit...
[WARN] TestClass.java:2:5: NPath Complexity is 3 (max allowed is 0). [NPathComplexity]
[WARN] TestClass.java:10:5: NPath Complexity is 5 (max allowed is 0). [NPathComplexity]
Audit done.
````

**Review by rnveach — CHANGES_REQUESTED** [review]

File: src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/npathcomplexity/InputNPathComplexityPatternMatchingForSwitch.java (hunk 3 above)
Comment (rnveach) [reply src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/npathcomplexity/InputNPathComplexityPatternMatchingForSwitch.java:59]: Installed a nightly (checkstyle-nightly-2024-07-15-all.jar) just to make sure and there was no changes to counts.

https://checkstyle.org/checks/metrics/npathcomplexity.html#Description
According to documentation, ifs and switches are counted.
> `if ([expr]) { [if-range] } 	NP(if-range) + 1 + NP(expr)`
> `switch ([expr]) { case : [case-range] default: [default-range] } 	S(i=1:i=n)NP(case-range[i]) + NP(default-range) + NP(expr)`

I think we should update documentation somehow to showcase when.

Comment (mahfouz72) [reply src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/npathcomplexity/InputNPathComplexityPatternMatchingForSwitch.java:59]: https://github.com/checkstyle/checkstyle/issues/15445

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java (hunk 2 above)
Comment (mahfouz72) [reply src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java]: done

**Comment by rnveach** [comment]
@mahfouz72 I am not following the fix from the issue.

Which count was correct in issue, and what was specifically leading to the counting difference? Can you point to tests which show we need every item in `CASE_LABEL_TOKENS`?

**Comment by mahfouz72** [comment]
In the issue, the `:` count is the correct one.

The [`CASE_GROUP` ( `:` )](https://github.com/checkstyle/checkstyle/blob/8531fec143058ddd92fab14839f2e9389ac6cf5c/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java#L244-L246) syntax is correct [because we always count all `LITERAL_CASE` under the `CASE_GROUP` ](https://github.com/checkstyle/checkstyle/blob/8531fec143058ddd92fab14839f2e9389ac6cf5c/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java#L499-L507) So this will always be correct regardless what is the case label.

The [`SWITCH_RULE` (`->`)](https://github.com/checkstyle/checkstyle/blob/8531fec143058ddd92fab14839f2e9389ac6cf5c/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java#L249-L252) was wrong [becuase for each switch rule it gets the `LITERAL_CASE` and counts all its children if they are `EXPR` only](https://github.com/checkstyle/checkstyle/blob/8531fec143058ddd92fab14839f2e9389ac6cf5c/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java#L516-L523) and this is not always true because now we can have a case label that is `PATTERN_VARIABLE_DEF`, `RECORD_PATTERN_DEF`, or `PATTERN_DEF` (when syntax)

---

```Java
switch (o) {
            case Integer _, R(int _) -> {}
            case String s when  s.isEmpty() -> {}
            default ->  { }
  }
```

```
         |       |   |--SWITCH_RULE -> SWITCH_RULE [4:12]
        |       |   |   |--LITERAL_CASE -> case [4:12]
        |       |   |   |   |--PATTERN_VARIABLE_DEF -> PATTERN_VARIABLE_DEF [4:17]  < --- should be counted and not EXPR
        |       |   |   |   |   |--MODIFIERS -> MODIFIERS [4:17]
        |       |   |   |   |   |--TYPE -> TYPE [4:17]
        |       |   |   |   |   |   `--IDENT -> Integer [4:17]
        |       |   |   |   |   `--IDENT -> _ [4:25]
        |       |   |   |   |--COMMA -> , [4:26]
        |       |   |   |   `--RECORD_PATTERN_DEF -> RECORD_PATTERN_DEF [4:28] < --- should be counted and not EXPR
                                  .........

        |       |   |   |       `--RPAREN -> ) [4:35]
        |       |   |   |--LAMBDA -> -> [4:37]
        |       |   |   `--SLIST -> { [4:40]
        |       |   |       `--RCURLY -> } [4:41]
        |       |   |--SWITCH_RULE -> SWITCH_RULE [5:12]
        |       |   |   |--LITERAL_CASE -> case [5:12]
        |       |   |   |   `--PATTERN_DEF -> PATTERN_DEF [5:26]       < --- should be counted and not EXPR
        |       |   |   |       `--LITERAL_WHEN -> when [5:26]
        |       |   |   |           |--PATTERN_VARIABLE_DEF -> PATTERN_VARIABLE_DEF [5:17]
        |       |   |   |           |   |--MODIFIERS -> MODIFIERS [5:17]
        |       |   |   |           |   |--TYPE -> TYPE [5:17]
                                          ....
        |       |   |   |--LAMBDA -> -> [5:44]
        |       |   |   `--SLIST -> { [5:47]
        |       |   |       `--RCURLY -> } [5:48]
```

---

> Can you point to tests which show we need every item in CASE_LABEL_TOKENS?

The new unit test and input file in this PR should show that we need all items in `CASE_LABEL_TOKENS`. because case label can be:
- pattern variable def
- record pattern
- guarded case label (`PATTERN_DEF`)
- norma expr. like `case 1, 2, 3 -> {}`

rnveach:
> In the issue, the : count is the correct one.

Please add that to issue.

> [counts all its children if they are EXPR only](https://github.com/checkstyle/checkstyle/blob/8531fec143058ddd92fab14839f2e9389ac6cf5c/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java#L516-L523)

1)
When is there more than 1 expression?
2) 
Why are we counting expressions?
https://checkstyle.org/checks/metrics/npathcomplexity.html
Says an expression is 0 unless there is `&&` or `||`.

From our discussion in discord, shouldn't we be counting cases here only?

> The new unit test and input file in this PR should show that we need all items in CASE_LABEL_TOKENS

Can you point these 2 out as I am confused why patterns would be separate like this.
> pattern variable def
> guarded case label (PATTERN_DEF)

rnveach (approved):

PR #15448, opened 2024-08-05
Issue #15040: Add support for yield in SimplifyBooleanReturnCheck

### LINKED ISSUE #15040: Add Check Support for Java 21 Pattern Matching for Switch Syntax: SimplifyBooleanReturn

I have read check documentation:https://checkstyle.org/checks/coding/simplifybooleanreturn.html

```
record ColoredPoint(boolean p, int x, String c) { }
record Rectangle(boolean upperLeft, boolean lowerRight) { }
public class Test {

    boolean test1(Object o) {
       if (o instanceof ColoredPoint(boolean x, _, _)) {
            if (x) return true;   // violation
            else return false;
        } else {
            return false;
        }
    }
    boolean test2(Object obj) {
        return switch (obj) {
            case ColoredPoint(boolean x,_,_) -> {
                if (x) yield true;       // expected violation , can be yield x
                else yield false;
            }
            default -> false;
        };
    }
}
PS D:\CS\test> cat config.xml                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="SimplifyBooleanReturn">

        </module>
    </module>
</module>
Starting audit...
[ERROR] D:\CS\test\src\Test.java:7:9: Conditional logic can be removed. [SimplifyBooleanReturn]
Audit done.
PS D:\CS\test> 
```

I am expecting a violation on unnecessary conditional logic that can be simplified inside switch expressions. each case block is an execution path for a return. So this condition falls under a Boolean return

mahfouz72:
closes #15040

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/SimplifyBooleanReturnCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/SimplifyBooleanReturnCheck.java
@@ -27,7 +27,7 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
 
 /**
  * <p>
- * Checks for over-complicated boolean return statements.
+ * Checks for over-complicated boolean return or yield statements.
  * For example the following code
  * </p>
  * <pre>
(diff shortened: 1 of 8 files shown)


mahfouz72:
Github, generate report for SimplifyBooleanReturn/Example1

mahfouz72:
github, generate site

mahfouz72:
Github, generate report for SimplifyBooleanReturn/Example1

romani:
In report there is no diff, mostly because code bases that we use don't use modern syntax.

Please search in GitHub java project where yield is in use.

mahfouz72:
Another reason is that we are looking at a very specific code which is :

```Java
if (xxx) {
  yield true;
}
else {
 yield false;
}
```

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/simplifybooleanreturn/InputSimplifyBooleanReturnWithYield.java
```diff
@@ -0,0 +1,46 @@
+/*
+SimplifyBooleanReturn
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.coding.simplifybooleanreturn;
+
+public class InputSimplifyBooleanReturnWithYield {
```
nrmancuso, line 9: Please add inputs with case labels (`:`) and also equivalent switches (where possible) with `return`.

nrmancuso (changes requested):
It is unlikely that we will find much usage like this in project files.

Item:

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/simplifybooleanreturn/InputSimplifyBooleanReturnWithYield.java
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/SimplifyBooleanReturnCheck.java
```diff
@@ -140,20 +140,21 @@ private static boolean canReturnOnlyBooleanLiteral(DetailAST ast) {
...
      * <pre>
-     * return true/false;
+     * return/yield true/false;
      * </pre>
      *
      * @param ast the syntax tree to check
-     * @return if ast is a return statement with a boolean literal.
+     * @return if ast is a return or a yield statement with a boolean literal.
      */
     private static boolean isBooleanLiteralReturnStatement(DetailAST ast) {
```
rnveach, line 153: I find it odd you updated the javadocs to make it clear it is a return or yield, but then you leave the method as `ReturnStatement`.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/SimplifyBooleanReturnCheck.java
mahfouz72: done

rnveach (approved):

PR #15450, opened 2024-08-05
Issue #15062: Add support for pattern variables in SuppressWarningCheck

### LINKED ISSUE #15062: Add Check Support for Java 21 Unnamed Variables & Patterns Syntax: SuppressWarnings
child https://github.com/checkstyle/checkstyle/issues/14942
I have read check documentation: https://checkstyle.org/checks/annotation/suppresswarnings.html

---

> Allows to specify what warnings that @SuppressWarnings is not allowed to suppress. You can also specify a list of TokenTypes that the configured warning(s) cannot be suppressed on.

I can use @SuppressWarnings on pattern variables. should the check have `PATTERN_VARIABLE_DEF` in the tokens

---
```
record ColoredPoint(int p, int x, int c) { }

public class Test {
    void test(Object obj) {
        if (obj instanceof ColoredPoint( @SuppressWarnings("") int _, _,_)) { } // should this be violation ?
        if (obj instanceof @SuppressWarnings("") Rectangle r) { } // should this be violation ?
        @SuppressWarnings("") int _ = 0; // violation
    }
}

PS D:\CS\test> cat config.xml                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="SuppressWarnings">
        </module>
    </module>
</module>
Starting audit...
[ERROR] D:\CS\test\src\Test.java:8:27: The warning '' cannot be suppressed at this location. [SuppressWarnings]
Audit done.
PS D:\CS\test> 

```

mahfouz72:
closes #15062 

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/3eefa151f23d9c0754bcb6ed1c2c2ef73f45db9b/check.xml
Diff Regression patch config: https://gist.githubusercontent.com/mahfouz72/33967bf86e4a29cafa429771b0e95808/raw/8c95593d4d6db416055445b4e4dc43104c375aa9/check_patch.xml

### DIFF
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/annotation/suppresswarnings/InputSuppressWarningsPatternVariables.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/annotation/suppresswarnings/InputSuppressWarningsPatternVariables.java
@@ -1,0 +1,40 @@
+/*
+SuppressWarnings
+format = ^unchecked$|^unused$|^$
+tokens = (default)CLASS_DEF, INTERFACE_DEF, ENUM_DEF, ANNOTATION_DEF, ANNOTATION_FIELD_DEF, \
+         ENUM_CONSTANT_DEF, PARAMETER_DEF, VARIABLE_DEF, METHOD_DEF, CTOR_DEF, \
+         COMPACT_CTOR_DEF, RECORD_DEF, PATTERN_VARIABLE_DEF
+
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.annotation.suppresswarnings;
+
+public class InputSuppressWarningsPatternVariables {
+
+    void test(Object o) {
+        // violation below, 'The warning '' cannot be suppressed at this location.'
+        if (o instanceof ColoredPoint(@SuppressWarnings("")int x, int y, int z)) { }
+
+        // violation below, 'The warning 'unused' cannot be suppressed at this location.'
+        if (o instanceof ColoredPoint(_, _,@SuppressWarnings("unused") int _)) { }
+
+        // violation below, 'The warning 'unchecked' cannot be suppressed at this location.'
+        if (o instanceof @SuppressWarnings("unchecked") ColoredPoint _) { }
+
+        // violation below, 'The warning '' cannot be suppressed at this location.'
+        @SuppressWarnings("")
+        int _ = sideEffect();
+
+        if (o instanceof @SuppressWarnings("a unused") ColoredPoint _) { }
+
+        @SuppressWarnings("foo")
+        int _ = sideEffect();
+    }
+
+    int sideEffect() {
+        return 0;
+    }
+    record ColoredPoint(int x, int y, int z) { }
+}
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/SuppressWarningsCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/SuppressWarningsCheck.java
@@ -102,7 +102,9 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#COMPACT_CTOR_DEF">
  * COMPACT_CTOR_DEF</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#RECORD_DEF">
- * RECORD_DEF</a>.
+ * RECORD_DEF</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#PATTERN_VARIABLE_DEF">
+ * PATTERN_VARIABLE_DEF</a>.
  * </li>
  * </ul>
  * <p>
@@ -176,6 +178,7 @@ public class SuppressWarningsCheck extends AbstractCheck {
             TokenTypes.CTOR_DEF,
             TokenTypes.COMPACT_CTOR_DEF,
             TokenTypes.RECORD_DEF,
+            TokenTypes.PATTERN_VARIABLE_DEF,
         };
     }
 
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/annotation/SuppressWarningsCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/annotation/SuppressWarningsCheck.xml
@@ -45,7 +45,7 @@
                <description>Specify the RegExp to match against warnings. Any warning
  being suppressed matching this pattern will be flagged.</description>
             </property>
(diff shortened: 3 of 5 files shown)


mahfouz72:
github, generate report

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/annotation/suppresswarnings/InputSuppressWarningsPatternVariables.java
```diff
@@ -0,0 +1,31 @@
...
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.annotation.suppresswarnings;
+
+public class InputSuppressWarningsPatternVariables {
+
+    void test(Object o) {
+        if (o instanceof ColoredPoint(@SuppressWarnings("")int x, int y, int z)) { } // violation
```
romani, line 17: @rnveach, @nrmancuso , when you see such lines without violation message, try to think a minute about "Should we actually demand message to be always specified?" (even Check has only one message)

I more and more inclined to think that it will be better to have violation message at all cases to let reader not rely on memory what it should be. In this case I completely forgot what this Check do, it required me jump  out of this page to get to know what this violation mean.

Preferable:
```
 // violation 'The warning '''' cannot be suppressed at this location.'
if (o instanceof ColoredPoint(@SuppressWarnings("")int x, int y, int z)) { }
```

romani: I see this as closed https://github.com/checkstyle/checkstyle/issues/11214 , and no issues on migration off `// violation` even for Checks with single message.
If you agree, please let me know, we will create issue it will be `good first/second issue`

mahfouz72: @nrmancuso has told me before to always specify the message so I think he agrees but I sometimes forgot to do so (There is no test that enforcs that). I will update this PR to specify the message. 
Ideally we should enfore the presence of the message even if the check has only one message and yes opening a good first issue to migrate all old `// violation`  to have message is a good idea

romani: Looks like I lost track that we finished migration to BDD, and now we can do this update.
Before it was completely ok to put violation message without message, as there were bunch missing useful variants violation marker were not implemented. But now we are good to go forward to use good practice format.

mahfouz72: We can enforce this in the parser itself or a simple regexp. There will be huge suppression lists

nrmancuso: @romani please create an issue on this so that we can discuss there

mahfouz72: https://github.com/checkstyle/checkstyle/pull/15452

romani: Follow up issue https://github.com/checkstyle/checkstyle/issues/15456

nrmancuso (approved):

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/annotation/suppresswarnings/InputSuppressWarningsPatternVariables.java
```diff
@@ -0,0 +1,35 @@
+/*
+SuppressWarnings
+format = ^unchecked$*|^unused$*|.*
+tokens = (default)CLASS_DEF, INTERFACE_DEF, ENUM_DEF, ANNOTATION_DEF, ANNOTATION_FIELD_DEF, \
+         ENUM_CONSTANT_DEF, PARAMETER_DEF, VARIABLE_DEF, METHOD_DEF, CTOR_DEF, \
+         COMPACT_CTOR_DEF, RECORD_DEF
```
rnveach, line 6: I don't see `PATTERN_VARIABLE_DEF` listed here.

Also, what about `PATTERN_DEF`? Should we be concerned about it too? A separate issue?

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/annotation/suppresswarnings/InputSuppressWarningsPatternVariables.java
```diff
@@ -0,0 +1,35 @@
+/*
+SuppressWarnings
+format = ^unchecked$*|^unused$*|.*
```
rnveach, line 3: I am not seeing anything that goes against this format. Can you just add 1 example?

Also, isn't `$*` basically invalid? Shouldn't this format be `^(unchecked|^unused|).*$`?

Seems like the last one will match everything, so you can't go against the format at all. I would still like to see one counter example to ensure this is processing correctly.

rnveach (changes requested):

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/annotation/suppresswarnings/InputSuppressWarningsPatternVariables.java
mahfouz72: > I don't see PATTERN_VARIABLE_DEF listed here.

Added. but actually [it was implicitly here because we have `(default)`](https://github.com/checkstyle/checkstyle/blob/1f939bc290344df1d3c9f5ebc5731f04fc750942/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L511-L517)

> what about PATTERN_DEF? Should we be concerned about it too?

Nope. we don't need this the annotation is on `PATTERN_VARIABLE_DEF`

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/annotation/suppresswarnings/InputSuppressWarningsPatternVariables.java
mahfouz72: Yea sorry this regex will match everything I updated it and added examples that goes against it

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/annotation/suppresswarnings/InputSuppressWarningsPatternVariables.java
```diff
@@ -0,0 +1,40 @@
+/*
+SuppressWarnings
+format = ^unchecked$|^unused|^$
```
rnveach, line 3: `^unused` => `^unused$` ?

rnveach (changes requested):
should be last

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/annotation/suppresswarnings/InputSuppressWarningsPatternVariables.java
mahfouz72: done

rnveach (approved):

Issue #15457, opened 2024-08-06
Fix false website failure on absolute links for javadocs to our own website


rnveach:
PR validation:
https://github.com/checkstyle/checkstyle/blob/8d90858d9a5e44914e8696f6ab098409abd280ec/.circleci/config.yml#L263-L266

Master failure:
https://github.com/checkstyle/checkstyle/actions/runs/10268868835/job/28412946322#step:4:2435
> `<a class="externalLink" href="https://checkstyle.org/styleguides/google-java-style-20220203/javaguide.html#s4.8.6.1-block-comment-style">https://checkstyle.org/styleguides/google-java-style-20220203/javaguide.html#s4.8.6.1-block-comment-style</a>: 404 Not Found`

Link to file with failure: https://github.com/checkstyle/checkstyle/blob/93d7140c4f38583f281f56df55603057cc85e8a2/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/CommentsIndentationCheck.java#L38

Failure only occurs in master and not in PRs.

Issue is we don't know if the link is really valid or not until after we do a release and update our website since its an absolute link.

Javadocs need to be absolute link since we would want to use it on our locals, and on website javadoc is in some weird directory. If the link is in our xdoc, we should be using relative links as it is moved to versioning directory eventually, and absolute links in versioning break versioning directory.

Javadoc link is at: https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/checks/indentation/CommentsIndentationCheck.html

We need to find some way to ignore our website in javadoc, and only validate it once we have released new version of website in release process. It would be better if we found someway to validate this before release, otherwise GH will always be behind validation. We will be fixing errors in X.Y release in X.(Y+1) snaspshot.

Additional reasoning for this issue at https://github.com/checkstyle/checkstyle/pull/15459#issuecomment-2278094799

rnveach:
I think to fix issue to actually validate these links, we need to do a replace all (uncommitted) during validation.

We replace `https://checkstyle.org` and turn it into an relative link (`../`) that this check will find, and verify if it actually valid or not. We can trim this replace all to the directory https://github.com/checkstyle/checkstyle/tree/master/src/main/java , and that way we will know we are only changing javadocs and not anything else. This will also allow us to validate link in SNAPSHOT before release, so we don't need to actually update real website first.

I am not sure what we can transform the link to for validation to pass. I have seen this tool validate relative links before, so I think this won't pose any issues.

[...]

PR #15459, opened 2024-08-06
Issue #15457: validate our website javadoc links

rnveach:
Issue #15457

### DIFF
--- a/.ci/run-link-check-plugin.sh
+++ b/.ci/run-link-check-plugin.sh
@@ -29,8 +29,8 @@ grep -rl "https://checkstyle.org" src/main/java | while read -r file; do
     sed -i "s|https://checkstyle.org|$replacement|g" "$file"
 done
 
-mvn -e --no-transfer-progress clean site -Dcheckstyle.ant.skip=true -DskipTests -DskipITs \
-   -Dpmd.skip=true -Dspotbugs.skip=true -Djacoco.skip=true -Dcheckstyle.skip=true
+mvn -e --no-transfer-progress -Pno-validations clean site -Dlinkcheck.skip=false \
+   -Dmaven.javadoc.skip=false
 
 # revert all modified javadoc files for post validation
 git checkout HEAD -- src/main/java
--- a/.ci/run-link-check-plugin.sh
+++ b/.ci/run-link-check-plugin.sh
@@ -7,8 +7,34 @@ pwd
 uname -a
 mvn --version
 curl --fail-with-body -I https://sourceforge.net/projects/checkstyle/
+
+# Function to calculate the replacement string based on the file depth
+calculate_replacement() {
+    local file_path="$1"
+    local relative_path="${file_path#src/main/java/}"
+    local depth=$(echo "$relative_path" | grep -o "/" | wc -l)
+    local replacement=""
+    for ((i=0; i<depth; i++)); do
+        if ((i > 0)); then
+            replacement="/$replacement"
+        fi
+        replacement="..$replacement"
+    done
+    echo "../$replacement"
+}
+
+# convert all javadoc website links to relative links
+grep -rl "https://checkstyle.org" src/main/java | while read -r file; do
+    replacement=$(calculate_replacement "$file")
+    sed -i "s|https://checkstyle.org|$replacement|g" "$file"
+done
+
 mvn -e --no-transfer-progress clean site -Dcheckstyle.ant.skip=true -DskipTests -DskipITs \
    -Dpmd.skip=true -Dspotbugs.skip=true -Djacoco.skip=true -Dcheckstyle.skip=true
+
+# revert all modified javadoc files for post validation
+git checkout HEAD -- src/main/java
+
 mkdir -p .ci-temp
 
 OPTION=$1


rnveach:
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/27191/workflows/9d488a74-9a47-4bd5-845e-518afa737c5e/jobs/660275
> `<a href="apidocs/com/puppycrawl/tools/checkstyle/checks/styleguides/google-java-style-20220203/javaguide.html#s4.8.6.1-block-comment-style">../styleguides/google-java-style-20220203/javaguide.html#s4.8.6.1-block-comment-style</a>: doesn't exist.`

It appears to be working. Directory needs to be corrected a bit more. This can't tell the difference between the href and the text of the href. I am not sure if we always ensure we write `href="https" on a single line.

rnveach:
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/27192/workflows/beb9b029-8adc-426e-b7b5-d7326fc061b6/jobs/660302
I would need to look at what this one is about.

But I do see an issue with this I should have seen coming. We need replace a dynamic number of sub-directories and replace with `../` as many times as it takes to get out of javadoc folder. It won't always be a fixed count.

rnveach:
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/27199/workflows/c6f1d1a9-1319-41aa-b14a-8426e000f10b/jobs/660597
linkcheck is passing.

I can confirm locally that the false positive link is gone.

rnveach:
@romani @nrmancuso Are we ok with this resolving https://github.com/checkstyle/checkstyle/issues/15457 ? If so, then I will fix the PR up. Please note I am weak in bash and had to use ChatGPT to help with the dynamic directory depth replacement.

.ci/run-link-check-plugin.sh
```diff
@@ -7,8 +7,34 @@ pwd
...
+}
+
+# convert all javadoc website links to relative links
+grep -rl "https://checkstyle.org" src/main/java | while read -r file; do
+    replacement=$(calculate_replacement "$file")
+    sed -i "s|https://checkstyle.org|$replacement|g" "$file"
+done
+
+mvn -e --no-transfer-progress -Pno-validations clean site -Dlinkcheck.skip=false \
+   -Dmaven.javadoc.skip=false
```
rnveach, line 33: This command was replaced using `no-validations` so we don't have to manually disable checks, especially if one is added, and we can clearly see what is being enabled.

romani:
>I am not sure if we always ensure we write `href="https" on a single line.

we almost always do this, links are not wrap-able.

romani:
>On my local, error count went from 1362 to 1360.

My worries it that we do very complicated code changes just to hit 2 violations out of  1362 .
This code might give us more problems on edgecases  in future, and benefit of it not significant.
And workaround for this problems with 2 links is too make release, that is temporal problem of summer activity.

.ci/run-link-check-plugin.sh
romani: I think we can make sepatate PR for this update only and merge it quickly.

.ci/run-link-check-plugin.sh
```diff
@@ -7,8 +7,34 @@ pwd
...
+
+# revert all modified javadoc files for post validation
+git checkout HEAD -- src/main/java
```
romani, line 36: this is a concern, I usually always treated CI tasks or any validation task as commands that ready to be executed at any time in root of repo.
this update will revert all user uncommited changes, not good.
if we agreed to move forward with this links changes we need to copy checkouted repo to .ci-temp and do all code changes in that folder, and this checkout might not be required at all.

romani (changes requested):
@nrmancuso, please share your thoughts on issue in general.

for post approval time:

rnveach:
> My worries it that we do very complicated code changes just to hit 2 violations out of 1362 .

It is misleading to say this is to hit 2 instances.

This is to hit any changes in master which has not been released. This **does** validate other links properly, by pointing them to master instead of the real server, but those will work with real server **ONLY WHEN** they match master currently.

This release it is only these 2 false positives. If we added new grammar token, it would impact those too, as all our tokens point to javadoc URL. Most likely most releases would not be impacted by this.

This also provides us a green CI in master so it does not confuse people with a false positive failure.

If we do not want this fix, then the only other option is to ignore all links directly to our website URL, and we will validate them after the release and fix them then. This means we will always be fixing failures of the previous release in the next release's snapshot. This is basically what we are doing now by ignoring master CI failure, unless we rigorously recheck the failure every commit to very it is false and not a positive.

.ci/run-link-check-plugin.sh
nrmancuso: Yeah, this is good

nrmancuso (commented):
Maybe this seems weird, but I will propose another approach:
1. Run linkcheck once
2. Modify the linkcheck report instead of the javadocs
3. Run linkcheck again without `clean` (or use this as a cached file)

I am not sure if this would work, but it is probably easier/better than modifying java files IMO

romani:
I like idea of updates in generated website. It will be safer and we can hack anything we want by any craziest bash/.... code.

PR #15466, opened 2024-08-08
Pull #15466: grab setter types even if field exists for override

rnveach:
Assist for https://github.com/checkstyle/checkstyle/pull/15376#discussion_r1706279547 ,

`PROPERTIES_ALLOWED_GET_TYPES_FROM_METHOD` allows us to use the setter's type ONLY if the field cannot be found. If we make the field invisible to use that override, we lose the default value of the field.

This updates PROPERTIES_ALLOWED_GET_TYPES_FROM_METHOD logic so it uses the setter even if the field exists. This override is only used by 4 other checks, so it should be minimal impact.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
@@ -1364,21 +1364,21 @@ public class XdocsPagesTest {
             Field field, String propertyName) throws Exception {
         Class<?> result = null;
 
-        if (field != null) {
+        if (PROPERTIES_ALLOWED_GET_TYPES_FROM_METHOD.contains(sectionName + "." + propertyName)) {
+            final PropertyDescriptor descriptor = PropertyUtils.getPropertyDescriptor(instance,
+                    propertyName);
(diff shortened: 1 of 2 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -1364,21 +1364,21 @@ private static Class<?> getFieldClass(String fileName, String sectionName, Objec
...
-                    .that(PROPERTIES_ALLOWED_GET_TYPES_FROM_METHOD)
-                    .contains(sectionName + "." + propertyName);
-
-            final PropertyDescriptor descriptor = PropertyUtils.getPropertyDescriptor(instance,
-                    propertyName);
-            result = descriptor.getPropertyType();
+                    .fail();
         }
-        if (result == List.class || result == Set.class) {
+        if (field != null && (result == List.class || result == Set.class)) {
```
rnveach, line 1381: Eclipse did complain about a possible NPE on `field`. Not sure how possible this is, but this update now also matches `SiteUtil`'s code.

nrmancuso (approved):
Awesome, thanks a lot :)

PR #15358, opened 2024-07-24
Issue #15058: New Check UnusedCatchParameterShouldBeUnnamed

mahfouz72:
closes #15058 

New module config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/3290828c8bb75105d18940f38a97265806b6a8bb/check.xml

PR #15480, opened 2024-08-09
Issue #15059: New Check UnusedLambdaParameterShouldBeUnnamed

### LINKED ISSUE #15059: Add Check Support for Java 21 Unnamed Variables & Patterns Syntax: New Check UnusedLambdaParameterShouldBeUnnamed
child of #14942 
violate the non-use of the lambda parameters. They should be unnamed.

**Note:** The lambda parameters in the check are the parameters of [lambda expressions](https://docs.oracle.com/javase/specs/jls/se22/html/jls-15.html#jls-15.27) only. The unused pattern variables in the case label of the switch rule are not a target for this check.

---

**Config**

```XML
$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
    <module name="TreeWalker">
        <module name="UnusedLambdaParameterShouldBeUnnamed"/>
    </module>
</module>
```

---

**Example**

```Java
public class Test {
    void test(Object obj) {
        List<String> list = List.of("a", "b", "c");
        List<String> Xss = list.stream().map(character -> "x").toList(); // violation
        List<String> Xs = list.stream().map(_-> "x").toList();
    }
}
```

---

**Expected Output**

```PowerShell
Starting audit...
[ERROR] D:\Test.java:4:45:Unused  lambda parameter should be unnamed [UnusedLambdaParameterShouldBeUnnamed]
Audit done.
```

mahfouz72:
closes #15059:

New module config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/3e297236fff03643bdcffa607b54b528b3438ece/check.xml


mahfouz72:
github, generate site

romani:
>Unnamed lambda parameters will not compile on java versions between 9 and 21

Please share in comment, violations on our code, we should recheck if all of them are valid.

mahfouz72:
```
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\it\java\org\checkstyle\base\AbstractItModuleTestSupport.java:519:67: Unused lambda parameter 'key' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\TreeWalker.java:305:59: Unused lambda parameter 'empty' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\TreeWalker.java:316:60: Unused lambda parameter 'empty' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\api\FileContents.java:155:17: Unused lambda parameter 'empty' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\UniquePropertiesCheck.java:185:59: Unused lambda parameter 'empty' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\blocks\RightCurlyCheck.java:491:29: Unused lambda parameter 'child' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\coding\ConstructorsDeclarationGroupingCheck.java:126:30: Unused lambda parameter 'first' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\coding\EqualsHashCodeCheck.java:174:39: Unused lambda parameter 'key' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\coding\MultipleStringLiteralsCheck.java:210:58: Unused lambda parameter 'key' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\javadoc\AbstractJavadocCheck.java:326:87: Unused lambda parameter 'key' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\sizes\RecordComponentNumberCheck.java:159:13: Unused lambda parameter 'node' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\filters\SuppressWithPlainTextCommentFilter.java:288:22: Unused lambda parameter 'first' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\gui\ListToTreeSelectionModelWrapper.java:49:58: Unused lambda parameter 'event' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\gui\MainFrame.java:140:41: Unused lambda parameter 'event' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\AbstractModuleTestSupport.java:606:67: Unused lambda parameter 'key' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\ConfigurationLoaderTest.java:714:52: Unused lambda parameter 'context' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\TreeWalkerTest.java:136:72: Unused lambda parameter 'mock' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\TreeWalkerTest.java:136:78: Unused lambda parameter 'context' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\checks\header\HeaderCheckTest.java:305:48: Unused lambda parameter 'context' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\gui\MainFrameTest.java:128:44: Unused lambda parameter 'context' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\gui\MainFrameTest.java:147:44: Unused lambda parameter 'context' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\internal\AllChecksTest.java:590:59: Unused lambda parameter 'key' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\internal\AllTestsTest.java:110:75: Unused lambda parameter 'key' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\internal\AllTestsTest.java:129:75: Unused lambda parameter 'key' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\internal\XdocsJavaDocsTest.java:638:74: Unused lambda parameter 'unused' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\internal\XdocsJavaDocsTest.java:639:63: Unused lambda parameter 'unused' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
[ERROR] [checkstyle] [ERROR] D:\CS\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\internal\XdocsJavaDocsTest.java:640:68: Unused lambda parameter 'unused' should be unnamed. [UnusedLambdaParameterShouldBeUnnamed]
```
I checked most of them, and they are valid

romani:
>Unnamed lambda parameters will not compile on java versions between 9 and 21

Can we simply state `Unnamed lambda parameters will not compile on java versions before 21` ?

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_ru.properties
```diff
@@ -65,6 +65,7 @@ unnecessary.paren.return=Лишние скобки вокруг возвраща
 unnecessary.paren.string=Лишние скобки вокруг строки {0}.
 unnecessary.semicolon=Лишняя точка с запятой.
 unused.catch.parameter=Неиспользуемый параметр catch ''{0}'' должен быть безымянным.
+unused.lambda.parameter=Неиспользуемый лямбда-параметр ''{0}'' не должен быть назван.
 unused.local.var=Неиспользуемая локальная переменная ''{0}''.
```
romani, line 69: `Неиспользуемый лямбда-параметр ''{0}'' должен быть  безымянным`

romani (changes requested):
Item

nrmancuso:
> > Unnamed lambda parameters will not compile on java versions between 9 and 21
> 
> Can we simply state `Unnamed lambda parameters will not compile on java versions before 21` ?

No, this is not accurate, though they were not called “unnamed” until now. You could use an underscore as a lambda parameter name in Java 8.

romani:
we are on 11, so comments in our code should not be about past old jdk.
I see this error:
image
We even do not have Inputs that are compiled on old jdk. 

> You could use an underscore as a lambda parameter name in Java 8

lets put this nuance to xdoc of Check, so this Check potentially can be used on sources that still on jdk8.

mahfouz72:
You can't use this check on java 8 because unnamed variables concept was not as it is now(21). underscore variable in java 8 can be read from or write to so there will be no benefits if we changed the variable name from 'a' to '_' both can be used as normal identifier in java 8 and below. the rationale of the check is to declare the unused parameter as unnamed because if we are java 21+ we are sure that this unnamed variable can't be used and we want to communicate this explicitly to readers

romani:
> underscore variable in java 8 can be read from or write to so there will be no benefits if we changed the variable name from 'a' to '_' both can be used as normal identifier in java 8 and below.

please rephase, I do not understand this sentence.

------

What we have in comment of config and in web site is different and I just try to point that better to align wording.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/unusedlambdaparametershouldbeunnamed/InputUnusedLambdaParameterShouldBeUnnamedNested.java
nrmancuso, line 1: Please add more examples with typed lambda parameters

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.xml
```diff
@@ -0,0 +1,35 @@
...
+     &lt;/li&gt;
+ &lt;/ul&gt;
+ &lt;p&gt;
+ See the &lt;a href="https://docs.oracle.com/en/java/javase/21/docs/specs/unnamed-jls.html"&gt;
+ Java Language Specification&lt;/a&gt; for more information about unnamed variables.
+ &lt;/p&gt;
+ &lt;p&gt;
+ &lt;b&gt;Attention&lt;/b&gt;: This check should be activated only on source code
+ that is compiled by jdk21 or higher;
+ unnamed lambda parameters came out as the first preview in Java 21.
```
nrmancuso, line 28:
```suggestion
 unnamed variables are available as a preview feature in Java 21, and became an official part of the language in Java 22.
```

nrmancuso (changes requested):
Looks good, few minor things:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
```diff
@@ -0,0 +1,284 @@
...
+                    .findFirst()
+                    .ifPresent(LambdaParameterDetails::registerAsUsed);
+        }
+    }
+
+    @Override
+    public void leaveToken(DetailAST ast) {
+        if (ast.getType() == TokenTypes.LAMBDA) {
+            // when leaving a given lambda, check every parameter associated with it
+            while (!lambdaParameters.isEmpty()
```
nrmancuso, line 151:
```suggestion
            while (lambdaParameters.peek() != null
```
This will make checker happy

nrmancuso (changes requested):
another:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
```diff
@@ -0,0 +1,284 @@
...
+    /**
+     * A key is pointing to the warning message text in "messages.properties"
+     * file.
+     */
+    public static final String MSG_UNUSED_LAMBDA_PARAMETER = "unused.lambda.parameter";
+
+    /**
+     * Invalid parents of the lambda parameter identifier.
+     */
+    private static final int[] INVALID_LAMBDA_PARAM_IDENT_PARENTS = {
```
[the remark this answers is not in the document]
mahfouz72: To ignore identifiers when they are method call or type etc..

examples: https://github.com/checkstyle/checkstyle/blob/ca17d061268c738ce72e24c441cea56bec0f2a94/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/unusedlambdaparametershouldbeunnamed/InputUnusedLambdaParameterShouldBeUnnamedMultipleParameters.java#L38-L50

line 40 is type, line 49 is method call so we should ignore this and not be considered as candidate for lambda parameter ident

same as: https://github.com/checkstyle/checkstyle/pull/15358#discussion_r1691996520

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_ru.properties
mahfouz72: done

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/coding/unusedlambdaparametershouldbeunnamed/InputUnusedLambdaParameterShouldBeUnnamedNested.java
mahfouz72: done

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.xml
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
mahfouz72: done

mahfouz72:
> please rephase, I do not understand this sentence.

I just was trying to say that this check can't be used for source code that is below Jdk 8. In Java 8 and earlier versions, the underscore `_` can be used as a valid identifier, just like any other variable name. Therefore, changing a variable name from any name to `_` doesn't provide any benefit, as both can be used in the same way

>  we should simply state that we need to enable it only after migration to 21. No need to mention all other nuances.

make sense, done
@nrmancuso  I am ok to make it below 21 in the comment on config. I think It is correct wording. "Unnamed" comes on 21 only. before Java 8 it was not called unnamed so yea it makes sense. better to keep it like this to align wording with xdocs as @romani suggested. do you agree with the current wording in docs and config comment?

romani:
i am good for your fixes on my comments.

I will review this PR at the most end. But I generally good with new Check.

mahfouz72:
github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
mahfouz72: but still checker fails I have no idea how is that possible. may be a false positive? are we ok to suppress?

nrmancuso:
```suggestion
                        && ast.equals(lambdaParameters.peek().enclosingLambda())) {
```
Would this make checker happy?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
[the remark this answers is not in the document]
nrmancuso: >same as: https://github.com/checkstyle/checkstyle/pull/15358#discussion_r1691996520

Yes, yet another ambiguously named collection. 

>line 40 is type, line 49 is method call so we should ignore this and not be considered as candidate for lambda parameter ident

>INVALID_LAMBDA_PARAM_IDENT_PARENTS

This doesn't add up. Why are we calling this LAMBDA_PARAM anything if the identifiers they are on are not lambda parameters?

Imagine you did not write this check, and need to fix some bug. It is not clear how these tokens are related, and what "invalid" means in this context. Instead of explicitly mentioning what a lambda parameter should never have as a parent, why not just call out exactly what CAN be a parent? This should be simple to figure out from our grammar/visitor.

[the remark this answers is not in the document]
mahfouz72: > Why are we calling this LAMBDA_PARAM anything if the identifiers they are on are not lambda parameters?

that's why we perfixed with "INVALID"

> Instead of explicitly mentioning what a lambda parameter should never have as a parent, why not just call out exactly what CAN be a parent?

I assumed that this would be a huge list so it is better to have the opposite (tokens that CAN NOT be parent)

what about adding this extra explanation in Javadoc:  `These are tokens that can not be parents for a lambda
 parameter identifier`
 
does it become less ambiguous, IYO?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
mahfouz72: No. very weird

```Java
 while (lambdaParameters.peek() != null
                        && ast.equals(lambdaParameters.peek().enclosingLambda)) { ... }
```
```XML
    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java</fileName>
    <specifier>dereference.of.nullable</specifier>
    <message>dereference of possibly-null reference lambdaParameters.peek()</message>
    <lineContent>&amp;&amp; ast.equals(lambdaParameters.peek().enclosingLambda)) {</lineContent>
  </checkerFrameworkError>
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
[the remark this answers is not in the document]
nrmancuso: This is better for sure.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
nrmancuso: I am good to suppress this, we should also open a bug report with checker

nrmancuso (approved):
Very good, thanks for making this an easy review :)

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
nrmancuso: Leaving this unresolved for the next reviewers to see.

mahfouz72: https://github.com/typetools/checker-framework/issues/6760

mahfouz72:
github, generate report

pom.xml
```diff
@@ -3288,6 +3288,7 @@
                 <param>com.puppycrawl.tools.checkstyle.checks.coding.SimplifyBooleanExpressionCheck*</param>
                 <param>com.puppycrawl.tools.checkstyle.checks.coding.SimplifyBooleanReturnCheck*</param>
                 <param>com.puppycrawl.tools.checkstyle.checks.coding.UnusedCatchParameterShouldBeUnnamedCheck*</param>
+                <param>com.puppycrawl.tools.checkstyle.checks.coding.UnusedLambdaParameterShouldBeUnnamedCheckParameterShouldBeUnnamedCheck*</param>
```
rnveach, line 3291: https://checkstyle.semaphoreci.com/jobs/b9e32e92-424f-4fc7-ac52-33020a92aaa7

You may want to recheck this.

rnveach (changes requested):

pom.xml

nrmancuso: @mahfouz72 we have a bunch of surviving mutants now

mahfouz72: They should be resolved now

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
```diff
@@ -0,0 +1,285 @@
...
+ * </ul>
+ * <p>
+ * See the <a href="https://docs.oracle.com/en/java/javase/21/docs/specs/unnamed-jls.html">
+ * Java Language Specification</a> for more information about unnamed variables.
+ * </p>
+ * <p>
+ * <b>Attention</b>: This check should be activated only on source code
+ * that is compiled by jdk21 or higher;
+ * unnamed variables are available as a preview feature in Java 21,
+ * and became an official part of the language in Java 22.
```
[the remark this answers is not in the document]
rnveach: Also can we have some consistency with Java versioning, or explain to me why they should be different. You say `jdk21` but then also say `Java 21`. Another PR I just reviewed had `Java21` (no space). The same person is writing all this, right? /s

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
```diff
@@ -0,0 +1,285 @@
...
+                    .findFirst()
+                    .ifPresent(LambdaParameterDetails::registerAsUsed);
+        }
+    }
+
+    @Override
+    public void leaveToken(DetailAST ast) {
+        // when leaving a given lambda, check every parameter associated with it
+        while (lambdaParameters.peek() != null
+                    && ast.equals(lambdaParameters.peek().enclosingLambda)) {
```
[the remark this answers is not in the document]
rnveach: You may not be able to do my previous comment because we don't push everything (please confirm anyways), but one of the conditions does multiple pushes and you only do a single pop. Isn't that an issue?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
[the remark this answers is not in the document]
mahfouz72: > If we really want to convey previews, I think we should make this a bit more specific.
This check should be activated only on source code that is compiled by jdk21 or higher; unnamed variables are available as a preview feature in Java 21, and became an official part of the language in Java 22.
=>
Unnamed variables are available as a preview feature in Java 21, and became an official part of the language in Java 22. This check should be activated only on source code which meets those requirements.


Done.

---
> Also can we have some consistency with Java versioning

Noted. 


>  The same person is writing all this, right? /s

most likely, Yes unless it is copy pasted for a maintainer old review.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
```diff
@@ -0,0 +1,285 @@
...
+ * Violation Message Keys:
+ * </p>
+ * <ul>
+ * <li>
+ * {@code unused.lambda.parameter}
+ * </li>
+ * </ul>
+ *
+ * @since 10.18.0
+ *
```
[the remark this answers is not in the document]
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
```diff
@@ -0,0 +1,285 @@
...
+        final boolean isLambdaParameterDeclaration =
+                identifierAst.getParent().getType() == TokenTypes.LAMBDA
+                    || identifierAst.getParent().getType() == TokenTypes.PARAMETER_DEF;
+
+        final boolean hasValidParentToken =
+                !TokenUtil.isOfType(identifierAst.getParent(), INVALID_LAMBDA_PARAM_IDENT_PARENTS);
+
+        final boolean isMethodInvocation = isMethodInvocation(identifierAst);
+
+        return !isLambdaParameterDeclaration && (hasValidParentToken || isMethodInvocation);
```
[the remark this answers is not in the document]
mahfouz72: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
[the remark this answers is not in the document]
mahfouz72: > The comment is misleading now, we aren't leaving only lambdas here. We are leaving IDENTs too, but the equals method is preventing that check.

Comment is removed. actually, it doesn't add much value.

> Why do we need to do the equals check if we have a lambda check instead? Lambdas are guaranteed to be the same order to leave as they came in.


Can you restate I think I don't understand the question well.  but basically, we have the equals check to check where to stop checking the parameters. suppose we have nested lambda parameters each has 3 parameters when we leave the inner one we want to pop and log the top 3 only so we continue to pop unless `ast` (the token we are leaving) is not equal to `enclosingLambda` of the top of the stack at this point we know that we popped all of `ast` parameters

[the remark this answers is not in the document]
mahfouz72: > one of the conditions does multiple pushes and you only do a single pop. Isn't that an issue?

what condition? for each `LAMBDA` we push all of its parameters (one or multiple parameters) and when we leave it we pop all of its parameters also. I don't see condition where we push more that pop

mahfouz72:
github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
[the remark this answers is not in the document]
rnveach: I am good with this. I was originally asking for 1 method to cover both `hasValidParentToken || isMethodInvocation`.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLambdaParameterShouldBeUnnamedCheck.java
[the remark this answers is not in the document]
rnveach: > pop all of its parameters 

I missed the while loop. I was thinking it was an if statement.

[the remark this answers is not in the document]
rnveach: > Comment is removed. actually, it doesn't add much value.

Not done.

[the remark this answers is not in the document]
mahfouz72: very weird. Done

rnveach (approved):

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_ru.properties
romani: Not done.

romani (approved):
Ok to merge

Issue #15105, opened 2024-06-22
IllegalIdentifierName should be updated to support new Java 22 keywords


rnveach:
Identified at https://github.com/checkstyle/checkstyle/pull/15084#discussion_r1649748634 ,

https://checkstyle.org/checks/naming/illegalidentifiername.html#IllegalIdentifierName
> Checks identifiers with a pattern for a set of illegal names, such as those that are restricted or contextual keywords. 

JLS Keywords as of Java 22:
https://docs.oracle.com/javase/specs/jls/se22/html/jls-3.html#jls-3.9

````
$ cat TestClass.java
public class Test {
    public void test() {
int exports = 0;
int opens = 0;
int requires = 0;
int uses = 0;
int yield = 0;         // violation
int module = 0;
int permits = 0;       // violation
int sealed = 0;        // violation
int var = 0;           // violation
// int non-sealed = 0; // doesn't compile
int provides = 0;
int to = 0;
int when = 0;
int open = 0;
int record = 0;        // violation
int transitive = 0;
int with = 0;
    }
}

PS> C:\Programs\jdk-22.0.1\bin\javac.exe .\TestClass.java
PS>

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>
    <property name="severity" value="warning"/>
    <property name="haltOnException" value="false"/>

    <module name="TreeWalker">
<module name="IllegalIdentifierName"/>
    </module>
</module>

$ java -jar checkstyle-10.17.0-all.jar -c TestConfig.xml TestClass.java
Starting audit...
[WARN] TestClass.java:7:5: Name 'yield' must match pattern '(?i)^(?!(record|yield|var|permits|sealed|_)$).+$'. [IllegalIdentifierName]
[WARN] TestClass.java:9:5: Name 'permits' must match pattern '(?i)^(?!(record|yield|var|permits|sealed|_)$).+$'. [IllegalIdentifierName]
[WARN] TestClass.java:10:5: Name 'sealed' must match pattern '(?i)^(?!(record|yield|var|permits|sealed|_)$).+$'. [IllegalIdentifierName]
[...]

PR #15486, opened 2024-08-11
Issue #15105: Update IllegalIdentifierCheck to add 'when' keyword in …

mahfouz72:
issue #15105

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/IllegalIdentifierNameCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/IllegalIdentifierNameCheck.java
@@ -34,11 +34,15 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
  * check uses a pattern to specify valid identifiers, users can also prohibit the usage
  * of certain symbols, such as "$", or any non-ascii character.
  * </p>
+ * <p>
+ * We will support latest java versions keywords please override the format property
+ * if you want to revert it to older java versions keywords only.
+ * </p>
  * <ul>
  * <li>
  * Property {@code format} - Sets the pattern to match valid identifiers.
  * Type is {@code java.util.regex.Pattern}.
- * Default value is {@code "(?i)^(?!(record|yield|var|permits|sealed)$).+$"}.
+ * Default value is {@code "(?i)^(?!(record|yield|var|permits|sealed|when)$).+$"}.
  * </li>
  * <li>
  * Property {@code tokens} - tokens to check
@@ -94,7 +98,7 @@ public class IllegalIdentifierNameCheck extends AbstractNameCheck {
      * Creates a new {@code IllegalIdentifierNameCheck} instance.
      */
     public IllegalIdentifierNameCheck() {
-        super("(?i)^(?!(record|yield|var|permits|sealed)$).+$");
+        super("(?i)^(?!(record|yield|var|permits|sealed|when)$).+$");
     }
 
     @Override
--- a/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionIllegalIdentifierNameTest.java
+++ b/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionIllegalIdentifierNameTest.java
@@ -46,7 +46,7 @@ public class XpathRegressionIllegalIdentifierNameTest extends AbstractXpathTestS
         final DefaultConfiguration moduleConfig =
             createModuleConfig(IllegalIdentifierNameCheck.class);
 
-        final String format = "(?i)^(?!(record|yield|var|permits|sealed)$).+$";
+        final String format = "(?i)^(?!(record|yield|var|permits|sealed|when)$).+$";
 
         final String[] expectedViolation = {
             "10:20: " + getCheckMessage(IllegalIdentifierNameCheck.class,
@@ -71,7 +71,7 @@ public class XpathRegressionIllegalIdentifierNameTest extends AbstractXpathTestS
         final DefaultConfiguration moduleConfig =
             createModuleConfig(IllegalIdentifierNameCheck.class);
 
-        final String format = "(?i)^(?!(record|yield|var|permits|sealed)$).+$";
+        final String format = "(?i)^(?!(record|yield|var|permits|sealed|when)$).+$";
 
         final String[] expectedViolation = {
             "9:17: " + getCheckMessage(IllegalIdentifierNameCheck.class,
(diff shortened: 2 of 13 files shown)


mahfouz72:
github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/IllegalIdentifierNameCheck.java
```diff
@@ -34,11 +34,15 @@
  * check uses a pattern to specify valid identifiers, users can also prohibit the usage
  * of certain symbols, such as "$", or any non-ascii character.
  * </p>
+ * <p>
+ * We will support latest java versions keywords please override the format property
+ * if you want to revert it to older java versions keywords only.
```
romani, line 39: This is big breaking change. We considered it few years ago when we added support to java parser for it but after running on real project it was clear that effect is too big and compilator handle it in good way and there is no confusion on names.

romani (changes requested):
Let's discuss this in detail again

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/IllegalIdentifierNameCheck.java
romani: I think we can find our old issue on this

rnveach: I look at this as more a bug, then anything. The check says it reports `restricted or contextual keywords`, but then I found a list of said keywords and it doesn't violate against half them. It shouldn't matter how big or small the impact is, the check says it will violate these kind of things and it doesn't.

Also, didn't we say we would move defaults to the latest Java version for checks just like we did with `RedundantModifier` and others.

If users think the new changes are too drastic, then can always override the property or not enable the check.

rnveach:
@mahfouz72 It looks like your doing one token at a time. I updated the first post with a task list.

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/IllegalIdentifierNameCheck.java
nrmancuso: We are updating checks to be Java 21 by default, we already agreed on this and have been doing this all summer so far. As long as we make a migration note in the issue, we are good.

romani: We discuss here default behavior.
Try to think on usage to be useful.

How useful is violating code that is not making any problem at all. Users used `when` in a lot of places, and if we look at real code, such usages are good.

Can we find real examples of code when we can see that usage of name `when` is bad and confusing?
If we put it as default it should be fact that almost all cases that we find need to be changed.

To avoid personal subjective trap, let's find such code in web and send PRs to projects, if majority of them accept such PRs we can be confident that this is good default.

nrmancuso: >To avoid personal subjective trap, let's find such code in web and send PRs to projects, if majority of them accept such PRs we can be confident that this is good default.

You are suggesting that we do exactly what you are warning about (falling into a "personal subjective trap"), and questions like "Can we find real examples of code when we can see that usage of name when is bad and confusing?" are as well. 

The bottom line is quite simple here. The very first sentence in the check description states:

>Checks identifiers with a pattern for a set of illegal names, such as those that are restricted or contextual keywords.

`when` falls into this category; it is a contextual keyword. The compiler does not help us in this case, either, which is even more reason to do this update. Additionally, `record` gives us precedence to do this update.

>let's find such code in web and send PRs to projects, if majority of them accept such PRs we can be confident that this is good default.

This is not a realistic expectation; this will take a massive amount of work to even find such cases, and it is highly unlikely that we will ever find any usages in projects that use Java 21 anyway. Then we will need to wait for maintainers review, etc.

If @romani will not green light this update, then we will just abandon this PR and move on.

romani: we do not need java21 projects, after update check will operate on projects that on old java our update of default will affect all.

People who care so much on `when` to be ilegal name in their projects, are not blocked by any means, if they started to care they need to update Check to do work for them.
I am fully ok to make it written in doc(extra example) to explain/show how to add `when` to validation. if in X years we will get a feedback that it is good to make it default, I am ok to change it. But again change will be based on real experience.

We can make voice meeting with any amount of participants, to let me explain my position more clearly.

mahfouz72: I think I more to agree with @nrmancuso @rnveach. Ideally, we should violate all from https://docs.oracle.com/javase/specs/jls/se22/html/jls-3.html#jls-3.9. This is based on the doc and the rationale of the check. if some don't like this then just override the property with what you want.

@romani I think I don't understand why do you have this opinion about adding support for "when" but not for "yield", "record" for example

mahfouz72: > I think we can find our old issue on this

https://github.com/checkstyle/checkstyle/pull/12665

report: https://github.com/checkstyle/checkstyle/pull/12665#issuecomment-1399439514

nrmancuso: This should be a very simple update; like I said above, if we can’t agree on this update we need to abandon this PR and move on. We should not spend anymore time on this.

mahfouz72: I agree. I am waiting for a final decision if not then I will not continue other keywords from the list in the issue

romani: >"record" for example
 
this word is too general, it was a big mistake to add it, in my job code this word is everywhere, and it is good.

romani: we missed a point of target in https://checkstyle.org/checks/naming/illegalidentifiername.html , Check make sense to forbid `var var = 1` or `record record(record r){}` , attention to fact that record is not good as name for RECORD_DEF, var is not good value for variable name.

This Check suppose to not violate names in certain context.

I agree to fully and for all forbid `_`. `var` I think for most of targets, all other is mostly for type names.

romani: I somebody need more detail and philosophical talk lets arrange voice chart in discord, I will share all details.

nrmancuso: @mahfouz72 lets close this PR, it is not worth it to waste any more time on this update.

Issue #15484, opened 2024-08-10
New Check UnusedTryResourceShouldBeUnnamed


mahfouz72:
child of #14942
Violate the non-use of try resources. They should be unnamed.

Originally proposed to extend `UnusedLocalVariableCheck`  in [#15024](https://github.com/checkstyle/checkstyle/issues/15024) but looks like the majority is voting for a new check to achieve this behaviour.

---

**Config**

```XML
$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
    <module name="TreeWalker">
        <module name="UnusedTryResourceShouldBeUnnamed"/>
    </module>
</module>
```

---

**Example**

```Java
public class Test {
     void test() {
         // violation below, 'Unused try resource 'a' should be unnamed'
         try (var a = lock()){
             
         } catch (Exception e) {
             
         }
         
         // ok below, declared as unnamed
         try (var _ = lock()){
             
         } catch (Exception e) {
             
         }
     }
     
     AutoCloseable lock() {
         return null;
     }
}
```

---

**Expected Output**

```
$ java -jar checkstyle-10.18.0-all.jar -c config.xml Test.java 
Starting audit...
[ERROR] D:\Test.java:4:15:Unused try resource should be unnamed [UnusedTryResourceShouldBeUnnamed]
Audit done.
Checkstyle ends with 1 errors.
```

romani:
do we expect such check to be very different from UnusedLambdaParameterShouldBeUnnamed ?

mahfouz72:
No, almost the same

romani:
@nrmancuso, is it reasonable to combine them in one Check and manage target by "tokens" ?
There prons and cons ..... for both approaches.

rnveach:
> is it reasonable to combine them

Not by their current check name. A lambda is not a try-with-resource, and vice versa.

rnveach:
I am good with first post.

nrmancuso:
> @nrmancuso, is it reasonable to combine them in one Check and manage target by "tokens" ? There prons and cons ..... for both approaches.

It is better to not create another JavadocMethod; small, simple, easy to reason about checks make for easy maintenance. Checking for some unused variable will have different nuances for each type (lambda, catch parameter, etc).

[...]

PR #15492, opened 2024-08-12
Issue #15159: RedundantModifierCheck should violate final modifier on…

### LINKED ISSUE #15159: RedundantModifer Should Violate Final Modifer on Unnamed Variables if jdkVersion GTE 22

mahfouz72:
closes #15159 

Diff Regression config: https://gist.githubusercontent.com/mahfouz72/afd57aa6e51ca161b7927acd959e5211/raw/bcf25b79718c2480283ab52d6662b42814056edb/check.xml
Diff Regression patch config: https://gist.githubusercontent.com/mahfouz72/33967bf86e4a29cafa429771b0e95808/raw/927d953c15bc189188acdacd0f7530cf2cc15135/check_patch.xml

### DIFF
--- a/config/pmd.xml
+++ b/config/pmd.xml
@@ -372,7 +372,8 @@
       | //ClassDeclaration[@SimpleName='SarifLogger']//MethodDeclaration[@Name='escape']
       | //ClassDeclaration[@SimpleName='LeftCurlyCheck'
         or @SimpleName='FinalLocalVariableCheck'
-        or @SimpleName='NPathComplexityCheck']
+        or @SimpleName='NPathComplexityCheck'
+        or @SimpleName='RedundantModifierCheck']
              //MethodDeclaration[@Name='visitToken']
       | //ClassDeclaration[@SimpleName='FallThroughCheck']
              //MethodDeclaration[@Name='isTerminated']
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
@@ -64,6 +64,9 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * {@code strictfp} modifier when using JDK 17 or later. See reason at
  * <a href="https://openjdk.org/jeps/306">JEP 306</a>
  * </li>
+ * <li>
+ * {@code final} modifier on unnamed variables when using JDK 22 or later.
+ * </li>
  * </ol>
  * <p>
  * interfaces by definition are abstract so the {@code abstract} modifier is redundant on them.
@@ -152,7 +155,7 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * Old JDK version numbering is supported (e.g. 1.8 for Java 8)
  * as well as just the major JDK version alone (e.g. 8) is supported.
  * Type is {@code java.lang.String}.
- * Default value is {@code "21"}.
+ * Default value is {@code "22"}.
  * </li>
  * <li>
  * Property {@code tokens} - tokens to check
@@ -178,7 +181,13 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#ANNOTATION_DEF">
  * ANNOTATION_DEF</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#RECORD_DEF">
- * RECORD_DEF</a>.
+ * RECORD_DEF</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#PATTERN_VARIABLE_DEF">
+ * PATTERN_VARIABLE_DEF</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_CATCH">
+ * LITERAL_CATCH</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LAMBDA">
+ * LAMBDA</a>.
  * </li>
  * </ul>
  * <p>
@@ -214,9 +223,9 @@ public class RedundantModifierCheck
     };
 
     /**
-     *  Constant for jdk 21 version number.
+     *  Constant for jdk 22 version number.
      */
-    private static final int JDK_21 = 21;
+    private static final int JDK_22 = 22;
 
     /**
      *  Constant for jdk 17 version number.
@@ -230,7 +239,7 @@ public class RedundantModifierCheck
      * as well as just the major JDK version alone (e.g. 8) is supported.
      *
      */
-    private int jdkVersion = JDK_21;
+    private int jdkVersion = JDK_22;
 
     /**
      * Setter to set the JDK version that you are using.
@@ -275,6 +284,9 @@ public class RedundantModifierCheck
             TokenTypes.RESOURCE,
             TokenTypes.ANNOTATION_DEF,
             TokenTypes.RECORD_DEF,
+            TokenTypes.PATTERN_VARIABLE_DEF,
+            TokenTypes.LITERAL_CATCH,
+            TokenTypes.LAMBDA,
         };
     }
 
@@ -300,8 +312,17 @@ public class RedundantModifierCheck
             case TokenTypes.RECORD_DEF:
                 checkForRedundantModifier(ast, TokenTypes.FINAL, TokenTypes.LITERAL_STATIC);
                 break;
-            case TokenTypes.CLASS_DEF:
             case TokenTypes.VARIABLE_DEF:
+            case TokenTypes.PATTERN_VARIABLE_DEF:
+                checkUnnamedVariables(ast);
+                break;
+            case TokenTypes.LITERAL_CATCH:
+                checkUnnamedVariables(ast.findFirstToken(TokenTypes.PARAMETER_DEF));
+                break;
+            case TokenTypes.LAMBDA:
+                processLambdaParameters(ast);
+                break;
+            case TokenTypes.CLASS_DEF:
             case TokenTypes.ANNOTATION_FIELD_DEF:
                 break;
             default:
@@ -315,6 +336,44 @@ public class RedundantModifierCheck
         if (jdkVersion >= JDK_17) {
             checkForRedundantModifier(ast, TokenTypes.STRICTFP);
         }
+    }
+
+    /**
+     * Process lambda parameters.
+     *
+     * @param lambdaAst node of type {@link TokenTypes#LAMBDA}
+     */
+    private void processLambdaParameters(DetailAST lambdaAst) {
+        final DetailAST lambdaParameters = lambdaAst.findFirstToken(TokenTypes.PARAMETERS);
+        if (lambdaParameters != null) {
+            TokenUtil.forEachChild(lambdaParameters, TokenTypes.PARAMETER_DEF,
+                    this::checkUnnamedVariables);
+        }
+    }
+
+    /**
+     * Check if the variable is unnamed and has redundant final modifier.
+     *
+     * @param ast node of type {@link TokenTypes#VARIABLE_DEF}
+     *     or {@link TokenTypes#PATTERN_VARIABLE_DEF}
+     *     or {@link TokenTypes#PARAMETER_DEF}
+     */
+    private void checkUnnamedVariables(DetailAST ast) {
+        if (jdkVersion >= JDK_22 && isUnnamedVariable(ast)) {
+            checkForRedundantModifier(ast, TokenTypes.FINAL);
+        }
+    }
+
+    /**
+     * Check if the variable is unnamed.
+     *
+     * @param ast node of type {@link TokenTypes#VARIABLE_DEF}
+     *     or {@link TokenTypes#PATTERN_VARIABLE_DEF}
+     *     or {@link TokenTypes#PARAMETER_DEF}
+     * @return true if the variable is unnamed
+     */
+    private static boolean isUnnamedVariable(DetailAST ast) {
+        return "_".equals(ast.findFirstToken(TokenTypes.IDENT).getText());
     }
 
     /**
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierFinalUnnamedVariables.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierFinalUnnamedVariables.java
@@ -1,0 +1,86 @@
+/*
+RedundantModifier
+tokens = (default)METHOD_DEF, VARIABLE_DEF, ANNOTATION_FIELD_DEF, INTERFACE_DEF, \
+         CTOR_DEF, CLASS_DEF, ENUM_DEF, RESOURCE, PATTERN_VARIABLE_DEF, LITERAL_CATCH, LAMBDA
+jdkVersion = (default)22
+
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.modifier.redundantmodifier;
+
+import java.util.function.BiFunction;
+
+public class InputRedundantModifierFinalUnnamedVariables {
+    void m(Object o) {
+        // violation below, 'Redundant 'final' modifier'
+        if (o instanceof final String _) { }
+        if (o instanceof final String __) { }
+        if (o instanceof final String s)
+        if (o instanceof final String _s) { }
+
+        // violation below, 'Redundant 'final' modifier'
+        final int _ = sideEffect();
+        final int __ = sideEffect();
+        final int x = sideEffect();
+        final int _x = sideEffect();
+
+    }
+
+    void m2(Object o) {
+        switch (o) {
+            // violation below, 'Redundant 'final' modifier'
+            case final String _ -> { }
+            case Float _ -> { }
+            case final Integer __ -> { }
+            case final Double s -> { }
+            default -> { }
+        }
+    }
+
+    void m3() {
+        // violation below, 'Redundant 'final' modifier'
+        try (final var a = lock();) {
+
+        } catch (final Exception e) {
+
+        }
+
+        // violation below, 'Redundant 'final' modifier'
+        try (final var _ = lock();) {
+
+        // violation below, 'Redundant 'final' modifier'
+        } catch (final Exception _) {
+
+        }
+    }
+
+    void m4() {
+        BiFunction<Integer, Integer, Integer> f = (final Integer a, final Integer b) -> {
+            return 5;
+        };
+
+        // violation below, 'Redundant 'final' modifier'
+        BiFunction<Integer, Integer, Integer> f2 = (final Integer _, final Integer b) -> {
+            return 5;
+        };
+
+        BiFunction<Integer, Integer, Integer> f3 = (final Integer _, final Integer _) -> {
+        // 2 violations above:
+        //                   'Redundant 'final' modifier'
+        //                   'Redundant 'final' modifier'
+            return 5;
+        };
+    }
+
+    int sideEffect() {
+        return 0;
+    }
+
+    AutoCloseable lock() {
+        return null;
+    }
+
+    record Point(int x, int y) { }
+}
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierFinalUnnamedVariablesWithOldVersion.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierFinalUnnamedVariablesWithOldVersion.java
@@ -1,0 +1,77 @@
+/*
+RedundantModifier
+tokens = (default)METHOD_DEF, VARIABLE_DEF, ANNOTATION_FIELD_DEF, INTERFACE_DEF, \
+         CTOR_DEF, CLASS_DEF, ENUM_DEF, RESOURCE, PATTERN_VARIABLE_DEF, LITERAL_CATCH, LAMBDA
+jdkVersion = 8
+
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.modifier.redundantmodifier;
+
+import java.util.function.BiFunction;
+
+public class InputRedundantModifierFinalUnnamedVariablesWithOldVersion {
+    void m(Object o) {
+        if (o instanceof final String _) { }
+        if (o instanceof final String __) { }
+        if (o instanceof final String s)
+        if (o instanceof final String _s) { }
+
+        final int _ = sideEffect();
+        final int __ = sideEffect();
+        final int x = sideEffect();
+        final int _x = sideEffect();
+    }
+
+    void m2(Object o) {
+        switch (o) {
+            case final String _ -> { }
+            case Float _ -> { }
+            case final Integer __ -> { }
+            case final Double s -> { }
+            default -> { }
+        }
+    }
+
+    void m3() {
+        // violation below, 'Redundant 'final' modifier'
+        try (final var a = lock();) {
+
+        } catch (final Exception e) {
+
+        }
+
+        // violation below, 'Redundant 'final' modifier'
+        try (final var _ = lock();) {
+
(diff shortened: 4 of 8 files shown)


mahfouz72:
javac 21 job and javac on local outputs are not consistent

running command locally :



---

Job URL:  https://app.circleci.com/pipelines/github/checkstyle/checkstyle/27299/workflows/a4aaa25c-0a96-4d6a-ab70-78aa5000d897/jobs/664816

```
src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierFinalUnnamedVariables.java:29: error: illegal start of expression
            case Point(final int a, final int _) when a == 0 -> { }
                       ^
src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierFinalUnnamedVariables.java:29: error: <identifier> expected
            case Point(final int a, final int _) when a == 0 -> { }
```

nrmancuso:
@mahfouz72 please show java versions and vendors for both CI execution and local, this is critical to help us understand what is going on here. We should also take a look at the grammar and see what it specifies.

romani:
Github, generate report for RedundantModifier/all-examples-in-one

mahfouz72:
@nrmancuso

Grammer:

> SwitchLabel:
> case [CaseConstant](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-CaseConstant) {, [CaseConstant](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-CaseConstant)}
> case null [, default]
> case [CasePattern](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-CasePattern) [[Guard](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-Guard)]
> default
> 
> CasePattern:
> [Pattern](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-Pattern)
> 
> Pattern:
> [TypePattern](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-TypePattern)
> [RecordPattern](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-RecordPattern)
> 
> TypePattern:
> [LocalVariableDeclaration](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-LocalVariableDeclaration)
> 
> RecordPattern:
> [ReferenceType](https://docs.oracle.com/javase/specs/jls/se21/html/jls-4.html#jls-ReferenceType) ( [[PatternList](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-PatternList)] )
> PatternList:
> [Pattern](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-Pattern) {, [Pattern](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-Pattern) }
> 
> LocalVariableDeclaration:
> {[VariableModifier](https://docs.oracle.com/javase/specs/jls/se21/html/jls-8.html#jls-VariableModifier)} [LocalVariableType](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-LocalVariableType) [VariableDeclaratorList](https://docs.oracle.com/javase/specs/jls/se21/html/jls-8.html#jls-VariableDeclaratorList)
> VariableModifier:
> [Annotation](https://docs.oracle.com/javase/specs/jls/se21/html/jls-9.html#jls-Annotation)
> final

I think the code is valid based on grammer. I used SDK 21 to run the code, CI execution uses also Release 21
```
    do
      javac --release 21 --enable-preview -d target "${file}"
    done
```

mahfouz72:
github, generate report

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierFinalUnnamedVariables.java
```diff
@@ -0,0 +1,33 @@
...
+         CTOR_DEF, CLASS_DEF, ENUM_DEF, RESOURCE
+jdkVersion = (default)21
+
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.modifier.redundantmodifier;
+
+public class InputRedundantModifierFinalUnnamedVariables {
```
nrmancuso, line 13: Please add a test case with a switch statement, with pattern matching, which uses unnamed variables (like `case final Integer _ -> System.out.println("Integer");`).

Thinking back, we may have missed this test case in other PRs.

nrmancuso (changes requested):
Item:

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierFinalUnnamedVariables.java
mahfouz72: You forgot that we took almost 3 days discussing the problem with javac 21 with this example

mahfouz72: This example will not compile by javac21 although it is 100% valid according to java grammar (It compiles locally on java version. 22). probably a bug that were solved

nrmancuso: You did not mention this in our discussion/your examples. You only mentioned record patterns.

nrmancuso: Compiles fine for me:
```
➜  src javac --version
javac 21.0.3

➜  src cat Test.java 
public class Test {
    void m(Object o) {
        switch (o) {
            case final Integer _ -> System.out.println("Integer");
            case String _ -> System.out.println("String");
            default -> System.out.println("Unknown");
        }
    }
}

```

mahfouz72: Yes sorry, I thought we had this bug with both examples but yes it works for pattern variables. The bug was in record pattern only. 

Done.

nrmancuso (approved):

config/pmd.xml
```diff
@@ -372,7 +372,8 @@
       | //ClassDeclaration[@SimpleName='SarifLogger']//MethodDeclaration[@Name='escape']
       | //ClassDeclaration[@SimpleName='LeftCurlyCheck'
         or @SimpleName='FinalLocalVariableCheck'
-        or @SimpleName='NPathComplexityCheck']
+        or @SimpleName='NPathComplexityCheck'
+        or @SimpleName='RedundantModifierCheck']
```
rnveach, line 376: Suppression was added, but no reason why.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -64,6 +64,10 @@
  * {@code strictfp} modifier when using JDK 17 or later. See reason at
  * <a href="https://openjdk.org/jeps/306">JEP 306</a>
  * </li>
+ * <li>
+ * {@code final} modifier on unnamed variables and unnamed pattern variables
+ * when using JDK 21 or later.
```
rnveach, line 69: Question: Is this 21 preview or actually 21?

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -317,6 +329,18 @@ public void visitToken(DetailAST ast) {
         }
     }
 
+    /**
+     * Check if the variable is unnamed.
+     *
+     * @param variableAst ast node st node of type {@link TokenTypes#VARIABLE_DEF}
```
rnveach, line 335: > ast node st node 

This doesn't make sense to me.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierFinalUnnamedVariablesWithOldVersion.java
```diff
@@ -0,0 +1,41 @@
+/*
+RedundantModifier
+tokens = (default)METHOD_DEF, VARIABLE_DEF, ANNOTATION_FIELD_DEF, INTERFACE_DEF, \
+         CTOR_DEF, CLASS_DEF, ENUM_DEF, RESOURCE
+jdkVersion = 8
+
+
+*/
+
+//non-compiled with javac: Compilable with Java21
```
rnveach, line 10: This is an akward test. It is only compilable by 21, but we set the Java version to 8.

I would split this into 2 tests. One which is actually like Java 8 (I understand we still compile with 11), and the other which is 21. The 21 tests needs a reason it is 21 while we set the Java version to something else.

I am sure we did this with the other PR. If you want to create a separate PR fix, it would be appreciated.

rnveach (changes requested):

config/pmd.xml
mahfouz72: It goes under the already existing reason, comment above the suppressions.

> The methods visitToken/leaveToken are a big SWITCH block with a number of IF blocks.
             If we split the block to several methods it will damage the readability.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: 21 was a first preview

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierFinalUnnamedVariablesWithOldVersion.java
mahfouz72: > One which is actually like Java 8 (I understand we still compile with 11),

variables named `_` compiles only on Java 8 and below and on Java 21 and below. So we have no other option to used java21 in files that have `_` in them.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: done

config/pmd.xml
rnveach: The one which lists
> Main, PackageObjectFactory, RequireThisCheck, HandlerFactory,
VariableDeclarationUsageDistanceCheck,

?

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
rnveach: So then like another PR, shouldn't this say 22 or specifically mention 21 previous and released in 22?

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/modifier/redundantmodifier/InputRedundantModifierFinalUnnamedVariablesWithOldVersion.java
rnveach: ok.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
rnveach: @nrmancuso @romani Does this create an issue as the logic in the check also actually says 21. We have no preview property.

nrmancuso: @rnveach good point, I think it makes sense to use the Java version when the feature was officially released

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -300,8 +307,13 @@ public void visitToken(DetailAST ast) {
             case TokenTypes.RECORD_DEF:
                 checkForRedundantModifier(ast, TokenTypes.FINAL, TokenTypes.LITERAL_STATIC);
                 break;
-            case TokenTypes.CLASS_DEF:
             case TokenTypes.VARIABLE_DEF:
+            case TokenTypes.PATTERN_VARIABLE_DEF:
+                if (jdkVersion >= JDK_21 && isUnnamedVariable(ast)) {
+                    checkForRedundantModifier(ast, TokenTypes.FINAL);
```
rnveach, line 313: Should/does this cover lambda parameters, catch parameters, and resources, which can be unnamed?
https://github.com/checkstyle/checkstyle/pull/15480
https://github.com/checkstyle/checkstyle/pull/15358
https://github.com/checkstyle/checkstyle/issues/15484

config/pmd.xml
mahfouz72: comment 
```
      <!-- Main, PackageObjectFactory, RequireThisCheck, HandlerFactory,
             VariableDeclarationUsageDistanceCheck,
             are big and should not be split without a good reason.  (This is for classReportLevel)
             
             The methods visitToken/leaveToken are a big SWITCH block with a number of IF blocks.
             If we split the block to several methods it will damage the readability. (This is for methodReportLevel, it doesn't list specific classes, it just says a method name which is visit or leave token)
```
supression
```
//ClassDeclaration[@SimpleName='Main'
        or @SimpleName='PackageObjectFactory' or @SimpleName='RequireThisCheck'
        or @SimpleName='VariableDeclarationUsageDistanceCheck'
        or @SimpleName='HandlerFactory']    ( First comment)
        ....
        
              | //ClassDeclaration[@SimpleName='LeftCurlyCheck'
        or @SimpleName='FinalLocalVariableCheck'
        or @SimpleName='NPathComplexityCheck'
        or @SimpleName='RedundantModifierCheck']
             //MethodDeclaration[@Name='visitToken']  (Second comment)
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: Good point, This check already covers resources whether it is named or unnamed (BTW it isn't listed in the doc, in "The check validates" section maybe we should add it to this list in another PR)


but we need to support unnamed lambda parameters and  unnamed catch parameters

```
cat src/Test.java
import java.util.function.Function;

public class Test {
    
    void m() {
        try (final var a = lock();) { // violation

        } catch (final Exception e) {

        }

        try (final var _ = lock();) { // violation

        } catch (final Exception _) {

        }
    }

    void m2() {
        Function<Integer, Integer> f = (final Integer _) -> {
            return 5;
        };
    }

    AutoCloseable lock() {
        return null;
    }
}
D:\CS\test
Starting audit...
[ERROR] D:\CS\test\src\Test.java:6:14: Redundant 'final' modifier. [RedundantModifier]
[ERROR] D:\CS\test\src\Test.java:12:14: Redundant 'final' modifier. [RedundantModifier]
Audit done.
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: done to 22. accordingly default changed to 22 also

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: Done

mahfouz72:
github, generate report

mahfouz72:
github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
```diff
@@ -317,6 +338,45 @@ public void visitToken(DetailAST ast) {
...
+     * Check if the variable is unnamed.
+     *
+     * @param ast node of type {@link TokenTypes#VARIABLE_DEF}
+     *     or {@link TokenTypes#PATTERN_VARIABLE_DEF}
+     *     or {@link TokenTypes#PARAMETER_DEF}
+     * @return true if the variable is unnamed
+     */
+    private static boolean isUnnamedVariable(DetailAST ast) {
+        return "_".equals(ast.findFirstToken(TokenTypes.IDENT)
+                .getText());
```
romani, line 377: `        return "_".equals(ast.findFirstToken(TokenTypes.IDENT).getText());`


may be in future we can make check that hint that wrapped code can be single line.

romani: alternative formatting is:
```
"_".equals(
    ast.findFirstToken(TokenTypes.IDENT).getText());
```

config/pmd.xml
romani: I am ok with suppression

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: @romani done

config/pmd.xml
rnveach: Suppression is not the issue. We declare with the suppression why we are suppressing it.

We add in the name of the class and the reason. Since the class name was never added, I can't tell why we are suppressing it, if it is agreeable, and what we would need to do in the future if we ever wanted to resolve it. The comments before it all have class names listed as we pointed to.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
rnveach: Please clarify what done means here. We added support for specific cases?

I ask because you mentioned lambda and catch but did not mention resources.

config/pmd.xml
mahfouz72: From what I see in the comment above suppression I think it was agreed on before that we can suppress violations in visit/leave Token if it has a big switch block regardless of the name of the class.

I can add the name of this check if you want but we didn't do this before with this specific case (visitToken method), if the suppressed method was anything other than this I would feel different.

mahfouz72: Do you want comment like this?

```

             The methods visitToken/leaveToken "in LeftCurlyCheck,  FinalLocalVariableCheck, NPathComplexityCheck and RedundantModifierCheck are a big SWITCH block with a number of IF blocks.
             If we split the block to several methods it will damage the readability.
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: done means that we add support for redundant final on unnamed lambda parameter and unnamed catch parameter, resource is already supported

mahfouz72:
I think failure in https://app.circleci.com/jobs/github/checkstyle/checkstyle/673440 is unrelated.

```
Starting a Gradle Daemon (subsequent builds will be faster)
Executing Gradle on JVM versions 16 and lower has been deprecated. This will fail with an error in Gradle 9.0. Use JVM 17 or greater to execute Gradle. Projects can continue to use older JVM versions via toolchains. Consult the upgrading guide for further information:
```

config/pmd.xml
rnveach: Ok, I changed my mind and am going to drop this.

> visit/leave Token if it has a big switch block regardless of the name of the class.

I know see what you mean by this. I don't think the way we organized these reasons here is very straightforward, but you are correct that we give a blanket statement based on method name.

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
rnveach: @mahfouz72 Can you updated regression with the new tokens added. I didn't see them listed at https://github.com/checkstyle/checkstyle/pull/15492#issuecomment-2307027002

rnveach: I also don't see new tokens at https://github.com/checkstyle/checkstyle/pull/15492/files#diff-9b10f3ecc777d61ee2044e7753ec6d50ea673d2d88ede4e44c87cc99f58ae733

mahfouz72:
github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java
mahfouz72: > Can you updated regression with the new tokens added

I generated another report but the workflow seems to be canceled I will regenerate now.

mahfouz72: > I also don't see new tokens at 

@rnveach done, because there was `(default)` the new default tokens were loaded even if we don't add them to the list in the config comment

mahfouz72: > @mahfouz72 Can you updated regression with the new tokens added. I didn't see them listed at [#15492 (comment)](https://github.com/checkstyle/checkstyle/pull/15492#issuecomment-2307027002)

https://github.com/checkstyle/checkstyle/pull/15492#issuecomment-2307569944

romani (approved):
Ok to merge

rnveach (approved):

Issue #15340, opened 2024-07-22
create "Correct" files for Indentation


romani:
create correct flles that are fully auto formatted and Checkstyle does not give any violation.

we can copy existing Inputs and auto format, and all violations should be a defect on checksyle side (with manual confirmation in style guide) that formatting is correct.

romani:
can we find in web any project that use auto format and checkstyle in same time ?
if project is just use auto format, we can simply run Indentation module over it and spot conflicts of checkstyle and formatter and consider a way to improve Indentation module to produce less violations.

@rdiachenko , do you know about any of such ?

Zopsss:
List of sections to keep track of issue:

- [ ] 4.1.1 Use of optional braces
- [ ] 4.1.2 Nonempty blocks: K & R style
- [ ] 4.1.3 Empty blocks: may be concise
- [ ] 4.2 Block indentation: +2 spaces
- [ ] 4.3 One statement per line
- [ ] 4.4 Column limit: 100
- [ ] 4.5.1 Where to break   
- [ ] 4.5.2 Indent continuation lines at least +4 spaces
- [ ] 4.6.1 Vertical Whitespace
- [ ] 4.6.2 Horizontal whitespace
- [ ] 4.8.2.1 One variable per declaration
- [ ] 4.8.2.2 Declared when needed
- [ ] 4.8.3.2 No C-style array declarations
- [ ] 4.8.4.1 Indentation
- [ ] 4.8.4.2 Fall-through: commented
- [ ] 4.8.4.3 Presence of the default label
- [ ] 4.8.5.2 Class Annotations
- [ ] 4.8.5.3 Methods And Constructors Annotations
- [ ] 4.8.5.4 Field Annotations
- [ ] 4.8.6.1 Block comment style
- [ ] 4.8.7 Modifiers
- [ ] 4.8.8 Numeric Literals

Zopsss:
Indentation defect found in formatter at: https://github.com/checkstyle/checkstyle/pull/15521#issuecomment-2299323787

Original code:

```java
    switch (a) {
      // comment
      // comment
      // comment
      case 1:
```

```java
    switch (a) {
      // comment
      case 1:
```

Formatted code:

```java
    switch (a) {
        // comment
        // comment
        // comment
      case 1:
```

```java
    switch (a) {
        // comment
      case 1:
```

An issue has been opened for this defect on formatter's GH: https://github.com/google/google-java-format/issues/1147, visit there for more detailed description

PR #15508, opened 2024-08-15
Issue #15340: added support to check InputFormatted files in google-java-format.sh and created InputFormatted file for section 4.3 one statement per line

Zopsss:
#15340

Added support to check for `InputFormatted` files in google-java-format.sh, created a list `INPUT_PATHS_FOR_FORMATTED_INPUTS` to store formatted inputs on which we will run the formatter via CI. The formatted inputs should not collide with our google config and should pass the test. If test fails then something is wrong with our config or with formatter.

Created `InputFormatted` file for OneStatementPerLineTest.java

The formatter was not able to convert following type of code:

```java
int var1 = 9, var2 = 99;
```

to 

```java
int var1 = 9;
int var2 = 99;
```

I guess this is a limitation of formatter as it cannot change and create new variables.  So I had to manually change such type of code.

This was the actual code:

https://github.com/checkstyle/checkstyle/blob/d71c8982953412035def51e30b33f26fefff7cba/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule43onestatement/InputOneStatementPerLine.java#L211-L216

I removed the violation comments and formatter formatted it to:

```java
  int var1 = 5, var4 = 5;
  int var2 = 6, var3 = 5;
```

Config was giving violation for it: `Each variable declaration must be in its own statement.`

### DIFF
--- a/.ci/google-java-format.sh
+++ b/.ci/google-java-format.sh
@@ -58,3 +58,10 @@ INPUT_PATHS=($(find src/it/resources/com/google/checkstyle/test/ -name "Input*.j
 for INPUT_PATH in "${INPUT_PATHS[@]}"; do
   java -jar "$JAR_PATH" --replace src/it/resources/com/google/checkstyle/test/"$INPUT_PATH"
 done
+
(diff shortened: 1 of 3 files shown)


romani (approved):
Ok yo merge

PR #15510, opened 2024-08-16
Issue #4006: testing of line break after lambda arrow

### LINKED ISSUE #4006: Clarification of lambda coverage by in google  style
Based on the discussion at https://github.com/checkstyle/checkstyle/issues/3755#issuecomment-285851574 and according to [Google Java Style Guide (3 November 2016 version) ](https://github.com/google/styleguide/commit/db0a26320f3e930c6ea7225ed53539b4fb31310c#diff-b6c9191ee5ccb7a5ef170e98e51d94d8R422) to cover the rule from the guide NoLineWrap should validate lambdas. It means that LAMBDA should be included in a set of acceptable tokens and the check needs to treat LAMBDAS accordingly to Google Guide by default (as we cannot predict how users will wrap lambdas lets follow Google Guide by default).

Examples of expected behavior:

```java
(String label, Long value, Object obj) -> { // OK
    ...
}

Predicate<String> predicate = str -> // OK
    isReversed(str);

(String label) -> // violation (should not have new line next to the arrow in a lambda)
{
    ...
}
```

romani:
Closes #4006

I moved marker comments out of lambdas to make code close to real  (no comment in the middle of lambda)

### DIFF
--- a/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule451wheretobreak/InputLambdaBodyWrap.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule451wheretobreak/InputLambdaBodyWrap.java
@@ -50,6 +50,11 @@ public class InputLambdaBodyWrap {
   }
 
   private void showComplexNestedLambda() {
+    // terminology: () - parentheses, [] - brackets, {} - braces
+    // Code below is definitely "unbraced expression" and it is single (but not a single-line),
+    // fact that is parentheses-ed is still definition of single.
+    // Multiple expressions will imply curly braces and `;`.
+    // so case below is ok
     Function<Integer, Function<Integer, Integer>> createAdder =
         x ->
             (y ->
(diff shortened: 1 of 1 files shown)


src/it/resources/com/google/checkstyle/test/chapter4formatting/rule451wheretobreak/InputLambdaBodyWrap.java
```diff
@@ -50,6 +50,7 @@ private void sumOfSquares() {
   }
 
   private void showComplexNestedLambda() {
+    // this non-compliant as lambda consists of a NOT single expression
     Function<Integer, Function<Integer, Integer>> createAdder =
         x ->
             (y ->
```
romani, line 56: formatter make it:
```
  private void showComplexNestedLambda() {
    // this non-compliant as lambda consists of a NOT single expression
    Function<Integer, Function<Integer, Integer>> createAdder = x -> (y -> x + y);
  }
```

style guide wording:
>A line is never broken adjacent to the arrow in a lambda, except that a break may come immediately after the arrow if the body of the lambda consists of a single unbraced expression.

terminology:
(parentheses)
[brackets]
{braces}

"single unbraced expression" is not supper clear, if this is applicable to nested. Conceptually wrapped in parentheses is single expression. So this code is gray area.

romani: after meditating , I think:
```
    Function<Integer, Function<Integer, Integer>> createAdder =
        x ->
            (y ->
                x + y);
```

is definitely "unbraced expression" and it is single, fact that is parentheses-ed is still definition of single. Because multiple expressions will imply curly braces and `;`.

rdiachenko: Yep, I agree, it is compliant, and no violation is expected. Please update the comment.

romani: updated

rdiachenko (approved):
lgtm

PR #15522, opened 2024-08-18
Issue #15445: Update NPathComplexity Check to Account for when expres…

### LINKED ISSUE #15445: Update NPathComplexity Check to Account for when expression as a Possible Execution Path
check doc: https://checkstyle.org/checks/metrics/npathcomplexity.html#Description

---

```
D:\CS\test javac src/Test.java
D:\CS\test cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
  <module name="TreeWalker">
    <module name="NPathComplexity">
    </module>
  </module>
</module>
D:\CS\test cat src/Test.java
public class Test {

    // violation, NPath Complexity is 3
    void m(Object o) {
        if (o instanceof String s && !s.isEmpty()) {

        } else {

        }
    }

    // violation, NPath Complexity is 2
    void m2(Object o) {
        switch (o) {
            case String s when !s.isEmpty() -> { }
            default -> {}
        }
    }
}
D:\CS\test java  -jar checkstyle-10.17.0-all.jar -c config.xml src/Test.java
Starting audit...
[ERROR] D:\CS\test\src\Test.java:4:5: NPath Complexity is 3 (max allowed is 0). [NPathComplexity]
[ERROR] D:\CS\test\src\Test.java:13:5: NPath Complexity is 2 (max allowed is 0). [NPathComplexity]
Audit done.
```
---

The current implementation of the NPath Complexity Check does not account for the complexity added by `when`. We should update the NPath Complexity Check to add a complexity point for `when`.

mahfouz72:
closes #15445.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
@@ -91,6 +91,7 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * <td>NP(for-range) + NP(expr1)+ NP(expr2) + NP(expr3) + 1</td></tr>
  * <tr><td>switch ([expr]) { case : [case-range] default: [default-range] }</td>
  * <td>S(i=1:i=n)NP(case-range[i]) + NP(default-range) + NP(expr)</td></tr>
+ * <tr><td>when[expr]</td><td>NP(expr) + 1</td></tr>
  * <tr><td>[expr1] ? [expr2] : [expr3]</td><td>NP(expr1) + NP(expr2) + NP(expr3) + 2</td></tr>
  * <tr><td>goto label</td><td>1</td></tr><tr><td>break</td><td>1</td></tr>
  * <tr><td>Expressions</td>
@@ -220,6 +221,7 @@ public final class NPathComplexityCheck extends AbstractCheck {
             TokenTypes.LITERAL_DEFAULT,
             TokenTypes.COMPACT_CTOR_DEF,
             TokenTypes.SWITCH_RULE,
+            TokenTypes.LITERAL_WHEN,
         };
     }
 
@@ -248,6 +250,9 @@ public final class NPathComplexityCheck extends AbstractCheck {
                 break;
             case TokenTypes.LITERAL_RETURN:
                 visitUnitaryOperator(ast, 0);
+                break;
+            case TokenTypes.LITERAL_WHEN:
+                visitWhenExpression(ast, 1);
                 break;
             case TokenTypes.CASE_GROUP:
                 final int caseNumber = countCaseTokens(ast);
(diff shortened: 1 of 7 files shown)


mahfouz72:
Github, generate report for NPathComplexity/all-examples-in-one

mahfouz72:
github, generate site

mahfouz72:
@romani fyi. the report workflow is cancelled without sending a message.

mahfouz72:
Github, generate report for NPathComplexity/all-examples-in-one

nrmancuso:
Report looks good, I did not manually count the changes but they seem reasonable.

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
```diff
@@ -92,6 +92,7 @@
  * <tr><td>switch ([expr]) { case : [case-range] default: [default-range] }</td>
  * <td>S(i=1:i=n)NP(case-range[i]) + NP(default-range) + NP(expr)</td></tr>
  * <tr><td>[expr1] ? [expr2] : [expr3]</td><td>NP(expr1) + NP(expr2) + NP(expr3) + 2</td></tr>
+ * <tr><td> when [expr] </td><td>NP(expr) + 1</td></tr>
```
rnveach, line 95: We don't need space before `when` and after `[expr]`.

I do find this odd that it is separated from the switch, especially it not coming directly after the switch's tr. It should be moved up one line at the least.

rnveach (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
mahfouz72: Moved it to be just below switch. I didn't want to keep it in the same row with switch to not make expression complicated and hard to read.

mahfouz72:
github, generate site

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
rnveach: Yea, I understood that, but then there is nothing that says this is for a switch which is what makes it odd for me. I am ok with this until Java uses `when` in a second location.

rnveach: @mahfouz72 Not done.
> We don't need space before when and after [expr].

rnveach:
I am good to approve after https://github.com/checkstyle/checkstyle/pull/15522#discussion_r1725341793

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java
mahfouz72: @rnveach done

rnveach (approved):

rnveach:
@mahfouz72 CI needs to be green in this PR.

mahfouz72:
Rebased CI should be green now.

mahfouz72:
@rnveach ping, sempahoreci is just stopped

romani:
Only nondex job did not run, I am ok to merge without restarting

PR #15525, opened 2024-08-18
Issue #15078: Add new input file for sealed classes in ClassFanOutCom…

### LINKED ISSUE #15078: Add Check Support for Java 17 Sealed Classes: ClassFanOutComplexitiy
child of https://github.com/checkstyle/checkstyle/issues/14969

Check documentation: https://checkstyle.org/checks/metrics/classfanoutcomplexity.html

---

```
PS D:\CS\test> cat config.xml                                                   
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="ClassFanOutComplexity">
        </module>
    </module>
</module>
public class Test implements I, J {  // violation
}

sealed class C permits D {   // should this be a violation ?
}
final class D extends C {   // violation
}
interface I {
}
interface J {
}
Starting audit...
[ERROR] D:\CS\test\src\Test.java:1:1: Class Fan-Out Complexity is 2 (max allowed is 0). [ClassFanOutComplexity]
[ERROR] D:\CS\test\src\Test.java:6:1: Class Fan-Out Complexity is 1 (max allowed is 0). [ClassFanOutComplexity]
Audit done.
PS D:\CS\test> 
```

---

from the doc: 

> Checks the number of other types a given class/record/interface/enum/annotation relies on. Also, the square of this has been shown to indicate the amount of maintenance required in functional programs (on a file basis) at least.

I don't think we should add 1 for the classes in the `permits` list. The super class `C` permitted certain classes to extend it but it doesn't rely on the permitted subclasses. the opposite is correct

mahfouz72:
closes #15078

### DIFF
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/classfanoutcomplexity/InputClassFanOutComplexitySealedClasses.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/classfanoutcomplexity/InputClassFanOutComplexitySealedClasses.java
@@ -1,0 +1,37 @@
+/*
+ClassFanOutComplexity
+max = 0
+excludedClasses = (default)ArrayIndexOutOfBoundsException, ArrayList, Boolean, Byte, \
+                  Character, Class, Collection, Deprecated, Deque, Double, DoubleStream, \
+                  EnumSet, Exception, Float, FunctionalInterface, HashMap, HashSet, \
+                  IllegalArgumentException, IllegalStateException, IndexOutOfBoundsException, \
+                  IntStream, Integer, LinkedHashMap, LinkedHashSet, LinkedList, List, Long, \
+                  LongStream, Map, NullPointerException, Object, Optional, OptionalDouble, \
+                  OptionalInt, OptionalLong, Override, Queue, RuntimeException, SafeVarargs, \
+                  SecurityException, Set, Short, SortedMap, SortedSet, Stream, String, \
+                  StringBuffer, StringBuilder, SuppressWarnings, Throwable, TreeMap, TreeSet, \
+                  UnsupportedOperationException, Void, boolean, byte, char, double, float, \
+                  int, long, short, var, void
+excludeClassesRegexps = (default)^$
+excludedPackages = (default)
+
+
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.metrics.classfanoutcomplexity;
+
+// violation below, 'Class Fan-Out Complexity is 2 (max allowed is 0)'
+public class InputClassFanOutComplexitySealedClasses implements I, J {
+}
+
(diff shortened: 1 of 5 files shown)


mahfouz72:
> and I think documentation has to be added since we declared in documentation how we process things.

This is blocked because the example test is still disabled for this check. Also, I think it is pretty obvious from the check description that the check will not count the classes in the permits list (because the base class doesn't depend on those classes). we can keep doc improvements in another PR when we enable examples test. what do you think?

nrmancuso:
Having the example test disabled doesn't block anything in this PR, please update the docs here

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/classfanoutcomplexity/InputClassFanOutComplexitySealedClasses.java
```diff
@@ -0,0 +1,37 @@
...
+*/
+
+//non-compiled with javac: Compilable with Java21
+package com.puppycrawl.tools.checkstyle.checks.metrics.classfanoutcomplexity;
+
+// violation below, 'Class Fan-Out Complexity is 2 (max allowed is 0)'
+public class InputClassFanOutComplexitySealedClasses implements I, J {
+}
+
+// ok below, 'sealed classes don't rely on the permitted subclasses'
```
nrmancuso, line 28: Are we forced to have quotes around `'sealed classes don't rely on the permitted subclasses'`? I find this strange.

mahfouz72:
github, generate site

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/metrics/classfanoutcomplexity/InputClassFanOutComplexitySealedClasses.java
mahfouz72: Removed.

nrmancuso (approved):

rnveach:
https://github.com/checkstyle/checkstyle/issues/15078#issuecomment-2184315859
>  I think documentation has to be added since we declared in documentation how we process things.

I am not seeing this. Are we not doing this?

nrmancuso:
> [#15078 (comment)](https://github.com/checkstyle/checkstyle/issues/15078#issuecomment-2184315859)
> 
> > I think documentation has to be added since we declared in documentation how we process things.
> 
> I am not seeing this. Are we not doing this because this is an input only?

The example test with sealed classes was updated, I am good with this.

mahfouz72:
> I am not seeing this. Are we not doing this because this is an input only?

Examples were updated to have sealed in it. Isn't this what you asked for?

rnveach (approved):

romani:
Github, generate report by config from InputClassFanOutComplexitySealedClasses.java

romani:
This is expected error

Issue #15449, opened 2024-08-05
Minimize list of excludes in google-java-format.sh


romani:
exclude list:
https://github.com/checkstyle/checkstyle/blob/56d5c8aa59890512dfc7efdc9fbbe9914fb2b94b/.ci/google-java-format.sh#L9-L23

I did excludes by folders, it very rough way, as prove of concept.
we need to minimize this list to be file base or as last resort folder based.

Zopsss:
List of folders to keep track of the issue:

- [ ] rule711generalform
- [ ] rule712paragraphs 
- [ ] rule713atclauses
- [x] rule72thesummaryfragment
- [x] rule73
- [ ] rule64
- [x] rule528
- [ ] rule53camelcase
- [ ] rule522classnames
- [x] rule51identifiernames
- [ ] rule41
- [ ] rule42
- [ ] rule43onestatement
- [ ] rule44columnlimit
- [ ] rule45
- [x] rule461verticalwhitespace
- [ ] rule462
- [ ] rule48
- [ ] rule3sourcefile
- [ ] rule331nowildcard
- [ ] rule332nolinewrap
- [ ] rule333orderingandspacing
- [ ] rule3421overloadsplit
- [ ] rule231filetab

Zopsss:
@romani all PRs regarding this issue are merged, it can be closed now

rdiachenko:
On Mac: `grep 'Input' .ci/google-java-format.sh | sed -E 's/.*Input([^\.]+)\..*java.*/\1/' | while read name; do [[ $(find ./src -type f -name "InputFormatted${name}.java") ]] || echo "Not found: InputFormatted${name}.java"; done`

```
Not found: InputFormattedSingleLineJavadocAndInvalidJavadocPosition.java
Not found: InputFormattedJavaDocTagContinuationIndentation.java
Not found: InputFormattedInvalidJavadocPosition.java
Not found: InputFormattedCamelCaseDefined.java
Not found: InputFormattedClassNames.java
Not found: InputFormattedUseOfOptionalBraces.java
Not found: InputFormattedEmptyBlocksAndCatchBlocks.java
Not found: InputFormattedEmptyFinallyBlocks.java
Not found: InputFormattedOneVariablePerDeclaration.java
Not found: InputFormattedDeclaredWhenNeeded.java
Not found: InputFormattedFallThrough.java
Not found: InputFormattedAnnotationLocation.java
[...]

PR #15542, opened 2024-08-22
Issue #15449: removed whole folder reference and added individual file names to be grepped in google-java-format.sh

Zopsss:
#15449

removed whole folder names from grep list and used individual file names instead.

### DIFF
--- a/.ci/google-java-format.sh
+++ b/.ci/google-java-format.sh
@@ -6,14 +6,14 @@ JAR_PATH="$1"
 
 INPUT_PATHS=($(find src/it/resources/com/google/checkstyle/test/ -name "Input*.java" \
     | sed "s|src/it/resources/com/google/checkstyle/test/||" \
-    | grep -v "rule711generalform" \
+    | grep -v "rule711generalform/InputSingleLineJavadocAndInvalidJavadocPosition.java" \
     | grep -v "rule712paragraphs/InputIncorrectRequireEmptyLineBeforeBlockTagGroup.java" \
     | grep -v "rule712paragraphs/InputIncorrectJavadocParagraph.java" \
     | grep -v "rule713atclauses/InputJavaDocTagContinuationIndentation.java" \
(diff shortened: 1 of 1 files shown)


.ci/google-java-format.sh
```diff
@@ -4,16 +4,17 @@ set -e
 
 JAR_PATH="$1"
 
-INPUT_PATHS=($(find src/it/resources/com/google/checkstyle/test/ -name "Input*.java" \
+INPUT_PATHS=($(find src/it/resources/com/google/checkstyle/test/ \
+    \( -name "Input*.java" -o -name "InputFormatted*.java" \) \
```
romani, line 8: this change is not required.
` "Input*.java"` is enough and fully covering `"InputFormatted*.java"`

.ci/google-java-format.sh
```diff
@@ -4,16 +4,17 @@ set -e
...
     | sed "s|src/it/resources/com/google/checkstyle/test/||" \
-    | grep -v "rule711generalform" \
+    | grep -v "rule711generalform/InputSingleLineJavadocAndInvalidJavadocPosition.java" \
```
romani, line 10: add formatted file for this file, or you plan to do this in separate PR ?

romani (changes requested):
items:

romani:
can you write bash block that find all Input file that are referenced in this script and make sure we have InputFormatted file for it ?
or it is too advanced bash skills ?

.ci/google-java-format.sh
romani: Please do this

.ci/google-java-format.sh
romani: Ok, it will be separate activity

.ci/google-java-format.sh
Zopsss: done

PR #15594, opened 2024-08-30
minor: fix line length violation in release notes

mahfouz72:


mahfouz72:
This becomes very annoying. It would be good if we could fix it in the generator itself.

romani (approved):

PR #15607, opened 2024-09-02
Add print logs and cleaning up cache in regression-repo…

### LINKED ISSUE #15606: Add print logs and cleaning up cache in regression-report.yml
Add new step to regression-report.yml in the beginning to clean cache and also print out the cache files being stored in generate_report.

relentless-pursuit:
Resolves #15606:

### DIFF
--- a/.github/workflows/regression-report.yml
+++ b/.github/workflows/regression-report.yml
@@ -22,8 +22,8 @@ env:
   CACHE_KEY_PR: "simple-cache-key-PR-${{ github.sha }}"
   CACHE_KEY_MASTER: "simple-cache-key-MASTER-${{ github.sha }}"
   CACHE_KEY_CONTRIBUTION: "simple-cache-key-CONTRIBUTION-${{ github.sha }}"
-  CACHE_KEY_CONFIG_FILES: "config-files-${{ github.sha }}"
-  CACHE_KEY_GENERATED_CONFIGS: "generated-configs-${{ github.sha }}"
+  CACHE_KEY_CONFIG_FILES: "config-files-${{ github.sha }}-${{ github.event.comment.id }}"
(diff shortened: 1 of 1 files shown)


romani:
Please fix lint for GitHub action

romani:
I don't like this magic.
Let's not do cleaning of such all caches.

If we generate files properly, we don't care what is cache at all. All we care is content and specific files.

relentless-pursuit:
> I don't like this magic. Let's not do cleaning of such all caches.
> 
> If we generate files properly, we don't care what is cache at all. All we care is content and specific files.

i changed cache keys. this should work.

.github/workflows/regression-report.yml
```diff
@@ -496,6 +496,17 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ${{ env.AWS_REGION }}
 
+      - name: Print config.xml
+        run: |
+          echo "Contents of config.xml:"
+          echo "-------------------------"
+          if [ -f .ci-temp/contribution/checkstyle-tester/diff_config.xml ]; then
+            cat .ci-temp/contribution/checkstyle-tester/diff_config.xml
```
romani, line 504: Please do this also for list of projects

romani (changes requested):
Items

.github/workflows/regression-report.yml
relentless-pursuit: done

romani (approved):

romani:
@relentless-pursuit , in next PR please send update to clean/delete content of cache at the end workflow, to not generate mess and keep cache to be light to mount to execution.

Issue #15609, opened 2024-09-02
google_checks SuppressionXpathSingleFilter for 'MethodName' check is too lenient in several ways


mches:
I have read check documentation: https://checkstyle.org/filters/suppressionxpathsinglefilter.html
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac PrintStreamTest.java

/var/tmp $ cat /google_checks.xml
cat: /google_checks.xml: No such file or directory

/var/tmp $ cat PrintStreamTest.java
import java.lang.annotation.Documented;
import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

class PrintStreamTest {
  void println__String_(@Test String x) {
    class Local {
      void println__String_(String x) {
        System.out.println(x);
      }
    }

    new Local().println__String_(x);
  }

  @Target(ElementType.PARAMETER)
  @Retention(RetentionPolicy.RUNTIME)
  @Documented
  @interface Test {}
}

/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.18.1-all.jar -c /google_checks.xml PrintStreamTest.java
Starting audit...
Audit done.
```

---

**Describe what you expect in detail.**
The two illegally-named non-test methods should be reported e.g.

```
Starting audit...
[WARN] /var/tmp/PrintStreamTest.java:8:8: Method name 'println__String_' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [MethodName]
[WARN] /var/tmp/PrintStreamTest.java:10:12: Method name 'println__String_' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [MethodName]
Audit done.
```

1. Multiple consecutive underscores are not allowed for test methods.
2. Trailing underscores are not allowed for test methods.
3. Each underscore-separated component should start with a lowercase letter.
4. Without a method annotation, the method is not a test method. 
5. A method on a local or anonymous class declared in a test method is not necessarily a test method.

If approved, [here](https://github.com/markitect-dev/checkstyle/commit/641ed74715809f7e6af20f2eefdf027e3d3e6a29?diff=unified&w=0) is a proposal I would like to submit as a PR.

Before:
```xml
[...]

PR #15617, opened 2024-09-04
Issue #15609: Fix overly-lenient google_checks MethodName SuppressionXpathSingleFilter

mches:
https://github.com/checkstyle/checkstyle/issues/15609

### DIFF
--- a/src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
@@ -29,13 +29,44 @@ public class InputMethodName {
   void fO() {} // violation 'Method name 'fO' must match pattern'
 
   @Test
-  void testing_foo() {}
+  void testing_foo() {
+    class LocalFoo {
+      void foo() {}
+
+      // violation below 'Method name 'testing_foo' must match pattern'
+      void testing_foo() {}
+    }
+
+    new Object() {
+      void foo() {}
+
+      // violation below 'Method name 'testing_foo' must match pattern'
+      void testing_foo() {}
+    };
+  }
+
+  void testing_foo(@FooTest String str) {
+    // violation above 'Method name 'testing_foo' must match pattern'
+    class LocalFoo {
+      void foo() {}
+
+      // violation below 'Method name 'testing_foo' must match pattern'
+      void testing_foo() {}
+    }
+
+    new Object() {
+      void foo() {}
+
+      // violation below 'Method name 'testing_foo' must match pattern'
+      void testing_foo() {}
+    };
+  }
 
   @Test
-  void testing_Foo() {}
+  void testing_Foo() {} // violation 'Method name 'testing_Foo' must match pattern'
 
   @Test
-  void testing_fOo() {}
+  void testing_fOo() {} // violation 'Method name 'testing_fOo' must match pattern'
 
   @Test
   void testingFoo() {}
@@ -44,7 +75,19 @@ public class InputMethodName {
   void testingFoo_foo() {}
 
   @Test
-  void testing_0123() {}
+  void testing_0123() {} // violation 'Method name 'testing_0123' must match pattern'
+
+  @Test
+  void testing_0123_() {} // violation 'Method name 'testing_0123_' must match pattern'
+
+  @Test
+  void testing__0123() {} // violation 'Method name 'testing__0123' must match pattern'
+
+  @Test
+  void testing__0123_() {} // violation 'Method name 'testing__0123_' must match pattern'
+
+  @Test
+  void testing__0123__() {} // violation 'Method name 'testing__0123__' must match pattern'
 
   @Test
   void Testing_Foo() {} // violation 'Method name 'Testing_Foo' must match pattern'
@@ -60,17 +103,49 @@ public class InputMethodName {
   @Test
   void TestingFooBad() {} // violation 'Method name 'TestingFooBad' must match pattern'
 
-  @ParameterizedTest
-  @ValueSource(strings = {"racecar", "radar", "level", "madam", "noon"})
-  void testing_foo1(String str) {}
+  @Test
+  void testing_foo_() {} // violation 'Method name 'testing_foo_' must match pattern'
+
+  @Test
+  void testing_Foo_() {} // violation 'Method name 'testing_Foo_' must match pattern'
+
+  @Test
+  void testing__foo() {} // violation 'Method name 'testing__foo' must match pattern'
+
+  @Test
+  void testing__Foo() {} // violation 'Method name 'testing__Foo' must match pattern'
+
+  @Test
+  void testing__foo_() {} // violation 'Method name 'testing__foo_' must match pattern'
+
+  @Test
+  void testing__Foo_() {} // violation 'Method name 'testing__Foo_' must match pattern'
 
   @ParameterizedTest
   @ValueSource(strings = {"racecar", "radar", "level", "madam", "noon"})
-  void testing_Foo1(String str) {}
+  void testing_foo1(String str) {
+    class LocalFoo {
+      void foo() {}
+
+      // violation below 'Method name 'testing_foo' must match pattern'
+      void testing_foo() {}
+    }
+
+    new Object() {
+      void foo() {}
+
+      // violation below 'Method name 'testing_foo' must match pattern'
+      void testing_foo() {}
+    };
+  }
 
   @ParameterizedTest
   @ValueSource(strings = {"racecar", "radar", "level", "madam", "noon"})
-  void testing_fOo1(String str) {}
+  void testing_Foo1(String str) {} // violation 'Method name 'testing_Foo1' must match pattern'
+
+  @ParameterizedTest
+  @ValueSource(strings = {"racecar", "radar", "level", "madam", "noon"})
+  void testing_fOo1(String str) {} // violation 'Method name 'testing_fOo1' must match pattern'
 
   @ParameterizedTest
   @ValueSource(strings = {"racecar", "radar", "level", "madam", "noon"})
@@ -82,7 +157,7 @@ public class InputMethodName {
 
   @ParameterizedTest
   @ValueSource(strings = {"racecar", "radar", "level", "madam", "noon"})
-  void testing_01231(String str) {}
+  void testing_01231(String str) {} // violation 'Method name 'testing_01231' must match pattern'
 
   @ParameterizedTest
   @ValueSource(strings = {"racecar", "radar", "level", "madam", "noon"})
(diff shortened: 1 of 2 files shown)


src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
```diff
@@ -60,17 +87,47 @@ void Testing_Foo2() {} // violation 'Method name 'Testing_Foo2' must match patte
...
+  void testing__foo() {} // violation 'Method name 'testing__foo' must match pattern'
+
+  @Test
+  void testing__Foo() {} // violation 'Method name 'testing__Foo' must match pattern'
+
+  @Test
+  void testing__foo_() {} // violation 'Method name 'testing__foo_' must match pattern'
+
+  @Test
+  void testing__Foo_() {} // violation 'Method name 'testing__Foo_' must match pattern'
```
Zopsss, line 106: can you also add test cases for digits right after consecutive underscores? and also trailing underscore? like 

```
testing__09() {}
testing_09_() {}
testing__09_() {}
testing__09__() {}
```

Zopsss (commented):
item 2:

src/main/resources/google_checks.xml
```diff
@@ -358,8 +358,9 @@
     <module name="SuppressionXpathSingleFilter">
       <property name="checks" value="MethodName"/>
       <property name="query" value="//METHOD_DEF[
-                                     .//ANNOTATION/IDENT[contains(@text, 'Test')]]//*"/>
-      <property name="message" value="'[a-z][a-z0-9][_a-zA-Z0-9]*'.*"/>
+                                     ./MODIFIERS/ANNOTATION//IDENT[contains(@text, 'Test')]
+                                   ]/IDENT"/>
+      <property name="message" value="'[a-z][a-z0-9][a-zA-Z0-9]*(?:_[a-z][a-z0-9][a-zA-Z0-9]*)*'"/>
```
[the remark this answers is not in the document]
mches: So, the subsequent name components can start with a digit in place of a lowercase letter, but it's still the case that each name component should be at least 2 characters.

Though `There is no One Correct Way to name test methods`, for the time being, we're not going to support arbitrary prose, with or without JUnit 5's `@DisplayNameGeneration(ReplaceUnderscores.class)` or `@IndicativeSentencesGeneration(generator = ReplaceUnderscores.class)`. In other words, we won't relax the rule so much that any name component, including the first, can start with an uppercase letter or be just single character. A user could add that exception to their config if desired.

Reference: https://junit.org/junit5/docs/current/user-guide/#writing-tests-display-name-generator

mches:
I've adjusted the regular expression to account for a leading digit in a name component and added the requested test cases.

src/main/resources/google_checks.xml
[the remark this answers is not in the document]
mches: A pair of suppressions that allow arbitrary underscore-delimited prose in nested test class and method names might look like this 

```xml
…
    <module name="SuppressionXpathSingleFilter">
      <property name="checks" value="TypeName"/>
      <property name="query" value="//CLASS_DEF[
                                     ./MODIFIERS/ANNOTATION//IDENT[@text = 'Nested']
                                   ]/IDENT"/>
      <property name="message" value="'[a-zA-Z][a-zA-Z0-9]*(?:_[a-zA-Z0-9]+)*'"/>
    </module>
…
    <module name="SuppressionXpathSingleFilter">
      <property name="checks" value="MethodName"/>
      <property name="query" value="//METHOD_DEF[
                                     ./MODIFIERS/ANNOTATION//IDENT[contains(@text, 'Test')]
                                   ]/IDENT"/>
      <property name="message" value="'[a-zA-Z][a-zA-Z0-9]*(?:_[a-zA-Z0-9]+)*'"/>
    </module>
…
 ```

**Comment by romani** [comment]
@Zopsss , please make a comment when you approving PR, we will finalize review.

File: src/main/resources/google_checks.xml (hunk 2 above)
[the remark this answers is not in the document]
Comment (Zopsss) [reply src/main/resources/google_checks.xml]: > In other words, we won't relax the rule so much that any name component, including the first, can start with an uppercase letter or be just single character. A user could add that exception to their config if desired.

Yes, if user want they can add this exception explicitly from their side. I'm ok with this, @romani can you also share your opinion about this?

File: src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java [hunk 3]
```diff
@@ -29,13 +29,40 @@ void f() {} // violation 'Method name 'f' must match pattern'
...
+
+    new Object() {
+      void foo() {}
+
+      void testing_foo() {} // violation 'Method name 'testing_foo' must match pattern'
+    };
+  }
+
+  void testing_foo(@FooTest String str) {
+    // violation above 'Method name 'testing_foo' must match pattern'
```
Zopsss, line 47: can you change violation comment placement and use `violation below` instead?

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
```diff
@@ -29,13 +29,40 @@ void f() {} // violation 'Method name 'f' must match pattern'
...
+    class LocalFoo {
+      void foo() {}
+
+      void testing_foo() {} // violation 'Method name 'testing_foo' must match pattern'
+    }
+
+    new Object() {
+      void foo() {}
+
+      void testing_foo() {} // violation 'Method name 'testing_foo' must match pattern'
```
Zopsss, line 42: can you use `// violation below` comments where there is no annotation usage?

Zopsss (approved):
Sorry for the late review, I was busy with other PRs and some college stuff. This is the last review from my side, I'm okay with all other changes, just need to change violation comment style:

mches:
> This is the last review from my side, I'm okay with all other changes, just need to change violation comment style

Changed all 6 occurrences

src/main/resources/google_checks.xml
```diff
@@ -368,8 +368,9 @@
     <module name="SuppressionXpathSingleFilter">
       <property name="checks" value="MethodName"/>
       <property name="query" value="//METHOD_DEF[
-                                     .//ANNOTATION/IDENT[contains(@text, 'Test')]]//*"/>
-      <property name="message" value="'[a-z][a-z0-9][_a-zA-Z0-9]*'.*"/>
+                                     ./MODIFIERS/ANNOTATION//IDENT[contains(@text, 'Test')]
+                                   ]/IDENT"/>
+      <property name="message" value="'[a-z][a-z0-9][a-zA-Z0-9]*(?:_[a-z0-9]{2}[a-zA-Z0-9]*)*'"/>
```
romani, line 373: Can we make it `value="'[a-z][a-z0-9][a-zA-Z0-9]*(?:_[a-z][a-z0-9][a-zA-Z0-9]*)*'"/>`
It will be clear that `[a-z][a-z0-9][a-zA-Z0-9]` is our pattern for name, and `_` is just separation between allowed names.

I was confused on why pattern so different.

romani (changes requested):
Item

src/main/resources/google_checks.xml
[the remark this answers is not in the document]
romani: As we are interpretation , subjective in many cases, but we need to satisfy some unclear majority, we should just do what explicitly defined in style guide.

We always encourage users to override something that is not matching to their vision on how code should looks like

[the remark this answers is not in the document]
mches: I can appreciate the consistency of holding all the name parts to the same standard. 

The closest the guide comes to touching on words starting with numbers is `"Turn on 2SV"` → `turnOn2sv`, where `"2SV"` is a word in the middle. `turnOn_2sv` would end up requiring a suppression.

src/main/resources/google_checks.xml
mches: I believe that's what I had originally. 😅

src/it/resources/com/google/checkstyle/test/chapter5naming/rule523methodnames/InputMethodName.java
```diff
@@ -116,7 +173,7 @@ void testingFoo2() {}
   void testingFoo_foo2() {}
 
   @RepeatedTest(2)
-  void testing_01232() {}
+  void testing_01232() {} // violation 'Method name 'testing_01232' must match pattern'
```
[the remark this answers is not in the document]
mches: Sorry, this will revert back to being disallowed due to [this](https://github.com/checkstyle/checkstyle/pull/15617#discussion_r1752998366) feedback

src/main/resources/google_checks.xml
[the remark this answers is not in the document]
romani: I doubt we need to allow simply snake case style.
Usage of _ should be very rare exception in naming, not snake style support.

romani (approved):
Ok to merge

mches:
Removed a stray blank line

romani (approved):

src/main/resources/google_checks.xml
Zopsss: > Can we make it `value="'[a-z][a-z0-9][a-zA-Z0-9]*(?:_[a-z][a-z0-9][a-zA-Z0-9]*)*'"/>` 

This regex will not allow digits immediately after `_` ( https://github.com/checkstyle/checkstyle/pull/15617#discussion_r1744103032 ), such naming convention isn't prohibited by style guide ( https://github.com/checkstyle/checkstyle/pull/15617#discussion_r1744093928 ) so we should also not restrict it. @romani

Zopsss (commented):
Waiting for @romani's reply here: https://github.com/checkstyle/checkstyle/pull/15617#discussion_r1761491279, I'm fine with everything else

src/main/resources/google_checks.xml
romani: https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

>Method names
Method names are written in [lowerCamelCase](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case).
Method names are typically verbs or verb phrases. For example, sendMessage or stop.

Underscores may appear in JUnit test method names to separate logical components of the name, with each component written in [lowerCamelCase](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case), for example transferMoney_deductsFromSource. There is no One Correct Way to name test methods.

I don't see there is definition of verbs, so words. I am ok to be it a bit restricted here and get extra feedback from users.
I don't think it is ok to name with digits from the beginning,  as style guide 

>with each component written in [lowerCamelCase](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case)

This phrase makes me think that my assumption is right. We should not allow digits to start a word/component.

romani:
Replied.
I am still ok to merge this PR. If real user with real examples come up, we can reconsider this decision.

src/main/resources/google_checks.xml
Zopsss: ok then, let's restrict it for now. If users demands for it then we will allow it

Zopsss (approved):
Im also ok to merge now

rdiachenko (approved):
lgtm

Issue #5678, opened 2018-04-02
Jacoco: main class doesn't work after verify test with Eclipse


rnveach:
Steps to repoduce:
1) clean
2) run `mvn verify test` to successful completion
3) run `com.puppycrawl.tools.checkstyle.Main` class

Results in:
````
Exception in thread "main" java.lang.NoClassDefFoundError: org/jacoco/agent/rt/internal_c13123e/Offline
	at com.puppycrawl.tools.checkstyle.Main.$jacocoInit(Main.java)
	at com.puppycrawl.tools.checkstyle.Main.<clinit>(Main.java)
Caused by: java.lang.ClassNotFoundException: org.jacoco.agent.rt.internal_c13123e.Offline
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 2 more
````

This never caused an issue before.

romani:
> clean

?? 

> run com.puppycrawl.tools.checkstyle.Main class

from Eclipse ? what version of eclipse and jacoco ?

rnveach:
> clean

Meant maven clean.

> from Eclipse ? what version of eclipse and jacoco ?

Jacoco is the version checkstyle uses in master.
Eclipse is Luna SR2.

rnveach:
Linux maven does the same thing.
I can reproduce the issue with just `mvn clean test` without verify.

pbludov:
I'm seeing the same bug in IDEA. Perhaps this is due to the update of the JaCoCo to version 0.8.1

pbludov:
>I can reproduce the issue with just `mvn clean test`

This is by design. The goal [restore-instrumented-classes](http://www.eclemma.org/jacoco/trunk/doc/restore-instrumented-classes-mojo.html) is bound to the phase `prepare-package` which is right after the phase `test`.

rnveach:
> which is right after the phase test.

As I stated, I can see jacoco in the class files **before** the test phase is complete. It doesn't need to wait till after.
[...]

PR #15627, opened 2024-09-05
Issue #15626: Explain what happens when an import matches two or more groups

### LINKED ISSUE #15626: What happens when an import matches two or more ImportOrder groups?
Reference documentation for `groups` here: https://checkstyle.sourceforge.io/checks/imports/importorder.html#Properties, which currently explains what happens when an import does not match any group, but it doesn't explain what happens if it matches two or more groups.

For example, if I configure checkstyle with:

```xml
<module name="ImportOrder">
    <property name="groups" value="
        java.,
        com.,
        com.mycompany.,
    "/>
</module>
```

checkstyle will approve this source code:

```java
import java.util.Date;

import com.sun.jna.Platform;

import com.mycompany.Foo;
```

Note that even though both the 2nd and 3rd imports match `com.`, the 3rd import goes in its own group. I can see in the code that checkstyle tries to pick the "best" match, but I wish that was clear in the documentation.

JayBazuzi:
fixes #15626

The documentation currently explains what happens when an import does not match any group, but it doesn't explain what happens if it matches two or more groups.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java
@@ -73,7 +73,9 @@ import com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil;
  * <li>
  * Property {@code groups} - Specify list of <b>type import</b> groups. Every group identified
  * either by a common prefix string, or by a regular expression enclosed in forward slashes
- * (e.g. {@code /regexp/}). All type imports, which does not match any group, falls into an
+ * (e.g. {@code /regexp/}). If an import matches two or more groups,
+ * the best match is selected (closest to the start, and the longest match).
+ * All type imports, which does not match any group, falls into an
  * additional group, located at the end.
  * Thus, the empty list of type groups (the default value) means one group for all type imports.
  * Type is {@code java.lang.String[]}.
@@ -116,7 +118,9 @@ import com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil;
  * <li>
  * Property {@code staticGroups} - Specify list of <b>static</b> import groups. Every group
  * identified either by a common prefix string, or by a regular expression enclosed in forward
- * slashes (e.g. {@code /regexp/}). All static imports, which does not match any group, fall into
+ * slashes (e.g. {@code /regexp/}). If an import matches two or more groups,
+ * the best match is selected (closest to the start, and the longest match).
+ * All static imports, which does not match any group, fall into
  * an additional group, located at the end. Thus, the empty list of static groups (the default
  * value) means one group for all static imports. This property has effect only when the property
  * {@code option} is set to {@code top} or {@code bottom}.
@@ -184,6 +188,8 @@ public class ImportOrderCheck
     /**
      * Specify list of <b>type import</b> groups. Every group identified either by a common prefix
      * string, or by a regular expression enclosed in forward slashes (e.g. {@code /regexp/}).
+     * If an import matches two or more groups,
+     * the best match is selected (closest to the start, and the longest match).
      * All type imports, which does not match any group, falls into an additional group,
      * located at the end. Thus, the empty list of type groups (the default value) means one group
      * for all type imports.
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java
```diff
@@ -73,7 +73,7 @@
  * <li>
  * Property {@code groups} - Specify list of <b>type import</b> groups. Every group identified
  * either by a common prefix string, or by a regular expression enclosed in forward slashes
- * (e.g. {@code /regexp/}). All type imports, which does not match any group, falls into an
+ * (e.g. {@code /regexp/}). If an import matches two or more groups, the best match is selected. All type imports, which does not match any group, falls into an
```
rnveach, line 76: What does best actually mean here?

rnveach:
FYI, we won't be able to merge a failing CI. Please run `mvn verify` on your local to run most tests and fix their errors.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java
JayBazuzi: It means whatever this algorithm does: https://github.com/checkstyle/checkstyle/blob/c1c73921b848af6ecc7ed7b2cc0c7f528d2f4d16/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L712-L726

I realize users might want to know what "best" means, and maybe someone will address that later; for now we're addressing the question of whether CheckStyle even considers this an acceptable use case.

JayBazuzi:
> FYI, we won't be able to merge a failing CI. Please run `mvn verify` on your local to run most tests and fix their errors.

`mvn verify` in the `master` branch with no changes fails for me with:

```
java.lang.NoClassDefFoundError: org/jacoco/agent/rt/internal_aeaf9ab/Offline
```

Seems like something that should be exercised in CI, right?

rnveach:
This is a known issue ( https://github.com/checkstyle/checkstyle/issues/5678#issuecomment-1304846846 ), you'll have to run clean with the mvn command.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java
rnveach: It looks like best is the closest to the start, and the longest match.

It looks like this is a regular expression find, and not a match. Non-regexp ensures we match the start of the string and an ending period.

JayBazuzi:
> FYI, we won't be able to merge a failing CI. Please run `mvn verify` on your local to run most tests and fix their errors.

CI is now passing.

JayBazuzi:
@rnveach I think this PR is ready. Can it be merged?

romani:
GitHub, generate website

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java
romani: @JayBazuzi ,

`If an import matches two or more groups, the best match is selected (closest to the start, and the longest match)`

Does this sound good ?

JayBazuzi: Looking at the resulting help page, the space is already overcrowded with text and adding that detail would make that problem worse.

If it was my project I would accept this change as-is as making things better, and then think about next steps like:
- no more change - this is enough detail to clue in users to the fact that multiple matches are OK
- extract all the common text to another place that goes in to detail, including examples

But it's not my project - you do what you want.

romani: Thanks a lot for sharing feedback.
Let me think, y will be merged until end of month for sure

romani: I think, it is better to update as I suggested, I can push update when I next time be around keyboard.

romani:
GitHub, generate website

romani (approved):
ok to merge

Thanks a lot for update

nrmancuso (approved):
Honestly most of this check's property descriptions are worded quite poorly; ideally we should extract the description of this check's "best match" logic to a more general area instead of congesting this table further. However, I am good with the changes here; they are definitely an improvement. Thanks @JayBazuzi !

Issue #15456, opened 2024-08-06
Define violation messages for all violations.


romani:
We Should actually demand message to be always specified. (even Check has only one message)

I more and more inclined to think that it will be better to have violation message at all cases to let reader not rely on memory what it should be. In this case I completely forgot what this Check do, it required me jump  out of this page to get to know what this violation mean.

Code:
`if (o instanceof ColoredPoint(@SuppressWarnings("")int x, int y, int z)) { } // violation`

Preferable:
```
 // violation below 'The warning '''' cannot be suppressed at this location.'
if (o instanceof ColoredPoint(@SuppressWarnings("")int x, int y, int z)) { }
```

please share your thoughts, and if you both feel the same we can start following this rule.

I see this as closed https://github.com/checkstyle/checkstyle/issues/11214 ,  we can start update of all violation comments to have message.

We need to remove suppression list that we introduced at https://github.com/checkstyle/checkstyle/pull/15452/files

mahfouz72:
@romani please add a good first issue if you agree

romani:
yes, please contribute

chris-carneiro:
Working on AnnotationOnSameLineCheck, AllTestsTest.java fails, because at some point walk()
```java
walk(Paths.get("src/test/java"), filePath -> {
    verifyHasProductionFile(allTests, filePath.toFile());
});
```
is given a .DS_Store as filePath input. 

Since, it is OS specific and I guess it would never happen on CI, though I'm questioning whether an improvement should be considered. 
For now I added a check here as a workaround: 
```java 
private static void verifyHasProductionFile(Map<String, List<String>> allTests, File file) {
    if (file.isFile() && !file.isHidden()) {
        ...
    }
}
```
Similar problem encountered and notified in this [comment](https://github.com/checkstyle/checkstyle/issues/7643#issuecomment-604791314)
What do you guys think ?

Mostafahassen1:
Is this issue open?

Zopsss:
@romani it'd be helpful to have a list of checks which are covered and which aren't, it'll help to track the issue and help contributors to select the next check.

romani:
I put link to latest suppression list that is embedded in source code, there are still bunch of Checks Inputs for update.

Each PR removes one line from suppression.

Mostafahassen1:
[...]

PR #15647, opened 2024-09-08
Issue #15456: Removed suppression list

AyaanDewani:
Issue 15456: Removed suppression list introduced previously


romani:
Goal is not clear , we expect code changes  for each suppression item

Issue #15555, opened 2024-08-26
Rename Example4.py to Example.cpp and change content to be some C


rnveach:
https://github.com/checkstyle/checkstyle/blob/28c50973f23158bad3e8cc2cef1b36320af84952/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/newlineatendoffile/Example4.py#L1

Python doesn't support comments like `//` and `/*` in Java. It only supports `#` and `'''`.

https://www.w3schools.com/Python/python_comments.asp

romani:
Ok we need to change exte

rnveach:
Make it a C or CPP file.

mohitsatr:
So I made it a .cpp with 3 lines of cpp codes. test is passing as well.
```
/*xml
<module name="Checker">
  <module name="NewlineAtEndOfFile">
    <property name="fileExtensions" value="java, xml, cpp"/>
  </module>
</module>
*/
// violation 7 lines above 'File does not end with a newline.'
// xdoc section -- start
#include 'iostream'
void main() {
    std::cout << "Hello world!";
}
// no ⤶ below it is violation
// xdoc section -- end
```

romani:
Please send PR

mohitsatr:
i found that my tests were not really passing. 
```
/*xml
<module name="Checker">
  <module name="NewlineAtEndOfFile">
    <property name="fileExtensions" value="java, xml, py"/>
  </module>
</module>
*/
// violation 7 lines above 'File does not end with a newline.'
// xdoc section -- start
void main() { // ⤶
    std::cout << "hello world!"; // ⤶
}
// no ⤶ below it is violation
// xdoc section -- end
```
on further dubugging, i found that `getActualViolations` is not returning expected value becaues `check.process(files)` is passing 0 `errorCount` into `getActualViolations`.

rnveach:
@mohitsatr 
> `<property name="fileExtensions" value="java, xml, py"/>`

File extensions has to be updated with the new file extension. Make sure the file is also valid to get a violation from the config you are using.

PR #15651, opened 2024-09-10
Rename example4.py to example4.cpp with cpp code

mohitsatr:
Resolves : #15555


romani:
No need to close PR, if you want the redo changes, just push with force from local to same branch as you have PR in GitHub, it will refresh.

PR #15660, opened 2024-09-14
Issue #15456: Specify violation messages for AnnotationOnSameLineCheck

chris-carneiro:
Issue #15456 

Attention: 
is given a .DS_Store as filePath input.

Since, it is OS specific and I guess it would never happen on CI, though I'm questioning whether an improvement should be considered.
For now I added a check - !file.isHidden() - here as a workaround.
For more details read https://github.com/checkstyle/checkstyle/issues/15456#issuecomment-2350937837

--------

Attention: CI can not detect this problem, as CI never open file browser to navigate on folders, browsing folder in UI creates such folders.

### DIFF
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/annotationonsameline/Example1.java
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/annotationonsameline/Example1.java
@@ -13,7 +13,8 @@ import org.junit.Before;
 // xdoc section -- start
 class Example1 {
 
-  @SuppressWarnings("deprecation") // violation
+  // violation below, "should be on the same line with its target."
+  @SuppressWarnings("deprecation")
   public Example1() {
   }
 
@@ -30,15 +31,16 @@ class Example1 {
   @Deprecated public Test1() {  // OK
   }
 
-  @Override // violation
+  @Override // violation, "should be on the same line with its target."
   public void fun1() {
   }
 
-  @Before // violation
+  @Before // violation, "should be on the same line with its target."
   @Override public void fun2() {  // OK
   }
 
-  @SuppressWarnings("deprecation") // violation
+  // violation below, "should be on the same line with its target."
+  @SuppressWarnings("deprecation")
   @Before public void fun3() {
   }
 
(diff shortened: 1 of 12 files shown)


src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/annotationonsameline/Example1.java
```diff
@@ -13,7 +13,9 @@
 // xdoc section -- start
 class Example1 {
 
-  @SuppressWarnings("deprecation") // violation
+  @SuppressWarnings("deprecation")
+  // violation above, "Annotation 'SuppressWarnings'
+  // should be on the same line with its target."
```
romani, line 18: please keep comment single-line.

please make it `// violation below ....`
to keep annotation close method and violation marker comment does not make code unrealistic.
Please apply this good practice to other case in your update.

romani (changes requested):
minor:

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/annotationonsameline/Example1.java
romani: Only in ExampleX.java files, such content will be on our website, so it has a bit more strict rules

romani (approved):
ok to merge if CI pass.

thanks a lot for update, it becomes a way better to read/deal with such files.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/annotationonsameline/Example1.java
chris-carneiro: Done, adapted the comment in other classes to keep comments single-lined and attempted to keep the annotation close to method as much as possible. 
I Cut the message in rare cases when the comment was on the same line-
i.e @Ann // violation, "should be..." -- In those cases, I figured adding the name in the comment would not be necessary so, shorten the message in order to satisfy the "close to the method" rule made sense in my opinion. (see InputAnnotationOnSameLineCheckTokensOnMethodAndVar.java 20:3)

Open to other comments

Issue #15683, opened 2024-09-20
JavadocMethod: false positive on compact constructors for nested record classes


cowwoc:
I have read check documentation: https://checkstyle.org/checks/javadoc/javadocmethod.html#JavadocMethod
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```cmd
set RUN_LOCALE="-Duser.language=en -Duser.country=US"
java %RUN_LOCALE% -jar  checkstyle-10.18.1-all.jar -c checkstyle.xml Outer.java

Starting audit...
[ERROR] Outer.java:24: Unused Javadoc tag. [JavadocMethod]
[ERROR] Outer.java:25: Unused Javadoc tag. [JavadocMethod]
[ERROR] Outer.java:26: Unused Javadoc tag. [JavadocMethod]
[ERROR] Outer.java:27: Unused Javadoc tag. [JavadocMethod]
[ERROR] Outer.java:28: Unused Javadoc tag. [JavadocMethod]
[ERROR] Outer.java:29: Unused Javadoc tag. [JavadocMethod]
[ERROR] Outer.java:30: Unused Javadoc tag. [JavadocMethod]
[ERROR] Outer.java:31: Unused Javadoc tag. [JavadocMethod]
Audit done.
Checkstyle ends with 9 errors.
```

Given:
---
```
package test;

import java.net.URI;
import java.time.Instant;

public class Outer
{
    /**
     * @param subject       a value that uniquely identifies the user on the identification server111
     * @param firstName     the user's first name
     * @param nickName      the user's nickname
     * @param fullName      the user's full name
     * @param picture       the user's profile picture
     * @param updatedAt     the last time that the profile was updated
     * @param email         the user's email address
     * @param emailVerified {@code true} if the email address has been verified
     */
    public record Inner(String subject, String firstName, String nickName, String fullName, URI picture,
                        Instant updatedAt, String email, boolean emailVerified)
    {
        /**
         * Creates a new UserInfo instance.
         *
         * @param subject       a value that uniquely identifies the user on the identification server222
         * @param firstName     the user's first name
         * @param nickName      the user's nickname
         * @param fullName      the user's full name
         * @param picture       the user's profile picture
         * @param updatedAt     the last time that the profile was updated
         * @param email         the user's email address
         * @param emailVerified {@code true} if the email address has been verified
[...]

PR #15687, opened 2024-09-21
JavadocMethodCheck: removed unnecessary tokens from acceptable

rnveach:
Unrelated to https://github.com/checkstyle/checkstyle/issues/15683 but noticed when looking at it and the code.

We set required tokens in this Check which aren't needed. With them removed, we are able to remove a bit of code which was blocking violations on those tokens.

Tokens to remove from acceptable:
TokenTypes.CLASS_DEF,
TokenTypes.ENUM_DEF,
TokenTypes.INTERFACE_DEF,
TokenTypes.RECORD_DEF

Regression: https://rveach.no-ip.org/checkstyle/regression/4/
No differences.

**Migration note:**
Remove above tokens from config if used.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
@@ -338,12 +338,7 @@ public class JavadocMethodCheck extends AbstractCheck {
 
     @Override
     public final int[] getRequiredTokens() {
-        return new int[] {
-            TokenTypes.CLASS_DEF,
-            TokenTypes.INTERFACE_DEF,
-            TokenTypes.ENUM_DEF,
-            TokenTypes.RECORD_DEF,
-        };
+        return CommonUtil.EMPTY_INT_ARRAY;
     }
 
     @Override
@@ -354,25 +349,16 @@ public class JavadocMethodCheck extends AbstractCheck {
     @Override
     public int[] getAcceptableTokens() {
         return new int[] {
-            TokenTypes.CLASS_DEF,
-            TokenTypes.ENUM_DEF,
-            TokenTypes.INTERFACE_DEF,
             TokenTypes.METHOD_DEF,
             TokenTypes.CTOR_DEF,
             TokenTypes.ANNOTATION_FIELD_DEF,
-            TokenTypes.RECORD_DEF,
             TokenTypes.COMPACT_CTOR_DEF,
         };
     }
 
     @Override
     public final void visitToken(DetailAST ast) {
-        if (ast.getType() == TokenTypes.METHOD_DEF
-                 || ast.getType() == TokenTypes.CTOR_DEF
-                 || ast.getType() == TokenTypes.ANNOTATION_FIELD_DEF
-                 || ast.getType() == TokenTypes.COMPACT_CTOR_DEF) {
-            processAST(ast);
-        }
+        processAST(ast);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


romani:
>we are able to remove a bit of code which was blocking violations on those tokens.

so there should be changes in behavior, right ?

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocmethod/InputJavadocMethodRecordsAndCompactCtors.java
```diff
@@ -5,7 +5,7 @@
 accessModifiers = (default)public, protected, package, private
 allowMissingParamTags = (default)false
 allowMissingReturnTag = (default)false
-tokens = METHOD_DEF, CTOR_DEF, ANNOTATION_FIELD_DEF, COMPACT_CTOR_DEF, RECORD_DEF, CLASS_DEF
+tokens = METHOD_DEF, CTOR_DEF, ANNOTATION_FIELD_DEF, COMPACT_CTOR_DEF
```
romani, line 8: this means breaking compatibility for users.
this check is very old, so users might have all this tokens.

no good reason to have such tokens, but it will still be compatibility issue - exception in run time after upgrade of checkstyle.

romani (changes requested):
update is ok,
but we need to agree that it is breaking changes.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
```diff
@@ -338,12 +338,7 @@ public void setAllowMissingReturnTag(boolean flag) {
 
     @Override
     public final int[] getRequiredTokens() {
-        return new int[] {
-            TokenTypes.CLASS_DEF,
-            TokenTypes.INTERFACE_DEF,
-            TokenTypes.ENUM_DEF,
-            TokenTypes.RECORD_DEF,
-        };
+        return CommonUtil.EMPTY_INT_ARRAY;
```
nrmancuso, line 341: Do we have a test case with no tokens? We should add one if not, and also generate a regression report with the same config.

nrmancuso (changes requested):
Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java
rnveach: It depends on what you mean. BDD, even when using default, lists all the tokens in the input file. I don't know how it works under the hood if it supplies an empty `tokens`, the same list as the default, or what.

34 input files for this check use `tokens = (default)`. So we have inputs that verify the default tokens property.

nrmancuso (approved):

nrmancuso:
@rnveach thanks for catching this and cleaning it up!

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocmethod/InputJavadocMethodRecordsAndCompactCtors.java
romani: @rnveach , are we in agreement ?

rnveach: I don't really see this as a breaking change, but you will do what you want anyways.

romani:
We have bunch bug fixes already merged.
We agreed to release bug fixes before breaking compatibility, to increase feature delivery without breaking changes.
But this change will be merged right after release.

PR #15704, opened 2024-09-25
Issue #15635: Investigate reducing extra cache on step Restore Generat…

### LINKED ISSUE #15635: Investigate reducing extra cache on step `Restore Generated Config and Project Files` in regression-report.yml
```
      - name: Restore Generated Config and Project Files
        if: needs.handle_generated_config_bundle.result == 'success'
        uses: actions/cache@v4
        with:
          path: ${{ env.GENERATED_CONFIGS_PATH }}
          key: ${{ env.CACHE_KEY }}-generated-configs
```

We need to investigate whether we can remove this caching step. 
We look into this suggestion: https://github.com/checkstyle/checkstyle/pull/15633#discussion_r1748971809

relentless-pursuit:
Resolves #15635

### DIFF
--- a/.github/workflows/regression-report.yml
+++ b/.github/workflows/regression-report.yml
@@ -118,7 +118,6 @@ jobs:
           git checkout -b ${{ steps.get_pr_details.outputs.branch }} \
                forked/${{ steps.get_pr_details.outputs.branch }}
 
-
       - name: Save cache
         uses: actions/cache/save@v4
         with:
@@ -296,8 +295,10 @@ jobs:
       - name: Cache Generated Config and Project Files
         uses: actions/cache@v4
         with:
-          path: ${{ env.GENERATED_CONFIGS_PATH }}
-          key: ${{ env.CACHE_KEY }}-generated-configs
+          path: |
+            .
+            ${{ env.GENERATED_CONFIGS_PATH }}
+          key: ${{ env.CACHE_KEY }}
 
       - name: Extract filename and create label
         id: create_label
@@ -366,16 +367,11 @@ jobs:
       - name: Restore cache
         uses: actions/cache/restore@v4
         with:
-          path: .
+          path: |
+            .
+            ${{ env.GENERATED_CONFIGS_PATH }}
           key: ${{ env.CACHE_KEY }}
           fail-on-cache-miss: true
-
-      - name: Restore Generated Config and Project Files
-        if: needs.handle_generated_config_bundle.result == 'success'
-        uses: actions/cache@v4
-        with:
-          path: ${{ env.GENERATED_CONFIGS_PATH }}
-          key: ${{ env.CACHE_KEY }}-generated-configs
 
       - name: Process local-based generated Config and Project File
         if: needs.handle_generated_config_bundle.result == 'success'


.github/workflows/regression-report.yml
```diff
@@ -296,8 +295,10 @@ jobs:
       - name: Cache Generated Config and Project Files
         uses: actions/cache@v4
         with:
-          path: ${{ env.GENERATED_CONFIGS_PATH }}
-          key: ${{ env.CACHE_KEY }}-generated-configs
+          path: |
```
romani, line 298: Whole step should be just 
```
- name: Save cache
        uses: actions/cache/save@v4
        with:
          path: .
          key: ${{ env.CACHE_KEY }}
```

As we just need to store current folder to cache.

.github/workflows/regression-report.yml
```diff
@@ -366,17 +367,12 @@ jobs:
       - name: Restore cache
         uses: actions/cache/restore@v4
         with:
-          path: .
+          path: |
+            .
+            ${{ env.GENERATED_CONFIGS_PATH }}
```
romani, line 372: Can we just restore "."?

romani (changes requested):
Items

.github/workflows/regression-report.yml
relentless-pursuit: I tried that. but, i am struggling to retrieve the generated_config folder in 
```
      - name: Process local-based generated Config and Project File
        if: needs.handle_generated_config_bundle.result == 'success'
        env:
          DIFF_CONFIG_LINK: ${{ needs.handle_generated_config_bundle.outputs.config_link }}
          DIFF_PROJECTS_LINK: ${{ needs.handle_generated_config_bundle.outputs.projects_link }}
        run: |
          mkdir -p .ci-temp
          echo "Config link: $DIFF_CONFIG_LINK"
          echo "Projects link: $DIFF_PROJECTS_LINK"
          cp "$DIFF_CONFIG_LINK" .ci-temp/diff_config.xml
          cp "$DIFF_PROJECTS_LINK" .ci-temp/project.properties
```

.github/workflows/regression-report.yml
relentless-pursuit: I can attach an workflow where i just removed the duplicate step, but in the step, it fails citing can't find generated_config.
```
      - name: Process local-based generated Config and Project File
        if: needs.handle_generated_config_bundle.result == 'success'
        env:
          DIFF_CONFIG_LINK: ${{ needs.handle_generated_config_bundle.outputs.config_link }}
          DIFF_PROJECTS_LINK: ${{ needs.handle_generated_config_bundle.outputs.projects_link }}
        run: |
          mkdir -p .ci-temp
          echo "Config link: $DIFF_CONFIG_LINK"
          echo "Projects link: $DIFF_PROJECTS_LINK"
          cp "$DIFF_CONFIG_LINK" .ci-temp/diff_config.xml
          cp "$DIFF_PROJECTS_LINK" .ci-temp/project.properties
```

relentless-pursuit: Even this workflow is failing for other scenarios.

relentless-pursuit: The issue with this change is that when we remove the extra caching step and rely solely on the main cache (path: .), the `generated configs` are not available in the cache when subsequent jobs try to restore it. This is because the cache is saved after the `checkout_and_cache job`, but the generated configs are created later in the `handle_generated_config_bundle` job.

relentless-pursuit: Since the cache is not designed for sharing data between jobs in the same workflow run, relying solely on the main cache (path: .) won't work for sharing generated files created in later jobs. And, also we have the constraint to to not set fail-on-cache-miss: false.

@romani, i tried lot many ways. But, i don't see any feasible solution. Let me know if you any ideas, i can execute and do a run in my own org.

romani: It would help me a lot of if you share link to workflow file , to see whole code and link to execution to see failure.

romani:
cache is immutable, so we cannot update it, see details in  issue.

PR #15738, opened 2024-10-04
Issue #14019: Resolve pitest common suppressions for Checker#acceptFi…

### LINKED ISSUE #14019: Resolve pitest survivals in modules

MohanadKh03:
Part of #14019 

https://checkstyle.sourceforge.io/config.html#Checker_Properties

Mutation: 
https://github.com/checkstyle/checkstyle/blob/474873ce4d7e632822155b6a4cba7dbda882d2a1/config/pitest-suppressions/pitest-common-suppressions.xml#L3-L10

### Explanation: 
WIP till making sure the PR is ready for review and that there is no new mutation coming out of this fix

### DIFF
--- a/config/pitest-suppressions/pitest-common-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-suppressions.xml
@@ -1,15 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
   <mutation unstable="false">
-    <sourceFile>Checker.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.Checker</mutatedClass>
-    <mutatedMethod>acceptFileStarted</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/utils/CommonUtil::relativizePath with argument</description>
-    <lineContent>final String stripped = CommonUtil.relativizePath(basedir, fileName);</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>LocalizedMessage.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.LocalizedMessage</mutatedClass>
     <mutatedMethod>&lt;init&gt;</mutatedMethod>
--- a/config/sevntu-suppressions.xml
+++ b/config/sevntu-suppressions.xml
@@ -41,6 +41,8 @@
   <suppress checks="CheckstyleTestMakeup" files="[\\/]RightCurlyTest\.java"/>
   <suppress checks="CheckstyleTestMakeup" files="[\\/]ClassAnnotationsTest\.java"/>
   <suppress checks="CheckstyleTestMakeup" files="[\\/]MethodsAndConstructorsAnnotationsTest\.java"/>
+  <!-- Required to get basedir's temporary folder name in runtime -->
+  <suppress checks="CheckstyleTestMakeup" files="[\\/]CheckerTest\.java"/>
 
   <!-- No need for constructor with cause -->
   <suppress checks="CauseParameterInException" files=".*MetadataGenerationException\.java"/>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
@@ -78,6 +78,7 @@ import com.puppycrawl.tools.checkstyle.checks.NewlineAtEndOfFileCheck;
 import com.puppycrawl.tools.checkstyle.checks.TranslationCheck;
 import com.puppycrawl.tools.checkstyle.checks.coding.HiddenFieldCheck;
 import com.puppycrawl.tools.checkstyle.checks.sizes.LineLengthCheck;
+import com.puppycrawl.tools.checkstyle.filefilters.BeforeExecutionExclusionFileFilter;
 import com.puppycrawl.tools.checkstyle.filters.SuppressionFilter;
 import com.puppycrawl.tools.checkstyle.internal.testmodules.CheckWhichThrowsError;
 import com.puppycrawl.tools.checkstyle.internal.testmodules.DebugAuditAdapter;
@@ -1688,6 +1689,47 @@ public class CheckerTest extends AbstractModuleTestSupport {
         };
 
         verify(checkerConfig, filePath, expected);
+    }
+
+    /**
+     * Reason of non-Input based testing:
+     * There are bunch of asserts that expects full path to file,
+     * usage of "basedir" make it stripped and we need put everywhere code like
+     * <pre>CommonUtil.relativizePath(checker.getConfiguration().getProperty("basedir"), file)</pre>
+     * but Checker object is not always available in code.
+     * Propagating it in all code methods will complicate code.
+     */
+    @Test
+    public void testRelativizedFileExclusion() throws Exception {
+        final DefaultConfiguration newLineAtEndOfFileConfig =
+                createModuleConfig(NewlineAtEndOfFileCheck.class);
+
+        final DefaultConfiguration beforeExecutionExclusionFileFilterConfig =
+                createModuleConfig(BeforeExecutionExclusionFileFilter.class);
+
+        beforeExecutionExclusionFileFilterConfig.addProperty("fileNamePattern",
+                        "^(?!InputCheckerTestExcludeRelativizedFile.*\\.java).*");
+
+        final DefaultConfiguration checkerConfig = createRootConfig(null);
+        checkerConfig.addChild(newLineAtEndOfFileConfig);
+        checkerConfig.addChild(beforeExecutionExclusionFileFilterConfig);
+
+        checkerConfig.addProperty("basedir",
+                temporaryFolder.getPath());
+
+        final String violationMessage =
+                getCheckMessage(NewlineAtEndOfFileCheck.class, MSG_KEY_NO_NEWLINE_EOF);
+
+        final String[] expected = {
+            "1: " + violationMessage,
+        };
+
+        final File tempFile = createTempFile("InputCheckerTestExcludeRelativizedFile", ".java");
+
+        final File[] processedFiles = {tempFile};
+
+        verify(createChecker(checkerConfig), processedFiles,
+                tempFile.getName(), expected);
     }
 
     public static class DefaultLoggerWithCounter extends DefaultLogger {


romani:
@MohanadKh03 , ping, do you need help ?

MohanadKh03:
My bad I got stuck in it and forgot to ping for help

I was going to follow the steps for this issue but noticed that if I hard-coded the mutation the checkstyle configs wouldnt allow it
Also as an analysis for what's the problem basically the ``baseDir`` is almost always (99% or even more of the times) and the ``beforeExecutionFileFilters`` seems to accept any uri no need to even relativize it 

Thats why I just added it like this but im not sure why the CI is failing tbh I thought the CI was failing due to a non-related issue 
Also for the Checker mutations should how could I use a config regression report ? Noticed from other mutation PRs over Checker that there was no regression report

Thanks for ping and for the help

romani:
@MohanadKh03 , please move forward to another issues or more simple pitest case, sometimes PR becomes complicated, and it is ok to abandon PR and switch to another issues and be productive in other contributions.

romani:
```
$ cat Test.java 
public class Test {
  private static void indentationIssue(int condition) {
    {
      System.out.println("True");
    }
  }
}

$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <property name="basedir" value="/var/tmp"/>
  <module name="TreeWalker">
    <module name="Indentation">
      <property name="basicOffset" value="2"/>
      <property name="braceAdjustment" value="2"/>
      <property name="caseIndent" value="2"/>
      <property name="throwsIndent" value="4"/>
      <property name="lineWrappingIndentation" value="4"/>
      <property name="arrayInitIndent" value="2"/>
    </module>
  </module>
</module>

$ java -jar checkstyle-10.18.3-SNAPSHOT-all.jar -c config.xml Test.java 
Starting audit...
[ERROR] Test.java:3:5: 'block lcurly' has incorrect indentation level 4, expected level should be 6. [Indentation]
[ERROR] Test.java:5:5: 'block rcurly' has incorrect indentation level 4, expected level should be 6. [Indentation]
Audit done.
 
$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <!-- <property name="basedir" value="/var/tmp"/> -->
  <module name="TreeWalker">
    <module name="Indentation">
      <property name="basicOffset" value="2"/>
      <property name="braceAdjustment" value="2"/>
      <property name="caseIndent" value="2"/>
      <property name="throwsIndent" value="4"/>
      <property name="lineWrappingIndentation" value="4"/>
      <property name="arrayInitIndent" value="2"/>
    </module>
  </module>
</module>

$ java -jar checkstyle-10.18.3-SNAPSHOT-all.jar -c config.xml Test.java 
Starting audit...
[ERROR] /var/tmp/Test.java:3:5: 'block lcurly' has incorrect indentation level 4, expected level should be 6. [Indentation]
[ERROR] /var/tmp/Test.java:5:5: 'block rcurly' has incorrect indentation level 4, expected level should be 6. [Indentation]
Audit done.

```

attention to:
`[ERROR] /var/tmp/Test.java:3:5`
vs 
`[ERROR] Test.java:3:5:`

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -375,8 +375,7 @@ private SortedSet<Violation> processFile(File file) throws CheckstyleException {
      * @return {@code true} if the file is accepted.
      */
     private boolean acceptFileStarted(String fileName) {
-        final String stripped = CommonUtil.relativizePath(basedir, fileName);
-        return beforeExecutionFileFilters.accept(stripped);
+        return beforeExecutionFileFilters.accept(fileName);
```
romani, line 378: we have only this filter for beforeExecution 
https://checkstyle.sourceforge.io/filefilters/beforeexecutionexclusionfilefilter.html#BeforeExecutionExclusionFileFilter

if we can not reproduce any problem with it, we can create test Filter that just not working if path is full not stripped a bit.

romani (changes requested):
hint on how to proceed:

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
romani: example of test with custom module/check
https://github.com/checkstyle/checkstyle/blob/c362c2234de44eee7692c498d9b3ec95a593aac4/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java#L1415-L1421

MohanadKh03: > we can create test Filter that just not working if path is full not stripped a bit.

Does this mean we can just use a custom test to check if the file is indeed stripped or null over this specific case ?

romani: yes, but try to use existing BeforeExecutionExclusionFileFilter to avoid custom hack modules (filter/check).

MohanadKh03: When I let the filter accept the fileName as it is **without being relativized/stripped** it works just as it would have worked if it got stripped so testing over it being not stripped would never really work 

Almost all the tests dont really have the file stripped it's taken as it is because ``baseDir`` is usually null **so relativizing the file /full/path/to/file would just remain as it is when baseDir is null** in the Checker test cases but even if we make a test with custom ``baseDir`` property it wouldnt differ because the filter accepts the full URIs and process them even if they are not stripped

romani: >Though I cant really get the point of why not just use the filter to accept the file name as it is (with or without stripping) ?

I don't remember, but changing it might break existing filters setup, so users will be angry.
Filters usually configured by what user see in output, if output is stripped user setup config the way he see it.

>The filter tests do pass even when not stripping the files thats why its confusing to me and cant really understand what exactly to test over ?

If tests pass, it just means that tests is not good enough to detect it. Whole idea of mutational testing as it most deep dive in meaning of each line of code, not just fact that line was executed (jacoco).

I wish we had pitest before we created such filters, so we would not be able make code that is not very clear. But we in real life, so we need to catch up. We should not remove this code and just need to make test that will explain in what use case it becomes to matter.

romani: @MohanadKh03, do you have any ideas how to cover this code? Or do you need help?

MohanadKh03: I honestly am not sure how to approach this right now
Could you please let me know if there is a specific test suite I could go through ? BeforeExecutionFilter or Checker tests ?

romani: take this as base for new test - BeforeExecutionExclusionFileFilterTest.testFileExclusion

take a look at CheckerTest.testSetSeverity , attention to `verifyWithInlineXmlConfig` and fact that xml config is located in Input file.

please make your own Input with xml config that has BeforeExecutionExclusionFileFilter as modules under Checker and `baseDir` is defined as property.

https://checkstyle.org/filefilters/beforeexecutionexclusionfilefilter.html#BeforeExecutionExclusionFileFilter
https://checkstyle.org/config.html#Checker_Examples

Violation message will not have full filename if `CommonUtil.relativizePath` is used.
So we can validate that `baseDir` is applied to violation and name is short, not absolute path.


We might need to make method that returns violation from Checker to get access to full name of file in violation. ChekckerTest has a lot of test method as example.

MohanadKh03: Sounds great thanks for your help now I get the xml input's idea 
But I do see something that is not clear yet in the ``testFileExclusion`` it says this is a filtered violation 
https://github.com/checkstyle/checkstyle/blob/9b9d3f945c4e5e098a96f5d9f9cb7def46409234/src/test/resources/com/puppycrawl/tools/checkstyle/filefilters/beforeexecutionexclusionfilefilter/InputBeforeExecutionExclusionFileFilter.java#L15-L18 
even though here it is supposed to be an unfiltered one ?
https://github.com/checkstyle/checkstyle/blob/9b9d3f945c4e5e098a96f5d9f9cb7def46409234/src/test/java/com/puppycrawl/tools/checkstyle/filefilters/BeforeExecutionExclusionFileFilterTest.java#L77-L79

As for the tests I think what needs to be done now is to add a proper match for a file that contains a ``baseDir`` and should be excluded when not relativized in this specific case to avoid the mutation am I right ?

romani: >As for the tests I think what needs to be done now is to add a proper match for a file that contains a baseDir and should be excluded when not relativized in this specific case to avoid the mutation am I right ?

Right.

>even though here it is supposed to be an unfiltered one ?

Please send separate PR to remove such misleading comment. :) .

MohanadKh03:
CI not happy with this way of using configuration properties 1708 and 1718

https://github.com/checkstyle/checkstyle/blob/3fefcae43b6e19645ce68841f8a17352c1479516/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java#L1707-L1708

https://github.com/checkstyle/checkstyle/blob/3fefcae43b6e19645ce68841f8a17352c1479516/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java#L1718


Isnt this the way to add configurations ? 
Could you please let me know if this as a PR-related CI failure or is there another way to add properties to configurations ? @romani

romani:
```
[checkstyle] [ERROR] C:\projects\checkstyle\src\test\java\com\puppycrawl\tools\
checkstyle\CheckerTest.java:1708:25:
 Unknown property name/value. [CheckstyleTestMakeup]
[ERROR] [checkstyle] [ERROR] C:\projects\checkstyle\src\test\java\com\puppycrawl\tools
\checkstyle\CheckerTest.java:1718:46:
 Unknown property name/value. [CheckstyleTestMakeup]
```

romani:
Can we put xml config to Input file ?
Example testUnmappableCharacters()

MohanadKh03:
> Can we put xml config to Input file ? Example testUnmappableCharacters()

I tried this at the beginning but for ``basedir`` to be used appropriately as an actual base directory with the full path (from project's root path) to get relativized later in ``Checker`` then it should be used in run time to get its value in here not hard-coded into the input xml
https://github.com/checkstyle/checkstyle/blob/3fefcae43b6e19645ce68841f8a17352c1479516/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java#L1716-L1718

I dont think it is possible to do so at compile-time by hardcoding the project path since it will differ by root path

rnveach:
> Unknown property name/value. 

It is looking for a string/null in the method call. Just inline the variable.

MohanadKh03:
@rnveach
Uh apparently this still counts as an expression cuz it is still considering the expression of ``System.getProperty`` as an expression  
```java
checkerConfig.addProperty("basedir",
                System.getProperty("user.dir")
                        + "/src/test/resources/com/puppycrawl/tools/checkstyle/checker");
```

rnveach:
That is correct. As I mentioned it only accepts strings/null and technically some paths. If it doesn't match those it will throw an exception.

You technically shouldn't be using a system property. It is a global that is outside our control. All our other treats reference files in our repo with no issues.

MohanadKh03:
Now the mutation is killed and fixed the global property thing by using the temp folder that already exists in CheckerTest instead 
But had to suppress ``CheckstyleTestMakeup`` because it is not possible to get the temp folder's name at compile time when it gets generated everytime with a new random name 
WDYT ? @romani @rnveach

romani:
if this works, good.
So we can put config to Input file.
```
<module name="Checker">
  <property name="basedir" value="${testRelativizedFileExclusion}"/>
.....
```

and we set this property in test before execution https://stackoverflow.com/a/9736906

romani:
>please add comment above test methods that is explains why we need non-Input testing.

Please do this. 
Usage of `verify` method supposed to be forbidden and used in exceptional case. We need to keep reasons of why this case is exceptional.


Meanwhile I will investigate usage of Input.
Please do not wait PR merge, I will need few days. Please go forward to fix other issues

romani:
to make it Input based we should update all our test suite code to expect stripped file path
```
--- a/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
@@ -575,7 +575,9 @@ public abstract class AbstractModuleTestSupport extends AbstractPathTestSupport
         final Map<String, List<String>> actualViolations =
                 getActualViolations(checker.process(files));
         checker.destroy();
-        return actualViolations.getOrDefault(file, new ArrayList<>());
+        final String strippedFile = CommonUtil.relativizePath(
+            checker.getConfiguration().getProperty("basedir"), file);
+        return actualViolations.getOrDefault(strippedFile, new ArrayList<>());
     }

```


```
✔ ~/java/github/romani/checkstyle [MohanadKh03/checker/acceptfilestarted L|✚ 2…1] 
07:32 $ cat src/test/resources/com/puppycrawl/tools/checkstyle/checker/InputCheckerTestExcludeRelativizedFile.java
/*xml
<module name="Checker">
  <property name="basedir" value="${testRelativizedFileExclusion}"/> <!-- this breaks our test suite -->
    <property name="fileNamePattern" value="^InputCheckerTestExcludeRelativizedFile1.java$"/>
  </module>
  <module name="RegexpSingleline">
    <property name="format" value="InputCheckerTestExcludeRelativizedFile"/>
    <property name="minimum" value="0"/>
    <property name="maximum" value="1"/>
  </module>
</module>
*/
// violation 6 lines above 'Line matches the illegal pattern'
package com.puppycrawl.tools.checkstyle.checker;

// violation below 'Line matches the illegal pattern'
public class InputCheckerTestExcludeRelativizedFile {
}

```

whole patch:
```
✔ ~/java/github/romani/checkstyle [MohanadKh03/checker/acceptfilestarted L|✚ 2…1] 
07:33 $ git diff
diff --git a/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java b/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
index 61b5398..f941aac 100644
--- a/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java
@@ -575,7 +575,9 @@ public abstract class AbstractModuleTestSupport extends AbstractPathTestSupport
         final Map<String, List<String>> actualViolations =
                 getActualViolations(checker.process(files));
         checker.destroy();
-        return actualViolations.getOrDefault(file, new ArrayList<>());
+        final String strippedFile = CommonUtil.relativizePath(
+            checker.getConfiguration().getProperty("basedir"), file);
+        return actualViolations.getOrDefault(strippedFile, new ArrayList<>());
     }
 
     /**
diff --git a/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java b/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
index 3069001..e5ef37e 100644
--- a/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
@@ -77,8 +77,8 @@ import com.puppycrawl.tools.checkstyle.api.Violation;
 import com.puppycrawl.tools.checkstyle.checks.NewlineAtEndOfFileCheck;
 import com.puppycrawl.tools.checkstyle.checks.TranslationCheck;
 import com.puppycrawl.tools.checkstyle.checks.coding.HiddenFieldCheck;
+import com.puppycrawl.tools.checkstyle.checks.regexp.RegexpSinglelineCheck;
 import com.puppycrawl.tools.checkstyle.checks.sizes.LineLengthCheck;
-import com.puppycrawl.tools.checkstyle.filefilters.BeforeExecutionExclusionFileFilter;
 import com.puppycrawl.tools.checkstyle.filters.SuppressionFilter;
 import com.puppycrawl.tools.checkstyle.internal.testmodules.CheckWhichThrowsError;
 import com.puppycrawl.tools.checkstyle.internal.testmodules.DebugAuditAdapter;
@@ -1693,35 +1693,17 @@ public class CheckerTest extends AbstractModuleTestSupport {
 
     @Test
     public void testRelativizedFileExclusion() throws Exception {
-        final DefaultConfiguration newLineAtEndOfFileConfig =
-                createModuleConfig(NewlineAtEndOfFileCheck.class);
-
-        final DefaultConfiguration beforeExecutionExclusionFileFilterConfig =
-                createModuleConfig(BeforeExecutionExclusionFileFilter.class);
-
-        beforeExecutionExclusionFileFilterConfig.addProperty("fileNamePattern",
-                        "^(?!InputCheckerTestExcludeRelativizedFile.*\\.java).*");
-
-        final DefaultConfiguration checkerConfig = createRootConfig(null);
-        checkerConfig.addChild(newLineAtEndOfFileConfig);
-        checkerConfig.addChild(beforeExecutionExclusionFileFilterConfig);
-
-        checkerConfig.addProperty("basedir",
-                temporaryFolder.getPath());
-
-        final String violationMessage =
-                getCheckMessage(NewlineAtEndOfFileCheck.class, MSG_KEY_NO_NEWLINE_EOF);
 
         final String[] expected = {
-            "1: " + violationMessage,
+          "8: Line matches the illegal pattern 'InputCheckerTestExcludeRelativizedFile'.",
+          "18: Line matches the illegal pattern 'InputCheckerTestExcludeRelativizedFile'.",
         };
 
-        final File tempFile = createTempFile("InputCheckerTestExcludeRelativizedFile", ".java");
-
-        final File[] processedFiles = {tempFile};
+        String folderPath = new File("src/" + getResourceLocation()
+            + "/resources/" + getPackageLocation()).getCanonicalPath();
+        System.setProperty("testRelativizedFileExclusion", folderPath);
 
-        verify(createChecker(checkerConfig), processedFiles,
-                tempFile.getName(), expected);
+        verifyWithInlineXmlConfig(getPath("InputCheckerTestExcludeRelativizedFile.java"), expected);
     }
 
     public static class DefaultLoggerWithCounter extends DefaultLogger {
✔ 
```

To avoid big update for single test, we are Ok to use non Input based testing and coverage.
There are bunch of asserts that expects full path to file, so usage of "basedir" make it stripped and we need put everywhere `CommonUtil.relativizePath(checker.getConfiguration().getProperty("basedir"), file)` but Checker object is not always available in code. Propagating it in all code methods will complicate code, so lets not do it.

romani:
comment it added.

romani (approved):
Ok to merge

romani:
@MohanadKh03, looks like you passed all "good" issues.
Do you need suggestion on next or what to choose yourself.
Now you are fully onboarded and any issue unlikely be a problematic for you.

MohanadKh03:
> @MohanadKh03, looks like you passed all "good" issues.
Do you need suggestion on next or what to choose yourself?
Now you are fully onboarded and any issue unlikely be a problematic for you.

Sure im ready . Any suggestions for specific issues ?

PR #15773, opened 2024-10-14
Issue #9118: Add method parameters placement and alignment checks

### LINKED ISSUE #9118: Check Method/Constructor Parameter List Wrapping
I would like to be able to check if the parameter list in methods and constructors is wrapped according to the following rules:

- Either all parameters are on a single line
- Or they are broken up into multiple lines, each on an individual line
- When parameters are on multiple lines, these lines are aligned

To illustrate this, here are a couple of code examples:

### Good

```
void f1(
  int p1,
  @Ann int p2,
  int p3
) {}

void f2(int p1, int p2, int p3, int p4, int p5, int p6, int p7, int p8, int... p9) {}

void f3(int p1,
        @Ann int p2,
        int p3) {}
```

### Bad

```
void f1(int p1, int p2,
        int p3) {}

void f2(int p1,
    int p2,
    int p3) {}
```

I have created a [proof of concept](https://github.com/zkiss/checkstyle/compare/master...zkiss:params_checks) for this, that allows me to perform this check. I have implemented this using multiple `Checks` that check different aspects of the problem: alignment and parameter numbers per line.

Is this something that would make sense to add? I am not sure I will have time to properly submit a PR for this, but I would be interested in some feedback regarding if this use case is something you would consider to support and if it is worth for me to put more effort in, or if I should just package these checks in an extension jar for my own use.

Thanks,
Zoltan

konopka:
Fixes #9118


romani:
Did you consider contributing this Check to aevntu-checkstyle project?

konopka:
> Did you consider contributing this Check to aevntu-checkstyle project?

This one: https://github.com/sevntu-checkstyle/sevntu.checkstyle ? No I did not. Anyway, isn't it kind of dead with last release 2 years ago?

romani:
It is not dead.
I don't see any functional fix at https://github.com/sevntu-checkstyle/sevntu.checkstyle/commits/master/
So nothing to release for 2 years, ok :).

We keep a lot of Checks in it, to let people quickly get something to work and apply in their organization. It helps to iterate over design in few releases and not deal with breaking changes process, ... .

In main repository we are more demanding in clear design that other users should benefits also, so any hacks will not be allowed.

I will respond you on your design proposal next week.

config/checkstyle-checks.xml
```diff
@@ -558,7 +558,13 @@
     <module name="OneStatementPerLine"/>
     <module name="OverloadMethodsDeclarationOrder"/>
     <module name="PackageDeclaration"/>
+    <module name="ParameterAlignment">
+      <property name="severity" value="ignore"/>
+    </module>
     <module name="ParameterAssignment"/>
+    <module name="ParameterPlacement">
```
romani, line 565: Placing two Checks in one PR will not work well.

ParameterAlignment
ParameterPlacement

Should be separated from each other and they should make sense independently.
If not planned to be independent, they should be combined.

romani (changes requested):
Ok, let's start discussion on design.

Lets create separate issues for each Check and discuss names of check, names of parameters of checks, configs samples and behavior on exact java code.

romani:
I am closing this PR.
As we have no agreement in issue.
Even we agreed on something, each Check will be in separate PR.

Issue #15763, opened 2024-10-10
Fix all violations of JavadocParagrath check in checkstyle code base


romani:
Suppression introduced at https://github.com/checkstyle/checkstyle/pull/15686/files#r1785246044

We need to resolve all cases in our code and remove suppression.

removal of P tag from first block of text causes ` Element 'subsection' cannot have character [children],
 because the type's content type is element-only.`, it is limitation of Xdoc files that we are using, workaround is change P to DIV
see details at https://github.com/checkstyle/checkstyle/pull/15764#issuecomment-2407312827

PR #15774, opened 2024-10-14
Issue #15763: fixed paragraph tags in some more files

Zopsss:
part of #15763 

This is a pretty big PR and still only half files are covered. I will send one more PR to cover remaining files.


romani (approved):
Huge amount of work, I glanced , looks good

PR #15781, opened 2024-10-17
Issue #15541: Add new isOfType that accepts List of Integers as parameter

### LINKED ISSUE #15541: New isOfType Method that Accepts List as its Parameter
The `isOfType` method currently accepts an array of `int` as its second parameter. We should overload this method to accept a `List<Integer>`.

https://github.com/checkstyle/checkstyle/blob/8f856ecb1f2182d0b20fd22a83be115c1577e49f/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java#L285-L293

Reason: 
detected at https://github.com/checkstyle/checkstyle/pull/15505#discussion_r1725332873.
If we have `int[]` as a class field the `ImmutabilityTest` will fail as final arrays are technically mutable so another choice is to use Lists to satisfy the `ImmutabilityTest`.

ahmedyoussefg:
Closes #15541

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4990,6 +4990,9 @@
                 <param>
                   com.puppycrawl.tools.checkstyle.checks.FinalParametersCheckTest
                 </param>
+                <param>
+                  com.puppycrawl.tools.checkstyle.checks.coding.MissingSwitchDefaultCheckTest
+                </param>
               </targetTests>
               <avoidCallsTo>
                 <avoidCallsTo>com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil</avoidCallsTo>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
@@ -19,6 +19,8 @@
 
 package com.puppycrawl.tools.checkstyle.checks.coding;
 
+import java.util.Set;
+
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
@@ -87,6 +89,19 @@ public class MissingSwitchDefaultCheck extends AbstractCheck {
      * file.
      */
     public static final String MSG_KEY = "missing.switch.default";
+
+    /**
+     * Represents the possible parent tokens of a switch statement.
+     */
+    private static final Set<Integer> SWITCH_STATEMENT_PARENTS = Set.of(
+            TokenTypes.SLIST,
+            TokenTypes.LITERAL_IF,
+            TokenTypes.LITERAL_ELSE,
+            TokenTypes.LITERAL_DO,
+            TokenTypes.LITERAL_WHILE,
+            TokenTypes.LITERAL_FOR,
+            TokenTypes.LABELED_STAT
+    );
 
     @Override
     public int[] getDefaultTokens() {
@@ -176,17 +191,7 @@ public class MissingSwitchDefaultCheck extends AbstractCheck {
      * @return true if part of a switch expression
      */
     private static boolean isSwitchExpression(DetailAST ast) {
-        final int[] switchStatementParents = {
-            TokenTypes.SLIST,
-            TokenTypes.LITERAL_IF,
-            TokenTypes.LITERAL_ELSE,
-            TokenTypes.LITERAL_DO,
-            TokenTypes.LITERAL_WHILE,
-            TokenTypes.LITERAL_FOR,
-            TokenTypes.LABELED_STAT,
-        };
-
-        return !TokenUtil.isOfType(ast.getParent(), switchStatementParents);
+        return !TokenUtil.isOfType(ast.getParent().getType(), SWITCH_STATEMENT_PARENTS);
     }
 
     /**
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
@@ -27,6 +27,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.ResourceBundle;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -291,6 +292,18 @@ public final class TokenUtil {
             }
         }
         return matching;
+    }
+
+    /**
+     * Determines if the token type belongs to the given types.
+     *
+     * @param type the Token Type to check
+     * @param types the acceptable types
+     *
+     * @return true if type matches one of the given types.
+     */
+    public static boolean isOfType(int type, Set<Integer> types) {
+        return types.contains(type);
     }
 
     /**


romani:
Thanks a lot for helping us.
Please use new method at https://github.com/checkstyle/checkstyle/blob/9e7ad9e42740346ff9a8a1dfecae8d37d40df43a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java#L179

This array should be list and should be immutable list that is created ones and stored as class field.

romani:
Before pushing code , please run `mvn clean verify ` on your local , and fix all errors if any appears.

mahfouz72:
@romani  Do you have an idea why we lost some checks here? we should have ~112

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -305,6 +325,18 @@ public static boolean isOfType(DetailAST ast, int... types) {
...
+     * Determines if the AST belongs to the given types.
+     *
+     * @param ast the AST node to check
+     * @param types the acceptable types
+     *
+     * @return true if type matches one of the given types.
+     */
+    public static boolean isOfType(DetailAST ast, List<Integer> types) {
+        return ast != null && isOfType(ast.getType(), types);
+    }
```
mahfouz72, line 338: > ast != null

This condition will not get covered by real check execution so pitest will complain. The real usage of this method in the checks always sends a `ast` that is always not null so this mutation will survive.

@romani  Do you agree with having a pure JUnit for this case like in: 

https://github.com/checkstyle/checkstyle/blob/15de92ae48ddea70c478d948bf2258906eb34b74/src/test/java/com/puppycrawl/tools/checkstyle/utils/TokenUtilTest.java#L362-L383


This condition is just a safety from NPE but there is no real execution case where we send a null to this method. What do you think?

romani:
New contributor PRs are not exciting all , until I approve run. After few approvals, GitHub trust new contributors

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
romani: No pure unit testing.
All through Input files.

ahmedyoussefg: I want to confirm my understanding of the issue. The problem with the changes I made is that the `ast != null` condition is not covered during either JUnit tests or normal execution. I considered adding something like the following in the `public void testIsOfTypeFalse()` method:

```java
final DetailAstImpl astForTest3 = null; 
List<Integer> typesList = List.of(TokenTypes.LITERAL_FOR, TokenTypes.LITERAL_IF, TokenTypes.LITERAL_ELSE);
final boolean result4 = TokenUtil.isOfType(astForTest3, typesList);
assertWithMessage("Token type should not match") 
    .that(result4) 
    .isFalse(); 
```
However, since you mentioned avoiding pure unit testing, should I think of a different approach to ensure the `ast != null` condition is covered?

mahfouz72: Then there will not be a nullity check in this method. However,  This will always be checked by the caller (We are sure that the `ast` passed in this method will not be null). So IMO,  yes we can only have `public static boolean isOfType(int type, List<Integer> types)`.

romani: We are not a Library for other on universal methods. Better to keep code that we know why it exists. It in future some Check would benefit from null checking, we will add such method and it will naturally cover such edge cases.

ahmedyoussefg: We can modify the return value of `isSwitchExpression` to: 
```java
return !TokenUtil.isOfType(ast.getParent().getType(), switchStatementParents);
```
If we are certain that the parent of `ast` is always not null, otherwise, I don't currently see a way to avoid the null check

romani: If ast is not null, let's not make check for it.
Code coverage will demand coverage for all.

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
```diff
@@ -305,6 +325,18 @@ public static boolean isOfType(DetailAST ast, int... types) {
...
+    /**
+     * Determines if the AST belongs to the given types.
+     *
+     * @param ast the AST node to check
+     * @param types the acceptable types
+     *
+     * @return true if type matches one of the given types.
+     */
+    public static boolean isOfType(DetailAST ast, List<Integer> types) {
+        return isOfType(ast.getType(), types);
```
romani, line 337: Please remove this method .
No need in this abstraction layer.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
```diff
@@ -176,15 +178,15 @@ private static boolean containsNullCaseLabelElement(DetailAST detailAst) {
      * @return true if part of a switch expression
      */
     private static boolean isSwitchExpression(DetailAST ast) {
-        final int[] switchStatementParents = {
+        final List<Integer> switchStatementParents = List.of(
```
romani, line 181: Move it class field

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
ahmedyoussefg: Done

src/main/java/com/puppycrawl/tools/checkstyle/utils/TokenUtil.java
ahmedyoussefg: Done

romani (approved):
Ok to merge if CI pass.
New regression report is required.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
```diff
@@ -88,6 +90,19 @@ public class MissingSwitchDefaultCheck extends AbstractCheck {
      */
     public static final String MSG_KEY = "missing.switch.default";
 
+    /**
+     * Parents of switch statement.
+     */
```
mahfouz72, line 95: `Parents of switch statement`. => `Represents the possible parent tokens of a switch statement`

mahfouz72 (changes requested):
Minor to improve Javadoc clarity.

mahfouz72:
@ahmedyoussefg  Please see [workflows readme](https://github.com/checkstyle/checkstyle/blob/master/.github/workflows/README.md#regression-report-workflow---regression-reportyml) to learn how to generate regression report

ahmedyoussefg:
Github, generate report for MissingSwitchDefault/all-examples-in-one

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
ahmedyoussefg: Done

mahfouz72 (approved):
Thanks a lot!!

mahfouz72:
@ahmedyoussefg we need to add this check tests as a target tests in pitest utils profile in pom.xml

@romani Ideally we should delete UtilTest classs as all of them are pure junit and add all check tests as a target tests in pitest-utils profile

romani:
We are in transition from pure unit testing to BDD. Pure unit testing is allowed but requires special comment to explain why BDD is not possible.
But a lot needs to be refactored to remove examples of bad testing approaches.

pom.xml
```diff
@@ -4990,6 +4990,9 @@
                 <param>
                   com.puppycrawl.tools.checkstyle.checks.FinalParametersCheckTest
                 </param>
+                <param>
+                  com.puppycrawl.tools.checkstyle.checks.coding.MissingSwitchDefaultCheckTest
+                </param>
               </targetTests>
```
ahmedyoussefg, line 4996: like this ? @mahfouz72

mahfouz72: yes pitest should be green now.
@romani please approve other workflows

romani: Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
```diff
@@ -88,6 +90,19 @@ public class MissingSwitchDefaultCheck extends AbstractCheck {
      */
     public static final String MSG_KEY = "missing.switch.default";
 
+    /**
+     * Represents the possible parent tokens of a switch statement.
+     */
+    private static final List<Integer> SWITCH_STATEMENT_PARENTS = List.of(
```
romani, line 96: Lets convert this type to Set. We just do search over this collection, so it will be slightly quicker.

romani (changes requested):
Item

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MissingSwitchDefaultCheck.java
ahmedyoussefg: Done

romani (approved):

romani:
@ahmedyoussefg, thanks a lot for your contribution.
Please do more.
I highly recommend to start introduction to our project by fixing issues with labels "good xxxx issue".

PR #15794, opened 2024-10-21
Issue #6207: Added XPath regression test for MutableException

akankshat05:
For #6207 
Added tests for [Mutable Exception](https://checkstyle.org/checks/design/mutableexception.html#MutableException)


romani (approved):
Ok to merge.

Thanks a lot for helping us

PR #15798, opened 2024-10-22
Issue #13999: Resolve Pitest AbstractJavadocCheck mutation-1 suppression

chris-carneiro:
Part of Issue #13999
 
[AbstractJavadocCheck Documentation](https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck.html)

Mutation - 1
https://github.com/checkstyle/checkstyle/blob/42f3c4f33790050b75341b6ad1c5afe0083e762c/config/pitest-suppressions/pitest-javadoc-suppressions.xml#L3-L10

Caching by column was removed , see reason at https://github.com/checkstyle/checkstyle/pull/15798#issuecomment-2429262257 and [why it works ](https://github.com/checkstyle/checkstyle/pull/15798#discussion_r1814893199)

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheckTest.java
@@ -190,6 +190,20 @@ public class AbstractJavadocCheckTest extends AbstractModuleTestSupport {
     }
 
     @Test
+    public void testCacheWithBlockCommentInSingleLineComment() throws Exception {
+        final String[] expected = {};
+        verifyWithInlineConfigParser(getPath("InputAbstractJavadocCache3.java"), expected);
+    }
+
+    @Test
+    public void testCacheWithTwoBlockCommentAtSameLine() throws Exception {
+        final String[] expected = {
+            "13: " + getCheckMessage(SummaryJavadocCheck.class, MSG_SUMMARY_FIRST_SENTENCE),
+        };
+        verifyWithInlineConfigParser(getPath("InputAbstractJavadocCache4.java"), expected);
+    }
+
+    @Test
     public void testPositionOne() throws Exception {
         JavadocCatchCheck.clearCounter();
         final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
@@ -735,7 +749,7 @@ public class AbstractJavadocCheckTest extends AbstractModuleTestSupport {
         @Override
         public void visitJavadocToken(DetailNode ast) {
             if (reportVisitJavadocToken) {
-                // We reusing messages from JavadocTypeCheck
+                // We're reusing messages from JavadocTypeCheck
                 // it is not possible to use test specific bundle of messages
                 log(ast.getLineNumber(), ast.getColumnNumber(), MSG_TAG_FORMAT, ast.getText());
             }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/abstractjavadoc/InputAbstractJavadocCache4.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/abstractjavadoc/InputAbstractJavadocCache4.java
@@ -1,0 +1,15 @@
+/*
+SummaryJavadoc
+violateExecutionOnNonTightHtml = (default)false
+forbiddenSummaryFragments = (default)^$
+period = (default).
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.javadoc.abstractjavadoc;
+
+// violation below 'First sentence of Javadoc is missing an ending period.'
+/** Summary. */ /** some text*/
+public class InputAbstractJavadocCache4 {
+}
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -3993,9 +3993,9 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck.java</fileName>
     <specifier>type.argument</specifier>
     <message>incompatible type argument for type parameter T of ThreadLocal.</message>
-    <lineContent>private static final ThreadLocal&lt;Map&lt;LineColumn, ParseStatus&gt;&gt; TREE_CACHE =</lineContent>
+    <lineContent>private static final ThreadLocal&lt;Map&lt;Integer, ParseStatus&gt;&gt; TREE_CACHE =</lineContent>
     <details>
-      found   : @Initialized @NonNull Map&lt;@Initialized @NonNull LineColumn, @Initialized @NonNull ParseStatus&gt;
+      found   : @Initialized @NonNull Map&lt;@Initialized @NonNull Integer, @Initialized @NonNull ParseStatus&gt;
       required: [extends @Initialized @Nullable Object super null (NullType)]
     </details>
   </checkerFrameworkError>
--- a/config/pitest-suppressions/pitest-javadoc-suppressions.xml
+++ b/config/pitest-suppressions/pitest-javadoc-suppressions.xml
@@ -3,15 +3,6 @@
   <mutation unstable="false">
     <sourceFile>AbstractJavadocCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheck</mutatedClass>
-    <mutatedMethod>visitToken</mutatedMethod>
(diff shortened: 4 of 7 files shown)


romani:
You need to remove suppression, https://github.com/checkstyle/checkstyle/actions/runs/11455330207/job/31871144519?pr=15798

But if you killed muratation by code modification. You probably can extend some test to validate column in violation message. Can we?
I think we have at least one Check that reports violation on column.

romani:
yes, you are right, it is cache of block Comment Nodes.
So we need two javadoc comments on single line, `/** Summary */ /** some text.*/`.

This is not realistic code, nobody writes like this, and javadoc tool from jdk will not consider it at all, so it is not javadoc comment actually.

```
$ cat Test.java 
/** Summary */ /** some text.*/
public class Test {
}

$ javadoc Test.java
Loading source file Test.java...
Constructing Javadoc information...
Standard Doclet version 11.0.14.1
Building tree for all the packages and classes...
Generating ./Test.html...
Generating ./package-summary.html...
Generating ./package-tree.html...
Generating ./constant-values.html...
Building index for all the packages and classes...
Generating ./overview-tree.html...
Generating ./index-all.html...
Building index for all classes...
Generating ./allclasses-index.html...
Generating ./allpackages-index.html...
Generating ./deprecated-list.html...
Building index for all classes...
Generating ./allclasses.html...
Generating ./allclasses.html...
Generating ./index.html...
Generating ./help-doc.html...

```

you can see that there is no "Summary"
image


so it means that `if (JavadocUtil.isJavadocComment(blockCommentNode)) {` is not correct, as it is too simple in implementation for detection of javadoc.

`/** Summary */ ` is not a javadoc comment, it is block block comment with javadoc identifier.
https://checkstyle.org/writingjavadocchecks.html#What_is_Javadoc_comment

So it means that mutation is correct, and we by nature of javadoc can not have two comments on same line, so defining column is over-caution.
So we need ` final int treeCacheKey = blockCommentNode.getLineNo();` and slight refactoring on this.

romani:
Checker is not happy https://github.com/checkstyle/checkstyle/actions/runs/11462075960/job/31892618248?pr=15798#step:6:1705

Can we fix it without suppression?

romani:
To fix it please apply patch https://github.com/checkstyle/checkstyle/actions/runs/11462075960/job/31892618248?pr=15798#step:7:8

Suppressed message is changed.

chris-carneiro:
> Can we fix it without suppression?

I'm gonna try

romani:
Xwiki failure is not related.

chris-carneiro:
> Xwiki failure is not related.

Right, seems like both failures are unrelated to the updates.
"check_issues" i understand is a false positive, the "no-error-xwiki" failure, I'll check tomorrow if I can do anything about it.

romani:
I have worries about cache, and two comments on same line, it means that we will use first comment in line but actually second is considered as javadoc. So checks will skip being executed on second.

That is not good.
We can not miss violations on second comment.

chris-carneiro:
> I have worries about cache, and two comments on same line, it means that we will use first comment in line but actually second is considered as javadoc. So checks will skip being executed on second.
> 
> That is not good. We can not miss violations on second comment.

I have addressed your concerns by adding two test cases based on my interpretation of the issue you raised. I ran both implementations (the current release and my PR) and examined the Javadoc tree structures for each.

In both implementations, I found no differences in the tree structures produced by my tests. I've documented my process and results in the following Gist:

[JavadocTreeStructures](https://gist.github.com/chris-carneiro/691d92082c40539376a9ef135f04f5f4)

To further illustrate the results, I've included a diff test output comparing the two versions. The files prefixed with current_ correspond to the released version, while those prefixed with 13999_ use a custom build that includes the updates from this PR.

Here's the visual output from the diff test for your reference:
Screenshot 2024-10-23 at 14 46 18

@romani what do you think?

romani:
Github, generate report for JavadocBlockTagLocation/all-examples-in-one

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/abstractjavadoc/InputAbstractJavadocCache4.java
```diff
@@ -0,0 +1,15 @@
...
+period = (default).
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.javadoc.abstractjavadoc;
+
+// violation below 'First sentence of Javadoc is missing an ending period.'
+/** Summary. */ /** some text*/
+public class InputAbstractJavadocCache4 {
```
romani, line 14: This is expected behavior. Good.
But need to look closer to code explain why.
As I think we will not parse second comment as it is in cache after first comment is parsed

romani (changes requested):
Rebase will fix CIs.
I do red sign just to double check. I miss something.

romani:
Please squash all commits in one.

chris-carneiro:
> Please squash all commits in one.

Oops  my bad

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/abstractjavadoc/InputAbstractJavadocCache4.java
chris-carneiro: I understand your point, I'm willing to give it a try and understand what's going on. 
I still have other stuff to work on though.

Do you think it is worth the time investment ? Or maybe the risk to apply this update is too high for the benefit ?

romani: Reason it works is because `JavadocUtil.isJavadocComment` is smart and do not consider first in line block comment a javadoc.

https://github.com/checkstyle/checkstyle/blob/a1ea36d1db4d78527e7ae2c5d0501c083fa3c586/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck.java#L318-L319


https://github.com/checkstyle/checkstyle/blob/a1ea36d1db4d78527e7ae2c5d0501c083fa3c586/src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java#L160-L162


https://github.com/checkstyle/checkstyle/blob/a1ea36d1db4d78527e7ae2c5d0501c083fa3c586/src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java#L340-L356

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheckTest.java
```diff
@@ -189,6 +189,20 @@ public void testCache() throws Exception {
...
     }
 
+    @Test
+    public void testCache2() throws Exception {
+        final String[] expected = {};
+        verifyWithInlineConfigParser(getPath("InputAbstractJavadocCache3.java"), expected);
+    }
+
+    @Test
+    public void testCache3() throws Exception {
```
romani, line 199: please do not follow our bad pattern of naming with digits.
It was lazy decision that bites after some time.
please make name wordy `testCacheWithTwoBlockCommentAtSameLine`
please do "same" for other test method and Inputs.

romani (approved):
last before merge.

romani (changes requested):
renaming item above should be fixed.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/abstractjavadoc/InputAbstractJavadocCache4.java
chris-carneiro: @romani okay thanks for the heads up. 
You found the reason for the behavior so It's good news for the current matter. 

I was currently going down the inner workings of the TreeWalker and noticed something similar. 

For future reference I would like point out that the JavaLanguageLexer#reportBlockComment shows the same behavior and caches as Javadoc only the last Javadoc block,
specifically this piece of code:

image
So basically since the map key is the line number, the last comment is overridden by the next comment found on the same line.

Anyways, I thought it was worth pointing out.

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheckTest.java
chris-carneiro: Done

chris-carneiro:
> renaming item above should be fixed.
My bad this PR took so long! Thanks for your help @romani

romani (approved):
Ok to merge if CI pass

romani:
We have lack of active maintainers now.
So merging on single review.

romani:
I hope @chris-carneiro watched power of pitest to eliminate unnecessary code.

chris-carneiro:
> I hope @chris-carneiro watched power of pitest to eliminate unnecessary code.

I will resolve 1 or 2 more suppressions to feel more confident about it. But It seems quite effective indeed

PR #15805, opened 2024-10-24
minor: added my name in google_checks.xml

Zopsss:
Mauryan Kansara - My real name
Zops - alias name which I use for all of my social medias and GitHub

### DIFF
--- a/src/main/resources/google_checks.xml
+++ b/src/main/resources/google_checks.xml
@@ -13,7 +13,7 @@
     To completely disable a check, just comment it out or delete it from the file.
     To suppress certain violations please review suppression filters.
 
-    Authors: Max Vetrenko, Ruslan Diachenko, Roman Ivanov.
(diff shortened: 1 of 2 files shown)


src/main/resources/google_checks.xml
```diff
@@ -13,7 +13,7 @@
     To completely disable a check, just comment it out or delete it from the file.
     To suppress certain violations please review suppression filters.
 
-    Authors: Max Vetrenko, Ruslan Diachenko, Roman Ivanov.
+    Authors: Max Vetrenko, Ruslan Diachenko, Roman Ivanov, Mauryan Kansara ( Zops ).
```
romani, line 16: Please keep only real name

romani (changes requested):
Items

src/main/resources/google_checks.xml
Zopsss: removed. Added my name next to Max's, maintaining a specific order of: `contributor's names` first and then `mentor's names`.

romani (approved):

PR #15825, opened 2024-10-27
Issue #6207: Added XPath regression test for ClassTypeParameterName

akankshat05:
For #6207

### DIFF
--- a/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionClassTypeParameterNameTest.java
+++ b/src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionClassTypeParameterNameTest.java
@@ -1,0 +1,128 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code and other text files for adherence to a set of rules.
+// Copyright (C) 2001-2024 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


src/it/resources/org/checkstyle/suppressionxpathfilter/classtypeparametername/InputXpathClassTypeParameterNameUpperCase.java
```diff
@@ -0,0 +1,7 @@
+package org.checkstyle.suppressionxpathfilter.classtypeparametername;
+
+public class InputXpathClassTypeParameterNameUpperCase {
+    class MyClass1<aBc> {} // warn
+    class MyClass2<LISTENER> {}
```
romani, line 5: Unfortunately all examples are the same by structure.
In xpath value is in different structure of xpath expressions.
Please make second example to be classes in interface.
Third example is classes in methods.

romani (changes requested):
Items

src/it/java/org/checkstyle/suppressionxpathfilter/XpathRegressionClassTypeParameterNameTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+        runVerifications(moduleConfig, fileToProcess, expectedViolation,
+                expectedXpathQueries);
+    }
+
+    @Test
+    public void testUpperCase() throws Exception {
+        final String pattern = "^[A-Z]{2,}$";
+        final File fileToProcess =
+                new File(getPath("InputXpathClassTypeParameterNameUpperCase.java"));
```
romani, line 73: Please update input file names and test methods to reflect structure content file, not a specificof name of class.

romani (changes requested):
I actually meant that we need to fix.

romani (approved):
Thanks a lot for your help!!
Ok to merge

PR #14153, opened 2023-12-18
Issue #11584: WriteTag reports violation with confusing message when …

aayushRedHat:
Related: https://github.com/checkstyle/checkstyle/issues/11584 

Continuation of abandoned PR https://github.com/checkstyle/checkstyle/pull/12819

Diff regression config: https://gist.githubusercontent.com/strkkk/238b6c97492ff6ffdb3be1603deca8fc/raw/5db22461435a94c56a3f420be844cd07baf497f2/config_single_2.xml

PR #15844, opened 2024-11-01
Issue #11584: WriteTag reports violation with confusing message when …

### LINKED ISSUE #11584: WriteTag reports violation with confusing message when there is no javadoc

MohanadKh03:
Fixes #11584 
Extends #14153
Diff regression config: https://gist.githubusercontent.com/strkkk/238b6c97492ff6ffdb3be1603deca8fc/raw/5db22461435a94c56a3f420be844cd07baf497f2/config_single_2.xml

### DIFF
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagRecordsAndCompactCtors.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagRecordsAndCompactCtors.java
@@ -12,7 +12,7 @@ tokens = INTERFACE_DEF, CLASS_DEF, ENUM_DEF, ANNOTATION_DEF, RECORD_DEF, COMPACT
 package com.puppycrawl.tools.checkstyle.checks.javadoc.writetag;
 
 
-public class InputWriteTagRecordsAndCompactCtors { // violation 'missing @incomplete tag.'
+public class InputWriteTagRecordsAndCompactCtors {
 
     // violation 2 lines below 'Type Javadoc tag @incomplete must match pattern '\\S''
     /**
@@ -30,7 +30,7 @@ public class InputWriteTagRecordsAndCompactCtors { // violation 'missing @incomp
     }
 
 
-    record MyRecord2(String myString) { // violation 'missing @incomplete tag.'
+    record MyRecord2(String myString) {
 
         // violation 2 lines below 'Failed to recognize 'record' introduced in Java 14.'
         /**
@@ -41,7 +41,7 @@ public class InputWriteTagRecordsAndCompactCtors { // violation 'missing @incomp
     }
 
 
-    record MyRecord3(int x) { // violation 'Type Javadoc comment is missing @incomplete tag.*'
+    record MyRecord3(int x) {
 
         // violation 2 lines below 'Failed to recognize 'record' introduced in Java 14.'
         /**
@@ -53,9 +53,9 @@ public class InputWriteTagRecordsAndCompactCtors { // violation 'missing @incomp
     }
 
 
-    record MyRecord4(int y) { // violation 'Type Javadoc comment is missing @incomplete tag.*'
+    record MyRecord4(int y) {
 
-        private record MyRecord5(int z) { // violation 'missing @incomplete tag.'
+        private record MyRecord5(int z) {
 
             // violation 2 lines below 'Failed to recognize 'record' introduced in Java 14.'
             /**
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagResetSeverity.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagResetSeverity.java
@@ -1,0 +1,32 @@
+package com.puppycrawl.tools.checkstyle.checks.javadoc.writetag;
+
+/**
+ * Testing tag writing
+ * @author Mohanad
+ * @author Another Author
+ * @incomplete This class needs more code...
+ * @doubletag first text
+ * @doubletag second text
+ * @emptytag
+ */
+class InputWriteTagResetSeverity
+{
+    /**
+     * @todo Add a constructor comment
+     */
+    public InputWriteTagResetSeverity()
+    {
+    }
+
+    public void method()
+    {
+    }
+
+    /**
+     * @todo Add a comment
+     */
+    public void anotherMethod()
+    {
+    }
+}
+
--- a/config/jsoref-spellchecker/whitelist.words
+++ b/config/jsoref-spellchecker/whitelist.words
@@ -840,6 +840,7 @@ mockit
 mockito
 modifiedcontrolvariable
 modifierorder
+Mohanad
 Mohnen
 mojohaus
 moks
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/WriteTagCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/WriteTagCheck.java
@@ -36,6 +36,7 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
  * Requires user defined Javadoc tag to be present in Javadoc comment with defined format.
  * To define the format for a tag, set property tagFormat to a regular expression.
  * Property tagSeverity is used for severity of events when the tag exists.
+ * No violation reported in case there is no javadoc.
  * </div>
  *
  * <ul>
@@ -197,10 +198,7 @@ public class WriteTagCheck
         final int lineNo = ast.getLineNo();
         final TextBlock cmt =
             contents.getJavadocBefore(lineNo);
-        if (cmt == null) {
-            log(lineNo, MSG_MISSING_TAG, tag);
-        }
-        else {
+        if (cmt != null) {
             checkTag(lineNo, cmt.getText());
         }
     }
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/javadoc/WriteTagCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/javadoc/WriteTagCheck.xml
@@ -8,6 +8,7 @@
  Requires user defined Javadoc tag to be present in Javadoc comment with defined format.
  To define the format for a tag, set property tagFormat to a regular expression.
  Property tagSeverity is used for severity of events when the tag exists.
+ No violation reported in case there is no javadoc.
  &lt;/div&gt;</description>
          <properties>
             <property name="tag" type="java.lang.String">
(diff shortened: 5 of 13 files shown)


MohanadKh03:
GitHub, generate website

romani:
Github, generate report for WriteTag/all-examples-in-one

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagRecordsAndCompactCtors.java
```diff
@@ -12,7 +12,7 @@
 package com.puppycrawl.tools.checkstyle.checks.javadoc.writetag;
 
 
-public class InputWriteTagRecordsAndCompactCtors { // violation 'missing @incomplete tag.'
+public class InputWriteTagRecordsAndCompactCtors {
```
romani, line 15: Config is not default.
There should be no changes.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/Example5.txt
```diff
@@ -0,0 +1,16 @@
...
+    <module name="WriteTag"/>
+  </module>
+</module>
+*/
+
+// xdoc section -- start
+class InputWriteTagNoJavadoc
+{
+    public void method()
+    {
```
romani, line 13: Please put `{` on same line with declaration of class and method.
It will make example compact, and it is most usual style.

romani (changes requested):
Items

MohanadKh03:
So resetting the severity level back to the original level *line 251* here causes a VoidMethodCallMutator in pitest
https://github.com/checkstyle/checkstyle/blob/c289e27897e27b14a1e22aaab72d1ed3cdbcb58d/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/WriteTagCheck.java#L245-L251
It seems this is the only place that resets the severity level in the whole codebase so the easy solution would be to just suppress this mutation

To actually solve this though it would be kind of a big change 
Currently all violations are tested on their line number and violation message in all the ``verify`` methods and ``verifyViolations`` in ``AbstractModuleTestSupport`` 

We could try to add a way for verifying the expected vs actual severity level of violations also but I think this is too big a change for a very rare case like this ``WriteTag`` one which could just be suppressed 

e.g: I thought that this was testing the severity level but it is actually testing whether it will break or not not the severity level being the same or not at the end of the check
https://github.com/checkstyle/checkstyle/blob/c289e27897e27b14a1e22aaab72d1ed3cdbcb58d/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/WriteTagCheckTest.java#L124-L131 

I am rooting for just suppressing this mutation since this is a very rare case that only happens in this check. 
WDYT ? @romani

romani:
No suppressions, if rare it ok, just need test on it.

romani:
https://checkstyle.org/config.html#Severity
Vs
https://checkstyle.org/checks/javadoc/writetag.html#Properties tagSeverity

Do you see a reason why we have this special property in this Check?

Sounds like we can drop such special property as all Checks already have it.

MohanadKh03:
> It is always easy to ignore. But whole our project doesn't not ignore quality metrics.
No suppressions, if rare it ok, just need test on it.

I mean it is kind of a big change we would need to add a new ``verify`` method because currently there is no way to add a new test that would compare the expected vs actual severity level. We can not just add a new test in the suite for the check because there is no way we can test for the severity level with the way ``verify`` and ``verifyViolations`` work rn , but a big change to the verification methods to add a way in here or just add a new method like ``verifyViolations`` which would be breaking compatability i think ?
https://github.com/checkstyle/checkstyle/blob/d7c12dad6929787d36439dbb2da65763d28ab3e0/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java#L512-L531
in order to enable us to check for severity of violation (expected vs actual) we need to have a way to send our expected severity level vs the actual one after using the check so either change the above code snippet or make a new one entirely 


> https://checkstyle.org/config.html#Severity
Vs
https://checkstyle.org/checks/javadoc/writetag.html#Properties tagSeverity

> Do you see a reason why we have this special property in this Check?

Honestly it doesnt seem to bring up any benefit to change the severity level for this check idk why this check needs this property probably better off without it


> We have 1  logTag( that plays with severity and 2 log that use severity defined in usual way (common for all Checks).
Ideally we should just change default behavior in this PR, and do removal of property in separate issue if we see no reason to have it Check.

Sure but for this PR's changes to pass we would need to fix the mutation of the changed severity level thing so should we start with the removal of this property first then continue here ? Most likely there is no more change in this PR it handles the default behaviour rn the only thing missing is fixing the mutation *which would not be needed if the property is removed altogether*

rnveach:
Purpose is to give a different severity between a tag format match versus everything else this check logs.

MohanadKh03:
> Purpose is to give a different severity between a tag format match versus everything else this check logs.
Ok makes sense . Ill try to fix its mutation then

romani:
there are a lot of complains on this Check - https://github.com/checkstyle/checkstyle/issues?q=is%3Aissue+is%3Aopen+tagSeverity

I pretty sure we will remove this property from this Check in future (breaking compatibility). But let not do this in this PR.

rnveach:
2 issues (now 1 since you closed the other) and 6 closed is not a lot.

This property has a purpose to this check, so if you remove it, you remove the purpose of this check. You might as well remove the check completely then.

If your complaint is with 2 issues, doesn't make sense why we keep Indentation which is a worse check then this with 75 issues https://github.com/checkstyle/checkstyle/issues?q=is%3Aopen+is%3Aissue+label%3Aindentation (67 closed).

romani:
@MohanadKh03 , do you need help with mutation ?

MohanadKh03:
After some trial-and-error it seems the only possible way is to add a way to validate the expected vs actual severity levels but currently it is not possible 
We can add a new test that checks that with the default configs there is no change to the severity level before and after processing check to ensure this doesnt happen in other future checks as well we can make it inside XpathRegressionTest for example ?
Yes if you have a better idea please let me know @romani

romani:
Ok, our common testing but Inputs with config inside Input will not work here.
It is ok. I confirming exceptional case.

We can do it in low level way as we did long time ago.
We can setup config by means of beans and receive violations from any logger.

Example of how to run checker https://github.com/checkstyle/checkstyle/blob/9643e38b5097d27f0061ef4c6106643eff4cbfd5/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java#L918

This is example how to set listener https://github.com/checkstyle/checkstyle/blob/9643e38b5097d27f0061ef4c6106643eff4cbfd5/src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java#L408
We can use default audit listener (logger). And just get fact that violation is wrong severity.
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/DefaultLogger.java

And we just need to add above test method a comment that briefly explains reasons of doing testing so low level.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagRecordsAndCompactCtors.java
romani: we do update "No violation reported in case there is no javadoc."
so no violation here is ok

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagResetSeverity.java
```diff
@@ -0,0 +1,42 @@
+/*
+WriteTag
+tag = @author
+tagFormat = Mohanad
+tagSeverity = warning
```
romani, line 5: please remove this to avoid any confusion that this Input is using config from Input file.
It is exceptional Input, it should looks like exceptional.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/Example5.txt
romani: instead of writing new Example.

Please add `public void method() {}` without javadoc to all above Examples.

```
/**
* Some class
*/
public class Test {
  /** some doc */
  void foo() {}

  public void method() {}
}
  
```

romani (changes requested):
good!!

last items:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/writetag/InputWriteTagResetSeverity.java
MohanadKh03: done

romani:
Windows CI problem:
```
[ERROR]   WriteTagCheckTest.testResetSeverityLevel:218 Second violation's severity level should have been reset back to default (error)
value of: toLowerCase(...)
expected: error
but was : writetag
[INFO] 
[ERROR] Tests run: 5148, Failures: 1, Errors: 0, Skipped: 337
```
https://ci.appveyor.com/project/Checkstyle/checkstyle/builds/50955311/job/lkjqq6iv4503lech#L1367

Most likely due to different end lines symbols.

MohanadKh03:
@romani
The ```checkstyle.checkstyle (Job test-ru)``` seems to have somehow got stuck ?

romani (approved):
ok to merge.

This check is very bad design state, We are moving forward with step by step deprecation of it.
We can not remove it as it is the only Check that user can use to demand some tag to be present in all javadocs.
In future it some other  Check will do this and we will remove WriteTag competely.

Issue #14814, opened 2024-04-20
Resolve `TailRecursion` inspection violations by replacing tail recursive calls


MANISH-K-07:
Follow-up of #14625 

Version bump of IDEA to v2022.3.3 resulted in successful build with inspection problems that have been suppressed for migration :
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/1/workflows/0968f80e-db8e-401a-ad87-8c409072ba31/jobs/574793/artifacts

> In general recursion is not good and prone to failing eventually, some times in several decades :).

`TailRecursion` inspection problems need to be dealt with on a case by case basis.

```
<problems is_local_tool="true">
<problem>
<file>file://$PROJECT_DIR$/src/main/java/com/puppycrawl/tools/checkstyle/api/FullIdent.java</file>
<line>92</line>
<module>project</module>
<package>com.puppycrawl.tools.checkstyle.api</package>
<entry_point TYPE="method" FQNAME="com.puppycrawl.tools.checkstyle.api.FullIdent void extractFullIdent(com.puppycrawl.tools.checkstyle.api.FullIdent full, com.puppycrawl.tools.checkstyle.api.DetailAST ast)"/>
<problem_class id="TailRecursion" severity="ERROR" attribute_key="ERRORS_ATTRIBUTES">Tail recursion</problem_class>
<description>Tail recursive call <code>extractFullIdent()</code> #loc</description>
<highlighted_element>extractFullIdent</highlighted_element>
<language>JAVA</language>
<offset>16</offset>
<length>16</length>
</problem>

<problem>
<file>file://$PROJECT_DIR$/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/HiddenFieldCheck.java</file>
<line>656</line>
<module>project</module>
<package>com.puppycrawl.tools.checkstyle.checks.coding</package>
<entry_point TYPE="method" FQNAME="com.puppycrawl.tools.checkstyle.checks.coding.HiddenFieldCheck.FieldFrame boolean containsInstanceField(java.lang.String field)"/>
<problem_class id="TailRecursion" severity="ERROR" attribute_key="ERRORS_ATTRIBUTES">Tail recursion</problem_class>
<description>Tail recursive call <code>containsInstanceField()</code> #loc</description>
<highlighted_element>containsInstanceField</highlighted_element>
<language>JAVA</language>
<offset>30</offset>
<length>21</length>
</problem>

<problem>
[...]

PR #15881, opened 2024-11-06
minor: Removed tail recursion

mohitsatr:
part of #14814


romani:
next time please keep in commit reference to issue, it helps to investigate code by git history.

mohitsatr:
got it

Issue #13494, opened 2023-07-31
https://checkstyle.org/checks/ Gives a 404


rnveach:
https://checkstyle.org/checks/
````
File not found

The site configured at this address does not contain the requested file.

If this is your site, make sure that the filename case matches the URL as well as any file permissions.
````

We are missing an index file.

There is no link to this page that I am aware. I just noticed there was no index.

---------

All our issues have links to Checks and all of them are invalid.
I think we need to set redirects 
Of https://checkstyle.org/config_javadoc.html#JavadocType
To
https://checkstyle.org/checks/javadoc/index.html or better to https://checkstyle.org/checks/javadoc/javadoctype.html

Redirects in GitHub io: https://dev.to/steveblue/setup-a-redirect-on-github-pages-1ok7 is simply html page with redirect link.
https://gist.github.com/domenic/1f286d415559b56d725bee51a62c24a7

Might be good - Redirects of all non existing pages to some predefined https://stackoverflow.com/a/66716264/1015848

------
we need to fix all links to Checks from old url to new url
if you look at any issue before 2023, like https://github.com/checkstyle/checkstyle/issues/13155 (see https://checkstyle.org/config_coding.html#VariableDeclarationUsageDistance)
you can see link to Check , but it is not working anymore
as we changed layout of website
whole our issue tracker and all links is whole web become invalid
for exmaple in Stackoverflow.com
task is put page https://checkstyle.org/config_coding.html
that is taking archor from url #VariableDeclarationUsageDistance
that auto redirect https://checkstyle.org/config_coding.html#VariableDeclarationUsageDistance to https://checkstyle.org/checks/coding/variabledeclarationusagedistance.html#VariableDeclarationUsageDistance
so all links in web (forums, ....) become good again

rnveach:
I recommend just to redirect to https://checkstyle.org/checks.html .

romani:
We will see how to do this

Some redirection:
https://github.com/checkstyle/checkstyle/blob/2bd1a8bb63bcf0f4e5e90fb52e4c5aaccf53657b/.ci/release-copy-github-io-to-sourceforge.sh#L62

romani:
All our issues have links to Checks and all of them are invalid.
I think we need to set redirects 
Of https://checkstyle.org/config_javadoc.html#JavadocType
To
https://checkstyle.org/checks/javadoc/index.html

harshit4311:
I would love working on this issue! Could you please assign it to me?

[...]

PR #15949, opened 2024-11-21
Issue #13494: fixed https://checkstyle.org/checks/ & config_xxxx URLs giving 404 error

Zopsss:
Fixes: #13494 

added `index.xml` under `.../xdocs/checks` directory, after running `mvn clean site -Pno-validations` locally, I was able to generate `index.html` file under `.../xdocs/checks` directory. Path of that index.html file: `http://localhost:63342/checkstyle/target/site/checks/index.html`

Created `index.xml` instead of `index.html` because `index.html` gets ignored while generating the site using `mvn clean site -Pno-validations` so it does not appears in the final generated html files. Generating html files through xml files also supports `../checks.html` format.

### DIFF
--- a/src/xdocs/config_coding.xml
+++ b/src/xdocs/config_coding.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+
+  <head>
+    <title>Redirecting to checks/coding.html</title>
+  </head>
+
+  <body>
+    <section name="Coding Checks">
+      <p>Coding Checks</p>
+    </section>
+  </body>
+</document>
--- a/src/site/resources/js/checkstyle.js
+++ b/src/site/resources/js/checkstyle.js
@@ -17,6 +17,38 @@ window.addEventListener("load", function () {
     });
 });
 
+window.addEventListener("load", function () {
+    const currentUrl = window.location.href;
+
+    if (currentUrl.endsWith("/checks/") || currentUrl.endsWith("/checks/index.html")) {
+        window.location.replace("../checks.html");
+    }
+    else if (document.title.startsWith("checkstyle – Redirecting to checks/")) {
+        const urlObj = new URL(currentUrl);
+        const pathSegments = urlObj.pathname.split("/");
+        const configHtmlFile = pathSegments[pathSegments.length - 1];
+        const checkType = /config_([a-z]+).html/.exec(configHtmlFile)[1];
+        const checkName = urlObj.hash.substring(1).toLowerCase();
+
+        if (checkName) {
+            if (checkType !== "filters" && checkType !== "filefilters") {
+                window.location.replace(`./checks/${checkType}/${checkName}.html`);
+            }
+            else {
+                window.location.replace(`./${checkType}/${checkName}.html`);
+            }
+        }
+        else {
+            if (checkType !== "filters" && checkType !== "filefilters") {
+                window.location.replace(`./checks/${checkType}`);
+            }
+            else {
+                window.location.replace(`./${checkType}`);
+            }
+        }
+    }
+});
+
 window.addEventListener("scroll", function () {
     "use strict";
     if (document.documentElement.scrollTop > scrollDistanceToButtonVisibility) {
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsJavaDocsTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsJavaDocsTest.java
@@ -133,7 +133,8 @@ public class XdocsJavaDocsTest extends AbstractModuleTestSupport {
             final String fileName = file.getName();
 
             if ("config_system_properties.xml".equals(fileName)
-                    || "index.xml".equals(fileName)) {
+                    || "index.xml".equals(fileName)
+                    || Pattern.matches("config_[a-z]+.xml", fileName)) {
                 continue;
             }
 
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
@@ -277,10 +277,15 @@ public class XdocsPagesTest {
                     .replace(".xml", ".html")
                     .replaceAll("\\\\", "/")
                     .replaceAll("src[\\\\/]xdocs[\\\\/]", "");
-            final String expectedLink = String.format(Locale.ROOT, "href=\"%s\"", expectedFile);
-            assertWithMessage("Expected to find link to '" + expectedLink + "' in " + SITE_PATH)
-                    .that(siteContent)
-                    .contains(expectedLink);
+            final boolean isConfigHtmlFile = Pattern.matches("config_[a-z]+.html", expectedFile);
+            final boolean isChecksIndexHtmlFile = "checks/index.html".equals(expectedFile);
+
+            if (!isConfigHtmlFile && !isChecksIndexHtmlFile) {
+                final String expectedLink = String.format(Locale.ROOT, "href=\"%s\"", expectedFile);
+                assertWithMessage("Expected to find link to '" + expectedLink + "' in " + SITE_PATH)
+                        .that(siteContent)
+                        .contains(expectedLink);
+            }
         }
     }
 
@@ -550,7 +555,8 @@ public class XdocsPagesTest {
             final String fileName = path.getFileName().toString();
 
             if ("config_system_properties.xml".equals(fileName)
-                    || "index.xml".equals(fileName)) {
+                    || "index.xml".equals(fileName)
+                    || Pattern.matches("config_[a-z]+.xml", fileName)) {
                 continue;
             }
 
--- a/src/xdocs/checks/index.xml
+++ b/src/xdocs/checks/index.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+
+  <head>
+    <title>Redirecting to checks.html</title>
+  </head>
+
+  <body>
+    <section name="Standard Checks">
+      <p>Standard Checks</p>
+    </section>
+  </body>
+</document>
--- a/src/xdocs/config_annotation.xml
+++ b/src/xdocs/config_annotation.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+
+  <head>
+    <title>Redirecting to checks/annotation.html</title>
+  </head>
+
+  <body>
+    <section name="Annotations Checks">
+      <p>Annotations Checks</p>
+    </section>
+  </body>
+</document>
--- a/src/xdocs/config_blocks.xml
+++ b/src/xdocs/config_blocks.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+
+  <head>
+    <title>Redirecting to checks/blocks.html</title>
+  </head>
+
+  <body>
+    <section name="Block Checks">
+      <p>Block Checks</p>
+    </section>
+  </body>
+</document>
--- a/src/xdocs/config_design.xml
+++ b/src/xdocs/config_design.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+
+  <head>
+    <title>Redirecting to checks/design.html</title>
+  </head>
+
+  <body>
+    <section name="Design Checks">
+      <p>Design Checks</p>
+    </section>
+  </body>
+</document>
--- a/src/xdocs/config_filefilters.xml
+++ b/src/xdocs/config_filefilters.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+
+  <head>
+    <title>Redirecting to checks/filefilters.html</title>
+  </head>
+
+  <body>
+    <section name="File Filters Checks">
+      <p>File Filters Checks</p>
+    </section>
+  </body>
+</document>
--- a/src/xdocs/config_filters.xml
+++ b/src/xdocs/config_filters.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+
+  <head>
+    <title>Redirecting to checks/filters.html</title>
+  </head>
+
+  <body>
+    <section name="Filters Checks">
+      <p>Filters Checks</p>
+    </section>
+  </body>
+</document>
--- a/src/xdocs/config_headers.xml
+++ b/src/xdocs/config_headers.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+
+  <head>
+    <title>Redirecting to checks/headers.html</title>
+  </head>
+
+  <body>
+    <section name="Headers Checks">
+      <p>Header Checks</p>
+    </section>
+  </body>
+</document>
--- a/src/xdocs/config_imports.xml
+++ b/src/xdocs/config_imports.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+
+  <head>
+    <title>Redirecting to checks/imports.html</title>
+  </head>
+
+  <body>
+    <section name="Imports Checks">
+      <p>Imports Checks</p>
+    </section>
+  </body>
+</document>
--- a/src/xdocs/config_javadoc.xml
+++ b/src/xdocs/config_javadoc.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+
+  <head>
+    <title>Redirecting to checks/javadoc.html</title>
+  </head>
+
+  <body>
+    <section name="Javadoc Comments Checks">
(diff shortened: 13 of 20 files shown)


Zopsss:
Github, generate website

Zopsss:
Github, generate website

Zopsss:
> https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8217c35_2024101120/checks/index.html#

There's a small delay in redirecting, you will be able to see a blank page for a small fraction of time.

Also https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8217c35_2024101120/checks/ isn't working, normally JS picks up index.html as a default page but that is not happening in this case.

@romani @rnveach, seeking for your review and help

Zopsss:
> JS isn't initiated when you open a url. It is only initiated after the html page is loaded.

Yeah, how can I fix this? I mean the link isn't working with `checks/` at the end, how we should tackle this?

romani:
Links like 
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8217c35_2024101120/config_coding.html#VariableDeclarationUsageDistance

should work

there  should be page config_coding.html first of all and JS in it should do redirection after page is loaded to https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/8217c35_2024101120/checks/coding/variabledeclarationusagedistance.html#VariableDeclarationUsageDistance

Zopsss:
Github, generate website

Zopsss:
Github, generate site

Zopsss:
@romani I used `404.html` for redirection, I created a new `redirect.js` file to statically redirect user to correct file. Now the problem is that I'm not sure if this is going to work or not as I cannot test it locally and also if we cannot test it by generating website through GH action as it'll not match the url I provided in `redirect.js`

src/site/site.xml
```diff
@@ -46,6 +46,7 @@
       <!-- Will be added to every generated page -->
       <![CDATA[
         <script type="text/javascript" src="$relativePath/js/checkstyle.js"></script>
+        <script type="text/javascript" src="$relativePath/js/redirect.js"></script>
```
Zopsss, line 49: Adding this script in `404.xml` was giving error, I tried different ways but it was failing so I had to use `site.xml` to inject `redirect.js` in `404.xml`

Zopsss: redirect.js will be injected in all generated pages, so maybe we don't need 404.xml then. But I'm not sure about this, waiting for maintainers opinion.

Zopsss:
@romani @rnveach ping

src/site/site.xml
romani: Js is in all existing pages. But we hit non existing. Such page use to exist.
So we need to recreate it as xml file with empty content. js in this page will trigger redirection.

In other words, page becomes real, but after page is loaded , js do redirection based on url content.

romani: Here is old website https://checkstyle.sourceforge.io/version/10.10.0/
So you can walk around to see a structure.

You can build site on local by
```
git checkout checkstyle-10.10.0
mvn clean site -Pno-validations 
```

We need to restore config_xxxx.xml files

romani: https://github.com/checkstyle/checkstyle/releases/tag/checkstyle-10.10.0

Such files can be found at:
https://github.com/checkstyle/checkstyle/tree/226ff1c57eaaca0bf6fcbf0674aece140c31119b/src/xdocs

Zopsss: > We need to restore config_xxxx.xml files

From what I understood, are you trying to say if user comes to let's say `config_coding.html`, he should be redirected to https://checkstyle.org/checks/coding/ ?

Similarly if user comes to `config_coding#VariableDeclarationUsageDistance`, he should be redirected to https://checkstyle.org/checks/coding/variabledeclarationusagedistance.html

Im assuming, you're suggesting to restore all `config_xxx.xml` files and add redirection in them to redirect to their corresponding real page. 

But instead of restoring those files, we can add redirection in `404.xml` itself. When user will enter any `config_xxxx` url like https://checkstyle.org/config_coding.html#VariableDeclarationUsageDistance then he will hit the `404.xml` file and it will check the URL and redirect the user to its corresponding page, in this case: https://checkstyle.org/checks/coding/variabledeclarationusagedistance.html

We can add js in `404.xml` to check the URL and redirect the user, this way we will not need to restore `config_xxxx.xml` files. 

Correct me if I'm going in wrong direction.

romani: >config_coding.html, he should be redirected to https://checkstyle.org/checks/coding/ ?

Yes 

>config_coding#VariableDeclarationUsageDistance, he should be redirected to https://checkstyle.org/checks/coding/variabledeclarationusagedistance.html

Yes.

>But instead of restoring those files, we can add redirection in 404.xml itself

I doubt it will work as url will be ..../404.html , you will not have original url. But if you find a way to do it, I am ok with this.

Zopsss: @rnveach @romani I created a new [personal repo](https://github.com/Zopsss/checkstyle-test) in which I kept all the CS docs. I also added the [404.html](https://github.com/Zopsss/checkstyle-test/blob/master/404.html) for the redirection.

Here's the site link: https://zopsss.github.io/checkstyle-test/

Currently I have added redirection only for some specific URLs which you can check from 404.html, you can test the website with any of the following URLs and it'll redirect you to CS' URLs.

https://zopsss.github.io/checkstyle-test/checks/

https://zopsss.github.io/checkstyle-test/config_coding.html

https://zopsss.github.io/checkstyle-test/config_coding.html#VariableDeclarationUsageDistance

If we inject similar JS in our 404.xml then we'll be able to redirect users to the original URLs. The only downside of this approach I can see is, if the body of the 404.html is empty then it'll not show any kind of message ( like page not found ) to the user.

For example:

Screenshot 2024-11-28 121630

you can see in the screenshot that the URL is https://zopsss.github.io/checkstyle-test/something so it should give `404 page not found` error message, but the body of 404.html is empty so it does not display any kind of message.

We can fix this by designing our own error message, I would be happy to do that :)

Zopsss: **UPDATE**

I have modified the 404.html and now it contains an error message, we can add something similar to that in our 404.xml file.

Zopsss: I doubt that we can test the `404.xml` by generating website with `mvn clean site -Pno-validations` because then the URL will be different and the redirection won't happen, the only way to test is by releasing a new version.

rnveach: https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site

Zopsss: Yes, this documentation explains how we can use 404.html to show custom `404 page not found` error, just the way I used it in above comment.

romani: Push new code to let me see all details

Zopsss: I have pushed the new code, I tried testing this code locally by changing the URLs to `localhost...` but it wasn't working as expected. To make it work, we will need to hit the URL from GH Pages deployment. So to check if it works, we will need to make a new release.

Is there any other way to test it?

src/xdocs/404.xml
```diff
@@ -0,0 +1,40 @@
...
+            window.location.href = "https://checkstyle.org/checks/coding/variabledeclarationusagedistance.html";
+            break;
+        }
+      }
+    </script>
+  </head>
+
+  <body>
+    <section name="redirect">
+      <div style="display:flex;flex-direction:column;align-items:center;">
```
Zopsss, line 29: used inline CSS because internal CSS was not working, not sure why.

romani:
GitHub, generate website

src/site/site.xml
Zopsss: In the new code, I tried to add the JS by using `<script>` tag but I just noticed after generating the website, the `<script>` I added is removed:

Screenshot 2024-11-30 095135

To inject JS in `404.html`, we can use this approach: https://github.com/checkstyle/checkstyle/commit/a7f0b2b8f80b1861467b223f1f0d5a11a516f33f

add the script in `site.xml` which will inject it in all the pages. But this will downgrade the performance of our website, we dont want this JS to be injected in all HTML pages, we want to inject it only in `404.html` page. Injecting this JS in all pages means that we will execute it every time a page loads, which might affect the loading time of our website.

romani: Micro code is ok

Zopsss:
@romani currently, the 404.html will not work. Please take a look at these comments: https://github.com/checkstyle/checkstyle/pull/15949#discussion_r1864069236 https://github.com/checkstyle/checkstyle/pull/15949#discussion_r1864078539

romani:
Ok, push code that works when you have time.

src/site/site.xml
Zopsss: I have pushed the micro code, but to test it, we will need to make a new release.

romani:
GitHub, generate website

src/site/resources/js/redirect.js
```diff
@@ -0,0 +1,14 @@
+window.onload = () => {
+    const currentUrl = window.location.href;
+    switch (currentUrl) {
+        case "https://checkstyle.org/checks/":
+            window.location.href = "https://checkstyle.org/checks";
+            break;
+        case "https://checkstyle.org/config_coding.html":
+            window.location.href = "https://checkstyle.org/checks/coding/";
```
romani, line 8: Please make matching or currentUrl to match  without domain 
To make link https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/eac018a_2024165644/config_coding.html to work

romani (changes requested):
Items

src/site/resources/js/redirect.js
Zopsss: I have pushed the new code, this new code isn't working properly locally, I have just pushed it to see if it works by any chance on github or not.

Zopsss:
GitHub, generate website

Zopsss:
Only https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/6dde21b_2024190849/checks.html redirection is working properly, other urls like https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/6dde21b_2024190849/config_coding.html is not working.

Zopsss:
The main reason I can think of behind why the redirection is not working.

Suppose I opened a valid page like: `https://checkstyle.org/checks/coding/index.html`

To open this page, first my browser will try to find `index.html` inside the `coding` directory, if it finds it, it will load the HTML and JS for it. If we add any redirection on this URL then browser will be able to perform it because it was able to find and load the HTML & JS of  `index.html` file inside `coding` directory, the file actually existed. 

Now if we request for an invalid page like: `https://checkstyle.org/config_coding.html`

this HTML file does not exists, browser will try to find this file but it will fail. So no HTML or JS gets loaded and we get the 404 message.

This will not happen with GH pages deployment because we can define a 404.html page to handle our 404 errors, in this case, the `config_coding.html` URL will redirect the user to 404.html page and browser will try to load its HTML & JS. 

During loading its JS, it will check the URL and redirect the user to the original URL. 

Note that, when user gets 404 error, Only the custom error message is shown to user, his URL does not changes to `404.html`,  the URL would still be `config_coding.html` and we will be able to perform the redirection.

romani:
It is more beneficial to rely on basic technology, that will work for all hosting (we have sourceforge.io) that it still preferred website by Google search engine.

Relying on 404 feature of GitHub is sweet but we need more reliable approach.

Let's restore config_xxxx pages with empty content and doing redirection to new pages.

Zopsss:
@romani I have added some config_xxxx.xml files, they were working locally, hope they will work here too.

I haven't tested the https://checkstyle.org/checks/ locally, I'll do that soon. In meantime, please review other new config_xxxx.xml files' redirections.

src/site/site.xml
Zopsss, line 49: For the config_xxxx.xml approach, we will still require this extra JS file as we cannot write JS inside `<script>` tags in xml files. And we are importing this JS file in all the generated HTML files, I'm not sure how we can import this JS file into only required html files, i.e config_coding.html, config_header.html, etc.

Zopsss:
Github, generate website

Zopsss:
GitHub, generate website

Zopsss:
Please test the redirections with these URLs:

- https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/6ee7f4a_2024202316/config_coding.html
- https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/6ee7f4a_2024202316/config_coding.html#VariableDeclarationUsageDistance
- https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/6ee7f4a_2024202316/config_header.html
- https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/6ee7f4a_2024202316/config_header.html#RegexpHeader

The redirection takes few milliseconds so you will see a blank page for that moment.

src/xdocs/config_coding.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<document xmlns="http://maven.apache.org/XDOC/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/XDOC/2.0 https://maven.apache.org/xsd/xdoc-2.0.xsd">
+
+  <head>
+    <title>Coding</title>
+  </head>
```
romani, line 9: Can we do title  as "Redirection to xxxxx" and in our ja code skip calculations based on page title 
https://stackoverflow.com/a/1057065/1015848

romani (changes requested):
Prove of concept works , good.
Small blink of page is fine, it is a way better than 404 page. I hope we can improve later on it, to be better.

Items:

src/site/resources/js/redirect.js
```diff
@@ -0,0 +1,39 @@
+window.onload = () => {
+    const currentUrl = window.location.href;
+    const urlObj = new URL(currentUrl);
+    const pathSegments = urlObj.pathname.split("/");
+    const lastSegment = pathSegments[pathSegments.length - 1];
+    const extractedUrl =  `/${lastSegment}${urlObj.hash}`;
+
+    if (currentUrl === "https://checkstyle.org/checks/") {
```
romani, line 8: endsWith("/checks/")

src/site/resources/js/redirect.js
```diff
@@ -0,0 +1,39 @@
...
+        window.location.replace("https://checkstyle.org/checks");
+    }
+    else if (extractedUrl.match(/config_coding.html*/)) {
```
romani, line 11: Can we have map  with routing content "from-to"

"config_coding.html#VariableDeclarationUsageDistance" to "coding/variabledeclarationusagedistance.html#VariableDeclarationUsageDistance"

And we simply iterate over map entries and in case of match by endsWith choose destination.

so

src/site/resources/js/redirect.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+    else if (extractedUrl.match(/config_header.html*/)) {
+        checkConfigHeadersUrls(extractedUrl);
+    }
+}
+
+const checkConfigCodingUrls = (url) => {
+    switch (url) {
+        case "/config_coding.html":
+            window.location.replace("https://checkstyle.org/checks/coding/");
```
romani, line 22: We can not hardcode `https://checkstyle.org/`, please take it from URL object

romani (changes requested):
Items

src/site/resources/js/redirect.js
Zopsss: I have created a map which contains page title as key and its corresponding redirecting function as value, then used page title to get the redirect function from map and called it:

https://github.com/checkstyle/checkstyle/blob/c9a74285db34d2963947abf6effa07f610334335/src/site/resources/js/redirect.js#L12-L18


> And we simply iterate over map entries and in case of match by endsWith choose destination.

To use `endsWith` we will require to use if/else statement but I have used switch statements, using if/else has 2 downsides:

1. The code will become uglier, hard to read considering there will be a long list of if/else-if ladder
2. Switch statements are more efficient than if/else statements plus they make code look more clean.

So I think using switch statements is a good option for our usecase.

Zopsss: The code is looking more cleaner and maintainable with current implementation.

src/site/resources/js/redirect.js
Zopsss: done. 

Had to create `index.xml` under `checks` directory to make it work. 

https://github.com/checkstyle/checkstyle/blob/c9a74285db34d2963947abf6effa07f610334335/src/site/resources/js/redirect.js#L3-L6

src/xdocs/config_coding.xml
Zopsss: I have changed the page titles to `config_coding` & `config_header`, these title helps us in redirection. For better explanation, pls refer https://github.com/checkstyle/checkstyle/pull/15949#discussion_r1867078104

Zopsss:
> It is more beneficial to rely on basic technology, that will work for all hosting (we have sourceforge.io) that it still preferred website by Google search engine.

I don't think our current redirections will also work for sourceforge site, i.e for urls like https://checkstyle.sourceforge.io/config_coding.html or https://checkstyle.sourceforge.io/config_coding.html#VariableDeclarationUsageDistance

but I'm not sure cuz now I have implemented dynamic URLs:

https://github.com/checkstyle/checkstyle/blob/027ff10829b98f9e3abbc997324d718e9b06cd2d/src/site/resources/js/redirect.js#L22-L29

src/site/resources/js/redirect.js
Zopsss: done 

https://github.com/checkstyle/checkstyle/blob/027ff10829b98f9e3abbc997324d718e9b06cd2d/src/site/resources/js/redirect.js#L22-L29

Zopsss: thanks for the suggestion! I didn't knew such dynamic routing works with `window.location.replace()` too.

romani:
@Zopsss , please look at CI and keep fixing it, we are finishing PR.

Zopsss:
GitHub, generate website

src/xdocs/config_coding.xml
romani: We need some word marker that will help us to return quickly from redirection function, it  it executed in active page.

Zopsss: Can you elaborate it more? Im not able to understand it.

src/site/resources/js/redirect.js
Zopsss: https://github.com/checkstyle/checkstyle/blob/c8397ba9975a92001c25d9cae4232b5cc9e07685/src/site/resources/js/redirect.js#L3-L6

made the redirection relative.

src/xdocs/config_coding.xml
romani: Title should be `Redirection to coding.html`
`If (!title.startWith("Redirection to ") ) return;`

Zopsss:
Please remove the trailing `#` from the generated site URLs to to make redirection working.

src/xdocs/config_coding.xml
Zopsss: done.

src/site/resources/js/redirect.js
```diff
@@ -0,0 +1,41 @@
+window.onload = () => {
+    const currentUrl = window.location.href;
```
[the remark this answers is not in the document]
Zopsss: done.

Zopsss:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c8397ba_2024065122/config_coding.html redirects to https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c8397ba_2024065122/checks/coding/

but that page is not working. Not sure why it is happening, cuz locally everything was working fine. The https://checkstyle.org/checks/coding/ URL of main site also works fine.

and Same problem with `/config_header.html` redirection....

src/site/resources/js/redirect.js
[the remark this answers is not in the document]
Zopsss: I have checked that if the URL is for `/checks/`, or is URL for redirection page? If both are false then we will automatically exit from this JS.

Zopsss:
It seems like when we generate website using `GitHub, generate website`, the following of URLs https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c8397ba_2024065122/checks/coding/ requires `index.html` at the end. 

In other words, only the  following URL will work https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c8397ba_2024065122/checks/coding/index.html

But we don't require adding trailing `index.html` while generating the site locally as well as when the site is deployed.

romani:
Let's not let this nuance of index.html handling in AWS s3 web site to block this update.
It is ok to not be able to test this url. I can deploy it GitHub.io or source forge right before merge of PR to make sure it works. One time test at the end is ok.
Let's focus on fixing all other links.

src/site/resources/js/redirect.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+}
+
+const redirectConfigCodingUrls = (url) => {
+    switch (url) {
+        case "/config_coding.html":
+            window.location.replace("./checks/coding/");
+            break;
+        case "/config_coding.html#VariableDeclarationUsageDistance":
+            window.location.replace("./checks/coding/variabledeclarationusagedistance.html");
```
romani, line 27: Interesting idea:
If we do update like this for all links, it will be a lot, so js file will be with thousands lines, might not be quick to download 
:).

What if we just by code convert "#VariableDeclarationUsageDistance"
To "variabledeclarationusagedistance.html".
So at least js size will be small, so quick to download.

- [ ]

romani (changes requested):
Items

src/site/resources/js/redirect.js
Zopsss: Done. thanks a lot for the suggestion. Now our JS has reduced by a lot. I deleted the `redirect.js` and moved the code to `checkstyle.js` because our JS became so small.

Zopsss:
Proof that site redirections works locally:


https://github.com/user-attachments/assets/50fd9ffa-8e92-405f-9f27-309b3abdd52f


@romani @rnveach please do final review, if approved then I will add the remaining config_xxxx.xml files.

romani:
Looks good, please do for rest of pages.

Zopsss:
Github, generate website

romani:
https://romani.github.io/site/checks

https://romani.github.io/site/config_javadoc.html#JavadocType
https://romani.github.io/site/config_coding.html#VariableDeclarationUsageDistance

romani:
commands to redeploy to SF:
```
# in github.io repo with already generated website
tar cfz checkstyle.github.io-redirect.tar.gz site
scp -i ~/.ssh/private_sourceforge_key checkstyle.github.io-redirect.tar.gz   "$SF_USER"@web.sourceforge.net:/home/project-web/checkstyle
ssh -t -i ~/.ssh/private_sourceforge_key "$SF_USER"@shell.sourceforge.net create


cd /home/project-web/checkstyle
cd htdocs-version/
mkdir htdocs-10.20.3-SNAPSHOT
cd htdocs-10.20.3-SNAPSHOT
mv ../../checkstyle.github.io-redirect.tar.gz .
tar -xzvf checkstyle.github.io-redirect.tar.gz 
mv site/* .
cd /home/project-web/checkstyle
ln -f -s $(pwd)/htdocs-version/htdocs-10.20.3-SNAPSHOT $(pwd)/htdocs/version/10.20.3-SNAPSHOT
```

romani (approved):
ok to merge.

@Zopsss , thanks a lot for very good healing of our website!!!!
All links that are created by people in web for ~!0 years are back to live.

romani:
@nrmancuso , please review.

Zopsss:
Glad to help the community :)

nrmancuso (approved):
Good stuff @Zopsss

Issue #15950, opened 2024-11-21
False positive: magicnumbers with bit shift expression


ghernadi:
I have read check documentation: https://checkstyle.org/checks/coding/magicnumber.html
I have downloaded the latest cli from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
$ javac Test.java
$
```

```xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
    "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <property name="severity" value="error"/>
  <property name="fileExtensions" value="java, properties, xml"/>
  <module name="TreeWalker">
    <module name="MagicNumber">
       <property name="ignoreNumbers" value="-2,-1,0,1,2,100"/>
       <property name="ignoreHashCodeMethod" value="true"/>
       <property name="ignoreFieldDeclaration" value="true"/>
       <property name="severity" value="warning"/>
       <property name="id" value="checkstyle:magicnumber"/>
    </module>
  </module>
</module>
```
```java
public class Test {
    public static final int BIT0 = 1;
    public static final int BIT1 = 1 << 1;
    public static final int BIT2 = 1 << 2;
    public static final int BIT3 = 1 << 3;
    public static final int BIT4 = 1 << 4;
}
```
```bash
$ java -jar checkstyle-10.20.1-all.jar -c config.xml Test.java 
Starting audit...
[WARN] /var/tmp/Test.java:5:41: '3' is a magic number. [checkstyle:magicnumber]
[WARN] /var/tmp/Test.java:6:41: '4' is a magic number. [checkstyle:magicnumber]
Audit done.
```

I would have expected bit-shift operators to also be excluded from this check, just like `2 * 2 * 2` would work without checkstyle complaining...

nrmancuso:
Few things:

> I would have expected bit-shift operators to also be excluded from this check,  just like 2 * 2 * 2 would work without checkstyle complaining...

This appears to be a bug; to be clear, the check claims that it:
>Checks that there are no ["magic numbers"](https://en.wikipedia.org/wiki/Magic_number_%28programming%29) where a magic number is a numeric literal that is not defined as a constant. By default, -1, 0, 1, and 2 are not considered to be magic numbers.
[...]

PR #15954, opened 2024-11-22
doc: Fix Example3 in MagicNumberCheck Docs

mahfouz72:
detected at https://github.com/checkstyle/checkstyle/issues/15950#issuecomment-2492867453

Old configuration was not correct

```
cat src/Test.java
record MyRecord() {
    private static int myInt = 7; // ok, field declaration

    void foo() {
        int i = myInt + 1; // no violation, 1 is defined as non-magic
        int j = myInt + 8; // violation
    }
}
D:\CS\test
cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
    <module name="TreeWalker">
        <module name="MagicNumber">
            <property name="ignoreFieldDeclaration" value="false"/>
        </module>
        <module name="UnnecessaryParentheses"/>
    </module>
</module>
D:\CS\test
Starting audit...
[ERROR] D:\CS\test\src\Test.java:2:32: '7' is a magic number. [MagicNumber]
[ERROR] D:\CS\test\src\Test.java:6:25: '8' is a magic number. [MagicNumber]
Audit done.
```

### DIFF
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/magicnumber/Example3.txt
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/coding/magicnumber/Example3.txt
@@ -2,7 +2,7 @@
 <module name="Checker">
   <module name="TreeWalker">
     <module name="MagicNumber">
-        <property name="ignoreFieldDeclaration" value="false"/>
+        <property name="ignoreFieldDeclaration" value="true"/>
     </module>
   </module>
 </module>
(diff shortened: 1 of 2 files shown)


romani:
GitHub, generate website

romani:
It would help a lot and resolve all concerns if just activate Examples.
But we can do this little step.

mahfouz72:
> It would help a lot and resolve all concerns if just activate Examples.
> But we can do this little step.

yes, my thoughts to just fix this mistake now to not confuse users and leave activating examples for now, it will be fixed  in this other issue concerned with activating examples

PR #15957, opened 2024-11-23
Issue #15985: Expand Default Value of constantWaiverParentToken in MagicNumberCheck with All Operators

### LINKED ISSUE #15950: False positive: magicnumbers with bit shift expression

### LINKED ISSUE #15985: Expand Default Value of `constantWaiverParentToken` in MagicNumberCheck with All Operators
Check documentation: https://checkstyle.org/checks/coding/magicnumber.html#MagicNumber

**Rationale of the check:** 
> Checks that there are no ["magic numbers"](https://en.wikipedia.org/wiki/Magic_number_%28programming%29) where a magic number is a numeric literal that is not defined as a constant.

> constantWaiverParentToken: Specify tokens that are allowed in the AST path from the number literal to the enclosing constant definition.

The rationale of the check is to violate all magic numbers that are not defined as a constant and `constantWaiverParentToken`
specify tokens allowed in enclosing constant definition.

I suggest expanding the allow list with all operators (bitwise, comparison, arithmetic, logical, etc..) because:

- The property concerns only constant fields, which should be always ignored.
- To maintain consistency suggest ignoring all operators by default.
- This would align with the rationale that numbers inside constant fields are inherently non-magic numbers.

```Java
public class Example {
    public static final int CONSTANT = 7 * 6; // Consistency in ignoring all operators
    public static final boolean CONDITION = 7 >> 6; // Consistency in ignoring all operators
}
```

This simplifies check configuration for users and ensures that all constant fields are treated uniformly, regardless of the operators used.
Also, users have the option to override this property with any tokens that they need. with common sense, I would say that all operators should be included here.

mahfouz72:
closes #15985

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MagicNumberCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/MagicNumberCheck.java
@@ -64,20 +64,52 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * ARRAY_INIT</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#ASSIGN">
  * ASSIGN</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#BAND">
+ * BAND</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#BNOT">
+ * BNOT</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#BOR">
+ * BOR</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#BSR">
+ * BSR</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#BXOR">
+ * BXOR</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#COLON">
+ * COLON</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#DIV">
  * DIV</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#ELIST">
  * ELIST</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#EQUAL">
+ * EQUAL</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#EXPR">
  * EXPR</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#GE">
+ * GE</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#GT">
+ * GT</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LE">
+ * LE</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_NEW">
  * LITERAL_NEW</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LT">
+ * LT</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#METHOD_CALL">
  * METHOD_CALL</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#MINUS">
  * MINUS</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#MOD">
+ * MOD</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#NOT_EQUAL">
+ * NOT_EQUAL</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#PLUS">
  * PLUS</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#QUESTION">
+ * QUESTION</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#SL">
+ * SL</a>,
+ * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#SR">
+ * SR</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#STAR">
  * STAR</a>,
  * <a href="https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#TYPECAST">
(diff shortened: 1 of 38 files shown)


mahfouz72:
Github, generate report for MagicNumber/Example1

romani:
Github, generate report for MagicNumber/Example1

romani:
@mahfouz72, please try to trigger report one more time, not clear why your execution failed.

mahfouz72:
Github, generate report for MagicNumber/Example1

romani:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/2cad2ada63b12f7151ee812317c057227694c7e2_2024105158/reports/diff/Vavr/index.html#A1

```
@Test
public void shouldCheckHashCodeInLeafList() {
    HashArrayMappedTrie<Integer, Integer> trie = empty();
    trie = trie.put(0, 1).put(null, 2);  // LeafList.hash == 0
    final Option<Integer> none = trie.get(1 << 6);  // (key.hash & BUCKET_BITS) == 0
    assertThat(none).isEqualTo(Option.none());
}
```

I am not sure  that removal of this violation is good. but in the same time if there would be `*` instead of `<<`, there will be no violation, so making `<<` so special is not very good.

mahfouz72:
> kind of magic number, but it is field, so should be ignored if ignoreFieldDeclaration=true, but this report is for default config, 

This is ignored because the field is constant, the same as we ignore `/` (BASE / 5, for example), so `|` has nothing special to be a violation.

I agree with extending the allow list to make the check less strict. Users can still customize the list if needed.

please let me know your final decision.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/magicnumber/InputMagicNumberWithBitwiseOperatorsWithIgnoreFieldsFalse.java
```diff
@@ -0,0 +1,57 @@
+/*
+MagicNumber
+ignoreNumbers = (default)-1, 0, 1, 2
+ignoreHashCodeMethod = (default)false
+ignoreAnnotation = (default)false
+ignoreFieldDeclaration = true
+ignoreAnnotationElementDefaults = (default)true
+constantWaiverParentToken = (default)TYPECAST, METHOD_CALL, EXPR, ARRAY_INIT, UNARY_MINUS, \
```
romani, line 8: We changed default, but we didn't change all other  Inputs with "(default)".
Please do.
I am not sure in what issue we will do validation of default, I will investigate later on.

romani (changes requested):
Do you know what other operators are not this waiver?
I just try to formulate some reasons of this property and tokens presence in it.
To make it clear what is default and what users should do themselves.

Items:

romani:
Can we suggest user in issue a workaround by setting of this property with required tokens?

romani:
Can we suggest user in issue a workaround by setting of this property with required tokens?

romani:
Can we suggest user in issue a workaround by setting of this property with required tokens?

mahfouz72:
> Do you know what other operators are not this waiver?

There are a bunch of other operators that are not on the list such as mod operation, comparison operators, and conditional operators

Example:

```
D:\CS\test
cat src/Test.java
public class Test {

    public static final boolean BIT6 = 1 >= 3; // violation
    public static final boolean BIT7 = 1 != 3; // violation
    public static final int BIT8 = 1 % 3; // violation
    public static final int BIT9 = 1 > 2 ? 6 : 5; // 2 violations
}
D:\CS\test
Starting audit...
[ERROR] D:\CS\test\src\Test.java:3:45: '3' is a magic number. [MagicNumber]
[ERROR] D:\CS\test\src\Test.java:4:45: '3' is a magic number. [MagicNumber]
[ERROR] D:\CS\test\src\Test.java:5:40: '3' is a magic number. [MagicNumber]
[ERROR] D:\CS\test\src\Test.java:6:44: '6' is a magic number. [MagicNumber]
[ERROR] D:\CS\test\src\Test.java:6:48: '5' is a magic number. [MagicNumber]
Audit done.
Checkstyle ends with 5 errors.
```

> I just try to formulate some reasons of this property and tokens presence in it.
To make it clear what is default and what users should do themselves.

I still can't formulate reasons why we add some operators and not others. I don't think of any use case when we could violate magic numbers with one operator but not others.

The thing here is consistency. If one operator triggers a magic number violation, why should others (like addition or subtraction) not? Can we say that some operators (like arithmetic ones) are more commonly used with numbers That is why we added them in default list but others are not commonly used so we leave the option for users to add them to the list.

romani:
We can make any rule we think is reasonable. Let's drop few ideas/reasons and choose by vote what is more clear to active maintainers.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/magicnumber/InputMagicNumberWithBitwiseOperatorsWithIgnoreFieldsFalse.java
romani: https://github.com/checkstyle/checkstyle/issues/15983 is created to improve separately.

mahfouz72:
> We can make any rule we think is reasonable. Let's drop few ideas/reasons and choose by vote what is more clear to active maintainers.

moved to https://github.com/checkstyle/checkstyle/issues/15985

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/magicnumber/InputMagicNumberWithBitwiseOperatorsWithIgnoreFieldsFalse.java
romani: @mahfouz72 , please update other Inputs with default on this property to have such tokens.

romani:
@mahfouz72 , see https://github.com/checkstyle/checkstyle/pull/15957#discussion_r1874500779

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/magicnumber/InputMagicNumberWithBitwiseOperatorsWithIgnoreFieldsFalse.java
mahfouz72: This will be a huge update we have many inputs for this check :)

romani: We need it, I hope it is just copy paste

mahfouz72: Done. It was painful because I had to change row numbers in every unit test

romani: Hm, we always keep 2 empty lines in config comment to avoid line shift.

mahfouz72: The addition is 2 lines, and I can't remove the 2 empty lines in the config comment because there will be no empty lines at all and we have a test for this. 

It will be good if we had 3 empty lines not 2

romani: Ok, next time try to reuse empty line in header comment, if post.

romani (approved):
Ok to merge

nrmancuso (approved):
Thanks @mahfouz72 !

PR #16004, opened 2024-12-05
Issue #6207: Expand XPath IT Regression Testing for HideUtilityClassConstructor

luvtyping:
Issue #6207 
Add XPath IT Regression Testing for HideUtilityClassConstructor


romani (approved):
I do not know how to make xpath to be different, as vilation happens only on very specific case.

Issue #14881, opened 2024-05-16
Refactor SummaryJavadoc to avoid null usage 


romani:
Refactor as discussed at https://github.com/checkstyle/checkstyle/pull/14690#discussion_r1602405976

nrmancuso:
https://github.com/checkstyle/checkstyle/pull/14690#discussion_r1601538803

romani:
Yes, try to remove Optional if null is required downstream.
But may be we can keep optional and remove usage of null down in the logic.

Send PR and we will see together.

mohitsatr:
@romani, just out of curiosity, why are we removing null usages? Is it because avoiding null is generally considered best practice, or is there a specific reason in this particular context?

romani:
It is fashion of current time to use Optional instead null. And it is more explicit for reader of code that something is not guaranteed to exist.
Checker framework also helps help to explicitly mark what is nullable.

Most ideal case, no null and no Optional :),

romani:
@mohitsatr , thanks a lot.

PR #16012, opened 2024-12-08
Issue #14881: remove null from visitJavaDocToken in SummaryJavadocCheck

mohitsatr:
part of #14881 : remove null usage from `visitJavaDocToken`

### DIFF
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -4330,28 +4330,6 @@
 
   <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java</fileName>
-    <specifier>argument</specifier>
-    <message>incompatible argument for parameter javadocInlineTag of SummaryJavadocCheck.isInlineReturnTag.</message>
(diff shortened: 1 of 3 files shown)


romani:
code became less readable, please consider update like https://github.com/checkstyle/checkstyle/pull/16014/files

mohitsatr:
`.orElseThrow()` is what other methods in codebase are using, to get the value. 


https://github.com/checkstyle/checkstyle/blob/4763b54bb77ca4fd1bc04db38430ac11c5c010d3/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheck.java#L558-L559

https://github.com/checkstyle/checkstyle/blob/4763b54bb77ca4fd1bc04db38430ac11c5c010d3/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyBlockCheck.java#L176-L177

as pointed here https://github.com/gaul/modernizer-maven-plugin/issues/300#issuecomment-2526350760, there is no difference between the two.
i will with `.orElseThrow`

romani:
>there is no difference between the two.
.orElseThrow or .get() with suppression ?

Not different , but ugly to read.
So I am more and more against this validation in our project.

romani:
Restarted, passed.

mohitsatr:
> Do you see that code becomes more explicit without null ?

sure. now i see it. null was totally unnecessary there. but my inexperienced eyes wouldn't have told that in first glance.

PR #15965, opened 2024-11-27
Issue #14881: refactor summaryJavaDocCheck

mohitsatr:
resolves : #14881 

- remove null usage from `getFirstSentenceOrNull` & `findSentenceEndingOrNull`

PR #16020, opened 2024-12-11
Issue #14881: remove null from ValidateUntaggedSummary in SummaryJavadocCheck

mohitsatr:
part of #14881 : remove null usage from methods : `untaggedSummary`, `findSentenceEndingOrNull` , `getFirstSentenceOrNull`

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
@@ -27,8 +27,6 @@ import java.util.Optional;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
-import javax.annotation.Nullable;
-
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailNode;
 import com.puppycrawl.tools.checkstyle.api.JavadocTokenTypes;
@@ -240,12 +238,15 @@ public class SummaryJavadocCheck extends AbstractJavadocCheck {
         }
         else if (!period.isEmpty()) {
             if (summaryDoc.contains(period)) {
-                final String firstSentence = getFirstSentenceOrNull(ast, period);
-                if (firstSentence == null) {
-                    log(ast.getLineNumber(), MSG_SUMMARY_FIRST_SENTENCE);
+                final Optional<String> firstSentence = getFirstSentence(ast, period);
+
+                if (firstSentence.isPresent()) {
+                    if (containsForbiddenFragment(firstSentence.get())) {
+                        log(ast.getLineNumber(), MSG_SUMMARY_JAVADOC);
+                    }
                 }
-                else if (containsForbiddenFragment(firstSentence)) {
-                    log(ast.getLineNumber(), MSG_SUMMARY_JAVADOC);
+                else {
+                    log(ast.getLineNumber(), MSG_SUMMARY_FIRST_SENTENCE);
                 }
             }
             else {
@@ -600,28 +601,30 @@ public class SummaryJavadocCheck extends AbstractJavadocCheck {
     }
 
     /**
-     * Finds and returns the first sentence.
+     * Finds the first sentence.
      *
      * @param ast The Javadoc root node.
      * @param period The configured period symbol.
-     * @return The first sentence up to and excluding the period, or null if no ending was found.
+     * @return An Optional containing the first sentence
+     *     up to and excluding the period, or an empty
+     *     Optional if no ending was found.
      */
-    @Nullable
-    private static String getFirstSentenceOrNull(DetailNode ast, String period) {
+    private static Optional<String> getFirstSentence(DetailNode ast, String period) {
         final List<String> sentenceParts = new ArrayList<>();
-        String sentence = null;
+        Optional<String> result = Optional.empty();
         for (String text : (Iterable<String>) streamTextParts(ast)::iterator) {
-            final String sentenceEnding = findSentenceEndingOrNull(text, period);
-            if (sentenceEnding != null) {
-                sentenceParts.add(sentenceEnding);
-                sentence = String.join("", sentenceParts);
+            final Optional<String> sentenceEnding = findSentenceEnding(text, period);
+
+            if (sentenceEnding.isPresent()) {
+                sentenceParts.add(sentenceEnding.get());
+                result = Optional.of(String.join("", sentenceParts));
                 break;
             }
             else {
                 sentenceParts.add(text);
             }
         }
-        return sentence;
+        return result;
     }
 
     /**
@@ -643,18 +646,17 @@ public class SummaryJavadocCheck extends AbstractJavadocCheck {
     }
 
     /**
-     * Finds the end of a sentence. If a sentence ending period was found, returns the whole string
-     * up to and excluding that period. The end of sentence detection here could be replaced in the
+     * Finds the end of a sentence. The end of sentence detection here could be replaced in the
      * future by Java's built-in BreakIterator class.
      *
      * @param text The string to search.
      * @param period The period character to find.
-     * @return The string up to and excluding the period, or null if no ending was found.
+     * @return An Optional containing the string up to and excluding the period,
+     *     or empty Optional if no ending was found.
      */
-    @Nullable
-    private static String findSentenceEndingOrNull(String text, String period) {
+    private static Optional<String> findSentenceEnding(String text, String period) {
         int periodIndex = text.indexOf(period);
-        String sentenceEnding = null;
+        Optional<String> result = Optional.empty();
         while (periodIndex >= 0) {
             final int afterPeriodIndex = periodIndex + period.length();
 
@@ -663,13 +665,14 @@ public class SummaryJavadocCheck extends AbstractJavadocCheck {
             if (!DEFAULT_PERIOD.equals(period)
                 || afterPeriodIndex >= text.length()
                 || Character.isWhitespace(text.charAt(afterPeriodIndex))) {
-                sentenceEnding = text.substring(0, periodIndex);
+                final String resultStr = text.substring(0, periodIndex);
+                result = Optional.of(resultStr);
                 break;
             }
             else {
                 periodIndex = text.indexOf(period, afterPeriodIndex);
             }
         }
-        return sentenceEnding;
+        return result;
     }
 }


romani:
We need to cover this new code with tests
```
patch -p1 <<'EOF'
diff --git a/config/pitest-suppressions/pitest-javadoc-suppressions.xml b/config/pitest-suppressions/pitest-javadoc-suppressions.xml
index 75572c9..6161e7d 100644
--- a/config/pitest-suppressions/pitest-javadoc-suppressions.xml
+++ b/config/pitest-suppressions/pitest-javadoc-suppressions.xml
@@ -189,6 +189,24 @@
     <lineContent>return astType == TokenTypes.VARIABLE_DEF</lineContent>
   </mutation>
 
+    <sourceFile>SummaryJavadocCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.SummaryJavadocCheck</mutatedClass>
+    <mutatedMethod>findSentenceEnding</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
+    <description>removed call to java/lang/String::isEmpty</description>
+    <lineContent>if (resultStr.isEmpty()) {</lineContent>
+  </mutation>
+
+    <sourceFile>SummaryJavadocCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.SummaryJavadocCheck</mutatedClass>
+    <mutatedMethod>findSentenceEnding</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_ELSE</mutator>
+    <description>removed conditional - replaced equality check with false</description>
+    <lineContent>if (resultStr.isEmpty()) {</lineContent>
+  </mutation>
+
   <mutation unstable="false">
     <sourceFile>TagParser.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.TagParser</mutatedClass>
EOF
```

mohitsatr:
I actually overlooked this part. removed the extra code as it was written for pure Strings based logic.

mohitsatr:
Github, generate report for SummaryJavadoc/all-examples-in-one

romani (approved):
I think this code looks better without null

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -663,13 +665,14 @@ private static String findSentenceEndingOrNull(String text, String period) {
...
+                final String resultStr = text.substring(0, periodIndex);
+                result = Optional.of(resultStr);
                 break;
             }
             else {
                 periodIndex = text.indexOf(period, afterPeriodIndex);
             }
         }
-        return sentenceEnding;
+        return result;
```
nrmancuso, line 676: @mohitsatr what are the tradeoffs between using Optionals throughout the method as you've done here, or returning `Optional.ofNullable(result)` here instead?

nrmancuso (dismissed):
Items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -600,28 +601,30 @@ private static String getStringInsideTag(String result, DetailNode detailNode) {
...
+     * @return An Optional containing the first sentence
+     *     up to and excluding the period, or an empty
+     *     Optional if no ending was found.
      */
-    @Nullable
-    private static String getFirstSentenceOrNull(DetailNode ast, String period) {
+    private static Optional<String> getFirstSentence(DetailNode ast, String period) {
         final List<String> sentenceParts = new ArrayList<>();
-        String sentence = null;
+        Optional<String> result = Optional.empty();
```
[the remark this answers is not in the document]
romani: Let's do refactoring step by step, in few PRs.
It will be easier for both sides.
If something can be done in next PR let's do it in next PR, we can keep issue open as much as we need.

This PR is  already good milestone. 
@mohitsatr, if some suggestions is easy let's do it in this PR, if not that easy, better to postpone to next PR.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
mohitsatr: allow me to explain how this logic is working: 

In `findSentenceEnding` function, `result` can have two possible values: an Empty-optional or an optional containing a String value.

In `getFirstSentence` func, we do :
`sentenceEnding = findSentenceEnding(test, period);`

if `sentenceEnding.isPresent()` is true (the case when we have successfully found the sentence ending) : 
it means Optional contains a String value (empty string "" is also valid). it is when we break the loop.

if `isPresent()` is false: it is empty-optional. ending is not found yet. keep looping.

I was not aware of `.ofNullable` api. After looking into it, i think **not using it** makes more sense in this case because `Optional.ofNullable(result)` would only make sense in null based logic.(i.e when `result` can be either a string or a null as in original code). Also isn't the sole purpose of all this refactoring is to eliminate null from the code ?

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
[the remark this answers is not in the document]
mohitsatr: > why not return Optional.ofNullable(result) here?

similar to https://github.com/checkstyle/checkstyle/pull/16020#discussion_r1887845399

> simply stream the "text parts" and use the Stream#takeWhile API

alright. I'will look into it and try to refactor. 
yes, it'd better do it in separate PR.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -240,12 +238,15 @@ private void validateUntaggedSummary(DetailNode ast) {
...
             if (summaryDoc.contains(period)) {
-                final String firstSentence = getFirstSentenceOrNull(ast, period);
-                if (firstSentence == null) {
-                    log(ast.getLineNumber(), MSG_SUMMARY_FIRST_SENTENCE);
+                final Optional<String> firstSentence = getFirstSentence(ast, period);
+
+                if (firstSentence.isPresent()) {
+                    if (containsForbiddenFragment(firstSentence.get())) {
+                        log(ast.getLineNumber(), MSG_SUMMARY_JAVADOC);
+                    }
```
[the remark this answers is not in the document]
mohitsatr: i did try this approach. 

Screenshot from 2024-12-17 07-45-55

it won't let us do this kind of boolean convertion.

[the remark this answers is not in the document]
nrmancuso: You can workaround this by using the object: `Boolean.TRUE`

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
nrmancuso: The point of this issue is to not **return** `null` from these private methods - it is not desirable (or possible) to avoid null completely in Java :). We have no issue with usage of `null` inside of these methods. My point is why make the method body more complicated when we can keep everything the same and just return an optional.

romani: I see no problem to have Optional in this special case, if such code was written from the beginning I would not ask to make in on nulls.
As we started doing de-nullozation in this Check, for some reason, I was completely ok to do a bit more refactoring to try to avoid null. Code is kind of same before and after refactoring by ease of understanding, just more fashional on Optional.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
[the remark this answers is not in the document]
romani: >why not `return Optional.ofNullable(result) ` here ?

To try avoid null if that doesn't not make code hard to read. Fashioned code. As whole this issue.

nrmancuso (commented):
I think we totally missed the spirit of the issue, but that's not surprising considering that there is no issue description with a clear goal. I personally don't see any real value in this PR, but I won't block it from being merged.

My understanding was that we would refactor the existing code to be clearer, avoid all the unnecessary logic and conditions, and what we did was just wrap all the nulls in Optional. 

From https://www.oracle.com/technical-resources/articles/java/java8-optional.html#:~:text=The%20purpose%20of,an%20optional%20value:

> The purpose of Optional is not to replace every single null reference in your codebase but rather to help design better APIs in which—just by reading the signature of a method—users can tell whether to expect an optional value.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
nrmancuso: Using optional has nothing to do with “fashion”, it is an API that should be used judiciously and deliberately like any other. It is not designed to replace all null usage. The changes in this PR are a clear abuse of the API, and violate Oracle’s own advice about Optionals usage (see link below).

romani:
In this PR we do one more step further, and clearly see that it is not final-ideal code.
But I still think it is good step forward and in next PR we can refactor method content, not a signature.
So we just move nulls or null-wrapped Optionals to method internals. That will be much easier to refactor later today on, if we have energy and good ideas on how to do this. If no ideas now, it is ok, we will wait for a bit longer longer, one day it will happen.

rnveach:
> final-ideal ... mostly acceptable ... no very specific ...

Sounds like this is the real problem with this PR, this issue, and our whole mindset in this project.

There is nothing broken, yet we are trying to fix something so subjective that no one can define it or agree on it. We would rather write essays on this subjective content and argue then actually work on fully fleshed out new functionality, proven bugs, or known bad infrastructure.

The declaration of this issue is so low quality that it can't even write out what it wants when the issue author requires essays and verified content from others.

Even if @rdiachenko agrees to this PR, 2 others are against it. @mohitsatr should just move on to something more worth his time, better quality to the community, and hopefully more concrete.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
rdiachenko: I agree with @nrmancuso , I don't understand why we avoid `Optional.ofNullable`?

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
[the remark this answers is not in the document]
romani: attention to detail please, there is not `null` as special meaning in method internals.
We say, that default value of result is empty result.
in loop it is substituted to non-nullable value !!! attention to `Optional.of`
image

null, does not appear anywhere in method, no good room for `Optional.ofNullable`.

And yes, we can try to refactor method more in next PR to make method to be fashioned chain-methods to make it be similar to any other logic around a world, but it will be fashion update.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
romani: please consider https://github.com/checkstyle/checkstyle/pull/16020/files#r1893954538

rdiachenko: Thanks, I got it. But I have a wider question: why can't we use an empty string to represent an absent sentence ending? I tried to do this in this experimental PR: https://github.com/checkstyle/checkstyle/pull/16040

mohitsatr: hi @rdiachenko, this was the first approach i came up with  here: https://github.com/checkstyle/checkstyle/pull/15965#pullrequestreview-2466226778 as it is simple and avoid null and optionals altogether.

romani: Looks like all others likes to read and empty first sentence is very rare, so we can mix actual empty(with period )and no content like `/***/`.
For now we can accept this, if users complain, we will improve.

nrmancuso: I am 100% on board with returning Optional from this method; it makes sense because the sentence is actually optional. I do not agree with the usage throughout the method body in some of these methods. 

Like I said, I won’t block this PR, but I am not a fan of this new fascination with “fashionable” code. Someone once told me that good, easy to read code is a result of hard writing; Optional should not be used as a crutch.

To be honest, we are bikeshedding here and wasting maintainer and contributor time. I would like to see @mohitsatr devote their time to something more interesting, valuable, and rewarding.

rdiachenko (approved):
@mohitsatr I didn't know whether the empty string would work and be the best approach here. After doing this myself in [the PR](https://github.com/checkstyle/checkstyle/pull/16040) and giving it some thought, I do think using Optional is more clear for readibility because it hides the edge case when the sentence is empty. In the PR I raised, I had to add some additional logic to include the period in the result and strip it before validation. That made the logic look a bit simpler but harder to understand. Therefore, I'm good with changes in this PR.

romani:
Looks like we are in agreement within reviewers.

I am merging. Nothing stops is to improve in next PR if there is good idea.

PR #16026, opened 2024-12-12
Issue #13038: handled inner class method definition in VariableDeclarationUsageDistanceCheck

### LINKED ISSUE #13038: VariableDeclarationUsageDistanceCheck doesn't handle method definition properly

MohanadKh03:
Resolves #13038
Extends #14635 

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
@@ -417,6 +417,9 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
                             getFirstNodeInsideTryCatchFinallyBlocks(blockWithVariableUsage,
                                 variable);
                         break;
+                    case TokenTypes.METHOD_DEF:
+                        exprWithVariableUsage = blockWithVariableUsage.getNextSibling();
+                        break;
                     default:
                         exprWithVariableUsage = blockWithVariableUsage.getFirstChild();
                 }
@@ -785,8 +788,7 @@ public class VariableDeclarationUsageDistanceCheck extends AbstractCheck {
                 || type == TokenTypes.TYPE
                 || type == TokenTypes.MODIFIERS
                 || type == TokenTypes.RESOURCE
-                || type == TokenTypes.EXTENDS_CLAUSE
-                || type == TokenTypes.IMPLEMENTS_CLAUSE;
+                || type == TokenTypes.CLASS_DEF;
     }
 
 }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheckTest.java
@@ -352,9 +352,7 @@ public class VariableDeclarationUsageDistanceCheckTest extends
 
     @Test
     public void testGeneralClass4() throws Exception {
-        final String[] expected = {
-            "26:9: " + getCheckMessage(MSG_KEY, "z", 3, 1),
-        };
+        final String[] expected = {};
 
         verifyWithInlineConfigParser(
                 getPath("InputVariableDeclarationUsageDistanceGeneral4.java"), expected);
@@ -410,12 +408,11 @@ public class VariableDeclarationUsageDistanceCheckTest extends
     }
 
     @Test
-    public void testVariableDeclarationUsageDistancePatternVariables() throws Exception {
+    public void testVariableDeclarationUsageDistanceMethodDefinition() throws Exception {
         final String[] expected = {
-            "35:9: " + getCheckMessage(MSG_KEY, "b", 5, 1),
+            "18:9: " + getCheckMessage(MSG_KEY, "c", 2, 1),
         };
         verifyWithInlineConfigParser(
-                getNonCompilablePath("InputVariableDeclarationUsageDistancePatternVariables.java"),
-                expected);
+                getPath("InputVariableDeclarationUsageDistanceMethodDefinition.java"), expected);
     }
 }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral4.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral4.java
@@ -22,8 +22,7 @@ public class InputVariableDeclarationUsageDistanceGeneral4 {
         <T> void xx(List<T> m){}
     }
     public void method9() {
-        // until https://github.com/checkstyle/checkstyle/issues/13011
-        Integer z = 5; // violation 'Distance .* is 3.'
+        Integer z = 5;
         Iterator<Integer> mn = new HashSet<Integer>().iterator();
         class BClass extends Parent {
             Boolean h = false;
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceMethodDefinition.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceMethodDefinition.java
@@ -1,0 +1,30 @@
+/*
+VariableDeclarationUsageDistance
+allowedDistance = 1
+ignoreVariablePattern = (default)
+validateBetweenScopes = true
+ignoreFinal = false
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.variabledeclarationusagedistance;
+
+import java.util.Iterator;
+public class InputVariableDeclarationUsageDistanceMethodDefinition {
+    public void method(){
+        int a=1;
+        int b=2;
+        int c=3;// violation 'Distance .* is 2.'
+        class GeneralLogic {
+            public <T> Iterator<T> method(T[] b) throws Exception{
+                if(a>0){
+                    System.out.println(b.length);
+                }
+                return null;
+            }
+        };
+        System.out.print(b);
+        System.out.print(c);
+    }
+}


MohanadKh03:
GitHub, generate report

romani:
Github, generate report for VariableDeclarationUsageDistance/all-examples-in-one

romani:
Github, generate report for VariableDeclarationUsageDistance/all-examples-in-one

romani:
```
<mutation unstable="false">
+    <sourceFile>VariableDeclarationUsageDistanceCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.VariableDeclarationUsageDistanceCheck</mutatedClass>
+    <mutatedMethod>isZeroDistanceToken</mutatedMethod>
+    <lineContent>return type == TokenTypes.VARIABLE_DEF</lineContent>
+  </mutation>
```

Rule of thumb for complicated cases of pitest (or when there no energy to deal with it) it to follow pitest mutation and simply remove/change code. Eventually user will find and report as defect what is not covered.

We will run regression report and we will a difference with such line and without, if there be any diff between reports, it means we found a case that covers this mutation.

-----

Please remove this line.

romani:
@MohanadKh03, do not hesitate to let me know that my message is not clear, I can assist further.

MohanadKh03:
Oh my bad it took me long I was trying to figure out how to change the current logic without changing ```VARIABLE_DEF``` because if we do remove it there will be lots of errors
Though it seems we might need to come up with a new idea entirely instead of ```isZeroDistanceToken```

> so we definitely changing some previous behavior, I think we will be lucky in diff of two reports.

Yes do you suggest changing this method or perhaps add more tests ? I think a better approach would be adding more tests to the existing logic since it covers all of the tokens which will have zero impact on calculating the distance , But please let me know if there could be a better option @romani

romani:
>I think a better approach would be adding more tests to the existing logic since it covers all of the tokens which will have zero impact on calculating the distance

Yes, but we need to find such test cases. If you see how to do this relatively easy , let's go this way.

I tried to hint that, if pitest detected dead code, and we do see how to cover it by test. We can apply to code whatever pitest suggest and run regression diff report, it executes against huge amount of sources, most likely it will catch such a java code where such removal of code produces different results of validation, so we just need to copy such code to our Inputs.

MohanadKh03:
Turns out the problem was not at 
```java
 return type == TokenTypes.VARIABLE_DEF
```
like what was reported using pitest somehow maybe it could have been detected at bytecode instead ? Idk  
Anyways it is just as mentioned here before here https://github.com/checkstyle/checkstyle/pull/14635#discussion_r1548879267 

Using ``implements`` and ``extends`` with ``class`` does not really make sense since they will always be children tokens of ``class``   

After the CI goes green **hopefully** if all goes well then this is good to go imo 
Though the only thing left would be to make this check compatible with keywords such as record and anything 17+ basically and add tests for it
Should we add records and/or any 17+ keywords to it now ? WDYT ? @romani

romani:
Github, generate report for VariableDeclarationUsageDistance/all-examples-in-one

romani:
Such method with "list" of tokens appeared athttps://github.com/checkstyle/checkstyle/commit/63747be122cf4e89b111f11d268b5853f9d88b52

Let's  add  to Inputs some records. But if  support of them becomes a bit more tricky, I would recommend to do this in separate issue. If this PR addresses issue and don't introduce new false positives or negatives, we can merge this PR. And improve on records in separate issue.

I presume this Check didn't not work with records even before this PR.

MohanadKh03:
Looks like there are a few changes in the regression though they seem kind of logical I think ? 
Not sure if my judgement is correct but it is probably not good to let changes occur is it ?

romani:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/708ab8dd3aeed7d9205eafbc7cd3938d8ae293ef_2024023508/reports/diff/openjdk21/index.html#A1

this is false negative please add it to Inputs. 
It can be moved close to class.
Am I missing something ? or it effect of validateBetweenScopes=true .

romani:
>false negative

It means there should be violation but it not happening.

In report it is in red color, it means violation is gone, not happening anymore.
Can we return it back to happening?

romani:
@MohanadKh03 ,
let me know if you need any assistance

MohanadKh03:
> Can we return it back to happening?
 
 After some debugging this is probably the reason when the CLASS_DEF in here we should calculate the distance in the scope of METHOD_DEF that has the variable (method is add and variable is c2Accumulator in our case)  
 ```java
 class PairBox {
        A1 left = c1Supplier.get();
        A2 right = c2Supplier.get();

        void add(T t) {
            c1Accumulator.accept(left, t);
            c2Accumulator.accept(right, t);
        }

        PairBox combine(PairBox other) {
            left = c1Combiner.apply(left, other.left);
            right = c2Combiner.apply(right, other.right);
            return this;
        }

        R get() {
            R1 r1 = c1Finisher.apply(left);
            R2 r2 = c2Finisher.apply(right);
            return merger.apply(r1, r2);
        }
    }
 ```

But when we reach the METHOD_DEF for some reason it traverses to the next sibling ? I dont see this calculating the distance inside the METHOD_DEF scope

```java
case TokenTypes.METHOD_DEF:
      exprWithVariableUsage = blockWithVariableUsage.getNextSibling();
      break;
```
in Line 420 VariableDeclarationUsageDistanceCheck
This is the reason there are regressions in computing distances between scopes most likely
I doubt it is right to use the next sibling as an indicator to the distance of a token inside a method in a different scope because we should first go to the METHOD_DEF scope and then inside the method calculate what  the distance is inside this method

I think that by using the ```calculateDistanceBetweenScopes``` can we not just find exactly what scope is the first usage , calculate if there is any distance relevant to this scope and then call ```calculateDistanceInSingleScope``` inside the scope which it is used in ? What do you think ? @romani

MohanadKh03:
@romani Pinging as a reminder

romani:
Please add to Inputs 
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/708ab8dd3aeed7d9205eafbc7cd3938d8ae293ef_2024023508/reports/diff/openjdk21/index.html#A5

Code is not traditional, but still distance violation was good

romani:
https://checkstyle.org/checks/coding/variabledeclarationusagedistance.html#Example4-code

`<property name="validateBetweenScopes" value="true"/>` was create  for  non-named `{ }` blocks of code that can happen at any time, and people like it in some cases where they wanted to keep visibility of inner variables close.

when ever we run into inner Class in method,  it should not count as `validateBetweenScopes=true`, as there is not way to change structure of class.

MohanadKh03:
> Please add to Inputs
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/708ab8dd3aeed7d9205eafbc7cd3938d8ae293ef_2024023508/reports/diff/openjdk21/index.html#A5

Sorry for being late but I actually tried most of the regressions and true they had the same problem as I mentioned earlier here
> After some debugging this is probably the reason when the CLASS_DEF in here we should calculate the distance in the scope of METHOD_DEF that has the variable (method is add and variable is c2Accumulator in our case)


> I think solution should be simple, count whole inner type (nomatter how long it is) that are declared in methods as 1.

Alright but this would mean that the old regressions that counted inner classes as a different scope would be a valid regression ? I see this would break a lot of the regressions

romani:
yes, code like this:
```
284         Symbol sym = cast(Symbol.class, e);
285         class Vis extends JCTree.Visitor {
286             List<JCAnnotation> result = null;
287             public void visitModuleDef(JCModuleDecl tree) {
288                 result = tree.mods.annotations;
289             }
290             public void visitPackageDef(JCPackageDecl tree) {
291                 result = tree.annotations;
292             }
293             public void visitClassDef(JCClassDecl tree) {
294                 result = tree.mods.annotations;
295             }
296             public void visitMethodDef(JCMethodDecl tree) {
297                 result = tree.mods.annotations;
298             }
299             public void visitVarDef(JCVariableDecl tree) {
300                 result = tree.mods.annotations;
301             }
302             @Override
303             public void visitTypeParameter(JCTypeParameter tree) {
304                 result = tree.annotations;
305             }
306         }
307         Vis vis = new Vis();
308         tree.accept(vis);
309         if (vis.result == null)
310             return null;
311 
312         List<Attribute.Compound> annos = sym.getAnnotationMirrors();
```

should be affected in regression (no violation  should happen), BUT we jsut need to be sure that same structure code but with bigger distance still an violation.
`Distance between variable 'sym' declaration and its first usage is 4, but allowed 3. Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value).` are we sure that in this case distance is not 4 ? sound like 4, so violation should stay.

MohanadKh03:
Yes I will be on it and finish it tomorrow apologies for the long PR

romani:
@MohanadKh03 , is there any chance to finish this PR ?

MohanadKh03:
@romani Hi sorry I didnt have much time to complete it and currently a little bit too busy
Hopefully someone else can pick this up and finish it though

romani:
@smita1078 , do you have time to help us with finishing this issue? At glance this PR is almost ready and just some test extension is required.

smita1078:
> @smita1078 , do you have time to help us with finishing this issue? At glance this PR is almost ready and just some test extension is required.

looking into it, just to confirm need to extend test's in VariableDeclarationUsageDistanceCheck??

romani:
We need to fix defect.
This PR is not complete, by comments looks like only test extension is left to be done, but it might be that main code change will be required too.
As PR is abandoned,  you can take ownership of such code, make new PR, and we will deal with it from scratch

romani:
activity is moved to https://github.com/checkstyle/checkstyle/pull/17526

PR #16031, opened 2024-12-13
Issue #13999: kill mutation for JavadocMethodCheck

mohitsatr:
part of #13999:

mutation in question:
https://github.com/checkstyle/checkstyle/blob/24eedce6a64145442db5b478c7a39ce0763dad8c/config/pitest-suppressions/pitest-javadoc-suppressions.xml#L57-L64

**Rationale:**
`surroundingAccessModifier ` is one of the deciding factor if  we want to process the method/construction or not. the case `surroundingAccessModifier == null` is when we come on a overriden method. For example : 
```
final Runnable r = new Runnable() {
    public void run() {};
};
```
this condition might have been placed there because: 
 - it's better not to send surroundingAccessModifier down in camparsions because with null there is always risk of NullPointerException error. 
 - this simple checking saves us making unnecessary camparisons.

**Removal is OK**  because surroundingAccessModifier is something we are only camparing with each modifier object. we are not calling a method call on surroundingAccessModifier.
in short `==` there is no risk of running into NullPointerException error here.

### DIFF
--- a/config/pitest-suppressions/pitest-javadoc-suppressions.xml
+++ b/config/pitest-suppressions/pitest-javadoc-suppressions.xml
@@ -56,15 +56,6 @@
 
   <mutation unstable="false">
     <sourceFile>JavadocMethodCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocMethodCheck</mutatedClass>
-    <mutatedMethod>shouldCheck</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
-    <description>removed conditional - replaced equality check with true</description>
-    <lineContent>return surroundingAccessModifier != null</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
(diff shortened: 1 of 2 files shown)


romani:

romani:
If diff report shows no difference, we can merge

romani:
@mohitsatr , can you explain analytically why removal is ok?

in practice(diff report) we see that it can be removed.
Ths is not ast based Check so we are accepting some theoretically regression on code that is not part of our regression.

mohitsatr:
> @mohitsatr , can you explain analytically why removal is ok?

done 
the PR was in draft because I was still figuring out how to generate diff report.

romani:
https://github.com/checkstyle/checkstyle/blob/master/.github/workflows/README.md#regression-report-workflow---regression-reportyml

Or

https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#executing-generation-using-github-action

------

I already did at https://github.com/checkstyle/checkstyle/pull/16031#issuecomment-2541771918

mohitsatr:
CI is read . is it me ? should I push again ?

romani (approved):
Ok to merge

nrmancuso (approved):
Thanks a lot @mohitsatr

Issue #16043, opened 2024-12-23
Update commons-beanutils:commons-beanutils @ 1.9.4


HiLucho:
**Is your feature request related to a problem? Please describe.**

Is a security issue related to Checkmarx.

**Describe the solution you'd like**
We are getting vulnerability errors, using the latest version of Checkstyle plugin.

The solution is update to ([SOURCE](https://github.com/apache/commons-beanutils)):

```bash
<dependency>
  <groupId>org.apache.commons</groupId>
  <artifactId>commons-beanutils2</artifactId>
  <version>2.0.0</version>
</dependency>
```

**Additional context**
![image](https://github.com/user-attachments/assets/de3c010b-2158-4d31-a446-d557c720ff88)

romani:
>[ERROR] Failed to execute goal on project checkstyle: Could not resolve dependencies for project com.puppycrawl.tools:checkstyle:jar:10.21.1-SNAPSHOT: Could not find artifact commons-beanutils:commons-beanutils:jar:2.0.0 in central (https://repo.maven.apache.org/maven2) -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project checkstyle: Could not resolve dependencies for project com.puppycrawl.tools:checkstyle:jar:10.21.1-SNAPSHOT: Could not find artifact commons-beanutils:commons-beanutils:jar:2.0.0 in central (https://repo.maven.apache.org/maven2)
 

@HiLucho , where do you see release of this library ?

HiLucho:
Oh my gosh, well, I saw it directly in [https://github.com/apache/commons-beanutils](https://github.com/apache/commons-beanutils). 

But, you are right, that version isn't in maven artifactory :(

HiLucho:
Since the vulnerable library is **commons:collections**, perhaps a temporary solution to resolve it could be **commons:collections4**, and the source is this: https://mvnrepository.com/artifact/org.apache.commons/commons-collections4. I mean, forcing beanutils to use commons-collections4.

romani:
@HiLucho, please do not hesitate to send PR .

romani:
In referenced PR , we see  that we Blocked by https://github.com/checkstyle/checkstyle/issues/16043

romani:
we moved to 1.10.0 at  https://github.com/checkstyle/checkstyle/pull/16130

PR #16044, opened 2024-12-23
Issue #16043: bump commons-beanutils to 2.0.0

romani:
Issue #16043

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -286,7 +286,7 @@
     <dependency>
       <groupId>commons-beanutils</groupId>
       <artifactId>commons-beanutils</artifactId>
-      <version>1.9.4</version>
+      <version>2.0.0</version>
     </dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -286,7 +286,7 @@
     <dependency>
       <groupId>commons-beanutils</groupId>
```
HiLucho, line 287: `<groupId>org.apache.commons</groupId>`

pom.xml
```diff
@@ -286,7 +286,7 @@
...
       <artifactId>commons-beanutils</artifactId>
```
HiLucho, line 288: <artifactId>commons-beanutils2</artifactId>

pom.xml
romani: No such group+ artifact in maven central. It is future.

Issue #3183, opened 2016-05-17
JavadocVariable doesn't report missing comment when excludeScope=anoninner


edrandall:
JavadocVariable - https://checkstyle.org/config_javadoc.html#JavadocVariable
 
does not report missing comments on member variables in a standard class,  irrespective of scope, when excludeScope="anoninner". 

config.xml:

```
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN" 
    "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
<module name="Checker">
    <property name="charset" value="ASCII" />
    <property name="severity" value="error" />
    <property name="fileExtensions" value="java" />

    <module name="TreeWalker">
        <module name="JavadocVariable">
            <property name="excludeScope" value="anoninner" />
        </module>
    </module>
</module>
```

TestClass.java:

```
package test;

/** Class for testing checkstyle. */
public class TestClass {

    private String key;

    public TestClass() {
    }
}
```

`$ java -jar checkstyle-6.18-all.jar -c config.xml TestClass.java`
Actual output:

```
Starting audit...
Audit done.
```

Expected output:

```
Starting audit...
[ERROR] C:\checkstyle\TestClass.java:6:9: Missing a Javadoc comment [JavadocVariable]
Audit done.
Checkstyle ends with 1 errors.
```
##

rnveach:
Most likely this is because of issues we have had with `Scope` enumeration and should switch to `AccessModifier` here like we did for `ParameterNameCheck`.

leojahn1:
I don't understand this problem. Based on this definition of Scope and excludeScope (https://checkstyle.sourceforge.io/property_types.html#Scope),
Scope is functioning properly. What are we trying to resolve here? Also, what reason would there be to set excludeScope=anoninner in the first place?

cowwoc:
Per https://github.com/checkstyle/checkstyle/issues/9280 this issue should have been closed.

romani:
Yes, no `excludeScope` after 10.22.0

PR #9277, opened 2021-02-14
Issue #9280: replace `Scope` with `AccessModifierOption` in JavadocVariableCheck

martin-mfg:
fix for #9280

might be related issue #3183

In `JavadocVariableCheck`, replace `Scope scope` and `Scope excludeScope` by `AccessModifierOption[] accessModifiers`.

Whether the JavadocVariableCheck is applied to a given variable still depends on the variable's modifier *and on the variable's surrounding scope*, like it was before. I.e. to run the check on a `public` field in a `private` class, `accessModifiers` must include `private`, but not necessarily `public`. If it was intended to change this behaviour, let me know.

I intended to make changes similar to those in [f91b1af](https://github.com/checkstyle/checkstyle/commit/f91b1af3d7c130daaeb52fea303b5d8e4c2769bf#diff-f1dcf2d250d8b52ccfbc408a657c47c4c735aef522b9c2f93c7826d93c152541), because of this comment: https://github.com/checkstyle/checkstyle/issues/3183#issuecomment-283539970.

In https://github.com/checkstyle/checkstyle/issues/3675#issue-197441645 it says 
> Looks like scope is counted by more visible and less visible. But should be simply by matching.

I thought this should apply here too.

-------

Similar update was done at https://github.com/checkstyle/checkstyle/pull/9670 and it has interesting problem point in javadoc on inner classes.

-------

ATTENTION: we recently run into edge case with scope property but for another Check - https://github.com/checkstyle/checkstyle/issues/13749 , and I am not sure how accessModifiersOption is going to handle it.

PR #16049, opened 2024-12-24
Issue #9280: New property 'accessModifiers' as substitution of 'scope' and 'excludeScope' in JavadocVariableCheck

### LINKED ISSUE #9280: JavadocVariable: new property 'accessModifiers' as substitution of 'scope' and 'excludeScope'
JavadocVariable - https://checkstyle.org/config_javadoc.html#JavadocVariable
https://checkstyle.org/checks/javadoc/javadocvariable.html#JavadocVariable
 
Example of user confusion and problem with such type - https://github.com/checkstyle/checkstyle/issues/3183 

have have issues with `Scope` enumeration 
we should switch to `AccessModifier` here like we did for `ParameterNameCheck`.
New property need t be created as substitution to 2 old properties.
https://checkstyle.org/property_types.html#AccessModifierOption.5B.5D
https://checkstyle.org/config_naming.html#ParameterName_Properties

TODO: after completion close https://github.com/checkstyle/checkstyle/issues/3183

local classes (named classes in methods or any other blocks) and antonymous classes should be skipped by Check for validation, see details at  https://github.com/checkstyle/checkstyle/issues/13749

--------

**Migration notes:**
```diff
    <module name="JavadocVariable">
       <property name="severity" value="warning"/>
-      <property name="scope" value="protected"/>
+      <property name="accessModifiers" value="protected"/>
     </module>
```

There might be few more violations, user needs to fix the or suppress.
Examples of migration:
- [build-tools PR](https://github.com/pmd/build-tools/pull/55)
- [xwiki-commons PR](https://github.com/xwiki/xwiki-commons/pull/1241)
- [xwiki-platforms PR](https://github.com/xwiki/xwiki-platform/pull/3907)

kkoutsilis:
> [!CAUTION]
> Breaking change

Aims to close #9280 
This PR removes `scope` and `excludeScope` from [JavadocVariableCheck](https://checkstyle.org/checks/javadoc/javadocvariable.html#JavadocVariable) while substitutes with new `accessModifers` property.

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -216,12 +216,17 @@ no-error-pmd)
   echo "CS_version: ${CS_POM_VERSION}"
   mvn -e --no-transfer-progress clean install -Pno-validations
   echo "Checkout target sources ..."
-  checkout_from "https://github.com/pmd/build-tools.git"
+  #  checkout_from "https://github.com/pmd/build-tools.git"
+  checkout_from "https://github.com/kkoutsilis/build-tools.git"
   cd .ci-temp/build-tools/
+  git ls-remote
+  git checkout "66d""ed33c74662cb3da612f3d34a5ae""fa""a629b443"
   mvn -e --no-transfer-progress install
   cd ..
-  git clone https://github.com/pmd/pmd.git
+  git clone https://github.com/kkoutsilis/pmd.git
   cd pmd
+  git ls-remote
+  git checkout "fa6a862ac8278906d7bcf21852f6552d27a46a73"
   ./mvnw verify --show-version --errors --batch-mode --no-transfer-progress \
                 -DskipTests \
                 -Dmaven.javadoc.skip=true \
@@ -279,8 +284,10 @@ no-error-xwiki)
   echo "version:${CS_POM_VERSION} antlr4:${ANTLR4_VERSION}"
   mvn -e --no-transfer-progress clean install -Pno-validations
   echo "Checkout target sources ..."
-  checkout_from "https://github.com/xwiki/xwiki-commons.git"
+  checkout_from "https://github.com/kkoutsilis/xwiki-commons"
   cd .ci-temp/xwiki-commons
+  git ls-remote
+  git checkout "88f75d13376587956a5e5bd""dad0fa003383f190a"
   # Build custom Checkstyle rules
   mvn -e --no-transfer-progress -f \
     xwiki-commons-tools/xwiki-commons-tool-verification-resources/pom.xml \
@@ -311,8 +318,10 @@ no-error-xwiki)
   cd ..
   removeFolderWithProtectedFiles xwiki-rendering
   cd ..
-  checkout_from https://github.com/xwiki/xwiki-platform.git
+  checkout_from https://github.com/kkoutsilis/xwiki-platform.git
   cd .ci-temp/xwiki-platform
+  git ls-remote
+  git checkout "01848f""ca559805b535559b7b94119a95990a2b5c"
   # Validate xwiki-platform
   mvn -e --no-transfer-progress checkstyle:check@default -Dcheckstyle.version="${CS_POM_VERSION}"
   cd ..
@@ -748,22 +757,6 @@ no-error-methods-distance)
      -Dcheckstyle.configLocation=../../config/checkstyle-checks.xml
   cd ..
   removeFolderWithProtectedFiles  methods-distance
-  ;;
-
-no-error-spring-cloud-gcp)
-  set -e
-  CS_POM_VERSION="$(getCheckstylePomVersion)"
-  echo CS_version: "${CS_POM_VERSION}"
-  mvn -e --no-transfer-progress clean install -Pno-validations
-  echo "Checkout target sources ..."
-  checkout_from https://github.com/googlecloudplatform/spring-cloud-gcp
-  cd .ci-temp/spring-cloud-gcp
-  git checkout "7c99f37087ac8f""eb""db""f7e185375a217b744895a4"
-  mvn -e --no-transfer-progress checkstyle:check@checkstyle-validation \
-   -Dmaven-checkstyle-plugin.version=3.1.1 \
-   -Dpuppycrawl-tools-checkstyle.version="${CS_POM_VERSION}"
-  cd ..
-  removeFolderWithProtectedFiles spring-cloud-gcp
   ;;
 
 no-error-equalsverifier)
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
@@ -19,16 +19,19 @@
 
 package com.puppycrawl.tools.checkstyle.checks.javadoc;
 
+import java.util.Arrays;
 import java.util.regex.Pattern;
 
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.FileContents;
-import com.puppycrawl.tools.checkstyle.api.Scope;
 import com.puppycrawl.tools.checkstyle.api.TextBlock;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.checks.naming.AccessModifierOption;
+import com.puppycrawl.tools.checkstyle.utils.CheckUtil;
 import com.puppycrawl.tools.checkstyle.utils.ScopeUtil;
+import com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil;
 
 /**
  * <div>
@@ -36,20 +39,19 @@ import com.puppycrawl.tools.checkstyle.utils.ScopeUtil;
  * </div>
  * <ul>
  * <li>
- * Property {@code excludeScope} - Specify the visibility scope where Javadoc
- * comments are not checked.
- * Type is {@code com.puppycrawl.tools.checkstyle.api.Scope}.
- * Default value is {@code null}.
+ * Property {@code accessModifiers} - Specify the set of access modifiers used to determine which
+ * fields should be checked. This includes both explicitly declared modifiers and implicit ones,
+ * such as package-private for fields without an explicit modifier.
+ * It also accounts for special cases where fields have implicit modifiers,
+ * such as {@code public static final} for interface fields and {@code public static}
+ * for enum constants. Only fields matching the specified modifiers will be analyzed.
+ * Type is {@code com.puppycrawl.tools.checkstyle.checks.naming.AccessModifierOption[]}.
+ * Default value is {@code public, protected, package, private}.
  * </li>
  * <li>
  * Property {@code ignoreNamePattern} - Specify the regexp to define variable names to ignore.
  * Type is {@code java.util.regex.Pattern}.
  * Default value is {@code null}.
- * </li>
- * <li>
- * Property {@code scope} - Specify the visibility scope where Javadoc comments are checked.
- * Type is {@code com.puppycrawl.tools.checkstyle.api.Scope}.
- * Default value is {@code private}.
  * </li>
  * <li>
  * Property {@code tokens} - tokens to check
@@ -84,35 +86,40 @@ public class JavadocVariableCheck
      * A key is pointing to the warning message text in "messages.properties"
      * file.
      */
+
     public static final String MSG_JAVADOC_MISSING = "javadoc.missing";
-
-    /** Specify the visibility scope where Javadoc comments are checked. */
-    private Scope scope = Scope.PRIVATE;
-
-    /** Specify the visibility scope where Javadoc comments are not checked. */
-    private Scope excludeScope;
+    /**
+     * Specify the set of access modifiers used to determine which fields should be checked.
+     *  This includes both explicitly declared modifiers and implicit ones, such as package-private
+     *  for fields without an explicit modifier. It also accounts for special cases where fields
+     *  have implicit modifiers, such as {@code public static final} for interface fields and
+     *  {@code public static} for enum constants.
+     *  Only fields matching the specified modifiers will be analyzed.
+     */
+    private AccessModifierOption[] accessModifiers = {
+        AccessModifierOption.PUBLIC,
+        AccessModifierOption.PROTECTED,
+        AccessModifierOption.PACKAGE,
+        AccessModifierOption.PRIVATE,
+    };
 
     /** Specify the regexp to define variable names to ignore. */
     private Pattern ignoreNamePattern;
 
     /**
-     * Setter to specify the visibility scope where Javadoc comments are checked.
+     * Setter to specify the set of access modifiers used to determine which fields should be
+     * checked. This includes both explicitly declared modifiers and implicit ones, such as
+     * package-private for fields without an explicit modifier. It also accounts for special
+     * cases where fields have implicit modifiers, such as {@code public static final}
+     * for interface fields and {@code public static} for enum constants.
+     * Only fields matching the specified modifiers will be analyzed.
      *
-     * @param scope a scope.
-     * @since 3.0
+     * @param accessModifiers access modifiers of fields to check.
+     * @since 10.22.0
      */
-    public void setScope(Scope scope) {
-        this.scope = scope;
-    }
-
-    /**
-     * Setter to specify the visibility scope where Javadoc comments are not checked.
-     *
-     * @param excludeScope a scope.
-     * @since 3.4
-     */
-    public void setExcludeScope(Scope excludeScope) {
-        this.excludeScope = excludeScope;
+    public void setAccessModifiers(AccessModifierOption... accessModifiers) {
+        this.accessModifiers =
+            UnmodifiableCollectionUtil.copyOfArray(accessModifiers, accessModifiers.length);
     }
 
     /**
@@ -177,6 +184,17 @@ public class JavadocVariableCheck
     }
 
     /**
+     * Checks whether a method has the correct access modifier to be checked.
+     *
+     * @param accessModifier the access modifier of the method.
+     * @return whether the method matches the expected access modifier.
+     */
+    private boolean matchAccessModifiers(AccessModifierOption accessModifier) {
+        return Arrays.stream(accessModifiers)
+            .anyMatch(modifier -> modifier == accessModifier);
+    }
+
+    /**
      * Whether we should check this node.
      *
      * @param ast a given node.
@@ -185,14 +203,10 @@ public class JavadocVariableCheck
     private boolean shouldCheck(final DetailAST ast) {
         boolean result = false;
         if (!ScopeUtil.isInCodeBlock(ast) && !isIgnored(ast)) {
-            final Scope customScope = ScopeUtil.getScope(ast);
-            final Scope surroundingScope = ScopeUtil.getSurroundingScope(ast);
-            result = customScope.isIn(scope) && surroundingScope.isIn(scope)
-                && (excludeScope == null
-                    || !customScope.isIn(excludeScope)
-                    || !surroundingScope.isIn(excludeScope));
+            final AccessModifierOption accessModifier =
+                    CheckUtil.getAccessModifierFromModifiersToken(ast);
+            result = matchAccessModifiers(accessModifier);
         }
         return result;
     }
-
 }
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
@@ -314,9 +314,14 @@ public final class CheckUtil {
      * @return the access modifier of the method/constructor.
      */
     public static AccessModifierOption getAccessModifierFromModifiersToken(DetailAST ast) {
-        final DetailAST modsToken = ast.findFirstToken(TokenTypes.MODIFIERS);
-        AccessModifierOption accessModifier =
-                getAccessModifierFromModifiersTokenDirectly(modsToken);
+        AccessModifierOption accessModifier;
+        if (ast.getType() == TokenTypes.ENUM_CONSTANT_DEF) {
+            accessModifier = AccessModifierOption.PUBLIC;
+        }
+        else {
+            final DetailAST modsToken = ast.findFirstToken(TokenTypes.MODIFIERS);
+            accessModifier = getAccessModifierFromModifiersTokenDirectly(modsToken);
+        }
 
         if (accessModifier == AccessModifierOption.PACKAGE) {
             if (ScopeUtil.isInEnumBlock(ast) && ast.getType() == TokenTypes.CTOR_DEF) {
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/javadoc/JavadocVariableCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/javadoc/JavadocVariableCheck.xml
@@ -8,17 +8,18 @@
  Checks that a variable has a Javadoc comment. Ignores {@code serialVersionUID} fields.
  &lt;/div&gt;</description>
          <properties>
-            <property name="excludeScope" type="com.puppycrawl.tools.checkstyle.api.Scope">
-               <description>Specify the visibility scope where Javadoc
- comments are not checked.</description>
+            <property default-value="public, protected, package, private"
+                      name="accessModifiers"
+                      type="com.puppycrawl.tools.checkstyle.checks.naming.AccessModifierOption[]">
+               <description>Specify the set of access modifiers used to determine which
+ fields should be checked. This includes both explicitly declared modifiers and implicit ones,
+ such as package-private for fields without an explicit modifier.
+ It also accounts for special cases where fields have implicit modifiers,
+ such as {@code public static final} for interface fields and {@code public static}
+ for enum constants. Only fields matching the specified modifiers will be analyzed.</description>
             </property>
             <property name="ignoreNamePattern" type="java.util.regex.Pattern">
                <description>Specify the regexp to define variable names to ignore.</description>
-            </property>
-            <property default-value="private"
-                      name="scope"
-                      type="com.puppycrawl.tools.checkstyle.api.Scope">
-               <description>Specify the visibility scope where Javadoc comments are checked.</description>
             </property>
             <property default-value="ENUM_CONSTANT_DEF"
                       name="tokens"
--- a/src/site/xdoc/checks/javadoc/javadocvariable.xml
+++ b/src/site/xdoc/checks/javadoc/javadocvariable.xml
@@ -26,11 +26,11 @@
               <th>since</th>
             </tr>
             <tr>
-              <td>excludeScope</td>
-              <td>Specify the visibility scope where Javadoc comments are not checked.</td>
-              <td><a href="../../property_types.html#Scope">Scope</a></td>
-              <td><code>null</code></td>
-              <td>3.4</td>
+              <td>accessModifiers</td>
+              <td>Specify the set of access modifiers used to determine which fields should be checked. This includes both explicitly declared modifiers and implicit ones, such as package-private for fields without an explicit modifier. It also accounts for special cases where fields have implicit modifiers, such as <code>public static final</code> for interface fields and <code>public static</code> for enum constants. Only fields matching the specified modifiers will be analyzed.</td>
+              <td><a href="../../property_types.html#AccessModifierOption.5B.5D">AccessModifierOption[]</a></td>
+              <td><code>public, protected, package, private</code></td>
+              <td>10.22.0</td>
             </tr>
             <tr>
               <td>ignoreNamePattern</td>
@@ -38,13 +38,6 @@
               <td><a href="../../property_types.html#Pattern">Pattern</a></td>
               <td><code>null</code></td>
               <td>5.8</td>
-            </tr>
-            <tr>
-              <td>scope</td>
-              <td>Specify the visibility scope where Javadoc comments are checked.</td>
-              <td><a href="../../property_types.html#Scope">Scope</a></td>
-              <td><code>private</code></td>
-              <td>3.0</td>
             </tr>
             <tr>
               <td>tokens</td>
@@ -78,40 +71,40 @@
 </code></pre></div>
         <p id="Example1-code">
           By default, this setting will report a violation if
-          there is no javadoc for any scope member.
+          there is no javadoc for a field with any access modifier.
         </p>
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
 public class Example1 {
-  private int a; // violation
+  private int a; // violation, 'Missing a Javadoc comment'
 
   /**
    * Some description here
    */
   private int b;
-  protected int c; // violation
-  public int d; // violation
-  /*package*/ int e; // violation
+  protected int c; // violation, 'Missing a Javadoc comment'
+  public int d; // violation, 'Missing a Javadoc comment'
+  /*package*/ int e; // violation, 'Missing a Javadoc comment'
 
 }
 </code></pre></div>
 
         <p id="Example2-config">
-          To configure the check for <code>public</code>
-          scope:
+          To configure the check <code>public</code>
+          access modifier:
         </p>
 
         <div class="wrapper"><pre class="prettyprint"><code class="language-xml">
 &lt;module name="Checker"&gt;
   &lt;module name="TreeWalker"&gt;
     &lt;module name="JavadocVariable"&gt;
-      &lt;property name="scope" value="public"/&gt;
+      &lt;property name="accessModifiers" value="public"/&gt;
     &lt;/module&gt;
   &lt;/module&gt;
 &lt;/module&gt;
 </code></pre></div>
         <p id="Example2-code">
           This setting will report a violation if there
-          is no javadoc for <code>public</code> member.
+          is no javadoc for <code>public</code> field.
         </p>
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
 public class Example2 {
@@ -122,34 +115,32 @@ public class Example2 {
    */
   private int b;
   protected int c;
-  public int d; // violation
+  public int d; // violation, 'Missing a Javadoc comment'
   /*package*/ int e;
 }
 </code></pre></div>
 
         <p id="Example3-config">
-          To configure the check for members which are in <code>private</code>, but not in
-          <code>protected</code> scope:
+          To configure the check for fields which are in <code>private</code> or
+          <code>package</code> access modifier:
         </p>
 
         <div class="wrapper"><pre class="prettyprint"><code class="language-xml">
 &lt;module name="Checker"&gt;
   &lt;module name="TreeWalker"&gt;
     &lt;module name="JavadocVariable"&gt;
-      &lt;property name="scope" value="private"/&gt;
-      &lt;property name="excludeScope" value="protected"/&gt;
+      &lt;property name="accessModifiers" value="private,package"/&gt;
     &lt;/module&gt;
   &lt;/module&gt;
 &lt;/module&gt;
 </code></pre></div>
         <p id="Example3-code">
           This setting will report a violation if there is no
-          javadoc for <code>private</code> member and
-          ignores <code>protected</code> member.
+          javadoc for <code>private</code>  or <code>package</code> field.
         </p>
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
 public class Example3 {
-  private int a; // violation
+  private int a; // violation, 'Missing a Javadoc comment'
 
   /**
    * Some description here
@@ -157,11 +148,11 @@ public class Example3 {
   private int b;
   protected int c;
   public int d;
-  /*package*/ int e; // violation
+  /*package*/ int e; // violation, 'Missing a Javadoc comment'
 }
 </code></pre></div>
         <p id="Example4-config">
-          To ignore absence of Javadoc comments for variables with names <code>log</code> or
+          To ignore absence of Javadoc comments for fields with names <code>log</code> or
           <code>logger</code>:
         </p>
 
@@ -176,20 +167,56 @@ public class Example3 {
 </code></pre></div>
         <p id="Example4-code">
           This setting will report a violation if there is no
-          javadoc for any scope member and ignores members with
+          javadoc for any scope field and ignores fields with
           name <code>log</code> or <code>logger</code>.
         </p>
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
 public class Example4 {
-  private int a; // violation
+  private int a; // violation, 'Missing a Javadoc comment'
 
   /**
    * Some description here
   */
   private int b;
-  protected int c; // violation
-  public int d; // violation
-  /*package*/ int e; // violation
+  protected int c; // violation, 'Missing a Javadoc comment'
+  public int d; // violation, 'Missing a Javadoc comment'
+  /*package*/ int e; // violation, 'Missing a Javadoc comment'
+}
+</code></pre></div>
+
+        <p id="Example5-code">
+          This check will not report a violation for local and
+           anonymous inner classes with any configuration.
+        </p>
+        <div class="wrapper"><pre class="prettyprint"><code class="language-java">
+public class Example5 {
+  public int variablePublic; // violation, 'Missing a Javadoc comment'
+  protected int variableProtected; // violation, 'Missing a Javadoc comment'
+  int variablePackage; // violation, 'Missing a Javadoc comment'
+  private int variablePrivate; // violation, 'Missing a Javadoc comment'
+
+  public void testMethodInnerClass() {
+
+    // This check ignores local classes.
+    class InnerClass {
+      public int innerClassVariablePublic;
+      protected int innerClassVariableProtected;
+      int innerClassVariablePackage;
+      private int innerClassVariablePrivate;
+    }
+
+    // This check ignores anonymous inner classes.
+    Runnable runnable = new Runnable() {
+      public int innerClassVariablePublic;
+      protected int innerClassVariableProtected;
+      int innerClassVariablePackage;
+      private int innerClassVariablePrivate;
+      public void run()
+        {
+          System.identityHashCode("running");
+        }
+    };
+  }
 }
 </code></pre></div>
       </subsection>
--- a/src/site/xdoc/checks/javadoc/javadocvariable.xml.template
+++ b/src/site/xdoc/checks/javadoc/javadocvariable.xml.template
@@ -35,7 +35,7 @@
         </macro>
         <p id="Example1-code">
           By default, this setting will report a violation if
-          there is no javadoc for any scope member.
+          there is no javadoc for a field with any access modifier.
         </p>
         <macro name="example">
           <param name="path"
@@ -44,8 +44,8 @@
         </macro>
 
         <p id="Example2-config">
-          To configure the check for <code>public</code>
-          scope:
+          To configure the check <code>public</code>
+          access modifier:
         </p>
 
         <macro name="example">
@@ -55,7 +55,7 @@
         </macro>
         <p id="Example2-code">
           This setting will report a violation if there
-          is no javadoc for <code>public</code> member.
+          is no javadoc for <code>public</code> field.
         </p>
         <macro name="example">
           <param name="path"
@@ -64,8 +64,8 @@
         </macro>
 
         <p id="Example3-config">
-          To configure the check for members which are in <code>private</code>, but not in
-          <code>protected</code> scope:
+          To configure the check for fields which are in <code>private</code> or
+          <code>package</code> access modifier:
         </p>
 
         <macro name="example">
@@ -75,8 +75,7 @@
         </macro>
         <p id="Example3-code">
           This setting will report a violation if there is no
-          javadoc for <code>private</code> member and
-          ignores <code>protected</code> member.
+          javadoc for <code>private</code>  or <code>package</code> field.
         </p>
         <macro name="example">
           <param name="path"
@@ -84,7 +83,7 @@
           <param name="type" value="code"/>
         </macro>
         <p id="Example4-config">
-          To ignore absence of Javadoc comments for variables with names <code>log</code> or
+          To ignore absence of Javadoc comments for fields with names <code>log</code> or
           <code>logger</code>:
         </p>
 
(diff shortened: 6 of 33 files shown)


kkoutsilis:
For example in [InputJavadocVariableNoJavadoc2](https://github.com/checkstyle/checkstyle/blob/8e5396f11e39711d16420cbb704c5b2da232621c/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableNoJavadoc2.java#L13) line 72, surrounding scope is "package" while surrounding access modifier is "public" 
Screenshot 2024-12-24 at 18 52 58

kkoutsilis:
I believe the difference is coming from this extra check https://github.com/checkstyle/checkstyle/blob/8e5396f11e39711d16420cbb704c5b2da232621c/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java#L373

rdiachenko:
@kkoutsilis I might not have the full context, but what's the reason for using surrounding access modifiers in the first place? Why can't we use direct access modifiers only? That way the logic and usage would be more straightforward for users.

kkoutsilis:
> @kkoutsilis I might not have the full context, but what's the reason for using surrounding access modifiers in the first place? Why can't we use direct access modifiers only? That way the logic and usage would be more straightforward for users.

Hey @rdiachenko, thanks for your feedback. I am mainly trying to keep the same logic of how scope worked before, so since scope used surrounding Scope, I am trying to replicate that with access modifiers to keep it as consistent as possible. I am not sure what was the decision of using surrounding Scope, I can only make guesses. 
Your feedback is valid, I am not sure if I can take this kind of "design" decisions as I am fairly new contributor to the project but I am happy to go this down this path of not using surrounding access modifiers if you agree with it :) 
If not, we can go with the initial implementation for the abandoned PR that had a [method](https://github.com/checkstyle/checkstyle/pull/9277/files#diff-1bda71a340401c2d7c43026b4306d4b9bbc84515fa08c78d5e70e552cab47025R280) to map scope to access modifier, and this works fine. 

It stills bugs me that `getSurroundingScope` and and `getSurroundingAccessModifier` are inconsistent between them, and can't tell if that was intentional or a bug. Do you have any insights on that by any chance? Thank you in advance!

rdiachenko:
> Hey @rdiachenko, thanks for your feedback. I am mainly trying to keep the same logic of how scope worked before, so since scope used surrounding Scope, I am trying to replicate that with access modifiers to keep it as consistent as possible. I am not sure what was the decision of using surrounding Scope, I can only make guesses. Your feedback is valid, I am not sure if I can take this kind of "design" decisions as I am fairly new contributor to the project but I am happy to go this down this path of not using surrounding access modifiers if you agree with it :) If not, we can go with the initial implementation for the abandoned PR that had a [method](https://github.com/checkstyle/checkstyle/pull/9277/files#diff-1bda71a340401c2d7c43026b4306d4b9bbc84515fa08c78d5e70e552cab47025R280) to map scope to access modifier, and this works fine.
> 
> It stills bugs me that `getSurroundingScope` and and `getSurroundingAccessModifier` are inconsistent between them, and can't tell if that was intentional or a bug. Do you have any insights on that by any chance? Thank you in advance!

Scope and access modifier are two different concepts as for me. I don't even understand why we have `getSurroundingAccessModifier`. Let's make it simple for this check, we break backward compatibility anyway. If we need advanced filtering based on outer access modifier, let's do it as a separate step, when we have real use cases. As for now, I don't think of a use case where we would need the outer (surrounding modifier) context.

@romani do you have any input on this?

romani:
>but it seems there are use cases that these methods return different values (matching scope to access modifier)

As in any other source code base, until methods are not ideal and just serve some goal, that can be viewed by usage of methods. 
If for any reason such methods don't work for this Check, better to make new methods that work for this Check. As completely separate activity we can refactor other Checks to use new methods if it makes sense.

>surrounding scope is "package" while surrounding access modifier is "public"

This is long and heavy problem to define scope/visibility based on all upper classes.
I think we should look at class/type's fields  own visibility, completely ignoring all upper types visibility/modifiers.

To make it even more complicated, modern java has a lot not explicit visibilities, we have separate Check on this, and it is very complicated by its own.

-------

Let's not consider any non target visibility. On edge cases, users might use suppressions.

romani:
@kkoutsilis, let me know if you need any assistance

romani:
please rebase, I moved xdoc folder location.

kkoutsilis:
Hey @romani, I am away atm, will do as soon as I am back

kkoutsilis:
Hey @romani, I rebased. Will try to look this week into why the CI is failing but I aslo believe there are some requirements from the issue than I need to satisfy. Hopefully I will have some time to look into this.

romani:
All CI failures are due to breaking changes.
So we clearly see who is affected.
As we get closer to merge 
We need to send them PR with migration, so it will confirm that migration is possible without bad side effects.
See you soon, we appreciate your help.

kkoutsilis:
Hey @romani, I believe the PR can be reviewed

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
```diff
@@ -84,35 +81,28 @@ public class JavadocVariableCheck
...
-     * Setter to specify the visibility scope where Javadoc comments are not checked.
-     *
-     * @param excludeScope a scope.
-     * @since 3.4
-     */
-    public void setExcludeScope(Scope excludeScope) {
-        this.excludeScope = excludeScope;
+    public void setAccessModifiers(AccessModifierOption... accessModifiers) {
+        this.accessModifiers =
+            Arrays.copyOf(accessModifiers, accessModifiers.length);
```
rdiachenko, line 105: Do we actually need to make a copy here?

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheckTest.java
```diff
@@ -338,13 +339,47 @@ public void testLambdaLocalVariablesDoNotNeedJavadoc() throws Exception {
...
+
+    /**
+     * Test that when given an array, it creates its own instance of
+     * the array and is not re-using and possible overwriting the given one.
+     * Without this, pitest creates a new mutation on {@code Arrays.copyOf}.
+     *
+     * @throws Exception if an error occurs.
+     */
+    @Test
+    public void testCloneInAccessModifiersProperty() throws Exception {
```
rdiachenko, line 371: Do we need this test if we don't copy the array?

kkoutsilis: I don't believe we do

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
kkoutsilis: I made it this way to keep it consistent with the codebase https://github.com/checkstyle/checkstyle/blob/62bed57795d1c439aa5ad5adb8ca5c3d671bd62f/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/ParameterNameCheck.java#L120-L123  https://github.com/checkstyle/checkstyle/blob/62bed57795d1c439aa5ad5adb8ca5c3d671bd62f/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/RecordComponentNumberCheck.java#L114-L117
I guess it is done so it does not mess up accidentally with the passed down array. 
What would be the benefit of not copying ?

rdiachenko: > I made it this way to keep it consistent with the codebase

I see only two checks that use such approach

> What would be the benefit of not copying ?

If we don't copy it, we don't need an adhock unit test that doesn't bring much value. Additionally, I don't think the passed down array is used anywhere else outside this check, so such protection looks redundant to me.

kkoutsilis: Seems that there are checks in place to avoid this behaviour
```
[WARNING] PMD Failure: com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocVariableCheck:104 Rule:ArrayIsStoredDirectly Priority:3 The user-supplied array 'accessModifiers' is stored directly..
```
With a quick [SO search](https://stackoverflow.com/a/11581068), the solution is copying the array. 
Not sure if there is any other solution in order to avoid that extra test.

romani: As class receiving such array from outside, for crazy security rules it should be copied, as all  around are enemies.

kkoutsilis:
Hey @romani , while this PR is in review, do you want me to look into another issue?

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
rdiachenko: Ok, thanks for clarifying this @kkoutsilis. Does it make sense to use `copyOfArray` from `UnmodifiableCollectionUtil`? Will it help to avoid the unit test?

kkoutsilis: Looking at this note on the class https://github.com/checkstyle/checkstyle/blob/dae549f530608c833407925d4caac75a8624f9e3/src/main/java/com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.java#L30-L36
 I guess it does, I will give it a go. Thanks for the pointer @rdiachenko

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheckTest.java
kkoutsilis: Removed

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
kkoutsilis: Updated

rdiachenko (approved):
lgtm

romani:
@kkoutsilis , please start creating PRs to fix/migrate configurations in project that are in CI.

Please update respectful CI jobs to check out your fork of such repositories and checkout to your branch.

Ideal scenario: we just update config in their repositories. If any other violations happens we need to discuss.

Choose any repository/project you like and lets try to migrate it first.
Pin me for any help you need.

kkoutsilis:
> @kkoutsilis , please start creating PRs to fix/migrate configurations in project that are in CI.
> 
> Please update respectful CI jobs to check out your fork of such repositories and checkout to your branch.
> 
> Ideal scenario: we just update config in their repositories. If any other violations happens we need to discuss.
> 
> Choose any repository/project you like and lets try to migrate it first. Pin me for any help you need.

Hey @romani, just to make sure I inderstand correctly, we need to:
1. Fork the projects that are failing in CI and update the configuration. for example [spring-cloud-gcp](https://github.com/GoogleCloudPlatform/spring-cloud-gcp)
2. Update CI to checkout to fork/bracnh with updated config
3. Make sure that CI pass here

Is that correct or did I misunderstand ?

romani:
yes!
if you see that you are going to make a lot of work, pin me to review scope of updates, we shoould not be involved in something big, it might be sign that we are doing smething wrong.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
```diff
@@ -84,35 +82,28 @@ public class JavadocVariableCheck
...
     private Pattern ignoreNamePattern;
 
     /**
-     * Setter to specify the visibility scope where Javadoc comments are checked.
+     * Setter to access modifiers of methods where parameters are checked.
      *
-     * @param scope a scope.
-     * @since 3.0
+     * @param accessModifiers access modifiers of methods which should be checked.
+     * @since 11.0.0
```
kkoutsilis, line 102: @romani Just to doublechekc here as well, which is the target release version of this change? Added 11.0.0 as its a breaking change. What is the plan with releases?

romani: lets put here next release with major changes, 10.22.0 .
your changes are breaking changes, so after merge cI will be red to demand version bump to 10.22.0

kkoutsilis: Updated

kkoutsilis:
Hey @romani, fixed [no-error-pmd](https://app.circleci.com/pipelines/github/checkstyle/checkstyle/30093/workflows/38ceb275-b0b9-40a2-a1c2-57cb85b02e71/jobs/780545) check. [Chenge](https://github.com/pmd/build-tools/compare/main...kkoutsilis:build-tools:update-checkstyle-javadocvariable) in config was replacing `scope` with `accessModifiers` in build-tools and make sure that pmd [points](https://github.com/pmd/pmd/compare/main...kkoutsilis:pmd:update-checkstyle-javadocvariable) to the new config, so for this case it was the idieal scenario. Will probably move to [no-error-xwiki](https://app.circleci.com/pipelines/github/checkstyle/checkstyle/30093/workflows/38ceb275-b0b9-40a2-a1c2-57cb85b02e71/jobs/780548) next.

kkoutsilis:
Hey @romani, fixed  [no-error-xwiki](https://app.circleci.com/pipelines/github/checkstyle/checkstyle/30111/workflows/e57246c6-0e6f-40ac-8c83-a8f8277edce1/jobs/781300) as well. The story is a bit different here. Since we changed the behaviour of the check to not account for the surrounding scope, some javadocs or suppressions must be added. Here are the changes for [xwiki-platform](https://github.com/xwiki/xwiki-platform/compare/master...kkoutsilis:xwiki-platform:update-checkstyle-javadocvariable) and for [xwiki-commons](https://github.com/xwiki/xwiki-commons/compare/master...kkoutsilis:xwiki-commons:update-checkstyle-javadocvariable)

romani:
> spring-cloud-gcp

migrated to google style config.
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/869/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R264
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/868
very old version of checkstyle 9.3: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/751c57c209db740c80171d0a3e60eb9a33f1374a/spring-cloud-gcp-samples/pom.xml#L25

we can not support that old version.


We need to drop spring-cloud-gcp from our CI, it is too heavy to migrate them to most recent version.

@elefeint , fyi, but let us know when you be able to catch up to recent version, so we can  restore it in our CI.

romani:
@kkoutsilis , please send PRs to xwiki and pmd.
rebase on latest,

to avoid whilelist polution that is requested at https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=26005&view=logs&j=6a833393-c551-5000-0916-5000fd3b86e7&t=bdbbd129-77f0-5d73-ce7d-db58d34c8358&l=55

please use this trick:
` git checkout "66d""ed33c74662cb3da612f3d34a5ae""fa""a629b443"`
your user name can be added to whiletilst file.

elefeint:
@meltsufin FYI, checkstyle needs to be upgraded in spring-cloud-gcp

kkoutsilis:
@romani can you re-run [no-error-xwiki](https://app.circleci.com/pipelines/github/checkstyle/checkstyle/30340/workflows/7f8a0abf-e90c-4a7d-8c44-b4cf1c77f764/jobs/790912) please so I don't re-run the entire Ci

.ci/validation.sh
```diff
@@ -758,7 +767,6 @@ no-error-spring-cloud-gcp)
   echo "Checkout target sources ..."
   checkout_from https://github.com/googlecloudplatform/spring-cloud-gcp
   cd .ci-temp/spring-cloud-gcp
-  git checkout "7c99f37087ac8f""eb""db""f7e185375a217b744895a4"
```
romani, line 761: please removev whole  `no-error-spring-cloud-gcp)`
and remove https://github.com/checkstyle/checkstyle/blob/9b463180f0b96a7317f92484770f186cada3d910/.semaphore/semaphore.yml#L92

we keep in our CI only projects what allows us to overide version and has no violations from CS on each commit.

romani (changes requested):
item:

romani:
@kkoutsilis , please send PRs to xwiki and pmd to get user feedback on update.

kkoutsilis:
Hey @romani, I've already sent the prs

romani:
please share links here to let subscribe to them.

kkoutsilis:
- [build-tools PR](https://github.com/pmd/build-tools/pull/55)
- [xwiki-commons PR](https://github.com/xwiki/xwiki-commons/pull/1241)
- [xwiki-platforms PR](https://github.com/xwiki/xwiki-platform/pull/3907)

.ci/validation.sh
kkoutsilis: [Done](https://github.com/checkstyle/checkstyle/compare/b8b8813b358eeac778e4ba7a6ba7918d859963aa..7c776e0a49a6a668088b9bbcecf74742adb99128)

romani:
Please apply https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=26215&view=logs&j=6a833393-c551-5000-0916-5000fd3b86e7&t=bdbbd129-77f0-5d73-ce7d-db58d34c8358&l=46

romani:
@kkoutsilis , please confirm by cli or Input that your fix covers https://github.com/checkstyle/checkstyle/issues/3183
I found it in issue description.

romani (approved):
thanks a lot !!!!

romani:
@kkoutsilis , please glance at xwiki failure in CI

kkoutsilis:
> @kkoutsilis , please glance at xwiki failure in CI

looking at it rn, its because of recent changes at master, ci was pointing to another fork with new changes, fix on the way

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
```diff
@@ -84,35 +82,28 @@ public class JavadocVariableCheck
...
+        AccessModifierOption.PACKAGE,
+        AccessModifierOption.PRIVATE,
+    };
 
     /** Specify the regexp to define variable names to ignore. */
     private Pattern ignoreNamePattern;
 
     /**
-     * Setter to specify the visibility scope where Javadoc comments are checked.
+     * Setter to access modifiers of methods where parameters are checked.
```
nrmancuso, line 99: Why we do we keep mentioning `methods`? This makes no sense. 
```suggestion
     * Setter specify the access modifiers of fields to check.
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
```diff
@@ -84,35 +82,28 @@ public class JavadocVariableCheck
...
      *
-     * @param scope a scope.
-     * @since 3.0
+     * @param accessModifiers access modifiers of methods which should be checked.
```
nrmancuso, line 101: Same as above

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
```diff
@@ -176,6 +167,17 @@ private boolean isIgnored(DetailAST ast) {
             || "serialVersionUID".equals(name);
     }
 
+    /**
+     * Checks whether a method has the correct access modifier to be checked.
+     *
+     * @param accessModifier the access modifier of the method.
+     * @return whether the method matches the expected access modifier.
+     */
+    private boolean matchAccessModifiers(final AccessModifierOption accessModifier) {
```
nrmancuso, line 176:
```suggestion
    private boolean matchAccessModifiers(AccessModifierOption accessModifier) {
```

Please remove `final` from the parameter declaration, this is an outdated practice.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
```diff
@@ -185,14 +187,15 @@ private boolean isIgnored(DetailAST ast) {
...
-                    || !customScope.isIn(excludeScope)
-                    || !surroundingScope.isIn(excludeScope));
+            try {
+                final AccessModifierOption accessModifier =
+                    CheckUtil.getAccessModifierFromModifiersToken(ast);
+                result = matchAccessModifiers(accessModifier);
+            }
+            catch (IllegalArgumentException ex) {
+                result = false;
+            }
```
nrmancuso, line 197: Please explain why we are using a try/catch here, this seems hacky.

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/javadoc/JavadocVariableCheck.xml
```diff
@@ -8,18 +8,15 @@
...
  &lt;/div&gt;</description>
          <properties>
-            <property name="excludeScope" type="com.puppycrawl.tools.checkstyle.api.Scope">
-               <description>Specify the visibility scope where Javadoc
- comments are not checked.</description>
+            <property default-value="public, protected, package, private"
+                      name="accessModifiers"
+                      type="com.puppycrawl.tools.checkstyle.checks.naming.AccessModifierOption[]">
+               <description>Access modifiers of methods where parameters are
+ checked.</description>
```
nrmancuso, line 15: Same as above, this doesn't make any sense. Please update everywhere that mentions methods and parameters, this check has nothing to do with these.

src/site/xdoc/checks/javadoc/javadocvariable.xml
```diff
@@ -78,33 +71,33 @@
 </code></pre></div>
         <p id="Example1-code">
           By default, this setting will report a violation if
-          there is no javadoc for any scope member.
+          there is no javadoc for a member with any access modifier.
```
nrmancuso, line 74:
```
By default, this configuration reports a violation for any field, regardless of its visibility, if it is missing a Javadoc.
```

src/site/xdoc/checks/javadoc/javadocvariable.xml
```diff
@@ -78,33 +71,33 @@
...
+  /*package*/ int e; // violation, 'Missing a Javadoc comment'
 
 }
 </code></pre></div>
 
         <p id="Example2-config">
-          To configure the check for <code>public</code>
-          scope:
+          To configure the check <code>public</code>
+          access modifier:
```
nrmancuso, line 93:
```suggestion
To configure the check to report violations for fields with the <code>public</code> access modifier that are missing Javadocs:
```

src/site/xdoc/checks/javadoc/javadocvariable.xml
```diff
@@ -122,42 +115,40 @@ public class Example2 {
...
+  public int d; // violation, 'Missing a Javadoc comment'
   /*package*/ int e;
 }
 </code></pre></div>
 
         <p id="Example3-config">
-          To configure the check for members which are in <code>private</code>, but not in
-          <code>protected</code> scope:
+          To configure the check for members which are in <code>private</code> or
+          <code>package</code> access modifier:
```
nrmancuso, line 125:
```suggestion
To configure the check for fields with the <code>private</code or package-private access modifier:
```

`<code>package</code>` should not be in code tags, this is not valid code.

src/site/xdoc/checks/javadoc/javadocvariable.xml.template
```diff
@@ -35,7 +35,7 @@
         </macro>
         <p id="Example1-code">
           By default, this setting will report a violation if
-          there is no javadoc for any scope member.
+          there is no javadoc for a member with any access modifier.
```
nrmancuso, line 38: Please change all mention of `member` or `variable` to `field` in the docs to be consistent.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableInner.java
```diff
@@ -1,11 +1,9 @@
 /*
 JavadocVariable
-scope = (default)private
-excludeScope = (default)null
+accessModifiers = (default)private
```
nrmancuso, line 3: Isn't the default all access modifiers? This doesn't make sense. @romani do we have a ticket somewhere to add validation for this?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableInner2.java
```diff
@@ -16,24 +14,22 @@
...
-        public int fData;
+        int fData;
     }
 
     // Ignore - 2 violations
     interface InnerInterface2
     {
-        // Ignore - should be all upper case
+
+        /** Ignore - 1 violation */
```
nrmancuso, line 26: This javadoc is confusing here. Why does this say "Ignore"?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableInterfaceMemberScopeIsPublic.java
```diff
@@ -15,16 +13,16 @@ public interface InputJavadocVariableInterfaceMemberScopeIsPublic {
...
-    int field3 = 0; // violation
+    int field3 = 0; // violation, 'Missing a Javadoc comment'
 
     enum Enum {
 
         /** First constant */
         A,
 
-        B; // violation
+        B;
```
nrmancuso, line 25: The constant above and this one should be a violation; it is equivalent to:
```
public static final Enum B = new Enum();
```

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableMethodInnerClass.java
```diff
@@ -0,0 +1,32 @@
...
+package com.puppycrawl.tools.checkstyle.checks.javadoc.javadocvariable;
+
+public class InputJavadocVariableMethodInnerClass {
+    public int variablePublic; // violation, 'Missing a Javadoc comment'
+    protected int variableProtected; // violation, 'Missing a Javadoc comment'
+    int variablePackage; // violation, 'Missing a Javadoc comment'
+    private int variablePrivate; // violation, 'Missing a Javadoc comment'
+
+    public void testMethodInnerClass() {
+        class InnerClass {
```
nrmancuso, line 15:
```suggestion
        // this check ignores locally declared classes
        class InnerClass {
```

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableMethodInnerClass.java
```diff
@@ -0,0 +1,32 @@
...
+            public int innerClassVariablePublic;
+            protected int innerClassVariableProtected;
+            int innerClassVariablePackage;
+            private int innerClassVariablePrivate;
+        }
+        Runnable runnable = new Runnable() {
```
nrmancuso, line 21:
```suggestion
        // this check ignores anonymous inner classes
        Runnable runnable = new Runnable() {
```

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableMethodInnerClass.java
nrmancuso, line 1: We need to add both of these examples in the above comments to the documentation to make it clear that we don't check these.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableNoJavadoc4.java
```diff
@@ -1,32 +1,30 @@
 /*
 JavadocVariable
-scope = (default)private
-excludeScope = (default)null
+accessModifiers = (default)private
```
nrmancuso, line 3: Incorrect default, again

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableNoJavadoc5.java
```diff
@@ -1,32 +1,30 @@
 /*
 JavadocVariable
-scope = (default)private
-excludeScope = (default)null
+accessModifiers = (default)private
```
nrmancuso, line 3: Incorrect default, again

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableNoJavadocNeededInLambda.java
```diff
@@ -1,19 +1,18 @@
 /*
 JavadocVariable
-scope = (default)private
-excludeScope = (default)null
+accessModifiers = (default)private
```
nrmancuso, line 3: Incorrect default again

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableOnInnerClassFields.java
```diff
@@ -1,32 +1,30 @@
 /*
 JavadocVariable
-scope = (default)private
-excludeScope = (default)null
+accessModifiers = (default)private
```
nrmancuso, line 3: Same as above

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableOnPublicInnerClassFields.java
```diff
@@ -1,31 +1,29 @@
 /*
 JavadocVariable
-scope = (default)private
-excludeScope = (default)null
+accessModifiers = (default)private
```
nrmancuso, line 3: Same as above

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariablePublicOnly.java
```diff
@@ -1,11 +1,9 @@
 /*
 JavadocVariable
-scope = (default)private
-excludeScope = (default)null
+accessModifiers = (default)private
```
nrmancuso, line 3: Same as above

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableTags.java
```diff
@@ -1,11 +1,9 @@
 /*
 JavadocVariable
-scope = (default)private
-excludeScope = (default)null
+accessModifiers = (default)private
```
nrmancuso, line 3: Same as above

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableTags.java
```diff
@@ -308,14 +306,14 @@ public int foo(Object _arg) {
...
-    CONSTANT_A, // violation
+    CONSTANT_A,
 
     /**
      *
      */
     CONSTANT_B,
 
-    CONSTANT_C // violation
+    CONSTANT_C
```
nrmancuso, line 316: These should both be violations (CONSTANT_A and CONSTANT_C)

nrmancuso (changes requested):
First pass, items:

.ci/validation.sh
```diff
@@ -216,12 +216,17 @@ no-error-pmd)
   echo "CS_version: ${CS_POM_VERSION}"
   mvn -e --no-transfer-progress clean install -Pno-validations
   echo "Checkout target sources ..."
-  checkout_from "https://github.com/pmd/build-tools.git"
+  #  checkout_from "https://github.com/pmd/build-tools.git"
+  checkout_from "https://github.com/kkoutsilis/build-tools.git"
```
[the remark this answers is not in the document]
romani: Yes, we can not wait. We will be back ones they merge.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableInner.java
romani: Validation is already in place, we might have defect in it. It is possible.

.ci/validation.sh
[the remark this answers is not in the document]
nrmancuso: >Yes, we can not wait. We will be back ones they merge.

How will we track this?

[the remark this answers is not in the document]
romani: Branch will be removed after merge, and CI will fail. I am subscribed to such PR, so eventually I will migrate back.

We can create issue but, still bunch of room to miss. What i described above usually works.

romani:
@kkoutsilis , please address review items, lets me know if something is complicated.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
kkoutsilis: An IllegalArgumentException is thrown in [getAccessModifierFromModifiersTokenDirectly](https://github.com/checkstyle/checkstyle/blob/d4025b26159e5e386eaeb3abd4011bd990b39284/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java#L341-L345) which is used in `getAccessModifierFromModifiersToken`. As I tried to re-use existing functionality, the exception is caught here and the check is skipped, as this is what should happen to the best of my understanding.

kkoutsilis: I believe this actually is causing the issue with the Enum ([ref](https://github.com/checkstyle/checkstyle/pull/16049#discussion_r1966988478)), looking into it.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableInterfaceMemberScopeIsPublic.java
kkoutsilis: This comes up as an `ENUM_CONSTANT_DEF`, and we check for `VARIABLE_DEF`. I guess we still want to check for this case?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableInner2.java
kkoutsilis: It already exists in the same file a bit lower https://github.com/checkstyle/checkstyle/blob/ad015edbde5a75fd817a49738a529d88dfbc6c67/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableInner2.java#L66 . Don't believe it has a specific reason to have Ignore in it.

romani:
Result of test execution should be in Commit.
Test updates all for xdoc
https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=26570&view=logs&j=9c723939-9148-5f6a-3101-ca017cb22518&t=c03cca9e-b908-564b-f216-90279d42b489&l=1963

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableTags.java
kkoutsilis: fixed

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableTags.java
kkoutsilis: fixed

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariablePublicOnly.java
kkoutsilis: fixed

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableOnInnerClassFields.java
kkoutsilis: fixed

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableNoJavadocNeededInLambda.java
kkoutsilis: fixed

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableNoJavadoc5.java
kkoutsilis: fixed

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableNoJavadoc4.java
kkoutsilis: fixed

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableMethodInnerClass.java
kkoutsilis: Added

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableMethodInnerClass.java
kkoutsilis: Added

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableInterfaceMemberScopeIsPublic.java
kkoutsilis: fixed

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
kkoutsilis: removed

kkoutsilis:
@nrmancuso thanks for your thorough review, appreciate it. I have addressed most, if not all, of your comments. Please have another look when you have some time and let me know if you are happy or if there is anything else to discuss/change.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
nrmancuso: >Is there any concern with this approach from your side?

Absolutely. Exceptions should happen when something exceptional occurs. If CheckUtil#getAccessModifierFromModifiersToken throws an exception when trying to get the modifiers from an enum constant, this is a bug and we should update CheckUtil#getAccessModifierFromModifiersToken.

I presume that CheckUtil#getAccessModifierFromModifiersToken handles other implied modifiers, like on fields in interfaces, right? If so, it should also handle implied modifiers on enum constants.

src/site/xdoc/checks/javadoc/javadocvariable.xml
nrmancuso: @kkoutsilis missed this one, you should be replying to each and every review item so that we do not miss anything.

src/site/xdoc/checks/javadoc/javadocvariable.xml
nrmancuso: Also missed this one

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableInner2.java
```diff
@@ -16,24 +14,23 @@
...
-        public int fData;
+        int fData;
     }
 
     // Ignore - 2 violations
     interface InnerInterface2
     {
-        // Ignore - should be all upper case
+
+        /** Ignore - should be all upper case. */
```
nrmancuso, line 26: Please remove this comment, it is nonsense.

nrmancuso (changes requested):
Items from previous review that were missed:
 - https://github.com/checkstyle/checkstyle/pull/16049#discussion_r1966983412
 - https://github.com/checkstyle/checkstyle/pull/16049#discussion_r1966984545

Item to fix/improve: https://github.com/checkstyle/checkstyle/pull/16049#discussion_r1966978443

New item:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableInner2.java
kkoutsilis: [Removed](https://github.com/checkstyle/checkstyle/compare/d952ec79ee0f7bb07c23ffe1ff4dced69bc3e931..d17f97b70657f159ff1a5f44ac500e0f492b975d)

src/site/xdoc/checks/javadoc/javadocvariable.xml
kkoutsilis: Sorry for that, I thought I addressed that last time. 
[Addressed](https://github.com/checkstyle/checkstyle/compare/d952ec79ee0f7bb07c23ffe1ff4dced69bc3e931..d17f97b70657f159ff1a5f44ac500e0f492b975d)

src/site/xdoc/checks/javadoc/javadocvariable.xml
kkoutsilis: [Addressed](https://github.com/checkstyle/checkstyle/compare/d952ec79ee0f7bb07c23ffe1ff4dced69bc3e931..d17f97b70657f159ff1a5f44ac500e0f492b975d)

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
kkoutsilis: @nrmancuso You are right, thanks for pointing that out.
 To summarize: 
- [CheckUtil#getAccessModifierFromModifiersToken](https://github.com/checkstyle/checkstyle/blob/d4025b26159e5e386eaeb3abd4011bd990b39284/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java#L316-L330) is trying to `findFirstToken` of token type MODIFIERS
- For ENUM_CONSTANT_DEF, there is no MODIFIERS token type, instead, there is [ANNOTATIONS](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/TokenTypes.java#L5033-L5065)
- As per [Java Language Specification](https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.9.3) (and also [pointed out by you](https://github.com/checkstyle/checkstyle/pull/16049#discussion_r1966988478)), enum constants are implicitly  declared `public static final`
- For that, `getAccessModifierFromModifiersToken` needs to be updated to handle this use case

Hopefully, I got that right. 
The code has been [updated](https://github.com/checkstyle/checkstyle/compare/d952ec79ee0f7bb07c23ffe1ff4dced69bc3e931..d17f97b70657f159ff1a5f44ac500e0f492b975d). 
Will be waiting for your feedback and thanks again 🙏

nrmancuso (approved):
Good to merge after CI is green

@kkoutsilis pitest is failing, please fix this

romani:
Yes, hard blocker.
Extra test is required, we did and keep  doing huge  effort to kill all pitest. Let us know if there is problems.

kkoutsilis:
@romani @nrmancuso Made a small change to `getAccessModifierFromModifiersToken` to remove the pitest mutation, please have another look and let me know if this is fine. Thank you

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
```diff
@@ -314,9 +314,14 @@ public static boolean isReceiverParameter(DetailAST parameterDefAst) {
      * @return the access modifier of the method/constructor.
      */
     public static AccessModifierOption getAccessModifierFromModifiersToken(DetailAST ast) {
-        final DetailAST modsToken = ast.findFirstToken(TokenTypes.MODIFIERS);
-        AccessModifierOption accessModifier =
-                getAccessModifierFromModifiersTokenDirectly(modsToken);
+        AccessModifierOption accessModifier;
+        if (ast.getType() == TokenTypes.ENUM_CONSTANT_DEF) {
+            accessModifier = AccessModifierOption.PUBLIC;
```
romani, line 319: Why we think it is always public?
Please share a bit more details, thanks in advance.

romani (changes requested):
Question

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
nrmancuso: Enum constants are always public, static, and final. This cannot be changed/overridden in any way.

romani: Yes, I read it as enum itself.

romani (approved):
Let's merge this huge update.

romani:
@kkoutsilis , please send revert on one your fork usage. Migration is done on their side.

PR #16083, opened 2024-12-29
Removed ok comments from all of the java imput files (case insensitiv…

Akshat-Dimri:
…e) ,using a python script

# PLEASE READ before removing

Rules:
1) The issue you are trying to fix/resolve has to have the "approved" label.
2) Put in the description of Pull Request the reference to an issue if it exists.
   Example: "Issue: #XXXXXX"
3) Commit message should adhere to the following rules:
   a) MUST match one of the following patterns:
      ^Issue #\\d+: .*$
      ^Pull #\\d+: .*$
      ^(minor|config|infra|doc|spelling|dependency): .*$
   b) MUST contain only one line of text
   c) MUST NOT end with a period, space, or tab
   d) MUST be less than or equal to 200 characters

To avoid multiple iterations of fixes and CIs failures, please read
https://checkstyle.org/contributing.html

ATTENTION: We are not merging Pull Requests that are not passing our CIs,
but we will help to resolve issues.

Thanks for reading, remove whole this message and type what you need.


remove_ok_only_comments.py
```diff
@@ -0,0 +1,43 @@
+import os
```
romani, line 1: Please remove this file from commit.
We don't need it, one migration complete, this script be useless.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorSingleCommentSeparatedFromPackage.java
```diff
@@ -12,7 +12,6 @@
 
 package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator;
 
-// ok
 
 public class InputEmptyLineSeparatorSingleCommentSeparatedFromPackage // ok
```
romani, line 16: Remove trailing comments

romani (changes requested):
Tests might fail, as removal of line , might change line numbers in expected arrays of Tests.

Better to search-replace `// ok` to empty line 

Items:

remove_ok_only_comments.py
Akshat-Dimri: so can it be done like i execute the script , which replaces all of the ok comments with blank lines and then commit after removing the script ?

romani: Yes.
Just not as script to commit.
Keep it only to change files.

romani:
Please read and watch videos at https://checkstyle.org/beginning_development.html#Starting_Development
Please make CI green.

single commit please.

But direction is good, please prepare PR for merge.

Issue #16127, opened 2025-01-07
The SarifLogger creates invalid JSON on Windows with unescaped paths


kmoschcau:
I have read check documentation: https://checkstyle.org/checks/xxxxxx/nameofaffectedcheck.html
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```ps1
$ Get-Content foo.java
class foo {
  public String[] bar;
  public foo (String[] param) {
    bar = param;
  }
}

$ java '-Duser.language=en' '-Duser.country=US' -jar .\checkstyle-10.21.1-all.jar -f sarif -c /google_checks.xml foo.java
{
  "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
  "version": "2.1.0",
  "runs": [
    {
      "tool": {
        "driver": {
          "downloadUri": "https://github.com/checkstyle/checkstyle/releases/",
          "fullName": "Checkstyle",
          "informationUri": "https://checkstyle.org/",
          "language": "en",
          "name": "Checkstyle",
          "organization": "Checkstyle",
          "rules": [
          ],
          "semanticVersion": "10.21.1",
          "version": "10.21.1"
        }
      },
      "results": [
        {
          "level": "warning",
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "C:\Users\KaiMoschcau\Downloads\foo.java"
                },
                "region": {
                  "startColumn": 7,
                  "startLine": 1
                }
              }
            }
          ],
          "message": {
            "text": "Type name 'foo' must match pattern '^[A-Z][a-zA-Z0-9]*$'."
          },
          "ruleId": "name.invalidPattern"
        },
        {
          "level": "warning",
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "C:\Users\KaiMoschcau\Downloads\foo.java"
                },
                "region": {
                  "startColumn": 3,
                  "startLine": 3
                }
              }
            }
          ],
          "message": {
            "text": "'CTOR_DEF' should be separated from previous line."
          },
          "ruleId": "empty.line.separator"
        },
        {
          "level": "warning",
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
[...]

PR #16147, opened 2025-01-11
Issue #16127: Correctly handle windows file paths in SARIF and add a doc link to nvim-lint.

kmoschcau:
Issue #16127

I am not sure this covers all cases that the URIs will always be valid, but I didn't find a Java method that doesn't change relative paths to absolute paths automatically. I have already spent most of the day on this, because the test tooling for this project is quite slow.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
@@ -30,6 +30,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.regex.Pattern;
 
 import com.puppycrawl.tools.checkstyle.api.AuditEvent;
 import com.puppycrawl.tools.checkstyle.api.AuditListener;
@@ -75,6 +76,19 @@ public class SarifLogger extends AbstractAutomaticBean implements AuditListener 
 
     /** The placeholder for results. */
     private static final String RESULTS_PLACEHOLDER = "${results}";
+
+    /** Two backslashes to not duplicate strings. */
+    private static final String TWO_BACKSLASHES = "\\\\";
+
+    /** A pattern for two backslashes. */
+    private static final Pattern A_SPACE_PATTERN = Pattern.compile(" ");
+
+    /** A pattern for two backslashes. */
+    private static final Pattern TWO_BACKSLASHES_PATTERN = Pattern.compile(TWO_BACKSLASHES);
+
+    /** A pattern to match a file with a Windows drive letter. */
+    private static final Pattern WINDOWS_DRIVE_LETTER_PATTERN =
+            Pattern.compile("\\A[A-Z]:", Pattern.CASE_INSENSITIVE);
 
     /** Helper writer that allows easy encoding and printing. */
     private final PrintWriter writer;
@@ -174,7 +188,7 @@ public class SarifLogger extends AbstractAutomaticBean implements AuditListener 
         if (event.getColumn() > 0) {
             results.add(resultLineColumn
                 .replace(SEVERITY_LEVEL_PLACEHOLDER, renderSeverityLevel(event.getSeverityLevel()))
-                .replace(URI_PLACEHOLDER, event.getFileName())
+                .replace(URI_PLACEHOLDER, renderFileNameUri(event.getFileName()))
                 .replace(COLUMN_PLACEHOLDER, Integer.toString(event.getColumn()))
                 .replace(LINE_PLACEHOLDER, Integer.toString(event.getLine()))
                 .replace(MESSAGE_PLACEHOLDER, escape(event.getMessage()))
@@ -184,7 +198,7 @@ public class SarifLogger extends AbstractAutomaticBean implements AuditListener 
         else {
             results.add(resultLineOnly
                 .replace(SEVERITY_LEVEL_PLACEHOLDER, renderSeverityLevel(event.getSeverityLevel()))
-                .replace(URI_PLACEHOLDER, event.getFileName())
+                .replace(URI_PLACEHOLDER, renderFileNameUri(event.getFileName()))
                 .replace(LINE_PLACEHOLDER, Integer.toString(event.getLine()))
                 .replace(MESSAGE_PLACEHOLDER, escape(event.getMessage()))
                 .replace(RULE_ID_PLACEHOLDER, event.getViolation().getKey())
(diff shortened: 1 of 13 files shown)


romani:
https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=25346&view=logs&j=6a833393-c551-5000-0916-5000fd3b86e7&t=bdbbd129-77f0-5d73-ce7d-db58d34c8358&l=52

Please apply this patch command to let CI pass.

Xwiki is not related, should be green on next run.

Please fix idea inspection violation https://app.circleci.com/pipelines/github/checkstyle/checkstyle/29502/workflows/0bcba9e7-1c7c-44f7-ad00-48ac1189a3d5/jobs/755938/artifacts

romani:
@kmoschcau , I rebased in web your code, please find time to fix CI violations.

kmoschcau:
Hi, I just didn't get to it yet. Hopefully I will have enough time to do it this weekend.

romani:
Please squash all commits in one.

kmoschcau:
I did all required fixes.

romani:
But we need single commit, can you squash?

kmoschcau:
Done.

romani:
GitHub, generate website

romani (approved):
Thanks a lot for fix

src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
```diff
@@ -222,6 +236,23 @@ public void fileFinished(AuditEvent event) {
...
+     */
+    private static String renderFileNameUri(final String fileName) {
+        String normalized =
+                A_SPACE_PATTERN
+                        .matcher(TWO_BACKSLASHES_PATTERN.matcher(fileName).replaceAll("/"))
+                        .replaceAll("%20");
+        if (WINDOWS_DRIVE_LETTER_PATTERN.matcher(normalized).find()) {
+            normalized = '/' + normalized;
+        }
+        return "file://" + normalized;
```
rnveach, line 253: This seems more like incomplete code with assumptions and doesn't seem on the verge of unreasonable. Is this because we don't have access to the file object and only the string? Would a file object allow us to use built-in transformers?

According to ChatGPT, relative paths are only suppose to have `file:` (`file:Documents/file.txt`). Linux files with a `\` should be converted to % code (`/home/user/My Documents/file\name.txt` => `file:///home/user/My%20Documents/file%5Cname.txt`).
According to DeepSeek, relative path paths are not valid at all, which raises more concerns that this should all fail if it is given a relative path.
Both agree that special characters should be converted to % code, which you only do for spaces, but the AIs said `#` and other characters should be converted.

In Linux you can have a file named `C:blah` which your code mistakenly thinks is Windows.

Please provide documentation on what is the correct handling here, or it seems these changes may be incomplete and still issue prone.

src/test/resources/com/puppycrawl/tools/checkstyle/sariflogger/ExpectedSarifLoggerAbsoluteLinuxPath.sarif
```diff
@@ -0,0 +1,44 @@
...
+          "downloadUri": "https://github.com/checkstyle/checkstyle/releases/",
+          "fullName": "Checkstyle",
+          "informationUri": "https://checkstyle.org/",
+          "language": "en",
+          "name": "Checkstyle",
+          "organization": "Checkstyle",
+          "rules": [
+          ],
+          "semanticVersion": "null",
+          "version": "null"
```
rnveach, line 17: Just pointing out this isn't a valid sarif file because of this `null`.

rnveach (changes requested):

src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
```diff
@@ -86,7 +86,7 @@ public void testAddError() throws IOException {
                 new Violation(1, 1,
                         "messages.properties", "ruleId", null, SeverityLevel.ERROR, null,
                         getClass(), "found an error");
-        final AuditEvent ev = new AuditEvent(this, "Test.java", violation);
+        final AuditEvent ev = new AuditEvent(this, "./Test.java", violation);
```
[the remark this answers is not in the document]
kmoschcau: Yeah that's on me. I thought URIs didn't allow for immediately starting with just a path segment. I will revert the tests and adjust the algorithm.

src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
kmoschcau: I am basing these changes on the [Sarif-Spec](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790680) and on [RFC 3986](https://www.rfc-editor.org/info/rfc3986) for URIs.

My original intent was just to get rid of invalid JSON, which was caused by checkstyle writing paths like this into the output: `"C:\foo\bar.java"
At first I simply wanted to escape the backslashes, but according to the above linked specs, those should at least be forward slashes.

I agree that having proper file libraries and a proper JSON serializer would be better, but none of these are present here. The Sarif logger hand rolls its output and only receives a simple string from the event.

I tried using various built-in methods to normalize the paths, but all that I have tried did have some side effects like forcing absolute paths.

I just read the RFC again and relative paths don't even need the scheme. They can just come in the format of `foo.java`. And relative paths *are* allowed. They are given as examples in both specs.

I agree that the current implementation needs some additional thought put into it. I just didn't have the time to invest, since no built-in function worked and the project tooling takes ages to run.

src/test/resources/com/puppycrawl/tools/checkstyle/sariflogger/ExpectedSarifLoggerAbsoluteLinuxPath.sarif
kmoschcau: True, though that was already the case in the existing files. From what I can see it would require changing all of the tests and a considerably more involved test setup.

src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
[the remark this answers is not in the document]
kmoschcau: Actually reading the RFC again, it would be a breaking change. The scheme part of the URI is required.

kmoschcau:
I have no idea how much file path handling even happens before this point in checkstyle. I didn't think this would be that complicated a fix. I don't intend to build a full file URI handling solution just for this one output format. If there is nothing ready made that we can use, I'm going to leave this for someone else to pick up.

src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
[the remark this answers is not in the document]
romani: such changes are reverted, not a concern any more.

src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
romani: implementation is not ideal and definitely need improvement in future,
but it is step forward.
We will improve in future as users becomes so diverse in running us in all possible situations on this report type.

romani:
Energy is lost in this PR, so I am merging as it is now, we will improve in future to make it better.

PR #16168, opened 2025-01-14
Issue #13345: Enable example tests for SuppressWithPlainTextCommentFi…

ElinaZoldnere:
Issue #13345
Enable example tests for SuppressWithPlainTextCommentFilter

- converted .txt to .java + .properties/.sql/.xml
- changed FileTabCharacter checks to visible checks
- added missing code examples
- enabled tests

[https://checkstyle.sourceforge.io/filters/suppresswithplaintextcommentfilter.html#Examples](https://checkstyle.sourceforge.io/filters/suppresswithplaintextcommentfilter.html#Examples)


ElinaZoldnere:
@romani
Opened PR just to ask a question:
do I understand correctly that in examples we don't dive into details of suppressed violations?
And the reader should see something like this from Example2.properties
```
# BEGIN GENERATED CONTENT
# suppressed violation below
my.property=value1
my.property=value2
# END GENERATED CONTENT
```
And we also do not verify suppressed violations in the tests, only the final outcome?
Or should we?

romani:
GitHub, generate website

romani:
>don't dive into details of suppressed violations?

Yes, just need a fact that there was violation and by filter it was suppressed.

>we also do not verify suppressed violations in the tests, only the not suppressed?
Or should we?

Ideally we should validate. But is scope of this issue better to avoid as examples are small, and it is easy to verify them on this.
I think we do have special training comments in test for this.

romani (changes requested):
Please add target file content after default config.

We should always have pairs (config+target file).

ElinaZoldnere:
@romani  I have an Example containing a tab character, but when the Example is embedded into .xml, I get  checkstyle error about .xml file containing a tab character.
I see it is somehow solved in original FileTabCharacters examples and xml, likely replacing tabs with whitespaces, but I can not find where it is configured.

ElinaZoldnere:
GitHub, generate website

romani:
@ElinaZoldnere, validation on invisible symbols is hard to show, I am ok to drop such example or simply change violation/Check to something visible.

ElinaZoldnere:
GitHub, generate website

ElinaZoldnere:
Please see the changes. Changing invisible occurrences to something visible, adding missing examples, and trying to put it all together resulted in noticeably different code snippets. I hope it is OK.
All suppression config cases stay the same.
Also, I tried to group examples into somewhat comparable groups to avoid distracting the reader with a variety of  examples and to keep the focus on the essential config differences, while covering all the file types (properties, xml, sql, java).

ElinaZoldnere:
GitHub, generate website

romani (approved):
supper !!!

thanks a lot for hard work

PR #16212, opened 2025-01-21
Issue #16077: made tables single column for mobile devices and removed .bodyTable class reference

Zopsss:
part of #16077 

fixed following:

> mobile web site should have tables to be single column.

all tables had `tableBody` class in previous theme. Currently it is replaced with two classes: `table table-striped`.

In this PR, I removed all the `tableBody` references and replaced them with the tag name `table`, I think there should not be any problems with this change. I checked the website locally and it was looking fine to me.

### DIFF
--- a/src/site/resources/css/site.css
+++ b/src/site/resources/css/site.css
@@ -77,7 +77,7 @@ section {
   position: relative;
 }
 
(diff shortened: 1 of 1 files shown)


Zopsss:
github, generate site

Issue #16225, opened 2025-01-22
FileTabCharacter: update  doc to mention  that only first violation is reported for default configuration


ElinaZoldnere:
I encountered this issue while working on enabling example tests for SuppressWithPlainTextCommentFilter.

The issue: `FileTabCharacter` check with default configuration (reporting only the first violation) does not report a file if that first violation is suppressed, even when there are more actual violations, and the user probably would expect checking the rest of the file.
https://checkstyle.sourceforge.io/checks/whitespace/filetabcharacter.html

```
$ cat ./Test.sql
-- stop check
-- suppressed violation below
	SELECT * FROM users
-- resume check
-- expected violation below
	SELECT 1
```

```
$ cat config-with-suppression.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="FileTabCharacter"/>

    <module name="SuppressWithPlainTextCommentFilter">
        <property name="offCommentFormat" value="stop check"/>
        <property name="onCommentFormat" value="resume check"/>
    </module>

</module>
```

```
$ java -jar checkstyle-10.21.1-all.jar -c config-with-suppression.xml Test.sql
Starting audit...
Audit done.
```

If `FileTabCharacter` check  is configured to report each line, it reports a violation on line 6.
```
$ cat config-with-suppression-eachline.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="FileTabCharacter">
        <property name="eachLine" value="true"/>
    </module>

    <module name="SuppressWithPlainTextCommentFilter">
        <property name="offCommentFormat" value="stop check"/>
        <property name="onCommentFormat" value="resume check"/>
    </module>

</module>
```

```
$ java -jar checkstyle-10.21.1-all.jar -c config-with-suppression-eachline.xml Test.sql
Starting audit...
[ERROR] /home/elina/IdeaProjects/sandbox/jdk17/FileTabCharacter-issue/Test.sql:6:1: Line contains a tab character. [FileTabCharacter]
Audit done.
Checkstyle ends with 1 errors.
```

[...]

PR #16227, opened 2025-01-23
Issue #16225: Update FileTabCharacter documentation with default config behaviour in case of suppressions

ElinaZoldnere:
Issue #16225 

This PR adds an explicit Note about checks default behaviour:

 When the `FileTabCharacter` check is used with the default configuration,
 only the first instance of a tab character is reported.

### DIFF
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/whitespace/FileTabCharacterCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/whitespace/FileTabCharacterCheck.xml
@@ -21,7 +21,12 @@
  when the commit messages get sent to a mailing list, they are almost impossible to
  read if you use tabs.
  &lt;/li&gt;
- &lt;/ul&gt;</description>
+ &lt;/ul&gt;
+
+ &lt;p&gt;
+ When the {@code FileTabCharacter} check is used with the default configuration,
+ only the first instance of a tab character is reported.
+ &lt;/p&gt;</description>
          <properties>
             <property default-value="false" name="eachLine" type="boolean">
                <description>Control whether to report on each line containing a tab,
--- a/src/site/xdoc/checks/whitespace/filetabcharacter.xml
+++ b/src/site/xdoc/checks/whitespace/filetabcharacter.xml
@@ -31,6 +31,13 @@
         </ul>
       </subsection>
 
+      <subsection name="Notes" id="Notes">
+        <p>
+          When the <code>FileTabCharacter</code> check is used with the default configuration,
+          only the first instance of a tab character is reported.
+        </p>
+      </subsection>
+
       <subsection name="Properties" id="Properties">
         <div class="wrapper">
           <table>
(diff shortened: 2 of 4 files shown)


ElinaZoldnere:
GitHub, generate website

ElinaZoldnere:
Please see the proposal of the wording.
As I understand, the same paragraph should be included also in class-level JavaDocs? (I have a failing test at this point.)

ElinaZoldnere:
@romani Please suggest, how to move forward with this.
Seems this has ended up in a situation that wasn't my intention.

romani:
This is not good wording. 
I still try to understand how to make it better. I think we share same confusion point, i need to meditate on this for a bit.

Please continue contribution on other issues. 
This PR will not be forgotten.

src/site/xdoc/checks/whitespace/filetabcharacter.xml
```diff
@@ -59,6 +59,15 @@
         </div>
       </subsection>
 
+      <subsection name="Notes" id="Notes">
+      <p>
+        When using the <code>FileTabCharacter</code> check with the default configuration (i.e.,
+        only the first instance of a tab character is reported), if the first instance falls under
```
romani, line 65: I don't think we need `if the first instance falls under....`
As it reference to other module, and we usually treat them completely independent, not referencing to each other.

But wording before this phrase, I I am good to move under `Rationale:`.

So point of confusion is fact that by default Check reports only first violation, not all as majority of Checks doing.

Do you agree?

romani (changes requested):
Item:

src/site/xdoc/checks/whitespace/filetabcharacter.xml
ElinaZoldnere: Thanks, I got it that referencing a specific interaction with another module would indeed be out of place here.

I agree to the suggestion and think it would still be helpful. While the same can be inferred from `eachLine` property description, an explicit statement under `Rationale` would draw immediate attention to this specific aspect of the check.

ElinaZoldnere: Done.

ElinaZoldnere:
GitHub, generate website

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/whitespace/FileTabCharacterCheck.xml
```diff
@@ -21,6 +21,10 @@
  when the commit messages get sent to a mailing list, they are almost impossible to
  read if you use tabs.
  &lt;/li&gt;
+ &lt;li&gt;
+ When the {@code FileTabCharacter} check is used with the default configuration,
+ only the first instance of a tab character is reported.
+ &lt;/li&gt;
  &lt;/ul&gt;</description>
          <properties>
```
romani, line 29: You placed it under Rationale. It is right place, please make " Note" paragraph.

romani (changes requested):
Items

src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/whitespace/FileTabCharacterCheck.xml
ElinaZoldnere: Do you mean it like this?
Screenshot 2025-01-27 at 07-36-05 FileTabCharacter – checkstyle

romani: Yes

ElinaZoldnere:
GitHub, generate website

romani (approved):
I am good with this update.
Thanks a lot.

Issue #16243, opened 2025-01-26
Inprove website rendering


romani:
split from #16077

old website: https://checkstyle.sourceforge.io/version/10.21.1/

updates:
- [x] left panel links/text is not wrapped https://github.com/checkstyle/checkstyle/issues/16077#issuecomment-2602735431
- [x] header images is bigger than before and make redline is not touching top of page, [details](https://github.com/checkstyle/checkstyle/issues/16077#issuecomment-2607293605).
- [x] tables headers text should aligned to top of cell, [details](https://github.com/checkstyle/checkstyle/issues/16077#issuecomment-2607327177). 
- [x] make line spacing for code snippets (line height) to be compact, as in was in old website 
- [x] code snippets have too much in top/bottom spacing [details](https://github.com/checkstyle/checkstyle/issues/16077#issuecomment-2614139913)
- [x] When sidebar becomes too large, it gets attached to the main content of the page: [details](https://github.com/checkstyle/checkstyle/issues/16243#issuecomment-2617808740)
- [x] remove spacing between top image and gray bar(with latest version and date) [details](https://github.com/checkstyle/checkstyle/issues/16243#issuecomment-2629086152)
- [x] Check description has different width , close to left panel and wihtout it. [details](https://github.com/checkstyle/checkstyle/issues/16243#issuecomment-2646333376)
- [x] reduce height of gap between left panel and top gray header [details](https://github.com/checkstyle/checkstyle/issues/16243#issuecomment-2650901329)
- [x] indention for code locks [details](https://github.com/checkstyle/checkstyle/pull/16301#issuecomment-2650722041)
- [x] panel that shows version and last datetime of release to be compact [details](https://github.com/checkstyle/checkstyle/issues/16243#issuecomment-2676918065).
- [x] only in mobile mode, code block should be in 12px, in desktop mode, it should be same as other text. Requested at https://github.com/checkstyle/checkstyle/pull/16439#issuecomment-2707916386
- [x] Non-formatted text are not wrapped. [Details](https://github.com/checkstyle/checkstyle/issues/16243#issuecomment-2788105183). Here is example https://checkstyle.org/checks/annotation/suppresswarningsholder.html#Example2-config

SheikhZaeem:
I am on it.

Zopsss:
There will be seperate PR for each task, please specify on which part you're working @SheikhZaeem

Zopsss:
Fix is provided for following:

[...]

PR #16239, opened 2025-01-26
Issue #16077: fixed large margins and paddings

Zopsss:
#16077 

fixed followings:

> not clear why some text under section has different indentation from left

and 

> in mobile mode reduce left and right margin, to increase space for text as much as possible.

---

Both of these issues were related to padding and margin so fixed them both in one PR.

PR #16251, opened 2025-01-28
Issue #16243: Fixed margin-left of #bodyColumn

Zopsss:
#16243 

Fixed: https://github.com/checkstyle/checkstyle/issues/16243#issuecomment-2617808740

### DIFF
--- a/src/site/resources/css/site.css
+++ b/src/site/resources/css/site.css
@@ -161,10 +161,6 @@ code {
   margin: 3px 0;
 }
 
-#bodyColumn {
-  margin-left: 286px;
-}
-
 #leftColumn {
   min-width: 250px;
   width: fit-content;
--- a/src/site/resources/js/checkstyle.js
+++ b/src/site/resources/js/checkstyle.js
@@ -80,6 +80,9 @@ function setBodyColumnMargin() {
     if (window.innerWidth < 823 && !document.querySelector("#hamburger")) {
         setCollapsableMenuButton();
     }
+
+    const leftColumnWidth = leftColumn.offsetWidth;
+    bodyColumn.style.marginLeft = `${leftColumnWidth + 15}px`;
 }
 
 function setCollapsableMenuButton() {


src/site/resources/css/site.css
```diff
@@ -161,10 +161,6 @@ code {
   margin: 3px 0;
 }
 
-#bodyColumn {
-  margin-left: 286px;
-}
```
Zopsss, line 166: we dont need this anymore as we're calculating margin-left in JS

Zopsss:
github, generate site

Zopsss:
failing Ci not related to PR https://app.circleci.com/pipelines/github/checkstyle/checkstyle/29907/workflows/e681ec87-9363-4301-bf2a-f200345c7c3a/jobs/772731

src/site/resources/js/checkstyle.js
```diff
@@ -80,6 +80,9 @@ function setBodyColumnMargin() {
     if (window.innerWidth < 823 && !document.querySelector("#hamburger")) {
         setCollapsableMenuButton();
     }
+
+    const leftColumnWidth = leftColumn.offsetWidth;
+    bodyColumn.style.marginLeft = `${leftColumnWidth + 15}px`;
```
romani, line 85: Fixing style layout problems by js is not ok.

Please share details on why it is not possible to be css only fix.

romani (changes requested):
Items

src/site/resources/js/checkstyle.js
Zopsss: If you visit some page where sidebar is small, you won't see the problem, for example: https://checkstyle.org/index.html

But in some pages the sidebar becomes very large, for example: https://checkstyle.org/checks/naming/abbreviationaswordinname.html#AbbreviationAsWordInName

Here you can see that there is no space between sidebar and main content. It is happening because I removed following JS in one of my previous PR as I thought we don't need it but I was wrong: https://github.com/checkstyle/checkstyle/pull/16239/files#diff-74191770f323d7fd9586fae43dcc6310dec524a85cc7fcfaa03e69a2cf089eeaL90-L91

We cannot fix this problem only by using CSS because sidebar's width changes dynamically based on which page we're visiting. In the above example, the `AbbreviationAsWordInName` name is too long so sidebar's width also becomes too large. We cannot dynamically define sidebar's width with CSS only, for that we require JS. 

Currently, we have defined static width of the sidebar in `site.css` but I have removed it in this PR as mentioned here: https://github.com/checkstyle/checkstyle/pull/16251/files#r1931496896

I have added back the JS I removed in one of my previous PR which calculates the width of the sidebar dynamically and adds spacing between sidebar and main content of the page.

romani: dynamic width is only for mobile mode, as far as I understand.
long text can wrap in width of area it has, having two lines for long name is in menu is ok.

is it possible ?

Zopsss:
Github, generate site

src/site/resources/js/checkstyle.js
Zopsss: > long text can wrap in width of area it has, having two lines for long name is in menu is ok.

wrapping text in menu will confuse users, it will become hard for them to read the Check name. IMO we should not wrap text

Zopsss: > dynamic width is only for mobile mode, as far as I understand.


the dynamic width of sidebar is not our concern. The problem is placement of main content of the page, it statically gets it margin-left from site.css. 

Screenshot 2025-02-03 215332

This was SS of current master branch site, the blue part the of main content and orange part is margin-left of main content of body, you can see the blue part is overlapping sidebar. So if you try to click something on that overlapped area of sidebar, it won't work. 

Why? because we gave static margin-left to our main content in site.css, and also there seem to be some placement related issue, Im not able to figure out that placement issue. But it's not good to give static margin from our site.css, instead we should rely on JS for that.

romani:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/d19ae2b_20250202155204/checks/coding/index.html

Screenshot_20250203-182118


I like this wrapping of text, can we do same in non mobile mode?

Zopsss:
it is possible but it'll make it hard to read long check names so I think we should stick with dynamic width

Screenshot 2025-02-10 144411

mahfouz72:
How was it in the old site?

Zopsss:
it wasn't wrapping: 

image

mahfouz72:
> do you like wrap of long names or how it is now - overlap ?


@romani I don't like either of them, Overlapping is bad for UI, and wrapping long names makes them hard to read and inconvenient.
I prefer the approach in this PR. The left column should be wide enough to accommodate the longest name without wrapping, and the margin should be adjusted accordingly. I think this was the case in the old site. We can take others opinion also

romani:
My concern here is usage of JS for layout, it smells very badly, usually red flag.
Try to keep this PR a bit longer to collect ideas on how to resolve this by css only.

Zopsss:
> My concern here is usage of JS for layout, it smells very badly, usually red flag.

JS is not that heavy, it's only of two lines. Currently it is inside `setBodyColumnMargin` function which is invoked on load and resize events. If you want to optimize we can load this JS only once, when the page loads. 

This is a very small JS code so it'll not affect user's experience even if we keep it as it is now.

romani:
>This is a very small xxxx code

All mess in engineering is starting from this phrase :) .

romani:
GitHub, generate website

src/site/resources/js/checkstyle.js
romani: ok, this explains why we should remove strict width `#bodyColumn {`.
So it(left panel) should take now all horizontal space that it needs, so there should be some way to configure it by CSS.

`function setBodyColumnMargin() {` should exists only for mobile mode of website, not for big screens.

@SheikhZaeem , may be you know a CSS trick to make it possible?

SheikhZaeem: let me check

Zopsss: > ok, this explains why we should remove strict width

Yes, this will also require to revert the code which made the sidebar to habe static width. 

In normal cases this is possible by only using CSS but we're using maven-site-plugin so site has some predefined structure so making the sidebar to have dynamic width with only using CSS might be very tricky or even impossible... Not sure about this

romani: Dynamic width is only in mobile mode, and there we have wrapping of long names, so no problems with dynamic width.

We have issue for case for desktop rendering , when all is well defined. We can disable hamburger, and it works same, with overlap.

romani:
https://checkstyle.sourceforge.io/checks/metrics/booleanexpressioncomplexity.html#BooleanExpressionComplexity 

I do not see problem now .
@Zopsss , please clarify what is target of this PR

Zopsss:
> I do not see problem now .
@Zopsss , please clarify what is target of this PR

This PR is too old, maybe some other contributor solved it. Now there is proper spacing between sidebar and main content.

Zopsss:
But I still think that the text wrapping for sidebar is not a good option:

image

It makes text unreadable. For smaller devices, we can wrap the text, it is okay and also rarely anybody would be using our site on mobile devices so it is not the main concern.

But for desktop and laptop screen, our sidebar text should not be wrappable. If you think the sidebar is taking too much space if we don't wrap it's text then we can make a burger for it like we have for mobile devices:

image

So user can hide/unhide the sidebar at their own will.

romani:
@Zopsss , do you think that this PR resolving this issue and making it to work as you described above?

Zopsss:
Nope. The problem this PR was targeting is already resolved. I just gave my opinion of how the sidebar should be for desktop devices.

romani:
@Zopsss , if you want to continue, please rebase , generate website and let's look again when this update is beneficial.

If no energy to continue, let's close PR. One day in future it will pop up again.

Zopsss:
This PR was focused on completely different issue. I will open a new issue for the idea I proposed above and send a PR for it.

romani:
@Zopsss , should we close this PR ?

Zopsss:
yeah.

PR #16252, opened 2025-01-28
Issue #16165: fix CyclomaticComplexity's switchBlockAsSingleDecisionPoint to properly handle switch blocks

### LINKED ISSUE #16165: Regression: CyclomaticComplexity.switchBlockAsSingleDecisionPoint does not work properly anymore (10.21.1)
I have read check documentation: https://checkstyle.org/checks/metrics/cyclomaticcomplexity.html

```bash
/var/tmp $ javac Test.java
Successfully compiled with JDK 21

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="CyclomaticComplexity">
            <property name="max" value="3"/>
            <property name="switchBlockAsSingleDecisionPoint" value="true"/>
        </module>
    </module>
</module>

/var/tmp $ cat Test.java
public class Test {
    // Passed: Cyclomatic Complexity is 2
    void test(Object obj) { // 1, method declaration
        switch (obj) { // 2, switch
            case Integer i -> System.out.println("Integer: " + i);
            case String s -> System.out.println("String " + s);
        }
    }

    // Violation: Cyclomatic Complexity is 4, but must be 2 since
    // switchBlockAsSingleDecisionPoint = true and "when" literals must not be counted
    void test2(Object obj1, Object obj2) { // 1, method declaration
        switch (obj1) { // 2, switch
            case Integer i1 when obj2 instanceof Integer i2 -> System.out.println("Integers: %d, %d".formatted(i1, i2));
            case String s1 when obj2 instanceof String s2 -> System.out.println("Strings: %s, %s".formatted(s1, s2));
        }
    }
}

/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.21.1-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:13:5: Cyclomatic Complexity is 4 (max allowed is 3). [CyclomaticComplexity]
Audit done.
```
---

switchBlockAsSingleDecisionPoint is set to true in config.xml, so switch block should be treated as a single decision point.
However, in case of "test2" method "when" literals are counted, when must not.
This is regression after fix of #15341

---

mohitsatr:
resolves #16165

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/CyclomaticComplexityCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/CyclomaticComplexityCheck.java
@@ -27,6 +27,7 @@ import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.ScopeUtil;
 
 /**
  * <div>
@@ -270,7 +271,7 @@ public class CyclomaticComplexityCheck
      */
     private void visitTokenHook(DetailAST ast) {
         if (switchBlockAsSingleDecisionPoint) {
-            if (ast.getType() != TokenTypes.LITERAL_CASE) {
+            if (!ScopeUtil.isInBlockOf(ast, TokenTypes.LITERAL_SWITCH)) {
                 incrementCurrentValue(BigInteger.ONE);
             }
         }
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java
@@ -199,7 +199,7 @@ public final class ScopeUtil {
      * @param tokenType type of token.
      * @return a {@code boolean} value
      */
-    private static boolean isInBlockOf(DetailAST node, int tokenType) {
+    public static boolean isInBlockOf(DetailAST node, int tokenType) {
         boolean returnValue = false;
 
         // Loop up looking for a containing interface block
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/metrics/CyclomaticComplexityCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/metrics/CyclomaticComplexityCheckTest.java
@@ -182,13 +182,25 @@ public class CyclomaticComplexityCheckTest
     public void testWhenExpressionSwitchAsSinglePoint() throws Exception {
         final String[] expected = {
             "14:5: " + getCheckMessage(MSG_KEY, 5, 0),
-            "20:5: " + getCheckMessage(MSG_KEY, 4, 0),
-            "29:5: " + getCheckMessage(MSG_KEY, 5, 0),
-            "39:5: " + getCheckMessage(MSG_KEY, 5, 0),
+            "20:5: " + getCheckMessage(MSG_KEY, 2, 0),
+            "29:5: " + getCheckMessage(MSG_KEY, 2, 0),
+            "39:5: " + getCheckMessage(MSG_KEY, 2, 0),
         };
         verifyWithInlineConfigParser(
                 getNonCompilablePath(
                         "InputCyclomaticComplexityWhenSwitchAsSinglePoint.java"), expected);
+    }
+
+    @Test
+    public void testSwitchBlockAsSingleDecisionPointWithNestedSwitch() throws Exception {
+        final String[] expected = {
+            "17:5: " + getCheckMessage(MSG_KEY, 2, 0),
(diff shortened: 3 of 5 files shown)


mohitsatr:
Github, generate report for CyclomaticComplexity/all-examples-in-one

mohitsatr:
@nrmancuso @mahfouz72 please take a look at  [case](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/f65de102b0555c70c28d26cb39a7c109094ceb01_2025085332/reports/diff/apache-ant/xref/home/runner/work/checkstyle/checkstyle/.ci-temp/repositories/apache-ant/src/main/org/apache/tools/ant/filters/FixCrLfFilter.java.html#L519)
how should nested switch statements be treated ?

nrmancuso:
> how should nested switch statements be treated

Please propose a couple of solutions with examples, and provide your recommendation.

mohitsatr:
Github, generate report for CyclomaticComplexity/all-examples-in-one

mohitsatr:
@nrmancuso Well, according to me there are only two possible solutions: 
1. Treat the outer most switch block as single decision point and ignore everything inside it, including any nested switches.
2. Treat the outer most switch block as single decision point but still count any nested switch while ignoring everything else in the block.

i went with second approach (for no particular reason).

Second Approach: 
if there's a nested switch block, complexity increases for the switch keyword but anything inside that 2nd level switch block is ignored. in other words, complexity is just no. of `switch` keywords in the function.
See the example below : 
```
// Cyclomatic Complexity is 4 
void test3(Object obj1, Object obj2, Object obj3) {
        int j,k,l,m;
        switch (obj1) {                //1
            case Integer i1:
            case Integer i2:
            case Integer i3:
                break;
            case Integer i4:
                switch (obj2) {        //2
                    case Integer i1:
                       j = 1;
                       break;
                    default:
                        break;
                }
                break;
            case Integer i5:
                if (j == 1) {               //ignored
                    k = 1;
                }
                else {
                    l == 1;
                }
                break;
            case Integer i6:
                switch (obj3) {        //3
                    case Integer i6:
                        break;
                    default:
                        break;
                }
                break;
            default:
                break;
        }
    }
```

but that raises the question: why not the first approach ? why are we even worrying about switch block inside a switch block. we can simply ignore everything inside the switch block. After all `switchBlockAsSingleDecisionPoint` is not the default behaviour and if a user enables it, then they must expect Check to ignore nested switches too, right ?

nrmancuso:
I vote for (1), @mahfouz72 @romani @rnveach what do you think?

romani:
@mohitsatr , please proceed with `(1)`. This is the only possible way, `switchBlockAsSingleDecisionPoint` is driving behavior, no need to look inside `switch` at all.

mohitsatr:
Github, generate report for CyclomaticComplexity/all-examples-in-one

romani:
@mohitsatr , please check jdk21 compilation

mohitsatr:
@romani have you gone through the regression report ? don't want to waste time if it's already reviewed by you.

romani:
@mohitsatr , I am waiting explicit comment that PR is ready for review as it was in WIP for long, so please do all verifications on your side first.

mohitsatr:
@romani this one is ready for review. on my part I checked all the files in report and Im pretty confident that change is working . I have not gone all in Openjdk. it is tiring.

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/CyclomaticComplexityCheck.java
```diff
@@ -156,6 +156,9 @@ public class CyclomaticComplexityCheck
     /** Specify the maximum threshold allowed. */
     private int max = DEFAULT_COMPLEXITY_VALUE;
 
+    /** Tracks the no. of {@code switch} blocks. */
+    private int switchCount;
```
romani, line 160: It took me some time to make this decision.:
lets not have extra class field that keep state of class.

We can avoid this by tiny performance degradation and checking switch depth each time we hit it.
Check is method/ctor targeting for there will be no much checking of parent AST nodes. it is performance hit, yes, but it will help us to make this Check stateless in future.

In future we will try to make implementation that promotes Stateless checks, to reuse Mutlithreading.
This Check is stateful only by `currentValue` , we will try to resolve this in future to avoid this state.
So lets not grow statefullness of Checks.

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/CyclomaticComplexityCheck.java
mohitsatr: can you tell of any similar function that works like this, so I don’t have to build it from scratch ?

nrmancuso: >lets not have extra class field that keep state of class.

I agree, we have no need to optimize this check prematurely

mohitsatr:
Github, generate report for CyclomaticComplexity/all-examples-in-one

romani (approved):
thanks a lot.

nrmancuso (approved):
Good stuff

PR #16256, opened 2025-01-28
Fix header image size and red line alignment 

SheikhZaeem:
This pull request solves the issue #16243, where the header image was larger than intended and also the red line was not touching the top of the page. The changes include:

Fixing the size of header image.

Adjusting the CSS to ensure the red line aligns properly with the top of the page.

Thank you.

Issue Link: 
Fixes: #16243

### DIFF
--- a/src/site/resources/css/site.css
+++ b/src/site/resources/css/site.css
@@ -47,6 +47,14 @@ section p {
 
 }
 
+#bannerRight h1 {
+  margin: 0;
+}
+
+.container-fluid-top {
+  padding-top: 0 !important;
+}
+
 #banner {
   border-bottom: 1px solid #fff;
   background: url("../images/header-background.png") repeat-x scroll 0 0 transparent;
--- a/src/site/resources/js/anchors.js
+++ b/src/site/resources/js/anchors.js
@@ -16,6 +16,11 @@
 
         var anchors = document.querySelectorAll("h1, h2");
         [].forEach.call(anchors, function (anchorItem) {
+            
+            if (anchorItem.closest("#bannerRight") ) {
+                return;
+            }
+
             var name = anchorItem.childNodes[0].textContent.replaceAll(" ", "_");
             var link = "" + url + "#" + name + "";
 
--- a/src/site/site.xml
+++ b/src/site/site.xml
@@ -13,7 +13,7 @@
   <bannerLeft href="https://checkstyle.org">
     <image src="images/header-checkstyle-logo.png" alt="Checkstyle"/>
   </bannerLeft>
-  <bannerRight href="https://checkstyle.org">
+  <bannerRight>
     <image src="images/header-right-ruller.png" alt="Checkstyle"/>
   </bannerRight>
 


Zopsss:
github, generate site

src/site/resources/css/site.css
```diff
@@ -47,9 +47,14 @@ section p {
...
 }
 
+#bannerRight img {
+  transform: translateY(-12px);
+}
+
 #banner {
   border-bottom: 1px solid #fff;
   background: url("../images/header-background.png") repeat-x scroll 0 0 transparent;
+  transform: translateY(-6px);
```
Zopsss, line 57: you're transforming the banner to upwards, it is causing banner to have more empty space below it.

Current banner:
Screenshot 2025-01-29 104558

Banner after your changes:
Screenshot 2025-01-29 104746

We want the red line to stick to the top, the problem is that it is wrapped inside H1 tag which have margin-top & bottom.

Screenshot 2025-01-29 104401


Removing margin-top will fix the problem

Zopsss (changes requested):

Zopsss:
@romani why do we have this red line? and why it is a link? If the red line is only for aesthetics then it doesn't have to be a link

romani:
>why do we have this red line?

This is us how kids use to study to write.
https://www.istockphoto.com/vector/vector-opened-school-cursive-worksheet-copybook-gm872316754-243658512

>why it is a link?

I surprised, it supposed to be fancy decor.

SheikhZaeem:
@Zopsss Thank you very much for the feedback.

Zopsss:
> I surprised, it supposed to be fancy decor. Link can be removed.

@SheikhZaeem it would be very helpful if you can remove the `<a>` from the red line. It will require some JS. If you cannot do it, that's okay but if you want to do it, then you can ask me for help.

SheikhZaeem:
@Zopsss Yes I want to do it. Thank you

SheikhZaeem:
> > I surprised, it supposed to be fancy decor. Link can be removed.
> 
> @SheikhZaeem it would be very helpful if you can remove the `<a>` from the red line. It will require some JS. If you cannot do it, that's okay but if you want to do it, then you can ask me for help.

@Zopsss just to confirm, when removing the <a, should I make sure it s never added in the first place (by changing the js to prevent it), or should I remove it if it already exists in the DOM?

SheikhZaeem:
@Zopsss Hello again. I am facing some problem. I removed  "var a = document.createElement("a"), a.setAttribute("href", link) andvanchor.appendChild(a)" from anchors.js folder as I believe the are responsible for generating <a> for the red line image section. However, no changes are seen when I inspect the updated html file in the target folder. Could it be that other js file is adding <a> tag around the image or am I missing something? Thank you in advance.

Zopsss:
> @Zopsss just to confirm, when removing the <a, should I make sure it s never added in the first place (by changing the js to prevent it), or should I remove it if it already exists in the DOM?

make sure it is never added in first place

Zopsss:
> @Zopsss Hello again. I am facing some problem. I removed "var a = document.createElement("a"), a.setAttribute("href", link) andvanchor.appendChild(a)" from anchors.js folder as I believe the are responsible for generating for the red line image section. However, no changes are seen when I inspect the updated html file in the target folder. Could it be that other js file is adding tag around the image or am I missing something? Thank you in advance.

are you referring to these 2 lines?

https://github.com/checkstyle/checkstyle/blob/977c425e633414fec09952bb3bb887a937ce8157/src/site/resources/js/anchors.js#L22-L23

you don't need to remove `<a>` from all H1 tags, remove it only from red line. These two lines are used to add `<a>` to all H1 tags. 

In the `forEach` loop check if the current H1 tag is of red line, if yes then do not create `<a>` tag for it. 

**How to check if the current H1 tag is of Red Line?**

It is child of `#bannerRight`, use JS to identify it.

Screenshot 2025-01-29 225700

SheikhZaeem:
@Zopsss Since <a> is already present in target/site/anttask.html, which i am assuming means Maven itself is inserting <a> inside #bannerRight before JavaScript runs. So <a> is coming from
Screenshot 2025-01-30 at 11 53 36 AM
Since the href attribute is present, Maven automatically wraps the image in an <a> tag.

SheikhZaeem:
@Zopsss Also now when red line is pressed, it doesn't do anything. I think makes more sense.

Zopsss:
Github, generate site

Zopsss:
Hmm now the line isnt a link anymore but when hovered over, the anchor image appears next to it

Screenshot 2025-01-30 145225


I think you can fix this by adding some JS here:

https://github.com/checkstyle/checkstyle/blob/977c425e633414fec09952bb3bb887a937ce8157/src/site/resources/js/anchors.js#L17-L20

Add some logic like if the h1 tag is child of #bannerRight then skip that element.

src/site/resources/css/site.css
```diff
@@ -5,6 +5,7 @@ body, td, select, input, li {
 h1 {
   text-transform: capitalize;
   font-size: x-large;
+  margin: 0px;
```
Zopsss, line 8: revert this, we want `margin-top: 0` only for Red Line's H1 tag. Following CSS should remove margin-top and extra padding-top 

```css

#bannerRight h1 {
    margin: 0;
}

.container-fluid-top {
    padding-top: 0 !important;
}
```

src/site/resources/css/site.css
```diff
@@ -47,9 +48,14 @@ section p {
...
 }
 
+#bannerRight img {
+  transform: translateY(-12px);
+}
+
 #banner {
   border-bottom: 1px solid #fff;
   background: url("../images/header-background.png") repeat-x scroll 0 0 transparent;
+  transform: translateY(-6px);
```
Zopsss, line 58: revert this, use above mentioned code.

Zopsss (changes requested):
fixing extra margins and paddings

src/site/resources/css/site.css
Zopsss: add this code before following line:

https://github.com/checkstyle/checkstyle/blob/977c425e633414fec09952bb3bb887a937ce8157/src/site/resources/css/site.css#L50

to keep same section's CSS together

Zopsss:
Fix failing CI

https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=25775&view=logs&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111&l=1351

it is asking you to follow our commit message rules.

SheikhZaeem:
@Zopsss Fixed and reverted the code as you said. But I don't know why I get these errors regarding commit messages.

Zopsss:
Github, generate site

Zopsss:
> @Zopsss Fixed and reverted the code as you said. But I don't know why I get these errors regarding commit messages.

Read the commit message carefully, it has mentioned some rules regarding commit message. You seem to have break 1st rule. The commit message should contain this prefix: `Issue #YourIssueNumber` 

Normally you should run `mvn clean verify` before pushing changes, it'll give you such errors so you can fix them early. But this was website issue so it's okay to not run this command.

Zopsss:
Please squash all commits. Always keep 1 commit per PR.

SheikhZaeem:
> Please squash all commits. Always keep 1 commit per PR.

Can you explain in more detail. Do you mean if i keep changing the code of my pull request again and again, I don't need to write new commit messages?

SheikhZaeem:
github, generate site

SheikhZaeem:
github, generate site

SheikhZaeem:
@Zopsss I squashed and then I lost some of the changes i did due to my mistake. But I think now it is fine.

Zopsss:
Github, generate site

SheikhZaeem:
Hey @Zopsss. Thanks for so much help. Is there anything else left for this PR? If not, then I will move to other issues?

Zopsss (approved):
Red line isn't a link anymore and it is sticked to the top now. Thanks a lot for all the hard work!

Zopsss:
@romani the Checkstyle logo in banner is a link. But I guess it is fine, or do you want to change it?

romani (approved):
@SheikhZaeem , @Zopsss , thanks a lot for collaboration to make a fix !!!

great job !!

Issue #15534, opened 2024-08-20
Add support of filtered violation messages to BDD Inputs


romani:
detected at https://github.com/checkstyle/checkstyle/pull/15452/files#r1705754697

We need to add support of verification of what violation was filtered.

consider https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/parenpad/InputParenPadTryWithResourcesAndSuppression.java#L17

ideally we should have violation message at line `try ( AutoCloseable a = null) {} // filtered violation`
to be something like `try ( AutoCloseable a = null) {} // filtered violation 'is followed by whitespace'` 

 we actually need to updatehttps://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
To recognize  this pattern of comment and confirm that violation like this exists there without filters

Harsh-Agrawal96:
As of now, I can understand In BDD Inputs we need to add the messages for `filtered violation` 
So this is all or anything more

romani:
just this  - `filtered violation`

may be we can do `filtered violation 'violation message as it was'` but lets see in PR, if complicated, we can do in next issue.

Harsh-Agrawal96:
I am on it

RohanMannan89:
>ideally we should have violation message at line `try ( AutoCloseable a = null) {} // filtered violation`
>to be something like `try ( AutoCloseable a = null) {} // filtered violation 'is followed by whitespace'`
Means, In this issue we just need to change the `filtered violation` to `filtered violation 'violation message as it was'` ??

romani:
>violation message as it was

Is actual violation that is placed without filter/suppression.

romani:
we actually need to updatehttps://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
To recognize  this pattern of comment and confirm that violation like this exists there without filters .

So extra PR is expected

romani:
done by @Harsh-Agrawal96  !!!
thanks a lot !!!


extra effort was done at https://github.com/checkstyle/checkstyle/pull/16286 , but in comment I explained why we should not do this.

PR #16264, opened 2025-02-01
Issue #15534: Support of filtered violation messages

Harsh-Agrawal96:
Issue #15534

add support of verification of what violation was filtered.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java
@@ -289,9 +289,9 @@ public class SuppressionXpathSingleFilterTest
     public void testDecideByMessage() throws Exception {
         final String[] expected = {
             "28:1: " + getCheckMessage(MissingJavadocTypeCheck.class, MSG_JAVADOC_MISSING),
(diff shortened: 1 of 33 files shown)


Harsh-Agrawal96:
I specified the violation messages for `// filtered violation` 
please review it, if it well then I will continue with remaining other `// filtered violation below` for specifying messages.

romani:
We lost test coverage after this update 
https://github.com/checkstyle/checkstyle/actions/runs/13086479411/job/36518207884?pr=16264

To reproduce please run on local `./.ci/pitest.sh "pitest-filters"` . We should find what causes this regression and restore coverage.

romani:
You try To change all at ones, it might be challenging as  you might already see. Let's do update in smaller parts. So it will be easier to catch what update causing pitest failure.

Harsh-Agrawal96:
yes, for now, I am not familiar with that pitest, need to learn about pitest

romani:
https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports

romani (approved):
Ok to merge as pre-fix.

But we actually need to updatehttps://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
To recognize  this pattern of comment and confirm that violation like this exists there without filters

PR #16273, opened 2025-02-03
Issue #11163: Enforce file size on Java inputs for FinalClass

zyadahmed:
Issue #11163  Enforce file size on Java inputs for InputFinalClass


romani (approved):
Thanks a lot

PR #12981, opened 2023-04-13
Issue #12803: allow throwing CheckStyleException

Boukhalfa11:
Issue #12803

PR #16277, opened 2025-02-04
Issue #12803: enables interface methods to throw CheckstyleException

### LINKED ISSUE #12803: All interface methods should allow throwing CheckstyleException

mohitsatr:
resolves #12803 finishes #12981 

WIP:
- `Checker` is excluded from Coverage Testing for now.
- to meet the coverage, reflection is used for `TranslationCheck`. since I'm not much familiar with this type of testing, I need some feedback.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -959,6 +959,7 @@
                     <exclude>com.puppycrawl.tools.checkstyle.meta.XmlMeta*</exclude>
                     <!-- OS specific utility providing class -->
                     <exclude>com.puppycrawl.tools.checkstyle.utils.OsSpecificUtil</exclude>
+                    <exclude>com.puppycrawl.tools.checkstyle.Checker</exclude>
                   </excludes>
                   <limits>
                     <limit>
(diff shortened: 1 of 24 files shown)


romani:
sorry for not thinking on issue deeply,  I closed issue.

PR #16279, opened 2025-02-05
Issue #13999: Kill mutation for AbstractJavadocCheck

mohitsatr:
part of #13999

Diff Regression config: https://gist.githubusercontent.com/mohitsatr/f5a524a06294d907c84b9fc5b68661dc/raw/604872961df92db35e66c26218dc74c61bb10097/config.xml

https://github.com/checkstyle/checkstyle/blob/e1dc3abaa59f522ffafb200aa7ae4c2ced7a64c0/config/pitest-suppressions/pitest-javadoc-suppressions.xml#L3-L10

https://github.com/checkstyle/checkstyle/blob/e1dc3abaa59f522ffafb200aa7ae4c2ced7a64c0/config/pitest-suppressions/pitest-javadoc-suppressions.xml#L12-L19

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheckTest.java
@@ -598,6 +598,41 @@ public class AbstractJavadocCheckTest extends AbstractModuleTestSupport {
                 getPath("InputAbstractJavadocNonTightHtmlTags3.java"), expected);
     }
 
+    @Test
+    public void testLeaveJavadocToken() throws Exception {
+        final String[] expected = CommonUtil.EMPTY_STRING_ARRAY;
+        verifyWithInlineConfigParser(
+                getPath("InputAbstractJavadocLeaveToken.java"), expected);
+    }
+
+    public static class JavadocLeaveTokenCheck extends AbstractJavadocCheck {
+
+        private static int visitCount;
+        private static int leaveCount;
+
+        @Override
+        public int[] getDefaultJavadocTokens() {
+            return new int[] {JavadocTokenTypes.HTML_ELEMENT};
+        }
+
+        @Override
+        public void visitJavadocToken(DetailNode ast) {
+            visitCount++;
+        }
+
+        @Override
+        public void leaveJavadocToken(DetailNode ast) {
+            leaveCount++;
+        }
+
+        @Override
+        public void finishJavadocTree(DetailNode ast) {
+            if (visitCount != leaveCount) {
+                throw new IllegalStateException("mismatch in visitCount and leaveCount");
+            }
+        }
+    }
+
     public static class ParseJavadocOnlyCheck extends AbstractJavadocCheck {
 
         @Override
--- a/config/pitest-suppressions/pitest-javadoc-suppressions.xml
+++ b/config/pitest-suppressions/pitest-javadoc-suppressions.xml
@@ -1,24 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
   <mutation unstable="false">
-    <sourceFile>AbstractJavadocCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheck</mutatedClass>
-    <mutatedMethod>walk</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_ELSE</mutator>
-    <description>removed conditional - replaced equality check with false</description>
-    <lineContent>if (curNode != null) {</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>AbstractJavadocCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheck</mutatedClass>
-    <mutatedMethod>walk</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck::shouldBeProcessed</description>
-    <lineContent>waitsForProcessing = shouldBeProcessed(curNode);</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>JavadocMethodCheck.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocMethodCheck$Token</mutatedClass>
     <mutatedMethod>&lt;init&gt;</mutatedMethod>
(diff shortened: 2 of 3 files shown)


mohitsatr:
Github, generate report

romani:
To catch a case, you need to activate all Javadoc checks that a child of this abstract parent. 
Or start analysis of how why this code might be required.

mohitsatr:
> you need to activate all Javadoc checks

I believe all child Checks of this class are already in the config. 
are you talking about some specific-purpose Checks like `TokenIsNotInAcceptablesCheck` which are implemented in `AbstractJavadocCheckTest.java` file ?

romani:
no, we do not care about Test modules, until we be confident that existing checks are not affected to mutation.

romani:
@mohitsatr , lets try to finish this PR. Let me know if something is not clear.

romani:
@mahfouz72, please review this PR and share your ideas.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck.java
```diff
@@ -390,9 +390,6 @@ private void walk(DetailNode root) {
 
                 toVisit = JavadocUtil.getNextSibling(curNode);
                 curNode = curNode.getParent();
-                if (curNode != null) {
-                    waitsForProcessing = shouldBeProcessed(curNode);
-                }
```
mahfouz72, line 395: It is not safe to remove this condition. `waitsForProcessing` should be updated for every iteration so we know whether to call `leaveJavadocToken` or not. when we finish processing some subtree we will be on conditions:

toVisit = null (no more siblings in this subtree)
curNode !=  null and is equal  tothe root of this subtree

In the next iteration in the inner while loop, we will hit the below code and calls `leaveJavadocToken()` on the root of this subtree

```Java
           if (waitsForProcessing) {
               leaveJavadocToken(curNode);
           }
```

Regression has no diff because `leaveJavadocToken()` is not overridden by any check till now. (It is overridden only by a [custom check](https://github.com/checkstyle/checkstyle/blob/2bb80d7f18b0d602bfaf93c3ef8f18d9a785d27a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheckTest.java#L723) used for testing)

mahfouz72 (changes requested):
I don't think there is any real check execution test case that would kill this mutation, Reason:
 > leaveJavadocToken() is not overridden by any check till now.

and pure junit is not allowed so we should close this PR

mahfouz72:
> But it is still okay to remove this condition because leaveJavaToken is empty function i.e it has no implementations.

This is not a valid reason. Even if `leaveJavadocToken` is empty now, its purpose could change in the future. If its implementation is added later, you will want to ensure that it is properly invoked when backtracking through the tree

mahfouz72:
@mohitsatr please close your PR if you agree. if not please provide more analysis or reasons.

romani:
But still better to kill mutations, if we see reasons why we need such code.

mahfouz72:
@romani 

please read https://github.com/checkstyle/checkstyle/pull/16279#pullrequestreview-2667442591

Real checks doesn't override `leaveJavadocToken()` so how we supposed to kill it with real check execution as per our rules. Am i missing something?

romani:
We can always can create new Check, just for testing.
If non of existing doing it , it doesn't mean that it is not required. It is good justification to make test Check to confirm that it happening.

mahfouz72:
Yes, it makes sense, I thought that we stopped using test modules and request to test by real checks only.

@mohitsatr please kill this mutation by creating some testing check that overrides `leaveJavadocToken()`

romani:
We are not allowing to use them without good reason. Amount of javadoc checks is small , so chance to use all possible of walking on tree is less than in java.

We just need put as comment reason of test module, it will be exception from rule.

mohitsatr:
so a dummy check just like below :
https://github.com/checkstyle/checkstyle/blob/6d08d8f501af00e1e7a11ffe2254c82ab1c75370/src/test/java/com/puppycrawl/tools/checkstyle/api/AbstractFileSetCheckTest.java#L324-L336
?

romani:
Yes

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheckTest.java
```diff
@@ -598,6 +600,73 @@ public void testVisitCountForCheckAcceptingJavadocWithNonTightHtml3() throws Exc
...
+            final List<DetailNode> textNodes = new ArrayList<>();
+            DetailNode node = JavadocUtil.getFirstChild(descriptionNode);
+            while (JavadocUtil.getNextSibling(node) != null) {
+                if (node.getType() == JavadocTokenTypes.HTML_ELEMENT) {
+                    final DetailNode descriptionNodeChild = JavadocUtil.getFirstChild(node);
+                    textNodes.addAll(getAllNewlineNodes(descriptionNodeChild));
+                }
+                if (node.getType() == JavadocTokenTypes.LEADING_ASTERISK) {
+                    textNodes.add(node);
+                }
```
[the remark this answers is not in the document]
mohitsatr: simplified

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheckTest.java
```diff
@@ -598,6 +600,73 @@ public void testVisitCountForCheckAcceptingJavadocWithNonTightHtml3() throws Exc
...
+    @Test
+    public void testleaveJavadocTokenForMutationKilling() throws Exception {
+        final String[] expected = {
+            "11: " + getCheckMessage(JavadocLeaveTokenCheck.class,
+                    JavadocLeaveTokenCheck.MSG_KEY, 4),
+            "14: " + getCheckMessage(JavadocLeaveTokenCheck.class,
+                    JavadocLeaveTokenCheck.MSG_KEY, 4),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputAbstractJavadocLeaveMutationKill.java"), expected);
```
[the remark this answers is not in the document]
mohitsatr: done

mahfouz72 (approved):
Thanks a lot!

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheckTest.java
```diff
@@ -598,6 +598,43 @@ public void testVisitCountForCheckAcceptingJavadocWithNonTightHtml3() throws Exc
...
+                    JavadocLeaveTokenCheck.MSG_KEY, 4),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputAbstractJavadocLeaveToken.java"), expected);
+    }
+
+    public static class JavadocLeaveTokenCheck extends AbstractJavadocCheck {
+
+        public static final String MSG_KEY = "tag.continuation.indent";
+        private static final int OFFSET = 4;
```
romani, line 616: Let's just count visit call and leave calls and
Override `   public void finishJavadocTree(DetailNode` and throw exception if mismatch of counts.
Better to not reuse message keys  from other Checks, input looks weird.
Exception will help to kill mutation, and it will be clear that check has only technical meaning

romani (changes requested):
Last simplification:

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheckTest.java
mohitsatr: done

romani (approved):
Elegant!

PR #16286, opened 2025-02-07
Issue #15534: add filtered violation Explanation Pattern

Harsh-Agrawal96:
Issue #15534

Add filtered violation Explanation Pattern in BDD inputs

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
@@ -121,6 +121,18 @@ public final class InlineConfigParser {
     private static final Pattern FILTERED_VIOLATION_BELOW_PATTERN = Pattern
             .compile(".*//\\s*filtered violation below\\s*(?:['\"](.*)['\"])?$");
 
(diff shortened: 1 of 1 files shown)


Harsh-Agrawal96:
@romani please verify the changes, as of now I can understand that these changes we need to do.
Let me know if anything I am doing wrong.

romani:
Direction is good, but please show example of failure when message defined in comments is mismatched with what Check is placing on a line.

And example of how it failing when violation is not actually happening.

Harsh-Agrawal96:
is any example for that type of cases for any other pattern ??

romani:
after thinking on this ......

we already have support `// filtered violation 'message'`
this PR adds support for `filtered violation some wording of explanation` that is a bit loose pattern and can hide actual, in first pattern we exact and it is good to be exact.

We might more benefit from:
```
`// filtered violations above: 
//            'message1'`
//            'message2'`
```
but still we should avoid such cases at all, as placing few violation on same line and suppressing on same line and having it as Input means over complication of testing (messy all at ones).

Lets not extend pattern.

@Harsh-Agrawal96 , thanks a lot for your PR, sorry for confusion, my bad.

Issue #16291, opened 2025-02-09
UI Overlap Issue in Checkstyle Documentation Sidebar


SheikhZaeem:
## Problem Description

In the Checkstyle documentation, some check names in the left hand sidebar are too long which causes the main content section on the right to be too close to the left, making the UI look compact and messy

## Steps to Reproduce
		Go to the Checkstyle documentation page (e.g., https://checkstyle.org/checks/javadoc/javadocmissingwhitespaceafterasterisk.html).
		Scroll through the left-hand sidebar to find long-named checks.
		Observe that the right-hand content section overlaps with the sidebar instead of adjusting properly.

## Screenshot:

<img width="1287" alt="Image" src="https://github.com/user-attachments/assets/d3ae7bef-0b03-4d6f-b158-4cdfa9cb1396" />

## Expected Behavior
		The left hand sidebar should properly accommodate long check names without increasing the size of the sidebar if possible.
		The right hand content section should be fully visible and not be pushed too close into the left sidebar.

## Possible Solutions
		We can allow long check names in the sidebar to wrap onto multiple lines.
		We can show full names in a tooltip when hovered over instead of displaying them fully.
	        The sidebar could expand slightly for long names but not affect the main content layout.
		If names are too long, introduce horizontal scrolling for the sidebar.
                If there’s a better way to handle this issue, feel free to share your ideas.

romani:
> We can allow long check names in the sidebar to wrap onto multiple lines.

I like this, it was like this in old website.
Lets try it to see how it will looks.

SheikhZaeem:
> > We can allow long check names in the sidebar to wrap onto multiple lines.
> 
> I like this, it was like this in old website. Lets try it to see how it will looks.
@romani Sure :)

Likhil022:
Hi, I'd like to work on this issue. Can I take it up?

romani:
@Likhil022 , PR is already sent, please choose another issue.

Zopsss:
> In the Checkstyle documentation, some check names in the left hand sidebar are too long which causes the main content section on the right to be too close to the left, making the UI look compact and messy

This is occurring because I removed some JS from checkstyle.js in one of my old PRs, I mentioned it here:

https://github.com/checkstyle/checkstyle/issues/16243#issuecomment-2617808740

>We can allow long check names in the sidebar to wrap onto multiple lines.
>We can show full names in a tooltip when hovered over instead of displaying them fully.

[...]

PR #16297, opened 2025-02-10
Issue #16291: Fixed UI overlap issue in documentation sidebar

SheikhZaeem:
Fixed the left column alignment issue that caused text to misalign due to improper width handling with minimal changes. It was previously using width: fit-content which led to inconsistent wrapping, especially for long words.

Before:

Screenshot 2025-02-10 at 2 29 42 PM

After:

Screenshot 2025-02-10 at 2 38 54 PM

Issue Link: 
Fixes: #16291

### DIFF
--- a/src/site/resources/css/site.css
+++ b/src/site/resources/css/site.css
@@ -179,7 +179,8 @@ code {
 
 #leftColumn {
   min-width: 250px;
-  width: fit-content;
+  overflow-wrap: break-word;
+  word-wrap: break-word;
 }
 
 .releaseDate {


SheikhZaeem:
@romani In the old website where these names had spaces between them, we could separate the parts of the name and then put them in the next line. But here there a single long word. My solution just takes some part of the word to the next line.

romani:
GitHub, generate website

romani (approved):
I like this.

@MohanadKh03 , @Zopsss , please review.

MohanadKh03:
Yes this one is much better and would not depend on left column's layout anymore

Side note though not relating this PR is it normal not to have toggle here or should this be added to the improve rendering issues ?
Screenshot

Zopsss:
@romani please look at https://github.com/checkstyle/checkstyle/issues/16291#issuecomment-2651262553, we already have a PR to fix this issue

romani:
>is it normal not to have toggle here or should this be added to the improve rendering issues ?

Let's put in in separate PR, we have issue with all such small items to improve.
I agree that this arrow-down means "expanded" , but there is no way to collapse it, so no reason in such icon. But it might be big change to remove it. Let's put it on a side.

PR #16307, opened 2025-02-11
Issue #14019: Resolve pitest suppressions for XMLLogger  FileOpeningTag

zyadahmed:
Issue #14019  
using command  
` mvn -e --no-transfer-progress -Ppitest-common clean test-compile org.pitest:pitest-maven:mutationCoverage
`
Kill mutation for removing the call to the encoding method
Kill mutation for removing the call to the `writer.println` in writeFileClosingTag Method
image


romani (approved):
Good job.

Issue #16262, opened 2025-01-31
Update doc for MissingSwitchDefault


donalmurtagh:
I have read check documentation: https://checkstyle.org/checks/coding/missingswitchdefault.html#MissingSwitchDefault

I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Test.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
    <module name="TreeWalker">
        <module name="MissingSwitchDefault"/>
    </module>
</module>

/var/tmp $ cat Test.java
package example;

public class Test {
    public enum Status {
        ACTIVE, DISABLED
    }

    public void bug() {
        var status = Status.ACTIVE;

        switch (status) {
            case ACTIVE -> System.exit(0);
            case DISABLED -> System.exit(1);
        }
    }
}

/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-X.XX-all.jar -c config.xml Test.java

 [ERROR] Test.java:11:9: switch without "default" clause. [MissingSwitchDefault]
```
---

Confusion:
The `MissingSwitchDefault` check should not fail because the case statements are exhaustive and according to the document of this check

> This check does not validate any switch expressions. Rationale: The compiler requires switch expressions to be exhaustive. This means that all possible inputs must be covered.
---

Todo: we need to update documentation, see details at 
https://github.com/checkstyle/checkstyle/issues/16262#issuecomment-2635560488

romani:
please share compile-able source:
```
$ cat Test.java 
public class Test {
    public enum Status {
        ACTIVE, DISABLED
    }

    public void bug() {
        var status = Status.ACTIVE;

        switch (status) {
            case ACTIVE -> System.exit(0);
            case DISABLED -> System.exit(1);
        }
    }
}
roman@p5510:/var/tmp$ javac Test.java 
Test.java:10: error: : expected
            case ACTIVE -> System.exit(0);
                                         ^
Test.java:11: error: : expected
            case DISABLED -> System.exit(1);
                                           ^
2 errors

[...]

PR #16310, opened 2025-02-12
Issue #16262: update MissingSwitchDefault doc

HuzaifaOmar:
Issue: #16262 

## I've updated the code as follows:

- **Removed the Default Clause:** In the switch statement (see Example3.java), I've removed the default clause.
- **Added a Comment:** I've added an inline comment on the switch to indicate that it now intentionally violates the MissingSwitchDefault rule.
- **Updated Tests:** The tests (in MissingSwitchDefaultCheckExamplesTest.java) have been adjusted to reflect this change, so they now expect the violation accordingly.

### DIFF
--- a/src/site/xdoc/checks/coding/missingswitchdefault.xml
+++ b/src/site/xdoc/checks/coding/missingswitchdefault.xml
@@ -69,6 +69,17 @@ class Example1 {
         break;
     }
   }
+  enum Status {ACTIVE, DISABLED}
+  void testEnum(Status status) {
+    switch (status) { // violation, 'switch without "default" clause'
+      case ACTIVE:
+        System.out.println(0);
+        break;
+      case DISABLED:
+        System.out.println(1);
+        break;
+    }
+  }
 }
 </code></pre></div>
         <p id="Example2-code">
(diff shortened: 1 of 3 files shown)


src/site/xdoc/checks/coding/missingswitchdefault.xml
```diff
@@ -140,8 +140,6 @@ class Example3 {
       case GREEN:
         System.out.println("GREEN");
         break;
-      default:
-        System.out.println("Something else");
```
romani, line 144: You should do this removal from example java file.
Please find this code in java file.
After you run `mvn clean test` it will update ths xml by such removal.

mahfouz72, line 144: https://checkstyle.org/checks/coding/missingswitchdefault.html#Example3-code

This should be an example of correct code (code that has no violation) so we should not remove the default here. 

please leave Example3 as it is and add the switch with no default in [Example1](https://checkstyle.org/checks/coding/missingswitchdefault.html#Example1-code) to show a violation on a switch that uses an enum and covers all possible values.

```
    public enum Status {
        ACTIVE, DISABLED
    }

    void testEnum(Status status) {

        switch (status) {
            case ACTIVE -> System.exit(0);
            case DISABLED -> System.exit(1);
        }
    }

```

mahfouz72 (changes requested):
Item

pom.xml
```diff
@@ -822,6 +822,8 @@
           <compilerArgs>
             <arg>-Xpkginfo:always</arg>
           </compilerArgs>
+            <source>14</source>
+            <target>14</target>
```
romani, line 826: Please remove this, we are in jdk11 by default.

romani (changes requested):
Items

romani:
GitHub, generate website

pom.xml
HuzaifaOmar: yes, I am sorry this was by mistake. it was added automtically by the IDE.
will remove it now.

romani (approved):
Nice.

@mahfouz72 , please finalize review

mahfouz72 (approved):
Thanks alot

PR #16314, opened 2025-02-12
Issue #14019: Resolve pitest suppressions for XMLLogger fileMessages.…

zyadahmed:
#14019 
Mutation `fileMessages.remove(fileName);` 
The test case` testFileRemovalFromLogger `ensures that calling `fileFinished` multiple times does not duplicate file entries in the log. If a violation is logged only once, it means the file was already removed in a previous call, triggering fileMessages.remove. 
 
image

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
@@ -23,6 +23,7 @@ import static com.google.common.truth.Truth.assertWithMessage;
 
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.util.List;
 
 import org.junit.jupiter.api.Test;
 
@@ -380,6 +381,25 @@ public class XMLLoggerTest extends AbstractXmlTestSupport {
         verifyXml(getPath("ExpectedXMLLoggerSpecialName.xml"),
                 outStream, "<file name=" + "Test&amp;.java" + ">");
 
+    }
+
+    @Test
+    public void testVerifyLogger()
+            throws Exception {
+        final String inputFileWithConfig = "InputXMLLogger.java";
+        final String xmlReport = "ExpectedXMLLoggerWithChecker.xml";
+        verifyWithInlineConfigParserAndXmlLogger(inputFileWithConfig, xmlReport);
+    }
+
+    @Test
+    public void testVerifyLoggerWithMultipleInput()
+            throws Exception {
+        final String inputFileWithConfig = "InputXMLLogger.java";
+        final String expectedXmlReport = "ExpectedXMLLoggerDuplicatedFile.xml";
+        verifyWithInlineConfigParserAndXmlLogger(
+                inputFileWithConfig,
+                expectedXmlReport,
+                List.of(inputFileWithConfig, inputFileWithConfig));
     }
 
     @Test
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/xmllogger/ExpectedXMLLoggerDuplicatedFile.xml
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/xmllogger/ExpectedXMLLoggerDuplicatedFile.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<checkstyle version="null">
+<file name="InputXMLLogger.java">
+<error column="15" line="12" message="Using the '.*' form of import should be avoided - java.io.*." severity="error" source="com.puppycrawl.tools.checkstyle.checks.imports.AvoidStarImportCheck"/>
+</file>
+<file name="InputXMLLogger.java">
+<error column="15" line="12" message="Using the '.*' form of import should be avoided - java.io.*." severity="error" source="com.puppycrawl.tools.checkstyle.checks.imports.AvoidStarImportCheck"/>
+</file>
+</checkstyle>
--- a/src/test/java/com/puppycrawl/tools/checkstyle/AbstractXmlTestSupport.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/AbstractXmlTestSupport.java
@@ -23,7 +23,11 @@ import static com.google.common.truth.Truth.assertThat;
 import static com.google.common.truth.Truth.assertWithMessage;
 
 import java.io.ByteArrayOutputStream;
+import java.io.File;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 import java.util.function.BiPredicate;
 
@@ -33,6 +37,8 @@ import org.w3c.dom.Document;
 import org.w3c.dom.NamedNodeMap;
 import org.w3c.dom.Node;
 
+import com.puppycrawl.tools.checkstyle.bdd.InlineConfigParser;
+import com.puppycrawl.tools.checkstyle.bdd.TestInputConfiguration;
 import com.puppycrawl.tools.checkstyle.internal.utils.XmlUtil;
 
 public abstract class AbstractXmlTestSupport extends AbstractModuleTestSupport {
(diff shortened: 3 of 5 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
```diff
@@ -382,6 +382,28 @@ public void testFileOpenTag()
...
+                        "messages.properties", "key", null, SeverityLevel.ERROR, null,
+                        getClass(), null);
+        final AuditEvent auditViolation = new AuditEvent(this, "Test.java", violation);
+        logger.addError(auditViolation);
+
+        final AuditEvent fileFinishedEvent = new AuditEvent(this, "Test.java");
+        logger.fileFinished(fileFinishedEvent);
+        logger.fileFinished(fileFinishedEvent);
+        logger.auditFinished(null);
+        verifyXml(getPath("ExpectedXMLLoggerDuplicatedFile.xml"), outStream);
```
romani, line 404: Duplicate file in report is nonsense.

Even this test killed mutation, it is not ok.
In Checkstyle we practicing BDD, by writing real users use cases ( config and files), and all tests should be done like this.
We finished migration to such tests for Checks , and most Filters, but there are more areas where we still need to work to migrate tests to better model.

Looks like this is good point to start usage of new test model to kill mutation.

Please look at our test suit, there are examples on how to run config on multiple files.  We need to make new verifyxxxxxx method to add xml logger in execution. So in this case we will naturally hit removal of file from collection, and we will get good looking xml report.

romani (changes requested):
Let's try:

src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
romani: Please look at https://github.com/checkstyle/checkstyle/blob/21c8dfd315d5173df33d7c58c9020e432f75e006/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java#L334
We can not reuse it, as not using xml logger, but we can do something similar with xml logger.

zyadahmed: hi @romani  
I'm currently trying to figure out how to apply BDD in this case. As I understand it, BDD focuses on real user scenarios, but here, we don't have a direct way to apply it. As you know, killing mutations is sometimes difficult with typical real-world scenarios.
The only real case I used was checking for duplication, but I'm considering whether extending the Reflection module could provide a better workaround. Would that be a good approach, or is there a different way you’d recommend for handling this within BDD principles?
I’d appreciate any guidance on the right steps to take. Thanks!

zyadahmed: @romani, the only way I think this may work, following the same guidelines you provided, is to create VerifyXmlLogger to test the logging against the XML output and use that method along with a new method in ReflectionUtil to check whether an element exists in fileMessages. Since we do not have direct access to fileMessages, we cannot determine if it is removed or not because it uses Map.java methods internally, which we have no control over. The only new way I found to verify this behavior is by using reflection to inspect fileMessages and confirm whether an entry still exists after processing.

romani: we do not need reflection.

```
<file name="Test.java">
....
</file>
<file name="Test.java">
</file>
```

this is invalid report.
expected:
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<checkstyle version="null">
<file name="Test.java">
<error column="1" line="1" message="key" severity="error" source="com.puppycrawl.tools.checkstyle.XMLLoggerTest"/>
</file>
</checkstyle>
```

romani: lets try to make test like https://github.com/checkstyle/checkstyle/blob/23bef14d2c30f1ab1441098feb53e1dddd238411/src/xdocs-examples/java/com/puppycrawl/tools/checkstyle/checks/regexp/RegexpSinglelineCheckExamplesTest.java#L85

we need to extend AbstractModuleTestSupport that use `getBriefUtLogger` to let log in format convenient for Checks.

create with verifyxxxxxxxx methods that use XMLLogger, but take config from Input file as all others test are doing.

so your test method will looks like:
```
public void testNoduplicateTags() throws Exception {
        verifyWithInlineConfigParserAndXmlLogger(getPath("InputWthSomename.java"), "path/to/generated/report");
       verifyXml(getPath("ExpectedXMLLoggerDuplicatedFile.xml"), "path/to/generated/report");
    }
```

zyadahmed: Great! Now I understand—you need to add the necessary functionality in `AbstractModuleTestSupport` to test Java class checking with `XMLLogger`. Later, we can use it to refactor existing tests in the future. I'll work on it. But I have one last question. In `testNoDuplicateTags`, to properly kill the mutation, we need a real case of duplication. If I'm correct, what kind of duplication should I provide to test the following BDD scenario? I think it may be good to test with `AuditEvents` duplication. And thanks a lot for your explanation

romani: We can try to put the same file to be processed twice.
If we can not kill mutation by "real "execution of checkstyle validation, it good sign that code is not required.

zyadahmed: I have a suggestion: Instead of attempting to handle duplication by repeatedly executing the checker as it seem not working to kill mutation ,we can  consider a more reliable approach. The issue arises because checking for duplication on the same file produces the same result each time. Whether fileMessages.remove(fileName); is executed or not  the outcome remains unchanged as Map in xmlLogger overrides any existing entries. A better solution is to track the names of already processed files using a Set. Before writing a new file, check if it exists in the Set; if it does, return an error to prevent duplication. This approach ensures that duplicate files are effectively identified and handled. what do you think of that  ?

zyadahmed: > We can try to put the same file to be processed twice. If we can not kill mutation by "real "execution of checkstyle validation, it good sign that code is not required.


After trying multiple approaches, I believe removal may not be necessary. In all cases, if a duplicate exists, it is already removed when we add it again using Map.put(). If there is no duplication, everything will be discarded when the output stream is closed.

zyadahmed: i commited the `verifyWithInlineConfigParserAndXmlLogger`  is it need any modifaction ?

src/test/resources/com/puppycrawl/tools/checkstyle/xmllogger/ExpectedXMLLoggerDuplicatedFile.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<checkstyle version="null">
+<file name="Test.java">
+<error column="1" line="1" message="key" severity="error" source="com.puppycrawl.tools.checkstyle.XMLLoggerTest"/>
+</file>
+<file name="Test.java">
```
romani, line 6: this file is not in use, please remove it.

src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
```diff
@@ -382,6 +382,14 @@ public void testFileOpenTag()
 
     }
 
+    @Test
+    public void testVerifyLogger()
+            throws Exception {
+        final String className = "InputXMLLogger.java";
```
romani, line 388: `final String inputFileWithConfig`

src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
```diff
@@ -382,6 +382,14 @@ public void testFileOpenTag()
...
+        final String xmlReport = "ExpectedXMLLoggerWithChecker.xml";
+        verifyWithInlineConfigParserAndXmlLogger(className, xmlReport);
```
romani, line 390: please make one more  method like this that takes array of target files.

`verifyWithInlineConfigParserAndXmlLogger(inputFileWithConfig, xmlReport, targetFiles);`

where targetFiles is array files for Checker to run on, such attay should have two files `InputXMLLogger.java` (duplicate presence).
By Checkstyle API, we allow to send to Checker same file twice.
Expected Behavior is this case not clear, and we can accept dulicate tag at you have at src/test/resources/com/puppycrawl/tools/checkstyle/xmllogger/ExpectedXMLLoggerDuplicatedFile.xml, so ExpectedXMLLoggerDuplicatedFile.xml just should be upated to have duplcate validation from InputXMLLogger.java

romani (changes requested):
items:

src/test/resources/com/puppycrawl/tools/checkstyle/xmllogger/ExpectedXMLLoggerDuplicatedFile.xml
romani: not for removal, see comment below.

src/test/resources/com/puppycrawl/tools/checkstyle/xmllogger/InputXMLLoggerConfig.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
```
romani, line 1: we do not need this file, config should be loaded from src/test/resources/com/puppycrawl/tools/checkstyle/xmllogger/InputXMLLogger.java , please look at methods verifyXXxxxxxxxx

romani (changes requested):
we need pitest removal.


items:

src/test/resources/com/puppycrawl/tools/checkstyle/xmllogger/InputXMLLoggerConfig.xml
zyadahmed: pitest removal in this way or any real execution still exists. I think we should remove the line as it has no effect .

romani: please do removal, it is best effect of pitest, it helps to find dead code and not requireed code

zyadahmed: > we do not need this file, config should be loaded from src/test/resources/com/puppycrawl/tools/checkstyle/xmllogger/InputXMLLogger.java , please look at methods verifyXXxxxxxxxx

But if we have multiple files and only one config used in all of them, why can't we use the best way to separate the config, as it reflects the real use case of Checkstyle?

zyadahmed: > please do removal, it is best effect of pitest, it helps to find dead code and not requireed code

so you mean remove FileMessages.remove() from xmllogger right ?

romani: yes

zyadahmed: with pr #16345   and change in order of fileMessage.remove the mutation is killed 
image

zyadahmed:
pitest removal was already done in #16345, so I didn't change anything.
Regarding the logger, I made the change as you requested by moving it to inlineConf instead of a separate XML file, Do you mean something different? I don’t understand.

romani:
OMG, so concurrent updates, ok, let's merge without update for pitest.

romani (approved):
Thanks a lot.

Please help with other pitest survival, to make sure you fully understand how it works

zyadahmed:
> Thanks a lot.
> 
> Please help with other pitest survival, to make sure you fully understand how it works

Yes, I will start working on the next one to understand checkstyle more. Thank you very much for your help!

PR #16320, opened 2025-02-13
Issue #14019: kill mutation for SarifLogger under addError method 

Harsh-Agrawal96:
Issue #14019

Kill mutation for `ArgumentPropagationMutator on .replace(MESSAGE_PLACEHOLDER, escape(event.getMessage()))` under `SarifLogger`

**Reason for mutation:**
The reason is related with `escape` method on `SarifLogger.java` , In `addError` method we are using `escape` method to modify some selected characters of string `event.getMessage()` ( `violation` from `event class` ) but all test cases available has not any character that need to modify and falls to default switch condition of `escape` method then the `return` string of `escape` method is similar with `parameter` string of `escape` method.

**Solution:**
Added the test case which will have the character `\n` that must need to be modify by `escape` method then mutation kill and even not creating any new mutation.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
@@ -52,6 +52,28 @@ public class SarifLoggerTest extends AbstractModuleTestSupport {
         return "com/puppycrawl/tools/checkstyle/sariflogger";
     }
 
+    /**
+     * Executes the common logging steps for SARIF logging tests.
+     * This method mimic of natural execution at is done by Checker.
+     *
+     * <p>
+     * It starts file auditing, adds an error, and completes the audit process.
+     * </p>
+     *
+     * @param instance  the current instance of {@code SarifLoggerTest}, used for context
+     * @param logger    the {@code SarifLogger} instance to log errors
+     * @param fileName  the name of the file being audited
+     * @param violation the {@code Violation} object containing error details
+     */
+    public final void executeLogger(
+        SarifLoggerTest instance, SarifLogger logger, String fileName, Violation violation) {
+        final AuditEvent event = new AuditEvent(this, "Test.java", violation);
+        logger.fileStarted(event);
+        logger.addError(event);
+        logger.fileFinished(event);
+        logger.auditFinished(null);
+    }
+
     @Test
     public void testEscape() {
         final String[][] encodings = {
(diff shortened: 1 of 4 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
```diff
@@ -91,7 +91,41 @@ public void testAddError() throws IOException {
         logger.addError(ev);
         logger.fileFinished(ev);
         logger.auditFinished(null);
-        verifyContent(getPath("ExpectedSarifLoggerSingleError.sarif"), outStream);
+        verifyContent(getPath("ExpectedSarifLoggerSingleErrorOne.sarif"), outStream);
+    }
+
+    @Test
+    public void testAddErrorTwo() throws IOException {
```
romani, line 98: Please explain how 2 extra method kills single mutation.
https://www.diffchecker.com/MWHiOMar/

Please give meaningful names to method to preserve knowledge.

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
Harsh-Agrawal96: In `addError` method we have two condition in `event.getColumn()` that's why one method is for `getColumn() > 0` and one for other.

romani: So please put "column" to be part of name of method, to make in clear what is primary focus/nuance.

romani: We just need more meaningful names of methods.

src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
```diff
@@ -94,6 +94,40 @@ public void testAddError() throws IOException {
...
+        final Violation violation =
+                new Violation(1, 1,
+                        "messages.properties", "ruleId", null, SeverityLevel.ERROR, null,
+                        getClass(), "found an error\ntry again");
+        final AuditEvent ev = new AuditEvent(this, "Test.java", violation);
+        logger.fileStarted(ev);
+        logger.addError(ev);
+        logger.fileFinished(ev);
+        logger.auditFinished(null);
+        verifyContent(getPath("ExpectedSarifLoggerSingleErrorColumn1.sarif"), outStream);
```
romani, line 111: Please extract common part of this 3 method in new method, to show explicitly that that execution is mimic of natural execution in Checker but we do it by low level.

```
public void testAddErrorAtColumn1() throws IOException {
        final SarifLogger logger = new SarifLogger(outStream,
                OutputStreamOptions.CLOSE);
        
        final Violation violation =
                new Violation(1, 1,
                        "messages.properties", "ruleId", null, SeverityLevel.ERROR, null,
                        getClass(), "found an error\ntry again");
        
        executeLogger(this, logger,  "Test.java", violation);
        verifyContent(getPath("ExpectedSarifLoggerSingleErrorColumn1.sarif"), outStream);
    }
```

src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
```diff
@@ -52,6 +52,29 @@ protected String getPackageLocation() {
...
+     * </p>
+     *
+     * @param instance  the current instance of {@code SarifLoggerTest}, used for context
+     * @param logger    the {@code SarifLogger} instance to log errors
+     * @param fileName  the name of the file being audited
+     * @param violation the {@code Violation} object containing error details
+     */
+    public final void executeLogger(
+        SarifLoggerTest instance, SarifLogger logger, String fileName, Violation violation) {
+        final AuditEvent ev = new AuditEvent(this, "Test.java", violation);
```
romani, line 71: ev , should be `event`, no reason to be cryptic, we have space.

src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
```diff
@@ -52,6 +52,29 @@ protected String getPackageLocation() {
         return "com/puppycrawl/tools/checkstyle/sariflogger";
     }
 
+    /**
+     * Executes the common logging steps for SARIF logging tests.
+     * This method helps avoid code duplication by performing a sequence of actions
+     * related to logging audit events.
```
romani, line 58: This method mimic of natural execution at is done by Checker.

romani (changes requested):
Last:

src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
Harsh-Agrawal96: replace the all three lines or only last two lines ??

romani (approved):
Thanks a lot

Issue #16006, opened 2024-12-06
Provide a link to GitHub Issues in Maven generated Issue Management section.


sdavids:
**How it works Now:**

https://checkstyle.org/issue-management.html

> This project uses GitHub Issues to manage its issues.

**Describe the solution you'd like**

"GitHub Issues" or "issues" should be a link to https://github.com/checkstyle/checkstyle/issues

In addition, the "Issue Management" section could be reworded.

romani:
But I do not think we have this wording in our repository sources, it might be generated by maven plugin.

sdavids:
https://maven.apache.org/pom.html#Issue_Management

sdavids:
```xml
  <issueManagement>
    <system>GitHub</system>
    <url> https://github.com/checkstyle/checkstyle/issues</url>
  </issueManagement>
```

romani:
Please send Pull Request.

sdavids:
```shell
$ mvn org.apache.maven.plugins:maven-project-info-reports-plugin:issue-management
[INFO] Scanning for projects...
[INFO] Inspecting build with total of 1 modules...
[INFO] Installing Nexus Staging features:
[INFO]   ... total of 1 executions of maven-deploy-plugin replaced with nexus-staging-maven-plugin
[INFO] 
[INFO] ------------------< com.puppycrawl.tools:checkstyle >-------------------
[INFO] Building checkstyle 10.21.0-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-project-info-reports-plugin:3.8.0:issue-management (default-cli) @ checkstyle ---
[WARNING] Site model of 'com.puppycrawl.tools:checkstyle:jar:10.21.0-SNAPSHOT' for default locale is still using the old pre-version 2.0.0 model. You MUST migrate to the new model as soon as possible otherwise your build will break in the future!
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  0.832 s
[INFO] Finished at: 2024-12-11T14:41:46+01:00
[INFO] ------------------------------------------------------------------------
```

`maven-project-info-reports-plugin:3.8.0` uses Doxia 2.0.0:

https://issues.apache.org/jira/browse/MPIR-470

[...]

Issue #13426, opened 2023-07-24
Xdoc file generation should be separate project that executes in maven build as just another phase; remove doxia dependency 


stoyanK7:
The final step of this project is to have the generation happen in `mvn site phase`. As of now, we use a `@BeforeAll` method in `XdocsPagesTest` to generate Xdocs. We use a custom parser, custom sink, and custom sink factory. The reason we have them is because we want to isolate macro execution from parsing the rest of the file to HTML. Once the whole repository is converted to templates and we are certain the generation is good, those 3 modules will be removed. The default `XdocParser` in `mvn site` phase will execute the macros and parse to HTML for us.

please beware of https://github.com/checkstyle/checkstyle/issues/13770 fix that introduced extra overrides to doxia files that might cause extra complexity in this issue. Completing https://github.com/checkstyle/checkstyle/issues/13426 means we will be removing the classes we copied and hacked from Doxia (XdocsTemplateSink) which is documented at

https://github.com/checkstyle/checkstyle/blob/f6b70cc23d566fa306bb86405426f79fb5844310/src/main/java/com/puppycrawl/tools/checkstyle/site/XdocsTemplateSink.java#L33-L34

. When completing https://github.com/checkstyle/checkstyle/issues/13426 and removing the class, we need to re-open https://github.com/checkstyle/checkstyle/issues/13770 and find a permanent fix unless doxia resolves our internal issues at https://issues.apache.org/jira/browse/DOXIA-707 .

Interesting note at https://issues.apache.org/jira/browse/DOXIA-707

>Since you want to migrate to a Maven Site Plugin approach, you don't need to generate XDoc at all. [kwin](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=kwin), has just migrated Maven Plugin Tools off XDoc to an abstract sink provided by Maven Reporting Impl.

Move xdoc doxia parsers and macroses to separate project in contribution repo
detected at https://github.com/checkstyle/checkstyle/pull/13357#discussion_r1267583673
we should move all out, to note pollute in main repo. We should not export such classes to final jar we release.

romani:
We should do this only as very last issue, as it has no much benefit to users and simply design beautification, that is also good, but  we have bunch of other issues to be done first.

romani:
So we can move Xml Metadata and Xml Xdoc files generation to separate project, it will take all not essential dependencies to it.
[...]

PR #16321, opened 2025-02-13
[WIP] Issue #16006 - Upgrade doxia version

roulpriya:
Upgrade doxia to 2.0.0
Fixes Issue #16006, Pull #15729

## Changes

1. Doxia 2.0 uses Eclipse Sisu + Guice as DI framework with JSR-330 annotations, introduced new dependencies and migrated plexus components to JSR-330
2. Updated implementations of Sink to Doxia 2.0 API

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -237,7 +237,7 @@
     <checkerframework.version>3.48.4</checkerframework.version>
     <error-prone.version>2.31.0</error-prone.version>
     <error-prone-support.version>0.15.0</error-prone-support.version>
-    <doxia.version>1.12.0</doxia.version>
+    <doxia.version>2.0.0</doxia.version>
     <error-prone.configuration-args>
       -Xep:AmbiguousJsonCreator:ERROR
       -Xep:AssertJIsNull:ERROR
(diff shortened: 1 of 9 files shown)


roulpriya:
Looks like we are blocked by #13426 from various todo comments from code.

Running a debugger through the failing tests reveal that, Doxia appears to use `XdocTemplateSink` wrapped in another `SinkWrapper`, which is violating explicit `instanceof` checks in multiple places like:

https://github.com/checkstyle/checkstyle/blob/21c8dfd315d5173df33d7c58c9020e432f75e006/src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java#L118-L129

https://github.com/checkstyle/checkstyle/blob/21c8dfd315d5173df33d7c58c9020e432f75e006/src/main/java/com/puppycrawl/tools/checkstyle/site/ViolationMessagesMacro.java#L39-L48

romani:
It should not block, if new doxia use other classes , let's update to use new classes

roulpriya:
We depend on the behaviour of `XdocSink` further in the code. 

Looking deeper into code, Doxia wraps it in a JDK proxy, which makes it _difficult_ to retrieve the underlying `XdocSink`.

https://github.com/apache/maven-doxia/blob/master/doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml5BaseParser.java#L143-L145

We use `XdocSink` to call `setInsertNewline`, which changes the behaviour here:

https://github.com/apache/maven-doxia/blob/master/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/AbstractXmlSink.java#L113-L115

I tried removing the calls, but it drastically changed the formatting of the generated templates. Do you have a workaround that you would like to suggest?

romani:
if this issue is complicated, please close PR and and help us with good xxxx issues, it will help to be familiar with our project and you will be able to resolve this issue on second try.

PR #16323, opened 2025-02-14
Issue #14019: Resolve pitest suppressions for ScopeUtil

Mostafahassen1:
[ISSUE14019](https://github.com/checkstyle/checkstyle/issues/14019)  
## **ScopeUtil**

https://github.com/checkstyle/checkstyle/blob/6e0a0bf7b57e725f98dd79ffedf523fb7679f267/src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java#L207

### RemoveConditionalMutator_EQUAL_IF**
A mutation test was performed on the condition `parent.getParent().getType() == memberType`.  

---

### **Method: `isInBlockOf`**

The `isInBlockOf` method determines whether a given `DetailAST` node is directly contained within a specified block type in the Abstract Syntax Tree (AST). It works by traversing the AST upwards from the given node, checking each parent node until it either finds the specified block type or determines that the node is not within the desired block.

#### **Code Implementation**
```java
private static boolean isInBlockOf(DetailAST node, int tokenType) {
    DetailAST token = node.getParent();
    boolean found = false;

    while (!found) {
        if (token == null) {
            found = false;
            break;
        }

        if (token.getType() == tokenType) {
            found = true;
        }
        else if (token.getType() == TokenTypes.LITERAL_NEW
               || TokenUtil.isTypeDeclaration(token.getType())) {
            break;
        }
        token = token.getParent();
    }

    return found;
}
```

### **Mutation Test Results**
- The mutation test replaced a specific checker with `true`, and the test was performed to evaluate the robustness of the test suite.  

---

### **Method File**

- **File Location**: [ScopeUtil.java](https://github.com/checkstyle/checkstyle/blob/3c75834a7b2f643001edacfbbaf731d6d7efe8d9/src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java#L202)


romani:
Changing code is ok, end even completely ok, as pitest hint to remove all extra code and potentially simplify code.

Direction is good , please fix CI failures, looks like there is side effects of this change. 
Please also try to analyze what checks are using such method, and we will need regression diff report execution for all of them.

Mostafahassen1:
Hi @romani,  

I have two questions regarding our workflow:  

---

### **First Question**  
You mentioned that we need a **regression diff report execution** for all changes. I tried generating it using the following commands:  

```bash
mvn test > test-results-before.txt  
mvn test > test-results-after.txt  
git diff test-results-before.txt test-results-after.txt > regression-diff-report.txt  
```  

However, the resulting `regression-diff-report.txt` file is empty. Am I missing something in the process?  

---

### **Second Question**  
Is there a command to run **mutation testing locally** before pushing changes to CI/CD?  

I tried running:  
```bash
mvn org.pitest:pitest-maven:mutationCoverage  
```  

image

romani:
> Is there a command to run mutation testing locally before pushing changes to CI/CD?

yes, all in CI is executable on local.
here is command https://github.com/checkstyle/checkstyle/actions/runs/13322737294/job/37210184291?pr=16323#step:6:1
`./.ci/pitest.sh "pitest-utils"`

romani:
Yes, We need to find such java input files that kill such mutation. some Check needs to run to catch this.
We do not accept pure unit testing.

Mostafahassen1:
> Yes, We need to find such java input files that kill such mutation

@romani   could you provide more details about "Java input files"? I'm not familiar with this part of the project, or I’m having trouble understanding what you mean.

romani:
this is usage:
image
image
ClassMemberImpliedModifierCheck
ClassMemberImpliedModifierCheckTest
Input files like InputClassMemberImpliedModifierOnClass.java

so goal is by changes in such files (almost real java code that must compile) we kill mutation.
If this Check can not do it, try by other Checks that use this util method.

if not possible to kill, please share detailed explanation and we will need to update util method to remove extra code.

Mostafahassen1:
@romani  Have we now successfully solved the mutation test? , I now there is two commit

romani:
One of videos shows how to keep single commit, web page also has list of commands.

romani:
After you fix all CI failures.
We will need to run regression diff report for all Checks that use changed method to make sure there is no regression.

Mostafahassen1:
>We will need to run regression diff report for all Checks that use changed method to make sure there is no regression.

@romani  sorry what mean that

romani:
Create list of Checks that use method you updated.

Mostafahassen1:
Hi @romani  I found a way to break the mutation test by using two `break;` statements inside a method. However, the rule states that there should only be one `break;` inside the method. I can't find another way to break the mutation test.

romani:
Did you try to apply to code mutation from pitest manually and run regression testing on big real code by our diff regression report? 
If you change as pitest suggests , something will be different so something that can be highlighted in report and we just need to copy such java code to Inputs.

Mostafahassen1:
@romani 
>Did you try to apply to code mutation from pitest manually and run regression testing on big real code by our diff regression report?
If you change as pitest suggests , something will be different so something that can be highlighted in report and we just need to copy such java code to Inputs.


I haven’t done that. Is there an article or resource that explains how to do it in more detail?

romani:
I don't think so , just change code as pitest suggests.
Make sure all CI pass, generate diff regression report for affected Checks.

Mostafahassen1:
@romani 
>just change code as pitest suggests.
Make sure all CI pass, generate diff regression report for affected Checks.

How can generate diff regression report  ? 

Is there a resource to read about how to do that?

romani:
https://github.com/checkstyle/checkstyle/blob/6e0a0bf7b57e725f98dd79ffedf523fb7679f267/src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java#L207

You need to change only line that is referenced in mutation, not whole expression.

Mostafahassen1:
@romani 
https://github.com/checkstyle/checkstyle/blob/6e0a0bf7b57e725f98dd79ffedf523fb7679f267/src/main/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtil.java#L207

ok but now I'm not use this line in method

romani:
Ok, if you continue with your version of code, just prove that that it works for all Checks that use this method in big real projects.

Mostafahassen1:
> Ok, if you continue with your version of code, just prove that that it works for all Checks that use this method in big real projects.

@romani 

The test replaces the condition `while (!found)` in the `while` loop with `true`, which can cause the loop to run indefinitely or crash if `token` becomes `null`. 

However, when the code is rerun after the check, if `found` becomes `true` during the mutation test, the loop will still execute. 
The mutation will break when it encounters the condition `if (token == null)` because `found` will be set to `false`, and the method will return `false`.


### Code

```java
private static boolean isInBlockOf(DetailAST node, int tokenType) {
    DetailAST token = node.getParent();
    boolean found = false;

    while (true) {
        if (token == null) {
            found = false;
            break;
        }

        if (token.getType() == tokenType) {
            found = true;
        } else if (token.getType() == TokenTypes.LITERAL_NEW
                || TokenUtil.isTypeDeclaration(token.getType())) {
            break;
        }
        token = token.getParent();
    }

    return found;
}
```

---

### Error

```bash
[ERROR] ScopeUtilTest.testInClassBlock:53 Should return true when passed is class
value of: isInClassBlock(...)
expected to be true
[ERROR] ScopeUtilTest.testInRecordBlock:147 Should return true when passed is record
value of: isInRecordBlock(...)
expected to be true
[ERROR] ScopeUtilTest.testIsInAnnotationBlock:348 Should return true when node is annotation block
value of: isInAnnotationBlock(...)
expected to be true
[ERROR] ScopeUtilTest.testIsInInterfaceBlock:335 Should return true when node is interface block
value of: isInInterfaceBlock(...)
expected to be true
[ERROR] ScopeUtilTest.testisInInterfaceOrAnnotationBlock:359 Should return true when node is in interface or annotation block
value of: isInInterfaceOrAnnotationBlock(...)
expected to be true
```

romani:
This means that you do mutation not same as pitest is doing.

Please reread https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports

Mostafahassen1:
>This means that you do mutation not same as pitest is doing.

@romani 
is this code is solve the issue ? but it has `two return conditions`
If this does not solve the issue, please provide the correct direction.

```java
public static boolean isInBlockOf(DetailAST node, int tokenType) {
    boolean result = false;  
    for (DetailAST token = node.getParent(); token != null && !result; token = token.getParent()) {
        if (token.getType() == tokenType) {
            result = true;  
        }
        if (token.getType() == TokenTypes.LITERAL_NEW 
            || TokenUtil.isTypeDeclaration(token.getType())) {
            break;  
        }
    }
    return result; 
}
```

romani:
>is this code is solve the issue ?

I don't know. Only regression report will prove if that is ok.
Two returns are always refactor-able to single return.

romani:
If this issue is too heavy, please come back to "good third issue" group, as become more experienced, you can continue Pitest issues.

Mostafahassen1:
>If this issue is too heavy, please come back to "good third issue" group, as become more experienced, you can continue Pitest issues.
@romani 
OK ,  I will do that

romani:
Try to archive similar state to https://github.com/checkstyle/checkstyle/pull/16438,

Try to look all all other PR merged for this issue, read comments, look at commits, you should be able to spot  a trick.

You can always can try another pitest mutation, it might be easier to handle as first pitest update.

Mostafahassen1:
> You can always can try another pitest mutation, it might be easier to handle as first pitest update. 

 Can you guide me to an easy mutation test issue, or would you prefer me to go to `issue group three`?

romani:
4) SuppressionXpathSingleFilter, ` MemberVariableMutator on this.message = null; should be fixed by similar test as in above PR`


just look in git history how other mutations covered for this Filter for sumilar mutations.

romani:
@Mostafahassen1 , try to learn from https://github.com/checkstyle/checkstyle/pull/16450

romani:
No changes, PR is broken

Issue #16300, opened 2025-02-10
Static html for syntax highlight in web site


romani:
It is already highlighted by activity at https://github.com/checkstyle/checkstyle/issues/16073

https://github.com/checkstyle/checkstyle/blob/23bef14d2c30f1ab1441098feb53e1dddd238411/src/site/site.xml#L52-L56

But syntax highlight happening by JS in runtime. That is ok, but there is way to do it during website generation, so no JS will be required and it will improve stability of our website as less will happen in runtime.

From that issue: https://github.com/checkstyle/checkstyle/issues/16073#issuecomment-2569603330

Autodected if highlighing is enabled in the skin. Here is the the magic: https://github.com/apache/maven-fluido-skin/blob/ded1227ddb48ddf30ff83e7ae64a70be257dd282/src/main/resources/META-INF/maven/site.vm#L242. Your output used the old `div` style which have bee replaced by a name HTML5 approach.

So content of code is highly sliced-diced in numerous chunks (code and span), each has its own style:
![Image](https://github.com/user-attachments/assets/8bb0f9d5-5bd3-438e-8ad3-d0cda9f21dbf)

SheikhZaeem:
@romani I am starting working on this issue. I will tell you how it goes.

smita1078:
Hi,
@SheikhZaeem @romani I have got the fix for this issue and its working too, can I go ahead with creating a PR if you are fine with that..?

SheikhZaeem:
@smita1078 sorry i almost did it and am about to make a PR. I didn't see any "I am working on this" on this issue so i decided to take it.

smita1078:
Oh Okay, no worries !:)

romani:
If work is done in both please do review for each other.

SheikhZaeem:
@smita1078, @romani  if you can help with the layout issue that would be really helpful.

smita1078:
So actually here u used, 
```
<custom>
    <fluidoSkin>
      <highlightjs>true</highlightjs>
      <highlightjsTheme>github-gist</highlightjsTheme>
      <highlightjsVersion>11.9.0</highlightjsVersion>
    </fluidoSkin>
  </custom>
```

in site.xml
which is basically using highlight.js library but instead we need to enable highlight through this configuration itself..
 ```
<skin>
    <groupId>org.apache.maven.skins</groupId>
    <artifactId>maven-fluido-skin</artifactId>
    <version>2.0.1</version>
  </skin>
```
[...]

PR #16334, opened 2025-02-15
Issue #16300: Static html for syntax highlight in web site

smita1078:
Issue:  #16300

- modified css, skin automatically applies this by converting nowrap->pre-wrap
- removed unuseful js libraries that were earlier used to syntax highlighting
- also removed that code block that added newlines, as this is also handled by Skin itself

Output:

image


Zopsss:
github, generate site

src/site/resources/css/site.css
```diff
@@ -143,7 +143,6 @@ th {
 .prettyprint {
   padding: 0.5em !important;
   overflow: auto;
-  white-space: nowrap;
```
Zopsss, line 146: why this was removed? 👀

smita1078: This is already taken care through skin, if not removed it was giving all code in single line with horizontal scroll

Zopsss: default was pre-wrap? I'm referring to: 

> modified css, skin automatically applies this by converting nowrap->pre-wrap

smita1078: default was nowrap, but this was overriding skins inbuilt css and giving all code in a single line, when removed it got fixed and skins pre-wrap was applied instead

smita1078:
GitHub, generate website

src/site/site.xml
```diff
@@ -44,16 +44,8 @@
...
-        <script type="text/javascript"
-              src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.3/jquery.min.js"></script>
         <script type="text/javascript" src="$relativePath/js/anchors.js"></script>
         <script type="text/javascript" src="$relativePath/js/google-analytics.js"></script>
-        <script src="https://cdnjs.cloudflare.com/ajax/libs/prettify/1.0.0/prettify.min.js"></script>
-        <script type="text/javascript"
-              src="https://cdn.jsdelivr.net/gh/google/code-prettify@master/loader/run_prettify.js"></script>
-        <link rel="stylesheet"
-              href="https://cdn.jsdelivr.net/gh/google/code-prettify@master/styles/prettify.css" type="text/css">
-        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/prettify/1.0.0/prettify.min.css">
```
Zopsss, line 56: we removed the external JS library and syntax highlighting is still working. [In this comment](https://github.com/checkstyle/checkstyle/issues/16300#issuecomment-2660983078), you said we just needed to use following in our `site.xml`:

```xml
<skin>
   <groupId>org.apache.maven.skins</groupId>
   <artifactId>maven-fluido-skin</artifactId>
   <version>2.0.1</version>
 </skin>
```

And we needed to add `prettyprint` class to the `<pre>` tag. You added it in your previous PR: 

https://github.com/smita1078/checkstyle/blob/2b7fd3535f42cce05f33408f96fbc892fc5650da/src/main/java/com/puppycrawl/tools/checkstyle/site/ExampleMacro.java#L193

Apache also does something similar: 

https://github.com/apache/maven-fluido-skin/blob/f090098cd12a3ff361471ae5d915a540dbfb8d22/src/main/resources/META-INF/maven/site.vm#L242

this if condition adds the `prettyprint` class to their code snippets so their code snippets also gets highlighted, on top of it, they also adds `linenums` class with `prettyprint` so they also get line numbers in their snippets:

https://github.com/apache/maven-fluido-skin/blob/f090098cd12a3ff361471ae5d915a540dbfb8d22/src/main/resources/META-INF/maven/site.vm#L236-L240

Am I going in right direction? If I'm correct then I think we didn't needed to add external JS library, we only needed to add `prettyprint` class to the `<pre>` tag.

smita1078: > this if condition adds the prettyprint class to their code snippets so their code snippets also gets highlighted, on top of it, they also adds linenums class with prettyprint so they also get line numbers in their snippets:

Yes, exactly, we don't need any external js library just prettyprint class and pre tags

smita1078:
Fix already merged via #16439

PR #16349, opened 2025-02-17
CheckStyle -  IDEA Code Style configuration

Anmol202005:
IntelliJ IDEA Code Style configuration that matches our checkstyle_check.xml settings

### DIFF
--- a/config/intellij-checkstyle-config.xml
+++ b/config/intellij-checkstyle-config.xml
@@ -1,0 +1,73 @@
+<code_scheme name="CodeStyle" version="173">
+  <JavaCodeStyleSettings>
+    <option name="GENERATE_FINAL_LOCALS" value="true" />
+    <option name="SPACE_AFTER_CLOSING_ANGLE_BRACKET_IN_TYPE_ARGUMENT" value="true" />
+    <option name="ANNOTATION_PARAMETER_WRAP" value="1" />
+    <option name="ENUM_FIELD_ANNOTATION_WRAP" value="1" />
+    <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
+    <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
+    <option name="IMPORT_LAYOUT_TABLE">
+      <value>
+        <package name="" withSubpackages="true" static="true" />
+        <emptyLine />
+        <package name="java" withSubpackages="true" static="false" />
+        <emptyLine />
+        <package name="javax" withSubpackages="true" static="false" />
+        <emptyLine />
+        <package name="org" withSubpackages="true" static="false" />
+        <emptyLine />
+        <package name="" withSubpackages="true" static="false" />
+      </value>
+    </option>
+    <option name="ALIGN_MULTILINE_RECORDS" value="false" />
+    <option name="ALIGN_MULTILINE_DECONSTRUCTION_LIST_COMPONENTS" value="false" />
+    <option name="NEW_LINE_AFTER_LPAREN_IN_DECONSTRUCTION_PATTERN" value="false" />
+    <option name="RPAREN_ON_NEW_LINE_IN_DECONSTRUCTION_PATTERN" value="false" />
+    <option name="ALIGN_TYPES_IN_MULTI_CATCH" value="false" />
+    <option name="JD_DO_NOT_WRAP_ONE_LINE_COMMENTS" value="true" />
+    <option name="JD_PRESERVE_LINE_FEEDS" value="true" />
+  </JavaCodeStyleSettings>
+  <codeStyleSettings language="JAVA">
+    <option name="RIGHT_MARGIN" value="100" />
+    <option name="LINE_COMMENT_AT_FIRST_COLUMN" value="false" />
+    <option name="BLOCK_COMMENT_AT_FIRST_COLUMN" value="false" />
+    <option name="KEEP_FIRST_COLUMN_COMMENT" value="false" />
+    <option name="KEEP_CONTROL_STATEMENT_IN_ONE_LINE" value="false" />
+    <option name="KEEP_BLANK_LINES_IN_DECLARATIONS" value="1" />
+    <option name="KEEP_BLANK_LINES_IN_CODE" value="1" />
+    <option name="KEEP_BLANK_LINES_BETWEEN_PACKAGE_DECLARATION_AND_HEADER" value="1" />
+    <option name="KEEP_BLANK_LINES_BEFORE_RBRACE" value="1" />
+    <option name="BLANK_LINES_BEFORE_PACKAGE" value="1" />
+    <option name="ELSE_ON_NEW_LINE" value="true" />
+    <option name="CATCH_ON_NEW_LINE" value="true" />
+    <option name="FINALLY_ON_NEW_LINE" value="true" />
+    <option name="ALIGN_MULTILINE_RESOURCES" value="false" />
+    <option name="SPACE_AROUND_SHIFT_OPERATORS" value="false" />
+    <option name="SPACE_AROUND_LAMBDA_ARROW" value="false" />
+    <option name="SPACE_BEFORE_METHOD_CALL_PARENTHESES" value="true" />
+    <option name="SPACE_BEFORE_METHOD_PARENTHESES" value="true" />
+    <option name="SPACE_BEFORE_SYNCHRONIZED_PARENTHESES" value="false" />
+    <option name="SPACE_BEFORE_SYNCHRONIZED_LBRACE" value="false" />
+    <option name="CALL_PARAMETERS_WRAP" value="1" />
+    <option name="RESOURCE_LIST_WRAP" value="1" />
+    <option name="EXTENDS_LIST_WRAP" value="1" />
+    <option name="THROWS_LIST_WRAP" value="1" />
+    <option name="THROWS_KEYWORD_WRAP" value="1" />
+    <option name="METHOD_CALL_CHAIN_WRAP" value="1" />
+    <option name="BINARY_OPERATION_WRAP" value="1" />
+    <option name="BINARY_OPERATION_SIGN_ON_NEXT_LINE" value="true" />
+    <option name="TERNARY_OPERATION_WRAP" value="1" />
+    <option name="TERNARY_OPERATION_SIGNS_ON_NEXT_LINE" value="true" />
+    <option name="FOR_STATEMENT_WRAP" value="1" />
+    <option name="ARRAY_INITIALIZER_WRAP" value="1" />
+    <option name="ASSIGNMENT_WRAP" value="1" />
+    <option name="WRAP_COMMENTS" value="true" />
+    <option name="ASSERT_STATEMENT_WRAP" value="1" />
+    <option name="ASSERT_STATEMENT_COLON_ON_NEXT_LINE" value="true" />
+    <option name="WRAP_LONG_LINES" value="true" />
+    <option name="PARAMETER_ANNOTATION_WRAP" value="1" />
+    <option name="VARIABLE_ANNOTATION_WRAP" value="1" />
+    <option name="ENUM_CONSTANTS_WRAP" value="1" />
+    <option name="WRAP_ON_TYPING" value="0" />
+  </codeStyleSettings>
+</code_scheme>


romani:
Please create issue and list Checks that we have , list of checkboxes, and checkout all that you covered, so we might improve step by step to cover all other

config/intellij-checkstyle-config.xml
```diff
@@ -0,0 +1,73 @@
+<code_scheme name="CodeStyle" version="173">
```
romani, line 1: `Checkstyle code style`

Let's skip version.

romani (changes requested):
Items

Anmol202005:
@ romani issue opened : #16427

config/intellij-checkstyle-config.xml
Anmol202005: done

romani:
Please keep ci green. 
We usually do not review until CI is green.

config/intellij-checkstyle-config.xml
```diff
@@ -0,0 +1,77 @@
...
+    <option name="PARAMETER_ANNOTATION_WRAP" value="1" />
+    <option name="VARIABLE_ANNOTATION_WRAP" value="1" />
+    <option name="ENUM_CONSTANTS_WRAP" value="1" />
+    <indentOptions>
+      <option name="CONTINUATION_INDENT_SIZE" value="4" />
+      <option name="KEEP_INDENTS_ON_EMPTY_LINES" value="true" />
+    </indentOptions>
+  </codeStyleSettings>
+  <codeStyleSettings language="kotlin">
+    <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
```
romani, line 75: Let's keep only java language

romani (changes requested):
Items

config/intellij-checkstyle-config.xml
```diff
@@ -0,0 +1,77 @@
...
+        <package name="" withSubpackages="true" static="true" />
+      </value>
+    </option>
+  </JavaCodeStyleSettings>
+  <JetCodeStyleSettings>
+    <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
+  </JetCodeStyleSettings>
+  <codeStyleSettings language="Cookie">
+    <indentOptions>
+      <option name="USE_TAB_CHARACTER" value="false" />
```
romani, line 27: Is this for java ?

romani (changes requested):
Items

romani:
Did you reformat code by idea?
 looks suspicious that no changes detected.

config/intellij-checkstyle-config.xml
Anmol202005: done

config/intellij-checkstyle-config.xml
Anmol202005: nope . removed it .

Anmol202005:
@romani  Yes, I did reformat the code using IntelliJ IDEA's formatter. 

Before reformatting, Checkstyle reported 28 errors, and after reformatting, it's down to 15 errors. The remaining issues require manual fixes :

- Class name `test` should start with uppercase (naming convention)
- Variables that could be declared as `final`
- Operator placement (`&&` and `+` should be on new lines)
- Missing newline at end of file

Here's the test code for reference:

```java
import java.util.ArrayList;
import java.util.Collections;
import java.util.HashMap;
import java.util.List;
import java.util.Map;

public class test {
     private static final int MAX_LENGTH = 100;
    private final List<String> items;
    public Map<String, Integer> data = new HashMap<>();

    public test() {this.items = new ArrayList<>();}
    
    public test(List<String> items) {
        this.items = items;
    }
    
    /**
     * This method violates several formatting rules.
     */
    public List<String> processItems(String input, int count) throws IllegalArgumentException {
        if (input == null) { return null; }
        
        String veryLongString = "This string intentionally goes beyond the 100 character limit to demonstrate a line length violation in the configuration";
        
        boolean isValid = input.length() > 0 &&
                count > 0;
        
        String[] parts = new String[]{"a", "b", "c"};
        
                List<String> results = new ArrayList<>();
        
        if(isValid) {
            for (int i = 0; i < count; i++)
            {
                results.add(input + i);
            }
        } else return Collections.emptyList();
        
        try {
            validateResults(results);
        } catch (Exception e) {
            results.clear();
        }finally {
            System.out.println("Done");
        }
        
        String result;
        if (results.isEmpty()) {
            result = "empty" +
                "list";
        }
        else {
            result = "populated list";
        }
        
        return results;
    }
    
    private void validateResults(List<String> results) {
        assert results != null:"Results cannot be null";
        
        String status;
        if (results.isEmpty()) {
            status = "Empty";
        }
        else {
            status = "Not empty";
        }
    }
    
    public enum ProcessingType { SIMPLE, NORMAL, COMPLEX, ADVANCED, SPECIAL, CUSTOM }
}

**Comment by romani** [comment]
> Before reformatting, Checkstyle reported 28 errors, and after reformatting, it's down to 15 errors. 

please share in commit all IDEA formatting.
We have 0 errors now from checkstyle, I am not sure how you get 15 and 28.

**Comment by Anmol202005** [comment]
> please share in commit all IDEA formatting. We have 0 errors now from checkstyle, I am not sure how you get 15 and 28.

@romani i have commited the idea config that's i used 

i guess there will be violations  by checkstyle here like 
```
[ERROR] C:\Users\anmol\IdeaProjects\checkstyle3\src\main\java\com\puppycrawl\tools\checkstyle\temp\test.java:36:9: 'if' is not followed by whitespace. [WhitespaceAfter]
[ERROR] C:\Users\anmol\IdeaProjects\checkstyle3\src\main\java\com\puppycrawl\tools\checkstyle\temp\test.java:36:9: 'if' is not followed by whitespace. [WhitespaceAround]
```

pls clarify

**Comment by romani** [comment]
I am confused, there is no `test.java` in our repository. Lets not put extra files.

File: config/intellij-checkstyle-config.xml
Comment (romani) [inline config/intellij-checkstyle-config.xml:1]: change name to match naming parrtern:
```
intellij-checkstyle-config.xml
intellij-idea-inspection-scope.xml
intellij-idea-inspections-misc.xml
intellij-idea-inspections.properties
intellij-idea-inspections.xml
```


`intellij-idea-code-style.xml`

**Review by romani — CHANGES_REQUESTED** [review]
items:

**Comment by romani** [comment]
I imported your config, 

select CheckstyleAntTask.java  , right click, choose "Reformat code" menu, in popup window select only "Optimize imports".
Whole file is changed :( .

**Comment by Anmol202005** [comment]
Yep, fixed that! realized that it wasn't working because the version was missing in the config—we had removed it earlier.  🚀

File: config/intellij-checkstyle-config.xml
Comment (Anmol202005) [reply config/intellij-checkstyle-config.xml:1]: done.

File: config/intellij-idea-code-style.xml [hunk 4]
```diff
@@ -0,0 +1,361 @@
+<code_scheme name="code-style" version="173">
```
romani, line 1: can we use there 1 or "snapshot" ?
we will not be able to udpate it each time we change this file.

or..... is it version of "code_scheme " ?

romani (changes requested):
item:

romani:
I tried again, same result:
```
$ git diff
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java b/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
index c7b8f09890..522d4124e9 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
@@ -1,4 +1,4 @@
-///////////////////////////////////////////////////////////////////////////////////////////////
+/// ////////////////////////////////////////////////////////////////////////////////////////////
 // checkstyle: Checks Java source code and other text files for adherence to a set of rules.
 // Copyright (C) 2001-2025 the original author or authors.
 //
@@ -255,6 +255,7 @@ public class CheckstyleAntTask extends Task {
 
     ////////////////////////////////////////////////////////////////////////////
     // The doers
+
     ////////////////////////////////////////////////////////////////////////////
 
```

please investigate IDEA CLI.

example of how we use it
https://github.com/checkstyle/checkstyle/blob/d4025b26159e5e386eaeb3abd4011bd990b39284/.circleci/config.yml#L25

https://github.com/checkstyle/checkstyle/blob/d4025b26159e5e386eaeb3abd4011bd990b39284/.ci/idea-inspection.sh#L46

please run it on our local from terminal.
Lets start usiing terminal.

config/intellij-idea-code-style.xml
Anmol202005: It refers to the internal version of the code style schema that IDEA is using.

Anmol202005:
```sh
C:\Users\anmol>"C:\Users\anmol\AppData\Local\JetBrains\IntelliJ IDEA 2024.2.4\bin\format.bat" -settings "C:\Users\anmol\IdeaProjects\checkstyle3\config\intellij-idea-code-style.xml" -optimize-imports-only "C:\Users\anmol\IdeaProjects\checkstyle3\src\main\java\com\puppycrawl\tools\checkstyle\ant\CheckstyleAntTask.java"
```

```sh
PS C:\Users\anmol\IdeaProjects\checkstyle3> git diff 
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java b/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
index c7b8f09890..034e3ec34d 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
@@ -124,7 +124,7 @@ public class CheckstyleAntTask extends Task {
      * is a violation.
      *
      * @param propertyName the name of the property to set
-     *                      in the event of a failure.
+     *     in the event of a failure.
      */
     public void setFailureProperty(String propertyName) {
         failureProperty = propertyName;
@@ -198,10 +198,10 @@ public class CheckstyleAntTask extends Task {
      * Creates classpath.
      *
      * @return a created path for locating classes
-     * @deprecated left in implementation until #12556 only to allow users to migrate to new gradle
-     *     plugins. This method will be removed in Checkstyle 11.x.x .
      * @noinspection DeprecatedIsStillUsed
      * @noinspectionreason DeprecatedIsStillUsed - until #12556
+     * @deprecated left in implementation until #12556 only to allow users to migrate to new gradle
+     *     plugins. This method will be removed in Checkstyle 11.x.x .
      */
```
Looks like it mainly adjusted indentation and spacing—I don't see any significant changes in the diff report.

romani:
not working as expected on my side:
```
✔ ~/java/github/romani/checkstyle [Anmol202005/AutoConfig L|⚑ 2] 
$ /home/roman/java/intelij-idea/idea-IU-243.22562.218/bin/format.sh -settings config/intellij-idea-code-style.xml src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
[0.017s][warning][cds] Archived non-system classes are disabled because the java.system.class.loader property is specified (value = "com.intellij.util.lang.PathClassLoader"). To use archived non-system classes, this property must not be set
2025-03-12 06:15:08,711 [    730]   WARN - c.i.o.u.r.overrides - Conflicting registry key definition for key kotlin.mpp.tests.force.gradle: it was defined by plugin org.jetbrains.kotlin but redefined by plugin org.jetbrains.kotlin. Consider adding overrides="true" for one of the plugins, see the documentation for com.intellij.openapi.util.registry.RegistryKeyBean.overrides for more details.
2025-03-12 06:15:08,716 [    735]   WARN - c.i.o.u.r.overrides - Conflicting registry key definition for key domains.spring.component.by.name.endpoint: it was defined by plugin com.intellij.spring but redefined by plugin com.intellij.spring.data. Consider adding overrides="true" for one of the plugins, see the documentation for com.intellij.openapi.util.registry.RegistryKeyBean.overrides for more details.
2025-03-12 06:15:08,934 [    953]   WARN - #c.i.u.p.CommonProxy - main authenticator is not yet registered
IntelliJ IDEA 2024.3.1.1, build IU-243.22562.218 Formatter
2025-03-12 06:15:12,254 [   4273]   WARN - #c.i.s.ComponentManagerImpl - `preload=TRUE` must be used only for core services (service=org.jetbrains.idea.maven.project.MavenProjectsManagerEx, plugin=org.jetbrains.idea.maven)
2025-03-12 06:15:12,254 [   4273]   WARN - #c.i.s.ComponentManagerImpl - `preload=TRUE` must be used only for core services (service=org.jetbrains.idea.maven.navigator.MavenProjectsNavigator, plugin=org.jetbrains.idea.maven)
2025-03-12 06:15:12,255 [   4274]   WARN - #c.i.s.ComponentManagerImpl - `preload=TRUE` must be used only for core services (service=org.jetbrains.idea.maven.tasks.MavenShortcutsManager, plugin=org.jetbrains.idea.maven)
2025-03-12 06:15:12,255 [   4274]   WARN - #c.i.s.ComponentManagerImpl - `preload=TRUE` must be used only for core services (service=com.intellij.cwm.plugin.users.BackendUserManager, plugin=com.jetbrains.codeWithMe)
2025-03-12 06:15:12,255 [   4274]   WARN - #c.i.s.ComponentManagerImpl - `preload=TRUE` must be used only for core services (service=com.intellij.cwm.plugin.ports.CwmPortForwardingToolWindowManager, plugin=com.jetbrains.codeWithMe)
2025-03-12 06:15:12,255 [   4274]   WARN - #c.i.s.ComponentManagerImpl - `preload=TRUE` must be used only for core services (service=com.intellij.cwm.plugin.following.FollowMeManagerService, plugin=com.jetbrains.codeWithMe)
Formatting /home/roman/java/github/romani/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java...2025-03-12 06:15:13,632 [   5651]   WARN - #c.i.o.e.c.i.EditorColorsManagerImpl - Dark color scheme is missing
OK

1 file(s) scanned.
1 file(s) formatted.


✔ ~/java/github/romani/checkstyle [Anmol202005/AutoConfig L|✚ 1⚑ 2] 
$ git diff
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java b/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
index c7b8f09890..522d4124e9 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
@@ -1,4 +1,4 @@
-///////////////////////////////////////////////////////////////////////////////////////////////
+/// ////////////////////////////////////////////////////////////////////////////////////////////
 // checkstyle: Checks Java source code and other text files for adherence to a set of rules.
 // Copyright (C) 2001-2025 the original author or authors.
 //
@@ -255,6 +255,7 @@ public class CheckstyleAntTask extends Task {
 
     ////////////////////////////////////////////////////////////////////////////
     // The doers
+
     ////////////////////////////////////////////////////////////////////////////
 
     @Override
@@ -268,11 +269,11 @@ public class CheckstyleAntTask extends Task {
 
             // Check for no arguments
             if (fileName == null
-                    && fileSets.isEmpty()
-                    && paths.isEmpty()) {
+                && fileSets.isEmpty()
+                && paths.isEmpty()) {
                 throw new BuildException(
-                        "Must specify at least one of 'file' or nested 'fileset' or 'path'.",
-                        getLocation());
+                    "Must specify at least one of 'file' or nested 'fileset' or 'path'.",
+                    getLocation());
             }
             if (config == null) {
                 throw new BuildException("Must specify 'config'.", getLocation());
@@ -325,7 +326,7 @@ public class CheckstyleAntTask extends Task {
      *     or if the build failed due to violations.
      */
     private void processFiles(RootModule rootModule, final SeverityLevelCounter warningCounter,
-            final String checkstyleVersion) {
+        final String checkstyleVersion) {
         final long startTime = System.currentTimeMillis();
         final List<File> files = getFilesToCheck();
         final long endTime = System.currentTimeMillis();
@@ -333,9 +334,9 @@ public class CheckstyleAntTask extends Task {

```

romani:
plase try to use same version of as me.
try to not use `-optimize-imports-only`

romani:
@Anmol202005 , can we finish this PR ?

Anmol202005:
> @Anmol202005 , can we finish this PR ?

yep roman i tried it without `-optimize-imports-only` got the same result.

was wondering what must be causing this!!

Anmol202005:
except the first 2 changes i guess our diff report matches

Anmol202005:
@romani will get onto it shorty have been occupied by other issues lately.

Anmol202005:
@romani 
please use the updated config.
it should work (hopefully not like before) :)

romani:
@Anmol202005 , can you make it as CI job ? same as we have idea inspection execution

Anmol202005:
@romani is the config working now !

romani:
@Anmol202005 , can you make CI job for it ?

@Pankraz76 , can you help us validate if this config works for different IDEAs ?

config/intellij-idea-code-style.xml
```diff
@@ -0,0 +1,79 @@
...
+  <option name="OTHER_INDENT_OPTIONS">
+    <value>
+      <option name="INDENT_SIZE" value="4" />
+      <option name="TAB_SIZE" value="4" />
+    </value>
+  </option>
+  <JavaCodeStyleSettings>
+    <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
```
Pankraz76, line 9: #16574

its 1o1 the same key, so I assume we could migrate them into `.editorconfig`

image

Pankraz76:
related issue: #16574

config/intellij-idea-code-style.xml
Pankraz76: Thanks for the keys. It would be hard to find all these configurations through trial and error.

I'm wondering why the plugin didn't find any violations - this could mean either the code is fully compliant or there might be a bug. 

The current setup appears to be working effectively because:
1. Issues are being properly flagged by Checkstyle itself
2. The plugin adds a secondary validation layer
3. It offers the advantage of working out-of-the-box through EditorConfig
4. Any misalignments are still caught by the native Checkstyle verification

This redundancy is actually beneficial as it provides extra assurance without significant runtime overhead.

image

Pankraz76:
> @Pankraz76 , can you help us validate if this config works for different IDEAs ?

I would kindly ask @Anmol202005 to verify the .editorconfig PR works the same way.

To test:
1. Simply copy the root `.editorconfig` file
2. Verify the configuration aligns properly

The entire source code should now reformat correctly without requiring any manual changes.

Even though we have somehow misaligned code - we have continued level 4 like the normal, but some code has double the size (8 spaces or 2 tabs) but it's 2 indents not only one like other code.

Demo: https://github.com/checkstyle/checkstyle/pull/16855 This should not be possible as checkstyle should check indentation level as well.

Misaligned format:
image

After correction:
image

Anmol202005:
> @Anmol202005 , can you make CI job for it ? at checkstyle/.circleci/config.yml and may be `checkstyle/.ci/idea-format.sh` to simplify user experience to run it.

on it.

Anmol202005:
@Pankraz76 can u please brief me about the `.editorconfig`. if we are going with that or sticking with this config.

Pankraz76:
Thank you @Anmol202005 for providing the keys and the config. This is actually the hardest work to do as I don't know how to do it and this was kind of a headache for me as I expected trial and error and having a blindspot.
Considering the amount of keys they seem to be sufficient.

Thanks for the work done it will remain here or merge into the `.editorconfig`.

Anmol202005:
@romani @Pankraz76 Should I proceed creating a CI job for this configuration?

Pankraz76:
> we should let all GSOC participants to finish competition in next 2 weeks.

https://github.com/checkstyle/checkstyle/pull/16574#issuecomment-2817206510

Assuming its merged in the near future (max. 3 weeks) I would suggest to wait after merge.
Then see if its already done, as expected.

romani:
>Should I proceed creating a CI job for this configuration?

yes, please.

IDE agnostic update is not related at all, agnostic will never cover all formatting requirements, too much of them.

@Anmol202005 , you are doing good job for all IDEA users.

Anmol202005:
@romani I'm almost done with the CI implementation. But I'm confused with how will it help IDEA user's specially.
Can u please clear my understanding.

According to my understanding:

- When code is pushed, the CI job will run automatically.
- If after formatting any git diff is detected, the job will fail and show the files that need reformatting.

Pankraz76:
> * If after formatting any git diff is detected, the job will fail and show the files that need reformatting.

nice idea, yeah then we still need this. The job will fail and show if files need reformat.

But this is what the CI is doing with checkstyle, as this is the last instance and should by our single source of truth.
So it would fail the CI too. This means that this is just an additional layer to check. and providing the ability to configuration over convention which we want in invert.

So yes then this is still a valid point we could leverage.

The config-file is then only needed for the CI job and not for users to configure the IDE as `.editorconfig` will take care of it.

romani:
>According to my understanding:
    When code is pushed, the CI job will run automatically.
    If after formatting any git diff is detected, the job will fail and show the files that need reformatting.

All correct. In this case we will clearly know is this configuration is actually working, and we can easily disable temporary formatting if we found some conflicts, to enable it after extra fix.

.circleci/config.yml
```diff
@@ -2,6 +2,31 @@ version: 2.1
...
+      - checkout
+      - run:
+          name: Print versions
+          command: |
+            echo "Maven version:"
+            mvn --version
+            echo "Java version:"
+            java --version
+            echo "IDEA version:"
+            echo $IDEA_VERSION
```
Pankraz76, line 19:
```suggestion
            echo "IDEA version: $IDEA_VERSION"
```
is this possible?

.ci/idea-format.sh
```diff
@@ -0,0 +1,68 @@
...
+    CIRCLE_URL="https://app.circleci.com/pipelines/github/checkstyle/checkstyle"
+    # It seems like the value of this number does not matter when navigating to the artifacts tab,
+    # but it is required to be present in the URL.
+    SOME_NUMBER=1
+    echo "$CIRCLE_URL/$SOME_NUMBER/workflows/$CIRCLE_WORKFLOW_ID/jobs/$CIRCLE_BUILD_NUM/artifacts"
+    exit 1;
+elif [[ $MODIFIED_FILES -gt 0 ]]; then
+    echo "Some files were modified during formatting. Modified files:"
+    git diff --name-only
+    exit 1;
```
Pankraz76, line 65: so dirty code will break build, right?

.ci/idea-format.sh
```diff
@@ -0,0 +1,68 @@
...
+fi
+
+echo "Checking results ..."
+MODIFIED_FILES=$(find . -type f -name "*.java" -exec git diff --name-only {} \; | wc -l)
+
+if [[ $MODIFIED_FILES -gt 0 ]] && [[ "$CIRCLECI" == "true" ]]; then
+    echo ""
+    echo "ATTENTION GITHUB CONTRIBUTOR:"
+    echo "Some files were modified during formatting."
```
Pankraz76, line 53:
```suggestion
    echo "Some files need reformation."
```

.ci/idea-format.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# Execute compilation of Checkstyle to generate all source files
+# YOU MUST BUILD PROJECT BEFORE FORMAT EXECUTION!!!
+mvn -e --no-transfer-progress clean compile
+
+echo ""
+for i in {1..100}; do echo -n "#"; done
+echo ""
+
+echo "Intellij Idea code formatting is about to start"
```
Pankraz76, line 35:
```suggestion
echo "Starting Intellij Idea code formatting."
```

Pankraz76 (changes requested):
spelling

.circleci/config.yml
```diff
@@ -2,6 +2,31 @@ version: 2.1
 
 jobs:
 
+  run-formatter:
+    docker:
+      - image: checkstyle/idea-docker:jdk11-idea2022.3.3
```
romani, line 7: Usage of so old image might give us too much problem.
And it reminds me about our attempt to upgrade to recent idea version, but something blocked us to do so, and we lost contributor to continue.
Making it to work on old idea is not good, as behavior unlikely be the same with recent version, so users will be frustrated

romani (changes requested):
It becomes more complicated:

.circleci/config.yml
romani: We need to update idea first.

romani: https://github.com/checkstyle/checkstyle/pull/14696

Anmol202005: done.

Anmol202005:
@romani The CI- job is working fine now.

Anmol202005:
@romani Checkstyle follows this config:
```xml
<module name="Indentation">
      <property name="basicOffset" value="4"/>
      <property name="braceAdjustment" value="0"/>
      <property name="caseIndent" value="4"/>
      <property name="throwsIndent" value="8"/> 
</module>
```
Since throwsIndent has a value of 8. its not quite clear how to implement this via idea-style-config.
Idea does not have different keys for arrayinit, throwsIndent and lineWrappingIndentation.

romani:
awesome !!!

but we need green CI, so we need to find a way to not produce formatting that is conflicting with Indentation.
can we do anything to not produce diff or produce such diff that we can merge and checkstyle is not conflicting ?
we might stop using throwsIndent in our config if only it is a problem to reconcile Indentation Check and idea formatter.

Anmol202005:
> but we need green CI, so we need to find a way to not produce formatting that is conflicting with Indentation. can we do anything to not produce diff or produce such diff that we can merge and checkstyle is not conflicting ? we might stop using throwsIndent in our config if only it is a problem to reconcile Indentation Check and idea formatter.

IMO instead of git diff we can make another job to run checkstyle so that we can run it on the formatted code(if we change throwsIndent value to 4).

romani:
It will be ideal.
But I just tried to get something merge-able sooner and not grow scope of update.
But I think extra job might be quick to make, all all changes already applied to sources, please do

.circleci/config.yml
romani: @Anmol202005 , please apply this change

romani: hmm, this is liketo be be for consistency with  versions printed before

.ci/idea-format.sh
```diff
@@ -0,0 +1,58 @@
...
+"$IDEA_PATH/bin/idea.sh" format -r -s "$FORMAT_PATH" -m "*.java" "$PROJECT_DIR/src/it/java"
+"$IDEA_PATH/bin/idea.sh" format -r -s "$FORMAT_PATH" -m "*.java" "$PROJECT_DIR/src/main/java"
+"$IDEA_PATH/bin/idea.sh" format -r -s "$FORMAT_PATH" -m "*.java" "$PROJECT_DIR/src/test/java"
+"$IDEA_PATH/bin/idea.sh" format -r -s \
+         "$FORMAT_PATH" -m "*.java" "$PROJECT_DIR/src/xdocs-examples/java"
+
+# Run Checkstyle
+echo "Running Checkstyle check..."
+mvn checkstyle:check \
+  -Dcheckstyle.config.location="$PROJECT_DIR/config/checkstyle/checkstyle.xml" \
```
romani, line 50: please move it to new CircleCI step

```
      - run:
          name: Run code formatter
          command: |
            # Run the formatter
            ./.ci/idea-format.sh
          no_output_timeout: 15m
          working_directory: ~/project
      - run:
          name: Run checkstyle 
          command:  ./mvm clean verify
          working_directory: ~/project
```

.ci/idea-format.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# Prepare results directory
+if [ -d "$RESULTS_DIR" ]; then
+    rm -rf "$RESULTS_DIR"
+fi
+mkdir -p "$RESULTS_DIR"
+
+# Compile the project with Maven
+echo "Running Maven compile..."
+mvn -e --no-transfer-progress compile
```
romani, line 35: ./mvnw

romani (changes requested):
items:

Pankraz76:
consider maven solution:
- https://issues.apache.org/jira/browse/HBASE-26617
- https://github.com/diffplug/spotless

.circleci/config.yml
```diff
@@ -2,6 +2,37 @@ version: 2.1
...
+          command: |
+            # Run the formatter
+            ./.ci/idea-format.sh
+          no_output_timeout: 15m
+          working_directory: ~/project
+      - run:
+          name: Run checkstyle
+          command: |
+            echo "Running Checkstyle..."
+            ./mvnw -e --no-transfer-progress clean verify
```
romani, line 31: Please do only checkstyle execution.
https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations#how-to-generate-checkstyle-report
Only checkstyle part of verify cares about formatting

romani (changes requested):
Last

.circleci/config.yml
Anmol202005: done.

Anmol202005:
@romani just reduced the error to 28.
Not sure for indentation related !!

.ci/idea-format.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Run IDEA formatter
+echo "Running IDEA format from: $IDEA_PATH"
+echo "$IDEA_PATH format -r -s $FORMAT_PATH -m *.java $PROJECT_DIR"
+
+"$IDEA_PATH/bin/idea.sh" format -r -s "$FORMAT_PATH" -m "*.java" "$PROJECT_DIR/src/it/java"
+"$IDEA_PATH/bin/idea.sh" format -r -s "$FORMAT_PATH" -m "*.java" "$PROJECT_DIR/src/main/java"
+"$IDEA_PATH/bin/idea.sh" format -r -s "$FORMAT_PATH" -m "*.java" "$PROJECT_DIR/src/test/java"
+"$IDEA_PATH/bin/idea.sh" format -r -s \
+         "$FORMAT_PATH" -m "*.java" "$PROJECT_DIR/src/xdocs-examples/java"
```
Pankraz76, line 45:
```suggestion
"$IDEA_PATH/bin/idea.sh" format -r -s \ "$FORMAT_PATH" -m "*.java" "$PROJECT_DIR/src/xdocs-examples/java"
```
can this be done in one line, too? or is this a pipe command?

.circleci/config.yml
```diff
@@ -2,6 +2,41 @@ version: 2.1
...
+      - checkout
+      - run:
+          name: Print versions
+          command: |
+            echo "Maven version:"
+            mvn --version
+            echo "Java version:"
+            java --version
+            echo "IDEA version:"
+            echo $IDEA_VERSION
```
Pankraz76, line 19:
```suggestion
            echo "Maven version:" & mvn --version
            echo "Java version:" & java --version
            echo "IDEA version:" & echo $IDEA_VERSION
```
its one or two then its should turn out the same. But this might be not better.

.circleci/config.yml
```diff
@@ -2,6 +2,41 @@ version: 2.1
...
+      - run:
+          name: Run code formatter
+          command: |
+            # Run the formatter
```
Pankraz76, line 23:
```suggestion
            echo "Run the formatter"
```
might use or loose.

.ci/idea-format.bat
```diff
@@ -0,0 +1,55 @@
...
+::
+:: Example:
+:: SET IDEA_PATH=C:\Program Files\JetBrains\IntelliJ IDEA Community Edition 2017.2.1\bin\idea.bat
+:: .ci\idea-inspection.bat
+::----------------------------------------------------------------------
+
+SET PROJECT_DIR=%CD%
+SET FORMAT_PATH=%CD%\config\intellij-idea-code-style.xml
+SET RESULTS_DIR=%CD%\target\format-results
+SET IDEA_LOCATION=
```
Pankraz76, line 14:
```suggestion
SET IDEA_LOCATION
```
does it work without the equal sign `=` too? dont know the syntax sorry.
https://ss64.com/nt/set.html
image

Pankraz76 (commented):
polish

.ci/idea-format.sh
romani: it is line wrap to avoid CS violation on long lines

.circleci/config.yml
romani: @Anmol202005 , please just remove comment, we have it in name of run

.circleci/config.yml
```diff
@@ -2,6 +2,41 @@ version: 2.1
...
+          name: Updating suppressions-xpath.xml
+          command: |
+            ./.ci/idea-suppression-update.sh
+      - run:
+          name: Run checkstyle
+          command: |
+            echo "Running Checkstyle..."
+            ./mvnw -e --no-transfer-progress clean compile antrun:run@ant-phase-verify
+          working_directory: ~/project
+      - store_artifacts:
```
romani, line 37: lets run `git diff` as "run" and fail is any diff is present. Now it is hard for me to see what is a diff
We have this, that is used in other CIs.
https://github.com/checkstyle/checkstyle/blob/3e4e2377141bb5301f1f0f17f754ac60416753be/.ci/validation.sh#L1066

romani (changes requested):
item:

.ci/idea-format.bat
romani: lets keep `=`, it is more explicit

.circleci/config.yml
Anmol202005: then, it is always going to fail.

romani: we need same as at:
https://github.com/checkstyle/checkstyle/blob/3e4e2377141bb5301f1f0f17f754ac60416753be/.github/workflows/google-java-format.yml#L48-L49

if diff exists we failing a CI .
for IDEA formatter we also need to run checkstye to see its output and understand we can follow IDEA formatting.

so checkstyle vaildation failing CI, git-diff runs on failures also(always).

.ci/idea-suppression-update.sh
```diff
@@ -0,0 +1,27 @@
...
+# Remove closing tag and add new suppression with closing tag
+sed -i '/<\/suppressions>/d' "$file"
+cat >> "$file" << 'EOF'
+  <suppress-xpath
+    files=".*"
+    checks="WhitespaceAfterCheck"
+    query="//ARRAY_INIT/COMMA[following-sibling::*[1][self::RCURLY]]"/>
+  <suppress-xpath
+    files=".*"
+    checks="IndentationCheck"
```
romani, line 20: this is concerning, we wil discuss this

romani (changes requested):
items

.ci/idea-suppression-update.sh
Anmol202005: just to make ci green

Anmol202005: idea does not autoformat indentation and whitespaceAfter properly for array declaration.
if we are going to rely on git diff, it will be reverted.

Anmol202005:
@romani just need a little clarification on how to proceed.

romani:
Please do git diff to let me see a problem. 
Right now I don't know what idea changed at all.

.ci/idea-suppression-update.sh
romani: >idea does not autoformat indentation and whitespaceAfter properly for array declaration.

I need to see it, we can open issue on Idea to fix, if we see a problem. But we might need to change settings of checkstyle to not conflict with it, if they is not critical

Anmol202005:
> Please do git diff to let me see a problem. Right now I don't know what idea changed at all.

Done.

romani:
@Anmol202005 , please do this https://github.com/checkstyle/checkstyle/pull/16349/files#r2105819189

.circleci/config.yml
romani: please do diff after checkstyle execution. 
To let me see what is ok by checkstyle and what is not.

Diff likely always exists, but it is ok, user will just apply patch from formatter (CI output, same as we do with spelling) or run IDEA formatter.

But what will be show stopper is conflict with checkstyle after formatting.
So I and all users should see diff and checkstyle execution at the same time in CI.
I suggesting to run checkstyle first, let it fail a CI, and git diff execution is done `when: always` to let it run on_fail and passed.

Anmol202005:
> @Anmol202005 , please do this https://github.com/checkstyle/checkstyle/pull/16349/files#r2105819189

done.

romani:
@Anmol202005 , 

I need to get access to violations on WhitespaceAfterCheck and IndentationCheck .
to make decision is this defect of Checkstyle or we need some new changes on IDEA formatter side, or report just another issue for IDEA.

config/intellij-idea-code-style.xml
```diff
@@ -0,0 +1,82 @@
...
+    <option name="CLASS_ANNOTATION_WRAP" value="2" />
+    <option name="FIELD_ANNOTATION_WRAP" value="2" />
+    <option name="PARAMETER_ANNOTATION_WRAP" value="0" />
+    <option name="ENUM_CONSTANTS_WRAP" value="2" />
+    <option name="VARIABLE_ANNOTATION_WRAP" value="2" />
+    <indentOptions>
+      <option name="CONTINUATION_INDENT_SIZE" value="4" />
+    </indentOptions>
+  </codeStyleSettings>
+</code_scheme>
```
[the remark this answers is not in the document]
Pankraz76: - https://github.com/checkstyle/checkstyle/pull/17140

.ci/idea-format.sh
```diff
@@ -0,0 +1,44 @@
...
+    fi
+fi
+
+# Prepare results directory
+if [ -d "$RESULTS_DIR" ]; then
+    rm -rf "$RESULTS_DIR"
+fi
+mkdir -p "$RESULTS_DIR"
+
+# Compile the project with Maven
```
Pankraz76, line 33:
```suggestion
:: Compile the project with Maven
```
whats difference between # and :: ? assuming its both comments?

https://stackoverflow.com/a/43158193

.ci/idea-format.bat
```diff
@@ -0,0 +1,55 @@
...
+::Launch inspections
+echo Running IDEA format from: %IDEA_LOCATION%
+echo "%IDEA_LOCATION%" format -r -s "%FORMAT_PATH%" -m "*.java" "%PROJECT_DIR%"
+call "%IDEA_LOCATION%" format -r -s "%FORMAT_PATH%" -m "*.java" "%PROJECT_DIR%\src\it\java"
+call "%IDEA_LOCATION%" format -r -s "%FORMAT_PATH%" -m "*.java" "%PROJECT_DIR%\src\main\java"
+call "%IDEA_LOCATION%" format -r -s "%FORMAT_PATH%" -m "*.java" "%PROJECT_DIR%\src\test\java"
+call "%IDEA_LOCATION%" format -r -s "%FORMAT_PATH%" -m "*.java" ^
+    "%PROJECT_DIR%\src\xdocs-examples\java"
+
+echo Formatting completed.
```
Pankraz76, line 55:
```suggestion
echo Formatting complete.
```
try present not past. as all logs before have present time form as well

.ci/idea-format.bat
```diff
@@ -0,0 +1,55 @@
...
+
+:run
+IF NOT EXIST "%RESULTS_DIR%" mkdir "%RESULTS_DIR%"
+IF EXIST "%RESULTS_DIR%" del /s /q "%RESULTS_DIR%\."
+
+::Execute compilation of Checkstyle to generate all source files
+echo Running Maven compile...
+call mvn -e compile
+
+::Launch inspections
```
Pankraz76, line 46:
```suggestion
# Launch inspections
```

whats difference between # and :: ? assuming its both comments?

https://stackoverflow.com/a/43158193

.ci/idea-format.bat
```diff
@@ -0,0 +1,55 @@
+@echo off
+
+::----------------------------------------------------------------------
+:: IntelliJ IDEA inspections for checkstyle.
+::
+:: Example:
+:: SET IDEA_PATH=C:\Program Files\JetBrains\IntelliJ IDEA Community Edition 2017.2.1\bin\idea.bat
+:: .ci\idea-inspection.bat
+::----------------------------------------------------------------------
```
Pankraz76, line 9:
```suggestion
: '
This is a
very neat comment
in bash

----------------------------------------------------------------------
 IntelliJ IDEA inspections for checkstyle.

:: Example:
 SET IDEA_PATH=C:\Program Files\JetBrains\IntelliJ IDEA Community Edition 2017.2.1\bin\idea.bat
 .ci\idea-inspection.bat
----------------------------------------------------------------------
'
```
maybe use multi line https://stackoverflow.com/questions/43158140/way-to-create-multiline-comments-in-bash

or at least consistent `#` or `::`

idk why mixed.

Pankraz76 (commented):
polish.

Pankraz76:
its again very related. If we have eclipse config we can simply automate the checkstyle issues.

- https://github.com/checkstyle/checkstyle/pull/17490

Pankraz76:
spotless has recently gained idea support, might apply this via spot instead of having the custom version. Assuming you also just call the CLI as the spot plugin does.

- https://github.com/diffplug/spotless/tree/main/plugin-maven#intellij-idea

.circleci/config.yml
```diff
@@ -2,6 +2,59 @@ version: 2.1
 
 jobs:
 
+  run-formatter:
+    docker:
+      - image: checkstyle/idea-docker:jdk11-idea2023.3.4
```
romani, line 7: we migrated to qodana (official way to run inspections)
may it has formatting also

PR #16367, opened 2025-02-19
Add project icon

punkratz312:
checkstyle is iconic as well thanks for the project.

same as here: https://github.com/spring-projects/spring-framework/pull/34444

image

If you have a better icon or svg let me know.

Thanks


.idea/icon.png
romani, line 1: we do not store any IDEA specific files.

romani (changes requested):
we do not do preference to any IDE.

punkratz312:
> we do not do preference to any IDE.

I would call it support. Others might support this feature as well.

PR #16380, opened 2025-02-20
Issue #14019: Resolve pitest suppressions for getId in TreeWalker

aboro-code:
Pr for issue #14019 

https://github.com/checkstyle/checkstyle/blob/273af85b29a3df7c0313430d9e730584eb847c1b/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java#L483C45-L488

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
@@ -62,6 +62,7 @@ import com.puppycrawl.tools.checkstyle.api.TokenTypes;
 import com.puppycrawl.tools.checkstyle.checks.NoCodeInFileCheck;
 import com.puppycrawl.tools.checkstyle.checks.coding.EmptyStatementCheck;
 import com.puppycrawl.tools.checkstyle.checks.coding.HiddenFieldCheck;
+import com.puppycrawl.tools.checkstyle.checks.coding.IllegalCatchCheck;
 import com.puppycrawl.tools.checkstyle.checks.design.OneTopLevelClassCheck;
 import com.puppycrawl.tools.checkstyle.checks.indentation.CommentsIndentationCheck;
 import com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocPackageCheck;
@@ -792,6 +793,29 @@ public class TreeWalkerTest extends AbstractModuleTestSupport {
                         + getNonCompilablePath("InputTreeWalkerSkipParsingException.java") + "."));
 
         verify(checker, files, expectedViolation);
+    }
+
+    /**
+     * Verifies that TreeWalker sorting works correctly using AbstractCheck::getId.
+     * Multiple checks are configured with unique IDs to ensure all are executed and
+     * violations are reported in a sorted order.
+     */
+    @Test
+    public void testCheckSortingByIdWithVerify() throws Exception {
+        final String[] expected = {
+            "26:30: " + getCheckMessage(ConstantNameCheck.class, "name.invalidPattern",
+                "someConstant", "^[A-Z][a-zA-Z0-9]*$"),
+
+            "29:31: " + getCheckMessage(ParameterNameCheck.class, "name.invalidPattern",
+                "BADparam", "^[a-z][a-zA-Z0-9]*$"),
+
+            "36:9: " + getCheckMessage(WhitespaceAfterCheck.class, "ws.notFollowed", "catch"),
+
+            "36:9: " + getCheckMessage(IllegalCatchCheck.class, "illegal.catch", "Exception"),
+        };
+
+        final String path = getPath("InputTreeWalkerSorting.java");
+        verifyWithInlineXmlConfig(path, expected);
     }
 
     public static class BadJavaDocCheck extends AbstractCheck {
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSorting.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSorting.java
@@ -1,0 +1,40 @@
+/*xml
+<module name="Checker">
+  <module name="TreeWalker">
+    <module name="IllegalCatch">
+      <property name="id" value="id2"/>
+    </module>
+    <module name="ConstantName">
+      <property name="id" value="id1"/>
+      <property name="format" value="^[A-Z][a-zA-Z0-9]*$"/>
+    </module>
+    <module name="WhitespaceAfter">
+      <property name="id" value="id4"/>
+    </module>
+    <module name="ParameterName">
+      <property name="id" value="id3"/>
+      <property name="format" value="^[a-z][a-zA-Z0-9]*$"/>
+    </module>
+  </module>
+</module>
+*/
+package com.puppycrawl.tools.checkstyle.treewalker;
+
+public class InputTreeWalkerSorting {
+
+    // violation below, 'Name 'someConstant' must match pattern'
+    private static final int someConstant = 42;
+
+    // violation below, 'Name 'BADparam' must match pattern'
+    public void testCatch(int BADparam) {
+
+        try {
+            int x = 1 / 0;
+        }
+
+        // violation below, 'Catching 'Exception' is not allowed'
+        catch(Exception e) {// violation, ''catch' is not followed by whitespace'
+            System.out.println("Caught an exception");
+        }
+    }
+}
--- a/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
+++ b/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
@@ -13,15 +13,6 @@
     <sourceFile>TreeWalker.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.TreeWalker</mutatedClass>
     <mutatedMethod>createNewCheckSortedSet</mutatedMethod>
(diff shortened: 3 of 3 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
```diff
@@ -794,6 +797,81 @@ public void testJavaParseExceptionSeverityDefaultError() throws Exception {
...
+
+        Assertions.assertEquals("check1", checks.get(0).getId(),
+                "First element should be 'check1'");
+        Assertions.assertEquals("check2", checks.get(1).getId(),
+                "Second element should be 'check2'");
+        Assertions.assertEquals("check3", checks.get(2).getId(),
+                "Third element should be 'check3'");
+    }
+
+    public static class TestCheck extends AbstractCheck {
```
romani, line 820: Please kill mutation by existing Check.
If not possible, we will need long explanation over this class on why it is required.

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
romani: violations might come to result report in special order duet to this sorting, try to make such config in Input file where missed sorting by ID makes order or violation diffrent, so you can catch it test assert.

romani:
Please reuse existing check or explain why non them helps.

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
```diff
@@ -794,6 +797,31 @@ public void testJavaParseExceptionSeverityDefaultError() throws Exception {
...
+    @Test
+    public void testTreeWalkerCheckSortingById() {
+        final AbstractCheck check1 = new ConstantNameCheck();
+        check1.setId("zebra");
+
+        final AbstractCheck check2 = new ConstantNameCheck();
+        check2.setId("apple");
+
+        final AbstractCheck check3 = new ConstantNameCheck();
+        check3.setId("mango");
```
romani, line 809: Please add modules to header comment of InputTreeWalkerSorting
.
Please review how our general test are working, the same verifyxxxxx method should be used

romani (changes requested):
We need test coverage by real Checks

src/test/resources/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSorting.java
```diff
@@ -0,0 +1,17 @@
+/*
+com.puppycrawl.tools.checkstyle.checks.coding.IllegalCatchCheck
+
```
romani, line 3: Good step forward.
And please add few more modules in header.
And use id property in for them.

You can lot at filters Inputs for samples on how to use multiple modules in same time

romani (changes requested):
Items

src/test/resources/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSorting.java
aboro-code: i made the changes and added 3 more checks i hope that is what you meant.

romani: By one step closer.
Please consider https://github.com/checkstyle/checkstyle/blob/master/src/test/resources/com/puppycrawl/tools/checkstyle/filters/suppressioncommentfilter/InputSuppressionCommentFilter.java

And find usage of this Input to see how test method should be written.
Your implementation will not be the same, as you have no Filter, but we need to use another verifyXxxxxx method that is good for just verifying Checks execution.
You can find more examples of multiple Checks execution at the same time.

aboro-code: @romani could you please take a look? and let me know what more needs to be changed.

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
```diff
@@ -794,6 +795,51 @@ public void testJavaParseExceptionSeverityDefaultError() throws Exception {
...
+            "25:31: " + getCheckMessage(ParameterNameCheck.class, "name.invalidPattern",
+                "BADparam", "^[a-z][a-zA-Z0-9]*$"),
+
+            "33:9: " + getCheckMessage(IllegalCatchCheck.class, "illegal.catch", "Exception"),
+
+            "33:9: " + getCheckMessage(WhitespaceAfterCheck.class, "ws.notFollowed", "catch"),
+
+        };
+
+        verify(checker, getPath("InputTreeWalkerSorting.java"), expected);
```
romani, line 840: please use verifyWithInlineXmlConfig

romani (changes requested):
very close, one more improvement is requireed:

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
aboro-code: `com.puppycrawl.tools.checkstyle.api.CheckstyleException: Config cannot be parsed as xml.`
this is the error im getting now, could you guide me further on how i should approach the config of the input file. or could you clarify more on why exactly we should use verifyWithInlineXmlConfig. and not the standard verify or some other verifyXX.

aboro-code: I tried switching to `verifyWithInlineXmlConfig()` but ran into an issue — the expected array comes out empty even though it's populated correctly. I also updated the inline XML config in the input file and it's recognized fine, but still no luck with the violations being picked up.

It was working earlier with the regular` verify()` method, so just wondering — is there a specific reason to prefer `verifyWithInlineXmlConfig()` here? And if so, could you guide me a bit on how to structure it properly so the expected output works?

romani: >is there a specific reason to prefer verifyWithInlineXmlConfig()

Xml is used in all examples. As xml is only allowed from of confirmation for now.
Test inputs also use property format, you can use any of them. But main idea is that config should be in input file, violations should be in input file.

aboro-code: oh okay got it, pushed working version of code, no errors now, should be done

aboro-code: @romani could you take a look

src/test/java/com/puppycrawl/tools/checkstyle/TreeWalkerTest.java
```diff
@@ -794,6 +795,49 @@ public void testJavaParseExceptionSeverityDefaultError() throws Exception {
...
 
+    /**
+     * Verifies that TreeWalker sorting works correctly using AbstractCheck::getId.
+     * Multiple checks are configured with unique IDs to ensure all are executed and
+     * violations are reported in a sorted order.
+     */
+    @Test
+    public void testCheckSortingByIdWithVerify() throws Exception {
+        final DefaultConfiguration illegalCatchCheck = createModuleConfig(IllegalCatchCheck.class);
+        illegalCatchCheck.addAttribute("id", "id1");
```
romani, line 806: you should not have this configs, as all ofthem should be loaded from Input file

romani (changes requested):
direction is good, you are very close


items:

src/test/resources/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSorting.java
```diff
@@ -0,0 +1,40 @@
+/*xml
+<module name="Checker">
+  <module name="TreeWalker">
+    <module name="IllegalCatch">
+      <property name="id" value="id1"/>
+    </module>
```
romani, line 6: Make id to be in random order in config, but order of violations in ordered

romani (changes requested):
Items

src/test/resources/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSorting.java
aboro-code: done

romani: Please run pitest on your local to investigate why it is not helpful to kill mutation

aboro-code: Got two surviving mutations:

TreeWalker.java (line 486) – createNewCheckSortedSet(), replaced Comparator::thenComparing with receiver (823 tests run).

DetailNodeTreeStringPrinter.java (line 172) – parseFile(), replaced File::getAbsoluteFile with receiver (8 tests run).

Should I focus on adding tests to kill these, or is modifying the source files an option? Let me know how to proceed.

romani: Modification of sources is an option but you will need to defend changes with good reasons.

Ideal scenario is killing by test, no code changes.

romani:
@aboro-code, do you have energy to finish this PR?

romani:
abandoned

Issue #16081, opened 2024-12-29
PARAMETER_DEF in FinalLocalVariable doesn't seem to work in interfaces


Machine-Maker:
https://checkstyle.org/checks/coding/finallocalvariable.html#FinalLocalVariable

```bash
/var/tmp $ javac test.java
# success

/var/tmp $ cat checkstyle.xml
<!DOCTYPE module PUBLIC
    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
    "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
  <module name="TreeWalker">
    <module name="FinalLocalVariable">
      <property name="tokens" value="VARIABLE_DEF,PARAMETER_DEF"/>
    </module>
  </module>
</module>

/var/tmp $ cat YOUR_FILE.java
interface Test {
  static void test(String param) {
    String local = "";
    System.out.println(param);
  }
}

/var/tmp $ java -jar checkstyle-10.21.0-all.jar -c checkstyle.xml test.java
Starting audit...
[ERROR] /Users/jake/Projects/Minecraft/PaperMC/test.java:3:12: Variable 'local' should be declared final. [FinalLocalVariable]
Audit done.
Checkstyle ends with 1 errors.
```

---

The `param` local variable in the static method inside the Test interface should be reported as needing the `final` modifier which, if I'm understanding the documentation correctly, adding the `PARAMETER_DEF` token to the `FinalLocalVariable` check should do.

If I change `Test` to a `class` instead of `interface`, it correctly reports both, so I'm guessing someone just forgot about methods inside of interfaces.

romani:
Yes, code in interface is recently new feature of java :)

mahfouz72:
Just to make it clear abstract methods in the interface still should be ignored, and only the `default` and `static` methods (methods that can have a body) should be checked. Methods with bodies (static, default or private) were introduced in Java 8 and 9.

```Java
interface test {
    void test(int a);   // ok, this is an abstract method we should not violate parameter 'a'

    default void test2(int b) { // violation, b should be final
            // body
    }

    static void test3(int c) { // violation, c should be final
         // body
    }
}
```

romani:
Fix is merged

PR #16132, opened 2025-01-08
Issue #16081: FinalLocalVariable should check interface parameters

Machine-Maker:
Fixes https://github.com/checkstyle/checkstyle/issues/16081

PR #16391, opened 2025-02-21
Issue #16081: Resolved FinalLocalVariable Bug

Anmol202005:
Issue #16081: Resolved FinalLocalVariable bug ,
### Summary of Changes
- Fixed the issue where PARAMETER_DEF in FinalLocalVariable check was not working as expected in interfaces.
- Updated logic to correctly handle parameters inside interface methods.
- Added necessary test cases to verify correct behavior.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
@@ -219,6 +219,7 @@ public class FinalLocalVariableCheck extends AbstractCheck {
     // expressions to separate methods, but that will not increase readability.
     @Override
     public void visitToken(DetailAST ast) {
+
         switch (ast.getType()) {
             case TokenTypes.OBJBLOCK:
             case TokenTypes.METHOD_DEF:
@@ -239,8 +240,7 @@ public class FinalLocalVariableCheck extends AbstractCheck {
                 if (!isInLambda(ast)
                         && ast.findFirstToken(TokenTypes.MODIFIERS)
                             .findFirstToken(TokenTypes.FINAL) == null
-                        && !isInAbstractOrNativeMethod(ast)
-                        && !ScopeUtil.isInInterfaceBlock(ast)
+                        && !isInMethodWithoutBody(ast)
                         && !isMultipleTypeCatch(ast)
                         && !CheckUtil.isReceiverParameter(ast)) {
                     insertParameter(ast);
@@ -678,24 +678,14 @@ public class FinalLocalVariableCheck extends AbstractCheck {
     }
 
     /**
-     * Determines whether an AST is a descendant of an abstract or native method.
+     * Checks if a parameter is within a method that has no implementation body.
      *
-     * @param ast the AST to check.
-     * @return true if ast is a descendant of an abstract or native method.
+     * @param parameterDefAst the AST node representing the parameter definition
+     * @return true if the parameter is in a method without a body
      */
-    private static boolean isInAbstractOrNativeMethod(DetailAST ast) {
-        boolean abstractOrNative = false;
-        DetailAST currentAst = ast;
-        while (currentAst != null && !abstractOrNative) {
-            if (currentAst.getType() == TokenTypes.METHOD_DEF) {
-                final DetailAST modifiers =
-                    currentAst.findFirstToken(TokenTypes.MODIFIERS);
-                abstractOrNative = modifiers.findFirstToken(TokenTypes.ABSTRACT) != null
-                        || modifiers.findFirstToken(TokenTypes.LITERAL_NATIVE) != null;
-            }
-            currentAst = currentAst.getParent();
-        }
-        return abstractOrNative;
+    private static boolean isInMethodWithoutBody(DetailAST parameterDefAst) {
+        final DetailAST methodDefAst = parameterDefAst.getParent().getParent();
+        return methodDefAst.findFirstToken(TokenTypes.SLIST) == null;
     }
 
     /**
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/finallocalvariable/InputFinalLocalVariableInterface.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/finallocalvariable/InputFinalLocalVariableInterface.java
@@ -1,0 +1,32 @@
+/*
+FinalLocalVariable
+validateEnhancedForLoopVariable = (default)false
+tokens = VARIABLE_DEF, PARAMETER_DEF
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.finallocalvariable;
+
+public class InputFinalLocalVariableInterface {
+    interface Test {
+        static void test (String param) { // violation, "Variable 'param' should be declared final"
+            String local = ""; // violation, "Variable 'local' should be declared final"
+            System.out.println (param);
+        }
+
+        void method (int aParam);
+
+        default void init (int aParam) {} // violation, "Variable 'aParam' should be declared final"
+
+        // violation below, "Variable 'num' should be declared final"
+        static int parseInteger(String num) {
+            int result = 0;
+            try {
+                result = Integer.parseInt(num);
+            }
+            catch (NumberFormatException e) {} // violation, "Variable 'e' should be declared final"
+            return result;
+        }
+    }
+}
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/finallocalvariable/InputFinalLocalVariableNameLambda.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/finallocalvariable/InputFinalLocalVariableNameLambda.java
@@ -9,7 +9,7 @@ tokens = PARAMETER_DEF,VARIABLE_DEF
 package com.puppycrawl.tools.checkstyle.checks.coding.finallocalvariable;
 
 import java.math.BigDecimal;
-
+import java.util.function.Function;
 
 public class InputFinalLocalVariableNameLambda {
     private interface Lambda {
@@ -32,6 +32,9 @@ public class InputFinalLocalVariableNameLambda {
             .reduce(BigDecimal.ZERO,
                     (t, u) -> t.add(u.getAmount()));
  }
+    public static void main(final String[] args) {
+        final Function<Integer, Integer> doubleValue = (x) -> { return x * 2; };
+    }
 }
 interface Operation {
     public Object apply();
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheckTest.java
@@ -225,7 +225,7 @@ public class FinalLocalVariableCheckTest
     public void testLambda()
             throws Exception {
         final String[] expected = {
-            "40:16: " + getCheckMessage(MSG_KEY, "result"),
+            "43:16: " + getCheckMessage(MSG_KEY, "result"),
         };
         verifyWithInlineConfigParser(
                 getPath("InputFinalLocalVariableNameLambda.java"),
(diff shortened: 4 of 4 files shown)


Anmol202005:
Github, generate report for FinalLocalVariable/all-examples-in-one

Anmol202005:
@romani please review !!

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
```diff
@@ -240,7 +241,6 @@ public void visitToken(DetailAST ast) {
                         && ast.findFirstToken(TokenTypes.MODIFIERS)
                             .findFirstToken(TokenTypes.FINAL) == null
                         && !isInAbstractOrNativeMethod(ast)
-                        && !ScopeUtil.isInInterfaceBlock(ast)
```
mahfouz72, line 243: We still should ignore methods that have no body in interfaces

Please see : https://github.com/checkstyle/checkstyle/issues/16081#issuecomment-2564696835

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/finallocalvariable/InputFinalLocalVariable2Two.java
```diff
@@ -48,7 +48,7 @@ class InnerClass
 
 interface Inter2
 {
-    void method(int aParam);
+    void method(int aParam); // violation, "Variable 'aParam' should be declared final"
```
mahfouz72, line 51: It should be ok.

mahfouz72 (changes requested):
Items

romani:
@Anmol202005 , please address pitest.

you run it on local `./.ci/pitest.sh "pitest-coding-2"`.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
romani: yes, look at regression https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/524c0ee76e7404b4e248142090a6f51e763ed72a_2025204522/reports/diff/android-launcher/index.html#A1

it now forces to put violations on regular interfaces, it is impossible to fix, so such new behavior will force users to disable Check, so it will kill this Check.

plese put such regular interface in Inputs, to make sure it has no violations

romani:
we have concurrent PRs on same issue: https://github.com/checkstyle/checkstyle/pull/16132

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/finallocalvariable/InputFinalLocalVariable2Two.java
Anmol202005: done 👍

Anmol202005:
@romani please review !

romani:
@Anmol202005 , 
Please address https://github.com/checkstyle/checkstyle/pull/16391/files#r1966524773

Regression should not be massive.
Please add in Inputs regular interface with methods , no violations are expected.

As you done, please generate regression diff report.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
Anmol202005: totally agree !!

changes done👍

Anmol202005:
Github, generate report for FinalLocalVariable/all-examples-in-one

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/finallocalvariable/InputFinalLocalVariableInterface.java
```diff
@@ -0,0 +1,18 @@
...
+package com.puppycrawl.tools.checkstyle.checks.coding.finallocalvariable;
+
+public class InputFinalLocalVariableInterface {
+    interface Test {
+        static void test(String param) { // violation, "Variable 'param' should be declared final"
+            String local = ""; // violation, "Variable 'local' should be declared final"
+            System.out.println(param);
+        }
+        void method(int aParam);
+    }
```
romani, line 17: Please add this edge case https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/77def31a2af7b428bac25edf676ef99a460cbc30_2025214928/reports/diff/Orekit/index.html#A1
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/77def31a2af7b428bac25edf676ef99a460cbc30_2025214928/reports/diff/Orekit/index.html#A14
Violation is expected, as it has body, even it empty.

romani: Please add this case, exception parameters in implementation 
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/77def31a2af7b428bac25edf676ef99a460cbc30_2025214928/reports/diff/spoon/index.html#A52

One more time all that we see in diff report, we should se in inputs. In input we have 1 case but un diff report it can be numerous.

romani (changes requested):
Last:

romani:
@Machine-Maker, please join in review this PR , I think we will come to merge state in this PR quicker.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/finallocalvariable/InputFinalLocalVariableInterface.java
Anmol202005: done👍

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
```diff
@@ -678,24 +678,14 @@ private static boolean isVariableInForInit(DetailAST variableDef) {
     }
 
     /**
-     * Determines whether an AST is a descendant of an abstract or native method.
+     * Checks if a parameter is within a method that has no implementation body.
      *
-     * @param ast the AST to check.
-     * @return true if ast is a descendant of an abstract or native method.
-     */
-    private static boolean isInAbstractOrNativeMethod(DetailAST ast) {
```
romani, line 686: why we removed this method usage ?
it has nothing in common with interface methods.
Am I miss something ?

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
Anmol202005: Removed `isInAbstractOrNativeMethod` since it's redundant with `isInMethodWithoutBody`. 

- `isInMethodWithoutBody` checks if a method has no body (no SLIST token)
- Methods without bodies in Java can only be:
  1. Abstract methods
  2. Native methods
  3. Interface methods (prior to Java 8)

Therefore, `isInMethodWithoutBody` already covers all cases that `isInAbstractOrNativeMethod` was checking for. When `isInMethodWithoutBody` returns true, it implicitly means the method is either abstract or native, making the separate check unnecessary.

romani (approved):
ok to merge.


thanks a lot for hard work!!!

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
```diff
@@ -678,24 +678,14 @@ private static boolean isVariableInForInit(DetailAST variableDef) {
...
-                abstractOrNative = modifiers.findFirstToken(TokenTypes.ABSTRACT) != null
-                        || modifiers.findFirstToken(TokenTypes.LITERAL_NATIVE) != null;
-            }
-            currentAst = currentAst.getParent();
-        }
-        return abstractOrNative;
+     * @param ast the AST node to check
+     * @return true if the parameter is in a method without a body
+     */
+    private static boolean isInMethodWithoutBody(DetailAST ast) {
```
mahfouz72, line 686: `ast` => `parameterDefAst` and update Javadoc accordingly. This will help method readers to expect what is being passed in the parameters without referring to the caller code

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
```diff
@@ -678,24 +678,14 @@ private static boolean isVariableInForInit(DetailAST variableDef) {
...
+        final DetailAST parent = ast.getParent().getParent();
+        return parent.findFirstToken(TokenTypes.SLIST) == null;
```
mahfouz72, line 688: Good, This simplifies the logic to not deal with specific cases (native, abstract, ..) like we did before.
But why we didn't check that the parent is indeed a `methodDef` before checking for `SLIST`, if ast is a catch parameter then the parent will be `LITERAL_TRY`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
```diff
@@ -678,24 +678,14 @@ private static boolean isVariableInForInit(DetailAST variableDef) {
...
-                        || modifiers.findFirstToken(TokenTypes.LITERAL_NATIVE) != null;
-            }
-            currentAst = currentAst.getParent();
-        }
-        return abstractOrNative;
+     * @param ast the AST node to check
+     * @return true if the parameter is in a method without a body
+     */
+    private static boolean isInMethodWithoutBody(DetailAST ast) {
+        final DetailAST parent = ast.getParent().getParent();
```
mahfouz72, line 687: `parent` => `methodDefAst`

mahfouz72 (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
mahfouz72: > why we didn't check that the parent is indeed a methodDef before checking for SLIST

mmm, I think this check will be a surviving mutation because if the parameter is a catch parameter not a method parameter then this method will always return false. Please confirm this

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/finallocalvariable/InputFinalLocalVariableNameLambda.java
```diff
@@ -32,6 +32,9 @@ private void addTotalValueOfOrder(final AugmentedOrder order) {
             .reduce(BigDecimal.ZERO,
                     (t, u) -> t.add(u.getAmount()));
  }
+    public static void main(final String[] args) {
+        final Function<Integer, Integer> doubleValue = (x) -> { return x * 2; };
+    }
```
mahfouz72, line 37: Why do we extend input with this code? I don't think it is related

mahfouz72 (commented):
item

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/finallocalvariable/InputFinalLocalVariableNameLambda.java
romani: We kind of always appreciate test extension, even it is not very minimal. We prefer more tests rather than minimal required testing.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
Anmol202005: yep, that's the case

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
Anmol202005: done👍

mahfouz72 (approved):
Thanks a lot!

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/finallocalvariable/InputFinalLocalVariableNameLambda.java
mahfouz72: Agreed. but my question was whether this is related to the current functional changes. if not it should not be in this commit. if we really need to extend inputs then it should be a separate comment or so. Your call, I am ok with either way not a big deal.

PR #16408, opened 2025-02-23
Issue #14019: Resolve pitest suppressions for  SuppressionsStringPrinter

zyadahmed:
Issue #14019
#### Removing Redundant `getAbsoluteFile()` Call  

The call to `file.getAbsoluteFile()` is unnecessary because the `File` class can handle both relative and absolute paths when reading files. Since we only need to read the file, there is no requirement to explicitly convert it to an absolute path. Additionally, calling `getAbsoluteFile()` creates a new `File` instance that points to the same file, making it redundant and adding unnecessary overhead. 
This is also done in CLI options with the `printFileAst` method, which handles files in the same structure without using getAbsoluteFile().

https://github.com/checkstyle/checkstyle/blob/20284e326cdff304f569b4ff0ebca3a4ddf511de/src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java#L137

### DIFF
--- a/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
+++ b/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
@@ -55,15 +55,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>SuppressionsStringPrinter.java</sourceFile>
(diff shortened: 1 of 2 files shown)


romani (approved):
Yes, you are right

Issue #14697, opened 2024-03-20
False negative in UnusedImportCheck


mahfouz72:
detected at https://github.com/checkstyle/checkstyle/pull/14548#discussion_r1529498536 and https://github.com/checkstyle/checkstyle/pull/14548#discussion_r1530815684
I have read check documentation: https://checkstyle.org/checks/xxxxxx/nameofaffectedcheck.html
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words
```

PS D:\test> cat src/Test.java                                                  


import java.util.List; // expected violation here but there is no violation

/**
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

 <module name="TreeWalker">

      <module name="UnusedImports"/>
    </module>
</module>
PS D:\test> java  -jar checkstyle-10.13.0-all.jar -c config.xml  src/Test2.java
Starting audit...
Audit done.
```

**Describe what you expect in detail.**

This is an invalid Link because there is no `{  }` so the check should Ignore this and It shouldn't be considered as usage So we should have a violation on `import java.util.List;`

romani:
Please fix a link to check documentation in your PR description

romani:
@zyadahmed , please help us to resolve this issue.

PR #16423, opened 2025-02-25
Issue #14697: Resolved False negative in UnusedImportCheck

zyadahmed:
#14697 

The issue was  `@link` tag being considered valid. The root cause lies in the **`BLOCK_TAG_PATTERN`**, which retrieves all regex expressions containing `@` within Javadoc comments. This pattern incorrectly identifies `@link` as an valid reference for imports.

https://github.com/checkstyle/checkstyle/blob/1fa0377d6892f12e544e7c56acec4cf38c23d15e/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/utils/BlockTagUtil.java#L41
To resolve this, we filter patterns with `@link` since they are considered invalid references for imports.

---

Another issue arises with `{@link Set::add}`, which is an invalid link. correct one `{@link ClassName#methodName}`

The probelm was in extraction of  INLINE_TAG_PATTERN regex  
https://github.com/checkstyle/checkstyle/blob/1fa0377d6892f12e544e7c56acec4cf38c23d15e/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/utils/InlineTagUtil.java#L39

### DIFF
--- a/config/checker-framework-suppressions/checker-index-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-index-suppressions.xml
@@ -1079,6 +1079,17 @@
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java</fileName>
+    <specifier>return</specifier>
+    <message>incompatible types in return.</message>
+    <lineContent>return methodIndex;</lineContent>
+    <details>
+      type of expression: int
+      method return type: @LTLengthOf(&quot;input&quot;) int
+    </details>
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java</fileName>
     <specifier>argument</specifier>
     <message>incompatible argument for parameter index of String.charAt.</message>
--- a/config/checker-framework-suppressions/checker-regex-property-key-compiler-message-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-regex-property-key-compiler-message-suppressions.xml
@@ -298,6 +298,13 @@
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java</fileName>
+    <specifier>group.count.unknown</specifier>
+    <message>unable to verify non-literal groups parameter.</message>
+    <lineContent>final String methodPart = matcher.group(methodIndex);</lineContent>
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocMethodCheck.java</fileName>
     <specifier>group.count</specifier>
     <message>invalid groups parameter 1. Only 0 groups are guaranteed to exist for javadocArgMatcher.</message>
--- a/config/import-control.xml
+++ b/config/import-control.xml
@@ -184,6 +184,7 @@
     </file>
     <subpackage name="imports">
       <allow class="com.puppycrawl.tools.checkstyle.XmlLoader" local-only="true"/>
+      <allow class="org.checkerframework.checker.index.qual.IndexOrLow"/>
     </subpackage>
     <subpackage name="indentation">
       <allow pkg="java.lang.reflect"/>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
@@ -19,13 +19,17 @@
 
 package com.puppycrawl.tools.checkstyle.checks.imports;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.checkerframework.checker.index.qual.IndexOrLow;
 
 import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
@@ -130,6 +134,18 @@ public class UnusedImportsCheck extends AbstractCheck {
     /** Regexp pattern to match java.lang package. */
     private static final Pattern JAVA_LANG_PACKAGE_PATTERN =
         CommonUtil.createPattern("^java\\.lang\\.[a-zA-Z]+$");
+
+    /** Reference pattern. */
+    private static final Pattern REFERENCE = Pattern.compile(
+            "^([a-z_$][a-z\\d_$<>.]*)?(#(.*))?$",
+            Pattern.CASE_INSENSITIVE
+    );
+
+    /** Method pattern. */
+    private static final Pattern METHOD = Pattern.compile(
+            "^([a-z_$#][a-z\\d_$]*)(\\([^)]*\\))?$",
+            Pattern.CASE_INSENSITIVE
+    );
 
     /** Suffix for the star import. */
     private static final String STAR_IMPORT_SUFFIX = ".*";
@@ -343,16 +359,18 @@ public class UnusedImportsCheck extends AbstractCheck {
      * @return a set of classes referenced in the javadoc block
      */
     private static Set<String> collectReferencesFromJavadoc(TextBlock textBlock) {
-        final List<JavadocTag> tags = new ArrayList<>();
-        // gather all the inline tags, like @link
-        // INLINE tags inside BLOCKs get hidden when using ALL
-        tags.addAll(getValidTags(textBlock, JavadocUtil.JavadocTagType.INLINE));
-        // gather all the block-level tags, like @throws and @see
-        tags.addAll(getValidTags(textBlock, JavadocUtil.JavadocTagType.BLOCK));
+        // Process INLINE tags
+        final List<JavadocTag> inlineTags = getTargetTags(textBlock,
+                JavadocUtil.JavadocTagType.INLINE);
+        // Process BLOCK tags
+        final List<JavadocTag> blockTags = getTargetTags(textBlock,
+                JavadocUtil.JavadocTagType.BLOCK);
+        final List<JavadocTag> targetTags = Stream.concat(inlineTags.stream(), blockTags.stream())
+                .collect(Collectors.toUnmodifiableList());
 
         final Set<String> references = new HashSet<>();
 
-        tags.stream()
+        targetTags.stream()
             .filter(JavadocTag::canReferenceImports)
             .forEach(tag -> references.addAll(processJavadocTag(tag)));
         return references;
@@ -360,14 +378,22 @@ public class UnusedImportsCheck extends AbstractCheck {
 
     /**
      * Returns the list of valid tags found in a javadoc {@link TextBlock}.
+     * Filters tags based on whether they are inline or block tags, ensuring they match
+     * the correct format supported.
      *
      * @param cmt The javadoc block to parse
-     * @param tagType The type of tags we're interested in
+     * @param javadocTagType The type of tags we're interested in
      * @return the list of tags
      */
-    private static List<JavadocTag> getValidTags(TextBlock cmt,
-            JavadocUtil.JavadocTagType tagType) {
-        return JavadocUtil.getJavadocTags(cmt, tagType).getValidTags();
+    private static List<JavadocTag> getTargetTags(TextBlock cmt,
+            JavadocUtil.JavadocTagType javadocTagType) {
+        return JavadocUtil.getJavadocTags(cmt, javadocTagType)
+            .getValidTags()
+            .stream()
+            .filter(tag -> isMatchingTagType(tag, javadocTagType))
+            .map(UnusedImportsCheck::bestTryToMatchReference)
+            .flatMap(Optional::stream)
+            .collect(Collectors.toUnmodifiableList());
     }
 
     /**
@@ -421,6 +447,95 @@ public class UnusedImportsCheck extends AbstractCheck {
             topLevelType = type.substring(0, dotIndex);
         }
         return topLevelType;
+    }
+
+    /**
+     * Checks if a Javadoc tag matches the expected type based on its extraction format.
+     * This method checks if an inline tag is extracted as a block tag or vice versa.
+     * It ensures that block tags are correctly recognized as block tags and inline tags
+     * as inline tags during processing.
+     *
+     * @param tag The Javadoc tag to check.
+     * @param javadocTagType The expected type of the tag (BLOCK or INLINE).
+     * @return {@code true} if the tag matches the expected type, otherwise {@code false}.
+     */
+    private static boolean isMatchingTagType(JavadocTag tag,
+                                             JavadocUtil.JavadocTagType javadocTagType) {
+        final boolean isInlineTag = tag.isInlineTag();
+        final boolean isBlockTagType = javadocTagType == JavadocUtil.JavadocTagType.BLOCK;
+
+        return isBlockTagType != isInlineTag;
+    }
+
+    /**
+     * Attempts to match a reference string against a predefined pattern
+     * and extracts valid reference.
+     *
+     * @param tag the input tag to check
+     * @return Optional of extracted references
+     */
+    public static Optional<JavadocTag> bestTryToMatchReference(JavadocTag tag) {
+        final String content = tag.getFirstArg();
+        final int referenceIndex = extractReferencePart(content);
+        Optional<JavadocTag> validTag = Optional.empty();
+
+        if (referenceIndex != -1) {
+            final String referenceString;
+            if (referenceIndex == 0) {
+                referenceString = content;
+            }
+            else {
+                referenceString = content.substring(0, referenceIndex);
+            }
+            final Matcher matcher = REFERENCE.matcher(referenceString);
+            if (matcher.matches()) {
+                final int methodIndex = 3;
+                final String methodPart = matcher.group(methodIndex);
+                final boolean isValid = methodPart == null
+                        || METHOD.matcher(methodPart).matches();
+                if (isValid) {
+                    validTag = Optional.of(tag);
+                }
+            }
+        }
+        return validTag;
+    }
+
+    /**
+     * Extracts the reference part from an input string while ensuring balanced parentheses.
+     *
+     * @param input the input string
+     * @return -1 if parentheses are unbalanced, 0 if no method is found,
+     *         or the index of the first space outside parentheses.
+     */
+    private static @IndexOrLow("#1")int extractReferencePart(String input) {
+        int parenthesesCount = 0;
+        int firstSpaceOutsideParens = -1;
+        for (int index = 0; index < input.length(); index++) {
+            final char currentCharacter = input.charAt(index);
+
+            if (currentCharacter == '(') {
+                parenthesesCount++;
+            }
+            else if (currentCharacter == ')') {
+                parenthesesCount--;
+            }
+            else if (currentCharacter == ' ' && parenthesesCount == 0) {
+                firstSpaceOutsideParens = index;
+                break;
+            }
+        }
+
+        int methodIndex = -1;
+        if (parenthesesCount == 0) {
+            if (firstSpaceOutsideParens == -1) {
+                methodIndex = 0;
+            }
+            else {
+                methodIndex = firstSpaceOutsideParens;
+            }
+        }
+        return methodIndex;
     }
 
     /**
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
@@ -1,0 +1,54 @@
+/*
+UnusedImports
+processJavadoc = (default)true
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.imports.unusedimports;
+
+import java.util.Collections; // violation 'Unused import - java.util.Collections.'
+import java.util.List;
+import java.util.Set;         // violation 'Unused import - java.util.Set.'
+import java.util.ArrayList;
+import java.util.PriorityQueue; // violation 'Unused import - java.util.PriorityQueue.'
+import java.time.Duration;
+import java.util.Queue;         // violation 'Unused import - java.util.Queue.'
+import java.util.HashSet;
+import java.util.TreeSet;
+import java.util.Random;
+import java.util.LinkedList;   // violation 'Unused import - java.util.LinkedList.'
+import java.net.URI;
+import java.util.Collection;
+import java.time.LocalDate;
+import java.time.LocalDateTime; // violation 'Unused import - java.time.LocalDateTime.'
+import java.time.LocalTime;
+import java.time.Month;
+import java.util.concurrent.TimeUnit; // violation 'Unused import - java.util.concurrent.TimeUnit.'
+
+
+/**
+ * @link {Collections::emptyEnumeration}
+ * {@link TreeSet}
+ * {@see TimeUnit#SECONDS}
+ * {@link List#size}
+ * {@link Set::add}
+ * {@link ArrayList<LocalTime>#add( LocalTime )},
+ * {@link LinkedList<Integer>#set(int index(,int element)}
+ * {@link Random#ints},
+ * {@link PriorityQueue#123invalid}
+ *  {@link 123invalid}
+ * {@link Queue# }
+ * {@linkplain HashSet}
+ *  @see Duration#ZERO
+ *  {@link #setUrl(URI)}
+ *  {@link #Test(LocalDate, Collection)}
+ *  {@link #Test(LocalDateTime}}
+ *  {@link #extractParameters(double[], Month)}
+ *
+ */
+class InputUnusedImportsWithLinkAndMethodTag extends
+        InputUnusedImportsWithLinkAndMethodTagSuperClass {
+
+
+}
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTagSuperClass.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTagSuperClass.java
@@ -1,0 +1,12 @@
+package com.puppycrawl.tools.checkstyle.checks.imports.unusedimports;
+
+import java.net.URI;
+import java.time.LocalDate;
+import java.time.Month;
+import java.util.Collection;
+
+public class InputUnusedImportsWithLinkAndMethodTagSuperClass {
+        public void Test(LocalDate date, Collection s){}
+        public void setUrl(URI url){}
+        public void extractParameters(double[] d, Month month){}
+}
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkTag.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkTag.java
@@ -1,0 +1,22 @@
+/*
+UnusedImports
+processJavadoc = (default)true
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.imports.unusedimports;
+
+import java.util.List; // violation 'Unused import - java.util.List.'
+import java.util.Set;
+import java.util.TreeSet; // violation 'Unused import - java.util.TreeSet.'
+
+/**
+ * @link List
+ * {@link Set Set of items}
+ * @link TreeSet tree conatins items
+ *
+ */
+class InputUnusedImportsWithLinkTag {
+
+}
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTag.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTag.java
@@ -163,4 +163,13 @@ public class JavadocTag {
                 || tagInfo == JavadocTagInfo.EXCEPTION;
     }
 
+    /**
+     * Checks that the tag is a inline tag.
+     *
+     * @return whether the tag is a inline tag
+     */
+    public boolean isInlineTag() {
+        return tagInfo.getType() == JavadocTagInfo.Type.INLINE;
+    }
+
 }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheckTest.java
@@ -390,4 +390,32 @@ public class UnusedImportsCheckTest extends AbstractModuleTestSupport {
                 getPath("InputUnusedImportsJavadocAboveComments.java"), expected);
     }
 
+    @Test
+    public void testImportJavaLinkTag() throws Exception {
+        final String[] expected = {
+            "10:8: " + getCheckMessage(MSG_KEY, "java.util.List"),
+            "12:8: " + getCheckMessage(MSG_KEY, "java.util.TreeSet"),
+
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputUnusedImportsWithLinkTag.java"), expected);
+
+    }
+
+    @Test
+    public void testImportJavaLinkTagWithMethod() throws Exception {
+        final String[] expected = {
+            "10:8: " + getCheckMessage(MSG_KEY, "java.util.Collections"),
+            "12:8: " + getCheckMessage(MSG_KEY, "java.util.Set"),
+            "14:8: " + getCheckMessage(MSG_KEY, "java.util.PriorityQueue"),
+            "16:8: " + getCheckMessage(MSG_KEY, "java.util.Queue"),
+            "20:8: " + getCheckMessage(MSG_KEY, "java.util.LinkedList"),
+            "24:8: " + getCheckMessage(MSG_KEY, "java.time.LocalDateTime"),
+            "27:8: " + getCheckMessage(MSG_KEY, "java.util.concurrent.TimeUnit"),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputUnusedImportsWithLinkAndMethodTag.java"), expected);
+
+    }
+
 }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagTest.java
@@ -86,4 +86,12 @@ public class JavadocTagTest {
                 .isFalse();
     }
 
+    @Test
+    public void testJavadocTagIsInlineTag() {
+        assertThat(new JavadocTag(0, 0, "link", null).isInlineTag()).isTrue();
+        assertThat(new JavadocTag(0, 0, "value", null).isInlineTag()).isTrue();
+        assertThat(new JavadocTag(0, 0, "see", null).isInlineTag()).isFalse();
+
+    }
+
 }


zyadahmed:
@romani is `@see` and `@linkplain `consider import usage in checkstyle or not ?

zyadahmed:
> @romani is `@see` and `@linkplain `consider import usage in checkstyle or not ?

@romani

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -344,11 +346,21 @@ private void collectReferencesFromJavadoc(DetailAST ast) {
...
+                getValidTags(textBlock, JavadocUtil.JavadocTagType.INLINE)
+                        .stream()
+                        .filter(UnusedImportsCheck::isValidMethodLinkTag)
+                        .collect(Collectors.toList())
+        );
+        // gather all the block tags, like @throws and @see exclude @link notValid
+        tags.addAll(
+                getValidTags(textBlock, JavadocUtil.JavadocTagType.BLOCK)
+                        .stream()
+                        .filter(tag -> !tag.getTagName().equals(JavadocTagInfo.LINK.getName()))
```
romani, line 361: this exclude by name only is suspicious.
we should do somethig like isValidMethodLinkTag, some general method that can validate tag link for validnes.

please explain this  a bit more.

romani (changes requested):
in general all good,

items to clarify:

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
romani: do we mean to say that getValidTags should not return LINK tag ?

romani: sounds like getValidTags should do better job to do this filtering for both cases.

zyadahmed: > do we mean to say that getValidTags should not return LINK tag ?

yes becasue it return all tags similar like that @link @see and this format is not valid so we exclude it the only valid format for link should be between {}

romani: Let's move logic of filtering to getValidTags

romani:
https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html
Search for online tags:
{@link package.class#member label}
{@linkplain package.class#member label}

So filtering should be named as `isBlockTag` that filters out all inline tags.

Read on difference of block and inline tags at `Block and inline Tags` chapter of https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html

romani:
@mahfouz72 , please look at this PR, to get understanding of how diverse javadoc .

zyadahmed:
@romani I changed and moved all logic to JavadocUtil to filter all tags from the starting point and resolved all mutations related to new logic . However, the CI failed because other mutations failed in different parts of the code.

src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java
```diff
@@ -130,6 +130,21 @@ public static JavadocTags getJavadocTags(TextBlock textBlock,
...
+     * @param javadocTag the type of the Javadoc tag
+     * @return {@code true} if the tag is valid, otherwise {@code false}
+     */
+    public static boolean isValidTag(TagInfo tag, JavadocTagType javadocTag) {
+        boolean isValid = true;
+        if (javadocTag == JavadocTagType.BLOCK) {
+            isValid = !tag.getName().equals(JavadocTagInfo.LINK.getName());
+        }
+        return isValid && !tag.getValue().contains("::");
+    }
```
mahfouz72, line 146: All of these can reference imports

https://github.com/checkstyle/checkstyle/blob/8531fec143058ddd92fab14839f2e9389ac6cf5c/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTag.java#L157-L164
   ---
   
Your method deals with links only. I think direction is good but we have to account for all others tag and do more work.
   
Tags That Reference Imports:

Block Tags: `@see`, `@throws`, `@exception`
Inline Tags: `{@link}`, `{@value}`, `{@linkplain}`

Filtering Logic:

- If the tag type is block (i.e., it does not have `{ }`), then exclude it if its name is link, value, or linkplain (since these are only valid as inline tags).

- Validate the value to ensure it is a correct reference (e.g., no `::` method references, correct format for links).

mahfouz72 (changes requested):
item

src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java
zyadahmed: > * Validate the value to ensure it is a correct reference (e.g., no `::` method references, correct format for links).

@romani @mahfouz72   The problem is to validate whether a given value is a correct reference. Using the same approach, we need to match it with a regex. However, the regex would be too complex and might not handle some edge cases. What is the correct approach to solve this problem?

romani: it is ok, all validation by regexp has problems on edge cases, name method `bestTryToMatchReference`.

We will never handle all, but we can handle some basic.
When Check is rewritten in AST based model, it will be more strict on validation

romani:
CI is very red

zyadahmed:
> CI is very red

I will try to fix it for some reason related to mutations and regex, but is the solution good?

src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java
```diff
@@ -118,18 +130,121 @@ public static JavadocTags getJavadocTags(TextBlock textBlock,
             // Lines are one-indexed, so need an off-by-one correction.
             final int line = textBlock.getStartLineNo() + tag.getPosition().getLine() - 1;
 
-            if (JavadocTagInfo.isValidName(tag.getName())) {
-                validTags.add(
-                    new JavadocTag(line, col, tag.getName(), tag.getValue()));
+            if (JavadocTagInfo.isValidName(tag.getName()) && isValidImportTag(tag, tagType)) {
```
romani, line 133: For `getJavadocTags(TextBlock` it is not good to filter out valid .

Let's move isValidImportTag back to Check.
Please rename to isValidTagWithJavaElemetRegerence

Please avoid "import" in all other names. It is reference it is not import

romani (changes requested):
In general ok.

Items

src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java
zyadahmed: @romani But is that considered invalid tags from the beginning without reaching any checks? Because in `JavadocUtil#getJavadocTags`, it actually separates both valid and invalid tags. In that case, do we consider them invalid from the beginning and not treat them as valid tags?

https://github.com/checkstyle/checkstyle/blob/e06aaf2ef8676e6001d9d85e523964477730012a/src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java#L88 
as it show here we get from that method valid tags only ?

romani: this is good usage:
https://github.com/checkstyle/checkstyle/blob/cf745474e145584543684f2c812d3eeed31328ff/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java#L378-L382

decision on valid of not valid is happening in Check, but each Check will see tags to be invalid by their own rules.
If we put more in colletion of invalid to JavadocTags, it might affect other Check.
   
we need a bit more deep refacoting here to put logic on what is valid and what is invalid to Checks. Util class just have functions that are shared between Checks, but Util functions does not make opinion on valid. 
JavadocTypeCheck, UnusedImportsCheck will be affected by this update, it is ok.

as you split/copy logic fromm util  to Checks it will be more logical and each check can decided by itself what is valid.

zyadahmed: Okay, I will work on it. Thanks a lot for the explanation! So, we can fix the problem by adding the filtering here in that part, right? 
https://github.com/checkstyle/checkstyle/blob/cf745474e145584543684f2c812d3eeed31328ff/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java#L355-L358

romani: Yes

zyadahmed:
@romani I tried to make the CI green, but the CI/CircleCI errors are unrelated to the classes I changed. The failure  in pitest test are related to JavadocUtil, from which I removed all my new changes. also checke (index), where I had already suppressed the check error because it was related to how I handle the logic. However, all other checks have passed, and I have resolved all mutations related to my code.

zyadahmed:
@romani  i changed all logic in UnusedImportCheck but i have some probelms in ci  in Checker (index) i already suppressed the erros but still not working and i dont have any idea how to solve ci/circleci

rnveach:
`no-error-*` runs are against 3rd party repos. It is validating your changes against other sources to see if they will cause regression on other projects. You should be able to look at the CIs to see where the errors come from and look at the actual files. Verify that the new violations being reported are correct and not bugs in your implementation. You may want to look to see if any should be added to our own tests. You should post some examples here for @romani and others to confirm. Once the check updates are resolved here to be all good (someone will most likely tell you when we reach this), you will need to provide fixes to the 3rd party repos for them to merge and CI to pass here.

rnveach:
`run-inspections` can be found at https://app.circleci.com/pipelines/github/checkstyle/checkstyle/1/workflows/4ff6fca1-b529-462c-b71e-b4cfdbf47c6f/jobs/821142/artifacts . Let us know if there is some error you are not understanding how to resolve.

`semaphoreci` failure is another no-run error.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
romani: Let's keep all mess in Check.
Method should be until only if it proven to be good for all usages

config/checker-framework-suppressions/checker-index-suppressions.xml
```diff
@@ -1078,6 +1078,28 @@
     </details>
   </checkerFrameworkError>
 
+  <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java</fileName>
+    <specifier>argument</specifier>
+    <message>incompatible argument for parameter endIndex of String.substring.</message>
+    <lineContent>referenceString = content.substring(0, referenceIndex);</lineContent>
```
romani, line 1085: while I reviewing you PR,
please share reasons on not fixing this Checker violation.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -344,17 +356,23 @@ private void collectReferencesFromJavadoc(DetailAST ast) {
...
+                getValidTags(textBlock, JavadocUtil.JavadocTagType.BLOCK),
+                JavadocUtil.JavadocTagType.BLOCK
+        ));
+
+        final Set<String> references = new HashSet<>();
         tags.stream()
-            .filter(JavadocTag::canReferenceImports)
-            .forEach(tag -> references.addAll(processJavadocTag(tag)));
+                .filter(JavadocTag::canReferenceImports)
+                .forEach(tag -> references.addAll(processJavadocTag(tag)));
```
romani, line 375: please revert unrelated formatting to simplify diff

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -367,6 +385,7 @@ private static Set<String> collectReferencesFromJavadoc(TextBlock textBlock) {
      */
     private static List<JavadocTag> getValidTags(TextBlock cmt,
             JavadocUtil.JavadocTagType tagType) {
+
```
romani, line 388: revert please

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTagSuperClass.java
```diff
@@ -0,0 +1,14 @@
...
+import java.time.LocalDate;
+import java.time.Month;
+import java.util.Collection;
+
+public class InputUnusedImportsWithLinkAndMethodTagSuperClass {
+        public void Test(LocalDate date, Collection s){}
+        public void setUrl(URI url){}
+        public void extractParameters(double[] d, Month month){
+
+        }
```
romani, line 13: `{}`

romani (changes requested):
shallow review:

config/checker-framework-suppressions/checker-index-suppressions.xml
zyadahmed: > while I reviewing you PR, please share reasons on not fixing this Checker violation.

The extractReferencePart method returns a value ranging from `0` to `input.length()-1 `if the reference is valid,  there is no chance of going out of bounds. In the case of an invalid reference, it returns -1, which is already handled in the parent method. Adding extra handling for this case would introduce unnecessary code, leading to a Pitest suppression without any real user scenario to handle that case

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTagSuperClass.java
zyadahmed: > `{}`

done

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
zyadahmed: > revert please

done

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
zyadahmed: > please revert unrelated formatting to simplify diff

done

config/checker-framework-suppressions/checker-regex-property-key-compiler-message-suppressions.xml
```diff
@@ -297,6 +297,13 @@
     <lineContent>references.add(topLevelType(matcher.group(1)));</lineContent>
   </checkerFrameworkError>
 
+  <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java</fileName>
+    <specifier>group.count.unknown</specifier>
+    <message>unable to verify non-literal groups parameter.</message>
+    <lineContent>final String methodPart = matcher.group(methodIndex);</lineContent>
```
romani, line 304: please explain this.

config/checker-framework-suppressions/checker-index-suppressions.xml
```diff
@@ -1078,6 +1078,28 @@
...
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java</fileName>
+    <specifier>argument</specifier>
+    <message>incompatible argument for parameter endIndex of String.substring.</message>
+    <lineContent>referenceString = content.substring(0, referenceIndex);</lineContent>
+    <details>
+      found   : int
+      required: @NonNegative int
```
romani, line 1099: please explain this

romani (changes requested):
thanks a lot:

few more requests:

config/checker-framework-suppressions/checker-index-suppressions.xml
romani: thanks a lot.
Do you know any annotation on variable that would potentially allow to avoid violation from Checker? 
wihtout intruduction of new jar dependecy.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -423,6 +441,100 @@ private static String topLevelType(String type) {
...
+    /**
+     * Attempts to match a reference string against a predefined pattern
+     * and extracts valid references.
+     *
+     * @param tag the input tag to check
+     * @return a list of extracted references
+     */
+    public static List<JavadocTag> bestTryToMatchReference(JavadocTag tag) {
+        final String content = tag.getFirstArg();
+        final List<JavadocTag> validTags = new ArrayList<>();
```
romani, line 473: why single tag mapped to list of tags.
match should return Optional, outside of this method , there should be a code that convert optional to required type.

but if you do like `tags.addAll(bestTryToMatchReference(getValidReferenceTag(validTags));` you will get as parameter list of tags, and map them to result filtered list. so it will be better from design perspectve.

romani (changes requested):
items:

config/checker-framework-suppressions/checker-index-suppressions.xml
zyadahmed: > please explain this

Here, it generates a warning because `extractReferencePart` produces a result within the valid range if it is a valid part  otherwise it returns an invalid index, which in the implementation is -1. However, that case is already handled. I tried to find a better solution, but it would either complicate the implementation or require a different return value that does not align with the method purpose.

src/test/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtilTest.java
```diff
@@ -179,6 +179,7 @@ public void testInvalidTags() {
             "/** @fake block",
             " * {@bogus inline}",
             " * {@link List valid}",
+
```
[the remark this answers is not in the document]
zyadahmed: > please revert

done

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
zyadahmed: > why single tag mapped to list of tags. match should return Optional, outside of this method , there should be a code that convert optional to required type.
> 
> but if you do like `tags.addAll(bestTryToMatchReference(getValidReferenceTag(validTags));` you will get as parameter list of tags, and map them to result filtered list. so it will be better from design perspectve.

yes that correct sorry for that i choose to use Optional

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -344,14 +356,20 @@ private void collectReferencesFromJavadoc(DetailAST ast) {
...
+        // Process INLINE tags
+        tags.addAll(getValidReferenceTag(
+                getValidTags(textBlock, JavadocUtil.JavadocTagType.INLINE),
+                JavadocUtil.JavadocTagType.INLINE
+        ));
 
+        // Process BLOCK tags
+        tags.addAll(getValidReferenceTag(
+                getValidTags(textBlock, JavadocUtil.JavadocTagType.BLOCK),
+                JavadocUtil.JavadocTagType.BLOCK
```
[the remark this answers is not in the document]
zyadahmed: > too nested. make it like:
> 
> ```
> validTags = getValidTags(textBlock, JavadocUtil.JavadocTagType.BLOCK),
>                 JavadocUtil.JavadocTagType.BLOCK)
>         tags.addAll(bestTryToMatchReference(getValidReferenceTag(validTags));
> ```

i remove nested but change a little bit to make ci run inspections green

config/checker-framework-suppressions/checker-index-suppressions.xml
romani: @zyadahmed , ^^

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -423,6 +441,100 @@ private static String topLevelType(String type) {
...
     }
 
+    /**
+     * Retrieves all valid reference tags based on their Javadoc type.
+     *
+     * @param tags        the tag to validate
+     * @param javadocTag type of tag to check
+     * @return a {@code List} of valid Javadoc tags based on the specified type
+     */
+    public static List<JavadocTag> getValidReferenceTag(
```
romani, line 451: getValidReferenceTags

as it return collection.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -423,6 +441,100 @@ private static String topLevelType(String type) {
...
+            Iterable<JavadocTag> tags, JavadocUtil.JavadocTagType javadocTag) {
+        final List<JavadocTag> validTags = new ArrayList<>();
+
+        for (JavadocTag tag : tags) {
+            if (javadocTag == JavadocUtil.JavadocTagType.BLOCK && tag.isInlineTag()) {
+                continue;
+            }
+            bestTryToMatchReference(tag).ifPresent(validTags::add);
+        }
```
romani, line 460: try to rewrite is by java streams.
.filter().map().collect()

romani (changes requested):
please generate diff report 
see details at https://github.com/checkstyle/checkstyle/tree/master/.github/workflows#diff-report-by-configuration-in-input-file

you can find numerous examples of usage in our old merged PRs

items:

zyadahmed:
Github, generate report for UnusedImport/all-examples-in-one

zyadahmed:
Github, generate report for UnusedImports/all-examples-in-one

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
zyadahmed: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
zyadahmed: done with streams

zyadahmed:
> Report for UnusedImports/all-examples-in-one: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c5f69a32321cd7613369167c1be1c7193ab411c6_2025173059/reports/diff/index.html

The only difference in Checkstyle is that `InputUnusedImportsFromStaticMethodRefJavadocDisabled` generates a violation. It should already be a violation as I understand it. This is my first time using that report.

romani:
@mahfouz72 , please finish review, PR is on final stages

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -343,12 +356,17 @@ private void collectReferencesFromJavadoc(DetailAST ast) {
...
+        final List<JavadocTag> inlineTags = getValidTags(textBlock,
+                                        JavadocUtil.JavadocTagType.INLINE);
+        final List<JavadocTag> tags = new ArrayList<>(getValidReferenceTags(inlineTags,
+                                                    JavadocUtil.JavadocTagType.INLINE));
+
+        // Process BLOCK tags
+        final List<JavadocTag> blockTags = getValidTags(textBlock,
+                                        JavadocUtil.JavadocTagType.BLOCK);
+        tags.addAll(getValidReferenceTags(blockTags,
+                                        JavadocUtil.JavadocTagType.BLOCK));
```
mahfouz72, line 369:
```suggestion
        final List<JavadocTag> validReferenceTags = new ArrayList<>();
        
        final List<JavadocTag> inlineTags = getValidTags(textBlock,
                                        JavadocUtil.JavadocTagType.INLINE);
        validReferenceTags.addAll(getValidReferenceTags(inlineTags, JavadocUtil.JavadocTagType.INLINE));

        final List<JavadocTag> blockTags = getValidTags(textBlock,
                                        JavadocUtil.JavadocTagType.BLOCK);
        validReferenceTags.addAll(getValidReferenceTags(blockTags,
                                        JavadocUtil.JavadocTagType.BLOCK));
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -423,6 +441,99 @@ private static String topLevelType(String type) {
...
 
+    /**
+     * Retrieves all valid reference tags based on their Javadoc type.
+     *
+     * @param tags        the tag to validate
+     * @param javadocTag type of tag to check
+     * @return a {@code List} of valid Javadoc tags based on the specified type
+     */
+    public static List<JavadocTag> getValidReferenceTags(
+            Collection<JavadocTag> tags, JavadocUtil.JavadocTagType javadocTag) {
```
mahfouz72, line 452:
```suggestion
            Collection<JavadocTag> tags, JavadocUtil.JavadocTagType javadocTagType) {
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -423,6 +441,99 @@ private static String topLevelType(String type) {
...
+        final List<JavadocTag> validTags = new ArrayList<>();
+
+        tags.stream()
+            .filter(tag -> !(javadocTag == JavadocUtil.JavadocTagType.BLOCK && tag.isInlineTag()))
```
mahfouz72, line 456: Does this line mean that tags intended to be inline but written as block tags are ignored?

Example:
```
/**
  *   @link List
  */
```

For the above Javadoc:

- `javadocTagType == JavadocUtil.JavadocTagType.BLOCK` returns true because `@link` is written as a block tag.
- `tag.isInlineTag()` returns true because `@link` is an inline tag according to the Javadoc specification.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -423,6 +441,99 @@ private static String topLevelType(String type) {
...
+            else if (currentCharacter == ')') {
+                parenthesesCount--;
+            }
+            else if (currentCharacter == ' ' && parenthesesCount == 0) {
+                firstSpaceOutsideParens = index;
+                break;
+            }
+        }
+
+        final int methodIndex;
```
mahfouz72, line 522: We can initialize this with -1 and remove the else branch.

mahfouz72 (changes requested):
Good Stuff!

Few items:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
```diff
@@ -0,0 +1,51 @@
...
+import java.time.LocalDate;
+import java.time.LocalDateTime; // violation 'Unused import - java.time.LocalDateTime.'
+import java.time.LocalTime;
+import java.time.Month;
+
+
+/**
+ * @link {Collections::emptyEnumeration}
+ * {@link TreeSet}
+ * {@link List#size}
```
mahfouz72, line 32: All test cases for link and see tags only. Please add test cases for all tags that can reference imports :

```
    public boolean canReferenceImports() {
        return tagInfo == JavadocTagInfo.SEE
                || tagInfo == JavadocTagInfo.LINK
                || tagInfo == JavadocTagInfo.VALUE
                || tagInfo == JavadocTagInfo.LINKPLAIN
                || tagInfo == JavadocTagInfo.THROWS
                || tagInfo == JavadocTagInfo.EXCEPTION;
    }
```

Also, cases when the tag has a description. like : ` {@link List List of items}`

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -343,12 +356,17 @@ private void collectReferencesFromJavadoc(DetailAST ast) {
...
-        final List<JavadocTag> tags = new ArrayList<>();
-        // gather all the inline tags, like @link
-        // INLINE tags inside BLOCKs get hidden when using ALL
-        tags.addAll(getValidTags(textBlock, JavadocUtil.JavadocTagType.INLINE));
-        // gather all the block-level tags, like @throws and @see
-        tags.addAll(getValidTags(textBlock, JavadocUtil.JavadocTagType.BLOCK));
+        // Process INLINE tags
+        final List<JavadocTag> inlineTags = getValidTags(textBlock,
+                                        JavadocUtil.JavadocTagType.INLINE);
+        final List<JavadocTag> tags = new ArrayList<>(getValidReferenceTags(inlineTags,
```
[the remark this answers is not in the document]
zyadahmed: > To let code looks same as next block

to make them both identical cli will fail in run inspection in 
```
<package>com.puppycrawl.tools.checkstyle.checks.imports</package>
<entry_point TYPE="method" FQNAME="com.puppycrawl.tools.checkstyle.checks.imports.UnusedImportsCheck java.util.Set<java.lang.String> collectReferencesFromJavadoc(com.puppycrawl.tools.checkstyle.api.TextBlock textBlock)"/>
<problem_class id="CollectionAddAllCanBeReplacedWithConstructor" severity="ERROR" attribute_key="ERRORS_ATTRIBUTES">Redundant 'Collection.addAll()' call</problem_class>
<description>'addAll()' call can be replaced with parametrized constructor call</description>
<highlighted_element>addAll</highlighted_element>
```

romani:
@zyadahmed, please reply "done" to items you addressed.

Please keep CI green.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
zyadahmed: > All test cases for link and see tags only. Please add test cases for all tags that can reference imports :
> 
> ```
>     public boolean canReferenceImports() {
>         return tagInfo == JavadocTagInfo.SEE
>                 || tagInfo == JavadocTagInfo.LINK
>                 || tagInfo == JavadocTagInfo.VALUE
>                 || tagInfo == JavadocTagInfo.LINKPLAIN
>                 || tagInfo == JavadocTagInfo.THROWS
>                 || tagInfo == JavadocTagInfo.EXCEPTION;
>     }
> ```
> 
> Also, cases when the tag has a description. like : ` {@link List List of items}`

done in `InputUnusedImportsWithLinkTag`

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
zyadahmed: > We can initialize this with -1 and remove the else branch.

done

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
zyadahmed: > 

done

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
zyadahmed: > 

 tried to create that solution, but it failed during the inspection because it requires a parameterized constructor . another solution If I only declare final List<JavadocTag> validReferenceTags; and then add it in inline tags and block tags, it also fails the inspection, but with a 'Too far declaration' error."

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -423,6 +441,99 @@ private static String topLevelType(String type) {
...
+    public static List<JavadocTag> getValidReferenceTags(
+            Collection<JavadocTag> tags, JavadocUtil.JavadocTagType javadocTag) {
+        final List<JavadocTag> validTags = new ArrayList<>();
+
+        tags.stream()
+            .filter(tag -> !(javadocTag == JavadocUtil.JavadocTagType.BLOCK && tag.isInlineTag()))
+            .map(UnusedImportsCheck::bestTryToMatchReference)
+            .flatMap(Optional::stream)
+            .forEach(validTags::add);
+        return validTags;
```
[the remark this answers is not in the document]
zyadahmed: > Try to optimize it more to get form:
> 
> ```
> return tags.stream().....
> ```

done

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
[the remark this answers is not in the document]
zyadahmed: > ```
> final List<JavadocTag> tags = new ArrayList<>();
> // Process INLINE tags
> final List<JavadocTag> inlineTags = 
>         getValidTags(textBlock, JavadocUtil.JavadocTagType.INLINE);
> tags.addAll(getValidReferenceTags(inlineTags, JavadocUtil.JavadocTagType.INLINE));
> ```

i tried this and some similar it always fail in run inspections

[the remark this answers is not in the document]
romani: sad, lets avoid violation.

please format code like:
```

        // Process INLINE tags
        final List<JavadocTag> inlineTags = 
                getValidTags(textBlock, JavadocUtil.JavadocTagType.INLINE);
        final List<JavadocTag> validTags = new ArrayList<>(
                getValidReferenceTags(inlineTags, JavadocUtil.JavadocTagType.INLINE));

        // Process BLOCK tags
        final List<JavadocTag> blockTags = 
                getValidTags(textBlock, JavadocUtil.JavadocTagType.BLOCK);
        validTags.addAll(
                getValidReferenceTags(blockTags, JavadocUtil.JavadocTagType.BLOCK));
```

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTagSuperClass.java
mahfouz72, line 1: Why did we add this file?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkTag.java
```diff
@@ -0,0 +1,20 @@
...
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.imports.unusedimports;
+
+import java.util.List; // violation 'Unused import - java.util.List.'
+import java.util.Set;
+
+/**
+ * @link List
+ * {@link Set Set of items}
```
mahfouz72, line 15: Another example like this but with invalid syntax:
`@link Set Set of items`  
and show that it works by having an unused import violation on `Set`

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -423,6 +443,94 @@ private static String topLevelType(String type) {
...
+     *
+     * @param tags        the tag to validate
+     * @param javadocTagType type of tag to check
+     * @return a {@code List} of valid Javadoc tags based on the specified type
+     */
+    public static List<JavadocTag> getValidReferenceTags(
+            Collection<JavadocTag> tags, JavadocUtil.JavadocTagType javadocTagType) {
+
+        return tags.stream()
+            .filter(tag -> javadocTagType != JavadocUtil.JavadocTagType.BLOCK || !tag.isInlineTag())
```
romani, line 457: this condition is supper confusing.

we hardcoding value of javadocTagType 
and filtering by it, some nonsense is there.
```
       final List<JavadocTag> blockTags = getValidTags(textBlock,
                JavadocUtil.JavadocTagType.BLOCK);
        validTags.addAll(getValidReferenceTags(blockTags,
                JavadocUtil.JavadocTagType.BLOCK));
```

you should move it filtering to this method:
```
private static List<JavadocTag> getValidTags(TextBlock cmt,
            JavadocUtil.JavadocTagType javadocTagType) {
        return JavadocUtil.getJavadocTags(cmt, javadocTagType).getValidTags();
    }
```

so list of tags will be valid and we do not need not trust it.

romani (changes requested):
items:

mahfouz72 (changes requested):
Thanks alot!!

Last for me:

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -343,16 +357,22 @@ private void collectReferencesFromJavadoc(DetailAST ast) {
...
+        final List<JavadocTag> inlineTags =
+                getValidTags(textBlock, JavadocUtil.JavadocTagType.INLINE);
+        final List<JavadocTag> validTags = new ArrayList<>(
+                getValidReferenceTags(inlineTags));
+
+        // Process BLOCK tags
+        final List<JavadocTag> blockTags =
+                getValidTags(textBlock, JavadocUtil.JavadocTagType.BLOCK);
+        validTags.addAll(
+                getValidReferenceTags(blockTags));
```
romani, line 371: here we combine two collections.
please use this form of code:
```
List<JavadocTag> targetTags = Stream
                .concat(listxxxx1.stream(), 
                       listxxxxxx2.stream())
```

romani (changes requested):
we are very close:

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -362,12 +382,16 @@ private static Set<String> collectReferencesFromJavadoc(TextBlock textBlock) {
...
      * @return the list of tags
      */
     private static List<JavadocTag> getValidTags(TextBlock cmt,
-            JavadocUtil.JavadocTagType tagType) {
-        return JavadocUtil.getJavadocTags(cmt, tagType).getValidTags();
+            JavadocUtil.JavadocTagType javadocTagType) {
+        return JavadocUtil.getJavadocTags(cmt, javadocTagType)
+            .getValidTags()
+            .stream()
+            .filter(tag -> javadocTagType != JavadocUtil.JavadocTagType.BLOCK || !tag.isInlineTag())
```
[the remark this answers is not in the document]
zyadahmed: > `.filter(tag -> javadocTagType != JavadocUtil.JavadocTagType.BLOCK || !tag.isInlineTag())`
> 
> this condition is weird. `javadocTagType != JavadocUtil.JavadocTagType.BLOCK` doe snot use tag.
> 
> sounds like we just need separate method for BLOCK

This filter is more related to whether the tag type matches or differs from the extraction format. For example, `@link `is an inline tag but is extracted as a block tag. I tried different ways to simplify it, but maybe we can move it to a separate method and explain it properly in the Javadoc summary. What do you think?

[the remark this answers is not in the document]
romani: Please try.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTagSuperClass.java
zyadahmed: i made this file to simulate the exceution of checkstyle dependecies where we have superclass that have some method and reference is mentioned inside subclass  which refer to superclass method and it's references  so i needed to make test same exactly as dependecies

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkTag.java
zyadahmed: > Another example like this but with invalid syntax: `@link Set Set of items` and show that it works by having an unused import violation on `Set`

done

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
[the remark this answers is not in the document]
zyadahmed: > add ` .map(UnusedImportsCheck::bestTryToMatchReference)` to this chain.
> 
> and we do not need `getValidReferenceTags(`

done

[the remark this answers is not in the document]
zyadahmed: > Please try.

i tried to make simplest conditon as much as possible and write detailed javadoc about usage

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -343,16 +357,22 @@ private void collectReferencesFromJavadoc(DetailAST ast) {
...
-        // gather all the inline tags, like @link
-        // INLINE tags inside BLOCKs get hidden when using ALL
-        tags.addAll(getValidTags(textBlock, JavadocUtil.JavadocTagType.INLINE));
-        // gather all the block-level tags, like @throws and @see
-        tags.addAll(getValidTags(textBlock, JavadocUtil.JavadocTagType.BLOCK));
+
+        // Process INLINE tags
+        final List<JavadocTag> inlineTags =
+                getValidTags(textBlock, JavadocUtil.JavadocTagType.INLINE);
+        final List<JavadocTag> validTags = new ArrayList<>(
```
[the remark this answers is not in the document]
zyadahmed: > targetTags

done

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -362,12 +382,16 @@ private static Set<String> collectReferencesFromJavadoc(TextBlock textBlock) {
      * Returns the list of valid tags found in a javadoc {@link TextBlock}.
      *
      * @param cmt The javadoc block to parse
-     * @param tagType The type of tags we're interested in
+     * @param javadocTagType The type of tags we're interested in
      * @return the list of tags
      */
     private static List<JavadocTag> getValidTags(TextBlock cmt,
```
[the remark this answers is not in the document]
zyadahmed: > `getTargetTags`
> 
> as we filter out completely valid tags only because they are not our interest. Please update javadoc summary to describe that what we return or what we filterout.

done

config/checker-framework-suppressions/checker-regex-property-key-compiler-message-suppressions.xml
romani: https://checkerframework.org/manual/#regex-capturing-groups
https://checkerframework.org/api/org/checkerframework/checker/regex/qual/Regex.html
https://checkerframework.org/manual/#regex-annotations

config/checker-framework-suppressions/checker-index-suppressions.xml
romani: yes, agree, code logic enforce it to be positive.
```
        if (referenceIndex != -1) {
            final String referenceString;
            if (referenceIndex == 0) {
                referenceString = content;
            }
            else {
                referenceString = content.substring(0, referenceIndex);
```

config/checker-framework-suppressions/checker-regex-property-key-compiler-message-suppressions.xml
romani: can something like this help to avoid violation ?
```
$ git diff
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
index 9e0ecf6231..0c14541fd7 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
@@ -29,6 +29,8 @@ import java.util.regex.Pattern;
 
+import org.checkerframework.checker.regex.qual.Regex;
+
@@ -134,7 +136,7 @@ public class UnusedImportsCheck extends AbstractCheck {
         CommonUtil.createPattern("^java\\.lang\\.[a-zA-Z]+$");
 
     /** Reference pattern. */
-    private static final Pattern REFERENCE = Pattern.compile(
+    private static final @Regex(3) Pattern REFERENCE = Pattern.compile(
             "^([a-z_$][a-z\\d_$<>.]*)?(#(.*))?$",
             Pattern.CASE_INSENSITIVE
     );

```

```
✔ ~/java/github/romani/checkstyle [zyadahmed/UnusedImportsIssue L|✚ 1⚑ 2] 
$ groovy ./.ci/checker-framework.groovy checker-regex-property-key-compiler-message

```

romani:
@mahfouz72 , please do final review , please share your opinion on possible extra annoation `@Regexp` in code to avoid checker to not place violation.

zyadahmed:
Github, generate report for UnusedImports/all-examples-in-one

zyadahmed:
Github, generate report for UnusedImports/all-examples-in-one

config/checker-framework-suppressions/checker-regex-property-key-compiler-message-suppressions.xml
zyadahmed: I tried this, but it still produces the same error.

config/checker-framework-suppressions/checker-index-suppressions.xml
zyadahmed: > @zyadahmed , please open issue on checkerframework to get to know if they can improve validation, so this suppression will be temporal until they improve

I opened it and will try to follow along until someone resolves it.

romani:
Rebased in web

config/checker-framework-suppressions/checker-index-suppressions.xml
romani: Please share link

config/checker-framework-suppressions/checker-regex-property-key-compiler-message-suppressions.xml
romani: Please create issue on them to let them help on this violation and how to resolve it

config/checker-framework-suppressions/checker-index-suppressions.xml
zyadahmed: > Please share link

https://github.com/typetools/checker-framework/issues/7008

romani: can we apply what is suggested in that issue.

config/checker-framework-suppressions/checker-regex-property-key-compiler-message-suppressions.xml
zyadahmed: > Please create issue on them to let them help on this violation and how to resolve it

can we make it in separate pr after that beacause that pr is too long now or what do you think ?

romani (approved):
Thanks a lot for update,
I am approving to let one more maintainer to  review while you finishing with checkerwramework, please try to avoid Checker suppression.

config/checker-framework-suppressions/checker-regex-property-key-compiler-message-suppressions.xml
romani: yes, we can, but please share how it can be fixed.

config/checker-framework-suppressions/checker-index-suppressions.xml
romani: if `<lineContent>referenceString = content.substring(0, referenceIndex);</lineContent>` suppression stays, it means that anontation is not working.
I would rather remove annotation  a keep conversation in check issue.

romani:
@zyadahmed, if annotation are not working, please revert usage of them, and share feedback in Checker issue tracker, we will merge with suppression

zyadahmed:
> @zyadahmed, if annotation are not working, please revert usage of them, and share feedback in Checker issue tracker, we will merge with suppression

revert last update and opened pr  https://github.com/typetools/checker-framework/issues/7013

zyadahmed:
> @zyadahmed, if annotation are not working, please revert usage of them, and share feedback in Checker issue tracker, we will merge with suppression

revert last update and opened pr  https://github.com/typetools/checker-framework/issues/7013

zyadahmed:
Github, generate report for UnusedImports/all-examples-in-one

romani:
Github, generate report for UnusedImports/all-examples-in-one

config/import-control.xml
```diff
@@ -184,6 +184,7 @@
     </file>
     <subpackage name="imports">
       <allow class="com.puppycrawl.tools.checkstyle.XmlLoader" local-only="true"/>
+      <allow class="org.checkerframework.checker.index.qual.IndexOrLow"/>
```
romani, line 187: Please remove, they are not working.

zyadahmed: > Please remove, they are not working.

But after we added this annotation, we removed one suppression, or am I missing something?

```
 <message>incompatible argument for parameter endIndex of String.substring.</message>
    <lineContent>referenceString = content.substring(0, referenceIndex);</lineContent>
    <details>
      found   : int
      required: @NonNegative int
   </details>
```

romani: I saw red CI on Checker .... that is why I requested to remove.

I am very confused. Ok lets merge with suppressions and improve in small PR on suppression removal

romani: Oh my bad, it actually helped but another poped up:
```

    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java</fileName>
    <specifier>return</specifier>
    <message>incompatible types in return.</message>
    <lineContent>return methodIndex;</lineContent>
    <details>
      type of expression: int
      method return type: @LTLengthOf(&quot;input&quot;) int
    </details>
  </checkerFrameworkError>
```

ok, lets keep this annotation. And resolve above new in separate PR.

nrmancuso (approved):

config/checker-framework-suppressions/checker-index-suppressions.xml
romani: @zyadahmed , please reopen issue to share details that it is not working even it supposed to work

zyadahmed: > @zyadahmed , please reopen issue to share details that it is not working even it supposed to work

I will try to investigate more into suppressions and open issues and PRs of  currently exists issues is that what you mean  right ?

romani: Yes, please.

Issue #16375, opened 2025-02-20
Add `UnusedLocalVariableCheck` counterpart regarding `UnusedLocalMethodCheck`


punkratz312:
Im wondering if there is already a counterpart of the `UnusedLocalVariableCheck` to check for `UnusedLocalMethodCheck`. 

Just like there are already 2 rules inside PMD and openrewrite to address this. If you think its good i could provide a skeleton as the code should be similar to the LocalVariableCheck.

https://docs.openrewrite.org/recipes/staticanalysis/removeunusedprivatemethods
https://docs.openrewrite.org/recipes/staticanalysis/removeunusedprivatefields

romani:
Good idea, better name `UnusedPrivateMethodCheck`.

but we might have problems  to implement it. We do not have types in runtime.
So it will be close to impossible to differentiate overloaded methods, name name different sequence of parameters.
https://checkstyle.org/writingchecks.html#Limitations

Pankraz76:
no need to waste effort onto checkstyle anymore.

Pretenting to have 100% quality and not passing prone and PMD is a very good and hilarious joke, but not some real considerable software.

PR #16430, opened 2025-02-26
`RSPEC-S1144: UnusedLocalMethodCheck`

punkratz312:
## Skeleton for: [checkstyle/checkstyle#16375](https://github.com/checkstyle/checkstyle/issues/16375)

I don't know how to implement this. I thought this should be quite simple as it's kind of copy-and-paste from another `UnusedLocalVariableCheck`. We could cover the happy path, and the method overloading stuff can come later—if it's even possible, as the types seem to be absent at runtime. But since this is only some edge case, we could still find the majority of the unused stuff dangling around.

At least the tests are working, and I gained knowledge on how to integrate a new rule. So the skeleton is there; it just needs a pro user to adjust the implementation. Thanks!

[RSPEC-S1144](https://sonarsource.github.io/rspec#/rspec/S1144)


punkratz312:
@romani This first draft seems to work, but the test does not behave as expected.  
Even though the implementation is correct when manually debugging, it appears that the tests somehow execute an old state of the code.  

Strangely, even when passing `5` as expected in the error, the log indicates that the expectation is empty: 

image

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedPrivateMethodCheck.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public void beginTree(DetailAST root) {
+        declaredMethods.clear();
+        methodCalls.clear();
+    }
+
+    @Override
+    public void visitToken(DetailAST ast) {
+        if (ast.getType() == TokenTypes.METHOD_DEF) {
+            declaredMethods.add(ast.findFirstToken(TokenTypes.IDENT).getText());
```
punkratz312, line 62: only check for privat method

src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java
```diff
@@ -540,9 +540,9 @@ private static void verifyViolations(String file, List<TestInputViolation> testI
         final List<Integer> expectedViolationLines = testInputViolations.stream()
                 .map(TestInputViolation::getLineNo)
                 .collect(Collectors.toUnmodifiableList());
-        assertWithMessage("Violation lines for %s differ.", file)
-                .that(actualViolationLines)
-                .isEqualTo(expectedViolationLines);
+//        assertWithMessage("Violation lines for %s differ.", file)
+//                .that(actualViolationLines)
+//                .isEqualTo(expectedViolationLines);
```
punkratz312, line 545:
```suggestion
        assertWithMessage("Violation lines for %s differ.", file)
                .that(actualViolationLines)
                .isEqualTo(expectedViolationLines);
```

PR #16438, opened 2025-02-27
Issue #14019: Kill mutation for getMissedHtmlTag

ahmedyoussefg:
For #14019 

# Mutation

```xml
  <mutation unstable="false">
    <sourceFile>JavadocDetailNodeParser.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.JavadocDetailNodeParser</mutatedClass>
    <mutatedMethod>getMissedHtmlTag</mutatedMethod>
    <description>removed call to java/util/Deque::pop</description>
    <lineContent>htmlTagNameStart = stack.pop();</lineContent>
  </mutation>

```

# Modules

- `JavadocDetailNodeParser`

# Steps

- First, I hardcoded the mutation
- I ran `mvn clean verify` and all tests passed
- I generated the pitest report and there were no new surviving mutations
- I generated the diff report for `JavadocChecks` and there were no diff

# Analysis
- After hardcoding the mutation, I analyzed its behaviour and the impact on the code:
image

1. Removal of `!stack.isEmpty()` check:
    1. I’ve reviewed the scenarios where the token type will be `HTML_TAG_NAME`, the stack is empty, and the previous token type is not `START`: 
        - For example, in cases like `</test>`, the method `getMissedHtmlTag()` is not invoked because `recognitionEx.getCtx()` is **not** an instance of `JavadocParser.HtmlTagContext`.
    - This suggests that removing the `stack.isEmpty()` check does not introduce incorrect behavior under currently handled scenarios.
    - **However**, there might still be an unhandled edge case (in both codes). @romani , maybe this could be an issue?
        - If a user writes something like this `<test test="hello">content` in Javadoc (attribute name is same as tag name), it will crash the program. so maybe we should add `&& prevToken == JavadocTokenTypes.SLASH`?
image (1)

2. Removal of `htmlTagNameStart` , because `stack.top()` will always provide the same logic (which is the latest unclosed html tag).

Attention:
updated method is used only in when parse exception happening, so it is not that easy to reproduce, and behavior on excpeption is not deterministic and users unlikely to be affected.
```
        catch (ParseCancellationException | IllegalArgumentException exc) {
            ParseErrorMessage parseErrorMessage = null;

            if (exc.getCause() instanceof FailedPredicateException
                    || exc.getCause() instanceof NoViableAltException) {
                final RecognitionException recognitionEx = (RecognitionException) exc.getCause();
                if (recognitionEx.getCtx() instanceof JavadocParser.HtmlTagContext) {
                    final Token htmlTagNameStart = getMissedHtmlTag(recognitionEx); <!----- here is target merthod
                    parseErrorMessage = new ParseErrorMessage(
                            errorListener.offset + htmlTagNameStart.getLine(),
                            MSG_JAVADOC_MISSED_HTML_CLOSE,
                            htmlTagNameStart.getCharPositionInLine(),
                            htmlTagNameStart.getText());
                }
            }
```

### DIFF
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -441,7 +441,7 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/JavadocDetailNodeParser.java</fileName>
     <specifier>dereference.of.nullable</specifier>
     <message>dereference of possibly-null reference stack.peek()</message>
-    <lineContent>if (stack.peek().getText().equals(token.getText())) {</lineContent>
+    <lineContent>&amp;&amp; stack.peek().getText().equals(token.getText())) {</lineContent>
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
--- a/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
+++ b/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
@@ -1,15 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
   <mutation unstable="false">
-    <sourceFile>JavadocDetailNodeParser.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.JavadocDetailNodeParser</mutatedClass>
-    <mutatedMethod>getMissedHtmlTag</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to java/util/Deque::pop</description>
-    <lineContent>htmlTagNameStart = stack.pop();</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
     <sourceFile>TreeWalker.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.TreeWalker</mutatedClass>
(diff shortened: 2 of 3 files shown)


romani:
>should I analyse the code more to understand why hardcoding is OK?

Yes, pitest killer of useless code. But we should think twice before removing.

ahmedyoussefg:

ahmedyoussefg:
@romani  I've updated the description to clarify why the mutation is okay, please review when you have time.

romani:
>However, there might still be an unhandled edge case (in both codes). @romani , maybe this could be an issue?
    If a user writes something like this <test test="hello">content in Javadoc (attribute name is same as tag name), it will crash the program. so maybe we should add && prevToken == JavadocTokenTypes.SLASH?

yes, please update code to be robust, when all is bad we need to throw regular Chekcstyle Parser exception and nothing else, or ideally let Checks to run is they can.

ahmedyoussefg:
> > However, there might still be an unhandled edge case (in both codes). @romani , maybe this could be an issue?
> > If a user writes something like this content in Javadoc (attribute name is same as tag name), it will crash the program. so maybe we should add && prevToken == JavadocTokenTypes.SLASH?
> 
> yes, please update code to be robust, when all is bad we need to throw regular Chekcstyle Parser exception and nothing else, or ideally let Checks to run is they can.

@romani, would it be preferable to address this in a separate issue and PR?

romani:
yes, I did not noticed that you reproduced this excpeption on released version, so it is not related to your code changes.
please create new issue on excpetion.

romani (approved):
thanks a lot for hard work and investigation on this.

PR #16470, opened 2025-03-04
Issue #14019: kill mutation of SarifLogger readResource Method

Harsh-Agrawal96:
Issue #14019

Kill mutation for `SarifLogger` `NonVoidMethodCallMutator on final String version = SarifLogger.class.getPackage().getImplementationVersion();`

**Reason for Mutation:**
On `readResource` method under `SarifLogger.java`, we are writing the `buffer` on `result` variable in some `n iterations of while loop` but by the mutation it makes the `initial assignment` of `length` variable as `0` and this will not affect the `result` variable so all test cases passed.

**Kill mutation:**
Makes the `initial assignment` of `length` variable as `0` by own. By this, all logic remains as usual, and it can't change because it will affect the `result` variable, and test cases will fail.

### DIFF
--- a/config/pitest-suppressions/pitest-common-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-suppressions.xml
@@ -19,15 +19,6 @@
   </mutation>
 
   <mutation unstable="false">
-    <sourceFile>SarifLogger.java</sourceFile>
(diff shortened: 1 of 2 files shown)


Harsh-Agrawal96:
@romani or someone please review it, waiting for response.

PR #16492, opened 2025-03-07
Issue #14019: Kill mutation for getImplementationVersion in runCli

SteLeo1602:
Issue: #14019

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Main.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Main.java
@@ -121,7 +121,7 @@ public final class Main {
         try {
             final ParseResult parseResult = commandLine.parseArgs(args);
             if (parseResult.isVersionHelpRequested()) {
-                System.out.println(getVersionString());
+                printVersionToSystemOutput();
             }
             else if (parseResult.isUsageHelpRequested()) {
                 commandLine.usage(System.out);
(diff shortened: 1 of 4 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/Main.java
```diff
@@ -156,13 +156,24 @@ else if (parseResult.isUsageHelpRequested()) {
...
     }
 
+    /**
+     * Prints version string when the user requests version help (--version or -V).
+     *
+     * @noinspection UseOfSystemOutOrSystemErr
+     * @noinspectionreason UseOfSystemOutOrSystemErr - driver class for Checkstyle requires
+     *      usage of System.out and System.err
+     */
+    private static void printVersionToSystemOutput() {
```
romani, line 166: Confirmed moving out as this version method doesn't work without jar. We need to exclude it from pitest.

src/main/java/com/puppycrawl/tools/checkstyle/Main.java
```diff
@@ -330,8 +341,6 @@ else if (hasSuppressionLineColumnNumber) {
             }
             if (LOG.isDebugEnabled()) {
                 LOG.debug("Checkstyle debug logging enabled");
-                LOG.debug("Running Checkstyle with version: "
-                        + Main.class.getPackage().getImplementationVersion());
```
romani, line 334: Not working without jar and no real benefits to have it not possible to cover in pitest.
We prefer pitest over this debug log.

romani (approved):
Good to merge

Issue #16233, opened 2025-01-25
test execution is failing on non EN locales


romani:
test execution is failing on non EN locales:
```
$ ./.ci/validation.sh test-de
....
[ERROR] Failures: 
[ERROR]   PackageObjectFactoryTest.testCreateModuleWithTryInAllRegisteredPackages:537 
Invalid exception message
value of              : getMessage()
expected to start with: Unable to instantiate 'PackageObjectFactoryTest$MockClass' class, 
it is also not possible to instantiate it as com.puppycrawl.tools.checkstyle.PackageObjectFactoryTest$MockClass, 
PackageObjectFactoryTest$MockClassCheck, 
com.puppycrawl.tools.checkstyle.PackageObjectFactoryTest$MockClassCheck

but was               : Fehler beim Instanziieren der Klasse 'PackageObjectFactoryTest$MockClass'. 
Es war auch nicht möglich, sie als com.puppycrawl.tools.checkstyle.PackageObjectFactoryTest$MockClass,
 PackageObjectFactoryTest$MockClassCheck, com.puppycrawl.tools.checkstyle.PackageObjectFactoryTest$MockClassCheck
 zu instanziieren. Bitte prüfen Sie, dass der Klassenname als kanonischer 
Name angegeben wurde oder lesen Sie, wie Sie abgekürzte Namen konfigurieren:
 https://checkstyle.org/config.html#Packages. Bitte prüfen Sie auch, dass der angegebene 
ClassLoader des Checkers richtig konfiguriert ist.
[INFO] 
[ERROR] Tests run: 5142, Failures: 1, Errors: 0, Skipped: 53

```

Lalithsaireddy:
To set up the default error message to english of all Non-EN Locale . We need to implement @AfterClass and @BeforeClass for PackageObjectFactoryTest (for setuploacle and restorlocale) and to test it public void test testCreateModuleWithTryInAllRegisteredPackages()

romani:
If you see solution , please send PR

romani:
@AmitKumarDeoghoria, please help us here

smita1078:
Can I try this issue with the suggested solution.

> To set up the default error message to english of all Non-EN Locale . We need to implement [@afterclass](https://github.com/afterclass) and @BeforeClass for PackageObjectFactoryTest (for setuploacle and restorlocale) and to test it public void test testCreateModuleWithTryInAllRegisteredPackages()

[...]

PR #16500, opened 2025-03-08
Issue #16233: test execution is failing on non EN locales

AmitKumarDeoghoria:
Solves : #16233
Reverted https://github.com/checkstyle/checkstyle/pull/16234 to verify

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
@@ -42,10 +42,13 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.jupiter.api.Test;
 import org.mockito.MockedStatic;
 import org.mockito.Mockito;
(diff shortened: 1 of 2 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
```diff
@@ -65,9 +68,21 @@
...
 
+    @BeforeClass
+    public static void setupLocale() {
+        defaultLocale = Locale.getDefault();
+        Locale.setDefault(Locale.ENGLISH);
+    }
+
+    @AfterClass
+    public static void restoreLocale() {
+        Locale.setDefault(defaultLocale);
```
romani, line 83: Is this required? To restore original locale .

Our test needs English locale to let messages in Input files to work properly.

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
AmitKumarDeoghoria: Hey @romani , I added the code to restore the original locale to maintain test isolation and prevent any side effects on other test cases that might rely on the system’s default locale. But if we are always using English and there are no dependencies on the system locale now or in the future in other files, we can remove it. So should I remove it?

romani: We use to be locale based .
But after we migrated to Input based tests with messages inside, it becomes impossible, so we stick to EN locale.

Ok , let's keep it, it should not break anything, this should easier to help identify locale problems in other tests.

romani (approved):
Ok to merge thanks a lot

Issue #8807, opened 2020-09-07
SuppressWithPlainTextCommentFilter is slow on files with multiple errors


basinilya:
Check documentation: https://checkstyle.org/config_filters.html#SuppressWithPlainTextCommentFilter

```
> javac BigJava.java
>

> type ..\checkstyle\checkstyle.xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE module PUBLIC "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN" "https://checkstyle.org/dtds/configuration_1_3.dtd">
<!--
    This configuration file was written by the eclipse-cs plugin configuration editor
-->
<!--
    Checkstyle-Configuration: project-local
    Description: none
-->
<module name="Checker">
  <property name="severity" value="warning"/>
  <module name="TreeWalker">
  </module>
  <module name="LineLength">
    <property name="severity" value="warning"/>
    <property name="max" value="120"/>
    <property name="fileExtensions" value="java" />
  </module>
  <module name="SuppressWithPlainTextCommentFilter">
    <property name="offCommentFormat" value="CHECKSTYLE\.OFF\: ([\w\|]+)"/>
    <property name="onCommentFormat" value="CHECKSTYLE\.ON\: ([\w\|]+)"/>
    <property name="checkFormat" value="$1"/>
  </module>
</module>


> type BigJava.java
public interface BigJava {
    int XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX0 = 0;
... full file attached
    int XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX2999 = 0;
}


> set "RUN_LOCALE=-Duser.language=en -Duser.country=US"
> java %RUN_LOCALE% -jar checkstyle-8.36-all.jar -c config.xml BigJava.java
Starting audit...
[WARN] C:\11\eclipse-cs-254-bad\src\BigJava.java:3: Line is longer than 120 characters (found 134). [LineLength]
...
[...]

Issue #12242, opened 2022-10-01
AuditEvents: Allow to Pass File Source


rnveach:
Identified at https://github.com/checkstyle/checkstyle/pull/12151#discussion_r985102182 , 

[Filters](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/Filter.java#L34) have no way to collect the file source from the [AuditEvents](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java#L44-L48) as the class does not carry that information and there is no other interactions to get this information from,

[TreeWalkerAuditEvents](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalkerAuditEvent.java#L35) do actually collect the file source and are able to pass that on to [TreeWalkerFilters](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalkerFilter.java#L35).

We should expand normal AuditEvents to also send the source to be used in filters when it concerns violations.

Harsh-Agrawal96:
I am on it

PR #12151, opened 2022-09-01
Issue #11939: New filter SuppressWithNearbyTextFilter

stoyanK7:
Resolves #11939

---
Diff Regression config: https://gist.githubusercontent.com/stoyanK7/c2a70dcd9dfc7ab1e689e54e3feba577/raw/ae91f59f217a8804d7d07cbbcfc7c8383108b490/config.xml
Diff Regression patch config: https://gist.githubusercontent.com/stoyanK7/934c2f859f7f9f97559bb339b3287deb/raw/edcc70b3257675a04e350f7ca8d614bd3c9cee06/config.xml

PR #16505, opened 2025-03-08
Issue #12242: expand AuditEvents for fileContent

Harsh-Agrawal96:
Issue #12242

Update the `AuditEvent` for fileContent to use in Filters

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/AuditEvent.java
@@ -44,6 +44,8 @@ public final class AuditEvent {
     private final String fileName;
     /** Violation associated with the event. **/
     private final Violation violation;
+    /** The file contents. */
+    private final FileContents fileContents;
 
     /**
      * Creates a new instance.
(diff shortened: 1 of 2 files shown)


Harsh-Agrawal96:
Update the `AuditEvent` for `fileContents`, now where to use this `fileContets`
Is there is any similar way on CS codebase to how to proceed with `fileContents`.

Harsh-Agrawal96:
@romani  please verify
Update the `AuditEvent` for `fileContents`, now where to use this `fileContets`
Is there is any similar way on CS codebase to how to proceed with fileContents.

romani:
@Harsh-Agrawal96 , thanks a lot for update.
Please make CI green and what first usage will be for fileContent to be good example of how valuable is this.
can we get some performance improvement ?

Harsh-Agrawal96:
>issue is identified at https://github.com/checkstyle/checkstyle/pull/12151#discussion_r985102182 , so we should use this new class field in that filter to improve performance (to avoid reload of file from disk).

After fix Issue #8807 we don't need to use `fileContent` in that and `SuppressWithNearbyTextFilter` uses `fileText` So what other places to use fileContent.

romani:
@Harsh-Agrawal96 , I updated issue description.
I still we need it, as it would help us to not to cache of file content, as it is already in memory and event has reference to it.

PR #16449, opened 2025-03-01
Issue #16378 : Validate inputs in config

Kushagra-Sedwal:
Issue #16378 : Validate inputs in config

Draft pr

Issue #16378, opened 2025-02-20
Validate that all Inputs mentioned all default properties in config


romani:
detected at [src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/InputJavadocVariableMethodInnerClass.java](https://github.com/checkstyle/checkstyle/pull/16049/files#diff-8548d219436cb758627c138f8ebff8413f3daa528b6d504350dd96f22b643a35R2)

we have already validation that if user put `(default)` before property value, it is validated against default value in Check.
It is done at https://github.com/checkstyle/checkstyle/issues/15983 .
Validation logic is at InlineConfigParser.java

TODO:
upadate InlineConfigParser.java to validate that all properties are specified in Input file. if all properties are default, all proprties should be mentioned in Input file and all has prefix `(default)`

Kushagra-Sedwal:
hey!! @romani  I would like to work on this issue can you specify the properties to be specified in the InlineConfigParser.java?

romani:
During parsing of Input file config, there is place where we know name of module.
At that place , we should put method that you need to create that loads module metadata file to take list of properties from it, metadata file is not present, we load module java class by Scrapper and take list of properties from it.
After that do validation that all properties mentioned and if value equals to default there should be prefix `(default)`.

Kushagra-Sedwal:
I've created method to load metadata and validate its properties but I am having issue 

<img width="1174" alt="Image" src="https://github.com/user-attachments/assets/3241d318-2c3e-4000-8c0b-58e7f8660fb8" />
this is the method implementation can you review this and help me where I am going wrong

romani:
Please send PR to let me see all details

Kushagra-Sedwal:
> Please send PR to let me see all details

done

kornilov-mr:
It seems like @Kushagra-Sedwal has abandoned this issue. I would like to work on it instead, however there is still one thing i don't understand about this issue. 
@romani if i were to implement the validation (that all properties are specified), would it also mean that every test that does use inlineParser without all properties would fail? That is exactly what happened with Kushagra-Sedwal suggestion (who got  [2991 errors](https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=26519&view=logs&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111) on test )

[...]

PR #16507, opened 2025-03-09
Issue #16378: Updated InlineConfigParser to validate that all properties are specified

kornilov-mr:
Issue #16378: Updated InlineConfigParser.java to validate that all properties are specified in a test file

The same solution as in this [pull request](https://github.com/checkstyle/checkstyle/pull/16449), except it passes validation.
At least this validation locally:
```bash
./.ci/validation.cmd run_checkstyle
```

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java
@@ -49,6 +49,9 @@ import com.puppycrawl.tools.checkstyle.ConfigurationLoader;
 import com.puppycrawl.tools.checkstyle.PropertiesExpander;
 import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
 import com.puppycrawl.tools.checkstyle.api.Configuration;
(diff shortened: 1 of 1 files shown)


Kushagra-Sedwal:
Hello @kornilov-mr I had not abandoned the issue but since you have started working on this I'd like tell you that the java file also requires validation if the default properties have prefix (default)

romani (changes requested):
We need suppression list.
Suppression should be until new issue that we need to create

kornilov-mr:
@romani Can you please elaborate on what do you mean by suppression list?
Because the only thing i can think of by now, Is making some .xml with all tests that should be skipped by maven in the test phase, which seems to me too radical. though i think there is a different approach, that i am missing

romani:
Can we make it same as https://github.com/checkstyle/checkstyle/blob/894538171c7f3f1f2246c51aad3dd583d32fca5e/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L201-L208 by modules/folder?

As I described at https://github.com/checkstyle/checkstyle/issues/16378#issuecomment-2708504378

Issue #6784, opened 2019-05-26
Create integration tests in CI to alarm on breaking compatibility


romani:
Create CI build items to prevent case like https://github.com/checkstyle/checkstyle/issues/3417#issuecomment-495998320

We probably can clone samples repo and use just built snapshot version.

As breaking changes are confirmed to be introduced, we will disable such build items till sevntu update.

Gradle is already added: https://github.com/search?q=repo%3Acheckstyle%2Fcheckstyle+"no-exception-samples-gradle"&type=code 

TODO: maven and Ant samples to add to CI, similar as we did with gradle sample.

romani:
@AmitKumarDeoghoria, please help us with this issue.
It is a bit more challenging than any before, take some time to handle it, pin me when you are stuck.

nrmancuso:
Closed via #16510

Issue #14086, opened 2023-12-01
semaphore CI is failing with FAILED DOWNLOADS for junit-platform-commons-1.10.1.jar 


romani:
failure: https://checkstyle.semaphoreci.com/jobs/6026b7a4-5eca-40c7-b411-8d0286d549fd is stable
details:
```
[ivy:retrieve] 	==== local-maven-2: tried 00:44
[ivy:retrieve] 	  /home/semaphore/.m2/repository/org/junit/platform/junit-platform-commons/1.10.1/junit-platform-commons-1.10.1.jar 00:44
[ivy:retrieve] 		:::::::::::::::::::::::::::::::::::::::::::::: 00:44
[ivy:retrieve] 		::              FAILED DOWNLOADS            :: 00:44
[ivy:retrieve] 		:: ^ see resolution messages for details  ^ :: 00:44
[ivy:retrieve] 		:::::::::::::::::::::::::::::::::::::::::::::: 00:44
[ivy:retrieve] 		:: org.junit.platform#junit-platform-engine;1.10.1!junit-platform-engine.jar 00:44
[ivy:retrieve] 		:: org.opentest4j#opentest4j;1.3.0!opentest4j.jar 00:44
[ivy:retrieve] 		:: org.junit.platform#junit-platform-commons;1.10.1!junit-platform-commons.jar 00:44
[ivy:retrieve] 		::::::::::::::::::::::::::::::::::::::::::::::
```

there might be some damage in `/home/semaphore/.m2/` so download is failing.
I dod not see a way to clean this folder, no caches management in UI.

romani:
we had similar issue at https://github.com/checkstyle/checkstyle/issues/12705

romani:
I tried to not use maven cache .... but issue is the same failure to download artifact 
https://checkstyle.semaphoreci.com/workflows/65ba10a4-7259-4388-b305-d1cbddce09de?pipeline_id=a9dc30f1-cad4-4574-986a-a8ad0e4a51c5
artifact is at maven https://mvnrepository.com/artifact/org.junit.platform/junit-platform-commons/1.10.1

I will send message to semaphore support.

romani:
Activated back, no special fixes, but we are on jdk17 now

PR #16510, opened 2025-03-09
Issue #6784: Create integration tests in CI to alarm on breaking comp…

AmitKumarDeoghoria:
Solves #6784
Reverted: #14087

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -230,6 +230,21 @@ workflows:
           name: "no-exception-samples-gradle"
           image-name: "cimg/openjdk:11.0.19"
           command: "./.ci/no-exception-test.sh no-exception-samples-gradle"
+      - validate-with-maven-script:
+          name: "no-exception-samples-maven"
+          image-name: "cimg/openjdk:11.0.19"
+          command: "./.ci/no-exception-test.sh no-exception-samples-maven"
+
+      - validate-with-maven-script:
+          name: "no-exception-samples-ant"
+          image-name: "circleci/openjdk:11-jdk"
+          command: |
+            export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
+            sudo apt update
+            sudo apt install -y --reinstall ca-certificates-java
+            sudo apt install -y ant
+            sudo update-ca-certificates -f
+            ./.ci/no-exception-test.sh no-exception-samples-ant
 
   idea:
     jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -238,13 +238,7 @@ workflows:
       - validate-with-maven-script:
           name: "no-exception-samples-ant"
           image-name: "circleci/openjdk:11-jdk"
-          command: |
-            export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
-            sudo apt update
-            sudo apt install -y --reinstall ca-certificates-java
-            sudo apt install -y ant
-            sudo update-ca-certificates -f
-            ./.ci/no-exception-test.sh no-exception-samples-ant
+          command: "./.ci/no-exception-test.sh no-exception-samples-ant"
 
   idea:
     jobs:
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -52,8 +52,6 @@ blocks:
                 - .ci/validation.sh verify-regexp-id
                 - .ci/no-exception-test.sh guava-with-google-checks
                 - .ci/no-exception-test.sh guava-with-sun-checks
-                # until https://github.com/checkstyle/checkstyle/issues/14086
-                # - .ci/no-exception-test.sh no-exception-samples-ant
                 - .ci/validation.sh nondex
                 # until https://github.com/checkstyle/checkstyle/issues/9807
                 # - mvn -e --no-transfer-progress clean package -Passembly,no-validations
--- a/.ci/no-exception-test.sh
+++ b/.ci/no-exception-test.sh
@@ -258,6 +258,24 @@ no-exception-samples-gradle)
   removeFolderWithProtectedFiles checkstyle-samples
   ;;
 
+no-exception-samples-maven)
+  CS_POM_VERSION="$(getCheckstylePomVersion)"
+  echo 'CS_POM_VERSION='"${CS_POM_VERSION}"
+  mvn -e --no-transfer-progress -B install -Pno-validations
+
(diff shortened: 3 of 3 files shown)


AmitKumarDeoghoria:
Hey @romani  can you review, If I am on the right direction?

romani:
something is related 
https://github.com/checkstyle/checkstyle/issues/14086

at least we need removal from:
https://github.com/checkstyle/checkstyle/blob/2e45390d4cfbab0981973eff07d6e5c09d4e12d4/.semaphore/semaphore.yml#L55C25-L56

.semaphore/semaphore.yml
```diff
@@ -52,8 +52,7 @@ blocks:
                 - .ci/validation.sh verify-regexp-id
                 - .ci/no-exception-test.sh guava-with-google-checks
                 - .ci/no-exception-test.sh guava-with-sun-checks
-                # until https://github.com/checkstyle/checkstyle/issues/14086
-                # - .ci/no-exception-test.sh no-exception-samples-ant
+                - .ci/no-exception-test.sh no-exception-samples-ant
```
romani, line 55: If we put all to circleci, why we activate it at semaphore?

I think we should just delete such lines

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ jobs:
   run-inspections:
     docker:
       - image: checkstyle/idea-docker:jdk11-idea2022.3.3
-
+    resource_class: medium
```
romani, line 8: Please explain reason

romani (changes requested):
Items

.semaphore/semaphore.yml
AmitKumarDeoghoria: I initially activated Semaphore CI as a fallback while transitioning everything to CircleCI. I wanted to ensure we had a backup in case anything broke and also compare build times and reliability between the two. Now that CircleCI is handling everything smoothly, keeping Semaphore active just adds unnecessary complexity. So, I will clean it now.

.circleci/config.yml
AmitKumarDeoghoria: I added the resource_class to prevent the build from being killed due to insufficient memory or CPU, ensuring resource-intensive tasks like Maven builds have enough resources to complete successfully.

romani: Can you share link to failure in CI for lack of resources

AmitKumarDeoghoria: https://app.circleci.com/pipelines/github/checkstyle/checkstyle/31107/workflows/247ab1f1-55f0-4066-8678-aae46a69c29e/jobs/823011

romani: this kill means nothing special, it might be some cancelation of jobs polisy triggered it.

https://app.circleci.com/pipelines/github/checkstyle/checkstyle?branch=pull%2F16510

please revert such update on sizes, lets give it one more chance. Such projects are super small, no extra resources should be required.

AmitKumarDeoghoria: Done.

romani (approved):
thanks a lot,

ok to merge.

AmitKumarDeoghoria:
@nrmancuso can you approve too please.

.circleci/config.yml
```diff
@@ -230,6 +230,15 @@ workflows:
...
           image-name: "cimg/openjdk:11.0.19"
           command: "./.ci/no-exception-test.sh no-exception-samples-gradle"
+      - validate-with-maven-script:
+          name: "no-exception-samples-maven"
+          image-name: "cimg/openjdk:11.0.19"
+          command: "./.ci/no-exception-test.sh no-exception-samples-maven"
+
+      - validate-with-maven-script:
+          name: "no-exception-samples-ant"
+          image-name: "circleci/openjdk:11-jdk"
```
nrmancuso, line 240: All `circleci/*` images are deprecated, why are we using this?

nrmancuso (changes requested):
Item:

.circleci/config.yml
romani: no warnigns, still works :).

nrmancuso: > no warnigns, still works :).

Yes, there are warnings, and unless we have some valid reason to do so, we should use the newer images since they are hosted by CircleCI themselves and are faster.

We should use only `cimg/*` images

AmitKumarDeoghoria: Hey @nrmancuso  @romani  I tried updating it to cimg/*
But on doing that job is failing with CI link : https://app.circleci.com/pipelines/github/checkstyle/checkstyle/31339/workflows/3090d6dd-dc53-4a99-b3a9-d93b4ae7d9b3/jobs/832790
It was not able to find the image of ant in docker, then I tried installing it with commands  manually but there was some certification cache issue as given in below CI Link : https://app.circleci.com/pipelines/github/checkstyle/checkstyle/31341/workflows/f51bd578-419c-4bc3-a464-4fef413ea727/jobs/832824
Even after reinstalling the certs its not working. I think we should go ahead with the one which is working.

AmitKumarDeoghoria: I tried using cimg/openjdk:11.0.19

nrmancuso: Please try an older one from around the same time that the CircleCI dockerhub image was created.

AmitKumarDeoghoria: @nrmancuso I believe  Cimg does not comes with pre-installed with tools like Ant, only the bare essentials needed for running Java applications. On Installing ant  manually under command section in cimg , there were some certification cache issue. circleci/openjdk may be better in this cases because it comes pre-installed with tools like Ant, Maven, and Gradle, reducing setup time and complexity for builds that rely on them. Please let me know the next steps what should we go forward with.

nrmancuso: >I believe 

This is the beauty of open source: you have no need to believe. You only need to visit the repo where the dockerfiles live and take a look for yourself :)

https://github.com/CircleCI-Public/cimg-openjdk/blob/main/11.0/Dockerfile

It does not appear that `ant` is installed, and after looking at the dockerfile, it will probably be a bit annoying to reinstall the certs and install ant each time. Let us proceed with the deprecated image.

nrmancuso (approved):

.circleci/config.yml
AmitKumarDeoghoria: Agreed. I did Investigate after multiple trials and errors, I think 'believe' would be a wrong word but since I am new to some of the things, I don't want to assert anything If I am not completely sure. But Really Thanks for the guidance and I would keep it in mind from next time. I will get better.

PR #16526, opened 2025-03-11
Issue #11163: Enforce file size on InputLineLengthSimple

Abdelrhmansersawy:
Issue #11163


Abdelrhmansersawy:
@romani 
I noticed that there are two seemingly identical test classes in the Checkstyle repository:  

- `src/test/resources/com/puppycrawl/tools/checkstyle/checks/sizes/linelength/InputLineLengthSimple.java`  
- `src/test/resources/com/puppycrawl/tools/checkstyle/checks/sizes/linelength/InputLineLengthSimple1.java`  

Would you recommend removing `InputLineLengthSimple1.java` to clean up the code, or is there a specific reason for keeping both?

romani:
Let's keep them for now. There are a lot of duplicates in tests , but to some times it is not exactly duplicate and cover some time nuances.
For now we are ok with this.

PR #16533, opened 2025-03-11
Issue #16482: check-performance-regression workflow should test for j…

### LINKED ISSUE #16482: check-performance-regression workflow does not check Javadoc grammar performance
The current [benchmark config](https://github.com/checkstyle/checkstyle/blob/master/config/benchmark-config.xml) uses the `NoCodeInFileCheck`, which prevents Javadoc from being parsed. As a result, Javadoc grammar is not included in the benchmark analysis, leading to incomplete validation of checkstyle parsing performance.

mahfouz72:
closes #16482

### DIFF
--- a/.ci/check-performance-regression.sh
+++ b/.ci/check-performance-regression.sh
@@ -2,19 +2,24 @@
 
 set -e
 
+# Get the baseline seconds and config file from arguments
+if [ "$#" -ne 2 ]; then
+  echo "Missing arguments!"
+  echo "Usage: $0 BASELINE_SECONDS CONFIG_FILE"
+  exit 1
+fi
+
+BASELINE_SECONDS=$1
+CONFIG_FILE=$2
+
 # max difference tolerance in %
 THRESHOLD_PERCENTAGE=10
-# baseline of execution time in seconds
-BASELINE_SECONDS=384.02
-
 # JDK version
-JDK_VERSION=17
+JDK_VERSION=21
 # sample project path
 SAMPLE_PROJECT="./.ci-temp/jdk$JDK_VERSION"
 # suppression file
 SUPPRESSION_FILE="./config/projects-to-test/openjdk$JDK_VERSION-excluded.files"
-# config file
-CONFIG_FILE="./config/benchmark-config.xml"
 
 # execute a command and time it
 # $TEST_COMMAND: command being timed
@@ -42,7 +47,7 @@ execute_benchmark() {
 
   # add suppressions to config file
   sed -i "/  <!-- Filters -->/r $SUPPRESSION_FILE" \
-        $CONFIG_FILE
+        "$CONFIG_FILE"
 
   for ((i = 0; i < NUM_EXECUTIONS; i++)); do
     local CMD=(java -jar "$JAR_PATH" -c "$CONFIG_FILE" \
--- a/.github/workflows/check-performance-regression.yml
+++ b/.github/workflows/check-performance-regression.yml
@@ -22,7 +22,22 @@ concurrency:
 jobs:
   test:
     if: github.repository == 'checkstyle/checkstyle'
+    strategy:
+      matrix:
+        target:
+          - java
+          - javadoc
+        include:
+          - target: java
+            BASELINE_SECONDS: 445.76
+            CONFIG_FILE: './config/benchmark-config.xml'
+          - target: javadoc
+            BASELINE_SECONDS: 570.62
+            CONFIG_FILE: './config/benchmark-javadoc-config.xml'
     runs-on: ubuntu-latest
+    env:
+      BASELINE_SECONDS: ${{ matrix.BASELINE_SECONDS }}
+      CONFIG_FILE: ${{ matrix.CONFIG_FILE }}
     steps:
       - name: Set up JDK 11
         uses: actions/setup-java@v4
@@ -33,11 +48,11 @@ jobs:
       - name: Checkout Pull Request Code
         uses: actions/checkout@v4
 
-      - name: Clone JDK 17 Repo
+      - name: Clone JDK 21 Repo
         uses: actions/checkout@v4
         with:
-          repository: openjdk/jdk17
-          path: ./.ci-temp/jdk17
+          repository: openjdk/jdk21
+          path: ./.ci-temp/jdk21
 
       - name: Setup local maven cache
         uses: actions/cache@v4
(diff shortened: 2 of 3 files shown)


mahfouz72:
```
Execution Time Baseline: 400 s
Average Execution Time: 528.08 s
Execution Time Difference: 32.0200%
Difference exceeds the maximum allowed difference (32.0200%      > 10%)!
```

A baseline for javadoc should be 530

.github/workflows/check-performance-regression.yml
```diff
@@ -22,7 +22,16 @@ concurrency:
...
+    strategy:
+      matrix:
+        grammar:
+          - java
+          - javadoc
     runs-on: ubuntu-latest
+    env:
+      BASELINE_SECONDS: ${{ matrix.grammar == 'java' && 384.02 || 530 }}
+      CONFIG_FILE: ${{ matrix.grammar == 'java' && './config/benchmark-config.xml'
+                                    || './config/benchmark-javadoc-config.xml' }}
```
nrmancuso, line 34: Please do something like:
```suggestion
strategy:
  matrix:
    target:
      - java
      - javadoc
    include:
      - target: java
        BASELINE_SECONDS: 384.02
        CONFIG_FILE: './config/benchmark-config.xml'
      - target: javadoc
        BASELINE_SECONDS: 530
        CONFIG_FILE: './config/benchmark-javadoc-config.xml'
    runs-on: ubuntu-latest
    env:
       BASELINE_SECONDS: ${{ matrix.BASELINE_SECONDS }}
       CONFIG_FILE: ${{ matrix.CONFIG_FILE }}
```

nrmancuso (changes requested):
@mahfouz72 are we able to bump the jdk version?

Items:

nrmancuso:
@mahfouz72 can you test this easily in your fork?

.github/workflows/check-performance-regression.yml
mahfouz72: Done

mahfouz72:
> are we able to bump the jdk version?

@nrmancuso Done, time increased by 16%

```
Execution Time Baseline: 384.02 s
Average Execution Time: 445.76 s
Execution Time Difference: 16.0700%
Difference exceeds the maximum allowed difference (16.0700%      > 10%)!
```

Javadoc run was canceled without outputs. I will rerun

@nrmancuso I assume there is no need to test on the fork and we can prove it is working here right?

mahfouz72:
Javadoc after bumping Jdk version

```
Execution Time Baseline: 530 s
Average Execution Time: 570.62 s
Execution Time Difference: 7.6600%
Difference is within the maximum allowed difference (7.6600%      <= 10%).
```

nrmancuso:
> @nrmancuso I assume there is no need to test on the fork and we can prove it is working here right?

Yeah, for some reason I thought we never ran workflows from a contributors fork

nrmancuso:
@rnveach please review

config/benchmark-javadoc-config.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+  "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
+<module name="Checker">
+  <!-- Filters -->
+
```
romani, line 6: remove this two lines

config/benchmark-javadoc-config.xml
```diff
@@ -0,0 +1,12 @@
...
+  <property name="severity" value="ignore"/>
+  <module name="TreeWalker">
+    <module name="JavadocBlockTagLocationCheck">
+    </module>
```
romani, line 10: `<module name="JavadocBlockTagLocationCheck"\>`

romani (changes requested):
minors:

config/benchmark-javadoc-config.xml
mahfouz72: This is needed because the sed command uses `<!-- Filters -->` as a search pattern to insert suppressions after this line


https://github.com/checkstyle/checkstyle/blob/d1f9cbdd15defda8acfc8da4887ee73c1adf42dd/.ci/check-performance-regression.sh#L48-L50

config/benchmark-javadoc-config.xml
mahfouz72: done

romani (approved):
ok to merge from me

.ci/check-performance-regression.sh
```diff
@@ -2,19 +2,24 @@
...
+CONFIG_FILE=$2
+
 # max difference tolerance in %
 THRESHOLD_PERCENTAGE=10
-# baseline of execution time in seconds
-BASELINE_SECONDS=384.02
-
 # JDK version
-JDK_VERSION=17
+JDK_VERSION=21
```
rnveach, line 18: Maybe strange question, would it make more sense to just pull the JDK version from the active Java version? I know they aren't interconnected, but would be one less thing to remember to update and keep in synch.

rnveach (approved):

.ci/check-performance-regression.sh
mahfouz72: I don't get it why do we need to keep this variable and active java version in synch? They are not connected as you said we should test performance on new grammer like jdk21 but we are on jdk 11. Can you clarify more I am not sure I understand

rnveach: It would just be to reduce the number of areas we need to update or remember to update.

By keeping them in synch we could just update one and the other will automatically switch over and we won't have to do anything more (as long as CI is green). Most times we have multiple places to update, we forget to update something nested deeper and just rely on first things we find and a green CI. Green CI doesn't always mean everything is updated if we don't have double validation. This is what started our XDoc validation. XDocs easily got out of date when we updated checks as there were no double checks that things were kept in synch.

It was just an idea, thats why I didn't put it as change needed or anything.

rnveach: Thinking some more, this could cause more issues. We may get different results on local runs if we rely on execution environment. However, this is environment specific since we time results, so this whole check may have issues with local running.

mahfouz72:
I think it just needs a restart unrelated failure

rnveach:
@nrmancuso  CI is fully green.

PR #16561, opened 2025-03-14
Issue #11163: Enforce file size on LocalVariableName

Mahmoud-Ayman-Saleh:
Issue #11163: Enforce file size on LocalVariableName


romani (approved):
thanks a lot

Issue #16361, opened 2025-02-18
SarifLoggerTest.java to use verifyWithInlineConfigParserAndLogger


romani:
After merge of https://github.com/checkstyle/checkstyle/pull/16314

We have ability to update all/most test methods in SarifLoggerTest.java to use verifyWithInlineConfigParserAndLogger (this method should be created to accept any logger)

convert usage of `verifyContent` to `verifyWithInlineConfigParserAndLogger` or similar method

Update is required for:
- [ ] [sarif](https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java)
- [ ] [output](https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/DefaultLoggerTest.java)

Please send PR for each test method separately, to ease review.

Example of update https://github.com/checkstyle/checkstyle/pull/16698/files

Goal: use it as much as possible, ideally in all test methods.

Harsh-Agrawal96:
I am working on this

ElinaZoldnere:
I'd like to try this.

romani:
@ElinaZoldnere , thanks a lot.
We missed you a lot. Thanks a lot to be back.

romani:
Initial PR is merged to have method that use Sarif for output. Further migration of test methods is expected in next PRs.

ElinaZoldnere:
@romani Regarding `DefaultLoggerTest` - there are many test cases not based on violations, but rather on exceptions occuring during checking process. In real life these seems quite specific cases - getting an exception on compilable java file using valid configuration. At least I do not have ideas beyond feeding some invalid file name directly to the checker.  (Such tests woudn't be so nice as with sarif, but likely can be done.) 
Could you comment on what should be the focus here?

romani:
https://github.com/checkstyle/checkstyle/blob/9369673bb8023556b87e8a01ac51508286de6daf/src/test/java/com/puppycrawl/tools/checkstyle/DefaultLoggerTest.java#L122

Looks like this testing exception during execution of checkstyle. We can reproduce it by Input that is not compile able java class, and create exception on parsing.

ElinaZoldnere:
@romani 
Had a question. There are already about 150 LOC specifically related to loggers in `AbstractModuleTestSupport`. Would it make sense to move them into a dedicated extension like `AbstractLoggerTestSupport`?
[...]

PR #16584, opened 2025-03-16
Issue 16361: Update SarifLoggerTest.java to use verifyWithInlineConfigParserAndLogger

ElinaZoldnere:
Issue #16361 

- created a method `verifyWithInlineConfigParserAndLogger`  in `AbstractModuleTestSupport` for loggers with one output stream. One test case in `SarifLoggerTest` is converted to the new approach as an example.

### DIFF
--- a/config/checkstyle-non-main-files-checks.xml
+++ b/config/checkstyle-non-main-files-checks.xml
@@ -57,7 +57,7 @@
     <property name="id" value="lineLength"/>
     <!-- catch lines above 100 symbols -->
     <property name="format"
-             value="^(?!(.*value=.*|.*href=|.*http(s)?:|import |(.* )?package |.* files=|.*\.dtd| \* \{@code| \* com\.)).{101,}$"/>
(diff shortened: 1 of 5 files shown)


ElinaZoldnere:
@romani 
Please, see the direction. I am not sure, if I like it, especially having ByteArrayOutputStream as a separate argument in the abstract method alongside the logger.

romani:
Direction is good. 
Let's finish PR with update only of one test method, after merge of first PR, migration will be separate effort, and could be done in parallel.

ElinaZoldnere:
Please see the solution based on sarif test requirements.

For DefaultLoggerTest, a separate method is likely needed, with two output streams and two corresponding expected output files (not created yet). I thought mixing everything into a single method would make it unnecessarily complex for cases where only a single stream is needed, like sarif.

romani (approved):
That a lot, I appreciate that you keep PRs small . It helps me a lot.

PR #16586, opened 2025-03-17
Issue #43: Javadoc should be detected above block comment

### LINKED ISSUE #43: JavadocMethod: Javadoc Not Detected Above Multiline Comments
Check: http://checkstyle.sourceforge.net/config_javadoc.html#JavadocMethod

When programming in Java with JML annotations (http://www.jmlspecs.org/), the typical convention is to place the JML annotations - which are, for the purposes of any non-JML-aware tools, specially formatted Java comments - directly above the method they are annotating. 

A Javadoc comment also needs to be placed above the method it is documenting. Most Javadoc processing tools (including "javadoc" itself with the standard doclet) have no problem ignoring non-Javadoc comments that occur between the Javadoc comment for a method and the method declaration. However, Checkstyle will fail to detect the Javadoc comment for a method if there is a multi-line (but not a single-line) comment between the Javadoc comment and the method declaration. For example, Checkstyle correctly detects the Javadoc in this case:

```
/**
 * A Javadoc comment.
 */
//@ A JML Annotation
public void foo() {}
```

but not in this case:

```
/**
 * A Javadoc comment.
 */
/*@ A JML Annotation */
public void foo() {} 
```

It would be very helpful for Checkstyle to be able to detect such Javadoc comments. 

Examples of JML JavaDoc: https://sourceforge.net/apps/trac/jmlspecs/browser/TeachingMaterials/trunk/CAV2007Tutorial/examples/Actor.java?rev=2675 (https://sourceforge.net/apps/trac/jmlspecs/wiki/TeachingMaterials)
## 

This change impacts the following checks:
UnusedImports
JavadocMethod
JavadocStyle
JavadocType
JavadocVariable
MissingJavadocMethod
MissingJavadocType
WriteTag

mahfouz72:
closes #43

Diff Regression config: https://gist.githubusercontent.com/mohitsatr/f5a524a06294d907c84b9fc5b68661dc/raw/604872961df92db35e66c26218dc74c61bb10097/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java
@@ -216,12 +216,33 @@ public final class FileContents implements CommentListener {
         // Lines start at 1 to the callers perspective, so need to take off 2
         int lineNo = lineNoBefore - 2;
 
-        // skip blank lines
-        while (lineNo > 0 && (lineIsBlank(lineNo) || lineIsComment(lineNo))) {
+        // skip blank lines and comments
+        while (lineNo > 0 && (lineIsBlank(lineNo) || lineIsComment(lineNo)
+                            || lineInsideBlockComment(lineNo + 1))) {
             lineNo--;
         }
 
         return javadocComments.get(lineNo);
+    }
+
+    /**
+     * Checks if the specified line number is inside a block comment.
+     * This method scans through all block comments (excluding Javadoc comments)
+     * and determines whether the given line number falls within any of them
+     *
+     * @param lineNo the line number to check
+     * @return {@code true} if the line is inside a block comment (excluding Javadoc comments)
+     *          , {@code false} otherwise
+     */
+    private boolean lineInsideBlockComment(int lineNo) {
+        final Collection<List<TextBlock>> values = clangComments.values();
+        return values.stream()
+            .flatMap(List::stream)
+            .filter(comment -> !javadocComments.containsValue(comment))
+            .anyMatch(comment -> {
+                return lineNo >= comment.getStartLineNo()
+                                        && lineNo <= comment.getEndLineNo();
+            });
     }
 
     /**
--- a/src/test/java/com/puppycrawl/tools/checkstyle/api/FileContentsTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/api/FileContentsTest.java
@@ -202,11 +202,11 @@ public class FileContentsTest {
                 .that(fileContents.getJavadocBefore(4).toString())
                 .isEqualTo(new Comment(new String[] {"/** A */"}, 0, 1, 7).toString());
         assertWithMessage("Invalid comment")
-                .that(fileContents.getJavadocBefore(5))
-                .isNull();
+                .that(fileContents.getJavadocBefore(5).toString())
+                .isEqualTo(new Comment(new String[] {"/** A */"}, 0, 1, 7).toString());
         assertWithMessage("Invalid comment")
-                .that(fileContents.getJavadocBefore(6))
-                .isNull();
+                .that(fileContents.getJavadocBefore(6).toString())
+                .isEqualTo(new Comment(new String[] {"/** A */"}, 0, 1, 7).toString());
     }
 
     @Test
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocmethod/InputJavadocMethodAboveComments.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocmethod/InputJavadocMethodAboveComments.java
@@ -1,0 +1,102 @@
+/*
+JavadocMethod
+validateThrows = true
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.javadoc.javadocmethod;
+
+public class InputJavadocMethodAboveComments {
+
+    /**
+     * A Javadoc comment.
+     * @return 0
+     */
+    //@ A JML Annotation
+    public int foo() throws Exception { // violation, 'Expected @throws tag for 'Exception''
+        return 0;
+    }
+
+    /**
+     * A Javadoc comment.
+     * @return 0
+     */
+    /*@ A JML Annotation */
+    public int foo2() throws Exception { // violation, 'Expected @throws tag for 'Exception''
+        return 0;
+    }
+
+    /**
+     * A dangling Javadoc comment.
+     * @throws Exception exception
+     * @return 0
+     */
+    /**
+     * A Javadoc comment.
+     * @throws Exception exception
+     * @return 0
+     */
+    public int foo3() throws Exception {
+        return 0;
+    }
+
+    /**
+     * A dangling Javadoc comment.
+     * @throws Exception exception
+     * @return 0
+     */
+    /**
+     * A Javadoc comment.
+     * @return 0
+     */
+    public int foo4() throws Exception {  // violation, 'Expected @throws tag for 'Exception''
+        return 0;
+    }
+
+    /**
+     * A Javadoc comment.
+     * @return 0
+     */
+    @SuppressWarnings("unchecked")
+    /* generic type warnings are suppressed for this method
+    because of reason xyz */
+    public int method() throws Exception { // violation, 'Expected @throws tag for 'Exception''
+        return 0;
+    }
+
+    /**
+     * A Javadoc comment.
+     * @return 0
+     */
+    /* generic type warnings are suppressed for this method
+    because of reason xyz */
+    @SuppressWarnings("unchecked")
+    public int method2() throws Exception {  // violation,'Expected @throws tag for 'Exception''
+        return 0;
+    }
+
+    /**
+     * A Javadoc comment.
+     * @return 0
+     */
+    // generic type warnings are suppressed for this method because of reason xyz
+    @SuppressWarnings("unchecked")
+    public int method3() throws Exception { // violation, 'Expected @throws tag for 'Exception''
+        return 0;
+    }
+
+    /**
+     * A Javadoc comment.
+     * @return 0
+     */
+    @SuppressWarnings("unchecked")
+    // generic type warnings are suppressed for this method because of reason xyz */
+    public int method4() throws Exception { // violation, 'Expected @throws tag for 'Exception''
+        return 0;
+    }
+
+    /*@ A JML Annotation */
+    public int foo5() {
+        return 0;
+    }
+}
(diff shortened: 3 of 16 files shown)


mahfouz72:
Github, generate report

nrmancuso:
@mahfouz72 if we have two “Javadocs” above a method, how does this work out?

mahfouz72:
@nrmancuso The closer to the method is the Javadoc to check.

https://github.com/checkstyle/checkstyle/blob/dc6d7f1e1a996cf6c70e80958a4ec97dac9803fd/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocmethod/InputJavadocMethodAboveComments.java#L27-L45

nrmancuso (approved):
Good stuff :)

nrmancuso:
@rnveach please take a look

rnveach:
You certainly found an old issue to look at.

nrmancuso:
> You certainly found an old issue to look at.

I was thinking the same 😂 i don’t think I’ve dealt with anything less than four digits before.

rnveach (commented):
https://github.com/checkstyle/checkstyle/issues/43#issuecomment-569840264
> FileContents.getJavadocBefore and this is used by 8 checks, so each one is affected by this bug

Can you provide us a list so it can be put in the first issue of the post.

src/test/java/com/puppycrawl/tools/checkstyle/api/FileContentsTest.java
```diff
@@ -202,11 +202,8 @@ public void testReportBlockCommentJavadoc() {
                 .that(fileContents.getJavadocBefore(4).toString())
                 .isEqualTo(new Comment(new String[] {"/** A */"}, 0, 1, 7).toString());
         assertWithMessage("Invalid comment")
-                .that(fileContents.getJavadocBefore(5))
-                .isNull();
-        assertWithMessage("Invalid comment")
-                .that(fileContents.getJavadocBefore(6))
-                .isNull();
+                .that(fileContents.getJavadocBefore(5).toString())
+                .isEqualTo(new Comment(new String[] {"/** A */"}, 0, 1, 7).toString());
```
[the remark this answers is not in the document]
rnveach: I am not fully following this comment. Your comment and code says "A", but then the screenshot has "B" and technically the javadoc in the screenshot is "This is Javadoc".

src/test/java/com/puppycrawl/tools/checkstyle/api/FileContentsTest.java
```diff
@@ -202,11 +202,8 @@ public void testReportBlockCommentJavadoc() {
                 .that(fileContents.getJavadocBefore(4).toString())
                 .isEqualTo(new Comment(new String[] {"/** A */"}, 0, 1, 7).toString());
         assertWithMessage("Invalid comment")
-                .that(fileContents.getJavadocBefore(5))
-                .isNull();
-        assertWithMessage("Invalid comment")
-                .that(fileContents.getJavadocBefore(6))
-                .isNull();
```
rnveach, line 209: Also why did we remove 6? It looks like it was for `5, 0, 5, 6`.

src/test/java/com/puppycrawl/tools/checkstyle/api/FileContentsTest.java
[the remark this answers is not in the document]
mahfouz72:
```Java
       final FileContents fileContents = new FileContents(new FileText(new File("filename"),
                Arrays.asList("/** A */", "", "//", "/**/", "/* B */")));
```

In the test, the order of comments indicates that the Javadoc before line 6 is `/** A */`. but the test was incorrect in stating that there was no Javadoc before (because we were not skipping the block comments). After correcting this, I updated the test accordingly.

As proof, I attached a screenshot from IDEA using the same comment order as in the test, which shows that IDEA correctly renders the Javadoc for the method.

To make things identical just replace `/** This is Javadoc */` by `/** A */` in the screenshot.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocmethod/InputJavadocMethodAboveComments.java
```diff
@@ -0,0 +1,91 @@
...
+    /*@ A JML Annotation */
+    public int foo2() { // violation, '@return tag should be present and have description.'
+        return 0;
+    }
+
+    /**
+     * A Javadoc comment.
+     */
+    /**
+     * A Javadoc comment.
```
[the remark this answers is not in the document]
mahfouz72: Done

src/test/java/com/puppycrawl/tools/checkstyle/api/FileContentsTest.java
mahfouz72: It was 1 4 5 6. Looks like I removed 6 by mistake I don't remember. Restored
4, 5, 6 have the same expected output

src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java
```diff
@@ -216,14 +216,35 @@ public TextBlock getJavadocBefore(int lineNoBefore) {
         // Lines start at 1 to the callers perspective, so need to take off 2
         int lineNo = lineNoBefore - 2;
 
-        // skip blank lines
-        while (lineNo > 0 && (lineIsBlank(lineNo) || lineIsComment(lineNo))) {
+        // skip blank lines and comments
+        while (lineNo > 0 && (lineIsBlank(lineNo) || lineIsComment(lineNo)
+                            || lineInsideBlockComment(lineNo + 1))) {
```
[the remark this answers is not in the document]
mahfouz72: Inside this method, I used the clangComments collection and the lineNos in this collection is shifted by 1 from the file text. That's why I added 1.

mahfouz72:
> Can you provide us a list so it can be put in the first issue of the post.

UnusedImports
JavadocMethod
JavadocStyle
JavadocType
JavadocVariable
MissingJavadocMethod
MissingJavadocType
WriteTag

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocmethod/InputJavadocMethodAboveComments.java
```diff
@@ -0,0 +1,91 @@
...
+    @SuppressWarnings("unchecked") // violation, '@return tag should be present.*.'
+    public int method3() {
+        return 0;
+    }
+
+    /**
+     * A Javadoc comment.
+     */
+    @SuppressWarnings("unchecked") // violation, '@return tag should be present.*.'
+    // generic type warnings are suppressed for this method because of reason xyz */
```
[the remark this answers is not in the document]
rnveach: No, I just missed this. Sorry.

src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java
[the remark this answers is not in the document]
rnveach: > I used the clangComments collection and the lineNos in this collection is shifted by 1 from the file text. That's why I added 1.

Is this a bug in our code? I believe some places still have issues with 0 versus 1 in our code.

rnveach (approved):

src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java
```diff
@@ -216,14 +216,35 @@ public TextBlock getJavadocBefore(int lineNoBefore) {
...
+    /**
+     * Checks if the specified line number is inside a block comment.
+     * This method scans through all block comments (excluding Javadoc comments)
+     * and determines whether the given line number falls within any of them
+     *
+     * @param lineNo the line number to check
+     * @return {@code true} if the line is inside a block comment (excluding Javadoc comments)
+     *          , {@code false} otherwise
+     */
+    public boolean lineInsideBlockComment(int lineNo) {
```
romani, line 237: please make it private.

We should not grow our API contract to users.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocmethod/InputJavadocMethodAboveComments.java
```diff
@@ -0,0 +1,91 @@
...
+
+package com.puppycrawl.tools.checkstyle.checks.javadoc.javadocmethod;
+
+public class InputJavadocMethodAboveComments {
+
+    /**
+     * A Javadoc comment.
+     */
+    //@ A JML Annotation
+    public int foo() { // violation, '@return tag should be present and have description.'
```
romani, line 15: from user perspective it is not clear what is considered as javadoc comment. may be `//@ A JML Annotation` ?

```
    /**
     * A Javadoc comment.
     * @return result
     */
    //@ A JML Annotation
    public int foo() throws Exception { // violation, '@throw tag should be present and have description.'
```

let update to this to give more hints that return is found, so javadoc is second comment from method signature.
same for all others.

romani (changes requested):
item:

romani:
> UnusedImports
JavadocMethod
JavadocStyle
JavadocType
JavadocVariable
MissingJavadocMethod
MissingJavadocType

lets add one javadoc above comment to Input of each of this check. as they are affected.

mahfouz72:
The report is already there for all Javadoc checks

https://github.com/checkstyle/checkstyle/pull/16586#issuecomment-2730044420

src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java
mahfouz72: done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocmethod/InputJavadocMethodAboveComments.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * A dangling Javadoc comment.
+     * @return 0
+     */
+    /**
+     * A Javadoc comment.
+     */
+    public int foo4() { // violation, '@return tag should be present and have description.'
```
romani, line 47: Please put exception to all javadoc, to make it clear by return what comment is Javadoc.

Please do for all

romani (changes requested):
Last

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocmethod/InputJavadocMethodAboveComments.java
mahfouz72: done

romani (approved):
Thanks a lot !!!

PR #16587, opened 2025-03-17
Issue #14019: Remove suppression for pitest-ant profile

Rohanraj123:
part of #14019

**Explaination :** 
Since no function was there that returns maxErrors variable directly, I thought of adding getter method for that. But Then I realised that there is no actual dependency on maxErrors field because it was never populated using setter function. It was just for testing purpose. Even in production code processFiles() func. It was useless to have the logic that I removed because maxErrors is always 0 ( default value ). Not sure whether removing the fields and logic is a good practise or not. 
cc @romani 

**Targeted Mutation:** 
https://github.com/checkstyle/checkstyle/blob/afba154a6a5be1ec41accd829170140332759950/config/pitest-suppressions/pitest-ant-suppressions.xml#L183-L190

Diff Regression config: https://gist.githubusercontent.com/Rohanraj123/b4b513e27c22e9a961f16d802f9c22e6/raw/f405dbbe958f1caff54a9aaafd9d6f0056930bc9/gistfile1.txt

Diff Regression projects: https://gist.githubusercontent.com/Rohanraj123/7aaecef3f81122b27a046f3743a1618d/raw/655c618a8911504977a9d59d343c59abd9221341/gistfile1.txt

Report label:Issue#16587-report

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
@@ -356,8 +356,8 @@ public class CheckstyleAntTask extends Task {
         // Handle the return status
         if (!okStatus) {
             final String failureMsg =
-                    "Got " + numErrs + " errors and " + numWarnings
-                            + " warnings.";
+                    "Got " + numErrs + " errors (max allowed: " + maxErrors + ") and "
+                            + numWarnings + " warnings.";
             if (failureProperty != null) {
                 getProject().setProperty(failureProperty, failureMsg);
             }
--- a/config/pitest-suppressions/pitest-ant-suppressions.xml
+++ b/config/pitest-suppressions/pitest-ant-suppressions.xml
@@ -152,13 +152,4 @@
     <description>removed call to org/apache/tools/ant/DirectoryScanner::getBasedir</description>
     <lineContent>logIndex, fileNames.length, scanner.getBasedir()), Project.MSG_VERBOSE);</lineContent>
   </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>CheckstyleAntTask.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask</mutatedClass>
-    <mutatedMethod>setMaxErrors</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable maxErrors</description>
-    <lineContent>this.maxErrors = maxErrors;</lineContent>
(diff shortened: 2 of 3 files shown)


Rohanraj123:
GitHub, generate report

Rohanraj123:
GitHub, generate report

Rohanraj123:
@romani PTAL at PR description.

Rohanraj123:
GitHub, generate report

Rohanraj123:
GitHub, generate report

Rohanraj123:
Github, generate report

Rohanraj123:
Github, generate report

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
```diff
@@ -102,9 +102,6 @@ public class CheckstyleAntTask extends Task {
     /** The name of the properties file. */
     private File properties;
 
-    /** The maximum number of errors that are tolerated. */
-    private int maxErrors;
```
romani, line 106: https://checkstyle.org/anttask.html#Parameters

it is publicly visible paramter, we can not remove it, it is used by users.

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
Rohanraj123: What do you think about [this](https://github.com/checkstyle/checkstyle/pull/16587#issuecomment-2729255811) workaround for testing?

romani: Reflection is absolutely last resort, with big explanation comment above method.

Rohanraj123:
**Issue Summary:**
The private field maxErrors has no getter, that makes direct testing difficult.

The logic checks:
`
final boolean okStatus = numErrs <= maxErrors && numWarnings <= maxWarnings;`
If `numErrs > maxErrors,` an exception is thrown.
**Mutation Testing Concern:**

If the setter logic is removed, maxErrors defaults to 0.

Since numErrs is always at least 1 then an exception is always thrown.
That means in even in the case mutation happens and numErrs is 1. Exception can be thrown. hence It is important to respect the presence of maxErrors var in logic through tests.

Problem: The exception message does not include maxErrors, so we can't verify its impact.

`Testing Approaches:`
Reflection (Not good practise) :  Can access maxErrors, but violates encapsulation.

Modify Exception Message (Preferred) :  Include maxErrors in the failure message, allowing validation in tests.

Rohanraj123:
@romani PTAL

romani (approved):
a way better.

I hope you getting on how to make good tests, without hack to make it covered.

PR #16588, opened 2025-03-17
Issue #14019: kill mutation of encode(event.getMessage()) in XMLLogger

Mahmoud-Ayman-Saleh:
Issue #14019: Kill mutation for XMLLogger ArgumentPropagationMutator on + encode(event.getMessage())

Reason for Mutation: In the writeFileError method of XMLLogger.java, we normally encode the error message to escape special XML characters (like <, >, and &). However, the mutation replaces the call to encode(event.getMessage()) with event.getMessage(), so special characters aren’t escaped. As a result, the XML output may be malformed, yet all test cases pass.

Kill mutation: Simulate the mutation by directly using event.getMessage() in place of encode(event.getMessage()). This change exposes the missing encoding of special characters, causing tests that expect correctly escaped XML to fail

@romani waiting for your review

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
@@ -224,6 +224,13 @@ public class XMLLoggerTest extends AbstractXmlTestSupport {
     }
 
     @Test
+    public void testAddErrorWithEncodedMessage() throws Exception {
+        final String inputFileWithConfig = "InputXMLLoggerEncodedMessage.java";
+        final String expectedXmlReport = "ExpectedXMLLoggerEncodedMessage.xml";
+        verifyWithInlineConfigParserAndXmlLogger(inputFileWithConfig, expectedXmlReport);
+    }
+
+    @Test
     public void testAddErrorOnZeroColumns() throws Exception {
         final XMLLogger logger = new XMLLogger(outStream, OutputStreamOptions.CLOSE);
         logger.auditStarted(null);
--- a/config/pitest-suppressions/pitest-common-suppressions.xml
+++ b/config/pitest-suppressions/pitest-common-suppressions.xml
@@ -8,13 +8,4 @@
     <description>Removed assignment to member variable args</description>
     <lineContent>this.args = null;</lineContent>
   </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>XMLLogger.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.XMLLogger</mutatedClass>
-    <mutatedMethod>writeFileError</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.ArgumentPropagationMutator</mutator>
-    <description>replaced call to com/puppycrawl/tools/checkstyle/XMLLogger::encode with argument</description>
-    <lineContent>+ encode(event.getMessage())</lineContent>
-  </mutation>
 </suppressedMutations>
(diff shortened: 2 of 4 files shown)


Rohanraj123:
Hey @Mahmoud-Ayman-Saleh I think you should follow the steps mentioned [here](https://github.com/checkstyle/checkstyle/issues/12341).

As per the community, changing into tests always better than changing the business logic. And hence we first always try to find tests and then If not found then we can change the business logic. And its better to go step by step. It gets easier!!
Hope it will help!!

src/main/java/com/puppycrawl/tools/checkstyle/XMLLogger.java
```diff
@@ -208,7 +208,7 @@ private void writeFileError(AuditEvent event) {
                 + event.getSeverityLevel().getName()
                 + "\"");
         writer.print(" message=\""
-                + encode(event.getMessage())
+                + event.getMessage()
```
romani, line 211: Find inspiration at this update https://github.com/checkstyle/checkstyle/commit/d11b600d2aad3a8ba1ef76a252024758bbdc5edc

romani (changes requested):
Items

Mahmoud-Ayman-Saleh:
@romani please review, I’m not entirely sure what is causing the CI failure

romani:
https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=27251&view=logs&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111&l=1352

very readable error message.
main point, there was no error before your changes, you changing something and start to fail, try to investgate

Rohanraj123:
> @romani please review, I’m not entirely sure what is causing the CI failure

Hey @Mahmoud-Ayman-Saleh In the CI : 
```
expected: Test XML encoding: <tag> & "quotes" & 'apos'; Existing entities: & <
but was : Test XML encoding: <tag> & "quotes" & apos; Existing entities: &amp; &lt;
```
This shows that expected and actual message was not same. Just change the message according the actual one. You can also check these things on local before sending next push using` mvn clean verify`.

Mahmoud-Ayman-Saleh:
Thanks @Rohanraj123 and @romani for your help :)
 I had already built the project locally many times, and I encountered this error before. I actually tried changing the message, but at that time, the same error kept appearing, even though I modified it. However, I missed a small detail I needed to update something in `ExpectedXMLLoggerEncodedMessage.xml`
There were also some minor issues, like an extra space. I fixed everything, built the project locally successfully, and just pushed the changes. The CI should pass now!

src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
```diff
@@ -223,6 +223,26 @@ public void testAddErrorModuleId() throws Exception {
...
+            "test-module", getClass(),
+            "Test XML encoding: <tag> & \"quotes\" & 'apos'; Existing entities: & <"
+        );
+
+        final AuditEvent ev = new AuditEvent(this, "Test.xml", violation);
+        logger.addError(ev);
+        logger.auditFinished(null);
+
+        // Verify the XML output matches the expected escaped values
+        verifyXml(getPath("ExpectedXMLLoggerEncodedMessage.xml"), outStream);
```
romani, line 243: Very close to final.
But please use verifyWithInlineConfigParserAndXmlLogger

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
Mahmoud-Ayman-Saleh: resolved

romani (approved):
Awesome.

I hope you see how elegant your solution

PR #16590, opened 2025-03-17
Issue #14019: Remove suppression for pitest-ant profile : retrieveAll…

Rohanraj123:
part of #14019 

**Explaination:**
Usage of `scanner.getDir()` is being done twice dynamically but only second usage has been covered by tests. And since both the function calls returns the same value then we can store it once and reuse it everytime we need that will ensure the testing of variable in the second time by test that was testing the s`canner.baseDir()` call the second time. 
This function has not very much of dependencies of other modules hence I dont think it affect other places in the code. 

**Target mutation :**
https://github.com/checkstyle/checkstyle/blob/aac3e77caa63afd4a9c0340688bf486705bbac5f/config/pitest-suppressions/pitest-ant-suppressions.xml#L174-L181

### DIFF
--- a/config/pitest-suppressions/pitest-ant-suppressions.xml
+++ b/config/pitest-suppressions/pitest-ant-suppressions.xml
@@ -174,15 +174,6 @@
   <mutation unstable="false">
     <sourceFile>CheckstyleAntTask.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask</mutatedClass>
-    <mutatedMethod>retrieveAllScannedFiles</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to org/apache/tools/ant/DirectoryScanner::getBasedir</description>
-    <lineContent>logIndex, fileNames.length, scanner.getBasedir()), Project.MSG_VERBOSE);</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>CheckstyleAntTask.java</sourceFile>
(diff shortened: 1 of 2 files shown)


Rohanraj123:
Closing this for now. Will reopen after the previous work!!

Rohanraj123:
@romani Please take a look. Since method is not integrated in any Checks, hence unit test is the only option. Soon open more PRs for Pitest where I will parse input files and apply some checks to test the mutated method.

src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
```diff
@@ -173,6 +173,34 @@ public final void testPathsFileWithLogVerification() throws IOException {
...
+
+        final FileSet fileSet = new FileSet();
+        fileSet.setFile(new File(getPath(FLAWLESS_INPUT)));
+        antTask.addFileset(fileSet);
+
+        antTask.scanFileSets();
+
+        final List<MessageLevelPair> loggedMessages = antTask.getLoggedMessages();
+
+        final String expectedMessage = new File(getPath(".")).getAbsolutePath();
```
romani, line 195: expectedPath

As it is not whole message

romani (changes requested):
Items

Rohanraj123:
Hey @romani , With this PR I think I  have pushed several PRs on PITEST. And would love to move furthur with more general fixes and Issues related to CheckerFramework. So that I can improve my understanding of different Checks and Filters and Checkstyle in general.

Would love to hear from you what are some points to focus on while moving ahead and making my contributions step one level forward.

romani (approved):
Thanks a lot

PR #16595, opened 2025-03-18
Issue #11163: Enforce upper ell file size

ZiadAbdElFatah:
**Issue #11163** 

Refactored InputUpperEllSemantic.java file into some files to reduce it's size

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/UpperEllCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/UpperEllCheckTest.java
@@ -48,7 +48,7 @@ public class UpperEllCheckTest
     public void testWithChecker()
             throws Exception {
         final String[] expected = {
-            "96:40: " + getCheckMessage(MSG_KEY),
+            "29:40: " + getCheckMessage(MSG_KEY),
         };
         verifyWithInlineConfigParser(
                 getPath("InputUpperEllSemantic.java"), expected);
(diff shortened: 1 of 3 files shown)


romani:
Single commit

src/test/java/com/puppycrawl/tools/checkstyle/checks/UpperEllCheckTest.java
```diff
@@ -48,7 +48,7 @@ public void testGetRequiredTokens() {
     public void testWithChecker()
             throws Exception {
         final String[] expected = {
-            "96:40: " + getCheckMessage(MSG_KEY),
+            "29:40: " + getCheckMessage(MSG_KEY),
```
romani, line 51: All new input files should have 
```
/*
UpperEll
```

And added to Test class in special test methods.

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/checks/UpperEllCheckTest.java
ZiadAbdElFatah: @romani 

I added 
```
/*
UpperEll
```

And about the test methods, there is nothing to test in these files except the one I made.

There is only one check on upper ell in **InputUpperEllSemantic.java** file.

romani: Yes, weird.
Just remove all Inputs that are not referenced in Test class.

ZiadAbdElFatah: Ok but all these classes and methods were in InputUpperEllSemantic.java file, I think they are important.

romani: You have two options:
- remove them 
- add them all to individual test methods

ZiadAbdElFatah: @romani 
I removed them and it works fine.

romani (approved):
Thanks a lot

PR #16607, opened 2025-03-19
Issue #14019: Remove suppression for pitest-ant profile : execute()

Rohanraj123:
part of #14019 

**Target Mutation :** 

https://github.com/checkstyle/checkstyle/blob/5495106a50a0fdb3a4f9bee1280c14b4a5417dea/config/pitest-suppressions/pitest-ant-suppressions.xml#L93-L109


Rohanraj123:
@romani PTAL

romani (approved):
Thanks a lot.

Junit testing is unusual than in other parts but it is ok, we will think on better forms later on

Rohanraj123:
> Thanks a lot.
> 
> Junit testing is unusual than in other parts but it is ok, we will think on better forms later on

Am currently looking at the predefined tests, to use those componentsin new tests written.

PR #16614, opened 2025-03-20
Issue #14019: Remove last pitest-utils suppresion for isInBlockOf()

Rohanraj123:
part of #14019

**Explaination** : 
So basically `!returnValue` was being used in the loop logic to ensure that loop ends when we find the required tokenType. Now If mutator mutates this line `!returnValue to true`, Loop will keep on running upto root node. 
Now the problem is that returnValue will always return true if it founds the required tokenType that makes it harder to test.
We can stop the loop from running inside the if condition when we find the correct tokenType using `break` statement directly inside the if conditional, Hence it is a unnecessary code. 

Target Mutation : 
https://github.com/checkstyle/checkstyle/blob/9369673bb8023556b87e8a01ac51508286de6daf/config/pitest-suppressions/pitest-utils-suppressions.xml#L1-L11

Diff Regression config: https://gist.githubusercontent.com/Rohanraj123/cb0a90c5fd0962c02c5035a66e13d625/raw/fa3bb5ce3ac93c607b0de6b69c8f9b698b96a4eb/gistfile1.txt


src/test/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtilTest.java
```diff
@@ -375,6 +375,18 @@ public void testisInInterfaceOrAnnotationBlock() {
                 .isFalse();
     }
 
+    @Test
+    public void testIsInBlockOf() {
+      DetailAST standaloneNode = getNode(TokenTypes.LITERAL_NEW);
+      try {
+          ScopeUtil.isInBlockOf(standaloneNode, TokenTypes.OBJBLOCK);
```
romani, line 382: wedo not allow pure unit tests any more.  
please do coverage by means of some Check and some real compilable Input file.

romani (changes requested):
items:

src/test/java/com/puppycrawl/tools/checkstyle/utils/ScopeUtilTest.java
Rohanraj123: Was thinking for the same. Thanks for the guidance.!!

Rohanraj123:
Github, generate report

Rohanraj123:
Github, generate report

Rohanraj123:
Closing this PR as the fix already merged

PR #16628, opened 2025-03-22
Pull #16628: Try SLL prediction mode for fast javadoc parsing

mahfouz72:
> This is the fastest prediction mode, and provides correct results for many grammars

```Java
public enum PredictionMode {
	/**
	 * The SLL(*) prediction mode. This prediction mode ignores the current
	 * parser context when making predictions. This is the fastest prediction
	 * mode, and provides correct results for many grammars. This prediction
	 * mode is more powerful than the prediction mode provided by ANTLR 3, but
	 * may result in syntax errors for grammar and input combinations which are
	 * not SLL.
	 *
	 * <p>
	 * When using this prediction mode, the parser will either return a correct
	 * parse tree (i.e. the same parse tree that would be returned with the
	 * {@link #LL} prediction mode), or it will report a syntax error. If a
	 * syntax error is encountered when using the {@link #SLL} prediction mode,
	 * it may be due to either an actual syntax error in the input or indicate
	 * that the particular combination of grammar and input requires the more
	 * powerful {@link #LL} prediction abilities to complete successfully.</p>
	 *
	 * <p>
	 * This prediction mode does not provide any guarantees for prediction
	 * behavior for syntactically-incorrect inputs.</p>
	 */
	SLL,
        .....
```
Diff Regression config: https://gist.githubusercontent.com/mahfouz72/33967bf86e4a29cafa429771b0e95808/raw/f621465b3c7dd6e64ac603d396570bf92a1a69d6/check_patch.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/JavadocDetailNodeParser.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/JavadocDetailNodeParser.java
@@ -34,6 +34,7 @@ import org.antlr.v4.runtime.ParserRuleContext;
 import org.antlr.v4.runtime.RecognitionException;
 import org.antlr.v4.runtime.Recognizer;
 import org.antlr.v4.runtime.Token;
+import org.antlr.v4.runtime.atn.PredictionMode;
 import org.antlr.v4.runtime.misc.Interval;
 import org.antlr.v4.runtime.misc.ParseCancellationException;
 import org.antlr.v4.runtime.tree.ParseTree;
@@ -168,6 +169,9 @@ public class JavadocDetailNodeParser {
         final CommonTokenStream tokens = new CommonTokenStream(lexer);
 
         final JavadocParser parser = new JavadocParser(tokens);
+
+        // set prediction mode to SLL to speed up parsing
+        parser.getInterpreter().setPredictionMode(PredictionMode.SLL);
 
         // remove default error listeners
         parser.removeErrorListeners();
--- a/.github/workflows/check-performance-regression.yml
+++ b/.github/workflows/check-performance-regression.yml
@@ -32,7 +32,7 @@ jobs:
             BASELINE_SECONDS: 445.76
             CONFIG_FILE: './config/benchmark-config.xml'
           - target: javadoc
-            BASELINE_SECONDS: 570.62
+            BASELINE_SECONDS: 481.41
             CONFIG_FILE: './config/benchmark-javadoc-config.xml'
     runs-on: ubuntu-latest
     env:
--- a/config/jsoref-spellchecker/whitelist.words
+++ b/config/jsoref-spellchecker/whitelist.words
@@ -1221,6 +1221,7 @@ sivakumar
 Slawinski
 slf
 slist
+SLL
 SLo
 Slurper
 slurpersupport
--- a/pom.xml
+++ b/pom.xml
@@ -4718,6 +4718,14 @@
                 <avoidCallsTo>
                   org.antlr.v4.runtime.BailErrorStrategy
                 </avoidCallsTo>
+                <!-- Setting the prediction mode to SLL is an intentional optimization
+                 to speed up parsing.
+                 Since this is a deterministic change to the parser's behavior,
+                 no normal check execution test case will be able to kill the mutation, as the test
+                 will always pass regardless of the parsing strategy. -->
+                <avoidCallsTo>
+                  org.antlr.v4.runtime.atn.ParserATNSimulator
+                </avoidCallsTo>
               </avoidCallsTo>
               <coverageThreshold>100</coverageThreshold>
               <mutationThreshold>98</mutationThreshold>


mahfouz72:
Interesting!
```
Execution Time Baseline: 570.62 s
Average Execution Time: 474.83 s
Execution Time Difference: 16.7800%
Difference exceeds the maximum allowed difference (16.7800%      > 10%)!
```

mahfouz72:
Since the Javadoc grammar contains many ambiguities and many nested rules, it may not be safe to rely on single lookahead parsing. However, if we can rewrite the grammar to reduce nesting and ambiguity, we could significantly improve parsing time by using SLL parsing

@nrmancuso @rnveach @romani 
What do you think?

rnveach:
This is a 16% deduction right?

What are the downsides of this mode? I assume they didn't enable it by default for a reason.

Just curious if we also seen what this does on the Java side?

mahfouz72:
> This is a 16% deduction right? Maybe consider running the performance multiple times to confirm it is a stable deduction.

Yes, 100 seconds deduction which is 16%. I restarted CI by pushing again to see if this reduction is stable.

> What are the downsides of this mode? I assume they didn't enable it by default for a reason.

From Javadoc of this option:

> If a syntax error is encountered when using the {@link #SLL} prediction mode, it may be due to either an actual syntax error in 
> the input or indicate that the particular combination of grammar and input requires the more  powerful {@link #LL} prediction 
> abilities to complete successfully

> 	 This prediction mode does not provide any guarantees for prediction
	 behavior for syntactically-incorrect inputs.</p>

As per my understanding SLL mode doesn't handle ambiguities well. If a grammar contains multiple possible parse paths for a given input, SLL mode will fail. and It doesn't backtrack, once it hits an ambiguity, it stops. As it works with a single lookahead and not taking the whole parser context into account


> Just curious if we also seen what this does on the Java side?

I didn't try this may ne we should try it in another PR but I think that Java can be more complex than Javadoc which might make it require the full LL approach for correct parsing.  @nrmancuso can confirm this 


We can use SLL mode as the default parsing strategy, and if it encounters any issues or fails to parse, we can switch to LL mode.

mahfouz72:
Github, generate report

rnveach:
> If a syntax error is encountered ... indicate that the particular combination of grammar and input requires the more powerful {https://github.com/link #LL} prediction abilities to complete successfully

I believe we need regression without suppressing grammar failures on all our regression projects. We need to know if this change is now increasing grammar failures.

mahfouz72:
Github, generate report

mahfouz72:
Github, generate report

romani:
if you think this config is frequently to use you can put it as folder to https://github.com/checkstyle/test-configs and re-use at any time.

mahfouz72:
Time decreased by 16% so I reduced the baseline.
The report (without suppressing grammar failures) shows no difference, so parsing is unaffected.
I think it is ok to suppress this surviving mutation as we just use a setter and can't think of a test case that specific to this line

please see if this is good and update is approved I can do similar approach for java parsing

romani:
We are going to zero mutation phase, some most weird  should be covered by pure unit testing, if no reasonable coverage can be provided.

nrmancuso (approved):
Very good, I tried this when I implemented the Java grammar and it did not work at the time.  Maybe some improvements have been made since then, we should at least try it out again.

I’m don’t see any point in holding up this PR over a pitest suppression.

rnveach (approved):

src/main/java/com/puppycrawl/tools/checkstyle/JavadocDetailNodeParser.java
```diff
@@ -112,6 +113,7 @@ public ParseStatus parseJavadocAsDetailNode(DetailAST javadocCommentAst) {
 
         try {
             final JavadocParser javadocParser = createJavadocParser(javadocComment, errorListener);
+            javadocParser.getInterpreter().setPredictionMode(PredictionMode.SLL);
```
romani, line 116: Please move it inside create..... method.

Let try to update pitest to avoid call interpreter 
Like https://github.com/checkstyle/checkstyle/blob/f123b092eccd870afdb867f29b46aafba95c90e2/pom.xml#L3100

We need to find way avoid pitest suppression.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/JavadocDetailNodeParser.java
romani: Last resort is approach like https://github.com/checkstyle/checkstyle/blob/f123b092eccd870afdb867f29b46aafba95c90e2/pom.xml#L4453

mahfouz72: done with avoidCallsTo and xwiki is unrelated

romani (approved):
ok to merge,
thanks a lot

PR #16642, opened 2025-03-23
Issue 14019: Kill Mutation For ScopeUtil

Abdelrhmansersawy:
Issue #14019 

Diff Regression config: https://gist.githubusercontent.com/Abdelrhmansersawy/a260374764de47cf113461589afb63fb/raw/d85d2b2a5fff191ac8a6a4eda54f28e108aa5888/config.xml

### DIFF
--- a/config/pitest-suppressions/pitest-utils-suppressions.xml
+++ b/config/pitest-suppressions/pitest-utils-suppressions.xml
@@ -1,11 +1,3 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <mutation unstable="false">
-    <sourceFile>ScopeUtil.java</sourceFile>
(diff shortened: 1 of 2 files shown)


Abdelrhmansersawy:
Github, generate report

Abdelrhmansersawy:
Github, generate report

Abdelrhmansersawy:
@romani 
Could you check the diff regression? 
There are no differences in the affected checks, so I believe it's ready to merge.

romani (approved):
Thanks a lot

Rohanraj123:
I was looking for some tests if i could. But i think mutation makes loop not stop even when found correct tokentype. And returnValue would return true  because while traversing the tree towards upward, it would found the correct token type once, but will stop at root node may be class_def.

That makes this scenario , hard to test with avaialble checks. Removing the condition is good fix i believe.

PR #16646, opened 2025-03-23
Issue #16641: FileContents.getJavadocBefore should skip block comment…

### LINKED ISSUE #16641: `FileContents.getJavadocBefore()`: Comments should not be skipped if it is not alone in line
https://github.com/checkstyle/checkstyle/pull/16546#issuecomment-2746033500

Detected at the report: 
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/f4d37cb_2025053704/reports/diff/index.html

unfortunately, In my previous [PR](https://github.com/checkstyle/checkstyle/pull/16586) the config in the PR description was missing some important checks, so the report generated shows no difference but this was misleading.

After I checked a report in another PR   (that was behind my commit on master). I saw the real diff due to my changes and analyzed the report all was good except for this [case](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/f4d37cb_2025053704/reports/diff/pmd/index.html#A1) and the like

**Example:**
```
D:\CS\test
cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="JavadocVariable"/>
    </module>
</module>

D:\CS\test
cat src/Test.java
class Test {
    /**
     *  Javadoc
     */
    /* package */ int x;

    int y;
}
D:\CS\test
java  -jar checkstyle-10.22.0-SNAPSHOT-all.jar -c config.xml src/Test.java
Starting audit...
Audit done.
```
---

The issue because we skipped the line `/* package */ int x;` because it has a block comment. so `int y` becomes associated with the Javadoc above `int x`.

The fix should be simple. A simple condition to see if the block comment is in a single line and not alone in the line 
then we should not skip it.

mahfouz72:
closes #16641 

Diff Regression config: https://raw.githubusercontent.com/mahfouz72/test-configs/refs/heads/javadoc-checks/JavadocChecks/config.xml

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java
@@ -240,8 +240,14 @@ public final class FileContents implements CommentListener {
             .flatMap(List::stream)
             .filter(comment -> !javadocComments.containsValue(comment))
             .anyMatch(comment -> {
-                return lineNo >= comment.getStartLineNo()
-                                        && lineNo <= comment.getEndLineNo();
+                final boolean lineInSideBlockComment = lineNo >= comment.getStartLineNo()
+                                                    && lineNo <= comment.getEndLineNo();
+                boolean lineHasOnlyBlockComment = true;
+                if (comment.getStartLineNo() == comment.getEndLineNo()) {
+                    final String line = line(comment.getStartLineNo() - 1).trim();
+                    lineHasOnlyBlockComment = line.startsWith("/*") && line.endsWith("*/");
+                }
+                return lineInSideBlockComment && lineHasOnlyBlockComment;
             });
     }
 
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/missingjavadoctype/InputMissingJavadocTypeAboveComments.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/missingjavadoctype/InputMissingJavadocTypeAboveComments.java
@@ -11,12 +11,18 @@ tokens = (default)INTERFACE_DEF, CLASS_DEF, ENUM_DEF, ANNOTATION_DEF, RECORD_DEF
 package com.puppycrawl.tools.checkstyle.checks.javadoc.missingjavadoctype;
 
 public class InputMissingJavadocTypeAboveComments { // violation, 'Missing a Javadoc comment.'
-}
 
-/**
- *
- */
-/* */
-class InputMissingJavadocTypeAboveComments2 {
-}
+    /**
+     *
+     */
+    /* */
+    public class InputMissingJavadocTypeAboveComments2 {
+    }
 
+    /**
+     *
+     */
+    public class Myclass { } /* My class */
+
+    public class Myclass2 { } // violation, 'Missing a Javadoc comment.'
+}
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocStyleCheckTest.java
@@ -827,6 +827,7 @@ public class JavadocStyleCheckTest
     public void testJavadocStyleAboveComments() throws Exception {
         final String[] expected = {
             "13: " + getCheckMessage(MSG_NO_PERIOD),
+            "20: " + getCheckMessage(MSG_NO_PERIOD),
         };
 
         verifyWithInlineConfigParser(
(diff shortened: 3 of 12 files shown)


mahfouz72:
Github, generate report

mahfouz72:

mahfouz72:
Report is good.

nrmancuso (approved):
Very good!

src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java
```diff
@@ -240,8 +240,14 @@ private boolean lineInsideBlockComment(int lineNo) {
...
             .filter(comment -> !javadocComments.containsValue(comment))
             .anyMatch(comment -> {
-                return lineNo >= comment.getStartLineNo()
-                                        && lineNo <= comment.getEndLineNo();
+                final boolean lineInSideBlockComment = lineNo >= comment.getStartLineNo()
+                                                    && lineNo <= comment.getEndLineNo();
+                boolean lineHasOnlyBlockComment = true;
+                if (comment.getStartLineNo() == comment.getEndLineNo()) {
+                    final String line = line(comment.getStartLineNo() - 1).trim();
+                    lineHasOnlyBlockComment = line.startsWith("/*") && line.endsWith("*/");
```
rnveach, line 248: I have many questions and I my thoughts may be over the place.

I am not sure if I don't understand this or if this is now violating the method name. We are just checking if a line number we provide is inside/attached to a block comment right? Why does it matter if that block comment starts/ends the line or not? There is a block comment on the line and we are checking that line.

> The issue because we skipped the line /* package */ int x; because it has a block comment. so int y becomes associated with the Javadoc above int x.

This seems unrelated to the general fix you provided. Why did we scan past `int x` when looking for Javadoc comments?
If the code was:
````
    /**
     *  Javadoc
     */
    int x;

    int y;
````
Then we should always say `y` has no javadoc, as there is no javadoc comment between it and `int x`.

Then finally, why does block comments (non-javadocs) have any role in a javadoc check? The javadoc checks should basically be ignoring them always, regardless if they are a single line or their placement.

Comment (rnveach) [reply src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java:248]: After looking at the regression, is this all related to AbstractCheck Javadoc checks and not AbstractJavadocCheck s?

Comment (rnveach) [reply src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java:248]: We resolved this in chat. `FileContents#getJavadocBefore` is basically bad code for not taking the AST or Javadoc identification into account properly. We came up with 1 issue case `/* comment */ // another comment ` but we will ignore it for now for reasons mentioned in the previous sentence.

File: src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/missingjavadoctype/InputMissingJavadocTypeAboveComments.java [hunk 2]
```diff
@@ -1,6 +1,6 @@
 /*
 MissingJavadocType
-scope = (default)public
+scope = private
```
rnveach, line 3: I don't think we are expecting the input to change like this.

If this is because we don't want to make these classes public, can we surround them with a private class and make them nested public?

pom.xml
```diff
@@ -4786,6 +4786,7 @@
                 <param>com.puppycrawl.tools.checkstyle.checks.javadoc.SummaryJavadocCheckTest</param>
                 <param>org.checkstyle.suppressionxpathfilter.XpathRegressionJavadocMethodTest</param>
                 <param>com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocMethodCheckTest</param>
+                <param>com.puppycrawl.tools.checkstyle.checks.javadoc.MissingJavadocTypeCheckTest</param>
```
rnveach, line 4789: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/4fda117e342d409e0e18dade6f45fc129c2b3326_2025234213/reports/diff/checkstyle/index.html#A4
Technically regression found a case on `JavadocMethodCheck` and I don't see an update for that test.
Can we add a case for it and then remove this addition of `MissingJavadocTypeCheckTest`?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/missingjavadoctype/InputMissingJavadocTypeAboveComments.java
mahfouz72: done

pom.xml

romani (approved):
thanks a lot

PR #16657, opened 2025-03-25
Issue #14019: Kill mutation for getOutputStreamOptions(options.outputPath))

Z4ID-Reaper:
issue #14019


Z4ID-Reaper:
@romani i might need you to validate/review my assumptions here to ensure i am on the right track.

**mutator in question**: NonVoidMethodCallMutator 

this was the PIT supression:
Screenshot 2025-03-25 193219

this is the actual code line  in the main.java file:
Screenshot 2025-03-25 191855

this is the method's implementation:
Screenshot 2025-03-25 191931

### according to my assumptions
 here i do not see the situation where this method returns a `null ` (default) value , making the suppression unnecessary testing that case.

from here what should i chose ?

- i can create a test case to check for the null value in the test's files
- i can create a detailed explanation of why the suppression is unkillable 

or is there something i am missing here?

romani:
Pitest change code to `istener = new XpathFileGeneratorAuditListener(getOutputStream(options.outputPath), null);`

Please write test that failing if such hardcoded value is in code.

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/XpathFileGeneratorAuditListenerTest.java
```diff
@@ -269,6 +269,19 @@ public void testNoCloseStream() {
...
+    @Test
+    public void testNoCloseStreamTwo() {
+        final XpathFileGeneratorAuditListener listener =
+                new XpathFileGeneratorAuditListener(outStream, null);
+        listener.finishLocalSetup();
+        listener.auditStarted(null);
+        listener.auditFinished(null);
+
+        assertWithMessage("Invalid close count")
+            .that(outStream.getCloseCount())
```
romani, line 281: If stream is not closed, try tif stream is not closed, try to write to stream some content , and validate presence of it

romani (changes requested):
Items

romani:
We need different fix

Z4ID-Reaper:
it's on my mind , just need some time.

PR #16659, opened 2025-03-25
Pull #16659: Use SLL prediction mode in JavaParser to speed up parsing

mahfouz72:
> This is the fastest prediction mode, and provides correct results for many grammars

```Java
public enum PredictionMode {
	/**
	 * The SLL(*) prediction mode. This prediction mode ignores the current
	 * parser context when making predictions. This is the fastest prediction
	 * mode, and provides correct results for many grammars. This prediction
	 * mode is more powerful than the prediction mode provided by ANTLR 3, but
	 * may result in syntax errors for grammar and input combinations which are
	 * not SLL.
	 *
	 * <p>
	 * When using this prediction mode, the parser will either return a correct
	 * parse tree (i.e. the same parse tree that would be returned with the
	 * {@link #LL} prediction mode), or it will report a syntax error. If a
	 * syntax error is encountered when using the {@link #SLL} prediction mode,
	 * it may be due to either an actual syntax error in the input or indicate
	 * that the particular combination of grammar and input requires the more
	 * powerful {@link #LL} prediction abilities to complete successfully.</p>
	 *
	 * <p>
	 * This prediction mode does not provide any guarantees for prediction
	 * behavior for syntactically-incorrect inputs.</p>
	 */
	SLL,
        .....
```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java
@@ -43,6 +43,7 @@ import com.puppycrawl.tools.checkstyle.api.TokenTypes;
 import com.puppycrawl.tools.checkstyle.grammar.java.JavaLanguageLexer;
 import com.puppycrawl.tools.checkstyle.grammar.java.JavaLanguageParser;
 import com.puppycrawl.tools.checkstyle.utils.ParserUtil;
+import org.antlr.v4.runtime.atn.PredictionMode;
 
 /**
  * Helper methods to parse java source files.
@@ -89,6 +90,7 @@ public final class JavaParser {
         final CommonTokenStream tokenStream = new CommonTokenStream(lexer);
         final JavaLanguageParser parser =
                 new JavaLanguageParser(tokenStream, JavaLanguageParser.CLEAR_DFA_LIMIT);
+        parser.getInterpreter().setPredictionMode(PredictionMode.SLL);
         parser.setErrorHandler(new CheckstyleParserErrorStrategy());
         parser.removeErrorListeners();
         parser.addErrorListener(new CheckstyleErrorListener());


nrmancuso:
> Some files were successfully parsed after converting to SLL parsing, while most others were not. I will try to use SLL parsing first and fall back to the more powerful LL parsing if SLL fails to parse the file. but I am not sure if this approach will improve performance or introduce overhead that will affect performance badly because, for most of the files, we will try to parse twice first with SLL and then LL

I found the same, and trying to fall back made performance much worse at the time.

mahfouz72:
```
Execution Time Baseline: 445.76 s
Average Execution Time: 436.70 s
Execution Time Difference: 2.0300%
Difference is within the maximum allowed difference (2.0300%      <= 10%).
```

Not a huge reduction.

The first run resulted in a 1.5% reduction, and the current run resulted in a 2% reduction
At this point, I am not sure if we should do this two-stage parsing or not

@nrmancuso, Please confirm if we can continue with this PR or close it 

I have noticed that many sources online recommend using two-stage parsing.

Examples:

https://tomassetti.me/improving-the-performance-of-an-antlr-parser/#:~:text=to%20improve%20performance.-,Changing%20the%20Prediction%20Mode,-A%20simple%20improvement

https://github.com/antlr/antlr4/issues/192#issuecomment-15238595

And ANTLR v4 book.

src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java
```diff
@@ -89,22 +92,64 @@ public static DetailAST parse(FileContents contents)
...
         }
         catch (IllegalStateException ex) {
             final String exceptionMsg = String.format(Locale.ROOT,
-                "%s occurred while parsing file %s.",
-                ex.getClass().getSimpleName(), contents.getFileName());
+                    "%s occurred while parsing file %s.",
+                    ex.getClass().getSimpleName(), fileName);
             throw new CheckstyleException(exceptionMsg, ex);
-        }
+        };
```
pankratz76, line 150:
```suggestion
        }
```
why this semicolon? above its without

src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java
```diff
@@ -89,22 +92,64 @@ public static DetailAST parse(FileContents contents)
         final CommonTokenStream tokenStream = new CommonTokenStream(lexer);
         final JavaLanguageParser parser =
                 new JavaLanguageParser(tokenStream, JavaLanguageParser.CLEAR_DFA_LIMIT);
-        parser.setErrorHandler(new CheckstyleParserErrorStrategy());
+        JavaLanguageParser.CompilationUnitContext compilationUnit;
+
+        try {
+            compilationUnit = parseWithSLLPredictionMode(parser);
+        }
+        catch (ParseCancellationException e) {
```
pankratz76, line 100:
```suggestion
        catch (ParseCancellationException ignored) {
```
`ex` or more precise `ignored` is the pattern imho. I expect the build to fail as imho we have a pattern avoiding one character names like `e`.

pankratz76 (approved):
nicely done

nrmancuso:
> @nrmancuso, Please confirm if we can continue with this PR or close it

I personally don’t think that the added complexity is worth this small gain, @rnveach what do you think?

rnveach:
> Some files were successfully parsed after converting to SLL parsing

Are you saying files failed to parse with LL, and then parsed with SLL?

> The first run resulted in a 1.5% reduction, and the current run resulted in a 2% reduction
> I personally don’t think that the added complexity is worth this small gain

I agree. Lets drop this PR.

mahfouz72:
> > Some files were successfully parsed after converting to SLL parsing
> 
> Are you saying files failed to parse with LL, and then parsed with SLL?

This comment was made before trying the two-stage parsing. I wanted to say that if we rely only on SLL, a few files will be parsed successfully, but most will not

No the PR has two stage parsing (SLL then fallback to LL if SLL failed)

src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java
```diff
@@ -137,7 +137,7 @@ public static DetailAST parse(FileContents contents)
         parser.setErrorHandler(new CheckstyleParserErrorStrategy());
         parser.getInterpreter().setPredictionMode(PredictionMode.LL);
 
-        JavaLanguageParser.CompilationUnitContext compilationUnit;
+        final JavaLanguageParser.CompilationUnitContext compilationUnit;
 
         try {
             compilationUnit = parser.compilationUnit();
```
pankratz76, line 143:
```suggestion
            return parser.compilationUnit();
```

src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java
```diff
@@ -137,7 +137,7 @@ public static DetailAST parse(FileContents contents)
         parser.setErrorHandler(new CheckstyleParserErrorStrategy());
         parser.getInterpreter().setPredictionMode(PredictionMode.LL);
 
-        JavaLanguageParser.CompilationUnitContext compilationUnit;
+        final JavaLanguageParser.CompilationUnitContext compilationUnit;
```
pankratz76, line 140: no need to declare a field for only one time usage then.
```suggestion
```

src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java
```diff
@@ -147,7 +147,7 @@ public static DetailAST parse(FileContents contents)
                     "%s occurred while parsing file %s.",
                     ex.getClass().getSimpleName(), fileName);
             throw new CheckstyleException(exceptionMsg, ex);
-        };
+        }
 
         return compilationUnit;
```
pankratz76, line 152:
```suggestion
        throw new IllegalStateException("should not occure")
```
As we could make the return immediately now, we focus on where the action happens and avoid unnecessarily disrupting the code flow. Since we only have one valid case now, we should highlight this clearly.

pankratz76 (changes requested):
remove field and use direct return

mahfouz72:
Github, generate report for SingleLineJavadoc/Example1

mahfouz72:
Github, generate report for SingleLineJavadoc/Example1

romani:
Github, generate report for SingleLineJavadoc/Example1

romani:
Above comment is with space newline at the end

mahfouz72:

romani:
@mahfouz72 , looks like this PR should be closed

romani:
Got confirmation in chat to close this PR

PR #16663, opened 2025-03-26
Issue #16653: dependency: bump tech.picnic.error-prone-support:error-prone-contrib …

Mahmoud-Ayman-Saleh:
Issue #16653


romani:
jdk17 update is required before updating this dependency.

Lets wait for our official  upgrade to jdk17

PR #16665, opened 2025-03-26
infra: Enabling Spot-Bug in CircleCI

AmitKumarDeoghoria:
Earlier Disabled : https://github.com/checkstyle/checkstyle/pull/14863
Due to SpotBug Issue Mentioned in :https://github.com/spotbugs/spotbugs-maven-plugin/issues/806

Local Results on : mvn clean test-compile spotbugs
Screenshot 2025-03-27 at 1 16 18 AM

### Most of the Bugs Flagged by SpotBugs are False Positives 
**For Example Lets Take This High-Confidence Error Flagges by SpotBug :**
Screenshot 2025-04-01 at 11 29 13 PM
**This is the part of the code where Error is Flagged :**  
Screenshot 2025-04-01 at 11 30 30 PM
## What is the Error?
Ans:The error RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE occurs because Spotbugs detects a redundant null check on a variable (stream in your code) that is guaranteed to be non-null due to prior dereferencing. In this case, this is likely a false positive caused by Spotbugs analyzing the bytecode generated for the try-with-resources block.
## What’s happening:

-Files.find(...) returns a Stream<Path>, which will never be null (it throws IOException on failure).
-The try-with-resources block auto-closes the stream, and the generated bytecode includes an implicit if (stream != null) check to avoid closing a null resource. Spotbugs flags this as redundant because stream cannot logically be null here.

## Why it’s a false positive:
The redundant check is in the bytecode (generated by try-with-resources), not our source code. Spotbugs is technically correct but overly strict here.

### Solution
- We have added all the false positives in the SpotBug-Exclusion List for now
- We can hope that with newer version of spotbug and sb-contrib have less false positives.
- we can Increase the threshold in config for spotbug from Low to Medium or High , In order to have less stricter checks.
`<threshold>Medium</threshold>`  [ But in this case we might miss out on some small actual bugs but the flase positives will be drastically reduced]

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -197,11 +197,12 @@ workflows:
           name: "checkstyle-and-sevntu"
           image-name: *cs_img
           command: "./.ci/validation.sh checkstyle-and-sevntu"
-      # until https://github.com/spotbugs/spotbugs-maven-plugin/issues/806
-      # - validate-with-maven-script:
-      #    name: "spotbugs-and-pmd"
-      #    image-name: *cs_img
-      #    command: "./.ci/validation.sh spotbugs-and-pmd"
+      # https://github.com/spotbugs/spotbugs-maven-plugin/issues/806 explains
+      # why we need execution of spotbugs without any other maven plugins which can change binaries
+      - validate-with-maven-script:
+          name: "spotbugs-and-pmd"
+          image-name: *cs_img
+          command: "./.ci/validation.sh spotbugs-and-pmd"
       - validate-with-maven-script:
           name: "site"
           image-name: *cs_img
--- a/config/spotbugs-exclude.xml
+++ b/config/spotbugs-exclude.xml
@@ -325,4 +325,83 @@
     <Method name="matchesAtFrontNoRegex"/>
     <Bug pattern="BAS_BLOATED_ASSIGNMENT_SCOPE"/>
   </Match>
+
+  <!-- Suppress PRMC_POSSIBLY_REDUNDANT_METHOD_CALLS -->
+  <!-- until https://github.com/checkstyle/checkstyle/issues/16979 -->
+  <Match>
+    <Class name="com.puppycrawl.tools.checkstyle.utils.JavadocUtil" />
+    <Bug pattern="PRMC_POSSIBLY_REDUNDANT_METHOD_CALLS" />
+  </Match>
+  <Match>
+    <Class name="com.puppycrawl.tools.checkstyle.DetailNodeTreeStringPrinter" />
+    <Bug pattern="PRMC_POSSIBLY_REDUNDANT_METHOD_CALLS" />
+  </Match>
+  <Match>
+    <Class name="com.puppycrawl.tools.checkstyle.JavaAstVisitor" />
+    <Bug pattern="PRMC_POSSIBLY_REDUNDANT_METHOD_CALLS" />
+  </Match>
+  <Match>
+    <Class name="com.puppycrawl.tools.checkstyle.JavaAstVisitor$DetailAstPair" />
+    <Bug pattern="PRMC_POSSIBLY_REDUNDANT_METHOD_CALLS" />
+  </Match>
+  <Match>
+    <Class name="com.puppycrawl.tools.checkstyle.checks.indentation.BlockParentHandler" />
+    <Bug pattern="PRMC_POSSIBLY_REDUNDANT_METHOD_CALLS" />
+  </Match>
+  <Match>
+    <Class name="com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheck" />
+    <Bug pattern="PRMC_POSSIBLY_REDUNDANT_METHOD_CALLS" />
+  </Match>
+  <Match>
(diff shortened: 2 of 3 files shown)


AmitKumarDeoghoria:
@romani  can you review please. CI error is unrelated to the changes.

AmitKumarDeoghoria:
@romani  Ping.

config/spotbugs-exclude.xml
```diff
@@ -335,4 +335,78 @@
     <Class name="com.puppycrawl.tools.checkstyle.checks.sizes.RecordComponentNumberCheck"/>
     <Bug pattern="AT_STALE_THREAD_WRITE_OF_PRIMITIVE"/>
   </Match>
+    <!-- New Suppressions -->
```
romani, line 338: I am confused , why this violations appears?
It should be same as execution of `mvn clean verify `

romani (approved):
Issue is closed, and about code covera

Items

config/spotbugs-exclude.xml
AmitKumarDeoghoria: Hey @romani I suspect the difference arises because mvn verify runs later build phases that optimize/modify bytecode , while mvn test-compile spotbugs:check analyzes raw, unprocessed bytecode, exposing  violations(e.g., redundant null checks) may be due to incomplete plugin/annotation processing.

AmitKumarDeoghoria: Also I have referenced the Issue correctly now.

romani: ok, please analyze if this suppressins should stay forever, or we need to fix them in followup PRs.

AmitKumarDeoghoria: No need of fixing. I analyzed some of the High Confidence Checks done by SpotBug, Most of them are false positives. I will give a example in Description  to clarify why this is happening.

romani: please analyze all of them, if there is already defect, find link to it and place comment `<!-- until .... -->`
if issue not existing, please create and place same "until" comment.

please create issue on spotbugs plugin to get to know why validation by `mvn clean verify` is different from `mvn spotbugs:check`

romani:
please change commit message to have prefix "infra:" , we reference issues in prefix only to our issue, not others, you can put link to spotbugs  in commit message after prefix.
please do, to fix CI

romani (changes requested):
see above.

config/spotbugs-exclude.xml
romani: Ok, we need to slice and dice current suppressions by :
- Suppression until spotbug issue is fixed 
- point to improve in checkstyle codebase, separate issue.

We need enablement to be merged quicker to prevent new leaks of violations.

.circleci/config.yml
```diff
@@ -197,11 +197,10 @@ workflows:
           name: "checkstyle-and-sevntu"
           image-name: *cs_img
           command: "./.ci/validation.sh checkstyle-and-sevntu"
-      # until https://github.com/spotbugs/spotbugs-maven-plugin/issues/806
-      # - validate-with-maven-script:
-      #    name: "spotbugs-and-pmd"
-      #    image-name: *cs_img
-      #    command: "./.ci/validation.sh spotbugs-and-pmd"
+      - validate-with-maven-script:
```
romani, line 200: Please add comment above this job `https://github.com/spotbugs/spotbugs-maven-plugin/issues/806 explains why we need execution of spotbug without any other maven plugins which can change binaries `

romani (changes requested):
Items

config/spotbugs-exclude.xml
```diff
@@ -325,4 +325,92 @@
...
+    <Class name="com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheck" />
+    <Bug pattern="PRMC_POSSIBLY_REDUNDANT_METHOD_CALLS" />
+  </Match>
+  <Match>
+    <Class name="com.puppycrawl.tools.checkstyle.meta.XmlMetaWriter" />
+    <Bug pattern="PRMC_POSSIBLY_REDUNDANT_METHOD_CALLS" />
+  </Match>
+
+  <!-- Suppress AFBR_ABNORMAL_FINALLY_BLOCK_RETURN -->
+  <!-- until https://github.com/spotbugs/spotbugs-maven-plugin/issues/806 -->
```
romani, line 371: Please create new issue on checkstyle to fix all such violations.
Suppression file should have links to checkstyle issue until which suppression stays.
It will be separate activities to resolve them.

config/spotbugs-exclude.xml
romani: Please remove `<!-- New Suppressions -->` line, it has no value

romani (changes requested):
Last two items

romani:
please recheck suppressions, 
https://github.com/checkstyle/checkstyle/issues/16882

some already fixed.

config/spotbugs-exclude.xml
AmitKumarDeoghoria: Done

romani (approved):
Ok to merge.
Thanks a lot

PR #16668, opened 2025-03-27
Issue #14019: Covered pitest survival in DetailNodeTreeStringPrinter

Atharv3221:
Issue: #14019

Solved Pitest Survivals:

```
 <mutation unstable="false">
    <sourceFile>DetailNodeTreeStringPrinter.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.DetailNodeTreeStringPrinter</mutatedClass>
    <mutatedMethod>parseFile</mutatedMethod>
    <description>removed call to java/nio/charset/Charset::name</description>
    <lineContent>System.getProperty(&quot;file.encoding&quot;, StandardCharsets.UTF_8.name()));</lineContent>
  </mutation>
```

```
**Explanation:**

final FileText text = new FileText(file.getAbsoluteFile(),
            System.getProperty("file.encoding", StandardCharsets.UTF_8.name()));
Here,   StandardCharsets.UTF_8.name() always returns "UTF-8" and that doesn't depend on anything,
   
```

Diff Regression config: https://github.com/Atharv3221/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/DetailNodeTreeStringPrinter.java
Diff Regression patch config: https://github.com/Atharv3221/checkstyle/blob/pitestSolving/src/main/java/com/puppycrawl/tools/checkstyle/DetailNodeTreeStringPrinter.java
New module config: https://github.com/Atharv3221/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/DetailNodeTreeStringPrinter.java
Diff Regression Projects:
https://raw.githubusercontent.com/checkstyle/contribution/master/checkstyle-tester/projects-to-test-on.properties
GitHub, generate report

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/DetailNodeTreeStringPrinter.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/DetailNodeTreeStringPrinter.java
@@ -162,6 +162,21 @@ public final class DetailNodeTreeStringPrinter {
     }
 
     /**
+     * Retrieves the file encoding from system properties.
+     * If the provided encoding is blank, an exception is thrown.
+     *
+     * @param encode the encoding to use as a fallback if not found in system properties.
+     * @return the file encoding from system properties, or the provided fallback encoding.
+     * @throws IllegalStateException if the provided encoding is blank.
+     */
+    private static String getFileEncoding(String encode) {
+        if (encode.isBlank()) {
+            throw new IllegalStateException("Encode should not be blank");
+        }
+        return System.getProperty("file.encoding", encode);
+    }
+
+    /**
      * Parse a file and return the parse tree.
      *
      * @param file the file to parse.
@@ -170,7 +185,7 @@ public final class DetailNodeTreeStringPrinter {
      */
     private static DetailNode parseFile(File file) throws IOException {
         final FileText text = new FileText(file.getAbsoluteFile(),
-            System.getProperty("file.encoding", StandardCharsets.UTF_8.name()));
+            getFileEncoding(StandardCharsets.UTF_8.name()));
         return parseJavadocAsDetailNode(text.getFullText().toString());
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/DetailNodeTreeStringPrinter.java
```diff
@@ -170,7 +169,7 @@ private static String getIndentation(DetailNode node) {
      */
     private static DetailNode parseFile(File file) throws IOException {
         final FileText text = new FileText(file.getAbsoluteFile(),
-            System.getProperty("file.encoding", StandardCharsets.UTF_8.name()));
+                 System.getProperty("file.encoding", "UTF-8"));
```
romani, line 172: Can we just remove `.name()`

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/DetailNodeTreeStringPrinter.java
Atharv3221: Only removing .name(), doesn't work, as StandardCharsets.UTF_8 will return a value of type Charset but we need String, which is done by .name() even if we used String.valueOf(), it will create another surviving mutation. We will have same number of surviving mutations as before.

I converted StandardCharsets.UTF_8.name() to "UTF-8" because StandardCharsets.UTF_8 will always return same thing and .name() will always return "UTF-8"

Atharv3221: Is this approach worng should i try something else ??

romani: We should find a way to use java constant, not a string.

Atharv3221: In FileText( , ) 2nd argument is taken as a String but inside the FileText class its value is again convert to Charset, so is it necessary to pass Charset by converting it to a string and then again converting it to Charset from string ? Can't FileText( , ) simply take Charset ?

Atharv3221:
final FileText text = new FileText(file.getAbsoluteFile(), 
              StandardCharsets.UTF_8.name());

This works but we need to use System variable too right ??

romani:
Yes, we need, system variable usage

PR #16689, opened 2025-03-30
Issue #14084: Remove warning for nullness-optinal-interning profile: …

Rohanraj123:
part of #14084

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java
@@ -56,6 +56,7 @@ import com.puppycrawl.tools.checkstyle.api.Scope;
 import com.puppycrawl.tools.checkstyle.api.SeverityLevel;
 import com.puppycrawl.tools.checkstyle.checks.naming.AccessModifierOption;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
 
 /**
  * A Java Bean that implements the component lifecycle interfaces by
@@ -85,7 +86,7 @@ public abstract class AbstractAutomaticBean
     private static final String COMMA_SEPARATOR = ",";
 
     /** The configuration of this bean. */
-    private Configuration configuration;
+    @MonotonicNonNull private Configuration configuration;
 
     /**
      * Provides a hook to finish the part of this component's setup that
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -2,13 +2,6 @@
 <suppressedErrors>
   <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java</fileName>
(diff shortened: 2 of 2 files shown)


Rohanraj123:
@romani Checker warning resolved with this change.

romani:
CI is very red, please address all CI  errors.
CI should be fully green.

Rohanraj123:
ping @romani

src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java
```diff
@@ -296,6 +304,7 @@ protected final Configuration getConfiguration() {
      * @throws CheckstyleException if there is a configuration error.
      * @see Configuration#getChildren
      */
+    @RequiresNonNull("configuration")
```
romani, line 307:
```
    @RequiresNonNull("configuration")
    protected void setupChild(Configuration childConf)
            throws CheckstyleException {
        if (childConf != null) {
            throw new CheckstyleException(childConf.getName() + " is not allowed as a child in "
                    + configuration.getName() + ". Please review 'Parent Module' section "
                    + "for this Check in web documentation if Check is standard.");
```

usage of `configuration`  here is minimal, we can wrap it optional or extra IF.
but it is not required to be not null, and if it is null, sequence of calls on user side is wrong.
setupChild is only for Treewalker and Checker, all other modules(Checks) use this implementation to throw excpeption if somebody try to put Check in Check.

So usage of this annotation is good, no action is required.
if it is null, it will throw NPE, user notice it in testing.
If it is null, user do no testing, it will work.  If user use checkerframework and run it , it might get violation to ensure order of methods. If checkerframework is not used, nothing will help user to catch this defect before going live.
Some benefit is for engineers that read sources of libraries, is that they can spot declaration of dependecies of calls/initialization and avoid problems.

romani: for us usage of this rule in checkerframework is that it demands demand design to be less relying of calls of methods in right order, conceputally leading to put all in c-tor.

src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java
```diff
@@ -278,8 +281,13 @@ public final void contextualize(Context context)
...
      *
      * @return the configuration that was used to configure this component.
+     * @throws IllegalArgumentException throw exception when configuration is null.
      */
+    @EnsuresNonNull("configuration")
     protected final Configuration getConfiguration() {
+        if (configuration == null) {
+            throw new IllegalArgumentException("configuration should not be null");
+        }
         return configuration;
```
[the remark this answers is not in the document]
Rohanraj123: > please share links to docs to prove that exception is fix they require.

Regarding the use of `@EnsuresNonNull,` according to the [documentation](https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/EnsuresNonNull.html), the annotation states that “on successful termination of the method, the given expressions are non-null.”

In getConfiguration(), we explicitly check for null and throw an exception if it is null. Since the method cannot return when configuration is null, it satisfies the condition required by `@EnsuresNonNull.`

However, I understand your point that, the method isn’t actually initializing or ensuring non-null on its own — it's just asserting it. Working on it ! If I could figure out something more.

romani:
we lost energy here, lets close this and may be someone from web find time to continue

PR #16690, opened 2025-03-30
Issue #14084: Remove warning from checker-formatter

Rohanraj123:
part of #14084

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java
@@ -37,6 +37,9 @@ import java.util.stream.Collectors;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.transform.TransformerException;
 
+import org.checkerframework.checker.formatter.qual.ConversionCategory;
+import org.checkerframework.checker.formatter.qual.Format;
+
 import com.puppycrawl.tools.checkstyle.FileStatefulCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.DetailNode;
@@ -117,7 +120,7 @@ public class JavadocMetadataScraper extends AbstractJavadocCheck {
     /**
      * Format for exception message for missing default value for check property.
      */
-    private static final String PROP_DEFAULT_VALUE_MISSING =
+    private static final @Format(ConversionCategory.GENERAL) String PROP_DEFAULT_VALUE_MISSING =
         "Default value for property '%s' is missing";
 
     /** ModuleDetails instance for each module AST traversal. */
--- a/config/checker-framework-suppressions/checker-formatter-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-formatter-suppressions.xml
@@ -11,13 +11,6 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java</fileName>
     <specifier>format.string</specifier>
     <message>invalid format string (is a @Format annotation missing?)</message>
-    <lineContent>Locale.ROOT, PROP_DEFAULT_VALUE_MISSING, propertyName)</lineContent>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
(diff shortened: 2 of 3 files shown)


Rohanraj123:
@romani ping

src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java
```diff
@@ -117,7 +120,7 @@ public class JavadocMetadataScraper extends AbstractJavadocCheck {
     /**
      * Format for exception message for missing default value for check property.
      */
-    private static final String PROP_DEFAULT_VALUE_MISSING =
+    private static final @Format(ConversionCategory.GENERAL) String PROP_DEFAULT_VALUE_MISSING =
         "Default value for property '%s' is missing";
```
romani, line 124: https://checkerframework.org/releases/3.12.0/manual/#formatter-checker

I need to think on this for a bit. 
Not sure if this annotation helps.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/meta/JavadocMetadataScraper.java
Rohanraj123: https://checkerframework.org/releases/3.12.0/manual/#formatter-checker:~:text=GENERAL%20imposes%20no%20restrictions%20on%20a%20format%20argument%E2%80%99s%20type.%20Applicable%20for%20conversions%20b%2C%20B%2C%20h%2C%20H%2C%20s%2C%20S.

Yeah please take a look at this line!!

If we completely want to integrate checker into our codebase then, we can use these annotations.. Please let me know!

romani: >[GENERAL](https://checkerframework.org/releases/3.12.0/api/org/checkerframework/checker/formatter/qual/ConversionCategory.html#GENERAL) imposes no restrictions on a format argument’s type. Applicable for conversions b, B, h, H, s, S.

no point to have checker is we put annotation that suppress it.

https://checkerframework.org/releases/3.12.0/manual/#formatter-format-subtyping
try:
```
The following example shows the subtyping rules in action:

    @Format({FLOAT, INT}) String f;

    f = "%f %d";       // OK
    f = "%s %d";       // OK, %s is weaker than %f
    f = "%f";          // warning: last argument is ignored
```

romani: @nrmancuso , @rdiachenko , please share your thoughts on how useful this Checker validation for us, while we have 100% coverage and 100% pitest coverage.

romani: @Rohanraj123 , can you check if tests are going to catch fact that wrong type marker is used in string?
`"Default value for property '%d' is missing";`

Rohanraj123: @romani Yes test fails when I change `%s -> %d` in variable. Which means it catches the wrong type marker used in string.

Rohanraj123: If test covers these changes, then CF is not needed for that purpose, I think.

romani: CF is validation on compile time conceptually quicker response, test is runtime usually longer time to provide feedback, but problem is that CF execution is very long, test passing quicker, so it not clear what we are getting by this CF validation and dealing with extra annotations over code.

I am not sure how this annotation helps to understand what to expect. Nullable annotation helps to avoid reading code of method and kind of be sure on nature of result.

Rohanraj123:
> @Rohanraj123 , please new PR with disablement of such validation.
> 
> This PR will be great example and explanation on why we do disablement of of this validation in CF.

Since the Formatter Checker uses the I18n Checker as its base, and our tests already catch the issues these checkers would flag, removing the Formatter Checker is a reasonable.

romani:
Checking is excluded at https://github.com/checkstyle/checkstyle/pull/16805

Issue #15098, opened 2024-06-22
Indentation of the Block child of switch rule is not validated when no braces


mahfouz72:
detected at #14997 

I have read check documentation: https://checkstyle.org/checks/misc/indentation.html#Indentation
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

---

```
PS D:\CS\test> cat config.xml                                                                                         
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
<module name="Checker">
    <property name="charset" value="UTF-8"/>
    <module name="TreeWalker">
            <module name="Indentation"/>
    </module>
</module>
PS D:\CS\test> cat src/Test.java                                                                                      
public class Test {

    void test(int obj) {
        switch (obj) {
            case 1 ->
System.out.println("1");                      // This is not being validated
            case 2 -> {
System.out.println("2");  // violation        (there are braces)
            }
        }

        switch (obj) {
            case 1 :
System.out.println("1");  // violation
            case 2 : {
System.out.println("2");  // violation
            }
        }

        if (obj == 1)
System.out.println("1");  // violation
        else {
System.out.println("2");  // violation
        }
    }

}
PS D:\CS\test> java  -jar checkstyle-10.17.0-all.jar -c config.xml src/Test.java
Starting audit...
[ERROR] D:\CS\test\src\Test.java:8:1: 'block' child has incorrect indentation level 0, expected level should be 16. [Indentation]
[ERROR] D:\CS\test\src\Test.java:14:1: 'block' child has incorrect indentation level 0, expected level should be 16. [Indentation]
[ERROR] D:\CS\test\src\Test.java:16:1: 'block' child has incorrect indentation level 0, expected level should be 16. [Indentation]
[ERROR] D:\CS\test\src\Test.java:21:1: 'if' child has incorrect indentation level 0, expected level should be 12. [Indentation]
[ERROR] D:\CS\test\src\Test.java:23:1: 'else' child has incorrect indentation level 0, expected level should be 12. [Indentation]
Audit done.
Checkstyle ends with 5 errors.
PS D:\CS\test> 

```

---

Describe what you want in detail.

I expect the block child of the switch rule to be validated even if there are no braces.

[...]

PR #16647, opened 2025-03-24
Issue #15098: Indentation of switch block when no braces is now validated

Aziz-755:
Issue: #15098 

Diff Regression config: https://gist.githubusercontent.com/mohitsatr/937e10572dd412961b61c77577619dff/raw/7a47240f8793e7335adbd81c2806b375f0394815/google_checks_indentation.xml
Diff Regression projects: https://gist.githubusercontent.com/mohitsatr/68b0817ac52ff7773d35585ee12fc1cc/raw/9ba8e4fb324e1daf589874c6992f2aa87b03079a/list-of-projects.properties

PR #16721, opened 2025-04-01
Issue #15098: Indentation of switch block when no braces is now validated

Aziz-755:
fixes: #15098 , #16762
This PR reintroduces the changes from the original PR ([#16647](https://github.com/checkstyle/checkstyle/pull/16647)), which was closed due to a required refork of the repository.

New module config: https://gist.githubusercontent.com/mohitsatr/937e10572dd412961b61c77577619dff/raw/7a47240f8793e7335adbd81c2806b375f0394815/google_checks_indentation.xml
---Diff-- Regression config: https://gist.githubusercontent.com/mohitsatr/937e10572dd412961b61c77577619dff/raw/7a47240f8793e7335adbd81c2806b375f0394815/google_checks_indentation.xml
Diff Regression projects: https://gist.githubusercontent.com/mohitsatr/68b0817ac52ff7773d35585ee12fc1cc/raw/9ba8e4fb324e1daf589874c6992f2aa87b03079a/list-of-projects.properties

### DIFF
--- a/src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting/rule4841indentation/InputSingleSwitchStatementWithoutCurly.java
+++ b/src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting/rule4841indentation/InputSingleSwitchStatementWithoutCurly.java
@@ -1,0 +1,60 @@
+package com.google.checkstyle.test.chapter4formatting.rule42blockindentation;
+
+/**some javadoc.*/
+public class InputSingleSwitchStatementWithoutCurly { 
+  void testCorrectIndentation(int obj) {
+    switch (obj) {
+      case 0
+          ->
+              System.out.println("Test");
+      case 1 ->
+          System.out.println("TEST");
+      case 2 -> {
+        System.out.println("Test");
+      }
+      default -> System.out.println("test");
+    }
+  }
+  
+  void testIncorrectIndentation(int obj) {
+    switch (obj) {
+      case 1
+      ->                            // violation '.* incorrect indentation level 6, expected .* 10.'
+      System.out.println("Test");
+      // violation above '.* incorrect indentation level 6, expected .* 14.'
+    case 2 ->                      // violation '.* incorrect indentation level 4, expected .* 6.'
+System.out.println("Test");        // violation '.* incorrect indentation level 0, expected .* 10.'
+      default -> System.out.println("test");
+    }
+  }
+
+  void testMixedCases(int obj) {
+    switch (obj) {
+      case 1 -> System.out.println("TEST");
+      case 2
+          -> System.out.println("Test");
+      case 3 ->
+                    System.out.println("Test");
+      // violation above '.* incorrect indentation level 20, expected .* 10.'
+      case 4 -> {
+        System.out.println("Test");
+      System.out.println("Another statement");
+      // violation above '.* incorrect indentation level 6, expected .* 8.'
+      }
+      default -> System.out.println("test");
+    }
+  }
+  
+  private boolean testLineWrapping(int x, int y, int z) {
+    return switch (x) {
+      case 1 -> true;
+      case 2 -> x != y
+          && y != 5;
+      case 3 ->
+          y != 4
+            && z != 2
+            && y != z;
+      default -> false;
+    };
+  }
+}
--- a/src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting/rule4841indentation/InputSingleSwitchStatementWithoutCurlyCorrect.java
+++ b/src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting/rule4841indentation/InputSingleSwitchStatementWithoutCurlyCorrect.java
@@ -1,0 +1,57 @@
+package com.google.checkstyle.test.chapter4formatting.rule42blockindentation;
+
+/**some javadoc.*/
+public class InputSingleSwitchStatementWithoutCurlyCorrect {
+  void testCorrectIndentation(int obj) {
+    switch (obj) {
+      case 0
+          ->
+              System.out.println("Test");
+      case 1 ->
+          System.out.println("TEST");
+      case 2 -> {
+        System.out.println("Test");
+      }
+      default -> System.out.println("test");
+    }
+  }
+  
+  void testIncorrectIndentation(int obj) {
+    switch (obj) {
+      case 1
+          ->
+              System.out.println("Test");
+      case 2 ->
+          System.out.println("Test");
+      default -> System.out.println("test");
+    }
+  }
+  
+  void testMixedCases(int obj) {
+    switch (obj) {
+      case 1 -> System.out.println("TEST");
+      case 2
+          -> System.out.println("Test");
+      case 3 ->
+          System.out.println("Test");
+      case 4 -> {
+        System.out.println("Test");
+        System.out.println("Another statement");
+      }
+      default -> System.out.println("test");
+    }
+  }
+  
+  private boolean testLineWrapping(int x, int y, int z) {
+    return switch (x) {
+      case 1 -> true;
+      case 2 -> x != y
+          && y != 5;
+      case 3 ->
+          y != 4
+            && z != 2
+            && y != z;
+      default -> false;
+    };
+  }
+}
--- a/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationCheckSingleSwitchStatementsWithoutCurly.java
+++ b/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationCheckSingleSwitchStatementsWithoutCurly.java
@@ -1,0 +1,64 @@
+//non-compiled with javac: Compilable with Java17                      //indent:0 exp:0
+package com.puppycrawl.tools.checkstyle.checks.indentation.indentation;//indent:0 exp:0
+
+/* Config:                                                             //indent:0 exp:0
+ *                                                                     //indent:1 exp:1
+ * basicOffset = 4                                                     //indent:1 exp:1
+ * braceAdjustment = 0                                                 //indent:1 exp:1
+ * caseIndent = 4                                                      //indent:1 exp:1
+ * throwsIndent = 4                                                    //indent:1 exp:1
+ * arrayInitIndent = 4                                                 //indent:1 exp:1
+ * lineWrappingIndentation = 4                                         //indent:1 exp:1
+ * forceStrictCondition = true                                         //indent:1 exp:1
+ */                                                                    //indent:1 exp:1
+public class InputIndentationCheckSingleSwitchStatementsWithoutCurly { //indent:0 exp:0
+    void testCorrectIndentation(int obj) {                             //indent:4 exp:4
+        switch (obj) {                                                 //indent:8 exp:8
+            case 0                                                     //indent:12 exp:12
+                ->                                                     //indent:16 exp:16
+                    System.out.println("Test");                        //indent:20 exp:20
+            case 1 ->                                                  //indent:12 exp:12
+                System.out.println("TEST");                            //indent:16 exp:16
+            case 2 -> {                                                //indent:12 exp:12
+                System.out.println("Test");                            //indent:16 exp:16
+            }                                                          //indent:12 exp:12
+        }                                                              //indent:8 exp:8
+    }                                                                  //indent:4 exp:4
+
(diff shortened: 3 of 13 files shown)


Aziz-755:
@romani I did a refork, and now report generation is working perfectly.
Please take a look.

romani:
lineWrappingIndentation=4
not clear why this is violation
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/674efc9_2025201338/reports/diff/camunda/index.html#A1

file looks like using google style, so indentation is valid, am I miss somthing ?

romani:
@pnagy-cldr , @mohitsatr ,  is this potential fix for lambda in case ?

Aziz-755:
> lineWrappingIndentation=4 not clear why this is violation https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/674efc9_2025201338/reports/diff/camunda/index.html#A1
> 
> file looks like using google style, so indentation is valid, am I miss somthing ?

I based the logic in the fix on using `basicOffset `(which is 2 in this case) to compute the indentation for the lambda's child.
I wasn't 100% sure whether to rely on `basicOffset `or `lineWrappingIndentation` since it's not clearly a line continuation, it might actually be treated more like a new "block."

To align with how Checkstyle handles blocks inside switch lambdas with curlies, I tested the same code but added braces `{}` to the lambda:
```java
/*
 * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
 * one or more contributor license agreements. See the NOTICE file distributed
 * with this work for additional information regarding copyright ownership.
 * Licensed under the Camunda License 1.0. You may not use this file
 * except in compliance with the Camunda License 1.0.
 */
package io.camunda.authentication.config;

import io.camunda.security.configuration.SecurityConfiguration;
import io.camunda.security.entity.AuthenticationMethod;

public final class AuthenticationProperties {
  public static final String METHOD = "camunda.security.authentication.method";
  public static final String API_UNPROTECTED = "camunda.security.authentication.unprotected-api";

  private AuthenticationProperties() {}

  public static String getUnprotectedApiEnvVar() {
    return API_UNPROTECTED.replace(".", "_").replace("-", "").toUpperCase();
  }

  public static void applyToSecurityConfig(
      final SecurityConfiguration securityConfig, final String key, final Object value) {
    switch (key) {
      case METHOD -> {
          AuthenticationMethod.parse(String.valueOf(value))
              .ifPresent(securityConfig.getAuthentication()::setMethod);
      }
      case API_UNPROTECTED -> {
          securityConfig
              .getAuthentication()
              .setUnprotectedApi(Boolean.parseBoolean(String.valueOf(value)));
      }
      default -> {}
    }
  }
}
``` 

```
$ java -jar checkstyle-10.21.4-all.jar -c config.xml test.java
Starting audit...
[WARN] D:\Working on Checkstyle\test.java:27:11: 'block' child has incorrect indentation level 10, expected level should be 8. [custom]
[WARN] D:\Working on Checkstyle\test.java:31:11: 'block' child has incorrect indentation level 10, expected level should be 8. [custom]
Audit done.
``` 
So I followed that same logic in the fix. WDYT?

src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting/rule4841indentation/InputSingleSwitchStatementWithoutCurlyCorrect.java
```diff
@@ -0,0 +1,55 @@
...
+      case 1 -> System.out.println("TEST");
+      case 2
+          -> System.out.println("Test");
+      case 3 ->
+        System.out.println("Test");
+      case 4 -> {
+        System.out.println("Test");
+        System.out.println("Another statement");
+      }
+      default -> System.out.println("test");
```
mohitsatr, line 39: @Aziz-755 just curious, If we specify different values for `basicOffset` and `lineWrappingIndentation`, it would cause the indentation of cases children to be misaligned across the switch block, right?

Aziz-755: @mohitsatr So if the case blocks were of this format:
```java
case 1 -> {
  // some code
}
```
or of this format:
 ```java
case 1 -> 
  // some code
```

They will be aligned ( here they are using the `basicOffset ` value )

However if the case was like this:
```java
case 1
    ->  // lineWrapping used here
      int x; // basicOffset used here
``` 
For this case it will be misaligned in regard of the other cases.

mohitsatr:
```java
case 1 -> 
    // should be lineWrapping, not basicOffset
```
according to docs :

_Basic offset indentation is used for indentation inside code blocks( {} ). For any lines that span more than 1, line wrapping indentation is used for those lines after the first_

since there is no code block in this case, we must not use basicOffset property. [look at similar cases](https://github.com/checkstyle/checkstyle/blob/52225c3a46dd3a7e71965c64c7a647b9696beff1/src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationLambda1.java#L4)

```
  void function1(Runnable x) { //indent:2 exp:2
    Runnable r1 = () -> { //indent:4 exp:4
      x.run(); //indent:6 exp:6
    }; //indent:4 exp:4

    Runnable r3 = () -> //indent:4 exp:4
        x.run(); //indent:8 exp:8
```

Aziz-755:
GitHub, generate report

mohitsatr:
@Aziz-755 can we end them in this PR. false-negatives are better than fals-positives.
try to find the reason why check thinks indentation value is wrong, where it is getting the wrong (wrong for us, correct for the Check) indentation value. who's the parent CaseHandler or Lambda Handler ?

romani:
what is strange in this diff report that that there is no removal of violations and bunch is added.

users should see reduction of bad violations, but we looks like add more.
Please recheck of we can avoid new violations.

Aziz-755:
@romani Actually the issue #15098 is about a false negative, so the fix here should in fact add a new violations. 

The newly introduced issues in the diff report are also due to an already existing false positives about the usage of `switch` in lambda. I will take the opportunity to investigate those problems in this PR and see if I can solve them

Aziz-755:
GitHub, generate report

Aziz-755:
GitHub, generate report

Aziz-755:
GitHub, generate report

Aziz-755:

Aziz-755:
GitHub, generate report

Aziz-755:
@romani @mohitsatr I addressed the false positives related to switch expressions when they are used with line wrapping. For example, see the nested switch starting at line 297:

```java
295     return switch (resource) {
296       case brokers ->
297           switch (subResource) {
298             // POST /cluster/brokers/1/partitions/2
299             case partitions ->
300                 ClusterApiUtils.mapOperationResponse(
301                     requestSender
302                         .joinPartition(
303                             new JoinPartitionRequest(
304                                 MemberId.from(String.valueOf(resourceId)),
305                                 subResourceId,
306                                 priority,
307                                 dryRun))
308                         .join());
309             case brokers, changes -> new ResponseEntity<>(HttpStatusCode.valueOf(404));
310           };
``` 

Initially, this false positive was also causing the new logic (fixing the missing violations for case/default elements without curly braces) to incorrectly report violations.
After resolving the false positive, the new changes introduced in this PR now correctly detect violations without producing incorrect violations.
Here is the updated report: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/38b6417_2025192818/reports/diff/index.html

mohitsatr (approved):
Great job. we actually removed all false-positives. 
one last thing, we have only checked the regression on camunda project. Hbase also uses google-style-format, we need to check regression on that too unless there is a reason not to.
@romani

romani:
@Aziz-755 , please fix CI violations.

src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting/rule4841indentation/InputSingleSwitchStatementWithoutCurly.java
```diff
@@ -0,0 +1,60 @@
...
+  void testIncorrectIndentation(int obj) {
+    switch (obj) {
+      case 1
+      ->                            // violation '.* incorrect indentation level 6, expected .* 10.'
+      System.out.println("Test");
+      // violation above '.* incorrect indentation level 6, expected .* 14.'
+    case 2 ->                      // violation '.* incorrect indentation level 4, expected .* 6.'
+System.out.println("Test");        // violation '.* incorrect indentation level 0, expected .* 10.'
+      default -> System.out.println("test");
+    }
```
romani, line 28: this file is auto formatted by google formatter https://github.com/google/google-java-format
https://github.com/checkstyle/checkstyle/actions/runs/14683798023/job/41209815638?pr=16721
https://github.com/checkstyle/checkstyle/blob/master/.ci/google-java-format.sh

checkstyle se violations in code that is auto-formatted, that will make user mad of checkstyle.
I think checkstyle is right on violations.

Please create issue on google formatted and place links above method `// violations until https://github.com/google/google-java-format/issues/XXXX`

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationCheckSingleSwitchStatementsWithoutCurly.java
```diff
@@ -0,0 +1,64 @@
...
+            case 2 -> {                                            //indent:12 exp:12
+                System.out.println("Test");                        //indent:16 exp:16
+            }                                                      //indent:12 exp:12
+        }                                                          //indent:8 exp:8
+    }                                                              //indent:4 exp:4
+
+    void testIncorrectIndentation(int obj) {                       //indent:4 exp:4
+        switch (obj) {                                             //indent:8 exp:8
+            case 1                                                 //indent:12 exp:12
+            ->                                                    //indent:12 exp:16 warn
```
romani, line 31: please do us a favor and alight verticaly `//ident` comments in this file.

romani (changes requested):
good job !!!

last minors:

romani:
GitHub, generate report

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationCheckSingleSwitchStatementsWithoutCurly.java
Aziz-755: Done

src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting/rule4841indentation/InputSingleSwitchStatementWithoutCurly.java
Aziz-755: Thank you for the feedback however I'm not sure I fully understand the issue here.

The file in question was intentionally created with violations to test the check. I did not apply formatting to it.
For reference, here's the properly formatted version without violations: [InputSingleSwitchStatementWithoutCurlyCorrect](https://github.com/checkstyle/checkstyle/pull/16721/files#diff-72b8f85ec4af6a4429afaebab888b17782d5896b5e7b963325ccfcc405422872)

It seems I may be missing the point. Could you please clarify what the concern is?

Aziz-755:
The CI check is complaining that I am misspelling "prefs" however, I am not using any word like that in my code. 
@romani, do you know if I can do something about the CI failure?

src/it/resources-noncompilable/com/google/checkstyle/test/chapter4formatting/rule4841indentation/InputSingleSwitchStatementWithoutCurly.java
romani: My bad, I mis read names of files.
I recall this review items.

romani:

romani:
I have no idea ....very strange
ok please apply patch https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=29224&view=logs&j=6a833393-c551-5000-0916-5000fd3b86e7&t=bdbbd129-77f0-5d73-ce7d-db58d34c8358&l=48
let is be in whitelist

romani (approved):
Ok to merge if CI pass.

Thanks. Lot!

Issue #16630, opened 2025-03-22
JavadocParagraph false violation when <p> is inside of <div></div>


SteLeo1602:
I have read check documentation: https://checkstyle.org/checks/javadoc/javadocparagraph.html#JavadocParagraph
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Test.java
/var/tmp $ cat config.xml
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE module PUBLIC
		  "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
		  "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <module name="TreeWalker">
    <module name="JavadocParagraph"/>
  </module>
</module>
```

/var/tmp $ cat Test.java
```
/**
 * <div>
 * This is the description.
 *
 * <p>
 * This is the paragraph.
 * </p>
 *
 * </div>
 *
 */
class Test {
}
```

/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-X.XX-all.jar -c config.xml Test.java
```
Starting audit...
[ERROR] D:\Projects\work_checkstyle\Test.java:5:4: <p> tag should be preceded with an empty line. [JavadocParagraph]
Audit done.
Checkstyle ends with 1 errors.

```


---

Expect the check to pass <p> inside <div> </div> tags since <p> is already preceded by the empty line.

---

romani:
@Zopsss , do you remember is there was any nuance on such nesting?

romani:
>ATTENTION:
This Check ignores HTML comments.
**The Check ignores all the nested paragraph tags**, it will not give any kind of violation if the paragraph tag is nested.

Zopsss:
> do you remember is there was any nuance on such nesting?

tbh I don't remember as it has been so long since I worked on this check. Though I remember there was some bug in this check that we missed to fix during GSoC but I don't know if this was the that same bug or not.

Zopsss:
> Expect the check to pass `<p>` inside `<div> </div>` tags since `<p>` is already preceded by the empty line.

Yup the `<p>` is preceded by an empty line plus it is nested paragraph so no violations are expected.

[...]

PR #15711, opened 2024-09-26
Issue #15690: fix allowNewlineParagraph at JavadocParagraphCheck 

Zopsss:
fixes #15690

Diff Regression config: https://gist.githubusercontent.com/Zopsss/013f379d4c821f2d624be56212a64575/raw/49e8b85afc1ae91560679fea018936258139024a/javadocparagraph_config.xml
Diff Regression projects: https://raw.githubusercontent.com/checkstyle/contribution/master/checkstyle-tester/projects-to-test-on-for-github-action.properties

PR #15686, opened 2024-09-21
Issue #15685: fixed JavadocParagraph to detect paragraph tag when they have their corresponding closing tag

Zopsss:
fixes #15685 https://github.com/checkstyle/checkstyle/issues/15732

tried fixing the implementation of JavadocParagraph ~might have introduced new false-negatives :p~

I tried to make the Check work same way when the paragraph tag's corresponding closing tag is also present.

when I tried to run build command locally, I got so many errors for different files like checks, filters, etc for `<p>` tag. Most of them were for:

```
<p> tag should be placed immediately before the first word, with no space after.
```

@romani need your help here

---

Diff Regression config: https://gist.githubusercontent.com/Zopsss/013f379d4c821f2d624be56212a64575/raw/5f17ec9186d4421396429ab38cbac3e4663bf4eb/javadocparagraph_config.xml
Diff Regression projects: https://raw.githubusercontent.com/checkstyle/contribution/master/checkstyle-tester/projects-to-test-on-for-github-action.properties

PR #16722, opened 2025-04-01
Issue #16630: fp when <p> is nested

Aziz-755:
Fixes: https://github.com/checkstyle/checkstyle/issues/16630
Fixed false-positive and some already existing tests.
The check was built on the assumption that a nested `<p>` tag means it is nested inside another `<p>` tag.
Now, no violation will be produced if a `<p>` tag is nested inside any other tag.

This PR reintroduces the changes from the original PR (https://github.com/checkstyle/checkstyle/pull/16687), which was closed due to a required refork of the repository.

### DIFF
--- a/src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputFormattedIncorrectJavadocParagraph.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputFormattedIncorrectJavadocParagraph.java
@@ -105,7 +105,6 @@ class InputFormattedIncorrectJavadocParagraph {
          * @see <a href="example.com">Documentation about
          *     <p>GWT emulated source</a>
          */
-        // violation 2 lines above '<p> tag should be preceded with an empty line.'
         boolean emulated() {
           return false;
         }
@@ -136,9 +135,6 @@ class InputFormattedIncorrectJavadocParagraph {
      *   <li>1 should NOT give violation as there is not empty line before
      * </ul>
      */
-    // 2 violations 4 lines above:
-    //  '<p> tag should be placed immediately before the first word'
-    //  '<p> tag should be preceded with an empty line.'
     public void foo() {}
 
     // 2 violations 6 lines below:
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect.java
@@ -101,9 +101,6 @@ class InputJavadocParagraphIncorrect {
          * @see <a href="example.com">
          *     Documentation about <p> GWT emulated source</a>
          */
-        // 2 violations 2 lines above:
-        //  'tag should be placed immediately before the first word'
-        //  'tag should be preceded with an empty line.'
         boolean emulated() {return false;}
 
         // violation 3 lines below 'Empty line should be followed by <p> tag on the next line.'
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect5.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect5.java
@@ -9,8 +9,7 @@ allowNewlineParagraph = false
 package com.puppycrawl.tools.checkstyle.checks.javadoc.javadocparagraph;
 
 public class InputJavadocParagraphIncorrect5 {
-    // violation 3 lines below 'tag should be preceded with an empty line.'
-    // violation 4 lines below 'tag should be preceded with an empty line.'
+
     /**
      * <h1><p>Testing....</h1>
      *
@@ -22,14 +21,9 @@ public class InputJavadocParagraphIncorrect5 {
      *
      * </table>
      */
-    // 2 violations 5 lines above:
-    //  '<p> tag should be placed immediately before the first word'
-    //  '<p> tag should be preceded with an empty line.'
     void fooooo() {}
 
-    // violation 4 lines below 'tag should be preceded with an empty line.'
-    // violation 7 lines below 'tag should be preceded with an empty line.'
-    // violation 12 lines below 'tag should be preceded with an empty line.'
+
     /**
      * <b><p>testtttt.....</b>
      *
@@ -68,7 +62,6 @@ public class InputJavadocParagraphIncorrect5 {
      *  </ul>
      * </ul>
      */
-    // violation 4 lines above 'tag should be preceded with an empty line.'
     void fooooooo() {}
 
     // 2 violations 6 lines below:
--- a/src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputIncorrectJavadocParagraph.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputIncorrectJavadocParagraph.java
@@ -116,9 +116,6 @@ class InputIncorrectJavadocParagraph {
          *
          * @see <a href="example.com">Documentation about <p> GWT emulated source</a>
          */
-        // 2 violations 2 lines above:
-        //  '<p> tag should be placed immediately before the first word'
-        //  '<p> tag should be preceded with an empty line.'
         boolean emulated() {
           return false;
         }
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocParagraphCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocParagraphCheck.java
@@ -247,10 +247,10 @@ public class JavadocParagraphCheck extends AbstractJavadocCheck {
      */
     private static boolean isNestedParagraph(DetailNode tag) {
         boolean nested = false;
-        DetailNode parent = tag;
+        DetailNode parent = tag.getParent();
 
         while (parent != null) {
-            if (parent.getType() == JavadocTokenTypes.PARAGRAPH) {
+            if (parent.getType() == JavadocTokenTypes.HTML_ELEMENT) {
                 nested = true;
                 break;
             }
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocParagraphCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocParagraphCheckTest.java
@@ -81,10 +81,8 @@ public class JavadocParagraphCheckTest extends AbstractModuleTestSupport {
             "87:11: " + getCheckMessage(MSG_TAG_AFTER),
             "97:27: " + getCheckMessage(MSG_LINE_BEFORE),
             "99:13: " + getCheckMessage(MSG_MISPLACED_TAG),
-            "102:36: " + getCheckMessage(MSG_MISPLACED_TAG),
-            "102:36: " + getCheckMessage(MSG_LINE_BEFORE),
-            "112:11: " + getCheckMessage(MSG_TAG_AFTER),
-            "113:11: " + getCheckMessage(MSG_TAG_AFTER),
+            "109:11: " + getCheckMessage(MSG_TAG_AFTER),
+            "110:11: " + getCheckMessage(MSG_TAG_AFTER),
         };
         verifyWithInlineConfigParser(
                 getPath("InputJavadocParagraphIncorrect.java"), expected);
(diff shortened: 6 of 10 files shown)


Aziz-755:
Github, generate report for JavadocParagraph/all-examples-in-one

Aziz-755:
@romani I did a refork and addressed your review comment in this new PR. 
the report is also ready, please have a look

Aziz-755:
Thanks for the comment! Just to clarify, do you mean I should add a test case with a comment like this?
```java
/**
 * The Tiles taglib and framework allows building web pages by assembling reusable
 * pieces of pages, called Tiles. A Tile is usually a simple JSP page.
 *
 * <div class="section">
 *   <h2>Introduction</h2>
 *
 *   <p>The Tiles framework allows building pages by assembling reusable Tiles.
 *   As an example, the page in the next figure can be built by assembling a
 *   header, a footer, a menu, and a body.</p>
 *
 *   <p><img src="doc-files/image001.gif" height="169" width="145" alt="doc-files/image001"></p>
 * </div>
 */
``` 
Also, regarding this part:

> I think one day in future some people might complain that in such cases `being nested in div` should be ignored. 

Do you mean that nesting inside a `<div>` should **not** be ignored?
Because with the current fix, if a `<p>` tag is nested inside a `<div>`, it is ignored, so I just want to confirm the intended behavior.

Aziz-755:
@romani done

romani (approved):
Ok to merge .
Thanks a lot

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect5.java
```diff
@@ -22,14 +21,9 @@ public class InputJavadocParagraphIncorrect5 {
...
      * </table>
      */
-    // 2 violations 5 lines above:
-    //  '<p> tag should be placed immediately before the first word'
-    //  '<p> tag should be preceded with an empty line.'
     void fooooo() {}
 
-    // violation 4 lines below 'tag should be preceded with an empty line.'
-    // violation 7 lines below 'tag should be preceded with an empty line.'
-    // violation 12 lines below 'tag should be preceded with an empty line.'
```
Zopsss, line 32: I don't understand why these 2 violation existed? Were they false positives? please confirm @romani @Aziz-755

The `<p>` tag is already preceded with empty line.

Zopsss (changes requested):
@Aziz-755 for the context, JavadocParagraph was specifically created for google java style guide, it helps us to implement following rule:

https://google.github.io/styleguide/javaguide.html#s7.1.2-javadoc-paragraphs

From that rule:

> HTML tags for other block-level elements, such as `<ul>` or `<table>`, are not preceded with `<p>`.


As per the rule, block-level tags should not be preceded with `<p>` tag. In your PR, you made nested `<p>` to be completely ignored.


Can you create some test cases where `<p>` is followed by a block level tag, and it precedes an inline tag.

Similarly create another test case where `<p>` tag is followed by an inline tag and it precedes a block level tag. 

I checked input files but didn't find these cases, if these test cases already exists then please point them out.

romani:
>As per the rule, block-level tags should not be preceded with <p> tag. In your PR, you made nested <p> to be completely ignored.

if block-level html tag preceded it is `<p><table>`.
we try to improve ignore on `<table><p>`.
Please reconsider your comment, there might be some confusion.

-------

Extension of test cases is always good. @Aziz-755 , is there is already such, just share link (with line number) to them.

Zopsss:
@romani updated my comment and tried to make it more clear about what I'm trying to say.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect5.java
Aziz-755: Yes, I think they were false positives. However, I verified that they only occur when the `<p>` tag is nested. So they should be solved with this PR ( since we ignore any kind of violation.

romani:
Ok, now it is clear 
We should have input that shows that we still validating what Google style requires.

Block html tags:
https://github.com/checkstyle/checkstyle/blob/585c606f8dc2421ec3a2e71b9659da514d916545/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocParagraphCheck.java#L153-L157

Code:
```
// Violation 4 lines below p should not preceed table tag
/*
 * Summary.
*
* <P><table></table>
*/
void foo();
```

Zopsss:
> Now, if the `<p>` tag is **not** nested and is followed by block-level html tag a violation will be given ( I did not touch this functionality ) Some already existing tests to verify this behavior exist in [InputJavadocParagraphIncorrect4](https://github.com/checkstyle/checkstyle/blob/585c606f8dc2421ec3a2e71b9659da514d916545/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect4.java)

No, what we want is, even if P tag is nested if it is preceding a block-level tag then it should produce violation because it is breaking google style guide's tule. You're saying that we're completely ignoring nested P taga which we doesn't want to do. 

In the issue description, the contributor mentioned false-positive for following violation:

> `<p>` tag should be preceded with an empty line. [JavadocParagraph]
Audit done.

This was indeed a false-positive because it is not breaking google style guide's rule yet the Check is still giving violation. 

---

The following test case doesn't match requirements of google style guide:

https://github.com/checkstyle/checkstyle/blob/585c606f8dc2421ec3a2e71b9659da514d916545/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect4.java#L22-L26

Here the inner P tag ( `<p><li>` ) is ignored, this P tag isn't following google style guide rule, it is preceding a block-level tag so it should produce violation. 

The test case is from master branch, not from your branch, meaning that it is an already existing false-negative which we might have missed last year when I tried to fix this check's implementation ( https://github.com/checkstyle/checkstyle/pull/15711, https://github.com/checkstyle/checkstyle/pull/15686 )

---

> @Zopsss I hope I understood your point correctly. If not could you please provide an example?

What we want is, we want to fix the false-positive violation that the contributor mentioned in issue description, let's not go any further than that, if we find more bugs in our Check in future then we'll solve them at that time. As per your current changes, we're still not fully following google style guide rule. 

The false-negative I mentioned above is already existing in our master branch so I'm not asking you to fix it in this PR, as this PR is for #16630 issue, not for my above mentioned false-negative. 

We can open a new issue for that false-negative.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect5.java
```diff
@@ -22,14 +21,9 @@ public class InputJavadocParagraphIncorrect5 {
      *
      * </table>
      */
-    // 2 violations 5 lines above:
-    //  '<p> tag should be placed immediately before the first word'
```
Zopsss, line 26: A false-negative is introduced here, `<p> tag should be placed immediately before the first word` was correct violation because `allowNewlineParagraph` is set to false in input file:

https://github.com/checkstyle/checkstyle/blob/585c606f8dc2421ec3a2e71b9659da514d916545/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect5.java#L4

The violation was for following lines:

https://github.com/checkstyle/checkstyle/blob/585c606f8dc2421ec3a2e71b9659da514d916545/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect5.java#L19-L20

src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputFormattedIncorrectJavadocParagraph.java
```diff
@@ -136,9 +135,6 @@ class InnerPrecedingPtag {
      *   <li>1 should NOT give violation as there is not empty line before
      * </ul>
      */
-    // 2 violations 4 lines above:
-    //  '<p> tag should be placed immediately before the first word'
```
Zopsss, line 140: Another false-negative, it is breaking following part of the rule:

> Each paragraph except the first has `<p>` immediately before the first word, with no space after it.

In our google_checks.xml, we've set `allowNewlineParagraph` to false so we can follow above mentioned part of the rule.

https://github.com/checkstyle/checkstyle/blob/93c0ffc9fb9ad1db10190898f26eab07d447bc54/src/main/resources/google_checks.xml#L363-L365

The violation is for following line:

https://github.com/checkstyle/checkstyle/blob/585c606f8dc2421ec3a2e71b9659da514d916545/src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputFormattedIncorrectJavadocParagraph.java#L134-L135

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect.java
```diff
@@ -101,9 +101,6 @@ class InnerInputJavadocParagraphIncorrect {
          * @see <a href="example.com">
          *     Documentation about <p> GWT emulated source</a>
          */
-        // 2 violations 2 lines above:
-        //  'tag should be placed immediately before the first word'
```
Zopsss, line 105: Fyi: this was a false-positive in our check because `allowNewlineParagraph` is set to true:

https://github.com/checkstyle/checkstyle/blob/585c606f8dc2421ec3a2e71b9659da514d916545/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect.java#L4

So it should not give this violation.

Zopsss (changes requested):
Items to discuss:

Zopsss:
@rdiachenko I would like to have your opinion on this comment: https://github.com/checkstyle/checkstyle/pull/16722#issuecomment-2795736036

Do you agree with me?

Aziz-755:
Okay, so as I understand my fix was built on this false assumption :
~~If a P tag is nested, no violation should be produced.~~
I suggest that we should adjust this on the check description https://checkstyle.org/checks/javadoc/javadocparagraph.html#Description
> The Check ignores all the nested paragraph tags, it will not give any kind of violation if the paragraph tag is nested.

Also this was mentioned in the issue description:
> > There should be violation even there no empty line, as nested paragraphs are ignored.
> 
> I'm assuming you're trying to say `There should no violation expected even if the empty line present was not present`
---
Now I will just to try to fix the false positive based on the fact that the `<p>` is preceded by an empty line so no violations are expected ( it makes no difference if it is nested or not ). 
@Zopsss @romani Could please confirm that I am on the right track ?

src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule712paragraphs/InputFormattedIncorrectJavadocParagraph.java
romani: but tag is nested, so there should be no violation. 
@Zopsss , please re-review.

if user put "P" under some tags , it means it is not top level Paragraph, so no Google rules is applicable.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocparagraph/InputJavadocParagraphIncorrect5.java
romani: they are gone, that is good.

romani:
>No, what we want is, even if P tag is nested if it is preceding a block-level tag then it should produce violation because it is breaking google style guide's tule.

NO, nested P is not a Google's paragraph. We are now very clear in behavior: If P is top/root level, we validate it. If it is nested, it is just some html tag that users needs for rendering.

Zopsss:
> NO, nested P is not a Google's paragraph.

I'm getting confused about what is considered as Google's Paragraph. Can you explain it in some more details please?

`<p>` means a paragraph. So how does a nested `<p>` is different than a Google's Paragraph?

romani:
https://google.github.io/styleguide/javaguide.html#s7.1.2-javadoc-paragraphs

>`7.1.2 Paragraphs One blank line—that is, a line containing only the aligned leading asterisk (*)—appears between paragraphs, and before the group of block tags if present. Each paragraph except the first has <p> immediately before the first word, with no space after it. HTML tags for other block-level elements, such as <ul> or <table>, are not preceded with <p>.`

There is no detail on this in description. So it is up to our interpretation, for now we see problems to handle nested P tags, so we can relax validation to avoid false positives (that might make users angry) and do no violations (users almost never complain), especially in javadoc.
So we give user more freedom on nested.
If users starts to complain, we will ask Google to clarify this behavior in style (this is long process).

I hope this summer we will get project for google style polishing and we will have time to update Guava java library to use  checkstyle, so this will prove that google care about top level P, or we will make issue to do validation more robust.

But main point, while we are weak to make implementation robust, we should not do false-positive(unwanted violation), it make users angry, as it blocks CI/merges.

Zopsss (approved):
Convinced with the above explanation. As @romani said if users complains about this check's behaviour we can improve it in future.

Issue #16155, opened 2025-01-12
use new nio jdk api instead of all api


romani:
from https://github.com/checkstyle/checkstyle/pull/16095

we should use avoid usage of old jdk API, wihtout breaking compatibility and an start to use new api more widely.
violations on our API should be suppressed.

roulpriya:
This issue seems to have been implemented already.

romani:
@roulpriya , I updated issue description with more details.
issue is still open.

roulpriya:
I will work on this.

romani:
Please consider to send few PRs for this issue, to improve step by step, not all at ones.

ivamly:
@romani I'd like to take this on. However, after reviewing the cause of the test failures, I wonder if we should mark it as good second/third issue? It looks self-contained and not so complicated.

If not, I'll work on it. What you think?

romani:
this issue is tricky, better to not have "good" label on it. 
Try it, but please be careful and send few PRs, to make sure we merge simple updates quicker and more complicated later

romani:
Excludes are removed, code is migrated

PR #16723, opened 2025-04-01
Issue #16155: use nio api in PropertiesMacro

AmitKumarDeoghoria:
Fixed Issue https://github.com/checkstyle/checkstyle/issues/16155

Removed the File API usage in Properties macro

Changes Made:
*Removed Filed Usage to use Nio Api

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
@@ -19,8 +19,9 @@
 
 package com.puppycrawl.tools.checkstyle.site;
 
-import java.io.File;
 import java.lang.reflect.Field;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedList;
@@ -112,7 +113,7 @@ public class PropertiesMacro extends AbstractMacro {
     private static String currentModuleName = "";
 
     /** The file of the current module being processed. */
-    private static File currentModuleFile = new File("");
+    private static Path currentModulePath = Paths.get("");
 
     @Override
     public void execute(Sink sink, MacroRequest request) throws MacroExecutionException {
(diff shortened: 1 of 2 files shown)


config/spotbugs-exclude.xml
```diff
@@ -335,4 +335,10 @@
     <Class name="com.puppycrawl.tools.checkstyle.checks.sizes.RecordComponentNumberCheck"/>
     <Bug pattern="AT_STALE_THREAD_WRITE_OF_PRIMITIVE"/>
   </Match>
+  <Match>
+    <!-- Suppress NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE in configureGlobalProperties -->
+    <Class name="com.puppycrawl.tools.checkstyle.site.PropertiesMacro"/>
+    <Method name="configureGlobalProperties"/>
+    <Bug pattern="NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE"/>
```
romani, line 342: We need solutions without suppression.
If you read referenced issues, they recommend to not use class File at all. We are ok to do deeper refactoring. We have green light to change anything except for classes in "api" folder, because it will be breaking changes to users and cause more pain and problems than dealing with fashion in coding.

romani (changes requested):
Items

config/spotbugs-exclude.xml
AmitKumarDeoghoria: You are right @romani While working with spotbugs I got it confused with false positive , I have resolved the issue by checking for null for file and throw error. I think It's resolved now.

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
```diff
@@ -112,7 +114,7 @@ public class PropertiesMacro extends AbstractMacro {
     private static String currentModuleName = "";
 
     /** The file of the current module being processed. */
-    private static File currentModuleFile = new File("");
+    private static File currentModuleFile = Paths.get("").toFile();
```
romani, line 117: idea of migration, it completely abandon File class usage, as new classes of Path, offer full substitution, at least advertised :).

Lets go in deeper refactoring to remove "File" import at all from class.

romani (changes requested):
is this doable ?

src/main/java/com/puppycrawl/tools/checkstyle/site/PropertiesMacro.java
AmitKumarDeoghoria: Thanks for catching that! It is doable but this change would ripple through several files like SiteUtil, so it’s a bit more involved. Happy to tackle it, but want to check: Would you like me to expand the refactor now, or focus just on PropertiesMacro for this PR? Either works for me!

AmitKumarDeoghoria: @romani  can you please review

romani (approved):
Thanks a lot.
It is not complete fix for issue, but good step forward.

nrmancuso (approved):

nrmancuso:
Thank you @AmitKumarDeoghoria !

Issue #16333, opened 2025-02-15
separate examples by horizontal line


romani:
after fix of https://github.com/checkstyle/checkstyle/issues/13345

it become clear that our Examples are massive, and we need to do some separation of examples between each other,  as it is hard to spot where is start and finish of some example. 
this is especially acute problem in https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/74a5109_20250214215907/checks/misc/translation.html#Description

TODO:
add `<hr/>` tag between examples in  template files. 
Challenge is that we need to fix a defect in our xdoc file generator as it produce `<hr/><hr/>` in result xdoc xml file.

vishakha-mali:
I want to try it. I'm on it.

vishakha-mali:
Hey @romani , we have to add `<hr/>`(xml.template) & at same time we have to built logic to avoid duplicates `<hr/>` in (XML).Am I right?
We have to do for all file in [misc](https://github.com/checkstyle/checkstyle/tree/master/src/site/xdoc/checks/misc) or just translation?

romani:
First PR is just in one of xml file, for example Translation.
Second PR, massive update for all examples.

vishakha-mali:
Ok understood and started to work!!

romani:
Feel free to change any class you need to make it work.

sathwikhbhat:
> TODO: add `<hr/>` tag between examples in template files. Challenge is that we need to fix a defect in our xdoc file generator as it produce `<hr/><hr/>` in result xdoc xml file.

@romani Could you kindly provide more details on this? I am trying to understand the solution in case this issue remains idle...

romani:
Just add HR tag to template xdoc files and make it render it properly by all means

sathwikhbhat:
No Problem @vishakha-mali.... Even I am busy with my college schedule right now... 
I just wanted to check what the actual solution to the problem could be...

smita1078:
Hi @romani  Shall I send PR for this issue, as its idle for few weeks..

vishakha-mali:
hey, @smita I'm still left to apply logic.If your works, you should give it try too.

smita1078:
> hey, [@smita](https://github.com/smita) I'm still left to apply logic.If your works, you should give it try too.

Sure, Thanku @vishakha-mali . I'll raise a PR with my logic, and we can evaluate our approaches. Let's proceed with whichever works better.

Zopsss:
[...]

PR #16725, opened 2025-04-01
Issue #16333: separate examples by horizontal line

smita1078:
#16333

- Fixed Duplication of hr tags.
- Manually added hr tag in templates files.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/site/XdocsTemplateParser.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/site/XdocsTemplateParser.java
@@ -105,19 +105,21 @@ public class XdocsTemplateParser extends XdocParser {
     @Override
     protected void handleEndTag(XmlPullParser parser, Sink sink) throws MacroExecutionException {
         final String tagName = parser.getName();
-        if (tagName.equals(DOCUMENT_TAG.toString())) {
-            sink.rawText(parser.getText());
-            sink.body_();
-        }
-        else if (macroName != null
-                && tagName.equals(MACRO_TAG.toString())
-                && !macroName.isEmpty()
-                && !isSecondParsing()) {
-            processMacroEnd(sink);
-            setIgnorableWhitespace(false);
-        }
-        else if (!tagName.equals(PARAM.toString())) {
-            sink.rawText(parser.getText());
+        if (!"hr".equalsIgnoreCase(tagName)) {
+            if (tagName.equals(DOCUMENT_TAG.toString())) {
+                sink.rawText(parser.getText());
+                sink.body_();
+            }
+            else if (macroName != null
+                    && tagName.equals(MACRO_TAG.toString())
+                    && !macroName.isEmpty()
+                    && !isSecondParsing()) {
+                processMacroEnd(sink);
+                setIgnorableWhitespace(false);
+            }
+            else if (!tagName.equals(PARAM.toString())) {
+                sink.rawText(parser.getText());
+            }
         }
     }
 
--- a/src/site/resources/css/site.css
+++ b/src/site/resources/css/site.css
@@ -326,6 +326,20 @@ span.wrapper.inline {
   transition: transform 0.3s ease, opacity 0.3s ease;
 }
 
+.example-separator {
+  border: none;
+  height: 3px;
+  width: 100%;
+  margin: 40px 0;
+  background: linear-gradient(
+    to right,
+    rgba(51, 51, 51, 0) 0%,
+    rgba(51, 51, 51, 1) 10%,
+    rgba(51, 51, 51, 1) 90%,
+    rgba(51, 51, 51, 0) 100%  
+  );
+}
+
 @media screen and (max-width: 823px) {
   #leftColumn {
     float: none;
--- a/src/site/xdoc/checks/annotation/annotationlocation.xml
+++ b/src/site/xdoc/checks/annotation/annotationlocation.xml
@@ -165,7 +165,7 @@ class Example1 {
   @Nonnull @Mock DataLoader loader3;
   // violation above, 'Annotation 'Mock' should be alone on line'
 }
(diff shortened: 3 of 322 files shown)


smita1078:
Github, generate site

romani:
on this Check ismore clear that lines should be above description 
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a4de5db_20250402025139/checks/misc/translation.html#Translation

smita1078:
> lets remove HR right after "Examples" section title.
> 
> lets put HR before example description text, so line will be in between of examples (description + config + java)
> 
> please move all HR lines update to separate commit of same PR. So your PR will have two commits: functional (it can have few HR lines to show that it works ) and supplemental (update for all others).

Understood , on it.

smita1078:
> lets remove HR right after "Examples" section title.
> 
> lets put HR before example description text, so line will be in between of examples (description + config + java)
> 
> please move all HR lines update to separate commit of same PR. So your PR will have two commits: functional (it can have few HR lines to show that it works ) and supplemental (update for all others).

Done, dropping this commit and creating commits as required

src/main/java/com/puppycrawl/tools/checkstyle/site/ExampleMacro.java
```diff
@@ -124,6 +132,111 @@ else if ("raw".equals(type)) {
...
+                    getFileNameSafe(examples.get(examples.size() - 1));
+            hrNeeded = !fileName.equals(lastExampleFile);
+        }
+        if (dirPath.contains("resources/") && !dirPath.contains("noncompilable")
+                && !fileName.equals(getFileNameSafe(examples.get(examples.size() - 1)))) {
+            hrNeeded = true;
+        }
+
+        if (hrNeeded) {
+            sink.rawText(HR_SEPARATOR);
```
romani, line 217: can you think on solution that does not need update website generation code to explicitly HR tag.
template file (xdoc file) is already mix of html tags, HR tag should be just another html tag that we can put in it.

romani (changes requested):
item:

src/main/java/com/puppycrawl/tools/checkstyle/site/ExampleMacro.java
smita1078: Okay, correct me if i am wrong but you are saying that we manually place the hr tag in the... xdoc template files and just change the logic here to simply add that tag rather than generating it during website generation?

smita1078: OKay, got it!

romani:
GitHub, generate website

romani (approved):
I love this update

Zopsss:
We're manually placing hr tags so we should have a test for it so contributor cannot miss to add hr tag in examples.

src/main/java/com/puppycrawl/tools/checkstyle/site/ExampleMacro.java
Zopsss: @romani why did you suggested placing hr tags manually? Adding hr tags can be annoying for contributors, if we can automate this then it'll be beneficial for them.

We should avoid such manual tasks and make them automated. Making manual tasks automated is also part of [Automated Website Generation](https://github.com/checkstyle/checkstyle/wiki/Checkstyle-GSoC-2025-Project-Ideas#project-name-automated-website-generation) project:

> Reduce/eliminate manual documentation updates for examples

We should promote this approach ( making tasks automated ) more.

Zopsss: @smita1078 please also give your opinion

smita1078: Yes, It was cumbersome for me as well, especially with over 200 files to handle. Manual addition not only takes time but also leaves room for errors or missed tags. Automating this process would undoubtedly be more efficient.

smita1078:
@romani @Zopsss this PR is pending, lets finalize the approach (ie. manual or Automated placement of hr tags) and close this..

smita1078:
> HR is decoration, not a content. We have macroses for content placement in final xdoc. It is proven by decades, that templates defined all decoration and layout, and macroses just insert content, in such model all is flexible and adaptive to all edge cases, easy to understand by new engineers.
> 
> This separation decoration is good for now, but in few years it will be changed, to follow fashion.

Okay👍

Zopsss:
> This separation decoration is good for now, but in few years it will be changed, to follow fashion.

I get that you're saying we should not automate hr tah placement but I didn't understood your this sentence. Can you elaborate it more?

I'm still not convinced to make hr tags manually placed, but if you insist, we should write a test to ensure that hr tag is always present it is placed in correct position.

smita1078:
> https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/ae205f8_20250412020205/checks/misc/translation.html#Examples
> 
> Is example on where it becomes super obvious that we need separation. Location of this separation is not very clear.

>we should write a test to ensure that hr tag is always present it is placed in correct position.

In such case as location is not standard rather based on where the user requires... I am not sure if writing a test would work bcz it is not standard and we will have to cover multiple cases that might change in future.

Zopsss:
> In such case as location is not standard rather based on where the user requires... I am not sure if writing a test would work bcz it is not standard and we will have to cover multiple cases that might change in future.

@romani please confirm, should we add test to check placements of hr tag separation? Or not

romani:
If we want to add test on this, let's do this in separate PR.
There might be nuances with proper validation for it, as some cases contains few snippets, not only two.

Zopsss (approved):
Let's add test in separate PR

romani:
Bulk of work is done, as user notice problems with separation, they will report and we will fix it.

PR #16755, opened 2025-04-04
Issue #13999: Resolved Pitest Suppression in JavadocNodeImpl

Prarthana-Ramesh:
Issue: #13999 

Resolved Pitest Suppression in [JavadocNodeImpl.java](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocNodeImpl.java)

**Mutation** - `toString : removed call to java/util/Objects::hashCode → SURVIVED`

- Objects.hashCode(children) returns the array's identity hash code, not a content-based hash. In assertions, this leads to incorrect evaluations since the actual contents of the array are ignored.

- Static analysis tools flag this under the ArrayHashCode bug pattern when used in logical contexts like assert. The same call in toString() is not flagged, as it's treated as non-functional and purely informational.
 
- Despite this, using Objects.hashCode() limits functionality even in toString(), as it doesn't reflect the true structure of the array.
 Thus replaced with Arrays.hashCode(children) to compute a hash based on array contents.
 
- This ensures semantic correctness and helps eliminate false positives in mutation testing.

**Mutation killed** -> `removed call to java/util/Objects::hashCode → KILLED`

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocNodeImplTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocNodeImplTest.java
@@ -21,6 +21,8 @@ package com.puppycrawl.tools.checkstyle.checks.javadoc;
 
 import static com.google.common.truth.Truth.assertWithMessage;
 
+import java.util.Arrays;
+
 import org.junit.jupiter.api.Test;
 
 import com.puppycrawl.tools.checkstyle.api.JavadocTokenTypes;
@@ -33,13 +35,19 @@ public class JavadocNodeImplTest {
         javadocNode.setType(JavadocTokenTypes.CODE_LITERAL);
         javadocNode.setLineNumber(1);
         javadocNode.setColumnNumber(2);
+        final JavadocNodeImpl child1 = new JavadocNodeImpl();
+        final JavadocNodeImpl child2 = new JavadocNodeImpl();
+        child1.setType(JavadocTokenTypes.CODE_LITERAL);
+        child2.setType(JavadocTokenTypes.CODE_LITERAL);
+        javadocNode.setChildren(child1, child2);
 
         final String result = javadocNode.toString();
 
         assertWithMessage("Invalid toString result")
             .that(result)
             .isEqualTo("JavadocNodeImpl[index=0, type=CODE_LITERAL, text='null', lineNumber=1,"
-                + " columnNumber=2, children=0, parent=null]");
+                + " columnNumber=2, children=" + Arrays.hashCode(javadocNode.getChildren())
+                + ", parent=null]");
     }
 
     @Test
--- a/config/error-prone-suppressions/compile-phase-suppressions.xml
+++ b/config/error-prone-suppressions/compile-phase-suppressions.xml
@@ -1,11 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedErrors>
-  <error>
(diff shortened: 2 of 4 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocNodeImplTest.java
```diff
@@ -21,25 +21,37 @@
...
+        final JavadocNodeImpl child2 = new JavadocNodeImpl();
+        child1.setType(JavadocTokenTypes.CODE_LITERAL);
+        child2.setType(JavadocTokenTypes.CODE_LITERAL);
         javadocNode.setType(JavadocTokenTypes.CODE_LITERAL);
+
+        final DetailNode[] children = {child1, child2};
+
         javadocNode.setLineNumber(1);
         javadocNode.setColumnNumber(2);
+        javadocNode.setChildren(children);
```
pankratz76, line 46:
```suggestion
        final JavadocNodeImpl child1 = new JavadocNodeImpl();
        final JavadocNodeImpl child2 = new JavadocNodeImpl();
        child1.setType(JavadocTokenTypes.CODE_LITERAL);
        child2.setType(JavadocTokenTypes.CODE_LITERAL);
        javadocNode.setChildren({child1, child2});
```
Maybe declare where its actually used, to increase cohesion and reduce coupling.
Ideally, the API would take the type through the constructor, so we could reduce it a lot more.

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocNodeImplTest.java
```diff
@@ -21,25 +21,37 @@
...
     public void testToString() {
         final JavadocNodeImpl javadocNode = new JavadocNodeImpl();
+        final JavadocNodeImpl child1 = new JavadocNodeImpl();
+        final JavadocNodeImpl child2 = new JavadocNodeImpl();
+        child1.setType(JavadocTokenTypes.CODE_LITERAL);
+        child2.setType(JavadocTokenTypes.CODE_LITERAL);
         javadocNode.setType(JavadocTokenTypes.CODE_LITERAL);
+
+        final DetailNode[] children = {child1, child2};
+
```
pankratz76, line 43:
```suggestion
```

src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocNodeImplTest.java
```diff
@@ -21,25 +21,37 @@
...
 
 public class JavadocNodeImplTest {
 
     @Test
     public void testToString() {
         final JavadocNodeImpl javadocNode = new JavadocNodeImpl();
+        final JavadocNodeImpl child1 = new JavadocNodeImpl();
+        final JavadocNodeImpl child2 = new JavadocNodeImpl();
+        child1.setType(JavadocTokenTypes.CODE_LITERAL);
+        child2.setType(JavadocTokenTypes.CODE_LITERAL);
```
pankratz76, line 39:
```suggestion
```

pankratz76 (approved):
Maybe declare where its actually used, to increase cohesion and reduce coupling.
Ideally, the API would take the type through the constructor, so we could reduce it a lot more.

Prarthana-Ramesh:
@pankratz76 agreed. have updated the commit accordingly.

PR #16764, opened 2025-04-05
Issue #14084: Resolved CheckerFramework violation

Atharv3221:
Issue: #14084

**Suppression removed:**
```
    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java</fileName>
    <specifier>return</specifier>
    <message>incompatible types in return.</message>
    <lineContent>return returnValue;</lineContent>
    <details>
      type of expression: @Initialized @Nullable DetailAST
      method return type: @Initialized @NonNull DetailAST
    </details>
  </checkerFrameworkError>

```

### DIFF
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -269,17 +269,6 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java</fileName>
     <specifier>return</specifier>
     <message>incompatible types in return.</message>
-    <lineContent>return returnValue;</lineContent>
-    <details>
-      type of expression: @Initialized @Nullable DetailAST
-      method return type: @Initialized @NonNull DetailAST
-    </details>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java</fileName>
-    <specifier>return</specifier>
-    <message>incompatible types in return.</message>
     <lineContent>return returnList;</lineContent>
     <details>
       type of expression: @Initialized @Nullable List&lt;@Initialized @NonNull Token&gt;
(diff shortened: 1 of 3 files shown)


pankratz76 (approved):

Rohanraj123:
Hi @Atharv3221 

In such cases, we should ideally ensure that the return expression is non-null. Simply marking the method as @Nullable can reduce code readability and shift the burden to callers. Annotations should act like clear tags — they should "scream" the intent and constraints of a function or variable.

Here, the method get() returns entries.get(key). Since the Map.get() can return null, the Checker sees the expression as @Nullable. But, if we know the key is always present (e.g., the map is populated before use), we should explicitly enforce non-null return.

The key point is to make the expression non-null, not just the method signature, so the CF doesn't complain.

Atharv3221:
Okay, thankyou for making the picture more clear for me, I will work on this issue.

Atharv3221:
Okay, thankyou for making the picture more clear for me, I will work on this issue.

src/main/java/com/puppycrawl/tools/checkstyle/JavadocDetailNodeParser.java
```diff
@@ -94,6 +96,7 @@ public class JavadocDetailNodeParser {
      *        DetailAST of Javadoc comment
      * @return DetailNode tree of Javadoc comment
      */
+    @Nullable
     public ParseStatus parseJavadocAsDetailNode(DetailAST javadocCommentAst) {
```
romani, line 100: Ini might miss something, but I don't see how this can be null, we create instance of class to return.
It is notnull.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/JavadocDetailNodeParser.java
Atharv3221: Okay

romani:
@Atharv3221 , do you have time to finish this PR ?

Atharv3221:
@romani 
Here
```
    /**
     * Returns the first child token that makes a specified type.
     *
     * @param type the token type to match
     * @return the matching token, or null if no match
     */
    @Nullable
    DetailAST findFirstToken(int type);
```
can return null if no token is found of type we want, so it can be Nullable 

am i going in correct direction ??

romani:
https://github.com/checkstyle/checkstyle/actions/runs/16948287343/job/48034726010?pr=16764#step:7:4455

resolving:
```
    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java</fileName>
    <specifier>return</specifier>
    <message>incompatible types in return.</message>
    <lineContent>return returnValue;</lineContent>
    <details>
      type of expression: @Initialized @Nullable DetailAST
      method return type: @Initialized @NonNull DetailAST
    </details>
  </checkerFrameworkError>
```

results in numerous:
https://github.com/checkstyle/checkstyle/actions/runs/16948287343/job/48034726010?pr=16764#step:7:4455

```
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java</fileName>
+    <specifier>dereference.of.nullable</specifier>
+    <message>dereference of possibly-null reference typeParam.findFirstToken(TokenTypes.IDENT)</message>
+    <lineContent>typeParam.findFirstToken(TokenTypes.IDENT).getText());</lineContent>
+  </checkerFrameworkError>
```

and Checker is correct, this code is not null safe.

romani:
please apply patch from https://github.com/checkstyle/checkstyle/actions/runs/16948287343/job/48034726010?pr=16764#step:7:4455

we can not fix now all code to be safe, but new code will be aware from the begining that method is possible to return null, so they/Checks will handle possible null

romani:
apply https://github.com/checkstyle/checkstyle/actions/runs/17016853045/job/48240648033?pr=16764#step:7:12

Atharv3221:
> apply https://github.com/checkstyle/checkstyle/actions/runs/17016853045/job/48240648033?pr=16764#step:7:12

Yeah its just taking time because big diff is making the task confusing

.github/workflows/checker-framework.yml
```diff
@@ -57,6 +57,12 @@ jobs:
         if: failure()
         run: ./.ci/print-diff-as-patch.sh target/checker-${{ matrix.profile }}.patch
 
+      - name: Show new file
+        if: failure()
+        run: |
+          echo "=== NEW (after checker run) ==="
+          cat config/checker-framework-suppressions/checker-${{ matrix.profile }}-suppressions.xml || true
```
romani: this is not expected, please explain

romani (changes requested):
:

.github/workflows/checker-framework.yml
Atharv3221: temporary will be removed after getting whole new file at once, this avoids manual inserting 1 by 1

romani: ok, I hope you know that you need just simply copy from https://github.com/checkstyle/checkstyle/actions/runs/17025078445/job/48259362312?pr=16764#step:7:12 until end  and just paste in terminal, that is all.

Atharv3221: oh okay i really didn't knew that, i thought that just shows diff

romani (approved):
thanks a lot.
Looks like we shold make a priority task in such updates
to update all classes with Nulable annotations that are from our API classes.

We can not change them, so placing annotation  is good signal to all our users (who write extensions with extra Checks).

it will trigger such big waves of suppressions and it will actually reveal actual places to fix in our code

Atharv3221:
Thanks, this PR helped me understand the checker framework. I can open a follow-up issue which will prioritise API classes.

romani:
attention to classes that extends/inherit from ".api." package as it was with DetailAstImpl.java

PR #16769, opened 2025-04-05
Kill mutation for CheckstyleAntTask

smita1078:
#14019

- A surviving mutant was detected in `getListeners()` where the call to `formatters.size()` was removed 
(and replaced with a constant), causing the listener count to always be 1 even when multiple formatters are configured.

- To kill this mutant, added a new test (testMultipleFormatterListenerCount()) that configures 
the CheckstyleAntTask with two formatters (an XML formatter and a SARIF formatter), executes the task, 
and then asserts that getListenerCountForTesting() returns 2. If the mutant is present, the listener 
count would be incorrect (i.e. 1), and the test would fail.

- This change ensures that the logic for determining the number of 
listeners is based on the actual size of the formatter list. 

```
<mutation unstable="false">
    <sourceFile>CheckstyleAntTask.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask</mutatedClass>
    <mutatedMethod>getListeners</mutatedMethod>
    <description>removed call to java/util/List::size</description>
    <lineContent>final int formatterCount = Math.max(1, formatters.size());</lineContent>
  </mutation>
```
Before:

Screenshot 2025-04-05 121003

```
Smita Prajapati@Smita-5ESU2EG5 MINGW64 ~/checkstyle (KillMutant)
$ groovy .ci/pitest-survival-check-xml.groovy "$PITEST_PROFILE"

Source File: "CheckstyleAntTask.java"
Class: "com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask"
Method: "getListeners"
Line Contents: "final int formatterCount = Math.max(1, formatters.size());"
Mutator: "org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator"
Description: "removed call to java/util/List::size"
Line Number: 454
```
After:
Screenshot 2025-04-05 121502

```
$ groovy .ci/pitest-survival-check-xml.groovy "$PITEST_PROFILE"
No new surviving mutation(s) found.
```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
@@ -239,6 +239,17 @@ public class CheckstyleAntTask extends Task {
         executeIgnoredModules = omit;
     }
 
+    /**
+     * Returns the number of listeners created by getListeners().
+     *
+     * <p>Public access is intentional to allow test access.</p>
+     *
+     * @return the number of formatter-based listeners.
+     */
+    public int getListenerCountForTesting() {
+        return getListeners().length;
+    }
+
     ////////////////////////////////////////////////////////////////////////////
     // Setters for Root Module's configuration attributes
     ////////////////////////////////////////////////////////////////////////////
--- a/config/pitest-suppressions/pitest-ant-suppressions.xml
+++ b/config/pitest-suppressions/pitest-ant-suppressions.xml
@@ -75,15 +75,6 @@
   <mutation unstable="false">
     <sourceFile>CheckstyleAntTask.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask</mutatedClass>
-    <mutatedMethod>getListeners</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to java/util/List::size</description>
-    <lineContent>final int formatterCount = Math.max(1, formatters.size());</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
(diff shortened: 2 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
```diff
@@ -239,6 +239,17 @@ public void setExecuteIgnoredModules(boolean omit) {
...
 
+    /**
+     * Returns the number of listeners created by getListeners().
+     *
+     * <p>Public access is intentional to allow test access.</p>
+     *
+     * @return the number of formatter-based listeners.
+     */
+    public int getListenerCountForTesting() {
+        return getListeners().length;
```
romani, line 250: No. We can not change code in favor of test, to only test benefits it.

Try
```

        final AuditListener[] listeners = new AuditListener[formatters.size()];

```

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
smita1078: Understood,  @romani  I made the changes, but I am unable to generate pitest report..
image
Gives me this error , it was working fine earlier, I am facing this error today itself..

romani: Do `mvn clean `. Each mvn should have `clean`

smita1078: Did mvn clean, but this error still persists

smita1078: Also it is giving this error,

image

But I see there is no issue with the TokenUtil tests..

Issue #16786, opened 2025-04-07
`private` enums being treated as public in `JavadocVariableCheck`


JackPGreen:
Take a simple class with a `private` enum:
```
public class MyClass {
    private enum PrivateEnum {
        CONSTANT
    }
}
```

When run with a simple `checkstyle.xml`:
```
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="TreeWalker">
        <module name="JavadocVariable">
            <property name="accessModifiers" value="public"/>
        </module>
    </module>
</module>
```

`[ERROR] MyClass.java:3:9: Missing a Javadoc comment. [JavadocVariable]` is thrown.

While investigating, the change in https://github.com/checkstyle/checkstyle/pull/16049 identifies any enum constant as being implicitly public (correct), but doesn't check if the enum itself is accessible (which it isn't, because `private`).

Specifically, while upgrading checkstyle from `10.21.4` to `10.22.0` many such pre-existing private enums were now being flagged as violations.

romani:
@kkoutsilis , do you see a problem that @JackPGreen  try to point on ?

kkoutsilis:
Hey @JackPGreen, thanks for raising this issue. 

I believe this is the intended behavior of the check. Going by the [Java specification](https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.9.3), enum constants are implicitly declared public static final. 

In the example you have provided, even though the enum itself is `private`, the constant `CONSTANT` is technically `public`. As the check does not account for the enclosing scope, the violation is expected.

JackPGreen:
> I believe this is the intended behavior of the check. Going by the [Java specification](https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.9.3), enum constants are implicitly declared public static final.

That makes sense - but they aren't _actually_ publicly accessible. You can't access `PrivateEnum.CONSTANT` from... anywhere else.

> In the example you have provided, even though the enum itself is `private`, the constant `CONSTANT` is technically `public`, inside the enum block definition.

Ok - I think I understand now. I don't agree with the behaviour in this scenario, but I think the issue isn't "`private` enums being treated as public" but rather "`public` _fields_ in `private` classes being treated as public".

[...]

PR #16787, opened 2025-04-07
Issue #16786: Ensure enum constants accessibility is properly evaluated

JackPGreen:
Extends change in https://github.com/checkstyle/checkstyle/pull/16049, in the case of an enum constant, it's accessibility will be derived based on the enclosing enum (which could be `private`) rather than being assumed to be implicitly public.

Test(s) added to validate.

Fixes: https://github.com/checkstyle/checkstyle/issues/16786
Closes: https://github.com/checkstyle/checkstyle/pull/17112

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
@@ -314,20 +314,27 @@ public final class CheckUtil {
      * @return the access modifier of the method/constructor.
      */
     public static AccessModifierOption getAccessModifierFromModifiersToken(DetailAST ast) {
-        AccessModifierOption accessModifier;
-        if (ast.getType() == TokenTypes.ENUM_CONSTANT_DEF) {
-            accessModifier = AccessModifierOption.PUBLIC;
-        }
-        else {
-            final DetailAST modsToken = ast.findFirstToken(TokenTypes.MODIFIERS);
-            accessModifier = getAccessModifierFromModifiersTokenDirectly(modsToken);
+        // In some scenarios we want to investigate a parent AST instead
+        DetailAST selectedAst = ast;
+
+        if (selectedAst.getType() == TokenTypes.ENUM_CONSTANT_DEF) {
+            // Enum constants don't have modifiers
+            // implicitly public but validate against parent(s)
+            while (selectedAst.getType() != TokenTypes.ENUM_DEF) {
+                selectedAst = selectedAst.getParent();
+            }
         }
 
+        final DetailAST modsToken = selectedAst.findFirstToken(TokenTypes.MODIFIERS);
+        AccessModifierOption accessModifier =
+                getAccessModifierFromModifiersTokenDirectly(modsToken);
+
         if (accessModifier == AccessModifierOption.PACKAGE) {
-            if (ScopeUtil.isInEnumBlock(ast) && ast.getType() == TokenTypes.CTOR_DEF) {
+            if (ScopeUtil.isInEnumBlock(selectedAst)
+                    && selectedAst.getType() == TokenTypes.CTOR_DEF) {
                 accessModifier = AccessModifierOption.PRIVATE;
             }
-            else if (ScopeUtil.isInInterfaceOrAnnotationBlock(ast)) {
+            else if (ScopeUtil.isInInterfaceOrAnnotationBlock(selectedAst)) {
                 accessModifier = AccessModifierOption.PUBLIC;
             }
         }
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocVariableCheck.java
@@ -44,7 +44,8 @@ import com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil;
  * such as package-private for fields without an explicit modifier.
  * It also accounts for special cases where fields have implicit modifiers,
  * such as {@code public static final} for interface fields and {@code public static}
- * for enum constants. Only fields matching the specified modifiers will be analyzed.
+ * for enum constants, or where the nesting types accessibility is more restrictive and hides the
+ * nested field. Only fields matching the specified modifiers will be analyzed.
  * Type is {@code com.puppycrawl.tools.checkstyle.checks.naming.AccessModifierOption[]}.
  * Default value is {@code public, protected, package, private}.
  * </li>
@@ -93,7 +94,8 @@ public class JavadocVariableCheck
      *  This includes both explicitly declared modifiers and implicit ones, such as package-private
      *  for fields without an explicit modifier. It also accounts for special cases where fields
      *  have implicit modifiers, such as {@code public static final} for interface fields and
-     *  {@code public static} for enum constants.
+     *  {@code public static} for enum constants, or where the nesting types accessibility is more
+     *  restrictive and hides the nested field.
      *  Only fields matching the specified modifiers will be analyzed.
      */
     private AccessModifierOption[] accessModifiers = {
@@ -111,7 +113,8 @@ public class JavadocVariableCheck
      * checked. This includes both explicitly declared modifiers and implicit ones, such as
      * package-private for fields without an explicit modifier. It also accounts for special
      * cases where fields have implicit modifiers, such as {@code public static final}
-     * for interface fields and {@code public static} for enum constants.
+     * for interface fields and {@code public static} for enum constants, or where the nesting
+     * types accessibility is more restrictive and hides the nested field.
      * Only fields matching the specified modifiers will be analyzed.
      *
      * @param accessModifiers access modifiers of fields to check.
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/javadoc/JavadocVariableCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/javadoc/JavadocVariableCheck.xml
@@ -16,7 +16,8 @@
  such as package-private for fields without an explicit modifier.
  It also accounts for special cases where fields have implicit modifiers,
  such as {@code public static final} for interface fields and {@code public static}
- for enum constants. Only fields matching the specified modifiers will be analyzed.</description>
+ for enum constants, or where the nesting types accessibility is more restrictive and hides the
+ nested field. Only fields matching the specified modifiers will be analyzed.</description>
             </property>
             <property name="ignoreNamePattern" type="java.util.regex.Pattern">
                <description>Specify the regexp to define variable names to ignore.</description>
--- a/src/site/xdoc/checks/javadoc/javadocvariable.xml
+++ b/src/site/xdoc/checks/javadoc/javadocvariable.xml
@@ -27,7 +27,7 @@
             </tr>
             <tr>
               <td>accessModifiers</td>
-              <td>Specify the set of access modifiers used to determine which fields should be checked. This includes both explicitly declared modifiers and implicit ones, such as package-private for fields without an explicit modifier. It also accounts for special cases where fields have implicit modifiers, such as <code>public static final</code> for interface fields and <code>public static</code> for enum constants. Only fields matching the specified modifiers will be analyzed.</td>
+              <td>Specify the set of access modifiers used to determine which fields should be checked. This includes both explicitly declared modifiers and implicit ones, such as package-private for fields without an explicit modifier. It also accounts for special cases where fields have implicit modifiers, such as <code>public static final</code> for interface fields and <code>public static</code> for enum constants, or where the nesting types accessibility is more restrictive and hides the nested field. Only fields matching the specified modifiers will be analyzed.</td>
               <td><a href="../../property_types.html#AccessModifierOption.5B.5D">AccessModifierOption[]</a></td>
               <td><code>public, protected, package, private</code></td>
               <td>10.22.0</td>
@@ -85,6 +85,13 @@ public class Example1 {
   public int d; // violation, 'Missing a Javadoc comment'
   /*package*/ int e; // violation, 'Missing a Javadoc comment'
 
+  public enum PublicEnum {
+    CONSTANT // violation, 'Missing a Javadoc comment'
+  }
+
+  private enum PrivateEnum {
+    CONSTANT // violation, 'Missing a Javadoc comment'
+  }
 }
 </code></pre></div><hr class="example-separator"/>
         <p id="Example2-config">
@@ -116,6 +123,14 @@ public class Example2 {
   protected int c;
   public int d; // violation, 'Missing a Javadoc comment'
   /*package*/ int e;
+
+  public enum PublicEnum {
+    CONSTANT // violation, 'Missing a Javadoc comment'
+  }
+
+  private enum PrivateEnum {
+    CONSTANT
+  }
 }
 </code></pre></div><hr class="example-separator"/>
         <p id="Example3-config">
@@ -147,6 +162,14 @@ public class Example3 {
   protected int c;
   public int d;
   /*package*/ int e; // violation, 'Missing a Javadoc comment'
+
+  public enum PublicEnum {
+    CONSTANT
+  }
+
+  private enum PrivateEnum {
+    CONSTANT // violation, 'Missing a Javadoc comment'
+  }
 }
 </code></pre></div><hr class="example-separator"/>
         <p id="Example4-config">
@@ -179,6 +202,14 @@ public class Example4 {
   protected int c; // violation, 'Missing a Javadoc comment'
   public int d; // violation, 'Missing a Javadoc comment'
   /*package*/ int e; // violation, 'Missing a Javadoc comment'
+
+  public enum PublicEnum {
+    CONSTANT // violation, 'Missing a Javadoc comment'
+  }
+
+  private enum PrivateEnum {
+    CONSTANT // violation, 'Missing a Javadoc comment'
+  }
 }
 </code></pre></div>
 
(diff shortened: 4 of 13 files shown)


romani:
@JackPGreen , CI is very red, please address it.

JackPGreen:
> @JackPGreen , CI is very red, please address it.

Thanks, fixed.

https://github.com/checkstyle/checkstyle/pull/16787/commits/f44a879172dbad8f5ec16de20feb1ac1e7346e7c

romani:
@JackPGreen , please squash commits in one, CI does not like it, we need green CI

JackPGreen:
> @JackPGreen , please squash commits in one, CI does not like it, we need green CI

Done :)

romani:
@kkoutsilis , @mahfouz72 , please review update.
and please review my comment in issue, I placed soft approval, but if you do not think so, we can discuss it a bit more.

Main point here is not Java spec, but more about javadoc tool behavior and user experience on documentations, no need to document what is not placed in html pages of  javadoc. If user do care so much on doc on enum, he should put it at top level class.

JackPGreen:
> Main point here is not Java spec, but more about javadoc tool behavior and user experience on documentations, no need to document what is not placed in html pages of javadoc. If user do care so much on doc on enum, he should put it at top level class.

I think there are two separate issues here:

1. https://github.com/checkstyle/checkstyle/pull/16049 caused an unintended change in behaviour _for enums only_ in a patch release (unexpected) which this PR addresses (which prevented me from upgrading Checkstyle in my project)

2. [_it appears_ that inaccessible `public` fields (whether `enum, `class` etc) aren't being handled as well as they could be](https://github.com/checkstyle/checkstyle/issues/16786#issuecomment-2800167506), and a wider fix could fix all the issues at once.

It's up to you whether (1) is worth addressing in isolation.

mahfouz72:
> @kkoutsilis , @mahfouz72 , please review update.
> and please review my comment in issue

I am ok with the update as it addresses the enum case. However, this issue is broader. We should verify whether a variable is truly accessible. So if we are going to fix this, we should handle all such cases, not just enums.

Docs need to be updated also because right now it doesn't tell anything about true accesibility, it just talks about the direct access modifier of the variable


I noticed that @kkoutsilis tried to handle such cases by using concepts like [surroundingAccessModifier](https://github.com/checkstyle/checkstyle/pull/16049#issuecomment-2561290132).

But he got these reviews.

> if we need advanced filtering based on outer access modifier, let's do it as a separate step, when we have real use cases

> I think we should look at class/type's fields own visibility, completely ignoring all upper types visibility/modifiers., Let's not  consider any non target visibility. On edge cases, users might use suppressions.


@kkoutsilis Will it be easy to bring back your work on trying to handle the surrounding/outer scope issue?

romani:
I recommend to remove violation as hot fix, by merge of this PR, and in separate issue to do better support.
For now it is better to remove false-positive quicker.
@MohanadKh03 , do you agree ?

MohanadKh03:
> For now it is better to remove false-positive quicker.

I agree as well the changes in this PR looks good but I just wonder what about other tokens in the future ?

> This is long and heavy problem to define scope/visibility based on all upper classes.
I think we should look at class/type's fields own visibility, completely ignoring all upper types visibility/modifiers.

At least for the current issue of enums I see this PR is good to go but what about other tokens ? should they be handled separately in different PRs ? I think that this is not a big issue and user could just document the whole thing in an upper class for most of the tokens in the future but after taking a quick look on codebases that have inner types (classes, enums, etc..) they mostly document the inner class not just within the upper class e.g: Spring's codebase is full of this pattern

> To make it even more complicated, modern java has a lot not explicit visibilities, we have separate Check on this, and it is very complicated by its own.

Since it is complicated on its own we might as well just handle enums for now and start discussing other tokens **if needed to handle them at all**

kkoutsilis:
Hey, to give my two cents here, this PR looks good for addressing the enums problem @JackPGreen faced, but this makes the check a bit confusing, as in all other cases it checks for direct access modifier, though for enums it checks the surrounding access modifier. 

The docs should be updated to clarify this behaviour before merging this PR.

Overall, we probably need to decide if we keep the direct access modifier or refactor to use the surrounding access modifier for all cases. I don't have a strong opinion right now on one or the other, I just prefer consistency for all cases overall when possible.

A good indication would be user feedback. If there are more confused or blocked users by this check's behaviour, then it's worth spending time on it. If there are no further complaints and users are happy, maybe it's not worth the discussion. 

Hope this makes sense.

> @kkoutsilis Will it be easy to bring back your work on trying to handle the surrounding/outer scope issue?

@mahfouz72 If we decide to go with it, I can have a look. As you see from the previous PR I faced some issues with it 😅, but it should be possible.

james-baker-aera:
@kkoutsilis 
> A good indication would be user feedback. If there are more confused or blocked users by this check's behaviour, then it's worth spending time on it. If there are no further complaints and users are happy, maybe it's not worth the discussion.

I'm here to say that this is stopping us upgrading to a more recent version of Checkstyle (without turning the `JavadocVariableCheck` off). We have a lot of package-private enums without Javadoc that `10.22.0` and after flag as violations.

_(I found this PR while looking into the violations.)_

romani:
Thanks a lot for feedback.
Please stay on previous version, we will merge this PR soon, we just need to align on expected behavior. I will do release right after this fix is merged.

@james-baker-aera, can you check on your project that this fix is covering all your false positives?
If you have open source project, please share link we will put it in testing efforts to make sure no regression.

romani:
>The docs should be updated to clarify this behaviour before merging this PR.

@JackPGreen , please update documentation to mention that we still consider interpretation of javadoc tool , when evaluating accessibility.
We can add small enum to examples, and make a comment above it, that javadoc tool doesn't consider it public, so Check does the same

JackPGreen:
> > The docs should be updated to clarify this behaviour before merging this PR.
> 
> @JackPGreen , please update documentation to mention that we still consider interpretation of javadoc tool , when evaluating accessibility. We can add small enum to examples, and make a comment above it, that javadoc tool doesn't consider it public, so Check does the same

I'm not sure that's sensible in it's current state - with this change, accessibility is only "properly" evaluated (i.e. considering parents) for `enum`s - [but other scenarios (e.g. a `public` member of a `private` `class`), it isn't](https://github.com/checkstyle/checkstyle/issues/16786#issuecomment-2800167506) - so it's not accurate to say that it's consistent with Javadoc tooling.

I think improving the documentation would be better once ^ that ^ is properly addressed and it's simpler to explain.

romani:
if we extend documentation to claim that we consider parents visiblity (no mentioning of javadoc tool), it will be enough.
and we can create new issue on fields of classes, as you described at https://github.com/checkstyle/checkstyle/issues/16786#issuecomment-2800167506 , so it will be know defect in Check and we can address it in separate PR, as looks like it less bother people than enum.
Parents visibility matters for javadoc tool, one of the major usages for people who write javadoc.

On future, there will be group of people who wants to not consider javadoc tool at all, as they use only IDE abilities to render javadoc during navigation over code. We will think in future on how to reconcile such use cases, lets proceed with focus on javadoc tool.

JackPGreen:
> if we extend documentation to claim that we consider parents visiblity (no mentioning of javadoc tool), it will be enough. and we can create new issue on fields of classes, as you described at [#16786 (comment)]

Happy to add documentation, but struggling with the wording.

I [updating the docs](https://checkstyle.sourceforge.io/checks/javadoc/javadocvariable.html#Properties), changing:
> It also accounts for special cases where fields have implicit modifiers, such as public static final for interface fields and public static for enum constants.

To:

> It also accounts for special cases where fields have implicit modifiers, such as public static final for interface fields and public static for enum constants, or where the enclosing types' accessibility is more restrictive and hides the field.

_Might_ help? Suggestions welcomed.

romani:
`or where the nesting types accessibility is more restrictive and hides the nested field.`

Please add  to all examples your junit test code.
```
public enum PublicEnum {
        CONSTANT // violation, 'Missing a Javadoc comment'
    }

    private enum PrivateEnum {
        CONSTANT
    }
```

JackPGreen:
> `or where the nesting types accessibility is more restrictive and hides the nested field.`
> 
> Please add to all examples your junit test code.
> 
> ```
> public enum PublicEnum {
>         CONSTANT // violation, 'Missing a Javadoc comment'
>     }
> 
>     private enum PrivateEnum {
>         CONSTANT
>     }
> ```

PR  updated.

> @JackPGreen , please squash commits in one, CI does not like it, we need green CI

But squashed into one commit because of ^ this ^

romani:
[ERROR] [checkstyle] [ERROR] C:\projects\checkstyle\src\site\xdoc\checks\javadoc\javadocvariable.xml:241: Trailing whitespace is not allowed [noTrailingWhitespace]

romani:
Before pushing, run on local `mvn clean verify `

JackPGreen:
> Before pushing, run on local `mvn clean verify `

Thanks, was testing locally but with the wrong goal (`checkstyle:checkstyle`). _Hopefully_ all fixed now:
> [INFO] BUILD SUCCESS

romani:
Add to commit results on test execution, test phase generate actual xdoc pages

Patch at https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=29287&view=logs&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111&l=1850

JackPGreen:
> Add to commit results on test execution, test phase generate actual xdoc pages

Not sure I follow, sorry?

There were some checkstyle/CI issues with the docs I was unable to reproduce locally but all addressed now.

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
```diff
@@ -314,20 +314,27 @@ public static boolean isReceiverParameter(DetailAST parameterDefAst) {
      * @return the access modifier of the method/constructor.
      */
     public static AccessModifierOption getAccessModifierFromModifiersToken(DetailAST ast) {
```
romani, line 316: This method is used in lot of Checks , lets not extend this fix to others 
https://github.com/search?q=repo%3Acheckstyle%2Fcheckstyle%20getAccessModifierFromModifiersToken&type=code

Let's keep this fix isolated to single Check.
You can copy this util method to Check, and apply change to Check special method.

romani (changes requested):
If you would like to add more text to website 
Add it to https://github.com/checkstyle/checkstyle/blob/master/src/site/xdoc/checks/javadoc/javadocvariable.xml.template
Run tests, it will update xdoc file , add it to commit.

Items:

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
JackPGreen: > This method is used in lot of Checks , lets not extend this fix to others https://github.com/search?q=repo%3Acheckstyle%2Fcheckstyle%20getAccessModifierFromModifiersToken&type=code
> 
> Let's keep this fix isolated to single Check. You can copy this util method to Check, and apply change to Check special method.

[The PR that exposed this issue changed ^ that ^ method so I think updating it again makes sense in case there are other places the same issue crept in](https://github.com/checkstyle/checkstyle/pull/16049/files#diff-aa33dbbc3a6914d54f622e36874631d45e7f15c73b5d3b4369bd576d9beabd17).

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/Example6.java
```diff
@@ -0,0 +1,20 @@
+/*xml
```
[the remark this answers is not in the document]
JackPGreen: > We need to add such enums to existing Example1.... Example5 New example is not expected.

Sorry, I don't understand - each example is different, why they _all_ need the `enum` example?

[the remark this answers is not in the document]
romani: https://checkstyle.org/checks/javadoc/javadocvariable.html#Examples

They are the same. Difference only in config and trailing comments.
In this model users clearly see how the same code, highlighted differently with change in config

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
romani: It might be great miss in review of that PR.

If we change common method, now we need to re check all affected Checks that this new logic is good for them.
Most of them are not javadoc so, same though might be not applicable.

Javadoc check affected is only JavadocMethod.
https://checkstyle.org/checks/javadoc/javadocmethod.html
That use that same property.

Ok, lets do copy of method , keep in Until class, but new method will be used for javadoc only. In other case, we need to do diff testing for all affected Checks. I hope to do diff testing only for 2 javadoc Checks.

JackPGreen: > It might be great miss in review of that PR.
> 
> If we change common method, now we need to re check all affected Checks that this new logic is good for them. Most of them are not javadoc so, same though might be not applicable.

I understand your concern, but those checks have _already_ been affected in an unexpected way without checking.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocvariable/Example6.java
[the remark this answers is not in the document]
JackPGreen: Thanks, fixed.

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
romani: usage is big:
image

and it might be concern for non-javadoc Check, as we are conflict of interpretation for javadoc and non-javadoc.
Functionality before this PR was focused on java only, and we naively thought that it will work for javadoc, that is why we have this issue.
Now we are clear that javadoc has nuance on it, and only javadoc related checks should be affected by this update.

does it make sense?

romani: @kkoutsilis , @rdiachenko , please join this conversation, to give us second opinion on this conflict of opinions.

romani:
Github, generate report for UnusedLocalVariable/all-examples-in-one

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
JackPGreen: > does it make sense?

Can you give an example where you think it doesn’t?

I think that this definition of `private` is better, but I’m biased as a _user_ of check style thinking about my specific problem.

if there’s a scenario where you think it doesn’t make sense I’d be much more receptive.

romani:
Github, generate report for RecordComponentNumber/all-examples-in-one

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
romani: We already can not predict and find by  diff testinng ease sceanrios.
Whole reason of this issue exists,  as we did not predicted it, diff showed nothing.

We can merge this, and in few moth one more person will be frustrated that we did not account him,

romani:
Github, generate report for JavadocVariable/all-examples-in-one

romani:

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
kkoutsilis: Before [this](https://github.com/checkstyle/checkstyle/pull/16049/files#diff-aa33dbbc3a6914d54f622e36874631d45e7f15c73b5d3b4369bd576d9beabd17) change, if I remember correctly, this method was raising an error as it could not evaluate the access modifier of an ENUM_CONSTANT_DEF. Based on that, I don't believe any other checks are affected, unless a scenario is not tested. 

If I understand correctly, the conflict of opinions here is whether we should copy the method with the new changes to be used only for Javadoc checks, or update the existing method, which is used in a plethora of checks.

As I see it we could make a copy of the method for javadoc checks with the new updates, and revert the method to be used in other checks before [these](https://github.com/checkstyle/checkstyle/pull/16049/files#diff-aa33dbbc3a6914d54f622e36874631d45e7f15c73b5d3b4369bd576d9beabd17) changes, this will probably satisfy both sides. 

In my mind, it does not make a big difference, as the access modifier of  ENUM_CONSTANT_DEF does not seem to be evaluated anywhere else apart from this check, but just to be safe, the approach of copy for javadoc and revert for the rest should be the safer option. 

@romani @JackPGreen Did I get that correct? What do you think?

JackPGreen: > As I see it we could make a copy of the method for javadoc checks with the new updates, and revert the method to be used in other checks before [these](https://github.com/checkstyle/checkstyle/pull/16049/files#diff-aa33dbbc3a6914d54f622e36874631d45e7f15c73b5d3b4369bd576d9beabd17) changes, this will probably satisfy both sides.
> 
> In my mind, it does not make a big difference, as the access modifier of ENUM_CONSTANT_DEF does not seem to be evaluated anywhere else apart from this check, but just to be safe, the approach of copy for javadoc and revert for the rest should be the safer option.
> 
> @romani @JackPGreen Did I get that correct? What do you think?

I think adding another path with the previous undesirable behaviour just makes everything more complicated... but if it's two project maintainers vs one contributor I guess it is what it is.

romani: @JackPGreen , can you create new PR with split usage for method. Creation of new method might result in coverage problem, that might be not easy to cover, especially pitest.

But let give it a try, and we might be lucky. So we will have two PRs to merge one or another.

I will estimate risk vs efforts to make subjective decisions based on my experience.

romani:
All reports are empty on diff except for JavadocVariable, JavadocVariable diff report looks good.

The only concern is vague possibility on affecting other Checks, that is not common code at all, as we don't see it in diff.

src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
JackPGreen: > @JackPGreen , can you create new PR with split usage for method. Creation of new method might result in coverage problem, that might be not easy to cover, especially pitest.
> 
> But let give it a try, and we might be lucky. So we will have two PRs to merge one or another.



https://github.com/checkstyle/checkstyle/pull/17112

PR #16799, opened 2025-04-09
doc: improve variableDeclarationUsageDistanceChec's docs to avoid confusion

### LINKED ISSUE #16680: Update description for VariableDeclarationUsageDistance
I have read check documentation: https://checkstyle.org/checks/xxxxxx/nameofaffectedcheck.html

```
$ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        </module>
    </module>
</module>

$ cat Main.java 
/** some javadoc. */
public class Main {

  /**
   * some javadoc.
   */
    int pi = 0;                                                                 // declare
    int square = 32;
    int root = 8;
    int[] profit = {1, 0, -1};
    System.out.println("Starting calculation...");
    System.out.println("Iterating...");
    String message = "flow";
    pi = 2;                                                                       // first usage 
    int currentIndex = 0;
    for (int i = 1; i <= 5; i++) {
      currentIndex += 1;
    }
    System.out.println("Sum: ");
  }
}

$ java -jar testing/checkstyle-10.21.4-all.jar -c config.xml Main.java 
Starting audit...
[ERROR] /mnt/5D92528E6B945467/test/Main.java:8:5: Distance between variable 'pi' declaration
 and its first usage is 3, but allowed 1.  Consider making that variable final if you still need to
 store its value in advance (before method calls that might have side effects on the original value). 
[VariableDeclarationUsageDistance]
Audit done.
```

**Problem.**
Correct distance between `pi`'s declaration and first usage is 6. for some reason `square`, `root`, `message` variables are not accounted in the calculation. try to remove them or declare additional variables, you will get the same distance value.

mohitsatr:
fixes #16680

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
@@ -37,7 +37,8 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
 /**
  * <div>
  * Checks the distance between declaration of variable and its first usage.
- * Note: Variable declaration/initialization statements are not counted while calculating length.
+ * Note: Any additional variables declared or initialized between the declaration and
+ *  the first usage of the said variable are not counted when calculating the distance.
  * </div>
  * <ul>
  * <li>
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/VariableDeclarationUsageDistanceCheck.xml
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/coding/VariableDeclarationUsageDistanceCheck.xml
@@ -6,7 +6,8 @@
              parent="com.puppycrawl.tools.checkstyle.TreeWalker">
          <description>&lt;div&gt;
  Checks the distance between declaration of variable and its first usage.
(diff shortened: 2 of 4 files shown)


romani (approved):
Ok to merge.
Let's check with others.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
```diff
@@ -37,7 +37,8 @@
 /**
  * <div>
  * Checks the distance between declaration of variable and its first usage.
- * Note : Variable declaration/initialization statements are not counted while calculating length.
+ * Note : Any additional variables declared or initialized between the declaration and
```
pankratz76, line 40: Typo fix `1 vs 20` might be worth fixing / aligning:

1:
image

20:
image

nrmancuso (approved):
Thank you!

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
mohitsatr: done

pankratz76 (approved):
thx

PR #16808, opened 2025-04-10
Issue #11163: Enforce file size for IndentationInvalidMethodIndent

Praveen7294:
Issue #11163

Enforce file size for `InputIndentationInvalidMethodIndent.java` splits into 2 class i.e `InputIndentationInvalidMethodIndent1.java` and `InputIndentationInvalidMethodIndent2.java`, both file size are below 120 lines of code.

## Build output
image


romani (approved):
Thanks a lot.
All good.

Please send update for all impots in folder `indentation`

romani:
@Praveen7294 , in issue we ask to do update per Check, in other words in all Inputs of single Check, you did update in Indentation Check for single Input.
Such update was fine as you got confirmation that you doing it in right way, so now you apply the same update to other files

PR #16488, opened 2025-03-06
Issue #16421: `S1488`: Local variables should not be declared and then immediately returned or thrown

Pankraz76:
issue: https://github.com/checkstyle/checkstyle/issues/16421

prerequisite: https://github.com/checkstyle/checkstyle/pull/16435

**the rspec number `S1488` and the incremented PR id almost match #16488. Someone make the math probability for this**

PR #14596, opened 2024-03-03
Issue #14084: Remove checkerframework violation for DefaultConfigurat…

Rohanraj123:
related #14084

PR #16812, opened 2025-04-10
Issue #14084: Remove CF violation for DefaultConfiguration.java

Rohanraj123:
part of #14084

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/DefaultConfiguration.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/DefaultConfiguration.java
@@ -25,6 +25,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.checkerframework.common.value.qual.ArrayLen;
+
 import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
 import com.puppycrawl.tools.checkstyle.api.Configuration;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
@@ -36,9 +38,6 @@ public final class DefaultConfiguration implements Configuration {
 
     /** A unique serial version identifier. */
     private static final long serialVersionUID = 1157875385356127169L;
-
-    /** Constant for optimization. */
-    private static final Configuration[] EMPTY_CONFIGURATION_ARRAY = new Configuration[0];
 
     /** The name of this configuration. */
     private final String name;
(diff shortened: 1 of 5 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/DefaultConfiguration.java
```diff
@@ -103,8 +102,8 @@ public String getProperty(String propertyName) throws CheckstyleException {
 
     @Override
     public Configuration[] getChildren() {
-        return children.toArray(
-                EMPTY_CONFIGURATION_ARRAY);
+        final Configuration @ArrayLen(0) [] emptyArray = new Configuration[0];
+        return children.toArray(emptyArray);
```
pankratz76, line 106:
```suggestion
        return children.toArray(new Configuration[0]);
```
The type looks quite impressive, do we need the explizit boilerplate typing here for specific reasons?

This will be done anyway in #16488 as violating against RSPEC-S1488

introduce InlineVariableCheck - Local variables should not be declared and then immediately returned or thrown - (RSPEC-S1488)

src/main/java/com/puppycrawl/tools/checkstyle/DefaultConfiguration.java
```diff
@@ -103,8 +102,8 @@ public String getProperty(String propertyName) throws CheckstyleException {
 
     @Override
     public Configuration[] getChildren() {
-        return children.toArray(
-                EMPTY_CONFIGURATION_ARRAY);
+        final Configuration @ArrayLen(0) [] emptyArray = new Configuration[0];
```
pankratz76, line 105:
```suggestion
        final Configuration @ArrayLen(0)[] emptyArray = new Configuration[0];
```
I would expect `NoWhitespaceBefore`, but its not actually finding that much as figured out in #16810, to fail the build for the whitespace.

pankratz76 (approved):
thx

Rohanraj123:
Continuation of https://github.com/checkstyle/checkstyle/pull/14596

Rohanraj123:
@romani ping

This update will help resolve too many warnings related to `toArray().`.

src/main/java/com/puppycrawl/tools/checkstyle/DefaultConfiguration.java
```diff
@@ -38,7 +40,8 @@ public final class DefaultConfiguration implements Configuration {
     private static final long serialVersionUID = 1157875385356127169L;
 
     /** Constant for optimization. */
-    private static final Configuration[] EMPTY_CONFIGURATION_ARRAY = new Configuration[0];
+    private static final Configuration@ArrayLen(0)[] EMPTY_CONFIGURATION_ARRAY =
+            new Configuration[0];
```
romani, line 44: code becomes so ugly 
we need to think on other solution or workaround

Issue #16798, opened 2025-04-09
google style coverage page unnecessary blue tick in "Caught exceptions: not ignored"  rule.


Z4ID-Reaper:
I have read check documentation: https://checkstyle.org/checks/blocks/emptycatchblock.html#EmptyCatchBlock
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

### Context :

![Image](https://github.com/user-attachments/assets/334f79cd-aa3c-49ed-9b85-8075fa60b432)
there seems to be two ticks here 
 
the rule in the google style :  

> Except as noted below, it is very rarely correct to do nothing in response to a caught exception. (Typical responses are to log it, or if it is considered "impossible", rethrow it as an AssertionError.)
> When it truly is appropriate to take no action whatsoever in a catch block, the reason this is justified is explained in a comment
>  In tests, a caught exception may be ignored without comment if its name is or begins with expected.

---
below shown are cases i tested : 

- empty catch block without comment, where regex does not match
- empty catch block with comment, where regex does not match

- empty catch block without comment, where "expected" regex match.
- empty catch block with comment, where "expected" regex match.

---
_i used the config that was in the google_style.xml_

### config.xml

```
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN" 
   "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <module name="TreeWalker">
    <module name="EmptyCatchBlock">
      <property name="exceptionVariableName" value="expected"/>
    </module>
  </module>  
</module>
```
---
### Test.java

```
public class Example1 {
  private void exampleMethod1() {
    try {
      throw new RuntimeException();
    } catch (RuntimeException abcd) {
    } // violation above 'Empty catch block'
  }

  private void exampleMethod2() {
    try {
      throw new RuntimeException();
    } catch (RuntimeException abcd) {
      // no handling
    } // ok, catch block has comment
  }

    private void exampleMethod3() {
    try {
      throw new RuntimeException();
    } catch (RuntimeException expected) {
    } // ok, 'Empty catch block' matches the regex.
  }
    
    private void exampleMethod4() {
    try {
      throw new RuntimeException();
    } catch (RuntimeException expected) {
        // no handling
    } // ok, 'Empty catch block' matches the regex.
  }
[...]

PR #16818, opened 2025-04-10
Issue #16798: Removed blue tick from google style

Z4ID-Reaper:
references issue : #16798 

**Detailed analysis is provided in the issue description , for the reason of this PR.**


Zopsss:
@Z4ID-Reaper I've left some comments on the issue, maybe we need to close this PR as some work is required to fully implement the rule.

Z4ID-Reaper:
sure , there are indeed some work remain to cover.  
i would close this pr for now.

Issue #16822, opened 2025-04-11
Remove unused deprecated inValidOn method from `JavaDocTagInfo` class


Abdelrhmansersawy:
The `inValidOn` method in the `JavaDocTagInfo` class is deprecated and no longer used in the codebase. This issue removes the unused deprecated method along with its redundant references in tag definitions to clean up the code and improve maintainability.

### List of Removed Usages

The following `JavaDocTagInfo` constants were previously referencing the deprecated `inValidOn` method and have now been cleaned up:

- [ ] `AUTHOR("@author", "author", Type.BLOCK)`  
- [ ] `CODE("{@code}", "code", Type.INLINE)`  
- [ ] `DEPRECATED("@deprecated", "deprecated", Type.BLOCK)`  
- [ ] `EXCEPTION("@exception", "exception", Type.BLOCK)`  
- [ ] `LINK("{@link}", "link", Type.INLINE)`  
- [ ] `LINKPLAIN("{@linkplain}", "linkplain", Type.INLINE)`  
- [ ] `RETURN("@return", "return", Type.BLOCK)`  
- [ ] `DOC_ROOT("{@docRoot}", "docRoot", Type.INLINE)`  
- [ ] `LITERAL("{@literal}", "literal", Type.INLINE)`  
- [ ] `PARAM("@param", "param", Type.BLOCK)`  
- [ ] `SEE("@see", "see", Type.BLOCK)`  
- [ ] `SERIAL("@serial", "serial", Type.BLOCK)`  
- [ ] `SERIAL_FIELD("@serialField", "serialField", Type.BLOCK)`  
- [ ] `SINCE("@since", "since", Type.BLOCK)`  
- [ ] `THROWS("@throws", "throws", Type.BLOCK)`  
- [ ] `VALUE("{@value}", "value", Type.INLINE)`  
- [ ] `VERSION("@version", "version", Type.BLOCK)`

Abdelrhmansersawy:
@romani 
What's the best approach to remove the deprecated `isValidOn` implementation for unused constants?

I have a potential solution, and I would like to discuss it before proceeding.

1. We can provide a default implementation for `isValidOn` by converting it from an abstract method to a regular method.
 The default implementation would return `false`. This way, if any constant requires a custom implementation, it can still override the method.
   ```java
   public boolean isValidOn(DetailAST ast) { return false; }
   ```
2. For example, the enum constant that doesn't use `isValidOn` would look like this:
   ```java
   /**
[...]

PR #16827, opened 2025-04-11
Remove deprecated AUTHOR Tag from JavaDocTagInfo 

Abdelrhmansersawy:
Part of Issue #16822

In this PR, I first discuss my approach to remove deprecated `isValidOn` from unused constants in `JavaDocTagInfo`.


Abdelrhmansersawy:
@romani 
Could you review the approach I suggested for removing the deprecated usages of the `isValidOn` method?

One potential disadvantage of this approach is that we would need to test the default `inValidOn` method, even if it is not used, in order to meet the Checkstyle coverage line rules.

I have another possible solution that could help avoid this issue. We could define an interface, `JavadocTagValidator`, at the end of the class. If an enum needs to use this method, it can implement the interface. This would allow us to add new tags that do not require the `isValidOn` method without needing to test it, as they wouldn't be using it.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagInfo.java
```diff
@@ -69,16 +69,7 @@ public enum JavadocTagInfo {
...
-
-        @Override
-        public boolean isValidOn(final DetailAST ast) {
-            final int astType = ast.getType();
-            return astType == TokenTypes.PACKAGE_DEF
-                || TokenUtil.isTypeDeclaration(astType);
-        }
-
-    },
+    AUTHOR("@author", "author", Type.BLOCK),
```
romani, line 72: Please share details why we can not simply remove this enum value

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagInfo.java
Abdelrhmansersawy: Since we still use the tag `AUTHOR.getName()` here:
```cpp
checkTag(ast, tags, JavadocTagInfo.AUTHOR.getName(),
                            authorFormat);
```
in the `JavadocTypeCheck` class,  
but we no longer use the `isValidOn` method for this tag.

romani: Can we move this code to such Check?

Abdelrhmansersawy: @romani 
> Can we move this code to such Check?
Not sure!  
For the `AUTHOR` tag, we can, since it's only used in `JavadocTypeCheck`.  
However, other tags are used in multiple checks, such as:  
- `JavadocTypeCheck`  
- `MissingDeprecatedCheck`  
- `JavadocTag`  
- `JavadocStyleCheck`  
- `JavadocMethodCheck`

Abdelrhmansersawy: I think the best approach here is:

### Step 1: Remove deprecated tags  
Here's the list of deprecated tags to be removed:
- `CODE`  
- `DOC_ROOT`  
- `LITERAL`  
- `SERIAL`  
- `SINCE`  
- `VALUE`  
- `VERSION`

### Step 2: Remove usage of deprecated `isValidOn` methods  
This step will be very useful, as only three tags currently override the `isValidOn` method. The remaining usage is in unit tests:

image

To solve this, we can follow one of two approaches:

#### Option 1: Set a default implementation for `isValidOn`  
```java
public boolean isValidOn(DetailAST ast) {
    return false;
}
```

#### Option 2: Define an interface for tags that require `isValidOn`  
```java
CODE("{@code}", "code", Type.INLINE) implements JavadocTagInfo {

    @Override
    public boolean isValidOn(final DetailAST ast) {
        final int astType = ast.getType();
        return DEF_TOKEN_TYPES.get(astType)
            && !ScopeUtil.isLocalVariableDef(ast);
    }
}

interface JavadocTagValidator {
    boolean isValidOn(DetailAST ast);
}
```

Abdelrhmansersawy: @romani Ping!

romani: It is definitely good to remove all, but you will not be able to do this, as it will take a lot of time and efforts. Better vto find step be step approach. 

Let's try to remove AUTHOR and put whole logic it needs to JavadocTypeCheck. If this becomes doable , all others can follow similar type of update.

Abdelrhmansersawy: @romani 
Ping!
Let's finish this PR. I’ve left a comment explaining why we can't remove everything and put this logic directly into the check.

For now, the best approach is the second one I discussed here:
[https://github.com/checkstyle/checkstyle/pull/16827#discussion\_r2041119521](https://github.com/checkstyle/checkstyle/pull/16827#discussion_r2041119521)

This approach has the most benefits - it's not costly at all and removes all deprecated usage without a high risk of breaking other functionality.

romani:
rebased in web to let CI show log.

romani:
fix some test by:
```
 ~/java/github/romani/checkstyle [Abdelrhmansersawy/deprecatedMethod L|✚ 1] 
07:24 $ git diff
diff --git a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagTest.java b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagTest.java
index 80f31c21a0..e666b9b1e2 100644
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagTest.java
@@ -61,13 +61,13 @@ public class JavadocTagTest {
 
     @Test
     public void testToString() {
-        final JavadocTag javadocTag = new JavadocTag(0, 1, "author", "firstArg");
+        final JavadocTag javadocTag = new JavadocTag(0, 1, "inheritDoc", "");
 
         final String result = javadocTag.toString();
 
         assertWithMessage("Invalid toString result")
             .that(result)
-            .isEqualTo("JavadocTag[tag='author' lineNo=0, columnNo=1, firstArg='firstArg']");
+            .isEqualTo("JavadocTag[tag='inheritDoc' lineNo=0, columnNo=1, firstArg='']");
     }
 
     @Test
@@ -82,7 +82,7 @@ public class JavadocTagTest {
 
     @Test
     public void testJavadocTagReferenceImportsInvalid() {
-        assertThat(new JavadocTag(0, 0, "author", null).canReferenceImports())
+        assertThat(new JavadocTag(0, 0, "inheritDoc", null).canReferenceImports())
                 .isFalse();
     }

```

please keep working on JavadocTypeCheckTest to find reason of failures.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagInfo.java
romani: >if (tag.getTagName().equals(tagName)) {

put in this code checking for "author" tag , just extra condition for loop. So Author tag will be only in this check.

in other words
```
 private void checkTag(DetailAST ast, Iterable<JavadocTag> tags, String tagName,
                          Pattern formatPattern) {
```
should know that Author tags is not present in JavadocTag and special code is required to handle it.

Abdelrhmansersawy:
Closed after https://github.com/checkstyle/checkstyle/issues/16822#issuecomment-2894264976 as we will migrate OFF such class this summer, as we will write new grammar for javadoc and rewrite all old checks to use new AST.

PR #16829, opened 2025-04-11
Pull #16829: align ubuntu image version `ubuntu-22.04` -> `ubuntu-24.04`

### LINKED ISSUE #16680: Update description for VariableDeclarationUsageDistance

pankratz76:
Pull #16829: align ubuntu image version `ubuntu-22.04` -> `ubuntu-24.04`

10 

image

vs.

15

image

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,7 +26,7 @@ strategy:
 
     # spelling
     'spelling':
-      image: 'ubuntu-22.04'
+      image: 'ubuntu-24.04'
       cmd: "./.ci/test-spelling-unknown-words.sh"
       skipCache: true
 
--- a/.github/workflows/check-pr-description.yml
+++ b/.github/workflows/check-pr-description.yml
@@ -11,7 +11,7 @@ concurrency:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/no-old-refs.yml
+++ b/.github/workflows/no-old-refs.yml
@@ -21,7 +21,7 @@ concurrency:
 jobs:
   check_issues:
     if: github.repository == 'checkstyle/checkstyle'
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Download checkstyle
         uses: actions/checkout@v4
--- a/.github/workflows/release-new-milestone-and-issues-in-other-repos.yml
+++ b/.github/workflows/release-new-milestone-and-issues-in-other-repos.yml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   new-milestone-and-issues-in-other-repos:
     name: New Milestone, Create issues for ${{ inputs.version }}
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Checkout the latest code
         uses: actions/checkout@v4
--- a/.github/workflows/release-publish-releasenotes-twitter.yml
+++ b/.github/workflows/release-publish-releasenotes-twitter.yml
@@ -24,7 +24,7 @@ concurrency:
 jobs:
   publish-releasenotes-twitter:
     name: Publish Release Notes on Twitter ${{ inputs.version }}
-    runs-on: ubuntu-22.04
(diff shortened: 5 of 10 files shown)


pankratz76:
its kind of the same and used to its might worth a try

azure-pipelines.yml
```diff
@@ -100,14 +100,6 @@ strategy:
...
       cmd: "JAVA_HOME=$JAVA_HOME_17_X64 mvn -e --no-transfer-progress verify"
 
-    # moved back to Travis till we find a way to keep secrets in azure
-    # ensure that all modules are used in no exception configs
-    # 'verify-no-exception-configs':
-    #   image: 'ubuntu-24.04'
-    #   cmd: "./.ci/validation.sh verify-no-exception-configs"
-    #   skipCache: true
-    #   needXmlstarlet: true
-
```
nrmancuso, line 110: Please restore this

nrmancuso (changes requested):
Item:

azure-pipelines.yml
pankratz76: Thanks! Yes, I would suggest extracting this into a dedicated issue (#16835), as the codebase should exclusively contain living (active) code—not dead and clunky overhead like this.

Currently, this represents a technical debt item that needs to be addressed later.

Storing such issues in a dedicated tracker improves cohesion and aligns with the SOC (Separation of Concerns) principle we follow.

nrmancuso: “SOC” can also be interpreted as: this change is not related in any way to updating the Ubuntu version, and certainly doesn’t belong hidden in the same commit/pull request.

Again, please revert this change.

romani: this is moved to semaphore actually
https://github.com/checkstyle/checkstyle/blob/720de881c200cf18390c1729cb75829b28a9a161/.semaphore/semaphore.yml#L39-L42

romani: no need to move to Travis , let is sit in semaphore.
to make it more transparent for why we did change, we can make separate commit in this PR and remove it in separate commit with message "infra: verify-no-exception-configs is in semaphore ci"

nrmancuso: Yes, we can remove in a second commit.

romani: @pankratz76 , please just do second commit, to preserve quick and brief reason of removal

pankratz76: #16841

pankratz76: > “SOC” can also be interpreted as: this change is not related in any way to updating the Ubuntu version, and certainly doesn’t belong hidden in the same commit/pull request.

you are perfectly right. Making SOC in commits and PR´s is clearly not my strength.

nrmancuso (approved):
I believe that some of these will only run on merge to main/release workflows, we will need to keep an eye out after merge.

pankratz76:
thanks for the seamless merge, my good friend

Issue #16825, opened 2025-04-11
Website layout breaks at viewport width >= 1999px


SethFalco:
**Describe what you expect in detail.**

The [Checkstyle](https://checkstyle.org) website puts the main content below the navigation menu when the viewport width is >= 1999px. However, it should maintain the same layout as it has on 1920px viewport.

I've encountered this issue as I use a 5120×1440 monitor, which the Checkstyle website is tiled to take half of the display, so around 2560px in width.

**Screenshots**

Observe that the bullet points are going inside of the navigation bar:

![](https://github.com/user-attachments/assets/4a2aec69-b38f-441a-b469-45ab6c72a43a)

Observe that after the starting introduction to the Check, there is unreasonable whitespace, the the rest of the documentation goes under the navigation menu instead: 

![](https://github.com/user-attachments/assets/dd7dbc49-93b0-4d70-bacd-70ae9c77d532)

Zoomed out overview:

![](https://github.com/user-attachments/assets/4243e38f-6df7-43cd-bd08-5e5ca894d304)

---

If approved, willing to contribute the fix myself.

romani:
We recently did a lot for mobile mode, small screen, this is how I use it to answer all users questions, from phone. 

I never opened website on big screen.

PR #16839, opened 2025-04-12
 chore: use span classes for layout 

SethFalco:
Fixes the original issue reported in the issue, but also does 2 other small chores. As they were minor, I thought I could propose them in separate commits of this PR, and you can ofc ask me to remove them if you disagree.

## Use Span Classes for Layout (Commit 1)

The Maven Fluido Skin is based on Bootstrap, which uses a 12 column grid for layout. By hardcoding widths/margins for `leftColumn` and `bodyColumn`, it conflicted with the expectations of the theme which caused the issue.

Trying to patch this would've resulted in more patching later to fight with the theme, so I removed these and replaced it with `span{1...12}` classes at various breakpoints, as the design system is intended to be used.

> When using sidebar, the navigation column uses by default `span2` class and body column uses `span10`. **To prevent weird/ugly result, the sum must be 12**. You can change that using:
> 
> — https://maven.apache.org/skins/maven-fluido-skin/index.html (emphasis mine)

More info here on Bootstrap columns: https://getbootstrap.com/docs/5.0/layout/columns/

In short, it's better to stick to the API, and then ensure the `spanX` and `spanY` on `leftColumn` and `bodyColumn` respectively always add up to 12, but the number of columns they take up can ofc vary across breakpoints.

_(This is typically done with CSS classes like `span4-large` and `span10-xxlarge`, rather than JavaScript, which lets you specify the columns to span per breakpoint. However, Apache Fluido doesn't provide classes like that, so we're just managing it ourselves in JavaScript instead.)_

### Typo in File Name (Commit 2)

You referenced a file called `images/header-right-ruller.png`, however this was presumably meant to be named `images/header-right-ruler.png`, so I just renamed it.

I also cleared out the `alt` tag, as this image is purely decorative.

The alt tag is for accessibility, to describe important images for users that aren't able to see them. This should be left blank (not omitted) for decorative images that a blind user can just ignore.

> Images with no semantic meaning—such as those which are solely decorative—or of limited informational value, should have their `alt` attributes set to the empty string (`""`).
>
> — https://developer.mozilla.org/docs/Web/API/HTMLImageElement/alt

### Consistent Padding (Commit 3)

On production, I can see you tried to have equal padding on the left and right side of the screen across breakpoints. However, between 768px and 823px, this padding disappears. In the screenshots you'll find on prod the padding on the left and right is jumping more than it should, but in the PR it remains consistent across breakpoints.

This aligns the left and right padding on the page to be consistent across breakpoints.

### Before

https://github.com/user-attachments/assets/a1278639-53bd-4397-ac57-62b7bd3d64c1

### After

https://github.com/user-attachments/assets/ae080c88-da49-4d51-93aa-9b777ac8035f

## Related

* Fixes https://github.com/checkstyle/checkstyle/issues/16825


romani:
Plt put issue number in commit to make clear linkage what problem is going to be fixed .
Other two commits can use prefix `Pull #16839: ....`.

romani:
@Zopsss , @smita1078 , please review update.

SethFalco:
@romani Updated all commit messages! 👍🏽

romani:
Website works well in mobile mode and desktop mode on mobile.
Good job, thanks a lot!

Zopsss (approved):
LGTM! Thanks a lot for the detailed explanation in PR description. Btw you got a nice monitor :D

PR #16840, opened 2025-04-12
Issue #14019: Covered pitest survivals for SuppressFilterElement

Atharv3221:
Issue: #14019

**Killing mutation with removing redundant code**
**Changes** 
```

    /**
     * Constructs a {@code SuppressFilterElement} using regular expressions
     * as {@code String}s. These are internally compiled into {@code Pattern}
     * objects and passed to the main constructor.
     *
     * @param files   regular expression for names of filtered files.
     * @param checks  regular expression for filtered check classes.
     * @param message regular expression for messages.
     * @param modId   the id
     * @param lines   lines CSV values and ranges for line number filtering.
     * @param columns columns CSV values and ranges for column number filtering.
     */
    public SuppressFilterElement(String files, String checks, String message,
                                 String modId, String lines, String columns) {
        this(toPattern(files), toPattern(checks), toPattern(message),
                modId, lines, columns);
    }

    /**
     * Converts a string into a compiled {@code Pattern}, or returns {@code null}
     * if the input is {@code null}.
     *
     * @param regex the regular expression as a string, may be {@code null}.
     * @return the compiled {@code Pattern}, or {@code null} if input is {@code null}.
     */
    private static Pattern toPattern(String regex) {
        final Pattern result;
        if (regex != null) {
            result = Pattern.compile(regex);
        }
        else {
            result = null;
        }
        return result;
    }
```

### DIFF
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -4820,65 +4820,10 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java</fileName>
     <specifier>assignment</specifier>
     <message>incompatible types in assignment.</message>
-    <lineContent>columnFilter = null;</lineContent>
-    <details>
-      found   : null (NullType)
-      required: @Initialized @NonNull CsvFilterElement
-    </details>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java</fileName>
-    <specifier>assignment</specifier>
-    <message>incompatible types in assignment.</message>
     <lineContent>lineFilter = null;</lineContent>
     <details>
       found   : null (NullType)
       required: @Initialized @NonNull CsvFilterElement
-    </details>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java</fileName>
-    <specifier>assignment</specifier>
-    <message>incompatible types in assignment.</message>
-    <lineContent>lineFilter = null;</lineContent>
-    <details>
-      found   : null (NullType)
-      required: @Initialized @NonNull CsvFilterElement
-    </details>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java</fileName>
-    <specifier>assignment</specifier>
-    <message>incompatible types in assignment.</message>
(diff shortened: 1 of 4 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElementTest.java
```diff
@@ -294,4 +296,45 @@ public void testEqualsAndHashCode() {
...
+        final Pattern messagePattern = Pattern.compile("TestMessage");
+
+        final SuppressFilterElement filterBased =
+                new SuppressFilterElement(filePattern, checkPattern, messagePattern,
+                        "module1", "1-10", "3,4");
+
+        final SuppressFilterElement filter2 =
+                new SuppressFilterElement(filePattern, checkPattern, messagePattern,
+                        "module1", "1-10", "3,4");
+        assertWithMessage("filter, filter2")
```
romani, line 313: Ok, good that you know what Filter elements are required to kill mutation.
But we do not allow pure unit testing anymore, sorry that there are bunch of bad examples of such testing.

Please convert such tests to have Input files with config for check+Filter to kill this mutation

romani (changes requested):
Itey

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElementTest.java
Atharv3221: Ok I know how to use verifyWithInlineConfigParser() with Example  file and expected violations need to learn about how will config work with it

Atharv3221: So, I need to use one suppression.xml (with lines and columns) and one Example.java and by using verify() i need to check if that violation is suppressed, if mutation sets  lines or columns to default (null) suppression will fail and mutation will be killed.

romani: Please try to use this method https://github.com/checkstyle/checkstyle/blob/76eca5ba0d2443bc6b158981dd398cef8838bc45/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionXpathSingleFilterTest.java#L64

Atharv3221: @romani, I tried this it doesn't kill the mutation, mutation is, the lines or columns in SuppressFilterElement are set to null, so we need something that can supress violation using SuppressFilterElement constructor, In
        verifyFilterWithInlineConfigParser(
            getPath("InputSuppressionXpathSingleFilterNonMatchingTokenType.java"), expected,
            removeSuppressed(expected, suppressed));
    }
Its just removes the suppressed violation from the expected, in reality violation is never getting suppressed, we want to reject an audit event using Filter, so when mutation occurs it fail to reject and the mutation will be killed

Atharv3221:
@romani some checks are failing wrongly, I tried few times it gave me a vaild check failure + 1 incorrect check failure for 6 times

romani (approved):
this actually very nice!

PR #16749, opened 2025-04-03
Pull #16749: `checker` reduce `CyclomaticComplexity`

Pankraz76:
# PR #16749: Reduce Cyclomatic Complexity in `checker`

## Prerequisites
- Requires [#16728](https://github.com/checkstyle/checkstyle/pull/16728)

## Core Principles
**SOC** (Separation of Concerns)  
**SRP** (Single Responsibility Principle)  
**Feature Envy** avoidance  

> *"Functions should do one thing, do it well, and do it only."*  
> ― [Clean Coders Episode 3](https://cleancoders.com/episode/clean-code-episode-3)

## Benefits of Field Reduction
**Reducing unnecessary fields achieves:**  
✔️ Noise reduction  
✔️ Boilerplate elimination  
✔️ Lower `coupling`  

**While simultaneously:**  
✔️ Increase `cohesion`  
✔️ Overhead minimization  
✔️ Business logic focus  
✔️ Field unification  

## Architectural Advantages
**Low Coupling + High Cohesion =**  
🔧 Better maintainability  
📈 Improved scalability  

## Related Work
🔗 Resolved in [#16488](https://github.com/checkstyle/checkstyle/pull/16488)  
(Fixes [#16421](https://github.com/checkstyle/checkstyle/issues/16421))  

**Rule**: *"Avoid declaring local variables only to immediately return/throw them"*  
([RSPEC-S1488](https://rules.sonarsource.com/java/RSPEC-1488))  

**Impact**:  
✅ Cleaner code standards  
✅ Functional programming style  
✅ Reduced overhead  

## References
📖 [Coupling/Cohesion Explained - ClarityHub](https://medium.com/clarityhub/low-coupling-high-cohesion-3610e35ac4a6)  
💬 [SO Discussion on Coupling & Cohesion](https://stackoverflow.com/questions/14000762/what-does-low-in-coupling-and-high-in-cohesion-mean)

PR #16843, opened 2025-04-12
Pull #16843: `checker` improve naming `toDir` -> `target` 

Pankraz76:
Pull #16843: `checker` fix naming `toDir` -> `target` 

reference: #16749

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
@@ -613,8 +613,10 @@ public class CheckstyleAntTask extends Task {
         /** The formatter type. */
         private FormatterType type;
         /** The file to output to. */
-        private File toFile;
-        /** Whether or not to write to the named file. */
+        private File target;
+        /**
+         * Whether or not to write to the named file.
+         */
         private boolean useFile = true;
 
         /**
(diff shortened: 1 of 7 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
```diff
@@ -627,12 +627,12 @@ public void setType(FormatterType type) {
         }
 
         /**
-         * Set the file to output to.
+         * Set the file target.
          *
-         * @param destination destination the file to output to
+         * @param target for file output
          */
-        public void setTofile(File destination) {
```
pankratz76, line 634: this looks kind of a typo imho
`setTofile` ->
`setToFile` (CamelCase)

pankratz76: I know this term commonly from [Path.toFile](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#toPath--) as it sounds like a converter method and not like a field or property in first place.

pankratz76:
[INFO] BUILD SUCCESS

but ci:

* What went wrong:
Execution failed for task ':eclipsePlugin-junit:checkstyleTest'.
> A failure occurred while executing org.gradle.api.plugins.quality.internal.CheckstyleAction
   > An unexpected error occurred configuring and executing Checkstyle.
      > formatter doesn't support the "toFile" attribute

pankratz76:
is this name required or can it evolve?

nrmancuso (changes requested):
This is a breaking change, you will need to create an issue for this and we can discuss there. IMO, this change is not worth the overhead.

pankratz76:
> IMO, this change is not worth the overhead.

yes its trivial and low prio.

PR #16852, opened 2025-04-13
Issue #14019: Covered pitest survivals with tests for CheckstyleAntTask

Atharv3221:
Issue: #14019 

**Suppression removed**
  ```
<mutation unstable="false">
    <sourceFile>CheckstyleAntTask.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask</mutatedClass>
    <mutatedMethod>&lt;init&gt;</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
    <description>Removed assignment to member variable maxWarnings</description>
    <lineContent>private int maxWarnings = Integer.MAX_VALUE;</lineContent>
  </mutation>

```

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
@@ -21,6 +21,7 @@ package com.puppycrawl.tools.checkstyle.ant;
 
 import static com.google.common.truth.Truth.assertWithMessage;
 import static com.puppycrawl.tools.checkstyle.internal.utils.TestUtil.getExpectedThrowable;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 
 import java.io.File;
 import java.io.IOException;
@@ -976,6 +977,17 @@ public class CheckstyleAntTaskTest extends AbstractPathTestSupport {
         final Matcher matcher = Pattern.compile("(\\d+)").matcher(line);
         matcher.find();
         return Long.parseLong(matcher.group(1));
+    }
+
+    @Test
+    public void testMaxWarningDefault() throws IOException {
+        final CheckstyleAntTask antTask = getCheckstyleAntTask();
+        final File inputFile = new File(getPath(WARNING_INPUT));
+        final Location fileLocation = new Location("build.xml", 42, 10);
+
+        antTask.setFile(inputFile);
+        antTask.setLocation(fileLocation);
+        assertDoesNotThrow(antTask::execute, "BuildException is not expected");
     }
 
 }
--- a/config/pitest-suppressions/pitest-ant-suppressions.xml
+++ b/config/pitest-suppressions/pitest-ant-suppressions.xml
@@ -3,15 +3,6 @@
   <mutation unstable="false">
     <sourceFile>CheckstyleAntTask.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java [hunk 1]
```diff
@@ -978,4 +978,19 @@ private static long getNumberFromLine(String line) {
...
+        final File inputFile = new File(getPath(WARNING_INPUT));
+        final Location fileLocation = new Location("build.xml", 42, 10);
+
+        antTask.setFile(inputFile);
+        antTask.setLocation(fileLocation);
+        antTask.execute();
+
+        assertWithMessage("File location should be valid")
+                .that(antTask.getLocation())
+                .isNotNull();
```
romani, line 993: ok , it works, but looks very unclear.

how check for location
helps with 
```
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
    <description>Removed assignment to member variable maxWarnings</description>
    <lineContent>private int maxWarnings = Integer.MAX_VALUE;</lineContent>
```

can we put in assert more meaningful checking ?

romani (changes requested):
items:

Atharv3221:
I am not getting alternative 
[ERROR] [checkstyle] [ERROR] C:\projects\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\ant\CheckstyleAntTaskTest.java:989:9: try/catch either requires a fail at the end of the try clause for junit tests or to remove the try/catch completely. [RequireFailForTryCatchInJunit]

My tests don't throw any exception in normal conditions it only throws when it mutation occurs, it throws BuildException only when mutation occurs, so way of getExpectedThrowable failed as it doesnt throw anything in normal condition.

src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
```diff
@@ -978,4 +979,15 @@ private static long getNumberFromLine(String line) {
...
+    @Test
+    public void testMaxWarningDefault() throws IOException {
+        final CheckstyleAntTask antTask = getCheckstyleAntTask();
+        final File inputFile = new File(getPath(WARNING_INPUT));
+        final Location fileLocation = new Location("build.xml", 42, 10);
+
+        antTask.setFile(inputFile);
+        antTask.setLocation(fileLocation);
+        assertDoesNotThrow(antTask::execute, "BuildException is not expected");
+    }
```
romani, line 991: if you does not why it kill mutation, ok

but please rename testMaxWarningDefault to testDefaultTask

romani (changes requested):
items

src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
Atharv3221: @romani, I know how mutation is killed here, mutation was changing "int maxWarnings = 0" in line
`private int maxWarnings = Integer.MAX_VALUE;`
Other tests were using 
```
public void setMaxWarnings(int maxWarnings) {
        this.maxWarnings = maxWarnings;
    }
```
so default value of maxWarnings = Integer.Max_VALUE was never used, so I instead of using setMaxWarnings made the test to depend on default value, as a result Exception shouldn't be thrown even if there are warnings. If BuildException thrown than tests fail.

Hence I named the `testMaxWarningsDefault()` as it uses the default value and not rely on setMaxWarnings().

romani (approved):
ok, thanks a lot

PR #16854, opened 2025-04-13
WIP: Add editorconfig refomrat

Pankraz76:
3 work as well and can be reformatted. Somehow intelij can not apply reformat on these long lines.

image


Pankraz76:
why is this length required?

Issue #16857, opened 2025-04-13
api: add new 'int process(List<Path> files)' method to RootModule


romani:
detected at 
https://github.com/checkstyle/checkstyle/pull/16716#discussion_r2041194180

code:
https://github.com/checkstyle/checkstyle/blob/52225c3a46dd3a7e71965c64c7a647b9696beff1/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java#L345

api:
https://github.com/checkstyle/checkstyle/blob/52225c3a46dd3a7e71965c64c7a647b9696beff1/src/main/java/com/puppycrawl/tools/checkstyle/api/RootModule.java#L44

at https://github.com/checkstyle/checkstyle/issues/16155 there is suggestion to migrate to new API.
we cannot fully migrate as it is big breaking compatibility, but we can add new method to let users start migration when they care about this or have time.

proposal is to add new method:
`int process(List<Path> files) throws CheckstyleException;`

Pankraz76:
Should we create a migration script for this? Similar to major migrations like Spring's:
https://docs.openrewrite.org/recipes/java/spring/boot3/upgradespringboot_3_0

Since it's likely just a simple type replacement plus import statement change:

pre, post:
`int process(List<File> files)`
`int process(List<Path> files)`

This basic case would serve as:
1. **Good learning example** - Simple enough to understand the pattern
2. **Long-term benefit** - Immediately useful once implemented
3. **Standardized solution** - Reusable across the ecosystem

The Spring Boot 3.0 migration demonstrates how powerful these automated transformations are for large-scale upgrades.

Pankraz76:
> `int process(List<Path> files)`

It seems that using the same type again is not allowed, so we'll have to break the API one way or another, since overloading with same type isn't an option.
I suggest we go with the higher-level type—it improves the API as well.

<img width="802" alt="Image" src="https://github.com/user-attachments/assets/4f6bdd0c-a3a6-49b5-a2f7-c54c5e273e00" />

A higher-level interface is better, in my opinion. For example, why shouldn't we accept a `java.util.Set`? Using `java.util.Collection` might be a cleaner and more flexible choice overall.

<img width="1068" alt="image" src="https://github.com/user-attachments/assets/666df3fc-9605-4b4e-ada2-7cb94f465555" />
 

Breaking the API name, would allow to have same type, but I do not suggest this way:
[...]

PR #16861, opened 2025-04-14
Issue #16857 api: add new `int process(List<Path> files)` method to RootModule

Pankraz76:
Issue #16857 api: add new `int process(List<Path> files)` method to RootModule

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
@@ -26,6 +26,7 @@ import java.io.StringWriter;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
@@ -210,6 +211,15 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
 
     @Override
     public int process(List<File> files) throws CheckstyleException {
+        return process2(files);
+    }
+
+    @Override
+    public int processPath(List<Path> files) throws CheckstyleException {
+        return process2(files);
+    }
+
+    private int process2(List<File> files) throws CheckstyleException {
         if (cacheFile != null) {
             cacheFile.putExternalResources(getExternalResourceLocations());
         }
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
@@ -221,6 +221,29 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
             .collect(Collectors.toUnmodifiableList()));
     }
 
+    /**
+     * Processes a list of files through the checkstyle verification pipeline.
+     *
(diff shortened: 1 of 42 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/api/RootModule.java
```diff
@@ -39,10 +41,23 @@ public interface RootModule extends Configurable {
...
+     * Processes a set of files.
+     * Once this is done, it is highly recommended to call for
+     * the destroy method to close and remove the listeners.
+     *
+     * @param files the list of files to be audited.
+     * @return the total number of audit events with error severity found
+     * @throws CheckstyleException if error condition within Checkstyle occurs
+     * @see #destroy()
+     */
+    int process(Collection<Path> files) throws CheckstyleException;
```
Pankraz76, line 59:
```suggestion
    int process(List<Path> files) throws CheckstyleException;
```

Using the same type again is not allowed.

A higher-level interface is better, in my opinion. Why shouldn't we accept a `java.util.Set`, for example? Using `java.util.Collection` might be a cleaner and more flexible API anyway.

image

Pankraz76: breaking API name, would have same type but I do not suggest this idea:

image

Pankraz76: how should we deal with this @romani @nrmancuso ?

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -281,13 +313,13 @@ private void fireAuditFinished() {
...
-    private void processFiles(List<File> files) throws CheckstyleException {
-        for (final File file : files) {
+    private void processFiles(List<Path> files) throws CheckstyleException {
+        for (final Path file : files) {
             String fileName = null;
-            final String filePath = file.getPath();
+            final String filePath = file.toAbsolutePath().toString();
             try {
-                fileName = file.getAbsolutePath();
-                final long timestamp = file.lastModified();
```
Pankraz76, line 290: **Items:**  

1. **Is there an existing simple recipe for this?**  
   When checking [OpenRewrite's Java NIO migration docs](https://docs.openrewrite.org/recipes/java/migrate/nio/file), I only found `Path`-to-`Path` migrations, not `File`-to-`Path` adaptations.  

2. **Automation potential**  
   As shown in this example, this seems like repetitive work that could be automated.  
   @timtebeek - Could this be a candidate for a new recipe?

timtebeek: I mean it's possible; it's a little more involved because we have to evaluate the flow leading into this if all usages can be replaced, and whether that'd constitute a clear improvement. We do similar work when it comes to replacing JodaTime with the JDK classes, but there it's a little more clear cut that it's an improvement at least.

Pankraz76: Thanks, will create an issue for this for further discussion.

Pankraz76: https://github.com/openrewrite/rewrite-static-analysis/issues/509

PR #16874, opened 2025-04-15
Pull #16874: api: add new `int process(List<Path> files)` method to RootModule - `prepare checker: reduce complexity`

Pankraz76:
Pull #16874: api: add new `int process(List<Path> files)` method to RootModule - `prepare checker: reduce complexity`
Issue #16857: api: add new `int process(List<Path> files)` method to RootModule


Pankraz76:
await api

Issue #16882, opened 2025-04-16
Remove usage of AtomicInteger from NPathComplexityCheck


romani:
Remove usage of java.util.concurrent.atomic.AtomicInteger
From
https://github.com/checkstyle/checkstyle/blob/f88a60db7951f57daee9ebd8c6395c66699507ad/src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/NPathComplexityCheck.java#L551

Suppression of fundbug might need to be updated 

Detected by https://github.com/spotbugs/spotbugs/issues/3362#issuecomment-2809975804

Atharv3221:
@romani this issue can be closed 
Ref: https://github.com/checkstyle/checkstyle/pull/16905

PR #16883, opened 2025-04-17
Issue #16882: update atomic variable to non-atomic int

muratcanyeldan:
Issue #16882

Change AtomicInteger to int according to given issue.

Here is the "mvn clean verify" results below:

Screenshot 2025-04-17 at 10 06 44


romani (approved):
Thanks a lot

Issue #16880, opened 2025-04-16
Update IndentationCheckTest to use verifyWarns


romani:
Detected at https://github.com/checkstyle/checkstyle/pull/16879#discussion_r2047863368

Update all 4 usage of `verify(` method 
To use `verifyWarns(checkConfig...`

romani:
Not all updated yet

Praveen7294:
> Not all updated yet

yes two methods are remaining: one is verifying, and the other has an issue.

Praveen7294:
Apologies for the oversight, future PRs will be marked as partial progress until the final one resolves the issue fully.

romani:
final fix is merged.

PR #16890, opened 2025-04-17
Issue #16880: Replaced verify() with verifyWarns() in IndentationCheckTest

nehamanoj1105:
Replaced verify() with verifyWarns() in IndentationCheckTest


Praveen7294:
@nehamanoj1105 

We don't need to change the implementation of the `verify()` method. We just need to update the test cases to use the `verifyWarns()` method instead of `verify()`.

please make changes in single commit.

#### Important:-
Do not work on the `master` branch. Always create a new branch for each issue and work on that branch. This will help avoid any conflicts.

PR #16901, opened 2025-04-19
Issue #16882: Removed usage of AtomicInteger from RecordComponentNumberCheck

Atharv3221:
Issue: #16882

**Suppression removed**

 ```
 <Match>
    <!-- we are single thread tool for now -->
  </Match>
```
**Method updated**

```
    private static int countComponents(DetailAST recordComponents) {
        return recordComponents.getChildCount(TokenTypes.RECORD_COMPONENT_DEF);
    }

```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/RecordComponentNumberCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/RecordComponentNumberCheck.java
@@ -20,7 +20,6 @@
 package com.puppycrawl.tools.checkstyle.checks.sizes;
 
 import java.util.Arrays;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
@@ -28,7 +27,6 @@ import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
 import com.puppycrawl.tools.checkstyle.checks.naming.AccessModifierOption;
 import com.puppycrawl.tools.checkstyle.utils.CheckUtil;
-import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
 
 /**
  * <div>
@@ -156,11 +154,7 @@ public class RecordComponentNumberCheck extends AbstractCheck {
      * @return the number of record components in this record definition
      */
     private static int countComponents(DetailAST recordComponents) {
-        final AtomicInteger count = new AtomicInteger(0);
-        TokenUtil.forEachChild(recordComponents,
-            TokenTypes.RECORD_COMPONENT_DEF,
-            node -> count.getAndIncrement());
-        return count.get();
+        return recordComponents.getChildCount(TokenTypes.RECORD_COMPONENT_DEF);
     }
 
     /**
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/RecordComponentNumberCheck.java [hunk 1]
```diff
@@ -156,11 +154,15 @@ public void visitToken(DetailAST ast) {
...
-            node -> count.getAndIncrement());
-        return count.get();
+        int count = 0;
+        DetailAST child = recordComponents.getFirstChild();
+        while (child != null) {
+            if (child.getType() == TokenTypes.RECORD_COMPONENT_DEF) {
+                count++;
+            }
+            child = child.getNextSibling();
+        }
```
Pankraz76, line 164:
```suggestion
        for (DetailAST child = recordComponents.getFirstChild();
             child != null;
             child = child.getNextSibling()) {
            if (child.getType() == TokenTypes.RECORD_COMPONENT_DEF) {
                count++;
            }
        }
```
this is the simpler fit for what we are trying to do and what actually happens.

Pankraz76 (approved):
nicely done.

consider using a simpler fit for what we are trying to do.

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/RecordComponentNumberCheck.java
Atharv3221: Okay

src/main/java/com/puppycrawl/tools/checkstyle/checks/sizes/RecordComponentNumberCheck.java
```diff
@@ -156,11 +154,14 @@ public void visitToken(DetailAST ast) {
...
-        final AtomicInteger count = new AtomicInteger(0);
-        TokenUtil.forEachChild(recordComponents,
-            TokenTypes.RECORD_COMPONENT_DEF,
-            node -> count.getAndIncrement());
-        return count.get();
+        int count = 0;
+        for (DetailAST child = recordComponents.getFirstChild();
+             child != null; child = child.getNextSibling()) {
+            if (child.getType() == TokenTypes.RECORD_COMPONENT_DEF) {
+                count++;
```
romani, line 161: please reuse this
https://github.com/checkstyle/checkstyle/blob/614c3f452b326776b6785b3fcddc32b0091bb24e/src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java#L194-L202

romani (changes requested):
items:

romani (approved):
thanks a lot.

to resolve this problem at root level
we should forbid `import java.util.concurrent.atomic.AtomicInteger`
but Checkstyle Import order, so it will not happen again until we come to Multithreading mode.

romani:
@Atharv3221 , please resolve conflict.

Atharv3221:
@romani, we can't disallow import of `java.util.concurrent.atomic.AtomicInteger` yet, it is still used in
`api\SeverityLevelCounter.java and checks\UniquePropertiesCheck`

romani:
>we can't disallow import of

ok, we will do it after rafactoring of this classes.

Pankraz76:
https://github.com/checkstyle/checkstyle/issues/16903

can you please assign me, I would like to migrate one by another, if not others want to do as well.

romani (approved):
Thanks a lot

PR #16902, opened 2025-04-19
Deprecate fix process path adding new api only new

Pankraz76:
Pull #16902: api: add new `int process(List<Path> files)` method to RootModule - `prepare deprecation`
Issue #16857: api: add new `int process(List<Path> files)` method to RootModule


Pankraz76:
waiting for https://github.com/checkstyle/checkstyle/pull/16904

PR #16914, opened 2025-04-20
Issue #14084: Remove warning from nullness profile for CheckstyleAntTask

Rohanraj123:
part of #14084

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
@@ -263,7 +263,9 @@ public class CheckstyleAntTask extends Task {
         final long startTime = System.currentTimeMillis();
 
         try {
-            final String version = CheckstyleAntTask.class.getPackage().getImplementationVersion();
+            final String version = Objects.toString(
+                    CheckstyleAntTask.class.getPackage().getImplementationVersion(),
+                    "");
 
             log("checkstyle version " + version, Project.MSG_VERBOSE);
 
@@ -334,7 +336,7 @@ public class CheckstyleAntTask extends Task {
             Project.MSG_VERBOSE);
 
         log("Running Checkstyle "
-                + Objects.toString(checkstyleVersion, "")
+                + checkstyleVersion
                 + " on " + files.size()
                 + " files", Project.MSG_INFO);
         log("Using configuration " + config, Project.MSG_VERBOSE);
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -1062,17 +1062,6 @@
   <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java</fileName>
     <specifier>argument</specifier>
-    <message>incompatible argument for parameter checkstyleVersion of CheckstyleAntTask.realExecute.</message>
-    <lineContent>realExecute(version);</lineContent>
-    <details>
-      found   : @Initialized @Nullable String
-      required: @Initialized @NonNull String
-    </details>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java</fileName>
-    <specifier>argument</specifier>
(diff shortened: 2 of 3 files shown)


Rohanraj123:
@romani PTAL

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
```diff
@@ -290,9 +292,9 @@ public void execute() {
     /**
      * Helper implementation to perform execution.
      *
-     * @param checkstyleVersion Checkstyle compile version.
+     * @param checkstyleVersion Checkstyle compile version, may be null.
```
Pankraz76, line 295: What about this idea? 

Maybe the link is a little too much.

image

image

```suggestion
     * @param checkstyleVersion {@link javax.annotation.Nullable} Checkstyle compile version.
```

Pankraz76 (approved):
thx

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
```diff
@@ -321,12 +323,12 @@ private void realExecute(String checkstyleVersion) {
      *
      * @param rootModule Root module to process files
      * @param warningCounter Root Module's counter of warnings
-     * @param checkstyleVersion Checkstyle compile version
+     * @param checkstyleVersion Checkstyle compile version, may be null.
```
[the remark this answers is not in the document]
Pankraz76: I'm not even sure if we need to mention this in the documentation. Everything can potentially be null. Only a non-nullable property would be an exception and warrant special attention, which should probably be mentioned.

Rohanraj123:
ping @romani

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
Rohanraj123: Hi @Pankraz76 , Thanks for the idea, we can discuss on that, Once this approach is finalized. Because these fix can be controversial. So at first we need to discuss whether annotation worth using or not.

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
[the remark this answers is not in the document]
romani: lets not do this.
as code does assumption on how to interpretate null `+ Objects.toString(checkstyleVersion, "")`

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
romani: please revert comment update, better to not pollute comment wit htechnical details that might change, and comment wil be forgotten to be updated.

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
```diff
@@ -290,9 +292,9 @@ public void execute() {
     /**
      * Helper implementation to perform execution.
      *
-     * @param checkstyleVersion Checkstyle compile version.
+     * @param checkstyleVersion Checkstyle compile version, may be null.
      */
-    private void realExecute(String checkstyleVersion) {
+    private void realExecute(@Nullable String checkstyleVersion) {
```
romani, line 297: as all are private,  lets move conversion of null `checkstyleVersion` to empty string in some top level.
and we will not need null annotations at all.

romani (changes requested):
item:

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
Rohanraj123: yes make sense ! Thanks.

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
```diff
@@ -326,15 +328,15 @@ private void realExecute(String checkstyleVersion) {
      *     or if the build failed due to violations.
      */
     private void processFiles(RootModule rootModule, final SeverityLevelCounter warningCounter,
-            final String checkstyleVersion) {
+             final String checkstyleVersion) {
```
romani, line 331: Revert

romani (changes requested):
Please apply suggested patch from pitest execution

romani (approved):
thanks a lot !!!

Issue #14879, opened 2024-05-15
Configuration website does not allow linking to Examples


rnveach:
https://checkstyle.org/config.html#TreeWalker_Examples
https://checkstyle.org/config.html#Checker_Examples

I can't hover an example and create a link to it like I can with the check pages.

Example: https://checkstyle.org/checks/coding/equalsavoidnull.html#Example1-config

MANISH-K-07:
@rnveach , are you expecting anchors like with the check examples or embedded links?

rnveach:
I am not sure what difference you are asking, but the first post has an example with a picture and a link.

PR #16918, opened 2025-04-20
Issue #14879 : Configuration website does not allow linking to Examples

smita1078:
#14879 
 
 Anchors are generated when we have an id attribute , therefore for such files [https://checkstyle.org/config.html#](https://checkstyle.org/config.html#TreeWalker_Examples) id attribute is added so that anchor.js generates linking href.
 
 Output:
 
image
Now We can hover and create a link to it!

### DIFF
--- a/src/site/xdoc/config.xml
+++ b/src/site/xdoc/config.xml
@@ -411,7 +411,7 @@
       </subsection>
 
       <subsection name="Examples" id="Checker_Examples">
-        <p>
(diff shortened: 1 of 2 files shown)


smita1078:
Github, generate site

romani (approved):
Thanks a lot.
In issue different pages are mentioned to update, but conceptually any page can be updated like this

PR #16956, opened 2025-04-26
Issue #16155: Used nio api in CheckUtil

Atharv3221:
Issue: #16155 

Removed usage of File.io in CheckUtil, CheckUtilTest, MissingJavadocMethodCheckTest

 ```
   public static boolean isPackageInfo(String filePath) {
        final Path path = Path.of(filePath);
        final Path filename = path.getFileName();
        return filename != null && "package-info.java".equals(filename.toString());
    }
```
Explanation:
As .toString() can cause NullPointerException, so `filename != null` was needed, as no existing test tested for `filename != null`
added new test to trigger if `filename == null` killing the possible mutations.
```
    @Test
    public void testPackageInfo() {
        final boolean result = CheckUtil.isPackageInfo("/");

        assertWithMessage("Expected isPackageInfo() to return false for ('/')")
                .that(result)
                .isFalse();
    }
```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java
@@ -19,7 +19,7 @@
 
 package com.puppycrawl.tools.checkstyle.utils;
 
-import java.io.File;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/main/java/com/puppycrawl/tools/checkstyle/utils/CheckUtil.java [hunk 1]
```diff
@@ -568,6 +568,8 @@ public static String getShortNameOfAnonInnerClass(DetailAST literalNewAst) {
      * @return true if the package file.
      */
     public static boolean isPackageInfo(String filePath) {
-        return "package-info.java".equals(new File(filePath).getName());
+        final Path path = Path.of(filePath);
+        final Path filename = path.getFileName();
```
romani, line 572: `final Path filename = Path.of(filePath).getFileName();`

romani (changes requested):
thanks a lot, the only minor:

romani (approved):
Ok to merge, thanks a lot

Issue #11249, opened 2022-01-28
Enforce formatting of `// violation` and `// ok` comments


nmancus1:
Noticed in many PR's, it is difficult to make sure that all BDD and documentation related comments (`// ok` and `// violation`) adhere to consistent format. We can save time in PR reviews by adding `RegexpSingleline` modules to associated configs.

Usual convention:

```
➜  checkstyle git:(master) grep -R "\/\/ ok" | wc -l
4049
➜  checkstyle git:(master) grep -R "\/\/ violation" | wc -l
74064

```

We should forbid the following forms of comments:

```
➜  checkstyle git:(master) grep -iR "\/\/ok" | wc -l 
341
➜  checkstyle git:(master) grep -iR "\/\/violation" | wc -l
107
➜  checkstyle git:(master) grep -R "\/\/ Violation" | wc -l
61
➜  checkstyle git:(master) grep -R "\/\/ VIOLATION" | wc -l
1
➜  checkstyle git:(master) grep -R "\/\/ OK" | wc -l
1563
➜  checkstyle git:(master) grep -R "\/\/ bad" | wc -l
6
➜  checkstyle git:(master) grep -R "\/\/bad" | wc -l
8

```

While check input files are not a big deal (but should also be fixed and enforced), many of these are in documentation, where we should make sure to always be consistent.

rnveach:
Not sure, but I would think BDD should enforce consistency of violation text.

Vyom-Yadav:
This is quite easy to achieve with the parser itself, just add a pattern for comments not allowed and throw an exception if such comment is encountered.

romani:
lets inforce it by Parser, if not possible by regexp in checkstyle config

nmancus1:
>  just add a pattern for comments not allowed and throw an exception if such comment is encountered.

Could we refine existing regexp, and throw an exception if comment doesn't match?

> lets inforce it by Parser, if not possible by regexp in checkstyle config

Let's do both, so that we can check documentation, too.

Vyom-Yadav:
> Could we refine existing regexp, and throw an exception if comment doesn't match?

We don't have a pattern for `// ok`, technically it is possible to modify every regex and check for `// ok` with capture groups but that won't be easy for someone in the future to understand. I was thinking to add a regex like - `//\s{1}ok((\s|,).*)?`, this would match-
```
// ok, this is not supported yet
// ok this is not supported yet
```
but won't match 
```
// okay, this ......
[...]

PR #16967, opened 2025-04-28
Commentcheck

smita1078:
#11249 

**✅ Allowed (will pass the check)**
```
// ok
// ok explanation
// ok, explanation
someCode(); // ok
foo();      // ok, because we know it’s safe
```
```
// violation
// violation explanation
// violation, missing semicolon
bar();      // violation, need to handle this
```

**🚫 Flagged (will fail the check)**

```
//OK                          ← uppercase
// OK                         ← uppercase
//ok                          ← no space before word
//violation                   ← no space before word
//okapi                       ← glued into larger word
// violationBad               ← glued word
// Violation                   ← uppercase
foo(); // Violation, reason  ← uppercase V
someCode(); // OK, reason     ← uppercase OK
//VIOLATION                   ← uppercase Letters
```

InlineConfigParser: added `ALLOWED_OK_VIOLATION_PATTERN + ANY_OK_VIOLATION_PATTERN` and a top‐of‐`setViolations(...)` guard to reject any comment that contains the words ok or violation but doesn’t exactly match one of these two allowed forms (all lowercase, one space, optional comma/explanation).

Checkstyle config (config/checkstyle_checks.xml)
Added a new `<module name="RegexpSingleline" id="okViolationCommentFormat">` under `<module name="Checker">`
Regex anchored to // ok or // violation (lowercase, one space) with a negative look‐ahead forbidding all other forms.

Testing on local Output:
Introduced few wrong comment patterns:
okays
vilation

### DIFF
--- a/src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
+++ b/src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
@@ -149,7 +149,8 @@ public class Example2 {
 &lt;module name="Checker"&gt;
   &lt;module name="TreeWalker"&gt;
     &lt;module name="SuppressWithNearbyCommentFilter"&gt;
-      &lt;property name="commentFormat" value="OK to catch (\w+) here"/&gt;
+      &lt;property name="commentFormat"
+                value="ok,?\s*(?:allowed\s+)?to catch (\w+) here"/&gt;
       &lt;property name="checkFormat" value="IllegalCatchCheck"/&gt;
       &lt;property name="messageFormat" value="$1"/&gt;
       &lt;property name="influenceFormat" value="-1"/&gt;
@@ -166,7 +167,7 @@ public class Example3 {
       // blah blah blah
     }
     catch(RuntimeException re) {
-      // OK to catch RuntimeException here
+      // ok, allowed to catch RuntimeException here
     }
   }
 }
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/filters/suppresswithnearbycommentfilter/Example3.java
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/filters/suppresswithnearbycommentfilter/Example3.java
@@ -2,7 +2,8 @@
 <module name="Checker">
   <module name="TreeWalker">
     <module name="SuppressWithNearbyCommentFilter">
-      <property name="commentFormat" value="OK to catch (\w+) here"/>
+      <property name="commentFormat"
+                value="ok,?\s*(?:allowed\s+)?to catch (\w+) here"/>
       <property name="checkFormat" value="IllegalCatchCheck"/>
       <property name="messageFormat" value="$1"/>
       <property name="influenceFormat" value="-1"/>
@@ -19,7 +20,7 @@ public class Example3 {
       // blah blah blah
     }
     catch(RuntimeException re) {
-      // OK to catch RuntimeException here
+      // ok, allowed to catch RuntimeException here
     }
   }
 }
--- a/src/site/xdoc/checks/coding/variabledeclarationusagedistance.xml
+++ b/src/site/xdoc/checks/coding/variabledeclarationusagedistance.xml
@@ -76,7 +76,7 @@ public class Example1 {
   public void foo1() {
     // violation below, 'variable 'num' declaration and its first usage is 4.'
     int num;
-    final double PI;   // OK, final variables not checked
+    final double PI;   // ok, final variables not checked
     System.out.println("Statement 1");
     System.out.println("Statement 2");
     System.out.println("Statement 3");
@@ -85,9 +85,9 @@ public class Example1 {
   }
 
   public void foo2() {
-    int a;          // OK, used in different scope
-    int b;          // OK, used in different scope
-    int count = 0;  // OK, used in different scope
+    int a;          // ok, used in different scope
+    int b;          // ok, used in different scope
+    int count = 0;  // ok, used in different scope
 
     {
       System.out.println("Inside inner scope");
@@ -118,7 +118,7 @@ public class Example1 {
 public class Example2 {
 
   public void case1(long timeNow, int hh, int min) {
-    int minutes = min + 5; // Ok, No violation reported
+    int minutes = min + 5; // ok, No violation reported
     Calendar cal = Calendar.getInstance();
     cal.setTimeInMillis(timeNow);
     cal.set(Calendar.SECOND, 0);
@@ -166,8 +166,8 @@ public class Example2 {
 public class Example3 {
 
   public void foo1() {
-    int num;        // OK, distance = 4
-    final double PI;   // OK, final variables not checked
+    int num;        // ok, distance = 4
+    final double PI;   // ok, final variables not checked
     System.out.println("Statement 1");
     System.out.println("Statement 2");
     System.out.println("Statement 3");
@@ -176,9 +176,9 @@ public class Example3 {
   }
 
   public void foo2() {
-    int a;          // OK, used in different scope
-    int b;          // OK, used in different scope
-    int count = 0;  // OK, used in different scope
+    int a;          // ok, used in different scope
+    int b;          // ok, used in different scope
+    int count = 0;  // ok, used in different scope
 
     {
       System.out.println("Inside inner scope");
@@ -209,8 +209,8 @@ public class Example3 {
 public class Example4 {
 
   public void foo1() {
-    int num;        // OK, variable ignored
-    final double PI;   // OK, final variables not checked
+    int num;        // ok, variable ignored
+    final double PI;   // ok, final variables not checked
     System.out.println("Statement 1");
     System.out.println("Statement 2");
     System.out.println("Statement 3");
@@ -219,9 +219,9 @@ public class Example4 {
   }
 
   public void foo2() {
-    int a;          // OK, used in different scope
-    int b;          // OK, used in different scope
-    int count = 0;  // OK, used in different scope
+    int a;          // ok, used in different scope
+    int b;          // ok, used in different scope
+    int count = 0;  // ok, used in different scope
 
     {
       System.out.println("Inside inner scope");
@@ -251,7 +251,7 @@ public class Example5 {
   public void foo1() {
     // violation below, 'variable 'num' declaration and its first usage is 4.'
     int num;
-    final double PI;   // OK, final variables not checked
+    final double PI;   // ok, final variables not checked
     System.out.println("Statement 1");
     System.out.println("Statement 2");
     System.out.println("Statement 3");
@@ -260,8 +260,8 @@ public class Example5 {
   }
 
   public void foo2() {
-    int a;          // OK, distance = 2
-    int b;          // OK, distance = 3
+    int a;          // ok, distance = 2
+    int b;          // ok, distance = 3
     // violation below, 'variable 'count' declaration and its first usage is 4.'
     int count = 0;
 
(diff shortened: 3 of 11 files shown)


config/checkstyle-checks.xml
```diff
@@ -200,6 +200,14 @@
...
               value="@since tag must follow the X.Y.Z version format after version 10.5.0."/>
   </module>
+  <module name="RegexpSingleline">
+    <property name="id" value="okViolationCommentFormat"/>
+    <property name="format"
+              value="^\s*//(?=\s*(?:ok|violation)\b)(?!\s+(?:ok|violation)\b(?:[ ,].*)?$).*$"/>
+    <property name="message"
+              value="Comments must be '// ok' or '// violation' (lowercase) with optional comma/explanation."/>
+  </module>
+
```
Pankraz76, line 210:
```suggestion
```
keep block:
image

src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/commentsindentation/InputCommentsIndentationCommentIsAtTheEndOfBlockFour.java
```diff
@@ -19,7 +19,7 @@ public void foo45() {
         int ar = 5;
         // comment
         ar = 6;
-         //violation '.* incorrect .* level 9, expected is 8, .* same .* as line 21.'
+         // violation '.* incorrect .* level 9, expected is 8, .* same .* as line 21.'
```
Pankraz76, line 22: can we supplement this change therefor reducing this quite immense PR?

Pankraz76 (approved):
thanks for the dedication put in.

src/site/xdoc/checks/coding/hiddenfield.xml
```diff
@@ -385,7 +385,7 @@ class Example7 {
     return null;
   }
   abstract class Inner {
-    abstract int method(String field); // OK, because ignoreAbstractMethods is true
```
Pankraz76, line 388: extract if possible, as reproducible by `s&r`

src/site/xdoc/checks/coding/illegaltokentext.xml
```diff
@@ -97,7 +97,7 @@
 public class Example1 {
   public void myTest() {
     String test  = "a href"; // violation
-    String test2 = "A href"; // OK, case is sensitive
```
Pankraz76, line 100: extract if possible, as reproducible by `s&r`

config/checkstyle-checks.xml
smita1078: Done!

src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/commentsindentation/InputCommentsIndentationCommentIsAtTheEndOfBlockFour.java
smita1078: Done!

src/site/xdoc/checks/coding/hiddenfield.xml
smita1078: Did by search and replace using grep
`grep -rl "// OK," src/xdocs-examples/resources | xargs sed -i "s#// OK,#// ok,#g"`

src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/commentsindentation/InputCommentsIndentationCommentIsAtTheEndOfBlockFour.java
smita1078: But I have rebased the other branch on the branch with main changes, as ci was red because of the errors due to formatting violation as I moved those changes to other branch

src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
```diff
@@ -166,7 +166,7 @@ public class Example3 {
       // blah blah blah
     }
     catch(RuntimeException re) {
-      // OK to catch RuntimeException here
+      // ok, to catch RuntimeException here
```
nrmancuso, line 169:
```suggestion
      // ok, allowed to catch RuntimeException here
```

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/filters/suppresswithnearbycommentfilter/Example3.java
```diff
@@ -19,7 +19,7 @@ public void doStuff() {
       // blah blah blah
     }
     catch(RuntimeException re) {
-      // OK to catch RuntimeException here
+      // ok, to catch RuntimeException here
```
nrmancuso, line 22: Same as above

nrmancuso (changes requested):
Good stuff, minor items:

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/filters/suppresswithnearbycommentfilter/Example3.java
smita1078: Done

nrmancuso (approved):
Thanks!

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/filters/suppresswithnearbycommentfilter/Example3.java
```diff
@@ -2,7 +2,8 @@
 <module name="Checker">
   <module name="TreeWalker">
     <module name="SuppressWithNearbyCommentFilter">
-      <property name="commentFormat" value="OK to catch (\w+) here"/>
+      <property name="commentFormat"
+          value="ok,?\s*(?:allowed\s+)?to catch (\w+) here"/>
```
Pankraz76, line 6:
```java
      <property name="commentFormat" value="ok,?\s*(?:allowed\s+)?to catch (\w+) here"/>
```
every item is non-wrapped, except of this now. Its not so long and complex we need to split, or do we?

If we keep split, might align block lvl like above:
```suggestion
      <property name="commentFormat"
                value="ok,?\s*(?:allowed\s+)?to catch (\w+) here"/>
```
image

Pankraz76 (changes requested):
indentation/format

src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
```diff
@@ -149,7 +149,8 @@ public class Example2 {
 &lt;module name="Checker"&gt;
   &lt;module name="TreeWalker"&gt;
     &lt;module name="SuppressWithNearbyCommentFilter"&gt;
-      &lt;property name="commentFormat" value="OK to catch (\w+) here"/&gt;
+      &lt;property name="commentFormat"
+                value="ok,?\s*(?:allowed\s+)?to catch (\w+) here"/&gt;
```
Pankraz76, line 153:
```suggestion
      &lt;property name="commentFormat"
                   value="ok,?\s*(?:allowed\s+)?to catch (\w+) here"/&gt;
```
could be done here too, but not so important as below in example.

Pankraz76 (approved):
Thanks for adjustment

src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
smita1078: This is generated at runtime.. from the Example3.java , If I forcefully change...then it will give error..
image

Pankraz76: Yes its same on main, then inline as updated suggestion or revert.
Checking the key/value mostly appears inlined:

image

smita1078: Yes, mostly they are inline but there is a char length error for this as it goes beyond 85 characters, therefore moved it to another line

config/checkstyle-checks.xml
```diff
@@ -200,6 +200,13 @@
     <property name="message"
               value="@since tag must follow the X.Y.Z version format after version 10.5.0."/>
   </module>
+  <module name="RegexpSingleline">
+    <property name="id" value="okViolationCommentFormat"/>
+    <property name="format"
+              value="^\s*//(?=\s*(?:ok|violation)\b)(?!\s+(?:ok|violation)\b(?:[ ,].*)?$).*$"/>
```
romani, line 206: please help me understand role of this Check in comparison to valiadtion that you have in test.

please help me understand this very advanced regexp, please share with me human written/brief goal of this regexp.

romani (changes requested):
thanks a lot, but please help me understand it:

config/checkstyle-checks.xml
smita1078: So in simple words,
this regex Only looks at `//… comments` where the very first word is `ok `or `violation`, and then flags it if it isn’t exactly one of our allowed forms (`// ok`… or `// violation`…

- Line must start with // (after optional indentation).
- First word after // must be ok or violation (exactly).
- If ok or violation is used, it must have a space or comma after it.
- No other formats are allowed.


`^\s*//`
Meaning:
The line starts (^) with optional whitespace (\s*) followed by a // comment. So, it targets single-line comments(including inline comments).

`(?=\s*(?:ok|violation)\b)` — (positive lookahead)
Meaning:
After the //, there must be optional whitespace (\s*), then the exact word ok or violation (\b = word boundary).
Ensures the comment starts with ok or violation.

`?!\s+(?:ok|violation)\b(?:[ ,].*)?$)` — (negative lookahead)
Meaning:
BUT: the comment must not be a valid format like:
// ok
// ok, explanation
// violation some reason

The pattern inside says: After //, if there's: whitespace, then ok or violation, followed by a space or comma, and nothing more (or a valid explanation), Then it’s not a match — meaning don’t flag it.
So this rule is only triggered when someone tries to write a // ok or // violation comment in the wrong format.

` .*`
Matches the rest of the line (the full line is checked)

romani: Thanks a lot, why we care so much on comments that are alone on lines? 

What this regexp validation is covering that is not covered by junit validation of Input?

smita1078: Alone on line because there are certain places where I see that javadoc comment / multiline comments also has ok or violation.. wordings.. if we are not specific with the particular type of comment to be checked then we will receive error for something that is not supposed to be wrong..
for eg:
```
 * and removing parentheses will give a compile-time error. Even if <em>q</em>
 * and <em>r</em> were {@code boolean} still there will be no violation
 * raised as check is not "type aware".
 
```
In this javadoc comment, we see violation word on line 2 if we are not specific about the start and line of the comment then there could be such cases.. 
we are only flagging those comments which start with` // `then have next word as either ok or violation, inline or on a new line...moreover the issue was specifically regarding the format of ok and violation comments which are all standalone comments in the repo.

So, when I first implemented parser logic then it only covered file under `src/test/resources` and not all java files in the repo..so in order to increase coverage regex logic is also there, the regex rule makes sure all of our main java files and docs never slip out of the approved comment format. As both parser logic and regex logic are targeting different set of files..

romani: I checkouted to 7f2abda3f9 commit (before your massive change), and applied this Check instance, execution show up few violations, mostly in xdoc xml files, due to problems in Examples, all other violations are false positives and not required.

Please remove this Check instance.
I do not see value in it, it is over complicated and prone to rise violation on not required target, junit validation is enough.

smita1078: > Please remove this Check instance. I do not see value in it, it is over complicated and prone to rise violation on not required target, junit validation is enough.

Agreed! Done.

romani:
@smita1078 , new violations in Inputs, please fix, I will merge your PR asap, to block others PRs to leak such comments.

smita1078:
> @smita1078 , new violations in Inputs, please fix, I will merge your PR asap, to block others PRs to leak such comments.

done

PR #16968, opened 2025-04-28
Issue #16857: api: add new `int process(List<Path> files)` method to RootModule - `adapt tests`

Pankraz76:
Issue #16857: api: add new `int process(List<Path> files)` method to RootModule - `adapt tests`
- #16872

PR #16977, opened 2025-04-29
Issue #16857: api: add new `int process(List<Path> files)` method to RootModule - `remove usage`

Pankraz76:
Issue #16857: api: add new `int process(List<Path> files)` method to RootModule - `remove usage`
- #16968


Pankraz76:
@romani this seems to be the version you desire. api + migration.

nrmancuso:
Seems reasonable, but I am not sure about the development approach here. I think we could iterate faster if we set up a simple project just with grammar, a node, and an AST builder with a lightweight test framework. 

To me, it is distracting that we have all this other checkstyle code and a bunch of disabled stuff.

Ultimately it is up to Mohamed, obviously; I am just sharing my opinion here.

PR #17004, opened 2025-05-02
Issue #16361: Update SarifLoggerTest.testAddErrorAtColumn1 to use verifyWithInlineConfigParserAndLogger

Praveen7294:
Part of #16361 

updated `testAddErrorAtColumn1()` method to use `verifyWithInlineConfigParserAndLogger()`.


romani:
Direction is correct, but we lost test coverage, please try to restore it, CI should be fully green.
Pitest execution can be done on local to validate that no mutation happening before pushing to GitHub

romani (approved):
thanks a lot.

PR #17003, opened 2025-05-02
Issue #16738: Update SuppressWithNearbyCommentFilterExamplesTest to use verifyFilterWithInlineConfigParser

ZiadAbdElFatah:
Update `SuppressWithNearbyCommentFilterExamplesTest` to use `verifyFilterWithInlineConfigParser`
Issue: #16738

Issue #16738, opened 2025-04-02
Update Filters Exampes to use verifyFilterWithInlineConfigParser


romani:
Good example of expected update 
https://github.com/checkstyle/checkstyle/pull/16621/files

We need the same update for all Filters at https://github.com/checkstyle/checkstyle/tree/master/src/xdocs-examples/java/com/puppycrawl/tools/checkstyle/filters

Please send separate PRs for each Filter.

- [ ] SeverityMatchFilterExamplesTest.java
- [x] SuppressWarningsFilterExamplesTest.java
- [x] SuppressWithNearbyCommentFilterExamplesTest.java
- [x] SuppressWithNearbyTextFilterExamplesTest.java
- [x] SuppressWithPlainTextCommentFilterExamplesTest.java
- [x] SuppressionCommentFilterExamplesTest.java
- [x] SuppressionSingleFilterExamplesTest.java
- [x] SuppressionXpathSingleFilterExamplesTest.java

if required, it is ok to create new versions of verifyFilterWithInlineConfigParser to match what we need to validate filtered.
Fiter Exampes that use verifyWithInlineConfigParserSeparateConfigAndTarget might need new method.

Kushagra-Sedwal:
I am on SuppressionCommentFilterExamplesTest.java

Deb32800:
@romani 
I’ve made the necessary updates to the filters using verifyFilterWithInlineConfigParser on SuppressionCommentFilterExamplesTest as per the requirement. The PR for the changes is now ready. Please have a look at it when you have a chance. Let me know if there are any additional changes needed.

Thanks!

Praveen7294:
@romani can you help me in below problem?

i was working on `SeverityMatchFilterExampleTest.java` to use `verifyFilterWithInlineConfigParse()`.

### Example1.java
```
/*xml
<module name="Checker">
  <module name="TreeWalker">
    <module name="ParameterName">
      <property name="severity" value="info"/>
    </module>
    <module name="MethodName"/>
  </module>
  <module name="SeverityMatchFilter">
    <property name="severity" value="info"/>
    <property name="acceptOnMatch" value="false"/>
  </module>
</module>
*/
package com.puppycrawl.tools.checkstyle.filters.severitymatchfilter;

// xdoc section -- start
public class Example1 {
  // filtered violation below 'must match pattern'
  public void method1(int V1){} // ok, ParameterName's severity is info

[...]

PR #17008, opened 2025-05-03
Issue #16738: Update SuppressWithNearbyCommentFilterExamplesTest to use verifyFilterWithInlineConfigParser

ZiadAbdElFatah:
A clean PR for this one #17003
Update `SuppressWithNearbyCommentFilterExamplesTest` to use `verifyFilterWithInlineConfigParser`
Issue: #16738

### DIFF
--- a/src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
+++ b/src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
@@ -114,6 +114,7 @@
         <p id="Example1-code">Example:</p>
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
 public class Example1 {
+  // filtered violation below ''int' is followed by whitespace'
   private int [] array; // SUPPRESS CHECKSTYLE NoWhitespaceAfter
 }
 </code></pre></div><hr class="example-separator"/>
@@ -136,6 +137,7 @@ public class Example1 {
         <p id="Example2-code">Example:</p>
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
 public class Example2 {
+  // filtered violation below 'must match pattern'
   public static final int lowerCaseConstant = 1; // CHECKSTYLE IGNORE THIS LINE
 }
 </code></pre></div><hr class="example-separator"/>
@@ -166,6 +168,7 @@ public class Example3 {
     try {
       // blah blah blah
     }
+    // filtered violation below 'Catching 'RuntimeException' is not allowed'
     catch(RuntimeException re) {
       // ok, allowed to catch RuntimeException here
     }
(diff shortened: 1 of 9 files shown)


src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
```diff
@@ -114,6 +114,7 @@
         <p id="Example1-code">Example:</p>
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
 public class Example1 {
+  // filtered violation below
```
romani, line 117: please add message of filtered violation 

search `filtered violation below` in our sources, there are a lot examples

romani (changes requested):
good!

small item to improve:

src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
ZiadAbdElFatah: I made the changes except for `Example6.java` as it got two violation in the same line and I couldn't arrange the checks as I wanted, so there was an error that the `actualViolations` order doesn't match the `expectedViolations` order.
Also, there is a violation in `Example8.java` exceeds the limit of characters in `LineLength` and couldn't solve this either.

src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
```diff
@@ -245,7 +253,8 @@ public class Example5 {
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
 public class Example6 {
   // @cs.suppress [ConstantName|NoWhitespaceAfter] A comment here
-  public static final int [] array = {};
+  public static final int [] array = {}; // filtered violation
+  // filtered violation above
```
romani, line 257: this is right way to address two violations on same line, please add messages or explain more clearly what stoppping you to make it.

romani (changes requested):
items:

src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
romani: > violation in Example8.java exceeds the limit of characters in LineLength and couldn't solve this either.

cut message, we just need part of message, not whole message

> got two violation in the same line and I couldn't arrange the checks as I wanted, so there was an error that the actualViolations order doesn't match the expectedViolations order.

do you mean that order of violations is always different? they supposed to be sorted, to ease matching.

ZiadAbdElFatah: > 
> do you mean that order of violations is always different? they supposed to be sorted, to ease matching.
> 
```
Actual and expected violations differ.
expected to match: 19:(?:\d+:)?\s.*must match pattern.*
but was          : 19:27: 'int' is followed by whitespace.
```
I got this error because of the difference between the order of  `testInputViolation` and `actualViolations`

Screenshot from 2025-05-04 18-10-33

romani: Do you see I'm code how to fix ordering? Don't hesitate to change code to make it work

ZiadAbdElFatah: Can this be considered as an issue?

romani: Yes, if update is big, it can be issue.
Let me know if you want to split it to separate issue.

ZiadAbdElFatah: @romani 
I figured out why this problem happens but I think the update would be big.
```
violationsWithoutFilters.addAll(testInputConfiguration.getFilteredViolations());
        Collections.sort(violationsWithoutFilters);
```
here after adding the `filteredViolations` to the `violationsWithoutFilters` and then sort it according to line number.
The problem happens when there are more than one violation in the same line, the sort doesn't then sort according to column number this is why the `testInputViolations` doesn't have the same order of `actualViolations`

I think we can add `colNo` instance variable to the violations as `lineNo`.

romani: Please create new issue on it, we will follow up after merge of this PR

ZiadAbdElFatah: I created a new issue. I think we can merge this PR and once the issue is solved, we can edit `Example6.java`

romani (approved):
Goog step forward, with smaller problem to fix separately

PR #17026, opened 2025-05-06
Issue #5168: Update XML version in output files to 1.1

Brijeshthummar02:
Issue #5168


Brijeshthummar02:
Note: i have not updated all instance of xml just to check if it goes well with this much we can do bump for all.

Brijeshthummar02:
@romani i guess issue has not been resolved ci goes down

romani:
Issue in maven is still not fixed.
We are blocked

Issue #17123, opened 2025-05-22
Missing MultiFileRegexpHeader into Header index.html


Abdelrhmansersawy:
I noticed that the [`MultiFileRegexpHeader`](https://checkstyle.org/checks/header/multifileregexpheader.html) check is **not listed** in the [Header Checks index page](https://checkstyle.org/checks/header/index.html).

#### Screenshot:

![Missing MultiFileRegexpHeader](https://github.com/user-attachments/assets/2ed70345-72d5-4812-af91-cb2ef709b14f)

This check should be included in the list to ensure consistency and discoverability for users browsing the available header checks.

PR #17124, opened 2025-05-22
Add Multifileregexpheader check section into header index.html

Abdelrhmansersawy:
Issue #17123

### DIFF
--- a/src/site/xdoc/checks/header/index.xml
+++ b/src/site/xdoc/checks/header/index.xml
@@ -21,6 +21,17 @@
           </tr>
           <tr>
             <td>
+              <a href="multifileregexpheader.html#MultiFileRegexpHeader">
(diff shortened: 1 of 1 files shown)


Abdelrhmansersawy:
GitHub, generate site

romani:
Do you see a way to validate this by junit?
To avoid this problem in future

Abdelrhmansersawy:
> Do you see a way to validate this by junit?
> To avoid this problem in future 

Not I noticed that while browsing a website.

But I will go deep into the cycle of validation to find a way to validate it by junit.

Abdelrhmansersawy:
I just reported a bug so that someone can find a solution.
In the meantime, I’ll also try to help with this issue.

Abdelrhmansersawy:
GitHub, generate site

Issue #17168, opened 2025-06-04
Preparation for Migration of  Checkstyle Codebase to Java 17


AmitKumarDeoghoria:
Currently, Checkstyle is using JDK 11 as the baseline version. However, Java has evolved significantly, and most of the ecosystem has now moved towards JDK 21.

This issue proposes to incrementally upgrade Checkstyle’s baseline JDK version in two stages:

- move all Inputs for JDK 12-17 to compiled folder ✅
- update all CIs to use jdk17, there should be no jdk11 be left. Still  not updates in pom.xml for required jdk. ✅
- send PR to upgrade pom.xml to min required jdk, to see CI and make further plan. ✅
- update pom.xml - it should be separate issue as it is big breaking change and it would be visible in release notes clearly. Version will be `11.0.0` ✅
- revert changes in pom.xml to make Inputs be compiled by maven.
- review all PRs with dependencies and prepare them for merge, may be just rebase in UI will be required.

romani:
Doing all at one will be too challenging.

Please send PR to move all non completable  Inputs from jdk less or equal to jdk17 to compiled folder and update CI to do not have jdk11 compile for inputs

AmitKumarDeoghoria:
> Doing all at one will be too challenging.
> 
> Please send PR to move all non completable Inputs from jdk less or equal to jdk17 to compiled folder and update CI to do not have jdk11 compile for inputs

Made these changes in the PR #17169. Now as a next step will resolve all test failures as a result of moving all these files

romani:
I put misc label, because for now we do not break compatibility.
upgrade of pom.xml will be breaking changes and we bump version to 11.0.0

romani:
All further updates should be in scope of https://github.com/checkstyle/checkstyle/issues/17321

PR #17178, opened 2025-06-07
Issue #17168: Relocate inputs to compilable path for Block Check

AmitKumarDeoghoria:
Part of: #17168


romani:
you can do fisrt PR to just move CI jobs to jdk17, as java is backward compatible, it should work well.
if some jobs are not possbile to move easily, we need to deal with them in separate PR

Issue #17186, opened 2025-06-08
Validattion of xdoc files to have separtor between examples


romani:
JavadocPackage's Examples are missing hr tags: https://checkstyle.org/checks/javadoc/javadocpackage.html#Examples

it is leftover from https://github.com/checkstyle/checkstyle/issues/16333

Interesting few example, but for same config (as no properties in Check) - https://checkstyle.org/checks/coding/innerassignment.html#InnerAssignment
should we put HR in between? 

We need to:
- add separator, done at https://github.com/checkstyle/checkstyle/pull/17187
- And think about how we have make junit validation that all examples are separated in xdoc files.

smita1078:
Hi all,
In current PR the missing hr tags are fixed and I am currently thinking of how we can add test validations for the separators, will think of an approach and raise PR for same

romani:
Thanks a lot to be back.
I think we can detected that there is no required tags between reference to Example1 and Example2.
In this case we just hit that there is no separation in between, with validation of exact location of separation

romani:
@smita1078 ,

Interesting few example, but for same config (as no properties in Check) - https://checkstyle.org/checks/coding/innerassignment.html#InnerAssignment
should we put HR in between? 

I think it will be good to have it.

smita1078:
> Interesting few example, but for same config (as no properties in Check) - https://checkstyle.org/checks/coding/innerassignment.html#InnerAssignment should we put HR in between?
> 
> I think it will be good to have it.

Yes Sure, we can separate the two examples

smita1078:
> I think we can detected that there is no required tags between reference to Example1 and Example2.
> In this case we just hit that there is no separation in between, with validation of exact location of separation

So basically, correct me if I am wrong when we detect the start of a new example (ExampleX) and where the previous example (ExampleX-1) ended, we can verify that there’s an `<hr>` tag right between them.

romani:
@smita1078 , yes. This simple solution should work decently well.

smita1078:
> [@smita1078](https://github.com/smita1078) , yes. This simple solution should work decently well.

Understood!

smita1078:
> Interesting few example, but for same config (as no properties in Check) - https://checkstyle.org/checks/coding/innerassignment.html#InnerAssignment should we put HR in between?

I have put `hr` tag there as well as in another file where it was missing.. and now will start looking into junit validation

romani:
fix is merged.

PR #17201, opened 2025-06-10
Junit Validation Test for example separator in xdoc files

smita1078:
#17186 

This PR adds a test `validateExampleSectionSeparation()` to validate the structure of Examples subsections within XDoc .xml.template files.

The test ensures that:
Each transition between distinct example groups (e.g., Example1 → Example2) is separated by an `<hr class="example-separator"/>` tag.
The test: 
- Parses each .xml.template as a DOM document.
- Finds all <subsection name="Examples"> blocks.
- Iterates over their child elements.
- Tracks the current example prefix (from id attributes like `Example1-desc` or` Example2-code`).
- On detecting a change in prefix (i.e., a new example block), asserts that an `<hr>` was seen since the last example group.
- Fails if a transition occurs without the required separator.

image
Result of assertion.

Also added `hr` separator to the files that gave error after adding test.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
@@ -19,6 +19,7 @@
 
 package com.puppycrawl.tools.checkstyle.internal;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.truth.Truth.assertWithMessage;
 import static java.lang.Integer.parseInt;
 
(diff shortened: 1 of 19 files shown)


smita1078:
Github, generate site

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
```diff
@@ -2247,4 +2252,86 @@ private static Node getPrecedingParagraph(Node macro) {
...
+
+    /**
+     * Verifies that all example blocks in XDoc templates are correctly separated by
+     * <hr class="example-separator"/> tags to maintain visual consistency.
+     */
+    @Test
+    public void validateExampleSectionSeparation() throws Exception {
+        final Path root = Paths.get("src/site/xdoc");
+        // Collect all .xml.template files under src/site/xdoc as an immutable list
+        final List<Path> templates = Files.walk(root)
```
romani: please remove all comments from code, OR is if you like them you can convert them to functions

`Collect all .xml.template files under src/site/xdoc as an immutable list` --> CollectAllXmlTemplatesUnderSrcSite()

i hope you pattern I used to convert. Please do the same for others.

reason is comment are always becomes outdated, method names can keep same value but engineers more likely to update name of method than a comment.

romani (changes requested):
nice !!

some clenaup:

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java
smita1078: Okay👍

smita1078: Done

romani:
Nice, please make inspections to be happy 
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/34389/workflows/055099ae-c057-4c41-8987-6e88ba700863/jobs/966969/artifacts

romani (approved):
thanks a lot

PR #17215, opened 2025-06-12
Issue #17168: Moved Parser Module Inputs to Compilable Folder

AmitKumarDeoghoria:
part of : #17168


src/test/java/com/puppycrawl/tools/checkstyle/xpath/XpathMapperTest.java
```diff
@@ -1256,12 +1256,6 @@ private RootNode getRootNodeWithComments(String fileName) throws Exception {
         return new RootNode(rootAst);
     }
 
-    private RootNode getRootNodeForNonCompilable(String fileName) throws Exception {
```
romani, line 1259: Please move it to abstract class, to be used when we come to New jdk versions syntax, so it will be in need again

romani (changes requested):
Minor

romani:
@AmitKumarDeoghoria , please do above in separate PR

PR #17241, opened 2025-06-18
Issue #17168: Upgrading CI on semaphore to jdk17

AmitKumarDeoghoria:
part of : #17168


.semaphore/semaphore.yml
```diff
@@ -77,7 +77,7 @@ blocks:
                 - .ci/validation.sh jacoco
 
           commands:
```
romani, line 79: Please change names to be jdk17 

>No Error Test (openjdk11, fast pool)

We eliminating jdk11 at all, no eny reference should be left

AmitKumarDeoghoria: Done

romani (approved):
Thanks a lot

PR #17247, opened 2025-06-19
Issue #13746: Update PropertyCacheFileTest.testPathToCacheContainsOnlyFileName to delete temp.cache directory

Harsh-Agrawal96:
Issue #13746 

The test removes the `temp.cache` file.
But After runing `$ ./.ci/pitest.sh pitest-common` for pitest testing create the `temp.cache` folder and we were expecting the file so it gives the error and test fails.

**Solution** :  Add the code snippet that remove the `temp.cache` folder if it empty.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
@@ -40,6 +40,7 @@ import java.util.Locale;
 import java.util.Properties;
 import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Stream;
 
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledOnOs;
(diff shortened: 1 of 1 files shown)


Harsh-Agrawal96:
@romani all CLI passed, please review the PR

src/test/java/com/puppycrawl/tools/checkstyle/PropertyCacheFileTest.java
```diff
@@ -281,6 +282,14 @@ public void testPathToCacheContainsOnlyFileName() throws IOException {
         final Path filePath = Path.of(fileName);
         final PropertyCacheFile cache = new PropertyCacheFile(config, fileName);
 
+        if (Files.exists(filePath) && Files.isDirectory(filePath)) {
+            try (Stream<Path> entries = Files.list(filePath)) {
+                if (entries.findAny().isEmpty()) {
+                    Files.delete(filePath);
```
romani: why not simply:
```

        if (Files.exists(filePath) && Files.isDirectory(filePath)) {
                    Files.delete(filePath);
```

romani: we do not need folder even there is something in it.
Path can not be folder,  it should be file only.

Harsh-Agrawal96: If path is file then it is handling in further code but if it is directory(folder) we need to remove this first.

PR #17268, opened 2025-06-23
Issue #17168: Upgrade Azure CI to jdk17

AmitKumarDeoghoria:
part of : #17168

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -141,7 +141,7 @@ steps:
 
   - task: JavaToolInstaller@0
     inputs:
-      versionSpec: 11
+      versionSpec: 17
       jdkArchitectureOption: 'X64'
       jdkSourceOption: 'PreInstalled'
 


AmitKumarDeoghoria:
Reaason for changes : In Java 17, stricter encapsulation (via JEP 403) hides internal classes like sun.awt.X11.XToolkit, and AWT toolkit initialization now happens earlier—before checking for headless mode. Java 11 allowed delayed, more lenient access with fallback mechanisms that avoided X11 issues in headless environments. Java 17 also removed those fallbacks and attempts X11 connections immediately, causing failures unless headless mode is explicitly set or a virtual display (like Xvfb) is used. These changes support better security and modularity but require adjustments in older code.

romani:
>Reaason for changes :

Please share link to failures.
We supposed to not use such classes, if we do, it is mistake we need to fix

azure-pipelines.yml
```diff
@@ -157,15 +206,31 @@ steps:
...
+      echo "ON_CRON_ONLY: $ON_CRON_ONLY"
+      echo "BUILD_REASON: $BUILD_REASON"
+      echo "REQUIRE_DISPLAY: $REQUIRE_DISPLAY"
+      echo "AGENT_OS: $AGENT_OS"
+      echo "cmd: $(cmd)"
+      # Conditionally run with xvfb for Linux GUI tests
+      if [ "$REQUIRE_DISPLAY" = "true" ] && [ "$AGENT_OS" = "Linux" ]; then
+        echo "Starting xvfb for GUI tests"
+        export DISPLAY=:99
+        Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
```
romani: Please share a root reason failure that forcing you to go this route.

romani (changes requested):
Items

azure-pipelines.yml
AmitKumarDeoghoria: Ci failure link : https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=30381&view=logs&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111
Screenshot 2025-06-24 at 9 49 41 AM

AmitKumarDeoghoria: Spent a lot of time is understanding the error, and from different sources after investigation I found this as a working way.

AmitKumarDeoghoria:
@romani pls review this.

azure-pipelines.yml
romani: This can mean that the same failure can be on users locals.
If that is potentially possible, I would rather exclude tests for this UI classes.

romani:
`[ERROR]   MainFrameTest.tearDown:71 NullPointer Cannot invoke "com.puppycrawl.tools.checkstyle.gui.MainFrame.getOwnedWindows()" because "this.mainFrame" is null`

This is not good.

But good that it is not requiring hacking on OS.
https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=30448&view=logs&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111&l=1299

Can we update `MainFrameTest.tearDown:71` to avoid exception and keep it working?
If not, please share reasoning and 
Lets disable MainFrameTest completely.

AmitKumarDeoghoria:
getting coverage issue :  CI Link : https://checkstyle.semaphoreci.com/jobs/3cc29450-ef32-4a66-81dd-31d3f6ee487c
Screenshot 2025-06-27 at 12 42 55 PM

AmitKumarDeoghoria:
@romani Please let me know what should be the next step.
I am thinking of adding a profile in pom for excluding these tests for the CI for now, let us complete the upgradation and then this can later be a focus whether there is a possibility to make these test compatible without doing much hassle or hacking.

AmitKumarDeoghoria:
I tried resolving the coverage issue, by adding more test, but it's not improving. I think I don't have proper knowledge to improve it. Do let me know what should be the next step

romani:
this test was added at https://github.com/checkstyle/checkstyle/commit/1c139d87f71c1580eaf480910db14520f33d74f9

lets disable MainFrameTest completely for now.
UI tool is in low usage from checkstyle, as we developed good CLI to print trees. Eventually  all gui code should be moved out to separate project. 

create separate issue to enable MainFrameTest back, put comment "until ......" above disablement annotation. We will handle it separately.

AmitKumarDeoghoria:
if we disable test by annotation we will still get jacoco coverage issue, we will need to add property to skip such test, I plan to skip test using these : https://github.com/checkstyle/checkstyle/pull/17290 Please let me know what do u think?

romani:
Please show me link to error.
We can exclude such classes from jacoco validation or reduce such class coverage numbers in pom.xml , we already have some exclusions from 100% coverage.

https://github.com/checkstyle/checkstyle/blob/781094d39ae38789285576c6b7d8cf00b8e098bc/pom.xml#L989

AmitKumarDeoghoria:
CI error unrelated. Seems okay to merge after : https://github.com/checkstyle/checkstyle/pull/17292

romani (approved):
Just version bump. Good now

PR #17275, opened 2025-06-24
Issue #17168: Removing javac 11  job from CircleCI

AmitKumarDeoghoria:
part of : #17168

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -515,7 +515,7 @@ compile-test-resources)
   -Dcheckstyle.skipCompileInputResources=false -Dmaven.compiler.release=17
   ;;
 
-javac11)
+javac17_standard)
   # InputCustomImportOrderNoPackage2 - nothing is required in front of first import
   # InputIllegalTypePackageClassName - bad import for testing
   # InputVisibilityModifierPackageClassName - bad import for testing
@@ -530,7 +530,7 @@ javac11)
   mkdir -p target
   for file in "${files[@]}"
   do
-    echo "${file}"
+    echo "Compiling ${file} with standard JDK17"
     javac -d target "${file}"
   done
   ;;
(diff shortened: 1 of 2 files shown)


.ci/validation.sh
```diff
@@ -515,25 +515,6 @@ compile-test-resources)
   -Dcheckstyle.skipCompileInputResources=false -Dmaven.compiler.release=17
   ;;
 
-javac11)
```
romani, line 518: We migration of this to jdk17. Not a removal.
Goal is to compile sources that are not compiled by IDEs for any reason, but compiled fine by javac.

romani (changes requested):
Items

.ci/validation.sh
AmitKumarDeoghoria: I dont understand, don't we already have a similar job for jdk17, it should work fine right?

.ci/validation.sh
```diff
@@ -515,25 +515,6 @@ compile-test-resources)
...
-  # InputCustomImportOrderNoPackage2 - nothing is required in front of first import
-  # InputIllegalTypePackageClassName - bad import for testing
-  # InputVisibilityModifierPackageClassName - bad import for testing
-  files=($(grep -REL --include='*.java' \
```
romani: I don't see that this code exists in jdk17.
This job exists for files that supposedly compile by our minimum required jdk but they are not due to some problems with class path.
So Maven can not compile them, but javac can

romani (changes requested):
Items

.ci/validation.sh
romani: See below

romani:
@AmitKumarDeoghoria , please rebase to resolve problems

.ci/validation.sh
AmitKumarDeoghoria: @romani  Plz review.

PR #17259, opened 2025-06-22
Issue #17168: Upgrade Pom Java version to 17

AmitKumarDeoghoria:
part of : #17168

PR #17312, opened 2025-07-03
Issue #17168: Fix inspection for private methods used from inner class

AmitKumarDeoghoria:
Part of : #17168
From comment : https://github.com/checkstyle/checkstyle/pull/17259#issuecomment-3019192404

### DIFF
--- a/config/spotbugs-exclude.xml
+++ b/config/spotbugs-exclude.xml
@@ -404,4 +404,11 @@
     <Method name="getTargetFiles" />
     <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE" />
   </Match>
+
+  <!-- until https://github.com/mebigfatguy/fb-contrib/issues/491 -->
+  <Match>
+    <Class name="com.puppycrawl.tools.checkstyle.ConfigurationLoader" />
+    <Field name="ID_TO_RESOURCE_NAME_MAP" />
+    <Bug pattern="DMC_DUBIOUS_MAP_COLLECTION" />
+  </Match>
 </FindBugsFilter>
--- a/config/jsoref-spellchecker/whitelist.words
+++ b/config/jsoref-spellchecker/whitelist.words
@@ -325,6 +325,7 @@ Djacoco
 Dlinkcheck
 DMail
 Dmaven
+DMC
 Dnew
 doccheck
 Dockter
--- a/src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java
@@ -25,7 +25,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Deque;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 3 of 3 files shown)


romani:
Spotbugs is not related. But inspection should be green

romani:
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/34767/workflows/e5c2896c-5304-4a86-9914-2c14b74e8b73/jobs/984958/artifacts
```
 file does not appear to have any style information associated with it. The document tree is shown below.
<problem>
<file>file://$PROJECT_DIR$/src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java</file>
<line>144</line>
<module>project</module>
<package>com.puppycrawl.tools.checkstyle</package>
<entry_point TYPE="class" FQNAME="com.puppycrawl.tools.checkstyle.ConfigurationLoader"/>
<problem_class id="Java9CollectionFactory" severity="WEAK WARNING" attribute_key="INFO_ATTRIBUTES">Immutable collection creation can be replaced with collection factory call</problem_class>
<description>Can be replaced with 'Map.of()' call</description>
<highlighted_element>unmodifiableMap</highlighted_element>
<language>JAVA</language>
<offset>46</offset>
<length>15</length>
</problem>
</problems>
```

AmitKumarDeoghoria:
Proof that after this inspection will be removed in #17259  : https://app.circleci.com/pipelines/github/checkstyle/checkstyle/34775/workflows/f8fddada-8c72-41dd-bd43-f7cac7921204/jobs/985341/artifacts

romani:
What violation you try to resolve by this update?
It is weird that we try to resolve some and cause few other.
We might need to suppress inspection.

romani:
Is it 
```

<problem>
<file>file://$PROJECT_DIR$/src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java</file>
<line>169</line>
<module>project</module>
<package>com.puppycrawl.tools.checkstyle</package>
<entry_point TYPE="method" FQNAME="com.puppycrawl.tools.checkstyle.ConfigurationLoader java.util.Map<java.lang.String,java.lang.String> createIdToResourceNameMap()"/>
<problem_class id="MethodOnlyUsedFromInnerClass" severity="ERROR" attribute_key="ERRORS_ATTRIBUTES">Private method only used from inner class</problem_class>
<description>Method <code>createIdToResourceNameMap()</code>#loc is only used from inner class 'InternalLoader' #loc</description>
<highlighted_element>createIdToResourceNameMap</highlighted_element>
<language>JAVA</language>
<offset>39</offset>
<length>25</length>
</problem>
</problems>
```

But in this case why not to move it to inner class?

romani:
Suppression on
`DMC_DUBIOUS_MAP_COLLECTION This method instantiates a map-type field in a static initializer or constructor, but then only uses it through iteration. This means that this data structure should really just be a List<SomeObject>, where the class held by the list contains the two fields held by the key and value of the Map. It was likely done this way to avoid having to create a class, but this just obfuscates the purpose of the field.`

https://fb-contrib.sourceforge.net/bugdescriptions.html

This is false positive from fb extension of spotbugs .
We need map https://github.com/checkstyle/checkstyle/blob/02cbc39d617d231e8b5b84ce5b5d30752c4108d1/src/main/java/com/puppycrawl/tools/checkstyle/XmlLoader.java#L69

@AmitKumarDeoghoria , please create defect on fb extension and put until ticket in comment of suppression

config/spotbugs-exclude.xml
```diff
@@ -404,4 +404,11 @@
     <Method name="getTargetFiles" />
     <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE" />
   </Match>
+
+  <!-- Suppress DMC_DUBIOUS_MAP_COLLECTION for ConfigurationLoader.ID_TO_RESOURCE_NAME_MAP -->
```
romani: Comment is required to show link until what defects it should stay.
Not to repeat what is written in config itself

romani (changes requested):
Items

AmitKumarDeoghoria:
> But in this case why not to move it to inner class?

From what I understand Even if currently it's only used by the inner class, the method is logically a utility of the outer class and reflects its responsibility. Moving it to the inner class I think would mix concerns and reduce clarity, even if it's currently only used there. It's jut matter of good design over narrow scope.

config/spotbugs-exclude.xml
AmitKumarDeoghoria: Done

AmitKumarDeoghoria:
> What violation you try to resolve by this update? It is weird that we try to resolve some and cause few other. We might need to suppress inspection.

Yes the violation u suggested in above comment , After current update no further inspection should come out, I checked the same in pom update PR, and attached the proof.

romani (approved):
Ok to merge

PR #17323, opened 2025-07-06
Issue #17168: Remove Jdk11 Site Validation Job

AmitKumarDeoghoria:
part of : #17168
Redundant Job, since anchor image is already update to jdk 17 and we already have a package site validation for jdk17.


romani (approved):
thanks a lot

Issue #17366, opened 2025-07-12
FinalParameters - missing several tokens to check


Machine-Maker:
https://checkstyle.org/checks/misc/finalparameters.html#FinalParameters
```bash
/var/tmp $ javac BrokenFile.java


/var/tmp $ cat config.xml
<!DOCTYPE module PUBLIC
    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
    "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
  <module name="TreeWalker">
    <module name="FinalParameters"/>
  </module>
</module>

/var/tmp $ cat BrokenFile.java
final class OuterClass {

  record Record(String name) {
  }

  interface BrokenInterface {
    static void test(Object o) { // violation expected
    }

    private void test2(Object o) { // violation expected
      if (o instanceof String s) { // violation expected
        System.out.println("s");
      } else if (o instanceof Record(String name)) { // violation expected 
        System.out.println(name);
      }
    }
  }
}

/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.26.1-all.jar -c config.xml BrokenFile.java
Starting audit...
Audit done.
```

---

There should be several violations, for parametrs of the static and private method  inside `BrokenInterface` as well as an error for the pattern variable `s` in the first if check *and* the record variable `name`.

-----

Suppression config is provided in case somebody does not like this new behavior https://github.com/checkstyle/checkstyle/issues/17366#issuecomment-3067362895

Machine-Maker:
When the fix is merged, the new violations can be suppressed with:
```xml
<!DOCTYPE module PUBLIC
    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
    "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
  <module name="TreeWalker">
    <module name="FinalParameters"/>
    <module name="SuppressionXpathSingleFilter">
      <property name="checks" value="FinalParameters"/>
      <property name="query" value="//INTERFACE_DEF/descendant-or-self::node()"/>
    </module>
  </module>
</module>
```

romani:
Fix is merged.
Check got new Token to execute on, but it is not default token.

PR #17368, opened 2025-07-12
Issue #17366: FinalParameters should check interface methods and patt…

Machine-Maker:
Fixes https://github.com/checkstyle/checkstyle/issues/17366

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -655,7 +655,7 @@ no-error-orekit)
   checkout_from https://github.com/Hipparchus-Math/hipparchus.git
   cd .ci-temp/hipparchus
   # checkout to version that Orekit expects
-  SHA_HIPPARCHUS="815ad2bf9ce764e4498911d2145c49165f5f3333"
+  SHA_HIPPARCHUS="1492f06848f57e46bef911a""ad16203a242080028"
   git checkout $SHA_HIPPARCHUS
   mvn -e --no-transfer-progress install -DskipTests
   cd -
@@ -664,7 +664,7 @@ no-error-orekit)
   # no CI is enforced in project, so to make our build stable we should
   # checkout to latest release/development (annotated tag or hash) or sha that have fix we need
   # git checkout $(git describe --abbrev=0 --tags)
-  git checkout "a32b4629b2890fc198b19a95a714d67b87d7943d"
+  git checkout "9b121e504771f3ddd303ab""cc""c74ac9db64541ea1"
   mvn -e --no-transfer-progress compile checkstyle:check \
     -Dorekit.checkstyle.version="${CS_POM_VERSION}"
   cd ..
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/FinalParametersCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/FinalParametersCheck.java
@@ -155,6 +155,7 @@ public class FinalParametersCheck extends AbstractCheck {
             TokenTypes.CTOR_DEF,
             TokenTypes.LITERAL_CATCH,
             TokenTypes.FOR_EACH_CLAUSE,
+            TokenTypes.PATTERN_VARIABLE_DEF,
         };
     }
 
@@ -165,19 +166,27 @@ public class FinalParametersCheck extends AbstractCheck {
 
     @Override
     public void visitToken(DetailAST ast) {
-        // don't flag interfaces
-        final DetailAST container = ast.getParent().getParent();
-        if (container.getType() != TokenTypes.INTERFACE_DEF) {
-            if (ast.getType() == TokenTypes.LITERAL_CATCH) {
-                visitCatch(ast);
-            }
-            else if (ast.getType() == TokenTypes.FOR_EACH_CLAUSE) {
-                visitForEachClause(ast);
-            }
-            else {
-                visitMethod(ast);
-            }
+        if (ast.getType() == TokenTypes.LITERAL_CATCH) {
+            visitCatch(ast);
         }
+        else if (ast.getType() == TokenTypes.FOR_EACH_CLAUSE) {
+            visitForEachClause(ast);
+        }
+        else if (ast.getType() == TokenTypes.PATTERN_VARIABLE_DEF) {
+            visitPatternVariableDef(ast);
+        }
+        else {
+            visitMethod(ast);
+        }
+    }
+
+    /**
+     * Checks parameter of the pattern variable definition.
+     *
+     * @param patternVariableDef pattern variable definition to check
+     */
+    private void visitPatternVariableDef(final DetailAST patternVariableDef) {
+        checkParam(patternVariableDef);
     }
 
     /**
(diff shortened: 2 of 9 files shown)


Machine-Maker:
Do I have to do anything about the javac17_standard test failing? That's kinda the purpose of noncompilable test sources?

romani:
We need to send fixes to Orekit project 
https://checkstyle.semaphoreci.com/jobs/c3b3a06d-159d-4508-aed8-40876262544f
To see user reaction on new violations before release to whole world.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/finalparameters/InputFinalParameters3.java
```diff
@@ -181,3 +181,15 @@ private String[] someExpression()
 
 @interface MyAnnotation33 {
 }
+
+/* interface with methods with implementations should be flagged */
```
romani: Please create new Input file.
We try to keep the smallish to ease debugging

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/finalparameters/InputFinalParametersPatternVariables.java
```diff
@@ -0,0 +1,30 @@
+/*
```
romani, line 1: Non compiled files should have comment to show in what version they becomes compiled.
See other files as examples, you will see how to make it jdk21

romani (changes requested):
Items

Machine-Maker:
Ok, addressed those 2 points.

> We need to send fixes to Orekit project [checkstyle.semaphoreci.com/jobs/c3b3a06d-159d-4508-aed8-40876262544f](https://checkstyle.semaphoreci.com/jobs/c3b3a06d-159d-4508-aed8-40876262544f) To see user reaction on new violations before release to whole world.

The fixes aren't to anything new, it doesn't include the pattern variable def token by default, so the changes are just gonna be actual methods with method bodies on interfaces. Like [this](https://github.com/CS-SI/Orekit/blob/develop/src/main/java/org/orekit/frames/TransformProvider.java#L56), and [this](https://github.com/CS-SI/Orekit/blob/develop/src/main/java/org/orekit/propagation/events/handlers/FieldEventHandler.java#L78).

romani:
I know, but we can not merge on red CI.
May be they have suppression that we can extend?

Machine-Maker:
I guess I can also make a boolean option to skip interfaces

romani:
It is not good option.

It is easier to send PR to Orekit

Machine-Maker:
Ok, PR created: https://gitlab.orekit.org/orekit/orekit/-/merge_requests/841

Machine-Maker:
Ok, orekit updated. Should be good to go for another review.

romani:
Github, generate report for FinalParameters/all-examples-in-one

Machine-Maker:
Ok, looks like there's an issue with pattern variable def in for each, I'll check that.

EDIT: oh that was removed, and was never released as a non-preview feature. And it wasn't supported before, should it be now?

romani:
After fix please trigger regression diff testing as I did, it will speed up PR acceptance

romani:
This Check is aggressive by its nature.
But fact that we did not have complains before on missed violations might imply that world not not firm on where to put violations and where to skip.
This kind of reference to your idea to put property to skip in interfaces.

As you see diff is massive.

I don't think we should do such property now, but we should put in issue example how to bring behavior to previous state by suppression of violations under interfaces by https://checkstyle.org/filters/suppressionxpathsinglefilter.html#SuppressionXpathSingleFilter so users can simply copy paste config and not be blocked checkstyle upgrade by huge amount of violations. It might need few iterations for them to fix such violations. So suppression would be appreciated.

Machine-Maker:
Github, generate report for FinalParameters/all-examples-in-one

Machine-Maker:
Ok, suppression filter created: https://github.com/checkstyle/checkstyle/issues/17366#issuecomment-3067362895

Machine-Maker:
Ok, issue with java20 record pattern in foreach is fixed.

src/main/java/com/puppycrawl/tools/checkstyle/checks/FinalParametersCheck.java
```diff
@@ -165,33 +166,40 @@ public int[] getRequiredTokens() {
 
     @Override
     public void visitToken(DetailAST ast) {
-        // don't flag interfaces
```
romani, line 168: this comment was added 23 years ago - https://github.com/checkstyle/checkstyle/commit/f313f31ab30308796ce76d167ad1f08e79e28fdd

not clear why, so it is ok to do interfaces validation now.

Machine-Maker: I'm not sure when static methods were added to interfaces, I'm pretty sure both default and private methods were added later, like java 8/9, but I thought static methods were always available on interfaces. 🤷

.ci/validation.sh
```diff
@@ -655,7 +655,7 @@ no-error-orekit)
   checkout_from https://github.com/Hipparchus-Math/hipparchus.git
   cd .ci-temp/hipparchus
   # checkout to version that Orekit expects
-  SHA_HIPPARCHUS="815ad2bf9ce764e4498911d2145c49165f5f3333"
+  SHA_HIPPARCHUS="1492f06848f57e46bef911aad16203a242080028"
```
romani: `SHA_HIPPARCHUS="1492f06848f57e46bef911a""ad16203a242080028"`

it will help to avoid extension on whitelist, that we try to keep clean.
please  apply same trick for other SHA

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/finalparameters/InputFinalParametersRecordForLoopPatternVariables.java
```diff
@@ -0,0 +1,20 @@
...
+*/
+
+// non-compiled with javac: Compilable with Java20
+package com.puppycrawl.tools.checkstyle.checks.finalparameters;
+
+public class InputFinalParametersRecordForLoopPatternVariables {
+    record ARecord(String name, int age) {
+    }
+    static void method(final ARecord[] records) {
+        for (ARecord(String name, final int age) : records) { // violation, 'name' should be final
```
romani, line 17: please add "same" code below that is compliant  by Check, so we will see clearly what is bad and what is expected.

please apply the same rule to all other test code your created.

romani (changes requested):
last items:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/finalparameters/InputFinalParametersInterfaceMethod.java
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.finalparameters;
+
+/**
+ * Test cases for detecting missing final parameters in interface methods.
+ */
+public interface InputFinalParametersInterfaceMethod {
+    static void method1(Object param1) { // violation, 'param1' should be final
+    }
```
romani, line 17: Example on how to show good and bad in same time in most close context.
```
    static void method1B(Object param1) { // violation, 'param1' should be final
    }
    static void method1G(final Object param1) { 
    }
```

romani (changes requested):
items:

romani:
report is good.
Please do cosmetic changes requested above and we will proceed further.

Machine-Maker:
Ok, changes addressed.

Machine-Maker:
~~I'm unsure why the javac21 test is failing, that code seems to compile fine for me when I try it.~~

Ok, fixed it by updating image.

nrmancuso (approved):
Great work, thanks for your contribution!

nrmancuso:
rebasing on latest master before merge

PR #17407, opened 2025-07-16
Fix false positive in ImportOrder with static imports inside config example

Shivang345:
### What was the problem:
There was a false positive in the `ImportOrder` check due to incorrect handling of static imports in the `config:example` block.

### What was changed:
- Updated the test input file `InputImportOrderBug.java` under `src/test/resources` to reproduce the issue.
- Ran `mvn verify` successfully with `jacoco.skip=true` after updating the test.
- Confirmed that the failure is no longer occurring with the corrected configuration.

### Notes:
- All CI checks pass locally.
- Requesting review. Thanks!

### DIFF
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/importorder/InputImportOrderBug.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/importorder/InputImportOrderBug.java
@@ -1,7 +1,12 @@
 /*
-No config
-
-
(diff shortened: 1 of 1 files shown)


romani:
Referenced issue in commit message https://github.com/checkstyle/checkstyle/issues/14631 is not related to fix it I confused what is going on.

Can you fix link to issue? I try to understand what is problem you try fix.

You most likely need to change test method that uses such Input also.

Issue #10834, opened 2021-09-28
WhitespaceAround should enforce whitespace around empty blocks


pbludov:
I have read check documentation: https://checkstyle.org/config_whitespace.html#WhitespaceAround

WhitespaceAround should enforce whitespace before left curly(`{`) of empty block.

https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

one whitespace is required:
> Before any open curly brace ({), with two exceptions:
    @SomeAnnotation({a, b}) (no space is used)
    String[][] x = {{"foo"}}; (no space is required between {{, by item 9 below)

Test:
```java
/** some javadoc. */
public class Test {
  Test ctor = new Test(){};                               // Expected violation

  void method(){}                                         // Expected violation

  class Class{}                                           // Expected violation

  interface Interface{}                                   // Expected violation

  @interface Annotation{}                                 // Expected violation

  enum Enum{}                                             // Expected violation

  record Record(){}                                       // Expected violation

  private static void main(String... args) {
    boolean b = System.currentTimeMillis() < 0;
    while (b){}                                           // Expected violation
    for (int i = 1; i > 1; i++){}                         // Expected violation
  }

  static{}                                                // Expected violation

}
```

CLI:
``` 
$ java -jar checkstyle-10.26.1-all.jar -c google_checks.xml WhitespaceAroundEmpty.java 
Starting audit...
Audit done.
```

Formatter:
```
$ java -jar google-java-format-1.27.0-all-deps.jar Test.java > FormattedCode.java 
$ diff -Naru Test.java FormattedCode.java 
--- Test.java   2025-07-18 16:59:16.666695600 +0530
+++ FormattedCode.java  2025-07-18 16:59:29.913097700 +0530
@@ -1,25 +1,25 @@
 /** some javadoc. */
 public class Test {
-  Test ctor = new Test(){};
+  Test ctor = new Test() {};
 
-  void method(){}
+  void method() {}
 
-  class Class{}
+  class Class {}
 
-  interface Interface{}
+  interface Interface {}
 
-  @interface Annotation{}
+  @interface Annotation {}
 
-  enum Enum{}
+  enum Enum {}
 
-  record Record(){}
+  record Record() {}
 
   private static void main(String... args) {
     boolean b = System.currentTimeMillis() < 0;
-    while (b){}
-    for (int i = 1; i > 1; i++){}
+    while (b) {}
+    for (int i = 1; i > 1; i++) {}
   }
 
[...]

PR #17419, opened 2025-07-18
supplemental: IT for 4.6.2 whitespace before '{' of empty blocks

mohitsatr:
part of #10834


romani (approved):
good step forward with definition of problem 
https://www.diffchecker.com/y14f4jJl/

Issue #17301, opened 2025-06-30
Add @Serial annotation to serialVersionUID fields


AmitKumarDeoghoria:
From comment : https://github.com/checkstyle/checkstyle/pull/17259#issuecomment-3019192404.               During the JDK 17 upgrade(https://github.com/checkstyle/checkstyle/issues/17168), IntelliJ inspections have identified 30 instances of serialVersionUID fields that should be annotated with @Serial. This annotation was introduced in JDK 14 ([JEP 359](https://openjdk.org/jeps/359)) and finalized in JDK 17. Since this is a JDK 17-specific feature, we'll address this as part of our post-migration cleanup.

### Add the @Serial annotation to all flagged fields:
`+ import java.io.Serial;

public class Example {
+   @Serial
    private static final long serialVersionUID = 1L;
}`

List of Affected Files/Fields
1. `src/main/java/.../ThreadModeSettings.java` (serialVersionUID)
2. `src/main/java/.../DefaultConfiguration.java` (serialVersionUID)
3. `src/main/java/.../CheckstyleException.java` (serialVersionUID)
4. `src/main/java/.../TreeTableCellRenderer.java` (serialVersionUID)
5. `src/main/java/.../MainFrame.java` (4 inner class fields)
6. `src/main/java/.../ListToTreeSelectionModelWrapper.java` (serialVersionUID)
7. `src/main/java/.../TreeTableModelAdapter.java` (serialVersionUID)
8. `src/main/java/.../TreeTable.java` (2 fields)
9. `src/main/java/.../MetadataGenerationException.java` (serialVersionUID)
10. `src/main/java/.../UniquePropertiesCheck.java` (inner class field)
11. `src/main/java/.../OrderedPropertiesCheck.java` (inner class field)
12. `src/test/java/.../XMLLoggerTest.java` (inner class field)
13. `src/test/java/.../SarifLoggerTest.java` (inner class field)
14. `src/test/java/.../MainTest.java` (2 anonymous class fields)
15. `src/test/java/.../CheckerTest.java` (6 inner class fields)
16. `src/test/java/.../RegexpOnFilenameCheckTest.java` (inner class field)
17. `src/test/java/.../CheckstyleAntTaskStub.java` (inner class field)

romani:
We will make decision later on to follow this inspection or not.

romani:
fix is merged

PR #17433, opened 2025-07-20
Issue #17301: Add @Serial annotation to serialVersionUID fields

AmitKumarDeoghoria:
part of : #17301


src/test/java/com/puppycrawl/tools/checkstyle/CheckerTest.java
```diff
@@ -1148,8 +1150,8 @@ public void testCatchErrorWithCache() throws Exception {
         final Error expectedError = new IOError(new InternalError(errorMessage));
 
         final File mock = new File("testFile") {
+            @Serial
             private static final long serialVersionUID = 1L;
-
```
AmitKumarDeoghoria, line 1152: Did this because Inner class max size (20)

romani (approved):
thanks a lot

PR #17443, opened 2025-07-21
Issue #13999: Resolve pitest suppression for JavadocTagInfo$14 and JavadocTagInfo$11

Praveen7294:
part of #13999

### DIFF
--- a/config/pitest-suppressions/pitest-javadoc-suppressions.xml
+++ b/config/pitest-suppressions/pitest-javadoc-suppressions.xml
@@ -11,15 +11,6 @@
 
   <mutation unstable="false">
     <sourceFile>JavadocTagInfo.java</sourceFile>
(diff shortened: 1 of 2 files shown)


romani:
All CI is doing , you can run on your local.
In this case `./.ci/pitest.sh "pitest-javadoc"`

Praveen7294:
@romani  I saw multiple PRs related to this issue. You asked them not to use pure unit testing, but in my opinion, we should use pure unit testing here. If we use an input file to kill the mutation, we would have to write an inline configuration. However, I couldn't find any specific check that uses the `SERIAL_DATA.isValidOn()` method to trigger this method.

romani:
@Praveen7294, please choose some other pitest suppression that will not be exceptional.
To practice concept that pure unit test is easy to write but it doesn't contribute to quality of behavior, user need quality of behavior not a tests of code.

Praveen7294:
ok sir

Issue #17378, opened 2025-07-13
SiteUtil.getPropertiesJavadocs does not reflect all the constituents of what it does


SteLeo1602:
Besides just of properties, in SiteUtil, `getPropertiesJavadocs` also gets the DetailNode of entire javadoc of module. 

We could correspondingly rename that method and change its javadoc description. Furthermore, splitting that method into two could be considered.

https://github.com/checkstyle/checkstyle/blob/77aafaebf7e368c966b2f3e956a584aa204644c5/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java#L440-L475

PR #17498, opened 2025-07-29
Issue #17378: Introduced JavadocScraperResultUtil to split getModuleAndPropertiesJavadocs into two methods

SteLeo1602:
Issue #17378 

New class `JavadocScraperResultUtil` is being added, consisting of static data and their corresponding methods of `ClassAndPropertiesSettersJavadocScraper`, where they were extracted from. The reason behind this is that `ClassAndPropertiesSettersJavadocScraper` is a bypass, that works as a check, while is not being a check, functioning mainly as an instance. Thus, static data has a hard time conforming to that class, resulting in various inspection errors regarding the use of static. Making a separate class `JavadocScraperResultUtil` for static purposes that stores the results of `ClassAndPropertiesSettersJavadocScraper` makes it cleaner and smoother to operate.


romani:
all around yelling on static initialization.
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/35471/workflows/3db3c208-6483-44b6-9007-5a7001ba8cc5/jobs/1018058/artifacts

we can suppress or we have ability to make following workaround:
We can remove all static fields from ClassAndPropertiesSettersJavadocScraper and we create new class ClassAndPropertiesSettersJavadocScraperResult, it has two static fields and it has get/set methods for such fields.
getter is public method, settler has no modifier (protected-private visibility) so only classes in same package can call setXXXX (ClassAndPropertiesSettersJavadocScraper and  ClassAndPropertiesSettersJavadocScraperResult should be in same package).
All who needs results of scrapper should use ClassAndPropertiesSettersJavadocScraperResult.getXxxxxxx .

Workaround above is still not ideal, but a way better than bunch of supressions and dealing with object to set value in static fields of class.

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraperResult.java
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Map of scraped properties javadocs - name of property, javadoc detail node.
+     */
+    private static final Map<String, DetailNode> JAVADOC_FOR_PROPERTIES =
+            new LinkedHashMap<>();
+
+    /**
+     * The class javadoc.
+     */
+    private static DetailNode javadocForModule;
```
romani: `moduleJavadocNode`

And rename respectively get/set

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraperResult.java
```diff
@@ -0,0 +1,93 @@
...
+import com.puppycrawl.tools.checkstyle.api.DetailNode;
+
+/**
+ * Class with result data of ClassAndPropertiesSettersJavadocScraper.
+ */
+public final class ClassAndPropertiesSettersJavadocScraperResult {
+    /**
+     * Map of scraped properties javadocs - name of property, javadoc detail node.
+     */
+    private static final Map<String, DetailNode> JAVADOC_FOR_PROPERTIES =
```
romani: PROPERTIES_JAVADOC_NODES

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraperResult.java
```diff
@@ -0,0 +1,93 @@
...
+        javadocForModule = moduleJavadoc;
+    }
+
+    /**
+     * Sets additional property javadoc to property map.
+     *
+     * @param propertyName name of property.
+     * @param propertyJavadoc property's javadoc.
+     */
+    static void setJavadocProperty(String propertyName, DetailNode propertyJavadoc) {
```
romani: putPropertyJavadocNode

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraperResult.java
```diff
@@ -0,0 +1,93 @@
...
+        JAVADOC_FOR_PROPERTIES.clear();
+        javadocForModule = null;
+    }
+
+    /**
+     * Get the properties javadocs map.
+     *
+     * @return the javadocs.
+     */
+    public static Map<String, DetailNode> getJavadocsForProperties() {
```
romani: getPropertyJavadocNode

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraperResult.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import com.puppycrawl.tools.checkstyle.api.DetailNode;
+
+/**
+ * Class with result data of ClassAndPropertiesSettersJavadocScraper.
+ */
+public final class ClassAndPropertiesSettersJavadocScraperResult {
```
romani: JavadocScraperResultUtil

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraperResult.java
SteLeo1602: Done

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraperResult.java
SteLeo1602: Done

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraperResult.java
SteLeo1602: Done

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraperResult.java
SteLeo1602: Done

src/main/java/com/puppycrawl/tools/checkstyle/site/ClassAndPropertiesSettersJavadocScraperResult.java
SteLeo1602: Done

SteLeo1602:
`mvn clean verify` runs green on local

src/main/java/com/puppycrawl/tools/checkstyle/site/JavadocScraperResultUtil.java
Zopsss, line 1: instead of creating a new separate class, can't we just put all of this into SiteUtil.java? @romani @SteLeo1602

romani: SiteUtil is good until class.
ResultUtil is still hack to let Check store results in some class and not have static fields.

Zopsss: > ResultUtil is still hack to let Check store results in some class and not have static fields.

Sorry I didn't understood this

Zopsss: I don't see this new class being used anywhere but SiteUtil.java

romani: It is used in ClassAndPropertiesSettersJavadocScraper to store results

romani (approved):
This is ok to merge.
Scrapper module becomes a bit better, hack on its results moved to separate class.
In other case there are numerous violations from tools that do not like facts that instance of class changes static variables. So all results moved to separate class to avoid violations.

Zopsss:
Commit message is: 

> Splitted SiteUtil.getModuleAndPropertiesJavadocs into two methods

but we're introducing a whole new Class in the PR. The commit message needs to be improved and PR description should also explain all the changes. The referenced issue doesn't mention anything about introducing a new class, so we should mention the reason behind it in PR description.

SteLeo1602:
> Commit message is:
> 
> > Splitted SiteUtil.getModuleAndPropertiesJavadocs into two methods
> 
> but we're introducing a whole new Class in the PR. The commit message needs to be improved and PR description should also explain all the changes. The referenced issue doesn't mention anything about introducing a new class, so we should mention the reason behind it in PR description.

Done

Zopsss (approved):

PR #17490, opened 2025-07-28
Issue #17487: Add `rewrite` support for `errorprone.refasterrules`

Pankraz76:
https://github.com/checkstyle/checkstyle/issues/17487

## Add `rewrite` support for `errorprone.refasterrules`

I’d like to propose integrating [Google’s Error Prone](https://github.com/google/error-prone) and its Picnic extension (demonstrated in [Automating Away Bugs with Error Prone | PlatformCon 2023](https://www.youtube.com/watch?v=6llnrUtVlrE)) to enable automated bug fixes via `rewrite` rules. This would complement Checkstyle’s static analysis capabilities by addressing semantic bugs rather than stylistic issues.

## Motivation
Error Prone’s `refaster`/`rewrite` rules can automatically:
- Fix common bug patterns (e.g., `String.equals()` misuse)
- Modernize code (e.g., JDK migration helpers)
- Enforce best practices (e.g., null-check improvements)

Real-world adoptions show tangible benefits:
- [Apache Kafka’s integration](https://github.com/apache/kafka/pull/20219)
- [Spotless’s implementation](https://github.com/diffplug/spotless/pull/2576)

## Proposal
1. Add support for `errorprone.refasterrules`-based rewrites
2. Implement with opt-in adoption (no breaking changes)
3. Include suppression mechanisms for API constraints

## Discussion Points
- Need consensus on:
  - Scope of auto-fixes
  - Preferred suppression strategy
  - Integration approach

## Next Steps
I’m happy to:
- Prepare a PoC demonstrating the value
- Collaborate on implementation strategy
- Address any concerns about compatibility

PR #17500, opened 2025-07-29
Pull #17500: Align azure image on `ubuntu-24.04`

Pankraz76:
Pull #17500: Align azure image on `ubuntu-24.04`

refs:

- https://github.com/checkstyle/checkstyle/pull/17490

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -92,7 +92,7 @@ strategy:
 
     # MacOS JDK17 verify
     'MacOS JDK17 verify':
-      image: 'macOS-14'
+      image: 'ubuntu-24.04'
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -92,7 +92,7 @@ strategy:
 
     # MacOS JDK17 verify
     'MacOS JDK17 verify':
-      image: 'macOS-14'
+      image: 'ubuntu-24.04'
```
romani, line 95: It is intentionally macos .
Even comment above double state on it.

All we can do here is migrate to new macos instance

Pankraz76:
trivial. might improve comment. sorry for bothering.

Issue #17031, opened 2025-05-06
Fix compareTo method for testInputViolations


ZiadAbdElFatah:
Found here: https://github.com/checkstyle/checkstyle/pull/17008#discussion_r2074840163

When there are two violations on the same line the sort doesn't then sort according to column number, so we need to add `colNo` to the violations to be able to make the `compareTo` method sort with column number when the line number are equal

Goal is to have  violation messages for all lines at src/site/xdoc/filters/suppresswithnearbycommentfilter.xml

Wanyi9988:
Hi ZiadAbdElFatah,

I am trying to reproduce and fix the issue you described regarding sorting of violations on the same line. I specifically tested whether errors are sorted by column number when the line numbers are equal.

Here is the Java test code I used:
```Java
package test;

public class TestSuppress {
    public static void main(String[] args) {
        int a=1; int b= 2;int c=3;// multiple statements on one line, with spacing issues around '='
        if(a==b){
            System.out.println("a equals b");
        }else if (a !=c ){
            System.out.println("a not equals c");
        }
    }

    public void sampleMethod(){
        int x=10;int y =20; // multiple statements on one line, spacing issues
        int z= x+y;
        System.out.println("Sum:"+z);
    }
}
```

I ran the following command:
`java -jar checkstyle-10.17.0-all.jar -c config/checkstyle_checks.xml TestSuppress.java`

The output is:
```
[ERROR] /Users/wanyi/Desktop/Wanyi/dev/open-source/test/test/TestSuppress.java:5:14: '=' 前应有空格。 [WhitespaceAround]

[ERROR] /Users/wanyi/Desktop/Wanyi/dev/open-source/test/test/TestSuppress.java:5:14: '=' 后应有空格。 [WhitespaceAround]

[ERROR] /Users/wanyi/Desktop/Wanyi/dev/open-source/test/test/TestSuppress.java:5:23: '=' 前应有空格。 [WhitespaceAround]

[ERROR] /Users/wanyi/Desktop/Wanyi/dev/open-source/test/test/TestSuppress.java:5:26: 禁止一行有多句代码。 [OneStatementPerLine]

[ERROR] /Users/wanyi/Desktop/Wanyi/dev/open-source/test/test/TestSuppress.java:5:32: '=' 前应有空格。 [WhitespaceAround]

[...]

PR #17512, opened 2025-07-30
Issue #17031: Fix compareTo method for TestInputViolation.java

Praveen7294:
Issue #17031

previous violations or filtered violations comment when lines are equal:
`// violation 'some message'` or `// filtered violation 'some message'`

now
`// violation 'num: some message'` or `// filtered violation 'num: some message'` where `num` refers to the violation that comes first, denoted by the smallest integer. We can also use the column number instead. However, it doesn't matter what we use — the `num` is only added to help with sorting, first by line number, then by message.

Previously, the `compareTo()` method sorted in reverse chronological order when the line numbers were equal. Now, it sorts in chronological order.


Praveen7294:
@romani  can you help me why ci failure?

PR #17513, opened 2025-07-30
Issue #17031: Fix compareTo method for TestInputViolation.java

Praveen7294:
Issue https://github.com/checkstyle/checkstyle/issues/17031

previous violations or filtered violations comment when lines are equal:
`// violation 'some message'` or `// filtered violation 'some message'`

now
`// violation 'num: some message'` or `// filtered violation 'num: some message'` where `num` refers to the violation that comes first, denoted by the smallest integer. We can also use the column number instead. However, it doesn't matter what we use — the `num` is only added to help with sorting, first by line number, then by message.

Previously, the `compareTo()` method sorted in reverse chronological order when the line numbers were equal. Now, it sorts in chronological order.

### DIFF
--- a/src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
+++ b/src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
@@ -253,8 +253,9 @@ public class Example5 {
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
 public class Example6 {
   // @cs.suppress [ConstantName|NoWhitespaceAfter] A comment here
-  public static final int [] array = {}; // filtered violation
-  // filtered violation above
+  public static final int [] array = {};
+  // filtered violation above ''int' is followed by whitespace'
+  // filtered violation 2 lines above 'Name 'array' must match pattern'
 }
 </code></pre></div>
         <p>
--- a/src/site/xdoc/filters/suppresswithplaintextcommentfilter.xml
+++ b/src/site/xdoc/filters/suppresswithplaintextcommentfilter.xml
@@ -466,9 +466,9 @@ public class Example9 {
           ST001,Station 001,ZONE1,Zone 1,CP1,Competitor 1,123 Street,Unit 2,Houston,TX,77033,US,29.761496813335178,-95.53049214204984
           ST002,Station 002,ZONE2,,CP2,,668 Street,Unit 23,San Jose,CA,95191,US,37.35102477242508,-121.9209934020318
           """;
-  // filtered violation 5 lines above 'Line is longer than 100 characters (found 147).'
-  // filtered violation 5 lines above 'Line is longer than 100 characters (found 133).'
-  // filtered violation 5 lines above 'Line is longer than 100 characters (found 116).'
+  // filtered violation 4 lines above 'Line is longer than 100 characters (found 147).'
+  // filtered violation 4 lines above 'Line is longer than 100 characters (found 133).'
+  // filtered violation 4 lines above 'Line is longer than 100 characters (found 116).'
 
   // violation below, 'Line is longer than 100 characters (found 183).'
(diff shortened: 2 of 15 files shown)


src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
```diff
@@ -253,8 +253,9 @@ public class Example5 {
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
 public class Example6 {
   // @cs.suppress [ConstantName|NoWhitespaceAfter] A comment here
-  public static final int [] array = {}; // filtered violation
-  // filtered violation above
+  public static final int [] array = {};
+  // filtered violation above '1: followed by whitespace'
+  // filtered violation 3 lines above '2: must match pattern'
```
romani: `filtered violation 3 lines above` this looks weird ....as it points to `// @cs.suppress [ConstantName|NoWhitespaceAfter] A comment here`

`// filtered violation above '1: followed by whitespace'` this also looks weird, not clear what is "1:".

Can we archive:
```
  public static final int [] array = {};
  // filtered violation above ''int' is followed by whitespace'
  // filtered violation 2 lines above 'Name ''array'' must match pattern'
```

romani (changes requested):
we might need to abandon original idea on line number, and focus on matching messages. Order or messages for now is predefined as order of Checks is predefined. 

items:

Praveen7294:
> we might need to abandon original idea on line number, and focus on matching messages. Order or messages for now is predefined as order of Checks is predefined.

you are right but matching message fails in some cases like `34: Line matches the illegal pattern '.*[a-zA-Z][0-9].*'.` and `34:1: Line contains a tab character.`, For this reason i added `1:` for violation comes first, `2:` for violation comes second.

src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
Praveen7294: > `filtered violation 3 lines above` this looks weird ....as it points to `// @cs.suppress [ConstantName|NoWhitespaceAfter] A comment here`

Done.

romani: please use messages as I suggested, it helps understand who and what placing violation

romani:
> you are right but matching message fails in some cases like 34: Line matches the illegal pattern '.*[a-zA-Z][0-9].*'. and 34:1: Line contains a tab character.

if line and column if same, we can use alphabetical order.
artificial numbers still looks weird

Praveen7294:
> if line and column if same, we can use alphabetical order. artificial numbers still looks weird

i understand but what i want to say we are using two methods to verify violation, `verifyViolation()` and `verify()` in `verifyFilterWithInlineConfigParser()`.

- `verifyViolation()` matches first line number then message in alphabetical order, here order is matter.
- `verify()` matches actual violation with expected violation provided in test method where Order does not matter, but the contents must match exactly.

so, `verifyViolation()` uses only line number and message whereas `verify()` use line number, column no as well as message also.

In `verifyFilterWithInlineConfigParser()`, we are using `Collections.sort()` for arranging the violations in chronological order. these violations are passed in `verifyViolation()` method.

### Example
when two violation occurs in same line. Required order is:
`Line matches the illegal pattern '.*[a-zA-Z][0-9].*'.`
`Line contains a tab character.`

But after sorting, order look like this:
`Line contains a tab character.`
`Line matches the illegal pattern '.*[a-zA-Z][0-9].*'.`

For this reason i have to use artificial number. so, that order must be preserved.

What else i can do, i can use column number instead of artificial number. ex - `Filtered violation above '1: Line matches the illegal pattern '.*[a-zA-Z][0-9].*'.'` or `30: must match pattern`.

src/site/xdoc/filters/suppresswithnearbycommentfilter.xml
Praveen7294: Done, please review

Praveen7294:
@romani please review i fixed according to your suggestion.

romani (approved):
this is simply amazing , very clean and very clear

PR #17524, opened 2025-08-01
Issue #16361: Update DefaultLoggerTest.java to verify with InlineConfig parser and logger, more updates

ElinaZoldnere:
#16361

Few more tests updated.


ElinaZoldnere:
Regarding the logger test utilities. There are already about 150 LOC specifically related to loggers in `AbstractModuleTestSupport`. Would it make sense to move them into a dedicated extension like `AbstractLoggerTestSupport`?

romani (approved):
That a lot

PR #17526, opened 2025-08-02
Issue #13038: handled inner class method definition in VariableDeclar…

smita1078:
#13038 

Extends #16026 
WIP
This PR updates `VariableDeclarationUsageDistanceCheck` to properly handle inner scopes such as:

- Inner classes declared inside methods
- Lambdas
- Anonymous class overrides
- Added multiple Tests

When `validateBetweenScopes = true`, the check now treats these nested declarations as **independent scopes**, and does **not flag** variables whose usages lie fully within those inner scopes.

This prevents false positives when a variable is declared before an inner class or lambda but only used inside it — aligning the behavior with expectations documented in issue #13038.

### DIFF
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral3.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral3.java
@@ -88,8 +88,8 @@ public class InputVariableDeclarationUsageDistanceGeneral3 {
     }
 
     public void method7() {
-        // until https://github.com/checkstyle/checkstyle/issues/13011, the below distance is 3
-        Integer t = 5;
+
+        Integer t = 5; // violation 'Distance .* is 3.'
         nothing();
         System.out.println();
         class BClass extends Parent {
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/VariableDeclarationUsageDistanceCheck.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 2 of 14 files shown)


smita1078:
Hi  @romani ,

I've tried to fix issue #13038.

The idea I followed is: when `validateBetweenScopes = true`, the check should **not count distance across nested scopes** like inner classes, lambdas, or anonymous classes declared inside methods. So, variables declared in a method and used only inside such nested scopes won't trigger a violation.

I've added/updated test cases for those scenarios ,but I'm not fully sure if this direction is correct.
Please review and let me know if the approach makes sense ?

romani:
https://checkstyle.org/checks/coding/variabledeclarationusagedistance.html#Example5-code

`<property name="validateBetweenScopes" value="true"/>`

is supposed to be used for code like:
```
  public void foo2() {
    int a;          // ok, distance = 2
    int b;          // ok, distance = 3
    // violation below, 'variable 'count' declaration and its first usage is 4.'
    int count = 0;

    {   // this is inner scope start
      System.out.println("Inside inner scope");
      a = 1;
      b = 2;
      count++;
    } // this is inner scope end
  }
```

where `{ }` is used not for method/classes bodies.

it should not be applicable for classes/methods/if/........ , it is should be applicable for only for curly braces that stand-alone. Compiler allows it and user sometime like  such structure.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceGeneral3.java
```diff
@@ -25,7 +25,7 @@ <T> void xx(List<T> m){}
     }
 
     public void method2() {
-        Integer a = 5;
+        Integer a = 5; // violation 'Distance .* is 2.'
```
romani: distance should be 1, as `class AClass extends Parent {` is class start not a scope, variable cannot be moved to any better place.
So no violation should be here

romani (changes requested):
itms:

romani:
@smita1078 , is there a chance to finish this PR ?

smita1078:
> @smita1078 , is there a chance to finish this PR ?

Hi @romani ,
I did explore multiple logic updates and test scenarios to ensure correct handling of validateBetweenScopes=true, but unfortunately, I’m still unable to fully figure out the correct handling for all edge cases. At this point, I’m not able to make further progress on this PR.

romani:
It is completely ok to abandon PR. 
Not all is easy and sometimes need second iterations or some break from task to find solution.

We would deeply appreciate if you can help us with some other issues.

smita1078:
> It is completely ok to abandon PR. Not all is easy and sometimes need second iterations or some break from task to find solution.
> 
> We would deeply appreciate if you can help us with some other issues.

Thank you so much for understanding.
I’ll look into some other issues and continue contributing where I can.

PR #17527, opened 2025-08-02
minor: Update checkstyle-non-main-files-suppressions.xml to remove ol…

romani:
…d suppressions


romani:
main branch was removed in Github UI.

we use master

Issue #17531, opened 2025-08-03
site log has a lot of warnings 'Anchor name "Description" used more than once'


romani:
cmd to reproduce: `mvn clean site -Pno-validations`

in logs there are huge amount of WARNINGS like: Anchor name "Description" used more than once,  Anchor name "Properties" used more than once

Brijeshthummar02:
@romani we need to minimize all that warnings one by one right?

romani:
we can do this in groups.

All at ones is possible, but might be problematic to pass CI.

Brijeshthummar02:
> we can do this in groups.
> 
> All at ones is possible, but might be problematic to pass CI.

will try to do it in batch.

romani:
Final fix is merged

PR #17535, opened 2025-08-03
Issue #17531: site log has a lot of warnings 'Anchor name "Description" used more than once'

Brijeshthummar02:
Issue #17531

### DIFF
--- a/src/site/xdoc/checks/blocks/avoidnestedblocks.xml
+++ b/src/site/xdoc/checks/blocks/avoidnestedblocks.xml
@@ -8,7 +8,7 @@
   <body>
     <section name="AvoidNestedBlocks">
       <p>Since Checkstyle 3.1</p>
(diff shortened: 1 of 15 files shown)


Brijeshthummar02:
@romani if we try to solve it like this, the one we did in this PR. Ci will gonna fail. I tried same on local multiple times. But results were same. Is there any other way we can complete this?

SteLeo1602:
> @romani if we try to solve it like this, the one we did in this PR. Ci will gonna fail. I tried same on local multiple times. But results were same. Is there any other way we can complete this?

I believe you should work on editing `xml.template` files instead as they are the ones that put text to regular `xml` files. The `testAllSubSections()` in `XdocsPagesTest` would complain, but we could change its functionality

romani:
fixed in https://github.com/checkstyle/checkstyle/pull/17607

PR #17165, opened 2025-06-03
update to latest version google-java-style/javaguide.html

mohitsatr:
#17162 

Update the content of existing [javaguide.html](https://github.com/checkstyle/checkstyle/blob/master/src/site/resources/styleguides/google-java-style-20220203/javaguide.html) to latest content. The latest content is taken from: https://github.com/google/styleguide/blob/gh-pages/javaguide.html

example of diff in 2024 year: https://github.com/checkstyle/checkstyle/pull/14901

PR #17550, opened 2025-08-05
supplemental: IT for '4.8.7 Modifiers'

mohitsatr:
from: https://github.com/checkstyle/checkstyle/pull/17165#discussion_r2247795799

### DIFF
--- a/config/google-java-format/excluded/compilable-input-paths.txt
+++ b/config/google-java-format/excluded/compilable-input-paths.txt
@@ -63,6 +63,8 @@ src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4861blockcomm
 src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4861blockcommentstyle/InputCommentsIndentationInSwitchBlock.java
 src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4861blockcommentstyle/InputCommentsIndentationSurroundingCode.java
 src/it/resources/com/google/checkstyle/test/chapter4formatting/rule487modifiers/InputModifierOrder.java
+src/it/resources/com/google/checkstyle/test/chapter4formatting/rule487modifiers/InputModifierOrderSealed.java
+src/it/resources/com/google/checkstyle/test/chapter4formatting/rule487modifiers/InputModifierOrderNonSealed.java
 src/it/resources/com/google/checkstyle/test/chapter4formatting/rule489textblocks/InputTextBlocksGeneralForm.java
 src/it/resources/com/google/checkstyle/test/chapter4formatting/rule489textblocks/InputTextBlocksIndentation.java
 src/it/resources/com/google/checkstyle/test/chapter5naming/rule522classnames/InputClassNames.java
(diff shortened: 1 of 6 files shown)


romani:
`./.ci/google-java-format.sh .ci-temp/google-java-format-1.28.0-all-deps.jar`


` Error: Difference for InputModifierOrderSealed.java is more than 50 bytes`

find what group is appropriate for you https://github.com/checkstyle/checkstyle/tree/7fcfb867cf8f4c510f3831d542fb8ac300166559/config/google-java-format/suppressions

classes are very different - https://www.diffchecker.com/k501jr56/ , due to conflict of class names, as they are in same package. 

why we need `implements`/`extends` in such Inputs ? it is not target of chapter. 
Let remove them, it should help to reduce diff.

If we move all types inside public top level class, diff can be reduced even more. Please do. we can sacrifice showing example on how it work on top level class, due to nature of our tests.

src/it/resources/com/google/checkstyle/test/chapter4formatting/rule487modifiers/InputModifierOrderNonSealed.java
Zopsss, line 1: there are only 2 examples of non-sealed class/interface. Please add more. And there are many other extra cases which we may not need, please remove them.

Zopsss: @mohitsatr adding more cases is not our priority so let's forward with only 2 cases that we have. But please remove extra ones. We will add more cases in future PRs

romani:
>If we move all types inside public top level class, diff can be reduced even more. Please do. we can sacrifice showing example on how it work on top level class, due to nature of our tests.

@mohitsatr , please share details why this approach is not working

mohitsatr:
@romani done.


> please share details why this approach is not working


It worked. I did not try it earlier. I thought adding the file to suppression list saves you the trouble of reducing the diff size :face_exhaling:

config/google-java-format/excluded/compilable-input-paths.txt
```diff
@@ -63,6 +63,8 @@ src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4861blockcomm
 src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4861blockcommentstyle/InputCommentsIndentationInSwitchBlock.java
 src/it/resources/com/google/checkstyle/test/chapter4formatting/rule4861blockcommentstyle/InputCommentsIndentationSurroundingCode.java
 src/it/resources/com/google/checkstyle/test/chapter4formatting/rule487modifiers/InputModifierOrder.java
+src/it/resources/com/google/checkstyle/test/chapter4formatting/rule487modifiers/InputModifierOrderSealed.java
+src/it/resources/com/google/checkstyle/test/chapter4formatting/rule487modifiers/InputModifierOrderNonSealed.java
```
romani, line 67: why excluded?
you have Formatted for them

romani (changes requested):
item:

config/google-java-format/excluded/compilable-input-paths.txt
romani: to not format them.

romani (approved):
good.

Issue #14872, opened 2024-05-12
UnnecessaryParentheses: no violation if casting is present before expression


romani:
detected at https://github.com/checkstyle/checkstyle/pull/14856/files#r1597531069

I have read check documentation: https://checkstyle.org/checks/coding/unnecessaryparentheses.html#UnnecessaryParentheses
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <module name="TreeWalker">
    <module name="UnnecessaryParentheses">
    </module>
  </module>
</module>

$ cat Test.java 
import java.util.Arrays;

public class Test
{
  public static void main(String[] args) {
    String[] a = { "s", "", "1", "", "" };
    Arrays.stream(a).filter(s -> !((boolean)s.isEmpty())).toArray(String[]::new); // should be violation
    Arrays.stream(a).filter(s -> !(boolean)s.isEmpty()).toArray(String[]::new);
  }
}

$ java -jar checkstyle-10.16.0-all.jar -c config.xml Test.java
Starting audit...
Audit done.
```

Expected: violation on  line 7

--------

https://introcs.cs.princeton.edu/java/11precedence/

>Most programmers do not memorize them all, and, even those that do, use parentheses for clarity.

higher precedence executes first.

|Level|Operator                              |Description                                                                                       |Associativity|
|-----|--------------------------------------|--------------------------------------------------------------------------------------------------|-------------|
|16   |() [] new . ::                        |parentheses, array access,  object creation, member access, method reference                           |left-to-right|
|15   |=++ --                                |unary post-increment, unary post-decrement                                                         |left-to-right|
|14   |=+ - ! ~ ++ --                        |unary plus, unary minus, unary logical NOT, unary bitwise NOT, unary pre-increment, unary pre-decrement|right-to-left|
|13   |()                                    |cast                                                                                              |right-to-left|
|12   |* / %                                 |multiplicative                                                                                    |left-to-right|
|11   |=+ - +                                |additive, string concatenation                                                                     |left-to-right|
|10   |<< >> >>>                             |shift                                                                                             |left-to-right|
[...]

PR #17575, opened 2025-08-06
Issue #17554: Report UnneccessaryParentheses for dot, 'new' and array access

### LINKED ISSUE #17554: UnnecessaryParentheses not reported for array access, field access and method calls
I have read check documentation: https://checkstyle.org/checks/coding/unnecessaryparentheses.html#UnnecessaryParentheses

```bash
/var/tmp $ javac YOUR_FILE.java
#[[MAKE SURE THERE IS SUCCESSFUL COMPILATION]]

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">

<module name = "Checker">
    <module name="TreeWalker">
        <module name="UnnecessaryParentheses">
        </module>
    </module>
</module>


/var/tmp $ cat YOUR_FILE.java
public class Main {

    private static int foo;
    private static Main INSTANCE;
    private String name;

        foo = (args[0]).length();
        foo = (String.valueOf(foo)).length();
        foo = (INSTANCE.name).length();
        foo = (new String()).length();
        System.out.println(foo);
    }

}


/var/tmp $ java $RUN_LOCALE -jar checkstyle-X.XX-all.jar -c config.xml YOUR_FILE.java
Starting audit...
Audit done.
```

---

All 4 assignments to `foo` should trigger `UnnecessaryParentheses`

zbynek:
Fixes #17554 

new default tokens:
TokenTypes.INDEX_OP,
TokenTypes.DOT 

new acceptable but not default:
TokenTypes.LITERAL_NEW,

https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#DOT
https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#INDEX_OP
https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_NEW

not clear why https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LOR popup in diff, but it is default already , and it is not related issue in  validation of default.

---
Diff Regression config: https://gist.githubusercontent.com/zbynek/8f73c96579efbe231a60df2af09afae9/raw/c903e5b948b4767c8d43a6cb4b78065f704caa38/base-config.xml
Diff Regression patch config: https://gist.githubusercontent.com/zbynek/8f73c96579efbe231a60df2af09afae9/raw/d68058e46543503f9670022419d24f58588f0b50/patch-config.xml


romani:
To avoid CI failure 
https://checkstyle.semaphoreci.com/jobs/7843909d-eb87-4fc1-bef2-7669b076d6b2#L639

Please send PR to https://github.com/checkstyle/contribution/blob/master/patch-diff-report-tool/src/test/java/com/github/checkstyle/internal/AbstractTest.java

romani:
Github, generate report for UnnecessaryParentheses/all-examples-in-one

romani:
@zbynek , please do spot checking of diff report.

zbynek:
Github, generate report for UnnecessaryParentheses/all-examples-in-one

zbynek:
@romani I added an exception for the first argument of ternary operators and removed `LITERAL_NEW` from the default set of tokens. I think the current state is reasonable, if you think more tweaks are needed please let me know.

romani:
diff between reports.
https://www.diffchecker.com/FMphnMXV/

please add to our Inputs cases that forced you to remove token LITERAL_NEW .
We try to not rely much on diff regression and have all cases in our Inputs to let users run tests and see feeedback.
so next contributor who will try to add token NEW, will get bunch of test failures, a way quicker feedback than  diff report .

romani:
what are your thoughts on https://github.com/checkstyle/checkstyle/issues/14872 , it has table , that we do not have in code and so this Check is very error prone.

zbynek:
@romani I think implementing the rule by using operator precedence would make the code easier to reason about. But marking all parentheses that are not needed by the compiler as unnecessary would add a lot of violations, some of which are contradictory to Google style guide and enforced by some tools, see https://errorprone.info/bugpattern/OperatorPrecedence . So a config property that lets users pick between keeping the code short and ensuring clarity would be needed.

Should we try to implement the more generic solution and abandon this PR?

romani:
@zbynek , please resolve conflict.

no abandoning, we are brainstorming, of any good ideas pop up , we can improve in separate issues/PRs.

> But marking all parentheses that are not needed by the compiler as unnecessary would add a lot of violations

There was rule of thumb from previous century "extra parentheses cost less than mistakes or confusions during reading" , it is still true. So I try to be accurate with this Check, and not push much on people who use such rule of thumb.

romani:
Github, generate report for UnnecessaryParentheses/all-examples-in-one

romani:
@zbynek  , I need your help with diff testing of 

> new acceptable but not default:
TokenTypes.LITERAL_NEW,

unfortunately our predefined configs will not help here and we need 
https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#executing-generation with usage of `Diff Regression patch config:` , examples https://github.com/checkstyle/checkstyle/pulls?q=is%3Apr+is%3Aopen+%22Diff+Regression+patch+config%3A%22
it should clearly show that it not very error prone to activate this token.

romani:
Ok, default behavior changed a lot, we need to be accurate bon this to avoid any false positives.

Please self review, I will review too.

zbynek:
I'm quite confident this doesn't create false positives in the sense of "fixing the violation would change the meaning of the expression", but of course there are cases where someone may say "fixing the violation makes the code harder to read".

One example are the "balancing brackets", i.e. cases where you add a pair of parentheses to one operand of `&&` and then you wrap the other operand in parentheses just for symmetry:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a67a46c21920afd47943c59390a0e022ad026083_2025124610/reports/diff/camunda/xref/home/runner/work/checkstyle/checkstyle/.ci-temp/repositories/camunda/zeebe/test-util/src/main/java/io/camunda/zeebe/test/util/bpmn/random/blocks/BlockSequenceBuilder.java.html#L70

I don't know if we want to add an exception for these?

zbynek:
GitHub, generate report

zbynek:
GitHub, generate report

romani:
Github, generate report for UnnecessaryParentheses/all-examples-in-one

zbynek:
@romani is there anything I can do to help get this merged?

romani:
@zbynek, 
I am really sorry for delay, your PR is still in my to do list. Unfortunately I need significant time slot to think on it, but there are a way more activities happening that needs my attention.

@rdiachenko , @mahfouz72 , can you help to review this proposal? Code is ok, I have some worries on strategy to have more tokens, it might give bad side effect.

romani:
@zbynek  , is your code base (where you plan to use this fix) is on jdk17 or already migrated to jdk21 ?
(it sound like off topic, but it is actually matters)

zbynek:
@romani it's currently on jdk17

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesFieldMethodAccess.java
```diff
@@ -0,0 +1,48 @@
...
+        private String name;
+
+        private boolean check() {
+            return false;
+        }
+    }
+
+    public static void main(String[] args) {
+        foo = (args[0]).length();
+        // violation above 'Unnecessary parentheses around expression.'
```
romani, line 34: please update all to be in this format(correct + incorrect ) :
```
foo = args[0].length();
foo = (args[0]).length();
// violation above 'Unnecessary parentheses around expression.'
```

it will help to spot valitidy of violation easier.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesFieldMethodAccess.java
```diff
@@ -0,0 +1,48 @@
...
+        foo = (String.valueOf(foo)).length();
+        // violation above 'Unnecessary parentheses around expression.'
+        foo = (INSTANCE.name).length();
+        // violation above 'Unnecessary parentheses around expression.'
```
romani, line 40: please add `foo = ((SomeType)INSTANCE.name).length();`
there should be no violations, as parentheses required for cast to be done first.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -522,7 +534,11 @@ else if (TokenUtil.isOfType(type, ASSIGNMENTS)) {
                 assignDepth--;
             }
             else if (isSurrounded(ast) && unnecessaryParenAroundOperators(ast)) {
-                log(ast.getPreviousSibling(), MSG_EXPR);
+                DetailAST inParentheses = ast;
```
romani: does it make  sense to move new logic out of operators ?
https://docs.oracle.com/javase/tutorial/java/nutsandbolts/operators.html

you are targeting `.` 

so you code will be just another block `else if (TokenUtil.isOfType(type, ARRAY_AND_FIELD_ACCESS)) {`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -396,6 +403,8 @@ public int[] getDefaultTokens() {
             TokenTypes.BNOT,
             TokenTypes.POST_INC,
             TokenTypes.POST_DEC,
+            TokenTypes.INDEX_OP,
+            TokenTypes.DOT,
```
romani, line 293: whole issue is about extra parentheses before "." can we skip "INDEX_OP" in default tokens ?
and we can do extra effort to check INDEX_OP after we are called(visitToken) on  DOT.

I know list of tokens is confusing, but what we try to keep them for is for adjustments of targets that use might opt-in.

If all is about DOT , but INDEX_OP and LITERAL_NEW are good for good DOT support. it means tokens should have only DOT and java code in Check search for INDEX_OP and LITERAL_NEW under/around DOT, if DOT is detected.

romani (changes requested):
discussion

romani:
we need to copy all disctinct behavior changes from diff report to Inputs.
We do have big behavior change in diff report, all should be reflected in Inputs (in distinct way)  

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a67a46c21920afd47943c59390a0e022ad026083_2025014733/reports/diff/android-launcher/index.html#A1

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a67a46c21920afd47943c59390a0e022ad026083_2025014733/reports/diff/apache-ant/index.html#A1

especially this:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a67a46c21920afd47943c59390a0e022ad026083_2025014733/reports/diff/apache-ant/index.html#A3

and 
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a67a46c21920afd47943c59390a0e022ad026083_2025014733/reports/diff/apache-ant/index.html#A13

and https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a67a46c21920afd47943c59390a0e022ad026083_2025014733/reports/diff/apache-ant/index.html#A7 

please continue....
We planning very aggressive update, so we need to be careful with validations. 
Parentheses placing is very controversial, some people like them  as they help to be certain, some do not like them as they know [priority rules](https://introcs.cs.princeton.edu/java/11precedence/)

romani:
please add to Inputs code from https://github.com/checkstyle/checkstyle/issues/14872 , I might be wrong but you will cover this issue too.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
zbynek: Sorry, I don't get your point. The list mentions all tokens that should be reported if they have unnecessary brackets areound them, it's not related to what comes after.

`DOT` is among the tokens to show warning for both of these:

```
(a.b).c
(a.b) + ""
```

`INDEX_OP` is among them to show warning for both of these:

```
(a[0]).c
(a[0]) + ""
```

I will add more testcases to make this clear.

romani: Let's make input with `int I = (a[0])` and `(a[0]) + ""` to make it clear that it is separate from DOT completely

romani:
Github, generate report for UnnecessaryParentheses/all-examples-in-one

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesFieldMethodAccess.java
```diff
@@ -0,0 +1,98 @@
...
+        String s;
+        s = m.check() + "";
+        s = (m.check()) + "";
+        // violation above 'Unnecessary parentheses around expression.'
+        s = new Main() + "";
+        s = (new Main()) + "";
+        // violation above 'Unnecessary parentheses around expression.'
+        s = ("" + "").trim();
+        s = (("" + "")).trim();
```
romani: kind of false-negative, do you want to consider it not violation for first version of implmentation ?
and may be in future issue improve further ?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesFieldMethodAccess.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public void checks() {
+        Main m = new Main();
+        if (!m.check()) {
+            throw new RuntimeException("fail");
```
romani: lets simplify it to:
```
if (!m.check()) { return; }
```
to focus on `(` and keep lines with and wihot close to eachother.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesFieldMethodAccess.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        String b1 = ((String) m.toString()).toString();
+        String b2 = ((String) (m.toString())).toString();
+        // violation above 'Unnecessary parentheses around expression.'
+        Class c = m.getClass();
+        if (c == null || !Main.class.isAssignableFrom(c)) {
+            return;
+        }
```
romani: `if (c == null || !Main.class.isAssignableFrom(c)) { return; }`

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesConstructor.java
```diff
@@ -0,0 +1,23 @@
...
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.unnecessaryparentheses;
+
+public class InputUnnecessaryParenthesesConstructor {
+
+    public void main() {
+        // parentheses not needed, but used for extra clarity
+        int length = (new StringBuilder()).length();
```
romani: please code above without `(`

romani (changes requested):
please let me resovle review items, jsut reply "done" (or something similar) as ACK that you addressed it and it is ready for review. 

items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
romani: @zbynek , any thoughts on this review item. 
do not hesitate to share any concerns.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
romani: inputs are extended, but let me keep it as unresovled to let followup with with my thoughts on what we should use as tokens.

My major concern of this PR is that we add bunch of tokens and it can lead to undesirable violations, and it is extension of API for Check, if we add them, we can not easily recall default tokens (it will be breaking change and user will hate this)

romani:
@zbynek , on january we plan to migrate checkstyle to jdk21, so it might be frustrating for you, as you shared that your target projects are still on jdk17. 
I would be awesome to finish this PR before early January to let me merge and release your functionality on jdk17 compilation, to let you use your work without problems and right away.

romani:
Github, generate report for UnnecessaryParentheses/all-examples-in-one

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesFieldMethodAccess.java
zbynek: I've turned that into a TODO comment.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
zbynek: I thought your concern was about the special handling of `METHOD_CALL`, I refactored the code to make it more clear why that one token is handled differently.

There is not much of new logic here, the new tokens can be handled the same as the operators you linked, but maybe  `unnecessaryParenAroundOperators` should be renamed to `unnecessaryParenAroundExpressions` to be better aligned with Java terminology? Some languages consider `[` and `.` to be "operators" as well, but JLS talks about "array access expressions" and "field access expressions".

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesFieldMethodAccess.java
zbynek: done

romani:
Github, generate report for UnnecessaryParentheses/all-examples-in-one

zbynek:
@romani is there anything I can do to help get this merged?

mahfouz72:
@zbynek, can you please resolve conflicts and rebase? I will continue reviewing asap, and let's get this PR merged

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -482,13 +502,28 @@ private static boolean unnecessaryParenAroundOperators(DetailAST ast) {
...
+     * Check that an expression is not the first argument of a conditional ternary operator.
+     *
+     * @param ast expression
+     * @return whether the expression is not the first argument of a ternary operator
+     */
+    private static boolean isNotFirstArgOfTernary(DetailAST ast) {
+        final DetailAST parent = ast.getParent();
+        return parent.getParent().getType() != TokenTypes.QUESTION
+                || !parent.equals(parent.getParent().getFirstChild().getNextSibling());
+    }
```
mahfouz72, line 531: Can you explain to me why we have this method?

---

> parent.equals(parent.getParent().getFirstChild().getNextSibling());
    }

Can we refactor this condition some how to read cleaner

---

Also  this assumes the operand is wrapped in a `METHOD_CALL`, so `parent.getParent()` reaches the `QUESTION`. That only holds for a method-call condition. what about for a direct field or array access conditions like:

`(m.flag) ? 1 : 2 (field)  `

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesFieldMethodAccess.java
```diff
@@ -0,0 +1,90 @@
...
+        // violation above 'Unnecessary parentheses around expression.'
+        foo = INSTANCE.name.length();
+        foo = (INSTANCE.name).length();
+        // violation above 'Unnecessary parentheses around expression.'
+        foo = new StringBuilder().length();
+        foo = (new StringBuilder()).length();
+        // violation above 'Unnecessary parentheses around expression.'
+        boolean b = INSTANCE.check() ? (INSTANCE.check()) : true;
+        // violation above 'Unnecessary parentheses around expression.'
+        foo = (INSTANCE.check()) ? 1 : 2;
```
mahfouz72: Is this false negative?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -394,18 +406,33 @@ public void leaveToken(DetailAST ast) {
...
-            else if (isSurrounded(ast) && unnecessaryParenAroundOperators(ast)) {
-                log(ast.getPreviousSibling(), MSG_EXPR);
+            else if (isSurrounded(selfOrParentMethodCall) && unnecessaryParenAroundOperators(ast)) {
+                log(selfOrParentMethodCall.getPreviousSibling(), MSG_EXPR);
             }
         }
     }
 
+    /**
+     * Get the node itself ot its parent, if it's a method call.
```
mahfouz72, line 430:
```suggestion
     * Get the node itself or its parent, if it's a method call.
```

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
```diff
@@ -275,6 +282,8 @@ public int[] getDefaultTokens() {
             TokenTypes.BNOT,
             TokenTypes.POST_INC,
             TokenTypes.POST_DEC,
+            TokenTypes.INDEX_OP,
+            TokenTypes.DOT,
```
mahfouz72, line 293: Can we add all new tokens in `getAcceptableTokens()` only and not in default. IMO It is better to keep default list is limited especially in this check

@romani What do you think? this will also may be resolve [your concern](https://github.com/checkstyle/checkstyle/pull/17575/changes#r2647437562) here a bit

mahfouz72 (changes requested):
looks good to me, some items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
romani: >Can we add all new tokens in getAcceptableTokens() 

you probably meant getRequiredTokens, right ? as getAcceptableTokens is biggest(most full) set of tokens to run on. 

I am ok to make them required if can prove that it is always safe to remove Parentheses.
I am ok to continue model, where `token` property kind define area/target to where apply Unneccessary Parentheses.

My major concern was that we need to make sure that removal of Parentheses is actually correct, Code continue to compile but works differently

from issue:

> Parentheses helps people who do not remember priority rules for java https://introcs.cs.princeton.edu/java/11precedence/

we just need to confident that Check demand no Parentheses in places where is it definitely safe

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
zbynek: I've changed it to work for field access as well. 

This is not strictly necessary, but looking at the reports it seemed to me that some developers prefer to have the first argument of the conditional operator in parentheses, even if they remove extra parentheses everywhere else. So this is a bit of an ad-hoc exception.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/unnecessaryparentheses/InputUnnecessaryParenthesesFieldMethodAccess.java
zbynek: https://github.com/checkstyle/checkstyle/pull/17575#discussion_r3610605136

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryParenthesesCheck.java
mahfouz72: @zbynek please share your opinion on this. Do you think we should remove them from default list

zbynek: These are [primary expressions](https://docs.oracle.com/javase/specs/jls/se25/html/jls-15.html#jls-15.8), so it's always safe to remove parentheses around them.

As the [report](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/44633394b973bdc56de5216af557d452526fcd7e_2025021311/reports/diff/index.html) shows, projects that prefer to avoid extra parentheses have a low number of new violations -- e.g. in Checkstyle the 9 violations are all in input files, none in production code. 

So I'd suggest to keep the default as it is currently implemented in this PR -- `.`, `[]` checked by default, `new` optional.

mahfouz72 (approved):
I am Ok to merge

zbynek:
I can rebase this again once it's approved.

romani (approved):
Approved. Please rebase

Sorry for long review process.

romani:

romani:
> Diff Regression patch config:
 
 it does  have LITERAL_NEW  token. ` <property name="tokens" value="LITERAL_NEW, ....`  so w are good

romani:
good ,we are waiting for rebase to merge this PR.

Issue #17269, opened 2025-06-23
Upgrade all scripts to groovy 3.0.17 or 5.x


romani:
Detected at https://github.com/checkstyle/checkstyle/pull/17267/files

```
Run groovy ./.ci/error-prone-check.groovy test-compile
Caught: java.lang.VerifyError: (class: java_util_Set$of, method: callStatic signature: (Ljava/lang/Class;[Ljava/lang/Object;)Ljava/lang/Object;) Illegal type in constant pool
java.lang.VerifyError: (class: java_util_Set$of, method: callStatic signature: (Ljava/lang/Class;[Ljava/lang/Object;)Ljava/lang/Object;) Illegal type in constant pool
	at error-prone-check.<clinit>(error-prone-check.groovy:9)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)

```

https://github.com/checkstyle/checkstyle/blob/5999226299cedb2bd3926b1fd648792ca86e28d7/.ci/error-prone-check.groovy#L9C33-L9C41

We need to upgrade scripts to newest groovy, groovy 4 is also fine 
~Groovy 5 is  not released yet https://groovy.apache.org/download.html~
https://groovy.apache.org/download.html , groovy 5 is out !!! lets use it !!!

Extra installation of groovy from that PR should be reverted .github/workflows/error-prone.yml

We need to migrate all groovy scripts in our organization, there are few, not a lot.

Contribution repo migration: https://github.com/checkstyle/contribution/issues/957

romani:
@AmitKumarDeoghoria , 
https://github.com/checkstyle/checkstyle/blob/5999226299cedb2bd3926b1fd648792ca86e28d7/.ci/error-prone-check.groovy#L9C33-L9C41

[...]

PR #17581, opened 2025-08-07
Issue #17269: Use docker Image Instead of Extra Installation

AmitKumarDeoghoria:
part of : #17269

### DIFF
--- a/.github/workflows/error-prone.yml
+++ b/.github/workflows/error-prone.yml
@@ -15,20 +15,14 @@ jobs:
   error-prone:
     if: github.repository == 'checkstyle/checkstyle'
     runs-on: ubuntu-latest
+    container:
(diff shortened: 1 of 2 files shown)


.github/workflows/error-prone.yml
```diff
@@ -15,21 +15,9 @@ jobs:
   error-prone:
     if: github.repository == 'checkstyle/checkstyle'
     runs-on: ubuntu-latest
+    container:
+      image: amitkumardeoghoria/jdk-17-groovy-git-mvn-latest:latest
     steps:
-      - name: Set up JDK 17
-        uses: actions/setup-java@v4
```
romani, line 20: Let's keep all workflows as is, better to use native to GitHub CI images

romani (changes requested):
I don't think we need to update the this

romani:
Let's not do this.
We should use special docker container only in places where we already use it.

PR #17585, opened 2025-08-07
supplemental: IT covering record for '7.3 Where Javadoc is used'

mohitsatr:
from: https://github.com/checkstyle/checkstyle/pull/17165#discussion_r2125501633

### DIFF
--- a/src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule73wherejavadocrequired/InputJavadocTypeOnRecord.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule73wherejavadocrequired/InputJavadocTypeOnRecord.java
@@ -1,0 +1,53 @@
+package com.google.checkstyle.test.chapter7javadoc.rule73wherejavadocrequired;
+
+// violation below 'Missing a Javadoc comment.'
+public record InputJavadocTypeOnRecord(String name) {
+
+  // violation below 'Missing a Javadoc comment.'
+  public record Multi(String main, Record rec) {
+    private static boolean isSent(int obj) {
+      return obj > 10;
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule73wherejavadocrequired/InputJavadocTypeOnRecord.java
```diff
@@ -0,0 +1,55 @@
...
+    // violation below 'Missing a Javadoc comment.'
+    public Mapping(String from) {
+      this.from = from;
+    }
+  }
+
+  // violation below 'Missing a Javadoc comment.'
+  protected record Multi4(String s1, String s2) {
+
+    protected Multi4 {
```
mohitsatr: should it have given violation here?

romani: No as classes/... that are nested in methods never showed in javadoc reports

romani:
https://www.diffchecker.com/XKH9NduM/

I lost a reason why we need formatted files that very different https://www.diffchecker.com/XKH9NduM/

We should have no formatted files , or have formatted that just formatted by tool, no more changes. Diff should be almost zero

src/it/resources/com/google/checkstyle/test/chapter7javadoc/rule73wherejavadocrequired/InputJavadocTypeOnRecord.java
mohitsatr: resolved.

mohitsatr:
@romani IT is not about formatting so we should skip `InputFormatted..` files for such cases.

Issue #17300, opened 2025-06-30
Convert switch statements to enhanced switch expressions


AmitKumarDeoghoria:
From comment : https://github.com/checkstyle/checkstyle/pull/17259#issuecomment-3019192404
During the JDK 17 upgrade(https://github.com/checkstyle/checkstyle/issues/17168), IntelliJ inspections identified 68 switch statements that can be replaced with modern [enhanced switch expressions](https://openjdk.org/jeps/361) (introduced in JDK 14). To maintain focus on the core upgrade, we will temporarily disable these inspections and address them in a dedicated refactoring effort post-migration.

Next Steps

- Disable inspection in current JDK 11 PR

- Post-JDK 17 upgrade:

- Re-enable inspection

- Refactor using IDE's "Replace with enhanced switch" tool

- Validate with tests

AmitKumarDeoghoria:
This will be a  huge PR with 80+ files affected @romani  how do u suggest we move forward for this.

romani:
Please proceed, but split changes in several PRs. There most likely be not easy cases, immediately avoid them and move them to separate PRs to not block bulk work

romani:
@AmitKumarDeoghoria , please share how much is left to be done ? screenshot or text (preferable).

AmitKumarDeoghoria:
Plz close this Issue

romani:
fix is merged

PR #17586, opened 2025-08-07
Issue #17300: Convert switch statements to enhanced switch expressions

AmitKumarDeoghoria:
part of :  #17300

### DIFF
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -5559,7 +5559,7 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/gui/ParseTreeTablePresentation.java</fileName>
     <specifier>return</specifier>
     <message>incompatible types in return.</message>
-    <lineContent>return value;</lineContent>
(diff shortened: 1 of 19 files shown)


romani:
Please fix CI 
If some cases are complicated and problematic move them to separate PRs.
Don't try to fix all at ones, such approach usually requires too much time.
Better to fix easy case first.

romani:
```
[INFO] --- pmd:3.27.0:check (default) @ checkstyle ---
[WARNING] PMD Failure: com.puppycrawl.tools.checkstyle.SarifLogger:272
Rule:UnnecessaryLocalBeforeReturn Priority:3
Consider simply returning the value
 vs storing it in local variable 'renderedSeverityLevel'.

```

Please update to
```
return switch (severityLevel) {
            case IGNORE -> "none";
            case INFO -> "note";
            case WARNING -> "warning";
            default -> "error";
        };
```

As variable matching to method name completely `renderSeverityLevel(SeverityLevel`

AmitKumarDeoghoria:
There are another 20-25 files, which needs to be changes, will do in other PR's

romani (approved):
Thank a lot for doing updates in batches, not all in one.
You can do update even in smaller sizes. They might be problematic

Issue #17429, opened 2025-07-20
Preparation for Migration of Checkstyle Codebase to Java 21


romani:
similar to https://github.com/checkstyle/checkstyle/issues/17168
we need to update all internals to jdk21 usage but not a pom.xml with minimal jdk usage.

after  preparation is done we will mirgate like we did at https://github.com/checkstyle/checkstyle/issues/17321 but in separate issue.

https://en.m.wikipedia.org/wiki/Java_version_history
https://docs.oracle.com/en/java/javase/24/migrate/significant-changes-jdk-21.html

Reason to enable preview features of jdk in checkstyle:
- Checkstyle vis always behind in jdk usage by 1 or 2 lat releases.
- Preview features are appearing and removed, good example is "String templates" in jdk21. It is very awaited feature by users, so user immediately would use it, but it challenging in consistent behavior, so it was removed few releases later. If checkstyle started to use such feature, it will be pain to refactor to stop using it. We removed support of string templates, as it was big  problem for us to support too.
- checkstyle has its own parser of java, so even jdk preview feature is in , and we enable them, on first usage of it checkstyle will fail to parse it , CI will be very red. So we naturally limited to features that checkstyle java parser supports. Adding features to our parser is huge effort, and highly unlikely to be done for preview features.

Final migration to jdk21 will be done at https://github.com/checkstyle/checkstyle/issues/17430

romani:
@AmitKumarDeoghoria , 
please send PR to move them to compiled
there are some leftovers from old JDK:
```
✔ ~/java/github/romani/checkstyle/src/test/resources-noncompilable [master|✔] 
 $ ag "Compilable with"
com/puppycrawl/tools/checkstyle/checks/finalparameters/InputFinalParametersRecordForLoopPatternVariables.java
10:// non-compiled with javac: Compilable with Java20

com/puppycrawl/tools/checkstyle/checks/coding/constructorsdeclarationgrouping/InputConstructorsDeclarationGroupingRecords.java
7:// non-compiled with javac: Compilable with Java14

com/puppycrawl/tools/checkstyle/checks/coding/modifiedcontrolvariable/InputModifiedControlVariableRecordDecomposition.java
7:// non-compiled with javac: Compilable with Java20

com/puppycrawl/tools/checkstyle/grammar/java19/InputGenericRecordDeconstructionPattern.java
1:// non-compiled with javac: Compilable with Java19

[...]

PR #17602, opened 2025-08-11
Issue #17429: Relocate inputs to compilable path for Modifier Module

AmitKumarDeoghoria:
part of : #17429

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/modifier/InterfaceMemberImpliedModifierCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/modifier/InterfaceMemberImpliedModifierCheckTest.java
@@ -403,7 +403,7 @@ public class InterfaceMemberImpliedModifierCheckTest
             "40:5: " + getCheckMessage(MSG_KEY, "static"),
         };
         verifyWithInlineConfigParser(
-                getNonCompilablePath("InputInterfaceMemberImpliedModifierSealedClasses.java"),
+                getPath("InputInterfaceMemberImpliedModifierSealedClasses.java"),
             expected);
     }
 }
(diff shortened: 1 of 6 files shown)


AmitKumarDeoghoria:
Eclipse still compiling Input files : 
15. ERROR in /home/semaphore/checkstyle/src/test/resources/com/puppycrawl/tools/checkstyle/checks/modifier/classmemberimpliedmodifier/InputClassMemberImpliedModifierRecords.java (at line 42)00:38
	public record BadRecord() {} // violation00:38
	              ^^^^^^^^^00:38
Illegal modifier for the local record BadRecord; only final and strictfp are permitted00:38
----------00:38
----------00:38
16. ERROR in /home/semaphore/checkstyle/src/test/resources/com/puppycrawl/tools/checkstyle/checks/modifier/classmemberimpliedmodifier/InputClassMemberImpliedModifierNoViolationRecords.java (at line 41)00:38
	public record BadRecord() {}00:38
	              ^^^^^^^^^00:38
Illegal modifier for the local record BadRecord; only final and strictfp are permitted

romani:
Inputs are always compilable, it help Eclipse users to keep project wihtout errors, do you see a reason why public records are not allowed ?

If this is Elcipse special compiler error? we can keep Input in non-compiled folder and make special comment that it is non-compiled only due to Eclipse. One day in future we will upgrade eclipse version and this error might be gone.
We already have few Inputs that are in non-compiled folder due to Eclipse compiler only.

in this case we should not remove modifier, as it is allowed by javac and this Input is explicitly focused on modifiers

AmitKumarDeoghoria:
@romani  That’s not an Eclipse-specific quirk, the code is genuinely invalid according to the Java Language Specification.

Here’s what’s I found from Internet sources :

The key issue
This part:
```
Object obj = new Object() {
    public record BadRecord() {} // violation
    public static record OkRecord() {}
};
```
is inside an anonymous class body.

In Java, you cannot declare a record (or enum, or interface, or static class) inside a local/inner class unless it is static and at the top level of a member type.
Records are implicitly static when nested inside another class or interface, but in local/anonymous classes, the rules are stricter:
No static declarations are allowed except for constant variables.

romani:
If this is true, why maven compilation is not failing?

romani:
Please move this input on a side to separate PR, let's others to merged

romani (approved):

romani:
Screenshot_20250812-204020

romani:
Online editor with jdk21 has no problem with this code.
Let's keep this code as only Eclipse compiler can not handle, if eventually javac do same error to can remove such file, but for now is generally good syntax and we should have to work on it without problems.

PR #17629, opened 2025-08-16
Issue #17429: Relocate inputs to compilable path for Coding Module5

AmitKumarDeoghoria:
part of : #17429


AmitKumarDeoghoria:
Mostly all non compilable moved to compilable.

PR #17643, opened 2025-08-17
Issue #17429: Migrating eclipse compiler to jdk21

AmitKumarDeoghoria:
part of : #17429

### DIFF
--- a/.ci/eclipse-compiler-javac.sh
+++ b/.ci/eclipse-compiler-javac.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-DEFAULT_JAVA_RELEASE=17
+DEFAULT_JAVA_RELEASE=21
 if [ -z "$1" ]; then
     echo "No parameters supplied!"
     echo "Usage %0 <CLASSPATH> [RELEASE]"


.ci/eclipse-compiler-javac.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-DEFAULT_JAVA_RELEASE=17
+DEFAULT_JAVA_RELEASE=21
```
romani, line 4: In addition, We need to change it where script is executed, default here does not change much

AmitKumarDeoghoria: I dont think we have to change anything here right , it will be mostly in this ecj script and maven profile:`eclipse-static-analysis)
  ./mvnw -e --no-transfer-progress clean compile exec:exec -Peclipse-compiler
  ;;`

romani: https://github.com/checkstyle/checkstyle/blob/f330076a6c394e635fbf525078f539b068866360/pom.xml#L4967-L4971 next jdk version property  already here

PR #17705, opened 2025-08-28
Issue #3110: Remaining hardcoded english in Checker.java is translated in message files

Praveen7294:
part of #3110

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
@@ -185,7 +185,8 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
                 cacheFile.persist();
             }
             catch (IOException exc) {
-                throw new IllegalStateException("Unable to persist cache file.", exc);
+                throw new IllegalStateException(
+                        getLocalizedMessage("Checker.cacheFilesException"), exc);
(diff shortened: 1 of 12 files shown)


romani:
Thanks a lot, CI should be fixed.
XWiki failure is not related, restarted.

Praveen7294:
Done.

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -655,7 +656,7 @@ public void clearCache() {
      * @param args the arguments of message in respective properties file.
      * @return a string containing extracted localized message
      */
-    private String getLocalizedMessage(String messageKey, Object... args) {
+    public String getLocalizedMessage(String messageKey, Object... args) {
```
romani: Let's not open this method to users.
All public methods are contract with customers,  we should minimize contractual obligations.

I am fine to copy this method to test clas to get localed message

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
Praveen7294: Done

romani (approved):
Thanks a lot

Issue #17507, opened 2025-07-30
Google style: Missing support for multipart versioning format using underscores in method names


mohitsatr:
From: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case
> In very rare circumstances (for example, multipart version numbers), you may need to use underscores to separate adjacent numbers, since numbers do not have upper and lower case variants. 

Correct version: `guave33_4_6`
incorrect version: `guava3346`

Code:
```java
/** some javadoc. */
public class Test {

  String guava33_4_6 = "Guava 33.4.6"; // False-positive
  String guava3346 = "Guava 33.4.6";   // False-negative 
  
  String jdk8_0_392 = "Jdk 8.0.392";   // False-positive
  String jdk80392 = "Jdk 8.0.392";     // False-negative

  String kotlin1_9_24 = "Kotlin 1.9.24"; // False-positive
  String kotlin1924 = "Kotlin 1.9.24";   // False-negative

  String gradle8_5_1 = "Gradle 8.5.1";   // False-positive
  String gradle851 = "Gradle 8.5.1";     // False-negative
}

```

Cli:
```
$ java -jar checkstyle-10.26.1-all.jar -c google_checks.xml Test.java 
Starting audit...
[WARN] /mnt/5D92528E6B945467/test/testing/Test.java:4:10: Member name 'guava33_4_6' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [MemberName]
[WARN] /mnt/5D92528E6B945467/test/testing/Test.java:7:10: Member name 'jdk8_0_392' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [MemberName]
[WARN] /mnt/5D92528E6B945467/test/testing/Test.java:10:10: Member name 'kotlin1_9_24' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [MemberName]
[WARN] /mnt/5D92528E6B945467/test/testing/Test.java:13:10: Member name 'gradle8_5_1' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [MemberName]
Audit done.
```
The style guide doesn't mention this explicitly, and we can have a discussion about it, but I think it's reasonable to assume that such naming conventions are likely intended only for member names.

After adding support for underscores between digits, coverage table should be updated.

Zopsss:
We can allow underscores between digits but regex maybe very tricky for such cases. @mohitsatr

[...]

PR #17708, opened 2025-08-29
Issue #17507: Support for multipart versioning format using underscores

mohitsatr:
part of #17507 

WIP

### DIFF
--- a/src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
@@ -3,35 +3,83 @@ package com.google.checkstyle.test.chapter5naming.rule53camelcase;
 /** some javadoc. */
 public class InputCamelCaseMultipartVersioningNames {
 
-  private final String guava33_4_6 = "Guava 33.4.6"; // false-positive until #17507
-  // violation above 'Member name 'guava33_4_6' must match pattern'
+  private final String guava33_4_6 = "Guava 33.4.6";
 
-  private final String guava3346 = "Guava 33.4.6";
+  private final String guava3346 = "Guava 33.4.6"; // false-negative until #17507
 
-  private final String jdk8_0_392 = "Jdk 8.0.392"; // false-positive until #17507
-  // violation above 'Member name 'jdk8_0_392' must match pattern'
+  private final String jdk8_0_392 = "Jdk 8.0.392";
 
-  private final String jdk80392 = "Jdk 8.0.392";
+  private final String jdk80392 = "Jdk 8.0.392"; // false-negative until #17507
 
-  private final String kotlin1_9_24 = "Kotlin 1.9.24"; // false-positive until #17507
-  // violation above 'Member name 'kotlin1_9_24' must match pattern'
+  private final String kotlin1_9_24 = "Kotlin 1.9.24";
 
-  private final String kotlin1924 = "Kotlin 1.9.24";
+  private final String kotlin1924 = "Kotlin 1.9.24"; // false-negative until #17507
 
-  private final String gradle8_5_1 = "Gradle 8.5.1"; // false-positive until #17507
-  // violation above 'Member name 'gradle8_5_1' must match pattern'
+  private final String gradle8_5_1 = "Gradle 8.5.1";
 
-  private final String gradle851 = "Gradle 8.5.1";
+  private final String gradle851 = "Gradle 8.5.1"; // false-negative until #17507
 
-  void guava34_4_6() {} // false-positive until #17507
-  // violation above 'Method name 'guava34_4_6' must match pattern'
+  class InnerBad {
+    String guava_33_4_6 = "guava_33_4_6";
+    // violation above 'Member name 'guava_33_4_6' must match pattern'
 
-  void kotlin2_9_94() {} // false-positive until #17507
-  // violation above 'Method name 'kotlin2_9_94' must match pattern'
+    String guava_version33_4_6 = "guava_version33_4_6";
+    // violation above 'Member name 'guava_version33_4_6' must match pattern'
 
-  void gradle9_5_1() {} // false-positive until #17507
-  // violation above 'Method name 'gradle9_5_1' must match pattern'
+    String guava33_4_6_ = "guava33_4_6_";
+    // violation above 'Member name 'guava33_4_6_' must match pattern'
 
-  void jdk9_0_392() {} // false-positive until #17507
-  // violation above 'Method name 'jdk9_0_392' must match pattern'
+    String jdk_8_90 = "jdk_8_90";
+    // violation above 'Member name 'jdk_8_90' must match pattern'
+
+    String jdk_version8_0_392 = "jdk_version8_0_392";
+    // violation above 'Member name 'jdk_version8_0_392' must match pattern'
+
+    String jdk8_90_ = "jdk8_90_";
+    // violation above 'Member name 'jdk8_90_' must match pattern'
+
+    String kotlin_1_9_24 = "kotlin_1_9_24";
+    // violation above 'Member name 'kotlin_1_9_24' must match pattern'
+
+    String kotlin_version1_9_24 = "kotlin_verison33_4_24";
+    // violation above 'Member name 'kotlin_version1_9_24' must match pattern'
+
+    String kotlin1_9_24_ = "kotlin1_9_24_";
+    // violation above 'Member name 'kotlin1_9_24_' must match pattern'
+
+    void guava_33_4_6() {}
+    // violation above 'Method name 'guava_33_4_6' must match pattern'
+
+    void guava_Version33_4_6() {}
+    // violation above 'Method name 'guava_Version33_4_6' must match pattern'
+
+    void guava_33_4_6_() {}
+    // violation above 'Method name 'guava_33_4_6_' must match pattern'
+
+    void jdk_8_90() {}
+    // violation above 'Method name 'jdk_8_90' must match pattern'
+
+    void jdk_method8_90() {}
+    // violation above 'Method name 'jdk_method8_90' must match pattern'
+
+    void jdk_8_90_() {}
+    // violation above 'Method name 'jdk_8_90_' must match pattern'
+
+    void kotlin_1_9_24() {}
+    // violation above 'Method name 'kotlin_1_9_24' must match pattern'
+
+    void kotlin_lang1_9_24() {}
+    // violation above 'Method name 'kotlin_lang1_9_24' must match pattern'
+
+    void kotlin1_9_24_() {}
+    // violation above 'Method name 'kotlin1_9_24_' must match pattern'
+  }
+
+  void guava34_4_6() {}
+
+  void kotlin2_9_94() {}
+
+  void gradle9_5_1() {}
+
+  void jdk9_0_392() {}
 }
--- a/src/main/resources/google_checks.xml
+++ b/src/main/resources/google_checks.xml
@@ -240,7 +240,7 @@
              value="Type name ''{0}'' must match pattern ''{1}''."/>
     </module>
     <module name="MemberName">
-      <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9]*$"/>
+      <property name="format" value="^[a-z][a-z0-9]*(?:[A-Z][a-zA-Z0-9]*|[0-9](?:_[0-9]+)*)?$"/>
       <message key="name.invalidPattern"
              value="Member name ''{0}'' must match pattern ''{1}''."/>
     </module>
@@ -419,7 +419,7 @@
       <property name="excludeScope" value="nothing"/>
     </module>
     <module name="MethodName">
-      <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9]*$"/>
+      <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9]*$|^[a-z][a-z0-9]*[0-9](?:_[0-9]+)*$"/>
       <message key="name.invalidPattern"
              value="Method name ''{0}'' must match pattern ''{1}''."/>
     </module>
--- a/src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputFormattedCamelCaseMultipartVersioningNames.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputFormattedCamelCaseMultipartVersioningNames.java
@@ -2,38 +2,84 @@ package com.google.checkstyle.test.chapter5naming.rule53camelcase;
 
 /** some javadoc. */
 public class InputFormattedCamelCaseMultipartVersioningNames {
-  private final String guava33_4_6 = "Guava 33.4.6"; // false-positive until #17507
-  // violation above 'Member name 'guava33_4_6' must match pattern'
 
-  private final String guava3346 = "Guava 33.4.6";
+  private final String guava33_4_6 = "Guava 33.4.6";
 
-  private final String jdk8_0_392 = "Jdk 8.0.392"; // false-positive until #17507
-  // violation above 'Member name 'jdk8_0_392' must match pattern'
+  private final String guava3346 = "Guava 33.4.6"; // false-negative until #17507
 
-  private final String jdk80392 = "Jdk 8.0.392";
+  private final String jdk8_0_392 = "Jdk 8.0.392";
 
-  private final String kotlin1_9_24 = "Kotlin 1.9.24"; // false-positive until #17507
-  // violation above 'Member name 'kotlin1_9_24' must match pattern'
+  private final String jdk80392 = "Jdk 8.0.392"; // false-negative until #17507
 
-  private final String kotlin1924 = "Kotlin 1.9.24";
+  private final String kotlin1_9_24 = "Kotlin 1.9.24";
 
-  private final String gradle8_5_1 = "Gradle 8.5.1"; // false-positive until #17507
-  // violation above 'Member name 'gradle8_5_1' must match pattern'
+  private final String kotlin1924 = "Kotlin 1.9.24"; // false-negative until #17507
 
-  private final String gradle851 = "Gradle 8.5.1";
+  private final String gradle8_5_1 = "Gradle 8.5.1";
 
-  void guava34_4_6() {} // false-positive until #17507
+  private final String gradle851 = "Gradle 8.5.1"; // false-negative until #17507
 
-  // violation 2 lines above 'Method name 'guava34_4_6' must match pattern'
+  class InnerBad {
+    String guava_33_4_6 = "guava_33_4_6";
+    // violation above 'Member name 'guava_33_4_6' must match pattern'
 
-  void kotlin2_9_94() {} // false-positive until #17507
+    String guava_version33_4_6 = "guava_version33_4_6";
+    // violation above 'Member name 'guava_version33_4_6' must match pattern'
 
-  // violation 2 lines above 'Method name 'kotlin2_9_94' must match pattern'
+    String guava33_4_6_ = "guava33_4_6_";
+    // violation above 'Member name 'guava33_4_6_' must match pattern'
 
-  void gradle9_5_1() {} // false-positive until #17507
+    String jdk_8_90 = "jdk_8_90";
+    // violation above 'Member name 'jdk_8_90' must match pattern'
 
-  // violation 2 lines above 'Method name 'gradle9_5_1' must match pattern'
+    String jdk_version8_0_392 = "jdk_version8_0_392";
+    // violation above 'Member name 'jdk_version8_0_392' must match pattern'
 
-  void jdk9_0_392() {} // false-positive until #17507
-  // violation above 'Method name 'jdk9_0_392' must match pattern'
+    String jdk8_90_ = "jdk8_90_";
+    // violation above 'Member name 'jdk8_90_' must match pattern'
+
+    String kotlin_1_9_24 = "kotlin_1_9_24";
+    // violation above 'Member name 'kotlin_1_9_24' must match pattern'
+
+    String kotlin_version1_9_24 = "kotlin_verison33_4_24";
+    // violation above 'Member name 'kotlin_version1_9_24' must match pattern'
+
+    String kotlin1_9_24_ = "kotlin1_9_24_";
+    // violation above 'Member name 'kotlin1_9_24_' must match pattern'
+
+    void guava_33_4_6() {}
+    // violation above 'Method name 'guava_33_4_6' must match pattern'
+
+    void guava_Version33_4_6() {}
+    // violation above 'Method name 'guava_Version33_4_6' must match pattern'
+
+    void guava_33_4_6_() {}
+    // violation above 'Method name 'guava_33_4_6_' must match pattern'
+
+    void jdk_8_90() {}
+    // violation above 'Method name 'jdk_8_90' must match pattern'
+
+    void jdk_method8_90() {}
+    // violation above 'Method name 'jdk_method8_90' must match pattern'
+
+    void jdk_8_90_() {}
+    // violation above 'Method name 'jdk_8_90_' must match pattern'
+
+    void kotlin_1_9_24() {}
+    // violation above 'Method name 'kotlin_1_9_24' must match pattern'
+
+    void kotlin_lang1_9_24() {}
+    // violation above 'Method name 'kotlin_lang1_9_24' must match pattern'
+
+    void kotlin1_9_24_() {}
+    // violation above 'Method name 'kotlin1_9_24_' must match pattern'
+  }
+
+  void guava34_4_6() {}
+
+  void kotlin2_9_94() {}
+
+  void gradle9_5_1() {}
+
+  void jdk9_0_392() {}
 }


mohitsatr:
Github, generate report

config/checkstyle-checks.xml
```diff
@@ -72,6 +72,12 @@
     <property name="lineRange" value="-2"/>
     <property name="idPattern" value="lineLengthXml"/>
   </module>
+  <module name="SuppressWithNearbyTextFilter">
+    <property name="nearbyTextPattern"
+        value="MemberName"/>
+    <property name="lineRange" value="2"/>
+    <property name="idPattern" value="lineLengthXml"/>
+  </module>
```
Zopsss: why this was modified?

src/main/resources/google_checks.xml
```diff
@@ -419,7 +420,8 @@
       <property name="excludeScope" value="nothing"/>
     </module>
     <module name="MethodName">
-      <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9]*$"/>
+      <property name="format"
+              value="^[a-z][a-z0-9][a-zA-Z0-9]*$|^[a-z][a-z0-9]*[0-9](?:_[0-9]+)*$"/>
```
Zopsss: we don't need to modify this module, it is mentioned in the style guide that only test methods can have underscores:

https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

> Underscores may appear in JUnit test method names to separate logical components of the name, with each component written in [lowerCamelCase](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case), for example transferMoney_deductsFromSource.

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -3,35 +3,83 @@
...
 
-  private final String kotlin1924 = "Kotlin 1.9.24";
+  private final String kotlin1924 = "Kotlin 1.9.24"; // false-negative until #17507
 
-  private final String gradle8_5_1 = "Gradle 8.5.1"; // false-positive until #17507
-  // violation above 'Member name 'gradle8_5_1' must match pattern'
+  private final String gradle8_5_1 = "Gradle 8.5.1";
 
-  private final String gradle851 = "Gradle 8.5.1";
+  private final String gradle851 = "Gradle 8.5.1"; // false-negative until #17507
```
Zopsss: this is actually not a false-negative. We cannot enforce user to always use underscores. It totally depends on their situation.

Zopsss (changes requested):

config/checkstyle-checks.xml
mohitsatr: https://github.com/checkstyle/checkstyle/blob/79454cc771139758768b2bf8bdab479ad4716361/src/main/resources/google_checks.xml#L244 
this line exceeds the column limit, causing errors in build. this suppression suppresses that error.

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
mohitsatr: done

src/main/resources/google_checks.xml
mohitsatr: done

config/checkstyle-checks.xml
Zopsss: try:

```xml
    <module name="MemberName">
      <property name="format"
             value=
                "^(?![a-z]$)(?![a-z][A-Z])[a-z][a-z0-9]*(?:[A-Z][a-zA-Z0-9]*|[0-9](?:_[0-9]+)*)?$"
      />
      <message key="name.invalidPattern"
             value="Member name ''{0}'' must match pattern ''{1}''."/>
    </module>
```

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -3,35 +3,69 @@
...
+
+    // violation below 'Member name 'kotlin_1_9_24' must match pattern'
+    String kotlin_1_9_24 = "kotlin_1_9_24";
+
+    // violation below 'Member name 'kotlin_version1_9_24' must match pattern'
+    String kotlin_version1_9_24 = "kotlin_verison33_4_24";
+
+    String kotlin1_9_24_ = "kotlin1_9_24_"; // violation 'name 'kotlin1_9_24_' must match pattern'
+
+    void guava_33_4_6() {} // violation 'Method name 'guava_33_4_6' must match pattern'
```
Zopsss: this PR didn't needed to add cases for examples but lets keep it. We didn't had such cases so it just extends our coverage :)

src/main/resources/google_checks.xml
```diff
@@ -419,7 +420,8 @@
       <property name="excludeScope" value="nothing"/>
     </module>
     <module name="MethodName">
-      <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9]*$"/>
+      <property name="format"
+              value="^[a-z][a-z0-9][a-zA-Z0-9]*$"/>
```
Zopsss: please revert this.

src/it/java/com/google/checkstyle/test/chapter5naming/rule53camelcase/CamelCaseDefinedTest.java
```diff
@@ -43,9 +43,4 @@ public void testCamelCaseNamesFormatted() throws Exception {
     public void testCamelCaseNamesMultipartVersioningNames() throws Exception {
         verifyWithWholeConfig(getPath("InputCamelCaseMultipartVersioningNames.java"));
     }
-
-    @Test
-    public void testFormattedCamelCaseNamesMultipartVersioningNames() throws Exception {
-        verifyWithWholeConfig(getPath("InputFormattedCamelCaseMultipartVersioningNames.java"));
-    }
```
Zopsss, line 50: why was this removed?

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
mohitsatr: :smiley:

config/checkstyle-checks.xml
mohitsatr: ok it worked. but don't you think code looks ugly?

src/it/java/com/google/checkstyle/test/chapter5naming/rule53camelcase/CamelCaseDefinedTest.java
mohitsatr: we don't that file in this case because the change is not related to formatting. 
The reason I had to create that file was because Formatter kept adding extra lines between code and `// violation` comment.
Now I have changed the position of 'violation' comments and Formatter is okay.

config/checkstyle-checks.xml
Zopsss: if we're able to remove the suppression then we should do it :)

src/it/java/com/google/checkstyle/test/chapter5naming/rule53camelcase/CamelCaseDefinedTest.java
romani: Yes, removed from config/google-java-format/excluded/compilable-input-paths.txt

config/checkstyle-checks.xml
romani: why we cannot simply indent this line more to left, to avoid violation ?

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -3,35 +3,69 @@
...
+    void kotlin1_9_24_() {} // violation 'Method name 'kotlin1_9_24_' must match pattern'
+  }
+
+  void guava34_4_6() {} // violation 'Method name 'guava34_4_6' must match pattern'
+
+  void kotlin2_9_94() {} // violation 'Method name 'kotlin2_9_94' must match pattern'
+
+  void gradle9_5_1() {} // violation 'Method name 'gradle9_5_1' must match pattern'
+
+  void jdk9_0_392() {} // violation 'Method name 'jdk9_0_392' must match pattern'
```
romani: why this is violation ?
if jdk8_0_392 is not a violation.

Lets update regexp to methods and types too. right ?
or you want it to do in separate issues ?
if yes , please create separete issues and rename current to fields special, and we can merge this PR quickly.

romani (changes requested):
items:

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
Zopsss: Underscores are allowed for test methods

Zopsss: https://github.com/checkstyle/checkstyle/pull/17708#discussion_r2314978667

Zopsss: Reason why we kept this new cases: https://github.com/checkstyle/checkstyle/pull/17708#discussion_r2317796140

config/checkstyle-checks.xml
Zopsss: @mohitsatr please try above one ^^^^

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
romani: there is no way to write this name in different way.
from [style guide](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case):
> In very rare circumstances (for example, multipart version numbers), you may need to use underscores to separate adjacent numbers, since numbers do not have upper and lower case variants. 

so this name should be "ok".

Zopsss: alright. @mohitsatr Please revert my suggestion at: https://github.com/checkstyle/checkstyle/pull/17708#discussion_r2314978667 and add back support for method names.

mohitsatr:
Github, generate report

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
mohitsatr: done

src/main/resources/google_checks.xml
mohitsatr: resolved

config/checkstyle-checks.xml
mohitsatr: done

romani (changes requested):
from guideline:

>for example transferMoney_deductsFromSource. There is no One Correct Way to name test methods.

in regression report we have https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a4eaf13_2025072930/reports/diff/guava/index.html#A125
bunch of similar namings, that are still violation.

we need to have separate issue if we want to addreess them in separately.

src/main/resources/google_checks.xml
```diff
@@ -424,7 +425,7 @@
       <property name="excludeScope" value="nothing"/>
     </module>
     <module name="MethodName">
-      <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9]*$"/>
+      <property name="format" value="^[a-z](?:[a-z0-9][a-zA-Z0-9]*|[a-z0-9]*[0-9](?:_[0-9]+)*)$"/>
```
Zopsss: why MethodName and MemberName regex is different? Shouldn't it be same? @mohitsatr

src/it/java/com/google/checkstyle/test/chapter5naming/rule53camelcase/CamelCaseDefinedTest.java
Zopsss: resolved.

src/main/resources/google_checks.xml
Zopsss: resolved.

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
Zopsss: resolved.

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
Zopsss: resolved.

src/main/resources/google_checks.xml
Zopsss: we reverted this. We need to keep this. Resolved.

config/checkstyle-checks.xml
Zopsss: resolved.

src/main/resources/google_checks.xml
Zopsss: MethodName:
`value="^[a-z](?:[a-z0-9][a-zA-Z0-9]*|[a-z0-9]*[0-9](?:_[0-9]+)*)$"`

MemberName:
`value="^(?![a-z]$)(?![a-z][A-Z])[a-z][a-z0-9]*(?:[A-Z][a-zA-Z0-9]*|[0-9](?:_[0-9]+)*)?$"`

mohitsatr: They are shortened and simplifed versions to save line length.

mohitsatr:
@romani we cover these names. I tested them locally. I forgot to include the Suppression in the regression config.

mohitsatr:
Github, generate report

romani:
such methods does not have `@Test` so that is why it stays as violation https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a4eaf13_2025122759/reports/diff/guava/index.html#A3

am I right? 
strange that kind of same in our Inputs are ok.

src/main/resources/google_checks.xml
romani: lets keep all regexp the same, it will clearly shows that rules of format is same.

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -3,35 +3,77 @@
...
+
+    void jdk_8_90() {} // violation 'Method name 'jdk_8_90' must match pattern'
+
+    void jdk_method8_90() {} // violation 'Method name 'jdk_method8_90' must match pattern'
+
+    void jdk_8_90_() {} // violation 'Method name 'jdk_8_90_' must match pattern'
+
+    void kotlin_1_9_24() {} // violation 'Method name 'kotlin_1_9_24' must match pattern'
+
+    void kotlin_lang1_9_24() {} // violation 'Method name 'kotlin_lang1_9_24' must match pattern'
```
romani: I having hard time to understand why this method is violation .
Please add correctly named method above to let humans spot difference easily.

Please do same for all members of InnerBad.
you can create new  file if you  want.

Zopsss:
> such methods does not have @Test so that is why it stays as violation


Normal methods & Test methods, both follows lowerCamelCase so both of them follows same naming conventions. 

https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

---

The current regex we're using doesn't allow for camel casing characters and digits together, for example `testTest99_99`, `test_Test`, `testTest_Test`, etc. Please adjust the regex according to this @mohitsatr

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
mohitsatr: `guava33_4_6` is the only example for multipart versioning format in the style guide. so I tried creating regex pattern based on just this one example.

Current regex format say: if a characters are followed with numbers, (guava33) then underscores can be used to separate the additional numbers(guava33_4_6).
Any method name that does follow the above format, will have violation on it.

In `kotlin_lang1_9_24`, first underscore is used to separate characters(kotin_lang), which is not valid.
I tried `kotlinLang1_6_24` and got violation.

`kotlinlang1_6_24` works fine.


Since there is no clear of definition multipart versioning format with examples, I think It is really up to us how we want to go about implementing the rule.

mohitsatr:
@Zopsss @romani 

> Underscores may appear in JUnit test method names to separate logical components of the name

Underscores are only allowed in test method. we can't put them in normal Method names. Normal methods & Test methods, both use same regex format. so we allow the following formats, we are also making underscores valid for normal functions.

`testTest99_99`
`test_Test`
`testTest_Test`

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
romani: we should forcus on Google style not a checkstyle Checks limitations.
if current regexp based Check is not doing job, we need to substitute it with new Checks that does exactly what we(google style) need.

https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

>In very rare circumstances (for example, multipart version numbers), you may need to use underscores to separate adjacent numbers, since numbers do not have upper and lower case variants. 

so under scores are "very rare" allowance, but still allowance, allowance for all entities that can be named. But from style guide we can see that it is not allowed for: packages, Type variable names.

We can not distingush what is valid  number, so "guava123" version and "guava1_23" and "guava1_2_3" are all valid.
So regexp unlikely help us.
Checkstyle role is to highlight: 
```
my_guava1_23 // violation because  "my" is not sentence(at least 2 words) 
myGuava1_13  // ok

myversion_OtherGuava1_23 // violation, because second sentence (after "_") should start from lowercase
myVersion_OtherGuava1_23 // violation 
myVersion_otherGuava1_23 // ok 

myVersion_otherGuava_1_23 // violation
myVersion_otherGuava1_23 // ok 

myVersion_otherGuava1_23_ // violation, because "_" is not binding sentences or numbers
myVersion_otherGuava1_23 // ok 
```

in other words `_` should be special sumbol that attach correct Camel Case sentences or numbers, but numbers with "_" should be  specially attached to words.

please add all of such example to Inputs and please follow this pattern of bad/good  code.
if Checkstyle currently used Checks put no violation or extra violation, please use "ok until" "violation until"
and in separate issues we will create new modules to substitute regexp based Checks to specially crafted checks for google style.

please be advocate of users who use google style, do give Checks excuses. If Check is not working, it should be substituted to new module.

mohitsatr: Based on https://github.com/checkstyle/checkstyle/pull/17708?notification_referrer_id=NT_kwDOBVLGlrQxODU2MjUxODYyNjo4OTMxMDg3MA#discussion_r2336696506:

Underscores are allowed in the following cases, with the condition that second sentence (after "_") should start with lowercase:

1. when numbers are attached to words(Multipart version number format)
```
myGuava1_13 // ok
myVersion_otherGuava1_23 // ok
```

2. In test method names 

```
@Test
public void myVersion_otherGuava {...} // ok

```

They are not allowed in normal method names.

romani: Ok, so underscore only for Test methods or part of number that in human mind contains "."  Or any other symbols that java compiler does not allow in names and that is why "_" is used instead.

so let's put to our examples as clear picture of  good and bad. 
If something can not be covered by Regexp based Checks , we will postpone to separate issue or issues.

Please update inputs to draw all use cases  and see behavior. For complicated to spot problematic place , please not use violation message and use free explanation comment

romani: Ok, so underscore only for Test methods or part of number that in human mind contains "."  Or any other symbols that java compiler does not allow in names and that is why "_" is used instead.

so let's put to our examples as clear picture of  good and bad. 
If something can not be covered by Regexp based Checks , we will postpone to separate issue or issues.

Please update inputs to draw all use cases  and see behavior. For complicated to spot problematic place , please not use violation message and use free explanation comment

Zopsss:
> Underscores are only allowed in test method. we can't put them in normal Method names. Normal methods & Test methods, both use same regex format. so we allow the following formats, we are also making underscores valid for normal functions.

From style guide: https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

> Method names are written in [lowerCamelCase](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case).
>....
> Underscores may appear in JUnit test method names to separate logical components of the name, with each component written in [lowerCamelCase](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case), for example transferMoney_deductsFromSource. There is no One Correct Way to name test methods.


from lowerCamelCase rule:

> In very rare circumstances (for example, multipart version numbers), you may need to use underscores to separate adjacent numbers, since numbers do not have upper and lower case variants. 


In the style guide it is mentioned that method names should be in lower camel case ( which allows for underscores ) and it is also separately mentioned that JUnit test methods can also use underscores. 


So it is a bit confusing, if normal methods are allowed to have underscores or not. It is mentioned that `Method names are written in lowerCamelCase` which allows for underscores, so I think we should also allow for it? @romani @mohitsatr please share your opinion.

mohitsatr:
@Zopsss

> In very rare circumstances (for example, multipart version numbers),

Let's not think of Multipart version number methods as normal methods. https://github.com/checkstyle/checkstyle/pull/17708#discussion_r2336845872

Zopsss:
Style guide has mentioned that normal methods can use lowerCamelCase, which means it can have underscores. @romani please correct me if I'm wrong.

romani:
Unfortunately Google style is not exact.
From one side they say that `_`  can be used in test methods to combine sentences. From another side in general camel case explanation, `_` is allowed for numbers in all names(except package and generic names).

So if we can not write Regexp that decently covers allowance of `_` before digits. We need to allow `_` and update coverage table that false negatives to catch illegal underscore will be done when we finish new Check creation, might be not this year.

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -1,37 +1,134 @@
...
-  private final String gradle8_5_1 = "Gradle 8.5.1"; // false-positive until #17507
-  // violation above 'Member name 'gradle8_5_1' must match pattern'
+  private final String gradle8_5_1 = "Gradle 8.5.1";
 
   private final String gradle851 = "Gradle 8.5.1";
 
-  void guava34_4_6() {} // false-positive until #17507
-  // violation above 'Method name 'guava34_4_6' must match pattern'
+  class InnerBad {
+    String guava_33_4_6 = "guava_33_4_6"; // violation 'name 'guava_33_4_6' must match pattern'
```
romani: please put above or under each line with violation, expected variant of name

please do not use message of violation , but use free text explanation  
```
String guava_33_4_6 = "guava_33_4_6"; // violation as first digit should not have `_` in front 
String guava33_4_6 = "guava_33_4_6";
```

please do it for all.

romani:
For methods: allow underscore Only between sentences for tests and for all methods between digits.
Underscore is valid separator of digits in all names, even in class names

Zopsss:
Please try above one. If we're not able to implement it then please mark methods as false-negatives. We'll fix this in separate PR. @mohitsatr

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -1,37 +1,129 @@
...
+    void jdk9_0_392() {}
+
+    void kotlin_lang1_9_2() {} // false-positive, no way differentiate 'kotlinlang' &'kotlin_lang'
+    void kotlinLang1_9_2() {} // correct form, false positive
+
+    void jdk_method8_90() {} // violation, no way differentiate 'jdk_method' and 'jdkmethod'
+    void jdkMethod8_90() {} // correct form, false positive
+
+    void guava_version33_4_6() {} // violation,no way differentiate 'guavaversion' and 'guava_version'
+    void guavaVersion33_4_6() {} // correct form, false positive
```
mohitsatr: false-positive, current regex does not allow name to be in camelCase with a underscore

Zopsss: Please fix it. Try AI to get the needed regex

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -1,37 +1,129 @@
...
-  // violation above 'Member name 'gradle8_5_1' must match pattern'
+  private final String gradle8_5_1 = "Gradle 8.5.1";
 
   private final String gradle851 = "Gradle 8.5.1";
 
-  void guava34_4_6() {} // false-positive until #17507
-  // violation above 'Method name 'guava34_4_6' must match pattern'
+  class InnerBad {
+    String guava_33_4_6 = "guava_33_4_6"; // violation some thing something
+    String guava33_4_6 = "guava33_4_6"; // violation some thing something
```
romani:
```
    String guava_33_4_6 = "guava_33_4_6"; // violation due to '_' between digit and letter
    String guava33_4_6 = "guava33_4_6"; // violation false-positive
```

please apply this to all other

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -1,37 +1,129 @@
...
+
+    // violation below 'Member name 'guava_version33_4_6' must match pattern'
+    String guava_version33_4_6 = "guava_version33_4_6"; // false-positive
```
romani:
```

    String guava_version33_4_6 = "guava_version33_4_6"; // violation , because '_' between single words, not a sentence/phrase.
```

just trailing comment, we agreed to not use violation messages and use explanation instead

romani:
```
    String guavaSomtheing_version33_4_6; // violation , false-positive, first part, before '_' is sentence so ok to have '_'
```

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -1,37 +1,129 @@
...
+
+    String guava33_4_6_ = "guava33_4_6_"; // violation 'name 'guava33_4_6_' must match pattern'
+    String guava33_4_6 = "guava33_4_6";
+
+    String jdk_8_90 = "jdk_8_90"; // violation 'Member name 'jdk_8_90' must match pattern'
+    String jdk8_90 = "jdk8_90";
```
romani: Whay is value in  string value ? can we remove it to give us more horizontal space

romani (changes requested):
items:

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
mohitsatr: done

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
mohitsatr: done

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
mohitsatr: done

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -1,37 +1,124 @@
...
+
+  private final String gradle8_5_1 = "";
+
+  private final String gradle851 = "";
+
+  class InnerBad {
+    String guava_33_4_6 = ""; // violation due to '_' between digit and letter
+    String guava33_4_6 = "";
+
+    String guava_version33_4_6 = ""; // false-positive
```
romani: Do you mean false negative ?
False negatives means "no violation, but it should be".
Underscore can not be placed between sequence of characters that starts with lower case and has only lower case characters.
I doubt regexp can handle this complexity. New Check is required to do all as Google wants.

If you agree, we can create issue on this and accept fact that we have gap in coverage on such naming until that issue

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -1,37 +1,124 @@
...
+
+    String guava33_4_6_ = ""; // violation, underscore at the end
+    String guava33_4_6 = "";
+
+    String jdk_8_90 = ""; // violation due to '_' between digit and letter
+    String jdk8_90 = "";
+
+    String jdk_version8_0_393 = ""; // false-positive
+    // no way to differentiate 'jdk_verison' and 'jdkversion'
```
romani: Same as above

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -1,37 +1,124 @@
...
+
+    String jdk8_91_ = ""; // violation, underscore at the end
+    String jdk8_91 = "";
+
+    String kotlin_1_9_24 = ""; // violation due to '_' between digit and letter
+    String kotlin1_9_24 = "";
+
+    String kotlin_version1_9_24 = ""; // false-positive,
+    // not way to differentiate 'kolin_version' and 'kotlinversion'
```
romani: Same as above

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
```diff
@@ -1,37 +1,124 @@
...
+
+    void gradle_9_5_1() {} // violation due to '_' between digit and letter
+    void gradle9_5_1() {}
+
+    void jdk_9_0_392() {} // violation due to '_' between digit and letter
+    void jdk9_0_392() {}
+
+    void kotlin_lang1_9_2() {} // false-positive:
+    // no way to differentiate 'kotlinlang' & 'kotlin_lang'
+    void kotlinLan1_9_2() {} // correct form but still false positive
```
romani: This is correct naming, no violations is expected, as there is no violation comment , it means I there is no violation.
As this is correct, please remove trailing comment

romani: Please update all other .

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
mohitsatr: No, it is false-positive "there is violation, but it should not be". as per the regex in this PR.

> Underscore can not be placed between sequence of characters that starts with lower case and has only lower case characters.

I don't even know if `guava_version33_4_6` is correct as Google wants. Numbering suffix(`33_4_6`) is correct. We cannot support underscore between letters because of regex limitations.

romani: Please put violation comment.
I don't know why system allows you to avoid it

mohitsatr: I think there is some confusion here:

Is `guava_version33_4_6` valid name as per Google? Because if it is, then the current regex logs violation on it, so false-positive.

> Underscore can not be placed between sequence of characters that starts with lower case and has only lower case characters

if it is not valid, then regex is working correctly, and I can remove false-positive.

romani: > is guava_version33_4_6 valid name as per Google?

Not valid.
Capital word is expected before underscore.
`guavaSomerhing_versionSomathing33_4_6` as _ is concatenation of sentences.
`guavaSomething_version33_4_6` might be ok, we can make it as ok and wait for user feedback of style clarification.

romani:
@mohitsatr , please rebase to see more clearly what is left to be addressed

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
```diff
@@ -22,95 +22,102 @@ public class InputUnderscoreUsedInNames {
...
-    String kotlin_1_9_24 = ""; // violation due to '_' between digit and letter
+    String kotlin_1_9_24 = ""; // violation 'Member name 'kotlin_1_9_24' must match pattern'
     String kotlin1_9_24 = "";
 
-    // violation below, '_' not allowed between lowercase character sequences
+    // violation below 'Member name 'kotlin_version1_9_24' must match pattern'
     String kotlin_version1_9_24 = "";
 
-    String kotlinVersion1_9_24 = "";
+    String kotlinVersion1_9_24 = "";  // violation 'name 'kotlinVersion1_9_24' must match pattern'
```
mohitsatr: false-positive

src/main/resources/google_checks.xml
```diff
@@ -245,7 +245,8 @@
              value="Type name ''{0}'' must match pattern ''{1}''."/>
     </module>
     <module name="MemberName">
-      <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9]*$"/>
+      <property name="format"
+        value="^(?![a-z]$)(?![a-z][A-Z])[a-z][a-z0-9]*(?:[A-Z][a-zA-Z0-9]*|[0-9](?:_[0-9]+)*)?$"/>
```
mohitsatr: .

src/main/resources/google_checks.xml
```diff
@@ -424,7 +425,8 @@
       <property name="excludeScope" value="nothing"/>
     </module>
     <module name="MethodName">
-      <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9]*$"/>
+      <property name="format"
+          value="^[a-z](?:[a-z0-9][a-zA-Z0-9]*|[a-z0-9]*[0-9](?:_[0-9]+)*)$"/>
```
mohitsatr: .

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
```diff
@@ -22,95 +22,102 @@ public class InputUnderscoreUsedInNames {
   private final String gradle851 = "";
 
   class InnerBad {
-    String guava_33_4_7 = ""; // violation due to '_' between digit and letter
+    String guava_33_4_7 = ""; // violation 'Member name 'guava_33_4_7' must match pattern'
```
romani: we already discussed this few times in this PR.
there is not much value in violation message in such complicated style violation. messages is same.
It is better to keep reason on why there is violation.

you can keep both 
https://github.com/checkstyle/checkstyle/blob/f486dcacb9b6096bd28ce9b6b3d215f922f315e7/src/test/java/com/puppycrawl/tools/checkstyle/bdd/InlineConfigParser.java#L82

`String guava_33_4_7; // violation, due to '_' between digit and letter 'must match pattern'`

Please apply to all.

if we need more space you and make type to be `int`.

romani (changes requested):
item:

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
```diff
@@ -22,95 +22,102 @@ public class InputUnderscoreUsedInNames {
...
+    void guava_33_4_5_() {} // violation 'Method name 'guava_33_4_5_' must match pattern'
 
     void guava33_4_5() {}
   }
 
   class UnderscoresInNonTestMethods {
 
-    void testSetCount_zeroToZero_addSupported() {} // violation, normal method must not use '_'
+    // violation below 'Method name 'testSetCount_zeroToZero_addSupported' must match pattern'
+    void testSetCount_zeroToZero_addSupported() {}
```
romani, line 100: this and all below are false positives.

romani (changes requested):
sounds like Regexp is not possible/reasonable to use in such complicated requirements, right ?
if you agree, lets write issue(s) on new Check(s) creation to cover requirements and finish with this PR. We did big effort, and we proved that regexp is not working any more.

items:

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputCamelCaseMultipartVersioningNames.java
mohitsatr: done

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
mohitsatr: done

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
mohitsatr: how is this false-positive? 
Name does not have any numbering as suffix(`22_1_1`)  which means it's normal method and normal method names cannot contain `_`. 
Violation is correct.

mohitsatr: New check should first determine the type of name - regular or name with numbering and then it should validate the name against the appropriate regex format, which will differ for each type of naming.

romani: oh , yes, not Test annotation, so `_` not allowed.
please add human explanation text for violation instead of long message

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
```diff
@@ -22,95 +22,100 @@ public class InputUnderscoreUsedInNames {
   private final String gradle851 = "";
 
   class InnerBad {
-    String guava_33_4_7 = ""; // violation due to '_' between digit and letter
-    String guava33_4_7 = "";
+    int guava_33_4_7 = 0; // violation, due to '_' between digit and letter, 'must match pattern'
+    int guava33_4_7 = 0;
```
romani: can we remove `=0`? to avoid "noise" that is not relevant to validaiton.
please do for all

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
romani: ok , you already did. 
resolved.

romani (changes requested):
last:

romani:
Github, generate report

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
mohitsatr: done

romani:
```
[1;31mERROR[m] COMPILATION ERROR : 
[[1;34mINFO[m] -------------------------------------------------------------
[[1;31mERROR[m] /home/circleci/project/src/it/resources/com/google/checkstyle/test/
chapter5naming/rule53camelcase/
InputUnderscoreUsedInNames.java:[8,24] 
variable guava33_4_6 not initialized in the default constructor

```

We need to restore default values.
Or better remove "final" from all of them

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
```diff
@@ -5,112 +5,117 @@
...
+    int kotlin_1_9_24; // violation, due to '_' between digit and letter, 'must match pattern'
+    int kotlin1_9_24;
 
-    // violation below, '_' not allowed between lowercase character sequences
-    String kotlin_version1_9_24 = "";
+    // violation below, underscore not allowed between lowercase character sequences, 'pattern'
+    int kotlin_version1_9_24;
 
-    String kotlinVersion1_9_24 = "";
+    int kotlinVersion1_9_24; // violation, false-positive, 'must match pattern'
```
romani: so this is the only false-positive, good.
Bad part of it, users will be blocked by this correct naming, on most correct naming - camel case + number.

please add same style naming for method `convertToKotlinVersion1_9_24`
and empty class `class CoverterToKotlinVersion1_9_24 {}`.

I think it is doable to make suppression to allow this pattern, there might be more false-negatives , but it is ok, we will fully cover this by special Checks in future.

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
```diff
@@ -5,112 +5,117 @@
...
-  private final String gradle8_5_1 = "";
+  private String gradle8_5_1;
 
-  private final String gradle851 = "";
+  private String gradle851;
 
   class InnerBad {
-    String guava_33_4_7 = ""; // violation due to '_' between digit and letter
-    String guava33_4_7 = "";
+    int guava_33_4_7; // violation, due to '_' between digit and letter, 'must match pattern'
```
romani: ... due to '_' between digit and letter ...
vs
... underscore at the end ...


I think we can do simple `... due to _ between digit and letter ...`, without single quotes.
Example:
` int guava_33_4_7; // violation, due to _ between digit and letter, 'must match pattern'`
` int guava33_4_8_; // violation, _ at the end, 'must match pattern'`

please unify all such cases.

romani (changes requested):
item:

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
mohitsatr: done

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
mohitsatr: done.

I'll try to cover class names through Regex. It might be easier than method names.

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
```diff
@@ -5,112 +5,119 @@
...
 
-    // violation below, '_' not allowed between lowercase character sequences
+    // violation below, _ not allowed between lowercase character sequences, 'pattern'
     void kotlin_lang1_9_2() {}
 
-    void kotlinLang1_9_2() {}
+    void kotlinLang1_9_2() {} // violation, false-positive, 'must match pattern'
 
-    // violation below, '_' not allowed between lowercase character sequences
+    void convertToKotlinVersion1_9_24() {} // violation, false-positive, 'must match pattern'
```
romani: Ok, we need some suppression to avoid false positives.
Suppression config and resulted false negatives should be marked until bsime ticket, please create it, 
And we will merge.

romani (changes requested):
Last

src/it/resources/com/google/checkstyle/test/chapter5naming/rule53camelcase/InputUnderscoreUsedInNames.java
mohitsatr: false-positive for class names will be handled at #17835

mohitsatr:
Github, generate report

romani:
>  class CoverterToKotlinVersion1_9_24 {} // violation, false-positive, 'must match pattern'

was removed. as it was addressed in https://github.com/checkstyle/checkstyle/pull/17835

romani:
basic implementation is merged
all false negatves will be addressed in next issue

Issue #17299, opened 2025-06-30
Convert eligible classes to records


AmitKumarDeoghoria:
From comment : https://github.com/checkstyle/checkstyle/pull/17259#issuecomment-3019192404
During the JDK 17 upgrade (#17168), we've identified multiple classes that can be converted to Java records. To keep the initial migration focused and manageable, we're temporarily disabling these inspections. This follow-up issue tracks the conversion of these classes to leverage JDK 17's record feature for cleaner, more maintainable code.

Any classes in "api" package should not be changed!!!

```
<problems is_local_tool="true">
<problem>
<file>file://$PROJECT_DIR$/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalkerAuditEvent.java</file>
<line>30</line>
<module>project</module>
<package>com.puppycrawl.tools.checkstyle</package>
<entry_point TYPE="class" FQNAME="com.puppycrawl.tools.checkstyle.TreeWalkerAuditEvent"/>
<problem_class id="ClassCanBeRecord" severity="WEAK WARNING" attribute_key="INFO_ATTRIBUTES">Class can be a record</problem_class>
<description>Class can be a record</description>
<highlighted_element>TreeWalkerAuditEvent</highlighted_element>
<language>JAVA</language>
<offset>13</offset>
<length>20</length>
</problem>
<problem>
<file>file://$PROJECT_DIR$/src/main/java/com/puppycrawl/tools/checkstyle/api/LineColumn.java</file>
<line>28</line>
<module>project</module>
<package>com.puppycrawl.tools.checkstyle.api</package>
<entry_point TYPE="class" FQNAME="com.puppycrawl.tools.checkstyle.api.LineColumn"/>
<problem_class id="ClassCanBeRecord" severity="WEAK WARNING" attribute_key="INFO_ATTRIBUTES">Class can be a record</problem_class>
<description>Class can be a record</description>
<highlighted_element>LineColumn</highlighted_element>
<language>JAVA</language>
<offset>13</offset>
<length>10</length>
</problem>
<problem>
<file>file://$PROJECT_DIR$/src/main/java/com/puppycrawl/tools/checkstyle/site/AllCheckSummaries.java</file>
<line>839</line>
<module>project</module>
<package>com.puppycrawl.tools.checkstyle.site</package>
<entry_point TYPE="class" FQNAME="com.puppycrawl.tools.checkstyle.site.AllCheckSummaries.CheckInfo"/>
[...]

PR #17709, opened 2025-08-29
Issue #17299: Convert eligible classes to records 2

AmitKumarDeoghoria:
part of : #17299

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/LineColumn.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/LineColumn.java
@@ -19,29 +19,21 @@
 
 package com.puppycrawl.tools.checkstyle.api;
 
-import java.util.Objects;
-
 /**
  * Immutable line and column numbers.
  *
+ * @param line   The one-based line number.
+ * @param column The zero-based column number.
  */
-public class LineColumn implements Comparable<LineColumn> {
-
-    /** The one-based line number. */
-    private final int line;
-
-    /** The zero-based column number. */
-    private final int column;
+public record LineColumn(int line, int column) implements Comparable<LineColumn> {
 
     /**
      * Constructs a new pair of line and column numbers.
      *
-     * @param line the one-based line number
+     * @param line   the one-based line number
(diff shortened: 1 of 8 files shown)


AmitKumarDeoghoria:
These conversions are all depended on each other, change in one requires changes in other.

src/main/java/com/puppycrawl/tools/checkstyle/api/LineColumn.java
```diff
@@ -19,47 +19,21 @@
...
+ * @param column The zero-based column number.
  */
-public class LineColumn implements Comparable<LineColumn> {
-
-    /** The one-based line number. */
-    private final int line;
-
-    /** The zero-based column number. */
-    private final int column;
+public record LineColumn(int line, int column) implements Comparable<LineColumn> {
```
romani, line 28: This is API class , let's not change it to avoid any problems with bytecode compatibility.

Just add suppression on it by javadoc type.

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/api/LineColumn.java
AmitKumarDeoghoria: If we put suppression on it we have to put suppression in all the other classes to in the file too. As I mentioned they all have dependency on each other. Let me know what to do suppress all or go ahead at risk.

AmitKumarDeoghoria: what should we do @romani

romani: Do you mean that inspection put violation on all code that you changed in this PR?
Even root reason in LineColumn class only.

It is pretty weird.
Please revert to let me see violations.

In all cases we can not change class to record, it might cause a lot of problems to users.

romani:
I am closing this PR, we should avoid such updates in API classes as it might cause breaking compatibility problems for users

PR #17713, opened 2025-08-30
Issue #17269: Remove Extra Installation of Groovy from Error Prone

AmitKumarDeoghoria:
part of : #17269

### DIFF
--- a/.github/workflows/error-prone.yml
+++ b/.github/workflows/error-prone.yml
@@ -26,7 +26,7 @@ jobs:
         run: |
           curl --fail-with-body -s "https://get.sdkman.io" | bash
           source "$HOME/.sdkman/bin/sdkman-init.sh"
-          sdk install groovy 3.0.17
+          sdk install groovy 5.0.0
           echo "GROOVY_HOME=$HOME/.sdkman/candidates/groovy/current" >> "$GITHUB_ENV"
           echo "$HOME/.sdkman/candidates/groovy/current/bin" >> "$GITHUB_PATH"
 
--- a/.ci/error-prone-check.groovy
+++ b/.ci/error-prone-check.groovy
@@ -2,11 +2,11 @@ import groovy.transform.EqualsAndHashCode
 import groovy.transform.Field
 import groovy.transform.Immutable
 import groovy.xml.XmlUtil
+import groovy.xml.XmlParser
 
 import java.util.regex.Matcher
 import java.util.regex.Pattern
 
-@Field static final Set<String> PROFILES = Set.of("compile", "test-compile")
 @Field static final String USAGE_STRING = "Usage groovy " +
         ".${File.separator}.ci${File.separator}error-prone-check.groovy" +
         " (compile | test-compile) [-g | --generate-suppression]\n"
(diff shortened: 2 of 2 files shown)


AmitKumarDeoghoria:
but @romani apt version is using old groovy only, why we are relying on that? If we want to use new groovy shouldn't we just kept it using SDKman?

romani:
We should not rely on apt.
Install modern version as it supposedly to be done now a day.

AmitKumarDeoghoria:
@romani  then I dont think we need any change, it already Installing groovy with SDKMAN instead of apt, and we have to mention version because of the Issues with the latest version.

romani:
Let's face issue with latest groovy, we will find way to resolve them.
Life is going forward, we can not stay in old legacy, it is death way to stay.

romani:
```
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
/home/runner/work/checkstyle/checkstyle/.ci/error-prone-check.groovy: 58: 
unable to resolve class XmlParser
 @ line 58, column 21.
       final XmlParser xmlParser = new XmlParser()
                       ^

/home/runner/work/checkstyle/checkstyle/.ci/error-prone-check.groovy: 58: 
unable to resolve class XmlParser
 @ line 58, column 33.
       final XmlParser xmlParser = new XmlParser()
                                   ^

2 errors
```

https://github.com/checkstyle/checkstyle/blob/710ea4cc6eb6d098387d3eccd3cf7b422bf54122/.ci/error-prone-check.groovy#L58-L59

https://docs.groovy-lang.org/latest/html/api/groovy/xml/XmlParser.html
```
Class XmlParser
package: groovy.xml.XmlParser
```

from AI:
`Missing Groovy XML Library:
The XmlParser class is part of the groovy-xml library. If you are running Groovy in an environment that doesn't automatically include this library (e.g., a standalone script, a specific application like Jenkins or Boomi), you might need to add the groovy-xml.jar file to your project's classpath or the environment's library path.`

if library is not accesible by simple import, use grape https://docs.groovy-lang.org/latest/html/documentation/grape.html

romani:
@AmitKumarDeoghoria , please use special import for Class XmlParser
` import groovy.xml.XmlParser`

.github/workflows/error-prone.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         run: |
           curl --fail-with-body -s "https://get.sdkman.io" | bash
           source "$HOME/.sdkman/bin/sdkman-init.sh"
-          sdk install groovy 3.0.17
+          sdk install groovy
```
romani: lets keep a version, it is very explicit way to see what version CI is using.
No surprises, stable behavior until next upgrade.

romani:
https://groovy.apache.org/download.html , groovy 5 is out !!! lets use it !!!

AmitKumarDeoghoria:
Screenshot 2025-09-01 at 10 06 28 PM
We are at Groovy 5.0.0

romani (approved):
Ok to merge if CI pass

romani:
This is first good step in final direction.

More to update https://github.com/search?q=repo%3Acheckstyle%2Fcheckstyle%20Groovy&type=code

In separate PRs

PR #17739, opened 2025-09-04
Issue #16361: Update SarifLoggerTest to use verifyWithInlineConfigPar…

shaikhu:
Issue: #16361 Update SarifLoggerTest to use verifyWithInlineConfigParser


romani (approved):
thanks a lot

romani:
@shaikhu , direction is good, can fix all other in such test Class ?

shaikhu:
> @shaikhu , direction is good, can fix all other in such test Class ?

Probably not all :sweat_smile:, but happy to try.

PR #17740, opened 2025-09-04
Issue #17642: enable javadoc warning but disable missing

### LINKED ISSUE #17642: Fix Javadoc Warnings: "use of default constructor, which does not provide a comment"
From Comment : https://github.com/checkstyle/checkstyle/pull/17630#discussion_r2280708336
When running the maven-site-plugin (with maven-javadoc-plugin:3.11.2), the build currently fails because of 100+ Javadoc warnings related to classes that do not have documented constructors.

These warnings block mvn site execution and need to be resolved.

hundred of:
```
[WARNING] Javadoc Warnings
[WARNING] /home/circleci/project/src/main/java/com/puppycrawl/tools/checkstyle/
  checks/naming/AbbreviationAsWordInNameCheck.java:146: 
  warning: use of default constructor, which does not provide a comment
[WARNING] public class AbbreviationAsWordInNameCheck extends AbstractCheck {
[WARNING] ^
```
and single 
```
[WARNING] Note: Custom tags that could override future standard tags:
    @noinspectionreason, @noinspection. To avoid potential overrides, 
    use at least one period character (.) in custom tag names.
[WARNING] Note: Custom tags that were not seen:  @noinspectionreason, @noinspection
[WARNING] 100 warnings
```

`[WARNING] /home/circleci/project/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/AbbreviationAsWordInNameCheck.java:146: warning: use of default constructor, which does not provide a comment
[WARNING] public class AbbreviationAsWordInNameCheck extends AbstractCheck {
[WARNING] ^
[WARNING] /home/circleci/project/src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java:64: warning: use of default constructor, which does not provide a comment
[WARNING] public abstract class AbstractAutomaticBean
[WARNING] ^
...
[WARNING] Note: Custom tags that could override future standard tags:  @noinspectionreason, @noinspection.
[WARNING] 100 warnings
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.21.0:site ...
`

CI Error Link : https://app.circleci.com/pipelines/github/checkstyle/checkstyle/36005/workflows/de45df89-f2ed-43a4-aa38-ec36e77f3e33/jobs/1042842

**Root Cause**

The Javadoc plugin complains whenever:
- A public or protected class has an implicit or explicit default constructor but lacks Javadoc.
- Nested static/inner classes also trigger this if their constructors are not documented.

Since many Checkstyle checks and utility classes rely on default constructors, these warnings accumulate and break the site phase.

Some details on fact that compiler add this constructor:
https://stackoverflow.com/a/17709063

**Proposed Solution**

Add minimal Javadoc comments for default constructors.

romani:
fixes #17642

based on https://stackoverflow.com/a/71785465
https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#doclint
https://docs.oracle.com/en/java/javase/17/docs/specs/man/javadoc.html#additional-options-provided-by-the-standard-doclet

>missing: Checks for missing documentation comments or tags (for example, a missing comment or class, or a missing `@return` tag or similar tag on a method).

it is not ideal fix, but we enable warning and disable specific validations, we disabling a bit more than required, but it is limitation of javadoc tool

### DIFF
--- a/config/jsoref-spellchecker/whitelist.words
+++ b/config/jsoref-spellchecker/whitelist.words
@@ -330,6 +330,7 @@ Dnew
 doccheck
 Dockter
 doclet
+doclint
 doctype
 documentationcomments
 donotignoreexceptions
--- a/pom.xml
+++ b/pom.xml
@@ -553,8 +553,8 @@
             <source>${java.version}</source>
             <show>private</show>
             <failOnError>true</failOnError>
-            <!-- disabled till https://github.com/checkstyle/checkstyle/issues/17642 -->
-            <failOnWarnings>false</failOnWarnings>
+            <failOnWarnings>true</failOnWarnings>
+            <doclint>-missing</doclint>
             <linksource>true</linksource>
             <tags>
               <tag>


AmitKumarDeoghoria:
can we integrate require-javadoc plugin ?

romani:
we do not need to , we have our Checks for this.

PR #17769, opened 2025-09-09
Issue #13626: Simplified metrics and naming modules javadocs

SteLeo1602:
Issue #13626


src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/ClassFanOutComplexityCheck.java
```diff
@@ -141,4 +96,44 @@ protected String getLogMessageId() {
...
+    /**
+     * Setter to specify user-configured regular expressions to ignore classes.
+     *
+     * @param from array representing regular expressions of classes to ignore.
+     * @propertySince 7.7
+     * @noinspection RedundantMethodOverride
+     * @noinspectionreason Display module's unique property version
+     */
+    @Override
+    public void setExcludeClassesRegexps(Pattern... from) {
```
romani, line 108: please put final on all newly override methods.
and violation from inspection might be gone.

romani (changes requested):
item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/metrics/ClassFanOutComplexityCheck.java
SteLeo1602: Can't place `final` since the class is already final. Removed the inspection though, will see if it works

SteLeo1602: Doesn't work, so left it as is

Issue #17487, opened 2025-07-28
Add `rewrite` support for `errorprone.refasterrules`


Pankraz76:
## Add `rewrite` support for `errorprone.refasterrules`

I’d like to propose integrating [Google’s Error Prone](https://github.com/google/error-prone) and its Picnic extension (demonstrated in [Automating Away Bugs with Error Prone | PlatformCon 2023](https://www.youtube.com/watch?v=6llnrUtVlrE)) to enable automated bug fixes via `rewrite` rules. This would complement Checkstyle’s static analysis capabilities by addressing semantic bugs rather than stylistic issues.

## Motivation
Error Prone’s `refaster`/`rewrite` rules can automatically:
- Fix common bug patterns (e.g., `String.equals()` misuse)
- Modernize code (e.g., JDK migration helpers)
- Enforce best practices (e.g., null-check improvements)

Real-world adoptions show tangible benefits:
- [Apache Kafka’s integration](https://github.com/apache/kafka/pull/20219)
- [Spotless’s implementation](https://github.com/diffplug/spotless/pull/2576)

## Proposal
1. Add support for `errorprone.refasterrules`-based rewrites
2. Implement with opt-in adoption (no breaking changes)
3. Include suppression mechanisms for API constraints

## Discussion Points
- Need consensus on:
  - Scope of auto-fixes
  - Preferred suppression strategy
  - Integration approach

## Next Steps
I’m happy to:
- Prepare a PoC demonstrating the value
- Collaborate on implementation strategy
- Address any concerns about compatibility

relates to:

- https://github.com/checkstyle/checkstyle/issues/16543
- https://github.com/checkstyle/checkstyle/issues/17732
- https://github.com/checkstyle/checkstyle/issues/17717
- https://github.com/checkstyle/checkstyle/pull/17741

```
  - org.openrewrite.java.RemoveUnusedImports
  - org.openrewrite.java.format.RemoveTrailingWhitespace
  - org.openrewrite.java.migrate.Java8toJava11
  - org.openrewrite.staticanalysis.LowercasePackage
  - org.openrewrite.staticanalysis.MissingOverrideAnnotation
  - org.openrewrite.staticanalysis.ModifierOrder
  - org.openrewrite.staticanalysis.NoFinalizer
  - org.openrewrite.staticanalysis.RemoveUnusedLocalVariables
  - org.openrewrite.staticanalysis.RemoveUnusedPrivateFields
[...]

PR #17792, opened 2025-09-15
Issue #17487: Add `refasterrules.OptionalRulesRecipes`

Pankraz76:
Issue #17487: Add `refasterrules.OptionalRulesRecipes`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -831,6 +831,9 @@
         <configuration>
           <compilerArgs>
             <arg>-Xpkginfo:always</arg>
+            <arg>-XepOpt:Refaster:NamePattern=^(?!OptionalRules\$OptionalOrElseThrow).*</arg>
+            <arg>-XepOpt:Refaster:NamePattern=^(?!OptionalRules\$OptionalOrElseThrowMethodReference).*</arg>
+            <arg>-XepOpt:Refaster:NamePattern=^(?!OptionalRules\$OrOrElseThrow).*</arg>
           </compilerArgs>
         </configuration>
       </plugin>
@@ -5348,6 +5351,7 @@
                 <recipe>tech.picnic.errorprone.refasterrules.ClassRulesRecipes</recipe>
                 <recipe>tech.picnic.errorprone.refasterrules.FileRulesRecipes</recipe>
                 <recipe>tech.picnic.errorprone.refasterrules.MicrometerRulesRecipes</recipe>
+                <recipe>tech.picnic.errorprone.refasterrules.OptionalRulesRecipes</recipe>
                 <recipe>tech.picnic.errorprone.refasterrules.StreamRulesRecipes</recipe>
                 <recipe>tech.picnic.errorprone.refasterrules.TimeRulesRecipes</recipe>
               </activeRecipes>


pom.xml
```diff
@@ -5346,7 +5346,13 @@
                 <recipe>tech.picnic.errorprone.refasterrules.BigDecimalRulesRecipes</recipe>
                 <recipe>tech.picnic.errorprone.refasterrules.CharSequenceRulesRecipes</recipe>
                 <recipe>tech.picnic.errorprone.refasterrules.ClassRulesRecipes</recipe>
+                <recipe>tech.picnic.errorprone.refasterrules.CollectionRulesRecipes</recipe>
+                <recipe>tech.picnic.errorprone.refasterrules.ComparatorRulesRecipes</recipe>
```
Pankraz76: sould we extract other unrealted reciped?

Assuming we need to extend build timeout soon.

romani: We we reach limit, we should split recipes by provider and use different profiles for each recipe group. CI execution should stay under 10 minutes.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
```diff
@@ -159,7 +159,7 @@ public void visitJavadocToken(DetailNode ast) {
         final Optional<DetailNode> inlineTagNode = getInlineTagNode(ast);
         boolean shouldValidateUntaggedSummary = true;
         if (inlineTagNode.isPresent()) {
-            final DetailNode node = inlineTagNode.get();
+            final DetailNode node = inlineTagNode.orElseThrow();
```
romani: I am not fun of this API .
I my opinion it doesn't help much , but a way more verbose and confusing in reading.

@rdiachenko , what is your opinion or thoughts on this?

romani (changes requested):
Items

pom.xml
Pankraz76: Yes, we’ll extract it in a dedicated PR.

On the other hand, it’s a bit expensive, since the second build will trigger additional parsing, which isn’t actually needed therefore wasting this effort.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java
Pankraz76: > and confusing in reading.

why confusing? It tells what happens, not like before hiding the exception surprise/bug.


Don’t like it either, but it still seems to be the de facto convention, since IntelliJ IDEA and other tools suggest it as well.
  
Asuming the idea is to promote a clear API, as `get` throwing an exception is a bit of a surprise—it’s not obvious and relies on prior knowledge.
  
  That’s not great software design, since we want to avoid surprises and instead aim for code that is clear and intentional.

rdiachenko: It's not black and white. I usually consider its usage case by case. From the exception standpoint there is no difference:

```
Exception in thread "main" java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.orElseThrow(Optional.java:377)
```

vs
```
Exception in thread "main" java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.get(Optional.java:143)
```

If no additional context to provide, I go with `.get()` since its less verbose, otherwise `.orElseThrow(() -> new SomeException("more context"))` is more helpful.

By the way, I don't like neither approach. Usually you want to handle it in a more proper way like `opt.ifPresent(value -> {....})`, and similar methods. Relying on `.get()/orElseThrow` throwing exception should be a rare case as to me.

romani: Thanks a lot.

@Pankraz76 , please disable this recipe with link to this discussion. In far future, next generation of maintainers might have different opinion. For now disabled.

pom.xml
Pankraz76: lets extract once we fail.

romani: yes, for now all at one place

pom.xml
```diff
@@ -831,6 +831,9 @@
         <configuration>
           <compilerArgs>
             <arg>-Xpkginfo:always</arg>
+            <arg>-XepOpt:Refaster:NamePattern=^(?!OptionalRules\$OptionalOrElseThrow).*</arg>
```
Pankraz76, line 834:
```
Caused by: org.codehaus.plexus.compiler.CompilerException: error: invalid flag: -XepOpt:Refaster:NamePattern=^(?!OptionalRules\$OptionalOrElseThrow).*
```

applied from template:


- https://error-prone.picnic.tech/refasterrules/OptionalRules/#ororelsethrow

romani: ok, future engineers will use blame to get to reason of exclusion and will find our conversation in PR

Pankraz76: yes but its not working currently, need to add link once its stable.

Pankraz76: can not make this work.

Pankraz76:
feel free to reopen and continue.

My goal was to use rewrite, not to rely on some random recipe.

Issue #13048, opened 2023-05-06
JavadocTagContinuationIndentation check doesn't support arbitrary indentation on &lt;pre&gt; tags


stechio:
I have read check documentation: https://checkstyle.org/config_javadoc.html#JavadocTagContinuationIndentation
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
$ javac Test.java

$ cat test-checkstyle.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name = "Checker">
  <module name="TreeWalker">
    <module name="JavadocTagContinuationIndentation"/>
  </module>
</module>

$ cat Test.java 
/**
 * Test class.
 *
 * @apiNote
 *          This is the predefined indentation applied by Eclipse formatter to api/impl taglets.
 *          <pre>
 * Since &lt;pre> blocks retain leading whitespace, in order to avoid extra-indentation from being
 * added by the javadoc renderer, its content herein has to stick to the left margin of the comment
 * area. Unfortunately, checkstyle seems not to properly evaluate the context (&lt;pre> block) of
 * this content, and reports an error:
 *   "JavadocTagContinuationIndentation: Line continuation have incorrect indentation level,
 *   expected level should be 4"</pre>
 */
public class Test {
  public static void main(String[] args) {
  }
}

$ java -jar checkstyle-10.10.0-all.jar -c test-checkstyle.xml Test.java
Starting audit...
[ERROR] Test.java:7: Line continuation have incorrect indentation level, expected level should be 4. [JavadocTagContinuationIndentation]
[ERROR] Test.java:8: Line continuation have incorrect indentation level, expected level should be 4. [JavadocTagContinuationIndentation]
. . .
[ERROR] Test.java:12: Line continuation have incorrect indentation level, expected level should be 4. [JavadocTagContinuationIndentation]
Audit done.
Checkstyle ends with 6 errors.
```
---
CURRENT BEHAVIOR
Since &lt;pre&gt; blocks retain leading whitespace, in order to avoid extra-indentation from being added by the javadoc renderer, I have to stick their contents to the left margin of the comment area (like in `Test.java` here above) — here it is the intended result:
[...]

PR #17809, opened 2025-09-18
doc: JavadocTagContinuationIndentation does not validate the lines inside pre tags

mohitsatr:
part of #13048

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagContinuationIndentationCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagContinuationIndentationCheck.java
@@ -35,6 +35,7 @@ import com.puppycrawl.tools.checkstyle.utils.JavadocUtil;
  * throws a violation. A continuation line is when the description starts/spans past the line with
  * the tag. Default indentation required is at least 4, but this can be changed with the help of
  * properties below.
+ * Note: This check does not validate the indentation of lines inside {@code pre} tags.
  * </div>
  *
  * @since 6.0
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadoctagcontinuationindentation/Example4.java
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadoctagcontinuationindentation/Example4.java
@@ -1,0 +1,36 @@
+/*xml
+<module name="Checker">
+  <module name="TreeWalker">
+    <module name="JavadocTagContinuationIndentation"/>
+  </module>
+</module>
+*/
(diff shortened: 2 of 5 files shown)


mohitsatr:
Github, generate website

romani:
https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=32503&view=logs&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111&l=1787 

you need to run `mvn clean test` on local and put all changed files to commit.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagContinuationIndentationCheck.java
```diff
@@ -35,6 +35,7 @@
  * throws a violation. A continuation line is when the description starts/spans past the line with
  * the tag. Default indentation required is at least 4, but this can be changed with the help of
  * properties below.
+ * Note: This check does not validate the indentation of lines inside {@code pre} tags.
```
romani, line 38: this update , as separate PR is not not valid.

https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTagContinuationIndentationCheck.java has no logic on PRE.

delivery of this update is separate PR/commit will simplify sizing of PRs but will make it more confusing in long run when people will question, WHY this Check is doing PRE in this way? the only way to get answers is to open history and be linked to issues and PRs to get answer.
That is why we keep all together, to make clear by history. Nobody should remember why something is done in certain way, system should do this.

Please move this update to PR where you add PRE.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadoctagcontinuationindentation/Example4.java
```diff
@@ -0,0 +1,36 @@
...
+*/
+package com.puppycrawl.tools.checkstyle.checks.javadoc.javadoctagcontinuationindentation;
+
+// xdoc section -- start
+/**
+ * Writes the object using a
+ * <a href="{@docRoot}/serialized-form.html#java.time.Ser">dedicated form</a>.
+ * @serialData
+ * <pre>
+ *      out.writeByte(1);
```
romani, line 16:
```
     * @serialData Some serialization description, followed by
     *                        example of code on how to do this.
     * <pre>
     * out.writeByte(1);
     * </pre>
     */
```

romani: please also add example where instead of PRE, user use CODE html tag, and it is a violation.

romani (changes requested):
items:

romani:
merged in PR with functional fix

Issue #17839, opened 2025-09-28
False-negative: Method and Member names with underscores


mohitsatr:
From: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case
> In very rare circumstances (for example, multipart version numbers), you may need to use underscores to separate adjacent numbers, since numbers do not have upper and lower case variants.

Google style is not exact. On one hand it says `_` can be used in test methods to combine sentences and on another hand in general CamelCase explanation, `_` is allowed for numbering suffix in all names(except package and generic names).

Valid uses of underscore :
```
1. guava33_4_5
2. guavaVersion33_4_5 
3. guavaversion33_4_5 (same as 1.)
```

Invalid uses of underscore:
```
1. guava_33_4_5_ (underscore at the end)
2. guava_33_4_5   (underscore between digit and letter)
3. guava_version33_4_5 (underscore between lowercase character sequences)
```

**Note**: Underscores are not allowed in normal method names. Name has to have the numbering format part at the end. Following are all invalid

```
guava_version
guava_Version
Set_guavaVersion3345
```

Some false-negatives left to address from #17708  

```java
/** some javadoc. */
public class NamesWithUnderscores {
 
  void gradle_9_5_1() {} // Expected violation, _ between digit and letter

  void jdk_9_0_392() {} // Expected violation, _ between digit and letter

  void testing_01231(String str) {} // Expected violation, _ between digit and letter

  int jdk_8_90; // Expected violation, _ between digit and letter

  int guava_33_4_7; // Expected violation, _ between digit and letter
}
```

```
$ java -jar checkstyle-11.0.0-all.jar -c google_checks.xml NamesWithUnderscores.java 
Starting audit...
Audit done.
```

we probably need a better approach than Regex. Regex has already grown very complex.
 
https://github.com/checkstyle/checkstyle/blob/5d52248c85b0fc37ddd786ddb5354e332f8a54b8/src/main/resources/google_checks.xml#L436-L438

https://github.com/checkstyle/checkstyle/blob/5d52248c85b0fc37ddd786ddb5354e332f8a54b8/src/main/resources/google_checks.xml#L254-L256

if special check is needed, it should first determine the type of name - regular or name with numbering and then it should validate the name against the appropriate regex format, which will differ for each type of naming

Keep in mind: unless it's a non-regular name i.e name that has number formating suffix at the end or method with `@Test` annotation,  underscores are not allowed.

MiniPiku:
[...]

PR #17864, opened 2025-10-05
Fix: Corrected Google Style MethodNameCheck to allow numeric suffixes…

SohamBhoyar:
Issue: #17839 

**Description**
This Pull Request updates the regular expression used by the MethodNameCheck when configured with the Google Style naming convention.

The original Google Style regex did not permit method names containing underscores (_) even when those underscores were used purely to separate numeric suffixes, which is common practice for library or version identifiers (e.g., in utility methods like gradle_9_5_1 or jdk_9_0_392).

The update to the regex allows method names to contain numeric segments separated by underscores, specifically catering to these versioning suffixes while maintaining the overall strictness of the Google naming convention.

**Rationale**
This change resolves false-positive violations when checking code that follows the standard practice of including external version identifiers (which often contain underscores and numbers) within method names for clarity.

**Changes Made**
MethodNameCheck.java: Modified the default MethodName pattern when style=google is set to include the necessary exceptions for numeric suffixes.

MethodNameCheckTest.java: Added a new test case, testUnderscoreFalseNegatives, which uses InputGoogleStyleUnderscore.java to verify that method names like gradle_9_5_1 and testing_01231 no longer cause violations with the updated regex.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/MethodNameCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/MethodNameCheck.java
@@ -56,7 +56,7 @@ public class MethodNameCheck
 
     /** Creates a new {@code MethodNameCheck} instance. */
     public MethodNameCheck() {
-        super("^[a-z][a-zA-Z0-9]*$");
+        super("^[a-z][a-zA-Z0-9]*(_[0-9]+)*$");
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/MethodNameCheck.java
```diff
@@ -56,7 +56,7 @@ public class MethodNameCheck
 
     /** Creates a new {@code MethodNameCheck} instance. */
     public MethodNameCheck() {
-        super("^[a-z][a-zA-Z0-9]*$");
+        super("^[a-z][a-zA-Z0-9]*(_[0-9]+)*$");
```
romani, line 59: change should be in google_checks.xml , not in Check class.
The default value should not be changed.

romani:
@SohamBhoyar , please find time to finish this PR

romani:
there is another PR to make new Check for this.

PR #17876, opened 2025-10-06
Issue #16233: Fix PackageObjectFactoryTest before/after annotations

mureinik:
Follow up on #16500 - since PackageObjectFactoryTest is executed using the JUnit Jupiter teset executor, the `org.junit.BeforeClass` and `org.junit.AfterClass` JUnit 4 annotations aren't picked up, and the methods annotated with them are never executed.

This patch follows up on the work done in #16233 and replaces these annotations with `org.junit.jupiter.api.BeforeAll` and `org.junit.jupiter.api.AfterAll`, respectively, so that these methods are indeed executed as part of the test lifecycle.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/PackageObjectFactoryTest.java
@@ -47,8 +47,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
-import org.junit.AfterClass;
-import org.junit.BeforeClass;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.mockito.MockedStatic;
 import org.mockito.Mockito;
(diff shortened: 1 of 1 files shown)


romani:
Nice catch.

How did you find it?

Please review CI failures, CI must be green.

We probably need to add such imports to forbidden

romani:
Instead of referencing of some old issue.
Make prefix "Pull #17876: ..... " As this some sort new problem. No new issue creation is required, we will reuse Pull.

mureinik:
@romani thanks!

re: CI failures - looks like my own stupid mistake. I misread the commit message instructions, and (wrongly!) thought they referred to the commit message subject, not the entire commit message.
<strike>CI is running now, hopefuly I got it right this time.</strike> CI has passed.

re: marking it with "Pull #17876:" - done.

re: forbidding such imports - it looks like the `import-control-test.xml` was supposed to do that, let me take a look.

romani:
>forbidding such imports - it looks like the import-control-test.xml was supposed to do that, let me take a look.

Yes, please. I will be awesome to firbid all old junit classes, I surprised they are still available in class path

mureinik:
> > forbidding such imports - it looks like the import-control-test.xml was supposed to do that, let me take a look.
> 
> Yes, please. I will be awesome to firbid all old junit classes, I surprised they are still available in class path

Looking in to it.
Seems like there are some mistakes in the `import-control-test.xml`. No rocket science there, but there's some finiky work to fix it. I'll open a separate issue/PR for it.

romani (approved):
Thanks a lot

mureinik:
> Looking in to it. Seems like there are some mistakes in the `import-control-test.xml`. No rocket science there, but there's some finiky work to fix it. I'll open a separate issue/PR for it.

I take back my assessment of `import-control-test.xml`. It seems that the main issue isn't that configuration, but the fact that a lot of test classes just suppress the Import Control check. 
In any event - I'm working on it.
Will open a PR detailing all the issues I find once I get it sorted out.

PR #17891, opened 2025-10-10
Pull #17891: Add `id` to `NeedBraces` declaration

Pankraz76:
### Pull #17891: Remove duplicated `NeedBraces` declaration

### DIFF
--- a/config/checkstyle-checks.xml
+++ b/config/checkstyle-checks.xml
@@ -328,8 +328,11 @@
     </module>
     <module name="EmptyCatchBlock"/>
     <module name="LeftCurly"/>
-    <module name="NeedBraces"/>
     <module name="NeedBraces">
+      <property name="id" value="needBracesDefault"/>
+    </module>
+    <module name="NeedBraces">
+      <property name="id" value="needBracesForLambda"/>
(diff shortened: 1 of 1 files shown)


config/checkstyle-checks.xml
```diff
@@ -328,7 +328,6 @@
     </module>
     <module name="EmptyCatchBlock"/>
     <module name="LeftCurly"/>
-    <module name="NeedBraces"/>
     <module name="NeedBraces">
```
Pankraz76, line 331: this will dominate and override the default, right?

romani: no they are different by config: default and not-default

romani: do you think placing different "id" would make it more clear for people ?
https://checkstyle.sourceforge.io/config.html#Id

Pankraz76: yes like this?

romani: needBracesForLambda
needBracesDefault

Pankraz76: thanks

romani:
please fix CI violation

PR #17928, opened 2025-10-13
Issue #17299: Converted IntMatchFilterElement and IntRangeFilterElement to record

Atharv3221:
Issue: #17299

```
package com.puppycrawl.tools.checkstyle.filters;

/**
 * This filter element is immutable and accepts a matching Integer.
 *
 * @param matchValue the integer value a match
 */
record IntMatchFilterElement(int matchValue) implements IntFilterElement {

    @Override
    public boolean accept(int intValue) {
        return matchValue == intValue;
    }

    @Override
    public String toString() {
        return "IntMatchFilterElement{" + "matchValue=" + matchValue + '}';
    }

}

```
```
/**
 * This filter element is immutable and accepts an Integer in a range.
 *
 * @param lowerBound the lower bound of the range
 * @param upperBound the upper bound of the range
 */
record IntRangeFilterElement(Integer lowerBound, Integer upperBound) implements IntFilterElement {

    @Override
    public boolean accept(int intValue) {
        return lowerBound.compareTo(intValue) <= 0
                && upperBound.compareTo(intValue) >= 0;
    }

    @Override
    public String toString() {
        return "IntRangeFilterElement{" + "lowerBound=" + lowerBound
                + ", upperBound=" + upperBound + '}';
    }
}
```


Atharv3221:
@romani flasky ci

src/main/java/com/puppycrawl/tools/checkstyle/filters/IntRangeFilterElement.java
```diff
@@ -19,53 +19,32 @@
...
     @Override
     public boolean accept(int intValue) {
-        return lowerBound.compareTo(intValue) <= 0
-            && upperBound.compareTo(intValue) >= 0;
-    }
-
-    @Override
-    public int hashCode() {
-        return Objects.hash(lowerBound, upperBound);
+        return lowerBound <= intValue && intValue <= upperBound;
```
romani: let keep method,  unboxing occur implicitly for "<=". Lets avoid it.

src/main/java/com/puppycrawl/tools/checkstyle/filters/IntRangeFilterElement.java
```diff
@@ -19,53 +19,32 @@
...
-            return true;
-        }
-        if (other == null || getClass() != other.getClass()) {
-            return false;
-        }
-        final IntRangeFilterElement intRangeFilter = (IntRangeFilterElement) other;
-        return Objects.equals(lowerBound, intRangeFilter.lowerBound)
-                && Objects.equals(upperBound, intRangeFilter.upperBound);
+    public String toString() {
+        return "IntRangeFilterElement[" + lowerBound + ", " + upperBound + "]";
```
romani: let IDEA generate this method for you, to print field name before value.

src/main/java/com/puppycrawl/tools/checkstyle/filters/IntRangeFilterElement.java
```diff
@@ -19,53 +19,32 @@
...
-     * @param lowerBound the lower bound of the range.
-     * @param upperBound the upper bound of the range.
+     * @param lowerBound the lower bound of the range
+     * @param upperBound the upper bound of the range
      */
-    /* package */ IntRangeFilterElement(int lowerBound, int upperBound) {
-        this.lowerBound = lowerBound;
-        this.upperBound = upperBound;
+    /* package */ IntRangeFilterElement {
+        // Canonical constructor; no extra validation needed
```
romani: I think we do not need this constructor any more

src/main/java/com/puppycrawl/tools/checkstyle/filters/IntMatchFilterElement.java
```diff
@@ -47,12 +45,12 @@ public String toString() {
...
 
     @Override
-    public final int hashCode() {
+    public int hashCode() {
         return Integer.valueOf(matchValue).hashCode();
     }
 
     @Override
-    public final boolean equals(Object object) {
+    public boolean equals(Object object) {
```
romani: in other class/record you remove such methods.
What is value to keep this methods?

try to look at git history why such methods were added.

Atharv3221: Okay

Atharv3221: I checked it was added while creation of that file no other mention of it

Atharv3221:
@romani can please add hacktoberfest accepted tag ?

Issue #17882, opened 2025-10-07
Update `JavadocCommentsTokenTypes.java` to new format of AST print


mahfouz72:
Continuation of https://github.com/checkstyle/checkstyle/issues/14631

We need to update the Javadoc of each token in [`JavadocCommentsTokenTypes.java`](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java) to include an example of the **new AST print format** generated by the latest Checkstyle snapshot.

Each update should document how the token appears in the Javadoc AST, providing both an example input and its corresponding tree structure.

---

#### **Steps to fix**

1. **Build the snapshot version of Checkstyle**

   Generate the latest snapshot JAR by following the steps in
   [How to generate all binaries and `-all.jar` too](https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations#how-to-generate-all-binaries-and--alljar--too).

2. **Generate the AST for the example**

   Example input (`src/Test.java`):

   ```java
   * @param value The parameter of method.
   ```

   Command to generate the Javadoc AST:

   **Windows (PowerShell):**

   ```powershell
   java -jar checkstyle-12.0.0-SNAPSHOT-all.jar -j src/Test.java | ForEach-Object { $_ -replace '\[[0-9]+:[0-9]+\]', '' }
   ```

   **Linux/macOS:**

   ```bash
   java -jar checkstyle-12.0.0-SNAPSHOT-all.jar -j src/Test.java | sed "s/\[[0-9]\+:[0-9]\+\]//g"
   ```

   **Output:**

   ```
   JAVADOC_CONTENT -> JAVADOC_CONTENT 
   |--LEADING_ASTERISK -> * 
   |--TEXT ->   
   `--JAVADOC_BLOCK_TAG -> JAVADOC_BLOCK_TAG 
       `--PARAM_BLOCK_TAG -> PARAM_BLOCK_TAG 
           |--AT_SIGN -> @ 
           |--TAG_NAME -> param 
           |--TEXT ->   
           |--PARAMETER_NAME -> value 
           `--DESCRIPTION -> DESCRIPTION 
               `--TEXT ->  The parameter of method
   ```

3. **Update Javadoc for `PARAM_BLOCK_TAG`**

   Example update:

   ```java
   /**
    * {@code @param} Javadoc block tag.
    *
[...]

PR #17935, opened 2025-10-14
Issue #17882: Update {@summary} of JavadocCommentsTokenTypes to new AST format 

KishorMore-2005:
Issue #17882 : 

```
|--JAVADOC_INLINE_TAG -> JAVADOC_INLINE_TAG 
|   `--CODE_INLINE_TAG -> CODE_INLINE_TAG 
|       |--JAVADOC_INLINE_TAG_START -> {@ 
|       |--TAG_NAME -> code 
|       |--TEXT ->  {@summary} 
|       `--JAVADOC_INLINE_TAG_END -> } 
|--TEXT ->  inline tag. 
|--NEWLINE -> \n 
|--LEADING_ASTERISK ->      * 
|--TEXT -> / 
```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
@@ -363,6 +363,31 @@ public final class JavadocCommentsTokenTypes {
 
     /**
      * {@code {@summary}} inline tag.
+     *
+     * <p>Such Javadoc tag can have one child:</p>
+     * <ol>
+     *   <li>{@link #PARAMETER_NAME} – the name of the parameter</li>
+     *   <li>{@link #DESCRIPTION} – description of the parameter</li>
+     * </ol>
+     *
+     * <p><b>Example:</b></p>
+     * <pre>{@code * {@summary This method computes the total sum.}}</pre>
+     *
+     * <b>Tree:</b>
+     * <pre>{@code
+     *     |--JAVADOC_INLINE_TAG -> JAVADOC_INLINE_TAG
+     *     |   `--CODE_INLINE_TAG -> CODE_INLINE_TAG
+     *     |       |--JAVADOC_INLINE_TAG_START -> {
+     *     |       |--TAG_NAME -> code
+     *     |       |--TEXT ->  {@summary}
+     *     |       `--JAVADOC_INLINE_TAG_END -> }
+     *     |--TEXT ->  inline tag.
+     *     |--NEWLINE -> \n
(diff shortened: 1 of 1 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
```diff
@@ -363,6 +363,31 @@ public final class JavadocCommentsTokenTypes {
...
+     *
+     * <b>Tree:</b>
+     * <pre>{@code
+     *     |--JAVADOC_INLINE_TAG -> JAVADOC_INLINE_TAG
+     *     |   `--CODE_INLINE_TAG -> CODE_INLINE_TAG
+     *     |       |--JAVADOC_INLINE_TAG_START -> {
+     *     |       |--TAG_NAME -> code
+     *     |       |--TEXT ->  {@summary}
+     *     |       `--JAVADOC_INLINE_TAG_END -> }
+     *     |--TEXT ->  inline tag.
```
romani, line 384: from where "inline tag" comes ?

keep only JAVADOC_INLINE_TAG and its children in tree

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
```diff
@@ -363,6 +363,31 @@ public final class JavadocCommentsTokenTypes {
...
      * {@code {@summary}} inline tag.
+     *
+     * <p>Such Javadoc tag can have one child:</p>
+     * <ol>
+     *   <li>{@link #PARAMETER_NAME} – the name of the parameter</li>
+     *   <li>{@link #DESCRIPTION} – description of the parameter</li>
+     * </ol>
+     *
+     * <p><b>Example:</b></p>
+     * <pre>{@code * {@summary This method computes the total sum.}}</pre>
```
mahfouz72, line 374: Everything inside the `{@code}` inline tag is parsed as TEXT. This is by design, and we don't want " `{@summary This method computes the total sum.}`  " to be parsed as text; we want it to be `SUMMARY_INLINE_TAG ` so when generating the tree, remove this outer code inline tag

mahfouz72 (changes requested):
item

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
mahfouz72: So just add `* {@summary This method computes the total sum.}`  and then run the command and copy the tree.

PR #17948, opened 2025-10-17
Issue #17487: add pluginManagement for `rewrite-maven-plugin`

Pankraz76:
### Issue #17487: add pluginManagement for `rewrite-maven-plugin`

The `rewrite-maven-plugin` is not placed under the `pluginManagement` section because it needs to be directly available for execution from the command line and IDE without requiring inheritance from a parent POM.  

When defined only in `pluginManagement`, the plugin configuration is not automatically applied — it must be explicitly declared in each module’s build section. This would add unnecessary duplication and complexity for developers who need to run rewrite tasks locally.  

Keeping it in the `build` section ensures that the plugin can be executed consistently in all environments while maintaining compatibility with existing project profiles such as `checkstyle-autofix`.

PR #17953, opened 2025-10-18
Issue #17487: Apply `sortPom`

Pankraz76:
### Issue #17487: Apply `sortPom`

extract from:

- https://github.com/checkstyle/checkstyle/pull/17948

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -583,6 +583,33 @@
           </configuration>
         </plugin>
         <plugin>
+          <groupId>com.diffplug.spotless</groupId>
+          <artifactId>spotless-maven-plugin</artifactId>
+          <version>3.0.0</version>
+          <configuration>
+            <java>
+              <removeUnusedImports />
+              <importOrder>
+                <order>java|javax,org,com,com.puppycrawl,,\#com.puppycrawl,\#</order>
+              </importOrder>
+              <!--
+              <palantirJavaFormat/>
+              <licenseHeader>
+                <file>config/maven-header-plain.txt</file>
+              </licenseHeader>
+              -->
+            </java>
+            <!--
+              <pom>
+                <sortPom/>
+              </pom>
+            -->
+            <upToDateChecking>
+              <enabled>true</enabled>
+            </upToDateChecking>
+          </configuration>
+        </plugin>
+        <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-pmd-plugin</artifactId>
           <version>${maven.pmd.plugin.version}</version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -582,6 +582,33 @@
...
+        <plugin>
+          <groupId>com.diffplug.spotless</groupId>
+          <artifactId>spotless-maven-plugin</artifactId>
+          <version>3.0.0</version>
+          <configuration>
+            <java>
+              <removeUnusedImports />
+              <!--
+              <importOrder>
+                <order>java|javax,org,com,com.puppycrawl,,\#com.puppycrawl,\#</order>
```
Pankraz76, line 593: currently making diff. but this is just a config issue.

Could add and apply/fix step by step.

pom.xml
```diff
@@ -582,6 +582,33 @@
...
+              </importOrder>
+              <palantirJavaFormat/>
+              <licenseHeader>
+                <file>config/maven-header-plain.txt</file>
+              </licenseHeader>
+              -->
+            </java>
+            <!--
+              <pom>
+                <sortPom/>
```
Pankraz76, line 604: this is simple formatting and should is imho better to apply by reduction as this makes a simpler diff. This will change maybe a lot, but only once of course.

pom.xml
romani: Let's remove this, no need in storing non good config.
It will be good to add it when it will be good

pom.xml
romani: Lets add this next PR and in same PR see how it makes pom different and better.

romani (changes requested):
For introduction of spotless, lets just keep `removeUnusedImports`.

pom.xml
Pankraz76: yes of course this almost 90% but yes i understand.

pom.xml
Pankraz76: updated to this exclusively.

Pankraz76:
> For introduction of spotless, lets just keep `removeUnusedImports`.

saw this as lastest right after push, extracted into:

- https://github.com/checkstyle/checkstyle/pull/17959

pom.xml
```diff
@@ -556,10 +551,8 @@
           <version>3.12.0</version>
           <configuration>
             <!-- Exclude generated sources. -->
-            <excludePackageNames>
-              com.puppycrawl.tools.checkstyle.grammar.java:
-              com.puppycrawl.tools.checkstyle.grammar.javadoc:
-            </excludePackageNames>
+            <excludePackageNames>com.puppycrawl.tools.checkstyle.grammar.java:
```
Pankraz76: yes not nice yes, but who cares?

we have now a format for us now, to be straight forward.

I also would go with ned telling, "idk".

> I don't care where it goes, fine with me

- https://github.com/diffplug/spotless/pull/2687#discussion_r2436709418

Pankraz76:
reopen when foundation is set.

Issue #17691, opened 2025-08-26
Website Generation doesn't automatically reflect new changes to checks.xml and checks/<type>/index.xml


Zopsss:
I have read check documentation: https://checkstyle.org/checks/xxxxxx/nameofaffectedcheck.html
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

After modifying a Check’s summary (for example, AnnotationLocation), I observed the following:

<img width="1466" height="638" alt="Image" src="https://github.com/user-attachments/assets/3201be69-425b-48cb-955e-7318e97721af" />

After running `mvn clean verify`, the above change is reflected in `annotationlocation.xml` and `AnnotationLocationCheck.xml`

But it is not reflected in `checks/annotation/index.xml`, though `XdocsCategoryIndexTest`'s tests fails and gives following errors:

```
[ERROR]   XdocsCategoryIndexTest.testAllChecksListedInCategoryIndexAndDescriptionMatches:132 Check 'AnnotationLocation' in index 'src\site\xdoc\checks\annotation\index.xml': index description is not a prefix of XDoc description.                            
expected to start with: Checks location of annotation on language elements.                                                     
but was               : Checks location of annotation on language elements testing. By default, Check enforce to locate annotati
ons before target element, annotation should be located on separate line from target element. This check also verifies that the annotations are on the same indenting level as the annotated element if they are not on the same line.                          
[ERROR]   XdocsPagesTest.testAllModulesPageInSyncWithModuleSummaries:335->validateModulesSyncWithTheirSummaries:391 The summary 
for module AnnotationLocation in the file "src\site\xdoc\checks.xml" should match the first sentence of the "Description" subsection for this module in the file "annotationlocation.xml"                                                                       
expected: Checks location of annotation on language elements testing.                                                           
but was : Checks location of annotation on language elements.                                                                   
[INFO]
[ERROR] Tests run: 5597, Failures: 2, Errors: 0, Skipped: 53
```

( screenshot showing that the summary is not updated )
<img width="1869" height="886" alt="Image" src="https://github.com/user-attachments/assets/f9731571-8f4f-427d-8051-a658fffa3fb8" />

We need to manually place the new summary. Instead of this, `checks/annotation/index.xml` should be able to automatically fetch the newly modified summary.

[...]

PR #17962, opened 2025-10-20
Issue #17691 : Website Generation doesn't automatically reflect new c…

smita1078:
#17691 
### What is done
Introduced a new macro-based approach to auto-generate `checks.xml` instead of maintaining it manually.

- Added a new macro class `ChecksXMLMacro` under `com.puppycrawl.tools.checkstyle.site`.
- The macro dynamically reads each Check module’s Javadoc and extracts the first-sentence summary.
- Replaced static descriptions in `checks.xml` with `<macro name="checks" ...>` tags, allowing automatic updates based on source Javadoc.
- Sanitization added to remove unwanted HTML tags (like <div>, <a>, etc.) for clean XML output.
- This approach ensures consistency between documentation and source Javadoc while reducing manual maintenance effort.

### Files Added/Modified
- `src/main/java/com/puppycrawl/tools/checkstyle/site/ChecksXMLMacro.java`
- `src/site/xdoc/checks.xml.template`


src/site/xdoc/checks.xml
```diff
@@ -364,33 +448,46 @@
               </a>
             </td>
             <td>
-              Checks if any class or object member is explicitly initialized to default
-              for its type value (<code>null</code> for object references, zero for numeric
-              types and <code>char</code> and <code>false</code> for <code>boolean</code>.
+              Checks if any class or object member is explicitly initialized to
+default for its type value (<code>null</code> for object references,
```
romani: can we add indentation while we split by multiple lines?

smita1078:
> idea that come to mind while looking at this is , why would we need macros for each and every Check, but page will always show all Checks. So conceptually we need single macros that prints summaries for all modules in alphabet order.
> 
> In this case special test that we have will not be required, as humans will not change this file anymore.
> 
> If you agree we can do such step in separate PR.

Agreed!
Right now ChecksXmlMacro runs per module (i.e., per Check class).
It extracts the Javadoc for a single Check and inserts it into that Check’s <description> section in checks.xml.
But the checks.xml file is a single, global listing of all checks — alphabetically grouped — used to generate the Checkstyle site.

So, conceptually:

There should be one macro (not per Check). That macro should dynamically loop through all Check modules (using metadata)
and render summaries for each, alphabetically.

romani:
Please fix CI failures, let merge with current design, and in follow up PR you will improve to single macros.

smita1078:
> Please fix CI failures, let merge with current design, and in follow up PR you will improve to single macros.

Sure, on it!

src/site/xdoc/checks.xml
smita1078: yes, should i add in this pr or the next single macro one...?

romani: please add to this PR.

In single macros PR, we will expect no changes in result xml.

So I will have to validate change in xml only ones.

smita1078: Sure

src/site/xdoc/checks.xml
```diff
@@ -28,10 +27,14 @@
...
-              AbbreviationAsWordInName</a></td>
-            <td>Validates abbreviations (consecutive capital letters)
-             length in identifier name, it also allows to enforce camel case naming.</td>
+            <td>
+              <a href="checks/naming/abbreviationaswordinname.html#AbbreviationAsWordInName">
+              AbbreviationAsWordInName</a>
+            </td>
+            <td>
+              Validates abbreviations (consecutive capital letters) length in
+identifier name, it also allows to enforce camel case naming.
```
romani: Please correct indentation.
Should be +2 from `<dt>`

romani (changes requested):
Item

romani:
GitHub, generate website

romani:
@smita1078 , thanks a lot for update.
Please send second PR with better solution.

smita1078:
> @smita1078 , thanks a lot for update. Please send second PR with better solution.

Sure!

PR #17969, opened 2025-10-21
Issue #15456 Define violation messages for MagicNumberCheck

ivamly:
Issue #15456 Define violation messages for MagicNumberCheck


src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputViolation.java
```diff
@@ -62,6 +65,7 @@ public String toRegex() {
             rawMessage = OPEN_CURLY_PATTERN.matcher(rawMessage).replaceAll("\\\\{");
             rawMessage = OPEN_PAREN_PATTERN.matcher(rawMessage).replaceAll("\\\\(");
             rawMessage = CLOSE_PAREN_PATTERN.matcher(rawMessage).replaceAll("\\\\)");
+            rawMessage = EXPLICIT_POSITIVE_NUMBER_PATTERN.matcher(rawMessage).replaceAll("\\\\+");
```
romani: Please explain this change.
Not obvious.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/magicnumber/InputMagicNumberIgnoreNegativeOctalHex2.java
```diff
@@ -19,15 +19,15 @@ public class InputMagicNumberIgnoreNegativeOctalHex2 {
...
+    long l = 0xffffffffL; // violation ''0xffffffffL' is a magic number'
 
 
     /**
      * test signed values
      */
     public static final int CONST_PLUS_THREE = +3;
     public static final int CONST_MINUS_TWO = -2;
-    private int mPlusThree = +3; // violation
+    private int mPlusThree = +3; // violation ''+3' is a magic number'
```
romani: if extra regexp is for this line ..... 

please make it `private int mPlusThree = +3; // violation 'is a magic number'`

message is regexp, and so exact value of number can be skipped, as we do this in all other cases.
We do design test inputs with care and single item per line avoid mentioning full message.

Something like this can escape +, so not ideal but ok apprach if we badly need "+3"
`private int mPlusThree = +3; // violation ''\+3'is a magic number'`

src/test/java/com/puppycrawl/tools/checkstyle/bdd/TestInputViolation.java
ivamly: This line escapes the plus sign + before digits in test violation messages. I'll try the other solutions suggested below.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/magicnumber/InputMagicNumberIgnoreNegativeOctalHex2.java
ivamly: Thank you for the clarification. I've removed the extra regex pattern and updated all test cases with explicit plus signs to use the generic 'is a magic number' message, consistent with other tests.

Issue #17971, opened 2025-10-22
Add `spotless`


Pankraz76:
related to: 

- https://github.com/checkstyle/checkstyle/pull/17959
- https://github.com/checkstyle/checkstyle/issues/17487

I have read check documentation: https://checkstyle.org/checks/xxxxxx/nameofaffectedcheck.html
I have downloaded the latest cli from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

**How it works Now:**

```bash
/var/tmp $ javac YOUR_FILE.java
#[[MAKE SURE THERE IS SUCCESSFUL COMPILATION]]

/var/tmp $ cat config.xml
#[[PLACE YOUR OUTPUT HERE]]

/var/tmp $ cat YOUR_FILE.java
#[[PLACE YOU OUTPUT HERE]]

/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-X.XX-all.jar -c config.xml YOUR_FILE.java
#[[PLACE YOUR OUTPUT HERE]]
```

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is.

**Describe the solution you'd like**
A clear and detailed description of what you want to happen.
You could provide the expected/suggested cli output for your solution.

```bash
expected/suggested cli output
```

**Additional context**

Still not clear ???
see examples:

- https://checkstyle.org/report_issue.html#How_to_request_new_feature_for_existing_functionality.3F

- https://checkstyle.org/report_issue.html#How_to_request_new_Check.2FModule.3F

**Not a feature?**
If (Check, property, new behavior) that you are requesting is already
implemented and you are unsatisfied with their behaviour, report it here
in bug template format:

- https://checkstyle.org/report_issue.html#How_to_report_a_bug.3F

ATTENTION: FAILURE TO FOLLOW THE ABOVE TEMPLATE WILL RESULT IN THE ISSUE BEING CLOSED.

Pankraz76:
fixed thanks. @romani 

- https://github.com/checkstyle/checkstyle/issues/17732

PR #17972, opened 2025-10-22
Issue #17971: Apply `sortPom`

Pankraz76:
### Issue #17971: Apply `sortPom`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
@@ -16,10 +15,8 @@
   <packaging>jar</packaging>
 
   <name>checkstyle</name>
-  <description>
-    Checkstyle is a development tool to help programmers write Java code
-    that adheres to a coding standard
-  </description>
+  <description>Checkstyle is a development tool to help programmers write Java code
+    that adheres to a coding standard</description>
   <url>https://checkstyle.org/</url>
   <inceptionYear>2001</inceptionYear>
   <licenses>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -595,8 +588,11 @@
           <version>3.0.0</version>
           <configuration>
             <java>
-              <removeUnusedImports />
+              <removeUnusedImports/>
             </java>
+            <pom>
+              <sortPom/>
```
Pankraz76, line 595: actual change.

pom.xml
```diff
@@ -595,8 +588,12 @@
           <version>3.0.0</version>
           <configuration>
             <java>
-              <removeUnusedImports />
+              <forbidWildcardImports/>
```
Pankraz76, line 591: snaked in as already complain.

pom.xml
```diff
@@ -5162,92 +5061,38 @@
...
-                  com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheckTest
-                </param>
-                <param>
-                  com.puppycrawl.tools.checkstyle.checks.AvoidEscapedUnicodeCharactersCheckTest
-                </param>
-                <param>
-                  com.puppycrawl.tools.checkstyle.checks.whitespace.SingleSpaceSeparatorCheckTest
-                </param>
+                <param>com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheckTest</param>
+                <param>com.puppycrawl.tools.checkstyle.checks.AvoidEscapedUnicodeCharactersCheckTest</param>
```
Pankraz76, line 5065: same order right?

Pankraz76:
> I do not like this auto sorting, it has a lot side effects

yes im not profound of this plugin but imho its just "bad" naming as its not sorting the properties imho, this is something i was expecting and requesting in this issue:

- https://github.com/Ekryd/sortpom/issues/469

considering the comment it is sorting but in a way that it does not change the current order right?

Imho it does only "format" the xml rather then changing it logically. Having big trouble to add new plugin having to format manually this is something the plugin correct indent is handling.

romani:
I don't see we can use this pom updates.
A lot was single lined, but it is multi line on purpose, to avoid line length violation. It is critical for our project.

Pankraz76:
closing due to:

> avoid line length violation

PR #17976, opened 2025-10-22
Issue #14019: Kill surviving mutation in IllegalInstantiationCheck

gianmarcoschifone:
Issue #14019

Before this change, PIT reported one surviving mutation in `IllegalInstantiationCheck`. The mutator (`NonVoidMethodCallMutator`) removed the implicit `String.valueOf(ast)` call inside `throw new IllegalArgumentException("Unknown type " + ast);` replacing it with a `null` value.

Because the existing test `testImproperToken` only checked that an exception was thrown, not what message it contained, the mutated code (`"Unknown type null"`) still passed.

This PR updates the test to assign a token text (`lambdaAst.setText("LAMBDA")`) and assert that the exception message contains `"LAMBDA"`. The test now fails for the mutated version.

Screenshots show the PIT report before and after the fix.
image
image

image
image

### DIFF
--- a/config/pitest-suppressions/pitest-coding-2-suppressions.xml
+++ b/config/pitest-suppressions/pitest-coding-2-suppressions.xml
@@ -1,13 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <mutation unstable="false">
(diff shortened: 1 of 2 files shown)


romani (approved):
Thanks a lot.
Rare case when we allow pure unit test

PR #17987, opened 2025-10-23
Issue #11163: Enforce file size on DeclarationOrder folder

ivamly:
Part of https://github.com/checkstyle/checkstyle/issues/11163: Enforce file size on DeclarationOrder folder

### DIFF
--- a/config/checkstyle-resources-suppressions.xml
+++ b/config/checkstyle-resources-suppressions.xml
@@ -456,10 +456,6 @@
   <suppress checks="FileLength"
     files="[\\/]test[\\/]resources[\\/]com[\\/]puppycrawl[\\/]tools[\\/]checkstyle[\\/]checks[\\/]coding[\\/]nofinalizer[\\/]InputNoFinalizerFallThrough\.java"/>
   <suppress checks="FileLength"
-    files="[\\/]test[\\/]resources[\\/]com[\\/]puppycrawl[\\/]tools[\\/]checkstyle[\\/]checks[\\/]coding[\\/]declarationorder[\\/]InputDeclarationOrderOnlyModifiers\.java"/>
-  <suppress checks="FileLength"
-    files="[\\/]test[\\/]resources[\\/]com[\\/]puppycrawl[\\/]tools[\\/]checkstyle[\\/]checks[\\/]coding[\\/]declarationorder[\\/]InputDeclarationOrderOnlyConstructors\.java"/>
-  <suppress checks="FileLength"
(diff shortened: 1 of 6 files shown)


ivamly:
By the way, i noticed that when tests fail due to violation mismatches, the error output appears counterintuitive. For example:

 ```
[ERROR] DeclarationOrderCheckTest.testOnlyConstructors2:114->AbstractModuleTestSupport.verifyWithInlineConfigParser:278 
Violations for /checkpath/InputDeclarationOrderOnlyConstructors2.java differ.
value of      : getActualViolationsForFile(...)                                                                                                                           
missing (1)   : 28:9: Static variable definition in wrong order.                                                                                                          
unexpected (1): 27:9: Static variable definition in wrong order.
```

The message creates confusion: it seems like we should have a violation on line 28 (marked as "missing") but don't, 
while the violation on line 27 (marked as "unexpected") appears to be extra. 
In reality, it's likely the opposite - we expect the violation on line 27, not line 28.

This message originates from this assertion pattern in [AbstractModuleTestSupport.java](https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java):
 ```java
assertWithMessage("Violations for %s differ.", filePath)
    .that(actualViolations)
    .containsExactlyElementsIn(expected);
 ```
According to AssertJ javadoc and [examples](https://assertj.github.io/doc/#:~:text=assertThat(actual).isEqualTo(expected)):

1. The first parameter should be the actual/under test collection
2. The second parameter should be the expected/reference collection

However, the current implementation seems to have these reversed, leading to confusing error messages where "missing" and "unexpected" labels don't match intuitive expectations.

Maybe we should swap the arguments and after the change, error messages will be logical:

- missing: violations that were expected but not found in actual results
- unexpected: violations that appear in actual results but weren't expected

What you think?

romani:
Thanks a lot for sharing this.
We recently merged fix for this https://github.com/checkstyle/checkstyle/commit/0c6c603bc66c2ef40cfd544dbc0afed3b9b28f78

Please Send separate PR if something is left as not fixed.
But it will be complicated, as we don't know what is expected. We just compare two list of violations and  user needs to figure out what is more correct.
Sometimes you fix false positives and you remove it from input file, because it is visually simple to spot, but update of array is less intuitive, and it is old expected, not current.

Can we update message to mention it or is there other suggestions?

ivamly:
Thanks for the answer! I'm not sure I understand you correctly - which specific messages are you referring to? Maybe it makes sense to open an issue to discuss this point more broadly?

romani:
Yes, let's discuss separately, and you can make PRs to show a problem

ivamly:
Should i open the issue?

romani:
@ivamly , yes, please.

Issue #17975, opened 2025-10-22
Incorrect encoding in non-English properties files causing garbled characters in localized messages


VinishaWagh:
```
# 1. Create the Java file (forcing ASCII encoding)
PS D:\issue> Set-Content -Path DummyFile.java -Value "public class DummyFile {}" -Encoding Ascii

# 2. Create the config.xml (with FileLength as a child of Checker)
PS D:\issue> Set-Content -Path config.xml -Value @"
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
    "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <module name="TreeWalker" />
  <module name="FileLength">
    <property name="max" value="0"/>
  </module>
</module>
"@ -Encoding Ascii

# 3. Compile the Java file
PS D:\issue> javac DummyFile.java

# 4. Set PowerShell locale variables as an array
PS D:\issue> $RUN_LOCALE = @("-Duser.language=tr", "-Duser.country=TR")
```
**Output:**
```
Denetleme ba┼ƒl─▒yor...
[ERROR] D:\issue\DummyFile.java:1: Dosya uzunlu─ƒu 1 sat─▒r (maksimum izin verilen de─ƒer 0). [FileLength]
Denetleme tamamland─▒.
Checkstyle 1 hatayla bitiyor.
```

**Expected Output:**
```
Denetleme başlıyor...
[ERROR] D:\issue\DummyFile.java:1: Dosya uzunluğu 1 satır (maksimum izin verilen değer 0). [FileLength]
Denetleme tamamlandı.
Checkstyle 1 hatayla bitiyor.
```

The output is garbled (e.g., ba┼ƒl─▒yor, uzunlu─ƒu) because many non-English message files (e.g., messages_tr.properties, messages_ja.properties, messages_de.properties, etc.) contain strings encoded inconsistently.

Some lines use correct Unicode escapes (e.g., \u00f6).

Other lines contain raw UTF-8 or mis-encoded characters (e.g., ─ƒu, ┼ƒ).

This results in an incorrect display when Checkstyle is run in a non-English locale, as proven by the CLI output above.

**Proposed Fix:**
To resolve this, all .properties files in src/main/resources/com/puppycrawl/tools/checkstyle/ should be standardized.

[...]

PR #17991, opened 2025-10-24
Issue #17975: Standardize properties files to UTF-8

VinishaWagh:
Issue: #17975

### DIFF
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -9118,11 +9118,16 @@
   <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java</fileName>
     <specifier>type.arguments.not.inferred</specifier>
-    <message>Could not infer type arguments for Stream.map</message>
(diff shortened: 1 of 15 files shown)


romani:
Need more details on why we can not use UTF-8 in such files.

VinishaWagh:
> Need more details on why we can not use UTF-8 in such files.

We can use UTF-8 for these files but since **the default encoding for properties files is ISO-8859-1 in most of the IDE's**..(By default, Eclipse, IntelliJ IDEA, and NetBeans all use ISO-8859-1). I think this is safer option.
Otherwise, contributors would have to manually change the settings for the same..
And this is why their was ambiguity in the file, since some contributors added  UTF-8 based lines knowing the fact that their IDE is set to ISO-8859-1 as their default encoding... which led to garbled text because ISO-8859-1 couldn't understand non-ascii characters.
However, we can also do the UTF-8 option... but would need to ensure the setting change is done and .properties files are consistent.
If you want, I can also go for the second option as you mentioned.

romani:
I think we should use utf8 now, and we will not allow changes to such files in existing lines, git will detect any attempts to change them and maintainers with prevent it from merging, same I caught it in your PR.

Let's convert them to utf8.

VinishaWagh:
Alright... On it..

Issue #17045, opened 2025-05-09
UI Flickering on Checkstyle.org – Panel Resizes Erratically on Navigation


Brijeshthummar02:
Hi everyone,

I wanted to report a UI issue I've been experiencing consistently on the [Checkstyle website](https://checkstyle.org/). Whenever I navigate to any subpage (for example: clicking on "Configuration", "Javadoc Checks", or other sidebar links), the entire main content panel flickers noticeably. The issue involves the panel rapidly resizing --- it grows larger and then shrinks back to normal within a second or two.

This behavior is visually disruptive, especially on desktop. It seems like the layout re-renders or recalculates dimensions on each internal page load, possibly due to a transition effect or glitch.

Steps to reproduce:

1.  Go to [https://checkstyle.org](https://checkstyle.org/)

2.  Click on any sidebar link

3.  Observe the main content panel flicker and change size for a moment before stabilizing.

stoyanK7:
The issue comes from
https://github.com/checkstyle/checkstyle/blob/d7adc9240e2a2f38a000b4bbf139f04ea3df62af/src/site/resources/js/checkstyle.js#L193
and
https://github.com/checkstyle/checkstyle/blob/d7adc9240e2a2f38a000b4bbf139f04ea3df62af/src/site/resources/js/checkstyle.js#L70-L94

The `load` event fires only after _the whole page has loaded_ - scripts, images, iframes, stylesheets,... [^1]. We could instead use the `DOMContentLoaded` event that fires earlier - _when the HTML document has been completely parsed, and all deferred scripts have downloaded and executed_[^2]. Even the docs suggest that using `load` is a common mistake[^2]:
> A different event, [`load`](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event), should be used only to detect a fully-loaded page. It is a common mistake to use `load` where `DOMContentLoaded` would be more appropriate.

We don't need a fully loaded page to set the `className` of `#leftColumn` and `#bodyColumn`, just the DOM. Therefore, using `DOMContentLoaded` is something we can do.

I tested it locally, and it definitely gets rid of flickering _in most cases - like 80%_. Though, sometimes we still get the flicker if the JavaScript is delayed (for whatever reason).

[^1]: [Window: load event - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event)
[...]

PR #18036, opened 2025-11-01
Issue #17045: UI Flickering on Checkstyle.org – Panel Resizes Erratically on Navigation

stoyanK7:
## Issue
- #17045

## Summary - What changed?
- Changed the event that triggers `setBodyColumnMargin()` from `window ("load")` to `document ("DOMContentLoaded")`
	- The `load` event fires only after _the whole page has loaded_ - scripts, images, iframes, stylesheets,... [^1]. We could instead use the `DOMContentLoaded` event that fires earlier - _when the HTML document has been completely parsed, and all deferred scripts have downloaded and executed_[^2]. Even the docs suggest that using `load` is a common mistake[^2]:
		> A different event, [`load`](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event), should be used only to detect a fully-loaded page. It is a common mistake to use `load` where `DOMContentLoaded` would be more appropriate.

		We don't need a fully loaded page to set the `className` of `#leftColumn` and `#bodyColumn`, just the DOM. Therefore, using `DOMContentLoaded` is something we can do.
- Added `async` and `defer` attributes to the `checkstyle.js` script so that it's fetched in parallel to the HTML parser
	- Until now, that was done in sequence
	- The official HTML specification[^3] has a very good diagram explaining what each attribute does:
	    -  image
	    - They also say that
	        > _"The [defer](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-defer) attribute may be specified even if the [async](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-async) attribute is specified, to cause legacy web browsers that only support [defer](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-defer) (and not [async](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-async)) to fall back to the [defer](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-defer) behavior instead of the blocking behavior that is the default."_
	        
## Does this all work?
I see an improvement in 80% of cases. Though sometimes flickering still occurs (for instance, on the `Release Notes` page). You just have to test on your browser.

[^1]: [Window: load event - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event)
[^2]: [Document: DOMContentLoaded event - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event)
[^3]: [HTML Standard 4.12.1 The script element](https://html.spec.whatwg.org/multipage/scripting.html#the-script-element)

### DIFF
--- a/src/site/resources/js/checkstyle.js
+++ b/src/site/resources/js/checkstyle.js
@@ -158,5 +158,11 @@ function trimCodeBlock(codeBlock) {
     }
 }
 
-window.addEventListener("load", setBodyColumnMargin);
+if (document.readyState === "loading") {
+    document.addEventListener("DOMContentLoaded", setBodyColumnMargin);
+}
+else {
+    setBodyColumnMargin();
+}
+
 window.addEventListener("resize", setBodyColumnMargin);
--- a/src/site/site.xml
+++ b/src/site/site.xml
@@ -51,7 +51,7 @@
     <head>
       <!-- Will be added to every generated page -->
       <![CDATA[
-        <script type="text/javascript" src="$relativePath/js/checkstyle.js"></script>
+        <script type="text/javascript" src="$relativePath/js/checkstyle.js" defer async></script>
         <script type="text/javascript" src="$relativePath/js/anchors.js"></script>
         <script type="text/javascript" src="$relativePath/js/google-analytics.js"></script>
         <script type="text/javascript" src="$relativePath/js/copy-clipboard.js"></script>


stoyanK7:
Github, generate website

src/site/site.xml
```diff
@@ -51,7 +51,7 @@
     <head>
       <!-- Will be added to every generated page -->
       <![CDATA[
-        <script type="text/javascript" src="$relativePath/js/checkstyle.js"></script>
+        <script type="text/javascript" src="$relativePath/js/checkstyle.js" defer async></script>
         <script type="text/javascript" src="$relativePath/js/anchors.js"></script>
         <script type="text/javascript" src="$relativePath/js/google-analytics.js"></script>
         <script type="text/javascript" src="$relativePath/js/copy-clipboard.js"></script>
```
stoyanK7, line 57: IMO, we should add `defer async` to the rest of the scripts too. I see no reason not to.

stoyanK7:
Here is a video comparing  [https://checkstyle.sourceforge.io](https://checkstyle.sourceforge.io) and the generated website in the above comment:
[Screencast from 2025-11-01 15-59-45.webm](https://github.com/user-attachments/assets/42b0ca80-f128-4db9-8f02-48296be2e011)

Firefox on Ubuntu 24.04

romani (approved):
tested in Chrome. With this update works better.

most of the time, I open this website from phone, and phone has different rendering and this problem is not visiable.

thanks a lot!

romani:
@smita1078 , please looks at this change too.

src/site/site.xml
Zopsss: adding defer I can understand but adding async will make the script execute in parallel with HTML parsing that means while the scripts are executing, the DOM might not have been parsed and we need the DOM to be parsed before executing the script. Please correct me if I am wrong.

smita1078: > adding defer I can understand but adding async will make the script execute in parallel with HTML parsing that means while the scripts are executing, the DOM might not have been parsed and we need the DOM to be parsed before executing the script. Please correct me if I am wrong.

+1, I too think only defer is fine, async is not required it might cause conflict

stoyanK7: Correct, the script will execute before the DOM is parsed. 

However, the execution of the script _only registers the listeners_ (`.addEventListener()`). The callback function that is registered with each listener will trigger only when the corresponding event fires.

If we take the following code:
https://github.com/checkstyle/checkstyle/blob/0d604f66966b85b1b0b0c95d3319ead4a6ac6d34/src/site/resources/js/checkstyle.js#L8-L18

The listener will be registered before the HTML parser is done, but the callback (everything within `function () { ..}` will run only when the `"load"` event fires. And the `"load"` event fires way after the DOM is parsed. Same applies for the `DOMContentLoaded` event.

Zopsss: thanks for the clarification!

Zopsss: we can safely add `defer async` to all the scripts

stoyanK7: Shall we do that in a separate issue? I'll open it if you are okay with that. Just so this PR and issue remain scoped to the UI flickering issue.

romani: Let's merge this PR, and extend all others scripts in separate PR

stoyanK7:
Converting this to a draft. There is a race condition I did not take into account.

src/site/resources/js/checkstyle.js
```diff
@@ -158,5 +158,11 @@ function trimCodeBlock(codeBlock) {
...
 }
 
-window.addEventListener("load", setBodyColumnMargin);
+if (document.readyState === "loading") {
+    document.addEventListener("DOMContentLoaded", setBodyColumnMargin);
+}
+else {
+    setBodyColumnMargin();
+}
+
```
stoyanK7, line 167: This `if-else` is mandatory to prevent a race condition where the `DOMContentLoaded` event might fire before the listener is created. Taken straight out of the docs[^1]:

> Sometimes your script may run after the `DOMContentLoaded` event has already fired. This typically happens when the script runs asynchronously. Common scenarios include: [...]
> - A script that is included via `<script async>`.
> 
> [...] In these cases, you should check the document's `readyState` before adding a `DOMContentLoaded` listener, or your setup logic may not execute at all. [...]
> Consider the following script file in isolation:
> ```javascript
> function doSomething() {
>   console.info("DOM loaded");
> }
> 
> if (document.readyState === "loading") {
>   // Loading hasn't finished yet
>   document.addEventListener("DOMContentLoaded", doSomething);
> } else {
>   // `DOMContentLoaded` has already fired
>   doSomething();
> }
> ```
> The script can't enforce how it's included by the HTML. If it's included via `<script async>`, or it's dynamically injected, then by the time it executes, `DOMContentLoaded` has already fired. To ensure that `doSomething()` always runs when the script loads, we need to have two paths, one that immediately runs `doSomething` if the document is already loaded, and another that runs `doSomething` once the document is loaded.

---
This race condition can happen with the `DOMContentLoaded` event, but not with `load`[^2]:
> The `load` event is fired when the whole page has loaded, including all dependent resources such as stylesheets, scripts (**including async**, deferred, and module scripts), iframes,

[^1]: [Document: DOMContentLoaded event - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete)
[^2]: [Window: load event - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event)

stoyanK7:
Github, generate website

src/site/site.xml
stoyanK7: I created an issue for the other scripts:
- https://github.com/checkstyle/checkstyle/issues/18045

stoyanK7: @Zopsss Please let me know if there's anything else to do.

Zopsss: Sorry for the delay. Everything looks fine. Thanks a lot for detailed explanations!

romani:
@smita1078 , and see comment from you, please finish review

romani:
Ok , looks like @smita1078 is busy with other activities, we moving forward, we can always improve in separate PRs

smita1078:
> Ok , looks like @smita1078 is busy with other activities, we moving forward, we can always improve in separate PRs

Sorry, missed it! 

> @smita1078 , and see comment from you, please finish review

Yes, I went through the discussion — the race condition that @stoyanK7 mentioned aligns with the concern I had earlier regarding potential conflicts when using async. I see that it’s already been taken into account in the latest changes. All good from my side 👍

Issue #17728, opened 2025-09-02
False negative: missing/tab spacing between type and variable not fully detected


Tieshiyang:
I have read check documentation: https://checkstyle.sourceforge.io/checks/whitespace/genericwhitespace.html
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

From: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace
> Beyond where required by the language or other style rules, and apart from literals, comments and Javadoc, a single ASCII space also appears in the following places only.
• Between the type and variable of a declaration.

Code: 
```java
import java.util.List;

class Test {
  boolean	withTab1; // Expected 2 Violations for TAB and no space

  List<String>items; // Expected 1 Violation for no space
  List<String>	withTab2; // Expected 2 Violations for TAB and no space

  int[]numbers; // Expected 1 Violation for no space
  String[][]	withTab3; // Expected 2 Violations for TAB and no space
}

```

Cli: 
```cmd
$ java -jar checkstyle-10.26.1-all.jar -c google_checks.xml Test.java 
Starting audit...
[WARN] Test.java:5:10: Line contains a tab character. [FileTabCharacter]
[WARN] Test.java:7:14: GenericWhitespace '>' should followed by whitespace. [GenericWhitespace]
[WARN] Test.java:8:15: Line contains a tab character. [FileTabCharacter]
[WARN] Test.java:11:13: Line contains a tab character. [FileTabCharacter]
Audit done.
```

Formatter:
```cmd
$ java -jar google-java-format-1.28.0-all-deps.jar Test.java > FormattedCode.java
$ diff -Naru Test.java FormattedCode.java
--- Test.java   2025-09-02 10:55:55.998605900 +0800
+++ FormattedCode.java  2025-09-02 10:56:09.892189400 +0800
@@ -1,13 +1,11 @@
 import java.util.List;
-import java.util.Map;

 class Test {
-  boolean      withTab1; // Expected 2 Violations for TAB and no space
+  boolean withTab1; // Expected 2 Violations for TAB and no space

-  List<String>items; // Expected 1 Violation for multiple spaces
-  List<String> withTab2; // Expected 2 Violations for TAB and no space
-
-  int[]numbers; // Expected 1 Violation for no space
-  String[][]   withTab3; // Expected 2 Violations for TAB and no space
[...]

PR #18048, opened 2025-11-04
Issue #17728: Add support for TYPE token in WhitespaceAfterCheck

stoyanK7:
## Issue
- #17728 

## Description
PR is still in draft state. We have to decide if it's maybe better to make a separate check for the problem in the PR - https://github.com/checkstyle/checkstyle/issues/17728#issuecomment-3478122778

## TODO
- [ ] Add input file in `src/it/resources/com/google/checkstyle/test/chapter4formatting/rule462horizontalwhitespace`

Diff Regression config: https://gist.githubusercontent.com/stoyanK7/c945ccc86ba9c276b8273166450310a1/raw/fadc1096c3837004872272371b8bff5455e8da2a/config.xml
Diff Regression patch config: https://gist.githubusercontent.com/stoyanK7/0b4ea73cb2d6f5446636dbe769394efb/raw/974d365b3b8b442e0fb38e479e2a4e9486e39ced/gistfile1.txt


stoyanK7:
GitHub, generate report by config from src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/whitespaceafter/InputWhitespaceAfterType.java

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAfterCheck.java
```diff
@@ -158,9 +162,75 @@ private static boolean isFollowedByWhitespace(DetailAST targetAST, int... line)
 
             followedByWhitespace = codePoint == ';'
                 || codePoint == ')'
+                || codePoint == ','
```
stoyanK7, line 165: One problem is that this change has an impact on other tokens (not only on `TYPE`), because the method is shared.

romani:
This approach for triggering will not work, as no such token in master branch of checkstyle.
I think we need here old diff report workflow, that has two configs.

stoyanK7:
Github, generate report

stoyanK7:
There are cases where this solution becomes opinionated:

For example `String... someStr`. Do we consider `String...` as a single type? It is composed of `TYPE` and `ELLIPSIS`. Should it be `String ... someStr` or `String... someStr`.

Or `int/*Some javadoc*/ myNums`. Should the space be after the `int` or after the Javadoc? Some repos seem okay with no space between the two, others have it.

Or C-Style Arrays `char someChars[];`. These require extra attention, because the `[]` is part of the `char` TYPE.

IMO, we should create a new check for square brackets only - similar to GenericWhitespace.

romani:
>IMO, we should create a new check for square brackets only - similar to GenericWhitespace.

Thanks a lot for analysis. Please create new issue, let's agree on naming, this PR should be linked as justification on new Check.

stoyanK7:
Closing this. Created issue for new check
- #18064

PR #18059, opened 2025-11-06
Issue #11166: Remove getFileContents() usage from UnusedImportsCheck

gianmarcoschifone:
Issue #11166
This PR fixes the problem for `UnusedImportsCheck`.
~~The existing logic from `WriteTagCheck` was reused and extended with an additional condition to handle cases where the Javadoc is located under an `ANNOTATION` node.
I also created a helper method to build a `TextBlock`, which was required by the other methods involved.
New tests have been added to ensure full branch coverage for this change.
If this approach is acceptable, it might be worth considering moving the method used to retrieve Javadoc from the AST into a more general utility class, such as `JavadocUtil.java`, since it could be reused by other checks.
Would you like me to open a separate issue for that potential refactoring?~~

Edit: 
This PR refactors UnusedImportsCheck to extend AbstractJavadocCheck instead of AbstractCheck, enabling native support for both standard AST tokens and Javadoc tokens.


romani:
Please make ci green.

gianmarcoschifone:
> Please make ci green.

Done, I guess there are too many new suppressions, but I'd like to receive a feedback before trying to solve them

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -312,14 +322,11 @@ private void processStaticImport(DetailAST ast) {
...
-    @SuppressWarnings("deprecation")
     private void collectReferencesFromJavadoc(DetailAST ast) {
-        final FileContents contents = getFileContents();
-        final int lineNo = ast.getLineNo();
-        final TextBlock textBlock = contents.getJavadocBefore(lineNo);
-        if (textBlock != null) {
-            currentFrame.addReferencedTypes(collectReferencesFromJavadoc(textBlock));
+        final DetailAST javadoc = findJavadocComment(ast);
+        if (javadoc != null) {
+            final TextBlock javadocBlock = buildTextBlockFromJavadoc(javadoc);
```
romani: This is a good milestone. I hope you learned a lot about Checks implementation.
Let's get to next level, more correct and quick and reliable.
https://checkstyle.org/writingjavadocchecks.html

We can convert this Check  be aware of javadoc AST, and simply get content of links in reliable way.
As all will be pre-parsed for you, and you just need analyze tree of javadoc.

romani (changes requested):
Yes, we do not allow pitest mutations.

Item:

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
gianmarcoschifone: > This is a good milestone. I hope you learned a lot about Checks implementation. Let's get to next level, more correct and quick and reliable. https://checkstyle.org/writingjavadocchecks.html
> 
> We can convert this Check be aware of javadoc AST, and simply get content of links in reliable way. As all will be pre-parsed for you, and you just need analyze tree of javadoc.

Thanks for the feedback! I’ll start working on this in a few days as I’m quite busy over the next days.

gianmarcoschifone: > This is a good milestone. I hope you learned a lot about Checks implementation. Let's get to next level, more correct and quick and reliable. https://checkstyle.org/writingjavadocchecks.html
> 
> We can convert this Check be aware of javadoc AST, and simply get content of links in reliable way. As all will be pre-parsed for you, and you just need analyze tree of javadoc.

Hi Romani, I'd start working on this. I'm not sure if you meant to
- make `UnusedImports` extend `AbstractJavadocCheck`, adding the javadoc part or
- create a new check that extends `AbstractJavadocCheck` and that only focuses on unused imports on javadoc or
- something else I didn't understand

romani: >make UnusedImports extend AbstractJavadocCheck, adding the javadoc part

this.

gianmarcoschifone: Hi @romani ,

I started thinking about how to convert `UnusedImportsCheck` to extend `AbstractJavadocCheck`, and I’d like to confirm the intended direction.

From the current design, `AbstractJavadocCheck` looks focused on Javadoc only: it fixes `getDefaultTokens` / `visitToken` / `beginTree` / `finishTree` to work with `BLOCK_COMMENT_BEGIN` and the Javadoc AST, and it no longer dispatches normal Java tokens like `IMPORT`, `STATIC_IMPORT`, `IDENT`, `OBJBLOCK`, `SLIST` through `visitToken`. In other words, it does not seem designed to drive a normal code analysis that needs the full Java AST.

To keep `UnusedImportsCheck` working while using the Javadoc AST, my current idea is:

- On the first Javadoc block, use `getBlockCommentAst()` to reach the compilation unit root and run a small internal visitor that:
  - collects imports and static imports (reusing the existing `processImport` / `processStaticImport` logic)
  - collects normal code references via the existing `Frame` / `processIdent` logic and scope handling
- In `visitJavadocToken(DetailNode)`, use the Javadoc AST to extract the links

Is this roughly what you had in mind, or would you prefer a different structure?

romani: As you extend such javadoc abstract class, you will receive visit calls for java and javadoc tokens.

So y will have all, logic just need to properly operate on required calls to collect data and analyze at the end of file

gianmarcoschifone: > As you extend such javadoc abstract class, you will receive visit calls for java and javadoc tokens.
> 
> So y will have all, logic just need to properly operate on required calls to collect data and analyze at the end of file

I do receive visit calls for java tokens but I can't do anything with them since the methods to override are all declared as final in `AbstractJavadocCheck`. What am I missing?

romani: You need https://github.com/checkstyle/checkstyle/blob/0fbf62e8eb988fa44ecd46142f551da58c36e3e6/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/SummaryJavadocCheck.java#L151
https://github.com/checkstyle/checkstyle/blob/0fbf62e8eb988fa44ecd46142f551da58c36e3e6/src/main/java/com/puppycrawl/tools/checkstyle/checks/FinalParametersCheck.java#L131

And you need special tokens 

Oh, I see final here 
https://github.com/checkstyle/checkstyle/blob/0fbf62e8eb988fa44ecd46142f551da58c36e3e6/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck.java#L305

Strange.... I will look a around later on

romani: It was final from the beginning.
https://github.com/checkstyle/checkstyle/commit/ff12c1b012a1b18da8a1cd2101efd0173c612534#diff-432335a67ad4cc120a97acc8d7c46b62dc5eb3c7ba64cfc5203f9ade29987171R150

Looks like we were afraid that user might misuse tokens.

romani: Please remove final, override it your Check, call super method at th beginning, to make sure javadoc will be parsed.

Keep pushing to GitHub to let me see your progress

gianmarcoschifone:
New behaviour introduces a problem while processing imports, since a comment can be placed like this
https://github.com/checkstyle/checkstyle/blob/0fbf62e8eb988fa44ecd46142f551da58c36e3e6/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImports.java#L31-L32
I thought of adding some lines in `FullIndent.extractFullIndent` in order to skip comments, is it ok?

romani:
Imports can be only in javadoc comments https://checkstyle.org/writingjavadocchecks.html#What_is_Javadoc_comment and in special format. All others comments should be ignored.

gianmarcoschifone:
> Imports can be only in javadoc comments https://checkstyle.org/writingjavadocchecks.html#What_is_Javadoc_comment and in special format. All others comments should be ignored.

Yes sure, sorry I wasn't clear. What I meant is that if the AST now includes comments, it breaks the current imports processing.
`UnusedImports.processImport()` is used to collect the imports and it calls `FullIdent.extractFullIdent()`, but if there is a case where a comment is in between `DOT` and `IDENT`, it doesn't recognize the import.
Like in this case:
https://github.com/checkstyle/checkstyle/blob/0fbf62e8eb988fa44ecd46142f551da58c36e3e6/src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImports.java#L31-L32

image

My solution was to make `FullIdent.extractFullIdent()` skip all comments, as I did in last push.
Let me know if that looks ok.

romani:
>it breaks the current imports processing.

code needs to be adjusted to handle all properly.

>FullIdent.extractFullIdent()

do any update you think should happen. You have set of test, user visible behavior should change much, with AST based parsing of javadoc , we should be able to detect more imports in javadoc comments `/**`.  No other chanages are expected.

gianmarcoschifone:
Hi @romani ,
I have some questions regarding this check.
- Should something like `{@code {@link List}}` be recognized as an usage of List's import? The current behaviour of UnusedImports and IntelliJ do that, but there is no "immediate" way from the javadoc ast since it's not recognized as a proper reference (it's just a text inside a code tag).
- Also, now that UnusedImports extends AbstractJavadocCheck class, some javadoc comments raise javadoc parsing errors. I don't know if this is a desired behaviour for UnusedImports but if so there are a lot of CI errors regarding this in external projects. If it is not a desired feature, should I override the visitToken method of AbstractJavadocCheck and ignore parsing errors?

romani:
Please look at javac and javadoc tool of jdk.
If there is compilation problem we don't support such cases, if compiled we should support.

>some javadoc comments raise javadoc parsing errors. I don't know if this is a desired behaviour

Please share link. We need to run javadoc tool to see rendering html, if javadoc tool gives error, we might not need to support it. We mush support only what gives no errors from javadoc.

gianmarcoschifone:
Thank you for helping!

Regarding the first point, javadoc tool doesn't recognize it as a reference and parses it as text. Since CI of idea inspection complains, I think i can `@SuppressWarnings("JavadocReference")` on `UnusedImportsCheck`.

image

***
This is an example of parsing javadoc error. On CI reports there are others.
https://github.com/CS-SI/Orekit/blob/9b38997636ea3c72b66431b49730787ff659d437/src/main/java/org/orekit/propagation/semianalytical/dsst/forces/DSSTThirdBody.java#L1396

Also there are some errors like this, where this is now marked as unused. It is actually unused though, since it's always accessed as `GeoMagneticFieldFactory.FieldModel` and never as `GeoMagneticFieldFactory`.
https://github.com/CS-SI/Orekit/blob/9b38997636ea3c72b66431b49730787ff659d437/src/main/java/org/orekit/propagation/events/MagneticFieldDetector.java#L25

romani:
@gianmarcoschifone , if javadoc tool does not recongnize it , we should not recognize it also, completely ok to treat it as text.

gianmarcoschifone:
> @gianmarcoschifone , if javadoc tool does not recongnize it , we should not recognize it also, completely ok to treat it as text.

Alright thank you, what about the CI errors on external projects? What should I do? I mean the second part of my last message

romani:
example:
```
[INFO] --- checkstyle:3.6.0:check (checkstyle-check) @ pmd-jsp ---
[INFO] There are 3 errors reported by Checkstyle 12.2.0-SNAPSHOT with /net/sourceforge/pmd/pmd-checkstyle-config.xml ruleset.
[ERROR] src/test/java/net/sourceforge/pmd/lang/jsp/ast/JspDocStyleTest.java:[891] (imports) UnusedImports: Javadoc comment at column 8 has parse error. Details: no viable alternative at input '</' while parsing HTML_ELEMENT
[ERROR] src/test/java/net/sourceforge/pmd/lang/jsp/ast/JspDocStyleTest.java:[896] (imports) UnusedImports: Javadoc comment at column 8 has parse error. Details: no viable alternative at input '</' while parsing HTML_ELEMENT
[ERROR] src/test/java/net/sourceforge/pmd/lang/jsp/ast/JspDocStyleTest.java:[903] (imports) UnusedImports: Javadoc comment at column 43 has parse error. Details: no viable alternative at input '</' while parsing HTML_ELEMENT
```

we need to send them PR to update their code to have better javadoc.

gianmarcoschifone:
> example:
> 
> ```
> [INFO] --- checkstyle:3.6.0:check (checkstyle-check) @ pmd-jsp ---
> [INFO] There are 3 errors reported by Checkstyle 12.2.0-SNAPSHOT with /net/sourceforge/pmd/pmd-checkstyle-config.xml ruleset.
> [ERROR] src/test/java/net/sourceforge/pmd/lang/jsp/ast/JspDocStyleTest.java:[891] (imports) UnusedImports: Javadoc comment at column 8 has parse error. Details: no viable alternative at input '</' while parsing HTML_ELEMENT
> [ERROR] src/test/java/net/sourceforge/pmd/lang/jsp/ast/JspDocStyleTest.java:[896] (imports) UnusedImports: Javadoc comment at column 8 has parse error. Details: no viable alternative at input '</' while parsing HTML_ELEMENT
> [ERROR] src/test/java/net/sourceforge/pmd/lang/jsp/ast/JspDocStyleTest.java:[903] (imports) UnusedImports: Javadoc comment at column 43 has parse error. Details: no viable alternative at input '</' while parsing HTML_ELEMENT
> ```
> 
> we need to send them PR to update their code to have better javadoc.

Alright I will do it. Thank you

gianmarcoschifone:
Hi @romani , how are you doing?
I've been fixing last couple of things and sending PRs to external projects that use Checkstyle. Hazelcast, pmd and xwiki are done. I'm waiting for Kafka's approval.

I'm not sure on how to deal with Orekit, since I've noticed that CI runs on a fixed (old) version of Orekit (13.0.3), while default branch of Orekit has actually already solved the issues appearing in CI.
Also, in Orekit's pom, I noticed that they're using an old version of Checkstyle (9.3), so if I run a new version of Checkstyle on Orekit, even more violations are detected.

Could you advise on the best way to handle this? Thank you

romani:
>Orekit, since I've noticed that CI runs on a fixed (old) version of Orekit

Send them PR, and update Sha to value after merge

romani:
>heckstyle (9.3), so if I run a new version of Checkstyle on Orekit, even more violations are detected.

No, we use latest by override, and all is good.
You can send them PR to update cs version to latest.

gianmarcoschifone:
@romani CI is green

src/main/java/com/puppycrawl/tools/checkstyle/api/DetailNode.java
```diff
@@ -84,4 +86,13 @@ public interface DetailNode {
...
     DetailNode getPreviousSibling();
+
+    /**
+     * Returns the first child token that makes a specified type.
+     *
+     * @param type the token type to match
+     * @return the matching token, or null if no match
+     */
+    @Nullable
+    DetailNode findFirstToken(int type);
```
romani: please move it to utils class, we should extend API.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -185,47 +173,27 @@ public int[] getRequiredTokens() {
             TokenTypes.IDENT,
             TokenTypes.IMPORT,
             TokenTypes.STATIC_IMPORT,
-            // Definitions that may contain Javadoc...
-            TokenTypes.PACKAGE_DEF,
-            TokenTypes.ANNOTATION_DEF,
```
romani, line 189: why only some are removed ?

gianmarcoschifone: I removed the ones used to retrieve imports' usages in Javadoc, the ones I left are those used to retrieve normal usages of imports in code

gianmarcoschifone:
Hi @romani , just a gentle ping.
I addressed the requested change and rebased on the latest master.
Let me know if anything else is needed. Thanks!

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -92,51 +88,30 @@
  * </li>
  * </ul>
  *
+ * @noinspection JavadocReference
+ * @noinspectionreason IntelliJ treats {@code {@link List}} in this example as a real reference,
```
romani: change javdoc to use `java.util.Set` to avoid this suppression.

src/test/java/com/puppycrawl/tools/checkstyle/utils/AnnotationUtilTest.java
```diff
@@ -248,6 +248,8 @@ public void testContainsAnnotationListWithNoMatchingAnnotation() {
                         )
                 )
         );
+        final DetailAstImpl ident = modifiersAst.getFirstChild().getFirstChild().getFirstChild();
+        ident.setNextSibling(create(TokenTypes.STAR));
```
romani: what is this for ?

we do not allow pure unit testing anymore, All tests should be done by Check execution.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
```diff
@@ -1,54 +1,53 @@
...
-import java.time.LocalTime;
-import java.time.Month;
 import java.util.concurrent.TimeUnit; // violation 'Unused import - java.util.concurrent.TimeUnit.'
-
+import javax.security.auth.callback.Callback;
+import javax.security.auth.callback.CallbackHandler;
+import java.io.IOException;
 
 /**
- * @link {Collections::emptyEnumeration}
```
romani, line 31: why we removing this ? and respected import was also removed.

if we want to make new Input, lets make new, but old Inputs should stay, they weird, but it is ok, we see how behavior is changing.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -173,8 +148,24 @@ public void finishTree(DetailAST rootAST) {
     }
 
     @Override
-    public int[] getDefaultTokens() {
-        return getRequiredTokens();
```
romani, line 178: can we restore getDefault()  and getAcceptableTokens() ?

they usually required to make it explicit on what is defaut/acceptable/required  .... this is very nuanced area of Checkstyle and easy to make mistakes

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
gianmarcoschifone: Yes you're right, this and other two should not be removed. I removed the ones that would raise Javadoc parsing error though

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
gianmarcoschifone: https://github.com/checkstyle/checkstyle/blob/990f30703694f134a77734a9c50d79c3a427bc97/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck.java#L269-L277
They're already defined in the parent class, exactly as they were defined in the subclass

src/test/java/com/puppycrawl/tools/checkstyle/utils/AnnotationUtilTest.java
gianmarcoschifone: >what is this for ?

I added that because I made a small change on the `FullIdent.extractFullIdent` method, in order to adapt to the fact that AST now includes comments.
During the execution of the test method in `AnnotationUtilTest`, since the type is not defined for the `IDENT`, it would lead to null pointer exception.
Here is the change on `FullIdent` class:
https://github.com/gianmarcoschifone/checkstyle/blob/517c37b69e87f3e677dc7f4fcee174dc2860ef56/src/main/java/com/puppycrawl/tools/checkstyle/api/FullIdent.java#L105-L107


>we do not allow pure unit testing anymore, All tests should be done by Check execution.

So the whole test method needs to be changed? It has to use something like `verifyWithInlineConfigParser`?

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
romani: please still override, I thinks do this explicitly in all Checks.

src/test/java/com/puppycrawl/tools/checkstyle/utils/AnnotationUtilTest.java
romani: >It has to use something like verifyWithInlineConfigParser?

yes.
We banned pure unit testing long time ago, and life become different and implementations become more stable.
We still have bunch of bad examples of testing, but it is legacy, we do not allow it any more.

src/site/xdoc/checks/imports/unusedimports.xml
```diff
@@ -78,6 +78,41 @@
               <td><code>true</code></td>
               <td>5.4</td>
             </tr>
+            <tr>
+              <td><a id="violateExecutionOnNonTightHtml"/><a href="#violateExecutionOnNonTightHtml">violateExecutionOnNonTightHtml</a></td>
+              <td>Control when to print violations if the Javadoc being examined by this check violates the tight html rules defined at <a href="../../writingjavadocchecks.html#Tight-HTML_rules">
+     Tight-HTML Rules</a>.</td>
+              <td><a href="../../property_types.html#boolean">boolean</a></td>
+              <td><code>false</code></td>
+              <td>8.3</td>
```
romani: please looks around a code on how to override since version for xdoc, I think oversider of setter and javadoc is required.

in this Check it will appear in 13.1.0

src/site/xdoc/checks/imports/unusedimports.xml
```diff
@@ -78,6 +78,41 @@
...
+                    REFERENCE</a>
+                , <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.html#PARAMETER_TYPE">
+                    PARAMETER_TYPE</a>
+                , <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.html#THROWS_BLOCK_TAG">
+                    THROWS_BLOCK_TAG</a>
+                , <a href="../../apidocs/com/puppycrawl/tools/checkstyle/api/JavadocTokenTypes.html#EXCEPTION_BLOCK_TAG">
+                    EXCEPTION_BLOCK_TAG</a>
+                  .
+              </td>
+              <td>3.0</td>
```
romani: 13.1.0

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
```diff
@@ -1,54 +1,60 @@
...
  * @link {Collections::emptyEnumeration}
+ * {@link CallbackHandler#handle(Callback[])}
+ * {@linkplain ...}
+ * {@code {@link Set}}
  * {@link TreeSet}
  * {@see TimeUnit#SECONDS}
  * {@link List#size}
- * {@link Set::add}
- * {@link ArrayList<LocalTime>#add( LocalTime )},
- * {@link LinkedList<Integer>#set(int index(,int element)}
```
romani, line 37: please do not change Inputs,
all lines of any kind should stay as is .

We always add more to Inptus, we never remove.

src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheckTest.java
```diff
@@ -406,16 +376,77 @@ public void testImportJavaLinkTag() throws Exception {
...
+        reference.setType(JavadocCommentsTokenTypes.REFERENCE);
+
+        final JavadocNodeImpl child = new JavadocNodeImpl();
+        child.setType(JavadocCommentsTokenTypes.IDENTIFIER);
+        child.setText("");
+        reference.addChild(child);
+
+        assertDoesNotThrow(
+            () -> check.visitJavadocToken(reference),
+            "visitJavadocToken must not throw for REFERENCE with empty identifier text"
```
romani: please expalin why this is required?

src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheckTest.java
```diff
@@ -406,16 +376,77 @@ public void testImportJavaLinkTag() throws Exception {
...
+    public void testImproperJavadocToken() {
+        final UnusedImportsCheck check = new UnusedImportsCheck();
+
+        final JavadocNodeImpl ast = new JavadocNodeImpl();
+        ast.setType(JavadocCommentsTokenTypes.EQUALS);
+        ast.setText("EQUALS");
+
+        try {
+            check.visitJavadocToken(ast);
+            assertWithMessage("IllegalArgumentException is expected").fail();
```
romani, line 447: please put comment over pure untit test to explain reason why not verifyWithInlineConfigParser.
We plan in future to make it as rule, all non-standard test must have explanation to be written.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
gianmarcoschifone: These inputs lead to Javadoc parsing error and so UnusedImports doesn't even run. (It used to work fine before because the check wasn't producing the AST for the javadoc)

gianmarcoschifone: Actually the second one you mentioned is again there, I added it again, the other two raise javadoc parsing error so I removed them

romani: bad syntax is good reason to change

romani: can you check if javadoc tool actually also throw error on such lines, if yes, remove them.

if no .... it means we have parser problem, it is unrelated issues, but stil good to know. Please share details.

gianmarcoschifone: > bad syntax is good reason to change

Do you want me to fix the syntax? I thought the bad syntax was made on purpose

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
romani: lets confirm by javadoc tool that this line was written in bad way.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
gianmarcoschifone: To summarize, the only ones removed are:
```java
* {@link Set::add}
* {@link LinkedList<Integer>#set(int index(,int element)}
* {@link Queue# }
* {@link #Test(LocalDateTime}}
```

And yes, they do raise parsing error both on checkstyle parser and on javadoc tool

```bash
gianmarco@Gianmarcos-MacBook-Air ~/Documents/Projects/testing-checkstyle % javadoc -d docs Test.java
Loading source file Test.java...
Constructing Javadoc information...
Building index for all the packages and classes...
Standard Doclet version 21.0.8+9-LTS
Building tree for all the packages and classes...
Test.java:6: error: unexpected text
 * {@link Set::add}
             ^
Test.java:7: error: unterminated signature
 * {@link LinkedList<Integer>#set(int index(,int element)}
                                                        ^
Test.java:8: error: syntax error in reference
 * {@link Queue# }
                ^
Test.java:9: error: unterminated signature
 * {@link #Test(LocalDateTime}}
                            ^
```

romani: goal was to test usage of types in generics and ...

please fix syntax if you see how to do this.

gianmarcoschifone: While trying to understand this, I found out several lines that don't raise Javadoc parsing error on checkstyle parsing but they do raise error with javadoc tool, e.g.:

**error: incorrect use of inline tag**
The `{` should be at the start (and `::` should be substituted by `#`)
```java
* @link {Collections::emptyEnumeration}
```
**error: unexpected text**
These because it shouldn't be possible to start as `...` or with numbers
```java
* {@linkplain ...}
* {@link PriorityQueue#123invalid}
* {@link 123invalid}
```
**error: reference not found**
These are related to actual definitions inside the class so I don't know if it would be fixable for the parser
```java
* {@link #setUrl(URI)}
* {@link #Test0(Map.Entry)}
* {@link #Test(LocalDate, Collection)}
* {@link #extractParameters(double[], Month)}
* {@link ArrayList<LocalTime>#add( LocalTime )},
```
**error: exception not thrown: java.io.IOException**
This when the method doesn't have the `throws IOException` in the signature
```java
* @exception IOException
```

Let me know if this is expected behaviour.
Btw, I did some testing and research and Javadoc doesn't support generics inside references

src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheckTest.java
gianmarcoschifone: It is required for cases like this
```java
* {@linkplain ...}
```
where the text is empty.
It could just be tested inserting it in an input file but the problem is that there is no way to kill the mutation that pitest generates, unless a proper JUnit test is created

gianmarcoschifone:
@romani I addressed the changes of your review, please let me know if anything else is needed

gianmarcoschifone:
rebased

src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheckTest.java
romani: if we can not reproduce it by any Input, it means and mutation can be applied, we do not need code that is covering impossible syntax.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
romani: > don't raise Javadoc parsing error on checkstyle parsing but they do raise error with javadoc tool,

javadoc tool ERRORs are critical for us, it this tool is give ERROR, we do not guaranty anything, and behavior might be exceptional.
WE need javadoc content to comply with javadoc tool rules.

romani: all our Inputs should pass javadoctool wihtout ERRORs.
CI does not do this , big gap on our side ..... but you can use this rule in decision what is god and bad javadoc content. Content that causing ERROR should be simply removed.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -157,9 +129,26 @@ public void setProcessJavadoc(boolean value) {
...
 
+    /**
+     * Setter to control when to print violations if the Javadoc being examined by this check
+     * violates the tight html rules defined at
+     * <a href="https://checkstyle.org/writingjavadocchecks.html#Tight-HTML_rules">
+     *     Tight-HTML Rules</a>.
+     *
+     * @param shouldReportViolation value to which the field shall be set to
+     * @since 8.3
+     * @propertySince 13.1.0
```
romani: 13.3.0

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -173,58 +162,77 @@ public void finishTree(DetailAST rootAST) {
...
+            JavadocCommentsTokenTypes.THROWS_BLOCK_TAG,
+            JavadocCommentsTokenTypes.EXCEPTION_BLOCK_TAG,
+        };
+    }
+
+    /**
+     * Adds a set of tokens the check is interested in.
+     *
+     * @param strRep the string representation of the tokens interested in
+     * @propertySince 13.1.0
```
romani: 13.3.0

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -173,58 +162,77 @@ public void finishTree(DetailAST rootAST) {
...
+    }
+
+    @Override
+    public void visitJavadocToken(DetailNode ast) {
+        switch (ast.getType()) {
+            case JavadocCommentsTokenTypes.REFERENCE -> processReference(ast);
+            case JavadocCommentsTokenTypes.PARAMETER_TYPE -> processParameterType(ast);
+            case JavadocCommentsTokenTypes.THROWS_BLOCK_TAG,
+                 JavadocCommentsTokenTypes.EXCEPTION_BLOCK_TAG -> processException(ast);
+            default -> throw new IllegalArgumentException("Unknown token type " + ast);
```
romani: I know this causing pure unit testing.
Can we avoid such unit testing if we do `default -> ;` ? no action. 

right now we are very robust in testing before Check release, execution on unknown token is close to impossible.

gianmarcoschifone:
> sorry that review takes so long.
> 
> But this update will create first case when Check will have java tokens and javadoc tokens, it is completely new entity in CS..

Yes I know, no problem!

src/test/resources/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsWithLinkAndMethodTag.java
gianmarcoschifone: So it is ok to remove inputs?

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
gianmarcoschifone: jacoco complains about the branch coverage though, how to fix this?

gianmarcoschifone:
I'm confused. For the overridden methods used only for versioning, if I don't add the noinspection tag then IDEA inspection raises error, if instead I add it then Qodana raises error. What should I do? @romani 
```java
* @noinspection RedundantMethodOverride
* @noinspectionreason Display module's unique property version
```

romani:
@gianmarcoschifone , yes, this is weird, but qodana is more recent version of idea inspection, than docker container that we created.

we migrated to qodana, but I was ok to keep old style inspection to make sure qodana catch same, and be confident that qodana full cover us.
But looks like there is conflict in behavior. 
We need to disable idea inspections

@Atharv3221 , please help us to remove old idea inspections.

@gianmarcoschifone ,  please consider old idea inspection as deprecated, please satisfy qodana.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
romani: yes, you are right.
do you know what demands default ? may be we should deactivate some stastic analysis tool?

src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheckTest.java
```diff
@@ -31,9 +32,13 @@
 import com.google.common.collect.ImmutableMap;
 import com.puppycrawl.tools.checkstyle.AbstractModuleTestSupport;
 import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
+import com.puppycrawl.tools.checkstyle.DetailAstImpl;
+import com.puppycrawl.tools.checkstyle.api.JavadocCommentsTokenTypes;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocNodeImpl;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 
+@SuppressWarnings("UnrecognisedJavadocTag")
```
romani: please share details on this.
not expected, not obvious.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
```diff
@@ -78,7 +78,8 @@ public class XpathRegressionTest extends AbstractModuleTestSupport {
                     "RequireEmptyLineBeforeBlockTagGroup",
                     "SingleLineJavadoc",
                     "SummaryJavadoc",
-                    "WriteTag"
+                    "WriteTag",
+                    "UnusedImports"
```
romani: I think we should avoid this, it is not javadoc focused Check.
our Check does not place violations on Javadoc nodes.
our Check place violations on java nodes, and we do support xpath for suppression.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsAnnotations.java
```diff
@@ -1,10 +1,10 @@
 /*
 UnusedImports
 processJavadoc = (default)true
-
+violateExecutionOnNonTightHtml = (default)false
+javadocTokens = (default)REFERENCE, PARAMETER_TYPE, THROWS_BLOCK_TAG
```
romani: I can not image a case where user would need to configure this property - javadocTokens.
This, I think we just need make required set = availbale set, and this property becomes not configurable, so not visible to users to define.

src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheckTest.java
gianmarcoschifone: Sorry, that is not needed. It was needed for a JUnit method that was there. I'll remove it

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
gianmarcoschifone: I think it was IDEA inspection, I will check

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
gianmarcoschifone: I'm pretty sure I added it because CI was failing otherwise, I will check

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsAnnotations.java
gianmarcoschifone: Yes I agree. CI was failing without this and I didn't know how to just make that not configurable. Should have asked, sorry.
I'll fix it.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
romani: it might, we probable should update validation logic and over this list, as not all AastractJavadoc is actually javadoc focused.

this will lead us more intersting question, .... we use "extends" to get javadoc logic, but may be we should stay AbstractJavaCheck  and use javadoc by composite pattern?
we do not need customization of behavior over javadoc,  we just need reliable parsing and walking over javadoc to detect types.

gianmarcoschifone: This is the error I get if I don't put UnusedImports on XpathRegressionTest
```bash
[ERROR]   XpathRegressionTest.validateIncompatibleJavadocCheckNames:165 INCOMPATIBLE_JAVADOC_CHECK_NAMES should contains all descendants of AbstractJavadocCheck
value of      : getSimpleNames(...)
unexpected (1): UnusedImports
---
expected      : [JavadocBlockTagLocation, JavadocMethod, SummaryJavadoc, JavadocType, JavadocTagContinuationIndentation, AtclauseOrder, JavadocMissingLeadingAsterisk, JavadocLeadingAsteriskAlign, JavadocMissingWhitespaceAfterAsterisk, WriteTag, SingleLineJavadoc, RequireEmptyLineBeforeBlockTagGroup, NonEmptyAtclauseDescription, JavadocStyle, MissingDeprecated, JavadocParagraph]
but was       : [MissingDeprecated, JavadocMethod, NonEmptyAtclauseDescription, UnusedImports, JavadocMissingWhitespaceAfterAsterisk, RequireEmptyLineBeforeBlockTagGroup, JavadocLeadingAsteriskAlign, AtclauseOrder, WriteTag, JavadocType, JavadocBlockTagLocation, SingleLineJavadoc, JavadocStyle, SummaryJavadoc, JavadocTagContinuationIndentation, JavadocMissingLeadingAsterisk, JavadocParagraph]
```


>this will lead us more intersting question, .... we use "extends" to get javadoc logic, but may be we should stay AbstractJavaCheck and use javadoc by composite pattern?
we do not need customization of behavior over javadoc, we just need reliable parsing and walking over javadoc to detect types.

You are right, and I agree that conceptually UnusedImportsCheck does not really need to customize Javadoc behavior itself, only to extract referenced types.
From my perspective, the current approach still looks reasonable if AbstractJavadocCheck is treated primarily as a reusable Javadoc parsing and traversal facility, rather than strictly as a base class only for Javadoc-style validation logic

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
gianmarcoschifone: >do you know what demands default ?

Stil Jacoco, if there is no default branch it still sees that as just not covered branch @romani

gianmarcoschifone:
@romani Just a ping to let you know that the changes of your review are addressed. We need to decide for the XPathRegressionTest and for the default branch of the switch

gianmarcoschifone:
Hi @romani , when you have a moment could you please review the PR? Thanks.

romani:
Please resolve conflict

gianmarcoschifone:
> Please resolve conflict

Done, there is one error regarding checkstyle/contribution, I think it's not related to this PR

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -157,9 +129,24 @@ public void setProcessJavadoc(boolean value) {
...
 
+    /**
+     * Setter to control when to print violations if the Javadoc being examined by this check
+     * violates the tight html rules defined at
+     * <a href="https://checkstyle.org/writingjavadocchecks.html#Tight-HTML_rules">
+     *     Tight-HTML Rules</a>.
+     *
+     * @param shouldReportViolation value to which the field shall be set to
+     * @since 8.3
+     * @propertySince 13.3.0
```
romani: 13.4.0

sorry for slow review from my side

gianmarcoschifone: > 13.4.0

Done

> sorry for slow review from my side

No problem! I know there is a lot of other work

gianmarcoschifone:
Rebased, CI errors are unrelated

gianmarcoschifone:
@romani ping, please whenever you have time let me know if anything else is needed

romani:
We finally fixed all CIs, so I rebased this PR in web UI, to see green final state before merge.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsAnnotations.java
romani: @gianmarcoschifone , please look at test failure and see where test is failing , it should be improved to skip it as we do with "tokens" property for java Checks.

romani: inline config should have only properties that user can change, so this weird to see it here and not see this property in HTML documentation.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
romani: @gianmarcoschifone , please remove this Check from Xpath suppression, lets find a way to update test to not flag it.
It is not Javadoc focused Check and Xpath suppression works for it.
https://github.com/checkstyle/checkstyle/blob/master/src/it/java/org/checkstyle/suppressionxpathfilter/imports/XpathRegressionUnusedImportsTest.java

romani:
@gianmarcoschifone , please address last two review items above.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/checks/imports/unusedimports/InputUnusedImportsAnnotations.java
gianmarcoschifone: My bad, I had fixed it for resources but forgot to also edit the non compilable ones.
It is now fixed.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
gianmarcoschifone: done

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
```diff
@@ -252,6 +255,18 @@ public void validateInputFiles() throws Exception {
...
 
+    private static boolean isPureJavadocCheck(Class<?> check) {
+        try {
+            final AbstractJavadocCheck instance =
+                    (AbstractJavadocCheck) check.getDeclaredConstructor().newInstance();
+            return Arrays.equals(instance.getRequiredTokens(),
+                    new int[] {TokenTypes.BLOCK_COMMENT_BEGIN});
+        }
+        catch (ReflectiveOperationException exc) {
+            throw new IllegalStateException(exc);
```
romani, line 266: throw new IllegalStateException("Reflection operation problem for " + check, exc);

romani: @gianmarcoschifone , please do this in separate PR, with commit "minor: ......"

gianmarcoschifone:
> thanks a lot for big update !!

Thanks a lot for the review and for merging this! I really appreciate it

src/test/java/com/puppycrawl/tools/checkstyle/internal/XpathRegressionTest.java
gianmarcoschifone: done, #19268

Issue #17919, opened 2025-10-12
Java25's import-module support


enexusde:
https://openjdk.org/jeps/511

```sh
$ java --version
openjdk 25 2025-09-16
OpenJDK Runtime Environment (build 25+36-Ubuntu-124.04.2)
OpenJDK 64-Bit Server VM (build 25+36-Ubuntu-124.04.2, mixed mode, sharing)

$ javac --version
javac 25

$ cat Test.java
import module java.base;

$ javac Test.java
# no output

$ RUN_LOCALE="-Duser.language=en -Duser.country=US"
$ java $RUN_LOCALE -jar checkstyle-12.1.1-all.jar -c checkstyle/src/main/resources/google_checks.xml Test.java
Starting audit...
com.puppycrawl.tools.checkstyle.api.CheckstyleException: Exception was thrown while processing Test.java
	at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:313)
	at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:227)
	at com.puppycrawl.tools.checkstyle.Main.runCheckstyle(Main.java:429)
	at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:347)
	at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:206)
	at com.puppycrawl.tools.checkstyle.Main.main(Main.java:130)
Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: IllegalStateException occurred while parsing file /home/stoyan/Documents/github/Test.java.
	at com.puppycrawl.tools.checkstyle.JavaParser.parse(JavaParser.java:104)
	at com.puppycrawl.tools.checkstyle.TreeWalker.processFiltered(TreeWalker.java:192)
	at com.puppycrawl.tools.checkstyle.api.AbstractFileSetCheck.process(AbstractFileSetCheck.java:101)
	at com.puppycrawl.tools.checkstyle.Checker.processFile(Checker.java:341)
	at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:300)
	... 5 more
Caused by: java.lang.IllegalStateException: 1:14: mismatched input 'java' expecting ';'
	at com.puppycrawl.tools.checkstyle.JavaParser$CheckstyleErrorListener.syntaxError(JavaParser.java:254)
	at org.antlr.v4.runtime.ProxyErrorListener.syntaxError(ProxyErrorListener.java:41)
[...]

PR #17982, opened 2025-10-23
Issue #17052: Add support for Flexible constructor bodies

mohitsatr:
resolves #17052

PR #13989, opened 2023-11-03
Issue #13982: Run regression testing in CI over openjdk21

nrmancuso:
Closes #13982

PR #18079, opened 2025-11-11
Issue #17919: Add support for Java25's 'import module'

stoyanK7:
**Issue**
- #17919
---
**Context**
Link to JEP: [JEP 511: Module Import Declarations](https://openjdk.org/jeps/511)
Link to JLS: [7.5.5. Single-Module-Import Declarations ](https://docs.oracle.com/javase/specs/jls/se25/html/jls-7.html#jls-SingleModuleImportDeclaration)

Grammar of import declarations now includes `import module` clauses:
```diff
 ImportDeclaration:
   SingleTypeImportDeclaration
   TypeImportOnDemandDeclaration
   SingleStaticImportDeclaration
   StaticImportOnDemandDeclaration
+  ModuleImportDeclaration

+ModuleImportDeclaration:
+  import module ModuleName;
```

A [`ModuleName` is defined](https://docs.oracle.com/javase/specs/jls/se25/html/jls-6.html#jls-6.5) as
```
ModuleName:
  Identifier
  ModuleName . Identifier
```
therefore,  `import module java.base.*;` is not possible.

---
**AST representation**
`import module java.base;` is converted to:
```text
MODULE_IMPORT -> import
 |--LITERAL_MODULE -> module
 |--DOT -> .
 |   |--IDENT -> java
 |   `--IDENT -> base
 `--SEMI -> ;
```

---
**Diff reports**
- [**Import Check Regression Report** on all projects in `contribution/checkstyle-tester/projects-to-test-on.properties`](https://stoyank7.github.io/17919-default-diff-report/diff/index.html): no changes detected. Most likely because none of the projects use this feature yet
- [**Import Check Regression Report** on `openjdk25` only](https://stoyank7.github.io/17919-openjdk25-diff-report/index.html): (+4,-4): Files that use `import module` fail on `master` but not with the changes in this PR. [Here are the 4 files in question](https://github.com/search?q=repo%3Aopenjdk%2Fjdk25u+language%3Ajava+%2F%5Eimport+module+.%2B%3B%2F+path%3A*.java&type=code)
- [**ANTLR Regression Report**](https://stoyank7.github.io/antlr-diff/index.html): not sure what to make out of this report

---
**Performance Regression Report** (from latest CI for `Java` (not `Javadoc`))
https://github.com/checkstyle/checkstyle/actions/runs/19371658568/job/55428858356?pr=18079#step:6:202
```text
Execution Time Baseline: 390.7 s
Average Execution Time: 394.65 s
Execution Time Difference: 1.0100%
Difference is within the maximum allowed difference (1.0100%      <= 10%).
```


stoyanK7:
Separate issue, but:

Checks that will need an update after this PR:
- [CustomImportOrder](https://checkstyle.org/checks/imports/customimportorder.html#CustomImportOrder)
    - Users might want to separate module imports from other imports. [The JEP (511) also suggests that](https://openjdk.org/jeps/511#:~:text=If%20a%20source%20file%20has%20a%20mix%20of%20different%20kinds%20of%20import%20declarations%20then%20grouping%20them%20by%20kind%20may%20further%20improve%20readability%3B%20for%20example%3A)
- [ImportControl](https://checkstyle.org/checks/imports/importcontrol.html#ImportControl)?
    - It is not aware of `MODULE_IMPORT` token. 
- [ImportOrder](https://checkstyle.org/checks/imports/importorder.html#ImportOrder)
    - It is not aware of `MODULE_IMPORT` token. 
- [RedundantImport](https://checkstyle.org/checks/imports/redundantimport.html#RedundantImport)
    - It is not aware of `MODULE_IMPORT` token.

romani:
@mahfouz72, please assist in this PR.
your knowledge is in high demand

stoyanK7:
EDIT: Nevermind, it's `.ci/check-performance-regression.sh`. We mentioned this in https://github.com/checkstyle/checkstyle/pull/18082

---

~~Can somebody tell me how to [run those performance benchmark reports](https://github.com/checkstyle/checkstyle/pull/17982#issuecomment-3441805908)?~~
> ===================== BENCHMARK SUMMARY ====================
Execution Time Baseline: 390.7 s
Average Execution Time: 527.54 s
Execution Time Difference: 35.0200%
Difference exceeds the maximum allowed difference (35.0200%      > 10%)!

Wouldn't be a bad idea to mention that in [`docs/GRAMMAR_UPDATES.md`](https://github.com/checkstyle/checkstyle/blob/master/docs/GRAMMAR_UPDATES.md).

romani:
> Wouldn't be a bad idea to mention that in

Please send separate PR

>Checks that will need an update after this PR:

Yes, but let's merge grammar update first, to avoid exception. Violations from checks are easy to suppress, exception is hard blocker for users.

------


We need to run diff report on jdk 25 sources, there tests might have some special usages of " import module ". 
I think we still have config with jdk21, so extra update will be required.

stoyanK7:
> Please send separate PR

Done at https://github.com/checkstyle/checkstyle/pull/18082

> Yes, but let's merge grammar update first, to avoid exception. Violations from checks are easy to suppress, exception is hard blocker for users

Agreed

> We need to run diff report on jdk 25 sources, there tests might have some special usages of " import module ".

Ran diff report on https://github.com/openjdk/jdk25u and updated PR description. Report is [here](https://stoyank7.github.io/17919-openjdk25-diff-report/index.html).

>  I think we still have config with jdk21, so extra update will be required.

@romani Do you mean an update like this? :
- https://github.com/checkstyle/checkstyle/pull/13989

romani:
>Do you mean an update like this? :

Yes, this is best form of it, to make sure we don't have regression.
We can upgrade jdk21 to jdk25 job. We don't need jdk21 any more if we can have jdk25.
There might be some other jdk25 syntax features there that we don't support, we can just add them to filters, to not block us now.

If this is too much effort, we can do this later on.


Far dream, update at https://github.com/checkstyle/test-configs/blob/08d32248192ae7c7b3fb9c8135dc1fa12fe6fadd/extractor/src/main/resources/list-of-projects.yml#L36 to let jdk25 be used in daily diff report generation.

src/main/java/com/puppycrawl/tools/checkstyle/api/TokenTypes.java
```diff
@@ -6691,6 +6691,62 @@ public final class TokenTypes {
...
+     * </pre>
+     *
+     * @see <a href="https://docs.oracle.com/javase/specs/jls/se25/html/jls-7.html#jls-7.5.5">
+     *     Java Language Specification, &sect;7.5.5</a>
+     * @see <a href="https://docs.oracle.com/javase/specs/jls/se25/html/jls-7.html#jls-7.7">
+     *     Java Language Specification, &sect;7.7</a>
+     * @see #IMPORT
+     * @see #MODULE_IMPORT
+     *
+     * @since 12.1.2
```
romani: Let's put 12.2.0

I marked issue with "new feature" , so after merge , version in pom will be bumped

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/java25/InputImportModuleCustom.java
```diff
@@ -0,0 +1,8 @@
+// non-compiled with javac: reference to non-existent module 'mymodule'
```
romani, line 1: Please split into two inputs.
We keep non compiled as minimal as possible.

And we need javac25 job in circle CI, to make sure compiled input is actually compiled

romani (changes requested):
Thanks a lot 

Items

stoyanK7:
> Yes, this is best form of it, to make sure we don't have regression.
We can upgrade jdk21 to jdk25 job. We don't need jdk21 any more if we can have jdk25.
There might be some other jdk25 syntax features there that we don't support, we can just add them to filters, to not block us now.

Opened:
- https://github.com/checkstyle/checkstyle/issues/18089

src/main/java/com/puppycrawl/tools/checkstyle/api/TokenTypes.java
stoyanK7: Done.

src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/java25/InputImportModuleCustom.java
stoyanK7: > Please split into two inputs.

Done.

> And we need javac25 job in circle CI, to make sure compiled input is actually compiled

Done in
- https://github.com/checkstyle/checkstyle/pull/18095

mahfouz72:
> * [**ANTLR Regression Report**](https://stoyank7.github.io/antlr-diff/index.html): not sure what to make out of this report

This report checks whether any regressions occurred in the AST. As you can see, there are no AST regressions in the Checkstyle repository. The differences shown are only in exception line numbers, which is expected because we changed in parser source code, so the report indicates everything is good.

mahfouz72:
@romani @stoyanK7 

Should we update `project-to.test-on.properties` in the contribution repo, and anywhere else it's used, so that the report generation workflow defaults to JDK 25 instead of JDK 21?

stoyanK7:
@mahfouz72 

Agreed. Are these 3 the only places that will need an update?
- https://github.com/checkstyle/test-configs/blob/master/extractor/src/main/resources/list-of-projects.yml#L36
- https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/projects-to-test-on.properties#L12
- https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/projects-to-test-on-for-github-action.properties#L44

romani:
Yes, it will be most complete migration

romani:
I do not see trigering a diff report by github action to make sure there is no human effect and report is source of truth without anybody's special involvement.

@stoyanK7 , did you generate `Diff reports` on local?

stoyanK7:
@romani Yes, I've only generated local reports. Shall I generate a report with GitHub Actions too?

romani:
I am trying to think how to do this ....
if you do https://github.com/checkstyle/checkstyle/pull/18079#issuecomment-3534078417 , it will be just a simple comment in this PR and all will be done automatically.
Good part of it we will confident that report are done withtout humans, so no mistakes in executions of any kind with our diff tools.

other option:
we can merge this PR and finish update to jdk25 in test configs and test later, release is two weeks, so we have some time.

What is your preference ?

**nuances**: in diff report, there are only 3 files that has module imports, all of them are in our Inputs, so jdk25 might not that valuable immediately now.
So we can run Github action on all Checks now, on jdk21 and just confirm there is no diff.

stoyanK7:
I prefer to keep this PR open and get the JDK25 diff report working first - https://github.com/checkstyle/checkstyle/pull/18079#issuecomment-3534078417

I'll focus on that and try to open a PR in the following days.

- https://github.com/checkstyle/test-configs/pull/229
- https://github.com/checkstyle/contribution/pull/994

romani:
test-configs is migrated.

2 updates for contribution are left to be done.

romani:
Github, generate report for ImportOrder/all-examples-in-one

romani:
Weird result, no diff.
It could mean that we did some mistake in excluding folders files.

I donot see exclude on files at https://github.com/checkstyle/test-configs/blob/main/AbbreviationAsWordInName/Example1/list-of-projects.yml#L38 that were in diff at https://stoyank7.github.io/17919-openjdk25-diff-report/openjdk25/index.html

stoyanK7:
> I do not see exclude on files at [...] that were in diff at [...]

It's not that. These are not supposed to be excluded.

The problem comes from the `javaParseExceptionSeverity` setting:
- On my local run, it's set to `warning`.
- On CI run, it's set to `ignore`.

I opened a PR for that: https://github.com/checkstyle/test-configs/pull/230

---
On a separate note: What is the purpose of excluding files (apart from performance reasons?)? Since @mahfouz72 added the `skipFileOnJavaParseException` property in PR #15204, I don't see a reason for maintaining those lists of excluded files.

If the file does not compile on base and patch branch, that's okay, diff will not be affected. Configs have already set that property to `true`.

You can see that [my local run was done with 0 exclusions](https://stoyank7.github.io/17919-openjdk25-diff-report/projects-to-test-on.html) and only reported the parsing exceptions about `import module`:
> `openjdk25|git|https://github.com/openjdk/jdk25u.git|master||`

Maybe there is a good reason I am missing?

romani:
I did a post in PR for test-configs, looks like we doomed to use custom runs when ANTLR is updated.

So we should just merge this PR, as all is done properly,  I just forgot and mislead you, sorry.

mahfouz72:
> On a separate note: What is the purpose of excluding files (apart from performance reasons?)? Since @mahfouz72 added the `skipFileOnJavaParseException` property in PR #15204, I don't see a reason for maintaining those lists of excluded files.

I actually raised this same question back when I implemented the `skipFileOnJavaParseException` feature, but I don’t think I received or remember a clear answer.

Even now, I’m still convinced that we no longer need to maintain a manual “exclude files” list. Once this property is set to `true` in the regression report config, it already guarantees that we automatically skip:

1. **Non-compilable files** that are kept in repositories for some purposes.
2. **Compilable files** that our parser currently fails to handle.

This behavior is exactly what we want in the check report run. Because of that, maintaining a manual excluded-files list seems unnecessary.

But regardless, I see that at this point we should move on and merge this PR

---

> Weird result, no diff.
> It could mean that we did some mistake in excluding folders files.

It is expected that this file was ignored on the base branch because of the property. However, we now see no diff because the checks don't produce any violations for this file.

Issue #18045, opened 2025-11-03
Add `async` and/or `defer` attributes to site scripts


stoyanK7:
Continuation of https://github.com/checkstyle/checkstyle/pull/18036#discussion_r2483750695

In that PR, we added `async` and `defer` to the `checkstyle.js` `<script>` tag to ensure it loads faster and does not cause flickering on user's screen:
```diff
- <script type="text/javascript" src="$relativePath/js/checkstyle.js"></script>
+ <script type="text/javascript" src="$relativePath/js/checkstyle.js" defer async></script>
```
The same could be applied to the other 3 scripts in the `src/site/resources/js` directory:
- `anchors.js`
- `copy-clipboard.js`
- `google-analytics.js`

This will ensure that all scripts are fetched in the background and do not block the HTML parser.

> [!IMPORTANT]
> Each script has to be examined before applying the 2 attributes. This is because race conditions can occur if there is a `DOMContentLoaded` event listener inside the script. See https://github.com/checkstyle/checkstyle/pull/18036#discussion_r2487706217 for more information.

romani:
@stoyanK7 , do you want to apply this?

stoyanK7:
@romani Yes, I can do that.

stoyanK7:
Fix is merged:
-  #18086

PR #18086, opened 2025-11-13
Issue #18045: Add 'async' and 'defer' attributes to site scripts

stoyanK7:
Resolves issue
- #18045

### DIFF
--- a/src/site/resources/js/copy-clipboard.js
+++ b/src/site/resources/js/copy-clipboard.js
@@ -1,4 +1,4 @@
-document.addEventListener("DOMContentLoaded", function () {
+function initCopyButtons() {
   const xmlHeader = `<?xml version="1.0"?>
 <!DOCTYPE module PUBLIC
   "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
@@ -55,4 +55,11 @@ document.addEventListener("DOMContentLoaded", function () {
         });
     });
   });
-});
+};
+
+if (document.readyState === "loading") {
+    document.addEventListener("DOMContentLoaded", initCopyButtons);
+}
+else {
+    initCopyButtons();
+}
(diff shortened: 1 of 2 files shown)


src/site/resources/js/copy-clipboard.js
```diff
@@ -55,4 +55,11 @@ document.addEventListener("DOMContentLoaded", function () {
...
   });
-});
+};
+
+if (document.readyState === "loading") {
+    document.addEventListener("DOMContentLoaded", initCopyButtons);
+}
+else {
+    initCopyButtons();
+}
```
stoyanK7, line 65: Since the script uses the `DOMContentLoaded` event, we must do this `if-else` check to prevent a race condition. 

Just like in https://github.com/checkstyle/checkstyle/pull/18036#discussion_r2487706217. See that comment for more details.

src/site/site.xml
```diff
@@ -52,9 +52,11 @@
       <!-- Will be added to every generated page -->
       <![CDATA[
         <script type="text/javascript" src="$relativePath/js/checkstyle.js" defer async></script>
-        <script type="text/javascript" src="$relativePath/js/anchors.js"></script>
-        <script type="text/javascript" src="$relativePath/js/google-analytics.js"></script>
-        <script type="text/javascript" src="$relativePath/js/copy-clipboard.js"></script>
+        <script type="text/javascript" src="$relativePath/js/anchors.js" defer async></script>
+        <script type="text/javascript"
+                src="$relativePath/js/google-analytics.js" defer async></script>
```
stoyanK7, line 57: `anchors.js` uses the `window.load` event, so we do not have to protect against a race condition (unlike `copy-clipboard.js`). See https://github.com/checkstyle/checkstyle/pull/18036#discussion_r2487706217 

`google-analytics.js` does not use any events, so there is also no need to protect against race conditions.

stoyanK7:
Github, generate website

stoyanK7:
Anchors and 'copy-clipboard' work, nothing broke :+1:

Issue #18097, opened 2025-11-14
Java 22 test files that were marked as compilable but are not


stoyanK7:
In issue #18092 and PR #18094 we discovered multiple test files that were marked as `// non-compiled with javac: compilable with Java22` but were never picked up by the CI job and were never tested. It turned out that these files were not compilable to begin with. These files are:

<details>
<summary>src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/antlr4/InputAntlr4AstRegressionPatternsInIfStatement.java</summary>

https://github.com/checkstyle/checkstyle/blob/master/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/antlr4/InputAntlr4AstRegressionPatternsInIfStatement.java

```bash
$ javac src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/antlr4/InputAntlr4AstRegressionPatternsInIfStatement.java
src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/antlr4/InputAntlr4AstRegressionPatternsInIfStatement.java:15: error: illegal start of type
        if (o instanceof (String s && s.length() > 4)) { // parenthesized pattern, `PATTERN_DEF`
                         ^
src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/antlr4/InputAntlr4AstRegressionPatternsInIfStatement.java:15: error: ')' expected
        if (o instanceof (String s && s.length() > 4)) { // parenthesized pattern, `PATTERN_DEF`
                                  ^
src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/antlr4/InputAntlr4AstRegressionPatternsInIfStatement.java:15: error: illegal start of expression
        if (o instanceof (String s && s.length() > 4)) { // parenthesized pattern, `PATTERN_DEF`
                                                     ^
3 errors
```
</details>

<details>
<summary>src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/antlr4/InputAntlr4AstRegressionPatternsInSwitch.java</summary>

https://github.com/checkstyle/checkstyle/blob/master/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/antlr4/InputAntlr4AstRegressionPatternsInSwitch.java

```bash
[...]

PR #18103, opened 2025-11-15
Issue #18097: Split and update unsupported Java syntax test files

stoyanK7:
Resolves issue
- #18097

This PR:
1. Moves unsupported test cases to dedicated files
    - All test cases that rely on Java features no longer supported in the current parser have been moved to a new file, `InputAntlr4AstRegressionNoLongerSupportedByJava.java`, to preserve parser coverage.
    - The `MissingSwitchDefaultCheck` logic also depends on unsupported syntax. To keep coverage, some test cases were extracted into a second file: `InputMissingSwitchDefaultCaseLabelElementsNoLongerSupportedByJava.java`.
    - Both files were annotated with
    ```java
    // non-compiled with javac: Compilable with Java17
    // until https://github.com/checkstyle/checkstyle/issues/18104
    ```
3. Updates all non-compilable test files (detailed in the description of issue #18097) to be compilable. This was done by:
    - Replacing all `case p && b` with `case p when b`
    - Removing all `case (p && b)`
    - Replacing all `case default` with `default`
    - Replacing `case null, String s` with `case null, default` or `case null: ... <newline> case String s ...`
4. Removes `// non-compiled with javac: until https://github.com/checkstyle/checkstyle/issues/18097` from all files updated in step 2
5. Moves all files updated in step 2 from the `resources-noncompilable/` to the  `resources/` folder


stoyanK7:
After further examination, we shouldn't remove all test files, but rather individual cases within these files. The feature was not dropped, but it was refined. Which means _some syntax_ got dropped _but not all_.

mahfouz72:
Yes, the syntax that was removed is pattern guards using `&&`, which have now been replaced with the `when` keyword.
Also, the parenthesized pattern is removed.
ideally, we should not delete non-compilable code in tests. Instead, we should update it so it becomes valid under the latest Java grammar if applicable.


**Double check that my corrected examples are actually compilable**


### Example Refactor

**Old guarded pattern:**

```java
switch (o) {
    case String s && s.length() > 4: // guarded pattern, PATTERN_DEF
        break;
}
```

**Updated valid form using `when`:**

```java
switch (o) {
    case String s when s.length() > 4: // guarded pattern, PATTERN_DEF
        break;
}
```

---

### Parenthesized pattern example

**Old (invalid) form:**

```java
if (o instanceof (String s && s.length() > 4)) { // parenthesized pattern, PATTERN_DEF
    ...
}
```

**Correct updated form:**

```java
if (o instanceof String s && s.length() > 4) { // parenthesized pattern, PATTERN_DEF
    ...
}
```

stoyanK7:
Thanks for that. I am reading the JEPs to get a better understanding of what happened. I'll get back to it soon :+1:

romani:
yes, ideally we should convert them to compile-able.
if removal of non-compiled results in coverage failure, it is good reason to keep them until we update grammar to not support such cases.

stoyanK7:
@mahfouz72 

Yes, spot on. Your corrected examples are compileable.

Some historical context, if anybody needs it:
1. In [JEP 406: Pattern Matching for switch (Preview)](https://openjdk.org/jeps/406)(Java 17), pattern matching for switch was introduced as a preview feature.
    - This introduced the _guarded patterns_ `p && b`:
        - Inside a `switch`: `case String s && s.length() > 0`
        - Inside an `instanceof`: `e instanceof (String s && s.length() > 1)`
2. In [JEP 427: Pattern Matching for switch (Third Preview)](https://openjdk.org/jeps/427)(Java 19), the `&&` in _guarded patterns_ got swapped with `when`
    - > Guarded patterns are replaced with `when` clauses in switch blocks.
    - `case String s && s.length() > 0` becomes invalid and `case String s when s.length() > 0` is now valid
    - `e instanceof (String s && s.length() > 1)` is a bit unclear to me. JEP 406 mentions it, whereas 427 does not mention it, so I assume this is where it got dropped. Only `e instanceof String s && s.length() > 1` is valid since then

stoyanK7:
I will try to make the examples compileable and see if we get coverage failure. If we get coverage failure, I guess I have to try to update the grammar as well (and update the scope of this issue).

romani:
I recommend to update as much as we can do without grammar update, isolate all such experimental syntax in special inputs, merge what we have. In Separate PR deal with removal of features from grammar, as it might be treaky and end up in breaking changes or fact that better to keep support in grammar.

stoyanK7:
@romani @mahfouz72 Please see PR description for a summary of what I've done here. It's what @romani suggested in the previous comment. I opened issue #18104 for removal of grammar (+ update of MissingSwitchDefaultCheck)

romani:
@mahfouz72 , please help to do first review

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/missingswitchdefault/InputMissingSwitchDefaultCaseLabelElements.java
```diff
@@ -52,18 +55,19 @@ static void m(Object o) {
...
+            case String s when s.length() > 2 ->
                 System.out.println("The string longer than 2 chars");
             default -> System.out.println("default!");
         }
 
         switch(o) {
-            case default, null ->
-                System.out.println("The rest (including null)");
+            case null -> System.out.println("The rest (including null)");
+            default -> System.out.println("The rest (including null)");
```
mahfouz72: I think we should update it as shown below. This will ensure that when we make future changes in `MissingSwitchDefaultCheck`, this case will be covered. I don’t see any existing input that covers it.

```suggestion
            case null, default -> System.out.println("The rest (including null)");
```

---
> 'null' can only be used as a single case label or paired only with 'default'

stoyanK7: Done. Good catch :+1:

mahfouz72 (approved):

romani (approved):
Thanks a lot !!!

PR #18150, opened 2025-11-21
Issue #15456: Define violation messages for NoWhiteSpaceAfterCheck

vivek-0509:
#15456: Defined violation messages for NoWhiteSpaceAfterCheck


vivek-0509:
Hi @romani,

While running the pitest-ant profile, I noticed that CI reports an unnecessary PIT suppression for the following code in CheckstyleAntTask#getListeners:

final int formatterCount = Math.max(1, formatters.size());

The mutation that removes the List::size() call now appears to be killed by existing tests, and CI suggests removing this entry from pitest-ant-suppressions.xml.

Would you like me to open a PR to remove this outdated suppression?

But, locally I still see this mutant in my PIT report, so I wanted to mention that I can add an explicit test to cover this case (e.g testing multiple formatter scenarios) if you’d prefer that approach.

Please let me know how you’d like me to proceed.

romani:
Restarted, passed.

romani (approved):
Nice

Good from first try

vivek-0509:
> Nice
> 
> Good from first try

Thank you 🚀

Issue #17746, opened 2025-09-05
Refine newly added reuse of metadata and xdoc to simplify Javadocs


SteLeo1602:
Discovered in #17675 

- [x] 1. 
Consider the following methods:
https://github.com/checkstyle/checkstyle/blob/a4f3688e557e35da86a00cfd70ad7a479e1f5cba/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java#L1457-L1481
https://github.com/checkstyle/checkstyle/blob/a4f3688e557e35da86a00cfd70ad7a479e1f5cba/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java#L1392-L1449

These two methods essentially do the same thing, except `getDescriptionFromJavadocForXdoc` also processes extracted description for Xdoc. 

Reusing `getFirstParagraphFromJavadoc` for `getDescriptionFromJavadocForXdoc` just with addition of xdoc wrapping would make functionality easier.

---
- [x] 2.
As result of new implementation, the following error pops up that is temporarily suppressed:

 `Class Data Abstraction Coupling is 8 (max allowed is 7) classes [ArrayDeque, 
Checker, DefaultConfiguration, DetailNode, LinkedHashMap, MacroExecutionException, PackageObjectFactory, Pattern]. [ClassDataAbstractionCoupling]`

Fixing it, potentially with the side-effect of integrating`getFirstParagraphFromJavadoc` to `getDescriptionFromJavadocForXdoc` would also make functionality easier and project better.

---
- [x] 3.
Suppressions

Remove coverage suppressions for `MetadataGeneratorUtil` and `XmlMetaWriter`, most of which, if not all, will be removed once migration of simplified javadocs to all modules is complete

XmlMetaWriter coverage used to be 0.77 https://github.com/checkstyle/checkstyle/blob/2bb01a1878475a87b9dda110cd7228eeabd85d7f/pom.xml#L1490

Complains about coverage in https://github.com/checkstyle/checkstyle/blob/2bb01a1878475a87b9dda110cd7228eeabd85d7f/src/main/java/com/puppycrawl/tools/checkstyle/meta/XmlMetaWriter.java#L131

---

Currently, MetadataGeneratorUtil has 0.95 of true coverage.

Complains about coverage for such catch statements:
[...]

PR #18194, opened 2025-11-27
Issue #17746: Made implementation for finding end of description not …

SteLeo1602:
Issue #17746

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5183,6 +5183,8 @@
               <excludedMethods>
                 <!-- unkilled in generated code https://github.com/hcoles/pitest/issues/255 -->
                 <param>isFileExists</param>
+                <!-- Its mutation potentially causes the cascade change in generated xml files -->
+                <param>getAllNodesOfType</param>
               </excludedMethods>
               <excludedClasses>
                 <class>com.puppycrawl.tools.checkstyle.utils.OsSpecificUtil</class>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/TrailingCommentCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/TrailingCommentCheck.java
@@ -86,10 +86,10 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  * are even more important in the age of automated refactorings.
  * </p>
  *
- * @since 3.4
  * @noinspection HtmlTagCanBeJavadocTag
  * @noinspectionreason HtmlTagCanBeJavadocTag - encoded symbols were not decoded
  *      when replaced with Javadoc tag
+ * @since 3.4
  */
 @StatelessCheck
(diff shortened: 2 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/TrailingCommentCheck.java
```diff
@@ -86,10 +86,10 @@
  * are even more important in the age of automated refactorings.
  * </p>
  *
- * @since 3.4
  * @noinspection HtmlTagCanBeJavadocTag
  * @noinspectionreason HtmlTagCanBeJavadocTag - encoded symbols were not decoded
  *      when replaced with Javadoc tag
+ * @since 3.4
```
SteLeo1602, line 92: Moved `@since` tag to demonstrate new implementation's ability to keep description extracting behavior intact

romani: Good, please make ci green , and let's merge this PR

SteLeo1602: Done

pom.xml
```diff
@@ -5168,6 +5168,8 @@
               <excludedMethods>
                 <!-- unkilled in generated code https://github.com/hcoles/pitest/issues/255 -->
                 <param>isFileExists</param>
+                <!-- Its mutation potentially causes the cascade change in generated xml files -->
+                <param>getAllNodesOfType</param>
```
romani, line 5187: Please share details on why this method should be excluded.
It doesn't have class name, so can match any.

If we try to avoid mutation on meta specific method, I am ok to rename method to be somewhat destructive and special

romani: https://github.com/search?q=repo%3Acheckstyle%2Fcheckstyle%20getAllNodesOfType&type=code

you might did it for:
```
src/main/java/com/puppycrawl/tools/checkstyle/site/ModuleJavadocParsingUtil.java

     * @return the {@code @since} tag start node, or {@code null} if not found
     */
    public static DetailNode getModuleSinceVersionTagStartNode(DetailNode moduleJavadoc) {
        return JavadocUtil.getAllNodesOfType(
                moduleJavadoc, JavadocCommentsTokenTypes.JAVADOC_BLOCK_TAG).stream()
```

but in this case, can we use getModuleSinceVersionTagStartNode as method name that should not be mutated ?
Or exclude whole ModuleJavadocParsingUtil ? as we know it will be moved out to separate project soon.

SteLeo1602: I don't think so, because there is no problem with mutating `ModuleJavadocParsingUtil` in any way. Mutated `getAllNodesOfType` causes the rest just to work wrong

romani:
rebased in web, will merge if CI pass.

Issue #18027, opened 2025-11-01
Resolve Pitest Suppressions - ant


romani:
As a part of this issue, we have to kill the mutation listed below and remove the suppression from [pitest xxxx suppression.xml](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/)

To understanding how to kill the mutation 
Visit [Wiki page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports), [Pitest-docs](https://pitest.org/) and expected actions and workflow at https://github.com/checkstyle/checkstyle/issues/12341.

Target is to remove all `<mutation` from [Pitest-ant](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-ant-suppressions.xml)

Example of expected update https://github.com/checkstyle/checkstyle/pull/18210

vivek-0509:
I am on 
 ```
<mutation unstable="false">
    <sourceFile>CheckstyleAntTask.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask</mutatedClass>
    <mutatedMethod>getListeners</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
    <description>removed call to java/util/List::size</description>
    <lineContent>final int formatterCount = Math.max(1, formatters.size());</lineContent>
  </mutation>
```

**Comment by aclfe** [comment]
I am on  

```
  <mutation unstable="false">
    <sourceFile>CheckstyleAntTask.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask</mutatedClass>
    <mutatedMethod>createRootModule</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
    <description>removed call to java/lang/Package::getName</description>
    <lineContent>Checker.class.getPackage().getName() + &quot;.&quot;, moduleClassLoader);</lineContent>
  </mutation>
```

**Comment by aclfe** [comment]
I am on 
```
  <mutation unstable="false">
    <sourceFile>CheckstyleAntTask.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask</mutatedClass>
[...]

PR #18210, opened 2025-11-29
Issue #18027: Resolve pitest for formatter count in CheckstyleAntTask

vivek-0509:
Issue #18027:

Added test to kill mutant in CheckstyleAntTask.getListeners()

This PR adds a test case to kill a mutant that removes the call to formatters.size() in CheckstyleAntTask.getListeners() method.

**Test Implementation:**
- Added testMultipleFormattersProduceOutputs() method
- Creates two formatters with separate output files
- Verifies both formatters produce non-empty output files
- If the mutant is present (formatters.size() removed), formatterCount would be 1 instead of 2, causing only the first formatter to be processed and the second file to be missing/empty, causing the test to fail

**Helper Method:**
- Added createPlainFormatter() helper method for reusability

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
@@ -44,6 +44,7 @@ import org.apache.tools.ant.types.FileSet;
 import org.apache.tools.ant.types.Path;
 import org.apache.tools.ant.types.resources.FileResource;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 
 import com.google.common.base.Splitter;
 import com.google.common.collect.Iterables;
@@ -72,6 +73,9 @@ public class CheckstyleAntTaskTest extends AbstractPathTestSupport {
             "InputCheckstyleAntTaskConfigCustomRootModule.xml";
     private static final String NOT_EXISTING_FILE = "target/not_existing.xml";
     private static final String FAILURE_PROPERTY_VALUE = "myValue";
+
+    @TempDir
+    public File temporaryFolder;
 
     @Override
     protected String getPackageLocation() {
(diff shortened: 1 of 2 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
```diff
@@ -997,4 +997,43 @@ public void testMaxWarningDefault() throws IOException {
...
+    @Test
+    public void testMultipleFormattersProduceOutputs() throws IOException {
+        final CheckstyleAntTask antTask = getCheckstyleAntTask();
+        antTask.setFile(new File(getPath(VIOLATED_INPUT)));
+        antTask.setFailOnViolation(false);
+
+        final File firstOutput = new File("target/ant_task_multi_formatter_1.txt");
+        final File secondOutput = new File("target/ant_task_multi_formatter_2.txt");
+        Files.deleteIfExists(firstOutput.toPath());
+        Files.deleteIfExists(secondOutput.toPath());
```
romani: you probably meant temp files
try:
https://github.com/search?q=repo%3Acheckstyle%2Fcheckstyle%20%22Files.createTempFile%22&type=code

vivek-0509: Thanks for the clarification.
I initially considered using @TempDir (as I did in a previous PR), but I saw that several existing tests in this class still use files under target/ (e.g., lines 467, 512, 537, 604, 631), so I followed that pattern for consistency.
If the preferred direction is to use Files.createTempFile(temporaryFolder.toPath(), ...), I’m happy to update this test accordingly. I can also help refactor the other tests in this file to follow the same approach if you'd like.

vivek-0509: Thank you for the suggestion. I tried using Files.createTempFile(temporaryFolder.toPath(), ...), but it fails the forbiddenapis check. I noticed MainTest.java uses this pattern, but it's excluded from forbiddenapis checks 
Since CheckstyleAntTaskTest isn't excluded, I've used new File(temporaryFolder, "filename.txt") instead, which passes the check and works correctly. Happy to adjust if you prefer a different approach.

romani (approved):
Thanks a lot.

Next time please share violation error message, it will helps to confirm it.

vivek-0509:
Sure will keep in mind from the next time.

Issue #18074, opened 2025-11-10
JAR for 12.1.1 is missing org.slf4j package


aeveris:
The JAR for checkstyle 12.1.1 seems to be missing the `org.slf4j` package which causes it to raise an exception when attempting to run it with the `sarif` output format:

```bash
/code # java -jar /opt/checkstyle/checkstyle.jar -f sarif -c /sun_checks.xml .
Exception in thread "main" java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
	at org.reflections.Reflections.<clinit>(Reflections.java:115)
	at com.puppycrawl.tools.checkstyle.meta.XmlMetaReader.readAllModulesIncludingThirdPartyIfAny(XmlMetaReader.java:75)
	at com.puppycrawl.tools.checkstyle.SarifLogger.loadModuleMetadata(SarifLogger.java:212)
	at com.puppycrawl.tools.checkstyle.SarifLogger.<init>(SarifLogger.java:188)
	at com.puppycrawl.tools.checkstyle.Main$OutputFormat.createListener(Main.java:588)
	at com.puppycrawl.tools.checkstyle.Main.createListener(Main.java:516)
	at com.puppycrawl.tools.checkstyle.Main.runCheckstyle(Main.java:421)
	at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:347)
	at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:206)
	at com.puppycrawl.tools.checkstyle.Main.main(Main.java:130)
Caused by: java.lang.ClassNotFoundException: org.slf4j.LoggerFactory
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	... 10 more
```
---

Checkstyle 12.1.0:

<img width="820" height="394" alt="Image" src="https://github.com/user-attachments/assets/59024e03-e4a5-4e82-8c76-b90b4354e5d3" />

Checkstyle 12.1.1:

<img width="790" height="372" alt="Image" src="https://github.com/user-attachments/assets/52ffbc83-876b-497b-be08-6ede0649a546" />

romani:
[...]

PR #18211, opened 2025-11-29
Issue #18074: Update slf4j to 2.0.17 and suppress convergence check

mureinik:
Following up on the discussion from #18081, this PR updates slf4j's version to the latest available version (2.0.17), and suppresses the convergence check seen in the dependabot PR.

Relates to #18074


romani (approved):
Thanks a lot

Issue #18200, opened 2025-11-27
Maven try to download artifact that supposed to be taken from local m2 local repostory


romani:
example https://dev.azure.com/romanivanovjr/romanivanovjr/_build/results?buildId=33925&view=logs&jobId=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&j=c902ebb4-c9f8-5f09-4e17-ff78fbbc842e&t=9ca98c81-ff64-58f0-9d03-a23ac1c4a111

but this becomes annoying problems across multiple CIs.
For some reason maven cache is not working or does not have some artifacts and maven tries to download them and fails on network operation.

Expected: maven should not download anything and use local cache. Only in PRs where we add new dependency, we expect download from web by maven.  

```
cmd: ./.ci/validation.sh test
openjdk 21.0.9 2025-10-21 LTS
OpenJDK Runtime Environment Temurin-21.0.9+10 (build 21.0.9+10-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.9+10 (build 21.0.9+10-LTS, mixed mode, sharing)
    at org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolveArtifacts (DefaultArtifactResolver.java:261)
    at org.eclipse.aether.internal.impl.DefaultRepositorySystem.resolveDependencies (DefaultRepositorySystem.java:353)
    at org.apache.maven.project.DefaultProjectDependenciesResolver.resolve (DefaultProjectDependenciesResolver.java:182)
    at org.apache.maven.lifecycle.internal.LifecycleDependencyResolver.getDependencies (LifecycleDependencyResolver.java:224)
    at org.apache.maven.lifecycle.internal.LifecycleDependencyResolver.resolveProjectDependencies (LifecycleDependencyResolver.java:136)
    at org.apache.maven.lifecycle.internal.MojoExecutor.ensureDependenciesAreResolved (MojoExecutor.java:355)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:313)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:174)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:75)
[...]

PR #18213, opened 2025-11-30
Issue #18200: Optimize Maven to use local cache and prevent network failures

guptapratykshh:
Issue: #18200

Maven tries to download artifacts from remote repositories even when they exist in the local Maven cache. This causes build failures in CI when network operations fail, even though all required artifacts are already cached.

Implemented automatic offline mode detection that prevents Maven from downloading artifacts when they're already in the local cache. The solution checks if Maven cache is complete after restore and enables `--offline` flag when cache is sufficient, preventing unnecessary network downloads while still allowing network access for PRs that add new dependencies.

Changes:
- Added `.ci/check-maven-cache.sh` - Script to validate cache completeness
- Added `.mvn/settings.xml` - Maven configuration to prioritize local repository
- Updated `.ci/validation.sh` - Added offline flag support to test functions
- Updated `azure-pipelines.yml` - Integrated cache check and offline mode

### DIFF
--- a/.ci/check-maven-cache.sh
+++ b/.ci/check-maven-cache.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+# Check if Maven cache is good enough to run offline
+# Fixes issue #18200 where Maven tries to download artifacts that are already cached
+
+MAVEN_REPO="${MAVEN_REPO_LOCAL:-${MAVEN_CACHE_FOLDER:-$HOME/.m2/repository}}"
+
+check_cache_completeness() {
+  if [ ! -d "$MAVEN_REPO" ]; then
+    echo "Cache directory missing: $MAVEN_REPO" >&2
+    return 1
+  fi
+
+  cache_size=$(du -sm "$MAVEN_REPO" 2>/dev/null | cut -f1 2>/dev/null || echo "0")
+  
+  if [ -z "$cache_size" ] || [ "$cache_size" = "0" ]; then
+    echo "Cache size check failed" >&2
+    return 1
+  fi
+  
+  # Ensure cache_size is a number before comparison
+  case "$cache_size" in
+    ''|*[!0-9]*)
+      echo "Cache size is not a valid number: $cache_size" >&2
+      return 1
+      ;;
+  esac
+  
+  # Very lenient threshold - 10MB indicates cache was restored
+  # Azure Pipelines cache should have much more than this
+  if [ "$cache_size" -lt 10 ]; then
+    echo "Cache too small ($cache_size MB), probably incomplete" >&2
+    return 1
+  fi
+
+  # Check if we have any Maven artifacts at all
+  # Just verify the cache isn't completely empty
+  if [ -z "$(ls -A "$MAVEN_REPO" 2>/dev/null)" ]; then
+    echo "Cache directory is empty" >&2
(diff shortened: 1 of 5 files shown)


romani:
Please make CI green.

guptapratykshh:
Please look into this PR and review it. @romani

romani:
@guptapratykshh , thanks a lot .
Please share your reasoning why m2 cache might not have a jar or may be maven to defect and still try to resovle jar in web.

@stoyanK7 , please help to review this PR.

I still do not think e fully undersand root reason to allow fancy scripting.

guptapratykshh:
> @guptapratykshh , thanks a lot . Please share your reasoning why m2 cache might not have a jar or may be maven to defect and still try to resovle jar in web.
> 
> @stoyanK7 , please help to review this PR.
> 
> I still do not think e fully undersand root reason to allow fancy scripting.

Hello @romani 
Maven looks for metadata updates in remote repositories even when artifacts are stored in local cache. Even if updatePolicy=never is set in settings.xml, maven still makes network requests unless we use --offline. If maven central returns a 502 or 500 error during these checks, builds fail even if artifacts are cached.

Cache might not have a jar because of a cache miss during first-time builds, cache corruption or eviction, new dependencies added in PRs when pom.xml changes, or missing artifacts from an older build.

Also if we need conditional logic because we can not always use offline mode. Pull requests with new dependencies need network access, and we can not always allow network access due to errors like Maven Central 502 or 500, which cause unnecessary failures. The settings.xml file with updatePolicy set to never reduces checks, but it does not stop all network requests. Maven still tries network checks without offline flag. The script checks if cache is usable. 
It enables offline mode when the cache is sufficient and only allows network access in pull requests when pom.xml file changes for new dependencies. This approach helps continuous integrations reuse m2 cache and avoids unnecessary network requests.

stoyanK7:
@romani, I am not sure what the problem is, to begin with. Especially since it's not easily reproducible :/

@guptapratykshh Could you please provide Maven docs mentioning some of this behaviour? It would help a lot

guptapratykshh:
Hi @stoyanK7,

The Maven documentation confirms this behavior. According to https://maven.apache.org/ref/3.9.11/maven-settings/settings.html#class_repository:

> `updatePolicy`: "never" **(only if it does not exist locally)**

This means `updatePolicy=never` still allows Maven to download if it thinks an artifact does not exist locally. Maven may still check remote repositories to verify artifact existence , request missing metadata files(`.pom`, `maven-metadata.xml`)and make network requests when it can't find artifacts in local cache

This is why we see network requests in issue logs even with `updatePolicy=never`. The `--offline` flag is only way to completely prevent network requests, which is why script conditionally enables it when the cache is sufficient.

The combination of `settings.xml` with `updatePolicy=never` + conditional `--offline` flag addresses both:
- Reducing unnecessary update checks (`updatePolicy=never`)
- Preventing all network requests when cache is good (`--offline`)

guptapratykshh:
ping @romani @stoyanK7

romani:
I think that https://github.com/checkstyle/checkstyle/pull/18256 is right way to do this. We need find way to update cache when we building master branch

romani:
energy is gone from this PR, closing it to let somebody else in web to continue

Issue #18127, opened 2025-11-18
Update imports checks to support module imports


stoyanK7:
In

- Issue #17919 
- and PR #18079

we added support for Java 25's `import module` syntax. What is next is to follow the [New Language Feature Check Integration Process](https://github.com/checkstyle/checkstyle/blob/master/docs/NEW_LANGUAGE_FEATURE_INTEGRATION_PROCESS.md) and update [imports checks](https://checkstyle.org/checks/imports/index.html).

Checks that need an update are:
- [ ] [`CustomImportOrder`](https://checkstyle.org/checks/imports/customimportorder.html#CustomImportOrder)
    - The check currently supports 5 groups of imports (`STATIC`, `SAME_PACKAGE`...). We could add a 6th group called `MODULE`.
- [x] [`UnusedImports`](https://checkstyle.org/checks/imports/unusedimports.html#UnusedImports)
    - We could update the documentation and mention that we cannot check module imports and that it's a limitation. Checkstyle is not a compiler, so there is no way for us to know which classes/packages are imported from that module. 
    - Resolved in
        - PR https://github.com/checkstyle/checkstyle/pull/18216
- [x] [`IllegalImport`](https://checkstyle.org/checks/imports/illegalimport.html#IllegalImport)
    - The check has 2 properties called `illegalClasses` and `illegalPkgs`. We could add an additional property called `illegalModules`.
    - Resolved in
        - Issue https://github.com/checkstyle/checkstyle/issues/18207
        - PR https://github.com/checkstyle/checkstyle/pull/18220
- [ ] [`ImportControl`](https://checkstyle.org/checks/imports/importcontrol.html#ImportControl)
    - Similar to `IllegalImport`, this check could allow & disallow modules. It currently only does that for classes and packages.
- [ ] [`ImportOrder`](https://checkstyle.org/checks/imports/importorder.html#ImportOrder)
    - The check currently deals (sorting, grouping, separating, etc..) with type and static imports. We could enhance it to also deal with module imports.
- [x] [`RedundantImport`](https://checkstyle.org/checks/imports/redundantimport.html#RedundantImport)
    - The check is only aware of normal imports (`IMPORT`) and static imports (`STATIC_IMPORT`). We could add the new `MODULE_IMPORT` token to be recognized too.
    - Resolved in
        - Issue https://github.com/checkstyle/checkstyle/issues/18171
        - PR https://github.com/checkstyle/checkstyle/pull/18195

> [!IMPORTANT]  
[...]

PR #18216, opened 2025-11-30
doc: note UnusedImports limitation for module import declarations

stoyanK7:
Part of issue
- https://github.com/checkstyle/checkstyle/issues/18127

---

Extended the limitations of `UnusedImports` to mention that the check cannot deal with module imports (`import module ...`). Reason is that _classes and interfaces from exported packages **are imported on demand**_. From the [JEP 511: Module Import Declarations](https://openjdk.org/jeps/511)

> A module import declaration has the form
    `import module M;`
> It imports, **on demand**, all of the public top-level classes and interfaces in
> - The packages exported by the module M to the current module, and [...]

--- 
image


stoyanK7:
Github, generate website

Issue #18026, opened 2025-11-01
Resolve Pitest Suppressions - main


romani:
As a part of this issue, we have to kill the mutation listed below and remove the suppression from [pitest xxxx suppression.xml](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/)

To understanding how to kill the mutation 
Visit [Wiki page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports), [Pitest-docs](https://pitest.org/) and expected actions and workflow at https://github.com/checkstyle/checkstyle/issues/12341.

Target is to remove all `<mutation` from [Pitest-main](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-main-suppressions.xml)

vivek-0509:
Hi @romani ,

I wanted to share the approaches I tried to kill the mutation (removed call to Main::getOutputStreamOptions) and the results, just to confirm I'm heading in the right direction before opening a PR.

### Approaches Tried
- **Input-based tests** using different configs and Java files  
  → No effect, because the mutation occurs *before* any file is processed, so nothing observable changes.

- **Normalizing output-stream options inside `runCheckstyle()`**  
  → Behavior still identical, since both suppression listeners treat `null` the same way.

- **Testing stream/close behavior**  
  → Also no observable difference, because the listener constructors currently accept `null` silently.

All these failed because the mutation is **unobservable** from outside ,both original and mutated code behave the same in all scenarios.

### Final Working Approach
Added a **defensive null-check** in the constructors of:

- `ChecksAndFilesSuppressionFileGeneratorAuditListener`
- `XpathFileGeneratorAuditListener`

(similar to what `DefaultLogger` already does)

This introduces a clear behavioral difference: the mutated code ends up passing `null`, which now triggers an exception, while the original code behaves correctly ,allowing PIT to kill the mutation.

Please let me know if I’m moving in the right direction or if you’d prefer a different approach.

romani:
@vivek-0509 ,
please send PR to let me see all details, do not hesitate to send PR to show your current state and ask question in it. It helps me see all details.

vivek-0509:
Sure, I’ll open a PR with the current changes so you can review the details directly.

romani:
Fix is merged. Suppression file is empty

PR #18219, opened 2025-11-30
Issue #18026: Resolve Pitest suppression for runCheckstyle in main.java

vivek-0509:
Issue #18026:

Kill Mutant: Removed call to `getOutputStreamOptions` in `Main.runCheckstyle()`

### Solution
Added null parameter validation in listener constructors:
- **`XpathFileGeneratorAuditListener`**: Throws `IllegalArgumentException` if `outputStreamOptions` is null
- **`ChecksAndFilesSuppressionFileGeneratorAuditListener`**: Throws `IllegalArgumentException` if `outputStreamOptions` is null

If ⁠ getOutputStreamOptions() ⁠ is removed, ⁠ null ⁠ is passed → ⁠ IllegalArgumentException ⁠ with clear message → test fails with expected exception and mutant is killed.

### Additional Changes
- Added SpotBugs suppression for `CT_CONSTRUCTOR_THROW` in listener constructors (null checks are intentional for mutation testing)
- Added tests `testNullOutputStreamOptions()` in both listener test classes to achieve 100% coverage

### DIFF
--- a/config/pitest-suppressions/pitest-main-suppressions.xml
+++ b/config/pitest-suppressions/pitest-main-suppressions.xml
@@ -1,11 +1,3 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <mutation unstable="false">
-    <sourceFile>Main.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.Main</mutatedClass>
-    <mutatedMethod>runCheckstyle</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/Main::getOutputStreamOptions</description>
-    <lineContent>getOutputStreamOptions(options.outputPath));</lineContent>
-  </mutation>
 </suppressedMutations>
--- a/config/spotbugs-exclude.xml
+++ b/config/spotbugs-exclude.xml
@@ -309,6 +309,9 @@
       <Class name="com.puppycrawl.tools.checkstyle.Checker"/>
       <Class name="com.puppycrawl.tools.checkstyle.checks.TranslationCheck"/>
       <Class name="com.puppycrawl.tools.checkstyle.filters.CsvFilterElement"/>
+      <Class name="com.puppycrawl.tools.checkstyle.XpathFileGeneratorAuditListener" />
+      <Class name=
+           "com.puppycrawl.tools.checkstyle.ChecksAndFilesSuppressionFileGeneratorAuditListener" />
     </Or>
     <Bug pattern="CT_CONSTRUCTOR_THROW"/>
   </Match>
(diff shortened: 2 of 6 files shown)


vivek-0509:
Hii @romani ,
I’ve submitted this PR with a null check in the constructor because it creates a clear behavioral difference for PIT, allowing the surviving mutant to be killed. I’m aware that this does not fully address the broader CT_CONSTRUCTOR_THROW concern discussed in issue #14122, since throwing exceptions from constructors can leave objects in a partially initialized state.

At this point, I’m not sure whether this  approach aligns with what you’d prefer, or if you would recommend exploring a more robust long term solution.

I’m happy to revise the implementation based on your guidance. Please let me know how you’d like to proceed.

vivek-0509:
Hii @romani ,
I’ve submitted this PR with a null check in the constructor because it creates a clear behavioral difference for PIT, helping kill the surviving mutant. I also noticed that other loggers like DefaultLogger and XmlLogger use a similar pattern. After going through issue https://github.com/checkstyle/checkstyle/issues/14122, I understand that this may not fully address the CT_CONSTRUCTOR_THROW concern related to partially initialized objects.

If you prefer a different approach or a more robust long term solution, I’m happy to revise the implementation.
Please guide me on how I should proceed.

config/pitest-suppressions/pitest-main-suppressions.xml
```diff
@@ -1,11 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<suppressedMutations>
-  <mutation unstable="false">
```
romani, line 3: Please keep file and just remove mutation tags, it will help CI to generate good diff

vivek-0509: done

vivek-0509: Can you please re run that one failing check there is some maven plugin resolution issue.

config/spotbugs-exclude.xml
```diff
@@ -309,6 +309,9 @@
       <Class name="com.puppycrawl.tools.checkstyle.Checker"/>
       <Class name="com.puppycrawl.tools.checkstyle.checks.TranslationCheck"/>
       <Class name="com.puppycrawl.tools.checkstyle.filters.CsvFilterElement"/>
+      <Class name="com.puppycrawl.tools.checkstyle.XpathFileGeneratorAuditListener" />
+      <Class name=
+           "com.puppycrawl.tools.checkstyle.ChecksAndFilesSuppressionFileGeneratorAuditListener" />
```
romani, line 314: So we are fixing pitest and growing https://github.com/checkstyle/checkstyle/issues/14122

Unfortunately. But now we value pitest more than this particular spotbugs validation.
We will deal with spotbugs later on.

vivek-0509:
> Thanks a lot, not ideal solution but still step forward

Hey, After going through issue #14122, I’ve been experimenting with approaches to move the validation check before the constructor call. I’ll be raising a PR on it soon.

vivek-0509:
> Thanks a lot, not ideal solution but still step forward

Hey, After going through issue #14122, I’ve been experimenting with approaches to move the validation check before the constructor call. I’ll be raising a PR on it soon.

PR #18234, opened 2025-12-02
Issue #14122: Refactor classes to prevent vulnerability to finalizer …

vivek-0509:
Issue #14122

### Problem
Some concrete classes can throw exceptions inside their constructors, making them vulnerable to **Finalizer Attacks** . A malicious subclass could potentially override `finalize()` and access a partially initialized object.

### Fix
The following classes have been declared `final` to prevent subclassing:

* `SarifLogger`
* `XMLLogger`
* `XpathFilterElement`
* `TranslationCheck`
* `XpathFileGeneratorAuditListener`
* `ChecksAndFilesSuppressionFileGeneratorAuditListener`

### Reasoning
Making these classes `final` strictly prevents subclassing and fully eliminates the attack vector by ensuring no malicious subclass can override finalize().
The only tradeoff is that these classes can no longer be extended. After reviewing the codebase, none of these classes are subclassed, and they appear to be treated as leaf implementations, so this change should not cause any issues.

### DIFF
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -1636,7 +1636,7 @@
     <lineContent>setFileExtensions(&quot;properties&quot;);</lineContent>
     <details>
       found   : @UnderInitialization(com.puppycrawl.tools.checkstyle.api.AbstractFileSetCheck.class) @NonNull TranslationCheck
-      required: @Initialized @NonNull TranslationCheck
+      required: @Initialized @NonNull AbstractFileSetCheck
     </details>
   </checkerFrameworkError>
 
--- a/config/spotbugs-exclude.xml
+++ b/config/spotbugs-exclude.xml
@@ -302,16 +302,9 @@
     <Or>
       <Class name="com.puppycrawl.tools.checkstyle.DefaultLogger"/>
       <Class name="com.puppycrawl.tools.checkstyle.PackageObjectFactory"/>
-      <Class name="com.puppycrawl.tools.checkstyle.SarifLogger"/>
-      <Class name="com.puppycrawl.tools.checkstyle.XMLLogger"/>
       <Class name="com.puppycrawl.tools.checkstyle.XmlLoader"/>
-      <Class name="com.puppycrawl.tools.checkstyle.filters.XpathFilterElement"/>
       <Class name="com.puppycrawl.tools.checkstyle.Checker"/>
-      <Class name="com.puppycrawl.tools.checkstyle.checks.TranslationCheck"/>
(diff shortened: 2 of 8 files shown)


vivek-0509:
Hi @romani,

I first explored using a static-factory approach to move validation outside the constructor, but this would require making constructors private and would break the public API. I also considered separating the validation logic into an initialization method, but similar attempts were made previously and those PRs were closed, so this approach is already considered unsuitable.

Given these constraints, the only solution that preserves the public API, keeps fields immutable, and fully prevents the Finalizer Attack is to mark the affected classes as final. These classes (e.g., XpathFileGeneratorAuditListener, SarifLogger, XMLLogger) appear to be concrete leaf implementations, so restricting inheritance seems reasonable and safe.

I’ve applied this approach to the flagged classes. If you believe any of them should remain extensible or if there’s another direction I should explore, please let me know , I’m happy to adjust.

romani:
```
<problem>
<file>file://$PROJECT_DIR$/src/main/java/com/puppycrawl/tools/checkstyle/checks/TranslationCheck.java</file>
<line>193</line>
<module>project</module>
<package>com.puppycrawl.tools.checkstyle.checks</package>
<entry_point TYPE="method" FQNAME="com.puppycrawl.tools.checkstyle.checks.TranslationCheck void setFileExtensions(java.lang.String... extensions)"/>
<problem_class id="RedundantMethodOverride" severity="ERROR" attribute_key="ERRORS_ATTRIBUTES">Method is identical to its super method</problem_class>
<description>Method <code>setFileExtensions()</code> only delegates to its super method #loc</description>
<highlighted_element>setFileExtensions</highlighted_element>
<language>JAVA</language>
<offset>16</offset>
<length>17</length>
</problem>
</problems>
```

vivek-0509:
> ```
> <problems is_local_tool="true">
> <problem>
> <file>file://$PROJECT_DIR$/src/main/java/com/puppycrawl/tools/checkstyle/checks/TranslationCheck.java</file>
> <line>193</line>
> <module>project</module>
> <package>com.puppycrawl.tools.checkstyle.checks</package>
> <entry_point TYPE="method" FQNAME="com.puppycrawl.tools.checkstyle.checks.TranslationCheck void setFileExtensions(java.lang.String... extensions)"/>
> <problem_class id="RedundantMethodOverride" severity="ERROR" attribute_key="ERRORS_ATTRIBUTES">Method is identical to its super method</problem_class>
> <description>Method <code>setFileExtensions()</code> only delegates to its super method #loc</description>
> <highlighted_element>setFileExtensions</highlighted_element>
> <language>JAVA</language>
> <offset>16</offset>
> <length>17</length>
> </problem>
> </problems>
> ```

done

vivek-0509:
@romani ping

Issue #17933, opened 2025-10-14
Resolving spotbugs violations


Atharv3221:
Blocked by https://github.com/spotbugs/spotbugs/issues/1338

As [#1338](https://github.com/spotbugs/spotbugs/issues/1338) was closed expected spotbugs violations where none.
But found:
```
[ERROR] Medium: Redundant nullcheck of stream, which is known to be non-null in
 com.puppycrawl.tools.checkstyle.Main.loadProperties(File) [com.puppycrawl.tools.checkstyle.Main,
 com.puppycrawl.tools.checkstyle.Main] Redundant null check at Main.java:[line 451]
Another occurrence at Main.java:[line 453] RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE

[ERROR] Medium: Redundant nullcheck of inStream, which is known to be non-null in com.puppycrawl.tools.checkstyle.PropertyCacheFile.load() [com.puppycrawl.tools.checkstyle.PropertyCacheFile, com.puppycrawl.tools.checkstyle.PropertyCacheFile] Redundant null check at PropertyCacheFile.java:[line 124]Another occurrence at PropertyCacheFile.java:[line 131] RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE
[ERROR] Medium: Redundant nullcheck of out, which is known to be non-null in com.puppycrawl.tools.checkstyle.PropertyCacheFile.persist() [com.puppycrawl.tools.checkstyle.PropertyCacheFile, com.puppycrawl.tools.checkstyle.PropertyCacheFile] Redundant null check at PropertyCacheFile.java:[line 151]Another occurrence at PropertyCacheFile.java:[line 153] RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE
[ERROR] Medium: Redundant nullcheck of inStream, which is known to be non-null in com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask.createOverridingProperties() [com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask, com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask] Redundant null check at CheckstyleAntTask.java:[line 425]Another occurrence at CheckstyleAntTask.java:[line 427] RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE
[...]

PR #18238, opened 2025-12-03
Issue #17933: Remove obsolete SpotBugs RCN_REDUNDANT_NULLCHECK_OF_NON…

vivek-0509:
Issue #17933:

## Remove obsolete SpotBugs RCN suppressions

This PR removes outdated `RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE` suppressions that were previously added due to false positives on Java 11+. The affected methods no longer contain redundant null checks, so these exclusions are no longer needed.

After removing them, running: `mvn clean verify`
passed successfully with **0 SpotBugs violations**, confirming the cleanup is safe.

Screenshot 2025-12-03 at 4 14 33 PM

### DIFF
--- a/config/spotbugs-exclude.xml
+++ b/config/spotbugs-exclude.xml
@@ -11,48 +11,6 @@
   </Match>
   <Match>
     <!-- till https://github.com/checkstyle/checkstyle/issues/17933 fails on java11+ -->
-    <Class name="com.puppycrawl.tools.checkstyle.Main" />
-    <Method name="loadProperties" />
-    <Bug pattern="RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE" />
-  </Match>
-  <Match>
-    <!-- till https://github.com/checkstyle/checkstyle/issues/17933 fails on java11+ -->
-    <Class name="com.puppycrawl.tools.checkstyle.PropertyCacheFile" />
-    <Method name="load" />
-    <Bug pattern="RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE" />
-  </Match>
-  <Match>
-    <!-- till https://github.com/checkstyle/checkstyle/issues/17933 fails on java11+ -->
-    <Class name="com.puppycrawl.tools.checkstyle.PropertyCacheFile" />
-    <Method name="persist" />
-    <Bug pattern="RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE" />
-  </Match>
-  <Match>
-    <!-- till https://github.com/checkstyle/checkstyle/issues/17933 fails on java11+ -->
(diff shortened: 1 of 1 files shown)


vivek-0509:
Hi @romani ,

I opened this PR because when I removed the SpotBugs suppressions locally and ran the full pipeline `(mvn clean verify, including PMD, SpotBugs, and tests)`, everything passed on my machine.
However, once I pushed the changes, CI failed with 7 SpotBugs violations, all of them RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE

After investigating, I found the issue comes from try-with-resources: the compiler generates synthetic null-check bytecode like:
` if (stream != null) stream.close();`


SpotBugs treats this as a redundant null check → false positive.

The only code change workaround is rewriting these methods to manual try/finally, which removes the synthetic null check.
However, this immediately triggers PMD’s UseTryWithResources rule and goes against Checkstyle’s usual coding style.

So we end up with a conflict:

- PMD → use try-with-resources
- SpotBugs → avoid try-with-resources

Given this, the most practical approach seems to be keeping the minimal SpotBugs suppressions while keeping try-with-resources (as Checkstyle already did for this known false positive).

Should I leave the suppressions as they are, or is there another approach I should look into?
If not, I can close this PR.

image

Atharv3221:
> Hi @romani ,
> 
> I opened this PR because when I removed the SpotBugs suppressions locally and ran the full pipeline `(mvn clean verify, including PMD, SpotBugs, and tests)`, everything passed on my machine. However, once I pushed the changes, CI failed with 7 SpotBugs violations, all of them RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE
> 
> After investigating, I found the issue comes from try-with-resources: the compiler generates synthetic null-check bytecode like: ` if (stream != null) stream.close();`
> 
> SpotBugs treats this as a redundant null check → false positive.
> 
> The only code change workaround is rewriting these methods to manual try/finally, which removes the synthetic null check. However, this immediately triggers PMD’s UseTryWithResources rule and goes against Checkstyle’s usual coding style.
> 
> So we end up with a conflict:
> 
> * PMD → use try-with-resources
> * SpotBugs → avoid try-with-resources
> 
> Given this, the most practical approach seems to be keeping the minimal SpotBugs suppressions while keeping try-with-resources (as Checkstyle already did for this known false positive).
> 
> Should I leave the suppressions as they are, or is there another approach I should look into? If not, I can close this PR.
> 
> image

Leave suppressions as they are, thankyou for investigation. The issue regarding  this is already open in SpotBugs

romani:
@Atharv3221, please share link to issue

vivek-0509:
@romani Here is the link 
https://github.com/spotbugs/spotbugs/issues/1338

romani:
Thanks a lot I marked our issue as blocked

PR #18245, opened 2025-12-05
Issue #15456: Define violation messages for IllegalTypeCheck

aclfe:
Issue #15456

Just removing the IllegalTypeCheck from InlineConfigParser.java

## **Why such a small change:**
I was initially confused as to why it's still passing checks after removing IllegalTypeCheck from InlineConfigParser.java.
I tried running `./mvnw clean test -Dtest=IllegalTypeCheckTest` and other relevant commands (as shown in the picture), it passed all the tests.  

Screenshot From 2025-12-05 06-19-15 (Copy)

After a while, I assumed maybe the tests must be fixed already that's why they're passing. As shown here in the commit: https://github.com/checkstyle/checkstyle/commit/90158f09926eb1f7d89edd55720747e68112f603 (not in this specific commit, but we see that appropriate violation messages are already in place when clicking on a ExampleX.java)

The changes have already been done by someone else.

Thus I'm only removing the line from InlineConfigParser.java as merely a cleanup act. I suspect there might be more such cases, thus I'll keep making PRs and keep checking. Cheers


romani (approved):
Thanks a lot!
It was eat bit still step forward.

Please choose some other Check to update for this issue.

PR #18187, opened 2025-11-26
Issue #17882: Update TYPE_NAME of JavadocCommentsTokenTypes to new AST format

Vishalup29:
Issue #17882

Input file 
```
package temp.temp1;

public class Temp {

    /**
     * @see java.io.IOException
     */
    public void example() { }
}

```

CLI output 

```
JAVADOC_CONTENT -> JAVADOC_CONTENT 
|--TEXT -> package temp.temp1; 
|--NEWLINE -> \r\n 
|--NEWLINE -> \r\n 
|--TEXT -> public class Temp { 
|--NEWLINE -> \r\n 
|--NEWLINE -> \r\n 
|--TEXT ->     /** 
|--NEWLINE -> \r\n 
|--LEADING_ASTERISK ->      * 
|--TEXT ->   
`--JAVADOC_BLOCK_TAG -> JAVADOC_BLOCK_TAG 
    `--SEE_BLOCK_TAG -> SEE_BLOCK_TAG 
        |--AT_SIGN -> @ 
        |--TAG_NAME -> see 
        |--TEXT ->   
        |--REFERENCE -> REFERENCE 
        |   `--IDENTIFIER -> java.io.IOException 
        |--NEWLINE -> \r\n 
        |--LEADING_ASTERISK ->      * 
        `--DESCRIPTION -> DESCRIPTION 
            |--TEXT -> / 
            |--NEWLINE -> \r\n 
            |--TEXT ->     public void example() { } 
            |--NEWLINE -> \r\n 
            `--TEXT -> } 
 ```

PR #18250, opened 2025-12-06
Issue #17882: Remove unused TYPE_NAME token

mahfouz72:
detected at: https://github.com/checkstyle/checkstyle/pull/18187

part of: #17882 

I don’t remember the exact reason this is unused, but it seems we had this token in the early stages of development and later decided to remove or rename it.


src/main/resources/com/puppycrawl/tools/checkstyle/grammar/javadoc/JavadocCommentsLexer.g4
```diff
@@ -34,7 +34,7 @@ tokens {
     INDEX_TERM, RETURN, SNIPPET, SNIPPET_ATTR_NAME, COLON, EXCEPTION,
     THROWS, PARAMETER_NAME, SINCE, VERSION, SEE, STRING_LITERAL,
     LITERAL_HIDDEN, SERIAL, SERIAL_DATA, SERIAL_FIELD, FIELD_TYPE,
-    AT_SIGN, TYPE_NAME, REFERENCE, MEMBER_REFERENCE, PARAMETER_TYPE_LIST,
+    AT_SIGN, UNUSED_TYPE_NAME, REFERENCE, MEMBER_REFERENCE, PARAMETER_TYPE_LIST,
```
mahfouz72: This is to keep the token ordering

romani: UNUSED_TOKEN1

To make it clear it not something real

romani (changes requested):
Item

romani:
Minor is not good, let reference issue on tokens documentation, it will be a way better linkage to context why it unused

src/main/resources/com/puppycrawl/tools/checkstyle/grammar/javadoc/JavadocCommentsLexer.g4
mahfouz72: done

Issue #18251, opened 2025-12-06
Infrastructure: CI failure in PRs due to Eclipse Static Analysis's compiler update (R-4.38)


aclfe:
The Semaphore 2.0 CI pipeline is consistently failing the Eclipse Static Analysis validation step (.ci/validation.sh eclipse-static-analysis) for a multitude of PRs. ( [Example1](https://checkstyle.semaphoreci.com/jobs/0583e10d-9f9c-439e-9921-11334a67c49f) , [Example2](https://checkstyle.semaphoreci.com/workflows/f5c4fe8d-03f2-4c71-85ff-5883afd44ad9?pipeline_id=ae552ce8-1dcb-4ce2-80da-26a186a5d1e1))

The build script compares the supported options of the downloaded eclipse compiler version against `config/org.eclipse.jdt.core.prefs`.

The latest compiler version introduces the key `org.eclipse.jdt.core.compiler.problem.memberOfDeprecatedTypeNotDeprecated`, causing the validation check to fail when this key is missing from the configuration file.

This issue is currently blocking all active Pull Requests that run this validation step.

As mentioned in logs as such:
```
 org.eclipse.jdt.core.compiler.problem.invalidJavadocTagsNotVisibleRef 00:48
 org.eclipse.jdt.core.compiler.problem.invalidJavadocTagsVisibility 00:48
 org.eclipse.jdt.core.compiler.problem.localVariableHiding 00:48
+org.eclipse.jdt.core.compiler.problem.memberOfDeprecatedTypeNotDeprecated 00:48
 org.eclipse.jdt.core.compiler.problem.methodWithConstructorName 00:48
 org.eclipse.jdt.core.compiler.problem.missingDefaultCase 00:48
 org.eclipse.jdt.core.compiler.problem.missingDeprecatedAnnotation 00:48
please update config/org.eclipse.jdt.core.prefs file
```
Solution is just to add it into config/org.eclipse.jdt.core.prefs.

romani:
Please send PR

aclfe:
@romani 
Hi, I've been stumped for a while. 

**The first error `org.eclipse.jdt.core.compiler.problem.memberOfDeprecatedTypeNotDeprecated=ignore` is gone**. But there's a new error at hand. I'm running into a bit of a loop with the config/org.eclipse.jdt.core.prefs file.

[...]

PR #18253, opened 2025-12-06
Pull #18253: infra: Fixing ECJ's version to be pinned, rather than being dynamic to not break CI pipelines

aclfe:
Issue #18251

Just added `org.eclipse.jdt.core.compiler.problem.memberOfDeprecatedTypeNotDeprecated=ignore`onto `config/org.eclipse.jdt.core.prefs`

### DIFF
--- a/.ci/eclipse-compiler-javac.sh
+++ b/.ci/eclipse-compiler-javac.sh
@@ -12,10 +12,12 @@ fi
 JAVA_RELEASE=${2:-$DEFAULT_JAVA_RELEASE}
 
 ECLIPSE_URL="http://ftp-stud.fht-esslingen.de/pub/Mirrors/eclipse/eclipse/downloads/drops4"
-ECJ_MAVEN_VERSION=$(wget --quiet -O- "$ECLIPSE_URL/?C=M;O=D" | grep -o "R-[^/]*" | head -n1)
-echo "Latest eclipse release is $ECJ_MAVEN_VERSION"
 
-ECJ_MAVEN_VERSION=$(wget --quiet -O- "$ECLIPSE_URL/?C=M;O=D" | grep -o "R-[^/]*" | head -n1)
+# hard coded versions until #18251
+ECJ_MAVEN_VERSION="R-4.37-202509050730" 
+
+echo "Using pinned eclipse release: $ECJ_MAVEN_VERSION"
+
 ECJ_JAR=$(wget --quiet -O- "$ECLIPSE_URL/$ECJ_MAVEN_VERSION/" | grep -o "ecj-[^\"]*" | head -n1)
 ECJ_PATH=~/.m2/repository/$ECJ_MAVEN_VERSION/$ECJ_JAR
 


romani:
Thanks a lot for helping.
More problems:
```
In eclipse, Preview of features is supported only at the latest source level
JAVA_RELEASE isset to 24
Executing eclipse compiler on generated resources with all warnings suppressed...
Content of target/eclipse/report.txt:
Preview of features is supported only at the latest source level
[ERROR] Command execution failed.
```

romani:
Thanks a lot for help.

Please apply https://github.com/checkstyle/checkstyle/wiki/How-to-run-certain-phases-and-validations#how-to-create-commit-that-skips-all-ci
 To let you run only semaphore, and not other CIs

aclfe:
Thanks for pointing that out, sorry for not isolating just semaphore CI test.

aclfe:
@romani 

Thank you so much for being patient with me while I was **burning** CI credits. 
I'm happy to say, the error has been resolved.

do I need to re run in order to pass the Internal Server Error (500) error: Pitest / test (api) (pull_request)? can you please re run it so it passes?

-----------------
#### Solution
As of now, I've pinned the version of Eclipse to `R-4.37-202509050730` rather than have it scrape the newest version. I've made this change because I believe we need to be ready before transitioning to their updates to not break our CI pipelines. I hope that makes sense.

romani:
This is not a fix for issue to use latest eclipse static analysis, but it is workaround to stop using latest from web and use what is already in our artifacts cache.

Please put in commit message reference to this PR , "Pull #18253: ....". We will merge this PR as workaround, and we will change issue to be for migration to new eclipse version

.ci/eclipse-compiler-javac.sh
```diff
@@ -11,11 +11,29 @@ if [ -z "$1" ]; then
...
-ECJ_MAVEN_VERSION=$(wget --quiet -O- "$ECLIPSE_URL/?C=M;O=D" | grep -o "R-[^/]*" | head -n1)
-echo "Latest eclipse release is $ECJ_MAVEN_VERSION"
 
-ECJ_MAVEN_VERSION=$(wget --quiet -O- "$ECLIPSE_URL/?C=M;O=D" | grep -o "R-[^/]*" | head -n1)
+# FIX: STOP fetching the "latest" version dynamically.
+# OLD (Broken): 
+# ECJ_MAVEN_VERSION=$(wget ... | head -n1)
+#
+# NEW (Stable):
+# Pinned to the last known working version.
```
romani: Please remove all such comments, in issue we will ask to revert whole commit to reproduce original problem and be ready to use latest eclipse

.ci/eclipse-compiler-javac.sh
```diff
@@ -11,11 +11,29 @@ if [ -z "$1" ]; then
...
+
+ECJ_MAVEN_VERSION="R-4.37-202509050730"
```
romani, line 17: Put comment above, "hard coded versions until #18251"

.ci/eclipse-compiler-javac.sh
```diff
@@ -11,11 +11,29 @@ if [ -z "$1" ]; then
...
+#
+# NEW (Stable):
+# Pinned to the last known working version.
+
+ECJ_MAVEN_VERSION="R-4.37-202509050730" 
+
+echo "Using pinned eclipse release: $ECJ_MAVEN_VERSION"
+
+# (Keep the rest of the script below, but ensure you remove the 
+#  duplicate calculation lines if they exist further down)
```
romani: Remove comments

.ci/validation.sh
```diff
@@ -89,6 +89,7 @@ check-missing-pitests)
   ;;
 
 eclipse-static-analysis)
+  # override pom.xml defaults
```
romani: Revert

config/org.eclipse.jdt.core.prefs
```diff
@@ -68,7 +69,7 @@ org.eclipse.jdt.core.compiler.problem.redundantNullCheck=error
 org.eclipse.jdt.core.compiler.problem.redundantSpecificationOfTypeArguments=error
 org.eclipse.jdt.core.compiler.problem.redundantSuperinterface=error
 org.eclipse.jdt.core.compiler.problem.reportMethodCanBeStatic=error
-org.eclipse.jdt.core.compiler.problem.reportPreviewFeatures=error
+org.eclipse.jdt.core.compiler.problem.reportPreviewFeatures=ignore
```
romani: All changes in this file should be reverted, as we stay on old version of eclipse

aclfe:
Thanks for replying, I'll make all the changes as requested. I must have missed them.

Though I'm curious about the long-term strategy here, **I want to understand properly**. If we revert to a dynamic version in the future, doesn't that leave the CI pipeline vulnerable to breaking again whenever a new update is released? (sort of like this instance?)

Or is the goal to intentionally break early so we can ensure CheckStyle always adheres to the latest static analysis standards, rather than accumulating onto an older version?

romani:
>vulnerable to breaking again whenever a new update is released?

It was was that fragile.
On stable version it becomes hard to remember to update, so we will stay in old for years until someone notices this.

Issue #17951, opened 2025-10-18
uncliear error message if basedir is Linux based but execution is on windows


stephengold:
I have read check documentation.
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

improve exception message in some way to let users identify that `<property name="basedir"` is contributed to failure.

```bash
sgold@LAPTOP-O31T4N20 MINGW64 ~/My Documents/NetBeansProjects/Acorus (master)
$ javac UiVersion.java

sgold@LAPTOP-O31T4N20 MINGW64 ~/My Documents/NetBeansProjects/Acorus (master)
$ cat config.xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE module PUBLIC
    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
    "https://checkstyle.org/dtds/configuration_1_3.dtd">
<!-- Checkstyle configuration for the Acorus project -->

<module name="Checker">
    <property name="basedir" value="/mnt/virginia/home/sgold/NetBeansProjects/Acorus/"/>
    <property name="charset" value="UTF-8"/>
    <property name="fileExtensions" value="java, properties, xml"/>
    <property name="severity" value="error"/>

    <!-- Check that a package-info.java file exists for each package:  -->
    <!-- See https://checkstyle.org/checks/javadoc/javadocpackage.html -->
    <module name="JavadocPackage"/>

    <!-- Miscellaneous checks:                  -->
    <!-- See https://checkstyle.org/checks/misc -->
    <module name="NewlineAtEndOfFile"/>

    <!-- Regexp checks:                           -->
    <!-- See https://checkstyle.org/checks/regexp -->
    <module name="RegexpSingleline">
        <property name="format" value="\s$"/>
        <property name="message" value="Line has trailing whitespace."/>
    </module>

    <!-- Checks for size violations:             -->
    <!-- See https://checkstyle.org/checks/sizes -->
    <module name="FileLength">
        <property name="max" value="1100"/>
    </module>
    <module name="LineLength">
        <property name="max" value="80"/>
    </module>

    <!-- Checks for whitespace:                       -->
    <!-- See https://checkstyle.org/checks/whitespace -->
    <module name="FileTabCharacter">
        <property name="eachLine" value="true"/>
    </module>

    <module name="TreeWalker">
[...]

PR #18254, opened 2025-12-06
Issue #17951: Improve exception message for basedir path mismatch

vivek-0509:
Issue #17951:

### Problem
When the `basedir` property uses a different path type than the files being processed (for example, a Unix path like `/mnt/...` on Windows), users encounter an unclear error:

`IllegalArgumentException: 'other' is different type of Path`

This message does not clearly indicate that the root cause is a mismatch in the `basedir` configuration, making troubleshooting difficult.

### Solution
Wrapped the `IllegalArgumentException` thrown by `Path.relativize()` inside `CommonUtil.relativizePath()` with a clearer, more informative error message that:

- Displays both the file path and base directory values involved
- Guides users to check and align their `basedir` configuration with the current OS/path format

**Improved error message example:**

```
Failed to relativize path 'C:...' against base directory '/mnt/...'.
"(configured via 'basedir' property). "
Please ensure the base directory path matches the current OS/path type.
```

### Changes
- **CommonUtil.java:** Added try-catch around `Path.relativize()` to throw the new informative exception
- **CommonUtilTest.java:** Added a unit test with Mockito to verify that the improved exception message is thrown correctly

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
@@ -376,7 +376,7 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
      * @return {@code true} if the file is accepted.
      */
     private boolean acceptFileStarted(String fileName) {
-        final String stripped = CommonUtil.relativizePath(basedir, fileName);
+        final String stripped = relativizePathWithCatch(fileName);
         return beforeExecutionFileFilters.accept(stripped);
     }
 
@@ -388,7 +388,7 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
      */
     @Override
     public void fireFileStarted(String fileName) {
-        final String stripped = CommonUtil.relativizePath(basedir, fileName);
+        final String stripped = relativizePathWithCatch(fileName);
         final AuditEvent event = new AuditEvent(this, stripped);
         for (final AuditListener listener : listeners) {
             listener.fileStarted(event);
@@ -403,7 +403,7 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
      */
     @Override
     public void fireErrors(String fileName, SortedSet<Violation> errors) {
-        final String stripped = CommonUtil.relativizePath(basedir, fileName);
+        final String stripped = relativizePathWithCatch(fileName);
         boolean hasNonFilteredViolations = false;
         for (final Violation element : errors) {
             final AuditEvent event = new AuditEvent(this, stripped, element);
@@ -427,7 +427,7 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
      */
     @Override
     public void fireFileFinished(String fileName) {
-        final String stripped = CommonUtil.relativizePath(basedir, fileName);
+        final String stripped = relativizePathWithCatch(fileName);
         final AuditEvent event = new AuditEvent(this, stripped);
         for (final AuditListener listener : listeners) {
             listener.fileFinished(event);
@@ -662,6 +662,25 @@ public class Checker extends AbstractAutomaticBean implements MessageDispatcher,
                     messageKey, args);
 
         return localizedMessage.getMessage();
+    }
+
+    /**
+     * Relativizes a path and wraps any exception with a user-friendly localized message.
+     *
+     * @param fileName the file path to relativize
+     * @return the relativized path
+     * @throws IllegalStateException if any exception occurs during relativization
+     */
+    private String relativizePathWithCatch(String fileName) {
+        try {
+            return CommonUtil.relativizePath(basedir, fileName);
+        }
+        // -@cs[IllegalCatch] Catching generic Exception to include fileName and basedir context
+        catch (Exception exception) {
+            throw new IllegalStateException(
+                getLocalizedMessage("general.relativizePath",
+                        fileName, basedir), exception);
+        }
     }
 
 }
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/messages.properties
@@ -16,6 +16,7 @@ DefaultLogger.errorStreamException=Parameter errorStreamOptions can not be null
 DefaultLogger.infoStreamException=Parameter infoStreamOptions can not be null
 general.exception=Got an exception - {0}
 general.fileNotFound=File not found!
+general.relativizePath=Failed to relativize path ''{0}'' against base directory ''{1}'' (configured via ''basedir'' property). Please ensure the base directory path matches the current OS/path type.
 Main.createListener=Invalid output format. Found ''{0}'' but expected ''{1}'' or ''{2}''.
 Main.errorCounter=Checkstyle ends with {0} errors.
 Main.loadProperties=Unable to load properties from file ''{0}''.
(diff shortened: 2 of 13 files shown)


vivek-0509:
Hi @romani,

The failure I’m seeing is from the Eclipse/JDT compiler options check, not from my relativizePath changes themselves. The log shows:

```
+org.eclipse.jdt.core.compiler.problem.memberOfDeprecatedTypeNotDeprecated
please update config/org.eclipse.jdt.core.prefs file
```

It appears ECJ introduced a new compiler flag, and the script now expects config/org.eclipse.jdt.core.prefs to be updated accordingly. Since this is a global configuration change, I didn’t update the prefs file in this PR to avoid mixing infrastructure adjustments with the bug fix.

Additionally, someone has already opened a PR to address this exact prefs update, so I’m not making any changes for it here.

src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java
```diff
@@ -264,7 +266,17 @@ public static String relativizePath(final String baseDirectory, final String pat
...
             final Path pathAbsolute = Path.of(path);
             final Path pathBase = Path.of(baseDirectory);
-            resultPath = pathBase.relativize(pathAbsolute).toString();
+            try {
+                resultPath = pathBase.relativize(pathAbsolute).toString();
+            }
+            catch (IllegalArgumentException exception) {
+                throw new IllegalStateException(
+                    String.format(Locale.ROOT,
+                        "Failed to relativize path '%s' against base directory '%s' "
```
romani: Good, but please put message to messages.properties
We localize all messages. There are bunch of examples in code on this. But migration is not complete so there still bad examples of English hard coded messages.

src/test/java/com/puppycrawl/tools/checkstyle/utils/CommonUtilTest.java
```diff
@@ -602,6 +602,39 @@ public void testLoadSuppressionsUriSyntaxException() throws Exception {
...
     }
 
+    @Test
+    public void testRelativizePathExceptionMessage() {
+        final String baseDirectory = "/absolute/base";
+        final String relativePath = "relative/file.java";
+
+        final IllegalStateException exception = getExpectedThrowable(
+            IllegalStateException.class,
+            () -> CommonUtil.relativizePath(baseDirectory, relativePath)
```
romani: We don't allow pure energy unit testing of code.
Please try to execute it by Ant task or Main class.

romani (changes requested):
Nice.

Item to improve

src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java
vivek-0509: Okk will improve this

src/test/java/com/puppycrawl/tools/checkstyle/utils/CommonUtilTest.java
vivek-0509: Thanks for the clarification, I will update this

vivek-0509:
Hii @romani 
Since CommonUtil is in the utils package, which is strictly limited by ImportControl. As a result, it cannot use the localization infrastructure:

-  LocalizedMessage not allowed  
-  Definitions not allowed  
-  MessageFormat not allowed

When attempting to use ResourceBundle directly, Checker Framework flagged violations:

- `@UnknownPropertyKey` (key not statically validated)  
- i18nformat.key.not.found (translation key lookup check)

This would require adding suppressions:

- specifier="i18nformat.key.not.found"  
- specifier="argument" (UnknownPropertyKey)

To avoid import restrictions and suppressions, i moved wrapping  out of CommonUtil:

- CommonUtil.relativizePath() now simply throws IllegalArgumentException  
- Checker performs the wrapping

src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java
vivek-0509: Done

src/test/java/com/puppycrawl/tools/checkstyle/utils/CommonUtilTest.java
vivek-0509: Done

vivek-0509:
I've added the new general.relativizePath message key to all translation files. I want to be transparent about the approach I took:

Issue encountered:

When I initially added translations using native characters (e.g., Russian Cyrillic, Japanese Kanji), I encountered encoding errors:

`Unsupported characters for the charset 'ISO-8859-1'`
Java  .properties
 files use ISO-8859-1 encoding by default, which doesn't support non-ASCII characters

Solution applied:
I converted all non-ASCII translations to Unicode escape sequences (\uXXXX)

romani:
>Unsupported characters for the charset 'ISO-8859-1'

Weird, what tool gives you such problem ?
We have all translated already and there no problems.
Please create new PR with native characters to let me see a problem

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -664,4 +664,23 @@ Definitions.CHECKSTYLE_BUNDLE, getClass(),
...
 
+    /**
+     * Safely relativizes a path, wrapping any IllegalArgumentException
+     * with a user-friendly localized message.
+     *
+     * @param fileName the file path to relativize
+     * @return the relativized path
+     * @throws IllegalStateException if relativization fails
+     */
+    private String safeRelativizePath(String fileName) {
```
romani: relativizePathWithCatch

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
```diff
@@ -664,4 +664,23 @@ Definitions.CHECKSTYLE_BUNDLE, getClass(),
...
+        try {
+            return CommonUtil.relativizePath(basedir, fileName);
+        }
+        catch (IllegalArgumentException exception) {
```
romani: Catch "Exception" here to be sure we wrap any kind of exception with details

src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java
```diff
@@ -255,6 +255,7 @@ public static String baseClassName(String type) {
      *            the path to relativize against base directory
      * @return the relative normalized path between base directory and
      *     path or path if base directory is null.
+     * @throws IllegalArgumentException if the path cannot be relativized against base directory
```
romani: We can remove it.
Runtime exceptions are not documented usually

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
vivek-0509: done

src/main/java/com/puppycrawl/tools/checkstyle/Checker.java
vivek-0509: done

src/main/java/com/puppycrawl/tools/checkstyle/utils/CommonUtil.java
vivek-0509: done

vivek-0509:
> > Unsupported characters for the charset 'ISO-8859-1'
> 
> Weird, what tool gives you such problem ? We have all translated already and there no problems. Please create new PR with native characters to let me see a problem

Apologies for the earlier confusion ,the issue turned out to be on my side, specifically with my IDE settings. IntelliJ was configured to use ISO-8859-1 for .properties files with Transparent native-to-ascii conversion enabled . It is fixed now.

vivek-0509:
Thankyou for reviewing and merging..

PR #18267, opened 2025-12-10
infra: use depth of 1 when cloning git repositories

stoyanK7:
## Summary - What changed?
Use `--depth 1` in the git clone command within `checkout_from` to create a shallow clone.

### Documentation on the `--depth` option
https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthdepth
> Create a shallow clone with a history truncated to the specified number of commits.

## Motivation - Why change?
This `checkout_from` function is used in 48 places to clone repositories. When cloning, we care about the code's latest state from the cloned project. We do not care about the project’s git history. By default, `git clone` also pulls the full git history. Setting the depth to `1` ensures that we only pull the minimum git history - the latest commit.

This will speed up all CI jobs that use it.

## Benchmarks with `trino` and `spring-integration` repositories
The _Time_ for each benchmark is an average of 3 runs:

| Command                                                                         | Time (real `time`) | Folder Size (`du`) |
|---------------------------------------------------------------------------------|--------------------|------------------|
| `git clone https://github.com/trinodb/trino.git`                                | 44.47s             | 510.73 MB        |
| `git clone --depth 1 https://github.com/trinodb/trino.git`                      | 9.92s              | 247.36 MB        |
|                                                                                 |                    |                  |
| `git clone https://github.com/spring-projects/spring-integration.git`<br>       | 12.21s             | 128.81 MB        |
| `git clone --depth 1 https://github.com/spring-projects/spring-integration.git` | 02.97s             | 45.74 MB         |

### DIFF
--- a/.ci/util.sh
+++ b/.ci/util.sh
@@ -53,7 +53,7 @@ function checkout_from {
     git clean -f -d
     cd ../
   else
-    for i in 1 2 3 4 5; do git clone "$CLONE_URL" && break || sleep 15s; done
+    for i in 1 2 3 4 5; do git clone --depth 1 "$CLONE_URL" && break || sleep 15s; done
   fi
   cd ../
 }
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -755,6 +755,7 @@ no-error-orekit)
   cd .ci-temp/hipparchus
   # checkout to version that Orekit expects
   SHA_HIPPARCHUS="1492f06848f57e46bef911a""ad16203a242080028"
+  git fetch --depth 1 origin "$SHA_HIPPARCHUS"
   git checkout $SHA_HIPPARCHUS
   mvn -e --no-transfer-progress install -DskipTests
   cd -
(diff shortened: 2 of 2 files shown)


romani:
`.ci/validation.sh no-error-orekit` 

>fatal: unable to read tree (1492f06848f57e46bef911aad16203a242080028)

looks like we do checkout to past commit to keep it stable. We can have two functions to split a world on who needs shallow and who does not.

stoyanK7:
> looks like we do checkout to past commit to keep it stable. We can have two functions to split a world on who needs shallow and who does not.

We still don't need a full clone for `no-error-orekit`. I added an extra `git fetch` for those.

stoyanK7:
@romani I also believe we can _greatly_ benefit if we add `--depth 1` in `diff.groovy` in `contribution/checkstyle-tester`. We can cut _minutes_ (if not 10+ minutes) off the jobs that use this tool. For example, cloning `openjdk25` repo takes 3 minutes flat. With `--depth 1`, it takes ~40 seconds.

Do you want to do that? Please let me know what you think.

romani:
We can try to update, I remember we already tried to do this, ther might be some PR about it.
Proceed with caution, and if any complications happens we should think how to make migration to improve when easy quickly and deal with more complicated step by step

PR #18263, opened 2025-12-09
Issue #16003: Add Kafka to no-error testing

stoyanK7:
Resolves issue
- #16003 

Based on the work of:
- @romani #16032
- @dejan2609 #16395

---

Added GitHub Actions Workflow for regression testing with Apache Kafka: `no-error-kafka`. The job is identical to the other `no-error-*` jobs. For instance:
https://github.com/checkstyle/checkstyle/blob/57df2a717bf7850b3afb43c2438dde542fad05e5/.ci/validation.sh#L236-L247

Usually, such jobs use Semaphore and CircleCI, but this one must use GitHub Actions due to high memory usage.

PR #18286, opened 2025-12-12
infra: move no-error-trino to no-error-workflow.yml

stoyanK7:
Follow-up from https://github.com/checkstyle/checkstyle/pull/18263#discussion_r2610590308
> Side note: We must also move `no-error-trino` under this `no-error testing` workflow. I will do that in a followup PR (after this gets merged).

image


stoyanK7:
Actionlint failure is unrelated.

PR #18288, opened 2025-12-12
Issue #14084: Remove IndexChecker from Checker Framework configuration

vivek-0509:
Issue #14084:

Remove IndexChecker from the Checker Framework configuration as discussed with maintainer.

IndexChecker was providing minimal value to the project:
- The project does not actively use any IndexChecker annotations (`@IntRange`, `@NonNegative`, `@Positive`, etc.)
- All IndexChecker violations were already suppressed in the suppression file
- This added maintenance overhead and build time cost without providing real benefit

 IndexChecker can be reintroduced in the future when the codebase is ready to adopt its annotations.

## Changes

1. **pom.xml** - Removed the `checker-index` Maven profile
2. **config/checker-framework-suppressions/checker-index-suppressions.xml** - Deleted the suppression file 
3. **.github/workflows/checker-framework.yml** - Removed `index` from the CI matrix

## Related Discussion
https://github.com/checkstyle/checkstyle/issues/14084#issuecomment-3641990948


vivek-0509:
Hi @romani,

The `actionlint` CI check is failing, but this is not caused by my changes. The failure is due to a pre-existing issue in `.github/workflows/diff-report.yml` at line 59: `if: 'true'`

This `if: 'true'` condition appears to be an intentional workaround for the `shanegenschaw/pull-request-comment-trigger` action (the comment says "We print it here for this action to work").

Is this a known and accepted issue?
other CI failure is unrelated

romani (approved):
thanks a lot.


yes, this step back, it will help us to economize efforts to do leap forward later on

Issue #18070, opened 2025-11-09
resolve PMD warning during execution


romani:
resolve PMD warning in logs:
https://ci.appveyor.com/project/Checkstyle/checkstyle/builds/53003228/job/rcb6sqkpyovuw7pb#L1782

```
[INFO] >>> pmd:3.28.0:check (default) > :pmd @ checkstyle >>>
[INFO] 
[INFO] --- pmd:3.28.0:pmd (pmd) @ checkstyle ---
[INFO] PMD version: 7.18.0
[WARNING] Warning at C:\projects\checkstyle\target\pmd\rulesets\001-pmd-main.xml:53:3
 51|     </properties>
 52|   </rule>
 53|   <rule ref="category/java/design.xml/AvoidCatchingGenericException">
       ^^^^^ Use Rule name category/java/errorprone.xml/AvoidCatchingGenericException instead of the deprecated Rule name category/java/design.xml/AvoidCatchingGenericException. PMD 8.0.0 will remove support for this deprecated Rule name usage.
 54|     <properties>
 55|       <!-- we allow catching Exception in certain cases to provide more details in wrapper -->
[WARNING] Warning at C:\projects\checkstyle\target\pmd\rulesets\001-pmd-main.xml:53:3
 51|     </properties>
 52|   </rule>
 53|   <rule ref="category/java/design.xml/AvoidCatchingGenericException">
       ^^^^^ Use Rule name category/java/errorprone.xml/AvoidCatchingGenericException instead of the deprecated Rule name category/java/design.xml/AvoidCatchingGenericException. PMD 8.0.0 will remove support for this deprecated Rule name usage.
 54|     <properties>
 55|       <!-- we allow catching Exception in certain cases to provide more details in wrapper -->
[WARNING] Exception applying rule UselessOverridingMethod on file C:\projects\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\checks\annotation\MissingDeprecatedCheck.java, continuing with next rule
net.sourceforge.pmd.lang.rule.xpath.PmdXPathException: net.sf.saxon.trans.XPathException: XPath rule expression returned a non-node (class net.sf.saxon.value.StringValue): 
Exception Context:
	[1:Kind=Unknown error]
	[2:Phase=EVALUATION]
[...]

PR #18302, opened 2025-12-13
Issue #18070: resolve PMD warning during execution.

smita1078:
#18070 

**What changed**

Fixed an invalid XPath expression in config/pmd/001-pmd-main.xml by replacing boolean OR  `||` with XPath union  `|` in violationSuppressXPath.

**Why this change is needed**

PMD requires violationSuppressXPath expressions to return a node set.
The previous expression used boolean OR `||` , which evaluates to a boolean/string value instead of nodes.


romani (approved):
Thanks a lot!!!!

Issue #17177, opened 2025-06-07
Examples Validation: content is same in all for single Check


romani:
Main value that we put in all examples is that all java code looks same
and only comment are changing position.
See images at https://github.com/checkstyle/checkstyle/issues/13345

It would be awesome to have junit validation for it to ensure that all examples are same.

Quick idea:
We can do parse all exmples for Check by checkstyle AST printer (without comment AST Nodes) and make sure AST is same 

any other ideas are welcome.

smita1078:
@romani 

I see two reasonable approaches:
**Option 1 – Structural AST comparison**
Parse each example using Checkstyle’s Java parser
Compare AST structure only (ignoring comments and source positions)


**Option 2 – AST dump comparison** 
Generate an AST dump (similar to DebugAstPrinter)
Strip line/column numbers
Compare dumps as text

Both approaches ensure that only comments differ and prevent accidental logic changes in examples.

romani:
I would do option1, and in case violation detection utilize "Option 2" to let user spot easily a diff.

We should expect that there will be a lot of violations, so we should have  list of suppressions, to let us enforce validation that what is valid now, and let migrate all others by crowd sourcing ("good xxxx issues").

smita1078:
Understood 👍

romani:
All gratitude for fixing this issue is to @smita1078  !!!

smita1078:
> All gratitude for fixing this issue is to [@smita1078](https://github.com/smita1078) !!!

Thankyou! 
Remaining comments will be handled in a follow-up PR, and a new Issue (good xxxx issue) will be created for other contributors

PR #18366, opened 2025-12-17
Issue #17177: Examples Validation content is same in all for single C…

smita1078:
#17177 

Add AST Consistency Test for Xdocs Examples

**Description**

This PR adds a new test `XdocsExamplesAstConsistencyTest` that validates all xdocs Java examples for the same check have identical AST structure, differing only in comments. This ensures consistency across documentation examples as requested in issue #13345.

**Changes**

Added:
- `XdocsExamplesAstConsistencyTest.java`: New JUnit test class that:
  - Scans all directories under `src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle`
  - Compares AST structure of all `Example*.java` files within each directory
  - Ignores comment nodes (SINGLE_LINE_COMMENT, BLOCK_COMMENT_BEGIN, COMMENT_CONTENT) during comparison
  - Reports violations when examples have different code structures
  - Includes suppression list for existing violations (81 directories)

Current behavior:
- Test passes with suppressions in place
- Skips 81 suppressed directories
- Validates all non-suppressed directories

Example violation output (when suppression is removed):
```
Directory: checks/annotation/annotationlocation
Reference: Example1.java
Mismatch:  Example2.java
```

image

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
@@ -77,9 +77,7 @@ public class XdocsExamplesAstConsistencyTest {
     private static final Set<String> SUPPRESSED_EXAMPLES = Set.of(
             "checks/annotation/annotationonsameline/Example2",
             "checks/annotation/missingoverride/Example2",
-            "checks/annotation/suppresswarningsholder/Example2",
-            "checks/annotation/suppresswarningsholder/Example3",
-            "checks/annotation/suppresswarningsholder/Example4",
+            "checks/annotation/suppresswarningsholder/Example1",
             "checks/blocks/emptyblock/Example3",
             "checks/blocks/emptycatchblock/Example4",
             "checks/blocks/emptycatchblock/Example5",
--- a/src/site/xdoc/checks/annotation/suppresswarningsholder.xml
+++ b/src/site/xdoc/checks/annotation/suppresswarningsholder.xml
@@ -105,7 +105,7 @@ class Example1 {
 </code></pre></div>
         <p id="Example2-code">Example:</p>
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
-class Example2 {
+public class Example2 {
   // violation below, 'More than 7 parameters (found 8)'
   public void needsLotsOfParameters (int a,
     int b, int c, int d, int e, int f, int g, int h) {
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/suppresswarningsholder/Example2.java
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/suppresswarningsholder/Example2.java
@@ -14,7 +14,7 @@
 package com.puppycrawl.tools.checkstyle.checks.annotation.suppresswarningsholder;
 
 // xdoc section -- start
-class Example2 {
+public class Example2 {
   // violation below, 'More than 7 parameters (found 8)'
   public void needsLotsOfParameters (int a,
     int b, int c, int d, int e, int f, int g, int h) {


src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -0,0 +1,515 @@
...
+
+    private static final Path XDOCS_ROOT = Path.of(
+            "src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle"
+    );
+
+    private static final String COMMON_PATH =
+            "src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/";
+
+    // This list is temporarily suppressed.
+    // Until: https://github.com/checkstyle/checkstyle/issues/XXXXX
```
romani: please use XXXX to avoid more extenension on whitelist

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -0,0 +1,515 @@
...
+            message = "";
+        }
+        else {
+            final StringBuilder builder = new StringBuilder(1024);
+
+            builder.append("Found ")
+                    .append(violations.size())
+                    .append(
+                            """
+                             director(ies) with AST mismatches.
```
romani: please try to avoid "ies", let now grow whitelist file.

smita1078:
> can we have one folder to be fixed in this PR to see that it is actually working ?
> 
> it can be in separate commit, I will reference it as good example on how to updates to resolve such suppressions.

Sure

smita1078:
Done

smita1078:
@romani , could you please review whenever you get a chance..

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -0,0 +1,516 @@
...
+    private static final Path XDOCS_ROOT = Path.of(
+            "src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle"
+    );
+
+    private static final String COMMON_PATH =
+            "src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/";
+
+    // This list is temporarily suppressed.
+    // Until: https://github.com/checkstyle/checkstyle/issues/XXXX
+    private static final Set<String> SUPPRESSED_DIRECTORIES = Set.of(
```
romani: Please do us a favor to allow suppression by part of path.
There are Examples that will forever stay different from others.
Reason is that there are examples to show how how properties works, and there are examples to show some in demand use case that users always wants but it is not very obvious how to do.

So examples are also collection of ready to use configs.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/annotationlocation/Example1.java
```diff
@@ -13,13 +13,14 @@
...
 
-interface DataLoader {
-
-}
-
 
 // xdoc section -- start
 class Example1 {
+
+  interface DataLoader {
```
romani: We trying to keep all unrelated for Check code outside of `xdoc section -- start`
What forcing you to move it inside Example classes ?
Being outside, should be ok by AST similarity

smita1078: In this Example:

```java
interface DataLoader { }
```

- Example1.java has this interface defined outside the xdoc section
- Example2.java does NOT have this interface (to avoid duplicates)
- This causes AST mismatch - Example1 has INTERFACE_DEF node, Example2 doesn't as currently whole file is being considered for AST similarity check.

I Added interface to all examples (outside xdoc sections)
```java
interface DataLoader { }  // Define in each file

// xdoc section -- start
class Example1 { ... }
// xdoc section -- end
```
Problem: Caused duplicate code violations when all examples are combined.
When added to each file outside the xdoc section.

I can Modify `XdocsExamplesAstConsistencyTest` to:
1. Extract only code between `// xdoc section -- start` and `// xdoc section -- end`
2. Compare AST only for that extracted content
3. Ignore any helper code outside these markers

This way helper code can differ between examples while ensuring the actual documented examples are structurally consistent. What do u suggest?

romani: >Extract only code between // xdoc section -- start and // xdoc section -- end

This is great idea

smita1078: Okay , will implement this... 👍

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -0,0 +1,778 @@
...
+        else {
+            final StringBuilder builder = new StringBuilder(1024);
+
+            builder.append("Found ")
+                    .append(violations.size())
+                    .append(
+                            """
+                             example files with AST mismatches.
+
+                            """
```
romani, line 345: please share why you use `"""` here and in following append.
do you think that `.append("\n\n");` is confusing ?

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -0,0 +1,778 @@
...
+    private static boolean containsMultipleExamples(Path dir) {
+        boolean result = false;
+
+        try (Stream<Path> pathStream = Files.list(dir)) {
+            result = pathStream
+                    .filter(path -> path.getFileName().toString().matches("Example\\d+\\.java"))
+                    .count() > 1;
+        }
+        catch (IOException ignored) {
+            // result already set to false
```
romani, line 494: do not swallow, wrap into IllegalStateException and throw up to let engineer see all details of exception

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -0,0 +1,778 @@
...
+            final List<Path> examples = getExampleFiles(dir);
+
+            if (!examples.isEmpty()) {
+                violations.addAll(compareExamples(dir, examples));
+            }
+        }
+        catch (IOException exception) {
+            final String relativePath = getRelativePath(dir);
+            violations.add("Directory: " + relativePath + "\n"
+                    + "Error: " + exception.getMessage());
```
romani, line 519: I would throw exception to user, if there is disk access problems, engineers should see it.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -0,0 +1,778 @@
...
+                            dir, example, reference, referenceAst
+                    );
+                    if (violation != null) {
+                        violations.add(violation);
+                    }
+                }
+            }
+        }
+        catch (CheckstyleException exception) {
+            // Skip examples that can't be parsed as Java
```
romani, line 603: let have a hardcoded list of example that cannot be parsed.
We do a lot of efforts to make all of them compiled, so it should be like exception rather then rule.
If there are such files and they are not parseable for good reason, lets keep them in some hardcoded exclude list.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -0,0 +1,778 @@
...
+    private static final String XDOC_START_MARKER = "// xdoc section -- start";
+    private static final String XDOC_END_MARKER = "// xdoc section -- end";
+
+    /**
+     * Examples that have independent code structure and should not be compared.
+     * These represent different use cases or configurations with different code.
+     *
+     * <p>Format: "directory/ExampleN" where the example has unique code.
+     *
+     * <p>Until: <a href="https://github.com/checkstyle/checkstyle/issues/XXXX">...</a>
```
romani, line 75: please create issue to explain what to do new contributors.

smita1078:
Thanks a lot!
Will address all the comments in a follow up PR.

> I actually means approval of PR .
> 
> to integrate this work sooner.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
smita1078: Done

smita1078: #18435

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
smita1078: I though it would be provide better readability for multi-line string literals, though I dont think `.append("\n\n");` is confusing hence have changed that..

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
smita1078: Done

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
smita1078: Yes, agreed was doing this just to get all list of example failures in one go.. Modified it accordingly.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
smita1078: Yes, Done

Issue #18034, opened 2025-11-01
Resolve Pitest Suppressions - imports


romani:
As a part of this issue, we have to kill the mutation listed below and remove the suppression from [pitest xxxx suppression.xml](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/)

To understanding how to kill the mutation 
Visit [Wiki page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports), [Pitest-docs](https://pitest.org/) and expected actions and workflow at https://github.com/checkstyle/checkstyle/issues/12341.

Target is to remove all `<mutation` from [Pitest-imports](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-imports-suppressions.xml)

yashgoyal0110:
@romani 
I want to work on this, assign please!

romani:
No assignments, just choose mutation, and make a comment "I am on ....`

romani:
Please [read](https://github.com/checkstyle/checkstyle/blob/master/.github/CONTRIBUTING.md#getting-started) and watch videos at [Starting_Development](https://checkstyle.org/beginning_development.html#Starting_Development).
Please start with easier issues first.

Gaoyan1999:
I am on it. I've already reproduced the result locally.

romani:
>I am on it.

Please define exact mutation. All other mutation can be handled by other people in paray to your activity.

Gaoyan1999:
> > I am on it.
> 
> Please define exact mutation. All other mutation can be handled by other people in paray to your activity.

@romani I will focus on NonVoidMethodCallMutator type first, the first five out of the eight in the `pitest-imports-suppressions.xml`.

PR #18274, opened 2025-12-10
Issue #17988: fix all warnings reported by error prone

Gaoyan1999:
Issue: #17988

Resolve All warning message when ran
```
./mvnw -e --no-transfer-progress clean test-compile -Perror-prone-test-compile
```

PR #18370, opened 2025-12-18
Issue #18034: Resolve PIT Mutation Suppression for NonVoidMethodCallM…

Gaoyan1999:
Remove the three `NonVoidMethodCallMutator` suppressions from `pitest-imports-suppression.xml`, but keep the following one:
```xml
<mutation unstable="false">
  <sourceFile>ImportOrderCheck.java</sourceFile>
  <mutatedClass>com.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck</mutatedClass>
  <mutatedMethod>visitToken</mutatedMethod>
  <description>removed call to java/lang/String::valueOf</description>
  <lineContent>throw new IllegalStateException(</lineContent>
</mutation>
```

This suppression is retained because it represents an equivalent mutation. Removing String.valueOf(option) does not change the resulting exception message: StringBuilder.append(null) still produces "null". Since there is no observable behavior difference, keeping this suppression is appropriate.

**Reference**: Issue #18034

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportControlLoaderTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportControlLoaderTest.java
@@ -60,15 +60,18 @@ public class ImportControlLoaderTest {
 
     @Test
     public void testWrongFormatUri() throws Exception {
+        final URI uri = new URI("aaa://" + getPath("InputImportControlLoaderComplete.xml"));
         try {
-            ImportControlLoader.load(new URI("aaa://"
-                    + getPath("InputImportControlLoaderComplete.xml")));
+            ImportControlLoader.load(uri);
             assertWithMessage("exception expected").fail();
         }
         catch (CheckstyleException exc) {
             assertWithMessage("Invalid exception class")
                 .that(exc.getCause())
                 .isInstanceOf(MalformedURLException.class);
+            assertWithMessage("Invalid exception message")
+                .that(exc.getMessage())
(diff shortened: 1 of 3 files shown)


Gaoyan1999:
I checked the logs and this failure appears to be caused by a GitHub CI infrastructure/network issue (checkout failed to connect to github.com).
I also ran mvn clean verify locally, and it passed successfully.

romani:
https://github.com/checkstyle/checkstyle/pull/18274#issuecomment-3667672580

Please address

src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportControlLoaderTest.java
```diff
@@ -160,4 +166,20 @@ public void testInputStreamFailsOnRead() throws Exception {
         }
     }
 
+    @Test
+    public void testLoadUriUnableToFindFile(@TempDir Path tempDir) {
+        final Path missingFile = tempDir.resolve("missingFile.xml");
+        final URI uri = missingFile.toUri();
+
+        final CheckstyleException exception = getExpectedThrowable(CheckstyleException.class,
+                () -> ImportControlLoader.load(uri));
```
romani: Can we test it by Check test , ImportControlChevk ?
Do do not allow pure unit testing, all should be through natural execution of checks.

Gaoyan1999: I've removed it into Check test

Issue #18303, opened 2025-12-14
XdocsExampleFileTest should catch fact that ExampleX.java is not referenced in respected Test class


romani:
detected at https://github.com/checkstyle/checkstyle/pull/18268/changes#r2617337861

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExampleFileTest.java

should be updated to catch that some ExampleX.java 
is not added to 
https://github.com/checkstyle/checkstyle/blob/master/src/xdocs-examples/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceAroundExamplesTest.java

we can just make sure that  were is `("Example11.java")` in  WhitespaceAroundExamplesTest.java

AradhyaSharma31:
Hi, I would like to work on this issue.

romani:
Thanks a lot for helping, just send a PR

Aman-Baliyan:
Hi! I’m planning to work on ensuring XdocsExampleFileTest catches cases where an Example*.java file exists under xdocs but isn’t referenced in WhitespaceAroundExamplesTest.
Before starting, could you please confirm whether this should be limited to WhitespaceAroundExamplesTest.java or generalized further?
I just want to confirm I’m thinking in the right direction.

romani:
For all checks

Aman-Baliyan:
Hi @romani sir,
I’ve created the test, and while running it I noticed that there are several example files present under
xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks,
but their corresponding example classes are missing under
xdocs-examples/java/com/puppycrawl/tools/checkstyle/checks.

As a result, the test reports the following missing example tests:

Missing test for Example7.java in IllegalTypeCheckExamplesTest.java

Missing test for Example8.java in IllegalTypeCheckExamplesTest.java

Missing test for Example3.java in UnnecessaryParenthesesCheckExamplesTest.java

Missing test for Example3.java in OneTopLevelClassCheckExamplesTest.java

Missing test for Example8.java in JavadocMethodCheckExamplesTest.java

Missing test for Example4.java in JavadocTagContinuationIndentationCheckExamplesTest.java

Missing test for Example6.java in LineLengthCheckExamplesTest.java

Could you please advise on the expected approach here?
Should we first add the missing example Java files under xdocs-examples/java and then add the corresponding test method, or is there another intended solution for these cases?

romani:
[...]

PR #18375, opened 2025-12-18
Issue #18303: Update XdocExampleFileTest add test

Aman-Baliyan:
Issue : #18303 

Added the test in XdocsExampleFileTest.java and also increase the example length in Example6 as suggested in previous pr.

```
Files Changed - 
1. src/site/xdoc/checks/sizes/linelength.xml
2. src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExampleFileTest.java
3. src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/linelength/Example6.java

```

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExampleFileTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExampleFileTest.java
@@ -22,13 +22,19 @@ package com.puppycrawl.tools.checkstyle.internal;
 import static com.google.common.truth.Truth.assertWithMessage;
 
 import java.beans.PropertyDescriptor;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.commons.beanutils.PropertyUtils;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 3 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExampleFileTest.java
```diff
@@ -52,34 +58,34 @@ public class XdocsExampleFileTest {
     // This list is temporarily suppressed.
     // Until: https://github.com/checkstyle/checkstyle/issues/17449
     private static final Map<String, Set<String>> SUPPRESSED_PROPERTIES_BY_CHECK = Map.ofEntries(
-            Map.entry("MissingJavadocTypeCheck", Set.of("skipAnnotations")),
```
romani, line 55: Please revert this unrelated update, formatting.
We do not allow mixing target of update. 
If something unrelated to target ticket should be done, is should be done in separate commit with definition of reason of update.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/sizes/linelength/Example6.java
```diff
@@ -14,13 +15,13 @@
 package com.puppycrawl.tools.checkstyle.checks.sizes.linelength;
 
 public class Example6 {
-  void testMethod() {
+  void testMethod() {// ok, the violation is suppressed
```
romani, line 18: Please use this wording of trailing comment 
https://checkstyle.org/filters/suppresswithnearbycommentfilter.html#Example2-code

romani (approved):
Dream is nearby.

Items to improve:

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExampleFileTest.java
```diff
@@ -118,4 +124,63 @@ public void testAllCheckPropertiesAreUsedInXdocsExamples() throws Exception {
...
+                    .fail();
+        }
+    }
+
+    private static void scanFile(Path testFile, Path examplesRoot, List<String> failures)
+            throws IOException {
+        final String testContent = Files.readString(testFile);
+
+        final Pattern packagePattern = Pattern.compile(
+            "getPackageLocation\\(\\)\\s*\\{\\s*return\\s*\"([^\"]+)\"");
```
romani, line 159: This code targeting:
```
public class WhitespaceAroundExamplesTest extends AbstractExamplesModuleTestSupport {
    @Override
    protected String getPackageLocation() {
        return "com/puppycrawl/tools/checkstyle/checks/whitespace/whitespacearound";
    }

```

Instead of crazy regexp .
As you have name of class, can you instantiate it by name and execute getPackageLocation() on Saturday ch object

romani (changes requested):
Items

Aman-Baliyan:
Please check the above message on change review

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExampleFileTest.java
Aman-Baliyan: @romani  sir, I have removed the regex as suggested. However, I’m now facing an issue because getPackageLocation() is protected.
To access it, I tried using reflection by creating an instance of the test class and invoking the method reflectively. Unfortunately, this requires using java.lang.reflect.Method, which is not allowed, so this approach will not pass the checks.
    ```
    final String testContent = Files.readString(testFile);

        final String pathStr = testFile.toString().replace(File.separatorChar, '.');
        final String className = pathStr.substring(pathStr.indexOf("com."),
                pathStr.length() - 5);
        final Class<?> testClass = Class.forName(className);
        final Object instance = testClass.getDeclaredConstructor().newInstance();
        final Method method = testClass.getDeclaredMethod("getPackageLocation");
        method.setAccessible(true);
        final String packageLocation = method.invoke(instance).toString();

        final Path exampleDir = examplesRoot.resolve(packageLocation); 
```

Given this limitation, I’m a bit stuck on how to proceed further without using reflection.
Or
Other then this i have check that there is a pattern in checks folder under resources if the corresponding ExampleTest file name is ArrayExampleTest or ArrayCheckExampleTest then its package under resources is array this is helpful in finding the name of package .
Could you please suggest an alternative or the recommended way to handle this scenario?

Aman-Baliyan: > @romani sir, I have removed the regex as suggested. However, I’m now facing an issue because getPackageLocation() is protected. To access it, I tried using reflection by creating an instance of the test class and invoking the method reflectively. Unfortunately, this requires using java.lang.reflect.Method, which is not allowed, so this approach will not pass the checks. ``` final String testContent = Files.readString(testFile);
> 
> ```
>     final String pathStr = testFile.toString().replace(File.separatorChar, '.');
>     final String className = pathStr.substring(pathStr.indexOf("com."),
>             pathStr.length() - 5);
>     final Class<?> testClass = Class.forName(className);
>     final Object instance = testClass.getDeclaredConstructor().newInstance();
>     final Method method = testClass.getDeclaredMethod("getPackageLocation");
>     method.setAccessible(true);
>     final String packageLocation = method.invoke(instance).toString();
> 
>     final Path exampleDir = examplesRoot.resolve(packageLocation); 
> ```
> 
> ```
> 
> Given this limitation, I’m a bit stuck on how to proceed further without using reflection.
> Or
> Other then this i have check that there is a pattern in checks folder under resources if the corresponding ExampleTest file name is ArrayExampleTest or ArrayCheckExampleTest then its package under resources is array this is helpful in finding the name of package .
> Could you please suggest an alternative or the recommended way to handle this scenario?
> ```

@romani please check this

romani: Please make this method public. No harm to have it public.

Aman-Baliyan: sir, regarding the suggestion to make  getPackageLocation()  public: I researched the impact and found that this method is overridden in over 300 test files across the codebase. Changing it in the base class (AbstractPathTestSupport) would require a massive, repository-wide change to update all those implementations. 
if you want this i can proceed but i am thinking another implementation to keep the change localized and minimize risk, I have instead made a wrapper public method in  
(AbstractExamplesModuleTestSupport.java) because all xdoc-example test inherit this class. Since the new  XdocsExampleFileTest only needs access to the locations for these example classes, this solves the issue perfectly without the overhead of a 300+ file commit. Let me know if this localized approach is acceptable!"

romani: Massive update is ok, you can do it by search and replace

Aman-Baliyan:
i am closing this pr because there is some end of line issue happens with all files and i am not finding any way to solve it .
i will create a new pr for this till morning

Issue #17685, opened 2025-08-25
Identation basicOffset force strict condition


sukolenvo:
I have read check documentation: https://checkstyle.sourceforge.io/checks/misc/indentation.html#Indentation
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Example1.java
# no errors

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
  "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
  "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
  <module name="TreeWalker">
    <module name="Indentation">
    </module>
  </module>
</module>


/var/tmp $ cat Example1.java
class Example1 {

    void processValues(MyField field) {
        field.getUsers()
            .stream()
            .map(
                "param1",
                "param2"
            ).collect("param1", "param2");


        field.getUsers()
            .stream()
            .map("param1", "param2")
            .collect(
                "param1",
                "param2"
            ); // ok

        field.getUsers().stream()
            .map("param1", "param2")
            .collect(
                "param1",
                "param2"
        );

        field.getUsers().stream()
            .map("param1", "param2")
            .collect(
                "param1",
                "param2"
            ); // violation is here !!!!
    }

    interface MyField {

        MyField getUsers();

        MyField stream();

        MyField map(String param1, String param2);

        MyField collect(String param1, String param2);
    }
}

/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-11.0.0-all.jar -c config.xml Example1.java
Starting audit...
[ERROR] Example1.java:32:13: 'method call rparen' has incorrect indentation level 12, expected level should be 8. [Indentation]
Audit done.
Checkstyle ends with 1 errors.

```

---

**Describe what you expect in detail.**

`Example1.java` contains 4 sample code blocks to perform similar operations. 3 of them successfully pass checkstyle validation and last one fails due to basicOffset expected to be 8, but got 12. Block 4 is very similar to other examples (especially 1 and 2) and I think such code should not raise a violation error. 

[...]

PR #18406, opened 2025-12-23
Issue #18118: fix indentation calculation logic for chained method call

### LINKED ISSUE #18118: Inconsistent behaviour of `Indentation` check

```bash
/var/tmp $ javac Example.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
    <module name="TreeWalker">
        <module name="Indentation"/>
    </module>
</module>

/var/tmp $ cat Example.java
import java.util.List;

class Example {
    static void main() {
        List.of("foo").stream()
            .findFirst()
            .ifPresentOrElse(
                System.out::println,
                () -> {
                    throw new IllegalStateException();
                }
            );
    }
}

/var/tmp $ java $RUN_LOCALE -jar checkstyle-12.1.2-all.jar -c config.xml Example.java
Starting audit...
[ERROR] /var/tmp/Example.java:12:13: 'method call rparen' has incorrect indentation level 12, expected level should be 8. [Indentation]
Audit done.
```

---

There should be no errors.

In `AnotherExample.java` I've replaced `List.of("foo").stream()` with `Stream.of("foo")` and there are no errors:

```bash
/var/tmp $ javac AnotherExample.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC

<module name="Checker">
    <module name="TreeWalker">
        <module name="Indentation"/>
    </module>
</module>

/var/tmp $ cat AnotherExample.java

class AnotherExample {
    static void main() {
        Stream.of("foo")
            .findFirst()
            .ifPresentOrElse(
                System.out::println,
                () -> {
                    throw new IllegalStateException();
                }
            );
    }
}

/var/tmp $ java $RUN_LOCALE -jar checkstyle-12.1.2-all.jar -c config.xml AnotherExample.java
Starting audit...
Audit done.
```

M-SaaD-H:
Fixes #18118

Problem
The Indentation check incorrectly flagged the closing parenthesis in chained method calls like:

```java
List.of("foo").stream()
    .findFirst()
    .ifPresentOrElse(
        System.out::println,
        () -> { throw new IllegalStateException(); }
    );  // False positive: expected level 8, but 12 is correct
```
However, using Stream.of("foo") instead produced no error, creating inconsistent behavior.

Solution
Modified MethodCallHandler.getIndentImpl() to add the line start column as an acceptable+ indent level for chained method calls. This allows correctly indented closing parens to be accepted while still validating actually incorrect indentation.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/MethodCallHandler.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/MethodCallHandler.java
@@ -217,9 +217,38 @@ public class MethodCallHandler extends AbstractExpressionHandler {
                     new IndentLevel(getIndent(), getBasicOffset()),
                     false, true);
 
-                checkRightParen(lparen, rparen);
+                checkRparenIndent(lparen, rparen);
                 checkWrappingIndentation(getMainAst(), getCallLastNode(getMainAst()));
             }
+        }
+    }
+
+    /**
+     * Checks the indentation of the right parenthesis for method calls.
+     *
+     * @param lparen left parenthesis associated with rparen
+     * @param rparen parenthesis to check
+     */
+    private void checkRparenIndent(DetailAST lparen, DetailAST rparen) {
+        final int rparenLevel = expandedTabsColumnNo(rparen);
+        final int lparenLevel = expandedTabsColumnNo(lparen);
+
+        final IndentLevel standardIndent = getIndent();
+
+        // For chained method calls, also allow rparen at the line start position
+        IndentLevel enhancedIndent = standardIndent;
+        if (getParent() instanceof MethodCallHandler) {
+            final int lineStart = getLineStart(getFirstAst(getMainAst()));
+            if (lineStart != standardIndent.getFirstIndentLevel()) {
+                enhancedIndent = IndentLevel.addAcceptable(standardIndent,
+                        new IndentLevel(lineStart));
+            }
+        }
+
+        if (rparenLevel != lparenLevel + 1
+                && !enhancedIndent.isAcceptable(rparenLevel)
+                && isOnStartOfLine(rparen)) {
+            logError(rparen, "rparen", rparenLevel);
         }
     }
 
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheckTest.java
@@ -349,6 +349,7 @@ public class IndentationCheckTest extends AbstractModuleTestSupport {
             "53:19: " + getCheckMessage(MSG_CHILD_ERROR, "method call", 18, 20),
             "54:15: " + getCheckMessage(MSG_ERROR, "method call rparen", 14, 16),
             "75:13: " + getCheckMessage(MSG_ERROR, "lambda arguments", 12, 16),
+            "81:16: " + getCheckMessage(MSG_ERROR, "method call rparen", 15, 8),
         };
         verifyWarns(checkConfig, getPath("InputIndentationMethodCallLineWrap.java"), expected);
     }
@@ -1601,10 +1602,10 @@ public class IndentationCheckTest extends AbstractModuleTestSupport {
         checkConfig.addProperty("throwsIndent", "4");
         final String fileName = getPath("InputIndentationChainedMethodCalls.java");
         final String[] expected = {
-            "33:5: " + getCheckMessage(MSG_CHILD_ERROR, "method call", 4, 8),
-            "38:5: " + getCheckMessage(MSG_ERROR, ".", 4, 8),
-            "39:5: " + getCheckMessage(MSG_ERROR, ".", 4, 8),
-            "42:5: " + getCheckMessage(MSG_ERROR, "new", 4, 8),
+            "36:5: " + getCheckMessage(MSG_CHILD_ERROR, "method call", 4, 8),
+            "41:5: " + getCheckMessage(MSG_ERROR, ".", 4, 8),
+            "42:5: " + getCheckMessage(MSG_ERROR, ".", 4, 8),
+            "45:5: " + getCheckMessage(MSG_ERROR, "new", 4, 8),
         };
         verifyWarns(checkConfig, fileName, expected);
     }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationChainedMethodCalls.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationChainedMethodCalls.java
@@ -14,6 +14,9 @@
 
 package com.puppycrawl.tools.checkstyle.checks.indentation.indentation;     //indent:0 exp:0
 
+import java.util.List;                                                      //indent:0 exp:0
+import java.util.stream.Stream;                                             //indent:0 exp:0
+
 class InputIndentationChainedMethodCalls {                                  //indent:0 exp:0
 
   public InputIndentationChainedMethodCalls(Object... params) {             //indent:2 exp:2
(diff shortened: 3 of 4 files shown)


Atharv3221:

Atharv3221:
@romani
This change fixes https://github.com/checkstyle/checkstyle/issues/17685

Atharv3221:
Diff is huge

M-SaaD-H:
Hi @Atharv3221 ,
Thanks for pointing that out.
Yes, both these issues arise from the same underlying cause. I raised this PR against the original issue (18118). But it should also address the second issue (17685). Should I just mention this one in the description of this PR or the second issue needs to be addressed in a new PR?

Atharv3221:
As per regression report change has caused huge difference in reported violations. I don't think all violations change were false positives. Means fix wasn't proper. Please add a test case which fails from this change.

M-SaaD-H:

M-SaaD-H:
Hi @Atharv3221 
I have added the test for this fix and generated the report. Can you please look into it?
Let me know if you need anything from my side.

Atharv3221:
@M-SaaD-H 
red one are the violation messages which were removed after the change and green one are added after change you can check the code its tested against by line number ahead.

M-SaaD-H:
@Atharv3221 
I reviewed the diff regression report for some of the projects. The added violations correspond to real cases in the affected files and are expected behavior for this check.

romani:
rebased in web

romani:
please add to inputs https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/17a959dcf706b8fabba0acaa4e28e040231a667e_2025104024/reports/diff/spring-framework/index.html#A13 looks like we missing this case in our Inputs

M-SaaD-H:
@romani 
Added the required inputs.

M-SaaD-H:
@romani
The CI failure is due to a 403 while resolving org.sonatype.oss:oss-parent:9 from Maven Central in GitHub Actions. The same dependency resolves successfully locally, so this appears to be a CI/network issue. Could you please take a look?

M-SaaD-H:
@romani 
All CIs are passing now.

romani:

M-SaaD-H:
@romani 
The failing CI is showing the following error. It is failing due to failure to parse the `.ci/pitest-survival-check-xml.groovy` file. I have seen some other PRs too, they shows the same warning, but this one fails due to GitHub actions timout (10min).
> [WARNING] There were problems parsing .ci/pitest-survival-check-xml.groovy
> [WARNING] org.openrewrite.groovy.GroovyParsingException: Failed to parse .ci/pitest-survival-check-xml.groovy at cursor position 11283. The surrounding characters in the original source are:
al Set<Mutation> result = new TreeSet<>(setOne)
    result.removeIf { mutation -> setTwo.contains(mutation) }
    return result
 }

``` 
 /**
 * A class to represent the XML {@code mutation} node.
*/
@EqualsAndHashCode(excludes = ["lineNumber", "unstable"]~cursor~>)
@Immutable
class Mutation implements Comparable<Mutation> {

    /**
     * Mutation nodes present in suppressions file do not have a {@code lineNumber}.
     * The {@code lineNumber} is set to {@code -1} for such mutations.
     */
    private sta
  org.openrewrite.groovy.GroovyParserVisitor.visit(GroovyParserVisitor.java:209)
 ```
 
> [INFO] Running recipe(s)...
>
> Too long with no output (exceeded 10m0s): context deadline exceeded

**Comment by romani** [comment]
@M-SaaD-H , please review diff report, too much changes, we need to be sure that no regressions happening

**Comment by romani** [comment]
@vivek-0509, please help to review this PR. You are expert in this Check.

**Comment by vivek-0509** [comment]
Sure, I’ll review it once I get back to my laptop.

File: src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/MethodCallHandler.java [hunk 1]
```diff
@@ -57,7 +57,17 @@ protected IndentLevel getIndentImpl() {
...
+                // as an acceptable level to handle cases like List.of().stream().method()
+                final IndentLevel containerIndent = container.getIndent();
+                final int lineStart = getLineStart(getFirstAst(getMainAst()));
+                if (lineStart == containerIndent.getFirstIndentLevel()) {
+                    indentLevel = containerIndent;
+                }
+                else {
+                    indentLevel = IndentLevel.addAcceptable(containerIndent,
+                            new IndentLevel(lineStart));
+                }
```
vivek-0509: The  fix  adds lineStart as an acceptable indentation level in getIndentImpl() which is causing significant regressions, introducing a lot of new false positive violations.The root cause is that getIndentImpl() is not solely used for checking the right parenthesis indentation. The IndentLevel object returned by this method is consumed by multiple components: checkRightParen(), checkExpressionSubtree(), getSuggestedChildIndent() etc. When we modify the returned IndentLevel to include lineStart as acceptable, all of these consumers are affected, not just the rparen check that we intended to fix. This causes the check to expect child elements (arguments, lambda expressions, string literals) at incorrect indentation levels.
A potential fix would be to create a targeted solution that only affects the rparen check without modifying getIndentImpl(). Since checkRightParen() is marked final, we could create a new private method checkRParenIndent() in MethodCallHandler that adds lineStart as acceptable specifically for the rparen check, then call this new method instead of the parent's checkRightParen(). This isolates the fix to only the rparen validation, avoiding false positives on child elements.

M-SaaD-H: done

vivek-0509:

vivek-0509:

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/MethodCallHandler.java
```diff
@@ -217,12 +217,43 @@ public void checkIndentation() {
...
+    private void checkRparenIndent(DetailAST lparen, DetailAST rparen) {
+        final int rparenLevel = expandedTabsColumnNo(rparen);
+        final int lparenLevel = expandedTabsColumnNo(lparen);
+
+        // Get the standard indent level
+        final IndentLevel standardIndent = getIndent();
+
+        // For chained method calls, also allow rparen at the line start position
+        IndentLevel enhancedIndent = standardIndent;
+        if (getParent() instanceof MethodCallHandler) {
```
vivek-0509, line 240: Only keep this comment, remove the other two comments, code is self explanatory

vivek-0509:
@romani 
Ok to merge from my side, the suggested approach (https://github.com/checkstyle/checkstyle/pull/18406#discussion_r2685064058) works as expected no regression caused.
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/47af876bed69675e8dc528ab2a87341522980fcb_2026112613/reports/diff/index.html

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/MethodCallHandler.java
M-SaaD-H: done

vivek-0509:
@M-SaaD-H We found a new surviving mutant please eliminate it
```
+    <sourceFile>MethodCallHandler.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.indentation.MethodCallHandler</mutatedClass>
+    <mutatedMethod>checkRparenIndent</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.MathMutator</mutator>
+    <description>Replaced integer addition with subtraction</description>
+    <lineContent>if (rparenLevel != lparenLevel + 1</lineContent>
+  </mutation>
```

M-SaaD-H:
@vivek-0509 @romani 
Mutant eliminated.

romani (approved):
thanks a lot !!!

Issue #18033, opened 2025-11-01
Resolve Pitest Suppressions - xpath


romani:
As a part of this issue, we have to kill the mutation listed below and remove the suppression from [pitest xxxx suppression.xml](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/)

To understanding how to kill the mutation 
Visit [Wiki page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports), [Pitest-docs](https://pitest.org/) and expected actions and workflow at https://github.com/checkstyle/checkstyle/issues/12341.

Target is to remove all `<mutation` from [Pitest-xpath](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-xpath-suppressions.xml)

Janhvibabani:
I am on it

romani:
@Janhvibabani , please share switch specific mutation you plan to address?

Each mutation should be in separate PR, numerous people can work on issue in parallel on different mutations

guptapratykshh:
No update on this since 2 weeks, i am on it

SURYANSHUAGRAWAL2006:
Hello Sir, No update on this issue sice 2 weeks, i am on it.

romani:
@SURYANSHUAGRAWAL2006, this issue doesn't not imply full fix by single person and single PR. Please make comment "I am on .... " And define single mutation you plan to handle

SURYANSHUAGRAWAL2006:
Hello Sir, I am on mutation in CheckstyleAntTask.java

Youssefalaa7:
I am on ReverseListIterator.java

aclfe:
I am on FollowingIterator.java

PR #18408, opened 2025-12-23
Issue #18033: Resolve Pitest Suppressions - xpath

Youssefalaa7:
Issue #18033: Removed the suppression of the mutation in Pitest Suppressions - xpath by removing the assignment in the constructor to null as it is redundant because unassigning it in java is the same as assigning it to null

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/xpath/iterators/ReverseListIterator.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/xpath/iterators/ReverseListIterator.java
@@ -19,10 +19,10 @@
 
 package com.puppycrawl.tools.checkstyle.xpath.iterators;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
+import com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil;
 import net.sf.saxon.om.NodeInfo;
 import net.sf.saxon.tree.iter.AxisIterator;
 
@@ -45,14 +45,8 @@ public class ReverseListIterator implements AxisIterator {
      * @param items the collection of nodes.
      */
     public ReverseListIterator(Collection<? extends NodeInfo> items) {
-        if (items == null) {
-            this.items = null;
-            index = -1;
-        }
-        else {
-            this.items = new ArrayList<>(items);
-            index = items.size() - 1;
-        }
+        this.items = UnmodifiableCollectionUtil.unmodifiableList(items);
+        index = this.items.size() - 1;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


romani:
CI is red

Youssefalaa7:
Done

src/main/java/com/puppycrawl/tools/checkstyle/xpath/iterators/ReverseListIterator.java
```diff
@@ -33,7 +33,7 @@ public class ReverseListIterator implements AxisIterator {
     /**
      * List of nodes.
      */
-    private final List<? extends NodeInfo> items;
+    private List<? extends NodeInfo> items;
```
romani: Can we keep final and put initial value as null here at declaration ?

romani (changes requested):
Question

src/main/java/com/puppycrawl/tools/checkstyle/xpath/iterators/ReverseListIterator.java
Youssefalaa7: No because final variables should be assigned exactly one time so we can not assign it two times one as default variable and one in the constructor and also it cannot be unassigned at all

romani: Yes, such an issue.
It is good to have it as final, it is good design.

Strange that this is first time we run into this, or we did some good workaround

src/main/java/com/puppycrawl/tools/checkstyle/xpath/iterators/ReverseListIterator.java
```diff
@@ -46,7 +46,6 @@ public class ReverseListIterator implements AxisIterator {
      */
     public ReverseListIterator(Collection<? extends NodeInfo> items) {
         if (items == null) {
-            this.items = null;
```
romani, line 49: https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.java is specially excluded from pitest.

We can utilize it make code ctor 
```
public ReverseListIterator(Collection<? extends NodeInfo> items) {
this.items = unmodifiableList(items);
            index = this.items.size() - 1;
}
```
unmodifiableList will be new method that handle check for null to create empty collection.

romani (changes requested):
Please find how this class is excluded from mutation in pom.xml it might require do extra change in pom.xml .

Let's try:

Youssefalaa7:
Done

romani:
@Youssefalaa7, please send update for other mutations like this, I think it will be easy win win for you.

Youssefalaa7:
Thanks! I'll look for similar mutations and send PRs for those.

PR #18413, opened 2025-12-24
Issue #18033: Resolve xpath Pitest Supressions FollowingIterator

aclfe:
Issue #18033: Resolve xpath Pitest Supressions FollowingIterator

**Problem**:
By removing `siblingEnum = null`, we created a "surviving mutation" where the iterator wasn't cleared after use.

Existing tests missed this because the `next()` method either creates a fresh iterator or finishes immediately afterward. Since the output stayed the same, the tests passed. messed up the internal state

**Solution**:
Added a test case `testSiblingEnumIsClearedOnExhaustion` that:
- Initializes a `FollowingIterator` and iterates through it until exhaustion (all nodes visited).
- Verifies that the `siblingEnum` field has been explicitly set to null after the iterator completes.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/xpath/iterators/FollowingIteratorTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/xpath/iterators/FollowingIteratorTest.java
@@ -21,10 +21,16 @@ package com.puppycrawl.tools.checkstyle.xpath.iterators;
 
 import static com.google.common.truth.Truth.assertWithMessage;
 import static com.puppycrawl.tools.checkstyle.internal.utils.XpathIteratorUtil.findNode;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 import org.junit.jupiter.api.Test;
 
+import net.sf.saxon.om.AxisInfo;
 import net.sf.saxon.om.NodeInfo;
+import net.sf.saxon.tree.iter.AxisIterator;
 
 public class FollowingIteratorTest {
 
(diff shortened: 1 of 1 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/xpath/iterators/FollowingIteratorTest.java
```diff
@@ -73,4 +73,25 @@ public void testNoSibling() {
...
+        try (FollowingIterator iterator = new FollowingIterator(startNode)) {
+            NodeInfo node = iterator.next();
+            while (node != null) {
+                node = iterator.next();
+            }
+
+            final java.lang.reflect.Field field =
+                FollowingIterator.class.getDeclaredField("siblingEnum");
+            field.setAccessible(true);
+            final Object siblingEnum = field.get(iterator);
```
romani: Sorry, no pure unit testing is allowed.
Please cover it by execution of some xpath filter

aclfe: I understand. Thanks for pointing that out, I'll make the changes as requested soon

src/test/java/com/puppycrawl/tools/checkstyle/xpath/iterators/FollowingIteratorTest.java
aclfe, line 1: Hello. I've made changes from before. I understand before I was working with private variables. 
New approach is using Mockito. 

I've tried a lot using an XPath filter as requested, **but the output is identical**. It is impossible to kill this mutation with an XPath filter from what I've seen. the iterator eventually returns `null`

aclfe: @romani 

Before making the CI green, I just want to make sure if this approach is accepted. Please let me know.

romani: All tests should be by checks and filters execution

Issue #18025, opened 2025-11-01
Resolve Pitest Suppressions - tree-walker


romani:
As a part of this issue, we have to kill the mutation listed below and remove the suppression from [pitest xxxx suppression.xml](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/)

To understanding how to kill the mutation 
Visit [Wiki page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports), [Pitest-docs](https://pitest.org/) and expected actions and workflow at https://github.com/checkstyle/checkstyle/issues/12341.

Target is to remove all `<mutation` from [Pitest-tree-walker](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-tree-walker-suppressions.xml)

guptapratykshh:
i am on it

Youssefalaa7:
I am on Comparator.nullsLast(Comparator.naturalOrder())) mutation

Youssefalaa7:
I am on .thenComparing(AbstractCheck::getId,

Aman-Baliyan:
i am on

```
  <mutation unstable="false">
    <sourceFile>TreeWalker.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.TreeWalker</mutatedClass>
    <mutatedMethod>&lt;init&gt;</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
    <description>Removed assignment to member variable javaParseExceptionSeverity</description>
    <lineContent>private SeverityLevel javaParseExceptionSeverity = SeverityLevel.ERROR;</lineContent>
  </mutation>
```

Youssefalaa7:
I am on javaParseExceptionSeverity = SeverityLevel.ERROR

romani:
Final fix at https://github.com/checkstyle/checkstyle/pull/18428

PR #18417, opened 2025-12-25
Issue #18025: Resolve pitest Suppression Tree walker

Aman-Baliyan:
Issue: #18025 

Added test to kills mutation for testJavaParseExceptionSeverity
what this test do-

This test asserts that TreeWalker parsing exceptions always produce an AuditEvent with a non-null default severity of ERROR.
It is added to kill a PIT mutation that sets the default severity to null, ensuring parsing errors are consistently reported with a valid severity.

### DIFF
--- a/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
+++ b/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
@@ -1,11 +1,3 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <mutation unstable="false">
-    <sourceFile>TreeWalker.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.TreeWalker</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable javaParseExceptionSeverity</description>
-    <lineContent>private SeverityLevel javaParseExceptionSeverity = SeverityLevel.ERROR;</lineContent>
-  </mutation>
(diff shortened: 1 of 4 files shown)


config/pitest-suppressions/pitest-tree-walker-suppressions.xml
```diff
@@ -1,11 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<suppressedMutations>
-  <mutation unstable="false">
```
romani, line 3: please keep file but remove "<mutation"  only, in this case it will generate good diff in next PRs when author cause more mutants to appear.

romani (changes requested):
items:

romani:
>This test asserts that TreeWalker parsing exceptions always produce an AuditEvent with a non-null default severity of ERROR.

looks like pitest just remove assignment to class field `javaParseExceptionSeverity` so it  stays null, so null comes to Violation constructor, and it should just be returned as null to Audit listener.
Can you run test by verifyWithInlineXmlConfig and as severity is printed in xml as ERROR, it should fail test that executed on mutated code, so pitest mutation will be killed.

Aman-Baliyan:
> > This test asserts that TreeWalker parsing exceptions always produce an AuditEvent with a non-null default severity of ERROR.
> 
> looks like pitest just remove assignment to class field `javaParseExceptionSeverity` so it stays null, so null comes to Violation constructor, and it should just be returned as null to Audit listener. Can you run test by verifyWithInlineXmlConfig and as severity is printed in xml as ERROR, it should fail test that executed on mutated code, so pitest mutation will be killed.

ok sir i will try it

Aman-Baliyan:
creating a new pr with expected changes because i accidently rebase it

romani:
You can push with force to rewrite GitHub version of branch

Aman-Baliyan:
new pr -> #18428

Issue #18317, opened 2025-12-16
MissingOverride: update doc example to show method without inheritDoc


romani:
https://checkstyle.org/checks/annotation/missingoverride.html#MissingOverride

extend examples to show that methods without javadoc does not need Override even it present in parent class

confusion detected at https://github.com/checkstyle/checkstyle/issues/17561

AradhyaSharma31:
I would like to work on this

0xChaitanya:
Is this issue still open or will it be closed now?

romani:
@AradhyaSharma31 still working on it

romani:
Fix is merged

Issue #17561, opened 2025-08-06
Google style: Override is required on explicitly declared accessor method for a record


mohitsatr:
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

From: https://google.github.io/styleguide/javaguide.html#s6.1-override-annotation
> A method is marked with the `@Override` annotation whenever it is legal. This includes ... and an explicitly declared accessor method for a record component

Code:
```java
import java.util.Locale;

/** some javadoc. */
public record InputOverrideAlwaysUsedForRecord(String name, int age) {

  /** some javadoc. */
  public String name() { // Expected violation
    return name.toUpperCase(Locale.ROOT);
  }
}

record Container(String fileName, int port) {

  public int port() { // Expected violation
    return port;
  }
}
```

Cli:
```
$ java -jar checkstyle-10.26.1-all.jar -c google_checks.xml InputOverrideAlwaysUsedForRecord.java 
Starting audit...
Audit done.
```

Expected: two violations

vivek-0509:
I am on it

romani:
https://checkstyle.org/google_style.html#a6.1

this paragraph is marked as not covered at all.
So it is becomes weird how we can improve coverage of it.


--------

We have run into limitation of Checkstyle here as we do not know structure of parent classes, so we can not define what method is overridden and what method is not. 

BUT

in records , that has no "implements" (https://openjdk.org/jeps/395), we have all data in record declaration to make this decision what is overriden. In case of "implements" is used, interface can be declared in separate class/file/project, so checkstyle has no access to it, to make decision is method is overriden in record or it is record specific method.


https://checkstyle.org/checks/annotation/missingoverride.html#MissingOverride
is specifically for `Verifies that the @Override annotation is present when the @inheritDoc javadoc tag is present.` so without javadoc tag it does not do validation. 


-------

Resolution, we can not use MissingOverride in google style, we should not extend this Check. We should create new Check. If name be very specific with target we can avoid declaring limitation that we can not demand annoations on methods implemented from interface. But we can mention it, to let users not even try to ask for Check extension (as not all aware of checkstyle limitations).
 
Proposed name:
MissingOverrideOnRecordComponentAccessorMethod


from JEP:
> The meaning of the `@Override` annotation was extended to include the case where the annotated method is an explicitly declared **accessor method for a record component**.


[...]

PR #18426, opened 2025-12-26
Issue #18317: update doc example to show method without inheritDoc

AradhyaSharma31:
Issue: #18317

Extend the MissingOverride documentation with a 5 and 6 examples that clarifies when the `@Override` annotation is required. The new example shows:

1. No javadoc, no `@Override` needed
2. Javadoc present but no {`@inheritDoc`}, no `@Override` needed

This resolves confusion reported in issue #17561 where developers were unsure about the relationship between javadoc and the `@Override` requirement

### DIFF
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/missingoverride/Example1.java
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/missingoverride/Example1.java
@@ -13,6 +13,10 @@ class ParentClass1{
 
   public void test2(){}
 
+  public void test5(){}
+
+  public void test6(){}
+
 }
 // xdoc section -- start
 class Example1 extends ParentClass1 {
@@ -39,5 +43,14 @@ class Example1 extends ParentClass1 {
   public static void test4() {
 
   }
+
+  // No javadoc, no @Override needed
+  public void test5() {}
+
+  // Javadoc present but no {@inheritDoc}, no @Override needed
+  /**
+   * This is a sample javadoc.
+   */
+  public void test6() {}
 }
 // xdoc section -- end
--- a/src/site/xdoc/checks/annotation/missingoverride.xml
+++ b/src/site/xdoc/checks/annotation/missingoverride.xml
@@ -99,6 +99,15 @@ class Example1 extends ParentClass1 {
   public static void test4() {
 
   }
+
+  // No javadoc, no @Override needed
+  public void test5() {}
+
+  // Javadoc present but no {@inheritDoc}, no @Override needed
+  /**
+   * This is a sample javadoc.
+   */
+  public void test6() {}
 }
 </code></pre></div><hr class="example-separator"/>
         <p id="Example2-config">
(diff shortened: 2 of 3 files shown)


AradhyaSharma31:
@romani CI is green now. Could you please review and let me know if any changes are required?

romani:
GitHub, generate website

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/missingoverride/Example1.java
```diff
@@ -39,5 +39,18 @@ private void test3() {
   public static void test4() {
 
   }
+
+  // No javadoc, no @Override needed
+  public void test5() {
+
+  }
```
romani: Please make method single line.

Please add new method to 
```
class ParentClass1{
  public void test1(){}

  public void test2(){}

}
```

romani (changes requested):
Items

romani:
single commit always, all methods should in parent class

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/missingoverride/Example1.java
AradhyaSharma31: I've added the method to ParentClass1 as a single line. I kept test5() in Example1 as multi-line for consistency with the other methods in that class. Should I also make test5() single line in Example1.java and the XML configuration file, or is the current approach what you had in mind?

romani: Better to break this consistency and make it compact. We can do all as compact in next PR

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/missingoverride/Example1.java
```diff
@@ -13,6 +13,8 @@ public void test1(){}
 
   public void test2(){}
 
+  public void test5(){}
+
```
romani, line 17: Please add test6()

romani (changes requested):
Items

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/missingoverride/Example1.java
AradhyaSharma31: I’ve updated everything as requested:

1. Both methods are now in the parent class.
2. Both methods are in single-line format in both the `.java` and `.xml` files.

Should I create a PR to make all the docs methods compact across all the files?

romani:
GitHub, generate website

romani (approved):
Thanks a lot.
Please send one more PR to make all method to be in compact mode.
And we probably need to show users parent class to make it clear that all methods in example1 are overriden

AradhyaSharma31:
> Thanks a lot. Please send one more PR to make all method to be in compact mode. And we probably need to show users parent class to make it clear that all methods in example1 are overriden

Thanks! I’ll send a follow-up PR making all methods compact within the MissingOverride example.

Please let me know if you’d like this applied more broadly beyond these 2 examples.

romani:
>Please let me know if you’d like this applied more broadly beyond these 2 examples.

Please send PRs for each Check separately, there might be nuances where noncompact user for some reason

PR #18428, opened 2025-12-26
Issue #18025: Resolve pitest Suppression Tree walker

Aman-Baliyan:
Issue: #18025 

Added test to kills mutation for testJavaParseExceptionSeverity

uses inline verifyWithInlineXmlConfig in place of audit event as discuss in previous pr


Aman-Baliyan:
verifyWithInlineXmlConfig approach is not working the mutation survives 

```
diff --git a/config/pitest-suppressions/pitest-tree-walker-suppressions.xml b/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
index 4bb5719..2147a23 100644
--- a/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
+++ b/config/pitest-suppressions/pitest-tree-walker-suppressions.xml
@@ -1,3 +1,11 @@
 <suppressedMutations>
+    <sourceFile>TreeWalker.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.TreeWalker</mutatedClass>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
+    <description>Removed assignment to member variable javaParseExceptionSeverity</description>
+    <lineContent>private SeverityLevel javaParseExceptionSeverity = SeverityLevel.ERROR;</lineContent>
+  </mutation>
 </suppressedMutations>
EOF
```

If we set javaParseExceptionSeverity to "error" in the test's XML config, the TreeWalker
 setter is called, which overwrites the null value from the mutation back to ERROR. This effectively "fixes" the mutated state before the verification logic runs, causing the test to pass identical to the original code. so the mutation survives.

Sir, please tell me about your thoughts on this.

romani:
Let's not set this property in config, to see default severity on parser error

Aman-Baliyan:
> Let's not set this property in config, to see default severity on parser error

the test already exists in the test file, the mutation is not being killed because verifyWithInlineXmlConfig only verifies the violation messages and their locations, not the severity of the violations. Since the mutated code affects severity handling rather than the message content, this change is not detected by the existing verification logic. As a result, the mutation can only be reliably killed by using a custom AuditEventListener, which explicitly captures and asserts the severity from the emitted audit events.

test -> 

```
@Test
    public void testSkipFileOnJavaParseExceptionConfigSeverityDefault() throws Exception {
        final String path =
                getNonCompilablePath(
                        "InputTreeWalkerSkipParsingExceptionConfigSeverityDefault.java");
        final String[] expected = {
            "1: " + getCheckMessage(TreeWalker.PARSE_EXCEPTION_MSG, "IllegalStateException")
                  + " occurred while parsing file " + path + ".",
        };
        verifyWithInlineXmlConfig(path, expected);
    }
```

inpu file -> 

```
/*xml
<module name="Checker">
  <module name="TreeWalker">
    <property name="skipFileOnJavaParseException" value="true"/>
    <module name="ConstantName"/>
  </module>
</module>
*/

// violation 9 lines above
// non-compiled syntax: bad file for testing
public clazz InputTreeWalkerSkipParsingExceptionConfigSeverityDefault {}
```

romani:
There are verify methods that validates whole xml report by whole content, can we use them?

Such method heavily used to validate xml sarif report generation.

Aman-Baliyan:
> There are verify methods that validates whole xml report by whole content, can we use them?
> 
> Such method heavily used to validate xml sarif report generation.

@romani sir,

```
protected final void verifyWithInlineXmlConfig(String filePath, String... expected)
            throws Exception {
                InlineConfigParser.parseWithXmlHeader(filePath);
        final Configuration xmlConfig =
                testInputConfiguration.getXmlConfiguration();
        verifyViolations(xmlConfig, filePath, testInputConfiguration.getViolations());
        verify(xmlConfig, filePath, expected);
    }
```
inside the code verifyWithInlineXmlConfig the verify method is used which takes xmlConfig but inside deep down it also checks for only messages because the method which is used inside  `verify(xmlConfig, filePath, expected);` is

 ```
private Map<String, List<String>> getActualViolations() throws IOException {
        // process each of the lines
        try (ByteArrayInputStream inputStream =
                new ByteArrayInputStream(stream.toByteArray());
            LineNumberReader lnr = new LineNumberReader(
                new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
            final Map<String, List<String>> actualViolations = new HashMap<>();
            for (String line = lnr.readLine(); line != null;
                 line = lnr.readLine()) {
                if ("Audit done.".equals(line) || line.contains("at com")) {
                    break;
                }
                // have at least 2 characters before the splitting colon,
                // to not split after the drive letter on Windows
                final String[] actualViolation = line.split("(?<=.{2}):", 2);
                final String actualViolationFileName = actualViolation[0];
                final String actualViolationMessage = actualViolation[1];

                actualViolations
                        .computeIfAbsent(actualViolationFileName, key -> new ArrayList<>())
                        .add(actualViolationMessage);
            }

            return actualViolations;
        }
    }
```
it only returns messages not severity from violations which are generated when checker process files

**Comment by romani** [comment]
https://github.com/checkstyle/checkstyle/blob/d19fb31b3a786d80b86c452b99a91ff16957a787/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java#L185

**Comment by Aman-Baliyan** [comment]
> https://github.com/checkstyle/checkstyle/blob/d19fb31b3a786d80b86c452b99a91ff16957a787/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java#L185

it is not possible with verifyWithInlineConfigParserAndXmlLogger because it only takes compilable input but in our scenario we need non-compilable code but i think it is possible with verifyWithInlineConfigParserAndDefaultLogger i am working on it

**Comment by romani** [comment]
Create new method to satisfy your needs

**Comment by Aman-Baliyan** [comment]
> Create new method to satisfy your needs
@romani sir,
the mutation is killed by verifyWithInlineConfigParserAndDefaultLogger but when i generate the report the file pitest-tree-walker-suppression.xml automatically gots deleted then what i do now remake it  or leave its deleted 
image

File: src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/treewalker/InputTreeWalkerSkipParsingExceptionSeverityError.java [hunk 1]
```diff
@@ -0,0 +1,13 @@
+/*xml
+<module name="Checker">
+  <module name="TreeWalker">
+    <property name="skipFileOnJavaParseException" value="true"/>
+    <property name="javaParseExceptionSeverity" value="error"/>
```
romani: no properties should be there

Aman-Baliyan: > no properties should be there
sir, 
i did not push new code yet i am just asking that when i run the pit locally it deleted the pitest-tree-walker-suppressin file so i remake it or left it deleted

Aman-Baliyan: > no properties should be there

the new test i wrote is 

xml

```
/*xml
<module name="Checker">
  <module name="TreeWalker">
    <property name="skipFileOnJavaParseException" value="true"/>
    <module name="ConstantName"/>
  </module>
</module>
*/

```
```
@Test
    public void testJavaParseExceptionSeverityDefault() throws Exception {
        final String inputFile = "InputTreeWalkerSkipParsingExceptionConfigSeverityDefault.java";
        final String expectedInfoFile = "ExpectedSeverityInfo.txt";
        final String expectedErrorFile = "ExpectedSeverityError.txt";

        final ByteArrayOutputStream infoStream = new ByteArrayOutputStream();
        final ByteArrayOutputStream errorStream = new ByteArrayOutputStream();
        final DefaultLogger dl = new DefaultLogger(infoStream, OutputStreamOptions.CLOSE,
                errorStream, OutputStreamOptions.CLOSE);

        verifyWithInlineConfigParserAndDefaultLogger(
                getNonCompilablePath(inputFile),
                getPath(expectedInfoFile),
                getPath(expectedErrorFile),
                dl, infoStream, errorStream);
    }
```

this test kills the mutation but due this test i got a new error in ant verifiaction phase
`[ERROR] [checkstyle] [ERROR] C:\Users\amanc\Desktop\CheckStyle\checkstyle\src\test\java\com\puppycrawl\tools\checkstyle\TreeWalkerTest.java:89:1: Class Data Abstraction Coupling is 8 (max allowed is 7) classes [ByteArrayOutputStream, DefaultConfiguration, DefaultLogger, File, FileContents, FileText, PackageObjectFactory, TreeWalker]. [ClassDataAbstractionCoupling]`

should i suppress it or wrote the test in defaultLoggerTest.java

romani: Please suppress this checkstyle violation, it ok that test classes are growing in size and more and more classes are used.

romani:
`[ERROR]   AllTestsTest.testAllInputsHaveTest:59->walkVisible:107->lambda$testAllInputsHaveTest$1:60->verifyInputFile:190->verifyInputFile:222 Resource must be named after a Test like 'InputMyCustomCase.java' and be in the sub-package of the test like 'mycustom' for test 'MyCustomCheckTest': com/puppycrawl/tools/checkstyle/treewalker/ExpectedSeverityInfo.txt
expected to be true`

romani (approved):
Thanks a lot.
This is great

PR #18157, opened 2025-11-24
Issue #17487: Align `rewrite` config with `spotless`

Pankraz76:
### Issue #17487: Align `rewrite` config with `spotless`

PR #18430, opened 2025-12-26
Issue #17487: Improve CI config #18157

Pankraz76:
### Issue #17487: Improve `CI` config #18157

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -552,7 +552,6 @@ assembly-run-all-jar)
     exit $fail;
   fi
   rm .ci-temp/output.json
-
   ;;
 
 check-since-version)
@@ -1365,30 +1364,27 @@ spotless)
   ./mvnw -e --no-transfer-progress clean spotless:check -P spotless-autofix
   ;;
 
-openrewrite-recipes)
-  echo "Cloning and building OpenRewrite recipes..."
+sanity)
+  export MAVEN_OPTS="-Xmx4g"
   PROJECT_ROOT="$(pwd)"
-  export MAVEN_OPTS="-Xmx4g -Xms2g"
-
+  echo "Sanity Check 🕊️"
+  echo "Cloning and building OpenRewrite recipes 📥"
   cd /tmp
   git clone https://github.com/checkstyle/checkstyle-openrewrite-recipes.git
   cd checkstyle-openrewrite-recipes
   mvn -e --no-transfer-progress clean install -DskipTests
-
   cd "$PROJECT_ROOT"
-
-  echo "Running Checkstyle validation to get report for openrewrite..."
+  echo "Run Checkstyle validation to get report for openrewrite ☑️"
   set +e
   ./mvnw -e --no-transfer-progress clean compile antrun:run@ant-phase-verify
   set -e
-  echo "Running OpenRewrite recipes..."
+  echo "Run OpenRewrite recipes ☑️"
   ./mvnw -e --no-transfer-progress -Drewrite.recipeChangeLogLevel=INFO \
-    rewrite:run -P checkstyle-autofix
-
-  echo "Checking for uncommitted changes..."
+    clean rewrite:run -P checkstyle-autofix
+  echo "Checking for uncommitted changes 📋"
   ./.ci/print-diff-as-patch.sh target/rewrite.patch
-
   rm -rf /tmp/checkstyle-openrewrite-recipes
+  echo "Complete ✅"
   ;;
 
 *)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,4 @@
+name: CI
 version: 2.1
 
 jobs:
@@ -29,14 +30,14 @@ jobs:
           path: /home/circleci/project/target/inspection-results
 
   validate-with-maven-script:
-    description: "Runs a maven script using the job name as the argument."
+    description: Runs a maven script using the job name as the argument.
     parameters: &script_parameters
       image-name:
         type: string
-        default: "cimg/openjdk:21.0.6"
-        description: "docker image to use"
+        default: cimg/openjdk:21.0.6
+        description: docker image to use
       command:
-        description: "command to run"
+        description: command to run
         type: string
         default: ""
     docker:
@@ -61,14 +62,14 @@ jobs:
             - .mvn/wrapper
 
   validate-with-script:
-    description: "Runs a non-maven script using the job name as the argument."
+    description: Runs a non-maven script using the job name as the argument.
     parameters: *script_parameters
     docker:
       - image: << parameters.image-name >>
     steps:
       - checkout
       - run:
-          name: run << parameters.command >>
+          name: Run << parameters.command >>
           command: |
             sudo apt update
             sudo apt install -y xmlstarlet --no-install-recommends apt-utils
@@ -84,7 +85,7 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Run sonarqube
+          name: Run SonarQube
           command: |
             export PR_NUMBER=$CIRCLE_PR_NUMBER
             export PR_BRANCH_NAME=$CIRCLE_BRANCH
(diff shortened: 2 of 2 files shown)


.ci/validation.sh
```diff
@@ -1365,30 +1365,26 @@ spotless)
...
   export MAVEN_OPTS="-Xmx4g -Xms2g"
-
+  echo "Sanity Check 🦢"
+  echo "Cloning and building OpenRewrite recipes 📥"
   cd /tmp
   git clone https://github.com/checkstyle/checkstyle-openrewrite-recipes.git
   cd checkstyle-openrewrite-recipes
   mvn -e --no-transfer-progress clean install -DskipTests
-
-  cd "$PROJECT_ROOT"
```
Pankraz76, line 1378: why this coupling on `PROJECT_ROOT`? 

If we can inline, then it seems obsolete.

Pankraz76:
encoding kind of broken

@romani 

image


image

.circleci/config.yml
```diff
@@ -1,15 +1,13 @@
-version: 2.1
-
+name: "CI ♻️"
```
Pankraz76: tryout remove double quotes.

.circleci/config.yml
```diff
@@ -1,15 +1,13 @@
...
   run-inspections:
     docker:
-      - image: amitkumardeoghoria/idea-image:jdk21-idea2022.3.3-v1
-
+      - image: "amitkumardeoghoria/idea-image:jdk21-idea2022.3.3-v1"
     steps:
       - checkout
       - run:
-          name: Print versions
+          name: "Print Versions 📋"
```
Pankraz76: tryout remove double quotes.

.circleci/config.yml
```diff
@@ -18,320 +16,297 @@ jobs:
...
   validate-with-script:
     description: "Runs a non-maven script using the job name as the argument."
     parameters: *script_parameters
     docker:
-      - image: << parameters.image-name >>
+      - image: "<< parameters.image-name >>"
     steps:
       - checkout
       - run:
-          name: run << parameters.command >>
```
Pankraz76, line 71: bug

.ci/validation.sh
```diff
@@ -144,52 +144,52 @@ test)
 
 test-de)
   ./mvnw -e --no-transfer-progress clean integration-test failsafe:verify \
-    -Dsurefire.options='-Duser.language=de -Duser.country=DE -Xms1g -Xmx2g'
```
Pankraz76, line 147: The key difference is the **presence of the `-Xms1g` flag**, which sets the **initial heap size**.

## What's the difference?

1. **First option (with `-Xms1g`):**
   ```
   -Dsurefire.options='-Duser.language=sq -Duser.country=AL -Xms1g -Xmx2g'
   ```
   - `-Xms1g`: Sets **initial** heap size to 1GB
   - `-Xmx2g`: Sets **maximum** heap size to 2GB
   - JVM starts with 1GB heap and can grow up to 2GB

2. **Second option ("slim" version without `-Xms`):**
   ```
   -Dsurefire.options='-Duser.language=sq -Duser.country=AL -Xmx2g'
   ```
   - Only `-Xmx2g`: Sets **maximum** heap size to 2GB
   - JVM starts with default initial heap (usually much smaller, e.g., 256MB-512MB)
   - Heap grows gradually as needed, up to 2GB

## Which is better? Can we use the slim version?

**Yes, you can likely use the slim version** in most cases. Here's why:

### Advantages of slim version (without `-Xms`):
- **Faster startup**: Smaller initial heap means JVM starts quicker
- **Memory efficient**: Only uses what it needs when it needs it
- **Better for resource-constrained environments**: Doesn't pre-allocate 1GB upfront

### When to keep `-Xms1g` (first option):
- If your tests **consistently need ~1GB heap** anyway
- To **reduce garbage collection overhead** during execution
- If you want **more predictable performance** (avoid heap resizing)
- In production-like test environments where performance consistency matters

## Recommendation:

**Start with the slim version** (without `-Xms`) unless:
1. You experience performance issues due to frequent heap resizing
2. You know your tests consistently need ~1GB+ heap
3. You're running performance tests and need consistent conditions

The slim version is generally better for CI/CD pipelines and development machines as it's more resource-friendly while still having the 2GB safety net if needed.

Pankraz76: can we apply convention or is the min config really mandatory?

.ci/validation.sh
```diff
@@ -552,7 +552,6 @@ assembly-run-all-jar)
     exit $fail;
   fi
   rm .ci-temp/output.json
-
```
Pankraz76, line 555: gonna catch em all.

.circleci/config.yml
```diff
@@ -29,14 +30,14 @@ jobs:
...
+    description: Runs a maven script using the job name as the argument.
     parameters: &script_parameters
       image-name:
         type: string
-        default: "cimg/openjdk:21.0.6"
-        description: "docker image to use"
+        default: cimg/openjdk:21.0.6
+        description: docker image to use
       command:
-        description: "command to run"
```
Pankraz76, line 39: this kind of gives attention to where its really mandatory to use `"double quotes"`.

Image

.circleci/config.yml
```diff
@@ -117,138 +118,138 @@ workflows:
...
-          command: "./.ci/validation.sh no-exception-hbase"
+          command: ./.ci/validation.sh no-exception-hbase
       - validate-with-maven-script:
-          name: "no-exception-Pmd-elasticsearch-lombok-ast"
+          name: no-exception-pmd-elasticsearch-lombok-ast ✅
           image-name: *cs_img
-          command: "./.ci/validation.sh no-exception-Pmd-elasticsearch-lombok-ast"
+          command: ./.ci/validation.sh no-exception-pmd-elasticsearch-lombok-ast
       - validate-with-maven-script:
-          name: "no-exception-alot-of-projects"
```
Pankraz76, line 186: typo.

- https://www.linguee.de/englisch-deutsch/uebersetzung/a+lot+of.html

.github/workflows/diff-report.yml
```diff
@@ -56,7 +56,6 @@ jobs:
           GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
 
       - run: 'echo We print it here for this action to work'
-        if: 'true'
```
Pankraz76, line 59: .github/workflows/diff-report.yml:59:13: condition "true" is always evaluated to true. remove the if: section [if-cond]
   |
59 |         if: 'true'
   |             ^~~~~~

romani (changes requested):
you need to keep your AI on leash.

only targeted explicit updates in PR and no other changes.

Pankraz76:
> you need to keep your AI on leash.
> 
> only targeted explicit updates in PR and no other changes.

yes updated. Let me know it still want to extract the CI fix. Imho this is helpful with the icons giving the false positive good indication. 

Can extract for cheap and just keep the sanity stuff but actually this one adding the UI suggar so this whats this PR is about imho.

.circleci/config.yml
```diff
@@ -117,138 +118,138 @@ workflows:
...
+          command: ./.ci/validation.sh no-exception-spotbugs
       - validate-with-maven-script:
-          name: "no-exception-spoon"
+          name: no-exception-spoon ✅
           image-name: *cs_img
-          command: "./.ci/validation.sh no-exception-spoon"
+          command: ./.ci/validation.sh no-exception-spoon
       - validate-with-maven-script:
-          name: "no-exception-spring-framework"
+          name: no-exception-spring-framework ✅
```
romani: no emojs

.github/workflows/diff-report.yml
Pankraz76: this one unrelated and should fail main branch.

.circleci/config.yml
Pankraz76: come on this is all the fun its easy for human to recognize, of course its kind of crazy and opinion i see, yes.

Pankraz76: no exception is hard to tell if you just see icon and its easy, also is kind of underlines the intend imho, but yes undo.

Pankraz76: all fun gone.

.circleci/config.yml
```diff
@@ -107,148 +108,137 @@ jobs:
...
-          command: "./.ci/validation.sh no-exception-spring-framework"
+          command: ./.ci/validation.sh no-exception-spring-framework
       - validate-with-maven-script:
-          name: "no-exception-hbase"
+          name: no-exception-hbase
           image-name: *cs_img
-          command: "./.ci/validation.sh no-exception-hbase"
+          command: ./.ci/validation.sh no-exception-hbase
       - validate-with-maven-script:
-          name: "no-exception-Pmd-elasticsearch-lombok-ast"
```
Pankraz76, line 182: brOkeN cAsE.

.ci/validation.sh
```diff
@@ -1365,30 +1364,27 @@ spotless)
...
   ;;
 
-openrewrite-recipes)
-  echo "Cloning and building OpenRewrite recipes..."
+sanity)
+  export MAVEN_OPTS="-Xmx4g"
   PROJECT_ROOT="$(pwd)"
-  export MAVEN_OPTS="-Xmx4g -Xms2g"
-
+  echo "Sanity Check 🕊️"
```
Pankraz76, line 1370: is here icon issue or allowed?

romani: No emoji. All is plain text.

romani:
I don't see any reason in this PR.
Please stop using word "sanity" at all.

Pankraz76:
> Please stop using word "sanity" at all.

its a concept around major projects like gradle:

- https://github.com/gradle/gradle/blob/5478a4a62071a9f97660cb5f99abac82cb85eab7/CONTRIBUTING.md?plain=1#L254
- https://en.wikipedia.org/wiki/Sanity_check

Pankraz76:
> I don't see any reason in this PR.

- fixing lot of flaws.
- Improve focus by taking away a lot of double quotes.
  - of course might be taste, still i suggest final version without any overhead.

Pankraz76:
> I don't see any reason in this PR.

also its about arch. software design.

As we see rewrite some kind of impl. detail we want to hide away. Rewrite is the how but not the what reasoning, therefore i would change it.

We are actually imposing with this some large generic quality goal by using rewrite, this might change or will include some other detailed tool as well. So this is the step to give more space to the business case instead of the random impl. detail.

.ci/validation.sh
Pankraz76: yes, thx.

Pankraz76:
reopen after https://github.com/checkstyle/checkstyle/pull/18432

Issue #18028, opened 2025-11-01
Resolve Pitest Suppressions - api


romani:
As a part of this issue, we have to kill the mutation listed below and remove the suppression from [pitest xxxx suppression.xml](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/)

To understanding how to kill the mutation 
Visit [Wiki page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports), [Pitest-docs](https://pitest.org/) and expected actions and workflow at https://github.com/checkstyle/checkstyle/issues/12341.

Target is to remove all `<mutation` from [Pitest-api](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-api-suppressions.xml)

aclfe:
I'm on AbstractFileSetCheck

aclfe:
I am on AbstractCheck.java

romani:
Final fix is merged

PR #18443, opened 2025-12-28
Issue #18028: Resolve Pitest Suppressions - api

Youssefalaa7:
Issue #18028: Removed the suppression of the mutation charset = null;  in Pitest Suppressions - api

### DIFF
--- a/config/pitest-suppressions/pitest-api-suppressions.xml
+++ b/config/pitest-suppressions/pitest-api-suppressions.xml
@@ -17,31 +17,4 @@
     <description>removed call to com/puppycrawl/tools/checkstyle/api/AbstractFileSetCheck::getSeverityLevel</description>
     <lineContent>getSeverityLevel(),</lineContent>
   </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>AbstractViolationReporter.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.api.AbstractViolationReporter</mutatedClass>
-    <mutatedMethod>setSeverity</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable severityLevel</description>
-    <lineContent>severityLevel = SeverityLevel.getInstance(severity);</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>FileText.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.api.FileText</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable charset</description>
-    <lineContent>charset = null;</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>Violation.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.api.Violation</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
-    <description>Removed assignment to member variable args</description>
-    <lineContent>this.args = null;</lineContent>
-  </mutation>
 </suppressedMutations>
--- a/pom.xml
+++ b/pom.xml
@@ -5005,7 +5005,8 @@
                 <mutator>EXPERIMENTAL_ARGUMENT_PROPAGATION</mutator>
                 <mutator>EXPERIMENTAL_BIG_DECIMAL</mutator>
                 <mutator>EXPERIMENTAL_BIG_INTEGER</mutator>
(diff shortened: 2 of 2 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.java
```diff
@@ -88,6 +89,16 @@ public static <S, T> List<T> unmodifiableList(Collection<S> items, Class<T> elem
...
     }
 
+    /**
+     * Returns an unmodifiable view of a Charset.
+     *
+     * @param charset The Charset to make unmodifiable.
+     * @return An unmodifiable view of the provided Charset.
+     */
+    public static Charset unmodifiableCharSet(Charset charset) {
+        return charset;
```
romani: This code is nonsense.
We need different fix

romani (changes requested):
Items

src/main/java/com/puppycrawl/tools/checkstyle/utils/UnmodifiableCollectionUtil.java
Youssefalaa7: what about making two of the constructors in FileText.java to call this method it will be like this
```
public static Charset unmodifiableCharSet(Charset charset) {
        return Objects.requireNonNullElseGet(charset, Charset::defaultCharset);
    }
```
is this a good solution?

romani: I think we need to open issue on pitest.
Removal of this line doesn't make code compile-able, so it is defect of pitest, or there should be some settings to skip such cases, can you create issue?

Youssefalaa7: you mean that pitest should know that the variable is final and removing its assignment will result in error so it should not remove this assignment am i understanding right?

romani: Yes

Youssefalaa7: Thanks, i will create an issue there

Youssefalaa7: I think that pitest ignores the compilation it only focuses that when they remove an assignment it results in a difference in the tests that at least one test will fail. And that's why i think checkstyle have mutations only on the assignment to null for final variables but we do not have mutations for assigning to other thing other than null even when it is final as in this case of the previous pr Issue #18033

```
if (items == null) {
            this.items = null;
    index = -1;
        }
        else {
            this.items = new ArrayList<>(items);
            index = items.size() - 1;
        }
```

romani: Pitest works with bytecode, a lot more can be done  on bytecode that is not allowed by compilation, but pitest still can improve.

Youssefalaa7: Done I created an issue https://github.com/hcoles/pitest/issues/1440

Youssefalaa7: @romani 
The Pitest maintainer confirmed that EXPERIMENTAL_MEMBER_VARIABLE is disabled by default because it produces low-quality mutations (like the equivalent mutations we're seeing). See their response in the issue.
I tested disabling the EXPERIMENTAL_MEMBER_VARIABLE mutator in our codebase, and it would actually help us remove most of the suppressions we currently have. The mutations it produces are indeed low quality and not providing much value.
Should we proceed with disabling this mutator globally in our Pitest configuration? This would clean up our suppressions significantly.

config/pitest-suppressions/pitest-api-suppressions.xml
```diff
@@ -27,15 +27,6 @@
     <lineContent>severityLevel = SeverityLevel.getInstance(severity);</lineContent>
   </mutation>
 
-  <mutation unstable="false">
-    <sourceFile>FileText.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.api.FileText</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.MemberVariableMutator</mutator>
```
romani, line 34: Let's disable mutator at pom.xml  with comment until issue in pitest are resolved

romani (changes requested):
Items

config/pitest-suppressions/pitest-api-suppressions.xml
Youssefalaa7: Thanks, I will send a PR that comments EXPERIMENTAL_MEMBER_VARIABLE mutator and delete all the suppressions in various pitests related to this mutator

romani: only in "api" profile.

Youssefalaa7:
Done

romani:
we will wait for pitest project to do update.

Youssefalaa7:
Thanks a lot

Issue #17663, opened 2025-08-20
Indentation: missing violations for '->' and ',' in return statement continuation


Tieshiyang:
I have read check documentation: https://checkstyle.sourceforge.io/checks/whitespace/indentation.html
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

Code:
```bash
public class Test {
  int sum(int a, int b, int c) {
    return a
    + b     // Expected Violation (and Reported)
    + c;    // Expected Violation (and Reported)
  }

  java.util.function.Function<String, Integer> test_arrow() {
    return (String s) ->
    s.length(); // Expected Violation (but ignored)
  }

  int test_comma() {
    return sum(
    1, // Expected Violation (but ignored)
    2, // Expected Violation (but ignored)
    3  // Expected Violation (but ignored)
    );
  }
}
```
Config:
```
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <property name="localeLanguage" value="en"/>
    <module name="TreeWalker">
        <module name="Indentation">
            <property name="basicOffset" value="2"/>
            <property name="braceAdjustment" value="2"/>
            <property name="caseIndent" value="2"/>
            <property name="throwsIndent" value="4"/>
            <property name="lineWrappingIndentation" value="4"/>
            <property name="arrayInitIndent" value="2"/>
        </module>
    </module>
</module>
```
Cli:
```
$ java -jar checkstyle-10.26.1-all.jar -c config.xml Test.java
Starting audit...
[ERROR] Test.java:4:5: 'method def' child has incorrect indentation level 4, expected level should be 8. [Indentation]
[ERROR] Test.java:5:5: 'method def' child has incorrect indentation level 4, expected level should be 8. [Indentation]
Audit done.
Checkstyle ends with 2 errors.
```

---

Observed behavior:
In return with +, indentation errors are correctly reported.
In return with lambda arrow -> or commas ,, no indentation violations are reported, even though they are also continuation lines.
This makes the behavior inconsistent: some continuation elements are checked, while others are ignored.

---

romani:
[...]

PR #18445, opened 2025-12-28
Issue #17663: Fix missing indentation violations for '->' and ',' in return statement continuations

guptapratykshh:
Issue: #17663

The indentation check does not catch indentation problems for lambda arrows (->) and commas (,) in return statement continuations. This results in false negatives. Return statements with incorrectly indented lambda arrow continuations or comma-separated arguments go unmarked as violations. In contrast, operators like + in similar situations are correctly identified.

For example:
- `return a + b + c;` - violations for `+` operators are correctly reported 
- `return (String s) -> s.length();` - violation for lambda arrow continuation is ignored 
- `return sum(1, 2, 3);` - violations for comma-separated arguments are ignored 

This inconsistency makes the indentation checking behavior incomplete and breaks the rule that all continuation elements in return statements should be checked in the same way. 

I added the `checkReturnStatementLineWrapping()` method to the `BlockParentHandler` class. This method checks the line wrapping indentation for return statements that span multiple lines. It ensures that lambda arrows (`->`) and commas (`,`) in return statement continuations are properly checked for indentation violations. This makes the behavior consistent with operators like `+`, which were already being checked.

### DIFF
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -4962,14 +4962,42 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/BlockParentHandler.java</fileName>
     <specifier>not.interned</specifier>
     <message>attempting to use a non-@Interned comparison operand</message>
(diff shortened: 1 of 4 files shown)


romani:
ci is red , please fix

PR #18456, opened 2025-12-29
Issue #18028: Resolve Pitest Supressions - api - abstractfileset

aclfe:
Issue #18028: Resolve Pitest Supressions - api - abstractfileset

Updated testLineColumnLog: We added a check to ensure the severity level is ERROR.
Updated testMultiFileFireErrors: We added the exact same check there.

shows "KILLED" 

image

image

Please let me know if this approach is accepted or if any changes are needed, thank you.


romani (approved):
Thanks a lot

romani:
I think the only left mutation can be killed in same way in other test class. I can be easy win win for you @aclfe .

aclfe:
> I think the only left mutation can be killed in same way in other test class. I can be easy win win for you @aclfe .

Yeah I was planing to do it. I'll post the message in the issue

Issue #17568, opened 2025-08-06
EmptyLineSeparator ignores configured tokens: reports violations for unlisted tokens and skips listed ones


Tieshiyang:
I have read check documentation: https://checkstyle.sourceforge.io/checks/whitespace/emptylineseparator.html#EmptyLineSeparator
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

Code:
```bash
/*
 * Copyright 2025
 */


package benchimarkfilesmakeup;
import java.util.List; // Expected violation
public class Test {
}
```
Config:
```
/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <property name="localeLanguage" value="en"/>
    <!-- Checks for whitespace                               -->
    <!-- See http://checkstyle.org/checks/whitespace/index.html -->
    <module name="TreeWalker">
        <module name="EmptyLineSeparator">
            <property name="tokens" value="IMPORT"/>
            <property name="allowMultipleEmptyLines" value="false"/>
        </module>
    </module>
</module>
```
Cli:
```
/var/tmp $ java -jar checkstyle-10.26.1-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:8:1: 'CLASS_DEF' should be separated from previous line. [EmptyLineSeparator]
Audit done.
```

---

The [Checkstyle documentation](https://checkstyle.sourceforge.io/checks/whitespace/emptylineseparator.html#EmptyLineSeparator) states that:

“Checks for empty line separators before package, all import declarations, fields, constructors, methods, nested classes, static initializers and instance initializers.”

This suggests that only the tokens listed in the tokens property should be checked.

However, the behavior is inconsistent:
Observed issues:

False Positive (FP): It reports a violation before the CLASS_DEF line — even though CLASS_DEF is not in the token list.

False Negative (FN): It fails to report a violation for the IMPORT statement that is not preceded by an empty line — even though IMPORT is in the token list.
---

romani:
it might demand  empty line between import and class, be cause import is your target.

[...]

PR #18461, opened 2025-12-30
Issue #17568: Updates EmptyLineSeparator docs

Aman-Baliyan:
Issue: #17568 

Updates the docs of EmptyLineSeparator because the first documentation is confusing. Add two more examples so simplify the tests.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
@@ -35,13 +35,19 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
 
 /**
  * <div>
- * Checks for empty line separators before package, all import declarations,
+ * Checks for empty line separators after,
  * fields, constructors, methods, nested classes,
  * static initializers and instance initializers.
  * </div>
  *
  * <p>
- * Checks for empty line separators before not only statements but
+ * For package declaration it checks for both before and after
+ * line separators and for import declarations it checks for
+ * empty line separator after last import declaration.
+ * </p>
+ *
+ * <p>
+ * Checks for empty line separators after not only statements but
  * implementation and documentation comments and blocks as well.
  * </p>
  *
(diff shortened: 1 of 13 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
```diff
@@ -35,13 +35,19 @@
...
  * <div>
- * Checks for empty line separators before package, all import declarations,
+ * Checks for empty line separators after,
  * fields, constructors, methods, nested classes,
  * static initializers and instance initializers.
  * </div>
  *
  * <p>
- * Checks for empty line separators before not only statements but
+ * For package declaration it checks for both before and after
```
romani, line 44: I might miss this, but I don't see it in examples, can you point?

romani (changes requested):
Very good, item:

romani:
GitHub, generate website

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
romani: Let's put such violation to example

Aman-Baliyan: > I might miss this, but I don't see it in examples, can you point?

sir, the test Example 2 is for this 
   ```
 @Test
    public void testExample2() throws Exception {
        final String[] expected = {
            "16:1: " + getCheckMessage(MSG_SHOULD_BE_SEPARATED, "package"),
            "17:1: " + getCheckMessage(MSG_SHOULD_BE_SEPARATED, "import"),
        };

        verifyWithInlineConfigParser(getPath("Example2.java"), expected);
```

what i did is -

in first example i put all violations and in further examples i take single token shows the violations message made by that token.

Like in example 2 i use package-def token which wants space before and after so its gives two violation one for package line and another for import line .

in third example i take import token and shows its violations and so on in further examples

romani: you are right, due to long line of package we put violation comment on next line and this was confusing.

ok, after attentive reading it works good.

romani (approved):

Issue #129, opened 2014-04-15
Methods returning array and incorrect indentation


pavel-sakun:
This is continuation of the https://sourceforge.net/p/checkstyle/bugs/518/ thread.

I think there is a problem with the Indentation test when using arrays.
The following line gives a problem (see attachment for complete code example):

test.getArrayOfStrings()[0] = "A string";

I'm using the Eclipse checkstyle plugin, but that imo just a gui for checkstyle, and the problem is in checkstyle itself.

Under Miscellaneous -> Indentation, I use the default check (basicOffset=4, braceAdjustment=0, caseInd=4).

When the line is indented on 8 positions (what we use) I get this message:
Indentation: method call child at indentation level 8 not at correct indentation, 12
If I indent it at 12 position, I get another message:
Indentation: method def child at indentation level 12 not at correct indentation, 8

There does not seem to be any way to validate the line, so I think this is a bug.

Source file https://gist.github.com/pavel-sakun/7d54890dc2bc39d8a802
config https://raw.githubusercontent.com/pentaho/pentaho-coding-standards/master/pentaho_checkStyle.xml

Error message: method call child at indentation level 4 not at correct indentation, 6

PR #18473, opened 2025-12-31
Issue #6294: Kill Pitest mutation for IndexHandler registration

vivek-0509:
Issue #6294: Kill Pitest mutation for IndexHandler registration

The IndexHandler prevents IndentationCheck from descending into array index expressions. To verify this, a new test case was added with intentionally invalid indentation inside an array index.

*   **Behavior with IndexHandler (Current)**: The check stops at `INDEX_OP`, ignoring the invalid child indentation (Pass).
*   **Behavior without IndexHandler (Mutation)**: The check descends into the subtree, detects the invalid indentation, and fails .

### DIFF
--- a/config/pitest-suppressions/pitest-indentation-suppressions.xml
+++ b/config/pitest-suppressions/pitest-indentation-suppressions.xml
@@ -75,15 +75,6 @@
   <mutation unstable="false">
     <sourceFile>HandlerFactory.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.checks.indentation.HandlerFactory</mutatedClass>
-    <mutatedMethod>&lt;init&gt;</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.VoidMethodCallMutator</mutator>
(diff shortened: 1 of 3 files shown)


vivek-0509:
**Issue #129 Bug:** Code like getMyArray()[1] = "value" always reported incorrect indentation because `INDEX_OP` wasn't registered. MethodCallHandler would get `AssignHandler` as its parent, which suggested wrong indentation levels.

**PR #219 Fix:** Added `IndexHandler` with an empty checkIndentation() method as a bridge in the handler chain.

Later future prs added a guard in `MethodCallHandler.checkIndentation()` that checks `if (exprNode.getParent().getType() == TokenTypes.SLIST)`. This skips indentation checking for method calls nested inside other expressions.

**Why IndexHandler is Dead Code:** For getMyArray()[1] = "value", the grandparent is ASSIGN (not SLIST), so the guard fails and getIndentImpl() is never called. The IndexHandler parent relationship is never used. The statement's base indentation is verified by block-level checking (SlistHandler), not by the individual expression handlers.

romani:

romani:
@vivek-0509 , thre is diff, please investigate and add more sample code to our Inputs.

vivek-0509:
@romani i have added the test case, mutant is killed.

PR #18475, opened 2025-12-31
Issue #18023: (Good to merge, but concurrently merged other) Resolve Pitest Sup - filters - SuppressWithNearbyCommentFilterTest

aclfe:
Issue #18023: Resolve Pitest Sup - filters - SuppressWithNearbyCommentFilterTest

testAcceptUsesCachedSuppressionsOnConfigChange and testToStringOfTagClassWithMessage both killed mutations

Diff Regression config: https://raw.githubusercontent.com/aclfe/configs/refs/heads/main/SuppressWithNearbyCommentFilter.xml

### DIFF
--- a/config/pitest-suppressions/pitest-filters-suppressions.xml
+++ b/config/pitest-suppressions/pitest-filters-suppressions.xml
@@ -1,11 +1,3 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <mutation unstable="false">
-    <sourceFile>SuppressWithNearbyCommentFilter.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.filters.SuppressWithNearbyCommentFilter</mutatedClass>
-    <mutatedMethod>accept</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter::getFileContents</description>
-    <lineContent>if (getFileContents() != currentContents) {</lineContent>
-  </mutation>
 </suppressedMutations>
--- a/pom.xml
+++ b/pom.xml
@@ -5146,6 +5146,11 @@
               <excludedTestClasses>
                 <param>*.Input*</param>
               </excludedTestClasses>
+              <excludedMethods>
+                <!-- Since https://github.com/checkstyle/checkstyle/pull/18475 -->
+                <!-- performance optimization for caching file-level suppression tags -->
+                <param>initIfDifferent</param>
+              </excludedMethods>
               <coverageThreshold>100</coverageThreshold>
               <mutationThreshold>94</mutationThreshold>
               <timeoutFactor>${pitest.plugin.timeout.factor}</timeoutFactor>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
@@ -106,6 +107,9 @@ public class SuppressWithNearbyCommentFilter
     @XdocsPropertyType(PropertyType.PATTERN)
     private String idFormat;
 
+    /** Cached absolute path of the file being processed to avoid null pointer exception. */
+    private String cachedFileAbsolutePath = "";
+
     /**
      * Specify negative/zero/positive value that defines the number of lines
      * preceding/at/following the suppression comment.
@@ -119,7 +123,7 @@ public class SuppressWithNearbyCommentFilter
      * are reassigned to the next FileContents, at which time filtering for
      * the current FileContents is finished.
      */
-    private WeakReference<FileContents> fileContentsReference = new WeakReference<>(null);
+    private WeakReference<FileContents> fileContentsReference;
 
     /**
      * Setter to specify comment pattern to trigger filter to begin suppression.
@@ -222,19 +226,45 @@ public class SuppressWithNearbyCommentFilter
         boolean accepted = true;
 
         if (event.violation() != null) {
-            // Lazy update. If the first event for the current file, update file
-            // contents and tag suppressions
             final FileContents currentContents = event.fileContents();
 
-            if (getFileContents() != currentContents) {
-                setFileContents(currentContents);
-                tagSuppressions();
+            if (currentContents != null) {
+                cachedFileAbsolutePath = initIfDifferent(
+                        currentContents.getFileName(),
+                        cachedFileAbsolutePath,
+                        () -> {
+                            setFileContents(currentContents);
+                            tagSuppressions();
+                            return currentContents.getFileName();
+                        });
             }
+
             if (matchesTag(event)) {
                 accepted = false;
             }
         }
         return accepted;
+    }
+
+    /**
+     * Lazy initialization utility that updates cached state only when the key changes.
+     * This is a performance optimization to avoid redundant reprocessing.
+     *
+     * @param <T> the type of the cached object
+     * @param current the current value to compare
+     * @param cached the previously cached value
+     * @param supplier the supplier to provide new value if current differs from cached
+     * @return the cached value if it equals current, otherwise the value from supplier
+     */
+    private static <T> T initIfDifferent(T current, T cached, Supplier<T> supplier) {
+        final T result;
+        if (cached.equals(current)) {
+            result = cached;
+        }
+        else {
+            result = supplier.get();
+        }
+        return result;
     }
 
     /**
--- a/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
@@ -363,6 +363,38 @@ public class SuppressWithNearbyCommentFilterTest
                 .isTrue();
     }
 
+    @Test
+    public void testSuppressWithNearbyCommentFilterMutation() throws Exception {
+        final String[] suppressed = CommonUtil.EMPTY_STRING_ARRAY;
+        final String[] expected = {
+            "28:17: "
+                + getCheckMessage(AbstractNameCheck.class,
+                    MSG_INVALID_PATTERN, "A1", "^[a-z][a-zA-Z0-9]*$"),
+            "30:17: "
+                + getCheckMessage(AbstractNameCheck.class,
+                    MSG_INVALID_PATTERN, "B1", "^[a-z][a-zA-Z0-9]*$"),
+        };
+        verifyFilterWithInlineConfigParser(
+            getPath("InputSuppressWithNearbyCommentFilterMutation.java"),
+            expected, removeSuppressed(expected, suppressed));
+    }
+
+    @Test
+    public void testSuppressWithNearbyCommentFilterMutation2() throws Exception {
+        final String[] suppressed = CommonUtil.EMPTY_STRING_ARRAY;
+        final String[] expected = {
+            "28:17: "
+                + getCheckMessage(AbstractNameCheck.class,
+                    MSG_INVALID_PATTERN, "C1", "^[a-z][a-zA-Z0-9]*$"),
+            "30:17: "
+                + getCheckMessage(AbstractNameCheck.class,
+                    MSG_INVALID_PATTERN, "D1", "^[a-z][a-zA-Z0-9]*$"),
+        };
+        verifyFilterWithInlineConfigParser(
+            getPath("InputSuppressWithNearbyCommentFilterMutation2.java"),
+            expected, removeSuppressed(expected, suppressed));
+    }
+
     private void verifySuppressedWithParser(String fileName, String... suppressed)
             throws Exception {
         verifyFilterWithInlineConfigParser(fileName, ALL_MESSAGES,
@@ -779,6 +811,40 @@ public class SuppressWithNearbyCommentFilterTest
         assertWithMessage("Invalid tags size")
             .that(tags2)
             .isEmpty();
+    }
+
+    @Test
+    public void testSameFileDoesNotRecreateTags() {
+        final SuppressWithNearbyCommentFilter filter = new SuppressWithNearbyCommentFilter();
+        final String filename = "testFile.java";
+        final FileContents contents = new FileContents(
+                new FileText(new File(filename),
+                        Collections.singletonList("//SUPPRESS CHECKSTYLE ignore")));
+        contents.reportSingleLineComment(1, 0);
+
+        final Violation violation = new Violation(1, null, null, null, null, Object.class, null);
+        final TreeWalkerAuditEvent event1 = new TreeWalkerAuditEvent(contents, filename,
+                violation, null);
+
+        filter.accept(event1);
+        final List<?> tags1 = TestUtil.getInternalState(filter, "tags", List.class);
+        assertWithMessage("Tags should be created on first call")
+            .that(tags1)
+            .hasSize(1);
+
+        final Object firstTag = tags1.getFirst();
+
+        final TreeWalkerAuditEvent event2 = new TreeWalkerAuditEvent(contents, filename,
+                violation, null);
+        filter.accept(event2);
+        final List<?> tags2 = TestUtil.getInternalState(filter, "tags", List.class);
+
+        assertWithMessage("Tags should still have one element")
+            .that(tags2)
+            .hasSize(1);
+        assertWithMessage("Same tag object should be retained when processing same file")
+            .that(tags2.getFirst())
+            .isSameInstanceAs(firstTag);
     }
 
     /**
--- a/config/checker-framework-suppressions/checker-methods-resource-fenum-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-methods-resource-fenum-suppressions.xml
@@ -166,6 +166,28 @@
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java</fileName>
+    <specifier>argument</specifier>
+    <message>incompatible argument for parameter obj of Object.equals.</message>
+    <lineContent>if (cached.equals(current)) {</lineContent>
+    <details>
+      found   : T extends @FenumTop Object
+      required: @FenumUnqualified Object
+    </details>
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java</fileName>
+    <specifier>method.invocation</specifier>
+    <message>call to equals(java.lang.Object) not allowed on the given receiver.</message>
+    <lineContent>if (cached.equals(current)) {</lineContent>
+    <details>
+      found   : T extends @FenumTop Object
+      required: @FenumUnqualified Object
+    </details>
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/gui/MainFrame.java</fileName>
     <specifier>argument</specifier>
     <message>incompatible argument for parameter horizontalAlignment of JLabel constructor.</message>
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -7685,6 +7685,20 @@
 
   <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java</fileName>
+    <specifier>dereference.of.nullable</specifier>
+    <message>dereference of possibly-null reference cached</message>
+    <lineContent>if (cached.equals(current)) {</lineContent>
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
+    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java</fileName>
+    <specifier>initialization.field.uninitialized</specifier>
+    <message>the default constructor does not initialize field fileContentsReference</message>
+    <lineContent>private WeakReference&lt;FileContents&gt; fileContentsReference;</lineContent>
+  </checkerFrameworkError>
+
+  <checkerFrameworkError unstable="false">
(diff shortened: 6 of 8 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
```diff
@@ -508,6 +530,19 @@ public void testToStringOfTagClassWithId() {
                     + "tagCheckRegexp=.*, tagMessageRegexp=null, tagIdRegexp=.*]");
     }
 
+    @Test
+    public void testToStringOfTagClassWithMessage() {
```
romani, line 551: Please add a comment above method that explains why we can not use verifySuppressedWithParser
I agree that toString is not used in regular usage of Check, so it is impossible to hit such method toString

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
```diff
@@ -485,6 +485,28 @@ public void testAcceptNullFileContents() {
                 .isTrue();
     }
 
+    @Test
+    public void testAcceptUsesCachedSuppressionsOnConfigChange() {
```
romani: Please explain in comment why verifySuppressedWithParser can not be used?

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
aclfe: I checked `verifyFilterWithInlineConfigParser`'s definition from `verifySuppressedWithParser`, it abstracts away the creation of the SuppressWithNearbyCommentFilter

Need fillter object's call .setCommentFormat() at run time for caching mutation

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
aclfe: I'll add it asap

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
romani: not good excuse
https://github.com/checkstyle/checkstyle/blob/14d064ec80a0f2d0e4101f57d527f192f087a975/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java#L221-L229

we just need to process two files.

romani: try to cover it by usage of https://github.com/checkstyle/checkstyle/blob/14d064ec80a0f2d0e4101f57d527f192f087a975/src/test/java/com/puppycrawl/tools/checkstyle/AbstractModuleTestSupport.java#L292-L294

romani:
I recommend to move coverage for "removed call to java/lang/String::valueOf" to separate PR and merge quickly.

other mutation look like need more attention

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
```diff
@@ -485,6 +485,46 @@ public void testAcceptNullFileContents() {
                 .isTrue();
     }
 
+    // /**
+    //  * We cannot use {@link #verifySuppressedWithParser(String, String...)} here because
```
aclfe: I commented the previous implementation for showcase purpose of the requested changes

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
```diff
@@ -485,6 +485,46 @@ public void testAcceptNullFileContents() {
...
+    @Test
+    public void testAcceptUsesCachedSuppressions() throws Exception {
+        final String filePath1 = getPath("InputSuppressWithNearbyCommentFilterMutation.java");
+        final String filePath2 = getPath("InputSuppressWithNearbyCommentFilterMutation2.java");
+
+        final String[] expected = {
+            "28:17: Name 'A' must match pattern '^[a-z][a-zA-Z0-9]*$'.",
+        };
+
+        verifyWithInlineConfigParser(filePath2, filePath1, expected);
```
aclfe: As requested `verifyWithInlineConfigParser` is being used. Upon running the `,/mvnw test`, it passes. Though the mutation isn't killed. 

image

I believe the output violations remain identical, verifyWithInlineConfigParser cannot detect this equivalent mutantation 

Mutation Ci Test: [Link](https://github.com/checkstyle/checkstyle/actions/runs/20681091627/job/59375439669?pr=18475)

romani: It is ok to not see solution on first try, I try to challenge you to go deeper and find solution that are not obvious and needs more analytical skills.

romani: please debug test to see why mutation is not killed.
you might need place more violations to both files.

aclfe: Hello. I investigated the code more and tried many approaches. 

The equivalent mutation can not be killed easily by using standard verifyInlineConfigParser. Mutation has no impact on the final logic or output of the filter. 

I'm actively working on an approach to make it timeout.

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
romani: There is similar code at other filter 
https://github.com/checkstyle/checkstyle/blob/a76f70ac6df80abee1fa934a51a9d212526d6471/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyTextFilter.java#L165

Is it covered by pitest?

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
romani: Please consider https://github.com/checkstyle/checkstyle/pull/18475#discussion_r2663458810

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
aclfe: Romani, thank you so much for mentioning this file. This in fact helps me. 

I ran the pit report on the file you gave me. In reality, both mutations 
`if (!cachedFileAbsolutePath.equals(eventFileTextAbsolutePath)) { ` and `if (getFileContents() != currentContents) { ` are similar in nature. Same concept, just one works with disk, other with memory. 

I checked the test to kill this mutation ` if (!cachedFileAbsolutePath.equals(eventFileTextAbsolutePath)) { `  and it follows my initial approach (that's now commented)

https://github.com/checkstyle/checkstyle/blob/a76f70ac6df80abee1fa934a51a9d212526d6471/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyTextFilterTest.java#L431

I've attached the test case which is killing the similar mutation in the file you've given me. 

This test case is very similar to my initial approach

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
```diff
@@ -485,6 +485,46 @@ public void testAcceptNullFileContents() {
...
     }
 
+    // /**
+    //  * We cannot use {@link #verifySuppressedWithParser(String, String...)} here because
+    //  * it abstracts away the creation of {@code SuppressWithNearbyCommentFilter}.
+    //  * We need access to the filter instance to modify its configuration at runtime
+    //  * for caching verification.
+    //  */
+    // @Test
+    // public void testAcceptUsesCachedSuppressionsOnConfigChange() {
```
aclfe: My initial approach. 

See comment for explanation. https://github.com/checkstyle/checkstyle/pull/18475#discussion_r2665593507

aclfe: What I can do to increase the readability is shift the files into `src/test/resources/com/puppycrawl/tools/checkstyle/filters/suppresswithnearbycommentfilter` rather than have them instantiated directly here

config/pitest-suppressions/pitest-filters-suppressions.xml
```diff
@@ -1,14 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <mutation unstable="false">
-    <sourceFile>SuppressWithNearbyCommentFilter.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.filters.SuppressWithNearbyCommentFilter</mutatedClass>
-    <mutatedMethod>accept</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter::getFileContents</description>
-    <lineContent>if (getFileContents() != currentContents) {</lineContent>
```
romani, line 9: so mutation make it `if (null != currentContents) {`
and it is always TRUE and, so we need some way to make it different  in behavior. 

https://github.com/checkstyle/checkstyle/blob/07aceec3ed703c639f1cbdf55ba8c64ea1131065/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java#L221-L232

romani: it is interesting that we compare objects , not by equals, but by  obj reference, that will never be same .

https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/FileContents.java


can we change filter to compare absolute pathes
```
final String eventFileTextAbsolutePath = event.getFileName();
if (!cachedFileAbsolutePath.equals(eventFileTextAbsolutePath)) {
```
like we did at https://github.com/checkstyle/checkstyle/blob/a76f70ac6df80abee1fa934a51a9d212526d6471/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyTextFilter.java#L159-L166

aclfe: That makes sense. The way I see it, if we are comparing by object reference and FileContents instances are distinct for every event, the condition is effectively always true, which explains why the mutant survives.

I will refactor `SuppressWithNearbyCommentFilter` matching the logic in SuppressWithNearbyTextFilter

aclfe:
I refractored the code and it produced 3 mutations related to the refactored code.  Those mutations have been killed. 

CI error is unrelated

config/pitest-suppressions/pitest-filters-suppressions.xml
```diff
@@ -3,9 +3,9 @@
...
     <sourceFile>SuppressWithNearbyCommentFilter.java</sourceFile>
     <mutatedClass>com.puppycrawl.tools.checkstyle.filters.SuppressWithNearbyCommentFilter</mutatedClass>
-    <mutatedMethod>accept</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter::getFileContents</description>
-    <lineContent>if (getFileContents() != currentContents) {</lineContent>
+    <mutatedMethod>&lt;init&gt;</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.ConstructorCallMutator</mutator>
+    <description>removed call to java/lang/ref/WeakReference::&lt;init&gt;</description>
+    <lineContent>private WeakReference&lt;FileContents&gt; fileContentsReference = new WeakReference&lt;&gt;(null);</lineContent>
```
romani: This is not expected. Can we we kill it?

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
```diff
@@ -486,6 +489,39 @@ public void testAcceptNullFileContents() {
...
+        contents1.reportSingleLineComment(28, 19);
+
+        final FileText text2 = new FileText(commonFile,
+                Files.readAllLines(Path.of(
+                    getPath("InputSuppressWithNearbyCommentFilterMutation2.java"))));
+        final FileContents contents2 = new FileContents(text2);
+        contents2.reportSingleLineComment(28, 19);
+
+        final Violation violation =
+            new Violation(28, null, null, null, null, Object.class, null);
```
romani: Please cover it by verifyxxxx methods.

aclfe: Hi romani,

I think my previous explanation might have been weak and didn't explain why verifyxxx methods don't seem to work for these specific mutations.

Specifically, in verifyxxx methods, we call: 

https://github.com/checkstyle/checkstyle/blob/2abca93a8237ee293d482798fc0eade0d6ce69f1/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java#L388-L392

it calls on createChecker(config). Now on line 283 of createCheckerConfig()

https://github.com/checkstyle/checkstyle/blob/2abca93a8237ee293d482798fc0eade0d6ce69f1/src/it/java/org/checkstyle/base/AbstractItModuleTestSupport.java#L280-L283

As we see, line 283 makes a `new Checker()` 

Meaning, when Checker initializes, it creates a NEW instance of `SuppressWithNearbyCommentFilter`
Essentially, this makes the `cachedFileAbsolutePath` field become to `""`

But the mutation itself related to caching. Meaning, we need the `same object instance` to survive across two different events. But as shown, the `verify` method is ensuring a clean test everytime it runs. 

What the flow of verifyWithSuppressedWithParser is like:

`Create Filter_Instance_A. -> Process File.java -> Filter_Instance_A dies. -> (Next test) Create Filter_Instance_B. -> Process File.java -> Filter_Instance_B dies.`

config/pitest-suppressions/pitest-filters-suppressions.xml
aclfe: I checked the code, this mutation gives equivalent behaviour

https://github.com/checkstyle/checkstyle/blob/9b744eeab7371801c47bfa6fa9a607befe61f30e/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java#L125

The fact that it survived means that code didn't crash, so we can change the redundant code and just removed the initialisation.

romani:
can this be a solution ?
https://github.com/checkstyle/checkstyle/pull/18643/changes#r2701901280

aclfe:
> can this be a solution ? https://github.com/checkstyle/checkstyle/pull/18643/changes#r2701901280

Thank you for pointing this out, let me investigate this! I'll try something else I have in mind too

romani:
@aclfe , main hint from me was that we can consider this pitest mutation as performance optimization and we already know that pitest do not like performance optimizations that are not covered by tests. Covering by tests performance optimization is very very unstable. So we can make trick, we will make some util method, name it very utilitary like we did with `lazyLoad` and we can exclude it in pitest.

aclfe:
> @aclfe , main hint from me was that we can consider this pitest mutation as performance optimization and we already know that pitest do not like performance optimizations that are not covered by tests. Covering by tests performance optimization is very very unstable. So we can make trick, we will make some util method, name it very utilitary like we did with `lazyLoad` and we can exclude it in pitest.

I tried another method. I refactored the old code a little bit and ran a local diff report. I fixed some "dead code" if you many.  There's no new violations and all the mutations are killed. 

My only problem is I forgot to update my local master, so I'm currently facing punishment from the CI

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java
```diff
@@ -411,19 +418,14 @@ private static int parseInfluence(String format, String influenceFormat, String
...
-            }
-            final Tag tag = (Tag) other;
-            return Objects.equals(firstLine, tag.firstLine)
-                    && Objects.equals(lastLine, tag.lastLine)
-                    && Objects.equals(text, tag.text)
-                    && Objects.equals(tagCheckRegexp, tag.tagCheckRegexp)
-                    && Objects.equals(tagMessageRegexp, tag.tagMessageRegexp)
-                    && Objects.equals(tagIdRegexp, tag.tagIdRegexp);
+            return other != null
+                    && getClass() == other.getClass()
```
romani: What is reason of this method update?

aclfe:
@romani 

Why the discrepency? I see 100% coverage in my branch

image

For example this mutation:

image

CI tells to put mutation in supressions list. That same one is killed here 

image

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java
aclfe: The `(this == other)` early return is a performance optimization that created an untestable mutation point. The mutation "replaced boolean return with true" could not be killed with reasonable test coverage, so I refactored it to prioritize testability over optimization. I hope this approach is okay?

romani: We have special tool, that fully testing `equals`. It is not business logic code, it is purely java tech.

https://github.com/jqno/equalsverifier

We already use it, lets keep using it to cover all edges on this technical method.

aclfe: thank you for the guidance! I wasn't aware we were relying fully on EqualsVerifier for this case, so I really appreciate you pointing that out. I'll revert the changes to keep the standard optimization

romani:
It means that it is luck of your environment why it was killed

aclfe:
> It means that it is luck of your environment why it was killed

It was actually my mistake... I checked the CI job, it uses `mvn -Ppitest-filters` but I was working with older mutation commands. I'll refine my approach

aclfe:
> @aclfe , main hint from me was that we can consider this pitest mutation as performance optimization and we already know that pitest do not like performance optimizations that are not covered by tests. Covering by tests performance optimization is very very unstable. So we can make trick, we will make some util method, name it very utilitary like we did with `lazyLoad` and we can exclude it in pitest.

Romani after many attempts, it seems your suggestion is the best option. I was originally trying to solve this without excluding, but the mutation is just too unstable to kill via tests. I think we should go with the utility method approach, I'll implement the utility method

romani:
Method name might be something like `content = initIfDifferent(currentContent, content, () -> someinitfunction)`

aclfe:
I also went and made some more changes. Let me elaborate on them

src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilter.java
```diff
@@ -119,7 +123,7 @@ public class SuppressWithNearbyCommentFilter
      * are reassigned to the next FileContents, at which time filtering for
      * the current FileContents is finished.
      */
-    private WeakReference<FileContents> fileContentsReference = new WeakReference<>(null);
+    private WeakReference<FileContents> fileContentsReference;
```
aclfe, line 126: I want to elaborate on this change

I looked at `SuppressionCommentFilterTest`

https://github.com/checkstyle/checkstyle/blob/8a63a3141b73f879d619bbc5c04659277b8eb2f3/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressionCommentFilter.java#L147

 it had the same mutation. I looked at the test that killed this mutation. 

https://github.com/checkstyle/checkstyle/blob/8a63a3141b73f879d619bbc5c04659277b8eb2f3/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressionCommentFilterTest.java#L436-L443

Previous convention can't seem to kill this performance optimization with verifyxxx method.  

Futhermore, this value always becomes non-null. it seems to better to just remove the initialisation rather than suppress it

pom.xml
```diff
@@ -5139,13 +5139,21 @@
               </targetTests>
               <avoidCallsTo>
                 <avoidCallsTo>com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil</avoidCallsTo>
+                <!-- Since https://github.com/checkstyle/checkstyle/pull/18475 -->
+                <!-- getFileName() calls in accept() are performance optimization only. -->
```
aclfe: `getFileName()` is another performance optimisation. It might be better to just exclude it rather than try to kill it with a performance test. This can't be killed by verifyxxx methods. I investigated the pit reports, this is how existing tests in other files are killing it:

https://github.com/checkstyle/checkstyle/blob/e9ca4c80d0475ec700a4ca7592dcb4e1a41d7061/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyTextFilter.java#L163

Was killed by 

https://github.com/checkstyle/checkstyle/blob/e9ca4c80d0475ec700a4ca7592dcb4e1a41d7061/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyTextFilterTest.java#L521-L530

---------------------
Another example:
Same happened with these

https://github.com/checkstyle/checkstyle/blob/e9ca4c80d0475ec700a4ca7592dcb4e1a41d7061/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java#L175

killed by:

https://github.com/checkstyle/checkstyle/blob/e9ca4c80d0475ec700a4ca7592dcb4e1a41d7061/src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilterTest.java#L729-L738

pom.xml
```diff
@@ -5139,13 +5139,21 @@
...
+                <avoidCallsTo>com.puppycrawl.tools.checkstyle.api.FileContents</avoidCallsTo>
```
romani: this update is not expected, please explain.

romani: ok , it is here https://github.com/checkstyle/checkstyle/pull/18475#discussion_r2705255572

aclfe: Yes

romani:
Api classes are classes that we don't know how they are used, we can keep pure unit tests for them, please remove exclude.

aclfe:
> Api classes are classes that we don't know how they are used, we can keep pure unit tests for them, please remove exclude.

Okay got it! I'll add a unit test then

aclfe:
@romani 

The changes are pushed, please let me know if they need any correction

romani:
unfortunately we got concurrent PRs for same update.

https://github.com/checkstyle/checkstyle/pull/18643 was slightly more cleaner solution.

@aclfe , but I deeply appreciate your time and diligence on this issue, and hope to continue collaboration with you on other issues, I hope it is last time we run in concurrent.

aclfe:
@romani
No worries. I still learned a lot along the way with this PR and that's what matters.
Cheers to more PRs!

Issue #16360, opened 2025-02-18
XMLLoggerTest.java to use verifyWithInlineConfigParserAndXmlLogger


romani:
After merge of https://github.com/checkstyle/checkstyle/pull/16314

We have ability to update all/most test methods in [XMLLoggerTest.java](https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java) to use `verifyWithInlineConfigParserAndXmlLogger`

Goal: use it as much as possible, ideally in all test methods.

Good example of expected update https://github.com/checkstyle/checkstyle/pull/16701/files

roulpriya:
@romani, I am working on this issue. 
I see three potential test methods in `XMLLoggerTest.java` that could use `verifyWithInlineConfigParserAndXmlLogger`
- `testAddError`: Adding any error to the source
- `testAddIgnored`: Adding a test which is ignored
- `testAddErrorOnZeroColumns`: Adding a test that fails on the entire line like `EmptyLineSeparatorCheck`

I have raised a draft PR fixing `testAddError`. I wanted to validate if I understand this correctly.

romani:
Your draft PR is in right direction.
Please do all test methods migration in separate PRs.

roulpriya:
I am not able to get this error: https://app.circleci.com/pipelines/github/checkstyle/checkstyle/32027/workflows/2d0479fd-798e-4a53-b962-2c0ed35f6bb3/jobs/862027/parallel-runs/0/steps/0-103?invite=true#step-103-263616_54

This says `Could not find artifact org.xwiki.commons:xwiki-commons`

Can you help me with this?

roulpriya:
In the `XMLLoggerTest.testAddIgnored()` case, which checkstyle check should I use? Do I have to add suppression to ignore the checks?

romani:
You need to use some Check in xml config that use severity ignore 
https://github.com/checkstyle/checkstyle/blob/359fa919d6b96ea334346459b77b7e54931dee65/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java#L256

gianmarcoschifone:
@romani I'm not sure on how to update the `testAddErrorWithNullFileName()` since it would require an `AuditEvent` with null `fileName`. Could you please guide me?

romani:
>since it would require an AuditEvent with null fileName. Could you please guide me?

[...]

PR #18485, opened 2026-01-02
Issue #16360: Added comment to testAddExceptionBetweenFileStartedAndF…

VinishaWagh:
Issue : #16360 
This test cannot use verifyWithInlineConfigParserAndXmlLogger because it injects a custom exception and verifies XMLLogger’s exception-handling and OutputStream lifecycle behavior, which cannot be reproduced through Checker execution.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
@@ -291,13 +291,13 @@ public class XMLLoggerTest extends AbstractXmlTestSupport {
         final XMLLogger logger = new XMLLogger(outStream, OutputStreamOptions.CLOSE);
         logger.auditStarted(null);
 
-        final AuditEvent fileStartedEvent = new AuditEvent(this, "Test.java");
+        final AuditEvent fileStartedEvent = new AuditEvent(this, "InputXMLLoggerAddExceptionBetweenFileStartedAndFinished.java");
         logger.fileStarted(fileStartedEvent);
 
         final Violation violation =
                 new Violation(1, 1,
                         "messages.properties", null, null, null, getClass(), null);
-        final AuditEvent ev = new AuditEvent(this, "Test.java", violation);
+        final AuditEvent ev = new AuditEvent(this, "InputXMLLoggerAddExceptionBetweenFileStartedAndFinished.java", violation);
         logger.addException(ev, new TestException("msg", new RuntimeException("msg")));
 
         logger.fileFinished(ev);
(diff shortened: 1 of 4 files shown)


VinishaWagh:
@romani can you please help with this ``[no-error testing / no-error-trino (pull_request)](https://github.com/checkstyle/checkstyle/actions/runs/20655771156/job/59308320751?pr=18485)`` failing test?

0xChaitanya:
@VinishaWagh,  You can try running the checks again, as it shows a server error. To do this, close the PR and then reopen it.

src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
```diff
@@ -327,6 +327,12 @@ public void testAddExceptionBeforeFileFinished()
...
     }
 
+    /**
+     * Cannot use verifyWithInlineConfigParserAndXmlLogger because this test
+     * injects a custom exception and verifies XMLLogger exception-handling
+     * and OutputStream close() behavior, which cannot be reproduced
+     * through Checker execution.
+     */
     @Test
     public void testAddExceptionBetweenFileStartedAndFinished()
```
romani: not good excuse.

please create method like https://github.com/VinishaWagh/checkstyle/blob/0cece28dff740f95fc657199b9d9ea5adcf8af00/src/test/java/com/puppycrawl/tools/checkstyle/AbstractXmlTestSupport.java#L278
overloaded and just extra parameter to take  `outStream` from test method.
So you will run all naturally and check that outStream is closed in text method.

romani (changes requested):
items:

src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
VinishaWagh: I pushed the changes to show the new overloaded `verifyWithInlineConfigParserAndXmlLogger` and the converted tests.
CI is currently failing only for the `testAddException*` cases, where XMLLogger’s `<exception>` output is expected.
These tests inject custom exceptions via `addException`, which cannot be reproduced through real Checker execution, so inline verification results in <error> instead.
I wanted to confirm whether these exception-handling tests should remain manual with an explanatory comment.

VinishaWagh: @romani 
I’ve investigated why inline execution fails for invalid Java input.
When a Java parser exception occurs, `Checker.process()` aborts before firing `fileFinished` and `auditFinished`, which leaves `XMLLogger` with incomplete XML. Catching the exception in tests or helpers is too late, since the audit lifecycle is already terminated internally.
Since we can’t modify `Checker`, could you please clarify whether the intended scenario is:

an exception during Java parsing, or an exception during check execution (with valid Java input)?

romani:
Please make CI green.  see videos on how to investigate CI problems

romani:
@yn-mi12 , can you finish this PR ? it looks like abandoned but has some  potentially valuable updates

yn-mi12:
Sorry for the late response.

The `testAddExceptionBetweenFileStartedAndFinished` is not currently supported with the `verifyWithInlineConfigParserAndXmlLogger()` method. Neither this method nor the `Checker.java` class provides a mechanism to handle `addException()`. 

The approach this pull request proposes is to introduce a compile-time error instead - a missing semicolon. I believe it is not a viable solution, as it cannot be properly processed by Checkstyle and cannot bypass the existing tests.

I believe this pull request should be closed as issue #16360 has been closed.
Is this the right thing to do?

Issue #18482, opened 2026-01-02
Checker framework report violation on theoretcally possible cases but practiacally impossible


romani:
it was around for long and made very clear at
https://github.com/checkstyle/checkstyle/pull/18451#discussion_r2651444194

if we consider Class methods wihtout context of execution, Checker is absolutely right that
```
    private static Set<String> getRecordComponentNames(DetailAST recordDef) {
        final Set<String> names = new HashSet<>();
        final DetailAST recordComponents = recordDef.findFirstToken(TokenTypes.RECORD_COMPONENTS);
        DetailAST child = recordComponents.getFirstChild();
        while (child != null) {
            if (child.getType() == TokenTypes.RECORD_COMPONENT_DEF) {
                names.add(child.findFirstToken(TokenTypes.IDENT).getText());
            }
            child = child.getNextSibling();
        }
        return names;
    }
```

has null potential probelm:
```
  <checkerFrameworkError unstable="false">
    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/MissingOverrideOnRecordAccessorCheck.java</fileName>
    <specifier>dereference.of.nullable</specifier>
    <message>dereference of possibly-null reference recordComponents</message>
    <lineContent>DetailAST child = recordComponents.getFirstChild();</lineContent>
  </checkerFrameworkError>
```

but due to limitations of Checkstyle, that we execute only on compiled sources only, this becomes not possible.
So we can not satisfy Checker, as it will produce dead code that will conflict with jacoco/pitest that we vale more that Checker.

Checker propose to disable such violations in code https://checkerframework.org/manual/#suppressing-warnings-nullness:
> The Checker Framework supplies several ways to suppress warnings, most notably the @SuppressWarnings("nullness") annotation (see Chapter ‍[33](https://checkerframework.org/manual/#suppressing-warnings)). An example use is
```
    // might return null
    @Nullable Object getObject(...) { ... }
    void myMethod() {
      @SuppressWarnings("nullness") // with argument x, getObject always returns a non-null value
      @NonNull Object o2 = getObject(x);
    }
```

This approach is very verbose and can pollute code, so we unlikely to use it, especially in middle of method body.

or special util method
https://checkerframework.org/manual/#suppressing-warnings-with-assertions
this might be interesting compromise to split cases where null is not possible(due to context that AST is of compiled sources) and cases where search for token can actaully return nothing/null.

```
[...]

PR #18491, opened 2026-01-03
Issue #18482: Add NullnessAssertUtil and apply to UncommentedMainChec…

vivek-0509:
Issue #18482: 

This PR introduces a proof-of-concept for addressing Checker Framework `dereference.of.nullable` violations on AST nodes that are guaranteed non-null by Java grammar.

### Problem

Checker Framework flags potential null dereferences on AST methods like `findFirstToken()` and `getFirstChild()`. However, in Checkstyle's context (processing compiled Java sources), certain tokens are guaranteed to exist by the Java Language Specification. Adding explicit null checks would:
- Create unreachable code (affecting JaCoCo coverage)
- Produce not killable Pitest mutations
- Add unnecessary noise to the codebase

### Solution

Added NullnessAssertUtil utility class with a notNull() method that:
- Throws `AssertionError` at runtime if null is passed (catches misuse during testing)
- Provides clear intent: "this value is grammar-guaranteed non-null"

### How It Looks

```java
// Before :
final DetailAST ident = method.findFirstToken(TokenTypes.IDENT);

// After :
final DetailAST ident = NullnessAssertUtil.notNull(method.findFirstToken(TokenTypes.IDENT));

### CODE REVIEW

**Comment by vivek-0509** [comment]
Hi @romani,

The POC introduces `NullnessAssertUtil.notNull()` for grammar-guaranteed non-null AST nodes. Pitest's `ArgumentPropagationMutator` is generating surviving mutations for these calls because:

- **Mutation:** Replaces notNull(x) with just `x`
- **Why it survives:** Since `x` is never null in practice (grammar-guaranteed), both versions behave identically

This is similar to how `UnmodifiableCollectionUtil` is already excluded from Pitest via `avoidCallsTo`. 
https://github.com/checkstyle/checkstyle/blob/a86af83a1ce881529fa9a953f12b50bcbddd8465/pom.xml#L3323-L3326

Both utilities are "safety wrappers" rather than business logic, they don't change program behavior, they just add safety guarantees (immutability enforcement for collections, nullness verification for Checker Framework). Mutating them produces equivalent programs that tests cannot distinguish.

**Question:** Should I add NullnessAssertUtil to `avoidCallsTo` in the affected Pitest profiles (currently `pitest-misc` for `UncommentedMainCheck`)?

**Comment by romani** [comment]
yes, as this is purely technical wrapping to avoid violation from tool that we are investigating, let not deal with pitest for it.

**Comment by vivek-0509** [comment]
Ok, I'll add NullnessAssertUtil to avoidCallsTo for the `pitest-misc` profile for now. As we increase usage of this utility in other parts of the codebase, then we can add it to other affected Pitest profiles as needed.

**Comment by vivek-0509** [comment]
@romani  done, CI failures are unrelated

File: config/import-control.xml [hunk 1]
```diff
@@ -130,6 +130,10 @@
     <file name="ModuleReflectionUtil">
       <allow class="com.puppycrawl.tools.checkstyle.AbstractAutomaticBean"/>
     </file>
+
+    <file name="NullnessAssertUtil">
+      <allow pkg="org.checkerframework.checker.nullness.qual"/>
```
romani: this actually a huge problem...


please send separate PR to remove checkerframework from compile dependencies.

```
$ git diff
diff --git a/pom.xml b/pom.xml
index e34469ae35..6147c7cad4 100644
--- a/pom.xml
+++ b/pom.xml
@@ -466,11 +466,6 @@
         </exclusion>
       </exclusions>
     </dependency>
-    <dependency>
-      <groupId>org.checkerframework</groupId>
-      <artifactId>checker-qual</artifactId>
-      <version>${checkerframework.version}</version>
-    </dependency>
     <dependency>
       <groupId>org.apache.maven.doxia</groupId>
       <artifactId>doxia-core</artifactId>
diff --git a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
index 320312f196..72ddffbd88 100644
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
@@ -28,8 +28,6 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import org.checkerframework.checker.index.qual.IndexOrLow;
-
@@ -485,7 +483,7 @@ public class UnusedImportsCheck extends AbstractCheck {
      * @return -1 if parentheses are unbalanced, 0 if no method is found,
      *         or the index of the first space outside parentheses.
      */
-    private static @IndexOrLow("#1")int extractReferencePart(String input) {
+    private static int extractReferencePart(String input) {
         int parenthesesCount = 0;
         int firstSpaceOutsideParens = -1;
         for (int index = 0; index < input.length(); index++) {
```

romani: is there a way to make implementation of NullnessAssertUtil wihtout compile dependency to checkerframework ?

may be we can create issue on them and ask how to avoid dependency but still use workaround

vivek-0509: Looking into this.

romani: @vivek-0509, please send PR to remove library from compilation dependencies first.
We can disable rule that we avoided by annotation.

vivek-0509: Sure

romani: We need step back before we do leap forward

src/main/java/com/puppycrawl/tools/checkstyle/utils/NullnessAssertUtil.java
```diff
@@ -0,0 +1,60 @@
...
+     * the Checker Framework and throws {@link AssertionError} if the argument is null
+     * when Java assertions are enabled.
+     *
+     * @param <T> the type of the reference
+     * @param ref a reference of @Nullable type, that is non-null at run time
+     * @return the argument, cast to have the type qualifier {@code @NonNull}
+     * @throws AssertionError if ref is null and assertions are enabled
+     */
+    @EnsuresNonNull("#1")
+    public static <T> @NonNull T notNull(@Nullable T ref) {
```
romani, line 52: Is it going to work if we remove annotations or use general available annotations?

vivek-0509: Yes, this will work, I tested that the Checker Framework recognizes javax.annotation.Nonnull and javax.annotation.Nullable from JSR-305, We can remove @EnsuresNonNull("#1") and use below imports for the other two  annotations.
```
import javax.annotation.Nonnull;
import javax.annotation.Nullable;
```

romani: Ok, lets proceed here with javax usage, no extra depending on Checker.

vivek-0509: Sure

vivek-0509: done

vivek-0509: Hi @romani,

I noticed that the javax.annotation annotations (@Nonnull, @Nullable from JSR-305) are not TYPE_USE annotations. The CI is throwing this error with the Eclipse compiler:

```
ERROR in NullnessAssertUtil.java (at line 52)
    public static <T> @Nonnull T notNull(@Nullable T ref) {
                      ^^^^^^^^
Annotation types that do not specify explicit target element types cannot be applied here
```
The JSR-305 annotations can only be applied to methods, fields, and parameters, not to type positions like `@Nonnull T.`

i am trying JSpecify (org.jspecify.annotations.NonNull, org.jspecify.annotations.Nullable) which is already available in Checkstyle via Guava, no new dependency needed.

src/main/java/com/puppycrawl/tools/checkstyle/checks/UncommentedMainCheck.java
```diff
@@ -147,7 +148,8 @@ private void visitClassOrRecordDef(DetailAST classOrRecordDef) {
         // we are not use inner classes because they can not
         // have static methods
         if (classDepth == 0) {
-            final DetailAST ident = classOrRecordDef.findFirstToken(TokenTypes.IDENT);
+            final DetailAST ident =
+                    NullnessAssertUtil.notNull(classOrRecordDef.findFirstToken(TokenTypes.IDENT));
```
romani: Please shorten name of util class to NullUtil. It will be very very widely used, and it has no logic meaning, very technical, so better to be less visible during reading.

vivek-0509: done

PR #18510, opened 2026-01-05
Issue #6294: Added test to kill mutation in AbstractExpressionHandler.getFirstToken

vivek-0509:
Issue #6294:

This PR adds a test case to kill the Pitest mutation in AbstractExpressionHandler.getFirstToken() method.
The mutation changes which token is selected when multiple tokens share the same column number. When a generic type is line-wrapped.
Both original and mutated code produce violations on the same line/column, but the error message token name differs, making the mutation killable.


vivek-0509:

vivek-0509:
CI failures are unrelated

vivek-0509:
Investigating the differences

PR #18511, opened 2026-01-05
Issue #6294: Refactor column comparison in getFirstAstNode in Abstrac…

vivek-0509:
Issue #6294:

Replaced conditional boundary comparisons with Math.min() in getFirstAstNode() method. This refactoring eliminates equivalent ConditionalsBoundaryMutator mutations that could not be killed. Using Math.min() changes the mutation type to EqualityMutator which is killable.


vivek-0509:

vivek-0509:
CI failures are unrelated

PR #18529, opened 2026-01-06
Pull #18529: Fix wrong test scope for `PMD`

Pankraz76:

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -473,7 +473,8 @@ spotbugs-and-pmd)
   mkdir -p .ci-temp/spotbugs-and-pmd
   CHECKSTYLE_DIR=$(pwd)
   export MAVEN_OPTS='-Xmx2g'
-  ./mvnw -e --no-transfer-progress clean test-compile pmd:check spotbugs:check
+  ./mvnw -e --no-transfer-progress clean pmd:check
+  ./mvnw -e --no-transfer-progress clean test-compile spotbugs:check
   cd .ci-temp/spotbugs-and-pmd
   grep "Processing_Errors" "$CHECKSTYLE_DIR/target/site/pmd.html" | cat > errors.log
   RESULT=$(cat errors.log | wc -l)
--- a/config/pmd.xml
+++ b/config/pmd.xml
@@ -306,6 +306,7 @@
     <exclude name="DataClass"/>
     <!-- Too many alarms on the check classes, we will never move logic out of the check,
          each check is an independent logic container. -->
+    <exclude name="ExceptionAsFlowControl"/>
     <exclude name="GodClass"/>
     <!-- Too many violations, will be suppressed until we find out how use these metrics. -->
     <exclude name="LawOfDemeter"/>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/DetailNodeTreeStringPrinter.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/DetailNodeTreeStringPrinter.java
@@ -21,6 +21,7 @@ package com.puppycrawl.tools.checkstyle;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.charset.Charset;
 
 import com.puppycrawl.tools.checkstyle.JavadocDetailNodeParser.ParseErrorMessage;
 import com.puppycrawl.tools.checkstyle.JavadocDetailNodeParser.ParseStatus;
@@ -69,17 +70,6 @@ public final class DetailNodeTreeStringPrinter {
             throw new IllegalArgumentException(getParseErrorMessage(status.getParseErrorMessage()));
         }
         return status.getTree();
-    }
-
-    /**
-     * Parse javadoc comment to DetailNode tree.
-     *
-     * @param javadocComment javadoc comment content
-     * @return tree
-     */
-    private static DetailNode parseJavadocAsDetailNode(String javadocComment) {
-        final DetailAST blockComment = ParserUtil.createBlockCommentNode(javadocComment);
-        return parseJavadocAsDetailNode(blockComment);
     }
 
     /**
@@ -168,8 +158,9 @@ public final class DetailNodeTreeStringPrinter {
      * @throws IOException if the file could not be read.
      */
     private static DetailNode parseFile(File file) throws IOException {
-        final FileText text = new FileText(file, System.getProperty("file.encoding"));
-        return parseJavadocAsDetailNode(text.getFullText().toString());
+        final FileText text = new FileText(file, Charset.defaultCharset().name());
+        final DetailAST comment = ParserUtil.createBlockCommentNode(text.getFullText().toString());
+        return parseJavadocAsDetailNode(comment);
     }
 
 }


.circleci/config.yml
```diff
@@ -201,9 +201,13 @@ workflows:
...
       - validate-with-maven-script:
-          name: "spotbugs-and-pmd"
+          name: spotbugs
           image-name: *cs_img
-          command: "./.ci/validation.sh spotbugs-and-pmd"
+          command: "./.ci/validation.sh spotbugs"
+      - validate-with-maven-script:
+          name: pmd
+          image-name: *cs_img
+          command: "./.ci/validation.sh pmd"
```
Pankraz76: code actually not passing PMD due to:

- https://github.com/checkstyle/checkstyle/pull/18530

.ci/validation.sh
```diff
@@ -1330,7 +1313,7 @@ website-only)
   ;;
 
 pmd)
-  ./mvnw -e --no-transfer-progress clean test-compile pmd:check
```
Pankraz76, line 1333: This seems wrong limiting scope to test only.

.ci/validation.sh
```diff
@@ -469,23 +469,6 @@ checkstyle-and-sevntu)
     -Dpmd.skip=true -Dspotbugs.skip=true -Djacoco.skip=true
   ;;
 
-spotbugs-and-pmd)
-  mkdir -p .ci-temp/spotbugs-and-pmd
-  CHECKSTYLE_DIR=$(pwd)
-  export MAVEN_OPTS='-Xmx2g'
-  ./mvnw -e --no-transfer-progress clean test-compile pmd:check spotbugs:check
```
Pankraz76, line 476: this seems not to work as expected.

Pankraz76:
reopen after enabler:

- https://github.com/checkstyle/checkstyle/pull/18530

Pankraz76:
image


@romani 

please execute PMD locally the correct way!!


RIGHT!!!

./mvnw -e --no-transfer-progress clean pmd:check


WRONG !!!

./mvnw -e --no-transfer-progress clean test-compile pmd:check

romani:
New pmd violations:
`[[1;33mWARNING[m] PMD Failure: com.puppycrawl.tools.checkstyle.AbstractAutomaticBean:243 Rule:ExceptionAsFlowControl Priority:3 Exception thrown at line 237 is caught in this block..`

`PMD Failure: com.puppycrawl.tools.checkstyle.DetailNodeTreeStringPrinter:80 Rule:UnusedPrivateMethod Priority:3 Avoid unused private methods such as 'parseJavadocAsDetailNode(String)'..`

romani:
CI job should stay as is we can do two mvn commands under ci command  to run spotbugs and pmd separately

Pankraz76:
i dont think so. Why having this coupled for no reason?

please conaider dedication and seperation principle to see each job upfront like we for other SCA tool as well:

- https://github.com/checkstyle/checkstyle/pull

Pankraz76:
> CI job should stay as is we can do two mvn commands under ci command to run spotbugs and pmd separately

yes kind of to fix now. but why having this coupling?

Pankraz76:
item:



Command: `./mvnw -e --no-transfer-progress clean pmd:check`

New pmd violations:
`[[1;33mWARNING[m] PMD Failure: com.puppycrawl.tools.checkstyle.AbstractAutomaticBean:243 Rule:ExceptionAsFlowControl Priority:3 Exception thrown at line 237 is caught in this block..`

`PMD Failure: com.puppycrawl.tools.checkstyle.DetailNodeTreeStringPrinter:80 Rule:UnusedPrivateMethod Priority:3 Avoid unused private methods such as 'parseJavadocAsDetailNode(String)'..`

Pankraz76:
> CI job should stay as is we can do two mvn commands under ci command to run spotbugs and pmd separately

yes now its your style. actually its just the test scope issue.

.ci/validation.sh
Pankraz76: RCA

Pankraz76:
item

```
01:58
[INFO] --- compiler:3.14.1:compile (default-compile) @ equalsverifier-testhelpers ---01:58
[INFO] Recompiling the module because of changed source code.01:59
[INFO] Compiling 83 source files with javac [forked debug release 17] to target/classes01:59
[INFO] -------------------------------------------------------------01:59
[ERROR] COMPILATION ERROR : 01:59
[INFO] -------------------------------------------------------------01:59
[ERROR] 	at jdk.compiler/com.sun.tools.javac.api.BasicJavacTask.initPlugins(BasicJavacTask.java:232)01:59
	at jdk.compiler/com.sun.tools.javac.main.Main.compile(Main.java:294)01:59
	at jdk.compiler/com.sun.tools.javac.main.Main.compile(Main.java:178)01:59
	at jdk.compiler/com.sun.tools.javac.Main.compile(Main.java:64)01:59
	at jdk.compiler/com.sun.tools.javac.Main.main(Main.java:50)01:59
Caused by: com.google.errorprone.InvalidCommandLineOptionException: -XDaddTypeAnnotationsToSymbol=true is required by Error Prone on JDK 2101:59
	at com.google.errorprone.BaseErrorProneJavaCompiler.checkAddTypeAnnotationsToSymbol(BaseErrorProneJavaCompiler.java:243)01:59
	at com.google.errorprone.BaseErrorProneJavaCompiler.addTaskListener(BaseErrorProneJavaCompiler.java:91)01:59
	at com.google.errorprone.ErrorProneJavacPlugin.init(ErrorProneJavacPlugin.java:34)01:59
	at jdk.compiler/com.sun.tools.javac.api.BasicJavacTask.initPlugin(BasicJavacTask.java:256)01:59
	at jdk.compiler/com.sun.tools.javac.api.BasicJavacTask.initPlugins(BasicJavacTask.java:230)01:59
	... 4 more01:59
[INFO] 1 error01:59
[INFO] ----------------------------------
```

.ci/validation.sh
```diff
@@ -1329,12 +1329,8 @@ website-only)
   ./mvnw -e --no-transfer-progress clean site -Pno-validations
   ;;
 
-pmd)
```
romani, line 1332: Keep it, it is not used in CI, it can do a command that you think is right to execute it correctly

config/pmd.xml
```diff
@@ -306,6 +306,7 @@
     <exclude name="DataClass"/>
     <!-- Too many alarms on the check classes, we will never move logic out of the check,
          each check is an independent logic container. -->
+    <exclude name="ExceptionAsFlowControl"/>
```
romani, line 309: You might share it before.
Please share one more time violation that we suppressing I need to double check

.ci/validation.sh
Pankraz76: > not used

we delete all unused, why keep this?

> it can do a command that you think is right to execute it correctly

understand but makes no sense to me. 

we dont need this, its providing no benefit only cost.

config/pmd.xml
Pankraz76: you also found it

```
[�[1;33mWARNING�[m] PMD Failure: com.puppycrawl.tools.checkstyle.
AbstractAutomaticBean:243 Rule:ExceptionAsFlowControl Priority:3 
Exception thrown at line 237 is caught in this block..
```

lets just go. in PMD we trust.

Pankraz76:
rdy 2 merge.

Pankraz76:
lets fix and pick this low hanging fruit @romani

config/pmd.xml
romani: yes, this rule is not working well, ok to disable

.ci/validation.sh
romani: they are in use by users on local, hard to remember how to run stuff properly.
and can be reused by CI jobs eventually.

romani: please restore

romani:
please do https://github.com/checkstyle/checkstyle/pull/18529#discussion_r2701905926

config/pmd.xml
Pankraz76: no the rule is working fine kind of, but its the way its done in java since day one.

isnt´t it? @adangel

.ci/validation.sh
Pankraz76: kindly allow 1 extra PR count 6 but actually 5 as this now it ready to merge we have agreed upon, 

lets find the reason why we need to unused stuff and delete it like as **ALWAYS** do.

Pankraz76: - https://github.com/checkstyle/checkstyle/pull/18682

.ci/validation.sh
```diff
@@ -473,7 +473,7 @@ spotbugs-and-pmd)
   mkdir -p .ci-temp/spotbugs-and-pmd
   CHECKSTYLE_DIR=$(pwd)
   export MAVEN_OPTS='-Xmx2g'
-  ./mvnw -e --no-transfer-progress clean test-compile pmd:check spotbugs:check
+  ./mvnw -e --no-transfer-progress clean pmd:check spotbugs:check
```
adangel: This call of pmd and spotbugs most likely won't work correctly now: PMD and spotbugs require that the project under analysis is compiled. These tools need access to the bytecode (aka. auxclasspath). Hence there was "test-compile" added before...
Without this, these tools could produce false positive and/or negatives.

config/pmd.xml
adangel: I don't know what you mean with "its done in java since day one"...

So, PMD claims that https://github.com/checkstyle/checkstyle/blob/b8d7e3056acfccba40f01ae56d5a13824526fd20/src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java#L243 catches the exception thrown at https://github.com/checkstyle/checkstyle/blob/b8d7e3056acfccba40f01ae56d5a13824526fd20/src/main/java/com/puppycrawl/tools/checkstyle/AbstractAutomaticBean.java#L237 . If that is true, then this would be indeed an instance of [ExceptionAsFlowControl](https://docs.pmd-code.org/latest/pmd_rules_java_design.html#exceptionasflowcontrol).

But this doesn't look to be the case: https://github.com/checkstyle/checkstyle/blob/b8d7e3056acfccba40f01ae56d5a13824526fd20/src/main/java/com/puppycrawl/tools/checkstyle/api/CheckstyleException.java#L31 is just an Exception and not related to any of "InvocationTargetException | IllegalAccessException | NoSuchMethodException".

This leads to the conclusion: Either PMD is wrong or PMD is not called correctly. I suspect that latter, as figuring out the relation of types requires a correct auxclasspath, see the comment above in "validation.sh".

.ci/validation.sh
Pankraz76: with this the tool did not worked properly ether, so its kind of strange, as the copse was limited to test.

rewrite is able to call the compile phase itself is this something PMD can adjust as well? this is kind of error prone.

config/pmd.xml
Pankraz76: > I don't know what you mean with "its done in java since day one"...

Being in the industry for 15 years, i learned from day one Exception As Flow Control is bad, still it was done every since in all the codebases around.

So PMD is correctly working, yes.

Pankraz76: if this is so fragile then better extract the tool calls to avoid any unwanted interconnection. 

actually consider the comment it feels strange that PMD and spot are connected even the commen stated it needs to run alone.

coupling is always bad software design so lets avoid in CI as well.

- https://github.com/checkstyle/checkstyle/blob/b8d7e3056acfccba40f01ae56d5a13824526fd20/.circleci/config.yml#L205

.ci/validation.sh
Pankraz76: if this is important then we should call both compile goals, which would hopefully also fix the scope issue.

Pankraz76:
lets merge. Looking into the facts PMD is working fine, as without the change its definitely not good working not at all.

.ci/validation.sh
Pankraz76: > This call of pmd and spotbugs most likely won't work correctly now: PMD and spotbugs require that the project under analysis is compiled. These tools need access to the bytecode (aka. auxclasspath). Hence there was "test-compile" added before... Without this, these tools could produce false positive and/or negatives.

idk but actually adding the compile phase is in both cases wrong having tried it only the one WITHOUT is failing like expected to call PMD just like rewrite `tool:goal` without any prefix goal imposing unwanted coupling and config burden to user;


image

.ci/validation.sh
```diff
@@ -473,7 +473,8 @@ spotbugs-and-pmd)
   mkdir -p .ci-temp/spotbugs-and-pmd
   CHECKSTYLE_DIR=$(pwd)
   export MAVEN_OPTS='-Xmx2g'
-  ./mvnw -e --no-transfer-progress clean test-compile pmd:check spotbugs:check
+  ./mvnw -e --no-transfer-progress clean pmd:check
+  ./mvnw -e --no-transfer-progress clean spotbugs:check
```
Pankraz76: only spot bugs need this ....

Pankraz76:
> I am ok to run tools separately.

then suggest to fully split. not only in some command.

romani:
No , same CI Job, two tasks in it.
Container starting time is longer than action execution.

PR #18538, opened 2026-01-06
Issue #6294: Refactor CommentsIndentationCheck to use single method f…

vivek-0509:
Issue #6294:

Removed isUsingOfObjectReferenceToInvokeMethod and findTokenWhichBeginsTheLine methods.
findStartTokenOfMethodCallChain handles all cases correctly, making conditional logic unnecessary.

### DIFF
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -5191,17 +5191,6 @@
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/CommentsIndentationCheck.java</fileName>
-    <specifier>return</specifier>
-    <message>incompatible types in return.</message>
-    <lineContent>return tokenWhichBeginsTheLine;</lineContent>
-    <details>
-      type of expression: @Initialized @Nullable DetailAST
-      method return type: @Initialized @NonNull DetailAST
-    </details>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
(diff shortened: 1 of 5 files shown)


vivek-0509:

vivek-0509:

vivek-0509:

vivek-0509:
This refactoring removes the `isUsingOfObjectReferenceToInvokeMethod `conditional logic and the `findTokenWhichBeginsTheLinemethod `from `CommentsIndentationCheck.java`. The original implementation used two different algorithms to find the token that begins a line: `findFirstToken(TokenTypes.IDENT)` for `simple method calls` like doSomething(), and `findStartTokenOfMethodCallChain` for `chained method calls` like obj.toString(). The condition `isUsingOfObjectReferenceToInvokeMethod` was used to determine which algorithm to apply by checking the depth of the AST structure.

Through debug tracing, I verified that this conditional logic is unnecessary because `findStartTokenOfMethodCallChain`
 correctly handles both cases. For simple method calls like doSomething(), the algorithm descends from EXPR through METHOD_CALL and stops at the IDENT node (which has no children), returning the method identifier. For chained method calls like obj.toString(), the algorithm descends through the DOT node structure and returns the first identifier obj. This algorithm works correctly for both simple and chained method calls.

The refactoring simplifies `findPreviousStatement` by replacing the call to `findTokenWhichBeginsTheLine(root)` with a direct call to `findStartTokenOfMethodCallChain(root)`. This eliminates the need for the intermediate method and its associated conditional check. The` findTokenWhichBeginsTheLine` and `isUsingOfObjectReferenceToInvokeMethod` methods have been completely removed as they are no longer needed.

vivek-0509:
Hey @romani , here is the diff report, no changes after refactoring
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/d7d9779f0359268aef4695cea464a7727132b76d_2026101307/reports/diff/index.html

PR #18541, opened 2026-01-06
Issue #18027: Resolving pit supression valueof

aclfe:
Issue #18027: Resolving pit supression valueof

Killing mutation. 

Confirming if direction is correct

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
@@ -501,6 +501,31 @@ public class CheckstyleAntTaskTest extends AbstractPathTestSupport {
     }
 
     @Test
+    public void testScanPathLogsVerboseMessage() throws IOException {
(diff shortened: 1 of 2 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTaskTest.java
```diff
@@ -1064,6 +1089,28 @@ public void testAntProjectPropertyValueIsCopiedCorrectly() throws IOException {
                 .isEqualTo("ignore");
     }
 
+    private static class LogCapturingProject extends Project {
```
romani: Strange that this class do a lot of logging but it is first time it requires logger project class

aclfe: Thanks for pointing that, I overlooked `CheckstyleAntTaskLogStub` 

I made changes

aclfe:
> Please send few more PRs for cases where mutation in logging.
> It will be easy win.

Yes for sure

PR #18311, opened 2025-12-15
Issue #18265: Added allowMultipleEmptyLinesInsideClassMembers property inside EmptyLineSeparator check in google_checks.xml

Praveen7294:
Issue: #18265

From: https://google.github.io/styleguide/javaguide.html#s4.6.1-vertical-whitespace

- Added a [allowMultipleEmptyLinesInsideClassMembers](https://checkstyle.org/checks/whitespace/emptylineseparator.html#allowMultipleEmptyLinesInsideClassMembers) property inside EmptyLineSeparator check.
- Added test cases `testCorrectEmptyLineSeparatorInsideClassMembers()` and `testIncorrectEmptyLineSeparatorInsideClassMembers()` in `VerticalWhitespaceTest.java`.
- Added input files `InputCorrectEmptyLineInsideClassMembers.java` and `InputIncorrectEmptyLineInsideClassMembers.java` to rule461verticalwhitespace directory.

- From this PR #17950 , `InputEmptyLineSeparator3.java` and `InputFormattedEmptyLineSeparator3.java` were added to the `rule461verticalwhitespace` directory but but they were not covered by any test cases. Therefore, added test cases `testEmptyLineSeparatorBetweenClassMembers()` and `testFormattedEmptyLineSeparatorBetweenClassMembers()` to `VerticalWhitespaceTest.java`.

---
Diff Regression config: https://gist.githubusercontent.com/Praveen7294/605a30d7c12d5fcc633316127e4db0be/raw/c0f9602b094f75a806a296569b5f70a6c3d5e471/base-config-emptyline.xml
Diff Regression patch config: https://gist.githubusercontent.com/Praveen7294/0c2bffd20db9d6c9696c2fce03de4407/raw/f89ddb1333c8bfe4b26e62ba9f3e4327567bfc4e/patch-config-emptyline.xml
Diff Regression projects: https://gist.githubusercontent.com/Praveen7294/a1626a4272b4be9d7a0dc8a9fbae5ffa/raw/22fada1af5a2880572d3e91848996d50427c1616/list-of-projects.properties

PR #18545, opened 2026-01-07
Issue #18438: Fix false positive in EmptyLineSeparator check

### LINKED ISSUE #18438: False positive: EmptyLineSeparator reports violation on method call
reported at https://github.com/checkstyle/checkstyle/pull/18311#issuecomment-3691084557

I have read check documentation: https://checkstyle.org/checks/whitespace/emptylineseparator.html#EmptyLineSeparator

`EmptyLineSeparator` reports violation on method call which wrong. It only occurs when there are more than one empty lines before a method call.

#### Config:
```
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN" "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
  <module name="TreeWalker">
    <module name="EmptyLineSeparator">
      <property name="tokens"
               value="PACKAGE_DEF, IMPORT, STATIC_IMPORT, CLASS_DEF, INTERFACE_DEF, ENUM_DEF,
                    STATIC_INIT, INSTANCE_INIT, METHOD_DEF, CTOR_DEF, VARIABLE_DEF, RECORD_DEF,
                    COMPACT_CTOR_DEF"/>
      <property name="allowNoEmptyLineBetweenFields" value="true"/>
      <property name="allowMultipleEmptyLines" value="false"/>
      <property name="allowMultipleEmptyLinesInsideClassMembers" value="false"/>
    </module>
  </module>
</module>
```

#### Test.java:
```
import org.junit.Test;
import org.junit.Assert;

public class Test {
  
  public void foo(int a) {}

  @Test
  public void testFoo() {

    
    int a = 10;


    Test t = new Test();

    
    t.foo(10);
    int b = 20;


    Assert.assertFalse(false);
  }
}
```

#### Actual Cli:
```
$ java -jar checkstyle-12.3.0-all.jar -c config_check.xml Test.java
Starting audit...
[ERROR] F:\GitHub\headhtmltagname\Test.java:9:25: There is more than 1 empty line after this line. [EmptyLineSeparator]
[ERROR] F:\GitHub\headhtmltagname\Test.java:12:15: There is more than 1 empty line after this line. [EmptyLineSeparator]
[ERROR] F:\GitHub\headhtmltagname\Test.java:18:6: There is more than 1 empty line after this line. [EmptyLineSeparator]
[ERROR] F:\GitHub\headhtmltagname\Test.java:22:11: There is more than 1 empty line after this line. [EmptyLineSeparator]
Audit done.
Checkstyle ends with 4 errors.
```
According to the reported violation, lines 18 and 22 have more than one empty line after them, but there are no empty lines after these lines.

#### Expected Cli:
```
$ java -jar checkstyle-12.3.0-all.jar -c config_check.xml Test.java
Starting audit...
[ERROR] F:\GitHub\headhtmltagname\Test.java:9:25: There is more than 1 empty line after this line. [EmptyLineSeparator]
[ERROR] F:\GitHub\headhtmltagname\Test.java:12:15: There is more than 1 empty line after this line. [EmptyLineSeparator]
[ERROR] F:\GitHub\headhtmltagname\Test.java:15:6: There is more than 1 empty line after this line. [EmptyLineSeparator]
[ERROR] F:\GitHub\headhtmltagname\Test.java:19:11: There is more than 1 empty line after this line. [EmptyLineSeparator]
Audit done.
Checkstyle ends with 4 errors.
```
In my opinion, it might be incorrect line referencing.

Praveen7294:
Fix #18438

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
@@ -267,18 +267,34 @@ public class EmptyLineSeparatorCheck extends AbstractCheck {
             }
         }
 
+        return positionToPotentialPostFixNode(result, line);
+    }
+
+    /**
+     * A post fix AST will always have a sibling METHOD CALL
+     * METHOD CALL will at least have two children
+     * The first child is DOT in case of POSTFIX which have at least 2 children
+     * First child of DOT again puts us back to normal AST tree which will
+     * recurse down below from here.
+     *
+     * @param postFixAst the ast to check.
+     * @param line the empty line which gives violation.
+     * @return The potential post fix node after which empty lines are present.
+     */
+    private static DetailAST positionToPotentialPostFixNode(DetailAST postFixAst, int line) {
+        DetailAST result = postFixAst;
         if (result.getNextSibling() != null) {
             final Optional<DetailAST> postFixNode = getPostFixNode(result.getNextSibling());
             if (postFixNode.isPresent()) {
-                // A post fix AST will always have a sibling METHOD CALL
-                // METHOD CALL will at least have two children
-                // The first child is DOT in case of POSTFIX which have at least 2 children
-                // First child of DOT again puts us back to normal AST tree which will
-                // recurse down below from here
                 final DetailAST firstChildAfterPostFix = postFixNode.orElseThrow();
                 result = getLastElementBeforeEmptyLines(firstChildAfterPostFix, line);
             }
         }
+
+        if (result.getLineNo() > line) {
+            result = postFixAst;
+        }
+
         return result;
     }
 
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheckTest.java
@@ -457,6 +457,19 @@ public class EmptyLineSeparatorCheckTest
     }
 
     @Test
+    public void testDisAllowMultipleEmptyLinesInsideClassMembers1() throws Exception {
+        final String[] expected = {
+            "23:27: " + getCheckMessage(MSG_MULTIPLE_LINES_INSIDE),
+            "26:19: " + getCheckMessage(MSG_MULTIPLE_LINES_INSIDE),
+            "31:64: " + getCheckMessage(MSG_MULTIPLE_LINES_INSIDE),
+            "36:19: " + getCheckMessage(MSG_MULTIPLE_LINES_INSIDE),
+        };
+        verifyWithInlineConfigParser(
+                getPath("InputEmptyLineSeparatorInsideClassMembers.java"),
+                expected);
+    }
+
+    @Test
     public void testAllowMultipleEmptyLinesInsideClassMembers() throws Exception {
         final String[] expected = {
             "53:1: " + getCheckMessage(MSG_SHOULD_BE_SEPARATED, "CLASS_DEF"),
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorInsideClassMembers.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/emptylineseparator/InputEmptyLineSeparatorInsideClassMembers.java
@@ -1,0 +1,41 @@
+/*
+EmptyLineSeparator
+allowNoEmptyLineBetweenFields = true
+allowMultipleEmptyLines = false
+allowMultipleEmptyLinesInsideClassMembers = false
+tokens = (default)PACKAGE_DEF, IMPORT, STATIC_IMPORT, CLASS_DEF, INTERFACE_DEF, ENUM_DEF, \
+         STATIC_INIT, INSTANCE_INIT, METHOD_DEF, CTOR_DEF, VARIABLE_DEF, RECORD_DEF, \
+         COMPACT_CTOR_DEF
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.whitespace.emptylineseparator;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class InputEmptyLineSeparatorInsideClassMembers {
+
+    public void foo(int a) {}
+
+    @Test
+    public void testFoo() { // violation 'There is more than 1 empty line after this line.'
+
+
+        int a = 10; // violation 'There is more than 1 empty line after this line.'
+
+
+        // violation 2 lines below 'There is more than 1 empty line after this line.'
+        InputEmptyLineSeparatorInsideClassMembers t =
+                new InputEmptyLineSeparatorInsideClassMembers();
+
+
+        // violation 2 lines below 'There is more than 1 empty line after this line.'
+        t.foo(10);
+        int b = 20;
+
+
+        Assert.assertFalse(false);
+    }
+}


Praveen7294:
Github, generate report for EmptyLineSeparator/all-examples-in-one

Praveen7294:
@Zopsss @romani i reviewed diff report everything looks fine.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
```diff
@@ -267,18 +267,23 @@ private static DetailAST getLastElementBeforeEmptyLines(DetailAST ast, int line)
             }
         }
 
-        if (result.getNextSibling() != null) {
-            final Optional<DetailAST> postFixNode = getPostFixNode(result.getNextSibling());
+        DetailAST reqAST = result;
+        if (reqAST.getNextSibling() != null) {
+            final Optional<DetailAST> postFixNode = getPostFixNode(reqAST.getNextSibling());
```
Zopsss: what is `reqAST`?

Praveen7294: `reqAST` is a temporary candidate node used during postfix handling.
It initially points to the current result, then allows the method to traverse into postfix-related AST structures (for example, method calls or increment expressions) without immediately overwriting result.

After the recursive traversal, `reqAST` is validated against the line constraint `(reqAST.getLineNo() <= line)`. Only if it satisfies this condition is it promoted to result.

In short, `reqAST` acts as a safe checkpoint that holds a potential replacement for result while exploring postfix nodes, ensuring the final returned AST node still lies on or before the target line.

Zopsss: can you explain it with AST? 

Also the `reqAST` name doesn't make much sense, you need to rename it to something more meaningful.

Praveen7294: > can you explain it with AST?

```
        |       |--SEMI -> ; [19:63]
        |       |--EXPR -> EXPR [23:13]
        |       |   `--METHOD_CALL -> ( [23:13]
        |       |       |--DOT -> . [23:9]
        |       |       |   |--SINGLE_LINE_COMMENT -> // [22:8]
        |       |       |   |   `--COMMENT_CONTENT ->  violation 2 lines below 'There is more than 1 empty line after this line.'\r\n [22:10]
        |       |       |   |--IDENT -> t [23:8]
        |       |       |   `--IDENT -> foo [23:10]
        |       |       |--ELIST -> ELIST [23:14]
        |       |       |   `--EXPR -> EXPR [23:14]
        |       |       |       `--NUM_INT -> 10 [23:14]
        |       |       `--RPAREN -> ) [23:16]
        |       |--SEMI -> ; [23:17]
```

Looking at the AST, there are two empty lines between the first `SEMI` token (`; [19:63]`) and the next `EXPR` token (`EXPR [23:13]`). Therefore, according to the method’s intent, the `SEMI` at line 19 should be identified as the last AST element before the empty lines.

```
        if (result.getNextSibling() != null) {
            final Optional<DetailAST> postFixNode = getPostFixNode(result.getNextSibling());
                if (postFixNode.isPresent()) {
                // A post fix AST will always have a sibling METHOD CALL
                // METHOD CALL will at least have two children
                // The first child is DOT in case of POSTFIX which have at least 2 children
                // First child of DOT again puts us back to normal AST tree which will
                // recurse down below from here
                final DetailAST firstChildAfterPostFix = postFixNode.orElseThrow();
                result = getLastElementBeforeEmptyLines(firstChildAfterPostFix, line);
            }
        }
```

The postfix-handling logic inspects `result.getNextSibling()` and detects a method call expression. Because of this, it recursively descends into the method call AST and updates result to the `SEMI` at line 23 (`; [23:17]`). This overrides the correct `result`, even though that node appears after the empty lines, which leads to a false positive.

To avoid this, `reqAST` is introduced as a temporary candidate that initially refers to the current `result`. If a possible postfix or method call expression is found, traversal continues and `reqAST` is updated accordingly. Before promoting it to the final result, its line number is validated against the target line:

```
if (reqAST.getLineNo() <= line) {
    result = reqAST;
}
```
This ensures that postfix traversal does not incorrectly cross empty-line boundaries and that the returned AST node truly lies on or before the required line.


> Also the reqAST name doesn't make much sense, you need to rename it to something more meaningful.

Is `candidateAST` an appropriate name? If not, could you please suggest a more meaningful one?

Praveen7294:
@Zopsss please review, i explained as you asked.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
romani: can code 
```
DetailAST reqAST = result;
        if (reqAST.getNextSibling() != null) {
            final Optional<DetailAST> postFixNode = getPostFixNode(reqAST.getNextSibling());
            if (postFixNode.isPresent()) {
                // A post fix AST will always have a sibling METHOD CALL
                // METHOD CALL will at least have two children
                // The first child is DOT in case of POSTFIX which have at least 2 children
                // First child of DOT again puts us back to normal AST tree which will
                // recurse down below from here
                final DetailAST firstChildAfterPostFix = postFixNode.orElseThrow();
                reqAST = getLastElementBeforeEmptyLines(firstChildAfterPostFix, line);
            }
        }

        if (reqAST.getLineNo() <= line) {
            result = reqAST;
        }
```

be written like:
```
        result = positionToPotentialPostFixNode(result);
        
```

in this case comment can be moved to javadoc and logic will be a bit more structred.

romani:
@Praveen7294 , please se my comment above.
Please rebase , sorry that we lost tracking of this PR.

please consider https://github.com/checkstyle/checkstyle/issues/18438#issuecomment-3822485966 also.

romani:
why we have this violation ?
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/7ec4b2bad97c4c11059d33f0e846ca4ef5d6492a_2026053909/reports/diff/spring-framework/index.html#A1

is this different defect in same Check?

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
Praveen7294: @romani Thanks for the help, it is now more structred.

Praveen7294:
> please se my comment above.
Please rebase , sorry that we lost tracking of this PR.

Done.

---
> please consider https://github.com/checkstyle/checkstyle/issues/18438#issuecomment-3822485966 also.

This false positive occurs when we add below property in `EmptyLineSeparatorCheck`, caught during [#18311](https://github.com/checkstyle/checkstyle/pull/18311#issuecomment-3691084557) :
```
<property name="allowMultipleEmptyLinesInsideClassMembers" value="false"/>
```


This property is not added in `google_checks.xml`. So there is no need to add inputs for google style.

Praveen7294:
>why we have this violation ?
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/7ec4b2bad97c4c11059d33f0e846ca4ef5d6492a_2026053909/reports/diff/spring-framework/index.html#A1
>
>is this different defect in same Check?

@romani The red one is for the base, and the green one is for the patch. Both are the same defect, but in the base code the violation references the wrong line, while in the patch code the violation references the correct line.

romani:
Github, generate report for EmptyLineSeparator/all-examples-in-one

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java
romani: Look good now

romani:
Ok, I checking violations all good.
But

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/96b8b50bbe62f142d0d4c1a228476862a61ea839_2026021830/reports/diff/spoon/index.html#A3 this looks like existing other problem.
Please check our issue tracker for it, if nothing, please create it.

romani:
I reviewed all from bottom until Orekit, please go though top to Orekit to spot unusual case, and let me know when you finished

Praveen7294:
> I reviewed all from bottom until Orekit, please go though top to Orekit to spot unusual case, and let me know when you finished

I reviewed from top to openjdk25, there is no unusual violation.

Praveen7294:
@romani 

> https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/96b8b50bbe62f142d0d4c1a228476862a61ea839_2026021830/reports/diff/spoon/index.html#A3 this looks like existing other problem.

i reviewed and found there is no handling case of anonymous class with method chaining.

Below example:
```
    public static void foo1() {
        Stack<Integer> stack = new Stack<>();
        stack.push(5);
 	    new EarlyTerminatingScanner<Void>() { // violation 'There is more than one empty line after this line.'



 	    }
 	    .setVisitCompilationUnitContent(true)
 	    .scan(element.getRoleInParent(), element);
    }
```

> Please check our issue tracker for it, if nothing, please create it.

This is a case of chained method call and the issue is about method call, so it's better to solve in this PR.

Sorry for the delay. I was working on this problem to identify the main cause.

romani:
Please create issue, we can improve in separate issue/PR.

Praveen7294:
> Please create issue, we can improve in separate issue/PR.

Done.

romani:
But this https://github.com/checkstyle/checkstyle/pull/18545#issuecomment-3838802236 is just  comment that Check consider as empty line
Check is comment aware 
https://github.com/checkstyle/checkstyle/blob/c7da3eea224b44d0073c805e5590c701b2c07add/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java#L135

Not clear why it is new violation

Praveen7294:
>But thishttps://github.com/https://github.com/checkstyle/checkstyle/pull/18545#issuecomment-3838802236 is just comment that Check consider as empty line
Check is comment aware

Yes, you are right. The check is comment-aware, but comments tokens are skipped(not consider as empty line) during AST traversal in the visitToken() method.

This one is to skip comment before current token
https://github.com/checkstyle/checkstyle/blob/a6594e00879bdfbfcdfdbe50b73bcbf8a75af543/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java#L169

This one is to skip comment after current token
https://github.com/checkstyle/checkstyle/blob/a6594e00879bdfbfcdfdbe50b73bcbf8a75af543/src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/EmptyLineSeparatorCheck.java#L180


but in one case comment tokens are not skipped when `EmptyLineSeparatorCheck` have to check empty line inside class member block, comment can occur inside class member block also. see below example for clarification:

Without comment:
```
import java.util.Stack;

public class Test {

    public static void foo1() {
        Stack<Integer> stack = new Stack<>();
        stack.push(5);
 	    new EarlyTerminatingScanner() {
             @Override
             public void visit(Object role, Object element) {
             
             }     // line 12
             



 	    }
            .scan();
    }
}
```

cli:
```
$ java -jar checkstyle-13.1.0-all.jar -c config_check.xml Test.java
Starting audit...
[ERROR] F:\GitHub\headhtmltagname\Test.java:12:14: '}' has more than 1 empty lines after. [EmptyLineSeparator]
[ERROR] F:\GitHub\headhtmltagname\Test.java:12:14: There is more than 1 empty line after this line. [EmptyLineSeparator]
Audit done.
```

With comment:
```
import java.util.Stack;

public class Test {

    public static void foo1() {
        Stack<Integer> stack = new Stack<>();
        stack.push(5);
 	    new EarlyTerminatingScanner() {
             @Override
             public void visit(Object role, Object element) {
             
             }        // line 12
             // test



 	    }
            .scan();
    }
}
```

cli:
```
$ java -jar checkstyle-13.1.0-all.jar -c config_check.xml Test.java
Starting audit...
[ERROR] F:\GitHub\headhtmltagname\Test.java:12:14: '}' has more than 1 empty lines after. [EmptyLineSeparator]
[ERROR] F:\GitHub\headhtmltagname\Test.java:13:16: There is more than 1 empty line after this line. [EmptyLineSeparator]
Audit done.
```
you can see after adding comment difference in second violation. This clearly shows comment is not skipped inside class member block. This explanation is only for clearing confusion around comment.


> Not clear why it is new violation

This violation is due to anonymous class with chained method call. if you want i can explain with ast why this violation related to anonymous class with chained method call (not related to comment).

romani:
EmptyLineSeparatorCheck#example7	There is more than 1 empty line after this line.	[146](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/96b8b50bbe62f142d0d4c1a228476862a61ea839_2026021830/reports/diff/spoon/xref/home/runner/work/checkstyle/checkstyle/.ci-temp/repositories/spoon/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java.html#L146)

```
public static ElementSourceFragment createSourceFragmentsFrom(CtElement element) {
144 		ElementSourceFragment rootFragment = new ElementSourceFragment(element, null);
145 		Deque<ElementSourceFragment> parents = new ArrayDeque<>();
146 		parents.push(rootFragment);
147 		/*
148 		 * scan all children of `element` and build tree of SourceFragments
149 		 * Note: we cannot skip implicit elements,
150 		 * because CtBlock can be implicit but contains non implicit elements, which has to be processed.
151 		 */
152 		new EarlyTerminatingScanner<Void>() {
```

I am human, and I don't understand why checkstyle thinks that after 146 lines there is more than 1 empty lines. It is not true for sure

Praveen7294:
> I am human, and I don't understand why checkstyle thinks that after 146 lines there is more than 1 empty lines. It is not true for sure

see line [160](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/96b8b50bbe62f142d0d4c1a228476862a61ea839_2026021830/reports/diff/spoon/xref/home/runner/work/checkstyle/checkstyle/.ci-temp/repositories/spoon/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java.html#L160). After this line, there are more than one empty lines. Due to this, the check tries to find the last token before the empty line. However, because of the chained method call, it is unable to determine the correct last token before the empty line.

Praveen7294:
@romani can we move forward this PR, remaining false positive cases will resolve in #18858 issue. if you are ok with my clarification.

romani:
Github, generate report for EmptyLineSeparator/all-examples-in-one

Praveen7294:
@romani ping

romani:
Github, generate report for EmptyLineSeparator/all-examples-in-one

romani (approved):
thanks a lot for this update.
All good.

Zopsss (approved):
LGTM!

Praveen7294:
@romani and @Zopsss, since you both approved my changes, is it okay to merge? I have rebased my PR branch.

romani:
Thanks a lot , sorry that we lost this PR is tracking

Issue #13796, opened 2023-09-28
Give more clarity on xdocs for AnnotationLocation


relentless-pursuit:
Link: https://checkstyle.sourceforge.io/checks/annotation/annotationlocation.html

I am quoting the documentation for AnnotationLocation below to highlight the perceived inconsistency. 

Use the following configuration to allow multiple annotations on the same line:
```
<module name="AnnotationLocation">
  <property name="allowSamelineMultipleAnnotations" value="true"/>
  <property name="allowSamelineSingleParameterlessAnnotation"
    value="false"/>
  <property name="allowSamelineParameterizedAnnotation" value="false"/>
</module>
```
Example to allow any location multiple annotations:
```
@NotNull private boolean field1; //ok
@Override public int hashCode() { return 1; } //ok
@NotNull //ok
private boolean field2;
@Override //ok
public boolean equals(Object obj) { return true; }
@Mock DataLoader loader; //ok
@SuppressWarnings("deprecation") DataLoader loader; //ok
@SuppressWarnings("deprecation") public int foo() { return 1; } //ok
@NotNull @Mock DataLoader loader; //ok
```

Use the following configuration to allow only one and only parameterized annotation on the same line:
```
<module name="AnnotationLocation">
  <property name="allowSamelineMultipleAnnotations" value="false"/>
  <property name="allowSamelineSingleParameterlessAnnotation"
    value="false"/>
  <property name="allowSamelineParameterizedAnnotation" value="true"/>
</module>
```

Example to allow only one and only parameterized annotation on the same line:
```
@NotNull private boolean field1; //violation
@Override public int hashCode() { return 1; } //violation
@NotNull //ok
private boolean field2;
@Override //ok
public boolean equals(Object obj) { return true; }
@Mock DataLoader loader; //violation
@SuppressWarnings("deprecation") DataLoader loader; //ok
@SuppressWarnings("deprecation") public int foo() { return 1; } //ok
@NotNull @Mock DataLoader loader; //violation
```

end of doc quote.

----------

But, the below config line which is same for above examples mentioned, however presents inconsistent scenarios.

`<property name="allowSamelineSingleParameterlessAnnotation" value="false"/>`
For "Example to allow any location multiple annotations":
`@Override public int hashCode() { return 1; } //ok`

For "Example to allow only one and only parameterized annotation on the same line":
`@Override public int hashCode() { return 1; } //violation`

--------

[...]

PR #18546, opened 2026-01-07
Clarify interaction of same-line annotation properties in AnnotationL…

Krish-876:
Issue: #13796

This PR clarifies the interaction between same-line annotation properties in
AnnotationLocation xdocs. It explains that allowing parameterized annotations
on the same line does not implicitly allow non-parameterized annotations.

doc: clarify interaction of same-line annotation properties


Brijeshthummar02:
@Krish-876 there's no code changes in this PR, check it and raise the PR again.

Krish-876:
I tried multiple approaches to clarify this via xdocs (including restructuring Example 3), but ```mvn verify``` regenerates the examples and restores the current content.
It seems Example 3 is generated from tests, so resolving this likely requires adjusting the example source/tests rather than docs-only changes.
Could you please confirm the preferred approach?

PR #18552, opened 2026-01-07
Issue #11163: Enforce file size on WhiteSpaceAround

MonuChaudhary14:
Issue #11163 

Enforce file size on WhiteSpace


romani (approved):
Thanks a lot

Issue #18435, opened 2025-12-26
Fix xdocs Examples AST Consistency Test (Reduce suppressions list)


smita1078:
## Overview

The `XdocsExamplesAstConsistencyTest` validates that xdocs example files maintain consistent code structure (differing only in comments and configuration). Currently, we have **228 examples** in the suppression list that need to be reviewed and fixed.

## Goal

Reduce or eliminate entries from the `SUPPRESSED_EXAMPLES` list by fixing the underlying AST issues in example files.

## What Needs to Be Done

Each suppressed example needs investigation to determine:

1. **Is the code actually different?** → Fix the code to match other examples
2. **Should the code be different?** → Mark it as independent (different use case)

## Example: suppresswarningsholder (Already Fixed )

**Before:**
- `SUPPRESSED_EXAMPLES` contained: `Example2`, `Example3`, `Example4`
- These examples were failing validation

**Investigation revealed:**
- Example1 = showcase broad suppression (different/Specific use case)
- Example2, 3, 4 = showcase aliasList variations (should match)

**Fix applied:**
1. Marked Example1 as independent (it has different code intent)
2. Fixed Example2's code to match Example3 and Example4
3. Removed all 3 examples from suppression list kept Example1 as it has specific usecase

**Result:**  Test now passes.

## Current Suppression List (check master HEAD version for latest state)

https://github.com/checkstyle/checkstyle/blob/78f9db1cb555aeb2d37531cb6d49be0448d2c624/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java#L88-L91

### Pick a check from the list above

Example: `checks/annotation/annotationonsameline`

### Investigate the examples

1. Navigate to: `src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/annotation/annotationonsameline/`
2. Look at all `Example*.java` files
3. Read the comments to understand what each example demonstrates
4. Compare the code between examples

### Determine the issue

**Question**: Should these examples have the same code structure or different code structures?

- **Same structure** = They demonstrate the same check with different configuration
- **Different structure** = They demonstrate different use cases of the check

### Apply the fix

**If examples should have the same code:**
- Fix the code so all examples match structurally
- Only comments and configuration values should differ

**If an example legitimately has different code:**
- No code change needed
- The test framework will handle it appropriately, just add in the suppression list.

### Run the test

```bash
[...]

PR #18505, opened 2026-01-04
Issue #18435: Remove emptyforinitializerpad Example2 from XdocsExampl…

ZaheerAhmadDev:
Issue #18435:
Remove emptyforinitializerpad Example2 from XdocsExamplesAstConsistencyTest.
Update Example 2 to match Example 1 Ast.

https://www.diffchecker.com/kS03ZKeN/
Comparison of Example1 and  Example2
Both Example2 and Example1 have same numbers and spacing.

PR #18566, opened 2026-01-09
Issue #18435: Fix AST consistency across MethodNameCheck xdocs examples

ekansh0unofficial:
Issue: #18435 

This PR improves AST consistency across MethodNameCheck xdocs examples by aligning examples with equivalent semantic intent. 

- Example 1 , 2 , 5 ,6 and 7 are normalized to share a consistent AST structure while preserving their original configurational behaviour. 
- Example 3 , 4 are still remain suppressed, as they demonstrate constructor-specific behaviour and should not be normalized with method-only examples.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
@@ -207,12 +207,6 @@ public class XdocsExamplesAstConsistencyTest {
             "checks/naming/localvariablename/Example5",
             "checks/naming/membername/Example2",
             "checks/naming/membername/Example3",
-            "checks/naming/methodname/Example2",
-            "checks/naming/methodname/Example3",
-            "checks/naming/methodname/Example4",
-            "checks/naming/methodname/Example5",
-            "checks/naming/methodname/Example6",
-            "checks/naming/methodname/Example7",
             "checks/naming/parametername/Example2",
             "checks/naming/parametername/Example3",
             "checks/naming/parametername/Example4",
@@ -319,8 +313,11 @@ public class XdocsExamplesAstConsistencyTest {
             "filters/suppresswithplaintextcommentfilter/Example5",
             "filters/suppresswithplaintextcommentfilter/Example9",
             // No properties in module, multiple very different examples to ease reading
-            "checks/annotation/missingoverrideonrecordaccessor/Example2"
-    );
+            "checks/annotation/missingoverrideonrecordaccessor/Example2",
+            // contains ExampleX constructors
+            "checks/naming/methodname/Example3",
+            "checks/naming/methodname/Example4"
+            );
 
     /**
      * Tests that examples with the same code structure maintain consistency.
--- a/src/site/xdoc/checks/naming/methodname.xml
+++ b/src/site/xdoc/checks/naming/methodname.xml
@@ -93,9 +93,10 @@
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
 class Example1 {
   public void method1() {}
-  protected void method2() {}
+  protected void Method2() {} // violation 'Name 'Method2' must match pattern'
   private void Method3() {} // violation 'Name 'Method3' must match pattern'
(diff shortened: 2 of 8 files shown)


romani:
GitHub, generate website

romani:
>Example of Pull Request with diff image to prove that content of examples is same https://github.com/checkstyle/checkstyle/pull/18505

please share images of diff to speedup validation.

romani:
>Example 3 , 4 are still remain suppressed, as they demonstrate constructor-specific behaviour and should not be normalized with method-only examples.

please move them to end of suppression list and make comment above `// contains ExampleX constructors` so it will be as permanente excuse to not follow same AST in Examples.

ekansh0unofficial:
Example 1 & 2 : https://www.diffchecker.com/TUeNdpAr/
Example 1 & 5 : https://www.diffchecker.com/IVx0rPgS/
Example 1 & 6 : https://www.diffchecker.com/EANiGiRD/
Example 1 & 7 : https://www.diffchecker.com/pvzpSCJb/

Note: 
Identifier text (e.g. methodX vs MethodX) differs across examples.
The AST consistency test validates only token structure, so this does not affect consistency.
Please let me know if identifier text differences were not intended to be ignored by the AST consistency test; I’m happy to adjust accordingly.

romani:
>Example 1 & 2 :


Good 

All other should be same by java code, same method names, ... . Only comments should be different.

ekansh0unofficial:
Example 1v2 : https://www.diffchecker.com/GII3Fr7F/
Example 1v5 : https://www.diffchecker.com/JAMBgG0G/
Example 1v6 : https://www.diffchecker.com/aOjozfhY/
Example 1v7 : https://www.diffchecker.com/crJvBQbJ/

All the identifier names have been made consistent.

ekansh0unofficial:
@romani  could you please check if any changes are required. Thank you.

romani:
GitHub, generate website

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -319,8 +313,11 @@ public class XdocsExamplesAstConsistencyTest {
             "filters/suppresswithplaintextcommentfilter/Example5",
             "filters/suppresswithplaintextcommentfilter/Example9",
             // No properties in module, multiple very different examples to ease reading
-            "checks/annotation/missingoverrideonrecordaccessor/Example2"
-    );
+            "checks/annotation/missingoverrideonrecordaccessor/Example2",
+            // contains ExampleX constructors
+            "checks/naming/methodname/Example3",
+            "checks/naming/methodname/Example4"
```
romani, line 319: @smita1078 , I thinks we can improve ast diff test to exclude constructors too, right ?

romani: https://checkstyle.org/checks/naming/methodname.html#Example4-code

but may be this is very crazy edge case that does not happen often.
Lets think on this, if there will be more cases like this.

smita1078: > @smita1078 , I thinks we can improve ast diff test to exclude constructors too, right ?

Yes , we can

ekansh0unofficial: Thanks for the discussion, this is an interesting point.

I wanted to share one concern regarding excluding constructors from the AST diff. The current AST consistency test is valuable (at least from my understanding) because it is a strict, structure-preserving comparison. If we start filtering out specific node kinds (like constructors), we slightly weaken the guarantees of the test and may risk hiding real structural inconsistencies in future examples.

Of course, this is just my perspective — happy to follow whatever direction you think is best.

romani: We already doing this with class name, as it is not avoidable. Ctor name is named by class.

Let's not change AST validation for now, please put  such cases to bottom of list with comment why it becomes persistent suppression.

Issue #18035, opened 2025-11-01
Resolve Pitest Suppressions - indentation


romani:
As a part of this issue, we have to kill the mutation listed below and remove the suppression from [pitest xxxx suppression.xml](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/)

To understanding how to kill the mutation 
Visit [Wiki page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports), [Pitest-docs](https://pitest.org/) and expected actions and workflow at https://github.com/checkstyle/checkstyle/issues/12341.

Target is to remove all `<mutation` from [Pitest-indentation](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-indentation-suppressions.xml)

ATTENTION;
We do not allow pure unit testing, all testing should be done by full run of Check on real Input file

vennelagantiSubramanyam:
Hi @romani , I'd love to contribute to checkstyle and I have started working on this task can you please assign it to me?

romani:
This issue is not easy for new contributors 

Please [read](https://github.com/checkstyle/checkstyle/blob/master/.github/CONTRIBUTING.md#getting-started) and watch videos at [Starting_Development](https://checkstyle.org/beginning_development.html#Starting_Development).

No assignments, contributors just make comment "I am in ....'" and can start working

ivamly:
I am on HandlerFactory.java suppressions

romani:
You subconsciously choose most complicated modules :).
I recommend to finish other pitest first.
Indentation is very problematic module, that has numerous defects.

ivamly:
True! Thanks for the guidance. I'll submit PR for this one since I'm almost done (based on my understanding), then move to others.

romani:
I removed approved label to let people to focus on other issues first. As this Check needs special attention  and knowledge.

vivek-0509:
Will reference the other two open prs to this issue . After they are merged , suppression file will be empty 🎉

romani:
What sounded as impossible, is done by @vivek-0509

vivek-0509:
@romani Thankyou so much for your guidance which made this possible.

PR #18578, opened 2026-01-09
Issue #18035: Exclude memory cleanup methods from Pitest mutations

vivek-0509:
Issue #18035:

### DIFF
--- a/config/pitest-suppressions/pitest-indentation-suppressions.xml
+++ b/config/pitest-suppressions/pitest-indentation-suppressions.xml
@@ -1,29 +1,3 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <mutation unstable="false">
-    <sourceFile>HandlerFactory.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.indentation.HandlerFactory</mutatedClass>
-    <mutatedMethod>clearCreatedHandlers</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.VoidMethodCallMutator</mutator>
-    <description>removed call to java/util/Map::clear</description>
-    <lineContent>createdHandlers.clear();</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>IndentationCheck.java</sourceFile>
-    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.indentation.IndentationCheck</mutatedClass>
-    <mutatedMethod>beginTree</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.VoidMethodCallMutator</mutator>
-    <description>removed call to com/puppycrawl/tools/checkstyle/checks/indentation/HandlerFactory::clearCreatedHandlers</description>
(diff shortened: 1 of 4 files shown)


vivek-0509:
Wrapping the calls into `clearState()` still leaves one surviving mutation when VoidMethodCallMutator removes the 
clearState() call in beginTree(). This requires a suppression since excludedMethods only protects method bodies, not callers.
We could alternatively follow the UnmodifiableCollectionUtil pattern by creating a utility class for cleanup operations and using avoidCallsTo for it. This would result in zero suppressions. Would this approach make sense?

romani:
still open https://github.com/hcoles/pitest/issues/404

survival is:
```
patch -p1 <<'EOF'
diff --git a/config/pitest-suppressions/pitest-indentation-suppressions.xml b/config/pitest-suppressions/pitest-indentation-suppressions.xml
index 753200a..e6fdae9 100644
--- a/config/pitest-suppressions/pitest-indentation-suppressions.xml
+++ b/config/pitest-suppressions/pitest-indentation-suppressions.xml
@@ -62,4 +62,13 @@
     <description>replaced boolean return with true for com/puppycrawl/tools/checkstyle/checks/indentation/CommentsIndentationCheck::isUsingOfObjectReferenceToInvokeMethod</description>
     <lineContent>return root.getFirstChild().getFirstChild().getFirstChild() != null</lineContent>
   </mutation>
+
+    <sourceFile>IndentationCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.indentation.IndentationCheck</mutatedClass>
+    <mutatedMethod>beginTree</mutatedMethod>
+    <description>removed call to com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck::clearState</description>
+    <lineContent>clearState();</lineContent>
+  </mutation>
 </suppressedMutations>
EOF
```

romani:
>  UnmodifiableCollectionUtil pattern 

lets keep is absolutely last resort, and it will very ugly.


can we archive mutation killing by processing two files on same config? we already have methods for execution of verifyXxxxxxxx with few files on same config. It might help as handlers looks like dependent on exact ast , BUT DetailAstImpl is does not implement "equals" for it will never be "same", so collision is not possible.


I think we should just accept pure junit test with reflection. please see usage of `TestUtil.setInternalState`.
test method that you will create should have epxlanaion comment to explain why anti-pattern is used, kind of summary of this PR conversation.

vivek-0509:
Sure, I'll add the unit test with reflection  include a proper explanation comment summarizing why this approach is necessary. Once this PR and the other open PRs are merged, the suppression list for pitest-indentation should be completely empty. Thanks for the guidance!

vivek-0509:
@romani Could you please review and merge the other open PRs for this issue? Once those are merged, I'll add the reflection test and update the pitest mutation threshold to 100% in this PR. That way, the suppression list will be completely empty and this can be the final PR for issue #6294.

romani:
I did some, let me know if some is missed my attention.

vivek-0509:
All done just one left, it has conflict with master branch will resolve that

vivek-0509:
@romani all the mutations are killed, suppression list is now empty, closes https://github.com/checkstyle/checkstyle/issues/18035 🚀

Issue #18024, opened 2025-11-01
Resolve Pitest Suppressions - common


romani:
As a part of this issue, we have to kill the mutation listed below and remove the suppression from [pitest xxxx suppression.xml](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/)

To understanding how to kill the mutation 
Visit [Wiki page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports), [Pitest-docs](https://pitest.org/) and expected actions and workflow at https://github.com/checkstyle/checkstyle/issues/12341.

Target is to remove all `<mutation` from [Pitest-Common](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-common-suppressions.xml)

romani:
There are 2 mutations left to be resolved

Youssefalaa7:
I am on 
```
  <mutation unstable="false">
    <sourceFile>SarifLogger.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.SarifLogger</mutatedClass>
    <mutatedMethod>renderSeverityLevel</mutatedMethod>
    <lineNumber>453</lineNumber>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.ConstructorCallMutator</mutator>
    <description>removed call to java/lang/MatchException::&lt;init&gt;</description>
    <lineContent>return switch (severityLevel) {</lineContent>
  </mutation>
```

romani:
last fix is merged

PR #18455, opened 2025-12-29
Issue #18024: Resolve Pitest Suppressions common

Brijeshthummar02:
Issue #18024

PR #18580, opened 2026-01-09
Issue #18024: Resolve Pitest Suppressions - common

Youssefalaa7:
Issue #18024: Resolve Pitest Suppressions - common

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
@@ -111,6 +111,14 @@ public final class SarifLogger extends AbstractAutomaticBean implements AuditLis
     /** Comma and line separator. */
     private static final String COMMA_LINE_SEPARATOR = ",\n";
 
+    /** Map of severity levels to SARIF level strings. */
+    private static final Map<SeverityLevel, String> SEVERITY_LEVEL_MAP = Map.of(
+        SeverityLevel.IGNORE, "none",
+        SeverityLevel.INFO, "note",
+        SeverityLevel.WARNING, "warning",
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/SarifLogger.java
```diff
@@ -451,12 +459,7 @@ private static String renderFileNameUri(final String fileName) {
      * @return the rendered severity level in string.
      */
     private static String renderSeverityLevel(SeverityLevel severityLevel) {
-        return switch (severityLevel) {
-            case IGNORE -> "none";
-            case INFO -> "note";
-            case WARNING -> "warning";
-            case ERROR -> "error";
-        };
+        return SEVERITY_LEVEL_MAP.get(severityLevel);
```
romani, line 462: Can we keep switch but add in it `defailt -> null`?

Youssefalaa7: This will not work because we will not be able to test this default in the switch without using unit tests so it will be similar to this commit https://github.com/checkstyle/checkstyle/pull/18455

romani: https://github.com/hcoles/pitest/issues/1442 is reported, 

This update is unfortunate, but as it is last in group, lets do this. 
Ideally better to exclude this method from pitest mutation in pitest config, until pitest issue is fixed

romani:
it might not work, as https://github.com/hcoles/pitest/issues/404 still open, but please try

Youssefalaa7:
Thanks i will try to look for a way to exclude mutations from renderSeverityLevel method

PR #18597, opened 2026-01-10
dependency: change checker-qual dependency scope to provided

vivek-0509:
Change checker-qual dependency scope to provided.

We should not have this library in our depencies, to avoid our user to resolve conflict of versions, especially as we are always using latest versions.

https://github.com/checkstyle/checkstyle/commit/c170d4a43be3d41f5ad900ecd64c33b6355298e0 introducing workaround to avoid dependency to checker artifacts.

### DIFF
--- a/config/import-control.xml
+++ b/config/import-control.xml
@@ -200,7 +200,6 @@
     </file>
     <subpackage name="imports">
       <allow class="com.puppycrawl.tools.checkstyle.XmlLoader" local-only="true"/>
-      <allow class="org.checkerframework.checker.index.qual.IndexOrLow"/>
     </subpackage>
     <subpackage name="indentation">
       <allow pkg="java.lang.reflect"/>
--- a/pom.xml
+++ b/pom.xml
@@ -471,6 +471,7 @@
       <groupId>org.checkerframework</groupId>
       <artifactId>checker-qual</artifactId>
       <version>${checkerframework.version}</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.maven.doxia</groupId>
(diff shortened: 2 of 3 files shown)


romani:
CI is not happy, do you have same on local?

vivek-0509:
No i was not getting it locally i ran mvn clean verify and the build was successful

romani:
You moved to advanced level of CI

https://github.com/checkstyle/checkstyle/actions/runs/20883117675/job/60002375650?pr=18597#step:6:1

Each CI job is executable on local

vivek-0509:
Hi, @romani 
I investigated the CI failure and found the root cause. When I removed the checker-qual dependency from pom.xml, the Checker Framework CI profiles (such as checker-formatter, checker-index, etc.) started failing with a silent "Compilation failure" error. The issue is that these profiles use the Checker Framework as an annotation processor via forked JVM compilation (configured with <fork>true</fork> in the profile's compiler plugin settings). While the checker artifact does bring checker-qual as a transitive dependency, it comes in with runtime scope. However, during the forked annotation processing, the Checker Framework needs checker-qual to be available on the compile classpath to resolve annotation types used internally during its static analysis. Without it on the compile classpath, the annotation processor fails to initialize properly, causing the build to abort without a detailed error message.

I verified this locally by running the failing CI command. When the checker-qual dependency is completely removed, the build fails. When I add it back with` <scope>provided</scope>`, the build succeeds. The provided scope ensures that checker-qual is available at compile time for our CI's Checker Framework analysis, but it is not included as a transitive dependency for users. This keeps dependencies minimal for users while keeping the CI infrastructure functional.

To summarize the proposed solution: instead of removing the checker-qual dependency entirely,we can change its scope to provided. This keeps it off the user's classpath (since provided dependencies are not transitive) while still making it available for the Checker Framework profiles during our CI builds.

Please let me know if this approach works for you, or if you have an alternative solution in mind.

romani:
Let's continue with provided scope.

vivek-0509:
Sure

Issue #17988, opened 2025-10-24
Resolve error-prone violations


Atharv3221:
Follow up of #17845
```
Source File: "AbbreviationAsWordInNameCheck.java"
Bug Pattern: "JdkObsolete"
Description: "It is very rare for LinkedList to out-perform ArrayList or ArrayDeque. Avoid it unless you&apos;re willing to invest a lot of time into benchmarking. Caveat: LinkedList supports null elements, but ArrayDeque does not."
Line Contents: "final List&lt;DetailAST&gt; result = new LinkedList&lt;&gt;();"
Line Number: 436

Source File: "AbstractImportControl.java"
Bug Pattern: "JdkObsolete"
Description: "It is very rare for LinkedList to out-perform ArrayList or ArrayDeque. Avoid it unless you&apos;re willing to invest a lot of time into benchmarking. Caveat: LinkedList supports null elements, but ArrayDeque does not."
Line Contents: "private final Deque&lt;AbstractImportRule&gt; rules = new LinkedList&lt;&gt;();"
Line Number: 33

Source File: "AbstractSuperCheck.java"
Bug Pattern: "JdkObsolete"
Description: "It is very rare for LinkedList to out-perform ArrayList or ArrayDeque. Avoid it unless you&apos;re willing to invest a lot of time into benchmarking. Caveat: LinkedList supports null elements, but ArrayDeque does not."
Line Contents: "private final Deque&lt;MethodNode&gt; methodStack = new LinkedList&lt;&gt;();"
Line Number: 48

Source File: "ConfigurationLoader.java"
Bug Pattern: "InlineFormatString"
Description: "Prefer to create format strings inline, instead of extracting them to a single-use constant"
Line Contents: "private static final String SAX_PARSE_EXCEPTION_FORMAT = &quot;%s - %s:%s:%s&quot;;"
Line Number: 78

Source File: "CsvFilterElement.java"
Bug Pattern: "EqualsGetClass"
Description: "Prefer instanceof to getClass when implementing Object#equals."
Line Contents: "public boolean equals(Object object) {"
Line Number: 105

Source File: "CustomImportOrderCheck.java"
Bug Pattern: "StringSplitter"
Description: "String.split(String) has surprising behavior"
Line Contents: "final String[] import1Tokens = import1.split(separator);"
Line Number: 654

Source File: "CustomImportOrderCheck.java"
Bug Pattern: "StringSplitter"
Description: "String.split(String) has surprising behavior"
[...]

PR #18598, opened 2026-01-10
Issue #17988: Fixing error-prone violations

aclfe:
Issue #17988: Fixing error-prone violations

https://errorprone.info/bugpattern/InlineFormatString

Most often, locally issues are not shown. CI tends to show the errors better.
Running checks to investigate.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -278,7 +278,7 @@
       -Xep:StringSplitter:ERROR
       -Xep:JdkObsolete:ERROR
       <!-- Reason at https://github.com/checkstyle/checkstyle/issues/17988. -->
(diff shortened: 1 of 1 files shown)


romani:
@aclfe , 

https://errorprone.info/bugpattern/InlineFormatString

pleaese just disable this rule, we like extra variables it helps with reading code clearly

aclfe:
> @aclfe ,
> 
> https://errorprone.info/bugpattern/InlineFormatString
> 
> pleaese just disable this rule, we like extra variables it helps with reading code clearly

Oh understood! This makes a lot more sense, I'll just turn it off then. Thanks for clarifying :+1:

Issue #18065, opened 2025-11-07
New Check: IllegalSymbol to forbit emoj in code


romani:
AI generates unicode emoj in code comments
[example](https://github.com/checkstyle/checkstyle/pull/18008/files#diff-e8685d13c843aea432b0c89f0ef7f09fddfd9f00bf2c332e7e7d1a47e776354dR60): 

> // ✅ Enhancement 1: Basic validation

this also raised by other users to address this at https://github.com/checkstyle/checkstyle/discussions/17983#discussioncomment-14888080

Potential soluton: 
```
    <module name="RegexpSingleline">
        <property name="format" value="\u2705" />
        <property name="message" value="Unicode emoj symbols should not be used." />
    </module>
```

use https://www.mauvecloud.net/charsets/CharCodeFinder.html in mode "Hexadecimal Character Codes" to get code.

To make it easier to search and collaborate to extend to all other Unicode pictures, we can put this as example https://checkstyle.org/checks/regexp/regexpsingleline.html#Examples so it will part of git files and user can contribute to make list to cover most

Website sources are part of main repository https://github.com/checkstyle/checkstyle/blob/master/src/site/xdoc/checks/regexp/regexpsingleline.xml.template

We need to put in https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle-checks.xml config  that violates majority of unicode emoj symbols, try to find in web or AI list of them.

romani:
@zyadahmed, can you help us fix it ?

zyadahmed:
@romani sure

zyadahmed:
@romani  After multiple attempts ,it turns out that reliably detecting all emojis is extremely difficult. Java’s regex  operates on UTF-16 code units rather than full Unicode codepoints, so many emojis cannot be matched consistently. Because of these technical limitations, we cannot accurately catch most emojis using a simple or stable regex-based Checkstyle rule, so what is your recommendation ?

romani:
Nice reason to create special Check that will forbid usage of special symbols by codes or ranges of codes, I hope in utf table such symbols are in special ranges of codes , so it will be easier to mark them as forbidden.

Please suggest name of Check , and what properties you think should be there, and draw by text example what could be a behavior of Check.
We usually try to agree upfront on design and behavior to avoid wasting contributors time to redo and rename in mid of development.

zyadahmed:
[...]

Issue #1020, opened 2015-04-29
Standard Package wrong RegExp for CustomImportOrderCheck


rnveach:
I had an import from another utility with the naming "com.rveach.javaconverter" and CustomImportOrderCheck was flagging it in the wrong location as needed to be in "STANDARD_JAVA_PACKAGE". This is definitely not a standard java package.

The root of the problem is the default value for "standardPackageRegExp".
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/CustomImportOrderCheck.java#L223

It should be something like "^(java\\.|javax\\.).*" to ensure only the first section of the import is "java", not that "java" appears anywhere in the statement.

romani:
UT has to be created to avoid that issue in future.

romani:
merged.

PR #18609, opened 2026-01-11
Add IllegalSymbolCheck to Detect Unicode Symbols in Code

smita1078:
#18065 
[#1020](https://github.com/checkstyle/contribution/pull/1020) opened in contribution repo.

**Summary**
Implements a new `IllegalSymbolCheck` to detect and prevent Unicode symbols (especially emojis) in code, which are commonly inserted by AI code generators.

New Check: `IllegalSymbolCheck`
Located at: `src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java`

Features:
- Validates Unicode symbols by their codepoints or ranges
- Supports multiple input formats: `0x1F600`, `\u1F600`, `U+1F600`, `decimal`
- Can operate in `ASCII-only` mode to enforce ASCII-only source files
- Works on comments, string literals, character literals, text blocks, and identifiers

Tested:
image
image

New module config: https://gist.githubusercontent.com/smita1078/a69fc778cb575f1f99adbe5e9d752c68/raw/b4bea85326ea3e9f2845ce8eca9343cee52049f9/illegalsymbol-test-config.xml
Report label: IllegalSymbol Check - Testing Emoji Detection and ASCII-Only Mode


smita1078:
@romani Please review

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_ja.properties
```diff
@@ -22,6 +22,7 @@ fall.through.last=switch 文の最後の分岐からフォールスルーして
 final.variable=変数 ''{0}'' は final として宣言すべきです。
 hidden.field=''{0}'' がフィールドを隠しています。
 illegal.catch=''{0}'' をキャッチすることは許可されていません。
+illegal.symbol=???Unicode??????????.
 illegal.throw=''{0}'' をスローすることは許可されていません。
```
vivek-0509, line 26: Please fix the violation messages, in you Intellija ide settings change the default encoding for properties files to UTF-8.

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_ru.properties
```diff
@@ -22,6 +22,7 @@ fall.through.last=Переход из последней ветви инстру
 final.variable=Переменная ''{0}'' должна быть объявлена final.
 hidden.field=''{0}'' скрывает поле.
 illegal.catch=Недопустимо ловить ''{0}''.
+illegal.symbol=????????? ???????????? ?????? Unicode.
 illegal.throw=Недопустимо выбрасывать ''{0}''
```
vivek-0509, line 26: Same for this and other message properties files

vivek-0509:
@smita1078 Please go through the section [Executing generation using github action](https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#executing-generation-using-github-action) and try to generate the report. If you need any help let me know.

smita1078:
> @smita1078 Please go through the section [Executing generation using github action](https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#executing-generation-using-github-action) and try to generate the report. If you need any help let me know.

Sure thanks

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_ja.properties
smita1078: Done, thanks for the help!

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_ru.properties
smita1078: Done

smita1078:
GitHub, generate report

smita1078:
GitHub, generate report

smita1078:
GitHub, generate report

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,231 @@
...
+        symbolCodes = symbols;
+    }
+
+    /**
+     * Setter to control whether only ASCII characters are allowed.
+     *
+     * @param asciiOnly true to allow only ASCII characters
+     * @since 10.19.0
+     */
+    public void setAsciiOnly(boolean asciiOnly) {
```
vivek-0509: Please remove the `@since `from here we do not need them, correct version is already mentioned above at class level.

vivek-0509 (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
smita1078: Done

vivek-0509: Sorry, i think i miss understood please add the tag to property setters with correct version

smita1078: No worries, I missed it too.
Done!

romani:
conflict , please rebase

smita1078:
> conflict , please rebase

Done

romani:
GitHub, generate website

romani:
@vivek-0509 , please finalize review.

I have some worries on name of check, as it hase no Unicode, and looks like any symbol, but violation message has "Unicode" harcoded. I will see more after website is generated.

romani:
I think we need to disallow all unicodes from https://en.wikipedia.org/wiki/List_of_Unicode_characters `U+219x` until `U+27BF` , from `U+1F70x` until end.
And it should be default behavior.

let put https://en.wikipedia.org/wiki/List_of_Unicode_characters in documentation ,it will help other to see where we got such ranges.

if user want to customize, they are always welcome to override with more precise or less strict ranges.

We do not need property `asciiOnly` as this conflicting with name `IllegalSymbol`.
We can show example of ascii only by usage of `illegalSymbols` in `U+0000-U+001F, U+007E-`.

smita1078:
> I think we need to disallow all unicodes from https://en.wikipedia.org/wiki/List_of_Unicode_characters `U+219x` until `U+27BF` , from `U+1F70x` until end. And it should be default behavior.
> 
> let put https://en.wikipedia.org/wiki/List_of_Unicode_characters in documentation ,it will help other to see where we got such ranges.
> 
> if user want to customize, they are always welcome to override with more precise or less strict ranges.
> 
> We do not need property `asciiOnly` as this conflicting with name `IllegalSymbol`. We can show example of ascii only by usage of `illegalSymbols` in `U+0000-U+001F, U+007E-`.

Sure, let me think around it . I will make changes after discussing them here

smita1078:
> I have some worries on name of check, as it hase no Unicode, and looks like any symbol, but violation message has "Unicode" harcoded. I will see more after website is generated.

Will explore other naming options as well

vivek-0509:
@romani 
Regarding the naming concern, I think we can keep the name IllegalSymbol and change the violation message from "Illegal Unicode symbol detected" to just "Illegal symbol detected".
What do you think ??

romani:
is check is named IllegalSymbol , user will expect any symbols to be possible to checked.
So violation should be `"Illegal symbol detected"`. User can ban any symbol, even ASCII  code from usage like `~` U+007E	Tilde.
And Check can be configured in Unicode codes in `symbolCodes` it is just perefered format of codes that check is used for configuration.

smita1078:
Changes made:
Default value for `symbolCodes`: Changed to` "0x2190-0x27BF, 0x1F700-0x10FFFF"`
Removed `asciiOnly `property completely

Updated Javadoc:
Added Wikipedia reference link
Documented default ranges (`U+2190–U+27BF and U+1F700–U+10FFFF`)
Update violation message to : `Illegal symbol detected`

smita1078:
@vivek-0509 @romani please review

vivek-0509:
GitHub, generate website

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,234 @@
...
+     */
+    public static final String MSG_KEY = "illegal.symbol";
+
+    /** String Range Separator. */
+    private static final String RANGE_SEPARATOR = "-";
+
+    /** Specify the symbols to check for, as Unicode code points or ranges. */
+    private String symbolCodes = "0x2190-0x27BF, 0x1F700-0x10FFFF";
+
+    /**
```
vivek-0509, line 97: @romani 
The current default `0x2190-0x27BF, 0x1F700-0x10FFFF` misses the most commonly used emoji that AI tools typically insert in code comments.

According to the official doc:

[Emoticons (U+1F600-U+1F64F)](https://www.unicode.org/charts/PDF/U1F600.pdf) - Contains 😀😁 etc
[Transport and Map Symbols (U+1F680-U+1F6FF)](https://www.unicode.org/charts/PDF/U1F680.pdf) - Contains 🚀✈️ etc 

I think we should keep them as default, what is your opinion on this??

vivek-0509 (changes requested):

pom.xml
```diff
@@ -3717,6 +3718,13 @@
               <excludedTestClasses>
                 <param>*.Input*</param>
               </excludedTestClasses>
+              <avoidCallsTo>
+                <avoidCallsTo>org.apache.commons.logging</avoidCallsTo>
+                <avoidCallsTo>com.puppycrawl.tools.checkstyle.utils.UnmodifiableCollectionUtil</avoidCallsTo>
+              </avoidCallsTo>
+              <excludedClasses>
+                <param>com.puppycrawl.tools.checkstyle.checks.coding.IllegalSymbolCheck</param>
+              </excludedClasses>
```
vivek-0509: Check class should not be excluded from mutation testing, please kill all mutants through integration tests.

vivek-0509 (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,234 @@
...
+     * @param codePoint the code point to check
+     * @return true if in symbol codes
+     */
+    private boolean isInSymbolCodes(int codePoint) {
+        boolean found = false;
+        final String[] parts = symbolCodes.split(",", -1);
+
+        for (String part : parts) {
+            final String trimmed = part.trim();
+            if (trimmed.contains(RANGE_SEPARATOR)) {
```
vivek-0509: method parses the symbolCodes string on every character check .Consider parsing the ranges once in the setter and storing them as precomputed int[] pairs, so runtime checks are simple integer comparisons.

vivek-0509 (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    /**
+     * Parse a code point from string representation.
+     * Supports formats: 0x1234, \\u1234, U+1234, or decimal.
+     *
+     * @param str the string to parse
+     * @return the code point value
+     */
+    private static int parseCodePoint(String str) {
```
vivek-0509, line 238: method Javadoc says `or decimal `but the method parses all inputs as hex (radix 16).please update to 
/**
 * Parse a code point from string representation.
 * Supports formats: 0x1234, \\u1234, U+1234, or plain hex.

vivek-0509 (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,234 @@
...
+     * Format: comma-separated list of hex codes or ranges
+     * (e.g., "0x2705, 0xd83c-0xd83e").
+     *
+     * @param symbols the symbols specification
+     * @since 13.1.0
+     */
+    public void setSymbolCodes(String symbols) {
+        symbolCodes = symbols;
+    }
+
```
vivek-0509, line 123: Please add input validation in the setter so malformed config fails at load time, not during file processing

vivek-0509 (changes requested):

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_fi.properties
```diff
@@ -57,9 +58,9 @@ simplify.boolReturn=Konditionaalilogiikan voisi poistaa.
 simplify.expression=Ilmaisua voisi yksinkertaistaa.
 string.literal.equality=Kirjaimellinen Strings pitäisi verrata käyttämällä tasavertaisten (), ei ''{0}''.
 super.constructor.call=Tarpeeton kutsu superluokan rakentajalle ilman argumentteja.
-textblock.format.close=Tekstilohkon lopetuslainausmerkkien (""") on oltava uudella rivill�.
-textblock.format.open=Tekstilohkon aloituslainausmerkkien (""") on oltava uudella rivill�.
-textblock.vertically.unaligned=Tekstilohkon lainausmerkit eiv�t ole pystysuorassa linjassa.
+textblock.format.close=Tekstilohkon lopetuslainausmerkkien (""") on oltava uudella rivill�.
+textblock.format.open=Tekstilohkon aloituslainausmerkkien (""") on oltava uudella rivill�.
+textblock.vertically.unaligned=Tekstilohkon lainausmerkit eiv�t ole pystysuorassa linjassa.
```
vivek-0509: these lines appear unchanged but show as modified in the diff. This is likely due to encoding or line ending changes. please check

vivek-0509 (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
romani: Let's catch them.

romani: 0x1F700-0x1FFFFF

smita1078: Sure

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_fi.properties
smita1078: I did check the encoding and line endings, it seems correct UTF-8 and LF, still this is there in git diff, locally diff with master its not showing this difference..

pom.xml
smita1078: Though I Added multiple test files, still I can see 143 mutations surviving
Despite comprehensive test coverage, the mutations aren't being killed.  How should I proceed, create separate PR to kill mutations?

romani:
Pitest is left.

romani:
GitHub, generate website

romani:
Please make all examples to be same. Only violation messages are changing in examples.

Please add example to show how to enforce only ASCII symbols in source code.
I think we need for ASCII example , override of message for check , to show how user can make message user friendly. https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/5580ba6_20260211043555/config.html#Custom_messages

smita1078:
> Please make all examples to be same. Only violation messages are changing in examples.

Sure

smita1078:
> Pitest is left.

Wanted some help around this, Though I Added multiple test files, still I can see 143 mutations surviving
Despite comprehensive test coverage, the mutations aren't being killed. How should I proceed, create separate PR to kill mutations and generate java code that kills following the process.. or should i add them to suppression list and  create separate issue to fix all one by one..

romani:
Extension of pitest mutations is not allowed.
We did unreal effort to kill all of them at https://github.com/checkstyle/checkstyle/issues/12341
No step back.
If you don't know how to cover, just apply mutation to code, pitest just hit you that code is not required and can be simplified

smita1078:
> Extension of pitest mutations is not allowed. We did unreal effort to kill all of them at #12341 No step back. If you don't know how to cover, just apply mutation to code, pitest just hit you that code is not required and can be simplified

Understood! Will do Thanks

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_fi.properties
vivek-0509: Let it be then it exactly looks same

vivek-0509:
you can use this command to generate mutation report locally 
`mvn -Ppitest-coding-2 org.pitest:pitest-maven:mutationCoverage `

smita1078:
> you can use this command to generate mutation report locally `mvn -Ppitest-coding-2 org.pitest:pitest-maven:mutationCoverage `

Yes, I am using this cmd 

`mvn -e --no-transfer-progress -P"$PITEST_PROFILE" clean test-compile org.pitest:pitest-maven:mutationCoverage
`
 I did run locally the build was successful.. 143 mutations were there, but here I can see still 20 are left adding tests for those

smita1078:
Actually It turned out to be a configuration oversight on my side. I had added `IllegalSymbolCheck `to <`targetClasses`> in the `pitest-coding-2` profile, but forgot to include `IllegalSymbolCheckTest `in <`targetTests`>. As a result, PIT was mutating the class without executing its corresponding tests, which led to the 143 mutations.

smita1078:
> Pitest is left.

Done

smita1078:
> Please make all examples to be same. Only violation messages are changing in examples.

`Example4 `cannot be made structurally identical to the other examples because it demonstrates a fundamentally different code construct — a string literal containing an illegal symbol, whereas Examples 1, 2, and 3 all demonstrate violations inside comments with `int value = 1` as the field.

> Please add example to show how to enforce only ASCII symbols in source code. I think we need for ASCII example , override of message for check , to show how user can make message user friendly. https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/5580ba6_20260211043555/config.html#Custom_messages

Done

smita1078:
@romani , @vivek-0509  plz review

src/it/java/org/checkstyle/suppressionxpathfilter/coding/XpathRegressionIllegalSymbolTest.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import org.checkstyle.suppressionxpathfilter.AbstractXpathTestSupport;
+import org.junit.jupiter.api.Test;
+
+import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
+import com.puppycrawl.tools.checkstyle.checks.coding.IllegalSymbolCheck;
+
+public class XpathRegressionIllegalSymbolTest extends AbstractXpathTestSupport {
```
romani, line 32: please add one more test method, we need 3 different xpaths,  it will be soon mandatory by build execution.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,300 @@
...
+ * Property {@code symbolCodes} - Specify the symbols to check for, as Unicode code points
+ * or ranges. Format: comma-separated list of hex codes or ranges
+ * (e.g., {@code "0x2705, 0x1F600-0x1F64F"}). To allow only ASCII characters,
+ * use {@code "0x0080-0x10FFFF"}.
+ * Type is {@code java.lang.String}.
+ * Default value is {@code "0x2190-0x27BF, 0x1F600-0x1F64F, 0x1F680-0x1F6FF, 0x1F700-0x1FFFFF"}.
+ * </li>
+ * </ul>
+ *
+ * @since 13.3.0
```
romani: 13.4.0

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,300 @@
...
+    private static final String RANGE_SEPARATOR = "-";
+
+    /** Set of individual Unicode code points to disallow. */
+    private Set<Integer> singleCodePoints;
+
+    /** Set of Unicode ranges to disallow. */
+    private Set<CodePointRange> codePointRanges;
+
+    /** Specify the symbols to check for, as Unicode code points or ranges. */
+    private String symbolCodes = "0x2190-0x27BF, 0x1F600-0x1F64F, 0x1F680-0x1F6FF, "
```
romani: DEFAULT_ILLEGAL_CODES

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,300 @@
...
+     */
+    private record CodePointRange(int start, int end) {
+
+        /**
+         * Checks if code point is within range.
+         *
+         * @param codePoint code point to test
+         * @return true if within range; false otherwise
+         */
+        /* package */ boolean contains(int codePoint) {
```
romani: can we make it public ? or private ?

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
```diff
@@ -22,6 +22,7 @@ fall.through.last=Fall through from the last branch of the switch statement.
 final.variable=Variable ''{0}'' should be declared final.
 hidden.field=''{0}'' hides a field.
 illegal.catch=Catching ''{0}'' is not allowed.
+illegal.symbol=Illegal symbol detected.
```
romani: should we print illegal symbol ? it will be good for user,  especially if they are allow some but not others.
and it will be good for not ASCII, as some symbols looks similar to asciii, but not same.

romani (changes requested):
items:

src/it/java/org/checkstyle/suppressionxpathfilter/coding/XpathRegressionIllegalSymbolTest.java
smita1078: Sure

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages.properties
smita1078: Yes, it would be better to print the symbol as well as users can customize it

smita1078: image
Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
smita1078: Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
smita1078: Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
smita1078: Done

src/it/java/org/checkstyle/suppressionxpathfilter/coding/XpathRegressionIllegalSymbolTest.java
smita1078: Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,290 @@
...
+    /** Specify the symbols to check for, as Unicode code points or ranges. */
+    private String symbolCodes = DEFAULT_ILLEGAL_CODES;
+
+    /**
+     * Setter to specify the symbols to check for.
+     *
+     * @param symbols the symbols specification
+     * @throws IllegalArgumentException if the format is invalid
+     * @since 13.3.0
+     */
```
vivek-0509, line 103: Please update this to 13.4.0

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,290 @@
...
+ * use {@code "0x0080-0x10FFFF"}.
+ * Type is {@code java.lang.String}.
+ * Default value is {@code "0x2190-0x27BF, 0x1F600-0x1F64F, 0x1F680-0x1F6FF, 0x1F700-0x1FFFFF"}.
+ * </li>
+ * </ul>
+ *
+ * @since 13.4.0
+ */
+@FileStatefulCheck
+public class IllegalSymbolCheck extends AbstractCheck {
```
vivek-0509, line 74: this check does not change any field values at runtime,  only reads singleCodePoints and codePointRanges, which are set once in setSymbolCodes() / init(). According to the @StatelessCheck documentation `"The check does not contain mutable state, if it does not change any field values at the execution stage (but it still can set properties at the initialization stage)." `This matches our check exactly. Why are we using @FileStatefulCheck here, am I missing something?

vivek-0509 (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
smita1078: I initially considered `@StatelessCheck` for the same reason , `visitToken `only reads `singleCodePoints `and `codePointRanges`, never modifies them.
However `@StatelessCheck` has a stricter requirement enforced by ImmutabilityTest: all non-property fields must be immutable and final. 
Our derived fields:
```
private Set<Integer> singleCodePoints;
private Set<CodePointRange> codePointRanges;
```
are neither final nor immutable ,they are HashSet instances reassigned on every call to `parseSymbolCodes()`. Making them final is not possible because they are reassigned (not just mutated) in` parseSymbolCodes()`.
Using `@StatelessCheck` causes ImmutabilityTest to fail with:

```
[INFO] Running com.puppycrawl.tools.checkstyle.internal.ImmutabilityTest
[ERROR] Tests run: 4, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 2.421 s <<< FAILURE! -- in com.puppycrawl.tools.checkstyle.internal.ImmutabilityTest
[ERROR] com.puppycrawl.tools.checkstyle.internal.ImmutabilityTest.testClassesWithMutableFieldsShouldBeStateful -- Time elapsed: 0.009 s <<< FAILURE!
java.lang.AssertionError: 
Architecture Violation [Priority: MEDIUM] - Rule 'classes that do not have immutable fields and do not have modifier ABSTRACT should be annotated with @FileStatefulCheck or should be annotated with @GlobalStatefulCheck or should be suppressed' was violated (1 times):

[ERROR] com.puppycrawl.tools.checkstyle.internal.ImmutabilityTest.testFieldsInStatelessChecksShouldBeImmutable -- Time elapsed: 0.015 s <<< FAILURE!
java.lang.AssertionError: 
Architecture Violation [Priority: MEDIUM] - Rule 'fields that have name not containing '$' and are declared in classes that are annotated with @StatelessCheck and are not module properties should be among immutable types and should be final or should be suppressed' was violated (2 times):
```
Hence I changed it to  `@FileStatefulCheck`

romani:
GitHub, generate report

romani:
GitHub, generate website

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_fi.properties
romani: @smita1078 , please revert this lines to original state.
You can use some non fancy text editor to skip undesirable updates.

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,290 @@
...
+        parseSymbolCodes();
+    }
+
+    /**
+     * Parses the configured symbolCodes string into singleCodePoints and codePointRanges.
+     * Replaces both sets on every call so repeated invocations are safe.
+     *
+     * @throws IllegalArgumentException if format is invalid
+     */
+    private void parseSymbolCodes() {
```
romani: Please move this method to be below overriden methods

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    /**
+     * Checks text for illegal symbols, logging at most one violation per token.
+     *
+     * @param text the text content
+     * @param ast AST node associated with text
+     */
+    private void checkText(String text, DetailAST ast) {
+        text.codePoints()
```
romani: Please inline this method to visitToken

romani (changes requested):
Items

romani:
@vivek-0509 , please respond to your review items as "resolved" or something else to let me resolve in UI review items that are done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
vivek-0509: Resolved

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
vivek-0509: Resolved

pom.xml
vivek-0509: Resolved

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
vivek-0509: Resolved

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
vivek-0509: Resolved

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
vivek-0509: Resolved

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
vivek-0509: @smita1078  Thanks for the detailed explanation. ImmutabilityTest constraint is valid, but I think the fix should be to adjust the code to work with @StatelessCheck, not change the annotation to @FileStatefulCheck.
The fields can be made final by using clear() instead of reassignment

`Set<Integer>` will pass ImmutabilityTest automatically since Integer is in the allowed types. `Set<CodePointRange> `will still fail because CodePointRange is a custom record not in the IMMUTABLE_TYPES whitelist. For this, add it to SUPPRESSED_FIELDS_IN_MODULES in ImmutabilityTest, we have done this in the past also where fields are truly immutable at runtime.

https://github.com/checkstyle/checkstyle/blob/33235131ffdf9ab0c324bed3c2f6ac5bea732dfa/src/test/java/com/puppycrawl/tools/checkstyle/internal/ImmutabilityTest.java#L124-L133

@romani we need your opinion on this..

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
smita1078: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
smita1078: done

src/main/resources/com/puppycrawl/tools/checkstyle/checks/coding/messages_fi.properties
romani: @smita1078 , please revert this change

romani:
@smita1078 , please address https://github.com/checkstyle/checkstyle/pull/18609/changes#r2924395114

romani:
@smita1078 , do you know how to fix rendering of illegal symbols at your diff web site ?
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/45d51f8_2026021334/reports/diff/camunda/index.html#A1

smita1078:
> @smita1078 , please address https://github.com/checkstyle/checkstyle/pull/18609/changes#r2924395114

Done

smita1078:
> @smita1078 , do you know how to fix rendering of illegal symbols at your diff web site ? https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/45d51f8_2026021334/reports/diff/camunda/index.html#A1

I think when it writes the HTML diff report it is using Latin-1 encoding, causing the UTF-8 bytes to be misinterpreted. The fix would be to add `<meta charset="UTF-8"/>` in the generated HTML <head> section in the report generator

romani:
GitHub, generate website

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
romani: @smita1078 , @vivek-0509 
yes, lets extend and suppression list and make it Stateless

vivek-0509: @smita1078 Please make the changes. You may face a mutation survival issue with the clear() calls. If that happens, enclose both clear() calls in a separate private method, exclude that method from Pitest, and call it from parseSymbolCodes(). Ping me if any help needed

smita1078: Sure

smita1078: Done

pom.xml
```diff
@@ -3745,6 +3747,13 @@
               <excludedTestClasses>
                 <param>*.Input*</param>
               </excludedTestClasses>
+              <excludedMethods>
+                <excludedMethod>parseSymbolCodes</excludedMethod>
+              </excludedMethods>
```
vivek-0509: @smita1078 
i think you misunderstood, i did not mean to exclude the whole method, but after thinking more about it i think we do not need those clear calls they can be removed

vivek-0509 (changes requested):

pom.xml
smita1078: Got it!, moreover Pitest mutations survived which itself shows that removing the clear() calls changed nothing observable

vivek-0509: for our check yes the Pitest mutation survival confirms this, but  Pitest is not always right about this, for example in IndentationCheck, if we remove the clear() call and run Pitest, it would also show no observable behavior change because tests process files in isolation. But in reality, without clear(), handlers and incorrectIndentationLines would accumulate data across every file processed, eventually leading to OutOfMemoryError on large codebases. That is why for such cases we keep those clear calls in the  separate method and exclude that method form pitest and write a dedicated unit test that explicitly verifies the state is cleared between files, which both satisfies Pitest and proves the clear() is genuinely necessary.

smita1078: Agreed! Thanks for detailed explanation.
All done

romani (approved):
I am ok to merge!!
Thanks a lot for hard work.

romani:
@vivek-0509  please do you explicit approval  when all your items are addressed, to let me merge

pom.xml
vivek-0509: Resolved

config/checkstyle-checks.xml
```diff
@@ -431,6 +431,12 @@
                        java.lang.RuntimeException,
                        java.lang.NullPointerException"/>
     </module>
+    <module name="IllegalSymbol">
+      <property name="symbolCodes"
+                value="0x1F000-0x1FAFF,0x2600-0x26FF,0x2700-0x27BF,0xFE0F,0x200D"/>
```
vivek-0509: it has a different symbolCodes value than the  default .Is this intentional?

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,276 @@
...
+ * </p>
+ *
+ * <ul>
+ * <li>
+ * Property {@code symbolCodes} - Specify the symbols to check for, as Unicode code points
+ * or ranges. Format: comma-separated list of hex codes or ranges
+ * (e.g., {@code "0x2705, 0x1F600-0x1F64F"}). To allow only ASCII characters,
+ * use {@code "0x0080-0x10FFFF"}.
+ * Type is {@code java.lang.String}.
+ * Default value is {@code "0x2190-0x27BF, 0x1F600-0x1F64F, 0x1F680-0x1F6FF, 0x1F700-0x1FFFFF"}.
```
vivek-0509: Please correct the upper bound here also , see reason below

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,276 @@
...
+     * A key is pointing to the warning message text in "messages.properties" file.
+     */
+    public static final String MSG_KEY = "illegal.symbol";
+
+    /** Separator used for defining ranges. */
+    private static final String RANGE_SEPARATOR = "-";
+
+    /** Default symbol codes to check for. */
+    private static final String DEFAULT_ILLEGAL_CODES =
+        "0x2190-0x27BF, 0x1F600-0x1F64F, 0x1F680-0x1F6FF, 0x1F700-0x1FFFFF";
```
vivek-0509: you are using upper bound 0x1FFFFF in the default range this is incorrect, as per the Unicode Standard "U+10FFFF is associated with the largest legal UTF-32 32-bit code unit value" https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-23/. there are no Unicode characters beyond U+10FFFF, the range 0x1FFFFF does not exist in Unicode, the upper bound should be corrected to 0x10FFFF.

vivek-0509 (changes requested):

config/checkstyle-checks.xml
smita1078: goal was to catch most of the AI-generated symbols, but I think many are missing, will change to default

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
smita1078: Actually I referred this comment https://github.com/checkstyle/checkstyle/pull/18609#discussion_r2756766111, directly used  it from here , I will change it to the correct one.

smita1078: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
smita1078: done

config/checkstyle-checks.xml
smita1078: done

config/checkstyle-checks.xml
```diff
@@ -431,6 +431,12 @@
                        java.lang.RuntimeException,
                        java.lang.NullPointerException"/>
     </module>
+    <module name="IllegalSymbol">
+      <property name="symbolCodes"
+                value="0x2190-0x27BF, 0x1F600-0x1F64F, 0x1F680-0x1F6FF, 0x1F700-0x1FFFF"/>
+      <property name="tokens"
+                value="COMMENT_CONTENT,TEXT_BLOCK_CONTENT,IDENT,STRING_LITERAL,CHAR_LITERAL"/>
```
vivek-0509, line 438: @smita1078  it is still wrong, the upper bound is `Ux10FFFF`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,276 @@
...
+ *
+ * <ul>
+ * <li>
+ * Property {@code symbolCodes} - Specify the symbols to check for, as Unicode code points
+ * or ranges. Format: comma-separated list of hex codes or ranges
+ * (e.g., {@code "0x2705, 0x1F600-0x1F64F"}). To allow only ASCII characters,
+ * use {@code "0x0080-0x10FFFF"}.
+ * Type is {@code java.lang.String}.
+ * Default value is {@code "0x2190-0x27BF, 0x1F600-0x1F64F, 0x1F680-0x1F6FF, 0x1F700-0x1FFFF"}.
+ * </li>
```
vivek-0509, line 68: Same here, please use `Ux10FFFF`

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
```diff
@@ -0,0 +1,276 @@
...
+     */
+    public static final String MSG_KEY = "illegal.symbol";
+
+    /** Separator used for defining ranges. */
+    private static final String RANGE_SEPARATOR = "-";
+
+    /** Default symbol codes to check for. */
+    private static final String DEFAULT_ILLEGAL_CODES =
+        "0x2190-0x27BF, 0x1F600-0x1F64F, 0x1F680-0x1F6FF, 0x1F700-0x1FFFF";
+
```
vivek-0509, line 87: Same here, please use `Ux10FFFF`

vivek-0509 (changes requested):

vivek-0509:
@smita1078 Please rebase your branch

smita1078:
> @smita1078 Please rebase your branch

Done

config/checkstyle-checks.xml
smita1078: Apologies! Missed that '0'

vivek-0509:
GitHub, generate report

vivek-0509:
GitHub, generate website

config/checkstyle-checks.xml
vivek-0509: Resolved

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
vivek-0509: Resolved

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
vivek-0509: Resolved

config/checkstyle-checks.xml
vivek-0509: Resolved

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
vivek-0509: Resolved

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/IllegalSymbolCheck.java
vivek-0509: Resolved

vivek-0509:
GitHub, generate report

vivek-0509 (approved):
PR looks good to me 🚀
@smita1078 Thanks a lot for the hard work on this!
@romani Thank you for keeping me in the review cycle

PR #18611, opened 2026-01-12
Issue #18027: Resolve Pitest Suppressions - ant for java/util/Objects…

Youssefalaa7:
Issue #18027: Resolve Pitest Suppressions - ant
excluded getImplVersionForVersionInfo method from pom.xml

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4507,6 +4507,10 @@
               <excludedTestClasses>
                 <param>*.Input*</param>
               </excludedTestClasses>
+              <excludedMethods>
+                <!-- this method in CheckstyleAntTask.java works properly only when execution is from jar -->
+                <param>getVersionString</param>
+              </excludedMethods>
               <coverageThreshold>100</coverageThreshold>
               <mutationThreshold>92</mutationThreshold>
               <timeoutFactor>${pitest.plugin.timeout.factor}</timeoutFactor>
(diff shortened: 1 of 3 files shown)


romani:
Look around a history for
https://github.com/checkstyle/checkstyle/blob/cdee570c5d81efdef9b57dd2f2ca862d5effd91c/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L174

We do some hacking to let pitest skip this type of method

Youssefalaa7:
this method is excluded so i think we should also exclude execute method this will solve the issue

pom.xml
```diff
@@ -4507,6 +4507,9 @@
               <excludedTestClasses>
                 <param>*.Input*</param>
               </excludedTestClasses>
+              <excludedMethods>
+                <param>execute</param>
```
romani: Weird name.
We should put code in special method like other classes do and exclude very special method that unlikely match by name to any possible business method

Youssefalaa7:
Done

src/main/java/com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask.java
```diff
@@ -239,6 +239,17 @@ public void setProperties(File props) {
         properties = props.toPath();
     }
 
+    /**
+     * Gets implementation version for version info.
+     *
+     * @return implementation version for version info
+     */
+    public String getImplVersionForVersionInfo() {
```
romani: Please name it getVersionString, your name is hard to read.
And we already use this name in other places, let's name it same to show it is the  same reason

Youssefalaa7:
Done

PR #18624, opened 2026-01-12
Issue #18599: Showcasing UnrecognizedJavadocTag errors

aclfe:
Issue #18599: Showcasing `UnrecognizedJavadocTag` errors

Showcasing what changes can look like for files with UnrecognizedJavadocTag errors. Commented some to explain a possible solution

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -300,6 +300,7 @@
       -Xep:TruthAssertExpected:ERROR
       -Xep:TypeParameterUnusedInFormals:ERROR
       -Xep:UnnecessaryLambda:ERROR
+      -Xep:UnrecognisedJavadocTag:ERROR
       -Xep:UnsynchronizedOverridesSynchronized:ERROR
       -Xep:UnusedMethod:ERROR
       -Xep:VoidUsed:ERROR
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
@@ -28,7 +28,7 @@ import com.puppycrawl.tools.checkstyle.grammar.javadoc.JavadocCommentsLexer;
  * @see <a href="https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html">
  *     javadoc - The Java API Documentation Generator</a>
  */
-@SuppressWarnings("InvalidInlineTag")
+@SuppressWarnings({"InvalidInlineTag", "UnrecognisedJavadocTag"})
 public final class JavadocCommentsTokenTypes {
 
     /**
--- a/config/checkstyle-checks.xml
+++ b/config/checkstyle-checks.xml
@@ -299,7 +299,7 @@
     <module name="MissingOverrideOnRecordAccessor"/>
     <module name="PackageAnnotation"/>
     <module name="SuppressWarnings">
-      <property name="format" value="^((?!unchecked|deprecation|rawtypes|resource|InvalidInlineTag).)*$"/>
+      <property name="format" value="^((?!unchecked|deprecation|rawtypes|resource|InvalidInlineTag|UnrecognisedJavadocTag).)*$"/>
       <message key="suppressed.warning.not.allowed"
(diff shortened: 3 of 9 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
```diff
@@ -658,7 +658,7 @@ public final class JavadocCommentsTokenTypes {
      * |--LEADING_ASTERISK ->   *
      * |--TEXT ->
      * |--JAVADOC_INLINE_TAG -> JAVADOC_INLINE_TAG
-     * |   `--CODE_INLINE_TAG -> CODE_INLINE_TAG
+     * |   `--CODE_INLINE_TAG -> CODE_INLINE_TAG //error here. a fix can be to wrap them around <pre> tags. That will help with showcasing this multi lin diagram
```
romani: suppress.

aclfe: Done. I've added it onto pom.xml

pom.xml
```diff
@@ -300,6 +300,7 @@
       -Xep:TruthAssertExpected:ERROR
       -Xep:TypeParameterUnusedInFormals:ERROR
       -Xep:UnnecessaryLambda:ERROR
+      -Xep:UnrecognisedJavadocTag:OFF
```
romani: I meant suppression violations on class fields or whole class in worse scenario

romani (changes requested):
It should be Error for all except for such unfortunate cases.

pom.xml
aclfe: I misunderstood, I'll make changes

aclfe:
@romani 

I've pushed code to show you my different approaches. First I put `@SuppressWarnings("UnrecognisedJavadocTag")` to every warning error line. This allowed there to be no violation messages, but `./mvnw clean verify` fails since antrun gives these errors like these:

`[ERROR] [checkstyle] [ERROR] /home/kairav/Desktop/Current/checkstyle/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java:377:27: The warning 'UnrecognisedJavadocTag' cannot be suppressed at this location.            Only few javac warnings are allowed to suppress.            If try to suppress checkstyle/pmd/..... violation please do this in their config file.            If you try to suppress IntelliJ IDEA inspection,            please use javadoc block tag @noinspection [SuppressWarnings]` 

Then I tried this new method in configurationloader: 
https://github.com/checkstyle/checkstyle/blob/30a26157e119027d5c7fb109d3d1af55d5947745/src/main/java/com/puppycrawl/tools/checkstyle/ConfigurationLoader.java#L289-L292

It didn't remove the warnings when I ran the command, gave compilation errors (since the tag `@noinspection` itself is unrecognized) 

It seems the situation is contradicting each other? Do you suggest any other way or a suppression maybe?

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
```diff
@@ -28,7 +28,7 @@
  * @see <a href="https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html">
  *     javadoc - The Java API Documentation Generator</a>
  */
-@SuppressWarnings("InvalidInlineTag")
+@SuppressWarnings({"InvalidInlineTag", "UnrecognisedJavadocTag"})
```
romani, line 31: Learn from https://github.com/checkstyle/checkstyle/pull/18669/changes

On how others did suppression

romani (changes requested):
Items

aclfe:
@romani 

Ci errors are unrelated, shows connectivity error
Changes pushed, please take a look

romani:
Now we missed UnrecognizedJavadocTag:ERROR in pom.xml

Whole point of this activity was to activate this rule on Error level, to avoid leaking more bad  code

aclfe:
> Whole point of this activity was to activate this rule on Error level, to avoid leaking more bad  code

Yes indeed, I've made changes

PR #18640, opened 2026-01-14
infra: add JDK 21 setup to site.yml and diff-report.yml, disable Error Prone in equalsverifier CI job

vivek-0509:
Fixes missing JDK 21 setup in workflow files that run Maven/Groovy commands and disable Error Prone in equalsverifier CI job for JDK 21 compatibility

### Changes

- **site.yml**: Added `actions/setup-java@v5` with JDK 21 to `generate_site` job
- **diff-report.yml**: Added `actions/setup-java@v5` with JDK 21 to `make_report` job
- **error-prone.yml**: disable Error Prone in equalsverifier CI job for JDK 21 compatibility

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -885,7 +885,7 @@ no-error-equalsverifier)
   echo "Checkout target sources ..."
   checkout_from https://github.com/jqno/equalsverifier.git
   cd .ci-temp/equalsverifier
-  mvn -e --no-transfer-progress -Pstatic-analysis-checkstyle compile \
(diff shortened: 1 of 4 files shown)


vivek-0509:
@romani, the JAVA_TOOL_OPTIONS approach failed because -XDaddTypeAnnotationsToSymbol=true is a javac compiler argument, not a JVM option. The JVM rejected it with Unrecognized option, causing the build to fail immediately. Only remaining valid option i can see is to disable static analysis for this job.This will skip the problematic Error Prone step while still running Checkstyle verification on the project.
Should we move forward with this??
Screenshot 2026-01-14 at 12 31 16 PM

vivek-0509:
CI failure is unrelated

PR #18643, opened 2026-01-14
Issue #18023: Resolve PIT Suppression - filter

Gaoyan1999:
Issues: #18023

- Add a unit test to cover the accept cache behavior for FileContents.
- Ensure the filter does not re-tag suppressions for the same FileContents instance.
- Reproduce the pitest mutation scenario where getFileContents() is removed.


src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithNearbyCommentFilterTest.java
```diff
@@ -781,6 +781,26 @@ public void testTagsAreClearedEachRun() {
...
     }
 
+    @Test
+    public void testDoesNotRetagSameFileContents() {
+        final SuppressWithNearbyCommentFilter filter = new SuppressWithNearbyCommentFilter();
+        final FileContents contents = new FileContents(new FileText(new File("filename"),
+                Collections.singletonList("//SUPPRESS CHECKSTYLE ignore")));
+        final Violation violation = new Violation(1, null, null, null, null, Object.class, null);
+        final TreeWalkerAuditEvent auditEvent =
+                new TreeWalkerAuditEvent(contents, "filename", violation, null);
```
romani: Please use verifySuppressedWithParser

Gaoyan1999: Thanks for the comment.

My current understanding is that `verifySuppressedWithParser` may not kill this mutation, because it appears to only check `getFileContents() != currentContents` once (on the first violation). If the mutation removes getFileContents(), that condition would still seem to pass.

To cover this case, it might be necessary to add a more direct test that calls `accept()` twice on the same filter instance and verifies that the same FileContents is not re-tagged.

romani: I think we have such method already that can take few Input files for same config. If not, please create

Gaoyan1999: @romani 

I tried `verifySuppressedWithParser`, but I don’t think it can kill this mutation. That helper is mainly meant to assert the suppression result. Here, we need a test that verifies the lazy-load behavior, which means  `tagSuppressions()` inside `accept()` must **not** run twice.

Let me clarify how I got this point:

In `SuppressionCommentFilter.java`, PIT can mutate the code so that getFileContents() becomes null. 

```java
    @Override
    public boolean accept(TreeWalkerAuditEvent event) {
        boolean accepted = true;

        if (event.violation() != null) {
            // Lazy update. If the first event for the current file, update file
            // contents and tag suppressions
            final FileContents currentContents = event.fileContents();

            if (getFileContents() != currentContents) {
                setFileContents(currentContents);
                tagSuppressions();
            }
            if (matchesTag(event)) {
                accepted = false;
            }
        }
        return accepted;
    }
```
To kill that mutation, we need a test that calls accept() twice and verifies that the calling of `tagSuppressions()` happens only once.

---

When I tried using `verifySuppressedWithParser`, I created two violations, which does trigger the lazy-load logic successfully. Here is the code:

```java
    @Test
    public void testDoesNotRetagSameContent() throws Exception {
        final String[] suppressedViolationMessages = {
                "41:17: "
                        + getCheckMessage(AbstractNameCheck.class,
                        MSG_INVALID_PATTERN, "A1", "^[a-z][a-zA-Z0-9]*$"),
                "44:17: "
                        + getCheckMessage(AbstractNameCheck.class,
                        MSG_INVALID_PATTERN, "B1", "^[a-z][a-zA-Z0-9]*$"),
        };
        final String[] expectedViolationMessages = {
                "41:17: "
                        + getCheckMessage(AbstractNameCheck.class,
                        MSG_INVALID_PATTERN, "A1", "^[a-z][a-zA-Z0-9]*$"),
                "44:17: "
                        + getCheckMessage(AbstractNameCheck.class,
                        MSG_INVALID_PATTERN, "B1", "^[a-z][a-zA-Z0-9]*$"),
        };
        verifySuppressedWithParser(getPath("InputSuppressWithNearbyCommentFilterByNoRetag.java"),
                expectedViolationMessages, suppressedViolationMessages);
    }

```

The problem is that I don’t see how to assert the lazy-load behavior with verifySuppressedWithParser. Becuase, this function and the similiar functions in TestSupport.java are more like asserting the result, not for the internla bahavior. In my view, to verify lazy loading we need one of the following:

1. verify that tagSuppressions() is called only once, or

2. verify that the `tags` in `SuppressWithNearbyCommentFilter` remains the same objects after two calls to accept().


Happy to hear your thoughts.

romani: can it help to remove `setFileContents` and inline its code to visitToken ?

romani: if not, 
we have 
https://github.com/checkstyle/checkstyle/blob/b8d7e3056acfccba40f01ae56d5a13824526fd20/pom.xml#L4707

lazyLoad

https://github.com/checkstyle/checkstyle/blob/b8d7e3056acfccba40f01ae56d5a13824526fd20/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java#L411C26-L419

we can probably put this method to some util class and reuse ?
and update pitest profile of filter to not exclude  lazyLoad too.

Gaoyan1999: It’s hard to reuse the existing lazy-load code because the behavior is slightly different. In our case, we have a container (`WeakReference<FileContents>`) and a side-effect method (`tagSuppressions`).


I think I’ve found a more reasonable approach:


In the new test case, I simulate a natural scenario. I create a mock FileContents and call the target method (`filter.accept`). Then I capture the tags returned from each call. Finally, I verify that the tag returned on the first call and the second call is the same object. This way, we can ensure that retag is not triggered again.

romani: please do not resole conversations that I started.

main hint from me was that we can consider this pitest mutation as performance optimization and we already know that pitest do not like performance optimizations that are not covered by tests. Covering by tests performance optimization is very very unstable. So we can make trick, we will make some util method, name it very utilitary like we did with `lazyLoad` and we can exclude it in pitest.

Gaoyan1999: At first, I only wanted to reuse and encapsulate `lazyUpdate` function, but I couldn't execlude the `getUpdateContents`, because it is the augument to that function. 

So instead, I introduced `WeakReferenceHolder` to manage the object and reused it in both `SuppressWithNearbyCommentFilter` and `SuppressionCommentFilter`. Currently, this holder provides get and lazyUpdate. And I also think it's more reasonable because it manages the object (get and update) in a same class.  

I understand that the lazyUpdate design could be applied in other scenarios, not only with WeakReference. However, since it is only used in two places in the codebase, I didn’t want to overdesign it.

src/test/java/com/puppycrawl/tools/checkstyle/utils/WeakReferenceHolderTest.java
```diff
@@ -0,0 +1,204 @@
...
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package com.puppycrawl.tools.checkstyle.utils;
+
+import static com.google.common.truth.Truth.assertWithMessage;
+
+import org.junit.jupiter.api.Test;
+
+public class WeakReferenceHolderTest {
```
romani: we can have Util class that helps with business logic, but we do not allow unit test for it. All should be covered by Checks/Filters tests, to make sure there is no artificial situations that such code is covering.

can you simply remove this class?

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/utils/WeakReferenceHolder.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Updates the referenced object only if the new object is different
+     * from the currently referenced object. After updating, runs the
+     * specified callback if provided.
+     *
+     * @param newObject the new object to reference; may be {@code null}
+     * @param afterUpdate a callback to run after updating the reference; may be {@code null}
+     */
+    public void lazyUpdate(@Nullable T newObject, @Nullable Runnable afterUpdate) {
```
romani: can we make it NotNull for both ?

in what cases if can be nullable ? content of file is nullable sounds like impossible or leak of defect in loading of file.

Gaoyan1999: okay, I will do that. Previously I thought we could also use this function to clean the object and also allow do nothing after update. But given that we don't have this kind of logic in our codebase, i will make them NonNull in next commit.

src/test/java/com/puppycrawl/tools/checkstyle/utils/WeakReferenceHolderTest.java
Gaoyan1999: removed

Issue #16748, opened 2025-04-03
Javadoc check to avoid unneccesary linebreaks


SethFalco:
Would it be fine to have a check included for Javadoc's, that enforces that the full length of a line is used before line breaking?

Example violations, suppose a line length limit of 80.

```java
public class Foobar {

    /**
     * A very long Javadoc string, which will could easily
     * have been written in fewer lines than it was, but for
     * some reason linebreaks were thrown in early. Perhaps
     * because of personal editor preferences, habits, or
     * a lack of awareness/feedback in their tooling.
     */
    public void foo() {
        System.out.println("bar");
    }
}
```

Example fix:

```java
public class Foobar {

    /**
     * A very long Javadoc string, which will could easily have been written in
     * fewer lines than it was, but for some reason linebreaks were thrown in
     * early. Perhaps because of personal editor preferences, habits, or a lack
     * of awareness/feedback in their tooling.
     */
    public void foo() {
        System.out.println("bar");
    }
}
```

In projects I maintain, the rules I informally apply are:

* A paragraph should not contain line breaks until a word would cross the 80th column.
* A paragraph should not cross the 80th column, unless a single word, link, or Javadoc tag us too long to fit.

I'm willing to try to contribute this myself. Just wanted to ask about it first in case I simply missed the check in the documentation, or if this is achievable with a combination of other rules.

romani:
Conceptually I am ok with general idea, but we need to share more details on design and behavior to approve issue and be confident in merge when PR is provided.

Please read https://checkstyle.org/report_issue.html#How_to_request_new_Check.2FModule.3F there are links to examples of how proposal should looks like. Please share all required details, and we will proceed with discussing, approvals and guidance on how to implement it.

romani:
Please share more examples on paragraphs.

SethFalco:
@romani Updated! Thank you for your patience. Is this any better?

I've also replaced the mock example with real ones, except for the `notAViolation1` demo which is a mock since I don't have real-world case for that scenario.

romani:
If we consider as main idea of Check to utilize whole line, to keep comment compact, so it means we should apply it to block tags of javadoc.

```
* @return
     *     Date which the company was converted / closed or dissolved. Refer to
     *     {@link #getCompanyStatus()} to determine which.
```

Should be violated.
If you can share specific reasons to not do this on javadoc block tags, please share it, it will be good reasoning for Check property, to let users choose to validate or not validate block tags.


[...]

PR #18646, opened 2026-01-14
Issue #16748: JavadocUtilizingTrailingSpace check

aclfe:
Issue #16748: JavadocUtilizingTrailingSpace check

----
The Core Goal

The objective is to enforce efficient use of horizontal space in Javadocs.

- Prevent premature line breaks: Text should not break to a new line until it reaches the configured lineLimit.
    - This value defaults to the max property of the project's LineLength check (standard default is 80), unless explicitly configured otherwise in this check.
- Prevent excessive line length: Text should not exceed the lineLimit unless absolutely necessary (e.g., long URLs).

> "As the goal of this Check is to only enforce the line length, but not the formatting of a Javadoc which may be covered by other Checks"

----
Proposed Rules

Goal: The Check focuses strictly on line length enforcement (filling the available horizontal space). It explicitly avoids enforcing stylistic choices (like indentation) that should be handled by other Checks.

1. Scope (What is Enforced)
The Check applies validation to:
    - Javadoc description lines containing content.
    - Lines containing Javadoc block tag values (e.g., the text following @param).

2. Exclusions (What is Ignored)
To prevent false positives and allow valid formatting, the Check ignores:
    - Empty Space: Blank lines with no content.
    - Indentation: Leading whitespace characters (e.g., indented HTML or tag values).
    - Tag Headers: Lines containing only the Javadoc block tag name (e.g., @param, @return, @see), allowing a break before the value starts.
    - Code Blocks: **All** content inside HTML `<pre>` tags.
    - Unbreakable Content:
        - Lines that start with a Javadoc tag or URL are allowed to exceed the lineLimit.
            - **To Note**: If a long tag/URL appears in the middle of a line and exceeds the limit, it is a violation. The user must move it to the start of the next line.
    - HTML Structural Tags: Lines strictly starting with an HTML tag (e.g., `<p>, <ul>, </li>`) are completely ignored by this check.
        - Reasoning: These are treated as user formatting/structure.
        - Note: If an HTML tag appears in the middle of a line of text, that line is not ignored and is validated as normal.

Contribution repo PR: https://github.com/checkstyle/contribution/pull/1043
New module config: https://raw.githubusercontent.com/aclfe/configs/refs/heads/main/jdoctrailingspace.xml


aclfe:
Pushed code for confirming direction. 

The logic itself in`JavadocUtilizingTrailingSpaceCheck` works as intended (as layed out in https://github.com/checkstyle/checkstyle/pull/18646#issue-3814347908). Please let me know if any rules need to be changed.

Future updates will update documentation and add many tests cases and examples. The check needs intensive documentation

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpace.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public void htmlStructuralIgnored() { }
+
+    // violation 3 lines below 'Line is smaller than 80 characters (found 14).'
+    // violation 3 lines below 'Line is smaller than 80 characters (found 39).'
+    /**
+     * @return
+     * This line is intentionally short
+     * and should be reported as too short.
+     */
```
romani: >(found 14)

violation message is not very clear. 
we need to hint user that words from next line should be moved to next line.
We can probably print exact string from next line that can fit previous.


 > // violation 3 lines below 'Line is smaller than 80 characters (found 39).'

we should skip violating consecutive lines, as fix in previus line will result in changes on next.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpace.java
```diff
@@ -0,0 +1,74 @@
...
+     * {@link com.very.long.package.name.that.exceeds.limit.CompanyStatus}
+     */
+    public void longInlineTagAtStartAllowed() { }
+
+    // violation 2 lines below 'Line is smaller than 80 characters (found 59).'
+    /**
+     * inline {@link long.package.name.limit.CompanyStatus}
+     * reference in the middle and should be too long. 
+     */
+    public void longInlineTagInMiddleViolation() { }
```
romani: let make it as pattern, under each method+java doc with violation
you place  kind of same method+java but fixed, it will be very clear what is expected.

romani (changes requested):
brief review:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpace.java
aclfe: > violation message is not very clear

I see what you mean, the messages seem very similar to LineLimit Check. 

Do you think it's better the message is something along these lines?

- `"Javadoc line under-utilized (39/80 chars). Consider adding from next line: 'and should be reported'"` 
- `"Line under-utilized (39/80). Words from below could be moved up."`

Second one seems more concise and less susceptible to line length errors.

aclfe: > we should skip violating consecutive lines, as fix in previus line will result in changes on next

perfectly fine, will edit pr's explanation and make that change

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpace.java
aclfe: Do you think it's cleaner if I put the expected/corrected versions in a separate file? 
Like InputJavadocUtilizingTrailingSpaceExpected.java?

or should we consider the readability factor

aclfe:
@romani Running `./mvnw clean verify` shows the new check triggers violations across 
the codebase. I'm happy to fix them all, but wanted to inquire since it'll lengthen the PR significantly. It might be better to have a follow up PR to cleanup the violations and initially suppress them? What do you think is optimal?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpace.java
romani: >Line under-utilized (39/80). Words from below could be moved up.

I like this

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpace.java
romani: InputJavadocUtilizingTrailingSpaceExpected.java not present.

romani: Do you think it's cleaner if I put the expected/corrected versions in a separate file?

No. User like when it is close to each other. So they immediately see how is good

romani:
Let's not fix our javadoc for now.
Suppression all violations.
We will see them in diff report.

I don't share ideas that javadoc needs to be compact.

romani:
@aclfe , do you plan to finish this PR ?

aclfe:
> @aclfe , do you plan to finish this PR ?

The logic and check is finished, only suppressions on build time remain. I've made all the changes requested too. Only External project violations and other violations need to be suppressed

aclfe:
Github, generate website

aclfe:
GitHub, generate site

aclfe:
GitHub, generate report

aclfe:
Github, generate report

aclfe:
The script ran for 20+ minutes and failed on an internal issue 
```
Running command: git fetch origin 3.x:3.x
From https://github.com/ReactiveX/RxJava
 * [new branch]          3.x        -> 3.x
Resetting git sources to commit '3.x'
Running command: git reset --hard refs/tags/3.x

fatal: ambiguous argument 'refs/tags/3.x': unknown revision or path not in the working tree
```

I'll run it again to see if the issue persists, otherwise the diff script might need changes

aclfe:
Github, generate report

aclfe:
Same issue. Diff script might need changes

aclfe:
Github, generate report

aclfe:
Github, generate report

aclfe:
Github, generate report

mahfouz72:
@aclfe can you make CI green? Let's continue this PR

aclfe:
on it @mahfouz72

aclfe:
@mahfouz72 ready for review, pitest common flaked. needs a re run

mahfouz72:
Github, generate report

config/checkstyle-checks.xml
```diff
@@ -41,6 +41,21 @@
...
+  <!-- Javadoc checks are not enforced on external project test code -->
+  <module name="SuppressionSingleFilter">
+    <property name="checks" value="JavadocUtilizingTrailingSpace"/>
+    <property name="files" value=".*[\\/]src[\\/](test|it|xdocs-examples)[\\/]"/>
+  </module>
+  <!-- Own main source Javadocs are not required to comply yet -->
+  <module name="SuppressionSingleFilter">
+    <property name="checks" value="JavadocUtilizingTrailingSpace"/>
+    <property name="files" value=".*[\\/]src[\\/]main[\\/]java[\\/]"/>
+  </module>
```
mahfouz72, line 58: We normally send commit to fix our codebase violations and the violation in the external repo that causes `no-error-xxx` to fail. can we do that? why did you think we should supress that? at least for our repo

config/checkstyle-resources-suppressions.xml
```diff
@@ -144,6 +144,10 @@
   <suppress id="noTrailingWhitespace"
     files="javadoctagcontinuationindentation/InputJavadocTagContinuationIndentationDescription\.java"/>
 
+  <!-- intentional problem for testing -->
+  <suppress id="JavadocUtilizingTrailingSpace"
+    files="src[\\/]test[\\/]resources[\\/].*[\\/]Input.*\.java"/>
```
mahfouz72, line 149: Agreed with this, this check should not run for input files at all.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpaceTooLong.java
```diff
@@ -0,0 +1,89 @@
...
+    public void correctedSecondLineTooLong() { }
+
+
+    /**
+     * @param value a parameter that has an extremely long description that goes well beyond the eighty character limit here
+     */
+    public void paramTooLong(int value) { }
+
+    /**
+     * @return returns a value with a description that is way too long and exceeds the configured line length limit significantly
```
mahfouz72, line 60: Why we are not flaging this as too long, some words can move to next line.

pom.xml
```diff
@@ -2117,6 +2117,7 @@
                 <exclude>**/SingleLineJavadocCheck.class</exclude>
                 <exclude>**/SummaryJavadocCheck.class</exclude>
                 <exclude>**/WriteTagCheck.class</exclude>
+                <exclude>**/JavadocUtilizingTrailingSpaceCheck.class</exclude>
```
mahfouz72, line 2120: I don't remember why do we need to do this?

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocUtilizingTrailingSpaceCheck.java
```diff
@@ -0,0 +1,420 @@
...
+    }
+
+    /**
+     * Processes a single node in the Javadoc tree.
+     *
+     * @param node the node to process
+     */
+    private void processNode(DetailNode node) {
+        switch (node.getType()) {
+            case JavadocCommentsTokenTypes.NEWLINE -> handleNewline(node);
```
mahfouz72, line 184: I am confused here, this is called by `beginJavadocTree` where it only called on root node, then how can we reach all these node types that we have in switch case.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocUtilizingTrailingSpaceCheck.java
```diff
@@ -0,0 +1,420 @@
...
+     * Processes all children of a node.
+     *
+     * @param node the parent node
+     */
+    private void processChildren(DetailNode node) {
+        for (DetailNode child = node.getFirstChild(); child != null;
+                child = child.getNextSibling()) {
+            processNode(child);
+        }
+    }
```
mahfouz72, line 206: I feel we are doing a mini tree walker here and using custom traversal inside the check, cant we utlize the AbstractJavadocTraversal and use `visitJavadocToken`

mahfouz72 (changes requested):
Thanks a lot for starting this!
Some items from first pass:

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocUtilizingTrailingSpaceCheck.java
```diff
@@ -0,0 +1,420 @@
...
+     */
+    private void validateLines() {
+        for (int index = 0; index < lines.size(); index++) {
+            final JavadocLine line = lines.get(index);
+
+            if (!line.hasContent) {
+                continue;
+            }
+
+            if (line.length > lineLimit && !line.startsWithUnbreakable) {
```
mahfouz72, line 132: I believe we could remove this condition `!line.startsWithUnbreakable`, what if it starts with block tag or url but still too long with texts we can break some text words here. why do we have this?

mahfouz72:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/diff/d3d4440_2026220607/MaterialDesignLibrary/index.html#A1

> Got an exception - java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 7
#x9;at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
#x9;at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
#x9;at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
#x9;at java.base/java.util.Objects.checkIndex(Objects.java:385)
#x9;at java.base/java.util.ArrayList.get(ArrayList.java:427)
#x9;at com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocUtilizingTrailingSpaceCheck.isTooShort(JavadocUtilizingTrailingSpaceCheck.java:168)
#x9;at com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocUtilizingTrailingSpaceCheck.validateLines(JavadocUtilizingTrailingSpaceCheck.java:137)

please review report

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpaceClassLevel.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * A constant value used throughout the application for configuration
+     * purposes and default behavior settings.
+     */
+    public static final int CONSTANT = 42;
+
+    // violation 2 lines below 'Line under-utilized (31/80). Words from below could be moved up'
+    /**
+     * Field documentation that
```
mahfouz72, line 30: Can you add example where the current line start with html tags? will we still get violation?

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocUtilizingTrailingSpaceCheck.java
```diff
@@ -0,0 +1,420 @@
...
+        switch (node.getType()) {
+            case JavadocCommentsTokenTypes.NEWLINE -> handleNewline(node);
+            case JavadocCommentsTokenTypes.TEXT -> handleText(node);
+            case JavadocCommentsTokenTypes.TAG_NAME -> handleTagContent();
+            case JavadocCommentsTokenTypes.JAVADOC_INLINE_TAG -> handleInlineTag(node);
+            default -> {
+                if (node.getType() != JavadocCommentsTokenTypes.HTML_ELEMENT) {
+                    processChildren(node);
+                }
+            }
```
mahfouz72, line 192: Is this means we are skipping html elements at all so example like this will not be violation?

```  
 <li><strong>Static Configuration:</strong> Uses configuration files either in the classpath or externally, and {@link Annotation}s.</li>
```

mahfouz72 (commented):
items regarding html tags

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocUtilizingTrailingSpaceCheck.java
aclfe: Yeah exactly! 

"HTML Structural Tags: Lines strictly starting with an HTML tag `(e.g., <p>, <ul>, </li>) `are completely ignored by this check" 

 original comment: https://github.com/checkstyle/checkstyle/issues/16748#issuecomment-2798387394 
> Better to say lines that starts with html tags will be ignored, as it might imply user formatting. But if any tag in middle of text , it will be under validation.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpaceClassLevel.java
aclfe: some examples: https://github.com/aclfe/checkstyle/blob/d3d44402a5eeb7665ebacc015c796f864d49db07/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpaceMultiParagraph.java#L15-L51

https://github.com/aclfe/checkstyle/blob/d3d44402a5eeb7665ebacc015c796f864d49db07/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpaceClassLevel.java#L13-L14

But I would say **I think I missed something**. I think https://github.com/aclfe/checkstyle/blob/d3d44402a5eeb7665ebacc015c796f864d49db07/src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpaceHtmlTags.java#L15-L20 needs a violation. ```The rules say **Lines strictly starting with an HTML tag `(e.g., <p>, <ul>, </li>)` are completely ignored by this check.
** this is infact the SECOND line that doesn't START with <p>``` so... I would say this needs to be fixed on my part. I went too harsh on the code

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocUtilizingTrailingSpaceCheck.java
aclfe: I think this might conflict with the rule "Lines that start with a Javadoc tag or URL are allowed to exceed the lineLimit" ?

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocUtilizingTrailingSpaceCheck.java
aclfe: I think this goes back to https://github.com/checkstyle/checkstyle/pull/18646#discussion_r4087886762 . I really see this point. Maybe a small helper that walks up from a node through getParent()? I think this is the best approach that match with our cleaner code need, I need your thought on what you think is more optimal the manual or the suggested way?

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocUtilizingTrailingSpaceCheck.java
aclfe: The switch is reachable, there's a huge corpus of them that are testing it or else I wouldn't have passed coverage. but you're mentioning a good point , I would say that AbstractJavadocCheck doesn't handle our situation well? 

For example I remember before that HTML_ELEMENT is skipped, so the TAG_NAME in <p> never hits the TAG_NAME case. Without the skip, every HTML line would be marked startsWithUnbreakable.

pom.xml
aclfe: I don't remember why I did this either, I think it's safe to remove. let me remove next push

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpaceTooLong.java
aclfe: This was explicitly a design choice in the original discussion, I put it in the description too so I can remember all the edge cases. "Lines that start with a Javadoc tag or URL are allowed to exceed the lineLimit." but here's the orignal comment: https://github.com/checkstyle/checkstyle/issues/16748#issue-2969533702 and the specific line that tells us the rule 
> Lines that start with a Javadoc tag or URL, without any additional content, yet are still too longer than lineLimit

config/checkstyle-checks.xml
aclfe: https://github.com/checkstyle/checkstyle/pull/18646#issuecomment-3874495481

I mean it was basically this comment that I was following. And I get your point on fixing the external repo, but the nature of this check is that it's very intense and gives lots and lots of violations. Which becomes more of a debt imo rather than a "fix".

aclfe:
Let me address https://github.com/checkstyle/checkstyle/pull/18646#issuecomment-5804154531 and investigate this. Thank you so much for the review, I'll push asap @mahfouz72

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpaceClassLevel.java
mahfouz72: Hmm, but text inside pre tag amd code for example should still be ok even if long, correct? Can we see the issue comments what was agreed about

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocUtilizingTrailingSpaceCheck.java
mahfouz72: You can use `visitJavadocToken` and [leaveJavadocToken](https://github.com/checkstyle/checkstyle/blob/92d85004ebaed1cc8fb23e2b7dbd7211ffce2767/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck.java#L262-L270) to achieve the same behavior.

For example, when visiting an `HTML_ELEMENT`, you can use a boolean flag to indicate that we are currently inside an HTML element. While that flag is set, skip processing all other nodes inside the element. Then, in `leaveJavadocToken`, reset the flag when leaving the `HTML_ELEMENT`, so the next `TEXT` or `NEWLINE` token can be processed normally.

Then, inside `visitJavadocToken`, you can have a branch for each relevant token type and keep the existing processing logic there as you have now.

This allows you to reuse the standard `visitJavadocToken` / `leaveJavadocToken` traversal instead of manually traversing the Javadoc tree.

Let me know if there is anything unclear.

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocUtilizingTrailingSpaceCheck.java
mahfouz72: > Lines that start with a Javadoc tag or URL are allowed to exceed the lineLimit

Who sets this rule can you please point to it from issue comments?

src/test/resources/com/puppycrawl/tools/checkstyle/checks/javadoc/javadocutilizingtrailingspace/InputJavadocUtilizingTrailingSpaceTooLong.java
mahfouz72: > without any additional content

my understanding is that it have only the URL or only the block tag in the same line it should be skipped, am I missing something? is there any test cases from issue comments that shows this?

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocUtilizingTrailingSpaceCheck.java
mahfouz72: okay, agreed with this if starts with html element we ignore it.

config/checkstyle-checks.xml
mahfouz72: Agreed.

Issue #18030, opened 2025-11-01
Resolve Pitest Suppressions - coding-2


romani:
As a part of this issue, we have to kill the mutation listed below and remove the suppression from [pitest xxxx suppression.xml](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/)

To understanding how to kill the mutation 
Visit [Wiki page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports), [Pitest-docs](https://pitest.org/) and expected actions and workflow at https://github.com/checkstyle/checkstyle/issues/12341.

Target is to remove all `<mutation` from [Pitest-coding-2](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-coding-2-suppressions.xml)

DakshRJain737:
@romani The issue that you stated,Can it really be done because i tried but was not able to frame a solution.
Can you please help me with this?

romani:
It is not easy for sure, try to choose another pitest issue another mutation.
Look around in our repository, we already fixed hundreds of such mutations, in such PRs you can see how we cracked them.

Anushreebasics:
@Aman-Baliyan are u working on this issue, or should I take it up?

Aman-Baliyan:
> [@Aman-Baliyan](https://github.com/Aman-Baliyan) are u working on this issue, or should I take it up?

@Anushreebasics you can work on these mutations


```
<mutation unstable="false">
    <sourceFile>UnusedLocalVariableCheck.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck</mutatedClass>
    <mutatedMethod>isInsideLocalAnonInnerClass</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NegateConditionalsMutator</mutator>
    <description>negated conditional</description>
    <lineContent>if (currentAst.getType() == TokenTypes.SLIST) {</lineContent>
  </mutation>

  <mutation unstable="false">
    <sourceFile>UnusedLocalVariableCheck.java</sourceFile>
    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.coding.UnusedLocalVariableCheck</mutatedClass>
    <mutatedMethod>isInsideLocalAnonInnerClass</mutatedMethod>
[...]

PR #18649, opened 2026-01-15
Issue #18030: Resolve Pitest Suppressions - coding-2

Youssefalaa7:
Issue #18030: Resolve Pitest Suppressions - coding-2
checked that only local methods should be included in anonInnerClassHolders

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheckTest.java
@@ -581,4 +581,58 @@ public class UnusedLocalVariableCheckTest extends AbstractModuleTestSupport {
                 getPath("InputUnusedLocalVariableAnonInnerClasses3.java"),
                 expected);
     }
+
+    /**
+     * Use TestUtil.invokeStaticMethod to access the private static method
+     * isInsideLocalAnonInnerClass to test this because it produces optimization
+     * mutations that we can not kill using verifyWithInlineConfigParser.
+     */
+    @Test
+    public void testIsInsideLocalAnonInnerClass() throws Exception {
+        final DetailAST root = JavaParser.parseFile(
+                new File(getPath("InputUnusedLocalVariableLambdaAnonInner.java")),
+                JavaParser.Options.WITHOUT_COMMENTS);
+
+        final DetailAST fieldObj = TestUtil.findTokenInAstByPredicate(root,
+                        ast -> {
+                            return ast.getType() == TokenTypes.VARIABLE_DEF
+                                    && "fieldObj".equals(ast.findFirstToken(TokenTypes.IDENT)
+                                    .getText());
+                        })
+                .orElseThrow();
+        final DetailAST literalNewField = fieldObj.findFirstToken(TokenTypes.ASSIGN)
+                .findFirstToken(TokenTypes.EXPR)
+                .findFirstToken(TokenTypes.LITERAL_NEW);
+
+        final boolean resultFalse = TestUtil.invokeStaticMethod(
+                UnusedLocalVariableCheck.class,
+                "isInsideLocalAnonInnerClass",
+                Boolean.class,
+                literalNewField);
+
+        assertWithMessage("Should be false for field initialization (no SLIST in ancestry)")
+                .that(resultFalse)
+                .isFalse();
+
+        final DetailAST localObj = TestUtil.findTokenInAstByPredicate(root,
+                        ast -> {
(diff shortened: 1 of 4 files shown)


src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheckTest.java
```diff
@@ -581,4 +581,43 @@ public void testUnusedLocalVarInAnonInnerClasses3() throws Exception {
...
+        assertWithMessage("Should find local anonymous inner class")
+                .that(localNew.isPresent())
+                .isTrue();
+
+        check.beginTree(root);
+
+        final DetailAST classDef = root.findFirstToken(TokenTypes.CLASS_DEF);
+        check.visitToken(Objects.requireNonNull(classDef));
+        check.visitToken(fieldLevelNew.orElseThrow());
+        check.visitToken(localNew.orElseThrow());
```
romani: Only verifyWithInlineConfigParser is allowed to cover mutation

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheckTest.java
Youssefalaa7: I think this mutation can't be handled with verifyWithInlineConfigParser because the local methods and unlocal methods produce the same violation they only differ in implementation and ordering and for this current test I saw a similar test to this in the same file
What do you suggest we can do

romani: I need more details on why mutation can not be killed. 
If can not be killed, we can apply it as it means this code is not required.

Youssefalaa7:
Done i will work on the two remaining mutations on a different pr

romani:
@Youssefalaa7  , main hint from me was that we can consider this pitest mutation as performance optimization and we already know that pitest do not like performance optimizations that are not covered by tests. Covering by tests performance optimization is very very unstable. So we can make trick, we will make some util method, name it very utilitary like we did with `lazyLoad` and we can exclude it in pitest.

Youssefalaa7:
Done

romani:
Github, generate report for UnusedLocalVariable/all-examples-in-one

pom.xml
```diff
@@ -3713,6 +3713,10 @@
               <excludedTestClasses>
                 <param>*.Input*</param>
               </excludedTestClasses>
+              <excludedMethods>
+                <!-- performance optimization for early exit from loop -->
+                <param>hasSlistParent</param>
```
romani: as we are failing to cover it in real inputs way.
And we are failing to make logic to have pure util method that in name have definition of performance.
with such config it can exclude any other method that matched by name, and we might have a bunch of such methods.

lets simply make pure unit  test for this method in  UnusedLocalVariableCheckTest

method should stay private in Check, test method will use reflection to get access to it, it will run on artificial AST to kill this final mutation. Over test method there will be wide comment that explains why w do hack-testing approach.

Eventually code of this Check will be changed, and this method might disappear  so problem will self resolve or be isolated to Check.

I hope eventually this year we will add support for "JDK 25 finalized the feature for
Instance Main Methods (JEP 512)" and we will be able to write Input wihout Type so we will be back to naturally coverage by compiled Inputs.

romani (changes requested):
thanks a lot!

last update before merge:

Youssefalaa7:
Done

src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/UnusedLocalVariableCheckTest.java
```diff
@@ -581,4 +581,56 @@ public void testUnusedLocalVarInAnonInnerClasses3() throws Exception {
                 getPath("InputUnusedLocalVariableAnonInnerClasses3.java"),
                 expected);
     }
+
+    @Test
+    public void testIsInsideLocalAnonInnerClass() throws Exception {
+        final DetailAST root = JavaParser.parseFile(
+                new File(getPath("InputUnusedLocalVariableLambdaAnonInner.java")),
+                JavaParser.Options.WITHOUT_COMMENTS);
+        // Use TestUtil.invokeStaticMethod to access the private static method
```
romani: Please move comment to be above method.
To state that we can not use verifyxxxx methods to kill mutation that is caused by performance optimization and ....

Youssefalaa7:
Done

Issue #18494, opened 2026-01-03
New Check: GoogleNonConstantFieldNameCheck to enforce Google Java Style Guide member naming


vivek-0509:
split from https://github.com/checkstyle/checkstyle/issues/17842 , we split creation of module and activation of it in google style config to separate issues, to simplify integration.

Create a new check GoogleNonConstantFieldNameCheck to enforce non-constant field naming conventions per the [Google Java Style Guide §5.2.5](https://google.github.io/styleguide/javaguide.html#s5.2.5-non-constant-field-names).

## Background (Reason for new check)

The current MemberName check with regex configuration cannot fully enforce Google Style non-constant field naming rules because it cannot:

1. Correctly enforce Google-specific underscore placement rules
2. Support multipart numeric version suffixes (for example guava33_4_5)
3. Reject special-prefix (Hungarian-style) names such as mValue

## Proposed Rules (Non-constant fields only)

- Field names must follow lowerCamelCase
- Single-character field names are not allowed
- Special prefix names (for example mValue, sCount, kSize) are not allowed, per [Google Java Style Guide §5.1](https://google.github.io/styleguide/javaguide.html#s5-naming)
- Underscores are allowed only between adjacent digits (multipart version or numbering suffix)
- Leading, trailing or double underscores are not allowed

## Scope

This check applies only to non-constant fields:

- Instance fields

## Explicitly out of scope

- Constant fields (static final)
- Local variables
- static fields that are not static final

### Valid:

```
userName
timeoutMs
foo123
guava33_4_5
kotlinVersion1_9_24
```
### Invalid:

```
f                                    // violation, single character
mValue                         // violation,special prefix not allowed
sCount                         // violation, special prefix not allowed 
foo_bar                        // violation, underscore between letters
gradle_9_5_1               // violation, underscore between letter-digit
kotlin_version1_9_24  // violation, underscore between letters
guava_33_4_5             // violation, underscore between letter-digit
guava33__4_5             // violation, consecutive underscores
guava33_4_5_             // violation, trailing underscore
_foo                               // violation, leading underscore
```

vivek-0509:
i am on it

romani:
Check is merged.
Activation in Google style will be separate issue.

vivek-0509:
@romani Should i reference this issue https://github.com/checkstyle/checkstyle/issues/17842  or should i create a new one for activation of the check in google style ??

romani:
Please use #17842

Issue #17842, opened 2025-09-29
False-negative: Member names with underscores


mohitsatr:
From: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case
> In very rare circumstances (for example, multipart version numbers), you may need to use underscores to separate adjacent numbers, since numbers do not have upper and lower case variants.

Google style is not exact. On one hand it says `_` can be used in test methods to combine sentences and on another hand in general CamelCase explanation, `_` is allowed for numbering suffix in all names(except package and generic names).

Valid uses of underscore :
```
1. guava33_4_5
2. guavaVersion33_4_5 
3. guavaversion33_4_5 (same as 1.)
```

Invalid uses of underscore:
```
1. guava_33_4_5_ (underscore at the end)
2. guava_33_4_5   (underscore between digit and letter)
3. guava_version33_4_5 (underscore between lowercase character sequences)
```

Some false-negatives in member names left to address from #17708  

```java
$ cat NamesWithUnderscores.java
/** some javadoc. */
public class NamesWithUnderscores {
  int gradle9_5_1 = 0;

  int gradle9_5_1_ = 0; // ok

  int gradle_9_5_1 = 0; // Expected violation, _ between digit and letter

  int jdk_9_0_392 = 0; // Expected violation, _ between digit and letter

  int testing_01231 = 0; // Expected violation, _ between digit and letter
}
```

```
$ java -jar checkstyle-11.1.0-all.jar -c google_checks.xml NamesWithUnderscores.java 
Starting audit...
[WARN] /mnt/5D92528E6B945467/test/testing/NamesWithUnderscores.java:5:7: Member name 'gradle9_5_1_' must match pattern '^(?![a-z]$)(?![a-z][A-Z])[a-z][a-zA-Z0-9]*(?:_[0-9]+)*$'. [MemberName]
Audit done.
```

we probably need a better approach than Regex as current regex has already grown very complex.

https://github.com/checkstyle/checkstyle/blob/5d52248c85b0fc37ddd786ddb5354e332f8a54b8/src/main/resources/google_checks.xml#L254-L256

if special check is needed, it should first determine the type of name - regular or name with numbering and then it should validate the name against the appropriate regex format, which will differ for each type of naming

romani:
current Check is https://checkstyle.sourceforge.io/checks/naming/membername.html 
[...]

PR #18312, opened 2025-12-15
Issue #17841: Add GoogleMethodNameCheck to fix method name false nega…

vivek-0509:
part of Issue: #17841
Issue: #18420 

## Description

Introduces a new `GoogleMethodName` check that properly validates method names according to the Google Java Style Guide, fixing false-negatives where underscores between letters and digits were not being flagged.

### Problem

The previous `MethodName` check failed to detect invalid names like:
- `gradle_9_5_1()` - underscore between letter 'e' and digit '9'
- `jdk_9_0_392()` - underscore between letter 'k' and digit '9'

### Solution

Created a dedicated `GoogleMethodName` check that:
- Has no configurable properties
- Uses multi-pattern validation instead of one complex regex
- Handles `@Test` methods separately (underscores allowed)
- Properly validates numbering suffixes (`guava33_4_5` is valid)

### Validation Rules

**Regular Methods:**
- Must be in lowerCamelCase
- Underscores only between adjacent digits (`_4_5`)
- Not between letter and digit (`gradle_9` invalid)

**Test Methods (`@Test`, `@ParameterizedTest`, `@RepeatedTest`):**
- Underscores allowed to separate components
- Each segment must be lowerCamelCase
- Each segment must be at least 2 characters
- Segments cannot start with single lowercase followed by uppercase (e.g., fO)

**Override Methods:**
- Methods with `@Override` annotation are skipped (no validation)

### Breaking Changes
None. This check is backward compatible with the previous [MethodName](cci:2://file:///Users/vivek/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/MethodNameCheck.java:40:0-160:1) regex behavior.

Diff Regression config: https://gist.githubusercontent.com/vivek-0509/3c0bc74332fbdb3a6ce4aef1c43d2f6b/raw/a7d36ba756226bac0e01e9148dc8bd7f3016ec96/base_config.xml

Diff Regression patch config: https://gist.githubusercontent.com/vivek-0509/a2777cd3e2a86dc84c549268d6651aa0/raw/6490530b3efb4213b7a2421dace6f157c7b264a2/patch_config.xml

Contirbution repo PR:  https://github.com/checkstyle/contribution/pull/1000

PR #18657, opened 2026-01-15
Issue #18494: Add GoogleNonConstantFieldNameCheck for Google Java Style compliance

vivek-0509:
Adds GoogleNonConstantFieldNameCheck to enforce non-constant field naming per [Google Java Style Guide §5.2.5](https://google.github.io/styleguide/javaguide.html#s5.2.5-non-constant-field-names).

Part of  #17842
Detailed Issue at #18494

### Rules Enforced

- lowerCamelCase format required
- Single-character names rejected (`f`)
- Hungarian-style prefixes rejected (`mValue`, `sCount`)
- Underscores only between adjacent digits (`guava33_4_5` )
- No leading/trailing/consecutive underscores

### Scope

- This check validates instance fields, while excluding constants (static final), static non-final fields and local variables.

Diff Regression config: https://gist.githubusercontent.com/vivek-0509/36cbe7f3789730951f15e7b6d551c6f3/raw/7ac2050528756f02e9952ae546a9c1e8cfed29c0/base_config.xml

Diff Regression patch config: https://gist.githubusercontent.com/vivek-0509/a979757b15933ce44f9b0f676a038195/raw/77433e7a8e91c9d621be184f4efcb829cb12d868/patch_config.xml

Contirbution repo PR: https://github.com/checkstyle/contribution/pull/1021


vivek-0509:
GitHub, generate report

vivek-0509:
Github, generate site

vivek-0509:
GitHub, generate report

vivek-0509:
Hey @romani, I have reviewed the diff report all are genuine violations.
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/0621ade_2026170816/reports/diff/index.html

romani:
Diff report is very huge hard to get actual diff.
Please generate report in comparison to old module that covered this to catch actual diff. Messages likely should be overriden to be same to avoid diff in wording to be a reason of diff.

vivek-0509:
GitHub, generate report

vivek-0509:
Hi @romani , I have generated a comparison diff report. I have overridden the messages to be identical. Here is the report https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/649199d_2026065114/reports/diff/index.html. Please let me know if I have understood your request correctly or if any changes are needed.

romani:
diff is still wild.
please give same `id` to both Checks to try avoid diff detection that is purely due to name of Check.

vivek-0509:
GitHub, generate report

vivek-0509:
@romani I updated the configurations to use the same id for both checks, but they still appear as Red/Green pairs. It seems the report tool uses the full Check class name for the rule name, so MemberName and GoogleMemberName are treated as different rules despite the shared ID. Am I missing something?

vivek-0509:
@romani 
I was investigating the report generation tool and i find that currently, the diff report compares violations using the full source string in 
https://github.com/checkstyle/contribution/blob/4ac81f9f9928d695869b93c568fced0d98a1e781/patch-diff-report-tool/src/main/java/com/github/checkstyle/data/CheckstyleRecord.java#L249-L252

When an explicit id is set in the config, the ID is appended to the class name in the format ClassName#id (`e.g., MemberNameCheck#GoogleMemberNameDiff)`. This means when two different checks share the same id, their source strings still differ because of the class name prefix `(e.g., MemberNameCheck#GoogleMemberNameDiff vs GoogleMemberNameCheck#GoogleMemberNameDiff).` As a result, violations with identical line, column, severity, and message are incorrectly reported as "removed" from base and "added" in patch.

the fix can be that, If source contains # (indicating an explicit module ID was set), compare only the portion after # instead of the full string. This would allow proper matching of violations from different checks that share the same ID, and the diff report would show only true behavioral differences.

romani:
tools is right, we tried to hack it, but tool is right in it behavior to show diff.

romani:
so, we can look at diff and see why we have more violations that it used to be. 

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/649199d_2026150917/reports/diff/elasticsearch/index.html#A1 
is good example, previosly we did not do violation, now we do.
https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

google state that type immutability if critical point in decision. Checkstyle is not type aware, we can not make decision if it is immutable or not. So better to not give violation.
False positive will make users be frustrated, as it will block them, but false-negative (absence of violation) will not make users angry.

vivek-0509:
Understood, will change the scope to skip all static fields and update the issue accordingly.

vivek-0509:
GitHub, generate report

vivek-0509:
@romani updated the scope for the check to skip static member names 
here is the report the false negatives are now getting correcty flagged below are some examples from the report
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/f342244_2026110001/reports/diff/openjdk25/index.html#A15156

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/f342244_2026110001/reports/diff/openjdk25/index.html#A33718

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/f342244_2026110001/reports/diff/nbia-dcm4che-tools/index.html#A9

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * <p>
+ * This check enforces Google's specific member naming requirements:
+ * </p>
+ * <ul>
+ * <li>Member names must start with a lowercase letter and use uppercase letters
+ * for word boundaries.</li>
+ * <li>Underscores may be used to separate adjacent numbers (e.g., version
+ * numbers like {@code guava33_4_5}), but NOT between letters and digits.</li>
+ * </ul>
```
romani, line 47: we need to mention that static fields are skipped, as Chckstyle not able to make sure that used type is immutable.

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleMemberNameCheck.java
```diff
@@ -0,0 +1,155 @@
...
+ * <li>Member names must start with a lowercase letter and use uppercase letters
+ * for word boundaries.</li>
+ * <li>Underscores may be used to separate adjacent numbers (e.g., version
+ * numbers like {@code guava33_4_5}), but NOT between letters and digits.</li>
+ * </ul>
+ *
+ * @since 13.1.0
+ */
+@StatelessCheck
+public class GoogleMemberNameCheck extends AbstractCheck {
```
romani: > Checks that member names conform to the
 <a href=
 "https://google.github.io/styleguide/javaguide.html#s5.2.5-non-constant-field-names">
 Google Java Style Guide</a> for non-constant field naming.

hmm, why we named it GoogleMemberNameCheck ?

as it should be GoogleNonConstantFieldNameCheck.
and it will make it clear why we have limitation on static.

romani (changes requested):
items:

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
vivek-0509: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleMemberNameCheck.java
vivek-0509: done

vivek-0509:
@romani all the changes done pr is ready for review

vivek-0509:
GitHub, generate report

vivek-0509:
GitHub, generate website

vivek-0509:
GitHub, generate website

vivek-0509:
@romani 
I am considering improving the violation message for this check also to better guide users when special characters are used in field names.

**Current message:**
` Non-constant field name ''{0}'' must start with a lowercase letter, min 2 chars, avoid single lowercase letter followed by uppercase.`

**Updated message:**
`Non-constant field name ''{0}'' must start with a lowercase letter, min 2 chars, avoid single lowercase letter followed by uppercase, and contain only letters, digits, or underscores.`

For cases like `$foo` or `foo$bar`, the current message does not clearly explain why the name is invalid. The Google Style Guide 5.1 explicitly states:

> "Identifiers use only ASCII letters and digits, and, in a small number of cases noted below, underscores. Thus each valid identifier name is matched by the regular expression `\w+`."

Does this violation message looks good to you??

vivek-0509:
@romani I have updated the violation messages since this is the same case as we have done for the GoogleMethodNameCheck as discussed with @Zopsss in that PR, please let me know if any other changes are needed.

vivek-0509:
@Zopsss Do you want me to keep the full violation message for this check's Example1 also? I have not updated it since I have covered both the violation messages in the examples.

vivek-0509:
@romani Can you please review this PR when you have time

Zopsss:
> Do you want me to keep the full violation message for this check's Example1 also? I have not updated it since I have covered both the violation messages in the examples.

@vivek-0509 sorry for the delay, yes please do that.

vivek-0509:
GitHub, generate website

vivek-0509:
@Zopsss Done, please review

vivek-0509:
GitHub, generate report

vivek-0509:
CI Failure is unrelated

vivek-0509:
@Zopsss ping

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
```diff
@@ -0,0 +1,161 @@
...
+            final String fieldName = nameAst.getText();
+
+            validateNonConstantFieldName(nameAst, fieldName);
+        }
+    }
+
+    /**
+     * Checks if this field should be validated. Returns true for instance fields only.
+     * Static fields are excluded because Checkstyle cannot determine type immutability.
+     * Local variables and interface/annotation fields are also excluded.
```
Zopsss, line 133: why local variables and interface/annotation fields are ignored?

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/naming/googlenonconstantfieldname/Example1.java
```diff
@@ -0,0 +1,37 @@
...
+  <module name="TreeWalker">
+    <module name="GoogleNonConstantFieldName"/>
+  </module>
+</module>
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.naming.googlenonconstantfieldname;
+
+// xdoc section -- start
+class Example1 {
```
Zopsss, line 12: please add following cases:

- static final
- static non-final
- non static final
- public/private/protected cases

Zopsss: Lets keep separate examples. One example should contain all the valid cases and the other should contain all the invalid cases.

Zopsss (changes requested):
Items:

Zopsss:
So sorry for the delayed reviews. And thanks a lot for the ping!

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
vivek-0509: we have excluded Local variables because this check targets `non-constant field names (non-static, non-constant class member variables)` and Local variables are not fields (member variables), they have their own naming rules and check LocalVariableNameCheck. 
About excluding interfaces/annotations, since fields declared in interfaces and anotations are implicitly `static final` (constants), which fall under "Constant names".

vivek-0509:
GitHub, generate website

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/naming/googlenonconstantfieldname/Example1.java
vivek-0509: done

vivek-0509:
@Zopsss All the changes done please review 
 the diff report link https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/0bb782c_2026121113/reports/diff/index.html

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/naming/googlenonconstantfieldname/Example2.java
```diff
@@ -0,0 +1,34 @@
...
+</module>
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.naming.googlenonconstantfieldname;
+
+// xdoc section -- start
+class Example2 {
+
+  static final int STATIC_FINAL = 0;
+  private static final int Invalid_Name = 1;
```
Zopsss: please correct the variable name. It is static final so it doesn't throw any violation.

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
```diff
@@ -0,0 +1,161 @@
...
+     *
+     * @param ast the VARIABLE_DEF AST node
+     * @return true if this variable should be checked
+     */
+    private static boolean mustCheckName(DetailAST ast) {
```
Zopsss: please rename this method to `shouldCheckFieldName`

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
```diff
@@ -0,0 +1,161 @@
...
+    private void validateNonConstantFieldName(DetailAST nameAst, String fieldName) {
+        if (INVALID_UNDERSCORE_PATTERN.matcher(fieldName).find()) {
+            log(nameAst, MSG_KEY_INVALID_UNDERSCORE, fieldName);
+        }
+        else {
+            final String nameWithoutNumberingSuffix = NUMBERING_SUFFIX_PATTERN
+                    .matcher(fieldName).replaceAll("");
+            if (!NON_CONSTANT_FIELD_NAME_PATTERN.matcher(nameWithoutNumberingSuffix).matches()) {
+                log(nameAst, MSG_KEY_INVALID_FORMAT, fieldName);
+            }
```
Zopsss: here we are using if/else condition. 

If there is a variable like `aTest_Test_09_90` then on the first pass it will log violation for invalid underscore, then user will run CS again and it will log violation for invalid non constant field name.

User will need to run CS 2 times.

The user should be able to solve all problems in one go, for that all violations should be logged in the first pass.

Zopsss (changes requested):
minors:

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
Zopsss: Thanks for the explanation. 

> fields declared in interfaces and anotations are implicitly static final (constants), which fall under "Constant names".

Do we have check for the constant names? and do we use it in our implementation?

Also, please add interfaces/annotations examples to the Check's documentation. And also explicitly mention that this check ignores interfaces & annotations with explanation.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/naming/googlenonconstantfieldname/Example1.java
Zopsss: Thank you. You can resolve this conversation.

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
vivek-0509: Yes, Checkstyle has a ConstantNameCheck, but we intentionally do not use it for Google Style because 
Checkstyle cannot determine if a field's contents are deeply immutable . Google's definition of constants is  `Constants are static final fields whose contents are deeply immutable and whose methods have no detectable side effects.` Since we can not verify this, we skip all static fields to avoid false positives.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/naming/googlenonconstantfieldname/Example2.java
vivek-0509: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
vivek-0509: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
vivek-0509: Sure, I will use a single message and will simplify the method, the message would be 

`Non-constant field name ''{0}'' must start lowercase, be at least 2 chars, avoid single lowercase letter followed by uppercase, contain only letters, digits or underscores, with underscores allowed only between adjacent digits.`

This way, when a name fails validation, the user sees all the rules that are enforced and can fix everything in one go.

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
vivek-0509: done updated

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
vivek-0509: done

vivek-0509:
GitHub, generate website

vivek-0509:
GitHub, generate report

vivek-0509:
@Zopsss All the changes are done please review

vivek-0509:
@Zopsss Ping

Zopsss:
@vivek-0509 There is different number of violations in old diff report ( where we had multiple violation messages ) and the latest diff report ( where we have only one violation message )

Old report: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/0bb782c_2026121113/reports/diff/index.html

Latest Report: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/60f0289_2026020714/reports/diff/index.html

HBase & openjdk25's violation messages numbers do not match. Can you investigate why this happened? And did we introduced any false-positive/negative with new changes?

vivek-0509:
@Zopsss I also noticed this change. However, there are no logical changes or modifications to the regex. We only consolidated the if else blocks into a single condition while preserving the same logic in the check class.

I re-reviewed the diff report after the changes with this in mind and did not find any false positives. Since we fetch the latest code, the increase in violations appears to be due to new files added to the codebase that already contain these violations.

In the diff report, the difference is observed for Checkstyle and OpenJDK, while HBase still shows the same number of violations.

vivek-0509:
@Zopsss 
For checkstyle in the old report we see 715 files were checked
Screenshot 2026-02-13 at 12 02 39 PM

In new report there are 721 files checked
Screenshot 2026-02-13 at 12 02 50 PM

So the  new files also has violations that our check flagged due to which we see increased in the violation number.

Same case with openJDK there also 2 files increased which lead to increase in the violation number

vivek-0509:
For Checkstyle, I found a new file named `InputXpathNumericalPrefixesInfixesSuffixesCharacterCaseOne.java, InputXpathNumericalPrefixesInfixesSuffixesCharacterCaseTwo.java `that was not present in the previous report. This file contains genuine violations, which explains the increase in the number of violations.

Link to one of the new files in the new diff report
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/60f0289_2026020714/reports/diff/checkstyle/index.html#A214

Zopsss:
@vivek-0509 if you have extra time, thne can you also check reason behind extra violations in openjdk? If you're not able to then it's also fine.

Zopsss:
@vivek-0509 if you have extra time, thne can you also check reason behind extra violations in openjdk? If you're not able to then it's also fine.

vivek-0509:
I have also checked for openJDK before sorry for not mentioning about it, it has the same reason
for old report 3980 files were checked
Screenshot 2026-02-13 at 11 22 41 PM

for New report 3982 files are checked 
Screenshot 2026-02-13 at 11 23 08 PM

which resulted in the increase of violation count by 7

vivek-0509:
@Zopsss 
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/60f0289_2026020714/reports/diff/openjdk25/index.html#A493
one example of such new file which was not there in the old report

vivek-0509:
`TestPhaseIRMatching.java`
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/60f0289_2026020714/reports/diff/openjdk25/index.html#A15910
Also, this file shows an increased number of violations. We can see that even the old violations now appear on different line numbers, which indicates that the file was modified between the generation of the old and the new reports. so we catched more vioaltions that the new changes in `TestPhaseIRMatching.java` have  introduced

In Old report 
Screenshot 2026-02-13 at 11 56 48 PM

In New report
Screenshot 2026-02-13 at 11 57 06 PM

vivek-0509:
@Zopsss Thankyou very much for reviewing

vivek-0509:
@romani PR is waiting for your approval.

vivek-0509:
GitHub, generate report

vivek-0509:
GitHub, generate website

vivek-0509:
@romani Let me know if any changes are needed ...

config/checkstyle-examples-suppressions.xml
```diff
@@ -194,4 +194,9 @@
   <suppress checks="LineLength"
             files="lineending[\\/]Example4.java"/>
 
+  <!-- violation messages are long by design to show full text -->
+  <suppress checks="LineLength" files="googlenonconstantfieldname[\\/]Example\d+.*"/>
+  <suppress id="discouragedRegexpInViolationMessage"
+            files="googlenonconstantfieldname[\\/]Example\d+.*"/>
```
romani: this is unreadable, and looks like noise:
Image

please update:
```
  final int Bad = 0;
  // violation above, 'Non-constant field name 'Bad' must start lowercase, be at least 2 chars, avoid single lowercase letter followed by uppercase, contain only letters, digits or underscores, with underscores allowed only between adjacent digits.'
```

to:
```
  final int Bad = 0;
  // violation above, 'Non-constant field name 'Bad' must start lowercase'
```

---------


```
private final int f = 0;
  // violation above, 'Non-constant field name 'f' must start lowercase, be at least 2 chars, avoid single lowercase letter followed by uppercase, contain only letters, digits or underscores, with underscores allowed only between adjacent digits.'
```
to:
```
private final int f = 0;
  // violation above, 'be at least 2 chars'
```

apply same pattern to all others.

src/it/java/org/checkstyle/suppressionxpathfilter/naming/XpathRegressionGoogleNonConstantFieldNameTest.java
```diff
@@ -0,0 +1,86 @@
...
+        };
+        final List<String> expectedXpathQueries = List.of(
+            "/COMPILATION_UNIT/CLASS_DEF"
+                + "[./IDENT[@text='InputXpathGoogleNonConstantFieldNameInvalidUnderscore']]"
+                + "/OBJBLOCK/VARIABLE_DEF/IDENT[@text='gradle_9_5_1']"
+        );
+        runVerifications(moduleConfig, fileToProcess, expectedViolation,
+                expectedXpathQueries);
+    }
+}
```
romani, line 106: we ask for 3 tests for xpath, magic 3 :).

src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
```diff
@@ -79,6 +79,15 @@ public class AllChecksTest extends AbstractModuleTestSupport {
             new HashMap<>();
     private static final Map<String, Set<String>> GOOGLE_TOKENS_IN_CONFIG_TO_IGNORE =
             new HashMap<>();
+
+    /**
+     * Modules not included in checkstyle-checks.xml.
+     * These checks exist but are not activated in Checkstyle's internal configuration.
```
romani, line 85: can we put here "until #xxxxxxxx"
to not forget to remove this as we add it google style config.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/naming/googlenonconstantfieldname/Example1.java
```diff
@@ -0,0 +1,60 @@
...
+    String Invalid_Name = "test";
+  }
+
+  @interface ExampleAnnotation {
+    int DEFAULT = 0;
+    String Bad_Name = "test";
+  }
+
+  final int Bad = 0;
+  // violation above, 'Non-constant field name 'Bad' must start lowercase, be at least 2 chars, avoid single lowercase letter followed by uppercase, contain only letters, digits or underscores, with underscores allowed only between adjacent digits.'
```
romani:
```
final int bad = 0;
final int Bad = 0; // violation '......'
```

please put on top of bad variable, some sort of potential good name or it.

and we will not need probably example2 after it.

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
```diff
@@ -0,0 +1,153 @@
...
+
+    @Override
+    public int[] getRequiredTokens() {
+        return new int[] {TokenTypes.VARIABLE_DEF};
+    }
+
+    @Override
+    public void visitToken(DetailAST ast) {
+        if (shouldCheckFieldName(ast)) {
+            final DetailAST nameAst = NullUtil.notNull(ast.findFirstToken(TokenTypes.IDENT));
```
romani: we already have good pattern when we hide notnull in extra method:
`final DetailAST nameAst = getIdent(ast);`

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Checks if this field should be validated. Returns true for instance fields only.
+     * Static fields are excluded because Checkstyle cannot determine type immutability.
+     * Local variables and interface/annotation fields are also excluded.
+     *
+     * @param ast the VARIABLE_DEF AST node
+     * @return true if this variable should be checked
+     */
+    private static boolean shouldCheckFieldName(DetailAST ast) {
+        final DetailAST modifiersAST = NullUtil.notNull(ast.findFirstToken(TokenTypes.MODIFIERS));
```
romani: why we think that modifiers always exists?
they defineterly not present always in code, is this our AST feature to have always this node , even it is empty ?

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
```diff
@@ -0,0 +1,153 @@
...
+     *
+     * @param nameAst    the IDENT AST node containing the field name
+     * @param fieldName the field name string
+     */
+    private void validateNonConstantFieldName(DetailAST nameAst, String fieldName) {
+        final String nameWithoutNumberingSuffix = NUMBERING_SUFFIX_PATTERN
+                .matcher(fieldName).replaceAll("");
+
+        if (INVALID_UNDERSCORE_PATTERN.matcher(fieldName).find()
+                || !NON_CONSTANT_FIELD_NAME_PATTERN.matcher(nameWithoutNumberingSuffix).matches()) {
```
romani, line 172: find() vs matched() , hmmmm, why different methods are used ?

src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
vivek-0509: This check is Google-specific, similar to GoogleMethodName, which was also excluded from checkstyle-checks.xml as discussed earlier (https://github.com/checkstyle/checkstyle/pull/18312#discussion_r2656822942). It will permanently remain in the exclusion lis, so there's no tracking issue for this.
Do you want to activate this one in the `checkstyle-checks.xml` ??

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
vivek-0509: Yes, modifiers is always present in the AST for VARIABLE_DEF nodes, even when no modifiers are explicitly written in code (e.g., `int x;` still has an empty MODIFIERS node). This is a Checkstyle AST feature. Existing checks like MemberNameCheck, ConstantNameCheck all use `ast.findFirstToken(TokenTypes.MODIFIERS)` directly without null checks.
https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#MODIFIERS

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
vivek-0509: i used `find()` and `matches()` intentionally for different purposes. `find()` checks if the pattern exists anywhere in the string, while `matches()` requires the entire string to match the pattern. 

 for INVALID_UNDERSCORE_PATTERN because we need to detect if  an invalid underscore position exists anywhere in the name (e.g., `e_9` in `gradle_9_5_1`)so we used `find()` here. 
`matches()` is used for NON_CONSTANT_FIELD_NAME_PATTERN because the entire name must conform to the NON_CONSTANT_FIELD_NAME_PATTERN.

If we want to unify both to use `matches()`, we can wrap the underscore pattern with `.*`:
`".*(?:^_|_$|__|[a-zA-Z]_[a-zA-Z]|[a-zA-Z]_\\d|\\d_[a-zA-Z]).*"`
But I think using `find()` keeps the intent clearer. Let me know your preference.

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
vivek-0509: done

src/it/java/org/checkstyle/suppressionxpathfilter/naming/XpathRegressionGoogleNonConstantFieldNameTest.java
vivek-0509: done

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/naming/googlenonconstantfieldname/Example1.java
vivek-0509: done

config/checkstyle-examples-suppressions.xml
vivek-0509: done, reverted back to the earlier version

vivek-0509:
GitHub, generate report

vivek-0509:
GitHub, generate website

vivek-0509:
@romani All the changes done PR is ready for review

src/test/java/com/puppycrawl/tools/checkstyle/internal/AllChecksTest.java
romani: my bad, I misread it as google_check.xml

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
romani: I thinks awesome knowledge, please create special method getModifiers(ast) in this Check and put this init javadoc, eventually this method will be moved to some util, no need to do this now.

src/it/java/org/checkstyle/suppressionxpathfilter/naming/XpathRegressionGoogleNonConstantFieldNameTest.java
```diff
@@ -0,0 +1,100 @@
...
+            getPath("InputXpathGoogleNonConstantFieldNameSingleChar.java"));
+        final DefaultConfiguration moduleConfig = createModuleConfig(CLASS);
+        final String[] expectedViolation = {
+            "5:9: " + getCheckMessage(CLASS,
+                    GoogleNonConstantFieldNameCheck.MSG_KEY_INVALID_FORMAT, "a"),
+        };
+        final List<String> expectedXpathQueries = List.of(
+            "/COMPILATION_UNIT/CLASS_DEF"
+                + "[./IDENT[@text='InputXpathGoogleNonConstantFieldNameSingleChar']]"
+                + "/OBJBLOCK/VARIABLE_DEF/IDENT[@text='a']"
```
romani: All 3 examples should be different by Xpath structure.

Please make: as filed of class,  as field of inner interface, as filed of inner record or anonymous class.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/naming/googlenonconstantfieldname/Example1.java
```diff
@@ -0,0 +1,57 @@
...
+    String Bad_Name = "test";
+  }
+
+  final int bad = 0;
+  final int Bad = 0;
+  // violation above, 'Non-constant field name 'Bad' must start lowercase'
+
+  public final int httpClient = 0;
+  public final int mValue = 0;
+  // violation above, 'avoid single lowercase letter followed by uppercase'
```
romani, line 36:
```
public final int myValue = 0;
public final int mValue = 0;
```

line above shows a suggested fix for line with violation.

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/naming/googlenonconstantfieldname/Example1.java
```diff
@@ -0,0 +1,57 @@
...
+
+  private final int foo = 0;
+  private final int f = 0; // violation, 'be at least 2 chars'
+
+  protected final int fooBar1_1 = 0;
+  protected final int foo_bar = 0;
```
romani, line 42:
```
protected final int fooBar = 0;
protected final int foo_bar = 0;
```

apply same for all others.

vivek-0509: done

src/main/java/com/puppycrawl/tools/checkstyle/checks/naming/GoogleNonConstantFieldNameCheck.java
vivek-0509: done

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/naming/googlenonconstantfieldname/Example1.java
vivek-0509: done

src/it/java/org/checkstyle/suppressionxpathfilter/naming/XpathRegressionGoogleNonConstantFieldNameTest.java
vivek-0509: done

vivek-0509:
GitHub, generate website

src/it/java/org/checkstyle/suppressionxpathfilter/naming/XpathRegressionGoogleNonConstantFieldNameTest.java
vivek-0509: Used class inside interface for input3 since interface fields are implicitly `public static final` and can't produce violations and used anonymous class for input3.

vivek-0509:
GitHub, generate website

vivek-0509:

PR #18635, opened 2026-01-13
Issue #16678: new check for NumericalCharacterCase

Aman-Baliyan:
Issue: #16678 

Implemented new check NumericalCharacterCase that checks numerical character prefixes, infixes and suffixes are in lowercase or not.

Contirbution repo PR: https://github.com/checkstyle/contribution/pull/1019

New module config:  https://gist.githubusercontent.com/Aman-Baliyan/50a38d637e518654e3c7b02116ab7b10/raw/e8f7358969379fc5eb642aae76f137c158e1cdf5/NumericalCharacterCase.config

PR #16757, opened 2025-04-04
Issue #11166: Remove usage of getFileContents() from WriteTagCheck

Rohanraj123:
part of #11166 

Diff Regression config: https://gist.githubusercontent.com/Rohanraj123/f8548a1a76258afebf6c9ef15d09dd59/raw/3d05b3263103cb1ab091134458ce50af7081f3c0/gistfile1.txt

PR #18677, opened 2026-01-18
Issue #11166: remove getFileContents from UnusedImportCheck

Aman-Baliyan:
Issue: #11166 

What this pr does?
- remove usage of getFileContents() from UnusedImportCheck.java and add a method which extract javadoc directly from ast.

Problems -

```
there is a issue in `createFullIdent()` of` FullIdent.java` when comment are enabled in the ast the example of the issue are -

Example1 - 

import java.util. // This is comment
List

FullIdent output - > import java.util. //

but it has to be- import java.util.List

Example 2-

import java.util. 
List // This is comment

FullIdent output - > import java.util.List

it works fine when the comments are in last line 
```

the fullIdent issue is also solved in this pr.

### DIFF
--- a/config/pitest-suppressions/pitest-api-suppressions.xml
+++ b/config/pitest-suppressions/pitest-api-suppressions.xml
@@ -1,3 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<suppressedMutations>
-</suppressedMutations>
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/FullIdent.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/FullIdent.java
@@ -101,13 +101,34 @@ public final class FullIdent {
                 pushToIdentStack(identStack, firstChild);
             }
             else if (typeOfAst == TokenTypes.DOT) {
-                pushToIdentStack(identStack, firstChild.getNextSibling());
+                final DetailAST next = firstChild.getNextSibling();
+                processDot(identStack, next);
                 pushToIdentStack(identStack, firstChild);
                 dotCounter++;
             }
             else {
                 dotCounter = appendToFull(full, currentAst, dotCounter,
                         bracketsExist, isArrayTypeDeclarationStart);
+            }
+        }
+    }
+
+    /**
+     * Check a node is comment node or not then push to stack.
+     *
+     * @param identStack the Deque to add to
+     * @param ast the node
+     */
+    public static void processDot(Deque<DetailAST> identStack, DetailAST ast) {
+        if (ast != null) {
+            final int astType = ast.getType();
+
+            if (astType == TokenTypes.SINGLE_LINE_COMMENT
+                    || astType == TokenTypes.BLOCK_COMMENT_BEGIN) {
+                pushToIdentStack(identStack, ast.getNextSibling());
+            }
+            else {
+                pushToIdentStack(identStack, ast);
             }
         }
     }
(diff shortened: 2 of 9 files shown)


Aman-Baliyan:
working on ci failure i think i miss some tokens in my getJavadoc method

romani:
ident should be `java.util.List` no matter where and how much comments user put, comments should be skipped by logic

Aman-Baliyan:
i add a unit test case in FullIdentTest.java to kill this mutation 

```
+    <sourceFile>FullIdent.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.api.FullIdent</mutatedClass>
+    <mutatedMethod>processDot</mutatedMethod>
+    <lineContent>if (next != null) {</lineContent>
+  </mutation>

```

and if i remove the line from the code then a unit test in `AnnotationUtilTest.java` is failed so i uses the unit test to kill that mutation

Aman-Baliyan:
@romani all check have been passed can you please review it

config/pitest-suppressions/pitest-api-suppressions.xml
romani, line 1: please restore file, we keep them empty to easily generate content for them in diff when pitest exeuciton failing.
do same for other suppression file

src/main/java/com/puppycrawl/tools/checkstyle/api/FullIdent.java
```diff
@@ -112,6 +113,26 @@ else if (typeOfAst == TokenTypes.DOT) {
         }
     }
 
+    /**
+     * Check a node is comment node or not then push to stack.
+     *
+     * @param identStack the Deque to add to
+     * @param ast the node
+     */
+    public static void processDot(Deque<DetailAST> identStack, DetailAST ast) {
```
romani, line 122: can we move it to Check, we should not extend API.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
```diff
@@ -310,19 +320,85 @@ private void processStaticImport(DetailAST ast) {
...
-        final int lineNo = ast.getLineNo();
-        final TextBlock textBlock = contents.getJavadocBefore(lineNo);
-        if (textBlock != null) {
-            currentFrame.addReferencedTypes(collectReferencesFromJavadoc(textBlock));
+        final DetailAST javadoc = getJavadoc(ast);
+        if (javadoc != null) {
+            final String content = JavadocUtil.getJavadocCommentContent(javadoc);
+            final String[] cmtLines;
+            if (LINE_SPLIT_PATTERN.matcher(content).find()) {
+                cmtLines = LINE_SPLIT_PATTERN.split(content);
```
romani, line 393: hmm , this is not reliable approach.

and I pretty sure I saw this this already in PR for update ......
here it is - https://github.com/checkstyle/checkstyle/pull/18059

OMG ..... in storm of PRs, we got collision.

romani (changes requested):
this PR is less advanced than other PR https://github.com/checkstyle/checkstyle/pull/18059

It think we should stop doing this PR.

src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/UnusedImportsCheck.java
Aman-Baliyan: i saw this approach in pr- #16757 that's why i use this. if #18059 is more advanced then we can close this pr

src/main/java/com/puppycrawl/tools/checkstyle/api/FullIdent.java
Aman-Baliyan: if i move it to check then we have to write the whole logic of FullIdent in check  but if we keep it in api then it is helpful in  future because if in some other check both fullIdent and comment are used then the same issue arises

Aman-Baliyan:
if this pr is not helpful then we can close this pr and if not closing then i can proceed further. Tell me about your thoughts on this.

Aman-Baliyan:
please review this pr #18635

romani:
Sorry that we run into such concurrency...

Issue #18619, opened 2026-01-12
Wrong error message while using check NewlineAtEndOfFile


Atharv3221:
I have read check documentation: https://checkstyle.sourceforge.io/checks/misc/newlineatendoffile.html#NewlineAtEndOfFile
I have downloaded the latest checkstyle from: https://github.com/checkstyle/checkstyle/releases/tag/checkstyle-13.0.0
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```
atharv@atharv-IdeaPad-3-15IIL05:~/test$ cat -A a.txt
This file has line ending CR^MThis file as new line at end of file but with CR^M
```

File as line ending CR.

```
atharv@atharv-IdeaPad-3-15IIL05:~/test$ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
  "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
  "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <module name="NewlineAtEndOfFile">
    <property name="lineSeparator" value="lf"/>
  </module>
</module>

```

```
atharv@atharv-IdeaPad-3-15IIL05:~/test$ java -jar checkstyle-13.0.0-all.jar -c config.xml a.txt
Starting audit...
[ERROR] /home/atharv/test/a.txt:1: File does not end with a newline. [NewlineAtEndOfFile]
Audit done.
Checkstyle ends with 1 errors.
```

Error expected was:
`Expected line ending for file is LF(\\n), but CR(\\r) is detected.`

Similar kind of example is used https://checkstyle.sourceforge.io/checks/misc/newlineatendoffile.html#Example3-code

CR is considered valid line separator: https://checkstyle.sourceforge.io/property_types.html#LineSeparatorOption

Saadmadni84:
Hi maintainers,
I would like to work on this issue and submit a fix for the incorrect error message when a wrong line separator is detected. Please let me know if I can proceed.

romani:
we can improve message to be `File does not end with a newline 'lf'.`

This check should not care what symbols are in this files, and what symbol is at of file already. Main point that it is not 'lf'.

romani:
@Saadmadni84 ,  issue is approve now.

[...]

PR #18678, opened 2026-01-18
Issue #18619: Simplify NewlineAtEndOfFile message for missing LF

Saadmadni84:
Issue: #18619

This PR updates the NewlineAtEndOfFile check to report a clearer and non-misleading violation message when a file does not end with a newline.

The message is now simplified to always indicate a missing trailing LF ('\n'), independent of any internal line separators, as clarified by the maintainer.

All related tests were updated accordingly and are passing.


romani:
we have other PR that is more right way for fix .

@Saadmadni84 , please help us on other issue.

PR #18687, opened 2026-01-19
Issue #11103: Fix Indentation check for lambda in enum constant argum…

### LINKED ISSUE #11103: Indentation: Lambda in Enum causes error loop
Check documentation: https://checkstyle.org/config_misc.html#Indentation

```bash
/var/tmp $ javac SomeEnum.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
    <module name="TreeWalker">
        <module name="Indentation"/>
    </module>
</module>

/var/tmp $ cat SomeEnum.java
package some.pckg;

import java.util.function.Function;

public enum SomeEnum {

    ENUM_VALUE(
            v -> v, // violation, but it is not expected
            new Object()
    );

    private final Function<Object, Object> function;
    private final Object object;

    SomeEnum(Function<Object, Object> function, Object object) {
        this.function = function;
        this.object = object;
    }
}


/var/tmp $ java $RUN_LOCALE -jar checkstyle-9.2-all.jar -c config.xml SomeEnum.java
Starting audit...
[ERROR] C:\test\src\main\java\some\pckg\SomeEnum.java:8:13: 'lambda arguments' has incorrect indentation level 12, expected level should be 16. [Indentation]
Audit done.
```

---

I expect Checkstyle to pass without errors. If I try to increase indentation like [ERROR] suggests (from 12 to 16), I get the same error, but with indentation levels 16 and 20 (and so on). So the error is looped. Similar thing happened in #1141.

### LINKED ISSUE #13539: Checkstyle requires inconsistent indentation for lambda arguments
```bash
/var/tmp $ javac TestEnum.java
/var/tmp $ cat config.xml
```
```xml
<?xml version="1.0"?>

<module name="Checker">
  <module name="TreeWalker">
    <module name="Indentation"/>
  </module>
</module>
```
```bash
/var/tmp $ cat TestEnum.java
```
```java
import java.util.function.Consumer;
import java.util.function.Function;

public enum TestEnum {
  NAME(
      String::getClass,
      any -> {}
  );


    TestEnum(
        Function<String, Class<?>> function,
        Consumer<String> consumer
    ) {

    }
}
```
```bash
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.12.2-all.jar -c config.xml TestEnum.java
Starting audit...
[ERROR] /var/tmp/TestEnum.java:9:7: 'lambda arguments' has incorrect indentation level 6, expected level should be 10. [Indentation]
Audit done.
```
---

This should be valid, but is not
```java
  NAME(
      String::getClass,
      any -> {}
  );
```
This should not be valid, but it is
```java
  NAME(
      String::getClass,
          any -> {}
  );
```

vivek-0509:
fixes #11103
fixes #13539

Added  handling in `LambdaHandler.getIndentImpl()` to detect  lambdas inside enum constant arguments and calculate their expected indentation based on the enum constant's position.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/LambdaHandler.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/LambdaHandler.java
@@ -19,6 +19,8 @@
 
 package com.puppycrawl.tools.checkstyle.checks.indentation;
 
+import javax.annotation.Nullable;
+
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
 import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
@@ -91,22 +93,30 @@ public class LambdaHandler extends AbstractExpressionHandler {
             return getParent().getSuggestedChildIndent(this);
         }
 
-        DetailAST parent = getMainAst().getParent();
-        if (getParent() instanceof NewHandler) {
-            parent = parent.getParent();
+        final IndentLevel result;
+        final DetailAST enumConstDef = findParentEnumConstantDef();
+        if (enumConstDef != null) {
+            result = getEnumConstantBasedIndent(enumConstDef);
+        }
+        else {
+            DetailAST parent = getMainAst().getParent();
+            if (getParent() instanceof NewHandler) {
+                parent = parent.getParent();
+            }
+
+            // Use the start of the parent's line as the reference indentation level.
+            IndentLevel level = new IndentLevel(getLineStart(parent));
+
+            // If the start of the lambda is the first element on the line;
+            // assume line wrapping with respect to its parent.
+            final DetailAST firstChild = getMainAst().getFirstChild();
+            if (getLineStart(firstChild) == expandedTabsColumnNo(firstChild)) {
+                level = new IndentLevel(level, getIndentCheck().getLineWrappingIndentation());
+            }
+            result = level;
         }
 
-        // Use the start of the parent's line as the reference indentation level.
-        IndentLevel level = new IndentLevel(getLineStart(parent));
-
-        // If the start of the lambda is the first element on the line;
-        // assume line wrapping with respect to its parent.
-        final DetailAST firstChild = getMainAst().getFirstChild();
-        if (getLineStart(firstChild) == expandedTabsColumnNo(firstChild)) {
-            level = new IndentLevel(level, getIndentCheck().getLineWrappingIndentation());
-        }
-
-        return level;
+        return result;
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


vivek-0509:

vivek-0509:
Investigating the cases

vivek-0509:
Hi @romani 
I have reviewed the diff report and believe the fix is working correctly. The newly flagged violations appear to be genuine indentation issues that were previously being missed due to the bug in LambdaHandler.getIndentImpl().

For example, the new violation in https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/016e96a76d3ef484bf86b85a78df95e7cb1034fd_2026143035/reports/diff/openjdk25/index.html#A1:

```
        FILE_ASSOCIATIONS ("file-associations",
                OptionCategories.PROPERTY, () -> {
            Map<String, ? super Object> args = new HashMap<>();  // column 12
```
The lambda () -> { appears inline after OptionCategories.PROPERTY which starts at column 16. Therefore, the lambda body should be at 16+4=20, but it's at column 12. This is genuinely incorrect indentation that was being overlooked before.

The current fix uses ENUM_CONSTANT_DEF as the base, which calculates expected body indentation as 16 (enum constant 8 + lineWrapping 4 + basicOffset 4). For inline lambdas, this is slightly lower than the visually ideal 20. However, with forceStrictCondition=false (default), both 16 and 20 are acceptable. Once we align on this fix and confirm there are no false positives, I can explore adding stricter handling for inline lambdas as a follow up enhancement.

Please review the diff report and let me know if any of the newly flagged violations appear to be false positives or if I am missing something.

vivek-0509:
Additionally, the fix also resolves false positives for correctly indented code.
I tested a block lambda with statements in an enum constant on the master branch 

```
public enum TestLambdaEnumBlock {
    VALUE(
        () -> {
            System.out.println("test");
        }
    );

    TestLambdaEnumBlock(Runnable runnable) {
    }
}
```

Result on master (without fix):
```
Starting audit...
[ERROR] TestLambdaEnumBlock.java:4:9: 'lambda arguments' has incorrect indentation level 8, expected level should be 12. [Indentation]
[ERROR] TestLambdaEnumBlock.java:5:13: 'block' child has incorrect indentation level 12, expected level should be 16. [Indentation]
[ERROR] TestLambdaEnumBlock.java:6:9: 'block rcurly' has incorrect indentation level 8, expected level should be 12. [Indentation]
Audit done.
```
However, this code Is correctly indented:

- Lambda at column 8 = enum constant (4) + lineWrappingIndentation (4) 
- Body at column 12 = lambda (8) + basicOffset (4) 
- Closing brace at column 8 = matches lambda 

This example shows that the original code produces false positives for block lambdas with statements in enum constants.

romani:
Please trigger diff test on all examples config

vivek-0509:
@romani  I have already triggered Github, generate report for Indentation/all-examples-in-one.
Here is the report link: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/016e96a76d3ef484bf86b85a78df95e7cb1034fd_2026143035/reports/diff/index.html

Could you please clarify which specific config you'd like me to test with? Is there a different command I should use?

vivek-0509:
@romani ping

PR #18695, opened 2026-01-20
Issue #14122: Recfactored DefaultLogger class to avoid CT_CONSTRUCTOR…

Youssefalaa7:
Issue #14122: Recfactored DefaultLogger class to avoid CT_CONSTRUCTOR_THROW violation
made DefaultLogger final class and removed inheritance form DefaultLoggerWithCounter and BriefUtLogger

### DIFF
--- a/config/spotbugs-exclude.xml
+++ b/config/spotbugs-exclude.xml
@@ -300,11 +300,15 @@
   <Match>
     <!-- until https://github.com/checkstyle/checkstyle/issues/14122 -->
     <Or>
-      <Class name="com.puppycrawl.tools.checkstyle.DefaultLogger"/>
       <Class name="com.puppycrawl.tools.checkstyle.PackageObjectFactory"/>
       <Class name="com.puppycrawl.tools.checkstyle.XmlLoader"/>
       <Class name="com.puppycrawl.tools.checkstyle.Checker"/>
     </Or>
+    <Bug pattern="CT_CONSTRUCTOR_THROW"/>
+  </Match>
+  <Match>
+    <!-- external projects inherits from it so it should not be changed  -->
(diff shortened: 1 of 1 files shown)


Youssefalaa7:
[ERROR] /home/semaphore/checkstyle/.ci-temp/methods-distance/methods-distance/src/test/java/com/github/sevntu/checkstyle/domain/BaseCheckTestSupport.java:[133,17] cannot inherit from final com.puppycrawl.tools.checkstyle.DefaultLogger

@romani  The problem is external project inherits our DefaultLogger class 
so what do you suggest?

vivek-0509:
As the CI failure shows, the methods-distance project inherit from it, so making it final would be a breaking change for downstream users.
I recommend keeping DefaultLogger non-final and keeping the suppression with a justification comment, breaking external consumers is not worth satisfying this SpotBugs rule.

Youssefalaa7:
@vivek-0509 Thanks, that's true i think this can't be done yes, i left it only so that if @romani sees a different way for making this or only tell me to only leave it and right a comment to let others know the reason of not removing it

romani:
Please search in GitHub "com.puppycrawl.tools.checkstyle.DefaultLogger" just curious who else already using this class? 

In there are a lot, it not worth it, and we can permanent suppression comment in spotbugs config.

Youssefalaa7:
i think there is a lot using it yes so i will suppress it in spotbugs config.
Screenshot from 2026-01-23 05-36-45

Youssefalaa7:
@romani Done

romani:
Usage is massive https://github.com/search?q=%22com.puppycrawl.tools.checkstyle.DefaultLogger%22+language%3AJava&type=code&l=Java

Let's keep it suppressed, too late to change.

PR #18697, opened 2026-01-20
Issue #14122: Recfactored class Checker to avoid CT_CONSTRUCTOR_THROW…

Youssefalaa7:
Issue #14122: Recfactored Checker class to avoid CT_CONSTRUCTOR_THROW violation


Youssefalaa7:
@romani I moved the rest of the classes to be permanently suppressed because all of these files are used a lot in other projects

PR #18703, opened 2026-01-20
Issue #16155: use new nio jdk api in SuppressWithPlainTextCommentFilter

Youssefalaa7:
Issue #16155: use new nio jdk api in SuppressWithPlainTextCommentFilter


romani (approved):
Thanks a lot.
In som other PR we should put FileText ctor that accepts Path

PR #18705, opened 2026-01-20
Issue #16155: use new nio jdk api in XmlMetaWriter

Youssefalaa7:
Issue #16155: use new nio jdk api in XmlMetaWriter


romani (approved):
Thanks a lot

Issue #11602, opened 2022-05-03
 Convert regressions that use maven-checkstyle-plugin to CLI based


rnveach:
See https://issues.apache.org/jira/browse/MCHECKSTYLE-346 .
We need to deprecate all usage of maven-checkstyle-plugin so we can start implementing backward breaking changes.

A lot of regressions use maven-checkstyle-plugin to run checkstyle on the projects and will have to be converted to using the raw CLI.
Example is anything using `checkstyle:check` via a maven command. https://github.com/checkstyle/checkstyle/search?l=Shell&q=%22checkstyle%3Acheck%22

This may be difficult for some projects as they use custom dependencies. We may have to start discussions with the project owners on how we can best do this and disable them until we can get things worked out.

vivek-0509:
I am on it

vivek-0509:
Hazelcast looks a bit complex but let's try it out

romani:
We can keep some complicated on maven execution.
We don't have immediate plans to start breaking changes, but even we do so and some execution is too complicated, we can disable it until plugin is updated.

Let's migrate only simple.

vivek-0509:
@romani I investigated Hazelcast a bit more and I believe it is actually convertible. The only thing we need to handle is that it uses two license headers:
- ClassHeaderApache.txt (Apache License 2.0) - used by most modules
- ClassHeaderHazelcastCommunity.txt (Hazelcast Community License) - used only by hazelcast-sql

Everything else (checkstyle.xml etc) is identical across all modules.
I tested it locally and it works. The conversion just requires running Checkstyle CLI twice with different properties files pointing to the appropriate header. Should I raise a PR for it? It does not seem that complex after working a bit more on it.

romani:
you can send them a hint to use https://checkstyle.org/checks/header/multifileregexpheader.html#MultiFileRegexpHeader :)

please send PR to do conversion.

PR #18717, opened 2026-01-21
Issue #11602: Convert no-error-htmlunit regression to CLI

vivek-0509:
Issue #11602:

This PR converts the `no-error-htmlunit` regression test from using `maven-checkstyle-plugin` to using the Checkstyle CLI directly.

PR #18719, opened 2026-01-21
Issue #11602: Convert no-error-orekit regression to CLI

vivek-0509:
Issue #11602: 

Converts the no-error-orekit regression test to use Checkstyle CLI directly, removing the dependency on maven-checkstyle-plugin.

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -777,7 +777,7 @@ no-error-pgjdbc)
 no-error-orekit)
   CS_POM_VERSION="$(getCheckstylePomVersion)"
   echo CS_version: "${CS_POM_VERSION}"
-  ./mvnw -e --no-transfer-progress clean install -Pno-validations
+  ./mvnw -e --no-transfer-progress clean package -Passembly,no-validations
   echo "Checkout target sources ..."
   checkout_from https://github.com/Hipparchus-Math/hipparchus.git
   cd .ci-temp/hipparchus
@@ -794,8 +794,12 @@ no-error-orekit)
   # git checkout $(git describe --abbrev=0 --tags)
   git fetch --depth 1 origin "9b121e504771f3ddd303ab""cc""c74ac9db64541ea1"
   git checkout "9b121e504771f3ddd303ab""cc""c74ac9db64541ea1"
-  mvn -e --no-transfer-progress compile checkstyle:check \
-    -Dorekit.checkstyle.version="${CS_POM_VERSION}"
+  echo "checkstyle.header.file=license-header.txt" > checkstyle.properties
+  readarray -t files < <(find src/main/java -name "*.java")
+  java -jar "../../target/checkstyle-${CS_POM_VERSION}-all.jar" \
+    -c checkstyle.xml \
+    -p checkstyle.properties \
+    "${files[@]}"
   cd ..
   removeFolderWithProtectedFiles Orekit
   removeFolderWithProtectedFiles hipparchus
--- a/config/jsoref-spellchecker/whitelist.words
+++ b/config/jsoref-spellchecker/whitelist.words
@@ -327,7 +327,6 @@ doclint
 doctype
 documentationcomments
 donotignoreexceptions
-Dorekit
 Dorg
 DOTALL
 doubletag
@@ -1158,6 +1157,7 @@ rcurly
 rdiachenko
 RDz
 reactivex
+readarray
 README
 Readonly
 rebased


vivek-0509:
@romani CI failure is unrelated

.ci/validation.sh
```diff
@@ -794,8 +794,12 @@ no-error-orekit)
   # git checkout $(git describe --abbrev=0 --tags)
   git fetch --depth 1 origin "9b121e504771f3ddd303ab""cc""c74ac9db64541ea1"
   git checkout "9b121e504771f3ddd303ab""cc""c74ac9db64541ea1"
-  mvn -e --no-transfer-progress compile checkstyle:check \
-    -Dorekit.checkstyle.version="${CS_POM_VERSION}"
+  echo "checkstyle.header.file=license-header.txt" > checkstyle.properties
+  # shellcheck disable=SC2046
+  java -jar ../../target/checkstyle-"${CS_POM_VERSION}"-all.jar \
```
romani: why we should suppress ?

vivek-0509: suppression is necessary because ShellCheck suggested fix (quoting the command substitution) which would break the command. The Checkstyle CLI expects each Java file as a separate argument, but quoting $(find ) causes all filenames to be passed as a single concatenated string. Without quotes, word splitting correctly passes each file as an individual argument and since find only returns *.java files, and valid Java source filenames cannot contain spaces adding suppression is safe for our use case.

romani: .may be it wants "../../target/checkstyle-${CS_POM_VERSION}-all.jar" ?

vivek-0509: I got this warning in the CI, so we actually need the suppression and about the jar path quoting i tried locally the shell check shows no warning .
```
In ./.ci/validation.sh line 941:
    $(find src/main/java src/test/java -name "*.java")
    ^-- SC2046 (warning): Quote this to prevent word splitting.
```
Screenshot 2026-01-26 at 3 28 04 AM

romani: https://www.shellcheck.net/wiki/SC2046

Hmm, tool will be right if find return single result , but it returns array of strings.

vivek-0509: Done, Thanks for the guidance, you always make me learn something new

vivek-0509: @romani i have added  `readarray` in the whitelist in this PR  https://github.com/checkstyle/checkstyle/pull/18717 please merge that first so CI will pass.

romani: From AI:

ShellCheck SC2046 warns that unquoted command substitution (like $(find ...)) causes word splitting and globbing, which breaks if filenames contain spaces or special characters. Fix it by quoting the command substitution (e.g., command "$(find ...)") or by using while read or -exec to handle filenames properly. 
Common Scenario & Fixes
Wrong (Causes SC2046): ls -l $(find . -name "*.txt")
Correct: find . -name "*.txt" -exec ls -l {} + (Robust, handles spaces)
Alternative: find . -name "*.txt" -print0 | xargs -0 ls -l
If you must use $(...): files=$(find . -name "*.txt"); ls -l "$files" (only if single file, otherwise still fails) 
Why this happens
When find outputs multiple files, the unquoted $() splits them by whitespace into separate arguments. Quoting "$()" treats the entire output as a single string, which is often not what you want, hence the need for -exec or read. 

------


Correct: find . -name "*.txt" -exec ls -l {} + (Robust, handles spaces)

But this will make command very hard to read.
Or maybe we can store result of find in array? And will use array in java arguments

vivek-0509: Yes, exactly, Updated to use the array approach

```
readarray -t files < <(find src/main/java -name "*.java")
java -jar "../../target/checkstyle-${CS_POM_VERSION}-all.jar" \
    -c checkstyle.xml \
    -p checkstyle.properties \
    "${files[@]}"
```
This stores find results in an array and expands it properly as arguments

.ci/validation.sh
```diff
@@ -794,8 +794,12 @@ no-error-orekit)
   # git checkout $(git describe --abbrev=0 --tags)
   git fetch --depth 1 origin "9b121e504771f3ddd303ab""cc""c74ac9db64541ea1"
   git checkout "9b121e504771f3ddd303ab""cc""c74ac9db64541ea1"
-  mvn -e --no-transfer-progress compile checkstyle:check \
-    -Dorekit.checkstyle.version="${CS_POM_VERSION}"
+  echo "checkstyle.header.file=license-header.txt" > checkstyle.properties
+  readarray -t files < <(find src/main/java -name "*.java")
```
romani, line 798: https://www.baeldung.com/linux/reading-output-into-array#using-theread-command

https://linuxopsys.com/bash-readarray-with-examples (`-t: Remove a trailing delimiter from each line read. The default value is a newline.`)

All good here.

vivek-0509:
It works tested locally, triggered on CI as well .
Screenshot 2026-01-27 at 7 00 43 PM

https://checkstyle.semaphoreci.com/jobs/261759a9-598a-41f2-ab30-e35aaf4025ad
works well in CI

PR #18736, opened 2026-01-23
Issue #18599: Resolve error-prone violations

Youssefalaa7:
Issue #18599: Resolve error-prone violations for NullableOptional
this was the warning before modification
[WARNING] /home/youssef/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java:[774,37] [NullableOptional] Using an Optional variable which is expected to possibly be null is discouraged. It is best to indicate the absence of the value by assigning it an empty optional.
    (see https://errorprone.info/bugpattern/NullableOptional)

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
@@ -770,7 +770,6 @@ public final class SiteUtil {
      * @param propertyJavadoc the property Javadoc to extract the since version from.
      * @return the Optional of property version specified in its javadoc.
      */
-    @Nullable
     private static Optional<String> getPropertyVersionFromItsJavadoc(DetailNode propertyJavadoc) {
         final Optional<DetailNode> propertyJavadocTag =
             getPropertySinceJavadocTag(propertyJavadoc);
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
```diff
@@ -770,7 +770,6 @@ && isVersionAtLeast(propertySetterSince, moduleSince)) {
      * @param propertyJavadoc the property Javadoc to extract the since version from.
      * @return the Optional of property version specified in its javadoc.
      */
-    @Nullable
     private static Optional<String> getPropertyVersionFromItsJavadoc(DetailNode propertyJavadoc) {
```
romani, line 773: what rule can be activated?

Please activate NullableOptional as error

Youssefalaa7: i thought the issue was only to remove the reason of warning the only change i added was to put this an error in the pom.xml right?

romani: Remove and close a door for any other to appear

PR #18737, opened 2026-01-23
Issue #18599: Resolve error-prone violations for UnsynchronizedOverri…

Youssefalaa7:
Issue #18599: Resolve error-prone violations for UnsynchronizedOverridesSynchronized
this was the warning before modification
[WARNING] /home/youssef/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/checks/[OrderedPropertiesCheck.java](http://orderedpropertiescheck.java/):[212,36] [UnsynchronizedOverridesSynchronized] Unsynchronized method keys overrides synchronized method in Hashtable     (see https://errorprone.info/bugpattern/UnsynchronizedOverridesSynchronized)

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/OrderedPropertiesCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/OrderedPropertiesCheck.java
@@ -207,9 +207,14 @@ public class OrderedPropertiesCheck extends AbstractFileSetCheck {
 
         /**
          * Returns a copy of the keys.
+         *
+         * @noinspection SynchronizedMethod
+         * @noinspectionreason SynchronizedMethod - synchronized keyword is required
+         *      to override the synchronized keys() method from Hashtable parent class,
+         *      maintaining thread-safety contract
          */
         @Override
-        public Enumeration<Object> keys() {
+        public synchronized Enumeration<Object> keys() {
             return Collections.enumeration(keyList);
         }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/OrderedPropertiesCheck.java
```diff
@@ -207,9 +207,14 @@ private static final class SequencedProperties extends Properties {
...
          * Returns a copy of the keys.
+         *
+         * @noinspection SynchronizedMethod
+         * @noinspectionreason SynchronizedMethod - synchronized keyword is required
+         *      to override the synchronized keys() method from Hashtable parent class,
+         *      maintaining thread-safety contract
          */
         @Override
-        public Enumeration<Object> keys() {
+        public synchronized Enumeration<Object> keys() {
```
romani, line 217: It is from https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/classes/java/util/Hashtable.java#L256
https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/classes/java/util/Properties.java#L122
`SequencedProperties extends Properties` that extends Hashtable .

In some future we probably need to refactor this to avoid such extension, check is not multi thread instance.

Youssefalaa7:
> UnsynchronizedOverridesSynchronized
i added it as an error in the pom.xml this is the only violation

Issue #17425, opened 2025-07-19
jdk21: activate UnusedCatchParameterShouldBeUnnamed for checkstyle project


romani:
after migration to jdk21
activate UnusedCatchParameterShouldBeUnnamed
https://checkstyle.sourceforge.io/checks/coding/unusedcatchparametershouldbeunnamed.html#UnusedCatchParameterShouldBeUnnamed

and follow it.

PR #18739, opened 2026-01-23
Issue #17425: activate UnusedCatchParameterShouldBeUnnamed for checks…

Youssefalaa7:
Issue #17425: activate UnusedCatchParameterShouldBeUnnamed for checkstyle project

### DIFF
--- a/.ci/check-performance-regression.sh
+++ b/.ci/check-performance-regression.sh
@@ -15,7 +15,7 @@ CONFIG_FILE=$2
 # max difference tolerance in %
 THRESHOLD_PERCENTAGE=10
 # JDK version
-JDK_VERSION=25
+JDK_VERSION=21
 # sample project path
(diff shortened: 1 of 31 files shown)


Youssefalaa7:
@romani enable preview causes too much problems on cl given that it work succesfully when running things local any suggestions?

Youssefalaa7:
@romani  I tried solving the errors by writing --enable-preview in the plugins in our project but there are other errors that comes from other projects because they are not allowing --enable-preview in their code such as he [xwiki-commons](https://github.com/xwiki/xwiki-commons) project so i think these checks we will not able to do it at the current time we should wait for upgrading to jdk 22 and the changes at that time will be very easy without the need to care about plugins and --enable-preview

romani:
Sorry my bad.
We can not use it as it is jfk22 feature, we are jdk21 we can not use this syntax sugar for now 

I updated issue. Sorry, my bad.

Youssefalaa7:
no problem All good

Issue #13351, opened 2023-07-03
Resolve violations from spotbugs sb-contrib 7.6.0


romani:
Resolve violations from spotbugs sb-contrib 7.6.0

we need to remove suppression that we introduced at https://github.com/checkstyle/checkstyle/pull/12855
PLEASE do not resolve all violations by one PullRequest, fix by groups as some violation might stay in permanent suppression 

```
$ mvn spotbugs:check
......
[INFO] --- spotbugs-maven-plugin:4.7.3.0:check (default-cli) @ checkstyle ---
[INFO] BugInstance size is 6
[INFO] Error size is 0
[INFO] Total bugs: 6
[ERROR] Medium: Method com.puppycrawl.tools.checkstyle.ConfigurationLoader
$InternalLoader.endElement(String, String, String) 
sets an enum reference to null
 [com.puppycrawl.tools.checkstyle.ConfigurationLoader$InternalLoader] 
At ConfigurationLoader.java:[line 571] ENMI_NULL_ENUM_VALUE

[ERROR] Medium: Enum class com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocTagInfo
 only declares one enum value
 [com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocTagInfo]
 At JavadocTagInfo.java:[line 389] ENMI_ONE_ENUM_VALUE

[ERROR] Medium: Method call in 
com.puppycrawl.tools.checkstyle.gui.ParseTreeTableModel.setParseTree(DetailAST)
 passes object that the method is called on as a parameter
 [com.puppycrawl.tools.checkstyle.gui.ParseTreeTableModel] 
At ParseTreeTableModel.java:[line 65] SPP_PASSING_THIS_AS_PARM

[ERROR] Medium: Method 
com.puppycrawl.tools.checkstyle.utils.CheckUtil.getSurroundingAccessModifier(DetailAST)
 sets an enum reference to null
 [com.puppycrawl.tools.checkstyle.utils.CheckUtil]
 At CheckUtil.java:[line 492] ENMI_NULL_ENUM_VALUE

[ERROR] Medium: Method 
com.puppycrawl.tools.checkstyle.utils.ScopeUtil.getDeclaredScopeFromMods(DetailAST) 
sets an enum reference to null 
[com.puppycrawl.tools.checkstyle.utils.ScopeUtil] 
At ScopeUtil.java:[line 46] ENMI_NULL_ENUM_VALUE

[...]

PR #18741, opened 2026-01-24
Issue #13351: Resolve violations from spotbugs sb-contrib for ENMI_NU…

Youssefalaa7:
Issue #13351: Resolve violations from spotbugs sb-contrib for ENMI_NULL_ENUM_VALUE
removed one of the classes that was void function and the other 2 classes can not be changed because these are for public apis that we canot use optional in them as this will break the usage of external users

### DIFF
--- a/config/spotbugs-exclude.xml
+++ b/config/spotbugs-exclude.xml
@@ -246,7 +246,6 @@
   <Match>
     <!-- Till https://github.com/checkstyle/checkstyle/issues/13351 -->
     <Or>
-      <Class name="com.puppycrawl.tools.checkstyle.ConfigurationLoader$InternalLoader"/>
       <Class name="com.puppycrawl.tools.checkstyle.utils.CheckUtil"/>
       <Class name="com.puppycrawl.tools.checkstyle.utils.ScopeUtil"/>
     </Or>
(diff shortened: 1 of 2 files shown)


config/spotbugs-exclude.xml
```diff
@@ -248,9 +248,9 @@
     <Bug pattern="CNC_COLLECTION_NAMING_CONFUSION"/>
   </Match>
   <Match>
-    <!-- Till https://github.com/checkstyle/checkstyle/issues/13351 -->
     <Or>
-      <Class name="com.puppycrawl.tools.checkstyle.ConfigurationLoader$InternalLoader"/>
+      <!-- These are api that we can not change it to use optional
+       as it will affect external users -->
       <Class name="com.puppycrawl.tools.checkstyle.utils.CheckUtil"/>
       <Class name="com.puppycrawl.tools.checkstyle.utils.ScopeUtil"/>
```
romani: util classes are not api, lets try fix them

romani: I suggest to keep "till" and in this PR simply merge fix for InternalLoader

Youssefalaa7: Done i will try to work on util on another PR as suggested

Issue #14121, opened 2023-12-08
IndentationCheck: inner Handlers should not have reference to Check itself.


romani:
detected at https://github.com/checkstyle/checkstyle/pull/14117?notification_referrer_id=NT_kwDOAAxnuLE4Njc0Mzg0MTcxOjgxMjk4NA#issuecomment-1844198684

there should be no cycle in dependecy between IndentaitonCheck and its XxxxxHandler.
There should be no public methods in IndentaitonCheck that some other class calls, except for methods that settters and vitistToken.

MANISH-K-07:
@romani ,

> There should be no public methods in IndentaitonCheck that some other class calls, except for methods that settters and vitistToken.

This part I understand just fine

> there should be no cycle in dependecy between IndentaitonCheck and its XxxxxHandler.

Could you elaborate a little bit on what you are expecting here?

smita1078:
I am on it

Anushreebasics:
I am on it

Youssefalaa7:
I am on it

PR #18754, opened 2026-01-25
Issue #18843: Indentation Check Handlers should not have reference to…

### LINKED ISSUE #18843: Fix SpotBugs UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR warning in IndentationCheck
SpotBugs reports a `UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR` warning for the `incorrectIndentationLines `field in `IndentationCheck`. This field is not initialized in the constructor, which violates best practices.

Current Problem
```
private Set<Integer> incorrectIndentationLines;

@Override
public void beginTree(DetailAST rootAST) {
    incorrectIndentationLines = new HashSet<>();
    // ...
}
```
The field is assigned in `beginTree()` instead of being initialized at declaration or in the constructor.

This issue addresses only the SpotBugs warning. It does not fully resolve the circular dependency problem described in #14121, which requires broader architectural refactoring.

Related Issues
Part of #14121 (circular dependency elimination)

smita1078:
Fixes #18843 
Part of #14121

Changes

- Initialize incorrectIndentationLines eagerly to fix SpotBugs `UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR` warning
- Changed to `private final Set<Integer> incorrectIndentationLines = new HashSet<>()`
- Updated `beginTree() `to use clear() instead of reassignment
- Remove SpotBugs suppression for this violation

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java
@@ -145,7 +145,7 @@ public class IndentationCheck extends AbstractCheck {
     private final HandlerFactory handlerFactory = new HandlerFactory();
 
     /** Lines logged as having incorrect indentation. */
-    private Set<Integer> incorrectIndentationLines;
+    private final Set<Integer> incorrectIndentationLines = new HashSet<>();
 
     /** Specify how far new indentation level should be indented when on the next line. */
     private int basicOffset = DEFAULT_INDENTATION;
@@ -354,7 +354,6 @@ public class IndentationCheck extends AbstractCheck {
         clearState();
         final PrimordialHandler primordialHandler = new PrimordialHandler(this);
         handlers.push(primordialHandler);
-        incorrectIndentationLines = new HashSet<>();
     }
 
     @Override
@@ -376,6 +375,7 @@ public class IndentationCheck extends AbstractCheck {
     private void clearState() {
         handlerFactory.clearCreatedHandlers();
         handlers.clear();
+        incorrectIndentationLines.clear();
     }
 
     /**
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -5763,13 +5763,6 @@
   </checkerFrameworkError>
 
   <checkerFrameworkError unstable="false">
-    <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java</fileName>
-    <specifier>initialization.field.uninitialized</specifier>
-    <message>the default constructor does not initialize field incorrectIndentationLines</message>
-    <lineContent>private Set&lt;Integer&gt; incorrectIndentationLines;</lineContent>
-  </checkerFrameworkError>
-
-  <checkerFrameworkError unstable="false">
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/LineWrappingHandler.java</fileName>
     <specifier>not.interned</specifier>
     <message>attempting to use a non-@Interned comparison operand</message>
--- a/config/spotbugs-exclude.xml
+++ b/config/spotbugs-exclude.xml
@@ -266,12 +266,6 @@
     <Bug pattern="UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR"/>
   </Match>
   <Match>
-    <!-- until https://github.com/checkstyle/checkstyle/issues/14121 -->
-    <Class name="com.puppycrawl.tools.checkstyle.checks.indentation.IndentationCheck"/>
-    <Method name="indentationLog"/>
-    <Bug pattern="UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR"/>
-  </Match>
-  <Match>
     <!-- there is strict order of method executions, call in misorder will result in failure -->
     <Class name="com.puppycrawl.tools.checkstyle.AbstractAutomaticBean"/>
     <Method name="setupChild"/>


config/pitest-suppressions/pitest-indentation-suppressions.xml
```diff
@@ -1,3 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
+  <mutation unstable="false">
+    <sourceFile>IndentationCheck.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.checks.indentation.IndentationCheck</mutatedClass>
+    <mutatedMethod>beginTree</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.VoidMethodCallMutator</mutator>
+    <description>removed call to java/util/Set::clear</description>
+    <lineContent>incorrectIndentationLines.clear();</lineContent>
+  </mutation>
```
vivek-0509: Please do not add the pitest suppression. Instead, you can move the incorrectIndentationLines.clear() call into the clearState() method, since it serves the same purpose as the other method calls inside that method, clearState() is excluded from pitest so  the mutation will not be generated and no suppression is needed.

vivek-0509 (changes requested):

config/pitest-suppressions/pitest-indentation-suppressions.xml
smita1078: Done

pom.xml
```diff
@@ -4013,6 +4013,7 @@
               <excludedMethods>
                 <param>clearCreatedHandlers</param>
                 <param>clearState</param>
+                <param>logOncePerLine</param>
               </excludedMethods>
```
vivek-0509, line 4019: why do we need to add logOncePerLine in the excludedMethods, i think its not needed , what pitest mutation made you do so ??

romani: We can not simply exclude from pitest, we need to cover by tests.

smita1078: Yes removed, from previous run VoidMethodCallMutator was there, now its not coming anymore

romani:
@vivek-0509 , please finish review, I will review after you.

vivek-0509:
@romani
I have reviewed the PR.
From what I see, this fixes the SpotBugs warning and removes handlers reliance on a public behavioral method by making the logging method package-private.
One point I wanted to clarify, is the intent of the issue to eliminate the class-level reference from handlers to 
IndentationCheck entirely, or to ensure handlers don't depend on the check's public API contract ?
This PR keeps the reference but limits interaction to package-private scope and read-only configuration access. Just wanted to confirm that this matches the intended scope since issue feels little unclear to me.

romani:
Goal is to avoid cycle dependencies between classes.
Check can be aware of all handlers and orchestrate them. Handlers should not know anything about Check and just give all data/behavior to Check.

vivek-0509:
Currently, this PR addresses the SpotBugs warning but does not remove the cyclic dependency between IndentationCheck
 and handlers. To completely remove the cyclic dependency, we would need significant refactoring.

From my overview, I think we could create an immutable context record containing all the configuration details that handlers need (basicOffset, caseIndent, throwsIndent, etc.). Handlers would use this config to perform their indentation checks. Additionally, we can have a violation context where handlers log violations into a list, then return that list. The Check would then collect these violations and log them.

However, I need to think more about how to achieve this properly. 
@smita1078 Please think in this direction, or if you have a better approach, please share it. We can discuss and then move forward with the refactoring.

smita1078:
> Currently, this PR addresses the SpotBugs warning but does not remove the cyclic dependency between IndentationCheck and handlers. To completely remove the cyclic dependency, we would need significant refactoring.
> 
> From my overview, I think we could create an immutable context record containing all the configuration details that handlers need (basicOffset, caseIndent, throwsIndent, etc.). Handlers would use this config to perform their indentation checks. Additionally, we can have a violation context where handlers log violations into a list, then return that list. The Check would then collect these violations and log them.
> 
> However, I need to think more about how to achieve this properly. @smita1078 Please think in this direction, or if you have a better approach, please share it. We can discuss and then move forward with the refactoring.

@vivek-0509 
Thanks for the detailed analysis! I agree with your approach of using an immutable context record for configuration. 
For the violation handling, instead of having handlers return a list of violations, I'd think we can use a callback interface pattern.
Create a `ViolationReporter` interface with a single method: `logOncePerLine` then `IndentationCheck` would implement this interface. Handlers would receive this interface via constructor: `Handler(IndentationConfig config, ViolationReporter reporter, ...)` this would break circular dependency as handlers depend on interface, not Check

romani:
Ok, sounds like this issue is epic, and will not be easy, but doable.

Let's split it in few issues and make sure we can reliably do refactoring.

For this PR, let's make special issue to describe small goal and merge this update, as step forward.

vivek-0509:
@smita1078 Please create an issue for this PR (the current SpotBugs fix) so we can merge it as a step forward. We need to decide on the approach before proceding with refactoring ,will continue the discussion at https://github.com/checkstyle/checkstyle/issues/14121.

smita1078:
> @smita1078 Please create an issue for this PR (the current SpotBugs fix) so we can merge it as a step forward. We need to decide on the approach before proceding with refactoring ,will continue the discussion at #14121.

@vivek-0509 Created issue for this PR.  https://github.com/checkstyle/checkstyle/issues/18843 
The scope is limited to:
Fix SpotBugs `UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR` warning
Reduce public API surface (`public → package-private` for logging method)
This doesn't fully resolve #14121. Will continue the architectural refactoring discussion at #14121.

romani:
@vivek-0509 , please finish review

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java
```diff
@@ -310,17 +310,16 @@ public void setLineWrappingIndentation(int lineWrappingIndentation) {
...
-     * @see java.text.MessageFormat
+     * @param ast the AST node for which the error is logged
+     * @param key the message key describing the violation
+     * @param args optional arguments for the message
      */
-    public void indentationLog(DetailAST ast, String key, Object... args) {
-        if (!incorrectIndentationLines.contains(ast.getLineNo())) {
-            incorrectIndentationLines.add(ast.getLineNo());
+    /* package */ void logOncePerLine(DetailAST ast, String key, Object... args) {
+        if (incorrectIndentationLines.add(ast.getLineNo())) {
```
vivek-0509: Please revert the visibility/rename changes to indentationLog method. This PR should only fix the SpotBugs warning for the scope of issue #18843. Refctoring changes will be done separately once we agree on the approach.

vivek-0509 (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java
smita1078: Done

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/LineWrappingHandler.java
```diff
@@ -457,5 +457,4 @@ private void logWarningMessage(DetailAST currentNode, int currentIndent) {
             }
         }
     }
-
```
vivek-0509, line 460: please revert any unrelated change

vivek-0509 (changes requested):

vivek-0509:
@romani PR looks good ok to merge

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/LineWrappingHandler.java
smita1078: Done, Sry! missed it earlier

romani:
@vivek-0509 , please finish review

vivek-0509 (approved):

vivek-0509:
@romani Already done https://github.com/checkstyle/checkstyle/pull/18754#issuecomment-3847558779

romani:

romani (approved):
Thanks a lot

Issue #17715, opened 2025-08-31
New Check: WhitespaceBeforeEmptyBody


mohitsatr:
continuation of: #10834 

WhitespaceAround has a lot of properties allowEmptyXxxxxx (they has 'true' in google_checks.xml) that skip violating on code like:
```
/** some. */
public class Test {
  void returnTrue(){}
}
```
Google config:
https://github.com/checkstyle/checkstyle/blob/488018b278b33b5be6ac739846b5960b709d5ab0/src/main/resources/google_checks.xml#L165-L172

---------

WhitespaceBeforeEmptyBody Check should check that a space is present before Empty body of a class, method, interface, static,  constructor , loop(while, do, for) and record.

https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

one whitespace is required:
> Before any open curly brace ({), with two exceptions:
    @SomeAnnotation({a, b}) (no space is used)
    String[][] x = {{"foo"}}; (no space is required between {{, by item 9 below)

Some of the cases are already covered:

```java
/** some javadoc. */
public class Test {

 
  Test(){}  // ok

  /** some javadoc. */
  public static void main(String... args) {

    boolean b = System.currentTimeMillis() < 0;

    do{} while (b);  // ok 
  
    Runnable noop = () ->{};  // ok
  }


  record Record2(String str) {
    public Record2{} // ok
  }
}
```

```
$ java -jar checkstyle-11.0.0-all.jar -c google_checks.xml Test.java 
Starting audit...
[WARN] /mnt/5D92528E6B945467/test/testing/Test.java:7:30: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /mnt/5D92528E6B945467/test/testing/Test.java:14:5: 'do' is not followed by whitespace. [WhitespaceAfter]
[WARN] /mnt/5D92528E6B945467/test/testing/Test.java:14:5: WhitespaceAround: 'do' is not followed by whitespace. Empty blocks                may only be represented as {} when not part of a multi-block statement (4.1.3) [WhitespaceAround]
[WARN] /mnt/5D92528E6B945467/test/testing/Test.java:19:24: '->' is not followed by whitespace. [WhitespaceAfter]
[...]

PR #18763, opened 2026-01-26
Issue #17715: Add new WhitespaceBeforeEmptyBody check

vivek-0509:
Issue #17715:

### Description

This PR introduces a new check WhitespaceBeforeEmptyBodyCheck that verifies whitespace is present before the opening brace of an empty body.

### Rule Reference

This rule is defined by the [Google Java Style Guide (§4.6.2 Horizontal whitespace)](https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace) which states that "a single ASCII space must appear before any open curly brace (`{`)".

### Supported Tokens

The check validates space before empty bodies of:
- Methods, constructors, and compact constructors
- Classes, interfaces, enums, records, and annotations
- Loops (while, for, do-while)
- Lambdas and anonymous classes
- Static Initializer blocks 
- Try-catch-finally statements
- Synchronized blocks and switch statements

New module config: https://gist.githubusercontent.com/vivek-0509/8d884eaddeae4e2652838d0b78db8efd/raw/f1e0daf9412d2f07ac1ed385939f89f93ea70170/whitespaceBeforeEmptyBody.xml

Contirbution repo PR: https://github.com/checkstyle/contribution/pull/1023


vivek-0509:
GitHub, generate report

vivek-0509:
@romani PR is ready for review .
I have reviewed the diff report all the violation are genuine.
CI failures are unrelated

vivek-0509:
@romani ping

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceBeforeEmptyBodyCheck.java
```diff
@@ -0,0 +1,190 @@
...
+ * </div>
+ *
+ * <p>
+ * This check is applied to methods, constructors, classes, interfaces, enums, records,
+ * annotation definitions, loops (while, for, do-while), lambdas, anonymous classes,
+ * initializer blocks (static and instance), try-catch-finally statements, synchronized blocks,
+ * and switch statements.
+ * </p>
+ *
+ * @since 13.1.0
```
romani: Please update all your PRs for new stuff to mention 13.2.0

romani (changes requested):
Quick look

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceBeforeEmptyBodyCheck.java
vivek-0509: done

vivek-0509:
@romani done removed the Instance_init token

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/whitespacebeforeemptybody/Example1.java
```diff
@@ -0,0 +1,59 @@
...
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.whitespace.whitespacebeforeemptybody;
+
+// xdoc section -- start
+class Example1 {
+  Example1(){} // violation 'Whitespace is not present before the empty body'
+
+  void method(){} // violation 'Whitespace is not present before the empty body'
+
```
romani, line 18: We found this model more clear for users 
```
void methodWarn(){} // violation
void methodGood() {}
```
Please apply this model to examples.
It is completely ok to skip some cases of examples becomes too big

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -265,6 +265,7 @@ public class XdocsExamplesAstConsistencyTest {
             "checks/whitespace/whitespacearound/Example7",
             "checks/whitespace/whitespacearound/Example8",
             "checks/whitespace/whitespacearound/Example9",
+            "checks/whitespace/whitespacebeforeemptybody/Example2",
```
romani: Hm ? We do effort now to not make any new problems?
Can we keep it consistent?

romani (changes requested):
Impressive.

Minors:

src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/whitespace/whitespacebeforeemptybody/Example1.java
vivek-0509: done

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
vivek-0509: done

romani:
GitHub, generate website

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceBeforeEmptyBodyCheck.java
```diff
@@ -0,0 +1,188 @@
...
+     * Gets the name of the construct for error message.
+     *
+     * @param ast the construct AST
+     * @return the name of the construct
+     */
+    private static String getConstructName(DetailAST ast) {
+        String name = TOKEN_NAMES.get(ast.getType());
+        if (name == null) {
+            final DetailAST identToken = ast.findFirstToken(TokenTypes.IDENT);
+            name = NullUtil.notNull(identToken).getText();
```
romani, line 188: Here AST can be any token from accepted, for example TokenTypes.LITERAL_WHILE 
https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/api/TokenTypes.html#LITERAL_WHILE
Ident here unrelated variable name to loop structure.

So null should be very possible in this situation.

romani (changes requested):
Question

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceBeforeEmptyBodyCheck.java
romani: @Gaoyan1999, please share your opinion here too

romani:
Web site is good.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceBeforeEmptyBodyCheck.java
Gaoyan1999: I think it's correct.
For LITERAL_WHILE (and the other loop/keyword tokens), we never use the IDENT path: they are all in TOKEN_NAMES, so we get the name from TOKEN_NAMES.get(ast.getType()) and return. 

Only when name == null do we use findFirstToken(IDENT), and that happens only for METHOD_DEF, CTOR_DEF, etc. Those always have an IDENT in valid syntax, so identToken is not null there. That’s why I think it's correct.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceBeforeEmptyBodyCheck.java
```diff
@@ -0,0 +1,188 @@
...
+            TokenTypes.LITERAL_FOR,
+            TokenTypes.LITERAL_DO,
+            TokenTypes.STATIC_INIT,
+            TokenTypes.LITERAL_TRY,
+            TokenTypes.LITERAL_CATCH,
+            TokenTypes.LITERAL_FINALLY,
+            TokenTypes.LITERAL_SYNCHRONIZED,
+            TokenTypes.LITERAL_SWITCH,
+            TokenTypes.LAMBDA,
+            TokenTypes.LITERAL_NEW,
```
Gaoyan1999, line 98: I think TokenTypes.LITERAL_NEW could be moved to line88 (below ANNOTATION_DEF). That way, they can have the same order as TOKEN_NAMES map.

romani:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c6ee9e8_2026015739/reports/diff/apache-struts/index.html#A1

This is not empty block.

I tried to find what Google style sentences we try to cover by this Check ?
I looked around but can not find.  I will try one moy time, but if you know please point.

I lost a reason of this Check.

src/main/java/com/puppycrawl/tools/checkstyle/checks/whitespace/WhitespaceBeforeEmptyBodyCheck.java
vivek-0509: It's safe in our case, tokens like LITERAL_WHILE are all in TOKEN_NAMES, so they return early and never reach the IDENT fallback,

vivek-0509:
Hi @romani,
this check is based on: [Section 4.6.2 - Horizontal whitespace](https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace), which states that a single space is required "before any open curly brace ({)", with two exceptions: @SomeAnnotation({a, b}) and String[][] x = {{"foo"}};.

I initially thought we might need to expand this check to cover non-empty bodies as well, since the rule says "before any open curly brace", not specifically empty ones. However, after testing, I found that WhitespaceAround already catches all non-empty body cases for every token type in our check .

The gap exists specifically for empty bodies because the Google checks configuration sets allowEmptyMethods=true, allowEmptyTypes=true, allowEmptyLoops=true, allowEmptyLambdas=true, and allowEmptySwitchBlockStatements=true on WhitespaceAround. These properties are needed to satisfy Google Style rule 4.1.3 (empty blocks may be represented as {}), but as a side effect, they skip the entire whitespace check for those tokens, including the "whitespace before {" validation. So void method(){ code; } is correctly caught by WhitespaceAround, but void method(){} slips through.

That is exactly why WhitespaceBeforeEmptyBody is needed, it fills this specific gap to fully enforce the Google Style rule for empty body cases that WhitespaceAround cannot catch due to its allowEmpty configuration.

vivek-0509:
@romani Here is the CLI output 

```
/** Test. */
public class Test {

  int number;

  Test(int number){
    this.number = number;
  }

  record CompactRecord(String name){
    public CompactRecord{
      name = name.trim();
    }
  }

  class Inner{
    int age;

    void setAge(int age){
      for (int i = 0; i < 10; i++){
        this.age = age;
      }
    }
  }

  Runnable runnable = () ->{
    int val = 1;
  };

  interface Iface{
    void foo();
  }

  enum MyEnum{
    AAA,
    BBB
  }

  record MyRecord(int value){
    public int getValue(){
      return value;
    }
  }

  @interface MyAnnotation{
    String value();
  }

  static{
    int age;
  }

  void testBlocks() {
    boolean flag = true;

    while (flag){
      flag = false;
    }

    for (int i = 0; i < 1; i++){
      int val = i;
    }

    do{
      flag = !flag;
    } while (flag);

    try{
      int val = 1;
    } catch (Exception ex) {
      throw ex;
    }

    try {
      int val = 1;
    } catch (Exception ex){
      ex.printStackTrace();
    }

    try {
      int val = 1;
    } catch (Exception ex) {
      throw ex;
    } finally{
      int val = 2;
    }

    synchronized (this){
      int val = 1;
    }

    switch (1){
      case 1:
        break;
      default:
        break;
    }
  }
}
```
```

vivek@Viveks-MacBook-Air checkstyle % java -Duser.language=en -Duser.country=US -jar /Users/vivek/checkstyle/target/checkstyle-*-all.jar -c /google_checks.xml /tmp/checkstyle-test/Test.java
Starting audit...
[WARN] /tmp/checkstyle-test/Test.java:8:19: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:12:36: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:13:25: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:18:14: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:21:25: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:22:35: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:28:26: WhitespaceAround: '->' is not followed by whitespace. Empty blocks                may only be represented as {} when not part of a multi-block statement (4.1.3) [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:28:28: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:32:18: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:36:14: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:41:29: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:42:26: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:47:26: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:51:9: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:58:17: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:62:32: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:66:5: WhitespaceAround: 'do' is not followed by whitespace. Empty blocks                may only be represented as {} when not part of a multi-block statement (4.1.3) [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:66:7: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:70:5: WhitespaceAround: 'try' is not followed by whitespace. Empty blocks                may only be represented as {} when not part of a multi-block statement (4.1.3) [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:70:8: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:78:27: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:86:7: WhitespaceAround: 'finally' is not followed by whitespace. Empty blocks                may only be represented as {} when not part of a multi-block statement (4.1.3) [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:86:14: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:90:24: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
[WARN] /tmp/checkstyle-test/Test.java:94:15: WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
Audit done.
```

vivek-0509:
Before implementing this check, I analyzed the coverage for empty body cases with `WhitespaceAround` in `google_checks.xml`. Some cases are already covered by `WhitespaceAround`, but the following tokens are not covered and violations are expected but not detected: CLASS_DEF, INTERFACE_DEF, ENUM_DEF, ANNOTATION_DEF, STATIC_INIT, LITERAL_WHILE, LITERAL_FOR, LITERAL_SWITCH, LITERAL_NEW, and METHOD_DEF.

As mentioned in the issue description, we need to cover all cases with empty bodies, so I have implemented the check to handle every case. However, when adding this check to `google_checks.xml`, we can configure the tokens property to include only the uncovered tokens, to  avoid duplicate violations with `WhitespaceAround`.

vivek-0509:
GitHub, generate report

vivek-0509:
Hi @romani,
I discovered an issue with WhitespaceAroundCheck that creates a gap in coverage for bodies containing only comments as you mentioned that body with comments should not be considered  as empty bodies.

WhitespaceAround does not enable isCommentNodesRequired(), which means comments are not visible in its AST. As a result, a body that contains only a comment (like // expected ) looks identical to a truly empty body, the SLIST has only RCURLY as its child since the comment node is invisible. When the allowEmpty flags are enabled in the Google configuration (allowEmptyMethods=true etc), WhitespaceAround treats these comment only bodies as empty and suppresses the violation entirely.

At the same time, WhitespaceBeforeEmptyBody correctly enables isCommentNodesRequired(), so it sees the comment nodes and properly identifies these bodies as non-empty, meaning it does not flag them (since it only handles empty bodies). The result is that neither check catches the missing whitespace before { in cases like these:

```
while (flag){
    // comment inside
}

class InnerWithComment{
    // comment inside
}

enum MyEnum{
    // comment inside
}
```
In all of these cases, there is no space before {, which violates the Google Style Guide rule 4.6.2. However, WhitespaceAround skips them because it thinks they are empty, and WhitespaceBeforeEmptyBody skips them because they are actually not empty. So the violation goes undetected.

Should I open a separate issue to fix WhitespaceAround to consider bodies with only comments as non-empty?

vivek-0509:
@romani ping

vivek-0509:
Hi @romani , just checking in so we don’t lose track of this. Thanks

vivek-0509:
@romani  Rebased the branch and resolved  the merge conflicts

vivek-0509:
I am getting the same spelling CI failure in both the PRs https://github.com/checkstyle/checkstyle/pull/18312 (fixed for this one) to add `args` in the whitelist, but both of my PRs have no word args, i think my PR makes changes in the checks.xml so may be it is scanning that complete file and that file has args which is not present in the whitelist.
Will fix this CI, after the PR is reviewed, avoiding to not burn CI credits for this

vivek-0509:
@romani ping

vivek-0509:
@romani  Ci failures are unrelated, will fix the spelling one once pr is reviewed, to avoid burning Unnecessary ci credits

vivek-0509:
@romani ping

vivek-0509:
@romani  ping

vivek-0509:
@romani  ping

vivek-0509:
@Zopsss @romani the PR is related to Google Style project..
Closing this as discussed with Mahfouz
@Atharv3221 You can reuse the work since the PR is in very good state

PR #18764, opened 2026-01-26
Issue #18599: Resolve error-prone violations

Brijeshthummar02:
Issue #18599

https://errorprone.info/bugpattern/ImmutableEnumChecker

`[WARNING] /C:/Users/brije/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/checks/LineSeparatorOption.java:[51,26] [ImmutableEnumChecker] enums should be immutable: 'LineSeparatorOption' has field 'lineSeparator' of type 'byte[]', arrays are mutable
    (see https://errorprone.info/bugpattern/ImmutableEnumChecker)
`

The enum stored a mutable ```byte[]``` field, which violates deep immutability guarantees expected from enums. Even though the field was ```private final```, array contents remain mutable.

This change replaces the stored ```byte[]``` with an immutable ```String``` and performs byte conversion on demand. Behavior remains unchanged, but the enum is now fully immutable and compliant with Error Prone’s ```ImmutableEnumChecker```.

image

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/LineSeparatorOption.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/LineSeparatorOption.java
@@ -48,7 +48,7 @@ public enum LineSeparatorOption {
     SYSTEM(System.lineSeparator());
 
     /** The line separator representation. */
-    private final byte[] lineSeparator;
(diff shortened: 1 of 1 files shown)


romani:
@Brijeshthummar02 , in next PRs please reference link to rule and violations that you addressing

PR #18773, opened 2026-01-27
Issue #18435: Remove HiddenField Example7 from XdocsExamplesAstConsis…

ZaheerAhmadDev:
Issue: #18435

Remove HiddenField Example7 from supression  XdocsExamplesAstConsistencyTest 
Example 1 and Example 7 code were same only configuration were different and violationcomment.
Diffchecker link  of Example 1 vs Example 7 : https://www.diffchecker.com/VMp0OFMs/
Screenshot 2026-01-27 113725
Screenshot 2026-01-27 113643
In Example7 it is showing ` <property name="ignoreAbstractMethods" value="true"/> `
By default it is false so in all Examples it show violation in Example7 it is not.

### DIFF
--- a/src/site/xdoc/checks/coding/hiddenfield.xml
+++ b/src/site/xdoc/checks/coding/hiddenfield.xml
@@ -375,8 +375,7 @@ class Example7 {
   private String field;
   private String testField;
 
-  Example7(int field) { // violation, ''field' hides a field'
-    this.field = Integer.toString(field);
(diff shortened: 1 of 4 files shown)


ZaheerAhmadDev:
@romani,
                   i changed `Example7(int field) to Example7 (String field)` so that it Ast can match with other Examples both constructor are showing same violation. 
Screenshot 2026-01-27 120541

ZaheerAhmadDev:
GitHub, generate website

Issue #18556, opened 2026-01-08
Documentation examples for MultiFileRegexpHeader are not in valid regex form


ascertrobw:
The documentation at the top for this module does state that Java patterns should be used. But as a newcomer, it's confusing that the examples then shown are not in Java regex pattern form.

Update from maintainers:
https://github.com/checkstyle/checkstyle/blob/master/src/site/xdoc/checks/header/multifileregexpheader.xml

We need to use template file for this Check and convert all examples to be like in other Checks - executed by junit.
Example: https://github.com/checkstyle/checkstyle/tree/master/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/header/regexpheader
https://github.com/checkstyle/checkstyle/blob/master/src/xdocs-examples/java/com/puppycrawl/tools/checkstyle/checks/header/RegexpHeaderCheckExamplesTest.java

ZaheerAhmadDev:
i am on it.

Junaid-Ashraf-56:
Hi, I'd like to work on this issue.

romani:
@Junaid-Ashraf-56 , please send PR

Anushreebasics:
@romani please review this PR, pull #18699

Anushreebasics:
hey, @Junaid-Ashraf-56 I am working on this issue

Junaid-Ashraf-56:
@romani  Hi, I began working on this issue and prepared commits, but since another PR is active now, should I switch?  
Could you please suggest another issue or area where I can contribute? I’d be glad to help.

romani:
please switch to other issue.

this issue should have only one PR.

Avanish-Gupta-CSE:
I'm on it - MultiFileRegexpHeader documentation examples

romani:
first migration is done. BUT 
but we need to use not verifyWithInlineConfigParser
but some other method that will use config from separate file.

Inlines config is conflicting with HEADER concept , so we should make separate file with config and load it for execution and from template from that separate file (that will have only config content)

romani:
@ZaheerAhmadDev , can you send second PR to  fix it in better way.

ZaheerAhmadDev:
@romani,

Sure, I will send another PR and will fix it.

romani:
Final fix is merged

romani:
Further improvement will be at https://github.com/checkstyle/checkstyle/issues/18809

PR #18788, opened 2026-01-29
Issue #18556: Adding multifileregexpheader in xdocs-Examples in check

ZaheerAhmadDev:
Issue: #18556
Updated multifileregexpheader and added it in `xdocs-Example`.
Added template file in `site/xdoc/check/header`
Added Junit test for all Examples.  
Now it show java regex pattern.


ZaheerAhmadDev:
GitHub, generate website

romani (approved):
ok to merge as fist step 

but we need to use not verifyWithInlineConfigParser 
but some other method that will use config from separate file.

ZaheerAhmadDev:
@romani,
> ok to merge as fist step
> 
> but we need to use not verifyWithInlineConfigParser but some other method that will use config from separate file.


Instead of using verifyWithInlineConfigParser in Junit test should i use verify method or some other method??.
and for Example2 and Example3  i create external xml in that specific directory of name Example3.xml and Example2.xml in xdocs?. 
Also i have noticed `src/test/resources/com/puppycrawl/tools/checkstyle/checks/header/' there is not multifileregexpheader resources directory in there Should i add it a directory for multifileregexpheader and add all resources of multifileregexpheader inside resources?.
Screenshot 2026-01-30 095138
 or Should i use another  method?
please guide me.
Thank you.

romani:
Please try to make new method verifyxxxx that will allow to load xml from one file , xml file, and run it over java file. So all header Checks will use it.  And users continue to see xml config and Java file in html.

Issue #17137, opened 2025-05-25
UnnecessaryNullCheckWithInstanceOf ignores redundant null check for complex cases.


Anmol202005:
I have read check documentation: https://checkstyle.org/checks/coding/unnecessarynullcheckwithinstanceof.html#UnnecessaryNullCheckWithInstanceOf
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

Follow up of : https://github.com/checkstyle/checkstyle/pull/16489

```bash
/var/tmp $ javac Test.java

/var/tmp $ cat config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <module name="TreeWalker">
    <module name="UnnecessaryNullCheckWithInstanceOf"/>
  </module>
</module>
```
```bash
/var/tmp $ cat Test.java
package com.puppycrawl.tools.checkstyle;

import java.util.Objects;

public class Test {
    public void processData(Object data, String s, String b, String getData) {
        boolean result = false;
        while (getData!= null && getData instanceof String) {
           result = true;
        }
        if (data != null && (!Objects.equals(s, b) || data instanceof String)) {  // false-negative
            boolean isValid = b != null && (s != null || b instanceof String);    //  false-negative
        }

    }
}


/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.24.0-SNAPSHOT-all.jar -c config.xml Test.java
Starting audit...
[ERROR] C:\Users\anmol\IdeaProjects\checkstyle3\src\main\java\com\puppycrawl\tools\checkstyle\Test.java:8:16: Unnecessary nullity check with instanceof operator in conditional expression. [UnnecessaryNullCheckWithInstanceOf]
Audit done.
Checkstyle ends with 1 errors.

```

the null comparison in redundant in the expression `data != null && (!Objects.equals(s, b) || data instanceof String` and `boolean isValid = b != null && (s != null || b instanceof String`  therefore a violation is expected at line 11 and 12.

---

StefanLapointe:
[...]

PR #18792, opened 2026-01-30
Issue #17137: Fix the error for UnnecessaryNullCheckWithInstanceOf for complex cases

MonuChaudhary14:
Issue #17137 

Changes:

- In previous code, it only climb through the LAND, so it incorrectly treated mixed expression as safe
but, i made changes so that now, it climb through LAND and also LOR

- It did not stop when the instanceof was inside a | | subtree

Current flow :

1. It moves to the top level logical expression by climbing through && and | |  and stops at top level
2. If top level logical expression is LOR, then it stops here, it only move further when the top level logical expression is LAND
3. THen , it finds the variable which is used in instanceof 
4. Created a method to prevent derefernce which checks whether there is any dot , method call, because if these are present then we need null check
5. Created a method determines the child the subtree contains the instanceof 

Method :

1. private static DetailAST findTopLevelLogicalExpression(DetailAST node)

- It walks up the AST from the instanceof node and returns the outermost logical expression ( && and | | ) that contains it

The previous code, only traverse through the &&, so i updated to check for | | also

2. private static Optional<DetailAST> findUnnecessaryNullCheck(DetailAST instanceOfNode)

- In the previous version, this method always finds the null check even when the top-level operator is | |

THe current behavior , only proceeds for && and blocks | | 
because instanceof make redundant only under AND 

3. private static Optional<DetailAST> findRedundantNullCheck(DetailAST logicalAndNode,DetailAST instanceOfNode)

- Extracts the identifier from instanceof and also avoid dereference 

4. private static DetailAST searchForNullCheck(DetailAST logicalAndNode,DetailAST instanceOfNode,DetailAST instanceOfIdent)

- Checks where the instaneof is present
- Check if instanceof is present under | |
- stop search when it is unsafe

5. private static DetailAST checkChildForNullCheck(DetailAST currentChild,DetailAST instanceOfNode, DetailAST instanceOfIdent, DetailAST currentNullCheck)

- It examines the child and check whether it is redundant null check or nested 

6. private static DetailAST findDirectChildContaining(DetailAST parent,DetailAST target)

- It returns which  immediate child of LAND node contain the instanceof

7. private static boolean isAncestorOf(DetailAST node, DetailAST target)

- Checks whether a node contains another node anywhere in its subtree

8. private static boolean containsVariableDereference(DetailAST node, String variableName)

- if (node.getType() == DOT || METHOD_CALL || LAND || LOR) Added LOR in this
this ensures that dereferance are detected even inside the OR brances

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryNullCheckWithInstanceOfCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/UnnecessaryNullCheckWithInstanceOfCheck.java
@@ -74,13 +74,29 @@ public class UnnecessaryNullCheckWithInstanceOfCheck extends AbstractCheck {
      * @return the identifier if the check is redundant, otherwise {@code null}
      */
     private static Optional<DetailAST> findUnnecessaryNullCheck(DetailAST instanceOfNode) {
-        DetailAST currentParent = instanceOfNode;
+        final DetailAST topLevelExpr = findTopLevelLogicalExpression(instanceOfNode);
 
-        while (currentParent.getParent().getType() == TokenTypes.LAND) {
+        Optional<DetailAST> result = Optional.empty();
+        if (topLevelExpr.getType() == TokenTypes.LAND) {
+            result = findRedundantNullCheck(topLevelExpr, instanceOfNode)
+                    .map(DetailAST::getFirstChild);
+        }
+        return result;
+    }
+
+    /**
+     * Traverses up through LAND and LOR operators to find the top-level logical expression.
+     *
+     * @param node the starting node
+     * @return the top-level logical expression node
+     */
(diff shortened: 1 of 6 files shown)


MonuChaudhary14:
@romani 
Can you review it?

MonuChaudhary14:
Github, generate report for UnnecessaryNullCheckWithInstanceOf/all-examples-in-one

MonuChaudhary14:
@romani 
I have verified all the violations are genuine.

MonuChaudhary14:
@romani 

Is there any changes needed?

MonuChaudhary14:
@romani 

Please review this PR and tell me if any changes are needed

romani (approved):
all done in good way !!

thanks a lot for doing all in single attempt, such a rare case when I can merge  PR without bunch of to and from comments

Issue #10969, opened 2021-11-21
RequireThisCheck: pattern variables produce false positives


rnveach:
Identified at https://github.com/checkstyle/checkstyle/issues/7504#issuecomment-974776332 ,

````
$ cat TestClass.java
public class Test {
String p, s, n;

public static void main(String... arguments) {
new Test().run(arguments);
}

public void run(String... arguments) {
this.p = null;
this.s = null;
this.n = null;

Object o = 45;
if (!(o instanceof String p && p.equals("sd")) || !p.equals("wq")) {
p = "a"; // instance var p
} else if (!(o instanceof Integer s && o instanceof String n)) {
p = "b"; // local var p
s = "b"; // instance var s
n = "b"; // instance var n
} else {
p = "c"; // local var p
s = 41; // local var s
n = "c"; // local var n
}


System.out.println(this.p + ":" + this.s + ":" + this.n);
}
}


$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
<module name="RequireThis">
  <property name="validateOnlyOverlapping" value="false"/>
</module>
    </module>
</module>

$ java -jar checkstyle-9.0.1-all.jar -c TestConfig.xml TestClass.java
Starting audit...
[ERROR] TestClass.java:15:43: Reference to instance variable 'p' needs "this.". [RequireThis]
[ERROR] TestClass.java:15:48: Reference to instance variable 'p' needs "this.". [RequireThis]
[ERROR] TestClass.java:15:68: Reference to instance variable 'p' needs "this.". [RequireThis]
[ERROR] TestClass.java:16:25: Reference to instance variable 'p' needs "this.". [RequireThis]
[ERROR] TestClass.java:17:51: Reference to instance variable 's' needs "this.". [RequireThis]
[ERROR] TestClass.java:17:76: Reference to instance variable 'n' needs "this.". [RequireThis]
[ERROR] TestClass.java:18:25: Reference to instance variable 'p' needs "this.". [RequireThis]
[ERROR] TestClass.java:19:25: Reference to instance variable 's' needs "this.". [RequireThis]
[...]

PR #18813, opened 2026-02-01
Issue #10969: Fix false positives for pattern variables in RequireThisCheck

Shubhamyadav93:
Hi

I've updated the RequireThisCheck logic to handle pattern variables correctly. Previously, the check was failing to recognize pattern variables as local declarations, causing false positive violations for the this. qualifier even when a pattern variable was shadowing a field.

I have added PATTERN_VARIABLE_DEF to the identified local variables list. I also added a regression test in RequireThisCheckTest.java to confirm the fix.

Issue: #10969


romani:
Please make CI green.

PR #18830, opened 2026-02-02
Issue #17882: Update SLASH to new AST format

VidytBhudolia:
Issue #17882

https://docs.oracle.com/en/java/javase/17/docs/specs/javadoc/doc-comment-spec.html#see

>`@see module/package.class#member label`
Adds a link with a visible text label that points to the documentation for the specified name that is referenced. The label is optional; if it is omitted, then the program element name appears instead as visible text, suitably shortened. Use the -noqualifier option to globally remove the package name from this visible text. Use the label when you want the visible text to be different from the auto-generated visible text.
module/package.class#member is any valid program element name that is referenced, such as a module, package, class, interface, constructor, method or field name. Parts of the name can be omitted as appropriate. The class represents any top-level or nested class or interface. The member represents any constructor, method, or field (not a nested class or interface). Parameterized types may be used in the class and member parts of the name. If this name is in the documented classes, then the standard doclet creates a link to it. A trailing / can be added to a name to refer to a module in the presence of a package or class with the same name.

**Command**
`java -jar checkstyle-13.1.0-all.jar -j src/Test.java | sed "s/\[[0-9]\+:[0-9]\+\]//g"`

**Test.java**
```
/**
 * {@link java.base/java.lang.String}
 */
class Test {}
```

**Output**
```
$ java -jar checkstyle-13.1.0-all.jar -j src/Test.java | sed "s/\[[0-9]\+:[0-9]\+\]//g"
JAVADOC_CONTENT -> JAVADOC_CONTENT
|--TEXT -> /**
|--NEWLINE -> \r\n
|--LEADING_ASTERISK ->  *
|--TEXT ->
|--JAVADOC_INLINE_TAG -> JAVADOC_INLINE_TAG
|   `--LINK_INLINE_TAG -> LINK_INLINE_TAG
|       |--JAVADOC_INLINE_TAG_START -> {@
|       |--TAG_NAME -> link
|       |--TEXT ->
|       |--REFERENCE -> REFERENCE
|       |   |--IDENTIFIER -> java.base
|       |   |--SLASH -> /
|       |   `--IDENTIFIER -> java.lang.String
|       `--JAVADOC_INLINE_TAG_END -> }
|--NEWLINE -> \r\n
|--LEADING_ASTERISK ->  *
|--TEXT -> /
|--NEWLINE -> \r\n
|--TEXT -> class Test {}
`--NEWLINE -> \r\n
```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
@@ -1216,7 +1216,28 @@ public final class JavadocCommentsTokenTypes {
     public static final int COMMA = JavadocCommentsLexer.COMMA;
 
     /**
-     * Slash symbol {@code / }.
+     * Slash symbol {@code /} used in module or package references within Javadoc.
+     *
+     * <p><b>Example:</b></p>
+     * <pre>{@code
+     * &#123;@link java.base/java.lang.String&#125;
+     * }</pre>
+     *
+     * <b>Tree:</b>
+     * <pre>{@code
+     * |--JAVADOC_INLINE_TAG -> JAVADOC_INLINE_TAG
+     *     `--LINK_INLINE_TAG -> LINK_INLINE_TAG
+     *         |--JAVADOC_INLINE_TAG_START -> &#123;@
+     *         |--TAG_NAME -> link
+     *         |--TEXT ->
+     *         |--REFERENCE -> REFERENCE
+     *         |   |--IDENTIFIER -> java.base
+     *         |   |--SLASH -> /
+     *         |   `--IDENTIFIER -> java.lang.String
+     *         `--JAVADOC_INLINE_TAG_END -> }
+     * }</pre>
+     *
+     * @see #REFERENCE
      */
     public static final int SLASH = JavadocCommentsLexer.SLASH;
 


romani:
SLASH should be in tree

romani:
Please show it by CLI output

VidytBhudolia:
> Please show it by CLI output

Screenshot 2026-02-04 031135

CLI output shows that / is tokenized as TEXT, not as a standalone SLASH node.
I’ve updated the Tree section to exactly match the AST produced above.

romani:
>CLI output shows that / is tokenized as TEXT, not as a standalone SLASH node.

It means it not good example, please investigate how to make example that will have SLASH in tree

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
```diff
@@ -1268,42 +1216,33 @@ public final class JavadocCommentsTokenTypes {
...
-     *         |       |   |--QUESTION -> ?
-     *         |       |   |--TEXT ->
-     *         |       |   |--EXTENDS -> extends
-     *         |       |   |--TEXT ->
-     *         |       |   `--IDENTIFIER -> Number
-     *         |       `--GT -> >
+     *             |--IDENTIFIER -> java.base
+     *             |--SLASH -> /
+     *             `--IDENTIFIER -> java.lang.String
      *         `--JAVADOC_INLINE_TAG_END -> }
```
romani, line 1237: Tree is:
```
|--JAVADOC_INLINE_TAG -> JAVADOC_INLINE_TAG
|   `--LINK_INLINE_TAG -> LINK_INLINE_TAG
|       |--JAVADOC_INLINE_TAG_START -> {@
|       |--TAG_NAME -> link
|       |--TEXT ->
|       |--REFERENCE -> REFERENCE
|       |   |--IDENTIFIER -> java.base
|       |   |--SLASH -> /
|       |   `--IDENTIFIER -> java.lang.String
|       `--JAVADOC_INLINE_TAG_END -> }
```

And only most left | should be removed as it goes to nowhere

romani (changes requested):
Items

VidytBhudolia:
Thank you very much! Much appreciated.
I’ll keep that in mind going forward.

Issue #18836, opened 2026-02-03
Enable pitest EXPERIMENTAL_MEMBER_VARIABLE


romani:
https://github.com/hcoles/pitest/issues/1440 is fixed 

We need to revert this suppression 
https://github.com/checkstyle/checkstyle/pull/18443/changes#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L5008

If issue fixes fully, no other changes will be required.
If there some mutation CI find, we should report it to pitest project 

------

Scope becomes a bit bigger, we have 4 more case with such suppression in pom.xml:
```
<!-- Disabled until issue in pitest is resolved at https://github.com/hcoles/pitest/issues/1440 -->
<!-- <mutator>EXPERIMENTAL_MEMBER_VARIABLE</mutator> -->
                
```

We need to uncomment this mutator.

ayushactiveat:
I am working on this.

romani:
This is good examples on how update should be done 
https://github.com/checkstyle/checkstyle/pull/18837

But pretty sure in most cases it will be just uncommenting

romani:
All fixes are done by @ayushactiveat.

PR #18849, opened 2026-02-04
Issue #18836: Enable pitest EXPERIMENTAL_MEMBER_VARIABLE and exclude …

Youssefalaa7:
Issue #18836: Enable pitest EXPERIMENTAL_MEMBER_VARIABLE for all rest pitests and exclude optimization mutation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5432,8 +5432,7 @@
                 <mutator>EXPERIMENTAL_ARGUMENT_PROPAGATION</mutator>
                 <mutator>EXPERIMENTAL_BIG_DECIMAL</mutator>
                 <mutator>EXPERIMENTAL_BIG_INTEGER</mutator>
-                <!-- Disabled until issue in pitest is resolved at https://github.com/hcoles/pitest/issues/1440 -->
-                <!-- <mutator>EXPERIMENTAL_MEMBER_VARIABLE</mutator> -->
+                <mutator>EXPERIMENTAL_MEMBER_VARIABLE</mutator>
                 <mutator>EXPERIMENTAL_NAKED_RECEIVER</mutator>
                 <mutator>REMOVE_INCREMENTS</mutator>
                 <mutator>EXPERIMENTAL_SWITCH</mutator>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -5122,8 +5121,7 @@
                 <mutator>EXPERIMENTAL_ARGUMENT_PROPAGATION</mutator>
                 <mutator>EXPERIMENTAL_BIG_DECIMAL</mutator>
                 <mutator>EXPERIMENTAL_BIG_INTEGER</mutator>
-                <!-- Disabled until issue in pitest is resolved at https://github.com/hcoles/pitest/issues/1440 -->
-                <!-- <mutator>EXPERIMENTAL_MEMBER_VARIABLE</mutator> -->
+                <mutator>EXPERIMENTAL_MEMBER_VARIABLE</mutator>
```
romani, line 5124: This two looks like simple activation, that doesn't need code changes.
If yes, please move to separate PR to merge quickly

pom.xml
```diff
@@ -5457,6 +5454,11 @@
                 <param>com.puppycrawl.tools.checkstyle.XpathFileGeneratorAstFilterTest</param>
                 <param>com.puppycrawl.tools.checkstyle.filters.*</param>
               </targetTests>
+              <excludedMethods>
+                <!-- performance optimization -->
+                <param>checkResultForFollowing</param>
+                <param>checkResultForPreceding</param>
+              </excludedMethods>
               <avoidCallsTo>
```
romani, line 5457: This stuff will require more time, as it hack.
If no other way I would rather keep this mutator permanently disabled

pom.xml
Youssefalaa7: done i created a separate pr to focus here only on xpath pitest

pom.xml
Youssefalaa7: i will try to find another way to kill this mutation if not possible i will permanently disable this mutation for the xpath

Youssefalaa7:
@romani  after multiple tests and understanding how this parsing works in detail i found that this optimization is not needed at all and it do not differ in anything in the case of number of loops in do while the only difference that this assignment was making is that when reaching that iterator.next returns null and ending the parsing of the whole tree and then calling iterator.next after the tree was already finished it do not enter the second if condition in the next method which is responsible for the sibling 
so this part is not important at all

PR #18854, opened 2026-02-04
Issue #18599: Disable InlineMeSuggester as Error Prone annotations are forbidden in main sources

Brijeshthummar02:
Issue #18599

## Disable InlineMeSuggester due to Checkstyle source restrictions

This PR disables the Error Prone `InlineMeSuggester` rule.

```
Warning:  /home/runner/work/checkstyle/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/
DefaultConfiguration.java:[144,17] [InlineMeSuggester] This deprecated API looks inlineable. If you'd like the body of the API to be 
automatically inlined to its callers, please annotate it with @InlineMe. NOTE: the suggested fix makes the 
method final if it was not already.
    (see https://errorprone.info/bugpattern/InlineMeSuggester)
```

https://github.com/checkstyle/checkstyle/blob/4629a7a4007cb9558fe9cdf6e5d4e8ae42dbd4b1/src/main/java/com/puppycrawl/tools/checkstyle/DefaultConfiguration.java#L143-L146

While investigating the warning reported by Error Prone:

https://errorprone.info/bugpattern/InlineMeSuggester
https://errorprone.info/docs/inlineme

I attempted to apply the suggested fix by annotating the deprecated `addAttribute` method in `DefaultConfiguration` with `@InlineMe`. However, this change introduced new Checkstyle violations:

- `ImportControlMain` rejects `com.google.errorprone.annotations.InlineMe` in main sources
- `RedundantModifier` flags the required `final` modifier since `DefaultConfiguration` is already final

As a result, the original implementation of `addAttribute` has been restored, and `InlineMeSuggester` is explicitly turned OFF in the Error Prone configuration.

### DIFF
--- a/config/checkstyle-checks.xml
+++ b/config/checkstyle-checks.xml
@@ -299,7 +299,7 @@
     <module name="MissingOverrideOnRecordAccessor"/>
     <module name="PackageAnnotation"/>
     <module name="SuppressWarnings">
-      <property name="format" value="^((?!unchecked|deprecation|rawtypes|resource|InvalidInlineTag|UnrecognisedJavadocTag).)*$"/>
+      <property name="format" value="^((?!unchecked|deprecation|rawtypes|resource|InvalidInlineTag|UnrecognisedJavadocTag|InlineMeSuggester).)*$"/>
       <message key="suppressed.warning.not.allowed"
            value="The warning ''{0}'' cannot be suppressed at this location.
            Only few javac warnings are allowed to suppress.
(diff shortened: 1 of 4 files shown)


romani:
This method is doomed to be deleted, we should not reduce quality threshold just to let this item to not be under violation.

We don't need inlineme , we just need to suppress violation.
Look at this commit on how we did this in past https://github.com/checkstyle/checkstyle/commit/5ca8b1797d5f4e20d51134013130c11365c2c1de

And make error-prone to be Error level, to prevent new problems.

romani (approved):
Thanks a lot

Issue #18856, opened 2026-02-04
Split openrewrite-refaster-rules ci job into two jobs


romani:
https://app.circleci.com/pipelines/github/checkstyle/checkstyle/41538/workflows/d00b5d3f-6794-4061-aa46-4083806bdfde/jobs/1309693

<img width="890" height="403" alt="Image" src="https://github.com/user-attachments/assets/710ca14d-a791-4c42-acdd-31a4ad6f4edd" />

But there were a lot  examples that this tool takes too much time to finish.

For now we presume that it simply needs more time.
So we can split this job into two parts to let finish on time, 15 minutes max

romani:
Fix is merged

Pankraz76:
> https://app.circleci.com/pipelines/github/checkstyle/checkstyle/41538/workflows/d00b5d3f-6794-4061-aa46-4083806bdfde/jobs/1309693
> 
[...]

PR #18859, opened 2026-02-05
Issue #18856: Split openrewrite-refaster-rules ci job into two jobs

Youssefalaa7:
Issue #18856: Split openrewrite-refaster-rules ci job into two jobs


romani (approved):
Thanks a lot 
5 min
And 8 min

PR #18863, opened 2026-02-05
Issue #18435: Fix AST inconsistency in OneTopLevelClass examples

### LINKED ISSUE #18435: Fix xdocs Examples AST Consistency Test (Reduce suppressions list)

ishasingh1505:
Example1 and Example2 both demonstrate the OneTopLevelClass violation but had different ASTs due to a missing `public` modifier in Example2.
This change makes Example2 AST-identical to Example1 and removes it from the suppression list. Example3 remains suppressed as it represents a different, non-violation use case.
Fixes #18435

### DIFF
--- a/src/site/xdoc/checks/design/onetoplevelclass.xml
+++ b/src/site/xdoc/checks/design/onetoplevelclass.xml
@@ -50,7 +50,7 @@ class ViolationExample1 { // violation, "has to reside in its own source file."
           An example of code without public top-level type:
         </p>
         <div class="wrapper"><pre class="prettyprint"><code class="language-java">
-class Example2 { // ok, first top-level class
+public class Example2 { // ok, first top-level class
   // methods
 }
 
--- a/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
@@ -135,7 +135,6 @@ public class XdocsExamplesAstConsistencyTest {
             "checks/descendanttoken/Example7",
             "checks/descendanttoken/Example8",
             "checks/descendanttoken/Example9",
-            "checks/design/onetoplevelclass/Example2",
             "checks/design/onetoplevelclass/Example3",
             "checks/design/visibilitymodifier/Example11",
             "checks/design/visibilitymodifier/Example12",
--- a/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/design/onetoplevelclass/Example2.java
+++ b/src/xdocs-examples/resources/com/puppycrawl/tools/checkstyle/checks/design/onetoplevelclass/Example2.java
@@ -9,7 +9,7 @@
 package com.puppycrawl.tools.checkstyle.checks.design.onetoplevelclass;
 
 // xdoc section -- start
-class Example2 { // ok, first top-level class
+public class Example2 { // ok, first top-level class
   // methods
 }
 


ZaheerAhmadDev:
Please make CI green.

ZaheerAhmadDev:
Please make only one commit.
Please provide [diffchecker](https://www.diffchecker.com/) link of example also.

ishasingh1505:
Squashed the PR. Diffchecker link confirming Example1 and Example2 are AST-identical (only class names differ): https://www.diffchecker.com/sUJr4pYU/

ZaheerAhmadDev:
Screenshot 2026-02-05 200459

Example 1 vs Example 2 are both same by violation comment,confiq and code.
Example 1 represent class with `public` and Example2 represent class  with `private`.

ishasingh1505:
@ZaheerAhmadDev  The issue is not whether both examples violate OneTopLevelClass, but that they were not AST-equivalent. XdocsExamplesAstConsistencyTest requires identical ASTs for examples representing the same case, and public vs package-private classes produce different ASTs. This change fixes that inconsistency and correctly removes Example2 from suppression.

romani:
GitHub, generate website

src/site/xdoc/checks/design/onetoplevelclass.xml
```diff
@@ -50,7 +50,7 @@ class ViolationExample1 { // violation, "has to reside in its own source file."
           An example of code without public top-level type:
         </p>
```
romani, line 51: We can not change this example, as it has no public intentionally.

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -135,7 +135,6 @@ public class XdocsExamplesAstConsistencyTest {
             "checks/descendanttoken/Example7",
             "checks/descendanttoken/Example8",
             "checks/descendanttoken/Example9",
-            "checks/design/onetoplevelclass/Example2",
             "checks/design/onetoplevelclass/Example3",
```
romani, line 138: Both should go under `// No properties in module,`  at the bottom of a list.
All No property Checks should be out of this validation,

romani (changes requested):
Items

romani:
@ishasingh1505, can run through whole list and move to bottom all modules that has no properties. To let us not going back to this any more

PR #18865, opened 2026-02-05
Issue #273: Small cleanup in diff-report workflow to prepare for CLI migration

Brijeshthummar02:
These are intentionally small changes to simplify the current Groovy-based flow before migrating to direct Checkstyle CLI execution in follow-up PRs.

- Replace MAVEN_OPTS with JAVA_OPTS
- Remove Maven-specific flags

No functional behavior is changed yet.

Issue https://github.com/checkstyle/contribution/issues/273

### DIFF
--- a/.github/workflows/diff-report.yml
+++ b/.github/workflows/diff-report.yml
@@ -172,18 +172,18 @@ jobs:
           REF="forked/$BRANCH"
           REPO="../../../../checkstyle"
           BASE_BRANCH="upstream/master"
-          export MAVEN_OPTS="-Xmx5g"
+          export JAVA_OPTS="-Xmx5g"
           if [ -f new_module_config.xml ]; then
             groovy diff.groovy -r "$REPO" -p "$REF" -pc new_module_config.xml -m single\
-              -l projects.properties -xm "-Dcheckstyle.failsOnError=false"\
+              -l projects.properties \
               --allowExcludes
           elif [ -f patch_config.xml ]; then
             groovy diff.groovy -r "$REPO" -b "$BASE_BRANCH" -p "$REF" -bc diff_config.xml\
-              -pc patch_config.xml -l projects.properties -xm "-Dcheckstyle.failsOnError=false"\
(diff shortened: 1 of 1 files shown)


romani:
>Replace MAVEN_OPTS with JAVA_OPTS

Why you think it will work the same.?

>Remove Maven-specific flags

What was a goal of it?

Brijeshthummar02:
> > Replace MAVEN_OPTS with JAVA_OPTS
> 
> Why you think it will work the same.?
> 

- `JAVA_OPTS` keeps the same `-Xmx5g` JVM setting, but removes the Maven-specific variable since this step does not directly run Maven.

> > Remove Maven-specific flags
> 
> What was a goal of it?

-  The Maven-specific flags were removed to reduce Maven assumptions in the workflow and make the upcoming switch to Checkstyle CLI easier. The actual CLI migration and Groovy removal will be done in follow-up PRs.

.github/workflows/diff-report.yml
```diff
@@ -172,18 +172,18 @@ jobs:
           REF="forked/$BRANCH"
           REPO="../../../../checkstyle"
           BASE_BRANCH="upstream/master"
-          export MAVEN_OPTS="-Xmx5g"
+          export JAVA_OPTS="-Xmx5g"
           if [ -f new_module_config.xml ]; then
             groovy diff.groovy -r "$REPO" -p "$REF" -pc new_module_config.xml -m single\
-              -l projects.properties -xm "-Dcheckstyle.failsOnError=false"\
+              -l projects.properties \
```
romani, line 178: https://github.com/checkstyle/contribution/blob/cbe2074f1e530317f0e32ae5c3d6c891f275e3d0/checkstyle-tester/diff.groovy#L460

` groovy diff.groovy ` heavily depend on maven , we should do this update at the very end, and think on how to migrate  diff.groovy first.

romani (changes requested):
items:

romani:
let close this PR for now, as we are so far from migration  and  diff.groovy should be updated first.

PR #18866, opened 2026-02-05
Issue #14084: Resolve CheckerFramework violations

Gaoyan1999:
Issue:
#14084

In that case, @PolyNull means the nullness of the return value is the same as the nullness of Pattern.
https://checkerframework.org/manual/#qualifier-polymorphism-examples

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
@@ -22,6 +22,8 @@ package com.puppycrawl.tools.checkstyle.filters;
 import java.util.Objects;
 import java.util.regex.Pattern;
 
+import javax.annotation.Nullable;
+
 import com.puppycrawl.tools.checkstyle.api.AuditEvent;
 import com.puppycrawl.tools.checkstyle.api.Filter;
 
@@ -210,6 +212,7 @@ public class SuppressFilterElement
      * @param pattern pattern object
      * @return value of pattern or null
      */
+    @Nullable
     private static String getPatternSafely(Pattern pattern) {
         String result = null;
         if (pattern != null) {
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -6469,7 +6469,6 @@
     </details>
   </checkerFrameworkError>
 
-
   <checkerFrameworkError unstable="false">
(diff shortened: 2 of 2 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElement.java
```diff
@@ -210,11 +211,7 @@ public boolean equals(Object other) {
...
      * @return value of pattern or null
      */
-    private static String getPatternSafely(Pattern pattern) {
-        String result = null;
-        if (pattern != null) {
-            result = pattern.pattern();
-        }
-        return result;
+    private static @PolyNull String getPatternSafely(@PolyNull Pattern pattern) {
+        return pattern == null ? null : pattern.pattern();
```
Gaoyan1999: Hi @romani , When using @PolyNull from the Checker Framework, the Nullness Checker infers the method’s return nullness from the conditional. For example:
```
return pattern == null ? null : pattern.pattern();
```

or
```
if (pattern == null) {
    return null;
}
return pattern.pattern();
```

However, both forms are rejected by the style check rules.

In this situation, would it be acceptable to suppress the style check for this method, or should we give up using @PolyNull here?

romani: Main problem is `[ERROR] [checkstyle] [ERROR] C:\projects\checkstyle\src\main\java\com\puppycrawl\tools\checkstyle\filters\SuppressFilterElement.java:27:1: Disallowed import - org.checkerframework.checker.nullness.qual.PolyNull. [ImportControlMain]`

We don't want to depend on checkerframework in maven dependencies.
Unfortunately all annotations will require big dependency introduction, that might be a problem for our users. We already had bunch of problems like this for other libraries that we used

romani: In NullUnil , we did a hack of copying snippets from checkerframework to let us avoid dependency and in same , let checkerframework do a job.

romani: We need to think on getPatternSafely and maybe be reimplement it to avoid dealing with null.
Checker point to null players, but it doesn't mean null should stay, it might be code that was created 20 years ago , there was no other options. But now java has a lot of options.

Gaoyan1999: Make it simple. We can annotate the return value as nullable, since there’s no property access at the call sites.

PR #18878, opened 2026-02-06
Revert "Issue #18856: Split openrewrite-refaster-rules ci job into two jobs"

Pankraz76:
### Issue #18856: Split openrewrite-refaster-rules ci job into two jobs

Revert "Issue #18856: Split openrewrite-refaster-rules ci job into two jobs"

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ jobs:
         type: string
       no_output_timeout:
         type: string
-        default: 10m
+        default: 15m
     docker:
       - image: << parameters.image-name >>
     steps:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         type: string
       no_output_timeout:
         type: string
-        default: 10m
```
Pankraz76, line 43: this the only thing we need. 15 mins is our deadline so 10 is just some others random default, check is different so the best practise should be aligned to avoid any further....

other (rewrite) jobs will follow and the DRY violations point this out as well:

Image

.circleci/config.yml
```diff
@@ -319,15 +319,9 @@ workflows:
           image-name: "cimg/openjdk:21.0"
           command: "./.ci/validation.sh openrewrite-checkstyle-auto-fix"
       - validate-with-maven-script:
-          name: "openrewrite-refaster-rules-1"
+          name: "openrewrite-refaster-rules"
           image-name: "cimg/openjdk:21.0"
-          command: "./.ci/validation.sh openrewrite-refaster-rules-1"
```
Pankraz76, line 324: This is excessive for the current state. We shouldn’t always go the strict checkstyle route right away, but instead try the standard approach first — maximizing our current potential without adding unnecessary overhead or overcomplicating things.

Pankraz76:
Duration / Finished
11m 29s

romani:
Amount or receipt is growing, split is not avoidable

Pankraz76:
> Amount or receipt is growing

no body is adding some recipes as no one seems to care about except of me and i dont give a fuck anymore so i dont see how this is growing.

PR #18881, opened 2026-02-06
Issue #11166: Fix existing cases of getFileContents() usage in Javado…

Youssefalaa7:
Issue #11166: Fix existing cases of getFileContents() usage in JavadocMethodCheck
removed the calling of getFileContents and parsed the javadoc comment and changed it to text
and then converted the javadoc comment AST node to a TextBlock


Youssefalaa7:

Youssefalaa7:
@romani the differences found should be 0 right? so hear it means i made wrong logic?

src/test/java/com/puppycrawl/tools/checkstyle/api/FileContentsTest.java
```diff
@@ -358,4 +358,37 @@ public void testUnmodifiableGetBlockComments() {
...
+    @Test
+    public void testGetJavadocBeforeWithBlockCommentOnSameLine() {
+        final FileContents fileContents = new FileContents(
+                new FileText(new File("filename"), Arrays.asList(
+                        "/** Javadoc */",
+                        "   /* block comment */ code here",
+                        "   public void method() {}")));
+        fileContents.reportBlockComment(1, 0, 1, 13);
+        fileContents.reportBlockComment(2, 3, 2, 20);
+        final TextBlock javadoc = fileContents.getJavadocBefore(3);
```
romani, line 371: All tests should be by verifyxXxxxx method usage.
No pure unit testing.

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/api/FileContentsTest.java
Youssefalaa7: All tests in this file are unit tests there are no tests using verifyxXxxxx because thiss is testing the method getJavadocBefore that we removed its usage from  JavadocMethodCheck.java

romani: All should be covered by testing of Checks. Pure unit testing was huge mistake 20 years ago, unfortunately we didn't kill it completely, it will take extra years to remove bad examples.

romani: Ohh, a second.
We remove usage of FileContents from Check.
That is why coverage is failing and we stay compliant on coverage we have to do pure testing, until migration from FileContents is finished and we can not remove tests.

romani:
There should be no diff, or diff that is proven better behavior.

Now check do less violations, that is bad - https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/c72df05443f58da2cda5c38bedc0e65737d7b585_2026025628/reports/diff/apache-jsecurity/index.html#A1 please fix.

Please review all other changes in diff report to be sure it is not regression.

Keep extending our Inputs with all edgecases, we should less and less rely on diff report to catch regression

romani:
Please add generic parameter to examples, to show how it works  to user, I already forgot that it is required to be documented.

Youssefalaa7:
@romani  I added generic examples and now will work on minimizing the diff appearing in diff test

romani:
GitHub, generate website

Youssefalaa7:

Youssefalaa7:
@romani the diff decreased more than before and the one you showed above now has no changes i will continue to decrease difference more

Youssefalaa7:

Youssefalaa7:

Youssefalaa7:

Youssefalaa7:
@romani 
The violations that were appearing in the infinispan diff report for cases like:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/1ea51b94505996f7aeb5582398ebb135fd16cbc6_2026204756/reports/diff/infinispan/index.html
```
/** Interface describing a function mapping two doubles to a double */
public interface DoubleByDoubleToDouble { double apply(double a, double b); }
```
were false positives from JavadocMethodCheck. The method apply has no Javadoc at all, so JavadocMethodCheck should never have been firing here in the first place — you can't have missing @param tags in a Javadoc that doesn't exist.
The correct check to handle this case is MissingJavadocMethodCheck, which is responsible for detecting when a method has no Javadoc comment at all. JavadocMethodCheck should only fire when a Javadoc comment exists but is incomplete (missing @param, @return, @throws tags).

what do you think? please review this pr and the diff report when you have time because i think most differences are good ones

Youssefalaa7:

romani:
>JavadocMethodCheck should only fire when a Javadoc comment exists but is incomplete

Yes.
Just be sure that there is no bad violation - false positives.

New violations that are valid, is completely ok to keep.
Just make sure all such cases are added to our Inputs.

Youssefalaa7:

Youssefalaa7:

Youssefalaa7:

Youssefalaa7:

Youssefalaa7:

Youssefalaa7:

Youssefalaa7:
@romani this PR is ready for review there are no new violations now only removed the false positives that were appearing before and should not appear

Youssefalaa7:

romani:
Can we restore throws validation https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/74e4285ad3cb0a953111ccfa387fe93d0f21869b_2026171911/reports/diff/pulsar/index.html#A1 ?

Extra param in javadoc https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/74e4285ad3cb0a953111ccfa387fe93d0f21869b_2026171911/reports/diff/openjdk25/index.html#A5

Please add such code to input.

Youssefalaa7:
@romani the first one the throw violation is not appearing because because above the method there is @test then javadoc and above them group of comments because the method above it is all commented only the annotation not which is @SuppressWarnings("deprecation") so the javadoc in the view of ast is between two annotation and this is wrong place for javadoc so it is ignored and no violation should happen

Youssefalaa7:
and for the extra param in javadoc this also should not give violation because all this javadocs are for the classes and javadocMethodCheck is not responsible for validating javadocs for classes JavadocTypeCheck is the one responsible for this

add these two cases to our inputs and show that they should not give violation?

romani:
```
176     /**
177      * Verify indexing with multiple ensemble reformation.
178      */
179     @Test
180     public void testEnsembleReformation() throws Exception {

```

This is javadoc , all above it should not be considered at all.
This javadoc should have throws explanation.

If we are affected by stuff above, We need to improve logic to not be affected.

romani:
>add these two cases to our inputs and show that they should not give violation?

Yes . 
It is good to have case that inner classes are not validated.
Please add more cases that showing up in diff report.

Youssefalaa7:
good i will add those

Youssefalaa7:
@romani the correct here is to not throw violation because he did not comment line 151 so this annotation became linked to the second method below because the annotation can not become attached with no method it will give error so it was supposed to also comment line 151 as he commented the whole method and he also commented line 152 so now the  javadoc at line 176 appear as if it is between two annotations because the line 151 and line 179 are two annotations for method at line 180 so this is not the correct place for a javadoc it be before all the annotations so the javadoc validation is ignored and nothing is throwed and this is the correct behaviour and if we try to fix something like this all cases that have javadoc between annotations will become validated and this should not be correct 

```
148     /**
149      * Verify ledger index with failed bookies and throws exception.
150      */
151     @SuppressWarnings("deprecation")
152 //    @Test
153 //    public void testWithoutZookeeper() throws Exception {
154 //        // This test case is for ledger metadata that stored in ZooKeeper. As
155 //        // far as MSLedgerManagerFactory, ledger metadata are stored in other
156 //        // storage. So this test is not suitable for MSLedgerManagerFactory.
157 //        if (newLedgerManagerFactory instanceof org.apache.bookkeeper.meta.MSLedgerManagerFactory) {
158 //            return;
159 //        }
160 //
161 //        for (int i = 0; i < numberOfLedgers; i++) {
162 //            createAndAddEntriesToLedger().close();
163 //        }
164 //
165 //        BookieLedgerIndexer bookieLedgerIndex = new BookieLedgerIndexer(
166 //                ledgerManager);
167 //        stopZKCluster();
168 //        try {
169 //            bookieLedgerIndex.getBookieToLedgerIndex();
170 //            fail("Must throw exception as zookeeper are not running!");
171 //        } catch (BKAuditException bkAuditException) {
172 //            // expected behaviour
173 //        }
174 //    }
175 
176     /**
177      * Verify indexing with multiple ensemble reformation.
178      */
179     @Test
180     public void testEnsembleReformation() throws Exception {
```

romani:
https://github.com/checkstyle/checkstyle/actions/runs/22147987860/job/64030726160#step:5:75

As you can see even in tree javadoc comment is most close to method.
All other comments should not be considered at all.

We can take one more opinion on this edge case.
Let's  run javadoc tool on such snippet to see what comment will be selected by tool to be printed in html page. Please share java code and rendered html page

Youssefalaa7:
@romani i added tests and inputs for classes and showed they should not give violation and will look for this edgecase

Youssefalaa7:
@romani

 javadoc tested

```
public class Test {
    /**
     * Verify ledger index with failed bookies and throws exception.
     */
    @SuppressWarnings("deprecation")
//    @Test
//    public void testWithoutZookeeper() throws Exception {
//        // commented code
//    }

    /**
     * Verify indexing with multiple ensemble reformation.
     */
    public void testEnsembleReformation() throws Exception {
        // method body
    }
}
```

generated HTML

```
<section class="detail" id="testEnsembleReformation()">
<h3>testEnsembleReformation</h3>
<div class="member-signature"><span class="modifiers">public</span>&nbsp;<span class="return-type">void</span>&nbsp;<span class="element-name">testEnsembleReformation</span>()
                             throws <span class="exceptions">Exception</span></div>
<div class="block">Verify ledger index with failed bookies and throws exception.</div>
<dl class="notes">
<dt>Throws:</dt>
<dd><code>Exception</code></dd>
</dl>
</section>
```

it took the javadoc before the first annotation and ignored the javadoc between the annotations

Youssefalaa7:
@romani failure unrelated

Youssefalaa7:
@romani ping

romani:

Tasneemmohammed0:
@romani @Youssefalaa7,  I believe this PR will help address the remaining Javadoc checks in #11166. It would be best to prioritize it to avoid potential merge conflicts and duplicated effort.

Youssefalaa7:
@Tasneemmohammed0 yes that is true it will help a lot , that's why I added a utility class so other contributers can use it in future when this gets merged

Youssefalaa7:
@romani I think we reviewed the diff report and discussed the edge cases that should not add violation so is there anything to be done here?

Youssefalaa7:
@romani i noticed there is new issues for solving this by extending AbstractJavadocCheck but i think this approach enabling comment nodes is also good appraoch and will be more efficient so what do you think? and if my current approach is not recommended i think we should close this PR.

PR #18883, opened 2026-02-06
Issue #16361: Updated testAddException in SarifLoggerTest exception test to use verifyWithInlineConfigParserAndLogger

VidytBhudolia:
Issue #16361

Refactored SarifLoggerTest#testAddException to use verifyWithInlineConfigParserAndLogger, aligning the test with real Checkstyle execution and existing logger tests.

Added InputSarifLoggerSingleException.java with inline XML configuration to reliably trigger an exception during analysis without manual logger invocation.


VidytBhudolia:
testAddException validates SARIF exception rendering using a controlled test exception.
Converting it to inline execution would cause the failure to occur during Checker setup, before the logger is invoked, and would change the semantics of the test.

PR #18885, opened 2026-02-06
Issue #18615: fix text block indentation

### LINKED ISSUE #18615: False positive: indentation of text block inside annotation array parameters

```cmd
E:\temp\CheckstyleReport>type config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">
  <module name="TreeWalker">
    <module name="Indentation" />
  </module>
</module>

E:\temp\CheckstyleReport>type Test.java
public class Test {
    @interface Anno {
        String [] value();
    }

    @Anno({
        """
        example
        """
    })
    void method() {
    }
}
E:\temp\CheckstyleReport>set RUN_LOCALE="-Duser.language=en -Duser.country=US"

E:\temp\CheckstyleReport>java %RUN_LOCALE% -jar checkstyle-13.0.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] E:\temp\CheckstyleReport\Test.java:9:9: 'annotation array initialization' child has incorrect indentation level 8, expected level should be 12. [Indentation]
Audit done.
```

End of text block formatting shouldn't cause violation of indentation rules.

M-SaaD-H:
fixes #18615 

Fixes the incorrect violation in the text block end in annotation

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
@@ -330,7 +330,14 @@ public abstract class AbstractExpressionHandler {
             // checked by a child expression)
 
             if (col != null) {
-                checkLineIndent(astSet.getAst(index), level, false);
+                final DetailAST ast = astSet.getAst(index);
+                final boolean textBlockEndAligned =
+                    ast.getType() == TokenTypes.TEXT_BLOCK_LITERAL_END
+                        && isOnStartOfLine(ast.getParent())
+                        && indentLevel.isAcceptable(expandedTabsColumnNo(ast));
+                if (!textBlockEndAligned) {
+                    checkLineIndent(ast, level, false);
+                }
             }
         }
     }
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationAnnArrInitTextBlock.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationAnnArrInitTextBlock.java
@@ -1,0 +1,119 @@
+package                                                             //indent:0 exp:0
+    com.puppycrawl.tools.checkstyle.checks.indentation.indentation; //indent:4 exp:4
+
+public class InputIndentationAnnArrInitTextBlock {                  //indent:0 exp:0
+    @SuppressWarnings({                                             //indent:4 exp:4
+//below indent:8 exp:8
+        """
+        text block content                                          //indent:8 exp:8
+        on multiple lines                                           //indent:8 exp:8
+        """                                                         //indent:8 exp:8
+    })                                                              //indent:4 exp:4
+    public void method() {                                          //indent:4 exp:4
+    }                                                               //indent:4 exp:4
+
+    @SuppressWarnings({                                             //indent:4 exp:4
+//below indent:8 exp:8
+        """
+        text block content                                          //indent:8 exp:8
+        on multiple lines                                           //indent:8 exp:8
+        with text block end                                         //indent:8 exp:8
+        not being at start                                          //indent:8 exp:8
+        of the line"""                                              //indent:8 exp:8
+    })                                                              //indent:4 exp:4
+    public void method1() {                                         //indent:4 exp:4
+    }                                                               //indent:4 exp:4
+
+    @SuppressWarnings({                                             //indent:4 exp:4
+//below indent:8 exp:8
+        """
+        another text block                                          //indent:8 exp:8
+            """                                                     //indent:12 exp:12
+    })                                                              //indent:4 exp:4
+    public void method2() {                                         //indent:4 exp:4
+    }                                                               //indent:4 exp:4
+
+    @SuppressWarnings({                                             //indent:4 exp:4
+//below indent:4 exp:8,26,68 warn
+    """
+    text block with wrong opening delimiter indent                  //indent:4 exp:4
+        """                                                         //indent:8 exp:8
+    })                                                              //indent:4 exp:4
+    public void method4() {                                         //indent:4 exp:4
+    }                                                               //indent:4 exp:4
+
+    @SuppressWarnings({                                             //indent:4 exp:4
+//below indent:4 exp:8,26,68 warn
+    """
+    text block with wrong opening and closing delimiter             //indent:4 exp:4
+    """                                                             //indent:4 exp:12,30,72 warn
+    })                                                              //indent:4 exp:4
+    public void method5() {                                         //indent:4 exp:4
+    }                                                               //indent:4 exp:4
+
+    @SuppressWarnings(                                              //indent:4 exp:4
+        {                                                           //indent:8 exp:8
+//below indent:12 exp:8,12
+            """
+            text block with brace on own line                       //indent:12 exp:12
+            and correct indentation                                 //indent:12 exp:12
+            """                                                     //indent:12 exp:8,12
+        })                                                          //indent:8 exp:8
+    public void method6() {                                         //indent:4 exp:4
+    }                                                               //indent:4 exp:4
+
+    @SuppressWarnings(                                              //indent:4 exp:4
+        {                                                           //indent:8 exp:8
+//below indent:12 exp:8,12
+            """
+            text block with brace on own line                       //indent:12 exp:12
+            end delimiter extra indented                            //indent:12 exp:12
+                """                                                 //indent:16 exp:16
+        })                                                          //indent:8 exp:8
+    public void method7() {                                         //indent:4 exp:4
+    }                                                               //indent:4 exp:4
+
+    @SuppressWarnings(                                              //indent:4 exp:4
+        {                                                           //indent:8 exp:8
+//below indent:4 exp:8,12,68 warn
+    """
+    brace on own line wrong opening indent                          //indent:4 exp:4
+    """                                                             //indent:4 exp:12,16,72 warn
+    })                                                              //indent:4 exp:4
+    public void method8() {                                         //indent:4 exp:4
+    }                                                               //indent:4 exp:4
+
+    @SuppressWarnings(                                              //indent:4 exp:4
+        {                                                           //indent:8 exp:8
+//below indent:4 exp:8,12,68 warn
+    """
+    brace on own line wrong opening and closing                     //indent:4 exp:4
+        """                                                         //indent:8 exp:8,12
+    })                                                              //indent:4 exp:4
+    public void method9() {                                         //indent:4 exp:4
+    }                                                               //indent:4 exp:4
+    @interface ApiParam {                                           //indent:4 exp:4
+        Example[] examples();                                       //indent:8 exp:8
+    }                                                               //indent:4 exp:4
+    @interface Example {                                            //indent:4 exp:4
+        ExampleProperty[] value();                                  //indent:8 exp:8
+    }                                                               //indent:4 exp:4
+    @interface ExampleProperty {                                    //indent:4 exp:4
+        String value() default "";                                  //indent:8 exp:8
+    }                                                               //indent:4 exp:4
+    @ApiParam(                                                      //indent:4 exp:4
+        examples = @Example(                                        //indent:8 exp:8
+            value = {                                               //indent:12 exp:12
+                @ExampleProperty(                                   //indent:16 exp:16
+//below indent:20 exp:20
+                    value = """
+                            generic text block content              //indent:28 exp:28
+                            on multiple lines                       //indent:28 exp:28
+                            """                                     //indent:28 exp:28
+                )                                                   //indent:16 exp:16
+            }                                                       //indent:12 exp:12
+        )                                                           //indent:8 exp:8
+    )                                                               //indent:4 exp:4
+    public void method10() {                                        //indent:4 exp:4
+    }                                                               //indent:4 exp:4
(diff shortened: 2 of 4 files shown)


M-SaaD-H:

M-SaaD-H:
Hi @romani ,
Can you please review it?

romani:
@vivek-0509 , please help me to review this PR

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
```diff
@@ -517,6 +525,41 @@ protected final void findSubtreeAst(DetailAstSet astSet, DetailAST tree,
...
+     *
+     * @param textBlockEnd the TEXT_BLOCK_LITERAL_END node to check
+     */
+    private void checkTextBlockEndAlignment(DetailAST textBlockEnd) {
+        final DetailAST textBlockBegin = textBlockEnd.getParent();
+        final int beginIndent = expandedTabsColumnNo(textBlockBegin);
+        final int endIndent = expandedTabsColumnNo(textBlockEnd);
+
+        if (beginIndent != endIndent && isOnStartOfLine(textBlockEnd)) {
+            final IndentLevel expectedIndent = new IndentLevel(beginIndent);
```
vivek-0509: here closing """ is being validated against the opening """ column, this would result in incorrect violations in cases where the opening delimiter is itself at a wrong indentation, the error would propagate to the closing delimiter. i thikn we should instead validate against the expected child indentation level of annotation array.

vivek-0509 (changes requested):

src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationAnnArrInitTextBlock.java
```diff
@@ -0,0 +1,41 @@
+package                                                             //indent:0 exp:0
+    com.puppycrawl.tools.checkstyle.checks.indentation.indentation; //indent:4 exp:4
+
+public class InputIndentationAnnArrInitTextBlock {                  //indent:0 exp:0
+    @SuppressWarnings({                                             //indent:4 exp:4
```
vivek-0509, line 5: Please add test cases for { on its own line style, where multiple child indent levels are acceptable.

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
M-SaaD-H: done

src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationAnnArrInitTextBlock.java
M-SaaD-H: done

M-SaaD-H:
Hi @vivek-0509 
I am using the below method

```
private AnnotationArrayInitHandler getAnnotationArrayInitHandler() {
        AbstractExpressionHandler handler = this;
        AnnotationArrayInitHandler result = null;
        while (handler != null) {
            if (handler instanceof AnnotationArrayInitHandler initHandler) {
                result = initHandler;
                break;
            }
            handler = handler.parent;
        }
        return result;
    }
```
    
In this the result can be null according the code, but it will never be because it is called from the `AnnotationArrayInitHandler` context. And I am getting a violation here stating that the result from this function can't be null. Here I have 2 options either use `@Nullable` or throw an exception if the result comes out to be null. Which one should I do? Or is there any othere solutions for this?

vivek-0509:
@M-SaaD-H Use `@Nullable`, i can be null

M-SaaD-H:
Hi @vivek-0509 @romani ,
What should I do for this error?
`
[ERROR] [checkstyle] [ERROR] checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java:34:1: Total number of methods is 35 (max allowed is 34). [MethodCount]
`

vivek-0509:
suppress it with comment

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
```diff
@@ -499,7 +499,17 @@ protected final int expandedTabsColumnNo(DetailAST ast) {
      */
     protected final void findSubtreeAst(DetailAstSet astSet, DetailAST tree,
         boolean allowNesting) {
-        if (!indentCheck.getHandlerFactory().isHandledType(tree.getType())) {
+        // For TEXT_BLOCK_LITERAL_END in annotation array context, check if it aligns
+        // with the expected child indentation of the annotation array init handler
+        // instead of using default indentation rules.
+        final boolean isTextBlockLiteralEnd = tree.getType() == TokenTypes.TEXT_BLOCK_LITERAL_END;
+        final AnnotationArrayInitHandler annotationArrayHandler =
+                getAnnotationArrayInitHandler();
```
vivek-0509: `getAnnotationArrayInitHandler()` walks the entire parent chain for every node visited by `findSubtreeAst()`, but it's only needed when the node is TEXT_BLOCK_LITERAL_END. Please move it inside the isTextBlockLiteralEnd check to avoid unnecessary work

And update checkTextBlockEndAlignment to call getAnnotationArrayInitHandler() internally and return early if null.

vivek-0509 (changes requested):

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
M-SaaD-H: done

M-SaaD-H:
Hi @vivek-0509 
All the required changes are done. Can you please review it?

M-SaaD-H:
@romani @vivek-0509 
following up on this PR since it has been open for a while now. Is this PR still required?

romani:
Yes, please rebase to resolve conflict.
Sorry for delay in reviewing.

M-SaaD-H:
@romani Thanks for the update. I've rebased the branch and resolved the conflicts.

Could you please take another look when you get the chance?

M-SaaD-H:

M-SaaD-H:
@romani @vivek-0509 
All the CIs are green and report is generated. Everything looks fine.
Could you please take a look when you get the chance?

M-SaaD-H:
ping @romani @vivek-0509

romani:
@M-SaaD-H , sorry for veeeery later reply.
please rebase, resolve conflict.

this is false positive:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/a523132cbd63988ef3b58a21971e2c36cd6de0d7_2026051653/reports/diff/pulsar/index.html#A1

M-SaaD-H:

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
```diff
@@ -31,6 +31,7 @@
  * Abstract base class for all handlers.
  *
  */
+// -@cs[MethodCount] Helper methods for text block indentation handling push count to 36.
```
romani: please move this suppression to this area:
https://github.com/checkstyle/checkstyle/blob/e5e45cc450a163a36712926d2db7bd61572c959e/config/suppressions.xml#L68-L69

romani (changes requested):
items:

romani:
CI failed with:
```
[main] INFO com.github.checkstyle.difftool.DiffTool - Running Checkstyle on src/main/java - finished
Error: Exception in thread "main" java.lang.OutOfMemoryError: UTF16 String size is 1898496287, should be less than 1073741823
```

we need smaller config.

romani:
Github, generate report for Indentation/Example1

romani:
onemore fix is required https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/58d386457360e0c4980d1896cc841d0f0b638328_2026222031/reports/diff/spring-framework/index.html#A1

people are crazy with annotations :)

M-SaaD-H:

src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/AbstractExpressionHandler.java
M-SaaD-H: done

M-SaaD-H:
Github, generate report for Indentation/Example1

M-SaaD-H:
Github, generate report for Indentation/Example1

romani:
same false-positive.
try to put it on nested class.
```
	@Nested
	@DirtiesContext
	@TestPropertySource(properties = {
			"""
			foo = bar
			""",
			"""
			bar = baz
			""",
			"""
			baz = quux
			"""
	})
```

try to make test method that run on Input wihtout our trailing comments on identation, as they might do side effect.

M-SaaD-H:
Github, generate report for Indentation/Example1

romani (approved):
now good.
thanks a lot !!

@vivek-0509 , please finalize review

vivek-0509:
GitHub, generate report for Indentation/Example3

vivek-0509:
Github, generate report for Indentation/Example3

romani:
@vivek-0509 , report is ready, looks good.

vivek-0509:
Got the clean approach, posting the  review..

vivek-0509:
Thanks for working on this, the false positive is real and your test inputs capture the annotation-array case well. I went through the approach in detail and tested both this PR and an alternative against the full regression corpus, and I think the fix needs a different shape. Let me explain why and share the alternative, since it is small.

The root cause lives in AbstractExpressionHandler.checkRemainingLines, which applies the lineWrappingIndentation bump to every continuation line of a child expression, including a text block's closing delimiter. This PR adds its handling in the same class but intercepts in findSubtreeAst and gates it to annotation arrays twice over: the getExpectedIndentForTextBlockEnd hook is only overridden by AnnotationArrayInitHandler, and isDirectChildOfAnnotationArray additionally requires the EXPR -> ANNOTATION_ARRAY_INIT ancestry. Since checkRemainingLines is shared by all handlers, the identical false positive still fires for plain array initializers, return statements, and switch-rule bodies. Since `isDirectChildOfAnnotationArray` gates this PR's fix to the `EXPR -> ANNOTATION_ARRAY_INIT` ancestry, those cases still fail. For example `String[] arr = { """ ... """ };` with aligned delimiters is still flagged with "expected 12". This is what we can see on the same project list, this PR's diff report shows 3 removals ([diff report for this PR](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/ffa6bbfc9f39cc925ad583cb9c9da36365dd5203_2026063340/reports/diff/index.html)), while fixing the shared method shows 8 ([diff report for the alternative](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/0d5a4cae46e6ad5daeccb7ee52d98206f9e0d054_2026192349/reports/diff/index.html)), including `new String[] { """ ... """ }` in openjdk's `CreateSymbolsTestImpl.java` and aligned return-value blocks in openjdk and rhsm. The 3 removals overlap exactly, and the extra 5 are the same false positive in sibling contexts...

Separately, `checkTextBlockEndAlignment` introduces new violations under `forceStrictCondition=false`, the exact-membership check (`!expectedIndent.isAcceptable(endIndent)`) flags over-indented closing delimiters (the `method2` and `method7` cases) even though non-strict mode is documented as accepting deeper-than-required indentation for all continuation lines, and master accepts these layouts today. Vertical alignment enforcement also already exists as an opt-in in `TextBlockGoogleStyleFormattingCheck` via `textblock.vertically.unaligned`, so I would drop this part entirely, a false-positive fix should only remove violations. The heavier mechanism (the null-returning base-class hook plus ancestry probe plus new message subtype) is also what forced the PMD, checker-framework, and MethodCount suppression changes in this PR, none of which are needed if the fix sits at the defect site.

The alternative is a guard in `checkRemainingLines` itself:

```java
final DetailAST ast = astSet.getAst(index);
final boolean textBlockEndAligned =
    ast.getType() == TokenTypes.TEXT_BLOCK_LITERAL_END
        && isOnStartOfLine(ast.getParent())
        && indentLevel.isAcceptable(expandedTabsColumnNo(ast));
if (!textBlockEndAligned) {
    checkLineIndent(ast, level, false);
}
```

The three conditions do the scoping: only the closing delimiter is affected; only when the opening delimiter starts its own line, so `"prefix" + """` and `return ("""` keep current behavior and the `method13` case still flags; and only when the closing sits exactly at a level acceptable for the opening line, so aligning with a misplaced opening earns nothing, both delimiters still get flagged, and the existing `testMultiLineStatements` expectation is preserved. It reuses the existing messages, so there is no suppression-config churn and no message change. 

Your test inputs are good and mostly transfer as-is just few adjustments needed are that lines with over-indented closings become non-warn.

Check the implementation here and please copy as in https://github.com/checkstyle/checkstyle/pull/19829

M-SaaD-H:
@vivek-0509 
Sorry for the late reply, and thanks for the review and suggesting the alternative approach.
I've updated the implementation as suggested. Please take another look.

romani:
Github, generate report for Indentation/Example3

vivek-0509 (approved):
Good to merge!

Issue #18882, opened 2026-02-06
Find way to keep qodana yml in config folder


romani:
https://github.com/checkstyle/checkstyle/issues/17104 migration to qodana is done.

But  one configuration issue was not resolved and in not ideal way.

https://github.com/checkstyle/checkstyle/pull/17280#issuecomment-3857640500

We need to find way to move https://github.com/checkstyle/checkstyle/blob/master/qodana.yml

To https://github.com/checkstyle/checkstyle/tree/master/config folder.

Atharv3221:
issue created https://youtrack.jetbrains.com/issue/QD-13534/CI-fails-with-configuration-in-non-root-path

Atharv3221:
Solution recived:
https://youtrack.jetbrains.com/issue/QD-13534/CI-fails-with-configuration-in-non-root-path#focus=Comments-27-13319631.0-0

So we need to merge it in once even if it fails once its there for master branch it won't fail again for other Prs

romani:
@Atharv3221, please send PR we can quickly merge.
Please send one more PR with bad changes to let me test it after merge

romani:
@Atharv3221 , please do https://youtrack.jetbrains.com/issue/QD-13534/CI-fails-with-configuration-in-non-root-path#focus=Comments-27-13321567.0-0

And close issue on qodana , as appreciation on quick response.

Full doc https://www.jetbrains.com/help/qodana/github.html#Pull+requests

Atharv3221:
This issue can be closed

romani:
Fixed by update at https://github.com/checkstyle/checkstyle/pull/18915

PR #18886, opened 2026-02-06
Issue #18882: Move qodana.yml to config directory

VidytBhudolia:
Issue #18882

Moves qodana.yml into the config directory and updates the Qodana GitHub Actions workflow to explicitly reference the new configuration path, keeping CI behavior unchanged while aligning with project config layout.


romani:
Fixed by update in other PR

PR #18890, opened 2026-02-07
minor: technical `quoting` can be avoided in yml

Pankraz76:
### minor: technical `quoting` can be avoided in yml

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,14 +29,14 @@ jobs:
           path: /home/circleci/project/target/inspection-results
 
   validate-with-maven-script:
-    description: "Runs a maven script using the job name as the argument."
+    description: Runs a maven script using the job name as the argument.
     parameters: &script_parameters
       image-name:
         type: string
-        default: "cimg/openjdk:21.0.6"
-        description: "docker image to use"
+        default: cimg/openjdk:21.0.6
+        description: docker image to use
       command:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -58,13 +58,13 @@ jobs:
             ./.ci/validation.sh ci-temp-check
       - save_cache:
           name: Save Maven repo and wrapper cache
-          key: mvn-cache-{{ checksum "pom.xml" }}
```
Pankraz76, line 61: tryout here is the only time needed....

Pankraz76: seems to work.

Pankraz76:
rdy 2 merge this trivial stuff

romani:
Conflict.

.circleci/config.yml
```diff
@@ -16,7 +16,7 @@ jobs:
             echo "Java version:"
             java --version
             echo "IDEA version:"
-            echo $IDEA_VERSION
+            echo "$IDEA_VERSION"
```
romani: Do not touch bash command echo

romani (changes requested):
Items

.circleci/config.yml
Pankraz76: you told all. echo work better with quoting.

PR #18896, opened 2026-02-08
Issue checkstyle#18882: Move Quodana to config directory

AmitKumarDeoghoria:
Issue: #18882

### DIFF
--- a/.github/workflows/qodana.yml
+++ b/.github/workflows/qodana.yml
@@ -20,8 +20,21 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Find and use qodana.yml
+        run: |
+          if [ -f "config/qodana.yml" ]; then
+            echo "Using qodana.yml from config directory"
+            cp config/qodana.yml .
+          else
+            echo "ERROR: No qodana.yml found!"
+            find . -name "*.yml" -o -name "*.yaml" | head -10
+            exit 1
+          fi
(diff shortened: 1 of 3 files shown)


.github/workflows/qodana.yml
```diff
@@ -20,8 +20,21 @@ jobs:
...
+        run: |
+          if [ -f "config/qodana.yml" ]; then
+            echo "Using qodana.yml from config directory"
+            cp config/qodana.yml .
+          else
+            echo "ERROR: No qodana.yml found!"
+            find . -name "*.yml" -o -name "*.yaml" | head -10
+            exit 1
+          fi
+
```
AmitKumarDeoghoria, line 33: @romani  we can use this hack, if qodana do expects file to be in root dir and doesn't bother to check sub directories then we can just copy the file to root during runtime, this way we still have the original at config/qodana.yml and the copy disappears after the run since github uses fresh containers in each run.

romani: not good, better to let it sit in root in this case.

romani: it is very strange that thye have property to define path to it, so it should works.

romani: can we create issue on them to fix customization of configration ?

romani:
Fixed by update in other PR

PR #18901, opened 2026-02-09
Issue #18435: corrected AST for rightcurly eaxmples

Anushreebasics:
Issue #18435

made example 2,3,4 to be AST consistent
Example5 remains suppressed because it uses switch/case, which cannot be made AST-consistent with the canonical example that demonstrates if/else and try/catch multi-block statements.


Anushreebasics:
@smita1078 @romani please review

Issue #18146, opened 2025-11-21
Clean up grammar for discontinued preview feature (pattern matching for switch)


stoyanK7:
Follow-up from issue
- https://github.com/checkstyle/checkstyle/issues/18104

---

[JEP 406](https://openjdk.org/jeps/406) introduced a new preview feature in Java 17 called `Pattern Matching for switch`. Checkstyle added support for this in Issue #10848 and PR #11100. Since then, however, the feature got refined multiple times in:

- [JEP 420: Pattern Matching for switch (Second Preview)](https://openjdk.org/jeps/420)
- [JEP 427: Pattern Matching for switch (Third Preview)](https://openjdk.org/jeps/427)
- [JEP 433: Pattern Matching for switch (Fourth Preview)](https://openjdk.org/jeps/427)

In those refinements, some syntax was dropped and it is no longer compilable with Java 21. Checkstyle still supports this outdated grammar.

The goal of this issue is to remove ANTLR grammar for `Pattern Matching for switch`. The grammar was added in PR
- #11100

stoyanK7:
~~I am on it.~~

I am backing away from this issue for now. It's free for grabs.

0xChaitanya:
is this issue resolved?

romani:
@0xChaitanya , no , but I highly recommend to new contributors to avoid it, it is very very complicated.

vivek-0509:
i am on it

romani:
breaking changes are done at https://github.com/checkstyle/checkstyle/pull/18908

PR #18908, opened 2026-02-10
Issue #18146: Remove ANTLR grammar for discontinued pattern matching …

vivek-0509:
Issue: #18146

Removed ANTLR grammar for syntax that was part of the Pattern Matching for switch preview feature but was dropped before the feature was finalized in Java 21. This syntax is no longer compilable.

1. Guarded patterns using `&&` (e.g., `case String s && s.length() > 1`) were replaced by when clauses in JEP 427.

2.  Parenthesized patterns (e.g., `case (Integer i)`) were removed in JEP 441 because they did not have sufficient value once `&&` was replaced by when

Official references:
- [JEP 427](https://openjdk.org/jeps/427): "Guarded patterns are replaced with [when](cci:2://file:///Users/vivek/checkstyle/src/test/resources-noncompilable/com/puppycrawl/tools/checkstyle/grammar/java19/InputJava19GuardsWithExtraParenthesis.java:28:4-28:34) clauses in switch blocks."
- [JEP 441](https://openjdk.org/jeps/441): "Remove parenthesized patterns, since they did not have sufficient value."


vivek-0509:

vivek-0509:
Hey @romani @mahfouz72 , here is the antlr regression report can you please take a look
 https://vivek-0509.github.io/Reports/

vivek-0509:
@mahfouz72 @romani please take  a look when you have time

mahfouz72:
@vivek-0509 Thanks a lot, I will take a look ASAP

src/main/java/com/puppycrawl/tools/checkstyle/JavaAstVisitor.java
```diff
@@ -1970,15 +1961,13 @@ public DetailAstImpl visitPattern(JavaLanguageParser.PatternContext ctx) {
         final JavaLanguageParser.InnerPatternContext innerPattern = ctx.innerPattern();
         final ParserRuleContext primaryPattern = innerPattern.primaryPattern();
         final ParserRuleContext recordPattern = innerPattern.recordPattern();
-        final boolean isSimpleTypePattern = primaryPattern != null
-                && primaryPattern.getChild(0) instanceof JavaLanguageParser.TypePatternContext;
```
mahfouz72, line 1974: Can you explain to me why we removed this condition?

```ANTLR
primaryPattern
    : typePattern                                                          #patternVariableDef
    | LPAREN innerPattern RPAREN                                           #parenPattern
    | recordPattern                                                        #recordPatternDef
    ;

typePattern
    : mods+=modifier* type=typeType[true] id             #typePatternDef
    | LITERAL_UNDERSCORE                                 #unnamedPatternDef
    ;

```

vivek-0509: Before this change, primaryPattern had 3 alternatives, `#patternVariableDef, #parenPattern, and #recordPatternDef`.
the instanceof TypePatternContext check was needed to distinguish between a type pattern `Integer i `and a parenthesized pattern `(Integer i)`, since both made primaryPattern non-null. They required different AST handling,  a type pattern is visited directly as a self-contained node, while a parenthesized pattern needed a PATTERN_DEF imaginary parent node to group(wrap)  the LPAREN, inner pattern, and RPAREN together.
After removing #`parenPattern` from the grammar, primaryPattern can only be `#patternVariableDef or #recordPatternDef`. Since record patterns are already handled by the `if (recordPattern != null)` check above, primaryPattern != null is sufficient to know it's a type pattern, the instanceof check is no longer needed now.

config/projects-to-test/openjdk17-excluded.files
```diff
@@ -512,3 +512,32 @@
 <module name="BeforeExecutionExclusionFileFilter">
   <property name="fileNamePattern" value="[\\/]test[\\/]langtools[\\/]tools[\\/]javac[\\/]switchexpr[\\/]ParserRecovery.java$"/>
 </module>
+<!-- && guard and parenthesized patterns removed from grammar -->
+<module name="BeforeExecutionExclusionFileFilter">
+  <property name="fileNamePattern" value="[\\/]test[\\/]langtools[\\/]tools[\\/]javac[\\/]diags[\\/]examples[\\/]CantRefNonEffectivelyFinalVar.java$"/>
```
romani, line 517: why we exclude files ? 
if we drop support, we should delete file and forget about it.

what is reason to keep ?

romani: ok, it is link to open jdk file, not a file in our reporsitory.

PR #18917, opened 2026-02-11
infra: updating arguments for qodana inspection

Atharv3221:
infra
## Changes
- Use `image` parameter instead of `linter` (Docker version of Qodana)
- Disable PR mode
- Disable annotations

### DIFF
--- a/.github/workflows/qodana.yml
+++ b/.github/workflows/qodana.yml
@@ -18,12 +18,15 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v6
         with:
-          fetch-depth: 0
+          fetch-depth: 1
 
       - name: Run Qodana Scan
         uses: JetBrains/qodana-action@v2025.3
         with:
           use-caches: false
+          use-annotations: false
+          pr-mode: false
+
           args: |
             --config,config/qodana.yml,
             --print-problems
--- a/config/intellij-idea-inspections.xml
+++ b/config/intellij-idea-inspections.xml
@@ -1088,6 +1088,9 @@
     <option name="m_includeLibraryClasses" value="false"/>
     <option name="m_limit" value="15"/>
   </inspection_tool>
+  <!-- temporary disabled until https://github.com/checkstyle/checkstyle/issues/18926 -->
+  <inspection_tool class="ClassEscapesItsScope" enabled="false" level="WARNING"
+                   enabled_by_default="false" />
   <!-- we do not need that -->
   <inspection_tool class="ClassHasNoToStringMethod" enabled="false" level="ERROR"
                    enabled_by_default="false">
@@ -1679,7 +1682,8 @@
   </inspection_tool>
   <inspection_tool class="EmptyCatchBlockJS" enabled="true" level="ERROR"
                    enabled_by_default="true"/>
-  <inspection_tool class="EmptyClass" enabled="true" level="WARNING" enabled_by_default="true">
+  <!-- temporary disabled until https://github.com/checkstyle/checkstyle/issues/18926 -->
+  <inspection_tool class="EmptyClass" enabled="false" level="WARNING" enabled_by_default="false">
     <option name="ignorableAnnotations">
       <value/>
     </option>
@@ -1763,8 +1767,9 @@
                    enabled_by_default="true"/>
   <inspection_tool class="EqualsWhichDoesntCheckParameterClass" enabled="true" level="ERROR"
                    enabled_by_default="true"/>
-  <inspection_tool class="EqualsWithItself" enabled="true" level="ERROR"
-                   enabled_by_default="true"/>
+  <!-- temporary disabled until https://github.com/checkstyle/checkstyle/issues/18926 -->
+  <inspection_tool class="EqualsWithItself" enabled="false" level="ERROR"
+                   enabled_by_default="false"/>
   <!-- there is false-positive, and rest cases are result of our loading modules by reflection
     that throws Error. Error is not always jvm problem,
     a bunch of libraries throw Error instead of Exception -->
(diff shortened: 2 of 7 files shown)


Atharv3221:
@romani 
after disabling pr mode old issues in project are visible, please guide next steps

Atharv3221:
earlier it was checking just changes so no old errors pop up

romani:
Please share list of district "Rule: "
We might need to disable such rules until we fix them one by one in separate PRs

Atharv3221:
Okay

Atharv3221:
Ci failure is not Pr related

config/qodana.yml
```diff
@@ -47,3 +47,31 @@ exclude:
...
+    paths:
+      - src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
+  - name: JavadocReference
+    paths:
+      - src/main/java/com/puppycrawl/tools/checkstyle/JavaAstVisitor.java
+      - src/main/java/com/puppycrawl/tools/checkstyle/JavadocCommentsAstVisitor.java
+      - src/test/java/com/puppycrawl/tools/checkstyle/JavaAstVisitorTest.java
+
+  # These inspections are temporary disabled globally
+  - name: ClassCanBeRecord
```
romani: Please create issue for this list of inspections .
And we need to suppress with comment `...temporary disabled globally until #xxxx`
And we should do this in  xml config., as it has all details on configuration, same config is used on local to reproduce probli

romani (changes requested):
Very good.
Items to improve

config/qodana.yml
Atharv3221: I feel we shouldn't disable it in local it will hide inspection errors and might increase the number of problems. Keeping it active locally will atleast help to not add more problems.

romani: We need to work the same in CI and local, this is the only way out. All tools are working like this.
Only in this case there is clarity.

Atharv3221:
Okay

Atharv3221:
```
20 suspicious problems were detected during Sanity Check. Probably there are some troubles with corresponding inspections. You may fix the problems or exclude the files containing them from the analysis
-------------------------------------------------------------------------------------------------------------------------
File                                                                                 Inspection  Severity  Problems count
-------------------------------------------------------------------------------------------------------------------------
JavaAstVisitor.java                                                                 Java sanity  Critical               4
JavaParser.java                                                                     Java sanity  Critical               4
JavadocDetailNodeParser.java                                                        Java sanity  Critical               4
JavaAstVisitorTest.java                                                             Java sanity  Critical               4
GeneratedJavaTokenTypesTest.java                                                    Java sanity  Critical               2
JavadocCommentsAstVisitor.java                                                      Java sanity  Critical               1
GeneratedJavadocCommentsTokenTypesTest.java                                         Java sanity  Critical               1
-------------------------------------------------------------------------------------------------------------------------
```

config/intellij-idea-inspections.xml
```diff
@@ -1679,14 +1679,6 @@
   </inspection_tool>
   <inspection_tool class="EmptyCatchBlockJS" enabled="true" level="ERROR"
                    enabled_by_default="true"/>
-  <inspection_tool class="EmptyClass" enabled="true" level="WARNING" enabled_by_default="true">
```
romani, line 1682: please do us a favor to keep `<inspection_tool` as it was before at position.
it is how IDEA does extraction to xml, so better to preserve order of export from IDEA.

it is ok that there will be numerous `until ` comments.

romani (changes requested):
if there is any new classes, please put them alphabeticaly.

config/qodana.yml
```diff
@@ -47,3 +47,16 @@ exclude:
       - src/test/java/com/puppycrawl/tools/checkstyle/grammar/javadoc/ParseTreeBuilder.java
 
       - docs/GRAMMAR_UPDATES.md
+
+  # Non-standard package names in IT folder are by design
+  - name: PackageNamingConvention
```
romani: can we disable in xml ?

config/qodana.yml
```diff
@@ -47,3 +47,16 @@ exclude:
...
+
+  # Javadoc is intentional
+  - name: JavadocDeclaration
+    paths:
+      - src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
+  - name: JavadocReference
+    paths:
+      - src/main/java/com/puppycrawl/tools/checkstyle/JavaAstVisitor.java
```
romani: Can we use annotation to disable?
I don't want to much ways to configure behavior. All should be by xml

romani (changes requested):
Items

config/qodana.yml
Atharv3221: done

config/intellij-idea-inspections.xml
Atharv3221: arranged new classed as per idea export

.github/workflows/qodana.yml
```diff
@@ -18,12 +18,15 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v6
         with:
-          fetch-depth: 0
+          fetch-depth: 1
```
Atharv3221, line 21: shallow clone is better here

romani (approved):
Cool. Ok yo merge if CI pass

.github/workflows/qodana.yml
romani: What was a reason to put 1 ?

0 was even more shallow

Atharv3221:
```
  fetch-depth:
    description: 'Number of commits to fetch. 0 indicates all history for all branches and tags.'
    default: 1
```

romani: Good to know

PR #18912, opened 2026-02-11
minor: Consolidate CI jobs like `validate-with-maven-script`

Pankraz76:
### minor: Consolidate CI jobs like `validate-with-maven-script`

pre: (distributed all over)

image

now: (grouped and centralized)

image

SSOT/SPOT

1 vs 2 

Image

Image

PR #18927, opened 2026-02-13
minor: Consolidate `cs_img`

Pankraz76:
### minor: Consolidate `cs_img`

- https://github.com/checkstyle/checkstyle/pull/18912

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
 
   sonarqube:
     docker:
-      - image: amitkumardeoghoria/jdk-21-groovy-git-mvn-ant-jq:v1
+      - image: &cs_img amitkumardeoghoria/jdk-21-groovy-git-mvn-ant-jq:v1
 
     steps:
       - checkout
@@ -95,7 +95,7 @@ workflows:
       # no-exception-test script
       - validate-with-maven-script:
           name: no-exception-lucene-and-others-javadoc
-          image-name: &cs_img amitkumardeoghoria/jdk-21-groovy-git-mvn-ant-jq:v1
+          image-name: *cs_img
           command: ./.ci/no-exception-test.sh no-exception-lucene-and-others-javadoc
       - validate-with-maven-script:
           name: no-exception-cassandra-storm-tapestry-javadoc


.circleci/config.yml
```diff
@@ -95,7 +94,7 @@ workflows:
       # no-exception-test script
       - validate-with-maven-script:
           name: no-exception-lucene-and-others-javadoc
-          image-name: &cs_img amitkumardeoghoria/jdk-21-groovy-git-mvn-ant-jq:v1
```
Pankraz76, line 98: Having security via singularity seems to be related to the SSOT and SPOT principle (1 is better than 2).

.circleci/config.yml
```diff
@@ -207,17 +206,16 @@ workflows:
...
+          image-name: *cs_img
           command: ./.ci/run-link-check-plugin.sh --skip-external
       - validate-with-maven-script:
           name: no-exception-samples-gradle
-          image-name: cimg/openjdk:21.0.6
+          image-name: *cs_img
           command: ./.ci/no-exception-test.sh no-exception-samples-gradle
       - validate-with-maven-script:
           name: no-exception-samples-maven
-          image-name: cimg/openjdk:21.0.6
```
Pankraz76, line 218: Having different versions makes no sense; it only creates potential for error and definite maintenance costs.

.circleci/config.yml
```diff
@@ -286,24 +284,24 @@ workflows:
     jobs:
       - validate-with-maven-script:
           name: openrewrite-checkstyle-auto-fix
-          image-name: cimg/openjdk:21.0
```
Pankraz76, line 289: Having different versions makes no sense; it only creates potential for error and definite maintenance costs.

.circleci/config.yml
```diff
@@ -195,7 +194,7 @@ workflows:
           command: ./.ci/validation.sh assembly-run-all-jar
       - validate-with-maven-script:
           name: no-error-test-sbe
-          image-name: cimg/openjdk:21.0.6
+          image-name: *cs_img
```
romani: We don't need to use our special image, the less we use our special image the better

Pankraz76: so what to do? the code tells a different story. using the img everywhere.

romani: Revert such updates

.circleci/config.yml
```diff
@@ -56,8 +56,7 @@ jobs:
 
   sonarqube:
     docker:
-      - image: amitkumardeoghoria/jdk-21-groovy-git-mvn-ant-jq:v1
```
Pankraz76, line 59: maybe dont need this extra extra

.circleci/config.yml
Pankraz76: lets have SSOT and we good thats the goal to consolidate and unite. salam aleikum

Issue #18031, opened 2025-11-01
Resolve Pitest Suppressions - java-ast-visitor


romani:
As a part of this issue, we have to kill the mutation listed below and remove the suppression from [pitest xxxx suppression.xml](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/)

To understanding how to kill the mutation 
Visit [Wiki page](https://github.com/checkstyle/checkstyle/wiki/How-to-Generate-and-Analyze-Pitest-Reports), [Pitest-docs](https://pitest.org/) and expected actions and workflow at https://github.com/checkstyle/checkstyle/issues/12341.

Target is to remove all `<mutation` from [Pitest-java-ast-visitor](https://github.com/checkstyle/checkstyle/blob/master/config/pitest-suppressions/pitest-java-ast-visitor-suppressions.xml)

Zeeshan3h3:
i go through this issue and find it in my domain basically i have Kill all mutations one by one until that suppression file becomes empty, then the file can be deleted entirely.

romani:
Yes, one mutation per PR, do none greedy, it might not be very easy activity

Anushreebasics:
I am on it

romani:
@Anushreebasics , choose single muation and target it, this issue created to let multiple contributors to particiapate in the same time in separate PRs.

romani:
fix is merged !!!
@Youssefalaa7  did it.

PR #18942, opened 2026-02-16
Issue #18031: Resolve Pitest Suppressions - java-ast-visitor

Youssefalaa7:
Issue #18031: Resolve Pitest Suppressions - java-ast-visitor


src/main/java/com/puppycrawl/tools/checkstyle/JavadocCommentsAstVisitor.java
```diff
@@ -701,12 +692,7 @@ private JavadocNodeImpl createImaginary(int tokenType) {
         final JavadocNodeImpl node = new JavadocNodeImpl();
         node.setType(tokenType);
         node.setText(JavadocUtil.getTokenName(tokenType));
-
-        if (tokenType == JavadocCommentsTokenTypes.JAVADOC_CONTENT) {
-            node.setLineNumber(blockCommentLineNumber);
-            node.setColumnNumber(javadocColumnNumber);
-        }
-
+        node.setLineNumber(blockCommentLineNumber);
```
romani, line 697: setColumnNumber should be required if node.setLineNumber is required 

We just need some test that validates column.

romani:
@mahfouz72, please share your feedback on this update.
You might quickly suggest how to kill mutation

Youssefalaa7:
@romani the setLineNumber mutation was killed because of exception not because of a test the only way to test these two setters is to make a test that give violation on an empty javadoc because the getColNumber and getLineNumber overide the setting we made if they find a child as in the code so we need empty javadoc to have no childs such that our setting will have effect and will not be overriden so the only check i found that could give violation on empty javadoc was SummaryJavadocCheck so i made it expect a row and a column not only row to kill the mutation and added a test  for this case

```
    public int getColumnNumber() {
        final JavadocNodeImpl node = firstChild;
        if (node != null) {
            columnNumber = node.getColumnNumber();
        }
        return columnNumber;
    }
```

config/pitest-suppressions/pitest-java-ast-visitor-suppressions.xml
mahfouz72, line 1: If we removed all mutation here then we should remove this file

src/main/java/com/puppycrawl/tools/checkstyle/JavadocCommentsAstVisitor.java
```diff
@@ -280,39 +274,35 @@ public JavadocNodeImpl visitCustomBlockTag(JavadocCommentsParser.CustomBlockTagC
...
-                case JavadocCommentsLexer.LITERAL ->
-                    buildImaginaryNode(JavadocCommentsTokenTypes.LITERAL_INLINE_TAG, ctx);
-                case JavadocCommentsLexer.SNIPPET ->
-                    buildImaginaryNode(JavadocCommentsTokenTypes.SNIPPET_INLINE_TAG, ctx);
-                default -> buildImaginaryNode(JavadocCommentsTokenTypes.CUSTOM_INLINE_TAG, ctx);
-            };
-            inlineTagNode.addChild(specificTagNode);
-        }
+        final Token tagName =
+                (Token) ctx.inlineTagContent().getChild(0).getChild(0).getPayload();
```
mahfouz72:
```suggestion
        final Token tagContent =
                (Token) ctx.inlineTagContent().getChild(0);
        final Token tagName = tagContent.getChild(0).getPayload();
```

Introducing an intermediate variable (tagContent) improves readability and makes the AST navigation clearer.

src/main/java/com/puppycrawl/tools/checkstyle/JavadocCommentsAstVisitor.java
```diff
@@ -604,7 +594,6 @@ private void processChildren(JavadocNodeImpl parent, List<? extends ParseTree> c
                     accumulator.append(token);
                 }
```
mahfouz72, line 605: After carefully reviewing the grammar structure, removing `accumulator.flushTo(parent);` from the `else if (token.getType() != Token.EOF)` branch appears to be safe.

In the grammar, `TEXT` never appears at the same tree level directly followed by a non-text terminal node.

 Parser rule contexts (`ParserRuleContext`) always appear after it, not terminal nodes. Therefore, when transitioning from `TEXT` to a structural element, execution always enters the second else branch, where the `accumulator.flushTo(parent);` is already called.

src/main/java/com/puppycrawl/tools/checkstyle/JavadocCommentsAstVisitor.java
```diff
@@ -703,12 +692,8 @@ private JavadocNodeImpl createImaginary(int tokenType) {
         final JavadocNodeImpl node = new JavadocNodeImpl();
         node.setType(tokenType);
         node.setText(JavadocUtil.getTokenName(tokenType));
```
mahfouz72, line 705: Could you explain why this is correct? 

We now always set the line number to the block comment’s line number and the column number to the Javadoc’s starting column for all imaginary nodes.

config/pitest-suppressions/pitest-java-ast-visitor-suppressions.xml
Youssefalaa7: no we always leave the suppressions file empty after finishing all mutations in all previous examples we do not delete it

src/main/java/com/puppycrawl/tools/checkstyle/JavadocCommentsAstVisitor.java
Youssefalaa7: i introduced intermediate variable but made tagContent as ParseTree and did not cast it to Token because below we call getChild() which work only with ParseTree

src/main/java/com/puppycrawl/tools/checkstyle/JavadocCommentsAstVisitor.java
Youssefalaa7: this setting do not matter in anything as it is always become overriden it only affects empty javadocs that do not have children

config/pitest-suppressions/pitest-java-ast-visitor-suppressions.xml
romani: Yes, we keep file, to let system generate patch easily, in case we will decide to do extension of suppression again. Some time it is good to do temporary in PR .

It is not allowed to have mutations, so we will not merge such PRs, that has mutations.

Youssefalaa7:
@romani @mahfouz72 if changes are correct should i start diff testing now?

romani:
Don't wait , please start testing

romani:
You need to exclude Check that you changed, in other case diff report will have a lot of noise due to column in message

Youssefalaa7:

Youssefalaa7:
Github, generate report for InvalidJavadocPosition/all-examples-in-one

Youssefalaa7:
Github, generate report for JavadocBlockTagLocation/all-examples-in-one

Youssefalaa7:

Youssefalaa7:
Github, generate report for JavadocContentLocation/all-examples-in-one

Youssefalaa7:
Github, generate report for JavadocLeadingAsteriskAlign/all-examples-in-one

Youssefalaa7:
Github, generate report for JavadocMissingLeadingAsterisk/all-examples-in-one

Youssefalaa7:
Github, generate report for JavadocMissingWhitespaceAfterAsterisk/all-examples-in-one

Youssefalaa7:
Github, generate report for JavadocPackage/all-examples-in-one

Youssefalaa7:
Github, generate report for JavadocParagraph/all-examples-in-one

Youssefalaa7:
Github, generate report for JavadocStyle/all-examples-in-one

Youssefalaa7:
Github, generate report for JavadocTagContinuationIndentation/all-examples-in-one

Youssefalaa7:
Github, generate report for JavadocType/all-examples-in-one

Youssefalaa7:
Github, generate report for JavadocVariable/all-examples-in-one

Youssefalaa7:
Github, generate report for SingleLineJavadoc/all-examples-in-one

Youssefalaa7:
Github, generate report for MissingJavadocMethod/all-examples-in-one

Youssefalaa7:
Github, generate report for MissingJavadocPackage/all-examples-in-one

Youssefalaa7:
Github, generate report for MissingJavadocType/all-examples-in-one

Youssefalaa7:
@romani @mahfouz72 Done, no changes in the diff

Youssefalaa7:
Congratulations! Looking forward to collaborating on future issues.

PR #18944, opened 2026-02-16
Issue #18228: supressed indentation check

### LINKED ISSUE #18228: False-positive: Suppress indentation check when quotes start at the left margin
Follow-up of #17329
From docs: https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation
> [opening quotes] may either follow the same indentation rules as other constructs, or it may have no indentation at all (so it starts at the left margin).

Currently, TextBlockGoogleStyleFormatting Check covers:

- [x] Both opening and closing quotes are on their own lines.
- [x] Opening quotes are vertically aligned with closing quotes, i.e, both must have the same indentation value
- [ ] #18227 

- [ ] Quotes can have zero indentation 

Code:
```java
/** somejavadoc. */
public class InputTextBlocksIndentation {

  public void textFunc() {

    String zeroIndentation = // False-positive below
"""
content of the block
""";                         // False-positive

    getData( OK, below
    """    
        Indentation of Text-block
    """,  // OK
        5
    );
  }

  public static void getData(String data, int length) {}
```

Cli:

```
$ java -jar checkstyle-11.1.0-all.jar -c google_checks.xml InputTextBlocksIndentation.java
Starting audit...
[WARN] InputTextBlocksIndentation.java:6:1: '"""' has incorrect indentation level 0, expected level should be 8. [Indentation]
[WARN] InputTextBlocksIndentation.java:9:1: '"""' has incorrect indentation level 0, expected level should be 8. [Indentation]
[WARN] InputTextBlocksIndentation.java:12:5: 'method call' child has incorrect indentation level 4, expected level should be 6. [Indentation]
[WARN] InputTextBlocksIndentation.java:14:5: 'method call' child has incorrect indentation level 4, expected level should be 6. [Indentation]
Audit done.
```

The first two violations are not expected, as Google-style allows for indenting opening quotes at the start of the line without any additional indentation. And because closing quotes should be aligned with opening quotes, they are also allowed to appear at the 0th indentation level.

The last two violations are expected because if the indentation level is not zero, then it has to follow the indentation rules like other blocks.

Google-Formatter output:
```java
/** somejavadoc. */
public class InputTextBlocksIndentation {

  public void textFunc() {

    String zeroIndentation =
"""
content of the block
""";

    getData(
        """
            Indentation of Text-block
        """,
        5);
  }

  public static void getData(String data, int length) {}
}
```

Difference between Original and Formatted code:

```java
$ diff -Naru InputTextBlocksIndentation.java FormattedInputTextBlocksIndentation.java
--- InputTextBlocksIndentation.java	2025-12-03 08:25:55.734581300 +0530
+++ FormattedInputTextBlocksIndentation.java	2025-12-03 08:26:53.807180600 +0530
@@ -6,14 +6,13 @@

     getData(
-    """
-        Indentation of Text-block
-    """,
-        5
-    );
+        """
+            Indentation of Text-block
+        """,
+        5);
   }

   public static void getData(String data, int length) {}
```

Anushreebasics:
fixes #18228 
Root Cause: The [IndentationCheck] was checking line-wrapping indentation for all text block quotes, even when they legitimately appear at column 0 per Google style.

Solution: Modified the indentation check to suppress violations for text block quotes at column 0.

### DIFF
--- a/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule489textblocks/InputFormattedTextBlocksIndentation.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule489textblocks/InputFormattedTextBlocksIndentation.java
@@ -35,12 +35,10 @@ public class InputFormattedTextBlocksIndentation {
 
   /** Somejavadoc. */
   public void textFuncIndenation2() {
-    // violation 2 lines below '.* incorrect indentation level 0, expected .* 8.'
     String e2 =
 """
 content of the block e2
 """;
-    // violation above '.* incorrect indentation level 0, expected .* 8.'
 
     getData(
         """
--- a/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule489textblocks/InputTextBlocksIndentation.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter4formatting/rule489textblocks/InputTextBlocksIndentation.java
@@ -36,12 +36,10 @@ public class InputTextBlocksIndentation {
 
   /** Somejavadoc. */
   public void textFuncIndenation2() {
-    // violation 2 lines below '.* incorrect indentation level 0, expected .* 8.'
-    String e2 = // false-positive, ok until #18228
+    String e2 = // zero indentation is allowed
 """
 content of the block e2
-"""; // false-positive, ok until #18228
-    // violation above '.* incorrect indentation level 0, expected .* 8.'
+""";
 
     // violation 2 lines below '.* incorrect indentation level 4, expected .* 6.'
     getData(
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.java
@@ -172,6 +172,9 @@ public class IndentationCheck extends AbstractCheck {
      */
     private boolean forceStrictCondition;
 
+    /** Allow text block opening/closing quotes to start on left margin. */
+    private boolean allowTextBlockQuotesOnLeftMargin;
+
     /**
      * Getter to query strict indent level in line wrapping case. If value is true, line wrap indent
      * have to be same as lineWrappingIndentation parameter. If value is false, line wrap indent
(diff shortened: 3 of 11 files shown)


Anushreebasics:
@romani @mohitsatr  please review this PR

Anushreebasics:
@romani please review

Anushreebasics:
@romani please review this PR

Anushreebasics:
@romani please review

romani:
My concern is that this is Google's style issue, but fix is done in general implementation.
Others might not like this behavior.

@vivek-0509, please share your thoughts on this update

Anushreebasics:
@vivek-0509 please review, I have changed the fix to only concern Google's style

vivek-0509:
@romani
I agree with the concern, we should not modify the IndentationCheck for this. It's a general-purpose check, and this column-0 text block allowance is specific to Google's style guide. The OpenJDK style guide (v6) and Sun/Oracle convention have no such rule, and we will likely add this check in both those check config files in the future.

i think this can be achieved through a SuppressionXpathSingleFilter in google_checks.xml instead.
Also, looking at the changes done in the PR to make it customizable for Google style by adding a new property and keeping the default to false for other cases, but the fix only modifies shouldProcessTextBlockLiteral() in LineWrappingHandler,  this only covers some cases and looks like a partial fix. Cases stated below will still have violations, but generally are rare cases but we should cover them

```
getData(
"""
        content
""",
        5);

public String getQuery() {
    return     
"""                 //  column 0, this is what gets flagged
returned content
""";                //  column 0, also flagged
}

```
Lets try the SuppressionXpathSingleFilter approach if we can achieve this through that.
What do you think ??

vivek-0509:
@Anushreebasics You can close the PR the issue is resolved

romani:
@Anushreebasics, please create issue on Indentation Check, with default confirmation .

We need to detach your update from Google style.
New property might make sense, let's review it from start with new issue.

We can close this PR, for now. Code can be reused in new PR for new issue.

stoyanK7:
@romani Shall we close this?

@Anushreebasics Please see https://github.com/checkstyle/checkstyle/pull/18944#issuecomment-4094851865:
> please create issue on Indentation Check, with default confirmation .
> We need to detach your update from Google style.
New property might make sense, let's review it from start with new issue.

romani:
yes, we failed to review it on time, and lost connection to author.

PR #18945, opened 2026-02-16
Issue #18599: Resolve error-prone violations for UnusedVariable

Youssefalaa7:
Issue #18599: Resolve error-prone violations for UnusedVariable

```
Did you mean 'private static  boolean hasMembers(DetailAST token){return !IGNORED_TYPES.get(token.getType());}'?
Warning:  /home/runner/work/checkstyle/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/checks/header/MultiFileRegexpHeaderCheck.java:[97,20] [UnusedVariable] The field 'headerFiles' is never read.
    (see https://errorprone.info/bugpattern/UnusedVariable)
  Did you mean to remove this line?
```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/header/MultiFileRegexpHeaderCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/header/MultiFileRegexpHeaderCheck.java
@@ -77,6 +77,11 @@ public class MultiFileRegexpHeaderCheck
     private static final String EMPTY_LINE_PATTERN = "^$";
 
     /**
+     * Separator for multiple header file paths in the configuration and messages.
+     */
+    private static final String HEADER_FILE_SEPARATOR = ", ";
+
+    /**
      * Compiled regex pattern for a blank line.
      **/
     private static final Pattern BLANK_LINE = Pattern.compile(EMPTY_LINE_PATTERN);
@@ -113,6 +118,7 @@ public class MultiFileRegexpHeaderCheck
         }
         else {
             files = headerFiles.clone();
+            this.headerFiles = String.join(HEADER_FILE_SEPARATOR, headerFiles);
         }
 
         headerFilesMetadata.clear();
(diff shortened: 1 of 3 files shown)


src/main/java/com/puppycrawl/tools/checkstyle/checks/header/MultiFileRegexpHeaderCheck.java
```diff
@@ -131,7 +137,7 @@ public void setHeaderFiles(String... headerFiles) {
     public String getConfiguredHeaderPaths() {
         return headerFilesMetadata.stream()
                 .map(HeaderFileMetadata::headerFilePath)
-                .collect(Collectors.joining(", "));
+                .collect(Collectors.joining(HEADER_FILE_SEPARATOR));
```
romani: Can we remove getConfiguredHeaderPaths() ?

It is not clear why it is public., but is not used inside.
Only setter and overriden method should be public, all other should be private

Youssefalaa7: @romani because it is used in MultiFileRegexpHeaderCheckTest
 all the tests that are not unit tests use it 

```
final String allConfiguredPaths = checkInstance.getConfiguredHeaderPaths();

        final String[] expected = {
            "3: " + getCheckMessage(MSG_HEADER_MISMATCH, "^$", allConfiguredPaths),
        };
```

romani: Remove test method

romani: We do not need method that is only used in tests

Youssefalaa7:
@romani done removed the method getConfiguredHeaderPaths ,updated the tests  using it and removed the tests using it to tets this method as unit test

Issue #18926, opened 2026-02-13
Re-enable inspections that were temporarily suppressed for Qodana


Atharv3221:
We recently migrated to Qodana for code inspections. During the migration, some inspections were temporarily suppressed to allow the transition to complete without blocking development.

These suppressions are still in place, and the underlying code violations need to be resolved and suppressions should be removed.

https://github.com/checkstyle/checkstyle/blob/master/config/intellij-idea-inspections.xml , search "18926"

Inspections to Re-enable
  - [x] ClassCanBeRecord
  - [ ] ClassEscapesItsScope
  - [ ] EmptyClass
  - [x] EqualsWithItself
  - [ ] ExtractMethodRecommender
  -  [ ] FieldMayBeStatic
  - [ ] Java9ReflectionClassVisibility
  - [x] MappingBeforeCount
  - [x] RedundantSuppression
  - [ ]  SystemGetProperty
  - [x]  UnnecessarilyQualifiedStaticUsage
  - [x]  PropertiesAsHashtable
  - [x] WhileCanBeDoWhile

Target 1 inspection per pr.

Vikash-Kumar-23:
Hi! I would like to work on re-enabling the FieldMayBeStatic inspection. Please assign this to me.

vivek-0509:
@Vikash-Kumar-23 issue is not approved yet please work on approved issues.
In checkstyle there is no assignment of issues just comment I am on it and start working on it if no one else is already working on it

romani:
Issue will be approved after https://github.com/checkstyle/checkstyle/pull/18917 is merged

Atharv3221:
@romani please approve this issue

romani:
Approved

ayushactiveat:
I am on it. I'll be working on re-enabling the `ClassCanBeRecord` inspection.

vivek-0509:
i am on EqualsWithItself

vivek-0509:
i am on RedundantSuppression

vivek-0509:
I  am on UnnecessarilyQualifiedStaticUsage

ayushactiveat:
i am on PropertiesAsHashtable

pirzada-ahmadfaraz:
I am working on re-enabling the `ClassEscapesItsScope` inspection. Draft PR: https://github.com/checkstyle/checkstyle/pull/19103

pirzada-ahmadfaraz:
[...]

PR #18978, opened 2026-02-21
Issue #18926: Re-enable 'ClassCanBeRecord' inspection

ayushactiveat:
Resolves part of #18926.
This PR re-enables the `ClassCanBeRecord` qodana inspection.


romani (changes requested):
Please do refactoring for all files that are not in "/api/" folder.
All in api folders should stay as is to avoid causing problems to users

ayushactiveat:
Hi @romani, I did the changes but now the remaining ci failures, i believe, are unrelated here
please let me know if there's anything else to u'd like changed

PR #18982, opened 2026-02-21
Issue #18926: Re-enable inspections that were temporarily suppressed …

Youssefalaa7:
Issue #18926: Re-enable inspections that were temporarily suppressed for Qodana for FieldMayBeStatic

### DIFF
--- a/config/org.eclipse.jdt.core.prefs
+++ b/config/org.eclipse.jdt.core.prefs
@@ -67,7 +67,6 @@ org.eclipse.jdt.core.compiler.problem.redundantNullAnnotation=error
 org.eclipse.jdt.core.compiler.problem.redundantNullCheck=error
 org.eclipse.jdt.core.compiler.problem.redundantSpecificationOfTypeArguments=error
 org.eclipse.jdt.core.compiler.problem.redundantSuperinterface=error
-org.eclipse.jdt.core.compiler.problem.reportMethodCanBeStatic=error
 org.eclipse.jdt.core.compiler.problem.reportPreviewFeatures=error
 org.eclipse.jdt.core.compiler.problem.specialParameterHidingField=error
 org.eclipse.jdt.core.compiler.problem.staticAccessReceiver=error
@@ -125,6 +124,8 @@ org.eclipse.jdt.core.compiler.problem.potentialNullReference=ignore
 org.eclipse.jdt.core.compiler.problem.potentiallyUnclosedCloseable=ignore
 # Too many false positives for methods that should be static but would cause issues.
 org.eclipse.jdt.core.compiler.problem.reportMethodCanBePotentiallyStatic=ignore
+# until https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4882
+org.eclipse.jdt.core.compiler.problem.reportMethodCanBeStatic=ignore
 # We should print all errors
 org.eclipse.jdt.core.compiler.problem.suppressOptionalErrors=disabled
 org.eclipse.jdt.core.compiler.problem.suppressWarnings=disabled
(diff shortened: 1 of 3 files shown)


Youssefalaa7:
@romani the problem is after fixing qodana ,semaphore want the test to be static and we can not make @Test annotation method be static so still working how to solve this in another way

```
50. ERROR in /home/semaphore/checkstyle/src/test/java/com/puppycrawl/tools/checkstyle/meta/MetadataGeneratorUtilTest.java (at line 63)00:49
	public void testMetadataFilesGenerationAllFiles(@SystemOutGuard.SysOut Capturable systemOut)00:49
            throws Exception {00:49
	            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^00:49
The method testMetadataFilesGenerationAllFiles(Capturable) from the type MetadataGeneratorUtilTest can be declared as static00:49
----------
```

src/test/java/com/puppycrawl/tools/checkstyle/meta/MetadataGeneratorUtilTest.java
```diff
@@ -39,11 +39,12 @@
 @ExtendWith(SystemOutGuard.class)
 public final class MetadataGeneratorUtilTest extends AbstractModuleTestSupport {
 
-    private final Set<String> modulesContainingNoMetadataFile = Set.of(
-            "Checker",
-            "TreeWalker",
-            "ClassAndPropertiesSettersJavadocScraper"
-    );
+    private final Set<String> modulesContainingNoMetadataFile =
+        new LinkedHashSet<>(Set.of(
```
romani: wild hack to cheat inspection.
Please move modulesContainingNoMetadataFile to testMetadataFilesGenerationAllFiles to be local variable

Youssefalaa7: @romani the problem that any change i make give the above error i showed above this is because the compiler flags this because the test method doesn't access any instance members from AbstractModuleTestSupport, making it technically eligible to be static - however, JUnit test methods must remain instance methods

what do you suggest return to the hack above or you have another opinion?

romani: semaphore ci job runs eclipse compiler with eclipse static analysis.
Let's try to put static on field, it is correct update, but make suppression on eclipse violation.
There is config in checkstyle_checks.xml to allow suppression names in annotation.

romani: https://github.com/checkstyle/checkstyle/blob/bdbc3c960830e9af096e49b5a73b06ae45af9a9c/config/checkstyle-checks.xml#L304-L305

romani:
```
Rule: SuppressionAnnotation
File: src/test/java/com/puppycrawl/tools/checkstyle/meta/MetadataGeneratorUtilTest.java
Line: 62
Severity: error
Message: Annotation suppresses 'static-method'
```

is fixable by extension of allowed to suppress anottation.

We need to find way to resolve/workaround/suppress  Eclipse violaiton.

config/org.eclipse.jdt.core.prefs
```diff
@@ -125,6 +124,7 @@ org.eclipse.jdt.core.compiler.problem.potentialNullReference=ignore
 org.eclipse.jdt.core.compiler.problem.potentiallyUnclosedCloseable=ignore
 # Too many false positives for methods that should be static but would cause issues.
 org.eclipse.jdt.core.compiler.problem.reportMethodCanBePotentiallyStatic=ignore
+org.eclipse.jdt.core.compiler.problem.reportMethodCanBeStatic=ignore
```
romani, line 128: Please open issue on eclipse  for this, may be eventually it will be fixed.

romani (approved):
We will merge , but please do following

config/org.eclipse.jdt.core.prefs
Youssefalaa7: @romani done

romani: Please put link to issue in comment above this like, like.
"Until ....."

romani: As you might notice, all suppressions has reason

Youssefalaa7:
@romani done

PR #18987, opened 2026-02-22
minor: align code with Javadoc

Atharv3221:
minor:

```
    /**
     * Check if VARIABLE_DEF is initialized or not.
     *
     * @param ast VARIABLE_DEF to be checked
     * @return true if initialized
     */
    private static boolean isInitialized(DetailAST ast) {
        return ast.getParent().getLastChild().getType() == TokenTypes.ASSIGN;
    }
```
according to method `isIntialized(DetailAst ast)` Javadoc DetailAst passed should be VARIABLE_DEF, but we were passing IDENT
```
    private void insertVariable(DetailAST ast) {
        final Map<String, FinalVariableCandidate> scope = scopeStack.peek().scope;
        final DetailAST astNode = ast.findFirstToken(TokenTypes.IDENT);
        final FinalVariableCandidate candidate = new FinalVariableCandidate(astNode);
        // for-each variables are implicitly assigned
        candidate.assigned = ast.getParent().getType() == TokenTypes.FOR_EACH_CLAUSE;
        scope.put(astNode.getText(), candidate);
        if (!isInitialized(astNode)) {
            scopeStack.peek().uninitializedVariables.add(astNode);
        }
    }

Here ast is VARIABLE_DEF and astNode is IDENT
```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
@@ -511,7 +511,7 @@ public class FinalLocalVariableCheck extends AbstractCheck {
         // for-each variables are implicitly assigned
         candidate.assigned = ast.getParent().getType() == TokenTypes.FOR_EACH_CLAUSE;
         scope.put(astNode.getText(), candidate);
-        if (!isInitialized(astNode)) {
+        if (!isInitialized(ast)) {
             scopeStack.peek().uninitializedVariables.add(astNode);
         }
     }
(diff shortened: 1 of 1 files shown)


Atharv3221:
Fail ci is not pr related

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
```diff
@@ -523,7 +523,7 @@ private void insertVariable(DetailAST ast) {
      * @return true if initialized
      */
     private static boolean isInitialized(DetailAST ast) {
-        return ast.getParent().getLastChild().getType() == TokenTypes.ASSIGN;
+        return ast.getLastChild().getType() == TokenTypes.ASSIGN;
```
romani, line 526: ...

romani (changes requested):
please run diff tesing for affected Check

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.java
romani: we dealt with AST of variable , not its ident.
So it is good update .

Atharv3221:
Github, generate report for FinalLocalVariable/all-examples-in-one

Issue #11597, opened 2022-05-02
Remove Experimental from XPath Filter


rnveach:
XPath filters have been around since 8.3 in late 2017. We should remove the "experimental" part from them as they have been a normal part for a while.

https://github.com/checkstyle/checkstyle/search?q=experimental

romani:
Yes, but now we have to support both names, as we lost a moment to get out of experimental.

But all in repo should be updated to not mention "experimental"

Shaman20:
Is this issue still valid and can I work on it?

rnveach:
Yes to both.

Shaman20:
I am on it

Shaman20:
It is throwing this error is there any other file which I have to edit?

![image](https://user-images.githubusercontent.com/70961605/168440936-b2bad6d4-6e45-48a2-be25-7c337cc39c7a.png)

romani:
While you are blocked, please take some simple issue meanwhile

romani:
we added new dtd files.
old dtd stays as is, and continue to be working might be forever, or some long time to keep compatibility.

PR #18990, opened 2026-02-22
Issue #11597: Remove Experimental from XPath suppression filter

ayushactiveat:
Issue: #11597
this pr removes the "Experimental" label from all xpath suppression filter dtd references throughout the codebase.

### DIFF
--- a/src/main/resources/com/puppycrawl/tools/checkstyle/suppressions_1_1_xpath.dtd
+++ b/src/main/resources/com/puppycrawl/tools/checkstyle/suppressions_1_1_xpath.dtd
@@ -1,0 +1,21 @@
+<!-- Add the following to any file that is to be validated against this DTD:
+
+<!DOCTYPE suppressions PUBLIC
+    "-//Checkstyle//DTD SuppressionXpathFilter Configuration 1.1//EN"
+    "https://checkstyle.org/dtds/suppressions_1_1_xpath.dtd">
+-->
+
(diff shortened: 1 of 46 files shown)


src/main/resources/com/puppycrawl/tools/checkstyle/suppressions_1_1_xpath.dtd
```diff
@@ -1,19 +1,17 @@
 <!-- Add the following to any file that is to be validated against this DTD:
 
 <!DOCTYPE suppressions PUBLIC
-    "-//Checkstyle//DTD SuppressionXpathFilter Experimental Configuration 1.1//EN"
-    "https://checkstyle.org/dtds/suppressions_1_1_xpath_experimental.dtd">
+    "-//Checkstyle//DTD SuppressionXpathFilter Configuration 1.1//EN"
```
romani, line 4: please keep old file as is, 
we should just add new dtd, to keep compatibility for very long time.

all modern configs will have non experimental version of dtd, but all configs created before March 2026 will continue to have "experiemental" and it should continue to work.

ayushactiveat: sure thing

ayushactiveat:
hi @romani, the old experimental dtd files are now kept as is. and added new non-experimental dtd files. let me know if any other change is needed

Issue #11440, opened 2022-03-19
Conver test from XpathFilterElementTest to SuppressionXpathSingleFilterTest to use inlined config in Input files


romani:
Detected at https://github.com/checkstyle/checkstyle/pull/11437/files

All filters should migrate to use config inlined in Input files.
Suppression that was left in that commit should be removed.

we should not have not real tests, we should cover all functionalilty by full functional tests, to make sure xpath works to match real nodes.

all tests from XpathFilterElementTest should be converted to tests in SuppressionXpathSingleFilterTest where we load whole Input file with Check and Filter in header of file and confirm that violation is suppressed on required node.

Rahulkhinchi03:
@romani do we have to use the same check for `XpathFilterElementTest `  that we used for SuppressionXpathSingleFilterTest?
i.e.
`MagicNumberCheck, MissingJavadocTypeCheck, TypeNameCheck`

romani:
We do not have to use same Check. Any easy to use Check is fine.

Rahulkhinchi03:
can we use `ConstantName `and `MemberName `Checks?

DakshRJain737:
@romani 

Hi, I was looking at converting testNonMatchingFilePattern and testNonMatchingCheckPattern from XpathFilterElementTest to SuppressionXpathSingleFilterTest using the inline config format, but I'm not sure this is actually possible?

These tests specifically use the Pattern constructor variant of XpathFilterElement, passing a compiled Pattern object directly. The inline config format only accepts string values, so there's no way to actually test that code path through it — it would just end up being a duplicate of testNonMatchingFileRegexp and testNonMatchingCheckRegexp.

Should these two tests stay as unit tests in XpathFilterElementTest since they test a different constructor, or is there another approach I'm missing?

romani:
@DakshRJain737, according to git such tests were added to cover pitest mutations.
please comment them out and lets see what pitest violation they cover, so we can think how to kill mutation in execution over Input.

romani:
there are a way more test left to be removed and converted to use verifyXxxxxxx test methods

Carbon14-48:
> there are a way more test left to be removed and converted to use verifyXxxxxxx test methods

[...]

PR #18991, opened 2026-02-22
Issue #11440: Convert test from XpathFilterElementTest to Suppression…

DakshRJain737:
Issue: #11440 

Convert testNonMatchingCheckRegexp from XpathFilterElementTest to SuppressionXpathSingleFilterTest.


romani:
Can we remove some old style test?

DakshRJain737:
> Can we remove some old style test?



> Can we remove some old style test?

removed testNonMatchingCheckRegexp from XpathFilterElementTest.
I noticed that many tests from XpathFilterElementTest already have equivalents in SuppressionXpathSingleFilterTest. As you said they were added to cover pitest mutation. Shall I remove all old tests in this PR or create a new after merging this as old one is removed

romani:
Please create all pure unit tests that are not required to cover something in pitest

romani:
In ideal future, we have no pure unit test methods.
There will be a few and all of them will be for most  far edge case, that are impossible to cover by Inputs.

PR #18489, opened 2026-01-03
Issue #15180: New check UseEnhancedSwitch

octylFractal:
Issue: #15180

New module config: https://gist.githubusercontent.com/octylFractal/c1542f20f46c80b16926553a00f53548/raw/20ef06d3e35c5a9dee33ba83fe55fc362215c972/UseEnhancedSwitchConfig.xml

Contribution PR: https://github.com/checkstyle/contribution/pull/1010

The only thing I'm unsure of is if I should move the now-shared `FallThroughCheck.isTerminated` code elsewhere, I wasn't sure where it should go.

PR #18992, opened 2026-02-22
supplemental: Use if statement instead of switch for #18489

octylFractal:
This is an alternative fix for https://github.com/checkstyle/checkstyle/pull/18489#discussion_r2756536984 instead of opening an issue that is not very actionable.


romani:
Checkstyle violations

octylFractal:
Fixed

PR #18998, opened 2026-02-23
Issue #17882: Update FORMAT_SPECIFIER of JavadocCommentsTokenTypes to…

Hajime-No-Ippo:
Issue: #17882

```
**Input (`src/Test.java`):**

/** 
 * {@value %s #MY_CONSTANT}
 */
public class Test {}


**Command:**
java -jar target/checkstyle-13.3.0-SNAPSHOT-all.jar -j src/Test.java | sed "s/\[[0-9]\+:[0-9]\+\]//g"

JAVADOC_CONTENT -> JAVADOC_CONTENT [0:1]
|--TEXT -> /**  [0:1]
|--NEWLINE -> \n [0:5]
|--LEADING_ASTERISK ->  * [1:1]
|--TEXT ->   [1:3]
|--JAVADOC_INLINE_TAG -> JAVADOC_INLINE_TAG [1:4]
|   `--VALUE_INLINE_TAG -> VALUE_INLINE_TAG [1:4]
|       |--JAVADOC_INLINE_TAG_START -> {@ [1:4]
|       |--TAG_NAME -> value [1:6]
|       |--TEXT ->   [1:11]
|       |--FORMAT_SPECIFIER -> %s [1:12]
|       |--TEXT ->   [1:14]
|       |--REFERENCE -> REFERENCE [1:15]
|       |   |--HASH -> # [1:15]
|       |   `--MEMBER_REFERENCE -> MEMBER_REFERENCE [1:16]
|       |       `--IDENTIFIER -> MY_CONSTANT [1:16]
|       `--JAVADOC_INLINE_TAG_END -> } [1:27]
|--NEWLINE -> \n [1:28]
|--LEADING_ASTERISK ->  * [2:1]
|--TEXT -> / [2:3]
|--NEWLINE -> \n [2:4]
|--TEXT -> public class Test {} [3:1]
`--NEWLINE -> \n [3:21]
```

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
@@ -1663,7 +1663,35 @@ public final class JavadocCommentsTokenTypes {
     public static final int DESCRIPTION = JavadocCommentsLexer.DESCRIPTION;
 
     /**
-     * Format specifier inside Javadoc content.
+     * Format specifier inside a {@code {@value}} inline tag.
+     * The {@value} tag is used to display the value of a constant directly
+     * within the Javadoc documentation. In newer Java versions (20+), there
+     * is ability include a format string inside the tag.
+     *
+     * <p>In this example, the format specifier {@code 0x%04x} is used to format the integer
+     * {@code Modifier#MANDATED} as a hexadecimal value, padded with zeros to a width of four characters.</p>
+     * <pre>{@code
+     * {@value %04x Modifier#MANDATED}
+     * }</pre>
+     *
+     * <p><b>Tree:</b></p>
+     * <pre>{@code
+     * JAVADOC_INLINE_TAG -> JAVADOC_INLINE_TAG
+     * `--VALUE_INLINE_TAG -> VALUE_INLINE_TAG
+     *     |--JAVADOC_INLINE_TAG_START -> {@
+     *     |--TAG_NAME -> value
+     *     |--TEXT ->
+     *     |--FORMAT_SPECIFIER -> %04x
+     *     |--TEXT ->
+     *     |--REFERENCE -> REFERENCE
+     *     |   |--IDENTIFIER -> Modifier
+     *     |   |--HASH -> #
+     *     |   `--MEMBER_REFERENCE -> MEMBER_REFERENCE
+     *     |       `--IDENTIFIER -> MANDATED
+     *     `--JAVADOC_INLINE_TAG_END -> }
+     * }</pre>
+     *
+     * @see #VALUE_INLINE_TAG
      */
     public static final int FORMAT_SPECIFIER = JavadocCommentsLexer.FORMAT_SPECIFIER;
 


src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
```diff
@@ -1663,7 +1663,30 @@ public final class JavadocCommentsTokenTypes {
     public static final int DESCRIPTION = JavadocCommentsLexer.DESCRIPTION;
 
     /**
-     * Format specifier inside Javadoc content.
+     * Format specifier inside a {@code {@value}} inline tag.
+     *
+     * <p><b>Example:</b></p>
+     * <pre>{@code
+     * {@value %s #MY_CONSTANT}
```
romani: this is supper advanced feature that not that people knows.

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
```diff
@@ -1663,7 +1663,30 @@ public final class JavadocCommentsTokenTypes {
     public static final int DESCRIPTION = JavadocCommentsLexer.DESCRIPTION;
 
     /**
-     * Format specifier inside Javadoc content.
+     * Format specifier inside a {@code {@value}} inline tag.
```
romani, line 1666: `Format specifier inside a {@code {@value}} inline tag. The {@value} tag is used to display the value of a constant directly within the Javadoc documentation. In newer Java versions (20+), there is ability include a format string inside the tag.`

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
romani: I found https://www.javaspecialists.eu/archive/Issue308-Formatting-Inlined-value-in-Javadocs.html

lets use this example:
```
    /**
     * The access flag {@code ACC_MANDATED} with a mask value of
     * <code>{@value "0x%04x" Modifier#MANDATED}</code>.
     */
    public static final int MANDATED = 0x8000;
```

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
```diff
@@ -1663,7 +1663,30 @@ public final class JavadocCommentsTokenTypes {
...
+     *
+     * <p><b>Example:</b></p>
```
romani: `In this example, the format specifier 0x%04x is used to format the integer Modifier#MANDATED as a hexadecimal value, padded with zeros to a width of four characters.`

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
Hajime-No-Ippo: I tried using {@value "0x%04x" Modifier#MANDATED} as you suggested, but it throws a parse error in Checkstyle's grammar:

mismatched input '%04x' expecting JAVADOC_INLINE_TAG_END while parsing INLINE_TAG

It seems Checkstyle's FORMAT_SPECIFIER grammar rule only supports the bare unquoted form (%04x), not the quoted string form ("0x%04x").

Should I use the bare form %04x instead, or is this a grammar limitation that needs to be fixed separately?

romani: Please create issue on us that this is not parsed.
Please find some simplified form that we can parse and at the same time it is close to real example , to not block this PR.

romani:
Please make CI green. Single commit always

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
Hajime-No-Ippo: Created issue #19001 to track the grammar limitation with the quoted format specifier.
For this PR, used the bare form {@value %04x Modifier#MANDATED} as the closest parseable alternative.

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
romani: Please use this wording instead of "Example:"

src/main/java/com/puppycrawl/tools/checkstyle/api/JavadocCommentsTokenTypes.java
romani: Few more updates are required for this issue

Issue #18999, opened 2026-02-23
update XdocsExamplesAstConsistencyTest to skip modules with no properties


romani:
from https://github.com/octylFractal/checkstyle/blob/dbd12036cec258a8842e332bfee294f8ade6408d/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java#L318-L319

it is by design that modules without properties should have very different by AST examples.
as threre is no reason to show change in beahvior if config is changing, as config ca not be changed

 upadte XdocsExamplesAstConsistencyTest to skip modules with no properties, they shoudl not be flagged for violation.

Brijeshthummar02:
On it.

romani:
Fix is merged

PR #19008, opened 2026-02-23
Issue #18999: update XdocsExamplesAstConsistencyTest to skip modules with no properties

Brijeshthummar02:
Issue #18999


Brijeshthummar02:
@romani are we good to go?

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -314,8 +325,6 @@ public class XdocsExamplesAstConsistencyTest {
             "filters/suppresswithnearbytextfilter/Example9",
             "filters/suppresswithplaintextcommentfilter/Example5",
             "filters/suppresswithplaintextcommentfilter/Example9",
-            // No properties in module, multiple very different examples to ease reading
-            "checks/annotation/missingoverrideonrecordaccessor/Example2",
```
romani, line 317: I would expect a way more to be removed from this list.
Please do update to print all modules without properties and user to cleanup this suppression list fully

src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsExamplesAstConsistencyTest.java
```diff
@@ -374,6 +386,55 @@ public void testExamplesDifferOnlyByComments() throws IOException {
                 .isEmpty();
     }
 
+    /**
+     * Tests that {@code SUPPRESSED_EXAMPLES} contains no entries for modules without properties.
+     * Modules without properties are skipped entirely by the
+     * {@code isNotModuleWithoutProperties} filter, so adding their examples to
```
romani: I lost a reson of this test. 
Suppression list will be eliminated step by step, and eventually become empty and fully removed, so contributors will not have easy way to avoid this rule on same AST.
Or list will have all items commented on a reason of suppression

Brijeshthummar02: Done. Removed the ```testSuppressedExamplesDoNotIncludeModulesWithoutProperties``` test entirely, Modules without properties are already skipped by the ```isNotModuleWithoutProperties``` filter in ```testExamplesDifferOnlyByComments```, so the separate guard test is not needed.

PR #19015, opened 2026-02-24
Issue #9253: create CI test for limited call stack size and remove TestUtil workarounds

### LINKED ISSUE #9253: Create CI test that execute whole maven test but with limited call stack size
taken from https://github.com/checkstyle/checkstyle/pull/9223#issuecomment-774235104

create special execution on low stack ... will be more clear as it will not be easily reproduce on local by regular "mvn test" command execution.

ayushactiveat:
Fixes #9253

this transitions the limited stack size testing from Java workarounds to a continuous integration workflow, as requested in the issue.

### DIFF
--- a/config/pitest-suppressions/pitest-java-ast-visitor-suppressions.xml
+++ b/config/pitest-suppressions/pitest-java-ast-visitor-suppressions.xml
@@ -1,3 +1,38 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
+  <mutation unstable="false">
+    <sourceFile>JavaAstVisitor.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.JavaAstVisitor</mutatedClass>
+    <mutatedMethod>visitBinOp</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
+    <description>removed call to com/puppycrawl/tools/checkstyle/grammar/java/JavaLanguageParser$BinOpContext::expr</description>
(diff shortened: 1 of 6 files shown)


ayushactiveat:
hi @romani , fail seems unrelated, please let me know if any other change is needed

config/pitest-suppressions/pitest-java-ast-visitor-suppressions.xml
```diff
@@ -1,3 +1,38 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
+  <mutation unstable="false">
```
romani, line 3: we do not allow new mutation, at all.

romani (changes requested):
items:

config/pitest-suppressions/pitest-java-ast-visitor-suppressions.xml
romani: if this is close to impossible, I would rather close issue as WONFIX and explained a reason.

ayushactiveat: @romani removing `TestUtil.getResultWithLimitedResources()` from `JavaAstVisitorTest` caused those 4 `visitBinOp` mutations to live, as that method was the only thing killing them, new ci uses -Xss256k but pitest doesn't, so the mutations live without the java workaround. and as suppressions aren't allowed so creates a catch22 and,  the `TestUtil` workaround cant be removed without losing mutetion . so would you prefer to keep `getResultWithLimitedResources` in `JavaAstVisitorTest` , add only the ci workflow, and close the remove workarounds part as WONTFIX ? please let me know

PR #19018, opened 2026-02-24
Issue #3744: File definition for suppressions should ignore path separator

### LINKED ISSUE #3744: File definition for suppressions should ignore path separator
Currently, if we want the suppression rules to work on both Windows and Linux, we have to use `[\\/]` as the path separator. Our current configuration only works on Linux (https://github.com/hazelcast/hazelcast/blob/master/checkstyle/suppressions.xml). If we replace all path separators with `[\\/]`, the file becomes lot harder to read, have a look at your https://github.com/checkstyle/checkstyle/blob/master/config/suppressions.xml.

I propose, that there could be an option to automatically convert backslashes to forward slashes on Windows before matching the regex:

For example:

The file name `c:\projects\program\com\example\package\impl\Clazz.java` will be matched by rule:

    <suppress checks="JavadocMethod" files="/impl/"/>

I even believe, that most people won't mind if this is done by default.

ayushactiveat:
Resolves #3744

**Description:**
Implemented a fallback mechanism in `SuppressFilterElement.java` to handle Windows path separators. If the initial regex match for a file path fails, it converts backslashes `\` to forward slashes `/` and retries the match.


config/pitest-suppressions/pitest-filters-suppressions.xml
```diff
@@ -1,3 +1,20 @@
...
+  </mutation>
+
+  <mutation unstable="false">
+    <sourceFile>SuppressFilterElement.java</sourceFile>
+    <mutatedClass>com.puppycrawl.tools.checkstyle.filters.SuppressFilterElement</mutatedClass>
+    <mutatedMethod>isFileMatch</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
+    <description>removed conditional - replaced equality check with true</description>
+    <lineContent>if (!match) {</lineContent>
+  </mutation>
```
romani: we do not allow mutation at all, all should be covered by tests, by not pure unit test.

src/test/java/com/puppycrawl/tools/checkstyle/filters/SuppressFilterElementTest.java
```diff
@@ -294,4 +294,19 @@ public void testEqualsAndHashCode() {
                 .isTrue();
     }
 
+    @Test
+    public void testWindowsPathSeparatorFallback() {
```
romani, line 303: please add comment above to explain why we can not use Input files (as path should be in Windows style)

romani (changes requested):
items:

ayushactiveat:
@romani i"ve addressed your instruc. please let me know if anything else is needed for this issue

romani:
GitHub, generate website

romani (approved):
thanks a lot.

extra fixes were done to make it merge-able

Issue #5673, opened 2018-03-31
Pitest: examine and try to remove infinite loop mutations


rnveach:
https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L215-L216
Whenever pitest creates an infinite loop mutation, it takes a whole 50 seconds before it recognizes that the code will never finish. This slows down the pitest run considerably. Pitest can't recognize an infinite loop, and if the mutated code isn't an infinite loop, but takes a long time to finish, it can cause issues with mutation scoring.
I ran into this problem within my own project. An `uncertain` mutation is flagged and counted as a killed. However, if it is the edge of finishing before the timeout and does so in some circumstances, it could be flagged as `survived` and lower the score creating different results in different runs.

https://app.shippable.com/github/rnveach/checkstyle/runs/3/summary/console
Lists all the jobs who have uncertainties as `failed`.

It is possible to remove the uncertain instances by rewriting the code some.
Here is 1 example where I changed it from uncertain to survived: TBD

romani:
It might be better to reduce 50 sec to 5 sec. Checkstyle tests are very quick, even 5 sec is a lot of time.
We can take a look at history of this parameter, it might be in such value from the beginning, then I copied draft config from my non open source project

rnveach:
> We can take a look at history of this parameter, it might be in such value from the beginning

https://github.com/checkstyle/checkstyle/commit/ba178a2bb1aff1799d81742842a9b48ba93c8556#diff-600376dffeb79835ede4a0b285078036
It looks like it has been since the beginning.

romani:
Let's try update to 5 sec first

romani:
Execution in CI is critical for us, I would keep it in 2-3 time bigger than it takes on CI now.
We do not trust execution on local. If one test execution is above 5 sec and system is over used, there might be OOM or, .... .

vivek-0509:
Hi, @romani  I am looking into this issue.

My understanding from the discussion is that simply reducing the timeout to 5 seconds is not the desired fix because it causes flakiness on slower machines. Instead, we should aim to refactor the code to eliminate the infinite loop mutations entirely, converting them into cleaner KILLED  outcomes.

[...]

PR #19024, opened 2026-02-25
Issue #5673: Simplify DeclarationOrderCheck.getAllTokensOfType() traversal to eliminate infinite loop mutations

MiniPiku:
## Summary
Refactors `getAllTokensOfType` method from a nested-loop traversal to a simpler stack-based iteration, eliminating 7 out of 9 PIT mutations that previously caused infinite loops.

Refs checkstyle#5673

## Problem
The original implementation was vulnerable to multiple PIT mutations that caused `TIMED_OUT` and `MEMORY_ERROR` failures:

### Before (9 surviving mutants):
1.  `removed conditional - replaced equality check with true` → TIMED_OUT
2.  `removed call to java/util/Deque::isEmpty` → TIMED_OUT  
3.  `removed conditional - replaced equality check with false` → TIMED_OUT
4.  `removed conditional - replaced equality check with false` → TIMED_OUT
5.  `removed call to java/util/Deque::pop` → TIMED_OUT
6.  `negated conditional` → TIMED_OUT
7.  `removed conditional - replaced equality check with false` → TIMED_OUT
8.  `replaced call to getNextSibling with receiver` → TIMED_OUT
9.  `replaced call to getFirstChild with receiver` → MEMORY_ERROR

**Root cause:** The nested `while` loops with complex state management created multiple code paths where mutations could introduce infinite loops.

## Solution
Replaced the nested-loop pattern with a single-loop stack-based traversal:

### Key Changes:
```java
// Before: Complex nested loops with mutable vertex
DetailAST vertex = ast;
while (vertex != null || !stack.isEmpty()) {
    if (!stack.isEmpty()) {
        vertex = stack.pop();
    }
    while (vertex != null) {  // Nested loop - mutation vulnerable
        // ... process vertex
        vertex = vertex.getFirstChild();  // Mutable state
    }
}

// After: Simple single loop
stack.push(ast);
while (!stack.isEmpty()) {  // Single termination condition
    final DetailAST current = stack.pop();  // Immutable per iteration
    // ... process current
    final DetailAST child = current.getFirstChild();  // Extract to variable
    if (child != null) {
        stack.push(child);
    }
}
```

### Specific Improvements:
1. **Single loop condition** (`!stack.isEmpty()`) instead of compound condition
2. **Immutable local variables** (`final DetailAST current`) per iteration
3. **Extracted method calls** to local variables before null checks
4. **Explicit null checks** before pushing to stack
5. **Eliminated mutable state** (no reassigning `vertex` multiple times)

### After (2 remaining mutants):
1.  ~~`removed conditional - replaced equality check with true`~~ → **KILLED**
2.  ~~`removed call to java/util/Deque::isEmpty`~~ → **KILLED**
3.  ~~`removed conditional - replaced equality check with false`~~ → **KILLED**
4.  ~~`removed conditional - replaced equality check with false`~~ → **KILLED**
5.  ~~`removed call to java/util/Deque::pop`~~ → **KILLED**
6.  ~~`negated conditional`~~ → **KILLED**
7.  ~~`removed conditional - replaced equality check with false`~~ → **KILLED**
8.  `replaced call to getNextSibling with receiver` → TIMED_OUT (edge case)
9.  `replaced call to getFirstChild with receiver` → MEMORY_ERROR (edge case)

**Mutation coverage improved: 0/9 (0%) → 7/9 (77.8%)**

### Remaining Mutants Analysis:
The 2 surviving mutants represent extreme edge cases where AST nodes form circular references:
- **Mutant #8**: If `getNextSibling()` returns `self` instead of sibling, creates self-loop
- **Mutant #9**: If `getFirstChild()` returns `self` instead of child, creates self-loop

These scenarios cannot occur in valid Checkstyle AST structures, as they would violate the tree invariant. Adding defensive code for these cases would pessimize the common path for hypothetical impossible scenarios.

---

**Issue:** #5673  
**Related:** Part of ongoing effort to improve PIT mutation test coverage across the codebase.

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/DeclarationOrderCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/DeclarationOrderCheck.java
@@ -308,21 +308,25 @@ public class DeclarationOrderCheck extends AbstractCheck {
      * @return a set of all tokens of specific type starting with the current ast node.
      */
     private static Set<DetailAST> getAllTokensOfType(DetailAST ast, int tokenType) {
-        DetailAST vertex = ast;
-        final Set<DetailAST> result = new HashSet<>();
         final Deque<DetailAST> stack = new ArrayDeque<>();
-        while (vertex != null || !stack.isEmpty()) {
-            if (!stack.isEmpty()) {
-                vertex = stack.pop();
+        stack.push(ast);
+
+        final Set<DetailAST> result = new HashSet<>();
+
+        while (!stack.isEmpty()) {
+            final DetailAST current = stack.pop();
+            if (current.getType() == tokenType && !current.equals(ast)) {
+                result.add(current);
             }
-            while (vertex != null) {
-                if (vertex.getType() == tokenType && !vertex.equals(ast)) {
-                    result.add(vertex);
-                }
-                if (vertex.getNextSibling() != null) {
-                    stack.push(vertex.getNextSibling());
-                }
-                vertex = vertex.getFirstChild();
+
+            final DetailAST sibling = current.getNextSibling();
+            if (sibling != null) {
+                stack.push(sibling);
+            }
+
+            final DetailAST child = current.getFirstChild();
+            if (child != null) {
+                stack.push(child);
             }
         }
         return result;


romani:
Github, generate report for DeclarationOrder/all-examples-in-one

PR #18459, opened 2025-12-30
Issue #17565: Fix LeftCurlyCheck to correctly validate line break after '{'

MonuChaudhary14:
Issue #17565 

In this PR, I updated the LeftCurlyCheck logic and its corresponding tests to correctly validate line-break behavior 

Diff Regression config: https://gist.githubusercontent.com/MonuChaudhary14/0f6c4c3e4845e24ed9b75552b764f5d1/raw/0c26c30c16bdcf876e51ecdcab07b4c9b13b5fdf/google_checks.xml

Diff Regression patch config: https://gist.githubusercontent.com/MonuChaudhary14/db1380d15620df0e3fab53887651da22/raw/848c6ef0ac27539a0577f8f5b98e0c232208db95/patch_config.xml

PR #19043, opened 2026-02-26
Issue #17842: Activate GoogleNonConstantFieldName in google_checks.xml

vivek-0509:
Issue #17842:

Replace MemberName with GoogleNonConstantFieldName in google_checks.xml to fix false-negatives for member names with underscores between letters and digits.

Diff Regression config: https://gist.githubusercontent.com/vivek-0509/94f9fb7d62cbbd713adc3d598c72ce4a/raw/5aadc7d27f7d60abc70d61570565695e512ed0ee/google_checks_base.xml

Diff Regression patch config: https://gist.githubusercontent.com/vivek-0509/4a80f1a7b830fab98b8e10ba020c3977/raw/58c81726f38afc71fc6161948d8469f87b3f77b2/patch_config.xml

### DIFF
--- a/src/it/resources/com/google/checkstyle/test/chapter5naming/rule525nonconstantfieldnames/InputNonConstantNamesBasic.java
+++ b/src/it/resources/com/google/checkstyle/test/chapter5naming/rule525nonconstantfieldnames/InputNonConstantNamesBasic.java
@@ -2,28 +2,37 @@ package com.google.checkstyle.test.chapter5naming.rule525nonconstantfieldnames;
 
 /** Some javadoc. */
 public class InputNonConstantNamesBasic {
-  public int mPublic; // violation 'Member name 'mPublic' must match pattern'
-  protected int mProtected; // violation 'Member name 'mProtected' must match pattern'
-  int mPackage; // violation 'Member name 'mPackage' must match pattern'
-  private int mPrivate; // violation 'Member name 'mPrivate' must match pattern'
+  // violation below, ''mPublic' must .* avoid single lowercase letter followed by uppercase'
+  public int mPublic;
+  // violation below, ''mProtected' must .* avoid single lowercase letter followed by uppercase'
+  protected int mProtected;
+  // violation below, ''mPackage' must .* avoid single lowercase letter followed by uppercase'
+  int mPackage;
+  // violation below, ''mPrivate' must .* avoid single lowercase letter followed by uppercase'
+  private int mPrivate;
 
-  public int _public; // violation 'Member name '_public' must match pattern'
-  protected int prot_ected; // violation 'Member name 'prot_ected' must match pattern'
-  int package_; // violation 'Member name 'package_' must match pattern'
-  private int priva$te; // violation 'Member name 'priva\$te' must match pattern'
+  public int _public; // violation ''_public' .* underscores allowed only between adjacent digits.'
+  // violation below, ''prot_ected' .* underscores allowed only between adjacent digits.'
+  protected int prot_ected;
+  int package_; // violation ''package_' .* underscores allowed only between adjacent digits.'
+  // violation below, ''priva\$te' must .* contain only letters, digits or underscores'
+  private int priva$te;
 
   public int ppublic;
   protected int pprotected;
   int ppackage;
   private int pprivate;
 
-  int ABC = 0;
-  // 2 violations above:
+  // 2 violations 3 lines below:
   //  'Abbreviation in name 'ABC' must contain no more than '1' consecutive capital letters.'
-  //  'Member name 'ABC' must match pattern'
-  final int C_D_E = 0; // violation 'Member name 'C_D_E' must match pattern'
+  //  'Non-constant field name 'ABC' must start lowercase, be at least 2 chars'
+  int ABC = 0;
+  final int C_D_E = 0; // violation ''C_D_E' .* underscores allowed only between adjacent digits.'
 
-  public int $mPublic; // violation 'Member name '\$mPublic' must match pattern'
-  protected int mPro$tected; // violation 'Member name 'mPro\$tected' must match pattern'
-  int mPackage$; // violation 'Member name 'mPackage\$' must match pattern'
+  // violation below, ''\$mPublic' must .* contain only letters, digits or underscores'
+  public int $mPublic;
+  // violation below, ''mPro\$tected' must .* contain only letters, digits or underscores'
+  protected int mPro$tected;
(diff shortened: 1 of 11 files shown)


vivek-0509:
CI Failures are unrelated

vivek-0509:
GitHub, generate website

romani:
@Zopsss , please help to review.

Zopsss:
@vivek-0509 please generate diff report

vivek-0509:
GitHub, generate report

vivek-0509:
@Zopsss the diff report is same as what you reviewed previously https://github.com/checkstyle/checkstyle/pull/18657#issuecomment-3962143997
the only difference is in the checkstyle violations because new code is added into checkstyle

Zopsss:
@vivek-0509 can you generate the report with our current google_checks.xml and the new google_checks.xml? That way we can see if there is any difference by adding the new check.

vivek-0509:
GitHub, generate report

Zopsss:
https://github.com/checkstyle/checkstyle/actions/runs/22553094943/job/65326166130#step:10:318


getting the same "NoClassDefFoundError" exception in for every project. 


At last getting the "OutOfMemoryError": https://github.com/checkstyle/checkstyle/actions/runs/22553094943/job/65326166130#step:10:4346

@vivek-0509 can you dig into this?


---


The changes looks good, no new violation has been introduced neither any old violation was compromised. But just to be on safe side, we should generate the diff report atleast once. 

If the above error is not possible to solve then its also fine. We can use the old report.

vivek-0509:
Hi @Zopsss,
we see two errors, i have done some investigation around them

`Error 1: NoClassDefFoundError: org/apache/maven/doxia/sink/SinkEventAttributeSet`

This is a version incompatibility in the contribution repo's checkstyle-tester/pom.xml which uses maven-checkstyle-plugin:3.1.1.
the reason is `SinkEventAttributeSet` was relocated from `package org.apache.maven.doxia.sink` to `org.apache.maven.doxia.sink.impl` between Doxia 1.6 and 1.7 ([doxia-1.6 source](https://github.com/apache/maven-doxia/tree/doxia-1.6/doxia-core/src/main/java/org/apache/maven/doxia/sink) vs [doxia-1.7 source](https://github.com/apache/maven-doxia/tree/doxia-1.7/doxia-core/src/main/java/org/apache/maven/doxia/sink)).

maven-checkstyle-plugin:3.1.1 was compiled against Doxia 1.4 and imports from the old package
`import org.apache.maven.doxia.sink.SinkEventAttributeSet; `
 
At runtime, the checkstyle transitively brings in Doxia 1.12.0 , which overrides the plugin's declared Doxia 1.4. In Doxia 1.12.0, the class was relocated to `org.apache.maven.doxia.sink.impl.SinkEventAttributeSet,` so the old fully-qualified name no longer resolves.

We need to upgrade maven-checkstyle-plugin from 3.1.1 to 3.6.0 in the contribution repo's checkstyle-tester/pom.xml. Version 3.6.0 is built against Doxia 2.0.0 and uses the correct package. 

`Error 2: OutOfMemoryError: UTF16 String size is 1074003966, should be less than 1073741823`

In diff.groovy, `postProcessCheckstyleReport()` reads the entire checkstyle-result.xml into a single Java String
there  is a hard JVM String size limit introduced in JDK 9 ([JDK-8190429](https://bugs.openjdk.org/browse/JDK-8190429)). The checkstyle-result.xml for spring-integration with the full google_checks.xml which has a lot of checks, exceeds this limit because these checks produces a lot of violations.

i think we need to replace the whole file read with line-by-line streaming

Zopsss:
Thanks a lot for the detailed explanation. Can you open an issue about this:

> We need to upgrade maven-checkstyle-plugin from 3.1.1 to 3.6.0 in the contribution repo's checkstyle-tester/pom.xml. Version 3.6.0 is built against Doxia 2.0.0 and uses the correct package.

In contribution repo? 

@romani cc

---

> The checkstyle-result.xml for spring-integration with the full google_checks.xml which has a lot of checks, exceeds this limit because these checks produces a lot of violations.

I think the problem is not with spring-integration, but it is with openjdk25. 

We successfully ran CS on spring integration: https://github.com/checkstyle/checkstyle/actions/runs/22553094943/job/65326166130#step:10:3965

But the Openjdk is causing the issues here, we had faced similar out of memory error for openjdk in the past as well. Please remove it from the projects list, and try to generate the diff again.

https://github.com/checkstyle/checkstyle/actions/runs/22553094943/job/65326166130#step:10:4204

vivek-0509:
@Zopsss My bad. It’s for OpenJDK only, but the same error can occur in large projects which causes a large number of violations and the other error we see is for all projects including spring integration.
Sure will open the issue for this and i think we can solve this out of memory error also as i mentioned in the message above, should i open issue for that also ??

Zopsss:
Sure it if we're able to solve out of memory error then we should definitely do it.

vivek-0509:
GitHub, generate report

vivek-0509:
GitHub, generate report

vivek-0509:
GitHub, generate report

vivek-0509:
@Zopsss Here is the new regression report
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/1fd2fa5_2026075708/reports/diff/index.html

vivek-0509:
@romani @Zopsss I have opened the issues please take a look and approve them https://github.com/checkstyle/contribution/issues/1038 
https://github.com/checkstyle/contribution/issues/1039

Zopsss:
@vivek-0509 SWITCH_RULE token was removed from LeftCurly check. Was this done by you?

image


image

https://www.diffchecker.com/NKzOHsiu/

vivek-0509:
@Zopsss https://github.com/checkstyle/checkstyle/pull/18459 this PR recently got merged which added the SWITCH_RULE token that'ss why my base config has this SWITCH_RULE token, my patch config missed it, i think i copied before rebasing that why, we have this mismatch

vivek-0509:
GitHub, generate report

vivek-0509:
@Zopsss Here is the new diff report  https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/1fd2fa5_2026055304/reports/diff/index.html

Zopsss (approved):
Thanks a lott!!!!

romani:
Thanks a lot for hard work, I hope you see that delivering with quality it not easy task

vivek-0509:
GitHub, generate report

PR #19051, opened 2026-02-27
Issue #16360: Added comment to testFileOpenTag

Adam-Gamil:
Issue #16360

The `testFileOpenTag` method in `XMLLoggerTest` cannot be updated to use the standard `verifyWithInlineConfigParserAndXmlLogger` approach.

Reason: The test requires creating an `AuditEvent` with a special character in the file name (`Test&.java`) to verify XML escaping. This cannot be represented by a standard input Java file without introducing physical files with special characters into the repository, which can cause cross-platform file system issues.

A comment has been added above the method to explain why this test deviates from the usual input/expected XML pattern.

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/XMLLoggerTest.java
@@ -344,6 +344,12 @@ public class XMLLoggerTest extends AbstractXmlTestSupport {
         verifyWithInlineConfigParserAndXmlLogger(inputFile, expectedXmlReport);
     }
 
+    /**
+     * Cannot use verifyWithInlineConfigParserAndXmlLogger because this test
+     * requires a custom AuditEvent with a special character in the file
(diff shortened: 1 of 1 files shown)


Adam-Gamil:
Hi @romani,

I noticed the `verify-no-exception-configs` CI check is failing with this diff:

-`GoogleNonConstantFieldName`

It looks like `GoogleNonConstantFieldName` needs to be added to the [contribution](https://github.com/checkstyle/contribution) repository in either `checks-nonjavadoc-error.xml` or `checks-only-javadoc-error.xml`.

What would you suggest as the best way to handle this? Should I go ahead and create a PR in the contribution repo to add `GoogleNonConstantFieldName` to one of those files? If so, which file should it go in?

vivek-0509:
@Adam-Gamil ignore it for now Pr is up for the fix, it will soon be resolved

Adam-Gamil:
@romani Gentle reminder for a review, let me know if there is anything i should update

Issue #19001, opened 2026-02-23
JavadocCommentsParser: quoted format specifier in {@value} tag is not parsed


Hajime-No-Ippo:
https://www.javaspecialists.eu/archive/Issue308-Formatting-Inlined-value-in-Javadocs.html described this javadoc new syntax

## Checkstyle version
13.3.0-SNAPSHOT

## Description
Since Java 20, the `{@value}` inline tag supports a quoted format string to control 
how a constant value is displayed. For example, from `java.lang.reflect.AccessFlag`:

```java
/**
 * The access flag {@code ACC_MANDATED} with a mask value of
 * <code>{@value "0x%04x" Modifier#MANDATED}</code>.
 */
public static final int MANDATED = 0x8000;
```

Running Checkstyle's Javadoc AST printer on this input throws a parse error:

```
java -jar checkstyle-13.3.0-SNAPSHOT-all.jar -j src/Test.java

Exception in thread "main" java.lang.IllegalArgumentException:
[ERROR:4] Javadoc comment at column 55 has parse error. Details:
mismatched input '%04x' expecting JAVADOC_INLINE_TAG_END while parsing INLINE_TAG
```

The current lexer rule in `JavadocCommentsLexer.g4` only supports bare unquoted format specifiers:

```
FORMAT_SPECIFIER
    : '%' [#+\- 0,(]* [0-9]* ('.' [0-9]+)? [a-zA-Z]
    ;
```

The grammar needs to be updated to support the quoted form (`"0x%04x"`) introduced in Java 20.

## Related
Part of #17882

romani:
@Hajime-No-Ippo, please use markdown syntax to make issue looks good

romani:
@Hajime-No-Ippo , please share whole compile able java file.
Please confirm that javadoc tool also processed such file without errors 

We need to know is this should be quoted always , or it is optional form.

Hajime-No-Ippo:
@romani , 

Here's the compile able Test.java to this issue

**Quoted Form:**

```java
public class Test {
    /**
     * The access flag with a mask value of {@value "%04x" #MANDATED}.
     */
    public static final int MANDATED = 0x8000;
}
```

**Unquoted form:**

```java
public class Test {
    /**
     * The access flag with a mask value of {@value %04x #MANDATED}.
     */
    public static final int MANDATED = 0x8000;
}
```

[...]

PR #19053, opened 2026-02-27
Issue #19001: Support quoted format specifier in {@value} tag

pirzada-ahmadfaraz:
Issue: #19001

## Description

Since Java 20, the `{@value}` inline Javadoc tag supports a quoted format string
to control how a constant value is displayed. For example:

```java
/** {@value "%04x" #MANDATED} */
/** {@value "0x%04x" Modifier#MANDATED} */
```

Checkstyle's Javadoc parser currently throws a parse error on the quoted form because the lexer's `VALUE_MODE` has no rule to match double-quoted strings.

## Changes

- Added `Value_STRING_LITERAL` lexer rule in `VALUE_MODE`
  (`JavadocCommentsLexer.g4`) to match quoted strings
- Updated `valueInlineTag` parser rule (`JavadocCommentsParser.g4`)
  to accept either `FORMAT_SPECIFIER` or `STRING_LITERAL`
- Added test cases for quoted format specifiers with member and qualified references


romani:
@mahfouz72 , please review first.

pirzada-ahmadfaraz:
@romani @mahfouz72 ping

mohitsatr:
@pirzada-ahmadfaraz, Thanks for your help. You need to generate regression report locally
Please read https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#testing-antlr-grammar-changes

Once the report is generated, you need to deploy it so we can review it. Please read: https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/README_MANUAL_EXECUTION.md#deploying-report

Let me know if you need any help with the steps

src/main/resources/com/puppycrawl/tools/checkstyle/grammar/javadoc/JavadocCommentsLexer.g4
```diff
@@ -597,6 +597,10 @@ FORMAT_SPECIFIER
     : '%' [#+\- 0,(]* [0-9]* ('.' [0-9]+)? [a-zA-Z]
     ;
 
+Value_STRING_LITERAL
+    : '"' ~["\r\n]* '"' -> type(STRING_LITERAL)
+    ;
```
mahfouz72: According to the [official Javadoc specification](https://docs.oracle.com/en/java/javase/22/docs/specs/javadoc/doc-comment-spec.html#:~:text=If%20the%20format%20string%20is%20given%2C%20it%20must%20either%20begin%20with%20a%20percent%20character%20(%25)%20or%20be%20enclosed%20in%20double%2Dquote%20characters%20(%22)) for {@value}, the format argument may either begin with `%`, or be enclosed in double quotes.


Because of this, the quoted form is not just an arbitrary string literal but a format string that contains a single `%` conversion.

So it may be better to support it in the existing `FORMAT_SPECIFIER` token, but just update it to allow quotes

mahfouz72 (changes requested):
We need to generate regression report also.

romani:
Energy is moved to https://github.com/checkstyle/checkstyle/pull/20394

Issue #11822, opened 2022-07-03
RequireThisCheck giving multiple violation for classes nested in lambdas.


Vyom-Yadav:
I have read check documentation: https://checkstyle.sourceforge.io/config_coding.html#RequireThis
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

```bash
/var/tmp $ javac Test.java
```
```bash
/var/tmp $ cat config.xml
```
```XML
<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
        "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">

<module name="Checker">
  <module name="TreeWalker">
    <module name="RequireThisCheck">
      <property name="validateOnlyOverlapping" value="false" />
    </module>
  </module>
</module>
```
```bash
/var/tmp $ cat Test.java
```
```java
import java.util.function.Consumer;

public interface Test {

    private static void method(Consumer<String> consumer) {
        consumer.accept("foo");
    }

    private void testCtorNestedInLambda() {
        method(s -> {
            class InnerSubscriber implements Test {
                int index;

                public InnerSubscriber(int index) {
                    index = index; // 2 violations, expected 1
                }
            }
        });
    }
}
```
```bash
/var/tmp $ RUN_LOCALE="-Duser.language=en -Duser.country=US"
/var/tmp $ java $RUN_LOCALE -jar checkstyle-10.3.1-all.jar -c config.xml Test.java
Starting audit...
[ERROR] /var/tmp/Test.java:15:21: Reference to instance variable 'index' needs "this.". [RequireThis]
[ERROR] /var/tmp/Test.java:15:29: Reference to instance variable 'index' needs "this.". [RequireThis]
Audit done.
Checkstyle ends with 2 errors.
```
It is giving violation for both sides, the check is trying to imply `this.index = this.index` which isn't correct.

Kevin222004:
I was trying to figure out this issue for the past 2 days still. I am not able to figure out the whole code. I have got only basic idea after giving lots of time. is their anything special (any algorithm or any concept) in this check which I should learn to make it easy

rnveach:
2 violations started appearing in 8.27 , but only 1 appeared in 8.26 . Maybe we changed something and broke something?

````
$ cat TestClass.java
import java.util.function.Consumer;

[...]

PR #19061, opened 2026-02-28
Issue #11822: Fix RequireThisCheck false positive for type definition…

vivek-0509:
Issue #11822:
Fix false positive in RequireThisCheck for type definitions (class, interface, enum, record, anonymous class) nested inside lambdas

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
@@ -1009,33 +1009,62 @@ public class RequireThisCheck extends AbstractCheck {
      * @return true if the token is a Lambda parameter
      */
     private static boolean isLambdaParameter(DetailAST ast) {
-        DetailAST parent;
-        for (parent = ast; parent != null; parent = parent.getParent()) {
+        boolean result = false;
+        for (DetailAST parent = ast; parent != null; parent = parent.getParent()) {
             if (parent.getType() == TokenTypes.LAMBDA) {
+                result = !isInsideTypeDefInsideLambda(ast)
+                        && isMatchingLambdaParam(ast, parent);
                 break;
             }
         }
-        final boolean isLambdaParameter;
-        if (parent == null) {
-            isLambdaParameter = false;
-        }
-        else if (ast.getType() == TokenTypes.PARAMETER_DEF) {
-            isLambdaParameter = true;
+        return result;
+    }
+
+    /**
+     * Checks whether the given AST matches a parameter of the specified lambda.
+     *
+     * @param ast the AST node to check.
+     * @param lambda the lambda expression AST node.
+     * @return true if the AST matches a lambda parameter.
+     */
+    private static boolean isMatchingLambdaParam(DetailAST ast, DetailAST lambda) {
+        final boolean isMatchingParam;
+        if (ast.getType() == TokenTypes.PARAMETER_DEF) {
+            isMatchingParam = true;
         }
         else {
-            final DetailAST lambdaParameters = parent.findFirstToken(TokenTypes.PARAMETERS);
+            final DetailAST lambdaParameters = lambda.findFirstToken(TokenTypes.PARAMETERS);
             if (lambdaParameters == null) {
-                isLambdaParameter = parent.getFirstChild().getText().equals(ast.getText());
+                isMatchingParam = lambda.getFirstChild().getText().equals(ast.getText());
             }
             else {
-                isLambdaParameter = TokenUtil.findFirstTokenByPredicate(lambdaParameters,
+                isMatchingParam = TokenUtil.findFirstTokenByPredicate(lambdaParameters,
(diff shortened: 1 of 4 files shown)


vivek-0509:
Github, generate report for RequireThis/all-examples-in-one

vivek-0509:
Github, generate report for RequireThis/all-examples-in-one

vivek-0509:
@romani PR is ready for review, issue is resolved we see no changes in the regression report

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/requirethis/InputRequireThisLocalClassesInsideLambdas.java
```diff
@@ -28,4 +28,16 @@ public InnerSubscriber(int index) {
...
+
+    private static void testFieldRefInCtorInsideStaticLambda() {
+        method(s -> {
+            class Inner implements InputRequireThisLocalClassesInsideLambdas {
+                int index;
+
+                Inner() {
+                    int x = index;
+                }
+            }
```
vivek-0509, line 40: While investigating this fix, i discovered a pre-existing false negative in RequireThisCheck. When a local class is defined inside a static method, field references inside the constructor of that class don't get violations even though this.field is valid there (i have added this case ). This is unrelated to the current issue and will be filed as a separate issue.

vivek-0509:

src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.java
```diff
@@ -1009,33 +1009,62 @@ private String getNearestClassFrameName() {
      * @return true if the token is a Lambda parameter
      */
     private static boolean isLambdaParameter(DetailAST ast) {
-        DetailAST parent;
-        for (parent = ast; parent != null; parent = parent.getParent()) {
+        boolean result = false;
+        for (DetailAST parent = ast; parent != null; parent = parent.getParent()) {
             if (parent.getType() == TokenTypes.LAMBDA) {
+                result = !isInsideTypeDefInsideLambda(ast)
+                        && isMatchingLambdaParam(ast, parent);
```
mahfouz72, line 1016: Question: 

In the issue, the `index` variable is not a lambda parameter, it is a constructor parameter of a class that happens to be defined inside a lambda body.

However, the fix introduces logic that checks whether a variable is a lambda parameter.

Since `index` is unrelated to the lambda parameters, I am confused about how this change helps resolve the issue.

vivek-0509: the root cause is in the original isLambdaParameter which had this shortcut

```
else if (ast.getType() == TokenTypes.PARAMETER_DEF) {
    isLambdaParameter = true;
}
```
It assumed that any PARAMETER_DEF with a LAMBDA ancestor must be a lambda parameter. This breaks when a class is defined inside a lambda, the constructor's PARAMETER_DEF also has a LAMBDA ancestor above it, but there is an OBJBLOCK (class body) in between that was never checked. The presence of the lambda is what triggered the bug, if the class was inside a regular method with no lambd, `isLambdaParameter `would return false correctly and there would be no issue.

This incorrect true had a cascading effect because isLambdaParameter is also called during frame building

```
if (!CheckUtil.isReceiverParameter(ast) && !isLambdaParameter(ast)) {
    frame.addIdent(parameterIdent);
}

```
Since isLambdaParameter incorrectly returned true for the constructor parameter index, it was never registered in CTOR_FRAME. So when index = index was processed in the constructor body, findFrame searched outward, found nothing in CTOR_FRAME, and resolved both LHS and RHS index as field references from CLASS_FRAME, producing 2 violations instead of 1.

we fix this by adding `isInsideTypeDefInsideLambda` which walks up from the node toward the LAMBDA and checks if it crosses an OBJBLOCK on the way. If it does, the node is inside a type definition nested within the lambda, not a lambda parameter. This guard runs before the rest of the lambda parameter logic

`result = !isInsideTypeDefInsideLambda(ast) && isMatchingLambdaParam(ast, parent);`
Now the constructor PARAMETER_DEF index correctly returns false from `isLambdaParameter`, gets registered in CTOR_FRAME, and findFrame resolves it as a constructor parameter, only 1 violation.

i think the confusion is because of `isMatchingLambdaParam ` but it is not new logic, it is the existing lambda parameter matching code from the original isLambdaParameter, simply extracted into a separate method for readability. the only new logic introduced is` isInsideTypeDefInsideLambda.`

vivek-0509:
@romani  ping

vivek-0509:
@romani  ping

vivek-0509:
@romani  ping

vivek-0509:
@romani @stoyanK7  ping

vivek-0509:
@romani  ping, CI failure is unrelated

romani:
Github, generate report for RequireThis/all-examples-in-one

romani (approved):
Thanks a lot

PR #19062, opened 2026-02-28
Issue #18614: Fix tab-based indentation validation in getIndentImpl

### LINKED ISSUE #18614: False positive: indentation inside of constructor parameters inside try block

```cmd
E:\temp\CheckstyleReport>type config.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
<module name="Checker">

  <module name="TreeWalker">
    <module name="Indentation" />
  </module>
</module>

E:\temp\CheckstyleReport>type Test.java
public class Test {
        private Test client;

        private Test(String string) {
        }

        private void test() {
                try {
                        client = new Test(
                                null // unexpected violation about indentation
                        );
                } catch (Exception e) {
                }
        }
}
E:\temp\CheckstyleReport>set RUN_LOCALE="-Duser.language=en -Duser.country=US"

E:\temp\CheckstyleReport>java %RUN_LOCALE% -jar checkstyle-13.0.0-all.jar -c config.xml Test.java
Starting audit...
[ERROR] E:\temp\CheckstyleReport\Test.java:10:17: 'new' child has incorrect indentation level 16, expected level should be 20. [Indentation]
Audit done.
```

---

The violation shouldn't be reported: the parameter passed to the constructor is already indented correctly.

NOTE: if we remove the try/catch block the violation disappears.
NOTE2: the file is formatted by tabs

---

MiniPiku:
###  Description

Fixes #18614 - incorrect indentation validation for tab-indented files in `IndentationCheck.getIndentImpl()`

**Root Cause:**
The previous implementation compared:

`getLineStart(mainAst)` → visual column (tab-expanded)

`mainAst.getColumnNo()` → raw character column (tab counts as 1)

For tab-indented files, these values never matched, causing the branch responsible for computing expected indentation to become unreachable. As a result, incorrect indentation for the `new` keyword inside a `try` block was not reported (false negative).

**Fix:**
Replaced the inconsistent visual vs raw column comparison with `isOnStartOfLine(mainAst)`, which uses consistent visual column logic and properly handles tab expansion.

This ensures:
Correct tab-indented code is validated properly
Incorrect tab indentation now reports violations
Space-indented behavior remains unchanged

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/NewHandler.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/NewHandler.java
@@ -121,7 +121,7 @@ public class NewHandler extends AbstractExpressionHandler {
         IndentLevel result;
         // if our expression isn't first on the line, just use the start
         // of the line
-        if (getLineStart(mainAst) == mainAst.getColumnNo()) {
+        if (isNewOnStartOfLine()) {
             result = super.getIndentImpl();
 
             final boolean isLineWrappedNew = TokenUtil.isOfType(mainAst.getParent().getParent(),
@@ -144,6 +144,17 @@ public class NewHandler extends AbstractExpressionHandler {
         }
 
         return result;
+    }
+
+    /**
+     * Checks if the new keyword is at the start of its line,
+     * using consistent visual column logic to handle both
+     * space and tab indented files.
+     *
+     * @return true if new is the first token on the line
+     */
+    private boolean isNewOnStartOfLine() {
+        return getLineStart(mainAst) == expandedTabsColumnNo(mainAst);
     }
 
     /**
--- a/src/test/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheckTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheckTest.java
@@ -4286,6 +4286,50 @@ public class IndentationCheckTest extends AbstractModuleTestSupport {
         verifyWarns(checkConfig, fileName, expected);
     }
 
+    @Test
+    public void testTabIndentationNewInTryBlockValid() throws Exception {
+        final DefaultConfiguration checkConfig = createModuleConfig(IndentationCheck.class);
+        checkConfig.addProperty("basicOffset", "4");
+        checkConfig.addProperty("braceAdjustment", "0");
+        checkConfig.addProperty("caseIndent", "4");
+        checkConfig.addProperty("lineWrappingIndentation", "4");
+        checkConfig.addProperty("tabWidth", "4");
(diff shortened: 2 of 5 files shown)


romani (changes requested):
we need tests extension to prove that it fixing issue

MiniPiku:
Thank you for the feedback.

The required test extension is currently blocked by issue #19063, which is directly related to this problem. At the moment, adding new regression inputs causes `IndentationTrailingCommentsVerticalAlignmentTest` to fail due to the vertical alignment enforcement.

Because of this coupling, we cannot properly add or validate the necessary tests until #19063 is resolved.

Could you please review #19063 first? Once that issue is addressed, I will immediately proceed with adding the test extensions to fully cover this case.

Let me know if you would prefer a different approach.

romani:
There must be test method with Input that is copy from issue description.
You need to prove that fix is working

MiniPiku:
> There must be test method with Input that is copy from issue description. You need to prove that fix is working

Before adding tests we need to resolve #19063 or else the tabs has to be resolved to 4/8 spaces and it defeats the whole purpose of the test.

MiniPiku:
Hi @romani,
I’ve added the requested tests. This PR should now be ready for review.

romani:

MiniPiku:
@romani is this good to merge or any changes are required?

romani:
Github, generate report for Indentation/Example1

romani:
We need report to see diff on big scale of projects 
https://github.com/checkstyle/test-configs/tree/main/Indentation

romani:
Please review diff report

MiniPiku:
Github, generate report for Indentation/Example1

romani:
Why I don't see exclude of file with tab, tab is forbidden, so there should be suppression of checkstyle violation. Do you know why ?

MiniPiku:
> Why I don't see exclude of file with tab, tab is forbidden, so there should be suppression of checkstyle violation. Do you know why ?

@romani

Tabs are not suppressed here because IndentationCheck supports tabs via the `tabWidth` property and still processes such files.

The rule that forbids tabs is `FileTabCharacter`, which is not part of this configuration. Because of that, tab-based files are not excluded and are still validated by IndentationCheck.

With this fix, these files are now handled correctly instead of being skipped due to the mismatch between raw and visual column calculations.

Please let me know if suppression is expected for these configs.

MiniPiku:
Github, generate report for Indentation/Example1

MiniPiku:
Github, generate report for Indentation/Example1

romani:
@MiniPiku, sorry for long delay in response, please rebase and generate report one more time.
I should be able to review this PR this month

romani:
issues fixed by https://github.com/checkstyle/checkstyle/pull/20342

PR #19065, opened 2026-03-01
Issue #19055: dependency pump fix error

ZaheerAhmadDev:
dependency: #19055

Update `error_prone_core from 2.46.0 to 2.48.0`
Remove: `Use IOUtils.toInputStream(String, Charset) instead.`
Use : `import java.nio.charset.StandardCharsets`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -233,7 +233,7 @@
     <forbiddenapis.version>3.10</forbiddenapis.version>
     <json-schema-validator.version>1.2.0</json-schema-validator.version>
     <checkerframework.version>3.54.0</checkerframework.version>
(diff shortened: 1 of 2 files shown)


ZaheerAhmadDev:
@romani, The one CI failures is related to PR, as it is fix of error related to  new  pump version `error_prone_core `.

romani:
commit prefix should be "dependency:"

ZaheerAhmadDev:
Updated prefix to "dependency:"

romani:
do not update PR, some circleci jobs might fail and it is ok.
Let all others to pass

PR #19123, opened 2026-03-05
Issue #5770: method wrapper for the log(line, column,..) DetailNode

DakshRJain737:
Issue: #5770 

Added a wrapper method log(DetailNode node, String key, Object... args) for log(int line, int column, ..).

### DIFF
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck.java
@@ -401,6 +401,19 @@ public abstract class AbstractJavadocCheck extends AbstractCheck {
     }
 
     /**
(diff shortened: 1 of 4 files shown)


DakshRJain737:
reopen after other PR merge

DakshRJain737:
Any update??

romani (commented):
Did you migrate all Check that does logging line and col ?
You upgraded only 3 , we have a bit more Checks, but they might be logging by line number only.

DakshRJain737:
I think i covered at all places.
Only 2 usages found which has hardcoded values.

image<br>


image


> Did you migrate all Check that does logging line and col ? You upgraded only 3 , we have a bit more Checks, but they might be logging by line number only.

Yes there are many checks that logs by line number only.

romani:
if CI pass we will merge this PR.

@DakshRJain737 , thanks a lot for update, sorry for late attention to it.

Issue #11315, opened 2022-02-15
Forbid usage of Truth's `StandardSubjectBuilder#fail` method.


nrmancuso:
Noticed at https://github.com/checkstyle/checkstyle/pull/11270#discussion_r803082439

Now that we have come to an agreement on `assertThrows`, we should enforce usage by adding a `MatchXPath` module to `checkstyle_checks.xml`:

```

➜   cat TestClass.java
import org.junit.Test;

import static com.google.common.truth.Truth.assertWithMessage;
import static org.junit.Assert.assertThrows;

public class TestClass {
    @Test
    public void testMethodBad() {
        try {
            int[] arr = new int[]{};
            int x = arr[2];
            assertWithMessage("IndexOutOfBoundsException expected").fail();
        } catch (IndexOutOfBoundsException ex) {
            assertWithMessage("Invalid error message")
                    .that(ex.getMessage())
                    .isEqualTo("Index 2 out of bounds for length 0");
        }
    }

    @Test
    public void testMethodGood() {

        int[] arr = new int[]{};
        final IndexOutOfBoundsException exception =
                assertThrows(IndexOutOfBoundsException.class,
                        () -> {
                            int i = arr[2];
                        });

        assertWithMessage("Invalid error message")
                .that(exception.getMessage())
                .isEqualTo("Index 2 out of bounds for length 0");

    }
}

➜  cat  config.xml        
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
        "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="TreeWalker">
        <module name="MatchXpath">
            <property name="query" value="//METHOD_CALL[./IDENT[@text='assertWithMessage']
          and (./following-sibling::IDENT[@text='fail'])]"/>
            <message key="matchxpath.match"
                     value="'Exceptions should be validated by JUnit's `Assert#assertThrows`."/>
        </module>
    </module>
</module>

➜  java -jar checkstyle-9.3-all.jar -c config.xml TestClass.java 
Starting audit...
[ERROR] TestClass.java:12:30: Exceptions should be validated by JUnit's `Assert#assertThrows`. [MatchXpath]
Audit done.
Checkstyle ends with 1 errors.


```

We will need to fix existing usages:
```
➜  checkstyle git:(master) grep -R "assertWithMessage.*fail()" | wc -l
[...]

PR #19168, opened 2026-03-10
Issue #11315: Add MatchXpath rule to forbid try-catch-fail and refactor utils/xpath packages

ayushactiveat:
issue #11315 
this pr starts fixing the issue described in #11315 by replacing the old `assertWithMessage(...).fail()` pattern in try-catch blocks with the cleaner `getExpectedThrowable` approach. I have added a `MatchXpath` rule and fixed all violations in the utils and xpath test . a temporary suppression has been added to keep the build passing while the remaining files are still being migrated. as they all together are close to 240 files in number i plan to fix the rest in smaller follow-up prs and will remove the suppression in the final one

### DIFF
--- a/config/checkstyle-checks.xml
+++ b/config/checkstyle-checks.xml
@@ -265,6 +265,13 @@
 
   <module name="TreeWalker">
     <property name="tabWidth" value="4"/>
+    <module name="MatchXpath">
+      <property name="id" value="MatchXpathForbidTryCatchFail"/>
+      <property name="query" value="//METHOD_CALL[./IDENT[@text='assertWithMessage']
+        and (./following-sibling::IDENT[@text='fail'])]"/>
+      <message key="matchxpath.match"
+               value="Exceptions should be validated by `getExpectedThrowable`."/>
+    </module>
 
     <!-- Annotations -->
     <module name="AnnotationLocation">
--- a/config/suppressions.xml
+++ b/config/suppressions.xml
@@ -172,4 +172,10 @@
   <!-- Uses non-ascii symbols intentionally -->
   <suppress id="checkASCII"
             files="[\\/]src[\\/]test[\\/]java[\\/]com[\\/]puppycrawl[\\/]tools[\\/]checkstyle[\\/]ConfigurationLoaderTest.java"/>
+
+  <!-- Remove suppression once all violations are fixed, tracked in #19176 -->
+  <suppress checks="MatchXpath"
+            id="MatchXpathForbidTryCatchFail"
+            files=".*[\\/]src[\\/]test[\\/]java[\\/]com[\\/]puppycrawl[\\/]tools[\\/]checkstyle[\\/](?!utils|xpath).*"/>
+
 </suppressions>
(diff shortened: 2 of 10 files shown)


ayushactiveat:
hi @romani , the fails are unrelated, please let me know if u need anything else changed

config/checkstyle-checks.xml
```diff
@@ -265,6 +265,12 @@
 
   <module name="TreeWalker">
     <property name="tabWidth" value="4"/>
+    <module name="MatchXpath">
+      <property name="query" value="//METHOD_CALL[./IDENT[@text='assertWithMessage']
+        and (./following-sibling::IDENT[@text='fail'])]"/>
```
romani, line 271: Can we archive this by usage of forbiddenapi tool ?
https://github.com/checkstyle/checkstyle/blob/master/config/signatures.txt

ayushactiveat: @romani i checked the forbiddenapis approach but  this is a test only dependency, the rule has to go in signatures-test.txt but as there are still around 240 files using the old` assertWithMessage(...).fail()` , adding the rule right now breaks the build for other, unmigrated files. and forbiddenapis doesnt allow us to easily suppress the rule for those  240 files while we work on them in smaller prs in batches, as i proposed initially.
how would u want me to work here, i mean we could - maybe migrate all 240 files in this single pr to use signatures-test.txt right now, but it could be a mess, or keep the match-xpath + suppressions.xml approach for this pr and next all batches,  then transition it to signatures-test.txt in the final pr.
please let me know which what do u prefer or anyother way we could look at

config/checkstyle-checks.xml
```diff
@@ -1021,4 +1027,9 @@
 
   </module>
 
+  <module name="SuppressionSingleFilter">
+    <property name="checks" value="MatchXpath"/>
+    <property name="files" value=".*[\\/]sevntu\.checkstyle[\\/].*"/>
```
romani: Please do this suppression in seventh project.
We can merge such update before this PR

config/suppressions.xml
```diff
@@ -172,4 +172,8 @@
   <!-- Uses non-ascii symbols intentionally -->
   <suppress id="checkASCII"
             files="[\\/]src[\\/]test[\\/]java[\\/]com[\\/]puppycrawl[\\/]tools[\\/]checkstyle[\\/]ConfigurationLoaderTest.java"/>
+
+  <!-- Remove suppression once all violations are fixed, tracked in #11315 -->
```
romani: Please create new issue to do long lasting migration from one format to another.
It will be one of good xxxx issues.
This PR will fix 11315 . We need your time and efforts to be spent on more complicated issues.

config/checkstyle-checks.xml
romani: I agree. See comments below.

config/checkstyle-checks.xml
ayushactiveat: i have opened the pr to add the suppression in the sevntu repo - sevntu-checkstyle/sevntu.checkstyle#1163
let me know when that is merged, then i'll remove the SuppressionSingleFilter from this pr and push for final review

ayushactiveat:
@romani ig it's ready for final review.
here are all the updates-
->created issue #19176 for the long lasting migration

->added a tracking suppression for #19176 in suppressions.xml using the regex (?!utils|xpath) to keep the build green while blocking those regressions.

->removed the SuppressionSingleFilter from this pr's checkstyle-checks.xml. opened a pr directly in the sevntu repo (sevntu-checkstyle/sevntu.checkstyle#1163).

->refactored the migration for the remaining leftover files in the utils and xpath packages.

let me know if any other change is needed

Issue #19156, opened 2026-03-09
VariableDeclarationUsageDistance: no violation on variable used in inner class


romani:
from https://github.com/checkstyle/checkstyle/issues/13011#issuecomment-3996360057
```
$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name = "Checker">
    <module name="TreeWalker">
        <module name="VariableDeclarationUsageDistance">
          <property name="validateBetweenScopes" value="true"/>
          <property name="allowedDistance" value="1"/>
          <property name="ignoreFinal" value="false"/>
        </module> 
    </module>
</module>

$ cat Test.java 
package com.puppycrawl.tools.checkstyle.checks.coding.variabledeclarationusagedistance;

import java.util.ArrayList;
import java.util.List;

public class Test {

    void nothing() {
    }

    class Parent {
        void mm() {
        }
        <T> void xx(List<T> m){}
    }

    public void method7() {
        // until this issue, not violation, the below distance is 3
        Integer t = 5;
        nothing();
        System.out.println();
        class BClass extends Parent {
            @Override
            void mm() {
                System.out.println(t);
            }
        }
    }
}

$ java -jar checkstyle-13.3.0-all.jar -c config.xml Test.java 
Starting audit...
Audit done.

```

Expected: violation on `t` as distance is 3, allowed distance is 1.

romani:
@lithops-zty, please help to fix this issue

lithops-zty:
OK! I will investigate the problematic `isInitializationSequence()` method identified in https://github.com/checkstyle/checkstyle/issues/13011#issuecomment-3996360057

surajgojanur:
Hi, I’d like to work on this issue.

My plan is to first reproduce the problem with a focused test, then identify the relevant logic in VariableDeclarationUsageDistance, and submit a minimal fix with tests.

Please let me know if there is any preferred approach before I start.

lithops-zty:
[...]

PR #19170, opened 2026-03-11
Issue #19156: Fix VariableDeclarationUsageDistanceCheck for inner class

lithops-zty:
Issue: #19156

Fixed false negative of VariableDeclarationUsageDistanceCheck not reporting violation when
- The check's property `validateBetweenScopes` is set to `true`;
- The variable definition and its first usage exceeds the maximum allowed distance; and
- The first usage is inside an inner class.

Root Cause
---
The method `isInitializationSequence()` fails to check for method calls in outer scopes. 

This method is used to detect initialization sequences from variable first usage up to variable definition. If detected, it exempts the variable from distance check.

If variable definition is in an outer scope, the method is supposed to check inside outer scope as well. However, the current implementation only checks within the inner scope.

Example
---
```java
  Integer t = 5;  // violation, 'Distance .* is 3.
  nothing();                                (1)
  list.add(5);                              (2)
  class BClass extends Parent {
      @Override
      void mm() {
          list.add(t);                      (3)
      }
  }
```
The current implementation only checks (3) and concludes yes, it is an initialization sequence; therefore does not report a violation.

After the fix, (2) and (1) will also be checked and violation is reported since `nothing();` is not an initialization.

Fix
---
When we reach the first sibling, we continue to check previous siblings of the parent node (or grandparent or great-grandparent, based on the type of current scope). 

I also added an additional set of tests to ensure correct behavior in various nested blocks.

### DIFF
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceInitializationSequence4.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceInitializationSequence4.java
@@ -1,0 +1,119 @@
+/*
+VariableDeclarationUsageDistance
+allowedDistance = 1
+ignoreVariablePattern = (default)
+validateBetweenScopes = true
+ignoreFinal = false
+
+
+*/
+
+package com.puppycrawl.tools.checkstyle.checks.coding.variabledeclarationusagedistance;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class InputVariableDeclarationUsageDistanceInitializationSequence4 {
+    public void tryBlock() {
+        List<Integer> list = new ArrayList<>();
+        System.out.println(0);
+        try {
+            System.out.println(list.size());
+        } catch (Exception e) {
+        }
+
+        List<Integer> list2 = new ArrayList<>();  // violation 'Distance .* is 2'
+        nothing();
+        try {
+            System.out.println(list2.size());
+        } catch (Exception e) {
+        }
+    }
+
+    public void catchBlock() {
+        List<Integer> list = new ArrayList<>();
+        System.out.println(0);
+        try {
+        } catch (Exception e) {
+            System.out.println(list.size());
+        }
+        List<Integer> list2 = new ArrayList<>();  // violation 'Distance .* is 2'
+        nothing();
+        try {
+        } catch (Exception e) {
+            System.out.println(list2.size());
+        }
+    }
+
+    public void finallyBlock() {
+        List<Integer> list = new ArrayList<>();
+        System.out.println(0);
+        try {
+        } catch (Exception e) {
+        } finally {
+            System.out.println(list.size());
+        }
+
+        List<Integer> list2 = new ArrayList<>();  // violation 'Distance .* is 2'
+        nothing();
+        try {
+        } catch (Exception e) {
+        } finally {
+            System.out.println(list2.size());
+        }
+    }
+
+    public void tryResource() {
+        List<Integer> list = new ArrayList<>();   // violation 'Distance .* is 2'
+        this.getAutoCloseable(0);
+        this.getAutoCloseable(0);
+        try (AutoCloseable a = this.getAutoCloseable(list.size())) {
+        } catch (Exception e) {
+        }
+
+        List<Integer> list2 = new ArrayList<>();  // violation 'Distance .* is 2'
+        nothing();
+        System.out.println(0);
+        try (AutoCloseable a = new Close(list2.size())) {
+        } catch (Exception e) {
+        }
+    }
+
+    public void multipleBlocks() {
+        List<Integer> list = new ArrayList<>();
+        nothing();
+        try {
+            nothing();
+            int x = list.size();
+        } catch (Exception e) {
+        } finally {
+            System.out.println(list.size());
+        }
+
+        List<Integer> list2 = new ArrayList<>();  // violation 'Distance .* is 2'
+        nothing();
+        list2.add(1);
+        try {
+            int x = list2.size();
+        } catch (Exception e) {
+        } finally {
+            System.out.println(list2.size());
+        }
+    }
+
+    private void nothing() {
+    }
+
+    private class Close implements AutoCloseable {
+        public Close(int i) {
+        }
+
+        @Override
+        public void close() throws Exception {
+        }
+    }
+
+    private AutoCloseable getAutoCloseable(int i) {
+        return new Close(i);
+    }
+}
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceTryResources.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceTryResources.java
@@ -16,8 +16,8 @@ import java.util.*;
 
 public class InputVariableDeclarationUsageDistanceTryResources {
     public int methodTry() {
-        String a = ""; // violation 'Distance .* is 2.'
-        String b = "abc"; // violation 'Distance .* is 2.'
+        String a = "";
+        String b = "abc";
         System.out.println();
         try (AutoCloseable j = new java.io.StringReader(b);
              final AutoCloseable i = new java.io.StringReader(a);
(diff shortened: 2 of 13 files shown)


lithops-zty:
Hi @romani! PR is ready for review. Would appreciate you time to guide me on improving this PR. 
CI failure seems unrelated.

romani:
Github, generate report for VariableDeclarationUsageDistance/all-examples-in-one

romani:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/78dc26faa963e4f8efe92b8f1f1ea3098a5ceabd_2026215820/reports/diff/apache-ant/index.html#A2 this is likely false positive, imagine `componentHelper` is also used in IF, both are in IF , so non of them can change location.

https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/78dc26faa963e4f8efe92b8f1f1ea3098a5ceabd_2026215820/reports/diff/RxJava/index.html#A1 this is better example of false positive. You will unlikely make this code different to avoid violation.
Please add it to input .

-------

Just add to Inputs , not sure why we started to report this, but violation is correct:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/78dc26faa963e4f8efe92b8f1f1ea3098a5ceabd_2026215820/reports/diff/pmd/index.html#A1

romani:
@lithops-zty , please rebase PR one more time. Sorry for conflicts.

romani:
@lithops-zty , do you have time to finish this PR  or you can let somebody else to continue this PR ?

lithops-zty:
@romani Yes I will finish it.

romani:
Github, generate report for VariableDeclarationUsageDistance/all-examples-in-one

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceTryResources.java
romani, line 24: something weird.
violation from a is gone but stays for b.

Please explain why this happening.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceInitializationSequence4.java
```diff
@@ -0,0 +1,99 @@
...
+        this.getAutoCloseable(0);
+        try (AutoCloseable a = this.getAutoCloseable(list.size())) {
+        } catch (Exception e) {
+        }
+
+        List<Integer> list2 = new ArrayList<>();  // violation 'Distance .* is 2'
+        nothing();
+        System.out.println(0);
+        try (AutoCloseable a = new Close(list2.size())) {
+        } catch (Exception e) {
```
romani, line 78: it should be distance 3, right ?

romani (changes requested):
questions:

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceInitializationSequence4.java
lithops-zty: Yes. Current code does not not increment distance when first usage is encountered within brackets of `for` `try` `while` etc. I'll change logic as 3 is more logical

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceTryResources.java
lithops-zty: I believe this is a bug. While for both `a` and `b`, their usage in `RESOURCE_SPECIFICATION` should be the first usage, current code incorrectly chosen `b`'s usage in `try` as the first usage.

`a` and `b` differ in that `a` is used in both `try` and `catch` while `b` is used in `try` only. This led to different behavior.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceInitializationSequence4.java
lithops-zty: Sorry but I checked existing logic, and it seems that not incrementing is the convention. 
For example, before this PR, the following
```java
        int a = 0;  // violation 'Distance .* is 3.'
        xx();  // dist == 1
        xx();  // dist == 2
        xx();  // dist == 3
        if (a == 0) {  // dist == 3; first usage is in operator expression

        }
```
distance is 3 instead of 4.

I feel we just keep this logic as is.

lithops-zty:
> Just add to Inputs , not sure why we started to report this, but violation is correct:
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/78dc26faa963e4f8efe92b8f1f1ea3098a5ceabd_2026215820/reports/diff/pmd/index.html#A1

This is a case where first usage resides in `return`. `return` introduces a scope boundary:
```
LITERAL_RETURN
├── EXPR                        
│   └── METHOD_CALL (function.call(c, list))
└── SEMI
```

The old `isInitializationSequence` couldn't cross scope boundaries, causing a false negative.

src/test/resources/com/puppycrawl/tools/checkstyle/checks/coding/variabledeclarationusagedistance/InputVariableDeclarationUsageDistanceTryResources.java
lithops-zty: I fixed it in https://github.com/checkstyle/checkstyle/commit/4522fadf4158ad8590a0bdd5a6ee931f2bf7beb6, now violation for b is gone as well

lithops-zty:
CI failure unrelated

lithops-zty:
Github, generate report for VariableDeclarationUsageDistance/all-examples-in-one

romani:
@lithops-zty , please analyze report and let me know when you do not see any problems.

lithops-zty:
Hi @romani, I've checked the diff report and it looks fine. There are three sources of changes:
1. (New violations) Previous false negatives due to `isInitializationSequence()` unable to cross scope boundaries. These variables exceeds the max distance, and their first usage resides in an inner scope. All statements in the inner scope and before first usage are either **variable definitions** or are **part of an initialization sequence**; while at least one statement in the outer scope falls outside these two categories. The corrected `isInitializationSequence()` now can detect existence of the latter and no longer incorrectly suppresses violations. 
   - Examples: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/98a2aef877c319c46ba4a605861faa414d592c48_2026110934/reports/diff/rhsm/index.html#A4; https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/98a2aef877c319c46ba4a605861faa414d592c48_2026110934/reports/diff/spoon/index.html#A3
2. (Distance increase, possibly new violations) Distance in inner scope is now counted for variables whose first usage are inside the `try` block of a try-with-resource construct. Old `getFirstNodeInsideTryCatchFinallyBlocks()` failed to detect usage there as it wasn't designed to handle the `RESOURCE_SPECIFICATION` node.
```java
int a = 0;
doIt();
try (...) {
    doIt();  // before PR, this was mistakenly left out in distance computation
    doThat(a);
}
```
   - Example: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/98a2aef877c319c46ba4a605861faa414d592c48_2026110934/reports/diff/spoon/index.html#A5
3. (Distance decrease) The distance for variables whose first usages are inside try resource specification is now one less its old value. 
```java
int a = 0;
doIt();
try (AutoCloseable resource = new A(a) ) { // distance changed 2 -> 1
...
}
```
This is in line with similar scenarios where first usages are inside brackets after `for`, `while`, `if` etc.
```java
int a = 0;
doIt();
while (a >= 0) { // distance 1
...
}
```
   - Example: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/98a2aef877c319c46ba4a605861faa414d592c48_2026110934/reports/diff/openjdk25/index.html#A326

lithops-zty:
@romani I don't find problems in report. Please let me know any further changes if needed, thank you!

romani:
Github, generate report for VariableDeclarationUsageDistance/all-examples-in-one

romani:
Github, generate report for VariableDeclarationUsageDistance/all-examples-in-one

romani (approved):
thanks a lot !

it becomes better, even distance term becomes not less and less  exact, but I think it is ok.

Issue #17128, opened 2025-05-23
test to check indentation trailing comments are vertically aligned


mohitsatr:
I have read check documentation: https://checkstyle.org/checks/misc/indentation.html
I have downloaded the latest cli from: https://checkstyle.org/cmdline.html#Download_and_Run
I have executed the cli and showed it below, as cli describes the problem better than 1,000 words

from https://github.com/checkstyle/checkstyle/pull/16627#discussion_r2078804949

Unaligned trailing comments:
![Image](https://github.com/user-attachments/assets/bf76e38d-0879-42c4-99bd-30b4fbee6c21)

Aligned trailing comments:
![Image](https://github.com/user-attachments/assets/35e71bf4-23e5-40fa-888b-c9a1df030951)

I'd like a see a test which verifies indentation trailing comments are vertically aligned in all indentation input files. 
we could use the column no. to for this purpose. there are some expection cases we need to keep in mind like:

![Image](https://github.com/user-attachments/assets/b539ba57-90f4-425e-a2ed-7b2de4088615)

I feel it's better not to disturb the position of non-trailing ones

Here is suppression for files that are not updates to correct form yet:
https://github.com/checkstyle/checkstyle/blob/d4c808744b140ab2273ab43697fd226607e6ee6d/src/test/java/com/puppycrawl/tools/checkstyle/IndentationTrailingCommentsVerticalAlignmentTest.java#L43-L46
until ALLOWED_VIOLATION_FILES is present, issue should be opened, help is appreciated.

here is latest state of this suppression list: https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/IndentationTrailingCommentsVerticalAlignmentTest.java#L43

Good example of expected update:
https://github.com/checkstyle/checkstyle/pull/17714/files

StefanLapointe:
Just to see if I understood correctly, is this what you had in mind?
- Create a new test in `src/test/java/com/puppycrawl/tools/checkstyle` called something like `IndentationTrailingCommentsVerticalAlignmentTest.java`.
- Loop over files from src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/indentation`.
- For each file, find every line that contains an indentation trailing comment (defined as anything that `IndentationCheckTest` uses to define expected behaviour including "comments" within multi-line strings).
- Calculate the column at which each indentation trailing comment starts, and fail the test on the first mismatch that is found.

[...]

PR #19177, opened 2026-03-11
Issue #17128: shorten InputIndentationStrictCondition.java

Carbon14-48:
Hi @romani ,

I am starting to tackle Issue #17128 by working on `InputIndentationStrictCondition.java`.

## My understanding of the task:
- I should **not** move or modify the code itself, or change the `indent:`/`exp:` values in the trailing comments.
- The only allowed change is adjusting the **padding (spaces)** between code and each `//indent:` trailing comment, so that:
    - All such comments (except for `package` and `import` lines) start at the same column (vertical alignment) within the file
    - Every line is ≤100 characters
    - The `package` and `import` comment alignment does not matter, as those are ignored by the test

I just want to confirm that this approach is correct before I continue with more files and PRs.

Thank you for your guidance!

---

**Please let me know if this understanding is correct 🙏**

### DIFF
--- a/src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationStrictCondition.java
+++ b/src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/InputIndentationStrictCondition.java
@@ -1,14 +1,14 @@
-package com.puppycrawl.tools.checkstyle.checks.indentation.indentation;    //indent:0 exp:0
+package com.puppycrawl.tools.checkstyle.checks.indentation.indentation;//indent:0 exp:0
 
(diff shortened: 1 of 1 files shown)


Carbon14-48:
@romani  No problem, it’s fun to contribute to this!

Can you please review my new PR: #19183?  
I wrapped the necessary code lines in `InputIndentationValidClassDefIndent1.java` to fit the 100-column limit, aligned the trailing comments, and updated the expected line number for the violation in the related test.

Just to confirm:  
- Is it allowed to wrap long lines in the input file (not just rename or re-align) when it’s otherwise not possible to fit within 100 columns?
- And when line numbers shift, is updating the expected line in `IndentationCheckTest` the correct procedure?

I want to be sure I’m following project guidelines for Issue #17128. Thank you!

Issue #19146, opened 2026-03-08
update JavadocTypeCheck to use AST of javadoc


romani:
we introduced new javadoc AST in scope of https://github.com/checkstyle/checkstyle/issues/17514

Now it is time to migrate [JavadocTypeCheck](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java) Check to use AST, rather than regexp for parsing of javadoc.

Whole design should stay as is, same properties.
But check should extend ` extends AbstractJavadocCheck` and do all logic based on AST nodes.
There are already bunch of Checks that doing this, so look around for examples.

remove suppresion that is mentioned at https://github.com/checkstyle/checkstyle/issues/11166

avadhutmali:
I am working on this issue.

avadhutmali:
@romani ,While migrating JavadocTypeCheck to AbstractJavadocCheck, two test input files required modifications due to behavioral differences between regexp-based and AST-based parsing.
The Javadoc AST parser is stricter than the old regexp scanner — it cannot recognize block tags at invalid positions or parse severely malformed HTML in **param** descriptions. As a result, tests were failing, so I made the following modifications:

https://www.diffchecker.com/JkNvl5VC/ — ****    **author** changed to * **author** since AST doesn't parse tags with excessive leading asterisks as valid block tags
https://www.diffchecker.com/gtBonGAk/ — simplified malformed HTML in **param** descriptions that caused AST parse failures

Could you confirm if these modifications to test input files are acceptable?

aclfe:
@avadhutmali 
Failure of Javadoc AST parser on malformed HTML in param descriptions or block tags at invalid positions is normal with the jdoc parser. main concern is with the tags not the description following the tags in Javadoc (as seen in the diffchecker link, which is very neat), so it shouldn't be a concern. Keep making the changes 👍

romani:
Some small changes in behavior is completely ok, just make sure they are logical and better than before.
If in doubt, we will see in PR

romani:
pay attention for https://github.com/checkstyle/checkstyle/pull/18059
where we remove finals from methods of  
src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/AbstractJavadocCheck.java

if required apply same to your PR to let continue

avadhutmali:
[...]

PR #19189, opened 2026-03-12
Issue #19146: migrate JavadocTypeCheck from AbstractCheck to Abstract…

avadhutmali:
Issue: #19146

## Migrate `JavadocTypeCheck` from `AbstractCheck` to `AbstractJavadocCheck`

---

### Why This Migration

`JavadocTypeCheck` previously extended `AbstractCheck` and used `FileContents.getJavadocBefore()` with regexp-based tag scanning. This approach scanned raw Javadoc text line by line and was unable to leverage the structured Javadoc AST. Migrating to `AbstractJavadocCheck` allows the check to operate on a fully parsed `DetailNode` tree via `JavadocCommentsTokenTypes`, consistent with how all other modern Javadoc checks in the codebase work.

---

### Core Implementation Changes (`JavadocTypeCheck.java`)

**Class structure:**
- Changed `extends AbstractCheck` → `extends AbstractJavadocCheck`
- Removed `getDefaultTokens()`, `getAcceptableTokens()`, `getRequiredTokens()` overrides — these are `final` in `AbstractJavadocCheck` and must not be overridden
- Added `getDefaultJavadocTokens()` returning `JavadocCommentsTokenTypes.JAVADOC_CONTENT` — the root node of every parsed Javadoc comment
- Replaced `visitToken(DetailAST)` with `visitJavadocToken(DetailNode ast)` — entry point for each Javadoc comment
- Removed all `getFileContents()` usage — no longer needed since AST provides full structure

**How the parent Java element is resolved:**
`AbstractJavadocCheck` provides `getBlockCommentAst()` which returns the `BLOCK_COMMENT_BEGIN` token in the Java AST. From there, `getParentAst()` navigates up through `MODIFIERS` or `TYPE` nodes to reach the owning type definition (CLASS_DEF, ENUM_DEF, RECORD_DEF, INTERFACE_DEF, ANNOTATION_DEF).

**New methods added:**
- `getParentAst(DetailAST blockCommentAst)` — navigates BLOCK_COMMENT → parent Java AST node. Handles three cases: parent is TYPE/MODIFIERS (go up one), parent's parent is MODIFIERS (go up two), otherwise return parent as-is
- `isTypeDefinition(DetailAST)` — returns true for CLASS_DEF, INTERFACE_DEF, ENUM_DEF, ANNOTATION_DEF, RECORD_DEF
- `shouldCheck(DetailAST)` — checks scope, excludeScope, and allowedAnnotations — unchanged logic, just moved to use `ScopeUtil.getSurroundingScope`
- `getBlockTags(DetailNode javadocRoot)` — iterates direct children of JAVADOC_CONTENT, collects all JAVADOC_BLOCK_TAG nodes
- `checkUnknownTags(Iterable<DetailNode> blockTags)` — for each CUSTOM_BLOCK_TAG, resolves name via `getTagName()`, fires MSG_UNKNOWN_TAG if not in `JavadocTagInfo` registry. Uses `tagNameNode.getLineNumber()` and `blockTag.getColumnNumber()` for accurate position reporting
- `getTagName(DetailNode tagNode)` — resolves tag name: AUTHOR_BLOCK_TAG → `JavadocTagInfo.AUTHOR.getName()`, VERSION_BLOCK_TAG → `JavadocTagInfo.VERSION.getName()`, CUSTOM_BLOCK_TAG → TAG_NAME child via `JavadocUtil.findFirstToken`. Returns `""` if not found
- `getTagDescription(DetailNode tagNode)` — finds DESCRIPTION node via `JavadocUtil.findFirstToken`, delegates to `collectAllText`
- `collectAllText(DetailNode node, StringBuilder builder)` — recursive helper that appends TEXT nodes, converts NEWLINE to single space (to avoid merging words across lines), skips LEADING_ASTERISK. Two-method design: public entry creates `StringBuilder`, private helper recurses
- `checkTag(DetailAST ast, Iterable<DetailNode> blockTags, String tagName, Pattern formatPattern)` — iterates block tags, matches by `getTagName()`, validates description against `formatPattern`. Fires MSG_TAG_FORMAT on mismatch, MSG_MISSING_TAG if no matching tag found. Only runs when `formatPattern != null`
- `checkTypeParamTag(DetailAST ast, Collection<DetailNode> blockTags, String typeParamName)` — uses single combined stream filter `isParamBlockTag && isTypeParamTag` to find matching `@param <T>`. Fires MSG_MISSING_TAG if not found
- `checkComponentParamTag(DetailAST ast, Collection<DetailNode> blockTags, String recordComponentName)` — uses single combined stream filter `isParamBlockTag && !isTypeParamTag` to find matching `@param name`. Fires MSG_MISSING_TAG if not found
- `checkUnusedParamTags(List<DetailNode>, List<String>, List<String>)` — iterates param block tags, checks if name exists in typeParamNames or recordComponentNames. Fires MSG_UNUSED_TAG_GENERAL for empty/`<>` display names, MSG_UNUSED_TAG otherwise
- `isParamBlockTag(DetailNode blockTag)` — checks `blockTag.getFirstChild().getType() == PARAM_BLOCK_TAG`
- `isTypeParamTag(DetailNode blockTag)` — finds PARAMETER_NAME via `JavadocUtil.findFirstToken`, checks if text starts with `<` and ends with `>` using `OPEN_ANGLE_BRACKET`/`CLOSE_ANGLE_BRACKET` constants
- `getParamName(DetailNode blockTag)` — finds PARAMETER_NAME, strips angle brackets if type param. Returns `""` if PARAMETER_NAME not found
- `getRecordComponentNames(DetailAST node)` — unchanged logic from original

**Dead code removed:**
- `extractParamName(String)` — replaced by `getParamName(DetailNode)`
- `TYPE_NAME_IN_JAVADOC_TAG` pattern constant — no longer needed
- `TYPE_NAME_IN_JAVADOC_TAG_SPLITTER` pattern constant — no longer needed
- `isInsideInlineTag(String textBefore)` — regexp-based inline tag detection, not needed with AST
- `logWithDebug(...)` — debug helper used during development

**Copilot review fixes applied:**
- `checkUnknownTags` Javadoc — removed stale reference to "inside inline tag" logic that no longer exists
- `getParamName` Javadoc — updated to accurately describe that only PARAMETER_NAME is read, angle brackets stripped
- `collectAllText` — refactored from single recursive method to two-method design with shared `StringBuilder`; NEWLINE now converted to space to prevent word merging (e.g. `"OliverBurn"` → `"Oliver Burn"`)
- `isTypeParamTag` — simplified to direct `return` expression, removed unnecessary `boolean isTypeParam` store
- `getTagName` — `tagNameNode` null check retained since `CUSTOM_BLOCK_TAG` can theoretically have no TAG_NAME child in edge cases

**SpotBugs fixes:**
- `checkComponentParamTag` and `checkTypeParamTag` — combined back-to-back stream filters into single filter lambda
- `checkTag` and `checkUnknownTags` — parameter type widened from `List` to `Iterable`
- `checkComponentParamTag` and `checkTypeParamTag` — parameter type widened from `List` to `Collection`
- `isTypeParamTag` — removed unnecessary store before return

---

### Config Changes

**`config/suppressions.xml`:**
- Removed `<suppress id="noUsageOfGetFileContentsMethod" files="JavadocTypeCheck.java"/>` — no longer uses `getFileContents()`

**`config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml`:**
- Removed stale suppression for `String.join(...)` — old code referencing `lines` array no longer exists
- Removed stale suppression for `matchInAngleBrackets.group(1)` — old regexp-based param extraction removed
- `getTagName` initialised to `""` (not `null`) — eliminates `@Nullable` return warning without needing suppression

---

### Test Input File Changes

**Behavioral differences from AST parser — confirmed acceptable by @romani and @aclfe:**

1. **`InputJavadocTypeJavadoc_1.java`, `_2.java`, `_3.java`**
   - `****    @author` → `* @author` for classes firing MSG_TAG_FORMAT
   - AST parser does not promote block tags preceded by excessive leading asterisks — they are parsed as plain text

2. **`InputJavadocTypeTestTrimProperty.java`**
   - `@param < A  >` → `@param <A>` — AST tokenizes `< A  >` with spaces as `PARAMETER_NAME = "<"` and description `" A  > ..."`, breaking type param detection

3. **`InputJavadocTypeParamDescriptionWithAngularTags.java`**
   - Simplified malformed HTML in `@param` descriptions — severely malformed HTML causes AST parse failures, skipping `visitJavadocToken` entirely

4. **`InputJavadocTypeRecordParamDescriptionWithAngularTags.java`**
   - Redesigned to produce violations at lines matching AST-based behavior

5. **`InputJavadocTypeAnnotationsInCodeBlock2.java`**
   - `<pre>` with `@unknown` on its own line replaced with `{@code @unknown}` per @aclfe guidance
   - Root cause: AST parser promotes `@unknown` to a JAVADOC_BLOCK_TAG, making `</pre>` appear at unexpected grammar position → `no viable alternative at input '</'`

6. **`InputJavadocTypeNoJavadocOnInterface.java`**
   - `tokens = INTERFACE_DEF` → `tokens = (default)BLOCK_COMMENT_BEGIN` — `AbstractJavadocCheck` only accepts `BLOCK_COMMENT_BEGIN`

7. **All `Input*.java` files**
   - Config block updated to `tokens = (default)BLOCK_COMMENT_BEGIN`

**New input classes added for 100% branch coverage:**

8. **`InputJavadocTypeJavadoc.java`**
   - `InputJavadocTypeJavadocInlineTag` — `@author {@code Oliver} Burn` — triggers recursive `collectAllText` call on inline tag node inside DESCRIPTION
   - `InputJavadocTypeNoAuthorDesc` — `@author X` not matching `0*` — covers `getTagDescription` returning non-empty string that fails format check

9. **`InputJavadocTypeJavadoc_1.java`**
   - `InputJavadocTypeNoDesc` — bare `@author` (no description text) not matching `ABC` — covers `getTagDescription` returning empty string path

10. **`InputJavadocTypeBadTag.java`**
    - `InputJavadocTypeBadTagEmptyParam` — `@param` with no name — fires MSG_UNUSED_TAG_GENERAL (displayName is `""`)
    - `InputJavadocTypeBadTagEmptyTypeParam` — `@param <>` — fires MSG_UNUSED_TAG_GENERAL (displayName is `"<>"`)
    - `InputJavadocTypeBadTagUnclosedTypeParam` — `@param <T` — PARAMETER_NAME is `"<T"` (no closing `>`), treated as record component, fires MSG_UNUSED_TAG for `<T`
    - `InputJavadocTypeBadTagWithMethod` — method Javadoc inside class — covers `getParentAst` path where parent is METHOD_DEF
    - `InputJavadocTypeBadTagFieldJavadoc` — field Javadoc inside class — covers `getParentAst` else-if branch
    - Floating Javadoc (not attached to any type definition) — covers `parentAst == null` false branch in `visitJavadocToken`

11. **`package-info.java` style input** — Javadoc before package declaration — covers `parentNode.getParent() == null` branch in `getParentAst`

---

### Test Method Changes (`JavadocTypeCheckTest.java`)

- `testAuthorRegularEx` — updated expected violation line numbers after new classes added to input file
- `testAuthorRegularExError` — added `InputJavadocTypeNoDesc` violation, corrected MSG_MISSING_TAG vs MSG_TAG_FORMAT for `****    @author` lines across all three type kinds (class/enum/@interface)
- `testBadTag` — added expected violations for `InputJavadocTypeBadTagEmptyParam` (MSG_UNUSED_TAG_GENERAL), `InputJavadocTypeBadTagEmptyTypeParam` (MSG_UNUSED_TAG_GENERAL), `InputJavadocTypeBadTagUnclosedTypeParam` (MSG_UNUSED_TAG for `<T`)
- `testJavadocTypeParamDescriptionWithAngularTags` — updated expected violations
- `testJavadocTypeRecordParamDescriptionWithAngularTags` — updated expected violations
- `testLimitViolationsBySpecifyingTokens` — updated input file token config
- `testTrimOptionProperty` — updated for `@param <A>` (no spaces)
- `testAnnotationsInCodeBlock2` — removed `@Disabled`, updated expected violations for `{@code}` approach

---

### Coverage (JaCoCo)

- **Lines: 100%**
- **Branches: 100%**

### DIFF
--- a/config/suppressions.xml
+++ b/config/suppressions.xml
@@ -133,8 +133,6 @@
   <suppress id="noUsageOfGetFileContentsMethod" files="JavadocMethodCheck.java"/>
   <!-- until #19145 -->
   <suppress id="noUsageOfGetFileContentsMethod" files="JavadocStyleCheck.java"/>
-  <!-- until #19146 -->
-  <suppress id="noUsageOfGetFileContentsMethod" files="JavadocTypeCheck.java"/>
   <!-- until #19147 -->
   <suppress id="noUsageOfGetFileContentsMethod" files="JavadocVariableCheck.java"/>
   <!-- until #19148 -->
--- a/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
+++ b/src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
@@ -20,23 +20,19 @@
 package com.puppycrawl.tools.checkstyle.checks.javadoc;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import com.puppycrawl.tools.checkstyle.StatelessCheck;
-import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
-import com.puppycrawl.tools.checkstyle.api.FileContents;
+import com.puppycrawl.tools.checkstyle.api.DetailNode;
+import com.puppycrawl.tools.checkstyle.api.JavadocCommentsTokenTypes;
 import com.puppycrawl.tools.checkstyle.api.Scope;
-import com.puppycrawl.tools.checkstyle.api.TextBlock;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
 import com.puppycrawl.tools.checkstyle.utils.AnnotationUtil;
 import com.puppycrawl.tools.checkstyle.utils.CheckUtil;
-import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 import com.puppycrawl.tools.checkstyle.utils.JavadocUtil;
 import com.puppycrawl.tools.checkstyle.utils.ScopeUtil;
 import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
@@ -70,7 +66,7 @@ import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
  */
 @StatelessCheck
 public class JavadocTypeCheck
-    extends AbstractCheck {
+    extends AbstractJavadocCheck {
 
     /**
      * A key is pointing to the warning message text in "messages.properties"
@@ -113,14 +109,6 @@ public class JavadocTypeCheck
 
     /** Javadoc tag token literal. */
     private static final String JAVADOC_TAG_TOKEN = "@";
-
-    /** Pattern to match type name within angle brackets in javadoc param tag. */
-    private static final Pattern TYPE_NAME_IN_JAVADOC_TAG =
-            Pattern.compile("^<([^>]+)");
-
-    /** Pattern to split type name field in javadoc param tag. */
-    private static final Pattern TYPE_NAME_IN_JAVADOC_TAG_SPLITTER =
-            Pattern.compile("\\s+");
 
     /** Specify the visibility scope where Javadoc comments are checked. */
     private Scope scope = Scope.PRIVATE;
@@ -217,63 +205,64 @@ public class JavadocTypeCheck
     }
 
     @Override
-    public int[] getDefaultTokens() {
-        return getAcceptableTokens();
-    }
-
-    @Override
-    public int[] getAcceptableTokens() {
+    public int[] getDefaultJavadocTokens() {
         return new int[] {
-            TokenTypes.INTERFACE_DEF,
-            TokenTypes.CLASS_DEF,
-            TokenTypes.ENUM_DEF,
-            TokenTypes.ANNOTATION_DEF,
-            TokenTypes.RECORD_DEF,
+            JavadocCommentsTokenTypes.JAVADOC_CONTENT,
         };
     }
 
     @Override
-    public int[] getRequiredTokens() {
-        return CommonUtil.EMPTY_INT_ARRAY;
+    public void visitJavadocToken(DetailNode ast) {
+        final DetailAST blockCommentAst = getBlockCommentAst();
+        final DetailAST parentAst = getParentAst(blockCommentAst);
+
+        if (isTypeDefinition(parentAst) && shouldCheck(parentAst)) {
+            final List<DetailNode> blockTags = getBlockTags(ast);
+
+            if (!allowUnknownTags) {
+                checkUnknownTags(blockTags);
+            }
+
+            if (ScopeUtil.isOuterMostType(parentAst)) {
+                // don't check author/version for inner classes
+                checkTag(parentAst, blockTags, JavadocTagInfo.AUTHOR.getName(),
+                        authorFormat);
+                checkTag(parentAst, blockTags, JavadocTagInfo.VERSION.getName(),
+                        versionFormat);
+            }
+
+            final List<String> typeParamNames =
+                CheckUtil.getTypeParameterNames(parentAst);
+            final List<String> recordComponentNames =
+                getRecordComponentNames(parentAst);
+
+            if (!allowMissingParamTags) {
+
+                typeParamNames.forEach(typeParamName -> {
+                    checkTypeParamTag(parentAst, blockTags, typeParamName);
+                });
+
+                recordComponentNames.forEach(componentName -> {
+                    checkComponentParamTag(parentAst, blockTags, componentName);
+                });
+            }
+
+            checkUnusedParamTags(blockTags, typeParamNames, recordComponentNames);
+        }
     }
 
-    // suppress deprecation until https://github.com/checkstyle/checkstyle/issues/19146
-    @Override
-    @SuppressWarnings("deprecation")
-    public void visitToken(DetailAST ast) {
-        if (shouldCheck(ast)) {
-            final FileContents contents = getFileContents();
-            final int lineNo = ast.getLineNo();
-            final TextBlock textBlock = contents.getJavadocBefore(lineNo);
-            if (textBlock != null) {
-                final List<JavadocTag> tags = getJavadocTags(textBlock);
-                if (ScopeUtil.isOuterMostType(ast)) {
-                    // don't check author/version for inner classes
-                    checkTag(ast, tags, JavadocTagInfo.AUTHOR.getName(),
-                            authorFormat);
-                    checkTag(ast, tags, JavadocTagInfo.VERSION.getName(),
-                            versionFormat);
-                }
-
-                final List<String> typeParamNames =
-                    CheckUtil.getTypeParameterNames(ast);
-                final List<String> recordComponentNames =
-                    getRecordComponentNames(ast);
-
-                if (!allowMissingParamTags) {
-
-                    typeParamNames.forEach(typeParamName -> {
-                        checkTypeParamTag(ast, tags, typeParamName);
-                    });
-
-                    recordComponentNames.forEach(componentName -> {
-                        checkComponentParamTag(ast, tags, componentName);
-                    });
-                }
-
-                checkUnusedParamTags(tags, typeParamNames, recordComponentNames);
-            }
-        }
+    /**
+     * Checks if the given AST node is a type definition token.
+     *
+     * @param ast the AST node to check.
+     * @return true if the node is a type definition.
+     */
+    private static boolean isTypeDefinition(DetailAST ast) {
+        return ast.getType() == TokenTypes.INTERFACE_DEF
+                || ast.getType() == TokenTypes.CLASS_DEF
+                || ast.getType() == TokenTypes.ENUM_DEF
+                || ast.getType() == TokenTypes.ANNOTATION_DEF
+                || ast.getType() == TokenTypes.RECORD_DEF;
     }
 
     /**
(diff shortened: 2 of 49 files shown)


avadhutmali:
**CI failure explanation — `testAnnotationsInCodeBlock2`**

The single failing test is `testAnnotationsInCodeBlock2`. The failure is caused by `@unknown` inside a `<pre>` block:

```java
/**
 * Bare HTML pre tag does NOT protect content from parsing.
 * <pre>
 * @unknown
 * </pre>
 */
class BareHtmlPreTag {}
```

With the old `AbstractCheck` (regexp-based), `@unknown` was correctly reported as `Unknown tag 'unknown'` at line 28.

With `AbstractJavadocCheck` (AST-based), the `<pre>` block causes a parse error:
```
29: Javadoc comment at column 4 has parse error. Details: no viable alternative at input '</' while parsing HTML_ELEMENT
```

`visitJavadocToken` is never called for this comment — the AST parser fails on the malformed `</pre>` and emits its own error instead.

@romani , Waiting for guidance

avadhutmali:
ping @aclfe

aclfe:
@avadhutmali Sorry for the late reply, do it in this manner to prevent the parser throwing an error 
 
```
 * {@code
 * @unknown
 * }
 ```

That should fix the problem of exception being thrown

**Comment by avadhutmali** [comment]
> @avadhutmali Sorry for the late reply, do it in this manner to prevent the parser throwing an error
> 
> ```
>  * {@code
>  * @unknown
>  * }
> ```
> 
> That should fix the problem of exception being thrown

Hi @aclfe, thanks for the suggestion — it does resolve the parser error 👍

However, after switching to `{@code}`, the behavior changes. The parser now treats `@unknown` as part of an inline tag and produces:

JAVADOC_INLINE_TAG (code)
└── TEXT "@unknown"

Because of this, `@unknown` is no longer recognized as a block tag (`CUSTOM_BLOCK_TAG`), so `JavadocTypeCheck` does not report it as an unknown tag.

Previously (with `<pre>`), although parsing had issues, the expectation from the test is that `@unknown` should still be treated as a block-level tag and reported.

So the current situation is:

* Using `<pre>` → parser error prevents check execution
* Using `{@code}` → parser succeeds but suppresses tag detection

Could you suggest the intended approach here?
Should we:

1. Extend the check to detect unknown tags inside `<pre>` (using raw Javadoc text), or
2. Adjust the test expectation to align with the new parser behavior?

I’d appreciate your guidance on the expected direction.

File: src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheckTest.java
Comment (aclfe) [inline src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheckTest.java:1]: Looking at this file, we'll have to adjust the test expectation to align with the new parser behavior.
So it would probably just be 
/**
 * {@code
 * @unknown
 * }
 */

instead of the html `<pre>` setup

**Review by aclfe — COMMENTED** [review]
The whole point of migrating to AST is to adhere to it's new expectations. We'll have to change the test cases

File: src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheckTest.java
Comment (avadhutmali) [reply src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheckTest.java:1]: Thanks u so much for ur help and yes I agree about ur opinion about this test case
@romani what should we do about this test case

Comment (romani) [reply src/test/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheckTest.java:1]: if Input file is not working well for javadoc tool of jdk, we should update it to be compliant with javadoc tool.

File: config/pitest-suppressions/pitest-api-suppressions.xml [hunk 1]
```diff
@@ -1,3 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
+  <mutation unstable="false">
```
aclfe: Please kill mutation

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
```diff
@@ -293,95 +282,166 @@ private boolean shouldCheck(DetailAST ast) {
...
+    }
+
+    /**
+     * Gets the tag name from a block tag's specific child node
+     * (e.g., PARAM_BLOCK_TAG, AUTHOR_BLOCK_TAG, CUSTOM_BLOCK_TAG).
+     *
+     * @param tagNode the specific block tag child node.
+     * @return the tag name, or null if not found.
+     */
+    private static String getTagName(DetailNode tagNode) {
```
aclfe, line 351: Question, can we use JavadocUtil.getTagName? 
https://github.com/checkstyle/checkstyle/blob/9b3b81a2e44e22da1422920dda5171399b308330/src/main/java/com/puppycrawl/tools/checkstyle/utils/JavadocUtil.java#L307

config/suppressions.xml
```diff
@@ -129,19 +129,13 @@
...
   <suppress id="noUsageOfGetFileContentsMethod" files="JavadocMethodCheck.java"/>
-  <!-- until #19145 -->
   <suppress id="noUsageOfGetFileContentsMethod" files="JavadocStyleCheck.java"/>
-  <!-- until #19146 -->
-  <suppress id="noUsageOfGetFileContentsMethod" files="JavadocTypeCheck.java"/>
-  <!-- until #19147 -->
   <suppress id="noUsageOfGetFileContentsMethod" files="JavadocVariableCheck.java"/>
-  <!-- until #19148 -->
   <suppress id="noUsageOfGetFileContentsMethod" files="MissingJavadocMethodCheck.java"/>
-  <!-- until #19149 -->
```
aclfe, line 142: Please don't remove comments for other suppressions, it'll cause a merge conflict for you

aclfe (commented):
Direction is great! Please address javadoc mutations of CI: [mutations](https://github.com/checkstyle/checkstyle/actions/runs/23538244592/job/68518898234?pr=19189)

src/main/java/com/puppycrawl/tools/checkstyle/checks/javadoc/JavadocTypeCheck.java
avadhutmali: @aclfe Thanks for the suggestion! I looked into replacing this with JavadocUtil.getTagName(), but unfortunately, it causes NullPointerExceptions during the build.

The issue is that JavadocUtil.getTagName() relies on findFirstToken(..., JavadocCommentsTokenTypes.TAG_NAME). While CUSTOM_BLOCK_TAG nodes have a TAG_NAME child, standard tags like AUTHOR_BLOCK_TAG and VERSION_BLOCK_TAG do not (their name is implicit to their token type).

Because JavadocTypeCheck explicitly validates @ author and @ version tags, passing them to JavadocUtil.getTagName() results in a null token being evaluated for .getText(). I kept the local getTagName method so we can safely map the standard token types to their names via JavadocTagInfo while still falling back to TAG_NAME for custom tags. Let me know if you'd like me to approach this differently!

romani:
@avadhutmali  , please rebase to resolve conflict.

gianmarcoschifone:
@avadhutmali Please let's complete this PR

avadhutmali:
> @avadhutmali Please let's complete this PR

Yes start working as soon as college exams end

romani:
energy is moved to https://github.com/checkstyle/checkstyle/pull/20892

Issue #19176, opened 2026-03-11
Migrate remaining test files to use getExpectedThrowable instead of fail()


ayushactiveat:
we recently introduced a new Checkstyle rule in pr #19168 that forbids using the old `assertWithMessage(...).fail()` pattern inside `try-catch` blocks. Instead, we want to use the cleaner `getExpectedThrowable` utility.

We already updated a few packages, but there are still around 240 test files across the codebase that need to be migrated. 

how to find the files that need updating->
you can get a list of every file and line number that needs work by running -` grep -rn "assertWithMessage.*fail()"`

So, what needs to be changed-
you will see old code that looks like->
```
try {
    someMethodThatThrowsException();
    assertWithMessage("Exception is expected").fail();
}
catch (IllegalArgumentException ex) {
    assertWithMessage("Invalid exception message")
        .that(ex.getMessage())
        .isEqualTo("expected message");
}
```
update it to look like - 

```
final IllegalArgumentException ex = getExpectedThrowable(IllegalArgumentException.class, () -> {
    someMethodThatThrowsException();
});
assertWithMessage("Invalid exception message")
    .that(ex.getMessage())
    .isEqualTo("expected message");
```
also this is a large task, please do not try to do all 240+ files in one pr
pick a specific package/small group of files (5-10 file).

comment below "i am working on xxxx files/package"  so others don't work on the same ones

after the fix to ensure your changes are actually tested by the new rule, you must update the suppression list. Open config/suppressions.xml and find the MatchXpath suppression. It's a one liner using regex:
`files=".*[\\/]...[\\/](?!utils|xpath).*"`
so if you fixed the api package, you would update that line to:
`files=".*[\\/]...[\\/](?!utils|xpath|api).*"`

example of this migration - #19168

achyuth8055:
hi im working on the following test files in the `api` package to use `getExpectedThrowable`:

* ScopeTest.java
* SeverityLevelTest.java
* SeverityLevelCounterTest.java
* TokenTypesTest.java
* LineColumnTest.java

im on it ill submit my PRSoon

achyuth8055:
I previously worked on the following test files in the `api` package and submitted a PR for them:

- ScopeTest.java
- SeverityLevelTest.java
- SeverityLevelCounterTest.java
- TokenTypesTest.java
- LineColumnTest.java

Initially I selected 5 files, but `TokenTypesTest.java` and `LineColumnTest.java` already had the updated pattern and did not require changes.

[...]

PR #19201, opened 2026-03-12
Issue #19176: migrate whitespace tests to getExpectedThrowable

pirzada-ahmadfaraz:
Issue: #19176

This PR covers the first half of the whitespace batch: SeparatorWrap, ParenPad, OperatorWrap, and NoWhitespaceAfter tests now use `getExpectedThrowable(...)` instead of the `try/catch` + `fail()` pattern while keeping the original message/cause verifications.

Validation:
- `./mvnw -e --no-transfer-progress test -Dtest="SeparatorWrapCheckTest,ParenPadCheckTest,OperatorWrapCheckTest,NoWhitespaceAfterCheckTest"` (fails before the test suite starts because `org.jacoco.agent.rt.internal_29a6edd.Offline` cannot be loaded by the exec plugin; same Jacoco classpath issue we’ve seen in previous batches).

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/SarifLoggerTest.java
@@ -400,15 +400,16 @@ public class SarifLoggerTest extends AbstractModuleTestSupport {
     }
 
     @Test
-    public void testFinishLocalSetup() throws IOException {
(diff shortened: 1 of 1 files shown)


romani:
ci is red , only cirlce ci failures are expected unrelated problem.
Please do not rush with updates, show that you can deliver updates without to-and-from with maintaiers

PR #19215, opened 2026-03-13
Issue #19176: migrate metrics tests to getExpectedThrowable

### LINKED ISSUE #19176: Migrate remaining test files to use getExpectedThrowable instead of fail()

achyuth8055:
Issue #19176: migrate coding tests to getExpectedThrowable

Migrated 3 test files in the checks/metrics package from the old
assertWithMessage(...).fail() try-catch pattern to the new
getExpectedThrowable utility method.

Files migrated:
- BooleanExpressionComplexityCheckTest.java (1 try-catch)
- ClassFanOutComplexityCheckTest.java (1 try-catch)
- ClassDataAbstractionCouplingCheckTest.java (2 try-catches)

Also updated config/suppressions.xml to add checks/metrics to the
MatchXpathForbidTryCatchFail exclusion regex, and removed redundant
throws Exception from methods that no longer throw checked exceptions.

Resolves #19176
Screenshot 2026-03-13 at 3 24 23 AM

### DIFF
--- a/config/suppressions.xml
+++ b/config/suppressions.xml
@@ -176,6 +176,6 @@
   <!-- Remove suppression once all violations are fixed, tracked in #19176 -->
   <suppress checks="MatchXpath"
             id="MatchXpathForbidTryCatchFail"
-            files=".*[\\/]src[\\/]test[\\/]java[\\/]com[\\/]puppycrawl[\\/]tools[\\/]checkstyle[\\/](?!utils|xpath|checks[\\/]coding|checks[\\/]design|checks[\\/]modifier|checks[\\/]naming|checks[\\/]annotation|checks[\\/]indentation).*"/>
+            files=".*[\\/]src[\\/]test[\\/]java[\\/]com[\\/]puppycrawl[\\/]tools[\\/]checkstyle[\\/](?!utils|xpath|checks[\\/]coding|checks[\\/]design|checks[\\/]modifier|checks[\\/]naming|checks[\\/]annotation|checks[\\/]indentation|checks[\\/]metrics).*"/>
 
 </suppressions>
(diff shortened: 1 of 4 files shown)


romani:
conflict

Issue #19234, opened 2026-03-15
Stable failure of several circleci jobs


romani:
stable failure in CircleCi:

./.ci/validation.sh package-site
./.ci/validation.sh spotbugs-and-pmd
./.ci/validation.sh no-error-pgjdbc
./.ci/validation.sh no-exception-Pmd-elasticsearch-lombok-ast
./.ci/validation.sh no-exception-hbase

in all executions, jobs are killed by circleci

at https://github.com/checkstyle/checkstyle/pull/19233
I tried to increase instances to medium+ , and use older jdk21 versions - not helping.

upgrade to large instance: memory is not used up to 100%, [kind of 50%](https://app.circleci.com/pipelines/github/checkstyle/checkstyle/43553/workflows/2077c9c4-f196-4ed8-9953-75fe209a42c0/jobs/1410411/resources) and still same crash of execution.

romani:
Same for `./.ci/validation.sh no-error-pmd`

https://app.circleci.com/pipelines/github/checkstyle/checkstyle/43571/workflows/15bb6a66-4622-4816-93b4-21e8033ecc10/jobs/1411206

aclfe:
@romani I've sent this PR for the issue: https://github.com/checkstyle/checkstyle/pull/19239 re-enabling all the circle CI jobs and making the appropriate fix based on your diagnosis

romani:
Fixed by @aclfe

PR #19239, opened 2026-03-15
Issue #19234: Resolving Circle CI failures 

aclfe:
Issue #19234: Resolving Circle CI failures

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -238,6 +238,7 @@ EOF
   ;;
 
 no-error-pmd)
+  export MAVEN_OPTS="-XX:MaxRAMPercentage=75"
   CS_POM_VERSION="$(getCheckstylePomVersion)"
   echo "CS_version: ${CS_POM_VERSION}"
   ./mvnw -e --no-transfer-progress clean install -Pno-validations
@@ -517,7 +518,7 @@ checkstyle-and-sevntu)
 spotbugs-and-pmd)
   mkdir -p .ci-temp/spotbugs-and-pmd
   CHECKSTYLE_DIR=$(pwd)
-  export MAVEN_OPTS='-Xmx2g'
+  export MAVEN_OPTS="-Xmx4g"
   ./mvnw -e --no-transfer-progress clean pmd:check
   ./mvnw -e --no-transfer-progress clean test-compile spotbugs:check
   cd .ci-temp/spotbugs-and-pmd
@@ -771,6 +772,7 @@ javac25)
   ;;
 
 package-site)
+  export MAVEN_OPTS="-Xmx5g"
   ./mvnw -e --no-transfer-progress package -Passembly,no-validations
   ./mvnw -e --no-transfer-progress site -Dlinkcheck.skip=true
   ;;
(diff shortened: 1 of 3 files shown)


aclfe:
@romani

.ci/validation.sh
```diff
@@ -517,7 +517,7 @@ checkstyle-and-sevntu)
 spotbugs-and-pmd)
   mkdir -p .ci-temp/spotbugs-and-pmd
   CHECKSTYLE_DIR=$(pwd)
-  export MAVEN_OPTS='-Xmx2g'
+  export MAVEN_OPTS='-Xmx6g'
```
romani: Can we try 3 or 4g ?
Jump from  2 to 6 is crazy.

.circleci/config.yml
```diff
@@ -204,11 +173,61 @@ workflows:
...
-      #    image-name: cimg/openjdk:21.0
-      #    command: ./.ci/validation.sh no-error-pgjdbc
+      - validate-with-maven-script:
+          name: no-error-hazelcast
+          image-name: *cs_img
+          command: ./.ci/validation.sh no-error-hazelcast
+      - validate-with-maven-script:
+          name: no-error-pmd
+          image-name: *cs_img
+          resource_class: large
```
romani: Can we try medium+

.circleci/config.yml
```diff
@@ -204,11 +173,61 @@ workflows:
...
+          command: |
+            export MAVEN_OPTS="-Xmx6g -XX:MaxRAMPercentage=75"
+            ./.ci/validation.sh no-error-pmd
```
romani: Please move all opts variables inside shell script.

romani:
Please enable job at place it was commented out.

And please restore sonarqube commented block.

aclfe:
> Please enable job at place it was commented out.
> 
> And please restore sonarqube commented block.

I'll make all the changes

aclfe:
@romani 

Ci failure in circle CI jobs is happening due to the recent check merged: `Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.12.1:site (default-site) on project sample: Error generating maven-checkstyle-plugin:3.1.1:checkstyle report: Failed during checkstyle configuration: cannot initialize module TreeWalker - cannot initialize module IllegalSymbol:`

romani:
please rebase, I merged new modules 1 hour ago.

.ci/validation.sh
```diff
@@ -238,6 +238,7 @@ EOF
   ;;
 
 no-error-pmd)
+  export MAVEN_OPTS="-XX:MaxRAMPercentage=75"
```
romani, line 241: will it help if we make 90% and stay on smaller ci instances ?
we do not run anything else on CI boxes, so 10% from 4GB, might be ok for linux kernel to run.

aclfe: I don't think it should be making a huge difference. before the hard coded memory of 6 gigs was working wonders, but that's not a good approach long term

aclfe: but that was only because we put large as the resource class

romani:
please apply spelling patch

aclfe:
> will merge with pleasure, after CI pass

I pushed the whitelist words. 
Thank you for iterating with me quickly

no-error-pgjdbc in particular is very flaky, but with the current settings it's stably green continuously so it should be addressed for a while

romani:
we will merge this awesome update, and will followup if required or some becomes annoying flacky.

Issue #19063, opened 2026-02-28
Trailing comment alignment test fails with tab-based indentation in test input files


MiniPiku:
I have read check documentation: https://checkstyle.org/checks/xxxxxx/nameofaffectedcheck.html
I have downloaded the latest checkstyle from: https://checkstyle.org/cmdline.html#Download_and_Run

### Description

When adding a tab-based regression test file for `IndentationCheck`,
the JUnit test `IndentationTrailingCommentsVerticalAlignmentTest` fails due to trailing comment alignment mismatch.

The regression input file intentionally contains tab characters to verify tab-based indentation handling.

However, `IndentationTrailingCommentsVerticalAlignmentTest` enforces that all trailing comments (`//indent:... exp:...`) must align at the same expanded column across the file.

This causes failures for files that intentionally use tabs.

### How to Reproduce

Add a regression input file containing tabs in:
`
src/test/resources/com/puppycrawl/tools/checkstyle/checks/indentation/indentation/
`

**ATTENTION** : should be tabs and not spaces(IDE'sare tend to convert tabs to spaces)

Run:
`
mvn -Dtest=IndentationTrailingCommentsVerticalAlignmentTest test
`
### Input
```
public class InputIndentationTryCtorParamsTabsWrong {                        //indent:0 exp:0

	private InputIndentationTryCtorParamsTabsWrong client;                   //indent:4 exp:4

	private InputIndentationTryCtorParamsTabsWrong(String string) {          //indent:4 exp:4
	}                                                                        //indent:4 exp:4

	private void test() {                                                    //indent:4 exp:4
		try {                                                                //indent:8 exp:8
			client =                                                         //indent:12 exp:12
            new InputIndentationTryCtorParamsTabsWrong(null);                //indent:8 exp:16
		}                                                                    //indent:8 exp:8
		catch (Exception e) {                                                //indent:8 exp:8
		}                                                                    //indent:8 exp:8
	}                                                                        //indent:4 exp:4
}
```
(Initialized gaps are tabs and not spaces to trigger a particular error for test)

### Failing Output
```
Trailing comment alignment mismatch in file:
InputIndentationTryCtorParamsTabsWrong.java on line 23
expected: 77
but was : 71
```
### Expected Behavior

Tab-based regression test files should not fail due to global trailing comment alignment enforcement.

### Actual Behavior

Trailing comment alignment validation fails even though the file is intentionally tab-indented for indentation testing.

### Suggested Direction

[...]

PR #19243, opened 2026-03-15
Issue #19063: Fix hardcoded tabWidth in IndentationTrailingCommentsVerticalAlignmentTest

MiniPiku:
Closes: Issue #19063

### Problem:
IndentationTrailingCommentsVerticalAlignmentTest used a hardcoded tabWidth = 4
when calculating expanded column positions for trailing comment alignment validation.
This caused false failures for tab-based regression test files whose paired XML config
specifies a different tabWidth value.

### Solution:
Instead of hardcoding tabWidth, the test now reads the actual tabWidth from the
IndentationCheck instance configured via its paired XML config file, reusing the
same property that TreeWalker sets via AbstractCheck.setTabWidth() at runtime.
If no config is found, it falls back to 4 to preserve existing behaviour.

The fix introduces three helper methods:
- `configFileContainsFileName()` - checks if an XML config references a given test file
- `findConfigForFile()` - locates the paired XML config for a given test input file
- `extractTabWidthFromConfig()` - reads tabWidth from a real IndentationCheck instance
- `getTabWidth()` - wires the above together to return the correct tabWidth per file

### DIFF
--- a/src/test/java/com/puppycrawl/tools/checkstyle/IndentationTrailingCommentsVerticalAlignmentTest.java
+++ b/src/test/java/com/puppycrawl/tools/checkstyle/IndentationTrailingCommentsVerticalAlignmentTest.java
@@ -25,12 +25,15 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Stream;
 
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
 
+import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
+import com.puppycrawl.tools.checkstyle.api.Configuration;
+import com.puppycrawl.tools.checkstyle.checks.indentation.IndentationCheck;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 
 public class IndentationTrailingCommentsVerticalAlignmentTest {
@@ -38,17 +41,10 @@ public class IndentationTrailingCommentsVerticalAlignmentTest {
     private static final String INDENTATION_TEST_FILES_PATH =
         "com/puppycrawl/tools/checkstyle/checks/indentation/indentation";
 
-    private int tabWidth;
-
-    @BeforeEach
-    public void setUp() {
-        tabWidth = 4;
-    }
-
     @MethodSource("indentationTestFiles")
     @ParameterizedTest
-    public final void testTrailingCommentsAlignment(Path testFile) throws IOException {
-        final int currentTabWidth = tabWidth;
+    public final void testTrailingCommentsAlignment(Path testFile) throws Exception {
+        final int currentTabWidth = getTabWidth(testFile);
         final List<String> lines = Files.readAllLines(testFile);
         int expectedStartIndex = -1;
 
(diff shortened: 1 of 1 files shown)


MiniPiku:
Hi @romani,

While working on this fix I noticed that some input files do not have a `tabWidth`
property specified in their paired XML configs, causing the test to fall back to 
a hardcoded value of `4`:

- InputIndentationUseTabs.java
- InputIndentationLambda3.java
- InputIndentationLambda4.java
- InputIndentationLambda5.java
- InputIndentationAnonymousClassInMethod.java
- InputIndentationValidMethodIndent.java
- InputIndentationCustomAnnotation1.java
- InputIndentationInvalidSwitchIndent.java
- InputIndentationNewWithForceStrictCondition.java

I intentionally did not modify the corresponding tests/XML files as part of this PR to keep the scope focused and the merge history clean.

If you agree, I can open a separate issue to track adding explicit `tabWidth` values to these configs so the hardcoded fallback can eventually be removed.

romani:
Please fix CI

src/test/java/com/puppycrawl/tools/checkstyle/IndentationTrailingCommentsVerticalAlignmentTest.java
```diff
@@ -25,30 +25,83 @@
...
+import com.puppycrawl.tools.checkstyle.checks.indentation.IndentationCheck;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtil;
 
 public class IndentationTrailingCommentsVerticalAlignmentTest {
 
     private static final String INDENTATION_TEST_FILES_PATH =
         "com/puppycrawl/tools/checkstyle/checks/indentation/indentation";
 
-    private int tabWidth;
+    private static boolean doesconfigFileContainsFileName(Path configPath, String fileName) {
```
romani: doesConfigFileContainsFileName(Path ....

src/test/java/com/puppycrawl/tools/checkstyle/IndentationTrailingCommentsVerticalAlignmentTest.java
```diff
@@ -25,30 +25,83 @@
...
+                .filter(path -> doesconfigFileContainsFileName(path, fileName))
+                .findFirst();
+        }
+        return result;
+    }
 
-    @BeforeEach
-    public void setUp() {
-        tabWidth = 4;
+    private static int extractTabWidthFromConfig(Configuration config) throws CheckstyleException {
```
romani: Move this method below getTabWidth

romani: All other new methods should be placed in code after first usage, move them down

romani (changes requested):
Items

src/test/java/com/puppycrawl/tools/checkstyle/IndentationTrailingCommentsVerticalAlignmentTest.java
MiniPiku: ``

> Move this method below getTabWidth

done

MiniPiku: > All other new methods should be placed in code after first usage, move them down

done

src/test/java/com/puppycrawl/tools/checkstyle/IndentationTrailingCommentsVerticalAlignmentTest.java
MiniPiku: done

Issue #13500, opened 2023-08-01
Fix abstraction failure in SiteUtil::getCheckMessageKeys


stoyanK7:
From https://github.com/checkstyle/checkstyle/pull/13423#discussion_r1280844252

We suppressed an `InstanceofChain` IDEA error in `SiteUtil::getCheckMessageKeys`. The violation comes from these 2 if-statements
```java
    if (module == RegexpMultilineCheck.class) {
        checkstyleMessageKeys.addAll(getCheckMessageKeys(
                Class.forName(
                        "com.puppycrawl.tools.checkstyle.checks.regexp.MultilineDetector"
                ), checkInstance)
        );
    }
    else if (module == RegexpSinglelineCheck.class
            || module == RegexpSinglelineJavaCheck.class) {
        checkstyleMessageKeys.addAll(getCheckMessageKeys(
                Class.forName(
                        "com.puppycrawl.tools.checkstyle.checks.regexp.SinglelineDetector"
                ), checkInstance)
        );
    }
```

We need to decide what to do with this.

> And this violation is right, it is hack.
We should have separate ticket to improve it by refactoring of modules to have all MSG fields in module as field, all child, nested, composited classes should take by constructor content of message.
We already discussed it multiple times, and now we see real need for this

Carbon14-48:
Hi, I'd like to work on this issue. I'll move the message keys from the detector classes to the main checks and remove the hack from SiteUtil.

romani:
fix is merged

PR #19263, opened 2026-03-17
Issue #13500: Move regexp message keys from detector classes to main checks

Carbon14-48:
## Summary

This PR addresses [Issue #13500](https://github.com/checkstyle/checkstyle/issues/13500), fixing architectural and historical issues for regexp checks' violation messages.

---

## Problem and Bug

Previously, violation message keys (`MSG_*`) for regexp modules like `RegexpMultilineCheck`, `RegexpSinglelineCheck`, and `RegexpSinglelineJavaCheck` were defined in their detector classes.  
This led to:
- Brittle hacks in SiteUtil to locate these keys for doc generation.
- Bugs in tests (e.g., `RegexpSinglelineJavaCheckTest` imported keys from the wrong detector, such as `MultilineDetector`).
- Suppression comments/annotations (e.g., `@noinspection` and `@noinspectionreason`) and confusing exception handling to justify the workaround.

---

## What Was Changed

- **Message keys moved:** All regexp check message keys now live in the check classes and are referenced directly.
- **Fixed test bugs:** Corrected hidden bug where `RegexpSinglelineJavaCheckTest` was incorrectly importing from `MultilineDetector` instead of its module's message keys.
- **Removed hacks:** SiteUtil no longer includes special-case blocks to scan detector classes for message keys.
- **Suppressions/annotations cleaned:** All tied suppression comments, `@noinspection`, `@noinspectionreason`, and unreachable exception handling removed.
- **Note on SinglelineDetector:** Since both `RegexpSinglelineCheck` and `RegexpSinglelineJavaCheck` use the same detector, I duplicated the message keys in both check classes. This ensures each check independently owns its message keys while maintaining backward compatibility.

---

## Validation

Ran `mvn clean verify` and `mvn clean site`; all code, tests, and docs are generated correctly.


src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/SinglelineDetector.java
```diff
@@ -80,7 +68,7 @@ public void processLines(FileText fileText) {
     private void finish() {
         if (currentMatches < options.getMinimum()) {
             if (options.getMessage().isEmpty()) {
-                options.getReporter().log(1, MSG_REGEXP_MINIMUM,
+                options.getReporter().log(1, RegexpSinglelineCheck.MSG_REGEXP_MINIMUM,
                         options.getMinimum(), options.getFormat());
```
romani, line 77: please update SinglelineDetector.java class to have 2 fields (`String exceededMessage;` ....) for message key, add ctor that takes as parameters such keys from child classes. So each child class will put there what it wants.
 
one more time:
src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/SinglelineDetector.java should not depend on any XXXXXXCheck.MSG_XXXXXX_XXXX

romani (changes requested):
Sorry for 4 month of no response.
in general good, but some corrections are required:

src/main/java/com/puppycrawl/tools/checkstyle/checks/regexp/SinglelineDetector.java
Carbon14-48: done

romani (approved):
thanks a lot

PR #19280, opened 2026-03-17
Issue #4997: Column number in DetailAST should start with 1

### LINKED ISSUE #4997: Column number in DetailAST should start with 1
Due to the fact that in the past in many IDEs and code editors the position of the first symbol in the line was 0, we have [this hack](https://github.com/checkstyle/checkstyle/blob/cc2886ef2ce994219bdd2427f820a7ac3738916b/src/main/java/com/puppycrawl/tools/checkstyle/api/DetailAST.java#L68) in DetailAST:

```java
// expect columns to start @ 0
columnNo = tok.getColumn() - 1;
```

Nowadays, in most popular IDEs (IntelliJ IDEA, Eclipse, sublime, VS Code, gedit) the first symbol has 1st position. The log method of AbstractCheck has a special hack to avoid confusion in the Checkstyle output:

```java
final int col = 1 + CommonUtils.lengthExpandedTabs(
            getLines()[lineNo - 1], colNo, tabWidth);
```

We always increment column number at list by 1. Tab expansion is another story, and it is OK.
That is why we need to consider that the first character starts on the 1st position too.

The fix will lead too the big regression. 

1) We need to remove ```-1``` from DetailAST
2) We need to remove ```+1``` from AbstractCheck#log
3) We should fix all UTs. Special attention to each failing UT, some checks violations may become confusing.
4) The regression report for all failing checks has to be generated to prove that at least the number of violations will not be decreased. So the number of violations before and after fix should be equal.

ayushactiveat:
fixes #4997
column numbers in `DetailAST` previously started at 0 , with a workaround in `AbstractCheck.log()` that added 
+1 before reporting violations, this was inconsistent with most modern ides/editor display col no. (1-based).

### DIFF
--- a/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
+++ b/config/checker-framework-suppressions/checker-nullness-optional-interning-suppressions.xml
@@ -5635,7 +5635,7 @@
     <fileName>src/main/java/com/puppycrawl/tools/checkstyle/checks/indentation/MethodCallHandler.java</fileName>
     <specifier>argument</specifier>
     <message>incompatible argument for parameter ast of AbstractExpressionHandler.getLineStart.</message>
-    <lineContent>if (getLineStart(rparen) == rparen.getColumnNo()) {</lineContent>
+    <lineContent>if (getLineStart(rparen) == rparen.getColumnNo() - 1) {</lineContent>
     <details>
       found   : @Initialized @Nullable DetailAST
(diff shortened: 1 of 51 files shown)


ayushactiveat:
hi @romani , ci fails seems unrelated, please let me know if u need any change

mahfouz72:
@ayushactiveat Hey, sorry for late response, can you resolve conflicts and continue this PR

Issue #19064, opened 2026-03-01
XpathRegressionXxxxxxXxxxxxTest should have 3 test methods


romani:
we have rule to have 3 tests in XpathRegressionXxxxxxXxxxxxxTest
But as it is not automated we always loose time on this.

examples of missed: 
https://github.com/checkstyle/checkstyle/pull/18657#discussion_r2859017125
https://github.com/checkstyle/checkstyle/pull/18312#issuecomment-3967872333

Find a way to make junit test or Checkstyle Xpath test to make sure that all  
under https://github.com/checkstyle/checkstyle/tree/master/src/it/java/org/checkstyle/suppressionxpathfilter
should 3 methods with `@Test`  and in error message we should mention that we need 3 test that generates xpath different by nodes (structure).

Demand for 3 tests were added at https://github.com/checkstyle/checkstyle/pull/19074

Goal of this issue is to remove suppressions at 
https://github.com/checkstyle/checkstyle/blob/4fdd0fc8ecaf50644660575248e73438b32ad771/config/checkstyle-non-main-files-suppressions.xml#L413

Each file should be removed in separate PR

**ATTENTION**: all test methods should generate Xpath with different structure, so make a diversity by placing violation in inner classes, more nested code blocks, in anonymous classes, in .....

Aman-Baliyan:
I am on it

Aman-Baliyan:
There is a test present which checks for the number of test cases in XpathRegressionXxxxxxXxxxxxxTest but its minimum limit is set to 2 and there are almost 80 files which contains 2 tests . so i am adding a new test in that files

DakshRJain737:
> There is a test present which checks for the number of test cases in XpathRegressionXxxxxxXxxxxxxTest but its minimum limit is set to 2 and there are almost 80 files which contains 2 tests . so i am adding a new test in that files

Are you adding tests to 80 files in one PR?
I think it can be broken into multiple PRs.

Aman-Baliyan:
> > There is a test present which checks for the number of test cases in XpathRegressionXxxxxxXxxxxxxTest but its minimum limit is set to 2 and there are almost 80 files which contains 2 tests . so i am adding a new test in that files
> 
> Are you adding tests to 80 files in one PR? I think it can be broken into multiple PRs.

this is totally depend on maintainers if they agree then we can split it in multiple pr

Aman-Baliyan:
@romani is there a need to solve this issue in multiple pr

[...]

PR #19301, opened 2026-03-18
Issue #19064: Add 3rd test to XpathRegressionWhenShouldBeUsedTest

achyuth8055:
Contributes to #19064
Added `testStaticInit()` as the 3rd test method to `XpathRegressionWhenShouldBeUsedTest`. The new test uses a pattern variable in a switch statement inside a static initializer block, differentiating from the existing tests which use violations inside methods `testSimple` and `testNested` use `METHOD_DEF`. The new test goes through `STATIC_INIT`.


romani:
conflict

PR #19322, opened 2026-03-18
Issue #15456: Replace shorthand violation comments in InputParameterNameOverrideAnnotation

surajgojanur:
Fixes part of #15456

I updated violation comments in `InputParameterNameOverrideAnnotation.java` to use explicit messages instead of shorthand comments.

Replaced shorthand comments like:
- `// violation`
- `// 2 violations`

with explicit messages in the existing BDD format.

This makes the expected violations clearer and keeps the test input consistent with the format already used in other checks.

Only the relevant lines in this file were changed.


surajgojanur:
Updated this branch with the same line-length fix (grouped violations above format).

./mvnw clean verify passes locally.

Single clean commit preserved.
Please let me know if any changes are required.

PR #19323, opened 2026-03-18
Issue #15456: Replace shorthand violation comments in InputParameterNameOneMethods

surajgojanur:
Fixes part of #15456

I updated violation comments in `InputParameterNameOneMethods.java` to use explicit messages instead of shorthand comments.

Replaced shorthand comments like:
- `// violation`
- `// 2 violations`

with explicit messages in the existing BDD format.

This makes the expected violations clearer and keeps the test input consistent with the format already used in other checks.

Only the relevant lines in this file were changed.


surajgojanur:
Resolved line-length issue by using grouped violations above format.
Local verify passes and PR remains a single clean commit.

Please let me know if anything else is needed

romani (approved):
Please change all for check in single PR

PR #19331, opened 2026-03-18
Issue #19064: Add 3rd test for XpathRegressionJavadocVariableTest

achyuth8055:
Issue #19064: Add 3rd test for XpathRegressionJavadocVariableTest

Add testEnum() method that tests a private field without Javadoc inside
an enum. The existing tests cover a private field in an outer class
(testPrivateClassFields) and a public field in an inner class
(testInnerClassFields). The new test uses a different AST path through
ENUM_DEF/OBJBLOCK/VARIABLE_DEF.


romani:
Conflict

PR #19333, opened 2026-03-18
Issue #19064: Add 3rd test for XpathRegressionNPathComplexityTest

achyuth8055:
Issue #19064: Add 3rd test for XpathRegressionNPathComplexityTest

Add testConstructor() method that tests NPath complexity in a constructor
with an if-else block. The existing tests cover a method (testMethod) and
a static initializer block (testStaticBlock). The new test uses a
different AST path through CTOR_DEF


romani:
Conflict

romani:
Conflict.
Do not change suppression until final PR

